PR #4, opened 2014-07-10
Add support for vmware_fusion provider

hiremaga:
- The old precise64 box does not work with vmware_fusion
- The new one is from vagrantcloud.com and should work with both providers, it also does not need an explicit box url

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,5 @@
 Vagrant.configure("2") do |config|
-  config.vm.box = "precise64"
-  config.vm.box_url = "http://files.vagrantup.com/precise64.box"
+  config.vm.box = "hashicorp/precise64"
 
   config.vm.network "forwarded_port", guest: 8080, host: 8080 # atc
   config.vm.network "forwarded_port", guest: 5637, host: 5637 # glider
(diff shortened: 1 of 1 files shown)


wchrisjohnson:
How is the box used with the vmware/fusion provider? I'm attempting to get concourse/lite going to use this tutorial (https://github.com/starkandwayne/concourse-tutorial) and when I try to `vagrant up`, either with or without the provider attribute, it fails like so:

The box you're attempting to add doesn't support the provider
you requested. Please find an alternate box or use an alternate
provider. Double-check your requested provider to verify you didn't
simply misspell it.

If you're adding a box from HashiCorp's Atlas, make sure the box is
released.

Name: concourse/lite
Address: https://atlas.hashicorp.com/concourse/lite
Requested provider: ["vmware_desktop", "vmware_fusion", "vmware_workstation"]

vito:
We no longer build VMware Fusion images.

wchrisjohnson:
Thanks @vito - that I understand. Is it possible for me to use https://github.com/concourse/concourse-lite and the the vmware.json template, to build an image myself using packer, etc. ? I'm part way down that path but can't seem to get packer-bosh to install... 

I'm not expecting a full set of instructions, just whether or not what I need is in that repo to build a vmware box. This isn't my primary area of expertise...

vito:
you may want to use both that and the virtual box template as a reference.
with 1.0 we switched from packer-bosh to simply embedding the binaries,
which should make for a simpler building process.

wchrisjohnson:
Thanks @vito

PR #14, opened 2015-02-18
Bump garden linux to 0.117.0

hsiliev:
New installation via vagrant-bosh plugin using garden linux 0.34 fails to run Diego Inigo tests with http response 500.


vito:
thanks!

PR #15, opened 2015-02-19
move gate repository url to https, as with all other submodule urls

avade:


vito:
Thanks! Mind submitting this against the `develop` branch instead?

PR #34, opened 2015-05-13
Use c4.large instances types for compilation.

robdimsdale:
- instead of c3.
- use 30GB of ephemeral disk.


zachgersh:
@robdimsdale just a point of clarification - why is the ephemeral disk being included?

robdimsdale:
@zachgersh c4.x instance types do not have any ephemeral disk - we have to bind our own if we need it (which we do)

zachgersh:
LGTM then - @vito ok to merge?

vito:
Yep

PR #50, opened 2015-05-21
Update fly install instructions

mooingcat:
Now that no-pipeline screen exists, the binaries are in the centre, not bottom-right.


vito:
good point. thanks!

PR #51, opened 2015-05-21
Add pipeline config for redeploying concourse.

robdimsdale:
[#94528472]

### DIFF
--- a/ci/pipelines/concourse-redeploy.yml
+++ b/ci/pipelines/concourse-redeploy.yml
@@ -1,0 +1,62 @@
+groups:
+- name: concourse
+  jobs:
+  - deploy-concourse
+
+resources:
+- name: concourse-config
+  type: git
+  source:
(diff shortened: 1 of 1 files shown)


vito:
Awesome thanks! A few notes:

* Does the second `put` actually work? I thought we didn't respect `on: [failure]` if the previous step errored (which is how `put`s fail).
* Can we make the params be in the format of `foo-bar` for consistency with the other pipelines?
* Can probably remove the pipeline group since there's only one job in the whole thing.
* We should probably make the stemcell more generic so people can use this on vsphere to. So I think we can just rename `aws-stemcell` to `stemcell` and make the stemcell name parameterizable.

robdimsdale:
Makes sense. I made the changes above.

i also opened #52 to enable retrying from tasks that have exited with error.

vito:
perfect, thanks!

drnic:
Thanks Rob!

PR #57, opened 2015-06-10
change garden-linux version in bosh-lite.yml

elenasharma:
v0.205.0 is the latest working version; latest does not work.

### DIFF
--- a/manifests/bosh-lite.yml
+++ b/manifests/bosh-lite.yml
@@ -7,7 +7,7 @@ releases:
   - name: concourse
     version: latest
   - name: garden-linux
-    version: latest
+    version: 0.205.0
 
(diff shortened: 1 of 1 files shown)


vito:
Thanks! I think the cost of keeping this up-to-date outweighs the benefits though. This manifest is also used by developers, where we actually do want `latest` (since we want it to always use the releases we've just uploaded). And `latest` is correct so long as you've uploaded the correct versions before deploying (and nothing more recent).

I've updated the docs to suggest a better workflow:

http://concourse.ci/deploying-with-bosh.html#upload-the-concourse-and-garden-releases

elenasharma:
That makes sense. Thanks!

PR #59, opened 2015-06-13
Add vCloud example manifest

krishicks:


vito:
awesome, thanks. would you mind changing the release versions to `latest` to match the rest? they say latest since they're often used for local development where the user is uploading their own dev releases. (it's just expected that the versions they've uploaded will be compatible.)

the bosh-deployment resource will lock the versions down to the versions from the inputs, so if someone were to copy-paste this and use it in a pipeline it'd work fine too.

krishicks:
Updated.

vito:
thanks!

PR #66, opened 2015-06-16
Add KeyName parameter support for selecting an existing keypair

MarkKropf:
This is nice, since the hard coded bosh keypair isn't always desired


vito:
Thanks! Mind adding `"Default": "bosh"` to the param for convenience?

MarkKropf:
done @vito

vito:
Sweet, thanks!

PR #202, opened 2015-11-12
Add timeouts to SSH tunnel so we can reconnect when it fails.

mkb:
After conversation with @vito yesterday we added timeouts so that workers can reconnect when their connection fails.


mkb:
Closing as this is to the wrong branch. Sorry for the noise.

PR #253, opened 2016-01-13
Restrict access to InternalSecurityGroup.

simonjjones:
As discussed on Slack (https://concourseci.slack.com/archives/general/p1452624733008350) we would like to tighten our security group rules for the internal security group. Further to the discussion, it appears we needed to add access from the WebSecurityGroup as well as those mentioned.

Allow only the following sources to connect to InternalSecurityGroup on tcp and udp: InternalSecurityGroup, WebSecurityGroup, BOSHSecurityGroup.

The current configuration (InternalSecurityGroup is open to all sources on every port) causes amazon trusted advisor to raise a red warning.

Signed-off-by: Simon Jones <simon@cloudcredo.com>

### DIFF
--- a/manifests/cloudformation.json
+++ b/manifests/cloudformation.json
@@ -154,13 +154,13 @@
         "SecurityGroupEgress": [],
         "SecurityGroupIngress": [
           {
-            "CidrIp": "0.0.0.0/0",
+            "SourceSecurityGroupId": { "Ref": "WebSecurityGroup" },
             "IpProtocol": "tcp",
             "FromPort": "0",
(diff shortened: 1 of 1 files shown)


vito:
This looks good, thanks! I'm assuming you've tried it yourself?

simonjjones:
@vito We have yes, we successfully updated our existing cloud formation using this template, our concourse remains completely functional & in doing so we have removed the existing alert in Trusted Advisor.

xoebus:
Thanks!

Merged manually.

PR #265, opened 2016-01-21
Add support for newer AWS region on CloudFormation template

premist:
#### New region support

I modified CloudFormation template to support newer AWS region. I added support for two missing regions:

- EU (Frankfurt) region
- Asia Pacific (Seoul) region

#### Change NAT AMIs to use HVM virtualization type 

I changed NAT AMIs to use HVM Virtualization Types, since PV(Paravirtual) virtualization type is not supported in new regions.

#### Change instance type

I changed instance type to `m4.large` since M3 type is not supported in new regions. Since it's almost twice expensive compared to existing `m3.medium`, I thought about using cheaper instance type like `t2.medium`, but since it doesn't have a consistent performance I decided to use standard M4 instance type instead. Feel free to modify it to another instance type.

### DIFF
--- a/manifests/cloudformation.json
+++ b/manifests/cloudformation.json
@@ -63,6 +63,7 @@
       "ap-southeast-1": { "AMI": "ami-f22772a0" },
       "ap-southeast-2": { "AMI": "ami-3bae3201" },
       "ap-northeast-1": { "AMI": "ami-cd43d9cc" },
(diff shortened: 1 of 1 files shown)


xoebus:
Thanks!

xoebus:
Manually merged onto `develop`.

PR #275, opened 2016-02-10
Update concourse docs to include ATC_EXTERNAL_URL

zachgersh:
[#11346088]

Signed-off-by: Utako Ueda <uueda@pivotal.io>


vito:
thanks!

PR #290, opened 2016-02-17
`image_resource` `tag` should be a string

afeld:
When running the provided example using Concourse v0.72.1, I got the following error:

```
resource script '/opt/resource/check []' failed: exit status 1

stderr:
failed to read request: json: cannot unmarshal number into Go value of type string

errored
```

This changes the value in the example YAML to be a string rather than a number, which fixes the issue.


xoebus:
Thanks!

PR #296, opened 2016-02-25
Make 'attempts' doc a little clearer

benjamb:
Clearer to me at least.

(As discussed in slack) I was a little confused by the use of the word "retry" in the docs, as that seemed to imply the number of attempts was the number of times after the first attempt it would be retried, however this wasn't the case and it is intended to be the total number of times the step will be tried (with me so far? :) ).

Reword things in an attempt (heh) to make things clearer.


vito:
Looks good, thanks!

PR #304, opened 2016-03-02
point pivnet_resource link to docker image version for v0.74

cjcjameson:


vito:
thanks!

robdimsdale:
@cjcjameson thanks for this - we forgot to update these docs when we moved away from bosh releases as the preferred method of distribution.

PR #323, opened 2016-03-25
Add Azure example manifest

making:
Added for Azure

### DIFF
--- a/manifests/azure.yml
+++ b/manifests/azure.yml
@@ -1,0 +1,147 @@
+---
+name: concourse
+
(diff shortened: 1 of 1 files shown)


vito:
Thanks for this! We'll be switching to BOSH 2.0 very very soon though and probably removing all these per-infrastructure examples, as 2.0 lets us just have a single manifest, which we'll inline into the docs instead. So I'm gonna close this since we'd be removing it almost immediately after anyway. Sorry!

PR #326, opened 2016-03-25
add instructions about how to upgrade Concourse in Vagrant

afeld:
Apparently

> Vagrant can not and does not automatically download the updated box and update the machine because...updating the machine requires destroying it and recreating it

https://www.vagrantup.com/docs/boxes/versioning.html

Not ideal that the machine needs to be re-created, but it doesn't look like there's an easy way to upgrade otherwise.


afeld:
Bump!

vito:
Thanks! Sorry for the delay.

PR #330, opened 2016-03-27
add pipeline examples to Tutorials page

afeld:


afeld:
Looks like this page has now turned into [Tutorials](http://concourse.ci/tutorials.html), which the rest of these aren't. I still think it would be useful to list open source pipelines for reference...ideas about where that should live?

vito:
Not sure yet. I've thought about having a 'pipeline gallery' of sorts, which could also be useful for sharing reusable pipeline templates.

robdimsdale:
Not to take this conversation too far down a different track, but a pipeline gallery starts to answer the question of reusability and composability. Resources are deliverable, but there's no equivalent for task/job/pipeline config.

robdimsdale:
Not to take this conversation too far down a different track, but a pipeline gallery starts to answer the question of reusability and composability. Resources are deliverable, but there's no equivalent for task/job/pipeline config.

afeld:
Sooo, a standalone page then? Or combined with Tutorials?

vito:
Let's start with a listing directly on the tutorial page for now. Something like:

```
(current content)

You may also want to just dive right in to some living, breathing Concourse pipelines. These aren't accompanied by documentation explaining them, but it may be useful just to see some real-world use cases:

* (concourse pipeline)
* etc
```

afeld:
Updated!

vito:
Looks good, thanks for putting this together!

PR #342, opened 2016-03-31
Update README, getting started

vincentserpoul:
getting started link redirects to a 404
Nothing fancy, replaced it with introduction :)


vito:
thanks!

PR #343, opened 2016-04-01
Fixed doc typos

kshinn:
Going through the documentation I found a couple of things that may need correction. For the code sample, I'm not sure if the golang reference was in reference to the sample yaml. If it was, this may make things less confusing.


vito:
thanks!

PR #359, opened 2016-04-08
Reference cloudfoundry-community slack resource in docs

robdimsdale:
- existing docs point to resource that does not work with newer versions of concourse and is not maintained


vito:
thx

PR #377, opened 2016-04-15
Provide example for `external_url` property.

oppegard:
When we upgraded and added this property, we did not include the protocol. This caused the basic auth login page to generate invalid links.


vito:
Thanks!

There's an `example:` field you can specify alongside `description` which would be nicer.

vito:
LGTM, thanks for tweaking it.

PR #387, opened 2016-04-20
Update build configuration files to support IBM Power platform for postgresql_9.3 package during packaging process

barthy1:
Hey,
I am part of the team, porting Cloud Foundry projects to Power microprocessor architecture.
The problem we are working on that some packages need updated config.guess file to be built on IBM Power platform. 

The detailed information and discussion about implementation and reasons of this PR can be found in https://github.com/cloudfoundry/bosh/pull/892

Similar solutions were implemented for Bosh (https://github.com/cloudfoundry/bosh/pull/898) and cf-release (https://github.com/cloudfoundry/cf-release/pull/751)


src/config/config.guess
```diff
@@ -0,0 +1,1454 @@
...
+
+# Historically, `CC_FOR_BUILD' used to be named `HOST_CC'. We still
+# use `HOST_CC' if defined, but it is deprecated.
+
+# Portable tmp directory creation inspired by the Autoconf team.
+
+set_cc_for_build='
+trap "exitcode=\$?; (rm -f \$tmpfiles 2>/dev/null; rmdir \$tmp 2>/dev/null) && exit \$exitcode" 0 ;
+trap "rm -f \$tmpfiles 2>/dev/null; rmdir \$tmp 2>/dev/null; exit 1" 1 2 13 15 ;
+: ${TMPDIR=/tmp} ;
```
jfrank-pivotal: I'm going to assume that these tmp variables are defined on at least some target OS.  They aren't on my machine or a few different Linux containers I checked.  I realize that this embedded script has been around for a long time (there's a fix dated in 1994) and was checked into bosh and cf-release as well, so it's probably fine.

jfrank-pivotal:
Merged into `develop`

PR #392, opened 2016-04-21
Add smuggler generic resource framework to docs

keymon:
Add the resource https://github.com/redfactorlabs/concourse-smuggler-resource/ to the list of third party resources.

Although it is not a resource per sé, but a framework to quickly implement resources and lower the entry barrier.


vito:
Thanks!

PR #408, opened 2016-05-01
Add Matrix Notification Resource type

freelock:
Add Matrix Notification Resource type to Resource Types page of website


vito:
Thanks!

PR #414, opened 2016-05-03
Add GitHub Commit Status third-party resource

dpb587:
Made a resource to work with branch/PR statuses if you think a reference would be useful to the community.


vito:
neat, thanks!

PR #425, opened 2016-05-04
Add travis-ci resource to community resources list

ArthurHlt:


zachgersh:
I would actually love to know what  you are doing with a pipeline then triggers travis.

ArthurHlt:
@zachgersh you have actually many usecase:
- Use concourse as a pipeline for travis build
- trigger build on concourse when a project which you don't own ran a build on travis
- Make small part of a project which are not in concourse and have a travis integration be available on concourse without change anything

For the trigger part, it was not "expensive" to do so I did it but well I saw just one usecase for this:
When you have a project which have a lib in dependency and you want to be sure that your project is still compatible when a new version of this dependency arrive

vito:
Merged manually due to conflicts, thanks!

PR #431, opened 2016-05-05
Add ami-resource and debian-sources-resource

jdub:
ami-resource checks for new Amazon Machine Images (AMIs).

debian-sources-resource checks for updated Debian and Ubuntu compatible package sources.


vito:
Thanks!

PR #437, opened 2016-05-11
Fix documentation

takuan-osho:
According to [the comment in ATC source codes](https://github.com/concourse/atc/blob/master/atccmd/command.go#L60), `postgres://user:pass@10.0.32.0/concourse` is wrong for the parameter of `--postgres-data-source` option.

The end of the parameter strings should be `atc`, not `concourse`.


vito:
Neither are wrong; it's just whatever you named the database. I'd rather have `concourse` be the name in docs since if you're using the binaries the ATC is kind of abstracted away.

PR #478, opened 2016-06-15
Fix formatting of example manifest

kkallday:
Move vm_extension description above the key to follow format of other manifest key descriptions.

### DIFF
--- a/docs/setting-up/installing.any
+++ b/docs/setting-up/installing.any
@@ -331,9 +331,9 @@ Concourse's ideals, buckle up and head over to \reference{clusters-with-bosh}.
         instances: 1
         # replace with a VM type from your BOSH Director's cloud config
         vm_type: REPLACE_ME
-        vm_extensions:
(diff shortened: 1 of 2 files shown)


vito:
Seems fine as-is to me. The comment is attached to a single extension to replace, not the whole set of them.

kkallday:
Hey @vito. In the integration tests for bbl (bosh-bootloader: https://github.com/pivotal-cf-experimental/bosh-bootloader/), we are deploying a bosh director and using it to deploy concourse. We are using a manifest packaged with our tests however we would like to pull the manifest from the concourse repo and replace values. This change would make it easier for us to do string replacement along with cleaning up the format of the comment.

vito:
That feels like a weird coupling to me. Structurally updating the YAML rather than textually would be much more reliable. You'd be dealing with breakage all the time as we tweak comments/whitespace/etc.

kkallday:
Yeah, thats a good idea. Thanks @vito!

PR #487, opened 2016-06-19
Update README.md

jadekler:


vito:
It's correct as-is; Garden is the interface, Guardian is a particular implementation. Thanks though!

PR #490, opened 2016-06-21
Clarify use of from flag in check-resource CLI command

cunnie:
- We were initially confused when the `ref:` prefix didn't work with the
  bosh-io-release resource
- Our goal was to point out that the prefix is resource-dependent

Signed-off-by: Brian Cunnie <bcunnie@pivotal.io>

### DIFF
--- a/docs/using-concourse/fly-cli.any
+++ b/docs/using-concourse/fly-cli.any
@@ -425,6 +425,10 @@ flag. Once you've learned what the commands do, you may want to consult
   This can be useful for collecting versions that are older than the current
   ones, given that a newly configured resource will only start from the latest
   version.
+
+  Note the \code{ref:} prefix is resource-dependent. For example, the
+  \italic{bosh-io-release} resource might use \code{version:11.2} in place of
+  \code{ref:abcdef}.
 }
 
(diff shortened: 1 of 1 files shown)


docs/using-concourse/fly-cli.any
```diff
@@ -415,16 +415,21 @@ flag. Once you've learned what the commands do, you may want to consult
...
 
   To check from a particular version, including the given version, append the
-  \code{--from} flag like so:
+  \code{--from} flag. For example, to check for version '11.2' and above on
+  the \italic{bosh-io-release} resource, use the following:
 
   \codeblock{bash}{
   $ fly -t example check-resource --resource my-pipeline/my-resource \\
-      --from ref:abcdef
+      --from version:11.2
```
vito: I prefer using `ref:abcdef` as the example; `version:` looks more like a keyword than `ref` to me. We should call out that it's for the `git` resource though.

vito:
thanks! left some feedback

cunnie:
@vito sorry for the delay. Per your suggestion we narrowed the scope of our change.

vito:
no worries! looks good, thanks.

PR #508, opened 2016-07-05
Add link to devpi-resource

mdomke:
The [devpi server](http://doc.devpi.net/latest/) is a PyPI mirror and development package server for Python packages. The [concourse-devpi-resource](https://github.com/mdomke/concourse-devpi-resource) allows it to track and upload packages from a devpi server.


vito:
thanks!

PR #509, opened 2016-07-05
Add rsync resource link

mrsixw:
Add rsync resource link


vito:
Merged manually, thanks!

PR #512, opened 2016-07-06
flight-school needs an `image_resource` early on

suhlig:
With the switch to Garden runC the `build.xml`, without adding an `image_resource` early on, fails with this error:

```
$ fly -t lite execute -c build.yml
targeting http://192.168.100.4:8080

executing build 219
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 40960    0 40960    0     0  13.4M      0 --:--:-- --:--:-- --:--:-- 19.5M
initializing
volume graph is disabled
errored
```

### DIFF
--- a/docs/tutorials/flight-school.any
+++ b/docs/tutorials/flight-school.any
@@ -110,10 +110,7 @@ the \reference{installing}{installation guide} before coming back here.
 
   image_resource:
     type: docker-image
-    source:
-      repository: ruby
-      tag: 2.3.0
-  }
+    source: { repository: busybox }
 
   inputs:
   - name: flight-school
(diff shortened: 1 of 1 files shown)


docs/tutorials/flight-school.any
```diff
@@ -108,6 +108,13 @@ the \reference{installing}{installation guide} before coming back here.
   \codeblock{yaml}{
   platform: linux
 
+  image_resource:
+    type: docker-image
+    source:
+      repository: ruby
+      tag: 2.3.0
+  }
```
vito: don't think you want this curly here

vito:
good catch, thanks for the PR. left a comment inline.

some of the later prose should also be updated, but we can pull this in and work on it.

suhlig:
I am happy to adjust the rest of the doc as part of this PR, but wasn't sure if it's ok to remove the whole 'This message is a little obscure...' section, because that error would not occur with the Ruby image.

If you still want to keep this section, what would be the base image that has no bash?

vito:
`busybox` should do. That's what it resulted in before. Adding this changes the semantics a little, as before it would add a config the user didn't realize they needed, but it's at least useful to show the changes that result from switching the image out.

suhlig:
I updated the following docs via another commit, wasn't that much of a change.

vito:
merged manually and updated the rest of the doc. thanks!

PR #553, opened 2016-07-22
add pagerduty resource type

srbry:
Support teams like to know when a concourse job fails, a common tool used is pagerduty hence having a pagerduty notifications resource is extremely useful.

Example:
```
jobs:
- name: health-check
  public: true
  serial: true
  plan:
  - get: pipeline
    trigger: true
  - get: every-2-minutes
    trigger: true
  - task: test
    file: pipeline/tasks/test.yml
    params:
      CONCOURSE_URL: {{concourse-url}}
    on_failure:
      put: pagerduty-alert
      params:
        description: {{description}}
        incident_key: {{incident_key}}
```


vito:
merged manually, thanks!

PR #554, opened 2016-07-22
Adding google drive resource

jpatel-pivotal:


vito:
Is the note about only supporting tar/txt files still true? I don't see any reference to that in the code. Seems like a funny limitation. (Very cool otherwise though!)

PR #561, opened 2016-07-29
Reduce size of baggageclaim btrfs image file.

alext:
This was previously created to be the same size as the `/var/vcap/data`
filesystem. This is a problem because this filesystem is also used by
other things (bosh packages and jobs, garden's graph and depot, logs
etc).

When the btrfs volume starts to get full, this can lead to errors when the
btrfs driver gets a `no space on device` error when writing to the
sparse image file. This in turn causes a kernel error, and the loopback
filesystem gets remounted read-only.

To reduce the likelihood of this, this changes the init script to allocate
10Gig less than the filesystem size. This PR also adds a property to allow
overriding the size for cases when 10Gig of overhead isn't enough.

### DIFF
--- a/jobs/baggageclaim/spec
+++ b/jobs/baggageclaim/spec
@@ -37,3 +37,10 @@ properties:
     description: |
       Environment name to specify for errors emitted to Yeller.
     default: ""
+
+  volumes_image_size_mb:
+    description: |
+      The size in megabytes of the volumes image to allocate. This is allocated
+      under /var/vcap/data, so should leave enough space for everything else.
+      Defaults to the filesystem size minus 10Gig.
+    default: -1
--- a/jobs/baggageclaim/templates/baggageclaim_ctl.erb
+++ b/jobs/baggageclaim/templates/baggageclaim_ctl.erb
@@ -30,7 +30,7 @@ case $1 in
     chown -R vcap:vcap $LOG_DIR
 
     IMAGE_PATH=/var/vcap/data/baggageclaim/volumes.img
-    IMAGE_SIZE=$(df -m /var/vcap/data | tail -n1 | awk '{print $2}')
+    IMAGE_SIZE=$(($(df -m /var/vcap/data | tail -n1 | awk '{print $2}') - 10 * 1024))
     mkdir -p $(dirname ${IMAGE_PATH})
 
     VOLUMES_DIR=/var/vcap/data/baggageclaim/volumes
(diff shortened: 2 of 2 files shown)


vito:
The theory behind the current sizing is that by creating a sparse file the same size as the disk, we'll run out of disk space the same way we would if we didn't need to create the file in the first place (e.g. if the full disk was already `btrfs`). It should only take up as much actual space as it's containing, not its full size. 

Even with this change, running out of space should happen in the same way, and the filesystem would be made readonly. (This should only be happening to the btrfs loopback.) You'd have 10GB free disk available for the rest of the components, but I don't really see that as a dramatic improvement given that any of them could just as easily use up the remaining disk space (e.g. Garden), and we're not bothering to give them a limit.

I could be convinced that sparse files don't quite behave the way we expect them to, though. Are you seeing it fill up significantly ahead of time? I've been curious around things like fragmenting of data leading to the disk filling up more quickly.

alext:
This is something that was exposed while we were dealing with https://github.com/concourse/concourse/issues/565, which lead to the `/var/vcap/data/baggageclaim/volumes` filesystem filling up.

> The theory behind the current sizing is that by creating a sparse file the same size as the disk, we'll run out of disk space the same way we would if we didn't need to create the file in the first place

This doesn't match our experience. The failure mode depends on which filesystem fills up first. The layers in play are:

```
baggageclaim
  |
  v
/var/vcap/data/baggageclaim/volumes filesystem
  |
  v
btrfs driver (loopback mount of /var/vcap/data/baggageclaim/volumes.img)
  |
  V
/var/vcap/data filesystem
```


When `/var/vcap/data` runs out of space first, it's the btrfs driver that gets the "no space left on device" error, which leads to a kernel error and the `/var/vcap/data/baggageclaim/volumes` filesystem being remounted readonly. At this point, baggageclaim can't do anything further (either releasing expired volumes, or creating new ones).

When the `/var/vcap/data/baggageclaim/volumes` filesystem runs out of space, it's baggageclaim that receives the "no space left on device" error. This does not cause the filesystem to be remounted readonly. In this situation, some operations in baggageclaim will fail, but not everything. It will still be possible for baggageclaim to release existing volumes, and the system will eventually recover by itself once some space is freed up.

Reducing the size of the sparse file means that we'll get the second failure mode (as long as there's enough headroom). We chose 10Gig based on the amount of space used by other components in a deployment using this bosh manifest - around 2.7Gig was used by the bosh jobs/packages, we've seen Garden use up to around 3Gig, and we want a bit of headroom for logs and other things.

vito:
Interesting, thanks for the additional info - didn't know about the discrepancy. This sounds reasonable, then. May want to keep an eye on that 10GB number but it may be enough given that pretty much all data used by Concourse should be in the volumes, with only some amount of constant overhead per container in `/var/vcap/data/garden`.

PR #571, opened 2016-08-03
Remove unecessary PATH modification in .envrc

tsaleh:
The `.envrc` file adds `$GOPATH/bin` to the `$PATH`, but that directory doesn't exist.  I expect this is cargo culted.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,2 +1,1 @@
 export GOPATH=$PWD
-export PATH=$PWD/bin:$PATH


vito:
Not cargo-culted, it gets created by e.g. `go install`. We could add a `.gitkeep` instead so your shell doesn't warn you.

tsaleh:
Ah, in that case, just kill this PR.

Cheers,
Tammer Saleh
+1 (702) 336-3381

PR #576, opened 2016-08-03
Add alternative email resource.

mdomke:
The original email resource ([pivotal-cf/email-resource](https://github.com/pivotal-cf/email-resource)) didn't include a
MTA and was therefore more complicated to configure and setup as I think
it should be.


vito:
neat - thanks!

PR #597, opened 2016-08-22
Add open build service resource to resource types

manno:
We create a resource to control Open Build Service from Concourse. Would it be possible to add it to the resource types page in the documentation?


vito:
neat! thanks.

PR #609, opened 2016-08-29
Fixing docker composer postgres link

shinji62:
Seems the postgres connection chain is not pass correctly.

Just putting on the same line with double quote seems to fix the "sslmode is invalid" error


docs/setting-up/installing.any
```diff
@@ -295,8 +295,7 @@ a tool that meshes well with Concourse's ideals, buckle up and head over to
       CONCOURSE_BASIC_AUTH_USERNAME: concourse
       CONCOURSE_BASIC_AUTH_PASSWORD: changeme
       CONCOURSE_EXTERNAL_URL: "$\{CONCOURSE_EXTERNAL_URL\}"
-      CONCOURSE_POSTGRES_DATA_SOURCE: |
```
vito: How about adding a `-` to the end of this instead? The value's a bit long otherwise; it'd run off the side of this page:

https://concourse.ci/docker-repository.html

vito:
actually I'll just do the fix I suggested myself since I already have it open. good catch though, thanks!

shinji62:
No pb .

PR #651, opened 2016-09-15
Added auth-duration as a configurable parameter to the atc job

geofffranks:
Took the new `--auth-duration` flag in `v2.1.0` and made it configurable in the BOSH release


vito:
someone beat ya to it: https://github.com/concourse/concourse/pull/638

thanks though 😄

geofffranks:
That's what I get for looking at v2.1.0 and not master. :)

PR #658, opened 2016-09-20
Update documentation and atc job for generic oauth scope verification

LinuxBozo:
This PR fixes an "oops, I forgot" that goes along with concourse/atc#112

### DIFF
--- a/docs/using-concourse/teams.any
+++ b/docs/using-concourse/teams.any
@@ -230,10 +230,12 @@ as).
 
   \section{Generic oAuth}{generic-oauth}{
     The \code{--generic-oauth-*} flags configure a generic oAuth provider which
-    performs no additional verification about the individual user signing in.
-    It should only be used with internal auth systems. If it were used to
-    configure Google or Twitter oAuth, for example, it would permit just about
-    every person on the internet to create pipelines. It'd be mighty generous.
+    performs no additional verification about the individual user signing in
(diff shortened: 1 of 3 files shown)


docs/using-concourse/teams.any
```diff
@@ -266,11 +268,24 @@ as).
           --generic-oauth-client-secret $CLIENT_SECRET \\
           --generic-oauth-auth-url https://oauth.example.com/o/oauth2/auth \\
           --generic-oauth-token-url https://oauth.example.com/o/oauth2/token \\
-          --generic-oauth-auth-url-param=scope:read
+          --generic-oauth-auth-url-param scope:read
```
vito: I wonder how this relates to the explicit scope flag? May be confusing having both of them there.

LinuxBozo: It's related in a way. During the OAuth process, the client can request a desired set of scopes. Some OAuth servers will simply fail the request if you don't request any scopes. When the user token is returned, that token includes the scopes that the user actually has. This is what the `-scope` flag is checking for verification.

LinuxBozo: Not redundant at all. Again, one is passed during the OAuth request (`generic-oauth-auth-url-param`) and the other is used for verification of the user in the response received (`generic-oath-scope`). Related, yet separate.

PR #677, opened 2016-10-03
Allow to specify Riemann tags

combor:
## What

This PR allows user to specify custom Rieman taggs so they can be used for metrics filtering. 
 
Tags should be a dictionary under Riemann key in ATC properties. 

```
  - name: atc
    properties:
      riemann:
        tags:
          foo: bar
```
## Why

We like to use disposable environments with one Concourse per env. Each concourse uses Riemann to send metrics to Datadog shared account. It's much easier to filter metrics if tags are set properly.

### DIFF
--- a/jobs/atc/spec
+++ b/jobs/atc/spec
@@ -284,3 +284,11 @@ properties:
     description: |
       An optional prefix for emitted Riemann services
     default: ""
+
+  riemann.tags:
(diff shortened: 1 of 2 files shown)


vito:
Could you re-submit this to `develop`?

combor:
Sure.

PR #678, opened 2016-10-03
Allow to specify Riemann tags

combor:
## What

This PR allows user to specify custom Rieman taggs so they can be used for metrics filtering. 
 
Tags should be a dictionary under Riemann key in ATC properties. 

```
  - name: atc
    properties:
      riemann:
        tags:
          foo: bar
```
## Why

We like to use disposable environments with one Concourse per env. Each concourse uses Riemann to send metrics to Datadog shared account. It's much easier to filter metrics if tags are set properly.

### DIFF
--- a/jobs/atc/spec
+++ b/jobs/atc/spec
@@ -284,3 +284,11 @@ properties:
     description: |
       An optional prefix for emitted Riemann services
     default: ""
+
+  riemann.tags:
+    description: |
+      An optional map of tags in key: value format
+    default: {}
+    example:
+      env: dev
+      foo: bar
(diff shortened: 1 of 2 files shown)


jobs/atc/spec
```diff
@@ -284,3 +284,11 @@ properties:
     description: |
       An optional prefix for emitted Riemann services
     default: ""
+
+  riemann.tags:
+    description: |
+      An optional list of tags in key: value format
```
vito: This should probably say 'map' or something rather than 'list'. List to me implies:

```
- env: dev
- foo: bar
```

vito:
Looks good, only one minor rewording requested. Sorry I didn't catch that earlier before having you reopen. :)

combor:
No worries. I'm a little bit confused how new github review functionality works so I replied twice.

vito:
looks good! thanks.

PR #681, opened 2016-10-03
update typo

7hunderbird:
Is there a way to automate in a pipeline a command line spell checker like `aspell`, etc. for the docs?

Here's one I noticed that can be corrected.

Cheers.


vito:
Sorry, mistook this for a different issue.

Yeah, spell-checking as part of the pipeline would be a good idea. We'd need to have a pretty large whitelist though.

PR #716, opened 2016-10-23
Document job hooks.

jmcarp:
As implemented in https://github.com/concourse/atc/pull/125.

### DIFF
--- a/docs/configuring/jobs.any
+++ b/docs/configuring/jobs.any
@@ -98,6 +98,19 @@ Each configured job consists of the following attributes:
   disabled.
 }
 
+\define-attribute{on_success: step}{
+  \italic{Optional.} Step to execute when the job succeeds.
+}
+
+\define-attribute{on_failure: step}{
+  \italic{Optional.} Step to execute when the job fails.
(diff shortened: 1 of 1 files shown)


docs/configuring/jobs.any
```diff
@@ -98,6 +98,19 @@ Each configured job consists of the following attributes:
...
 
+\define-attribute{on_success: step}{
+  \italic{Optional.} Step to execute when the job succeeds.
+}
+
+\define-attribute{on_failure: step}{
+  \italic{Optional.} Step to execute when the job fails.
+}
+
+\define-attribute{ensure: boolean}{
```
vito: Not a boolean.

Also these three things should probably link to their step equivalent. This can be done with `\reference{ensure}`. These should probably also have a different tag so as to not conflict with the step-wise docs. This can be done with `\define-attribute{ensure: step}{job-ensure}{...}`.

jmcarp:
Updated, let me know if anything else needs fixing.

vito:
Looks good, thanks!

PR #768, opened 2016-11-12
Add link to RSS resource

suhlig:
I created a resource for tracking RSS feeds. This PR is to create a link in the docs, pointing to the resource's github page.


vito:
neat! thanks.

PR #775, opened 2016-11-14
Add calendar resource

henrytk:
This pull request adds a [calendar resource](https://github.com/henrytk/calendar-resource) to the list of third party resources. The resource currently only supports Google Calendars, but could be extended to other providers.


vito:
neat!

PR #779, opened 2016-11-16
Alphabetize resource types for readability

gilmatic:
Simple PR to alphabetize user submitted Resource Types for improved readability. I had a teammate miss the Terraform resource type, thus the PR.


vito:
Thanks for the PR but I think this'll end up being a burden over time (people adding to the list will have to preserve the order, so it'd be just another reason to wait). Ctrl+F/Cmd+F is probably enough until we've got a proper marketplace.

PR #861, opened 2017-01-09
Fix docker-compose install key generation instructions

joonas:
It looks like what's on the [Docker Repository set up](https://concourse.ci/docker-repository.html) page is different from [what's set up in the repository](https://github.com/concourse/concourse-docker/blob/master/generate-keys.sh), and since the output of the script in the repository is what the [Dockerfile uses in the environment variables](https://github.com/concourse/concourse-docker/blob/master/Dockerfile#L11-L18), this reconciles the differences.

### DIFF
--- a/docs/setting-up/installing.any
+++ b/docs/setting-up/installing.any
@@ -313,13 +313,13 @@ a tool that meshes well with Concourse's ideals, buckle up and head over to
   \codeblock{bash}{
     mkdir -p keys/web keys/worker
 
-    ssh-keygen -t rsa -f ./keys/web/host_key -N ''
+    ssh-keygen -t rsa -f ./keys/web/tsa_host_key -N ''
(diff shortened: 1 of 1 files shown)


vito:
Related to https://github.com/concourse/concourse/commit/b4b1e68b66cab2f718cc06b08016702056c0f3df - looks like we moved in the opposite direction first? Not sure how to reconcile this.

joonas:
@vito - as an alternative approach, I updated the concourse docker image to use the correct paths in concourse/concourse-docker#5, I'd be fine with closing this in favor of merging that instead and shipping those new images that would work with the current documentation. WDYT?

vito:
fixed via https://github.com/concourse/concourse/commit/5fce134ff8cb7a6d3ba5a6bd8dc8bb6a8178bc4c and https://github.com/concourse/concourse/commit/f537433f53ecea6557c19e0e675d7b295c0259d9 - thanks!

PR #866, opened 2017-01-10
Use max/min-widths to enforce sidebar width.

### LINKED ISSUE #865: Broken CSS on single-page docs
# Bug Report

Single-page Docs appear broken when viewed on a browser window narrower than 1200px. Observed on both chrome and firefox.

screen shot 2017-01-10 at 22 56 52

asehra:
Fixes #865 

Before: 
before

After:
after


vito:
thanks!

PR #880, opened 2017-01-16
bump x/crypto to fix ssh deadlock

databus23:
This PR bumps the crypto dependency to pull in https://github.com/golang/crypto/commit/2e74c773682f59dc50a56475f7918dd8fa6dcaf8 (See https://github.com/golang/go/issues/18439 for some more details)

Hopefully this addresses #750.

/cc @vito

Issue #750, opened 2016-11-04
Workers stop heartbeating


databus23:
We are seeing workers disappearing for a couple of weeks now.
When running 2.2.1 we had multiple occurrences of all workers (darwin & linux) not reconnecting after an atc/tsa outage. They stopped heartbeating with a failure and never resumed heartbeating.

Now that we have upgraded to 2.4.0 workers start disappearing at random without tsa going down.
There are not even errors logged by the workers. They just stop heartbeating. Restarting the workers makes them reregister with tsa again.
The least log messages related to heartbeating are:
```
Nov 03 15:52:12 mo-a5b515ae6 concourse[30611]: {"timestamp":"1478188332.755291224","source":"tsa","message":"tsa.connection.forward-worker.heartbeat.start","log_level":1,"data":
Nov 03 15:52:12 mo-a5b515ae6 concourse[30611]: {"timestamp":"1478188332.758820534","source":"tsa","message":"tsa.connection.forward-worker.heartbeat.reached-worker","log_level":0,"data":{"remote":"10.46.14.60:50180","session":"8.1.202","took":"3.308641ms"}}
Nov 03 15:52:12 mo-a5b515ae6 concourse[30611]: {"timestamp":"1478188332.763972282","source":"tsa","message":"tsa.connection.forward-worker.heartbeat.done","log_level":1,"data":{"remote":"10.46.14.60:50180","session":"8.1.202","worker-address":"100.77.0.15:34107","worker-platform":"linux","worker-tags":""}}
```

netstat shows me that the workers still have a connection open to tsa:2222 but it seems like nothing is going over the wire anymore.
I'm not ruling out intermittent network problems in our infrastructure causing this but in my mind the workers should be able to recover from this in all cases as soon as the network is fine again.
Could it be that the heartbeating gets stuck and doesn't kill the tcp connection in all cases? I'm not familiar with how the heartbeating control loop works.

We are running concourse binaries. The web component runs in k8s and we have workers running on VMs outside of the k8s cluster.

ahume:
@databus23 Can I ask why you choose not to run workers in the k8s cluster? I'm currently looking in to doing this and am wondering if this is the wrong path?

databus23:
Just happened again for a darwin worker (2.4.0).
These are the last log messages, after that just silence:
```
[...]

PR #900, opened 2017-01-30
Bump golang.org/x/crypto to fix ssh deadlock

databus23:
This is a follow up for #880 which was rolled back due to https://github.com/golang/go/issues/18711 which was addressed by https://github.com/golang/crypto/commit/6fb066875c8002ccd49dbd9a16b522711cb8f59d.

This hopefully puts https://github.com/concourse/concourse/issues/750 to bed. Lets see.

/cc @vito


vito:
sweet, thanks!

PR #910, opened 2017-02-06
bump crypto, third try

databus23:
This is the third try to bump the `golang.org/x/crypto` dependency to fix an ssh deadlock uncovered in #750.

The two previous attempts #880 and  #900  had to be reverted because of regressions.

I believe those are now fixed in upstream by https://github.com/golang/crypto/commit/77014cf7f9bde4925afeed52b7bf676d5f5b4285 and https://github.com/golang/crypto/commit/641ab6b32049cabca26c30bf27baaae445bf4175.

Lets see what breaks next.


vito:
thanks for keeping track of this!

PR #949, opened 2017-03-03
Download pivotal products and stemcells

rsamban:
1. Implemented with bash scripts
2. Can download multiple products in one operation.
3. You must specify exact product version that needs to be downloaded.

### DIFF
--- a/docs/using-concourse/resource-types.any
+++ b/docs/using-concourse/resource-types.any
@@ -168,6 +168,8 @@ before using it!
     \hyperlink{https://github.com/suhlig/concourse-rss-resource}{RSS}
   }{
     \hyperlink{https://github.com/patrickcrocker/maven-resource}{Maven Resource}
+  }{
+    \hyperlink{https://github.com/rsamban/pivnet-resource}{Pivnet Resource}
   }
 
   \section{Adding to this list}{


zachgersh:
@rsamban - how does this differ from the official pivnet-resource that is already here:

https://github.com/pivotal-cf/pivnet-resource

I wouldn't actually want to see this merged as it would cause confusion between the two resources.

rsamban:
I agree, this is almost duplicate of https://github.com/pivotal-cf/
pivnet-resource

Only thing it adds is,
1. You can download stem cells (specific to your IaaS)
2. you can download multiple products with one get operation (handy when
you are installing/upgrading large installations and you don't want to
clutter your pipeline with several get operations)
3. You must specify the full version - so no accidental download/upgrade of
the products. (You can do same thing with official release, but you can
make mistakes)

I thought it could be alternate resource.

I am good either way,

Let me know

thanks
-Ramesh

chendrix:
@zachgersh @rsamban what about prefixing one or both of them with their maintainers?

The resource types list in the documentation isn't some sort of canonical location of "blessed resources". It's a stopgap solution because we don't have a kind of "resource marketplace" or any discoverability. I don't want to gatekeep some people's resource types.

rsamban:
I am fine with prefixing it with my github handle.

Let me know

thanks
-Ramesh

zachgersh:
@chendrix I disagree with this. Even if concourse doesn't want to maintain the "blessed" resources it in a way is because being on the official documentation gives that impression. I'd prefer that people use the official pivnet / bosh-stemcell resources if they are in need of what this resource provides.

rsamban:
@zachgersh <https://github.com/zachgersh> @chendrix
<https://github.com/chendrix>  I don't want this to be too complicated. I
am ok if my pull request is rejected.

I understand @chendrix <https://github.com/chendrix>'s point.

Cheers!

thanks
-Ramesh

vito:
@zachgersh I don't think the official docs give that impression. They have a call-out right above the list that basically says "use at your own risk". We merge additions to that list with very light overview.

Issue #879, opened 2017-01-14
ATC memory issue after upgrade 2.6.0


ahelal:
Hi there!

# Bug Report

After upgrading to 2.6.0 the ATC node started to fail after a while 6 hours or so. It keeps consuming memory till it fails. 

Prior to upgrade everything was fine version 2.5.1

I attach a memory graph 
![atc-memory](https://cloud.githubusercontent.com/assets/4069495/21957605/20d15862-da9a-11e6-86a4-557c4151f796.png)

* Concourse version: 2.6.0 and 2.7.0
* Deployment type: binary
* Infrastructure/IaaS: AWS t2.medium
* Did this used to work? yes
* Linux : Ubuntu 16.04 Kernel 4.4.0-62 

Only work around so far is to restart concourse web twice a day

wieslaw-gat:
I've also encountered this problem, `concourse web` keeps hogging more and more memory - the usage goes up to ~2GB after which I have to restart it before oomkiller reaps it.

This started happening after the upgrade to 2.6.0, earlier no such problems appeared.

ahelal:
I also attached the heap from pprof might be helpful https://gist.github.com/ahelal/5af9bd99f430ea88e89e7c76d2ceb95e

arwineap:
Not a whole lot to add, but I did verify that my install also started gobbling memory on the web node.

http://i.imgur.com/qu5cPhn.png

schubert:
FTW since we saw this and reported it in #892 we believe this event correlated with a job failing on a worker at the same time with the following failure:

```
! DETAIL:  could not fork new process for connection: Resource temporarily unavailable
```

Which I completely speculate could be caused by exhausting file descriptors on the worker which could be caused by the many stdout/stderr handlers open and perhaps not being cleaned up and therefore exhausting FD's on the worker while simultaneously holding onto memory over on the web VM leading to the memory exhaustion.

Or it was just a coincidence but I thought it would be worth mentioning the correlation.

cc @tom-meyer

dsharp-pivotal:
Since it's not mentioned so far here: the leak seems to be in the TSA process that runs alongside the ATC on the same node.

I noticed an uptick in memory usage on one of our web nodes around 16:00 PST today, so I looked at the TSA logs around then.

I've attached around 5k lines of logs. There do seem to be some differences in the logs around 16:00, though nothing I can make sense of.

Here's an interesting snippet from around the time:
```
[...]

PR #968, opened 2017-03-14
Restart TSA if it's leaking memory

tushar-dadlani:
This mitigates https://github.com/concourse/concourse/issues/879 , while a fix is developed.

### DIFF
--- a/jobs/tsa/monit
+++ b/jobs/tsa/monit
@@ -3,3 +3,4 @@ check process tsa
   start program "/var/vcap/jobs/tsa/bin/tsa_ctl start"
   stop program "/var/vcap/jobs/tsa/bin/tsa_ctl stop"
   group vcap
+  if memory > 80% for 6 cycles then restart


voelzmo:
ouch, brutal.

dsharp-pivotal:
Brutal? If the process is mis-behaving, this seems like a good safeguard. Even in the absence of #879, I think this is a good idea.

schubert:
@vito given this is a cheap safeguard to at least mitigate the problem any reason this couldn't make it into 2.7.1 while the root cause of #879 is figured out?

chendrix:
Hi all, we believe we've figured out and fixed #879 which will be included in 2.7.1

Because of that, I'd prefer to kick the can down the road a little bit and not add this into our monit file until we've had a larger discussion about our theme "Efficiency". As such, I'm closing this but I'm tagging the tracker story with `efficiency`.

PR #1015, opened 2017-04-17
docker compose compatibility for v.3

yaman:
Adding docker compose compatibility for v.3. Also managing service dependencies with depends_on. concourse-worker depends on concourse-web and concourse-db. concourse-worker startup fails If it tries to start before concourse-web.

### DIFF
--- a/docs/setting-up/installing.any
+++ b/docs/setting-up/installing.any
@@ -277,6 +277,9 @@ a tool that meshes well with Concourse's ideals, buckle up and head over to
   Compose}, create \code{docker-compose.yml}:
 
   \titled-codeblock{docker-compose.yml}{yaml}{
+  version: '3'
+
+services:
   concourse-db:
     image: postgres:9.5
     environment:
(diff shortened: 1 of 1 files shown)


vito:
Worker shouldn't depend on DB as it doesn't talk to it directly - is it enough for it to depend on Web, which itself depends on DB?

yaman:
@vito yes indeed. I thought that it would be safe for worker to depend on both web and db since worker will not be usable if db is not ready anyway. And **depends_on** field explicitly shows the worker service dependencies. I preferred being explicit rather than being implicit...

vito:
@yaman I don't feel that having `worker` depend on `db` is the right kind of explicit. It suggests that the worker component has a direct dependency on `db`, which it doesn't. For example, you may configure your `web` node to run against an external database - this has no bearing on how the `worker` nodes are configured, and would just mean you remove the `db` and its dependency from `web`. Transitive dependencies are a common thing that I'd expect Docker to resolve correctly, and it builds a better mental model to express the relationships as-the-are.

yaman:
@vito I will make the changes soon. Thanks for the insight.

yaman:
@vito done.

chendrix:
@yaman any updates to this PR?

andrewedstrom:
There were some whitespace issues, but we've gone ahead and brought in the changes. Thanks again for your PR!

Issue #939, opened 2017-02-27
Documentation states that `inputs` is required, but is it really?


kdvolder:
In the docs here: https://concourse.ci/running-tasks.html#inputs

It states that 'inputs' is a required attribute. However, looking at some tutorial examples, like this one:

https://concourse.ci/hello-world.html

We see that its example 'hello.yml' task does not specify 'inputs' at all yet it runs just fine. 

It seems the tutorial and the docs send mixed signals on the 'inputs' being required (and also fly cli doesn't seem to care if its missing)

PS: Some background. We are working on a prototype for a Concourse Pipeline/Task editor (which does some linting). Some initial feedback we received was around the errors being reported by our linter around 'required' properties. The user was working through tutorial materials like the 'hello-world' example and noticed the editor reporting errors on a 'working' pipeline file. 

We have simply based our checking of 'required' properties on what the concourse docs say, but are now unsure if the docs are trustworthy in this respect.

In general, are 'required' properties really required? Or are they (sometimes) just 'recommended'. Is it appropriate for a linter to show an error when they are missing? Or is a warning more appropriate? Or should we even bother checking them at all? 

Or perhaps, in this specific example, the documentation is just incorrect and 'inputs' is really just a n optional property.

We're excited about creating a nice editor for Concourse, but we could really use a little help with questions like these. So I hope raising a github issue is okay. If not, please advise on how/where we should try and get these questions answered. (Note: We have already tried raising our questions on the slack channel but haven't gotten much of a response there. I gather, perhaps the channel is just too busy for the questions to get noticed by the right person)

ClaraFuPivotal:
Inputs is just an optional property so you shouldn't need to put a error or warning for when it's not there. Thanks for catching that! You can either submit questions here or through slack, sorry about not getting your questions answered on there. Slack is kind of hard to track questions IMO.

kdvolder:
Thanks for answering. But, perhaps you shouldn't close the ticket without fixing the docs. I made inputs optional in our editor a long time ago. I just thought maybe someone might want to fix the docs too.

I just checked and it looks like it still says that inputs is required:

> inputs: [object]
>
> Required. The expected set of inputs for the task.

chendrix:
We closed the ticket after fixing the docs in https://github.com/concourse/concourse/commit/38e69ae4af800f07fbaa76196ea23c4b7d3bc976

chendrix:
Looks like this documentation won't get pushed until after we release.

[...]

PR #1046, opened 2017-05-08
Fix docs: `inputs` in task is optional.

kdvolder:
Follow up on issue #939.


chendrix:
We changed this already in https://github.com/concourse/concourse/commit/38e69ae4af800f07fbaa76196ea23c4b7d3bc976

I will republish the docs now.

PR #1185, opened 2017-05-20
Documentation: Clarify `get:` step

cjcjameson:
The documentation previously made it seem like `get:` always was the name matching what was declared up in the `resources:` section. But it confused people because it's not -- it's only that it typically does unless you use the `resource:` attribute.


vito:
thanks!

PR #1382, opened 2017-07-15
documentation: Clarify interaction between params pipeline vs task.yml

cjcjameson:
- Specifically not referencing pipeline implications in
  `running-tasks.any`
- Note that `\reference{foo}` refers to the `\title{foo}` field of the
  target page, and it's also the same as what you see with `.html` on
  the live site. It doesn't refer to the `*.any` filename

Resubmission of #1186

### DIFF
--- a/docs/introduction/concepts.any
+++ b/docs/introduction/concepts.any
@@ -38,6 +38,9 @@ becomes more involved.
 
   To learn more about configuring and running tasks, see
   \reference{running-tasks}.
+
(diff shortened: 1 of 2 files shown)


jtarchie:
Thanks for the PR @cjcjameson. We recently had some changes to how our documentation is done. Mind taking a look at the conflict. Sorry for the delay.

vito:
this has been manually merged in - sorry for the delay!

PR #1456, opened 2017-08-09
Fix documentation on how to upgrade the VM with bosh

SHyx0rmZ:
Previous commit referred to `delete-env` instead of `create-env`.


SHyx0rmZ:
Fixed by c427cb929fbdf544a7d2220228e21f0ea4df0637.

PR #1529, opened 2017-08-28
A resource that can trigger jobs using slack

ahelal:
Hi I would like to add a new resource to the list. A resource that can trigger jobs using slack


vito:
Love it. Folks have said in the past that the Slack notification resource is an example of somewhere that the resource model breaks down, since it only implements `out`, but this nicely completes the interface.

Issue #1201, opened 2017-05-23
Aggregate issue: 'too many open files'


vito:
Please provide:

* Which component is emitting this?
* What files are open? (`lsof -p <process id>`)
* Concourse version
* Deployment type (e.g. binary, Bosh, K8s, Docker)

henderjm:
We saw this a couple of weeks ago, so going by memory
* Which component is emitting this?
  - ATC
* What files are open? (lsof -p <process id>)
  - Socket connections
* Concourse version
  - 2.6.0

Resolution:
1. We initially tried to fix this by bumping our ATC instances, this worked for about a day but ran into the error the next day.
1. So instead, we deployed a [ulimit-release](https://github.com/pivotal-cf/ulimit-release) which increased our number of open files. Have not seen the problem since.

defsprite:
We're seeing a lot of open pipes using the standalone binaries with docker.
It *feels* like they are growing steadily, but I have no definite proof of that yet.

Docker version: `Docker version 17.05.0-ce, build 89658be`
Concourse version: `3.0.1`

```
root@2ba1459bb611:/# date; lsof -Xn 2>/dev/null | wc -l
Wed Jun 21 16:47:52 UTC 2017
1451891
```

When grepping for random pipe, we see something like this:

```
root@2ba1459bb611:/# cat /tmp/lsof-output.txt | grep 713087065
COMMAND     PID   TID     USER   FD      TYPE DEVICE  SIZE/OFF       NODE NAME
concourse     7           root  533r     FIFO   0,10       0t0  713087065 pipe
concourse     7     8     root  533r     FIFO   0,10       0t0  713087065 pipe
concourse     7     9     root  533r     FIFO   0,10       0t0  713087065 pipe
concourse     7    10     root  533r     FIFO   0,10       0t0  713087065 pipe
concourse     7    11     root  533r     FIFO   0,10       0t0  713087065 pipe

<lots more here, TID keeps growing>

concourse     7 65519     root  533r     FIFO   0,10       0t0  713087065 pipe
concourse     7 65522     root  533r     FIFO   0,10       0t0  713087065 pipe
concourse     7 65526     root  533r     FIFO   0,10       0t0  713087065 pipe
runc      45372           root    1w     FIFO   0,10       0t0  713087065 pipe
runc      45372  9047     root    1w     FIFO   0,10       0t0  713087065 pipe
runc      45372 11925     root    1w     FIFO   0,10       0t0  713087065 pipe
[...]

PR #1530, opened 2017-08-28
Add option to increase ATC open files limit

gerhard:
The default 1024 open files limit is known to cause 'too many open files' errors

We don't want to change the default ATC open files limit for everyone,
we're simply adding the option to pick a better value when required.
Having measured and understood our very specific case, 10240 was
determined to be adequate for us. We hope that you will also measure and
understand the optimal value for your use-case.

re concourse/concourse#1201

[#150434196]

### DIFF
--- a/jobs/atc/spec
+++ b/jobs/atc/spec
@@ -452,3 +452,9 @@ properties:
       information about the user. The only metric collected is the version of
       concourse.
     default: false
+
+  atc.open_files_limit:
+    description: |
+      Set ATC (build scheduler & web interface) open files limit.
+      In high contention environments, the default limit is known to cause 'too many open files' errors.
+    default: 1024
--- a/jobs/atc/templates/atc_ctl.erb
+++ b/jobs/atc/templates/atc_ctl.erb
@@ -90,6 +90,8 @@ case $1 in
 
     setcap cap_net_bind_service=+ep /var/vcap/packages/atc/bin/atc
 
+    ulimit -n <%= p("atc.open_files_limit") %>
+
     exec chpst -u vcap:vcap /var/vcap/packages/atc/bin/atc \
       --bind-ip <%= p("bind_ip") %> \
       --bind-port <%= p("bind_port") %> \


pivotal-issuemaster:
@gerhard Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/concourse&pullRequestId=1530)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/concourse&pullRequestId=1530) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@gerhard Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/concourse&pullRequestId=1530)!

jtarchie:
@gerhard, thanks for the PR!

There is a [bosh-release](https://github.com/pivotal-cf/ulimit-release) that gives . We are trying to stray away from [too many knobs and dials](https://github.com/concourse/concourse/wiki/Anti-Patterns#knobs) from the OS level.

Please give that a try to solve your issue.

gerhard:
I agree that **how** open file limits are configured is not a Concourse release concern. This is what it means to use ulimit release to set open file limits for ATC:

1. `nofile limits in /etc/security/limits.d` will not apply to the ATC job, since the atc process runs through `chpst`
1. all monit jobs running on the instance that runs ATC will have the same open file limit
1. if monit process restarts, open file limits will not be set correctly until pre-start from the ulimit job runs. There is no guarantees when `pre-start` will run, only that it will run at least once.

I believe that ulimit release does what it can, open file limits is clearly a BOSH concern since all dependencies that have the authority to set them are BOSH dependencies. Moving this discussion to a better place, thank you @jtarchie

gerhard:
This has just bit us again today. [Created a ulimit release that can be consumed remotely]() and set both soft & hard file limit to 10240.

@jtarchie @vito I believe [bpm-release](https://github.com/cloudfoundry-incubator/bpm-release/blob/master/docs/config.md#job-configuration) is a much better alternative to addressing this long-term. Thoughts?

PR #1725, opened 2017-10-17
Stop always skipping TLS verification for credential managers

aeijdenberg:
Previous code incorrectly passed a boolean value to a boolean flag.

This is wrong because the flag parser silently ignores values for boolean flags, and simply sets the value to true.

(I have a separate ATC PR to return an error when an unrecognized positional argument like this is received).

As such all bosh deployments of atc are skipping TLS verification whether you want it or not for CredHub and Vault.

There's a good chance this change will break existing installations that thought they were verifying TLS, but really weren't, and may not be configured correctly to do so.

### DIFF
--- a/jobs/atc/templates/atc_ctl.erb
+++ b/jobs/atc/templates/atc_ctl.erb
@@ -223,7 +223,9 @@ case $1 in
       <% if_p("credhub.tls.ca_cert") do |ca_cert| %> \
       --credhub-ca-cert /var/vcap/jobs/atc/config/credhub_ca_cert \
       <% end %> \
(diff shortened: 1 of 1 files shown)


aeijdenberg:
@vito - Hi Alex, please see the above (and linked ATC PR) - this will get harder to fix as more deployments use credential managers, as this config causes ATC to *always* bypass TLS validation, even when the user explicitly tells it not to.

vito:
Ack, thanks and sorry for the delay!

aeijdenberg:
Thanks @vito !

PR #1849, opened 2017-11-24
Add prometheus properties.

jmcarp:
BOSH properties or it didn't happen.


TimSimmons:
yay! :+1: I had no idea this needed to happen.

vito:
thanks!

Issue #1862, opened 2017-12-01
3.7.0 bosh deployed worker has an empty known_hosts


camattin:
# Bug Report

The following can also be handy:

* Concourse version: 3.7.0
* Deployment type: BOSH
* Infrastructure/IaaS: vSphere
* Browser (if applicable): n/a
* Did this used to work? Yes

The changes from [#1834](http://github.com/concourse/concourse/issues/1834) appear to have broken the 3.7.0 bosh deployed worker. Relevant snippet from the worker bosh manifest:

```
  - name: groundcrew
    release: concourse
    consumes: {baggageclaim: {from: <workername>}}
    properties:
      tsa:
        worker_key:
          private_key: |
            <WORKER PEM PRIVATE KEY>
          public_key: "ssh-rsa <worker public key>"
        host: <tsa IP>
        host_public_key: "ssh-rsa <tsa public key>"
```

This results in an empty known_hosts file on the worker, so then he cannot talk back to the tsa, of course.

Our only fix as of now is to put the workers back to 3.6.0. As the generated* pieces were previously responsible for creating the known_hosts file in 3.6.0, it seems that this is the relevant change which created this new issue?

andrewedstrom:
Same issue. The known_hosts problem made our external worker show up as `stalled`. In order to fix it we had to go into credhub to find the new tsa public key that was generated, then manually insert it into `/var/vcap/data/jobs/groundcrew/f8aef18bdc2124983afb14ffa9ea5874f9e2ff8e/config/known_hosts` on our external worker vm

vito:
Already fixed by https://github.com/concourse/concourse/pull/1867 - thanks for the report!

For now you I would stick the workers to 3.6.0; there's no required upgrade for 3.7.0, compatibility-wise. The fix will be shipped soon.

PR #1867, opened 2017-12-05
Render known_hosts entry

christianang:
This fixes the `known_hosts` rendering problem in https://github.com/concourse/concourse/issues/1862.

On [jobs/groundcrew/templates/known_hosts.erb#L7](https://github.com/concourse/concourse/blob/v3.7.0/jobs/groundcrew/templates/known_hosts.erb#L7), the string that is supposed to be rendered by the erb is not within a scope where the erb can see it. We had to add a variable outside the block and set that for it to be brought within scope and rendered correctly.

@oozie and @christianang

### DIFF
--- a/jobs/groundcrew/templates/known_hosts.erb
+++ b/jobs/groundcrew/templates/known_hosts.erb
@@ -3,14 +3,16 @@
   tsa_port = nil
   tsa_host_public_key = nil
 
+  known_host = ""
   if_p("tsa.host", "tsa.port", "tsa.host_public_key") do |host, port, public_key|
-    "[#{host}]:#{port} #{public_key}"
+    known_host = "[#{host}]:#{port} #{public_key}"
   end.else do
     tsa = link("tsa")
     port = tsa.p("bind_port")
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@christianang Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/concourse&pullRequestId=1867)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/concourse&pullRequestId=1867) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@christianang Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/concourse&pullRequestId=1867)!

vito:
Ah, maybe the if_p and else methods don't return their branch's result. Weird. Thanks for looking at this!

jtarchie:
We should take this to the *danger zone* and write a top gun test for it!

PR #1868, opened 2017-12-05
Wrap this to interpolate variable

sharms:


jtarchie:
Doh! Good catch. Thanks for fixing it quickly.

PR #1895, opened 2017-12-13
Update spec with SSM parameter store credential management

surajbarkale:
This PR depends on https://github.com/concourse/atc/pull/229

### DIFF
--- a/jobs/atc/spec
+++ b/jobs/atc/spec
@@ -507,3 +507,42 @@ properties:
     description: |
       Client secret for CredHub authorization.
     default: ""
+
+  ssm.aws_region:
+    description: |
+      AWS region to use for fetching SSM parameters. Enviroment variable AWS_REGION is used
+      if this flag is not provided. This flag or the environment variable *must* be specified
+      for SSM secret backend to be enabled. When AWS_SDK_LOAD_CONFIG environment variable is
+      set, shared configuration files (~/.aws/config & ~/.aws/credentials) are used to load
+      this flag and credential flags. A profile within shared config file may be selected by
+      specifying it in AWS_PROFILE or AWS_DEFAULT_PROFILE environment variables.
+    default: ""
+  ssm.aws_access_key:
+    description: |
+      AWS Access key ID used as credentials for accessing SSM parameters. If this flag is not specified,
+      AWS_ACCESS_KEY_ID or AWS_ACCESS_KEY environment variables are used. On EC2 instances, the metadata
+      service is used to get credentials when none are specified.
+    default: ""
+  ssm.aws_secret_key:
+    description: |
+      AWS Secret Access Key used as credentials for accessing SSM parameters. If this flag is not specified,
+      AWS_SECRET_ACCESS_KEY or AWS_SECRET_KEY environment variables are used. On EC2 instances, the metadata
+      service is used to get credentials when none are specified.
+    default: ""
+  ssm.aws_session_token:
+    description: |
+      AWS Session Token used as credentials for accessing SSM parameters. If this flag is not specified,
+      AWS_SESSION_TOKEN environment variable is used. On EC2 instances, the metadata service is used to
+      get credentials when none are specified.
+    default: ""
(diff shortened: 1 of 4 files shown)


jobs/atc/spec
```diff
@@ -507,3 +507,42 @@ properties:
...
+      AWS_SECRET_ACCESS_KEY or AWS_SECRET_KEY environment variables are used. On EC2 instances, the metadata
+      service is used to get credentials when none are specified.
+    default: ""
+  ssm.aws_session_token:
+    description: |
+      AWS Session Token used as credentials for accessing SSM parameters. If this flag is not specified,
+      AWS_SESSION_TOKEN environment variable is used. On EC2 instances, the metadata service is used to
+      get credentials when none are specified.
+    default: ""
+  ssm.pipe_secret_template:
```
vito: Left a similar comment in the ATC PR, but I think we shouldn't abbreviate this, and it should be "pipeline".

jobs/atc/spec
```diff
@@ -507,3 +507,42 @@ properties:
...
+
+  ssm.aws_region:
+    description: |
+      AWS region to use for fetching SSM parameters. Enviroment variable AWS_REGION is used
+      if this flag is not provided. This flag or the environment variable *must* be specified
+      for SSM secret backend to be enabled. When AWS_SDK_LOAD_CONFIG environment variable is
+      set, shared configuration files (~/.aws/config & ~/.aws/credentials) are used to load
+      this flag and credential flags. A profile within shared config file may be selected by
+      specifying it in AWS_PROFILE or AWS_DEFAULT_PROFILE environment variables.
+    default: ""
```
vito: These defaults can be removed since you're using `if_p` anyway, and they're not really defaults. :)

jobs/atc/spec
```diff
@@ -507,3 +507,42 @@ properties:
...
+  ssm.aws_access_key:
+    description: |
+      AWS Access key ID used as credentials for accessing SSM parameters. If this flag is not specified,
```
vito: Might as well not mention flags or env vars since this is the only way to set them.

surajbarkale: There are 4 ways in which AWS credentials can be provided:

1. Network interface: Available on EC2 instances and requires no configuration from user.
2. Credentials file in `~/.aws/`: Users with simple AWS setups will have these.
3. Environment variables
4. Command line flags

Out of these, only the last one requires some explicit code. All others are [provided by AWS SDK](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/sessions.html) and require no special implementation. I imagine people will require support for all of these methods (I have used all of them since I started working with AWS last year). I suggest either keeping the documentation or including a link to AWS SDK which explains various options.

vito: Sure, but with BOSH, no one should be doing anything but BOSH properties or EC2 instance metadata. The VMs are hands-off after deploying, and there should be no fussing with `~/.aws` or setting environment variables or anything.

I would not recommend documenting other methods since this is the BOSH release itself. I think it would only confuse people consuming the release.

jobs/atc/spec
```diff
@@ -507,3 +507,38 @@ properties:
     description: |
       Client secret for CredHub authorization.
     default: ""
+
+  aws-ssm.aws_region:
```
vito: I think this should be `aws_ssm.region` (and the following properties should be changed similarly). Underscores are more conventional in BOSH and we removed the duplicate prefix in the CLI.

jobs/atc/spec
```diff
@@ -507,3 +507,38 @@ properties:
...
+    description: |
+      AWS region to use for fetching SSM parameters. Enviroment variable AWS_REGION is used
```
vito: I appreciate that these docs are somewhere but IMO the BOSH release is the wrong place. With BOSH, a user will only ever be setting properties in the manifest, never configuring environment variables or setting up `~/.aws` on the VMs or passing flags to CLIs - properties are a higher level of abstraction, where the user doesn't have to care how they work.

PR #1975, opened 2018-01-22
Added bbr-atcdb job for use with backup-and-restore-sdk-release

rkoster:
inspired by: https://github.com/cloudfoundry/uaa-release/tree/develop/jobs/bbr-uaadb
depends on: https://github.com/cloudfoundry-incubator/backup-and-restore-sdk-release

### DIFF
--- a/jobs/bbr-atcdb/spec
+++ b/jobs/bbr-atcdb/spec
@@ -1,0 +1,56 @@
+---
+name: bbr-atcdb
+
+templates:
+  config.json.erb: config/config.json
+  backup.sh.erb: bin/bbr/backup
+  restore.sh.erb: bin/bbr/restore
+
+packages: []
+
+consumes:
+- name: db
+  type: postgresql
+  optional: true
+- name: postgres
+  type: database
+  optional: true
+  
+properties:
+  postgresql_database:
+    description: |
+      Name of the database to use from the `postgresql` link.
+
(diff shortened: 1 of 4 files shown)


vito:
Would you mind cluing me in on some more details here? How is this used? What does it do?

rkoster:
I would like to be able to use bbr to backup and restore the postgres database, used by atc, with bbr. Since backup and restore of a postgres database is a common usecase the bbr team has created the backup-and-restore-sdk-release, details on using it can be found [here](http://docs.cloudfoundry.org/bbr/bbr-devguide.html#bbr-utilities). So to integrate with the sdk atc should provide a `config.json` with database credentials and some wrapper scripts which invoke the sdk.

jama-pivotal:
For additional context, when @rkoster mentions bbr I believes he means the Bosh Backup and Restore tool https://github.com/cloudfoundry-incubator/bosh-backup-and-restore

shinji62:
+1 for this one. @vito

RomRider:
It would help a lot to use only `bbr` to backup anything bosh related @vito! So +1 for this one!

vito:
I'll just merge this as-is for now since I'm not even sure that the code involved in this PR is what would need to change to resolve the above issue (it seems to just be glue). But I don't think we can document this as "supported" yet.

PR #2027, opened 2018-02-13
Add cookie secure flag.

jmcarp:
Goes with https://github.com/concourse/skymarshal/pull/1 and https://github.com/concourse/atc/pull/249; see patch to skymarshal for details.


vito:
thanks!

Issue #2029, opened 2018-02-14
downgrade_db.erb uses esc function that is not defined


calebwashburn:
Hi there!

Please pick the appropriate section (Feature Request or Bug Report) and delete the rest of this. Note that "missing support for X" is a feature request, not a bug!

# Feature Request

## What challenge are you facing?

Rather than jumping to a solution, describe the problem you're facing, and provide context. It helps us more to understand your problem first. This way we may be able to connect the dots between your issue and others and perhaps come up with a more general proposal, rather than tacking on feature after feature.

## A Modest Proposal

If you've got a suggestion, feel free to place it here - we don't mean to stifle your creativity! It also helps to describe how you arrived at this idea and why it'd work well for you.

# Bug Report

Bug reports are pretty free-form; just replace this with whatever. You can also help us triage the issue by including steps to reproduce, expected results, and the actual result. Help us help you!

The following can also be handy:

* Concourse version:
* Deployment type (BOSH/Docker/binary):
* Infrastructure/IaaS:
* Browser (if applicable):
* Did this used to work?

rkoster:
also run into this issue, while digging I found: https://github.com/concourse/concourse/commit/5814e61e9995dadeaec309a1da8cbaf421cd804e, so apparently a similar issue has passed ci before, maybe time to investigate?

vito:
Ugh. Templatizing Bash with Ruby is a nightmare for reasons like this. I'm guessing the helper functions end up defined globally (as `def` does at the toplevel in Ruby), so if other templates with `def esc` were evaluated prior to that template, it'll silently work.

We'll issue a patch release. Sorry about that.

tahaozket:
@vito Have you released a patch for the bosh release? If you haven't is there any way to install (maybe a release-candidate?) Concourse 3.9 via bosh release without waiting patch? We were eager to try btrfs improvements on our deployment but this issue blocked us :(

PR #2030, opened 2018-02-14
downgrade_db.erb uses esc function that is not defined

calebwashburn:
esc function is not available to to downgrade_db.erb, adding that function.

Opened issue - https://github.com/concourse/concourse/issues/2029


vito:
sorry, didn't see this and pushed https://github.com/concourse/concourse/commit/31d54b646e2b813f8bd938bb5f24ad31bbe36e34 - thanks though!

Issue #1914, opened 2017-12-21
Missing support for generic oauth ca cert in bosh release


scnewma:
# Feature Request

## What challenge are you facing?

We are trying to setup the Concourse bosh release to add support for generic oauth with the Concourse main team.  Our OAuth provider uses a self-signed certificate, and the current bosh release does not seem to have the option to pass in a ca certificate when configuring auth.

## A Modest Proposal

It would be great to have a `generic_oauth.ca_cert` option in the bosh release to allow users to pass in this option when setting up generic oauth via the bosh release.

youngm:
This issue should be fixed now and can be closed.

PR #2064, opened 2018-03-02
Support generic oauth ca cert.

youngm:
PR for issue: https://github.com/concourse/concourse/issues/1914

Adds support for setting the generic oauth ca cert via bosh property.

### DIFF
--- a/jobs/atc/templates/atc_ctl.erb
+++ b/jobs/atc/templates/atc_ctl.erb
@@ -16,6 +16,7 @@ LOG_DIR=/var/vcap/sys/log/atc
 SIGNING_KEY=/var/vcap/jobs/atc/config/token_signing_key
 PIDFILE=$RUN_DIR/atc.pid
 CF_CA_CERT=/var/vcap/jobs/atc/config/cf_ca_cert
+GENERIC_OAUTH_CA_CERT=/var/vcap/jobs/atc/config/generic_oauth_ca_cert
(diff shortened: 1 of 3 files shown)


jobs/atc/templates/generic_oauth_ca_cert.erb
```diff
@@ -0,0 +1 @@
+<%= p("generic_oauth.ca_cert", "") %>
```
cppforlife: dont need "" here if default is set in spec.

youngm:
Thanks @cppforlife I removed `""` from the spec to match most of the other CA config.

jobs/atc/templates/generic_oauth_ca_cert.erb
youngm: Thanks.  Removed "" from spec.

jobs/atc/templates/atc_ctl.erb
```diff
@@ -185,6 +186,7 @@ case $1 in
       <% end %> \
       --generic-oauth-token-url <%= esc(p("generic_oauth.token_url")) %> \
       --generic-oauth-display-name <%= esc(p("generic_oauth.display_name")) %> \
+      --generic-oauth-ca-cert $GENERIC_OAUTH_CA_CERT \
```
vito: this should be passed conditionally since if it's not set it'll be pointing to an empty file, which isn't valid

youngm:
Thanks for catching that @vito  I made the requested change.

vito:
thanks!

PR #2125, opened 2018-04-03
Add mkdir for /var/vcap/data/baggageclaim

z4ce:
Fixes error where baggaeclaim will not start and
baggageclaim_ctl.log has error `chmod: cannot access
 '/var/vcap/data/baggageclaim': No such file or directory`

### DIFF
--- a/jobs/baggageclaim/templates/baggageclaim_ctl.erb
+++ b/jobs/baggageclaim/templates/baggageclaim_ctl.erb
@@ -30,6 +30,7 @@ case $1 in
     chown -R vcap:vcap $LOG_DIR
 
     umask 0022
+    mkdir -p /var/vcap/data/baggageclaim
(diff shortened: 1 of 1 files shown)


vito:
Ack, good catch. Thanks! I wonder how our pipeline and deployment hasn't run in to this.

albertoleal:
@vito and @z4ce Our Concourse is failing to start because of this. Are you planning to cut 3.10.1 with this change?

thanks!

z4ce:
@albertoleal not sure when @vito is going to cut a new release but as a workaround just run `bosh -d concourse ssh worker -c 'sudo mkdir -p /var/vcap/data/baggageclaim'`

albertoleal:
Thanks @z4ce, we're using that workaround in the meantime.

Issue #318, opened 2016-03-16
Support for custom resource in fly execute


tom-meyer:
We just came across an issue when trying to `fly execute` where one of the inputs is a user-defined resource_type. Concourse gives us this for the output of the `get` for our custom git-lfs resource:

```
no workers satisfying: resource type 'git-lfs'

available workers:
  - platform 'linux'
  - platform 'linux'
  - platform 'linux'
  - platform 'linux'
```
Our custom resource is defined like this:
```
resource_types:
- name: git-lfs
  type: docker-image
  source:
    repository: "docker.ci.eng.pivotal.io:443/git-lfs-resource"
```

muv:
+1 just got the same

benmoss:
Yep, not clear how to make this work since this is specified in the pipeline.

mtekel:
+1

combor:
+1

vito:
We had planned on getting to this "soon" at one point but a ton of more urgent things have come up. At this point it'd probably be quicker for someone to contribute this as a PR.

Here's what we were planning to do:

`fly execute -j` should support the custom resource types defined in the job's pipeline.

Let's have it so you can submit a build with a set of custom resource types. This command would then get the config for the pipelines, pluck off the custom resource types, and tack it onto the request body.

When creating a build, rather than just submitting a plan, a JSON payload like the following should be submitted:

```yaml
{
  "resource_types": [], # the resource types
  "plan": {} # the plan
}
```

Where `"plan"` is today's payload, and `"resource_types"` is a `atc.ResourceTypes`: https://github.com/concourse/atc/blob/master/config.go#L67

Should be relatively easy to add; all the internal bits in place, just gotta pass that slice around.

vito:
Update on previous comment: turns out you can already submit a build with a set of versioned resource types, though it works a bit differently: you have to add the whole set of them to each `get`, `put`, or `task` in the plan.

So, all we need now is an endpoint to get that set of resource types from the API. Something like:

```
https://ci.concourse.ci/api/v1/teams/main/pipelines/main/resource_types
```

This would return [`VersionedResourceTypes`](https://github.com/concourse/atc/blob/master/config.go#L87), which is an array of...versioned resource types.

`fly execute` would then make that API call and tack them on to the task, get, and put steps in its build plan.

Not sure about UX for this. `fly execute --resource-types-from <pipeline>`? `-p` for short? (i.e. "execute in the context of this pipeline")

[...]

PR #2153, opened 2018-04-13
for issue #318 support custom resource in fly execute

xtremerui:
While working on this issue I found when providing a local input by -i, it already supports the input in custom type. So the work I needed to do was to fix the error `base resource type not found` when running `fly execute` with `-j` and one of the input is in custom resource type.


xtremerui:
No actual code changes in this parent repo. All submodule changes are made into PR in their repo individually so will close this one for now.

PR #2170, opened 2018-04-20
Add ability to specify default and maximum number of build logs to retain

### LINKED ISSUE #1417: Configurable default build_logs_to_retain
# Feature Request

We’ve been concerned with concourse’s default behavior to keep an indefinite amount of build history by default. 

Because we have some builds that produce a lot of logs, and since we don't always remember to set build_logs_to_retain, concourse falls over for some DB migrations during upgrades. This is because the migration becomes bottlenecked for several minutes/hours by the IO requirements of reading a bajillion lines of build log. 

Setting a build log limit is optional
https://concourse.ci/single-page.html#build_logs_to_retain

And doing so resolves the issue, but I have to do it for every pipeline. 

## A Modest Proposal

The default build log limit is infinity and unchangeable. 

I believe the default build log limit should be configurable globally, by specifying the value that the team that deploys concourse desires to be the default value, whether it's 100, 10000, or -1.

aeijdenberg:
Fixes #1417

### DIFF
--- a/jobs/atc/spec
+++ b/jobs/atc/spec
@@ -563,11 +563,11 @@ properties:
   build_log_retention.default:
     description: |
       Default (can be overriden by job) number of build logs to retain,
-      0 means retain all (database will grow indefinitely).
-    default: 0
+      0 (or not set) means retain all (database will grow indefinitely).
+    example: 100
   build_log_retention.maximum:
     description: |
       If set, this will cap the maximum number of build logs to retain for any job,
       capping any value set in a job itself or the build_log_retention.default.
-      0 means no maximum is specified.
-    default: 0
+      0 (or not set) means no maximum is specified.
+    example: 1000
--- a/jobs/atc/spec
+++ b/jobs/atc/spec
@@ -559,3 +559,15 @@ properties:
     description: |
       AWS SecretsManager secret name template used to resolve team specific secrets.
     default: /concourse/{{.Team}}/{{.Secret}}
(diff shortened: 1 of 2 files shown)


jobs/atc/spec
```diff
@@ -559,3 +559,15 @@ properties:
     description: |
       AWS SecretsManager secret name template used to resolve team specific secrets.
     default: /concourse/{{.Team}}/{{.Secret}}
+
+  build_log_retention.default:
+    description: |
+      Default (can be overriden by job) number of build logs to retain,
+      0 means retain all (database will grow indefinitely).
+    default: 0
```
vito: Rather than having empty defaults at the BOSH spec level, could we omit the defaults here and just conditionally pass the flags? The 0-value interpretation is kind of an implementation detail and could be confusing once it's exposed at higher levels, since really 0 means "don't do this feature".

iirc we tend to just conditionally pass flags for all the other properties.

aeijdenberg: Thanks @vito - fix as suggested (I think).

vito:
thanks!

giva01121:
Hello, I see that this file is no more available. Still is there a way to set a default global property for the builds log retention?

chenbh:
@giva01121 the bosh related stuff has been moved into its own repo: https://github.com/concourse/concourse-bosh-release/blob/master/jobs/web/spec#L1548

PR #2266, opened 2018-06-09
Add properties for emitting to Datadog to ATC

SHyx0rmZ:
These seemed to be missing.

### DIFF
--- a/jobs/atc/spec
+++ b/jobs/atc/spec
@@ -471,7 +471,6 @@ properties:
     description: |
       If configured, detailed metrics will be emitted to the specified Datadog Agent's
       dogstatsd server.
-    default: ""
 
   datadog.agent_port:
     description: |
@@ -481,7 +480,6 @@ properties:
   datadog.prefix:
     description: |
       An optional prefix for emitted Datadog events.
-    default: ""
 
   riemann.host:
     description: |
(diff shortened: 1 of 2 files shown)


jobs/atc/spec
```diff
@@ -467,6 +467,22 @@ properties:
       Environment name to specify for errors emitted to Yeller.
     default: ""
 
+  datadog.agent_host:
+    description: |
+      If configured, detailed metrics will be emitted to the specified Datadog Agent's
+      dogstatsd server.
+    default: ""
```
vito: Don't need a default here since we're using `if_p` in the template.

jobs/atc/spec
```diff
@@ -467,6 +467,22 @@ properties:
...
+
+  datadog.agent_port:
+    description: |
+      Port of the Datadog Agent's dogstatsd server to emit events to.
+    default: 8125
+
+  datadog.prefix:
+    description: |
+      An optional prefix for emitted Datadog events.
+    default: ""
```
vito: Same as above

SHyx0rmZ:
Oh, totally missed the notification for this. Fixed.

vito:
thanks!

SHyx0rmZ:
Thanks for merging!

PR #2315, opened 2018-06-22
mention bindata

pivotal-jamie-klassen:
The means of generating bindata is not uniform across all the repos, so I kept the advice pretty general.

For example, in the web repo technically you can run `scripts/generate-bindata`, but in my experience
it is much more common to run `make clean all` (in fact, I wonder whether the build instructions for elm should be mentioned in this document?). At the moment there is no `generate-bindata` in the skymarshal repo, but I think @pivotal-jwinters may be on the case.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -177,6 +177,8 @@ to synchronize with multiple PRs in flight. We'll take it from there. If your
 changes involve multiple components, though (`atc` and `fly` for example), be
 sure to let us know in each PR.
 
+### Counterfeiter fakes
+
 When adding / modifying methods to the ATC API, you may need to regenerate
 models and data structures used for testing using Counterfeiter. To do so,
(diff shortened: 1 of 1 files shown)


vito:
Good call. How do ya feel about putting this in each repo's `CONTRIBUTING.md` instead though? That way it can be more specific to each repo. I think the toplevel doc is good for establishing the general dev workspace/workflow but then each repo should have its own document.

pivotal-jwinters:
https://github.com/concourse/skymarshal/commit/a659b4250f9a8d5b262acdad960e448e36c7981b adds the generate bindata script in skymarshal

pivotal-jamie-klassen:
Closing, replaced by https://github.com/concourse/atc/pull/288, https://github.com/concourse/web/pull/6, and https://github.com/concourse/skymarshal/pull/3

PR #2447, opened 2018-07-31
make it clear that you need 10 or bcrypt strength

DaxterM:
I got an error when i created a password with a strength lower then 10 in the atc logs. This will make it clear that 10 or higher is required


vito:
thanks!

PR #2451, opened 2018-08-01
Add github.com/mattn/go-shellwords

cappyzawa:
`github.com/mattn/go-shellwords` is used in [concourse/baggageclaim/kernel/kernel_darwin.go#L12](https://github.com/concourse/baggageclaim/blob/df945db3059d4f9ace9284994f37b625c3fd864e/kernel/kernel_darwin.go#L12).

But, It was not included in submodules.

A Compile error occurred when I ran `scripts/test` on macOS.
```
kernel/kernel_darwin.go:12:2: cannot find package "github.com/mattn/go-shellwords" in any of:
        /usr/local/Cellar/go/1.10.3/libexec/src/github.com/mattn/go-shellwords (from $GOROOT)
        /Users/cappyzawa/go/src/github.com/concourse/concourse/src/github.com/mattn/go-shellwords (from $GOPATH)
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -440,3 +440,6 @@
 	path = src/code.cloudfoundry.org/credhub-cli
 	url = https://github.com/cloudfoundry-incubator/credhub-cli
 	branch = master
+[submodule "src/github.com/mattn/go-shellwords"]
+	path = src/github.com/mattn/go-shellwords
+	url = https://github.com/mattn/go-shellwords.git


vito:
It looks like it's not pulled in because:

1. That import is only needed on the Darwin platform, but...
2. The only use of that package comes from a file that only builds on Linux: https://github.com/concourse/baggageclaim/blob/master/baggageclaimcmd/driver_linux.go#L29

So it never gets picked up in the dependency chain.

I think I can just clean the code up so we don't need that dependency anymore. Thanks for bringing this to our attention! I'll close the PR once I'm done.

cappyzawa:
Thanks for description:+1:

PR #2479, opened 2018-08-09
Fix ca_cert templates

ArthurHlt:
Spec was declaring certificates for authentication as `<auth_system>_auth.ca_cert` but all templates doesn't use prefix `_auth` to retrieve certificate.

In addition, `<auth_system>_auth.ca_cert` are typed as `certificate` so `<auth_system>_auth.ca_cert.certificate` in template seems to be necessary.

For the moment, it is impossible to set `ca_cert` on `cf_auth` for example. When setting `cf_auth.ca_cert`, atc can't start and give this kind of error:

```
server: Failed to open connector cf: failed to open connector: failed to create connector cf: no certs found in root CA file "/var/vcap/jobs/atc/config/cf_ca_cert
```

This PR rewrite key to find in template as `<auth_system>_auth.ca_cert.certificate`.

### DIFF
--- a/jobs/atc/templates/cf_ca_cert.erb
+++ b/jobs/atc/templates/cf_ca_cert.erb
@@ -1,1 +1,1 @@
-<%= p("cf.ca_cert", "") %>
+<%= p("cf_auth.ca_cert.certificate", "") %>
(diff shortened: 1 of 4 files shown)


vito:
Oi, yeah those were totally broken before. Thanks for fixing!

PR #2480, opened 2018-08-09
Add LDAP configuration properties for ATC

JamesClonk:


JamesClonk:
I've tested it in a self-built concourse dev-release, the LDAP feature of atc/dex seems to work out of the box. :+1:

jobs/atc/templates/ldap_ca_cert.erb
```diff
@@ -0,0 +1 @@
+<%= p("ldap_auth.ca_cert", "") %>
```
vito: I think this'll need to be `ldap_auth.ca_cert.certificate` since the type for the property is `certificate`.

JamesClonk: @vito Hmm, I think it's fine like that. github_auth, cf_auth, etc. also only directly have `ca_cert`, I just copied their structure. Or is it wrong there?

vito: @JamesClonk They were all actually broken - https://github.com/concourse/concourse/pull/2479 fixed them. :)

JamesClonk:
@vito my bad, just saw the other merged PR (https://github.com/concourse/concourse/pull/2479).
Changed it accordingly. :smiley:

jobs/atc/templates/ldap_ca_cert.erb
JamesClonk: @vito changed it :+1:

vito:
thanks!

PR #2486, opened 2018-08-10
Fix add_local_users to produce correct output for switch

z4ce:
Fixes `add_local_users` in the bosh release. Currently using `add_local_users` results in an error in the `atc.stderr.log` of `server: no connectors specified`. This is caused by a malformed user entry in the atc_ctl template. The current template generates a line like:

```
--add-local-user \{\"concourse\"\=\>\"\$6\$Wbj2fLWp2\$/j4TZfKxDUpUfKIBfpB7ghQPcccNpaa1mX.bptgiednPA9DJpkRQYEdiESziPCvfNSTLVufmD8Marr3W81\"\} \
```

After the patch the line looks like:
```
--add-local-user concourse:\$6\$Wbj2fLWp2\$/j4TZfKxDUpUfKIBfpB7ghQPcccNpaa1mX.bptgiednPA9DJpkRQYEdiESziPCvfNSTLVufmD8Marr3W81
```

### DIFF
--- a/jobs/atc/templates/atc_ctl.erb
+++ b/jobs/atc/templates/atc_ctl.erb
@@ -145,7 +145,7 @@ case $1 in
       --log-db-queries \
       <% end %> \
       <% p("add_local_users").each do |v| %> \
-      --add-local-user <%= esc(v) %> \
+      --add-local-user <%= esc(v.keys[0]) + ":" + esc(v.values[0]) %> \
       <% end %> \
       --auth-duration <%= p("auth_duration") %> \
       --github-client-id <%= esc(p("github_auth.client_id")) %> \


vito:
The property takes a list of `user:pass` pairs, not a hash:

https://github.com/concourse/concourse/blob/master/jobs/atc/spec#L162-L168

We could change it to a hash (seems cleaner) but that would be backwards-incompatible. :/

z4ce:
Yes, the original version template produces a hash which doesn't work. This PR updates the template to produce `user:pass` pairs.

vito:
@z4ce By hash I meant YAML object. You're doing `add_local_users: {foo: bar}`, but the template needs `add_local_users: [foo:bar]` (a list of strings).

z4ce:
Ah, I see now. I will close this.

PR #2621, opened 2018-09-25
don't shell to ssh-keygen to generate rsa keypairs

tkellen:
this resolves an issue where newer versions of openssh (7.8p1-1)
cause these tests to fail.


vito:
thanks!

PR #2622, opened 2018-09-25
force usage of /tmp for postgres socket file

tkellen:
this makes it so the user running tests does not need write
access to /run/postgresql on distros that change it from the
postgresql default of /tmp


vito:
Thanks!

Issue #2588, opened 2018-09-13
GC containers & volumes from the DB if a worker reports that it no longer has them


topherbullock:
when a worker reports a list of existing containers/volumes back to the ATCs, we should diff that set against what the DB thinks should be there and remove the ones the worker didn't report having

This will help with some of the issues when a worker goes away and comes back with different state:
- Task fails with "config file not found" after restarting Docker service #1796
- failed to find created volume in baggageclaim #1550
- non-BOSH worker operation lifecycle #1457

Gotcha ( potential refactoring / documentation ) [gc.Destroyer's methods](https://github.com/concourse/atc/blob/master/gc/destroyer.go#L16-L17) take in a list of _reported_ handles, not the list of handles to destroy :)

vito:
Make sure there are no race conditions during container creation, since a `CREATING` row will be in the db and it may not be on the worker yet. Maybe only remove `CREATED` entries?

cirocosta:
Hey,

@mhuangpivotal and I started working on this issue, having tackled the first part of it (GCing containers).

Our approach is the following:

- on [`ReportWorkerContainers`](https://github.com/concourse/concourse/blob/1fc286e812a80f175eecd1018b42ec280dcf8253/atc/api/containerserver/report.go#L11) (endpoint touched by `tsa` when the [worker's `gc-interval` triggers](https://github.com/concourse/concourse/blob/1fc286e812a80f175eecd1018b42ec280dcf8253/worker/sweeper/sweep_runner.go#L50-L62)), mark the reported containers with `last_seen = now()`;
- on [`ContainerCollector`](https://github.com/concourse/concourse/blob/1fc286e812a80f175eecd1018b42ec280dcf8253/atc/gc/container_collector.go#L47) being triggered, remove missing containers (those that didn't report that they're still alive within a deadline - now set to `gc-interval * 3`).

Some notes:

- should we emit specific metrics regarding these GCd containers?
- should we document that the TTL should not be smaller than the worker GC-interval? Should we enforce that in `atccmd`? (if it's smaller, then we can have a race between `gc` and `reportcontainers`); and
- the migration to add the `last_seen` column does not retroactively mark the containers as last seen, so currently missing containers will never be collected by the GC.

The work can be found at the branch feature/2588, to be pushed after we tackle Volume as well (should follow the same pattern).

Wdyt? Could we be missing an edge case?

thx!

vito:
[...]

PR #2623, opened 2018-09-25
GC containers & volumes if worker reports no longer include them

mhuangpivotal:
concourse/concourse#2588


topherbullock:
Tried this out on `docker-compose` and it doesn't seem to be working:

1. created a new worker "boop" and ran a resource check on it, so it had one container [`fd9c189e-84aa-4f2a-587d-26541cb8435f`]
1. Backgrounded the worker process and removed everything from the garden depot
1. Restarted the worker and peered into the db:
```
concourse=# select handle, worker_name,missing_Since from containers;
                handle                | worker_name  | missing_since
--------------------------------------+--------------+---------------
 ff6565c0-7a50-464f-4b79-ad749bbc8829 | 26a1dd15ab92 |
 fd9c189e-84aa-4f2a-587d-26541cb8435f | boop         |
```
1. After some time missing since was never updated even though `gaol list` (https://github.com/contraband/gaol) also showed an empty Garden

Taking a deeper look into why

topherbullock:
Turns out I need to `git` gud and learn 2 `docker-compose`... 

Confirmed that it does in fact work with the steps above once I was actually running the correct code

Issue #2674, opened 2018-10-12
Emit metrics for locks held in the DB


topherbullock:
## What challenge are you facing?
As an operator its often hard to know when a component stops working because a lock in the DB cannot be acquired ( due to another lock being held, or DB issues)

## What would make this better?
Concourse could emit metrics for the type of lock and object id being acquired and released by the [LockDB](https://github.com/concourse/concourse/blob/master/atc/db/lock/lock.go#L122-L123)

## Are you interested in implementing this yourself?
![AWW YISS](https://media.giphy.com/media/9PsgUfyrNiKK4/giphy.gif)

topherbullock:
@concourse/pivotal Picking this up today for Freedom Friday

vito:
implemented by https://github.com/concourse/concourse/pull/2681

xtremerui:
Tested with local influxdb and grafana. Could see the metric `lock-held` on grafana datasource. Accepted.

cirocosta:
Hey,

Regarding the `objectId` field - is there a use case where one would create a graph that makes use of it?

From my understanding, that's an ID that we generate for specific "pieces" of the DB that we want to lock, right?

Could you please clarify how one can use that field?

thx!

PR #2681, opened 2018-10-12
Add metrics for locks

topherbullock:
More annoying to add than expected, since the metrics package imports the db package (for good reason), so needed to inject a "`LogFunc`" into the `LockFactory`.


topherbullock:
Summoning @TimSimmons and @databus23 for their opinions on using a `GaugeVec` for Prometheus  metrics, since you're both more knowledgable than I am about Prometheus. 😄

atc/metric/metrics.go
```diff
@@ -415,3 +416,79 @@ func (event ResourceCheck) Emit(logger lager.Logger) {
...
+			Attributes: map[string]string{
+				"type":      event.LockType,
+				"object_id": strconv.Itoa(event.ObjectID),
+			},
+		},
+	)
+}
+
+func LogLockAcquired(logger lager.Logger, lockID lock.LockID) {
+	logger.Debug("released")
```
cirocosta: whoops? "acquired" maybe 😁

cirocosta:
Nice!

Something that crossed my mind: would it be interesting to metrify the act of not being able to acquire a lock?

In the sense that we'd be able to quickly spot that a particular type of lock is being requested but never being acquired.

wdyt?

thx!

atc/metric/metrics.go
topherbullock: oops

topherbullock:
> (I went ahead and resolved the conflicts to kick the build again.)
> 
@vito 
THANKS, BOO

jama-pivotal:
linking to #2674

PR #2701, opened 2018-10-18
Don't fetch rows that we won't modify

aeijdenberg:
We're in the middle of trying to track down a really weird issue whereby `atc` appears to hang on startup, but only when (a) we're on Xenial instead of Trusty, (b) we're in this environment, not our other one and (c) when Postgres SSL verification is enabled.

(The above is irrelevant to this PR, but I share the context in case anyone has some ideas to throw at us...)

Along the way I noticed that our `builds` table has 1.5 million entries, each with a `NULL` `engine_metadata` field.

No point fetching each of these rows upon server start when we skip over them a few lines later if a `NULL` value is found.

Now back to find a solution for our original issue...

### DIFF
--- a/atc/db/open.go
+++ b/atc/db/open.go
@@ -129,6 +129,7 @@ func encryptPlaintext(logger lager.Logger, sqlDB *sql.DB, key *encryption.Key) e
 			SELECT id, ` + col + `
 			FROM ` + table + `
 			WHERE nonce IS NULL
+			AND ` + col + ` IS NOT NULL
 		`)
 		if err != nil {
 			return err


vito:
@aeijdenberg Looks unrelated. We'll have to fix that flake. Gonna go ahead and merge this, thanks!

aeijdenberg:
@vito - thanks for merging. By the way, I'm pretty sure we have finally tracked down our perf issue, and it's an interesting one. We believe it's due to a Linux kernel bug in 4.15.0-36 which is what is used by Xenial 97.19 (and maybe others?).

What we were seeing with `tcpdump` is that sometimes, intermittently, maybe halfway through fetching those 1.5M rows, that the tcp window size would reduce to 1, making fetching results from Postgres effectively grind to a slow crawl.

We're going to move to a different version of Xenial that doesn't have this buggy kernel.

PR #2707, opened 2018-10-21
Introduce a global max-in-flight limit

jchesterpivotal:
Essentially, this commit is intended to prevent Concourse from scheduling builds when workers are under heavy load. By doing so it should be possible to reduce the number of deployed workers.

Deliberately excluded features that can be added in future:

* Configurability for setting of container limit and utilisation limit.
* Having workers declare their own container limits or utilisation.
* Scaling utilisation limits according to worker pool size (larger pools can have higher utilisation while retaining safety; variability falls as the workloads are pooled)
* Giving yes/no according to job size -- ie allowing smaller jobs through if there is still room.

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -533,8 +533,8 @@ func (cmd *RunCommand) constructAPIMembers(
 		cmd.ResourceTypeCheckingInterval,
 		cmd.ResourceCheckingInterval,
 		engine,
+		dbWorkerFactory,
 	)
-
 	radarScannerFactory := radar.NewScannerFactory(
 		resourceFactory,
 		dbResourceConfigCheckSessionFactory,
@@ -729,6 +729,7 @@ func (cmd *RunCommand) constructBackendMembers(
 		cmd.ResourceTypeCheckingInterval,
 		cmd.ResourceCheckingInterval,
 		engine,
+		dbWorkerFactory,
 	)
 	dbWorkerLifecycle := db.NewWorkerLifecycle(dbConn)
 	dbResourceCacheLifecycle := db.NewResourceCacheLifecycle(dbConn)
--- a/atc/pipelines/radar_scheduler_factory.go
+++ b/atc/pipelines/radar_scheduler_factory.go
@@ -30,6 +30,7 @@ type radarSchedulerFactory struct {
 	resourceTypeCheckingInterval      time.Duration
 	resourceCheckingInterval          time.Duration
 	engine                            engine.Engine
+	workerFactory                     db.WorkerFactory
 }
 
 func NewRadarSchedulerFactory(
@@ -38,6 +39,7 @@ func NewRadarSchedulerFactory(
 	resourceTypeCheckingInterval time.Duration,
 	resourceCheckingInterval time.Duration,
 	engine engine.Engine,
+	workerFactory db.WorkerFactory,
 ) RadarSchedulerFactory {
 	return &radarSchedulerFactory{
 		resourceFactory:                   resourceFactory,
@@ -45,6 +47,7 @@ func NewRadarSchedulerFactory(
 		resourceTypeCheckingInterval:      resourceTypeCheckingInterval,
 		resourceCheckingInterval:          resourceCheckingInterval,
 		engine:                            engine,
+		workerFactory:					   workerFactory,
 	}
 }
 
@@ -79,12 +82,13 @@ func (rsf *radarSchedulerFactory) BuildScheduler(pipeline db.Pipeline, externalU
 		pipeline,
 		inputconfig.NewTransformer(pipeline),
 	)
+
 	return &scheduler.Scheduler{
 		Pipeline:    pipeline,
 		InputMapper: inputMapper,
 		BuildStarter: scheduler.NewBuildStarter(
 			pipeline,
-			maxinflight.NewUpdater(pipeline),
+			maxinflight.NewUpdater(pipeline, rsf.workerFactory),
 			factory.NewBuildFactory(
 				pipeline.ID(),
 				atc.NewPlanFactory(time.Now().Unix()),
--- a/atc/scheduler/maxinflight/updater.go
+++ b/atc/scheduler/maxinflight/updater.go
@@ -3,37 +3,62 @@ package maxinflight
 import (
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc/db"
+	"math"
 )
 
 //go:generate counterfeiter . Updater
+
+const (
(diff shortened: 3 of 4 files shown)


marco-m:
Hello, as a Concourse operator, it would be helpful if this feature had also some documentation and some metrics (for example how close is my Concourse to being throttled / what is the number of tasks being in this virtual queue) :-)

jchesterpivotal:
I agree that some metrics are the best fast-follow feature. Eg:

* Total pending builds (per second)
* Pending builds due to job max-in-flight, pending due to serial groups, pending due to global max-in-flight (per second)
* Timestamps for each span (in this case, pending->active, ideally by millisecond)

Little's Law predicts that given totals and time spans, it will be possible to calculate average flow rate.

A note: we should not to call this a queue, as that means a guarantee of ordering. We shouldn't promise that, as there are scheduling options that will perform better if Concourse is able to reorder builds. That's part of the reason I chose "global max-in-flight" as my description.

jchesterpivotal:
I agree, metrics are a good fast-follow feature. Eg:

* Pending builds due to serial groups, due to job max-in-flight or due to global max-in-flight (per second)
* Total pending builds (per second)
* Build spans such as pending->active.
* Total containers running (per second)
* Calculated safe max (per second)
* Calculated utilisation (per second)

A note: I don't want to call it a queue, because that has an implicit guarantee of order. We should leave it open for Concourse to adopt other scheduling disciplines in future that reorder builds to improve performance or efficiency. The name "global max-in-flight" was chosen in part for this reason.

itsdalmo:
This PR ensures that no more containers will be scheduled if there is already `(250 x number of workers) * 90%`  containers running - but are there any mechanisms in place to ensure that, if we have two workers, all 450 containers can't be placed on just one of the worker?

jchesterpivotal:
> if we have two workers, all 450 containers can't be placed on just one of the worker?

I attempted no such safeguard because that would require deeper changes to the scheduler. My explicit goal here is to get something simple that is better than the status quo.

> Perhaps it's an idea to call it something other than "global max in flight" also? Since we already have max_in_flight which restricts the parallelism of a specific job, while the functionality in this PR would prevent any new container from being placed if the workers are overloaded? I.e. the two are not related and should not be confused?

You'll note that the moment it's not user-facing at all. I provided no dials. The 90% value arbitrarily; the "correct" value cannot be determined without advance knowledge of the size of the worker pool and the variability of the build load. But it's less than 100%, which is where things reliably become unreliable.

The reason I went for max-in-flight is because that's where I added my logic and because it's analogous. Concourse buffers builds when a job reaches a max-in-flight defined by the pipeline author. Buffering builds because of a global limit seems to fit the same rule.

marco-m:
@itsdalmo 

> are there any mechanisms in place to ensure that, if we have two workers, all 450 containers can't be placed on just one of the worker?

In addition to what @jchesterpivotal already mentioned, remember that there are two container placement strategy available. The default will tend to schedule on the same worker, the other, random, does random placement. We have random placement in production since months and it does what is written on the box: random placement, which is better than nothing.

```
--container-placement-strategy=[volume-locality|random]
[$CONCOURSE_CONTAINER_PLACEMENT_STRATEGY]
Method by which a worker is selected during container placement. (default: volume-locality)
```

itsdalmo:
> In addition to what <at>jchesterpivotal already mentioned, remember that there are two container placement strategy available. The default will tend to schedule on the same worker, the other, random, does random placement.

Thanks! I'm aware of random container placement, but I have not enabled it because I fear that the inconsistent cache-hits would degrade build speeds a lot in certain cases. With >5 workers and random placements, we can't rely on caches for anything right?

marco-m:
I think in this case you have to pick your poison, based on your workload. I chose to distribute more evenly the tasks, since my build tasks can easily bring to their knees any EC2 instance type. It also depends on how long your build lasts compared to how long it takes to download a docker image.

jchesterpivotal:
Yes, worth noting that this change will be mostly effective on the random placement strategy. It probably won't help for the cache-aware placement strategy. That will require some deeper changes to the scheduling logic to be selective about workers and I wanted to keep it as simple as possible on this first change.

vito:
Phew, now that 5.0 is out it's time to dust off these PRs and get back to reviewing. Sorry about the delay!

@topherbullock Where do you think this sits in the grand scheme of queueing in Runtime? Was this blocked on refactoring (noted in #2928)?

jchesterpivotal:
I think #2928 would achieve what I want to achieve, which is preventing the ATC from blindly overloading workers, but would do so so in a more complete fashion. The reason I didn't do something like #2928 to begin with was that (a) I felt this PR would be easier to accept in the knowingly-dumb-but-very-simple form and (b) I wasn't confident making the necessary changes.

I think there's still a place for some of the followups.

topherbullock:
I like this as a way to ensure that new builds aren't scheduled if the cluster is overloaded, but I wonder whether this decision should be pushed deeper into the scheduler and be made for each step of a build which has already been scheduled to run.

I think overall the scheduler should be provided with more information about the resources it has available when making worker selection decisions, or deciding whether to run a step or not. #2928 still doesn't really capture that sentiment, and also suggests a very specific setting for users to tune based on their worker pool size ( which Concourse knows already ), and the resources available to individual workers ( which workers themselves could advertise )

vito:
This is effectively a "pump the brakes if we're at 90% capacity across all workers" as @itsdalmo noted, which will halt scheduling of new builds but won't prevent container over-allocation (especially in the case of a single overloaded worker). It's a simple approach but I think this needs to be pushed further down to really be useful.

For example, if I have a few empty workers and many nearly-full workers it would still end up over-loading the nearly full ones because the empty ones skewed the math. This is a realistic scenario; many deployments have tagged or Darwin/Windows workers which always have way fewer containers than the rest. That could be fixed by ignoring tagged/other-platform workers, but what if the build exclusively targets said workers? You could inspect the `plan` to find out, but some of that information isn't known until runtime (task configs loaded via `file:`)...

Because of these challenges and awkward questions, 'max in flight' ends up feeling like the wrong place to solve this. The real scheduling happens at the step level. We're finally getting around to queuing, the first instance being #3788. This is not yet motivated by resolving runtime resource load, but we should be able to apply our learnings to #3695 at least, which is currently prioritised in the Runtime backlog.

Now that #3695 is prioritised this feels like a stopgap solution which we'd later want to remove, so I'm leaning towards closing this and focusing our efforts on queuing at the step level. :slightly_smiling_face:

jchesterpivotal:
Yep, I think the time for this one passed, now that better alternatives are moving up the backlog. I'm looking forward to the exciting world of build buffering.

Issue #2542, opened 2018-08-29
Concourse doesn't tolerate slightly slow cred managers, making pipelines unreliable (e.g. Vault)


ralekseenkov:
# Bug Report

* Concourse version: both 3.14.1 and 4.0.0 are affected
* Deployment type (BOSH/Docker/binary): Helm
* Infrastructure/IaaS: Kubernetes
* Did this used to work? No

# Steps to reproduce
1. Point Concourse to cred manager under load (which takes a few seconds to respond)

2. Create pipeline which uses secrets in both resources and tasks:
    * resources (e.g. private keys to fetch git repos)
    * tasks (e.g. pass secrets to script as env variables)

3. Trigger pipeline:
```
fly -t target -j pipeline/start -w

started pipeline/start #1

resource1 2c3b6d68fdd9c0f7d80d521a855aa8f58f2585e6c5e5aa07d7bc55bb
Identity added: /tmp/git-resource-private-key (/tmp/git-resource-private-key)
Cloning into '/tmp/build/get'...
Fetching HEAD
152967c commit message
initializing
Finding variable 'consul_token': Get https://vault-host/v1/secret/a/b/c/concourse/pipeline/consul_token: net/http: TLS handshake timeout
errored
```

# Observed behavior
- if Concourse fails to get secrets from Vault in order to call a resource or task, it will fail the corresponding task/job
- for resources (which typically don't fail) it's possible to specify attempts (https://concourse-ci.org/attempts-step-modifier.html). so Concourse would try to get a resource again if Vault doesn't respond in time
- for tasks (where error code is way more important) there is no easy workaround. as https://concourse-ci.org/attempts-step-modifier.html would restart it from "errored" state (unable to get secret) as well as from "failed" state (which will mean 2nd execution of the task)
- this makes an issue in many cases, as Concourse fails on secret retrieval doesn't even get to running a task. while other systems work fine if Vault is slightly slow
- We thought that Vault cache in 4.0.0 might help, but it doesn't 

Desired behavior:
- when Concourse cannot get secret for running a resource or task, it should make a reasonable attempt to retry instead of failing right away
- backend timeout and number of retries should be configurable

ralekseenkov:
@vito any guidance on how this issue should be addressed? we really like to see this in the next release and open to contributing a fix

marco-m:
[...]

PR #2717, opened 2018-10-23
Added config params & retry logic for getting secrets. See #2542

ralekseenkov:
- every manager has SecretRetry with retry params (attempts + interval)
- SecretRetry gets passed down to Get() into each cred manager
- Get() now performs a configured number of retries instead of a single
  lookup and fail
- couple of unit tests added for retries (just for ssm, but they are
  all the same)

### DIFF
--- a/atc/creds/retryable_variables_factory.go
+++ b/atc/creds/retryable_variables_factory.go
@@ -1,0 +1,65 @@
+package creds
+
+import (
+	"fmt"
+	"github.com/cloudfoundry/bosh-cli/director/template"
+	"github.com/concourse/retryhttp"
+	"time"
+)
+
+type SecretRetryConfig struct {
+	Attempts int           `long:"secret-retry-attempts" default:"5"  description:"The number of attempts secret will be retried to be fetched, in case a retryable error happens."`
+	Interval time.Duration `long:"secret-retry-interval" default:"1s" description:"The interval between secret retry retrieval attempts."`
+}
+
+type RetryableVariablesFactory struct {
+	factory     VariablesFactory
+	retryConfig SecretRetryConfig
+}
+
+type RetryableVariables struct {
+	variables   Variables
+	retryConfig SecretRetryConfig
+}
+
+func NewRetryableVariablesFactory(factory VariablesFactory, retryConfig SecretRetryConfig) VariablesFactory {
+	return &RetryableVariablesFactory{factory: factory, retryConfig: retryConfig}
+}
+
+func (rvf RetryableVariablesFactory) NewVariables(teamName string, pipelineName string) Variables {
+	return RetryableVariables{variables: rvf.factory.NewVariables(teamName, pipelineName), retryConfig: rvf.retryConfig}
+}
+
+func (rv RetryableVariables) Get(varDef template.VariableDefinition) (interface{}, bool, error) {
+	r := &retryhttp.DefaultRetryer{}
+	for i := 0; i < rv.retryConfig.Attempts-1; i++ {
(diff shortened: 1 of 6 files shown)


ralekseenkov:
@vito as you recommended in https://github.com/concourse/concourse/issues/2542, adding retries to https://github.com/concourse/atc/blob/master/creds/eval.go#L10:6 revealed a number of issues:
1. it'd be global retry for a given template. e.g. if a template has 20 vars and failure rate for var lookup is 1/20, then these retries wouldn't be very helpful
2. a bigger problem is that eval() operates with interfaces and a very limited set of data, so passing ATC params or cred manager params (like # of retries and interval between retries) seems nearly impossible in this piece of code. hardcoding these values would obviously be pretty bad

so we settled on:
1. introducing a simple retry params struct
2. adding it to every cred manager
3. passing it down to Get()
4. doing retries in Get() on a per-variable basis, using a common method

in order for this to work, we also need to:
* merge https://github.com/concourse/retryhttp/pull/4
* update go.mod to point to a newer version of retryhttp

ralekseenkov:
@vito that makes a lot of sense, I should have done this this way from the very beginning.
pushed a new commit to address this:

* introduced RetryableVariablesFactory and RetryableVariables
* added unit tests for RetryableVariablesFactory
* retry config (attempts + interval) left on per-credmgr basis in case people want to configure different parameters for different credmgrs (if you feel global params would be better, it not a big deal to switch)
* bumped up the version of retryhttps

ralekseenkov:
@vito if we embed SecretRetryConfig into RunCommand/CredentialManagement, I assume retry parameters would become global.

right now it's done on cred mgr basis, i.e.:
```
--vault-secret-retry-attempts
--vault-secret-retry-interval
--credhub-secret-retry-attempts
--credhub-secret-retry-interval
...
```

are you asking to switch to global?

vito:
@ralekseenkov Yeah, I think it's fine to just have it global, since you can only configure one credential manager anyway. I don't think there's a need to configure it separately for each backend.

ralekseenkov:
ok, I switched it to global

```
    Credential Management:
          --secret-retry-attempts
          --secret-retry-interval
```

atc/creds/retryable_variables_factory.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	"fmt"
+	"github.com/cloudfoundry/bosh-cli/director/template"
+	"github.com/concourse/retryhttp"
+	"time"
+)
+
+type SecretRetryConfig struct {
+	Attempts int           `long:"secret-retry-attempts" default:"5"  description:"The number of attempts secret will be retried to be fetched, in case a retryable error happens."`
+	Interval time.Duration `long:"secret-retry-interval" default:"1s" description:"The interval between secret retry retrieval attempts."`
```
vito: Just noticed this is a fixed retry interval - what do you think about using exponential backoff, similar to the Vault manager's login backoff?:

https://github.com/concourse/concourse/blob/78e3a8912f2c1ff413633bf6b22e6d29e95c6acd/atc/creds/vault/reauther.go#L83-L91

Theoretically this would be a bit nicer if the credential manager is failing to handle a deluge of secret requests.

What's here is fine too (better than nothing for sure). Happy to merge this for now and leave exponential backoff for a later PR if you want to investigate it some more.

vito:
This is good to go so I'll just merge this in so you can try it out more easily. If the retry logic proves problematic it should be easy enough to switch to exponential backoff in a followup PR. This is still better than no retry at all. Thanks a bunch! :confetti_ball:

PR #2747, opened 2018-10-31
atc: set BUILD_URL and BUILD_URL_SHORT metadata envvars

norbertbuchmueller:
One often needs an URL to the Concourse build page, eg. to report a failed build to a Slack channel. It's not hard to construct it from the already existing metadata envvars, but that also means that the code that constructs the URL will encapsulate knowledge about the way Concourse build URLs are formulated (and then changing the URL structure of Concourse build pages would break that code).

This PR adds two new metadata envvars:
* BUILD_URL is the "long" URL (the one that contains the team, pipeline and job name) to the build page
* BUILD_URL_SHORT is the "short" URL (that only contains the BUILD_ID)

### DIFF
--- a/atc/engine/step_metadata.go
+++ b/atc/engine/step_metadata.go
@@ -35,5 +35,13 @@ func (metadata StepMetadata) Env() []string {
 		env = append(env, "BUILD_TEAM_NAME="+metadata.TeamName)
 	}
 
+	if metadata.ExternalURL != "" && metadata.TeamName != "" && metadata.PipelineName != "" && metadata.JobName != "" && metadata.BuildName != "" {
+		env = append(env, fmt.Sprintf("BUILD_URL=%s/teams/%s/pipelines/%s/jobs/%s/builds/%s", metadata.ExternalURL, metadata.TeamName, metadata.PipelineName, metadata.JobName, metadata.BuildName))
+	}
+
+	if metadata.ExternalURL != "" && metadata.BuildID != 0 {
+		env = append(env, fmt.Sprintf("BUILD_URL_SHORT=%s/builds/%d", metadata.ExternalURL, metadata.BuildID))
+	}
+
 	return env
 }
--- a/atc/engine/step_metadata_test.go
+++ b/atc/engine/step_metadata_test.go
@@ -24,6 +24,8 @@ var _ = Describe("StepMetadata", func() {
 				"BUILD_NAME=42",
 				"ATC_EXTERNAL_URL=http://www.example.com",
 				"BUILD_TEAM_NAME=some-team",
+				"BUILD_URL=http://www.example.com/teams/some-team/pipelines/some-pipeline-name/jobs/some-job-name/builds/42",
+				"BUILD_URL_SHORT=http://www.example.com/builds/1",
 			}))
 		})
 


pivotal-jwinters:
> but that also means that the code that constructs the URL will encapsulate knowledge about the way Concourse build URLs are formulated

I would argue that this same thing holds true for other components in concourse that shouldn't know anything about `web` routes.

vito:
@pivotal-jwinters That's been my thinking so far in the past few times this has come up. However, I could see this coming down to "which is worse?":

1. Everyone's pipelines manually constructing `web` routes and everyone having to change them when we change the scheme.
1. Concourse's left hand (runtime) knowing slightly too much about what the right hand is doing (web), but at least we can change things in one central location and mitigate (nullify?) the impact this change would have on everyone's pipelines.

For that reason I could see merging this as beneficial, even if it does add a bit of coupling. Maybe we can mitigate that coupling somehow? Kinda hard because it's Elm vs. Go code, but maybe we can come up with something ad-hoc. 🤷‍♂️

norbertbuchmueller:
@vito Exactly. I think it's better to duplicate the knowledge in one place (which now is in the same Git repo as the source of truth, ie. the `web` code lives) than duplicating it in many places (all the resource implementations and pipelines that need to construct the build URL).

> Maybe we can mitigate that coupling somehow?

Can we maybe somehow add a test (that tests against a running `web` instance)?

atc/engine/step_metadata.go
```diff
@@ -35,5 +35,13 @@ func (metadata StepMetadata) Env() []string {
 		env = append(env, "BUILD_TEAM_NAME="+metadata.TeamName)
 	}
 
+	if metadata.ExternalURL != "" && metadata.TeamName != "" && metadata.PipelineName != "" && metadata.JobName != "" && metadata.BuildName != "" {
+		env = append(env, fmt.Sprintf("BUILD_URL=%s/teams/%s/pipelines/%s/jobs/%s/builds/%s", metadata.ExternalURL, metadata.TeamName, metadata.PipelineName, metadata.JobName, metadata.BuildName))
```
vito: Would it make sense to just have one env var, set to either the long URL if it's in a pipeline or the short URL if it's a one-off build? Having to use one or the other could make this difficult to integrate with.

vito:
@norbertbuchmueller Yeah. Maybe this can be mitigated with a `web` acceptance test that uses the [`mock` resource](https://github.com/concourse/mock-resource) to print the value for the env, grabs the URL, and tries to browse to it.

If you're interested in giving that a shot, it would be a new test [here](https://github.com/concourse/concourse/tree/master/web/wats/test) - you would need to configure a pipeline like this:

```yaml
resources:
- name: foo
  type: mock
  source: {}

jobs:
- name: print-env
  plan:
  - put: foo
    params:
      version: foo
      print_env: true
```

...and then you could run a build, parse out the `env: BUILD_URL=(.*)`, and then try to browse to the URL. The test would just need to confirm by looking at the page content that it's viewing a build. (Alternatively I guess you could just check that the URL matches the browser's URL when viewing the build that prints the output.)

vito:
Side note: similar topic just came up here: https://github.com/concourse/concourse/pull/2759#discussion_r234652957

vito:
@norbertbuchmueller Heyo, do you have time to look at this anymore?

norbertbuchmueller:
> @norbertbuchmueller Heyo, do you have time to look at this anymore?

Hi, i'm afraid i cannot promise much. Not in the foreseeable future, at least. :-(

vito:
@norbertbuchmueller No problem, just checking in. I'm going to close this for now then but happy to review another PR in the future!

Issue #2415, opened 2018-07-25
Concourse Seems to Expose Secrets on Web Nodes


ryanaross:
# Bug Report

Bug reports are pretty free-form; just replace this with whatever. You can also help us triage the issue by including steps to reproduce, expected results, and the actual result. Help us help you!

The following can also be handy:

* Concourse version: 3.13.0
* Deployment type (BOSH/Docker/binary): BOSH / concourse-release

Hello Concourse devs, 

Great product, but am having an issue with plaintext communication between web and workers. 

Here is a sample of output from running 
`tcpdump -i eth0 'port 7777' -X`
On a 'web' instance:
```
17:15:25.387058 IP c3b9--asdgd-9f64-4b70-bfd4-903bd604509f.web.default.concourse-airlock.bosh.34000 > 1663-e497-4b98-a2323-8db2-e60b3997cc55.worker.default.concourse-airlock.bosh.7777: Flags [P.], seq 252:713, ack 196, win 229, options [nop,nop,TS val 16542092 ecr 16460566], length 461
	0x0000:  4500 0201 0fd3 4000 4006 f7f2 0a83 0e0a  E.....@.@.......
	0x0010:  0a83 0e22 84d0 1e61 c450 67ac ffeb 8fa9  ..."...a.Pg.....
	0x0020:  8018 00e5 3325 0000 0101 080a 00fc 698c  ....3%........i.
	0x0030:  00fb 2b16 7b22 7072 6f63 6573 735f 6964  ..+.{"process_id
	0x0040:  223a 2230 6236 6437 3735 622d 3732 3034  ":"xxxxxxx-xxxx
	0x0050:  2d34 6232 392d 3563 6235 2d64 6436 3061  -xxxx-xxxx-xxxxx
	0x0060:  3035 6662 6232 3022 2c22 736f 7572 6365  xxxxx","source
	0x0070:  223a 302c 2264 6174 6122 3a22 7b5c 2273  ":0,"data":"{\"s
	0x0080:  6f75 7263 655c 223a 7b5c 2261 6363 6573  ource\":{\"acces
	0x0090:  735f 6b65 795f 6964 5c22 3a5c 2250 5346  s_key_id\":\"X\",
	0x00b0:  5c22 6275 636b 6574 5c22 3a5c 2263 6f6e  \"bucket\":\"con
	0x00c0:  636f 7572 7365 2d62 6163 6b75 7073 5c22  course-backups\"
[...]

PR #2748, opened 2018-10-31
Refactor TSA to better support draining and rebalancing

vito:
This is probably the first substantial refactor the TSA has seen since its creation. The `worker` package is also much cleaner now, as a result of the TSA exposing a higher-level client that it can work with.

User-facing changes:

* Cleaned up the team authorized keys flag to just use a map, so now `--tsa-team-authorized-keys foo=path` is `--tsa-team-authorized-keys foo:path` instead.
* Removes support for directly registering workers. It wouldn't be too hard to add, but we're moving away from that registration flow anyway for security reasons (see #2415).
* TSA logs are no longer forwarded to the worker client. This was pretty confusing before (https://github.com/concourse/bin/issues/34). Instead, the client-side on the worker will log whenever a register/heartbeat event is received. This will be subject to the client-side log level configuration.


topgun/topgun_suite_test.go
```diff
@@ -491,7 +491,7 @@ func waitForWorkerInState(desiredStates ...string) string {
 		}
 
 		return workerName
-	}).ShouldNot(BeEmpty())
+	}).ShouldNot(BeEmpty(), "should have seen a worker in states: "+strings.Join(desiredStates, ", "))
```
topherbullock: Whoa..I had no idea this was a thing.

tsa/client.go
```diff
@@ -0,0 +1,551 @@
...
+	}
+
+	go proxyListenerTo(ctx, baggageclaimListener, opts.LocalBaggageclaimNetwork, opts.LocalBaggageclaimAddr)
+
+	eventsR, eventsW := io.Pipe()
+	defer eventsW.Close()
+
+	events := NewEventReader(eventsR)
+	go func() {
+		for {
```
topherbullock: Should this for also be watching the context for cancellations? Eventually `events.Next()` will return an `io.EOF`, but we could still be receiving and handling events after the context passed to `Register()` has been cancelled.

tsa/client.go
```diff
@@ -0,0 +1,551 @@
...
+
+func handleForwardedConn(ctx context.Context, rConn net.Conn, network string, addr string) {
+	logger := lagerctx.WithSession(ctx, "forward-conn", lager.Data{
+		"network": network,
+		"addr":    addr,
+	})
+
+	defer rConn.Close()
+
+	var lConn net.Conn
```
topherbullock: I'm assuming the "l" in `lConn` is "local"? Can we call it `localConn` to clarify that? Same for `rConn`.

topgun/topgun_suite_test.go
vito: yeah, much more helpful than 'expected string not to be empty' lol

tsa/client.go
```diff
@@ -0,0 +1,551 @@
...
+	for {
+		var err error
+		lConn, err = net.Dial("tcp", addr)
```
topherbullock: Does it make sense to `net.DialTimeout` here in case the local Garden or Baggageclaim are unresponsive? It will just retry again after 1 second, but maybe we should give up after some time?

worker/wait_group.go
```diff
@@ -0,0 +1,30 @@
...
+import "sync"
+
+type waitGroupWithCount struct {
+	sync.WaitGroup
+
+	countMutex sync.Mutex
+	count      int
+}
+
+func (bwg *waitGroupWithCount) Increment() {
```
topherbullock: I'm confused why the ptr receiver is named `bwg`; What's the "b", b?

bin/cmd/concourse/worker.go
```diff
@@ -118,6 +116,30 @@ func (cmd *WorkerCommand) Runner(args []string) (ifrit.Runner, error) {
...
+
+			// ensure we don't use baggageclaim's default retryhttp client; all
+			// traffic should be local, so any failures are unlikely to be transient.
+			// we don't want a retry loop to block up sweeping and prevent the worker
+			// from existing.
+			&http.Client{
+				Transport: &http.Transport{
+					// don't let a slow (possibly stuck) baggageclaim server slow down
+					// sweeping too much
+					ResponseHeaderTimeout: 1 * time.Minute,
```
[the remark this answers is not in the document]
vito: The timeout here is the time it'll wait for the server to respond to HTTP requests, which doesn't include dialing (that's just for estabilishing the connection).

It probably makes sense to have short dial timeouts, too, but this at least is primarily intended to handle the case of a locked-up Baggageclaim server, in which case dialing works fine, but API requests hang. I suppose a dial timeout would help in situations where Baggageclaim has too many open files to accept new connections.

tsa/client.go
vito: This shouldn't watch the context; if events are being sent, something has to be reading them or else the SSH connection may hang (on the Go side). Closing the context is also used as part of the TSA SSH protocol (it sends a signal), which could theoretically trigger subsequent events which may need to be read. Reading until EOF should be enough; it's guaranteed to happen because of the `defer sshClient.Close()` and `defer eventsW.Close()`.

worker/wait_group.go
vito: No idea. I copied this from its previous location (`worker/beacon`). I can rename it.

tsa/client.go
vito: sure

tsa/client.go
vito: Hmm good point but I don't know if it'll defend against much. This is gonna be dialing to a loopback address, and the only reason I can think of for that to hang is if the server has too many open files. In that case the timeout won't really help since it'll just try to dial it again in a second anyway. I'm leaning towards leaving it as-is if not just to have one less timeout value to remember.

Issue #2712, opened 2018-10-22
Add Error Log Metrics


scottietremendous:
**Summary**

For troubleshooting purposes, we should emit a metric for every logger.Error statement on Wings. 

**Solution**
We'll need to display/track this metric through some Dashboard (eg. Grafana+Influx) after we successfully emit the logger.Error metric. 

We'd want to look at the amount of error events over a period of time and setting up an alert around that.

cirocosta:
Hey,

Just some notes,

@YoussB and I were playing around with the idea of registering a new [Sink](https://github.com/cloudfoundry/lager#sinks), something in the lines of:

```go
package main

import (
	"fmt"
	"os"

	"code.cloudfoundry.org/lager"
)
 
type CustomSink struct{}

func (c *CustomSink) Log(f lager.LogFormat) {
	if f.LogLevel == lager.ERROR {
		fmt.Println("ooops, err!", f)
	}
}

func main() {
	customSink := CustomSink{}

	logger := lager.NewLogger("my-app")

	logger.RegisterSink(&customSink)
	logger.RegisterSink(lager.NewWriterSink(os.Stdout, lager.INFO))

	logger.Info("doing-stuff", lager.Data{"informative": true})
	logger.Error("doing-stuff", fmt.Errorf("error!!11"))
	logger.Info("doing-stuff", lager.Data{"informative": true})
}
```

The benefit of that is that we can increment the metrics in a single place and let the periodic collector take care of emitting those metrics.

(still experimenting with it)

cirocosta:
Hey,

Marking this as `paused` right now as I'm not actively working on it.

What's missing there is making sure that we don't extrapolate the cardinality in terms of possible error messages (see more https://github.com/concourse/concourse/pull/2754#issuecomment-443905349).

cirocosta:
Hey, 

Removing the `paused` label as on Friday I just got the code updated with an `opt-in` flag so we can eventually test this out before deciding about the `depth` of the errors to collect (see the discussion here: https://github.com/concourse/concourse/pull/2754#issuecomment-443905349)

Thanks!

cirocosta:
https://github.com/concourse/concourse/pull/2754

PR #2754, opened 2018-11-02
Adds custom log sink to periodically emit error log metrics

cirocosta:
Hey,

Tackling #2712, in the PR we make use of [`RegisterSink`](https://github.com/cloudfoundry/lager/blob/1a7fb7de44c33d258236e250189099e48a4c7430/logger.go#L13) to have the collection of error message metrics in a single place, without disrupting the current sink that gets registered (the `ReconfigurableSink`).

Something to take notice:

- In the case of an emitter like `influx`, each `log.Error` will incur in an emission at the moment that the log is performed (same behavior as the `http` metric collector we have) - this can probably be improved in a further PR (we could batch the writes).

Here's an example panel making use of it:

screen shot 2018-11-02 at 8 37 09 am

<div align="center"><i>The graph above shows the per-message count of errors, but we could also be displaying "per-host" aggregations so we can see a host that's showing an increase in the number of errors.</i></div>

<br />
<br />

Wdyt?

thx!


pivotal-issuemaster:
@cirocosta Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/concourse&pullRequestId=2754)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/concourse&pullRequestId=2754) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@cirocosta Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/concourse&pullRequestId=2754)!

atc/metric/error_sink_collector.go
```diff
@@ -0,0 +1,32 @@
...
+		logger: logger,
+	}
+}
+
+func (c *ErrorSinkCollector) Log(f lager.LogFormat) {
+	if f.LogLevel != lager.ERROR {
+		return
+	}
+
+	c.Lock()
```
cirocosta: Actually, we don't really need this lock 💭

cirocosta:
Hmm, actually, there's a problem:

- if an emitter suddenly can't emit metrics, we get into a loop (emitter logs `log.Error` because it can't emit, generating even more emissions).

Maybe we can have a single error that can be used by all emitters for such case (like, `metric.ErrFailedToEmit`) that we can inspect in the `Log` method (not generating an emission for that case).

vito:
What do you think about making this a periodic `Meter`, instead of one metric emitted on every error log? Similar to the following metrics:

https://github.com/concourse/concourse/blob/e7ee26a3219c7cfb281beba08cde9612a38d8e46/atc/metric/metrics.go#L12-L21

Might not be a huge deal but in situations where there are lots of errors being logged it could help cool off pressure from the metrics sink.

cirocosta:
Hey @vito,

I started with the idea of having a `map[string]Meter` which mapped from error to a particular meter, 
but end up going to a "one-error, one-metric" approach after looking at `HTTPResponseTime`.

My rationale for dropping that map was that maybe it'd be easier to optimize the emissions in a single
place than go on with a `map[string]Meter` as that'd require a bit more of synchronization than if we
did some batching there in the emitter (I thought about creating an issue about that but forgot 😬)

E.g., instead of making an influxdb emit for every emission, we could batch them in chunks and use the
BatchPoints thing with an actual batch of a bunch of points - right now we use batch but for a single point).

Given that Prometheus already does that "implicitly" by aggregating what comes in those `Inc`, `Set` and `Observe`,
eventually we could have all of our metrics moving toward that model.

*ps.: I'm really not familiar with the other emitters (newrelic, datadog and riemman), but I guess they're always
either "like influx", or "like prometheus".*


As that might take some time (and might even not be a good idea lol), I'll give a try at making use of meters 
and see how it goes - maybe there's a better way than what I tried before :thinking:

wdyt?

thx!

vito:
@cirocosta Ah, I didn't realize it was emitting the log messages along with the error; I thought the goal was just to track error counts. Since it's per-message the current approach makes more sense, though I'm a bir worried about event cardinality as there could be a bunch of different log messages (especially as some of them embed pipeline names/etc., which maybe we should clean up).

cirocosta:
Oh, that is true, the set of possible messages are way bigger than I expected (should've looked at it before):

```sh
cd ./atc

ag -Q '.Error("' --ignore "*_test.go" | \
	awk '{print $2}' | \
	awk -F "(" '{print $2}' | \
	sed 's/"//g' | \
	sed 's/,//g' | \
	sort -u | \
	wc -l
352
```

vito:
@cirocosta Yeah, I like the idea of taking the first N segments (for some N that's useful, so we'll probably need to check papertrail).

With one gotcha:

```
~/w/concourse (master) $ ag '\.Session\((?!"[a-z\-]+")'
atc/atccmd/command.go
1355:						logger.Session(pipeline.ScopedName("radar")),
1365:						Logger:    logger.Session(pipeline.ScopedName("scheduler")),
```

These dynamic session names are a terrible idea though and we should just remove them in favor of `lager.WithData`.

I think I'll just...do that now, actually.

vito:
forgot to reference this PR, but i cleaned that up in https://github.com/concourse/concourse/commit/fc7a31a8aada134fd909e2b5c787937ee863dd4d

cirocosta:
Hey, 

Quick update: I've just rebased it and added the `--capture-error-metrics` flag under the `metrics` flag group so one can opt-in if wanting to try out this feature.

Wdyt?

thx!

cirocosta:
Hey,

As an update to the last comment, I pushed a commit that:

- is rebased;
- includes a Prometheus metric for this; and
- adds a feature gate to `atccmd` so that people can opt-in.

With it activated, here's how it looks:

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 1b7dc7b03..0946bf08e 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -28,6 +28,9 @@ services:
     - CONCOURSE_ADD_LOCAL_USER=test:test,guest:guest
     - CONCOURSE_MAIN_TEAM_LOCAL_USER=test
     - CONCOURSE_LOG_LEVEL=debug
+    - CONCOURSE_PROMETHEUS_BIND_PORT=1313
+    - CONCOURSE_PROMETHEUS_BIND_IP=0.0.0.0
+    - CONCOURSE_CAPTURE_ERROR_METRICS=true

   worker:
     build:
```

```sh
curl localhost:1313/metrics -s | grep error_logs
# HELP concourse_error_logs Number of error logged
# TYPE concourse_error_logs counter
concourse_error_logs{message="atc.build-log-collector.tick.acquire.failed-to-register-in-db"} 1
concourse_error_logs{message="atc.build-tracker.track.failed-to-lookup-started-builds"} 1
concourse_error_logs{message="atc.collector.tick.acquire.failed-to-register-in-db"} 1
concourse_error_logs{message="atc.container-collector-worker-job-runner.failed-to-get-running-workers"} 1
concourse_error_logs{message="atc.list-all-pipelines.failed-to-get-all-visible-pipelines"} 1
concourse_error_logs{message="atc.list-resources.failed-to-get-dashboard-resources"} 1
concourse_error_logs{message="atc.list-teams.failed-to-get-teams"} 6
concourse_error_logs{message="atc.pipelines.failed-to-get-pipelines"} 1
concourse_error_logs{message="atc.pipelines.radar.failed-to-get-resource-types"} 1
concourse_error_logs{message="atc.pipelines.radar.failed-to-run-scan-resource"} 3
concourse_error_logs{message="atc.pipelines.scheduler.tick.failed-to-acquire-scheduling-lock"} 1
concourse_error_logs{message="atc.pipelines.scheduler.tick.lock.acquire.failed-to-register-in-db"} 1
concourse_error_logs{message="atc.pipelines.scheduler.tick.scheduling.try-start-next-pending-build.create.failed-to-construct-resource"} 1
concourse_error_logs{message="atc.pipelines.scheduler.tick.scheduling.try-start-next-pending-build.failed-to-fetch-resource"} 3
concourse_error_logs{message="atc.pipelines.scheduler.tick.scheduling.try-start-next-pending-build.failed-to-lookup-creating-container-in-garden"} 1
concourse_error_logs{message="atc.pipelines.scheduler.tick.scheduling.try-start-next-pending-build.failed-to-lookup-volume-in-bc"} 2
concourse_error_logs{message="atc.pipelines.scheduler.tick.scheduling.try-start-next-pending-build.find.failed-to-find-initialized-on"} 2
```

Which, "correlating" to the number of 500 status code, gets us an interesting view of what's going on in the "errors-side".

```sh
curl localhost:1313/metrics -s | grep 'status="500"' | grep count
concourse_http_responses_duration_seconds_count{method="GET",route="ListAllPipelines",status="500"} 1
concourse_http_responses_duration_seconds_count{method="GET",route="ListResources",status="500"} 1
concourse_http_responses_duration_seconds_count{method="GET",route="ListTeams",status="500"} 6
```

I decided to go with the full message for now so we can see in action how bad that gets, although I feel like this won't really be a *big* problem in terms of cardinality for those operating less than hundreds of ATCs.

Wdyt?

thx!

vito:
Sounds good! Kicked off a merge to re-trigger those errored CI builds.

vito:
For some reason GitHub is making me merge master in again, even though it should be able to merge cleanly. Time to wait for those builds again. :roll_eyes:

PR #2755, opened 2018-11-02
Fix fly binary instructions

deniseyu:
add correct `fly` dir, add `go get` step for Go dummies like me

### DIFF
--- a/fly/README.md
+++ b/fly/README.md
@@ -28,7 +28,8 @@ directory to setup your environment.
 3. You can now build the fly binary with go build:
 
   ```bash
-  cd src/github.com/concourse/fly
+  cd fly
+  go get ./...
   go build
   ```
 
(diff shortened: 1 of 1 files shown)


topherbullock:
@deniseyu Thanks for pointing out the issues with the readme! 
The `go get` step is optional, as `go build` will take care of it (tested on `go1.11.1 linux/amd64`)

If you want to clean more mess up for us.. 
This step can be removed entirely, as we no longer need direnv to set the GOPATH
```
2. Install direnv. Once installed you can cd in and out of the concourse directory to setup your environment.
```

vito:
Hmm this document may be a little redundant with the new contributing docs anyway, which cover this here: https://github.com/concourse/concourse/blob/master/.github/CONTRIBUTING.md#building-fly-and-targeting-your-local-concourse

We should probably clear our and consolidate these sub-READMEs before the next release.

deniseyu:
Honored to be the readme janitor

Issue #2745, opened 2018-10-31
Include Status Code to http response metrics


cirocosta:
## What challenge are you facing?

It's not very immediate to realize that a particular route is having trouble (e.g, raising a bunch of 500s to clients).

While https://github.com/concourse/concourse/issues/2712 provides a great view of errors happening in the system (we can keep track of increasing errors), with a metric per-endpoint, we can count the result of such errors as seen by the customers.

## What would make this better?

Add an extra field to the `HTTPResponseTime` metric from the `metric` package.

That'd probably require creating a wrapper for `http.ResponseWriter`, given that we can't retrieve the status code directly from it.

## Are you interested in implementing this yourself?

Sure!

cirocosta:
got it in 5.0! 😁

PR #2756, opened 2018-11-05
atc: Adds http response status code field to http response time metric

cirocosta:
Covering concourse/concourse#2745, this PR is about adding an extra tag to the fields of `HTTPResponseTime` so that we can also aggregate based on specific HTTP status codes of our interest.

### DIFF
--- a/atc/metric/emit.go
+++ b/atc/metric/emit.go
@@ -61,20 +61,28 @@ type eventEmission struct {
 	logger lager.Logger
 }
 
-var emissions = make(chan eventEmission, 1000)
+var emissions chan eventEmission
+
+func Deinitialize(logger lager.Logger) {
+	close(emissions)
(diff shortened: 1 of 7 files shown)


cirocosta:
whoops, just noticed that this is missing the Prometheus counterpart.

Coming with an update soon 👍

vito:
@cirocosta Should I just merge this for now and we can have a follow-up PR for prometheus support?

cirocosta:
That was a quick one, just rebased with the current master and added the field 😁  

thx!

cirocosta:
Thanks @vito for noticing the testflight problems!

@xtreme-sameer-vohra  and I went forward with that lib you suggested to make sure we don't get another of those cast problems; wdyt?

thx!

vito:
NEAT

Issue #438, opened 2016-05-13
Support for cc.xml


darrenhaken:
In order to use existing Build Radiators would Concourse ever support the cc.xml endpoint like other CI tools like Jenkins and GoCd?

vito:
Seems reasonable. We probably won't get to this ourselves, but a PR would be welcome. Is there an actual spec for the XML file format? I couldn't find one from a quick search.

darrenhaken:
@vito the best ive been able to find is: https://github.com/robertmaldon/cc_dashboard#summary

It's what we used with http://nevergreen.io/

darrenhaken:
@vito what is the process for PR's on Concourse? Also what language is it written in?

vito:
@darrenhaken Go. The repo this would apply to is https://github.com/concourse/atc - it would be a new API endpoint.

TimGrayden:
cc.xml is obsolete, and cctray was not updated for the last 5 years.

You can instead implement [CatLight open protocol](https://github.com/catlightio/catlight-protocol) and get amazing [cross-platform build monitor](https://catlight.io) that supports branches & build investigations.

darrenhaken:
This looks more like desktop notifications than a build monitor, am I missing something?






On Sat, Mar 18, 2017 at 10:38 PM +0000, "TimGrayden" <notifications@github.com> wrote:












cc.xml is obsolete, and cctray was not updated for the last 5 years.


You can instead implement CatLight open protocol and get amazing cross-platform build monitor that supports branches & build investigations.

TimGrayden:
You can make the window full screen and put it on a shared display for the whole team to see. But you're right - most people use it as a desktop notifier.

darrenhaken:
@TimGrayden It's not quite the same then as [KB4013429](Nevergreen) which I work on then which is designed for large screens (and also plays sounds). If there are alternatives to cc.xml I'd definitely like to explore them though due to it clearly not being supported in a long time.

cburgmer:
As a workaround https://github.com/BlacktoBlue/ConcourseCCtray might do it for you.

cburgmer:
I'm starting to work on a PR (https://github.com/cburgmer/concourse/commits/master). 

One issue I don't know how to tackle yet is how to handle authentication. AFAIK Concourse speaks OAuth 2.0, but most of the monitoring tools only do Basic Auth. Help welcome.

vito:
[...]

PR #2759, opened 2018-11-05
Support for cc.xml

cburgmer:
This PR tries to implement what's being asked for in #438.

Following the documentation in https://github.com/concourse/concourse/blob/master/.github/CONTRIBUTING.md#running-concourse one should yield the following.

Given the token from `~/.flyrc`:
```
$ curl 'http://localhost:8080/api/v1/teams/main/cc.xml' -H 'Authorization: Bearer YOUR_TOKEN'
```
```xml
<Projects>
  <Project activity="Sleeping" lastBuildLabel="5" lastBuildStatus="Success" 
           lastBuildTime="2018-11-05T20:02:57Z" name="example :: hello-world"
           webUrl="http://localhost:8080/teams/main/pipelines/example/jobs/hello-world"></Project>
</Projects>
```

I've assumed that an aborted build shall be marked as failed. Apologies for mixed tabs and spaces, I've trusted GoLand to choose the right thing to do, it might have failed me horribly.

The only way to currently access the endpoint is via a token retrieved from fly AFAIK (see above example). I wonder what Concourse wants to offer other apps when it comes to retrieving a long running session, preferably with easy user login. Currently I would see this as a blocker for integrating this new endpoint in my build monitor tools of choice.

### DIFF
--- a/atc/api/ccserver/get.go
+++ b/atc/api/ccserver/get.go
@@ -119,7 +119,7 @@ func (s *Server) buildProject(build db.Build, nextBuild db.Build, pipeline db.Pi
 	projectName := fmt.Sprintf("%s :: %s", pipeline.Name(), job.Name())
 	return Project{
 		Activity:		 activity,
-		LastBuildLabel:  fmt.Sprint(build.ID()),
+		LastBuildLabel:  fmt.Sprint(build.Name()),
 		LastBuildStatus: lastBuildStatus,
 		LastBuildTime:   build.EndTime().Format(time.RFC3339),
 		Name:            projectName,
--- a/atc/api/ccserver/get.go
+++ b/atc/api/ccserver/get.go
@@ -4,6 +4,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"net/http"
+	"time"
 
 	"code.cloudfoundry.org/lager"
 
@@ -13,7 +14,9 @@ import (
 
 type Project struct {
 	Name			string `xml:"name,attr"`
+	LastBuildLabel  string `xml:"lastBuildLabel,attr"`
 	LastBuildStatus	string `xml:"lastBuildStatus,attr"`
+	LastBuildTime   string `xml:"lastBuildTime,attr"`
 }
 
 type ProjectsContainer struct {
@@ -97,7 +100,9 @@ func buildProject(build db.Build, pipeline db.Pipeline, job db.Job) Project {
 
 	projectName := fmt.Sprintf("%s :: %s", pipeline.Name(), job.Config().Name)
 	return Project{
-		Name:            projectName,
+		LastBuildLabel:  fmt.Sprint(build.ID()),
 		LastBuildStatus: lastBuildStatus,
+		LastBuildTime:   build.EndTime().Format(time.RFC3339),
+		Name:            projectName,
 	}
 }
--- a/atc/api/ccserver/get.go
+++ b/atc/api/ccserver/get.go
@@ -17,15 +17,15 @@ import (
 type Project struct {
 	Activity        string `xml:"activity,attr"`
 	LastBuildLabel  string `xml:"lastBuildLabel,attr"`
-	LastBuildStatus	string `xml:"lastBuildStatus,attr"`
+	LastBuildStatus string `xml:"lastBuildStatus,attr"`
 	LastBuildTime   string `xml:"lastBuildTime,attr"`
-	Name			string `xml:"name,attr"`
-	WebUrl			string `xml:"webUrl,attr"`
+	Name            string `xml:"name,attr"`
+	WebUrl          string `xml:"webUrl,attr"`
 }
 
 type ProjectsContainer struct {
-	XMLName   xml.Name `xml:"Projects"`
-	Projects  []Project `xml:"Project"`
+	XMLName  xml.Name  `xml:"Projects"`
+	Projects []Project `xml:"Project"`
 }
 
 func (s *Server) GetCC(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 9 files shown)


pivotal-issuemaster:
@cburgmer Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/concourse&pullRequestId=2759)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/concourse&pullRequestId=2759) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@cburgmer Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/concourse&pullRequestId=2759)!

atc/api/ccserver/get.go
```diff
@@ -0,0 +1,136 @@
...
+	webUrl := s.createWebUrl([]string{
+		"teams",
+		pipeline.TeamName(),
+		"pipelines",
+		pipeline.Name(),
+		"jobs",
+		job.Name(),
+	})
+
+	projectName := fmt.Sprintf("%s :: %s", pipeline.Name(), job.Name())
```
vito: would `/` be a more conventional seperator? just judging from screenshots [here](http://ccmenu.org/) and [here](https://docs.travis-ci.com/user/cc-menu/)

atc/api/ccserver/get.go
```diff
@@ -0,0 +1,136 @@
...
+	return Project{
+		Activity:		 activity,
```
vito: nit: needs `go fmt` - won't block merge on this but you may want to configure `go fmt` on save with your editor for future PRs

atc/api/ccserver/get.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	var activity string
+	if nextBuild != nil {
+		activity = "Building"
+	} else {
+		activity = "Sleeping"
+	}
+
+	webUrl := s.createWebUrl([]string{
```
vito: between this and discussion in https://github.com/concourse/concourse/pull/2747 it may make sense to put these URL constructors somewhere more central - maybe on the `db.Build` interface?

atc/api/ccserver/get.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		logger.Error("failed-to-get-all-active-pipelines", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
+	var projects []Project
+
+	for _, pipeline := range pipelines {
+		jobs, err := pipeline.Jobs()
```
vito: It'd be more efficient to use [`pipeline.Dashboard`](https://github.com/concourse/concourse/blob/710a2fbccf0721f6fd000c58ddb8e939cd724370/atc/db/pipeline.go#L539) - this endpoint will be polled semi-frequently across possibly numerous developer machines, so every ounce of performance counts.

atc/api/ccserver/get.go
cburgmer: The `::` probably stems from GoCD. TeamCity seems to have switched from `::` to `/` in newer releases. TravisCI's slash does fit well into their URL schema. No strong opinions here.

cburgmer:
Halfway through the changes I had to upgrade to Golang 1.11.2 to solve their issue 27925 (although I'm unsure whether this helped, or cherry-picking ce8743a9afcd19b). I'm now left with a broken `ginkgo -r -p`, as go insists on touching `./go.{mod,sum}` and failing with `task.go:8:2: unknown import path "gopkg.in/yaml.v2"`. I'm not sure how to resolve this now. Happy to rebase on master if this makes this problem go away.

vito:
@cburgmer Not sure what version of Go you were on before, but if this is your first time with 1.11.x make sure it's cloned outside of your `$GOPATH` since projects using Go modules shouldn't be under that tree anymore.

It's also possible that `ginkgo -r -p` tried to fetch dependencies in parallel and corrupted your module cache (I *think* I've had that happen to me before). Make sure the repo is outside your `$GOPATH`, then try running `go clean -modcache; go mod download; ginkgo -r -p` to reset the fetched modules, fetch them serially (once), and then run the tests.

atc/api/ccserver/get.go
cburgmer: I read this as a TODO for after the PR. If there's anything you'd like me to do here, let me know please.

cburgmer:
Let me know if there's more feedback on the code. As for the topic as such, we can probably continue the conversation on the original issue.

cburgmer:
Just a note as the build has started failing: The merge of this PR is not blocked on the changes, but held back by the maintainers who only wish to maintain this feature should enough folks ask for it (hope I'm representing the sentiment correctly).

vito:
@cburgmer Sorry for leaving this in limbo, we were on the fence about this and let it sit over the holidays. I've pinged @pivotal-jwinters to review, thanks for your patience!

pivotal-jwinters:
@cburgmer did you solve your upstream auth issues with CCMenu?

cburgmer:
@pivotal-jwinters I did not proceed while this PR was pending so far, but doing so now.

Imho this PR should be considered independently of whether CCMenu supports OAuth or not right now, as cc.xml is consumed by more than this one app. Also, do consider that this is a bit of a chicken and egg situation here for both products :)

As for logging in via OIDC into Concourse, I haven't found much documentation. I did reverse engineer fly's exchange. Is using the `fly` clientId the right step forward for now for fetching tokens?

pivotal-jwinters:
Fair enough, just out of curiosity what other services use cc.xml?

As far as automating the OAuth flow, yeah, you can just use fly's public client.

At this point I think we can merge this, but just a heads up we're going to be revamping our APIs so things might change in the future.

vito:
I'm not sure why but the DCO/WIP checks don't seem to have fired for this. @cburgmer would you mind rebasing and pushing to kick everything off again?

cburgmer:
Sure. Rebased and pushed.

vito:
@cburgmer DCO but still isn't happy - looks like you need to run:

```
git rebase HEAD~13 --signoff
```

cburgmer:
Did the sign-off. I cannot reproduce either failures from testflight and unit locally.

vito:
@cburgmer Thanks! They look like flakes; I'll re-trigger.

PR #2767, opened 2018-11-06
WIP: Do not run version check in http context

timrchavez:
Defer the version check to outside the http context to reduce the risk
of the webhook connection from timing out.


pivotal-jwinters:
@timrchavez do you mind re-introducing some of those tests you removed? I think we'd like to at least test the expectations on the `fakePipeline` and `fakeScanner`. You can probably just wrap them in an `Eventually` block.

timrchavez:
Sure!

timrchavez:
@pivotal-jwinters Thanks for taking a look at this PR. I've re-added the tests. My local testing appears :+1:  but it appears that upstream Concourse prs test is pinned to a specific release right now, so it's not being tested there. Am I using `eventually` correctly?

atc/api/resources_test.go
```diff
@@ -1128,17 +1096,15 @@ var _ = Describe("Resources API", func() {
 						})
 
 						It("tries to scan with the latest version", func() {
-							Expect(fakeScanner.ScanFromVersionCallCount()).To(Equal(1))
+							Eventually(fakeScanner.ScanFromVersionCallCount()).Should(Equal(1))
```
pivotal-jwinters: This `Eventually` will block until it either receives the expect result or times out. You shouldn't need the other `Eventually` blocks below since the `fakeScanner` will already have been invoked.

atc/api/resources_test.go
```diff
@@ -1147,44 +1113,28 @@ var _ = Describe("Resources API", func() {
 						})
 
 						It("tries to scan with no version specified", func() {
-							Expect(fakeScanner.ScanFromVersionCallCount()).To(Equal(1))
+							Eventually(fakeScanner.ScanFromVersionCallCount()).Should(Equal(1))
```
pivotal-jwinters: Same here, you only need the first `Eventually`

atc/api/resources_test.go
vito: You also need to remove the `()` from the call otherwise it'll just be polling the result of the first call:

```							Eventually(fakeScanner.ScanFromVersionCallCount).Should(Equal(1))
```

timrchavez: Makes sense. Thanks!

timrchavez:
:shipit: :)?

Issue #454, opened 2016-05-31
Paramaterise task files


ymedlop:
I have a pipeline YAML with configuration variables (that are read correctly) and this file references a task in another YAML file. From this file, I cannot read the configuration variables in the pipeline's scope. I have created an example of this case [here](https://gist.github.com/ymedlop/fe7d9804684971cb24d3f091119acbed ). 

Is there a way for accessing the variables from the second YAML file?

gunta:
+1

zachgersh:
@ymedlop why don't you just move your image definition up into your task? I think this is what you are after:

https://concourse.ci/task-step.html#task-image

vito:
This'll become more viable with https://github.com/concourse/concourse/issues/291 - we may be able to have tasks run in the "context" of a pipeline's credential store/config source.

holycheater:
@zachgersh and I have "volume graph is disabled" error when I try to use `image: `.
https://github.com/concourse/concourse/issues/402#issuecomment-223619301
Someone says here its unsupported.
I'm running concourse worker from an official docker image.

clarafu:
@ymedlop The Concourse v3.3.0 introduced being able to template out the `source:` in the `image_resource` within the task file. Does this solve your issue or were you looking for only using a configuration file with `--load-from-vars`?

eedwardsav:
I have a similar issue as above...

When developing locally, I want to be able to run tasks using local docker images in a local insecure registry.

To accomplish this I can specify a local image_resource:

```
image_resource:
  type: docker-image
  source:
    repository: local-ci:5000/my/docker/image
    insecure_registries: ["local-ci:5000"]
    tag: latest
```

or I can leave out `image_resource` entirely, and specify it on the pipeline

**The problem with that approach** is that then I can't run the task without a pipeline, and the pipeline requires a proper input resource like a commit, which means to test a small change locally I still have to:
- commit changes to repo
- wait for local ci to see it (slow)
- execute it
- repeat...
when instead, it's a lot easier to just run the task using `fly execute` and ship the source files directly to it

The only approach I see right now to deal with this is to maintain **two copies** of my tasks, one for 'local dev' and one for 'central ci'... however, I consider this an anti-pattern, since I want to use the same set of config files for local or remote CI and simply bind them differently depending on the environment

(e.g. locally I want to use my local docker repo, but on a central ci server I'll push images to a central docker repo... however the task run logic should be the same... so why would I want to maintain it twice??)

[...]

Issue #2560, opened 2018-09-06
Allow a pipeline config to pass params to a task which are not defined in the task.yml


ljfranklin:
In Concourse v5.0 there is a breaking change planned where Concourse will raise an error if the pipeline config attempts to pass a param (environment variable) to a task if the task.yml does not explicitly define that param. This behavior is allowed in the current version of Concourse, but this causes some UX issues as Concourse is unable to verify typos in pipeline params and the current plan is to add validation on undeclared params in v5.0: #2510.

However, we currently have a use case for relying on specifying params in the pipeline that are not defined in the `task.yml`. We're creating a re-usable set of tasks that will be used across many pipelines. We want the authors of those pipelines to be able to specify env vars with a given prefix, e.g. `BOSH_some_variable`, and the `task.sh` will consume all env vars starting with that prefix similar to `bosh --vars-env BOSH interpolate some-template.yml`. So we have a shared task authored by one team with the pipeline env vars and the config template being authored by other teams. Would y'all consider somehow supporting this use case going forward?

As an initial proposal, the `task.yml` could opt into allowing undeclared params to be passed through to the task script:

```yaml
---
platform: linux

params: {}
allow_undeclared_params: true # <-------- add this key to opt back into old behavior

image_resource:
...
```

ljfranklin:
> allow params matching specific patterns (e.g. BOSH_*)

That would work for our use case, but I worry we're hitting diminishing returns in terms of validation at that point. It would catch a typo like `BSH_` but there might be other use cases where you do want to allow any params to be forwarded through regardless of prefix.

eedwards-sk:
I'm running into this same issue, as I'm working on some task files that need better parameterization (something that is sorely lacking with tasks, yet much better with resources... hmmm...)

I'm trying to do something similar, where I pass multiple values similar to how terraform does it with `TF_VAR_<variable>`, I'll be doing `BACKEND_CONFIG_<var>=<value>` so that I can turn it into a series of arguments `-backend-config=<var>=<value>`

Because the names of these params cannot be known in advance, and because there's no easy or good way to pass an array of values in through params, this change would negatively impact my workflow.

ogradyjd:
This is a non-started for my group as well.  We are creating a list of parameters in our pipeline config of which sub-sets are used by different tasks.  You are basically adding cruft if you are going to force every task to define all of those parameters now under threat of erroring out.   That means tasks developed by other teams would be forced to needlessly define those parameters with no real benefit.

[...]

Issue #2075, opened 2018-03-08
Setting pipeline credentials via `fly`


vito:
# Feature Request

## What challenge are you facing?

Right now I have to SSH on to my Vault VM and run `vault` commands from there. It's hokey and error-prone. Not having any simple and consistent way to do this makes credential management feel like a second-class feature, only accessible to operators.

## A Modest Proposal

Let's add some sort of `fly` command for setting credentials. And *maybe* listing and fetching them, but I'm a bit more hesitant to do that for security reasons (sending a single `write` is one thing, transmitting all the credentials out of the manager and over the network is another).

vito:
Good point!

baptiste-ls:
For a more user-friendly interface there's the [goldfish](https://github.com/Caiyeon/goldfish) project that supports many authentication types.

Also clients of Hashicorp have access to their Vault UI feature ($$$$).

vito:
Seems like there's not much pressure for this so I'll just close. I do want to make sure we're making a concerted effort to encourage users to use a credential manager, but from a security standpoint this may not be the best way. Documentation is a start.

PR #2770, opened 2018-11-08
added support for local task development and task params, see #454

ralekseenkov:
this fixes https://github.com/concourse/concourse/issues/454

it adds support for task parametrization, as well as extends 'fly execute' allowing to pass parameters to tasks

if you apply these changes and build concourse from source, you can use the following pipeline to test passing parameters to external tasks:
https://github.com/Aptomi/concourse-pipelines/tree/master/task-params-github-issue-454


atc/exec/task_config_source.go
```diff
@@ -121,9 +123,19 @@ func (configSource FileConfigSource) FetchConfig(logger lager.Logger, repo *work
 		return atc.TaskConfig{}, err
 	}
 
+	// process template of an external task, using vars provided from the pipeline
+	varsContents, err := yaml.Marshal(configSource.Plan.Vars)
```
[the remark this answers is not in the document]
ralekseenkov: done

fly/commands/execute.go
```diff
@@ -42,10 +46,14 @@ func (command *ExecuteCommand) Execute(args []string) error {
 		return err
 	}
 
-	taskConfigFile := command.TaskConfig
 	includeIgnored := command.IncludeIgnored
 
-	taskConfig, err := config.LoadTaskConfig(string(taskConfigFile), args)
+	taskConfigBytes, err := setpipelinehelpers.ATCConfig{}.NewConfig(command.TaskConfig, command.VarsFrom, command.Var, command.YAMLVar, false, false)
```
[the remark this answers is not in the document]
ralekseenkov: done

ralekseenkov:
@vito added another commit with refactoring

Overall - I agree, these helper packages don’t look too good. I made another attempt and tried to clean up/refactor the code to as much as I could:

* consolidated all templating code under atc/template (didn’t seem logical to put this package on the top level). preserved all logic for old-style templates, it wasn't a big deal
* removed redundant code - things like variables.go and file_vars_source.go. moved unit tests over to common template resolver under atc/template
* execute and set-pipeline now share a YamlTemplateWithParams helper, which helps translate command line params (-v, -y, -l) into template resolver params
* StaticConfigSource and FileConfigSource for tasks now only accept the required parameters instead of the whole task plan
* fixed interface of TemplateResolver, so marshaling of task vars is no longer required	
* moved some message printing helpers to display helpers, so they can be invoked from the right places
* speaking of setpipelinehelpers/ATCConfig… I removed quite a few methods from it, which didn’t belong there. so it’s a step forward toward getting rid of this struct all together

Overall, I think it’s much better now

atc/template/template_resolver.go
```diff
@@ -0,0 +1,79 @@
...
+	temp "github.com/concourse/concourse/fly/template"
+	"gopkg.in/yaml.v2"
+)
+
+type TemplateResolver struct {
+}
+
+func (templateResolver TemplateResolver) Resolve(configPayload []byte, paramPayloads [][]byte, flagVars template.StaticVariables, allowEmptyInOldStyleTemplates bool) ([]byte, error) {
+	var err error
+	if temp.Present(configPayload) {
```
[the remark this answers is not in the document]
ralekseenkov: agreed. not importing fly anymore here

atc/plan.go
```diff
@@ -100,6 +100,7 @@ type TaskPlan struct {
 
 	ConfigPath string      `json:"config_path,omitempty"`
 	Config     *TaskConfig `json:"config,omitempty"`
+	Vars       Params      `json:"vars,omitempty"`
```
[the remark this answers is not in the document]
ralekseenkov: yes

ralekseenkov:
rebased on top of the master

ralekseenkov:
@vito
* added testflight tests for external tasks (via set-pipeline and fly execute)
* added testflight test for external tasks with uninterpolated vars

overall, it's sad that there are no tests with mock cred managers which cover merging of CLI+cred mgr variables. not something that I can add easily

one remaining issue that I see is "fly execute" still doesn't fail early on tasks with uninterpolated variables. I don't have enough knowledge on how these one-off tasks are processed. may be you can help:
https://github.com/concourse/concourse/pull/2770/commits/16db4167dd245b6db5e8844e2edb6600cb98b013#diff-ec5e33929d580bba12daf7cb4c26be66R111

or may be we can leave it as is for the time being. this pull request is getting big and I'd really like to get it merged sooner rather than later

ralekseenkov:
@vito ok, I think I managed to get everything working on a long flight and pushed the last part of the fix, which unifies processing of "embedded" and "external" tasks.

summary:
* refactored the code which overrides task parameters. replaced MergedConfigSource with OverrideParamsConfigSource. completely removed all code responsible to merging task configs (copying over params from one struct to another was not very pretty... and not needed)
* unit tests became much clear. they no longer have to account for different combinations of A<->B merging, where one's config is nil, or one's params is nil, etc. as we don't need to merge configs anymore
* garden_factory, became slightly more clear and no longer has a complicated "if" statement
* refactored unit tests, ensuring that all cases of param overrides testing are moved to OverrideParamsConfigSource
* validateCredParams changed accordingly to correctly validate embedded and external tasks
* StaticConfigSource now interpolates template variables, as well as FileConfigSource. so basically FetchConfig() always returns task config with all template variables completely interpolated. TaskStep no longer has to deal with interpolation of task pieces
* fixed my last remaining test under testflight, so both "fly trigger-job" and "fly execute" fail consistently and early 

I also want to explicitly point out that I added a warning to OverrideParamsConfigSource, which says that vars is preferred over params (it will be displayed when when pipeline passes params to a task via 'params'). Params is basically a subset of what vars can do. We should encourage everyone to start using vars and advertise only one preferred way of doing things (not two). That will also drive up reusability of the tasks. If you have any concerns over deprecating params, let's discuss in Discord.

P.S. All tests seem to be passing locally with these changes.

fly/commands/internal/templatehelpers/yaml_template.go
```diff
@@ -0,0 +1,133 @@
...
+
+	evaluatedConfig, err := template.NewTemplateResolver(config, params).Resolve(false, allowEmpty)
+	if err != nil {
+		return nil, err
+	}
+
+	return evaluatedConfig, nil
+}
+
+func (yamlTemplate YamlTemplateWithParams) Validate(
```
vito: This feels out-of-place here since it's specifically doing pipeline config validation, and this object is used for both task configs and pipeline configs.

There's also CLI UI stuff in here. To be honest it probably should have just lived in the `validate-pipeline` command implementation. There are unit tests for this, but it looks like there are integration tests for `validate-pipeline` already, so maybe those are enough (or can be consolidated if not)?

atc/exec/task_step.go
```diff
@@ -310,7 +309,7 @@ func (action *TaskStep) containerSpec(logger lager.Logger, repository *worker.Ar
 	} else if config.ImageResource != nil {
 		imageSpec.ImageResource = &worker.ImageResource{
 			Type:    config.ImageResource.Type,
-			Source:  creds.NewSource(action.variables, config.ImageResource.Source),
+			Source:  creds.NewSource(boshtemplate.StaticVariables{}, config.ImageResource.Source),
```
vito: :+1: Makes sense, since this will now be resolved as part of fetching the config.

atc/exec/task_config_source.go
```diff
@@ -133,40 +132,51 @@ func (configSource FileConfigSource) Warnings() []string {
...
-func (configSource *MergedConfigSource) Warnings() []string {
-	warnings := []string{}
-	warnings = append(warnings, configSource.A.Warnings()...)
-	warnings = append(warnings, configSource.B.Warnings()...)
-	warnings = append(warnings, configSource.MergeWarnings...)
-
-	return warnings
+func (configSource OverrideParamsConfigSource) Warnings() []string {
+	if len(configSource.Params) > 0 {
+		return []string{fmt.Sprintf("overriding task parameters via 'params' is deprecated. use 'vars' instead")}
```
vito: :thinking: I don't think `vars:` really replaces `params:`, or at least we haven't decided to deprecate it at this point. Params will still be useful for things like tasks that have default values, plus allowing arbitrary env config (https://github.com/concourse/concourse/issues/2560):

**task.yml**

```
---
platform: linux

# ...

params:
  FOO: bar
```

**pipeline.yml**

```
task: foo
file: thing/task.yml
params: {FOO: buzz}
```

With `((vars))` it's all-or-nothing - defaults can't be specified.

I think it's safer at this point to introduce `vars:` and see how they're used for now. The concourse/rfcs#7 (future RFC) will be a good opportunity to reassess everything.

atc/api/configserver/save.go
```diff
@@ -165,48 +168,66 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
...
+					errs = multierror.Append(errs, err)
 				}
 
-				_, err = creds.NewTaskParams(vars, plan.TaskConfig.Params).Evaluate()
+				err = creds.NewTaskVarsValidator(credMgrVars, plan.TaskVars).Validate()
+				if err != nil {
+					errs = multierror.Append(errs, err)
+				}
+
+			} else if plan.TaskConfig != nil {
```
vito: :thinking: Supporting `vars:` with `config:` feels like more trouble than it's worth, because it's not really obvious which vars will be satisfied at `set-pipeline` time and which vars will be satisfied by `vars:` at runtime. In fact I could see this shooting devs in the foot if a var in `set-pipeline` unintentionally takes precedence over one that's also specified in `vars:`.

I think it would be simpler to only permit `vars:` with `file:`. This would also reduce the dependency between on `/exec` here which feels a little more coupled than usual:

```
~/w/concourse (master) $ ag /exec atc/api/
~/w/concourse (master) $
```

atc/exec/task_config_source.go
```diff
@@ -26,42 +28,32 @@ type TaskConfigSource interface {
...
-		return taskConfig, nil
+	byteConfig, err := json.Marshal(taskConfig)
+	if err != nil {
+		return atc.TaskConfig{}, fmt.Errorf("failed to marshal task config: %s", err)
 	}
 
-	if taskConfig.Params == nil {
-		taskConfig.Params = map[string]string{}
+	// process task template, using cred manager vars
+	byteConfig, err = template.NewTemplateResolver(byteConfig, configSource.Vars).Resolve(true, true)
```
vito: Minor point: this marshal-resolve-unmarshal dance would also go away if we restrict `vars:` to only work with `file:`.

atc/exec/garden_factory.go
```diff
@@ -135,21 +136,23 @@ func (factory *gardenFactory) Task(
...
+		// external task
+		taskConfigSource = FileConfigSource{ConfigPath: plan.Task.ConfigPath, Vars: taskVarsWithCredMgr}
+	} else {
+		// embedded task
+		taskConfigSource = StaticConfigSource{Config: plan.Task.Config, Vars: taskVarsWithCredMgr}
 	}
 
-	taskConfigSource = ValidatingConfigSource{ConfigSource: taskConfigSource}
+	// override params
+	taskConfigSource = OverrideParamsConfigSource{ConfigSource: taskConfigSource, Params: plan.Task.Params}
```
vito: :+1: Definitely simpler than `MergedConfigSource`, thanks!

testflight/task_vars_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+inputs:
+  - name: some-resource
+
+outputs:
+  - name: unwrapped-task-resource
+
+run:
+  path: sh
+  args: ["-c", "cat some-resource/task.yml | sed -e 's/START_VAR/((/' | sed -e 's/END_VAR/))/' > unwrapped-task-resource/task.yml"]
```
vito: Ha, yeah that's annoying, I can see why it's necessary.

Maybe we could extend the `mock` resource to allow base64-encoded file blobs or something for situations like this. (Don't worry about that in this PR, it's something we can clean up later. :slightly_smiling_face:)

atc/exec/task_config_source.go
ralekseenkov: agreed

atc/api/configserver/save.go
ralekseenkov: agreed. I don't know how to fully remove dependency on exec(), but agree that we shouldn't support vars for embedded tasks. will fix

atc/exec/task_config_source.go
ralekseenkov: ok, makes sense. let's keep them both

ralekseenkov:
@vito added commits to address your comments:
* pipeline config validation moved away from yamltemplate (thanks for spotting this)
* vars and plan params are supported only for file-based tasks
* keep both vars and params around without deprecating
* put back old warning for params into OverrideParamsConfigSource (<..> was defined in pipeline but missing from task file)
* code/UT cleanup

atc/exec/task_step.go
```diff
@@ -319,11 +318,6 @@ func (action *TaskStep) containerSpec(logger lager.Logger, repository *worker.Ar
 		imageSpec.ImageURL = config.RootfsURI
 	}
 
-	params, err := creds.NewTaskParams(action.variables, config.Params).Evaluate()
```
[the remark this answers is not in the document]
ralekseenkov: Thanks for catching this. I think there are no tests which cover these cases (e.g. cred manager variables interpolated in runtime, and then passed to external tasks via params). I think overriding and then interpolating is the way to go.

We can do it via:
FileConfigSource (remove interpolation from it) -> OverrideParamsConfigSource -> InterpolateTemplateConfigSource

Should be pretty clean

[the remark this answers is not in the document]
vito: Thankfully we do have tests for this, but it's in Topgun which doesn't run for PRs:

https://github.com/concourse/concourse/blob/master/topgun/pipelines/credential-management.yml#L87-L90

I 100% agree with your earlier mention that it'd be nice to have a more efficient suite to cover these things. I think https://github.com/concourse/concourse/issues/2075 plus a dummy dabase-backed credential manager would open the door for that should we decide to go through with it. (There are risks involved there, of course - my thinking is it would be a 'gateway drug' to real credential management, but I wouldn't want to compromise the security benefits people enjoy with the current approach.)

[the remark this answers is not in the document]
ralekseenkov: ok, that's definitely good to know! I believe topgun relies on bosh (I can understand why). either way, I wasn't able to run topgun successfully on a local machine. so that's why I was asking in Discord in the first place what's the exact procedure/exact set of sets a person should run locally for a big change before opening a PR

PR #2771, opened 2018-11-08
Adds ./hack/trace script

cirocosta:
Hey,

from the commit message:

```
The commit introduces the script `./hack/trace` whose
goal is to make debugging a running `web` or `worker` instance
possible through the use of [dlv][1].

It updates `Dockerfile` and `ci/Dockerfile` in order to not make
use of the default optimizations so that `dlv` can leverage the
most of the running process.

The script is capable to attach to either a worker or a web instance
by sharing the PID namespace of those. It can also display the source
code of the lines that match a given trace/breakpoint by having the
source code mounted as a volume in the expected locations.
```

One great enablement of using something like `dlv` during development is the ability to not need to stop a process and add `print`s everywhere - attach a break/tracepoint and there you go. 

Regarding the code itself, given that the Dockerfile under the root and `ci` are meant for development/testing only, I guess having those extra flags are not a problem, or am I wrong regarding that?

wdyt?

thx!

[1]: https://github.com/derekparker/delve

---

screen shot 2018-11-07 at 9 28 16 pm

### DIFF
--- a/hack/trace
+++ b/hack/trace
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+set -o errexit
+
+readonly DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+container_name=""
+
+case $1 in
+web)
+  container_name="concourse_web_1"
+  ;;
+
+worker)
+  container_name="concourse_web_1"
+  ;;
+
+*)
+  echo "Usage: trace (web|worker)"
+  exit 1
+  ;;
+esac
+
+tracee_pid=$(docker exec $container_name pidof concourse)
+
+docker build --tag dlv ./dlv
+
+docker run \
+  --interactive \
+  --pid=container:$container_name \
+  --privileged \
+  --rm \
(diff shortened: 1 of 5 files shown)


hack/trace
```diff
@@ -0,0 +1,35 @@
...
+  ;;
+esac
+
+tracee_pid=$(docker exec $container_name pidof concourse)
+
+docker build --tag dlv ./dlv
+
+docker run \
+  --interactive \
+  --pid=container:$container_name \
```
vito: NEAT

hack/trace
```diff
@@ -0,0 +1,35 @@
...
+readonly DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+container_name=""
+
+case $1 in
+web)
+  container_name="concourse_web_1"
+  ;;
+
+worker)
+  container_name="concourse_web_1"
```
danilopopeye: shouldn't be `concourse_worker_1` here?

Issue #236, opened 2015-12-17
Feature Request: Fetching the last N versions of a resource


dsabeti:
We're building a pipeline that needs to consume the 3 most recent versions of a resource. We were originally thinking that the "version" maintained by concourse internally would simply be a list of versions that the resource itself captures. Something like this:

| git | concourse version |
| --- | --- |
| a | [ a ] |
| b | [a, b] |
| c | [a, b, c] |
| d | [b, c, d] |

But perhaps a better way to do this is to simply request the last N versions from concourse itself. It would be great if we could write the following in our build plan:

``` yaml
- get: myResource
  num_versions: 3
```

with `num_versions` defaulting to 1.

Maybe people would prefer this:

``` yaml
- get: myResource
  additional_versions: 2
```

Another thing to consider is allowing the build plan choose a specific version relative to the most recent one. For example, getting the **_second**_ most recent version might look something like this:

``` yaml
- get: myResource
  versions_back: [ 1 ]
```

with `versions_back` defaulting to `[ 0 ]`. This would be useful if a resource version is yanked. There is something appealing about this solution, because it solves the first problem as well (you could specify `versions_back: [0, 1, 2]` to get the three most recent versions).

Anyway, we don't have a strong preference for naming, but these features, especially the first part (getting multiple versions) would be very helpful. Any thoughts?

jama-pivotal:
Hey there!

We're automatically closing this issue since the original poster (or another commenter) hasn't been active in the past six months. We therefore assume that this issue has been lost or resolved on its own. Close issues that remain inactive for a long period may get automatically locked.

Don't worry though; if this is in error, let us know with a comment and we'll be happy to reopen the issue.
Thanks!

dsabeti:
Hey there! I've moved on from the team that needed this in the first place, so I don't myself need the feature. Still, it seems like a reasonable feature request, so it would be nice for the issue to be re-opened, or to get a definitive "Will not implement" from the Concourse team.

Thanks!

PR #2783, opened 2018-11-12
fly: Add timestamps to build logs that print out from watch command

pivotal-kahin-ng:
Add timestamps to build logs to mimic build log data on web UI when running watch command.
As described in [#236](https://github.com/concourse/fly/issues/236).

### DIFF
--- a/fly/eventstream/render.go
+++ b/fly/eventstream/render.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"time"
 
 	"github.com/concourse/concourse/atc/event"
 	"github.com/concourse/concourse/fly/ui"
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@pivotal-kahin-ng Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/concourse&pullRequestId=2783)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=concourse/concourse&pullRequestId=2783) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@pivotal-kahin-ng Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=concourse/concourse&pullRequestId=2783)!

vito:
Thanks, this is a great idea! Maybe we should make this configurable? I could see it being somewhat noisy or unexpected if folks are using `fly watch` for things like batch `grep`ing.

I think there's also a good chance that this mangles output, as the log events aren't necessarily line-by-line. You could end up with timestamps printed either in the middle of a long line that was split into multiple events, or you could end up with only one timestamp printed for an event that had a bunch of lines.

I think you'd need some sort of writer that will keep track of timestamps as it receives events and injects the current timestamp just after each `\n` character. Kind of similar to Ginkgo's prefixed writer (...but please don't literally use that since it's meant for tests :stuck_out_tongue:): https://godoc.org/github.com/onsi/gomega/gexec#PrefixedWriter

pivotal-kahin-ng:
Thanks for the feedback! I'll investigate further

marco-m:
@pivotal-kahin-ng from somebody who resorts to take screenshots of the Concourse web UI to get the timestamps to document performance problems, you just won one case of beer :-) Thanks!

Issue #2555, opened 2018-09-04
`/api/v1/resources` and `/api/v1/jobs` return null instead of empty list


pivotal-jamie-klassen:
# Bug Report

When accessing these endpoints when unauthenticated or with no pipelines, the response body is just `null`. This makes life perhaps needlessly harder for the Elm code.

The following can also be handy:

* Concourse version: 4.1
* Deployment type (BOSH/Docker/binary): binary
* Infrastructure/IaaS: local
* Did this used to work? no, these endpoints were recently added for the dashboard

vito:
done by #2789!

PR #2789, opened 2018-11-14
atc: Return empty array instead of null

pivotal-kahin-ng:
Fix for [#2555]. 

Always creating an empty slice instead of zero-length slice.


vito:
thanks!

PR #2830, opened 2018-11-19
update contrib docs to mention GO111MODULE

ralekseenkov:
otherwise fly binary won't be built successfully


vito:
Thanks! This shouldn't be necessary though as long as you clone the repo outside of your `$GOPATH` - we should probably document that instead to prevent further confusion.

vito:
Went ahead and updated the document a bit.

PR #2858, opened 2018-11-22
fly: Add timestamps to build logs that print out from watch command

pivotal-kahin-ng:
Messed up my branching from this [PR](https://github.com/concourse/concourse/pull/2783)

@vito Is this what you had in mind when you mentioned configurability?


fly/eventstream/render.go
```diff
@@ -11,7 +11,11 @@ import (
 	"github.com/fatih/color"
 )
 
-func Render(dst io.Writer, src eventstream.EventStream) int {
+type RenderOptions struct {
+	ShowTimestamp bool
```
vito: 👍 I like the introduction of an options struct here rather than just a `bool`!

fly/eventstream/render.go
```diff
@@ -75,14 +81,17 @@ func Render(dst io.Writer, src eventstream.EventStream) int {
...
 			}
 
 			printColorFunc := printColor.SprintFunc()
-			fmt.Fprintf(dst, "%s\n", printColorFunc(e.Status))
+			eventLog = NewEventLog("%s\n", printColorFunc(e.Status), e.Time)
+			fmt.Fprint(dst, AdditionalFormatting(eventLog, options))
 
 			return exitStatus
 		}
+		fmt.Fprint(dst, AdditionalFormatting(eventLog, options))
```
vito: I think this would be less code if you wrapped `dst` with a single writer that kept track of the timestamp and conditionally added it at the start of each line. Sort of like the `PrefixedWriter`, but stateful and more domain-specific. Then all the `fmt.Printf`s don't have to change and you don't need the intermediary `EventLog` type.

Something like:

```go
dst := NewTimestampedWriter(dst, options.ShowTimestamps)

for {
  // ...
  switch e := ev.(type) {
  case event.Log:
    dst.SetTimestamp(e.Time)
    fmt.Fprint(dst, "%s", e.Payload)
  // ...
  }
}
```

I think the current approach works fine, just a suggestion. 🙂 I'm being especially careful with `fly` PRs because it's long overdue for some major refactoring in other areas. Hope that's OK!

fly/commands/watch.go
```diff
@@ -11,8 +11,9 @@ import (
 )
 
 type WatchCommand struct {
-	Job   flaghelpers.JobFlag `short:"j" long:"job"   value-name:"PIPELINE/JOB"   description:"Watches builds of the given job"`
-	Build string              `short:"b" long:"build"                               description:"Watches a specific build"`
+	Job   		flaghelpers.JobFlag `short:"j" long:"job"   	value-name:"PIPELINE/JOB"   description:"Watches builds of the given job"`
+	Build 		string              `short:"b" long:"build"                               	description:"Watches a specific build"`
+	Timestamp 	bool				`short:"t" long:"timestamp"								description:"Print with timestamp"`
```
vito: The short flag `-t` here will conflict with `-t` for `--target`. I'm having a hard time thinking of a better short flag though.

minor nitpick: `go fmt` on save?

pivotal-kahin-ng:
Thanks for the quick reply! It seems like I forgot to add an integration test 😅. I'll get that done right away!

`-t` actually doesn't conflict with the `--target` if users are following the usage example `fly [OPTIONS] watch [watch-OPTIONS]` but I can see how it could be confusing to have two `-t` in a command or if users like to put target their pipeline at the end of the command.

fly/commands/watch.go
pivotal-kahin-ng: How does `-u` sound for UTC or `-z` for timezone or `-i` for the second letter in the word?

pivotal-kahin-ng:
Seems like I have messed up my commit history.... let me fix that..

fly/eventstream/render.go
pivotal-kahin-ng: Updated it to be the way you visioned!

fly/commands/watch.go
```diff
@@ -11,8 +11,9 @@ import (
 )
 
 type WatchCommand struct {
-	Job   flaghelpers.JobFlag `short:"j" long:"job"   value-name:"PIPELINE/JOB"   description:"Watches builds of the given job"`
-	Build string              `short:"b" long:"build"                               description:"Watches a specific build"`
+	Job       flaghelpers.JobFlag `short:"j" long:"job"   	value-name:"PIPELINE/JOB"   description:"Watches builds of the given job"`
+	Build     string              `short:"b" long:"build"                               	description:"Watches a specific build"`
+	Timestamp bool                `short:"u" long:"timestamps"								description:"Print with timestamp"`
```
vito: This flag's a bit sneaky but that tends to happen with short flags. Fine with me since `-u` seems unlikely to collide with much at least.

fly/integration/watch_test.go
```diff
@@ -111,6 +111,31 @@ var _ = Describe("Watching", func() {
...
+						{ID: 2, Name: "2", Status: "started"},
+						{ID: 1, Name: "1", Status: "finished"},
+					}),
+				),
+				eventsHandler(),
+			)
+		})
+
+		It("watches the most recent one-off build with '--timestamp' configuration", func() {
+			watch("--timestamps")
```
vito: :thinking: These tests don't seem to verify anything timestamp-related, and it looks a little awkward to work timestamp tests in since `watch` just embeds its own assertions. If you're up for it, feel free to change how this is all done, but considering the meat of the functionality is already tested in the renderer, I'd be OK with merging this without the additional layer of tests.

fly/integration/watch_test.go
```diff
@@ -121,6 +146,10 @@ var _ = Describe("Watching", func() {
 		It("Watches the given build id", func() {
 			watch("--build", "3")
 		})
+
+		It("Watches the given build id with timestamp prefixed", func() {
```
vito: We can do without these additional tests (this + the rest in the file), since these flags don't really affect each other.

fly/ui/colors.go
```diff
@@ -6,7 +6,7 @@ var PendingColor = color.New(color.FgWhite)
 var StartedColor = color.New(color.FgYellow)
 var SucceededColor = color.New(color.FgGreen)
 var FailedColor = color.New(color.FgRed)
-var ErroredColor = color.New(color.FgWhite, color.BgRed, color.Bold)
+var ErroredColor = color.New(color.FgRed, color.Bold)
```
vito: Seems unrelated but I'm OK with it anyway.

pivotal-kahin-ng:
It would take me a bit of time to figure out how to rework the integration tests to be more flexible. Maybe it can be in a different PR?

My intent was for the time to show up as local time. I was under the impression that `func Unix(sec int64, nsec int64) Time` returned local time.

fly/commands/watch.go
pivotal-kahin-ng: Just noticed that the `build` command has `-t` as a short form for team 😶

vito: Heh, sure let's just go with `-t` then.

vito:
@pivotal-kahin-ng Looks like `time.Unix` does return local time. I think if we just change the flag to `-t` then (now that we're sure it's local) and clean up/remove the integration tests I can go ahead and merge this in.

Issue #869, opened 2017-01-12
New 'resources' command for fly


pcman312:
# Feature Request

## What challenge are you facing?

I am trying to pause all resources within concourse (yes, resources, not pipelines). In order to do that, I need a list of all of the resources currently configured so I can put them into the `pause-resource` command. Similar to the `pipelines` command, a `resources` command would be super useful to see a list of all resources (either globally or for a specific pipeline).

jama-pivotal:
Would also be nice to have `fly resource-versions` to support the pinning/unpinning features in #2702

mhuangpivotal:
`fly resources` command has been added in #2868.

It is no longer possible to pause a resource (even in the UI) because resource versions are now global (#2386). Instead, a resource can be pinned to a specific version (#2439, #2508).

PR #2868, opened 2018-11-26
fly: Add resources command to list all resources for given pipeline

pivotal-kahin-ng:
Implementation for this [issue](https://github.com/concourse/concourse/issues/869). 
Now sure how useful this will be since the `pause-resource` command is now removed? 

Shows resource name, paused/unpaused, and type of resource. Open to any suggestions on what else to show!


vito:
neat! looks good, thanks! :confetti_ball:

Issue #2499, opened 2018-08-16
Metrics for tracking number of workers by state


vito:
## What challenge are you facing?

I started seeing our per-worker container counts go up on Wings, and it took way too long to notice that it wasn't the number of containers that was increasing, it was the number of running workers that had gone down. A bunch of them were in `retiring` state, so they weren't being allocated any containers, resulting in the remaining workers being overloaded.

## A Modest Proposal

Having some way to notice workers in `retiring`, `landing`, `landed`, and `stalled` states (in addition to the happy state, `running`) would have made this way easier to notice.

Arguably we could have also noticed that the total container count hadn't spiked, which we can already do in our dashboards by summing up the existing containers metric (instead of grouping by worker). But being able to see workers stalling and un-stalling over time could also be a nice win long-term.

hprotzek:
Hi @vito is there any hidden complexity what I don't see? Why not just emit a new metric with the state string in `HeartbeatWorker` and then map the state to an `int` in the PrometheusEmitter. (didn't look at the other emitters). I am happy to help :-)

vito:
@hprotzek One complexity is that if a worker fails to heartbeat it'll transition to `stalled` state, which we'd also want metrics for (and that would inherently not come from `HeartbeatWorker`). I think this might just need to be its own periodic emitter.

cirocosta:
Oops, 

I just noticed that in https://github.com/concourse/concourse/commit/e7ee26a3219c7cfb281beba08cde9612a38d8e46 we missed the fact that `prometheus` has its own "switch" to decide which metrics to emit.

To quickly reproduce that:

1. enable Prometheus locally

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 1b7dc7b03..eaa90a482 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
     - .:/src
     command: web
     depends_on: [db]
-    ports: ["8080:8080"]
+    ports: ["8080:8080", "9391:9391"]
     environment:
     - CONCOURSE_POSTGRES_HOST=db
     - CONCOURSE_POSTGRES_USER=dev
@@ -28,6 +28,8 @@ services:
     - CONCOURSE_ADD_LOCAL_USER=test:test,guest:guest
     - CONCOURSE_MAIN_TEAM_LOCAL_USER=test
     - CONCOURSE_LOG_LEVEL=debug
[...]

Issue #2244, opened 2018-05-30
Prometheus metrics include dead workers


jmcarp:
# Bug Report

The prometheus metrics emitter never resets or deletes metrics for dead workers, so the prometheus endpoint reports container counts, volume counts, etc. for workers long after they're retired. We noticed this because one of our workers had a high container count right as it was retired, and we alert on high container counts, so we saw a phantom alert firing even though the worker it referenced didn't exist.

Not sure how to fix this with the current metrics interface, since the prometheus emitter doesn't know when workers get garbage collected.

cc @TimSimmons @databus23 

The following can also be handy:

* Concourse version: 3.13.0
* Deployment type (BOSH/Docker/binary): BOSH
* Infrastructure/IaaS: AWS
* Browser (if applicable): n/a
* Did this used to work? n/a

marco-m:
Given the very good explanation by @TimSimmons on the impedence mismatch between Concourse metrics architecture (designed to push) and Prometheus (who wants to pull), and given that I completely agree that additional code just add complexity and bugs, would it make sense to just require to use the Prometheus PushGateway ? This would 

1. simplify Concourse code dealing with Prometheus
2. resolve this bug

At work we collect all infra metrics with Prometheus and I would have no problems in adding a VM running the PushGateway only to collect sensible metrics for Concourse. If people complain they have to add a PushGateway, well, let them complain :-)

Opinions ?

vito:
Thanks for the info everyone!

I'm definitely seeing the friction here with Concourse's push style metrics vs. Prometheus's pull. I hope we can find a way in the future to not have so much custom (per-metric) code for the Prometheus backend. Right now it seems like every metric we add involves adding more code to the Prometheus engine to handle it, and I'm not at all confident that all engineers will remember to do that as we add them. I like to think of metrics as being in the same boat as logging, and want to be able to sprinkle them in cheaply.

Maybe we could have more information on the metrics types themselves so the Prometheus backend can generically process them? For example an `.EphemeralKey() string` marker which will result in the metric being reaped after a certain amount of inactivity from that same key? :man_shrugging:

vito:
Closing as https://github.com/concourse/atc/pull/277 was merged. Thanks!

marco-m:
Adding a comment to this closed bug because it seems to me that this thread contains good discussion about what Prometheus expects and the difficulty in making it work for Concourse.

[...]

PR #2869, opened 2018-11-26
atc: Adds missing worker state metric to prometheus emitter

cirocosta:
This PR adds a missing metric that was added before (worker state).

In the case of `prometheus`, it gets named as `concourse_worker_info`,
where `state` is just a label that we can use to aggregate on while
still having per-worker information (which can be further increased
later).

concourse/concourse#2499

With the feature in:

```sh
curl --silent localhost:9391/metrics | ag concourse_workers_info
# HELP concourse_workers_info Per-worker information
# TYPE concourse_workers_info gauge
concourse_workers_info{info="running",worker="ffbbd436b4a5"} 1
```

This way we're able to perform:

*ps.: tests were previously added in https://github.com/concourse/concourse/commit/ff2e45ff73cdf6cf3a9c3158420b1a00dd47a7a6, the issue where the metric (under `metrics.go`) was created.*

cc @YoussB

### DIFF
--- a/atc/metric/emitter/prometheus.go
+++ b/atc/metric/emitter/prometheus.go
@@ -27,6 +27,7 @@ type PrometheusEmitter struct {
 
 	workerContainers *prometheus.GaugeVec
 	workerVolumes    *prometheus.GaugeVec
+	workerInfo       *prometheus.GaugeVec
 
 	httpRequestsDuration *prometheus.HistogramVec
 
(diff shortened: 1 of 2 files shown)


vito:
thanks!

cirocosta:
Whoops, I noticed that this metric faces the same problem as reported in here: https://github.com/concourse/concourse/issues/2244

The gist of it:

>  With the default instrumentation approach (defining prometheus.{Counter|Gauge|Histogram}Vec objects and calling .Inc() or whatever on them), when you export a metric with a set of labels, it creates a timeseries that will be exported in the /metrics endpoint until the process is restarted (or, more specifically, the Prometheus Registerer object is re-instantiated or something, which you don't want to do). https://github.com/concourse/concourse/issues/2244#issuecomment-393187916

I'll follow up with a fix soon.

Sorry!

PR #2871, opened 2018-11-27
speed up dev builds

ralekseenkov:
download and cache go mod dependencies first in a docker layer, then build
go mod dependencies won't be downloaded the next time you rebuild

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,9 +19,16 @@ ENV CONCOURSE_SESSION_SIGNING_KEY /concourse-keys/session_signing_key
 ENV CONCOURSE_TSA_PUBLIC_KEY         /concourse-keys/tsa_host_key.pub
 ENV CONCOURSE_TSA_WORKER_PRIVATE_KEY /concourse-keys/worker_key
 
-# build Concourse
-COPY . /src
+RUN mkdir /src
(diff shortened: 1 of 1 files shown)


cirocosta:
Hey @ralekseenkov ,

Thanks for the contribution, but I this this has the very same effect of what is already done by `concourse/dev` (which is used by the root `Dockerfile`); is that right?

https://github.com/concourse/concourse/blob/71ee78fecc27d3d525d14ee4082b70cb04d5a21f/ci/dockerfiles/dev/Dockerfile#L42-L47

thx!

ralekseenkov:
@cirocosta my use case is testing Concourse changes locally using `docker-compose up --build`

the current behavior is that go mod dependencies are downloaded every time you run `docker-compose up --build`

with this fix, they are being cached in the docker layer, because it processes go.sum and go.mod first before the entire code base is copied under /src. so if go.sum and go.mod don't change, then go mod dependencies won't be re-downloaded. this fix seems pretty standard in Go 1.11 + Docker world

vito:
Ah, yeah same goal but this is kinda nice since it'll also work without requiring the occasional pull of `concourse/dev` in order to bring upstream module changes up to speed. Let's try it out!

Issue #2766, opened 2018-11-06
Workers endpoint returns 0 for `active_volumes`


topherbullock:
The `/api/v1/workers` endpoint is currently returning 0 for `active_volumes` on every worker returned. Since this API is used by `fly workers` that command also lists workers as having 0 active volumes.

pivotal-kahin-ng:
Currently when using `fly workers`, `active_volumes` will only show up if the user uses the `--json` configuration. Would it be beneficial for `volumes` to be one of the columns in the default `fly workers` table output?

cirocosta:
@topherbullock , 

I'd say this can be already accepted (it's still in Icebox, but I guess we can move it), wdyt?

thx!

topherbullock:
Closed as this is merged and accepted. thx @cirocosta :D

PR #2879, opened 2018-11-28
Fix workers endpoint to show active_volumes

pivotal-kahin-ng:
Fix for [#2766](https://github.com/concourse/concourse/issues/2766).

Add active volumes to heartbeat.


pivotal-kahin-ng:
Exactly that. Wasn't saved either!

vito:
Thanks!

Issue #2230, opened 2018-05-21
Shared webhook endpoint for multiple resources


wendorf:
# Feature Request

## What challenge are you facing?

We have one single GitHub repository that includes many different conceptual resources. For example, _n_ directories, one for each application we release. We use `source.paths` to generate versions for the resources only when the folder representing the contents of the resource has changed. This results in _n_ checks every `check_interval`, which puts a lot of strain on our workers.

We want to use `webhook_token`s to reduce strain (and setting `check_interval` to a very high number), but each resource has its own webhook URL, which means we must create a GitHub webhook for each resource. Not only is this an operational burden, but GitHub has a limit of 20 webhooks for a single repository (see [docs](https://developer.github.com/webhooks/)), preventing us from using this solution as our number of resources grows.

## A Modest Proposal

I would like to be able to define a single webhook URL that can trigger all _n_ resources. This would still result in _n_ checks on every push, but would greatly reduce the total number of checks that are run. Some YAML like this would be good:

```
resources:
- name: app-one
  type: git
  source:
    uri: git@github.com:org/repo.git
    branch: master
    paths: [app_one]
  webhook:
    name: MY_WEBHOOK
    token: WEBHOOK_TOKEN

- name: app-two
  type: git
  source:
    uri: git@github.com:org/repo.git
    branch: master
    paths: [app_two]
  webhook:
    name: MY_WEBHOOK
    token: WEBHOOK_TOKEN
```

Which would allow me to configure GitHub with a webhook path of `https://my-concourse.example.com/api/v1/teams/TEAM_NAME/pipelines/PIPELINE_NAME/webhooks/MY_WEBHOOK?webhook_token=WEBHOOK_TOKEN`.

An alternate structure:

```
webhooks:
- name: MY_WEBHOOK
  token: WEBHOOK_TOKEN

resources:
- name: app-one
  type: git
  source:
    uri: git@github.com:org/repo.git
    branch: master
    paths: [app_one]
  webhooks: [MY_WEBHOOK]

- name: app-two
  type: git
  source:
    uri: git@github.com:org/repo.git
    branch: master
    paths: [app_two]
  webhooks: [MY_WEBHOOK]
```

billimek:
👍 for this as well.  We end up creating a ton of webhooks on the github side and would love to have just one trigger a check for everything.

a459259:
+1 - We have a similar use-case, would make a lot of sense to have this in place...

haganbmj:
[...]

PR #2886, opened 2018-11-29
WIP: Shared Webhooks

christophermancini:
The scope of this PR is to improve the user experience and performance w.r.t. webhooks. Currently, webhooks maintain a one-to-one relationship with resources. Creating the need to have independent webhooks in code repository hosting systems such as GitHub for each branch and separately for PRs. If you use monorepos like we do at [DigitalOcean](https://blog.digitalocean.com/cthulhu-organizing-go-code-in-a-scalable-repo/), it can pose threats to performance for both Concourse and GitHub Enterprise at scale.

Inspired by the Shared Webhook idea proposed in #2230, I:

- added new `atc.Config` elements to support shared webhooks
- added a new API route of `/api/v1/resources/check/webhook`
- created a function, `CheckSharedWebHook`, in `check_webhook.go` to handle the new route
- left `webhook_token` in tact to maintain backwards compatibility with non-shared webhooks


vito:
Hmm it's really tempting to wait and see where https://github.com/concourse/rfcs/pull/15 goes. This, along with the async webhook checking change, makes me really worried about an unmanaged explosion in asynchronous (possibly redundant) resource checks.

/cc @pivotal-jwinters 

Side note: in light of https://github.com/concourse/concourse/issues/2895 I'm also a bit concerned about our scheduling approach, which that RFC mentions... I think after 5.x is out it'll be a good time to rethink these things (job/build and resource check scheduling) from first principles.

christophermancini:
Closing the PR for now per @vito's concerns.

Issue #1112, opened 2017-05-09
Jobs listing endpoint returns null for pipeline without jobs


chendrix:
Hey,

looks like `GET /api/v1/teams/:team_name/pipelines/:pipeline_name/jobs` returns `null` (text string) if a pipeline have 0 jobs. I suppose it should return an empty list with the content type properly set to `content-type/application-json`. What do you think?

---

``` http
GET /api/v1/pipelines/pipeline/jobs HTTP/1.1
Authorization: ***** Hidden credentials *****
(...)

HTTP/1.1 200 OK
Set-Cookie: ATC-Authorization=Basic bXl1c2VyOm15cGFzcw==; Path=/; Expires=Fri, 05 Aug 2016 14:22:30 GMT
X-Concourse-Version: 1.5.1
Date: Thu, 04 Aug 2016 14:22:30 GMT
Content-Length: 5
Content-Type: text/plain; charset=utf-8
Connection: close

null
```

having

```
+ fly -t <target> set-pipeline -c ./pipeline-3.yml -p pipeline -n
targeting http://<addr>

resources:
  resource repository has been added:
    name: repository
    type: git
    source:
      uri: https://github.com/cirocosta/test-project.git

pipeline created!
```

thx!

---
Moved from concourse/atc#97
/cc @cirocosta

vito:
fixed by #2909

PR #2909, opened 2018-12-04
atc: return non-zero array instead of null for pipelines with no jobs

pivotal-kahin-ng:
Fix for this issue: [1112](https://github.com/concourse/concourse/issues/1112). 

**FYI**: Original issue references `pipeline.yml` to only have a resource declared. Seems like that isn't applicable anymore since the user will get an error `"resource 'resourceName' is not used"`. Reproduce this issue by having only a `resource_types` in the `pipeline.yml`.


vito:
thanks!

Issue #2400, opened 2018-07-19
`fly get-pipeline` should show an error message if the pipeline is not found


ddadlani:
# Bug Report

Running `fly get-pipeline` when the pipeline does not exist, or when logged into the wrong team, returns empty values instead of an error message.

If I am logged in to team 'A' and I try to run `fly get-pipeline` for team 'B', I get the following output:
```
$  fly -t ci get-pipeline -p lala
groups: []
resources: []
resource_types: []
jobs: []
```

Instead, I would expect to see an error message telling me that I am unauthorized/pipeline does not exist.

Concourse/fly version: 3.14.1

vito:
fixed by https://github.com/concourse/concourse/pull/2933

PR #2933, opened 2018-12-11
fly: add check to see if pipeline exists

pivotal-kahin-ng:
Fix for [2400](https://github.com/concourse/concourse/issues/2400). 

Uses existing implementation from `PipelineConfig` which checks if pipeline is not found from the `GetConfig`.


vito:
thanks!

PR #2964, opened 2018-12-22
Fix based on lint except missing comment in skymarshal

cappyzawa:
There were lots of lint errors.
I fixed skymarshal's errors at the beginning (except missing comment...).

Before:
```
$ cd skymarshal
$ golint ./... | grep -v "comment" | wc -l
      76
```

After:
```
$ cd skymarshal
$ golint ./... | grep -v "comment" | wc -l
       0
```


cappyzawa:
hmmm..., the testflight is succeeded in my local env.

PR #2995, opened 2019-01-05
fix fly execute command

maplain:
go routine abortOnSignal is only executed once.

early exit on first signal will make following code never be executed, and the program will be stuck at that point.

### DIFF
--- a/fly/commands/execute.go
+++ b/fly/commands/execute.go
@@ -203,7 +203,6 @@ func abortOnSignal(
 	err := client.AbortBuild(strconv.Itoa(build.ID))
 	if err != nil {
 		fmt.Fprintln(ui.Stderr, "failed to abort:", err)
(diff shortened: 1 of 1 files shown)


vito:
Thanks for submitting this, but can you provide more information about what this fixes? How does it get stuck after an `os.Exit()` (which exits the program)?

maplain:
oh my bad, my fly is on 4.1.0. and it looks like it's already fixed.

Previously, it's a `return` here: https://github.com/concourse/fly/blob/v4.1.0-rc.20/commands/execute.go#L185 instead of an os.exit

this is the fix: https://github.com/concourse/concourse/commit/1e579cc62e977ea279d93528a45d4f78e593604d

closing this one.

vito:
Ah makes sense, np!

Issue #2988, opened 2019-01-03
Make periodic check container rebalancing duration configurable


vito:
see https://github.com/concourse/concourse/issues/2923

## What challenge are you facing?

Concourse rebalances check containers periodically to make sure they don't all end up one worker after a deploy. The duration that they rebalance is determined by worker uptime (so we rebalance more aggressively if there are newly provisioned workers), with an upper bounds of 1 hour. This upper bound is hardcoded and can result in painful re-cloning of large repos every hour.

## What would make this better?

Make it configurable? I could see two schemes here:

* After some value X (default 1 hour), back off on the rebalancing, only rebalancing after some value Y (configurable, default 1 hour to keep today's behavior).
* Simply add one value X which is the cap for the rebalancing, the same way the 1 hour configuration works today.

Not sure if one is better or worse, just that if you want to slow down the rebalancing more aggressively you may prefer the first option (otherwise things will keep rebalancing, just slower and slower until it hits your limit, say 24 hours).

PR #3030, opened 2019-01-11
Add web config for max duration of check container life cycle

pivotal-kahin-ng:
Add functionality for issue #[2988](https://github.com/concourse/concourse/issues/2988).


vito:
Hey, I think we're actually gonna do #3079 soon (which removes `check` container re-use entirely) so I think we can close this. Thanks though!

Issue #2957, opened 2018-12-20
Expiring the Auth Cookie shouild set 'HttpOnly' and 'Secure' fields


topherbullock:
In the logout flow we set the users cookie with an expiry in the past, but it does not include the `HttpOnly` and `Secure` attributes when setting that expiry.

Despite this being the logout flow (`/sky/logout`) and not exposing anything, we should still set those attributes as a best practice for all cookies.

vito:
fixed via https://github.com/concourse/concourse/pull/3031

PR #3031, opened 2019-01-11
Add httponly and secure attributes to skyserver calls

pivotal-kahin-ng:
Fix for issue #[2957](https://github.com/concourse/concourse/issues/2957)


vito:
For context, I think this was caught by an automated security scanner and that's what triggered the initial issue to be opened.

vito:
thanks! i'll go ahead and merge this

topherbullock:
@pivotal-jwinters @vito yeah, this isn't necessary, but worth it to ensure that future security scans don't flag it 😄

Issue #3036, opened 2019-01-14
Include team and tag in worker metrics


evanchaoli:
## What challenge are you facing?

In our cluster, we created a special worker pool with extra CPU and RAM in order to run heavy pipeline task, like "mvn build". All workers in this special pool are tagged with 'B747'.

We emit metrics to InfluxDB and use Graphana to monitor metrics. For worker metrics, I want to separate normal pool and the special pool into different charts. But according to the code https://github.com/concourse/concourse/blob/master/atc/metric/metrics.go#L126-L129, Concourse doesn't include `tag` in worker metrics, so that we are not able to split worker metrics based on tag.

One of my team member is going to submit a pull request, that will include both `team` and `tag` in worker metrics. I'm creating this issue to track the change.

## Are you interested in implementing this yourself?

Yes

cirocosta:
Hey @evanchaoli ,

Thanks for creating this issue!

I'd like to just add some two cents on the idea of having instead a separate metric that would output these extra informations.

<br />
From a Prometheus perspective, it seems fairly common to have a metric that exposes info about a particular instance/thing when you need to cross reference that later.

For instance, in the case of [`kubestate`](https://github.com/kubernetes/kube-state-metrics), there's [`kube_pod_info`](https://github.com/kubernetes/kube-state-metrics/blob/fd65206d7dbba19bbf78c6552375932447b80fc1/pkg/collectors/pod.go#L41-L66):

```
kube_pod_info{
	...
	namespace="kube-system",
	node="gke-cluster-1-preemptible-worker-pool-e3099e3f-6rd1",
	pod="kube-proxy-gke-cluster-1-preemptible-worker-pool-e3099e3f-6rd1",
	pod_ip="10.128.0.19",
	release="hush-house",
}	
```

Prometheus itself has `prometheus_build_info`:

```
prometheus_build_info{
	branch="HEAD",
	goversion="go1.11.3",
	instance="localhost:9090",
	job="prometheus",
	revision="dbd1d58c894775c0788470944b818cc724f550fb",
	version="2.6.0",
}	
```

That way, you can expose these extra information without having to add it to every metric, while still being possible to reference such when needed.

What I'm not sure, is if this is a common pattern in InfluxDB as well. Are you aware if that's the case? If so, WDYT about going through such direction?

Thx!

Any thoughts @marco-m ?

marco-m:
Summoning @databus23 for Prometheus-fu

evanchaoli:
[...]

PR #3037, opened 2019-01-14
Metric enhance: add team, tags for worker-volumes/worker-containers.

SimonXming:
For #3036:
* Add `team`, `tags` for worker-volumes
* Add `team`, `tags` for worker-containers

### DIFF
--- a/atc/metric/emitter/prometheus.go
+++ b/atc/metric/emitter/prometheus.go
@@ -448,9 +448,9 @@ func (emitter *PrometheusEmitter) workerContainersMetric(logger lager.Logger, ev
 		logger.Error("failed-to-find-platform-in-event", fmt.Errorf("expected platform to exist in event.Attributes"))
 		return
 	}
-	team, exists := event.Attributes["team"]
-	if !exists || team == "" {
-		logger.Error("failed-to-find-team-in-event", fmt.Errorf("expected team to exist in event.Attributes"))
+	team, exists := event.Attributes["team_name"]
+	if !exists {
+		logger.Error("failed-to-find-team-name-in-event", fmt.Errorf("expected team_name to exist in event.Attributes"))
 		return
 	}
 	tags, _ := event.Attributes["tags"]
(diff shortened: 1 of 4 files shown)


vito:
@cirocosta Wdyt of this in the context of Promethus and other cardinality-sensitive metrics backends?

cirocosta:
Hi @SimonXming ! Sorry for the very long delay, I'll be looking at it shortly!

SimonXming:
> Hi @SimonXming ! Sorry for the very long delay, I'll be looking at it shortly!

Thanks😄This feature is very useful for us to monitoring worker metrics.

atc/metric/metrics.go
```diff
@@ -126,6 +129,8 @@ func (event WorkerContainers) Emit(logger lager.Logger) {
 			Attributes: map[string]string{
 				"worker":   event.WorkerName,
 				"platform": event.Platform,
+				"team":     event.Team,
+				"tags":     strings.Join(event.Tags[:], " "),
```
cirocosta: Oh, I didn't know prometheus' text format allowed this (interestingly enough, it allows any utf8! cool! (see [exposition format](https://github.com/prometheus/docs/blob/5b4813eeb364d708278d54ee82cda15d2832b91f/content/docs/instrumenting/exposition_formats.md#comments-help-text-and-type-information))).

I wonder if this would be a problem with the other emitters though 🤔  Do you know if that'd be a problem for the other emitters supported?

atc/metric/emitter/prometheus.go
```diff
@@ -425,14 +425,20 @@ func (emitter *PrometheusEmitter) workerContainersMetric(logger lager.Logger, ev
 		logger.Error("failed-to-find-platform-in-event", fmt.Errorf("expected platform to exist in event.Attributes"))
 		return
 	}
+	team, exists := event.Attributes["team"]
+	if !exists || team == "" {
```
cirocosta: This will be a problem for those workers that don't have a team assigned to them, as in their registration/heart-beating, they'll reply with an empty string for `team`, thus making those who don't have per-team workers have any way of checking this metric.

Would it make more sense to do something like cAdvisor does with empty labels? https://github.com/google/cadvisor/issues/1951

That way we would have consistency in the label set, although I'm not entirely sure if this is the best & recommended way to go.

Wdyt?

atc/metric/metrics.go
SimonXming: Thanks for the heads-up. I do a little investigate here.
For `dogstatsd`:
- <https://docs.datadoghq.com/tagging/#defining-tags>
- It support slashes, likes `tags:tag_1/tag_2`

For `InfluxDB`
- <https://docs.influxdata.com/influxdb/v1.7/write_protocols/line_protocol_reference/#special-characters>
- For "tag values", slashes don't need to escapes by a backslash

For `lager` and `newRelic`
- I run a little test, using slashes also works fine

For `riemann`, seems it's using `github.com/golang/protobuf` for serialization.
- In my test, using slashes works fine

So I decide using slashes as tags separator. Wdyt?

atc/metric/emitter/prometheus.go
SimonXming: Thanks, I think you are right. I will ensure teamless worker also emit this metric.

marco-m:
hello @cirocosta kind reminder, this is another PR ready for your review, thanks :-)

cappyzawa:
I'm looking forward to this feature 😍

Issue #2980, opened 2019-01-02
Prometheus Locks state metric has unbound cardinality


cirocosta:
## Bug Report

Right now, `concourse_locks_state`(introduced in https://github.com/concourse/concourse/pull/2681) has a label that can get wild in terms of cardinality given that it exposes `objectId` that seems to be almost unbounded:

<details>
<summary>sample</summary>
<pre>
concourse_locks_state{lockType="Batch",objectId="-629780414"} 0
concourse_locks_state{lockType="Batch",objectId="-630800735"} 0
concourse_locks_state{lockType="Batch",objectId="-631227151"} 0
concourse_locks_state{lockType="Batch",objectId="-63332322"} 0
concourse_locks_state{lockType="Batch",objectId="-65043860"} 0
concourse_locks_state{lockType="Batch",objectId="-653188958"} 0
concourse_locks_state{lockType="Batch",objectId="-654479246"} 0
concourse_locks_state{lockType="Batch",objectId="-655227815"} 0
concourse_locks_state{lockType="Batch",objectId="-656004949"} 0
concourse_locks_state{lockType="Batch",objectId="-656531228"} 0
concourse_locks_state{lockType="Batch",objectId="-665949219"} 0
concourse_locks_state{lockType="Batch",objectId="-668941765"} 0
concourse_locks_state{lockType="Batch",objectId="-672275628"} 0
concourse_locks_state{lockType="Batch",objectId="-673747273"} 0
concourse_locks_state{lockType="Batch",objectId="-674451151"} 0
concourse_locks_state{lockType="Batch",objectId="-676382477"} 0
concourse_locks_state{lockType="Batch",objectId="-677811811"} 0
concourse_locks_state{lockType="Batch",objectId="-678350638"} 0
concourse_locks_state{lockType="Batch",objectId="-680122896"} 0
concourse_locks_state{lockType="Batch",objectId="-68260188"} 0
concourse_locks_state{lockType="Batch",objectId="-684084400"} 0
concourse_locks_state{lockType="Batch",objectId="-684971574"} 0
concourse_locks_state{lockType="Batch",objectId="-692834064"} 0
concourse_locks_state{lockType="Batch",objectId="-697969430"} 0
concourse_locks_state{lockType="Batch",objectId="-711165751"} 0
concourse_locks_state{lockType="Batch",objectId="-71154364"} 0
concourse_locks_state{lockType="Batch",objectId="-712431973"} 0
[...]

PR #3041, opened 2019-01-14
atc: remove objectId from locks held metric [#2980]

cirocosta:
tl;dr: this PR removes `objectId` from `locksheld` metric, but I *think* this metric *might* be more useful if we change it to keep track of lock acquiring & releasing durations.

---

Hey,

In this PR I try to address the problems that the *very* high cardinality of the [`LocksHeld`](https://github.com/concourse/concourse/pull/2681) might present by completely removing `objectId` from the final metric (which would get rid of that from all emitters - manually done for Prometheus).

```sh
curl -s localhost:1337/metrics | grep locks
# HELP concourse_locks_state Database locks held
# TYPE concourse_locks_state gauge
concourse_locks_held{type="Batch"} 0
concourse_locks_held{type="DatabaseMigration"} 0
```

While I see why one would like to have the `ObjectID` included (we could be able to see when the lock for "CreatingContainer" where `containerID=1` was acquired and when released), I'm not very sure whether having the metric without such dimension (`objectid`) is still all that relevant.

*ps.: If the intention of the metric was to answer the question of "are we having locks not being released for too long?", maybe a different type of metric could be used - one like `Histogram` where we could keep track of the duration from `acquire` to `release` for each lock type (keeping track of the individual ObjectID locks in-memory only).*

Wdyt?

thx!

cc @topherbullock


cirocosta:
Thanks!

Just added a new panel to hush-house that consumes it in a per-ATC manner:

screen shot 2019-01-17 at 8 12 15 am

Issue #3000, opened 2019-01-07
Bring back `concourse-rc` image to the pipeline


cirocosta:
## What challenge are you facing?

Previously, the pipeline would produce a release-candidate image that contained the final versions of `fly` (with the binaries of the 3 supported platforms), the binary, and base resources.

After the introduction of the new pipeline (https://ci.concourse-ci.org/teams/main/pipelines/concourse), that job went missing.

This would allow `hush-house` make use of an actual release-candidate version, having `fly` packed with it (so `fly sync` would work just fine), and not miss the default golang compiler optimizations (stripped out in `concourse/dev` for `dlv` debugging purposes).

## What would make this better?

Have the image produced in the pipeline.

## Are you interested in implementing this yourself?

Yep 😁

cirocosta:
Hey, assigning it to me to work on it today (freedom Friday).

cirocosta:
https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/rc-image/builds/7

PR #3042, opened 2019-01-14
ci: bring back `concourse-rc` image to the pipeline [#3000]

cirocosta:
Hey, 

This PR is intended to address the need for having an RC image being built in the pipeline (#3000) so that k8s-smoke, k8s-topgun, and hush-house can use that instead of `dev`.

#### Testing

To test it out, I used `fly execute` with the local changes, which seem to indicate that everything is working:

```sh
# Execute the `rc-image` task consuming some inputs from
# the bin-smoke job (which has similar inputs as `rc-image`).
fly -t ci \
  execute \
  -c ./ci/tasks/rc-image.yml \
  --input=concourse=./ \
  --inputs-from=concourse/bin-smoke \
  -p
```

ps.: when testing I included an extra `img login && img push cirocosta/concourse` so I could manually test the final artifact (the image) in k8s.

#### Image tag

if I understood correctly, [`registry-image`](https://github.com/concourse/registry-image-resource) seems to not support dynamic tag versioning right now (see https://github.com/concourse/registry-image-resource/issues/7  for some discussions around support for tagging built images), so the pipeline is default'ing to have the image being pushed to `concourse/concourse-rc:latest`.

#### How it looks

If you're wondering how the pipeline looks after the changes, check out below.

1. With develop & k8s & bosh:

first

2. With develop & k8s:

second

Please let me know if I missed anything,

Thx!

### DIFF
--- a/ci/dockerfiles/rc/Dockerfile
+++ b/ci/dockerfiles/rc/Dockerfile
@@ -1,0 +1,24 @@
+FROM ubuntu:bionic
+
+
+ENV \
+  CONCOURSE_RESOURCE_TYPES=/concourse-resource-types \
+  CONCOURSE_WORK_DIR=/worker-state \
+  CONCOURSE_GARDEN_DNS_PROXY_ENABLE=true
+
+
+RUN apt update && apt install -y \
+    ca-certificates \
+    btrfs-tools \
+    dumb-init \
+    iproute2
+
+
+ADD ./rc/bin/concourse /usr/local/bin
+ADD ./rc/bin/gdn /usr/local/bin
+ADD ./rc/resource-types /concourse-resource-types
+ADD ./rc/fly-assets /
+
+
+ENTRYPOINT ["dumb-init", "/usr/local/bin/concourse"]
+
--- a/ci/dockerfiles/rc/Dockerfile
+++ b/ci/dockerfiles/rc/Dockerfile
@@ -24,9 +24,6 @@ ENV CONCOURSE_WORKER_GARDEN_DNS_PROXY_ENABLE  true
 ENV CONCOURSE_WORK_DIR                /worker-state
 ENV CONCOURSE_WORKER_WORK_DIR         /worker-state
 
-# location for base resource types
-ENV CONCOURSE_RESOURCE_TYPES          /usr/local/concourse/resource-types
-
 # volume for non-aufs/etc. mount for baggageclaim's driver
 VOLUME /worker-state
 
--- a/ci/dockerfiles/rc/Dockerfile
+++ b/ci/dockerfiles/rc/Dockerfile
@@ -1,24 +1,42 @@
-FROM ubuntu:bionic
+FROM ubuntu:bionic AS ubuntu
+
+FROM ubuntu AS assets
+COPY ./rc/*.tgz /tmp
+RUN tar xzf /tmp/*tgz -C /usr/local
 
 
-ENV \
-  CONCOURSE_RESOURCE_TYPES=/concourse-resource-types \
-  CONCOURSE_WORK_DIR=/worker-state \
-  CONCOURSE_GARDEN_DNS_PROXY_ENABLE=true
+FROM ubuntu
 
+# 'web' keys
+ENV CONCOURSE_SESSION_SIGNING_KEY     /concourse-keys/session_signing_key
+ENV CONCOURSE_TSA_AUTHORIZED_KEYS     /concourse-keys/authorized_worker_keys
+ENV CONCOURSE_TSA_HOST_KEY            /concourse-keys/tsa_host_key
+
+# 'worker' keys
+ENV CONCOURSE_TSA_PUBLIC_KEY          /concourse-keys/tsa_host_key.pub
+ENV CONCOURSE_TSA_WORKER_PRIVATE_KEY  /concourse-keys/worker_key
+
+# enable DNS proxy to support Docker's 127.x.x.x DNS server
+ENV CONCOURSE_GARDEN_DNS_PROXY_ENABLE         true
+ENV CONCOURSE_WORKER_GARDEN_DNS_PROXY_ENABLE  true
+
+# auto-wire work dir for 'worker' and 'quickstart'
+ENV CONCOURSE_WORK_DIR                /worker-state
+ENV CONCOURSE_WORKER_WORK_DIR         /worker-state
+
+# location for base resource types
+ENV CONCOURSE_RESOURCE_TYPES          /usr/local/concourse/resource-types
+
+# volume for non-aufs/etc. mount for baggageclaim's driver
+VOLUME /worker-state
 
 RUN apt update && apt install -y \
-    ca-certificates \
     btrfs-tools \
+    ca-certificates \
     dumb-init \
     iproute2
 
-
-ADD ./rc/bin/concourse /usr/local/bin
-ADD ./rc/bin/gdn /usr/local/bin
-ADD ./rc/resource-types /concourse-resource-types
-ADD ./rc/fly-assets /
+COPY --from=assets /usr/local/concourse /usr/local/concourse
 
 
-ENTRYPOINT ["dumb-init", "/usr/local/bin/concourse"]
-
+ENTRYPOINT ["dumb-init", "/usr/local/concourse/bin/concourse"]
--- a/ci/tasks/rc-image.yml
+++ b/ci/tasks/rc-image.yml
@@ -1,0 +1,27 @@
+---
+platform: linux
+
+image_resource:
+  type: registry-image
+  source: {repository: concourse/builder}
+
+inputs:
+- name: concourse
+- name: linux-rc
+  path: rc
+
+outputs:
+- name: image
+
+params:
+  DOCKERFILE: concourse/ci/dockerfiles/rc/Dockerfile
+  REPOSITORY: concourse/concourse-rc
+  CONTEXT: .
+
+run:
+  path: /bin/bash
+  args:
+    - -ce
+    - |
+      tar xzf ./rc/concourse-*.tgz --strip-components=1 --directory=rc
+      build
--- a/ci/tasks/rc-image.yml
+++ b/ci/tasks/rc-image.yml
@@ -19,9 +19,4 @@ params:
   CONTEXT: .
 
 run:
-  path: /bin/bash
-  args:
-    - -ce
-    - |
-      tar xzf ./rc/concourse-*.tgz --strip-components=1 --directory=rc
-      build
+  path: build
--- a/ci/pipelines/concourse.yml
+++ b/ci/pipelines/concourse.yml
@@ -37,6 +37,7 @@ groups:
   - unit
   - lint
   - dev-image
+  - rc-image
   - testflight
   - watsjs
   - rc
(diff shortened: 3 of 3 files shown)


ci/dockerfiles/rc/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM ubuntu:bionic
```
vito: Seems like this replaces the [`concourse-docker` Dockerfile](https://github.com/concourse/concourse-docker/blob/master/Dockerfile) which was used in the previous pipeline. Maybe we should update that instead? The separate repo is kind of nice for categorizing Docker-specific issues.

Also I think there'll need to be some equivalent to how the `concourse-docker` Dockerfile wires up the worker/TSA/session signing keys.

ci/dockerfiles/rc/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+
+RUN apt update && apt install -y \
+    ca-certificates \
+    btrfs-tools \
+    dumb-init \
+    iproute2
+
+
+ADD ./rc/bin/concourse /usr/local/bin
```
vito: I think you can just do this:

```
ADD ./rc/concourse-*.tgz /usr/local
```

...which would remove the need to extract it in the task (`ADD` extracts local archives). This should result in `/usr/local/concourse/{bin,fly-assets,...}`.

The advantage of this is that it'll match the install instructions for the tarball distribution, and when installed in this way the binary can automatically find the `fly-assets`, `resource-types`, `bin/gdn` etc.

This is how the smoke tests install, too: 

https://github.com/concourse/concourse/blob/5dd0bcc92d24858af0830c27ad3c3d652513f8bb/ci/deployments/smoke/smoke.tf#L143

This install method is partly inspired by Go (which typically installs to `/usr/local/go`).

ci/dockerfiles/rc/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ENV \
+  CONCOURSE_RESOURCE_TYPES=/concourse-resource-types \
+  CONCOURSE_WORK_DIR=/worker-state \
+  CONCOURSE_GARDEN_DNS_PROXY_ENABLE=true
+
+
+RUN apt update && apt install -y \
+    ca-certificates \
+    btrfs-tools \
+    dumb-init \
```
vito: Neat, didn't know this was available in `apt` repos!

ci/tasks/rc-image.yml
```diff
@@ -0,0 +1,27 @@
...
+  DOCKERFILE: concourse/ci/dockerfiles/rc/Dockerfile
+  REPOSITORY: concourse/concourse-rc
+  CONTEXT: .
+
+run:
+  path: /bin/bash
+  args:
+    - -ce
+    - |
+      tar xzf ./rc/concourse-*.tgz --strip-components=1 --directory=rc
```
vito: this can be removed if you extract with `ADD` as mentioned above

ci/dockerfiles/rc/Dockerfile
cirocosta: Oh, true! I totally missed it, thanks for pointing that out, I think it indeed makes sense to keep it there! 

Going that way (not keeping it under `concourse/concourse`'s `ci` directory), do you think it'd be nice to have a way for people to build it outside our pipeline? E.g., by just `docker build -t blabla .` there in `concourse-docker`? 

As I see, that'd mean:

1. taking an `rc` URL as an arg (maybe w/ a default to the last stable release?);
2. passing a build arg with the RC when building in our pipeline (that'd mean having `--buld-arg` available in `concourse/builder`).

Wdyt?

ci/dockerfiles/rc/Dockerfile
cirocosta: Neat! Thanks!

cirocosta:
Thanks for looking into it!

From the questions you added, by the end I started thinking that maybe we should indeed keep the repository there for a reference of how one can use `docker-compose` to run Concourse, but maybe the actual process of building the image should live under `ci` here in `concourse/concourse`.

The rationale would be that:

- how you "run" it would fit different repositories (like, `concourse-bosh-deployment`, `helm/charts`, and `concourse/concourse-docker`); and
- how to "build" it would fit `concourse/concourse`, where the actual source code and pipelines live.

Do you think that makes sense?

I hope my replies are not too confusing!

Thanks!

ci/dockerfiles/rc/Dockerfile
vito: Hmmm I don't think it's super important to be able to `docker build` from the `concourse-docker` repo, personally. For local development they could always build from the `concourse` repo.

ci/dockerfiles/rc/Dockerfile
cirocosta: Weird, I was trying the `ADD` with wildcards here, and it seems like it doesn't extract - if I try it without wildcards though (passing the full name), it works just fine.

Looking at the buildkit test suite for the Dockerfile frontend (see https://github.com/tonistiigi/buildkit/blob/6cb4256e81dd20691d9c71a988d0791aee02e00b/frontend/dockerfile/dockerfile_test.go#L38-L49), it appears that they indeed don't have a test for such case. 

<details>
<summary>with wildcard</summary>
<pre>
docker build -t a .
[+] Building 2.9s (9/9) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                   0.0s
 => => transferring dockerfile: 1.09kB                                                                                                                 0.0s
 => [internal] load .dockerignore                                                                                                                      0.0s
 => => transferring context: 2B                                                                                                                        0.0s
 => [internal] load metadata for docker.io/library/ubuntu:bionic                                                                                       0.0s
 => CACHED [internal] helper image for file operations                                                                                                 0.0s
 => [1/3] FROM docker.io/library/ubuntu:bionic                                                                                                         0.0s
 => [internal] load build context                                                                                                                      0.0s
 => => transferring context: 1.41kB                                                                                                                    0.0s
 => CACHED [2/3] RUN apt update && apt install -y     btrfs-tools     ca-certificates     dumb-init     iproute2                                       0.0s
 => [3/3] ADD ./rc/*.tgz /usr/local                                                                                                                    1.9s
 => exporting to image                                                                                                                                 0.9s
 => => exporting layers                                                                                                                                0.9s
 => => writing image sha256:e0edec6723849d4aa97627de886d8fdfbcf43638b5001317a6e20a8c52dce258                                                           0.0s
 => => naming to docker.io/library/a                                                                                                                   0.0s

docker run --entrypoint /bin/sh a -c 'ls /usr/local | grep concourse'
concourse-5.0.0-rc.20-linux-amd64.tgz
</pre>
</details>

<details>
<summary>without wildcard</summary>
<pre>
docker build -t a .
[+] Building 0.1s (9/9) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                   0.0s
 => => transferring dockerfile: 1.12kB                                                                                                                 0.0s
 => [internal] load .dockerignore                                                                                                                      0.0s
 => => transferring context: 2B                                                                                                                        0.0s
 => [internal] load metadata for docker.io/library/ubuntu:bionic                                                                                       0.0s
 => [1/3] FROM docker.io/library/ubuntu:bionic                                                                                                         0.0s
 => [internal] helper image for file operations                                                                                                        0.0s
 => [internal] load build context                                                                                                                      0.0s
 => => transferring context: 1.41kB                                                                                                                    0.0s
 => CACHED [2/3] RUN apt update && apt install -y     btrfs-tools     ca-certificates     dumb-init     iproute2                                       0.0s
 => CACHED [3/3] ADD ./rc/concourse-5.0.0-rc.20-linux-amd64.tgz /usr/local                                                                             0.0s
 => exporting to image                                                                                                                                 0.0s
 => => exporting layers                                                                                                                                0.0s
 => => writing image sha256:84d4de2b138faf696517919c50dc4680c6584b0b75fa52e98daf41b73fe51a4f                                                           0.0s
 => => naming to docker.io/library/a                                                                                                                   0.0s

docker run --entrypoint /bin/sh a -c 'ls /usr/local | grep concourse'
concourse
</pre>
</details>

Pretty crazy

cirocosta: Now, with "regular docker":

<details>
<summary>with wildcards but without buildkit</summary>
<pre>
docker build -t a .
Sending build context to Docker daemon  189.9MB
Step 1/15 : FROM ubuntu:bionic
 ---> 1d9c17228a9e
Step 2/15 : ENV CONCOURSE_SESSION_SIGNING_KEY     /concourse-keys/session_signing_key
 ---> Using cache
 ---> 2b83f8901a32
Step 3/15 : ENV CONCOURSE_TSA_AUTHORIZED_KEYS     /concourse-keys/authorized_worker_keys
 ---> Using cache
 ---> 9a9c5bb963ac
Step 4/15 : ENV CONCOURSE_TSA_HOST_KEY            /concourse-keys/tsa_host_key
 ---> Using cache
 ---> d375a751074d
Step 5/15 : ENV CONCOURSE_RESOURCE_TYPES          /concourse-resource-types
 ---> Using cache
 ---> 3f29985eb47c
Step 6/15 : ENV CONCOURSE_TSA_PUBLIC_KEY          /concourse-keys/tsa_host_key.pub
 ---> Using cache
 ---> 3d8b26416bc8
Step 7/15 : ENV CONCOURSE_TSA_WORKER_PRIVATE_KEY  /concourse-keys/worker_key
 ---> Using cache
 ---> c3b605c29336
Step 8/15 : ENV CONCOURSE_GARDEN_DNS_PROXY_ENABLE true
 ---> Using cache
 ---> d7cbd9ddc87e
Step 9/15 : ENV CONCOURSE_WORKER_GARDEN_DNS_PROXY_ENABLE true
 ---> Using cache
 ---> 892f1397ed33
Step 10/15 : ENV CONCOURSE_WORK_DIR                /worker-state
 ---> Using cache
 ---> f4009a1d1ded
Step 11/15 : ENV CONCOURSE_WORKER_WORK_DIR         /worker-state
 ---> Using cache
 ---> ad87b391b231
Step 12/15 : VOLUME /worker-state
 ---> Using cache
 ---> 54a045c252a0
Step 13/15 : RUN apt update && apt install -y     btrfs-tools     ca-certificates     dumb-init     iproute2
 ---> Using cache
 ---> 702389661ca0
Step 14/15 : ADD ./rc/*.tgz /usr/local
 ---> c96116a680cc
Step 15/15 : ENTRYPOINT ["dumb-init", "/usr/local/bin/concourse"]
 ---> Running in d3f0a65d232f
Removing intermediate container d3f0a65d232f
 ---> 016d336bd83c
Successfully built 016d336bd83c
Successfully tagged a:latest

docker run --entrypoint /bin/sh a -c 'ls /usr/local | grep concourse'
concourse
</pre>
</details>

🤷‍♂️

vito: That's annoying! I guess as a workaround we could just rename the file to a stable name like `concourse.tgz` so we don't need to wildcard in the Dockerfile.

cirocosta: Ah, before I saw your comment, I had a patch locally using multi-stage builds, which end up achieving the goal in both cases (regular docker and buildkit). 

Do you think that's ok to keep?

thx!

vito: That sounds better, it'd be nice to just re-use the `builder` task without modifications.

ci/dockerfiles/rc/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# enable DNS proxy to support Docker's 127.x.x.x DNS server
+ENV CONCOURSE_GARDEN_DNS_PROXY_ENABLE         true
+ENV CONCOURSE_WORKER_GARDEN_DNS_PROXY_ENABLE  true
+
+# auto-wire work dir for 'worker' and 'quickstart'
+ENV CONCOURSE_WORK_DIR                /worker-state
+ENV CONCOURSE_WORKER_WORK_DIR         /worker-state
+
+# location for base resource types
+ENV CONCOURSE_RESOURCE_TYPES          /usr/local/concourse/resource-types
```
vito: this can be removed - the binary auto-discovers it:

https://github.com/concourse/concourse/blob/74730b03b769ae2164bae19747d326ca354faddc/bin/cmd/concourse/worker.go#L66

cirocosta: Aaah nice, thx!! Updated!

PR #3056, opened 2019-01-16
POC for fly curl

simonjohansson:
Howdie!

Sometimes we do stuff that maybe we should not. Stuff like programatically disable a resources version or doing other things with side effects that are not exposed via the fly cli. We maintain a fairly large Concourse cluster with ~550 pipelines so sometimes we just need to get our hands dirty.

For these cases it would be super neat to be able to

```
fly -t ${CONCOURSE_TEAM} curl -XPUT /api/v1/teams/${CONCOURSE_TEAM}/pipelines/${PIPELINE_NAME}/resources/version/versions/${ID}/disable
```

instead of doing janky bash/curl stuff

```
AUTH="Authorization: Bearer$(grep 'value: ' ~/.flyrc | cut -d: -f2)"
VERSION_URL="${CONCOURSE_URL}/api/v1/teams/${CONCOURSE_TEAM}/pipelines/${PIPELINE_NAME}/resources/version/versions"
curl -sSH "${AUTH}" -X PUT ${VERSION_URL}/${ID}/disable
```

Here is a untested POC that can do GET's. If this is something you would merge let me know and we can test and properly implement the functionality.

```
$ pwd
/Users/sjo4473/src/concourse/fly
$ go run main.go -t myTeam curl /api/v1/teams/myTeam/pipelines
HTTP/1.1 200 OK
Transfer-Encoding: chunked
Connection: keep-alive
Content-Type: application/json
Date: Wed, 16 Jan 2019 12:44:08 GMT
Server: nginx/1.15.6
Strict-Transport-Security: max-age=15724800; includeSubDomains
Vary: Accept-Encoding
X-Concourse-Version: 4.2.2
X-Download-Options: noopen


[{"id":738,"name":"myPipeline","paused":true,"public":false,"team_name":"myTeam"},......
```

### DIFF
--- a/fly/commands/curl.go
+++ b/fly/commands/curl.go
@@ -1,0 +1,51 @@
+package commands
+
+import (
+	"fmt"
+	"github.com/concourse/concourse/fly/commands/internal/curler"
+	"github.com/concourse/concourse/fly/rc"
+	"github.com/pkg/errors"
+	"strings"
+)
+
+type CurlCommand struct {
+}
+
+func (command *CurlCommand) validate(args []string) error {
+	if len(args) != 1 {
+		return errors.New("please provide a path you wish to curl")
+	}
+
+	if !strings.HasPrefix(args[0], "/api") {
+		return errors.New("path must start with /api")
+	}
+	return nil
+}
+
+func (command *CurlCommand) Execute(args []string) error {
+	target, err := rc.LoadTarget(Fly.Target, Fly.Verbose)
+	if err != nil {
+		return err
+	}
+
+	if err := target.Validate(); err != nil {
+		return err
+	}
+
+	if err := command.validate(args); err != nil {
+		return err
+	}
+
+	curl := curler.New(target.Client().HTTPClient(), target.URL())
+
+	body, respHeaders, err := curl.It(args[0])
+	if err != nil {
+		return err
+	}
+
+	fmt.Println(string(respHeaders))
+	fmt.Println(string(body))
+
+
+	return nil
+}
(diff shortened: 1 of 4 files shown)


vito:
I'm down. Just a thought: how would you feel about having this simply be a shim for exec'ing to `curl` with the token automatically set? And forwarding all the flags/etc.?

e.g.:

```sh
fly -t target curl /api/v1/builds -- -XPUT -H "Content-type: ..."
```

This would just run:

```sh
curl -H "Authorization: (token)" (target)/api/v1/builds -XPUT -H "Content-type: ..."
```

This would avoid needing to mimic `curl`s interface with our own flags and such.

simonjohansson:
@vito 
`Just a thought: how would you feel about having this simply be a shim for exec'ing to curl with the token automatically set? And forwarding all the flags/etc.?`

A coworker had the same idea, but what about our poor widows friends who might not have curl installed?

Looking at `cf curl` we should be able to clean up and borrow that implementation..

vito:
> A coworker had the same idea, but what about our poor widows friends who might not have curl installed?

I'm not too worried about this - `curl` is pretty easy to install on Windows: 

* https://chocolatey.org/packages/curl
* https://curl.haxx.se/windows/

...and we don't exactly expect everyone to be using this command in all kinds of different contexts, no? With `cf` it was a bigger problem because we couldn't make as many assumptions and I believe had a few documented features that required `cf curl`.

simonjohansson:
Closing this issue as we developed it on a coworkers fork. :)

PR #3080, opened 2019-01-17
fly curl command

robwhitby:
Hi, this adds the `fly curl` command, as discussed here: https://github.com/concourse/concourse/pull/3056

```fly curl PATH [-I] [-X METHOD] [-H HEADER] [-d DATA]```

Examples:
```
fly curl /api/v1/pipelines
fly curl /api/v1/teams/myteam/pipelines/mypipeline/config -X PUT -D @pipeline.yml -H 'application/yaml'
```

Options:
```
-X  HTTP method
-H  Custom headers to include in the request, flag can be specified multiple times
-d  HTTP data to include in the request body, or '@' followed by a file name to read the data from
-I  Output response headers
```

### DIFF
--- a/fly/commands/curl.go
+++ b/fly/commands/curl.go
@@ -1,0 +1,71 @@
+package commands
+
+import (
+	"fmt"
+	"github.com/concourse/concourse/fly/rc"
+	"io/ioutil"
+	"net/http/httputil"
+	"os"
+)
+
+type CurlCommand struct {
+	Method        string   `short:"X" description:"method" default:"GET"`
+	OutputHeaders bool     `short:"I" description:"output response headers"`
+	Headers       []string `short:"H" description:"HTTP Header"`
+	Body          string   `short:"d" description:"Body"`
+}
+
+func (command *CurlCommand) Execute(args []string) error {
+	target, err := rc.LoadTarget(Fly.Target, Fly.Verbose)
+	if err != nil {
+		return err
+	}
+
+	err = target.Validate()
+	if err != nil {
+		return err
+	}
+
+	if len(args) == 0 {
+		return fmt.Errorf("path required")
+	}
+
+	cr := CurlRequest{
+		Host:    target.URL(),
+		Path:    args[0],
+		Method:  command.Method,
+		Headers: command.Headers,
+		Body:    command.Body,
+	}
+
+	req, err := cr.CreateHttpRequest()
+	if err != nil {
+		return err
+	}
+
+	res, err := target.Client().HTTPClient().Do(req)
+	if err != nil {
+		return err
+	}
+
+	resBody, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return err
+	}
+
+	if res.StatusCode < 200 || res.StatusCode > 299 {
+		return fmt.Errorf("%s\n%s", res.Status, resBody)
+	}
+
+	if command.OutputHeaders {
+		headers, err := httputil.DumpResponse(res, false)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by simonjohansson** [comment]
@vito 

We had a look at just passing options to curl directly but we feel the user experience might suffer a bit.

Given an implementation like 

```golang
func (command *CurlCommand) Execute(args []string) error { 
  //validate
  return exec.Cmd{
    Path: pathToCurl,
    Args: append([]string{pathToCurl}, args...)
  }.Run()
}
```

Two issues,

### If we wish to pass args to curl as opposed to the fly we would have to send the end of options delimiter.

```
fly -t blah curl -- -H 'headerName: headerValue' -I /api/blah/bluh
```

Kinda confusing

### Supplying the path.
Since we need to construct the full URL from target.URL() + path we have three obvious options.

Three options
#### Passing it to the fly cli as option
```
fly -t blah curl --path /api/blah/bluh -- -H 'headerName: headerValue' -I
```

Weird that the path is part of the fly command, but the options are not.

#### Requiring that the first arg is a path
```
fly -t blah curl -- /api/blah/bluh -H 'headerName: headerValue' -I
```

Does not alight with how curl works, since..

#### Trying to guess where the path is
curl allows options to be passed either left or right of the url.

```
curl -I https://concourse-ci.org == curl https://concourse-ci.org -I
```

Given a call where the path is sandwitched between two options

```
fly -t blah curl -- -H 'headerName: headerValue' /api/blah/bluh -I
```
We need to loop trough the args and try to figure out which of them looks like a path.
We need to guess which of the args are infact the path.

vito:
First off thanks for fleshing this out in another PR!

I kinda figured the path would be the first (required) argument and the rest (if specified) would just be forwarded to `curl`:

```sh
fly -t blah curl /api/v1/foo -- -H 'Content-type: application-json' -d '{"foo":"bar"}'
```

Having the path separate doesn't feel weird to me because it's part of the reason for using `fly curl` instead of just `curl`. It's a little shim just to make `curl`ing easier. This could be an explicit part of the help-text by using `positional-args`:

```go
type CurlCommand struct {
	Args struct {
		Path   string `positional-arg-name:"PATH" required:"true"`
		Rest []string `positional-arg-name:"curl flags"`
	} `positional-args:"yes"`
}
```

Using `--` also doesn't feel that odd from a UNIXy standpoint - there's already precedent for it with `fly intercept`, and if this is actually a dev shim to make `curl`ing easier it makes a lot of sense. It would be nice if it wasn't necessary, but unfortunately it looks like that's a global setting and can't be set on just the `curl` command.

I guess one annoying thing with shelling out to `curl` would be needing to pass a file for the `--cacert` flag. That's currently embedded in `~/.flyrc` so you'd need to write it to a temporary file on each run. Kinda clunky but workable.

I just want to make sure we're not feeling a need to replicate `cf curl` - it's higher maintenance, has different motivations (i.e. portability), and most folks using `fly curl` won't even know what `cf curl` is. Coming at this without the CF context I'd expect a command called `curl` to run `curl`. :man_shrugging: For example I may at some point want to run something like `fly curl / -- -vvv` to debug routing issues.

simonjohansson:
@vito 
> First off thanks for fleshing this out in another PR!

Thanks for a detailed rejection :)

> I guess one annoying thing with shelling out to curl would be needing to pass a file for the --cacert flag. That's currently embedded in ~/.flyrc so you'd need to write it to a temporary file on each run. 

Under what circumstances is this true? My ~/.flyrc does not contain this.

> I just want to make sure we're not feeling a need to replicate cf curl

Fair enough.

PR #3081, opened 2019-01-17
Add Azure Key Vault credential provider

thomastaylor312:
Adds credential provider for Azure Key Vault. This has all needed unit tests (I think) and has been manually validated as well


thomastaylor312:
I'll be working on a PR for docs as well. Not sure if that need to be in place before this can be merged

thomastaylor312:
The TestFlight failure _seems_ like a flake, but I am not familiar enough with the systems to be completely sure

vito:
@thomastaylor312 I'll re-trigger `testflight` - it does look like a flake.

I probably won't get a chance to look at this much today or over the weekend, but it looks really cool - thanks for working on it! And thanks for submitting docs too!

thomastaylor312:
Rebased and pushed to fix merge conflict

vito:
Hey, just checking again - I probably won't have time to review this until after 5.0 is out the door, since we're focusing on that right now. :slightly_smiling_face: It's been a long time since we've shipped so I'm holding off on any last-minute big features for now.

vito:
@thomastaylor312 Once 5.0 is finally out I'd like to get back to a much more frequent release cadence. Something like 2 weeks between releases was typical before.

So maybe mid-February?

However disaster tends to happen after we sit on a release for too long, so I can't really provide a strong guarantee. :stuck_out_tongue:

vito:
Shockingly enough the 5.0 release has largely gone without incident, so I should be able to review this soon. Thanks for waiting!

atc/creds/keyvault/reader.go
```diff
@@ -0,0 +1,111 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/Azure/azure-sdk-for-go/services/keyvault/v7.0/keyvault"
+	"github.com/Azure/go-autorest/autorest"
+)
+
+// ErrSecretNotFound is the error returned when a secret cannot be found by the
+// reader
+var ErrSecretNotFound = fmt.Errorf("secret not found")
```
vito: Looks like this is unused? Probably 'cause `Get` just returns a `nil, false, nil` in this case anyway.

atc/creds/keyvault/reader.go
```diff
@@ -0,0 +1,111 @@
...
+	if !ok || e.Response == nil {
+		return false
+	}
+	return e.Response.StatusCode == http.StatusNotFound
+}
+
+// FakeReader is a wrapper that implements the SecretReader interface for
+// testing. Each function can be set independently depending on the needs of
+// the test
+type FakeReader struct {
```
vito: FWIW we typically use [`counterfeiter`](https://github.com/maxbrunsfeld/counterfeiter) for this kind of thing - it's used like this:

```go
// go:generate counterfeiter . InterfaceName

type MyInterface interface {
    // ...
}
```

and then `go generate ./...` will create `fakes/fake_my_interface.go`

we have a TODO to add this to the contributing docs (https://github.com/concourse/concourse/issues/3404)

atc/creds/keyvault/reader.go
```diff
@@ -0,0 +1,111 @@
...
+	return &keyVaultReader{
+		client:      client,
+		keyVaultURL: keyVaultURL,
+	}
+}
+
+// Get returns the value of the given secret name, a bool if it was found, and
+// an error if there was an issue fetching the secret
+func (k *keyVaultReader) Get(name string) (string, bool, error) {
+	// Right now we aren't timing out, we may want to add a configurable option
```
vito: Makes sense. I'm OK with this as-is for now. This may be something to consider generically rather than having each credential manager implement their own timeout flags.

Made a note for this here: https://github.com/concourse/rfcs/issues/5#issuecomment-472427713

atc/creds/keyvault/manager_test.go
```diff
@@ -0,0 +1,53 @@
+package keyvault
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+)
+
+func TestHealth(t *testing.T) {
+	t.Run("a working client should return good health", func(t *testing.T) {
```
vito: We typically use [`ginkgo`](https://github.com/onsi/ginkgo) for tests but I won't split hairs. (More info: https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md#testing-your-changes)

atc/creds/keyvault/keyvault.go
```diff
@@ -0,0 +1,74 @@
...
+	for i := range secrets {
+		variables[i] = varTemplate.VariableDefinition{
+			Name: secrets[i],
+		}
+	}
+	return variables, nil
+}
+
+func (k *keyVault) path(parts ...string) string {
+	return fmt.Sprintf("%s-%s-%s", k.PathPrefix, k.TeamName, strings.Join(parts, "-"))
```
vito: Is it possible to exploit this by putting a `-` in my team or pipeline name to access another team's credentials?

For example, `concourse-foo-bar-x` could either mean "team `foo-bar`, credential `x`" or "team `foo`, pipeline `bar`, credential `x`".

Is there another separator we could use?

To be honest we should probably just move credential manager configuration to the team-level (as mentioned in concourse/rfcs#5), but for now we have to watch out for this kind of thing.

atc/creds/keyvault/reader.go
thomastaylor312: Oh, thanks for the pointer. I'll update that

atc/creds/keyvault/manager_test.go
thomastaylor312: Yeah, I wasn't sure on that because some of them (I think the vault one) didn't use ginkgo. I'll change it over so it doesn't diverge from the core project

atc/creds/keyvault/keyvault.go
thomastaylor312: Unfortunately, Key Vault doesn't allow for other kinds of separators. Let me see what I can do to perform some validation because your concern is totally valid

vito: Dang, that's pretty strict. It doesn't even allow underscores. :eyes:

thomastaylor312: So I am guessing the only way we can handle this is by not allowing teams with a dash in them for secret names. Basically we'd remove all dashes from the team name and end up with `concourse-foobar-x` for team `foo-bar` and `concourse-foo-bar-x` for "team `foo`, pipeline `bar`, credential `x`". 

It is likely though that anyone using this will have a shared key vault anyway because there is no namespacing. Anyone who has edit access to the Key Vault will have access to any of the keys. If this is going to be a complete no-go from the security side of things, let me know. I think the other option would be to have each team with their own key vault (that the configured user for the credential provider has read access to) and we would change which key vault we are connecting to based on the team name

vito: Hmm, only problem then is teams `foo-bar` and `foobar` would end up sharing secrets. Granted, an operator would have had to knowingly create these ambiguous team names.

Well, here are some options I can think of:

1. Implement per-team credential manager configuration first. I'd really like to get around to that, but I don't think I can fit it in to our day-to-day backlogs since we've already got a few top priorities ("spaces", runtime resiliency, API and internal architecture cleanups). We still need to create an actual RFC for this (concourse/rfcs#5 is actually an RF-RFC), but maybe that won't take too long if we keep them small in scope. Would this be something you'd be interested in contributing by any chance? It's definitely a scope creep, so if not, that's perfectly OK - just thought I'd ask. 🙂 
1. Support pre-configuring different key vaults per-team, as you suggested. This would achieve a decent middle ground for now, but it kind of feels like a stopgap solution - it'd be painful to configure and relatively inflexible (enabling it for new teams would require redeploying Concourse).
1. Give up on multi-tenancy for this credential manager for now - forego the team and pipeline variable name scoping logic and just look up credentials by their exact name. This would at least make it obvious that it's one global scope, and it'd still be useful for folks that don't require multi-tenancy. We would just put a big warning on the docs saying this credential manager does not support scoped credentials.

I'm OK with either of these - the last idea is a little risky but I'm hoping by making the credential lookup rules so obviously different from the rest that users wouldn't mistakenly believe that it keeps teams away from each other's secrets.

vito: As a follow-up, I've started on an RFC for team-level credential manager configuration: https://github.com/concourse/rfcs/pull/21

thomastaylor312: So option 2 that you talked about _shouldn't_ require a redeploy. The idea is that the same service principal would need to be granted access, so it should be able to establish a new connection with the given key vault.

I do like the per-team credential manager idea, but I might not have the cycles to do it (you know, day job and all 🙂). Let me see if I can make some time, otherwise I'll probably opt for the stop gap (option 2)

thomastaylor312: Oh, though now that I am thinking about it more, that might not work... I believe all key vault names have to be unique and so someone with a team name like "devops" will quickly run into a naming collision. Let me see if I can find some cycles to do per-team credential management

vito: FWIW, because per-team credential management will be going through the RFC process, it may take a bit longer than you'd like compared to going with one of the short-term solutions here. More info on the whole process here: https://github.com/concourse/rfcs/blob/master/README.md

Any feedback on the RFC would be great in the meantime!

pivotal-jamie-klassen:
@thomastaylor312 now that #4600 has been merged, this opens up the possibility for Azure Key Vault to be used as a "pipeline-scoped-only" credential manager via `var_sources`... Maybe you'd like to take another look at this point?

thomastaylor312:
At this point I haven't touched it in long enough and am not currently doing any concourse work. Feel free to close the PR if that is easier and I'll keep the branch around for anyone who may want to take it on

muntac:
> At this point I haven't touched it in long enough and am not currently doing any concourse work. Feel free to close the PR if that is easier and I'll keep the branch around for anyone who may want to take it on

Closing PR as per comment above.

PR #3083, opened 2019-01-17
Added container type metrics

kcmannem:
ATC, TSA 

Each container is now classed as "build" or "check", and the number of containers in each class is measured as a metric with each heartbeat.

### DIFF
--- a/atc/metric/metrics.go
+++ b/atc/metric/metrics.go
@@ -136,7 +136,7 @@ func (event WorkerContainers) Emit(logger lager.Logger) {
 	emit(
 		logger.Session("build-containers-count"),
 		Event{
-			Name:  "build containers",
+			Name:  "worker build containers",
 			Value: event.BuildContainers,
 			State: EventStateOK,
 			Attributes: map[string]string{
@@ -148,7 +148,7 @@ func (event WorkerContainers) Emit(logger lager.Logger) {
 	emit(
 		logger.Session("check-containers-count"),
 		Event{
-			Name:  "check containers",
+			Name:  "worker check containers",
 			Value: event.CheckContainers,
 			State: EventStateOK,
 			Attributes: map[string]string{
--- a/atc/metric/metrics.go
+++ b/atc/metric/metrics.go
@@ -112,9 +112,11 @@ func (event SchedulingJobDuration) Emit(logger lager.Logger) {
 }
 
 type WorkerContainers struct {
-	WorkerName string
-	Platform   string
-	Containers int
+	WorkerName       string
+	Platform         string
+	ActiveContainers int
+	BuildContainers  int
+	CheckContainers  int
 }
 
 func (event WorkerContainers) Emit(logger lager.Logger) {
@@ -122,7 +124,7 @@ func (event WorkerContainers) Emit(logger lager.Logger) {
 		logger.Session("worker-containers"),
 		Event{
 			Name:  "worker containers",
-			Value: event.Containers,
+			Value: event.ActiveContainers,
 			State: EventStateOK,
 			Attributes: map[string]string{
 				"worker":   event.WorkerName,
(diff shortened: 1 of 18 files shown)


ddadlani:
Is `activeContainers` now basically just `buildContainers` + `checkContainers`? Does it make sense to keep `activeContainers` in the heartbeat?

atc/metric/metrics.go
```diff
@@ -111,24 +111,51 @@ func (event SchedulingJobDuration) Emit(logger lager.Logger) {
...
+			Name:  "build containers",
+			Value: event.BuildContainers,
+			State: EventStateOK,
+			Attributes: map[string]string{
+				"worker": event.WorkerName,
+			},
+		},
+	)
+
+	emit(
```
cirocosta: Hey,

Given that for `prometheus` we still need to specially treat the metrics in the emitter, it'd be nice to have a case for check containers https://github.com/concourse/concourse/blob/2c374aaaf6ebc619df4439ca6d21c68a59c77fde/atc/metric/emitter/prometheus.go#L333 (or did I miss something?)

Thx!

kcmannem:
@ddadlani not after #3025, since it'll be creating containers periodically which won't be accounted as check or build.

vito:
@kcmannem Do you have time to come back to this? Looks like we just need to add Prometheus support and have the heartbeater ignore containers that don't have properties?

vito:
Woops, now it needs a rebase. :sweat_smile:

kcmannem:
@vito done

cirocosta:
Hey,

I was trying this PR out and I'm not 100% if it's working as intended.

I've set the following pipeline:

```
resources:
  - name: image
    type: registry-image
    source:
      repository: busybox

jobs:
  - name: job
    plan:
      - get: image
      - task: something
        image: image
        config:
          platform: linux
          run:
            path: true
```

Then I triggered the job and waited for a while (such that the registration time would kick in and the metrics would be updated), but it seems like that doesn't really happen 🤔  

```sh
fly -t local workers --json | jq '.[] | .check_containers'
0
fly -t local containers --json | jq '.[] | select(.type == "check") | .id'
"a19439b8-4833-4462-7181-c1f35916f139"
```

I thought that it could be that `fly` would need to change something, but using `fly curl` reveals the same.

Am I missing something? 🤔  

Thanks!

ddadlani:
@cirocosta we tried the same on our local deployment and we don't see 0, we see null implying that the presenter doesn't allow you to see that info.

```
~/w/concourse (cont_metrics) $ fly -t dev workers --json | jq '.[] | .check_containers'
null
```

cirocosta:
Hey,

Sorry for the super slow feedback on this.

There's an error going on when it comes the time to emit the metrics that is preventing the metric from being exposed:

```sh
{"timestamp":"2019-03-28T14:06:40.474047000Z","level":"error","source":"atc","message":"atc.build-containers-count.emit.failed-to-find-platform-in-event","data":{"error":"expected platform to exist in event.Attributes","session":"1158.1"}}
{"timestamp":"2019-03-28T14:06:40.474197200Z","level":"error","source":"atc","message":"atc.check-containers-count.emit.failed-to-find-platform-in-event","data":{"error":"expected platform to exist in event.Attributes","session":"1159.1"}}
```

which makes sense, as we're not emitting the platform:

```
	emit(
		logger.Session("build-containers-count"),
		Event{
			Name:  "worker build containers",
			Value: event.BuildContainers,
			State: EventStateOK,
				"worker": event.WorkerName,
			},
		},
	)
```

Thanks!

kcmannem:
@cirocosta I think i fixed it this time. I looked at the output from localhost:9001 and saw the metric being emitted.

marco-m:
Hey @cirocosta in my understanding this is ready for your review, when you have time. Thanks :-)

kcmannem:
closed since, we're reconsidering to have the workers themselves report metrics instead of flowing through atc.

marco-m:
Hello @kcmannem 

> we're reconsidering to have the workers themselves report metrics instead of flowing through atc.

For my understanding: do you mean that you are considering having each worker expose Prometheus metrics ?

Sounds good to me, and more Prometheus-idiomatic.

kcmannem:
Yeah thats why my convo with @cirocosta led to regarding this pr. I'll reopen this if his direction changes.

Issue #3013, opened 2019-01-08
Feature-flag 'global resources' functionality


vito:
## What challenge are you facing?

Global resources (#2386) is great and all but we're now aware of two different situations where we *don't* want it: the time resource (to avoid stampeding herd of periodic jobs on workers) and resources that use IAM roles to determine resource access (#3002).

## What would make this better?

We should make it opt-in for now until we can more fully understand the impact of this feature and what changes may be necessary in order to make it safe to enable by default.

Most of the work for this is already done in #2908. I'm proposing that we add a flag to the ATC, `--enable-global-resources`, and have it default to **`false`** for now.

If the value is set to `false`, we should never share version history across resources regardless of whether the resource type configured `unique_version_history` (per #2908).

As this feature is primarily beneficial to large multi-tenant installations, it can be up to their operator's discretion whether they want to enable the feature or not.

clarafu:
Note that the flag only applies to having resources opt-in for sharing resource versions, resource type versions will always be shared even if the flag is applied. We decided to do this because resource type versions are not visible to users (since it pretty much always uses the latest version).

PR #3090, opened 2019-01-18
Add support for unique version history

### LINKED ISSUE #2908: Time resource shouldn't share version history

With #2386 we identified that the `time` resource would be adversely affected by a shared history, because it'll fire off a ton of builds at once if everyone configures the same interval.

In concourse/rfcs#1 we're planning to make this feature (shared history) opt-out, by allowing the resource's `info` script to include some configuration that disables it.

We don't want to do all of concourse/rfcs#1 just yet, so for now let's just hack in some way for v1 resources to opt-out.

Implementation note discussed with @clarafu: this could work by generating a random hash for resources that have opted-out and using it to generate a unique resource config for the resource. This would allow everything else to just keep working as-is (versions remain tied to resource configs, and we'll just have different resource configs).

clarafu:
fixes #2908

required for #3013


atc/db/resource.go
```diff
@@ -37,28 +37,31 @@ type Resource interface {
...
 
 	EnableVersion(rcvID int) error
 	DisableVersion(rcvID int) error
 
 	PinVersion(rcvID int) error
 	UnpinVersion() error
 
-	SetResourceConfig(lager.Logger, atc.Source, creds.VersionedResourceTypes) (ResourceConfig, error)
+	SetResourceConfig(lager.Logger, atc.Source, creds.VersionedResourceTypes) (ResourceConfigScope, error)
 	SetCheckError(error) error
```
vito: This should probably move to scopes, because of the following scenario:

* assume two teams have a resource config but with unique version history, i.e. different scopes
* there are no global workers
* team A has workers, team B doesn't
* team B tries to check, but fails because there are no workers, and sets the check error
* team A tries to check, works, clears out error

These two teams would be flipping the error on and off, and team A would see a bogus error.

vito:

PR #3094, opened 2019-01-18
Feature-flag shared version history

### LINKED ISSUE #3013: Feature-flag 'global resources' functionality

clarafu:
fixes #3013


vito:

PR #3101, opened 2019-01-21
Web: Refactoring effects from each separate screen into a single datatype

xtreme-james-cooper:
- We moved the effects from Dashboard, Pipeline, Build, etc into a single Effects file;
- We added a callback message type to handle returns from effects
- Also some misc refactoring in SubPage to take advantage of the cleaner new interface


vito:
Good to merge this once this goes green.

Do all the commits have the right authors? No big deal, just noticed a few in the middle may be misattributed due to a rebase or committing on a different machine something.

PR #3114, opened 2019-01-21
check containers can be placed on team workers

### LINKED ISSUE #2951: Allow check containers to be created on team scoped workers
Since #2910 has been completed, there have been a few questions about whether this behavior is valid, since there is a possibility that there are no "global" (non-team) workers available in a cluster.
(eg. https://github.com/concourse/concourse/blob/24127b20b46d6f563ccf13e5b39c3645292047ea/topgun/worker_landing_test.go#L219-L241)

We should ensure that check containers can be created on team scoped workers where necessary, and take into account the following:
1. Check containers for a team without team-scoped workers will be on global workers
1. There may be 2 check containers for identical resources if one already exists on a team-scoped worker - because container creation prefers team scoped workers - and a team with no team-scoped workers needs to check.
1. hijacking check containers is going to be changed to be admin only, so this will need to account for 1 and 2. We may want to add a command to debug the check which creates a fresh resource check container available to the admin user.

clarafu:
fixes #2951


vito:

PR #3119, opened 2019-01-22
bump hashicorp/go-msgpack to avoid bzr deps

vito:
effectively works around https://github.com/hashicorp/go-msgpack/pull/2 (they commented out the entire file upstream)


jefferai:
Yeah, when I was making the package go-modulized I was trying to figure out why it was pulling in so many deps, and it turns out it's because of comparitive benchmarks that the original package this was forked from ran, one of which was for bson. There's really no reason we need to be running those, so I commented the whole thing out.

PR #3121, opened 2019-01-22
web: Mac cli link now goes to 'darwin' rather than 'osx'

### LINKED ISSUE #3068: Web uses `osx` query param instead of `darwin`
## Bug Report

I noticed that with 5.x.x, `web` is using `osx` as the query parameter when the user clicks on the apple icon.

Given that `web` uses such query parameter to search for the file to serve from the filesystem, it ends up not being able to do so (as in our RCs we use - and have been using in 4.x series as well - `darwin` instead of OSX).

## Steps to Reproduce

1. go to https://ci.concourse-ci.org
2. click on the apple icon
3. see it failing: https://ci.concourse-ci.org/api/v1/cli?arch=amd64&platform=osx
4. verify that it should indeed fail as the file extracted under `fly-assets` is named `fly-darwin...` by opening any public `rc`: https://storage.googleapis.com/concourse-artifacts/rcs/concourse-5.0.0-rc.22-linux-amd64.tgz 

## Expected Results

Download the file

## Actual Results

No download 😁 

## Additional Context

To ensure that something similar like this doesn't happen again, should we add a test for it? Wdyt? 

## Version Info

* Concourse version: master 5.x (1dc23e7beb78c939a151d30423453d2576da94d1)
* Deployment type (BOSH/Docker/binary): docker & bosh

pivotal-jamie-klassen:
Fixes #3068

### DIFF
--- a/web/elm/src/Concourse/Cli.elm
+++ b/web/elm/src/Concourse/Cli.elm
@@ -1,12 +1,76 @@
-module Concourse.Cli exposing (downloadUrl, Cli(..))
+module Concourse.Cli exposing (Cli, clis, downloadUrl, iconUrl, id, label)
 
 
-downloadUrl : String -> String -> String
-downloadUrl arch platform =
-    "/api/v1/cli?arch=" ++ arch ++ "&platform=" ++ platform
+clis : List Cli
+clis =
+    [ OSX, Windows, Linux ]
 
 
 type Cli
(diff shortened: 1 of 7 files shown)


pivotal-jamie-klassen:
Gonna re-submit because this seems to be caused by a testflight flake.

Issue #2702, opened 2018-10-19
pinning/unpinning a resource version from fly


pivotal-jamie-klassen:
It seems appropriate, if we are going to be able to pin resource versions from the web UI (#2508) then we should also be able to do it from fly. I suspect the basic syntax might be

```
fly -t <target> pin-resource -r <pipeline-name>/<resource-name> -v <version>
fly -t <target> unpin-resource -r <pipeline-name>/<resource-name>
```

A potential problem here is that resource versions could be arbitrary JSON? I suspect it would be a bit unwieldy to say something like

```
fly -t dev pin-resource -r booklit/booklit -v {"ref": "723f16754b8939a486fcdd35d569a0686ff9ceec"}
```

so maybe it would make more sense to provide some kind of jsonpath/regex expression for `<version>`.

xtremerui:
In order to pin a resource, one must know the version first. 

For now in Fly there is no command for listing resources https://github.com/concourse/concourse/issues/869 or resource versions. The resource versions are only shown on web UI or fetched from api endpoint `/api/v1/teams/:team_name/pipelines/:pipeline_name/resources/:resource_name/versions`, whose example response for team `main` pipeline `concourse` resource `mock-resource` is

```
{
	id: 740908,
	type: "github-release",
	metadata: [{
			name: "name",
			value: "v0.5.1"
		},
		{
			name: "url",
			value: "https://github.com/concourse/mock-resource/releases/tag/v0.5.1"
		},
		{
			name: "body",
			value: ""
		},
		{
			name: "tag",
			value: "v0.5.1"
		},
		{
			name: "commit_sha",
			value: "70b8b0efe61d2eeb877f1f30720d972fdeecf0cd"
		}
	],
	resource: "mock-resource",
	version: {
		tag: "v0.5.1"
	},
	enabled: true
}
```
one would need to parse the version out programmatically.

chenbh:
Started working on this with @cirocosta, I believe we are going with 
```
fly -t <target> pin-resource -r <pipeline-name>/<resource-name> -i <version-id>
fly -t <target> unpin-resource -r <pipeline-name>/<resource-name>
```
Where you can get the `version-id` by calling
```
$ fly -t <target> resource-versions --resource=<pipeline-name>/<resource-name>

id      version                                       enabled
148973  ref:6e6da90ca9cef2b293482545ed0ac2fb5ca568bc  yes
[...]

PR #3131, opened 2019-01-23
Issue 2702/pin unpin

chenbh:
Fix for concourse/concourse#2702

### DIFF
--- a/atc/api/resourceserver/versionserver/pin_version.go
+++ b/atc/api/resourceserver/versionserver/pin_version.go
@@ -30,7 +30,11 @@ func (s *Server) PinResourceVersion(pipeline db.Pipeline) http.Handler {
 			return
 		}
 
-		err = resource.PinVersion(resourceConfigVersionID)
+		found, err = resource.PinVersion(resourceConfigVersionID)
+		if !found {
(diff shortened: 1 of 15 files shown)


vito:
@chenbh I think most of those concerns are addressed by the suggestion to query by partial fields. We could keep the existing pin-by-ID API, and just add an endpoint for querying versions by fields. The CLI would use that to get the ID to pin, and this would support pinning without fully specifying every field of the version.

The endpoint could be something like:

```
fly -t foo pin-resource-version -r bar/baz -v pr:3178
=> GET /api/v1/teams/foo/pipelines/bar/resources/baz/versions?filter=pr:3178
=> if not found, error; if multiple found, choose latest
=> PUT /api/v1/teams/foo/pipelines/bar/resources/baz/versions/12345/pin
```

I don't think we would really expect users to pass `--id` if this were implemented, though I guess it'd be easy to support alongside it (it'd just skip the lookup step).

vito:
@pivotal-jwinters Actually maybe we can write up an issue for the API part of this mentioned above?

vito:
Closing this for now as it's inactive and blocked on #3932

Issue #3002, opened 2019-01-07
Tagged resources should not re-use check containers from workers that don't have the configured tags


cirocosta:
## Bug Report

I noticed that it might be possible to a resource that can only be checked in a worker that's tagged with a specific tag, but have that never happen due to global resources taking just `source` in consideration (if I understood correctly).

For instance, consider the case where only `aws`-tagged workers are capable of performing certain actions (like, accessing S3 buckets because they have IAM Roles attached to them):

```yaml
- name: release-from-aws
  tags: [ aws ]
  type: s3
  source: &release-source
    bucket: sample-concourse-s3-resource
    initial_path: binary-1.0.0.tgz
    regexp: binary-(.*).tgz
    region_name: us-east-1
    skip_download: true

- name: release-from-gke
  type: s3
  source: *release-source
```

I'm not sure if this is a *bug* or not - should we consider `tags`?

## Steps to Reproduce

1. have two workers - one tagged, the other not
2. create a pipeline with two resources that share the same config, but differ by having `tags` in one of them
3. verify that the checks always happen in one of the workers (global resources!) - not necessarily in the one with the tag.

## Expected Results

I'm not really sure; eventually, happen over both? Although, I guess that could generate weird results if the worker is not only providing credentials. Maybe this could even end up leaking info somehow?

## Version Info

* Concourse version: master (c862322702eea2fd561b1a04de18ab1c1ef4810b)

vito:
Ugh, this is true.

I really regret supporting IAM roles; they're a perfect example of our 'worker state' anti-pattern, but too many people kept asking for them... 🙄

So is the fundamental problem here that the actual 'credentials' are associated to the machine, and not to the 'source'? That would mean that someone could just configure the same source without having access to it and then obtain the version history regardless of their worker's IAM configuration.

They at least couldn't fetch the bits, and couldn't access any cached bits, since they'd only be cached on workers that have IAM access to fetch them in the first place. But it is a scary information leak if they can see the version listing without credentials.

vito:
In the short-term it may be wise to at least disable resource version history sharing for the s3 resource (and any others that use IAM roles). Thankfully we just implemented support for that in #2908.

We could also make this functionality opt-in by default until we're sure it's safe.

vito:
There are two aspects to this:

1. Not re-using an untagged container for a tagged resource's check
1. Not sharing version history for tagged resources (assuming they may be tagged due to IAM roles or something that made the pipeline author decide to tag the resource)

[...]

PR #3133, opened 2019-01-23
Finish up tasks before v5.0.0

### LINKED ISSUE #2908: Time resource shouldn't share version history

### LINKED ISSUE #2951: Allow check containers to be created on team scoped workers

### LINKED ISSUE #3013: Feature-flag 'global resources' functionality

clarafu:
fixes #2908 (add unique version history)
fixes #3013 (feature flag global resources)
fixes #2951 (check containers placed on team workers)


atc/db/worker_factory.go
```diff
@@ -321,7 +321,20 @@ func (f *workerFactory) SaveWorker(atcWorker atc.Worker, ttl time.Duration) (Wor
 	return savedWorker, nil
 }
 
-func (f *workerFactory) FindWorkerForContainerByOwner(owner ContainerOwner) (Worker, bool, error) {
+func (f *workerFactory) FindWorkerForContainerByOwner(owner ContainerOwner, teamID int) (Worker, bool, error) {
```
vito: There seems to be a problem with this - in testing with a team worker, I noticed that a new container would be created every time I executed a check against the team's resources.

Here's my `docker-compose.yml` for setting up a team worker:

```yaml
version: '3'

services:
  db:
    image: postgres
    ports: ["6543:5432"]
    environment:
    - POSTGRES_DB=concourse
    - POSTGRES_USER=dev
    - POSTGRES_PASSWORD=

  web:
    build:
      context: .
      dockerfile: ${DOCKERFILE}
    volumes:
    - .:/src
    command: web
    depends_on: [db]
    ports: ["8080:8080"]
    environment:
    - CONCOURSE_POSTGRES_HOST=db
    - CONCOURSE_POSTGRES_USER=dev
    - CONCOURSE_POSTGRES_PASSWORD=dev
    - CONCOURSE_POSTGRES_DATABASE=concourse
    - CONCOURSE_EXTERNAL_URL=http://localhost:8080
    - CONCOURSE_ADD_LOCAL_USER=test:test,guest:guest
    - CONCOURSE_MAIN_TEAM_LOCAL_USER=test
    - CONCOURSE_LOG_LEVEL=debug
    - CONCOURSE_ENABLE_GLOBAL_RESOURCES=true

  worker:
    build:
      context: .
      dockerfile: ${DOCKERFILE}
    command: worker
    privileged: true
    depends_on: [web]
    ports: ["7777:7777", "7788:7788"]
    environment:
    - CONCOURSE_LOG_LEVEL=debug
    - CONCOURSE_TSA_HOST=web:2222
    - CONCOURSE_EPHEMERAL=${CONCOURSE_EPHEMERAL:-true}

      # avoid using loopbacks
    - CONCOURSE_BAGGAGECLAIM_DRIVER=overlay

      # so we can reach Garden/Baggageclaim for debugging
    - CONCOURSE_BIND_IP=0.0.0.0
    - CONCOURSE_BAGGAGECLAIM_BIND_IP=0.0.0.0

      # so that CI can configure a different network range that doesn't overlap
      # with the outer Garden
    - CONCOURSE_GARDEN_NETWORK_POOL

  team_worker:
    build:
      context: .
      dockerfile: ${DOCKERFILE}
    command: worker
    privileged: true
    depends_on: [web]
    environment:
    - CONCOURSE_TEAM=some-team

    - CONCOURSE_LOG_LEVEL=debug
    - CONCOURSE_TSA_HOST=web:2222
    - CONCOURSE_EPHEMERAL=${CONCOURSE_EPHEMERAL:-true}

      # avoid using loopbacks
    - CONCOURSE_BAGGAGECLAIM_DRIVER=overlay

      # so we can reach Garden/Baggageclaim for debugging
    - CONCOURSE_BIND_IP=0.0.0.0
    - CONCOURSE_BAGGAGECLAIM_BIND_IP=0.0.0.0

      # so that CI can configure a different network range that doesn't overlap
      # with the outer Garden
    - CONCOURSE_GARDEN_NETWORK_POOL
```

Then configure `some-team`:

```sh
$ fly -t l login -c http://localhost:8080 -u test -p test
$ fly -t l set-team -n some-team --local-user guest
```

Log in and set the following pipeline:

```yaml
groups: []
resources:
- name: booklit
  type: git
  source:
    uri: https://github.com/vito/booklit
- name: 3s
  type: time
  source: {interval: 3s}
resource_types: []
jobs:
- name: unit
  plan:
  - get: booklit
    trigger: true
  - get: 3s
  - task: test
    file: booklit/ci/test.yml
```

```sh
$ fly -t lst login -c http://localhost:8080 -u guest -p guest
$ fly -t lst set-pipeline -p booklit -c booklit.yml
```

Then trigger a few checks from the web UI.

There should only be two `check` containers - one for the `time` resource and one for the `git` resource. I saw one new container be created every time I ran the `check`. This can be verified by looking at the database:

```sh
$ ./hack/db
```

```sql
select rccs.id, rccs.expires_at - now() as expires_in, c.handle, c.worker_name, t.name as team_name from containers c, resource_config_check_sessions rccs, workers w left join teams t on w.team_id = t.id where w.name = c.worker_name and c.resource_config_check_session_id = rccs.id;
```

Side note: in my local testing I also configured the same pipeline under the `main` team. Not sure if that matters.

atc/db/worker_factory.go
```diff
@@ -332,13 +345,28 @@ func (f *workerFactory) FindWorkerForContainerByOwner(owner ContainerOwner) (Wor
 	}
 
 	ownerEq := sq.Eq{}
+	if teamWorkers > 0 {
+		ownerEq["t.id"] = teamID
+	}
```
vito: Maybe this would fix my comment?:

```go
	if teamWorkers > 0 {
 		ownerEq["t.id"] = teamID
 	} else {
 		ownerEq["t.id"] = nil
 	}
```

vito:
The migration timestamps will also have to be bumped again with #3138 merged.

clarafu:
All test cases
- [ ] When there is a team worker and the user belongs to the team, the check container will be placed on the team worker
- [ ] When there is a team worker with the same check container and the user belongs to the team, the check container on the team worker will be used

- [ ] When there is a team worker and the user does not belong to the team, the check container will not be created
- [ ] When there is a team worker with the same check container and the user does not belong to the team, the check container on the team worker will not be used

- [ ] When there is a team and global worker and the user belongs to the team, the check container will be placed on the team worker
- [ ] When there is a team with the same check container and a global worker and the user belongs to the team, the check container on the team worker will be used
- [ ] When there is a team worker and a global worker with the same check container and the user belongs to the team, the check container on the global worker will not be used

- [ ] When there is a team and global worker and the user does not belong to the team, the check container will be placed on the global worker
- [ ] When there is a team worker with the same check container and a global worker and the user does not belong to the team, the check container on the team worker will not be used
- [ ] When there is a team worker and a global worker with the same check container and the user does not belong to the team, the check container on the global worker will be used

- [ ] When there is a team worker and another team's worker and the user belongs to the first team, the check container will be placed onto the first team's worker
- [ ] When there is a team worker with the same check container and another team's worker the user belongs to the first team, the check container on the first team's worker will be used
- [ ] When there is a team worker and another team's worker with the same check container and the user belongs to the first team, the check container on the second team's worker will not be used

- [ ] When there is a tagged worker and the resource has matching tags, the check container will be placed onto the tagged worker
- [ ] When there is a tagged worker and the resource does not have matching tags, the check container will not be placed onto the tagged worker
- [ ] When there is a tagged worker and a global worker and the resource does not have matching tags, the check container will be placed on the global worker

- [ ] When there is a global worker and the resource has tags, the check container will not be placed onto the global worker

- [ ] When there are two matching tagged workers and one of the workers has the check container and the resource has matching tags, the check container on the tagged worker will be used

- [ ] When there is a team and tagged worker and the resource is part of the team and has matching tags, the check container will be placed onto the worker
- [ ] When there is a team and tagged worker and the resource is part of the team and does not have matching tags, the check container will not be placed onto the worker
- [ ] When there is a team and tagged worker and the resource is not part of the team and has matching tags, the check container will not be placed onto the worker

- [ ] When there is a team worker and the resource is part of the team and has tags, the check container will not be placed onto the worker

atc/db/migration/migrations/1548261635_create_table_resource_config_scopes.down.sql
```diff
@@ -0,0 +1,23 @@
+BEGIN;
+
+  ALTER TABLE resources
+    DROP COLUMN resource_config_scope_id;
+
+  ALTER TABLE resource_config_versions
+    DROP COLUMN resource_config_scope_id,
+    ADD COLUMN "resource_config_id" integer NOT NULL REFERENCES resource_configs (id) ON DELETE CASCADE,
```
vito: Upgrading to this with existing data fails because of this new column:

```
web_1     | failed to migrate database: 1 error occurred:
web_1     | 	* Migration '1548261635_create_table_resource_config_scopes.up.sql' failed: 2 errors occurred:
web_1     | 	* Transaction ALTER TABLE resource_config_versions
web_1     |     DROP COLUMN resource_config_id,
web_1     |     ADD COLUMN resource_config_scope_id integer NOT NULL REFERENCES resource_config_scopes (id) ON DELETE CASCADE,
web_1     |     ADD CONSTRAINT "resource_config_scope_id_and_version_md5_unique" UNIQUE ("resource_config_scope_id", "version_md5") failed, rolled back the migration
web_1     | 	* pq: column "resource_config_scope_id" contains null values
web_1     | 
web_1     | 
web_1     | 
web_1     | 
```

Looks like we need to deal with existing rows in `resource_config_versions`.

This only affects users upgrading from `master` since that whole table is new as of 5.0, so it should be empty when the migrations run. But maybe we should make the migration more resilient anyway, if not just for prod's sake.

What should we do here? Just truncate the table?

vito:
Acceptance:

* [ ] (#3013) Test that global resources is feature-flagged, and off by default; resources should each have their own scope.
* [ ] (#2908) Test that the `time` resource does not share version history, even if the feature flag is enabled.
* [ ] (#2951) Test behavior of global resource vrsions enabled with team-scoped workers.
    * [ ] Verified that https://github.com/concourse/concourse/pull/3133#pullrequestreview-195645273 is fixed.
    * [ ] Verified that https://github.com/concourse/concourse/pull/3133#pullrequestreview-196220484 is fixed.
    * [ ] Verified that things stabilize with one container, on whoever is keeping the `last_checked` interval up-to-date. Manually checking on the workerless team results in a new container on global workers, and manually checking on the team results in a new container on the team worker. Whichever one last checked ends up being the last container to live.
* [ ] (#3002) Test behavior with tagged resources.
    * [ ] A container should not be re-used for a tagged resource if its worker does not fit the same tags.
    * [ ] A container on a tagged worker should not be re-used for a resource that does not specify any tags.

atc/db/migration/migrations/1548261635_create_table_resource_config_scopes.down.sql
topherbullock: I think as much resiliency we can build into migrations is always good. If we ever wanted to release nightly stable version of Concourse it would be helpful, and in general migrations should be considered separately from what version they're in if only to maintain our own sanity when shipping and recommending upgrade paths.

It seems like `resource_config_versions` is a "meta" table with data generated by the scheduler, so truncating doesn't really lose anything Concourse can't regenerate later

PR #3147, opened 2019-01-24
atc, web: Adds pin commenting

### LINKED ISSUE #2875: Commenting on pin for Resource Pinning
This is related to the discussion going on in #2870 

With the introduction of resource pinning and this whole auto-pinning of paused resources on upgrades, it would be nice to have a messaging function to let users know why we pinned the resource for them, but also let users add their own annotations. This is also related to the larger epic of `annotation` stories, like #2025 

In this case, we're asking for a much simpler implementation:
* Allow users to submit a comment after they pin
* Display the comment for the duration of the pin
* When the pin is removed the comment goes away

Long term, I could see this tying into the larger story for a global commenting system

cc @Lindsayauchin

pivotal-jamie-klassen:
Fixes #2875


pivotal-jamie-klassen:
I confess we didn't think much about data integrity errors - eg I think you get a 500 if you try to comment on a resource that is not dynamically pinned, or if you try to pin a resource via fly that is already pinned.

vito:
@pivotal-jamie-klassen Looks like this has to be rebased now that I merged #2879

vito:
foo

vito:
@pivotal-jamie-klassen @xtreme-james-cooper looks like this needs to be rebased or merged, probably due to the other web refactors?

once this is up-to-date again I can review asap

vito:
awesome!

PR #3150, opened 2019-01-25
web: extract Subscription data type

pivotal-jamie-klassen:
this should simplify testing as we convert more Native/effect modules to ports. Now we have a user-defined type that we can make equality assertions about, instead of the opaque `Sub` type provided by Elm.


pivotal-jamie-klassen:
lol but we can't actually make equality assertions about Subscription yet because many of the constructors are higher-order functions.

vito:
@pivotal-jamie-klassen is this good to go? that last comment doesn't sound sure, heh

pivotal-jamie-klassen:
@vito it is! it will be a bit more gradual to pull the function parameters out of the Subscription types, but this is a good first step.

PR #3160, opened 2019-01-28
skip mod cache warming if 'replace' rule detected

vito:
otherwise this will blow up because the replacement directories aren't copied in


clarafu:
skip mod cache warming heh

PR #3161, opened 2019-01-28
ci: remove lint job

vito:
this makes more sense in the PR pipeline where we can compare to the upstream base

removing for now to unblock the pipeline since it's especially painful to unblock now that all changes to master come in through PRs


topherbullock:
Yeah, let's move this to the PR flow

Issue #3019, opened 2019-01-10
Show date in Concourse UI


evanchaoli:
## Bug Report

As the subject, Concourse UI doesn't show date, task log lines contain only time. And build info only shows like "2d 14h ago", which is hard to read especially when a job has a long history of builds.

## Steps to Reproduce

You can see this in every Concourse build.

## Expected Results

Add an option when start "concourse web", or on UI, to show date of builds.

## Version Info

* Concourse version: 4.2.2
* Deployment type (BOSH/Docker/binary): Binary
* Infrastructure/IaaS: VM
* Browser (if applicable): Any
* Did this used to work? No

marco-m:
I have to say I found this relative time like "2d 14h ago" confusing when troubleshooting, one has to do mental gymnastics to convert to an approximate absolute timestamp :-) I like this feature request.

evanchaoli:
Is there an ETA for the feature request? If you point out where, I mean which part of code, to do this change, I may contribute.

evanchaoli:
I just did a quick browse and found the change might be in https://github.com/concourse/concourse/blob/master/web/elm/src/BuildOutput.elm. I guess that should be small change to switch time to date in log lines. But unfortunately nobody from my team is familiar with ELM. Can you guys prioritize this request and get is shipped as early as possible?

vito:
> Can you guys prioritize this request and get is shipped as early as possible?

That's not how it works! 🙂 Everyone has their own feature requests, we can't simply do them all ourselves. This is an open-source project, and for the project to succeed *everyone* needs to take part when they can, not just our small team. Keep in mind how many different people use CI systems like Concourse and have their own different needs. It's impossible for one entity (company, person, whatever) to do everything for everyone as a top priority.

Please see the [contributing guide](https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md) if you need tips on getting started. That document goes over everything you need to know in order to make changes locally and test them out.

jama-pivotal:
Before jumping into this I've added the `design-snack` label to think this through a bit more. Specifically, there are a few areas where we would have to add dates:
* Dashboard thumbnail
* Build page
* In the timestamps somehow?
* Job page

Maybe we do some fancy rules, where once its past 24 hours, we switch over to the dates?

jesse-c:
> Maybe we do some fancy rules, where once its past 24 hours, we switch over to the dates?

If doing so, I'd recommend using `label`s to have the timestamp always available.

[...]

PR #3162, opened 2019-01-29
show date in build header instead of relative time

Twiknight:
This is a one line change for #3019  to make build events show date time in ISO8601 format.

I understand there might be more discussion on design side. But at least let's initiate something on the topic of date displaying.

### DIFF
--- a/web/assets/css/build.less
+++ b/web/assets/css/build.less
@@ -9,7 +9,7 @@
   span {
     /* "remove" element from layout to prevent repainting entire page */
     position: absolute;
-    width: 100px;
+    width: 180px;
     top: 0;
   }
 }
--- a/web/elm/src/BuildDuration.elm
+++ b/web/elm/src/BuildDuration.elm
@@ -28,8 +28,8 @@ view duration now =
                         -- https://github.com/elm-lang/elm-compiler/issues/1223
                         Duration.between (Date.toTime startedAt) (Date.toTime finishedAt)
                 in
-                [ labeledRelativeDate "started" now startedAt
-                , labeledRelativeDate "finished" now finishedAt
+                [ labeledVerboseDate "started" startedAt
+                , labeledVerboseDate "finished" finishedAt
                 , labeledDuration "duration" durationElmIssue
                 ]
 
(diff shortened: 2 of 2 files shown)


web/elm/src/Build/StepTree.elm
```diff
@@ -747,7 +747,7 @@ viewTimestamp hl id ( line, date ) =
         ]
         [ case date of
             Just date ->
-                Html.td [ class "timestamp", attribute "data-timestamp" (Date.Format.format "%H:%M:%S" date) ] []
```
vito: Woah, I don't think this is what we want. This will put a full timestamp including the year, month, and day on every single line of build output. Doesn't that seem like a bit much, especially on narrower screens?

These timestamps are designed for seeing where time is spent in between individual logs, and also for deep-linking into output. I really don't think you want them to have this much information. I thought this was going to be about the start/finish timestamps in the header:

screenshot from 2019-01-29 10-48-08

evanchaoli: Let me put a screen shot here:

concourseui-show-date

@vito I think YMD formatted date is not a problem to nowadays wide screen, and it more fits user's reading habit. Actually this issue was raised by our user (VMC team, the biggest and most important internal customer).

I agree the banner `started` and `finished` could be changed as well. We may help make a PR as well.

Twiknight:
@vito cancelled changes on log timestamp and modified start/finish banner as you recommended.

Following is a screenshot
screen shot 2019-01-31 at 3 29 21 pm

I left the original `formatRelativeDate` there in case someone would need it in the future.

Twiknight:
BTW, I see this message when running unit test
```
$ yarn test                                                                                                                                                                                                      ‹›
yarn run v1.13.0
$ cd web/elm && elm test
Success! Compiled 52 modules.
Successfully generated /dev/null
Success! Compiled 3 modules.
Successfully generated /Users/Shared/workspace/others/concourse/web/elm/elm-stuff/generated-code/elm-community/elm-test/elmTestOutput.js

elm-test 0.18.12
----------------

Running 801 tests. To reproduce these results, run: elm-test --fuzz 100 --seed 2010262503

failed to fetch resource: BadStatus { url = "", status = { code = 404, message = "" }, headers = Dict.fromList [], body = "" }

TEST RUN PASSED

Duration: 1479 ms
Passed:   801
Failed:   0

✨  Done in 42.33s.
```

This log 
```
failed to fetch resource: BadStatus { url = "", status = { code = 404, message = "" }, headers = Dict.fromList [], body = "" }
```
is generated by https://github.com/concourse/concourse/blob/618c8113a9beb0102dd526f1facceb8b1684994b/web/elm/src/Resource.elm#L233
Is the output side effect invoked by design?

Twiknight:
thx for the review and comments!

evanchaoli:
@vito Can you please take care of this? The change you requested has been resolved. We expect this change to be included in 5.0.

Issue #2233, opened 2018-05-23
Improved UI for manually triggered jobs


jama22:
This is a UI improvement originally proposed by user `simonjohansson` on our forums, thread here: https://discuss.concourse-ci.org/t/improved-ui-for-manually-triggered-jobs/146

Opening an issue here as a way to keep the work to be done recorded

cc @Lindsayauchin

jchesterpivotal:
I've previously pitched the idea of an "approvals resource" for this purpose. I feel like it would be a better fit into the existing model than a new kind of trigger.

simonjohansson:
@jchesterpivotal a resource would indeed be a nice way of modeling this, but it would not solve the core issue I believe, in that it would not be clear in the UI when it needs a trigger?

jchesterpivotal:
Well there you have me. But I can also argue this of any other triggering resource -- I'm wary of introducing special cases.

simonjohansson:
@jchesterpivotal 
> I'm wary of introducing special cases.

Absolutely, I totally agree. My suggestion in the first post of the discourse would apply to any job that have a `trigger: false` and where an "upstream" job have passed and/or resource have a new version

simonjohansson:
Howdie! 

Is there any update on when/if this is going to be worked on? 15 thumbs up and 5 hears seems to indicate that this is something many people want :)

simonjohansson:
Ping pong. :)

taylorsilva:
# Issue Bankruptcy

This issue is being closed as part of 2025's **Issue Bankruptcy** declaration. [See here for details](https://blog.concourse-ci.org/posts/2025-01-06-concourse-2024-review/)

If you think this issue still has merit then feel free to open a new issue and reference this one.

Issue #1707, opened 2017-10-10
Epic: spatial resource flows


vito:
# Feature Request

## What challenge are you facing?

Concourse makes it very easy to model changes over time. Resources do this for you; you point them at a given source of truth, and it'll let you know everything that happened.

Some workflows, however, can't just be modeled as change over time. Multiple branches of a repo, or pull requests to a repository, are over *space*, not *time*. They are parallel pipelines, which today are hard to manage, and impossible to correlate (you cannot fan-in).

Build matrixes are another example of wanting to run over many spaces (i.e. versions of a product). This can be done today, within a pipeline, but results in a massive pipeline with every combination explicitly configured.

## A Modest Proposal

This issue ties together #1172, #1176, and introduces the ability to have arbitrary build matrixes within one pipeline, which should dramatically improve the experience for people testing many variations/combinations.

Resources would have an additional interface to discover *spaces*. What these are depends on the resource. For the `git` resource, it would return each branch. For a resource tracking pull requests, it would return each PR number. For a resource tracking a semver'd product, it'd emit one space for each supported version line, at each granularity (i.e. `2, 2.1, 2.0, 1, 1.9, 1.8, ...`).

All versions would then be for a respective space, and Concourse would detect all versions across all spaces.

Jobs could then specify which spaces their `get` and `put` steps should run against. This would automatically "expand" into a job for each combination of spaces described by its steps. They can say `all` for all spaces (dynamically discovered), a list of static space names (e.g. `[2.0, 1.9]`), or a list of space name filters (e.g. `[feature/*]` for feature branches).

### Run against all branches of a repo, but only progressing one to the rest

In the simplest case, this makes it easy for me to have CI run against all branches in my side-project. It would just be a matter of adding `spaces: {booklit: all}` to the job:

```yaml
jobs:
- name: unit
  public: true
  spaces: {booklit: all}
  plan:
  - get: booklit
    version: every
    trigger: true
  - task: unit
    file: booklit/ci/test.yml

- name: rc
  public: true
  serial_groups: [version]
  plan:
  - get: booklit
    passed: [unit]
    trigger: true
  - get: version
    params: {pre: rc}
  - put: version
    params: {file: version/version}

- name: ship
  public: true
  serial_groups: [version]
  plan:
  - get: booklit
    passed: [rc]
  - get: final-version
    resource: version
    passed: [rc]
    params: {bump: final}
  - get: release-notes
  - put: version
    params: {file: final-version/version}
[...]

Issue #269, opened 2016-02-03
Show versions of inputs that would be used if a manual job would be triggered *now*


voelzmo:
The `promote` job in [our pipeline](https://main.bosh-ci.cf-app.com/pipelines/bosh-openstack-cpi) is triggered manually. As the fan-in of inputs and their `passed` requirements is non-trivial, I'd like concourse to help me with the work I'm currently doing in my head. Which is: figure out which versions of the inputs I would get if I hit the `+` button now.

So, ideally, somehow I would get a list of inputs as it is currently shown for running jobs as in this screenshot, for example
<img width="1321" alt="screen shot 2016-02-03 at 18 19 27" src="https://cloud.githubusercontent.com/assets/2256887/12790581/394fd6d2-caa3-11e5-942a-6039774f0314.png">

vito:
Related: https://github.com/concourse/concourse/issues/413#issuecomment-233076618 - need to think about UI/UX here and how it relates to the first to points in the linked comment.

/cc @kimeberz

jama-pivotal:
Hey there!

We're automatically closing this issue since the original poster (or another commenter) hasn't been active in the past six months. We therefore assume that this issue has been lost or resolved on its own. Close issues that remain inactive for a long period may get automatically locked.

Don't worry though; if this is in error, let us know with a comment and we'll be happy to reopen the issue.
Thanks!

cah-andrewfitzgerald:
Got here from #1172 -> #413 -> here.
This use case would still be nice for verifying release builds/other manual steps

vito:
Reopening this as it still comes up pretty frequently. I think it makes sense to tackle at some point, or at least track the need for it.

marco-m:
Would love this! :-)

voelzmo:
Beep boop! This issue is still interesting to me and apparently ("came up pretty frequently" in the past)[https://github.com/concourse/concourse/issues/269#issuecomment-339430867]. I'm not sure if that's the perfect candidate for auto-closing, but given the fact that nothing happened for more than 3 years, it might be okay.

vito:
@voelzmo I'll just re-open and mark it 'help wanted' so it stays around. I think this is one of those things that everyone agrees is a good idea, but there's not much to say about it, and it's kind of low-impact so our priorities have been elsewhere.

taylorsilva:
# Issue Bankruptcy

[...]

PR #3163, opened 2019-01-29
[POC] Super nasty rendering of jobs that needs manual triggering

simonjohansson:
Following discussion in https://discuss.concourse-ci.org/t/improved-ui-for-manually-triggered-jobs/146 and https://github.com/concourse/concourse/issues/2233 here is a naive stupid implementation of what we need.

No tests, absolutely no architectural consideration, no understanding of Elm so all logic is in the api and in the JS. Just meant as a POC to get the ball rolling.

Here is a video that shows it off with a brief explanation
https://vimeo.com/313990031


marco-m:
@simonjohansson video link is broken

vito:
Neat, this would definitely be a great feature to have!

I do have to warn ya that we've redesigned the pipeline UI for Spaces (#1707), so these changes will likely be shortlived if they're mostly in the current JS frontend.

I also noticed one odd thing in the POC - the job downstream of `b` was marked as 'needs-kicking' even though its upstream dependency on `b` was not yet satisfied. I think we'd need this to be integrated into the API somehow to communicate this more easily to the frontend.

Maybe the scheduler could mark the job as 'triggerable' somewhere here:

https://github.com/concourse/concourse/blob/b7c9f4611ca5fac7a3146f903b9d7742e786e1f5/atc/scheduler/scheduler.go#L76-L96

It could loop over the inputs, and if any of them have `FirstOccurrence` as `true` *and* none of them were marked `trigger: true`, it could mark the job as 'needs kick'. Then the API could just surface this as a boolean property in the job JSON.

This kind of work would probably still be relevant once 'spaces' ships - maybe we could start on that now and still implement something simple in the current web UI? It'll be easier to transfer it over to the new UI with most of the work done in the backend.

simonjohansson:
@marco-m thanks for the spot, updated the link :)

@vito 

>  the job downstream of b was marked as 'needs-kicking' even though its upstream dependency on b was not yet satisfied.

Implementation is full of these kind of bugs :) It would be preferable to as you say have this resolving further down the stack.

> Maybe the scheduler could mark the job as 'triggerable' somewhere here:

Thanks for the pointer, Ill see if I can do something more clever tomorrow!

marco-m:
@simonjohansson I watched the video. Very nice!
I was thinking, a test you might want to do is to validate that it behaves as expected also if a "new" version is pinned.

simonjohansson:
Is there any upcoming UI work to show what versions of a jobs resources would be used if a user would trigger it manually? Wondering if we need send more data over the wire that just `needs-kick`

vito:
@simonjohansson Nothing planned soon, but we do have another issue for precisely that: https://github.com/concourse/concourse/issues/269

The information is actually available in the code that I linked (`inputMapping`), so it could be neat to expose that through the API, too. You could even show what's new by exposing the 'FirstOccurrence` field.

simonjohansson:
@vito before moving on, there is two approaches we could take.

1. Clean up and test the POC, i.e just expose the data we need to the frontend and then do all the logic there.

2. Instead of passing along all the necessary data to the frontend just compute a value on the backend. something like:
```
// Again just POC, this needs to be tested, put in the correct place etc. etc.
func (s *Server) ListJobs(pipeline db.Pipeline) http.Handler {
		dashboard, err := pipeline.Dashboard()
		resources, _ := pipeline.Resources()
....
		graph := CreateGraph(dashboard, resources)

		for _, job := range dashboard {
			pj := present.Job(
				teamName,
				job.Job,
				job.FinishedBuild,
				job.NextBuild,
				job.TransitionBuild,
			)

			pj.NeedsKick =  NeedsAManualTrigger(job.Job.Name(), graph)
			jobs = append(jobs, pj)
		}
....
}

func NeedsAManualTrigger(jobName string, graph Graph) bool {
	//Magic
	return false
}

func CreateGraph(dashboard db.Dashboard, resources db.Resources) (graph Graph) {
	for _, resource := range resources {
		v, _, _, _ := resource.Versions(db.Page{Limit: 1})
		graph.Resources[resource.Name()] = v[0]
	}

	for _, job := range dashboard {
		lastSuccessfull, _, _ := job.Job.LastSuccessfulBuild()
		jobInputs := job.Job.Config().Inputs()
		buildInputs, _, _ := lastSuccessfull.Resources()

		graph.Jobs[job.Job.Name()] = GraphJob{
			JobInput:                  jobInputs,
			LastSuccessfulBuildInputs: buildInputs,
		}
	}

	return
}

type GraphJob struct {
	// Configured inputs for a job, contains relationships with other jobs
	// and if the build is manually triggered
	JobInput                  []atc.JobInput

	// Contains information about what versions the last successful build was run with
	LastSuccessfulBuildInputs []db.BuildInput
}

type Graph struct {
	Resources map[string]atc.ResourceVersion
	Jobs      map[string]GraphJob
}
```

We would not expose any data that we might need in the future, but then again, why would we, as we haven't started implementing those features yet :)

vito:
@simonjohansson The main problem is there's no way to know whether a job needs triggering from that part of the code, either. You could guess, which I think is fine for a proof-of-concept while figuring out the UI/UX, but I wouldn't want to ship it if it's misleading in some cases. As far as I can tell the knowledge of whether a job needs triggering is only available for a fleeting moment in the scheduler.

I think it wouldn't be too hard to just add a `bool` value on the job, which could be set in that part of the scheduler I linked. That'll be easier than exposing the actual inputs in the API - we can cross that bridge when we get to it. 🙂 You'd just need a migration to add the column to the `jobs` table and then a method to set it, called by the scheduler.

Happy to help along via PR reviews if you want to give it a go!

simonjohansson:
@vito thanks for the pointers, was a bit afraid of doing such an intrusive change, but with your blessing Ill look into it :)

simonjohansson:
@vito took some time, holiday and stuff, but Ive updated the POC with your suggestions with @stigtermichiel , please have a look ❤️

simonjohansson:
Regarding the failing ci task,
```
pr added migrations:
1550666656_add_needs_kick_to_jobs.down.sql
1550666656_add_needs_kick_to_jobs.up.sql
out of order with master.
new migrations in a pr must be strictly newer than those on master.
```

It seems to be whitespace related, if running 
```
diff -w "$expected_pr_migrations" <(sort -n "$actual_pr_migrations") >/dev/null || {
...
```

instead of 

```
diff "$expected_pr_migrations" <(sort -n "$actual_pr_migrations") >/dev/null || {
...
```
in `check-migration-order`

The test should pass

atc/db/migration/migrations/1550666656_add_needs_kick_to_jobs.up.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
```
vito: Need a corresponding `down` migration for this

atc/scheduler/scheduler.go
```diff
@@ -93,9 +93,34 @@ func (s *Scheduler) ensurePendingBuildExists(
...
+func firstOccurrenceInInputMapping(inputMapping algorithm.InputMapping) bool {
+	for _, iM := range inputMapping {
+		if iM.FirstOccurrence {
+			return true
+		}
+	}
+	return false
+}
+
+func jobIsManuallyTriggered(job db.Job) bool {
```
vito: I think might still be useful to mark something as 'needs triggering' even if it has some inputs with `trigger: true`. 🤔

atc/scheduler/scheduler.go
```diff
@@ -93,9 +93,34 @@ func (s *Scheduler) ensurePendingBuildExists(
...
+	if job.NeedsManualTrigger() != jobNeedsManualTrigger {
+		if err := job.SetNeedsManualTrigger(jobNeedsManualTrigger); err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
+func firstOccurrenceInInputMapping(inputMapping algorithm.InputMapping) bool {
```
vito: I think this could be removed with a slight refactor of the loop that goes over the inputs:

```go
var needsTrigger bool
for _, inputConfig := range job.Config().Inputs() {
	inputVersion, ok := inputMapping[inputConfig.Name]
	if !ok {
		continue
	}

	if !inputVersion.FirstOccurrence {
		continue
	}

	needsTrigger = true

	if inputConfig.Trigger {
		needsTrigger = false

		// ensure pending build exists

		break
	}
}

if job.NeedsManualTrigger() != needsTrigger {
	// set
}
```

simonjohansson: We tried this approach but in the end we both felt the readability suffered.

atc/scheduler/scheduler.go
simonjohansson: In what cases?

If we were to do that the frontend code would be more complicated, as we would have to again export the last version of which a input was running with to determine if a UI change would be necessary.

atc/db/migration/migrations/1550666656_add_needs_kick_to_jobs.up.sql
simonjohansson: Good catch 👍

atc/scheduler/scheduler.go
vito: Sometimes I configure jobs to only trigger on the inputs that are closer to my project, and leave things like non-critical external dependencies as `trigger: false`. For example, the four jobs in the second column in [our pipeline](https://ci.concourse-ci.org/teams/main/pipelines/concourse) have a non-trigger `postgres-image` input.

I think it would still be useful to surface to the user when there are new inputs available for that resource, even though the job can technically trigger on its own and pick it up. It may be something I notice shortly before shipping, and I may decide to pick up the new dependency for example.

I see this as an indicator of "if i were to trigger the job, would it have new inputs?" (#269) - whereas right now it seems to be more specifically "this job never triggers on its own, so let me know when it has new inputs to trigger". Which is actually the first time Concourse would special-case "manual-trigger-only" jobs.

simonjohansson: Ok, that seems like a nice use case.

I guess a concern would be that the UI might "flash" if doing this?

I.e 

1. Schedule iterates over jobs, one of which have one `trigger: true` and one `trigger: false` input, where the both inputs have first occurring versions.
2. The job gets updated with a `new-non-used-input` bool.
3. The ui fetches jobs data, and renderers some indicator
4. Time goes by
5. The job starts executing due to the `trigger: true` job
6. The indicator is removed since the job is now running.

So if looking at the pipeline dashboard view, you would see a job flash between states.

simonjohansson:
Updated the PR.

screenshot 2019-02-26 at 13 06 21

`hello-world` is now pink due to `every-minute` having a new version

simonjohansson:
@vito :)

vito:
@simonjohansson Sorry! Was focused on getting 5.0 out - I can take another look soon (probably today).

atc/db/job.go
```diff
@@ -51,9 +51,12 @@ type Job interface {
 	GetNextPendingBuildBySerialGroup(serialGroups []string) (Build, bool, error)
 
 	ClearTaskCache(string, string) (int64, error)
+
+	SetNewInputs(bool) error
+	NewInputs() bool
```
vito: nit: Could this be called `HasNewInputs`/`SetHasNewInputs` instead? `NewInputs` sounds like it would return the inputs. (This goes for the column name too.)

web/assets/css/_vars.less
```diff
@@ -15,6 +15,7 @@
 @base0E: #9b59b6; /* purple */
 @base0F: #8f4b2d; /* brown */
 @base0G: #3d4a58; /* 17 color palettes are 1 better than 16 */
+@base0H: #FF00EF; /* 18 color palettes are actually 1 better than 17 */
```
vito: lol

simonjohansson:
> @simonjohansson Sorry! Was focused on getting 5.0 out - I can take another look soon (probably today).

No worries, awesome to finally have the 5.0 release cut! 💯

atc/db/job.go
simonjohansson: Sure thing, Ill fix it tomorrow.

Lindsayauchin:
@simonjohansson @vito 

How about going from thin yellow halo that is static to thick pulsing yellow when the job is running?
Familiar to what is there currently and the static halo looks like it is waiting to be manually triggered. Thoughts?

Screen Shot 2019-03-11 at 5 57 43 PM

simonjohansson:
@Lindsayauchin good idea, I like it!

simonjohansson:
Renamed NewInputs ans SetNewInputs to HasNewInputs and SetHasNewInputs.

vito:
@Lindsayauchin Hmm I'm a little worried that at a glance that'll look like it's running and give me a heart attack (for jobs like `shipit`).

simonjohansson:
@vito @Lindsayauchin 

Im no designer nor CSS mastermind, hence the awful colour choice etc, but if ya'll agree on a design I can try to implement it.

vito:
@simonjohansson I feel bad leaving this PR hanging while we think about the UI, especially since it has a migration; those tend to get out-of-order when PRs hang around.

If you like, we could just merge this without the front-end changes and we take it from there. :slightly_smiling_face: I'll prioritize it highly in the UX backlog. You would just need to rename the migration so it has a newer timestamp and undo the `web/` changes. Having the API is valuable enough to start with.

If you'd like to try out another UI change, I'm still curious how my suggestion from https://github.com/concourse/concourse/pull/3163#pullrequestreview-212941878 would look (making the input lines a lighter grey/white). Seems like it'd be distinct enough that it wouldn't be confused with other states, and subtle enough to not cause annoying flickering. :man_shrugging: 

Any preference? Thanks!

simonjohansson:
Merge it :) I’ll revert the web stuff

simonjohansson:
@vito updated.

Do you want it squashed down to one commit?

vito:
@simonjohansson :turtle: Since the overall change ends up being pretty small, and it looks like the migration timestamp needs to be bumped again anyway (sorry), squashing it would be nice. :+1: 

I'll try to review this shortly after so the migration timestamp issue doesn't happen to you for a third time. Sorry about that. :slightly_smiling_face:

simonjohansson:
@vito no worries and sorry for slow resolution, was out and about.

Have a nice weekend!

PR #3168, opened 2019-01-29
hack/db: use pgcli if available

vito:
nicer ux, syntax highlighting, better completion, etc


vito:
@pivotal-jamie-klassen made both revisions!

PR #3178, opened 2019-01-30
web: consolidating urls, cleaning up deps

### LINKED ISSUE #3176: Consolidate dependencies for Elm 0.19 upgrade
In accordance with #3155 and especially @mhuangpivotal's comments, we should put all of our URL-related logic in one place where we can easily update it to 0.19, and remove any dead dependencies.

pivotal-jamie-klassen:
Fixes #3176


web/elm/src/Build.elm
```diff
@@ -107,7 +107,7 @@ init flags page =
                 , previousKeyPress = Nothing
                 , previousTriggerBuildByKey = False
                 , showHelp = False
-                , hash = flags.hash
```
xtreme-james-cooper: I'm going to admit to being kinda really steamed that we have to store this... but there's no way to pass it directly to the Build.Output, is there. :(

web/elm/src/Grid.elm
```diff
@@ -1,381 +0,0 @@
-module Grid exposing (Grid(..), MatrixCell(..), fromGraph, height, insert, toMatrix, width)
```
xtreme-james-cooper: This file doesn't seem to have to do with the main change... I assume it's just dead code removal?

web/elm/src/Routes.elm
```diff
@@ -1,84 +1,138 @@
...
 
 
 type Route
-    = Build String String String String
-    | Resource String String String
-    | Job String String String
-    | OneOffBuild String
-    | Pipeline String String
-    | Dashboard
+    = Build String String String String Highlight
```
xtreme-james-cooper: Ooooooooh, I like this change

web/elm/src/Routes.elm
```diff
@@ -89,159 +143,238 @@ buildRoute : Concourse.Build -> String
...
-    Dashboard |> toString
+    Dashboard Nothing |> toString
 
 
 dashboardHdRoute : String
 dashboardHdRoute =
     DashboardHd |> toString
 
 
+showHighlight : Highlight -> String
```
xtreme-james-cooper: ...Why is this now in routes? wouldn't it make more sense to keep it in Steptree, the only place it's used?

web/elm/src/Dashboard.elm
```diff
@@ -405,7 +405,9 @@ update msg model =
                             model
             in
             ( newModel
-            , [ FocusSearchInput, ModifyUrl (NewTopBar.queryStringFromSearch query) ]
+            , [ FocusSearchInput
+              , ModifyUrl (Routes.Dashboard (Just query) |> Routes.toString)
```
xtreme-james-cooper: Ewwwwww, |> in parens. 🤢 

`ModifyUrl (Routes.toString (Routes.Dashboard (Just query)))` or `Just query |> Routes.Dashboard |> Routes.toString |> ModifyUrl`, please.

(Or maybe make ModifyUrl take a route directly, and tostring it in runEffects?)

web/elm/src/NewTopBar.elm
```diff
@@ -50,17 +48,6 @@ query model =
             ""
 
 
-queryStringFromSearch : String -> String
```
xtreme-james-cooper: This is good, although it will bang into our newest top bar refactor

web/elm/src/Pipeline.elm
```diff
@@ -355,8 +351,22 @@ viewGroupsBar model =
...
+    { a
+        | selectedGroups : List String
+        , pipelineLocator : Concourse.PipelineIdentifier
+    }
+    -> Concourse.PipelineGroup
+    -> Html Msg
+viewGroup { selectedGroups, pipelineLocator } grp =
+    let
+        url =
+            Routes.Pipeline
```
xtreme-james-cooper: This seems... not as good as delegating to Routes.pipelineRoute like we used to.

web/elm/src/Pipeline.elm
```diff
@@ -511,52 +523,11 @@ getDefaultSelectedGroups pipeline =
...
-
-pipelineIdentifierFromModel : Model -> Maybe Concourse.PipelineIdentifier
-pipelineIdentifierFromModel model =
-    case model.pipeline of
-        RemoteData.Success pipeline ->
-            Just { teamName = pipeline.teamName, pipelineName = pipeline.name }
-
-        _ ->
-            Nothing
+    Routes.Pipeline
```
xtreme-james-cooper: Same as line 364 in this file.

web/elm/tests/BuildTests.elm
```diff
@@ -3,7 +3,7 @@ module BuildTests exposing (all)
 import Array
 import Build
 import Build.Models as Models
-import Build.Msgs as Msgs
+import Build.Msgs
```
xtreme-james-cooper: Is this the direction we want to go in general? always explicitly prefixed messages in tests?

web/elm/src/Build.elm
xtreme-james-cooper: The highlight, not the hash.

web/elm/src/Grid.elm
pivotal-jamie-klassen: yeah, we considered it part of the 'cleaning up deps'

web/elm/src/Routes.elm
pivotal-jamie-klassen: it's the inverse of `parseHighlight`, and my hope is that when we upgrade to 0.19 we will use https://package.elm-lang.org/packages/elm/url/latest/Url-Builder to construct links. It could still go where you're suggesting, but I guess it feels kinda nice for `Routes` to become the interface to the `Url` package in the future.

web/elm/tests/BuildTests.elm
pivotal-jamie-klassen: I'm still hopeful that in the long run our messages will be very flat types and there will be only one Msgs module... Are you thinking the alternative here would be to expose `Msg(..)` in the test? I don't really have a strong opinion on that one

web/elm/src/Build.elm
pivotal-jamie-klassen: I think the Right Thing to do here is merge all the `Build`-related models, potentially using extensible records to scope it in different modules... I guess we should poke around a bit more on this since we're touching it now anyway.

web/elm/src/Routes.elm
xtreme-james-cooper: Ahhh, I see. Cool.

web/elm/tests/BuildTests.elm
xtreme-james-cooper: No, just `Build.Msgs` vs `Build.Msgs as Msgs`. I think exposing would make it hard to read.

web/elm/src/Build.elm
xtreme-james-cooper: Yeah maybe. There's a bunch of them, though - it might make sense to do that as a separate, second pr.

PR #3186, opened 2019-01-31
Refactored "new top bar" -> "newest top bar"

pivotal-saman-alvi:
No features added, but many tests, and we do have feature parity


web/elm/src/NewTopBar/Styles.elm
```diff
@@ -26,79 +27,79 @@ pageHeaderHeight =
...
+                                    "stretch"
+
+                                Desktop ->
+                                    "center"
+
+                                BigDesktop ->
+                                    "center"
+
+                        Collapsed ->
+                            "flexStart"
```
pivotal-jamie-klassen: flex-start

web/wats/test/dashboard.js
```diff
@@ -168,3 +168,18 @@ test.skip('links to specific builds', async t => {
   await t.context.web.clickAndWait(`${group} .node[data-tooltip="passing"] a`, '.build-header');
   t.regex(await t.context.web.text(), /passing #1/);
 });
+
+test.only('dropdown is gone after selecting a specific status to filter', async t => {
```
pivotal-jamie-klassen: test.only prevents this from passing on CI!

web/wats/test/helpers/web.js
```diff
@@ -24,6 +24,7 @@ class Web {
     });
 
     this.page = await this.browser.newPage();
+    await this.page.setViewport({width: 1000, height: 1000})
```
pivotal-jamie-klassen: I'm OK with using this default and accepting that this means watsjs won't cover mobile resolutions by default, but puppeteer.launch takes a defaultViewport option ([docs](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions)), so it probably makes sense to set it above, rather than creating a page and resizing it every time you initialize a `Web`.

PR #3187, opened 2019-01-31
Refactor upgrade downgrade

vito:
Pushing this a bit early so I can increase the feedback loop for fixing the upgrade/downgrade tests - once this is pushed a dev image can be built and I can fix this more quickly.

This is broken info a few commits which should each be somewhat descriptive at least!


vito:
force-merging this to unbreak `concourse/dev`

PR #3188, opened 2019-02-01
Make --tsa-authorized-keys optional

tlwr:
What
---

- Makes the `--tsa-authorized-keys` flag optional, using a new function `getNumberOfWorkerSSHKeys`
- Adds tests for the `getNumberOfWorkerSSHKeys` function

Why
---

Right now if only `--tsa-team-authorized-keys` is used, `--tsa-authorized-keys` has to specified (e.g. `/dev/null`

Comments
---

I'm not sure if this conforms to the "style" of how Concourse is written as I do not have much experience of the codebase. There were no tests in `tsa/tsacmd` so I used `ginkgo bootstrap` as is present elsewhere.

Please let me know if I need to make further changes for this to be acceptable.


tsa/tsacmd/command.go
```diff
@@ -85,6 +85,10 @@ func (cmd *TSACommand) Runner(args []string) (ifrit.Runner, error) {
 		return nil, fmt.Errorf("failed to load team authorized keys: %s", err)
 	}
 
+	if cmd.getNumberOfWorkerSSHKeys(cmd.AuthorizedKeys, cmd.TeamAuthorizedKeys) == 0 {
```
vito: this could just be `len(cmd.Authorized) == 0 && len(cmd.TeamAuthorizedKeys) == 0` - don't think we need the helper function

come to think of it, should we maybe just skip this check and allow it to be started with no keys? that may be useful for e.g. debugging or temporarily disabling all keys (and still wanting to run the `web` node)

Issue #3164, opened 2019-01-29
Resource metadata can squish build output when text is too long


sohrab-:
## Bug Report

In the step (via web ui), when I expand the git resource section, the panels are splitting strangely. Note the log panel vs the commit info panel in below screenshots (half width and full width respectively).

![image](https://user-images.githubusercontent.com/4444588/51907082-f71a7780-2419-11e9-8cfe-9256aa89a820.png)

![image](https://user-images.githubusercontent.com/4444588/51907108-04cffd00-241a-11e9-8ad8-2516d1d6e782.png)

## Steps to Reproduce

1. Start Concourse via docker-compose: https://concourse-ci.org/docker-compose.yml
1. `git clone https://github.com/starkandwayne/concourse-tutorial.git`
1. `cd concourse-tutorial/tutorials/basic/pipeline-resources`
1. `fly --target tutorial login --concourse-url http://127.0.0.1:8080 -u test -p test`
1. `fly -t tutorial sp -c pipeline.yml -p hello-world`
1. `fly -t tutorial up -p hello-world`
1. Navigate to the build that ran in the web ui

Reproduced in both Chrome and Safari.

## Expected Results

For the resource section in the web UI to not look weird

## Actual Results

The resource section in the web UI _does_ look weird

## Version Info

* Concourse version: 4.2.2
* Deployment type (BOSH/Docker/binary): Docker
* Infrastructure/IaaS: Local (MacOS)
* Browser (if applicable): Chrome 71.0.3578.98 / Safari 12.0.3 (13606.4.5.3.1)
* Did this used to work? Yes, it used to not look weird

sohrab-:
Ok, it looks like it has to do with the long commit messages. Should this wrap around?

![image](https://user-images.githubusercontent.com/4444588/51908715-ae18f200-241e-11e9-93f9-6a9bd4e2531b.png)

vito:
Yep, we could probably word-wrap the metadata instead of squishing the output.

deniseyu:
Looks like this has been fixed and released in 5.0.0.

cirocosta:
Indeed! Thanks, @stigtermichiel !

<img width="1757" alt="Screen Shot 2019-05-17 at 8 35 55 AM" src="https://user-images.githubusercontent.com/3574444/57928486-de29bb00-787e-11e9-8420-8514833221ca.png">

PR #3193, opened 2019-02-01
Fix issue #3164: Squishing of build-step output

stigtermichiel:
I don't know too much about css, but I believe this fixes issue #3164 where metadata could squish the build output when too long.

screen shot 2019-02-01 at 14 33 08

Notice that `message` is aligned in the middle. Not sure if that is the desired look or not.

### DIFF
--- a/web/assets/css/production.less
+++ b/web/assets/css/production.less
@@ -233,8 +233,7 @@ li.prep-status {
 }
 
 .step-body .dictionary {
-  table-layout: fixed;
-  width: 50%;
+  max-width: 50%;
 }
 
 .dict-key {
@@ -246,6 +245,7 @@ li.prep-status {
 .dict-value {
   padding: 0;
   margin: 0;
+  word-break: break-all;
 }
 
 .resource-check-status .header {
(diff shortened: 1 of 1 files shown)


web/assets/css/production.less
```diff
@@ -232,6 +232,11 @@ li.prep-status {
   line-height: 28px;
 }
 
+.step-body .dictionary {
+  table-layout: fixed;
```
vito: What's this for? It seems to result in squishing the content a bit, since it tells the browser to keep the columns only as wide as the first row's columns, which can be short.

web/assets/css/production.less
```diff
@@ -232,6 +232,11 @@ li.prep-status {
...
+  width: 50%;
```
vito: Maybe this should be a `max-width` for cases where the metadata is actually short (no need to push the step output over).

web/assets/css/production.less
stigtermichiel: I made the changes you suggested. Much better like this.

pivotal-jamie-klassen:
Good choices - eventually as we pay down our tech debt this will probably be implemented with inline styles (rather than an external stylesheet) and covered with a unit test but it's a great quick fix.

vito:
thanks!

PR #3210, opened 2019-02-04
atc: fix syslog leaking goroutines

### LINKED ISSUE #3207: ATC syslog leaking goroutines
Related to #3140 

Syslog drainer spawns new goroutines to watch the db for build events but never closes them:
https://github.com/concourse/concourse/blob/11d1bc2c96952e06370b660d87c39d1557e136e5/atc/syslog/drainer.go#L55

chenbh:
fixes #3207 

`Build.Events()` spawns off a goroutine to do stuff with the db and needs to be closed.

### DIFF
--- a/atc/syslog/drainer.go
+++ b/atc/syslog/drainer.go
@@ -104,7 +104,7 @@ func (d *drainer) drainBuild(logger lager.Logger, build db.Build, syslog *Syslog
 
 			err = syslog.Write(d.hostname, tag, time.Unix(log.Time, 0), payload)
 			if err != nil {
-				logger.Error("failed-to-write-to-server", err, lager.Data{"tag": tag})
+				logger.Error("failed-to-write-to-server", err)
 				return err
 			}
 		}
--- a/atc/syslog/drainer.go
+++ b/atc/syslog/drainer.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"time"
 
+	"code.cloudfoundry.org/lager"
 	"code.cloudfoundry.org/lager/lagerctx"
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/atc/event"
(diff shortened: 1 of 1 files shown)


atc/syslog/drainer.go
```diff
@@ -49,51 +50,66 @@ func (d *drainer) Run(ctx context.Context) error {
...
-					err = syslog.Write(d.hostname, tag, time.Unix(log.Time, 0), payload)
-					if err != nil {
-						logger.Error("Syslog drainer sending to server error.", err)
-						return err
-					}
-				}
+func (d *drainer) drainBuild(logger lager.Logger, build db.Build, syslog *Syslog) error {
+	events, err := build.Events(0)
+	if err != nil {
+		logger.Error("Syslog drainer getting build events error.", err)
```
cirocosta: Hey, maybe instead of directly using the `logger` that gets passed as an argument, we'd be better of starting a session before so that one can more quickly jump to where the error log happens.

atc/syslog/drainer.go
```diff
@@ -49,51 +50,66 @@ func (d *drainer) Run(ctx context.Context) error {
...
+	// ignore any errors coming from events.Close()
+	defer db.Close(events)
+
+	for {
+		ev, err := events.Next()
+		if err != nil {
+			if err == db.ErrEndOfBuildEventStream {
+				break
 			}
+			logger.Error("Syslog drainer getting next event error.", err)
```
cirocosta: Making use of sessions, there's no need to prefix every log line with `Syslog drainer` given that we already have the context prefixed in the message (each `task` specified in a `session` gets appended with dots in the `message` field.

For instance, consider the following example where we have two actions being logged (`an-action` and `a-failed-action`) under the same session (as they share the same execution context):

```go
package main

import (
	"fmt"
	"os"

	"code.cloudfoundry.org/lager"
)

func main() {
	logger := lager.NewLogger("my-app")
	logger.RegisterSink(lager.NewWriterSink(os.Stdout, lager.DEBUG))

	sess := logger.Session("my-session")

	sess.Info("an-action")
	sess.Error("a-failed-action", fmt.Errorf("total failure"))
}
```

producing

```
{"timestamp":"1549320221.283070803","source":"my-app","message":"my-app.my-session.an-action","log_level":1,"data":{"session":"1"}}
{"timestamp":"1549320221.283279657","source":"my-app","message":"my-app.my-session.a-failed-action","log_level":2,"data":{"error":"total failure","session":"1"}}
```

Another detail: those actions don't have white spaces separating the fields, but instead, dashes - a pattern adopted across the whole codebase. 

See lager's [`Logger` interface](https://github.com/cloudfoundry/lager/blob/54c4f2530ddeb742ce9c9ee26f5fa758d6c1b3b6/logger.go#L16-L19) for more details on that.

Please let me know if you'd need more context!

atc/syslog/drainer.go
```diff
@@ -49,51 +50,66 @@ func (d *drainer) Run(ctx context.Context) error {
...
+			return err
+		}
 
-			err = build.SetDrained(true)
+		if ev.Event == "log" {
```
cirocosta: Wdyt of using `EventTypeLog` from `concourse/atc/event/types.go` instead of the string directly?

atc/syslog/drainer.go
```diff
@@ -49,51 +50,66 @@ func (d *drainer) Run(ctx context.Context) error {
...
-				logger.Error("Syslog drainer setting drained on build error.", err)
+				logger.Error("Syslog drainer sending to server error.", err)
 				return err
 			}
 		}
 	}
+
+	err = build.SetDrained(true)
+	if err != nil {
+		logger.Error("Syslog drainer setting drained on build error.", err)
```
cirocosta: Adding to the comments regarding prefixing the error actions, we can also evict the `error` suffix there (as the `log_level` is already telling that this is an error).

Across the codebase, I can recall the use of `failed` sometimes though (`failed-to-send-...` to be explicit that the action of "sending" something failed).

Wdyt?

atc/syslog/drainer.go
chenbh: All of the logging styles were from before so I do think this is a good time to unify it with the rest of the code base

atc/syslog/drainer.go
chenbh: The `logger` here is from https://github.com/concourse/concourse/blob/1f56328810f3b8e7f912b4da289550a231eba045/atc/syslog/drainer.go#L39
`drainBuild` is only used internally at https://github.com/concourse/concourse/blob/1f56328810f3b8e7f912b4da289550a231eba045/atc/syslog/drainer.go#L58

Since the logger is dependent on the context of the runner, I don't think we should store it in the `drainer struct` and the only option left is to pass it in as an arg

wdyt?

atc/syslog/drainer.go
vito: Yeah we should probably put logging style somewhere in our wiki.

cirocosta: @topherbullock got us covered! https://github.com/concourse/concourse/wiki/Concourse-Go-Style-Guide#logging

atc/syslog/drainer.go
```diff
@@ -39,61 +40,81 @@ func (d *drainer) Run(ctx context.Context) error {
...
 			}
 
-			err = build.SetDrained(true)
+			payload := log.Payload
+			tag := build.TeamName() + "/" + build.PipelineName() + "/" + build.JobName() + "/" + build.Name() + "/" + string(log.Origin.ID)
+
+			err = syslog.Write(d.hostname, tag, time.Unix(log.Time, 0), payload)
 			if err != nil {
-				logger.Error("Syslog drainer setting drained on build error.", err)
+				logger.Error("failed-to-write-to-server", err, lager.Data{"tag": tag})
```
cirocosta: Hey,

I have the impression that passing the `tag` there in the context of the log might end up being a duplication of what's already in the logging session (given that the session already contains pretty much all of the fields used to form the tag).

Wdyt of getting rid of that?

chenbh: Whoops, must've missed it!

PR #3218, opened 2019-02-05
WIP: keepalive worker ssh connections

vito:
note: i haven't tested this out with a failing worker yet. it would be nice to have topgun coverage for this sort of thing where we can simulate a hard shutdown. maybe we should try it out before merging?

i also snuck in a cleanup of the `.Accept` loop which is technically unrelated but the bare `return` on any error felt a bit scary in light of possible 'max open files' errors (which would previously cause the TSA to exit).

for the future: we should probably move to https://github.com/gliderlabs/ssh when we have time. once we're using that we can set idle timeout instead, which feels a *little* more intuitive (this is the only place where we have the *server* set tcpkeepalives, usually it's done via the dialer). an idle timeout should be sufficient since we can set it based on the heartbeat interval, which should keep the connection alive.

### DIFF
--- a/tsa/tsacmd/server.go
+++ b/tsa/tsacmd/server.go
@@ -83,22 +83,45 @@ func (forward ForwardedTCPIP) Wait() {
 	forward.Logger.Debug("drained")
 }
 
-func (server *server) Serve(listener net.Listener) {
+func (server *server) Serve(listener net.Listener) error {
+	var tempDelay time.Duration
 	for {
-		c, err := listener.Accept()
+		conn, err := listener.Accept()
 		if err != nil {
-			if !strings.Contains(err.Error(), "use of closed network connection") {
-				server.logger.Error("failed-to-accept", err)
+			if strings.Contains(err.Error(), "use of closed network connection") {
+				// shutting down
+				return nil
 			}
 
-			return
+			if ne, ok := err.(net.Error); ok && ne.Temporary() {
+				if tempDelay == 0 {
+					tempDelay = 5 * time.Millisecond
+				} else {
+					tempDelay *= 2
+				}
+
+				if max := 1 * time.Second; tempDelay > max {
+					tempDelay = max
+				}
+
+				server.logger.Error("failed-to-accept-temporarily", err, lager.Data{
+					"delay": tempDelay.String(),
+				})
+
+				time.Sleep(tempDelay)
+				continue
+			}
+
+			server.logger.Error("failed-to-accept", err)
+
+			return err
 		}
 
 		logger := server.logger.Session("connection", lager.Data{
-			"remote": c.RemoteAddr().String(),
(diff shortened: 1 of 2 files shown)


tsa/tsacmd/server.go
```diff
@@ -83,22 +83,57 @@ func (forward ForwardedTCPIP) Wait() {
...
 		if err != nil {
-			if !strings.Contains(err.Error(), "use of closed network connection") {
-				server.logger.Error("failed-to-accept", err)
+			if strings.Contains(err.Error(), "use of closed network connection") {
+				// shutting down
+				return nil
 			}
 
-			return
+			if ne, ok := err.(net.Error); ok && ne.Temporary() {
```
cirocosta: That's interesting, I didn't know what that had to do with the comment around the number of open files

```go
func (e Errno) Temporary() bool {
	return e == EINTR || e == EMFILE || e.Timeout()
}

//        EMFILE The per-process limit on the number of open file descriptors has been reached.
```

([syscall_unix.go](https://github.com/golang/go/blob/ccd9d9d4cef16618d84bae9cba9f1b264278c8ab/src/syscall/syscall_unix.go#L123-L125))

Neat!

tsa/tsacmd/server.go
```diff
@@ -83,22 +83,57 @@ func (forward ForwardedTCPIP) Wait() {
...
 		})
 
-		go server.handshake(logger, c)
+		if tc, ok := conn.(*net.TCPConn); ok {
+			err := tc.SetKeepAlive(true)
+			if err != nil {
+				logger.Error("failed-to-set-tcp-keepalive", err)
+			}
+
+			err = tc.SetKeepAlivePeriod(15 * time.Second)
```
cirocosta: Despite having looked at the other PR regarding keepalives yesterday, I got wondering about the number of probes before considering the connection idle. 

For those who might end up like me, check out @vito's commit message here https://github.com/concourse/concourse/commit/cac9c6d4435701757a96fb4f2a871cb4180d17d8 . tl;dr:  going with the default as Go doesn't expose a standard way of setting the number of probes.

tsa/tsacmd/server_runner.go
```diff
@@ -26,17 +26,16 @@ func (runner serverRunner) Run(signals <-chan os.Signal, ready chan<- struct{})
...
-		runner.server.Serve(listener)
+		exited <- runner.server.Serve(listener)
 	}()
 
 	for {
 		select {
-		case <-exited:
-			return nil
+		case err := <-exited:
+			return err
```
cirocosta: :+1:

vito:
@cirocosta Yeah, I think we could even write a `topgun` test for this that partitions the worker's network. I don't know what to assert exactly at this point, but the pipeline should eventually 'recover' and start using a different, healthy worker.

Issue #3003, opened 2019-01-07
Determine full set of core resources that we should bundle with Concourse


vito:
Let's re-assess which resource types should come with Concourse in 5.0+ - there are a few that are very situational and can probably be omitted.

pn-santos:
my 2c:

Generically speaking (dislaimer: I'm a minimalist), given how easy it is to add a resource that's not included, I think the resources included by default should be the ones that one would _very likely_ need to do anything useful and all resource which are related to projects _choices_ should be left as "add if needed".

For example I think `semver` should be left out. In cloud based CI/CD envs there's usually no need for semantic versioning, and semantic versioning itself is a _choice_ that not every project goes for.

Same for `s3`, it's kinda situational, some projects will have it others won't.

`pool` I find very useful, especially in scenarios where each microservice has it's pipeline which includes end-to-end tests and you want to make sure no 2 pipelines run them at the same time (we use it in non-pcf envs). And it's the only way to have "resource locks" across pipelines.

Agree with all other `✓`. Everything else (all `?` except `pool`) I think should be optional.

vito:
@pn-santos Agreed - I'd very much like to see Concourse ship with only very few "essential" resources and make `resource_types:` feel more "normal" and less like third-party/custom resources.

I think I agree with omitting `semver`, in part because we may end up reinventing the semver workflow once spaces lands, and it'll be easier to just introduce a new resource type.

I'm leaning towards omitting `pool` though since it actually is pretty situational in my experience. For example we don't make any use of it at all on our instance. It also feels a little awkward to use at times and may be in want of better Concourse primitives (i.e. some way to serialize pipeline segments), so I'm not sure about including it in core, even though it does have some perfectly fine use cases (e.g. for when you have separate Concourse clusters contending for a shared pool of resources).

So after all that we would end up with the following:

* `docker-image` and/or `registry-image` (in the future?) - because without this there's no way to use `resource_types:`
* `time` as timed triggers are pretty fundamental in other CI systems and it's cheap to include
* `git`, honestly just for convenience because in all likelihood most installations will use `git` these days

[...]

Issue #2781, opened 2018-11-12
Resource page: Responsive Web View


Lindsayauchin:
Breadcrumbs are misaligned on the resource page causing the login and resource name to crash into each other. 
![screenshot_20181106-162129](https://user-images.githubusercontent.com/27746614/48363684-85c9f480-e674-11e8-8abb-f9cd546f3cea.png)

pivotal-saman-alvi:
If you have a long pipeline name, we will also run into trouble with the pinned resources icon. Also something to take into consideration.

![screen shot 2018-11-14 at 2 57 26 pm](https://user-images.githubusercontent.com/8999128/48508888-b8fbb780-e81d-11e8-9692-4025022cdac6.png)

Issue #788, opened 2016-11-22
Resource types in a pipeline in the web UI are difficult to distinguish


jchesterpivotal:
# Feature Request

## What challenge are you facing?

When eyeballing a complex pipeline with many resources being passed from place to place, it becomes odious to add resource types (`foo-git`, `bar-s3`) to the name to indicate what is what. It's also difficult to visually inspect.

## A Modest Proposal

It would be nice if resources could publish an icon, which would be shown next to the resource name at each step in a pipeline. At a glance I could see which resources were git repos, S3 files, PivNet resources and so on.

As an intermediate step, a tooltip showing metadata about the resource, including its type and current version information, could be shown when I hover over a resource.

vito:
Related to https://github.com/concourse/atc/issues/64

jchesterpivotal:
I searched but didn't see that one. Do you want me to close this one?

vito:
I'm not inclined to close either yet; they're very close but slightly different. They both request resource type icons but the other also requests to show user-specified information (e.g. `.deb`/etc.).

I like this issue more for the resource type indicator alone, and the other one we can keep open for the extra bit (or close it if I decide it's too much).

jtarchie:
Related to https://github.com/concourse/concourse/issues/1120

jama-pivotal:
Merging in comments from #1120 and closing that one off 

We'll need to do a few things here before getting started:

- [ ] Get a sense of how those icons will be shown on the UI
- [ ] specify the preferred image dimensions and preferred formats
- [ ] define how the icons should be provided (special filename?)
- [ ] Update docs

CAFxX:
I agree, the more we build pipelines and the harder it gets to remember how each works by just looking at the dashboard. Just a small note, it may be worthwile to consider that each resource may provide configuration-dependent icons (e.g. for the cf resource I would love to have different icons for normal and b/g deploys -- and maybe tomorrow for additional ops that the cf resource may provide)

marco-m:
GoCD uses icons for resources, see for example the icons for git and hg:
![gocd-value-stream](https://user-images.githubusercontent.com/3075069/51964533-25727400-2467-11e9-9a66-388e43f25125.png)

vito:
@Lindsayauchin had some mockups in https://github.com/concourse/concourse/pull/3220 for adding tooltips. That seems like a good start?

marco-m:
[...]

PR #3220, opened 2019-02-05
Appends resource type to the name on the resource details page

### LINKED ISSUE #3169: Resource types displayed on details page
screen shot 2019-01-29 at 3 46 08 pm

It's hard to tell what resource type a resource is when looking at the resource details page.

It would be great to display a resource's current type in the web UI so all context is available. Right now, putting types in the resource name is the way you provide context to the metadata display on that page.

Heck yaw!

/cc @ddadlani

osis:
Resolves #3169

### DIFF
--- a/web/elm/src/Concourse.elm
+++ b/web/elm/src/Concourse.elm
@@ -605,6 +605,7 @@ type alias Resource =
     { teamName : String
     , pipelineName : String
     , name : String
+    , type_ : String
     , failingToCheck : Bool
     , checkError : String
     , checkSetupError : String
@@ -644,6 +645,7 @@ decodeResource =
         |: Json.Decode.field "team_name" Json.Decode.string
         |: Json.Decode.field "pipeline_name" Json.Decode.string
         |: Json.Decode.field "name" Json.Decode.string
+        |: Json.Decode.field "type" Json.Decode.string
         |: (defaultTo False <| Json.Decode.field "failing_to_check" Json.Decode.bool)
         |: (defaultTo "" <| Json.Decode.field "check_error" Json.Decode.string)
         |: (defaultTo "" <| Json.Decode.field "check_setup_error" Json.Decode.string)
--- a/web/elm/src/Resource.elm
+++ b/web/elm/src/Resource.elm
@@ -60,6 +60,7 @@ import Pinned exposing (ResourcePinState(..), VersionPinState(..))
 import Resource.Models as Models exposing (Model)
 import Resource.Msgs exposing (Msg(..))
 import Resource.Styles
+import Resource.Text
 import Routes
 import Spinner
 import StrictEvents
@@ -95,6 +96,7 @@ init flags =
                 , teamName = flags.teamName
                 , pipelineName = flags.pipelineName
                 , name = flags.resourceName
+                , type_ = ""
                 , checkStatus = Models.CheckingSuccessfully
                 , checkError = ""
                 , checkSetupError = ""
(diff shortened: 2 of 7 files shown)


osis:
Sorry, this is a WIP.

osis:
This should be stable now.

cc/ @pivotal-jamie-klassen @xtreme-james-cooper to take a peek whenever they have slack.

cirocosta:
That's pretty nice! 

I suppose not updating the `title` is desired. Is it?

screenshot from 2019-02-06 11-21-35

thx!

osis:
Naw, I didn't want to add it to the title. I figured it should match the object you interacted with to get to that page vs the details displayed on the page itself, but it's a loosely held opinion. /cc @Lindsayauchin 

Haha yea I went simple in terms of styling and Concourse is very flexible in terms of inputs. I wonder want would happen if someone had a "this-name-is-way-too-long-for-the-ui-to-handle" resource 😄.

I guess that might be handled in https://github.com/concourse/concourse/issues/2781.

Lindsayauchin:
@osis I worry about how long those titles will get! But, I like the freedom and the simplicity of the solution and it is resolving this pain for some users in a really simple way. Would this go away if we added an icon beside a resource or remain? https://github.com/concourse/concourse/issues/3169

osis:
@Lindsayauchin I'm not sure how you can introduce icons because there the amount of resource types can be infinite.

Lindsayauchin:
what if we limit the icon set to the core set of resources bundled with Concourse?

cirocosta:
Looking forward to 6.x where we might reduce significantly the number of resources we ship as core resources (see #3003), it seems like expanding that to custom resources would be a must 🤔 

It seems like having icons support would involve adding something like an `icon:` field somewhere like Helm charts do (https://github.com/helm/charts/blob/451214890c8138b48323449a414e54e6280f3b2c/stable/concourse/Chart.yaml#L5), which would mean having a new thing in the contract for resource implementors.

Would tooltips suffice? 😬

osis:
I would personally be a fan of dynamic text over static or dynamic icons.

- My brain explodes just thinking of the UX of putting base64 image in my pipeline yaml. 
- I would hope we don't evolve the design so that spirit of creating resources and the culture of plug-ability isn't treated like a second-class citizen in the UI of Concourse.

Lindsayauchin:
Tooltips would be the least disruptive to the current pipeline UI. 

Something like this?

screen shot 2019-02-07 at 2 50 38 pm

screen shot 2019-02-07 at 2 51 38 pm

osis:
I'm a fan of tooltips! I think that should be posted as an alternate solution for the pipelines view issue (#788). This is about where to put that info on the resource details page.

If you feel like would be better to put this on the pipelines view and the same information doesn't belong on the details page. Then we should chuck this PR into the garbage.

giphy

I yield to your design wisdom @Lindsayauchin 🙇

Lindsayauchin:
This tooltip could carry through to the resource page. 

screen shot 2019-02-07 at 6 07 54 pm

osis:
I think that would be great for consistency, but I also feel like a resource details page should clearly display the critical information that people need to see.

In summary, I'm going to sit on this fence. 🤷‍♂️

vito:
> My brain explodes just thinking of the UX of putting base64 image in my pipeline yaml.

Quick note: if/when we add support for this it would probably come from the resource type itself, perhaps from the `info` script in concourse/rfcs#1. I would rather not special-case core resources so having it part of the resource interface makes sense to me.

vito:
@osis I'm gonna close this for now and put https://github.com/concourse/concourse/issues/788 on our radar with the short-term goal of just adding tooltips. I can reopen if ya want.

Issue #2442, opened 2018-07-31
Allow concourse to support global secrets


StevenArmstrong:
## A Modest Proposal

Currently concourse only supports secrets scoped to a team or a pipeline as per the documentation 
E.g /concourse/main/foo 
      /concourse/team_a/mypipeline/foo
I have multiple common secrets that sit across multiple teams and I want to minimise the duplication in vault. 
It would be helpful if there was a  namespace
/concourse/global
This would allow secrets to be shared across teams if they are generic and minimise duplication that could be imported to all teams.

* Concourse version: 4.0.0
* Deployment type (BOSH/Docker/binary): binary
* Infrastructure/IaaS: Iaas
* Browser (if applicable): N/A
* Did this used to work? No

vito:
sounds like https://github.com/concourse/rfcs/issues/5 is relevant?

PR #3221, opened 2019-02-06
added shared path for vault credential lookup

hprotzek:
We have a lot of shared secrets, that we have to copy into each team folder. Also sometimes we create vault secrets during provisioning, where we are not aware of the available teams in Concourse. 

This PR adds an optional `shared-path` to the vault credential lookup. Existing behaviour of the credential lookup rules stays the same, except that if the path is configured, and credentials are neither found in `/concourse/TEAM_NAME/PIPELINE_NAME/foo_param` or `/concourse/TEAM_NAME/foo_param` it will check the shared path `/concourse/SHARED_PATH/foo_param`.

Resolves: https://github.com/concourse/concourse/issues/2442

### DIFF
--- a/atc/creds/vault/manager.go
+++ b/atc/creds/vault/manager.go
@@ -17,6 +17,7 @@ type VaultManager struct {
 	URL string `long:"url" description:"Vault server address used to access secrets."`
 
 	PathPrefix string `long:"path-prefix" default:"/concourse" description:"Path under which to namespace credential lookup."`
+	SharedPath string `long:"shared-path" description:"Path under which to lookup shared credentials."`
 
 	Cache    bool          `long:"cache" description:"Cache returned secrets for their lease duration in memory"`
 	MaxLease time.Duration `long:"max-lease" description:"If the cache is enabled, and this is set, override secrets lease duration with a maximum value"`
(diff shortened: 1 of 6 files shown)


hprotzek:
Hi, any change to get this merged?

hprotzek:
Hi, any change to get this merged?

vito:
@hprotzek Sorry for the delay, I held off on PR review for a bit to focus on getting 5.0 out the door. Now that it's out I'll get back to reviewing PRs this week. 🙂

brightzheng100:
Accessing some shared secrets across teams is a pretty common scenario and now it’s a real pain point as we have to duplicate through which introduces some operational overhead.
So should we extend and build this feature in a generic Concourse level, instead of backend credential manager specific?

Meanwhile, it would be nice if the shared path can be multiple instead of one for the sake of flexibility.

vito:
FWIW, I've started on a proposal that should eliminate the need for this: https://github.com/concourse/rfcs/pull/21

In the mean-time, if this is still a huge blocker I'm happy to review this PR and probably merge it as long as it's a small change with low security risk.

hprotzek:
@vito Thanks for the new proposal, I will look into it. Meanwhile could you review this PR, it would help us a lot if we can use it now and the change is very small and should not break any exiting config.

vito:
Merging - thanks!

gagsbd:
Hi,

Do we have similar feature for Credhub?

Thanks
Gags

PR #3225, opened 2019-02-06
User artifacts

### LINKED ISSUE #2826: Store user artifacts from fly execute in a volume
Currently when someone does a `fly execute` the API handler needs to coordinate user input streaming through the exec.Engine package. The `buidlserver` and `pipelineserver` in the below examples need to communicate through the same exec.Engine object so that they can transfer data over a channel.

https://github.com/concourse/concourse/blob/master/atc/api/buildserver/send_input.go#L61
https://github.com/concourse/concourse/blob/master/atc/api/pipelineserver/build.go#L39

We want to decouple the system by streaming user artifacts directly to a volume, and having the build plan grab those inputs from the volume.

pivotal-jwinters:
Fixes #2826


pivotal-jwinters:
FYI Before this gets merged we will need to bump the migration number.

cirocosta:
Hey @pivotal-jwinters ,

Talking about ATC's `peer-url` configuration with @topherbullock , it got us wondering whether that configuration would still be needed after this gets in.

Would it?

thx!

pivotal-jwinters:
ooo I hope so :), but to be honest I'm not sure. This PR removes the use case where the ATC needs to round robin the artifact upload to all its peers. But I don't know if thats the only usage for the peer url.

pivotal-jwinters:
It looks like its also used by the build tracker to know which host is tracking the build, but I wonder if `tracked_by` is still necessary.

pivotal-jwinters:
@cirocosta it looks like its not needed anymore. But this PR is already gigantic to lets remove it later.

topherbullock:
🔥🔥kill it with fire!🔥🔥

vito:
Superseded by #3307

PR #3243, opened 2019-02-07
Bionic golang builder image

YoussB:
As part of switching to `ubuntu:bionic` for the resources, we need an intermediate image to build `go` files (we were using `golang:alpine` before).

This adds a new job to build and push the `golang-builder` image to docker-hub under `concourse/golang-builder`

Go version is configured in `ci/dockerfiles/golang-builder/Dockerfile`

### DIFF
--- a/ci/dockerfiles/golang-builder/Dockerfile
+++ b/ci/dockerfiles/golang-builder/Dockerfile
@@ -2,26 +2,19 @@ FROM ubuntu:bionic
 
 # gcc for cgo
 RUN apt-get update && apt-get install -y --no-install-recommends \
-		g++ \
-		gcc \
-		libc6-dev \
-		make \
-		pkg-config \
-                wget \
-                git \
-                ca-certificates \
-	&& rm -rf /var/lib/apt/lists/*
+    build-essential \
+    git \
+    ca-certificates \
+  && rm -rf /var/lib/apt/lists/*
 
-ARG GOLANG_VERSION=1.11.5
+COPY ./golang-linux/go*.linux-amd64.tar.gz /go.linux-amd64.tgz
 
-RUN set -eux; \
-	url="https://golang.org/dl/go${GOLANG_VERSION}.linux-amd64.tar.gz"; \
-	wget -nv -O - "$url" | tar -C /usr/local -xz; \
-	export PATH="/usr/local/go/bin:$PATH"; \
-	go version
+RUN tar -C /usr/local -xzf /go.linux-amd64.tgz \
+  && rm /go.linux-amd64.tgz
 
 ENV GOPATH /go
 ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
+
 
 RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"
 WORKDIR $GOPATH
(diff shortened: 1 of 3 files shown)


ci/dockerfiles/golang-builder/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+                ca-certificates \
+	&& rm -rf /var/lib/apt/lists/*
+
+ENV GOLANG_VERSION 1.11.5
+
+RUN set -eux; \
+	url="https://golang.org/dl/go${GOLANG_VERSION}.linux-amd64.tar.gz"; \
+	wget -O go.tgz "$url" 2>/dev/null; \
+	tar -C /usr/local -xzf go.tgz; \
+	rm go.tgz; \
```
cirocosta: Hey,

These three lines can be done in a single one, e.g.:

```sh
curl -SL $url | tar xz -C /usr/local
```

(possible with `wget` too, I just don't recall the right flags :grin: )

ci/dockerfiles/golang-builder/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+		gcc \
+		libc6-dev \
+		make \
+		pkg-config \
+                wget \
+                git \
+                ca-certificates \
+	&& rm -rf /var/lib/apt/lists/*
+
+ENV GOLANG_VERSION 1.11.5
```
cirocosta: Hey, I think we could have some more flexibility making this an `ARG` instead.

For instance:

```
ARG GOLANG_VERSION=1.11.5
```

This would allow one to build the image for any golang version out there with `docker build --build-arg=GOLANG_VERSION=blabla`, something that we could support in `concourse/builder` in the future, allowing us to automate the build for this image based on Go releases.

Wdyt?

vito:
Hmm I'm not sure this is actually necessary - don't we only need the final stage (the one we actually package) to be based on Bionic?

chenbh:
I remember something about how an untrusted builder is capable of injecting something into the built binaries that will be put on the final stage

cc @topherbullock

topherbullock:
As I understand it, the concern is a statically built binary coming from a non-Bionic build image; that build step likely leverage libraries to do that building. Definitely a case of 'better safe than unsure', so we want to ensure we're using libraries which are secure and properly licensed.

ci/dockerfiles/golang-builder/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+		gcc \
+		libc6-dev \
+		make \
+		pkg-config \
+                wget \
+                git \
+                ca-certificates \
+	&& rm -rf /var/lib/apt/lists/*
+
+ARG GOLANG_VERSION=1.11.5
```
cirocosta:
```suggestion
ARG GOLANG_VERSION=1.12
```

vito:
@cirocosta Agreed - in fact maybe we should have it come in as a resource so we'll automatically bump it when new Go versions come out? That could help with #3479.

cirocosta:
@vito Totally! Could use a resource like [golang-windows](https://ci.concourse-ci.org/teams/main/pipelines/concourse/resources/golang-windows) 🤔

PR #3256, opened 2019-02-08
Prevent postgres deadlocking

clarafu:
When we update the cache indexes for the all the pipelines that use a
resource config which we found new versions for, postgres can get into a
deadlock when two resource configs try to update the same pipeline's
cache indexes at the same time. By running all the cache index updates
by itself (not in a transaction), we will prevent that from happening.

### DIFF
--- a/atc/db/resource_config_scope.go
+++ b/atc/db/resource_config_scope.go
@@ -91,12 +91,17 @@ func (r *resourceConfigScope) SaveVersions(versions []atc.Version) error {
 		}
 	}
 
-	err = bumpCacheIndexForPipelinesUsingResourceConfigScope(tx, r)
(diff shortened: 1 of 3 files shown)


atc/db/resource_config_scope.go
```diff
@@ -269,13 +274,39 @@ func incrementCheckOrder(tx Tx, r ResourceConfigScope, version string) error {
...
+		RunWith(conn).
+		Query()
+	if err != nil {
+		return err
+	}
+
+	var pipelines []int
+	for rows.Next() {
+		var pid int
+		err = rows.Scan(&pid)
```
vito: missing `err` check here

Issue #3241, opened 2019-02-07
Upgrading to 5.0.0 causes resource type checking become unrecoverable


clarafu:
After upgrading from 4.2.2 to 5.0.0, resources that use a custom resource type fail to check with `volume '<volume-id>' disappeared from worker <worker-handle>'`. These resource checks were never able to recover by itself. We observed that after we recreated the workers and manually triggered the checks, it seemed to check successfully. Not sure if we manually triggered the check before recreating, if that would've fixed it too? Even after recreating the workers, the resources were not able to recover on their own.

This was the pipeline that was used [link](https://github.com/concourse/drills/blob/master/longevity/shared-check-containers-across-teams/pipelines/check-containers-stress-test.yml)

vito:
Note: this only happens when the workers are upgraded without being re-created. A re-create is necessary for the deploying the 5.0 BOSH release because now that it uses the binary, it configures a `--work-dir` which is different from the old volumes location, so the old volumes aren't there.

This acutely affected resource types however, which is interesting - the rest of the resources all recovered (via #2588 which is new in 5.0). I won't block shipping 5.0 on this, though, because it still recovers better than in previous versions. We're going to document in the release notes that a worker recreate is required for BOSH users.

clarafu:
For now, a possible solution that we will spike on is having resource type radar scanning to not loop on acquiring the lock and just have it exit when it fails to acquire (like how it worked before). This solution is possible because we still share resource type versions even when global resources is disabled, so there is no real point to needing to require similar resource types between pipelines to fetch it's own versions.

vito:
> This acutely affected resource types however, which is interesting - the rest of the resources all recovered (via #2588 which is new in 5.0). I won't block shipping 5.0 on this, though, because it still recovers better than in previous versions. We're going to document in the release notes that a worker recreate is required for BOSH users.

Turns out this was wrong and it was just the thing @clarafu found - this will be in v5.0, then!

What's cool is after her fix was deployed everything recovered gracefully within a few minutes. So our release note won't even have to sound very dire.

cirocosta:
> a possible solution that we will spike on is having resource type radar scanning to not loop on acquiring the lock and just have it exit when it fails to acquire (like how it worked before).

Would this possibly remediate https://github.com/concourse/concourse/issues/3249?

vito:
fixed by #3275

PR #3275, opened 2019-02-12
Resource type scanner does not loop on acquiring lock

clarafu:
Instead of manually running a resource type scan within the resource
scanner when the resource uses a custom type with not versions, it will
wait on the resource type to receive a version from it's own resource
type scanning interval. This fixes the problem when many resources use
one resource type that has no versions yet, it will start off many scans
which all try to acquire the same lock. Resource type scanning also now
differs from the resource scanner because it does not loop if it fails
to acquire the lock, it will just exit with the error. We are able to do
this because even if global resources is disabled, resource type
versions are still shared within a concourse deployment so there is no
real need for them to pull in their own versions for each pipeline
resource.

[#3241]

### DIFF
--- a/atc/pipelines/radar_scheduler_factory.go
+++ b/atc/pipelines/radar_scheduler_factory.go
@@ -54,16 +54,6 @@ func (rsf *radarSchedulerFactory) BuildScanRunnerFactory(dbPipeline db.Pipeline,
 
 func (rsf *radarSchedulerFactory) BuildScheduler(pipeline db.Pipeline, externalURL string, variables creds.Variables) scheduler.BuildScheduler {
 
-	resourceTypeScanner := radar.NewResourceTypeScanner(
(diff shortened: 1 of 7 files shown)


atc/pipelines/radar_scheduler_factory.go
```diff
@@ -54,16 +54,6 @@ func (rsf *radarSchedulerFactory) BuildScanRunnerFactory(dbPipeline db.Pipeline,
 
 func (rsf *radarSchedulerFactory) BuildScheduler(pipeline db.Pipeline, externalURL string, variables creds.Variables) scheduler.BuildScheduler {
 
-	resourceTypeScanner := radar.NewResourceTypeScanner(
```
vito: yessss

topherbullock: yas

Issue #3184, opened 2019-01-31
Paused topbar UI


clarafu:
When I go to a pipeline that is paused, I see a black line between the user I am logged in as and the rest of the topbar which is all blue. Also, the pin icon is still in black which might seem a little out of place with all the other text and icons in the topbar being white.

<img width="1549" alt="screen shot 2019-01-31 at 11 23 00 am" src="https://user-images.githubusercontent.com/19290703/52068411-9e96c780-254a-11e9-92af-e9115d26c162.png">

Maybe we can remove that separation line or make it white as well, and also make the pin icon to be white when the pipeline is paused? 🤔 

Thanks!

jama22:
cc @Lindsayauchin

pivotal-saman-alvi:
It also highlights the wrong colour: 

![screen shot 2019-02-07 at 12 22 08 pm](https://user-images.githubusercontent.com/8999128/52430023-2a709c80-2ad3-11e9-8161-322a61d89701.png)

Lindsayauchin:
## Updates to the pin based on feedback on the styling. 

Can we update the default state for the pin to be white app wide? The dark grey seems to be not so successful and the feedback we are getting suggests we should iterate. 

![screen shot 2019-02-07 at 1 55 58 pm](https://user-images.githubusercontent.com/27746614/52435583-3878ea00-2ae0-11e9-936d-4db8c93e0cb0.png)

On a paused pipeline: 
![screen shot 2019-02-07 at 1 49 58 pm](https://user-images.githubusercontent.com/27746614/52435223-47ab6800-2adf-11e9-8c83-732c76a6af33.png)

The line separating the login from the pin should be almost white when the pipeline is paused this is the case for paused state on a pipeline view and when a job is paused. 

Can we also update the line that separates the login from the pin notificatio that is currently dark grey so we can dim down some of the contrast we have created in this area?

Please use #fff8, to add some alpha as the pure white is a bit too bright

![screen shot 2019-02-07 at 2 01 10 pm](https://user-images.githubusercontent.com/27746614/52435891-ec7a7500-2ae0-11e9-82d5-f4e6b086ff88.png)

## Hover State

While we are in this element it would be great to also update the hover state while we are there to dim down the contrast that we have now. 

Please use a white oval with alpha of 30% to update the style. 

[...]

PR #3279, opened 2019-02-12
web: fixed paused top bar colouration on pin and on login separator

xtreme-james-cooper:
Fixes issue #3184


pivotal-jamie-klassen:
the word 'issue' between 'fixes' and '#3184' meant that merging this didn't close the relevant issue!

Issue #2341, opened 2018-07-02
Maximum Build Time


kmacoskey:
# Feature Request

## What challenge are you facing?

It is a fairly common occurrence that builds become stuck/hung in some way within the specific context of what it is running. The execution hanging isn't specifically a concourse issue, but what is an issue is that these hung builds get ignored and then left running for days and sometimes weeks. This does cause problems when trying to drain workload on concourse VM's and it can potentially cause database issues if the hung build includes continuously emitting build output. 

## A Modest Proposal

It is unlikely that we could ever be able to avoid task execution from having the possibility of hanging or avoid the possibility of concourse users themselves failing to notice a long running build and stopping it. Therefore I propose addition of a configuration option to specify a maximum build time. This configuration option would allow us to ensure builds don't run longer then a few days.

marco-m:
@timsanwald the difference is that what proposed @kmacoskey is global, you set it once for the ATC command-line / env vars and all tasks in all pipelines don't have to be updated and cannot escape that global limit.

kmacoskey:
There is no way to enforce it that optional field being specified, nor a maximum we can specify per a given concourse deployment. Additionally I strongly think it should be unnecessary to need to add that field to every task of every pipeline in a deployment.

jama-pivotal:
Linking to #2249 as a part of operator visibility

matthughes:
Would love this as well.  For various reasons, we've had our Concourse or build have some spurious issue that causes builds to be stuck for hours.  I know our build should never take more than 30 minutes (worst possible case).  I want to consider that build failed if it takes longer.  Otherwise, there is no notification to the user.  I just now found out today our build has been stuck for two days because I rely on Slack notifications to inform me of build fails/successes.

jamesgeddes:
Has there been any progress on this?

taylorsilva:
#7426 will close this issue. It adds a configurable global timeout for get/put/task steps

PR #3283, opened 2019-02-13
atc: add global step timeout

christophermancini:
We have the shared the same pain as @kmacoskey in #2341, with tasks that can go unnoticed and running for days / weeks. I have created reports to help us catch jobs that have been running longer than 2hrs but I would prefer to have Concourse kill these tasks themselves.

This PR potentially resolves this issue by allowing the operator to set a `GlobalStepTimeout` on the web node, which defaults to `0s`.


kmacoskey:
+1 I fully support this. It is still an issue we have.

christophermancini:
I still need to finish the tests which I will be able to circle back to within the next few days.

christophermancini:
@vito I haven't been able to circle back to it as I am up against a big deadline to move all of our servers off of trusty to bionic. I should be done with that work within the next week or two and then I can finish up the tests, which is creating a test case in `testflight`.

vito:
@christophermancini Sounds good. FWIW I don't think we need a testflight test for this, since it depends on operator configuration and Testflight doesn't have any control over that. A TopGun test, which you've already added, seems appropriate. :+1:

christophermancini:
@vito oh, in that case, then this is good to go! minus the conflict that now exists. :(

christophermancini:
@vito resolved the conflict, you are free to review

xtremerui:
Hi @christophermancini are you still working on this PR? Thanks!

kirillbilchenko:
@vito @xtremerui @christophermancini are you guys planning to address this pr in any way, I think it's really useful property, which can help stabilise execution.

christophermancini:
@xtremerui @kirillbilchenko I intend to tackle this if no one gets to it first. Just have some higher priority things to attend to first. I should be able to get to it sometime in Q2.

clarafu:
@christophermancini Hi! I'd like to close this PR for now because it has been almost a year with no activity. Please feel free to open a new PR if you ever get the chance to tackle this, thanks!

PR #3286, opened 2019-02-13
web: fix layout of resource boxes on pipeline view

### LINKED ISSUE #3224: Pinned Resource Hover Style
When a resource has a pinned version, hovering over it in the pipeline view removes the padding between the pin icon and the text.

resource

* Concourse version: dev
* Infrastructure/IaaS: local
* Did this used to work? N/A

pivotal-jamie-klassen:
fixes #3224

### DIFF
--- a/web/assets/css/production.less
+++ b/web/assets/css/production.less
@@ -339,12 +339,8 @@ svg.pipeline-graph {
     stroke-width: 4px;
   }
 
-  .active.node rect {
-    filter: url(#embiggen);
-  }
-
   .active.node image {
-    transform: scale(1.09) translate(-5px, -1px);
+    transform: scale(1.06);
   }
 
   .active.node text {
--- a/web/public/index.js
+++ b/web/public/index.js
@@ -38,6 +38,50 @@ function redrawFunction(svg, jobs, resources, newUrl) {
       }
     })
 
+    function layoutGraphNode(graphNode, svgNode, isHighlighted) {
+      svgNode.classed({
+        active: isHighlighted
+      })
+      var scaleFactor = isHighlighted ? 1.06 : 1.00;
+
+      var padding = graphNode.padding() * scaleFactor;
+      var textNode = svgNode.select("text").node();
+      var imageNode = svgNode.select("image").node();
+      var height = graphNode.height() * scaleFactor;
+
+      var width;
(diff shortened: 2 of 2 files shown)


pivotal-jamie-klassen:
Trying to do this the Right Way on the `js-unit-tests` branch and realizing that it is pretty tricky to 'observe' (via DOM or SVG query methods, i.e. [getBBox](https://developer.mozilla.org/en-US/docs/Web/API/SVGGraphicsElement), [clientWidth](https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth), [scrollWidth](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollWidth), [getBoundingClientRect(...).width](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) or [window.getComputedStyle(...).width](https://developer.mozilla.org/en-US/docs/Web/API/Window/getComputedStyle)) the effect of an SVG filter (`dilate` in our case, via the [`#embiggen` element](https://github.com/concourse/concourse/blob/73221b3d902c1e5ed13296784163e93cc7129b1b/web/public/index.js#L268-L272)): long story short, almost all of the existing methods will only query the dimensions of an element _independent_ of its applied filters/transforms. This seems like a [reasonably well-established problem](https://stackoverflow.com/questions/9847684/get-the-real-size-of-a-svg-g-element) and while there may be tricks by which you can compute the resulting dimensions of an SVG element after it has a filter applied, it seems likely to be hard to reason about. From a pure TDD perspective this would discourage using filters at all, but maybe there's a way to think about the layout such that the `dilate` filter can be used alongside the required size/position changes for the text and pin icon. For now, I think we can agree that this is a magic number smell:

https://github.com/concourse/concourse/blob/master/web/assets/css/production.less#L347

vito:
@pivotal-jamie-klassen fyi this is kind of a small polish thing so I'm gonna take it out of the 5.0 milestone.

vito:
Just came back and took a look at this and it seems pretty broken (probably due to recent changes for resource icons). Do we still need this? It seems like the resource boxes are rendering OK now.

pivotal-jamie-klassen:
I'm fine closing this now. There's a better effort on the `js-unit-tests` branch anyway, which I might come back to at some point.

Issue #3251, opened 2019-02-08
Check containers accumulate on a single worker with the 'fewest-build-containers'


wagdav:
## Bug Report

We updated to Concourse 5.0 RC54 we see an accumulation of check containers.  The containers are placed to single workers until they reach 256 and pipelines start to fail with the error message:

```
insufficient subnets remaining in the pool
```

Following a proposed workaround in #847 we increased the number of available subnets by providing the `--network-pool=172.16.0.0/23` argument to Garden. This made the number of containers reach as high as ~350 on a single worker.

After a certain amount of time new error started to appear:
```
* iptables: create-instance-chains: iptables: Memory allocation problem.
```
The number of containers on a single worker went up to ~600. This error message was also reported in #3127 .

## Steps to Reproduce

Enable the "fewest-build-container" strategy and set pipelines with a lot of resource checks.  The number of build tasks or generated load by the task do not seem to directly influence the issue.

The number of pipelines required to trigger this issue depends on the available Linux workers.  We have ~100 pipelines and about ~20 Linux workers.

## Expected Results

Check containers distributed such that they don't kill workers.

## Actual Results

Check containers gravitate toward a specific worker until all its resources are exhausted.

## Possible explanation

The new "fewest-build-containers" strategy only applies to task containers, as expected. It seems that check containers fall back to Concourse's default behavior "volume locality", so they accumulate without limits on one or two workers.

A possible solution would be to use the random placement strategy for check containers. We've seen this working well 4.x using the random placement strategy. The check containers were evenly distributed among all the workers. 

## Additional Context

![containers](https://user-images.githubusercontent.com/514775/52484430-9841c480-2bb6-11e9-9d8e-e27ed170fffd.png)

This graph shows the number of containers for all of our workers as a function of time.  We can see  three distinct time periods:
* until 8:30 we have Concourse 4.2.1 with the random container placement strategy
* between 8:30 and 12:00 we have Concourse 5.0 RC54 with fewest-build-containers strategy.  Garden limits the number of containers at 256 and "insufficient subnets remaining in the pool" error messages are generated
* from 12:00 we have still Concourse 5.0 but we allow more subnets by providing `--network-pool=172.16.0.0/23`.  This way even more containers are placed on the worker and eventually it runs out of memory.

We can also see that the number of containers increase and decrease almost instantaneously. 

## Impact
Currently we are out of ideas on how to control this behavior and keep the system operational. We would be happy to provide more information that could help fixing this issue.

## Version Info

[...]

PR #3289, opened 2019-02-13
temporarily disable fewest-build-containers

vito:
it's causing a ruckus because of #3251


marco-m:
@vito is it time to bring the fewest-build-container back ? (Just in case you forgot)

vito:
@marco-m yep! opened a PR to revert it

PR #3291, opened 2019-02-13
Permanent unique version history for base types

clarafu:
If one worker registers with a unique base resource type, that base type gets persisted as unique permanatly.

### DIFF
--- a/atc/db/base_resource_type.go
+++ b/atc/db/base_resource_type.go
@@ -73,7 +73,7 @@ func (brt BaseResourceType) create(tx Tx, unique bool) (*UsedBaseResourceType, e
 		Suffix(`
 			ON CONFLICT (name) DO UPDATE SET
 				name = EXCLUDED.name,
-				unique_version_history = EXCLUDED.unique_version_history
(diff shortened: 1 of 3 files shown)


aledeganopix4d:
Hello, I've a question about this PR.
I've noted that the Worker version has been reset to **2.1** as it is in Concourse 4.2 for instance.
Does this means that the code change in https://github.com/concourse/concourse/pull/3291/files#diff-89e16b37f9593cad181ebae8aa00f73eR76 was everything needed to make the workers pre-5.0 compatible with 5.0?
Thank you!

vito:
@aledeganopix4d Yeah, exactly. We had bumped the worker version to 2.2 with the goal of allowing only new workers to register, so there wouldn't be competing values for whether it's unique or not. But bumping the worker version didn't actually help, since it doesn't prevent registration, it only prevents Concourse from using it.

So by having the value never go from `true` to `false` instead, we're able to bring the version back to 2.1, and existing 4.x workers should remain compatible (though in general you should always upgrade them to the latest anyway).

Issue #3005, opened 2019-01-07
Bring back bold fonts on parts of the Concourse web UI


jama-pivotal:
This is related to the conclusion of #2938 

Bring back some of the boldness of the fonts in the build logs and the resource page. 

In my sketch mocks, I am using an Inconsolata bold so if we could update a couple spots with some bold style to Inconsolata I think we are there!

Can we add `font-weight: 700;` ( or however you decide to implement bold style) to the body font on the following?:

### Pipeline view
_screenshot without style_ 

![screen shot 2019-01-07 at 11 36 53 am](https://user-images.githubusercontent.com/27746614/50780835-f276ff00-1271-11e9-9559-e456cad1b8ae.png)

_with   ` font-weight: 700;`_
![screen shot 2019-01-07 at 11 37 29 am](https://user-images.githubusercontent.com/27746614/50780789-d83d2100-1271-11e9-881c-f7ad1dc90ba8.png)

### Breadcrumbs
_screenshot without style_ 

![screen shot 2019-01-07 at 11 40 03 am](https://user-images.githubusercontent.com/27746614/50780468-fe15f600-1270-11e9-9e76-fa2be233dcc2.png)

_with   ` font-weight: 700;`_

![screen shot 2019-01-07 at 11 39 55 am](https://user-images.githubusercontent.com/27746614/50780475-02421380-1271-11e9-9574-f433bfd9c0c8.png)

### Resource page

![screen shot 2019-01-07 at 11 30 09 am](https://user-images.githubusercontent.com/27746614/50779913-a0cd7500-126f-11e9-9c5b-15f5641f3ef7.png)
_L screenshot without style  R screenshot with `font-weight: 700;`_

### Build page logs
This is also realted to https://github.com/concourse/concourse/issues/2765

On the Build page, the log output texts need a bit more contrast with some bold added to the font style. Adding the bold to the style attribute of `timestamped-content` td (in addition to adding it to the body) will help  improve the readability of the logs. 

![screen shot 2019-01-07 at 10 55 28 am](https://user-images.githubusercontent.com/27746614/50779750-23a20000-126f-11e9-94b5-d6f92eae0cfb.png)

_L screenshot without style  R screenshot with `font-weight: 700;`_

pivotal-jamie-klassen:
[...]

PR #3294, opened 2019-02-13
web: fix styles on group bar

### LINKED ISSUE #3285: groups bar on pipeline page is broken
screen shot 2019-02-13 at 12 02 22 pm

This broke from the mighty top bar refactors.

pivotal-jamie-klassen:
fixes #3285

### DIFF
--- a/web/assets/css/layout.less
+++ b/web/assets/css/layout.less
@@ -21,71 +21,6 @@ body,
   display: flex;
   flex-direction: column;
 
-  .top-bar {
-    position: fixed;
-    top: 0;
-    height: @top-bar-height;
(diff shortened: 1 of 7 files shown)


pivotal-jamie-klassen:
@vito I'm pretty sure the bold was added as part of #3005, which might have never been deployed to wings? As for the size change at first we thought it _was_ too big, but @Lindsayauchin didn't seem to mind - not sure what the opinion is now.

Lindsayauchin:
my mistake, there should be some distinction of hierarchy from the pipeline name to groups. If the pipeline name is 18px can we update the font in the groups label to be 14 px?

Issue #2438, opened 2018-07-30
Pipeline scheduling should never block on resource checks


vito:
# Feature Request

## What challenge are you facing?

The root cause of "my pipeline stopped scheduling" is pretty often to do with resource checks that the scheduler performs for pending builds that don't have pinned inputs.

## A Modest Proposal

We should somehow decouple pipeline scheduling from resource checking. Pipeline scheduling is supposed to be cheap and able to happen multiple times a minute. Doing checks in the middle increases risk of hanging and just generally slows things down.

pivotal-jwinters:
related to #1954

clarafu:
After some discussion, we will remove the force resource checks from the manually triggered builds and instead separate the responsibilities of the manually trigger button on the job into two separate buttons:
- a button that will force a check for all your resources that are inputs to your job
- another button that will manually trigger a new build with the current versions of each resource

This issue will be blocked on #3403.

cirocosta:
Hey @clarafu , do we also plan to have a separate flag for `check-resource`?

vito:
Hmm I'd like to think about this a bit more. Having a button that triggers without checking feels a little wishy washy - I could see it being misinterpreted as either 'trigger with newest available versions' (today's behavior) or 'retrigger THIS BUILD with same versions' (#413), but in reality it does neither.

What if we had the check button short-circuit the scheduling loop, so that if there *are* new versions, a new build will be queued up promptly, and then just went ahead and finally did #413 for the second button?

vito:
@cirocosta Also yeah, we'd need to think about that command - maybe even split it up so there's no chance for ambiguity. Something like `fly retrigger-build` and `fly check-and-schedule-job` (terrible name - I feel like we need a verb here for the 'check and schedule soonish' action).

vito:
Ok, after talking about this in IPM here's the idea:

* Introduce a 'sync job' button (to be named) that does the following:
    * Run a `check` for all resources that go into the job
    * After the `check`s complete, hits an API endpoint to "kick" scheduling for the job.
      * We'll need to think about this in the API, I think.
    * ...Somehow waits for a new build, which may or may not happen if the scheduler determines there's no new build to trigger.
      * We'll need to think about this in the API, I think.

* Do #413 

...we're running low on time so I'll fill in more details later.

vito:
Ok. So. That all led to a whole build page redesign.

Here's a short(er)-term plan for getting checking removed from the scheduler:

[...]

Issue #1585, opened 2017-09-12
Aborting a started build prior to an ATC tracking it will result in an orphaned, uncompleted build, still running


vito:
# Bug Report

I think [this block of code](https://github.com/concourse/atc/blob/e0434a2d057a261b17d654dbd8b110ba3b60a97c/engine/db_engine.go#L162-L170) does basically nothing, and hasn't since its inception.

In the case where a build is tracked and then the ATC is restarted, it's possible that a build is aborted prior to any ATC re-tracking it. This is easier to reproduce with two ATCs by doing the following:

* Start the build, make note of the ATC that's tracking it.
* Wait for the other ATC's tracking to fire (this is on a 10 second interval).
* Stop the ATC that's tracking it.
* Abort the build via the remaining ATC.

Right now, the `engineBuild.Abort` line will just write a signal to a channel that no one's reading from. To have gotten that far in the method, no one will have been tracking the build (the first thing it does is acquire a lock). In addition, no one will *ever* track the build by then, since after it acquires the lock it sets the status to `aborted`, excluding it from build tracking (which only looks at `started`).

This will result in a build's containers never actually being interrupted, and a build that is in status `aborted` but never marked `completed` in the database. This will result in the build's containers leaking, even, as we only mark `completed` builds non-interceptible.

vito:
Once this is done we can take advantage of `completed` being consistent by re-optimizing the pipeline dashboard query to look for `completed` rather than `status: [pending, started]`.

vito:
Here's an approach to fix this (discussed with @pivotal-jwinters):

* Instead of setting the build status, add an 'aborted' column and watch for that condition to become true, instead of watching for the status. This will let the build tracker still find aborted builds and ensure they all abort properly. The condition notifier already checks for this status prior to listening, so missed notifications should be fine.
* Migrate all aborted builds to set `completed` to `true`. This will allow us to just look for `completed` instead of 'started or pending'.

PR #3307, opened 2019-02-14
The Great Decoupling

### LINKED ISSUE #2826: Store user artifacts from fly execute in a volume

### LINKED ISSUE #2895: Scheduler race condition resulting in skipped versions (mainly with `version: every`)
## Bug Report

This `testflight` test fails, if you run it enough times:

https://github.com/concourse/concourse/blob/c3896f1ecd34d1b340d876997c4e4c46b75c9c29/testflight/resource_version_test.go#L45

Here's a CI build showing the failure: https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/testflight/builds/273

## Steps to Reproduce

```sh
cd concourse
docker-compose up --build
cd testflight
ginkgo -v -untilItFails -focus='runs builds with every version'
```

## Expected Results

The test should succeed reliably.

## Actual Results

I've seen it fail in under 100 runs locally. I also hacked the `AfterEach` to leave the pipeline around rather than destroying it, and when investigating I found some interesting results.

Build 1:

screencapture-localhost-8080-teams-testflight-pipelines-tf-pipeline-1-bc653c72-d5e4-406f-5ea5-b8c82697ca6b-jobs-some-passing-job-builds-1-2018-11-30-14_39_27

Build 2:

screencapture-localhost-8080-teams-testflight-pipelines-tf-pipeline-1-bc653c72-d5e4-406f-5ea5-b8c82697ca6b-jobs-some-passing-job-builds-2-2018-11-30-14_39_40

Build 3:

screencapture-localhost-8080-teams-testflight-pipelines-tf-pipeline-1-bc653c72-d5e4-406f-5ea5-b8c82697ca6b-jobs-some-passing-job-builds-3-2018-11-30-14_39_51

This shows that build 1 ran with `-guid1`, build 2 ran with `-guid2`, but then build 3 ran with...`-guid4`!

**HOWEVER**, if you look at the resource version history, it tells a different story!

screencapture-localhost-8080-teams-testflight-pipelines-tf-pipeline-1-bc653c72-d5e4-406f-5ea5-b8c82697ca6b-resources-some-resource-2018-11-30-14_40_22

This indicates that build 2 actually recorded `-guid3` as its input! This, I think, is where the bug begins; with `-guid3` recorded as an input, it makes sense for the third build to have "skipped" ahead to `-guid4`.

So why was `-guid3` recorded as an input when the build actually ran with `-guid2`?

I dunno. Figure it out!

## Version Info

* Concourse version: `c3896f1ecd34d1b340d876997c4e4c46b75c9c29`
* Infrastructure/IaaS: local
* Did this used to work? not sure, wouldn't be surprised if this has been a bug for a while

### LINKED ISSUE #3240: Remove API dependency on EXEC and Scheduler

concourse/rfcs#14

Once #3225 has been merged we can remove the ACTUAL dependency on the exec/scheduler packages. #3225 was just the plumbing work. The real benefit is being able to have our API handlers throw builds into the DB without having to `engineBuild.Resume` or `scheduler.TriggerImmediately`. Then let the `scheduler`/`build starter` take care of actually running them during the next tick.

- [x] `buildserver.CreateBuild` should no longer `enginebuild.Resume`. Instead just throw a started build into the db using the plan provided.

- [x] `pipelineserver.CreateBuild` is basically the same as above.

- [x] `jobserver.CreateJobBuild` should no longer `scheduler.TriggerImmediately`

- [x] `buildserver.AbortBuild` should no longer `enginebuild.Abort`. Instead we can just `dbbuild.MarkAsAborted` and a message will be sent to abort the build.

- [x] `jobserver.ListJobInputs` still makes a call to `scheduler.SaveInputMappings`. We should remove this side effect and just let it use whatever exists in `next_build_inputs`.

- [x] Success! The `scheduler.BuildScheduler` interface no longer needs `TriggerImmediately` or `SaveNextInputMapping`. Remove them.

- [x] Also remove the `scheduler`/`exec` dependencies in `atccmd.Command` that get injected into the api handler.

Edit:

- [x] Fix the hardcoded `exec.v2`
- [x] Change `engine_metadata` to `private_plan`?

### LINKED ISSUE #3267: Remove peer url

Once #3225 gets merged, we no longer need the `peer url`.

 Remove it.

pivotal-jwinters:
This PR is based off #3225 (which fixes #2826)

It also fixes #3240 

...and fixes #3267 

...and fixes #2895


atc/api/buildserver/abort.go
```diff
@@ -14,14 +13,7 @@ func (s *Server) AbortBuild(build db.Build) http.Handler {
...
 
-		engineBuild, err := s.engine.LookupBuild(aLog, build)
-		if err != nil {
-			aLog.Error("failed-to-lookup-build", err)
-			w.WriteHeader(http.StatusInternalServerError)
-			return
-		}
-
-		err = engineBuild.Abort(aLog)
+		err := build.MarkAsAborted()
```
vito: Related: https://github.com/concourse/concourse/issues/1585

Should I prioritize that soon while this is fresh in our heads?

atc/api/buildserver/server.go
```diff
@@ -16,9 +15,7 @@ type Server struct {
 	logger lager.Logger
 
 	externalURL string
-	peerURL     string
 
-	engine              engine.Engine
```
vito: dope

atc/api/jobserver/create_build.go
```diff
@@ -33,29 +32,10 @@ func (s *Server) CreateJobBuild(pipeline db.Pipeline) http.Handler {
...
-
-		resources, err := pipeline.Resources()
-		if err != nil {
-			logger.Error("failed-to-get-resources", err)
-			w.WriteHeader(http.StatusInternalServerError)
-			return
-		}
-
-		build, _, err := scheduler.TriggerImmediately(logger, job, resources, versionedResourceTypes)
+		build, err := job.CreateBuild()
```
vito: Shouldn't we make sure that this created build is marked 'triggered' so that it gets up-to-date resource versions? As discussed in https://github.com/concourse/concourse/issues/2438

It'd be nice if we kicked the scheduler loop here too. Fast is a feature.

atc/api/jobserver/list_inputs.go
```diff
@@ -28,21 +27,13 @@ func (s *Server) ListJobInputs(pipeline db.Pipeline) http.Handler {
...
-		scheduler := s.schedulerFactory.BuildScheduler(pipeline, s.externalURL, variables)
-
 		resources, err := pipeline.Resources()
 		if err != nil {
 			logger.Error("failed-to-get-resources", err)
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
 
-		err = scheduler.SaveNextInputMapping(logger, job, resources)
```
vito: :+1:

atc/api/pipelineserver/build.go
```diff
@@ -22,22 +21,13 @@ func (s *Server) CreateBuild(pipelineDB db.Pipeline) http.Handler {
 			return
 		}
 
-		build, err := pipelineDB.CreateOneOffBuild()
+		build, err := pipeline.CreateStartedBuild(plan)
```
vito: kicking the build tracker here would be nice

atc/atccmd/command.go
```diff
@@ -1029,7 +1005,7 @@ func (cmd *RunCommand) defaultBindIP() net.IP {
 	return net.ParseIP(URL)
 }
 
-func (cmd *RunCommand) defaultURL() flag.URL {
+func (cmd *RunCommand) DefaultURL() flag.URL {
```
vito: Should we call this `LocalURL` or something? I always get confused by the name.

atc/engine/db_engine.go
```diff
@@ -15,31 +14,29 @@ import (
 
 const trackLockDuration = time.Minute
 
-func NewDBEngine(engines Engines, peerURL string) Engine {
```
vito: dope

atc/db/migration/migrations/1551384520_fix_build_private_plan.up.go
```diff
@@ -0,0 +1,76 @@
...
+	rows, err := tx.Query("SELECT id, private_plan, nonce FROM builds WHERE private_plan IS NOT NULL")
+	if err != nil {
+		return err
+	}
+
+	builds := []build{}
+	for rows.Next() {
+
+		build := build{}
+		if err = rows.Scan(&build.id, &build.plan, &build.nonce); err != nil {
```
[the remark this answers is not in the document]
vito: (Same goes for the 'down' migration.)

atc/db/pipeline.go
```diff
@@ -50,7 +50,10 @@ type Pipeline interface {
 	GetBuildsWithVersionAsInput(int, int) ([]Build, error)
 	GetBuildsWithVersionAsOutput(int, int) ([]Build, error)
 	Builds(page Page) ([]Build, Pagination, error)
+
 	CreateOneOffBuild() (Build, error)
+	CreateStartedBuild(plan atc.Plan) (Build, error)
```
vito: Do we still need `CreateOneOffBuild`?

atc/db/team.go
```diff
@@ -43,16 +44,20 @@ type Team interface {
 	OrderPipelines([]string) error
 
 	CreateOneOffBuild() (Build, error)
+	CreateStartedBuild(plan atc.Plan) (Build, error)
```
vito: Do we still need `CreateOneOffBuild`?

atc/db/pipeline.go
```diff
@@ -972,6 +975,56 @@ func (p *pipeline) CreateOneOffBuild() (Build, error) {
...
+		"schema":       "exec.v2",
+		"private_plan": encryptedPlan,
+		"public_plan":  plan.Public(),
+		"nonce":        nonce,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	err = build.saveEvent(tx, event.Status{
```
vito: Saving an event here feels a little awkward. When does this normally get sent? Will it be sent twice now?

I wonder if it would be better to create it 'pending' and have the tracker also process pending one-off builds?

atc/db/team.go
```diff
@@ -647,6 +677,55 @@ func (t *team) CreateOneOffBuild() (Build, error) {
...
+		"schema":       "exec.v2",
+		"private_plan": encryptedPlan,
+		"public_plan":  plan.Public(),
+		"nonce":        nonce,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	err = build.saveEvent(tx, event.Status{
```
vito: same concern as above - i'm a little iffy on inserting these in 'started' state

atc/api/artifactserver/create.go
```diff
@@ -0,0 +1,68 @@
...
+		// of 'creating' volume.
+
+		// In the long run CreateVolume should probably return a 'creating'
+		// volume, but there are other changes needed in FindOrCreateContainer
+		// with the way we create volumes for a container
+
+		// I think leaving the race condition is fine for now. Worst case
+		// is a fly execute will fail and the user will need to rerun it.
+
+		artifact, err := volume.InitializeArtifact("", 0)
```
vito: 'create' -> 'initialize' -> 'stream in data' feels unintuitive - normally it's 'initialized' only when it has all its data. it seems like this is to paper over the GC race condition. also why are both arguments zero-values?

atc/db/worker_artifact.go
```diff
@@ -0,0 +1,119 @@
...
+	buildID   int
+	createdAt time.Time
+}
+
+func (a *artifact) ID() int              { return a.id }
+func (a *artifact) Name() string         { return a.name }
+func (a *artifact) BuildID() int         { return a.buildID }
+func (a *artifact) CreatedAt() time.Time { return a.createdAt }
+
+func (a *artifact) Volume(teamID int) (CreatedVolume, bool, error) {
```
vito: surprised this is given a team ID - will it ever have multiple volumes for different teams? or is this just for checking access?

atc/exec/task_step.go
```diff
@@ -440,8 +437,13 @@ func (action *TaskStep) registerOutputs(logger lager.Logger, repository *worker.
 
 		for _, mount := range volumeMounts {
 			if filepath.Clean(mount.MountPath) == filepath.Clean(outputPath) {
-				source := newTaskArtifactSource(mount.Volume)
-				repository.RegisterSource(worker.ArtifactName(outputName), source)
+				source := NewTaskArtifactSource(mount.Volume)
+				repository.RegisterSource(artifact.Name(outputName), source)
+
+				_, err := mount.Volume.InitializeArtifact(outputName, action.buildID)
```
vito: So rather than having a step that exists solely to proxy an artifact to the user, we initialize all task outputs as artifacts and just allow them to be downloaded via the API?

Does this mean that each build's task outputs now live for 12 hours? It seems like that could pile up and result in a lot more disk usage if so, assuming a build runs frequently enough.

atc/plan.go
```diff
@@ -28,7 +28,8 @@ type Plan struct {
 type PlanID string
 
 type UserArtifactPlan struct {
-	Name string `json:"name"`
+	ArtifactID int    `json:"artifact_id"`
+	Name       string `json:"name"`
 }
 
 type ArtifactOutputPlan struct {
```
vito: Should this be removed?

atc/scheduler/runner.go
```diff
@@ -22,15 +22,6 @@ type BuildScheduler interface {
 		resources db.Resources,
 		resourceTypes atc.VersionedResourceTypes,
 	) (map[string]time.Duration, error)
-
-	TriggerImmediately(
```
vito: dope

atc/worker_artifact.go
```diff
@@ -0,0 +1,8 @@
+package atc
+
+type WorkerArtifact struct {
+	ID        int    `json:"id"`
+	Name      string `json:"name"`
+	BuildID   int    `json:"build_id"`
+	CreatedAt int64  `json:"created_at"`
```
vito: Not actually sure what format we typically use for timestamps in the API, but we should check this to make sure we're consistent.

fly/commands/internal/executehelpers/inputs.go
```diff
@@ -127,18 +131,30 @@ func TaskInputsContainsName(inputs []atc.TaskInputConfig, name string) bool {
...
+	fact atc.PlanFactory,
+	team concourse.Team,
+	inputMappings []flaghelpers.InputPairFlag,
+	includeIgnored bool,
+) (map[string]Input, error) {
+
 	kvMap := map[string]Input{}
 
 	for _, i := range inputMappings {
+		artifact, err := Upload(team, i.Path, includeIgnored)
```
vito: I noticed we don't show upload/download progress anymore, now that the steps don't print it. It'd be nice to have that back.

atc/api/buildserver/create.go
```diff
@@ -22,22 +22,13 @@ func (s *Server) CreateBuild(team db.Team) http.Handler {
...
 		}
 
-		engineBuild, err := s.engine.CreateBuild(hLog, build, plan)
-		if err != nil {
-			hLog.Error("failed-to-start-build", err)
-			w.WriteHeader(http.StatusInternalServerError)
-			return
-		}
-
-		go engineBuild.Resume(hLog)
```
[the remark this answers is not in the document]
pivotal-jwinters: yeah can we add this to a general LISTEN/NOTIFY story? I think we could take this approach in a few more places.

atc/db/migration/migrations/1551384520_fix_build_private_plan.up.go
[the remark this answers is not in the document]
pivotal-jwinters: heh, yeah that makes sense. I'll batch it.

atc/api/accessor/accessor_test.go
```diff
@@ -651,12 +650,16 @@ var _ = Describe("Accessor", func() {
...
-		Entry("owner :: "+atc.SendInputToBuildPlan, atc.SendInputToBuildPlan, "owner", true),
-		Entry("member :: "+atc.SendInputToBuildPlan, atc.SendInputToBuildPlan, "member", true),
-		Entry("viewer :: "+atc.SendInputToBuildPlan, atc.SendInputToBuildPlan, "viewer", false),
+		Entry("owner :: "+atc.CreateArtifact, atc.CreateArtifact, "owner", true),
+		Entry("member :: "+atc.CreateArtifact, atc.CreateArtifact, "member", true),
+		Entry("viewer :: "+atc.CreateArtifact, atc.CreateArtifact, "viewer", false),
+
+		Entry("owner :: "+atc.GetArtifact, atc.GetArtifact, "owner", true),
+		Entry("member :: "+atc.GetArtifact, atc.GetArtifact, "member", true),
+		Entry("viewer :: "+atc.GetArtifact, atc.GetArtifact, "viewer", true),
```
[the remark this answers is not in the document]
pivotal-jwinters: hmm yeah maybe we should make it member only, there could be sensitive info in these artifacts.

atc/api/artifactserver/create.go
```diff
@@ -0,0 +1,68 @@
...
+
+		// This happens because CreateVolume returns a 'created' instead
+		// of 'creating' volume.
+
+		// In the long run CreateVolume should probably return a 'creating'
+		// volume, but there are other changes needed in FindOrCreateContainer
+		// with the way we create volumes for a container
+
+		// I think leaving the race condition is fine for now. Worst case
+		// is a fly execute will fail and the user will need to rerun it.
```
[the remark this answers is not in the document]
pivotal-jwinters: It gets GC'ed because it returns a created volume, with nothing keeping it around. I'll investigate to see if I can do some refactoring to get it to return a creating volume.

atc/api/artifactserver/create.go
```diff
@@ -0,0 +1,68 @@
...
+
+		artifact, err := volume.InitializeArtifact("", 0)
+		if err != nil {
+			hLog.Error("failed-to-initialize-artifact", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		err = volume.StreamIn("/", r.Body)
```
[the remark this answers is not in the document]
pivotal-jwinters: yep

atc/api/artifactserver/create.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/concourse/concourse/atc/worker"
+)
+
+func (s *Server) CreateArtifact(team db.Team) http.Handler {
+	hLog := s.logger.Session("create-artifact")
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+
+		// TODO: can probably check if fly sent us an etag header
```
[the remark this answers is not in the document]
pivotal-jwinters: yep

atc/api/buildserver/server.go
pivotal-jwinters: yep

atc/api/buildserver/abort.go
pivotal-jwinters: Maybe this could be something to look into when we do the broader LISTEN/NOTIFY work?

atc/api/artifactserver/get.go
```diff
@@ -0,0 +1,66 @@
...
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		defer reader.Close()
+
+		_, err = io.Copy(w, reader)
+		if err != nil {
+			logger.Error("failed-to-encode-artifact", err)
+			w.WriteHeader(http.StatusInternalServerError)
```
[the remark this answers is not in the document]
pivotal-jwinters: not sure what you mean? why does this write multiple headers?

atc/api/jobserver/create_build.go
pivotal-jwinters: This already happens here: https://github.com/concourse/concourse/blob/ca9fd7c33fe63291ef291d40c4bd9ccc2ba692d4/atc/db/pipeline.go#L254-L255

atc/api/jobserver/list_inputs.go
pivotal-jwinters: yep

atc/api/pipelineserver/build.go
pivotal-jwinters: LISTEN/NOTIFY work?

atc/atccmd/command.go
pivotal-jwinters: ¯\_(ツ)_/¯

atc/engine/db_engine.go
pivotal-jwinters: yep

atc/db/pipeline.go
pivotal-jwinters: ugh, yes. its used absolutely everywhere in the tests and I did not want to mess with that.

atc/db/team.go
pivotal-jwinters: so. many. tests.

atc/db/pipeline.go
pivotal-jwinters: I agree, but i think this is a larger discussion around build state. Seems like we need an intermediate `planned` state.

This means the `BuildStarter` actually becomes a `BuildPlanner`.

atc/db/team.go
pivotal-jwinters: RIP BuildStarter, long live BuildPlanner

atc/db/worker_artifact.go
pivotal-jwinters: yeah its just for checking access. its really unfortunate that teamID exists on the volume itself. It would be a lot nicer and more intuitive if each "higher level resource" had the teamID.

atc/exec/task_step.go
pivotal-jwinters: TBH I never even considered putting this in its own step (which is hilarious because I had to delete the old output artifact build step to make this change).

It wouldn't really help us with the lifecycle, but it would at least limit the number of `artifacts` that get created to only one-off builds.

atc/plan.go
pivotal-jwinters: nah, we still need it

atc/scheduler/runner.go
pivotal-jwinters: yep

atc/worker_artifact.go
pivotal-jwinters: from what i saw we aren't consistent lol

fly/commands/internal/executehelpers/inputs.go
pivotal-jwinters: erm yeah, i'll look into that

atc/api/artifactserver/create.go
pivotal-jwinters: yes yes and yes. we can have this flow make more sense if we figure out the race condition.

lets just pretend those zero values aren't there... they get used for artifacts from build outputs.

pivotal-jwinters:
- [ ] batch migration
- [ ] get artifact -> member only
- [ ] creating volume
- [ ] upload/download status
- [ ] artifact lifecycle

fly/commands/internal/executehelpers/inputs.go
pivotal-jwinters: I spoke with @xtremerui and he seems to think this got removed a while ago. If we want that back can we create a separate issue?

vito: It's there for me:

```
~/w/concourse (fix-dashboard-search) $ fly -t ci e -c ci/tasks/docker-compose-watsjs.yml -i concourse=. -p
executing build 4989469 at https://ci.concourse-ci.org/builds/4989469 
initializing
concourse: 16.54 MiB/s 0s                                                       
running concourse/ci/tasks/scripts/with-docker-compose concourse/ci/tasks/scripts/watsjs
/tmp/build/e55deab7/concourse /tmp/build/e55deab7
Creating network "concourse_default" with the default driver
Pulling db (postgres:)...
```

It's not a progress bar, but it still shows the upload rate and duration.

It shouldn't take long to restore it, and I'd like to avoid user-facing regressions on `master` that prevent us from shipping, so I don't want to merge without it. 😅 

You could probably just hoist the code from the user artifact step into `fly`:

https://github.com/concourse/concourse/blob/b20ee6770be162614a2d7e561c5bff095b81db93/atc/exec/user_artifact_step.go#L54-L62

atc/exec/task_step.go
vito: Hmm should we bring the artifact output step back, then? I wouldn't want to ship with all task outputs lasting for 12 hours.

atc/api/artifactserver/create.go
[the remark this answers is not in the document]
vito: Sounds fine.

atc/exec/task_step.go
pivotal-jwinters: Yeah working on that now. It adds a lot better symmetry :)

fly/commands/internal/executehelpers/inputs.go
pivotal-jwinters: ah cool! im going to pull this into fly execute instead of having it in the step.

atc/plan.go
pivotal-jwinters: err I originally thought you we're talking about the `UserArtifactPlan` because of how its highlighted. I now realize you actually meant the ArtifactOutputPlan, which coincidently we are bringing back because of other things you mentioned :)

pivotal-jwinters: glad it never got removed in the first place ;)

atc/db/migration/migrations/1551384520_fix_build_private_plan.up.go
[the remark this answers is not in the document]
pivotal-jwinters: I actually remember this now. I don't think this needs to be batched. The `private_plan` only exists for the duration that the build is running. We query for builds where `private_plan` IS NOT NULL so that shouldn't actually be very many things.

pivotal-jwinters:
@vito based on these commits I just pushed, and some comments, I think I addressed all your concerns. Let me know if you think there's something that I missed.

pivotal-jwinters:
Also fixed the double status. NOW it should be good to go.

atc/plan.go
vito: lol, yep 👍

atc/db/migration/migrations/1551384520_fix_build_private_plan.up.go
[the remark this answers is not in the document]
pivotal-jwinters: Currently there are only 42 builds on wings where this is true.

atc/api/buildserver/create.go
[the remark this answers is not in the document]
vito: created https://github.com/concourse/concourse/issues/3508

PR #3316, opened 2019-02-15
dependency script

cperry169:
This is a simple install script for Ubuntu 18.04 to setup a development environment for Concourse CI.

### DIFF
--- a/dev/install-dependencies-ubuntu/dependencies.sh
+++ b/dev/install-dependencies-ubuntu/dependencies.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+exec < /home/USER/concourse-ubuntu.log
+set -x
+set -e
+
+# snap
+sudo apt update
+sudo apt install snapd
+
+# yarn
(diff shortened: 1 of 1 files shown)


vito:
Hey, thanks for submitting this but I don't think we should maintain these kinds of things in the repo. Dev machines tend to be pretty opinionated and there can be any number of ways to install the dependencies depending on the OS, distribution, etc.

I think we should just have our contributing docs link to each tool's install instructions as they should always be up-to-date and already documented for the various platforms.

On the other hand, if folks do find this helpful I could see this being added in the Wiki. We could start on a category for that. I think the wiki already open for everyone to edit, so if you're interested, feel free to give it a shot.

Thanks again!

cperry169:
Can you add it to the Wiki

Sincerely,

Chad Perry
CompTIA A+, Network+, Server+, Linux+ SME
cperry169@hotmail.com
512.961.0291

Linkedin: www.linkedin.com/in/chadbperry
GitHub: https://github.com/cperry169

________________________________

PR #3319, opened 2019-02-15
atc/worker/: Add log to check if container lock is used

ddadlani:
Add log line to check if container locks are ever not acquired in containerProvider. We are using this as a metric to decide if we need to keep container locking since FindOrCreateContainer is always called within other held locks.

### DIFF
--- a/atc/worker/container_provider.go
+++ b/atc/worker/container_provider.go
@@ -171,6 +171,7 @@ func (p *containerProvider) FindOrCreateContainer(
 
 			if !acquired {
 				time.Sleep(creatingContainerRetryDelay)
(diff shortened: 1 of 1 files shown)


kcmannem:
@vito pls

atc/worker/container_provider.go
```diff
@@ -171,6 +171,7 @@ func (p *containerProvider) FindOrCreateContainer(
 
 			if !acquired {
 				time.Sleep(creatingContainerRetryDelay)
+				logger.Debug(fmt.Sprintf("did-not-acquire-creating-container-lock-%d", creatingContainer.ID()))
```
vito: this should use `lager.Data` instead of putting it in the message itself

ddadlani: thanks. Made the change

Issue #2592, opened 2018-09-15
Worker key generated on my machine doesn't work: "asn1: structure error"


TJSomething:
# Bug Report

* Concourse version: 4.1.0
* Deployment type (BOSH/Docker/binary): binary

I was testing out running a Concourse worker on my dev computer and I'm having an odd issue where Concourse will not accept a worker private key generated on my machine (Debian buster, with openssh-client 1:7.8p1-1), but it works if I generate it on another machine (Ubuntu xenial, with openssh-client 1:7.2p2-4ubuntu2.4), giving me the error:

```
invalid argument for flag `--tsa-worker-private-key' (expected *flag.PrivateKey): asn1: structure error: tags don't match (16 vs {class:1 tag:15 length:112 isCompound:true}) {optional:false explicit:false application:false private:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:false} pkcs8 @2
```

This is easy to work around, but really strange. As far as I can tell, this is a common problem with the with various Go applications with no solutions.

The full log, with some of my machine's details:

```
root@localhost:/opt/concourse# dpkg -S /usr/bin/ssh-keygen
openssh-client: /usr/bin/ssh-keygen
root@localhost:/opt/concourse# dpkg -l openssh-client
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                 Version         Architecture    Description
+++-====================-===============-===============-==============================================
ii  openssh-client       1:7.8p1-1       amd64           secure shell (SSH) client, for secure access t
root@localhost:/opt/concourse# ssh-keygen -t rsa -f worker_key_local -N ''
Generating public/private rsa key pair.
Your identification has been saved in worker_key_local.
Your public key has been saved in worker_key_local.pub.
The key fingerprint is:
SHA256:6XhtWTZZSpTZuN/L/UkfdZcVUAQdgcOSaHtwx0cSQyQ root@localhost
The key's randomart image is:
+---[RSA 2048]----+
|          .EB&XBo|
|         + ==*+o.|
|        . + +.+ .|
|         o o.+  o|
|        S . *. o+|
|       o . + .. =|
|      . o +   .oo|
|       . .    .o=|
|               .+|
+----[SHA256]-----+
[...]

Issue #2526, opened 2018-08-23
4.0.0 - docker web panic runtime error SIGSEGV


eedwards-sk:
# Bug Report

* Concourse version: 4.0.0
* Deployment type (BOSH/Docker/binary): Docker (web / db / linux worker) Darwin (worker)
* Infrastructure/IaaS:
* Browser (if applicable):
* Did this used to work? No

I'm trying to run a concourse stack locally with a CI build tool. The web container is ending up in a restart loop spammed with this in the logs. Seems to be uncaught exception of sorts?

```
goroutine 1 [running]:
crypto/rsa.(*PrivateKey).Public(0x0, 0x0, 0x0)
	/usr/local/go/src/crypto/rsa/rsa.go:92 +0x5
golang.org/x/crypto/ssh.NewSignerFromSigner(0x7f69ac53c470, 0xc420190408, 0xc420190408, 0x7f69ac53c470, 0xc420190408, 0xec4101)
	/tmp/build/9674af12/concourse/src/golang.org/x/crypto/ssh/keys.go:681 +0x35
golang.org/x/crypto/ssh.NewSignerFromKey(0x299b760, 0xc420190408, 0xec4178, 0x30, 0x2afd220, 0xc4206193f8)
	/tmp/build/9674af12/concourse/src/golang.org/x/crypto/ssh/keys.go:656 +0x163
github.com/concourse/tsa/tsacmd.(*TSACommand).configureSSHServer(0xc4201809c0, 0xc420387230, 0xc420591bb0, 0x1, 0x1, 0x0, 0x0, 0x0, 0xc422d46f00, 0xc4206caca0, ...)
	/tmp/build/9674af12/concourse/src/github.com/concourse/tsa/tsacmd/command.go:203 +0x18e
github.com/concourse/tsa/tsacmd.(*TSACommand).Runner(0xc4201809c0, 0xc420591890, 0x0, 0x1, 0x143ecd40, 0xc42083f360, 0xc4204aa800, 0x0)
	/tmp/build/9674af12/concourse/src/github.com/concourse/tsa/tsacmd/command.go:101 +0x249
main.(*WebCommand).Runner(0xc4207fa708, 0xc420591890, 0x0, 0x1, 0x2149481, 0xc4202e0100, 0xc42079d660, 0x21499d0)
	/tmp/build/9674af12/concourse/src/github.com/concourse/bin/cmd/concourse/web.go:67 +0x118
[...]

PR #3330, opened 2019-02-19
add generate-key command

### LINKED ISSUE #2779: Add 'concourse generate-key' command

Context: https://github.com/concourse/docs/pull/124

Apparently good ol' `ssh-keygen` isn't quite foolproof enough on some machines. It's also not really something we can assume on most Windows deployments.

It'd be easier and more portable if we just had a command to generate the keys. Then we could guarantee that Go would be happy with them, since we have more control over how they're generated in the first place.

To be honest this command could just straight-up mimick the behavior of `ssh-keygen`, maybe with opinionated defaults for heightened security (e.g. block size 4096?). I haven't put much thought into this though. We could make it super fancy and know to generate the set of keys that Concourse needs (signing key, ssh gateway (tsa) key, worker key), but that may be overkill since people may want to run it to just generate individual worker keys, too. So maybe it should just be one key per invocation.

If I only had the time!

vito:
this is intended to replace 'ssh-keygen' in our docs, and has the
following benefits:

* it's more portable - ssh-keygen isn't as easy to assume on Windows
* it's more likely to generate keys that our components can consume
* we can have safer defaults (e.g. key size 4096)

fixes #2779

see also #2526, #2592, concourse/docs#124


cmd/concourse/generate_key.go
```diff
@@ -0,0 +1,76 @@
...
+	"fmt"
+	"os"
+
+	"golang.org/x/crypto/ssh"
+)
+
+type GenerateKeyCommand struct {
+	Type string `short:"t"  long:"type"  default:"rsa"  choice:"rsa"  choice:"ssh"  description:"The type of key to generate."`
+
+	FilePath string `short:"f"  long:"filename"  required:"true"  description:"File path where the key shall be created."`
```
topherbullock:
```suggestion
	FilePath string `short:"f"  long:"filename"  required:"true"  description:"File path where the key shall be created. When generating ssh keys, the public key will be stored in a file with the same name but with '.pub' appended."`
```

cmd/concourse/command.go
```diff
@@ -18,6 +18,8 @@ type ConcourseCommand struct {
 
 	LandWorker   land.LandWorkerCommand     `command:"land-worker" description:"Safely drain a worker's assignments for temporary downtime."`
 	RetireWorker retire.RetireWorkerCommand `command:"retire-worker" description:"Safely remove a worker from the cluster permanently."`
+
+	GenerateKey GenerateKeyCommand `command:"generate-key" description:"Generate RSA keys for worker auth."`
```
topherbullock:
```suggestion
	GenerateKey GenerateKeyCommand `command:"generate-key" description:"Generate RSA key for use with Concourse components."`
```

Issue #3321, opened 2019-02-15
Mixed feelings about absolute time in build history.


bwinter:
The new switch to using absolute time on concourse job history seems like a step in the wrong direction to me. My general use case is to see how long something has been failing, the specific date seems less useful generally. I have a suspicion I will now just be forced to do this date math in my head every time I want to know how long ago something started failing.

Maybe there is a middle ground here or maybe I just need to spend time with this new UI either way, I wanted to get this conversation started.

bwinter:
Someone suggested a hybrid that uses github's technique. Namely showing "ago" notation on recent changes then switching to absolute times further out. This seems like a good compromise and seems to support my use cases.

vito:
related to #3162

/cc @analytically who left similar feedback here: https://github.com/concourse/concourse/commit/3544de50b0f073eaa864661396c67df1c1fd8854#commitcomment-32307862

bwinter:
Looks like this is the conversation that started the change: https://github.com/concourse/concourse/issues/3019

analytically:
Another option would be in-place replacing the date text when clicked. Flip from relative to absolute with a click on the time.

analytically:
And store the selection of relative/absolute client-side?

evanchaoli:
I like @analytically 's idea of "flipping". Seems like different people think in different perspectives. We initiated the absolution time change, because our users strongly asked for that. As we made the absolution change, we'll try to implement the "flip" feature.

vito:
@evanchaoli FWIW, before this change the absolute time was available as a tooltip on hover. Guessing that wasn't enough? Just checking to make sure it wasn't an oversight.

Twiknight:
The change was submitted by me. The reason is, relative time looks confusing in my use case.

Since there are supports on both side, I think it would be better if we can have a choice on how the history time looks like.

Well, 2 rough ideas in my mind is, 

1. Add tooltip to the history time, and show `some time ago` in the tooltip
   
<img width="546" alt="screen shot 2019-02-18 at 10 20 21 am" src="https://user-images.githubusercontent.com/11556597/52924262-fca71580-3366-11e9-97ba-15452842ca2e.png">

2. Add a switch in header or bottom bar to switch the time format.

![screen shot 2019-02-18 at 10 40 16 am](https://user-images.githubusercontent.com/11556597/52924892-a7203800-3369-11e9-8f25-bec6b08d5efa.png)

Any ideas?

[...]

PR #3333, opened 2019-02-20
use github notion on build history start/finish timestamp

Twiknight:
Signed-off-by: Twiknight <Twiknight@outlook.com>

Improve the timestamp display on build history according to https://github.com/concourse/concourse/issues/3321#issuecomment-464639912

Now if a time (any of started/finished) is
1. more than 24hrs ago -> show verbose date ( `"%b %d %Y %I:%M:%S %p"`), and use relative date (`sometime ago`) as tooltip
2. less than 24hrs ago -> show relative date and use verbose date as tooltip.

following is an example
screen shot 2019-02-20 at 11 07 03 am

### DIFF
--- a/web/elm/src/BuildDuration.elm
+++ b/web/elm/src/BuildDuration.elm
@@ -17,10 +17,10 @@ view duration now =
                 [ pendingLabel "pending" ]
 
             ( Just startedAt, Nothing ) ->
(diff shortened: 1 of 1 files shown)


vito:
Thanks!

PR #3334, opened 2019-02-20
fix task variable lookup order

ralekseenkov:
hey @vito

I've been testing task vars and was writing blog post on external tasks today. Found some oddness, which we need to fix. You actually mentioned this before, being unsure about the right order of passing params to the templating engine.

The current behavior is --- cred mgr first and then task vars. The problem with that approach is that it issues a lot of unnecessary lookups to cred mgrs, even though vars are specified (i.e. you specify a var and don't even expect Concourse to reach out to cred mgr, but it still goes there)

After thinking more about this, I think the behavior should be the opposite. If vars are specified, we should take a variable from there (it makes total sense. if it's specified explicitly, then take it! it overrides anything else). If it's not in vars, then we should defer lookup to cred manager.

This turned out to be a somewhat easy fix:
* now, the order in which you pass params, is the order in which they are taken
* there is only one place where we create an array of two values and it's `[]boshtemplate.Variables{boshtemplate.StaticVariables(plan.Task.Vars), credMgrVariables}`. the semantics makes total sense here
* added a comment to NewTemplateResolver() to clarify the interface, and added unit test which checks the order
* the only part of the code which still does processing of variables in reverse order is old style templates. and this code will hopefully be dropped sooner rather than later, along with support for old style templates...

### DIFF
--- a/atc/exec/garden_factory.go
+++ b/atc/exec/garden_factory.go
@@ -150,13 +150,13 @@ func (factory *gardenFactory) Task(
 		// external task - construct a source which reads it from file
 		taskConfigSource = FileConfigSource{ConfigPath: plan.Task.ConfigPath}
 
-		// use 'vars' from the pipeline + cred mgr variables for interpolation
+		// for interpolation - use 'vars' from the pipeline, and then fill remaining with cred mgr variables
 		taskVars = []boshtemplate.Variables{boshtemplate.StaticVariables(plan.Task.Vars), credMgrVariables}
 	} else {
 		// embedded task - first we take it
 		taskConfigSource = StaticConfigSource{Config: plan.Task.Config}
 
-		// use just cred mgr variables for interpolation
+		// for interpolation - use just cred mgr variables
 		taskVars = []boshtemplate.Variables{credMgrVariables}
 	}
 
--- a/fly/commands/internal/templatehelpers/yaml_template.go
+++ b/fly/commands/internal/templatehelpers/yaml_template.go
@@ -48,7 +48,21 @@ func (yamlTemplate YamlTemplateWithParams) Evaluate(
 	}
 
 	var params []boshtemplate.Variables
-	for _, path := range yamlTemplate.templateVariablesFiles {
+
+	// first, we take explicitly specified variables on the command line
+	vars := boshtemplate.StaticVariables{}
+	for _, f := range yamlTemplate.templateVariables {
+		vars[f.Name] = f.Value
+	}
+	for _, f := range yamlTemplate.yamlTemplateVariables {
+		vars[f.Name] = f.Value
+	}
+	params = append(params, vars)
+
+	// second, we take all files. with values in the files specified later on command line taking precedence over the
+	// same values in the files specified earlier on command line
+	for i := len(yamlTemplate.templateVariablesFiles) - 1; i >= 0; i-- {
+		path := yamlTemplate.templateVariablesFiles[i]
 		templateVars, err := ioutil.ReadFile(string(path))
 		if err != nil {
 			return nil, fmt.Errorf("could not read template variables file (%s): %s", string(path), err.Error())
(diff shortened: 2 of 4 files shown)


atc/exec/garden_factory.go
```diff
@@ -150,13 +150,13 @@ func (factory *gardenFactory) Task(
 		// external task - construct a source which reads it from file
 		taskConfigSource = FileConfigSource{ConfigPath: plan.Task.ConfigPath}
 
-		// use 'vars' from the pipeline + cred mgr variables for interpolation
+		// for interpolation - use 'vars' from the pipeline, and then fill remaining with cred mgr variables
 		taskVars = []boshtemplate.Variables{boshtemplate.StaticVariables(plan.Task.Vars), credMgrVariables}
```
vito: Hmm are you sure this shouldn't just be reversed? For example, the current semantics with `-l` are that the right-most side wins, even with `((vars))` in a pipeline:

```sh
~/w/concourse (fix-smoke) $ cat booklit.yml 
resources:
- name: booklit
  type: git
  source:
    uri: https://github.com/vito/booklit
    branch: ((branch))

jobs:
- name: unit
  plan:
  - get: booklit
  - task: test
    file: booklit/ci/test.yml

~/w/concourse (fix-smoke) $ cat develop.yml 
branch: develop

~/w/concourse (fix-smoke) $ cat master.yml 
branch: master

~/w/concourse (fix-smoke) $ fly -t l validate-pipeline -o -c booklit.yml -l develop.yml -l master.yml
jobs:
- name: unit
  plan:
  - get: booklit
  - file: booklit/ci/test.yml
    task: test
resources:
- name: booklit
  source:
    branch: master
    uri: https://github.com/vito/booklit
  type: git
```

The mental model is that they are appended and stacked on top of each other. So technically what's listed first is lowest precedence.

ralekseenkov: Yes. The command line semantics doesn't change - whatever specified on the right side wins. We want to preserve that. 

But, in the source code, I always felt a normal order should win:
* when you iterate over a slice or array, the default is to iterate in natural order and break from the loop if a match is found
* this is how boshtemplate.NewMultiVars(), for example, is implemented. so I always felt that whatever we pass into NewTemplateResolver() should not be reversed before passing to boshtemplate.NewMultiVars()
* and when I wrote `[]boshtemplate.Variables{boshtemplate.StaticVariables(plan.Task.Vars), credMgrVariables}`, the original intention was that plans vars should be tried first.

So I am trying to stick to my original plan. But if you strongly feel that we should follow "stacked on top of each other" approach in the source code too, I can make that change and change all comments to make sure the interface is clear.

ralekseenkov: P.S. I don't think this is going to solve the Vault slowness issue though. This only affects tasks with FileConfigSource, when an array with more than one element is passed to NewTemplateResolver(). Vault issue happens on regular tasks too, where an array consists of just one element with credential manager vars.

vito: Sounds fine, just wanted to make sure this didn't bubble up and change the existing CLI behavior. :+1:

fly/commands/internal/templatehelpers/yaml_template.go
```diff
@@ -48,7 +48,21 @@ func (yamlTemplate YamlTemplateWithParams) Evaluate(
...
+		vars[f.Name] = f.Value
+	}
+	for _, f := range yamlTemplate.yamlTemplateVariables {
+		vars[f.Name] = f.Value
+	}
+	params = append(params, vars)
+
+	// second, we take all files. with values in the files specified later on command line taking precedence over the
+	// same values in the files specified earlier on command line
+	for i := len(yamlTemplate.templateVariablesFiles) - 1; i >= 0; i-- {
```
vito: :+1: didn't notice this

PR #3345, opened 2019-02-20
web: better icons for unresolved step states

### LINKED ISSUE #3022: icons for unresolved step states on build page

While transitioning to the material design icon set in #2405, we noticed a lack of good icons to represent the following states for a step on the build page:
1. Pending - the build is currently running, but the step in question has not yet started.
1. Interrupted - the build stopped while the step in question was running.
1. Cancelled/Never reached - the build stopped before the step in question started.

Initial ideas are
1. Pending - circle?
1. Interrupted - dashed circle
1. Cancelled - circle with a diagonal line through it, like a "cancel" icon

👍

pivotal-jamie-klassen:
fixes #3022

There is only one usage of fontawesome left in the project, and that is on the build prep section when the status of a prep is unknown. I'd pretty much bet we can use a spinner there instead of `fa-circle-o-notch`.


Lindsayauchin:
Yes, let's replace `fa-circle-o-notch` with the new loading spinner.

PR #3352, opened 2019-02-20
fix checking bosh properties, clean up debug/healthcheck flags

vito:
this had an empty filter which ended up filtering out everything, rather
than filtering nothing


vito:
I went on to update this a bit to go over all the actually-missing envs and either add them to the BOSH release (in https://github.com/concourse/concourse-bosh-release/commit/96460b210ccba6a225ee97a385b0ba71f8f5c3f5) or list them under 'ignored' as appropriate. Once this PR is merged the 'check-props' job *should* go green finally. And then I think we should have it gate the `bosh-bump` job.

Issue #3266, opened 2019-02-11
Move Concourse resource base images to Bionic


cirocosta:
With the guidelines around standardizing the base images for the container images we ship around [`ubuntu:bionic`](https://hub.docker.com/_/ubuntu), this issue keeps track of the progress around those.

- [x] check resources not owned by the concourse team (baked resources and external resources).
- [x] Use an ubuntu-based Golang base for our builds
  - [x] review OK
  - [x] PR merged: concourse/concourse#3243
- [x] resources
  - [x] [git-resource](https://github.com/concourse/git-resource)
    - [x] migrate to builder/intermediate images ubuntu:bionic
    - [x] migrate actual image to ubuntu:bionic and build.
    - [x] run image tests successfully.
    - [x] run basic smoke tests.
    - ~~create a test suite for automated tests.~~
    - [x] review OK
    - [x] PR merged: concourse/git-resource#241
    - [x] released
  - ~~[archive](https://github.com/concourse/archive-resource)~~
    - deprecated
  - [x] [bosh-io-release-resource](https://github.com/concourse/bosh-io-release-resource) 
    - [x] review OK
    - [x] PR merged: concourse/bosh-io-release-resource#18
    - [x] released
  - [x] [bosh-io-stemcell-resource](https://github.com/concourse/bosh-io-stemcell-resource) 
    - [x] review OK
    - [x] PR merged: concourse/bosh-io-stemcell-resource#23
    - [x] released
  - [x] [cf-resource](https://github.com/concourse/cf-resource) 
    - [x] review OK
    - [x] PR merged: concourse/cf-resource#69
    - [x] released
  - [x] [github-release-resource](https://github.com/concourse/github-release-resource) 
    - [x] review OK
    - [x] PR merged: concourse/github-release-resource#82
    - [x] released
  - [x] [docker-image-resource](https://github.com/concourse/docker-image-resource) 
    - [x] review OK
    - [x] PR merged: concourse/docker-image-resource#256
    - [x] released
  - [x] [hg-resource](https://github.com/concourse/hg-resource) 
    - [x] review OK
    - [x] PR merged: concourse/hg-resource#11
    - [x] released
  - [x] Decide if we're shipping [mock-resource](https://github.com/concourse/mock-resource)
    - [x] [mock-resource](https://github.com/concourse/mock-resource)
       - [x] review OK
       - [x] PR merged: concourse/mock-resource#2
       - [x] released
[...]

PR #3364, opened 2019-02-21
build both alpine and ubuntu resources

chenbh:
Part of concourse/concourse#3266
Requires all resource repos to have a `dockerfiles/alpine/Dockerfile` and `dockerfiles/ubuntu/Dockerfile`

This will modify the resource pipelines to build both the `alpine` and `ubuntu` image when releasing to GitHub.

- the image with the `latest` tag on docker hub will remain alpine
- PR images will also remain alpine
- the GitHub release now has an extra tar, `resource-1.0.0-ubuntu.tgz` (the existing one will be renamed `resource-1.0.0-alpine.tgz`)
- builds will also be triggered on new releases of `ubuntu:bionic` docker image (similar to how it's currently triggered on new releases of `alpine:edge`

### DIFF
--- a/ci/pipelines/resources/template.jsonnet
+++ b/ci/pipelines/resources/template.jsonnet
@@ -389,7 +389,7 @@ local publish_job(bump) = {
       ]
     },
     {
-      name: "prs",
+      name: "prs-alpine",
       serial: true,
       public: true,
       plan: [
@@ -412,6 +412,7 @@ local publish_job(bump) = {
           resource: "resource-pr",
           params: {
             path: "resource-pr",
+            context: "status-alpine",
             status: "pending"
           },
           get_params: {fetch_merge: true}
@@ -421,13 +422,14 @@ local publish_job(bump) = {
           params: {
             load_base: "alpine-edge",
             tag: resource+"-resource/.git/id",
-            tag_prefix: "pr-",
+            tag_prefix: "pr-alpine-",
             dockerfile: resource+"-resource/dockerfiles/alpine/Dockerfile",
           } + build_params,
           on_failure: {
             put: "resource-pr",
             params: {
               path: "resource-pr",
+              context: "status-alpine",
               status: "failure"
             }
           },
@@ -435,6 +437,63 @@ local publish_job(bump) = {
             put: "resource-pr",
             params: {
               path: "resource-pr",
+              context: "status-alpine",
+              status: "success"
+            }
+          }
+        }
+      ]
+    },
+    {
+      name: "prs-ubuntu",
+      serial: true,
+      public: true,
+      plan: [
+        {
+          aggregate: [
+            {
+              get: "resource-pr",
+              trigger: true,
+              version: "every"
+            },
+            {
+              get: "ubuntu-bionic",
+              params: {save: true},
+              trigger: true
+            },
+          ] + extra_gets
+        },
+        {
+          put: resource+"-resource",
+          resource: "resource-pr",
+          params: {
+            path: "resource-pr",
+            context: "status-ubuntu",
+            status: "pending"
+          },
+          get_params: {fetch_merge: true}
+        },
+        {
+          put: "resource-image-dev-ubuntu",
+          params: {
+            load_base: "ubuntu-bionic",
+            tag: resource+"-resource/.git/id",
+            tag_prefix: "pr-ubuntu-",
+            dockerfile: resource+"-resource/dockerfiles/ubuntu/Dockerfile",
+          } + build_params,
+          on_failure: {
+            put: "resource-pr",
+            params: {
+              path: "resource-pr",
+              context: "status-ubuntu",
+              status: "failure"
+            }
+          },
+          on_success: {
+            put: "resource-pr",
+            params: {
+              path: "resource-pr",
+              context: "status-ubuntu",
               status: "success"
             }
           }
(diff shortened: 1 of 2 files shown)


chenbh:
Blocked until the PRs for `bionic-base-images` for every resource has been merged

chenbh:
@vito This should have minimal impact on the existing CI pipeline but still allow us to do releases with both `alpine` and `ubuntu`.

Should be merged the same time as:
concourse/bosh-io-release-resource#18
concourse/bosh-io-stemcell-resource#23
concourse/cf-resource#69
concourse/docker-image-resource#256
concourse/git-resource#241
concourse/github-release-resource#82
concourse/hg-resource#11
concourse/pool-resource#45
concourse/registry-image-resource#14
concourse/s3-resource#120
concourse/semver-resource#88
concourse/time-resource#36
concourse/tracker-resource#24

cc @YoussB

cirocosta:
Hey, a possibly stupid thought that crossed my mind - If we're very concerned about saving as many bytes as we can, we could perhaps adhere to a standard where all of our images have a `test` stage, and a final `resource` stage (is that the case already?) and then leverage `--target=$STAGE`.

The rationale is that we'd then be able to totally discard those stages that do not need their bits included in the final resource (like, `test` stuff).

One reason why this might be stupid: our `test` stages might not even be needed in most of the resources as they can fit in the `build` stage anyway.

Wdyt?

Thx!

cirocosta:
On a slightly different note, we could also leverage `squash`ing the layers if we end up using many `ADD`s and `RUN`s that add files that we end up deleting later (`squash` is not yet supported in buildkit - used by `img`, which is used by `concourse/builder`), but as we use `docker-image` in these pipelines, we could use such feature.

Although our `rootfs`'s that we ship in the bosh release and embedded in the `concourse/concourse(-rc)` are already "squashed" (they're the final view of the rootfs), it would present reductions for people consuming that through the registry (possibly relevant for #3003)?

YoussB:
Regarding https://github.com/concourse/concourse/pull/3364#issuecomment-467449863
 - These stages don't end up being part of the final image, so I think it is safe to keep them.
 - the point of having the tests run in a separate step is that they are built using the builder image and then run on the resource image itself, since many of the tests depend on packages to exist on the image (they are more integration tests).

chenbh:
Slight modification to the release process:
- the ubuntu image will now also be pushed to Docker Hub under `ubuntu` tag
- `latest`, `alpine`, and the version tags will still refer to the alpine image

new pipeline:
image

cirocosta:
Hi @chenbh ,

That's still missing running the `pr` on the Ubuntu image, isn't it?

Thanks!

chenbh:
Hey @cirocosta, from what I understand we want to keep the changes minimal for now and keep all the default stuff on `alpine` and the `pr` job is only used for unit tests which _shouldn't_ (hopefully) be affected by base OS

Also I believe it would have to be the repo/org admin who would have to add the checks to GitHub itself

cirocosta:
> the `pr` job is only used for unit tests which shouldn't (hopefully) be affected by base OS

That's true, the rootfs should not influence much, but given that for a PR, that's the only place where we check if the image even builds, I'd say that it's good to ensure that both are able to at least build 🤔

E.g., let's assume that there's a change that includes something in `alpine` through `apk` and the person submitting the PR assumes that the same package can be obtained through the same name via `apt` - although the PR would look good, it'd take some time from the reviewer to validate that it's indeed able to do what it promises. Does that make sense?

Thanks!

chenbh:
@cirocosta good point! That's been fixed now
- the pr images will be tagged using `pr-alpine-` and `pr-ubuntu-` on docker hub
- expanded checks to be `status-alpine` and `status-ubuntu` (instead of `status`)

ci/pipelines/resources/template.jsonnet
```diff
@@ -36,7 +36,6 @@ local build_params =
   else if resource == "concourse-pipeline" then
     {
       build: ".",
-      dockerfile: resource+"-resource/Dockerfile"
```
vito: Don't need this; I ended up closing https://github.com/concourse/concourse-pipeline-resource/pull/52. Either way this value will still need to be set so it knows where the `Dockerfile` is.

chenbh: The `dockerfile` value would be taken from the default place which is at https://github.com/concourse/concourse/blob/dbaac83eea403e425f71ac6387303f5117da2296/ci/pipelines/resources/template.jsonnet#L247-L252

chenbh:
Since `template.jsonnet` is also used to generate the pipelines for `concourse-pipeline`, `mock`, and `datadog-event` resources would it make sense to have the same dockerfile changes there as well (there were PRs for `concourse-pipeline` and `datadog-event` but I wasn't sure about `mock`)?

ci/pipelines/resources/template.jsonnet
vito: ahh didn't notice that :+1:

vito:
@chenbh sure, might as well I guess if we're looking to standardize on ubuntu in the long run anyway. i'll reopen the other PR.

chenbh:
Created PRs for the 3 resources we don't ship but would still be impacted by the `template.jsonnet` change.

 I feel the `mock-resource` change was kind of overkill but the alternative is to add potentially hard to understand logic into the template.

- https://github.com/concourse/datadog-event-resource/pull/7
- https://github.com/concourse/concourse-pipeline-resource/pull/52
- https://github.com/concourse/mock-resource/pull/2 

Also updated previous comment about the list of PRs we have open

PR #3368, opened 2019-02-22
web: split up NotFound so it isn't a msg dependency anymore

xtreme-james-cooper:
Pure creating-new-files refactor, no code changes.


pivotal-jamie-klassen:
Folding into #3401

PR #3369, opened 2019-02-22
web: refactored the build folder to be better structured

xtreme-james-cooper:
Build.Models is huge and bloaty. We already have a divided Dashboard file, Build is the next largest. It should be moved into subfolders too, IMO.

Pure code movement, no code changes, let alone feature work.


vito:
folding into https://github.com/concourse/concourse/pull/3401

PR #3373, opened 2019-02-24
Remove worker/pool client

clarafu:
The worker.Client interface only existed because there were parts of the
code that wanted to choose a worker and then find or create a container
but there were other parts that wanted to use a specific worker and then
find or create container on that worker. There were these two paths but
in order to use the same code, both the pool and worker objects were
adjusted in order to fit this worker Client interface. This commit
removes this dependency by having the pool object find or choose a
worker and then on the worker object find or create a container on that
chosen worker.

Also, get containers are now being placed with the chosen placement
strategy instead of always randomly placing them.


cirocosta:
Hey @clarafu 

Is this PR still a thing? I couldn't immediately find any other issue that this one is blocked on 🤔

Thanks!

vito:
It's done, I think we just wanted to run TopGun against it and wait until 5.0 before merging.

atc/atccmd/command.go
```diff
@@ -500,13 +500,16 @@ func (cmd *RunCommand) constructAPIMembers(
...
 	}
 
+	resourceFactory := resource.NewResourceFactory()
 	dbResourceCacheFactory := db.NewResourceCacheFactory(dbConn, lockFactory)
-	resourceFetcherFactory := resource.NewFetcherFactory(lockFactory, clock.NewClock(), dbResourceCacheFactory)
+	fetchSourceFactory := resource.NewFetchSourceFactory(dbResourceCacheFactory, resourceFactory)
+	resourceFetcher := resource.NewFetcher(clock.NewClock(), lockFactory, fetchSourceFactory)
 	dbResourceConfigFactory := db.NewResourceConfigFactory(dbConn, lockFactory)
 	imageResourceFetcherFactory := image.NewImageResourceFetcherFactory(
-		resourceFetcherFactory,
```
topherbullock: YES! Get outta here `FetcherFactory`!

atc/atccmd/command.go
```diff
@@ -500,13 +500,16 @@ func (cmd *RunCommand) constructAPIMembers(
 		return nil, err
 	}
 
+	resourceFactory := resource.NewResourceFactory()
 	dbResourceCacheFactory := db.NewResourceCacheFactory(dbConn, lockFactory)
-	resourceFetcherFactory := resource.NewFetcherFactory(lockFactory, clock.NewClock(), dbResourceCacheFactory)
+	fetchSourceFactory := resource.NewFetchSourceFactory(dbResourceCacheFactory, resourceFactory)
```
topherbullock: Looks like the behavior in `FetchSource` could be collapsed into `Fetcher`:
- we only ever instantiate them in `atccmd`
- `FetchSource`'s interface is only relied on by `Fetcher`

That would be.. 
SO FETCH

topherbullock: Might also be feasible to roll all of this up into the current resource factory, as its only there as a boundary for tests.. but I wouldn't block merging this on that change!

ddadlani:
Agreed with @topherbullock - I don't like how the `resource` package looks right now but that can be refactored later. Looks good otherwise!

PR #3374, opened 2019-02-24
ci: enable global resources on wings, bump version

vito:


vito:
(merging this since it's actually already been configured)

PR #3382, opened 2019-02-25
atc: add indexes for build input/output joins

vito:
these indexes greatly improve the time taken to load a pipeline's versions db


vito:
@pivotal-jamie-klassen It looks like the migration check is failing again, but for the opposite reason it failed before - now it *wants* a trailing linebreak in the `$expected_pr_migrations` file.

I think this is actually because if `$new_on_pr` is empty it shouldn't echo at all, but if it's non-empty it should echo with a trailing linebreak. It must have failed before (with no migrations added) because it would echo an empty string with a trailing linebreak when it should have just echoed nothing at all.

I'll just sneak the fix into my PR so the build goes green.

PR #3392, opened 2019-02-26
Build plan visibility

### LINKED ISSUE #2116: Logged out users can see build tasks
* Concourse version: 3.9.2

We noticed that when we refreshed the page on a build while logged out of our Concourse, we were able to briefly see the job information. This seems like a bug since once the page finishes loading it shows us the "Log in to view" message instead.

concourse-bug

pivotal-jwinters:
Don't show the build plan for private jobs.

fixes #2116


pivotal-jwinters:
Over to you @pivotal-jamie-klassen @xtreme-james-cooper

web/elm/src/Build/Build.elm
```diff
@@ -728,149 +805,176 @@ viewBuildPage model =
...
+
+body :
+    { currentBuild : Models.CurrentBuild
+    , authorized : Bool
+    , showHelp : Bool
+    }
+    -> Html Msg
+body { currentBuild, authorized, showHelp } =
+    Html.div [ class "scrollable-body build-body" ] <|
+        if authorized then
```
xtreme-james-cooper: Breaking this garbage 10,000-line function up. YESSSSSSSSSS

web/elm/src/Build/Output/Output.elm
```diff
@@ -100,60 +100,37 @@ handleStepTreeMsg action model =
 
 planAndResourcesFetched :
     Concourse.BuildId
-    -> Result Http.Error ( Concourse.BuildPlan, Concourse.BuildResources )
+    -> ( Concourse.BuildPlan, Concourse.BuildResources )
```
xtreme-james-cooper: ...Hrmmm. I get that the tuple is easier to pass from the caller, it just seems... odd.

pivotal-jwinters:
@pivotal-jamie-klassen mind rebasing this? The golang merge is pretty straight forward. I got part way through the elm stuff and wasn't sure what needed to be kept.

PR #3393, opened 2019-02-27
need to pull concourse/dev sometimes

ralekseenkov:
Improved instructions, so people have easier time rebuilding Concourse from source. We ran into this just recently (i.e. the need to pull `concourse/dev`).


Twiknight:
vote this up.

It waste me hours to find out why my changes would never take effect after the binary path moved in `concoruse/dev`

Issue #429, opened 2016-05-05
make urls in metadata clickable


cppforlife:
auto link any link looking things in resource metadata shown on build page.

xoebus:
We had some forward thinking links and markdown conventions for this at some point. Maybe those didn't make it.

evanchaoli:
Any progress on this? This is a feature we are looking for. I was going to open an issue for this, and then found this existing one has been here for long time. My team is willing to contribute to this feature.

PR #3398, opened 2019-02-28
make values starts with https or http clickable in build metadata

Twiknight:
A partial implementation for #429

It makes all strings that looks like a url in area `build -> resource -> metadata` clickable. 

This is very useful when your resource returns information from an external website, for example GitHub or GitLab like below case:

image

- It only handles url-like strings (String starts with "http://" or "https://"), since in elm 0.19 we'll have official url lib, it would be duplication to implement https://tools.ietf.org/html/rfc3986 here 

- It only works in build page with the resource metadata. The reason is, to make all url-like stuff clickable would be larger change that impacts several files and ten more functions. I chose to start with build metadata, where in most cases, you would want a clickable link. When we decide this is a proper way or find a better way, more changes can be applied to the rest parts.

### DIFF
--- a/web/elm/src/Build/StepTree/StepTree.elm
+++ b/web/elm/src/Build/StepTree/StepTree.elm
@@ -42,7 +42,7 @@ import Dict exposing (Dict)
 import DictView
 import Effects exposing (Effect(..))
 import Html exposing (Html)
-import Html.Attributes exposing (attribute, class, classList, href, style)
+import Html.Attributes exposing (attribute, class, classList, href, style, target)
 import Html.Events exposing (onClick, onMouseDown, onMouseEnter, onMouseLeave)
 import Routes exposing (Highlight(..), StepID, showHighlight)
 import Spinner
@@ -574,9 +574,25 @@ viewVersion version =
 
 viewMetadata : List MetadataField -> Html Msg
 viewMetadata =
-    List.map (\{ name, value } -> ( name, Html.pre [] [ Html.text value ] ))
-        >> Dict.fromList
-        >> DictView.view []
+        List.map
+            (\{ name, value } ->
+                ( name
+                , Html.pre []
+                    [ if String.startsWith "http://" value || String.startsWith "https://" value then
+                        Html.a
+                            [ href value
+                            , target "_blank"
+                            , style [ ( "text-decoration-line", "underline" ) ]
+                            ]
+                            [ Html.text value ]
+
+                      else
+                        Html.text value
+                    ]
+                )
+            )
+            >> Dict.fromList
+            >> DictView.view []
 
 
 viewStepState : StepState -> Html Msg
--- a/web/elm/tests/BuildTests.elm
+++ b/web/elm/tests/BuildTests.elm
@@ -1722,6 +1722,141 @@ all =
                             >> Query.has [ attribute <| Attr.src url ]
                     ]
                 ]
+            , describe "get step with metadata" <|
+                let
+                    httpURLText =
+                        "http://some-url"
+
+                    httpsURLText =
+                        "https://some-url"
+
+                    plainText =
+                        "plain-text"
+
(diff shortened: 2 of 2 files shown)


Twiknight:
@pivotal-jamie-klassen 
Thanks for the comments and the back-fill patch on my last pr. 
Actually I felt strange such it did not break any case when I was creating these patches.

I'll try to cover the changed lines with tests. But it might take some time for me to understand the test framework. (And your back-fill patch looks like a helpful example on my first sight : ) ).

I'll come back to you when I have some updates or I need some help.

pivotal-jamie-klassen:
@Twiknight glad to hear it! While you're looking at my follow-up PR maybe using `elm-format` will help you to more easily interpret the changes (as a lot of them were whitespace-related): https://github.com/concourse/concourse/pull/3425

vito:
folding into https://github.com/concourse/concourse/pull/3401

pivotal-jamie-klassen:
whoops, we told @vito the wrong PR.

Twiknight:
@pivotal-jamie-klassen I've  nearly done, but the test case looks horrible in code smell.
The problem is, I tried to pass several urls to the test function, but I could not find a utility in `elm-test` that applies to this case.

Currently it looks like below:
```
, test "get step metadata should show hyperlink on url-like content" <|
    \() -> Expect.all
        [ (\f -> f "http://metadata-url.com" ())
        , (\f -> f "https://metadata-url.com" ())
        ]
        (\url ->
            fetchPlanWithGetStep
                >> Build.update
                    (Build.Msgs.BuildEventsMsg <|
                        Build.Msgs.Events <|
                            Ok <|
                                Array.fromList
                                    [ Models.FinishGet
                                        { source = "stdout", id = "plan" }
                                        0
                                        (Dict.fromList [ ( "foo", "bar" ) ])
                                        [ { name = "url", value = url } ]
                                    ]
                    )
                >> Tuple.first
                >> Build.update (Build.Msgs.ToggleStep "plan")
                >> Tuple.first
                >> Build.view UserState.UserStateLoggedOut
                >> Query.fromHtml
                >> Query.find
                    [ containing [ text url ]
                    ]
                >> Query.has
                    [ tag "a"
                    , style [ ( "text-decoration-line", "underline" ) ]
                    , attribute <| Attr.target "_blank"
                    , attribute <| Attr.href url
                    ]
        )
```

Any ideas how can I make it readable?

pivotal-jamie-klassen:
@Twiknight in this case I would just duplicate the assertions, i.e. (`test "get step metadata should show hyperlink on value starting with 'http://'"` and `test "get step metadata should show hyperlink on value starting with 'https://'"`) as both cases impact the implementation logic. Maybe a case for *not* showing a link when the value starts with something else, too. I'd make the case here for [preferring duplication over the wrong abstraction](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction).

pivotal-jamie-klassen:
@Twiknight awesome!

evanchaoli:
@vito This code change has been approved, can you please help merge and include it in 5.0?

vito:
Merged a refactor and now there are merge conflicts here - sorry! Hopefully they're not too much of a pain to resolve.

@evanchaoli This didn't make it in to 5.0, so it'll be in the next release (5.1).

evanchaoli:
@vito 5.1 is fine. Thanks a lot.

Twiknight:
After the rebasing, seems the cases broke, I'm fixing and will update later

Twiknight:
@pivotal-jamie-klassen @vito ,

I've finished rewritten the cases to match current paradigm of data flow.

Any idea why a UI change would fail a Go test?

vito:
@Twiknight Looks like a flake. I've retriggered it and PR'd a fix for the flake: https://github.com/concourse/concourse/pull/3516

PR #3399, opened 2019-02-28
Vault timeout should not cause overall Concourse slowness

### LINKED ISSUE #3331: Job build takes a long time to start when Vault is not available
## Bug Report

If previously "+" button would instantly spin up a new build, now we are seeing CreateJobBuild takes exactly 5 seconds. You can see that in Chrome Developer Console, as well as in Grafana.

## Steps to Reproduce

Basically happens on any pipeline.

## Version Info

* Concourse version: 5.0 RC
* Deployment type (BOSH/Docker/binary): Kubernetes
* Infrastructure/IaaS: Azure AKS
* Did this used to work? Yes

If there is a way to profile it and capture some tracing info to see what takes long time, we are happy to help figure this out.

ralekseenkov:
Fixes #3331 by deferring Vault lookup to when it's really needed.

So that if Vault is not available:
* we don't slow down Concourse pipelines which don't require Vault 
* a meaningful error message is reported if Vault lookup fails

### DIFF
--- a/atc/creds/vault/vault.go
+++ b/atc/creds/vault/vault.go
@@ -1,10 +1,12 @@
 package vault
 
 import (
-	"path"
-
+	"errors"
+	"fmt"
 	"github.com/cloudfoundry/bosh-cli/director/template"
 	vaultapi "github.com/hashicorp/vault/api"
+	"path"
+	"time"
 )
 
 // A SecretReader reads a vault secret from the given path. It should
@@ -16,6 +18,7 @@ type SecretReader interface {
 // Vault converts a vault secret to our completely untyped secret
 // data.
 type Vault struct {
+	LoggedIn     <-chan struct{}
 	SecretReader SecretReader
 
 	PathPrefix   string
@@ -59,10 +62,27 @@ func (v Vault) Get(varDef template.VariableDefinition) (interface{}, bool, error
 	return evenLessTyped, true, nil
 }
 
+func (v Vault) getSecretReader() (SecretReader, error) {
+	// this will wait for up to 5 seconds, until login into Vault gets established
+	select {
+	case <-v.LoggedIn:
+		// if login into Vault is successful, v.LoggedIn channel will get closed
+		// therefore this select won't block anymore since successful login
+	case <-time.After(5 * time.Second):
+		return nil, errors.New("timeout connecting to vault")
+	}
+	return v.SecretReader, nil
+}
+
 func (v Vault) findSecret(path string) (*vaultapi.Secret, bool, error) {
-	secret, err := v.SecretReader.Read(path)
+	sr, err := v.getSecretReader()
 	if err != nil {
-		return nil, false, err
+		return nil, false, fmt.Errorf("unable to retrieve '%s': %s", path, err)
+	}
+
+	secret, err := sr.Read(path)
+	if err != nil {
+		return nil, false, fmt.Errorf("unable to retrieve '%s': %s", path, err)
 	}
 
 	if secret != nil {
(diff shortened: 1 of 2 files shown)


atc/creds/vault/vault.go
```diff
@@ -59,10 +65,31 @@ func (v Vault) Get(varDef template.VariableDefinition) (interface{}, bool, error
 	return evenLessTyped, true, nil
 }
 
+func (v Vault) getSecretReader() (SecretReader, error) {
+	v.once.Do(func() {
```
vito: At first I thought the `v.once` was unnecessary but it actually probably helps to not construct the `time.After`. 👍

atc/creds/vault/vault.go
```diff
@@ -59,10 +65,31 @@ func (v Vault) Get(varDef template.VariableDefinition) (interface{}, bool, error
...
+		select {
+		case <-v.LoggedIn:
+		case <-time.After(5 * time.Second):
+			v.timeout = true
```
vito: `v` is a non-pointer here so I think this assignment won't outlive this method call. That's fine for the first call but later calls won't run the `v.once` and so `v.timeout` will be false. Should this be changed to a pointer receiver?

ralekseenkov: Turns out timeout was only used in this method. So kept `v` as non-pointer and converted `timeout` into a local variable instead

vito: Hmm, that makes the assignment clearer but it'll still have the behavior of only attempting to wait on the log-in one time, and any subsequent calls will just start returning the secret reader even though it hasn't logged in yet (since `timeout` will always be `false`).

I guess I'm not clear on the flow here. There's also the fact that there's a new variables object constructed pretty often, so the `.once` is still gonna be evaluated each time. Maybe it should just be removed.

ralekseenkov: @vito here is the summary how I see it:
* getSecretReader() gets called every time a secret is retrieved
* when is gets called for the first time, we'd like to ensure that connection to vault has been established successfully, and give it up to 5 seconds
* if connection has been established successfully, subsequent calls should immediately return the corresponding initialized SecretReader
* if connection has not been established, subsequent calls should return immediately as well (it will be an un-authorized client / SecretReader and all secrets lookup will fail, which seems ok)

The main reason why sync.Once got added was not to avoid creation of time.After, but rather to account for the fact that select (without default clause) is blocking. Without sync.Once, select will read the status off v.LoggedIn for the first time, and then subsequent calls to select would block and wait once again for the timeout duration 

It's probably not obvious from the code, but I'm open to either adding more comments or possibly changing the implementation to make the code more readable.

vito: @ralekseenkov Ahh, now I understand the intent. `LoggedIn` is actually *closed* on the first successful log-in, though, not written to on each log-in:

https://github.com/concourse/concourse/blob/996b8d1671411d6d6a80558366bb98958b338f00/atc/creds/vault/reauther.go#L98-L100

So it shouldn't block, because closed channels always return their zero-value.

I think without the `Once` it'll wait up to 5 seconds on each call until the first successful log-in, and at that point it'll stop waiting as the `select` should just immediately unblock. Does that make sense? (Kinda hard to put this into words.)

ralekseenkov:
changed the code according to our conversation on discord
tested locally via docker compose, seems to work just fine
error message is slightly verbose, but should be ok for now
Screen Shot 2019-03-27 at 1 22 14 PM

vito:
@ralekseenkov Woops, I forgot about this and now there are conflicts from the other PR. Where would you like to go from here?

ralekseenkov:
let's abandon this change for now. if people run into this issue in the field, we can get this fixed later

PR #3401, opened 2019-02-28
Refactor/subscription ungenerify

pivotal-jamie-klassen:
Subscriptions are hard to test, because they all have functions living inside them. And their responses are all mixed up with regular active message responses.

We made subscriptions flat, made them return a flat `Delivery` type instead of the previous `Application.Msg` type, and also refactored the handling of effects in the subpages somewhat to make it easier to compose for instance the top bar.


pivotal-jamie-klassen:
Folded into #3428

PR #3402, opened 2019-02-28
skymarshal: make user claims configurable

### LINKED ISSUE #2683: Make 'username' claim name configurable for OAuth and OIDC
Per conversation with @pivotal-jwinters 

WIth Azure OIDC, username is always empty in the logs when user does login into Concourse.

There is already "oauth-auth-groups-key" and "oidc-auth-groups-key" parameters for how the "group" claim is called (default should be "groups").

For consistency, it might make sense to add "oauth-auth-username-key" and "oidc-auth-username-key" parameters for the "username" claim (default should be "username").

For Azure, we'd prob need to set it to "name" or "preferred_username". Here is the list of supported claims:
```
"claims_supported":["sub","iss","cloud_instance_name","cloud_instance_host_name","cloud_graph_host_name","msgraph_host","aud","exp","iat","auth_time","acr","nonce","preferred_username","name","tid","ver","at_hash","c_hash","email"]
```
Needs a Concourse change + Helm chart change.

Likely not. If anything, we are more interested in stability (we have issues related to resource checks hanging, worker stability, etc). Trying to reproduce other higher-priority Concourse issues right now.

pivotal-jwinters:
* You are able to configure user name for OIDC
* You are able to configure user name and ID for OAuth
* This commit also bumps dex

fixes #2683


vito:
What is the default behavior if these flags aren't specified?

pivotal-jwinters:
For OAuth: https://github.com/concourse/dex/blob/master/connector/oauth/oauth.go#L174-L180

For OIDC: https://github.com/concourse/dex/blob/master/connector/oidc/oidc.go#L259-L261

It's unfortunate that `user_name` and `username` are different, but for backwards compatibility we're stuck with those keys.

pivotal-jwinters:
Relevant commits to dex:
OIDC https://github.com/concourse/dex/commit/4530f12257d62a595fbc925d2224c04bcf767fb7
OAuth https://github.com/concourse/dex/commit/cb3c0d15de8f787591e37ce69bfa5dd4e22e9a80

pivotal-jwinters:
@vito yup makes sense. I'll add the defaults and rebase.

go.mod
```diff
@@ -47,16 +46,16 @@ require (
 	github.com/cloudfoundry/go-socks5 v0.0.0-20180221174514-54f73bdb8a8e // indirect
 	github.com/cloudfoundry/socks5-proxy v0.0.0-20180530211953-3659db090cb2 // indirect
 	github.com/cockroachdb/cmux v0.0.0-20170110192607-30d10be49292 // indirect
+	github.com/concourse/baggageclaim v1.3.5
```
vito: don't think this should be here (there's a 1.4.0 dependency on it below)

vito: actually I'll just fix this myself :sunglasses:

cappyzawa:
@pivotal-jwinters I am not familiar with certification, but I have one question.
Why is `UserIDKey` added to `skymarshal/skycmd/oauth_flags.go` but not to `skymarshal/skycmd/oidc_flags.go`?

I'm struggling because we need to set the value of the subject claim in oidc user.
If there is no problem even if `UserIDKey` is set, it is easy.

vito:
@cappyzawa Would you mind opening a new issue? It's hard to keep track of comments in closed/merged PRs.

cappyzawa:
@vito OK! I'll open a new issue😁

PR #3410, opened 2019-03-01
add 'shipit' and publish-* jobs

vito:
this is kinda important for 5.0 :stuck_out_tongue:


ci/pipelines/concourse.yml
```diff
@@ -782,20 +798,174 @@ jobs:
...
+    - bosh-topgun
+  - get: linux-rc
+    passed: [build-rc]
+  - get: windows-rc
+    passed: [build-rc]
+  - get: darwin-rc
+    passed: [build-rc]
+  - get: concourse-rc-image
+    passed: [k8s-topgun]
+  - get: charts
```
cirocosta: Hey, could you clarify why there's the need of having a `get` on `charts`? Is it just to pin down which version was used at the time?

Thanks!

ci/pipelines/concourse.yml
```diff
@@ -782,20 +798,174 @@ jobs:
...
 
-- name: push-docs
+- name: shipit
   public: true
+  serial_groups: [version]
+  plan:
+  - get: concourse
+    passed:
+    - build-rc
+    - k8s-topgun
```
cirocosta: neat! I like the fact that we end up having the `concourse-rc` image tested running in a containerized env before promoting it 

screen shot 2019-03-01 at 11 13 04 am

ci/pipelines/concourse.yml
```diff
@@ -782,20 +798,174 @@ jobs:
...
+      passed: [shipit]
+    - get: cbd
+    - get: version
+      passed: [shipit]
+    - get: bpm-release
+      passed: [shipit]
+    - get: postgres-release
+      passed: [shipit]
+  - task: bump-versions
+    file: cbd/ci/bump-versions.yml
```
cirocosta: Do we need to update https://github.com/concourse/concourse-bosh-deployment/blob/f26db87ba8837ef73935fc33f3038dcb0874a363/ci/scripts/bump-versions#L12-L19?

It still references `garden_runc` there 🤔

ci/pipelines/concourse.yml
vito: Yeah, the idea is for `shipit` to be the gateway and mark everything that we want to ship so that downstream publishing jobs know exactly what version to use. We don't currently do anything with `charts` but I figured we might at least want to know the sha that made it through when it comes time to manually publish.

ci/pipelines/concourse.yml
vito: Already updated on `develop`:

https://github.com/concourse/concourse-bosh-deployment/blob/develop/ci/scripts/bump-versions#L12-L19

vito:
@cirocosta Sounds good, I was thinking about doing that before actually. Do you think we can do the same for k8s? (k8s-check-props -> k8s-smoke)

vito:
Pushed aforementioned changes, also had k8s gate on the check-params job.

Here's how it all looks:

https://ci.concourse-ci.org/teams/main/pipelines/concourse2?groups=k8s&groups=bosh&groups=develop

PR #3415, opened 2019-03-01
topgun/k8s: add test to verify connectivity to external pg

### LINKED ISSUE #3106: topgun/k8s: Use of external postgresql database
With `hush-house` moving toward the use of an external PostgreSQL instead of the built-in Postgres dependent chart, It'd be good to have a test that attests the possibility of using external Postgres instances, especially the scene where one needs to provide some extra options regarding TLS (ca cert, etc).

Using the `stable/postgresql` for that (deploying it separately), we can also ensure that x-compatibility with such chart is achieved.

cirocosta:
fixes #3106

### DIFF
--- a/topgun/k8s/external_postgres_test.go
+++ b/topgun/k8s/external_postgres_test.go
@@ -50,7 +50,18 @@ var _ = Describe("External PostgreSQL", func() {
 		Wait(Start(nil, "kubectl", "delete", "namespace", namespace, "--wait=false"))
 	})
 
-	It("gets deployed", func() {
+	It("can have pipelines set", func() {
 		waitAllPodsInNamespaceToBeReady(namespace)
+
+		By("Creating the web proxy")
+		proxySession, atcEndpoint := startPortForwarding(
+			namespace, "service/"+releaseName+"-web", "8080")
+		defer proxySession.Interrupt()
+
+		By("Logging in")
+		fly.Login("test", "test", atcEndpoint)
+
+		By("Setting and triggering a dumb pipeline")
+		fly.Run("set-pipeline", "-n", "-c", "../pipelines/get-task.yml", "-p", "pipeline")
 	})
 })
--- a/topgun/k8s/external_postgres_test.go
+++ b/topgun/k8s/external_postgres_test.go
@@ -1,0 +1,56 @@
+package k8s_test
+
+import (
+	"fmt"
+	"path"
+
+	. "github.com/concourse/concourse/topgun"
+	. "github.com/onsi/ginkgo"
+)
+
+var _ = Describe("External PostgreSQL", func() {
+	var (
+		releaseName   string
+		pgReleaseName string
+		namespace     string
+	)
+
+	BeforeEach(func() {
+		releaseName = fmt.Sprintf("topgun-ep-%d-%d", GinkgoRandomSeed(), GinkgoParallelNode())
+		namespace = releaseName
+		pgReleaseName = releaseName + "-pg"
+
+		helmDeploy(pgReleaseName,
+			namespace,
+			path.Join(Environment.ChartsDir, "stable/postgresql"),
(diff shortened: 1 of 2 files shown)


topgun/k8s/external_postgres_test.go
```diff
@@ -0,0 +1,56 @@
...
+	})
+
+	AfterEach(func() {
+		helmDestroy(releaseName)
+		helmDestroy(pgReleaseName)
+		Wait(Start(nil, "kubectl", "delete", "namespace", namespace, "--wait=false"))
+	})
+
+	It("gets deployed", func() {
+		waitAllPodsInNamespaceToBeReady(namespace)
```
YoussB: I am not sure, should we add more test to make sure that interaction with the db is working correctly? 
like setting a pipeline and getting the pipeline config?

topgun/k8s/external_postgres_test.go
```diff
@@ -0,0 +1,56 @@
...
+			"--set=livenessProbe.periodSeconds=3",
+			"--set=persistence.enabled=false",
+			"--set=postgresqlDatabase=pg-database",
+			"--set=postgresqlPassword=pg-password",
+			"--set=postgresqlUsername=pg-user",
+			"--set=readinessProbe.initialDelaySeconds=3",
+			"--set=readinessProbe.periodSeconds=3",
+		)
+
+		deployConcourseChart(releaseName,
```
YoussB: might not be super relevant, should we move this part to the suite, since that is the basic deploy that we do in almost every test?

cirocosta: Hey, what do you mean by `this part`? The `releaseName` only, or the other settings?

topgun/k8s/external_postgres_test.go
cirocosta: Hey, that's seems like a pretty valid concern; my rationale was around the fact that `web` won't get ready (per the [`readinessProbe`](https://github.com/helm/charts/blob/7e9730490ac0aacabfe5236e60dec3f7b2653051/stable/concourse/values.yaml#L1124-L1127)) before all connectivity gets established (which includes performing the migrations as no endpoints will respond before that)  🤔 

To be more thorough then, should we:

- [’logging in’](https://github.com/concourse/concourse/blob/bf12d2f50c6b5d940bb80b60299f30bb10185f2d/topgun/k8s/kubernetes_creds_mgmt_test.go#L36-L37), and then
- [`set-pipeline`](https://github.com/concourse/concourse/blob/bf12d2f50c6b5d940bb80b60299f30bb10185f2d/topgun/k8s/kubernetes_creds_mgmt_test.go#L81)?

As I see, if nothing errors on such path, we should be good.

Wdyt?

thx!

topgun/k8s/external_postgres_test.go
YoussB: I meant deploying concourse chart.
my thought was that we create a `justbeforeeach` in the suite that deploys concourse chart using whatever params that we set here.

does this make any sense?

topgun/k8s/external_postgres_test.go
YoussB: Yeah, I didn't really think about the migrations. Maybe we don't need that?

topgun/k8s/external_postgres_test.go
cirocosta: hmmmm I see; my initial thoughts were more around not having too many things global (in `k8s_suite.go`) to avoid having to know what are the right variables to set when creating a new test 🤔 

Would you mind leaving that discussion to a separate issue? There we could elaborate more on it.

thanks!!

topgun/k8s/external_postgres_test.go
cirocosta: 👍 I'll include an extra step then!

cirocosta:
Updated @YoussB 😁

PR #3422, opened 2019-03-04
topgun/k8s: refactoring + disabling of persistence in tests

cirocosta:
Hey,

This PR is just about a quick refactoring regarding our k8s tests.

I noticed that in our tests we didn't have the `persistence.enabled` set up to be false, and we could leverage the `deployConcourseChart` in `baggageclaim_drivers_test` as there are no big specifics tunings there that can't be done with `deployConcourseChart`.

This PR also deals with the fact that some times the port forwarding might take some time to get set up, so it increases the `Eventually` timeout for such thing so that we might be able to have some less flakiness there.

Thanks!

---

Update: before merging - with two other PRs having been merged before this, it'd be good to have those tests updated to not use persistence as well as they're not testing lifecycle that requires persistence.

### DIFF
--- a/topgun/k8s/baggageclaim_drivers_test.go
+++ b/topgun/k8s/baggageclaim_drivers_test.go
@@ -47,6 +47,7 @@ var _ = Describe("Baggageclaim Drivers", func() {
 				"--set=concourse.web.kubernetes.enabled=false",
 				"--set=concourse.worker.baggageclaim.driver=" + c.Driver,
 				"--set=image=" + Environment.ConcourseImageName,
+				"--set=persistence.enabled=false",
 				"--set=postgresql.persistence.enabled=false",
 				"--set=web.livenessProbe.failureThreshold=3",
 				"--set=web.livenessProbe.initialDelaySeconds=3",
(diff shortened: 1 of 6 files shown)


cirocosta:
cc @YoussB

cirocosta:
>  The probelm is with using --wait=false. sometimes deleting a namespace conflicts with another namespace being deleted. At least this is how I understand it

Yeah ... I'm wondering whether we should instead just rely on an `Eventually` then 🤔 

The rationale for always cleaning the namespace is `helm delete` won't delete the namespace (see discussion here: https://github.com/helm/helm/issues/1464) as that'd mean that things like `StatefulSets`'s PVCs would be destroyed too (which goes against the idea of leaving PVCs around regardless of the pod still being there or not - see https://github.com/helm/helm/issues/1464#issuecomment-455320037)

Wdyt of using `Eventually` there then?

Thanks!

cirocosta:
With the many changes that have been happening lately, I'll close this in favor of some work that @YoussB will be doing regarding fixing some of those problems 👍  

Thanks!

PR #3425, opened 2019-03-04
web: add 'yarn format' to format Elm code

pivotal-jamie-klassen:
Also ran it across the codebase and it wound up organizing some imports.


xtreme-james-cooper:
LGTM

PR #3428, opened 2019-03-04
web: use a port for buildevents

### LINKED ISSUE #3229: Refactor build events to use a port instead of an effect manager
splitting out a sub-task of #3155.

pivotal-jamie-klassen:
Fixes #3229


pivotal-jamie-klassen:
Somewhere deep in these changes I think there's a regression -- now when I'm on the build page and I hit Command-L to focus the URL bar, I actually move backward in the build history.

vito:
@pivotal-jamie-klassen Would you like me to hold off on merging then or fix that in post? There are a few PRs based on this now so it might be easier to just merge and open a bug so they don't all have to be rebased.

pivotal-jamie-klassen:
@vito let's merge + open a bug. We've been thinking about using something other than the default Keyboard.downs subscription for keypresses anyway.

Issue #1971, opened 2018-01-20
Hide resource metadata for exposed pipeline


tenjaa:
# Feature Request

## What challenge are you facing?

We want to setup a build monitor that shows the status of our pipelines on a big screen.
Our solution was to get Raspeberry Pi, install dashkiosk and then just iterate through the pipelines we care about.
To be able to see our pipelines, we exposed them.

The problem is that with exposing the pipeline, we also expose all metadata of our resources. So emails, names and maybe some commits we do not want to be publicly available.

## A Modest Proposal

Add the option to not show metadata of resources when exposing pipelines:

`fly -t ci expose-pipeline -p my-pipeline -v "metadata=false"`

xtreme-conor-nosal:
There may be a use case for exposing metadata on a per-resource basis. The sample use case for `fly expose-pipeline` is for a contributor to an open source repo to be able to track their change through the project's pipeline. As an all-or-nothing flag, that visibility can't be provided without also exposing commits from other repos used by the pipeline, any buckets used to store artifacts, any servers backing a resource-pool, etc. There may be 'private infrastructure' behind the public pipeline.

xtreme-conor-nosal:
There are also security concerns around exposed metadata:
- I can see all the s3/gcp buckets in the pipeline, so I can check if any have misconfigured permissions
- resource pool metadata can expose server urls, if there's a CVE I can try to break in
- a resource maintainer may accidentally include a secret in metadata
- etc

vito:
I think this makes sense, at least as an option to set somehow. I think there are definitely use cases for showing it (i.e. pipelines for open-source projects) but I see why you'd want to hide it, too.

Maybe this is another property of the pipeline? Maybe you could mark resources `public: true` just like jobs, and hide metadata by default?

/cc @pivotal-jwinters - not sure if this is interesting for RBAC

mhuangpivotal:
We are thinking of adding a `public` property for resource as mentioned above. When this property is not set or set to `false`, we can hide either the entire resource page or just the metadata (so someone could still see resource versions and builds info). Would it make sense to show the resource versions regardless?

The metadata for the build inputs will still be visible if the job is set to `public: true`. This can be mitigated by setting `public: false` for the job. Would there a valid use case where the job is desired to be public but not the input metadata?

vito:
@Lindsayauchin We're gonna start off by just showing nothing, but maybe we want a UI to indicate that this information has been omitted because the user doesn't have access?

[...]

PR #3437, opened 2019-03-06
Hide resource metadata for exposed pipeline

mhuangpivotal:
* Added `Public` to resource config. Resource metadata is no longer returned in resource versions endpoint when `Public` is set to false or not set and the user is not authenticated.

#1971

### DIFF
--- a/atc/api/resourceserver/versionserver/list.go
+++ b/atc/api/resourceserver/versionserver/list.go
@@ -8,6 +8,8 @@ import (
 
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/atc/api/accessor"
+	"github.com/concourse/concourse/atc/api/present"
 	"github.com/concourse/concourse/atc/db"
 )
 
@@ -87,6 +89,11 @@ func (s *Server) ListResourceVersions(pipeline db.Pipeline) http.Handler {
 		w.Header().Set("Content-Type", "application/json")
 
 		w.WriteHeader(http.StatusOK)
+
+		acc := accessor.GetAccessor(r)
+		hideMetadata := !resource.Public() && !acc.IsAuthenticated()
+
+		versions = present.ResourceVersions(hideMetadata, versions)
 
 		err = json.NewEncoder(w).Encode(versions)
 		if err != nil {
--- a/atc/api/resourceserver/versionserver/list.go
+++ b/atc/api/resourceserver/versionserver/list.go
@@ -91,7 +91,7 @@ func (s *Server) ListResourceVersions(pipeline db.Pipeline) http.Handler {
 		w.WriteHeader(http.StatusOK)
 
 		acc := accessor.GetAccessor(r)
-		hideMetadata := !resource.Public() && !acc.IsAuthenticated()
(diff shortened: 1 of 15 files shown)


mhuangpivotal:
@pivotal-jamie-klassen The second part of this [issue](#1971) will be an indication in the UI. Do we want to create a separate issue for the UI component?

vito:
@mhuangpivotal What is the delta so far? Looks like it just `nil`s out metadata?

I'm fine with shipping that as the first step and doing the UX stuff after.

mhuangpivotal:
@vito that's right. The delta adds `Public` to resource config and `nil`s out metadata in the list resources API endpoint.

It would be good to have a separate issue to track the UI change since it'll require design.

I'm going to mark this PR ready for review.

atc/api/resourceserver/versionserver/list.go
```diff
@@ -88,6 +90,11 @@ func (s *Server) ListResourceVersions(pipeline db.Pipeline) http.Handler {
 
 		w.WriteHeader(http.StatusOK)
 
+		acc := accessor.GetAccessor(r)
+		hideMetadata := !resource.Public() && !acc.IsAuthenticated()
```
vito: `acc.IsAuthenticated()` isn't enough - I need to also be a member of the team. This just checks that I'm logged in at all.

mhuangpivotal: I think the authorization check happens in the auth handler:

https://github.com/concourse/concourse/blob/03b05e7fe33033cfcaa0a41aff634293943ac8e0/atc/wrappa/api_auth_wrappa.go#L94-L95

At the this point, the user is either not logged in or logged in to the correct team, and we only want to hide metadata if the resource is private and the user is not logged in. We didn't like we had to get the accessor here to check if the user is logged in, is there a better way to get the information?

vito: I could also be logged in and able to view the pipeline because it's exposed - in that case the metadata should be hidden. I think you can just use `IsAuthorized`: https://github.com/concourse/concourse/blob/03b05e7fe33033cfcaa0a41aff634293943ac8e0/atc/api/accessor/accessor.go#L13

mhuangpivotal: Ah, that makes sense. Pushed a fix.

atc/api/versions_test.go
```diff
@@ -86,12 +85,128 @@ var _ = Describe("Versions API", func() {
...
+								{
+									"id":2,
+									"enabled": false,
+									"version": {"some":"version"}
+								}
+							]`))
+						})
+					})
+
+					Context("when the user is authenticated", func() {
```
vito: do we need a test for `IsAuthorized` now? these both just show that it doesn't include version metadata, but we do want to include it if they're authorized

mhuangpivotal: These two tests are under the context "when not authorized". The "authorized" case is in a different context:

https://github.com/concourse/concourse/blob/7f05064caefc52487f95b30b695bb9c37ccfa601/atc/api/versions_test.go#L213

vito: ahh, makes sense - thanks!

Issue #2833, opened 2018-11-19
Worker should sweep containers and volumes in parallel


vito:
## What challenge are you facing?

If a worker has a bunch of containers and volumes coming and doing, the sweeper can be too slow to keep up.

## What would make this better?

Right now we delete all the containers one by one, and then delete all the volumes:

https://github.com/concourse/concourse/blob/2a2e30516336ab09974939572f7c45af82c3e4b0/worker/sweep_runner.go#L77-L97

We can/should do this in parallel with a reasonable max-in-flight for the containers.

There might not be much point in deleting the volumes themselves in parallel to each other, but we should at least be able to do it while we're destroying containers.

Note: if we *do* want to delete volumes concurrently to each other, we'll need to change BaggageClaim, since right now that's a single API call with a big list of handles.

vito:
FWIW it looks like we do destroy volumes in parallel, though I kind of wonder how effective that is and whether it's just unnecessary strain on disk. Would be nice to measure that.

https://github.com/concourse/baggageclaim/blob/8c6fe76a80e6bf0a5c8e651bea8877f32b98c2ae/api/volume_server.go#L257

topherbullock:
Oof, especially given some of @kcmannem 's discoveries around btrfs' ability to keep up in terms of block gc and fragmentation of the volumes we're throwing at it, and the frequency of volume creation / COW creation / deletion Concourse throws at it.

Maybe we should add some bounds to the parallelism here, rather than wildly spawning go routines

vito:
Yeah definitely.

Also: these should probably be parallelized on their own intervals. An implementation that simply GCs containers and volumes in parallel and waits for both to complete before the next tick could still end up slowing down container or volume GC if either end is slow enough. It would be better if there was a separate container GC interval and a separate volume GC interval.

ddadlani:
Containers and volumes now each have their own sweeper. Each sweeper still runs on a 30s interval, but since they are decoupled, volume deletion does not wait on container deletion and vice versa. 

In addition, volumes and containers are also deleted in parallel. We made changes to baggageclaim to expose the `DestroyVolume` method to the API to facilitate individual volume deletion.

[...]

PR #3440, opened 2019-03-06
Parallelize container and volume GC on worker

ddadlani:
Requires https://github.com/concourse/baggageclaim/pull/17 to be merged in beforehand.

Fixes issue #2833

### DIFF
--- a/cmd/concourse/worker.go
+++ b/cmd/concourse/worker.go
@@ -164,14 +164,25 @@ func (cmd *WorkerCommand) Runner(args []string) (ifrit.Runner, error) {
 		)
 
 		members = append(members, grouper.Member{
-			Name: "sweeper",
+			Name: "container-sweeper",
 			Runner: NewLoggingRunner(
-				logger.Session("sweeper"),
-				&worker.SweepRunner{
-					Logger:             logger.Session("sweeper-runner"),
+				logger.Session("container-sweeper"),
+				&worker.ContainerSweeper{
+					Logger:             logger.Session("container-sweeper"),
 					Interval:           cmd.SweepInterval,
 					TSAClient:          tsaClient,
 					GardenClient:       gardenClient,
+				},
+			),
+		})
+		members = append(members, grouper.Member{
+			Name: "volume-sweeper",
+			Runner: NewLoggingRunner(
+				logger.Session("volume-sweeper"),
+				&worker.VolumeSweeper{
+					Logger:             logger.Session("volume-sweeper"),
+					Interval:           cmd.SweepInterval,
+					TSAClient:          tsaClient,
 					BaggageclaimClient: baggageclaimClient,
 				},
 			),
--- a/cmd/concourse/worker.go
+++ b/cmd/concourse/worker.go
@@ -21,8 +21,6 @@ import (
 	"github.com/tedsuo/ifrit/sigmon"
 )
 
-const SweeperMaxInFlight = 5
-
 type WorkerCommand struct {
 	Worker WorkerConfig
 
@@ -42,7 +40,9 @@ type WorkerCommand struct {
 	HealthcheckBindPort uint16        `long:"healthcheck-bind-port"  default:"8888"     description:"Port on which to listen for health checking requests."`
 	HealthCheckTimeout  time.Duration `long:"healthcheck-timeout"    default:"5s"       description:"HTTP timeout for the full duration of health checking."`
 
-	SweepInterval time.Duration `long:"sweep-interval" default:"30s" description:"Interval on which containers and volumes will be garbage collected from the worker."`
+	SweepInterval               time.Duration `long:"sweep-interval" default:"30s" description:"Interval on which containers and volumes will be garbage collected from the worker."`
+	VolumeSweeperMaxInFlight    uint16        `long:"volume-sweeper-max-in-flight" default:"5" description:"Maximum number of volumes which can be swept in parallel."`
+	ContainerSweeperMaxInFlight uint16        `long:"container-sweeper-max-in-flight" default:"5" description:"Maximum number of containers which can be swept in parallel."`
 
 	RebalanceInterval time.Duration `long:"rebalance-interval" description:"Duration after which the registration should be swapped to another random SSH gateway."`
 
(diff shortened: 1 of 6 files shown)


worker/volume_sweeper.go
```diff
@@ -0,0 +1,98 @@
...
+		maxInFlight := make(chan int, sweeper.maxInFlight)
+
+		for _, handle := range volumeHandles {
+			maxInFlight <- 1
+			wg.Add(1)
+
+			go func(handle string) {
+				err := sweeper.baggageclaimClient.DestroyVolume(logger.Session("destroy-volumes"), handle)
+				if err != nil {
+					logger.Error("failed-to-destroy-volumes", err)
```
cirocosta: given that the failure could be localized to that volume in specific, wdyt of having that handle in the context of the message here?

worker/volume_sweeper.go
```diff
@@ -0,0 +1,98 @@
...
+
+		err := sweeper.tsaClient.ReportVolumes(ctx, handles)
+		if err != nil {
+			logger.Error("failed-to-report-volumes", err)
+		}
+	}
+
+	volumeHandles, err := sweeper.tsaClient.VolumesToDestroy(ctx)
+	if err != nil {
+		logger.Error("failed-to-sweep-volumes", err)
```
cirocosta: Should we change this action to something else? Seems a bit vague for what could actually be failing here 🤔

ddadlani:
Note: we do not have any tests for this because we were unsure of how to test these changes. Does anyone have any ideas/opinions on how to test this without essentially testing Go concurrency?

worker/volume_sweeper.go
ddadlani: Good catch, we didn't update that error message. Will do 👍

worker/volume_sweeper.go
ddadlani: It would mostly be a connection error of some sort, but that error would be surfaced as well. The log line would have something like `failed-to-sweep-volumes, err: <actual_err_msg>`. I think the `failed-to-sweep-volumes` is just to give context on where the following error occurred.

ddadlani: Updated the log messages to add handles. Is this better? Wdyt?

cirocosta: hmmmm I see, I was looking at it from the perspective that the error would come from the fact that *listing the volume handles to be destroyed* failed, instead of the whole action of sweeping them 🤔 

e.g, consider this panel:

screen shot 2019-03-07 at 12 04 28 pm

for something like that, I imagine that being able to see the more specific action that is failing (getting the list of volumes to destroy) can help troubleshooting more than a more generic one.

Wdyt?

ddadlani: Changed the error message in both container- and volume-sweeper to reflect that it's the listing volumes/containers that fails. Does this make it clearer?

worker/tsa_config.go
```diff
@@ -9,7 +9,7 @@ import (
 type TSAConfig struct {
 	Hosts            []string            `long:"host" default:"127.0.0.1:2222" description:"TSA host to forward the worker through. Can be specified multiple times."`
 	PublicKey        flag.AuthorizedKeys `long:"public-key" description:"File containing a public key to expect from the TSA."`
-	WorkerPrivateKey *flag.PrivateKey    `long:"worker-private-key" description:"File containing the private key to use when authenticating to the TSA."`
+	WorkerPrivateKey *flag.PrivateKey    `long:"worker-private-key" required:"true" description:"File containing the private key to use when authenticating to the TSA."`
```
vito: 🤔 This is kinda just snuck in here but I guess it's fine. The intent was to make it easier to see if the worker can even run on its own before worrying about to registering it with the cluster. I'm not sure if anyone knows about it though.

ddadlani: @vito yeah, we saw it as part of the refactor of `cmd/concourse/worker.go` to separate out the volume and container sweepers, but I guess we should have made it more explicit. Do you want us to make it a separate commit?

worker/volume_sweeper.go
cirocosta: oooh it does! thank you!!

worker/tsa_config.go
vito: That's OK, let's just see if anyone complains about this being required now. It might be more obvious this way.

vito:
@kcmannem @ddadlani baggageclaim 1.4.0 has been tagged, so this can be updated now!

kcmannem:
@vito done, passed

cmd/concourse/worker.go
```diff
@@ -21,6 +21,8 @@ import (
 	"github.com/tedsuo/ifrit/sigmon"
 )
 
+const SweeperMaxInFlight = 5
```
vito: Can we make this a flag instead with 5 as the default? It seems like this could vary based on the user's hardware capabilities.

ddadlani: We thought about that, but for example, for volume sweeping, we don't know what strain  might be placed on `btrfs` with parallel sweeping. We thought we should try it out on prod first. 

Would users even try different `SweeperMaxInFlight` values to figure out the optimal one for their setup? In our local tests, setting a bigger number didn't always mean better. But a user might be inclined to believe that raising it makes performance increase.

vito: Isn't this ultimately a magic number based on the (I'm assuming few) environments we've tested in?

Having a sensible default based on our testing is great, but forbidding users from configuring it so that it better utilizes their hardware feels like it's going too far.

Once Concourse ships we really have no idea what kind of hardware people will run it on. I'd guess in some cases people will be using pretty powerful machines with fast disks (maybe even ramdisk). They might not even be using `btrfs`, and they might be using a newer kernel than what we've tried. Since all this takes is converting a magic number const to a flag with a default, it seems worth doing if it allows power users to fine-tune things like this without waiting for another Concourse release for us to make it a flag.

If a user sees the flag and increases it expecting performance to improve and it *doesn't*, well, at least now they can eliminate that theory and look for bottlenecks in other areas. (There's also a chance someone will want to *decrease* this, no?)

ddadlani: @vito Fair enough. Do we want separate container- and volume-sweeper-max-in-flight values too? Perhaps called VolumeParallelSweepingLimit or something?

vito: @ddadlani 👍

How about `--container-sweeper-max-in-flight` and `--volume-sweeper-max-in-flight`. I'm not sure if we need separate flags, but I suppose it doesn't hurt. By my own logic I guess it'd better to have them available. Maybe someone would want to increase one but not the other? 🤔

ddadlani: @vito Yeah, if someone is going to tweak them they may as well do so separately. Added the changes in.

PR #3444, opened 2019-03-07
Fly prints the unpause-pipeline command after creating a pipeline

benchristel:
This makes it easier to unpause newly-set pipelines.

### DIFF
--- a/fly/commands/internal/setpipelinehelpers/atc_config.go
+++ b/fly/commands/internal/setpipelinehelpers/atc_config.go
@@ -2,6 +2,7 @@ package setpipelinehelpers
 
 import (
 	"fmt"
+	"github.com/concourse/concourse/fly/rc"
 	"net/url"
 	"os"
 
@@ -19,6 +20,7 @@ import (
 type ATCConfig struct {
 	PipelineName     string
 	Team             concourse.Team
+	TargetName       rc.TargetName
 	Target           string
 	SkipInteraction  bool
 	CheckCredentials bool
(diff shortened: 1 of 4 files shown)


fly/commands/set_pipeline.go
```diff
@@ -52,6 +52,7 @@ func (command *SetPipelineCommand) Execute(args []string) error {
 	atcConfig := setpipelinehelpers.ATCConfig{
 		Team:             target.Team(),
 		PipelineName:     pipelineName,
+		TargetName:       Fly.Target,
```
benchristel: Storing both the `TargetName` and `Target` in the `ATCConfig` is a little gross, IMO. Alternatively, the `ATCConfig` could contain a complete `Target` and `ATCConfig.Set()` could unpack the name and URL from it. Perhaps there is a reason (earlier validation of the target?) this is not desirable, though.

fly/commands/internal/setpipelinehelpers/atc_config.go
```diff
@@ -112,7 +118,7 @@ func (atcConfig ATCConfig) showPipelineUpdateResult(created bool, updated bool)
 		fmt.Printf("you can view your pipeline here: %s\n", targetURL.ResolveReference(pipelineURL))
 		fmt.Println("")
 		fmt.Println("the pipeline is currently paused. to unpause, either:")
-		fmt.Println("  - run the unpause-pipeline command")
+		fmt.Println("  - run the unpause-pipeline command:\n    " + atcConfig.UnpausePipelineCommand())
```
vito: tiny nitpick: might as well just give this its own `fmt.Println`

fly/commands/set_pipeline.go
vito: To be honest I've been wanting to rid us of this whole `ATCConfig` abstraction anyway as it seems to just add unnecessary indirection, but I'm OK with throwing a tiny stick onto a burning tire fire for now. 😉

benchristel:
@vito I've pushed another change to address your suggestion. If the build passes, let's merge :)

amilkh:
+1 I reviewed it with Ben and we saw it working!

benchristel:
@vito the build is passing, but Github wants another approving review, probably because I pushed a new commit. Can you take another look?

PR #3454, opened 2019-03-08
web: Make networking a bit cleaner

xtreme-james-cooper:
We've kind of mixed up domain models and the network calls. This breaks up the Concourse package into two packages, Concourse (maybe move Concourse.elm in and rename it to "Domain"?) and Network. Not sure where JSON parsing should reasonably go yet.

Also some misc cleanup around Effects and FlySuccess.


pivotal-jamie-klassen:
conflicts with master!

Issue #2847, opened 2018-11-21
Owners and Members should be able to Pause/Resume their Pipelines from the Pipeline view


jama22:
With the removal of the sidebar in Concourse 4.x, users have lost the ability to Pause or Resume their pipeline from the pipeline view. 

In the past, a user would be able to open the sidebar, find their pipeline name and use the play/pause buttons to pause/resume their pipelines. This is no longer possible and the user must now go back in their browser, find their pipeline on the dashboard and pause/resume their pipeline. That's pretty annoying

# Proposal
We introduce the Play/Pause buttons on the pipeline view so users can get this functionality back

# Design

cc @Lindsayauchin for Design help

Lindsayauchin:
![screen shot 2018-12-05 at 5 02 44 pm](https://user-images.githubusercontent.com/27746614/49547068-c3a2ed00-f8af-11e8-8fcd-9295fee8cede.png)
![screen shot 2018-12-05 at 5 02 35 pm](https://user-images.githubusercontent.com/27746614/49547069-c43b8380-f8af-11e8-87b4-4fe0a7359199.png)

PR #3475, opened 2019-03-11
web: add pause button to top bar of pipeline view

robwhitby:
Fixes https://github.com/concourse/concourse/issues/2847

Hey, this adds a play/pause button to the top bar so you don't have to go back to the dashboard to update. I don't know Elm very well so please let me know if I've done something weird!

Thanks
Rob


pivotal-jamie-klassen:
This is really great! It's very encouraging to see some really effective reuse of some of the abstractions we've been defining in this project (Effects, inline styles).

There's a couple of UI polish things to help with the user experience:
* The divider should be `#3d3c3c` when the pipeline is not paused, and `rgba(255, 255, 255, 0.5)` (which it currently is) when unpaused.
* Hover states: The icon itself should have 50% opacity when unhovered and 100% opacity when hovered in both the paused and unpaused cases. This is true except if the user is not authorized to pause/unpause the pipeline (i.e. they have the `viewer` role on the team to which the pipeline belongs). In that case, the icon should always have 50% opacity and should have a default cursor rather than a pointer cursor.
* Click behaviour: If the user is authorized to pause/unpause the pipeline, clicking the icon should turn it into a spinner of the same diameter while the API call is in progress, and the page state should update immediately upon the API call resolving (HTTP errors, at this point, generally do not get reported to the user). If the user is unauthorized (like the `viewer` case described above), clicking should do nothing. If the user is unauthenticated, clicking the icon should redirect to login.

Whatever of those you have the bandwidth and interest to do is awesome. We'll leave this open until we hear further from you, but either way this is much appreciated.

There's an example for the paused divider color at https://github.com/concourse/concourse/blob/master/web/elm/src/TopBar/Styles.elm#L260-L266 with tests at https://github.com/concourse/concourse/blob/master/web/elm/tests/TopBarTests.elm#L216-L221 and https://github.com/concourse/concourse/blob/master/web/elm/tests/TopBarTests.elm#L296-L301

There's an example of the hover and unauthorized/unauthenticated behaviour at https://github.com/concourse/concourse/blob/master/web/elm/src/Resource/Resource.elm#L960-L1006 (the resource check button) with tests at https://github.com/concourse/concourse/blob/master/web/elm/tests/ResourceTests.elm#L2631-L2708 and https://github.com/concourse/concourse/blob/master/web/elm/tests/ResourceTests.elm#L2719-L2811 and https://github.com/concourse/concourse/blob/master/web/elm/tests/ResourceTests.elm#L3062-L3136.

Finally, there's an example of changing to a spinner and back at https://github.com/concourse/concourse/blob/master/web/elm/src/Resource/Resource.elm#L1351-L1400 (the resource pin button) with tests at https://github.com/concourse/concourse/blob/master/web/elm/tests/ResourceTests.elm#L2407-L2427.

robwhitby:
great, thanks for the quick feedback. I will have a look in the next few days.

robwhitby:
yep agree with idea of splitting the rest into another PR. And thanks again for such detailed feedback, the project is quite daunting at first but I am really getting to like elm :)

PR #3476, opened 2019-03-11
Create .flyrc with 0600 permissions instead of 0777

### LINKED ISSUE #2822: .flyrc file should be explicitly locked down upon creation
When the `.flyrc` file is created the permissions reflect the current umask (my `.flyrc` file ended up with `rwxr-xr-x`).  This file contains sensitive information (bearer tokens) and must be locked down after it is created regardless of the current umask.  If the user wants to open it up later -- that's their business.  But when the file does not exist it should be created with restricted permissions (rw------).

agurney:
Fixes #2822

Previously, the `fly` command line tool would create its `.flyrc` file using default permissions `rwxrwxrwx` (0777), or fewer subject to the active `umask`, with 0755 being typical.  This is undesirable since the file is not actually executable, and also contains authentication information that should not be exposed to other users. Now, we will default to `rw-------` (0600). Due to the way that the underlying `open` syscall works, existing `.flyrc` files will retain their existing permissions: this only applies to newly created ones.

### DIFF
--- a/fly/rc/targets_test.go
+++ b/fly/rc/targets_test.go
@@ -70,7 +70,8 @@ var _ = Describe("Targets", func() {
 
 			Describe("when the file exists with 0755 permissions", func() {
 				BeforeEach(func() {
-					ioutil.WriteFile(flyrc, []byte{}, 0755)
+					err := ioutil.WriteFile(flyrc, []byte{}, 0755)
+					Expect(err).ToNot(HaveOccurred())
 				})
 
 				It("preserves those permissions", func() {
--- a/fly/rc/targets_test.go
+++ b/fly/rc/targets_test.go
@@ -59,6 +59,30 @@ var _ = Describe("Targets", func() {
 	})
 
 	Describe("SaveTarget", func() {
+		Context("when managing .flyrc", func() {
+			It("creates any new file with 0600 permissions", func() {
+				err := rc.SaveTarget("foo", "url", false, "main", nil, "")
+				Expect(err).ToNot(HaveOccurred())
+				fi, statErr := os.Stat(flyrc)
+				Expect(statErr).To(BeNil())
+				Expect(fi.Mode().Perm()).To(Equal(os.FileMode(0600)))
+			})
+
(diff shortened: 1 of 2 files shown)


marco-m:
Nice catch! 
Would it make sense to also be retroactive and replace `ioutil.WriteFile()` with something that removes the file before writing it, so that we always get the `0600` permissions ?

fly/rc/targets_test.go
```diff
@@ -59,6 +59,30 @@ var _ = Describe("Targets", func() {
...
+				err := rc.SaveTarget("foo", "url", false, "main", nil, "")
+				Expect(err).ToNot(HaveOccurred())
+				fi, statErr := os.Stat(flyrc)
+				Expect(statErr).To(BeNil())
+				Expect(fi.Mode().Perm()).To(Equal(os.FileMode(0600)))
+			})
+
+			Describe("when the file exists with 0755 permissions", func() {
+				BeforeEach(func() {
+					ioutil.WriteFile(flyrc, []byte{}, 0755)
```
vito: Super unlikely but I think it's still worth checking the `err` return value here. In the unlikely event that it fails it'll make the test failure much more likely. It'll also make our linter happier.

cirocosta:
Hi @agurney , thanks for the PR!

I noticed you signed your last commit, but to have the `DCO` step properly done, you'll need to have both commits signed (or, you could squash them both and sign that).

Thanks!

fly/rc/targets_test.go
agurney: Added this check.

agurney:
Also added the proper Signed-off-by.
I had a go at some logic for getting the permissions right on existing files, but it was getting really gnarly with different error cases. (Basic idea being to create a temp file, write into that, then atomically rename it onto the existing file if everything was good so far, or else clean it up if things went wrong.) I may have a separate attempt at implementing that, keeping this PR with the stated behavior.

marco-m:
@agurney thanks for the attempt anyway.

vito:
thanks!

Issue #3423, opened 2019-03-04
topgun/k8s: use of config file for main team roles configuration


cirocosta:
Hey,

With the new possibility of now configuring team auth for the `main` team through a configuration file, it'd be nice to have a test that ensures that:

- with a `ConfigMap` paired with the right file location under `concourse.web.auth.mainTeam.config` one can have the `main` team parameters configured.

For instance:

```yaml
roles:
- name: member
  local:
    users: ["local-user-previously-added"]
```

scottietremendous:
Want to add documentation into the helm readme around this feature. Also want to ensure that this is captured within OSS docs as well.

PR #3477, opened 2019-03-11
topgun/k8s: verify user roles with config map

YoussB:
fixes: concourse/concourse#3423

cc @pivotal-bin-ju


cirocosta:
`watsjs` test should be unrelated; just pinned it and re-run 👍

Issue #3487, opened 2019-03-12
Investigate Postgres CPU Utilization after migrating to 5.0.0


YoussB:
## Bug Report

After upgrading to 5.0.0 on Wings we noticed an increase in the CPU utilization on Google Cloud SQL 

<img width="943" alt="Screen Shot 2019-03-12 at 12 03 35 PM" src="https://user-images.githubusercontent.com/17758967/54223850-bdd02f80-44ce-11e9-832f-10936949f1b3.png">

## Additional Context

- Might have to do with extra joins added to SQL queries as of 5.0.0.

## Version Info

* Concourse version: 5.0.0
* Deployment type (BOSH/Docker/binary): Bosh, Google cloud SQL
* Infrastructure/IaaS: GCP

Typositoire:
We do have the same thing here. We're seeing a 10% increase load with 5.0.0

vito:
Addressing this in #3602 - closing!

Issue #3459, opened 2019-03-08
Investigate using version_id instead of version_md5


jwntrs:
## What challenge are you facing?

We reference the `version_md5` in a bunch of places, and it seems to really slow down our queries.

## What would make this better?

Can we use the `version_id` instead?
At the very least we should probably index `version_md5`.

clarafu:
`version_md5` is used so that the tables that reference resource versions only need to care about if the specific version (hash) exists instead of caring about concourse's "resource versioning" where versions can be gced from the database and brought back under different `version_id`s even though they are the exact same versions.

I created another issue (#3460) for investigating the performance differences with using various column types for the version references.

pivotal-jwinters:
If versions are unique and we create foreign key references to map them to scopes/spaces then we would never have to delete the version itself. We would only need to remove the mapping. This means that we would have persistent version_ids that we can use elsewhere.

Issue #3426, opened 2019-03-04
Investigate Wings trigger job slowdown


chenbh:
As part of the [Run Existing Job](https://github.com/concourse/oxygen-mask/blob/master/monitor.yml#L101-L148) SLI we expect to take less than 5 seconds to trigger a build on an existing job.
  
Between Wednesday Feb 27 at 10 pm and Saturday March 1 at 12 pm some pipelines took up to 24 secs to trigger.
![image](https://user-images.githubusercontent.com/17577073/53742299-50852480-3e66-11e9-9496-2b7272ce482e.png)

Notes:
- There was no deployment when the issue started occurring (the previous deployment was on the 22nd and the next on the 28th)
- The issue was resolved without a deployment (at least no automatic deployment was recorded)
- Issue only applied to certain pipelines: a job with 147575 builds took over 10 secs whereas a job with only 120 builds took less than a second
- Logs were not searchable past last 24 hours so not much to go on there

YoussB:
- We're not yet able to find a specific cause for the problem. Also, not having any logs is making things harder. 
- One possible cause can be network latency connecting to the db or db query time. If we had a metric that measures the time taken by a query to get executed, then we can tell if db or network are a cause of the issue.

cc: @pivotal-bin-ju

cirocosta:
> If we had a metric that measures the time taken by a query to get executed, then we can tell if db or network are a cause of the issue

@pivotal-jwinters had some similar thoughts around having metrics for such queries 🤔

YoussB:
@cirocosta 
I will echo @pivotal-bin-ju here, If we have such metric, it won't say there is a network latency for sure, yet it will indicate that there is a problem from the db end.

YoussB:
findings:
- there are some functions that run when `CreatejobBuild` is run, most importantly: `CreateBuild` under `job.go:493`.
- This function runs a transaction that creates build data in the db.

Possible causes:
- network latency between the `oxygen mask` pipeline and the actual job it is running.
- many pipelines running at the same time adding more transactions to the db.
- network latency between atc and db (especially if they are not hosted on the same network).

Possible Mitigation/further research:
- add a metric for the time taken by the db to respond to queries.

chenbh:
The problem came back at March 7th 2:23pm. 
![image](https://user-images.githubusercontent.com/17577073/53995903-95c68200-4104-11e9-9c06-7a1b138189bb.png)

[...]

PR #3488, opened 2019-03-12
Incremental performance gain for LoadVersionsDB

pivotal-jwinters:
Use a NOT IN instead of negative join for inputs/outputs

#3487 and #3426, oh jeez and #3459

This doesn't fully fix the problem, but it seems to be an improvement.

### DIFF
--- a/atc/db/pipeline.go
+++ b/atc/db/pipeline.go
@@ -696,16 +696,14 @@ func (p *pipeline) LoadVersionsDB() (*algorithm.VersionsDB, error) {
 		Join("builds b ON b.id = o.build_id").
 		Join("resource_config_versions v ON v.version_md5 = o.version_md5").
 		Join("resources r ON r.id = o.resource_id").
-		LeftJoin("resource_disabled_versions d ON d.resource_id = r.id AND d.version_md5 = v.version_md5").
 		Where(sq.Expr("r.resource_config_scope_id = v.resource_config_scope_id")).
+		Where(sq.Expr("(r.id, v.version_md5) NOT IN (SELECT resource_id, version_md5 from resource_disabled_versions)")).
 		Where(sq.NotEq{
 			"v.check_order": 0,
 		}).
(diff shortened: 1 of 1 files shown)


cirocosta:
@pivotal-jwinters sooooo, after this, we should see a drop in the `SchedulingLoadVersionsDuration` metric. Would you have an idea of how much gains we should expect?

https://github.com/concourse/concourse/blob/02a90a4d0e99ca00767bda0b83189215543efe45/atc/metric/metrics.go#L54-L57

https://github.com/concourse/concourse/blob/90a604c940b240a520b284e9a4ba5ad7062c4189/atc/scheduler/runner.go#L102-L111

pivotal-jwinters:
cool! ya! from my super simplistic tests, it should be 2x (ish) faster.

pivotal-jwinters:
err maybe more. We run a separate query for inputs and outputs. Each is about 2x faster so overall improvement could be better.

atc/db/pipeline.go
```diff
@@ -696,16 +696,14 @@ func (p *pipeline) LoadVersionsDB() (*algorithm.VersionsDB, error) {
 		Join("builds b ON b.id = o.build_id").
 		Join("resource_config_versions v ON v.version_md5 = o.version_md5").
 		Join("resources r ON r.id = o.resource_id").
-		LeftJoin("resource_disabled_versions d ON d.resource_id = r.id AND d.version_md5 = v.version_md5").
 		Where(sq.Expr("r.resource_config_scope_id = v.resource_config_scope_id")).
+		Where(sq.Expr("(r.id, v.version_md5) NOT IN (SELECT resource_id, version_md5 from resource_disabled_versions)")).
```
vito: My only concern with this is that it's a select across all teams/pipelines/etc., but the dataset is probably small enough that it won't matter.

pivotal-jwinters: yeah that was my thinking too

Issue #3102, opened 2019-01-21
topgun/k8s: ensure container limits work


cirocosta:
When not running on top of BOSH, we lack some of the control we have over base VM image (in our current testing infrastructure - [GKE](https://cloud.google.com/kubernetes-engine/) -, there's no such thing as our use of [Stemcells](https://bosh.cloudfoundry.org/docs/stemcell/)), which might be a problem for the way we set container limits as it might be the case that certain platforms don't enable things like cgroup swap accounting (e.g., see https://github.com/Azure/acs-engine/issues/3107).

Going with an approach like we do for baggageclaim drivers (see https://github.com/concourse/concourse/blob/1b1b9ef171f7bc3fbb81e27fa7afa256a3c707b0/topgun/k8s/baggageclaim_drivers_test.go#L90-L119) where we explicitly test the behavior for each possible VM image in GKE (Ubuntu and [Container Optimized](https://cloud.google.com/container-optimized-os/)), we can be more sure about how our platform coverage looks like.

YoussB:
@pivotal-bin-ju & I have created the needed tests, but it looks like setting `cgroups` configs for ubuntu is unsupported.

we got the following error:
```
runc run: exit status 1: container_linux.go:344: starting container process caused "process_linux.go:424: container init caused \"process_linux.go:390: setting cgroup config for procHooks process caused \\\"failed to write 1073741824 to memory.memsw.limit_in_bytes: open /sys/fs/cgroup/memory/garden/105b72a1-4b41-4c9c-5229-a651fc9590ee/memory.memsw.limit_in_bytes: permission denied\\\"\""
```
Do you know if there is anything that we might be missing?
@cirocosta @topherbullock 

will also push the changes to a feature branch `feature/3102-container-limits` so you can take a look.

thanks!

cirocosta:
The results @YoussB mentioned seems related to https://github.com/concourse/concourse/issues/3055 🤔

PR #3496, opened 2019-03-13
topgun/k8s: container limits

YoussB:
- added tests for container_limits for both ubuntu and cos, ubuntu doesn't seem to be working.
- added rand number generator to fix the `--wait=false` problem
- fixed tests naming

for the issue: concourse/concourse#3102

### DIFF
--- a/topgun/k8s/baggageclaim_drivers_test.go
+++ b/topgun/k8s/baggageclaim_drivers_test.go
@@ -2,6 +2,7 @@ package k8s_test
 
 import (
 	"fmt"
+
 	"github.com/onsi/gomega/gexec"
 
 	. "github.com/concourse/concourse/topgun"
@@ -35,10 +36,9 @@ var _ = Describe("Baggageclaim Drivers", func() {
 
 	DescribeTable("across different node images",
 		func(c Case) {
-			releaseName = fmt.Sprintf("topgun-bd-%s-%s-%d-%d",
-				c.Driver, c.NodeImage, randomGenerator.Int(), GinkgoParallelNode())
+			releaseName = fmt.Sprintf("topgun-bd-%s-%s-%d",
+				c.Driver, c.NodeImage, randomGenerator.Int())
 			namespace = releaseName
-
 
 			helmDeployTestFlags := []string{
 				"--set=concourse.web.kubernetes.enabled=false",
--- a/topgun/k8s/baggageclaim_drivers_test.go
+++ b/topgun/k8s/baggageclaim_drivers_test.go
@@ -36,35 +36,18 @@ var _ = Describe("Baggageclaim Drivers", func() {
 	DescribeTable("across different node images",
 		func(c Case) {
 			releaseName = fmt.Sprintf("topgun-bd-%s-%s-%d-%d",
-				c.Driver, c.NodeImage, GinkgoRandomSeed(), GinkgoParallelNode())
+				c.Driver, c.NodeImage, randomGenerator.Int(), GinkgoParallelNode())
 			namespace = releaseName
 
-			args := []string{
-				"upgrade",
-				"--force",
-				"--install",
-				"--namespace=" + namespace,
+
+			helmDeployTestFlags := []string{
 				"--set=concourse.web.kubernetes.enabled=false",
 				"--set=concourse.worker.baggageclaim.driver=" + c.Driver,
-				"--set=image=" + Environment.ConcourseImageName,
-				"--set=postgresql.persistence.enabled=false",
-				"--set=web.livenessProbe.failureThreshold=3",
-				"--set=web.livenessProbe.initialDelaySeconds=3",
-				"--set=web.livenessProbe.periodSeconds=3",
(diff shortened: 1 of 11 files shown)


topgun/k8s/baggageclaim_drivers_test.go
```diff
@@ -36,35 +36,18 @@ var _ = Describe("Baggageclaim Drivers", func() {
 	DescribeTable("across different node images",
 		func(c Case) {
 			releaseName = fmt.Sprintf("topgun-bd-%s-%s-%d-%d",
-				c.Driver, c.NodeImage, GinkgoRandomSeed(), GinkgoParallelNode())
+				c.Driver, c.NodeImage, randomGenerator.Int(), GinkgoParallelNode())
```
cirocosta: Oh, so the problem was that we were trying to delete a namespace that had a naming collision with something else?
I wonder if we should then just drop `GinkgoParallelNode()` to as the call to `.Int()` is going to give us a different number each time we call anyway. 

Wdyt?

thx!

topgun/k8s/container_limits_test.go
```diff
@@ -0,0 +1,99 @@
...
+		BeforeEach(func() {
+			nodeImage = "cos"
+		})
+
+		It("returns the configure default container limit", func() {
+			testLimits()
+		})
+	})
+
+	XContext("passing a config map location to the worker to be used by gdn", func() {
```
cirocosta: In the Baggageclaim drivers tests, we were instead asserting the negation (e.g., asserting that it fails to deploy) so that the test continuously run despite the fact that we expect the `false` condition in that case.

With an `XContext` we don't have that behaviour though - we'd need to manually deactivate `XContext` to see if the negative in such case still holds true.

Wdyt of going with an assertion that verifies that it doesn't work on `ubuntu`?

YoussB: makes sense. Deployment doesn't fail in this case, though. Running the task is the one that fails.

topgun/k8s/baggageclaim_drivers_test.go
YoussB: ok, I am also thinking about getting a smaller number ... this one is making the `releasename` huge.

YoussB:
@cirocosta Fixed, this should also unblock topgun.

topgun/k8s/external_worker_test.go
```diff
@@ -89,9 +89,8 @@ var _ = Describe("team external workers", func() {
 
 	AfterEach(func() {
 		helmDestroy(releaseName)
-		Wait(Start(nil, "kubectl", "delete", "namespace", namespace))
+		Wait(Start(nil, "kubectl", "delete", "namespace", namespace, "--wait=false"))
```
cirocosta: oops nice!

topgun/k8s/container_limits_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"github.com/onsi/gomega/gbytes"
+	"github.com/onsi/gomega/gexec"
+
+	. "github.com/concourse/concourse/topgun"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Garden Config", func() {
```
cirocosta: OOoh I just noticed that here we're using the `Garden Config` description.

topgun/k8s/container_limits_test.go
```diff
@@ -0,0 +1,99 @@
...
+			ShouldNot(HaveLen(0))
+	})
+
+	AfterEach(func() {
+		helmDestroy(releaseName)
+		Wait(Start(nil, "kubectl", "delete", "namespace", namespace, "--wait=false"))
+		Wait(proxySession.Interrupt())
+	})
+
+	Context("passing a config map location to the worker to be used by gdn", func() {
```
cirocosta: I guess this is the wrong description to use 😁

topgun/k8s/container_limits_test.go
```diff
@@ -0,0 +1,99 @@
...
+				"cat /sys/fs/cgroup/memory/memory.memsw.limit_in_bytes; cat /sys/fs/cgroup/cpu/cpu.shares",
+			)
+			<-hijackSession.Exited
+
+			Expect(hijackSession.ExitCode()).To(Equal(0))
+			Expect(hijackSession).To(gbytes.Say("1073741824\n512"))
+		})
+	})
+
+	Context("passing a config map location to the worker to be used by gdn", func() {
```
cirocosta: I guess this is the wrong description to use 😁

topgun/k8s/container_limits_test.go
```diff
@@ -0,0 +1,99 @@
...
+		// Context skipped as GKE ubuntu doesn't support container limits.
```
cirocosta: We can remove this comment now 👍

topgun/k8s/container_limits_test.go
YoussB: whoops ... the blind copy paste problem 😁

Issue #3494, opened 2019-03-13
downgrade job failed by missing worker when resource check


xtremerui:
It seems after it stop web node and migrate the db, the worker container is removed while it try to docker compose up the latest release (https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/downgrade/builds/158#L5c8673b0:477). 

However, since DB stays the same (other than the migration history), the now running web node still consider the removed worker alive (https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/downgrade/builds/158#L5c8673b0:513) and check-resource failed with non found worker afterwards(https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/downgrade/builds/158#L5c8673b0:529) since the check container still points to the removed worker. 

Ideally there should be a wait or something to confirm the web node aware of the removed worker and update the DB (mark it as stalled or remove it) thus the fly check-resource could use new container

xtremerui:
As #3510 merged, `downgrade` job is good again.

Issue #3079, opened 2019-01-17
Never re-use check containers


vito:
## What challenge are you facing?

Check containers are re-used for up to an hour (by default) so that resources can re-use cached state that they fetched (i.e. so a fresh `git clone` doesn't have to run all the time).

We limit their re-use to an hour so that they get rebalanced across workers periodically. However that means every hour forces a fresh `git clone`, which makes people want to extend their lifetime further (#2988).

This also means that for a bunch of resources, there will always be a bunch of containers across all the workers, leading to desperate measures like https://github.com/concourse/concourse/issues/3054 and in general a ton of questions because people are almost always surprised to see so many `check` containers.

We also are super careful to lock and make sure only one ATC is using a container at a time.

And we're careful about how these containers are re-used (e.g. across teams).

## What would make this better?

Let's not do that!

Can we instead create a one-time-use container every time we need to run a `check`, and use a volume to explicitly track the cached state between runs?

This volume could work like task caches - there would be one persisted on each worker that the `check` runs on. A copy of it would be mounted to the container, and that copy would become the new cache for subsequent runs.

Open questions:

* How quickly can containers be created/deleted? There should be some way to make this fast as Diego and Garden teams benchmarked the hell out of it back in the day. But we should check and make sure, because this will result in a lot more creations/deletes.
* How do we identify the GC/ownership of these containers now that they're one-off? We don't want them deleted while they're being used, but we do want them GC'd if there's some way to tell it's no longer in use.
* How do we handle the lifecycle of the cache volumes?

## Are you interested in implementing this yourself?

Yeah, actually, it sounds fun.

pivotal-jwinters:
I really like this idea. However I think this might be a good idea to rethink the way we model volume `uses` since this change will likely involve yet another another foreign key on the volumes table. 

At the moment, a `volume` (sometimes) belongs to a team, and defines its `use` through a foreign key reference (i.e. resource cache, task cache, etc). I see volumes as a lower level concept that shouldn't contain all this information. I think we should invert this relationship, and have the higher level concepts track both the team (where applicable) and volume foreign keys. This will make adding new volume `uses` much lighter weight, and let us query for the higher level objects, and join on the lower level objects.

@vito thoughts?

vito:
@pivotal-jwinters I wonder if the idea I threw around in https://github.com/concourse/concourse/pull/3025#issuecomment-454612856 meshes well with that.

[...]

PR #3504, opened 2019-03-14
keep fly check resource until it succeed in downgrade/upgrade job

xtremerui:
Refer to #3494 

There is a chance that `fly check resource` failed when looking up existing check container on an already-removed worker.

This won't be an issue anymore if https://github.com/concourse/concourse/issues/3079 is done. Since the very first `fly check resource` after downgrade would just spin up a new check container instead of using the old one.


vito:
I started on a cleanup of the Docker Compose setup that I think might remove the need for this, fwiw. I'll have something to push soon - just doing some local manual testing.

xtremerui:
As #3510 is merged this PR shouldn't be needed anymore.

PR #3505, opened 2019-03-14
Pipeline config API cleanup

vito:
There's a ton of cruft 'round these parts.

* We use `mapstructure` which makes declaring our own parsing rules very painful (e.g. `version` on a `get` step, task limit values).
* We define `yaml:`, `json:`, and `mapstructure:` fields which is verbose and ugly to maintain.
* The API supports `multipart/form-boundary` so the client can configure whether the pipeline is paused or not at configuration time. This is complicated and not even a `fly` CLI flag anymore.

The goal of this PR is to unify everything as YAML, since JSON is a subset of YAML anyway and then we can just define a single Unmarshaler for complex fields like [`version`](https://concourse-ci.org/get-step.html#get-step-version).


atc/container_limits_marshaller.go
```diff
@@ -23,20 +30,94 @@ func (c *ContainerLimits) UnmarshalJSON(limit []byte) error {
...
-		return err
+	var c ContainerLimits
+
+	var memoryBytes uint64
+	var uVal int
+	var err error
+
+	// the json unmarshaller returns numbers as float64 while yaml returns int
+	for key, val := range mapData {
+		if key == "memory" {
```
pivotal-jwinters: Can't you just look these up directly in the map?

atc/db/migration/migrations/1522178770_add_job_tags.up.go
```diff
@@ -14,8 +14,8 @@ type GroupConfig struct {
 type GroupConfigs []GroupConfig
 
 type Pipeline struct {
-	ID     int          `json:"id"`
-	Groups GroupConfigs `json:"groups,omitempty"`
+	ID     int
+	Groups GroupConfigs
```
pivotal-jwinters: Did you mean to remove these?

atc/container_limits_marshaller.go
vito: For posterity: probably, though this is actually just copied over from the decode hook; didn't write it myself.

atc/db/migration/migrations/1522178770_add_job_tags.up.go
vito: Yep, they don't actually do anything; the object was never used as JSON.

vito:
thanks!

PR #3506, opened 2019-03-14
ci: change topgun param BOSH_KEY to BOSH_SSH_KEY for clarity

kcmannem:
We got confused.


vito:
Doesn't this need a task or code change, too?

ddadlani:
Updated and tested with `fly execute`. Thanks for pointing it out!

PR #3510, opened 2019-03-14
rework how we use docker-compose in CI

vito:
* explicitly generate keys; this allows them to be shared between upgrades and downgrades
* don't use the dev Dockerfile and don't perform any building
* use an override file to swap out the image for upgrade/downgrade; don't use Quickstart
* stop using ephemeral workers, since this caused the workers to disappear under load
* set `stop_signal: SIGUSR2` so that workers retire during upgrade/downgrade
* pass the latest final concourse image as an input to upgrade/downgrade

This depends on `concourse/dev` changes so I don't think the builds for this PR will succeed. I verified locally by building and pushing to a different repo and changing references by hand. I was able to successfully run upgrade/downgrade/testflight.


ci/tasks/scripts/downgrade-test
```diff
@@ -11,25 +11,41 @@ source concourse/ci/tasks/scripts/docker-helpers.sh
...
 downgrade_to=$(docker run concourse/concourse:latest migrate --supported-db-version)
-docker-compose stop web
-docker-compose run web migrate --migrate-db-to-version $downgrade_to
+docker-compose exec web concourse migrate --migrate-db-to-version $downgrade_to
 
 # downgrade and verify
-docker-compose -f docker-compose-latest.yml up --remove-orphans -d
+docker-compose \
+  -f docker-compose.yml \
+  -f ci/overrides/docker-compose.ci.yml \
```
xtremerui:
```suggestion
```
Might not need this?

PR #3517, opened 2019-03-15
Update contributing guide - fix trace, add note about dependency forks

cirocosta:
Hey,

I noticed that we don't provide guides around how one can fork a dependency like `baggageclaim` and have that forked version being used in a `docker-compose build` scenario, so this PR tries to address that.

It also fixes the `./hack/trace` script that was missing some updates in the Dockerfile, as noticed by @kcmannem and @topherbullock  the other day 😁  

Thanks!


cirocosta:
note: although it's true that a `concourse` member could just push to a branch under `concourse/$dep` and then `go get github.com/concourse/$dep@$branch`, that's not what external contributors would do, so I didn't mention this other case. I also didn't write about the possible use of relative paths in the replace directive as that'd be useful either only through [`docker build`s with bind mounts](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/experimental.md#experimental-syntaxes) or local builds w/out docker (as far as I can see).

Wdyt?

Should we cover more scenarios?

Thanks!

PR #3528, opened 2019-03-16
rename --tsa-peer-ip to --tsa-peer-address

vito:
this can also be a DNS address, not just an IP

this looks backwards-incompatible but in practice this flag was always
auto-set by taking the host from `--peer-url`. that flag has now been
removed, and --tsa-peer-address is to be used instead, so we might as
well get the name right.


vito:
needed by https://github.com/concourse/concourse-bosh-release/commit/932c64ecbff44bb10eb468bcae3656a16499c604

vito:
Maybe we should just call this `--peer-address` instead (in the `concourse` binary) - it's a relatively prominent flag and I'd rather start making the binary feel more integrated and start removing prefixes like `--tsa-` so there's not so much flag churn if e.g. the ATC starts needing a peer address, too.

vito:
Gonna merge this to unblock the pipeline; I opened https://github.com/concourse/concourse/issues/3538 for the eventual cleanup (which we should do for 5.1).

PR #3536, opened 2019-03-18
WIP added support for vault kv engine v2

hprotzek:
Added support for vault k/v engine v2


hprotzek:
@vito Yes sure, we can survive with v1 for now :-)

vito:
👍 Thanks!

Issue #3508, opened 2019-03-14
Hasten state changes using LISTEN/NOTIFY


vito:
* Kick resource scanners when a job has been manually triggered
  * In combination with #2438 where we'd remove the checking and just wait for checks to have been done on their own
* Kick build trackers when a one-off build is submitted
  * Would speed up builds and bring us back to parity after #3307

vito:
Broke the third point out into https://github.com/concourse/concourse/issues/3596 instead

PR #3539, opened 2019-03-18
Don't block scheduling on resource scans 

mhuangpivotal:
Wait for all input resources to be checked instead of doing a check in place for manually triggered builds.

- leave the build in pending state if not all resources have finished
checking after the build creation time, unless the resource is pinned
- scheduler no longers needs the resource scanner
- add last_check_finished to resource_config_scopes to keep
track of when the last check finished
- add create_time to builds

fixes concourse/concourse#2438


mhuangpivotal:
Follow up issue: #3508 for the LISTEN/NOTIFY update.

Issue #3052, opened 2019-01-15
Decouple container and volume creation in FindOrCreateContainer


ddadlani:
Related to #2926.

After #3043, [ContainerProvider.FindOrCreateContainer](https://github.com/concourse/concourse/blob/f4075856f392c5cdfee44fbf6e1cdb7b0f607fa9/atc/worker/container_provider.go#L93) is easier to read and part of its responsibilities have been moved to [Worker.FindOrCreateContainer](https://github.com/concourse/concourse/blob/f4075856f392c5cdfee44fbf6e1cdb7b0f607fa9/atc/worker/worker.go#L146)

However, there are still some redundancies:
1. [containerProvider.createGardenContainer](https://github.com/concourse/concourse/blob/f4075856f392c5cdfee44fbf6e1cdb7b0f607fa9/atc/worker/container_provider.go#L304) still creates its own `gardenWorker`
1. `containerProvider` and `gardenWorker` are both passed a `volumeClient`, `db.Worker`, `gardenClient` and `imageFactory`, despite the fact that the `ContainerProvider`'s two exposed methods are only called from the `gardenWorker`.
1. If, in the future, we want to create the volumes at an earlier stage in the process, we need to move volume creation up one level since currently it is deeply intertwined with container creation.

This issue is meant to capture the work required to untangle the functionality of `gardenWorker` and `containerProvider` and maybe reduce the responsibilities of `containerProvider` as well.

ddadlani:
As part of the detangling, we moved all functionality of the `containerProvider` into the `gardenWorker`, however, this doesn't make it easier to remove the cyclic dependency. A lot of the logic requiring the `gardenWorker` in `containerProvider.createGardenContainer` is because of the creation or inflation of volumes, which requires a `gardenWorker`. This strengthens the argument for decoupling the volume lifecycle from the container lifecycle.

Currently, to create a db volume, you need a container ID, which means that the creation of a volume must come after the creation of a `creatingContainer`. Creating a volume without a container ID will cause the volume to be GCed because it is considered orphaned.

Our next steps will be to investigate adding lifecycle states and timestamps to the database `volumes` table. This may either be a separate `initializing` state, or a replication of the container lifecycle states (`creating`, `created`, `destroying`) depending on what works best. Volumes in the `initializing` or `creating` state would not be GCed. This would eliminate the need for a containerID in the volume.

This raises the following questions:
[...]

Issue #3301, opened 2019-02-14
fewest-build-containers: tasks in aggregate land on the same worker


wagdav:
## Bug Report

The "fewest-build-container" placement strategy always puts tasks within an aggregate block to the same worker.

## Steps to Reproduce

* enable the fewest-build-container strategy
* create a pipeline with an aggregate block

```
- name: job with aggregate
  plan:
  - aggregate:
    - task: build-on-ubuntu-16.04
    - task: build-on-ubuntu-18.04
```

Schedule some builds and observe the output of `fly containers`

## Expected Results

Tasks within an aggregate respect the fewest-build-container strategy. Sometimes they _may_ land on the same worker.

## Actual Results

Tasks within an aggregate _always_ land on the same worker.

## Additional Context

A sample output of `fly containers`:

```
63aebe28-b48c-4ac1-6d33-ba1a84300cac  wrk-lin-0-07b4efc347e46c064      pipeline-master                                   build                      81    84161  task   build-ubuntu-18.04                        n/a
71e38b7e-aaf0-42db-462e-7d2a361ec90a  wrk-lin-0-07b4efc347e46c064      pipeline-master                                   build                      81    84161  task   build-ubuntu-16.04                        n/a
[...]
ddd3b498-b3f8-44ce-4735-246937b4b884  wrk-lin-3-0544d8e8cdbc32852      pipeline-master                                   conan-pkg                  15    84179  task   build-ubuntu-16.04                        n/a
e720603b-537c-4c49-4be1-cdeb37402155  wrk-lin-3-0544d8e8cdbc32852      pipeline-master                                   conan-pkg                  15    84179  task   build-ubuntu-18.04                        n/a
```

## Version Info

* Concourse version: 5.0 RC 74
* Deployment type (BOSH/Docker/binary): binary
* Infrastructure/IaaS: AWS
* Browser (if applicable):
* Did this used to work? no.

kcmannem:
Thanks for pointing this out! It makes sense that we see this because the `fewest-build-containers` strategy depends upon the number of build containers on a worker. https://github.com/concourse/concourse/blob/master/atc/worker/placement.go#L75-L81

Since the aggregate steps are scheduled in parallel, the number of build containers on each worker is the same when scheduling all of them, and so the placement strategy selects the same worker for each one.
https://github.com/concourse/concourse/blob/master/atc/exec/aggregate.go#L22-L27

[...]

PR #3546, opened 2019-03-19
worker: remove ContainerProvider

ddadlani:
combine containerProvider and worker
add workerHelper which stores helper functions

fixes concourse/concourse#3052


vito:
😮

ddadlani:
Preliminary thoughts? 

One of the things we are not happy with is the length of createVolumes https://github.com/concourse/concourse/blob/955502744bc4c9c9b8db72c74bde5c4e80434d53/atc/worker/worker.go#L324-L465 and are looking to cut that down.

The main advantage that this refactor gives us is the isolation of `findOrInitializeContainer` https://github.com/concourse/concourse/blob/955502744bc4c9c9b8db72c74bde5c4e80434d53/atc/worker/worker.go#L187, which allows us to lock `CreatingContainer` creation for https://github.com/concourse/concourse/issues/3301

kcmannem:
Our other streams of work are based on this branch. We wanna integrate it into master. We pulled apart as much as we could without artifacts gc (volumes not dependent on containers). Another reason is to remove a the `CreatingContainerLock` which is unnecessary.

PR #3548, opened 2019-03-20
topgun/k8s -  Fix tests

YoussB:
- Increase default timeouts for eventually and consistently
- refactor namespace & releaseName to suite level
- fix rand generator seeding for releaseName/namespace
- fix missing svc prefix in main_team_role test

### DIFF
--- a/topgun/k8s/rebalance_test.go
+++ b/topgun/k8s/rebalance_test.go
@@ -2,6 +2,7 @@ package k8s_test
 
 import (
 	"fmt"
+	"math/rand"
 	"strings"
 	"time"
 
@@ -21,7 +22,7 @@ var _ = Describe("Worker Rebalancing", func() {
 	)
 
 	BeforeEach(func() {
-		releaseName = fmt.Sprintf("topgun-wr-%d", randomGenerator.Int())
+		releaseName = fmt.Sprintf("topgun-wr-%d-%d", rand.Int(), GinkgoParallelNode())
 		namespace = releaseName
 
 		deployConcourseChart(releaseName,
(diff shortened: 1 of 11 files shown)


topgun/k8s/rebalance_test.go
```diff
@@ -14,22 +13,21 @@ import (
...
+		setReleaseNameAndNamespace("wr")
 
 		deployConcourseChart(releaseName,
 			"--set=concourse.worker.ephemeral=true",
 			"--set=worker.replicas=1",
 			"--set=web.replicas=2",
 			"--set=concourse.worker.rebalanceInterval=5s",
-			"--set=concourse.worker.baggageclaim.driver=detect")
+			"--set=concourse.worker.baggageclaim.driver=detect",
+			"--set=web.env[0].name=CONCOURSE_TSA_PEER_ADDRESS",
```
cirocosta: Hey, why is this one needed? 

Is it only to circumvent the fact that we have new flags to be configured?

If that's the case, I'd suggest instead updating [`refs.json` in `concourse/charts`](https://github.com/concourse/charts/blob/maintenance/refs.json) to point to a branch where we can live a [Draft PR](https://github.blog/2019-02-14-introducing-draft-pull-requests/) to get the new flags for `5.1` in and then let the automation take care of producing the "merged view" of "upstream + new flags" 🤔

YoussB: @cirocosta 
Added a PR for updating the Helm Chart with the flag changes. Removed these temp changes from this PR.

cirocosta: Great! Thanks!!

cirocosta:
(for anyone wondering where that PR is: https://github.com/helm/charts/pull/12436)

PR #3552, opened 2019-03-20
fly progressbar switcharoo

vito:
* switches from cheggaaa/pb to mpb
* uploads inputs in parallel
* updates ATC CLI api to return `Content-Length` so that `fly sync` can render a proper progressbar

i considered writing inputs to disk so that we know the file size, but ultimately decided not to yet as we may need to consider disk usage in these situations.


fly/commands/internal/executehelpers/inputs.go
```diff
@@ -137,29 +141,68 @@ func GenerateLocalInputs(
...
+		}(i, mapping.Path)
+	}
 
-	kvMap := map[string]Input{}
-
-	for _, i := range inputMappings {
-		artifact, err := Upload(team, i.Path, includeIgnored)
-		if err != nil {
-			return nil, err
+	progress.Wait()
```
pivotal-jwinters: Can we push the progress bar into the upload function to be more symmetrical with download? And then if we want concurrent uploads we can implement this separate from the progress bar implementation?

vito: The progress bars have to be registered first otherwise it's a race condition similar to doing `wg.Add(1)` in a goroutine. Or are you suggesting we forego the parallel uploads for now and just move the progress bar stuff entirely into `Upload` and call it sequentially?

pivotal-jwinters: Honestly yeah maybe we just tackle the parallel uploads later. Downloads are sequential and its something we might want to consider there as well.

vito: What's the reasoning to not just do parallel upload now that we're switching to a progressbar library that supports it? To construct the bar in `Upload` I'll have to pass in more information just so the bar has an appropriate name (we only give it a path at the moment).

At this point I would rather change the downloads to run in parallel too.

pivotal-jwinters: It just seems weird to tie our concurrency decisions to a progress bar lib. To me the progress bar is just decoration.

It also used to be super clear that all we're doing is iterating over all the inputs and uploading them. Now it's much harder to reason about whats going on in that file.

vito: I'm not switching to concurrent upload/download because the library enforces it (it doesn't), I wanted to switch to concurrent just because it's nice to have, and the new library lets us do it easily. The progress bar is decoration, but if we want to have concurrent uploads/downloads the progress bar package has to be able to handle it properly, otherwise the output will be a mess.

I just pushed a refactor that should address the complicated looking upload code, though, and also makes downloads run in parallel too. :+1:

pivotal-jwinters: awesome!

PR #3560, opened 2019-03-21
time based retention for build log collector

hprotzek:
This adds a time based retention to the build log collector. So it's possible to configure retention count or time or both. If both are configured, the highest value wins: e.g. 10 builds, 1 day = will keep more then 10 builds if per day.


atc/atccmd/command.go
```diff
@@ -148,6 +148,9 @@ type RunCommand struct {
 	DefaultBuildLogsToRetain uint64 `long:"default-build-logs-to-retain" description:"Default build logs to retain, 0 means all"`
 	MaxBuildLogsToRetain     uint64 `long:"max-build-logs-to-retain" description:"Maximum build logs to retain, 0 means not specified. Will override values configured in jobs"`
 
+	DefaultBuildLogsDaysToRetain uint64 `long:"default-build-logs-days-to-retain" description:"Default days to retain build logs. 0 means unlimited"`
```
vito: The name here feels odd - how about `default-days-to-retain-build-logs` and `max-days-to-retain-build-logs`?

atc/job_config.go
```diff
@@ -10,6 +10,7 @@ type JobConfig struct {
 	SerialGroups         []string `yaml:"serial_groups,omitempty" json:"serial_groups,omitempty" mapstructure:"serial_groups"`
 	RawMaxInFlight       int      `yaml:"max_in_flight,omitempty" json:"max_in_flight,omitempty" mapstructure:"max_in_flight"`
 	BuildLogsToRetain    int      `yaml:"build_logs_to_retain,omitempty" json:"build_logs_to_retain,omitempty" mapstructure:"build_logs_to_retain"`
+	BuildLogsDaysToRetain int      `yaml:"build_logs_days_to_retain,omitempty" json:"build_logs_days_to_retain,omitempty" mapstructure:"build_logs_days_to_retain"`
```
vito: What do you think of something like this?:

```yaml
build_log_retention:
  builds: 10
  days: 5
```

We'd have to keep `build_logs_to_retain` for backwards-compatibility, but this would make it easier to add more retention policies without polluting the toplevel configuration too much. I could see someone wanting something like this for example:

```yaml
build_log_retention:
  failed_builds: 1000
  succeeded_builds: 10
```

Whaddaya think?

hprotzek:
Thanks @vito I will look into it and implement your feedback.

hprotzek:
Morning @vito finally I got time to add your suggestions. Build log retention can now be configured:
```
build_log_retention:
  builds: 10
  days: 5
```
The `old build_logs_to_retain` is still supported. When used both, a validation error will be thrown.

atc/gc/build_log_collector.go
```diff
@@ -123,6 +130,12 @@ func (br *buildLogCollector) Run(ctx context.Context) error {
 					}
 				}
 
+				if daysToRetainBuildLogs > 0 {
+					if build.StartTime().AddDate(0, 0, daysToRetainBuildLogs).After(time.Now()) {
```
vito: Should this be `build.FinishTime()`? It might matter for builds that take a very long time to run - if it runs for 2 days but I have a policy of 'reap after 1 day' I wouldn't want it to be immediately reaped.

atc/gc/build_log_retention_calculator.go
```diff
@@ -5,43 +5,69 @@ import (
...
-		maxBuildLogsToRetain:     maxBuildLogsToRetain,
+		defaultBuildLogsToRetain:     defaultBuildLogsToRetain,
+		maxBuildLogsToRetain:         maxBuildLogsToRetain,
+		defaultDaysToRetainBuildLogs: defaultDaysToRetainBuildLogs,
+		maxDaysToRetainBuildLogs:     maxDaysToRetainBuildLogs,
 	}
 }
 
-func (blrc *buildLogRetentionCalculator) BuildLogsToRetain(job db.Job) int {
+func (blrc *buildLogRetentionCalculator) BuildLogsToRetain(job db.Job) (int, int) {
```
vito: Maybe this could return `atc.BuildLogRetention`? It's not clear which `int` is which, and we already have the handy struct for it.

atc/validate.go
```diff
@@ -217,13 +217,33 @@ func validateJobs(c Config) ([]ConfigWarning, error) {
 			errorMessages = append(errorMessages, identifier+" has no name")
 		}
 
-		if job.BuildLogsToRetain < 0 {
+		if job.BuildLogRetention != nil && job.BuildLogsToRetain != 0 {
+			errorMessages = append(
+				errorMessages,
+				identifier+fmt.Sprintf(" can't use both build_log_retention and deprecated build_logs_to_retain"),
```
vito:
```suggestion
				identifier+fmt.Sprintf(" can't use both build_log_retention and build_logs_to_retain"),
```

We haven't deprecated it just yet - the validation makes sense but we haven't decided whether we want to deprecate the shorthand form. :slightly_smiling_face:

hprotzek:
Hi @vito ,I added your suggestions, thanks for the review.

PR #3566, opened 2019-03-21
Better abort tracking

### LINKED ISSUE #1585: Aborting a started build prior to an ATC tracking it will result in an orphaned, uncompleted build, still running

pivotal-jwinters:
This adds an aborted column to track aborted builds instead of immediately updating the build status to `aborted`. This makes sure that the build tracker will still pick up a build if it gets aborted before it ever starts.

fixes #1585


pivotal-jwinters:
@vito yeah makes sense. Seems to be a simple fix, since we already abort the build when it can't be started.

atc/db/worker_lifecycle.go
```diff
@@ -80,14 +80,7 @@ func (lifecycle *workerLifecycle) DeleteFinishedRetiringWorkers() ([]string, err
 		Join("containers c ON b.id = c.build_id").
 		Join("workers w ON w.name = c.worker_name").
 		LeftJoin("jobs j ON j.id = b.job_id").
-		Where(sq.Or{
-			sq.Eq{
-				"b.status": string(BuildStatusStarted),
-			},
-			sq.Eq{
-				"b.status": string(BuildStatusPending),
```
vito: Come to think of it I wonder if we ever needed to even check for pending builds. Why would a pending build have a container? :thinking:

atc/db/migration/migrations/1554304383_add_aborted_to_builds.up.sql
```diff
@@ -0,0 +1,7 @@
+BEGIN;
+
+  ALTER TABLE builds ADD COLUMN aborted boolean DEFAULT false NOT NULL;
+
+  UPDATE builds SET aborted = true, completed = true WHERE status = 'aborted';
```
vito: :+1: this is important so that old builds that were aborted but not completed don't gum up the API and make it think it's the 'next build'

PR #3313, opened 2019-02-14
Manual Trigger user Audit

loghen41:
This is in response in part to https://github.com/concourse/concourse/issues/2752

It only solves half of the issue, specifically focusing on manual triggers.

On a manual trigger, the user_name of the client is appended to the web log.
This serves as a start for an audit trail for manual triggers

Signed-off-by: Logan Hendricks <llhendricks@xactware.com>

PR #3577, opened 2019-03-22
Auditor

loghen41:
This is a single commit comprising all of the auditor changes from my very very broken PR made earlier.


marco-m:
hello, what is the context ? Is there an issue that explains this ?

loghen41:
@marco-m I apologize, I was working on a [pull request](https://github.com/concourse/concourse/pull/3313) for an auditing process that was being reviewed by @pivotal-jwinters . However, I had corrupted my git history on the original pull Request, and needed to open a new Pull request.  This one has a fixed git history now

atc/api/accessor/handler.go
```diff
@@ -3,31 +3,39 @@ package accessor
...
 	action        string
+	auditor       auditor.Auditor
 }
 
 func (h accessorHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	acc := h.accessFactory.Create(r, h.action)
 	ctx := context.WithValue(r.Context(), "accessor", acc)
 
+	if acc != nil {
```
pivotal-jwinters: This nil check shouldn't be necessary

atc/api/accessor/handler_test.go
```diff
@@ -26,7 +27,10 @@ var _ = Describe("Handler", func() {
 
 		dummyHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			innerHandlerCalled = true
-			access = r.Context().Value("accessor").(accessor.Access)
+
+			if r.Context().Value("accessor") != nil {
```
pivotal-jwinters: This shouldn't be necessary either

atc/auditor/auditor.go
```diff
@@ -0,0 +1,175 @@
...
+		EnableWorkerAuditLog:    EnableWorkerAuditLog,
+		EnableVolumeAuditLog:    EnableVolumeAuditLog,
+		logger:                  logger,
+	}
+}
+
+
+type Auditor interface {
+	LogAction(action string, userName string, r *http.Request)
+	ValidateAction(action string) bool
```
pivotal-jwinters: Does `ValidateAction` need to be part of this interface?

Also small nit, but can we rename `LogAction` to `Audit`.

atc/api/accessor/handler.go
loghen41: To explain this nil check, I found that the access object is actually a nil object on a variety of occasions.  For example, when an unauthenticated user is accessing the site, it technically has no access object.

in order to not throw errors in creating an audit log, I found that some level of nil checking was required to confirm that an access object existed.

I am however open to other solutions in order to handle this.

atc/auditor/auditor.go
loghen41: You know, you're right on ValidateAction, I'm really only using that locally.  Let me refactor that accordingly.  And yeah, totally fine to change the name of that method.

let me know if there anything else you would like me to respond to, and sorry it's required this much of an in depth review.

loghen41: @pivotal-jwinters In setting `ValidateAction` to not be a part of the interface, I wanted to know your team's policy regarding internal/external tests.  (as that method won't be public anymore) 

ie. In order to test ValidateAction, I would think to create one set of tests that is part of the `package auditor_test`, and a second set of tests that is part of `package auditor`

From what I could see, that's not a practice really followed anywhere else in the code base, so I wanted to ask: Do you prefer having only external tests? or am I ok to also build internal tests?

atc/api/accessor/handler_test.go
loghen41: See message above

atc/api/accessor/handler.go
pivotal-jwinters: It gets created on line 33. It will never be nil ;)

Also every request through the API goes through this handler. So every request will have a an `access` object, even unauthenticated users. Do you mean its nil in the tests?

atc/auditor/auditor.go
pivotal-jwinters: In general we only test exported methods because anything else is just an implementation detail. 

From the perspective of your tests, the only thing you really care about is if I `Audit` this event does it show up in my logs. You are already implicitly testing the `ValidateAction` method in these tests. So further explicit tests are not required.

atc/api/accessor/handler.go
loghen41: So, in an eating crow moment. I had remembered building this originally without a nil check on the handler and both the application and the tests were throwing errors when trying to run and perform audits on unauthenticated users.

However, as I was preparing a more in depth answer to your question, I can't reproduce the application errors I was seeing beforehand.  (which may mean that I either solved the issue, or am just remembering poorly).

It does look like the unit tests are throwing errors in both [the handler tests](https://github.com/loghen41/concourse/blob/auditor/atc/api/accessor/handler_test.go#L46) and throughout the [api test suite](https://github.com/loghen41/concourse/tree/auditor/atc/api) in general.

I'll remove this nil as it's performing correctly in the application, and adjust the unit tests accordingly.  Sorry to mis-remember this.

atc/auditor/auditor.go
loghen41: To be honest, that was my thought as well. (I've just worked a variety of organizations that have different ideologies when it comes to unit tests, so I tend to go to the most extreme version of unit testing, and scale down from there).

So I just hoped to find what your team would prefer.  I'll remove the explicit tests on the `ValidateAction` method.

atc/api/accessor/handler.go
loghen41: @pivotal-jwinters I just pushed up a change updating all of the unit tests, and removing unneeded nil checks.  Please feel free to review when you get a moment and let me know if you see anything else I missed. 

I'll mark all of these conversations as resolved in the meantime.

atc/api/cli_test.go
```diff
@@ -62,6 +64,12 @@ var _ = Describe("CLI Downloads API", func() {
 
 		err = windowsArchive.WriteZip(zipFile)
 		Expect(err).NotTo(HaveOccurred())
+
+		fakeaccess = new(accessorfakes.FakeAccess)
+	})
+
+	JustBeforeEach(func() {
+		fakeAccessor.CreateReturns(fakeaccess)
```
pivotal-jwinters: If this setup needs to happen in a few different tests it can probably be moved into the [api_suite_test.go](https://github.com/concourse/concourse/blob/master/atc/api/api_suite_test.go), but it's not a big deal.

loghen41: I can probably convert that over if you would like before things get merged in.  Let me know if you want me to tackle that on my end. or if you folks want to do it on your end.

loghen41:
My pleasure, 

thanks for taking the time to go through a bazillion iterations with me :)

atc/api/cli_test.go
loghen41: I went ahead and consolidated those into [api_suite_test.go](https://github.com/loghen41/concourse/blob/fded134857203080926b9b166e913802205f0a2d/atc/api/api_suite_test.go#L208) and removed them from the cli_test, builds_test and info_test.

You should be hopefully good to go. Unit tests passed on my end.I I'll resolve this conversation provided they pass on your end as well.

loghen41:
@pivotal-jwinters It looks like it failed on the `watsj` process, but from what I can review of the test, it doesn't appear to relate to what I was working on. 

Let me know what you need me to do on my end.

pivotal-jwinters:
Yeah probably just flakiness, I re-triggered the job.

vito:
Just a heads-up that we're on the verge of shipping 5.1 (probably Monday) so I'm going to hold off merging to avoid piling-on. I'll take a look shortly after!

loghen41:
@vito That's totally fine. Let me know after you're review if there's anything you need me to update as well.

loghen41:
@vito I assume you folks have been swamped on your end.  Let me know if there is anything on my end I can do to assist.

vito:
@loghen41 Went on vacation for a bit, back now! I'll review this today and try to get it in for 5.2. :slightly_smiling_face:

loghen41:
@vito all good! I hope it was a good vacation. Let me know if there is anything that needs tweaking, or anything I can do to assist on my end.

atc/auditor/auditor.go
```diff
@@ -0,0 +1,174 @@
...
+		return a.EnableVolumeAuditLog
+	default:
+		return false
+	}
+}
+
+func (a *auditor) Audit(action string, userName string, r *http.Request) {
+		err := r.ParseForm()
+		if err == nil &&  a.ValidateAction(action) {
+			a.logger.Info("audit", lager.Data{"command": action, "user": userName, "parameters": r.Form})
```
vito: Would you mind calling this `"action"` instead of `"command"`? Minor nit, but worth checking early since these logs will likely be machine processed at some point.

docker-compose.yml
```diff
@@ -28,6 +28,10 @@ services:
       CONCOURSE_EXTERNAL_URL: http://localhost:8080
       CONCOURSE_ADD_LOCAL_USER: test:test,guest:guest
       CONCOURSE_MAIN_TEAM_LOCAL_USER: test
+      CONCOURSE_ENABLE_BUILD_AUDITING: 'true'
+      CONCOURSE_ENABLE_JOB_AUDITING: 'true'
+      CONCOURSE_ENABLE_PIPELINE_AUDITING: 'true'
+      CONCOURSE_ENABLE_TEAM_AUDITING: 'true'
```
vito: Would you mind backing this change out? I don't think it's necessary as a dev default. Guessing this was checked in from local testing.

loghen41: For sure, I can pull this out, good catch.

atc/auditor/auditor.go
loghen41: Absolutely, I debated back and forth as to put it as "action", I can't remember why I landed on "command", but I can change it.

Issue #1788, opened 2017-11-07
Make web UI show task duration


JohannesRudolph:
# Feature Request

## What challenge are you facing?

Build performance is an important goal for us to keep developer productivity high. We'd like to track build performance to identify slow jobs / tasks in our pipeline. We're already using timeouts to discipline the team to maintain fast builds.

Jobs already have start/end times and duration displayed in the web ui, we'd like to have the same for tasks. 

## A Modest Proposal

Make the UI display task duration for finished tasks in the web ui. Without special support from the backend, the UI could just use the first and last event timestamp to calculate this data. Proper backend support may be more helpful though. 

(originally mentioned in https://github.com/concourse/atc/pull/219#issuecomment-342239963)

marco-m:
I think this is a very needed feature. Not only for the reasons pointed out by @JohannesRudolph, also to help troubleshooting when something is going bizarrely slow.

Lindsayauchin:
We could add it inline with the green chackmark on the task. 

Like so:
![Screen Shot 2019-03-08 at 3 16 45 PM](https://user-images.githubusercontent.com/27746614/54053427-36747a80-41b5-11e9-82ee-002298249278.png)

![Screen Shot 2019-03-08 at 3 16 41 PM](https://user-images.githubusercontent.com/27746614/54053434-396f6b00-41b5-11e9-8268-47bc11a0ac3d.png)

marco-m:
Another possibility would be to make it ever shorter, something like:

                                                                          [1m 06s]

Why ? because also resources are containers, and can take a lot of time to complete (for various reasons). If we follow this reasoning, then each "item" in the web page above, no matter if task or resource, could have on the right its duration.

If we follow this reasoning, then maybe from an implementation point of view it is simpler to do it incrementally, first with the tasks and in a second moment with the resources.

pivotal-jamie-klassen:
Fixed by #3579

PR #3579, opened 2019-03-23
display task duration on finished tasks

mockersf:
for #1788 

task duration is displayed on the right of finished tasks:
task duration


marco-m:
Yes! thanks @mockersf

mockersf:
I went with @marco-m proposition in #1788 for `[1m 06s]` after several tries:
* as `duration 1m 06s` like resources version, but then it wouldn't be possible to add it also to resources
* as `duration 1m 06s` next to the task name, which would be somewhat similar to how duration is displayed for the job but it feels to me like it is too... heavy?
* as the way metadata are displayed for a resource, but it hides the duration somewhat and it may be an issue when adding the duration for a resource

I'm open to suggestion too 😃 

For resource duration, I tried to see if it was the same kind of change but it seems events for resource ends (`FinishGet` and `FinishPut`) don't have a time at least in some cases

vito:
Ah, shame we don't include timestamps for those events. Would you be interested in adding that? I actually think it wouldn't be too hard; we'd just need to add the field to the `FinishGet`/`FinishPut` types and record the time here:

https://github.com/concourse/concourse/blob/90a604c940b240a520b284e9a4ba5ad7062c4189/atc/engine/put_delegate.go#L32-L37

...just like we do for tasks here:

https://github.com/concourse/concourse/blob/90a604c940b240a520b284e9a4ba5ad7062c4189/atc/engine/task_delegate.go#L61-L73

re: where to put the timestamp, it does seem like the UI is a bit cramped. The easiest way to layer it in would maybe be as a tooltip on the step's icon (the checkmark/X on the right). The disadvantage would be that you'd have to mouse-over to reveal the duration. Would that be OK? It would at least de-clutter the UI when you don't actually care, but it could also make it hard for slowness to 'jump out'.

marco-m:
@vito FYI in #1788, @Lindsayauchin added a thumbs up to the proposal of the `[1m 06s]` :-)

Lindsayauchin:
Looks similar to this issue. https://github.com/concourse/concourse/issues/1788 raised by @JohannesRudolph 

Some screens in that issue but, I will add them here for UI reference.

54053434-396f6b00-41b5-11e9-8268-47bc11a0ac3d

54053427-36747a80-41b5-11e9-82ee-002298249278

Lindsayauchin:
Proposing a design similar to issue #1788 that fits nicely into the current UI.  

Screen Shot 2019-03-26 at 2 35 42 PM

mockersf:
@Lindsayauchin that wouldn't work when also displaying durations for resources. With `[]` it gives:
resource duration

mockersf:
@vito I added duration for resources
* I added time to `FinishGet` and `FinishPut` events, that can be empty for previous versions of those events
* I added events `InitializeGet`, `StartGet`, `InitializePut` and `StartPut` similar to task events
* I removed remaining references to an `Initialize` event that were never called

mockersf:
I tried the tooltip, which has the nice side effect fo allowing more info to be displayed:
duration tooltip
initialization duration is selecting a worker, creating the container, ... and step is the duration spent in the actual step

Exist status code could be later added to this tooltip for example

vito:
@mockersf Nifty, I hadn't thought of timing breakdowns but that seems useful, especially when a significant chunk goes just to initialization. I wonder if there's any other useful info to put there in the future. :thinking: 

@marco-m @Lindsayauchin Wdyt of the tooltip UI/UX?

marco-m:
@vito at the beginning I was unsure, but the more I think about it, the more I find that the tooltip works well. I also like the timing breakdown. Thanks again @mockersf !

jchesterpivotal:
> Ah, shame we don't include timestamps for those events. Would you be interested in adding that?

I'm so excited for this change. In the data I scrape it's a noticeable hole.

Related wish: please give finer-grained timestamps. To many things can happen in a 1-second window, especially if caching makes various events "instant".

I'd ask for at least milliseconds, but I'd be happiest with nanoseconds (Golang's internal granularity).

mockersf:
> Related wish: regardless of how it's presented in the UI, please give finer-grained timestamps in the events streamed from the API. To many things can happen in a 1-second window, especially if caching makes various events "instant".
> 
> I'd ask for at least milliseconds, but I'd be happiest with nanoseconds (Golang's internal granularity).

Adding a more precise time in the event shouldn't be complicated, but there's an issue with format and backward compatibility:
* Changing the `time` field to be an int in nanoseconds would break existing clients and not be a standard
* Changing the `time` field to be a float in nanoseconds with the integer part for the second may break clients, but the UI would still work and gain precision. It would require additional computing on the server side
* Adding a new field could have one of the two format above, or use RFC3339 (Golang default) for a string format

what do you think ?

jchesterpivotal:
> what do you think ?

I prefer clock time elapsed since epoch. RFC3339 timestamps are calendar time, not clock time: they require external information for correct interpretation (timezones, daylight savings, leap seconds etc). Clock time ignores these additional constructs. A thousand years from now the nanoseconds elapsed from epoch until the time a stamp was taken will have exactly the same value. But calendars change all the time, a record kept as a calendrical timestamp is subject to unexpected -- and hard-to-detect -- corruption due to factors outside the record.

For example, the dates 3rd to the 13th of September, 1752, never happened in the British Empire. They were skipped due to a change from the Julian to Gregorian calender. If I had recorded a timestamp at the time (say 4th Sept 1752), it would become meaningless without additional processing. But if I had an epoch stamp (supposing nanoseconds since 1700, say), I would still have a universally reliable concept of when the event happened.

Both ways of expressing points in time are affected by clock drift, clock accuracy, clock precision and relativistic effects.

As for the format, I'd prefer an `int64` encoding nanoseconds since epoch (even if it's rounded off to whatever precision to OS allows). I don't think there is an compatibility guarantee for the API and clients can use `/v1/info` to see what version of Concourse they're talking to. Using an `int64` for nanoseconds gives approximately 580 years of headroom before it overflows.

mockersf:
I switched the `time` fields to be a time in nanoseconds, and the UI to display milliseconds details if it's below 1 second.

According to [Golang's doc](https://golang.org/pkg/time/#Time.UnixNano), it will start failing in 2263

jchesterpivotal:
> @jchesterpivotal Can you provide more details on why such high precision is important to you? I get that a lot can happen in a second, but transitioning to nanoseconds is nontrivial, and it's not like we're taking these measurements in a vacuum. 

Partly because of the limits of Postgresql interval types, which behave badly in interval queries when they are "empty". Nanoseconds are the finest granularity I can ever expect to get in my lifetime.

Right now millisecond resolution will probably avoid problems most of the time, but I will still need special cases because a lot of the time loading from the cache will be sub-millisecond due to file page caching. Microseconds are probably going to be enough for some time, but if you're going as far as microseconds ...

marco-m:
> There are a lot of moving parts behind a running build (including network calls to the database), and at the subsecond level it seems likely you'd get a whole lot of noise rather than a more precise signal

+1

jchesterpivotal:
I accept noise that will reduce the accuracy of my figures, but precision is still useful for me, because I need sufficient precision to create a reliable ordering of events with a lifetime > 0.

jchesterpivotal:
> Could you use the event ID instead? That's ultimately what determines the order of events from Concourse's perspective.

I miss out on the interval querying, which is super useful to me. If I have time intervals I can ask questions like "how many builds were active during <arbitrary window of time>?" 

But PostgreSQL, at least for version 9.6, interprets `[123,123)` as being `[123,)`. Then the `[123,)` doesn't show up properly in interval queries because it's not "between" anything. I can query occurs-before and occurs-after, but not occurs-during, occurs-overlapping etc.

mockersf:
> The nanosecond change kind of explodes the complexity/risk of this, but if you're willing to follow through with the changes necessary, then I'm cool with it. 👍

As there are more changes than what I did, I think it would be better to split the PR and keep this one just to display duration of steps in the UI, and then I could continue on moving to nanoseconds in another PR.

if you agree I'll remove the change to nanoseconds from this one and start another

jchesterpivotal:
> I think it would be better to split the PR 

Yes, I think so. I somewhat derailed the original purpose. Sorry :|

vito:
@mockersf @jchesterpivotal Sounds good. Happy to discuss nanosecond precision stuff in another PR. :+1:

mockersf:
@vito done, everything is back to seconds 🦊

vito:
@mockersf Sorry, looks like #3664 introduced some conflicts. Once this is rebased I'll keep an eye out for it and merge so you don't hit another conflict. :+1:

mockersf:
@vito it's rebased !

JohannesRudolph:
I may be missing something... but has this made its way into concourse 5.2.0? Because I don't see no task durations on my dashboard:-(

JohannesRudolph:
Ah... it's on hover. Nevermind.

Issue #165, opened 2015-10-14
Fail fast when one aggregated task fails


bendodd:
Given a set of more than one aggregated/concurrent tasks, can the job "fail fast" as soon as one of the tasks in the set fail? 

Especially useful when tests have been parallelised across multiple workers to reduce build time...

vito:
FWIW, I think this would be a pretty easy change to the [`aggregate` step](https://github.com/concourse/concourse/blob/master/atc/exec/aggregate.go). The only change necessary would be to give the sub-steps their own context and cancel it on the first `err` read from the `errs` channel here:

https://github.com/concourse/concourse/blob/02a90a4d0e99ca00767bda0b83189215543efe45/atc/exec/aggregate.go#L32-L34

This should be an optional param though. `fail_fast: true`?

vito:
This was implemented by https://github.com/concourse/concourse/pull/3580 :+1: - it'll be in 5.2

PR #3580, opened 2019-03-24
Add in_parallel step

### LINKED ISSUE #2628: Feature request: parameter to limit parallelism on aggregate steps

I have a pipeline which runs tests on pull requests for a monolithic repository containing terraform. Initially I wanted to use `aggregate:` so that the execution time stayed the same no matter how many directories were being tested. But as the number of directories increased, we encountered issues with the number of sockets that were opened on a given worker (and failing to open new ones).

Part of the problem might be that we are using a monolithic repository which means that a single volume containing the source code is needed by all the tasks in the `aggregate:` step, which means they tend to all end up on the same worker (Using `random` container placement is not really a good solution for us, since most other pipelines don't have this problem and would all be penalised from not having reliable caches).

Currently we have to manually batch the tasks together (while trying to balance the number of tasks running in parallel, with the execution time of each task :joy:) to avoid the issue...

I'd like to be able to specify a large number of tasks in an `aggregate:` step, and be able to limit how many of them should be running at any given time until all of them complete (in other words, I want to be able to limit the maximum parallelism of an aggregate step).

For example:
```yml
- aggregate:
  max_parallelism: 5
  - task: A
  ...
  - task: Z
```

Would start 5 random tasks to begin with, and then keep exactly 5 tasks running until they all had completed. `max_parallelism` would be equivalent to `max_in_flight` for jobs (and could probably have the same name?).

Would love to contribute, but I would need some hand-holding to even figure out where to start looking to implement these changes 😅

itsdalmo:
Closes #2628 by adding support for the following:

```yaml
jobs:
  - name: example
    plan:
    - get: every-minute
      trigger: true
    - in_parallel:
        limit: 1
        fail_fast: true
        steps:
        - task: echo
          config:
           ...
        - task: echo
          config:
            ...
```

InParallel ensures that `limit` steps are running at any one time and until all steps are complete. If `limit` is not specified, the default behavior is identical to an `aggregate` step and runs everything at once. Setting `limit` to 1 makes it behave like `do`. When cancelling a job, `in_parallel` will not schedule (`Run()`) the pending tasks and instead return the `ctx.Err()` as soon as possible. When setting `fail_fast`, `in_parallel` will abort all currently running tasks and refrain from scheduling pending tasks.

### Todo

- [ ] Revert changes to `examples/hello-world-every-minute.yml`.
- [x] Submit a PR to [concourse/docs](https://github.com/concourse/docs).
- [ ] Decide on YAML structure and naming (questions in [this comment](https://github.com/concourse/concourse/issues/2628#issuecomment-475908990)).

PS: Sorry for the big PR, but `parallel` is essentially a copy/paste of `aggregate` with very slight modifications. Was originally thinking I could refactor so both used the same underlying implementation, but I think that would only allow us to reuse the code in `exec` which does not have that big of an impact on the diff.

### DIFF
--- a/atc/config.go
+++ b/atc/config.go
@@ -298,6 +298,35 @@ type InParallelConfig struct {
 	FailFast bool         `yaml:"fail_fast,omitempty" json:"fail_fast,omitempty" mapstructure:"fail_fast"`
 }
 
+func (c *InParallelConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var data interface{}
+
+	err := unmarshal(&data)
+	if err != nil {
+		return err
+	}
+
+	switch actual := data.(type) {
+	case []interface{}:
+		if err := unmarshal(&c.Steps); err != nil {
+			return fmt.Errorf("failed to unmarshal parallel steps: %s", err)
+		}
+	case map[interface{}]interface{}:
+		// Used to avoid infinite recursion when unmarshalling this variant.
+		type target InParallelConfig
+
+		var t target
+		if err := unmarshal(&t); err != nil {
+			return fmt.Errorf("failed to unmarshal parallel config: %s", err)
+		}
+		c.Steps, c.Limit, c.FailFast = t.Steps, t.Limit, t.FailFast
+	default:
+		return fmt.Errorf("wrong type for parallel config: %v", actual)
+	}
+
+	return nil
+}
+
 // A PlanConfig is a flattened set of configuration corresponding to
 // a particular Plan, where Source and Version are populated lazily.
 type PlanConfig struct {
--- a/atc/config.go
+++ b/atc/config.go
@@ -308,6 +308,13 @@ type PlanConfig struct {
 	// corresponds to an Aggregate plan, keyed by the name of each sub-plan
 	Aggregate *PlanSequence `yaml:"aggregate,omitempty" json:"aggregate,omitempty" mapstructure:"aggregate"`
 
+	// a nested chain of steps to run in parallel
+	Parallel *PlanSequence `yaml:"in_parallel,omitempty" json:"in_parallel,omitempty" mapstructure:"in_parallel"`
+	// limit parallelism for a Parallel plan
+	MaxInParallel int `yaml:"max_in_parallel,omitempty" json:"max_in_parallel,omitempty" mapstructure:"max_in_parallel"`
+	// cancel a parallel step on first error
+	FailFast bool `yaml:"fail_fast,omitempty" json:"fail_fast,omitempty" mapstructure:"fail_fast"`
+
 	// corresponds to Get and Put resource plans, respectively
 	// name of 'input', e.g. bosh-stemcell
 	Get string `yaml:"get,omitempty" json:"get,omitempty" mapstructure:"get"`
--- a/atc/config.go
+++ b/atc/config.go
@@ -292,6 +292,12 @@ func (c InputsConfig) MarshalJSON() ([]byte, error) {
 	return json.Marshal("")
 }
 
+type InParallelConfig struct {
+	Steps    PlanSequence `yaml:"steps,omitempty" json:"steps" mapstructure:"steps"`
+	Limit    int          `yaml:"limit,omitempty" json:"limit,omitempty" mapstructure:"limit"`
+	FailFast bool         `yaml:"fail_fast,omitempty" json:"fail_fast,omitempty" mapstructure:"fail_fast"`
+}
+
 // A PlanConfig is a flattened set of configuration corresponding to
 // a particular Plan, where Source and Version are populated lazily.
 type PlanConfig struct {
@@ -309,11 +315,7 @@ type PlanConfig struct {
 	Aggregate *PlanSequence `yaml:"aggregate,omitempty" json:"aggregate,omitempty" mapstructure:"aggregate"`
 
 	// a nested chain of steps to run in parallel
-	Parallel *PlanSequence `yaml:"in_parallel,omitempty" json:"in_parallel,omitempty" mapstructure:"in_parallel"`
-	// limit parallelism for a Parallel plan
-	MaxInParallel int `yaml:"max_in_parallel,omitempty" json:"max_in_parallel,omitempty" mapstructure:"max_in_parallel"`
-	// cancel a parallel step on first error
-	FailFast bool `yaml:"fail_fast,omitempty" json:"fail_fast,omitempty" mapstructure:"fail_fast"`
+	InParallel *InParallelConfig `yaml:"in_parallel,omitempty" json:"in_parallel,omitempty" mapstructure:"in_parallel"`
 
 	// corresponds to Get and Put resource plans, respectively
 	// name of 'input', e.g. bosh-stemcell
--- a/atc/job_config_test.go
+++ b/atc/job_config_test.go
@@ -561,14 +561,16 @@ var _ = Describe("JobConfig", func() {
 				BeforeEach(func() {
 					jobConfig.Plan = atc.PlanSequence{
 						{
-							Parallel: &atc.PlanSequence{
-								{Get: "a"},
-								{Put: "y"},
-								{Get: "b", Resource: "some-resource", Passed: []string{"x"}},
-								{Get: "c", Trigger: true},
+							InParallel: &atc.InParallelConfig{
+								Steps: atc.PlanSequence{
+									{Get: "a"},
+									{Put: "y"},
+									{Get: "b", Resource: "some-resource", Passed: []string{"x"}},
+									{Get: "c", Trigger: true},
+								},
+								Limit:    1,
+								FailFast: true,
 							},
-							MaxInParallel: 1,
-							FailFast:      true,
 						},
 					}
 				})
@@ -600,18 +602,22 @@ var _ = Describe("JobConfig", func() {
 				BeforeEach(func() {
 					jobConfig.Plan = atc.PlanSequence{
 						{
-							Parallel: &atc.PlanSequence{
-								{
-									Parallel: &atc.PlanSequence{
-										{Get: "a"},
+							InParallel: &atc.InParallelConfig{
+								Steps: atc.PlanSequence{
+									{
+										InParallel: &atc.InParallelConfig{
+											Steps: atc.PlanSequence{
+												{Get: "a"},
+											},
+											Limit: 1,
+										},
 									},
-									MaxInParallel: 1,
+									{Get: "b", Resource: "some-resource", Passed: []string{"x"}},
+									{Get: "c", Trigger: true},
 								},
-								{Get: "b", Resource: "some-resource", Passed: []string{"x"}},
-								{Get: "c", Trigger: true},
+								Limit:    2,
+								FailFast: true,
 							},
-							MaxInParallel: 2,
-							FailFast:      true,
 						},
 					}
 				})
@@ -687,8 +693,10 @@ var _ = Describe("JobConfig", func() {
 						{
 							Aggregate: &atc.PlanSequence{
 								{
-									Parallel: &atc.PlanSequence{
-										{Put: "a"},
+									InParallel: &atc.InParallelConfig{
+										Steps: atc.PlanSequence{
+											{Put: "a"},
+										},
 									},
 								},
 								{Put: "b", Resource: "some-resource"},
--- a/atc/job_config_test.go
+++ b/atc/job_config_test.go
@@ -557,6 +557,88 @@ var _ = Describe("JobConfig", func() {
 				})
 			})
 
+			Context("when a simple parallel plan is the first step", func() {
+				BeforeEach(func() {
+					jobConfig.Plan = atc.PlanSequence{
+						{
+							Parallel: &atc.PlanSequence{
+								{Get: "a"},
+								{Put: "y"},
+								{Get: "b", Resource: "some-resource", Passed: []string{"x"}},
+								{Get: "c", Trigger: true},
+							},
+							MaxInParallel: 1,
+							FailFast:      true,
+						},
+					}
+				})
+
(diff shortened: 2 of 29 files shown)


atc/plan.go
```diff
@@ -5,6 +5,7 @@ type Plan struct {
 	Attempts []int  `json:"attempts,omitempty"`
 
 	Aggregate *AggregatePlan `json:"aggregate,omitempty"`
+	Parallel  *ParallelPlan  `json:"parallel,omitempty"`
```
vito: :thinking: The name feels slightly off; most of them are adjectives or verbs (except `task` I guess).

What do you think of `in_parallel`? I kinda like that as it fits alongside `on_success`/`do`/etc.

atc/config.go
```diff
@@ -306,6 +306,11 @@ type PlanConfig struct {
 	// corresponds to an Aggregate plan, keyed by the name of each sub-plan
 	Aggregate *PlanSequence `yaml:"aggregate,omitempty" json:"aggregate,omitempty" mapstructure:"aggregate"`
 
+	// a nested chain of steps to run in parallel
+	Parallel *PlanSequence `yaml:"parallel,omitempty" json:"parallel,omitempty" mapstructure:"parallel"`
+	// limit paralelism for a Parallel plan
+	MaxInParallel int `yaml:"max_in_parallel,omitempty" json:"max_in_parallel,omitempty" mapstructure:"max_in_parallel"`
```
vito: This name feels awkward but I'm not sure what else to suggest. Maybe it just sounds weird because I'm used to `max_in_flight`. I thought it might make sense to reuse the same name but @clarafu pointed it it might be confusing to have both set - someone could assume `max_in_flight: 3` at the job level is like setting it on all the steps in the plan. Maybe we should reserve 'in flight' terminology for builds, then? :thinking: If so this name is probably fine.

itsdalmo:
> My take is that the coverage may be a bit excessive owing to aggregate's storied history (it's one of the oldest step types). It might make sense to replace some of the more verbose/brittle tests with testflight tests instead. Or maybe they're actually fine and not the brittle/verbose tests I'm remembering from aggregate. Like I said I haven't reviewed thoroughly yet. 🙂

You are right, it is the very same tests as in `aggregate` 😂I'm happy to go over the tests and rewrite them so that we only keep essential tests; there are a lot of tests to ensure that other steps/hook work within a `parallel`, which are probably safe to remove since each step is just executed with `step.Run(ctx, state)`.

I can leave a comment on the tests I think can be deleted.

atc/scheduler/factory/factory_parallel_test.go
```diff
@@ -0,0 +1,210 @@
...
+						Name:                   "some other thing",
+						VersionedResourceTypes: resourceTypes,
+					}),
+				},
+			})
+			Expect(actual).To(Equal(expected))
+		})
+	})
+
+	Context("when I have nested parallel steps", func() {
```
itsdalmo: I don't see how this can fail if the buildFactory can produce a plan from a `ParallelPlan`. I.e. the implementation is already identifying `Parallel` and recursing to produce a plan for substeps that may include another `parallel`. Delete?

atc/scheduler/factory/factory_parallel_test.go
```diff
@@ -0,0 +1,210 @@
...
+							}),
+						},
+					}),
+				},
+			})
+			Expect(actual).To(Equal(expected))
+		})
+	})
+
+	Context("when I have a parallel with hooks", func() {
```
itsdalmo: Delete? If hooks work for any other step type, it should work for Parallel?

atc/scheduler/factory/factory_parallel_test.go
```diff
@@ -0,0 +1,210 @@
...
+						}),
+					}),
+				},
+				MaxInParallel: 1,
+			})
+			Expect(actual).To(Equal(expected))
+		})
+	})
+
+	Context("when I have a hook on a parallel", func() {
```
itsdalmo: Delete? If hooks work for any other step type, it should work for Parallel?

atc/job_config_test.go
```diff
@@ -502,6 +502,86 @@ var _ = Describe("JobConfig", func() {
 				})
 			})
 
+			Context("when a simple parallel plan is the first step", func() {
```
itsdalmo: I think this test, and the ones following, are mostly testing `get` in relation to `PlanSequence` and not `aggregate` or `parallel` specifically. But if the desire is to move from `aggregate` to `parallel` in the future, then I think it makes sense to just keep them?

itsdalmo: Same with the `exec_engine` tests. If the plan is to eventually move from `aggregate` to `parallel` then I think it makes sense to keep the tests; unless they can be cut completely?

atc/plan.go
itsdalmo: Makes sense, I've changed it to `in_parallel` 👍

atc/config.go
itsdalmo: I see what you mean, and I agree that it's best to keep `max_in_flight` separate since it could get a bit confusing. Especially when you consider indentation errors in the YAML, which would produce unexpected results if the keyword is recognised in both the job and plan? 

However, right now we have `in_parallel` and `max_in_parallel` - not sure if I like it or not 🤔If I'm thinking outside the box, I would suggest something like this:

- Add a `PlanSequence` in `atc.Config` which we call `steps`.
- Keep `in_parallel` but change it to take a pointer to an `int`.
- Add `in_sequence` and have it take a pointer to a `string`.

We'd use the pointers so that the values would be `nil` if the key was not present in the map, so we can return an error from `validate` if `steps` is specified without either `in_sequence` or `in_parallel`. Then we could represent all three desired functionalities:

```yaml
jobs:
  - name: example
    plan:
    - in_parallel: 3
      steps:
      - task: echo
        config:
         ...
    - in_parallel: # <- We see a 0, and set the limit to the length of the steps. Same as aggregate.
      steps:
      - task: echo
        config:
         ...
    - in_sequence: # <- Equivalent to "do"?
      steps:
      - task: echo
        config:
         ...
```

Might work 🤔

vito: I think the last two options there are a little awkward, since knowing how the YAML is parsed, those keys actually point to blank values, and Concourse's parser would have to be careful to notice that they're there at all (since they'd be set to either empty strings or nil - I don't remember).

The first option feels natural but forces you to pick a number in scenarios where you don't really care (unless we support the later two options).

I agree that `in_parallel` and `max_in_parallel` feel a little weird, but it at least allows the common case of "I just want things to run in parallel" to look simple:

```yaml
in_parallel:
- task: echo1
  file: config.yml
- task: echo2
  file: config.yml
- task: echo3
  file: config.yml
```

What if we allow two forms of `in_parallel`?

The first, simplest form which takes no configuration:

```yaml
in_parallel:
- task: echo1
  file: config.yml
- task: echo2
  file: config.yml
- task: echo3
  file: config.yml
```

...and in more complicated scenarios, it's instead an object with `limit` and `steps`:

```
in_parallel:
  limit: 3
  steps:
  - task: echo1
    file: config.yml
  - task: echo2
    file: config.yml
  - task: echo3
    file: config.yml
```

This could also be extended to support `fail_fast: true` (#165).

On the other hand I could see the repetition of `in_parallel` in `max_in_parallel` being a strength of the name, given that they *are* very closely tied together. :man_shrugging: 

Any preference?

atc/scheduler/factory/factory_parallel_test.go
vito: Agreed

atc/scheduler/factory/factory_parallel_test.go
vito: Agreed

atc/scheduler/factory/factory_parallel_test.go
vito: Agreed

atc/job_config_test.go
vito: I think these tests are just covering the recursing through all the different types of steps which contain other steps. It probably makes sense to keep it, since we need to recurse through the `Parallel` fields as well. This is necessary for finding all the 'job inputs', which are determined by looking through the entire plan for `get` steps.

atc/config.go
itsdalmo: I think in your example, `in_parallel` would only be able to cover the current `aggregate` and the proposed limited parallelisation, and in that case I would suggest we just stick with `in_parallel` and `max_in_parallel` - mainly because I suspect indentation might trip users up? 

We could still add `fail_fast` to this PR and perhaps throw an error if this key is present in a non-parallel context? However, it seems like the current `atc.Validate()` doesn't really do strict validation for the plan configuration, so perhaps it's fine to just ignore `fail_fast` if it's specified in another context? 🤔

jchesterpivotal:
I'd feel more comfortable if `aggregate` got an extra dial (`max_in_flight`, `throttle`, whatever), rather than having an entirely new step name that has overlapping functionality.

marco-m:
I agree with @jchesterpivotal here. Although the name `parallel` is more suitable than `aggregate`, having both names, besides code duplication, also adds user confusion, and I see already the same pipeline sporting both. +1 to add the max_in_flight to the existing aggregate.

pivotal-jwinters:
@vito I tend to agree with @jchesterpivotal as well. I know you wanted to migrate away from the `aggregate` step as per [this comment](https://github.com/concourse/concourse/issues/2628#issuecomment-472133539), but maybe we can just alias the name and deprecate the `aggregate` name over time?

itsdalmo:
The code changes are minimal, so all we currently need is a decision on:

- Add `parallel` step or extend `aggregate`?
- What should the config flag be called: `max_in_flight`, `max_in_parallel`, `limit` or `throttle`?

And keep in mind that if we extend `aggregate`, we have to add the above flag at the root of `atc.PlanConfig` (not nested inside a struct), less we will break existing pipelines. If we add `parallel` we are free to do whatever we want, at the cost of some code duplication and confusion (unless we are saying that `aggregate` is deprecated?).

I'm happy to implement this whatever way you guys want so as to get #2628 closed. Only thing I have to add is that `max_in_flight` might get "overloaded" if it is used in too many places.

vito:
@jchesterpivotal @marco-m @pivotal-jwinters I don't plan on supporting both `aggregate` and `in_parallel` forever - the point was to give users a reason to migrate to the new name by giving it more useful functionality (max-in-flight, fail-fast) and deprecate `aggregate` as `in_parallel` adoption grows.

Having an alias seems like it would be even more confusing - you've then shared the internal implementation, which users don't see, but duplicated *all* functionality, which users *do* see. Wouldn't it be clearer to have a deprecated, less-useful step and a newer step with a more meaningful name and more functionality?

marco-m:
@vito then I would suggest the following:
- introduce a check that returns error if the same pipeline has both `aggregate` and `in_parallel`. This check should be introduced in this PR or immediately after. The error message should mention that `aggregate` is deprecated.
- Deprecate `aggregate` immediately starting with 5.1. Add a check to `fly` that prints a deprecation warning and mentions _when_ aggregate will be removed.
- Remove `aggregate` in 6.0.

vito:
@marco-m Agreed. I'm not *too* sure about mixed `aggregate`/`in_parallel` being an error vs. a deprecation warning, but we at least have all the plumbing for emitting warnings.

@itsdalmo How's that sound?

Plumbing for deprecations:

https://github.com/concourse/concourse/blob/41f54f59330f2b6be55e05eb56166cb97494a64a/atc/validate.go#L27:17

->

https://github.com/concourse/concourse/blob/41f54f59330f2b6be55e05eb56166cb97494a64a/atc/api/configserver/save.go#L110

->

https://github.com/concourse/concourse/blob/41f54f59330f2b6be55e05eb56166cb97494a64a/fly/commands/internal/setpipelinehelpers/atc_config.go#L83

clarafu:
If we are trying to deprecate the `aggregate` step, then I would think that creating a new `parallel` step makes sense?

I feel like using the `max_in_flight` keyword again would be quite confusing especially if you have a very long pipeline config and you arn't sure if the `max_in_flight` is for the aggregate step or for the build plan, in which the keyword means two completely different things.

marco-m:
@clarafu I agree, `max_in_parallel` seems more appropriate.

itsdalmo:
@vito - I've added `fail_fast` in the latest commits. Struggled a bit with this one because it seems like `Run()` for the steps do not return an error when the task script fails - i.e. the only way I could get the expected behaviour was by using `s.Succeeded()`. Am I doing something wrong or is this working as intended?

In the current commit, `in_parallel` with `fail_fast: true` will interrupt currently running steps, stop scheduling pending tasks. This is the behaviour I would expect myself, so that is what I implemented, but perhaps you had something else in mind?

Also, I'm wondering if perhaps the configuration should _only_ be your second suggestion:

```yaml
in_parallel:
  limit: 3
  steps:
  - task: echo1
    file: config.yml
  - task: echo2
    file: config.yml
  - task: echo3
    file: config.yml
  fail_fast: true
```

Seems cleaner when related configuration goes in a map like this? What do you think?

itsdalmo:
@vito I added support for the shorthand form of `in_parallel` in the latest commit. I have not added custom serialization for YAML/JSON, so `get-pipeline` and `set-pipeline` will always return/diff `in_parallel: [ ... ]` as `in_parallel: { steps: [ ... ] }`. Returning the long form of the config feels clean to me since e.g. YAML anchors will be expanded on `get` and `set-pipeline` also - but let me know if you want it changed. Also added a deprecation warning if `aggregate` is set, as part of the validation.

Thoughts?

itsdalmo:
Thoughts on this @vito? Are we approaching mergeable? 😅

vito:
@itsdalmo Sorry! I was focusing on getting 5.1 out the door, which we were on the verge of for a while, but it stretched on for longer than expected. I'll take another look at this today and merge if it's good to go. :+1:

PR #3590, opened 2019-03-26
ci: release pipeline changes

vito:
These are all the changes that were necessary for 5.0.1. There's also a release template pipeline, which should be re-usable for future minor releases.

Each commit has a little more context in its description. :+1:

### DIFF
--- a/ci/pipelines/release.yml
+++ b/ci/pipelines/release.yml
@@ -2,6 +2,7 @@
 #
 #   ((release_major)) the MAJOR version, e.g. 5
 #   ((release_minor)) the MAJOR.MINOR version, e.g. 5.1
+#   ((concourse_smoke_deployment_name)) a unique name for the smoke bosh deployment
 #
 # the following git branches need to be created:
 #
--- a/ci/pipelines/release.yml
+++ b/ci/pipelines/release.yml
@@ -312,7 +312,7 @@ jobs:
       file: concourse/ci/tasks/k8s-delete.yml
       params:
         KUBE_CONFIG: ((kube_config))
-        RELEASE_NAME: concourse-smoke-patch # TODO can't use release_minor
+        RELEASE_NAME: ((concourse_smoke_deployment_name))
         CONCOURSE_IMAGE: concourse/concourse-rc
   - task: deploy
     image: unit-image
@@ -320,21 +320,21 @@ jobs:
     file: concourse/ci/tasks/k8s-deploy.yml
     params:
       KUBE_CONFIG: ((kube_config))
-      RELEASE_NAME: concourse-smoke-patch # TODO can't use release_minor
+      RELEASE_NAME: ((concourse_smoke_deployment_name))
       CONCOURSE_IMAGE: concourse/concourse-rc
   - task: k8s-smoke
     image: unit-image
     file: concourse/ci/tasks/k8s-smoke.yml
     params:
       KUBE_CONFIG: ((kube_config))
-      RELEASE_NAME: concourse-smoke-patch # TODO can't use release_minor
+      RELEASE_NAME: ((concourse_smoke_deployment_name))
       MAX_TICKS: 180
   - task: delete
     image: unit-image
     file: concourse/ci/tasks/k8s-delete.yml
     params:
       KUBE_CONFIG: ((kube_config))
-      RELEASE_NAME: concourse-smoke-patch # TODO can't use release_minor
+      RELEASE_NAME: ((concourse_smoke_deployment_name))
       CONCOURSE_IMAGE: concourse/concourse-rc
 
 - name: k8s-topgun
@@ -578,7 +578,7 @@ jobs:
       stemcells:
       - gcp-xenial-stemcell/*.tgz
       vars:
-        deployment_name: concourse-smoke-patch # TODO can't use release_minor
+        deployment_name: ((concourse_smoke_deployment_name))
   - task: discover-bosh-endpoint-info
     file: concourse/ci/tasks/discover-bosh-endpoint-info.yml
     image: unit-image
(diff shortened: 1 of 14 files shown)


ci/tasks/scripts/docker-semver-tags
```diff
@@ -0,0 +1,21 @@
...
+minor_version=$(echo $version | cut -d. -f1,2)
+
+latest_minor_version=$(echo $latest_version | cut -d. -f1,2)
+
+echo $version       >> tags/tags
+echo $minor_version >> tags/tags
+
+# if the version we're publishing is for the latest minor version, bump the
+# major and latest tags, too
+if [ "$minor_version" = "$latest_minor_version" ]; then
```
xtreme-sameer-vohra: Might be missing something but don't we want major & latest to always be bumped as well when publishing a new image ?

xtreme-sameer-vohra: Or we only apply those tags after sometime X manually when we feel the image has baked sufficiently ?

vito: It's a bit tricky. We don't want to bump `5` or `latest` if the current latest release is 5.2.x and we just shipped 5.1.1 from a release pipeline. That would effectively "downgrade" them to an older minor release. Does that make sense?

ci/pipelines/release.yml
```diff
@@ -0,0 +1,1192 @@
...
+    timeout: 1h
+    file: concourse/ci/tasks/docker-compose-watsjs.yml
+
+- name: upgrade
+  public: true
+  serial: true
+  plan:
+  - aggregate:
+    - get: concourse
+      passed: [unit, dev-image]
```
xtreme-sameer-vohra: @vito should we update `concourse.yml` to also include both job passed constraints ? Looks like it only contains `dev-image` at the moment

ci/pipelines/release.yml
```diff
@@ -0,0 +1,1192 @@
...
+      trigger: true
+    - get: dev-image
+      passed: [dev-image]
+      params: {format: oci}
+      trigger: true
```
xtreme-sameer-vohra: Same for this, should this also be in `concourse.yml` ?

ci/pipelines/release.yml
```diff
@@ -0,0 +1,1192 @@
...
+    image: unit-image
+    file: concourse/ci/tasks/upgrade-test.yml
+
+- name: downgrade
+  public: true
+  serial: true
+  plan:
+  - aggregate:
+    - get: concourse
+      passed: [unit, dev-image]
```
xtreme-sameer-vohra: Same comment as for `upgrade`

ci/pipelines/release.yml
```diff
@@ -0,0 +1,1192 @@
...
+      trigger: true
+    - get: dev-image
+      passed: [dev-image]
+      params: {format: oci}
+      trigger: true
```
xtreme-sameer-vohra: Same comment as for `upgrade`

ci/pipelines/release.yml
```diff
@@ -0,0 +1,1192 @@
...
+        KUBE_CONFIG: ((kube_config))
+        RELEASE_NAME: concourse-smoke-patch # TODO can't use release_minor
+        CONCOURSE_IMAGE: concourse/concourse-rc
+  - task: deploy
+    image: unit-image
+    input_mapping: {image-info: concourse-rc-image}
+    file: concourse/ci/tasks/k8s-deploy.yml
+    params:
+      KUBE_CONFIG: ((kube_config))
+      RELEASE_NAME: concourse-smoke-patch # TODO can't use release_minor
```
xtreme-sameer-vohra: @vito would it make sense to still parametrize this so that there isn't a chance of a conflict if multiple patch pipelines run simultaneously ?

ci/pipelines/release.yml
```diff
@@ -0,0 +1,1192 @@
...
+    params:
+      manifest: concourse/ci/deployments/bosh-smoke.yml
+      releases:
+      - concourse-release/*.tgz
+      - postgres-release/*.tgz
+      - bpm-release/*.tgz
+      stemcells:
+      - gcp-xenial-stemcell/*.tgz
+      vars:
+        deployment_name: concourse-smoke-patch # TODO can't use release_minor
```
xtreme-sameer-vohra: would it make sense to still parametrize this so that there isn't a chance of a conflict if multiple patch pipelines run simultaneously ?

ci/tasks/scripts/docker-semver-tags
xtreme-sameer-vohra: doh, yes, that does

xtreme-sameer-vohra:
Added some comments but overall lgtm
Some thoughts that may be beyond the scope of this story but worth considering;
- how we'd trim/modify the pipeline when Concourse gets bumped to a new Major.Minor (eg. 4.2.x wouldn't be shipping any docker image artifacts)
- Stemcell version to employ (boshio is always ahead of pivnet) - should we consider testing the latest on either and then drop boshio for a patch pipeline ?

vito:
@xtreme-sameer-vohra Pushed a few more commits that you may want to review, and integrated your suggestions too. :+1: 



> * how we'd trim/modify the pipeline when Concourse gets bumped to a new Major.Minor (eg. 4.2.x wouldn't be shipping any docker image artifacts)

Not sure what you mean by this; I don't think we need to trim them, they should still be able to publish Docker images. The 4.2.x pipeline for example would only push '4.2.5' and '4.2' tags - just not 'latest'. Although, we may want it to bump '4' - looks like I need to tweak that bump script. I'll do that now.

> * Stemcell version to employ (boshio is always ahead of pivnet) - should we consider testing the latest on either and then drop boshio for a patch pipeline ?

The plan for the release pipelines is to have them also continuously bump their dependencies (`gdn`, stemcells, etc.) so we can stay continuously integrated in case there's a CVE fix in one of them.

xtreme-sameer-vohra:
@vito lgtm 👍

Issue #1460, opened 2017-08-09
Add `on_error` attribute to allow outside sources to be notified of CI failure 


AdamDorwart:
# Feature Request

## What challenge are you facing?
When jobs fail that use a github PR resource, they get stuck with the last build badge that they were set with. To the user on Github, they see their job 'pending' forever. It would be nice to try to put a error badge on the build. `ensure` doesn't quite work because what badge should be put is different based on the outcome of the build.

Also another easy way to see the disconnect:
Concourse Job Events : Github Build Status
task -> pending
on_success -> success
on_failure -> failure
? -> error

## A Modest Proposal
Add an `on_error` attribute so jobs that have an error can be properly communicated to external services.

jtarchie:
This might be difficult in some cases. Sometimes the error state is because containers cannot be reached/created. If that error occurs, we have to watch out for the infinite loop.

vito:
This is hopefully something you'd only want for notifications. I don't think we should add an `on_error` - this probably points to more exploration needing to be done around the many UX problems resulting from using resources for notifications.

shinmyung0:
I also noticed this, when I am using something like the `pool` resource to manage cloud foundry spaces, sometimes my jobs will `error` due to mysterious `failed to create volume` messages. This results in the job not triggering the `on_failure` step that I have put to release the space's lock. 

Could we just have errors maybe triggering these `on_failure` events? Not sure what the proper way to do this currently is.

akamalov:
@vito is it possible to get an ETA on ```on_error``` attribute making it to Concourse release? Currently, one of our pipelines errors (not fails) on S3 upload. Since it does not fit to category ```on_failure```, there is no way we can notify the team that object upload has never taken place. We could definitely use this attribute in our production.

vito:
@akamalov There is no ETA - see https://github.com/concourse/concourse/issues/1460#issuecomment-333652366. The current plan is to improve support for notifications so that you don't need to attach all these hooks everywhere in the first place. We did some planning in concourse/rfcs#1 but later cut scope, so we'll probably tackle support for notifications as a new interface (and RFC) later on.

marco-m:
@vito, to be sure you considered this, I would like to restate what @shinmyung0 mentioned, this goes beyond notification and is more serious: if I use the Concourse pool resource to take a lock in a previous job and a subsequent job errors out, there is no way to release the lock.

vito:
[...]

PR #3591, opened 2019-03-26
Initial support for on_error

amanw:
This resolves: https://github.com/concourse/concourse/issues/1460

The UI aspects look correct but was unsure of specific changes to WATS that may be necessary.

Signed-off-by: Aman Wangde <Aman.wangde@emc.com>
Co-Authored by: James King <james.king@emc.com>


atc/exec/on_error.go
```diff
@@ -0,0 +1,47 @@
...
+func (o OnErrorStep) Run(ctx context.Context, state RunState) error {
+	stepRunErr := o.step.Run(ctx, state)
+	// with no error, we just want to return right away
+	if stepRunErr == nil {
+		return nil
+	}
+
+	// for all errors that aren't caused by an Abort, run the hook
+	if stepRunErr != context.Canceled {
+		o.hook.Run(context.Background(), state)
```
vito: Might want an explicit `_ =` here to make it clear that we're intentionally dropping the error.

Do you think it'd be interesting to return both errors if the error hook errors? That could be done with `multierror`. I have no real preference.

atc/scheduler/factory/build_factory.go
```diff
@@ -305,6 +305,11 @@ func (factory *buildFactory) applyHooks(cp constructionParams) (atc.Plan, error)
 		return atc.Plan{}, err
 	}
 
+	cp, err = factory.errorIfPresent(cp)
```
vito: FWIW, I think this is where precedence becomes important. If you want the `on_error` hook to only act in response to the original step, and none of its hooks, I think putting it first here makes sense. Right now it's after the `on_abort` hook but it looks like that's fine because the `on_abort` hook [drops any errors from its hook on the floor](https://github.com/concourse/concourse/blob/master/atc/exec/on_abort.go#L36).

On the other hand, if you want the hook to fire if anything errors, including success/failure hooks on the same step, you'd want to put it last. But I think having it respond only to the original step *probably* makes the most sense - if you have an `on_failure` hook for cleanup and an `on_error` hook for cleanup, it probably wouldn't help to have one cause the other.

atc/validate_test.go
```diff
@@ -1005,15 +1016,15 @@ var _ = Describe("ValidateConfig", func() {
 
 				It("throws a validation error", func() {
 					Expect(errorMessages).To(HaveLen(1))
-					Expect(errorMessages[0]).To(ContainSubstring("jobs.some-other-job.plan[0].get.some-resource.ensure.put.custom-name refers to a resource that does not exist ('some-missing-resource')"))
+					Expect(errorMessages[0]).To(ContainSubstring("jobs.some-other-job.plan[0].get.some-resource.error.put.custom-name refers to a resource that does not exist ('some-missing-resource')"))
 				})
 			})
 
-			Context("when a plan has an invalid step within an abort", func() {
```
vito: Looks like the test for `abort` went away? Or is this just a confusing diff?

web/elm/src/Build/StepTree/StepTree.elm
```diff
@@ -421,6 +427,9 @@ viewTree model tree =
         OnAbort { step, hook } ->
             viewHooked "abort" model step hook
 
+        OnError { step, hook } ->
+            viewHooked "error" model step hook
```
vito: This needs a corresponding `.hook-error` style here:

https://github.com/concourse/concourse/blob/444bd916ed46febe5b39809a94d8ecdb90730ca0/web/assets/css/modules.less#L58-L76

It should set the border color to `@base09` (orange).

atc/exec/on_error.go
amanw: Good point on the multierror. Will add that in.

atc/scheduler/factory/build_factory.go
amanw: In the case of `on_abort` dropping it errors, do we want to move this up higher in case that error ever gets handled? It is working for now but could become problematic at a later date.

atc/validate_test.go
amanw: Confusing diff. Abort test is here: https://github.com/amanw/concourse/blob/master/atc/validate_test.go#L974

web/elm/src/Build/StepTree/StepTree.elm
amanw: Missed, will fix

amanw:
@vito changes added per your feedback

atc/validate_test.go
king-jam: Link to right branch: https://github.com/amanw/concourse/blob/on_error/atc/validate_test.go

atc/scheduler/factory/build_factory.go
vito: Hmm maybe it's fine as it is. Either way there's kind of a danger of one breaking the other, since they both check the `err` return value right now, so we'll just need to be careful and trust our tests.

vito:
@amanw Woops, I just merged another PR that bumped us to Elm 0.19 so that caused conflicts here. Would you mind resolving them? :grimacing:

amanw:
@vito Had some git branch issues. Closing this PR and opened a new one from my fork's master. https://github.com/concourse/concourse/pull/3652

Issue #3538, opened 2019-03-18
Replace `--tsa-peer-address` with just `--peer-address`


vito:
See https://github.com/concourse/concourse/pull/3528#issuecomment-473916941

PR #3620, opened 2019-03-28
replace --tsa-peer-address with --peer-address

xtremerui:
concourse/concourse#3538

With moving this flag from tsa cmd to atc cmd, it will lose a way to specify a peer IP when starts `tsa` by its own. Is there any concern about that? Or is there a better way to handle this.

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -79,8 +79,9 @@ type ATCCommand struct {
 type RunCommand struct {
 	Logger flag.Lager
 
-	BindIP   flag.IP `long:"bind-ip"   default:"0.0.0.0" description:"IP address on which to listen for web traffic."`
-	BindPort uint16  `long:"bind-port" default:"8080"    description:"Port on which to listen for HTTP traffic."`
+	BindIP      flag.IP `long:"bind-ip"   default:"0.0.0.0" description:"IP address on which to listen for web traffic."`
+	BindPort    uint16  `long:"bind-port" default:"8080"    description:"Port on which to listen for HTTP traffic."`
+	PeerAddress string  `long:"peer-address" default:"127.0.0.1" description:"Network address of this web node, reachable by other web nodes. Used for forwarded worker addresses."`
 
 	TLSBindPort uint16    `long:"tls-bind-port" description:"Port on which to listen for HTTPS traffic."`
 	TLSCert     flag.File `long:"tls-cert"      description:"File containing an SSL certificate."`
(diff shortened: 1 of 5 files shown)


pivotal-jwinters:
I think you bring up a good point @xtremerui. Shouldn't we keep this as part of the TSA cmd since the configuration is actually on the TSA?

vito:
Yeah, I don't think we should move the flag to the ATC since it's not the component that needs it.

We should probably figure out what we want to do with these embedded `Command` structs. It made sense back in the day when ATC and TSA were deployed separately and had their own CLIs. Embedding the `Command` struct was a way to support that + support fusing them together into one binary. But now we only distribute the single binary, and at this point the embedding is kind of getting in the way and makes the codebase and CLI feel like a few heavy components duct-taped together and [hacked against their will](https://github.com/concourse/concourse/blob/0b10714d106c5b1b6f12e02268c967fe62ba4f5c/cmd/concourse/web.go#L24-L30).

A user isn't going to know what `--tsa-` is as long as we're documenting it as "the `web` node", and the repo structure is also [made confusing](https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md#developing-concourse) by having all these cutesy names at the toplevel.

So we should probably have a long-term plan for how we want to organize our components/commands/flags/etc. But that'll probably take a while, so in the short-term can we just toss a flag at the top of the `web` command and pass it down to the TSA? I don't want to require users to rename it again assuming we get rid of the `--tsa-` namespace in the future.

atc/atccmd/command.go
```diff
@@ -79,8 +79,9 @@ type ATCCommand struct {
 type RunCommand struct {
 	Logger flag.Lager
 
-	BindIP   flag.IP `long:"bind-ip"   default:"0.0.0.0" description:"IP address on which to listen for web traffic."`
-	BindPort uint16  `long:"bind-port" default:"8080"    description:"Port on which to listen for HTTP traffic."`
+	BindIP      flag.IP `long:"bind-ip"   default:"0.0.0.0" description:"IP address on which to listen for web traffic."`
+	BindPort    uint16  `long:"bind-port" default:"8080"    description:"Port on which to listen for HTTP traffic."`
+	PeerAddress string  `long:"peer-address" default:"127.0.0.1" description:"Network address of this web node, reachable by other web nodes. Used for forwarded worker addresses."`
```
vito: This can be moved to `WebCommand` since the ATC doesn't need it.

xtremerui: Done Thx!

Issue #2003, opened 2018-02-02
Missing documentation for Kubernetes and AWS ssm credential management


marco-m:
Hello,

documentation at https://concourse.ci/creds.html mentions only Vault and Credhub, while Concourse supports also AWS SSM (AWS Systems Manager Parameter Store) (#1676) and kubernetes.

In general, Concourse has more features than the ones documented, it is a pity because the majority of the users will never know :-)

willejs:
I am trying to use SSM at the moment, I am presuming that setting the ```CONCOURSE_AWS_SSM_REGION``` environment variable would enable it, but it seems like it doesent?

marco-m:
@vito ?

willejs:
I figured out, that you have to pass the flag to atc, and the environment variables `CONCOURSE_AWS_SSM_REGION` or `AWS_REGION` do not work currently. I havnt dug into it properly yet, i was making the assumption that every flag was available as an environment var prefied with `CONCOURSE_` which is wrong!

analytically:
I think the correct variable name is CONCOURSE_AWS_SSM_AWS_REGION

ralekseenkov:
@vito I've started reworking the code for credential managers to support caching and ran into this as well... docs for k8s credential manager completely missing from https://concourse-ci.org/creds.html. is there a docs person on Concourse team who could possibly work on adding the section?

eedwards-sk:
AWS SSM is documented https://concourse-ci.org/aws-ssm-credential-manager.html

However, it wasn't clear that you must provide the region to enable SSM + instance credentials:

>If no access key, secret key, or session token is provided, Concourse will attempt to use environment variables or the instance credentials assigned to the instance.

Maybe something like "to use instance credentials, provide only the `AWS_REGION`" or similar?

Does anyone know if it's actually `AWS_REGION` as per the docs, or `CONCOURSE_AWS_SSM_AWS_REGION` as per @analytically ?

vito:
I've at least corrected the `AWS_REGION` var to `CONCOURSE_AWS_SSM_REGION` but haven't had time to update the docs properly. I'd like to give them the same amount of TLC I gave to the Vault docs recently but just haven't had time. :slightly_frowning_face:

eedwards-sk:
bumping to keep open

vito:
FWIW k8s has docs now thanks to @cirocosta - https://concourse-ci.org/kubernetes-credential-manager.html

[...]

PR #3628, opened 2019-03-29
Cache secret lookups, fixes #2924

### LINKED ISSUE #2924: Limit API calls to secrets backend

Last month we had ~62 million requests to AWS secrets manager from Concourse. That is about 24 requests per second, every second in a 30 day period. We have 4x ATC's and 8x workers running roughly 400 containers (check and otherwise) at any point in the day, and ~500 secrets used by Concourse that are stored in secrets manager.

Without doing the math, the number of requests seem a bit excessive and represent a significant cost in AWS Secrets Manager, and added costs in Guard Duty for analysing all the secrets manager requests. For SSM I suspect it's very easy to get hit by the rate limit with this amount of requests.

The only option that come to mind is implementing a cache to avoid hitting the APIs directly with each request. Also, https://github.com/concourse/concourse/issues/2386 might already help reduce the number of API calls to the secret backend (but I have not read the implementation).

edit: Just read that global resources does not reduce the number of calls to the secret backend according to Vito https://github.com/concourse/concourse/issues/2386#issuecomment-426786169:
> Hmm I don't think this will actually help that case at all. Credential resolution is done prior to even determining whether to run the check or not, since we have to resolve the credentials to determine the resource config to check in the first place.

I'm interested, but find it hard to navigate the Concourse codebase and would need someone to point me in the right direction 🤗

ralekseenkov:
refactoring of secret managers + generic secret cache implementaiton

this will also serve as a good foundation for team-based credential managers RFC, as now you can introduce a composable secret manager (same way as caching and retrying is implemented), which can iterate over several underlying secret managers

```
* composable secret manager for team 1
           -> vault 1
           -> vault 2
           -> ssm
```                 

and you can add retries and caching at any layer. e.g. one global cache, or several caches per each credential manager, etc.

see commit description


topherbullock:
Hey @ralekseenkov 
Question about one of the commits in this PR: https://github.com/concourse/concourse/pull/3628/commits/c0cecccb6cffe6e75cfd530f798aeb2fe8d42dbf - was there a need to add the `gobin ...` directives vs installing `counterfeiter` to the `GOPATH/$bin`(https://github.com/maxbrunsfeld/counterfeiter#installing-counterfeiter-to-gopathbin), or is this change to ensure the version of the tool is tracked `go.mod`?

ralekseenkov:
@topherbullock my PR was created before Concourse team opted to install counterfeiter to $GOPATH/bin. I just rebased it on top of master and removed these changes

vito:
Taking a look at this now! There are merge conflicts but I've resolved them locally (it wasn't too bad). If all is good I'll just push the merge with the conflict resolutions. :+1: 

So far I've just kicked the tires at a surface level and confirmed that credentials are cached with a locally running Vault by inspecting the audit log to see when credentials are fetched.

The default cache duration is only one minute, which means credentials will still be re-fetched roughly every resource check (which also has a one minute interval). To me that makes sense, from the standpoint that it's a conservative default that will help when there are many builds running, while still preventing a credential rotated independent of its lease duration (i.e. manually) from impacting the system for too long, as the next resource check or two will pick it up. If users need to mitigate load they can always increase the `--secret-cache-duration`.

I haven't done any testing on lease durations yet, but its interaction with `--secret-cache-duration` makes sense to me (smaller duration wins).

Thanks for starting on all this, and thanks for your patience!

atc/atccmd/command.go
```diff
@@ -880,14 +880,20 @@ func (cmd *RunCommand) variablesFactory(logger lager.Logger) (creds.VariablesFac
...
 		}
 
 		break
 	}
-	return creds.NewRetryableVariablesFactory(variablesFactory, cmd.CredentialManagement.RetryConfig), nil
+
+	result := secretsFactory.NewSecrets()
+	result = creds.NewRetryableSecrets(result, cmd.CredentialManagement.RetryConfig)
+	if cmd.CredentialManagement.CacheConfig.Enabled {
+		result = creds.NewCachedSecrets(result, cmd.CredentialManagement.CacheConfig)
```
vito: :tada: composition!

atc/creds/cached_secrets.go
```diff
@@ -0,0 +1,73 @@
+package creds
+
+import (
+	"github.com/patrickmn/go-cache"
```
vito: nice that there was a package for this :+1:

atc/creds/cached_secrets.go
```diff
@@ -0,0 +1,73 @@
...
+	// otherwise, let's make a request to the underlying secret manager
+	value, expiration, found, err := cs.secrets.Get(secretPath)
+
+	// we don't want to cache errors, let the errors be retried the next time around
+	if err != nil {
+		return nil, nil, false, err
+	}
+
+	// here we want to cache secret value, expiration, and found flag too
+	// meaning that "secret not found" responses will be cached too!
```
vito: neat

atc/creds/secrets_factory.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+//go:generate counterfeiter . Secrets
+
+type Secrets interface {
+	// Every credential manager needs to be able to return (secret, secret_expiration_time, exists, error) based on the secret path
+	Get(string) (interface{}, *time.Time, bool, error)
+
+	// NewSecretLookupPaths returns an instance of lookup policy, which can transform pipeline ((var)) into one or more secret paths, based on team name and pipeline name
+	NewSecretLookupPaths(string, string) []SecretLookupPath
```
vito: :+1: nice idea - simplifies everyone's `Get` implementation

although, should this just be called `SecretLookupPaths`? it doesn't really feel like a constructor to me

atc/creds/credhub/credhub.go
```diff
@@ -87,14 +85,3 @@ func (c CredHubAtc) findCred(path string) (credentials.Credential, bool, error)
 
 	return cred, true, nil
 }
-
-func (c CredHubAtc) path(segments ...string) string {
-	return path.Join(append([]string{c.PathPrefix}, segments...)...)
-}
-
-func (c CredHubAtc) List() ([]template.VariableDefinition, error) {
```
vito: :+1: I like that folks don't have to implement this stub anymore

atc/creds/secretsmanager/secretsmanager.go
```diff
@@ -3,69 +3,60 @@ package secretsmanager
...
-			return nil, false, err
-		}
-		if found {
-			return value, true, nil
-		}
+// Get retrieves the value and expiration of an individual secret
+func (s *SecretsManager) Get(secretPath string) (interface{}, *time.Time, bool, error) {
+	value, expiration, found, err := s.getSecretById(secretPath)
+	if err != nil {
+		s.log.Error("unable to retrieve aws secret", err, lager.Data{
```
vito: this should be conventional formatting:

```suggestion
		s.log.Error("failed-to-fetch-aws-secret", err, lager.Data{
```

ralekseenkov:
Sounds good. I was trying to be careful, decoupling variable lookup logic (i.e. NewSecretLookupPaths) from secret retrieval. Hopefully nothing got broken as a result of this.

But there are cases like Kubernetes credential manager - it doesn't seem to have any tests and it's currently missing docs, see https://github.com/concourse/concourse/issues/2003. That makes it challenging to verify

ralekseenkov:
thanks :)

samgurtman:
I hate to do this, but would anyone have any guidance on back porting this to 5.1.0? I have a big of egg on my face after moving our credentials to SSM. I guess I can set the default resource check to much longer, but this seems like a way better solution.

ralekseenkov:
@samgurtman it might be easier to wait for 5.2.0 because maintaining your own forks is never fun. if you really want to backport, ping me on Discord and I can try to help

cirocosta:
Hey @ralekseenkov, I noticed that the kubernetes cred mgmt changed a bit, breaking the tests we have in place for such integration:

https://github.com/concourse/concourse/blob/37dd02b20741a9c90b362f3f4edeb918a674caa9/topgun/k8s/kubernetes_creds_mgmt_test.go#L73-L120

Seems related to the new idea of having the `:` separator here:

https://github.com/concourse/concourse/blob/7b846eeac9a5ca232d521f25b2c0c410e72e02de/atc/creds/kubernetes/kubernetes.go#L33-L35

Was there a specific reason for going this way?

Thanks!

cirocosta:
Oh, here's an example failure:

```
 running: kubectl create secret generic foo --namespace=topgun-k8s-cm-664743-1-main --from-literal=value=bar
secret/foo created
running: kubectl create secret generic caz --namespace=topgun-k8s-cm-664743-1-main --from-literal=baz=zaz
secret/caz created

fly -t concourse-topgun-k8s-1 set-pipeline -n -c ../pipelines/minimal-credential-management.yml -p pipeline
jobs:
  job unit has been added:
+ name: unit
+ plan:
+ - task: run
+   config:
+     platform: linux
+     image_resource:
+       type: registry-image
+       source:
+         repository: busybox
+     run:
+       path: /bin/sh
+       args:
+       - -c
+       - |
+         echo plain-secret=((foo))
+         echo nested-secret=((caz.baz))

...

failed to interpolate task config: 
  Finding variable 'foo': unable to split kubernetes secret path 
  into [namespace]:[secret]: topgun-k8s-cm-664743-1-/main/pipeline/foo
```

ralekseenkov:
@cirocosta it should be backwards compatible. I'll look into this today

PR #3639, opened 2019-03-29
Update ci pipeline k8s - Run smoke and topgun in parallel with the k8s-check-helm-params task

xtreme-sameer-vohra:
Currently, the K8s section of the pipeline is blocked on `check-params` failing.
While, this is a useful signal that needs to be addressed prior to shipping, it shouldn't block running smoke & topgun tests, as those could be worked on independently of params changes.

Hence, the PR switches the ordering of the jobs from;

`k8s-check-helm-params` => `k8s-smoke` => `k8s-topgun`

to 

`k8s-smoke` => `k8s-topgun` => `k8s-check-helm-params`

### DIFF
--- a/ci/pipelines/concourse.yml
+++ b/ci/pipelines/concourse.yml
@@ -306,33 +306,6 @@ jobs:
     image: unit-image
     file: concourse/ci/tasks/downgrade-test.yml
 
-- name: k8s-check-helm-params
-  public: true
-  max_in_flight: 1
-  plan:
-  - aggregate:
-    - get: concourse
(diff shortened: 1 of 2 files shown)


xtreme-sameer-vohra:
@cirocosta @YoussB 
Will re-submit so that the pipeline is;

`k8s-smoke` => `k8s-topgun`     =>| `shipit`
`k8s-check-helm-params`        => | `shipit`

where the tests & `check-params` are run in parallel and shipit expects all 3 to pass

PR #3645, opened 2019-03-30
Add pipeline-operator role

tlwr:
This is a _draft_ pull request for adding a new role `operator` role (which is not the best name). I didn't want to open an issue without having a go at implementing it myself.

I understand that in future versions of Concourse there is likely to be a more complete RBAC/permissions model, I still thought it worthwhile for discussion here.

At GDS we have an immediate need for this feature, and if nothing else this draft PR may serve as way of finding out if others have the same needs as us, or have solved this in other ways.

---

Context
===

At GDS we have teams which follow a "2 pairs of eyes" process before changes can go out to production, where Concourse is used for deploying the changes. Being able to change the configuration of the pipeline, or be able execute code "as" Concourse circumvents this process. We would like for people to be able to  "use" or "operate" a pipeline without being able to change its configuration.

Use/operate in this case means pause/unpause pipelines/resources/jobs, trigger/abort jobs, check resources, without changing the configuration of the pipelines, or being able to hijack/execute.

Example 1
---

(teams refer to teams of people rather than Concourse teams)

There is a large team of people (of about 30, divided into subteams). These people should all be able to operate the main deploy to production pipeline, but should not be able to change the configuration of the pipelines, or to hijack/execute. A much smaller set of people are allowed to change the configuration.

The eventual goal is any team or pipeline configuration change is done by Concourse, following the 2 pairs of eyes principle. At this point no individual would be allowed to make changes to teams/pipelines.

Presently everyone would be a team member, if this was accepted then everyone would be an operator.

Example 2
---

(teams refer to teams of people rather than Concourse teams)

There is a small team which provides infrastructure to other teams, they share a Concourse deployment. The team providing the infrastructure should have full control over the teams/pipelines, the team using the infrastructure should be able to operate the pipelines, but not change the configuration.

Presently the infrastructure team are members, and the other team are viewers, this causes friction where the one team has to contact the infrastructure team in order to pause/abort builds.

---

Pull request
===

This pull request adds a new role: `operator` which is allowed to perform:

- CreateBuild
- AbortBuild
- CreateJobBuild
- PauseJob
- UnpauseJob
- ClearTaskCache
- PauseResource
- UnpauseResource
- CheckResource
- CheckResourceWebHook
- CheckResourceType
- EnableResourceVersion
- DisableResourceVersion
- PinResourceVersion
- OrderPipelines
- PausePipeline
- UnpausePipeline
- ExposePipeline
- HidePipeline
- ReportWorkerContainers
- ReportWorkerVolumes

These were chosen somewhat arbitrarily. The need is to be able to Trigger/Abort jobs, Check resources, Pin/Pause/Unpause pipelines, resources, jobs. Expose/Hide and Order are probably superfluous.

I chose the name `operator` arbitrarily, and in hindsight it is confusing, however I am not sure what a better name would be.

---

I'm very motivated to get this merged, if it is something the Concourse team are willing to accept; if there is any further work required (e.g. documentation, additional tests), please let me know!


paroxp:
> Expose/Hide and Order are probably superfluous.

Agree. In fact Expose/Hide may be an unwanted action, especially when it's something that arguably could/should be done GitOps way 🤔 

> I chose the name `operator` arbitrarily, and in hindsight it is confusing, however I am not sure what a better name would be.

Did you consider `pilot`? :trollface: 

---

Looks good!

tlwr:
Despite your alleged intention of trolling I really like `pilot` as a name :)

I agree that expose/hide is not needed.

I'll wait for a bit more feedback, then update the PR with the rename and de-elevation of expose/hide.

xtremerui:
@pivotal-jwinters

pivotal-jwinters:
@tlwr This new role makes a lot of sense to me. We may have to take a second pass at the actions it gets mapped to though. For instance `CreatePipelineBuild` and `CreateBuild` are both used by `fly execute`, but in this PR one is mapped to `member` and the other is mapped to `operator`. Is that intentional?

I kind of like the names `owner` >  `editor` >  `member` > `viewer`, although this makes things more complicated because then we need to migrate anyone who is currently a `member` -> `editor`.

tlwr:
The role mapping was not intentional: I have checked and I think `CreateJobBuild` is the only one we need. (i.e. no hijack/execute, but able to trigger jobs). Although my knowledge of how the fly commands map to ATC API endpoints isn't great.

I was thinking about de-elevating member to read-only but for reasons you mentioned it would be quite disruptive. I'm a fan of `pilot` :D as a role that sits between `member` and `viewer`.

paroxp:
To move things forward, I've revisited the permissions for the new `pilot` (winkity wink) role, and reverted some of the changes:

    - atc.CreateBuild
    - atc.OrderPipelines
    - atc.PausePipeline
    - atc.UnpausePipeline
    - atc.ExposePipeline
    - atc.HidePipeline
    - atc.ReportWorkerContainers
    - atc.ReportWorkerVolumes

The above are now set back to `member` role as a minimum.

Let me know if these are cool with everyone, and @tlwr and I shall cleanup the commits before merging.

atc/integration/rbac_test.go
```diff
@@ -90,6 +90,37 @@ jobs:
 				})
 			})
 
+			Context("when the role is pilot", func() {
+				BeforeEach(func() {
+					team.Auth = atc.TeamAuth{
+						"pilot": map[string][]string{
+							"users":  []string{"local:o-user"},
```
tlwr:
```suggestion
							"users":  []string{"local:p-user"},
```

Following the integration test role user conventions

tlwr:
@paroxp thanks for this :)

---

I've added an extremely minor suggestion about integration test user naming.

---

I also have a question: currently `pilot` can pause/unpause resources and jobs but not pipelines.

I believe this is intentional for your use case because you are using the [gsp-concourse-pipeline-controller](https://github.com/alphagov/gsp-concourse-pipeline-controller) which has a gitops flow for pausing/exposing pipelines, and therefore non-admin/non-member users do not need these permissions.

For other use-cases I can see pausing/unpausing entire pipelines might be useful for `pilot`s, although resources/jobs _should_ be enough. Exposing/hiding should definitely be done by a controller-like thing though). (IMO should be a property in the pipeline config file)

---

Looks good to me - as above @paroxp and I will squash the commits when this gets some attention/validation from a Concourse core contributor.

---

```
           \  \     _ _
            \**\ ___\/ \
          X*#####*+^^\_\
            o/\  \
               \__\
```

atc/integration/rbac_test.go
paroxp: oof... nice spot. not a fan of that single letter representative :D

tlwr: :D

paroxp:
> currently `pilot` can pause/unpause resources and jobs but not pipelines.
> 
> I believe this is intentional for your use case because you are using the [gsp-concourse-pipeline-controller](https://github.com/alphagov/gsp-concourse-pipeline-controller) which has a gitops flow for pausing/exposing pipelines, and therefore non-admin/non-member users do not need these permissions.
> 
> For other use-cases I can see pausing/unpausing entire pipelines might be useful for `pilot`s, although resources/jobs _should_ be enough.

You're right. I've overlooked the job management. Coming to think of it, this also could be done with the piece of YAML instead of the clicky interface.

One scenario I can think of, is having a `destroy-the-world` pipeline being on pause and having been accidentally triggered without going through a nice process of approval and tests. But then, we should perhaps trust the user to stay away from the danger or not have the danger? :D

I'm easy either way. Changed.

tlwr:
I've cleaned up the commits to be just 2: the tests and the implementation.

The role is now `pilot` with the following permissions:

- AbortBuild
- CreateJobBuild
- PauseJob
- UnpauseJob
- ClearTaskCache
- PauseResource
- UnpauseResource
- CheckResource
- CheckResourceWebHook
- CheckResourceType
- EnableResourceVersion
- DisableResourceVersion
- PinResourceVersion
- PausePipeline
- UnpausePipeline

pivotal-jwinters:
ok what about `attendant`? 

My problem with `pilot` is that it doesn't actually describe the role, whereas `attendant` both describes the role and relates back to aviation.

vito:
I'm gonna be the boring one and recommend we don't use cutesy names for things like this if there's a boring name that's even slightly more clear. :sweat_smile: I'm just afraid of using a name that could be mis-interpreted for something as important as access control. I haven't looked deeply enough at this to judge whether that's the case though.

paroxp:
I'll take anything. Tell me what fits you best and I'll make the change :D Just want it merged _cries inside_.

tlwr:
Echoing @paroxp's comment, I'm absolutely ambivalent about the name, but having this role would be _super useful_ for us.

@vito does raise a good point that _too interesting_ names can be confusing in a security context.

What makes naming is difficult is that the roles are defined in relation to the team, rather than what permissions they have more generally. I.e. `owner` refers to team owner, `member` to team member. This kind of breaks down with `viewer`. As above, none of the role names have aviation themed names either.

A very descriptive, but overly verbose option is `restricted-member` which describes what the user is, what permissions they have relative to the team, and provides context relation to other roles, with the downside of being two words.

paroxp:
Howdy!

Been 20 days since last comment. I was wondering if there is anything we could do to move this along.

vito:
@paroxp Just got back from vacation and getting back up to speed. I'm going through PRs now and will hopefully get to this soon. :+1:

vito:
I think at this point my only objection is the name, tbh. I think part of the problem may be that 'member' is somewhat nondescript, but it may be a bit late to rename that. :thinking:

tlwr:
Yeah it is probably too late to change member, “restricted-member” is consistent with the other roles.

The other name we were thinking of is “clicker” although mainly as a joke.

Let me know what you think the name should be and I’ll update the PR; my vote is restricted-member.

It probably doesn’t matter about the name in the long term if proper RBAC comes along :D (we can help here)

atc/api/accessor/accessor.go
```diff
@@ -125,37 +127,37 @@ var requiredRoles = map[string]string{
...
 	atc.ListAllResources:              "viewer",
 	atc.ListResources:                 "viewer",
 	atc.ListResourceTypes:             "viewer",
 	atc.GetResource:                   "viewer",
-	atc.PauseResource:                 "member",
-	atc.UnpauseResource:               "member",
+	atc.PauseResource:                 "pipeline-operator",
+	atc.UnpauseResource:               "pipeline-operator",
 	atc.UnpinResource:                 "member",
 	atc.SetPinCommentOnResource:       "member",
```
vito: These both (`atc.UnpinResource`, `atc.SetPinCommentOnResource`) should probably be `pipeline-operator` too, since `atc.PinResourceVersion` lets them pin it.

vito:
@pivotal-jamie-klassen It looks like we need to make some changes to the web UI for this - the web UI currently inspects the token and checks the user's roles to determine whether buttons should be enabled, and it needs to be made aware of this new role otherwise `pipeline-operator`s won't be able to click things like the pipeline play/pause button:

https://github.com/concourse/concourse/blob/7b846eeac9a5ca232d521f25b2c0c410e72e02de/web/elm/src/UserState.elm#L23-L36

Do we just need to add an `isPipelineOperator` and use that in places where you don't have to be a member anymore?

Here are some directions we can go from here:

1. @tlwr makes those changes prior to merging this PR. They're in Elm, so the development and testing approach will be different from what you've learned so far, but hopefully covered enough by the [contributing docs](https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md).
2. We merge this first and make the necessary changes to `web` ourselves by prioritizing this in the UX backlog. This would leave `master` in a partially finished state if we *do* end up shipping before the web UI is done, but users could always just run `fly` commands for operations that are left disabled in the web UI.

Any preference, @tlwr?

Thanks!

tlwr:
I will take a look at this presently

tlwr:
Hi @vito , @paroxp and I are not Elm experts, we've added the minimum amount of code we think we need - please let us know if we have to make more amendments.

The code will support web UI buttons for pipeline-operator, which will all work (as far as we can tell), with the exception of Expose/Hide pipelines, which will be clickable in the UI but silently fail.

(sorry 🙇)

tlwr:
@pivotal-jamie-klassen thanks :)

vito:
Thanks a bunch!

PR #3646, opened 2019-04-01
use elm/url to determine url when display URL metadata

Twiknight:
Signed-off-by: Twiknight <Twiknight@outlook.com>

This is a follow up of my previous patch on URLs:  https://github.com/concourse/concourse/pull/3398

Since we are now elm:0.19, naive `startsWith` should be replaced a real url parser.

After the change, URL-like non-URLs will not be displayed as hyperlink.


vito:
Thanks! :+1:

PR #3648, opened 2019-04-01
ci: remove task caches on elm builds

pivotal-jamie-klassen:
hopefully this will remove the flakiness, and the builds seem to be super fast anyway.


vito:
related: https://github.com/elm/compiler/issues/1845

we could perhaps work around this by setting `$HOME` to a cache so Elm puts things there instead of `~/.elm` (which goes away between builds). but let's see how things perform with no caching first, since Elm compiles faster and that'd be fewer moving parts.

PR #3664, opened 2019-04-02
web: handle timezones

### LINKED ISSUE #2226: Setting the Time Zone
# Feature Request

Times throughout concourse (in tooltips/logs/etc) are not labeled with a time zone, and seem to be some mixture of local times of the client/server.

We'd really just like to be able to set it all to UTC so that there is no ambiguity when working on global solutions.

pivotal-jamie-klassen:
fixes #2226 

I've chosen to approach this PR as essentially a bugfix, as I took a shortcut and always showed UTC when upgrading to elm 0.19.

The task that requests the current time zone from the browser can fail. In this case we default to showing UTC. I think it may be worthwhile in general to surface the timezone in various places where we show timestamps, or at least to give some explanation of the default choice of UTC in the failure case. However I'd like to keep that out of the scope of this PR.


web/assets/css/build.less
```diff
@@ -57,14 +57,10 @@
     padding-right: 20px;
     flex-shrink: 0;
     user-select: none;
+    color: @grey-secondary;
+    display: inline-block;
 
-    &::before {
```
vito: :thinking: I can't remember why this was done this way before, but it seems to be fine either way. Looks like this makes it possible to test, so :+1:

PR #3407, opened 2019-02-28
topgun/k8s: ensure garden configuration can be properly set for the worker pods

cirocosta:
This ensures that our 5.x branch can have garden properties passed not only through environment variables as usual, but also ConfigMaps, which can be referenced through `CONCOURSE_GARDEN_CONFIG`.

fixes #3237
fixes #3405

PR #3669, opened 2019-04-02
cmd/smoke: fix garden -config flag, add coverage in bin-smoke

vito:
This fixes `--garden-config`/`$CONCOURSE_GARDEN_CONFIG` - previously it would fail because `gdn` accepts `-config`, not `--config`.

Added test coverage to `bin-smoke` that covers usage of this flag and tests network connectivity.

### DIFF
--- a/ci/deployments/smoke/garden/garden.ini
+++ b/ci/deployments/smoke/garden/garden.ini
@@ -1,0 +1,3 @@
+[server]
+dns-server = 8.8.8.8
+dns-server = 8.8.4.4
--- a/ci/deployments/smoke/smoke.tf
+++ b/ci/deployments/smoke/smoke.tf
@@ -164,6 +164,11 @@ resource "null_resource" "rerun" {
     source = "systemd/smoke-worker.conf"
   }
 
+  provisioner "file" {
+    destination = "/etc/concourse/garden.ini"
+    source = "garden/garden.ini"
+  }
+
   provisioner "remote-exec" {
     inline = [
       "set -e -x",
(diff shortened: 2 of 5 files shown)


vito:
@cirocosta Does this work? :thinking: https://github.com/concourse/charts/blob/920ab2e239e75a071a4d46b442aeb29e4852781b/stable/concourse/values.yaml#L1022

vito:
Wait...I'm confused, it looks like it was originally `-config` but now it's `--config` because you were seeing the same error I saw.

https://github.com/concourse/concourse/pull/3407

WHAT IS GOING ON

vito:
Maybe it's because `CONCOURSE_GARDEN_CONFIG` is being interpreted twice? Once for [this flag](https://github.com/concourse/concourse/blob/8bb85e9b0fa9abddc375f734349ce8a3439573db/cmd/concourse/worker_linux.go#L32) and then it ends up being passed as a flag here?:

https://github.com/concourse/concourse/blob/8bb85e9b0fa9abddc375f734349ce8a3439573db/cmd/concourse/worker_linux.go#L244-L283

I guess this would go for all of the `CONCOURSE_GARDEN_*` flags. Does that mean `CONCOURSE_GARDEN_USE_HOUDINI` is broken, since that'll end up passing `--use-houdini` to `gdn server`, and that's an unknown flag? :thinking:

vito:
Figured it out. https://github.com/concourse/concourse/pull/3407 was merged after 5.0.0/5.0.1, which is what I was testing with. Just confirmed that *it* actually fixes the error I was seeing. I'll update this to remove the `-config` change then.

PR #3671, opened 2019-04-02
Hasten the state

### LINKED ISSUE #3508: Hasten state changes using LISTEN/NOTIFY

pivotal-jwinters:
- send "build_started" notifications to build tracker
- send "resource_scan_ID" notifications to radar
- radar scanner now uses the resource ID instead of the resource Name... for reasons

fixes #3508


atc/db/notifications_bus.go
```diff
@@ -62,65 +66,35 @@ func (bus *notificationsBus) Listen(channel string) (chan bool, error) {
...
-					// already had notification queued up; no need to handle it twice
-				}
-			}
-		} else {
-			// alert all listeners of connection break so they can check for things
-			// they may have missed
-			for _, sinks := range bus.notifications {
-				for sink := range sinks {
-					select {
-					case sink <- false:
```
vito: :thinking: I don't see why this should be removed. This is to cover a possibility where ATC1 is tracking the build, loses connectivity, and a user aborts the build (possibly through another ATC whose connectivity is fine) and ATC1 misses the notification while it reconnects. If we ignore it and don't re-check the condition that ATC1 is looking for, the build won't be aborted.

So removing this means condition-notifiers can miss their conditions, and at least for build aborting that can be pretty important. It lowers resiliency. It's an edge case, but it'll happen eventually.

It's a little hairy but the comments should have made it pretty clear (I'm also kind of sad to see comments go away in these trickier places). Should we bring it back? Or is it actually not needed anymore because of something I haven't seen?

pivotal-jwinters: It's back now :)

Issue #3660, opened 2019-04-01
Generalising url option to fly commands


mockersf:
## What challenge are you facing?

When using `fly`, I have to specify the target and my pipeline / build / ... and remember the options.

## What would make this better?

`fly hijack` can take an `url` and not need anything else to know which target / pipeline / build to use. This option is only available for `hijack` command.
```
-u, --url=                    URL for the build, job, or check container to hijack
```
Generalising this option to other commands to get the target and other info needed would be helpful.
In my case, the two commands where it would help me the most is `get-pipeline` and `set-pipeline`

commands where `fly` could retrieve the target and pipeline from the URL:
- [ ] `checklist`
- [ ] `destroy-pipeline`
- [ ] `expose-pipeline`
- [ ] `get-pipeline`
- [ ] `hide-pipeline`
- [ ] `jobs`
- [ ] `pause-pipeline`
- [ ] `rename-pipeline`
- [ ] `resources`
- [ ] `set-pipeline`
- [ ] `unpause-pipeline`

commands where `fly` could retrieve the target and job from the URL:
- [ ] `pause-job`
- [ ] `trigger-job`
- [ ] `unpause-job`

commands where `fly` could retrieve the target and extra info from the URL:
- [ ] `abort-build` (build or pipeline / job)
- [ ] `check-resource` (pipeline and resource)
- [ ] `check-resource-type` (pipeline and resource type)
- [ ] `clear-task-cache` (pipeline and job, would still need step)
- [ ] `resource-versions` (pipeline and resource)
- [ ] `watch` (build or pipeline / job)

## Are you interested in implementing this yourself?

Yes ! I can start with the commands on pipelines

vito:
Sounds nice. It'd be great if there was some way to do this generically so all the commands don't require too much code to handle it. :thinking:

taylorsilva:
# Issue Bankruptcy

This issue is being closed as part of 2025's **Issue Bankruptcy** declaration. [See here for details](https://blog.concourse-ci.org/posts/2025-01-06-concourse-2024-review/)

If you think this issue still has merit then feel free to open a new issue and reference this one.

---

This still makes sense to do FYI

PR #3683, opened 2019-04-04
Generalising url option to fly commands

mockersf:
for #3660 

This is the smallest change I could think of use an url in all command generically.
What is happening:
- a light options parser, `UrlSetup`, is declared parsing just the url. If the url is present, it will be split in it's components and they will be added as environment variables if they are not already present
- the method to get the target either from the target option or the url has been moved from `HijackCommand` to `Fly`
- the `url` parameter has been moved from `HijackCommand` to `Fly`
- now a command only has to specify that it can take a parameter by an environment variable and it will be loaded from the url if the url has the information. 
- to get the target from the url instead of just the parameter, it's just a different method call

load order is url overloaded by environment variable overloaded by dedicated flag.

See [here](https://github.com/concourse/concourse/pull/3683/files#diff-aab25ef5684fe04b55e132c3bff60622) for all the changes to do to `get-pipeline` for it to work with an url

The only side effect is that pipeline can now also be specified by an environment variable, which I think is a good thing.

`get-pipeline` now can be called (pipeline is `test`, target is `test`, team is `main`):
```
fly -t test get-pipeline -p test
fly get-pipeline -u http://concourse_web_1:8080/teams/main/pipelines/test
fly get-pipeline -u http://concourse_web_1:8080/teams/main/ -p test
PIPELINE=test fly get-pipeline -u http://concourse_web_1:8080/teams/main/
PIPELINE=test fly -t test get-pipeline
```
and if none is specified, it will still give the correct error message:
```
fly -t test get-pipeline
error: the required flag `-p, --pipeline' was not specified
```

### DIFF
--- a/fly/commands/abort_build.go
+++ b/fly/commands/abort_build.go
@@ -9,8 +9,8 @@ import (
 )
 
 type AbortBuildCommand struct {
-	Job   flaghelpers.JobFlag `short:"j" long:"job" value-name:"PIPELINE/JOB"   description:"Name of a job to cancel"`
-	Build string              `short:"b" long:"build" required:"true" description:"If job is specified: build number to cancel. If job not specified: build id"`
+	Job   flaghelpers.JobFlag `short:"j" long:"job" value-name:"PIPELINE/JOB" env:"JOB"   description:"Name of a job to cancel"`
+	Build string              `short:"b" long:"build" required:"true"         env:"BUILD" description:"If job is specified: build number to cancel. If job not specified: build id"`
 }
 
 func (command *AbortBuildCommand) Execute([]string) error {
--- a/fly/commands/abort_build.go
+++ b/fly/commands/abort_build.go
@@ -19,11 +19,6 @@ func (command *AbortBuildCommand) Execute([]string) error {
 		return err
 	}
 
-	err = target.Validate()
-	if err != nil {
-		return err
-	}
-
 	var build atc.Build
 	var exists bool
 	if command.Job.PipelineName == "" && command.Job.JobName == "" {
--- a/fly/commands/abort_build.go
+++ b/fly/commands/abort_build.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/fly/commands/internal/flaghelpers"
-	"github.com/concourse/concourse/fly/rc"
 )
 
 type AbortBuildCommand struct {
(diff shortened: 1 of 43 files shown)


vito:
@mockersf Sorry for the delay on this, but it looks like there's a bit of commit cruft now and it's hard to tell where your change begins and ends. :sweat_smile: Would you mind rebasing? I can take a look shortly after. Thanks!

mockersf:
@vito it's rebased ! 👍

deniseyu:
hey @mockersf sorry it's taken us a while to get back to you  😞 

Thanks for adding so many unit tests for this! It looks pretty good to me, other than some whitespaces that were perhaps introduced by your editor. do you mind removing them and rebasing one more time?

xtreme-sameer-vohra:
Hey @mockersf,
We created #4272 which rebased off master again to get the changes proposed in this PR merged.
We will close this PR when the new one is accepted. 

Thanks

deniseyu:
copying over comment from #4272:

thanks @xtremerui for catching that! we thought env had previously been used, but it looks like it was actually added by #3683.

This actually ends up being a bigger UX change that should warrant some discussion. I think that adding environment variable setting for things like trigger-job could end up being surprising to a fly user.

I'm going to revert the env flags for now but leave in the URL-based targeting functionality for these commands.

Update: reverting the env support breaks this feature. I still am not a huge fan of using env var setting as a way to run fly commands. Running commands this way won't pollute the environment because it looks like it just happens for the child process, but they can be inherited from the shell session in potentially dangerous way. I just ran something like this:

export JOB=production/deploy
then ran fly trigger-job with no job parameter or URL passed in and it indeed kicked off a new build. If a user forgot to pass parameters then that job would be triggered, when they'd normally expect fly to error out and complain about missing parameters. (I know I do this sort of thing all the time just to see what parameters are required for a command!)

If possible we should stay consistent with fly's current design where all inputs are explicitly passed in (or sensibly defaulted in some cases, like if there is only one entity in a list). The previous implementation of hijack with URL support didn't read or set env vars, so maybe copying that model more closely would be a safer way to implement this feature. @vito @xtreme-sameer-vohra @xtremerui what do you think?

cc @mockersf

taylorsilva:
@mockersf do you have interest in trying to get this PR back in shape? We really dropped the ball on you with our extremely slow response time, sorry about that. Let me know how you want to proceed with this PR.

mockersf:
hey @taylorsilva !

I think you can close it, there has been a lot of changes since, it would probably be too painful to rebase now, and I'm not sure how to do it without going through the environment, I'm not proficient enough in Go to see another way to do this without a large refactoring of fly code (from my memory of it's code from 1 year ago...)
Actually, adding environment support to pass parameters was a nice feature for me, as at my work we have environment variables set depending on the current directory, so fly commands could have been less verbose then

Thank you!

taylorsilva:
Thanks for getting back to us! I'll close this PR then.

PR #3692, opened 2019-04-05
web: fix styles on groups bar

### LINKED ISSUE #3561: [ui] pipeline groups overlap if the browser windows isn't wide enough
## Bug Report

If a pipeline has many groups and the browser window is narrow enough to force wraparound, the second row of groups overlaps the first.

## Steps to Reproduce

1) Create a pipeline with many groups
2) Make the browser window narrow

## Expected Results

2 neat non-overlapping rows of groups

## Actual Results

rows overlap, reducing legibility

## Version Info

* Concourse version: 5.0.0
* Infrastructure/IaaS: GCP

pivotal-jamie-klassen:
fixes #3561

### DIFF
--- a/web/elm/src/Pipeline/Styles.elm
+++ b/web/elm/src/Pipeline/Styles.elm
@@ -2,7 +2,6 @@ module Pipeline.Styles exposing
     ( cliIcon
     , groupItem
     , groupsBar
-    , groupsList
     , pauseToggle
     , pinBadge
     , pinDropdownCursor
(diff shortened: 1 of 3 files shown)


web/elm/src/Pipeline/Styles.elm
```diff
@@ -23,16 +22,9 @@ groupsBar : List (Html.Attribute msg)
 groupsBar =
     [ style "background-color" Colors.groupsBarBackground
```
xtreme-james-cooper: Erm... why is this a block of styles and not a List of string string pairs? o.O

pivotal-jamie-klassen: [elm 0.19](https://discourse.elm-lang.org/t/style-from-v0-18-to-v0-19/1766)

PR #3705, opened 2019-04-09
topgun/k8s: external worker test to use separate deployments

YoussB:
- In order to test the worker only and web-only deployments, we updated
the external worker test to use separate deployments so we can be able
to make sure that the worker connects successfully to the ATC.

cc @taylorsilva

### DIFF
--- a/topgun/k8s/external_worker_test.go
+++ b/topgun/k8s/external_worker_test.go
@@ -10,7 +10,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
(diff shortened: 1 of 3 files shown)


cirocosta:
related: https://github.com/helm/charts/pull/12920

cirocosta:
Hey @YoussB , would you mind rebasing against `master`, please?

I'd like to try these changes with the rest of the suite before getting https://github.com/helm/charts/pull/12920 in.

Thanks!

YoussB:
you got it 😉

PR #3707, opened 2019-04-09
Refactor the engine

### LINKED ISSUE #1585: Aborting a started build prior to an ATC tracking it will result in an orphaned, uncompleted build, still running

pivotal-jwinters:
This PR collapses multiple engines into a single engine that deals with locking. It now contains a `StepBuilder` which is responsible for converting a `db.Build` (mostly an `atc.Plan`) into an `exec.Step`.

based off #3566 

fixes #1585


atc/db/build.go
```diff
@@ -172,15 +178,9 @@ func (b *build) ReapTime() time.Time          { return b.reapTime }
...
-
-func (b *build) IsRunning() bool {
-	switch b.status {
-	case BuildStatusPending, BuildStatusStarted:
-		return true
-	default:
-		return false
-	}
-}
+func (b *build) IsRunning() bool              { return !b.completed }
```
topherbullock: This change means we consider builds on which we've called `MarkAsAborted()` on as "Running", which is different from the existing behaviour.

topherbullock: An easy fix would be to set the completed bool in `MarkAsAborted()`, assuming its safe to consider an abort as a possible path for a build to be completed

atc/db/build.go
```diff
@@ -452,7 +453,7 @@ func (b *build) Delete() (bool, error) {
 // build was aborted before it was started.
 func (b *build) MarkAsAborted() error {
 	_, err := psql.Update("builds").
-		Set("status", string(BuildStatusAborted)).
+		Set("aborted", true).
```
topherbullock: If we're no longer setting a build's status to `BuildStatusAborted`, it may make sense to remove that type entirely and pull on that thread. There's a few api tests which mock out the build object and assume returning a status of `BuildStatusAborted`  is still what indicates the aborted-ness of a build.

atc/db/build.go
```diff
@@ -63,7 +65,7 @@ type Build interface {
 	TeamID() int
 	TeamName() string
 	Schema() string
-	PrivatePlan() string
+	PrivatePlan() atc.Plan
```
topherbullock: YAAAAASS

pivotal-jwinters:
@topherbullock this refactor is built off of #3566, which hopefully outlines the reasoning behind the changes we've made to aborting builds.

PR #3716, opened 2019-04-10
Ubuntu bosh release

pivotal-bin-ju:
We added a step in the job `bosh-bump` to replace `alpine` rootfs with `ubuntu` rootfs in the concourse linux-rc.
Link to a private note with more details is here: https://github.com/orgs/concourse/projects/33#card-19512105


cirocosta:
Hey @vito , updated 👍

PR #3742, opened 2019-04-16
Validate token if provided

### LINKED ISSUE #3558: Calls to APIs with invalid or expired tokens can return empty 200s rather than 401
For example, `/api/v1/pipelines`.

@vito explained that Concourse views these calls as *this is what you have access to given your status* and in that case not being logged in is not an error condition and suggested prefacing calls to resources like this with a call to `/sky/userinfo` to check the login status.

Unfortunately, we can't make calls to `/sky/userinfo` atomically with the follow on request so I  can never be sure if my token is expired when making a resource request or if the resource is truly empty.

I'd expect that a call to `/api/v1/pipelines` with an invalid/expired token returns 401.

pivotal-jwinters:
fixes #3558


pivotal-jwinters:
@pivotal-jamie-klassen this will likely need some dashboard work. We also might want to invalidate the cookie if a request is not authorized.

vito:
@pivotal-jwinters Is this blocked on dashboard work then? I think we set the cookie to expire at the same time that the token expires (give or take a second I guess), so maybe it would fail on one request or two and show the 'unexpected turbulence' but then be fine eventually and effectively show the same info it showed before?

pivotal-jwinters:
@vito yeah, I meant to make it a draft PR. Apparently you can't go from PR -> Draft after the fact.

pivotal-jamie-klassen:
Sorry, missed this discussion! dropping it into the UX backlog.

PR #3679, opened 2019-04-03
web: click eye icon to toggle pipeline visibility

pivotal-jamie-klassen:
fixes #2903

PR #3743, opened 2019-04-16
web: automatically add comment when pinning resource

### LINKED ISSUE #3650: Resource page: way to tell who pinned and when

The Concourse team, like many teams practicing continuous integration, sometimes has flaky builds. When a build fails due to flakiness, we have been using the pinning feature to re-trigger it with the same inputs. When somebody does this, it's usually done in a hurry and they often don't stick around to wait for the build to schedule and then un-pin. However, it's hard to be sure, and you we don't want to accidentally un-pin when we shouldn't.

It would be very helpful if the pin action automatically added a comment containing the pinning user's username and the time they pinned, so that teammates who wish to modify the same resource know who to ask about the pin. This feels fairly MVP and non-invasive.

Immediately upon pinning, I'd expect to see a screen like
Screen Shot 2019-04-01 at 9 54 53 AM
(showing ISO 8601 timestamp with user's local time zone)

👍

pivotal-jamie-klassen:
fixes #3650

branched off #3679, please merge it first.


pivotal-jamie-klassen:
@pivotal-jwinters this is the awkward 'two-phase' solution I mentioned. Let me
know whether you want to make the change to the resource pinning API to allow
setting both the comment and pin atomically -- probably doing the simplest thing
will require some changes to the UI code as well, and I wouldn't worry about 
backwards compatibility as I don't think we have any serious consumers of that 
endpoint yet.

pivotal-jwinters:
@pivotal-jamie-klassen yeah I think moving forward we should accept a body on the pin request with the contents of the comment (similar to the pin_comment request). If you start sending that body in the request on the web side it will just ignore it on the API side for now until we get a chance to update that endpoint.

vito:
@pivotal-jamie-klassen Looks like this needs a rebase. After that is it good-to-go?

PR #3753, opened 2019-04-18
ci/pipelines: renames major, minor and patch to publish

YoussB:
just a rename for ambiguity

### DIFF
--- a/ci/pipelines/resources/template.jsonnet
+++ b/ci/pipelines/resources/template.jsonnet
@@ -148,7 +148,7 @@ local create_release = {
 };
 
 local publish_job(bump) = {
-  name: bump,
+  name: "publish-" + bump,
   plan: [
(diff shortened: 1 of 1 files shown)


ci/pipelines/resources/template.jsonnet
```diff
@@ -148,7 +148,7 @@ local create_release = {
 };
 
 local publish_job(bump) = {
-  name: bump,
+  name: "Publish " + bump,
```
cirocosta: Hey, what about going with the same pattern we use for the main pipeline?

```suggestion
  name: "publish-" + bump,
```

https://github.com/concourse/concourse/blob/d64843ce1bbab2309213ff677e25aadb950fc33b/ci/pipelines/concourse.yml#L79-L81

PR #3766, opened 2019-04-22
fix quickstart to ignore --worker-tsa-worker-private-key flag

xtremerui:
set this flag to `required` false when running `concourse quickstart`

Signed-off-by: Rui Yang <ryang@pivotal.io>


cirocosta:
Hey @xtremerui ,

IIRC we used to perform some very basic testing against `quickstart` (maybe I'm wrong); do you think we should get back to doing that? It feels like it could prevent us from seeing some small issues like this. Wdyt?

Thanks!

xtremerui:
@cirocosta yeah it would be nice if there is a test to cover. Not sure if we had those tests before. But we could follow the suit here potentially https://github.com/concourse/concourse/tree/master/atc/integration

cirocosta:
Cool! To not pollute this PR with those thoughts I created a separate issue for it: https://github.com/concourse/concourse/issues/3774

PR #3767, opened 2019-04-22
Restrict SSH MAC algos to a smaller set

chenbh:
- go by default allows for some weak algorithms https://github.com/golang/crypto/blob/d864b10871cd4370fe574816b489c819c675ccc7/ssh/common.go#L69
- fixes https://github.com/concourse/concourse-bosh-release/issues/25

### DIFF
--- a/tsa/constants.go
+++ b/tsa/constants.go
@@ -1,0 +1,8 @@
+package tsa
+
+// CIS recommends a certain set of MAC algorithms to be used in SSH connections. This restricts the set from a more permissive set used by default by Go.
+// See https://infosec.mozilla.org/guidelines/openssh.html and https://www.cisecurity.org/cis-benchmarks/
+var AllowedMACs = []string{
+	"hmac-sha2-256-etm@openssh.com",
+	"hmac-sha2-256",
+}
--- a/tsa/client.go
+++ b/tsa/client.go
@@ -342,6 +342,10 @@ func (client *Client) dial(ctx context.Context, idleTimeout time.Duration) (*ssh
 	}
 
 	clientConfig := &ssh.ClientConfig{
+		Config: ssh.Config{
+			MACs: AllowedMACs,
+		},
+
 		User: "beacon", // doesn't matter
 
 		HostKeyCallback: client.checkHostKey,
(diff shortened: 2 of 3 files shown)


tsa/constants.go
```diff
@@ -0,0 +1,8 @@
+package tsa
+
+// CIS recommends a certain set of MAC algorithms to be used in SSH connections. This restricts the set from a more permissive set used by default by Go.
+// See https://infosec.mozilla.org/guidelines/openssh.html and https://www.cisecurity.org/cis-benchmarks/
+var AllowedMACs = []string{
+	"hmac-sha2-256-etm@openssh.com",
+	"hmac-sha2-256",
```
taylorsilva: Why only allow these two MAC's? What about the `512`'s?
From the link in the comment, this is mozilla's list:
```
hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,umac-128-etm@openssh.com,hmac-sha2-512,hmac-sha2-256,umac-128@openssh.com
```

chenbh: Go doesn't support it.......

chenbh: https://github.com/golang/crypto/blob/d864b10871cd4370fe574816b489c819c675ccc7/ssh/common.go#L69

topherbullock: @taylorsil

topherbullock:
@chenbh @mhuangpivotal @taylorsilva
Should we also configure a reduced set of 'Ciphers' on the config?

tsa/constants.go
topherbullock: Annoying that there are constants for cipher suites but not MACs 😐

chenbh: They're also `string` instead of tls's `uint16` arrays. But wait, SSH also has Ciphers (different from TLS cipher suites apparently) and those are a mix of hard coded strings and constants! https://github.com/golang/crypto/blob/d864b10871cd4370fe574816b489c819c675ccc7/ssh/common.go#L31

Oh did I mention that the Key Exchange Algorithms and Host Key Algorithms that are also part of SSH uses constants?

It's a big beautiful mess of things ¯\\_(ツ)_/¯

chenbh:
The way we've approached this problem is to restrict the MACs for now, eventually we should also restrict the Ciphers and centralize everything for TLS and SSH in one location in a separate PR

PR #3776, opened 2019-04-23
Update prod deployment

taylorsilva:
There are two issues in this PR. Each commit represents one issue. Let me know if this should be split into two PR's.

### Issue 1
https://github.com/pivotal-cf/concourse-support/issues/69
This is setting `volume_sweeper_max_in_flight` to 1. The default is 5 and the runtime team thinks this is currently the root cause of the issues we're seeing with prod. We have some logs that appear to support this theory.

### Issue 2
https://github.com/pivotal-cf/concourse-support/issues/70
We're moving secrets shared between the release pipelines into a shared path in vault. This commit is telling concourse where the shared path is. 
Secrets that aren't shared between release pipelines will remain in their pipeline specific paths.

### DIFF
--- a/ci/pipelines/concourse.yml
+++ b/ci/pipelines/concourse.yml
@@ -801,6 +801,7 @@ jobs:
       - cbd/cluster/operations/enable-global-resources.yml
       - cbd/cluster/operations/x-frame-options-unset.yml
       - cbd/cluster/operations/worker-volume-sweeper-max-in-flight.yml
+      - cbd/cluster/operations/vault-shared-path.yml
       - prod/prod/ops.yml
       vars_files:
       - cbd/versions.yml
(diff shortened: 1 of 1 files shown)


cirocosta:
Hey @taylorsilva , thanks for the PR!

Regarding `issue 1`, would you mind elaborating on whether there's a specific metric or some behavior that we should observe that is different from what is going on? Also, if that default value would introduce a regression to our customers, should we pay some more close attention to it change that to `1` like we're doing here?

cc @ddadlani  @kcmannem  (runtime)

Thanks!

ddadlani:
@cirocosta 
Changing the limit to 1 is an experiment, we do not know yet if it will change anything.
We can compare the flamegraphs once again and see if its still taking up a similar amount of cpu resources. If not, we should see the worker instability go away.

It shouldn't really be a regression, volumes were always being deleted one at a time prior to v5.1 with no direct problems. We parallelized it as an optimization for https://github.com/concourse/concourse/issues/2833

taylorsilva:
@cirocosta I have nothing to add to Divya's comment. Updated issue 1 in main comment with: 

> We have some logs that appear to support this theory.

which refers to the first log mentioned in https://github.com/pivotal-cf/concourse-support/issues/69

Issue #3782, opened 2019-04-24
Display colorized diff when setting team configuration


cirocosta:
## What challenge are you facing?

When performing a `fly set-team` in a team that has a bunch of configuration, it's not intuitive to see what are the differences between what is currently set and the desired state (that is being configured).

For those making use of the new `--config` flag that came out in the 5 series (see https://concourse-ci.org/managing-teams.html#setting-roles), there are more and more configurations to be made (if desired), so having a rich diff could facilitate the prevention of little errors.

## What would make this better?

Having a `diff` between both states computed and displayed when performing the action.

## Are you interested in implementing this yourself?

😁

PR #3801, opened 2019-04-28
Enable to print specified team config (fly get-team)

### LINKED ISSUE #3785: Add command `get-team`
[User Roles](https://concourse-ci.org/managing-teams.html#setting-roles) are added since 5 series.
That's a great feature, but it adds outputs to the `fly teams` command, making it harder to find information about a particular team.

For example, if I want to check the information of a specific team (A-team), I need to execute the following command.

```bash
$ fly -t <target> teams -d --json | jq '.[] | if .name == "A-team" then . else empty end'
```

Supporting `fly get-team` command.

I think at least the following needs to be implemented in order for this command to be implemented.
* `GET /api/v1/teams/:team_name` endpoint in the ATC.
* `func (client *client) GetTeam(teamName string)` in the go-concourse
* `GetTeamCommand` struct in the fly. 

Yes.

cappyzawa:
fix #3785 

I enabled to print **one** team config.
The team config is as follows.

```yaml                     
id: 2
name: mechanic
auth:
  owner:
    groups: []
    users:
    - local:mechanic
```

Admin users and users who belong to a team can view the team config.
And the viewer or higher authority can view it.

Before, the way to get one team config was below.
(concourseTeam is mechanic)
```
$ fly -t <target> teams -d --json | jq '.[] | if .name == "mechanic" then . else empty end'
{
  "id": 2,
  "name": "mechanic",
  "auth": {
    "owner": {
      "groups": [],
      "users": [
        "local:mechanic"
      ]
    }
  }
}
```

After these changes, the way is below.
```                                           
$ fly -t <target> get-team -n mechanic 
name/role         users                    groups
mechanic /member  local:mechanic   none
```

or

```
$ fly -t <target> get-team -n mechanic --json
{
  "id": 2,
  "name": "mechanic",
  "auth": {
    "owner": {
      "groups": [],
      "users": [
        "local:mechanic"
      ]
    }
  }
}
```

or

```
fly -t <target> curl api/v1/teams/mechanic
{"id":2,"name":"mechanic","auth":{"owner":{"groups":[],"users":["local:mechanic"]}}}
```

### edited
This command's output was changed based on https://github.com/concourse/concourse/pull/3801#discussion_r280564477

### DIFF
--- a/fly/commands/get_team.go
+++ b/fly/commands/get_team.go
@@ -1,24 +1,24 @@
 package commands
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
+	"sort"
+	"strings"
 
-	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/fly/commands/internal/displayhelpers"
 	"github.com/concourse/concourse/fly/rc"
-	yaml "gopkg.in/yaml.v2"
+	"github.com/concourse/concourse/fly/ui"
+	"github.com/fatih/color"
 )
 
 type GetTeamCommand struct {
 	Team string `short:"n" long:"team" required:"true" description:"Get configuration of this team"`
-	JSON bool   `short:"j" long:"json" description:"Print config as json instead of yaml"`
+	Json bool   `short:"j" long:"json" description:"Print config as json instead of yaml"`
 }
 
 func (command *GetTeamCommand) Execute(args []string) error {
-	asJSON := command.JSON
-	teamName := command.Team
-
 	target, err := rc.LoadTarget(Fly.Target, Fly.Verbose)
 	if err != nil {
 		return err
@@ -28,7 +28,7 @@ func (command *GetTeamCommand) Execute(args []string) error {
 		return err
 	}
 
-	config, found, err := target.Team().Team(teamName)
+	team, found, err := target.Team().Team(command.Team)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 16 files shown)


cappyzawa:
I think these changes will make the implementation of #3782 easier.

go-concourse/concourse/team.go
```diff
@@ -12,6 +12,7 @@ import (
 type Team interface {
 	Name() string
 
+	Config(teamName string) (atc.Team, bool, error)
```
vito: Seems like this should be called `Team` instead, since that's what we do for `.Pipeline/.Job`. :thinking: 

It's a *little* weird because it's already on a `Team` object, but it's at least consistent with the other methods, and kind of makes sense since you're trying to use team A as an access check when fetching team B.

fly/commands/get_team.go
```diff
@@ -0,0 +1,58 @@
...
+	return dumpTeam(config, asJSON)
+}
+
+func dumpTeam(config atc.Team, asJSON bool) error {
+	var payload []byte
+	var err error
+	if asJSON {
+		payload, err = json.Marshal(config)
+	} else {
+		payload, err = yaml.Marshal(config)
```
vito: I think it's a little weird that this shows the internal team ID since we don't currently expose that anywhere else. I don't feel too strongly about it though. This is the first `get-X` command that acts on an API resource as opposed to pipeline config so it just hasn't come up, I suppose.

go-concourse/concourse/team.go
cappyzawa: > It's a little weird because it's already on a Team object, but it's at least consistent with the other methods, and kind of makes sense since you're trying to use team A as an access check when fetching team B.

I think so.
I will change `Config` to `Team` :+1:

fly/commands/get_team.go
cappyzawa: Thanks for your feedback!

The output format of `get-team` command followed `teams`, not `get-pipeline`.

cappyzawa: Now, this command, `get-team`, become a filter to get a specified team from the output of `teams` as planned.

cappyzawa:
@vito I fixed based on the feedbacks.

Thanks!!

fly/commands/get_team.go
vito: I think it was OK before, it just felt weird to include the ID. 🙂 Happy to merge either way, this is something we can easily iterate on as the use case evolves.

cappyzawa: I referenced `fly teams` command.
This command's output is below. (with `--json`)

<details>
<summary>output</summary>
<pre>
<code>
fly -t lm teams --json      
[
  {
    "id": 3,
    "name": "mechanic",
    "auth": {
      "owner": {
        "groups": [],
        "users": [
          "local:mechanic"
        ]
      }
    }
  }
]
</code>
</pre>
</details>

If we should hide the ID, `get-team` and `teams` commands are fixed together.

> Happy to merge either way, this is something we can easily iterate on as the use case evolves.

If we can easily iterate on as the use case evolves, I want to complete changes in this PR. 😁

Is it OK?

vito: Showing the ID in `--json` is fine since that shows all the internal details - I only thought it was confusing in the default case (YAML). Not a big deal either way - I'll go ahead and merge. :)

Issue #2723, opened 2018-10-23
Make Concourse compatible with Credhub 2.1


jama22:
Investigate breaking changes from CredHub 2.1 and make Concourse compatible with CredHub 2.1

Relevant information from CredHub team:
Release Notes: https://github.com/pivotal-cf/credhub-release/releases/tag/2.0.0
API Documentation: http://credhub-api.cfapps.io/version/2.1/

- Read the docs
- Investigate and outline changes to Concourse required to get it working
- Update pipelines
- ???
- Run multi-tenant CredHub on Wings

---

[Update from @cirocosta]:

- [x] Get PR in - https://github.com/cloudfoundry-incubator/credhub-cli/pull/72
- [x] Test integration with both versions of Credhub server in `topgun`

jama22:
Here's additional context from an old conversation circa June 2018

As a part of bumping our major version, we will be making some backwards-incompatible API and manifest changes. The biggest ones are as follows:

- Authorization model changes: Identities will no longer be able to write to CredHub solely by virtue of authenticating. Instead, an explicit access control entry will need to be created, granting an identity read, write, delete, read_acl, and/or write_acl permissions to an identity, for a given namespace.
- `set` and `generate` changes: CredHub is simplifying the rules that determine when `set` or `generate` requests will do nothing or overwrite existing values.
- Minor manifest changes: The structure of our properties hash is changing slightly.
- Authorization on by default: The default value of `credhub.authorization.acls.enabled` is changing from false to true.
- For bosh manifests that deploy and configure CredHub, necessary changes will include:
- Restructuring the properties hash
- Adding explicit permissions for certain users to CredHub's properties hash (similarly to configuring UAA clients)
- CredHub will ensure that the migration from 1.9, 1.7 or 1.6 to 2.x will not result in data loss for users. Old permissions will be migrated such that the platform will continue to function throughout the upgrade.

scottietremendous:
Going to break down the first step into something more specific. We want to prioritize existing compatibility with our supported versions. We'd start with 4.2.3 and validate that this works. After that we'd probably want to check out the 5.x series. 3.13 will be out of scope for this at this moment.

cirocosta:
Hey, I just noticed that there's this other issue related to CredHub 2.x+: https://github.com/concourse/concourse/issues/2293

chenbh:
Findings so far:
[...]

PR #3804, opened 2019-04-29
Upgrade credhub-cli

mhuangpivotal:
Also update topgun credhub manifest to work for 2.3.0.

concourse/concourse#2723


vito:
I just un-pinned credhub-release in the pipeline, fyi. Guessing that's OK.

PR #3806, opened 2019-04-30
Added --external-garden-address flag to use non embedded garden

rkoster:
This PR addresses: https://github.com/concourse/concourse-bosh-release/issues/17

### DIFF
--- a/cmd/concourse/worker.go
+++ b/cmd/concourse/worker.go
@@ -50,6 +50,8 @@ type WorkerCommand struct {
 
 	Garden GardenBackend `group:"Garden Configuration" namespace:"garden"`
 
+	ExternalGardenAddress flag.URL `long:"external-garden-address" description:"garden server connection address, when specified will not use embedded garden"`
+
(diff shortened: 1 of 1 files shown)


cmd/concourse/worker.go
```diff
@@ -50,6 +50,8 @@ type WorkerCommand struct {
 
 	Garden GardenBackend `group:"Garden Configuration" namespace:"garden"`
 
+	ExternalGardenAddress flag.URL `long:"external-garden-address" description:"garden server connection address, when specified will not use embedded garden"`
```
vito: Can we rename this to `ExternalGardenURL`? In the other flags, 'Address' typically means IP or IP:Port.

```suggestion
	ExternalGardenURL flag.URL `long:"external-garden-url" description:"API endpoint of an externally managed Garden server to use instead of running one."`
```

vito:
Made the requested changes myself since we're on the verge of getting v5.2 out.

PR #3820, opened 2019-05-06
topgun/k8s: remove flakiness from worker-lifecycle test

### LINKED ISSUE #3795: topgun/k8s - Refactor worker lifecycle test
In this PR https://github.com/concourse/concourse/pull/3779 we fixed the flakiness of the worker lifecycle test by adding `sleep`'s. This is not an ideal solution and we should come up with a more robust test for the worker lifecycle.

cirocosta:
Hey,

By ensuring that we're able to rely less and less on time, this commit
addresses the concerns in https://github.com/concourse/concourse/issues/3795.

The idea here is to control the timing by explicitly writing to the file that the pipeline waits forever one.

Thus, the two cases come down to:

```
gracefully:
	set pipeline

	trigger job async

	wait for a task container to appear 
		(instead of waiting for `started`)

	start scale down to 0 replicas 
		(instead of killing the pod)

	see that the worker is retiring

	allow the task to finish
		(by writing to a file, instead of waiting for some time)

	see that build succeeded
	
	see that there are no more workers


ungracefully:
	set pipeline

	trigger job async

	wait for a task container to appear 
		(instead of waiting for `started`)

	start scale down to 0 replicas 
		(instead of killing the pod)

	see that there are no more workers

	see that the task was interrupted right away
```

Thanks!

---

closes #3795

### DIFF
--- a/topgun/k8s/worker_lifecycle_test.go
+++ b/topgun/k8s/worker_lifecycle_test.go
@@ -8,109 +8,129 @@ import (
 
 	. "github.com/concourse/concourse/topgun"
 	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 )
 
 var _ = Describe("Worker lifecycle", func() {
+
 	var (
-		proxySession        *gexec.Session
-		atcEndpoint         string
-		helmDeployTestFlags []string
+		proxySession *gexec.Session
+		atcEndpoint  string
+		gracePeriod  string
 	)
 
-	type Case struct {
-		TerminationGracePeriod    string
-		PipelineYamlFile          string
-		WorkerDisappearingTimeout time.Duration
-		RerunJob                  bool
-	}
+	JustBeforeEach(func() {
+		setReleaseNameAndNamespace("wl")
+
+		deployConcourseChart(releaseName,
+			`--set=worker.replicas=1`,
+			`--set=persistence.enabled=false`,
+			`--set=worker.terminationGracePeriodSeconds=`+gracePeriod,
+		)
+
+		waitAllPodsInNamespaceToBeReady(namespace)
+
+		By("Creating the web proxy")
+		proxySession, atcEndpoint = startPortForwarding(
+			namespace, "service/"+releaseName+"-web", "8080",
+		)
+
+		By("Logging in")
+		fly.Login("test", "test", atcEndpoint)
+
+		By("waiting for a running worker")
+		Eventually(func() []Worker {
+			return getRunningWorkers(fly.GetWorkers())
+		}, 2*time.Minute, 10*time.Second).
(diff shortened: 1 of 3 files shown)


cirocosta:
Regarding moving from table-driven to just regular `Context`-based testing, I did that mainly to make the boundaries a bit more clear:

```
------------------------------
• [SLOW TEST:77.813 seconds]
Worker lifecycle
/w/concourse/topgun/k8s/worker_lifecycle_test.go:14
  terminating the worker
  /w/concourse/topgun/k8s/worker_lifecycle_test.go:74
    gracefully
    /w/concourse/topgun/k8s/worker_lifecycle_test.go:81
      finishes tasks gracefully with termination
      /w/concourse/topgun/k8s/worker_lifecycle_test.go:102
------------------------------
• [SLOW TEST:92.074 seconds]
Worker lifecycle
/w/concourse/topgun/k8s/worker_lifecycle_test.go:14
  terminating the worker
  /w/concourse/topgun/k8s/worker_lifecycle_test.go:74
    ungracefully
    /w/concourse/topgun/k8s/worker_lifecycle_test.go:116
      interrupts the task execution
      /w/concourse/topgun/k8s/worker_lifecycle_test.go:121
```

Wdyt?

Thanks!

topgun/k8s/worker_lifecycle_test.go
```diff
@@ -8,109 +8,129 @@ import (
...
+				"statefulset", releaseName+"-worker", "--replicas=0",
+			)
+		})
+
+		Context("gracefully", func() {
+			BeforeEach(func() {
+				gracePeriod = "600"
+			})
+
+			JustBeforeEach(func() {
```
YoussB: Since this is part of the actual test, and there is only one test in this context, shouldn't this `JustBeforeEach` be part of the `It` 🤔 ?

topgun/k8s/worker_lifecycle_test.go
```diff
@@ -8,109 +8,129 @@ import (
...
+		}, 2*time.Minute, 10*time.Second).
+			Should(BeTrue())
+	})
+
+	AfterEach(func() {
+		cleanup(releaseName, namespace, proxySession)
+	})
+
+	Context("terminating the worker", func() {
+		JustBeforeEach(func() {
```
YoussB: do we need to have `JustBeforeEach` in the describe block and another in the Context block?
Since these 2 will run for all the tests, can we consolidate them?

cirocosta:
Thanks for the review, @YoussB !

I just finished updating the PR addressing the comments.

Wdyt?

Thanks!

YoussB:
@cirocosta not sure why I didn't get an email about your comment ... let me check.
on another note I was able to run the test multiple times without getting any failure, however, it looks like there is an interesting error with the `-untilItfails` flag that the generated random number for the `releaseName` and `namespace` is the same in every run.

PR #3823, opened 2019-05-06
Fix regression in k8s credential manager

### LINKED ISSUE #3821: Kubernetes credential manager regression - malformed secret path
## Bug Report

After the merge of #3628 , the Kubernetes credential manager tests (https://github.com/concourse/concourse/blob/37dd02b20741a9c90b362f3f4edeb918a674caa9/topgun/k8s/kubernetes_creds_mgmt_test.go) started failing:

```
running: kubectl create secret generic foo --namespace=topgun-k8s-cm-664743-1-main --from-literal=value=bar
secret/foo created
running: kubectl create secret generic caz --namespace=topgun-k8s-cm-664743-1-main --from-literal=baz=zaz
secret/caz created

fly -t concourse-topgun-k8s-1 set-pipeline -n -c ../pipelines/minimal-credential-management.yml -p pipeline
jobs:
  job unit has been added:
+ name: unit
+ plan:
+ - task: run
+   config:
+     platform: linux
+     image_resource:
+       type: registry-image
+       source:
+         repository: busybox
+     run:
+       path: /bin/sh
+       args:
+       - -c
+       - |
+         echo plain-secret=((foo))
+         echo nested-secret=((caz.baz))
```

There might be something in the caching logic that is not working for that credential manager.

## Steps to Reproduce

Run the credential manager test. 

## Expected Results

Continue working.

## Actual Results

Fails.

## Aditional context

I didn't go through the new caching logic, but it seems like the way that paths are passed to `Get` is not working correctly, at least when caching is not enabled. It might be that it works with caching, but I didn't test that yet. 

## Version Info

* Concourse version: master (https://github.com/concourse/concourse/commit/7b846eeac9a5ca232d521f25b2c0c410e72e02de)
* Deployment type (BOSH/Docker/binary): Kubernetes
* Did this used to work? yes

ralekseenkov:
Fixes #3821

Kubernetes secret manager was one of the two places I was worried about, because it didn't have any unit tests. I was moving methods between SecretsFactory and Secrets, and the right code ended up in a place where it was not being used :(

I moved it to the right place and added a basic unit test, which tests var -> secret name conversion.

### DIFF
--- a/atc/creds/kubernetes/kubernetes.go
+++ b/atc/creds/kubernetes/kubernetes.go
@@ -4,7 +4,6 @@ import (
 	"code.cloudfoundry.org/lager"
 	"fmt"
 	"github.com/concourse/concourse/atc/creds"
-	"path"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


cirocosta:
Hey @ralekseenkov,

Thanks so much for quickly getting the fix!

I just tested the branch against the k8s-related test suite, and it all works great :grin:

As a follow-up, I created #3825  to ensure that the variables are also properly set at the charts level and that we don't end up with possible future regressions on the caching functionality.

Thanks!

vito:
Thanks! 🙂

Issue #3811, opened 2019-05-01
Change Task Caches to be Worker Agnostic


kcmannem:
Following up on #3607, We need to remove the ownership of task_caches from runtime.

Currently we keep  track of which worker contains a certain task_cache. But if workers become a runtime concern, Core should keep track of which task_caches are available and should be invisible to runtime. Continuing our move to the "artifact" abstraction, we should move this responsibility over to core. This will mean that task_caches can no longer be worker specific. This will cause implementation changes where task_caches may be cached across multiple workers in the pool.

ddadlani:
Part 1 of this issue is to separate out the functionality of `worker_task_caches` into `task_caches` and `worker_task_caches`, similar to `resource_caches` and `worker_resource_caches`. Then we should be able to remove `worker_task_caches` as part of the work for #3607.

ddadlani:
@xtremerui You can merge into master, we rebase from that regularly and it shouldn't be too hard to do at this stage. 

Thank you!

PR #3830, opened 2019-05-08
atc: separate worker_task_cache table

xtremerui:
#3811

### DIFF
--- a/atc/db/migration/migrations/1557152441_separate_worker_task_cache.up.sql
+++ b/atc/db/migration/migrations/1557152441_separate_worker_task_cache.up.sql
@@ -1,0 +1,52 @@
+BEGIN;
+  CREATE TABLE task_caches (
+    id serial,
+    worker_name text,
+    job_id integer,
+    step_name text NOT NULL,
+    path text NOT NULL,
+    PRIMARY KEY ("id")
+  );
+  CREATE UNIQUE INDEX task_caches_job_id_step_name_path_uniq
+    ON task_caches (job_id, step_name, path);
+
+  CREATE INDEX task_caches_job_id ON task_caches USING btree (job_id);
+
+  ALTER TABLE ONLY task_caches
+    ADD CONSTRAINT task_caches_job_id_fkey FOREIGN KEY (job_id) REFERENCES jobs(id) ON DELETE CASCADE;
+
+  ALTER TABLE worker_task_caches
+    ADD COLUMN task_cache_id INTEGER;
+
+  ALTER TABLE worker_task_caches
+    ADD CONSTRAINT worker_task_caches_task_cache_fkey FOREIGN KEY (task_cache_id) REFERENCES task_caches (id) ON DELETE CASCADE;
+  CREATE INDEX worker_task_caches_task_cache_id ON worker_task_caches (task_cache_id);
+
+  WITH ins AS (
+    INSERT INTO task_caches (worker_name, job_id, step_name, path)
+    SELECT DISTINCT wtc.worker_name, wtc.job_id, wtc.step_name, wtc.path
+    FROM worker_task_caches wtc
+    RETURNING *
+  )
+  UPDATE worker_task_caches wtc
+  SET task_cache_id = ins.id
+  FROM ins
+  WHERE ins.worker_name = wtc.worker_name
+    AND ins.job_id = wtc.job_id
+    AND ins.step_name = wtc.step_name
+    AND ins.path = wtc.path;
+
+  DROP INDEX worker_task_caches_uniq;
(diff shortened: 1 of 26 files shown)


atc/db/task_cache_factory_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("TaskCacheFactory", func() {
+
+	Describe("FindOrCreate", func() {
+		Context("when there is no existing tash cache", func() {
+			It("creates resource cache in database", func() {
```
ddadlani: minor nitpick: task cache

atc/db/task_cache_factory_test.go
```diff
@@ -0,0 +1,89 @@
...
+			BeforeEach(func() {
+				usedTaskCache, err = taskCacheFactory.FindOrCreate(
+					defaultJob.ID(),
+					"some-step",
+					"some-path",
+				)
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			It("finds resource cache in database", func() {
```
ddadlani: same

atc/db/worker_task_cache.go
```diff
@@ -0,0 +1,113 @@
...
+		Where(sq.Eq{
+			"worker_name":   wtc.WorkerName,
+			"task_cache_id": wtc.TaskCache.ID(),
+		}).
+		RunWith(tx).
+		QueryRow().
+		Scan(&id)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			err = psql.Insert("task_caches").
```
ddadlani: why do we also insert a new task cache here? Would it not already exist since the `UsedTaskCache` exists?

xtremerui: When coming from volume.go `InitializeTaskCache`, the task cache is not in DB yet. So it has to create a new task cache first here. 

But I can see the reason that this part of the code doesn't belongs to worker task cache. Maybe, we should init the task cache in `InitializeTaskCache` first?

atc/db/migration/migrations/1557152441_separate_worker_task_cache.up.sql
```diff
@@ -0,0 +1,52 @@
...
+    ADD CONSTRAINT task_caches_job_id_fkey FOREIGN KEY (job_id) REFERENCES jobs(id) ON DELETE CASCADE;
+
+  ALTER TABLE worker_task_caches
+    ADD COLUMN task_cache_id INTEGER;
+
+  ALTER TABLE worker_task_caches
+    ADD CONSTRAINT worker_task_caches_task_cache_fkey FOREIGN KEY (task_cache_id) REFERENCES task_caches (id) ON DELETE CASCADE;
+  CREATE INDEX worker_task_caches_task_cache_id ON worker_task_caches (task_cache_id);
+
+  WITH ins AS (
```
clarafu: Not sure if you want to but maybe a migration test might be useful for this query that migrates data (along with the down migration one)

atc/db/worker_task_cache.go
```diff
@@ -0,0 +1,111 @@
...
+			Columns(
+				"worker_name",
+				"task_cache_id",
+			).
+			Values(wtc.WorkerName, wtc.TaskCache.ID()).
+			Suffix(`
+					ON CONFLICT (worker_name, task_cache_id) DO UPDATE SET
+						task_cache_id = ?
+					RETURNING id
+				`, id).
```
clarafu: 🤔This should probably be wtc.TaskCache.ID(), or you can also just use the EXCLUDED.task_cache_id instead of passing in an param so that it'll use the id value of the task_cache_id it conflicted with.

xtremerui: Good catch thx! There was a block before this to look for task cache by sql and was returning `id`. Now it is using the find() method instead.

atc/db/migration/migrations/1557152441_separate_worker_task_cache.up.sql
xtremerui: I have been thinking about this, IMHO the test here might not be very much needed:

1. The worker_task_caches table will be removed by runtime eventually 
2. The table is relatively simple 
3. The task cache data is not a key requirement for concourse being functioning properly. If in worst case all task cache data are lost during up/down migration, the task can still run (just that it will need to populate the task cache path again) 

What do you think @clarafu ? Thx!

clarafu: Makes sense to me, it's just my personal preference of usually writing tests for migrations that deal with moving data over but in this case I can see it being unnecessary. 👍

PR #3850, opened 2019-05-13
changing max volume sweep in flight to 3

ddadlani:


ddadlani:
3 has been more stable in prod/wings than 5, we don't know if the btrfs issues are caused by this or they're just correlated, but volume deletion is, in general, a heavier task than container creation so it's probably safest to delete fewer volumes in parallel.

Issue #1077, opened 2017-05-09
fly status command to provide high-level info about the ATC in question


chendrix:
It would be nice to get some high-level info about the ATC via a command like `fly status` - similar to `bosh status`

---
Moved from concourse/fly#20
/cc @robdimsdale

PR #3862, opened 2019-05-14
atc: info API yields the external URL and cluster name

agurney:
The `/api/v1/info` path, in addition to giving the server and worker versions, now also gives the "external URL" - the canonical entry point for the installation. This is useful in cases where the API is accessed by IP address, through a proxy, etc., and we want to find out the proper URL instead. It also now gives the user-facing cluster name, if one is configured, which might in the future be picked up by fly or other clients (see #1077).

This PR doesn't currently have any consumers of this information, but it seems probably useful for it to show up in `fly targets` listings, at least. I'm not sure if the right thing is to display the external URL alongside the configured target URL, or if instead we should be _replacing_ whatever the user specified with whatever the cluster reports in `/info` (e.g., I login with `-c http://10.20.30.40`, that cluster says its external URL is `https://ci.example.com`, and the latter is what gets saved in `.flyrc`). Likewise, I could imagine something like using the cluster name as a default target name (e.g. I could login without specifying `-t`, the cluster says its name is `dev`, so that's saved as target `dev`), but it's not obvious to me whether that's a good idea.

### DIFF
--- a/atc/api/api_suite_test.go
+++ b/atc/api/api_suite_test.go
@@ -32,6 +32,7 @@ var (
 	sink *lager.ReconfigurableSink
 
 	externalURL = "https://example.com"
+	clusterName = "Test Cluster"
 
 	fakeWorkerClient        *workerfakes.FakeClient
 	fakeVolumeRepository    *dbfakes.FakeVolumeRepository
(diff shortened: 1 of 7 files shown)


vito:
thanks!

pivotal-jamie-klassen:
Seeing this merged I want to not forget: the web UI should use this endpoint to fetch the cluster name now, instead of having the webhandler render it into the HTML template: #3909

PR #3864, opened 2019-05-14
cmd/concourse: fix worker private key flag

vito:
this flag should be required for the worker commands, and only optional
in 'quickstart'


vito:
thx!

Issue #3805, opened 2019-04-30
Increase highlighting of highlighted-line


tvon:
## What challenge are you facing?

When sharing a link to a specific line or a selection of lines the timestamps stand out but the related log lines blend in.

## What would make this better?

The UI background is `#3d3c3c`, changing the `.timestamped-line.highlighted-line` background to `#2d2c2c` (or some other color) is subtle but helps to make the relevant lines "pop" a bit more.

Before:
![Screen Shot 2019-04-30 at 12 36 20 AM](https://user-images.githubusercontent.com/25628/56941118-5010af00-6ae0-11e9-98e3-4099411ac50a.png)

After:
![Screen Shot 2019-04-30 at 12 36 27 AM](https://user-images.githubusercontent.com/25628/56941111-46874700-6ae0-11e9-891a-5e4d58e438e5.png)

## Are you interested in implementing this yourself?

I would take some familiarization with the UI code but I could figure it out if there is interest.  Or if someone with the domain knowledge wants to bang it out... whatever works.

vito:
@tvon I like it! I've marked this as 'good first issue' - [`CONTRIBUTING.md`](https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md) should be a good start if you're interested in PRing it. :+1:

cirocosta:
Thanks, @tvon ! https://github.com/concourse/concourse/pull/3866

PR #3866, opened 2019-05-14
Set a darker background on highlighted-line

tvon:
Darken `.highlighted-line` a bit, switching from `@base02` (`#3d3c3c`) to `@base01` (`#2a2929`) which helps highlight the highlighted-line(s) pop a bit more:

### Before
before

### After
after

Addresses https://github.com/concourse/concourse/issues/3805

### DIFF
--- a/web/assets/css/build.less
+++ b/web/assets/css/build.less
@@ -70,7 +70,7 @@
   }
 
   &.highlighted-line {
-    background: @base02;
+    background: @base01;
 
     .timestamp {
       color: @base0A;


vito:
@pivotal-jamie-klassen I suppose ideally this would be done in Elm using the new pattern for CSS, though for the moment it's a one-line change so I'm tempted to just pull it in. Any preference?

vito:
Sounds good - thanks @tvon!

PR #3867, opened 2019-05-15
topgun/k8s: reduce even more the flakiness of worker_lifecycle test

cirocosta:
We were previously not accounting for the fact that it might take a
little while for the build to be marked as succeeded.

That race used to make our tests fail quite a good amount of times.

### DIFF
--- a/topgun/k8s/worker_lifecycle_test.go
+++ b/topgun/k8s/worker_lifecycle_test.go
@@ -104,9 +104,12 @@ var _ = Describe("Worker lifecycle", func() {
 					Should(HaveLen(0))
 
 				By("seeing that the build succeeded")
-				startSession := fly.Start("builds", "-j", "some-pipeline/simple-job")
-				<-startSession.Exited
-				Expect(startSession.Out).To(gbytes.Say("succeeded"))
+				Eventually(func() *gbytes.Buffer {
+					startSession := fly.Start("builds", "-j", "some-pipeline/simple-job")
+					<-startSession.Exited
+					return startSession.Out
+				}, 1*time.Minute, 5*time.Second).
+					Should(gbytes.Say("succeeded"))
 			})
 		})
 
--- a/topgun/k8s/worker_lifecycle_test.go
+++ b/topgun/k8s/worker_lifecycle_test.go
@@ -104,12 +104,7 @@ var _ = Describe("Worker lifecycle", func() {
 					Should(HaveLen(0))
 
 				By("seeing that the build succeeded")
-				Eventually(func() *gbytes.Buffer {
-					startSession := fly.Start("builds", "-j", "some-pipeline/simple-job")
-					<-startSession.Exited
-					return startSession.Out
-				}, 1*time.Minute, 5*time.Second).
-					Should(gbytes.Say("succeeded"))
+				fly.Run("watch", "-j", "some-pipeline/simple-job")
 			})
 		})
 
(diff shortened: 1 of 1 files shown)


topgun/k8s/worker_lifecycle_test.go
```diff
@@ -104,9 +104,12 @@ var _ = Describe("Worker lifecycle", func() {
 					Should(HaveLen(0))
 
 				By("seeing that the build succeeded")
-				startSession := fly.Start("builds", "-j", "some-pipeline/simple-job")
-				<-startSession.Exited
-				Expect(startSession.Out).To(gbytes.Say("succeeded"))
```
vito: I think you could also run `fly watch -j some-pipeline/simple-job` and wait for it to exit.

cirocosta: oh, good point!

cirocosta:
I noticed that running the ungraceful stops, it seems like we don't receive the "disappeared" message anymore 🤔Or at least, in the multiple times I've run it, it's never saying that:

```
      Got stuck at:
          initializing
          fetching busybox@sha256:4fe8827f51a5e11bb83afa8227cbccb402df840d32c6b633b7ad079bc8144100
53071b97a884 [======================================] 738.1KiB/738.1KiB
          running sh -c until test -f /tmp/stop-waiting; do
            echo 'waiting for /tmp/stop-waiting to exist'
            sleep 1
          done

          echo done

          waiting for /tmp/stop-waiting to exist
          waiting for /tmp/stop-waiting to exist
          waiting for /tmp/stop-waiting to exist
          waiting for /tmp/stop-waiting to exist
          waiting for /tmp/stop-waiting to exist
          dial tcp 10.8.3.46:34707: connect: connection refused
          errored

      Waiting for:
          disappeared
```

Is this a behavior change in 5.2? 🤔  I don't remember seeing anything changing in this code path.

Thanks!

vito:
@cirocosta Not sure what the 'disappeared' message was, but if we're getting the above error earlier than usual perhaps there's some missing network retry logic? Might have to check the logs.

cirocosta:
@vito the current error makes sense, I was just surprised that we're not getting the `WorkerMissing` there: https://github.com/concourse/concourse/blob/45c135a1d4d0f653eaa61f9b69235b7965dc2f99/atc/worker/transport/errors.go#L5-L11

I was expecting to see that in more places 🤔  

```
atc/worker/transport/baggageclaim_round_tripper.go
32:			return nil, WorkerMissingError{WorkerName: c.workerName}

atc/worker/transport/baggageclaim_round_tripper_test.go
137:				Expect(err).To(Equal(transport.WorkerMissingError{WorkerName: "some-worker"}))

atc/worker/transport/garden_round_tripper.go
29:			return nil, WorkerMissingError{WorkerName: c.workerName}

atc/worker/transport/hijackable_client_test.go
94:			Expect(err).To(Equal(transport.WorkerMissingError{WorkerName: "some-worker"}))

atc/worker/transport/garden_round_tripper_test.go
135:				Expect(err).To(Equal(transport.WorkerMissingError{WorkerName: "some-worker"}))

atc/worker/transport/errors.go
5:type WorkerMissingError struct {
9:func (e WorkerMissingError) Error() string {

atc/worker/transport/hijackable_client.go
33:			return nil, nil, WorkerMissingError{WorkerName: c.workerName}


atc/worker/volume_client.go
98:	return fmt.Sprintf("volume '%s' disappeared from worker '%s'", e.Handle, e.WorkerName)
```

To summarize, I think checking for `errored` there is fine 👍 wdyt?

PR #3868, opened 2019-05-15
WIP: Fix #3855

### LINKED ISSUE #3855: It's difficult to set oidc users
related: https://github.com/concourse/concourse/pull/3402#issuecomment-487287372

OIDC users are set by the subject claim, but this value cannot be got without passing through the oidc client.
So, It's difficult to set the oidc users for operators.

If there is no problem even if `CONCOURSE_OIDC_USER_ID_KEY` is set, it is easy.

Yes.

cappyzawa:
fix #3855 

blocked by https://github.com/concourse/dex/pull/8 or https://github.com/dexidp/dex/pull/1448

### DIFF
--- a/skymarshal/skycmd/oidc_flags.go
+++ b/skymarshal/skycmd/oidc_flags.go
@@ -25,6 +25,7 @@ type OIDCFlags struct {
 	Scopes             []string    `long:"scope" description:"Any additional scopes that need to be requested during authorization"`
 	GroupsKey          string      `long:"groups-key" default:"groups" description:"The groups key indicates which claim to use to map external groups to Concourse teams."`
 	UserNameKey        string      `long:"user-name-key" default:"username" description:"The user name key indicates which claim to use to map an external user name to a Concourse user name."`
+	UserIDKey          string      `long:"user-id-key" description:"The user id key indicates which claim to use to map an external user id to a Concourse user id."`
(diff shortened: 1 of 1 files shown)


cappyzawa:
@vito https://github.com/dexidp/dex/pull/1448 is merged 🎉 

What should I do to merge this change into concourse/dex ?

cirocosta:
Hi @cappyzawa , thanks for the PR!

cc @pivotal-jwinters

pivotal-jwinters:
I'm actually a little hesitant to make this change (even though dex accepted the PR).

OIDC is a standard, that says the `sub` is the unique identifier for the user. The reason we made the `user_id` configurable in the OAuth connector is because there is no standard claim to identify the user.

In addition, what you are actually trying to achieve is a better way to identity the user, when doing set-team, so that you don't have to use some guid, is that correct? You can already do this by specifying the username, [which already is configurable](https://github.com/concourse/concourse/blob/master/skymarshal/skycmd/oidc_flags.go#L27).

cappyzawa:
@pivotal-jwinters Oh... I misunderstood that user name key is only related to the display of UI 🤦

Now, I try to login with `CONCOURSE_OIDC_USER_NAME_KEY`, then succeed.
Thanks for your useful information :bow:

@cirocosta Thanks!!

PR #3870, opened 2019-05-15
ci, web: fix dashboard preview performance

### LINKED ISSUE #3748: [performance regression] dashboard exhausting cpu
## Bug Report

Rendering the dashboard uses excessive amounts of cpu. Usage remains high even if the animations are outside the viewport, or if no jobs are running (no animations)

## Steps to Reproduce

Run many (>150) jobs simultaneously, view low density dashboard.

## Expected Results

Dashboard should not consume all cpu. Cpu usage should drop if animations are outside the viewport.

## Actual Results

Browser renders at < 5 FPS
Screen Shot 2019-04-17 at 12 04 58 PM

## Version Info

* Concourse version: 5.1.0
* Infrastructure/IaaS: GCP
* Did this used to work? Yes in 5.0.1

### LINKED ISSUE #3817: Add task in pipeline to benchmark UI
Try out elm-benchmark

pivotal-jamie-klassen:
fixes #3817
fixes #3748

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -323,6 +323,10 @@ even some automated fixes!
 Run `yarn format` to format the elm code according to the official Elm Style
 Guide. Powered by [elm-format](https://github.com/avh4/elm-format).
 
+### Elm benchmarking
+
+Run `yarn benchmark`.
+
 ### Running the acceptance tests (`testflight`)
 
 The `testflight` package contains tests that run against a real live Concourse.
--- a/ci/pipelines/prs.yml
+++ b/ci/pipelines/prs.yml
@@ -86,6 +86,11 @@ jobs:
     file: concourse-pr/ci/tasks/yarn-test.yml
     input_mapping: {concourse: concourse-pr}
     tags: [pr]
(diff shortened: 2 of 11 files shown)


cirocosta:
Hey @pivotal-jamie-klassen , 

For someone reviewing this - how could one verify it? Are you running a specific set of pipelines that are huge enough to make it slow and then comparing the before & after, or relying more on the benchmarking suite through `yarn benchmark`?

Thanks!

pivotal-jamie-klassen:
@cirocosta - @xtreme-james-thomson and I did both. For the most part the benchmark suite provided faster feedback. For reference, when we compared the old implementation to the tsort one currently on master, master was about 10 times slower on the sample data (less than 10 jobs), whereas the current benchmarks show performance differences of <5%. We did the same thing by creating the prod pipeline 100 times and profiling at the dashboard, seeing the compute-heavy section (yellow stuff before the purple) shorten by about a factor of 10.

cirocosta:
oooooh nice! That sounds great! 

Thanks for sharing!

PR #3878, opened 2019-05-17
Enable cert generation using Let's Encrypt

henderjm:
Instead of passing cert and key to serve over tls, the enable_autocert flag sets up ATC to get a certificate from Let's Encrypt.

Based on [this PR](https://github.com/concourse/atc/pull/199)

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -1006,7 +1006,7 @@ func (cmd *RunCommand) tlsConfig(logger lager.Logger, dbConn db.Conn) (*tls.Conf
 	if cmd.isTLSEnabled() {
 		tlsLogger := logger.Session("tls-enabled")
 		if cmd.EnableAutocert {
-			tlsLogger.Info("starting autocert manager")
+			tlsLogger.Debug("using-autocert-manager")
 
 			cache, err := newDbCache(dbConn)
 			if err != nil {
@@ -1017,12 +1017,11 @@ func (cmd *RunCommand) tlsConfig(logger lager.Logger, dbConn db.Conn) (*tls.Conf
 				Cache:      cache,
 				HostPolicy: autocert.HostWhitelist(cmd.ExternalURL.URL.Hostname()),
 				Client:     &acme.Client{DirectoryURL: cmd.ACMEURL.URL.String()},
-				ForceRSA:   true,
 			}
 			tlsConfig.NextProtos = append(tlsConfig.NextProtos, acme.ALPNProto)
 			tlsConfig.GetCertificate = m.GetCertificate
 		} else {
-			tlsLogger.Info("configuring tls using provided certs")
+			tlsLogger.Debug("loading-tls-certs")
 			cert, err := tls.LoadX509KeyPair(string(cmd.TLSCert), string(cmd.TLSKey))
 			if err != nil {
 				return nil, err
@@ -1097,12 +1096,6 @@ func (cmd *RunCommand) validate() error {
 			errs = multierror.Append(
 				errs,
 				errors.New("cannot enable autocert if --tls-cert or --tls-key are set"),
-			)
-		}
-		if cmd.TLSBindPort != 443 {
-			errs = multierror.Append(
-				errs,
-				errors.New("cannot enable autocert if external port is not 443"),
 			)
 		}
 	case cmd.TLSCert != "" && cmd.TLSKey != "":
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -58,6 +58,8 @@ import (
 	"github.com/tedsuo/ifrit/grouper"
 	"github.com/tedsuo/ifrit/http_server"
 	"github.com/tedsuo/ifrit/sigmon"
+	"golang.org/x/crypto/acme"
+	"golang.org/x/crypto/acme/autocert"
 
 	// dynamically registered metric emitters
 	_ "github.com/concourse/concourse/atc/metric/emitter"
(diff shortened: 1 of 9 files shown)


atc/atccmd/command.go
```diff
@@ -994,17 +999,36 @@ func (tripper mitmRoundTripper) RoundTrip(req *http.Request) (*http.Response, er
...
+			cache, err := newDbCache(dbConn)
+			if err != nil {
+				return nil, err
+			}
+			m := autocert.Manager{
+				Prompt:     autocert.AcceptTOS,
+				Cache:      cache,
+				HostPolicy: autocert.HostWhitelist(cmd.ExternalURL.URL.Hostname()),
+				Client:     &acme.Client{DirectoryURL: cmd.ACMEURL.URL.String()},
+				ForceRSA:   true,
```
vito: nit: looks like this is deprecated and ignored now: https://godoc.org/golang.org/x/crypto/acme/autocert#Manager

atc/atccmd/command.go
```diff
@@ -994,17 +999,36 @@ func (tripper mitmRoundTripper) RoundTrip(req *http.Request) (*http.Response, er
...
+			}
+			tlsConfig.NextProtos = append(tlsConfig.NextProtos, acme.ALPNProto)
+			tlsConfig.GetCertificate = m.GetCertificate
+		} else {
+			tlsLogger.Info("configuring tls using provided certs")
```
vito:
```suggestion
			tlsLogger.Debug("loading-tls-certs")
```

atc/atccmd/command.go
```diff
@@ -994,17 +999,36 @@ func (tripper mitmRoundTripper) RoundTrip(req *http.Request) (*http.Response, er
...
+	tlsConfig = atc.DefaultTLSConfig()
 
 	if cmd.isTLSEnabled() {
-		cert, err := tls.LoadX509KeyPair(string(cmd.TLSCert), string(cmd.TLSKey))
-		if err != nil {
-			return nil, err
-		}
+		tlsLogger := logger.Session("tls-enabled")
+		if cmd.EnableAutocert {
+			tlsLogger.Info("starting autocert manager")
```
vito: These log lines should match our [style guide](https://github.com/concourse/concourse/wiki/Concourse-Go-Style-Guide#logging) (and probably be at `Debug` level?):

```suggestion
			tlsLogger.Debug("using-autocert-manager")
```

atc/atccmd/command.go
```diff
@@ -1060,28 +1084,38 @@ func run(runner ifrit.Runner, onReady func(), onExit func()) ifrit.Runner {
...
+			)
+		}
+	case cmd.EnableAutocert:
+		if cmd.TLSCert != "" || cmd.TLSKey != "" {
+			errs = multierror.Append(
+				errs,
+				errors.New("cannot enable autocert if --tls-cert or --tls-key are set"),
+			)
+		}
+		if cmd.TLSBindPort != 443 {
```
vito: Where does this requirement come from? :thinking: This may be problematic for setups that run `web` on a higher port (>1024) and use a load balancer to forward to it.

atc/config.go
```diff
@@ -514,6 +514,9 @@ func DefaultTLSConfig() *tls.Config {
 
 		// Security team recommends a very restricted set of cipher suites
 		CipherSuites: []uint16{
+			tls.TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,
+			tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
+			tls.TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,
```
vito: Is this intentionally added or is this a merge artifact?

go.mod
```diff
@@ -45,6 +45,7 @@ require (
 	github.com/cloudfoundry/bosh-utils v0.0.0-20181224171034-c2cf699102bd // indirect
 	github.com/cloudfoundry/go-socks5 v0.0.0-20180221174514-54f73bdb8a8e // indirect
 	github.com/cloudfoundry/socks5-proxy v0.0.0-20180530211953-3659db090cb2 // indirect
+	github.com/concourse/atc v4.2.2+incompatible
```
vito: This looks bogus - bad import somewhere?

atc/config.go
henderjm: We had some issues with TLS handshakes with the suites configured so had to, somewhat reluctantly, add the most secure cipher suites [autocert uses](https://github.com/golang/crypto/blob/master/acme/autocert/autocert.go#L356). We put this through SSL Labs to verify the suites we added are in fact secure.

I'm open to alternative ideas.

vito: Sounds good, just making sure it wasn't a conflict as we happened to touch that code very recently:  https://github.com/concourse/concourse/commit/6f8880a7dcb87a0bee102820855959f7fd4b30f7

atc/atccmd/command.go
henderjm: This is probably a misunderstanding reading the autocerts Listeners which only listens on port 443. I have done some testing to make sure the ALPN challenge isn't restricted to having the external port on 443 so this check can be removed.

go.mod
henderjm: very bad! 😇

jpluscplusm:
Whilst I 100% welcome this feature (thanks @henderjm!), could I ask: @vito @topherbullock what's changed since @takeyourhatoff implemented this back in https://github.com/concourse/atc/pull/199? What caused the change of heart? :-)

vito:
@jpluscplusm I think it just took time to warm up to the idea of Let's Encrypt. It still seemed pretty new and unknown to us at the time. At this point the benefits of making TLS so easy to set up outweigh the technical things we pointed out back in the day.

Issue #3521, opened 2019-03-15
Viewing Large Build Logs Takes A Very Long Time


astrieanna:
I tried to find a similar issue, since I've seen this on every team I've been on in the past two years, but I failed to find one.

## What challenge are you facing?

Our tests have very noisy logs (as any large ginkgo test suite run in verbose does). When we load build pages, it takes a significant time to load the text of the logs. In some cases, it can be very hard to tell if the page is "stuck" while loading or just taking forever to load. The worst is when you get frustrated and click on a link (to another build or back to the pipeline) because you think the text is just not loading, and then there's a long pause, and then the text appears, and then you immediately end up on the page you clicked the link to -- you were so close to your goal of reading that text, but it's gone and you have to start over on the "waiting to load text" timer if you go back to the build page.

## What would make this better?

It would help if the text displayed faster. It's just text, so I assume it's not a network/download problem, but a display slowness. The current concourse interface seems to just be slow to display a giant chunk of text, and I'm not familiar enough with how the UI is structured or where the time is spent to suggest a fix for that (other than that someone should be profiling the UI code).

## Are you interested in implementing this yourself?

Not at the moment.

vito:
Yeah, this is a constant battle right now with our current architecture for the build logs. I feel this pain too, particular with builds that have pretty UIs like progress bars, which are cheap visually but costly at the transport level, as they work by constantly re-writing the same line in place.

The long and short of it is that ANSI escape sequences, the pseudo-protocol that determines how text renders in your terminal, are very hard to process "live" and render in a web browser. We've optimized the hell out of this in many different ways but there's only so much we can do without an architectural change. (Also, despite it being just text, sometimes it *is* a network/download speed problem, because of things like progress-bars and such.)

I'll be opening an issue soon to propose an architectural change that I think will help speed this up dramatically. For now I'll just throw this issue directly on the roadmap until I have time to write the other issue.

In the meantime, I have one bit of information to ask: is this a pain point you feel with finished builds, or also ones that are in-flight?

Thanks!

astrieanna:
It's both, in that opening either can be very slow. The in-progress ones are worse because the currently running step is open by default, so you don't even get a choice about whether those are the logs you're even interested in. The loading is less bad on finished builds, but I think that's largely because you only pay the waiting cost when you expect it (i.e. when opening the tests task).

[...]

PR #3894, opened 2019-05-21
stepTree: Disable autoexpand until #3521 is fixed.

kallisti5:


kallisti5:
Feel free to revert once #3521 is fixed

kallisti5:
Builds for me.
```
$ /home/kallisti5/Code/concourse/node_modules/elm/bin/elm make src/Build/StepTree/StepTree.elm
Success!                                                             
```

vito:
Does this result in a measurable performance improvement? :thinking: I thought most of the time spent was in processing the events, not rendering the output. The events are processed regardless of whether the step is expanded or collapsed.

kallisti5:
Part of it is just the browser showing huge of text from long builds.

Example data:
  https://build.haiku-os.org/buildbot/#/builders/1/builds/10/steps/16/logs/stdio

When concourse shows the build status, it seems to render all 17,000+ lines over and over which kills the browser performance. It's difficult to even scroll to the top of the page to "unexpand" the task.

After much clicking, if you're able to get to the top of the page, collapsing results in the page working as expected (except, a bit sluggish)

Also, this shouldn't be a client performance issue.  Running a Ryzen 1800x here with Chrome.. so that's a near best-case client.

kallisti5:
I've pushed out my initial work here:
  https://github.com/haiku/infrastructure/tree/master/playground/concourse

It assumes you're using the stock docker-compose deployment of concourse.

Once you're logged in via fly:
```
./apply-lifecycle.sh master secrets.source
```

Fire off a build of haiku / master-x86_64, and enjoy your browser crawling in a panic to keep up :-)

vito:
> When concourse shows the build status, it seems to render all 17,000+ lines over and over which kills the browser performance. It's difficult to even scroll to the top of the page to "unexpand" the task.

That sounds strange. Could be a performance regression. We do all kinds of lazy-rendering to prevent it from having to re-draw everything. I think that warrants investigation before collapsing steps by default. :thinking: Especially if the page remains sluggish after collapsing the step.

/cc @pivotal-jamie-klassen

pivotal-jamie-klassen:
We have validated that this is a decent stopgap with the following experiment:

We replaced [the openEventStream callback](https://github.com/concourse/concourse/blob/master/web/public/index.html#L202-L226) with
```javascript
app.ports.openEventStream.subscribe(function(config) {
  var openEvent, statusEvent, initializeEvent, finishTaskEvent, finalStatusEvent, endEvent
  // ... defining these dummy events
  var logEvent = {
    type: "event",
    target: {url: "http://localhost:8080/api/v1/builds/1/events"},
    data: `{
      "data":{
        "time":1560959925,
        "origin":{"id":"5d0a5ba8","source":"stderr"},
        "payload":"some long line of log output\\n"
      },
      "event":"log",
      "version":"5.1"
    }`
  };
  var buffer = [openEvent, statusEvent, initializeEvent]
    .concat(Array(10000).fill(logEvent))
    .concat([finishTaskEvent, finalStatusEvent, endEvent]);
  function flush() {
    if (buffer.length > 0) {
      app.ports.eventSource.send(buffer);
      buffer = [];
    }
  }
  setInterval(flush, 200);
});
```
which means that loading `http://localhost:8080/builds/1` will always mean a single Elm update loop occurs with a batch of 10000 identical events. We profiled build page with auto-expanding steps and saw this:

Screen Shot 2019-06-19 at 2 52 26 PM

We see ~2.7 seconds spent on
1. our Elm code processing the events
2. the Elm runtime updating the virtual DOM
3. the browser recomputing styles/repainting

Then we applied the change in this PR and saw this:
Screen Shot 2019-06-19 at 2 53 12 PM
We see ~1 second spent almost entirely on processing the batch of events with minimal effort on the Elm runtime or layout engine's part.

Furthermore, we have anecdotally observed that log lines don't actually take longer and longer to render as the output gets longer -- the real expense is just loading a huge amount of events on the first page load, so the above sample should be representative.

This gives a fair amount of confidence that there is no obvious inefficiency causing auto-expanded steps to be expensive; rather, we will need to think carefully about optimizing/offloading the three steps above. So @xtreme-james-thomson and I will open another PR that does the same as this one as well as some refactoring cleanup.

Upon discussion with @vito we agreed that this is a reasonable compromise to make in the interest of performance, and we could think of a couple other reasons why collapsing all steps on page load might be desirable.

vito:
> Upon discussion with @vito we agreed that this is a reasonable compromise to make in the interest of performance, and we could think of a couple other reasons why collapsing all steps on page load might be desirable.

The other reason is mainly that auto-expanding can actually make it difficult to see which steps failed, since the output pushes the rest of the steps off-screen. For example, if you've got an `in_parallel:` with 3 tasks and task 2 and 3 fail, you'd probably only see task 2 on first page load and have no idea task 3 also failed as its step header gets pushed offscreen.

So with that + the performance justification I think we can go forward with this change not only as a stopgap solution but as a potential UX improvement. :slightly_smiling_face: 

Thanks for the idea, @kallisti5!

kallisti5:
Glad this is getting fixed!

If you're looking to give a quick status overview, you could track the "number of log lines per build step". Jenkins and buildbot use these numbers to calculate a rough progress bar. (if the last run of this step was 1000 lines, then since this step has produced 200 lines it is likely ~20% done)

Anyway, that's a good enhancement for another ticket.   Thus unblocks haiku-os.org from potentially using concourse.ci for their (complex) builds.

Thanks!

vito:
@kallisti5 Nice, didn't know Haiku were evaluating Concourse! That's a cool project, I remember [hacking aroud with it](https://github.com/vito/MusicPlayer) a long while back.

Also this is out in v5.4 now. I think auto-collapsing for finished builds makes sense, though as a follow-up we can make it a bit easier to notice where the failure is. I opened #4111 for that.

Also, do you think there's a middle ground where we still auto-expand for in-flight steps? Or was that the source of the slowness? In my experience running builds are less laggy than finished ones, and it's otherwise nice to be able to just click in to a build and watch all the output from beginning to end. As it's implemented right now I have to expand each step.

Thanks!

kallisti5:
@vito you can look at our current deployment of concourse now for some ideas on how 5.4.0 reacts to large builds.  It's still slow, but somewhat better.

What's weird is when the build is running, you have to wait for each task to "repopulate"
(Each step spins as busy for 20+ seconds, then moves onto the next. It looks like the build is "still happening on those steps, but it is just "replaying the logs" behind the scenes)

This is the "longest" one:
https://ci.haiku-os.org/teams/nightly/pipelines/toolchain-master

kallisti5:
Also to compare:
https://build.haiku-os.org/buildbot/#/builders/1/builds/16

For all of buildbot's faults, their status page loads *really* quickly and is precise due to loading only the last ~40 log lines until you dig into details. I still think that's the "real" long-term answer.

PR #3902, opened 2019-05-22
#3301 adding lock around worker placement

deniseyu:

### DIFF
--- a/topgun/pipelines/lots-ata-time-2.yml
+++ b/topgun/pipelines/lots-ata-time-2.yml
@@ -9,84 +9,212 @@ jobs:
   plan:
   - get: timer
     trigger: true
-  - task: exit
-    config:
-      platform: linux
-      image_resource:
-        type: registry-image
-        source: {repository: concourse/dev}
-      run:
-        path: /bin/sh
-        args:
-        - -c
-        - |
-          exit 2
+  - in_parallel:
+    - task: exit
+      config:
+        platform: linux
+        image_resource:
+          type: registry-image
+          source: {repository: concourse/dev}
+        run:
+          path: /bin/sh
+          args:
+          - -c
+          - |
+            exit 2
+    - task: exit
+      config:
+        platform: linux
+        image_resource:
+          type: registry-image
+          source: {repository: concourse/dev}
+        run:
+          path: /bin/sh
+          args:
+          - -c
+          - |
+            exit 2
+    - task: exit
+      config:
+        platform: linux
+        image_resource:
+          type: registry-image
+          source: {repository: concourse/dev}
+        run:
+          path: /bin/sh
+          args:
+          - -c
+          - |
+            exit 2
 
 - name: exiter2
   plan:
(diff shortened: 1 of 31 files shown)


kcmannem:
dont merge yet, we messed up image pulling

pivotal-jwinters:
@kcmannem @deniseyu 

This seems kinda roundabout to me. You are basically saying find me a worker and create a container, then you expose back the worker just so that we can look up the container all over again. 

Why don't we just add a `CreateContainer` method in `worker.Client`, which returns a `Creating` container so that we can move closer to not caring about worker selection. Then we can move a bunch of the worker.FindOrCreateContainer logic, such as input streaming, into a method on the creating container itself (or something that takes in a creating container).

It will probably also help to separate the `FindOrCreate` logic into `Find` or `Create` and push that orchestration into core. Once we move to ephemeral check containers we will never be looking up an existing container, only ever creating new ones.

\cc @ddadlani

kcmannem:
@pivotal-jwinters I think the goal is to get to what you were saying. I agree its kind of weird right now. But this wasn't meant to be that big refactor but rather way to get things correct again and ease some pain points for users. This gives them something while we can work on longer term solutions after.

marco-m:
@wagdav FYI

wagdav:
@vito @ddadlani  would you please take a look at this PR? My understanding is that this would be a fix for #3301 and possibly other scheduler related problems. Thanks!

kcmannem:
@wagdav this change exposed a larger issue where checks error every now and then. Another recent change in how check sessions are handled also makes it difficult to patch this issue. We are still working on a fix for that error, after which we will merge this into master

kcmannem:
@wagdav we're gunna workaround the issue, this should be going in considering the tests pass.

kcmannem:
@ddadlani fixed the tests, we accidentally commit changes to the docker-compose. its ready for review now.

topgun/pipelines/lots-ata-time-2.yml
```diff
@@ -9,84 +9,212 @@ jobs:
...
-      image_resource:
-        type: registry-image
-        source: {repository: concourse/dev}
-      run:
-        path: /bin/sh
-        args:
-        - -c
-        - |
-          exit 2
+  - in_parallel:
```
kcmannem: @ddadlani instead of making a new topgun test cause its heavy, I decided to just update the existing one to now assert that in_parallel tasks get distributed evenly.

ddadlani: Sounds good

ddadlani:
fixes #3301

aledeganopix4d:
> fixes #3301

@ddadlani your comment didn't really close issue #3301, was that intended? (I thought that GitHub should close the issue with this type of comments...)

ddadlani:
@aledeganopix4d yes it should have. I think I had to comment that before I merged this PR, though. Thanks for pointing it out!

PR #3903, opened 2019-05-22
Change default container placement strategy to fewest build containers

kcmannem:
Depends on #3301 

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -115,8 +115,9 @@ type RunCommand struct {
 	ResourceCheckingInterval     time.Duration `long:"resource-checking-interval" default:"1m" description:"Interval on which to check for new versions of resources."`
 	ResourceTypeCheckingInterval time.Duration `long:"resource-type-checking-interval" default:"1m" description:"Interval on which to check for new versions of resource types."`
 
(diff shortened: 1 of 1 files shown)


ddadlani:
#3301 has been merged, once the modified topgun test passes this should be ok to merge

kcmannem:
with volume streaming being mostly free. The importance of volume locality is reduced largely. We should now use fewest-build-containers to balance load appropriately on default.

kcmannem:
kicked off the tests again, they should pass and this is ok to merge since volume changes are in now.

kcmannem:
Ready to go in

wagdav:
I believe this strategy is a better default than the volume locality.  However, #3301 is still open. My understanding is that currently only the `limit-active-task` strategy provides race-free worker selection.

nader-ziada:
@kcmannem  is there anything holding this back from being merged?

kcmannem:
@nader-ziada just waiting on someone other than me to merge it in. Feel free to approve/merge.

ddadlani:
Hi @wagdav you are correct, the race issue is still present, but we feel that `fewest-build-containers` is the most general solution we have right now. Even with the race condition, average utilization with `fewest-build-containers` is better than `volume-locality`, which is the current default.

ddadlani:
we're not going to switch to this right now, we're going to wait for a more correct solution for fewest-build

PR #3908, opened 2019-05-23
topgun: increase test coverage for bbr

cirocosta:
Previously, there were no tests covering the use of the link provided by
the Concourse deployment through `concourse_db` - we were relying solely
on the use of properties.

Now, with the introduction of that type of configuration, we needed
tests for it.


cirocosta:
ps.: depends on https://github.com/concourse/concourse-bosh-release/pull/35

Issue #3734, opened 2019-04-14
Display resource related builds in the build page


mockersf:
## What challenge are you facing?

When viewing a build in Concourse, I can't easily check other builds that may have been triggered by the same version of a resource, or builds that have been triggered by a new version of a resource produced by the current build. I have to remember the version of the resource, go to the resource page, select my version to view it's inputs to / outputs of

Resource version details in a build:
<img width="983" alt="Screenshot 2019-04-15 at 00 52 04" src="https://user-images.githubusercontent.com/8672791/56100425-d0e47e00-5f18-11e9-8009-bffb19bc1868.png">

Resource version details in the resource page:
<img width="983" alt="Screenshot 2019-04-15 at 00 52 21" src="https://user-images.githubusercontent.com/8672791/56100434-e0fc5d80-5f18-11e9-96fc-df66b9773df5.png">

## What would make this better?

Either display the same information in the build page when clicking on a resource, or provide a link to the resource page with the related version already opened.
The resource screen is using the ID from the `resource_config_versions` table which can easily be added to the response from the `api/v1/builds/<build-id>/resources` call for either linking or fetching the related information on click.

I think displaying "inputs to" on the build page when the version is clicked would work best, but it would push down resource logs and add more information to the build page.

## Are you interested in implementing this yourself?

Yes!

vito:
Sounds like a good idea. I'm not a huge fan of the current resource page (it's pretty janky and a little slapped-together), but I do agree that it'd be nice to be able to navigate to that page - it makes the pipeline feel more 'cohesive'. This could end up being even nicer with 'causality' (https://github.com/concourse/concourse/issues/1538).

Maybe this could be done by adding an individual-version view and having the build's version link to it?

pivotal-jamie-klassen:
@vito having a page at `/teams/:team/pipelines/:pipeline/resources/:resource/versions/:id` does feel pretty intuitive, now that I think about it. I could imagine some of the page components being similar to the build page (like a scrolling list of tabs for each version). Maybe a button in the top right to enable/disable the version (parallel to the pause button on the pipeline page)...

chenbh:
This was added in #6662 and released in 7.2.0

PR #3920, opened 2019-05-26
new view for a specific version of a resource

mockersf:
related to #3734 

New view that display all informations related to a specific version of a resource. It's a copy / paste of the resource view, but it will display only one version and it's expanded by default:
resource version

This screen can be reached by clicking the `#` in the resource view:
resource

Or the `#` in the build view on a  resource:
build

For me, the main utility is to be able to go from build -> resource version -> other builds that are using the same version of this resource

Other information that could be interesting to display on this view:
- the logs that produce this version, but I think that would be a bigger change to get the logs 
- anything else ? I'm open to suggestions.

If the view is OK I think there should be more work done for code deduplication between resource view and resource version view, but I didn't want to go into that before discussing what goes where

### DIFF
--- a/atc/api/builds_test.go
+++ b/atc/api/builds_test.go
@@ -685,12 +685,14 @@ var _ = Describe("Builds API", func() {
 						Expect(body).To(MatchJSON(`{
 							"inputs": [
 								{
+									"id": 0,
 									"name": "input1",
 									"version": {"version": "value1"},
 									"pipeline_id": 42,
 									"first_occurrence": true
 								},
 								{
+									"id": 0,
 									"name": "input2",
 									"version": {"version": "value2"},
 									"pipeline_id": 42,
@@ -699,10 +701,12 @@ var _ = Describe("Builds API", func() {
 							],
 							"outputs": [
 								{
+									"id": 0,
 									"name": "myresource3",
 									"version": {"version": "value3"}
 								},
 								{
+									"id": 0,
 									"name": "myresource4",
 									"version": {"version": "value4"}
 								}
--- a/atc/api/present/public_build_input.go
+++ b/atc/api/present/public_build_input.go
@@ -7,7 +7,7 @@ import (
 
 func PublicBuildInput(input db.BuildInput, pipelineID int) atc.PublicBuildInput {
 	return atc.PublicBuildInput{
-		ID:              input.ID,
+		ID:              input.VersionID,
 		Name:            input.Name,
 		Version:         atc.Version(input.Version),
 		PipelineID:      pipelineID,
@@ -17,7 +17,7 @@ func PublicBuildInput(input db.BuildInput, pipelineID int) atc.PublicBuildInput 
 
 func PublicBuildOutput(output db.BuildOutput) atc.PublicBuildOutput {
 	return atc.PublicBuildOutput{
-		ID:      output.ID,
+		ID:      output.VersionID,
 		Name:    output.Name,
 		Version: atc.Version(output.Version),
 	}
--- a/atc/api/present/public_build_input.go
+++ b/atc/api/present/public_build_input.go
@@ -7,6 +7,7 @@ import (
 
 func PublicBuildInput(input db.BuildInput, pipelineID int) atc.PublicBuildInput {
 	return atc.PublicBuildInput{
+		ID:              input.ID,
 		Name:            input.Name,
 		Version:         atc.Version(input.Version),
 		PipelineID:      pipelineID,
@@ -16,6 +17,7 @@ func PublicBuildInput(input db.BuildInput, pipelineID int) atc.PublicBuildInput 
 
 func PublicBuildOutput(output db.BuildOutput) atc.PublicBuildOutput {
 	return atc.PublicBuildOutput{
+		ID:      output.ID,
 		Name:    output.Name,
 		Version: atc.Version(output.Version),
 	}
--- a/atc/build_inputs_outputs.go
+++ b/atc/build_inputs_outputs.go
@@ -6,6 +6,7 @@ type BuildInputsOutputs struct {
 }
 
 type PublicBuildInput struct {
+	ID              int     `json:"id"`
 	Name            string  `json:"name"`
 	Version         Version `json:"version"`
 	PipelineID      int     `json:"pipeline_id"`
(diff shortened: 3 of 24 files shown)


pivotal-jwinters:
@vito what about using the version hash? We use that as the identifier elsewhere, and then we won't have to deal with weird characters in the url.

vito:
@pivotal-jwinters I think hashes would end up even being confusing when - for example if I'm trying to link to commit `1985b7128198358dde2390ec69ede303ff5962eb` the URL would be something like `e739c0e0fe1853d7a909e5657dceb78e` instead. :sweat_smile: We also don't expose the hashes in the UI or API - they're just an internal detail to make uniqueness easier in the database schema atm.

Since versions are currently expected to be fairly regular (simple strings) I think we can probably get away with having readable URLs. We may run into a similar challenge in the future for spaces since they'll also be routing based on 'config fragments' (assuming concourse/rfcs#24) like `branch:abcdef`.

mockersf:
1/ the "`#`" looks ok to me but it's not great, it can change

2/ yes, it's copy pasted and can be removed

3/ as the version is a free json object, it can get quite complicated, contains illegal characters in a URL, be very long (in a toy resource I once put a 5MB version just to see if it worked...)... and not be very readable once url encoded
I found this library that tries to do something clever to help on this subject : https://github.com/Sage/jsurl
sample json: (can't be used directly in a url, has to be encoded)
```
{"name":"John Doe","age":42,"children":["Mary","Bill"]}
```
url encoded json: (safe to use in a url but can get very long)
```
%7B%22name%22%3A%22John%20Doe%22%2C%22age%22%3A42%2C%22children%22%3A%5B%22Mary%22%2C%22Bill%22%5D%7D
```
encoded with jsurl: (looks readable and mostly safe, but made for query params, path params have additional contraints to url encode)
```
~(name~'John*20Doe~age~42~children~(~'Mary~'Bill))
```
base 64: (which is the safest from a url point of view, but not readable at all)
```
eyJuYW1lIjoiSm9obiBEb2UiLCJhZ2UiOjQyLCJjaGlsZHJlbiI6WyJNYXJ5IiwiQmlsbCJdfQ==
```
custom encoding: (a lot of special cases to think about. For example, what if a value contains a ","?)
```
name:John%20Doe,age:42,children:mary,bill
```

vito:
We ended up wanting another form of querying by version fields for `fly pin-resource-version`: https://github.com/concourse/concourse/pull/3131#issuecomment-459874740 - so I think we'll have to bite the bullet at some point and figure out how these should be passed to the API.

I've opened an issue for it here: https://github.com/concourse/concourse/issues/3932

It might be going a bit too far to block this PR on it, though. :thinking: Unless you're willing to take a swing at that, @mockersf, I'm fine with merging this with IDs in the URL for now.

pivotal-jamie-klassen:
> the version is a free json object
@mockersf hopefully a helpful clarification: it has to deserialize to a `map[string]string` https://github.com/concourse/concourse/blob/master/atc/resource_types.go#L7.

Lindsayauchin:
@mockersf  

**Feedback on the UX of Adding the # to the resource page:** 
I think we need to take a look at the UX implications of creating a button to link out in that area on the UI of the Resource page beside the pin toggle and the disable/enable toggle. 

In the Resource page, a link out to a new view could make sense on the **version** but, that currently expands and collapses the version meta data and inputs to/outputs.  If  we went down that path of adding a link out there to a new view we would have to create two clickable hotspots on that container. This could confuse users in the short term from the current expected behaviour of the expand and collapse reveal.

Just to get a better understanding of your current work flow. Your goal is to see a list of all builds that use a version of a resource from the build you are currently viewing. 

The pain you are currently experiencing is by navigating back to the pipeline view selecting that resource, going to the resource page then parsing out the version from that list, expanding that version to then view the builds this inputs to so you can navigate to those builds. 

The solutions presented could use some further exploration but, can you confirm my synopsis of your use case?

mockersf:
@Lindsayauchin yes you have it right

In the resource page, the link is on another button to not be on expand / collapse toggle. That's not very pretty, but I think the one on the build screen is worse, as there are no button there so I added the link in the expanse / collapse toggle space which is confusing

@vito I removed the check bar

pivotal-jamie-klassen:
Could work if the build steps had an arrow that you click to expand/collapse and then the version value itself was a link?

mockersf:
I'm not convinced with an array, as there is already a downward arrow on the build screen to mark a get step.

Maybe a small triangle that would turn when expanded?

pivotal-jamie-klassen:
@mockersf - I had to mentally `:s/array/arrow/` there, haha. Turning triangle is solid, or a plus/minus toggle would be distinct as well. Definitely worth getting @Lindsayauchin's input on this one.

Lindsayauchin:
@mockersf what do you think about adding the entry point for this new view as a link from a button that we put on the build step of that version-> expand that resource version -> button on the meta data area links out to the other builds that are using the same version of this resource? 

Screen Shot 2019-06-03 at 2 11 47 PM

vito:
IMO there's value in having a version-specific page anyway, and having the builds link directly to the versions used by it would make the pipeline feel more cohesive.

jama22:
@vito do you mean to say that this could lead to causality? mind == blown

vito:
@jama22 Yeah, that'd be a good page for it in the future. 👨💥

Lindsayauchin:
@mockersf @vito 

What if the resource version could be reach via link on the version string with an old school underline style? We would just need to turn off the expand collapse hotspot in that area of the resource container div. 

Screen Shot 2019-06-12 at 2 34 40 PM

Update the resource page and hide some of the unnecessary elements like the "checking successfully' and the refresh.  

Screen Shot 2019-06-12 at 2 42 49 PM

xtremerui:
Hi @mockersf  just a status check that are you still working on this PR?

jomsie:
Hi @mockersf! 

We've recently revamped our PR policy, as seen here: https://blog.concourse-ci.org/nobody-wants-a-stale-bot/#improving-pull-request-engagement.

I'm going to go ahead and close this PR for the time being, let me know if you'd like to continue working on it!

Also @matthewpereira I'm not sure if you've seen this PR, but if you have any opinions on it, feel free posting here!

PR #3922, opened 2019-05-27
atc: add aborting for a pending build

### LINKED ISSUE #3900: Abort a pending build is no op
## Bug Report

Screen Shot 2019-05-22 at 11 34 05 AM

When trying to abort a build that is pending due to resource checking failing, the request to abort is successful but the build is stuck in pending.
You can also see this in the backend `builds` table. The build is flagged to be aborted, but the status is stuck in pending.

## Steps to Reproduce

pipe.yml
```
resources:
- name: test
  type: registry-image
  source:
    repository: blah

jobs:
- name: test
  plan:
    - get: test

```
config the above pipeline, un-pause it, then trigger `test` job. Observe the build is in pending state with
```
waiting for a suitable set of input versions
test - no versions available
```
abort the build by clicking the `X` button. Observe the request is 204 successful but if you refresh the page notice the build is still in pending state.

So now if you change the repository to `busybox` to make the resource check successfully, the pending build will exit the pending state and then become aborted immediately.

## Expected Results

The build aborts immediately or shortly after clicking abort. Some kind of feedback that the abort button did something.

## Actual Results

The build is stuck in pending.... forever....

## Version Info

* Concourse version: 5.2.0
* Deployment type (BOSH/Docker/binary): bosh & docker (saw on both)
* Did this used to work? Not sure!

@xtremerui

xtremerui:
fixes #3900 in core/api. Still needs UI works.


xtremerui:
@pivotal-jamie-klassen feel free to have the UI changes in this PR or merge it to master first. Thx.

pivotal-jamie-klassen:
@xtremerui UI behaves almost exactly like an aborted build, except (due to the lack of a build plan) shows an infinite spinner on the build page:

Screen Shot 2019-05-27 at 4 22 24 PM
Screen Shot 2019-05-27 at 4 22 12 PM

We've never really had a design for this, so I think I'll just make a blank page instead of the (somewhat-harmful-and-misleading) forever spinner.

vito:
Seems like the unit tests are failing. Same failure happened twice - seems legit?

pivotal-jamie-klassen:
*update* I've left the UX in a not-actively-misleading-but-still-unpleasant state. When you watch the build from the prep stage and it gets cancelled, you see:
Screen Shot 2019-06-04 at 5 53 53 PM
and when you revisit a cancelled build, you see:
Screen Shot 2019-06-04 at 5 54 06 PM

I was hesitant to make to drastic of a UI choice on the fly like this, especially if the change is going to be reverted/overwritten with something new in the next major release. @Lindsayauchin you might have an idea and we can implement that if now is the right time. Otherwise we can leave it as is, or try for something basic, reusing something like the old "build-level error" area to show that there's nothing else to see here:
Screen Shot 2019-06-04 at 6 04 49 PM

@pivotal-jwinters @xtremerui @vito I'll leave this where it is an wait for more feedback. Pretty sure the tests should pass 🤞

atc/api/buildserver/plan.go
```diff
@@ -13,6 +13,10 @@ func (s *Server) GetBuildPlan(build db.Build) http.Handler {
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Content-Type", "application/json")
```
vito: this header should be set after the check below

pivotal-jamie-klassen:
OK, since i had to take another pass, I actually added the 'cancelled' message mentioned above.

Lindsayauchin:
@pivotal-jamie-klassen Sorry, I am late to this!

hmmmm, adding a cancelled state would have app wide implications so, perhaps for now this is something we scope down to what you have already implemented. 

Some quick UI tweaks: 
- [ ] Add back the black band where the loading message lives  to be consistent with where other messaging on the build lives. 

- [ ] 'build cancelled' instead of just cancelled (?) 

Screen Shot 2019-06-07 at 2 12 32 PM

PR #3937, opened 2019-05-29
Added support for InfluxDB batches of more than 1 and made the emission channel size configurable.

rudolfv:
@vito We still need to test the latest changes (making all the magic numbers configurable) against one of our concourse stacks, but I did manage to build a full release image in our slightly altered release pipeline. We are not running the upgrade or downgrade jobs, but everything else passes.


atc/metric/emit.go
```diff
@@ -66,7 +66,13 @@ var (
 	emissions       chan eventEmission
 )
 
-func Initialize(logger lager.Logger, host string, attributes map[string]string) error {
+func Initialize(logger lager.Logger, host string, attributes map[string]string, bufferSize int) error {
```
cirocosta: As we always expect `bufferSize` to be `uint` anyway, wdyt of changing the signature to take an `uint` instead of `int`?

```suggestion
func Initialize(logger lager.Logger, host string, attributes map[string]string, bufferSize uint32) error {
``` 

It seems to make that it'd be better to do the casting at the `make(chan ...` that comes later rather than at the moment that we pass the value to `Initialize` 🤔

atc/metric/emit.go
```diff
@@ -110,6 +116,9 @@ func Deinitialize(logger lager.Logger) {
 }
 
 func emit(logger lager.Logger, event Event) {
+	logger.Debug("emit-event", lager.Data{
```
cirocosta: We already have a way of having the metrics emitted logging to `stdout` through the `--emit-to-logs` variable - did you want this logline for the purposes of checking if metrics are still being emitted regardless of the configured emitter, or for the same purposes as `--emit-to-logs`?

atc/metric/emitter/influxdb.go
```diff
@@ -24,9 +26,22 @@ type InfluxDBConfig struct {
 	Password string `long:"influxdb-password" description:"InfluxDB server password."`
 
 	InsecureSkipVerify bool `long:"influxdb-insecure-skip-verify" description:"Skip SSL verification when emitting to InfluxDB."`
+
+	// https://github.com/influxdata/docs.influxdata.com/issues/454
```
cirocosta: Thanks for giving us the context on "why 5000"! That's very helpful.

Wdyt of leaving that to the PR though? As one could always trace the `5000` back to this PR, I'd say leaving that in the PR's comments (or even in the commit message) would be enough 🤔 In the codebase, we don't have many other cases where we add references like this.

atc/metric/emitter/influxdb.go
```diff
@@ -60,34 +81,56 @@ func (emitter *InfluxDBEmitter) Emit(logger lager.Logger, event metric.Event) {
...
+	batch = append(batch, event)
+	duration := time.Since(lastBatchTime)
+	if len(batch) > emitter.batchSize || duration > emitter.batchDuration {
+		logger.Debug("influxdb-pre-emit-batch", lager.Data{
+			"influxdb-batch-size": emitter.batchSize,
+			"current-batch-size": len(batch),
+			"influxdb-batch-duration": emitter.batchDuration,
+			"current-duration": duration,
+		})
+		go emitBatch(emitter, logger, batch)
```
cirocosta: I'm not very sure about making each emission a goroutine 🤔 If I understood correctly, this makes the buffer "infinite" as there's never really going to exist something that throttles the writes. Is this statement right? Maybe I got something wrong 😬

rudolfv:
> Thank you for the PR, @rudolfv !
> 
> I just added some questions / points 👍I feel a bit bad that we didn't have some testing around this area before 😬 Do you think it'd be worth adding some to this logic?
> 
> Thanks!

@cirocosta I am busy working on a unit test for this logic. Also see my replies to the other comments.

atc/metric/emit.go
rudolfv: I will make this change - makes sense to me to constrain it with a uint32 and do the cast at the last possible moment.

atc/metric/emit.go
rudolfv: This code is to check regardless of the configured emitter. If I remember correctly only one of the emitters can be configured at a time. So we cannot --emit-to-logs and InfluxDB emission configured at the same time.

atc/metric/emitter/influxdb.go
rudolfv: I will remove the comments and add as a comment here:

https://github.com/influxdata/docs.influxdata.com/issues/454
https://docs.influxdata.com/influxdb/v0.13/write_protocols/write_syntax/#write-a-batch-of-points-with-curl
5000 seems to be the batch size recommended by the InfluxDB team

atc/metric/emitter/influxdb.go
rudolfv: Making it a goroutine ensures that the construction of the batch points and writing to InfluxDB are immediately taken off the line of execution so that the emitter can continue to build up the next batch.

And this existing chunk of code in emit.go still ensures that events are passed on in a synchronous way to the InfluxDBEmitter after they have been read from the channel.

`func emitLoop() {
	for emission := range emissions {
		emission.logger.Debug("emit-event-loop", lager.Data{
			"event": emission.event,
		})
		emitter.Emit(emission.logger.Session("emit"), emission.event)
	}
}`

So there will only ever be one active batch in progress that will be submitted when either the size or duration limit is reached.

Does this answer your question? Or am I missing something?

atc/metric/emitter/influxdb_client.go
```diff
@@ -0,0 +1,34 @@
+package emitter
+
+import (
+	"time"
+
+	influxclient "github.com/influxdata/influxdb1-client/v2"
+)
+
+// This is a copy of the github.com/influxdata/influxdb1-client/v2/client.Client interface whose sole purpose is
```
rudolfv: I only added this file as a last resort. As stated in the counterfeiter docs for third party packages, I tried with
 //go:generate counterfeiter github.com/influxdata/influxdb1-client/v2/client.Client
and various other permutations of that, but none of them worked.

rudolfv:
@cirocosta This PR is ready for review again

cirocosta:
Hi @rudolfv !

Sorry for the very long delay, I'll get back to it very soon!

rudolfv:
@cirocosta Any updates on this?

rudolfv:
@vito @cirocosta I see the stale[bot] is about to close this. Do you have any updates for me?

vito:
@rudolfv Sorry! :slightly_frowning_face: I'll poke the issue so the stale bot backs off. We'll be sure to get to this soon. @cirocosta has been pretty busy with other commitments, and I've at least been busy putting together our roadmap. Unfortunately we don't have very consistent bandwidth for PR reviews - we're trying to improve on this as a project but it's slow going. I poked him about it yesterday, I'll try to get one of us to wrap this up soon depending on who's available.

rudolfv:
Thank you @vito!

atc/metric/emit.go
vito: I'm lukewarm on this; it seems like it'll be super noisy but I guess you can always filter it out. It's already at debug level but we have a decent amount of environments with debug logs enabled. Granted, we're used to maintaining filters to strip out the cruft.

If it was just added while you were developing and isn't useful anymore I would prefer we remove it, but if it was super useful in a pinch I'm fine with keeping it. :man_shrugging:

atc/metric/emit.go
```diff
@@ -132,13 +141,19 @@ func emit(logger lager.Logger, event Event) {
 
 	select {
 	case emissions <- eventEmission{logger: logger, event: event}:
+		logger.Debug("emit-event-write-to-channel", lager.Data{
```
vito: This on the other hand seems like it'd be a bit too noisy. :sweat_smile: Would you be OK with removing either this one or the above log?

(Also interesting that the above log line will run even with a `nil` emitter. Not sure if that's useful. I guess you'd notice it's not configured more quickly, but it also means those logs will show up even in places where metrics are undesired.)

atc/metric/emit.go
```diff
@@ -132,13 +141,19 @@ func emit(logger lager.Logger, event Event) {
...
+			"event": event,
+		})
 	default:
 		logger.Error("queue-full", nil)
 	}
 }
 
 func emitLoop() {
 	for emission := range emissions {
+		emission.logger.Debug("emit-event-loop", lager.Data{
```
vito: Also noisy - wouldn't this mean there are 3 logs emitted for every metric?

atc/metric/emitter/influxdb.go
```diff
@@ -46,48 +58,83 @@ func (config *InfluxDBConfig) NewEmitter() (metric.Emitter, error) {
...
+			"influxdb-batch-size": emitter.BatchSize,
+			"current-batch-size": len(batch),
+			"influxdb-batch-duration": emitter.BatchDuration,
+			"current-duration": duration,
+		})
+		emitter.SubmitBatch(logger)
 	}
 }
+
+func (emitter *InfluxDBEmitter) SubmitBatch(logger lager.Logger) {
```
vito: Just checking, is this thread-safe? I guess it is because all of the emits come from an emit loop? :thinking:

atc/metric/emitter/influxdb.go
```diff
@@ -46,48 +58,83 @@ func (config *InfluxDBConfig) NewEmitter() (metric.Emitter, error) {
...
+	batchToSubmit := make([]metric.Event, len(batch))
+	copy(batchToSubmit, batch)
+	batch = make([]metric.Event, 0)
+	lastBatchTime = time.Now()
+	go emitBatch(emitter, logger, batchToSubmit)
```
vito: I guess the idea behind this is to prevent a giant batch from submitting too slowly and causing the queue to fill up. Makes sense, since that was the original problem. But I wonder if batching would help mitigate that in and of itself, since submiting 5k at a time might cause less backpressure than 1 event 5000 times.

Have you observed any issues with this yet at large scale? It seems like the goroutines could potentially pile up due to a slow consumer.

I think @cirocosta had similar concerns on an earlier revision. I'm fine with how it is for now, but in the future we may want to add a max-in-flight or something. I guess at the end of the day slow consumers are hard to avoid, and this is why we've been thinking of standardizing on Prometheus. :sweat_smile:

cirocosta:
thaanks for looking at it @vito 🙌  sorry for dropping the ball on taking so long for doing it :( :( :(

rudolfv:
@vito I have removed all the extraneous logging. These are really all remnants of when I initially debugged the issue to find the exact point at which we were losing messages.

atc/metric/emitter/influxdb.go
rudolfv: @vito, yes, that is my understanding. The below chunk of code in emit.go will serialize the emission of events to the InfluxDB code. emitLoop -> emitter.Emit -> emitter.SubmitBatch

func emitLoop() {
	for emission := range emissions {
		emitter.Emit(emission.logger.Session("emit"), emission.event)
	}
}

atc/metric/emitter/influxdb.go
rudolfv: @vito We have not had any issues with this, but I understand your concerns. As a final test I would suggest that once we have merged this and built a release candidate, that we will spin up a concourse cluster similar to our production one and perform some load tests to give us a measure of confidence with all the 5.5 code changes.

rudolfv:
> @vito I have removed all the extraneous logging. These are really all remnants of when I initially debugged the issue to find the exact point at which we were losing messages.

Also, from running it in our production pipeline I can confirm that having debug mode on will create an insane amount of logging of which the usefulness is questionable. It makes sense to only add this again and do a custom build if we need to troubleshoot something specific further down the line.

atc/metric/emitter/influxdb.go
```diff
@@ -46,48 +58,82 @@ func (config *InfluxDBConfig) NewEmitter() (metric.Emitter, error) {
...
-	bp.AddPoint(point)
-
-	err = emitter.client.Write(bp)
+	err = emitter.Client.Write(bp)
 	if err != nil {
 		logger.Error("failed-to-send-points",
 			errors.Wrap(metric.ErrFailedToEmit, err.Error()))
 		return
 	}
+	logger.Info("influxdb-emitter-fork-influxdb-batch-emitted", lager.Data{
```
vito: just noticed this log line :sweat_smile: looks like more dev logging? (noticed the 'fork')

rudolfv: Thanks for catching this @vito. It was definitely dev logging. Removed and just waiting for the tests to pass.

rudolfv:
Thanks, @vito. Any ideas on when the 5.5 release is targeted for?

vito:
Don't know yet, I would assume a week or two at worst. There are few things left in the milestone.

rudolfv:
Thanks @vito, that's good news!

rudolfv:
@vito, where can I get hold of the 5.5 release candidates artifacts?

PR #3945, opened 2019-05-29
Add retries to the deploy step of bosh, as it is usually flaky

YoussB:
### Context:
- The `bosh topgun` task takes around 2 hours to run, and it is usually due to flakiness of running the `bosh deploy` step.
- @taylorsilva And I added a retry to the `bosh deploy` that is configurable from the task. We've currently set it to 3 times, so the deploy will be retried 3 times before admitting it is a legit failure (this excludes failures due to `timeouts`)


cirocosta:
Hey, I'm adding the `blocked` label to this one as it seems like we'd be better of going first over https://github.com/concourse/ci/issues/12 .

(please let me know if that's not true 😁 )

Thanks!

cirocosta:
Hey @YoussB ,

I'll close this one as we iterated quite a bit on topgun flakiness 😁  

thanks!

PR #3949, opened 2019-05-30
ci: wait for deployment rollout in k8s-deployment

cirocosta:
Previously, it could happen that `web` didn't have its deployment fully
complete, which used to make `port-forward` fail quite consistently.

By leveraging `rollout status` we can wait for the underlying pods to
get to a `READY` state (instead of plain `RUN`) before moving forward.


cirocosta:
oh oops, this was already pushed 😅  sorry!

PR #3954, opened 2019-05-31
ci: tag every commit that produces an RC build

pivotal-jamie-klassen:


vito:
Hmm is there another way to go about this? 🤔 This will end up with there being a *lot* of tags in the repo which would be kinda noisy. Tags are typically used more sparingly, e.g. for final versions. The [Releases page](https://github.com/concourse/concourse/releases for example practically treats tags like releases that haven't been given a title/description yet.

Is there another way we can make this information available to OSL? For example we used to put a commit sha in the blob filename for our resource types - I think for this very reason.

pivotal-jamie-klassen:
Good point! In this case, I'd like to suggest no longer using RC numbers, and simply versioning our pre-release builds with their associated commit hash. This changes the pipeline in a *slightly* more cross-cutting fashion so we'll take a bit of time to loop back on this PR.

vito:
@pivotal-jamie-klassen That meshes well with our plan for version-free 'branch' pipelines, so 👍

PR #3969, opened 2019-06-03
ci: move all CI config to its own repo

pivotal-jamie-klassen:


cirocosta:
Hey @pivotal-jamie-klassen ,

Thanks for the PR!

Would you mind elaborating a little on why are we performing this change of migrating it all to `concourse/ci`?

Thanks!

pivotal-jamie-klassen:
Hi @cirocosta; as you know, we have a few different release pipelines running very similar testing jobs: there's [master](https://ci.concourse-ci.org/teams/main/pipelines/concourse), [5.2.x](https://ci.concourse-ci.org/teams/main/pipelines/release-5.2.x), and [4.2.x](https://ci.concourse-ci.org/teams/main/pipelines/release-4.2.x). Now the principle guiding these "patching" pipelines is that they should be able to continuously exercise any updates to all upstream dependencies. This means that whenever we patch, we're favoring having everything updated over minimizing the size of the patch -- so if we're keeping the master pipeline in a condition where it can realistically test those updates, it makes things a bit easier if we can just share the configuration (tasks, scripts and images) that run those tests.

Furthermore, the Core/API and Runtime tracks both desire long-lived branches (such as the new algorithm in core, or ephemeral check containers in runtime) against which they will run the full suite of integration tests and deploy to a long-lived environment, measuring their long-term metrics.

With this level of reuse across so many pipelines, and with a mandate for "testing the latest whenever possible", a shared library of CI config seems like the best choice. Now we will try to parametrize these configs to allow for each release/branch's specific needs, but if necessary we're hoping to provide a means of [overriding the shared configs](https://github.com/orgs/concourse/projects/36#card-22307810) whenever necessary.

Issue #3939, opened 2019-05-29
"discovering new versions of ..." in build preparation should show 'spinning' while we're waiting on it


vito:
## What challenge are you facing?

Manually triggered builds will not run until all of their input resources have had a `check` perform after the build's creation timestamp. If one of the `check`s is stuck, I currently have no way of knowing besides manually going and checking all the resources.

## What would make this better?

We already have an entry in the build preparation that says 'discovering new versions of ...' but it looks like this is just always checked. Can we have it show a spinner until the resource's 'checked' timestamp is after the build's creation timestamp?

xtremerui:
i think this is exact the same as #3904

vito:
@xtremerui yep, i'll close the other one since this one's in flight

xtremerui:
Just a quick update for fast reproduce the issue. One can set up the pipeline and just disconnect from Internet to make the resource check hanging.

PR #3970, opened 2019-06-04
atc: more accurate build preparation

xtremerui:
for #3939

### DIFF
--- a/atc/db/build_preparation.go
+++ b/atc/db/build_preparation.go
@@ -15,7 +15,7 @@ type MissingInputReasons map[string]string
 const (
 	NoVersionsSatisfiedPassedConstraints string = "no versions satisfy passed constraints"
 	NoVersionsAvailable                  string = "no versions available"
-	NoResourceCheckFinished              string = "no resource check finished after build triggered"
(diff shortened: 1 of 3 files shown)


atc/db/build_preparation.go
```diff
@@ -15,6 +15,7 @@ type MissingInputReasons map[string]string
 const (
 	NoVersionsSatisfiedPassedConstraints string = "no versions satisfy passed constraints"
 	NoVersionsAvailable                  string = "no versions available"
+	NoResourceCheckFinished              string = "no resource check finished after build triggered"
```
vito: Should we change this to 'checking for latest available versions'?

Right now it kind of reads like an error state, but in practice we expect this to happen briefly for every manually triggered build, so it might be best to just make it look like a pending/loading/waiting condition and let the user decide if they've been waiting too long. wdyt?

```suggestion
	NoResourceCheckFinished              string = "checking for latest available versions"
```

vito: (asked in person :+1: - will merge my own suggestion)

PR #3977, opened 2019-06-04
ci: moving the bbr job to the atc

YoussB:
- the reason behind that is that the bosh link for the posrtgres db
(`concourse_db`) is provided by the web job and not the db job.
cc: @cirocosta


cirocosta:
🤔

I thought BOSH links were not supposed to be dependent on which instance group you put the job that gathers the link from is - is it not working with the current config?

cirocosta:
Hey,

We had few other BBR-related PRs this week that already covers the functionality, so I'm closing this one.

Thanks!

PR #3980, opened 2019-06-05
web: Discover cluster name by API call instead of template injection

### LINKED ISSUE #3909: web: use info endpoint to determine cluster name
With #3862 we basically validated the usefulness of exposing the cluster name via the API, so let's make that the only way to consume it, instead of this templating hackery:
https://github.com/concourse/concourse/blob/master/web/indexhandler/handler.go#L65-L69
https://github.com/concourse/concourse/blob/master/web/public/index.html#L87

agurney:
When the web UI displayed the cluster name before, it did so because of server-side templating. Instead, we can now retrieve it dynamically from the `/info` API call, storing it in the session. This reduces the necessary plumbing around propagating the value, and we lose a lot of `clusterName = ""` test boilerplate.

The `FetchVersion` effect, which invoked the `/info` path, is now called `FetchClusterInfo` because we are interested in more of that API's data; some more things are renamed in order to match.

Fixes #3909.

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -48,7 +48,6 @@ import (
 	"github.com/concourse/concourse/skymarshal/skycmd"
 	"github.com/concourse/concourse/skymarshal/storage"
 	"github.com/concourse/concourse/web"
-	"github.com/concourse/concourse/web/indexhandler"
 	"github.com/concourse/flag"
 	"github.com/concourse/retryhttp"
 	"github.com/cppforlife/go-semi-semantic/version"
@@ -611,7 +610,6 @@ func (cmd *RunCommand) constructAPIMembers(
 		return nil, err
 	}
 
-	indexhandler.ClusterName = cmd.Server.ClusterName
 	webHandler, err := webHandler(logger)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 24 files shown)


agurney:
Great point on that test. As you deduced, this is my first time with Elm and it is very much "bluff your way in Elm" as I tried to figure out language concepts (and this codebase) as I went along. I did have a pretty good time of it, thanks mainly to lots of help from the type checker giving _ridiculously good_ messages, and the code being clearly structured already. Thanks for explaining a bit more about how that particular test operates; I'll see if I can figure out a reasonable end-to-end replacement.

agurney:
@pivotal-jamie-klassen Sorry for the delay; I just pushed changes to address the end-to-end testing as you suggested (and rebased against current master). I removed the test in atccmd/ but added a new wats test, https://github.com/concourse/concourse/pull/3980/files#diff-950c08aada373fcb60d0cdf792551ff7R104, which does show the configured cluster name making it through to the rendered web page.

So the test coverage now demonstrates (1) cluster name is exposed on the info API, (2) the web UI asks for this info, (3) if we mock the info, whatever comes back makes it into the top bar, and now (4) end-to-end, if you spin up a cluster with `CONCOURSE_CLUSTER_NAME: dev` in the environment (per dockerfile), and log in, "dev" is indeed what you see.

Disclaimer: I am bluffing my way in wats even more than I am bluffing my way in general.

pivotal-jamie-klassen:
Great coverage, and wats test looks fine. I'm willing to take the bet that if we change the layout/tree structure of the top bar, we would expect to have to change this test too 😄

Issue #3992, opened 2019-06-07
Parallel Input Streaming


jama22:

kcmannem:
@cirocosta pointed out from the difficulties incurred by the rabbitmq team that streaming artifacts to a scheduled task takes a very long time. A large amount of inputs could lead to slow startup times, making it look like concourse scheduling is slow or lagging. We've addressed this in two ways; Gzip is a CPU heavy and also slower at compressing/decompressing resources for transfer so we switched it over to zstd #3880 and this where we will request resources to be streamed in all at once. This should mean the startup time is now limited to the time it takes for the largest resource being streamed..

ddadlani:
We could rate limit the number of parallel streams and expose that number as a command-line flag for configuration. We may need to test on prod/wings and have a benchmark to see the actual benefit that this provides.

We shouldn't merge this before #3819 which will reduce the cost of streaming, but we should benchmark both of these changes separately.

kcmannem:
running drills tests where we stream 6 heavy docker images to a task. Heres the findings...

| In Serial  | In Parallel |
| ------------- | ------------- |
| 7m 30s  | 3m 45s  |
| 7m 17s | 1m 43s  | 
| 4m 43s | 2m 34s  | 
| 7m 13s | 3m 34s  | 

Note:
-  I think we got really good volume locality on the second run so ignore this

Here are the network/cpu graphs of serial streams vs parallel streams, each hump relates to a build being run in the order of the table above.
![Screen Shot 2019-06-18 at 12 42 01 PM](https://user-images.githubusercontent.com/9054726/59702782-84dae180-91c6-11e9-9417-8f75db19bafc.png)
![Screen Shot 2019-06-18 at 12 43 41 PM](https://user-images.githubusercontent.com/9054726/59702881-bce22480-91c6-11e9-906c-b36ef35689d2.png)

![Screen Shot 2019-06-18 at 12 40 14 PM](https://user-images.githubusercontent.com/9054726/59702789-886e6880-91c6-11e9-9814-9996e3e04bc9.png)
![Screen Shot 2019-06-18 at 12 43 48 PM](https://user-images.githubusercontent.com/9054726/59702887-c10e4200-91c6-11e9-8bd3-b816b1390094.png)

cc @ddadlani

cirocosta:
Thanks for sharing! Do you have web CPU usage as well? Thanks!

kcmannem:
@cirocosta good thought, it does in fact consume cpu 
[...]

Issue #3928, opened 2019-05-27
Resource checking should be resilient to a 'stuck' worker


vito:
## What challenge are you facing?

[Prod](https://ci.concourse-ci.org) had its `concourse` resource stop checking for over four days. Investigating the stack dump revealed that a lock was held open and the goroutine holding it was stuck on this line:

https://github.com/concourse/concourse/blob/8d6219d0e1d82ef897838d229dbe4394dd31ecef/atc/resource/run_script.go#L136

```
goroutine 1021 [chan receive, 5834 minutes]:
github.com/concourse/concourse/atc/resource.(*resource).runScript(0xc039940220, 0x2607d20, 0xc01888ae40, 0x1ced6e0, 0x13, 0x0, 0x0, 0x0, 0x1b073c0, 0xc03959f700, ...)
	/tmp/build/70f2e240/concourse/atc/resource/run_script.go:136 +0x724
github.com/concourse/concourse/atc/resource.(*resource).Check(0xc039940220, 0x2607d20, 0xc01888ae40, 0xc0043bc480, 0xc004e71710, 0xc03959f6f0, 0x38384c8, 0x25e54e0, 0xc02c816900, 0x1cd7271)
	/tmp/build/70f2e240/concourse/atc/resource/resource_check.go:17 +0x11c
github.com/concourse/concourse/atc/radar.(*resourceScanner).check(0xc000e2cc60, 0x2627140, 0xc018487920, 0x2656480, 0xc046598f20, 0x26347a0, 0xc035b89f20, 0xc004e71710, 0xc01100c000, 0x8, ...)
	/tmp/build/70f2e240/concourse/atc/radar/resource_scanner.go:360 +0xbeb
github.com/concourse/concourse/atc/radar.(*resourceScanner).scan(0xc000e2cc60, 0x2627140, 0xc018487920, 0x765, 0x0, 0x0, 0x0, 0x0, 0x0)
	/tmp/build/70f2e240/concourse/atc/radar/resource_scanner.go:251 +0x7d6
github.com/concourse/concourse/atc/radar.(*resourceScanner).Run(0xc000e2cc60, 0x2627140, 0xc000e8dec0, 0x765, 0xdf8475801, 0x25c7ce0, 0xc00022e940)
	/tmp/build/70f2e240/concourse/atc/radar/resource_scanner.go:63 +0x97
[...]

PR #3993, opened 2019-06-07
adding parallel input streaming

kcmannem:

### DIFF
--- a/atc/worker/worker.go
+++ b/atc/worker/worker.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/concourse/baggageclaim"
 	"github.com/concourse/concourse/atc/metric"
+	"golang.org/x/sync/errgroup"
 
 	"code.cloudfoundry.org/garden"
 	"code.cloudfoundry.org/lager"
@@ -373,6 +374,21 @@ func (worker *gardenWorker) fetchImageForContainer(
 	return image.FetchForContainer(ctx, logger, creatingContainer)
 }
 
+type mountableLocalInput struct {
+	desiredCOWParent Volume
+	desiredMountPath string
+}
+
+type mountableRemoteInput struct {
+	desiredArtifact  ArtifactSource
+	desiredMountPath string
+}
+
+// creates volumes required to run any step:
+// * scratch
+// * working dir
+// * input
+// * output
 func (worker *gardenWorker) createVolumes(
 	logger lager.Logger,
 	isPrivileged bool,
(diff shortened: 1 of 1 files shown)


kcmannem:
@vito I want to rate limit the parallel streaming somehow (if we don't we'll risk killing network bandwidth on the atc) but don't want to add a user configurable maxinflight. Should I just ratelimit to 5 by default? (no merits on 5, it just feels right but it really depends on how big the resources are.)

atc/worker/worker.go
```diff
@@ -463,6 +470,92 @@ func (worker *gardenWorker) createVolumes(logger lager.Logger, isPrivileged bool
...
+func (worker *gardenWorker) cloneRemoteVolumes(
+	logger lager.Logger,
+	teamID int,
+	privileged bool,
+	container db.CreatingContainer,
+	nonLocals []mountableRemoteInput,
+) ([]VolumeMount, error) {
+	mounts := make([]VolumeMount, len(nonLocals))
+
+	var g errgroup.Group
```
ddadlani: Can we pass the context from `FindOrCreateContainer` down to this level so that we can create a group with context? E.g. ` g, ctx := errgroup.WithContext(ctx)`. This would allow us to cancel streaming if `FindOrCreateContainer` is cancelled.

atc/worker/worker.go
```diff
@@ -371,64 +387,55 @@ func (worker *gardenWorker) createVolumes(logger lager.Logger, isPrivileged bool
 
 	inputDestinationPaths := make(map[string]bool)
 
-	for _, inputSource := range spec.Inputs {
-		var inputVolume Volume
+	localInputs := make([]mountableLocalInput, 5)
+	nonlocalInputs := make([]mountableRemoteInput, 5)
```
ddadlani: Is there a reason to have a limit of 5 local/non-local inputs? for large tasks with lots of `get` steps before it, we could go over 5 inputs.

kcmannem: nope but it needs a intial value or nothing at all. Everytime `append` sees a full array, it reinitializes it so I was just making a good guess that most pipelines will have less than 5 inputs. Avoiding atleast 1 realloc when append is called for the first time.

atc/worker/worker.go
kcmannem: it looks like we don't pass contexts to FindOrCreateVolumes, and the context that errgroup.WithContext returns should be used by funcs that we want to cancel anytime the parent context cancels. Even StreamTo doesn't take in a context. I might be misunderstanding how this works but I don't think anything will be canceled unless we use the returned ctx.

ddadlani: I added a comment about using a context to #3928 since we should probably do that for both container and volume creation. Let's not do it for this issue.

kcmannem:
finished changes, tests passing.

vito:
The changeset looks good but I'm a little cautious to merge it in until I can see some metrics showing an improvement here. I brought this up in IPM but don't really remember where we landed. :sweat_smile:

I remember us saying how currently the bottleneck is with `gzip` compression - i.e. CPU bound, not network bound. So we're hoping to switch to `zstd` to lighten up the CPU load and make the bottleneck be network throughput, right? And the idea is that streaming in parallel allows us to get a lot more bits from A to B more quickly?

That last part is something I'd like to validate with metrics.

re: max-in-flight, would you want a global max-in-flight for parallel input streaming on the `web` node? If you limit it to 5 per-build or even per-step I don't think it'll help much once you have a bunch of builds running. It could still really easily saturate the network. :thinking: I guess the same is true for today, just not so extreme.

kcmannem:
@vito yeah this should go in after zstd so that we use less cpu. And currently we use very little of the bandwidth available to us. The idea is that the initialization time for a task has an upper bound of streaming the largest resource ( resource_a + resource_b vs max(resource_a, resource_b) ). We currently have drills running on this changes so I'll report back with the notable changes. We're streaming heavy docker images to one task as the test. What's interesting is that even in parallel we never spike in network usage (super strange) something is acting as limiter so we don't overload the network. Maybe how we flush the tcp buffers or maybe the fact that compressing is acting like a pseudo queue. From that data its hard to say if we really need a rate limiter of sorts.

kcmannem:
@vito the metrics are here, improvements look good and I don't think we're overloading the network.
https://github.com/concourse/concourse/issues/3992#issuecomment-503217341

Since we have a 10vm drills env already set up, why dont we just run strabo and see how bad the network gets?

kcmannem:
good improvements seen with a resource heavy pipeline. https://github.com/concourse/concourse/issues/3992#issuecomment-504036411

cirocosta:
Hey,

I just re-triggered the tests (were erroring due to the problems with our cred mgmt setup); is there anything else pending here?

Thank you!!

kcmannem:
@ddadlani ready for you to merge after tests pass

Issue #3788, opened 2019-04-24
Resource check queue


pivotal-jwinters:
Instead of checking resources synchronously we want to add them to a queue (table in the db). This way we get better visibility into what resource checks are happening, better de-duping of concurrent checks and allow for rate limiting.

pivotal-jwinters:
We've been investigating a two component approach: `scanner` and `checker`

### scanner
The scanner is responsible for determining if a new `check` should be added to the database. It should be concerned with things like paused jobs/pipelines, pinned resources, disabled resources, check interval, etc... 

### checker
The checker is responsible for running any checks (and type checks) that get added to the database. The check will scan the table and grab a lock to run the check. If it can't get the lock it will simply retry next time it runs. The checker should run regardless of any preconditions that the scanner might care about.

pivotal-jwinters:
One thing we need to keep in mind is the fact that the `resource_checks` table could get quite large. So we'll need to figure out how we're going to gc things.

vito:
@pivotal-jwinters Should they just be marked as completed after the check is done and/or removed?

Or, could we track the queue time and reap anything queued before the `last_checked` value?

pivotal-jwinters:
While working through this issue we've hit a bit of a wrinkle.

Ever since we introduced global resources, radar's behaviour has been a bit strange. Basically every `resource` still has its own check interval, but if two resources share the same `scope`, then only one will actually get the lock and run the check. This makes sense in the current context of all the things we need to run a check, however moving forward we want to remodel things a bit. 

In the context of our two component system (outlined above), the `scanner` will still interact at the `resource` level, but once we get to the `checker` we will have no more knowledge of which resource the check actually ties back to. We started thinking about how we could run a check with the actual resource:

> Since the `scope` doesn't actually know the `atc.Source` to run the check against, we'll need to store it in the `checks` table. This starts to look a LOT like a build `plan` for checks, so why not actually make it such.

This lead us down the path of [refactoring the exec package](https://github.com/concourse/concourse/pull/4000) so that we can add a `exec.CheckStep`. We are going to have to change a few more things along the way, but it seems like this is a pretty good path forward.

vito:
Notes from IPM regarding the handling of missing versions for parent resource types:

* Let's have `fly check-resource` fail fast if the parent resource type has no version.
* Let's add `fly check-resource --recursive` which will recurse into the parent types and perform checks for them, and then check the specified resource.
[...]

PR #4000, opened 2019-06-11
Refactor exec steps (get, put, task)

pivotal-jwinters:
Piling onto our existing PR, we also removed the use of creds types in all the lower levels of the image/resource/exec packages. This way interpolation only ever happens in the exec steps.


atc/db/resource.go
```diff
@@ -296,6 +297,35 @@ func (r *resource) SaveUncheckedVersion(version atc.Version, metadata ResourceCo
...
+	if err != nil {
+		return false, err
+	}
+
+	metadataJSON, err := json.Marshal(metadata)
+	if err != nil {
+		return false, err
+	}
+
+	_, err = psql.Update("resource_config_versions").
```
vito: :thought_balloon: this seems a little iffy/imprecise but it's probably fine.

just thinking about the case where the pipeline is re-configured while the build is in-flight, with a different resource config, which happens to produce the same version identifiers, but runs with a different type or something such that the metadata is not accurate to the new resource type

the likelihood and impact of this kind of edge case is so low that it doesn't seem like something to worry about

vito: would matching on `version_md5` be faster here?

cirocosta:
Hey @pivotal-jwinters ,

What's the rationale behind this PR? It seems like it contains some pretty significant changes, but I can't quickly tell from an outsider perspective 😬 

Thanks!!

pivotal-jwinters:
@cirocosta I added some more info [here](https://github.com/concourse/concourse/issues/3788#issuecomment-503569022)

atc/db/resource.go
pivotal-jwinters: Yeah good call on the `version_md5`. Updated.

atc/db/resource_config_factory.go
```diff
@@ -109,22 +108,17 @@ func (f *resourceConfigFactory) FindOrCreateResourceConfig(
...
-	resourceTypes creds.VersionedResourceTypes,
+	resourceTypes atc.VersionedResourceTypes,
 ) (ResourceConfigDescriptor, error) {
 	resourceConfigDescriptor := ResourceConfigDescriptor{
 		Source: source,
 	}
 
 	customType, found := resourceTypes.Lookup(resourceTypeName)
 	if found {
-		source, err := customType.Source.Evaluate()
```
vito: One concern I have here is that whereas before we would only evaluate the resource type that ends up being used, now we evaluate *all* of them.

Probably not a huge deal. Most resource types don't need credentials. :man_shrugging:

atc/db/team.go
```diff
@@ -839,12 +839,17 @@ func (t *team) FindCheckContainers(logger lager.Logger, pipelineName string, res
 		return nil, nil, err
 	}
 
+	resourceTypes, err := creds.NewVersionedResourceTypes(variables, versionedResourceTypes).Evaluate()
```
vito: (Yep this is what I was worried about, but again probably not a huge deal.)

atc/exec/get_step.go
```diff
@@ -130,7 +130,10 @@ func (step *GetStep) Run(ctx context.Context, state RunState) error {
 		return err
 	}
 
-	resourceTypes := creds.NewVersionedResourceTypes(variables, step.plan.VersionedResourceTypes)
+	resourceTypes, err := creds.NewVersionedResourceTypes(variables, step.plan.VersionedResourceTypes).Evaluate()
```
vito: (Here too.)

atc/exec/put_step.go
```diff
@@ -94,7 +94,10 @@ func (step *PutStep) Run(ctx context.Context, state RunState) error {
 		return err
 	}
 
-	resourceTypes := creds.NewVersionedResourceTypes(variables, step.plan.VersionedResourceTypes)
+	resourceTypes, err := creds.NewVersionedResourceTypes(variables, step.plan.VersionedResourceTypes).Evaluate()
```
vito: (...and here.)

atc/exec/task_step.go
```diff
@@ -126,7 +125,11 @@ func (step *TaskStep) Run(ctx context.Context, state RunState) error {
 	})
 
 	variables := creds.NewVariables(step.secrets, step.metadata.TeamName, step.metadata.PipelineName)
-	resourceTypes := creds.NewVersionedResourceTypes(variables, step.plan.VersionedResourceTypes)
+
+	resourceTypes, err := creds.NewVersionedResourceTypes(variables, step.plan.VersionedResourceTypes).Evaluate()
```
vito: (...and here.)

atc/exec/task_step.go
```diff
@@ -339,7 +342,7 @@ func (step *TaskStep) imageSpec(logger lager.Logger, repository *artifact.Reposi
 	} else if config.ImageResource != nil {
 		imageSpec.ImageResource = &worker.ImageResource{
 			Type:    config.ImageResource.Type,
-			Source:  creds.NewSource(boshtemplate.StaticVariables{}, config.ImageResource.Source),
```
vito: Ha, guessing this is 'cause it really just never supported interpolation here but the type demanded it. I guess this is a small win.

atc/radar/resource_scanner.go
```diff
@@ -154,10 +154,15 @@ func (scanner *resourceScanner) scan(logger lager.Logger, resourceID int, fromVe
 		return 0, err
 	}
 
-	versionedResourceTypes := creds.NewVersionedResourceTypes(
+	versionedResourceTypes, err := creds.NewVersionedResourceTypes(
 		scanner.variables,
 		resourceTypes.Deserialize(),
-	)
+	).Evaluate()
```
vito: (...and here.)

atc/radar/resource_type_scanner.go
```diff
@@ -115,10 +115,15 @@ func (scanner *resourceTypeScanner) scan(logger lager.Logger, resourceTypeID int
 		return 0, err
 	}
 
-	versionedResourceTypes := creds.NewVersionedResourceTypes(
+	versionedResourceTypes, err := creds.NewVersionedResourceTypes(
 		scanner.variables,
 		resourceTypes.Deserialize(),
-	)
+	).Evaluate()
```
vito: (...and here.)

atc/db/resource_config_factory.go
vito: I think the mitigation here is to just enable caching. That resource type is going to be actually evaluated and used *somewhere*, so it'll have been fetched anyway, and if you're using credential caching it'll only hit the credential manager once regardless of how many places call `.Evaluate` within that minute.

Issue #1309, opened 2017-06-20
Document fly completion


jmcarp:
## A Modest Proposal

Fly now supports tab completion for certain flags, but users will have to set up completion in their shells to make use of the feature. For example, my `.zshrc` includes this:

```
#compdef fly
autoload -U compinit && compinit
autoload -U bashcompinit && bashcompinit
_fly_bash_autocomplete() {
    args=("${COMP_WORDS[@]:1:$COMP_CWORD}")
    # Only split on newlines
    local IFS=$'\n'
    # Call completion (note that the first element of COMP_WORDS is
    # the executable itself)
    COMPREPLY=($(GO_FLAGS_COMPLETION=1 ${COMP_WORDS[0]} "${args[@]}"))
    return 0
}
complete -F _fly_bash_autocomplete fly
```

Some cli libraries can auto-generate these invocations for users, like https://github.com/alecthomas/kingpin#bashzsh-shell-completion, but go-flags doesn't. Or if fly were installed with homebrew, we could use a post-install hook to add completion scripts--but we don't. Seems like the options are to add sample bash/zsh invocations to the docs, or possibly update `sync` or another command to output completion scripts from `fly`.

suhlig:
FYI, I just looked at [`_cf-cli`](https://github.com/cloudfoundry/homebrew-tap/blob/master/cf-cli.rb), and simply re-using the same function works fine. I added `complete -F _cf-cli fly` to my `.bashrc`, and now fly has proper command completion.

Maybe fly / Concourse just needs to document this (pretty generic) function somewhere?

dohsimpson:
Thanks to @suhlig 's comment. Since I don't want to install cf-cli. What I did is the following:

1. Create a file like this

```
#!/bin/bash

# credit to: https://github.com/cloudfoundry/homebrew-tap/blob/master/cf-cli.rb
_fly_compl() {
    # All arguments except the first one
    args=("${COMP_WORDS[@]:1:$COMP_CWORD}")
    # Only split on newlines
    local IFS=$'\n'
    # Call completion (note that the first element of COMP_WORDS is
    # the executable itself)
    COMPREPLY=($(GO_FLAGS_COMPLETION=1 ${COMP_WORDS[0]} "${args[@]}"))
    return 0
}

complete -F _fly_compl fly
```

2. source that file in your bashrc or zshrc.

mattysweeps:
A common pattern among tools is to generate this with an `autocomplete` sub command.

[...]

PR #4012, opened 2019-06-13
Add `fly completion` as a command

deadNightTiger:
As mentioned in #1309, users have to add a snippet to their .bashrc to enable autocompletion. This PR packages the snippet into the `fly` CLI, so the `.bashrc` entry can be reduced to one-line:
```
source <(fly completion bash)
```
Or using the `/etc/bash_completion.d` directory:
```
fly completion bash > /etc/bash_completion.d/fly
```

### DIFF
--- a/fly/commands/completion.go
+++ b/fly/commands/completion.go
@@ -3,11 +3,8 @@ package commands
 import "fmt"
 
 type CompletionCommand struct {
-	Bash BashCompletionCommand `command:"bash"`
-	// TODO: support more shells
+	Shell string `long:"shell" required:"true" choice:"bash"` // add more choices later
 }
-
-type BashCompletionCommand struct {}
 
 // credits:
 // https://godoc.org/github.com/jessevdk/go-flags#hdr-Completion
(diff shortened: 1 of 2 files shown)


fly/commands/completion.go
```diff
@@ -0,0 +1,27 @@
+package commands
+
+import "fmt"
+
+type CompletionCommand struct {
+	Bash BashCompletionCommand `command:"bash"`
```
vito: Is `foo completion bash` a common convention in other CLIs? I only ask because the pattern elsewhere in `fly` is to use flags instead of arguments. So it'd be something like `fly completion --shell bash` instead, with `--shell` being something like this:

```suggestion
	Shell string `long:"shell" required:"true" choice:"bash"` // add more choices later
```

deadNightTiger: `kubectl` and `helm` have `completion bash`. `pip` has `completion --bash`, while `pipenv` has just `--completion` with autodetecting the running shell, and `terraform` does `-install-autocomplete`. I guess the convention is not as common as I thought :)

vito: Thanks for checking. :slightly_smiling_face:

Issue #3808, opened 2019-04-30
Test per-pipeline credential lookup using Kubernetes credential management


cirocosta:
Hey,

@mhuangpivotal  and I noticed that the k8s credential management tests are missing one use case that we support: having pipeline-scoped secrets.

https://github.com/concourse/concourse/blob/0ecb0381931e47ed9ac77091bdf0a92c21c2113b/topgun/k8s/kubernetes_creds_mgmt_test.go#L73

Having this test is important as people are probably already making use of such feature as it's already documented here: https://github.com/helm/charts/tree/c4107a5b7e32dd7040e64f219051eb9a2c7ca743/stable/concourse#kubernetes-secrets

Thanks!

taylorsilva:
# Issue Status
###### Keep editing this comment with the current status of the issue

***

### Task List

- [ ] Manually test that pipeline-scoped secrets work as expected
- [ ] Write tests to cover pipeline-scoped secrets
- [ ] Create a PR and wait for it to be merged

PR #4015, opened 2019-06-13
[topgun/k8s] Add test for per-pipeline k8s credentials

taylorsilva:
Fixes https://github.com/concourse/concourse/issues/3808

CC @pivotal-bin-ju

### DIFF
--- a/topgun/k8s/kubernetes_creds_mgmt_test.go
+++ b/topgun/k8s/kubernetes_creds_mgmt_test.go
@@ -72,14 +72,13 @@ var _ = Describe("Kubernetes credential management", func() {
 		})
 	})
 
-	Context("Consuming per-team k8s secrets", func() {
 
-		JustBeforeEach(func() {
-			// ((foo)) --> bar
-			createCredentialSecret(releaseName, "foo", "main", map[string]string{"value": "bar"})
 
-			// ((caz.baz)) --> zaz
-			createCredentialSecret(releaseName, "caz", "main", map[string]string{"baz": "zaz"})
+	Context("Consuming k8s credentials", func() {
+		var runsBuildWithCredentialsResolved = func(normalSecret string, specialKeySecret string) {
+			By("creating credentials in k8s credential manager")
+			createCredentialSecret(releaseName, normalSecret, "main", map[string]string{"value": "bar"})
+			createCredentialSecret(releaseName, specialKeySecret, "main", map[string]string{"baz": "zaz"})
 
 			fly.Run("set-pipeline", "-n",
 				"-c", "../pipelines/minimal-credential-management.yml",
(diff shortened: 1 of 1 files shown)


topgun/k8s/kubernetes_creds_mgmt_test.go
```diff
@@ -88,48 +87,106 @@ var _ = Describe("Kubernetes credential management", func() {
...
+
+		var disableTeamNamspaces = func() {
+			By("creating a namespace made by the user instead of the chart")
+			Run(nil, "kubectl", "create", "namespace", releaseName+"-main")
+			extraArgs = []string{
+				"--set=concourse.web.kubernetes.createTeamNamespaces=false",
+			}
+		}
+
+		var resetExtraArgs = func() {
```
cirocosta: Hey, do you think that we could get away from this?

I'm afraid that this adds some extra cognitive load for who's reading the tests as now you have to keep track of the fact that `extraArgs` can be reset to zero 🤔  

Wdyt? thanks!

topgun/k8s/kubernetes_creds_mgmt_test.go
```diff
@@ -88,48 +87,106 @@ var _ = Describe("Kubernetes credential management", func() {
...
+			extraArgs = []string{}
+		}
+
+
+		Context("using per-team credentials", func() {
+			normalSecret := "foo"
+			specialKeySecret := "caz"
```
cirocosta: Hey, wdyt of being more explicit in terms of naming here? E.g., what's "special" about it that "normal" doesn't have? 

Thanks!

topgun/k8s/kubernetes_creds_mgmt_test.go
```diff
@@ -88,48 +87,106 @@ var _ = Describe("Kubernetes credential management", func() {
...
 
+			By("seeing the credentials were resolved by concourse")
 			Expect(string(session.Out.Contents())).To(ContainSubstring("bar"))
 			Expect(string(session.Out.Contents())).To(ContainSubstring("zaz"))
 		}
 
-		Context("using the default namespace created by the chart", func() {
-			It("succeeds", runsBuildWithCredentialsResolved)
-		})
+		var runGetsCachedCredentials = func(normalSecret string, specialKeySecret string) {
```
cirocosta: Nice! It's pretty cool how much readable it becomes when looking at the context and specs descriptions once everything comes down to those methods.

I was wondering if we could get this out of the `context` block though, just like `createCredentialSecret` is separated. Wdyt?

thanks!

topgun/k8s/kubernetes_creds_mgmt_test.go
taylorsilva: Naming hard. The only thing different about `specialKeySecret` is that instead of it's key in the kv-map isn't `value` it's `baz`.
We'll think on it a bit, we weren't too happy with it either.

topgun/k8s/kubernetes_creds_mgmt_test.go
taylorsilva: We were able to remove this variable from the test!

topgun/k8s/kubernetes_creds_mgmt_test.go
taylorsilva: Renamed these. wdyt?

topgun/k8s/kubernetes_creds_mgmt_test.go
taylorsilva: Moved, looks better 👍

PR #4018, opened 2019-06-14
fly: remove "better go catch it lol"

gdamjan:
I guess the pun was fun when it was originally added, but seeing it more than once (also when you're angry) is annoying.

Let's just remove the pun and be more serious.

### DIFF
--- a/fly/main.go
+++ b/fly/main.go
@@ -62,7 +62,7 @@ func handleError(helpParser *flags.Parser, err error) {
 			fmt.Fprintln(ui.Stderr, "")
 			fmt.Fprintln(ui.Stderr, "    "+ui.Embolden("%s", netErr))
 			fmt.Fprintln(ui.Stderr, "")
-			fmt.Fprintln(ui.Stderr, "is the targeted Concourse running? better go catch it lol")
+			fmt.Fprintln(ui.Stderr, "is the targeted Concourse running?")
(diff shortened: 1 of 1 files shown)


marco-m:
I disagree with this PR. That pun is part of Concourse and I like it.

gdamjan:
YMMV, but it sounds really unprofessional and condescending to its users (and probably when they least need it)

jtarchie:
I believe that this reflects on a moment of nostalgia for most in the Concourse community. As a member who has been using concourse for a long time, this was subtle and smirking pun. Quick, witty, and unexpected. A moment to reflect, don't take things (too) seriously, you'll get through them.

There's a precedent for applications to have their own humour. This can be attributed to easter eggs, humorous 404 pages, etc. A personal favorite is the Konami code.

Every new user I've paired with in concourse has appreciated the joke. As at some point, for them, it is an error message.

I'm in the position to keep it. I like it.

xtremerui:
Have we settled to an agreement here? Could I close this PR @gdamjan ?

gdamjan:
> Have we settled to an agreement here? Could I close this PR @gdamjan ?

it seems *you* have an agreement, I personally don't find the argument "it's part of concourse" really convincing. But I should've expected this backlash when trying to change in-group jokes.

I wonder if I opened a PR adding jokes if that would be more part of concourse???

PR #4027, opened 2019-06-18
[topgun/bosh] Update tests for bosh links

taylorsilva:
Change the postgres credentials to be non-default. Existing tests will only pass if the bosh job in our release uses the bosh links correctly.

Blocked by https://github.com/concourse/concourse-bosh-release/pull/41

### DIFF
--- a/topgun/operations/bbr-concourse-link.yml
+++ b/topgun/operations/bbr-concourse-link.yml
@@ -1,3 +1,16 @@
+# use non-default credentials to ensure concourse_db bosh links are used
+- type: replace
+  path: /instance_groups/name=web/jobs/name=web/properties/postgresql/role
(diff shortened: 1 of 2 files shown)


cirocosta:
Hey,

If I understood correctly, nothing in the tests need to be changed because we set 0 properties there - it's all coming through bosh links, right?

Thanks!

taylorsilva:
@cirocosta correct! Nothing in the tests needed to change, just the ops-file. If you run the test without this PR https://github.com/concourse/concourse-bosh-release/pull/41 then the test will currently fail.

taylorsilva:
@cirocosta made more changes to the ops files. Latest commit message explains our thinking. let me know what you think.

PR #4028, opened 2019-06-18
only update resource version metadata if the new value is not null

### LINKED ISSUE #4005: resource version metadata is deleted
## Bug Report

Resource version metadata is being removed from the database. 

## Steps to Reproduce

Go to a resource page and expand the latest version. You should see metadata.

Click the `refresh` button to force check resource. Or execute the `fly check-resource` command.

## Expected Results

If there is no newer version nothing should happen

## Actual Results

The metadata of the latest version has been removed.

## Additional Context

I think there is something weird going on like a race between the resource's `in` and `check` scripts. The `check` script (for some reason?) doesn't populate `metadata` in the db.

The `saveResourceVersion` function does an UPSERT and overwrites the metadata each time - presumably deleting it if a `check` runs after an `in`.

https://github.com/concourse/concourse/blob/master/atc/db/resource_config_scope.go#L276

## Version Info

* Concourse version: 5.2

* Did this used to work? don't know

robwhitby:
fixes #4005


vito:
Thanks!

PR #4032, opened 2019-06-19
improve sidebar

### LINKED ISSUE #3872: Sidebar feedback round 2
The new sidebar looks great!! 

Feedback 1: It might help to use a different colour (maybe darker) font for the team names within the sidebar so that at a quick glance it's easy to differentiate between the teams and pipelines? 

Screen Shot 2019-05-15 at 9 33 36 PM

Feedback 2: `j`/`k` keybindings for switching between pipelines would be diggity dope. 

Feedback 3: Drag to reorder would be nice but maybe that should be a separate issue?

Thx!!

@vito and @clarafu

pivotal-jamie-klassen:
fixes #3872

# Release Note

```tex
\note{feature}{
  Improved the appearance of the sidebar (re-introduced in \ghrelease{v5.2.0}) 
  in response to
  \link{user feedback}{https://github.com/concourse/concourse/issues/3872}.
  Teams and pipelines are now easier to distinguish, and there are nicely styled
  tooltips for long items.
}
```


pivotal-jamie-klassen:
[useful comment on the originating issue](https://github.com/concourse/concourse/issues/3872#issuecomment-504141102)

pivotal-jamie-klassen:
@xtreme-james-thomson and I noticed that the tooltips seem to take a long time to appear when viewing a build with long output. It may be worth investigating this further before merging.

pivotal-jamie-klassen:
we investigated and discovered that the slow tooltip behaviour is because of #4061.

PR #4033, opened 2019-06-19
Remove unnecessary logger and lock interval

pivotal-jwinters:


pivotal-jamie-klassen:
At a glance this looks like a strict refactor. Is there any functional impact from this change? @pivotal-jwinters @xtremerui

xtremerui:
@pivotal-jamie-klassen nope

Issue #3932, opened 2019-05-28
Add API for querying versions by version fields


vito:
In order to support e.g. `fly pin-resource-version -r concourse/concourse -v ref:abcde`, we should have an API endpoint for finding version numbers by querying fields of the version. This idea first came about in https://github.com/concourse/concourse/pull/3131#issuecomment-459874740.

This API should implement partial field matching, since some versions contain auxiliary information that is difficult to provide as a user. For example, PR versions look like this:

![image](https://user-images.githubusercontent.com/1880/58494418-0c42b100-8143-11e9-891c-f0074b36e8e3.png)

...so in this case I'd like to be able to just say `-v pr:3931` without having to specify all the other fields.

This kinda relates to https://github.com/concourse/concourse/pull/3920 where we mulled over how to URL-encode versions. I'd like to make the common case look nice, e.g. just `ref:abcdef`. We'll have to deal with funny values in edge cases, and for those I guess we can just do ugly URL encoding.

Rough sketch of common + edge cases:

| version | url |
| - | - |
| `{"ref":"abcdef"}` | `/api/v1/teams/foo/pipelines/bar/resources/baz/versions?filter=ref:abcdef` |
| `{"ref":"some value"}` | `/api/v1/teams/foo/pipelines/bar/resources/baz/versions?filter=ref:some%20value` |
| `{"some ref":"some value"}` | `/api/v1/teams/foo/pipelines/bar/resources/baz/versions?filter=some%20ref:some%20value` |
| `{"ref":"some%value"}` | `/api/v1/teams/foo/pipelines/bar/resources/baz/versions?filter=ref:some%25value` |
| `{"ref":"some%value","another":"field"}` | `/api/v1/teams/foo/pipelines/bar/resources/baz/versions?filter=ref:some%25value&field=another:field` |

@pivotal-jwinters wdyt?

xtremerui:
One thing nice to have is a hint in response when nothing matches the field. Something like
`field not found. Available fields are base_sha, pr, message, url ...` taking pull request resource for example.

pivotal-jamie-klassen:
what about colons (`:`) in the keys or values of the version?

pivotal-jamie-klassen:
| version | url |
| - | - |
| `{"key:with:colon":"abcdef"}` | `/api/v1/teams/foo/pipelines/bar/resources/baz/versions?filter=key%3Awith%4Acolon:abcdef` |

vito:
[...]

PR #4048, opened 2019-06-21
Support querying resource versions by fields

xtremerui:
refer to https://github.com/concourse/concourse/issues/3932

a related side work is fixing `fly curl` with query params (since there is no concourse api needed passing query until this one) as before there was no way to pass query and if one using `fly curl /api/v1/path?param=value`, it will encode the `?` and produces a wrong url cause `404 not found`

### DIFF
--- a/atc/api/resourceserver/versionserver/list.go
+++ b/atc/api/resourceserver/versionserver/list.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc"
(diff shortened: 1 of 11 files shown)


atc/db/resource.go
```diff
@@ -384,6 +384,14 @@ func (r *resource) Versions(page Page) ([]atc.ResourceVersion, Pagination, bool,
 		WHERE r.id = $1 AND r.resource_config_scope_id = v.resource_config_scope_id AND v.check_order != 0
 	`
 
+	if len(versionFilter) != 0 {
+		filterJSON, err := json.Marshal(versionFilter)
+		if err != nil {
+			return nil, Pagination{}, false, err
+		}
+		query += fmt.Sprintf(` AND version @> '%s'`, string(filterJSON))
```
vito: We should use a param here to be safe. Someone could submit JSON with a `'` in it.

xtremerui: Ah thx for pointing it out!  Updated to use sql query param.

atc/api/resourceserver/versionserver/list.go
```diff
@@ -25,6 +26,20 @@ func (s *Server) ListResourceVersions(pipeline db.Pipeline) http.Handler {
...
+			logger.Error("failed-to-parse-request-form", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		fields := r.Form["filter"]
+		versionFilter := make(atc.Version)
+		for _, field := range fields {
+			vs := strings.SplitN(field, ":", 2)
+			versionFilter[vs[0]] = vs[1]
```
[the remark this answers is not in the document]
xtremerui: Done! Thx!

PR #4057, opened 2019-06-24
[MVP] max task containers

aledeganopix4d:
This is, of course, not final. But is the bare-barest minimum needed to implement the feature.
I'd like to share it with the team to have an idea of what do you guys think of the approach.

There are couple of values that are hardcoded, and some very rough logging, but beside these the code is working and doing what is supposed to.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 .idea
+.vscode
 *.iml
 node_modules
 *-packr.go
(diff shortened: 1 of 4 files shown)


cirocosta:
Hi, @aledeganopix4d !

Are you still looking for feedback on this one?

If so, what I'd recommend is keeping it as `draft` PR open and then ask for comments from other folks 😁 

Thanks!

aledeganopix4d:
Hi, @cirocosta,
thanks for your suggestion, but no, it was meant to have feedback from my team by opening against our fork, sorry for the noise :)

Issue #3803, opened 2019-04-29
File perms are not preserved in `fly execute`


Callisto13:
## Bug Report

Locally `execute`d builds often fail with:
```
Backend error: Exit status: 500, message: {"Type":"","Message":"runc exec: exit status 1: exec failed: container_linux.go:345: starting container process caused \"exec: \\\"cbug/dummy\\\": permission denied\"\n","Handle":"","ProcessID":"","Binary":""}
```

On my machine:
```
 cbug > ll ci
total 16
drwxr-xr-x  4 pivotal  staff  128 Apr 29 16:15 .
drwxr-xr-x  4 pivotal  staff  128 Apr 29 16:15 ..
-rwxr-xr-x  1 pivotal  staff   21 Apr 29 15:58 dummy
-rw-r--r--  1 pivotal  staff  173 Apr 29 16:15 dummy.yml
```

Inspection of the bind-mounted volume after a privileged build shows that permissions and ownership were not preserved\* and therefore the `run.path` executable could not be called.

On the worker both uploaded files have `0666` perms and are owned by `root`:
```
worker/28fe13ea-1333-497c-ac89-b6cb2eb51ab2:/var/vcap/data/worker/work/volumes# ll ./live/f6301c44-0ae1-479d-6eeb-4fb1d544aa1f/volume/ci/
total 8
drwxrwxrwx 1 root root  28 Apr 29 15:16 ./
drwxrwxrwx 1 root root  24 Apr 29 15:16 ../
-rw-rw-rw- 1 root root  21 Apr 29 14:58 dummy
-rw-rw-rw- 1 root root 173 Apr 29 15:15 dummy.yml
```

The parent directory, `ci`, is `0777`:
```
worker/28fe13ea-1333-497c-ac89-b6cb2eb51ab2:/var/vcap/data/worker/work/volumes# ll ./live/f6301c44-0ae1-479d-6eeb-4fb1d544aa1f/volume/
total 4
drwxrwxrwx 1 root root  24 Apr 29 15:16 ./
drwxr-xr-x 1 root root  72 Apr 29 15:16 ../
drwxrwxrwx 1 root root  28 Apr 29 15:16 ci/
```

While for a successful run, the Volume has been laid down with `0755` and the correct owner:

```
worker/28fe13ea-1333-497c-ac89-b6cb2eb51ab2:/var/vcap/data/worker/work/volumes# ll live/c6477ce5-a3f6-48a0-5dfa-c8bea76123ec/volume/ci
[...]

PR #4066, opened 2019-06-26
fix file permission when uploading local input during fly execute

xtremerui:
concourse/concourse#3803

when uploading local inputs for fly execute it will send "platform" of task config to create artifact endpoint. Then it will include platform info in worker spec for choosing a worker for artifact volume initializing.


xtremerui:
@pivotal-jwinters I rebased master and resolved the conflict. Should be good to merge now. Thx!

Issue #2928, opened 2018-12-10
Add ATC/worker flags to limit max build containers for workers


mhuangpivotal:
**Note: we want to refactor runtime code in #2926 before implementing this.**

With the `least-build-containers` container placement strategy implemented in #2577, we also want a way to limit the max number of build containers on the workers.

The proposal in #2577 wants to add `--max-tasks-per-worker` to ATC:

> if `max-tasks-per-worker` is set:
> 
> * If the current worker has less than `max-tasks-per-worker` tasks running, then dispatch to it the new task (and directly or indirectly update the priority queue).
> * If on the other hand the current worker has `>=` than `max-tasks-per-worker` tasks running, then since the list is a priority queue, all other workers will be in a similar situation, no need to keep traversing the list. Do not dispatch the runnable task and wait for the next event to wake up the scheduler.
> 
> If `max-tasks-per-worker` is not set:
> 
> * Take the current worker from the priority queue and dispatch the runnable task to it. This is still better than random placement because it dispatches to the matching worker will less running tasks.
>

This option requires `least-build-containers` to be set, otherwise it will error. 

Additionally,

> Add option `max-tasks-for-worker` to `concourse worker` and modify the scheduling algorithm on the ATC to allow to take care of differences between workers, while `max-tasks-per-worker` on the ATC would still be a maximum cap.

We may want to change the flag names to say `build-containers` instead of `tasks`, since the `least-build-containers` strategy includes `get`, `task` and `put` containers.

ddadlani:
@marco-m what exactly is the use case for having a varying maximum for build containers on each worker? We don't currently set container maximums, those are enforced directly by Garden. Would the fix in #3251 along with the `fewest-build-containers` strategy not satisfy your use case?

marco-m:
@ddadlani here is my understanding:

The fix in #3251, along with `fewest-build-containers`, would allow to spread evenly the task containers on the available workers. This would make `fewest-build-containers` usable in prod for us and would be a great improvement :-)

On the other hand, I think that this ticket, that stems from #2577, is about controlling the load. More in details: it would allow to control the max number of task containers on a given worker. For example: as an operator, if I know that more than, say, 2 task containers kill my workers, then I can set max-tasks-per-worker to 2. If the total number of runnable tasks is more than number_of_workers * max-tasks-per-worker, then the Concourse scheduler will not dispatch any task and wait for the next scheduler run / next event. This would provide a rough queue.

Does it makes sense ?

[...]

PR #4069, opened 2019-06-26
[POC] Limit tasks per worker.

aledeganopix4d:
This is a POC to attempt fixing #2928.
I have started by using the number of "Build containers" in a worker to evaluate if the worker can accept more tasks. However the lifetime of a container is longer than the lifetime of the task: this is especially evident in cases where a task fails or gets aborted.

To circumvent the above limitation I have added a column to the containers DB table: "running_task" that the task step sets to true when starts and to false when it completes. This allow to schedule tasks on workers that, effectively, are not executing a task.

As a side effect the strategy `fewest-build-containers` is more accurate, in the sense that now it counts the build containers on a worker that are active AND running a task.

With this draft PR I would like some feedback to understand if I am going in an acceptable direction, of course there are still many refinements to do before this could be a real PR, including but not limited to:

- Parametrize the maximum number of active tasks per worker as an option
- Set a limit on the amount of time the task can wait for a free worker
- Take into account the new column in the database migration
- Remove silly debug messages
- Add tests
- Tackle the corner case where tasks scheduled close to one another can still
end up on the same worker? (e.g. a job with in_parallel tasks).

What do you think?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,3 @@
-.DS_Store
-.idea
-*.iml
 node_modules
 *-packr.go
 tags
@@ -13,3 +10,9 @@ ci/deployments/*/.terraform/
 ci/deployments/*/keys/gcp.json
 ci/deployments/*/keys/id_rsa
 ci/deployments/*/keys/id_rsa.pub
+
+# Editors
+.DS_Store
+.idea
+.vscode
+*.iml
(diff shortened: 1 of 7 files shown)


aledeganopix4d:
I'm working on a different approach for the same issue.

PR #4086, opened 2019-07-02
Aligned long flag of get-team with other

cappyzawa:
long flag for the team name in `get-team` was different from others.

e.g. 
`set-team`'s long flag (`team-name`)
https://github.com/concourse/concourse/blob/c16d62872e4ae2055efe2a27cf4a9903d9d009f5/fly/commands/set_team.go#L27

`login`'s long flag (`team-name`)
https://github.com/concourse/concourse/blob/c16d62872e4ae2055efe2a27cf4a9903d9d009f5/fly/commands/login.go#L29

### DIFF
--- a/fly/commands/get_team.go
+++ b/fly/commands/get_team.go
@@ -14,8 +14,8 @@ import (
 )
 
 type GetTeamCommand struct {
-	Team string `short:"n" long:"team" required:"true" description:"Get configuration of this team"`
-	JSON bool   `short:"j" long:"json" description:"Print command result as JSON"`
+	TeamName string `short:"n" long:"team-name" required:"true" description:"Get configuration of this team"`
+	JSON     bool   `short:"j" long:"json" description:"Print command result as JSON"`
 }
 
 func (command *GetTeamCommand) Execute(args []string) error {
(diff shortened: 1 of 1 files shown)


xtremerui:
@vito while I like this consistence personally, it might potentially break folks script if they are using set-team with long flag. Not sure if it is totally worth in this case. Wdyt? Thx!

cappyzawa:
> it might potentially break folks script if they are using set-team with long flag. Not sure if it is totally worth in this case. 

It might be too much to ask but I think it's not `set-team` but `get-team`.
If it is `get-team`, I also agree, so I want to hear @vito's opinion.

cappyzawa:
@xtremerui @vito

I just noticed `builds` command's `-n` is also `team`...
```
$ fly builds -h
error: Usage:
  fly [OPTIONS] builds [builds-OPTIONS]

Application Options:
  -t, --target=                 Concourse target name
  -v, --version                 Print the version of Fly and exit
      --verbose                 Print API requests and responses
      --print-table-headers     Print table headers even for redirected output

Help Options:
  -h, --help                    Show this help message

[builds command options]
      -a, --all-teams           Show builds for the all teams that user has access to
      -c, --count=              Number of builds you want to limit the return to (default: 50)
          --current-team        Show builds for the currently targeted team
      -j, --job=PIPELINE/JOB    Name of a job to get builds for
          --json                Print command result as JSON
      -p, --pipeline=           Name of a pipeline to get builds for
      -n, --team=               Show builds for these teams
          --since=              Start of the range to filter builds
          --until=              End of the range to filter builds
```

Should I change this command's `--team` to `--team-name`?

cappyzawa:
@vito Thanks!!

PR #4087, opened 2019-07-02
Add worker and team completions

cappyzawa:
I added worker(`-w`) and team(`-n`) completions.

worker:
* `prune-worker -w`
* `land-worker -w`

team:
* `get-team -n`
* `set-team -n`
* `destroy-team -n`
* `rename-team -o`

### DIFF
--- a/fly/commands/destroy_team.go
+++ b/fly/commands/destroy_team.go
@@ -13,7 +13,7 @@ import (
 )
 
 type DestroyTeamCommand struct {
-	TeamName        flaghelpers.TeamFlag `short:"n" long:"team-name" required:"true"        description:"The team to delete"`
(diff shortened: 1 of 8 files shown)


fly/commands/destroy_team.go
```diff
@@ -27,7 +28,7 @@ func (command *DestroyTeamCommand) Execute([]string) error {
 		return err
 	}
 
-	teamName := command.TeamName
+	teamName := string(command.TeamName)
```
vito: a minor nitpick - I think a method on `flaghelpers.TeamFlag` would be safer than casting here, in case the composition of that type changes from a simples tring later. Here's one example of this pattern: https://github.com/concourse/flag/blob/master/directory.go#L29

cappyzawa: Thanks for your review!
I fixed it 👍

PR #4090, opened 2019-07-02
Update topgun/k8s test to use new hosts parameter

taylorsilva:
Changed topgun/k8s to use the new `hosts` parameter

Blocked by https://github.com/helm/charts/pull/15220

cc @zoetian


cirocosta:
Hey, as a follow up, `hush-house` will need to be updated too: https://github.com/concourse/hush-house/blob/eb4dbf1cdd79774fc23136e504dc6319946d8ddc/deployments/with-creds/worker/values.yaml#L51

PR #4092, opened 2019-07-03
adds '--all' to the pause and unpause commands

joshzarrabi:
- will pause and unpause all pipelines

before making changes to my concourse I like to pause every pipeline so I don't get stuck in a state where a worker starts to drain, and the change can happen more quickly. I wrote some hacky bash to make this happen but I thought it would be nice to add it into fly as a first class feature.

Signed-off-by: Josh Zarrabi <jzarrabi@pivotal.io>

### DIFF
--- a/fly/commands/pause_pipeline.go
+++ b/fly/commands/pause_pipeline.go
@@ -9,7 +9,8 @@ import (
 )
 
 type PausePipelineCommand struct {
-	Pipeline flaghelpers.PipelineFlag `short:"p"  long:"pipeline" required:"true" description:"Pipeline to pause"`
+	Pipeline flaghelpers.PipelineFlag `short:"p"  long:"pipeline" description:"Pipeline to pause"`
+	All      bool                     `short:"a"  long:"all"      description:"Pause all pipelines"`
 }
 
 func (command *PausePipelineCommand) Validate() error {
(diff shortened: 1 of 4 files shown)


xtremerui:
@jama22

deniseyu:
It looks like we have precedent for `--all` flags (`delete-target`, `prune-workers` for all stalled workers) so I think this is probably a reasonable extension for pause/unpause! Unless @jama22 or @vito disagrees!

joshzarrabi:
Hmm @clarafu I don't see a merge conflict? Maybe I'm missing it though.

deniseyu:
@joshzarrabi it's probably fine to skip rebasing; you're right that this should merge cleanly since these files don't change too often.
An edge case we saw elsewhere with `--all` flags was when a user runs the command with `--all` but no target specified, e.g. `fly pause-pipeline --all`. The least surprising behavior in that case is that `fly` should happily pause all the pipelines, rather than complaining about the absence of a target.
Would you mind adding integration tests for this edge case in both the pause and unpause files?
Sorry it's taken us a bit to get back to you on this!

joshzarrabi:
Hi, I'm actually a little bit confused about the ask. How should `fly` know which concourse to target if there is no `-t target` flag passed? Or am I reading the request wrong?

vito:
I would expect it to just fail if `-t` isn't specified. (Almost all commands do/should.)

joshzarrabi:
Okay should be ready! Thanks @vito

Issue #4035, opened 2019-06-19
Ability to enumerate active users


zoetian:
## Why we want to enhance this feature?
There's no accurate measure on the number of active users on a concourse cluster.

We proposed two ways of doing this:
1. Parsing Log. As @pivotal-jwinters suggested, we could filter out the event `atc.dex.event`
 from the log for the past 60 days.
2. New DB Table. @YoussB proposed to add a new database table `active_user` in current DB to update the user `last_login_date` to enumerate the active users. This would require additional changes to the current login flow. The table schema may look like below:

| username      | last_login_date |
| ----------- | ----------- |

And the Pros and Cons we (@pivotal-bin-ju and I) can think about for now:

|   | Pros       | Cons |
|----------- | ----------- | ----------- |
| 1. Parsing Log | - No need to modify source code <br> - Only need to parse the logs    | - Get access to the users' log <br> -  Analyzing Logs for 2 months of the size roughly 20G can be time consuming <br> - It is hard to keep the Log  from last 2 months for some users |
| 2. New DB Table| - Cleaner way to enumerate active users  |  - Extra effort on modifying the login flow and database schema <br> - Increase certain DB usage  |

***

### Issue Status

Working is being done on the [feature/4035-active-users](https://github.com/concourse/concourse/tree/feature/4035-active-users) branch

***

**Sidenote** (@taylorsilva): when logging in as a user that's not part of any team (`guest` user in docker-compose) we get a HTTP 400 back and a blank screen at the `/sky/callback` endpoint. Logs show `"error":"user doesn't belong to any team"`. Thinking this may be a regression. This note is to remind us to check this out and potentially report.
and fly:
```
fly -t local login -u guest -p guest
logging in to team 'main'

error: oauth2: cannot fetch token: 400 Bad Request
Response:
```

**Sidenote 2** (@cirocosta): Concourse allows operators to set the duration of the tokens that are generated (see https://github.com/concourse/concourse/blob/e5bcd88fa311d0c35342e9e34bbb7c7f7c72b10f/skymarshal/skycmd/flags.go#L48), potentially allowing folks to take a very long time to have to re-authenticate.

agurney:
From my perspective (large enterprise, many users) a logging approach is very reasonable, given that we already have infrastructure for log processing, security analytics / SIEM, etc. We are also interested in characterizing usage in finer detail - not just when logins happened, but what people did - which is the kind of thing that detailed logs can provide.

marco-m:
[...]

PR #4096, opened 2019-07-03
[atc/api] Enumerate Active Users

pivotal-bin-ju:
Fixes https://github.com/concourse/concourse/issues/4035

Looking to get some feedback on this code we have so far. Wondering about things like the error message we return [here](https://github.com/concourse/concourse/blob/89766859016262de13199e4798be12b145c01b30/atc/api/usersserver/get.go#L35) and just if we're on the right track so far with implementing this new endpoint.

CC @taylorsilva


xtremerui:
https://ci.concourse-ci.org/builds/6450778#L5d1bd07e:1

Think you would need to bump the migration timestamp. Thx! @pivotal-bin-ju

atc/db/user_factory.go
```diff
@@ -0,0 +1,104 @@
...
+
+		users = append(users, currUser)
+	}
+	return users, nil
+}
+
+func (f *userFactory) GetAllUsersByLoginDate(lastLogin time.Time) ([]User, error) {
+	rows, err := psql.Select("id", "username", "connector", "last_login").
+		From("users").
+		Where("last_login >= ?", lastLogin).
```
pivotal-jwinters: Can we change this to use `sq.GtOrEq{}`

atc/db/migration/migrations/1561054662_users_table.up.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+  CREATE TABLE users (
+    "id" serial NOT NULL PRIMARY KEY,
+    "username" text NOT NULL,
+    "connector" text NOT NULL,
+    "last_login" timestamp with time zone DEFAULT now() NOT NULL
+  );
+  ALTER TABLE ONLY users ADD CONSTRAINT user_unique UNIQUE (username,connector);
```
pivotal-jwinters: When we auth with dex we get back a unique `sub` in the token. Can we add a column to track this value and unique on it, instead of the username/connector combo.

atc/db/user_factory_test.go
```diff
@@ -0,0 +1,78 @@
...
+	JustBeforeEach(func() {
+		user, err = userFactory.CreateOrUpdateUser("test", "github")
+		Expect(err).ToNot(HaveOccurred())
+
+	})
+
+	Context("when user doesn't exist", func() {
+		It("Insert a user with last_login now()", func() {
+			Expect(user.Name()).To(Equal("test"))
+			Expect(user.LastLogin().Truncate(1 * time.Minute).String()).To(Equal(time.Now().Truncate(1 * time.Minute).String()))
```
pivotal-jwinters: FYI gomega has a temporally equivalent matcher:
https://onsi.github.io/gomega/#betemporallycomparator-string-compareto-timetime-threshold-timeduration

go-concourse/concourse/users.go
```diff
@@ -0,0 +1,10 @@
+package concourse
+
+import (
+	"github.com/concourse/concourse/atc"
+	"time"
+)
+
+func (client *client) ListActiveUsersSince(since time.Time) ([]atc.User, error) {
```
pivotal-jwinters: Is this used somewhere?

atc/db/user_factory.go
zoetian: Yes we can, thanks for pointing this out.

atc/db/migration/migrations/1561054662_users_table.up.sql
zoetian: Do you mean to add a new column `sub`?  Or completely replace the columns `username` or `connector`?
It seems like the current `username/connector` is more readable and allow us to trace the source the authentication.
WDYT?

go-concourse/concourse/users.go
zoetian: This is for `fly` and it's WIP.  We are planning to have a fly command that can print out all active users or some of the users since the given date.

YoussB:
@pivotal-jwinters, after @zoetian's update, do you have more comments are should we continue with the implementation of `fly`?

atc/db/migration/migrations/1561054662_users_table.up.sql
pivotal-jwinters: lets keep the username connector but also include a `sub` or `user_id` and unique on that

YoussB:
@pivotal-jwinters @xtremerui this should be ready for review now 👍

atc/db/migration/migrations/1563997651_users_table.up.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+  CREATE TABLE users (
+    "id" serial NOT NULL PRIMARY KEY,
+    "sub" text NOT NULL,
+    "username" text NOT NULL,
+    "connector" text NOT NULL,
+    "last_login" timestamp with time zone DEFAULT now() NOT NULL
+  );
+  ALTER TABLE ONLY users ADD CONSTRAINT user_unique UNIQUE (sub,username,connector);
```
pivotal-jwinters: You only need to unique on `sub` here.

taylorsilva:
We have one unit test failing: 
- https://ci.concourse-ci.org/teams/main/pipelines/prs/jobs/unit/builds/1487
- https://ci.concourse-ci.org/teams/main/pipelines/prs/jobs/unit/builds/1490

but it's only failing in CI. It doesn't fail on local. Thought it might be an issue with different ginkgo versions between ci and local but that was a dead-end. (ci and local using ginkgo v1.8.0)

Also, for some reason one test fails locally but not on ci 🤔 
```console
• Failure [3.316 seconds]
Web Command when --tsa-host-key is specified but no session signing key is provided [It] should fail due to missing db credentials
/Users/pivotal/workspace/concourse/cmd/concourse/concourse_test.go:77

  No future change is possible.  Bailing out early after 2.005s.
  Got stuck at:
      {"timestamp":"2019-07-30T22:08:21.646999000Z","level":"info","source":"atc","message":"atc.cmd.start","data":{"session":"1"}}
      {"timestamp":"2019-07-30T22:08:21.680895000Z","level":"info","source":"atc","message":"atc.cmd.finish","data":{"duration":142688,"session":"1"}}

  Waiting for:
      dial tcp 127.0.0.1:5432: connect: connection refused

  /Users/pivotal/workspace/concourse/cmd/concourse/concourse_test.go:78
```

taylorsilva:
Got all the tests passing!

@pivotal-jwinters @xtremerui one last review and then merge please 😄

atc/user.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"time"
+)
+
+type User struct {
+	ID        int       `json:"id,omitempty"`
+	Username  string    `json:"username,omitempty"`
+	Connector string    `json:"connector,omitempty"`
+	LastLogin time.Time `json:"last_login,omitempty"`
```
xtremerui: We are using int64 for timestamp in api for example [build](https://github.com/concourse/concourse/blob/5824dfd73cba6a3d9a256a2317f45a92478aa2b4/atc/build.go#L22). Could we make it the same for `LastLogin`? I think it will simplify some tests that related to timestamp as well.

taylorsilva: Latest commit changes the type to int64 now.

taylorsilva:
@pivotal-jwinters @xtremerui switched the `LastLogin` type from `time.Time` to `int64`. Please review again

Issue #4098, opened 2019-07-04
Resolve Failing Topgun Tests on PKS


scottietremendous:
[@cirocosta here]:

**Summary**

In our attempts to run Concourse with PKS, we've encountered a number of failing
[topgun tests][topgun]. 

At the moment, there are two classes of tests that fail:

- [baggageclaim drivers](https://github.com/concourse/concourse/blob/2c3e9dd527a332bd5816228ff67596da5bcf791f/topgun/k8s/baggageclaim_drivers_test.go)
- [container limits](https://github.com/concourse/concourse/blob/2c3e9dd527a332bd5816228ff67596da5bcf791f/topgun/k8s/container_limits_test.go)

In order the have [PKS tests in the main pipeline](https://github.com/concourse/ci/issues/40) successfully running, these
need to be addressed first.

**Acceptance Criteria**

- [ ] All failing tests passing.

**Extra context**

Those two classes of test have something in common - they require knowledge of
what is the underlying infrastructure that they run on.

E.g., to ensure that a particular set of baggageclaim drivers work across a
given set of node-level base images, we intentionally target a set of node pools
that contain a specific tag.

The number of pools to target then varies greatly:

- GKE allows us to have 3 different types of base images:
  - container optimized
  - container optimized (COS) with ContainerD
  - ubuntu
- PKS allows us to use only one - the Ubuntu-based stemcells.

[topgun]: https://github.com/concourse/concourse/tree/master/topgun/k8s

Thank you!

cirocosta:
List of failing tests:

```
Summarizing 8 Failures:

[Fail] Container Limits [JustBeforeEach] using cos as NodeImage returns the configure default container limit
/Users/pivotal/workspace/concourse/topgun/k8s/k8s_suite_test.go:172

[Fail] Container Limits [JustBeforeEach] using Ubuntu as NodeImage fails to set the memory limit
/Users/pivotal/workspace/concourse/topgun/k8s/k8s_suite_test.go:172

[Fail] Baggageclaim Drivers across different node images [It] with btrfs on cos
/Users/pivotal/workspace/concourse/topgun/k8s/k8s_suite_test.go:172

[Fail] Baggageclaim Drivers across different node images [It] with btrfs on ubuntu
/Users/pivotal/workspace/concourse/topgun/k8s/k8s_suite_test.go:172

[Fail] Baggageclaim Drivers across different node images [It] with overlay on cos
[...]

PR #4102, opened 2019-07-04
topgun/k8s: refactor tests to support targeting PKS

cirocosta:
Hey,

@taylorsilva and I started working on making the tests that currently fail on PKS (https://github.com/concourse/concourse/issues/4098#issuecomment-508549862) possible to run there.

Our approach is to conditionally `Skip()` based on the Kubernetes IAAS set (through environment variables). Trying to just fit another `if` in the current tests revealed that the code would become hard to reason about, so we started refactoring them:

- [x] `baggageclaim_drivers_test.go`
- [x] `container_limits_test.go`

Thanks!

### DIFF
--- a/topgun/k8s/baggageclaim_drivers_test.go
+++ b/topgun/k8s/baggageclaim_drivers_test.go
@@ -13,9 +13,9 @@ var _ = Describe("baggageclaim drivers", func() {
 	})
 
 	onPks(func() {
-		works("btrfs")
-		works("overlay")
-		works("naive")
+		baggageclaimWorks("btrfs")
+		baggageclaimWorks("overlay")
+		baggageclaimWorks("naive")
 	})
 
 	onGke(func() {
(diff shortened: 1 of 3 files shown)


YoussB:
I was able to get a full run today with the current codebase, there seems to be some flakiness in some tests that must be related to: https://github.com/concourse/ci/issues/44

YoussB:
makes sense, I can also add a fallback to GCP in case the variable is not set. So we can work on the other issue separately.

wdyt?

taylorsilva:
Wondering if we should be testing on both GKE and PKS? My gut tells me we should otherwise we'll have some stale tests if we only run tests on one k8s engine. 

Kinda outside the scope of this story though. Figured we could talk about this in https://github.com/concourse/ci/issues/40

YoussB:
@taylorsilva, I think we should be testing against both for sure, for the time being, until we figure out how we'll be testing on PKS I added the logic to default the testing to GKE if the environment variable is not set.

@cirocosta, it can safely be merged now, I guess.

Issue #3759, opened 2019-04-19
Performance regression on triggered jobs


freelock:
## Bug Report

The scheduler changes in 5.1.0 calls out that triggered jobs now must run a version check on every resource before the job is triggered. This makes Concourse practically unusable for us, as we trigger almost every run using a bot, and this has increased the startup time for each pipeline job from a few minutes to nearly 20 minutes. Across > 50 pipelines, a nightly triggered job now takes 18 hours instead of 3 1/2 -- meaning there's not much time left before the nightly job runs again.

## Steps to Reproduce

1. Our typical pipeline has ~19 resources. The vast majority of these are shared. We have started the binary Concourse Web with --enable-global-resources.
2. We run over 50 of these, generally 2 at a time (if there are more than 2 pipelines unpaused, our check worker historically has reached ridiculous load levels and basically locks up).
3. Unpause a pipeline, trigger one of the jobs that uses 4 resources -- one of them is unique to this pipeline, the other 3 shared across all other pipelines.
4. Wait for 15 - 20 minutes for the job to actually start. Even after this worker has all the containers/volumes from the pipeline being previously active.

## Expected Results

Our job should start within 5 minutes, if none of its resources are on the worker, or a minute or two if the other global resources are already present and have not exceeded some "freshness" age.

## Actual Results

I think after the job is triggered, all 19 resources in the pipeline are being created and checked. If they are being shared across other pipelines, it does not seem to be affecting the startup time...

## Additional Context

I think the ATC should go back to using existing resources in triggered jobs, if they have been checked previously (e.g. using the value in the resource's "check_every" parameter -- if the current resource on the worker is less than that value's age, use the cache version, and only fetch a fresher resource if the cached resource is older than that, or a "fly check-resource" has been called.

If the current behavior is preferred, then at least a startup option for the web?

## Version Info

* Concourse version: 5.1.0
* Deployment type (BOSH/Docker/binary): Binary
* Infrastructure/IaaS: GCE -- n1.standard web, n1.standard worker with local SSD disk for worker
* Browser (if applicable): n/a
* Did this used to work? Much better than now!

pivotal-jwinters:
@freelock manually triggered builds have actually always run a version check for every resource. The change in 5.1.0 only means that those checks don't block other builds from running anymore. Let's see if I can explain this with some pseudo-code.

Pre 5.1.0
```
for job in jobs
  if job.manuallyTriggered
    // this blocks other jobs from running
    for resource in job.resources
[...]

PR #4103, opened 2019-07-04
Fix missed check notifications

clarafu:
If the interval runner is not started up yet and we send a notification
to check a resource, the notification gets lost. This PR adds a
safeguard by having a condition notifier that will check a check
requested time value on the resources table to confirm whether or not we
need to run a check.

The resource type object was given the same notifier methods but are not used in order to maintain consistency between the two.

[#3759]


pivotal-jamie-klassen:
@clarafu can you rebase this? looks like the pr-resource [here](https://ci.concourse-ci.org/teams/main/pipelines/prs/resources/concourse-pr) missed or forgot this version.

clarafu:
@pivotal-jamie-klassen Okk, I rebased it!

vito:
@clarafu Looks like a migration snuck in to `master`, could you bump the migration number and re-push?

vito:
Actually after chatting with @pivotal-jwinters and @clarafu this is being superceded by #3788 - closing!

The plan is to use the new check queue for manually queued `check`s (`fly check-resource` and manually triggered jobs), and have a feature-flag for whether we rely on it for mission-critical continuous resource checking until we see it going well in production environments.

PR #4108, opened 2019-07-08
Add locking around steps.

aledeganopix4d:
This is my tentative at fixing #3301. Taking a pragmatic way and implementing a lock wherever needed.

All steps chose the worker serially by acquiring the same lock.

The lock is acquired in the task, put and get steps,
each implemented in the relative Run method.

The lock is released by the corresponding method that creates
the Container in the database just after that's done:
either `worker.FindOrCreateContainer` or `fetcher.Fetch`.

Resource_scanner, resource_type_scanner and image_resource_fetcher
do not implement the locking mechanism thus their processing is independent.

The unit tests are adjusted for the addition of the lock.

New test cases are probably required but I'd like to have feedback on the general direction I've taken
before working more on this.

@ddadlani @deniseyu @kcmannem 

### DIFF
--- a/atc/resource/fetcher.go
+++ b/atc/resource/fetcher.go
@@ -30,6 +30,7 @@ type Fetcher interface {
 		resourceTypes atc.VersionedResourceTypes,
 		resourceInstance ResourceInstance,
 		imageFetchingDelegate worker.ImageFetchingDelegate,
+		containerLock lock.Lock,
 	) (VersionedSource, error)
 }
 
@@ -60,6 +61,7 @@ func (f *fetcher) Fetch(
 	resourceTypes atc.VersionedResourceTypes,
 	resourceInstance ResourceInstance,
 	imageFetchingDelegate worker.ImageFetchingDelegate,
+	containerLock lock.Lock,
 ) (VersionedSource, error) {
 	containerSpec.Outputs = map[string]string{
 		"resource": ResourcesDir("get"),
@@ -71,6 +73,9 @@ func (f *fetcher) Fetch(
 	defer ticker.Stop()
 
 	versionedSource, err := f.fetchWithLock(ctx, logger, source, imageFetchingDelegate.Stdout())
+	if containerLock != nil {
+		containerLock.Release()
+	}
 	if err != ErrFailedToGetLock {
 		return versionedSource, err
 	}
--- a/atc/resource/resource_instance_fetch_source.go
+++ b/atc/resource/resource_instance_fetch_source.go
@@ -130,7 +130,6 @@ func (s *resourceInstanceFetchSource) Create(ctx context.Context) (VersionedSour
 		&worker.CertsVolumeMount{Logger: s.logger},
 	}
 
-
 	container, err := s.worker.FindOrCreateContainer(
 		ctx,
 		s.logger,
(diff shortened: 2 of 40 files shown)


atc/resource/fetcher.go
```diff
@@ -71,6 +73,9 @@ func (f *fetcher) Fetch(
 	defer ticker.Stop()
 
 	versionedSource, err := f.fetchWithLock(ctx, logger, source, imageFetchingDelegate.Stdout())
+	if containerLock != nil {
+		containerLock.Release()
+	}
```
ddadlani: the lock would also need to be released after the call to `fetchWithLock` on line 86 below.

ddadlani: the lock would need to be released after the second call to `fetchWithLock()` on line 86 since container creation could occur there, but this is susceptible to deadlock because we have 2 locks being held/released here.

atc/resource/resource_instance_fetch_source.go
```diff
@@ -139,6 +138,7 @@ func (s *resourceInstanceFetchSource) Create(ctx context.Context) (VersionedSour
 		s.session.Metadata,
 		s.containerSpec,
 		s.resourceTypes,
+		nil,
```
ddadlani: We should be passing the lock in here since we could be creating a `db.CreatingContainer` inside `FindOrCreateContainer`.

aledeganopix4d:
> Hi @aledeganopix4d
> 
> Thanks for working on fixing #3301!
> 
Hi!
Thank you for taking the time to review this PR!
> My biggest concern with this implementation is that all steps are effectively serialized. Some `get` or `task` steps could take quite a while and holding the lock for this time would mean that nothing else would get scheduled while a step is running, which seems too restrictive.
> 
> For the use case of #4076, I think it would be sufficient to have a `TaskStepLock` that is only required with that feature. What are your thoughts?
Yes, for #4076 it would be sufficient, however the bug reported with #3301 won't be solved for all use cases: `in_parallel` put and gets could still be landing on the same worker. 
> 
> We also have some incoming refactors (such as #4109) that will break this locking mechanism because we are trying to centralize container creation in one place (to enable things like scheduling and multiple runtimes).

I've noticed #4109 and I was indeed wondering how and when that refactor would impact my work with this PR and #4076. But my understanding is that I can keep working on the current HEAD as that rework might take a while?
> 
> Would it be sufficient for your use case to limit the lock to #4076?

Yes, in that case, probably, it's better if I create a single PR, that introduces `max-build-tasks-per-worker` and the minimum amount of locking required for that mechanism to work as intended.
Do you agree?

ddadlani:
Hi @aledeganopix4d,

We are planning to merge #4109 into master in the next couple of days. The only effect it should have on your PR is where the "active_tasks" number is incremented. In fact, you could likely move the calls to IncrementActiveTasks and DecrementActiveTasks and the new locking logic, into `RunTaskStep` in worker.Client and it could remain encapsulated in that part of the code.

What do you think?

ddadlani:
> Yes, in that case, probably, it's better if I create a single PR, that introduces max-build-tasks-per-worker and the minimum amount of locking required for that mechanism to work as intended.
Do you agree?

Yes, I agree

aledeganopix4d:
> Hi @aledeganopix4d,
> 
> We are planning to merge #4109 into master in the next couple of days. The only effect it should have on your PR is where the "active_tasks" number is incremented. In fact, you could likely move the calls to IncrementActiveTasks and DecrementActiveTasks and the new locking logic, into `RunTaskStep` in worker.Client and it could remain encapsulated in that part of the code.
> 
> What do you think?

That's perfect, thanks!

aledeganopix4d:
Superseded by #4118

Issue #3695, opened 2019-04-05
Better build scheduling / load distribution


ddadlani:
There have been multiple issues (#2577, #2928 etc) opened regarding Concourse's build/container scheduling and the container placement strategies currently present (`random`, `volume-locality` and `fewest-build-containers`) are not able to equitably distribute the workload because 

1) they are not aware of the actual load present on the workers (number of volumes and/or build containers are loose approximations)
2) there is no queueing logic, meaning that even if all the workers are busy, any new work will still be scheduled immediately, which can cause workers to fall over.

This is a meta-issue to discuss all the use cases and possible implementations for better work scheduling in Concourse.

qoomon:
A queue on concourse master would be awesome.
* Workers could then pull as many job as configured.
* Scaling would be much easier to implement, just monitor queue size and scale accordingly
  * `off-topic` would be great to have an option to shutdown workers after a specific count of job runs

ddadlani:
Doing some investigation into a `btrfs` issue on one of our workers, we came across [this thread](https://www.mail-archive.com/linux-btrfs@vger.kernel.org/msg45295.html) which suggests that scaling to >1000 volumes/subvolumes at a time is known to cause various issues. We might be able to mitigate the occurrence of these issues with a queue, since we can avoid scheduling more than ~1000 volumes on a worker.

kcmannem:
Thoughts on the scheduler:

We wrote placement strategies based on singular requirements wether its to pick a worker with the most locally required resources or one that has the fewest builds on it. We've known that no one strategy is best suited in any situation as tasks come in many shapes and sizes. In most situations we want to make placement decisions based on many requirements. These requirements vary based on the tasks users run and environmental state. Examples:
- Don't pick workers at max containers/volumes
- Don't pick workers in an error state for retries
- Don't pick workers with high load
- Don't pick a recently chosen worker

Some of these requirements have manifested themselves into strategies, some are not possible to do in our current flow. The main problem is that we can't merge them. We shouldn't look at worker placement as definitive, but rather providing options. Each strategy now returns a list of workers that meet the strategies criteria (filtering out ones that don't, so they are not picked). This gives the ability to stack and reorder these requirements in any way we want. 

If our strategy interface looked like this:
```[]worker -> Stragegy() -> []worker```
then this would be possible with our current strategies (lets ignore their correctness for now)
```[]worker -> fewestBuilds -> volumeLocality -> rand -> []worker```
[...]

PR #4109, opened 2019-07-08
Break out task step into core and runtime sides

ddadlani:
This is a draft PR to attempt splitting out the TaskStep into core-specific and runtime-specific concerns.

The idea is to put all runtime-specific concerns for all steps into the `worker.Client` so that we have a single place to optimize container placement and worker selection for stories such as #3695, #2928 and #3301. This PR attempts it only for the task step.

The tests do not currently pass, however this is a draft PR so that others can give their feedback on this approach. Fixing the tests is a work in progress.


atc/exec/task_step.go
```diff
@@ -188,135 +185,78 @@ func (step *TaskStep) Run(ctx context.Context, state RunState) error {
...
+			Status: status,
+			VolumeMounts: volumeMounts,
+			Err: err,
 		}
 
-		err = container.Stop(false)
-		if err != nil {
-			logger.Error("stopping-container", err)
-		}
+	}(results)
```
kcmannem: small change, I think closures capture the surrounding vars for you to use inside. You might not need to pass in results into the go func. This works too:
```
	foo := make(chan int, 1)
	go func() {
		foo <- 5
	}()
	fmt.Println(<-foo)
```

kcmannem:
I like this change

atc/exec/task_step.go
ddadlani: Hmm, I've been bitten by that before, because closures capture the surrounding variables at the time that the `go func()` is called. I guess that doesn't matter for channels.

Is it more readable/less likely to break after a refactor if we explicitly pass it in? If not, I'm happy to remove it

atc/worker/client.go
```diff
@@ -1,16 +1,38 @@
...
 type Client interface {
 	FindContainer(logger lager.Logger, teamID int, handle string) (Container, bool, error)
 	FindVolume(logger lager.Logger, teamID int, handle string) (Volume, bool, error)
 	CreateVolume(logger lager.Logger, vSpec VolumeSpec, wSpec WorkerSpec, volumeType db.VolumeType) (Volume, error)
+	RunTaskStep(
+		context.Context,
+		lager.Logger,
+		db.ContainerOwner,
+		ContainerSpec,
+		WorkerSpec,
```
nader-ziada: `WorkerSpec` and `ContainerPlacementStrategy` are Garden specific? are these fields optional?

atc/worker/client.go
```diff
@@ -1,16 +1,38 @@
...
+		ContainerPlacementStrategy,
+		db.ContainerMetadata,
+		ImageFetcherSpec,
+		TaskProcessSpec,
+		chan string,
+	) (int, []VolumeMount, error)
```
nader-ziada: Why is `Run` returning `VolumeMount`?

atc/worker/client.go
ddadlani: I don't see a way to avoid these for now. We may need to revisit this after we figure out how tags/teams and strategies are handled in k8s. We can make it optional for the k8s side of things.

`ContainerPlacementStrategy` might be able to be passed down at startup from `atccmd/command.go`. We could have a separate one for checks and for gets/puts/tasks, but this will depend on how all the other steps/checks are run.

atc/worker/client.go
ddadlani: Because we need to [registerOutputs](https://github.com/concourse/concourse/blob/392c00e931b5b81db6a8a01b926edaf6a2dfa087/atc/exec/task_step.go#L388-L433) in the TaskStep. Decoupling volumes/storage from Running a step is something we would need to look into, but in the absence of a separate Storage implementation, we would have to return volumes this way.

For example, we would likely register "outputs" beforehand with a specific storage implementation, and then de-register them if the task failed. That would allow us to not depend on VolumeMounts in the future.

atc/exec/task_step.go
```diff
@@ -188,135 +184,69 @@ func (step *TaskStep) Run(ctx context.Context, state RunState) error {
...
+			err = step.registerOutputs(logger, repository, config, result.VolumeMounts, step.containerMetadata)
+			if err != nil {
+				return err
+			}
 
-		<-exited
+			return ctx.Err()
 
-		return ctx.Err()
+		case result := <-results:
```
pivotal-jwinters: Can we unify this code path with the ctx.Done code path? 

I'm a little confused why we need both. [In the old task step](https://github.com/concourse/concourse/blob/master/atc/exec/task_step.go#L284-L318) it seemed really clear that we were waiting on either the [process to exit](https://github.com/concourse/concourse/blob/master/atc/exec/task_step.go#L299), or the [context to be cancelled](https://github.com/concourse/concourse/blob/master/atc/exec/task_step.go#L284) (which would prompt us to [stop the container](https://github.com/concourse/concourse/blob/master/atc/exec/task_step.go#L290) and then still wait for the [process to exit](https://github.com/concourse/concourse/blob/master/atc/exec/task_step.go#L295)).

It looks like this logic still exists in `RunTaskStep`, so just wondering why we need both here?

atc/exec/task_step.go
```diff
@@ -188,135 +184,69 @@ func (step *TaskStep) Run(ctx context.Context, state RunState) error {
...
-		err = container.Stop(false)
-		if err != nil {
-			logger.Error("stopping-container", err)
-		}
+	}(results)
+
+	for {
+		select {
+		case <-ctx.Done():
+			result := <-results
```
xtreme-sameer-vohra: Do we intend to block on this line ? Is there a chance where this might not return ever ?

atc/worker/pool.go
```diff
@@ -176,3 +180,37 @@ func (pool *pool) FindOrChooseWorker(
 
 	return workers[rand.Intn(len(workers))], nil
 }
+
+func (pool *pool) FindOrCreateContainer(
```
xtreme-sameer-vohra: Do we need this method ?

xtreme-sameer-vohra:
Overall, the `Task Step` looks much simpler than before and once this style of refactoring is applied to the other steps, they might similarly become much more cleaner, to the point, where there may arise 1 or 2 generic steps. 

One thing to watch out for is how `Client` would evolve as a result of this path ( once we add the other steps) . Might it become quite large with lots of state to initialize and private functions to `DRY` things up. This might in turn making testing much more complicated as it becomes a `thick` client.  In this regard, the current step design (eg. K8sTaskStep, GardenTaskStep ) would be smaller interfaces with dependencies can easily be injected for testing.

atc/exec/task_step.go
ddadlani: As per our discussion, we changed event handling to be handled in a separate goroutine. This allowed us to collapse the `select` block in question into just one code block (no channels required). Instead, event handling is the only thing requiring a channel now.

atc/exec/task_step.go
ddadlani: We did intend to block on it, but after some refactoring we were able to run `RunTaskStep` in the same goroutine so no channel blocking required

ddadlani:
@xtreme-sameer-vohra thanks for reviewing the PR!

> Overall, the `Task Step` looks much simpler than before and once this style of refactoring is applied to the other steps, they might similarly become much more cleaner, to the point, where there may arise 1 or 2 generic steps.
 
@nader-ziada and I were looking at the code, and we realized that the `Get` and `Put` steps and `Check`s also do a very similar thing, in that they select a worker, create a container and then [run a defined script](https://github.com/concourse/concourse/blob/8610e881837047770466256a7a26d3200bc28b07/atc/resource/resource_get.go#L26-L37) which is very similar to the setup for the task script. In the future, ideally we would be able to remove the concept of a "type" of step from runtime, and instead have a defined set of things (select worker, create container, run script, return result) that would allow us to run any kind of step.

> One thing to watch out for is how `Client` would evolve as a result of this path ( once we add the other steps) . Might it become quite large with lots of state to initialize and private functions to `DRY` things up. This might in turn making testing much more complicated as it becomes a `thick` client. In this regard, the current step design (eg. K8sTaskStep, GardenTaskStep ) would be smaller interfaces with dependencies can easily be injected for testing.

I agree. Currently `worker.Client` is making a lot of decisions on what to do. In the ideal future I would see it become `runtime.Client` which simply decides on what runtime backend to use and then invoke the required method on that. The code currently in `worker.Client` is more suited to being in a `runtime` implementation rather than in a client. Hopefully future refactor efforts enable this view.

atc/worker/pool.go
ddadlani: Whoops, no we don't

atc/exec/task_step.go
```diff
@@ -176,6 +173,26 @@ func (step *TaskStep) Run(ctx context.Context, state RunState) error {
...
+			ev := <-events
+			switch {
+			case ev.EventType == runtime.InitializingEvent:
+				step.delegate.Initializing(logger, config)
+
+			case ev.EventType == runtime.StartingEvent:
+				step.delegate.Starting(logger, config)
+
+			case ev.EventType == runtime.FinishedEvent:
+				step.delegate.Finished(logger, ExitStatus(ev.ExitStatus))
```
pivotal-jwinters: I think this `goroutine` is going to leak. You can either `return` after the finish event (which may or may not get called?) or you can [listen for events on a separate channel which gets closed when the surrounding method terminates](https://github.com/concourse/concourse/blob/master/atc/engine/engine.go#L175-L180)

ddadlani: @jwinters The channel is closed on line 235 below, would that not go through the `default` branch? I can put line 235 in a `defer` to ensure that it gets closed regardless of errors.

ddadlani: Changed it to use [range and close](https://tour.golang.org/concurrency/4) instead

atc/exec/task_step.go
```diff
@@ -181,6 +177,22 @@ func (step *TaskStep) Run(ctx context.Context, state RunState) error {
 		config.Limits.Memory = step.defaultLimits.Memory
 	}
 
+	events := make(chan runtime.Event, 1)
+	go func(logger lager.Logger, config atc.TaskConfig, events chan runtime.Event, delegate TaskDelegate) {
+		for ev := range events {
+			switch ev.EventType {
+			case runtime.InitializingEvent:
```
xtreme-sameer-vohra: Doesn't look like this event is called ?

atc/exec/task_step.go
```diff
@@ -181,6 +177,22 @@ func (step *TaskStep) Run(ctx context.Context, state RunState) error {
...
+				step.delegate.Initializing(logger, config)
+
+			case runtime.StartingEvent:
+				step.delegate.Starting(logger, config)
+
+			case runtime.FinishedEvent:
```
xtreme-sameer-vohra: Doesn't look like this event is called ?

atc/exec/task_step.go
```diff
@@ -193,192 +205,65 @@ func (step *TaskStep) Run(ctx context.Context, state RunState) error {
...
-	if err == nil {
-		logger.Info("already-exited", lager.Data{"status": exitStatusProp})
-
-		status, err := strconv.Atoi(exitStatusProp)
-		if err != nil {
-			return err
-		}
 
-		step.succeeded = (status == 0)
+	close(events)
```
xtreme-sameer-vohra: Should it be `RunTaskStep`'s responsibility to close the channel since it is the producer?

atc/exec/task_step.go
ddadlani: we'll address these changes in a follow up PR because it's been a while since this PR has been open and keeping it up to date with master is problematic.

ddadlani:
Another thing to consider is that for heavy tasks like Strabo, we will be creating a lot more goroutines since a new event goroutine is created for every task container

PR #4122, opened 2019-07-12
Update CONTRIBUTING.md for atc

taylorsilva:


atc/CONTRIBUTING.md
```diff
@@ -46,33 +37,8 @@ createdb atc
...
-cd ./web
-make -B
-```
-
-Finally you can run the ATC:
-
 ```
-cd ..
-go run cmd/atc/*.go --add-local-user test:test --main-team-local-user test
+go run cmd/atc/*.go run --add-local-user test:test --main-team-local-user test
```
nader-ziada: It's not clear that the use should run this command inside the `atc` folder

vito:
Hmm this is all cruft from before the 'one big repo' initiative. Should we just remove it?

taylorsilva:
@vito updated to just delete it

PR #4160, opened 2019-07-24
bump yarn dependencies

vito:
i just blindly ran 'yarn upgrade' - hopefully this silences all the security alerts

### DIFF
--- a/web/wats/yarn.lock
+++ b/web/wats/yarn.lock
@@ -64,10 +64,10 @@ abbrev@1:
   resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.1.1.tgz#f8f2c887ad10bf67f634f005b6987fed3179aac8"
   integrity sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==
 
-agent-base@^4.1.0:
-  version "4.2.0"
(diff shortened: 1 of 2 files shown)


jomsie:
So looks like the error on testflight is
```
error /tmp/build/13b0078c/concourse/node_modules/elmi-to-json: Command failed. 
Exit code: 1
Command: binwrap-install
...
ERR { Error: EEXIST: file already exists, mkdir '/tmp/build/13b0078c/concourse/node_modules/elmi-to-json/unpacked_bin'
    at Object.fs.mkdirSync (fs.js:885:18)
...
```

So I just tried a `yarn upgrade` on master locally, and got that exact issue.

And looking around I found
https://github.com/yarnpkg/yarn/issues/2629#issuecomment-283691556
and
https://github.com/yarnpkg/yarn/issues/2629#issuecomment-283511162

So I tried a `yarn upgrade --network-concurrency 1`, and it worked.

vito:
@jomsie I think the cache just had to be cleared. I've kicked off another build, looks like it got past that.

pivotal-jamie-klassen:
Looks like a bit more than `yarn upgrade` is required here - that lodash dependency shows up a bunch. Looks like the latest version of `elm-analyse` has updated its lodash dependency and i think we should drop the dependency on `chokidar-cli` as, personally, i never use it and it is becoming a bit of noise.

Only spent about 30 minutes looking at this this morning, and I can see that lodash appears transitively in a few other spots, so this might not be the last we've heard of it.

pivotal-jamie-klassen:
superceded by #4184

PR #4161, opened 2019-07-24
change latest migration version to the current timestamp

ddadlani:
the add_active_tasks_to_workers migration had a random number as the
version, which would cause future migrations to get skipped until their
versions passed this number.

concourse/concourse#2928

Co-authored-by: Rui Yang <ryang@pivotal.io>


ddadlani:
The `unit` job has failed because it checks to ensure migrations aren't removed. However, the purpose of this PR is to rename a migration, which means that it looks like a previous migration has been removed. Because of this, I am going to merge in this PR despite the unit failure. I ran the unit tests locally and tested running the migrations using `docker-compose` to be sure that this does not break.

Thanks to @xtremerui for pointing this out.

aledeganopix4d:
@ddadlani thank you for taking care of this. Admittedly I didn't get how the versions number where generated (and didn't think at timestamps -.-').
I meant to ask then forgot...

ddadlani:
@aledeganopix4d no problem! It was also my fault for not updating the wiki's CONTRIBUTING.md with migration information sooner. That is now also updated though, so future contributors will know how :)

Issue #3078, opened 2019-01-17
Paused version border moves content slightly


vito:
## Bug Report

![jitter](https://user-images.githubusercontent.com/1880/51328566-ac742500-1a41-11e9-87c7-ad2979cfb707.gif)

## Steps to Reproduce

Pause or unpause a resource version

Issue #2608, opened 2018-09-20
A "pipeline" view of tasks in a job


pn-santos:
## What challenge are you facing?

In some jobs it can be useful to run alternating "sections" with several parallel tasks and sequential ones. Mix that in with doing regular `put` actions with `on_failure` and `on_success` triggers, a fair amount of logs and the job UI page can very quickly become hard to take in at a glance (e.g. quickly identify which tasks failed when run in parallel).

A practical example: I have a job that builds github PRs and then runs several checks in parallel (using the output of the build). When a check fails it's a chore to scroll up and down the job page to be able to see the logs of the checks that failed.

## What would make this better?

It would be nice if there was a way to see a job the way we see pipelines with horizontal left-to-right layout, no logs, just name of tasks and their status: pass/fail/pending/in progress. This in a way that the tasks relationships were visually represented (i.e. easily identify parallel tasks, if they are `on_<condition>` triggered tasks, etc.).

Then by clicking a task, we'd be able to see only it's log.

I think a nice side effect would be that these screens would be faster to load (no need to load all the logs, just task names, relationships and status) and even when looking at one task, it would only have to load the logs for that task, not the entire job.

## Are you interested in implementing this yourself?

Unfortunately no, this is more of a frontend kind of thing, so not really part of my skill set.

jama-pivotal:
Very interesting! I think it would be a very useful view

pn-santos:
I still think this would be useful, but I have zero chance of making it happen myself.

vito:
Hmm, one change we made was to have all steps auto-collapse by default, even while they're running. Does that help with this? (#4051)

Interestingly there's been some talk about having *running* steps auto-expand/collapse, but that seems to be the root of your issue. /cc @eedwards-sk

One problem with the current approach, aside from whether or not the steps auto-expand, is that it can be a little hard to notice that a step is running. I wonder if we should give them a yellow border? :thinking: /cc @Lindsayauchin @pivotal-jamie-klassen @jomsie

pn-santos:
> Does that help with this?

Assuming logs are lazy loaded, yes that helps. 

That said the "pipeline" view I think would be much better are representing fan-in/fan-out steps in a job imo.

The vertical view can get complicated when there is `parallel/aggregate` + `do` nesting.

eedwards-sk:
[...]

PR #4164, opened 2019-07-25
web: add red border to failing steps

### LINKED ISSUE #4111: Make auto-collapsed failed steps easier to notice
Follow-up to #4058 / #4051 - failed steps no longer auto-expand, which is nice because it keeps them from shoving other failed steps offscreen, but I find it hard to notice *which* step failed.

For example:

image

That little ❌ is a bit hard to notice, mainly because I tend to look at the left side of the screen I guess, and it's pretty far from the step name making it hard to scan over and see what step it is.

Maybe we could make the whole header background for the step red? 😮 

/cc @pivotal-jamie-klassen @Lindsayauchin

jomsie:
fixes #4111

# Release Note
```tex
\note{feature}{
  We noticed after \link{PR #4058}{https://github.com/concourse/concourse/pull/4058} 
  (where build steps are collapsed by default) that it wasn't very easy to see failing steps.
  
  Now a failed step has a simple red border around its header and an erroring step has an
  orange border
  \link{#4164}{https://github.com/concourse/concourse/pull/4164} 
}
```

### DIFF
--- a/web/elm/src/Build/Styles.elm
+++ b/web/elm/src/Build/Styles.elm
@@ -146,12 +146,15 @@ stepHeader : StepState -> List (Html.Attribute msg)
 stepHeader state =
     [ style "display" "flex"
     , style "justify-content" "space-between"
-    , case state of
-        StepStateFailed ->
-            style "border" "1px solid #C24433"
+    , style "border" <|
+        "1px solid "
+            ++ (case state of
+                    StepStateFailed ->
+                        Colors.failure
 
-        _ ->
-            style "border" "none"
+                    _ ->
+                        Colors.frame
+               )
     ]
 
 
--- a/web/elm/src/Build/Styles.elm
+++ b/web/elm/src/Build/Styles.elm
@@ -19,6 +19,7 @@ module Build.Styles exposing
 
 import Application.Styles
 import Build.Models exposing (StepHeaderType(..))
+import Build.StepTree.Models exposing (StepState(..))
 import Colors
 import Concourse
 import Dashboard.Styles exposing (striped)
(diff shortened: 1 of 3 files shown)


web/elm/src/Build/Styles.elm
```diff
@@ -141,10 +142,16 @@ triggerTooltip =
...
+stepHeader : StepState -> List (Html.Attribute msg)
+stepHeader state =
     [ style "display" "flex"
     , style "justify-content" "space-between"
+    , case state of
+        StepStateFailed ->
+            style "border" "1px solid #C24433"
+
+        _ ->
+            style "border" "none"
```
vito: `none` vs. `1px` will result in content slightly jumping when the border is added - similar to https://github.com/concourse/concourse/issues/3078

jomsie: Good snag! Will fix!

web/elm/src/Build/Styles.elm
```diff
@@ -141,10 +142,19 @@ triggerTooltip =
...
+stepHeader state =
     [ style "display" "flex"
     , style "justify-content" "space-between"
+    , style "border" <|
+        "1px solid "
+            ++ (case state of
+                    StepStateFailed ->
+                        Colors.failure
+
+                    _ ->
```
vito: :ghost: Scope creeping, but can we do the same for errored steps? Those are in the same boat: they cause a build failure (well, error) but are hard to notice when auto-collapsed.

I don't think there are any other states we care about. :thinking:

pivotal-jamie-klassen: Here's a prototype, but let's get @Lindsayauchin in on this sooner rather than later.
Screen Shot 2019-07-31 at 3 21 45 PM

jomsie:
[wats test failing](https://ci.concourse-ci.org/builds/6627310#L5d41da29:721), running locally it passes when it doesn't timeout

web/elm/src/Build/Styles.elm
Lindsayauchin: @vito @pivotal-jamie-klassen Makes sense. I like it!

vito: @Lindsayauchin @pivotal-jamie-klassen More :ghost: scope creep: what do you think of adding a yellow border for running steps, since they're auto-collapsed too?

some context: https://github.com/concourse/concourse/issues/2608#issuecomment-517296718

cirocosta:
hey, I just wanted to say that this is amazing ❤️ ! thank you all! (@xtremerui shares the same thought!)

PR #4167, opened 2019-07-26
[bugfix] Do not trigger Garden error when asking for container property.

aledeganopix4d:
When querying Garden for a property that does not exist, Garden
logs an error message. This error message pollutes the Worker
logs as it doesn't really indicate that something wrong has
been going on.

By querying all properties and checking if the interested one
is among them we obtain the same result while avoiding Garden
to log an error.

### DIFF
--- a/atc/resource/run_script.go
+++ b/atc/resource/run_script.go
@@ -52,9 +52,10 @@ func (resource *resource) runScript(
 	}
 
 	if recoverable {
-		result, err := resource.container.Property(resourceResultPropertyName)
-		if err == nil {
-			return json.Unmarshal([]byte(result), &output)
+		result, _ := resource.container.Properties()
+		code := result[resourceResultPropertyName]
+		if code != "" {
+			return json.Unmarshal([]byte(code), &output)
 		}
 	}
 
--- a/atc/worker/client.go
+++ b/atc/worker/client.go
@@ -161,16 +161,18 @@ func (client *client) RunTaskStep(
 	}
 
 	// container already exited
-	exitStatusProp, err := container.Property(taskExitStatusPropertyName)
-	if err == nil {
-		logger.Info("already-exited", lager.Data{"status": exitStatusProp})
+	exitStatusProp, _ := container.Properties()
+	code := exitStatusProp[taskExitStatusPropertyName]
+	if code != "" {
+		logger.Info("already-exited", lager.Data{"status": taskExitStatusPropertyName})
 
-		status, err := strconv.Atoi(exitStatusProp)
+		status, err := strconv.Atoi(code)
 		if err != nil {
 			return TaskResult{-1, []VolumeMount{}, err}
 		}
 
 		return TaskResult{Status: status, VolumeMounts: container.VolumeMounts(), Err: nil}
+
 	}
 
 	processIO := garden.ProcessIO{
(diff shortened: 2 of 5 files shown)


ddadlani:
Hmm, I re-triggered the unit test suite but looks like it's consistently failing:
```
	 -------------------------------------------------------------------
	|                                                                   |
	|  Ginkgo timed out waiting for all parallel nodes to report back!  |
	|                                                                   |
	 -------------------------------------------------------------------

 exec timed out. path: ./atc/exec

Ginkgo ran 24 suites in 7m36.277437208s
Test Suite Failed
```
Could you look into that?

aledeganopix4d:
> Hmm, I re-triggered the unit test suite but looks like it's consistently failing:
> 
> ```
> 	 -------------------------------------------------------------------
> 	|                                                                   |
> 	|  Ginkgo timed out waiting for all parallel nodes to report back!  |
> 	|                                                                   |
> 	 -------------------------------------------------------------------
> 
>  exec timed out. path: ./atc/exec
> 
> Ginkgo ran 24 suites in 7m36.277437208s
> Test Suite Failed
> ```
> 
> Could you look into that?

Thanks for checking it.
I've indeed payed more attention to the tests which needed updating.
They now complete successfully on my machine. Unfortunately seems there's an issue with the CI...

aledeganopix4d:
@ddadlani I've rebased this PR on top of #4109 and fixed (I hope) all the tests.

ddadlani:
Would be nice if we could have garden simply not log the error, and instead send back the `found` boolean that the guardian property manager emits. 

But that would require a bunch of changes to both garden and guardian, so this is fine. I fixed a small conflict, will merge as soon as the tests pass 👍

atc/resource/run_script.go
```diff
@@ -52,9 +52,11 @@ func (resource *resource) runScript(
 	}
 
 	if recoverable {
-		result, err := resource.container.Property(resourceResultPropertyName)
-		if err == nil {
-			return json.Unmarshal([]byte(result), &output)
+		result, _ := resource.container.Properties()
+		for prop, code := range result {
```
xtreme-sameer-vohra: Hey @aledeganopix4d 
Does it make sense to use a key lookup rather than iterating over the map ?

atc/worker/client.go
```diff
@@ -161,16 +161,18 @@ func (client *client) RunTaskStep(
 	}
 
 	// container already exited
-	exitStatusProp, err := container.Property(taskExitStatusPropertyName)
-	if err == nil {
-		logger.Info("already-exited", lager.Data{"status": exitStatusProp})
+	exitStatusProp, _ := container.Properties()
+	for prop, code := range exitStatusProp {
```
xtreme-sameer-vohra: Hey @aledeganopix4d
Similarly, does it make sense to use a key lookup rather than iterating over the map here as well?

aledeganopix4d: Yes! Of course! :man_facepalming: 
Thanks!

aledeganopix4d:
> @aledeganopix4d looks good to me, but what is `task_queue.png` for? It looks like a screenshot of your Concourse?
> 
> I also re-triggered the tests, seems like they failed one more time.

ooooops :laughing:

Issue #4114, opened 2019-07-10
fly: 'delete-target' with command option -a not working as expected


Jayasankar-m:
## Bug Report
The fly command with 'delete-target' does not work as expected with -a command option

## Steps to Reproduce

create a target using fly command
```fly login -t <target> -c http://<url>:8080 -u <username> -p <password>```

try to delete all targets using command
```fly delete-target -a```
(without giving any target name)

## Expected Results
All targets should get deleted as per the documentation
https://concourse-ci.org/fly.html#fly-delete-target

## Actual Results
Following error is shown
```
no target specified. specify the target with -t or log in like so:

    fly -t (alias) login -c (concourse url)
```

## Additional Context
if a target is also specified as follows then all entries are getting removed
```fly -t <target> delete-target -a```

## Version Info

* Concourse version: 5.3.0

pivotal-bin-ju:
A PR was drafted for this: https://github.com/concourse/concourse/pull/4172/commits/cd77717cdeab44e4163c5e4209bf0a4ef96d7de5

PR #4172, opened 2019-07-26
delete all targets without target parameter

pivotal-bin-ju:
PR for https://github.com/concourse/concourse/issues/4114


deniseyu:
I added the test case, and also noticed that we didn't have a test case for when no target is specified. hope that's ok @pivotal-bin-ju!

jomsie:
i saw wats failed, probably because of the recent watsjs problems on the ui, so i restarted it with this pr

fly/integration/delete_target_test.go
```diff
@@ -74,6 +74,20 @@ var _ = Describe("Fly CLI", func() {
 				os.RemoveAll(tmpDir)
 			})
 
+			Context("when no fly target is specified", func() {
+				It("asks the user to specify a pipeline", func() {
```
vito:
```suggestion
				It("asks the user to specify a target or log in", func() {
```

xtremerui:
@vito could you review this one again?

vito:
Gonna close this out since it looks like the commit history got a little borked. Feel free to resubmit!

PR #4175, opened 2019-07-26
Minor refactor

xtreme-sameer-vohra:
- task_step - move event processing block closer to consumer
- worker/client.go - avoid hoisting to improve readability


ddadlani:
@xtreme-sameer-vohra FYI don't forget to run `git commit -s` instead of `git commit` so that we don't get DCO issues when trying to merge in to master :)

For this PR I manually set it to passing since reverting it and then rebasing is a pain due to how long we've left #4109 open.

PR #4182, opened 2019-07-29
backport zstd memory leak fix

pivotal-jamie-klassen:
planning to ship this as 5.4.1.


pivotal-jamie-klassen:
@kcmannem said i could do this. the release/5.4.x pipeline will prevent us from shooting ourselves in the foot.

atc/exec/get_step.go
```diff
@@ -301,13 +301,31 @@ func streamFileHelper(s interface {
...
+}
+
+func (frc fileReadMultiCloser) Read(p []byte) (n int, err error) {
+	return frc.reader.Read(p)
+}
+
+func (frc fileReadMultiCloser) Close() error {
+	for _, closer := range frc.closers {
+		err := closer.Close()
+		if err != nil {
```
dprotaso: You may want to close all the `io.Closers` before returning an error

kcmannem: thanks for the catch.

PR #4202, opened 2019-07-31
lidar: tracking checks in the database

### LINKED ISSUE #3788: Resource check queue

pivotal-jwinters:
fixes #3788 

This adds a new `lidar` package which changes the way resource checks happen. The responsibility of checking resources is now split between two components: `scanner`, `checker`. The scanner will iterate through every resource (and its corresponding type), and add checks to the `checks` table (if necessary). The `checker` will then run any checks that end up in this table. If you hit the check resource or check resource type api, we skip the `scanner` component and insert a check directly into the database, which then gets picked up by the `checker`. 

This feature can be turned on/off through the `CONCOURSE_ENABLE_LIDAR` flag. However, this flag will only toggle the internal resource scanning mechanism. Moving forward all API requests will be tracked in the `checks` table regardless of whether or not this feature is enabled. ~This means that we've had to change the default behaviour of fly. By default `fly check-resource` and `fly check-resource-type` are now asynchronous. You can make them synchronous by specifying the `--watch` flag. By default a resource will also now fail fast if its parent type does not have a version. You can change this behaviour by specifying the `--recursive` flag which will force a check of all parent types and wait for the parent checks to finish. If you want fly to behave the way it used to, you should specify both these flags.~ The default behaviour for fly is back to normal. Instead you can make your check `--async` and `--shallow` as desired.

An interesting point to note is `checks` are now very similar to `builds`. They have a plan. The plan contains a new `exec.CheckStep`, which is executed through the `engine.Engine`. In the future it probably makes sense for these two concepts to continue to converge, which sounds neat!

Also worth noting is that concourse performs **A LOT** of checks (like a lot a lot). This means the checks table will tend to grow very quickly. Be default checks get gc'ed every 6 hrs, but this interval can be configured by specifying a `CONCOURSE_GC_CHECK_RECYCLE_PERIOD`.


kcmannem:
Elon Musk: “Anyone relying on lidar is doomed.” https://arstechnica.com/cars/2019/08/elon-musk-says-driverless-cars-dont-need-lidar-experts-arent-so-sure/

pivotal-jwinters:
> Elon Musk: “Anyone relying on lidar is doomed.” https://arstechnica.com/cars/2019/08/elon-musk-says-driverless-cars-dont-need-lidar-experts-arent-so-sure/

“Experts: Maybe not“

kcmannem:
@pivotal-jwinters show him whats up

atc/engine/builder/builder.go
```diff
@@ -18,7 +17,8 @@ const supportedSchema = "exec.v2"
 type StepFactory interface {
 	GetStep(atc.Plan, exec.StepMetadata, db.ContainerMetadata, exec.GetDelegate) exec.Step
 	PutStep(atc.Plan, exec.StepMetadata, db.ContainerMetadata, exec.PutDelegate) exec.Step
-	TaskStep(atc.Plan, exec.StepMetadata, db.ContainerMetadata, exec.TaskDelegate, lock.LockFactory) exec.Step
```
ddadlani: why are we removing the lockFactory from TaskStep? We need it for `limit-active-tasks`.

pivotal-jwinters: We don't need to pass it in because the `stepFactory` already has a `lockFactory`.

https://github.com/concourse/concourse/pull/4202/files#diff-7e37c202e7044cc08233052392ecdb38R146

ddadlani: ah ok, missed that, thanks 👍

vito:
We're on the verge of shipping v5.5 so let's just hold off on merging until it's out.

atc/lidar/scanner.go
```diff
@@ -0,0 +1,134 @@
...
+
+	checkFactory         db.CheckFactory
+	secrets              creds.Secrets
+	defaultCheckTimeout  time.Duration
+	defaultCheckInterval time.Duration
+}
+
+func (s *scanner) Run(ctx context.Context) error {
+
+	lock, acquired, err := s.checkFactory.AcquireScanningLock(s.logger)
```
clarafu: More of a question, will having a scanning lock make it so that only one ATC can run scans on all resources at a time? And this makes it so that we can avoid having locks on the resource config?

atc/db/check_factory.go
```diff
@@ -0,0 +1,360 @@
...
+			resourceConfigScopeID,
+			schema,
+			CheckStatusStarted,
+			manuallyTriggered,
+			encryptedPayload,
+			nonce,
+			metadata,
+		).
+		Suffix(`
+			ON CONFLICT DO NOTHING
```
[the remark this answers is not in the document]
xtremerui: The conflict would only happen when it matches `CREATE UNIQUE INDEX resource_config_scope_id_key ON checks (resource_config_scope_id) WHERE status = 'started' AND manually_triggered = false;`

From an api request, the `manually_triggered` is `true` it will never error so one could create as many checks as he wants for same resource.

atc/lidar/scanner.go
pivotal-jwinters: The scanning lock only guards against multiple atcs creating checks. We still need the resource config locks to guard against multiple atcs running the actual check.

PR #4203, opened 2019-07-31
WIP - limit the number of concurrent parallel streams to 5

ddadlani:
This is an attempt to fix flakiness on Prod as seen in https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/bosh-smoke/builds/523, https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/bosh-prod-deploy/builds/294 and https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/k8s-topgun/builds/1515. 

We noticed that there was really high web network load at the times that these failures occurred (~1Gbps) so we're limiting the number of parallel streams to 5 (hardcoded) to see if that mitigates this.

### DIFF
--- a/atc/worker/worker.go
+++ b/atc/worker/worker.go
@@ -588,6 +588,8 @@ func (worker *gardenWorker) cloneRemoteVolumes(
 	// 		 this will only cancel the groups goroutines but not the requests.
 	g, _ := errgroup.WithContext(ctx)
 
+	maxInFlight := make(chan int, 5)
+
 	for i, nonLocalInput := range nonLocals {
 		// this is to ensure each go func gets its own non changing copy of the iterator
 		i, nonLocalInput := i, nonLocalInput
@@ -609,9 +611,13 @@ func (worker *gardenWorker) cloneRemoteVolumes(
 			"dest-worker": inputVolume.WorkerName(),
 		}
 
+
+		maxInFlight <- 1
+
 		g.Go(func() error {
 			err = nonLocalInput.desiredArtifact.StreamTo(logger.Session("stream-to", destData), inputVolume)
 			if err != nil {
+				<-maxInFlight
 				return err
 			}
 
@@ -620,9 +626,12 @@ func (worker *gardenWorker) cloneRemoteVolumes(
 				MountPath: nonLocalInput.desiredMountPath,
 			}
 
+			<-maxInFlight
 			return nil
 		})
 	}
+
+
 	if err := g.Wait(); err != nil {
 		return nil, err
 	}


vito:
:thinking: kinda scary that we get such weird failures with high parallelism. is there a theory as to what concretely causes it? are requests failing and then not being retried because stuff was partially streamed or something?

cirocosta:
This reminded me of something that I've done in the past with Docker to avoid
the docker daemon running out of resources 🤔 

Rather than letting `dockerd` have the same "preference" when it comes to
getting shares of CPU, we'd give it a higher priority (`cpushares=1024`),
leaving all of the other containers in the machine be owned by a separate
(sibling) cgroup with lower priority (`cpushares=768`) so that `dockerd`.

Swap `dockerd` by `concourse`, and that seems like a reasonable thing to do to
avoid the starvation, I guess 🤔 

wdyt?

kcmannem:
I want to wait a bit longer before we merge this in, see if this is still happening. It also masks the problem and can't debug more.

pivotal-jamie-klassen:
@kcmannem how do you feel about this now?

kcmannem:
@pivotal-jamie-klassen we don't have any other suspects atm but we don't have any concrete evidence either. I don't think we should block 5.4.1 on this though. I haven't seen anyone in the community complain about this. I've only seen someone running on PKS internally complain but it might not be caused by this issue.

atc/worker/worker.go
```diff
@@ -609,9 +611,13 @@ func (worker *gardenWorker) cloneRemoteVolumes(
 			"dest-worker": inputVolume.WorkerName(),
 		}
 
+
+		maxInFlight <- 1
+
 		g.Go(func() error {
```
xtreme-sameer-vohra: while, this PR didn't result in the change, is there a reason why the returned `error` isn't used ?

kcmannem: g.Go() manages the errors returned by all the go routines. When one of the goroutines returns an error. It is return by g.Wait() as seen in line 635. We use the error there.

xtreme-sameer-vohra: nvm, I just saw the impl of `g.Go` & `g.Wait`

kcmannem: an errgroup is basically a wait group that waits on errors instead of .Done()s to be called on a sync.waitgroup.

atc/worker/worker.go
```diff
@@ -609,9 +611,13 @@ func (worker *gardenWorker) cloneRemoteVolumes(
 			"dest-worker": inputVolume.WorkerName(),
 		}
 
+
+		maxInFlight <- 1
```
cirocosta: Hey, should this be here? I wonder if it should instead be right in the beginning of the `for` loop 🤔  

It seems to me that having it here, it feels like we'd end up with, e.g., 6 volumes created, while only 5 streaming ops go on (leaving that volume there wait for its turn). 

Although that might not be a big of a deal, it seems to me that "reserving access" right in the beginning would make it more "even" (5 volumes, while 5 ops go on).

(unless that was really the desire - having a "pre-created" volume there just waiting for its turn to get into the streaming phase - which I think seems like a valid reasoning).

thanks!!

kcmannem:
I'm going to close this pr. This was here on standby incase we found concrete evidence that stressing the network was causing problems. We haven't and I'm less convinced now then at the start that this is the root cause.

PR #4204, opened 2019-08-01
[WIP] Cleanup failed volume on pks

zoetian:
Trying to address: https://github.com/pivotal/concourse-ops/issues/110


cirocosta:
hey @taylorsilva , 

With the new PKS installation that we have, do you think this can be closed?

thx!

taylorsilva:
I think it might be. Going to try quickly deploying and deleting, I'll see if the PV's disappear.

taylorsilva:
PVC's and PV's get cleaned up now. They don't hang around or need to be manually deleted.

PR #4208, opened 2019-08-01
clearer errors and logging for limit-active-tasks

ddadlani:
As part of merging in changes from `limit-active-tasks` and refactoring the task step, we accidentally introduced an error in the logs where we tried to decrease the number of active tasks even if we were not using `limit-active-tasks`.

This PR fixes that, as well as cleaning up other logging and naming for better clarity.

### DIFF
--- a/atc/db/lock/lock.go
+++ b/atc/db/lock/lock.go
@@ -20,7 +20,7 @@ const (
 	LockTypeVolumeCreating
 	LockTypeContainerCreating
 	LockTypeDatabaseMigration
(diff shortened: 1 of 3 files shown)


aledeganopix4d:
@ddadlani Thank you so much for improving this.
I have a question: when I originally implemented the `decrease_active_task` function I thought a lot about doing so within the lock or not.
In the end I couldn't find any good reason to do it serially so I didn't.
Can I ask you what do you anticipate the problem could be by not doing so serially?
Thanks!

PR #4188, opened 2019-07-30
feat[gc/build-log-collector]: Added minimum_succeeded_builds to build_log_retention

evanchaoli:
When minimum_succeeded_builds is specified, at least such success builds' logs will
be retained, which will be helpful for developing and debugging a job.

This commit changes the original algorithm of build log collector a little bit:
- no longer specially handle build id 1
- use ReapTime field to decide if a build's log has been reaped
- FirstLoggedBuildID now points to the first build whoes log is not reapped

Unit tests are updated.

If some user requests to support min_fail_builds in future, it would be easy
to support.

Resove #4026
Close #4188

Issue #4017, opened 2019-06-14
Try the best to pass failed resource to on_failure and ensure steps.


evanchaoli:
## What challenge are you facing?

We are trying to implement a gating logic like this:

We develop a resource type to do code scan, based on the scan result and predefined gating criteria, the resource type decides if pass or fail the build. If gating fails, we also want to notify (email or slack) developers the scan result, so that we know why scan fails.

Pipeline code is like the following:

     resource_types:
     - name: CodeScan
       repository: ...

     resources:
     - name: scanner
        type: CodeScan

     jobs:
     - name: job1
       plan:
         ...
         - put: scanner # do scan in put, 
                       # generate scan result file in the following implied get,
                       # and gating (pass or fail) in get
           on_failure:
                put: email
                params:
                    email_body: scanner/report_file

The problem is, if gating fails (implied get fails), in the `on_failure` step, email resource won't have access to 'scanner' resource, so that email resource cannot send scan report out, though scan report has been generated.

## What would make this better?

Our current solution is to create two resource types, one does code scan, the other does gating. So that when gating fails, email resource can still get scan result file from the scanner resource. But using two resources to archive that is kinda overhead.

I think Concourse may try the best to pass failed resource to `on_failure` and `ensure` steps. So that error handler will not only know job fails, but also possible to know why job fails.

## Are you interested in implementing this yourself?

Not for now.

evanchaoli:
I have verified that, if a task followed by a on_failure, even if the task fails (exit non-zero), on_failure will still have access the task's outputs. But if a put step's implied get fails, on_failure will have no access the failed resource.

To ensure backward compatibility, I added a flag in GetStep, when the flag is set, even if resource.Get fails, it will still register the resource, so that on_failure step will have access to the failed resource. And the flag is only set by OnFailureStep. The change is done by PR #4212.

PR #4212, opened 2019-08-02
feat[exec]: enable on_failure to access failed resource

evanchaoli:
Without this feature, when a put/get follows an on_failure step, if put/get fails, on_failure will have no access to the failed resource, so that on_failure step has no idea why it failed. This feature enables on_failure to access the failed resource.

### DIFF
--- a/atc/exec/on_failure.go
+++ b/atc/exec/on_failure.go
@@ -26,6 +26,8 @@ func OnFailure(firstStep Step, secondStep Step) OnFailureStep {
 // If the first step fails (that is, its Success result is false), the second
 // step is executed. If the second step errors, its error is returned.
 func (o OnFailureStep) Run(ctx context.Context, state RunState) error {
+	o.updateGetStep(o.step)
+
 	err := o.step.Run(ctx, state)
 	if err != nil {
 		return err
@@ -42,4 +44,34 @@ func (o OnFailureStep) Run(ctx context.Context, state RunState) error {
 // completed successfully.
 func (o OnFailureStep) Succeeded() bool {
 	return o.step.Succeeded()
+}
+
+// Set GetStep's flag "registerUponFailure".
+func (o OnFailureStep) updateGetStep(step Step) {
+	switch step.(type) {
+	case AggregateStep:
+		for _, s := range ([]Step)(step.(AggregateStep)) {
+			o.updateGetStep(s)
+		}
+	case InParallelStep:
+		for _, s := range step.(InParallelStep).steps {
+			o.updateGetStep(s)
+		}
+	case LogErrorStep:
+		// Get is actually wrapped in a LogErrorStep, so we need to handle
+		// LogErrorStep.
+		o.updateGetStep(step.(LogErrorStep).Step)
+	case *GetStep:
+		step.(*GetStep).SetRegisterUponFailure(true)
+	case OnSuccessStep:
+		// For Put's implied Get, they are wired by an OnSuccessStep, Put as
+		// the first step, implied Get as the second step. That's reason why
+		// we need to handle both "step" and "hook".
+		o.updateGetStep(step.(OnSuccessStep).step)
+		o.updateGetStep(step.(OnSuccessStep).hook)
+	case *TimeoutStep:
+		o.updateGetStep(step.(*TimeoutStep).step)
+	case TryStep:
+		o.updateGetStep(step.(TryStep).step)
+	}
 }
--- a/atc/resource/resource_get.go
+++ b/atc/resource/resource_get.go
@@ -33,7 +33,9 @@ func (resource *resource) Get(
 		true,
 	)
 	if err != nil {
-		return nil, err
+		vs := NewGetVersionedSource(volume, version, []atc.MetadataField{})
+		version["_err"] = err.Error()
+		return vs, err
 	}
 
 	return NewGetVersionedSource(volume, vr.Version, vr.Metadata), nil
--- a/atc/resource/resource_instance_fetch_source.go
+++ b/atc/resource/resource_instance_fetch_source.go
@@ -172,25 +172,25 @@ func (s *resourceInstanceFetchSource) Create(ctx context.Context) (VersionedSour
 	)
 	if err != nil {
 		sLog.Error("failed-to-fetch-resource", err)
-		return nil, err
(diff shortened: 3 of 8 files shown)


evanchaoli:
This PR will solve issue #4017.

atc/resource/resource_instance_fetch_source.go
```diff
@@ -172,25 +172,25 @@ func (s *resourceInstanceFetchSource) Create(ctx context.Context) (VersionedSour
 	)
 	if err != nil {
 		sLog.Error("failed-to-fetch-resource", err)
-		return nil, err
+		return versionedSource, err
```
xtremerui: Is there a case here that both versionedSource and err are not nil?

Same for the rest of 3 changes of returning versionedSource instead of nil, is there a specific reason for that?

xtremerui: Is there a case that both versionedSource and err are not nil here?

atc/exec/on_failure.go
```diff
@@ -43,3 +45,33 @@ func (o OnFailureStep) Run(ctx context.Context, state RunState) error {
 func (o OnFailureStep) Succeeded() bool {
 	return o.step.Succeeded()
 }
+
+// Set GetStep's flag "registerUponFailure".
+func (o OnFailureStep) updateGetStep(step Step) {
```
vito: I don't really like this; seems like it would be constant maintenance as we add more step types.

atc/resource/resource_get.go
```diff
@@ -33,7 +33,9 @@ func (resource *resource) Get(
 		true,
 	)
 	if err != nil {
-		return nil, err
+		vs := NewGetVersionedSource(volume, version, []atc.MetadataField{})
+		version["_err"] = err.Error()
```
vito: This seems a little fishy.

evanchaoli:
> Hmm sorry but I don't really like this. Can you provide an example of using this? Maybe there's some other way to do what you're trying to do?

@vito  Thanks for the review. 

Typically, a resource's `put` is used to trigger some operation on a third party system, then implied `get` will fetch operation result (a set of output files) back.

There is a use case that, operation on third party system fails with some error info (files) generated so that the resource won't fail the `put`, instead, fails the implied `get` after `get` fetch error files back. Ideally, if `put` has an `on_failure`, then the resource will be accessed in `on_failure`, so that steps in `on_failure`, such as email notification may get error files from the resource and email the error to users.

But as the current Concourse implementation, once `put` or `get` fails, `on_failure` will have no access to the failed resource, thus not able to handle failure with detailed information.

This PR is to passing a failed resource to `on_failure`. Does it make sense? Is there any alternative solution?

vito:
Ah, OK, so you have assets in the failed resource container that you want to use to annotate a notification?

Would it be possible to pull the heavy lifting out into a task and use the resource solely for the actual push/pull side effects? That way your notifications would at least have access to the task outputs.

Even with this fix, I don't think it will help you if the `put` fails - in that case the `get` won't even run.

In general I'm hesitant to complicate any of the internals for this, especially if the prime focus is for notifications, since we're already planning out a whole different approach for notifications. Right now all I have to show for it is a pretty empty RFC (https://github.com/concourse/rfcs/pull/28) and a mention on the [v10 roadmap](https://blog.concourse-ci.org/core-roadmap-towards-v10/#rfc-28-notification-resources).

evanchaoli:
@vito Using a task will cause all users to write long and duplicate code in their pipelines, that's reason why we wrapper such operation as a resource. As a resource, a pipeline just simply make a `put` on the resource.

Yes, you are correct. If `put` fails, then the resource will not get anything, that's reason why I said "try best" to expose resource to `on_failure`. At least error handler will detect if the resource has something to investigate.

Let me give the other use case. We have a resource type that does code coverage scan. With the scan, we also want to do gating, like if code coverage is below 80%, fail the pipeline. Our original design is like, `put` performs code scan, implied `get` generates scan report, then check coverage rate, if below N%, fails. After we implemented that logic, we realized that, after `get` fails, `on_failure` and `ensure` have no access to the scan resource anymore. So that we have to introduce a separate gating resource type. This way, a pipeline has to use two resources to archive scan-gating. If this feature were there, then one scan resource is enough.

vito:
> @vito Using a task will cause all users to write long and duplicate code in their pipelines, that's reason why we wrapper such operation as a resource. As a resource, a pipeline just simply make a `put` on the resource.

Using a task doesn't mean you have to duplicate code. Tasks are just as easy to re-use as resources, it turns out. For example, the [`builder` task](https://github.com/concourse/builder-task) is pulled in via the `git` resource and used in our pipelines [like so](https://github.com/concourse/ci/blob/bd1b200ff44869926e1d3b31bea4a96a6ec2707f/pipelines/concourse.yml#L78-L82).

This ends up being approximately the same amount of work as using a custom resource type, while having the added benefit of tracking the task as a resource which you can version and notice when it changes.

In general I wouldn't recommend thinking of resources as a way to package and execute code. They are meant to have actual resource semantics, i.e. versioning external state.

evanchaoli:
OK, if you really don't like this, I'm ok to drop this PR. By the way, please help review my other PR #4188.

evanchaoli:
> In general I wouldn't recommend thinking of resources as a way to package and execute code. They are meant to have actual resource semantics, i.e. versioning external state.

In the use case I described above, a scan actually generates a version (a version of scan result). So it meets criteria of a resource.

I'm just curious, with this in mind, is Concourse going to deprecate those put-only resource types, like Email, Slack, and so on?

vito:
> I'm just curious, with this in mind, is Concourse going to deprecate those put-only resource types, like Email, Slack, and so on?

Once the resource interface is generalized (concourse/rfcs#24) I plan to support these in a way that doesn't shoehorn them into the 'versioned artifacts' workflow, which is the only workflow that Concourse really supports right now.

Instead of sprinkling `put`s all over your jobs, you would annotate your artifact resource with a notification resource, and Concourse would automatically run the notification resource's `put` action as the annotated resource is used in a pipeline.

> But if you strongly dislike it, I'm ok to drop this PR.

I appreciate the time put into it but I think it'd be best if we focus on the long-term goals for notifications instead. :slightly_smiling_face: Thanks for understanding.

PR #4215, opened 2019-08-02
Remove example dir

deniseyu:
The examples were moved out of this repo and into the docs a while back.

The pipeline manifest in this repo should not be used as a canonical pipeline, 
and examples of the time resource and parallel jobs exist in the documentation


README.md
```diff
@@ -80,8 +80,8 @@ Concourse instance to the next.
 
 ### Learn More
 
-* The [Official Site](https://concourse-ci.org) for documentation and
-  reference material.
+* The [Official Site](https://concourse-ci.org) for documentation,
+  reference material, and example pipelines (which no longer live in this repository).
```
taylorsilva: Seems weird to allude to "examples" and not point people to where they live. DO we have examples that live somewhere that we can directly link to? Or is this a yet-to-be-implemented thing?

deniseyu: It refers to this page: https://concourse-ci.org/examples.html 

It's in the top nav bar so I didn't link it, but can add the link if that would be better.

PR #4221, opened 2019-08-06
[bugfix] Do not increase active_tasks if the container is already present.

aledeganopix4d:
This fixes a bug where the active_tasks counter (used in the limit-active-tasks strategy)
would be increased even if the container running the task is already present in the worker.

The above bug can manifest, for instance, if the ATC restarts.

To detect if a container is already present on a worker a small helper function is added
to Pool and queried it only when usign the limit-active-tasks strategy.

A test case is added to ensure the active_tasks counter is not increased when the container
is already present on the worker.

@xtreme-sameer-vohra first noticed this issue while reviewing #4118.
If this PR is accepted I'll also remove from the docs the warning that workers needs to be restarted to reset the counter.

### DIFF
--- a/atc/worker/client.go
+++ b/atc/worker/client.go
@@ -255,10 +255,11 @@ func (client *client) chooseTaskWorker(
 	outputWriter io.Writer,
 ) (Worker, error) {
 	var (
-		chosenWorker    Worker
-		activeTasksLock lock.Lock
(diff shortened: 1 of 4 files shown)


xtreme-sameer-vohra:
Thanks for the PR @aledeganopix4d

PR #4229, opened 2019-08-06
atc: fix k8s secret fetching

vito:
also backfilled unit tests. the k8s library has a whole test fake system - pretty nice!

### DIFF
--- a/atc/creds/kubernetes/kubernetes.go
+++ b/atc/creds/kubernetes/kubernetes.go
@@ -9,14 +9,15 @@ import (
 	"github.com/concourse/concourse/atc/creds"
 
 	v1 "k8s.io/api/core/v1"
-	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
 type Kubernetes struct {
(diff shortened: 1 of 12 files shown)


atc/creds/kubernetes/kubernetes.go
```diff
@@ -54,7 +54,7 @@ func (k Kubernetes) Get(secretPath string) (interface{}, *time.Time, bool, error
 		return k.getValueFromSecret(secret)
 	}
 
-	k.logger.Info("kubernetes secret not found", lager.Data{
+	k.logger.Info("kubernetes=secret-not-found", lager.Data{
```
vito: fudge

pivotal-jamie-klassen:
To put it another way, I find that there's a bit of mental load involved in understanding how this solution (and it's associated tests) should fix the k8s-topgun failures we've been seeing.

vito:
@pivotal-jamie-klassen Good idea for the tests! I'll take a swing at that. 

The `Get` interface is the interface that was expected of the previously-external templating package we used. We use our own `vars` package now which is a stripped-down version. Now that we control all the code we can change it however we want. Given our use case we might not even need arbitrary-depth traversal; it could probably just be a secret name and optional field. I'm not really sure I want to pull on that thread as part of this fix though as it would affect all credential manager implementations. :sweat_smile:

vito:
@pivotal-jamie-klassen Ok, so I ended up just rewriting the whole test suite to use table-style tests. We might be able to extract a common harness for this across all credential managers somehow. The only thing that changes is the setup. :thinking: 

Also instead of renaming that `k` var I renamed `Kubernetes` to `Secrets` since it's already in the `kubernetes` package and specifically implements the `creds.Secrets` interface.

Issue #4200, opened 2019-07-31
High CPU on ATC due to lock acquisition from db


ebilling:
## Bug Report

We have experienced higher than expected cpu on the concourse web application for a long time and believe it is the same issue raised in https://github.com/concourse/concourse/issues/2346.  However, after running a trace and looking carefully over the code, I believe the issue is caused by too few open connections to the DB for the lock manager when using TLS connections. (see Additional Context for trace)

I tracked this down to a separate connection pool for the lock and the setup specifies only 1 max idle and max open connection in _RunCommand.constructLockConn_.  I am going to attempt to raise the number of allowed open and idle connections and see if this has a positive impact.

## Additional Context

(pprof) top 20
Showing nodes accounting for 0, 0% of 24s total
Dropped 666 nodes (cum <= 0.12s)
Showing top 20 nodes out of 298
      flat  flat%   sum%        cum   cum%
         0     0%     0%     17.01s 70.88%  github.com/tedsuo/ifrit.(*process).run
         0     0%     0%     14.51s 60.46%  github.com/concourse/concourse/atc/scheduler.(*Runner).Run
         0     0%     0%     14.49s 60.38%  github.com/concourse/concourse/atc/scheduler.(*Runner).tick
         0     0%     0%     12.22s 50.92%  github.com/cenkalti/backoff.Retry
         0     0%     0%     12.22s 50.92%  github.com/cenkalti/backoff.RetryNotify
         **0     0%     0%     12.21s 50.88%  database/sql.(*DB).conn
         0     0%     0%     12.20s 50.83%  database/sql.dsnConnector.Connect**
         0     0%     0%     12.20s 50.83%  github.com/concourse/concourse/atc/db.(*connectionRetryingDriver).Open
         0     0%     0%     12.20s 50.83%  github.com/concourse/concourse/atc/db.(*connectionRetryingDriver).Open.func1
         0     0%     0%     12.20s 50.83%  github.com/lib/pq.DialOpen
         0     0%     0%     12.17s 50.71%  github.com/lib/pq.(*conn).ssl
         0     0%     0%     12.07s 50.29%  github.com/lib/pq.ssl.func1
         0     0%     0%     12.07s 50.29%  github.com/lib/pq.sslVerifyCertificateAuthority
         0     0%     0%     10.30s 42.92%  github.com/concourse/concourse/atc/db.(*db).Exec
[...]

PR #4231, opened 2019-08-07
[#4200] reduces overhead of TLS db connections

ebilling:
From my testing with at the level of 6k requests I'm seeing a drop in CPU of approximately 30% by re-using database connections rather than letting them get reaped when they go idle.  For TLS connections re-establishing the connection is very expensive.


ebilling:
Sorry, picked up some changes I wasn't planning to take.  Will trim and resubmit.

vito:
FWIW I think you can also just `git push -f` and the PR will auto-update.

ebilling:
I took care of it.  Which branch would you prefer I do the pull request to?  Master or release/4.5.x?

vito:
@ebilling Late reply, but always `master`. :)

Issue #4152, opened 2019-07-23
Perform GC in parallel


vito:
In #3602 we noticed that during a `VACUUM` container counts would climb up and max out the workers.

This seems to imply that GC was held up somewhere. If containers are being created, queries are running to create those, so it's not like the whole table is locked up.

We currently run all of our collectors one after another: https://github.com/concourse/concourse/blob/8c5d0dc1a9c658b7ac8e978716d9d4d008707281/atc/gc/collector.go#L51

Maybe one of the collectors was blocked waiting on the `VACUUM`? Should we run these in parallel on independent loops instead?

Can we confirm the slowdown by observing logs on the `algorithm` environment? (This is time-sensitive, since they don't currently aggregate anywhere.)

vito:
Updating the title. The investigation led us to parallelizing GC on the algorithm branch, which proved pretty effective at fixing this. The main problem was that some collectors would just take a long time, and while they were GC'ing nothing else could GC.

There was a solid theory to the order that we would do GC sequentially, but doing them in parallel should be fine too - it just may take multiple ticks for objects to expire at times. The improved resiliency seems worth the trade-off.

Side note: we did notice that, on occasion, a single worker would still have its container count spike up. This seems to be a separate issue. With our fix implemented we never saw *all* workers spike up again, so it's still a valid fix. We dug into it and didn't really find a root cause. Container destroying was just taking longer - up to a couple minutes in some cases, whereas typically it takes a second or two. We don't have an explanation for the slowdown, but it was obvious that if container destroying was taking that long that it would snowball - there were 131 containers to GC in one tick, and if each one takes a couple minutes, it's not fast enough to overcome the rate of container creation in our test environment.

vito:
Also: would longer/shorter intervals make sense for some garbage-collectors vs. others?

Issue #4025, opened 2019-06-17
Give all of our components their own connection pools


vito:
We found that at large scale we would see GC starved of connections because pipeline scheduling was using them all.

Pipeline scheduling will use N connections; at least one per pipeline. GC only uses a handful, and this doesn't really change with more work. But if GC isn't able to acquire a connection, really bad things happen (mainly, workers fill up with containers/volumes).

Make sure that each pool also has a reasonable `MaxIdle` configured. The default is `2`, meaning that if a spike of pipeline scheduling a ton of connections may need to be opened. This is very costly if the database is configured with TLS. In #4232 we decided `MaxIdle` being half of `MaxOpen` is a reasonable arbitrary decision.

PR #4232, opened 2019-08-07
[#4200] prevent reconnect floods to Postgres

ebilling:
After many profiles, it became clear that at high transaction levels the system was spending substantial amounts of it's time reaping and re-establishing TLS connections in order to grab locks or run queries against Postgres.  

This change resulted in a 30% reduction in CPU for our web processes.  The dominate calls now are around marshaling and un-marshaling JSON.

Signed-off-by: Eric Billingsley <eric@calculi.com>

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -98,6 +98,8 @@ type RunCommand struct {
 
 	Postgres flag.PostgresConfig `group:"PostgreSQL Configuration" namespace:"postgres"`
 
+	MaxOpenConnections int `long:"max-conns" description:"The maximum number of open connections for a conneciton pool." default:"32"`
+
 	CredentialManagement creds.CredentialManagementConfig `group:"Credential Management"`
 	CredentialManagers   creds.Managers
 
@@ -401,12 +403,12 @@ func (cmd *RunCommand) Runner(positionalArguments []string) (ifrit.Runner, error
 
 	lockFactory := lock.NewLockFactory(lockConn, metric.LogLockAcquired, metric.LogLockReleased)
 
-	apiConn, err := cmd.constructDBConn(retryingDriverName, logger, 32, "api", lockFactory)
+	apiConn, err := cmd.constructDBConn(retryingDriverName, logger, cmd.MaxOpenConnections, "api", lockFactory)
 	if err != nil {
 		return nil, err
 	}
 
-	backendConn, err := cmd.constructDBConn(retryingDriverName, logger, 32, "backend", lockFactory)
+	backendConn, err := cmd.constructDBConn(retryingDriverName, logger, cmd.MaxOpenConnections, "backend", lockFactory)
 	if err != nil {
 		return nil, err
 	}
@@ -1153,7 +1155,7 @@ func (cmd *RunCommand) configureMetrics(logger lager.Logger) error {
 func (cmd *RunCommand) constructDBConn(
 	driverName string,
 	logger lager.Logger,
-	maxIdleConns int,
+	maxConn int,
 	connectionName string,
 	lockFactory lock.LockFactory,
 ) (db.Conn, error) {
@@ -1172,8 +1174,8 @@ func (cmd *RunCommand) constructDBConn(
 	}
 
 	// Prepare
-	dbConn.SetMaxOpenConns(2 * maxIdleConns)
-	dbConn.SetMaxIdleConns(maxIdleConns)
+	dbConn.SetMaxOpenConns(maxConn)
+	dbConn.SetMaxIdleConns(maxConn / 2)
 
 	return dbConn, nil
 }
@@ -1188,8 +1190,7 @@ func (cmd *RunCommand) constructLockConn(driverName string) (*sql.DB, error) {
 		return nil, err
 	}
 
-	dbConn.SetMaxOpenConns(10)
-	dbConn.SetMaxIdleConns(5)
+	dbConn.SetMaxIdleConns(1)
 	dbConn.SetConnMaxLifetime(0)
 
 	return dbConn, nil
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -1153,7 +1153,7 @@ func (cmd *RunCommand) configureMetrics(logger lager.Logger) error {
 func (cmd *RunCommand) constructDBConn(
 	driverName string,
 	logger lager.Logger,
-	maxConn int,
+	maxIdleConns int,
 	connectionName string,
 	lockFactory lock.LockFactory,
 ) (db.Conn, error) {
@@ -1172,7 +1172,8 @@ func (cmd *RunCommand) constructDBConn(
 	}
 
 	// Prepare
-	dbConn.SetMaxOpenConns(maxConn)
+	dbConn.SetMaxOpenConns(2 * maxIdleConns)
+	dbConn.SetMaxIdleConns(maxIdleConns)
 
 	return dbConn, nil
 }
@@ -1187,8 +1188,8 @@ func (cmd *RunCommand) constructLockConn(driverName string) (*sql.DB, error) {
 		return nil, err
 	}
 
-	dbConn.SetMaxOpenConns(1)
-	dbConn.SetMaxIdleConns(1)
+	dbConn.SetMaxOpenConns(10)
+	dbConn.SetMaxIdleConns(5)
 	dbConn.SetConnMaxLifetime(0)
 
 	return dbConn, nil
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -1190,6 +1190,7 @@ func (cmd *RunCommand) constructLockConn(driverName string) (*sql.DB, error) {
 		return nil, err
 	}
 
+	dbConn.SetMaxOpenConns(1)
 	dbConn.SetMaxIdleConns(1)
 	dbConn.SetConnMaxLifetime(0)
 
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -98,7 +98,7 @@ type RunCommand struct {
 
 	Postgres flag.PostgresConfig `group:"PostgreSQL Configuration" namespace:"postgres"`
 
-	MaxOpenConnections int `long:"max-conns" description:"The maximum number of open connections for a conneciton pool." default:"32"`
+	MaxOpenConnections int `long:"max-conns" description:"The maximum number of open connections for a connection pool." default:"32"`
 
 	CredentialManagement creds.CredentialManagementConfig `group:"Credential Management"`
 	CredentialManagers   creds.Managers


atc/atccmd/command.go
```diff
@@ -1187,8 +1188,8 @@ func (cmd *RunCommand) constructLockConn(driverName string) (*sql.DB, error) {
 		return nil, err
 	}
 
-	dbConn.SetMaxOpenConns(1)
```
vito: The lock connection has to be forced to 1 because we use session-level locks, which are associated to a connection. If we pool them, locks will potentially be lost when a connection in the pool is closed.

I don't think this would be the source of a lot of churn; it should just be one connection that stays open the whole time. :thinking: We never checked though.

atc/atccmd/command.go
```diff
@@ -1172,7 +1172,8 @@ func (cmd *RunCommand) constructDBConn(
 	}
 
 	// Prepare
-	dbConn.SetMaxOpenConns(maxConn)
+	dbConn.SetMaxOpenConns(2 * maxIdleConns)
```
vito: Not sure if we should bump this outright since folks may have tuned their Postgres server-side connection limit accordingly. Doubling the effective pool size on all the `web` nodes could may be surprising especially with multiple `web` nodes. Could we just do `SetMaxIdleConns(maxConn / 2)` instead?

Also, FYI we might end up re-doing our connection pooling; recently we've discussed giving critical things like GC their own connection pool (#4025) and that may turn into having per-backend-component pools with sizes based on the component's priority/usage, instead of having one big pool for everything. For if/when we get to that, what's the main learning here so we don't repeat the original mistake? Is it just to have a bit more breathing room with idle connections?

Thanks!

atc/atccmd/command.go
ebilling: This is actually the source of the first bottleneck.  If you look at the first set of calls in #4200 you'll see that AcquireLock is the dominate call.  This turns out to be because there is a race condition between giving up the lock from a previous call, getting a connection to the free list, and establishing new connections.  We found that because the connection pool manager is operating in a separate thread, the system was never able to pull from the idle connection list, so it would establish a new connection.  Then it had hit the max open and had to free the existing.

atc/atccmd/command.go
ebilling: This doesn't double the pool size, it doubles the max size.  If the system needs fewer connections it never gets to the limit.  The limit should be on the idle connections, not the open.  

For #4025, what's the benefit of creating separate pools?  Segmentation of connections creates more opportunities for bottlenecking rather than allowing for free flow.  

My thought was that the "right" thing to do is to only set the max Idle and let the max open go to a very high limit.  That allows the system to scale up and flex, but keeps a minimum amount of connections open so you don't get stuck in open, use, reap, repeat flows.  That's what killed performance on TLS.  If you aren't using encryption to the DB, then this isn't a big deal.  The setup time is low. But with TLS, you spend many milliseconds going back and forth, checking certificate chains, etc.  It's more expensive than the request, so you really need to err on the side of re-using that resource.

atc/atccmd/command.go
vito: I'm surprised it would establish a new connection and do all the TLS stuff even when it knows it's reached `MaxOpenConns` (1), though. I would have expected it to just wait for the open one to go back to idle or something.

Looking at #4200, I see `AcquireSchedulingLock` but not `AcquireLock`. The time could just as easly be coming from this `Exec` call, which also shows up in the profile but is using the normal connection pool, not the lock pool:

https://github.com/concourse/concourse/blob/ca0fd4cb1ed0b619fd1e083b5b3c827ffb669e63/atc/db/pipeline.go#L917-L922

I see that you confirmed the fix was effective, but I wonder if it would have been just as effective without changing the lock connection settings. Maybe it's really just a matter of bumping the idle connections for the other connection pools?

If not, and the lock pool is really the culprit, we'll need to find a way to fix it without bumping the pool, because it really needs to be a single dedicated connection for the locking semantics to work correctly.

The code that uses the lock connection is here: https://github.com/concourse/concourse/blob/master/atc/db/lock/lock.go

Maybe it should use a [`driver.Conn`](https://golang.org/pkg/database/sql/driver/#Conn) instead? That would avoid the pooling altogether. We'd need a way to re-establish the connection if it breaks, though.

atc/atccmd/command.go
vito: The idea is to prevent work from taking up the entire shared connection pool and starving other mission-critical components like garbage collection because the pool has hit `MaxOpenConns`. This can cause GC to not run, which can cause containers to pile up on the workers and then everything starts failing.

Our thinking - and this is somewhat backed by what we've observed in our test environments - is that each internal component has some ballpark number of connections it needs in order to do its job.

GC, for example, only needs around 2 when run in serial, and around 10 when run in parallel (assuming #4152).

The pipeline scheduler, on the other hand, needs a lot more than that depending on how many jobs there are to schedule. It's fine for scheduling work to "steal" connections from other scheduling work - things will just be scheduled more slowly - but it's not OK for scheduling work to "steal" from GC.

Also keep in mind most Postgres servers have a default limit of 100 open connections. It's not enough to just set the max idle pool and let everyone grab as many connections as they want. Someone will be starved eventually, whether it's client-side or server-side. So the connection pool has to be rationed somewhat carefully.

It's a tricky problem, though; it's fundamentally difficult to coordinate these limits on both sides, and there are probably other deeper ways to improve this situation. We just want sane defaults for now so things can't go off the rails. So, our idea was to just figure out what these reasonable connection counts (or maybe just 'ratios') are for each component and tune the defaults accordingly.

Does that make sense? :slightly_smiling_face:

ebilling: But since you have multiple webs the connection limits you impose on any one of them doesn't prevent you from hitting the 100 total.  In any case, you'll need to increase the number of connections on the DB to support a large number of calls.  I think we are bumping into an inherit shortcoming of the connection pooling system.  The ability to get a connection to the next request and bypass the idle queue is the real underlying problem.

atc/atccmd/command.go
ebilling: Here is the trace from the original code: [https://drive.google.com/file/d/1Pobv2FcULJykqHq5nsaMYRZsX3hGCSZ6/view?usp=sharing](5.4.0), and here is the trace with the changes: [https://drive.google.com/file/d/1Ep3KM9X4Cpc4cCXmpFAB0_c5KK1XrgQG/view?usp=sharing](5.4.0+#4232)

You can see the delta.  I first concentrated on the issue of the `AcquireSchedulingLock `(misnamed above by me).  Once I removed that as a bottleneck, the actual queries came up as an issue.  How this is manifesting on the system is that we end up with well over 100 (sometimes as high as 200) TCP connections in TIME_WAIT, which means they have been closed and the system is waiting to allow the port to be available again.  

I think your idea of using the `driver.Conn`is probably the right one, but I have to wonder whether it makes more sense to actually use a single pool and set the max Idle to be a reasonably high number (20-40).  The re-use ratio should be much higher and the connection pooling should start working as designed.  I understand that the default configuration of Postgres is to limit the number of connections to 100, but with multiple webs, and the backoff methodology already in place, I think that it doesn't really matter.  The goal should be to have the maximum number of requests serviced within the limits of the environment, and in a high workload environment, you will have to raise that limit on Postgres.

atc/atccmd/command.go
vito: You're definitely right about that, but it's still better to have *some* limit rather than *no* limit, isn't it? Without any limits you have no hope of preventing GC from being starved, since you'll hit the server-side limit at some point just because you received a lot of API requests or had a spike in pipeline scheduling. We have to do what we can to prevent that from turning into a cascading failure.

If I have N ATCs and they have a theoretical max pool usage of 64 connections (as it is today), I need to configure my Postgres to have a minimum of N * 64 connection pool. We just set ours to 200 and call it a day. That affords 3 `web` nodes which is a pretty decent number for large-scale environments.

Also bear in mind that these pool limits were *added* because people kept hitting the server-side limit. We did a very coarse-grain bucketing of the limit by giving the API 32 and the "backend" 32. We just want to be a bit smarter about how that's divided up by default, since 'backend' isn't fine-grained enough to prevent GC from getting wedged by other backend work.

I wonder how other systems enforce this kind of thing. In my ideal world it would all be enforced by the server so we don't have multiple 'sources of truth' and it doesn't matter how many `web` nodes you have. In a cursory search it looks like you can have per-user connection limits, so maybe you could have a separate user for critical things, but that's a lot to configure. :thinking:

atc/atccmd/command.go
vito: Looking at the graph, it looks like it's all coming from the non-lock connection pool. The graph shows the cost going through a `*countingConn` which is only added here:

https://github.com/concourse/concourse/blob/cfc598920835c03f8c3b7c68dd60f03b6e14cc2c/atc/atccmd/command.go#L1166

We can't use a pool of any sort because the nature of the queries we use to acquire/release. When releasing, the query has to be run on the same connection that acquired the lock. We use advisory locks and avoid holding transactions open, which we've observed causes an increase in database CPU usage the longer they're held open. Our solution was to just have a single dedicated connection for locks.

ebilling: I thought (admittedly, I didn't look as closely as you have) that there were two sets of mutexes around the calls to `AcquireSchedulingLock`.  Why does the actual connection matter at that point?  The connection does not actually create the lock in any real way does it?  The Mutex is what actually protects the ability to update/hold the lock.  What I saw was that the connection was not actually being re-used anyway.  As I understood it, the `*countingConn.Exec`  is the wrapper around all of the DB connection requests.  The size of the timings going into it, show that initially the big issue was around   `AcquireSchedulingLock`.  Once I cleared that it moved as the core issue was really around the number of maxIdleConnections.

atc/atccmd/command.go
ebilling: That's exactly why I took the position of maxOpen = 2 x maxIdle.  The thing we really want to configure is maxIdle as GC is a bursty problem.  It comes in with a lot of requests all at once and if there aren't enough sitting on the Idle queue, then it just starts making new ones.  

Here is my proposal:

1. Create one thread pool with a MaxIdleConnections queue size that is configurable, defaulted to 20.  We then set MaxOpenConnections to be a separately configurable parameter that defaults to 2x MaxIdleConnections. (alternatively we could set MaxIdleConnections to be 1/2 MaxOpenConnections), but the behavior we really want is to allow bursting and backoff, IMHO.

2. Locking should rely on the mutex, not the connection, so it should use the same connection pool.  After all, as soon as you have 2 webs, the connection is not your savior.

This gives us the flexibility of configuration for specific environments and puts defaults in place that are reasonable.  Sound like a plan?

atc/atccmd/command.go
vito: The lock is created using the connection here:

https://github.com/concourse/concourse/blob/12b4d7cfc178e0f0639e4b5b899b4d2f3784b012/atc/db/lock/lock.go#L213

The additional locking is just to prevent the same `web` node from acquiring the same lock twice, since Postgres will allow the same connection to acquire the same lock multiple times as long as it's released each time. This isn't what we want; we want it to wait on the original lock to be released.

`*countingConn` is only used for the non-lock connections; there's no need to count connections for the lock connection as we enforce that there is always only one anyway.

The fact that `AcquireSchedulingLock` showed up in the profile so much might just be because that operation happens once every 10 seconds for every pipeline, and it's the first database operation performed on that interval. It might also be spiky, i.e. many pipelines are being scheduled around the same time, which would lead to a bunch of connections being created. Increasing the max idle conns beyond the default of 2 as you did would help ease that spikiness, but I think it's just a coincidence that it was a '*Lock' call, and that the issue doesn't lie in the lock connection pool.

atc/atccmd/command.go
vito: Minor clarification: GC isn't bursty *yet* - right now it uses at most one or two connections. I'll be more bursty with #4152 (~10 connections) but not nearly as bursty as pipeline scheduling (N connections; at least 1 per pipeline). This difference in expected connection usage is why we want to separate the scheduling pool from the GC pool and give them reasonable sizes so they're not competing for resources.

> 1\. Create one thread pool with a MaxIdleConnections queue size that is configurable, defaulted to 20.  We then set MaxOpenConnections to be a separately configurable parameter that defaults to 2x MaxIdleConnections. (alternatively we could set MaxIdleConnections to be 1/2 MaxOpenConnections), but the behavior we really want is to allow bursting and backoff, IMHO.

Makes sense. We're just trying to fix the churning connections, so setting `MaxIdle` seems like the simplest thing to do.

FWIW, I don't think we should change the effective default max open connections. I don't want to surprise any Concourse operators and cause server-side limits to suddenly be hit (due to an increase) or cause components to be starved of connections (due to a decrease) where they weren't before. We're only trying to fix the connection churn, so I don't think changing anything else is worth the risk, at least not right now. We can re-assess these magic numbers when we do per-component connection pools (#4025). I'll make a note in that issue to be sure to configure `MaxIdle` appropriately so this doesn't regress.

> 2\. Locking should rely on the mutex, not the connection, so it should use the same connection pool.  After all, as soon as you have 2 webs, the connection is not your savior.

As clarified in https://github.com/concourse/concourse/pull/4232#discussion_r311694326 locking *does* use the connection, and I think that'll need to stay how it is.

With all that said, this sounds like a super simple change (just `SetMaxIdleConns(maxConn / 2)`. With a lot of thought/words behind it. :sweat_smile:

atc/atccmd/command.go
ebilling: As an aside, the actual observations are that we are bouncing between 21 and 48 open connections to the DB.  I'm still struggling to understand why treating the connection pools as independent is buying anything here.  

The reason I believe that the lock connection pool is important is that it is used frequently, and the mutexes around it guarantee that there is a backlog of requests coming in to capture the next lock.  This seems to result in the problem where there are no idle connections yet on the queue, and with the maxOpenConnections and maxIdleConnections set to 1, it was causing the situation where the pool manager had no choice but to establish a new connection for every request.  That's not good.  A pool size of 2 may be enough, but being set at 1 guarantees a race condition that results in re-establishing connections.

atc/atccmd/command.go
ebilling: Ok.  Thanks for all of the help on this.  There are already three connection pools (lockConn, apiConn and backendConn).  If you are going to add another for gcConn and potentially others, then making this configurable could get a bit unwieldy.  My personal vote would be to allow configuration of the total pool size and use the pool for all use cases.  Would be nice if there was a prioritized pool, but...

How would you like me to proceed?  I have this working for our needs right now, and we are happy with the changes so far.  Would you like me to make any specific changes, like make the default MaxOpenConnections a configuration parameter?

atc/atccmd/command.go
vito: As far as I can tell, if you have max connections set to 1, it's never going to open another connection. That would directly go against the documentation.

This is pretty easy to verify. Here's a gist demonstrating the behavior: https://gist.github.com/vito/55966975903a77ec31c0f36ace15fcc0

If you have `MaxOpen` set to 1 and something needs a new connection, it waits for the connection to become available again. It's not going to make a new connection; the limit is set to 1, and that would be 2 open connections.

atc/atccmd/command.go
vito: I think what you've just pushed looks good. :slightly_smiling_face: 

The simplest possible thing for #4025 is probably to just have a flag for each component and spend a lot of time making sure we get the defaults right so most operators don't think about it (...until they need to, and then they have all the knobs they want).

Some components only need a certain amount of connections no matter what the scale is, so those are pretty easy. Others, like the API and pipeline scheduling, are based more on usage patterns, so it's hard to have a reasonable default. But hey, hardcoding 32 got us this far!

atc/atccmd/command.go
ebilling: Thanks. I dug through the code, played with the example and I agree with the conclusion.  I've moved the lock back to it's original state, and just made the the two changes: 1.) allow MaxOpenConnections to be configurable via a flag, with same current default and 2.) set MaxIdleConnections to be 1/2 of MaxOpenConnections.  This should fix the issue of reconnections being a major issue, and allow for changes for particular environments.  I'm testing again on my side right now.

atc/atccmd/command.go
```diff
@@ -98,6 +98,8 @@ type RunCommand struct {
 
 	Postgres flag.PostgresConfig `group:"PostgreSQL Configuration" namespace:"postgres"`
 
+	MaxOpenConnections int `long:"max-conns" description:"The maximum number of open connections for a conneciton pool." default:"32"`
```
analytically: typo

ebilling: Thanks for catching.  Fixed.

atc/atccmd/command.go
```diff
@@ -1187,7 +1190,6 @@ func (cmd *RunCommand) constructLockConn(driverName string) (*sql.DB, error) {
 		return nil, err
 	}
 
-	dbConn.SetMaxOpenConns(1)
```
vito: I think this PR is good to go once this line is restored. :+1:

ebilling: done

Issue #4193, opened 2019-07-30
[atc/api] Increase scope of Admin privilege


taylorsilva:
Epic #4192 
### Description/Context

We want to update the API handlers that care about what teams a user belongs to. The accessor.IsAuthorized() should always return true if the user is Admin. Any handlers that care about the teams that a user is part of should check if the user is an admin and instead perform their action for/with all teams in mind.

Tip: search for usages of `acc.TeamNames()`, those are the handlers you likely need to change

### Task List

- Investigate and **potentially modify** the following methods to behave properly for an admin user:
  - [x] Modify ListAllPipelines
  - [x] Modify ListAllJobs
  - [x] Modify ListBuilds
  - [x] Modify ListAllResources
  - [x] Modify ListWorkers

### Acceptance Criteria
- [x] No handler should return unauthorized if a user is `Admin`
- [x] Web UI displays all pipelines when a user is `Admin`
- [ ] ~~Fly CLI should be able to log into any team and perform any action~~
  - will work on this in the Fly CLI story

taylorsilva:
# Issue Status

- Created a branch [`super-admin-user`][branch] with some wip code
- Our current implementation idea is that each API handler will have to do something like what we find in the `SetTeam()` function:

https://github.com/concourse/concourse/blob/5824dfd73cba6a3d9a256a2317f45a92478aa2b4/atc/api/teamserver/set.go#L55-L66

- For testing, when trying to add to the existing test suite it's really easy to start faking everything by accident. We'll need to be careful about how we test Admin behaviour

Next Step: Focus on testing and implementing a single API handler. We started with the `ListAllPipeline()` handler. We think it'll be good to focus on implementing just this one handler and seeing what we learn from it.

[branch]: https://github.com/concourse/concourse/tree/super-admin-user

taylorsilva:
We used this bash script (placed in the root of the concourse repo) after doing `docker-compose up --build -d` to set-up a team and private pipeline. Helped with quickly testing how the web ui is behaving.

```bash
#!/bin/bash

fly -t local login -u test -p test -n main
fly -t local set-team -n team1 --local-user=guest
fly -t local sp -p hello -c examples/hello-world-every-minute.yml

fly -t local login -u guest -p guest -n team1
fly -t local sp -p hello-team1 -c examples/hello-world-every-minute.yml
```

Issue #4195, opened 2019-07-30
[fly] Confirm functionality for Admin user


taylorsilva:
### Description
Given the changes done in the API story (#4193) add some tests to ensure the fly cli functions as expected for an admin.

### Acceptance Criteria
- [ ] Admin is not part of any team (besides main) but can log into any team
- [ ] All fly actions work for the admin

pivotal-bin-ju:
**DEV NOTE**

All the available commands have been categorized on levels. such as: `pipeline`, `team`, etc.
Please see the note from: https://github.com/concourse/concourse/issues/4240#issuecomment-521427454

scottietremendous:
We should add fly integration tests per command.

taylorsilva:
### Update Aug 20th

Spent most of the day wrapping up #4240 
Realized a smell with the testing we do for `fly/integration`. We're faking all the server responses which means we have to spend time making sure we craft the correct response. It would be nice if we faked this with the API code instead of re-writing the API code within each test.

pivotal-jamie-klassen:
@taylorsilva great point about fly integration tests. this hand-rolled fake
server is definitely a holdover from the days when fly was in its own repo.

zoetian:
note:

We also added the test cases in `testflight/admin_priviliges_test.go`

taylorsilva:
@pausebreathefly attaching it here
[super admin fly commands scopes.xlsx](https://github.com/concourse/concourse/files/4731864/super.admin.fly.commands.scopes.xlsx)

Issue #4194, opened 2019-07-30
[Web] Confirm UI functionality for Admin user


taylorsilva:
Epic: #4192

### Description
Given the changes done in the API story (#4193) add some tests to ensure the web UI functions as expected for an admin.

### Branch
`4194-super-admin-web-ui`

### PR
https://github.com/concourse/concourse/pull/4319

### Acceptance Criteria

Things we can't but should be able to do in the web UI for teams admin is not a member of:

- [x] pause/unpause a pipeline
- [x] force a check resource
- [x] expose a pipeline

pivotal-jamie-klassen:
For the most part I'd hope that if we have confidence in the API piece, we'll get confidence around the UI for free. 

I have one thought about the 'Admin should be able to see all teams' point: Should there maybe be an extra tag on dashboard teams for the teams you can see purely because you're an admin (i.e. a new `Admin` variant [here](https://github.com/concourse/concourse/blob/master/web/elm/src/Dashboard/Group/Tag.elm#L12-L16))? I don't think its necessary since we can probably assume the admins know that they're admins, but this seems like the only suitable place to ask the question.

PR #4238, opened 2019-08-08
atc/api: Super admin user first steps

taylorsilva:
tl;dr: Modify endpoint behavior to return everything when the user is an `Admin` i.e. owner of the main team

- We modified the following methods to behave properly for an admin user:
  - ListAllPipelines
  - ListAllJobs
  - ListBuilds
  - ListAllResources
  - ListWorkers
- Added the appropriate tests.
- this is currently functional, that the admin user is able to see all the pipelines when logging in from the web ui.
- fly hits team endpoints, so still not working as per this PR, we will address that in #4195.
- Also, applying actions to pipelines in the web ui (unpause, expose, trigger job), this should be addressed in #4194. 

cc: @youssb 

fixes: #4193

### DIFF
--- a/atc/api/accessor/accessor.go
+++ b/atc/api/accessor/accessor.go
@@ -42,19 +42,23 @@ func (a *access) Claims() jwt.MapClaims {
 }
 
 func (a *access) IsAuthorized(team string) bool {
+	if a.IsAdmin() {
+		return true
+	}
 	for teamName, teamRoles := range a.TeamRoles() {
-		if teamName == team {
-			for _, teamRole := range teamRoles {
-				if a.HasPermission(teamRole) {
-					return true
-				}
+		if teamName != team {
+			continue
+		}
+		for _, teamRole := range teamRoles {
+			if a.hasPermission(teamRole) {
+				return true
 			}
 		}
 	}
 	return false
 }
 
-func (a *access) HasPermission(role string) bool {
+func (a *access) hasPermission(role string) bool {
 	switch requiredRoles[a.action] {
 	case "owner":
 		return role == "owner"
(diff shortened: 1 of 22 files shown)


atc/db/resource_factory.go
```diff
@@ -40,12 +43,28 @@ func (r *resourceFactory) VisibleResources(teamNames []string) ([]Resource, erro
...
+		RunWith(r.conn).
+		Query()
+	if err != nil {
+		return nil, err
+	}
+
+	return r.parseResources(rows)
+}
+
+func (r *resourceFactory) parseResources(resourceRows *sql.Rows) ([]Resource, error) {
```
pivotal-jwinters: Typically we would call this `scanResources` and it would be a package level function (instead of being tied to the resourceFactory). That way it can be reused in other places if needed. See `scanResource` as an example.

atc/api/accessor/accessor.go
```diff
@@ -42,19 +42,23 @@ func (a *access) Claims() jwt.MapClaims {
...
+			if a.hasPermission(teamRole) {
+				return true
 			}
 		}
 	}
 	return false
 }
 
-func (a *access) HasPermission(role string) bool {
+func (a *access) hasPermission(role string) bool {
```
zoetian: Just curious, do we have to change this to lower case? Is there some reason to do so?

cirocosta: IIRC the reasoning behind the change was that there were no uses for `HasPermissions` outside the scope of the package

YoussB:
@pivotal-jwinters, updated and rebased. Please tell me if there is anything that is missing.

YoussB:
@pivotal-jwinters Will you please reapprove it?

PR #4243, opened 2019-08-08
pass context to baggageclaimclient.StreamIn/Out calls

### LINKED ISSUE #4242: Pass context into StreamIn/Out calls in ATC.
convert it

kcmannem:
fixes #4242 

### DIFF
--- a/atc/exec/task_config_source.go
+++ b/atc/exec/task_config_source.go
@@ -80,9 +80,8 @@ func (configSource FileConfigSource) FetchConfig(logger lager.Logger, repo *arti
 		return atc.TaskConfig{}, UnknownArtifactSourceError{sourceName, configSource.ConfigPath}
 	}
 
-	// TODO: idk if we really want to pass a context in here. This is a very
-	// 		 shortlived stream. It would pollute the TaskConfigSource Interface
-	//		 as all the FetchConfigs will be need to have this passed in.
+	// This context is not passed down yet because it would pollute the 
+	// TaskConfigSource Interface as all the FetchConfigs will be need to have this passed in.
 	stream, err := source.StreamFile(context.TODO(), logger, filePath)
 	if err != nil {
 		if err == baggageclaim.ErrFileNotFound {
--- a/atc/exec/task_config_source.go
+++ b/atc/exec/task_config_source.go
@@ -23,7 +23,7 @@ import (
 type TaskConfigSource interface {
 	// FetchConfig returns the TaskConfig, and may have to a task config file out
 	// of the artifact.Repository.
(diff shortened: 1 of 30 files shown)


kcmannem:
not a biggie change. In baggageclaim 1.6.3, the client is now required to pass a context for StreamIn/Out calls. This pr passes along context to all of these calls in the ATC.

atc/exec/task_config_source.go
```diff
@@ -79,7 +80,10 @@ func (configSource FileConfigSource) FetchConfig(logger lager.Logger, repo *arti
 		return atc.TaskConfig{}, UnknownArtifactSourceError{sourceName, configSource.ConfigPath}
 	}
 
-	stream, err := source.StreamFile(logger, filePath)
+	// TODO: idk if we really want to pass a context in here. This is a very
+	// 		 shortlived stream. It would pollute the TaskConfigSource Interface
+	//		 as all the FetchConfigs will be need to have this passed in.
+	stream, err := source.StreamFile(context.TODO(), logger, filePath)
```
kcmannem: look at this note and tell me if you agree

ddadlani: I'm cool with leaving it out for now, as long as we have a comment explaining why

vito: I'm leaning towards just implementing it. TaskConfigSource is only used by Tasks, right? You could also use `lagerctx` to pass the `lager.Logger` argument through the `Context` argument if you don't want to add another argument.

It's a short-lived stream but if we're intending to guard against an unreachable worker, it won't really matter how long the transfer takes. This is also one of the first things that happens in a task step, so it could be the first thing that hangs.

kcmannem:
@vito It shouldn't really change the responsiveness of aborting builds. It just makes sure an aborted build cleans up something it wasn't before.

vito:
@kcmannem I was imagining a scenario where there's a build running and transferring a large file from one step to the next - wouldn't this allow the transfer to be interrupted?

atc/exec/task_config_source.go
kcmannem: done

kcmannem:
This needs to be merged in before the overlay pr can go in since its in a higher version of bc https://github.com/concourse/baggageclaim/pull/30

pivotal-jamie-klassen:
this is more than I can scan at a glance... obviously it made it into v5.5.0 but does it need a release note? @kcmannem @ddadlani @clarafu @vito

kcmannem:
@pivotal-jamie-klassen  I don't think there was a known issue this addressed. We're just being more correct.

pivotal-jamie-klassen:
@kcmannem `release/undocumented` it is then.

Issue #2896, opened 2018-11-30
Consolidate multiple metrics emitters


jama22:
This is a spike to investigate the possibility of consolidating multiple metrics emitters...for easier maintainability

cc @cirocosta for input!

marco-m:
I suggest to have a look at https://www.robustperception.io/its-easy-to-convert-pull-to-push for an approach that supports in a natural way pull systems like Prometheus and push systems like InfluxDB.

cirocosta:
Hey, 

I wanted to highlight that I'd **really** like to get to the point where we can have all the metrics consolidated as pull-first with emissions (push) as a second thought like described in the article.

I was leaning towards waiting for something like [openmetrics](https://openmetrics.io/), which from [the announcement when they joined CNCF](https://www.linuxfoundation.org/blog/2018/08/cncf-to-host-openmetrics-in-the-sandbox/), seems to indicate that most of the vendors we support are interested in somehow contributing to it (prometheus, influx, and datadog).

Another idea was to just drop everything and go straight Prometheus, leaving to something like Telegraf the task of outputting those metrics to other systems.

From a k8s perspective, that'd be great: if you use datadog, leave `telegraf` as a sidecar container doing the job of scraping the Prometheus endpoint.

From BOSH, it seems like an addon could do that same job (not sure if this would be the right abstraction though).

Any thoughts on those?

thx!

vito:
@cirocosta I'd be fine with slimming down to Prometheus support if that seems OK with everyone. Prometheus seems to have the most traction/attention among our community so far.

We use InfluxDB ourselves but I'd be fine with switching over and seeing how it goes.

marco-m:
@cirocosta Here is my understanding and proposal.

In my understanding, OpenMetrics is about standardizing on the Prometheus exposition format. so (I think) it works only for a pull model. On the other hand, InfluxData is participating to OpenMetrics. Reading https://community.influxdata.com/t/monitoring-with-push-vs-pull-influxdb-adds-pull-support-with-kapacitor/795, it is already possible today to use Kapacitor to scrape a Prometheus endpoint and ingest data into InfluxDB.
If you are looking for ready-to-use metrics, it looks like that https://opencensus.io/ might be closer (and by the way OpenCensus also supports distributed tracing, which is another thing that would be wonderful to add to Concourse :-).

So my proposal:
- As Vito mention, focus on one metric approach, the pull model. Remove logic for push.
[...]

Issue #4038, opened 2019-06-20
datadog metric concourse.ci.build_started reported as build-id gauge instead of single count


eedwards-sk:
## Bug Report

`concourse.ci.build_started` is reported with a value of the build-id

it should instead be a `count` metric, with a value of `1`

## Steps to Reproduce

enable metrics emission to datadog

run a build

## Expected Results

value of metric is 1

## Actual Results

value of metric is the build id

## Additional Context

cannot graph it in a meaningful way

value should be count type (not gauge, and not deprecated counter), since then it can be graphed "builds over time", with tags to further refine the view

## Version Info

* Concourse version: 5.2.0
* Deployment type (BOSH/Docker/binary): binary
* Infrastructure/IaaS: AWS ASG EC2
* Browser (if applicable): N\A
* Did this used to work? Unknown

eedwards-sk:
This is still a needed tweak for metrics. It also might be straightforward to implement.

eedwards-sk:
Poke.

eedwards-sk:
Poke. Should be easy.

Issue #3196, opened 2019-02-01
multiple metrics violate datadog best practices and impact billing


eedwards-sk:
## What challenge are you facing?

Currently, metric tags emitted to datadog for the `http_response_time` metric include tags for the `path` which when correlating to build numbers or other dynamic paths, may become an ever-increasing list of tags, impacting billing and cardinality limits

e.g.:

`path:/api/v1/teams/main/pipelines/long/jobs/long-running-process/builds/29`

in addition the `build_finished` and `build_started` metrics include the `build_id`

because each tag on a metric becomes a new dimension in datadog, this means that:

- each `path` or `build_id`, which at the very least will be unique for each build, means that each build results in a new custom metric
- datadog billing charges based on custom metric use, meaning every single build can potentially increase billing cost
- because a build number is a counter, this will also start to push datadog cardinality limits as build counts increase, e.g. to 1000+
- any other highly dynamic paths would also push this limit

https://docs.datadoghq.com/tagging/#defining-tags
https://docs.datadoghq.com/developers/metrics/custom_metrics/#how-is-a-custom-metric-defined

## What would make this better?

- Stop tagging `http_response_time` metric with the `path`

- Stop tagging `build_started` and `build_finished` metrics with the `build_id`, and any other metrics adding tags with 'counters'  or dynamic / ever increasing values.

- Alternatively, identify the most dynamic sections of the metrics and consolidate them, e.g.:

`path:/api/v1/teams/main/pipelines/long/jobs/long-running-process/builds/29`

could be reduced to

`path:/api/v1/teams/main/pipelines/long/jobs/long-running-process/builds`

but if a user is generating a lot of dynamic pipelines, jobs, or tasks, it would still possibly be an excessive usage of custom metrics

- Alternatively, allow a metric blacklist in the configuration so I can stop reporting these metrics.

## Are you interested in implementing this yourself?

Theoretically Yes, I will help however I can, but I doubt I can do the implementation myself with my current knowledge of Go and the codebase.

eedwards-sk:
Based on my datadog billing page for usage, the top three custom metrics (by tag usage) from concourse are:

```
build_finished
build_started
http_response_time
```

eedwards-sk:
suggesting we remove or make configurable the submission of...

`path`:

https://github.com/concourse/concourse/blob/33d51223f4baf3f571d5e536b8992504fe9ee364/metric/metrics.go#L206-L241

`build_id`:

[...]

PR #4247, opened 2019-08-09
experiment: standardize on prometheus for metrics

cirocosta:
Hey,

I wanted to try out the idea of getting rid of all of our emitters and standardizing on
Concourse being a system that can be observed without having to worry about emitting its
measurements to others (e.g., adopting a pull-based approach and nothing more).

```
	
	something that cares 
	about the state of a		(e.g., prometheus, telegraf ...)
	concourse installation
	         |
	         |
	    "how are you doing?"
	         |
		 |
		 *---> concourse
		 	  http://concourse:$prom_port/metrics


```

At the moment, I didn't yet convert all the metrics that we have, and IMO, that should not
be done in a 1:1 mapping anyway - there might be a certain set of metrics that we just
don't care about, and others that we do, but we just never implemented.

Also, I tried to be super conservative when it comes to labeling in order to try to avoid
future problems with cardinality explosion, so all measurements (at the moment) have
bounded cardinality, which might mean that they're not super useful for very
context-specific metrics like "is this particular job being slower over time".

ps.: this is just an experiment to think about the area of [consolidating the emitters](https://github.com/concourse/concourse/issues/2896).

I didn't play with the Telegraf Prometheus input plugin (see
https://github.com/influxdata/telegraf/tree/master/plugins/inputs/prometheus) which should
"unlock" Prometheus-based metrics to those who can't just use Prometheus and need
something to emit metrics to their systems), but I'd be curious to know if other people
did!

Also, do you see this being problematic in any sense in the future? Is super high-context
relevant for you? (if so, should we start thinking about other approaches that are not
metrics that can drive this "high-context" observability - like distr tracing - into the
mix?)

I'd appreciate a lot your feedback!

Thanks!


Makefile
```diff
@@ -0,0 +1,54 @@
+all: run
+
```
cirocosta: (this is just for me to be able to more easily run this without having `web` running in a container 😅 )

```
make database
make worker
make run
```

(I should've named `run` `web`).

cirocosta:
In terms of "developer experience", it ends up looking like this for us:


1. add the definition of your measurement to `metrics.go`

```go
var (
	// other metrics ...

	ResourceChecksDuration = prometheus.NewHistogramVec(
		prometheus.HistogramOpts{
			Name: "concourse_resource_checks_duration_seconds",
			Help: "How long resource checks take",
		},
		[]string{LabelStatus},
	)
)

```

2. update the in-memory repr of your measurement

```go
metric.
	ResourceChecksDuration.
	WithLabelValues("failed").
	Observe(checkDuration.Seconds())
```

cirocosta:
Something I like about this is that instrumenting individual components of Concourse would
be simpler, as one would just go through that simple model of "I define a variable that
represents the measurement", and then "in my code, I increment it".

That'd mean having an easy time adding metrics to components like `tsa`, `beacon`,
`baggageclaim` ... etc

cirocosta:
```

	(same cluster, diff networks and diff proms)

	.--------network 1-----
	|
	|    .---.--prom--.-----.
	|    |   |        |     |
	|  atc1 tsa1     atc2 tsa2
	|
	|


	.--------network 2------------
	|
	|    .------.---prom2-.------.
	|    |      |         |      |
	|  beacon1 bc1      beacon2 bc2 ...



```


or


```

	(all managed in the same cluster & network)

	.--------network 1-----
	|
	|    .---.--prom--.-----.---------.------.---------.------.
	|    |   |        |     |         |      |         |      |
	|  atc1 tsa1     atc2 tsa2        |      |         |      |
	|  				beacon1 bc1      beacon2 bc2 ...


```


or


```

	(external workers connected to a given cluster)

	.--------network N------------
	|
	|    .---------------------------> datadog
	|    |        .------------------> datadog ...
	|    |        |
	|  telegraf telegraf
	|    |        |     
	|  beacon1   bc1    


```

atc/metric/metrics.go
```diff
@@ -1,581 +1,161 @@
...
-		Event{
-			Name:  "resource checked",
-			Value: 1,
-			State: state,
-			Attributes: map[string]string{
-				"pipeline":  event.PipelineName,
-				"resource":  event.ResourceName,
-				"team_name": event.TeamName,
-			},
+	ResourceChecksDuration = prometheus.NewHistogramVec(
```
cirocosta: here's an example of how you'd be defining the measurement

atc/radar/resource_scanner.go
```diff
@@ -373,30 +373,44 @@ func (scanner *resourceScanner) check(
...
-		TeamName:     scanner.dbPipeline.TeamName(),
-		Success:      err == nil,
-	}.Emit(logger)
 
 	if err != nil {
 		if rErr, ok := err.(resource.ErrResourceScriptFailed); ok {
 			logger.Info("check-failed", lager.Data{"exit-status": rErr.ExitStatus})
+			metric.
+				ResourceChecksDuration.
+				WithLabelValues("failed").
```
cirocosta: the corresponding example of how you leverage that measurement you defined in `metrics.go`

atc/atccmd/command.go
```diff
@@ -427,10 +417,10 @@ func (cmd *RunCommand) Runner(positionalArguments []string) (ifrit.Runner, error
 	}
 
 	members = append(members, grouper.Member{
-		Name: "periodic-metrics",
-		Runner: metric.PeriodicallyEmit(
-			logger.Session("periodic-metrics"),
-			10*time.Second,
+		Name: "metrics-handler",
```
cirocosta: all that you need to have an endpoint exporting prometheus metrics

kcmannem:
I like this.

"That'd mean having an easy time adding metrics to components like tsa, beacon,
baggageclaim ... etc"

^ thats gunna be super useful to us on runtime.

cirocosta:
@kcmannem  @ddadlani  I started collecting metrics from the workers:

Screen Shot 2019-08-26 at 8 56 11 AM

what other measurements would be nice to capture that are currently not possible to do today? 🤔

cirocosta:
Moving on, some things that could make this better:

1. use `promtheus.Timer`s when interacting with observers (histograms & summaries)
  - https://godoc.org/github.com/prometheus/client_golang/prometheus#Timer

2. potentially (?) move to summaries rather than using histograms
  - there's CPU and memory cost associated with performing the quantile
    computations in the client
    - how costly? I'm not very sure.
	- keeping a small Objectives map (quantiles to compute) would probably
	  make this not very CPU intensive, but, still, a stream needs to be
	  kept in memory

3. investigate how influx' telegraf prometheus input plays with these metrics
  - I'm really not sure how these metrics would look like when exported.
    - my guess is that counters & gauges would be "as expected", but histograms
      would look weird. moving to summaries could make it better though

cirocosta:
There are certain metrics here that are clearly lacking:

- number of teams
- number of pipelines
- ...

These metrics that are predominantelly statistics from the database, are missing, as there's not a very good way of having them exposed by a single component (e.g., it'd be weird to have two ATCs reporting the number of teams that they just got from the database).

For that case, I'm leaning towards having a second component (https://github.com/cirocosta/flight_recorder/ :sweat_smile:), rather than having a "periodic emitter" like we did before :thinking:

Wdyt?

thanks!

atc/gc/worker_collector.go
```diff
@@ -43,6 +46,9 @@ func (wc *workerCollector) Run(ctx context.Context) error {
 
 	if len(affected) > 0 {
 		logger.Info("marked-workers-as-stalled", lager.Data{"count": len(affected), "workers": affected})
+
+		// workers_stalled
```
cirocosta: Would it make more sense to have it as `workers_marked_stalled` as that's the actual action taking place? 🤔  

That way it'd make sense to be per-ATC (still being possible to be `sum`med up).

atc/worker/worker_helper.go
```diff
@@ -57,6 +59,13 @@ func (w workerHelper) createGardenContainer(
 			Env:        env,
 			Properties: gardenProperties,
 		})
+
+	metrics.
```
cirocosta: This is web's perspective of how long a container creation is taking.

I wonder if this would be more valuable if we included the worker name in the label set, so that we could tell what's the particular worker that is facing some long time to have the container created 🤔  

Naturally, in the best scenario we'd have distributed tracing in a way that would allow us to have all the context and separate between time taken as seen from the `web` and time taken by just `garden`, but I think what we have for now (+ having worker name in the label set) would be good enough

metrics/handler_wrapper.go
```diff
@@ -0,0 +1,31 @@
...
+		Handler: handler,
+	}
+}
+
+func (handler MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	metrics := httpsnoop.CaptureMetrics(handler.Handler, w, r)
+
+	HttpResponseDuration.
+		With(prometheus.Labels{
+			"code":  strconv.Itoa(metrics.Code),
```
cirocosta: I don't think there's a lot of value of having the plain `code` here - we could, instead, move towards `2xx`, reducing the amount of possible values that this can take down to 4 (`2xx`, `3xx`, `4xx`, `5xx`) - just what we care about

metrics/web.go
```diff
@@ -0,0 +1,141 @@
...
+import (
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+var (
+	slowBuckets = []float64{
+		1, 30, 60, 120, 180, 300, 600, 900, 1200, 1800, 2700, 3600, 7200, 18000, 36000,
```
cirocosta: ooh yeah, this bad boy here ends up generating 16 labels for each measurement that we end up adding 😅 moving to summaries, that would reduce quite a bit the load we put on the server (specially given that `slowBuckets` is used exclusively for `build` time - the metric with the highest cardinality).

metrics/web.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+func NewWebMetricsHandler() http.Handler {
+	registry := prometheus.NewRegistry()
+
+	for _, collector := range defaultCollectors {
+		registry.MustRegister(collector)
+	}
+
+	registry.MustRegister(
```
cirocosta: I'm still not very happy with having to not forget to add the variable here, but I couldn't think of a good way of automatically adding (without having to do something like what [`promauto`](https://github.com/prometheus/client_golang/blob/master/prometheus/promauto/auto.go) does - reimplementing the constructors).

gerhard:
This cannot come soon enough! The current metrics setup based on Influxdb is bursting - again. Throwing more CPUs & memory at it, until we have this new solution.

I discovered this while getting frustrated with the current metrics setup, used to create #4340

I highly recommend reaching out to @brian-brazil when you are ready with this, he has valuable insights into best practices: prometheus/docs#1414

Thanks for working on this @cirocosta !

brian-brazil:
A quick high level comment:

> These metrics that are predominantelly statistics from the database, are missing, as there's not a very good way of having them exposed by a single component

The /metrics of a process should expose information about the performance of the process itself, not be making RPCs to other systems to gather data or expose the data it's processing. So for example the MySQLd exporter exposes metrics about MySQLd performance - not raw data from tables. You'd have a different exporter for this. Kubelet vs kube-state-metrics is a good example of this.

That's not to say you can't expose such metrics on another endpoint (and sometimes you can cheat a little for some cheap metrics a given process knows about anyway) from the same process though.

gerhard:
I would enjoy starting to dog-food this in 2 weeks from now @cirocosta . Happy to delay if you need more time, but I am seriously intrigued about what you have going here 😉

cessien:
This is awesome! I was pretty unhappy with the cardinality for build duration that prometheus metrics offered so far. We're going to build this PR and try this out internally (CarGurus). I'll let you know what we like/ don't like about it

eedwards-sk:
I have concerns over this idea. I do not use prometheus, I use Datadog. I expect concourse to continue to support emitting statsd metrics that I can consume from my Datadog Agent.

The current datadog support is officially recognized by Datadog: https://docs.datadoghq.com/integrations/concourse_ci/

This means using the current statsd configuration allows me to consume concourse metrics into my Datadog account without any additional charges (even in spite of some cardinality issues that would otherwise be very expensive #3196 #4038)

I think losing compatibility or breaking the current contract that is documented by Datadog would be a painful regression.

In addition to not wanting to deal with prometheus, I also don't like the idea of having to manage or run a telegraf agent.

That's one more thing I'd now be responsible for configuring. Their current output plugin, assuming this is what would be used, could also possibly result in a break of the "free metrics" benefit of the current metrics namespace that Datadog represents. As per https://github.com/influxdata/telegraf/tree/master/plugins/outputs/datadog "Metrics are grouped by converting any _ characters to . in the Point Name.", which would mean needing to ensure that the namespace ended up identical to the default as it was before (`concourse.ci`). 

I'm also concerned if this means going to a "pull based" model rather than "push based", and how that affects system load and the frequency of reported metrics. The Datadog agent allows for high-frequency metric submission (udp statsd can receive lots of metrics per second) and automatically handles aggregation, sampling, batching and a bunch of other useful stuff to ensure metrics are captured and shipped.

If pull based, how does this affect the frequency of metric emissions? Does an agent need to poll aggressively to get more data points? Is the endpoint going to return a series of data points over a timespan and the agent is responsible for knowing what data points it already has?

cirocosta:
Hi @eedwards-sk, **thank you** a lot for putting time on sharing this feedback!

<br />

The current datadog support is officially recognized by Datadog: https://docs.datadoghq.com/integrations/concourse_ci/

That's a very valid point! It can definitely be a deal breaker as I can imagine
that this significantly reduces the expenses at the end of the day (which
matters a lot).

I wonder if in this case it's a matter of engaging with Datadog :thinking:

We expect to have newer metrics as the system changes, and improve them when we
see fit, so I can imagine that we'd like not to be too dependant on a vendor
expecting a restricted set of metrics. I see this is a win-win as being able to
better observe a system in DataDog would be something great for them too (as a
selling point).

> In addition to not wanting to deal with prometheus, I also don't like the idea
> of having to manage or run a telegraf agent.

Yeah .. I totally agree! I'd not like to do that either -  now there's one more
thing to break, to upgrade, to have security vulns ...

My view on this though is that despite that seeing like a problem, as the
industry moves forward towards standard formats (like OpenMetrics), we (as the
ones developing a CI system) are able to focus less and less on how to deal with
all these metric systems, and just implement the standard that everyone agrees
on, and let those that spend their entire they thinking about metrics, to
implement the agents that collect from those exporting in that standard format.

We can see that with Datadog itself, announcing that the Datadog agent is now
capable of ingesting metrics from those exposing them using the prometheus
exposition format (which OpenMetrics is pretty much entirely based on):

- https://www.datadoghq.com/blog/monitor-prometheus-metrics/

This is also true for NewRelic, another emitter that we currently have to
maintain:

- https://docs.newrelic.com/docs/integrations/prometheus-integrations

Although it might be a steep transiton, my view is that it's a good thing as
having less of these things that are not the core of what Concourse cares about,
allows us to have less of these edges full of dust that no one really owns.


> I'm also concerned if this means going to a "pull based" model rather than
> "push based", and how that affects system load and the frequency of reported
> metrics. The Datadog agent allows for high-frequency metric submission (udp
> statsd can receive lots of metrics per second) and automatically handles
> aggregation, sampling, batching and a bunch of other useful stuff to ensure
> metrics are captured and shipped.

> **If pull based, how does this affect the frequency of metric emissions? Does an
> agent need to poll aggressively to get more data points? Is the endpoint going
> to return a series of data points over a timespan and the agent is responsible
> for knowing what data points it already has?**

That's indeed a good point.

It might be that people could be using our metrics not as "just monitoring", but
as events that they rely on the exact timestamp of the submission for some
reason (e.g., we emit `build finished` metrics on a very granular basis, thus,
one could perhaps leverage that as something that tells exactly when a given job
failed), and indeed, with a pull-based model as I proposed here, that'd not be
possible as Prometheus (or any system that scrapes the endpoint) would be
essentially looking at snapshots of how the counters/gauges/etc look like at the
moment of scraping.

In this case, we'd indeed be breaking that functionality.

---

please let me know what you think - happy to discuss!

Thanks!

eedwards-sk:
@cirocosta Thanks for linking the prometheus article. Datadog is so big now a days I miss half the stuff they're doing.

Regarding the costs issue, as long as metrics continue to be prefixed with `concourse.ci.` then I believe Datadog would not charge for them. Although I'm not sure how metric namespaces work when using their prometheus (OpenMetrics) support, or how they get mapped as they're converted or processed.

I agree that concourse shouldn't have to carry the burden of knowing how to deal with a myriad of providers. I was under the impression that statsd is already a fairly generic protocol, which datadog supports (with some straightforward enhancement for dogstatsd / tags).

So I wonder if this is really just moving from one standard to another? If this new OpenMetrics approach works as well / better than statsd, I'm not opposed. Although it feels possibly like a k8s-driven change, and as much as the industry makes it seem, not everyone is using k8s, myself included. So perhaps I'm just not seeing this immediate need as clearly.

>It might be that people could be using our metrics not as "just monitoring", but
as events that they rely on the exact timestamp of the submission for some
reason (e.g., we emit build finished metrics on a very granular basis, thus,
one could perhaps leverage that as something that tells exactly when a given job
failed), and indeed, with a pull-based model as I proposed here, that'd not be
possible as Prometheus (or any system that scrapes the endpoint) would be
essentially looking at snapshots of how the counters/gauges/etc look like at the
moment of scraping.

Those are definitely things I look for out of metrics -- correlation is a very strong tool. Datadog reinforces that by letting me aggregate and correlate multiple metrics from e.g. the same host or hosts on the same timescale.

concourse-datadog-1

So it can be useful to see that a spike in CPU load was correlated to a build starting, or IOWait spiking when the container count increased.

If a polling agent operates regularly enough, then I assume it can still approximate the relative timeframe of when the events are occurring. However this still raises the issue of balancing poll frequency versus load due to polling (and serving a stats request).

brian-brazil:
> Although I'm not sure how metric namespaces work when using their prometheus (OpenMetrics) support, or how they get mapped as they're converted or processed.

They'll be charged as custom metrics. I don't think they've any integrations yet that use Prometheus under the covers, but I guess it's only a matter of time. Asking them is best.

> Although it feels possibly like a k8s-driven change

While Prometheus is commonly used on k8, it is not k8 specific in any way. Plus the major vendors all support the format, and it'll be more efficient than statsd.

> It might be that people could be using our metrics not as "just monitoring", but
as events that they rely on the exact timestamp of the submission for some
reason 

If you think this is likely, I'd suggest exporting gauge metrisc with the relevant timestamp(s).

eedwards-sk:
>They'll be charged as custom metrics. I don't think they've any integrations yet that use Prometheus under the covers, but I guess it's only a matter of time. Asking them is best.

Not necessarily. It depends on the namespace.

eedwards-sk:
@brian-brazil 

>Although it feels possibly like a k8s-driven change

>While Prometheus is commonly used on k8, it is not k8 specific in any way. Plus the major vendors all support the format, and it'll be more efficient than statsd.

That's what I mean by "k8s driven". I didn't say k8s specific :)

Can you clarify why it's more efficient than statsd? That sounds highly relevant to this discussion.

brian-brazil:
> Can you clarify why it's more efficient than statsd? That sounds highly relevant to this discussion.

https://www.robustperception.io/which-kind-of-push-events-or-metrics Statsd is (usually) events, Prometheus is metrics.

eedwards-sk:
@brian-brazil 

Thanks! That's a well summarized overview. I would still present the (anecdotal) perspective that, in practice, at scale, statsd has worked effectively in production for the past few years I've used it alongside Datadog, under a myriad of configurations and load.

I personally don't have any drivers to migrate away from it, and I'll continue to have numerous stacks and services using it for some time.

Thus I'll be interested to see what this looks like when going from OpenMetrics -> Datadog. If there is relative parity in terms of configuration complexity / management liability, I don't really care what is "inside the box", as long as it surfaces the info we want :)

brian-brazil:
> statsd has worked effectively in production for the past few years I've used it alongside Datadog, under a myriad of configurations and load.

I believe DataDog has a per-machine statsd, which avoids the usual issues from attempting to run a single statsd for many many machines.

> Thus I'll be interested to see what this looks like when going from OpenMetrics -> Datadog.

It'll be the exact same, it's already all supported AIUI. Datadog have contributed a number of performance improvements to the Prometheus Python client to facilitate their Prometheus/OpenMetrics integration.

pivotal-jwinters:
Admittedly I have not read through this entire thread so I might be missing some valuable context here, but after briefly chatting with @cirocosta it seems like we would be able to reduce the burden of adding new metrics if we added a `type` to our top level metric interface. This `type` would represent things like `counter`, or `histogram`.

That way each emitter can move away from mapping the event.name in favour of event.type. This means that the individual emitter would only ever change if a new type was added, which should be rare. Emitters would also be able to ignore types that don't translate between push or pull collection strategies.

cirocosta:
thanks for everyone who helped out expressing your thoughts & concerns! I'll close this one for now in favor of https://github.com/concourse/concourse/pull/5104

Issue #3989, opened 2019-06-07
web crash due to aborting during 'try' step followed by 'retry' step


hprotzek:
## Bug Report

We are running 3 web nodes on k8s (gke), after a rolling restart under high load, lots of running pipelines from 50 workers, the web nodes crashed multiple times with:

```
runtime error: invalid memory address or nil pointer dereference [signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0xaa585a]
at github.com/concourse/concourse/atc/exec.(*RetryStep).Succeeded (retry_step.go:47)
at github.com/concourse/concourse/atc/exec.OnSuccessStep.Succeeded (on_success.go:45)
at github.com/concourse/concourse/atc/exec.OnSuccessStep.Succeeded (on_success.go:45)
at github.com/concourse/concourse/atc/exec.OnFailureStep.Succeeded (on_failure.go:44)
at github.com/concourse/concourse/atc/engine.(*execBuild).Resume (engine.go:198)
```

## Steps to Reproduce

We couldn't reproduced it, but this was the second time we encountered this problem. 3 web nodes, 50 workers and lots of running pipelines.

## Version Info

* Concourse version: 5.2.0, 5.1.0
* Deployment type (BOSH/Docker/binary):  k8s
* Infrastructure/IaaS: gcp

cirocosta:
Hi @hprotzek ,

Thanks for filling the issue!

What do you mean by `5.2.0,5.1.0`? Did you see the very same error happening on both versions?

@ddadlani given that the problem is at `exec`, is this something that the runtime team could look at?

Thanks!

hprotzek:
@cirocosta yes we also saw it with version 5.1.0

vito:
Might need this tracked down to a pipeline config that caused it in order to reproduce it. `Succeeded` should only be called after `Resume` was called so that panic is interesting. 🤔

hprotzek:
We are using https://github.com/springernature/halfpipe to generate our pipelines. You can have a look at the examples: https://github.com/springernature/halfpipe/blob/master/e2e/artifacts/expected-pipeline.yml to see how most of our pipelines are structured.

josecv:
We also encountered this in our concourse cluster, which is also deployed on kubernetes, and also has 3 web pods.

The sequence of events seems to have been as follows:

* Introduced a change to add `attempts: 5` to a number of resource `get`s and `put`s in a number of pipelines.
[...]

PR #4252, opened 2019-08-09
Fail the try_step if the inner step was aborted [#3989].

josecv:
Signed-off-by: jose <jose@autonomic.ai>

### DIFF
--- a/atc/exec/try_step.go
+++ b/atc/exec/try_step.go
@@ -7,20 +7,23 @@ import (
 // TryStep wraps another step, ignores its errors, and always succeeds.
 type TryStep struct {
 	step Step
+	aborted bool
 }
 
 // Try constructs a TryStep.
 func Try(step Step) Step {
-	return TryStep{
+	return &TryStep{
 		step: step,
+		aborted: false,
 	}
 }
 
 // Run runs the nested step, and always returns nil, ignoring the nested step's
 // error.
-func (ts TryStep) Run(ctx context.Context, state RunState) error {
+func (ts *TryStep) Run(ctx context.Context, state RunState) error {
 	err := ts.step.Run(ctx, state)
 	if err == context.Canceled {
+		ts.aborted = true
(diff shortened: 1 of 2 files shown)


josecv:
I'm not a golang understander so please let me know if I've made any mistakes, I'll be glad to fix em up

xtremerui:
@vito the fix is good to me. However since there is a nice example in the issue that could reproduces the problem, is it worth to add a topgun test based on that? Could be a separate PR.

atc/exec/try_step.go
```diff
@@ -7,20 +7,23 @@ import (
 // TryStep wraps another step, ignores its errors, and always succeeds.
 type TryStep struct {
 	step Step
+	aborted bool
 }
 
 // Try constructs a TryStep.
 func Try(step Step) Step {
-	return TryStep{
+	return &TryStep{
```
pivotal-jwinters: How come we're changing this from a pointer?

vito: it's so the `ts.aborted = true` state persists through to the `Succeeded()` call

pivotal-jwinters: Why does wouldn't it get persisted in the pointer case?

vito: assuming you mean non-pointer case, `Run` is called with a copy of the struct as its receiver, so changes are only local to that function's copy

pivotal-jwinters: oh wow, I totally read this diff as going form ptr to value. haha don't mind me

josecv: I was under the impression this was necessary to mutate it later on, but if I'm wrong i can definitely fix it

pivotal-jwinters: no no its fine. I misread the diff and got confused for a second.

cirocosta:
Thanks a lot, @josecv !

PR #4267, opened 2019-08-12
atc: update contaner missing_since and worker start_time to timestamp…

xtremerui:
… with time zone


xtremerui:
> It makes sense to switch to timestamp with time zone, but is there a particular issue this is trying to fix? Just trying to get context :)

No an issue just to make consistent since I added a wiki [here](https://github.com/concourse/concourse/wiki/Concourse-Go-Style-Guide#timestamps)

> Also if we're changing the `containers` table to use `timestamp with time zone`, can we also change `volumes` for uniformity?

Yes volumes should be the same. Thx!

PR #4272, opened 2019-08-12
Fly url rebased

xtreme-sameer-vohra:
This PR is meant to complete PR #3683 
In addition to changes proposed in the original PR
- this rebases of latest master
- add additional error case for clarity

Thank you @mockersf for the PR

### DIFF
--- a/fly/commands/abort_build.go
+++ b/fly/commands/abort_build.go
@@ -9,8 +9,8 @@ import (
 )
 
 type AbortBuildCommand struct {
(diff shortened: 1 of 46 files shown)


deniseyu:
thanks @xtremerui for catching that! we thought `env` had previously been used, but it looks like it was actually added by #3683.
 
This actually ends up being a bigger UX change that should warrant some discussion. I think that adding environment variable setting for things like trigger-job could end up being surprising to a `fly` user (who hasn't been tripped up by an env var living in their shell session before?) 

I'm going to revert the `env` flags but leave in the URL-based targeting functionality for these commands.

cc @mockersf @vito @xtreme-sameer-vohra

Issue #4240, opened 2019-08-08
[fly] Super admin can log into any team


taylorsilva:
### Description
See title

### Acceptance Criteria
- [ ] a user that is a owner of the `main` team can log into any other team even when they're not explicitly part of that team
- [ ] User can perform any action after logging into another team they're not part of

zoetian:
**Dev Note**:

After we made the changes to `fly`, we found 2 issues:

1) admin would login to non-existed team w/o checking
```
fly -t local login -u test -p test -n non-exist-team
```

it would return 

```
logging in to team 'non-exist-team'

target saved
```

2) fly admin can see all builds without switching (i.e. login to) teams

```
> fly -t local login -u test -p test -n main
> fly -t local builds
```

for build command, admin users can now see all builds without switching to specific teams to see theirs builds. By default, it returns all the teams' builds for this admin user. We need to confirm if this is the expected feature.

pivotal-bin-ju:
**DEV NOTE**
> 1. admin would login to non-existed team w/o checking
got fixed. test cases also were populated as well.
> 2. fly admin can see all builds without switching (i.e. login to) teams
we've categorized all available `fly` commands on levels. such as `pipeline`, `team`, etc.
For the `super admin` user, the general idea is: 
- If a `fly` command act like team scoped with the api call like `/api/v1/teams/TEAMNAME/API-COMMAND-ENTRY`, `fly` cli should operate as the normal user  that owns that team. 
- If a `fly` command act like cluster scoped with the api call non-team-scoped, like `api/v1/builds`, `fly` cli should behave as `super admin` that is able to operate the cluster, return the records across team boundaries.

zoetian:
**Dev Note**:

 @YoussB and I ran all the concourse-wise commands for fly and documented the request type and the detailed api endpoint. 

other than that, we found a bug in the previous development build. when running: 

```bash
$ fly -t local login -n main
logging in to team 'main'

navigate to the following URL in your browser:

  http://localhost:8080/login?fly_port=51477

or enter token manually:
target saved
```
it can successfully login as the super admin user. However, when switching and login to the team it wasn't own:

```bash
$ fly -t local login -n team1
logging in to team 'team1'

navigate to the following URL in your browser:

  http://localhost:8080/login?fly_port=51482

or enter token manually:

could not reach the Concourse server called local:

    Get /api/v1/teams: unsupported protocol scheme ""

is the targeted Concourse running? better go catch it lol
```
[...]

PR #4273, opened 2019-08-12
Super admin login any team

zoetian:
Address: https://github.com/concourse/concourse/issues/4240


pivotal-jamie-klassen:
If this doesn't depend on #4238, I think it should be rebased without that
history -- I don't feel totally comfortable executing the cherry-pick
operation described here. However, it looks like the dependent PR will be merged
shortly so I'll just wait until that one is merged to review this one.

zoetian:
Hello, this PR is ready for review

pivotal-jamie-klassen:
while reviewing this PR, I haven been experimenting with accessing a
v3.14.1 cluster with this version of fly, to try to get more insight into this
snippet:

```go
func unmarshalToken(tokenValue string) (map[string]interface{}, error) {
	tokenContents := strings.Split(tokenValue, ".")
	if len(tokenContents) < 2 {
		// this is really bad and makes it hard to write proper integration tests
		return nil, nil
	}
```

I hope to finish this experiment tomorrow.

testflight/admin_priviliges_test.go
```diff
@@ -0,0 +1,173 @@
...
+                echo 'waiting for /tmp/stop-waiting to exist'
+                sleep 1
+              done
+              sleep 100
+              echo done
+              `), 0644)
+		Expect(err).NotTo(HaveOccurred())
+
+		oldTarget = flyTarget
+		flyTarget = adminFlyTarget
```
pivotal-jamie-klassen: mutating this unprotected global variable in a multithreaded context is very dangerous, and I think it is causing the failures we see on CI, where this test sets the variable in one thread, interfering with the other threads' ability to log in. I see this as a bit of a testing disaster that is nobody's fault 😢 ... I suspect it is safer to put something like this in the `SynchronizedBeforeSuite` and have this target just stick around for the duration of the whole testflight suite? there is also a `withFlyTarget` helper, which is also not thread-safe...

PR #4277, opened 2019-08-14
Do not acquire lock when decreasing task count.

aledeganopix4d:
The tasks count decreasing (used in the limit-active-tasks strategy) can be
executed concourrently as the DB transaction will make sure that the end
value is consistent.

Trying to acquire a lock when decreasing the tasks count in a system with
several flying task can easily result in a race to acquire lock and eventually
even a "de facto" dead lock.

We observed this issue on our Production CI where it eventually all just hanged, what happens is the following:

- New tasks acquire the lock to check if a worker is free and all are busy
- Completed tasks try to acquire a lock to decrease the task and free the worker
- More tasks enter the queue and try to acquire a lock
- The decrease-active-tasks is can perform its operation once in a blue moon making the situation worse and worse

@deniseyu 
@ddadlani

### DIFF
--- a/atc/worker/client.go
+++ b/atc/worker/client.go
@@ -143,7 +143,7 @@ func (client *client) RunTaskStep(
 	}
 
 	if strategy.ModifiesActiveTasks() {
-		defer decreaseActiveTasks(logger.Session("decrease-active-tasks"), lockFactory, chosenWorker)
+		defer decreaseActiveTasks(logger.Session("decrease-active-tasks"), chosenWorker)
 	}
 
 	container, err := chosenWorker.FindOrCreateContainer(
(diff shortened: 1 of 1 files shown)


ddadlani:
@aledeganopix4d interesting! I apologize for not answering your question about this in another thread earlier - I filed it away to respond to later and it completely slipped my mind.

We did not consider potential lock contention. The reason we added the lock was that in the case that between reading the number of tasks and actually selecting a worker, the number of tasks on a given worker might decrease. This is not a huge deal since in the next loop iteration the worker would be selected, but we were aiming for correctness.

If it truly is causing this much contention then I agree with you, we can just rely on DB locking as the failure case is not that much of a problem.

aledeganopix4d:
@ddadlani sorry for nagging: do you think we can merge this one?
We're eager to try it in prod!

PR #4280, opened 2019-08-14
web, atc, skymarshal: split large tokens across multiple cookies

### LINKED ISSUE #3141: Support large JWT token in web UI
## Bug Report

Currently web UI stores a JWT in cookie (`skymarshal_auth`). Most browsers support cookies of up to 4096 bytes.
If JWT is too large, access to `/sky/userinfo` returns 401 since `skymarshal_auth` is not stored in the cookie. Authentication itself succeeds but the user cannot see any information related to the user in Web UI.
This can eventually happen when the number of teams the user belongs is increasing in a large foundation since a JWT contains all team names.
We saw this issue when then number of teams was more than 200. Concourse operators belonged to all teams.

## Steps to Reproduce

1. Assign a lot of teams to a user that makes the size of JWT over 4096 bytes.
1. Login to the Web UI. Log message says "Login successfully" but user info does not show in the UI. (`fly login` works)

## Expected Results

User info can be seen in Web UI regardless of the size of JWT.
I'd suggest to use LocalStorage instead of cookies. 

## Actual Results

User info can not be seen in Web UI if JWT cannot be stored in cookie.

## Version Info

* Concourse version: 4.2.2
* Infrastructure/IaaS: OpenStack
* Browser (if applicable): Chrome / Safari / Firefox

pivotal-jamie-klassen:
fixes #3141

# Release Note

```tex
\note{fix}{
  The web UI used to \link{silently break}{https://github.com/concourse/concourse/issues/3141}
  when your token (which includes a potentially-long JSON-encoded string detailing all the 
  teams you are part of and what roles you have on them) was longer than the size
  of a single cookie (4096 bytes on most browsers!). This limit has been increased 15-fold, which
  should unblock most users on clusters with a lot of teams \ghpr{4280}.
}
```


evanchaoli:
@pivotal-jamie-klassen Will this be included in release 5.5? I don't see it in 5.5 release note.

jomsie:
@evanchaoli Hi! I think the release note is here: https://github.com/concourse/concourse/releases#v550-note-8

evanchaoli:
@jomsie Thanks a lot.

Issue #4283, opened 2019-08-15
fly set-pipeline command should honor where fly is invoked with prompt unpause-pipeline


evanchaoli:
## Bug Report

`fly set-pipeline` prompts `unpause-pipeline` command after a pipeline is set, which is super useful, like:

    the pipeline is currently paused. to unpause, either:
      - run the unpause-pipeline command:
        fly -t test-runway unpause-pipeline -p appcheck
      - click play next to the pipeline in the web ui

But a tiny problem is, it seems to hardcode "fly".  As a Concourse admin, I manager multiple Concourse clusters with different versions, so I have multiple versions of fly as well. I rename fly commands to `fly52`, `fly53` and so on ... Because "fly" is hardcoded, I cannot simply copy the hint command.

## Steps to Reproduce

1. Rename fly command to like `fly52` or whatever
2.`fly52 -t target set-pipeline ...`

Then you will see the hint unpause-pipeline command show "fly" instead of "fly52", but "fly52" is expected, so that user can directly copy the hint command and paste/execute it.

## Expected Results

Like:

    the pipeline is currently paused. to unpause, either:
      - run the unpause-pipeline command:
        fly52 -t test-runway unpause-pipeline -p appcheck
      - click play next to the pipeline in the web ui

## Actual Results

    the pipeline is currently paused. to unpause, either:
      - run the unpause-pipeline command:
        fly -t test-runway unpause-pipeline -p appcheck
      - click play next to the pipeline in the web ui

## Version Info

* Concourse version: 5.2, 5.3
* Deployment type (BOSH/Docker/binary): binary
* Infrastructure/IaaS: VM
* Browser (if applicable): Not related to this issue
* Did this used to work? No

PR #4284, opened 2019-08-15
fly: when prompt a fly command to run, use full fly path the is invoked instead of just hardcode "fly".

evanchaoli:
Test:

    $ fly/fly -t tt sp -c pipeline.yml -p pp
    ...
    the pipeline is currently paused. to unpause, either:
      - run the unpause-pipeline command:
        fly/fly -t tt unpause-pipeline -p pp
     - click play next to the pipeline in the web ui

Fixes: #4283


evanchaoli:
@vito Finally, build passed. :)

PR #4293, opened 2019-08-19
[BUGFIX] Release lock if pool.ContainerInWorker errors.

aledeganopix4d:
Whenever ContainerInWorker returned an error the Active Tasks Lock
would never be released, blocking any task to check for available
workers and thus to start.

Here the bug is fixed, by releasing the lock if the ContainerInWorker
returns an error.

Additionally the unit tests are expanded to verify that for every
lock acquisition a lock release is also called.

### DIFF
--- a/atc/worker/client.go
+++ b/atc/worker/client.go
@@ -14,6 +14,7 @@ import (
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/atc/db/lock"
 	"github.com/concourse/concourse/atc/runtime"
+	multierror "github.com/hashicorp/go-multierror"
 )
 
 const taskProcessID = "task"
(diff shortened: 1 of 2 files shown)


aledeganopix4d:
@ddadlani when you have a moment, can you check this small PR out?
Thanks!

atc/worker/client.go
```diff
@@ -278,6 +278,10 @@ func (client *client) chooseTaskWorker(
 			}
 			existingContainer, err = client.pool.ContainerInWorker(logger, owner, containerSpec, workerSpec)
 			if err != nil {
+				release_err := activeTasksLock.Release()
+				if release_err != nil {
+					return nil, release_err
```
cirocosta: Hey,

Wdyt of leveraging a `multierror` (e.g., https://github.com/concourse/concourse/blob/da466557298f5c4f146064a80065dca1cd07f9ac/atc/gc/volume_collector.go#L39) here? 

This way we wouldn't lose the original error that came from `client.pool.ContainerInWorker`.

Thanks!

aledeganopix4d: Yes! Good point!

PR #4299, opened 2019-08-19
Allow task.run.dir to be absolute

### LINKED ISSUE #4281: Provide full path to build directory via a variable

TL;DR: I need my environment variable `GOOGLE_APPLICATION_CREDENTIALS` (defined in `params`) to point to the **absolute path** of my build directory.

I have the following task:

```
platform: linux

image_resource:
  type: docker-image
  source:
    repository: gcr.io/kaniko-project/executor:latest

inputs:
- name: source
- name: credentials

outputs:
- name: image

run:
  path: /kaniko/executor
  args:
  - --dockerfile=source/Dockerfile
  - --context=source/
  - --tarPath=image/image.tar
  - --no-push
params:
  GOOGLE_APPLICATION_CREDENTIALS: credentials/google_application_credentials.json
```

- The `credentials` input is a folder that contains
- a `google_application_credentials.json` file, 
- which is used by `/kaniko/executor`, 
- via the environment variable `GOOGLE_APPLICATION_CREDENTIALS`, to authenticate against a private docker registry.

Unfortunately, when this task run, the environment variable value is a **relative** path (`credentials/google_application_credentials.json`) to the *actual* **build directory** in which the task is running (in this case `/tmp/build/sha1/`).

That is, at the time of execution the `/kaniko/executor` binary will try to read the file at `credentials/google_application_credentials.json`, which isn't a valid path since this environment variable is supposed to point to an absolute path.

I need my environment variable `GOOGLE_APPLICATION_CREDENTIALS` (defined in `params`) to point to the **absolute path** of my build directory.

The common pattern in the concourse eco-system is to write a thin shell wrapper around the executable provided by the docker image as such:

```
[OMMITED CONFIGURATION FOR EASE OF READ]

run:
  # Run a shell instead of the binary directly
  path: /bin/sh 
  args:
  - -c
  - |
    set -e
    readonly WORKSPACE_DIR="$(pwd)"
    export readonly GOOGLE_APPLICATION_CREDENTIALS="${WORKSPACE_DIR}/credentials/google_application_credentials.json"
    exec /kaniko/executor \
      --dockerfile=source/Dockerfile \
      --context=source/ \
      --tarPath=image/image.tar \
      --no-push
```

However, this solution **only works if the image contains a shell**.

In this instance, the `gcr.io/kaniko-project/executor:latest` does not contain a shell, and this solution does not apply.

Concourse-CI needs to provide a way for pipelines to be fed their build directory. My limited knowledge about concourse does not allow me to recommend one way to do it over another, however, as a concourse user, here's what would feel the most natural: a concourse provided variable like `((concourse-build-folder))`

```
[OMMITED CONFIGURATION FOR EASE OF READ]

run:
  path: /kaniko/executor
  args:
  - --dockerfile=source/Dockerfile
  - --context=source/
  - --tarPath=image/image.tar
  - --no-push
params:
  GOOGLE_APPLICATION_CREDENTIALS: ((concourse-build-folder))/credentials/google_application_credentials.json
```

A variable provided by concourse would be in line with the philosophy of preventing users from tying themselves to how concourse runs builds, yet, give them the possibility of discovering the `build path` at runtime, without having to resort to anti-pattern solutions.

I would be more than happy to help implementing any solution allowing me to write my pipelines without expecting the presence of a shell.

huguesalary:
Fixes #4281

### DIFF
--- a/atc/engine/builder/step_factory.go
+++ b/atc/engine/builder/step_factory.go
@@ -1,10 +1,6 @@
 package builder
 
 import (
-	"crypto/sha1"
-	"fmt"
-	"path/filepath"
-
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/atc/db/lock"
@@ -123,9 +119,6 @@ func (factory *stepFactory) TaskStep(
 	containerMetadata db.ContainerMetadata,
 	delegate exec.TaskDelegate,
 ) exec.Step {
-	sum := sha1.Sum([]byte(plan.Task.Name))
-	containerMetadata.WorkingDirectory = filepath.Join("/tmp", "build", fmt.Sprintf("%x", sum[:4]))
-
 	taskStep := exec.NewTaskStep(
 		plan.ID,
 		*plan.Task,
--- a/atc/exec/task_step.go
+++ b/atc/exec/task_step.go
@@ -2,6 +2,7 @@ package exec
 
 import (
 	"context"
+	"crypto/sha1"
 	"fmt"
 	"io"
 	"path"
@@ -15,6 +16,7 @@ import (
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/atc/db/lock"
 	"github.com/concourse/concourse/atc/exec/build"
+	"github.com/concourse/concourse/atc/resource"
 	"github.com/concourse/concourse/atc/runtime"
 	"github.com/concourse/concourse/atc/worker"
 	"github.com/concourse/concourse/vars"
@@ -181,6 +183,13 @@ func (step *TaskStep) Run(ctx context.Context, state RunState) error {
 		config.Limits.Memory = step.defaultLimits.Memory
 	}
 
+	if filepath.IsAbs(config.Run.Dir) {
+		step.containerMetadata.WorkingDirectory = config.Run.Dir
+	} else {
+		sum := sha1.Sum([]byte(step.plan.Name))
+		step.containerMetadata.WorkingDirectory = filepath.Join(resource.ResourcesDir(fmt.Sprintf("%x", sum[:4])), config.Run.Dir)
+	}
+
 	step.delegate.Initializing(logger)
 
 	workerSpec, err := step.workerSpec(logger, resourceTypes, repository, config)
@@ -196,7 +205,7 @@ func (step *TaskStep) Run(ctx context.Context, state RunState) error {
 	processSpec := runtime.ProcessSpec{
 		Path:         config.Run.Path,
 		Args:         config.Run.Args,
-		Dir:          config.Run.Dir,
+		Dir:          step.containerMetadata.WorkingDirectory,
 		StdoutWriter: step.delegate.Stdout(),
 		StderrWriter: step.delegate.Stderr(),
 	}
--- a/atc/worker/client.go
+++ b/atc/worker/client.go
@@ -5,7 +5,6 @@ import (
 	"crypto/sha256"
 	"fmt"
 	"io"
-	"path"
 	"strconv"
 	"time"
 
@@ -260,7 +259,7 @@ func (client *client) RunTaskStep(
 				Path: processSpec.Path,
 				Args: processSpec.Args,
 
-				Dir: path.Join(metadata.WorkingDirectory, processSpec.Dir),
+				Dir: processSpec.Dir,
 
 				// Guardian sets the default TTY window size to width: 80, height: 24,
 				// which creates ANSI control sequences that do not work with other window sizes


huguesalary:
Seems like I was a bit too eager to get this PR out, the code isn't enough to implement the feature it's trying to implement.

vito:
The value you'd need to change is here: https://github.com/concourse/concourse/blob/6f775419932491f565cc8f49b42b93753bc61965/atc/engine/builder/step_factory.go#L128

But you'll have to change it based on the task config, which can be loaded at runtime: https://github.com/concourse/concourse/blob/9f3235d418b9ea20d2de8013b8f17c658fb60071/atc/exec/task_step.go#L159

I think you might be able to find your way by just finding all the `metadata.WorkingDirectory` usage and updating them.

huguesalary:
testflight seems to fail, but I'm not sure it is because of my code

atc/engine/builder/step_factory.go
```diff
@@ -124,8 +124,18 @@ func (factory *stepFactory) TaskStep(
 	containerMetadata db.ContainerMetadata,
 	delegate exec.TaskDelegate,
 ) exec.Step {
-	sum := sha1.Sum([]byte(plan.Task.Name))
-	containerMetadata.WorkingDirectory = filepath.Join("/tmp", "build", fmt.Sprintf("%x", sum[:4]))
+
+	var wd string
+	configRunDir := plan.Task.Config.Run.Dir
```
vito: This will only work if the task has its config statically provided like this:

```yaml
task: foo
config: # ...
```

For this to also work with `file:` you'll need to modify `task_step.go` to determine this after the config is loaded instead.

huguesalary:
Thanks! 

I'll look into your comments. Open to any suggestions as to write this code better, since I'm really not familiar with the codebase.

deniseyu:
related docs PR that we'll probably want to amend once this is merged: https://github.com/concourse/docs/pull/241

vito:
@huguesalary It looks like the unit tests are failing; some tests are asserting against the working directory that they set up, but now it's being overridden via the new code.

Also it probably makes sense to add a `testflight` test for this too. Let me know if you need any help with that!

huguesalary:
Thanks @deniseyu 

@vito I tried modifying the unit tests for a few hours, but I have to admit those are over my head. I'd definitely appreciate some help for both the testflight and the unit tests.

deniseyu:
@huguesalary @vito I'll take a look at the tests and have a try at fixing them!

deniseyu:
It looks like there are some wiring problems that were introduced by the new filepath generation and persistence logic. I commented out [the absolute path branch of the conditional](https://github.com/concourse/concourse/pull/4299/files#diff-72838e316ad978ca133c3c2c1bfb6b31R178-R180), and also tried dropping the old logic (`filepath.Join("/tmp", "build", fmt.Sprintf("%x", sum[:4]))`) into the second branch of the conditional, which did not get the tests back to a working state.

It seems like updating `step.containerMetadata.WorkingDirectory` has further-reaching impacts, which is why seemingly unrelated tests are failing. For example, the `step.registerOutputs` call takes the metadata as an argument, and later on it [compares the metadata.WorkingDirectory with mountpaths in the list of associated volume mounts](https://github.com/concourse/concourse/blob/master/atc/exec/task_step.go#L395-L415) to determine if a new output source should be added. The tests towards the end of `task_step_test.go` are failing because the volume mount names haven't been similarly modified for absolute paths!

I don't have too much context on whether it's important to leave volume mount paths as is. But maybe if we extract out the path construction logic into a named function like `BuildWorkingDir` or something, we could keep the old behaviour intact by passing both the volume mount path and the new absolute filepath through the same function.

Otherwise most of the test failures look to be hardcoded strings, which we could potentially update but it would introduce the risk of more magic strings. So another advantage of splitting out the `if filepath.IsAbs` logic into a separate, stateless function is that we can use it in the tests to transform strings as well. And it would be easy to add unit tests for that new function!

Let me know if any of that doesn't make sense, and if you need extra tips on running the unit test suite locally!

huguesalary:
Thanks much for looking into this @deniseyu and the details, it all makes sense.

To add to what you said, I also noticed a few test that hardcode `/tmp/build/get` and `/tmp/build/put` and other similar paths instead of using `resource.ResourcesDir()` (although it technically does the exact same thing), those tests still pass, despite my changes, which I found surprising.

I think what I need to do next to keep this PR going is just setup my debugger. I saw the documentation about that and was able to get a debugger shell, but I'd prefer to get a remote vscode debugger setup. (If I get that working I'll submit a documentation PR).

Once I have a debugger I'll be much easier for me to understand how the code is architectured.

pivotal-jamie-klassen:
@huguesalary just yesterday #4593 was merged, which gives an option for connecting your IDE to the debugger

xtremerui:
@huguesalary if you could fix the conflict then we are good to go. Thx!

huguesalary:
@xtremerui oof it's been a few months since I submitted this PR, not sure where I left off unfortunatey. 

I'll try sometime this week/weekend to setup my development environment again and look at the conflict and at least fix that, but I just have a feeling that there might be more work needed than just fixing the conflict.

aoldershaw:
Hey @huguesalary, sorry for not staying more on top of this! I know it's been a while, but if you're still interested in getting this through, I'd be happy to try to help get the tests green.

I haven't yet tried rebuilding Concourse with these changes to confirm that it works, but I think you can achieve the behaviour you're looking for by moving [this change](https://github.com/concourse/concourse/pull/4299/files#diff-72838e316ad978ca133c3c2c1bfb6b31R186-R192) to [here](https://github.com/concourse/concourse/pull/4299/files#diff-c5a89110a80683460828b35613e51245L263) - something like the following:

`atc/worker/client.go:256`
```go
		...
		dir := processSpec.Dir
		if !filepath.IsAbs(processSpec.Dir) {
			dir = path.Join(metadata.WorkingDirectory, dir)
		}

		process, err = container.Run(
				...
				Dir: dir,
				...
		)
		...
```

And I think that may be the only change that's necessary ([this one](https://github.com/concourse/concourse/pull/4299/files#diff-7e37c202e7044cc08233052392ecdb38L126-L128) can be reverted).

I'll test this out soon to confirm it works.

aoldershaw:
Sure enough, this patch seems to work. With the following job config:

`job.yml`
```
platform: linux
image_resource:
  type: registry-image
  source: {repository: ubuntu}
run:
  path: pwd
  dir: /test
```

Running `fly -t dev execute -c job.yml` outputs:
```
initializing
running pwd
/test
succeeded
```
Changing `dir` to be a relative path instead still works:

`job.yml`
```
platform: linux
image_resource:
  type: registry-image
  source: {repository: ubuntu}
run:
  path: pwd
  dir: test
```

Outputs:

```
initializing
running pwd
/tmp/build/e55deab7/test
succeeded
```

huguesalary:
Hi @aoldershaw, thank you for looking into this!

> Sure enough, this patch seems to work

Do you mean it passes the test suite? (specifically the concourse-ci/unit).

Feel free to send me a patch and I'll gladly apply it to the branch and see if the tests work.

aoldershaw:
@huguesalary I haven't run it through all the tests, but the tests that were failing no longer seem to be.

> Sure enough, this patch seems to work

By this, I mean it passes the test suite (at least, the part that was failing) and tasks with `task.run.dir` as an absolute path seem to run in that directory (see the example tasks I provided [here](https://github.com/concourse/concourse/pull/4299#issuecomment-614044105)).

The patch can be found here: https://gist.github.com/aoldershaw/e4856270a3f7c6512436ac4e5f6d0026

By the way, we should add some tests for this feature. Let me know if you'd like some help with that!

aoldershaw:
Hi @huguesalary, I just wanted to follow up with you to see if you're still interested in working on this. No worries if you've been busy - I just wanted to check in.

Again, if you are interested in working on this, I'm more than happy to help you out however I can!

huguesalary:
Oops @aoldershaw indeed, I completely forgot about this! I won't be able to work on this at the moment unfortunately, sorry! Please feel free to do anything you feel is necessary to get it merged.

PR #4300, opened 2019-08-19
Add scroll in HD view

maxknee:
Adding `overflow: scroll`  for HD view

Signed-off-by: Max Knee <Max_Knee@comcast.com>

### DIFF
--- a/web/assets/css/dashboard.less
+++ b/web/assets/css/dashboard.less
@@ -69,6 +69,8 @@
 
     .pipeline-wrapper {
       display: flex;
+      overflow: scroll;
+      max-height: 95%;
     }
 
     .drop-area {


pivotal-jamie-klassen:
Hi @maxknee, I'm having some trouble understanding this change. The title
mentions scrolling behaviour on the HD dashboard, but the change here applies to
styles on the `.pipeline-wrapper` class. To the best of my knowledge this class
only appears in the non-HD case. Would you mind describing the problem you're
addressing here?

maxknee:
@pivotal-jamie-klassen I was seeing at that in HD view, if you have pipelines that overflow in the `pipeline-wrapper` class in HD was preventing the ability to see all pipelines.
Without `overflow: scroll;`
Screen Shot 2019-08-21 at 4 12 09 PM

With `overflow: scroll;`
 
Screen Shot 2019-08-21 at 4 12 40 PM

pivotal-jamie-klassen:
I can't seem to reproduce your scrolling issue. Using the code on maste,r when I have a lot of pipelines I see something like this on the HD dashboard:
Screen Shot 2019-08-23 at 5 44 33 PM
with a horizontal scroll. Furthermore when I deploy your change, I see a bunch of extraneous scrollbars on the non-HD dashboard:
Screen Shot 2019-08-23 at 5 43 56 PM

It's really got me wondering what version you were running when you noticed that vertical overflow issue, and whether you might try deploying the latest code from master and see if it recurs?

Issue #3972, opened 2019-06-04
Remove ability to see secret in build logs


ashleystendel:
# Feature request
## What challenge are you facing?
We currently use Concourse with Vault as our credential manager. The majority of tasks files are written in Bash. As secrets from Vault are set as environment variables, developers are able to either echo the environment variable or add a set -x to their script and be able to see all secrets.  

## What would make this better?
Redact all secrets coming from a credential manager. 

## Are you interested in implementing this yourself?
No.

jchesterpivotal:
Related: https://github.com/concourse/concourse/issues/1469

Previously I've pointed folks to `concourse-filter`: https://github.com/pivotal-cf-experimental/concourse-filter

I don't fully agree with @vito's reason in #1469; it might be worth revisiting if event delivery/rendering changes.

vito:
I'm still skeptical of having people rely on this behavior but it would be pretty nifty to do it automatically based on credential values that we know we fetched from the credential manager for the build.

ukabu:
I agree that we should not rely on this to be 100% reliable, but other CI tools already do that. Also this would prevent easy access to credentials by running one-off tasks (sure if you encode the credential, they will still show up). And this will prevent accidental exposure.
We could appease ou security people a bit by having that feature.

evanchaoli:
This feature is also important to our users. Redacting based on credential manager is a good idea. I think we can redact before store build logs into database, so that it also avoids leaking secrets via database.

evanchaoli:
I just read through related code today, and realized that it's not easy to redact credential manager stored secrets in build logs. The main problem is that, secret evaluation is done by a third party module `github.com/cloudfoundry/boah-ci/director/template`, and there are two restrictions from this module:

1. The function `Template.Evaluate()` doesn't return information of which vars are populated, so that in Concourse side, it won't know what will be replaced in build log.

2. We only want to replace credential manager stored secrets, but two types of variables, static and cred, might be passed to `Template.Evaluate()`. Obviously `Template.Evaluate()` should not handle Concourse specific data types.

From the other perspective, blindly redacting all credential manager stored things in build log might not be a good idea. I have seen some crazy users not only store secrets, but also configure files in Vault (because Concourse can only fetch vars from credential manager for standalone task files). So that things stored in Vault doesn't really secrets. Redacting all of them will 1) generated unexpected build log, 2) add too much workload to ATC.

[...]

PR #4311, opened 2019-08-22
feat: redacting secrets in build logs.

evanchaoli:
I have done an initial version. But since the change is big, I would call for an initial review from design perspective first before I spend time on updating tests.

My testing pipeline is:

    jobs:
    - name: test-cred
      public: true
      plan:
      - task: run
        config:
          platform: linux
          image_resource:
            type: registry-image
            source: {repository: busybox}
          run:
            path: sh
            args:
             - -exc
             - |
               echo The name is ((localname))
               echo The password is ((test.pass)), and say ((hello))

Where `((test.pass))` and `((hello))` are defined in Vault.

Pipeline execution screen shot:
redact

Highlights of the design:

* Tracks credential interpolation with VarsTracker, and makes it public.
* Adds a BuildStepDelegate parameter to TaskConfigSource.FetchConfig, so that it can install a build log filter to delegate.
* Redact credential with a filter function before output build logs in delegate.

Given not all users are sensitive with credential, and redact could lead to some performance drop, I'm thinking to add a CLI option to `concourse web` process to control if or not enable this redacting feature, by default, the switch will be off.

This PR will resolve issue #3972.

@vito Your comments are wanted.

### DIFF
--- a/atc/engine/builder/builder.go
+++ b/atc/engine/builder/builder.go
@@ -6,8 +6,10 @@ import (
 	"strings"
 
 	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/atc/creds"
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/atc/exec"
+	"github.com/concourse/concourse/vars"
 )
 
 const supportedSchema = "exec.v2"
@@ -26,22 +28,26 @@ type StepFactory interface {
 //go:generate counterfeiter . DelegateFactory
 
 type DelegateFactory interface {
-	GetDelegate(db.Build, atc.PlanID) exec.GetDelegate
-	PutDelegate(db.Build, atc.PlanID) exec.PutDelegate
-	TaskDelegate(db.Build, atc.PlanID) exec.TaskDelegate
+	GetDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.GetDelegate
+	PutDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.PutDelegate
+	TaskDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.TaskDelegate
 	CheckDelegate(db.Check, atc.PlanID) exec.CheckDelegate
-	BuildStepDelegate(db.Build, atc.PlanID) exec.BuildStepDelegate
+	BuildStepDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.BuildStepDelegate
 }
 
 func NewStepBuilder(
 	stepFactory StepFactory,
 	delegateFactory DelegateFactory,
 	externalURL string,
+	secrets creds.Secrets,
+	redactSecrets bool,
 ) *stepBuilder {
 	return &stepBuilder{
 		stepFactory:     stepFactory,
 		delegateFactory: delegateFactory,
 		externalURL:     externalURL,
+		secrets:         secrets,
+		redactSecrets:   redactSecrets,
 	}
 }
 
@@ -49,10 +55,11 @@ type stepBuilder struct {
 	stepFactory     StepFactory
 	delegateFactory DelegateFactory
 	externalURL     string
+	secrets         creds.Secrets
+	redactSecrets   bool
 }
 
 func (builder *stepBuilder) BuildStep(build db.Build) (exec.Step, error) {
-
 	if build == nil {
 		return exec.IdentityStep{}, errors.New("Must provide a build")
 	}
@@ -61,7 +68,8 @@ func (builder *stepBuilder) BuildStep(build db.Build) (exec.Step, error) {
 		return exec.IdentityStep{}, errors.New("Schema not supported")
 	}
 
-	return builder.buildStep(build, build.PrivatePlan()), nil
+	credVarsTracker := vars.NewCredVarsTracker(creds.NewVariables(builder.secrets, build.TeamName(), build.PipelineName()), builder.redactSecrets)
+	return builder.buildStep(build, build.PrivatePlan(), credVarsTracker), nil
 }
 
 func (builder *stepBuilder) CheckStep(check db.Check) (exec.Step, error) {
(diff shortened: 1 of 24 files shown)


atc/engine/builder/delegate_factory.go
```diff
@@ -357,9 +367,15 @@ func (writer *dbEventWriter) Write(data []byte) (int, error) {
 
 	writer.dangling = nil
 
+	payload := string(text)
+	if writer.filter != nil {
+		fmt.Println("EVAN: going through filter: %s", payload)
+		payload = writer.filter(payload)
```
vito: Filtering here may not always work as any credential could end up being split into separate `Write()` calls. This is more likely for multi-line credentials like SSH keys, but theoretically it could happen at any time, since `Write()` receives arbitrary chunks of output based largely on timing.

This aspect feels like a tricky problem to solve. :thinking:

A few ideas:

* Watch for segments of credentials instead of the entire value. For example, maybe for multi-line values we could redact each line independently, no matter where it shows up? This could also aid in cases where a multiline value is printed in a marshalled form, e.g. `"foo\nbar"` would result in `"[redacted **x**]\n[redacted **x**]"`.
* Watch out for things that match a prefix of a credential (I suppose up to some threshold). This would likely require buffering somewhere. Maybe there could be an `io.Writer` that "holds on" to the buffer until it doesn't match the prefix of any configured credentials?
* Both?

vars/template.go
```diff
@@ -264,6 +275,23 @@ func (t varsTracker) ExtraError() error {
 	return UnusedVarsError{Vars: names(unusedNames)}
 }
 
+var credTypeNameMatcher = regexp.MustCompile("creds\\.")
```
vito: What's this about?

evanchaoli: With my new implementation, this is no longer needed.

atc/engine/builder/delegate_factory.go
evanchaoli: I have an idea. For those multi-line secrets, can we just redact few characters for each line? Which would significantly mitigate random chunk problem.

evanchaoli:
Based on @vito 's comment, I enhanced the initial implementation. Highlights of the second implementation:

* Add a `CredVarsTracker` struct that wraps credential manager, and tracks interpolated secrets.
* Add a `CredVarsTracker` object in `engineBuild` struct, so that a build hold a single tracker.
* Pass `CredVarsTracker` object to all build step delegate.
* `CredVarsTracker` implements interface `Variables`, so that in get/put/task steps, `step.delegate.GetCredVarsTracker()` can be used as credential variables (original created by `creds.NewVariables`).
* With all above changes, the build's `CredVarsTracker` object holds interpolated secrets from all build steps, thus redact build logs.

My test pipeline:

    resource_types:
     - name: rt-tiny
       type: registry-image
       source:
         repository: evanli/tiny
         tag: v2
     
    resources:
    - name: tiny
      type: rt-tiny
      source:
       pass: ((test.pass))
    
    jobs:
    - name: test-cred
       public: true
       plan:
       - put: tiny
        params:
          key: ((hello))
     
       - task: task1
         config:
          platform: linux
          image_resource:
            type: registry-image
            source: {repository: busybox}
          outputs:
          - name: out
          run:
            path: sh
            args:
             - -exc
             - |
               echo The name is ((localname))
               echo The password is ((test.pass)), and say ((hello))
               echo ((hello)) > out/a.txt
    
       - task: task2
         config:
           platform: linux
           image_resource:
             type: registry-image
             source: {repository: busybox}
           inputs:
           - name: out
           run:
             path: sh
             args:
              - -exc
              - |
                cat out/a.txt

The resource `tiny` do nothing, just output `source` and `params` to build logs. And `task1` passes a secret to `task2` via file, then `task2` print the secrets to build log. So that this pipeline has covered redacts build logs of `get`, `put` and `task`.

redact2

From the above screen shot, we can see all secrets have been redacted.

@vito For the chunky build log issue, as you said, it's tricky to solve, and need more discussion.

vito:
> @vito For the chunky build log issue, as you said, it's tricky to solve, and need more discussion. I think that issue can be implemented with a separate PR. Given this PR changes a lot of files, rebasing the changes is a big pain. Given the feature is by default turned off, if you are ok with my current code, can we get this PR merged first? If you are ok with the current design and my suggestion about chunky problem, then I'm going to update unit tests and raise the final version of code change.

Totally fair. We can start small and iterate on it. :slightly_smiling_face: The new changes sound great, I'll try to review it today!

In the meantime, it looks like there are a couple build failures (I think fakes just need to be regenerated), and the DCO bot is unhappy: https://github.com/concourse/concourse/pull/4311/checks?check_run_id=212761804

atc/engine/builder/delegate_factory.go
```diff
@@ -277,25 +280,46 @@ func (d *taskDelegate) Finished(logger lager.Logger, exitStatus exec.ExitStatus)
...
 func (delegate *buildStepDelegate) ImageVersionDetermined(resourceCache db.UsedResourceCache) error {
 	return delegate.build.SaveImageResourceVersion(resourceCache)
 }
 
+type credVarsIterator struct {
+	line string
+}
+
+func (it *credVarsIterator) YieldCred(k, v string) {
+	it.line = strings.Replace(it.line, v, fmt.Sprintf("[redacted **%s**]", k), -1)
```
vito: Do you think there's any security concern with printing the var that was redacted here? :thinking: 

I can't think of a great example, but if someone were to be able to deduce (from context) the value that was redacted - e.g. a username - then the var may reveal what service the username is for. Granted, they'd still need the password, but I suspect some folks would be uncomfortable with any amount of information leakage.

If there's no real reason to print the var it might be safer to omit it.

atc/engine/builder/delegate_factory.go
vito: Not sure what you mean - do you mean 'chunk' the values that we redact, too? so e.g. 'abcdefghijklmnop' would redact 'abc', 'def', 'ghi', 'jkl' etc. individually?

atc/engine/builder/step_factory.go
```diff
@@ -21,7 +20,7 @@ type stepFactory struct {
 	resourceFetcher       fetcher.Fetcher
 	resourceCacheFactory  db.ResourceCacheFactory
 	resourceConfigFactory db.ResourceConfigFactory
-	secretManager         creds.Secrets
+	//secretManager         creds.Secrets
```
vito: (just make sure to kill this prior to merge!)

atc/engine/engine.go
```diff
@@ -29,12 +31,14 @@ type Runnable interface {
 //go:generate counterfeiter . StepBuilder
 
 type StepBuilder interface {
-	BuildStep(db.Build) (exec.Step, error)
+	BuildStep(db.Build, vars.CredVarsTracker) (exec.Step, error)
```
vito: :thinking: I wonder if we could push `creds.Secrets` and `redactSecrets` into the `StepBuilder` instead?

The `StepBuilder` could construct the `vars.CredVarsTracker` on its own since it's already receiving the `db.Build`:

```go
vars.NewCredVarsTracker(creds.NewVariables(secrets, build.TeamName(), build.PipelineName()), redactSecrets)
```

atc/engine/builder/builder.go
```diff
@@ -63,185 +64,185 @@ func (builder *stepBuilder) BuildStep(build db.Build) (exec.Step, error) {
 		return exec.IdentityStep{}, errors.New("Schema not supported")
 	}
 
-	return builder.buildStep(build, build.PrivatePlan()), nil
+	return builder.buildStep(build, build.PrivatePlan(), credVarsTracker), nil
```
vito: (Follow-up from previous comment:) this is where it could construct the `credVarsTracker` itself if it had the secrets

atc/exec/get_step.go
```diff
@@ -121,7 +118,8 @@ func (step *GetStep) Run(ctx context.Context, state RunState) error {
 
 	step.delegate.Initializing(logger)
 
-	variables := creds.NewVariables(step.secrets, step.metadata.TeamName, step.metadata.PipelineName)
+	//variables := creds.NewVariables(step.secrets, step.metadata.TeamName, step.metadata.PipelineName)
+	variables := step.delegate.GetCredVarsTracker()
```
vito: Stylistically we prefer to not have `Get`-prefixed methods - can we just rename this to `Variables()` and have it return `creds.Variables` interface since that's the only interface the steps care about?

atc/engine/builder/delegate_factory.go
evanchaoli: I don't intend to leave var in redacted text. As this is just the initial change, I printed var in order to make debugging easier. I'll remove var in the final commit.

atc/engine/engine.go
evanchaoli: The problem is, we want an `vars.CredVarsTracker` object to be global to a build. If each step constructs a `vars.CredVarsTracker` object, then we cannot solve the case like a task interpolates some secrets then pass secrets to second task and second task print secrets.

atc/engine/builder/delegate_factory.go
evanchaoli: Sorry for the confusion. Let me be clearer. Say a multi-line ssh key like:

     dkjffkfgkgusjdjjdjshhsjsjsjjshdhddjdhdhdhdjd
     skfjfjekjfk834899rfddf89823893282382303
     slsl00090809009fdjsdkjelkfdjkjdllsdlkklsdkld
    ...

We don't have redact the entire ssh key, as long as we redact partial of it, such as redacting first 8 chars of each line, then the key will lose its meaning:

     [**redacted**]sjdjjdjshhsjsjsjjshdhddjdhdhdhdjd
     [**redacted**]34899rfddf89823893282382303
     [**redacted**]ere009fdjsdkjelkfdjkjdllsdlkklsdkld
    ...

This way, as long as there is 1 line is redacted, the key lose its meaning, so that it significantly reduced possibility of not redacting due to random chunks of build logs.

From implementation perspective, we can randomly choose some pieces from a long secrets to redact.

vito: Ah, gotcha. I don't think users would be comfortable with any amount of information leakage as it weakens the entropy of the secret. And if we randomly remove pieces it would potentially reveal the full secret over enough runs.

atc/exec/get_step.go
```diff
@@ -121,7 +118,7 @@ func (step *GetStep) Run(ctx context.Context, state RunState) error {
 
 	step.delegate.Initializing(logger)
 
-	variables := creds.NewVariables(step.secrets, step.metadata.TeamName, step.metadata.PipelineName)
+	variables := step.delegate.Variables()
```
vito: It's nice that we don't have to use the `step.metadata` for this anymore. :+1:

atc/exec/get_step_test.go
```diff
@@ -219,6 +220,12 @@ var _ = Describe("GetStep", func() {
 			Expect(resourceInstance.LockName("fake-worker")).To(Equal(expectedLockName))
 		})
 
+		It("secrets are tracked", func() {
```
vito: This test and its `get` + `task` equivalent feel a little too low-level and don't add much confidence that credentials are actually redacted.

In fact I was able to comment out the line that does the redacting and no tests under `atc/` failed. Can we make these tests a bit higher-level?

atc/engine/builder/builder.go
evanchaoli: Moved `credVarsTracker` construction to `stepBuilder`.

atc/engine/builder/delegate_factory.go
evanchaoli: Ok, let's leave this discussion to a separate PR.

atc/exec/get_step.go
evanchaoli: Nice to hear that.

atc/exec/get_step_test.go
evanchaoli: I have added some test cases in `delete_factory_test.go`, so that redacting are tested. Now you should not be able to pass test if you comment out any line.

atc/engine/engine.go
evanchaoli: Good point. I revisited the code again, and realized you are totally right. Constructing a tracker in `BuildStep` is more elegant. Have removed tracker from `engine.go` and moved to `builder.go`.

evanchaoli:
With the lasted code I have just updated, the screen shot is:

redact3

atc/engine/builder/delegate_factory_test.go
```diff
@@ -456,5 +460,60 @@ var _ = Describe("DelegateFactory", func() {
...
+				delegate.Variables().Get(vars.VariableDefinition{Name: "source-param"})
+			})
+
+			Context("Stdout", func() {
+				JustBeforeEach(func() {
+					writer = delegate.Stdout()
+					writtenBytes, writeErr = writer.Write([]byte("ok super-secret-source ok"))
+				})
+
+				It("should be redacted", func() {
```
vito: :+1: Nice! These tests feel pretty descriptive and high-level.

atc/engine/engine.go
```diff
@@ -37,16 +39,17 @@ type StepBuilder interface {
...
 		release:       make(chan bool),
 		trackedStates: new(sync.Map),
 		waitGroup:     new(sync.WaitGroup),
 	}
 }
 
 type engine struct {
-	builder StepBuilder
+	builder       StepBuilder
+	secrets       creds.Secrets
```
vito: I think this can be rolled back now that the `StepBuilder` has them

atc/exec/check_step.go
```diff
@@ -21,14 +20,15 @@ type CheckStep struct {
 	plan              atc.CheckPlan
 	metadata          StepMetadata
 	containerMetadata db.ContainerMetadata
-	secrets           creds.Secrets
```
vito: Looks like this never needed to be here, thanks for cleaning it up!

Issue #4196, opened 2019-07-30
[fly] Add `--team-name` for all Fly commands


taylorsilva:
Epic #4192 

Currently the only way for users to interact with multiple teams with Fly is to create multiple targets.
This story will leave this workflow possible (users may have scripts that switch targets based on teams)
and will enable a new "single target, many teams" workflow.

If a user wants to change their default team, they can login with the same target into a different team.

Another thing to note is that we already have the `--team-name` and `-n` flags for some commands. We should continue to use both of these flags everywhere.

### Single-team scoped Commands
For single-team scoped commands the following should be possible for each command. This ensures a user's existing script/workflow will still work.

```bash
# use default team
fly -t ci <command> <command-options>
# use another team
fly -t ci <command> <command-options> --team-name team3
```

- [ ] abort-build
- [x] hijack
- [ ] resources
- [ ] check-resource-type
- [ ] check-resource
- [ ] resource-versions
- [ ] watch
- [ ] checklist
- [ ] clear-task-cache
- [ ] execute
- [ ] expose-pipeline
- [ ] destroy-pipeline
- [ ] get-pipeline
- [ ] set-pipeline
- [ ] hide-pipeline
- [ ] pause-pipeline
- [ ] unpause-pipeline
- [ ] rename-pipeline
- [ ] order-pipelines
- [x] jobs
- [x] trigger-job
- [x] pause-job
- [x] unpause-job

PR: https://github.com/concourse/concourse/pull/4406
**PR is ready for review and can be merged**

Remaining commands should be submitted as new PR's. One PR per command to make it easier for others to review.

### Multi-team scoped commands
For these commands, they currently behave like the previous single-team commands. They hit a `/api/v1/teams/...` endpoint and return the result. We should change the behaviour of the `--team-name` to accept **one or more** team names by declaring the flag multiple times.
We can also implement a `--all-teams` flag to just get data for all teams in Concourse.

The following should be possible for the below commands:
```bash
# use default team
fly -t ci <command> <command-options>
# return data for specified team
fly -t ci <command> <command-options> --team-name team1
# return data for all teams given
fly -t ci <command> <command-options> --team-name team1 --team-name team2 -n team3
# return data for all teams in Concourse
fly -t ci <command> <command-options> --all-teams
```

- [x] containers
- [ ] volumes
- [ ] pipelines
- [x] builds (`--all` has been implemented, enhancement for `--team`)

[...]

PR #4324, opened 2019-08-28
WIP: [Fly] Multi-team-scoped fly commands can display results by team name

zoetian:
fixes: #4196

This PR is for

`fly -t TARGET containers` (global api endpoints needed)

Any team-scoped commands should accept a `-n` or `--team` parameter.


fly/commands/volumes.go
```diff
@@ -1,7 +1,9 @@
 package commands
 
 import (
+	"errors"
 	"fmt"
+	"github.com/concourse/concourse/go-concourse/concourse"
```
zoetian: add a newline to separate the go std lib

ddadlani:
Can we squash the WIP commits (or remove the WIP if we would like the commit message to stay on master)?

zoetian:
> Can we squash the WIP commits (or remove the WIP if we would like the commit message to stay on master)?

Thanks for the CR! 🎉 😃

Till today this PR is still not done yet, we're still working on it. I really don't know why it shows that I requested the CR on runtime yesterday...., (consider that I was working on another `helm` track). But when we're done, we'll definitely squash everything into 1-2 commits.

fly/integration/suite_test.go
```diff
@@ -22,12 +24,24 @@ var flyPath string
 var homeDir string
 
 var atcServer *ghttp.Server
+var loginATCServer *ghttp.Server
```
zoetian: change this as well

atc/db/team.go
```diff
@@ -151,77 +151,7 @@ func (t *team) FindWorkerForVolume(handle string) (Worker, bool, error) {
...
-		return nil, err
-	}
-
-	containers, err = scanContainers(rows, t.conn, containers)
-	if err != nil {
-		return nil, err
-	}
-
-	return containers, nil
+	return getContainersByTeamIds([]int{t.id}, false, t.conn)
```
vito: Do we need to bother with skipping the team name here (the `false`)? What's the harm in just always including it?

fly/commands/containers.go
```diff
@@ -26,9 +31,33 @@ func (command *ContainersCommand) Execute([]string) error {
 		return err
 	}
 
-	containers, err := target.Team().ListContainers(map[string]string{})
-	if err != nil {
-		return err
+	if len(command.Teams) > 0 && command.AllTeams {
+		return errors.New("Cannot specify both --all-teams and --team")
```
vito: nit:

```suggestion
		return errors.New("cannot specify both --all-teams and --team")
```

fly/commands/fly.go
```diff
@@ -81,4 +81,9 @@ type FlyCommand struct {
 	Completion CompletionCommand `command:"completion" description:"generate shell completion code"`
 }
 
+type TeamsParam struct {
+	AllTeams bool     `short:"a" long:"all-teams" description:"Run the command for all teams"`
+	Teams    []string `short:"n" long:"team" description:"Run the command for a list of teams"`
```
vito: :thinking: The short-form here clashes with `-n/--non-interactive` and doesn't share any characters with the long-form (`--team`). Any other ideas?

In a perfect world we would free up `-t/--target` for `-t/--team` instead and use some other name to identify the Concourse endpoint. Maybe `-c/--concourse`, but that's used for `-c/--config` in some places. Maybe `-e/--env`? Doesn't look like that's used anywhere.

Changing `-t/--target` at this point would break literally everyone's muscle memory, so we might want to kick that can down the road until we redesign `fly` so people get used to a whole new scheme all at once.

How about we just omit the short-form for now? Right now `fly` is really designed around having a team in the target, so until we redo the UX entirely maybe we should just have people use the long form.

zoetian: Hey @vito, thanks for the review. I take a look at the current usage of the `-n`, and it is really confusing with the normal `--non-interactive`. However, for some of the fly users they may be used to login/ set-pipeline/ see builds on team with `-n` as `--team`. I personally think it would be better to keep it for now till ux team actually have settled the idea on which short form to use in the future to reduce the amount of work if anyone is using the script on any related commands. Is that okay? 🤯
also, updated to use `TeamsFlags`

vito: @zoetian Hmm I think we should just omit the short-form for now. I would rather have folks explicitly spell out `--team` than specify a short flag which means something different than they might expect. We can remove the `-n` shorthand for `--team` in the other commands for consistency.

UX team isn't responsible for `fly` anymore; it's maintained as part of the Core/API track, and I think consistency is more important than anything else at least until we get around to a new flag/command scheme.

xtremerui:
@taylorsilva this PR is part of an EPIC. What is the status atm? Do you still need this PR? Thx!

taylorsilva:
Changed title to WIP. I need to review this PR again, fix the conflicts, and potentially make changes to it as well before it's ready to merge.

taylorsilva:
Closing this PR. We'll think about how we want to implement these features later.

PR #4328, opened 2019-08-29
use JSON unmarshal for job config, switch to sigs.k8s.io/yaml

### LINKED ISSUE #4327: High CPU usage on ATC
## Bug Report

We've deployed a recent RC on our production environment, one based on 7c63f45bc1dd146b790e17fdad008e76d3487356.
Immediately after the deployment the CPU usage of the ATC jumped to ~99% and stays there almost constantly.
Whether before we would have an average usage around ~60% (both with v5.3.0 and an RC based on 56721145ebe9c3a02cf45183d86469758de55716)

## Steps to Reproduce

Deploying 7c63f45bc1dd146b790e17fdad008e76d3487356 from 56721145ebe9c3a02cf45183d86469758de55716, with all the pipelines/jobs/teams staying the same.

## Additional Context

So far the CPU usage is only seen through metrics, no impact on the final users.
We are testing the container placement strategy `limit-active-tasks` with a maximum `active_tasks == 1`, but I don't have reason to believe that's related.

I'm attaching a profile taken with pprof.

[pprof.concourse.samples.cpu.001.pb.gz](https://github.com/concourse/concourse/files/3556140/pprof.concourse.samples.cpu.001.pb.gz)

vito:
fixes #4327


pivotal-jwinters:
How come we're switching to this fork?

xtremerui:
> How come we're switching to this fork?

regression? I haven't tested myself but from the referenced issue the current yaml library causes high CPU @pivotal-jwinters

xtremerui:
I am trying to compare the cpu usage with/o this fix in my local docker compose with a set of 30 pipelines, which have 250 jobs in total. Can't really tell the difference in this scale. Both pprof profiles are normal and the `scanJob` is fast.

Will need to know the total number of jobs and the average size of config of each job in the failing case.

xtremerui:
A following set up seems good enough for the comparison.

Env: local docker
Pipelines: 100
jobs: 1000
Web UI: 3 dashboard opened at the same time

pprof result without fix:
Screen Shot 2019-08-29 at 3 34 12 PM

pprof result with fix:
Screen Shot 2019-08-29 at 3 35 49 PM

the ListAllJobs cpu usage reduced from ~4.8s to ~0.8s

vito:
The fork is unrelated to the CPU perf fix. The original dependency is unmaintained. The k8s fork is maintained.

atc/db/job.go
```diff
@@ -848,7 +847,7 @@ func scanJob(j *job, row scannable) error {
 	}
 
 	var config atc.JobConfig
-	err = yaml.Unmarshal(decryptedConfig, &config)
+	err = json.Unmarshal(decryptedConfig, &config)
```
pivotal-jwinters: So was this the performance culprit?

xtremerui: I think so... the pprof shows exactly that difference 
scanJob-> json.Unmarshal vs scanJob -> yaml.unmarshal

pivotal-jwinters: oh ya, neat!

vito: Yep the fix itself is just this line. :+1:

pivotal-jamie-klassen:
@vito needs a rebase, sorry!

PR #4333, opened 2019-08-30
Task lock team platform

aledeganopix4d:
The ActiveTask lock is acquired when a task wants to select a worker where to execute itself. The pools of workers where the task can run are separated by team and platform, thus having one lock for each
of these combination can improve the time it takes for unrelated, concurrent tasks to acquire a lock and
verify/reserve a worker where to be executed.

Also add `LockTypeActiveTasks` to the metrics as it was previously forgotten.

### DIFF
--- a/atc/db/lock/lock.go
+++ b/atc/db/lock/lock.go
@@ -49,8 +49,8 @@ func NewDatabaseMigrationLockID() LockID {
 	return LockID{LockTypeDatabaseMigration}
 }
 
-func NewActiveTasksLockID() LockID {
-	return LockID{LockTypeActiveTasks}
+func NewActiveTasksLockID(platform string, teamID int) LockID {
+	return LockID{LockTypeActiveTasks, lockIDFromString(platform + strconv.Itoa(teamID))}
 }
 
 //go:generate counterfeiter . LockFactory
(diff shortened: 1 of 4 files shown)


ddadlani:
Hi @aledeganopix4d, thanks for working on this!

I don't fully understand the motivation behind this change. How much improvement does the addition of `platform` and `team` provide? It shouldn't take that long to select a worker for the task and then release the lock, IMO. Could you share some insights regarding the issue this is meant to fix?

aledeganopix4d:
> Hi @aledeganopix4d, thanks for working on this!
> 
> I don't fully understand the motivation behind this change. How much improvement does the addition of `platform` and `team` provide? It shouldn't take that long to select a worker for the task and then release the lock, IMO. Could you share some insights regarding the issue this is meant to fix?

Hi @ddadlani,
Our deployment does not have enough teams to notice a difference with this PR.
The simple reasoning behind the change is this should be a flat improvement at "0 cost", but I might be wrong. If you're not convinced by the change we can skip this.

Thanks.

vito:
One problem: many Concourse instances have a pool of shared cross-team workers as well, so making the lock per-team won't help in that case.

ddadlani:
To add on to @vito's comment, if we're optimizing for the type of workers by `platform` and `team`, we should arguably also do `resourcetype` and `tags` but it honestly seems like an additional complication for not much benefit. I'm going to close this for now, at least until we see a good use case to optimize the lock in this way.

@aledeganopix4d would you be able to add the lock type to metrics as a different PR, though? We should tie that loose end up.

Thanks again!

PR #4335, opened 2019-08-30
CSS change to create sticky step headers on scroll

### LINKED ISSUE #4230: UX: Sticky step headers while scrolling through large log output
When I have a bunch of steps expanded each displaying large amounts of logs. I'd like to know which step's logs I'm currently reading. Much like how team headers are sticky on the dashboard view, it would be helpful to have step headers stick while scrolling between steps.

matthewpereira:
Fixes #4230

### DIFF
--- a/release-notes/v5.5.0.md
+++ b/release-notes/v5.5.0.md
@@ -119,3 +119,7 @@
 #### <sub><sup><a name="v550-note-29" href="#v550-note-29">:link:</a></sup></sub> fix
 
 * Scrollbars have been given a touch of the Concourse design language. It should no longer use browser defaults. #4355
+
+#### <sub><sup><a name="v550-note-30" href="#v550-note-30">:link:</a></sup></sub> fix
+
+* Build step headers are now sticky, staying at the top of the body on the build page instead of scrolling out of sight as you navigate long logs. This should make it easier to keep track of which step you're inspecting in complicated jobs. #4230
--- a/release-notes/v5.5.0.md
+++ b/release-notes/v5.5.0.md
@@ -120,6 +120,6 @@
 
 * Scrollbars have been given a touch of the Concourse design language. It should no longer use browser defaults. #4355
 
-#### <sub><sup><a name="v550-note-30" href="#v550-note-30">:link:</a></sup></sub> fix
+#### <sub><sup><a name="v550-note-30" href="#v550-note-30">:link:</a></sup></sub> feature
 
 * Build step headers are now sticky, staying at the top of the body on the build page instead of scrolling out of sight as you navigate long logs. This should make it easier to keep track of which step you're inspecting in complicated jobs. #4230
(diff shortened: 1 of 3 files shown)


jomsie:
It looks so good, I'm re-kickin' off the pipelines as prod died momentarily.

kcmannem:
@jomsie plz approve

web/assets/css/layout.less
```diff
@@ -39,7 +39,7 @@ body,
   .scrollable-body {
     width: 100%;
     box-sizing: border-box;
-    padding: 10px;
```
jomsie: This looks awesome when scrolling large amounts of logs, just one note about this padding removal, it makes the build headers kinda go to the edge of the screen.

Padding removal:
Screen Shot 2019-08-30 at 4 03 56 PM


Master:
Screen Shot 2019-08-30 at 3 59 39 PM

Looks like if you add double the margin as padding-right (20px), it might solve it! Let me know what you think!

jomsie:
looks great! can be merged after ci passes!

web/assets/css/layout.less
matthewpereira: Good catch @jomsie. Unfortunately, I think that adding `padding-right` will push the scrollbar off the screen, like this:

image

... and if we start using a `calc()` on the `width` property to fix that, we still get a messy gap where the scrollbar doesn't meet with the build list above it.

I chatted with @pivotal-jamie-klassen this morning about this, and we decided not to mess with the margin/padding for now. I've added a commit that puts `layout.less` back to normal, and results in this:

v4w4H2Wv3g

I think the visual impact of logs scrolling 'past' the sticky headers is minimal compared to the potential layout overhaul needed to solve the spacing in a way that doesn't impact scrollbars. We can look into the build page layout in order to clean up `position: fixed` elements and make it a little more flexible in the future, and make this look even cleaner at that point in time.

web/assets/css/production.less
```diff
@@ -202,8 +202,10 @@ li.prep-status {
 .build-step .header {
   cursor: pointer;
   clear: both;
-  position: relative;
```
matthewpereira: @jomsie Sorry to pull your attention back to this - I realized we had a duplicate `position` tag here. Can you re-approve?

release-notes/v5.5.0.md
```diff
@@ -119,3 +119,7 @@
 #### <sub><sup><a name="v550-note-29" href="#v550-note-29">:link:</a></sup></sub> fix
 
 * Scrollbars have been given a touch of the Concourse design language. It should no longer use browser defaults. #4355
+
+#### <sub><sup><a name="v550-note-30" href="#v550-note-30">:link:</a></sup></sub> fix
```
pivotal-jamie-klassen: One final nit. Can you call this a `feature` rather than a `fix`? Folks have never had this before!

matthewpereira: Updated. I'm used to different nomenclature and rules - I was interpreting this as non-visible or difficult-to-keep-track-of step headers being 'fixed' so that they're now visible or easy to keep track of.

Thanks for the correction!

Issue #555, opened 2016-07-22
Build plan tasks using file definition of task should validate params


Amit-PivotalLabs:
If I have a build plan with a task defined by a `file: task.yml`, and in my task step in my build plan, I specify some params that _aren't_ specified in the task.yml, I should see a nice validation error.

In other words, as a task.yml author, I want better validation/control over how build plans consuming my task.yml pass in params.

> Alex Suraci: you don't need to declare them, which i want to fix
> Alex Suraci: would prefer to have a validation there, e.g. to call out that a task is consuming an undeclared param

[[link](https://concourseci.slack.com/archives/general/p1469219969001805)]

Amit-PivotalLabs:
Has this been fixed?  Recently tried a `VAR1=val1 VAR2=val2 fly -t myci execute -c conf.yml`.  It picked up `VAR1` but not `VAR2`, only difference I could see is my `conf.yml` doesn't specify `VAR2`. If it has been fixed, great!  If not, maybe I'm doing something else wrong...

vito:
It works in `fly execute` because `fly` only selects values from your environment that are present in the task .yml - otherwise it'd be sending along `$HOME`, `$USER`, `$SSH_AGENT_PID`, etc.

It's still not fixed in the pipeline case.

Amit-PivotalLabs:
Makes sense, thanks.

jmcarp:
Is the idea to throw a validation error on undeclared params in https://github.com/concourse/atc/blob/master/task.go#L116-L134? I'd be down to give that a try if it seems like the right approach.

vito:
@jmcarp 🐢 yep

vito:
This was addressed by https://github.com/concourse/atc/pull/297 - thanks @edtan!

chendrix:
@vito please don't turn this into an error without a way of turning this validation off.

We have a use case for dynamically specifying ENV vars via pipeline templating.

PR #4355, opened 2019-09-03
adding styling to scrollbars

kcmannem:
The default scrollbar was hurting my eyes. So I tried applying the concourse styling to it. I used colors from the sidebar item active/inactive states.

Screen Shot 2019-09-03 at 4 15 00 PM

### DIFF
--- a/release-notes/v5.5.0.md
+++ b/release-notes/v5.5.0.md
@@ -111,3 +111,7 @@
 #### <sub><sup><a name="v550-note-28" href="#v550-note-28">:link:</a></sup></sub> fix
 
 * @xavierzwirtz also fixed `hg.pull` by passing the `sourceUri` in the Mercurial resource. This could fix some issues where Mercurial wasn't saving the full uri in the `hgrc` file concourse/hg-resource#13.
+
+#### <sub><sup><a name="v550-note-29" href="#v550-note-29">:link:</a></sup></sub> fix
+
+* Scrollbars have been given a touch of the Concourse design language. It should no longer use browser defaults. #4355
--- a/web/assets/css/globals.less
+++ b/web/assets/css/globals.less
@@ -9,6 +9,7 @@ body {
   margin: 0;
   background: @base02;
   color: @base05;
+  scrollbar-color: #555 rgba(0,0,0,0);
 }
 
 pre {
@@ -28,3 +29,18 @@ h3 {
   font-size: 14px;
 }
 
+::-webkit-scrollbar {
+  width: 10px;
+}
+
+::-webkit-scrollbar-track {
+  background: rgba(0, 0, 0, 0);
+}
+
+::-webkit-scrollbar-thumb {
+  background: #555;
+}
+
+::-webkit-scrollbar-thumb:hover {
+  background: #707070;
+}


kcmannem:
I don't think the code is actually correct, I don't really know if I even did the right thing. @pivotal-jamie-klassen help please.

kcmannem:
I think we can use these semantics/syntax to apply the webkit properties via elm-css. I just don't know where to put it.

```
legacyBorderRadius amount =
    Css.batch
        [ property "-moz-border-radius" amount
        , property "-webkit-border-top-left-radius" amount
        , property "-webkit-border-top-right-radius" amount
        , property "-webkit-border-bottom-right-radius" amount
        , property "-webkit-border-bottom-left-radius" amount
        , property "border-radius" amount
        ]

```

kcmannem:
So I wanted to add the style for every view but that meant making these style variable everywhere. I decided to just leave the stuff in globals.less, Lemme know if theres a globals.elm for me to move this to.

web/assets/css/globals.less
```diff
@@ -28,3 +28,18 @@ h3 {
   font-size: 14px;
 }
 
+::-webkit-scrollbar {
+  width: 10px;
+}
+
+::-webkit-scrollbar-track {
+  background: #3d3c3c00;
```
kcmannem: this is not a real color, its transparent notice the alpha value being 00. What's a better way to use transparency?

matthewpereira: I think `rgba(RRR, GGG, BBB, A);` is the best format for this.

In this case, if we want the opacity to be 0 on this element, I think you'd want this code:
`background: rgba(61, 60, 60, 0);`

(Technically you could replace the RGB values with whatever as long as the property is being set to fully transparent, I'm just translating literally from the hex color code above.)

kcmannem: ty

kcmannem:
@pivotal-jamie-klassen I don't think we can achieve the same look in firefox according to my googling and this answer https://stackoverflow.com/a/54101063, We would still have a white track.

 I've added release notes as well.

matthewpereira:
c2d8CX7bRq

^ This is what it looks like in Firefox 68 (on Mojave OSX, for what it's worth) when you set 
```css
body {
...
    scrollbar-color: #555 rgba(0,0,0,0);
}
```
This is experimental, draft-spec stuff (https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars) so we might want to separate it out for managing tech-debt reasons, if you want to add it in at all.

One line of CSS is pretty harmless in the long run, but I know you folks run a tight ship! 🚢

release-notes/v5.5.0.md
```diff
@@ -111,3 +111,7 @@
 #### <sub><sup><a name="v550-note-28" href="#v550-note-28">:link:</a></sup></sub> fix
 
 * @xavierzwirtz also fixed `hg.pull` by passing the `sourceUri` in the Mercurial resource. This could fix some issues where Mercurial wasn't saving the full uri in the `hgrc` file concourse/hg-resource#13.
+
+#### <sub><sup><a name="v550-note-29" href="#v550-note-29">:link:</a></sup></sub> fix
+
+* Scrollbars have been given a touch of the Concourse design language. It should no longer use browser defaults.
```
pivotal-jamie-klassen: Cute! Can you mention this PR number (#4355) in the note? I wonder if this will be a recurring problem

matthewpereira: Might be worth calling out `Scrollbars in Chrome` (and maybe FF if you add the above note's consideration to the PR) so that if there are Safari or Opera or whatever users out there they don't get confused.

pivotal-jamie-klassen:
> @pivotal-jamie-klassen I don't think we can achieve the same look in firefox according to my googling and this answer https://stackoverflow.com/a/54101063, We would still have a white track.
> 
> I've added release notes as well.

@kcmannem Well I was gonna say don't worry about Firefox, but looks like we got @matthewpereira the CSS wizard in here:

> ```css
> body {
> ...
>     scrollbar-color: #555 rgba(0,0,0,0);
> }
> ```
> 
> This is experimental, draft-spec stuff (https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Scrollbars) so we might want to separate it out for managing tech-debt reasons, if you want to add it in at all.
> 
> One line of CSS is pretty harmless in the long run, but I know you folks run a tight ship! 🚢

release-notes/v5.5.0.md
kcmannem: Maybe we can build a cli tool for this we do to autogen migrations

kcmannem:
noob question. Do i just add those properties to the body style in globals.less or do I need to add magic syntax to make sure it only gets applied for firefox and not chrome (I assume it'll just ignore and throw an error in the console?).

kcmannem:
ignore my last comment, i verified myself. Can confirm it works on both chrome and firefox now.

release-notes/v5.5.0.md
pivotal-jamie-klassen: 🤔 i like the sounds of that

pivotal-jamie-klassen: but @matthewpereira's point is a good one. I'm not sure that Opera or Safari are going to support the same bleeding-edge features that Firefox does.

PR #4357, opened 2019-09-04
fix go.mod to properly comply to 1.13 strict rules

cirocosta:
With the bump to go 1.13, there's a more strict validation against
the tags used in the [pseudo versions], meaning that it ends up really
reaching out to the source provider to validate if the sha matches the
tag (which, in this case, doesn't):

	git show v2.0.0-alpha.1
	commit d81cb85237595f720d83eda492bae8f6162fc5c0 (tag: v2.0.0-alpha.1)

By removing the tag, we can then let go perform the validation against
just the SHA.

[pseudo versions]: https://golang.org/cmd/go/#hdr-Pseudo_versions

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -219,7 +219,9 @@ require (
 	gotest.tools v2.1.0+incompatible // indirect
 	k8s.io/api v0.0.0-20171027084545-218912509d74
 	k8s.io/apimachinery v0.0.0-20171027084411-18a564baac72
-	k8s.io/client-go v2.0.0-alpha.0.0.20171101191150-72e1c2a1ef30+incompatible
+	k8s.io/client-go v0.0.0-20171101191150-72e1c2a1ef30
 	k8s.io/kube-openapi v0.0.0-20180731170545-e3762e86a74c // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -219,7 +219,9 @@ require (
 	gotest.tools v2.1.0+incompatible // indirect
 	k8s.io/api v0.0.0-20171027084545-218912509d74
 	k8s.io/apimachinery v0.0.0-20171027084411-18a564baac72
-	k8s.io/client-go v2.0.0-alpha.0.0.20171101191150-72e1c2a1ef30+incompatible
+	k8s.io/client-go v0.0.0-20171101191150-72e1c2a1ef30
 	k8s.io/kube-openapi v0.0.0-20180731170545-e3762e86a74c // indirect
 	layeh.com/radius v0.0.0-20190101232339-d3a4fc175dc9 // indirect
 )
+
+go 1.13
```
cirocosta: I'm not entirely sure if this is good to keep or not (would this block people from using non go1.13? if so, would that be bad?)

pivotal-jamie-klassen: let's keep the inexorable forward march of progress going. I don't think anybody needs to _develop_ concourse on pre-1.13?

PR #4359, opened 2019-09-04
Revert "use position fixed so tooltips can 'float' outside the container"

### LINKED ISSUE #4210: zany-looking build page tooltip
## Bug Report

css-transform seems to do something weird when the transform would put the element outside of its parents overflow area? Maybe something to do with the other style attributes on the tooltips.

## Steps to Reproduce

visit a build page with decently long output, like https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/dev-image/builds/948, and scroll down a bit, then hover over one of the step status icons to see a duration tooltip.

## Expected Results

a nicely behaved tooltip that is connected to its arrow.

## Actual Results

this scoundrel:
Screen Shot 2019-08-01 at 7 56 16 PM

## Additional context

I merged #4206 too hastily 😢 

## Version Info

* Concourse version: 5.4.0-rc (master)
* Deployment type (BOSH/Docker/binary): bosh
* Infrastructure/IaaS: gcp (prod)
* Browser (if applicable): chrome
* Did this used to work? yes, except for the ones at the top of the output

pivotal-jamie-klassen:
Reverts concourse/concourse#4206
Fixes #4210

### DIFF
--- a/web/elm/src/Build/Styles.elm
+++ b/web/elm/src/Build/Styles.elm
@@ -212,8 +212,9 @@ stepStatusIcon =
 
 firstOccurrenceTooltip : List (Html.Attribute msg)
 firstOccurrenceTooltip =
-    [ style "position" "fixed"
-    , style "transform" "translate(0, -100%)"
+    [ style "position" "absolute"
+    , style "left" "0"
+    , style "bottom" "100%"
     , style "background-color" Colors.tooltipBackground
     , style "padding" "5px"
     , style "z-index" "100"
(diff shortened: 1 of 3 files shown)


web/elm/src/Build/Styles.elm
```diff
@@ -212,8 +212,9 @@ stepStatusIcon =
 
 firstOccurrenceTooltip : List (Html.Attribute msg)
 firstOccurrenceTooltip =
-    [ style "position" "fixed"
-    , style "transform" "translate(0, -100%)"
+    [ style "position" "absolute"
```
kcmannem: absolute is always the correct position

Issue #4236, opened 2019-08-07
Enforce that all version fields are strings


vito:
Versions can be specified in a few places:

1. [`image_resource.version`](https://concourse-ci.org/tasks.html#image-resource-version)
1. [`version`](https://concourse-ci.org/resources.html#resource-version) on a resource definition
1. [`version`](https://concourse-ci.org/get-step.html#get-step-version) on a `get` step

We should make sure that all of these locations only accept a string value for each field.

Right now we're inconsistent. Cases 1 and 2 unmarshal into an `atc.Version` which is an alias to `map[string]string`, so those should be OK.

Case 3 however does some wacky meta-unmarshal stuff, since that field can technically be 3 things: "latest", "every", or a version object. It looks like our current implementation simply skips any non-string values, which is...well, that's *one* way to do it:

https://github.com/concourse/concourse/blob/12b4d7cfc178e0f0639e4b5b899b4d2f3784b012/atc/config.go#L126-L134

(Also, what's that `strings.TrimSpace` doing there??)

We should have all three bail out on any non-string values. We shouldn't do any coercion; that is the path to darkness (see https://github.com/concourse/concourse/issues/1407).

clarafu:
Closing this with the merge of #4371

PR #4371, opened 2019-09-04
Enforced all version fields to be strings

julia-pu:
This pr fixes the bug #4236. An error is now returned when there is a non-string value in the version config.

### DIFF
--- a/atc/config.go
+++ b/atc/config.go
@@ -128,7 +128,10 @@ func (c *VersionConfig) UnmarshalJSON(version []byte) error {
 		for k, v := range actual {
 			if s, ok := v.(string); ok {
 				version[k] = strings.TrimSpace(s)
+				continue
 			}
+
+			return fmt.Errorf("the value %v of %s is not a string", v, k)
 		}
 
 		c.Pinned = version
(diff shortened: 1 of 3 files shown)


atc/config.go
```diff
@@ -128,7 +128,10 @@ func (c *VersionConfig) UnmarshalJSON(version []byte) error {
 		for k, v := range actual {
 			if s, ok := v.(string); ok {
 				version[k] = strings.TrimSpace(s)
```
clarafu: I think Alex mentioned in the original issue for this PR that `TrimSpace(s)` was not needed?

xtremerui: It is somehow still useful ... it will trim these
`{"some":"        value         "}` to `{"some":"value"}`

It depends on how we want to process the value string , it we say they get exactly what they put in then yeah we should remove `TrimSpace`

clarafu: @vito pls more insight

vito: It just seems like a weasly bit of code that never had strong justification for being there in the first place. I don't see any reason why we would want it here and not literally everywhere else in the config, lol. For example someone could say `passed: ["job-a "]` but we don't bother sanitizing that.

PR #4377, opened 2019-09-05
remove reference to db.containerMetadat from resource package

kcmannem:
I really don't know why we had this resource.Session wrapper. It did nothing so I got rid of it. It removes references to `db.conatinerMetadata` from the resource package as a benefit as well.


ddadlani:
looks like the tests are failing?

kcmannem:
fixed

Issue #4348, opened 2019-09-03
FLY set-team documentation


niall-byrne:
## Bug Report

This is perhaps a nit, but was a bit of a pain:

Fly Documentation reports the following when running 'fly st --help`

```
Authentication (GitHub):
          --github-user=USERNAME              List of whitelisted GitHub users
          --github-org=ORG_NAME               List of whitelisted GitHub orgs
          --github-team=ORG_NAME:TEAM_NAME    List of whitelisted GitHub teams
```

`--github-team` does not actually appear to take a list as an argument.
Entering a comma separated list resulted in the entire string being put into the database as a "team name".

Found different language here: https://concourse-ci.org/github-auth.html
(Of course I looked here after the fact 😢  )

Suggest updating the CLI help text.

## Steps to Reproduce

```
fly -t prod st -n myteam --github-team=Pivotal:team-1,Pivotal:team-2,Pitvotal:team-3
```

## Expected Results

The list to be parsed and understood correctly by the CLI as three separate github teams.

## Actual Results

A single team was inserted into the db, the team name was the complete 'list' commas and all.  Github Team: `Pivotal:team-1,Pivotal:team-2,Pitvotal:team-3`

## Version Info

*Fly Version: 5.3.0

deniseyu:
hey @niall-byrne - this might be an issue with how the library we use to parse flags handles lists of arguments! to get multiple list items passed in, I think you have to specify the same flag a few times:

```
fly -t prod st -n myteam --github-team Pivotal:team-1 --github-team Pivotal:team-2
```

An example is in the integration test suite, with github-users rather than teams but the inputs are handled the same way: https://github.com/concourse/concourse/blob/master/fly/integration/set_team_test.go#L614-L622

It looks like the flags library does support delimiters, but only for arguments passed in through the environment: https://godoc.org/github.com/jessevdk/go-flags#Option and I'm not super keen to switch to env-based flags because an exported environment variable sitting in your shell session can cause surprising behavior.

I agree that the wording and example in the docs are confusing though, esp for folks who haven't used this particular flags library before. I'm not sure that it makes sense right now to add comma-parsing since it feels like functionality that should be supported by the flag-parsing library rather than doing string-munging in Fly. 

If you're up for opening a PR to fix the wording of the docs we'd welcome that!

niall-byrne:
@deniseyu Thanks, yep PR to fix documentation sounds like the right approach.  OK i'll see what I can do.

[...]

PR #4392, opened 2019-09-09
Documentation: https://github.com/concourse/concourse/issues/4348

niall-byrne:
In response to: https://github.com/concourse/concourse/issues/4348

The current wording seems to suggest a list is a valid input to any of these flags, where a single value is instead what is required.

I believe this brings the fly documentation more in line with: https://concourse-ci.org/github-auth.html


clarafu:
Actually there is a unit test that needs to be updated with the new messages https://github.com/concourse/concourse/blob/master/fly/integration/set_team_test.go#L436-L465. You can see the failure here https://ci.concourse-ci.org/builds/6906118.

niall-byrne:
Leave it with me, I'll circle back.

niall-byrne:
```
Ginkgo ran 12 suites in 3m0.181779165s
Test Suite Passed
```

I believe I have the correct set of tests passing this time, sorry about that.

PR #4395, opened 2019-09-09
upsert cert caches

### LINKED ISSUE #4326: acme/autocert managed certificate not updating
## Bug Report

We upgraded to concourse v5.3.0 back in June and started using the let's encrypt functionality. This has been running fine and we were happy.  However, The original certificates is expiring in a couple weeks now. We expected our certificate to update around 30 days before expiration, but it has not.

Looking at certificate transparency logs for our particular domain at https://crt.sh, we see the domain actually renewed around 30 days before expiration - and apparently again a week after that.  However our concourse is still serving the old certificate that expires in two weeks.

## Steps to Reproduce

Enable "Let's Encrypt" functionality and see that the certificate does not get updated.

## Expected Results

We expected the certificate to automatically update. 

## Additional Context

Looking into this we started seeing errors in the concourse db logs around 30 days prior to expiration:

```
ERROR:  duplicate key value violates unique constraint "cert_cache_pkey"
DETAIL:  Key (domain)=(OUR-FANCY-DOMAIN) already exists.
STATEMENT:  INSERT INTO cert_cache (domain, cert, nonce) VALUES ($1, $2, $3)
```

One hypothesis we have is that the new certificate just isn't getting stored in the cache so not refreshing.   In particular we see that the [query to store state in the database](https://github.com/concourse/concourse/blob/7c63f45bc1dd146b790e17fdad008e76d3487356/atc/atccmd/cert_cache.go#L25-L28) `cert_cache` table is an `INSERT` but perhaps it needs to be an [UPSERT](https://www.postgresql.org/docs/9.5/sql-insert.html#SQL-ON-CONFLICT) instead.

## Version Info

* Concourse version: 5.3.0
* Infrastructure/IaaS: GCP

pivotal-jwinters:
fixes #4326


release-notes/latest.md
```diff
@@ -0,0 +1,6 @@
+#### <sub><sup><a name="v551-note-1" href="#v551-note-1">:link:</a></sup></sub> fix
+
+* If you are using LetsEnrypt, you will want to upgrade to this version because it includes a fix for a bug where updating existing autocert/acme certificates fails
```
pivotal-jamie-klassen: let's mention the bug and this PR number for documentation best practices. Also typo.
```suggestion
* If you are using LetsEncrypt, you will want to upgrade to this version because it includes a fix for a [bug](https://github.com/concourse/concourse/issues/4326) where updating existing autocert/acme certificates fails #4395.
```

release-notes/latest.md
```diff
@@ -0,0 +1,6 @@
...
+
+tldr: upsert autocert certs.
```
pivotal-jamie-klassen: I don't think this is necessary

release-notes/latest.md
```diff
@@ -0,0 +1,6 @@
+#### <sub><sup><a name="v551-note-1" href="#v551-note-1">:link:</a></sup></sub> fix
```
pivotal-jamie-klassen: This is fine, but so you're aware these IDs can be anything unique, so the PR number or issue number is a natural choice, i.e.
```suggestion
#### <sub><sup><a name="v551-note-4326" href="#v551-note-4326">:link:</a></sup></sub> fix
```

PR #4398, opened 2019-09-10
feat: strengthen secret redaction to handle split secrets.

### LINKED ISSUE #4389: Strengthen secret redaction to handle split secrets
@evanchaoli got the ball rolling in https://github.com/concourse/concourse/pull/4311 by implementing a first pass of automatic credential redacting, however along the way we found one tricky problem.

There's a chance that Concourse receives "pieces" of a credential in separate `Write()` calls, at which point no redacting can occur as the full credential is never seen at once. This could happen if a sufficiently large credential is printed - say, a multiline SSH key. It might also just happen randomly; process output is received in arbitrary chunks, so it could be a timing problem too.

Relevant discussion: https://github.com/concourse/concourse/pull/4311#discussion_r319647503

evanchaoli:
Resolve #4389

Highlights:

1. Given build logs are in random chunk, first thing to do is normalize build logs to lines. If the last line doesn't contains new-line, then cache the last line.
2. Add a `Close()` method to `dbEventWriter`, because as we normalize log lines, if don't close a writer, then the last line might be lost.
3. After normalize log lines, we can treat a multi-line secret as multiple secrets, meaning redact each line of a secret separately. Consider that, users may print a ssh to log with each line having several leading spaces, in that case, we have no way to replace the kay as a whole.
4. I also did some changes that directs image pulling logs are redirected to stderr, and `get`/`put`'s user logs are redirected to stdout. So that we only do redaction on stdout.

The following screen shot shows my test result:
redact4

In this test, I managed to generate a 20K-line build log with some secrets embedded including multiple line secrets. As we can see, all secrets are redacted, and multiple secrets are redacted at line level.

### DIFF
--- a/atc/engine/builder/delegate_factory.go
+++ b/atc/engine/builder/delegate_factory.go
@@ -88,6 +88,10 @@ func (d *getDelegate) Starting(logger lager.Logger) {
 }
 
 func (d *getDelegate) Finished(logger lager.Logger, exitStatus exec.ExitStatus, info exec.VersionInfo) {
+	// PR#4398: close to flush stdout and stderr
+	d.Stdout().(io.Closer).Close()
+	d.Stderr().(io.Closer).Close()
+
 	err := d.build.SaveEvent(event.FinishGet{
 		Origin:          d.eventOrigin,
 		Time:            d.clock.Now().Unix(),
@@ -186,6 +190,10 @@ func (d *putDelegate) Starting(logger lager.Logger) {
 }
 
 func (d *putDelegate) Finished(logger lager.Logger, exitStatus exec.ExitStatus, info exec.VersionInfo) {
+	// PR#4398: close to flush stdout and stderr
+	d.Stdout().(io.Closer).Close()
+	d.Stderr().(io.Closer).Close()
+
 	err := d.build.SaveEvent(event.FinishPut{
 		Origin:          d.eventOrigin,
 		Time:            d.clock.Now().Unix(),
@@ -269,6 +277,10 @@ func (d *taskDelegate) Starting(logger lager.Logger, taskConfig atc.TaskConfig) 
 }
 
 func (d *taskDelegate) Finished(logger lager.Logger, exitStatus exec.ExitStatus) {
+	// PR#4398: close to flush stdout and stderr
+	d.Stdout().(io.Closer).Close()
+	d.Stderr().(io.Closer).Close()
+
 	err := d.build.SaveEvent(event.FinishTask{
 		ExitStatus: int(exitStatus),
 		Time:       time.Now().Unix(),
@@ -320,6 +332,8 @@ func NewBuildStepDelegate(
 		planID:          planID,
 		clock:           clock,
 		credVarsTracker: credVarsTracker,
+		stdout:          nil,
+		stderr:          nil,
 	}
 }
 
@@ -328,6 +342,8 @@ type buildStepDelegate struct {
 	planID          atc.PlanID
 	clock           clock.Clock
 	credVarsTracker vars.CredVarsTracker
+	stderr          io.Writer
+	stdout          io.Writer
 }
 
 func (delegate *buildStepDelegate) Variables() vars.CredVarsTracker {
@@ -342,8 +358,10 @@ type credVarsIterator struct {
 	line string
 }
 
-func (it *credVarsIterator) YieldCred(k, v string) {
-	it.line = strings.Replace(it.line, v, "[**redacted**]", -1)
+func (it *credVarsIterator) YieldCred(name, value string) {
+	for _, lineValue := range strings.Split(value, "\n") {
+		it.line = strings.Replace(it.line, lineValue, "[**redacted**]", -1)
+	}
 }
 
 func (delegate *buildStepDelegate) buildOutputFilter(str string) string {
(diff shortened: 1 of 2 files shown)


evanchaoli:
@vito wrt stdout/stderr, my idea is to make put/get/task consistent. There are two types of step ouput:
Type1: concourse generated output, like "pulling =======", or some underlying error printed by concourse code;
Type2: user script generated output, resource type code's stderr and task script's both stdout&stderr

I feel it would be helpful if we set a principle that type1 going to log-event(stderr) and type2 going to log-event(stdout). This way, though we mix stdout and stderr from task, I don't think it is a problem because users only build logs from GUI. Comparing to mixing concourse output with task's stderr, I believe mixing task's stdout&stderr is better.

The principle will also benefit future data analysis, because log-event(stdout) is system log and log-event(stderr) is user log.

Back to this PR, redacting secrets is CPU costly. A single line of "pulling ======== xx MB" showing on GUI is actually a lot of log writes, but they should not go through redaction process at all. Only user log holds risk of leaking secrets. With distinct system log and user log, we only do redaction against user log, so that it saves a lot of CPU resource.

vito:
@evanchaoli :thinking: That seems like a separate problem from the goals of this PR. It sounds like you want a way to separate the image fetching output from the task's output. I agree with that but I don't think it should be done by messing with and changing the meaning of `stdout`/`stderr`.

Right now resources must use `stdout` for writing the JSON response, which means resource type authors have to be careful not to accidentally print anything else to `stdout`. But the new resource protocol (whether it's concourse/rfcs#24 or concourse/rfcs#37) introduces a response *file*, freeing up resources to use both `stdout` and `stderr` for output, just like tasks.

At this point resource type authors (prototype authors?) will be able to print logs like normal, and it would be confusing to find upon inspection that your logs are somehow getting changed to `stderr`, and only when used as an `image_resource`. The impact of this is probably low but it feels like a subtle complication in a system with a lot of moving parts that I would rather avoid for the sake of contributor's ability to comprehend how everything works.

Similarly, it would be confusing to me if my task which prints to `stderr` has all of its output redirected to `stdout`. (In fact, I'm really not sure why that seems to be happening with `bash -x`, and I'd like to fix that.)

IMO we should just filter both streams regardless. Intuitively, both are just as likely to contain credentials: `echo $SECRET` with `set -x` will print it on `stdout` and `stderr`. And it's just as likely for tasks to print a whole bunch of output to either `stdout` or `stderr` - we get reports all the time of the build UI being slow to render such tasks - so we'll need to make it as efficient as possible, just as we would with the image fetching output.

I totally agree that it would be nice to be able to distinguish image fetching output from task output. I'd like to see it render differently in the UI, even. But I think that's a topic for a separate PR/issue, and I think it should be done in some other way so that `stdout`/`stderr` don't lose their original meaning. :slightly_smiling_face:

evanchaoli:
@vito Time difference really slows down our communication [Facepalm].

I agree one PR should focus on a single issue, which is a rule we are following in our daily development. But here for this PR, I want to distinguish image fetching output from task (actually including put/get steps) output is to make the redaction feature more efficient.

Now, there are already some confusion to resource type author and pipeline author:
* Image fetching output is mixed with step stderr output
* Task's stdout and stderr are mixed to stdout (which you want to fix, but IMO you don't have to for now because fixed this confusion just leads to the first confusion)

If we clearly state in documentation that like "output step's output going to log-event(stdiout)", I guess it should be ok to resource type author and pipeline author, as they anyway has no way to view stdout and stderr differently.

If we want to distinguish sys output (image fetch etc), and step's stdout/stderr, I think only way is to introduce a third log-event source, say "sys", other than "stdout" and "stderr". This can be together with the "protocol" feature.

But I guess it will take some time to get the "protocol" feature released. How about this way that let's open an issue on me, once the "protocol" feature is merged (before released), get/put get able to print to stdout, sys log, step's stdout and stderr are all distinguished, I can make a PR to add redaction back to stderr. So that now we can benefit improvement of redaction efficiency, and we won't forget to add redaction to stderr in future. I don't see I would move off my current project in short term, so I can promise to follow up the issue.

vito:
@evanchaoli Yeah the timezone difference is a bit rough. I'll try to reply earlier in the mornings so there's a little more overlap (though it's hard to make a 12hr difference work either way lol). Thanks for being patient.

Let me try to summarize my thoughts on this:

* This is overloading the terms `stdout`/`stderr` to mean something completely different. At that point we might as well not call them `stdout`/`stderr` at all.
* I don't agree that `stderr` should have different behavior, whether it's being used for 'system' logs or not. A task's image fetcher could accidentally leak credentials too. The `registry-image` resource type probably won't, but that's not the only resource type that can be used there. Since we're redacting `get` step output, we should redact `image_resource` output, too; they have the same semantics.
* Avoiding `stderr` for efficiency's sake feels like a short-term win considering:
  * many people have very verbose tasks and we'll have to make sure the redacting is efficient either way
  * some users (like us) actually fetch images as an explicit `get` step and then use [`image:`](https://concourse-ci.org/task-step.html#task-step-image) on a task step to use it, so we'll end up redacting the image fetching output anyway, but in the `get` step - not an `image_resource`
  * image resource fetching is cached, so eventually there won't be any output to redact anyway
  * we're in the process of switching to `containerd` and natively using OCI images, so the verbose output that you're seeing will go away as the resource will no longer have to unpack the rootfs.

I'm still in support of being able to distinguish image fetching logs from task output, which could be done by adding another field to the event origin to differentiate its stdio from the step's stdio:

https://github.com/concourse/concourse/blob/7e41cf6fbe4711bb71aa3760c9963b493eca456f/atc/event/events.go#L100-L103

...but this should probably be a separate PR since it'll involve more components (i.e. the web UI).

Does this make sense? :sweat_smile:

evanchaoli:
@vito OK, this is a good discussion. Let's move things forward. I can easily remove the stdout/stderr related changes. Just to confirm with you, if I update the PR by your Monday morning, is it possible to include this PR in 5.6? I wish to have a full functional redaction feature in 5.6.

evanchaoli:
@vito I have deleted stdout/stderr related changes, and now redacting on both stdout and stderr. Tests are updated accordingly. Please take a look again. Thanks a lot.

evanchaoli:
@vito Thank you so much for including this in 5.6.

PR #4403, opened 2019-09-10
Don't store secrets in the check plan

clarafu:
A topgun test was failing for lidar changes because we saved the check plan into the DB with variables interpolated and the topgun test checks to make sure there are no secrets stored in the DB. We changed it so that we will interpolate the vars in the check step instead.

### DIFF
--- a/atc/engine/builder/builder.go
+++ b/atc/engine/builder/builder.go
@@ -31,7 +31,7 @@ type DelegateFactory interface {
 	GetDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.GetDelegate
 	PutDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.PutDelegate
 	TaskDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.TaskDelegate
-	CheckDelegate(db.Check, atc.PlanID) exec.CheckDelegate
+	CheckDelegate(db.Check, atc.PlanID, vars.CredVarsTracker) exec.CheckDelegate
 	BuildStepDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.BuildStepDelegate
 }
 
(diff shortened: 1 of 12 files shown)


xtremerui:
Verified by acceptance test:

1. Setup local Vault with credentials and run docker compose concourse against it.
2. Config a pipeline with `uri: ((uri))` as source of a git resource
3. Verify by manually checking the resource, the plan column in `Checks` table should be populated with config without variable interpolation like below
```
{"id":"","check":{"type":"git","name":"booklit","source":{"uri":"((uri))"},"timeout":"1h0m0s","from_version":{"ref":"6198330aaa75d5be40aeb822a30277ffe9311838"}}}
```

Without this PR, the check plan will store an evaluated config like
```
{"id":"","check":{"type":"git","name":"booklit","source":{"uri":"https://github.com/vito/booklit"},"timeout":"1h0m0s","from_version":{"ref":"6198330aaa75d5be40aeb822a30277ffe9311838"}}}
```

atc/engine/builder/builder.go
```diff
@@ -31,7 +31,7 @@ type DelegateFactory interface {
 	GetDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.GetDelegate
 	PutDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.PutDelegate
 	TaskDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.TaskDelegate
-	CheckDelegate(db.Check, atc.PlanID) exec.CheckDelegate
+	CheckDelegate(db.Check, atc.PlanID, vars.CredVarsTracker) exec.CheckDelegate
```
vito: :+1: lol cool that this inconsistency ended up being fixed so quickly

pivotal-jamie-klassen:
@clarafu @pivotal-jwinters can we get a release note for this please?

pivotal-jamie-klassen:
This was a regression introduced by #4202, which is documented by #4475

PR #4404, opened 2019-09-10
Add a timeout when acquiring mutex lock

### LINKED ISSUE #4393: mutex lock timeout
In the `lockFactory`, when acquiring a lock, it can get stuck talking to the database. Future calls will block on the mutex.Lock. This can lead to an avalanche of goroutines from the `buildtracker`, `scheduler`, and `lidar` as each tick piles on more work.

We want to add a timeout to the mutex.Lock so that new goroutines will die quicker when they can't fetch the lock.

clarafu:
fixes #4393 

In the lockFactory, when acquiring a lock, it can get stuck talking to the database. Future calls will block on the mutex.Lock. This can lead to an avalanche of goroutines from the buildtracker, scheduler, and lidar as each tick piles on more work.

We want to add a timeout to the mutex.Lock so that new goroutines will die quicker when they can't fetch the lock.

### DIFF
--- a/atc/db/lock/lock.go
+++ b/atc/db/lock/lock.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"code.cloudfoundry.org/lager"
 )
@@ -65,9 +66,9 @@ type LockFactory interface {
 }
 
 type lockFactory struct {
-	db           LockDB
-	locks        lockRepo
-	acquireMutex *sync.Mutex
+	db    LockDB
+	locks lockRepo
+	mutex *mutex
 
 	acquireFunc LogFunc
 	releaseFunc LogFunc
@@ -91,7 +92,7 @@ func NewLockFactory(
 			locks: map[string]bool{},
 			mutex: &sync.Mutex{},
 		},
-		acquireMutex: &sync.Mutex{},
+		mutex: &mutex{make(chan struct{}, 1)},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


atc/db/lock/lock.go
```diff
@@ -147,42 +148,44 @@ type LockDB interface {
...
+	mutex  *mutex
 
 	acquired LogFunc
 	released LogFunc
 }
 
 func (l *lock) Acquire() (bool, error) {
-	l.acquireMutex.Lock()
-	defer l.acquireMutex.Unlock()
+	if l.mutex.Lock(time.Second) {
```
cirocosta: Hey, little nit - wdyt of having the `time.Second` here as a constant, for the purpose of being explicit that that is the timeout we're using? 

thanks!

atc/db/lock/lock_test.go
```diff
@@ -104,6 +104,36 @@ var _ = Describe("Locks", func() {
...
+
+			done := make(chan struct{})
+			readyToAcquire := make(chan struct{})
+
+			fakeLockDB := new(lockfakes.FakeLockDB)
+			factory := lock.NewTestLockFactory(fakeLockDB)
+
+			fakeLockDB.AcquireStub = func(id lock.LockID) (bool, error) {
+				select {
+				case <-time.After(time.Minute):
```
xtremerui: If default timeout is 1 second, should we make this test to a bit closer to that so if someone extend the default timeout this test could protect that? How about `time.Minute * 2`

atc/db/lock/lock.go
pivotal-jwinters: Fixed

atc/db/lock/lock_test.go
pivotal-jwinters: Updated

pivotal-jwinters:
@cirocosta @xtremerui do either of you have any general thoughts on this change? Any reservations about why doing this might be a bad idea?

PR #4406, opened 2019-09-10
[Fly] Single team scoped fly cmd

### LINKED ISSUE #4196: [fly] Add `--team-name` for all Fly commands

zoetian:
Resolves part 1/2 to fixes #4196

As mentioned in the above issue, this PR is to address the single-team-scoped fly commands (i.e. scoped by pipeline/ resource/ build etc.)

1. - [x]  hijack
2. - [x]  trigger-job
3. - [x]  pause-job
4. - [x]  unpause-job
5. - [x]  jobs

**This PR is ready for review**

We're going to continue implementing this type of change for the remaining commands (see #4196, the single-team commands section) in other PR's. For now, this can be reviewed/merged in.


cirocosta:
Hey @zoetian , 

As with just these four commands, we're already crossing the 1500+ lines of code of change, wdyt of splitting the next commands on other PRs? I imagine that doing so would make it simpler for reviewers to reason about the more isolated changes 🤔

Thanks!

cirocosta:
[ipm]: there might be some opportunity here for refactoring the `loginATCServer` so that we can DRY.

cirocosta:
[ipm]: let's wrap this PR around just these commands that we added, and then take some time on thinking on how we could gain more velocity with some refactoring to the testing

taylorsilva:
@concourse/core-api This PR is ready for review please and thanks 😄

fly/commands/fly.go
```diff
@@ -82,3 +85,14 @@ type FlyCommand struct {
 }
 
 var Fly FlyCommand
+
+type TeamFlag struct {
+	Team string `short:"n" long:"team-name" description:"Run command for specific team"`
```
vito: Same feedback as https://github.com/concourse/concourse/pull/4324#discussion_r333701608

fly/commands/fly.go
```diff
@@ -82,3 +85,14 @@ type FlyCommand struct {
...
+}
+
+func (t *TeamFlag) GetTeamTarget(target rc.Target) concourse.Team {
```
vito: nit: avoid naming methods `GetX`

fly/commands/hijack.go
```diff
@@ -63,10 +64,12 @@ func (command *HijackCommand) Execute([]string) error {
 	if err != nil {
 		return err
 	}
+
 	var chosenContainer atc.Container
 
 	if command.Handle != "" {
-		chosenContainer, err = target.Team().GetContainer(command.Handle)
+		team := command.TeamFlag.GetTeamTarget(target)
```
vito: tiny idea: since `TeamFlag` is embedded, this could just be `command.GetTeamTarget(target)`

fly/commands/trigger_job.go
```diff
@@ -30,11 +32,17 @@ func (command *TriggerJobCommand) Execute(args []string) error {
 		return err
 	}
 
-	build, err := target.Team().CreateJobBuild(pipelineName, jobName)
+	var build atc.Build
+	team := command.TeamFlag.GetTeamTarget(target)
+	build, err = team.CreateJobBuild(pipelineName, jobName)
 	if err != nil {
+		if command.Team == "" {
+			fmt.Println("hint: are you missing '--team' to specify the team for the build?")
```
vito: this seems like it could be confusing to print on any error; the others ain't so bad since it's an explicit 404

fly/commands/pause_job.go
```diff
@@ -22,16 +24,20 @@ func (command *PauseJobCommand) Execute(args []string) error {
...
 	if err != nil {
 		return err
 	}
 
 	if !found {
-		return fmt.Errorf("%s/%s not found\n", command.Job.PipelineName, command.Job.JobName)
+		if command.Team == "" {
+			fmt.Println("hint: are you missing '--team' to specify the team for the build?")
+		}
+		return fmt.Errorf("%s/%s not found\n", pipelineName, jobName)
```
vito: how about just amending this to say 'not found in team X'? rather than guessing with a hint maybe they'll notice the problem in the message

fly/commands/unpause_job.go
```diff
@@ -22,16 +24,20 @@ func (command *UnpauseJobCommand) Execute(args []string) error {
...
+	team := command.TeamFlag.GetTeamTarget(target)
+	found, err := team.UnpauseJob(pipelineName, jobName)
 	if err != nil {
 		return err
 	}
 
 	if !found {
-		return fmt.Errorf("%s/%s not found\n", command.Job.PipelineName, command.Job.JobName)
+		if command.Team == "" {
+			fmt.Println("hint: are you missing '--team' to specify the team for the build?")
```
vito: (same feedback as before; maybe just put the team in the error message)

zoetian: umm, after adding the team into the 404 message, I found it really confusing to see it. not sure if they can take it as a hint when debugging tho. 😛

taylorsilva: I think it makes sense to combine the error messages like @vito said. Right now, this error message would always be printed, even if the user is only part of one team. Combining the two messages makes sense and gives the user all the info they'd need to debug why their command failed.

Currently we'd see this:
```
% fly -t local unpause-job -j some-pipeline/some-job
hint: are you missing '--team' to specify the team for the build?
error: some-pipeline/some-job not found
```

If we combine it:
```
% fly -t local unpause-job -j some-pipeline/some-job
error: some-pipeline/some-job not found on team some-team
```

zoetian: Thanks! I already updated the code. Are you still seeing that output?

zoetian:
@concourse/core-api Sorry about the super long delay, this PR is ready for another round of review now. Thanks!

xtremerui:
@taylorsilva

pivotal-jamie-klassen:
We were playing around with this patch and noticed an unexpected behaviour:
```bash
$ fly -t dev pj -j doesnt/exist
error: doesnt/exist not found on team &{main %!s(*internal.connection=&{http://localhost:8080 0xc000273260 false 0xc0002732c0})}
```
Probably we'd rather see something more like
```
fly -t dev pj -j doesnt/exist
error: doesnt/exist not found on team main
```

PR #4417, opened 2019-09-11
fly pin resource version with version JSON

### LINKED ISSUE #2702: pinning/unpinning a resource version from fly

xtremerui:
Fixes #2702 

Example:
cmd
```
fly -t ci pin-resource-version -r pipeline/resource -v '{"tag":"1.1.3"}'
```
Available resource versions:
```
v1 {"tag":"1.1.3", "meta":"some-meta","commit":"12345"}
v2 {"tag":"1.1.3", "meta":"some-other-meta","commit":"23456"}
v3 {"tag":"1.2.0", "meta":"foo-meta","commit":"76543"}
```

will find resource versions v1 and v2 that have field ["tag":"1.1.3"] and pin the latest one v2


gaelL:
Great, do you think you could implement also `SetPinCommentOnResource` ?https://github.com/concourse/concourse/blob/master/atc/routes.go#L49

cirocosta:
Hey @xtremerui , I wonder if the syntax should then be the same for `check-resource` and `check-resource-type` when supplying `--from` 🤔  

From my understanding, there we're in the same situation - we want to do some action on a resource(-type) passing a given version.

Wdyt?

xtremerui:
> Hey @xtremerui , I wonder if the syntax should then be the same for `check-resource` and `check-resource-type` when supplying `--from` 🤔

In one hand I like the idea of consistent, in the other hand I actually think in both case `--from` should be `--version`. For example of `check-resource`, the resource interface definition [here](https://concourse-ci.org/implementing-resource-types.html#resource-check) actually uses `version` in the check request. 

Then if we say `--from` is a nicer way of saying check one resource from a version, then maybe we should do that for pin resource too i.e. pinning a resource `to` || `by` a version. `from` in this case is a bit strange IMHO.

clarafu:
I think Ciro's suggestion makes a lot of sense, if we are keeping the same type of syntax of `from` with check-resource it reads well as something similar to "check this resource from this version" and "pin this resource at this version". Maybe we can have the `pin-resource` flag be `at` instead?

But also maybe it makes more sense to just have the command be `pin-resource-version` and you pass it a `--version` flag since we arn't really pinning a resource, but we are pinning a resource version (rather than `check-resource` where we are actually running a check on the resource). It is a little weird to give `pin-resource-version` a `version` flag but I think the future plan for fly is to pass it arguments instead of flags which would make much more sense in this case.

clarafu:
@xtremerui Should we add a testflight test for this?

xtremerui:
@gaelL could you be more specific on what do you want to add to `SetPinCommentOnResource `?  This endpoint doesn't require a version since it will just simply find one resource by its name and set comment.

gaelL:
Yes the issue  #4428 seems to be the one which describe it.

xtremerui:
@pivotal-jamie-klassen  it will now print out a JSON string of the version. The version ID is removed from the print out.

fly/commands/pin_resource.go
```diff
@@ -0,0 +1,77 @@
+package commands
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/concourse/concourse/fly/commands/internal/displayhelpers"
```
pivotal-jamie-klassen:
```suggestion
	"github.com/concourse/concourse/atc"
	"github.com/concourse/concourse/fly/commands/internal/displayhelpers"
```

fly/commands/pin_resource.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/concourse/concourse/fly/commands/internal/flaghelpers"
+	"github.com/concourse/concourse/fly/rc"
+	"github.com/concourse/concourse/go-concourse/concourse"
+)
+
+type PinResourceCommand struct {
+	Resource flaghelpers.ResourceFlag `short:"r" long:"resource" required:"true" value-name:"PIPELINE/RESOURCE" description:"Name of the resource"`
+	Version  flaghelpers.JsonFlag     `short:"v" long:"version" description:"JSON string of at least one field of the version. The given key value pair(s) has to be an exact match but not all fields need to be provided. In the case of multiple resource versions matched, it will pin the latest one."`
```
pivotal-jamie-klassen:
```suggestion
	Version  *atc.Version     `short:"v" long:"version" description:"JSON string of at least one field of the version. The given key value pair(s) has to be an exact match but not all fields need to be provided. In the case of multiple resource versions matched, it will pin the latest one."`
```

fly/commands/pin_resource.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	err = target.Validate()
+	if err != nil {
+		return err
+	}
+
+	team := target.Team()
+
+	if command.Version.Value != nil {
```
pivotal-jamie-klassen:
```suggestion
	if command.Version != nil {
```

fly/commands/pin_resource.go
```diff
@@ -0,0 +1,77 @@
...
+		versions, _, found, err := team.ResourceVersions(command.Resource.PipelineName, command.Resource.ResourceName, concourse.Page{}, command.Version.Value)
```
pivotal-jamie-klassen:
```suggestion
		versions, _, found, err := team.ResourceVersions(command.Resource.PipelineName, command.Resource.ResourceName, concourse.Page{}, *command.Version)
```

fly/commands/pin_resource.go
```diff
@@ -0,0 +1,77 @@
...
+
+		if err != nil {
+			return err
+		}
+
+		if !found || len(versions) <= 0 {
+			displayhelpers.Failf("could not find version matching %s", command.Version.Raw)
```
pivotal-jamie-klassen:
```suggestion
			displayhelpers.Failf("could not find version matching %V", command.Version)
```

pivotal-jamie-klassen:
@xtremerui i just merged but forgot to ask for release notes!

xtremerui:
@pivotal-jamie-klassen i will add it and make a PR again.

Issue #4409, opened 2019-09-11
SPIKE web: send token manually when browser prevents automatically sending token


pivotal-jamie-klassen:
Following #4314, let's distinguish the case when fly cannot be reached (like when you are running fly on a remote server) from the case when your browser itself won't allow automatically sending the token.

Concretely, let's only show the `copy token to clipboard` button in the 'fly unreachable' case and only show a direct link in the 'browser prevented request' case.

**NOTE** we need to investigate whether these different errors can be distinguished in the `elm/http` library. There's a good chance that both types of error are [`NetworkError`s](https://package.elm-lang.org/packages/elm/http/2.0.0/Http#Error).

Even if the `elm/http` library swallows the difference between these errors, we may be able to find a (or write our own) library that gives more direct access to `XMLHttpRequest` - clearly there is a difference between `Failed to load resource: Could not connect to the server` and `XMLHttpRequest cannot load <url> due to access control checks.`

PR #4423, opened 2019-09-12
web: distinguish browser error from network error

pivotal-jamie-klassen:
fixes concourse/concourse#4409

### DIFF
--- a/web/elm/src/Message/Subscription.elm
+++ b/web/elm/src/Message/Subscription.elm
@@ -1,6 +1,7 @@
 port module Message.Subscription exposing
     ( Delivery(..)
     , Interval(..)
+    , RawHttpResponse(..)
     , Subscription(..)
     , runSubscription
     )
@@ -32,6 +33,16 @@ port reportIsVisible : (( String, Bool ) -> msg) -> Sub msg
 port sideBarStateReceived : (Maybe String -> msg) -> Sub msg
 
 
+port rawHttpResponse : (Maybe Json.Encode.Value -> msg) -> Sub msg
+
+
+type RawHttpResponse
+    = Success -- ()
+    | BadUrl String -- ({"badUrl": "the-bad-url./com"})
+    | Timeout -- ({"timeout": true})
+    | NetworkError String -- ({"networkError": "cannot load ... due to access checks"})
+
+
 type Subscription
     = OnClockTick Interval
     | OnMouse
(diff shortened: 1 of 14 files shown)


pivotal-jamie-klassen:
I think we can switch the JSON decoding with just using strings, and remove the BadUrl variant

web/elm/src/Message/Subscription.elm
```diff
@@ -32,6 +33,17 @@ port reportIsVisible : (( String, Bool ) -> msg) -> Sub msg
...
 
+port rawHttpResponse : (Json.Encode.Value -> msg) -> Sub msg
+
+
+type RawHttpResponse
+    = Success -- ()
+    | BadUrl String -- ({"badUrl": "the-bad-url./com"})
+    | Timeout -- ({"timeout": true})
+    | NetworkError -- ({"networkError": true})
+    | BrowserError -- ({"browserError": true})
```
pivotal-jamie-klassen: tests would be better than comments here

web/elm/tests/FlySuccessTests.elm
```diff
@@ -241,6 +271,19 @@ property query description assertion setup =
...
+-- subscription
+
+
+listensForTokenResponse : Property
+listensForTokenResponse setup =
+    test (setupDesc setup ++ ", listens for token response") <|
+        steps setup
+            >> Application.subscriptions
+            >> List.member Subscription.OnTokenSentToFly
+            >> Expect.equal True
```
pivotal-jamie-klassen: maybe a better 'list membership' expectation - this error message just says 'expected false to equal true' which is hard to understand in a hurry.

pivotal-jamie-klassen:
@Lindsayauchin @matthewpereira here's the error page when you're trying to `fly login` with Safari to a TLS-secured Concourse server. Maybe you folks can give some feedback about styles and copy?
Screen Shot 2019-09-16 at 10 52 00 AM

pivotal-jamie-klassen:
lindsays notes:

"however, your token could not be sent to fly because your browser blocked the attempt.

send token to fly directly < [make into button]

[more space]

if that fails, you will need to copy the token to your clipboard, return to fly, and paste your token into the prompt.

[copy token to clipboard]"

Lindsayauchin:
Screen for reference to the above note ^^

Screen Shot 2019-09-18 at 2 23 39 PM

pivotal-jamie-klassen:
@Lindsayauchin
Screen Shot 2019-09-18 at 4 29 52 PM

PR #4429, opened 2019-09-13
atc: fix lidar logging

pivotal-jwinters:
- add start/end logs for scanner and checker
- change check-already-exists to debug level


pivotal-jamie-klassen:
a minor logging change following #4202 - not going to document this specifically.

PR #4438, opened 2019-09-16
container: add worker name to logger data

christophermancini:
This meta data will greatly improve our ability to identify the origin of container / volume errors.


marco-m:
@christophermancini nice one!

Issue #4235, opened 2019-08-07
UX: preview input versions before triggering job build


jamieklassen:
1. move trigger button to left

![Screen Shot 2019-08-12 at 3 35 20 PM](https://user-images.githubusercontent.com/27746614/62892665-d6ab6e00-bd16-11e9-874e-79a32281ae2d.png)

1. add vertical divider between job and build details
1. change job section background to page background grey (hex#3D3C3C)
    * hover color
1. every five seconds call ListJobInputs and remember the data (test how this scales, it's possible this endpoint is expensive to hammer)
1. add down arrow to trigger button
1. when the trigger button is clicked, pop down the preview:

## Dropdown Active State:  
![Screen Shot 2019-08-23 at 3 54 15 PM](https://user-images.githubusercontent.com/27746614/63620043-4a266880-c5be-11e9-8f13-43574b06fdbf.png)

## Trigger Button Hover State:  

![Screen Shot 2019-08-23 at 3 55 23 PM](https://user-images.githubusercontent.com/27746614/63620132-8063e800-c5be-11e9-8a1a-809b06b4ac13.png)

## Trigger Button Selected State: 

![Screen Shot 2019-08-23 at 3 55 04 PM](https://user-images.githubusercontent.com/27746614/63620144-85289c00-c5be-11e9-9f0d-cd0ba20c9097.png)

    * if there is no data to preview, show a spinner
    * have a fixed-width preview
    * some delicate alignment stuff
    * frame-coloured background, slight transparency
    * clicked trigger button is frame color
    * 'trigger new build' button looks like 'save pin comment' button, acts like `+` currently does

1. click anywhere except preview to dismiss it

- [ ] Asset for new downward chevron [chevron_down_ic.svg.zip](https://github.com/concourse/concourse/files/3482360/chevron_down_ic.svg.zip)

Lindsayauchin:
Asset for new downward chevron
[chevron_down_ic.svg.zip](https://github.com/concourse/concourse/files/3482360/chevron_down_ic.svg.zip)

PR #4441, opened 2019-09-16
web: refactor build header

pivotal-jamie-klassen:
we started working on this to pursue #4235, but since switching priorities we want to avoid losing this work so let's just merge the refactoring part.


web/elm/src/Build/Output/Output.elm
```diff
@@ -103,115 +97,102 @@ planAndResourcesFetched buildId ( plan, resources ) model =
...
 handleEnvelopes :
     List BuildEventEnvelope
     -> OutputModel
-    -> ( OutputModel, List Effect, OutMsg )
+    -> ( OutputModel, List Effect )
 handleEnvelopes envelopes model =
     envelopes
         |> List.reverse
-        |> List.foldr handleEnvelope ( model, [], OutNoop )
+        |> List.foldr handleEnvelope ( model, [] )
```
kcmannem: cool function

kcmannem:
yo this is dope, I can read the code but I don't have experience to think more then just the changes look good at the surface. I learned a lot though. Thanks for including me, feel free to do it more often.

vito:
While accepting this I noticed we currently show a 'loading' spinner on a job that has no builds:



Opened this as a separate issue since it's probably unrelated to this refactor: https://github.com/concourse/concourse/issues/4493

pivotal-jamie-klassen:
@vito definitely not from this refactor -- that spinner has been there for years, and only became visible after the sidebar was introduced.

vito:
Noticed some weirdness: if I'm viewing a running build and switch quickly between a finished build and the running build, the header kind of...fuses? with the running build's header:



Rolling back the refactor, it looks like this *wasn't* a problem before:

pivotal-jamie-klassen:
@vito good catch, that's a real bug. I realized that #4487 doesn't actually depend on these changes, so let's not block it (or v5.6.0) on this one.

pivotal-jamie-klassen:
@vito fixed it... although i am still feeling that this component could use more tests.

jomsie:
i figure this'll finish this particular refactor, next we'll be using the algorithm-v3 branch to continue with #4235 (hopefully once that branch gets rebased from master)

Issue #3594, opened 2019-03-26
Panics in pipeline/build scheduling should not bring down the whole Concourse


vito:
Go's `net/http` is careful to recover panics in HTTP handlers and log them instead of crashing the whole server. Can we do the same for general pipeline scheduling and build tracking and anything else that's affected by user data?

One example of a panic: https://github.com/concourse/concourse/issues/3582

vito:
Note: we should definitely not use this to just paper over errors - we should make them very easy for operators to notice. `net/http` does this by logging them to `stderr` instead of `stdout` for example, so maybe we could do that too?

In the linked example, users would have been able to notice this because the build would never finish. They probably would have thought to open a bug, and at least would be able to abort the build to prevent it from hanging the whole pipeline.

marco-m:
I suggest to be careful before embarking along this path, which can be a slippery slope. The example in `net/http` is for user-supplied handlers, while here we are talking about Concourse core itself. Coming from Erlang "crash early" philosophy, I prefer Concourse crashing (I can set up an alarm on high restart frequency) than having no alarms due to a Concourse without one leg that uses a crutch to stand :-)

vito:
@marco-m Erlang's "crash early" are typically limited in scope - they don't bring down the *entire* Erlang VM and require a full system-level process restart, right?

As an operator I would rather set up alerts for panics like this than worry about a bug that allows users to submit pipelines that DOS my untrusted-multitenant Concourse instance.

marco-m:
@vito an Erlang crash begins indeed very limited, the granularity is the same as of a goroutine. Each Erlang "process" is inserted in a supervisor tree, which is what makes Erlang crash early powerful and effective. The crash begins at a leaf of the tree and, if the limited restart doesn't solve the problem, can bubble up the tree until it reaches the Erlang VM. Actually it can even go outside the Erlang VM and reboot the host :-)

You know better than me Concourse ;-), I am only suggesting to be careful.

vito:
@marco-m Agreed - we'll try to be conservative with this, I definitely don't want to sprinkle `recover()` everywhere and call it a day. :+1:

vito:
Adding this to 6.0 since I'm on mobile and can't add it to a project instead. Hopefully we get to it by then!

vito:
[...]

PR #4442, opened 2019-09-16
atc/resource/resource_put: prevent a nil VersionResult from crashing atc

nelsam:
We ran in to a *fun* instance where incorrectly configured resources in users' pipelines were crashing the `atc`.  This covers that case and prevents it from happening.

I didn't _really_ know what to include as an error message, so I just jotted down something that loosely makes sense.  I'm kinda expecting to have to change it during review.

### DIFF
--- a/atc/resource/resource_put.go
+++ b/atc/resource/resource_put.go
@@ -2,6 +2,7 @@ package resource
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/concourse/concourse/atc"
 )
@@ -21,9 +22,10 @@ func (resource *resource) Put(
 
 	vr := &VersionResult{}
 
+	path := "/opt/resource/out"
 	err := resource.runScript(
 		ctx,
-		"/opt/resource/out",
+		path,
 		[]string{resourceDir},
 		putRequest{
 			Params: params,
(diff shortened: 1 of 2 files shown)


christophermancini:
One of the crashes was triggered by a user testing changes to a pipeline and to avoid notifications to slack, redirected their slack YAML Anchors to the time resource. The second crash was caused by a user doing a put to a PR resource before the get step.

vito:
related: #3594

clarafu:
@christophermancini Would you be able to add a release note for this fix? We are trying out a process where we add a release note whenever we merge a PR. Thanks!!

nelsam:
d'oh, I knew I was forgetting something!

clarafu:
Don't worry about it, it's kinda just something new we are trying and I don't think we have documented the process anywhere yet aha

nelsam:
Yeah, just double-checked - I actually _saw_ that I was supposed to append to release notes in `latest.md` in the `CONTRIBUTING` guide, but just forgot about it in the PR.  I'll make another.

atc/resource/resource_put.go
```diff
@@ -21,9 +22,10 @@ func (resource *resource) Put(
 
 	vr := &VersionResult{}
 
+	path := "/opt/resource/out"
```
marco-m: This could be a `const` instead.

Why is immutable better? Because when as a reviewer I read `const`, I can relax and not worry if somewhere in the code this value is changed.

Issue #4341, opened 2019-09-02
The concourse_workers_containers Prometheus metric reports values from old, currently non-existent workers


kmdouglass:
## Bug Report

Four days ago we deployed the release candidate 7c63f45. Immediately after doing so we noticed that the `concourse_workers_containers` Prometheus metric began reporting a number of containers for workers that no longer exist but at one point did. These workers existed at one point because they were part of an AWS autoscaling group (ASG) and eventually were retired and destroyed as part of the natural lifecycle of the ASG.

Prior to this RC, the metric would not report the number of containers for previous workers; it would only report the number of containers for active workers. Now it reports the number of containers for both the current workers and previous but currently non-existent workers.

After four days of running the release candidate in production, we have accumulated a large number of time series for workers that no longer exist. This can be seen in the plot below:

![image](https://user-images.githubusercontent.com/3697676/64123592-1e1c9b80-cda5-11e9-8a3f-25c5e72c48c5.png)

The plot shows the number of time series that belong to the `concourse_workers_containers` metric on the y-axis and the time since ~2019/08/29 on the x-axis. (We deployed the RC around 8:30 AM on the 29'th of August.) The PromQL query is:

```
count(count by (worker)(concourse_workers_containers{worker!~".*mac-.*",worker!="deployer"}))
```

The inner `count by (worker)` essentially returns a vector of 1's, one for each time series returned by `concourse_workers_containers`, and the second count sums all of these 1's. The plot is constant over the weekend because we had no workers running. Once Monday, September 2 came and new workers were spawned, the number of time series began to grow again.

## Steps to Reproduce

1. Deploy RC 7c63f45 into production with a pool of workers that can dynamically grow and shrink over time.
2. Observe that the `concourse_workers_containers` Prometheus metric will continue to report a value for workers that existed at one point but no longer exist.

## Expected Results

I expect that the Prometheus `concourse_workers_containers` metric to report a number of containers for workers that exist at the time the metric was scraped. Old workers that no longer exist should not be included in the metric.

## Actual Results
The `concourse_workers_containers` metric reports both the number of containers for active workers and for workers that existed at one point but no longer exist.

## Additional Context

- The previous RC that we were running that DID NOT have this bug is 5672114.
- The output of `fly workers` is what I would expect; it only shows the workers currently registered with the ATC and not the old ones.
[...]

PR #4446, opened 2019-09-17
Issue 4341 - Fix the Prometheus metrics garbage collector

kmdouglass:
This PR addresses issue #4341 by proposing a fix to the Prometheus metrics garbage collector. As described in the issue, the problem with the GC is that the methods for deleting metrics from the Prometheus client library will fail when the labels that are provided as arguments do not form an exact match. Recently two additional labels were added to the `workerContainers` and `workerVolumes` metrics but the garbage collector logic was not updated to reflect these changes. As a result, the garbage collector now fails to clean up these metrics when a worker disappears.

The solution proposed in this PR is as follows:
- Maintain a map of the labels used for the `workerContainers`, `workerVolumes`, and `workerTasks` metrics as part of the `PrometheusEmitter`'s state.
- The map of labels is nested; the outer map is keyed by the worker name; the inner map is keyed by a serialized form of the metrics labels. The serialization is done because instances of `prometheus.Labels` themselves are not hashable and can therefore not be used in a set-like data structure such as `map[T]struct{}`
- During one cycle of the garbage collector, pass the labels that are associated with old workers to the [GaugeVec.Delete](https://godoc.org/github.com/prometheus/client_golang/prometheus#GaugeVec.Delete) method to remove the old metrics from memory. This avoids having to use the [GaugeVec.DeleteLabelValues](https://godoc.org/github.com/prometheus/client_golang/prometheus#GaugeVec.DeleteLabelValues) method whose input arguments need to be explicitly enumerated.
- The code as it was structured was not easily unit tested for a number of reasons. I created a small dummy interface that allowed unit testing the metric removal logic.

## Testing

- I added two unit tests to verify the logic for deleting old metrics
- I verified that metrics from stale workers are removed in the `docker-compose` environment by exposing port 9090 from the `web` service and setting a prometheus bind port and IP in the `web` command:

```yaml
<snip>
web:                                                                                                                                                                                                             
    build: .                                                                                                                                                                                                       
    volumes:                                                                                                                                                                                                       
    - .:/src                                                                                                                                                                                                       
    command: web --prometheus-bind-port=9090 --prometheus-bind-ip=0.0.0.0                                                                                                                                          
    depends_on: [db]                                                                                                                                                                                               
    ports:                                                                                                                                                                                                         
    - 8080:8080                                                                                                                                                                                                    
    - 9090:9090
<snip>
```

Then I killed the worker service and verified that the metric for the worker disappeared after 5 minutes by `curl`ing the metric from the ATC:

```
watch -n 30 curl -s localhost:9090 | grep ^[^#] | grep concourse_workers
```


atc/metric/emitter/prometheus.go
```diff
@@ -46,15 +48,32 @@ type PrometheusEmitter struct {
...
+	workerLastSeen         map[string]time.Time
+	mu                     sync.Mutex
 }
 
 type PrometheusConfig struct {
 	BindIP   string `long:"prometheus-bind-ip" description:"IP to listen on to expose Prometheus metrics."`
 	BindPort string `long:"prometheus-bind-port" description:"Port to listen on to expose Prometheus metrics."`
 }
 
+func serializeLabels(labels *prometheus.Labels) string {
```
marco-m: Needs a comment to explain why we have to do this.

marco-m:
@cirocosta this is ready for review. If you don't have time, could you please suggest somebody from the Concourse team? Thanks! We have some alarms that depend on this metrics and we are getting false positives daily :-)

marco-m:
@pivotal-jamie-klassen we enjoyed reading your review here in the team :-)

I think that one of the root causes of the mess is that Prometheus is a pull system, while Concourse has been designed thinking about push systems (like all other metrics systems except Concourse actually).

I remember various tickets and PRs in Concourse describing this problem of staleness. I pointed @cirocosta to this blog article that opened my eyes: https://www.robustperception.io/its-easy-to-convert-pull-to-push (the author of that blog is one of the Prometheus main committers).

Last but not least, thanks!

kmdouglass:
Indeed @pivotal-jamie-klassen , it was a bit tricky working on this bug, especially because I had heard that you guys were considering refactoring the metrics logic. I hope this PR can help bridge the gap until that happens.

Thanks a lot for the review!

cirocosta:
@marco-m that's a pretty good one! would you mind sharing your thoughts on https://github.com/concourse/concourse/pull/4247? 🤔  thanks!

thanks for the review, @pivotal-jamie-klassen !

Issue #3522, opened 2019-03-15
Support atomically switching pins


vito:
Occasionally I want to swap from one pinned version to another. Right now I have to unpin the old version and pin a new one which doesn't feel as nice as if I could just click the new one. It also has a race condition since something could be scheduled in between clicks.

vito:
As acceptance, I would like to run `pin-resource-version` twice and have the second run swap the pin.

xtremerui:
@pivotal-jamie-klassen @vito just want to confirm that in this issue it will only change the API part so one can pin a resource with a version without unpin it first.

 Do we need to create an issue for the UI work?

pivotal-jamie-klassen:
implemented by #4448 on the backend and #4561 on the frontend

PR #4448, opened 2019-09-17
API pin switching 

xtremerui:
implement the #3522


xtremerui:
Hi, @pivotal-jamie-klassen the backend changes are done. Would you prefer continue the UI work on the same branch or doing it as a separate issue?

pivotal-jamie-klassen:
@xtremerui let's split out another PR for the web work - this doesn't break anything

pivotal-jamie-klassen:
release note in #4564

PR #4463, opened 2019-09-18
only bump cache index when saving new versions

### LINKED ISSUE #4435: Lidar will always save versions everytime a check is run
After deploying lidar to our prod environment, we saw a huge spike in DB CPU 

image

From the metrics above, the first sharp increase and then plateau was when we first deployed with lidar and then the small spikes after were when we changed the checking interval to be 1h for all the resources.

We later figured out the problem to being that within lidar, the checker will run SaveVersions everytime https://github.com/concourse/concourse/blob/163450ad7fa0f3fcece4b361991b16c1114277da/atc/exec/check_step.go#L161
vs. radar which only ran SaveVersions when the versions returned has only one version and that version is the same as the one passed into the FromVersion
https://github.com/concourse/concourse/blob/a009293a9e868fd756c0e356402e23eed04ffe3a/atc/radar/resource_scanner.go#L412

This, as a result, caused all the checks to save versions and bump the pipeline cache index, which will cause the scheduler to run the super heavy LoadVersionsDB queries every single time a check is run. 

cc. @pivotal-jwinters

pivotal-jwinters:
fixes #4435

### DIFF
--- a/atc/db/resource_config_scope.go
+++ b/atc/db/resource_config_scope.go
@@ -78,8 +78,10 @@ func saveVersions(conn Conn, rcsID int, versions []atc.Version) error {
 
 	defer Rollback(tx)
 
+	var bumpCache bool
+
 	for _, version := range versions {
-		_, err = saveResourceVersion(tx, rcsID, version, nil)
+		newVersion, err := saveResourceVersion(tx, rcsID, version, nil)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pivotal-jamie-klassen:
This was a regression introduced by #4202, which is documented in #4475

Issue #4217, opened 2019-08-05
Improve web HTTP responses with `gzip` support


dpb587-pivotal:
## What challenge are you facing?

The UI is slow to load in targets with (very) large numbers jobs or bandwidth-limited connections.

Semi-related, the UI continues to request `jobs`/`pipelines` even before the last calls complete which causes large numbers of queued requests over time with limited connectivity (which exaggerates the impact of uncompressed resources; and might be worth tackling separately).

## What would make this better?

Enabling `gzip` compression for HTTP responses. It's common behavior of web servers and can significantly increase throughput.

When running Concourse behind a reverse proxy, I can enable `gzip` and see the following bandwidth differences on notable endpoints. The bandwidth impact is secondary; it's primary impact is typically faster load times.

Target | Resource | Concourse direct | Caddy reverse proxy w/ `gzip` |
-|-|-:|-:
| team | `pipelines` | 98 KB | 3.6 KB |
| team | `jobs` | 2764 KB | 186 KB |
| team | `resources` | 264 KB | 17 KB |
| wings | `pipelines` | 27 KB | 6 KB |
| wings | `jobs` | 622 KB | 62 KB |
| wings | `resources` | 93 KB | 10 KB |

I have been using the reverse proxy w/ `gzip` approach in my own environments for a while. Since many users access `web` directly, I think it's worth considering built-in support for broader benefit.

## Are you interested in implementing this yourself?

Sure. Pointers or concerns are welcomed.

vito:
Huh, I could have sworn we already did this, but I guess not. I'll chuck it in the API backlog; seems like low-hanging fruit.

PR #4470, opened 2019-09-19
gzip all api response

xtremerui:
Remove gzip compression from build events handler.
Instead, it compress all API response from top level.

This should help with general bandwidth saving and particularly the slowness loading of concourse dashboard.

Implements #4217


xtremerui:
Acceptance:

1. Run a local concourse and set couple pipelines with resources.
2. Open chrome dev tool `network` tab and investiage the size of the request and the actual content length by mouse hover the number of `size` column
3. With gzip enabled, all requests to API should indicates `gzip` as `content-encoding` in response header and the mouse hover on step 2 should show `aaa transfered over network. resource size: bbb`, where aaa should be smaller than bbb. The bigger the actual size bbb, the more the compression could save.

kcmannem:
https://medium.com/oyotech/how-brotli-compression-gave-us-37-latency-improvement-14d41e50fee4

this might make it even faster

kcmannem:
I don't really see a golang library so maybe we pass on this.

vito:
Would it be possible to just use https://github.com/nytimes/gziphandler instead of writing our own handler? It seems fairly popular + well-supported.

xtremerui:
@vito do we need any config provided by this library? 

```
type GzipResponseWriter struct {
	http.ResponseWriter
	index int // Index for gzipWriterPools.
	gw    *gzip.Writer

	code int // Saves the WriteHeader value.

	minSize int    // Specifed the minimum response size to gzip. If the response length is bigger than this value, it is compressed.
	buf     []byte // Holds the first part of the write before reaching the minSize or the end of the write.
	ignore  bool   // If true, then we immediately passthru writes to the underlying ResponseWriter.

	contentTypes []parsedContentType // Only compress if the response is one of these content-types. All are accepted if empty.
}
```
If we do, it still requires some tinker for concourse to utilize it. If we just gonna use the default handler, it does pretty much the same as this PR.

For latter, we could avoid increasing the binary size a bit?
We could still spike on it to compare.

vito:
I'm more worried about having to write and maintain the code ourselves than binary size. :slightly_smiling_face: 

I don't see why we would need to set those ourselves, but they expose them as options here: https://godoc.org/github.com/NYTimes/gziphandler#GzipHandlerWithOpts

xtremerui:
@vito I have committed the change but it failed testflight and watsjs by [hanging](https://ci.concourse-ci.org/teams/main/pipelines/prs/jobs/watsjs/builds/1958#L5d8a592d:299). I digged around and notice this in their code https://github.com/nytimes/gziphandler/blob/dd0439581c7657cb652dfe5c71d7d48baf39541d/gzip.go#L122

So, our tests hangs because it is watching for build events. In certain scenario where response is small enough (total size that is less than 1400 bytes) and no content-length specificed, the `buf` will hold flushing and wait until the `MinSize` is reached. In this case the api will not respond.

For concourse build events there is no way to tell how big the total size is beforehand (all other concourse APIs have content-length set), which means we would have to adjust the `MinSize` to 0 or a number to allow build events flushing disregarding the size. The reason that `MinSize` is introduced in the library is [here](https://github.com/nytimes/gziphandler/blob/dd0439581c7657cb652dfe5c71d7d48baf39541d/gzip.go#L32-L35), which I feel setting it to 0 is not ideal.

What do you think? Thx!

vito:
@xtremerui Maybe we could add a gzip wrappa that wraps the events endpoint with `MinSize` of 0 and leaves all others as the default?

xtremerui:
@vito updated with this good idea. It wont hang on small events anymore.

atc/wrappa/compression_wrappa.go
```diff
@@ -0,0 +1,40 @@
...
+func (wrappa CompressionWrappa) Wrap(handlers rata.Handlers) rata.Handlers {
+	wrapped := rata.Handlers{}
+
+	for name, handler := range handlers {
+
+		// always gzip for events
+		if name == atc.BuildEvents {
+			gzipEnforcedHandler, err := gziphandler.GzipHandlerWithOpts(gziphandler.MinSize(0))
+			if err != nil {
+				wrappa.Logger.Debug("Failed to create gzip handler with options", lager.Data{"error": err})
```
vito:
```suggestion
				wrappa.Logger.Error("failed-to-create-gzip-handler", err)
```

or just eliminate the logger and ignore the error since it should be impossible (it only errors on invalid options)

pivotal-jamie-klassen:
release note in #4564

Issue #4407, opened 2019-09-11
Fly curl cannot parse curl arguments


xtreme-sameer-vohra:
## Bug Report

`fly curl` with `curl options` fails to parse the options correctly and fails

## Steps to Reproduce

`fly -t local curl /api/v1/log-level "some-query-param" "-X"  "PUT"  "-d " "trace"`

## Expected Results
Set log-level to trace

## Actual Results
`error: unknown flag `X'`

## Version Info

* Concourse version: v5.5.x ( also expect this is an issue on older version of fly )

* Deployment type (BOSH/Docker/binary): n/a
* Infrastructure/IaaS: n/a
* Browser (if applicable): n/a
* Did this used to work? No

YoussB:
From the documentation on the website: 
https://concourse-ci.org/administration.html#fly-curl
You are supposed to pass `--` after the path in order for it to work. 

I tried doing that though but still doesn’t work 😁

wagdav:
I ran into this problem and I wanted to understand what's going on here.

It seems that since this https://github.com/concourse/concourse/commit/1d403d3fc94d8406b27c4f62ee71db41209eabf8 the invocation of fly curl changed.

The following works:
```
fly -t <target> curl /api/v1/log-level "" -- -X PUT -d info  -H "Content-Type: application/octet-stream"
```

To break it down into its components:
* `PATH`: `/api/v1/log-level` 
* `QUERY`: empty string, as this gets appended to the path with a `?`
* `--`: to separator for curl flags
* `-X PUT`, HTTP verb
* `-d info`, the body of the request (only debug, info, error and fatal are valid values)
* `-H "Content-Type: application/octet-stream`, by default the request uses `application/x-www-form-urlencoded` header which doesn't work in this case. The log-level handler expects a simple bytestream as the request body.

I remember that,
```
fly -t builder curl -- api/v1/log-level -X PUT -d "debug"
```
used to work, which was somewhat simpler :-)

xtreme-sameer-vohra:
Ah, I didn't know about `--`

PR #4473, opened 2019-09-20
[#4407] Update fly curl help text

xtreme-sameer-vohra:
Address #4407


pivotal-jamie-klassen:
Guessing that the flags library doesn't gracefully handle optional positional arguments? Maybe the query should be a flag

xtreme-sameer-vohra:
Yea, since query is an optional param, it might not make sense to have it as a positional argument. 
This PR was to just add the `--` delimiter   to differentiate between fly args vs. curl args to `fly curl -h` as its on the website but not in the cli itself.
Changing the query param should be a separate issue 🤔

deniseyu:
The description is better than no help text but it still confuses me lol. An example would maybe help. But we can do that iteratively!

Issue #4478, opened 2019-09-21
DNS response returns uncompressed when using Concourse Worker DNS


int-tt:
## Bug Report

Worker DNS does not do [Message compression](https://www.ietf.org/rfc/rfc1035.txt) when returning a response, so it returns with a large response

## Steps to Reproduce
execute task
https://gist.github.com/int-tt/e814b3886fc362ce2d03c11f2a86956d#file-use-worker-dns-yaml
## Expected Results
execute task or use `inttt/concourse:dev-set-compress` image 
https://gist.github.com/int-tt/e814b3886fc362ce2d03c11f2a86956d#file-external-dns-yaml

DNS reponse with message compression returned
```
; <<>> DiG 9.14.3 <<>> ecr-emulate.lnteger.dev @8.8.8.8
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 15180
;; flags: qr rd ra; QUERY: 1, ANSWER: 7, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 512
;; QUESTION SECTION:
;ecr-emulate.lnteger.dev.       IN      A

;; ANSWER SECTION:
ecr-emulate.lnteger.dev. 3      IN      CNAME   proxy-ap-n-proxylb-1r9iukj5vmg6a-733833415.lnteger.dev.
proxy-ap-n-proxylb-1r9iukj5vmg6a-733833415.lnteger.dev. 3 IN A 192.168.0.1
proxy-ap-n-proxylb-1r9iukj5vmg6a-733833415.lnteger.dev. 3 IN A 192.168.0.2
proxy-ap-n-proxylb-1r9iukj5vmg6a-733833415.lnteger.dev. 3 IN A 192.168.0.3
proxy-ap-n-proxylb-1r9iukj5vmg6a-733833415.lnteger.dev. 3 IN A 192.168.0.4
proxy-ap-n-proxylb-1r9iukj5vmg6a-733833415.lnteger.dev. 3 IN A 192.168.0.5
proxy-ap-n-proxylb-1r9iukj5vmg6a-733833415.lnteger.dev. 3 IN A 192.168.0.6

;; Query time: 8 msec
;; SERVER: 8.8.8.8#53(8.8.8.8)
;; WHEN: Sat Sep 21 11:05:47 UTC 2019
;; MSG SIZE  rcvd: 205
```

[...]

PR #4479, opened 2019-09-21
feat: enable dns message compression

int-tt:
fixed: #4478

### DIFF
--- a/worker/workercmd/dns.go
+++ b/worker/workercmd/dns.go
@@ -22,7 +22,8 @@ func (config DNSConfig) Server() (*dns.Server, error) {
 		for _, server := range resolvConf.Servers {
 			response, _, err := client.Exchange(r, fmt.Sprintf("%s:%s", server, resolvConf.Port))
 			if err == nil {
-				w.WriteMsg(response)
+				response.Compress = true
+				_ = w.WriteMsg(response)
 				break
 			}
 		}


kcmannem:
lets merge this later in the week for 5.7. We're planning on shipping 5.6 soon which includes lidar and best to avoid any last min changes.

int-tt:
Thanks!!!

worker/workercmd/dns.go
```diff
@@ -22,7 +22,8 @@ func (config DNSConfig) Server() (*dns.Server, error) {
 		for _, server := range resolvConf.Servers {
 			response, _, err := client.Exchange(r, fmt.Sprintf("%s:%s", server, resolvConf.Port))
 			if err == nil {
-				w.WriteMsg(response)
+				response.Compress = true
```
vito: Is there any problem with this unconditionally being set to `true`? I didn't know DNS compression was a thing, but I guess everything should be compatible with it; looks like it's been around for [a while](https://www.ietf.org/rfc/rfc1035.txt).

I guess we have to explicitly set this because the library transparently handled the compression and gave us a `*Msg`, but since we're going to re-transmit it we need to tell it to compress it again.

int-tt: 私は2つの理由で大丈夫だと認識しています。
* Dockerの内部DNSも常に圧縮をしているため。
https://github.com/docker/libnetwork/blob/master/resolver.go#L537
* ライブラリの内部で圧縮可能かどうか検査しているため
https://github.com/miekg/dns/blob/master/msg.go#L767

vito: sounds good! i'll go ahead and merge

Issue #4105, opened 2019-07-05
Allow to specify the text for a badge


rixx:
## What challenge are you facing?

All badges currently bear the title "build", next to the build status. A project will often have several relevant badges to display, but they look interchangeable.

## What would make this better?

I think the solution that has the least impact on people already using the badge feature would be to allow to specify the badge title via the URL (`?title=documentation`, for instance). This is how shields.io does it, too.

## Are you interested in implementing this yourself?

I haven't got any experience with go projects, but sure, if somebody helps me through the process, I'd be happy to work on this.

vito:
Makes sense! Seems like this would just need to be passed to the templates [here](https://github.com/concourse/concourse/blob/b1a1b19d6a7b032c487946a590fa4181739635ad/atc/api/jobserver/badge.go#L85). 👍

PR #4290, opened 2019-08-17
support custom title text for badge

atthavit:
Allow users to change the "build" text on badges with `?title=` param.

Fixes #4105

PR #4480, opened 2019-09-21
change: badge rendering considers "title" query param

ProvoK:
Hello!

This PR allows users to change the "build" text on badges with ?title= param.
Fixes https://github.com/concourse/concourse/issues/4105

I've opened another PR since https://github.com/concourse/concourse/pull/4290 were "stalled" and had a quite big bug (it modifies a reference to the badges, hence subsequent calls are affected) .

Some considerations: 
1. with title longer than 5/6 chars, the badge is quite ugly since its dimensions are baked in multiple places. Probably a different PR for making this rendering "responsive" is better in terms of isolation.
2. It could be nice to have something also for the "content" (aka "success|failed|etc"), maybe set from a query parameter like `?content=<some-regex>` that uses the actual job's output.

Signed-off-by: Vittorio Camisa <vittorio.camisa@gmail.com>

### DIFF
--- a/atc/api/jobs_test.go
+++ b/atc/api/jobs_test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"strings"
 	"time"
 
 	"github.com/concourse/concourse/atc"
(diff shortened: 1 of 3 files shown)


atc/api/jobs_test.go
```diff
@@ -577,6 +578,57 @@ var _ = Describe("Jobs API", func() {
...
+				})
+				It("html escapes title", func() {
+					response, err := client.Get(server.URL + "/api/v1/teams/some-team/pipelines/some-pipeline/jobs/some-job/badge?title=%24cov")
+					Expect(err).NotTo(HaveOccurred())
+
+					body, err := ioutil.ReadAll(response.Body)
+					Expect(err).NotTo(HaveOccurred())
+
+					Expect(strings.Contains(string(body), `
+      <text x="18.5" y="15" fill="#010101" fill-opacity=".3">$cov</text>
```
vito: `$cov` isn't HTML escaped; should the example input be something like `<cov>` instead so it shows the escaping?

atc/api/jobserver/badge.go
```diff
@@ -95,6 +98,15 @@ type badgeTemplateConfig struct {
 	StatusTextWidth string
 	Status          string
 	FillColor       string
+	Title           string
+}
+
+func enrichFromQuery(params url.Values, original *Badge) Badge {
```
vito: This code seems a little odd. I think we can just change `BadgeForBuild` to not return a pointer and then inline the `.title` attribute assignment, since then it would only be modifying a copy.

atc/api/jobserver/badge.go
```diff
@@ -4,23 +4,25 @@ import (
...
-	badgeAborted = Badge{width: 90, fillColor: `#8f4b2d`, status: `aborted`}
-	badgeErrored = Badge{width: 88, fillColor: `#fe7d37`, status: `errored`}
+	badgePassing = Badge{width: 88, fillColor: `#44cc11`, status: `passing`, title: `build`}
+	badgeFailing = Badge{width: 80, fillColor: `#e05d44`, status: `failing`, title: `build`}
+	badgeUnknown = Badge{width: 98, fillColor: `#9f9f9f`, status: `unknown`, title: `build`}
+	badgeAborted = Badge{width: 90, fillColor: `#8f4b2d`, status: `aborted`, title: `build`}
+	badgeErrored = Badge{width: 88, fillColor: `#fe7d37`, status: `errored`, title: `build`}
 )
 
 type Badge struct {
```
vito: Can we eliminate `badgeTemplateConfig` and just use this type directly? If we just make its fields and methods public (`.statusWidth`, `.statusTextWidth`) the template should "Just Work".

atc/api/jobs_test.go
ProvoK: Probably the test name isn't clear, because my intention was to verify that colling the endpoint with someting like  "badge?title=**%24cov**" resulted i "$cov" into the badge

atc/api/jobserver/badge.go
ProvoK: I agree with you, this kind of stuff are a bit odd with Go.

Actually, what you are saying was my very first though and implementation. 
I changed it back because I didn't want to do such a big change, but if you're ok with that I'm going to change it back :D

atc/api/jobserver/badge.go
ProvoK: I'll look into it! Seems very reasonable

ProvoK:
> Hey, sorry for the slow turnaround on this! Left a few changes to suggest if you've got time. 🙂 This endpoint doesn't get much attention so I'd like to sneak in some small cleanups if that's OK.

I should have changed all you requested but the test file

PR #4489, opened 2019-09-24
add user role

christophermancini:
- allows a role that can abort builds, trigger builds, and hijack containers to troubleshoot builds
- move hijack container down to new user role, meaning it adds perm to operator

If folks don't like the use of `user` for this role name, I am open to suggestions. Maybe `job-operator` or `smooth-operator`? The role `developer` was proposed by @deniseyu in the [comment below](https://github.com/concourse/concourse/pull/4489#issuecomment-534737023).

**NOTE:** I will update the release notes & documentation in this PR once the proposal has been approved.


pivotal-jamie-klassen:
I lol'd at smooth-operator. Your failing unit test comes from
https://github.com/concourse/concourse/blob/master/atc/api/accessor/accessor_test.go#L782

deniseyu:
What about "developer" as a name for the role? It is kind of open-ended 😬 but it mirrors existing concepts in Cloud Foundry and Kubernetes like "a user who can do read/write operations, but only in their operator-granted namespace". I'm making a bit of a guess here that the use case is for organizations where you'd have a centralized platform team admin'ing the Concourse deployment, and a number of dev teams using their own Concourse teams (and possibly their own workers).

@xtreme-sameer-vohra points out that hijacking could potentially escalate privileges. When you hijack into a container you can find all the credentials that the task or the step has access to. Not a showstopper necessarily, just something to keep in mind!

In a future where we have more granular role design in general, this could potentially be split into two roles, to make "hijackable" intent clearer: one role for "developer" and one for "hijacker".

christophermancini:
@deniseyu, you guessed correctly. We have a [monorepo](https://blog.digitalocean.com/cthulhu-organizing-go-code-in-a-scalable-repo/) and our team (CICD) maintains the pipelines that complete the standard tasks, e.g. `go vet/fmt/lint`, unit tests, functional tests, etc. I need the engineers who make changes to be able to hijack a build in order for them to be able troubleshoot why their build failed. However, I don't want them pausing / unpausing, pinning resources, renaming the pipeline, etc.

I am perfectly fine with `developer` for the role name, `user` was just the first thing that came to mind.

Rukenshia:
Is there a general roadmap for supporting more / different / customisable roles as well? We have a use case very similar to this, but definitely don't want the role to be able to hijack containers (basically a "Release coordinator" that just pushes the deploy button).
I work in a more traditional environment when it comes to our system role definitions, and have been struggling with compliance when it comes to concourse. Not sure if this is the right place to list all of them though, but I can imagine there are others out there with similar requirements.

deniseyu:
@Rukenshia there's some discussion about building out RBAC in this RFC: https://github.com/concourse/rfcs/blob/master/003-rbac/proposal.md

It would definitely be nice to move closer to a world where bundles of permissions could be composed and distributed to users in a more granular way. In terms of the [roadmap](https://blog.concourse-ci.org/core-roadmap-towards-v10/), it's not concretely scheduled yet but we're always listening to user signals like this to make prioritisation decisions.

deniseyu:
Hey folks, a small update here that may interest you:

@evanchaoli added PR #4657, which will enable operators to override the default mapping of roles to permissions. This might get you closer to the level of granularity that you're looking for, since you can take one of the existing user roles and revoke or add permissions as needed.

I get that adding a new role with a different set of defaults might still be useful though. Sorry it's taken us a while to get back to you on this; @christophermancini if you don't mind rebasing and renaming the role to `developer` then I'm happy to merge this. 

In terms of documentation, if you could add a comment to this issue https://github.com/concourse/docs/issues/277 with a brief description of the new role, then we can tackle the documentation bit as part of doing that work. It seems unfair to ask you to document the whole world of permissions and roles, since that isn't currently in our user-facing docs at all.

Please note that with the configurable RBAC PR merged in, the global list of permission names is in [a different file now](atc/api/accessor/role_action_map.go).

christophermancini:
@deniseyu excellent, TY for the update. I will discuss with our team to see what we want to do.

christophermancini:
Based on the new features coming down the pipe and our inability to push this forward, I am going to close this for now. TY Pivotal team for your support.

PR #4492, opened 2019-09-24
fly: always emit output in format-pipeline

agurney:
The `format-pipeline` command had logic to suppress output in the case where the formatted pipeline was unchanged from the input. This complicates use of the command in automated workflows, where we'd like to use `format-pipeline` as a canonicalization step. The new behavior lets the output be consumed unconditionally. For example, one can find out whether two pipelines - say, one stored in git and one retrieved from `fly get-pipeline` - are the same after `format-pipeline` has been applied to both.

Also add tests for the command, to demonstrate how a given input pipeline with various formatting oddities will be converted into a standard form, and that reformatting will preserve that form.

### DIFF
--- a/fly/integration/format_pipeline_test.go
+++ b/fly/integration/format_pipeline_test.go
@@ -1,0 +1,143 @@
+package integration_test
+
+import (
+	"io/ioutil"
+	"os"
+	"os/exec"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/onsi/gomega/gexec"
+)
+
+var _ = Describe("Fly CLI", func() {
+	Describe("format-pipeline", func() {
+		var (
+			configFile   *os.File
+			inputYaml    []byte
+			expectedYaml []byte
+		)
+
+		BeforeEach(func() {
+			var err error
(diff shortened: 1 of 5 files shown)


fly/integration/format_pipeline_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+				<-sess.Exited
+				Expect(sess.ExitCode()).To(Equal(0))
+
+				newYaml, err := ioutil.ReadFile(configFile.Name())
+				Expect(err).NotTo(HaveOccurred())
+				Expect(newYaml).To(MatchYAML(expectedYaml))
+			})
+
+			It("is idempotent", func() {
```
xtremerui: Is it a must to test idempotent here? IMHO the process is pretty straight forward and with integrity. Unless there is something fishy that might happen between yaml unmarshal and marshal?

agurney: It just seemed like a useful property to expect, albeit as you say there's no real reason to not expect it to happen.

agurney:
@xtremerui Do you want me to just remove that test?

xtremerui:
@agurney could you add a release note of it? Thx!

agurney:
Rebased with added release note

Issue #4484, opened 2019-09-23
Ordering of teams in the left hand menu


superbeeny:
## What challenge are you facing?

Our Concourse deployment has a lot of teams, with over a hundred in the left hand list it can be hard or sometimes impossible to find the one you want.

## What would make this better?

Ordering of teams, even just alphabetical would be awesome

## Are you interested in implementing this yourself?

Maybe :)

pivotal-jamie-klassen:
Applying an ordering like https://github.com/concourse/concourse/blob/master/web/elm/src/Dashboard/Group.elm#L53-L56
(which is basically sort by role and then alphabetically on team name to break ties) might be a fairly uncontroversial first pass.

Issue #4483, opened 2019-09-23
Teams with No Pipelines Do not appear on Left-hand-side List


keithkroeger:
## What challenge are you facing?

As a user, when I connect to Concourse, I don't see my team listed on the Left-hand-side navigation bar if I have no pipelines

## What would make this better?

Display my team and just note that I've no pipelines, please

PR #4494, opened 2019-09-24
api query transaction

### LINKED ISSUE #4373: API endpoints that uses multiple queries might result in terrible things
In the lidar drills environment, we found that if the API connection pool was maxed out in connections, the web memory usage increased at a crazy rate.

@vito theorizes that it is because for this specific case where the API requests were for ListAllJobs for the dashboard, it would result in three different db queries (from VisibleJobs call) which each acquire a connection (https://github.com/concourse/concourse/blob/master///atc/db/pipeline.go#L1045:20).

If there was a backlog of requests due to slowness with the db, most of the requests will wait to acquire a connection on each of these three calls. However instead of one request acquiring a connection, loading all the data, and returning, it will compete with future requests which *also* try to acquire a connection three times. While it's waiting for the second or third database call, it will be holding whatever data it has loaded so far, which leads to a memory leak.

A possible solution could be to have all API endpoints use only one transaction.

julia-pu:
Fixes #4373 

When multiple queries are made, changed the use of conn to use transactions instead.
Also cleaned up some queries that could be made in a single query instead.

### DIFF
--- a/atc/db/pipeline_factory.go
+++ b/atc/db/pipeline_factory.go
@@ -26,35 +26,18 @@ func NewPipelineFactory(conn Conn, lockFactory lock.LockFactory) PipelineFactory
 
 func (f *pipelineFactory) VisiblePipelines(teamNames []string) ([]Pipeline, error) {
 	rows, err := pipelinesQuery.
-		Where(sq.Eq{"t.name": teamNames}).
-		OrderBy("team_id ASC", "ordering ASC").
+		Where(sq.Or{
+			sq.Eq{"t.name": teamNames},
+			sq.Eq{"public": true},
+		}).
+		OrderBy("public ASC", "team_id ASC", "ordering ASC").
 		RunWith(f.conn).
 		Query()
 	if err != nil {
 		return nil, err
 	}
 
-	currentTeamPipelines, err := scanPipelines(f.conn, f.lockFactory, rows)
-	if err != nil {
-		return nil, err
-	}
-
-	rows, err = pipelinesQuery.
-		Where(sq.NotEq{"t.name": teamNames}).
-		Where(sq.Eq{"public": true}).
-		OrderBy("team_id ASC", "ordering ASC").
-		RunWith(f.conn).
-		Query()
-	if err != nil {
-		return nil, err
-	}
-
-	otherTeamPublicPipelines, err := scanPipelines(f.conn, f.lockFactory, rows)
-	if err != nil {
-		return nil, err
-	}
-
-	return append(currentTeamPipelines, otherTeamPublicPipelines...), nil
+	return scanPipelines(f.conn, f.lockFactory, rows)
 }
 
 func (f *pipelineFactory) AllPipelines() ([]Pipeline, error) {
--- a/atc/db/pipeline_factory.go
+++ b/atc/db/pipeline_factory.go
@@ -25,19 +25,48 @@ func NewPipelineFactory(conn Conn, lockFactory lock.LockFactory) PipelineFactory
 }
 
 func (f *pipelineFactory) VisiblePipelines(teamNames []string) ([]Pipeline, error) {
+	tx, err := f.conn.Begin()
+	if err != nil {
+		return nil, err
+	}
+
+	defer Rollback(tx)
+
 	rows, err := pipelinesQuery.
-		Where(sq.Or{
-			sq.Eq{"t.name": teamNames},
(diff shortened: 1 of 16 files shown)


atc/db/pipeline_factory.go
```diff
@@ -26,35 +26,21 @@ func NewPipelineFactory(conn Conn, lockFactory lock.LockFactory) PipelineFactory
 
 func (f *pipelineFactory) VisiblePipelines(teamNames []string) ([]Pipeline, error) {
 	rows, err := pipelinesQuery.
-		Where(sq.Eq{"t.name": teamNames}).
-		OrderBy("team_id ASC", "ordering ASC").
+		Where(sq.Or{
+			sq.Eq{"t.name": teamNames},
+			sq.And{
+				sq.NotEq{"t.name": teamNames},
```
clarafu: I think we can have it only check that it is `public: true` and remove the `t.name != teamNames`. 🤔

atc/db/pipeline_factory.go
```diff
@@ -26,35 +26,21 @@ func NewPipelineFactory(conn Conn, lockFactory lock.LockFactory) PipelineFactory
...
+				sq.Eq{"public": true},
+			},
+		}).
+		OrderBy("public ASC", "team_id ASC", "ordering ASC").
```
clarafu: For the ordering, I'm not sure if this will result in the teams that you passed in to be ordered first, and then the public ones after because there can be times when the team you passed in is also public, and that will order it as part of the `public = true` portion.

You might be able to double check that the web UI relies on the API for the ordering or it might order the teams itself after it receives the response, that way we don't need to worry about the ordering.

julia-pu: From the old code it seemed that we want the results to be ordered by public first, having the non-public ones at the beginning and then public ones after. I believe the unit tests and UI tests cover that case.

pivotal-jamie-klassen: The user experience this would affect is when viewing the sidebar as a non-super-admin; I can confirm that the dashboard does its own sorting. We didn't really lay out a requirement for the ordering in the sidebar though, so the UI definitely counts on the `/api/v1/pipelines` response to order pipelines. However #4483 and #4484 have just been opened, so there's definitely interest in defining this. It's your call if/when you want to make the simplification, but the sidebar for non-super-admins would definitely be affected if you drop that `public ASC` part.

atc/db/pipeline_factory.go
xtremerui: Removed that. Good call! Thx.

atc/db/pipeline_factory.go
xtremerui: reverted the changes done here. So UI could expect the same ordering of pipelines still.

xtremerui:
@clarafu this is a pure backend optimization( but we dont have benchmark). Do you want a specific release note for it? I dont recall we put such changes in release notes.

clarafu:
@xtremerui I think it would be good to put in how it might affect the users, like for example, this fix will help requests from the API be completed in one transaction, so as a result concourse deployments that are under high load will see that their fly commands or web UI will be loading faster because it completes the first request before trying to complete the next ones.

release-notes/latest.md
```diff
@@ -19,6 +19,10 @@
 
 * All API payloads are now gzipped. This should help save bandwidth and make the web UI load faster. #4470
 
+#### <sub><sup><a name="4494" href="#4494">:link:</a></sup></sub> feature
+
+* Optimize certain API requests to use single transaction of DB, instead of making DB connection for each sub-query. #4494
```
clarafu:
```suggestion
* When an API request is sent from the Web UI and there is no response, it sends another request every 5 seconds. This can clobber the database with a ton of the same requests and cause the first request to be delayed further because of the competition to grab a connection. This PR optimizes all API requests to use single transaction talking to the database, which will result in the queries for each request to be completed together.  #4494
```

vito: @clarafu Nothing wrong with adding too much info in release notes; I'd rather over-communicate than under-communicate as contributors will be reading these, too.

Suggesting a different wording though:

```suggestion
* API endpoints have been changed to use a single transaction per request, so that they become "all or nothing" instead of holding data in memory while waiting for another connection from the pool. This could lead to snowballing and increased memory usage as requests from the web UI (polling every 5 seconds) piled up. #4494
```

PR #4501, opened 2019-09-25
keep track of checks

pivotal-jwinters:
so we don't create more and more goroutines all doing the same thing

### DIFF
--- a/atc/lidar/checker.go
+++ b/atc/lidar/checker.go
@@ -2,6 +2,7 @@ package lidar
 
 import (
 	"context"
+	"sync"
 
 	"code.cloudfoundry.org/lager"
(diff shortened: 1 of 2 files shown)


atc/lidar/checker.go
```diff
@@ -25,6 +27,8 @@ type checker struct {
 
 	checkFactory db.CheckFactory
 	engine       engine.Engine
+
+	running *sync.Map
```
xtremerui: what is the memory impact for using this map? Is it possible that same check got picked up by different ATCs that creates overheads in memory?

pivotal-jwinters: This change is actually to limit the amount of memory used by each ATC by making sure we don't run duplicate checks.

PR #4503, opened 2019-09-26
Fixed the flake 'fly workers' test.

### LINKED ISSUE #4497: fly: `fly workers` integration test time flake
It is possible for enough time to pass between [this initialization](https://github.com/concourse/concourse/blob/master/fly/integration/workers_test.go#L141-L147) and [this assertion](https://github.com/concourse/concourse/blob/master/fly/integration/workers_test.go#L167-L173) that the assertion fails. [Proof](https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/unit/builds/1098#L5d8a5acd:69:92)

The Right Way™ to solve this problem is to refactor fly to use a fake clock that always returns the same time. However, the fly integration tests use [gexec.Build](https://github.com/concourse/concourse/blob/master/fly/integration/suite_test.go#L32) to compile fly and exercise it by shelling out to the compiled binary.

Instead of an 'integration' test like the above, we should be verifying the 'age' column works appropriately by externalizing our 'clock' dependency [here](https://github.com/concourse/concourse/blob/master/fly/commands/workers.go#L205), and then injecting a fake for tests. I'm thinking we should externalize the [stdout writer](https://github.com/concourse/concourse/blob/master/fly/commands/workers.go#L69) too.

evanchaoli:
Resolve #4497

### DIFF
--- a/fly/integration/workers_test.go
+++ b/fly/integration/workers_test.go
@@ -140,10 +140,10 @@ var _ = Describe("Fly CLI", func() {
 			BeforeEach(func() {
 				worker1StartTime = time.Now().Unix() - 2*day - 90*second
 				worker2StartTime = time.Now().Unix() - 1*day - 90*second
-				worker3StartTime = time.Now().Unix() - 10*hour - 3*minute - 5*second
-				worker4StartTime = time.Now().Unix() - 8*hour - 30*minute - 5*second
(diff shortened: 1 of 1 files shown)


evanchaoli:
@pivotal-jamie-klassen Please take a look this PR. With this change, I have run many times of fly integration tests on my local and never hit any failure.

PR #4507, opened 2019-09-26
Set duration to n/a if startTime is n/a

### LINKED ISSUE #4472: fly shows the wrong duration for cancelled builds
Looks like its computing time since the epoch? Probably fly is using some kind of integer type to represent those timestamps, so if the start time is not set, it defaults to "0 milliseconds since the epoch".

Screen Shot 2019-09-20 at 11 07 07 AM

I think it would make more sense to see `n/a` in the duration field, or even `cancelled` in the duration field.

iamjarvo:
Fixes #4472

Screen Shot 2019-09-26 at 2 05 43 PM

### DIFF
--- a/fly/commands/builds.go
+++ b/fly/commands/builds.go
@@ -254,7 +254,7 @@ func populateTimeCells(startTime time.Time, endTime time.Time) (ui.TableCell, ui
 		durationCell.Contents = endTime.Sub(startTime).String()
 	}
 
-	if startTime == zeroTime && endTime == zeroTime {
+	if startTime == zeroTime {
 		durationCell.Contents = "n/a"
 	}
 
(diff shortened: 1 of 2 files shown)


fly/commands/builds.go
```diff
@@ -254,7 +254,7 @@ func populateTimeCells(startTime time.Time, endTime time.Time) (ui.TableCell, ui
 		durationCell.Contents = endTime.Sub(startTime).String()
 	}
 
-	if startTime == zeroTime && endTime == zeroTime {
+	if startTime == zeroTime && endTime == zeroTime || startTime == zeroTime {
```
pivotal-jamie-klassen: glad that you found this logic! but I think `(A && B) || A == A` (by set theory, or by doing the truth table), so you can just say `startTime == zeroTime` here!

pivotal-jamie-klassen:
Also, the DCO bot will complain if your commit does not have a `Signed-off-by` line that matches the commit's author. This is part of our contributor agreement.

iamjarvo:
Thanks for the feedback, I will address your feedback

iamjarvo:
@pivotal-jamie-klassen made the changes you requested

pivotal-jamie-klassen:
release note in #4565

Issue #3985, opened 2019-06-05
Restrict allowed character sets for names in API (pipelines, jobs, resources, steps, etc)


vito:
## What challenge are you facing?

Tons of issues are opened all the time because various funny characters cause breakage in various places (e.g. `fly`, web UI):

* #2336
* #1892 
* #3941 
* #3787 
* #2211
* #3724
* ...probably like 10 others somewhere I can't find

We've never had any validation for these so folks have started to use all kinds of weird naming schemes which are difficult to support going forward, especially in contexts like `fly` where we often have our own notation in flags.

## What would make this better?

We should just decide on allowable characters and perform validation at pipeline submission time.

If I were to start Concourse over I would have restricted everything to `[a-z0-9-]+` and force everyone to use fairly simple job/pipeline names. I say `a-z` but I really mean alphabetical in whatever language the user speaks - I don't remember the regex notation for that.

I would even go as far as only allowing lowercase characters so we don't have to care about case-sensitivity, and allowing `-` but not `_` so we don't end up with ideological wars.

The advantage of having such a strict set of allowable characters would be that everyone's Concourse feels the same. This also allows us to use notation more freely, e.g. `/` for pipeline hierarchies (https://github.com/concourse/concourse/issues/532).

Are there any use cases out there for certain special characters in job/pipeline names?

How many people have I made angry with my recklessly narrow definition of allowed characters?

pn-santos:
At my company we've already sort of have the convention around using `a-z0-9-` names for pipelines, tasks, jobs and resource names. Although sometimes you see `A-Z` and `_` here and there but it's the minority.

Given that most of these names end up in a URI it makes it much nicer if they are URI "safe" and follow standard best-practices around URI naming schemes (lowercase everywhere, `-` instead of `_` and avoid all characters that would require percent encoding)

hfinucane:
We have a lot of long job names- "test cross-region VPC peering", "deploy to dev", that sort of thing. We could change them to work with no spaces and only dashes, but it might be annoying.

We have something else that's a little hairier- we have our own implementation of spaces using ERB, and for PRs we iterate over the task files in the 'ci' directory of the branch, and then create standalone jobs for each. Currently we're doing something like s/\//::/ to sanitize the names, but that would get less tenable if we were locked down to `a-zA-Z0-9-`.

ukabu:
[...]

PR #4508, opened 2019-09-26
Restrict names by [0-9a-z-] when saving pipeline config 

xtremerui:
names of pipeline, group, resource, resource type, job, get plan, put plan
and task plan are restricted to lower case alphanumeric and '-'

implements https://github.com/concourse/concourse/issues/3985

### DIFF
--- a/atc/api/configserver/save.go
+++ b/atc/api/configserver/save.go
@@ -84,15 +84,15 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	warnings, errorMessages := config.Validate()
+	pipelineName := rata.Param(r, "pipeline_name")
+	teamName := rata.Param(r, "team_name")
+
+	warnings, errorMessages := config.Validate(pipelineName)
 	if len(errorMessages) > 0 {
 		session.Info("ignoring-invalid-config")
 		s.handleBadRequest(w, errorMessages...)
 		return
 	}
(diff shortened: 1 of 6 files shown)


robwhitby:
Hi, 

I'm not in favour of this change for a couple of reasons. 

We have over 1000 pipelines that will have to be updated - which is a pain in itself, but after doing that a lot of jobs will lose their build history because the name has changed. 

Also our users like to use human readable names for jobs and especially tasks. e.g. `Deploy to Dev (onprem)`, `Run Tests` etc. etc. It helps make the pipeline understandable in the UI.

Maybe the restriction could be made a warning instead of an error? or the set of allowed characters expanded?

Thanks

xtremerui:
@robwhitby do you mind also add your comment to the original issue so folks can discuss more about it there? Certainly there are others who will share you thoughts as well. We could find a feasible way.

Thx!

clarafu:
This PR is basically only going to restrict users from setting new or existing pipelines from using anything other than lower case alphanumeric and dashes but it doesn't deal with existing untouched pipelines. @vito Are we going to strategize for how we are going to deal with converting existing pipelines with illegal names within the hierarchical pipelines issue?

cirocosta:
https://github.com/concourse/rfcs/pull/40

vito:
Let's put this down until concourse/rfcs#40 is resolved. 👍

PR #4516, opened 2019-09-27
add timeout to bclient for worker volumes gc

kcmannem:

### DIFF
--- a/worker/workercmd/worker.go
+++ b/worker/workercmd/worker.go
@@ -127,6 +127,10 @@ func (cmd *WorkerCommand) Runner(args []string) (ifrit.Runner, error) {
 				// sweeping too much
 				ResponseHeaderTimeout: 1 * time.Minute,
 			},
+			// we've seen destroy calls to baggageclaim hang and lock gc
+			// gc is periodic so we don't need to retry here, we can rely
+			// on the next sweeper tick.
+			Timeout: 5*time.Minute,
 		},
 	)
 


kcmannem:
I couldn't find an issue for this, if there was one.

worker/workercmd/worker.go
```diff
@@ -127,6 +127,10 @@ func (cmd *WorkerCommand) Runner(args []string) (ifrit.Runner, error) {
 				// sweeping too much
 				ResponseHeaderTimeout: 1 * time.Minute,
 			},
+			// we've seen destroy calls to baggageclaim hang and lock gc
+			// gc is periodic so we don't need to retry here, we can rely
+			// on the next sweeper tick.
+			Timeout: 5*time.Minute,
```
cirocosta: good point on not needing to retry 😁  

is there a particular reason for going with 5m though? Is that a common timeout that we use in other places? e.g., do we use timeouts at the transport level for gdn too?

Also, I'm curious if we should go with timeouts at the transport level, or standardizing on cancelling these at a higher level (e.g., using context cancellation) like we've been doing in other places 🤔  

thanks!

kcmannem: yeah we do, https://github.com/concourse/concourse/blob/master/atc/worker/gclient/factory.go#L54 . This is defaulted to 5mins. We don't really have arguments to pick 5 mins, it was just a hunch. 
Funny enough we never added timeouts for the bc client on the atc https://github.com/concourse/concourse/blob/master/atc/worker/db_worker_provider.go#L185 . Maybe we should reflect the changes there as well.

We've never really done context based timeouts for the garden clients as their interfaces don't provide context as an argument. We've actually imported the entire garden client source to add context for log streaming. We also added context for volume streaming. Looking back we should have added it to the entire interface.

Maybe we can make this a longer term goal to switch to context based cancelation for baggage claim as well since containerd does this as well.

deniseyu: I'd agree that while there are ways that we can "improve" the Garden client (aka fake-forking more of their code directly into Concourse 🙉) it might not be worthwhile since containerd will come with a client that's aware of `context`.

pivotal-jamie-klassen:
release note in #4560

PR #4518, opened 2019-09-29
feat: avoid to renew Vault lease every second if Vault server doesn't accept renew.

evanchaoli:
When tests with Vault dev server, renew lease is not allowed, thus ATC log is massed
with Vault renew error like:

    web_1     | {"timestamp":"2019-09-29T14:07:18.982725900Z","level":"error","source":"atc","message":"atc.credential-manager.renew.failed","data":{"error":"Error making API request.\n\nURL: PUT http://192.168.1.11:8200/v1/auth/token/renew-self\nCode: 400. Errors:\n\n* lease is not renewable","name":"vault","session":"5.46"}}

This commit will skip renew once detects such error.

### DIFF
--- a/atc/creds/vault/api_client.go
+++ b/atc/creds/vault/api_client.go
@@ -3,6 +3,7 @@ package vault
 import (
 	"fmt"
 	"path"
+	"strings"
 	"sync/atomic"
 	"time"
 
@@ -20,6 +21,8 @@ type APIClient struct {
 	authConfig AuthConfig
 
 	clientValue *atomic.Value
+
+	renewable bool
 }
 
 // NewAPIClient with the associated authorization config and underlying vault client.
@@ -32,6 +35,8 @@ func NewAPIClient(logger lager.Logger, apiURL string, tlsConfig *vaultapi.TLSCon
 		authConfig: authConfig,
 
 		clientValue: &atomic.Value{},
+
+		renewable: true,
 	}
 
 	client, err := ac.baseClient()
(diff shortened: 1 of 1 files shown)


xtremerui:
Do we need to consider a case when Vault allows lease later? I assume ATC would need to restart.

evanchaoli:
@xtremerui I don't think a Vault server would swing between need and not-need renew lease. This feature is most likely for dev Vault server only during develop Vault related features. In a production environment, it feature should not take effective as it works based on Vault returned error message.

evanchaoli:
@xtremerui Do you have further comment/concern?

xtremerui:
I agree that a Vault server wont be on/off with lease regularly. And a production Vault probably enables lease anyway (since we haven't seen one complain about that error). 

So to your point that it is mainly for dev, I don't feel strongly need of this change. You could have some other ways to filter out the log messages like pipe the log to a filter maybe?

evanchaoli:
@xtremerui I think this change is at least innocent. Yes we can filter log with many ways, but that all add overhead when run local tests. And logically, if Vault server returns an error of "lease is not renewable", client should not try to renew lease at all, right?

deniseyu:
I was wondering if there were any counterexamples to the implicit assumption that "if this error message is logged, then the secret is never renewable". From a quick search inside Vault's codebase of the string `lease is not renewable`, it looks like there are three conditions that could return this error message: https://github.com/hashicorp/vault/blob/6310462550443225ee511919dfa76d1f2e80672d/vault/expiration.go including insufficient user permissions. I don't have super strong context around Vault's internals so I don't know what's the likelihood of false positives being returned for each of these cases.

I think this PR is okay to merge because agreed that the risk seems pretty low. That being said I'd be careful of a one-way operation that flips the renewability to false, and using something configurable might be a safer way to implement this behavior. It looks like we already support making the retry interval configurable with the flag `CONCOURSE_VAULT_RETRY_INITIAL` which defaults to 1s. 

In the future, if this produces surprising behavior, maybe we could read that value and if it's set to 0, or a negative value that makes no sense, then the `Renew` loop only runs once? This would keep the onus on the operator to configure their dev environment explicitly.

evanchaoli:
@deniseyu Thanks a lot.

pivotal-jamie-klassen:
release notes in #4560

Issue #4337, opened 2019-09-02
UX/Runtime: Tell me what's happening when a step is spinning


kcmannem:
It's hard to really know what Concourse is doing when it shows you a loading spinner. This is because Concourse is doing multiple things in preparation to run the step. At a high level the loading spinner can represent any stage in [Picking a worker, Making a container, Mounting inputs, Attaching, etc...]. This is really long process that signals false positives to the users that something maybe wrong or give the impression that Concourse is slow.

In many products, when you see a loading spinner you can make a assumption for what its trying to do. The loading spinner only indicates one thing. If youtube is showing a loading spinner you can assume that its waiting for slices of video to come in. If you've scrolled to the bottom of your newsfeed and see a loading spinner, you know that more news posts are being fetched. We risk making the loading spinner even more ambiguous when the Kubernetes runtime comes in as the preparation stages there will be completely different.

Can we raise the visibility and provide feedback to users of what's happening?

The simplest approach is to write out the important Concourse debug logs (format them a bit so they're not json) to the StdOut/Err build event stream. This means they'll show up in User's step logs like how image fetching logs are shown. Or we can be a bit fancier and show something like this.
![Screen Shot 2019-09-01 at 7 43 40 PM](https://user-images.githubusercontent.com/9054726/64083942-ea356d80-ccf3-11e9-9473-9d0fa101c1f3.png)
This will prob take a bit more effort.

kcmannem:
cc @Lindsayauchin

maxknee:
#1891 I think is similar

Lindsayauchin:
@kcmannem This is a great idea! 

Looks like there is good user validation that this is needed system feedback for users.

What do you think of using an inline style with just a font colour update instead of using the tooltip UI style?

The rational is to keep the tooltip style as a label whereas I would view this as visibility of system feedback justifying a new style different from the tooltip.

@pivotal-jamie-klassen @jomsie  thoughts? 

![Screen Shot 2019-09-10 at 3 29 47 PM](https://user-images.githubusercontent.com/27746614/64644083-d965b580-d3df-11e9-925a-9cc11bedde40.png)

![Screen Shot 2019-09-10 at 4 52 22 PM](https://user-images.githubusercontent.com/27746614/64649685-62361e80-d3eb-11e9-8dda-5aa0f5221a2b.png)

kcmannem:
@Lindsayauchin I'm cool with this.

vito:
[...]

PR #4519, opened 2019-09-30
feat: show worker name, as well as platform, teams, tags, on GUI for put/get/task steps.

### LINKED ISSUE #4517: View Worker Job Ran On

Issues troubleshooting intermittent problems. We have a number of job intermittently failing and we would like to determine if this is when the jobs run on a specific worker / set of workers.

The ability to see where the job run will help us narrow down the root cause of our issues. Be that but ruling some out or by confirming some suspicions   

Within the job run output , display the worker name, ip and any other relevant data.

I would if I could, but I am not a programmer / developer.

evanchaoli:
Resolve #4517

Screen shot of my test result:

show-workers

### DIFF
--- a/atc/exec/get_step.go
+++ b/atc/exec/get_step.go
@@ -6,11 +6,11 @@ import (
 	"fmt"
 	"io"
 
-	"github.com/hashicorp/go-multierror"
-
 	"code.cloudfoundry.org/lager"
 	"code.cloudfoundry.org/lager/lagerctx"
 	"github.com/DataDog/zstd"
+	"github.com/hashicorp/go-multierror"
+
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/creds"
 	"github.com/concourse/concourse/atc/db"
@@ -189,6 +189,7 @@ func (step *GetStep) Run(ctx context.Context, state RunState) error {
 	if err != nil {
 		return err
 	}
+	step.delegate.Stderr().Write([]byte(fmt.Sprintf("Chosen worker %s: %s\n", chosenWorker.Name(), chosenWorker.Description())))
 
 	step.delegate.Starting(logger)
 
--- a/atc/exec/get_step_test.go
+++ b/atc/exec/get_step_test.go
@@ -2,6 +2,7 @@ package exec_test
 
 import (
 	"archive/tar"
+	"bytes"
 	"context"
 	"crypto/sha256"
 	"errors"
@@ -42,6 +43,8 @@ var _ = Describe("GetStep", func() {
 		fakeResourceCacheFactory *dbfakes.FakeResourceCacheFactory
 		fakeDelegate             *execfakes.FakeGetDelegate
 		getPlan                  *atc.GetPlan
+		stdoutBuf                *bytes.Buffer
+		stderrBuf                *bytes.Buffer
 
 		fakeVersionedSource       *resourcefakes.FakeVersionedSource
 		interpolatedResourceTypes atc.VersionedResourceTypes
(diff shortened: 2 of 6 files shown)


atc/exec/get_step.go
```diff
@@ -189,6 +189,7 @@ func (step *GetStep) Run(ctx context.Context, state RunState) error {
 	if err != nil {
 		return err
 	}
+	step.delegate.Stderr().Write([]byte(fmt.Sprintf("Chosen worker %s: %s\n", chosenWorker.Name(), chosenWorker.Description())))
```
xtremerui: what is the reasoning behind using `stderr` vs `stdout` here? Thx.

evanchaoli: There are some discussions in PR #4398 between @vito and me. I believe Concourse generated logs (sys log) should go to stderr. Going forward, @vito is thinking to add a separate log event channel to print Concourse generated logs.

xtremerui:
While I do like showing worker info for steps, I would like to put it into a more structure way like creating an event type for it or put the worker info into `StartGet` \ `StartPut` \ `StartTask`, so we could handle the output nicely in Web UI / Fly based on the event type. Just my 2 cents since it is more about runtime really @ddadlani .

kcmannem:
@xtremerui I agree, UX was planning to take on https://github.com/concourse/concourse/issues/4337 and we decided to keep stderr/out clean and structure the data in the db. I don't think we should be writing this kind of info there especially since we're currently working to integrate multiple runtime backends (Kubernetes). We need a generic way to show runtime metadata.

evanchaoli:
@xtremerui Current `StartGet` \ `StartPut` \ `StartTask` don't have a field to hold more info like this. Even if we add a field to them, it would also require UI side change to show such info. In future, some more fields, like `chosenWokrer`, could be added, but here I think what I'm adding is just a log line that shows on which workers a step will run.

evanchaoli:
@kcmannem I guess it would take time to get features you mentioned done. Once those are out, worker info is shown with some way else, we can revert this change. But this PR will provide a quick way for OPs people to see worker info, which would be super helpful.

kcmannem:
@evanchaoli I believe the UX team might be working on this next. Its high on their list of todos. This information is available to query from the db if needed. Users probably won't care about seeing every tasks runtime info but only care when concourse breaks. I don't want to optimize for second scenario when there is a workaround available.

evanchaoli:
@kcmannem  Looks like #4337 is going to show initialization info with a tooltip, so I want to confirm that, as tooltip will be refreshed as initialization going forward, will all information shown by tooltip will also print to build logs?

Also, @vito, what do you think about printing chosen worker to build log?

kcmannem:
@evanchaoli the point of #4337 is to better structure this kind of info in the db. It makes it available for both fly and the webpage to take advantage and display this UI in a more appropriate manner. I understand this is something a lot of users want (including ourselves), I've only pushed back since its something we're tackling very soon.

vito:
@evanchaoli I don't really like the idea of printing logs into the task's output; this should be done with progress or status indicators. I would rather everyone focus on #4337 (whether we get to it or someone in the community kicks it off), as we've already got the ball rolling in terms of design and strategy. We could maybe even re-use this pattern in order to split out the image fetching logs as a separate sub-task with its own progress. At that point we'll be in a world where `stdout`/`stderr` are purely what the task printed and we've got much better progress indicators for users.

In general we try to avoid "quick fix" bugfixes or "quick and dirty" features if there's already a plan for a better design in sight. I would rather work towards the end goal, because if we adopt quick-and-dirty solutions it can easily result in de-prioritization of the better version because the need is diminished.

evanchaoli:
OK, then I give up this PR.

Issue #1234, opened 2017-05-31
Movements applied to a pipeline view is applied to all pipelines.


benbotwin:
# Bug Report

* Concourse version: 3.1.0
* Deployment type: BOSH
* Infrastructure/IaaS: OpenStack
* Browser: Chrome
* Did this used to work? yes in 3.0.1

If you click and drag to move a pipeline around, the movement you applied doesn’t reset when you change the pipeline you’re looking at. The movement applied to the view of one pipeline is applied to all of them.

If I zoom way out of a pipeline and move it just a little bit I'm unable to find any other pipeline when I change views because they have moved just as much, but their view hasn't zoomed out to match. I need to refresh to see any of the pipelines after that because it's faster than dragging the view around looking for them.

cfryanr:
This problem also exists when switching between the tabs at the top of the UI which represent groups.

jama-pivotal:
This should be coming soon! we just closed out #1425 ! Thank you for your patience

PR #4539, opened 2019-10-02
Adds move command to concourse

### LINKED ISSUE #1164: As a Concourse operator with many pipelines, I would like to move a pipeline from the "main" team to a new team
# Feature Request

We have a concourse installation with many pipelines, most of which, our "main" team is not really responsible for. We've given a lot of Github Org/Teams access to our concourse, but we are ready to take advantage of concourse teams. We can create the team, but the existing pipelines would have to be deleted and re-created with the new team, losing all our precious build history.

## A Modest Proposal

Maybe, `fly set-pipeline-team -p some-pipeline -n some-team`

betarelease:
[#1164](https://github.com/concourse/concourse/issues/1164)

> Hi there! Thanks for submitting a pull request to Concourse! 
> If you haven't already, please take a look at our [Code of Conduct](https://github.com/concourse/concourse/blob/master/CODE_OF_CONDUCT.md) and our [Contributing Guide](https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md).
> 
> To help us review your PR, please fill in the following information.
> Feel free to delete any sections not applicable to your pull request.

# Existing Issue

Fixes #1164  .

# Why do we need this PR?
> No issue number? Please give us a few lines of context describing the need for this PR.

> What are the changes included in this PR? Feel free to add as many lines as needed below.

* Adding a move command to concourse/fly
* 
*

# Contributor Checklist
> Are the following items included as part of this PR?
- [x ] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/pipeline.go
+++ b/atc/db/pipeline.go
@@ -653,6 +653,40 @@ func (p *pipeline) Rename(name string) error {
 	return err
 }
 
+func (p *pipeline) Move(sourceTeamName, destinationTeamName string) error {
+	destinationTeamID, err := psql.Select("id").
+		From("teams").
+		Where(sq.Eq{
+			"name": destinationTeamName,
+		}).
+		RunWith(p.conn).Query()
+
+	_, err = psql.Update("pipelines").
+		Set("team_id", destinationTeamID).
+		Where(sq.Eq{
+			"id": p.id,
+		}).
+		RunWith(p.conn).
+		Exec()
+
+	if err != nil {
+		return err
+	}
+
+	_, err = psql.Update("builds").
+		Set("team_id", destinationTeamID).
+		Where(sq.Eq{
+			"id": p.id,
+		}).
+		RunWith(p.conn).
+		Exec()
+
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
 func (p *pipeline) Destroy() error {
 	_, err := psql.Delete("pipelines").
 		Where(sq.Eq{
--- a/atc/db/pipeline.go
+++ b/atc/db/pipeline.go
@@ -655,19 +655,29 @@ func (p *pipeline) Rename(name string) error {
 }
 
 func (p *pipeline) Move(destinationTeamName string) error {
+	tx, err := p.conn.Begin()
+	if err != nil {
+		return err
+	}
+
+	defer Rollback(tx)
+
 	destinationTeamID, err := psql.Select("id").
 		From("teams").
 		Where(sq.Eq{
 			"name": destinationTeamName,
 		}).
-		RunWith(p.conn).Query()
+		RunWith(tx).Exec()
+	if err != nil {
+		return err
+	}
 
 	_, err = psql.Update("pipelines").
 		Set("team_id", destinationTeamID).
 		Where(sq.Eq{
 			"id": p.id,
 		}).
-		RunWith(p.conn).
+		RunWith(tx).
 		Exec()
 
 	if err != nil {
@@ -679,13 +689,15 @@ func (p *pipeline) Move(destinationTeamName string) error {
 		Where(sq.Eq{
 			"id": p.id,
 		}).
-		RunWith(p.conn).
+		RunWith(tx).
 		Exec()
 
 	if err != nil {
 		return err
 	}
-	return nil
+
+	err = tx.Commit()
+	return err
 }
 
 func (p *pipeline) Destroy() error {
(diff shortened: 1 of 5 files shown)


betarelease:
Can you please review what I have so far. This is my WIP. I would like to know if this is going in the right direction and how you run tests locally so that I can run and verify. Thanks.

betarelease:
Updated the PR with the suggested way to check authorization. Also found a couple of other things wrongly wired. Please review.

atc/db/pipeline.go
```diff
@@ -653,6 +654,40 @@ func (p *pipeline) Rename(name string) error {
 	return err
 }
 
+func (p *pipeline) Move(destinationTeamName string) error {
+	destinationTeamID, err := psql.Select("id").
+		From("teams").
+		Where(sq.Eq{
+			"name": destinationTeamName,
+		}).
+		RunWith(p.conn).Query()
```
clarafu: These three queries should probably be done in one transaction in order to ensure that they are all committed together.

atc/db/pipeline.go
```diff
@@ -653,6 +654,40 @@ func (p *pipeline) Rename(name string) error {
...
+			"id": p.id,
+		}).
+		RunWith(p.conn).
+		Exec()
+
+	if err != nil {
+		return err
+	}
+
+	_, err = psql.Update("builds").
```
clarafu: I'm a little worried about this query because it will be updating all the builds within a pipeline, which can result in thousands or millions of rows being updated. Maybe it should be done in batches? Not too sure yet tho.. 🤔 I might want to do some bench marking to see how long it will take!

atc/db/pipeline.go
```diff
@@ -653,6 +654,40 @@ func (p *pipeline) Rename(name string) error {
...
+		Set("team_id", destinationTeamID).
+		Where(sq.Eq{
+			"id": p.id,
```
clarafu: This should be `"pipeline_id"` instead of `"id"`. The `"id"` represents a build id because you are updating the builds table.

atc/db/pipeline.go
```diff
@@ -653,6 +654,40 @@ func (p *pipeline) Rename(name string) error {
 	return err
 }
 
+func (p *pipeline) Move(destinationTeamName string) error {
+	destinationTeamID, err := psql.Select("id").
```
clarafu: We should be checking for if an error occurred here!

atc/db/pipeline.go
clarafu: `Query()` can return no rows without throwing an error so maybe we should use `QueryRow()` instead and then use `Scan(&destinationTeamID)` to grab out the team id. We should always be expecting a team id to return from this query so we should be erroring if there is no rows being returned here.

atc/db/pipeline.go
betarelease: Is it enough that we do a count check before making this update and abort with an error message if the count is more than 100 or something? That way we will know if this feature is in fact used by people with pipelines that have thousands of rows. If not atleast there will be some protection.

atc/db/pipeline.go
betarelease: Updated.

atc/db/pipeline.go
betarelease: done

atc/db/pipeline.go
betarelease: also added the transaction logic.

clarafu: If we do a count then abort, that would mean users that have more builds than our limit that we set wouldn't be able to use this `move` command. I think most of Concourse users have pipelines that have more than 100 or even 1000 builds 🤔 so I was more thinking of doing query in batches of 1000 builds. Doing it in batches will make it so that it doesn't lock up all the build rows while it's trying to run the update.

atc/db/pipeline.go
clarafu: Oh I actually meant the left side of this `sq.Eq`, it should be `"pipeline_id"` rather than `"id"` because right now it's going to only update one row from the builds table that has a build `"id"` of your `p.id` (pipeline id) value.

pivotal-izabela-gomes:
Hey @betarelease, I know it's been a long time, sorry about that. Did you get a chance to review the suggestions made up there? Just wondering if you are still interested in keep working on this. Thanks!

vito:
closing due to inactivity, but if you find the time again feel free to resubmit! thanks again.

PR #4544, opened 2019-10-02
Updating ./worker/README.md to give more information about the worker

### LINKED ISSUE #2860: Update worker README with information about its dependencies

A newcomer looking at the `./worker` package might not quickly realize where they should look for the actual implementations of either the baggageclaim or garden API.

e.g., we had under `concourse-bosh-release`:

```
* [Garden](https://github.com/cloudfoundry-incubator/garden) is a generic
  interface for orchestrating containers remotely on a worker
* [Baggageclaim](https://github.com/concourse/baggageclaim) is a server for
  managing caches and artifacts on the workers
```

although it makes more sense to have it under `./worker`.

Include in `./worker/README.md` a description of each of the possible implementations that are usually used as implementations of the Garden API (e.g., guardian and houdini), as well as a reference to Baggageclaim. 

yess 🙌

ctorboh15:
Fixes #2860

Updated the ./worker/README.md to be more descriptive of the worker module.

### DIFF
--- a/worker/README.md
+++ b/worker/README.md
@@ -14,10 +14,16 @@
   ## about
 
   *worker* is the workhorse of Concourse. It's responsible for the creation
-  and deletion of the containers in which the pipeline tasks are executed.
+  and deletion of the containers in which the pipeline operations
+  (get, check, put, task) are executed.
 
-  A worker node registers with the web node and is then used for executing
-  builds and performing resource checks. It doesn't really decide much on its own.
+  A worker node registers with the web node(s) and is then used for executing
+  builds and performing resource checks.
+
+  The ATC component in the web node(s) decides how to allocate containers
+  to workers that have been registered in the pool, using the configured
+  container-placement-strategy. It also manages the container deletion on the workers
+  via database calls and the Garden API on the workers.
 
   A worker node runs the following 2 GO API's
 
--- a/worker/README.md
+++ b/worker/README.md
@@ -1,5 +1,33 @@
 # worker
 
+*worker - container orchestrator/ worker artifact management/ build executor*
+
 ![](https://upload.wikimedia.org/wikipedia/commons/thumb/b/b8/Otto_Lilienthal_gliding_experiment_ppmsca.02546.jpg/640px-Otto_Lilienthal_gliding_experiment_ppmsca.02546.jpg)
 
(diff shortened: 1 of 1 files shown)


worker/README.md
```diff
@@ -1,5 +1,33 @@
...
+
+
+
+  ## about
+
+  *worker* is the workhorse of Concourse. It's responsible for the creation
+  and deletion of the containers in which the pipeline tasks are executed.
+
+  A worker node registers with the web node and is then used for executing
+  builds and performing resource checks. It doesn't really decide much on its own.
```
deniseyu: We can elaborate a bit more on `It doesn't really decide much on its own` - as a developer reading this, I'd be asking myself the question, "if the worker doesn't make decisions, then who does?" and the answer is that all container allocation logic happens in the ATC.

Perhaps one way to add a bit more detail is to say instead, "...performing resource checks. The ATC component in the web node(s) decides how to allocate containers to workers that have been registered in the pool, using the configured `container-placement-strategy`".

ctorboh15: I like those suggestions 👍 Should we also mention that ATC also handles container deletion as well? Re-reading this it sounds as though that the worker take care of container cleanup themselves when its more a GC step? Curious on your thoughts

deniseyu: Sure, happy for you to include mention of container deletion, that feels like a logical thing to include.

For more context (the below is probably too much detail to include in this level of documentation), it's a bit confusing because we're not as precise as we could be when we talk about "containers" and their lifecycle. You're absolutely right that the ATC initiates container deletion in the sense that the ATC will do the following things:

1. add/update a container record in its database as "creating"/"deleting"
2. send an API call along to the container's worker's Garden server asking, "please create/delete this container!"
3. update container record to "created", or delete the row in the case of deletion

But on the worker, the Garden server is the thing that actually gets rid of the container as a filesystem entity, when it hears the request from the ATC. So _technically_.. the last mile of deletion and creation is performed by something on the worker, but that's kind of an implementation detail that we shouldn't worry about too much when we're describing the responsibilities of components in the system. (Apologies if you already knew this context - I couldn't find a clear explanation anywhere in our [existing written documentation](https://concourse-ci.org/container-internals.html) so I figured I'd type it out!)

Issue #3600, opened 2019-03-27
GC Containers and volumes when present in the worker but not in the DB


cirocosta:
Hey,

Once we got https://github.com/concourse/concourse/issues/2588 in, we added support for GCing when there's a mismatch between what is in the worker and what is in the DB when

- DB has containers/volumes that a worker doesn't have,

But we **don't** do the other "side" of the diff:

- a worker having containers/volumes that the DB doesn't have.

This means that if `worker-0`, for some reason, has volumes that the DB doesn't have, those volumes will never be cleaned up unless someone manually discovers that diff and performs the removals.

 A scenario where this could occur:

- the user takes backup of the DB at `t0`,
- time goes on ... the worker has some volumes at `t1`,
- DB gets restored to `t0`,
- the worker will forever have the extra volumes there.

Do you think this could have repercussions in the way we create containers/volumes? E.g., in the case of ephemeral check containers?

Thanks!

vito:
Yeah, we'd have to be careful not to remove ephemeral or health-check containers/volumes. Maybe this could be done by marking the containers as ephemeral/health-check via properties?

cirocosta:
Hey,

Another that @YoussB and I were discussing:

- worker launched with `--ephemeral` starts
- containers and volumes are created
- network problems happen, worker "stalls", thus, gets removed from the db
- worker comes back, but its old volumes are never deleted

Does that make sense?

Thanks!

cirocosta:
(related: https://github.com/concourse/concourse/issues/2835)

YoussB:
@ddadlani and I saw the same issue when working with a customer that caused a big problem since most of the workers got overloaded by containers after an upgrade.

pivotal-jamil-shamy:
> @ddadlani and I saw the same issue when working with a customer that caused a big problem since most of the workers got overloaded by containers after an upgrade.

@ddadlani @YoussB @cirocosta 

To add more context for reference:
1. We were running concourse 5.2.1.
1. The average container count per worker before the upgrade was 100
1. We upgraded to concourse 5.5.1
1. We noticed that the container count per workers started to climb sharply.
1. After a few hours of finishing the upgrade, all the workers' container count was maxed out at 256.
1. No more builds were able to be scheduled.
1. After investigation, we found that there are many containers on the workers were not "known" by the DB, i.e. they were orphaned.
1. The only possible solution we were left with is to recreate ALL the workers VMs.

pivotal-jamil-shamy:
[...]

PR #4547, opened 2019-10-02
Skip over containers on stalled workers when deleting missing

### LINKED ISSUE #4536: Stalled workers' containers get cleaned up from ATC DB
## Bug Report

When a worker enters stalled state but has containers that were reported missing to the ATC, ATC will continue to count down the grace period and delete that container record after 5 minutes default (plus or minus 30 seconds for the GC tick). This leads to a state where a container still exists on the worker when it comes back online, but no record of the container exists on ATC anymore. ATC can continue sending traffic to that worker without knowledge of this state, and while Garden will reject additional container creation requests beyond the default limit of 250, it would be better if container capacity was synced up between the worker and ATC.

## Steps to Reproduce
1. modify a container row and insert a time value into the missing_since column.
2. turn off eth0 on the worker so this simulates a "stalled" worker
3. observe gc continue to clean up missing containers once the grace-period is up 

## Expected Results

not this. The ATC shouldn't delete container rows when a worker is in stalled state.

## Actual Results

it deletes the row

deniseyu:
* If containers from stalled workers are removed from ATC DB,
  then ATC will be out of sync with worker filesystem when
  worker returns to running state. This contributes to
  problem of mysteriously high container count on workers.

Addresses #4536

# Existing Issue

Fixes #4536.

# Why do we need this PR?

This PR solves one of a few possible edge cases that can contribute to upward-creeping container count on workers. This attempts to keep ATC DB's state closer to the filesystem reality of workers that are flipping between stalled and running state.

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/container_repository.go
+++ b/atc/db/container_repository.go
@@ -140,11 +140,19 @@ func (repository *containerRepository) FindDestroyingContainers(workerName strin
 }
 
 func (repository *containerRepository) RemoveMissingContainers(gracePeriod time.Duration) (int, error) {
+	stalledWorkers, err := stalledWorkerNames(repository.conn)
+	if err != nil {
+		return 0, err
+	}
+
 	result, err := psql.Delete("containers").
 		Where(
 			sq.And{
 				sq.Eq{
 					"state": []string{atc.ContainerStateCreated, atc.ContainerStateFailed},
+				},
+				sq.NotEq{
+					"worker_name": stalledWorkers,
 				},
 				sq.Gt{
 					"NOW() - missing_since": fmt.Sprintf("%.0f seconds", gracePeriod.Seconds()),
@@ -266,6 +274,26 @@ func (repository *containerRepository) FindOrphanedContainers() ([]CreatingConta
 	}
 
 	return creatingContainers, createdContainers, destroyingContainers, nil
+}
+
+func stalledWorkerNames(conn Conn) ([]string, error) {
+	var stalledWorkers []string
+	rows, err := psql.Select("name").Where(sq.Eq{"state": "stalled"}).From("workers").RunWith(conn).Query()
+	if err != nil {
+		return stalledWorkers, err
+	}
+	defer Close(rows)
+
+	for rows.Next() {
+		var workerName string
+		err = rows.Scan(&workerName)
+		if err != nil {
+			return []string{}, err
+		}
+
+		stalledWorkers = append(stalledWorkers, workerName)
+	}
+	return stalledWorkers, nil
 }
 
 func selectContainers(asOptional ...string) sq.SelectBuilder {
--- a/atc/db/container_repository.go
+++ b/atc/db/container_repository.go
@@ -140,23 +140,13 @@ func (repository *containerRepository) FindDestroyingContainers(workerName strin
 }
 
 func (repository *containerRepository) RemoveMissingContainers(gracePeriod time.Duration) (int, error) {
-	stalledWorkers, err := stalledWorkerNames(repository.conn)
-	if err != nil {
-		return 0, err
-	}
-
-	result, err := psql.Delete("containers").
+	result, err := psql.Delete("containers c USING workers w").
+		Where(sq.Expr("c.worker_name = w.name")).
 		Where(
 			sq.And{
-				sq.Eq{
-					"state": []string{atc.ContainerStateCreated, atc.ContainerStateFailed},
-				},
-				sq.NotEq{
-					"worker_name": stalledWorkers,
-				},
-				sq.Gt{
-					"NOW() - missing_since": fmt.Sprintf("%.0f seconds", gracePeriod.Seconds()),
-				},
+				sq.Expr(fmt.Sprintf("c.state='%s'", atc.ContainerStateCreated)),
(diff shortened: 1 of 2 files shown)


kcmannem:
lgtm, gonna do the docker test we did during our investigation tomorrow morning and accept this.

atc/db/container_repository.go
```diff
@@ -140,12 +140,20 @@ func (repository *containerRepository) FindDestroyingContainers(workerName strin
 }
 
 func (repository *containerRepository) RemoveMissingContainers(gracePeriod time.Duration) (int, error) {
+	stalledWorkers, err := stalledWorkerNames(repository.conn)
```
cirocosta: Heey,

I'm not very familiar with the code under `db`, but I think there's might a rece here given that we're performing those queries within the context of separate transactions - is this right? 

I wonder if this could/should be done in a single sql statement rather than performing a query, loading that in memory, and then using in a separate statement 🤔 

wdyt?

thx!

deniseyu: yea, I wanted to implement this as one atomic operation, but the challenge is that you can't do a left-join inside of a DELETE so unfortunately it has to happen in two steps, unless there's something I'm missing from the sparsely-documented sql library we're using...

cirocosta: oh, interesting! 

Being that the case, should we use the same transaction for that though? Is that a pattern that we use throught `db`? Or, perhaps, use subqueries (assuming that's permitted)?

thanks!!

deniseyu: yea that's a good point! we have a few other places in our codebase, like [here](https://github.com/concourse/concourse/blob/master/atc/db/check.go#L139-L175) where we do multiple stuff within one transaction so it's certainly not a brand new pattern. thanks @cirocosta for the catch 😄

atc/db/container_repository.go
```diff
@@ -140,17 +140,21 @@ func (repository *containerRepository) FindDestroyingContainers(workerName strin
 }
 
 func (repository *containerRepository) RemoveMissingContainers(gracePeriod time.Duration) (int, error) {
-	result, err := psql.Delete("containers").
+	result, err := psql.Delete("containers USING workers").
```
ddadlani: inspired by https://github.com/concourse/concourse/blob/ba5a5932f1921fd8c2d22d0246dae2af3a6abd78/atc/db/worker_task_cache.go#L97

atc/db/container_repository.go
ddadlani: we're using USING instead, inspired by https://github.com/concourse/concourse/blob/ba5a5932f1921fd8c2d22d0246dae2af3a6abd78/atc/db/worker_task_cache.go#L97

deniseyu:
this solution is much cleaner, thanks @ddadlani @kcmannem 🤘  i can't approve this since i opened it, but maybe @xtreme-sameer-vohra can!

i'm still not 100% happy with how those unit tests are written, they're just kind of confusing to read, but they can be iteratively improved later.

ddadlani:
FYI we've left the unit build pending because it's going to fail, once we push another commit fixing tests it should re-trigger

xtreme-sameer-vohra:
@deniseyu & @ddadlani 
With this fix, would a build ( & their corresponding steps ) that was running on a stalled worker, remain stuck in its previous state prior to the worker stalling indefinitely ?

Would that be desired ?

ddadlani:
@xtreme-sameer-vohra build status is not dependent on the existence of a container. The build status should update only if the process running in the container returns a status code. If a worker was stalled while a process is running, the build status would be stuck in its previous state regardless of these changes.

If the worker recovers from the stalled state, GC should kick in and ensure that worker and DB container state are synchronized. If the worker goes away, GC should clean up all the resources on that worker. This PR simply ensures that we don't make any assumptions about the state of resources on a stalled worker, since we don't have enough information.

kcmannem:
Don't merge this pr depends on #3600. I didn't want to add tests and remove tests in the span of 2 hours.

xtreme-sameer-vohra:
> @xtreme-sameer-vohra build status is not dependent on the existence of a container. The build status should update only if the process running in the container returns a status code. If a worker was stalled while a process is running, the build status would be stuck in its previous state regardless of these changes.
> 
> If the worker recovers from the stalled state, GC should kick in and ensure that worker and DB container state are synchronized. If the worker goes away, GC should clean up all the resources on that worker. This PR simply ensures that we don't make any assumptions about the state of resources on a stalled worker, since we don't have enough information.

Yep, that makes sense. Looking at the code https://github.com/concourse/concourse/blob/master/atc/worker/pool.go#L134 for a case where the web that initiated the build went away and another web took over tracking the build, `ContainerInWorker` will filter for `running` workers only.

atc/db/container_repository.go
kcmannem: Done

atc/db/container_repository.go
kcmannem: done

kcmannem:
@ddadlani just tested it manually again, it works.

PR #4548, opened 2019-10-02
fix(fly workers): showed wrong age if a worker is younger than 1 hour

### LINKED ISSUE #4545: fly workers command shows wrong age
In Concourse 5.6.0 the age column was added to the output of `fly workers`. Unfortunately this shows the wrong age, minutes are displayed as hours. @evanchaoli

evanchaoli:
Resolve #4545


evanchaoli:
@pivotal-jamie-klassen Sorry for the bug. Please take a look at this quick fix.

xtremerui:
I think the test that you removed in a previous PR is actually valid could you also put it back? The one that assert on `33h5m` (I think it should be `33m5s`)

evanchaoli:
@xtremerui I removed the test because of #4497. As `fly` runs only integration tests, for second-level test cases, as long as a test runs a little slow, it might fail.

PR #4556, opened 2019-10-03
atc: emit check finish event

### LINKED ISSUE #4551: missing lidar check metrics for non catch-all emitters
## Bug Report

Once we got Lidar in (see https://github.com/concourse/concourse/issues/3788), the metric that keeps track of the checks being performed was changed (see https://github.com/concourse/concourse/commit/e66e0302bed4edb3966c813a8f6b5da9eead5b7c).

As there were no emitter-specific changes, this means that those emitters that need specific formats (like Prometheus and NewRelic) do not get the events. 

## Steps to Reproduce

Run `concourse` with `CONCOURSE_PROMETHEUS_BIND_IP=0.0.0.0`, CONCOURSE_PROMETHEUS_BIND_PORT=9090`.

curl localhost:9391 | grep concourse_ | grep check

## Expected Results

Have a check metric.

## Actual Results

No check metric

## Additional Context

See https://github.com/concourse/concourse/blob/f580c2b1f59d95c862784521d710847830c41f1f/atc/metric/emitter/prometheus.go#L364-L367

ps.: I've been experimenting with getting rid of these per-emitter configurations that we must remember to do right now (see https://github.com/concourse/concourse/pull/4247)

## Version Info

* Concourse version: 5.6.0
* Did this used to work?: yes

xtremerui:
fixes #4551

### DIFF
--- a/atc/metric/emitter/prometheus.go
+++ b/atc/metric/emitter/prometheus.go
@@ -39,6 +39,7 @@ type PrometheusEmitter struct {
 	pipelineScheduled *prometheus.CounterVec
 
 	resourceChecksVec *prometheus.CounterVec
+	checksVec         *prometheus.CounterVec
 
 	schedulingFullDuration    *prometheus.CounterVec
 	schedulingLoadingDuration *prometheus.CounterVec
@@ -329,6 +330,17 @@ func (config *PrometheusConfig) NewEmitter() (metric.Emitter, error) {
 	)
 	prometheus.MustRegister(resourceChecksVec)
 
+	checksVec := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "concourse",
+			Subsystem: "lidar",
+			Name:      "checks_total",
+			Help:      "Counts the number of checks finished",
+		},
+		[]string{"scopeID", "checkName"},
+	)
+	prometheus.MustRegister(checksVec)
+
 	listener, err := net.Listen("tcp", config.bind())
 	if err != nil {
 		return nil, err
@@ -358,6 +370,7 @@ func (config *PrometheusConfig) NewEmitter() (metric.Emitter, error) {
 		pipelineScheduled: pipelineScheduled,
 
 		resourceChecksVec: resourceChecksVec,
+		checksVec:         checksVec,
 
 		schedulingFullDuration:    schedulingFullDuration,
 		schedulingLoadingDuration: schedulingLoadingDuration,
@@ -422,6 +435,8 @@ func (emitter *PrometheusEmitter) Emit(logger lager.Logger, event metric.Event) 
 		emitter.databaseMetrics(logger, event)
 	case "resource checked":
 		emitter.resourceMetric(logger, event)
+	case "check finished":
+		emitter.checkMetric(logger, event)
 	default:
 		// unless we have a specific metric, we do nothing
 	}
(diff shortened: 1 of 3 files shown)


xtremerui:
@cirocosta

atc/metric/emitter/prometheus.go
```diff
@@ -691,6 +693,21 @@ func (emitter *PrometheusEmitter) resourceMetric(logger lager.Logger, event metr
...
+		logger.Error("failed-to-find-resource-config-scope-id-in-event", fmt.Errorf("expected scope_id to exist in event.Attributes"))
+		return
+	}
+	checkName, exists := event.Attributes["check_name"]
+	if !exists {
+		logger.Error("failed-to-find-check-name-in-event", fmt.Errorf("expected check_name to exist in event.Attributes"))
+		return
+	}
+
+	emitter.resourceChecksVec.WithLabelValues(scopeID, checkName).Inc()
```
cirocosta: The definition of  `resourceChecksVec` is

```go

	resourceChecksVec := prometheus.NewCounterVec(
		prometheus.CounterOpts{
			Namespace: "concourse",
			Subsystem: "resource",
			Name:      "checks_total",
			Help:      "Counts the number of resource checks performed",
		},
		[]string{"team", "pipeline"},
```

but here it seems like we're passing `scopeID` and `checkName`, rather than `team` and `pipeline` 🤔  is this right?

cirocosta: np! here's the diff for `docker-compose.yml` that I used:

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 400598ee0..e055d324d 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -29,6 +29,9 @@ services:
       CONCOURSE_ADD_LOCAL_USER: test:test,guest:guest
       CONCOURSE_MAIN_TEAM_LOCAL_USER: test
       CONCOURSE_CLUSTER_NAME: dev
+      CONCOURSE_PROMETHEUS_BIND_IP: 0.0.0.0
+      CONCOURSE_PROMETHEUS_BIND_PORT: 9100
+      CONCOURSE_ENABLE_LIDAR: "true"
 
   worker:
     build: .
```

xtremerui: Yeah in the latest commit I tested prometheus emitter with that it should be good.

atc/metric/emitter/prometheus.go
```diff
@@ -305,6 +306,17 @@ func (config *PrometheusConfig) NewEmitter() (metric.Emitter, error) {
...
 	prometheus.MustRegister(resourceChecksVec)
 
+	checksVec := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "concourse",
+			Subsystem: "lidar",
+			Name:      "checks_total",
+			Help:      "Counts the number of checks finished",
+		},
+		[]string{"scopeID", "checkName"},
```
cirocosta: heey, do you think we really need these two?

Assuming that some of these timeseries dbs (like prometheus' tsdb) create a timeseries for each labelset (combination of those labels like `scopeID` and `checkName`), this can get super expensive without really bringing much value.

wdyt of  just dropping those two? 🤔  would an operator lose much? IMO it's worth dropping - we aggregate that with a `sum` in our dashboards anyway 😅  

thx!

cirocosta: (if we go without those two dimensions, we'd then go from `NewCounterVec` to `NewCounter`, and then in the place where we do the increments, we'd go for a simple `Inc()` instead of  chaining`WithLabelValues()`)

xtremerui:
@cirocosta updated with changes you suggested

PR #4559, opened 2019-10-04
Issue # 2958

ctorboh15:
Fixes # 2958

We now remove any api members that are not TLS specific in the even that we set the CONCOURSE_TLS_ONLY env variable

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -690,13 +690,7 @@ func (cmd *RunCommand) constructAPIMembers(
 		)})
 
 		if cmd.TLSOnly {
-			tmp := members[:0]
-			for _, apiMember := range members {
-				if apiMember.Name != "web" {
-					tmp = append(tmp, apiMember)
-				}
-			}
-			members = tmp
+			members = cmd.filterMembersForTLSOnly(members, "web")
 		}
 	}
 
@@ -1551,4 +1545,14 @@ func (cmd *RunCommand) appendStaticWorker(
 
 func (cmd *RunCommand) isTLSEnabled() bool {
 	return cmd.TLSBindPort != 0
+}
+
+func (cmd *RunCommand) filterMembersForTLSOnly(apiMembersArray []grouper.Member, memberNameToFilter string) []grouper.Member {
+	tmp := apiMembersArray[:0]
+	for _, apiMember := range apiMembersArray {
+		if apiMember.Name != memberNameToFilter {
+			tmp = append(tmp, apiMember)
+		}
+	}
+	return tmp
 }
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -87,6 +87,7 @@ type RunCommand struct {
 	BindIP   flag.IP `long:"bind-ip"   default:"0.0.0.0" description:"IP address on which to listen for web traffic."`
 	BindPort uint16  `long:"bind-port" default:"8080"    description:"Port on which to listen for HTTP traffic."`
 
+	TLSOnly     bool      `long:"tls-only" description:"Configure concourse to only accept HTTPS requests"`
 	TLSBindPort uint16    `long:"tls-bind-port" description:"Port on which to listen for HTTPS traffic."`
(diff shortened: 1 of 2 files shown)


deniseyu:
hey @ctorboh15 thanks for taking a look at this!! that issue is pretty old, and probably shouldn't have been labeled good-first-issue since the person who wrote it has left the team 🤔 but it's a good reminder for us to keep better track of how we use that label.

With this PR there would three possible states:

1. Support HTTP only
2. Support HTTPS only
3. Support both, but if port 443 is configured, then redirect traffic from 80 to 443

I wonder if we should make it clearer somewhere that enabling this flag would disable option 3 above entirely, because I can kind of imagine people being surprised at losing the redirect that browsers give them for free today. Not sure where's the best place to do this though. Open to your feedback!

Going to put some feedback about the filtering logic as inline change requests!

ctorboh15:
Im not really sure how the documentation for concourse is structured. I think it should be part of the ATC ReadME.md since it will be serving most of the web request. Better yet I think it would be beneficial to have documentation around how to deploy concourse. Like a list for available env vars that can be set or something to that effect.  To the point about redirects ... are we looking at this strictly from a web ui point of view? I can imagine scenarios where people try to deploy pipelines to concourse through concourse. In which case redirects could potentially start to get hairy.

deniseyu:
I think the redirect behavior mostly is just by browsers (i think it's Chrome/Firefox that does it? but not sure, been a while since I've done web stuff!!) for sure it's not good to rely on the redirect! Given that this flag would have to be explicitly turned on I don't think it's such a big deal.

Totally hear you on the documentation ambiguity. There's a tiny bit about how to get up and running in this repo's readme, but there isn't a dedicated section in the [official docs](https://concourse-ci.org/docs.html). There are a few environment flags mentioned in the page about the [web node](https://concourse-ci.org/concourse-web.html), but there's definitely room for improvement!

Anyways, I wouldn't necessarily expect the docs change to be part of this PR, just wanted to call out that down the line it might be useful! Thanks for thinking about it.

ctorboh15:
@deniseyu Is there anything you need from me?

atc/atccmd/command.go
```diff
@@ -687,6 +688,10 @@ func (cmd *RunCommand) constructAPIMembers(
 			httpsHandler,
 			tlsConfig,
 		)})
+
+		if cmd.TLSOnly {
+			members = cmd.filterMembersForTLSOnly(members, "web")
```
vito: I think it'd be cleaner to just conditionally append the non-TLS handler instead of filtering it out:

```go
if !cmd.TLSOnly {
	members = append(members, grouper.Member{
		Name: "web",
		Runner: http_server.New(
			cmd.nonTLSBindAddr(),
			httpHandler,
		),
	})
}
```

release-notes/latest.md
```diff
@@ -26,3 +26,6 @@
 #### <sub><sup><a name="4507" href="#4507">:link:</a></sup></sub> fix
 
 * @iamjarvo fixed a [bug](444://github.com/concourse/concourse/issues/4472) where `fly builds` would show the wrong duration for cancelled builds #4507.
+
+#### <sub><sup><a name="2958" href="#2958">:link:</a></sup></sub> feature
+* @ctorboh15 added support for a CONCOURSE_TLS_ONLY configuration that will remove all non TLS api members at startup
```
vito:
```suggestion
* @ctorboh15 added support for a CONCOURSE_TLS_ONLY configuration that will disable the non-TLS API listener.
```

atc/atccmd/command.go
ctorboh15: I thought of that but I added the filtering here to ensure that the user cant get into an a weird state where they might set TLS_ONLY to true but not have set the TLS_BIND_PORT etc. It seemed safer to filter than to do a conditional. I'm curious your thoughts on this

vito: Ah, yeah that makes sense. How about checking for that case in `validate()` instead? We do something similar here:

https://github.com/concourse/concourse/blob/6def8a4f5fe426150a967697ef51ee5fe5011e26/atc/atccmd/command.go#L1153-L1159

Issue #4562, opened 2019-10-04
Worker and DB container state can fall out of sync when Garden is slow to respond


kcmannem:
Looking at the containers listed in the worker. Many are not in the db.
```
worker/4fd44d6a-2376-4ec8-a187-25b25087e2a5:/var/vcap/data/worker/work/depot# ls -lt
total 692
drwxr-xr-x 2 root root 4096 Oct  4 18:45 db0ad945-cd43-4924-5359-aba00c3b38bd
drwxr-xr-x 3 root root 4096 Oct  4 18:45 15980398-ff0c-4ca8-50c4-c7cbf6d982b9
drwxr-xr-x 3 root root 4096 Oct  4 18:45 3c9ecd39-45b6-43ed-7bb4-bdf869a0ba31
drwxr-xr-x 3 root root 4096 Oct  4 18:45 826d3627-3d3c-4ee9-5844-abaf2255c9ad
drwxr-xr-x 3 root root 4096 Oct  4 18:45 6c12c532-9570-4fcd-707a-b3fa9f9e8b84
drwxr-xr-x 3 root root 4096 Oct  4 18:45 47bf16c7-4951-4f6b-5322-e23e455239bb
drwxr-xr-x 3 root root 4096 Oct  4 18:45 6c1b8539-e2ac-41fb-4522-98cd7a4a8c4b
drwxr-xr-x 3 root root 4096 Oct  4 18:45 068c55b0-68eb-428d-76ec-feb25420e044
drwxr-xr-x 3 root root 4096 Oct  4 18:45 59ab81b5-debe-4a7d-4c8f-5674cc26e826
drwxr-xr-x 3 root root 4096 Oct  4 18:45 9304b72b-0f37-4ee4-53b4-244276691f4d
drwxr-xr-x 3 root root 4096 Oct  4 18:45 65812853-3116-4f80-4f35-bcd3fdeeb989
drwxr-xr-x 3 root root 4096 Oct  4 18:45 2311313d-bb62-4188-4e68-5a0a4aa598a6
drwxr-xr-x 3 root root 4096 Oct  4 18:45 661fa51c-e91d-43df-656b-2968aea4ed0e
drwxr-xr-x 3 root root 4096 Oct  4 18:45 a58f5d52-c75d-477c-5751-9ce5fea39a89
drwxr-xr-x 3 root root 4096 Oct  4 18:45 d2bc537a-2de3-433e-744c-dff35c64e907
[...]

PR #4563, opened 2019-10-04
increase gclient http timeout for very slow create reqs

kcmannem:
temp fix for #4562


kcmannem:
I'm not going to merge this in anymore. We're moving towards enabling gc to cleanup untracked containers on workers. We didn't do this earlier since it would mask any issues our gc might have but  if we emit a metric when this occurs, it'll give us the same visibility without the failure mode. The timeout was also a step backwards since it could still fall back to an old scenario where resource checks/steps would hang randomly because of an unresponsive request.

PR #4564, opened 2019-10-04
add release note for gzip and pin switching

xtremerui:


pivotal-jamie-klassen:
so, i merged this but it's not going into v5.6.1. I will issue a followup to fix.

Issue #4425, opened 2019-09-13
'check' Output Should be Improved when Erroring


coofercat:
## What challenge are you facing?

Writing Resources is both simple and difficult. They have a fantastically simple API, but actually running and testing them can be tricky.

In the case of 'check' scripts, they can't routinely output any STDERR (unless they're returning an error), and so they are hard to debug if Concourse doesn't like the output it receives from them. A typical Concourse error message may be a single line like:

`cannot unmarshal number into go value of type string`

Whilst this does provide clues about what's wrong, it requires complete knowledge of the JSON in/out mechanism, accurate knowledge of the JSON being returned by the check script in exactly these circumstances, and of course understanding of the error itself.

## What would make this better?

Simply dump the JSON received by Concourse alongside the error. For example:

```
cannot unmarshal number into go value of type string

check returned this JSON:

[ {"version": 1234 } ]
```

Obviously, if it were possible to explicitly highlight the exact error, then all the better, but the above (+google!) hopefully offers enough clues to be able to figure out what to do to resolve the problem.

Likewise, if some means to output STDERR were available for checks, then this would also be beneficial, but the above seems like (I hope!) a simple solution that doesn't need extensive UI or code changes to implement.

vito:
Great idea! I think it would be a simple change to add here:

https://github.com/concourse/concourse/blob/5897aa409da2093b1c2cbf76ec1786deddba67a3/atc/resource/run_script.go#L135

This would apply to all resource actions, so it could help with debugging `get` and `put` as well.

lmorg:
@vito I appreciate this issue is now closed but I wonder if it is worth using `UnmarshalIndent()` for human readable printing, rather than regular old `Unmarshal` which would return a minified dump.

Issue #1038, opened 2017-05-02
Removing cached volumes


kmacoskey:
# Feature Request

## What challenge are you facing?

A problem arose such that we needed to change the git tag of a product release to an annotated tag without additional changes/commits to the repository. This change did not create a new version of the resource, therefore the build pipeline involved would only start new builds with the cached version of the repository. It was needed then to clear that cached resources from the concourse deployment.

We were able to identify the particular volumes with `fly volumes`. With those volume handles and worker id's, we did a baggageclaim api DELETE for each of those volumes on their respective workers. Following that we removed the corresponding rows in the database for the volumes in question. This successfully removed the volumes and the original pipeline in question was able to re-cache the git resource with the updated tag. 

This however had the unintended side effect that all pipelines in the same concourse deployment would occasionally error unpredictably with `error: cannot create volume`. We asked for assistance in the concourse slack and were told there is no clean way to remove cached volumes. The end result was that in order to return to a good working state we had to redeploy all of the workers of the deployment.  

## A Modest Proposal

The functionality and subsequent command in fly to delete a volume by a given volume handle:
`fly delete-volume [handle]`

joshzarrabi:
@kmacoskey We know that this is relatively unfriendly behavior, and we have a story to look into ways to improving our caching strategy. 

Right now concourse relies on strict semver versioning for all of your resources, and any deviation to this basically requires that you recreate your workers. We want to take away some of this pain, but we also don't want to encourage messy versioning, so its something we need to think about...

vito:
This problem arises when resource versions are updated-in-place. Concourse relies heavily on resource versions being immutable (not necessarily semver), for many reasons. If a version changes in the middle of a pipeline run, there's no guarantee that the things that ran before its mutation are still valid (i.e. that it still passes tests). Relying on immutability also allows us to implement caching generically, without resource authors having to implement "cache busting" detectors and such.

That being said, I do see this come up pretty often, so it's worth thinking about, as @joshzarrabi said. **HOWEVER**, this does not necessarily mean that we need a way to bust caches. If we add a `delete-volume`, or an `invalidate-cache`, that could paper over deeper issues, leaving them unsolved and leaving our users constantly running these commands. (The [`knobs`](https://github.com/concourse/concourse/wiki/Anti-Patterns#knobs) anti-pattern.) For example, it could be that the resource's versioning semantics aren't precise enough for your use case, or that you just need a different resource.

[...]

PR #4568, opened 2019-10-06
change: improved error message on malformed output for resources scripts

ProvoK:
Fixes https://github.com/concourse/concourse/issues/4425

Signed-off-by: Vittorio Camisa <vittorio.camisa@gmail.com>


atc/resource/run_script.go
```diff
@@ -131,8 +131,11 @@ func (resource *resource) runScript(
 				return err
 			}
 		}
-
-		return json.Unmarshal(stdout.Bytes(), output)
+		err := json.Unmarshal(stdout.Bytes(), output)
+		if err != nil {
+			return fmt.Errorf("%s. \nCommand returned this payload:\n %s", err, stdout.String())
```
vito: spacing/capitalization feels a little odd, here's a quick suggestion:

```suggestion
			return fmt.Errorf("%s\n\nwhen parsing resource response:\n\n%s", err, stdout.String())
```

```
cannot unmarshal number into go value of type string

when parsing resource response:

[ {"version": 1234 } ]
```

ProvoK:
@vito Done!
Happy to help :+1:  As you may have noticed, I'm looking around to help the cause. If you have any (not huge, I'm still totally new to the codebase) suggestion on what to pick next, just tell me!
You can find me also on Discord, same name :smile:

ProvoK:
> @ProvoK Awesome, thanks for contributing! This looks good now so I'll go ahead and merge. 
> 
> This feature would make a lot of people happy: [#1038 (comment)](https://github.com/concourse/concourse/issues/1038#issuecomment-466658736) - it's probably my top 'help wanted' issue.

@vito I'll try into that, there are involved a lot of changes but the scope is narrow. This kind of task will for sure help me explore a lot more of the codebase and the project, I'm intrigued :D

atc/resource/run_script.go
```diff
@@ -131,8 +131,11 @@ func (resource *resource) runScript(
 				return err
 			}
 		}
-
-		return json.Unmarshal(stdout.Bytes(), output)
+		err := json.Unmarshal(stdout.Bytes(), output)
+		if err != nil {
+			return fmt.Errorf("%s\n\nwhen parsing resource response:\n\n%s", err, stdout.String())
+		}
+		return err
```
evanchaoli: If err is not nil, then it returns at line 136, so should this line just return `nil`?

ProvoK: Yes, you're right. The only possible value for err is nil at line 138. I agree with you, it would more clear directly to return nil

PR #4577, opened 2019-10-07
GC unknown containers and volumes from workers

### LINKED ISSUE #3600: GC Containers and volumes when present in the worker but not in the DB

### LINKED ISSUE #4571: Add observability into when containers get out of sync between workers and DB
We don't currently delete containers that exist on the worker but not in the database, but we should have a metric and logs for when we see this occurring. 

We do not currently have any insight into why this may occur which makes it difficult to form any theories as to why we sometimes have more containers on the worker than in the db.

ddadlani:
# Existing Issue
Fixes #3600
Fixes #4571 

*  when reporting containers from worker, create destroying containers in the db for any unknown containers
*  these new destroying containers will then get gc-ed on the next gc tick
* metrics to be added in #4571 

# Contributor Checklist
- [X] Unit tests

# Reviewer Checklist
- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [x] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/api/containerserver/report.go
+++ b/atc/api/containerserver/report.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 
 	"code.cloudfoundry.org/lager"
+	"github.com/concourse/concourse/atc/metric"
 )
 
 func (s *Server) ReportWorkerContainers(w http.ResponseWriter, r *http.Request) {
@@ -52,6 +53,11 @@ func (s *Server) ReportWorkerContainers(w http.ResponseWriter, r *http.Request) 
 			"worker-name":   workerName,
 			"handles-count": numUnknownContainers,
 		})
+
+		metric.WorkerUnknownContainers{
+			WorkerName: workerName,
+			Containers: numUnknownContainers,
+		}.Emit(logger)
 	}
 
 	err = s.containerRepository.UpdateContainersMissingSince(workerName, handles)
--- a/atc/api/containerserver/report.go
+++ b/atc/api/containerserver/report.go
@@ -41,6 +41,13 @@ func (s *Server) ReportWorkerContainers(w http.ResponseWriter, r *http.Request) 
 		"num-handles": len(handles),
 	})
 
+	_, err = s.containerRepository.DestroyUnknownContainers(workerName, handles)
+	if err != nil {
+		logger.Error("failed-to-destroy-unknown-containers", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
 	err = s.containerRepository.UpdateContainersMissingSince(workerName, handles)
 	if err != nil {
 		logger.Error("failed-to-update-containers-missing-since", err)
--- a/atc/api/containerserver/report.go
+++ b/atc/api/containerserver/report.go
@@ -41,11 +41,17 @@ func (s *Server) ReportWorkerContainers(w http.ResponseWriter, r *http.Request) 
 		"num-handles": len(handles),
 	})
 
-	_, err = s.containerRepository.DestroyUnknownContainers(workerName, handles)
+	numUnknownContainers, err := s.containerRepository.DestroyUnknownContainers(workerName, handles)
 	if err != nil {
 		logger.Error("failed-to-destroy-unknown-containers", err)
 		w.WriteHeader(http.StatusInternalServerError)
 		return
+	}
+	if numUnknownContainers > 0 {
+		logger.Info("unknown-container-handles", lager.Data{
+			"worker-name":   workerName,
+			"handles-count": numUnknownContainers,
+		})
 	}
 
 	err = s.containerRepository.UpdateContainersMissingSince(workerName, handles)
--- a/atc/api/containerserver/report.go
+++ b/atc/api/containerserver/report.go
@@ -5,8 +5,9 @@ import (
 	"io/ioutil"
 	"net/http"
 
-	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc/metric"
+
+	"code.cloudfoundry.org/lager"
 )
 
 func (s *Server) ReportWorkerContainers(w http.ResponseWriter, r *http.Request) {
@@ -53,12 +54,12 @@ func (s *Server) ReportWorkerContainers(w http.ResponseWriter, r *http.Request) 
 			"worker-name":   workerName,
 			"handles-count": numUnknownContainers,
 		})
-
-		metric.WorkerUnknownContainers{
-			WorkerName: workerName,
-			Containers: numUnknownContainers,
-		}.Emit(logger)
 	}
+
+	metric.WorkerUnknownContainers{
+		WorkerName: workerName,
+		Containers: numUnknownContainers,
+	}.Emit(logger)
 
 	err = s.containerRepository.UpdateContainersMissingSince(workerName, handles)
 	if err != nil {
--- a/atc/api/volumeserver/report.go
+++ b/atc/api/volumeserver/report.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 
 	"code.cloudfoundry.org/lager"
+	"github.com/concourse/concourse/atc/metric"
 )
 
 // ReportWorkerVolumes provides an API endpoint for workers to report their current volumes
@@ -53,6 +54,10 @@ func (s *Server) ReportWorkerVolumes(w http.ResponseWriter, r *http.Request) {
 			"worker-name":   workerName,
 			"handles-count": numUnknownVolumes,
 		})
+		metric.WorkerUnknownVolumes{
+			WorkerName: workerName,
+			Volumes:    numUnknownVolumes,
+		}.Emit(logger)
 	}
 
 	err = s.repository.UpdateVolumesMissingSince(workerName, handles)
--- a/atc/api/volumeserver/report.go
+++ b/atc/api/volumeserver/report.go
@@ -42,6 +42,13 @@ func (s *Server) ReportWorkerVolumes(w http.ResponseWriter, r *http.Request) {
 		"handles-count": len(handles),
 	})
 
+	_, err = s.repository.DestroyUnknownVolumes(workerName, handles)
+	if err != nil {
+		logger.Error("failed-to-destroy-unknown-volumes", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
 	err = s.repository.UpdateVolumesMissingSince(workerName, handles)
 	if err != nil {
 		logger.Error("failed-to-update-volumes-missing-since", err)
--- a/atc/api/volumeserver/report.go
+++ b/atc/api/volumeserver/report.go
@@ -42,11 +42,17 @@ func (s *Server) ReportWorkerVolumes(w http.ResponseWriter, r *http.Request) {
 		"handles-count": len(handles),
 	})
 
-	_, err = s.repository.DestroyUnknownVolumes(workerName, handles)
+	numUnknownVolumes, err := s.repository.DestroyUnknownVolumes(workerName, handles)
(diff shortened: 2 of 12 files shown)


ddadlani:
Documentation will be completed as part of #4562

kcmannem:
@cirocosta would appreciate if you can check if we added metrics correctly.

atc/api/containerserver/report.go
```diff
@@ -41,6 +42,24 @@ func (s *Server) ReportWorkerContainers(w http.ResponseWriter, r *http.Request)
...
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if numUnknownContainers > 0 {
+		logger.Info("unknown-container-handles", lager.Data{
+			"worker-name":   workerName,
+			"handles-count": numUnknownContainers,
+		})
+
+		metric.WorkerUnknownContainers{
```
cirocosta: as we represent `unknown containers` as a gauge on pull-based "emitters" (prometheus at the moment), we can get into the following:

- t0: `numUnknownContainers == 10`
  - gauge = 10
- t1: `numUnknownContainers == 0`
  - gauge = 10
- t2: `numUnknownContainers == 0`
  - gauge = 10
- t3: `numUnknownContainers == 0`
  - gauge = 10

It'd be better if the metric emission happened outside that condition 🤔

atc/api/volumeserver/report.go
```diff
@@ -42,6 +43,23 @@ func (s *Server) ReportWorkerVolumes(w http.ResponseWriter, r *http.Request) {
 		"handles-count": len(handles),
 	})
 
+	numUnknownVolumes, err := s.repository.DestroyUnknownVolumes(workerName, handles)
+	if err != nil {
+		logger.Error("failed-to-destroy-unknown-volumes", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	if numUnknownVolumes > 0 {
```
cirocosta: (see comment on containers regarding having the metric emission outside the condition 😁 )

atc/metric/emitter/prometheus.go
```diff
@@ -374,6 +400,10 @@ func (emitter *PrometheusEmitter) Emit(logger lager.Logger, event metric.Event)
 		emitter.workerContainersMetric(logger, event)
 	case "worker volumes":
 		emitter.workerVolumesMetric(logger, event)
+	case "worker unknown containers": ////
```
cirocosta: whops, left the `/////`

atc/metric/emitter/prometheus.go
```diff
@@ -533,6 +563,31 @@ func (emitter *PrometheusEmitter) workersRegisteredMetric(logger lager.Logger, e
...
 
+func (emitter *PrometheusEmitter) workerUnknownContainersMetric(logger lager.Logger, event metric.Event) {
+	worker, exists := event.Attributes["worker"]
+	if !exists {
+		logger.Error("failed-to-find-worker-in-event", fmt.Errorf("expected worker to exist in event.Attributes"))
+		return
+	}
+
+	labels := prometheus.Labels{
+		"worker":   worker,
```
cirocosta: nit: formatting

atc/db/container_repository.go
```diff
@@ -362,3 +363,46 @@ func (repository *containerRepository) DestroyFailedContainers() (int, error) {
...
+	unknownHandles := diff(reportedHandles, dbHandles)
+
+	tx, err := repository.conn.Begin()
+	if err != nil {
+		return 0, err
+	}
+
+	defer Rollback(tx)
+
+	for _, unknownHandle := range unknownHandles {
```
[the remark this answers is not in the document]
kcmannem: we thought about this but yeah we didn't know how to do it with squirrel

atc/api/containerserver/report.go
kcmannem: good catch

atc/metric/emitter/prometheus.go
kcmannem: it was on purpose

atc/db/container_repository.go
```diff
@@ -362,3 +363,45 @@ func (repository *containerRepository) DestroyFailedContainers() (int, error) {
...
+
+func (repository *containerRepository) DestroyUnknownContainers(workerName string, reportedHandles []string) (int, error) {
+	dbHandles, err := repository.queryContainerHandles(sq.Eq{
+		"worker_name": workerName,
+	})
+	if err != nil {
+		return 0, err
+	}
+
+	unknownHandles := diff(reportedHandles, dbHandles)
```
cirocosta: should we stop here? as there are no `unknownHandles` to move to `destroying` 🤔

atc/db/container_repository_test.go
```diff
@@ -1003,4 +1003,77 @@ var _ = Describe("ContainerRepository", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(result.RowsAffected()).To(Equal(int64(1)))
+		})
+
+		JustBeforeEach(func() {
+			num, err = containerRepository.DestroyUnknownContainers(defaultWorker.Name(), workerReportedHandles)
+			Expect(err).ToNot(HaveOccurred())
+		})
+
+		Context("when there are containers on the worker that are not in the db", func() {
```
cirocosta: maybe this is captured somewhere, but, might be important to capture the case when the are no unknown handles in the db 😁

atc/db/container_repository.go
kcmannem: done

atc/db/container_repository_test.go
kcmannem: thanks for the catch!

atc/db/container_repository_test.go
```diff
@@ -1003,4 +1003,94 @@ var _ = Describe("ContainerRepository", func() {
...
+
+			It("does not affect containers in any other state", func() {
+				result, err := psql.Select("*").
+					From("containers").
+					Where(sq.Eq{"state": atc.ContainerStateCreated}).
+					RunWith(dbConn).Exec()
+
+				Expect(err).ToNot(HaveOccurred())
+				Expect(result.RowsAffected()).To(Equal(int64(1)))
+				Expect(num).To(Equal(2))
```
xtreme-sameer-vohra: its redundant but should we confirm the handle is "some-handle2"

atc/db/volume_repository.go
```diff
@@ -540,6 +542,53 @@ func (repository *volumeRepository) GetDestroyingVolumes(workerName string) ([]s
...
+
+	tx, err := repository.conn.Begin()
+	if err != nil {
+		return 0, err
+	}
+
+	defer Rollback(tx)
+
+	for _, unknownHandle := range unknownHandles {
+		_, err = psql.Insert("volumes").
```
clarafu: Do we want to do the same with the destroying container inserts where we do one large insert rather than a ton of single row inserts?

atc/db/volume_repository.go
```diff
@@ -540,6 +542,53 @@ func (repository *volumeRepository) GetDestroyingVolumes(workerName string) ([]s
...
+			Columns(
+				"handle",
+				"worker_name",
+				"state",
+			).
```
xtreme-sameer-vohra: Should we use an InsertBuilder outside the for loop similar to the pattern used for the container_repository.go

atc/db/container_repository.go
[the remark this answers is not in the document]
clarafu: noice!!! 👍

atc/db/volume_repository_test.go
```diff
@@ -908,4 +908,90 @@ var _ = Describe("VolumeFactory", func() {
...
+			})
+
+			It("does not affect volumes in any other state", func() {
+				result, err := psql.Select("*").
+					From("volumes").
+					Where(sq.Eq{"state": db.VolumeStateCreated}).
+					RunWith(dbConn).Exec()
+
+				Expect(err).ToNot(HaveOccurred())
+				Expect(result.RowsAffected()).To(Equal(int64(1)))
```
xtreme-sameer-vohra: Can we check that the volume handle is `some-handle2`

atc/db/volume_repository.go
clarafu: image

ddadlani:
we should add a description of `unknown containers` to the troubleshooting guide, just so that people know what it means and when to look at it.

xtreme-sameer-vohra:
Hey @ddadlani @kcmannem 
We need to update the other emitters as well.

xtreme-sameer-vohra:
With @cirocosta  validated that unknown containers and volumes on workers from the web's perspective are cleaned up

Steps
- on the worker, ran the following script
```
#!/bin/bash

set -o errexit
set -o nounset

main () {
    for i in $(seq 1 10); do
        create_container handle-$i
        create_volume handle-$i
    done
}

# create_container creates a container
#
# args:
#  1 - handle
#
create_container () {
    local handle=$1
    local rootfs=/tmp/$handle


    mkdir -p $rootfs
    curl -XPOST \
        localhost:7777/containers \
        -d "$(printf '{"handle":"%s", "rootfs":"raw://%s"}' $handle $rootfs)"
}

# create_volume creates a volume
#
# args:
#  1 - handle
#
create_volume () {
    local handle=$1
    
    curl -X POST \
     localhost:7788/volumes \
     -d "$(printf '{"handle":"%s", "strategy":{"type":"empty"}}' $handle)" \
      -H "Content-Type: application/json"
}

main
```
- monitored logs for messages 
```
web_1     | {"timestamp":"2019-10-18T19:24:43.763080800Z","level":"info","source":"atc","message":"atc.report-containers-for-worker.unknown-container-handles","data":{"handles-count":10,"name":"bce08d070240","session":"4209","worker-name":"bce08d070240"}}
web_1     | {"timestamp":"2019-10-18T19:24:43.763724100Z","level":"info","source":"atc","message":"atc.report-volumes-for-worker.unknown-volume-handles","data":{"handles-count":10,"name":"bce08d070240","session":"4211","worker-name":"bce08d070240"}}
```
- checked the state of garden and baggageclaim servers
```
curl localhost:7777/containers
curl localhost:7788/volumes
```
and observed unknown volumes and containers were deleted.

xtreme-sameer-vohra:
Validated Prometheus by configuring the endpoint and port and checking the `concourse_workers_unknown_containers` & `concourse_workers_unknown_volumes` metrics were being reported

kcmannem:
@xtreme-sameer-vohra thanks for checking and fixing prometheus as well

atc/db/container_repository.go
```diff
@@ -362,3 +363,49 @@ func (repository *containerRepository) DestroyFailedContainers() (int, error) {
...
+
+	if len(unknownHandles) == 0 {
+		return 0, nil
+	}
+
+	tx, err := repository.conn.Begin()
```
clarafu: nit: Not something you need to change but you probably don't need to open a transaction anymore, can probably just use the connection in the query.

PR #4580, opened 2019-10-08
allow HTTPS *only* configuration in ATC

### LINKED ISSUE #2958: ATC should allow HTTPS *only* configuration
Currently the ATC requires that the settings for HTTP are configured, and HTTPS is optional. We do redirect back to the configured `external_url`, which can "force" a redirect to `https://..`.

However, some users may want to configure their ATCs to not listen on HTTP at all.

ctorboh15:
Fixes #2958

added support for a CONCOURSE_TLS_ONLY configuration that will disable the non-TLS API listener.

** Note I had to reopen this as my other branch had a rogue unsigned commit that was getting out of hand **

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -87,6 +87,7 @@ type RunCommand struct {
 	BindIP   flag.IP `long:"bind-ip"   default:"0.0.0.0" description:"IP address on which to listen for web traffic."`
 	BindPort uint16  `long:"bind-port" default:"8080"    description:"Port on which to listen for HTTP traffic."`
 
+	TLSOnly     bool      `long:"tls-only" 			description:"Configure concourse to only accept HTTPS requests"`
 	TLSBindPort uint16    `long:"tls-bind-port" description:"Port on which to listen for HTTPS traffic."`
 	TLSCert     flag.File `long:"tls-cert"      description:"File containing an SSL certificate."`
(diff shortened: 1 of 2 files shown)


ctorboh15:
@vito Can you take a look at this. I implemented your suggested changes

atc/atccmd/command.go
```diff
@@ -1171,6 +1175,13 @@ func (cmd *RunCommand) validate() error {
 				errors.New("must specify HTTPS external-url to use TLS"),
 			)
 		}
+	case cmd.TLSOnly:
+		if !cmd.isTLSEnabled() || !cmd.LetsEncrypt.Enable {
+			errs = multierror.Append(
+				errs,
+				errors.New("must specify a --tls-bind-port, or --enable-lets-encrypt to use TLS "),
```
vito: This won't work because the `switch` statement run the `cmd.TLSBindPort == 0` (line 1157) instead. In fact it seems like a lot of these validations are kind of awkwardly overlapping as a result of using `switch`. :thinking: It's really one big chained `else-if`.

Using a `switch` statement here feels like it's not aging well. It could be refactored, or you could just put a `if cmd.TLSOnly` in the `cmd.TLSBindPort == 0` branch. I'd be fine with both; the latter is probably less work.

The error can then read `must specify --tls-bind-port in order to configure --tls-only`. I don't think you need to mention Let's Encrypt at all.

vito:
Closing due to inactivity, but feel free to reopen once you have time - thanks!

PR #4582, opened 2019-10-08
use basic garden client with timeout in container sweeper

### LINKED ISSUE #4467: Container and Volume sweeping relies on `gdn` and `baggageclaim` properly responding on time
## Bug Report

I was looking at the container sweeping logic, and it seems to me that if a
single call to volume sweeping takes forever to respond, no sweeping at all
will take place for the next ticks (as next ticks will not occur).

e.g., the runner is essentially something that runs on every tick if never
whenever unblocked:

```go
func (sweeper *containerSweeper) Run(...) {

    for {
        select {
        case <-timer.C:
            // cc: I assume this might take forever
            //
            sweeper.sweep(sweeper.logger.Session("tick"))

        case sig := <-signals:
            sweeper.logger.Info("sweep-cancelled-by-signal", lager.Data{"signal": sig})
            return nil
        }
    }
}
```

(see [containerSweeper.Run](https://github.com/concourse/concourse/blob/ac5f6b673ff6ab9cc45b4e315c81bddba1d13352/worker/container_sweeper.go#L40-L55))

looking at `sweeper.sweep`:

```go
func (sweeper *containerSweeper) sweep(logger lager.Logger) {
    ctx := lagerctx.NewContext(context.Background(), logger)

    // cc: grab all handles ... etc
    //
    containers, err := sweeper.gardenClient.Containers(garden.Properties{})

    // cc: grab the handles to destroy
    //
    containerHandles, err := sweeper.tsaClient.ContainersToDestroy(ctx)


    // cc: for each handle, submit to the "work queue" to delete
    //
    for _, handle := range containerHandles {
        maxInFlight <- 1
        wg.Add(1)

        go func(handle string) {
            err := sweeper.gardenClient.Destroy(handle)
            if err != nil {
                logger.WithData(lager.Data{"handle": handle}).
                    Error("failed-to-destroy-container", err)
            }

            <-maxInFlight
            wg.Done()
        }(handle)
    }

    // cc: this will wait until all finish, thus, we assume
    //     that they all eventually finish on a reasonable time
    //
    wg.Wait()
}
```

(see [containerSweeper.sweep](https://github.com/concourse/concourse/blob/ac5f6b673ff6ab9cc45b4e315c81bddba1d13352/worker/container_sweeper.go#L57-L97))

ps.: if a single deletion takes forever, the current tick's containers will all be deleted (except for that faulty one), but as we wait on the entirety of it, the next tick will never happen (my hypothesis)

## Steps to Reproduce

I didn't try reproducing this, but I can see that it'd be quite simple to do an e2e verification of this by modifying Houdini to not properly do the deletion.

## Expected Results

Not prevent further ticks in the case of a deletion being stuck

## Actual Results

Gets stuck if a deletion never returns (assumption). 

## Additional Context

Note.: we've recently made the web nodes more resilient by introducing broader
use of context cancellation (see [resource checking should be resilient to a 'stuck' worker](https://github.com/concourse/concourse/issues/3928)), but it seems like we didn't do the same for the worker side.

## Version Info

* Concourse version: 5.5.1
* Deployment type (BOSH/Docker/binary): any?

ddadlani:

# Existing Issue
Fixes #4467.

> What are the changes included in this PR? Feel free to add as many lines as needed below.

* add a DefaultGardenClientWithTimeout in `gclient`
* use the new garden client in the `worker` for the `container_sweeper`
* the baggageclaim client already has a timeout set

# Contributor Checklist

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/config.go
+++ b/atc/config.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	"golang.org/x/crypto/ssh"
 )
@@ -13,6 +14,9 @@ import (
 const ConfigVersionHeader = "X-Concourse-Config-Version"
 const DefaultPipelineName = "main"
 const DefaultTeamName = "main"
+
+// TIMEOUT CONSTANTS
+const GARDEN_CLIENT_HTTP_TIMEOUT = 5 * time.Minute
 
 type Tags []string
 
--- a/atc/worker/db_worker_provider.go
+++ b/atc/worker/db_worker_provider.go
@@ -4,6 +4,8 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/concourse/concourse/atc"
+
 	"code.cloudfoundry.org/clock"
 	"code.cloudfoundry.org/lager"
 	bclient "github.com/concourse/baggageclaim/client"
@@ -177,7 +179,7 @@ func (provider *dbWorkerProvider) NewGardenWorker(logger lager.Logger, tikTok cl
 		savedWorker.Name(),
 		savedWorker.GardenAddr(),
 		provider.retryBackOffFactory,
-		5*time.Minute,
+		atc.GARDEN_CLIENT_HTTP_TIMEOUT,
 	)
 
 	gClient := gcf.NewClient()
--- a/atc/worker/gclient/factory.go
+++ b/atc/worker/gclient/factory.go
@@ -70,3 +70,18 @@ func (gcf *gardenClientFactory) NewClient() Client {
 
 	return NewClient(NewRetryableConnection(connection.NewWithHijacker(hijackStreamer, gcf.logger)))
 }
+
+// Do not try any client method that requires hijack functionality (streaming logs)!
+func BasicGardenClientWithRequestTimeout(logger lager.Logger, requestTimeout time.Duration, address string) Client {
+	streamClient := &http.Client{
+		Timeout: requestTimeout,
+	}
+
+	streamer := &transport.WorkerHijackStreamer{
+		HttpClient:       streamClient,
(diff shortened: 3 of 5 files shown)


ddadlani:
If a container fails to delete, the gardenClient used should timeout the request after 5 minutes. 

This means that in future ticks, the sweeper will still try to delete the container. However, this is safe to do because guardian acquires a container lock to delete a container, and will return an error if we try to delete a container that is already being deleted.

https://github.com/cloudfoundry/garden/blob/f1dfd0e2a8c4dd9675f8a554ae757759afa40ca0/server/request_handling.go#L144-L147

cirocosta:
Hey,

Given that this is changing the behavior, it seems to me that we should be adding tests that at least falsifies "one can have a container deletion hanging forever" (assuming that's what we're trying to accomplish here) 🤔  

Wdyt?

thx!

kcmannem:
@cirocosta we have tested the timeout functionality previously with https://github.com/concourse/concourse/blob/master/atc/worker/gclient/client_stream_timeout_test.go

cirocosta:
so, it'd already not block forever? 🤔  I kinda lost on what this PR is fixing then 😬

ddadlani:
@cirocosta mentioned that we might want to test this better. Let's look into setting up a testing suite for this use case.

xtreme-sameer-vohra:
Adding a draft PR for tests here #4638

xtreme-sameer-vohra:
@ddadlani I tested it manually :)

PR #4583, opened 2019-10-08
Feature 4534 component tracking

### LINKED ISSUE #4534: Concourse deployment wide component tracking
The way concourse works right now, it will run each ATC component such as scheduler, lidar/radar, build tracker and GC once per tick **per ATC**. Meaning that if you have 4 ATCs, it will run GC 4 times every 30 seconds. This doesn't really make sense because each time the components run per tick, it will do a whole sweep of all the pipelines, builds or gc for the whole deployment, so running it 4 times within one interval (usually 30 seconds) does not make it much more effective in consideration to the huge load it adds. As a result, as users scale their ATCs horizontally, the more load they will be putting on their DB without much return. 

So in order to avoid this, we want to add deployment wide component tracking. This will ensure that each component will be only run once per interval tick.

A suggestion for the implementation is to have a table that has a column for `last_ran` and `enabled`. Each component will only run if `enabled` is true and the `last_ran` timestamp is less than the configured interval for that component. 

This change will also help the 6.0.0 migration where we need to batch migrate a whole new table, but we want to stop all components from running while we run that migration (#4214).

xtremerui:
# Existing Issue
Fixes #4534 

Added a table to track components last ran time and paused status. The interval of each component is seeded in DB during migration. While ATC starting up it will override those intervals with flag vars.

The runner (who runs the component) interval is default to 10s i.e. it will check if the component is paused every 10s. If component is unpaused it will proceed to check if the component run interval is reached. 

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


atc/lidar/interval_runner.go
```diff
@@ -56,30 +60,61 @@ func (r *intervalRunner) Run(signals <-chan os.Signal, ready chan<- struct{}) er
...
 	ctx, cancel := context.WithCancel(context.Background())
 
-	if err := r.runner.Run(ctx); err != nil {
-		r.logger.Error("failed-to-run", err)
-	}
-
 	for {
 		select {
 		case <-ticker.C():
+			component, _, err := r.componentFactory.Find(r.componentName)
```
cirocosta: Hey,

If I understood this right, this is inherintly racy in a way that would result
in the same behavior as before :thinking:

```
time 		atc1			atc2
0		find
1					find
2		oh not paused
3					oh not paused
4		interval elapsed
5					interval elapsed
6		runs
7					runs

```

in a way that we get two components running for the same interval.

Is that right? Or, am I missing another synchronization that might be happening
implicity somewhere else? 

Thanks!

atc/atccmd/command.go
```diff
@@ -76,6 +76,8 @@ import (
 var defaultDriverName = "postgres"
 var retryingDriverName = "too-many-connections-retrying"
 
+var runnerInterval = 10 * time.Second
```
cirocosta: nit: this could be a constant 😁

atc/component.go
```diff
@@ -0,0 +1,11 @@
+package atc
+
+const (
+	ComponentScheduler     string = "scheduler"
```
cirocosta: nit: the types are not really necessary

atc/db/component.go
```diff
@@ -0,0 +1,101 @@
...
+			return false, nil
+		}
+		return false, err
+	}
+
+	return true, nil
+}
+
+func (c *component) IntervalElapsed() bool {
+	inveral, _ := time.ParseDuration(c.interval)
```
cirocosta: I can imagine that the idea for avoiding checking the `err` is that "we were the ones who put it, thus, it's right", but, I wonder if we should check it anyway to make sure we don't get unexpected behavior that'd be very hard to narrow down.

I'd even consider a `panic(err)` if we'd like to not catch that parsing error in a higher level component

wdyt?

atc/builds/tracker_runner.go
```diff
@@ -24,14 +28,17 @@ type Notifications interface {
...
+	Notifications    Notifications
+	Interval         time.Duration
+	Clock            clock.Clock
+	Logger           lager.Logger
+	LockFactory      lock.LockFactory
+	ComponentFactory db.ComponentFactory
 }
 
 func (runner TrackerRunner) Run(signals <-chan os.Signal, ready chan<- struct{}) error {
+	close(ready)
```
ddadlani: why did we move the close to the top of the function? We think closing the ready chan signals to ifrit the runner is ready, but it looks like we're still doing some initializing after the call.

atc/builds/tracker_runner_test.go
```diff
@@ -62,85 +70,121 @@ var _ = Describe("TrackerRunner", func() {
...
 
-	It("tracks immediately", func() {
-		<-tracked
+	Context("when the interval elapses", func() {
+
+		JustBeforeEach(func() {
+			fakeClock.WaitForWatcherAndIncrement(interval)
+		})
+
+		Context("when the component is paused", func() {
```
ddadlani: we have this context (when the component is paused) in multiple contexts. Can we pull it out into a helper?

atc/lidar/interval_runner.go
xtremerui: There is a "task" (aka component) lock that already being using in `atc/lockrunner/runner.go` for gc, build-reaper and syslog-drainer. I think we should do the same for build-tracker and lidar scanner/checker as well. Will update the code accordingly.

atc/builds/tracker_runner_test.go
xtremerui: The reason these kind of logic all over each component (and thus the test) is the pipeline syncer (who spins off a scheduler for each pipeline). Since after we have algorithm-v3 merging in, pipeline syncer wil be removed and a runner for job level scheduler would be introduced, we would then be able to extract the component paused and interval elapsed logic to a centralize place. In that case the duplication will be gone.

So for now we would rather just let it be and hopelly clean it up as soon as algorithm-v3 is there.

atc/lidar/interval_runner.go
pivotal-jwinters: @xtremerui @cirocosta 

I don't think we need this extra locking because the underlying component does the locking itself: https://github.com/concourse/concourse/blob/master/atc/lidar/scanner.go#L43

pivotal-jwinters: I can see the argument for acquiring the lock first, but I'm really not worried about a race condition here. Plus there's no way to avoid that race for things like the build tracker, and resource checker which grab locks at a lower level for each individual build/check.

xtremerui: @pivotal-jwinters there is a lock here in lock runner 
https://github.com/concourse/concourse/blob/517bc6f3343bfbafa6c52b884f5b4bfced726d4c/atc/lockrunner/runner.go#L59

I'd say we could make them consistent at least? Could we remove that lock in this case?

pivotal-jwinters: Yeah the lock runner has its own lock because the underlying component doesn't do its own locking. I agree we should make this consistent. I like the idea of having all the locking in the runner, but that won't work for the build tracker and resource checker because they do locking at a lower level. Maybe each component should be responsible for its own locking and we get rid of the lock runner?

cirocosta: > I don't think we need this extra locking because the underlying component does the locking itself: https://github.com/concourse/concourse/blob/master/atc/lidar/scanner.go#L43

isn't that locking just the `run` though?

 I'm just afraid that locking just at that late stage would make us get to the situation that we had before where one would run immediately  after the other given that they could overlap the checking the interval + pause as those operations have no mutual exclusion (as shown in the ascii diagram) 🤔

pivotal-jwinters: @cirocosta it's locks all the way down.

> I guess we could lock on the entire component and then on each sub task, but that seems unnecessary.

After thinking about this more, I'm kinda backtracking on this. Maybe it is a good idea to lock at multiple levels. It's kinda like you said, lock the `run` then sub tasks can do their own locking if they need to.

pivotal-jwinters:
@ddadlani so what I said in https://github.com/concourse/concourse/pull/4583#issuecomment-541076851 no longer applies. 

Things changed in 3ada330 in order to address @cirocosta concerns over race conditions. Each component now holds a lock before running. Technically this mostly just moves locks around, but it does also introduce a top level lock for the build tracker and resource checker. It shouldn't introduce any of the problems you listed, since the lock isn't held for the duration of the `builds` or `checks` themselves.

atc/lidar/interval_runner.go
pivotal-jwinters: @cirocosta we now do the locking up front. This adds another layer of locking around certain components, but that seems like an ok tradeoff.

atc/builds/tracker_runner.go
pivotal-jwinters: just wanted to make it consistent with all the other runners

atc/db/migration/migrations/1569945021_add_components.up.sql
```diff
@@ -0,0 +1,31 @@
...
+      id serial PRIMARY KEY,
+      name text NOT NULL,
+      interval text NOT NULL,
+      last_ran timestamp WITH TIME ZONE,
+      paused boolean DEFAULT FALSE
+  );
+
+  CREATE UNIQUE INDEX components_name_key ON components (name);
+
+  INSERT INTO components(name, interval) VALUES
```
vito: Can/should we avoid this by turning the `UpdateIntervals` into an upsert/sync of all components? I didn't expect to maintain these as part of a migration. When we add a new migration would we need a migration to add the component?

atc/gc/build_log_collector.go
```diff
@@ -23,7 +23,7 @@ func NewBuildLogCollector(
 	batchSize int,
 	buildLogRetentionCalculator BuildLogRetentionCalculator,
 	drainerConfigured bool,
-) Collector {
+) *buildLogCollector {
```
vito: nit: Would you mind exporting these types? Having a public constructor return a private type has always felt odd to me. If we expose them it'll be easier to Godoc them in the future.

atc/gc/container_collector.go
```diff
@@ -27,7 +27,7 @@ func NewContainerCollector(
 	containerRepository db.ContainerRepository,
 	jobRunner WorkerJobRunner,
 	missingContainerGracePeriod time.Duration,
-) Collector {
+) *containerCollector {
```
vito: (same)

atc/gc/resource_cache_collector.go
```diff
@@ -11,7 +11,7 @@ type resourceCacheCollector struct {
 	cacheLifecycle db.ResourceCacheLifecycle
 }
 
-func NewResourceCacheCollector(cacheLifecycle db.ResourceCacheLifecycle) Collector {
+func NewResourceCacheCollector(cacheLifecycle db.ResourceCacheLifecycle) *resourceCacheCollector {
```
vito: (same)

atc/gc/resource_cache_use_collector.go
```diff
@@ -12,7 +12,7 @@ type resourceCacheUseCollector struct {
 	cacheLifecycle db.ResourceCacheLifecycle
 }
 
-func NewResourceCacheUseCollector(cacheLifecycle db.ResourceCacheLifecycle) Collector {
+func NewResourceCacheUseCollector(cacheLifecycle db.ResourceCacheLifecycle) *resourceCacheUseCollector {
```
vito: (same)

atc/gc/resource_config_check_session_collector.go
```diff
@@ -14,7 +14,7 @@ type resourceConfigCheckSessionCollector struct {
 
 func NewResourceConfigCheckSessionCollector(
 	configCheckSessionLifecycle db.ResourceConfigCheckSessionLifecycle,
-) Collector {
+) *resourceConfigCheckSessionCollector {
```
vito: (same)

atc/gc/resource_config_collector.go
```diff
@@ -11,7 +11,7 @@ type resourceConfigCollector struct {
 	configFactory db.ResourceConfigFactory
 }
 
-func NewResourceConfigCollector(configFactory db.ResourceConfigFactory) Collector {
+func NewResourceConfigCollector(configFactory db.ResourceConfigFactory) *resourceConfigCollector {
```
vito: (same)

atc/gc/volume_collector.go
```diff
@@ -19,7 +19,7 @@ type volumeCollector struct {
 func NewVolumeCollector(
 	volumeRepository db.VolumeRepository,
 	missingVolumeGracePeriod time.Duration,
-) Collector {
+) *volumeCollector {
```
vito: (same)

atc/gc/worker_collector.go
```diff
@@ -13,7 +13,7 @@ type workerCollector struct {
 	workerLifecycle db.WorkerLifecycle
 }
 
-func NewWorkerCollector(workerLifecycle db.WorkerLifecycle) Collector {
+func NewWorkerCollector(workerLifecycle db.WorkerLifecycle) *workerCollector {
```
vito: (same)

atc/lidar/interval_runner.go
```diff
@@ -29,25 +31,31 @@ func NewIntervalRunner(
 	runner Runner,
 	interval time.Duration,
 	notifications Notifications,
-	channel string,
+	componentName string,
+	lockFactory lock.LockFactory,
+	componentFactory db.ComponentFactory,
```
vito: neat that this is adopting useful component semantics

atc/lockrunner/runner.go
```diff
@@ -19,50 +17,100 @@ type Task interface {
...
-
-				err = lock.Release()
-				if err != nil {
-					lockLogger.Error("failed-to-release", err)
-					break
-				}
-			case <-signals:
-				return nil
-			}
+) *runner {
```
vito: This seems to have a lot of overlap with the interval runner. Do we need both? Is the only difference that this doesn't use a channel to short-circuit? (No big deal either way I guess.)

atc/pipelines/syncer.go
```diff
@@ -30,18 +32,36 @@ type runningPipeline struct {
...
 
 func (syncer *Syncer) Sync() {
+
+	component, _, err := syncer.componentFactory.Find(atc.ComponentScheduler)
+	if err != nil {
+		syncer.logger.Error("failed-to-get-component", err)
+		return
+	}
+
+	if component.Paused() {
```
vito: lol gross but there's a reason we're eliminating the syncer :+1:

atc/db/migration/migrations/1569945021_add_components.up.sql
pivotal-jwinters: yup good point, fixed in ef0fd2251332c7099be15e2e0498e87196dcfccd

atc/gc/build_log_collector.go
pivotal-jwinters: going to hold off on this for now as per our discussion

PR #4584, opened 2019-10-09
feat(fly): enable set-pipeline/validate-pipeline/execute's -c and -l options to take http/https URL.

evanchaoli:
# Why do we need this PR?

Current `fly set-pipeline/validate-pipeline/execute` can only be fed with a local files to `-c` and `-l` options. A user must have a local copy of YAML files in order to `set-pipeline`.

Consider the following use cases:

* I want to share a pipeline template YAML file to others. Otherwise have to copy the file to their local, which has to downsides: 1) not efficient, 2) I cannot ensure others not changing the file in their local.
* I want to `set-pipeline` from my other laptop, so I have to copy files to that laptop.

With this PR, `set-pipeline` may take a http/https URL as input files, so that:
* I can share a readonly URL to others
* Storing pipeline files to shared place, say git, then I can `set-pipeline` from any computer, only need to download a `fly` binary.

This PR is very important to my team. We are planning to build a set of pipeline templates, as a lot of CI templates are very similar, so that users may quickly setup their CI pipeline without understand details of Concourse and copy YAML files to their local.

Make `-c` and `-l` options can take http/https URL as input.

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/path_flag.go
+++ b/atc/path_flag.go
@@ -2,6 +2,8 @@ package atc
 
 import (
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"path/filepath"
 	"strings"
 
@@ -12,6 +14,11 @@ type PathFlag string
 
 func (path *PathFlag) UnmarshalFlag(value string) error {
 	if value == "" {
+		return nil
+	}
+
+	if isURL(value) {
+		*path = PathFlag(value)
 		return nil
 	}
 
@@ -33,7 +40,12 @@ func (path *PathFlag) UnmarshalFlag(value string) error {
 }
 
 func (path *PathFlag) Complete(match string) []flags.Completion {
-	matches, _ := filepath.Glob(match + "*")
+	var matches []string
+	if isURL(match) {
+		matches = []string{}
+	} else {
+		matches, _ = filepath.Glob(match + "*")
+	}
 	comps := make([]flags.Completion, len(matches))
 
 	for i, v := range matches {
@@ -41,4 +53,40 @@ func (path *PathFlag) Complete(match string) []flags.Completion {
 	}
 
 	return comps
+}
+
+func (path *PathFlag) ReadContent() ([]byte, error) {
+	var (
+		content []byte
+		err     error
+	)
+	if path.IsURL() {
+		res, err := http.Get(string(*path))
+		if err != nil {
+			return nil, err
+		}
+		defer res.Body.Close()
+		if res.StatusCode != http.StatusOK {
+			return nil, fmt.Errorf("%d: %s", res.StatusCode, http.StatusText(res.StatusCode))
+		}
+		content, err = ioutil.ReadAll(res.Body)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		content, err = ioutil.ReadFile(string(*path))
+		if err != nil {
+			return nil, err
+		}
+	}
+	return content, nil
+}
+
+func (path *PathFlag) IsURL() bool {
+	return isURL(string(*path))
+}
+
+func isURL(value string) bool {
+	value = strings.TrimSpace(value)
+	return strings.HasPrefix(value, "http://") || strings.HasPrefix(value, "https://")
 }
--- a/atc/path_flag_test.go
+++ b/atc/path_flag_test.go
@@ -1,0 +1,182 @@
+package atc_test
+
+import (
+	"fmt"
+	"github.com/concourse/concourse/atc"
+	"github.com/jessevdk/go-flags"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"os"
+)
+
+const content = "__content__"
+
+var _ = Describe("PathFlag", func() {
+	Describe("IsURL", func() {
+		It("return true when path is a http url", func() {
+			path := atc.PathFlag(" http://a.com ")
+			Expect(path.IsURL()).To(BeTrue())
+		})
+
+		It("return true when path is a https url", func() {
+			path := atc.PathFlag(" https://a.com ")
+			Expect(path.IsURL()).To(BeTrue())
+		})
+
+		It("return false when path is a file path", func() {
+			path := atc.PathFlag("a/b.txt")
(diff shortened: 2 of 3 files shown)


xtremerui:
First look at this it looks like the change affects more than just `set-pipeline` since concoruse use it for all other kind of file parsing e.g. `execute`, `validate-pipeline`. Is this PR targeting only `set-pipeline` or a bigger scope?

cirocosta:
Hi @evanchaoli ,

Thanks for submitting the PR!

Adding to the points raised by @xtremerui , it seems like [process substitution](https://www.gnu.org/software/bash/manual/html_node/Process-Substitution.html#Process-Substitution) already gives you what you want to accomplish, doesn't it?

```console
$ fly -t local set-pipeline -p mypipeline -c <(curl -SL https://something.com/pipeline.yml)
```

Thanks!

evanchaoli:
> ```
> $ fly -t local set-pipeline -p mypipeline -c <(curl -SL https://something.com/pipeline.yml)
> ```

@cirocosta My points:

* With my change, `-l` can also take variable files from remote.
* `curl` is not always installed on any Linux.
* Directly taking an URL is more user friendly than redirecting stdin.

evanchaoli:
> First look at this it looks like the change affects more than just `set-pipeline` since concourse use it for all other kind of file parsing e.g. `execute`, `validate-pipeline`. Is this PR targeting only `set-pipeline` or a bigger scope?

Our use cases only focus on `set-pipeline`. But to make this feature more generic, I can support big scope. I'll update the PR today.

evanchaoli:
@xtremerui I just checked code of `validate-pipeline` and `execute`, they all use `templatehelpers.YamlTemplateWithParams` to parse YAML files (both -c and -l), thus my change has applied to those two sub-commands as well. I also tested:

    $ /Users/chaol/concourse.fork/fly/fly -t dev vp -c http://engweb.<mycomany>.com/~chaol/simple.yml  -l http://engweb.<mycomany>.com/~chaol/cred.yml
    looks good

    $ /Users/chaol/concourse.fork/fly/fly -t dev execute -c http://engweb.<mycomany>.com/~chaol/simple.yml  -l http://engweb.<mycomany>/~chaol/cred.yml
    error: error unmarshaling JSON: while decoding JSON: json: unknown field "jobs"

The other sub-command `set-team` may also take a remote file, but I wants to focus on pipeline related commands in this PR.

ProvoK:
Hi @evanchaoli,
I've stumbled upon on this issue/PR and I would like to give my 2 cents.

I've encountered the same category of problems in other CLI/services/systems, and the best way was always "the linux way", that is _Do One Thing And Do It Well_.
In this case, other tools (eg: Gomplate, or even process substitution as @cirocosta outlined) are a natural fit and does not overcomplicate the fly CLI.

ProvoK:
> Hi @ProvoK, I appreciated the comment. However I still think `fly` to support URL is still helpful. `Gomplate` is a template template render tool, it also support URL based data source, why don't you think `Gomplate` should use "the linux way"?

Hi, touche'! 😄 
What I was trying to say is that, from my personal experience, features like this grows exponentially and become humongous. I feel that adding this will raise requests like "add http auth {{kind}}" and "support also {{some}} protocol" and more and more.
It's just my personal opinion, tough, and I **do not doubt** that it's useful per se!

I've quoted Gomplate because as a rendering tool, I've found it useful in rendering meta-pipelines.
Apart from that, Gomplate main raison d'etre is to render templates with <data> that could be fetched from multiple sources, which is a main feature of the tool. 

Still, I agree it's not the best example as a linux-well-made command (I've opened an issue about that some time ago)

vito:
I'm not a fan of this. :confused: Composability is a key aspect of Unixy CLIs. Not only does baking in functionality like this add more code to implement, test, and maintain, it can end up being a huge waste of engineering time and effort as we add more and more `curl` functionality (e.g. auth headers, timeouts, ...).

If my pipeline config, task config, pipeline vars, etc. lives at a URL I would just use `curl` to fetch it and then pass it to `fly`, probably with process substitution if it's a one-off command as @cirocosta noted.

If I don't have `curl`, I'll install it. And if I need to pass credentials, I can just use `curl -H "Authorization: ..."` instead of having to figure out whether `fly` supports auth headers.

This doesn't sound like a common enough workflow with a significant enough amount of pain to warrant the technical debt it would incur.

evanchaoli:
@vito I agree with spirit "composability", but I still want to add few points:

* What I'm adding is to enable both `-c` and `-l` options to take URLs. I don't see a "linux-way" to redirect two files into `fly`. Our our practice model, pipeline template comes from an URL, and part of variables (values of template) comes from the other URL, and secrets variables come from Vault. That's reason why we need this feature.
* I definitely agree we don't want to (and we don't have to) duplicate `curl` functions in `fly`. I believe 99% of URLs should be just simple HTTP GET. I'm just use a standard `Go` lib to to http get. If some user needs to handle complex HTTP things, then he can still use `curl` with the "linux-way". My change should not break the original "linux-way".
* `fly` also support Windows platform, which is not Unixy, though there are some solutions like `cygwin`. Will we suggest windows users to install `cygwin` just because he needs to get a pipeline yml file from remote? If I were the user, I would accept as `cygwin` is too heavy.

eedwards-sk:
> What I'm adding is to enable both -c and -l options to take URLs. I don't see a "linux-way" to redirect two files into fly

```
fly \
  -t foo \
  set-pipeline
    -p bar \
    -c <(curl --silent http://config.yaml) \
    -l <(curl --silent http://vars.yaml)
```

From an ops perspective, I agree with others in this thread -- downloading a config file is a step that I expect to occur outside of running a fly command.

Adding additional URI types is a slippery slope.

eedwards-sk:
> fly also support Windows platform, which is not Unixy, though there are some solutions like cygwin. Will we suggest windows users to install cygwin just because he needs to get a pipeline yml file from remote? If I were the user, I would accept as cygwin is too heavy.

There are plenty of windows solutions for downloading a file in a windows shell.

evanchaoli:
@eedwards-sk Thanks for the solution, I think that is good enough.

PR #4588, opened 2019-10-09
web: pin switching

### LINKED ISSUE #4561: web pin switching
Given I am on the resource page
And the resource is already pinned
And I am looking at a version in the versions list other than the pinned one
When I look at the pin button
I should be clickable

Given I am on the resource page
And the resource is already pinned
And I am looking at a version in the versions list other than the pinned one
When I click the pin button
Then a `PinResource` request should be made for the new version

This constitutes the web UI portion of #3522.

pivotal-jamie-klassen:
# Existing Issue

Fixes #4561.

* pin buttons on versions other than the pinned one are now clickable.
* clicking them will hit the PinVersion API appropriately
* there are reasonably-graceful visual transition states (spinners on both versions being switched between).

# Contributor Checklist
- [X] Unit tests
- [X] ~Integration tests (if applicable)~
- [X] ~Updated documentation (located at https://github.com/concourse/docs)~

# Reviewer Checklist
- [x] Code reviewed
- [x] Tests reviewed
- [x] Documentation reviewed
- [x] Release notes reviewed
- [x] PR acceptance performed

### DIFF
--- a/release-notes/latest.md
+++ b/release-notes/latest.md
@@ -19,9 +19,9 @@
 
 * All API payloads are now gzipped. This should help save bandwidth and make the web UI load faster. #4470
 
-#### <sub><sup><a name="4448" href="#4448">:link:</a></sup></sub> feature
+#### <sub><sup><a name="4448-4588" href="#4448-4588">:link:</a></sup></sub> feature
 
-* You can now use fly to pin a resource without unpinning it first. #4448
+* You can now pin a resource to different version without unpinning it first #4448, #4588.
 
 #### <sub><sup><a name="4507" href="#4507">:link:</a></sup></sub> fix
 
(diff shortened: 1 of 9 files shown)


xtremerui:
When there is a new version, after page auto refresh the new version shows on top while the last version got pushed to nowhere and disapeared. It will show up until a manual page fresh.

captured

Upon checking it is introduced in this PR only. Another interested behaviour is the first three versions act normal until there is a fouth one.

pivotal-jamie-klassen:
@xtremerui I think the root cause of your observation is this unfortunate copypasta: https://github.com/concourse/concourse/blob/master/web/elm/src/Network/Pagination.elm#L66-L73. During the course of investigating this we discovered the cause of some strange behaviours on the resource page as determined by the [permalink](https://github.com/concourse/concourse/blob/master/web/elm/src/Resource/Resource.elm#L710-L721) function, such as:
* if, at any point, the `ListVersions` api call returns no versions, all subsequent requests on autorefresh will be made with the query parameters `since=0&limit=100`
* otherwise, all autorefresh requests will be made with pagination parameters designed to always return the same list of versions (i.e. `from` the id of the first version returned and `limit` of the length of the list)

PR #4593, opened 2019-10-10
Debug helper script for the docker-compose rig.

radoslav-tomov:
Signed-off-by: Radoslav Tomov radko87@gmail.com

# Why do we need this PR?

It improves the dev experience by streamlining the setup of a debugging session when working with the local docker-compose setup.  

Added helper script, which will attach a debugger to either web or worker processes and bind to port  :2345 .

Example
` docker-compose up `
`./hack/debug web`

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


cirocosta:
Hi @radoslav-tomov , 

Thanks for submitting the PR!

We have [`hack/trace`](https://github.com/concourse/concourse/blob/ba5a5932f1921fd8c2d22d0246dae2af3a6abd78/hack/trace) which seems to achieve very similar goals that this PR does.

Why would we need this one specifically? Is it because it'd allow the usecase of attaching an IDE or something else to the session created within the container?

Also, are there any gotchas with doing that? e.g., symbol resolution, etc?

Thanks!

radoslav-tomov:
Hi @cirocosta, 

The change has been inspired by the hack/trace thus is almost identical.   It is useful for establishing a debug session and attaching IDE debugger. 

I for one use GoLand and have created 2 debug configurations: one for the web process and one for the worker. I have added extra configuration to execute the helper before the launch of the debug session. This way I have a single click experience if I want to launch a debug session.

Not sure about gotchas. Maybe the only one is that you can debug a single process at a time, as the port is hardcoded.

cirocosta:
oh, interesting! thanks for sharing!

<br />

A few of us here use Goland too, I can imagine how useful that can be 😁  

As the structure is mostly the same, aside from a little difference in behavior, wdyt of having an argument for `hack/trace` that one could provide to change the behavior (from an interactive session to opening the port)?

E.g., `hack/trace --listen`? In which case, it could then provide the extra arguments to start the listening 🤔  

Oh, and a small thing - it'd be nice to have [CONTRIBUTING.md](./CONTRIBUTING.md) updated too (as we document the current behavior of `hack/trace` there).

Wdyt?

thanks!

radoslav-tomov:
Hi @cirocosta, 

It sounds like a good idea to me and I would take it a step further by renaming the `hack/trace` to `hack/dlv`. I think this would remove the ambiguous term `trace` and make it much more clear to potential users.  What do you think about it?

Regards,
Rado

radoslav-tomov:
@cirocosta I've adopted your feedback. Can you check it out?

CONTRIBUTING.md
```diff
@@ -190,6 +190,27 @@ To trace a running worker instance:
...
+
+```sh
+$ ./hack/trace worker --listen 2345
+```
+
+After this is done, the final step is to connect your IDE to the debugger with the following parameters:
+* host: `localhost`
+* port: `2345` 
+
+For Go Land you can do so by going to Run | Edit Configurations… | + | Go Remote and fill in the parameters.
```
zoetian: Do you mind changing it to `GoLand` without the space? 😁

radoslav-tomov: @zoetian My bad. Done.

hack/trace
```diff
@@ -31,4 +42,4 @@ docker run \
   --rm \
   --tty \
   dlv \
-  attach $trace_pid
+  attach $trace_pid $listen_listen_cmd
```
cirocosta: whops, I think you meant `$listen_cmd`:

```suggestion
  attach $trace_pid $listen_cmd
````

**Comment by cirocosta** [comment]
Hi @radoslav-tomov , sorry for the super long delay - just a small change and then we're good to go 😁  

Thank you!

**Comment by radoslav-tomov** [comment]
@cirocosta good catch. I've had it fixed in my local branch, but most likely I missed pushing it. :d rookie mistake.

**Comment by cirocosta** [comment]
hahah, np! thank you!!

**Comment by huguesalary** [comment]
I'm using docker for mac and this doesn't seem to work on my end.

Running `./hack/trace worker --listen 2345` correctly builds and run a docker container running `dlv`, however, the container itself isn't listening to external connections, that is, `docker ps` shows this:

```
CONTAINER ID        IMAGE                       COMMAND                  CREATED             STATUS              PORTS                                            NAMES
07f1a7f0e0de        dlv                         "dlv attach 7 --head…"   9 seconds ago       Up 7 seconds                                                         musing_lichterman
0ef2217ae4da        concourse/concourse:local   "dumb-init concourse…"   3 minutes ago       Up 3 minutes        0.0.0.0:7777->7777/tcp, 0.0.0.0:7788->7788/tcp   concourse_worker_1
3816918da538        concourse/concourse:local   "dumb-init concourse…"   3 minutes ago       Up 3 minutes        0.0.0.0:8080->8080/tcp                           concourse_web_1
fa6be2e83ea4        postgres                    "docker-entrypoint.s…"   3 minutes ago       Up 3 minutes        0.0.0.0:6543->5432/tcp                           concourse_db_1
```

As you can see, there's no `PORTS` for the `dlv` container, which means I can not connect to the remote `dlv` process.

huguesalary:
I'm really bad at bash, but all that needs to be done is modify `hack/trace` and add the `-p $dlv_port:$dlv_port` to the `docker run`

pivotal-jamie-klassen:
@huguesalary lol good point: #4667

Issue #4595, opened 2019-10-10
web: simplify resource page pagination


pivotal-jamie-klassen:
As discussed [here](https://github.com/concourse/concourse/pull/4588#issuecomment-540635963), the resource page behaves in a few surprising ways with regards to version pagination. Let's propose something simpler here. In an ideal world there are two main user stories:
1. When I am looking at the resource page with no particular URL query parameters, I want to always up to 100 of the latest versions, and when the page autorefreshes I want new versions to flow in, potentially pushing down or causing older versions to disappear.
2. When I am looking at a specific page of versions (by visiting a URL with query parameters or by clicking the pagination chevrons), I expect the current URL to always show the exact same list of versions that I'm currently looking at.

As it stands we have two realities that make 2. above impossible: We specify pagination by anchoring the beginning or end of the page at a specific version id together with the page size, and the underlying list of versions can re-order depending on what gets returned from checks (@clarafu implemented this behaviour around the time global resources were introduced, as I recall).

The simplest proposal we can think to make right now would be not to have "permanent" pagination -- if you want to visit a specific page of versions you have to click your way there every time and the `since`, `until`, `from`, `to`, and `limit` query parameters to the resource page URL go away.

vito:
> and the underlying list of versions can re-order depending on what gets returned from checks (@clarafu implemented this behaviour around the time global resources were introduced, as I recall).

I don't think we should focus on this aspect when designing the UX as it's not a behavior we *expect* of resources. They're meant to have an immutable, stable ordering, so I don't think it's wrong for our UI to reflect that assumption.

I don't feel strongly about the rest, I just want to make sure our motivations are coming from the right place. :slightly_smiling_face:

PR #4596, opened 2019-10-10
[WIP] fix pagination params when building url

jomsie:
# Why do we need this PR?
Resource pagination doesn't exactly work as intended. The `to`, `from`, and `until` parameters were being overwritten with `limit`. That would lead to urls like `Request URL: https://ci.concourse-ci.org/api/v1/teams/main/pipelines/concourse/resources/concourse/versions?limit=79&limit=60846285`. Which just doesn't make sense!

# Contributor Checklist

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/release-notes/latest.md
+++ b/release-notes/latest.md
@@ -26,3 +26,7 @@
 #### <sub><sup><a name="4507" href="#4507">:link:</a></sup></sub> fix
 
 * @iamjarvo fixed a [bug](444://github.com/concourse/concourse/issues/4472) where `fly builds` would show the wrong duration for cancelled builds #4507.
+
+#### <sub><sup><a name="4596" href="#4596">:link:</a></sup></sub> fix
+
+* Pagination on the resources page will now properly respect the `to`, `from`, and `until` URL parameters #4596.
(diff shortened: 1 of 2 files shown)


xtremerui:
While I open the resource page starting from fresh and observe all incoming versions showing up, everything is good. And I notice the pagination params are `limit=100&since=0`.

However since I've got any number of versions later say 3 (version ids are 1, 3, 5), and I refresh the page, I notice the params become `limit=3&since=5`, which causing the version history stuck at showing 3 versions and no new version would able to auto show up. I think the logic of calculation `limit` is somehow wrong.

pivotal-jamie-klassen:
@xtremerui that is essentially [this observation](https://github.com/concourse/concourse/pull/4588#issuecomment-540635963) which gave rise to #4595 - looking in the code, the behaviour you described seems very much intended so we should zoom out and reassess.

jomsie:
i'm going to close this for now, can follow up in #4595

PR #4603, opened 2019-10-15
web: add re-run button to build page

### LINKED ISSUE #4589: web: add re-run button to build page
Before committing to the layout change proposed by @Lindsayauchin in #4222, let's first add the button at the right-hand side for non-running builds. This means that you will always see two buttons on the right-hand side of the build header, the leftmost of which will either be an abort button or a re-run button, depending on whether or not the current build is running.

In the happy path, clicking the button will send a POST request to the API route corresponding to the current build.

For now, assume this button behaves similarly to the abort button: it always appears clickable, and if the associated API call returns a 401 the browser redirects to login, and 403s are swallowed.

The icon on this button should be https://github.com/concourse/concourse/files/3498560/re-trigger_bttn.svg.zip, and the button's background color should match the status of the current build.

Later iterations may involve a two-click rerunning flow, but for now the click will just directly re-run. Let's think about a tooltip that describes what this button does in words - maybe "re-run with the same inputs"

pivotal-jamie-klassen:
# Existing Issue

Fixes #4589.

* re-run button appears on the right for non-running builds
* button has explanatory tooltip
* (tangentially related - we changed the 'manual triggering disabled in job config' tooltip to match styles)
* if you re-run a cancelled or pending build, it will silently give a 500 error
Screenshot:
Screen Shot 2019-12-16 at 4 54 26 PM

# Contributor Checklist
- [X] Unit tests
- [X] ~Integration tests (if applicable)~
- [x] Updated documentation: concourse/docs#262

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


zoetian:
@concourse/core-api rebased and everything still works. @Lindsayauchin @matthewpereira @vito the UX is not ideal when attempting to rerun a cancelled build (silent 500 error), but nothing breaks.

pivotal-jamie-klassen:
This is properly ready for review now - I expect the tests to pass, and you can hit `R` to rerun a build from  the keyboard! This is described in the keyboard help overlay on the build page as well.

PR #4604, opened 2019-10-15
[WIP] 1038 clear resource cache

### LINKED ISSUE #1038: Removing cached volumes

ProvoK:
Hi there,
after some time I understood a bit the flow between **fly** and **API server**, and its components.
Before digging into unit tests, query(ies) and so on, I'd like to know if I'm on the right track, in order to avoid duplication of effort later on.

# Existing Issue

Closes #1038 .

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/api/resourceserver/clear_cache.go
+++ b/atc/api/resourceserver/clear_cache.go
@@ -7,12 +7,27 @@ import (
 	"github.com/concourse/concourse/atc/db"
 )
 
-func (s *Server) ClearResourceCache(pipeline db.Pipeline) http.Handler {
+func (s *Server) ClearResourceCaches(pipeline db.Pipeline) http.Handler {
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		resourceName := r.FormValue(":resource_name")
 
-		clearedVersions := []string{"v1", "v2", resourceName}
+		clearedVersions := []string{resourceName, "v1"}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+
+		json.NewEncoder(w).Encode(clearedVersions)
+	})
+}
+
+func (s *Server) ClearResourceCache(pipeline db.Pipeline) http.Handler {
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		//resourceName := r.FormValue(":resource_name")
+		resourceConfigVersionID := r.FormValue(":resource_config_version_id")
+
+		clearedVersions := []string{resourceConfigVersionID}
 
 		w.Header().Set("Content-Type", "application/json")
 		w.WriteHeader(http.StatusOK)
--- a/atc/api/resourceserver/clear_cache.go
+++ b/atc/api/resourceserver/clear_cache.go
@@ -1,0 +1,22 @@
+package resourceserver
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/concourse/concourse/atc/db"
+)
+
+func (s *Server) ClearResourceCache(pipeline db.Pipeline) http.Handler {
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		resourceName := r.FormValue(":resource_name")
+
+		clearedVersions := []string{"v1", "v2", resourceName}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+
+		json.NewEncoder(w).Encode(clearedVersions)
+	})
+}
--- a/atc/routes.go
+++ b/atc/routes.go
@@ -30,7 +30,8 @@ const (
 	JobBadge       = "JobBadge"
 	MainJobBadge   = "MainJobBadge"
 
-	ClearTaskCache = "ClearTaskCache"
+	ClearTaskCache     = "ClearTaskCache"
+	ClearResourceCache = "ClearResourceCache"
 
 	ListAllResources     = "ListAllResources"
 	ListResources        = "ListResources"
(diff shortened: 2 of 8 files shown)


atc/routes.go
```diff
@@ -161,6 +163,9 @@ var Routes = rata.Routes([]rata.Route{
 	{Path: "/api/v1/teams/:team_name/pipelines/:pipeline_name/resources", Method: "GET", Name: ListResources},
 	{Path: "/api/v1/teams/:team_name/pipelines/:pipeline_name/resource-types", Method: "GET", Name: ListResourceTypes},
 	{Path: "/api/v1/teams/:team_name/pipelines/:pipeline_name/resources/:resource_name", Method: "GET", Name: GetResource},
+	{Path: "/api/v1/teams/:team_name/pipelines/:pipeline_name/resources/:resource_name/cache", Method: "DELETE", Name: ClearResourceCaches},
+	{Path: "/api/v1/teams/:team_name/pipelines/:pipeline_name/resources/:resource_name/cache/:resource_config_version_id", Method: "DELETE", Name: ClearResourceCache},
```
vito: nit: should probably be `/caches/`

atc/api/resourceserver/clear_cache.go
```diff
@@ -0,0 +1,37 @@
...
+
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		resourceName := r.FormValue(":resource_name")
+
+		clearedVersions := []string{resourceName, "v1"}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+
+		json.NewEncoder(w).Encode(clearedVersions)
```
vito: Returning the # of caches cleared would be nice; here's how the task cache clearing endpoint returns it:

https://github.com/concourse/concourse/blob/0e1b195bf964cf11e88c6cbd75775f5160cce1df/atc/api/jobserver/clear_task_cache.go#L58

fly/commands/clear_resource_cache.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/fly/commands/internal/flaghelpers"
+	"github.com/concourse/concourse/fly/rc"
+	"github.com/concourse/concourse/fly/ui"
+	"github.com/fatih/color"
+)
+
+type ClearResourceCacheCommand struct {
+	Resource flaghelpers.ResourceFlag `short:"r" long:"resource-type" required:"true" value-name:"PIPELINE/RESOURCE-TYPE" description:"Name of a resource to invalidate"`
```
vito:
```suggestion
	Resource flaghelpers.ResourceFlag `short:"r" long:"resource" required:"true" value-name:"PIPELINE/RESOURCE" description:"Resource to clear the cache from"`
```

fly/commands/clear_resource_cache.go
```diff
@@ -0,0 +1,57 @@
...
+	Version  *atc.Version             `short:"f" long:"from" required:"false" value-name:"VERSION" description:"Version of the resource cache to invalidate, e.g. digest:sha256@..."`
```
vito:
```suggestion
	Version  *atc.Version             `long:"version" required:"false" value-name:"VERSION" description:"Version of the resource cache to clear, e.g. ref:abcdef..."`
```

I removed the short form as there's ambiguity with `-v` for `--version`. I think in the long run we should redesign the CLI to take these values as arguments so we don't have so many overlapping short flags. :thinking: But that's a topic for another day. For now let's just have the long flag name.

atc/routes.go
clarafu: I think taking in a `resource_config_version_id` for the ClearResourceCache endpoint might not be worth the trouble because in the database, a resource cache points to a specific version and not a `resource_config_version_id`. When a user runs the fly command for clearing their resource cache, they pass in a version and then the fly command would have to convert that version to a `resource_config_version_id`, call the ClearResourceCache endpoint with it and then the endpoint would have to convert that id back to a version and use that to find the resource cache to delete. 

To that point, it might be easier to just have the `ClearResourceCaches` endpoint be able to also take in a `version` as a param. That way we can straight up use the version passed in from the user to find the resource cache associated to it. In this case, we would also not need to add a `ClearResourceCache` endpoint and just need `ClearResourceCaches`.

atc/routes.go
```diff
@@ -161,6 +163,9 @@ var Routes = rata.Routes([]rata.Route{
 	{Path: "/api/v1/teams/:team_name/pipelines/:pipeline_name/resources", Method: "GET", Name: ListResources},
 	{Path: "/api/v1/teams/:team_name/pipelines/:pipeline_name/resource-types", Method: "GET", Name: ListResourceTypes},
 	{Path: "/api/v1/teams/:team_name/pipelines/:pipeline_name/resources/:resource_name", Method: "GET", Name: GetResource},
+	{Path: "/api/v1/teams/:team_name/pipelines/:pipeline_name/resources/:resource_name/cache", Method: "DELETE", Name: ClearResourceCaches},
```
clarafu: One thing to keep in mind might be that in order to clear a resource cache, it has to not be used by anything. If we run a command to clear all the resource caches for a resource, that most likely will end up erroring because something is using a cache (for example, a build is currently running and using a cache that we are trying to delete).

I think it would be fine to error and have the fly command tell the user that something is currently using a cache that it is trying to delete, since we shouldn't be pulling out caches underneath builds that are using it. 🤔

clarafu:
@ProvoK Just wanted to know if you are still interested in working on this PR? We would love to collaborate with you to get it finished and polished! :)

clarafu:
Closing due to inactivity, please feel free to open a new PR if you ever get the chance to work on it again. Thanks!

PR #4605, opened 2019-10-15
Fix resource types tree lookup when resource type name and type are the same

### LINKED ISSUE #4599: Error when lidar enabled and resource and resource type have the same name
## Bug Report

In Concourse 5.6.0 with lidar enabled, resources error when named the same as a resource type.

```
checking failed
base resource type not found: slack
```

## Steps to Reproduce

Create a pipeline with a resource and resource type named the same:

```yaml
resource_types:
- name: slack
  type: registry-image
  source:
    repository: cfcommunity/slack-notification-resource
resources:
- name: slack
  type: slack
jobs:
- name: test
  plan:
  - put: slack
```

## Version Info

* Concourse version: 5.6.0 + lidar enabled
* Did this used to work? yes, works pre 5.6

This looks like the source of the problem, but I don't understand the context well enough to propose a fix: https://github.com/concourse/concourse/blob/master/atc/db/resource_type.go#L60

xtremerui:
fixes #4599

### DIFF
--- a/atc/db/resource_type.go
+++ b/atc/db/resource_type.go
@@ -57,7 +57,7 @@ type ResourceTypes []ResourceType
 func (resourceTypes ResourceTypes) Parent(checkable Checkable) (ResourceType, bool) {
 	for _, t := range resourceTypes {
 		if t.PipelineID() == checkable.PipelineID() {
-			if t.Name() != checkable.Name() && t.Name() == checkable.Type() {
+			if t != checkable && t.Name() == checkable.Type() {
 				return t, true
 			}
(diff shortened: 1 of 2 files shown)


xtremerui:
> Any way a test could be added for this?

@vito there is test that covers it already. https://github.com/concourse/concourse/blob/ca0d170a665a7e51c00831cbd450e963cf9b1e49/atc/db/resource_type_test.go#L171-L173

pivotal-jamie-klassen:
@pivotal-jwinters needs release note

PR #4616, opened 2019-10-18
detects inputs to a put step

### LINKED ISSUE #2692: Detect inputs to a put step
With #1202 implemented, it is sort of redundant to need to type out all the artifact inputs to a put step while they "most likely" have already written out the artifacts that are needed in the params. The only case we can think of for now that would not have the artifacts specified in the put params is when you would use a wildcard, which is covered by the "all" option for the inputs to a put configuration. 

We can add a "detect" option to the input configuration for a put step that will go through the params and grab out a list of all the artifacts that were specified within it and use those as the inputs. It might look something like this: 

```
- get: some-repo
- put: some-resource
  inputs: detect
  params:
    doesnt_matter: some-repo/hello
```

In this example we would take the `some-repo` from `some-repo/hello` and deduce that `some-repo` is an artifact to pull in as an input. This won't always correspond to an artifact, and in that case it should just be skipped. Note that the value `some-repo` should also be used to look up an artifact.

julia-pu:
Closes #2692 

Added "detect" option for a put step to detect artifacts in the params and use those as inputs instead of having to list all of the specific inputs.


atc/exec/put_inputs.go
```diff
@@ -70,6 +71,45 @@ func (i specificInputs) FindAll(artifacts *artifact.Repository) ([]worker.InputS
...
+	inputs []string
+}
+
+func NewDetectInputs(params map[string]interface{}) PutInputs {
+	var inputs []string
+	for _, value := range params {
+		switch actual := value.(type) {
+		case string:
+			input := actual
+			if idx := strings.IndexByte(actual, '/'); idx >= 0 {
```
ddadlani: do we also want to test this functionality, i.e. stripping slashes?

Also is there a reason we do it this way rather than strings.TrimSuffix?

cirocosta: I'd also add that it might be worth considering `path/fiilepath#SplitLiist` too 😁  https://golang.org/pkg/path/filepath/#SplitList (as we're dealing with file paths here)

vito: @ddadlani It's not stripping slashes here, the idea is to grab the first path segment, i.e. `foo` in `foo/bar/baz`, to identify (guess) the input name.

ddadlani: ah ok, thanks. That seems like it's definitely part of the functionality then, since we're coming up with an (expected) identifier. Having a test for it would help give context about that.

atc/exec/put_inputs.go
```diff
@@ -70,6 +71,45 @@ func (i specificInputs) FindAll(artifacts *artifact.Repository) ([]worker.InputS
...
+				input = actual[:idx]
+			}
+			inputs = append(inputs, input)
+		default:
```
clarafu: There are two more cases to handle:

```go
map[string]interface{}
[]interface{}
```

e.g.:

```yaml
params:
  globs: [foo/*.tgz]
  some_map: {key: bar/*.tgz}
```

And this can go n levels deep, so you might need to sprinkle some recursion on this logic!

vito:
Haha woops we should probably merge this. Adding to the 6.0 milestone.

/cc @clarafu

Issue #4457, opened 2019-09-18
Step runtime tooltip is hidden by build header bar


evanchaoli:
## Bug Report

See the screen shot:
<img width="621" alt="tooltip-is-hiden" src="https://user-images.githubusercontent.com/18585861/65119910-6b447280-da1f-11e9-987f-e44e6fbb66cb.png">

The first step's tooltip is hidden by the build header bar. We have noticed this issue since Concourse 5.2 (5.2 is our earliest version).

## Steps to Reproduce

Put the mouse cursor on the first step's right icon, and a tooltip should pop up and it would be hidden by header bar.

## Expected Results

Tooltip should be not hidden.

## Actual Results

Tooltip is hidden.

## Additional Context

This issue is across all Concourse versions since at least 5.2.

## Version Info

* Concourse version: 5.2-5.5
* Deployment type (BOSH/Docker/binary): K8S
* Infrastructure/IaaS: K8S
* Browser (if applicable): Chrome
* Did this used to work? No

pivotal-jamie-klassen:
related #3921. In general, I feel that the approach taken by the [sidebar](https://github.com/concourse/concourse/blob/master/web/elm/src/SideBar/Styles.elm#L254-L262) (the key part being `position: fixed`) probably scales to the greatest number of tooltip scenarios. While #4206 was a promising hack, it had some [unintended consequences](https://github.com/concourse/concourse/issues/4210) so the fix here will probably mean rolling up one's sleeves and wading into [some](https://github.com/concourse/concourse/blob/master/web/elm/src/SideBar/SideBar.elm#L67-L68) [complicated](https://github.com/concourse/concourse/blob/master/web/elm/src/Message/Effects.elm#L434-L436) [event](https://github.com/concourse/concourse/blob/master/web/elm/src/SideBar/SideBar.elm#L107-L115) [handling](https://github.com/concourse/concourse/blob/master/web/elm/src/Message/Effects.elm#L438-L440) [logic](https://github.com/concourse/concourse/blob/master/web/elm/src/SideBar/SideBar.elm#L120-L133) on the build page.

PR #4630, opened 2019-10-22
web: first-occurrence tooltip no longer hidden

### LINKED ISSUE #3921: tooltip for "new version" in build view goes under topbar
## Bug Report

When the first line of a build in the view is a new resource, the tooltip goes under the topbar. Same for tooltip with durations

## Steps to Reproduce

Go to a build with a new version of a resources that is used as the first step of a job. Mouse over the yellow arrow

## Expected Results

I can read the tooltip

## Actual Results

The tooltip is hidden under the topbar

## Additional Context

tooltip-topbar

It looks like it's a CSS problem with the overflow behaviour that has changed with the addition of the sidebar, but I'm not sure how to fix it

## Version Info

* Concourse version: 5.2
* Infrastructure/IaaS: kubernetes
* Browser (if applicable): Safari & Firefox
* Did this used to work? yes

pivotal-jamie-klassen:
# Existing Issue

Fixes #3921.

* first-occurrence tooltip is at the root of the DOM
* position:fixed and large z-index
* admittedly the fixed positioning is cumbersome, but it's immune to effects
  from restructuring any other part of the page
* sets a precedent for future tooltips to be handled by the Tooltip module

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


jomsie:
> seems legit
> 
> can't attach a screenshot since the tooltip goes away
> 
> oh well

a fun thing i learned when trying to debug a tooltip, in the `sources` tab of the chrome inspector, pressing f8 will auto breakpoint!

Screen Shot 2019-10-25 at 10 51 48 AM

pivotal-jamie-klassen:
You can also `yarn build-debug` and you'll get the elm debugger!

evanchaoli:
@pivotal-jamie-klassen Do you have a plan to fix #4457?

pivotal-jamie-klassen:
@evanchaoli it's not prioritized in the [UX roadmap](https://github.com/concourse/concourse/projects/21#column-6074000) but I'd definitely like to see it fixed as it's totally annoying. The reason I didn't want to undertake it right away is because the other tooltips basically involve text that can be derived directly from the [`DomID`](https://github.com/concourse/concourse/blob/master/web/elm/src/Message/Message.elm#L46-L80) value that shows up in the [`Session`](https://github.com/concourse/concourse/blob/master/web/elm/src/Application/Models.elm#L9-L19)'s [`.hovered`](https://github.com/concourse/concourse/blob/master/web/elm/src/Tooltip.elm#L14) property, whereas those duration tooltips involve data that depends on having received a bunch of build events, so there's a bit of refactoring required.

Here's my best guess at a good approach:

I think we'd have to expand the [`HoverState`](https://github.com/concourse/concourse/blob/master/web/elm/src/HoverState.elm#L15-L19) type to include some kind of variable content beyond just the `DomID` of the thing that is hovered, and we'd probably need to communicate the intended contents of the tooltip via the [`GetViewport`](https://github.com/concourse/concourse/blob/master/web/elm/src/Message/Effects.elm#L171) effect. Looking a little deeper I guess we could extend the `AlwaysShow` variant of [`TooltipPolicy`](https://github.com/concourse/concourse/blob/master/web/elm/src/Message/Callback.elm#L63-L65) to include data of a type something like
```
type TooltipContents = Text String | Table (List (String, String))
```
or even something more application-specific like
```
type TooltipContents = Text String | Duration {initialization : Timespan, step: Timespan}
```
(where `Timespan` is inspired from [`Build.Header.Views`](https://github.com/concourse/concourse/blob/master/web/elm/src/Build/Header/Views.elm#L71-L75))

All of that refactoring done, I think I'd modify [this test](https://github.com/concourse/concourse/blob/master/web/elm/tests/BuildTests.elm#L2971-L3015) to assert that a `GetViewport` message with the right `TooltipPolicy` (including the data durations) got triggered. At that point, driving the implementation in the `Tooltip` module and adding an extra case in the [`view` function](https://github.com/concourse/concourse/blob/master/web/elm/src/Tooltip.elm#L68-L90) seems like it would be pretty straightforward, in my estimation.

Any chance you or your team want to pursue this? Maybe @Twiknight has some spare cycles? 😝

evanchaoli:
> Any chance you or your team want to pursue this? Maybe @Twiknight has some spare cycles? 

@pivotal-jamie-klassen We'd like to help as in theory we are same company. Unfortunately @Twiknight is fully loaded recently.

PR #4591, opened 2019-10-09
ensure failed containers are cleaned from the workers

kcmannem:
Signed-off-by: Krishna Mannem <kmannem@pivotal.io>
Co-authored-by: Divya Dadlani <ddadlani@pivotal.io>

# Existing Issue
Fixes #4562 

# Changes proposed in this pull request
* when gc-ing failed containers, mark them as destroying rather than simply deleting the row
* this will ensure that if they failed due to a timeout on the gardenClient.Create call, the resulting garden container will be picked up by the sweeper.
* This should not cause any errors when deleting containers in Garden because the list of destroying containers depends upon the list of containers that the worker reports

# Contributor Checklist
- [X] Unit tests
- [ ] Updated documentation (located at https://github.com/concourse/docs)
- [ ] Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

PR #4635, opened 2019-10-23
Release 5.5.x pr 4591

pivotal-jwinters:
Backport #4591 to 5.5.x release branch


pivotal-jwinters:
@pivotal-jamie-klassen halp!

are those build failures the same yarn issue from yesterday? Is there a task cache that needs to be cleared?

PR #4617, opened 2019-10-19
Add new relic log info msg for non 2xx responses

xtreme-sameer-vohra:
# Why do we need this PR?
If New Relic is configured and the API responds with a non-2xx status code (eg. 403), no log messages are emitted in that event. 
It would be useful for the operator to be aware that the metrics weren't successfully emitted.

Steps to reproduce
- create a free account on New Relic
- generate an API key and configure Concourse
- the API request to New Relic fails with a 403, however, the error isn't logged

# Changes proposed in this pull request
Checks the API response's status code and logs messages that are outside 200-299 status codes.

# Contributor Checklist
> Are the following items included as part of this PR? Please delete checkbox items that don't apply.
- [ ] Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)

# Reviewer Checklist
> This section is intended for the core maintainers only, to track review progress. Please do not
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

PR #4641, opened 2019-10-23
metrics: batch newrelic emitter

### LINKED ISSUE #4540: Newrelic metrics emitter losing most events
## Bug Report

`CONCOURSE_METRICS_BUFFER_SIZE` does not have an effect with the Newrelic emitter. 

## Steps to Reproduce

Given a busy concourse deployment that emits metrics to Newrelic, the metrics are dropped after metrics buffer is full; **and it gets full in few seconds**. For example, even after I bump the buffer size to 7000 (default is 1000), the number of dropped messages will be the same (calculated that by counting the number metrics error messages in the logs).

I think the problem with this flag is that it gives a false sense of improving metrics collection. **I believe that If this flag is not combined with some kind of metrics batching, it will not have an effect on the metrics throughput.**

## Expected Results

Much fewer metrics being dropped.

## Actual Results

Most metrics are being dropped.

## Version Info

* Concourse version: 5.5.1
* Infrastructure/IaaS: Azure
* Did this used to work?: No

zoetian:
# Existing Issue
Fixes #4540.

# Why do we need this PR?

When using the concourse with newrelic emitter, the metrics is losing most events. The metrics buffer is full since it's only emitting single metric to newrelic server in an inefficient way.

* implemented our own `enqueue`, `flush`, `compress` and `emitBatch` functions to emit metrics data in batch
* added unit test for newrelic emitter

# Contributor Checklist
- [x] Unit tests
~~-[ ] Updated documentation (located at https://github.com/concourse/docs)~~

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/metric/emitter/newrelic.go
+++ b/atc/metric/emitter/newrelic.go
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	newrelicPayloadMaxSize = 1024 * 1024
+	newrelicPayloadMaxBytes = 1024 * 1024
 )
 
 type (
@@ -35,21 +35,16 @@ type (
 	}
 
 	NewRelicConfig struct {
-		AccountID          string        `long:"newrelic-account-id" description:"New Relic Account ID"`
-		APIKey             string        `long:"newrelic-api-key" description:"New Relic Insights API Key"`
-		ServicePrefix      string        `long:"newrelic-service-prefix" default:"" description:"An optional prefix for emitted New Relic events"`
-		CompressionEnabled bool          `long:"newrelic-metric-compression" description:"New Relic payload compression flag"`
-		FlushInterval      time.Duration `long:"newrelic-flush-interval" default:"60s" description:"Newrelic metric flush interval in seconds"`
+		AccountID         string        `long:"newrelic-account-id" description:"New Relic Account ID"`
+		APIKey            string        `long:"newrelic-api-key" description:"New Relic Insights API Key"`
+		ServicePrefix     string        `long:"newrelic-service-prefix" default:"" description:"An optional prefix for emitted New Relic events"`
+		EnableCompression bool          `long:"newrelic-metric-compression" description:"New Relic payload compression flag"`
+		FlushInterval     time.Duration `long:"newrelic-flush-interval" default:"60s" description:"Newrelic metric flush interval in seconds"`
 	}
 
 	singlePayload map[string]interface{}
 
 	batchPayload []byte
-
-	BatchEmitter interface {
-		Emit(logger lager.Logger, payload singlePayload)
-		Flush(logger lager.Logger)
-	}
 
 	batchEmitter struct {
 		client        *http.Client
@@ -90,7 +85,7 @@ func (config *NewRelicConfig) NewEmitter() (metric.Emitter, error) {
 			client:        client,
 			url:           fmt.Sprintf("https://insights-collector.newrelic.com/v1/accounts/%s/events", config.AccountID),
 			apikey:        config.APIKey,
-			compression:   config.CompressionEnabled,
+			compression:   config.EnableCompression,
 			flushInterval: config.FlushInterval,
 			lastEmitTime:  time.Now(),
 		},
@@ -226,13 +221,27 @@ func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
 	}
 }
 
+func (emitter *NewRelicEmitter) BufferPayloadSize() int {
+	size := 0
+	if emitter.batchEmitter.emitBuffer != nil {
+		size = len(emitter.batchEmitter.emitBuffer)
+	}
+	return size
+}
+
+func (emitter *NewRelicEmitter) SetUrl(url string) {
+	emitter.batchEmitter.url = url
+}
+
 func (emitter *batchEmitter) Flush(logger lager.Logger) {
-	if emitter.flushThreshold() {
-		batchPayload := emitter.flush()
-		if batchPayload != nil && len(batchPayload) > 0 {
-			emitter.updateLastEmitTime()
-			go emitter.emitBatch(logger, batchPayload)
-		}
+	if !emitter.flushThreshold() {
+		return
+	}
+
+	batchPayload := emitter.flush()
+	if batchPayload != nil && len(batchPayload) > 0 {
+		emitter.updateLastEmitTime()
+		go emitter.emitBatch(logger, batchPayload)
 	}
 }
 
@@ -255,7 +264,6 @@ func (emitter *batchEmitter) Emit(logger lager.Logger, payload singlePayload) {
 // enqueue new payload
 // if enqueue the new payload would exceed the max limit (1Mb)
 // return everything that's enqueued and enqueue the new payload
-// lock and unlock should be applied.
 func (emitter *batchEmitter) enqueue(newPayload singlePayload) (batchPayload, *loggableError) {
 	var (
 		tempBuff []byte
@@ -281,7 +289,7 @@ func (emitter *batchEmitter) enqueue(newPayload singlePayload) (batchPayload, *l
 	}
 
 	// when combined new payload exceeds 1MB
-	if payloadSize > newrelicPayloadMaxSize-2 { // reduce '[' and ']' for the json payload
+	if payloadSize > newrelicPayloadMaxBytes-2 { // reduce '[' and ']' for the json payload
 		buff = formatBatchEmitBuffer(emitter.emitBuffer)
 		emitter.emitBuffer = newPayloadData
 	} else {
@@ -293,7 +301,7 @@ func (emitter *batchEmitter) enqueue(newPayload singlePayload) (batchPayload, *l
 func (emitter *batchEmitter) bufferSize(buff []byte) (int, *loggableError) {
 	payloadSize := len(buff)
 	// for better performance, we only check compressed data size when the data in the buffer is larger than 1MB
-	if emitter.compression && payloadSize >= newrelicPayloadMaxSize-2 { // reduce size for '[' and ']'
+	if emitter.compression && payloadSize >= newrelicPayloadMaxBytes-2 { // reduce size for '[' and ']'
 		zippedBuffer, err := gZipBuffer(buff)
 		if err != nil {
 			return 0, newLoggableError("failed-to-gzip-payload", err)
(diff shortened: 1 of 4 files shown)


atc/metric/emitter/newrelic.go
```diff
@@ -1,41 +1,74 @@
...
+	"sync"
 	"time"
 
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc/metric"
 	"github.com/pkg/errors"
 )
 
+const (
+	NEWRELIC_PAYLOAD_MAX_SIZE            = 1024 * 1024
```
cirocosta: nit: trying to follow common conventions (e.g., Go std lib), we prefer camelcase 😁  e.g.: https://github.com/concourse/concourse/blob/7a6048b7d6a9124c891b58d9560a9af8be80edaf/tsa/commands.go#L3-L9

also, do these need to be exported? If I understood correctly, they're not necessary anywhere else 🤔

atc/metric/emitter/newrelic.go
```diff
@@ -1,41 +1,74 @@
 package emitter
 
 import (
+	"bufio"
 	"bytes"
+	"compress/gzip"
 	"encoding/json"
 	"fmt"
+	"github.com/concourse/flag"
```
cirocosta: nit: we usually separate these imports from stdlib - e.g.: https://github.com/concourse/concourse/blob/7a6048b7d6a9124c891b58d9560a9af8be80edaf/atc/db/build_factory.go#L3-L11

atc/metric/emitter/newrelic.go
```diff
@@ -1,41 +1,74 @@
...
 
 	NewRelicConfig struct {
-		AccountID     string `long:"newrelic-account-id" description:"New Relic Account ID"`
-		APIKey        string `long:"newrelic-api-key" description:"New Relic Insights API Key"`
-		ServicePrefix string `long:"newrelic-service-prefix" default:"" description:"An optional prefix for emitted New Relic events"`
+		AccountID          string `long:"newrelic-account-id" description:"New Relic Account ID"`
+		APIKey             string `long:"newrelic-api-key" description:"New Relic Insights API Key"`
+		ServicePrefix      string `long:"newrelic-service-prefix" default:"" description:"An optional prefix for emitted New Relic events"`
+		CompressionEnabled bool   `long:"newrelic-metric-compression" description:"New Relic payload compression flag"`
+		FlushInterval      int    `long:"newrelic-flush-interval" default:"60" description:"New Relic metric flush interval in seconds"`
```
cirocosta: nit: we typically use [`time.Duration`](https://golang.org/pkg/time/#Duration) for these things - e.g.: https://github.com/concourse/concourse/blob/7a6048b7d6a9124c891b58d9560a9af8be80edaf/atc/metric/emitter/influxdb.go#L31

atc/metric/emitter/newrelic.go
```diff
@@ -176,10 +242,125 @@ func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
...
+			err := emitter.emitBatch(batchPayload)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (db *batchEmitter) Emit(payload singlePayload) *loggableError {
```
cirocosta: I guess the intention was to use `emitter` here, rather than `db`? 😁

atc/metric/emitter/newrelic.go
```diff
@@ -85,37 +138,49 @@ func (emitter *NewRelicEmitter) simplePayload(logger lager.Logger, event metric.
...
-	if err != nil {
-		logger.Error("failed-to-serialize-payload", err)
-		return
+func (emitter *batchEmitter) emitBatch(payload batchPayload) *loggableError {
+	var (
+		payloadReader io.Reader
+		err           error
+	)
+
+	payload = append(append([]byte{'['}, payload...), ']')
```
cirocosta: is `payload` guaranteed to be a properly formatted json body that fits within a `[]` to form a proper json array?

I'm afraid this can lead to some trailing `,` problems 😬  

It seems weird to me to expact that `emitBatch` does the final wrapping of a properly formed payload, rather than just emitting it

atc/metric/emitter/newrelic.go
```diff
@@ -1,41 +1,74 @@
...
+	}
+
+	batchEmitter struct {
+		client        *http.Client
+		url           string
+		apikey        string
+		payloadQueue  batchPayload
+		lastEmitTime  time.Time
+		compression   bool
+		dataLock      *sync.Mutex
```
cirocosta: do we necessarily need locks on this?

e.g., the `influxdb` emitter is able to do well without as the places where one would usually use locks to synchronize access happen serially, thus, requiring no need for locks 🤔

atc/metric/emitter/newrelic_test.go
```diff
@@ -0,0 +1,99 @@
+package emitter
+
+import (
+	"code.cloudfoundry.org/lager"
+
+	"github.com/concourse/concourse/atc/metric"
+	"github.com/concourse/flag"
```
cirocosta: same as the other comment around imports 😁

atc/metric/emitter/newrelic.go
pivotal-bin-ju: Good catch! Definitely we should follow the convention.

atc/metric/emitter/newrelic.go
pivotal-bin-ju: Good suggestion. Recently I just learned, a redefined type like `time.Duration`, we could leverage its original type to do related operations(like `+`, `*`).

pivotal-bin-ju:
@cirocosta thanks for the comments.
Now the code(except the test case, I am going to figure it out tomorrow) is already refactored, ready for review.

atc/metric/emitter/newrelic_test.go
```diff
@@ -0,0 +1,100 @@
+package emitter
```
cirocosta: we usually don't have our tests sharing the same package as the package being tested, e.g.:

```console
$ ag -Q "package" **/*_test.go

scheduler/runner_test.go
1:package scheduler_test

scheduler/scheduler_suite_test.go
1:package scheduler_test
...
```

while this definitely restricts the surface that you can touch (only exported methods, ifaces, structs, vars, consts, etc) though, it'll make the tests be a more close approximation of "docs as code"

atc/metric/emitter/newrelic.go
```diff
@@ -177,10 +221,139 @@ func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
...
+func gZipBuffer(body []byte) (io.Reader, error) {
+	var err error
+
+	readBuffer := bufio.NewReader(bytes.NewReader(body))
+	buffer := bytes.NewBuffer([]byte{})
+	writer := gzip.NewWriter(buffer)
+
+	_, err = readBuffer.WriteTo(writer)
+	if err != nil {
+		return nil, err
```
cirocosta: in this case, `writer` is not being closed, right? - usually we put things like `writer.Close` in `defer` statements to avoid missing calling `Close()` on those returns

atc/metric/emitter/newrelic.go
```diff
@@ -177,10 +221,139 @@ func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
...
 		// concourse metric names, not our translation layer.
 		singlePayload["metric"] = event.Name
-		payload = append(payload, singlePayload)
+		emitter.batchEmitter.Emit(logger, singlePayload)
+
+	}
+}
+
+func (emitter *batchEmitter) Flush(logger lager.Logger) {
+	if emitter.flushThreshold() {
```
cirocosta: we tend to prefer early returns for these things, e.g.:

```go
if ! emitter.flushThreshold() {
  return
}

// ... etc etc
```

atc/metric/emitter/newrelic.go
```diff
@@ -177,10 +221,139 @@ func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
...
+	}
+	return buff
+}
+
+func (emitter *batchEmitter) updateLastEmitTime() {
+	emitter.lastEmitTime = time.Now()
+}
+
+func (emitter *batchEmitter) flushThreshold() bool {
+	return time.Now().Sub(emitter.lastEmitTime) > emitter.flushInterval
```
cirocosta: I was curious to see whether "time elapsed" calculations would be faster / slower than "ticker based" interval checking, and it turns out that "ticker based" seems to be faster lol

https://gist.github.com/cirocosta/9b19380ae14b1c2bf67438bb142b618a

BenchmarkTimeNow-12     15730425                73.0 ns/op
BenchmarkTicker-12      223836266                5.31 ns/op

I don't think we necessarily need to change this, just an interesting fact 😁

atc/metric/emitter/newrelic.go
```diff
@@ -13,29 +17,54 @@ import (
...
+		CompressionEnabled bool          `long:"newrelic-metric-compression" description:"New Relic payload compression flag"`
+		FlushInterval      time.Duration `long:"newrelic-flush-interval" default:"60s" description:"Newrelic metric flush interval in seconds"`
 	}
 
 	singlePayload map[string]interface{}
-	fullPayload   []singlePayload
+
+	batchPayload []byte
+
+	BatchEmitter interface {
```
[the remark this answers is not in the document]
zoetian: I think it's for the latter one, let me remove this interface...

pivotal-bin-ju:
Hi @cirocosta,
According to the our test, the performance is acceptable.
when data enqueue we do check the buffer size, if compression is not enabled, it just simply return the size of the buffer; if compression is enabled, it is going to check the compressed the data size once the data is greater than 1MB. https://github.com/concourse/concourse/blob/metric-newrelic/atc/metric/emitter/newrelic.go#L296
So, When the data does not  reach 1MB the efficiency is not pull down when compression is enabled.

Your suggestion is great! I totally agree.
Another thing that I did not figure out is how to tune the buffer size if we found the buffer is over than 1MB. (Actually it  is hard, since we do not check the return value from newrelic server until the PR https://github.com/concourse/concourse/pull/4617 gets merged.

taylorsilva:
Closing this PR because we're going with #4698

PR #4581, opened 2019-10-08
improve k8s-topgun resilience

pivotal-jamie-klassen:
# Existing Issue

Some improvements towards concourse/ci#182. I wouldn't say it completely fixes the issue at this point.

# Changes proposed in this pull request
There are two improvements in this PR:

1. we have found that `kubectl port-forward` seems to intermittently fail, and it has gotten worse lately. Accordingly we will not rely on it to broker our connections to the web nodes during topgun test cases; instead we actually have GKE deploy a load balancer to handle that traffic.
1. the BTRFS cases where failure was expected were making use of the wrong helper function, which meant that they could only pass in certain flaky circumstances.

# Contributor Checklist
- [X] ~Unit tests~
- [X] Integration tests (if applicable)
- [X] ~Updated documentation (located at https://github.com/concourse/docs)~
- [X] ~Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)~

# Reviewer Checklist
> This section is intended for the core maintainers only, to track review progress. Please do not
> fill out this section.
- [X] ~Code reviewed~
- [ ] Tests reviewed
- [X] ~Documentation reviewed~
- [X] ~Release notes reviewed~
- [X] ~PR acceptance performed~

PR #4643, opened 2019-10-24
Enable k8s-topgun to run inside k8s cluster

### LINKED ISSUE #4320: Address k8s-topgun flakiness - round 3
We had performed a round of improvements to k8s-topgun to improve its flaky behavior (https://github.com/concourse/concourse/issues/4124), and, while it got good for a while, it seems to be getting back to us 😬 

Screen Shot 2019-08-26 at 4 23 08 PM

Those "red" ones should definitely not be happening as we didn't perform any changes to the code that would make them fail.

See:

- https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/k8s-topgun/builds/1691#L5d6077bb:54:81
- https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/k8s-topgun/builds/1683#L5d607694:150:162

cirocosta:
# Existing Issue

fixes #4320
fixes https://github.com/concourse/ci/issues/182
closes #4581
closes #4613

# Why do we need this PR?

Running k8s-topgun outside of the k8s cluster has been flakey. These changes
allow us to run these tests inside or outside of the cluster with a new
environment variable `IN_CLUSTER`.

We created an `Endpoint` interface and switch implementations based on the value
of `IN_CLUSTER`. The `Endpoint` implementation always returns an address, which
is either:

1. an address of a port-forwarding session (you're running the test suite
outside the cluster)

2. an internal (to the cluster network) address of either a k8s service/pod
(you're running the test suite inside the cluster)

We also:

- removed `--wait` from the helm installation command. We already have a way of
  waiting for all pods to come up. Having `--wait` was causing the baggageclaim
  tests to fail sometimes because helm would exit with 1 when all we really want
  it to do is run a failed deployment.

- added higher `resources.requests` to schedule less containers to the same
  kubernetes node (so that we let the cluster better auto scale when needed)

- added worker readinessProbe that relies on the Concourse worker healthchecking
  mechanism (so that we know when `gdn` and `baggageclaim` are ready)

- increased timeouts to account for autoscaling (takes ~5m in most cases)

Aside from that, We didn't change the behaviour of any tests. We did refactor
the structure of some things just as errors came up.

# Contributor Checklist

- ~Unit tests~
  - test updates only
- [x] Integration tests (if applicable) (:sweat_smile:)
  - internal changes only
  - internal changes only

# Reviewer Checklist

> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/topgun/k8s/k8s_suite_test.go
+++ b/topgun/k8s/k8s_suite_test.go
@@ -37,13 +37,14 @@ type environment struct {
 	ConcourseImageTag    string `env:"CONCOURSE_IMAGE_TAG"`
 	FlyPath              string `env:"FLY_PATH"`
 	K8sEngine            string `env:"K8S_ENGINE" envDefault:"GKE"`
+	InCluster            bool   `env:"IN_CLUSTER" envDefault:"false"`
 }
 
 var (
-	Environment environment
-	fly         FlyCli
-	namespace   string
-	releaseName string
+	Environment            environment
+	endpointFactory        EndpointFactory
+	fly                    FlyCli
+	releaseName, namespace string
 )
 
 var _ = SynchronizedBeforeSuite(func() []byte {
@@ -79,7 +80,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 })
 
 var _ = BeforeEach(func() {
-	SetDefaultEventuallyTimeout(30 * time.Second)
+	SetDefaultEventuallyTimeout(90 * time.Second)
 	SetDefaultConsistentlyDuration(30 * time.Second)
 
 	tmp, err := ioutil.TempDir("", "topgun-tmp")
@@ -89,6 +90,11 @@ var _ = BeforeEach(func() {
 		Bin:    Environment.FlyPath,
 		Target: "concourse-topgun-k8s-" + strconv.Itoa(GinkgoParallelNode()),
 		Home:   filepath.Join(tmp, "fly-home-"+strconv.Itoa(GinkgoParallelNode())),
+	}
+
+	endpointFactory = PortForwardingEndpointFactory{}
+	if Environment.InCluster {
+		endpointFactory = AddressEndpointFactory{}
 	}
 
 	err = os.Mkdir(fly.Home, 0755)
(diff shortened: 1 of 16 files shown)


topgun/k8s/k8s_suite_test.go
```diff
@@ -139,12 +269,13 @@ func helmDeploy(releaseName, namespace, chartDir string, args ...string) *gexec.
 
 func helmInstallArgs(args ...string) []string {
 	helmArgs := []string{
-		"--set=web.livenessProbe.failureThreshold=3",
-		"--set=web.livenessProbe.initialDelaySeconds=3",
-		"--set=web.livenessProbe.periodSeconds=3",
-		"--set=web.livenessProbe.timeoutSeconds=3",
 		"--set=concourse.web.kubernetes.keepNamespaces=false",
+		"--set=concourse.worker.bindIp=0.0.0.0",
```
cirocosta: this is necessary for tests where we make direct requests to `gdn` - `kubectl port-forward` can forward to loopback addresses (127.0.0.1/8), but that's not true with a direct request to an iface.

topgun/k8s/k8s_suite_test.go
```diff
@@ -139,12 +269,13 @@ func helmDeploy(releaseName, namespace, chartDir string, args ...string) *gexec.
...
 		"--set=postgresql.persistence.enabled=false",
+		"--set=web.resources.requests.cpu=500m",
```
cirocosta: by placing `requests` but not limits, we let processes that need a lot of power (in a spike during startup, e.g.) to leverage that (as there are no actual limits being placed), while still leveraging the proper scheduling from k8s.

topgun/k8s/k8s_suite_test.go
```diff
@@ -139,12 +269,13 @@ func helmDeploy(releaseName, namespace, chartDir string, args ...string) *gexec.
...
+		"--set=worker.readinessProbe.httpGet.path=/",
```
cirocosta: without this, our checks for "is this worker ready" would simply always resolve to "true", regardless of that being true or not (as there is no default `readinessProbe` for the worker: see https://github.com/helm/charts/blob/ed71e4a2bde403c73efab8b4d601b3fd828b75e5/stable/concourse/values.yaml#L1574-L1577)

topgun/k8s/k8s_suite_test.go
```diff
@@ -200,16 +334,15 @@ func getPods(namespace string, flags ...string) []pod {
...
 
 func isPodReady(p pod) bool {
-	total := len(p.Status.ContainerStatuses)
-	actual := 0
-
-	for _, containerStatus := range p.Status.ContainerStatuses {
-		if containerStatus.Ready {
-			actual++
+	for _, condition := range p.Status.Conditions {
+		if condition.Type != "ContainersReady" {
```
cirocosta: by looking at `containersready` rather than "running", we ensure that we're only counting as "ready" something that's truly ready (i.e., pass the readiness probe).

cirocosta:
Sorry for the huge PR 😬 

This PR is meant to come together with https://github.com/concourse/ci/pull/191, which then leverages `IN_CLUSTER` and the fact that we now have a worker registered with `ci.concourse-ci.org` tagged as `k8s-topgun` (living in the `topgun` cluster).

topgun/k8s/k8s_suite_test.go
jomsie: this seems like a really good addition

topgun/k8s/k8s_suite_test.go
jomsie: just referenced https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions good read! good addition!

PR #4657, opened 2019-10-28
feat: make RBAC configurable.

### LINKED ISSUE #4632: feature request: make RBAC configurable

Currently role `operator` doesn't have permission of `hijack`, but our users are requesting that. Each company/organization may have its own operation model, I was considering to just add `hijack` permission to `operator`, but I'm afraid other users may be unhappy with that.

My proposal is to make role-permission mapping configuration. In most of cases, user won't completely redefine the mapping, instead, user may just make some small tuning based on the standard mapping.

We can make `concourse web` to take a RBAC file in YAML format. For example, if I just want to add `hijack` permission to `operator`, then file will look like:

     operator:
        - hijack

Then feed this file as

    $ concourse web --rbac-tuning rbac.yml ...

When ATC reads the file, it will add `hajack` permission to `operator`, and remove `hijack` from `member`.

Yes

evanchaoli:
# Existing Issue

Fixes #4632

# Why do we need this PR?

This PR will allow user to tuning role-action mapping.

Add a new CLI option `--config-rbac` that take a YAML file with the following format:

```yaml
pipeline-operator:
 - OrderPipelines
```

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


pivotal-jwinters:
@evanchaoli this is awesome thanks! The only thing I'm not sure of is the backing up and restoring the old roles. Is this going to be used for anything other than the unit tests?

evanchaoli:
> @evanchaoli this is awesome thanks! The only thing I'm not sure of is the backing up and restoring the old roles. Is this going to be used for anything other than the unit tests?

@pivotal-jwinters The restoring function should never be called from anywhere other than the unit tests. But current code cannot block that if somebody calls it. To enforce, I have two ideas:

* Explicitly indicate in the function comment that it should not be called from non-unit-test.
* In the restoring function, use `runtime.Caller()` to check caller, and caller file must be "_test.go" suffixed.

Do you have any suggestion?

evanchaoli:
@pivotal-jwinters I have just updated the commit and applied my two ideas, so that if the restoring function is called from a non-unit-test function, it will do nothing but only return an error.

evanchaoli:
> What if we just passed the role mapping data down through the accessorFactory instead? That way we don't have to test with global variables, and we don't need to worry about restoring anything in the unit tests?

@pivotal-jwinters To clarify, if you mean to maintain a local copy of `requiredRoles` in struct `access` and merge custom map into local `requiredRoles` when new an `access` object? 

That way, each `access` object will hold a roles map, is that a memory overhead?  Also, we will need to add a function `newAccess()`, and combine the global `requiredRoles` with custom map, which is a CPU overhead, then we also need to figure out some way to prevent from initializing an `access` object without calling `newAccess()`, which is an algorithm overhead. 

Please correct me if I understood you wrongly.

evanchaoli:
@pivotal-jwinters As you suggested, I now maintain a copy of customized action-role-map in accessor factory. Instead of passing the map to accessor, I create a new interface `ActionRoleMap` and only pass the interface to accessor. Please see if you are ok with this version.

atc/api/accessor/accessor_factory.go
```diff
@@ -52,3 +72,51 @@ func (a *accessFactory) validate(token *jwt.Token) (interface{}, error) {
...
 }
+
+type customRolesMapping map[string][]string
+
+func (a *accessFactory) CustomizeActionRoleMap(logger lager.Logger, configFilePath string) error {
+	if configFilePath == "" {
+		return nil
+	}
+
+	content, err := ioutil.ReadFile(configFilePath)
```
pivotal-jwinters: One last thing :), can we do all our file reading at the atccmd level instead? That way the accessorFactory can simply take a map of the custom roles, and it makes the tests a lot simpler.

evanchaoli: OK, I added a new method `accessor.ParseCustomActionRoleMap()` and called it from `atc/atccmd/command.go`.

pivotal-jwinters:
Also I've been thinking about what it might look like to allow for new roles, not just reconfiguring the existing roles that we have.

atc/api/accessor/role_action_map.go
```diff
@@ -0,0 +1,124 @@
...
+	atc.CreateArtifact:                "member",
+	atc.GetArtifact:                   "member",
+	atc.ListBuildArtifacts:            "viewer",
+}
+
+type CustomActionRoleMap map[string][]string
+
+//go:generate counterfeiter . ActionRoleMap
+
+type ActionRoleMap interface {
```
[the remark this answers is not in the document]
evanchaoli: Make sense. Let's split it into two interfaces.

evanchaoli:
> Also I've been thinking about what it might look like to allow for new roles, not just reconfiguring the existing ones. I don't think it would be too difficult, it might take a small change to fly to allow for free form role input.
> 
> It's probably something that we can consider in another PR though.

@pivotal-jwinters It would be great to support adding/removing roles, which is fully configurable RBAC. But I feel it would definitely need more considerations. For example, with current RBAC model, owner > team member > operator > viewer. If a user simply adds two roles, named "role1" and "role2", it would break the model. I'm not saying it would be hard to solve the problem, I'm just saying it should be well designed before we start to implement. For now, current PR is good enough for our use cases, I'm also working on a couple of more feature PRs (not raise yet), so I'd not extend scope of this PR.

evanchaoli:
@pivotal-jwinters I have addressed your latest comment, split `CustomizeActionRoleMap` method to a separate interface `ActionRoleMapModifier`. Please take a look again. Thanks.

atc/api/accessor/role_action_map.go
```diff
@@ -0,0 +1,129 @@
...
+	RoleOfAction(string) string
+}
+
+//go:generate counterfeiter . ActionRoleMapModifier
+
+type ActionRoleMapModifier interface {
+	CustomizeActionRoleMap(lager.Logger, CustomActionRoleMap) error
+}
+
+func ParseCustomActionRoleMap(filename string, mapping *CustomActionRoleMap) error {
```
deniseyu: hi @evanchaoli, the file parsing logic here is similar to how we validate and read from files in [this flags library](https://github.com/concourse/flag/blob/master/file.go). To use it, we input config values as `flag.File` rather than `string`, like this:

https://github.com/concourse/concourse/blob/84acaa94d6d7bf9bfd895770ede79ceb1c8dd6d4/atc/atccmd/command.go#L96

Another advantage of parsing inputs this way is that the existence of the file is validated earlier. Would you be up for refactoring this later on to leverage the existing library method? Let us know if you need support.

PR #4661, opened 2019-10-28
bump concourse/dex dependency

### LINKED ISSUE #4535: On a Concourse using cf-auth, when a user switches to the Auditor role, they lose access to their teams
**Summary**

Within cf-auth, a user has access to [three roles](https://docs.cloudfoundry.org/concepts/roles.html#roles) with their given org: Manager, Developer, and Auditor. It appears as though Concourse auth only recognizes the Developer role. If a user switches their role to Auditor within their cf-auth org, they lose access to their Concourse teams.

Speaking to @pivotal-jwinters, it seems as though their is a secondary API endpoint called `audited-spaces` for this role that Concourse will need to integrate with for this to work and consolidated all of the roles within spaces. 

Currently the CF auth connector hits the `v2/users/user-id/spaces` ([docs](https://apidocs.cloudfoundry.org/12.0.0/users/list_all_spaces_for_the_user.html)) endpoint to determine if a user is part of a space. This endpoint returns spaces that a user has a CF developer role for. It does not return spaces that a user is an auditor for though.

To get the spaces that a user is an auditor for we should also hit `v2/users/user-id/audited_spaces` ([docs](https://apidocs.cloudfoundry.org/12.0.0/users/list_all_audited_spaces_for_the_user.html)). We should then have the full list of spaces that a user belongs to.

**Side note**: During the call we did not confirm if a user with the CF Manager role also has the same issue. If it does then we may need to hit a _third_ API endpoint, `v2/users/user-id/managed_spaces` ([docs](https://apidocs.cloudfoundry.org/12.0.0/users/list_all_managed_spaces_for_the_user.html))

**Acceptance Criteria**

- As long as a user is within the proper org, they should have access to the Concourse teams they're auth'd against, regardless of their role being Manager, Developer, or Auditor.

pivotal-jwinters:
# Existing Issue
Fixes #4535.

# Why do we need this PR?
* includes the fix for cf connector that fetches all spaces for a user

# Contributor Checklist

# Reviewer Checklist
- [ ] Release notes reviewed

### DIFF
--- a/release-notes/latest.md
+++ b/release-notes/latest.md
@@ -8,7 +8,7 @@ There is no configuration required to take advantage of these new improvements.
 
 #### <sub><sup><a name="4535" href="#4535">:link:</a></sup></sub> feature, breaking
 
-* The cf auth connector used to only return spaces for the `developer` role. Now it returns spaces for `developer`, `auditor`, and `manager`.
+* The CloudFoundry auth connector, when configured to authorize users based on CF space membership, will now authorize space auditors and space managers in addition to space developers. This is a breaking change as any teams with CF space-based configuration may grant access to users that they wouldn't have before. #4661
 
 #### <sub><sup><a name="4480" href="#4480">:link:</a></sup></sub> feature
 
--- a/release-notes/latest.md
+++ b/release-notes/latest.md
@@ -6,6 +6,10 @@ This should help reduce the amount of duplicated work (when running more than 1 
 
 There is no configuration required to take advantage of these new improvements.
 
+#### <sub><sup><a name="4535" href="#4535">:link:</a></sup></sub> feature, breaking
+
+* The cf auth connector used to only return spaces for the `developer` role. Now it returns spaces for `developer`, `auditor`, and `manager`.
+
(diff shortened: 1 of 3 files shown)


release-notes/latest.md
```diff
@@ -1,3 +1,7 @@
+#### <sub><sup><a name="4535" href="#4535">:link:</a></sup></sub> feature, breaking
+
+* The cf auth connector used to only return spaces for the `developer` role. Now it returns spaces for `developer`, `auditor`, and `manager`.
```
vito: Might be better to show the user-level impact of this since it's a breaking change:

```suggestion
* The CloudFoundry auth connector, when configured to authorize users based on CF space membership, will now authorize space auditors and space managers in addition to space developers. This is a breaking change as any teams with CF space-based configuration may grant access to users that they wouldn't have before. #4661
```

Is that accurate? /cc @xtremerui

xtremerui: I think it is good as a warn to ops to double check the granted access.

ukabu:
https://github.com/concourse/concourse/issues/4535#issuecomment-548933195

PR #4666, opened 2019-10-28
atc: skip gzip on download cli endpoint

xtremerui:
Fixes below error and progress bar displaying wrongly when running `fly sync`.
```
warning: failed to parse Content-Length: strconv.ParseInt: parsing "": invalid syntax
```

# Why do we need this PR?
By https://github.com/concourse/concourse/pull/4470, the gzip handler removes `Content-length` during its process caused the above error

Skip gziphandler gzips `DownloadCLI` endpoint

# Contributor Checklist
- [x] ~Unit tests~

# Reviewer Checklist
- [x] Code reviewed
- [x] Tests reviewed
- [x] Release notes reviewed
- [x] PR acceptance performed

### DIFF
--- a/atc/wrappa/compression_wrappa.go
+++ b/atc/wrappa/compression_wrappa.go
@@ -21,19 +21,21 @@ func (wrappa CompressionWrappa) Wrap(handlers rata.Handlers) rata.Handlers {
 	wrapped := rata.Handlers{}
 
 	for name, handler := range handlers {
-
+		switch name {
 		// always gzip for events
-		if name == atc.BuildEvents {
+		case atc.BuildEvents:
 			gzipEnforcedHandler, err := gziphandler.GzipHandlerWithOpts(gziphandler.MinSize(0))
 			if err != nil {
 				wrappa.Logger.Error("failed-to-create-gzip-handler", err)
 			}
 
 			wrapped[name] = gzipEnforcedHandler(handler)
-			continue
+		// skip gzip as this endpoint does it already
+		case atc.DownloadCLI:
+			wrapped[name] = handler
+		default:
+			wrapped[name] = gziphandler.GzipHandler(handler)
 		}
-
-		wrapped[name] = gziphandler.GzipHandler(handler)
 	}
 
 	return wrapped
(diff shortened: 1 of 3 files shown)


pivotal-jamie-klassen:
Good point!

I'd love to see a bit better documentation/motivation for this decision... 🤔 

It's definitely important for `fly`'s ability to show a progress bar, and that feels like reasonable documentation. Otherwise seeing a PR that removes compression can feel like a step backwards! 😬 

How hard is it to add a test that the `Content-Length` header is on that endpoint? Do we have a testflight test that runs `fly sync`?

cirocosta:
> How hard is it to add a test that the Content-Length header is on that endpoint? 

oh, I don't think this PR is removing that - if I understood correctly, the `Content-Length` header used to be removed when having the bits sent via the gzip handler: 

```go
func (w *GzipResponseWriter) startGzip() error {
	// Set the GZIP header.
	w.Header().Set(contentEncoding, "gzip")


	// if the Content-Length is already set, then calls to Write on gzip
	// will fail to set the Content-Length header since its already set
	// See: https://github.com/golang/go/issues/14975.
	w.Header().Del(contentLength)
```

https://github.com/nytimes/gziphandler/blob/dd0439581c7657cb652dfe5c71d7d48baf39541d/gzip.go#L154-L157

By making `DownloadCLI` not go through gzip handler, that `content-length` header is then not removed anymore (a header that we were already setting):

https://github.com/concourse/concourse/blob/6d9ab0285aaf5e6e4ec15ca73075e21095fac90a/atc/api/cliserver/download.go#L60

Naturally, we'd be removing the compression of those bits (as `downloadserver` extracts the uncompressed file on the fly) 😬

pivotal-jamie-klassen:
@cirocosta my thinking is that we would not have seen the progress bar pain if we had a regression test that ensured that responses from `/api/v1/cli` always _had_ a `Content-Length` header -- since it's obviously the only endpoint that really _needs_ one. It was #4470 that broke this, so I'm asking for a test to protect us against similar cross-cutting API changes in the future. Does that make sense?

xtremerui:
I think having a testflight test for it is fair.

cirocosta:
ooh yeah, I totally agree on the testing 👍

vito:
@xtremerui Is this a regression that we'll need fixed for v5.7?

xtremerui:
@vito yes it is a regression introduced by the gzip change. I'd say the sync still works but annoying to see the warning. Thx!

xtremerui:
@pivotal-jamie-klassen testflight test and release note added

PR #4673, opened 2019-10-29
WIP: feat: add pipeline collector

### LINKED ISSUE #4578: Pipeline reaper

On some clusters, especially the experimental and staging clusters, we have frequently seen that some users deploy some pipelines and forget them, which leaves garbage pipelines on Concourse, making UI messy, wasting ATC's resource. 

I'm thinking to add a time-based pipeline reaper:

1. Add a CLI option `pipeline-reap-days`, default to 0. If set to 0, then disable the reaper.
2. When `pipeline-reap-days` is set to X days, if a pipeline has no new build triggered within X days, then reaper will pause the pipeline.
3. If a pipeline has been paused for X days, then reaper will destroy the pipeline.

Yes. But I'd like to get an indication of go from Concourse dev team before starting to work on the feature.

evanchaoli:
Added a CLI option `--pipeline-ttl` to concourse web. By default
this ttl is 0, meaning that pipeline collector won't work. When
ttl is set, if a non-paused has no any build within `ttl`, then
it will be paused; if a paused pipeline has no any build within
2*`ttl`, then it will be destroyed.

# Existing Issue

Fixes #4578 .

# Why do we need this PR?

It will help auto cleanup zombie pipelines, so that keep UI clean and save ATC resource.

* Added a pipelineCollector
* Also fixed a bug of `func getBuildsWithDates()` in `atc/db/build_factory.go`. Where if query by time returns no data, then it should not go on to call `getBuildsWithPagination`.

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -209,6 +209,8 @@ type RunCommand struct {
 	EnableRedactSecrets bool `long:"enable-redact-secrets" description:"Enable redacting secrets in build logs."`
 
 	ConfigRBAC string `long:"config-rbac" description:"Customize RBAC role-action mapping."`
+
+	PipelineTTL time.Duration `long:"pipeline-ttl" description:"A pipeline having no build for ttl will be paused, for 2*ttl will be destroyed. 0 means no ttl. (Example: 30m, 60h, valid time units are m and h)."`
 }
 
 type Migration struct {
@@ -966,6 +968,7 @@ func (cmd *RunCommand) constructGCMember(
 	fetchSourceFactory := fetcher.NewFetchSourceFactory(dbResourceCacheFactory, resourceFactory)
 	resourceFetcher := fetcher.NewFetcher(clock.NewClock(), lockFactory, fetchSourceFactory)
 	dbResourceConfigFactory := db.NewResourceConfigFactory(gcConn, lockFactory)
+	dbPipelineFactory := db.NewPipelineFactory(gcConn, lockFactory)
 	imageResourceFetcherFactory := image.NewImageResourceFetcherFactory(
 		dbResourceCacheFactory,
 		dbResourceConfigFactory,
@@ -1019,6 +1022,7 @@ func (cmd *RunCommand) constructGCMember(
 		atc.ComponentCollectorContainers:        gc.NewContainerCollector(dbContainerRepository, jobRunner, cmd.GC.MissingGracePeriod),
 		atc.ComponentCollectorCheckSessions:     gc.NewResourceConfigCheckSessionCollector(resourceConfigCheckSessionLifecycle),
 		atc.ComponentCollectorVarSources:        gc.NewCollectorTask(cmd.varSourcePool.(gc.Collector)),
+		atc.ComponentCollectorPipelines:         gc.NewPipelineCollector(dbPipelineFactory, cmd.PipelineTTL),
 	}
 
 	for collectorName, collector := range collectors {
@@ -1444,6 +1448,9 @@ func (cmd *RunCommand) configureComponentIntervals(componentFactory db.Component
 			}, {
 				Name:     atc.ComponentCollectorVarSources,
 				Interval: 60 * time.Second,
+			}, {
+				Name:     atc.ComponentCollectorPipelines,
+				Interval: 5 * time.Minute,
 			},
 		})
 }
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -1450,7 +1450,9 @@ func (cmd *RunCommand) configureComponentIntervals(componentFactory db.Component
 				Interval: 60 * time.Second,
 			}, {
 				Name:     atc.ComponentCollectorPipelines,
-				Interval: 5 * time.Minute,
+				// Pipeline collector is not a critical GC task, only run it
+				// every 10 GC cycles.
+				Interval: 10 * cmd.GC.Interval,
 			},
 		})
 }
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -1010,7 +1010,7 @@ func (cmd *RunCommand) constructGCMember(
 		time.Minute,
 	)
 
-	collectors := map[string]lockrunner.Task{
+	collectors := map[string]lockrunner.Task {
 		atc.ComponentCollectorBuilds:            gc.NewBuildCollector(dbBuildFactory),
 		atc.ComponentCollectorWorkers:           gc.NewWorkerCollector(dbWorkerLifecycle),
 		atc.ComponentCollectorResourceConfigs:   gc.NewResourceConfigCollector(dbResourceConfigFactory),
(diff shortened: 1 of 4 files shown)


atc/atccmd/command.go
```diff
@@ -1420,6 +1424,9 @@ func (cmd *RunCommand) configureComponentIntervals(componentFactory db.Component
 			}, {
 				Name:     atc.ComponentCollectorWorkers,
 				Interval: cmd.GC.Interval,
+			}, {
+				Name:     atc.ComponentCollectorPipelines,
+				Interval: 5 * time.Minute,
```
vito: Mind adding a flag for this or just using `cmd.GC.Interval`?

atc/gc/pipeline_collector.go
```diff
@@ -0,0 +1,88 @@
...
+	pipelines, err := p.pipelineFactory.AllPipelines()
+	if err != nil {
+		logger.Error("failed-to-get-pipelines", err)
+		return err
+	}
+	logger.Debug("after-get-all-pipelines", lager.Data{"total-pipeline-count": len(pipelines)})
+
+	for _, pipeline := range pipelines {
+		log := logger.Session("tick").WithData(lager.Data{"pipeline-id": pipeline.ID()})
+		if pipeline.Paused() {
```
vito: :thought_balloon: This could make it pretty easy to accidentally destroy your prod pipeline if you pause it while investigating an external issue that takes longer than expected to resolve. Paused pipelines will not produce any builds so they'd be subject to this garbage collector after enough time elapses.

What value do you think y'all would set for your own environment? I would imagine you'd need a pretty high TTL for pipelines to survive the weekend, too.

atc/gc/pipeline_collector.go
```diff
@@ -0,0 +1,88 @@
...
+				Since: int(time.Now().Unix()-int64(p.ttl.Seconds())),
+			}
+			builds, _, err := pipeline.BuildsWithTime(page)
+			if err != nil {
+				log.Error("failed-to-query-build-of-pipeline", err)
+			}
+			logger.Debug("running-query-builds", lager.Data{"buillds": len(builds)})
+			if len(builds) == 0 {
+				log.Info("pause-pipeline")
+				err := pipeline.Pause()
```
vito: I like the idea of a two-phase collector here, at least until we implement pipeline archiving. It's a decent cue to notice the pipeline is about to be removed.

atc/db/build_factory.go
```diff
@@ -178,7 +178,9 @@ func getBuildsWithDates(buildsQuery, minMaxIdQuery sq.SelectBuilder, page Page,
 
 		defer sinceRow.Close()
 
+		found := false
```
vito: Are there any tests to update/add for this or is it more of an optimization that's already technically exercised by existing tests and hard to test in isolation?

Would you mind splitting it out into a separate commit, just so it's clear that it's not strictly part of the pipeline collector logic?

evanchaoli: Sure, I make a separate PR to fix this bug.

evanchaoli:
> Most of my anxieties around this would be mitigated by archiving pipelines instead, but I understand not wanting to tackle that yet.

@vito This feature is by default turned off. And I think people should be very careful to turn on this feature on production. But this feature would be handy for experiment and staging clusters. For us, we would only turn on this feature on a playground cluster with a small TTL and on staging cluster with a bigger TTL. There have been a lot of zombie pipelines on our playground cluster, that's reason why I want this feature in place the earlier the merrier. We don't plan to use it on production.

atc/gc/pipeline_collector.go
evanchaoli: We would not use this feature on production cluster. From my current thinking, for playground (experiment) cluster, TTL could be 7 days; and for staging clusters, TTL could be 30 days.

atc/atccmd/command.go
evanchaoli: I don't feel it is worthy adding a new flag as this is not a critical collector task. `cmd.GC.Interval` seems to be too frequent for this task. If collectors all use `cmd.GC.Interval`, all GC tasks might start at same time, would it cause GC storm? 

I hard coded 5 minutes here because I saw there are a couple of other components like `atc.ComponentScheduler` and `atc.ComponentBuildReaper` also hard code their intervals. Maybe that's reason we should never leave bad samples in code. :) But I feel ok for this non-critical task. Are you ok? @vito

evanchaoli:
@vito I have just created pr #4682 to fix the db bug. Please merge that bug first, without that fix, pipeline reaper wouldn't work.

atc/gc/pipeline_collector.go
evanchaoli: Thanks.

atc/atccmd/command.go
evanchaoli: Changed to use `cmd.GC.Interval`.

aledeganopix4d:
@evanchaoli this is an very interesting PR!
I have a doubt: in my understanding the pipeline_collector is unable to understand if a pipeline was set and unpaused 5 minutes before it ran, am I correct?
If that is the case then it would be a bit annoying for the users to see their pipeline removed so quickly.
I have a proposal to cover this case here: #4809.
Let me please know what you think about it!
Thanks!

evanchaoli:
> I have a doubt: in my understanding the pipeline_collector is unable to understand if a pipeline was set and unpaused 5 minutes before it ran, am I correct?
> If that is the case then it would be a bit annoying for the users to see their pipeline removed so quickly.

@aledeganopix4d Basically TTL should be days rather than just minutes. The reason I created this PR is, there are a lot of zombie pipelines on our clusters, people tried to deploy pipelines and left them there. So in the implementation, pipeline_collector is by default switched off. For myself, I'll not turn on pipeine_collector on the production clusters, and only use it on experimental clusters.

evanchaoli:
Please hold on this review. I just realized a use case:

* A pipeline has no any build.
* After a TTL, reaper pauses the pipeline.
* Then the pipeline owner notices the pipeline is paused, and un-pauses the pipeline.

In this case, TTL should be counted since the user unpauses the pipeline.

vito:
FWIW, we're starting on the [archiving pipelines](https://project.concourse-ci.org/projects/MDc6UHJvamVjdDM4Mzc3ODI=) work, and it might actually be done in time for this to use it. :slightly_smiling_face:

clarafu:
@evanchaoli I wanted to confirm with you about if I'm understanding the logic behind the destroying of pipelines. When you say that if a paused pipeline has no any build within
2*ttl, then it will be destroyed, I'm not sure if that takes into account that paused pipelines cannot schedule any builds which will result in any pipelines that are paused for longer than 2*ttl to be destroyed. Is that the behaviour you intended? 

Also I wanted to sync up with you about if we can discuss having the pipelines be archived rather than destroyed. This will make it feel less intrusive and keeping around the history of the builds, but also helping clean up the clutter of pipelines. We have a chunk of work done for archiving pipelines, you can track it here https://project.concourse-ci.org/projects/MDc6UHJvamVjdDM4Mzc3ODI=. Not sure what you think about that?

clarafu:
I'm going to go ahead and close this first due to inactivity, please feel free to open a new PR if you would like to re-discuss the use case for this, thanks!

PR #4682, opened 2019-10-31
fix(db): fixed a bug where if query builds with a date that newer than all builds, it should return nothing but it actually returned all builds.

evanchaoli:
# Existing Issue

No

# Why do we need this PR?

As the subject, this bug causes `db.AllBuilds()` to return wrong builds, in turn causes the API `api/v1/teams/<team>/builds?timestamp=<unix time>` to return wrong builds if specified time is newer than all builds.

This bug also blocks my PR #4673.

* The bug is in function `getBuildsWithDates()`. If query by time returns no date, then the fucntion should just return rather than going on to call `getBuildsWithPagination`.

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/build_factory.go
+++ b/atc/db/build_factory.go
@@ -178,7 +178,9 @@ func getBuildsWithDates(buildsQuery, minMaxIdQuery sq.SelectBuilder, page Page, 
 
 		defer sinceRow.Close()
 
+		found := false
 		for sinceRow.Next() {
+			found = true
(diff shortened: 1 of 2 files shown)


evanchaoli:
Well, my team member @mgsolid also found this bug and opened issue #4683.

PR #4692, opened 2019-10-31
Adding lidar metrics

### LINKED ISSUE #4554: Add metrics for LIDAR
* [x] Size of check queue
* [x] Number of checks queued (per ATC)
* [x] Number of checks GC'd
* [x] Time it takes to run each check (start time - end time)
* [x] Time between queue and check started (create time - start time)
* [x] Check status (success/failure/error)

xtremerui:
# Existing Issue

Fixes #4554

# Why do we need this PR?
So we could have metrics available when Lidar is enabled.

Added below metrics for Lidar
 - `check_started`
 - `check_finished`
 - `check_enqueue`
 - `check_queue_size`
 - `checks_deleted`

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/check.go
+++ b/atc/db/check.go
@@ -145,13 +145,16 @@ func (c *check) Start() error {
 
 	defer Rollback(tx)
 
-	_, err = psql.Update("checks").
+	var startTime time.Time
+	err = psql.Update("checks").
 		Set("start_time", sq.Expr("now()")).
 		Where(sq.Eq{
 			"id": c.id,
 		}).
+		Suffix("RETURNING start_time").
 		RunWith(tx).
-		Exec()
+		QueryRow().
+		Scan(&startTime)
 	if err != nil {
 		return err
 	}
@@ -172,6 +175,8 @@ func (c *check) Start() error {
 		return err
 	}
 
+	c.startTime = startTime
+
 	return nil
 }
 
@@ -191,6 +196,7 @@ func (c *check) finish(status CheckStatus, checkError error) error {
 
 	defer Rollback(tx)
 
+	var endTime time.Time
 	builder := psql.Update("checks").
 		Set("status", status).
 		Set("end_time", sq.Expr("now()")).
@@ -202,9 +208,11 @@ func (c *check) finish(status CheckStatus, checkError error) error {
 		builder = builder.Set("check_error", checkError.Error())
 	}
 
-	_, err = builder.
+	err = builder.
+		Suffix("RETURNING end_time").
 		RunWith(tx).
-		Exec()
+		QueryRow().
+		Scan(&endTime)
 	if err != nil {
 		return err
 	}
@@ -232,6 +240,8 @@ func (c *check) finish(status CheckStatus, checkError error) error {
 	if err != nil {
 		return err
 	}
+
+	c.endTime = endTime
 
 	return nil
 }
--- a/atc/db/check.go
+++ b/atc/db/check.go
@@ -242,6 +242,7 @@ func (c *check) finish(status CheckStatus, checkError error) error {
 	}
 
 	c.endTime = endTime
+	c.status = status
 
 	return nil
 }
--- a/atc/db/check_lifecycle.go
+++ b/atc/db/check_lifecycle.go
@@ -10,7 +10,7 @@ import (
 //go:generate counterfeiter . CheckLifecycle
 
 type CheckLifecycle interface {
-	RemoveExpiredChecks(time.Duration) error
+	RemoveExpiredChecks(time.Duration) (int, error)
 }
 
 type checkLifecycle struct {
@@ -23,9 +23,9 @@ func NewCheckLifecycle(conn Conn) *checkLifecycle {
 	}
 }
 
-func (lifecycle *checkLifecycle) RemoveExpiredChecks(recyclePeriod time.Duration) error {
+func (lifecycle *checkLifecycle) RemoveExpiredChecks(recyclePeriod time.Duration) (int, error) {
 
-	_, err := psql.Delete("checks").
+	result, err := psql.Delete("checks").
 		Where(
 			sq.And{
 				sq.Gt{
@@ -37,5 +37,14 @@ func (lifecycle *checkLifecycle) RemoveExpiredChecks(recyclePeriod time.Duration
 		RunWith(lifecycle.conn).
 		Exec()
 
-	return err
+	if err != nil {
+		return 0, err
+	}
+
+	affected, err := result.RowsAffected()
+	if err != nil {
+		return 0, err
+	}
+
+	return int(affected), nil
 }
--- a/atc/db/dbfakes/fake_check_lifecycle.go
+++ b/atc/db/dbfakes/fake_check_lifecycle.go
@@ -2,29 +2,31 @@
 package dbfakes
 
 import (
-	"sync"
-	"time"
+	sync "sync"
+	time "time"
 
-	"github.com/concourse/concourse/atc/db"
+	db "github.com/concourse/concourse/atc/db"
 )
 
 type FakeCheckLifecycle struct {
-	RemoveExpiredChecksStub        func(time.Duration) error
+	RemoveExpiredChecksStub        func(time.Duration) (int, error)
 	removeExpiredChecksMutex       sync.RWMutex
 	removeExpiredChecksArgsForCall []struct {
 		arg1 time.Duration
 	}
 	removeExpiredChecksReturns struct {
-		result1 error
+		result1 int
+		result2 error
 	}
 	removeExpiredChecksReturnsOnCall map[int]struct {
-		result1 error
+		result1 int
+		result2 error
 	}
 	invocations      map[string][][]interface{}
 	invocationsMutex sync.RWMutex
 }
 
-func (fake *FakeCheckLifecycle) RemoveExpiredChecks(arg1 time.Duration) error {
+func (fake *FakeCheckLifecycle) RemoveExpiredChecks(arg1 time.Duration) (int, error) {
 	fake.removeExpiredChecksMutex.Lock()
 	ret, specificReturn := fake.removeExpiredChecksReturnsOnCall[len(fake.removeExpiredChecksArgsForCall)]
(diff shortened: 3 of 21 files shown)


atc/engine/engine.go
```diff
@@ -347,13 +347,16 @@ func (c *engineCheck) Run(logger lager.Logger) {
...
 	}
+	c.trackStarted(logger)
 
 	step, err := c.builder.CheckStep(c.check)
 	if err != nil {
 		logger.Error("failed-to-create-check-step", err)
 		return
 	}
 
+	defer c.trackFinished(logger)
```
pivotal-jwinters: Small nit but can we put this side by side with trackStarted?

atc/engine/engine.go
```diff
@@ -392,3 +395,42 @@ func (c *engineCheck) clearRunState() {
 	id := fmt.Sprintf("check:%v", c.check.ID())
 	c.trackedStates.Delete(id)
 }
+
+func (c *engineCheck) trackStarted(logger lager.Logger) {
+	found, err := c.check.Reload()
```
pivotal-jwinters: I wonder how valuable it will actually be to have the check pending duration. If we have to requery the db for this info maybe it isn't worth it?

atc/engine/engine.go
```diff
@@ -392,3 +395,42 @@ func (c *engineCheck) clearRunState() {
...
+
+	metric.CheckStarted{
+		CheckName:             c.check.Plan().Check.Name,
+		ResourceConfigScopeID: c.check.ResourceConfigScopeID(),
+		CheckPendingDuration:  c.check.StartTime().Sub(c.check.CreateTime()),
+	}.Emit(logger)
+}
+
+func (c *engineCheck) trackFinished(logger lager.Logger) {
+	found, err := c.check.Reload()
```
pivotal-jwinters: same here

atc/db/check.go
```diff
@@ -232,6 +240,8 @@ func (c *check) finish(status CheckStatus, checkError error) error {
 		return err
 	}
 
+	c.endTime = endTime
```
cirocosta: is it relevant to test that now `Start()` and `finish()` mutate the `struct` adding their `startTime` and `endTime` respectively?

atc/db/check_lifecycle.go
```diff
@@ -37,5 +37,14 @@ func (lifecycle *checkLifecycle) RemoveExpiredChecks(recyclePeriod time.Duration
...
+	if err != nil {
+		return 0, err
+	}
+
+	affected, err := result.RowsAffected()
+	if err != nil {
+		return 0, err
+	}
+
+	return int(affected), nil
```
cirocosta: hey, while it's fine to go from `int` to `int64` as 

>

(see https://golang.org/pkg/builtin/#int)

atc/db/dbfakes/fake_check_lifecycle.go
```diff
@@ -2,29 +2,31 @@
 package dbfakes
 
 import (
-	"sync"
-	"time"
+	sync "sync"
```
cirocosta: was that a formatting issue? I'm not sure we need to rename these imports to the same package names 🤔

cirocosta:
@pivotal-jwinters if that's something that we care about keeping track over time, should we have a more direct metric for specifically that? e.g., capturing the values from https://github.com/concourse/concourse/blob/342dc8b21a6bf841bfeaeb77fab67199186e9470/atc/lidar/scanner.go#L43-L53? Or, is the sum of `resources + resourceTypes` there not the same as the total number of `scopeIds` that one would get from counting the number of unique `scopeId`-labelled entries? (I might be interpreting this in the wrong way 😅 )

pivotal-jwinters:
@cirocosta yeah the sum(resources and resource_types) doesn't give you the accurate number of checks you should run. We do all the de-duplication behind the scenes, so only the unique number of scope_ids will give you that number.

atc/db/dbfakes/fake_check_lifecycle.go
xtremerui: It is a fake file that is generated we got no control here.

atc/db/check_lifecycle.go
xtremerui: I will have to do the int() convertion either here or when the meter.IncDelta is called [here](https://github.com/concourse/concourse/blob/9b9c15e5af33ab9c4176875e5485b48397b91e12/atc/metric/meter.go#L11)

We are using the IncDelta several places so I would prefer keep it as it is right now and have a seperate issue for int & int64 row affected counts.

atc/db/check.go
xtremerui: Tests are updated. Thx!

atc/db/dbfakes/fake_check_lifecycle.go
cirocosta: whoops that's true 😅

atc/db/check_lifecycle.go
cirocosta: sgtm! 👍

xtremerui:
@cirocosta I think updating the newrelic and prometheus is beyond the scope of the original issue for 2 reasons:

1. There are plenty of emitters that available and I believe at least `prometheus` emitter in concourse is one of those that created by community and got ported in as PR. I have seen couple changes of `prometheus` are from the community that IMHO we should keep it this way. Otherwise the work of maintaining all these emitters could grow heavily. 
2. Every emitter is different in implementation. As for myself who are not a experienced prometheus user, I don't feel my way of updating it to consume and convert concourse events is the best for the concourse prometheus users so I would prefer leave it to the community and hope someone could submit a PR for it.

Thx!

vito:
(Decided in IPM to do it as part of same PR.)

atc/metric/emitter/prometheus.go
```diff
@@ -334,13 +340,62 @@ func (config *PrometheusConfig) NewEmitter() (metric.Emitter, error) {
...
+	prometheus.MustRegister(checkEnqueueVec)
+
+	checkQueueVec := prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: "concourse",
+			Subsystem: "lidar",
+			Name:      "check_queue_size",
+			Help:      "Records the size of check queue",
+		},
+		[]string{},
```
cirocosta: if we're not passing any labels, it might be better to go with a regular Counter (not `Vec`)

atc/metric/emitter/prometheus.go
```diff
@@ -797,8 +884,30 @@ func (emitter *PrometheusEmitter) checkMetric(logger lager.Logger, event metric.
...
 	}
+	checkStatus, exists := event.Attributes["check_status"]
+	if !exists {
+		logger.Error("failed-to-find-check-status-in-event", fmt.Errorf("expected check_status to exist in event.Attributes"))
+		return
+	}
+
+	emitter.checksVec.WithLabelValues(scopeID, checkName, checkStatus).Inc()
+
+	switch checkStatus {
```
cirocosta: rather than going with 3 metrics, we could actually go with a single one - a histogram will always produce a `total`, and  given that we're already passing `checkStatus`, one would be able to do like `concourse_lidar_checks_duration_count`:

```go
// github.com/prometheus/client_golang/prometheus/histogram.go
//
func (h *histogram) Observe(v float64) {
        // ...

	// Increment count last as we take it as a signal that the observation
	// is complete.
	atomic.AddUint64(&hotCounts.count, 1)
}
```
(https://github.com/prometheus/client_golang/blob/333f01cef0d61f9ef05ada3d94e00e69c8d5cdda/prometheus/histogram.go#L291-L293)

given that histograms will also generate `N` timeseries  (being `N` the size of the number of buckets) which gets multiplied by the number of other possible labels, it'd be good to not have `scopeID` and `checkName` for this one then

cirocosta:
Hey, 

if I understood correctly, comparing to what was desired by the issue, we end up
with the following (w/ some suggestions):


- Number of checks queued (per ATC)
	- counter `check_enqueue{}`
		- better w/out checkName in prometheus

- Number of checks GC'd
	- counter `checks_deleted{}`
		- **missing in prometheus**

- Time it takes to run each check (start time - end time)
- Check status (success/failure/error)
	- histogram `check_duration_*{status=(succeded|failed|errored)}`
		- better w/out checkName and scopeId as this multiplies by the
		  number of buckets for the histogram

- Time between queue and check started (create time - start time)
	- do we have this one?


*extra*:
- Size of the queue as seen by an ATC
	- counter check_queue_size{}

Wdyt?

Thanks!

atc/lidar/scanner.go
```diff
@@ -112,6 +113,10 @@ func (s *scanner) check(checkable db.Checkable, resourceTypes db.ResourceTypes)
 		s.logger.Debug("check-already-exists")
 	}
 
+	metric.CheckEnqueue{
+		CheckName: checkable.Name(),
```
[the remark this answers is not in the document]
xtremerui: We only knows `checkable` name here. `scope_id` is not available in `checkable` interface.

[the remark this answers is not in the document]
xtremerui: `scope_id` is not available in `checkable` interface.

atc/metric/emitter/prometheus.go
xtremerui: I think from comment https://github.com/concourse/concourse/pull/4692#pullrequestreview-311275172 `scopeID` is useful still? 

`checkName` is not a must indeed.

xtremerui:
@cirocosta I don't think prometheus implements any periodic delete event right? For example there isn't prometheus implemetation of event "volumed deleted" adnd "container deleted".  Is there a reason for that? Does it apply to "checks deleted" as well?

atc/lidar/scanner.go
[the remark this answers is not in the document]
pivotal-jwinters: I think it should be fairly easy to add to the `checkable` interface. The `resource_type` query already joins the `scopes` table, I think you'll just need to scan it into a new property.

PR #4698, opened 2019-11-01
Add batching to new relic

### LINKED ISSUE #4540: Newrelic metrics emitter losing most events

taylorsilva:
# Existing Issue

Fixes #4540

This PR adds three flags to the NewRelic emitter:
- `CONCOURSE_NEWRELIC_BATCH_DURATION`
- `CONCOURSE_NEWRELIC_BATCH_SIZE`
- `CONCOURSE_NEWRELIC_COMPRESSION_DISABLED`

This change will allow users to adjust the batch size that the
NewRelic emitter uses with the `CONCOURSE_NEWRELIC_BATCH_SIZE` flag.

They can also adjust the frequency that data is emitted by
changing `CONCOURSE_NEWRELIC_BATCH_DURATION`.

We now compress the batch by default. Users can turn this off
by using the `CONCOURSE_NEWRELIC_COMPRESSION_DISABLED` flag.

This also includes the changes made by @xtreme-sameer-vohra in #4617 

CC: @pivotal-bin-ju @xtreme-sameer-vohra 

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/metric/emitter/newrelic.go
+++ b/atc/metric/emitter/newrelic.go
@@ -2,8 +2,11 @@ package emitter
 
 import (
 	"bytes"
+	"compress/gzip"
 	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"strings"
 	"time"
@@ -20,22 +23,29 @@ type (
 	}
 
 	NewRelicEmitter struct {
-		client     *http.Client
-		url        string
-		apikey     string
-		prefix     string
-		containers *stats
-		volumes    *stats
+		Client              *http.Client
+		Url                 string
+		apikey              string
+		prefix              string
+		containers          *stats
+		volumes             *stats
+		BatchSize           int
+		BatchDuration       time.Duration
+		CompressionDisabled bool
+		LastEmitTime        time.Time
+		NewRelicBatch       []NewRelicEvent
 	}
 
 	NewRelicConfig struct {
-		AccountID     string `long:"newrelic-account-id" description:"New Relic Account ID"`
-		APIKey        string `long:"newrelic-api-key" description:"New Relic Insights API Key"`
-		ServicePrefix string `long:"newrelic-service-prefix" default:"" description:"An optional prefix for emitted New Relic events"`
+		AccountID           string        `long:"newrelic-account-id" description:"New Relic Account ID"`
+		APIKey              string        `long:"newrelic-api-key" description:"New Relic Insights API Key"`
+		ServicePrefix       string        `long:"newrelic-service-prefix" default:"" description:"An optional prefix for emitted New Relic events"`
+		BatchSize           uint64        `long:"newrelic-batch-size" default:"2000" description:"Number of events to batch together before emitting"`
+		BatchDuration       time.Duration `long:"newrelic-batch-duration" default:"60s" description:"Length of time to wait between emitting until all currently batched events are emitted"`
+		CompressionDisabled bool          `long:"newrelic-batch-compression-disabled" description:"Compress the batch before emitting"`
 	}
 
-	singlePayload map[string]interface{}
-	fullPayload   []singlePayload
+	NewRelicEvent map[string]interface{}
 )
 
 func init() {
@@ -54,64 +64,22 @@ func (config *NewRelicConfig) NewEmitter() (metric.Emitter, error) {
 	}
 
 	return &NewRelicEmitter{
-		client:     client,
-		url:        fmt.Sprintf("https://insights-collector.newrelic.com/v1/accounts/%s/events", config.AccountID),
-		apikey:     config.APIKey,
-		prefix:     config.ServicePrefix,
-		containers: new(stats),
-		volumes:    new(stats),
+		Client:              client,
+		Url:                 fmt.Sprintf("https://insights-collector.newrelic.com/v1/accounts/%s/events", config.AccountID),
+		apikey:              config.APIKey,
+		prefix:              config.ServicePrefix,
+		containers:          new(stats),
+		volumes:             new(stats),
+		BatchSize:           int(config.BatchSize),
+		BatchDuration:       config.BatchDuration,
+		CompressionDisabled: config.CompressionDisabled,
+		LastEmitTime:        time.Now(),
+		NewRelicBatch:       make([]NewRelicEvent, 0),
 	}, nil
 }
 
-func (emitter *NewRelicEmitter) simplePayload(logger lager.Logger, event metric.Event, nameOverride string) singlePayload {
-	name := nameOverride
-	if name == "" {
-		name = strings.Replace(event.Name, " ", "_", -1)
-	}
-
-	eventType := fmt.Sprintf("%s%s", emitter.prefix, name)
-
-	payload := singlePayload{
-		"eventType": eventType,
-		"value":     event.Value,
-		"state":     string(event.State),
-		"host":      event.Host,
-		"timestamp": event.Time.Unix(),
-	}
-
-	for k, v := range event.Attributes {
-		payload[fmt.Sprintf("_%s", k)] = v
-	}
-	return payload
-}
-
-func (emitter *NewRelicEmitter) emitPayload(logger lager.Logger, payload fullPayload) {
-	payloadJSON, err := json.Marshal(payload)
-	if err != nil {
-		logger.Error("failed-to-serialize-payload", err)
-		return
-	}
-
-	req, err := http.NewRequest("POST", emitter.url, bytes.NewBuffer(payloadJSON))
-	if err != nil {
-		logger.Error("failed-to-construct-request", err)
-	}
-
-	req.Header.Add("Content-Type", "application/json")
-	req.Header.Add("X-Insert-Key", emitter.apikey)
-
-	resp, err := emitter.client.Do(req)
-	if err != nil {
-		logger.Error("failed-to-send-request",
-			errors.Wrap(metric.ErrFailedToEmit, err.Error()))
-		return
-	}
-
-	resp.Body.Close()
-}
-
 func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
-	payload := make(fullPayload, 0)
+	logger = logger.Session("new-relic")
 
 	switch event.Name {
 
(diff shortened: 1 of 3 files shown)


atc/metric/emitter/newrelic.go
```diff
@@ -54,64 +64,22 @@ func (config *NewRelicConfig) NewEmitter() (metric.Emitter, error) {
...
+		Url:                 fmt.Sprintf("https://insights-collector.newrelic.com/v1/accounts/%s/events", config.AccountID),
+		apikey:              config.APIKey,
+		prefix:              config.ServicePrefix,
+		containers:          new(stats),
+		volumes:             new(stats),
+		BatchSize:           int(config.BatchSize),
+		BatchDuration:       config.BatchDuration,
+		CompressionDisabled: config.CompressionDisabled,
+		LastEmitTime:        time.Now(),
+		NewRelicBatch:       make([]NewRelicEvent, 0),
```
cirocosta: Hey, it came to my mind that we **might** (in further iterations) leverage slices w/ a predefined capacity (so that we avoid having to allocate memory on demand as this increases in the `append(`s we perform 🤔  

The only concern I'd have is with the subsequent slice reset on `emitBatch`, where we'd need to do something like `batch = batch[:0]` to keep the capacity, but I'm not sure if that's "leaky" (w/ regards to memory) by default or not hmmm

(ps.: just a comment, I don't think we need that right now 😁 )

atc/metric/emitter/newrelic.go
```diff
@@ -173,16 +144,117 @@ func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
...
 		// didn't; therefore, be consistently inconsistent and use the
 		// concourse metric names, not our translation layer.
 		singlePayload["metric"] = event.Name
-		payload = append(payload, singlePayload)
+		emitter.NewRelicBatch = append(emitter.NewRelicBatch, singlePayload)
+	}
+
+	duration := time.Since(emitter.LastEmitTime)
+	if len(emitter.NewRelicBatch) >= emitter.BatchSize || duration >= emitter.BatchDuration {
+		logger.Debug("pre-emit-batch", lager.Data{
```
cirocosta: oh this can be sooo chatty - do you think this is necessary? or was this something used during dev that was just left?

atc/metric/emitter/newrelic.go
```diff
@@ -173,16 +144,117 @@ func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
...
+			"current-duration":   duration,
+		})
+		emitter.submitBatch(logger)
+	}
+}
+
+// NewRelic has strict requirements around the structure of the events
+// Keys must be alphanumeric and can contain hyphens or underscores
+// Values must be sting, int, or unix timestamps. No maps/arrays.
+func (emitter *NewRelicEmitter) transformToNewRelicEvent(logger lager.Logger, event metric.Event, nameOverride string) NewRelicEvent {
```
cirocosta: I don't think we need a `logger` here 😁   seems like deadcode

atc/metric/emitter/newrelic.go
```diff
@@ -173,16 +144,117 @@ func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
...
+	name := nameOverride
+	if name == "" {
+		name = strings.Replace(event.Name, " ", "_", -1)
+	}
+
+	eventType := fmt.Sprintf("%s%s", emitter.prefix, name)
```
cirocosta: I'd go with a plain `+` here rather than `Sprintf` (which would do all of the formatting work) as it'd not look terrible

```
BenchmarkFormatting-12           6101223               196 ns/op
BenchmarkAddition-12            13326642                87.4 ns/op
```

```go
func BenchmarkFormatting(b *testing.B) {
	var res string
	for n := 0; n < b.N; n++ {
		foo := strconv.Itoa(b.N)
		bar := strconv.Itoa(b.N)

		res = fmt.Sprintf("%s%s", foo, bar)
	}

	fmt.Fprintln(ioutil.Discard, res)
}

func BenchmarkAddition(b *testing.B) {
	var res string
	for n := 0; n < b.N; n++ {
		foo := strconv.Itoa(b.N)
		bar := strconv.Itoa(b.N)

		res = foo + bar
	}

	fmt.Fprintln(ioutil.Discard, res)
}
```

atc/metric/emitter/newrelic.go
```diff
@@ -173,16 +144,117 @@ func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
...
+	payload := NewRelicEvent{
+		"eventType": eventType,
+		"value":     event.Value,
+		"state":     string(event.State),
+		"host":      event.Host,
+		"timestamp": event.Time.Unix(),
+	}
+
+	for k, v := range event.Attributes {
+		payload[fmt.Sprintf("_%s", k)] = v
```
cirocosta: (same as above)

atc/metric/emitter/newrelic.go
```diff
@@ -173,16 +144,117 @@ func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
...
+func (emitter *NewRelicEmitter) emitBatch(logger lager.Logger, payload []NewRelicEvent) {
+	batch, err := emitter.marshalJSON(logger, payload)
+	if err != nil {
+		logger.Error("failed-to-marshal-batch", err)
+		return
+	}
+
+	req, err := http.NewRequest("POST", emitter.Url, batch)
+	if err != nil {
+		logger.Error("failed-to-construct-request", err)
```
cirocosta: oh, I think you want to return here (`req` would be nil after this err)

atc/metric/emitter/newrelic.go
```diff
@@ -173,16 +144,117 @@ func (emitter *NewRelicEmitter) Emit(logger lager.Logger, event metric.Event) {
...
+	}
+
+	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("X-Insert-Key", emitter.apikey)
+	if !emitter.CompressionDisabled {
+		req.Header.Add("Content-Encoding", "gzip")
+	}
+
+	resp, err := emitter.Client.Do(req)
```
cirocosta: I wonder if we should put a timeout on these requests 😬  

might be better in a different issue though - wdyt?

atc/metric/emitter/newrelic.go
taylorsilva: This was copied from the influxdb emitter 😬

atc/metric/emitter/newrelic.go
taylorsilva: That's fine. This is the old code, it just looks new because we moved the functions around.

atc/metric/emitter/newrelic.go
taylorsilva: (same comment for me as above as well)

atc/metric/emitter/newrelic.go
taylorsilva: @pivotal-bin-ju mentioned that we already set a timeout when we create the `client`. See line 61.

atc/metric/emitter/newrelic.go
```diff
@@ -20,22 +23,29 @@ type (
...
 	NewRelicConfig struct {
-		AccountID     string `long:"newrelic-account-id" description:"New Relic Account ID"`
-		APIKey        string `long:"newrelic-api-key" description:"New Relic Insights API Key"`
-		ServicePrefix string `long:"newrelic-service-prefix" default:"" description:"An optional prefix for emitted New Relic events"`
+		AccountID           string        `long:"newrelic-account-id" description:"New Relic Account ID"`
+		APIKey              string        `long:"newrelic-api-key" description:"New Relic Insights API Key"`
+		ServicePrefix       string        `long:"newrelic-service-prefix" default:"" description:"An optional prefix for emitted New Relic events"`
+		BatchSize           uint64        `long:"newrelic-batch-size" default:"2000" description:"Number of events to batch together before emitting"`
+		BatchDuration       time.Duration `long:"newrelic-batch-duration" default:"60s" description:"Length of time to wait between emitting until all currently batched events are emitted"`
+		CompressionDisabled bool          `long:"newrelic-batch-compression-disabled" description:"Compress the batch before emitting"`
```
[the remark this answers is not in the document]
taylorsilva: This was originally `CompressionEnabled`. The problem we ran into that resulted in the name `CompressionDisabled` is that we want compression on by default but you can't specify `default:"true"` with the `go-flags` library, as you correctly guessed. `go-flags` throws an error during compilation. We wanted to avoid having any checks around whether `CompressionEnabled` has been set or not; it just gets tricky to think about and you get those weird `!= true/false` type of `if` checks that are just annoying to reason about.

If we had access to this: https://godoc.org/github.com/jessevdk/go-flags#Option.IsSet
then we could do `CompressionEnabled` nicely. It's not clear to me how to pull this `Option` type in though.

I'm not happy that we settled on `CompressionDisabled` but that's what motivated us to stick with it.

taylorsilva:
@cirocosta changed as per your comments.

Still up for talking about `CompressionDisabled`. I'd appreciate help on it. I'm down for even removing it as an option all together as I don't think many people would ever bother not using it.

taylorsilva:
Made a change to the tests. We weren't verifying that each batch contained the correct number of events in it. This is now checked for each test.

skreddy6673:
@cirocosta Is there any New relic dashboard that i can refer just like Grafana dashboard?

cirocosta:
@skreddy6673, from us, no 😢  we're mainly using influxdb, datadog and prometheus across our deployments

that might be something to ask in the forum / discord:
- https://discuss.concourse-ci.org/
- https://discordapp.com/invite/MeRxXKW

skreddy6673:
Thanks @cirocosta 
 I'm using New relic and reached to some level where i replicated the Grafana Dashboard, but looking for more visibility and alerting, and to validate the New relic board.

PR #4699, opened 2019-11-01
revert #4661

pivotal-jamie-klassen:
# Why do we need this PR?
v5.7.0 was released with a breaking change and no way to opt out. We can afford to more carefully consider how to extend the CF auth connector without excluding certain use cases.

* downgrade dex

# Contributor Checklist
- [x] ~Unit tests~

# Reviewer Checklist
- [ ] Code reviewed
- [x] ~Tests reviewed~
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -43,7 +43,7 @@ require (
 	github.com/cloudfoundry/go-socks5 v0.0.0-20180221174514-54f73bdb8a8e // indirect
 	github.com/cloudfoundry/socks5-proxy v0.0.0-20180530211953-3659db090cb2 // indirect
 	github.com/concourse/baggageclaim v1.6.5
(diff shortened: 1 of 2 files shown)


evanchaoli:
@pivotal-jamie-klassen If we don't use CF auth, will it still be breaking?

pivotal-jamie-klassen:
@evanchaoli no, the 'breaking' aspect is that before 5.7.0, any roles specified with `--cf-space` would be granted only to users with the Developer role in that space. As of 5.7.0, the same configuration means that the same Concourse role would be granted to users with _any_ role in that space (including Auditor and Manager) -- https://github.com/concourse/concourse/releases#4535, https://github.com/concourse/concourse/issues/4535#issuecomment-548933195

PR #4702, opened 2019-11-04
make the team names clickable to filter the dashboard view

robwhitby:
# Why do we need this PR?
On the dashboard page I expect to be able to click on the team names to filter the view

*  team name headings are now links that apply the filter `team:<team-name>`

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/web/elm/tests/DashboardTests.elm
+++ b/web/elm/tests/DashboardTests.elm
@@ -866,6 +866,39 @@ all =
                     |> Common.queryView
                     |> Query.find [ class "dashboard-team-name-wrapper" ]
                     |> Query.has [ style "letter-spacing" ".2em" ]
+        , describe "clicking on team name"
+            [ test
+                "filters by team when there are pipelines"
+              <|
+                \_ ->
+                    whenOnDashboard { highDensity = False }
+                        |> givenDataAndUser
+                            (oneTeamOnePipeline "teamA")
+                            (userWithRoles [ ( "team", [ "owner" ] ) ])
+                        |> Tuple.first
+                        |> Common.queryView
+                        |> Query.find (teamHeaderSelector ++ [ containing [ text "teamA" ] ])
+                        |> Query.find [ class "dashboard-team-name"]
+                        |> Event.simulate Event.click
+                        |> Event.expect
+                            (ApplicationMsgs.Update <|
+                                Msgs.FilterMsg <|
+                                    "team:teamA"
+                            )
+            , test "does not filter by team when there are no pipelines" <|
+                \_ ->
+                    whenOnDashboard { highDensity = False }
+                        |> givenDataAndUser
+                            (apiData [ ( "teamA", [] ) ])
+                            (userWithRoles [ ( "team", [ "owner" ] ) ])
+                        |> Tuple.first
+                        |> Common.queryView
+                        |> Query.find (teamHeaderSelector ++ [ containing [ text "teamA" ] ])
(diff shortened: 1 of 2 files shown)


taylorsilva:
Gave this PR a test and works as expected, nice job! The team names don't appear any different on the dashboard and I'm no designer so not sure if we want something to help indicate to users that they can click the team names to filter. Maybe it's fine as-is and users will discover this on their own? @Lindsayauchin or @matthewpereira  would have some insight into this.

web/elm/tests/DashboardTests.elm
```diff
@@ -866,6 +866,39 @@ all =
                     |> Common.queryView
                     |> Query.find [ class "dashboard-team-name-wrapper" ]
                     |> Query.has [ style "letter-spacing" ".2em" ]
+        , describe "clicking on team name"
+            [ test
+                "filters by team when there are pipelines"
+              <|
```
zoetian: Hey @robwhitby, thanks for this PR! Do you mind doing a `yarn format` to make it easier on eyes? Thanks!

robwhitby: yep, no problem

matthewpereira:
Thank you for the PR!

I'll check in with @Lindsayauchin to chat about this. I get the feeling like there should be a way to cancel the action without moving your mouse all the way to the search bar in case of an accidental click.

I'm also curious to know how many users share this expectation about clicking team names to change filtering - this could cause confusion if we don't communicate what's going on in the right way. If we want to move forward with this sort of functionality, a dedicated button might be a better experience. I can imagine some users wanting to double click a team name to copy it to the clipboard, and having their dashboard filtering change in the process.

@robwhitby , could you speak about why you might be filtering out all the rest of the teams from view so that we can get a better idea of the scenario where this is most useful? Any insight into the part of your workflow where this would be useful would be helpful.

robwhitby:
Thanks. To be honest, when I added this I thought it was an obvious missing feature, I've seen many people click the team name expecting it to be a link, in the same way that you can click the pipeline name.

matthewpereira:
Hey @robwhitby! 

TLDR: Rather than create a new way for an element on the pipeline selection view to change the search/filtering state, I think it'll be more useful to scope navigation elsewhere and leave these team labels as labels.

Thanks for your patience with this issue. Over the past couple months we've been doing some extensive work on the dashboard UX, and have a number of initiatives that I think should cover some of the  in order to keep the experience from getting confusing. 

On the topic of navigation and displaying the right information when only some pipelines need to be visible, we're starting work on Favorite Pipelines (#5434) and have work ongoing for parts of the Archiving Pipelines work (#5055) which will tidy up the web interface, introduce new ways to interact with the sidebar, and change the set of pipelines that's visible in the CI monitor view. I'm hoping that these collateral changes will make it a lot easier for people to limit the scope of what they're seeing listed in new and useful ways, and speeding up the process of navigating to the pipelines they need - which I believe was part of the outcome that you were contributing to with this PR.

If you want to chat about the workflows that would be aided by scoping down the view to a single team, I'd still be happy to have that conversation with you - it'd be great to learn about any insights or specific needs you have for navigation so that we can make sure we're addressing those needs. Let me know!

PR #4703, opened 2019-11-04
Add release notes for dex downgrade

deniseyu:
[#4699](https://github.com/concourse/concourse/pull/4699)

for v5.7.1 patch release


release-notes/v5.7.1.md
```diff
@@ -0,0 +1,4 @@
+#### <sub><sup><a name="4699" href="#4699">:link:</a></sup></sub> fix
+
+* v5.7.0 added a breaking change to the CloudFoundry auth connector, which is reverted in this release.
```
vito: Maybe we should say that we're going to try to find a way to support both configurations, and that we're just reverting it until then.

PR #4707, opened 2019-11-04
atc/worker: make garden client timeout configurable

### LINKED ISSUE #4652: Timeout failures to provision containers
## Bug Report

We are seeing many timeout failures to provision containers. Referring to issue #4562, we believe we are facing a similar problem. However, we're not sure that the fix for issue #4562 will address the timeout issue. If the timeout is set at a constant 5 minutes (https://github.com/concourse/concourse/blob/0e1b195bf964cf11e88c6cbd75775f5160cce1df/atc/config.go#L19), but we are still hitting the timeout, our containers will never successfully provision.

## Steps to Reproduce

We are using a fairly large container image (0.5Gb), with overlay storage_driver. Some of our tasks require privileged containers. Also, we are scheduling many containers at once (random placement strategy), but our concourse deployment has 5 large (16cpu, 60Gb memory) workers. In addition, we are only running one pipeline (50ish jobs) on this deployment. We are running Concourse 5.6.0. 

## Expected Results

Either faster container provisioning, or a configurable container provision timeout.

## Actual Results

Many jobs are orange with the error message: `Post http://127.0.0.1:8080/containers: net/http: request canceled (Client.Timeout exceeded while awaiting headers)`. Also, our findings are consistent with the description in issue #4562.

## Version Info

* Concourse version: 5.6.0
* Deployment type: BOSH 
* Infrastructure/IaaS: GCP
* Did this used to work? yes, recently upgraded from 4 to 5.6.0

cirocosta:
# Existing Issue

Fixes #4652.

# Why do we need this PR?

In some cases documented in the issue, we've seen evidence of the hardcoded 5 minute timeout being reached.

Rather than using a hardcoded timeout of 5 minutes for http requests to the Garden server, both from the web and from the workers, this PR makes this value configurable. From an operational perspective, operators should have greater flexibility to determine what constitutes  a reasonable timeout.

Note: This timeout is configured in both the web and the worker because they use independent garden clients for different reasons. The web's timeout represents the time that the web will wait for Garden to spin up a build container (the UI will turn orange when this timeout is hit); whereas the worker's timeout is used internally for sweeping unused containers.

# Contributor Checklist

# Reviewer Checklist

- [x] Code reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [x] TODO: New config flags added? Ensure that they are added to the [BOSH](https://github.com/concourse/concourse-bosh-release) 

### DIFF
--- a/worker/workercmd/worker_linux.go
+++ b/worker/workercmd/worker_linux.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strings"
 	"syscall"
+	"time"
 
 	"code.cloudfoundry.org/lager"
 	"code.cloudfoundry.org/localip"
@@ -33,6 +34,8 @@ type GardenBackend struct {
 	GardenConfig flag.File `long:"config"               description:"Path to a config file to use for Garden. You can also specify Garden flags as env vars, e.g. 'CONCOURSE_GARDEN_FOO_BAR=a,b' for '--foo-bar a --foo-bar b'."`
 
 	DNS DNSConfig `group:"DNS Proxy Configuration" namespace:"dns-proxy"`
+
+	RequestTimeout time.Duration `long:"request-timeout" default:"5m" description:"How long to wait for requests to Garden to complete. 0 means no timeout."`
 }
 
 func (cmd WorkerCommand) LessenRequirements(prefix string, command *flags.Command) {
--- a/worker/workercmd/worker_nonlinux.go
+++ b/worker/workercmd/worker_nonlinux.go
@@ -4,6 +4,7 @@ package workercmd
 
 import (
 	"runtime"
+	"time"
 
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc"
@@ -11,7 +12,9 @@ import (
 	"github.com/tedsuo/ifrit"
 )
 
-type GardenBackend struct{}
+type GardenBackend struct {
+	RequestTimeout time.Duration `long:"request-timeout" default:"5m" description:"How long to wait for requests to Garden to complete. 0 means no timeout."`
+}
 
 type Certs struct{}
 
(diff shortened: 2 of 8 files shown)


worker/workercmd/worker_linux.go
```diff
@@ -33,6 +34,8 @@ type GardenBackend struct {
 	GardenConfig flag.File `long:"config"               description:"Path to a config file to use for Garden. You can also specify Garden flags as env vars, e.g. 'CONCOURSE_GARDEN_FOO_BAR=a,b' for '--foo-bar a --foo-bar b'."`
 
 	DNS DNSConfig `group:"DNS Proxy Configuration" namespace:"dns-proxy"`
+
+	RequestTimeout time.Duration `long:"request-timeout" default:"5m" description:"How long to wait for requests to Garden to complete. 0 means no timeout."`
```
kcmannem: sucks that we have to specify the same timeout in two different places. It's not obvious you'd have to change them in both places.

worker/workercmd/worker_nonlinux.go
```diff
@@ -4,14 +4,17 @@ package workercmd
...
+	"time"
 
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc"
 	"github.com/jessevdk/go-flags"
 	"github.com/tedsuo/ifrit"
 )
 
-type GardenBackend struct{}
+type GardenBackend struct {
```
kcmannem: instead of putting them in both worker_linux and worker_nonlinux, can we move this to worker_config. These are where common flags reside.

https://github.com/concourse/concourse/blob/0e1b195bf964cf11e88c6cbd75775f5160cce1df/worker/workercmd/worker.go#L27

cirocosta: gooood catch, we indeed thought about that, @kcmannem - we end up putting under `GardenBackend` to make it part of the Garden configuration group 🤔  otherwise, it'd feel kinda out of place (imo)

e.g., `DNSProxyEnable` is under `Garden` too (although yeah, we don't have that for non-linux)

cirocosta:
> your thoughts about moving timeout to a single location

what do you mean by that, more specifically? `worker_linux` vs `worker_nonlinux`, or `web` vs `worker`?

Regarding the former, I think it makes sense to be separated (as they're different timeouts), regarding the first, see comment above ^ 😁

Issue #1200, opened 2017-05-23
How do you use `set-pipeline`?


vito:
Currently `set-pipeline` is manually invoked by a user to submit a pipeline config from their local machine. This is great for rapid iteration, but also goes against a Concourse principle of having versioned, persistent, reproducible config.

Are you still using `set-pipeline` manually? Have you written scripts to automate it? Are you continuously configuring it using something like the [Concourse Pipelines resource](https://github.com/robdimsdale/concourse-pipeline-resource)? How are you handling credentials?

marco-m:
I am investigating Concourse since a week, with the idea of replacing a Jenkins deployment. The question you are asking is one of the main questions I had. I plan to use scripts and salt-stack exactly for reproducibility. I read about the pipeline resource and I want to investigate it more. Another question I had is why we need to specify the pipeline name in the command line, why not having the pipeline name in the YAML file ? The notion of naming things in the YAML files is already there for every concept. The `--pipeline` option could become required only if the pipeline name in the YAML file is not present...

vito:
@marco-m The name is not in the `.yml` so that you can use a single pipeline template to configure differently-parameterized and thus differently-named pipelines.

Chumper:
What about integrating seeding from a repository directly into concourse?

I would love to have the opportunity the state one or multiple repositories that concourse will then watch (or hook) and seed accordingly.
It doesn't need to have fancy diff mechanics, just the seeding would work fine.

That would make it easier to start with the correct "version everything" mindset.

marco-m:
@vito Thanks for the explanation regarding the lack of pipeline name in the configuration file. Makes sense.

freelock:
We use a bot that calls set-pipeline. Not sure how typical our setup is, but it's working well for us managing 60+ pipelines currently.

We have two pipeline yaml files that differ based on the platform being managed (we manage a lot of Drupal sites and a few WordPress sites -- we have a different pipeline per platform). All of the scripts and supporting tasks are in the same repo as our pipeline. We keep a separate credentials.yml file and a wrapper script that reads the secrets that need to be passed in as separate parameters when a pipeline is set.

We have one Matrix chatroom filled with bots -- our control bot, Concourse notifications, gitolite notifications, notifications from a couple different PM tools we use.

[...]

PR #4708, opened 2019-11-05
feat(atc): add set_pipeline step.

evanchaoli:
# Existing Issue

There is a RFC [concourse/rfc#31](https://github.com/concourse/rfcs/pull/31).

# Why do we need this PR?

This will enable to automatically reset a pipeline from itself.

* Set a new step `set_pipeline` as well as events of it.

Test pipeline:

```yml
resources:
- name: mygit
  type: git
  source:
    uri: git@gitlab.mycompany.com:chaol/fortest.git
    branch: master
    private_key: ((chaol-gitlab-private-key))

jobs:
- name: self-reset
  public: true
  plan:
  - get: mygit
    trigger: true
  - set_pipeline: sp
    params:
      pipeline_name: self
      config: mygit/test_self_reset.yml
      load_vars_from:
       - mygit/credential.yml

- name: normal-job
  public: true
  plan:
  - task: a-task
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: sh
        args:
         - -exc
         - |
           echo hello2
```

There are two jobs in this pipeline. The first job is used to self-reset the pipeline. The second job is a normal job.

The web UI should be updated also, my backend code has been able to successfully set-pipeline, but UI cannot display the entire build. @Twiknight is helping on the UI change.

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


atc/exec/set_pipeline_step.go
```diff
@@ -0,0 +1,229 @@
...
+
+	step.delegate.Starting(logger)
+
+	if spParams.PipelineName == "self" {
+		spParams.PipelineName = step.metadata.PipelineName
+	}
+
+	team, found, err := db.NewTeamFactory(step.dbConn, step.lockFactory).FindTeam(step.metadata.TeamName)
+	if err != nil {
+		logger.Error("EVAN-failed-to-find-team", err)
```
aledeganopix4d: I guess this and the following debug messages starting with `EVAN-` are not meant to be final? :)

evanchaoli: Of course not. This is still under WIP. :)

aledeganopix4d: Sorry, my bad, I've missed that :)

atc/exec/set_pipeline_step.go
```diff
@@ -0,0 +1,229 @@
...
+}
+
+// TaskStep executes a TaskConfig, whose inputs will be fetched from the
+// artifact.Repository and outputs will be added to the artifact.Repository.
+type SetPipelineStep struct {
+	planID      atc.PlanID
+	plan        atc.SetPipelinePlan
+	metadata    StepMetadata
+	delegate    SetPipelineDelegate
+	dbConn      db.Conn
```
pivotal-jwinters: In general, we've tried to stay away from exposing the db directly to each step type. Instead we use the delegate as a proxy to any database operations.

For example: 
https://github.com/concourse/concourse/blob/f6ec622e57d6559fffd676a5df16665745600a7a/atc/exec/check_step.go#L161

evanchaoli:
So far, I have made the feature functional and the existing unit tests pass. @Twiknight also has done UI part changes, unit tests not added yet.

Highlights of the change:

* Add `set_pipeline` step, which works similar to a `task` step, so that user may wrap `on_success`, `on_failure`, etc. steps around it to do error handling, notifying, and so on.
* `set_pipeline` the following parameters (reflecting to `fly set-pipeline` command):
  * `config`: path of pipeline YAML file. The file should come from other resource, like `git`
  * `pipeline_name`: pipeline name to set. If name is `self`, then do self set.
  * `var`: variables passing to pipeline. (`fly set-pipeline` also supports `--yaml-var` option, but I think it's not needed if we define vars in YAML.)
  * `load_vars_from`: variable files. The files should come from other resources, like `git`. 
  * `fail_when_no_diff`: When set to true, if there is pipeline change, the step will fail. Default is false. I add this flag in order to leave the flexibility to users.
* Move config diff code from `fly/command/internal/setpipelinehelper` to `atc/config_diff.go` because the `set_pipeline` step also needs to do pipeline diff.

The web UI looks like below:

self-set

Maybe we need a new icon for the `set_pipeline` step. However we are not good at art, so leaving that to some art talent to contribute with further PR.

@pivotal-jwinters and @vito Can you guys please take an initial review. If no significant comments, we'll start to add unit test cases. For this feature, maybe a testflight case is deserved.

atc/exec/set_pipeline_step.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, err
+	}
+
+	return byteConfig, nil
+}
+
+// FetchConfig streams pipeline configure file and var files from other resources
+// and construct an atc.Config object.
+func (s setPipelineArtifactSource) FetchConfig() (*atc.Config, error) {
+	config, err := s.streamInBytes(s.params.Config)
```
vito: I think we can overload the `file:` field on the step instead of using `params`, like so:

```yml
set_pipeline: pipeline-name
file: path/to/config.yml
```

I would rather not use `params` at all actually, since right now that's sort of a "black box" passed to resources. It feels weird for it to have an expected schema for certain step types.

atc/exec/set_pipeline_step.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+
+		fmt.Fprintf(stdout, "No diff found.\n")
+		step.succeeded = true
+		step.delegate.Finished(logger, ExitStatus(0), fromVersion, fromVersion)
+		return nil
+	}
+
+	fmt.Fprintf(stdout, "Updating the pipeline.\n")
+	pipeline, _, err = team.SavePipeline(spParams.PipelineName, *atcConfig, fromVersion, false)
```
vito: Per the original proposal, this should be using the step name as the pipeline name: https://github.com/vito/rfcs/blob/set-pipeline-step/031-set-pipeline-step/proposal.md#proposal

e.g.:

```yaml
set_pipeline: foo
file: ci/pipelines/foo.yml
```

In v10, assuming you have a project called `ci`, this could be shortened further to:

```yaml
set_pipeline: foo
```

With your proposed `self` behavior, it would just be `set_pipeline: self`.

How does that sound? :slightly_smiling_face:

atc/exec/set_pipeline_step.go
```diff
@@ -0,0 +1,244 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	diffExists := existingConfig.Diff(stdout, *atcConfig)
+	if !diffExists {
+		logger.Debug("no-diff")
+
+		if spParams.FailWhenNoDiff {
```
vito: IMO steps should never fail in this case; they should just silently no-op. I wouldn't even have a flag for it, since this is something Concourse is pretty opinionated about (builds being idempotent). For example I might have one job that configures a bunch of pipelines all at once; it would be pretty common for all but one to be a no-op.

atc/exec/set_pipeline_step.go
```diff
@@ -0,0 +1,244 @@
...
+	if err != nil {
+		return err
+	}
+
+	step.delegate.Starting(logger)
+
+	stdout := step.delegate.Stdout()
+
+	if spParams.PipelineName == "self" {
+		spParams.PipelineName = step.metadata.PipelineName
```
vito: I wonder if this syntax could be resolved at a higher level? :thinking: It feels weird to be peering into the metadata for this. For example, a one-off build could technically be constructed with a `set_pipeline` step, and one-off builds aren't associated to any pipeline, so this would end up being an empty string.

Maybe whoever constructs the step should be responsible for resolving `self` to the pipeline name?

atc/plan.go
```diff
@@ -137,6 +138,14 @@ type TaskPlan struct {
 	VersionedResourceTypes VersionedResourceTypes `json:"resource_types,omitempty"`
 }
 
+type SetPipelinePlan struct {
+	Name              string `json:"name"`
+	Params            Params `json:"params"`
```
vito: Mentioned this already but IMO we shouldn't be using `Params` for this. I would recommend having explicit fields here for `File` and the vars fields.

atc/plan.go
```diff
@@ -137,6 +138,14 @@ type TaskPlan struct {
...
+	ImageArtifactName string `json:"image,omitempty"`
+
+	VersionedResourceTypes VersionedResourceTypes `json:"resource_types,omitempty"`
```
vito: Don't think these are needed.

atc/exec/set_pipeline_step.go
```diff
@@ -0,0 +1,244 @@
...
+// and construct an atc.Config object.
+func (s setPipelineArtifactSource) FetchConfig() (*atc.Config, error) {
+	config, err := s.streamInBytes(s.params.Config)
+	if err != nil {
+		return nil, err
+	}
+
+	staticVarss := []vars.Variables{}
+	if len(s.params.Var) > 0 {
+		staticVarss = append(staticVarss, vars.StaticVariables(s.params.Var))
```
vito: In my fork I planned to use the step-level `vars:` config:

https://github.com/concourse/concourse/compare/feature/set-pipeline-step#diff-cd1e4bcee1dd2a9f163ac91a8232bcdeR327

There's no top-level `load_vars_from` at the moment but it could be added.

vito:
Also, since this started through the RFC process, we'll need to finish it through the RFC process as well: https://github.com/concourse/rfcs#resolution

It's fine to start working on it ahead of time, we just need to make sure what we actually deliver lines up with what everyone agreed on through the RFC. I don't expect any trouble here, it's just due diligence.

I'll try and finalize all of my v10 RFCs and start the resolution process soon, which will involve a blog post announcing it. I'll make note of this PR too. :slightly_smiling_face:

pivotal-jamie-klassen:
Big +1 on generic step events - our consumers of those events care about the different step types really. Like fly does show the task config before a task step but that could be made generic, I think.

atc/db/pipeline.go
```diff
@@ -214,6 +215,32 @@ func (p *pipeline) Reload() (bool, error) {
 	return true, nil
 }
 
+func (p *pipeline) Config() (atc.Config, error) {
+	jobs, err := p.Jobs()
+	if err != nil {
+		return atc.Config{}, fmt.Errorf("failed-to-get-jobs: %s", err)
```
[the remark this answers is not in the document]
evanchaoli: Oops! This code was moved from `atc/api/configserver/get.go`. Anyway, fixed.

atc/exec/set_pipeline_step.go
evanchaoli: @vito I now resolve `self` in the `EvaluatiePlan` method, please if you are ok with that.

atc/exec/set_pipeline_step.go
evanchaoli: I have moved `vars` to step-level and added `var_files`. To our use case, `var_files` should be mandatory.

atc/engine/builder/delegate_factory.go
```diff
@@ -294,6 +298,72 @@ func (d *taskDelegate) Finished(logger lager.Logger, exitStatus exec.ExitStatus)
...
+
+	logger.Debug("starting")
+}
+
+func (d *setPipelineDelegate) Finished(logger lager.Logger, exitStatus exec.ExitStatus, versionBefore db.ConfigVersion, versionAfter db.ConfigVersion) {
+	// PR#4398: close to flush stdout and stderr
+	d.Stdout().(io.Closer).Close()
+	d.Stderr().(io.Closer).Close()
+
+	err := d.build.SaveEvent(event.FinishSetPipeline{
```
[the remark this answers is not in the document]
evanchaoli: I have followed your comment, and made a generate build step delegate.

atc/exec/set_pipeline_step.go
evanchaoli: Sounds good.

atc/engine/builder/delegate_factory.go
```diff
@@ -403,6 +407,56 @@ func (delegate *buildStepDelegate) Stderr() io.Writer {
...
+	})
+	if err != nil {
+		logger.Error("failed-to-save-start-event", err)
+		return
+	}
+
+	logger.Debug("starting")
+}
+
+func (delegate *buildStepDelegate) Finished(logger lager.Logger, succeeded bool) {
```
vito: :+1:

atc/event/types.go
```diff
@@ -36,6 +36,15 @@ const (
 	// finished putting something
 	EventTypeFinishPut atc.EventType = "finish-put"
 
+	// initialize setting pipeline
```
vito:
```suggestion
	// initialize step
```

atc/event/types.go
```diff
@@ -36,6 +36,15 @@ const (
...
+	EventTypeInitialize atc.EventType = "initialize"
+
+	// started setting pipeline
```
vito:
```suggestion
	// started step
```

atc/event/types.go
```diff
@@ -36,6 +36,15 @@ const (
...
+	EventTypeStart atc.EventType = "start"
+
+	// finished setting pipeline
```
vito:
```suggestion
	// finished step
```

atc/exec/artifact/repository.go
```diff
@@ -71,3 +76,37 @@ func (repo *Repository) AsMap() map[Name]worker.ArtifactSource {
 
 	return result
 }
+
+func (repo *Repository) StreamFile(ctx context.Context, logger lager.Logger, path string) (io.ReadCloser, error) {
```
vito: thanks for carrying this over :)

atc/exec/set_pipeline_step.go
```diff
@@ -0,0 +1,243 @@
...
+
+func (s setPipelineSource) Validate() error {
+	if s.step.plan.File == "" {
+		return errors.New("file is not specified")
+	}
+
+	return nil
+}
+
+func (s setPipelineSource) EvaluatePlan(variables vars.Variables) error {
```
vito: This is interesting; what's the use case for evaluating all of these? I only ask because we don't do it for any of the other steps.

I don't think you need to do it for the `vars:` since any vars contained within will be recursively evaluated when resolving the pipeline config.

e.g.:

```yaml
set_pipeline: my-pipeline
vars:
  foo: ((bar))
```

In the pipeline, `((foo))` will resolve to `((bar))` which will then be resolved again to its value. Whereas here it's evaluated ahead-of-time so `((foo))` will resolve directly to the value.

So at least for `vars:` I don't think this is necessary as things should resolve the same. And aside from `vars:` I don't see a whole lot of value in evaluating the vars in `file:` or `var_files:`. :thinking:

atc/exec/set_pipeline_step.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+
+	s.step.plan.File = params["file"].(string)
+	s.step.plan.Vars = params["vars"].(map[string]interface{})
+	s.step.plan.VarFiles = []string{}
+	for _, ele := range params["var_files"].([]interface{}) {
+		s.step.plan.VarFiles = append(s.step.plan.VarFiles, ele.(string))
+	}
+
+	if s.step.plan.Name == "self" {
```
vito: Hmm I guess I was hoping this could be done in `atc/scheduler/factory/build_factory.go` so the `atc.Plan` object is always verbatim (no "self" magic), but it looks like that object doesn't have access to the pipeline name. Maybe it could be passed in?

I think this is a little important because `atc.Plan` types are meant to be fairly "dry" and can be submitted outside of the context of a pipeline, like with `fly execute`. So conceptually it makes sense to me for `self` to be pipeline-config magic that gets de-sugared into a regular build plan.

atc/set_pipeline.go
```diff
@@ -0,0 +1,55 @@
+package atc
```
vito: i think this file can be deleted now?

evanchaoli: Yes, I thought to delete it but forgot.

atc/exec/set_pipeline_step.go
evanchaoli: > This is interesting; what's the use case for evaluating all of these? I only ask because we don't do it for any of the other steps.
> 

I'm a little confused. I think all other steps evaluate `source` and `params`, for example:

```yml
resource:
- name: mygit
   type: git
   source:
       ...
       private_key: ((chaol-gitlab-private-key))

jobs:
- name: myjob
   plan:
   - task: mytask
      ...
      run:
          path: echo
          args:
          - Hello ((name))

    - task: othertask
       file: mygit/((path)/file.yml
```

Maybe you meant we don't evaluate fields other than `source` and `params`. But here I consider `file`, `var_files` etc. as params of `set_pipeline` step though I put it at top level per your suggestion.

> I don't think you need to do it for the `vars:` since any vars contained within will be recursively evaluated when resolving the pipeline config.
> 

For evaluating `vars` is, I initially thought for the self reset use case. Say, when I initially set the pipeline by `fly`, I run the command:

```bash
$ fly -t some sp -c pipeline.yml -p name -v foo=bar
```

Then self reset would still need to pass `foo`, like:

```yml

jobs:
- name: self-reset
  plan:
  - get: mygit
  - set-pipeline: self
     file: mygit/pipeline.yml
     vars:
        foo: ((foo))

- name: normal-job
   plan:
   - task:
      ...
       run: echo ((foo))
```

But, yes, you are right, in `noral-job`, `((foo))` has been interpolated during `fly sp`, the `set-pipeline` step even no need to have `vars` in this case. So I'll remove `vars` from evaluation.

I'll keep evaluation for `file` and `var_files`, which is required for pipeline template.

atc/exec/set_pipeline_step.go
evanchaoli: @vito I just changed the first argument of `BuildFactory.Create` from `JobConfig` to `Job`, so that we don't add extra parameter and we can figure out pipeline name from job.

evanchaoli:
@vito I just addressed all current review comments for backend. UI change is still ongoing.

evanchaoli:
Screen shot with the latest UI change:

set-pipeline2

atc/exec/set_pipeline_step.go
vito: There are two ways for vars to be expanded: static expansion done at `set-pipeline` time, which is non-semantic and happens client-side across the template, and dynamic expansion done at runtime, which is only done for specific attributes which support it, i.e. a resource's `source` and `params` (not general `params` - specifically `params` which deal with resources and `params` which are given to tasks as env vars).

The `vars` field doesn't need to be evaluated because any vars given at `set-pipeline` time will take precedence, and any vars being provided at runtime will just be recursively evaluated once they're pulled into the pipeline template (from `vars:`).

My only concern is that by evaluating all of the other fields this step is inconsistent with the other steps which don't. You don't need to explicitly implement evaluation if you're providing the vars statically at `set-pipeline` time, which it seems like would be the only use case for `file` and `var_files`.

For example, a task step won't evaluate any vars in its `file:` field. This example would only work if `path` is provided at `set-pipeline` time:

```yaml
plan:
- task: othertask
  file: mygit/((path))/file.yml
```

...but with this change, `file:` would be able to expand at runtime on the `set-pipeline` step, but there doesn't seem to be a strong reason for this behavior.

atc/scheduler/factory/build_factory.go
```diff
@@ -242,8 +248,22 @@ func (factory *buildFactory) constructUnhookedPlan(
 
 			VersionedResourceTypes: resourceTypes,
 		})
+
+	case planConfig.SetPipeline != "":
+		name := planConfig.SetPipeline
+		if name == "self" {
+			name = job.PipelineName()
```
pivotal-jwinters: Does the db.Job get threaded through everywhere just so we can do this?

atc/exec/set_pipeline_step.go
evanchaoli: @vito Maybe I over considered potential use case here. I agree to keep consistent behavior of `file`. Let me remove evaluation of `file` and `var_files` for now.

atc/scheduler/factory/build_factory.go
evanchaoli: @pivotal-jwinters Sorry for my English. I don't quite get your point. Can you please explain more? And what's your suggestion?

pivotal-jwinters: @evanchaoli sorry my fault, I should have been more clear. It just seems like there are a lot of changes that need to be made in order to pass the db.Job object all the way down to this low level. I just wanted to confirm that the db.Job object is only needed because of the pipeline name, and if it is, maybe there's a simpler way of getting that info here.

evanchaoli: @pivotal-jwinters Yeah, I wonder a simpler way too. Here the entry point is method `buildFactory.Create`. Obviously we cannot store a `db.Job` object in `buildFactory` because it is global and stateless. I initially thought to add `pipelineName` as an extra parameter to `buildFactory.Create`, but that would also need to thread the extra parameter through a lot of levels. So I decided to change the first parameter from `JobConfig` to `Job`, which at least keep count of parameters unchanged, and it make future change easier if some other info of job/pipeline is needed at any level.

pivotal-jwinters: @vito is the concept of `self` something we want to support? I always thought the purpose of this feature would be a gitops style pipeline configuration. I'm not sure I see the benefit of setting a pipeline as its executing itself. 

Maybe thats more a question for @evanchaoli. Whats your use case for setting a pipeline as it    executes?

evanchaoli: @pivotal-jwinters My original purpose of this PR is to support `set_pipeline: self`, we currently have no use case for setting another pipeline yet. Many of our users have experiences with GitLab CICD. One feature gap they complained to me is, with GitLab CICD, one check-in a YAML file to root of a repo, then a CICD pipeline setup/updated; but with Concourse, if a pipeline YAML has some change, the user has to run `fly set-pipeline` manually again. Many users store the pipeline YAML file in the same repo as product code, and they want the pipelines to be automatically updated when code repos are updated. This PR will fill up the gap. I have demoed my implementation to some of our users, and they loved it.

evanchaoli:
> Thanks for adding a testflight test! Are you waiting on [concourse/mock-resource#3](https://github.com/concourse/mock-resource/pull/3) before adding tests for the vars stuff? (I noticed nothing uses it yet.)

@vito Once the mock-resource PR is merged, I can add `var_files` and `vars` to the testflight test, that I think that is a  nice-to-have but a blocker, because `var_files` and `vars` have been tested in unit tests.

atc/exec/set_pipeline_step.go
```diff
@@ -0,0 +1,212 @@
...
+
+	if len(staticVarss) > 0 {
+		config, err = vars.NewTemplateResolver(config, staticVarss).Resolve(false, false)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	atcConfig := atc.Config{}
+	err = yaml.Unmarshal(config, &atcConfig)
```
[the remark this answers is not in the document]
evanchaoli: This is a beautiful refactor! Took it.

atc/api/config_test.go
```diff
@@ -908,30 +908,16 @@ jobs:
 						request.Body = gbytes.BufferWithBytes(remoraPayload)
 					})
 
-					It("returns 200", func() {
-						Expect(response.StatusCode).To(Equal(http.StatusOK))
+					It("returns 400", func() {
```
evanchaoli: @vito  Your `UnmarshalConfig` refactor seems a breaking change. Originally, unknown fields at top level are silently swallowed, but now error will be thrown to client. I think which should be fine, but still want to confirm with you.

evanchaoli:
@vito Front end test has been added by @Twiknight, so I have just removed "WIP". This PR has a lot of conflicts with #4600, once that one is merged, I will have to spend huge amount of efforts to rebase and resolve conflicts. (:

atc/config.go
```diff
@@ -22,6 +23,33 @@ type Config struct {
...
+	// a 'skeleton' of Config, specifying only the toplevel fields
+	type skeletonConfig struct {
+		Groups        interface{} `json:"groups,omitempty"`
+		Resources     interface{} `json:"resources,omitempty"`
+		ResourceTypes interface{} `json:"resource_types,omitempty"`
+		Jobs          interface{} `json:"jobs,omitempty"`
+	}
+
+	var stripped skeletonConfig
+	err := yaml.UnmarshalStrict(payload, &stripped)
```
vito: (replying to your other comment in this context instead)

Ah, you're right. It looks like `UnmarshalStrict` actually enforces disallowing unknown fields:

https://github.com/kubernetes-sigs/yaml/blob/4cd0c284b15f1735b8cc247df097d262b8903f9f/yaml.go#L42

I only kept this as `UnmarshalStrict` because the original code did that, but now that it's unmarshalling into a struct with expected fields it'll actually blow up on unknown toplevel keys.

I think we can just replace this with a `yaml.Unmarshal`, then:

```suggestion
	err := yaml.Unmarshal(payload, &stripped)
```

atc/config.go
```diff
@@ -22,6 +23,33 @@ type Config struct {
...
+
+	strippedPayload, err := yaml.Marshal(stripped)
+	if err != nil {
+		return err
+	}
+
+	return yaml.UnmarshalStrict(
+		strippedPayload,
+		&config,
+		yaml.DisallowUnknownFields,
```
vito: This can be removed since `UnmarshalStrict` already adds that option.

atc/plan.go
```diff
@@ -137,6 +138,13 @@ type TaskPlan struct {
 	VersionedResourceTypes VersionedResourceTypes `json:"resource_types,omitempty"`
 }
 
+type SetPipelinePlan struct {
+	Name     string                 `json:"name"`
+	File     string                 `json:"config"`
```
vito: typo?:

```suggestion
	File     string                 `json:"file"`
```

atc/scheduler/factory/build_factory.go
pivotal-jwinters: @evanchaoli everything you've mentioned here is exactly what we want to achieve with the [projects rfc](https://github.com/concourse/rfcs/pull/32) :), which will implement a full gitops workflow. This `set_pipeline` feature is only one step in that direction, and I feel like the self-referencing pipeline actually contradicts this model, because your pipeline needs to be manually set before you can update it. I would much rather see all this configuration be statically configured in one place so that we can avoid that manual step, and make it more portable across concourse installations.

We also do something similar to what your describing on [ci.concourse-ci.org](https://ci.concourse-ci.org/teams/main/pipelines/reconfigure-pipelines). The only difference is that you have a separate pipeline that configures a bunch of other pipelines.

evanchaoli: @pivotal-jwinters I think the concept of "project" is far more significant than `set_pipeline`. I feel `set_pipeline` can be considered as a standalone tool, and "project" uses it. Actually, a pipeline may use `set_pipeline` to set any other pipelines without projects.

A project bootstraps from a manual `set-project`; and a self-reset pipeline bootstraps from a manual `set-pipeline`. From this perspective, they are the same overhead. 

As far as I can see, a lot of our users have a single pipeline. Some pipelines are complex with multiple groups. The "project" allows users to split a big pipeline into multiple ones. But if or not users want to migrate existing working-fine pipelines to "project" is a problem.

So I don't think self-reset contradicts the "project" model unless Concourse would force all pipelines to reside in a "project". Users should have the flexibility to choose which model to go with, "project" or just pipeline.

pivotal-jwinters: @evanchaoli I don't mean to indicate that "projects" will be necessary in order to use the `set_pipeline` feature. I only bring it up because that RFC has a lot of info about how we want to think about configuration moving forward. 

I feel strongly that pipelines should not be responsible for updating themselves with magic keywords and dynamic configuration. Even though it's just the pipeline name, it greatly complicates the problem space and makes things much more difficult to reason about.

kcmannem: @evanchaoli I've been watching this PR from afar and I'm very much for a feature like this. I know a lot of concourse use cases are fulfilled by doing `fly set-pipeline` from a task. So it's great we have something built in to solve this problem.
However I agree with @pivotal-jwinters, I don't like the aspect of giving values in the KV map special meaning. It adds a flavor to the concourse yaml and pushes it in a direction yaml is not supposed to go.
`self` also introduces a recursive nature to pipelines which I'm not in favor of as Concourse already has a big learning curve and if users start depending on this, anyone new will feel a burden for learning this larger complexity.

evanchaoli: @pivotal-jwinters @kcmannem I'm from one side contributing this feature, and from the other side am thinking from a user's perspective. Say `self` is not supported, I'll still use the pipeline itself's name to do self reset, which should work fine same as using `self`. Later on my pipeline is renamed by some chance, then `set_pipeline` will create a new pipeline with the original name, which is definitely unexpected, it might take time to find root cause. At that time, I would again argue why `self` is not supported.

I commented on the [`set_pipeline` RFC for supporting `self`](https://github.com/concourse/rfcs/pull/31#issuecomment-546393485), and some user gave a thumb up, so I guess other users may also be interested in pipeline self reset.

> I feel strongly that pipelines should not be responsible for updating themselves with magic keywords and dynamic configuration. 

> self also introduces a recursive nature to pipelines which I'm not in favor of as Concourse already has a big learning curve and if users start depending on this, anyone new will feel a burden for learning this larger complexity.

I think in opposite. I planed to suggest "self-reset" as a best practice to our users. I always got questions from new users, "where should I store my pipeline yml files?", "is there a way to auto set pipeline when my pipeline has changes?", and so on ... Users usually is not satisfied with answers like "you can store pipeline yml file in anywhere". The best practice model I planned to suggest to our user is like: (to be honest, this is reason why I contribute this PR.)

* Store pipeline yml as well as task files in the product code repo.
* Every pipeline has a self-reset job in a separate group. (I'll have a piece of copy in my doc, so that user may just copies my code.)
* Debug pipeline on a non-production clusters (our platform has provided dev and staging clusters), manual `set-pipeline` during developing and debugging the pipeline.
* For a new pipeline, push yml to code repo, then do an initial `set-pipeline`.
* For existing pipeline, only push yml to code repo, then pipeline will auto reset itself.

As a product, I believe Concourse wants to hear voice from users, which is a way to attract more users to turn to Concourse, and make Concourse better.

vito: Hmm. So, I personally think `set_pipeline: self` is pretty easy to reason about and doesn't really seem out-of-place. It's a neat way of letting a pipeline self-update without its template having to contain its own name - an intentional decision that Concourse already made to support re-usable pipeline templates .

In a world where we don't already have Projects planned, I could totally see it being our first iteration before we started addressing the other problems that Projects solves (e.g. centralizing the full CI stack, keeping tasks in sync with the pipeline config, splitting up pipeline configs by sharing resources).

There is precedent in interpreting values a specific way (i.e. `version: every`, `version: latest`, `version: {"foo":"bar"}`) so I don't find it particularly strange to have another keyword. It kind of feels like a natural thing to wonder if such functionality exists once you know `set_pipeline:` is a thing: how do I make my pipeline update itself? And if your solution requires you to put a pipeline name in the template (or pass it in as a var), that would contrast with the original design of pipeline templates (which is to not contain a name).

In a future when Projects is implemented, I do think it may be a little redundant; Projects and `set_pipeline: self` are definitely overlapping in functionality. I think it might be confusing to have two competing approaches to the same problem, especially when one addresses more problems than the other. For that reason I'm kind of against having *both*, at least in principle.

In reality, `set_pipeline` is being implemented first, and Projects is an unknown amount of time away. So I could see the argument for implementing `set_pipeline: self` first and letting it become deprecated and (very) eventually removed once Projects lands. I think the only danger here would be if `set_pipeline: self` usage spreads so far that doing so would be too disruptive.

evanchaoli: >  I think the only danger here would be if `set_pipeline: self` usage spreads so far that doing so would be too disruptive.

I don't see this is a problem. I guess Concourse will not force all pipelines to reside in some "project". Because I'm afraid existing pipeline owners won't have motivations to migrate to "project" before they hit scale problem. So what we can do is, keep `set_pipeline: self` for standalone pipelines, and disable `set_pipeline: self` in "Projects". In future, if Concourse decides to deprecate standalone pipelines, then `set_pipeline: self` is naturally deprecated at the same time.

pivotal-jwinters: @evanchaoli @vito implementing a feature with a plan to deprecate it down the road doesn't seem like a very sustainable approach to me. And TBH it feels like we've been doing that lot lately. The problem with that approach is that it makes maintaining the code and making changes towards our actual roadmap much more difficult. Especially when all we're doing is hazarding a guess at how users might use this feature.

This feature is quite useful without the `self` reference. I propose we remove it, and if down the road if enough people ask for it, we can revisit implementing it in a way that better aligns with our roadmap.

vito: @pivotal-jwinters The problem with waiting until everything is done perfectly is that we withold useful features in the short-term for the sake of having everything perfect in the long-term. In the meantime, users who try Concourse may be disappointed by the lack of basic functionality, and may just use something else. Remember, we're not here to just strive for a perfect design, we're working on software for people to use, and those users exist *today*. It doesn't help anyone if Concourse v10 finally ships but by then everyone has adopted other solutions that addressed their needs more quickly.

My point was that this could have been a useful MVP before we even came up with the idea for 'projects'. By getting that MVP out more quickly we would have at least met the basic needs of users already and collected useful feedback/information about how it's used. Maybe we would have even come up with a different idea. Maybe we would have come up with 'projects' and ultimately deprecated `set_pipeline: self`. Who knows?

We're lucky to have a pretty far view of the future plans right now - but I don't think we should let that long-term vision cheat us and our users out of short-term wins if we can also identify a transition from the short-term to the long-term. We need to balance our ideal roadmap with the time it takes to get there, and be conscious of the fact that not many of those features in the roadmap have been put in front of actual users yet.

At this point this discussion is more attached to ideology than technical details - in this case the basic functionality is still there, and at the end of the day we're debating a feature that's easy to replace with simple templating. I just think it's been a mistake of ours to continuously chase "perfect" at the expense of "good". We're going to have to be able to contiuously iterate on, refactor, and deprecate Concourse functionality regardless. I think it's something to become *good at*, not something to *avoid*.

> The problem with that approach is that it makes maintaining the code and making changes towards our actual roadmap much more difficult.

I agree in general but honestly this is a pretty simple thing to reason about and implement. The code diff is super inflated by the test setup, and those tests were already known to be a pain in the butt to maintain and of questionable value.

> Especially when all we're doing is hazarding a guess at how users might use this feature.

It's not a guess, this is a user-requested feature that was made before this RFC was even known about. :sweat_smile: It was also a use case that has been discussed at length in #1200. If anything is guess right now, it's Projects.

To be honest, I don't really have a horse in this race. I don't think we should include it if no one's asking for it, but a user *has* asked for it, and I think it's justifiable.

I'm going to move this point under "open questions" so we can collect more feedback on it. Right now we seem to have 4 strongly opinionated people here and I'd rather wait until we can align on something before just making a snap decision one way or another. I would like to hear more from other people on this subject and just make sure everyone is aware of the trade-offs either way.

pivotal-jwinters: @vito I'm not suggesting we wait for perfect, I'm suggesting we make small incremental changes and see how people use them.

> I agree in general but honestly this is a pretty simple thing to reason about and implement.

I actually don't see it that way at all. We're taking what used to be static configuration and now saying that it can be updated at any point out from under us. That's breaking a very large boundary IMO.

Reading through #1200 it seems like the majority of people are asking for gitops. That seems like a much larger endorsement for `projects` than self-aware pipelines (at least to me anyway).

vito: > I'm not suggesting we wait for perfect, I'm suggesting we make small incremental changes and see how people use them.

Fair.

> I actually don't see it that way at all. We're taking what used to be static configuration and now saying that it can be updated at any point out from under us. That's breaking a very large boundary IMO.

I guess I just don't really see that boundary as being broken in a unique way by allowing `self`. Pipelines can always change at any point, whether it's user-initiated, initiated by a Project, initiated by a separate pipeline using `set_pipeline`, or initiated by the same pipeline, either by using `self` or by being given its own name. Concourse has had to support this since day 1 - it's why builds and their build plans are decoupled from pipelines.

evanchaoli: > @evanchaoli implementing a feature with a plan to deprecate it down the road doesn't seem like a very sustainable approach to me. 

@pivotal-jwinters I would predict that `self` will not be deprecated in a viewable long term. As I suggested in last comment, we can disable `self` in Projects, but keep `self` for standalone pipelines. Have anything been planed to deprecated standalone pipeline and enforce all pipelines to reside under projects?

> Especially when all we're doing is hazarding a guess at how users might use this feature.

Why do you consider "self" as a guessed feature? This PR is contributed by a REAL user. I have stated clearly that, I, AS A USER, need `self`. My motivation was only `set_pipeline: self` rather than `set_pipeline: other_pipeline`. I'm not randomly picking up some issue and contribute, I implement something by myself because our users need it, and we cannot see when that can be done from Concourse side. I represent a large group of users behind me.

vito: I've added this under 'Open Questions' here: https://github.com/vito/rfcs/blob/set-pipeline-step/031-set-pipeline-step/proposal.md#open-questions

evanchaoli:
> The UI changes are pretty uncontroversial. Purely in an effort to be helpful, I might suggest https://github.com/concourse/concourse/blob/master/web/public/images/ic-breadcrumb-pipeline.svg for the set-pipeline step icon - no pressure.

@pivotal-jamie-klassen If you look at https://github.com/concourse/concourse/pull/4708#issuecomment-552318466, I think we have changed to use that icon.

evanchaoli:
@vito @pivotal-jwinters @kcmannem Looks like it's hard to reach an agreement if or not "self" should be supported, while I believe we all agree that "set-pipeline" is very useful. To unblock this PR, I have removed "self". Please consider to include this PR in release 6.0.0. Once this PR is merged, I'll create a new PR to add "self" back, then you guys may take time to decide if or not accept that PR.

PR #4712, opened 2019-11-05
allow for role-based config in cf-auth

### LINKED ISSUE #4706: Fine-grained CF space groups for CF dex connector
* Add `org:space:auditor` and `org:space:manager` and `org:space:developer`.
* For backwards-compatibility, we'll have `org:space` mean `org:space:developer`. Or migrate it, if possible.

xtremerui:
# Existing Issue
Fixes #4706 .

# Why do we need this PR?
With change in 5.7 that done in PR #4661 concourse grants access for all roles `developer\auditor\manager` of a space in cf connector versus before it only allows role `developer`. This breaking change might cause problem for user who only want to grant access to space developer.

Allow role-based config in cf auth connector.
 - Define team auth using `--cf-space-all` to grant access to all roles under that space.
 - Define team auth using `--cf-space-developer` `--cf-space-auditor` `--cf-space-manager`to grant access to individual role

when using `fly set-team --config`, exmaple file could be:
```
roles:
- name: member
  cf:
    space_developers: ["org:space"]
    space_auditors: ["org:space"]
    space_managers: ["org:space"]
...
```

# Contributor Checklist
- [x] Unit tests
- [x] Updated documentation (located at https://github.com/concourse/docs) https://github.com/concourse/docs/pull/274
[ ] Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [x] New config flags added? Ensure that they are added to the [BOSH](https://github.com/concourse/concourse-bosh-release) 


xtremerui:
New config flags added to BOSH releases in PR https://github.com/concourse/concourse-bosh-release/pull/63

pivotal-jwinters:
@vito short answer yes, the long answer is a sad tale.

When we decode the config file, we use the much maligned `mapstructure`. We've never had any golang tags associated with these fields, so all of our existing configuration kinda just mashes everything together. For example the cf connector also accepts the `--cf-space-guid` flag which when used in the config file, needs to be expressed as `spaceguids`. Most auth configs have one word flags like `users` and `groups`, which is probably why no one noticed. We were going to change them all, but that wouldn't be backwards compatible.  So do we want to be consistent, or correct in this one case?

pivotal-jwinters:
@vito at first I though the LDAP connector had a bunch of flags that were all mashed together, but it turns out those are all for the top level provider, and not team config (so not subject to the config file). So it looks like spaceguids was the only mangled config. 

So we used our brains, and figured out a solution that seems to work:

vito:
@pivotal-jwinters oh nice! yeah the explicit `mapstructure:` tags seem like the way to go. :+1: for the backwards-compatible field too!

skymarshal/skycmd/cf_flags.go
```diff
@@ -81,7 +87,20 @@ func (flag *CFTeamFlags) GetUsers() []string {
...
+	}
+	for _, space := range flag.SpaceDevelopers {
+		groups = append(groups, fmt.Sprintf("%s:developer", space))
+	}
+	for _, space := range flag.SpaceAuditors {
+		groups = append(groups, fmt.Sprintf("%s:auditor", space))
+	}
+	for _, space := range flag.SpaceManagers {
+		groups = append(groups, fmt.Sprintf("%s:manager", space))
+	}
```
pivotal-jamie-klassen: Something makes me uneasy about this method of serializing these groups. It looks like spaces can have a colon in their name (based on [this regex](https://github.com/cloudfoundry/cloud_controller_ng/blob/0fc14205ce043c1f22a398c67ad080326f9ad115/app/models/runtime/space.rb#L10) and the fact that Ruby's `[[:punct:]]` pattern matches colons). Is there anything bad that can happen here if my space name has a colon in it?

xtremerui: Nothing bad. If dex returns a group like `some-org:some-space-has-c::olon` [here](https://github.com/concourse/dex/blob/8bbc2a6df1eaa27aa67e54c7948002c709ed5e0e/connector/cf/cf.go#L292), as long as in the team auth config the same string is set by `--cf-sapce-developer some-org:some-space-has-c::olon`, everything should work. We don't have restriction on team auth configuration.

vito:
Woops I actually had a review in progress for this which I forgot to submit... It might slow things down but maybe it'd be better to just request re-review of any reviewers with changes requested?

Here's what I was gonna suggest: I find the config language a bit confusing. For example:

```yaml
- name: member
  cf:
    space_developers: ["org:space"]
    space_auditors: ["org:space"]
    space_managers: ["org:space"]
    spaces_all: ["org:space"]
```

The weird thing to me is that the fields are named `space_foos` but the value that they contain aren't `foo`s, they're spaces. And `spaces_all` also doesn't really make sense to me but I get why it's there lol.

Would something like this be doable?:

```yaml
- name: member
  cf:
    space_roles: ["org:space:developer", "org:space:auditor", "org:space:manager"]
```

I'm not sure about `spaces_all` tbh, it's hard to come up with a good name for it and at the end of the day it's just shorthand for explicitly listing the roles, of which there are only 3 anyway. IMO defaulting to 'developer' was sensible.

pivotal-jwinters:
@vito oh woops sorry.

pivotal-jwinters:
What about something like this?

```
- name: member
  cf:
    spaces_developer_role: ["org:space"]
    spaces_auditor_role: ["org:space"]
    spaces_manager_role: ["org:space"]
    spaces_all_roles: ["org:space"]
```

vito:
@pivotal-jwinters Don't we kind of already give up control of group formatting by virtue of using `org:space` syntax? :thinking: Maybe there's value in just being explicit about it and documenting the group syntax (and in this case, available sub-group/roles) for each provider.

I'd be OK with something like your example, or this one which is a bit more verbose but seems super clear:

```yaml
- name: member
  cf:
    spaces_with_developer_role: ["org:space"]
    spaces_with_auditor_role: ["org:space"]
    spaces_with_manager_role: ["org:space"]
    spaces_with_any_role: ["org:space"]
```

(note: went with `any` instead of `all`)

vito:
I guess the weird thing about exposing it fully is then there's no point in even having the special yaml/flags?

PR #4752, opened 2019-11-12
add (incomplete) containerd garden backend alongside gdn

### LINKED ISSUE #4361: Implement our own bespoke `containerd` Garden backend
In the long run, we want replace Garden and just have our higher-level own worker API. In the short term, however, implementing this as a Garden backend will give us a shortcut to validating `containerd` with a fast feedback loop.

After this is done we can begin to transition away from Garden entirely. This will be recorded as a separate issue.

cirocosta:
### Existing Issue

fixes #4361

Although this ir faaar from a final full PR, the changes I propose here are
meant to verify a potential approach that we could take with regards to
implementing a containerd garden backend in a way that we can integrate to
`master` as fast as possible (i.e., pushing new code everyday, and build it up
one piece at a time).

My hypothesis here is that given that we already support different garden
backends (e.g., `houdini`), we could have "just" another one without having to
change any of the rest of the code, building this thing ("feature"-flagged)
without worries.

The PR itself includes:

- a separate runner (see `*workercmd.gardenRunner()`) that's activated by using
  `--garden-enable-containerd`
- a "containerd client" (`worker/backend/libcontainerd`) that's meant to be the
  interface that we interact with when communicating with `containerd` through
  its client library
  - not only it can protect us from minutia changed in the underlying lib, but
    it can also provide us with ways of mocking and verifying what we're doing
    (going TDD for the implementation of our backend)
- a very imcomplete backend that at least satisfies the `garden.Backend`
  interface
  - at the moment, only `Ping()` is implemented.

To test it out, modify `docker-compose.yml` to have `CONCOURSE_GARDEN_USE_CONTAINERD='true'`, or run `concourse worker` on linux with `--garden-use-containerd`, having `containerd` and the other necessary bits under `/usr/local/concourse/bin` (or specify `--garden-bin`).

thanks!


go.mod
```diff
@@ -218,3 +227,5 @@ require (
 )
 
 go 1.13
+
+replace github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
```
cirocosta: yeah, this looks weird, but for a good reason: [containerd: go mod friendly](https://github.com/containerd/containerd/issues/3031)

> Containerd currently causes issues for downstream projects using go mod. Since this is the official Go dependency management tool now, support would be nice.

worker/backend/container.go
```diff
@@ -0,0 +1,139 @@
...
+// If a host port is not given, a port will be acquired from the server's port
+// pool.
+//
+// If a container port is not given, the port will be the same as the
+// host port.
+//
+// The resulting host and container ports are returned in that order.
+//
+// Errors:
+// * When no port can be acquired from the server's port pool.
```
cirocosta: if you're wondering where these comments came from: they're straight from `garden` - we can totally remove them, but it seemed helpful for me to have them while implementing the methods. totally fine w/ removing though

Dockerfile
```diff
@@ -3,12 +3,30 @@
...
+
+RUN apt update && apt install -y curl
+
+RUN set -x && \
+	curl -sSL https://github.com/containerd/containerd/releases/download/v$CONTAINERD_VERSION/containerd-$CONTAINERD_VERSION.linux-amd64.tar.gz \
+		| tar -zvxf - -C /usr/local/concourse/bin --strip-components=1 && \
+	curl -sSL https://github.com/opencontainers/runc/releases/download/$RUNC_VERSION/runc.amd64 \ 
+		-o /usr/local/concourse/bin/runc && chmod +x /usr/local/concourse/bin/runc && \
+	curl -sSL https://github.com/containernetworking/plugins/releases/download/$CNI_VERSION/cni-plugins-linux-amd64-$CNI_VERSION.tgz \
+		| tar -zvxf - -C /usr/local/concourse/bin
```
cirocosta: this can be removed in a subsequent PR after getting  https://github.com/concourse/ci/pull/204 properly figured out - having this in this PR makes it "complete" though (can be tested with `CONCOURSE_GARDEN_USE_CONTAINERD=true`) from macos/etc

Issue #2802, opened 2018-02-08
Cannot intercept into resource


pontusarfwedson:
I have an issue that I cannot run `fly intercept` on one of my resources. It is the resource defined on
https://github.com/pontusarfwedson/concourse-bitbucket-pullrequest-resource so can be defined as stated in the README. What happens on when trying to intercept my pipeline is that all my other resources will get me into `bash-4.4` inside the container. However, when intercepting this resource it closes for some reason so I quickly end up back in my original terminal context. **The resource is working as expected in my pipeline.**

I have tried setting `--verbose` to my intercept command. This one shows similar output for both my working resources and this non-working one. How would one debug this?

**VERBOSE OUTPUT FROM INTERCEPTABLE RESOURCE**
```
2018/02/08 15:35:24 GET /api/v1/info HTTP/1.1
Host: -
User-Agent: Go-http-client/1.1
Accept-Encoding: gzip


2018/02/08 15:35:24 HTTP/1.1 200 OK
Content-Length: 43
Connection: keep-alive
Content-Type: text/plain; charset=utf-8
Date: Thu, 08 Feb 2018 14:35:24 GMT
Server: nginx/1.13.7
Strict-Transport-Security: max-age=15724800; includeSubDomains;
Vary: Accept-Encoding
X-Concourse-Version: 3.8.0
X-Content-Type-Options: nosniff
X-Download-Options: noopen
X-Xss-Protection: 1; mode=block

{"version":"3.8.0","worker_version":"1.2"}

2018/02/08 15:35:24 GET /api/v1/containers?pipeline_name=notifypipelinewithpr&resource_name=notify-dev-branch&type=check HTTP/1.1
Host: -
User-Agent: Go-http-client/1.1
Accept-Encoding: gzip


2018/02/08 15:35:24 HTTP/1.1 200 OK
Content-Length: 98
Connection: keep-alive
Content-Type: application/json
Date: Thu, 08 Feb 2018 14:35:24 GMT
Server: nginx/1.13.7
Strict-Transport-Security: max-age=15724800; includeSubDomains;
Vary: Accept-Encoding
X-Concourse-Version: 3.8.0
X-Content-Type-Options: nosniff
X-Download-Options: noopen
X-Xss-Protection: 1; mode=block

[{"id":"02d7bbf9-d556-45b1-67b3-d20a6d41faf1","worker_name":"concourse-worker-1","type":"check"}]
``` 

[...]

Issue #1073, opened 2017-05-09
hijacking should search harder for a shell


chendrix:
It seems to be that in the absence of a command, hijacking works by starting /bin/bash in the victim container. Some containers don't have a /bin/bash - for example, the Git resource. In this case, such a hijacking attempt fails.

One can explicitly specify /bin/sh as a command to get a shell. It would be a minor boon if the hijacking process did this automatically if there was no /bin/bash.

---
Moved from concourse/fly#73
/cc @tomwhoiscontrary

DRuggeri:
+1
Similarly, those using a minimal docker image such as alpine don't have /bin/bash

emcniece:
At the very least, it would be nice to have some sort of visible error indicating that `fly hijack` can't find the executable. Right now, failing hijacks print a 200 response with the `--verbose` flag and exit cleanly when the attempt fails. We had to inspect ATC logs to discover `runc exec: exit status 1: exec failed... executable file not found in $PATH`.

wagdav:
@emcniece  this is documented in #833

DRuggeri:
Well, I still think it is relevant, Mr. bot...

tomwhoiscontrary:
It's no longer important to me, because i've stopped using Concourse.  Maybe it's still important to some other people here, though.

tomwhoiscontrary:
I feel like if a bug has been open for two years, and nothing's been done about it, the thing to do isn't to ~~exterminate~~ close it (oh, you cards!), it's to promote it to some sort of Hall of Fame.

tomwhoiscontrary:
The most sensible way i can see to do this would be to change `atc.HijackProcessSpec` so that `Path` is an array, whose elements are alternatives, in preference order. If there is an explicit command given to hijack, make it a single-element array. Otherwise, make it `["/bin/bash", "/bin/sh"]`.

Then, in `Server::hijack`, put the [`request.Container.Run`](https://github.com/concourse/concourse/blob/1d71ab91a3be61214250efdfa4ceda31fe53efef/atc/api/containerserver/hijack.go#L217), call in a loop, trying each alternative in turn, and stopping as soon as one succeeds. Better than that would be to only try the next alternative if the error is specifically that the command isn't found. I don't know what that error type is, though, so i don't know if it's possible to tell that.

This is not exactly a beautiful change, but it is fairly contained. One field and two bits of code need changing.

[...]

Issue #833, opened 2016-12-22
`fly hijack` should provide details regarding missing executable


coreyti:
# Feature Request

## What challenge are you facing?

When `fly hijack`ing, my task picked up the wrong/default Docker image (due to a mistake on my end). That image does not contain `bash`. This fact lead to the `fly hijack`  attempt bailing out, with no output.

Examining the ATC logs, I was able to determine that the issue was indeed that `bash` was missing. Simplified log output:

```json
{
  "source": "atc",
  "message": "atc.hijack.hijack.failed-to-hijack",
  "log_level": 2,
  "data": {
    "error": "hijack: Backend error: Exit status: 500, message: {\"Type\":\"\",\"Message\":\"runc exec: exit status 1: exec failed: container_linux.go:247: starting container process caused \\\"exec: \\\\\\\"bash\\\\\\\": executable file not found in $PATH\\\"\\n\",\"Handle\":\"\"}\n",
    "process": {
      "path": "bash",
      "args": null,
      "env": [
        "<redacted>"
      ],
    }
  }
}
```

## A Modest Proposal

```
$ fly -t target hijack -b 123
Sorry, you can't do that. `bash` executable file not found in $PATH on the container
```

jagregory:
Confirming same behaviour here. Silent fail from fly, ATC logging:

```
{
  "error":"hijack: Backend error: Exit status: 500, message: {\"Type\":\"\",\"Message\":\"runc exec: exit status 1: exec failed: container_linux.go:259: starting container process caused \\\"exec: \\\\\\\"bash\\\\\\\": executable file not found in $PATH\\\"\\n\",\"Handle\":\"\"}\n"
  ...
}
```

eruvanos:
Any update on this issue?

marco-m:
The real problem is that `fly` returns an exit status of zero (all OK) so this makes it error-prone or impossible to use in scripts

marco-m:
@jama-pivotal just to be sure you are aware: this is actually a bug, not a feature request:

fly intercept /non/existing/command

will silently fail and report 0 as exit status.

Ticket #758, which more explicitly talks about bug, has been closed as duplicated of this one.

srkiNZ84:
This is also affecting us and is a difficult issue to investigate as we don't immediately think to search the ATC logs.

cirocosta:
Hey,

[...]

PR #4755, opened 2019-11-13
fly: default to intercepting with sh instead of bash

agurney:
# Existing Issue

Related to #1073, but this PR does something different from what that issue suggests.

#2802 is an example in the wild of someone running into problems with a target that doesn't have bash.

#833 covers how the existing error message was made to be visible (vs. hidden in ATC logs)

# Why do we need this PR?

The `fly hijack` / `fly intercept` command will try to invoke bash unless told otherwise. Not all containers have bash - in particular, alpine-based ones are quite common, and typically don't. The fly error message we get in this case is not very friendly:
```
Backend error: Exit status: 500, message: {"Type":"ExecutableNotFoundError","Message":"exec failed: container_linux.go:345: starting container process caused \"exec: \\\"bash\\\": executable file not found in $PATH\"\n","Handle":"","ProcessID":"","Binary":""}
```

While it does identify the problem, in practice many users are quite thrown on seeing it. It comes across as "something mysterious is deeply broken with the platform".

# Changes proposed in this PR

Defaulting to `sh` instead of `bash` should be more portable in the first instance, while not presenting many difficulties for people doing the interception. If I'm just `cd`ing and `ls`ing around, it works in the same way. If I test-invoke a bash script, its shebang line should be enough to get it started.

Issue #1073 suggests a probing process for finding different shells on the target. This could be a bit gnarly, if we have to create a succession of different process specs and try them in turn. A hacky workaround might be to default to something like
```
sh -c 'which -s bash && bash || sh'
```
which seems like it might also be confusing when it breaks, for not much advantage.

# Contributor Checklist
- [ ] Unit tests
- [X] Integration tests (if applicable)

doc note: some examples in https://github.com/concourse/docs/blob/master/lit/docs/install/worker.lit demonstrate a bash shell, but the intercept default-shell behavior isn't otherwise mentioned

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/fly/commands/hijack.go
+++ b/fly/commands/hijack.go
@@ -31,7 +31,7 @@ type HijackCommand struct {
 	StepType       string                   `          long:"step-type"                         description:"Type of step to hijack (e.g. get, put, task)"`
 	Attempt        string                   `short:"a" long:"attempt" value-name:"N[,N,...]"    description:"Attempt number of step to hijack."`
 	PositionalArgs struct {
-		Command []string `positional-arg-name:"command" description:"The command to run in the container (default: bash)"`
+		Command []string `positional-arg-name:"command" description:"The command to run in the container (default: sh)"`
 	} `positional-args:"yes"`
 }
 
@@ -314,7 +314,7 @@ func remoteCommand(argv []string) (string, []string) {
 
 	switch len(argv) {
 	case 0:
-		path = "bash"
+		path = "sh"
 	case 1:
 		path = argv[0]
 	default:
--- a/fly/integration/hijack_test.go
+++ b/fly/integration/hijack_test.go
@@ -26,7 +26,7 @@ var _ = Describe("Hijacking", func() {
 		hijacked = nil
 		workingDirectory = ""
 		user = "root"
-		path = "bash"
+		path = "sh"
 		args = nil
 	})
 
@@ -738,7 +738,7 @@ var _ = Describe("Hijacking", func() {
 
 		Context("when called with a specific path and args", func() {
 			BeforeEach(func() {
-				path = "sh"
+				path = "bash"
 				args = []string{"echo hello"}
 
 				containerArguments = "build_id=2&step_name=some-step"
@@ -748,7 +748,7 @@ var _ = Describe("Hijacking", func() {
 			})
 
 			It("hijacks and runs the provided path with args", func() {
-				hijack("-b", "2", "-s", "some-step", "sh", "echo hello")
+				hijack("-b", "2", "-s", "some-step", "bash", "echo hello")
 			})
 		})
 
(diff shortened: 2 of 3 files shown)


vito:
Kinda torn on this. It's pretty convenient to be able to hop straight into `bash`, which should work with all resource types bundled with Concourse, and should work in most one-off tasks that people use.

I totally agree that the error message is unhelpful. Maybe we could improve the API by returning a better message, and maybe even have `fly` catch and handle the error by falling back on `sh`?

The quicker fix is definitely to just change the default to `sh`, but to be honest I don't think it's worth the broader inconvenience for folks that have `bash` installed and use `fly intercept` regularly.

If you're interested in taking that approach, I can provide some pointers.

Here is where the error is thrown and propagated to `fly`:

https://github.com/concourse/concourse/blob/881ef4f07f4dadfdc44124a08582900558a28244/atc/api/containerserver/hijack.go#L217-L237

Instead of just bubbling up a generic error message response, we could check if it's a `garden.ExecutableNotFoundError` and return a special payload that `fly` could handle. This could be done by adding a `ExecutableNotFound` boolean field to [`atc.HijackOutput`](https://github.com/concourse/concourse/blob/90a604c940b240a520b284e9a4ba5ad7062c4189/atc/hijack_payload.go#L30-L35) and checking for it in `fly` here:

https://github.com/concourse/concourse/blob/881ef4f07f4dadfdc44124a08582900558a28244/fly/commands/internal/hijacker/hijacker.go#L123-L132

When `fly` sees the field it could try again with `sh` instead. And if that still fails, it can come up with its own error message.

This approach would improve the overall UX for running an invalid executable while supporting both `bash` and `sh` transparently. It'll take a bit more work but I'm happy to help if you're interested!

agurney:
Thanks for the tips on the underlying flow - there are definitely a few moving parts here. I was also thinking about trying to choose Powershell on Windows targets, which would fit in with the probing model (perhaps accompanied by noticing the container's worker's `platform` - which is accessible in the API even if not immediately available).

I am still at a bit of a loss for what is important about finding bash. My intuition is that if somebody is just typing regular old interactive commands at the prompt of an ephemeral container, there isn't much difference between them (e.g., the things listed in https://www.gnu.org/software/bash/manual/html_node/Bash-POSIX-Mode.html don't seem too misleading). Do you have a sense of what people are doing when they intercept which is unique to bash?

vito:
@agurney `sh` doesn't support a lot of quality-of-life things like tab completion. A lot of the time people will be intercepting a container running their tests and might need to run a bunch of commands, so it'd just be kind of annoying to have to do the extra hop into `bash` every time.

agurney:
Aaaah, that's a good call! I didn't realize tab completion was a difference. Anyway, trying to figure out the probe-and-guess-again flow :-)

PR #4766, opened 2019-11-14
[release/5.7.x] Fix topgun for release pipeline

taylorsilva:
Same problem that #4764 tried to fix

I just moved where the issue occurred in the last commit. Needed logic
around what `suite` should be equal to.

### DIFF
--- a/topgun/common/common.go
+++ b/topgun/common/common.go
@@ -175,11 +175,16 @@ type BoshInstance struct {
 func StartDeploy(manifest string, args ...string) *gexec.Session {
 	WaitForDeploymentAndCompileLocks()
 
+	var modifiedSuiteName string
+	if suiteName != "" {
+		modifiedSuiteName = "-" + suiteName
+	}
+
 	return SpawnBosh(
 		append([]string{
 			"deploy", manifest,
 			"--vars-store", filepath.Join(tmp, DeploymentName+"-vars.yml"),
-			"-v", "suite='-" + suiteName + "'",
+			"-v", "suite='" + modifiedSuiteName + "'",
 			"-v", "deployment_name='" + DeploymentName + "'",
 			"-v", "concourse_release_version='" + concourseReleaseVersion + "'",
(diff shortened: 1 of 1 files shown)


topgun/common/common.go
```diff
@@ -175,11 +175,16 @@ type BoshInstance struct {
...
+	if suiteName != "" {
+		modifiedSuiteName = "-" + suiteName
+	}
+
 	return SpawnBosh(
 		append([]string{
 			"deploy", manifest,
 			"--vars-store", filepath.Join(tmp, DeploymentName+"-vars.yml"),
-			"-v", "suite='-" + suiteName + "'",
+			"-v", "suite='" + modifiedSuiteName + "'",
```
cirocosta: would bosh complain about having a variable set to `""` ?

cirocosta: oh nice, it doesn't 😁  

```bash
bosh int ./concourse.yml  -v "suite=''" > a
bosh int ./concourse.yml  -v "suite='foo'" > b
git diff --no-index ./a ./b
```

```diff
diff --git a/./a b/./b
index c1929d6d4..37c1293bb 100644
--- a/./a
+++ b/./b
@@ -28,7 +28,7 @@ instance_groups:
         host_key: ((tsa_host_key))
         log_level: debug
         token_signing_key: ((token_signing_key))
-    release: concourse
+    release: concoursefoo
   name: web
   networks:
   - name: test
@@ -63,7 +63,7 @@ instance_groups:
       log_level: debug
       worker_gateway:
         worker_key: ((worker_key))
-    release: concourse
+    release: concoursefoo
   name: worker
   networks:
   - name: test
@@ -71,7 +71,7 @@ instance_groups:
   vm_type: test
 name: ((deployment_name))
 releases:
-- name: concourse
+- name: concoursefoo
   version: ((concourse_release_version))
 - name: bpm
   version: ((bpm_release_version))
```

taylorsilva: Good question! That was an assumption of mine.

PR #4773, opened 2019-11-15
[release/5.7.x] Remove hyphens from ops-files for bosh topgun

taylorsilva:
Some tests in bosh topgun are still failing because they use these
ops-files which still have the hyphen.

I should have looked at the commit where `SUITE` was introduced to
ensure I covered everything. This commit catches everything that was
missed and I checked the commit where `SUITE` was introduced to 
ensure nothing else will come up because of this change.


taylorsilva:
I also did an `ag` for suite and this was all it caught 😁 thanks for approving!

PR #4735, opened 2019-11-08
Adopt the new concourse helm chart repo.

pivotal-bin-ju:
the chart directory of the new chart (https://github.com/concourse/concourse-chart.git) starts from the root directory, so the hardcoded `stable/concourse` is not needed anymore.

We do still rely on some charts from the `helm/charts` repo, so the `CHARTS_DIR` was renamed to `HELM_CHARTS_DIR`.

Related to https://github.com/concourse/ci/pull/200

# Existing Issue
The `topgun` testsuite hardcoded the chart directory `stable/concourse`. the new chart directory starts from root of the repo https://github.com/concourse/concourse-chart.git

Fixes https://github.com/pivotal/concourse-ops/issues/117

# Why do we need this PR?
It is mandatory if we migrate concourse helm chart to the new repo.

# Changes proposed in this pull request
remove the hardcoded chart path `stable/concourse`, since the new chart is hosted at the root directory.

# Contributor Checklist
- [x] Test `topgun` manually.

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [ ] New config flags added? Ensure that they are added to the [BOSH](https://github.com/concourse/concourse-bosh-release) 
      and [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for the [integration tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env) (for example, if they are Garden configs that are not displayed in the `--help` text). 

cc @taylorsilva

PR #4775, opened 2019-11-15
fix k8s topgun test charts path

xtremerui:
since we are now using concourse/concourse-chart.

We would need to have this fix for all our release branches otherwise the job there won't work.


taylorsilva:
This change doesn't actually work with changes we've made in CI (as far as my understanding goes): https://github.com/concourse/ci/pull/200

https://github.com/concourse/concourse/pull/4735 will align master with how CI is working. This is just FYI, to spread knowledge about this change.

PR #4776, opened 2019-11-15
Improve help text for `fly order-pipelines`

### LINKED ISSUE #4760: fly: improve help text for `fly order-pipelines`

The help text for `fly order-pipelines`:
```
Usage:
  fly [OPTIONS] order-pipelines [order-pipelines-OPTIONS]

Application Options:
  -t, --target=              Concourse target name
  -v, --version              Print the version of Fly and exit
      --verbose              Print API requests and responses
      --print-table-headers  Print table headers even for redirected output

Help Options:
  -h, --help                 Show this help message

[order-pipelines command options]
      -p, --pipeline=        Name of pipeline to order
```

describes the `-p` flag as `Name of pipeline to order`, which really makes it seem like you're only supposed to pass this flag once. In fact, if you look at the [docs](https://concourse-ci.org/managing-pipelines.html#fly-order-pipelines), you see it can be passed multiple times. It would be great if I could learn this from the help text, rather than needing to refer to the docs site.

Something as simple as
```
[order-pipelines command options]
      -p, --pipeline=        Name of pipeline to order - this flag can be passed multiple times
```
The copy could probably be better than that example, but you get the idea.

:+1:

iamjarvo:
Fixes #4760

```
Usage:
  fly [OPTIONS] order-pipelines [order-pipelines-OPTIONS]

Application Options:
  -t, --target=              Concourse target name
  -v, --version              Print the version of Fly and exit
      --verbose              Print API requests and responses
      --print-table-headers  Print table headers even for redirected output

Help Options:
  -h, --help                 Show this help message

[order-pipelines command options]
      -a, --alphabetical     Order all pipelines alphabetically
      -p, --pipeline=        List of pipelines to order. This flag can be specified 1 or multiple times. For example, --pipeline pipeline-1 --pipeline pipeline-2
```

### DIFF
--- a/fly/commands/ordering_pipeline.go
+++ b/fly/commands/ordering_pipeline.go
@@ -14,7 +14,7 @@ var ErrMissingPipelineName = errors.New("Need to specify atleast one pipeline na
 
 type OrderPipelinesCommand struct {
 	Alphabetical bool                       `short:"a"  long:"alphabetical" description:"Order all pipelines alphabetically"`
-	Pipelines    []flaghelpers.PipelineFlag `short:"p" long:"pipeline" description:"Name of pipeline to order"`
+	Pipelines    []flaghelpers.PipelineFlag `short:"p" long:"pipeline" description:"List of pipelines to order. This flag can be specified 1 or multiple times. For example, --pipeline pipeline-1 --pipeline pipeline-2"`
 }
 
 func (command *OrderPipelinesCommand) Validate() ([]string, error) {


matthewpereira:
I agree with this PR in spirit, but think that we might benefit from being a bit more specific. With the `order-pipelines` command, the `--pipeline` flag does indeed take a single pipeline name, not a list; the disambiguation we're aiming to clear up seems to be pointed toward the fact that this command can take in multiple `--pipeline` flags.

Could we instead say something like:

```
      -p, --pipeline=        Name of pipeline to order. This flag needs to be set multiple times. For example, --pipeline=pipeline-1 --pipeline=pipeline-2
```

---

Also, in regard to this help text from `fly -h`:
```
order-pipelines      Orders pipelines (aliases: op)
```

Do you folks think this is also a good opportunity to be more specific about this description?

Maybe we could say something like:

```
order-pipelines      Orders multiple pipelines (aliases: op)
```

This could probably be word-smithed a little more, WDYT?

iamjarvo:
I agree with both points. 

Taking from another tool that has flags that can be set multiple times.
What do you think of?
```
Name of pipeline to order. This flag can be set multiple times. For example, --pipeline pipeline-1 --pipeline pipeline-2
```

For the description, how about?

```
Sets the order of the selected pipelines in the given order
```

matthewpereira:
Thanks for the feedback!

Re: `This flag can be set multiple times.` vs `This flag needs to be set multiple times.` or `This flag must be set multiple times.`, my understanding is that ordering a single pipeline will inevitably lead to unpredictable sorting (it will set the sort order for the pipeline to 0, rather than a relative position). @pivotal-jamie-klassen can you confirm this?

It feels like we should encourage users to always sort multiple pipelines at once - if this just sets an order property in the database, my feeling is that we shouldn't even allow the user to perform it for a single pipeline at a time, but maybe there are other use cases I haven't learned about.

Stylistically, I *think* fly's help text always uses a `=` whenever a flag is meant to be set with a value. For example, if we look at the help output for `fly teams -h`:

```
Application Options:
  -t, --target=              Concourse target name
  -v, --version              Print the version of Fly and exit
...
```
the `--target=` flag is listed with an equals sign since it's meant to be used with a value, whereas `--version` can be used on its own. In order to more clearly communicate this `--FLAG=VALUE` relationship I think `--pipeline pipeline-1` should be `--pipeline=pipeline-1`. WDYT?

For the description, `Sets the order of the selected pipelines in the given order` probably shouldn't have two instances of 'order' communicating different things in the same sentence. Maybe something like:

`order-pipelines      Sets the order for a selection of pipelines (aliases: op)`

🤔

iamjarvo:
I can update the example to use `=` I copied the example from https://concourse-ci.org/managing-pipelines.html#fly-order-pipelines which didn't didn't use `=`

Agree on using `order` twice

I wonder if this flag should be updated to ignore when 1 pipeline is being set?  Maybe that would make the description and flag more understandable? I was trying to think of a case when a user would want to order 1 pipeline.

matthewpereira:
Thanks for being patient on this one @iamjarvo . Busy week! :|

It might be best to open a separate issue/RFC to tackle the functionality of the `order-pipelines` flag, since this one could be merged if we stick to the scope of the help text on its own.

If you don't mind updating your code as discussed in the last comments, I can ping someone else to give it a final look over and merge. 💪

YoussB:
Hey @iamjarvo,

Thanks for your patience until we came back to you. 

It will be amazing if you can address @matthewpereira comments above so we can merge this in as soon as possible.

YoussB:
Hey @iamjarvo 
I am going to go ahead and close this for inactivity, but please reopen it at your own discretion.
We'll be more than happy to continue reviewing and see it through.

Thanks!

Issue #4783, opened 2019-11-18
epic: implement Garden API for containerd


deniseyu:
As an incremental step towards using `containerd` natively for Concourse workers,
we're going to implent a Garden backend that's backed by `containerd`.

On the worker side, the backend will be shipped alongside `gdn`, with the switch
between one and the other possible through the use of a feature flag:
`--garden-use-containerd`.

- [x] Backend
  - [x] `Start()`: https://github.com/concourse/concourse/pull/4752
  - [x] `Stop()`: https://github.com/concourse/concourse/pull/4779
  - [x] `Ping()`: https://github.com/concourse/concourse/pull/4752
  - [x] `Capacity()`: noop
  - [x] `Create()`: https://github.com/concourse/concourse/pull/4779
  - [x] `Destroy(handle)`: 
    - https://github.com/concourse/concourse/pull/4784
    - https://github.com/concourse/concourse/pull/4881
  - [x] `Containers()`: https://github.com/concourse/concourse/pull/4779
  - [x] `BulkInfo()`: noop
  - [x] `BulkMetrics()`: noop
  - [x] `Lookup()`
    - https://github.com/concourse/concourse/pull/4784
    - https://github.com/concourse/concourse/pull/4881
<br />

- [ ] Container
  - [x] `Handle()`: #4913
  - [x] `Stop()`: #5114
  - [x] `Info()`: #5114
  - [x] `Run()`: #5114
  - [x] `Attach()`: #5114
  - [x] `Property`-related: #5114
  - [x] `Lookup`: #5114
  - [ ] `SetGraceTime()`:
  - no-ops
    - [x] `CurrentLimits()` (Bandwidth, CPU, Disk)?
    - [x] `Limit*()` (Bandwidth, CPU, Disk)?
    - [x] `StreamIn()`
    - [x] `StreamOut()`
    - [x] `Metrics()`

<br />

- [x] Process
  - [x] `ID()`: #5114
  - [x] `Wait()`: #5114
  - [x] `SetTTY()`: #5114
  - [x] `Signal()`: #5114 

<br />

References:

* [Garden Container interface](https://github.com/cloudfoundry/garden/blob/master/container.go)
* [Houdini implementation](https://github.com/vito/houdini/blob/master/container.go)
* [Alex's spike code from before](https://github.com/concourse/concourse/blob/containerd-impl/gcontainerd/container.go)

deniseyu:
[...]

PR #4784, opened 2019-11-18
worker: add Destroy functionality

cirocosta:
### Why do we need this PR?

As part of https://github.com/concourse/concourse/issues/4783, this PR adds the
functionality required to get garden's `Destroy` working with containerd.

- [x] backend's Destroy high-level implementation
- [x] Destroy integration
- [x] backend's Lookup high-level implementation 
- [ ] ~Lookup integration~
- [x] Stop task within running container before container.Delete
- [x] ~add Dockerfile to help integration tests run on OSX~ split out to separate story #4850

### Contributor Checklist

- [x] Unit tests

### Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] PR acceptance performed
- ~Documentation reviewed~
- ~Release notes reviewed~
- ~New config flags added?~


worker/backend/libcontainerd/client.go
```diff
@@ -51,6 +51,8 @@ type Client interface {
 	) (
 		containers []containerd.Container, err error,
 	)
+
+	Destroy(ctx context.Context, handle string) error
```
cirocosta: we should add a comment that describes what's this used for

worker/backend/backend.go
```diff
@@ -106,7 +112,15 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
 //
 // Errors:
 // * TODO.
-func (b *Backend) Destroy(handle string) (err error) { return }
+func (b *Backend) Destroy(handle string) (err error) {
+	var ctx = namespaces.WithNamespace(context.Background(), b.namespace)
+
+	if handle == "" {
+		return InputValidationError{}
```
cirocosta: it might be useful to have a string passed to this, so that we can add some more context to the error here 🤔  

e.g.:

```go
InputValidationError{"missing handle"}
```

so that we can not only check the error type, but also reason about the low-level detail that caused the validation to fail

worker/backend/backend.go
```diff
@@ -20,6 +20,12 @@ type Backend struct {
 	namespace string
 }
 
+type InputValidationError struct{}
```
cirocosta: I wonder if we should get these down to an `errors.go` file 🤔  seems a bit disconnected from the purpose `backend.go`

wdyt?

worker/backend/libcontainerd/client.go
```diff
@@ -107,3 +109,6 @@ func (c *client) Version(ctx context.Context) (err error) {
 	_, err = c.containerd.Version(ctx)
 	return
 }
+ func (c *client) Destroy(ctx context.Context, handle string) (err error) {
+ 	return
```
pivotal-bin-ju: Does this return work? We declared an error type to return, but this line will return nothing.

cirocosta: yep, it's a no-op - from the commit message:

> - still need to actually invoke containerd's Destroy and add integration
tests

the current PR has only the high-level logic w/out the actual functionality in place yet 😁

worker/backend/backend.go
kcmannem: support

worker/backend/backend.go
kcmannem: runtime package has a types.go file in the `new-put-get-step` branch. I'd move it if there are many more errors we plan on adding.

deniseyu: oh true - I forgot about the new `runtime` package. That would be a nice place to put error types, as a source of truth for all common abstractions in runtime world 😎  We can do this after the refactor is merged in.

deniseyu:
End of Tuesday notes:

* We learned that in order to destroy a container with a Task inside it, we first have to shut down the Task. This is something that Guardian did for us behind the scenes, but with containerd we will have to do it ourselves. To minimize surprise, it would be good to bring across the same behaviours (e.g. which termination signal is used, how long are the timeouts, etc). Here is an example of how a container must be gracefully shut down using containerd's API: https://containerd.io/docs/getting-started/

* The task shutdown API calls are asynchronous, because containerd daemon can take any length of time to actually remove the container. To handle this we'll need to `Wait` on an exit code or some other signal.

* There is a `Container.Stop` method in the Garden API which we will eventually want to connect up to the task shutdown logic that we're currently implementing inside the Destroy. We'll also want to refactor that out into its own method.

* For tomorrow: let's experiment and see if we can bring up a Container without any task defined. Containerd blocks deletion when there's a running task, but what if there's no task at all??

deniseyu:
Here's a useful script to destroy all containers across all containerd namespaces:

https://gist.github.com/deniseyu/55c573026a83e3e1b0011201bcf7b3fc

💥

deniseyu:
End of Thursday notes (we should probably rename our branch 😭)

* We forgot that before deleting a container, we have to call `Task.Delete`. I've added that to the `case <- ctx.Done()` branch of the `Task.Kill` logic.

* That broke integration tests, which should now be fixed. We are very close to having the entire lifecycle of a container implemented!

* I made a failed attempt to unit-test channels. HALP ciro 

Still need to do:

* Properly test the non-timeout branch of `Task.Kill`. The rest of the unit tests currently pass, but only because the context timeout is being triggered -- this isn't the normal path and should probably not be the default behavior for "happy path" tests.

* Think about parameterizing the context cancellation deadline. It's a pain to test a 10s timeout (unless anyone knows a Go equivalent for [timecop](https://github.com/travisjeffery/timecop)?) 
Ciro mentioned that perhaps the `namespace` can hold onto it - still needs some investigation.

* We may want to dig deeper into exit codes, and what non-zero exit codes might imply. But maybe we don't care and just trust that things are dandy if no error is returned from containerd.

* Consider abstracting the `Task.Stop`  + `Task.Kill` logic into a named method that can become our implementation of Garden's `Container.Stop` interface

worker/backend/backend_test.go
```diff
@@ -414,3 +412,14 @@ func TestSuite(t *testing.T) {
...
 	})
 }
+
+func exitBeforeTimeout(ctx context.Context) (<- chan containerd.ExitStatus, error) {
+	c := make(chan containerd.ExitStatus, 1)
+	go func() {
+		es := containerd.NewExitStatus(0, time.Now(), nil)
+		c <- *es // beat the timeout!! i'm a bad person for referencing and yes i feel bad
+		close(c)
+		return
```
deniseyu: might not need this `return`

deniseyu:
Going to split out `Lookup` into a different task, because writing that integration test would be a good opportunity to build out `Handle()`. We also set up a Dockerfile for running containerd integration tests which will be a separate PR.

worker/backend/backend.go
deniseyu: We'll make a note to do this for all the error types when we're a little further along!

worker/backend/libcontainerd/client.go
deniseyu: done!

deniseyu:
End of Thursday notes:

@pivotal-bin-ju and I finished building out the local integration test tooling and we're able to run and develop the tests in Docker now, woo. We thought about adding a test for Lookup, but decided that since Destroy implicitly calls Lookup, for now there's not much value in adding an integration test for it - this may change if we want to do more interesting things as part of Lookup.

Next week we'll work on filling out the rest of the Garden client and container and process interfaces! Maybe once we get the process stuff working (can refer to Alex's spike branch) we'll be able to see a basic task running in the UI.

worker/backend/backend.go
```diff
@@ -3,14 +3,17 @@ package backend
...
 	bespec "github.com/concourse/concourse/worker/backend/spec"
+	"github.com/containerd/containerd"
 	"github.com/containerd/containerd/cio"
+	"github.com/containerd/containerd/errdefs"
 	"github.com/containerd/containerd/namespaces"
-	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/opencontainers/runtime-spec/specs-go"
 )
 
 var _ garden.Backend = (*Backend)(nil)
```
kcmannem: what does this mean?

worker/backend/backend.go
```diff
@@ -106,7 +115,68 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
 //
 // Errors:
 // * TODO.
-func (b *Backend) Destroy(handle string) (err error) { return }
+func (b *Backend) Destroy(handle string) error {
+	var ctx = namespaces.WithNamespace(context.Background(), b.namespace)
```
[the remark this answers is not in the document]
deniseyu: it's a style thing, but true - we can change this to be consistent with the `:=` style!

worker/backend/backend.go
```diff
@@ -106,7 +115,68 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
...
+	const maxTaskKillWaitTime = 10 * time.Second
```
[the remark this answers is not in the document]
deniseyu: we did this today, and it'll be included as part of the next PR!

worker/backend/backend.go
deniseyu: In the most recent version of go, if a package ends with `-go`, go imports will automatically ignore the suffix and import it as the * in *-go. I think Goland does this as part of running gofmt

worker/backend/backend.go
```diff
@@ -106,7 +115,68 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
...
+
+	if handle == "" {
+		return InputValidationError{}
+	}
+
+	container, err := b.client.GetContainer(ctx, handle)
```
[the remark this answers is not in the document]
deniseyu: yea, that's true - we can add this in the next PR where the client timeout is configurable, we'll just decorate every context at the start of each method with the timeout.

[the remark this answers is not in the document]
deniseyu: added it as a to-do item here: https://github.com/concourse/concourse/issues/4783#issue-524571495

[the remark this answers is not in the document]
xtreme-sameer-vohra: Task runs don't have timeouts, but pretty much every other operation does :)

worker/backend/backend.go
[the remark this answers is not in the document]
xtreme-sameer-vohra: Should this be a context.TODO to signal that we intend to change it ?

worker/backend/backend.go
```diff
@@ -106,7 +115,68 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
...
+		return err
+	}
+
+	task, err := container.Task(ctx, nil)
+	if err != nil {
+		if !errdefs.IsNotFound(err) {
+			return err
+		}
+
+		return b.client.Destroy(ctx, handle)
```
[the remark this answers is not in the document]
xtreme-sameer-vohra: What are the other error types besides `IsNotFound` ? Are all of them safe to call a `Destroy` on the container ?

worker/backend/backend.go
```diff
@@ -106,14 +113,75 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
...
+
+// killTasks kills a task on time, gracefully if possible, ungracefully if not.
+//
+func killTasks(ctx context.Context, task containerd.Task) error {
+	exitStatus, err := task.Wait(ctx)
+	if err != nil {
+		return err
+	}
+
+	err = task.Kill(ctx, syscall.SIGTERM)
```
xtreme-sameer-vohra: Based on the docs, its not clear, is `task.Kill` async like `task.Wait` ?

worker/backend/backend.go
```diff
@@ -106,14 +113,75 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
...
+	if err != nil {
+		return err
+	}
+
+	select {
+	case <-exitStatus:
+		_, err = task.Delete(ctx) // todo: we're swallowing exitcodes in both these forks, do we care?
+		return err
+	case <-ctx.Done():
+		err = task.Kill(ctx, syscall.SIGKILL) // should return GRPC DeadlineExceeded error type, wrapped up
```
xtreme-sameer-vohra: This ctx has already expired. So `task.Kill` may or may not be a noop. Does it make more sense to have another ctx?

worker/backend/backend.go
```diff
@@ -106,14 +113,75 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
...
+	}
+
+	err = task.Kill(ctx, syscall.SIGTERM)
+	if err != nil {
+		return err
+	}
+
+	select {
+	case <-exitStatus:
+		_, err = task.Delete(ctx) // todo: we're swallowing exitcodes in both these forks, do we care?
```
xtreme-sameer-vohra: There is an edge case where the task ran to completion prior to being killed and that result gets swallowed currently.

worker/backend/backend.go
```diff
@@ -106,14 +113,75 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
 //
 // Errors:
 // * TODO.
-func (b *Backend) Destroy(handle string) (err error) { return }
+func (b *Backend) Destroy(handle string) error {
+	ctx := namespaces.WithNamespace(context.Background(), b.namespace)
```
xtreme-sameer-vohra: Does it make sense for this to be a context.TODO as we intend it to be replaced ?

worker/backend/backend.go
```diff
@@ -106,14 +113,75 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
...
+	return nil
+}
 
 // Containers lists all containers filtered by Properties (which are ANDed together).
 //
 // Errors:
 // * None.
 func (b *Backend) Containers(properties garden.Properties) (containers []garden.Container, err error) {
-	var ctx = namespaces.WithNamespace(context.Background(), b.namespace)
+	ctx := namespaces.WithNamespace(context.Background(), b.namespace)
```
xtreme-sameer-vohra: Does it make sense for this to be a context.TODO as we intend it to be replaced ?

worker/backend/backend.go
```diff
@@ -148,7 +216,20 @@ func (b *Backend) BulkMetrics(handles []string) (metrics map[string]garden.Conta
 //
 // Errors:
 // * Container not found.
-func (b *Backend) Lookup(handle string) (container garden.Container, err error) { return }
+func (b *Backend) Lookup(handle string) (garden.Container, error) {
+	ctx := namespaces.WithNamespace(context.Background(), b.namespace)
```
xtreme-sameer-vohra: Does it make sense for this to be a context.TODO as we intend it to be replaced ?

worker/backend/backend.go
```diff
@@ -148,7 +216,20 @@ func (b *Backend) BulkMetrics(handles []string) (metrics map[string]garden.Conta
...
+	if handle == "" {
+		return nil, InputValidationError{}
+	}
+
+	_, err := b.client.GetContainer(ctx, handle)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Container{}, nil
```
xtreme-sameer-vohra: Missed the context on this, why do we return an empty container in this case ?

worker/backend/backend.go
```diff
@@ -106,14 +113,75 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
...
+	}
+
+	timeDelimitedContext, _ := context.WithTimeout(ctx, maxTaskKillWaitTime)
+	err = killTasks(timeDelimitedContext, task)
+	if err != nil {
+		return err
+	}
+
+	return b.client.Destroy(ctx, handle)
+}
```
pivotal-bin-ju: what about to refactor the logic as below (move `killTasks` into else branch, to have less return route path):
```
	task, err := container.Task(ctxWithTimeout, nil)
	
	if err != nil { //error occurs
		if !errdefs.IsNotFound(err) { //a real error but the task can not be found
			return ClientError { InnerError: err }
		}
                // do nothing if we could not find the task
	} else {
               // kill the task
		err = killTasks(ctxWithTimeout, task)
		if err != nil {
			return ClientError{ InnerError: err }
		}
	}
        // no matter the task exists or is killed, the container should be destroyed anyway. 
	err = b.client.Destroy(ctxWithTimeout, handle)
	if err != nil {
		return ClientError{ InnerError: err }
	}
```

worker/backend/backend.go
```diff
@@ -106,14 +113,75 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
 //
 // Errors:
 // * TODO.
-func (b *Backend) Destroy(handle string) (err error) { return }
+func (b *Backend) Destroy(handle string) error {
+	ctx := namespaces.WithNamespace(context.Background(), b.namespace)
+	const maxTaskKillWaitTime = 10 * time.Second
+
+	if handle == "" {
+		return InputValidationError{}
```
pivotal-bin-ju: would it be better if the handle check is at the first line of the function.?

worker/backend/backend.go
```diff
@@ -106,14 +113,75 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
...
+		return err
+	}
+
+	err = task.Kill(ctx, syscall.SIGTERM)
+	if err != nil {
+		return err
+	}
+
+	select {
+	case <-exitStatus:
```
pivotal-bin-ju: Maybe we should not ignore the exitCode. From the comment of the source code:
```
// ExitStatus encapsulates a process's exit status.
// It is used by `Wait()` to return either a process exit code or an error
```
It may not return the error immediately after we `SIGTERM` the task.

worker/backend/backend.go
pivotal-bin-ju: We may need another wait after we `SIGKILL`. the `task.Kill` will return the status of GRPC call.

worker/backend/backend_test.go
```diff
@@ -174,6 +177,226 @@ func (s *BackendSuite) TestContainersConversion() {
...
+func (s *BackendSuite) TestLookupGetContainerFails() {
+	s.client.GetContainerReturns(nil, errors.New("err"))
+	_, err := s.backend.Lookup("non-existent-handle")
+	s.Error(err)
+	s.Equal(err, errors.New("err"))
+}
+
+func (s *BackendSuite) TestLookupGetContainer() {
+	s.client.GetContainerReturns(new(libcontainerdfakes.FakeContainer), nil)
+	container, err := s.backend.Lookup("non-existent-handle")
```
xtreme-sameer-vohra: `non-existent-handle` could be changed to some existing container

worker/backend/backend_test.go
```diff
@@ -174,6 +177,226 @@ func (s *BackendSuite) TestContainersConversion() {
...
+	s.Equal(1, fakeTask.KillCallCount())
+	ctx, signal, _ := fakeTask.KillArgsForCall(0)
+	s.Equal(syscall.SIGTERM, signal)
+
+	namespace, found := namespaces.Namespace(ctx)
+	s.True(found)
+	s.Equal(testNamespace, namespace)
+}
+
+func (s *BackendSuite) TestDestroyWaitError() {
```
xtreme-sameer-vohra: TestDestroyWaitError  -> TestDestroyTaskWaitError

worker/backend/backend_test.go
```diff
@@ -174,6 +177,226 @@ func (s *BackendSuite) TestContainersConversion() {
...
+	exitChannel := make(chan containerd.ExitStatus) // this never returns
+	fakeTask.WaitReturns(exitChannel, nil)
+	s.client.GetContainerReturns(fakeContainer, nil)
+
+	fakeTask.KillReturnsOnCall(0, nil)
+	fakeTask.KillReturnsOnCall(1, errors.New("kill-again-error"))
+
+	err := s.backend.Destroy("some-handle")
+
+	s.Equal(2, fakeTask.KillCallCount())
```
xtreme-sameer-vohra: For correctness, assert that syscall.SIGKILL was passed to `fakeTask.Kill`

worker/backend/libcontainerd/client.go
```diff
@@ -51,6 +52,21 @@ type Client interface {
...
+	// GetContainer retrieves a created container that matches the specified handle.
+	//
+	GetContainer(
+		ctx context.Context,
+		handle string,
+	) (
+		container containerd.Container, err error,
+	)
+
+	// Destroy stops any running tasks on a container and remove the container.
```
xtreme-sameer-vohra: nit: removes

worker/backend/backend.go
deniseyu: good catch! Lookup isn't finished being implemented yet, because the Garden interface expects a `garden.Container` as a return value, but we haven't built out `backend/container.go` yet which means that we don't have anything that looks like a Garden container but does containerd stuff under the hood yet. We realized that too late into the process of writing this method though, so we just left it as is, since we wrote the unit tests around error handling already. We'll make a note to revisit this method when our container mullet is finished.

worker/backend/backend.go
deniseyu: yes

worker/backend/backend.go
deniseyu: I am not 100% sure about this - will have to check with @cirocosta later but I _think_ this is actually the code we want

worker/backend/backend_test.go
deniseyu: yes this test is bad right now -- related to the incomplete implementation of `Lookup`. Will be sorted out later 😂

deniseyu:
thanks @xtreme-sameer-vohra for the inline comments! We're going to fold them into the top of this branch: https://github.com/concourse/concourse/pull/4881 because in the last 2 days we restructured the backend and I'm a coward for merge conflicts 🏋️‍♀️

worker/backend/backend.go
xtreme-sameer-vohra: Just saw [this](https://github.com/concourse/concourse/pull/4784#issuecomment-558836584) so this makes sense now :)

worker/backend/backend.go
xtreme-sameer-vohra: The `context.Background` is never cancellable and is generally used as the root context, some more notes [here](https://golang.org/pkg/context/#Background)
If we have a plan to shift these cases from context.Background or context.TODO to a context with a timeout or a cancellable context, I am happy :)

worker/backend/backend.go
xtreme-sameer-vohra: thx

worker/backend/backend.go
xtreme-sameer-vohra: Chatted with @deniseyu @pivotal-bin-ju and we are using context with a timeout everywhere now

PR #4791, opened 2019-11-19
Remove update last scheduled test

clarafu:
Update last scheduled is already tested behaviourally within the pipeline factory test for PipelinesToSchedule. https://github.com/concourse/concourse/blob/release/6.0.x/atc/db/pipeline_factory_test.go#L141


vito:
Gonna fold this into #4741 since that fixes the testflight build which is preventing this PR from going green, while this one prevents #4741 from going green, which prevents this one from going green, which prevents #4741 from going green.

PR #4808, opened 2019-11-21
bug: fixed a lidar bug that caused ATC crash.

### LINKED ISSUE #4807: ATC crashes due to a bug of Lidar
## Bug Report

I have upgraded  two clusters to 5.7.0 last Monday with Lidar enabled. Today we noticed that some ATCs crashed a few times.  Trace stack:

```
panic: value method github.com/concourse/concourse/vars.UndefinedVarsError.Error called using nil *UndefinedVarsError pointer
goroutine 65423631 [running]:
github.com/concourse/concourse/vars.(*UndefinedVarsError).Error(0x0, 0xc0079e54c0, 0x1f1e166)
	<autogenerated>:1 +0x71
github.com/concourse/concourse/atc/db.(*resource).SetCheckSetupError(0xc002e55a20, 0x2830840, 0x0, 0x4218a5, 0x1dcb260)
	/tmp/build/1c3187db/concourse/atc/db/resource.go:280 +0x130
github.com/concourse/concourse/atc/lidar.(*scanner).setCheckError(0xc000c5b7c0, 0x2895fa0, 0xc00079d920, 0x28aad00, 0xc002e55a20, 0x2830840, 0x0)
	/tmp/build/1c3187db/concourse/atc/lidar/scanner.go:119 +0x4c
github.com/concourse/concourse/atc/lidar.(*scanner).Run.func1(0xc0021d43a0, 0xc000c5b7c0, 0xc00686ed20, 0x28c4a00, 0xc002e55a20, 0xc0025ba000, 0xbff, 0xe00)
	/tmp/build/1c3187db/concourse/atc/lidar/scanner.go:64 +0x189
created by github.com/concourse/concourse/atc/lidar.(*scanner).Run
	/tmp/build/1c3187db/concourse/atc/lidar/scanner.go:60 +0x2d9
```

## Steps to Reproduce

Not sure, this bug should be caused by some race condition.

Basically, needs to enable lidar, and make some resource check fails.

## Expected Results

ATC should never crash.

## Actual Results

It crashed intermittently.

## Additional Context

> (delete this section if not needed)

## Version Info

* Concourse version: 5.7.0, but I believe this bug has been there since Lidar was introduced
* Infrastructure/IaaS: K8S
* Browser (if applicable): NA
* Did this used to work? No

evanchaoli:
# Existing Issue

Fixes #4807

# Why do we need this PR?

The bug causes ATC to intermittently crash.

Problem code is:

```go
	for _, resource := range resources {
		waitGroup.Add(1)

		go func(resource db.Resource, resourceTypes db.ResourceTypes) {
			defer waitGroup.Done()

			err = s.check(resource, resourceTypes) // The bug is here!!!
			s.setCheckError(s.logger, resource, err)

		}(resource, resourceTypes)
	}
```

All go routines share the caller function's variable `err`, so that there is a race condition against `err` over all go routines. The fix is simple, just change the go routine to use a local `err`, which will avoid race condition.

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
> fill out this section.
- [x] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/lidar/scanner.go
+++ b/atc/lidar/scanner.go
@@ -60,7 +60,7 @@ func (s *scanner) Run(ctx context.Context) error {
 		go func(resource db.Resource, resourceTypes db.ResourceTypes) {
 			defer waitGroup.Done()
 
-			err = s.check(resource, resourceTypes)
+			err := s.check(resource, resourceTypes)
 			s.setCheckError(s.logger, resource, err)
 
 		}(resource, resourceTypes)


atc/lidar/scanner.go
```diff
@@ -60,7 +60,7 @@ func (s *scanner) Run(ctx context.Context) error {
 		go func(resource db.Resource, resourceTypes db.ResourceTypes) {
 			defer waitGroup.Done()
 
-			err = s.check(resource, resourceTypes)
+			err := s.check(resource, resourceTypes)
 			s.setCheckError(s.logger, resource, err)
```
pivotal-jamie-klassen: so access to the `err` variable was not thread-safe, but is `setCheckError`? I guess so, because we assume the database will synchronize across concurrent queries, but I'm not totally sure.

evanchaoli: @pivotal-jamie-klassen The problem is, when multiple go routines operates the same object `err`, there is a race condition. For example, in go routine 1, `check` returns an error, then `setCheckError` is called. Right when `err.Error()` is called in routing 1, routine 2 may set `err` to nil, then in routine 1, `err.Error()` will crash.

pivotal-jamie-klassen: From what I can see you still have a data race, though, with multiple goroutines sharing unsynchronized access to the scanner `s` here. In the scenario you're describing, this race gets pushed down to the database, so the `check_error` that gets written in the database will depend on which goroutine does the last write, or maybe which one gets the lock on that row, depending on how Postgres synchronization works (I'm not really sure). What I'm wondering is, have you thought this through? Should we be careful about what ends up getting written in the database, or are you convinced that we have sufficient synchronization?

evanchaoli: @pivotal-jamie-klassen `s.setCheckError()` is stateless, it just call stores the error to database, so there should be no unsynchronized access problem for `s` here.

cirocosta: (a side thought here): if this is the kind of thing that `-race` could catch - it'd probably be in our favor to have either smoke tests or topgun ones (that are "end-to-end" in nature) running with a `-race` bug 🤔

evanchaoli: @cirocosta I agree. But for this kind of bug, it's really not easy to trigger, so I'm afraid a unit or topgun test case would be hard to catch this bug.

evanchaoli:
@pivotal-jamie-klassen @cirocosta BTW, do you guys have idea why all tests hang up today? I pushed commits to several PRs, no any test run so far.

xtremerui:
@evanchaoli we are in the middle of migrate old CI to new CI that runs in K8s. The PR pipeline is one that not fully migrated yet because @cirocosta is working on finding a way in k8s to have iso network for untrusted PR workers.

PR #4757, opened 2019-11-14
Don't redact single-char secret. mitigate #4749.

evanchaoli:
# Existing Issue

Fixes #4749.

# Changes proposed in this pull request

This is a quick change to only redact secrets whose lengths are longer than 1.

# Contributor Checklist
> Are the following items included as part of this PR? Please delete checkbox items that don't apply.
- [x] Unit tests
- [ ] Integration tests (if applicable)
- [ ] Updated documentation (located at https://github.com/concourse/docs)
- [ ] Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)

# Reviewer Checklist
> This section is intended for the core maintainers only, to track review progress. Please do not
> fill out this section.
- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [ ] New config flags added? Ensure that they are added to the [BOSH](https://github.com/concourse/concourse-bosh-release) 
      and [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for the [integration tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env) (for example, if they are Garden configs that are not displayed in the `--help` text).

Issue #4749, opened 2019-11-11
((redacted)) showing in build logs


esauser:
## Bug Report

`((redacted))` is showing up occasionally in the build logs when secrets aren't present.  It consistently happens with certain output, but we don't know _why_ it happens.

## Steps to Reproduce

One example of where it always happens is when output the result from `terraform plan`.

```
# vault_jwt_auth_backend.keycloak will be updated in-place
  ~ resource "vault_jwt_auth_backend" "keycloak" ((redacted))
```

There is no secret there.  As far as we can tell, it's just whitespace and/or newline.  But it certainly doesn't happen on every whitespace or newline.

## Expected Results

```
# vault_jwt_auth_backend.keycloak will be updated in-place
  ~ resource "vault_jwt_auth_backend" "keycloak"
```

## Actual Results

```
# vault_jwt_auth_backend.keycloak will be updated in-place
  ~ resource "vault_jwt_auth_backend" "keycloak"
```

## Additional Context

We have the `CONCOURSE_ENABLE_REDACT_SECRETS` set to `true`.

## Version Info

* Concourse version: 5.7.0
* Deployment type (BOSH/Docker/binary): Binary
* Infrastructure/IaaS: GKE web, GCP Computer VM workers
* Did this used to work? Redacting secrets is new in 5.7.0.  We didn't start using it in 5.6.0 due to https://github.com/concourse/concourse/issues/4656

evanchaoli:
@esauser Is there any invisible chars defined in your credential manager? We only redact non-empty secrets as the code:

```go
func (it *credVarsIterator) YieldCred(name, value string) {
	for _, lineValue := range strings.Split(value, "\n") {
		lineValue = strings.TrimSpace(lineValue) // trim space from a secret first
		if lineValue != "" { // then only redact non-empty secret
			it.line = strings.Replace(it.line, lineValue, "((redacted))", -1)
		}
	}
}
```

esauser:
@evanchaoli noticed another instance this morning.

```
* google_project_iam_member.velero_server_role_binding: Error applying IAM policy for project "MY_PROJECT_NAME": Error setting IAM policy for project "MY_PROJECT_NAME": googleapi: Error 400: The role name must be in the form "roles/((redacted))role((redacted))", "organizations/((redacted))organization_id((redacted))/roles/((redacted))role((redacted))", or "projects/((redacted))project_id((redacted))/roles/((redacted))role((redacted))"., badRequest
```
should read
```
[...]

PR #4817, opened 2019-11-25
fix(atc): when secret redaction is not enabled, should not buffer build log lines

### LINKED ISSUE #4799: stdout forced to buffer in build logs in v5.6+
## Bug Report

We got reports from Concourse users that their terminal-based download progress bars never displayed in the build output until the download was completely finished. Since they were regularly running 10 gigabyte downloads, this was pretty troublesome UX.

We did some digging, and it seems that stdout doesn't get flushed appropriately in the build logs as of v5.6.0.

## Steps to Reproduce

Run
```bash
$ fly -t target execute -c <(cat <<EOF
platform: linux
image_resource:
  type: registry-image
  source: {repository: ubuntu}
run:
  path: bash
  args:
  - -c
  - |
    for i in {1..10}; do
      printf "$i\r"
      sleep 0.5
    done
    echo
EOF
)
```

and curl the events endpoint for the resulting build, something like

```bash
$ fly -t target curl /api/v1/builds/<build id>/events
```

## Expected Results

One event per loop iteration, like

```bash
# ... initial events ...

id: 49
event: event
data: {"data":{"time":1574267342,"origin":{"id":"5dd569c8"},"config":{"platform":"linux","image":"","run":{"path":"bash","args":["-c","for i in {1..10}; do printf \"$i\\r\"; sleep 0.5; done\necho\n"],"dir":""},"inputs":null}},"event":"start-task","version":"5.0"}

id: 50
event: event
data: {"data":{"time":1574267342,"origin":{"id":"5dd569c8","source":"stdout"},"payload":"1\r"},"event":"log","version":"5.1"}

id: 51
event: event
data: {"data":{"time":1574267343,"origin":{"id":"5dd569c8","source":"stdout"},"payload":"2\r"},"event":"log","version":"5.1"}

id: 52
event: event
data: {"data":{"time":1574267343,"origin":{"id":"5dd569c8","source":"stdout"},"payload":"3\r"},"event":"log","version":"5.1"}

# and so on ...
```

## Actual Results

one big buffered event, like
```
id: 25
event: event
data: {"data":{"time":1574267470,"origin":{"id":"5dd56a41","source":"stdout"},"payload":"1\r2\r3\r4\r5\r6\r7\r8\r9\r10\r\r\n"},"event":"log","version":"5.1"}
```

## Additional Context

It looks like the `dbEventWriterWithSecretRedaction` is always used (https://github.com/concourse/concourse/blob/master/atc/engine/builder/delegate_factory.go#L378-L389), and it buffers all bits before a newline (https://github.com/concourse/concourse/blob/master/atc/engine/builder/delegate_factory.go#L510).

Seems like secret redaction might not actually be opt-in? We didn't get all the way to the bottom of this, but maybe @evanchaoli will be able to help, having last changed this code? In any case, we noticed that [`newDbEventWriter`](https://github.com/concourse/concourse/blob/master/atc/engine/builder/delegate_factory.go#L420-L426) is never used.

## Version Info

* Concourse version: v5.6.0
* Deployment type (BOSH/Docker/binary): docker-compose
* Infrastructure/IaaS: local
* Did this used to work? yes, tested and working with v5.4 and v5.5.6

evanchaoli:
# Existing Issue

Fixes #4799 .

When secret redaction is not enabled, use `newDBEventWriter` to create stdout and stderr.

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/engine/builder/delegate_factory.go
+++ b/atc/engine/builder/delegate_factory.go
@@ -2,7 +2,6 @@ package builder
 
 import (
 	"io"
-	"io/ioutil"
 	"strings"
 	"time"
 	"unicode/utf8"
@@ -316,8 +315,19 @@ func (d *checkDelegate) SaveVersions(versions []atc.Version) error {
 	return d.check.SaveVersions(versions)
 }
 
-func (*checkDelegate) Stdout() io.Writer                                 { return ioutil.Discard }
-func (*checkDelegate) Stderr() io.Writer                                 { return ioutil.Discard }
+type discardCloser struct {
+}
+
+func (d discardCloser) Write(p []byte) (int, error) {
+	return len(p), nil
+}
+
+func (d discardCloser) Close() error {
+	return nil
+}
+
+func (*checkDelegate) Stdout() io.Writer                                 { return discardCloser{} }
+func (*checkDelegate) Stderr() io.Writer                                 { return discardCloser{} }
 func (*checkDelegate) ImageVersionDetermined(db.UsedResourceCache) error { return nil }
 func (*checkDelegate) Errored(lager.Logger, string)                      { return }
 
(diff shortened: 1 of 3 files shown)


evanchaoli:
@vito This PR, together with #4757 and #4808, maybe worthy a minor release of 5.6.1.

deniseyu:
Have to admit I don't have deep familiarity with this part of the code (but I wanted to build some context 😄) -- from reading the code it looks like @zoetian's observation in https://github.com/concourse/concourse/issues/4799 that redaction is always opted in is indeed the case, and switching the flag off [results in the DummyCredVarsTracker being used](https://github.com/concourse/concourse/pull/4817/commits/17eb00662e3c79e49ba4de272d58e6a9db2f4725#diff-68ac07b71ffa26113681f2aec23dca07R23-L30). Hard-coding the `Enabled()` return value as `true` further reinforces the idea that it's either redact or use the dummy.

I didn't review the last PR that merged this logic in so I could just be missing context, but it seems odd that if an operator disables this flag, then we send them down the path of using a dummy struct. But maybe it's just a naming issue, that the dummy struct isn't actually a placeholder or no-op object in the same sense that we use dummies in other parts of the codebase; in which case, we should rename it away from "dummy" and to something like "noRedact".

evanchaoli:
@deniseyu Let me explain the story a little bit. I implemented the secret redaction feature, and added a CLI option to switch the feature. The way to turn off secret redaction is to return `dummyCredVarsTracker` from `NewCredVarsTracker` when the second parameter `on` is false. The theory here is, `dummyCredVarsTracker` is no-op, if we don't track secret, then it won't redact any secret. 

The regression is in `buildStepDelegate`. Because `buildStepDelete` gets only an object of `CredVarsTracker`, it doesn't know if or not secret redaction is on or off, thus I always created `dbEventWriterWithSecretRedaction` object for `Stdout()` and `Stderr()`, which always buffers line.

With this change, I added `Enabled()` to interface `CredVarsTracker`, so that `buildStepDelegator` may know if var tracker is on or off, when off, use the old `dbEventWriter`, which doesn't buffer line, so that avoid regression.

If you concern with the name of `dummyCredVarsTracker`, I can rename it to `noOpTracker`. But I don't quite understand what did you mean by `to something like "noRedact" (which might also let us eliminate quite a bit of code in vars_tracker.go)`, can you please explain?

evanchaoli:
@pivotal-jwinters This PR is target to fix the regression. I think the issue #4799 is saying that even if they don't turn on secret redaction, build logs are still buffered. 

I agree that ideally we should resolve the issue by handling "\r" properly. We have seen a couple of use cases, #4749 and #4799, that my design of secret redaction doesn't cover. So I'd wait for more issue reports, and then think about a solution to try to cover all use cases.

Issue #1107, opened 2017-05-09
ATC Should have a healthcheck endpoint attached to it


chendrix:
Would be nice to have a dedicated healthcheck endpoint on the ATC that we could setup so that HA ATC deployments can have an ATC removed from the pool when it becomes unhealthy.

---
Moved from concourse/atc#160
/cc @zachgersh

Freakin:
Health check endpoint should also not redirect back to external URL

jdub:
And ideally not create a session. 😄

mjj209:
I found that the `/api/v1/info` returns a 200 without a re-direct. If this was documented, maybe this issue could be closed? Works great for running an HA concourse on GCP.

jdub:
Ooh, nice. Documenting (and perhaps commenting) it to ensure it doesn’t
change would be ace.

zachgersh:
@mjj209 - sounds nice but I'd prefer it was actually a dedicated `/health`. That way, as mentioned above we don't run into a situation where a comment is needed to describe what other things `/info` is used for.

radoslav-tomov:
It would be really cool if concourse had a dedicated health endpoint, which can basically give enough information on whether the cluster can function.  For our use case at least, I would hook it to Availability service, which would call it some interval of time and send alerts/track availability.

Something like `/api/v1/health`  and should cover at least the health of the workers and the DB. Ideally, it should not disclose any information and should be exposed without authentication.



Right now, we have managed to do similar by other means, but it is ugly and would be much better if w

alexbakar:
Added PoC for the health endpoint here: https://github.com/concourse/concourse/pull/4818
Please test and comment. Any feedback is welcome.

taylorsilva:
There's an RFC discussing this now: https://github.com/concourse/rfcs/pull/46

PR #4818, opened 2019-11-25
experiment: Health endpoint

alexbakar:
# Existing Issue
Addresses #1107

# Why do we need this PR?
This PR is a PoC for a health endpoint. The health endpoint is an HTTP endpoint exposed outside the Concourse cluster similarly to the info endpoint(that could be found at api/v1/info). In some scenarios a basic knowledge for a Concourse cluster health is needed( for example status of the DB and the worker nodes). With this PR it is demonstrated how we can give some basic knowledge about the DB and workers status. The endpoint is meant to be public, i.e. it doesn't require authentication. That's why it is mandatory to not reveal any sensitive information about the cluster.

Added new HTTP endpoint that could be found on api/v1/health. Upon request the ATC will check in the DB for the health check URL of every registered worker. Then it will ask every worker using the mentioned URL for its health. The check is actually the existing health check implemented in concourse/worker/healthchecker.go that is pinging the Garden and listing the BaggageClaim volumes. 
The response is in JSON format giving information about the DB and workers statuses. For every worker the name is displayed along with its corresponding status, which can be either "OK" or "NOK". For example this is the result for a cluster with two worker nodes:
`
{"db_status":"OK","workers_status":{"45c38d7c5d3f":"OK","7b0ddfcc9a16":"NOK"}}
`

where the health checker succeeded for worker "45c38d7c5d3f" and failed for "7b0ddfcc9a16" (due to Garden issue in my case).

When the worker is registered the information about its health endpoint is put into the DB in the newly added "healthcheck_url" column in workers table. The process happens in TSA similarly to the registration of Baggageclaim URL and Garden address, including the creation of the reverse tunnel.

To test the PR after cloning the repo you can build it up with:
yarn install
yarn build
docker-compose up --build -d
curl http://localhost:8080/api/v1/health

Some questions:
- should we return the http status (and corresponding message) from the workers check? For example like <worker_X>: 200 OK. Is more complex format necessary?
- currently there is no status of the ATC returned (assuming that if an ATC responds it is healthy enough). Should we make the checks more complex and provide such information?
- no explicit DB check is performed (assuming that if listing the workers from DB is working, then DB is healthy enough). Should we execute more complex checks and what?
- should we execute the health check on all of the workers registered no matter what their state is or only on those in "running" state?
- should we defend against DOS attacks and how?
- currently a health check timeout (local) can be set for when the checks are performed on each worker. Should we have a (global) timeout for when the checks are performed on the server side?

Any feedback is welcome.

# Contributor Checklist
Currently no tests added, no documentation updated. The idea for this PR is to get as much feedback as possible to help clarify the necessity and the requirements for such feature.

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


vito:
(related: https://github.com/concourse/rfcs/pull/46)

xtremerui:
close due to inacitivity. Feel free to reopen.

Issue #3607, opened 2019-03-27
Use the 'artifacts' abstraction for all volumes


ddadlani:
As part of #3424 , in order to create ephemeral containers directly in Garden that are not tracked by the ATC, we need to have a way to create persistent volumes (for the Garden container images) that are not owned by any db containers.

Currently, this is not possible because of the following lines:
https://github.com/concourse/concourse/blob/cdf9656f00b5ea8c4054d60de1a96fa5242f8468/atc/worker/image/image.go#L136-L140

Creating an image requires a container ID for GC purposes. Creating a resource cache also requires a container as its `ContainerOwner`.

We had previously discussed extending the `Artifacts` abstraction to encompass all types of volumes. Then, the `Artifact` would deal with GC logic etc, making it easier to create and manage new 'types' of volumes. We already introduced `worker_artifacts` for https://github.com/concourse/concourse/pull/3307.

This spike is to investigate whether extending this `artifacts` abstraction makes it easier to create image volumes for #3424, and to flesh out what it would look like.

ddadlani:
For now our thoughts are revolving around keeping worker_artifacts without any managed states, and leaving the 'Creating' and 'Created' states in the volume. We can just leave around artifacts that are not attached to volumes until the 12 hr mark is completed.

Artifacts are owned by builds and containers so we can on delete cascade once the container gets deleted. The difference from the current implementation is that we do not require a creating container to create an artifact in the new flow, which means that artifacts and volumes can be created before db containers.

ddadlani:
Artifact lifecycle flow:

 - an artifact can exist without an underlying volume. The assumption is that a volume will eventually be created that points to the artifact.
 - an artifact can exist without being associated to anything. If it never gets associated, it will be deleted after 12h
 - an artifact can be owned by the following objects:
 	- build
 	- resource cache
 	- task cache
 	- base resource type
 	- resource cert
 - once an artifact is owned by something, it becomes initialized. Initialized volumes can be DELETED in the following cases:
 	- owned by build: 
 		- status: aborted
 		- status: failed
 			- if build is not latest build for job
 		- status: errored
 			- if build is not latest build for job
 		- status: succeeded
 		- end_time: not null
 	- owned by resource cache
 		- resource cache is deleted
 	- owned by task cache
 	    - task cache is deleted
 	- owned by base resource type
 		- never
 	- owned by resource certs
 		- never
 	- owned by nothing
 		- always

ddadlani:
Quick update:

[...]

PR #4825, opened 2019-11-25
Refactor exec/*_step.go to remove exposure to runtime/garden components

### LINKED ISSUE #4265: Refactor the Get Step and Fetcher
# Goal of this refactor
Move Runtime specific concerns in `Put` & `Get` Steps into a client implementation. This is in service of #4052 which is to have a an interface for steps or implementation of steps that is agnostic of a specific runtime.

There are 2 main areas of decoupling, `execution` of a script (garden) and `storage` of binary data (baggageclaim)

For `execution`, the worker selection, container creation and running was moved from the `Put` step into the `worker client`, which has runtime specific logic. 

For `storage`, references to `worker.Volume` were removed and we proposed introducing the concept of `Artifact` & `ArtifactSource`, with the following definitions;
`Artifact` contains runtime agnostic metadata used by Concourse to manage the life cycle of anything that needs to be stored/retrieved.
`ArtifactSource` contains an concrete implementation to be able to store and retrieve data (eg. worker.Volume) 
Details [Iteration 1](https://github.com/concourse/concourse/issues/4265#issuecomment-523619482) & [Iteration 2](https://github.com/concourse/concourse/issues/4265#issuecomment-524925507)

# Why it took long and was complex ?
The `Get` step was particularly tricky for 2 reasons;
1) it requires a worker for the `Find` logic of the Fetcher, even though we don't create a container in that flow
2) the ImageResourceFetcher also uses the same Fetcher flow, and it needs to be fetched on a particular worker.
More details [here](https://github.com/concourse/concourse/issues/4052#issuecomment-511574187)

kcmannem:
# Existing Issue

Fixes #4265 

# Why do we need this PR?
Removes runtime/garden dependencies from exec and resource package.

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


deniseyu:
I'm going to leave some comments inline but I think we should operate with a heavy bias towards merging rather than blocking on changes. Unless I flag inline comments with "NEEDS FIX", assume that it's feedback that shouldn't block merging.

atc/exec/get_step.go
```diff
@@ -168,68 +150,59 @@ func (step *GetStep) Run(ctx context.Context, state RunState) error {
...
-		workerSpec,
-		step.strategy,
-	)
-	if err != nil {
-		return err
-	}
+	containerOwner := db.NewBuildStepContainerOwner(step.metadata.BuildID, step.planID, step.metadata.TeamID)
 
+	// TODO: Starting should be emitted just before the get script is executed rather than at this point
 	step.delegate.Starting(logger)
```
kcmannem: revisit this TODO?

atc/exec/put_inputs.go
```diff
@@ -51,40 +51,36 @@ func NewSpecificInputs(inputs []string) PutInputs {
...
 		})
 	}
 
 	return inputs, nil
 }
 
-type putInputSource struct {
-	name   artifact.Name
-	source worker.ArtifactSource
+type putFooBarInput struct {
```
kcmannem: TODO: dont forget this FooBar

atc/exec/task_config_source_test.go
```diff
@@ -38,11 +40,11 @@ var _ = Describe("TaskConfigSource", func() {
 					"a":               "b",
 					"evaluated-value": "((task-variable-name))",
 				},
-				Params: &atc.Params{
+				Params: atc.Params{
```
kcmannem: Curious about this change. I think previously we would use a pointer value as we need to interpolate vars in (()) and then rewrite the value into the map.

atc/runtime/types.go
```diff
@@ -1,12 +1,87 @@
 package runtime
```
kcmannem: Over time many of the runtime abstractions will fall in here, with k8s or worker package as the implementors. Every other package should only consume contents from runtime/types

atc/atccmd/command.go
```diff
@@ -798,19 +800,17 @@ func (cmd *RunCommand) constructBackendMembers(
 	engine := cmd.constructEngine(
 		pool,
 		workerClient,
-		resourceFetcher,
+		resourceFactory,
 		dbResourceCacheFactory,
 		dbResourceConfigFactory,
 		secretManager,
 		defaultLimits,
 		buildContainerStrategy,
```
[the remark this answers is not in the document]
deniseyu: They *might*, depending on what we decide for Kubernetes runtime v1. It might mean that we keep the abstraction of multiple workers, but one worker has the Kubernetes API slotted in as a client. This would let operators experimentally provision one Kubernetes cluster as a tagged worker, rather than setting up a new foundation for Kubernetes.. we should do some whiteboarding and deeper thinking as a team about this!

atc/engine/builder/step_factory.go
```diff
@@ -89,7 +86,7 @@ func (factory *stepFactory) PutStep(
 		factory.resourceFactory,
 		factory.resourceConfigFactory,
 		factory.strategy,
-		factory.pool,
+		factory.client,
```
deniseyu: What's the context behind removing pool from PutStep only but not for the others?

atc/exec/artifact_input_step.go
```diff
@@ -65,13 +69,16 @@ func (step *ArtifactInputStep) Run(ctx context.Context, state RunState) error {
...
+
 	logger.Info("register-artifact-source", lager.Data{
 		"artifact_id": buildArtifact.ID(),
-		"handle":      workerVolume.Handle(),
+		"handle":      art.ID(),
 	})
 
-	source := NewTaskArtifactSource(workerVolume)
-	state.Artifacts().RegisterSource(artifact.Name(step.plan.ArtifactInput.Name), source)
+	state.ArtifactRepository().RegisterArtifact(build.ArtifactName(step.plan.ArtifactInput.Name), &art)
```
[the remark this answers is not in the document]
deniseyu: I think in the meantime we could make the intention of this abstraction clearer by giving ArtifactRepository some clear method interface names and overcompensating on documentation

atc/exec/artifact_input_step.go
```diff
@@ -56,7 +59,8 @@ func (step *ArtifactInputStep) Run(ctx context.Context, state RunState) error {
 		return ArtifactVolumeNotFoundError{buildArtifact.Name()}
 	}
 
-	workerVolume, found, err := step.workerClient.FindVolume(logger, volume.TeamID(), volume.Handle())
+	// TODO-Later artifact_input_step shouldn't be looking up the volume on the worker
```
deniseyu: do you mind expanding on the eventual plan for this?

atc/exec/task_step.go
```diff
@@ -185,7 +183,7 @@ func (step *TaskStep) Run(ctx context.Context, state RunState) error {
 		return err
 	}
 
-	processSpec := worker.TaskProcessSpec{
+	processSpec := runtime.ProcessSpec{
```
pivotal-bin-ju: Are we still in Core? Would it be better to have a `func NewProcessSpec()` instead of exposing the `struct` here?

atc/exec/task_step.go
```diff
@@ -291,8 +273,8 @@ func (step *TaskStep) imageSpec(logger lager.Logger, repository *artifact.Reposi
 	return imageSpec, nil
 }
 
-func (step *TaskStep) containerInputs(logger lager.Logger, repository *artifact.Repository, config atc.TaskConfig, metadata db.ContainerMetadata) ([]worker.InputSource, error) {
-	inputs := []worker.InputSource{}
+func (step *TaskStep) containerInputs(logger lager.Logger, repository *build.Repository, config atc.TaskConfig, metadata db.ContainerMetadata) ([]worker.FooBarInput, error) {
```
pivotal-bin-ju: Why we do have `FooBarInput` here?  Should we have a more meaningful name?

atc/exec/build/repository.go
```diff
@@ -1,71 +1,68 @@
...
-// A RegisterableSource	artifact is an ArtifactSource which can be added to the registry
-type RegisterableSource interface {
-	worker.ArtifactSource
+//go:generate counterfeiter . RegisterableArtifact
+// A RegisterableArtifact is an Artifact which can be added to the registry
+type RegisterableArtifact interface {
+	runtime.Artifact
 }
 
 // RegisterSource inserts an ArtifactSource into the map under the given
```
deniseyu: pls update this documentation snippet!

atc/exec/build/repository.go
```diff
@@ -1,71 +1,68 @@
...
-// Repository is the mapping from a Name to an ArtifactSource.
+// Repository is the mapping from a ArtifactName to an Artifact.
 // Steps will both populate this map with new artifacts (e.g.  the resource
 // fetched by a Get step), and look up required artifacts (e.g.  the inputs
 // configured for a Task step).
 //
 // There is only one ArtifactRepository for the duration of a build plan's
 // execution.
 //
-// ArtifactRepository is, itself, an ArtifactSource. As an ArtifactSource it acts
```
deniseyu: thanks for removing this paragraph.. it was heckin confusing

atc/exec/build/repository.go
```diff
@@ -1,71 +1,68 @@
...
-type Name string
+type ArtifactName string
 
-// Repository is the mapping from a Name to an ArtifactSource.
+// Repository is the mapping from a ArtifactName to an Artifact.
 // Steps will both populate this map with new artifacts (e.g.  the resource
 // fetched by a Get step), and look up required artifacts (e.g.  the inputs
 // configured for a Task step).
 //
 // There is only one ArtifactRepository for the duration of a build plan's
```
deniseyu: Is it still true that ArtifactRepository is an in-memory data structure that lasts only for one build lifetime? I wonder if there's a way to make this relationship between an artifact and a build crystal clear.

atc/exec/build/repository.go
```diff
@@ -1,71 +1,68 @@
...
-		repo: make(map[Name]worker.ArtifactSource),
+		repo: make(map[ArtifactName]runtime.Artifact),
 	}
 }
 
-//go:generate counterfeiter . RegisterableSource
-// A RegisterableSource	artifact is an ArtifactSource which can be added to the registry
-type RegisterableSource interface {
-	worker.ArtifactSource
+//go:generate counterfeiter . RegisterableArtifact
```
deniseyu: I'm guessing that we had to name this something other than "Artifact" to avoid token clash, but this implies that there exists a world in which other artifacts exist which cannot be registered. This name (and the historical naming) is more confusing than not. Could we use a different descriptor that reveals more about what this will eventually be used for? Like... "PotentialBuildArtifact", or at least something that links to some Concourse domain concept?

atc/exec/check_step.go
```diff
@@ -142,9 +144,12 @@ func (step *CheckStep) Run(ctx context.Context, state RunState) error {
 	deadline, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
-	checkable := step.resourceFactory.NewResourceForContainer(container)
-
-	versions, err := checkable.Check(deadline, source, step.plan.FromVersion)
+	//TODO: Check if we need to add anything else to this processSpec
+	processSpec := runtime.ProcessSpec{
+		Path: "/opt/resource/check",
```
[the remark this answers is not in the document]
deniseyu: the GetStep uses this same datatype but it's much more populated: https://github.com/concourse/concourse/pull/4825/files#diff-514296ff114a3c668658b1f40922d758R154-R157

It'd be useful to have some docs around the runtime.ProcessSpec concrete type declaration

deniseyu:
Just jotting down some questions that @pivotal-bin-ju and I have, to remember to ask y'all later!

* In `atc/plan.go`, there is a comment asserting that ArtifactInput and ArtifactOutput are used for `fly execute`. We can't confirm from briefly scanning the code whether it's the case that they're *only* used for fly execute. Unsure if this was called out as a comment as a concern we will need to address later, or just an interesting observation?

atc/runtime/types.go
```diff
@@ -1,12 +1,87 @@
...
+		logDest io.Writer,
+		recoverable bool,
+	) error
+}
+
+type ProcessSpec struct {
+	Path         string
+	Args         []string
+	Dir          string
+	User         string
```
deniseyu: It looks like in check, get, put, and task, the `User` field here never gets filled out. Is there a reason we need it?

atc/exec/task_step.go
kcmannem: yup, its our latest TODO. Will update the pr with changes for this.

atc/exec/task_step.go
kcmannem: core is allowed to consume any method or struct in the runtime package. Is there a benefit to exposing the structs through constructors or factories. I couldn't see any, anyone have opinions/ideas on this?

atc/engine/builder/step_factory.go
kcmannem: good catch

atc/db/worker_artifact.go
```diff
@@ -12,6 +12,7 @@ import (
 
 //go:generate counterfeiter . WorkerArtifact
 
+// TODO This should be deprecated
 type WorkerArtifact interface {
```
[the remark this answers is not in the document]
deniseyu: might be another story for the future, but how much effort would it be to bring `fly execute`'s code path more in line with the path that regularly scheduled steps follow?

atc/engine/builder/delegate_factory_test.go
```diff
@@ -47,12 +48,12 @@ var _ = Describe("DelegateFactory", func() {
 	Describe("GetDelegate", func() {
 		var (
 			delegate   exec.GetDelegate
-			info       exec.VersionInfo
+			info       runtime.VersionResult
```
deniseyu: VersionResult -- to me this means the data fields are populated by a Get or Check. is that accurate? we should also consider adding some comments to where this type is declared in runtime/types

atc/exec/artifact_output_step.go
```diff
@@ -44,24 +44,29 @@ func (step *ArtifactOutputStep) Run(ctx context.Context, state RunState) error {
 
 	outputName := step.plan.ArtifactOutput.Name
 
-	source, found := state.Artifacts().SourceFor(artifact.Name(outputName))
+	buildArtifact, found := state.ArtifactRepository().ArtifactFor(build.ArtifactName(outputName))
```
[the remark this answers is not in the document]
kcmannem: i wonder if the problem here is the fact that more than half the sentence is comprised of the word artifact 🧐 artifact(artifact().arifact(artifact))

atc/exec/task_step.go
deniseyu: Bin and I chatted a bit offline about this - this might be a Philosophical Question about how we use Go lol. we typically have constructors that return pointers to private structs in other places of the code, but I _believe_ that's to add more control around how packages interact. In the case here, where we're explicitly setting up this package just to hold types that are designed to be sharable, I think there is less need to have a constructor

atc/exec/artifact_output_step_test.go
```diff
@@ -59,11 +66,11 @@ var _ = Describe("ArtifactOutputStep", func() {
 		})
 	})
 
-	Context("when the source exists", func() {
+	Context("when the artifact exists", func() {
 		Context("when the source is not a worker.Volume", func() {
```
deniseyu: guessing we need to update these test descriptor strings: `source` -> `artifact`

atc/exec/artifact_output_step.go
```diff
@@ -44,24 +44,29 @@ func (step *ArtifactOutputStep) Run(ctx context.Context, state RunState) error {
...
 		return ArtifactNotFoundError{outputName}
 	}
 
-	artifact, err := volume.InitializeArtifact(outputName, step.build.ID())
+	dbWorkerArtifact, err := volume.InitializeArtifact(outputName, step.build.ID())
 	if err != nil {
 		return err
 	}
 
 	logger.Info("initialize-artifact-from-source", lager.Data{
```
deniseyu: Is this log output still accurate? Looks like the term `source` was purged from elsewhere in this file

atc/exec/build/repository_test.go
```diff
@@ -0,0 +1,77 @@
...
+			Describe("ArtifactFor", func() {
+				It("yields the first artifact by the given name", func() {
+					artifact, found := repo.ArtifactFor("first-artifact")
+					Expect(artifact).To(Equal(firstArtifact))
+					Expect(found).To(BeTrue())
+				})
+
+				It("yields the second artifact by the given name", func() {
+					artifact, found := repo.ArtifactFor("second-artifact")
+					Expect(artifact).To(Equal(firstArtifact))
```
deniseyu: are you sure this test passes? the assertion here is that retrieving by "second-artifact" returns the firstArtifact

atc/exec/build/repository_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	Context("when a artifact is registered", func() {
+		var firstArtifact *runtimefakes.FakeArtifact
+
+		BeforeEach(func() {
+			firstArtifact = new(runtimefakes.FakeArtifact)
+			repo.RegisterArtifact("first-artifact", firstArtifact)
+		})
+
+		Describe("ArtifactFor", func() {
```
deniseyu: This test file probably be shortened by just testing the "multiple artifacts" route, since having one artifact is implicitly checked by having two.

atc/worker/fetcher.go
```diff
@@ -0,0 +1,170 @@
...
+		imageFetcherSpec.ResourceTypes,
+		containerSpec,
+		processSpec,
+		containerMetadata,
+		imageFetcherSpec.Delegate,
+	)
+
+	ticker := f.clock.NewTicker(GetResourceLockInterval)
+	defer ticker.Stop()
+
```
pivotal-bin-ju: Can we move ticker and it's defer after the error check return (just before the for loop) 
```
106 if err == nil || err != ErrFailedToGetLock {
107		return result, volume, err
108	}
```

atc/exec/artifact_output_step.go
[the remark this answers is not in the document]
deniseyu: the word 'artifact' no longer has any meaning to me 🤔

atc/exec/get_step.go
```diff
@@ -67,48 +65,27 @@ func NewGetStep(
...
-// If the worker does not have a VolumeManager, or if the worker does have a
-// VolumeManager but a cache for the version of the resource is not present,
-// the specified version of the resource will be fetched. As long as running
-// the fetch script works, Run will return nil regardless of its exit status.
-//
-// If the worker has a VolumeManager but did not have the cache initially, the
-// fetched ArtifactSource is initialized, thus warming the worker's cache.
-//
-// At the end, the resulting ArtifactSource (either from using the cache or
-// fetching the resource) is registered under the step's SourceName.
```
deniseyu: We should make a note to come back here and add better description of what's happening, once this refactor is finished. Definitely good to remove it for now since incorrect documentation is worse than no documentation.

atc/exec/put_inputs.go
```diff
@@ -28,13 +28,13 @@ func NewAllInputs() PutInputs {
 	return &allInputs{}
 }
 
-func (i allInputs) FindAll(artifacts *artifact.Repository) ([]worker.InputSource, error) {
-	inputs := []worker.InputSource{}
+func (i allInputs) FindAll(artifacts *build.Repository) ([]worker.FooBarInput, error) {
```
deniseyu: does this need a caller? doesn't look like `i` is used anywhere within the function

atc/exec/put_inputs.go
```diff
@@ -28,13 +28,13 @@ func NewAllInputs() PutInputs {
...
+	inputs := []worker.FooBarInput{}
```
deniseyu:
```suggestion
	inputs := make([]worker.FooBarInput, 0)
```

atc/exec/put_inputs.go
```diff
@@ -51,40 +51,36 @@ func NewSpecificInputs(inputs []string) PutInputs {
...
 }
 
-type PutResourceSource struct {
-	worker.ArtifactSource
-}
-
-func (source PutResourceSource) StreamTo(ctx context.Context, logger lager.Logger, dest worker.ArtifactDestination) error {
-	return source.ArtifactSource.StreamTo(ctx, logger, dest)
-}
+//func (source PutResourceSource) StreamTo(ctx context.Context, logger lager.Logger, dest worker.ArtifactDestination) error {
```
deniseyu: is this commented-out bit still needed?

atc/exec/put_step.go
```diff
@@ -109,7 +110,7 @@ func (step *PutStep) Run(ctx context.Context, state RunState) error {
 		putInputs = NewSpecificInputs(step.plan.Inputs.Specified)
 	}
```
deniseyu: I know this code is not new, but we might have an opportunity here to refactor this confusing control flow - I just caught it because the implementation of specific inputs and inputs is changing. Maybe one for the refactoring club!! @pivotal-jamie-klassen

atc/exec/put_inputs.go
```diff
@@ -51,40 +51,36 @@ func NewSpecificInputs(inputs []string) PutInputs {
...
+//	return source.StreamableArtifactSource.StreamTo(ctx, logger, dest)
+//}
```
deniseyu: For funzies I decided to see if I could figure out what a `FooBarInput` was just from reading this code. It sounds like it's either the full set, or a subset, of registered artifacts that are currently registered in the ArtifactRepository object, which may be used as part of the output generation process for exporting some artifact to the outside world during a Put. If that's true, then maybe the concept here is like a `StepProcessingInputs`, or `StepContainerInputs`, or `ActiveInputs`.

Not sure if you have historical context on this but I'm wondering why there's added complexity to cut down the set of inputs, especially when we're prepared to default it to the whole set when it's not specified. Is it related to performance?

On the whole I think it was a good idea to pull this logic out into its own file and introduce some interfaces. The same can probably be done for task step inputs later!

atc/exec/task_config_source.go
```diff
@@ -219,10 +218,10 @@ func (configSource ValidatingConfigSource) Warnings() []string {
 	return configSource.ConfigSource.Warnings()
 }
 
-// UnknownArtifactSourceError is returned when the artifact.Name specified by the
+// UnknownArtifactSourceError is returned when the artifact.ArtifactName specified by the
```
deniseyu:
```suggestion
// UnknownArtifactSourceError is returned when the build.ArtifactName specified by the
```

atc/exec/task_config_source_test.go
```diff
@@ -212,19 +222,19 @@ run: {path: a/file}
...
 			})
 		})
 
 		Context("when the file's artifact source cannot be found in the repository", func() {
 			It("returns an UnknownArtifactSourceError", func() {
-				Expect(fetchErr).To(Equal(UnknownArtifactSourceError{SourceName: "some", ConfigPath: "some/build.yml"}))
+				Expect(fetchErr).To(Equal(UnknownArtifactSourceError{SourceName: build.ArtifactName(artifactName), ConfigPath: artifactName + "/build.yml"}))
 			})
 		})
 	})
```
deniseyu: It's not letting me put an inline comment further down for some reason, but I think that with the change from pointers to values nested in the `TaskConfigSource`, it's worth adding an extra assertion to the unit test literally at the bottom of this entire file to confirm that `fetchedConfig.ImageResource.Params` also changes to take the interpolated values.

atc/exec/task_config_source_test.go
deniseyu: I think the interpolation logic is stateless now, and it returns an entirely new `TaskConfig` object rather updating nested pointers. It looks like it might be possible to update the address on that top-level pointer `TaskConfig *ImageResource` but in `task_step` and in the test, it looks like interpolation is performed, and the new value stored as a new variable that gets passed into later function calls. I don't see any evidence of `TaskConfig.ImageResource` being traversed for its inner values to be changed anymore though!

atc/exec/check_step.go
```diff
@@ -40,8 +42,8 @@ func NewCheckStep(
 	planID atc.PlanID,
 	plan atc.CheckPlan,
 	metadata StepMetadata,
-	containerMetadata db.ContainerMetadata,
 	resourceFactory resource.ResourceFactory,
+	containerMetadata db.ContainerMetadata,
```
pivotal-bin-ju: Why do we change the parameter order here?

kcmannem:
@deniseyu 
> * In `atc/plan.go`, there is a comment asserting that ArtifactInput and ArtifactOutput are used for `fly execute`. We can't confirm from briefly scanning the code whether it's the case that they're _only_ used for fly execute. Unsure if this was called out as a comment as a concern we will need to address later, or just an interesting observation?

yeah artifact input and output steps are only used for fly execute. Now that I think about it, I wonder if we can use these as a way for passing inputs/outputs between steps. Rather than have every step implement this logic individually. 🤔 

> * What's the future of the `worker_client`? Is the intention to slim this down and aggressively move FindOrCreate logic out of core entirely.. it seems like we might have to introduce some sort of blocking mechanism in order for Core to say "here is my container and process spec, figure the rest out" to the runtime, because the handover currently happens synchronously, and in the middle of all the `Run` functions.

The goal of worker_client was to eventually become a runtime_client with garden/k8s/etc implementations. The client interface should be moved to the runtime package at somepoint. As we've been discussing, the easiest way to plug in different backends is via the steps, thus the client contains RunXStep() methods. If we make the interface more granular later on, this can evolve into the things you talk about.

atc/exec/put_inputs.go
kcmannem: me and sameer had an impromptu discussion about this.

Heres the other options we considered:
```
DesiredInputs
InputRequests
ConfiguredInputs
InputClaims
InputRef
InputReferences
ArtifactReferences
```

We decided to change the implementation where FooBarInputs would be a map[DestinationPath]runtime.Artifact instead of a []FooBarInputs{artifact, destinationPath}

But we still didn't decide a name for this map.

atc/exec/get_step.go
```diff
@@ -168,68 +150,59 @@ func (step *GetStep) Run(ctx context.Context, state RunState) error {
...
-		return err
-	}
+	containerOwner := db.NewBuildStepContainerOwner(step.metadata.BuildID, step.planID, step.metadata.TeamID)
 
+	// TODO: Starting should be emitted just before the get script is executed rather than at this point
 	step.delegate.Starting(logger)
 
-	versionedSource, err := step.resourceFetcher.Fetch(
+	getResult, err := step.workerClient.RunGetStep(
 		ctx,
```
pivotal-bin-ju: 1. Is `RunXXXStep` a good naming?
2. Should we include those `RunXXXStep` in the interface `Client` or have a dedicated interface `ClientStepRunner` or something?

xtreme-sameer-vohra:
Current plan is to target merging this in after 5.8 is release so it has the opportunity to bake

xtreme-sameer-vohra:
Cleaned up history of commits
- removed concourse binary
- combined wip commits

atc/exec/set_pipeline_step.go
```diff
@@ -146,13 +155,22 @@ func (step *SetPipelineStep) Succeeded() bool {
...
-	repo   *artifact.Repository
+	repo   *build.Repository
 	step   *SetPipelineStep
+	client worker.Client
 }
 
 // streamInBytes streams a file from other resource and returns a byte array.
-func (s setPipelineSource) streamInBytes(path string) ([]byte, error) {
-	stream, err := s.repo.StreamFile(s.ctx, s.logger, path)
+func (s setPipelineSource) fetchPipelineBytes(path string) ([]byte, error) {
```
pivotal-bin-ju: comments need to be updated

atc/exec/get_step.go
```diff
@@ -163,6 +139,11 @@ func (step *GetStep) Run(ctx context.Context, state RunState) error {
 		ResourceTypes: resourceTypes,
 	}
 
+	imageSpec := worker.ImageFetcherSpec{
+		ResourceTypes: resourceTypes,
+		Delegate:      step.delegate,
```
pivotal-jwinters: Does this need more than just `stdout` and `stderr`?

atc/exec/get_step.go
```diff
@@ -163,6 +139,11 @@ func (step *GetStep) Run(ctx context.Context, state RunState) error {
...
+	}
+
 	resourceCache, err := step.resourceCacheFactory.FindOrCreateResourceCache(
```
pivotal-jwinters: at this point what's stopping us from moving the `resource_cache` over to the runtime side?

cirocosta:
Hey @kcmannem @xtreme-sameer-vohra ,

thank you'all for the effort on this!

<br />

Could you please update the description with some more info about the refactor?

I think it'd be great if it could reflect:

- what the PR is trying to accomplish
- why that's something needed
- how that's accomplished
  - what were the most significant changes?
  - is there specific parts of this refactor that other folks in the team should
    take a look at?

I think these points would give us reviewing the PR a bigger sense of whether
it's really achieving what it should and where to focus our attention to. 

<br />

thanks!!

atc/exec/build/repository.go
```diff
@@ -1,71 +1,68 @@
...
+// A RegisterableArtifact is an Artifact which can be added to the registry
+type RegisterableArtifact interface {
+	runtime.Artifact
```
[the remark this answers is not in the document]
xtreme-sameer-vohra: This is true, with the exception of a TaskCacheArtifact which uses a unique identified based on task attributes.

atc/exec/check_step.go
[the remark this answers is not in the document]
xtreme-sameer-vohra: This definitely could use more iteration to make it easier to read/understand, however, the `check_step` is functional as it sends up getting the source & params from the signature when the resource invokes the runscript - https://github.com/concourse/concourse/blob/new-put-get-steps/atc/resource/resource_check.go#L24

atc/worker/image/image_resource_fetcher.go
```diff
@@ -247,8 +262,11 @@ func (i *imageResourceFetcher) ensureVersionOfType(
 		return err
 	}
 
-	checkResourceType := i.resourceFactory.NewResourceForContainer(resourceTypeContainer)
-	versions, err := checkResourceType.Check(context.TODO(), resourceType.Source, nil)
+	processSpec := runtime.ProcessSpec{
+		Path: "/opt/resource/check",
+	}
+	checkResourceType := i.resourceFactory.NewResource(resourceType.Source, nil, resourceType.Version)
+	versions, err := checkResourceType.Check(context.TODO(), processSpec, resourceTypeContainer)
```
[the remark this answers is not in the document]
xtreme-sameer-vohra: Valid to make it a real ctx.
However, this has been using ctx.TODO since prior to the refactor - https://github.com/concourse/concourse/blame/master/atc/worker/image/image_resource_fetcher.go#L309

atc/worker/image/image_resource_fetcher.go
```diff
@@ -305,8 +323,11 @@ func (i *imageResourceFetcher) getLatestVersion(
 		return nil, err
 	}
 
-	checkingResource := i.resourceFactory.NewResourceForContainer(imageContainer)
-	versions, err := checkingResource.Check(context.TODO(), i.imageResource.Source, nil)
+	processSpec := runtime.ProcessSpec{
+		Path: "/opt/resource/check",
+	}
+	checkingResource := i.resourceFactory.NewResource(i.imageResource.Source, nil, i.imageResource.Version)
+	versions, err := checkingResource.Check(context.TODO(), processSpec, imageContainer)
```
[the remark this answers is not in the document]
xtreme-sameer-vohra: Valid to make it a real ctx.
However, this has been using ctx.TODO since prior to the refactor - https://github.com/concourse/concourse/blame/master/atc/worker/image/image_resource_fetcher.go#L309

atc/worker/image/image_factory.go
```diff
@@ -82,9 +82,7 @@ func (f *imageFactory) GetImage(
 
 	if imageSpec.ImageResource != nil {
 		var version atc.Version
-		if imageSpec.ImageResource.Version != nil {
-			version = *imageSpec.ImageResource.Version
-		}
+		version = imageSpec.ImageResource.Version
```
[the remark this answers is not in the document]
xtreme-sameer-vohra: There are 2 calls to `f.imageResourceFetcherFactory.NewImageResourceFetcher` and version is obtained from either `ResourceType.Version` or `ImageResource.Version`. So, this is not addressable in its current form.

atc/exec/artifact_input_step.go
xtreme-sameer-vohra: #3607

atc/exec/task_config_source_test.go
xtreme-sameer-vohra: thanks for letting us know

xtreme-sameer-vohra:
> Hey @kcmannem @xtreme-sameer-vohra ,
> 
> thank you'all for the effort on this!
> 
> 
> Could you please update the description with some more info about the refactor?
> 
> I think it'd be great if it could reflect:
> 
> * what the PR is trying to accomplish
> * why that's something needed
> * how that's accomplished
>   
>   * what were the most significant changes?
>   * is there specific parts of this refactor that other folks in the team should
>     take a look at?
> 
> I think these points would give us reviewing the PR a bigger sense of whether
> it's really achieving what it should and where to focus our attention to.
> 
> 
> thanks!!

Hey @cirocosta 
Yep, that totally makes sense. We had added that context to https://github.com/concourse/concourse/issues/4265

One question that is not completely addressed in the referenced issue above ;
- is there specific parts of this refactor that other folks in the team should
take a look at?

We'd like feedback/comments on the architectural changes and general style of de-coupling Runtime and Core modules. The `runtime` package was introduced as a place to locate `interfaces` & `structs` that we would expect all runtimes to leverage and could eventually be used to define the set of interfaces between Core & Runtime.

atc/worker/worker.go
```diff
@@ -583,9 +605,11 @@ func (worker *gardenWorker) cloneRemoteVolumes(
 		}
 
 		g.Go(func() error {
-			err = nonLocalInput.desiredArtifact.StreamTo(groupCtx, logger.Session("stream-to", destData), inputVolume)
-			if err != nil {
-				return err
+			if streamable, ok := nonLocalInput.desiredArtifact.(StreamableArtifactSource); ok {
```
cirocosta: Hey, 

was it intended to ignore the case where `nonLocalInput.desiredArtifact`'s type value does not match `StreamableArtifactSource`? i.e., do we want to proceed if that's not the case?

Just wondering if that's an "impossible statement" that we should treat as a programming mistake.

thanks!

atc/worker/container.go
```diff
@@ -138,3 +146,110 @@ func (container *gardenWorkerContainer) initializeVolumes(
...
+		}
+
+		err := json.Unmarshal(stdout.Bytes(), output)
+		if err != nil {
+			return fmt.Errorf("%s\n\nwhen parsing resource response:\n\n%s", err, stdout.String())
+		}
+		return err
+
+	case <-ctx.Done():
+		container.Stop(false)
```
cirocosta: from the Garden interface, `container.Stop(kill)` returns an error. Are we ignoring it on purpose? (I guess so).

If so, I wonder if we should signal that, e.g.:

```
_ = container.Stop(false)
```

thanks!

atc/worker/fetcher.go
```diff
@@ -0,0 +1,170 @@
...
+			return GetResult{}, nil, ctx.Err()
+		}
+	}
+}
+
+func (f *fetcher) fetchUnderLock(
+	ctx context.Context,
+	logger lager.Logger,
+	source FetchSource,
+	stdout io.Writer,
```
cirocosta: both `stdout` and `cache` are not being used in the implementation - was that intended?

atc/worker/fetcher.go
```diff
@@ -0,0 +1,170 @@
...
+	gardenWorker Worker,
+	containerSpec ContainerSpec,
+	processSpec runtime.ProcessSpec,
+	resource resource.Resource,
+	owner db.ContainerOwner,
+	imageFetcherSpec ImageFetcherSpec,
+	cache db.UsedResourceCache,
+	lockName string,
+) (GetResult, Volume, error) {
+	result := GetResult{}
```
cirocosta: if I understood correctly, we don't really need this line - `result` is already being created at line 98

atc/worker/client.go
```diff
@@ -369,3 +493,148 @@ type processStatus struct {
...
+		return PutResult{Status: status, VersionResult: runtime.VersionResult{}, Err: nil}
+	}
+
+	eventDelegate.Starting(logger)
+
+	vr, err = resource.Put(ctx, spec, container)
+	if err != nil {
+		if failErr, ok := err.(runtime.ErrResourceScriptFailed); ok {
+			return PutResult{failErr.ExitStatus, runtime.VersionResult{}, failErr}
+		} else {
```
cirocosta: IIRC, we tend to drop the `else` in these cases where there's an early return on an `if`.

cirocosta:
Hey,

I just pushed 2 changes to ensure that we have our code closer to the formatting
that Go projects are used to:

1. source formatting

a good number of files where missing a run of `go fmt`. IIRC, GoLand does not
run `go fmt` by default (I bet you were using goland :eyes:), I'd suggest in
that case configuring a precommit hook to ensure that we're pushing these
changes with the formatting on point - it's super easy to forget to run `go
fmt`.

2. import declarations

having the dot imports separated from the rest of the regular makes spotting
these easier than if they're grouped together with the rest of the imports - as
they should be used as an exception, such separation makes figuring out where an
exported thing comes from (although we should restrict those dot imports to
`ginkgo` only for DSL purposes).

if you have any concerns about those changes, please let me know!

thanks!

cirocosta:
@xtreme-sameer-vohra @kcmannem the testflight failure is pretty much what https://github.com/concourse/concourse/pull/4967 fixed - would you mind rebasing?

thx!

atc/exec/build/repository.go
```diff
@@ -0,0 +1,104 @@
...
+	repo.repoL.RLock()
+	for name, artifact := range repo.repo {
+		result[name] = artifact
+	}
+	repo.repoL.RUnlock()
+
+	return result
+}
+
+// func (repo *Repository) StreamFile(ctx context.Context, logger lager.Logger, path string) (io.ReadCloser, error) {
```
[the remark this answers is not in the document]
kcmannem: nope just garbage

atc/worker/worker.go
kcmannem: its an impossible case, the only types going in there should be of Streamable type. Its something that is hopefully temporary and gets fixed.

cirocosta: should we `panic`?

https://github.com/concourse/concourse/blob/cfe2a7ae78655c2e36f7a3afddb71961e0899ab5/atc/plan_factory.go#L65

atc/worker/fetcher.go
kcmannem: thanks for the catch, just residue from early experimenting on the branch

kcmannem:
@cirocosta thanks for your review, appreciate you pushing the formatting changes and explaining whats happening.

atc/db/volume.go
```diff
@@ -140,7 +140,7 @@ func (volume *creatingVolume) Failed() (FailedVolume, error) {
 }
 
 //go:generate counterfeiter . CreatedVolume
-
+// TODO-Later Consider separating CORE & Runtime concerns by breaking this abstraction up.
```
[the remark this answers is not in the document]
kcmannem: missed this one, but this is more of an epic that's known. I'll remove the todo as we don't need a reminder in the code.

atc/engine/builder/delegate_factory.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/atc/event"
 	"github.com/concourse/concourse/atc/exec"
+	"github.com/concourse/concourse/atc/runtime"
```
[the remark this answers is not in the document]
kcmannem: Shouldn't touch any in worker other than client. Whose interface will be moved to runtime at some point.

atc/worker/fetcher.go
kcmannem: done

atc/exec/build/repository.go
xtreme-sameer-vohra: I dont quite follow :(

kcmannem: @xtreme-sameer-vohra RegristerArtifact not RegisterSource

ddadlani:
Omg 🎉🎉🎉🎉

atc/exec/build/repository_test.go
xtreme-sameer-vohra: Good catch. This was due to the nature of how Ginkgo's Equal works using reflect.DeepEqual which is comparing values. As a result all new fakes of a type are considered to be equal by Ginkgo.

Addressed here - https://github.com/concourse/concourse/pull/4990

atc/exec/build/repository.go
xtreme-sameer-vohra: ah, i fooled myself with an outdated comment 😸

Issue #2936, opened 2018-12-12
Don't use `concourse` tokens.  


pivotal-jwinters:
## What challenge are you facing?

* [Users need to log out and log back in every time they change their auth config](https://github.com/concourse/concourse/issues/2441).
* Fly can't implement a proper oauth flow because `skymarshal` gets in the way.

Currently concourse issues its own auth tokens. 

Originally this happened for two reasons:
1) We encode all your team memberships into your token to make downstream requests more efficient. We figure out all your team memberships once, instead of upon each request. The drawback here is that since your teams live in your token. They don't change unless you log out and log back in.

2) `dex` doesn't support client credentials (yet). Any internal components (`tsa`) that need to talk to the API will generate their own auth tokens. They do this because they have no way of requesting a token from `dex`. Once `dex` supports the `client_credentials` grant type, each component can request a token through the proper channels.

## What would make this better?

We could use `dex` tokens directly... or we could let people configure whatever external auth provider they want. 

At the end of the day, all we need to know is:
* The unique identifier of the user/service
* What teams you should be a member of

This means _in theory_ we should be able to let people configure whatever auth provider they want, and we would provide a way of configuring how `claims` map to `teams`.

The way we do this is already quite free form. The `auth_config` for a `team` contains two fields for each `role`: `users` and `groups`. We take the `sub` and `groups` claims out of your token, iterate over every `role` for every `team`, and try and match your identity to anything thats been configured.

### So should we abandon dex?... No

`dex` is great because it has a really nice pluggable connector model. It lets you write custom logic to integrate with services that don't expose group information through a claim in your token. And it adds a very simple mechanism for maintaining a pool of local users. Plus, it's written in go, which means we can embed a dex server directly in the concourse binary. 

So instead of letting people configure whatever auth provider they want, maybe we just continue to use `dex` as the middle layer. It has connectors for `oidc` and `oauth2` support. This means dex would proxy groups from a 3rd party and return them to concourse in the `groups` claim (in a format that it will understand). Yes its an extra layer, but maybe that's fine.

Regardless of what we decide, I think we would want to continue to ship the concourse binary with _out-of-the-box_ `dex` support, and (it we so choose) give people the ability to bypass it.

pivotal-jwinters:
Couple prereqs
- [ ] Have the webserver handle all the cookie csrf logic #2963
- [ ] Dex needs to support the client credentials grant type 
- [ ] Make dex clients configurable

vito:
[...]

PR #4831, opened 2019-11-27
fix(atc/auditor): reduce audit logs

evanchaoli:
# Why do we need this PR?

We have turned off all audit logs, but realized audit logs are very noisy. A lot of system invoked actions are logged, such as:

```
{"timestamp":"2019-11-27T06:57:58.324099900Z","level":"info","source":"atc","message":"atc.audit","data":{"action":"ListAllPipelines","parameters":{},"user":""}}
{"timestamp":"2019-11-27T06:58:04.259709800Z","level":"info","source":"atc","message":"atc.audit","data":{"action":"ListAllPipelines","parameters":{},"user":""}}

{"timestamp":"2019-11-27T06:59:10.475297499Z","level":"info","source":"atc","message":"atc.audit","data":{"action":"ReportWorkerContainers","parameters":{"worker_name":["7EoikMcK4"]},"user":"system"}}
```

The noise makes it difficult to find useful info.

To not generate audit log if user is empty or "system".

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/auditor/auditor.go
+++ b/atc/auditor/auditor.go
@@ -78,9 +78,14 @@ func (a *auditor) ValidateAction(action string) bool {
 }
 
 func (a *auditor) Audit(action string, userName string, r *http.Request) {
-	err := r.ParseForm()
-	if err == nil && a.ValidateAction(action) {
-		a.logger.Info("audit", lager.Data{"action": action, "user": userName, "parameters": r.Form})
+	switch userName {
+	case "", "system":
+		return
+	default:
+		err := r.ParseForm()
+		if err == nil && a.ValidateAction(action) {
+			a.logger.Info("audit", lager.Data{"action": action, "user": userName, "parameters": r.Form})
+		}
 	}
 }
 


evanchaoli:
Hi @loghen41, just wonder your opinion as you are author of audit log.

logan-hendricks:
@evanchaoli that's kind of you to think to circle me in on this PR.  

So, to give some history, when I wrote the audit service, I debated this exact question for some time.  I eventually decided on the state it is in as it didn't seem correct to filter system level interactions, or unauthenticated level interactions from the audit trail in case users felt that was important "auditible" information. (for example, the company I worked for at the time wanted an audit log of what unauthenticated interactions were being accomplished.  

Looking at this code again right now, I definitely recognize it as an oversight on my part to not have that be a flaggable option.  I think what you have works, but you might get more advantage out of making this into a flag.  (however, that's just my opinion)

If you wanted to do that, it would simply require setting new user flag options in the [atccmd](https://github.com/concourse/concourse/blob/84acaa94d6d7bf9bfd895770ede79ceb1c8dd6d4/atc/atccmd/command.go#L184), so it could be passed in to the [auditor](https://github.com/concourse/concourse/blob/b09a805fd22d0f5bd06fef493920c416dc3bb2d8/atc/auditor/auditor.go#L12), and you could filter the logs according to the userName in [Audit](https://github.com/concourse/concourse/blob/b09a805fd22d0f5bd06fef493920c416dc3bb2d8/atc/auditor/auditor.go#L82)

The following is probably would I would do if I were constructing the PR myself... and represents entirely how I would handle this codewise.  Feel free to use or ignore any of this content.  

I would propose 3 new flags, (enableAuthenticatedUserAuditLogs, enableUnauthenticatedUserAuditLogs, enableSystemUserAuditLogs). 

I would then break out your switch:
```
	switch userName {
	case "", "system":
		return
	default:
```
into a new helper method to the name of **validateUser** within the [auditor](https://github.com/concourse/concourse/blob/b09a805fd22d0f5bd06fef493920c416dc3bb2d8/atc/auditor/auditor.go#L79)

It would probably look something like:
```
func (a *auditor) validateUser(userName string) bool {
	switch userName {
	case "":
            return a.enableUnauthenticatedUserAuditLogs
        case "system":
	    return a.enableSystemUserAuditLogs
	default:
            return a.enableAuthenticatedUserAuditLogs
}
```

This would allow it to be used in the same if block as the ValidateAction method, and it can just filter any user data before filtering the action.

Once again, I'll say that whatever you folks feel is best in the long term is fine, this would probably be what I would propose if I were writing it myself.

pivotal-jamie-klassen:
@evanchaoli Can you talk a bit more about your workflow/use-case? Like how are you storing and searching through these logs? It seems like this extra log noise shouldn't necessarily make it more difficult to find useful info.

I feel obligated to ask the question - should this be a feature of Concourse itself? If your cluster is busy enough and it's important enough for you to look through the logs, it's possible you'd be using a syslog sink/log reporting tool (something like stackdriver or logstash) that might allow you to configure your own filters. This argument could put a change like this in the category of [`quick-fix`](https://github.com/concourse/concourse/wiki/Anti-Patterns#quick-fix).

@evanchaoli I like @logan-hendricks's suggestion about making this configurable, but it also has the potential to be an operator [`knob`](https://github.com/concourse/concourse/wiki/Anti-Patterns#knobs).

Overall I'm a bit wary of the future-proof of those magic strings (`""` for unauthenticated and `"system"` for system accounts) and maintaining this new log-filtering feature that is coupled to them - I'd really like @pivotal-jwinters to weigh in given his knowledge of the burgeoning #2936, which could ultimately affect the `"system"` part.

evanchaoli:
@pivotal-jamie-klassen Yes, we can easily add whatever log filters at our side, but if we go back to the root, what should be audited? IMO, we only want to audit user behaviors. Like the logs lines I pasted, what do those `ListAllPipelines` with empty user tell? Also, "ReportWorkerContainers" with user "system" are just Concourse internal calls, why audit that? If that needs to be audited, why don't just watch ATC logs?

So the problem is not that we are not able to add filter at our side, instead, Concourse should only emit things that really need to be audited.

I believe internal API calls should not be audited. But I'm confused that where those `ListAllPipelines` with empty user come from? I suppose that are called by Concourse itself, it's  not really unauthorized. But when some real unauthorized calls coming in, how to distinct?

pivotal-jamie-klassen:
@evanchaoli It sounds like the purpose of this change is more philosophical (as in "what is _worth_  auditing?") than it is practical (as in "how can I more easily find the audit logs I'm interested in?"). I don't have a strong opinion about what _should not_ be audited, so I'll defer to folks with more of a stake in that use-case (like people who are using Concourse in production at public enterprises that have legal compliance requirements! It's possible you fit in this category yourself 😄).

PS. Taking a shot in the dark, I suspect those `ListAllPipelines` logs are the result of unauthenticated sessions using the dashboard page - maybe a CI monitor.

evanchaoli:
> (like people who are using Concourse in production at public enterprises that have legal compliance requirements! It's possible you fit in this category yourself 😄).
> 

Yes, all the PRs I created are to solve problems we hit in our environment (production, staging, and experimental clusters). But I know each company may practice differently. So this is not an urgent PR to merge, we may take time to see if more folks contribute their opinions. I'm also ok to add CLI options to control the behavior. Anyway, let's wait for some time and see.

> PS. Taking a shot in the dark, I suspect those `ListAllPipelines` logs are the result of unauthenticated sessions using the dashboard page - maybe a CI monitor.

Thanks, that explained, as we have exposed some pipelines. But which adds my confusion for auditing unauthorized actions. The reason why we do audit is to detect abnormal behaviors, sending those normal actions to audit logs just adds difficulty of detecting abnormal. Can such `ListAllPipelines` be treated as from user "system" ? So that this PR may only filter out "system" audit logs and keep unauthorized audit logs.

loghen41:
@evanchaoli First off. I apologize, I realize I was logged into my current work account when I made my comment earlier (which is @logan-hendricks ).  

I actually agree with @pivotal-jamie-klassen on both his points of the earlier recommendation I made being a potential knob.  And to his credit, I also was wary of the future proofing of the "system" and "" users.  If for those reasons the concourse team wants to leave it as is, I would be find protecting the system from being hampered from future growth.

I secondly agree with @pivotal-jamie-klassen that the questions proposed are fairly philosophical.  I'll give the responses as best as I can according to what I understood of the system when I wrote the change, as well as what the system "could" eventually become.  I can also tell you want some data behind your questions @evanchaoli , so I'll explain them in a high level of depth.  I apologize as this might get a bit wordy... but I'd like to frontload some of your concerns if possible. 

>  what should be audited? IMO, we only want to audit user behaviors. Like the logs lines I pasted, what do those ListAllPipelines with empty user tell? 

I think this is valid opinion.  However, in the list of features that Concourse allows, ListAllPipeline for an empty user does describe quite a few things:

To explain, an empty user basically means it's an **authenticated user**.  It is very possible to expose a pipeline or a team on a public view.  This means that unauthenticated users would be able to have access to see those pipelines.  While being able to see a pipeline may seem trivial, but according to the documentation on [exposing](https://concourse-ci.org/exposing.html). You can set multiple levels of permission for an unauthenticated viewer.  (They can see the pipeline, but not the builds).  This is useful if you want to keep build information private, but allow general pipeline history to be viewed.  If that level of granularity is required to be **audited** for business reasons, then it stands to reason you need to log out transactions for those unauthenticated users.

To expound the problem for future proofing.  I believe in some of the latest changes to Concourse (correct me if I'm wrong), but you can set a pipeline to also be triggered by unauthenticated users.  So if as an organization, I want the ability to see what actions are being done by these "empty" (genuinely unauthenticated) users, the only way to do so is to log those actions.  

> Also, "ReportWorkerContainers" with user "system" are just Concourse internal calls, why audit that? If that needs to be audited, why don't just watch ATC logs?

Realistically one of the major lackings that concourse has is that it is very difficult to know from an _audit_ perspective **what** triggered a pipeline.  Let me give you an example, as of right now, if you have a git resource that senses a new commit, and kicks off a new build.  You can find out what triggered the build if you do some pretty good hoop jumping.  You have to happen to know the git resource check # to search for the correct git resource log.  You then have to find the log that triggered the build.  As for now, the **build** was actually triggered by a git resource, but it's logged as a **system** action.  So yes, you could check the ATC logs, but ideally in the future, the auditor would have the ability to log which actual git commit triggered the build (and log that as the user instead of the system).  That level of change was quite large, and I opted instead to build this simple auditor as a potential first step to move towards a larger audit process.  (I also had a pipe dream of going and writing those later portions...... but other work related responsibilities took larger priority)

So the logging of "system" user actions is also completely correct within the audit log, as it declares all of the system related actions (which could realistically be auditible, if not very useful right now)

loghen41:
@evanchaoli I also just saw your second post here, and to add another voice on what is "auditible".  The main reason I wrote this PR was less for debugging related reasons, and more for the ability to capture whatever data was needed for potential Corporate/Enterprise Legal Audit purposes.  So, most of the decisions I made came from that realm.  

Basically, if a legal auditing team at an enterprise software company wanted to have X data, I wanted to make sure it was accessible. It mostly was trying to answer the question at the end of the day "who/what kicked off the build?"

the potential options include:
An authenticated User
An unauthenticated User
A system process (a git resources, s3 resource, etc...)

And as @pivotal-jamie-klassen recommended, I personally was using a system to dump all of the data into Splunk/Cloudwatch system. Which allowed it to be filtered/aggregated into a legal report.  

Using it for debugging purposes also can make sense. But as mentioned, the ATC already has quite a bit of debug information.  The audit delves more into the process of what an auditor/compliance portion of a company may want to have exposed.

evanchaoli:
@loghen41 Thanks for taking time putting all the explanation. You have done great job wrt the feature of audit log. This PR is just trying to make audit logs a little bit better. So discussion is good to have. :)

I'm not against to audit unauthorized behavior, I actually believe that's a very important part. My concern is, should UI invoked `ListAllPipelines` be treated as "unauthorized"? What about some real "unauthorized" hack comes? How to distinct a real "unauthorized" access and UI invoked "normal unauthorized" access?

I'm not all for using audit log for debugging purpose. I don't think audit log provides enough info for debugging, and logs are there for debugging, we also have metrics emitted. Audit logs are supposed to sent to certain people who care .

But I agree the current change of this PR is not perfect for merging. I also agree to add CLI options as you suggested. I'm even OK to discard this PR as long as we, including the Concourse dev team, have thought over.

loghen41:
@evanchaoli . I want to clarify, the logs are not being sent out because of **unauthorized** users.  (meaning they don't have access to the resources).

These are **unauthenticated** users.  Meaning they aren't logged in.. but are permitted to have access to the resource even though they are anonymous.

evanchaoli:
> @evanchaoli . I want to clarify, the logs are not being sent out because of **unauthorized** users. (meaning they don't have access to the resources).
> 
> These are **unauthenticated** users. Meaning they aren't logged in.. but are permitted to have access to the resource even though they are anonymous.
> 
> That being said. I'm not sure I understand your question here:
> 
> > What about some real "unauthorized" hack comes? How to distinct a real "unauthorized" access and UI invoked "normal unauthorized" access?
> 
> Can you go into more detail as to what you mean by that?

Well, my understanding of "unauthorized" is "not logged in and not permitted access" like attacks from hackers, those should definitely be recorded.

If "unauthorized" means "not logged in and permitted", then that's still normal user behavior. In that case, using CLI options to turn off "unauthorized" and "system" audit logs seems to be the best solution.

I still have the concern of distinct normal unauthorized and hacker unauthorized actions. But that could be out of scope of this PR, and I may create a new issue to discuss that.

pivotal-jamie-klassen:
@evanchaoli @loghen41 this has been an excellent discussion of feature narratives around auditing!  Thanks for your careful descriptions - let's make sure this PR doesn't get lost for the sake of future contributors working in this space.

pivotal-jwinters:
In light of #2936, I would actually like for us to move away from caring about things like `system` and `username`. These are non-standard claims in the OIDC world and if we want to support generic auth tokens then we should probably be relying on the `sub` claim. This means concourse would only care about two cases `authenticated` and `unauthenticated`. I can see the argument for wanting a knob to control whether or not we audit unauthenticated events, however  making sense of the actual `sub` (determining `system` vs `user`) should probably be done outside of concourse as @pivotal-jamie-klassen mentioned.

evanchaoli:
Let's discard this PR. I don't want to spend more time on this.

PR #4835, opened 2019-11-27
Request schedule when unpausing a job or pipeline

### LINKED ISSUE #4801: When a pipeline is unpaused, scheduling should be requested for all of its jobs
Similar to https://github.com/concourse/concourse/issues/4802

@clarafu says with 89% certainty that there's already a method to do this (mark all of a job's pipeline as `schedule_requested`).

### LINKED ISSUE #4802: When a job is unpaused, scheduling should be requested for the job
Building on https://github.com/concourse/concourse/issues/4717 - when a job is unpaused, it should request scheduling for itself.

julia-pu:
Depends on PR #4821 

Implements requesting schedule when a job or pipeline is unpaused.

fixes #4801 
fixes #4802


atc/db/pipeline.go
```diff
@@ -1070,3 +1086,46 @@ func resources(pipelineID int, conn Conn, lockFactory lock.LockFactory) (Resourc
 
 	return resources, nil
 }
+
+// The SELECT query orders the jobs for updating to prevent deadlocking.
+// Updating multiple rows using a SELECT subquery does not preserve the same
+// order for the updates, which can lead to deadlocking.
+func requestScheduleForJobsInPipeline(tx Tx, pipelineID int) error {
+	rows, err := psql.Select("id").
```
vito: Can this be simplified to a single query?:

```sql
UPDATE jobs
SET schedule_requested = now()
WHERE pipeline_id = $1
```

Where does the deadlock risk come from? :thinking: Was it something observed during testing?

PR #4841, opened 2019-11-28
add release notes for #4808 and #4817

vito:


vito:
made redundant by #4842

PR #4868, opened 2019-12-07
fly: show status of websocket dial when hijack err

### LINKED ISSUE #4849: Terrible output when trying to fly intercept but request is forbidden
If I'm a `viewer` and I try and hijack a container it fails with `websocket: bad handshake`. This is very confusing.

xtremerui:
# Existing Issue
Fixes #4849 

# Why do we need this PR?
Showing err when hijack fail is not enough for debugging

Show response status as well. So if an viewer wants to hijack into a container the error would be `403 Forbidden websocket: bad hanshake`. It would be obvious that it is about authorization failure.

# Reviewer Checklist
- [ ] Code reviewed
- [ ] PR acceptance performed


xtremerui:
Thx!

Issue #4795, opened 2019-11-20
lets encrypt ACMEv1 disabled


fredwangwang:
## Bug Report

The autocert client used by concourse is disabled by letsencrypt, resulting the following error message when running web with letsencrypt=true:
```
2019/11/20 01:45:30 http: TLS handshake error from 50.232.79.2:49366: acme/autocert: missing certificate
2019/11/20 01:45:33 http: TLS handshake error from 50.232.79.2:61681: 403 urn:acme:error:unauthorized: Account creation on ACMEv1 is disabled. Please upgrade your ACME client to a version that supports ACMEv2 / RFC 8555. See https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430 for details.
```

## Steps to Reproduce

Running web with letsencrypt function

## Expected Results

Certs obtained automatically

## Actual Results

error msg listed above

## Additional Context

> (delete this section if not needed)

## Version Info

* Concourse version: 5.7.1
* Deployment type (BOSH/Docker/binary): BOSH
* Infrastructure/IaaS: GCP

vito:
More context: https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430

suhlig:
Still getting the same old issue with `5.7.2`:

```
Dec 07 18:27:24 ci concourse-web[901]: 2019/12/07 18:27:24 http: TLS handshake error from [REDACTED]:52411: 403 urn:acme:error:unauthorized: Account creation on ACMEv1 is disabled. Please upgrade your ACME client to a version that supports ACMEv2 / RFC 8555. See https://community.letsencrypt.org/t/end-of-life-plan-for-acmev1/88430 for details.
```

Any hints to further debug this? 

Config:

```
export CONCOURSE_TLS_BIND_PORT="443"
export CONCOURSE_ENABLE_LETS_ENCRYPT="True"
```

vito:
Ack, looks like it's because we still have the v1 endpoint as the default value for the flag:

https://github.com/concourse/concourse/blob/c77d756b416274697000837f5712ec3e8bb744ce/atc/atccmd/command.go#L101

Sorry about that. At least the client has been upgraded to support v2 endpoints now.

Until we bump the default value, here's the config you can set to get it working:

```sh
export CONCOURSE_LETS_ENCRYPT_ACME_URL=https://acme-v02.api.letsencrypt.org/directory
```

suhlig:
[...]

PR #4869, opened 2019-12-08
update default let's encrypt acme endpoint

vito:
# Existing Issue

Fixes https://github.com/concourse/concourse/issues/4795 (again)

# Why do we need this PR?

The `autocert` client has been bumped to *support* the v2 API, but the flag still defaults to v1, so things are broken out-of-the-box unless the flag is explicitly set to the v2 endpoint.

* Updates the flag value to point to the v2 URL.

I haven't added test coverage here yet, but it's probably worth covering somehow. The client is bumped and the workaround is easy enough so maybe now we can take some time to try and figure this out. It'd be nice to have automatic coverage so we can notice breakage more quickly (and in this case notice...unworking fixage).

The challenge however is the test setup, since testing Let's Encrypt integration requires that the server under test is reachable by Let's Encrypt's servers - i.e. it needs to be publicly routable. There are also really low rate limits, but we can use their [staging environment](https://letsencrypt.org/docs/staging-environment/) instead which has much higher limits (30,000 registrations per week per domain instead of 50) and is actually intended for this kind of thing.

Should we just change one of our smoke test jobs to deploy with Let's Encrypt? :thinking: 

# Contributor Checklist

- [ ] Unit tests

# Reviewer Checklist

- [x] Code reviewed
- [x] Tests reviewed
- [x] Release notes reviewed
- [ ] PR acceptance performed


vito:
@pivotal-jamie-klassen Added release notes, and added a unit test that ensures the flag's default value matches the default value of the Let's Encrypt client, since we won't be able to catch that in smoke tests where we'll have to explicitly set the flag to their staging URL. :+1:

Issue #4877, opened 2019-12-10
Implement Garden Container interface using containerd container API


deniseyu:
We'll need to have the containerd API talk like a Garden container, [similar to how Houdini works](https://github.com/vito/houdini/blob/master/container.go), so that we can get something working end to end.

- [x] Implement `Handle()` #4913
- [x] Implement `Stop()` https://github.com/concourse/concourse/issues/4918
- [ ] Implement `Info()`
- [ ] Implement `Run()`
- [ ] Implement `Attach()`
- [ ] Implement `Property` management methods (`Property()`, `SetProperty()`, `Properties()`)
- [ ] Fix `Lookup` partial implementation in backend

Check if we still need these:

- [ ] Implement `Current*Limits()` (Bandwidth, CPU, Disk)?
- [ ] Implement `Limit*()` (Bandwidth, CPU, Disk)?
- [ ] Implement `StreamIn()`?
- [ ] Implement `StreamOut()`?
- [ ] Implement `Metrics()`?
- [ ] `SetGraceTime()`?

References:

* [Garden Container interface](https://github.com/cloudfoundry/garden/blob/master/container.go)
* [Houdini implementation](https://github.com/vito/houdini/blob/master/container.go)
* [Alex's spike code from before](https://github.com/concourse/concourse/blob/containerd-impl/gcontainerd/container.go)

cirocosta:
Hey,

I just updated https://github.com/concourse/concourse/issues/4783 to have all the methods so that we keep track of it all in that single issue.

Thanks!

PR #4881, opened 2019-12-11
Improve containerd implementation of Garden backend interface methods

deniseyu:
This PR incorporates code review feedback from #4784, refactors out error types, and adds additional edge case handling to the runtime backend. Some interface methods are not yet implemented (`BulkInfo`, `BulkMetrics`, `Lookup`, and `GraceTime`) because they either read properties of a concrete implementation of the Garden `Container` interface, or they need to return this concrete type. That work will be delivered after #4877.

Everything is behind a feature flag that isn't exposed yet in any packaging, so is safe to continuously merge into master.

# Existing Issue

Addresses #4783 .

* Create backend error types that can wrap client-side errors 
* Use context with timeout for all function calls that cross the backend/client boundary
* Add more error checks for Destroy edge cases
* Add godocs for backend methods

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] PR acceptance performed

### DIFF
--- a/worker/backend/backend.go
+++ b/worker/backend/backend.go
@@ -136,12 +136,12 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
 // Errors:
 // * TODO.
 func (b *Backend) Destroy(handle string) error {
-	ctx := namespaces.WithNamespace(context.Background(), b.namespace)
-	ctxWithTimeout, _ := context.WithTimeout(ctx, b.clientTimeout)
-
 	if handle == "" {
 		return InputValidationError{Message: "handle is empty"}
 	}
+
+	ctx := namespaces.WithNamespace(context.Background(), b.namespace)
+	ctxWithTimeout, _ := context.WithTimeout(ctx, b.clientTimeout)
 
 	container, err := b.client.GetContainer(ctxWithTimeout, handle)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


worker/backend/backend.go
```diff
@@ -112,36 +138,41 @@ func (b *Backend) Create(gdnSpec garden.ContainerSpec) (container garden.Contain
...
+		if err != nil {
+			return ClientError{ InnerError: err }
 		}
-
-		return b.client.Destroy(ctx, handle)
 	}
 
-	timeDelimitedContext, _ := context.WithTimeout(ctx, maxTaskKillWaitTime)
-	err = killTasks(timeDelimitedContext, task)
+	err = b.client.Destroy(ctxWithTimeout, handle)
```
kcmannem: is the reason we switched from killTasks to Destroy to avoid killing gracefully?

deniseyu: It's just a refactor - we only want to call `killTasks` if we can reference the container without error (line 165). This is kind of a complicated piece of logic that we attempted to explain better in the unit tests -- we moved `killTasks` up, so that Destroy gets in both of these circumstances:

1. A container is found but no Task was ever started inside it (line 161). This means that container creation was somehow interrupted, maybe by a network partition, so we want this container to be cleanly deletable later on by a GC. We fence off the "task does not exist" genre of error in 161.
1. A container is found and its Task was successfully stopped.

In both of these cases we want Destroy to be the next thing that happens. If no errors trigger early returns, then this should be the control flow now!

PR #4886, opened 2019-12-11
add version_md5 to resouce_cache table

xtremerui:
# Why do we need this PR?
look up by version_md5 (text)  is faster than version (jsonb) 

added `version_md5` column to `resource_caches` table. Use version_md5 when quering.

### DIFF
--- a/atc/db/resource_cache.go
+++ b/atc/db/resource_cache.go
@@ -60,13 +60,11 @@ func (cache *ResourceCacheDescriptor) findOrCreate(
 		err = psql.Insert("resource_caches").
 			Columns(
 				"resource_config_id",
-				"version",
 				"version_md5",
 				"params_hash",
 			).
 			Values(
 				resourceConfig.ID(),
-				cache.version(),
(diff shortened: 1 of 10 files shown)


atc/db/resource_cache.go
```diff
@@ -61,20 +61,22 @@ func (cache *ResourceCacheDescriptor) findOrCreate(
 			Columns(
 				"resource_config_id",
 				"version",
```
pivotal-jwinters: Shouldn't we remove this reference to `version`?

xtremerui: After discussion we found it would be easier to keep both version and version_md5 in the table. 
For example, in https://github.com/concourse/concourse/blob/9613e27cf55d58b81dd5439d0f05c93d721c5204/atc/db/resource_cache_factory.go#L128

if `version` column is dropped then it will need to make a query to resource_config_versions table with `version_md5`, it is not clear that if the version in  `resource_config_versions` is always sync with `resource_caches` table  in all cases.

PR #4907, opened 2019-12-16
Update main team config test

taylorsilva:
The next version of the chart makes the ConfigMap for you. The test can
now pass in the YAML for the main team instead of manually making the
ConfigMap first.

# Why do we need this PR?
k8s-topgun will break when the next major version of the chart is released because the expected value of `mainTeam.config` has changed: https://github.com/concourse/concourse-chart/pull/33#pullrequestreview-332844796

I locally ran this test on the `topgun` GKE cluster and it passed.


deniseyu:
We need to merge this into `release/5.7.x` too for the `helm-prs` pipeline to pick it up

PR #4916, opened 2019-12-17
implement Handle for backend/container

### LINKED ISSUE #4913: [Runtime] Implement Garden.Container.Handle() Method with the containerd backend
We use the `handle` as the unique identifier for containers in concourse and build business logic around it. Its a good first method to implement for the larger epic https://github.com/concourse/concourse/issues/4877

kcmannem:
Fixes #4913 

Signed-off-by: Bin Ju <bju@pivotal.io>


cirocosta:
> Not blocking, but wonder if, later on, we should add a better test for Containers now that it returns a set of real containers.

heey, is there an issue created for that?


thanks!

PR #4921, opened 2019-12-18
implement `Stop()` in backend/container

### LINKED ISSUE #4918: [Runtime] Implement Garden.Container.Stop() Method with the containerd backend
For the larger epic #4877
We use Stop() to send SIGTERM/SIGKILL to the process inside the container. NOTICE: a Stopped container is not a DESTROYED container, file copies are still feasible to or from a stopped container.

TODOs:
- [x] How can we operate a `containerd` garden container? Do we need a reference to the original `containerd` container, and delegate the request to `containerd` container?
- [x] Match the `Stop()` to `containerd` container's operation (<- Task kill and delete)
- [x] Test cases (unit & integration)
- [ ] Think about `Delete` container

pivotal-bin-ju:
fixes #4918

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


deniseyu:
Force-pushed to resolve merge conflicts

cirocosta:
Hey,

I don't think this logic is right according to the garden spec:

```
	// Stop stops a container.
	//
	// If kill is false, garden stops a container by sending the processes running inside it the SIGTERM signal.
	// It then waits for the processes to terminate before returning a response.
	// If one or more processes do not terminate within 10 seconds,
	// garden sends these processes the SIGKILL signal, killing them ungracefully.
	//
	// If kill is true, garden stops a container by sending the processing running inside it a SIGKILL signal.
	//
```

(see https://github.com/cloudfoundry/garden/blob/60b076cc47496373f5b7011b9c7200f229c5aee7/container.go#L13)

i.e.:

1. when issuing a `Stop()` with a `kill=false`, it should indeed send `SIGTERM`
(to give it a chance of gracefully finishing), but ultimatelly, it's supposed to
send a SIGKILL to try harder to termiante it.

2. when issuing a `Stop()`, it should wait for the process to be terminated



To see that in practice in today's `gdn`, we can do the following:

i. create a sample application that ignores `SIGTERM`


        #include <stdlib.h>
        #include <unistd.h>
        #include <stdio.h>
        #include <string.h>
        #include <signal.h>

        void
        handle_signal(int sig)
        {
                printf("ignoring %s\n", strsignal(sig));
        }

        int
        main(int argc, char** argv)
        {
                if (signal(SIGTERM, handle_signal) == SIG_ERR) {
                        perror("signal");
                        return 1;
                }

                while (1) {
                        pause();
                }

                return 0;
        }


ii. put it in a rootfs, then create a container of it and run a process

        gaol create --rootfs=raw:///rootfs --handle=test
        gaol run --command /a.out test


iii. run `container.Stop()` on it.

        time curl -XPUT \
                localhost:7777/containers/test/stop \
                -H "Content-Type: application/json" \
                -d '{}'


        real    0m11.024s
        user    0m0.004s
        sys     0m0.004s


From that, we can see that the `stop` request takes >= 10 seconds (the grace
period while it sends `SIGTERM`s). With something like [`killsnoop`] we can even
see the signals:

        TIME      PID    COMM             SIG
        17:26:09  17482  gdn              SIGTERM
        17:26:10  17482  gdn              SIGTERM
        17:26:11  17482  gdn              SIGTERM
        17:26:12  17482  gdn              SIGTERM
        17:26:13  17482  gdn              SIGTERM
        17:26:14  17482  gdn              SIGTERM
        17:26:15  20073  gdn              SIGTERM
        17:26:16  20073  gdn              SIGTERM
        17:26:17  20073  gdn              SIGTERM
        17:26:17  20073  gdn              SIGKILL       << booom!


[`killsnoop`]: https://github.com/iovisor/bcc/blob/8bb4e473bac87e65c9aaa51c9e90e2f98bb25acc/tools/killsnoop_example.txt


It's nice that those two things are already `killTasks` (in `backend.go`) are
already implemented though - I think we could reusa that here, with a slightly
variation on the case when we don't want to be graceful at first (`kill=true`).

Do you think we could reuse that?

thx!

worker/backend/container.go
```diff
@@ -1,39 +1,68 @@
...
-// It is possible to copy files in to and out of a stopped container.
-// It is only when a container is destroyed that its filesystem is cleaned up.
+// If kill is false, the backend stops a container by sending the processes running inside it the SIGTERM signal.
+// If kill is true, the processing is ungracefully terminated with a SIGKILL signal.
 //
 // Errors:
 // * None.
-func (c *Container) Stop(kill bool) (err error) { return }
+func (c *Container) Stop(kill bool) error {
+	task, err := c.containerdContainer.Task(c.ctx, nil)
```
cirocosta: If I understood correctly, the idea here is that `ctx` will come with a timeout, which would enforce a total duration to wait for the process to be stopped - is that right?

If so, I don't think that's what actually going on under the hood - when you create a context with a particular timeout, "the clock starts ticking": a deadline is set at that very exact moment, thus, time is already "running out" from that very moment.

For instance, consider the following snippet:

```go
func sleepHalfTheTime(ctx context.Context, total time.Duration) {
	time.Sleep(total >> 1)
}

func doSomethingUntilContextExpieration(ctx context.Context) {
	start := time.Now()
	defer func() {
		fmt.Printf("ELAPSED: %s\n", time.Now().Sub(start))
	}()

	<-ctx.Done()
}

func main() {
	total := 10 * time.Second

	ctx, _ := context.WithTimeout(context.TODO(), 10*time.Second)

	sleepHalfTheTime(ctx, total)
	doSomethingUntilContextExpieration(ctx)
}
```

this would result in:


```
ELAPSED: 5.000082213s
```

see [`WithTimeout` using `WithDeadline`](https://github.com/golang/go/blob/0ad368675bae1e3228c9146e092cd00cfb29ac27/src/context/context.go#L496)

```go
func WithTimeout(parent Context, timeout time.Duration) (Context, CancelFunc) {
	return WithDeadline(parent, time.Now().Add(timeout))
}
```

pivotal-bin-ju: I think that might be a topic how a garden container's `Stop()`  is called.
There is a bool parameter to indicate how the container is stoped. Can we assume the kill process is always `SIGTERM`, then `SIGKILL` here?
I guess the assumption may be:
- the caller is responsible for the context setup, like `WithTimeout` or `WithCancel`. If the context is a timeout context, the caller is responsible for the timeout duration as well.
- the caller is responsible for calling the `Stop()` with `true` or `false`. and here is example how garden stop() a container: https://github.com/cloudfoundry/garden/blob/044d7ac89d98e7ae70aacb86fce9ff434364bfe8/server/request_handling.go#L196

WDYT?

worker/backend/container.go
```diff
@@ -33,7 +42,36 @@ func (c *Container) Handle() (handle string) { return c.handle }
...
+	if err != nil {
+		return err
+	}
+
+	result, err := task.Delete(c.ctx)
+	if err != nil {
+		return err
+	}
+
+	if result != nil && result.Error() != nil && !errdefs.IsNotFound(result.Error()) {
```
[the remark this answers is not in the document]
pivotal-bin-ju: good catch. It has been fixed. my original intension is converting  the not-found-error to nil

cirocosta:
Hey @pivotal-bin-ju , would you mind also rebasing please? There have been some bumps to containerd too 😁  thanks!

cirocosta:
thanks for the PR! I'll close this is it's been superseded by #5114

Issue #4790, opened 2019-11-19
Re-runs should not become the 'last finished build'


vito:
Currently, if I have builds `7 8 9` and I click the job in the pipeline, it takes me to `9`.

Once I re-trigger `8`, the pipeline shows a 'running' animation, and clicking the job takes me to `8.1`. I feel like this isn't too unreasonable since I most likely want to see what's running.

However, once `8.1` completes, I would expect clicking the job to take me to `9`.

vito:
@Lindsayauchin @matthewpereira We should figure out the UX for this as it's kind of confusing - ping me when you have a moment!

vito:
* `next_build` should point to a rerun *only* if it's a rerun of the latest build
* `finished_build` should point to a rerun *only* if it's a rerun of the latest build

Right now we update `finished_build` whenever a build finishes; might need to make this conditional?

xtremerui:
@zoetian @pivotal-jamie-klassen could UX team provide some insight of the usage of `transtion_build` in `/api/v1/jobs` in dashboard? With the context of re-run builds and per vito's comment for `next_build` and `finished_build`, what behavior is expected from UX for 'transition_build' ? Thx.

xtremerui:
`transition_build` depends on the `finished_build`. By inheritance, 

`transition_build` should point to a status changed rerun only if it's a rerun of the latest build

PR #4926, opened 2019-12-18
atc: update rerun build behavior

xtremerui:
Implements https://github.com/concourse/concourse/issues/4790

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
- [x] Code reviewed
- [x] Tests reviewed


atc/db/build.go
```diff
@@ -1617,7 +1650,7 @@ func updateTransitionBuildForJob(tx Tx, jobID int, buildID int, buildStatus Buil
 		return nil
 	}
 
-	if latestStatus != buildStatus {
+	if latestStatus != buildStatus && (rerunID == 0 || rerunID == latestID) {
```
pivotal-jamie-klassen: This line started my investigation. I tried this change to discover the failing test:
```suggestion
	if latestStatus != buildStatus && (rerunID == latestID) {
```

atc/db/build_test.go
```diff
@@ -2233,3 +2339,17 @@ func convertToMD5(version atc.Version) string {
...
+	err := psql.Select(col).
+		From("jobs").
+		Where(sq.Eq{"id": jobID}).
+		RunWith(dbConn).
+		QueryRow().
+		Scan(&result)
+	Expect(err).ToNot(HaveOccurred())
+
+	return result
+}
```
pivotal-jamie-klassen: I think that this helper should not exist, for the same reason that we generally test public interfaces instead of implementation details - to help our tests remain flexible and resilient to refactors. I can imagine that the database schema that underlies these links (finished/next/transition) between builds and jobs could be refactored, say, to use a join table or a materialized view (I actually remember when the transition build idea was implemented via a materialized view). If this happens these tests will break for no good reason!

atc/db/build_test.go
```diff
@@ -377,6 +377,112 @@ var _ = Describe("Build", func() {
...
+						Expect(nextBuild).To(BeNil())
+					})
+
+					It("does not updates job latest finished build id", func() {
+						Expect(getJobBuildID(latestCompletedBuildCol, job.ID())).To(Equal(pdBuild.ID()))
+					})
+
+					It("does not updates transition build id", func() {
+						Expect(getJobBuildID(transitionBuildCol, job.ID())).To(Equal(pdBuild.ID()))
+					})
```
pivotal-jamie-klassen: This is the test case that got me so confused.

atc/db/build_test.go
```diff
@@ -377,6 +377,112 @@ var _ = Describe("Build", func() {
...
+		Context("rerunning a build", func() {
+			var (
+				pdBuild, pdBuild2, rrBuild db.Build
+				err                        error
+				latestCompletedBuildCol    = "latest_completed_build_id"
+				nextBuildCol               = "next_build_id"
+				transitionBuildCol         = "transition_build_id"
+			)
+
+			Context("when there is a pending build that is not a rerun", func() {
```
pivotal-jamie-klassen: This phrasing gives me some trouble - I _think_ that 'being a rerun' is more exceptional than not being a rerun. So I'd rather have a test like this just say `when there is a pending build`, and if it's relevant that the pending build is a rerun, we call it out in those cases.

Unless we really need a word for 'not a rerun`, in which case can we invent one? an 'original build' or something like that?

atc/db/build_test.go
xtremerui: The commit that moving from views to columns is here 2c47bef

@clarafu probably have better insights of this helper functions vs exposed interface discussion.

xtremerui:
@pivotal-jamie-klassen https://github.com/concourse/concourse/issues/4958 is created to track the build.go refactoring so we could merge this one to unblock 6.0.0 beta. Thx for the comment!

PR #4928, opened 2019-12-19
Fix k8s topgun 5.5.x

YoussB:
# Why do we need this PR?
- looks like some structuring has made the k8s topgun folder structure different leading to the failure of one of the tls termination tests.


YoussB:
I was able to get it to pass inside the task container after doing this change.

PR #4937, opened 2019-12-19
redo vault var source config flow

### LINKED ISSUE #4934: setting pipeline  with`var_source` of type `vault` with `auth_params` fails
## Bug Report

While performing acceptance on #4927 I configured a vault server with the approle auth backend. When I tried to set a pipeline that made use of this vault server as a var_source, I got some wacky errors.

## Steps to Reproduce

1. I used the following `docker-compose.override.yml`:

```yaml
version: '3'
services:
  vault:
    image: vault
    cap_add: [IPC_LOCK]
    ports: ["8200:8200"]
    volumes:
    - ./vault-config:/vault/config
    command: server
```

and the directory `./vault-config` contains the following `config.hcl`:

```hcl
listener "tcp" {
  address = "0.0.0.0:8200"
  tls_disable = true
}

storage "file" {
  path = "/vault/file"
}
```

With this configuration running against my local Docker server, I followed the [approle instructions from our docs](https://concourse-ci.org/vault-credential-manager.html#vault-approle-auth) to set up the approle auth backend. Then I tried to run `fly -t dev set-pipeline -p pipeline -c pipeline.yml` with the following `pipeline.yml`

```yaml
var_sources:
- name: vault
  type: vault
  config:
    url: http://vault:8200
    path_prefix: concourse
    auth_backend: approle
    auth_params:
      role_id: "<the role id i got when following instructions>"
      secret_id: "<the secret id i got when following instructions>"
jobs:
- name: test
  public: true
  plan:
  - task: echo-secret
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: echo
        args: ["((vault:text))"]
```

## Expected Results

```
pipeline created!
you can view your pipeline here: http://localhost:8080/teams/main/pipelines/pipeline
```

## Actual Results

```
could not reach the Concourse server called dev:

    Put http://localhost:8080/api/v1/teams/main/pipelines/pipeline/config: EOF

is the targeted Concourse running? better go catch it lol
```

## Additional Context

when I look in the web logs, I see
```
2019/12/19 16:01:50 http: panic serving 172.20.0.1:46490: interface conversion: interface {} is map[string]interface {}, not map[string]string
goroutine 1709 [running]:
net/http.(*conn).serve.func1(0xc000696460)
	/usr/local/go/src/net/http/server.go:1767 +0x149
panic(0x33f7de0, 0xc00089c390)
	/usr/local/go/src/runtime/panic.go:679 +0x1e0
github.com/concourse/concourse/atc/creds/vault.(*VaultManager).Config(0xc00062e1e0, 0xc00089c300)
```

so I'm thinking [this type cast](https://github.com/concourse/concourse/blob/master/atc/creds/vault/manager.go#L135) doesn't work.

## Version Info

* Concourse version: 5.8.0 RC (master)
* Deployment type (BOSH/Docker/binary): docker-compose
* Infrastructure/IaaS: local
* Did this used to work? no

### LINKED ISSUE #4935: `var_source` of type `vault` errors when CA contents are passed to `ca_cert` field
## Bug Report

While performing acceptance on #4927, I was following the instructions in [our docs](https://concourse-ci.org/vault-credential-manager.html) and ended up setting up a vault server with TLS enabled. So when I tried to run a test pipeline, I used the `ca_cert` field and instinctively pasted the CA I generated into that field as a multi-line YAML string. Unfortunately this gave an error!

## Steps to Reproduce

`docker-compose up` with the following `docker-compose.override.yml`:
```yaml
version: '3'
services:
  web:
    volumes:
    - ./vault-certs:/vault-certs
  vault:
    image: vault
    cap_add: [IPC_LOCK]
    ports: ["8200:8200"]
    volumes:
    - ./vault-config:/vault/config
    - ./vault-certs:/vault/certs
    command: server
```
where `./vault-config` contains this `config.hcl`:
```hcl
listener "tcp" {
  address = "0.0.0.0:8200"
  tls_cert_file = "/vault/certs/vault.crt"
  tls_key_file = "/vault/certs/vault.key"
}

storage "file" {
  path = "/vault/file"
}
```
and `./vault-certs` contains `vault-ca.crt` (the CA), `vault.crt` and `vault.key` (a certificate signed by that CA, and its private key).

Then follow the steps in our docs for [configuring the secrets engine](https://concourse-ci.org/vault-credential-manager.html#configuring-the-secrets-engine), and do a `vault write /concourse/main/text value='Hello World!'`.

Then run `fly -t dev set-pipeline -p pipeline -c pipeline.yml` where `pipeline.yml` is:
```yaml
var_sources:
- name: vault
  type: vault
  config:
    url: https://vault:8200
    path_prefix: concourse
    ca_cert: |
      -----BEGIN CERTIFICATE-----
      MIIE5jCCAs6... contents of vault-ca.crt
      -----END CERTIFICATE-----
    client_token: "<client token from https://concourse-ci.org/vault-credential-manager.html#vault-periodic-token>"
jobs:
- name: test
  public: true
  plan:
  - task: echo-secret
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: echo
        args: ["((vault:text))"]
```
Then unpause the pipeline and run `fly -t dev trigger-job -w -j pipeline/test`.

## Expected Results

```
started pipeline/test #1

initializing
fetching busybox@sha256:24fd20af232ca4ab5efbf1aeae7510252e2b60b15e9a78947467340607cd2ea2
322973677ef5 [======================================] 743.1KiB/743.1KiB
running echo Hello World!
Hello World!
succeeded
```

## Actual Results

```
started pipeline/test #1

create var_source 'vault' error: Error loading CA File: open -----BEGIN CERTIFICATE-----
MIIE5jC... contents of vault-ca.crt
-----END CERTIFICATE-----
: file name too long
errored
```

## Additional Context

I tried changing the value of `ca_cert` to `/vault-certs/vault-ca.crt` and it worked, but since I see a different parameter called [`ca_path`](https://github.com/concourse/concourse/blob/master/atc/creds/vault/manager.go#L121) I would expect that `ca_cert` should contain the contents of the cert itself, rather than a filesystem path.

## Version Info

* Concourse version: 5.8.0 RC (master)
* Deployment type (BOSH/Docker/binary): docker-compose
* Infrastructure/IaaS: local
* Did this used to work? no

vito:
# Existing Issue

Fixes #4934.

Fixes #4935.

* leverage mapstructure for type validation and coercion
* take literal values for ca/client cert/key instead of file paths
* backfill unit tests

# Contributor Checklist
- [x] Unit tests
- [x] ~~Integration tests (if applicable)~~ Vault topgun already exists and should verify that the flag config forms haven't changed backwards-incompatibly.
- [x] Updated documentation (located at https://github.com/concourse/docs)
- [x] ~~Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)~~

# Reviewer Checklist
- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [x] PR acceptance performed - verified that #4934 and #4935 are fixed. however, I'd honestly be willing to consider not fixing #4935 if it saved enough maintenance burden.

### DIFF
--- a/atc/creds/vault/api_client.go
+++ b/atc/creds/vault/api_client.go
@@ -1,13 +1,17 @@
 package vault
 
 import (
+	"crypto/tls"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"path"
 	"strings"
 	"sync/atomic"
 	"time"
 
 	"code.cloudfoundry.org/lager"
+	"github.com/hashicorp/go-rootcerts"
 	vaultapi "github.com/hashicorp/vault/api"
 )
 
@@ -18,7 +22,7 @@ type APIClient struct {
 
 	apiURL     string
 	namespace  string
-	tlsConfig  *vaultapi.TLSConfig
+	tlsConfig  TLSConfig
 	authConfig AuthConfig
 
 	clientValue *atomic.Value
@@ -27,7 +31,7 @@ type APIClient struct {
 }
 
 // NewAPIClient with the associated authorization config and underlying vault client.
-func NewAPIClient(logger lager.Logger, apiURL string, tlsConfig *vaultapi.TLSConfig, authConfig AuthConfig, namespace string) (*APIClient, error) {
+func NewAPIClient(logger lager.Logger, apiURL string, tlsConfig TLSConfig, authConfig AuthConfig, namespace string) (*APIClient, error) {
 	ac := &APIClient{
 		logger: logger,
 
@@ -174,7 +178,7 @@ func (ac *APIClient) setClient(client *vaultapi.Client) {
 func (ac *APIClient) baseClient() (*vaultapi.Client, error) {
 	config := vaultapi.DefaultConfig()
 
-	err := config.ConfigureTLS(ac.tlsConfig)
+	err := ac.configureTLS(config.HttpClient.Transport.(*http.Transport).TLSClientConfig)
 	if err != nil {
 		return nil, err
 	}
@@ -194,6 +198,73 @@ func (ac *APIClient) baseClient() (*vaultapi.Client, error) {
 	}
 
 	return client, nil
+}
+
+func (ac *APIClient) configureTLS(config *tls.Config) error {
+	if ac.tlsConfig.CACert != "" || ac.tlsConfig.CACertFile != "" || ac.tlsConfig.CAPath != "" {
+		rootConfig := &rootcerts.Config{
+			CAFile:        ac.tlsConfig.CACertFile,
+			CAPath:        ac.tlsConfig.CAPath,
+			CACertificate: []byte(ac.tlsConfig.CACert),
+		}
+
+		if err := rootcerts.ConfigureTLS(config, rootConfig); err != nil {
+			return err
+		}
+	}
+
+	if ac.tlsConfig.ClientCertFile != "" {
+		content, err := ioutil.ReadFile(ac.tlsConfig.ClientCertFile)
+		if err != nil {
+			return err
+		}
+
+		ac.tlsConfig.ClientCert = string(content)
+	}
+
+	if ac.tlsConfig.ClientKeyFile != "" {
+		content, err := ioutil.ReadFile(ac.tlsConfig.ClientKeyFile)
+		if err != nil {
+			return err
+		}
+
+		ac.tlsConfig.ClientKey = string(content)
+	}
+
+	if ac.tlsConfig.Insecure {
+		config.InsecureSkipVerify = true
+	}
+
+	var clientCert tls.Certificate
+	foundClientCert := false
+
+	switch {
+	case ac.tlsConfig.ClientCert != "" && ac.tlsConfig.ClientKey != "":
+		var err error
+		clientCert, err = tls.X509KeyPair([]byte(ac.tlsConfig.ClientCert), []byte(ac.tlsConfig.ClientKey))
+		if err != nil {
+			return err
+		}
+
+		foundClientCert = true
+	case ac.tlsConfig.ClientCert != "" || ac.tlsConfig.ClientKey != "":
+		return fmt.Errorf("both client cert and client key must be provided")
+	}
+
+	if foundClientCert {
+		// We use this function to ignore the server's preferential list of
+		// CAs, otherwise any CA used for the cert auth backend must be in the
+		// server's CA pool
+		config.GetClientCertificate = func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
+			return &clientCert, nil
+		}
+	}
+
+	if ac.tlsConfig.ServerName != "" {
+		config.ServerName = ac.tlsConfig.ServerName
+	}
+
+	return nil
 }
 
 func (ac *APIClient) clientWithToken(token string) (*vaultapi.Client, error) {
--- a/atc/creds/vault/manager.go
+++ b/atc/creds/vault/manager.go
@@ -10,58 +10,54 @@ import (
 	"code.cloudfoundry.org/lager"
 
 	"github.com/concourse/concourse/atc/creds"
-	vaultapi "github.com/hashicorp/vault/api"
+	"github.com/mitchellh/mapstructure"
 )
 
 type VaultManager struct {
-	URL string `long:"url" description:"Vault server address used to access secrets."`
+	URL string `mapstructure:"url" long:"url" description:"Vault server address used to access secrets."`
 
-	PathPrefix string `long:"path-prefix" default:"/concourse" description:"Path under which to namespace credential lookup."`
-	SharedPath string `long:"shared-path" description:"Path under which to lookup shared credentials."`
-	Namespace  string `long:"namespace"   description:"Vault namespace to use for authentication and secret lookup."`
+	PathPrefix string `mapstructure:"path_prefix" long:"path-prefix" default:"/concourse" description:"Path under which to namespace credential lookup."`
+	SharedPath string `mapstructure:"shared_path" long:"shared-path" description:"Path under which to lookup shared credentials."`
+	Namespace  string `mapstructure:"namespace" long:"namespace"   description:"Vault namespace to use for authentication and secret lookup."`
 
-	TLS  TLS
-	Auth AuthConfig
+	TLS TLSConfig  `mapstructure:",squash"`
+	Auth AuthConfig `mapstructure:",squash"`
 
 	Client        *APIClient
 	ReAuther      *ReAuther
 	SecretFactory *vaultFactory
 }
 
-type TLS struct {
-	CACert     string `long:"ca-cert"              description:"Path to a PEM-encoded CA cert file to use to verify the vault server SSL cert."`
+type TLSConfig struct {
+	CACert     string `mapstructure:"ca_cert"`
+	CACertFile string `long:"ca-cert"              description:"Path to a PEM-encoded CA cert file to use to verify the vault server SSL cert."`
 	CAPath     string `long:"ca-path"              description:"Path to a directory of PEM-encoded CA cert files to verify the vault server SSL cert."`
-	ClientCert string `long:"client-cert"          description:"Path to the client certificate for Vault authorization."`
-	ClientKey  string `long:"client-key"           description:"Path to the client private key for Vault authorization."`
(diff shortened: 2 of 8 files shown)


atc/creds/vault/manager_test.go
```diff
@@ -57,4 +67,157 @@ var _ = Describe("VaultManager", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+
+			serverCertBytes, err := serverCert.Export()
+			Expect(err).ToNot(HaveOccurred())
+
+			clientCertBytes, err := clientCert.Export()
+			Expect(err).ToNot(HaveOccurred())
+
+			caBytes, err := ca.Export()
+			Expect(err).ToNot(HaveOccurred())
```
pivotal-jamie-klassen: I've found myself doing this kind of dance (generating a TLS cert for vault subject to the constraints that the golang TLS clients demand) several times since I've started working on concourse full-time. I've done it with [bosh interpolate](https://bosh.io/docs/variable-types/#certificate), I've done it with the `openssl` CLI and cooking up a custom `req.conf`, and just today I learned about `certstrap` (from our [own docs](https://concourse-ci.org/vault-credential-manager.html) lol). Now that you've reinvented this wheel one more time, maybe we should just make a `concourse` subcommand that generates certs specifically for vault?

atc/creds/vault/manager.go
```diff
@@ -90,50 +86,27 @@ func (manager *VaultManager) MarshalJSON() ([]byte, error) {
...
 
-func toString(s interface{}) string {
-	if s == nil {
-		return ""
-	}
-	return s.(string)
-}
+func (manager *VaultManager) Config(config map[string]interface{}) error {
+	// apply defaults
+	manager.PathPrefix = "/concourse"
```
pivotal-jamie-klassen: This is a good catch, as it wasn't there in the initial implementation and I would get errors when specifying a `var_source` without `path_prefix`.

atc/creds/vault/manager_test.go
```diff
@@ -57,4 +67,157 @@ var _ = Describe("VaultManager", func() {
...
+			})
+
+			It("has sane defaults", func() {
+				Expect(configErr).ToNot(HaveOccurred())
+
+				Expect(manager.PathPrefix).To(Equal("/concourse"))
+				Expect(manager.Auth.RetryMax).To(Equal(5 * time.Minute))
+				Expect(manager.Auth.RetryInitial).To(Equal(time.Second))
+			})
+		})
```
pivotal-jamie-klassen: I was going to say this needs to be separately tested since I noticed the default `path_prefix` missing, but you obviously thought of it first.

atc/creds/vault/manager.go
```diff
@@ -90,50 +86,27 @@ func (manager *VaultManager) MarshalJSON() ([]byte, error) {
...
+	manager.Auth.RetryMax = 5 * time.Minute
+	manager.Auth.RetryInitial = time.Second
```
pivotal-jamie-klassen: It's too bad that these are duplicating the defaults as declared in the struct tags above - is there some way, perhaps through [reflection magic](https://golang.org/pkg/reflect/#Value.FieldByName), to determine the defaults dynamically? Is it worth the readability cost?

atc/creds/vault/api_client.go
```diff
@@ -196,6 +200,73 @@ func (ac *APIClient) baseClient() (*vaultapi.Client, error) {
 	return client, nil
 }
 
+func (ac *APIClient) configureTLS(config *tls.Config) error {
+	if ac.tlsConfig.CACert != "" || ac.tlsConfig.CACertFile != "" || ac.tlsConfig.CAPath != "" {
+		rootConfig := &rootcerts.Config{
+			CAFile:        ac.tlsConfig.CACertFile,
+			CAPath:        ac.tlsConfig.CAPath,
+			CACertificate: []byte(ac.tlsConfig.CACert),
+		}
```
pivotal-jamie-klassen: Should we mention somewhere, maybe in the docs, about the precedence of these fields? according to https://github.com/hashicorp/go-rootcerts/blob/master/rootcerts.go#L17-L28, `CAFile` has the highest precedence, followed by `CACertificate`, followed by `CAPath`.

atc/creds/vault/api_client.go
```diff
@@ -196,6 +200,73 @@ func (ac *APIClient) baseClient() (*vaultapi.Client, error) {
...
+		config.GetClientCertificate = func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
+			return &clientCert, nil
+		}
+	}
+
+	if ac.tlsConfig.ServerName != "" {
+		config.ServerName = ac.tlsConfig.ServerName
+	}
+
+	return nil
```
pivotal-jamie-klassen: whoa, deja vu https://github.com/hashicorp/vault/blob/2c21ef4df684d60030f6282997a0ad38f824bcaf/api/client.go#L190-L232. This part seemed like too much code with too few tests... and I bet the hashicorp codebase has tests for all these branches. Would it make sense to write the `tlsConfig.ClientCert` to a temporary file or something as a way to actually properly reuse their code?

This intuitively feels like the kind of place an unsuspecting person would just inject a little change without realizing that it's supposed to pretty much identically simulate the equivalent function in that `vault/api` package...

vito: The tests actually cover most of it, except for `insecure_skip_verify`, plus a few more 'sad path' cases could probably be added to ensure the values are respected. I hand-validated it by changing the values in the tests and watching them break but didn't commit to adding additional examples for them. I can go back and add them.

To be honest, this is all pretty basic TLS configuration that is either correct or totally broken; there's not much to simulate in `vault/api` that wouldn't be done in the exact same way in other projects or if we had just written it ourselves. It was just disappointing that I had to carry over so much just because their client wrapper doesn't support configuring via cert content instead of file paths everywhere. :cry:

atc/creds/vault/api_client.go
vito: I think we should actually just remove support for `CAPath` as we migrate to `var_sources` exclusive config, and at that point we'll only have to support `ca_cert`.

atc/creds/vault/manager.go
vito: I thought about I think it's actually more likely that we'll end up with different defaults in the long run. It doesn't make much sense to me to have a default path prefix now that we're configuring Vault per-pipeline. So I don't think the maintenance overhead of reflection code is worth the slight duplication, since it'll be temporary in two ways: the defaults may intentionally become different between the two, and we'd probably sooner go exclusively `var_sources` than have them accidentally fall out of sink.

atc/creds/vault/manager_test.go
vito: This could make sense since we have a `concourse generate-key` command, but I'm a little wary of explicitly reinventing (+ documenting and supporting) the wheel on certificate generation because that'll likely be much higher maintenance than private keys. I'd rather people just use `certstrap` since it's explicitly designed around making cert generation easy. We only did `generate-cert` because there were incompatibilities between tooling for generating and using keys across platforms/OSes/Linux distributions/etc.

IMO the general problem of cert generation difficulty actually is made worse by having wrappers hide away the complexity. At some point you're probably going to have to understand how it all works, and then un-learn what all those wrappers did for you, and whether they made the correct choices, and whether their practices were up-to-date.

atc/creds/vault/api_client.go
pivotal-jamie-klassen: yeah, i'll just update our internal wiki with certstrap instructions

atc/creds/vault/manager.go
pivotal-jamie-klassen: i thought about this more today, and you also already have the 'double-entry' effect of the test for the defaults, so why unnecessarily complicate the implementation...

vito:
I've captured both of those in a note in this backlog: https://project.concourse-ci.org/projects/MDc6UHJvamVjdDI5OTk2OTQ=

I'm still stewing up plans on where to go with `var_sources` - i.e. whether we replace the global credential manager config with global `var_sources` in v6.0, since that's already a major version bump, or whether we take more time and maybe do it in v7.0 or something.

Thanks for the review!

PR #4943, opened 2019-12-20
dev docker-compose quality-of-life improvements

vito:
two separate commits with their own explanations

i've been applying these changes locally and got tired of re-doing them over time, so i'm just gonna go ahead and submit this :P


vito:
Interesting. Testflight failed so I re-triggered it, but then it failed again. I think global resources actually breaks one of our testflight tests.

https://nci.concourse-ci.org/builds/137721#L5de94c99:709:724

The breakage probably has more to do with the test setup and maybe pollution than a legitimate regression.

docker-compose.yml
```diff
@@ -29,13 +29,14 @@ services:
       CONCOURSE_ADD_LOCAL_USER: test:test,guest:guest
       CONCOURSE_MAIN_TEAM_LOCAL_USER: test
       CONCOURSE_CLUSTER_NAME: dev
+      CONCOURSE_ENABLE_LIDAR: 'true'
```
cirocosta, line 32: funny to see that this is already outdated :sweat_smile:

cirocosta:
```suggestion

```

cirocosta: fixed during rebase

xtremerui:
@vito the testflight is failing because two resources are sharing the [same resource config](https://github.com/concourse/concourse/blob/85dc27c06bc89450aa802875c1cf2b39fa88d798/testflight/fixtures/caching.yml#L2-L11) when global resource is enabled. 

So the caching test could only fetching one version (cached1 in this case since it is newer) back instead of two
https://github.com/concourse/concourse/blob/85dc27c06bc89450aa802875c1cf2b39fa88d798/testflight/caching_test.go#L21-L22

Make the resource configs different for the two resources will fix the testflight.

vito:
gonna close this - I don't use the `depends_on` change anymore, since `docker-compose up -d --build web` works just as well to avoid restarting `worker`, now that the `Dockerfile` doesn't regenerate the keys each build (it was moved to earlier layers so it can be cached).

enabling global resources would be nice but we can address that whenever we're ready to un-feature-flag it.

PR #4945, opened 2019-12-21
fix(atc): change lidar checks to use resource_config_scope_id as lock id

### LINKED ISSUE #4940: bug(lidar): unexpected serial resource check
## Bug Report

Our users report a check slowness issue, some resource checks took more than 5 minutes, even than 20 minutes.

I looked into the issue, and found the bug is most probably at: https://github.com/concourse/concourse/blob/master/atc/db/check.go#L253:

```go
func (c *check) AcquireTrackingLock(logger lager.Logger) (lock.Lock, bool, error) {
	return c.lockFactory.Acquire(
		logger,
		lock.NewResourceConfigCheckingLockID(c.ResourceConfigID()), // <<==HERE
	)
}
```
Our cluster enables Lidar but global resource. Here `check` use `c.ResourceConfigID()` as lock ID, which causes all checks of those resources having same configure to run in serial order.

See a real number from one of our clusters:

```sql
select count(distinct(pipeline_id)) from resources where resource_config_id=49950
count
------
65
```

There are 65 pipelines having resources with same configure, which means, the 65 pipelines will only run builds in serial.

When I look at those resources' configure, they all use variables like:

```yml
resources:
- name: mygit
  url: xxxxxx
  branch: master
  private-key: ((gitlab-key))
```

Though pipelines are from different teams, and their private keys are different, but as they use same variable name, Concourse would consider them as same. (Note: I didn't actually look through code to figure out if Concourse interpolate variables before save data to `resource_configs`, but I assume Concourse is not able to do so, because if user updates Key in Vault, Concourse has no way to detect the change and update `resource_configs`.)

To quickly solve the check slowness issue, I think we should use `resource_id` as lock id when global resource is not turned on.

A deeper thought: will using variable in resource configure kill `global resource`? Given people often copy pipeline YAML from samples, resource definitions like the above are normal. If multiple pipelines of different teams have the same resource definition, though their `gitlab-key` from above sample are actually different, Concourse may wrongly consider them as same. From the doc https://concourse-ci.org/global-resources.html, `global resource` considers only `type` and `source` configuration, however `team` and `pipeline` should be put into consideration as well, because a variable could be interpolated from either pipeline path or team path from Vault.

@vito I think this is a serious issue, WDYT? If you confirm this is a bug, I might raise a quick fix that would use `resource id` as lock id when `global resource` is disabled, and hopefully the fix can be included in 5.8.0.

## Steps to Reproduce

1. Enable `lidar` and disable `global resource` on a cluster.
2. Create 20 pipelines, each has a git resource with same `source`, the job should `get` the resource with `trigger: true`
3. Push a commit to the git repo
4. Watch the 20 pipelines, jobs will run in serial.

## Expected Results

The 20 pipelines will be trigger to run in parallel. 

## Actual Results

Jobs will run in serial.

## Version Info

* Concourse version: 5.7.2 (previous version should have same issue)
* Deployment type (BOSH/Docker/binary):
* Infrastructure/IaaS:
* Browser (if applicable):
* Did this used to work?

evanchaoli:
# Existing Issue

Fixes #4940 .

# Why do we need this PR?

@vito @pivotal-jwinters If you agree with https://github.com/concourse/concourse/issues/4940#issuecomment-568136099, then the code change is quite nit.

This is NOT a workaround, I believe it should be the expected behavior. The change fits both `global resources` and `non global resources` modes.

Change to use `resource config scope id` as check lock id.

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/check.go
+++ b/atc/db/check.go
@@ -250,7 +250,7 @@ func (c *check) finish(status CheckStatus, checkError error) error {
 func (c *check) AcquireTrackingLock(logger lager.Logger) (lock.Lock, bool, error) {
 	return c.lockFactory.Acquire(
 		logger,
-		lock.NewResourceConfigCheckingLockID(c.ResourceConfigID()),
+		lock.NewResourceConfigCheckingLockID(c.ResourceConfigScopeID()),
 	)
 }
 


vito:
I'm going to close this as this will lead to a lot more containers for folks who don't have global resources enabled, which I think offsets the value of making the `check`s satisfied more quickly.

We're going to prioritize work to not re-use `check` containers in the first place, which should fix this "properly." I'm filling out a [new project](https://project.concourse-ci.org/projects/MDc6UHJvamVjdDM5Nzg3OTk=) for this - we'll get to it shortly after 6.0. I don't think it'll take too long.

Thanks!

evanchaoli:
It's ok to close this PR. Actually I was about to close it. Good to know you will prioritize "not-reuse-check-container".

PR #4948, opened 2019-12-23
bump containerd and its dependencies

cirocosta:
### Why do we need this PR?

containerd:
  - there've been two new versions (1.3.1 and 1.3.2) since the version
    that we were using before (1.3.0)
    - the biggeest update here is the support for runc 1.0.0-rc9, which
      contains mitigations for a CVE [^1]

cni:
  - although we don't use `cni` yet (didn't implement the networking
  bits), it contains a fix for a plugin that we will end up using[^2]

runc:
  - see `containerd` note[^3]

[^1]: https://github.com/containerd/containerd/releases
[^2]: https://github.com/containernetworking/plugins/releases
[^3]: https://github.com/opencontainers/runc/releases

* bump `containerd` in `go.mod`
* bump containerd and deps in `Dockerfile`

### Contributor Checklist

- ~Unit tests~
- ~Updated release notes (located at
  https://github.com/concourse/concourse/tree/master/release-notes)~

### Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


go.mod
```diff
@@ -10,8 +10,7 @@ require (
 	github.com/DataDog/datadog-go v3.2.0+incompatible
 	github.com/DataDog/zstd v1.4.0
 	github.com/Masterminds/squirrel v0.0.0-20190107164353-fa735ea14f09
-	github.com/Microsoft/go-winio v0.4.14 // indirect
```
xtreme-sameer-vohra: Hey @cirocosta 
Looks like we dropped `go-winio` in go.mod, however, in `go.sum` it just appears as though the sha changed. 
Are we missing an update in `go.sum`, where `go-winio` is removed ?

cirocosta: Hey @xtreme-sameer-vohra , I don't think an update is missing as the commit is literally the result of running `go mod tidy`, which is what can give us the full updated list:

```
go mod help tidy
usage: go mod tidy [-v]

Tidy makes sure go.mod matches the source code in the module.
It adds any missing modules necessary to build the current module's
packages and dependencies, and it removes unused modules that
don't provide any relevant packages. It also adds any missing entries
to go.sum and removes any unnecessary ones.
```

now, *why does this show in `go.sum` but not `go.mod`?* I'm not entirely sure.

My guess is that as this comes from `github.com/containerd/containerd`

```
go mod why github.com/Microsoft/go-winio
# github.com/Microsoft/go-winio
github.com/concourse/concourse/worker/backend
github.com/containerd/containerd/cio
github.com/Microsoft/go-winio
```

which is *not* module aware (it uses a `vendor` directory), Go is not keeping track of it as a module dependency, but because `go-winio` itself is a module, which has a SHA associated with it, it lets us verify that we're getting what we expect, even though its direct dependent (`containerd`) is not module aware.

PR #4950, opened 2019-12-27
Don't use concourse tokens

### LINKED ISSUE #2936: Don't use `concourse` tokens.

pivotal-jwinters:
fixes #2936

This PR removes our concept of `concourse` tokens and uses dex tokens directly for anything auth related. This means that we no longer encode all the concourse teams into the token, instead the token will contain all the 3rd party teams from the upstream service, and we do the claims -> concourse team mapping on each request, instead of at login.

One major change as a result of all this, is that the TSA can no longer mint its own tokens. This is probably a good thing for a number of reasons, but it means that concourse got a little more complicated to configure. A lot of that complexity gets hidden away in the `web` cmd, so at least the end user doesn't have to deal with it. They will still need to change some of their configuration, however, so this should be considered a **breaking change**. You can have a look at 6403cfcfabbfb30fc03fef1a49855d844cba611e to see what changes the user will need to make.

So this ended up being a bit bigger than I thought. It might be easier to review this PR one commit at a time. I've tried to leave detailed commit messages that attempts to explain the rationale behind each change.


tsa/tsacmd/command.go
```diff
@@ -39,7 +43,10 @@ type TSACommand struct {
 
 	ATCURLs []flag.URL `long:"atc-url" required:"true" description:"ATC API endpoints to which workers will be registered."`
 
-	SessionSigningKey *flag.PrivateKey `long:"session-signing-key" required:"true" description:"Path to private key to use when signing tokens in reqests to the ATC during registration."`
+	ClientID     string   `long:"client-id" required:"true" description:"Client used to fetch a token from the auth server"`
+	ClientSecret string   `long:"client-secret" required:"true" description:"Client used to fetch a token from the auth server"`
+	TokenURL     flag.URL `long:"token-url" required:"true" description:"Token endpoint of the auth server"`
+	Scopes       []string `long:"scope" description:"Scopes to request from the auth server"`
```
xtreme-sameer-vohra: This is not required as the TSA client only uses the client creds grant type.

skymarshal/dexserver/dexserver.go
```diff
@@ -15,14 +17,15 @@ import (
 )
 
 type DexConfig struct {
-	Logger       lager.Logger
-	IssuerURL    string
-	WebHostURL   string
-	ClientID     string
```
xtreme-sameer-vohra: was this ClientID and ClientSecret was used by skymarshal to talk to Dex ?

atc/api/accessor/accessor.go
```diff
@@ -14,52 +17,158 @@ type Access interface {
...
 }
 
-func (a *access) IsAuthorized(team string) bool {
-	if a.IsAdmin() {
-		return true
+func (a *access) TeamNames() []string {
+
+	teamNames := []string{}
+
+	isAdmin := a.IsAdmin()
```
xtreme-sameer-vohra: could this be cached on the access object ?
When `IsAuthorized` is called and user isn't an admin, this function is called repeatedly.

atc/api/accessor/accessor.go
```diff
@@ -14,52 +17,158 @@ type Access interface {
...
 	}
-	return jwt.MapClaims{}
+
+	return false
 }
 
-func (a *access) IsAuthorized(team string) bool {
-	if a.IsAdmin() {
-		return true
+func (a *access) TeamNames() []string {
```
xtreme-sameer-vohra: Would `AuthorizedTeamNames` be clearer ?

atc/api/accessor/accessor.go
```diff
@@ -14,52 +17,158 @@ type Access interface {
...
+			return true
 		}
-		for _, teamRole := range teamRoles {
-			if a.hasPermission(teamRole) {
-				return true
+	}
+	return false
+}
+
+func (a *access) teamRoles() map[string][]string {
```
xtreme-sameer-vohra: could be renamed to rolesForTeams to match function `rolesForTeam` and also imply its a map

tsa/tsacmd/command.go
pivotal-jwinters: The client credentials grant type still accepts scopes (in the spec), it's just in our case they aren't required by dex. I'd still like keep these here, in case our implementation of the client_credentials grant type in dex needs to change (it hasn't been merged upstream yet).

skymarshal/dexserver/dexserver.go
pivotal-jwinters: Yup we used to only need to configure one client for all of concourse, but now since we need to configure multiple clients (web, tsa). We just pass in a list of clients: https://github.com/concourse/concourse/pull/4950/files/cd34c1ccbeb2bc773aedf7bef60629a6083c7602..02fe4b927dc8cc3290842fac77e61c6bee3bb588#diff-097bd19334894da24754e934c93e4638R25

atc/api/accessor/accessor.go
pivotal-jwinters: Probably, however, the `TeamNames()` method isn't new and would need to be changed across a bunch of handlers, so I'm more inclined to leave it as is.

atc/api/accessor/verifier.go
```diff
@@ -0,0 +1,138 @@
...
+		keySetURL:  keySetURL,
+		audience:   audience,
+	}
+}
+
+type verifier struct {
+	sync.Mutex
+	httpClient  *http.Client
+	keySetURL   *url.URL
+	keySet      jose.JSONWebKeySet
```
xtreme-sameer-vohra: As an optimization, would it make sense to pass this in to the verifier as it could be cached higher up - for now even on bootstrapping as we don't expect this key to change dynamically ?

pivotal-jwinters: There's only ever one `verifier`, which is created in the atccmd: https://github.com/concourse/concourse/pull/4950/files#diff-67f9436106f0941622f0345dbf1c82a2R1675-R1685

fly/commands/login.go
```diff
@@ -172,7 +154,12 @@ func (command *LoginCommand) passwordGrant(client concourse.Client, username, pa
 		return "", "", err
 	}
 
-	return token.TokenType, token.AccessToken, nil
+	idToken, ok := token.Extra("id_token").(string)
+	if !ok {
+		return "", "", errors.New("invalid id_token")
```
xtreme-sameer-vohra: is it also possible that id_token might be missing so this could be `invalid or missing id_token`

atc/atccmd/command.go
```diff
@@ -1518,14 +1577,114 @@ func (cmd *RunCommand) constructHTTPHandler(
...
+	publicKeyPath, _ := url.Parse("/sky/issuer/keys")
+	publicKeyURL := cmd.ExternalURL.URL.ResolveReference(publicKeyPath)
+
+	validClients := []string{"fly"}
+	for clientId, _ := range cmd.Auth.AuthFlags.Clients {
+		validClients = append(validClients, clientId)
+	}
+
+	return accessor.NewVerifier(httpClient, publicKeyURL, validClients)
+}
```
xtreme-sameer-vohra: Could get the verifier to fetch the the `KeySet` here ?

pivotal-jwinters: This is part of the startup path. I'd rather fetch the keyset lazily on first request than block on startup.

atc/api/accessor/verifier.go
pivotal-jwinters: See https://github.com/concourse/concourse/pull/4950#discussion_r372619432

atc/atccmd/command.go
```diff
@@ -712,20 +744,37 @@ func (cmd *RunCommand) constructAPIMembers(
 				externalHost:  cmd.ExternalURL.URL.Host,
 				baseHandler:   authHandler,
 			},
+			tlsRedirectHandler{
```
xtreme-sameer-vohra: nit: cmd.ExternalURL.URL.Hostname() & cmd.ExternalURL.URL.Host can be pulled into vars to reduce visual noise since they are repeated a bunch of times

atc/atccmd/command.go
```diff
@@ -1107,7 +1156,7 @@ func (cmd *RunCommand) oldKey() *encryption.Key {
 	return oldKey
 }
 
-func webHandler(logger lager.Logger) (http.Handler, error) {
+func (cmd *RunCommand) constructWebHandler(logger lager.Logger) (http.Handler, error) {
```
xtreme-sameer-vohra: nit: this doesn't need cmd and can be a static func

atc/atccmd/command.go
```diff
@@ -1518,14 +1577,114 @@ func (cmd *RunCommand) constructHTTPHandler(
...
+				Handler:    webMux,
+				Middleware: middleware,
 			},
 		},
 	}
 
 	return httpHandler
 }
 
+func (cmd *RunCommand) constructLegacyHandler(
```
xtreme-sameer-vohra: nit: this doesn't need cmd and can be a static func

pivotal-jwinters:
There's a lot of personal stylistic nits here. Can we revisit these once we have all the major structural issues resolved?

atc/atccmd/command.go
```diff
@@ -1518,14 +1577,114 @@ func (cmd *RunCommand) constructHTTPHandler(
...
+	issuerURL := cmd.ExternalURL.URL.ResolveReference(issuerPath)
+	redirectURL := cmd.ExternalURL.URL.ResolveReference(redirectPath)
+
+	signingKey, err := loadOrGenerateSigningKey(cmd.Auth.AuthFlags.SigningKey)
+	if err != nil {
+		return nil, err
+	}
+
+	// Add public fly client
+	cmd.Auth.AuthFlags.Clients["fly"] = "Zmx5"
```
xtreme-sameer-vohra: Can we pull this out into constants ? 
There is another ref to the client id below `validClients := []string{"fly"}`

cmd/concourse/concourse_test.go
```diff
@@ -95,6 +95,9 @@ var _ = Describe("Web Command", func() {
 					"--debug-bind-port", strconv.Itoa(8000+GinkgoParallelNode()),
 					"--bind-port", strconv.Itoa(8080+GinkgoParallelNode()),
 					"--tsa-bind-port", strconv.Itoa(2222+GinkgoParallelNode()),
+					"--tsa-client-id", "client-id",
+					"--tsa-client-secret", "client-secret",
+					"--tsa-token-url", "http://localhost/token",
```
xtreme-sameer-vohra: should this be `/sky/issuer/token` ?

atc/api/accessor/verifier.go
xtreme-sameer-vohra: @cirocosta 's suggestion was to leverage something [once](https://golang.org/pkg/sync/#Once) to fetch lazily but only once.

xtreme-sameer-vohra:
> There's a lot of personal stylistic nits here. Can we revisit these once we have all the major structural issues resolved?

Yep, I just wanted to add comments so I don't forget, but they don't need to be resolved for merging. 

Finished looking through the src code. Next step was going to be to review the tests, however, I have a packed day tomorrow so we can ask if Ciro or Izabela can to look at it.

pivotal-jwinters:
@cirocosta ooo ya good catch with `fly userinfo`! Thanks!

Is there a reason you added this snippet?
```
> github.com/concourse/concourse/atc/api/accessor.(*verifier).Verify() ./atc/api/accessor/verifier.go:47 (PC: 0x19cb64b)
    42:	}
    43:
    44:	func (v *verifier) Verify(r *http.Request) (map[string]interface{}, error) {
    45:
    46:		header := r.Header.Get("Authorization")
=>  47:		if header == "" {
    48:			return nil, ErrVerificationNoToken
    49:		}
```

cirocosta:
oh, not at all - just sharing where we added the breakpoint to capture TSA's token

atc/api/accessor/accessor_test.go
```diff
@@ -1,900 +1,649 @@
...
-			It("returns as a system call", func() {
-				Expect(access.UserName()).To(Equal("system"))
+
+			It("returns empty", func() {
+				Expect(result).To(BeEmpty())
 			})
 		})
-		Context("when request has the system claim set as false", func() {
+
+		Context("when the token has a federated user_id", func() {
```
xtreme-sameer-vohra: should be `user_name`

cmd/concourse/web.go
```diff
@@ -72,24 +68,20 @@ func (cmd *WebCommand) Runner(args []string) (ifrit.Runner, error) {
...
 		cmd.TSACommand.ATCURLs = []flag.URL{cmd.RunCommand.DefaultURL()}
 	}
 
+	if cmd.TSACommand.TokenURL.URL == nil {
+		tokenPath, _ := url.Parse("/sky/issuer/token")
+		cmd.TSACommand.TokenURL.URL = cmd.RunCommand.DefaultURL().URL.ResolveReference(tokenPath)
+	}
+
+	cmd.RunCommand.Auth.AuthFlags.Clients[cmd.TSACommand.ClientID] = cmd.TSACommand.ClientSecret
+	cmd.RunCommand.Auth.AuthFlags.Clients[cmd.RunCommand.Server.ClientID] = cmd.RunCommand.Server.ClientSecret
```
xtreme-sameer-vohra: This should probably not happen in the `populateTSAFlagsFromATCFlags` method

atc/api/accessor/accessor_test.go
```diff
@@ -1,900 +1,649 @@
...
+					fakeTeam1.AuthReturns(atc.TeamAuth{
+						"owner": map[string][]string{
+							"users": []string{"some-connector:some-user-id"},
+						},
+					})
+				})
+
+				It("returns true", func() {
+					Expect(result).To(BeTrue())
+				})
```
pivotal-izabela-gomes: Should `pipeline-operator` role be tested when on an admin team also?

atc/api/accessor/accessor_test.go
```diff
@@ -1,900 +1,649 @@
...
-			access = accessorFactory.Create(req, "some-action")
-		})
-		Context("when valid token is set", func() {
-			It("returns true", func() {
-				Expect(access.IsAuthenticated()).To(BeTrue())
+				BeforeEach(func() {
+					fakeTeam1.NameReturns("some-team")
+					fakeTeam1.AuthReturns(atc.TeamAuth{
+						"member": map[string][]string{
+							"users": []string{"some-connector:some-user-id"},
```
pivotal-izabela-gomes: Should we also have tests for actions that require a specific role with TeamAuth using groups rather than users?

atc/api/accessor/verifier.go
xtreme-sameer-vohra: just checking for `ErrVerificationInvalidSignature`

pivotal-jwinters: sounds good!

atc/atccmd/command.go
pivotal-jwinters: done

atc/api/accessor/accessor_test.go
pivotal-jwinters: done

atc/api/accessor/accessor_test.go
pivotal-jwinters: done

atc/api/accessor/accessor_test.go
pivotal-jwinters: done

jomsie:
looks like watsjs test was a flake, re-running

xtremerui:
@jomsie those failures are not flake. The build events are not showing in brower thus breaks watsjs when there are tests assert on build event.

jomsie:
> @jomsie those failures are not flake. The build events are not showing in brower thus breaks watsjs when there are tests assert on build event.

ooh, gotcha, thanks for the update!

pivotal-jamie-klassen:
marking as blocked until v6.0.0 ships.

atc/atccmd/command.go
```diff
@@ -218,6 +229,9 @@ type RunCommand struct {
 	EnableRedactSecrets bool `long:"enable-redact-secrets" description:"Enable redacting secrets in build logs."`
 
 	ConfigRBAC string `long:"config-rbac" description:"Customize RBAC role-action mapping."`
+
+	SystemClaimKey    string   `long:"system-claim-key" default:"sub" description:"The token claim key to use when matching system-claim-values"`
+	SystemClaimValues []string `long:"system-claim-value" default:"ChBjb25jb3Vyc2Utd29ya2VyEgZjbGllbnQ" description:"Configure which token requests should be considered 'system' requests."`
```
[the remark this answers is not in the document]
pivotal-jwinters: What this configuration is saying is any token with a `sub` equal to `ChBjb25jb3Vyc2Utd29ya2VyEgZjbGllbnQ` should be considered a system token.

You should never have to change this value unless you change this:
https://github.com/concourse/concourse/blob/a8bcb4a8bb2f6f195f018e17d53b32eaf2c32919/tsa/tsacmd/command.go#L47

A responsible concourse operator only really needs to change the `CONCOURSE_TSA_CLIENT_SECRET` but we might have clients that want to change `CONCOURSE_TSA_CLIENT_ID` just for some added peace of mind.

So where does this value come from?

A token issued by dex to the tsa using the client credentials grant type looks like this:
```
{
   "iss":"http://localhost:8080/sky/issuer",
   "sub":"ChBjb25jb3Vyc2Utd29ya2VyEgZjbGllbnQ",
   "aud":"concourse-worker",
   "exp":1585664101,
   "iat":1585577701,
   "at_hash":"HucHOAM3G3tzgI-jhm_a-A"
}
```

The `sub` is the single unique identifier for this client/grant type combination. It ends up being a weird value because dex needs to maintain a unique namespace across all connectors so it hashes the user_id and the connector_id to get a sub thats guaranteed to be unique across all connectors. (All client credentials are granted under a `client` connector)

Instead of using the `sub` we could use the `aud` which will always have a human friendly value, however the `aud` doesn't change based on the grant type used to login. So if someone were to login as a user, then the `sub` would be different but the `aud` would always be `concourse-worker`. This might be fine though since this client is only ever use by the tsa for one specific purpose?

docker-compose.yml
```diff
@@ -29,6 +29,8 @@ services:
       CONCOURSE_ADD_LOCAL_USER: test:test,guest:guest
       CONCOURSE_MAIN_TEAM_LOCAL_USER: test
       CONCOURSE_CLUSTER_NAME: dev
+      CONCOURSE_CLIENT_SECRET: Y29uY291cnNlLXdlYgo=
+      CONCOURSE_TSA_CLIENT_SECRET: Y29uY291cnNlLXdvcmtlcgo=
```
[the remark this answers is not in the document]
pivotal-jwinters: Yeah these can be whatever, I only put base64 encoded values so people think of them as secrets.

atc/atccmd/command.go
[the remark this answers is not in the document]
vito: Gotcha, thanks for the background on that - I was wondering why the base64-encoded value looked so strange!

How would you feel about automatically deriving this value based on the configured `ClientID`? Basically just doing the `base64(ClientID+"client")` ourselves. The trade-off is that it's tightly coupled to Dex internals, but we at least wouldn't have to write awkward documentation around it. 🤷‍♂

Come to think of it - if Dex ever changes how this value is created, it would probably be less painful to change our tightly coupled code rather than update the documentation and instruct users to generate the value in the new way. 🤔 

wdyt?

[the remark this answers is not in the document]
pivotal-jwinters: We definitely could. My one hesitation is that I've been slowly trying to inch our configuration to a place where we could theoretically deploy concourse without dex. Like being able to point to an existing OIDC provider. In that world, adding these configurations that let you describe a 'system' request as a claim key/value pair is really useful.

What are your thoughts on changing the defaults to:
`CONCOURSE_SYSTEM_CLAIM_KEY=aud`
`CONCOURSE_SYSTEM_CLAIM_VALUE=concourse-worker`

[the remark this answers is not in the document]
vito: @pivotal-jwinters That sounds more usable, but would that also mean that if someone e.g. logs in via GitHub with a `concourse-worker` username, they would effectively have a system token?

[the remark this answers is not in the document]
pivotal-jwinters: The `aud` claim is always the client_id you use to login. All web logins use a `concourse-web` client so it would never conflict no matter what their user info.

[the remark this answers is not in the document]
vito: Ah OK. If there's no way to spoof it then yeah, that sounds good to me!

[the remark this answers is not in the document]
pivotal-jwinters: The one caveat is that all of this is configureable. So in theory someone could configure the web client and the tsa client to have the same client_id. In that case every token in the system would end up being a system token. Though for this to happen you would have to knowingly change these values, and presumably if you change these values, you know what you're doing.

PR #4954, opened 2019-12-31
nit: fix typo atc/db/component

xtreme-sameer-vohra:
Fix typo in atc/db/component.go


atc/db/component.go
```diff
@@ -58,13 +58,13 @@ func (c *component) Reload() (bool, error) {
 }
 
 func (c *component) IntervalElapsed() bool {
-	inveral, err := time.ParseDuration(c.interval)
+	interval, err := time.ParseDuration(c.interval)
```
kcmannem: good catch

PR #4963, opened 2020-01-03
Add a dumb YAML credential provider

tvon:
### Why do we need this PR?

The goal of this change is to provide a relatively simple way to get started using credentials in pipelines.  It works by specifying the path to a YAML file on the web node(s) with top-level `shared` and `teams` keys, the rest is up to the user.

This PR is known incomplete, I am creating it now to gauge interest.

A sample `credentials.yml`

```
---

shared:
  some_key: "some key value from shared secrets"
  registry:
    username: localuser
    password: localpass

teams:
  main:
    hello-world:
      some_key: "some_key value from team/pipeline"
    some_key: "some_key value from main team"
```

Then just set `CONCOURSE_LOCALFILE_PATH=/src/credentials.yml` in docker-compose and fire it up.

Some technical things to sort out:

* When does it make sense to read the file?  Currently reading it with `Get()` so the file can be updated without restarting concourse. - This should be optimized for performance, IMO.  If credential cache busting is a thing then that would make sense for re-reading the file.
* Should this go so far as encrypting the yaml values? - No, IMO.
* Is it worth puling in a json query package for this? - No, IMO.  Only want a few levels of depth (at most `((name.subkey))`), can sort that out without pulling in another lib.

* add `localfile` credential manager.

### Contributor Checklist
- [ ] Unit tests

### Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -70,6 +70,7 @@ import (
 	_ "github.com/concourse/concourse/atc/creds/credhub"
 	_ "github.com/concourse/concourse/atc/creds/dummy"
 	_ "github.com/concourse/concourse/atc/creds/kubernetes"
+	_ "github.com/concourse/concourse/atc/creds/localfile"
 	_ "github.com/concourse/concourse/atc/creds/secretsmanager"
 	_ "github.com/concourse/concourse/atc/creds/ssm"
 	_ "github.com/concourse/concourse/atc/creds/vault"
--- a/atc/creds/localfile/manager.go
+++ b/atc/creds/localfile/manager.go
@@ -18,25 +18,6 @@ type Manager struct {
 
 func (manager *Manager) Init(logger lager.Logger) error {
 
-	// Just go through the motions to make sure everything can be parsed/read/etc.
-	yamlDoc, err := ioutil.ReadFile(manager.Path)
-	if err != nil {
-		return err
-	}
-
-	jsonDoc, err := yaml.YAMLToJSON(yamlDoc)
-	if err != nil {
-		fmt.Printf("Error converting YAML to JSON: %s\n", err.Error())
-		return err
-	}
-
-	var someStruct map[string]interface{}
-	err = json.Unmarshal(jsonDoc, &someStruct)
-	if err != nil {
-		fmt.Printf("Error unmarshaling JSON: %s\n", err.Error())
-		return err
-	}
-
 	manager.SecretsFactory = &SecretsFactory{
 		path:   manager.Path,
 		logger: logger,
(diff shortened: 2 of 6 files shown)


tvon:
Hah, I somehow completely missed the dummy credentials manager.  For my current setup I just need to get a few secrets in there for the sake of experimenting with and demoing concourse against some private services.  I am not too concerned about the hot-reload, I just coded it that way out of convenience.

If we start using concourse for real then we would start with 15-20 secrets to store, including some certificates.  That sounds like it would be awkward on the CLI vs a yaml file (when I think about looping over a dict in ansible to generate the command).

So, this is not as immediately important to me now as it was yesterday.  I do like the idea of using a file vs CLI flags and can clean this up if there is interest.

pivotal-jamie-klassen:
Regardless of how this feature goes, I think your use-case (POCs to demo/pitch Concourse) is a fairly important archetype. I added an issue to the docs to track this.

When I think about maintenance costs, I feel obliged to suggest that either the config file provider or the dummy provider should exist, but not both. I'm realizing that they optimize for slightly different concerns, though.

One advantage I can imagine (though I'm not sure we're relying on it very heavily in our integration tests) with the dummy provider is testing. You can define some credentials statically and have no need to integrate with the filesystem to expect them to work; it's really nice and deterministic (no problem if your disk dies/hangs during the test run or if the file's contents change from underneath you). This is cool, but if it's the only justification it starts to smell a lot like [test damage](https://www.thoughtworks.com/insights/blog/test-induced-design-damage-fallacy-or-reality).

The other use case where dummy makes sense is in a pipeline-scoped [`var_source`](https://github.com/concourse/concourse/pull/4600). I think if I used a config file for this, I might have all these kinda "ephemeral" credential managers that need disk access -- potentially each needs access to its own file? Sounds a bit tricky.

On the other hand, the feature in this PR is pretty focused around a nice operator experience for this "demoing credential management" user narrative. This is a valuable property for a feature to have.

Finally I want to mention that maybe I'm over-inflating the maintenance cost of having both credential manager types, and supporting both isn't such a big deal.

tvon:
I really appreciate the thoughtful feedback, especially considering how ham-fisted my code is.

I might not be clear on how `var_sources` works, would that mean there could be multiple yaml-credential files configured with different scopes?  Would they all be configured on the web node(s)?  I can see how it might get out of hand, though I'm not sure how bad it really is (but then I don't know the concourse internals well).

A question about the dummy credential manager, how would one deal with a multi-line value like an SSH private key?

I do plan on getting back to this, I've been able to move along with the dummy store for now though.

pivotal-jamie-klassen:
> there could be multiple yaml-credential files configured with different scopes? Would they all be configured on the web node(s)?

> I don't see how that could realistically work with a yaml file unless you only specify some identifier for pre-configured file store.

Yep, these are the sorts of concerns I'm imagining too. Especially across multiple web nodes, the notion starts to get painful.

> how would one deal with a multi-line value like an SSH private key?

If you're using dummy as the cluster-wide credential manager and setting environment variables directly (you mentioned using ansible above so you might be able to achieve something a bit cleaner), I think you'd need to escape the newlines.

If you're configuring dummy as a var source, a bit like the example on https://concourse-ci.org/runtime-vars.html#var-sources, then you could just use ['block style'](https://yaml.org/spec/1.2/spec.html#id2795688) indicator for a multiline YAML string value.

Hope this helps!

tvon:
Closing this given the above discussion, thank you for your feedback.

PR #4972, opened 2020-01-06
add version column back to resource_caches

xtremerui:
to avoid trouble when running down migration if version can not be found for given version_md5

it will be used only when inserting.


xtremerui:
Since the testflight failure is irrelative I will merge this PR

Issue #4870, opened 2019-12-09
Enable "global resources" by default


vito:
We can't safely enable "global resources" until we have a plan for resource types that use things like IAM for credential management, so this depends on https://github.com/concourse/concourse/issues/3023

vito:
We also can't do this until we have a way for the `time` resource to not trigger every single `1m` interval at once. So that's blocked on something like 'trigger resources' (https://github.com/concourse/rfcs/pull/27) which don't collect a common history.

Issue #3023, opened 2019-01-10
Support for EC2 IAM roles in a way that allows us to safely share version history


vito:
Prerequisites:

* [ ] https://github.com/concourse/concourse/issues/5229

## What challenge are you facing?

With #2386 we started on a bunch of work to reduce the footprint of resources by sharing `check` containers and version history globally for equivalent resource definitions.

Resource definitions are considered equivalent if their `source:` (interpolated with credentials and  hashed for safety) and type are equivalent.

However, as was uncovered in #3002, there is one situation where the hashed interpolated source is not enough to determine whether version history should be shared: resources using IAM roles. These resources forego putting credentials in `source:` in favor of using EC2-configured IAM roles to grant  anything that runs on the workers access to the AWS resource automatically.

This has pretty scary implications on version history sharing. Because the `source:` does not contain the credentials, all it would take is one person that *does* have access (via their own workers) to successfully check, and then anyone else could configure the same `source:` and see the same version history without even having to configure the IAM roles.

Thankfully, they at least won't have access to the fetched bits. The `get` step would have to run on their own workers, which wouldn't have access, and so there would be no cache to re-use and no ability to fetch the bits. However, this is still a dangerous information leak.

## What would make this better?

I'm not sure! But the way resources use IAM roles today flies in the face of a couple Concourse anti-patterns:

* `anti/worker-state`: because the operator is configuring the workers specifically for particular workloads. This could backfire if they start to use those same workers to run un-trusted code like pull requests. Running on a worker should not automatically grant access to sensitive data!
* `anti/multi-source-of-truth` with a hint of `anti/contributor-burden`: each resource that deals with AWS now needs to support two ways of being configured: IAM roles and static configuration.

Is there some way we can make this relationship with IAM roles more explicit?

For example, instead of configuring the workers, could the operator configure the ATC with named IAM roles and explicitly permit certain teams to access those IAM roles, perhaps leveraging our existing credential manager support?

To be honest, my understanding of IAM roles is fairly loose, but as long as they can be named, we should be able to support this even in a multi-tenant environment, by configuring the ATC with something like this:

```
concourse web \
  --iam-secrets-team-roles team-name:role1
```

Then, assuming that the operator has already configured the `web` EC2 to have those named roles, pipelines belonging to `team-name` could use this credential manager like so:

```yaml
resources:
- name: my-bucket
  type: s3
  source:
    access_key_id: ((role1.access_key_id))
    secret_access_key: ((role1.secret_access_key))
[...]

Issue #2386, opened 2018-07-16
Equivalent resources defined across pipelines and teams should only correspond to a single version history


vito:
## What challenge are you facing?

Today, pipeline resource versions are collected in to a `versioned_resources` table. This table was predated the "life" epic (#629). It contains the following schema:

```
atc=# \d versioned_resources
                               Table "public.versioned_resources"
   Column    |  Type   | Collation | Nullable |                     Default                     
-------------+---------+-----------+----------+-------------------------------------------------
 id          | integer |           | not null | nextval('versioned_resources_id_seq'::regclass)
 version     | text    |           | not null | 
 metadata    | text    |           | not null | 
 type        | text    |           | not null | 
 enabled     | boolean |           | not null | true
 resource_id | integer |           |          | 
 check_order | integer |           | not null | 0
```

It points to `resource_id`, making this per-pipeline-resource. This means that multiple pipelines with the same resource configs will be redundantly collecting the same version/metadata information.

## A Modest Proposal

To be honest, this isn't a *huge* deal right now, aside from wasted database space and redundant checking. However if we make the relationship between a pipeline's resources and the abstract version history a bit tighter, there are actually a few benefits:

* We can reduce the amount of `check`ing required across pipelines for equivalent resources.
* We can reduce the amount of data recorded for equivalent resources.
* When a user changes their pipeline resource's configuration, the history will be "re-set" (ref. #145) to the new config, and should always be correct.
* There may be some as-yet-unknown improvements we can make to the database model by having a cleaner representation.
* As part of concourse/rfcs#1, we're going to start collecting *all* versions, not just starting from pipeline configuration time. There'll be a lot more data to record, so sharing it between duplicated resources will make things a lot more efficient.

### Enabling/Disabling versions

Enabling/disabling versions should remain scoped to pipeline resources, obviously. This can be done via a join table (`pipeline_resource_config_versions` or some such).

### Distinct check intervals

Now that we only check once per resource config, there's a little gotcha. Different pipelines can have varying `check_every` settings.

Here's one idea: record `last_checked` on the resource config, and have each pipeline's radar component just check if the `last_checked` is >= their interval. So, we'll check at the fastest defined frequency. Pipelines with longer `check_intervals` will have versions show up more quickly than expected, but that really shouldn't matter.

### Pausing pipeline resources

Currently, users pause pipeline resources with the intended effect that no new versions are collected and used for later builds. This is really awkward when other pipelines result in checking the config anyway.

[...]

PR #4973, opened 2020-01-07
atc(feat): var step

evanchaoli:
# Existing Issue

This PR implements RFC https://github.com/concourse/rfcs/pull/42.

# Why do we need this PR?

This PR adds a feature that simplify the way of passing data between steps.

My test pipeline:

```yaml
resource_types:
- name: rt-tiny
  type: registry-image
  source:
    repository: evanli/tiny2
    tag: v4

resources:
- name: tiny
  type: rt-tiny
  source:
    key: foobar

- name: tiny2
  type: rt-tiny
  source:
    key: foobar2

- name: data
  type: mock
  source:
    create_files:
      data.yml: |
        format: yml
      data.json: |
        {"format": "json"}

jobs:
- name: test
  public: true
  plan:
  - get: tiny # get tiny generates a file "a.txt" containing a single line "hello"
  - get: data

  - task: a-task
    input_vars:
    - name: tv
      file: tiny/a.txt
    - name: ty
      file: data/data.yml
    - name: tj
      file: data/data.json
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: sh
        args:
         - -exc
         - |
           echo tiny shows ((tv:value))
           echo data contains ((ty:format)) and ((tj:format))

  - get: tiny2
    input_vars:
    - name: tv
      file: tiny/a.txt
    params:
      data: ((tv:value))

  - put: tiny2
    input_vars:
    - name: tv
      file: tiny/a.txt
    params:
      path: ((tv:value))
```

Screen shot of a build:

input_vars_demo

This test shows that:

* `task`, `get` and `put` all work with `input_vars`
* `input_vars` takes plain text file, yaml file and json file. 

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


evanchaoli:
@vito Can you please take an initial look at the code change? If you don't have concern from the design perspective, I'll add unit tests for `get/put/task_step` and a test flight case.

vito:
Hey, sorry, still coming back online and catching up from the holidays. :slightly_smiling_face: 

I'm in a bit of analysis paralysis for the moment regarding a few var-related things, and this is kind of trapped in there too as a result. Basically I'm starting to think of `var_sources:` being implemented as prototypes, as another use case for the protocol.

There are a few motivators for this:

1. Running credential managers as prototypes on the *workers* makes a LOT of sense for things like acquiring EC2 IAM roles (https://github.com/concourse/concourse/issues/3023), which is a blocker for #4870.
1. Using prototypes would allow for the credential manager implementations to be developed, maintained, and upgraded independently of Concourse core. This alone would be a huge win and gives users a lot more power and flexibility.
1. The [`across` step](https://github.com/vito/rfcs/blob/spatial-resources/029-across-step/proposal.md) proposal is based the idea of `var_sources` being implemented as prototypes which *aren't* necessarily for credential management, instead using it to implement build matrixes - basically operating on "sets" of vars.
1. I'm also considering using `var_sources` as the approach for [trigger-only resources](https://github.com/concourse/rfcs/pull/27), by adding a `var:` step which points to a var source and uses it for triggering. I'm still drafting the RFC for this, but I'll try to get something out soon so you're not stuck with radio silence.

Building on that last motivator, I started to think about your `input_vars` RFC and see if the same thing could be represented by using this same `var:` step but load the var's value from a local file instead of a var source.

Your example would then look something like this:

```yml
jobs:
- name:
  plan:
  - get: tiny # get tiny generates a file "a.txt" containing a single line "hello"
  - get: data

  - var: tv
    file: tiny/a.txt
  - var: ty
    file: data/data.yml
  - var: tj
    file: data/data.json

  - task: a-task
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: sh
        args:
         - -exc
         - |
           echo tiny shows ((tv:value))
           echo data contains ((ty:format)) and ((tj:format))
```

But now the challenge would be how to safely and securely use prototypes for credential management,  the original use case for `var_sources`. This is an interesting technical challenge to approach cautiously. At this point I kind of took a step back to focus on the 6.0 beta. :sweat_smile: Apologies that this has kind of stepped on the toes of your proposal and dragged it out a bit. I would appreciate your feedback and questions on any of this!

evanchaoli:
@vito I believe you stand at a higher level to oversee that Concourse evolute alone a right direction, for which, I fully trust you, while I mostly focus on solving immediate pain points. So my major concern is that, a lot of things in your view (RFCs), like prototype, have no ETA yet, I really don't want them to block features like `input_vars`, as the feature is small but benefit a lot. I have received plenty of complains that users have to write additional code to parse files passed from previous steps. Especially for resource type authors, when they add a param `x`, they usually need to add an extra param `x_file`, which makes code redundant and hard to maintain.

I agree with motivators 1 and 2, only concern is users don't want to write credentials to disk. Also, I assume that current system-level credential manager should still be supported for backward compatibility.

For `across` step, it's good to add `for` loop function, some users asked for that to me. I think maybe we should only start from static value list. Using `var_source` with `across` will bring a lot of issues, so that leads `across` to be over complex and confused. IMO, an important sell-point of Concourse over competitor tools is "simple". The three basic steps, `get`, `put` and `task`, are easy to understand and use.

I kinda don't like `trigger-only` resource. The original request is just to add a `trigger_only: true` flag to skip executing `get`. I think that makes sense than `trigger-only` resource in certain use cases. For example, we have similar use cases: say a pipeline has a `git` resource, `job1` runs a build process triggered by the `git` resource, and `job2` runs test, it also triggered by the `git` resource and `passed: job1`, `job2` doesn't really need to fetch data from `git`. In other words, the `git` resource should do real `get` for `job1`, and be `trigger-only` for `job2`. For this kind of use case, adding `trigger_only: true` to `get_step` should be the simplest solution.

Adding a `var` step will also bring some problems to consider:

* Should `var` steps be order sensitive?
* Can a `var` refer to a previous `var`?
* Can a step (`get/put/task`) refer to a `var` defined after it?
* How to display `var` step on the web UI?
* ...

With my original design of https://github.com/concourse/rfcs/pull/42, `input_vars` is a simple feature, it's just a variant of existing `inputs`. But when you put it into the big picture, it becomes to something I suddenly cannot see when it can be really in place. Hmm, I don't know.

In summary, to me, I want to merge my current implementation as is, because looks like to me, the big picture still take time to reach, and I don't think `input_vars` will conflict to the bigger picture. For example, once `var` step is added, you may want to deprecate `input`. At that time, deprecate `input_vars` together with `input` should not a be a problem.

vito:
> I believe you stand at a higher level to oversee that Concourse evolute alone a right direction, for which, I fully trust you, while I mostly focus on solving immediate pain points. So my major concern is that, a lot of things in your view (RFCs), like prototype, have no ETA yet, I really don't want them to block features like `input_vars`, as the feature is small but benefit a lot. I have received plenty of complains that users have to write additional code to parse files passed from previous steps. Especially for resource type authors, when they add a param `x`, they usually need to add an extra param `x_file`, which makes code redundant and hard to maintain.

Makes sense! I totally understand the need, and I'm not trying to block it, I'm trying to reach consensus in order to avoid churn. Implementing this quickly and deprecating it later would probably be OK for your team, but it seems wise to avoid unnecessary churn for Concourse users, if it is indeed avoidable.

> I agree with motivators 1 and 2, only concern is users don't want to write credentials to disk. Also, I assume that current system-level credential manager should still be supported for backward compatibility.

100% agreed; that's the main technical challenge to me, and it might involve a change to the prototype interface. 🤔 

> For `across` step, it's good to add `for` loop function, some users asked for that to me. I think maybe we should only start from static value list. Using `var_source` with `across` will bring a lot of issues, so that leads `across` to be over complex and confused. IMO, an important sell-point of Concourse over competitor tools is "simple". The three basic steps, `get`, `put` and `task`, are easy to understand and use.

The `across` step was actually introduced with the sole purpose being to iterate over dynamic vars. This is how pipelines for branches and PRs will be dynamically configured. I definitely agree that static iteration is easiest to start with, but doing that on its own isn't enough for the v10 roadmap.

> I kinda don't like `trigger-only` resource. The original request is just to add a `trigger_only: true` flag to skip executing `get`. I think that makes sense than `trigger-only` resource in certain use cases. For example, we have similar use cases: say a pipeline has a `git` resource, `job1` runs a build process triggered by the `git` resource, and `job2` runs test, it also triggered by the `git` resource and `passed: job1`, `job2` doesn't really need to fetch data from `git`. In other words, the `git` resource should do real `get` for `job1`, and be `trigger-only` for `job2`. For this kind of use case, adding `trigger_only: true` to `get_step` should be the simplest solution.

The problem goes deeper than that. The `time` resource is an example of a trigger-only resource. However it has two problems which are much worse than the redundant fetching: 1. it's represented as a resource with an infinitely growing history, leading to more and more database utilization and scheduling overhead, 2. it triggers all downstream jobs at the same time, as they all share the same version history, leading to spiky load on workers. This is exacerbated by #2386, which is feature-flagged until we figure this out, so that a '10m' interval doesn't fire across the entire cluster at once.

So, my main goal is in solving that, not just skipping the fetching - that part's easy, and I don't really want to implement it in a way that sprinkles flags all over `get` steps. It feels weird to say "get it, but actually, don't get it" - from a declarative config standpoint.

> In summary, to me, I want to merge my current implementation as is, because looks like to me, the big picture still take time to reach, and I don't think `input_vars` will conflict to the bigger picture. For example, once `var` step is added, you may want to deprecate `input`. At that time, deprecate `input_vars` together with `input` should not a be a problem.

That's fine - we can merge it in and deprecate it later, but I was hoping to be able to come to a consensus before forcing so much churn upon our users. Changes to the pipeline config aren't something to take lightly - they're the single thing that guarantees a major version bump if there's a breaking change (like removing it after deprecating it). We could just leave it deprecated for a while, but it'd be better to avoid the whole problem.

These questions have pretty simple answers to me:

> * Should `var` steps be order sensitive?

Yes, as all steps are. They execute in the order they're written. A `var` step would just run and set its var.

> * Can a `var` refer to a previous `var`?

Don't know what this would mean. It'd only have a `file:` field really. If the step interpolates it, then sure, but I don't think there's precedent for that in other `file:` fields.

> * Can a step (`get/put/task`) refer to a `var` defined after it?

Nope; the step has to run first.

> * How to display `var` step on the web UI?

Just like the other steps.

---

I'm not proposing that we wait on prototypes, `var_sources`, etc. I'm seeing if we can land on a design that we can build on in the future to avoid breaking people's pipelines when we pick a different direction. For example, if we can agree on `var:/file:`, that's something that could be easily implemented *today*, and then *later* extended to collect vars from a var source, backwards-compatibly.

evanchaoli:
@vito Good to see that we are aligned on most of parts. We can continue discussing `trigger-only`  resource in its RFC. Now focus back on `input_vars`.

Several things from my mind are blocking me from agreeing on `var:/file`:

* As `var_source` is order insensitive, if we define `var` as order sensitive, it will look a little bit confused, though we can argue, steps within a job are order sensitive.
* I don't see what a `var` step can do. For a `var` step, I don't think we need to launch a container and run something in it. With a quick thought, even if go along with `var:/file`, the implementation could be like:
  * Get a list of `var` steps defined before the current step.
  * Handle the list of `var`s exactly same as my current code of handling `input_vars`.

So that a `var` step doesn't look like a real step, I don't know what can be shown on GUI for a var step. Maybe only thing we can do is to verify: if a given `file` exists and is valid. 

What do you think?

vito:
> * As `var_source` is order insensitive, if we define `var` as order sensitive, it will look a little bit confused, though we can argue, steps within a job are order sensitive.

Sorry, I think I don't understand what you mean by order-sensitive here. :thinking:

A `var` step would load the var's value from `file:`, optionally parse its YAML/JSON value, and set it in build-local vars under the step's given name. The var would then be available to subsequent steps.

We don't have a mechanism for build-local vars yet, but we'll need something like it for the `across` step. For the `var:` step, we could change the [`atc.BuildStepDelegate`](https://github.com/concourse/concourse/blob/95cf99207e03490c453c1718cd870f8eeff0776e/atc/exec/build_step_delegate.go#L13) to have a `.Variables()` that allows setting vars in an internal `vars.StaticVariables` that takes precedence over the rest. This way we don't have to change all the other step implementations - they'll automatically pick up the changes through their delegate.

Like the `set_pipeline` step, the `var` step with `file:` doesn't do anything with containers, but it still does something that can take time, fail, error, or succeed. It has to stream the file's contents from the worker, then optionally parse it, and then set the value (which, granted, can't fail unless we disallow shadowing or something). The worker could be slow or unreachable, the file could not exist, or the contents could be invalid YAML/JSON. So having the step show up in the UI makes sense to me. I would expect to see the same progress and error reporting semantics as the other steps.

evanchaoli:
input_vars2

@vito I just updated my implementation to do add a `var` step. As I just added the `set_pipeline` step, it's quite straightforward to add a new step of `var`, including UI changes. With the new `var` step, `get/put/task_step` code is not touched at all.

As `credTracker` lives across entire life of a build, it seems the best place to hold build-local vars. In `var` step, we just need to merge local vars into `credTracker`.

So the key point turns to be how to merge `var`s into `credTracker`. Given we should have a single `NamedVariables` object to hold all var sources, so I changed `vars.credVarsTracker` to hold a unNamed variables for global credential manager and a named varaible for var sources.  So that I can add a new function `CredVarsTracker.AddNamedVariables(string, Variables)` that can merge build-local var into `credTracker`.

Please see if you like this design. If you are ok with the design, I'll add unit tests and run some regression tests.

vito:
Ahh woops I just realized my example before wasn't quite what I meant. I forgot to update the `((var))` usage.

Thanks for trying out the new approach though, it's mostly there. :slightly_smiling_face: I really appreciate you taking the time to iterate on this as we work out the details!

Whereas your `input_vars` loads the files as named var sources, the `var` step that I meant to demonstrate sets them as anonymous local vars, not in a named var source. I don't really have an opinion about this yet, so I'm just going to think out loud here for a moment if you don't mind. :sweat_smile: 

Here's an updated example of what I was thinking:

```yaml
jobs:
- name:
  plan:
  - get: tiny # get tiny generates a file "a.txt" containing a single line "hello"
  - get: data

  - var: tv
    file: tiny/a.txt
  - var: ty
    file: data/data.yml
  - var: tj
    file: data/data.json

  - task: a-task
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: sh
        args:
         - -exc
         - |
           echo tiny shows ((tv))
           echo data contains ((ty.format)) and ((tj.format))
```

The only difference between this example and the earlier one is that `((tv:value))` becomes `((tv))`, `((ty:format))` becomes `((ty.format))`, and `((tj:format))` becomes `((tj.format))`.

To me the usage of the `var` step implies that it sets a single var, not a set of named vars. Otherwise we would call the step `vars`, but it would feel strange to design for the plural form since we know that there'd often be just a single value, and all other steps are singular.

I don't think you really gain/lose much by choosing one or another, because `((foo:bar.baz))` functions the same as be `((foo.bar.baz))`. The one mechanical difference is that by having it be a named var source you have to come up with a var name for single-value things, hence `((tv:value))` as opposed to `((tv))`.

One thing I *don't* like about using anonymous `((var))` syntax is that it's ambiguous with statically-configured `((vars))`. Take this example:

```yaml
jobs:
- name: blah-((name))
  plan:
  - get: foo
  - var: name
    file: foo/foo.txt
  - put: bar
    params:
      tag: ((name))
```

If I were to run `fly set-pipeline -v name=hello`, this would mistakenly set the build-local `((name))` at the same time. :thinking: Maybe we could prevent this through static validation, but it'd be kind of tricky since you'd need to perform this validation before any vars are interpolated - right now validation happens *after*.

Another option to think about may be to put these vars in a `build` or `self` variable scope or something, so you'd do `((build:tv))` and `((build:ty.format))`. I'm not super fond of this approach because it invents a magic keyword, and overloads `((foo:bar))` syntax to mean more than just reading from a named var source.

Another thing to keep in mind: whatever decision we make here would likely apply to the `across` step too, since it also sets build-local vars.

I wonder if we need a third syntax so that they're all completely unambiguous: `((foo))` would always be static, `((named:foo))` would always reference a var source, and some third syntax would be used for build-local variables. :thinking: 

I don't have a strong opinion here yet but I hope this helps you reason through it with me. I'll stop here so we can think about it some more. Thanks again for working through this with me!

evanchaoli:
@vito Ah, I never considered you wanted `ty.format`, as I always thought local vars (input_vars of my proposal) as named vars. I thought you picked `var` rather than `vars` because `vars` has been used as an attribute of `task`. With `var`, looks like you consider `ty.key1` and `ty.key2` as a single var, but to me, I always considered them as two. So I think `var` is an ok name, users should not get confused. Just like, `var_sources`, a var source may offer multiple vars, but we didn't pick up a name of `vars_sources`.

I still prefer to use named var for build-local vars. Because:

* It allows `across` to refer to a build-local var in same way as a var source.
* It allows a local var to overwrite var_source. (Note: I just meant a possibility, am not going to implement that in this PR.) As most of programming languages allow local variable to overwrite global variable, and pipeline authors are usually programmers, so I suppose pipeline author would not be surprised with local var overwriting var source.

Actually, TBH, I'm a little surprised with the syntax of `ty.format`. As I added `var_source`, from my mind, input_vars to be as named is very nature. In my mind, vars are just two types: named vars (var_source and build-local), and un-named (static vars and global credential managers).

vito:
> I thought you picked `var` rather than `vars` because `vars` has been used as an attribute of `task`.

Haha, didn't realize that. :sweat_smile: 

> With `var`, looks like you consider `ty.key1` and `ty.key2` as a single var, but to me, I always considered them as two. So I think var is an ok name, users should not get confused. Just like, `var_sources`, a var source may offer multiple vars, but we didn't pick up a name of `vars_sources`.

In `((foo:bar.baz))`, `foo` is a var source, `bar` is a var, and `baz` is just a field of the var. A var source is often a source of many vars, but its use case is to fetch one var at a time, as they're typically distinct from each other - hence it's a var source. I would only call it a "vars source" if it was being used to return sets of vars together or something.

`((ty.key1))` and `((ty.key2))` actually has to be thought of as a single var `ty` because in cases where the fields are generated, e.g. Vault's [AWS secrets engine](https://www.vaultproject.io/docs/secrets/aws/index.html), we have to fetch `ty` once, not twice, otherwise the access key and secret key won't work with each other.

The var syntax is documented here now btw, I backfilled a bunch of `((var))` docs upon merging `var_sources`: https://concourse-ci.org/vars.html

> Actually, TBH, I'm a little surprised with the syntax of `ty.format`. As I added `var_source`, from my mind, input_vars to be as named is very nature. In my mind, vars are just two types: named vars (var_source and build-local), and un-named (static vars and global credential managers).

Yeah this is the other way I tried to reason about it. Basically `((foo:bar))` always means it'll be evaluated at runtime. Right now that distinction doesn't hold true for `((foo))` which is either static or a global credential manager var, but if we're phasing out the global credential manager then it could make sense.

I still don't really like that we would end up with `((foo:value))` for simple values though, and I think having `foo` be either a var source or a singular build-local var is a bit confusing.

`((foo:bar.baz))` means "fetch `bar` from the `foo` source and read the `baz` field off of it."

A `var: user` step means "fetch a value from a local file and provide it as a local `user` var." Just as with normal vars, the content may be a simple value or a complex value. Assuming it's a simple value, I would expect to just use some syntax like `((user))`, as I would with other simple vars, whether it's static or from a credential manager. If it's complex, with `username` and `password` fields, I would expect to use `((user.username))` to just read the field of the var which has already been fetched.

So what confuses me about `((user:username))` is that if I attempt to build on my existing understanding of `foo:bar` syntax from var sources, it implies we need to fetch the `username` var from the `user` source. But in reality those values have already been fetched, which is different from how the syntax usually works. The actual behavior is much closer to just reading fields off of a fetched var.

I don't really see the benefits of allowing build-local vars to shadow var sources. With them behaving so differently that seems like it would almost always be a mistake. :confused: 

Maybe we actually just have different plans for how we expect to use this. To me, the `var` step would only be used to read one conceptual var where all the fields are related, e.g. a username/password, properties of single environment, or a single value. If it's being used to load up a bunch of arbitrary vars, I can see why you would expect that to act more like a var source, but I don't think that's really the core case we should be designing around - we usually prefer to design steps around the simplest case, i.e. the singular form - hence `set_pipeline` and not `set_pipelines`, even though folks often set many pipelines at once.

evanchaoli:
> `((ty.key1))` and `((ty.key2))` actually has to be thought of as a single var `ty` because in cases where the fields are generated, e.g. Vault's [AWS secrets engine](https://www.vaultproject.io/docs/secrets/aws/index.html), we have to fetch `ty` once, not twice, otherwise the access key and secret key won't work with each other.

An opposite sample is Vault. Say we have two keys under same path: `/concourse/ty/key1` and `/concourse/ty/key2`. Concourse implementation fetch `ty` once, but as a user, I define `key1` and `key2` in Vault as two different keys, I don't care if they are fetched in batch of individually at all. So I think user experience is a different thing than implementation.

My original purpose of `input_vars` is just a simplified form of `inputs`. We cannot what content a task may write to an output file, right? It could be a very complex JSON. Without `input_vars` or `var` step, the following steps have to write some code to read the file and parse JSON, which causes redundant code. Thus, to me, an input file is a "var source", it could hold a single key as `value`, or very complex bunches of keys.

If you feel `var` is a lit confused with complex JSON, how about rename to `var_source` that perfectly set user's expectation:

```yaml
jobs:
- name:
  plan:
  - get: tiny # get tiny generates a file "a.txt" containing a single line "hello"
  - get: data

  - var_source: tv
    file: tiny/a.txt
  - var_source: ty
    file: data/data.yml
  - var_source: tj
    file: data/data.json

  - task: a-task
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: sh
        args:
         - -exc
         - |
           echo tiny shows ((tv:value))
           echo data contains ((ty:format)) and ((tj:format))
```

Or `build_vars`, or `build_var_source`:

```yaml
  - build_vars: tv
    file: tiny/a.txt
  - build_var: ty
    file: data/data.yml
  - build_var: tj
    file: data/data.json
```

I'm slightly keen on `build_vars`, that implies those vars are build local, and might contain multiple vars. 

For `var_source`, my only worry is user might expect to define a build Vault given `var_source` is too similar to `var_sources`. But if that is a valid use case, it's easy to implement.

For `((foo:value))` that you don't like, I just feel ok. If you really/strongly don't like it, we may consider to not support plain text input file, so that `((foo:value))` is avoided. The reason I designed `((foo:value))` is to compatible with existing resource types and tasks. For example, current `git-resource`'s `get` writes a set of files, each of them contains a single information, like sha, committer. Once this feature is in-place, `git-resource` may change to write a single JSON file. So we may eventually deprecate `((foo:value))`. If we don't support `((foo:value))` from the beginning, that may force resource type authors and pipeline authors to speed up switching from plain text output files to JSON files.

vito:
> An opposite sample is Vault. Say we have two keys under same path: `/concourse/ty/key1` and `/concourse/ty/key2`. Concourse implementation fetch `ty` once, but as a user, I define `key1` and `key2` in Vault as two different keys, I don't care if they are fetched in batch of individually at all. So I think user experience is a different thing than implementation.

I'm confused by this. :thinking: Concourse doesn't fetch `ty` once, and it wouldn't in this case because that's not a key, it's just a part of a path. If you have credentials stored as `/concourse/ty/key1` and `/concourse/ty/key2` those are different keys, and they will be fetched separately, with syntax like `((ty/key1))` and `((ty/key2))`.

With the AWS secrets engine you have to do a single `read` call to fetch a credential that has `access_key` and `secret_key` fields. For [this example](https://www.vaultproject.io/docs/secrets/aws/index.html#usage) those fields would be read as `((aws/creds/my-role.access_key))` and `((aws/creds/my-role.secret_key))`, in which case Concourse has to know to fetch `aws/creds/my-role` a single time per some sort of evaluation scope (possibly per-build?) so that the credentials match each other.

This is not currently implemented; we haven't had to because it's currently impossible to use any secrets engine except for generic/kv, because of our strict credential path rules. Now that we're able to relax that with `var_sources`, we can implement this properly.

> For `((foo:value))` that you don't like, I just feel ok. If you really/strongly don't like it, we may consider to not support plain text input file, so that `((foo:value))` is avoided. The reason I designed `((foo:value))` is to compatible with existing resource types and tasks. For example, current `git-resource`'s `get` writes a set of files, each of them contains a single information, like sha, committer. Once this feature is in-place, `git-resource` may change to write a single JSON file. So we may eventually deprecate `((foo:value))`. If we don't support `((foo:value))` from the beginning, that may force resource type authors and pipeline authors to speed up switching from plain text output files to JSON files.

Ah I see, so you were mainly interested in the complex values case but added `:value` for backwards compatibility. Well, I guess that depends on what direction we want resource type authors to take.

Right now there's a kind of simplicity in writing single values to files. Scripts can easily consume them by just `cat`ing them into a var. I think that's a good thing; sometimes you're just writing a task that's just "CI glue" and it can be annoying to have to bake something like `jq` (as amazing as it is) into all of your CI images.

My gut feeling is that as long as information is being passed through the filesystem it's better for it to just be in simple files, as it assumes as little as possible about what is going to be consuming them. But when it comes to protocols and higher-level automation JSON makes a lot more sense.

I think we can avoid the `var` step entirely for reading values from fetched resources if we have a way for the [Resource Prototypes interface](https://github.com/concourse/rfcs/pull/38) `get` action to return arbitrary values in its response which can be read through build-local vars. This would be very similar to my proposal for the `var` step reading from `var_sources`.

Aside from the `((foo:value))` point, I still don't really think of these values as var sources given the reasoning in my previous comment. The `var` step fetches a value in the same way that a single var is fetched, whose fields can then be accessed in-memory, so it feels more like a var, not a var source. The "var" was fetched by reading the value - what really have after that is a "field source". :smile:

So, here is my preference:

1. Design the `var` step and the reading of values primarily around the "simple values" case, as this will work with today's resource types which almost always write simple values. Complex values would of course still be supported, by reading fields off of the var.
1. Consider updating the [Resource Prototypes interface](https://github.com/concourse/rfcs/pull/38) so that resource prototypes can return arbitrary fields which can be read through a build-local var named after the step.
1. For now, we can use regular `((var))` syntax, but later we may want to consider an explicit build-local var syntax.

Then we would have 3 different syntaxes to mean 3 clearly different types of parameterization:

1. `((foo))` is only for statically provided vars
1. `((source:foo))` is only for reading from a var source defined under `var_sources`
1. `((.foo))` or something for build-local vars

I think when it comes time to inherit and maintain a gigantic CI stack that makes use of all 3 kinds of parameterization, we'll thank ourselves for having made a distinction between them. :slightly_smiling_face: 

---

Thanks again for putting up with me. I wouldn't blame you for being annoyed at the friction here, I just want to be very conscious of the decisions we make as they will have a lasting impact on pipeline authors to come. I've got an image in my head that feels "right" to me so I'm pushing for it but I also want to make sure my reasoning is sound and makes sense to you. :slightly_smiling_face:

evanchaoli:
@vito Okey, as you have finalized your thoughts, I'm ok with that. Code change over my previous commit is tiny. I have updated the commit (amended as you hadn't read the previous commit yet). Please take look, if you don't have big comments, then I'll start to add tests.

Now my testing pipeline turns to:

```yaml
resource_types:
- name: rt-tiny
  type: registry-image
  source:
    repository: evanli/tiny2
    tag: v4

resources:
- name: tiny
  type: rt-tiny
  source:
    key: foobar

- name: tiny2
  type: rt-tiny
  source:
    key: foobar2

- name: data
  type: mock
  source:
    create_files:
      data.yml: |
        format: yml
      data.json: |
        {"format": "json"}

jobs:
- name: test
  public: true
  plan:
  - get: tiny
  - get: data

  - var: tv
    file: tiny/a.txt
  - var: ty
    file: data/data.yml
  - var: tj
    file: data/data.json

  - task: a-task
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: sh
        args:
         - -exc
         - |
           echo tiny shows ((tv))
           echo data contains ((ty.format)) and ((tj.format))

  - get: tiny2
    params:
      data: ((tv))

  - put: tiny2
    params:
      path: ((tv))
```

Screen shot is exact same as the previous post, I don't duplicate attach it.

atc/exec/var_step.go
```diff
@@ -0,0 +1,137 @@
...
+
+	"github.com/concourse/baggageclaim"
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/atc/exec/artifact"
+	"github.com/concourse/concourse/atc/exec/build"
+	"github.com/concourse/concourse/atc/worker"
+	"github.com/concourse/concourse/vars"
+)
+
+// SetPipelineStep sets a pipeline to current team. This step takes pipeline
```
vito:
```suggestion
// VarStep loads a value from a file and sets it as a build-local var.
```

(We'll just amend this later when we add more functionality.)

atc/exec/var_step.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil, err
+	}
+
+	byteConfig, err := ioutil.ReadAll(stream)
+	if err != nil {
+		return nil, err
+	}
+
+	varFromFile := vars.StaticVariables{}
+	switch filepath.Ext(filePath) {
```
vito: I think this is a really nice default but it might be problematic if someone wants to load the raw JSON content as the value and pass it around as a black box. This is common with GCP keys, which are JSON, but they're always passed around verbatim.

In my RFC I had planned to add a `format` field, which can be `raw`, `yaml`, or `json`. I could be easily convinced of having it infer the format from the filename by default, and letting the user explicitly say `format: raw` to disable it. (Or vice versa - `format: json` if the file has no extension.) I think I'll put this as an open question in the RFC, but write the RFC assuming we do this by default.

(As with the `set_pipeline` step, the RFC won't block this, it'll just be experimental until we finalize the RFC. :slightly_smiling_face:)

vars/vars_tracker.go
```diff
@@ -82,29 +84,11 @@ func (t *credVarsTracker) IterateInterpolatedCreds(iter CredVarsTrackerIterator)
...
-
-func (t dummyCredVarsTracker) IterateInterpolatedCreds(iter CredVarsTrackerIterator) {
-	// do nothing
+	return t.enabled
 }
 
-func (t dummyCredVarsTracker) Enabled() bool {
-	return false
+func (t *credVarsTracker) AddVar(v Variables) {
+	t.credVars = NewMultiVars([]Variables{v, t.credVars})
```
vito: An alternative could be to instantiate `credVars` with a [`vars.StaticVariables`](https://github.com/concourse/concourse/blob/master/vars/static_vars.go#L7) embedded that we then mutate in this method.

Something like this:

```go
func NewCredVarsTracker(credVars Variables, on bool) CredVarsTracker {
  localVars := StaticVariables{}

  return &credVarsTracker{
    localVars: localVars,

    // note: I can never remember the proper order here...
    credVars: NewMultiVars([]Variables{localVars, credVars}),

    // ...
  }
}

func (t *credVarsTracker) SetLocalVar(name string, val interface{}) {
  t.localVars[name] = val
}
```

Splitting hairs, but this would result in less allocation and traversal through nested `NewMultiVars`. And maybe if we figure out a distinct syntax for build-local vars we wouldn't even need to mess with `credVars`. :thinking:

web/elm/src/Build/StepTree/StepTree.elm
```diff
@@ -828,6 +837,9 @@ viewStepHeaderLabel headerType stepID =
 
                 StepHeaderSetPipeline ->
                     "set_pipeline:"
+
+                StepHeaderVar ->
+                    "var"
```
vito:
```suggestion
                    "var:"
```

atc/exec/var_step.go
evanchaoli: I thought about to add a `format` attribute, as this was an initial version calling for review from design perspective, I didn't add it. Let me add it in next commit.

vars/vars_tracker.go
evanchaoli: Comparing with secret redaction, I guess cost of nested `MultiVars` is tiny. I think let's only address this issue once you work out the final RFC. Maybe in the RFC, you will also address if local var should be redacted, maybe have an attribute for `var` step to indicate if a local var should be redacted.

evanchaoli:
@vito I have addressed your review comments and added unit tests as well as a testflight case. If you have no more comment, I think we can merge this PR. Once your RFC is out, we can use a new PR to address new things from your RFC.

vars/vars_tracker.go
vito: I only meant to suggest a refactor here because it seems kind of confusing to keep chaining `StaticVariables` that each contain only one entry. I wasn't really thinking about secret redacting. :confused: I guess we can re-assess this when we implement local vars, in which case having a distinct separation would be more important so you can't accidentally access a local var with normal var syntax. Not a big deal right now. :+1: 

I can see why might want an option to toggle redaction; good point. The var step could be used to read an innocuous value like a dynamically generated Docker tag, in which case you probably *want* that to be printed in your build for clarity. For safety reasons I'll probably have them redact by default though in the RFC. We can keep talking about this point there - I'll try to have something out today.

atc/config.go
```diff
@@ -477,6 +477,15 @@ type PlanConfig struct {
...
 	Version *VersionConfig `json:"version,omitempty"`
+
+	// name of 'var' step
+	Var string `json:"var,omitempty"`
+
+	// format of input file.
+	Format string `json:"format,omitempty"`
+
+	// if or not dump local vars in 'var' step.
+	Dump bool `json:"dump,omitempty"`
```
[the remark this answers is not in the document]
evanchaoli: @vito  I designed the attribute `dump`  is not for post-build debugging purpose, for that, `fly intercept` can be used. Instead, a var might be informational, kinda like metadata of `get/put`, but `task` doesn't have metadata. So `dump` is not only for debugging, but also provide a way for pipeline to display information. From this perspective, `dump` is more descriptive, because it's just dumping the variable value for informational purpose.

I thought to dump var with a better format like:

```
foo.bar = xxx
fool.baa.rr = yyy
```

But it could need more code to write, yeah, I was lazy yesterday :), also I was not sure if you would allow `dump` to go, so I just dumped original content of input file. If you are ok with `dump`, I can change to dump var with above better format.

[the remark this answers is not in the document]
vito: Ahh I see, so it's not really about debugging, it's more about having a way to show interesting/important values in the build output.

I wonder if we could relate this to the redacting behavior? For example, instead of having a flag to enable/disable redacting and a flag to enable/disable dumping of values, we could have a flag that lists which fields to both a) prevent redacting and b) immediately print in the `var` step output in the same way that `get`/`put` metadata is shown:

```yaml
var: branch
reveal: [name]
```

vito:
I just published a draft of my RFC: https://github.com/concourse/rfcs/pull/27 ([rendered](https://github.com/vito/rfcs/blob/trigger-resources/027-var-steps/proposal.md))

Here's the tl;dr of where I landed relative to this PR so far:

Instead of a single `var` step which can read from either local file or a var source, I decided to split it into two steps: `load_var` and `get_var`. I did this so that it's clear that `passed` constraints can only be used in the var source case (`get_var`), and just to make the steps easier to reason about as they're really doing two entirely different things. (For example, `get_var` will have a container but `load_var` will not.)

I also like that the names are now [imperative](https://en.wikipedia.org/wiki/Imperative_mood) (as with `get`/`put`/`set_pipeline`), instead of just the word `var`, and that `load_var` kind of builds on the `--load-vars-from` flag. (Only it's loading a single var. :wink: I guess we could add `load_vars` in the future now that the name doesn't conflict with `vars`.)

The `load_var` step is essentially what is implemented in this PR, except I went ahead and added a `sensitive: true/false` flag (default `true`) which we can use to control redacting behavior. Though this may change if the suggestion in https://github.com/concourse/concourse/pull/4973#discussion_r368598329 resonates with you.

For local var syntax, I've come around to the idea of using var source syntax, but in a kind of different way: vars set through `load_var`, `get_var`, and `across` can be put into a special var source named `.`, e.g. `((.:foo))`. I like this more than `((.foo))`, which I previously proposed, but has the downside of making the var access look like a field access. I also like that this makes use of `.` as a special identifier, because the "valid identifiers" RFC (https://github.com/concourse/rfcs/pull/40) forbids it for normal use as a var source name, and intuitively it kind of makes sense given that `.` already means things like "local directory."

If the RFC makes sense to you, I think we can continue here by just renaming the step to `load_var`. Happy to discuss further details either here or in the RFC! :slightly_smiling_face:

atc/config.go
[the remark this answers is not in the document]
evanchaoli: I think `reveal` is a good idea. But it would require more changes to store revealed code in database, and GUI code change to show metadata for `var` step. So, let me remove `dump` from this PR, and use a separate PR to add `reveal` to `var` step, it should not be an urgent feature.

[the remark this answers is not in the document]
vito: :+1: Sounds good!

Just an idea: maybe we can add a generic `Metadata` build event which sets the metadata for a given step? We added generic `Initialize/Start/Finish` events with the `set_pipeline` step, but `FinishGet` and `FinishPut` events still exist, with their only difference being that they include the version and metadata (and an `ExitStatus` but that could be simplified to `Succeeded`):

https://github.com/concourse/concourse/blob/4bde6a2d8c77c9cb512c0ec641183112b3019603/atc/event/events.go#L132-L138

Maybe those could be split out into a separate event? Not really sure what to do about the `Version`, since the `var` step won't really need that. Hmm. Maybe it could become an optional step label or something, i.e. `Label: map[string]string`. :thinking:

evanchaoli:
@vito I have just posted a new commit that renamed `var` step to `load_var`. I rename `dump` and added `insensitive` to `load_var`. In the RFC, you proposed `sensitive` and default to `true`. But from implementation perspective, it's hard to default a value to true, because when user doesn't define the parameter, Go will default an absent bool field to false, we have no way to distinct if user define `sensitive` to false or user not define `sensitive`. So I changed to use `insensitive`, and default to false.

vito:
@evanchaoli Ah yeah that can be kind of annoying. Here's one way to have a bool field default to true: https://play.golang.org/p/pkKGThGRfAH

I think 'insensitive' feels a little awkward because that's more commonly used to describe [being insensitive towards others](https://dictionary.cambridge.org/dictionary/english/insensitive). :smile:

I'm not sure how users feel about boolean fields that default to true. I think a lot of us developers try to avoid it for the reason you gave, but at least there's a reasonable workaround. But I think a user probably just cares more about the word/flag name being obvious, even if it means defaulting to `true`.

Maybe we can sidestep this whole issue by just calling it `reveal`? That will allow it to default to `false` anyway. We can have `reveal: true` mean "don't redact" and maybe the "show in metadata" functionality in the future. And then in a separate PR we can have it accept both a bool and an array of fields to show as medadata. :thinking:

evanchaoli:
@vito Using pointer is the way I can come out for defaulting a bool value to true, but that looks weird, so I tried to avoid it. I have changed to `insensitive` to `reveal`.

In the last commit that I have just pushed, I have tested with both secret-redaction turned on and off. So it should be ready for review/merge.

The current test pipeline code is:

```yaml
resource_types:
- name: rt-tiny
  type: registry-image
  source:
    repository: evanli/tiny2
    tag: v4

resources:
- name: tiny
  type: rt-tiny
  source:
    key: foobar

- name: tiny2
  type: rt-tiny
  source:
    key: foobar2

- name: data
  type: mock
  source:
    create_files:
      data.yml: |
        format: yml
      data.json: |
        {"format": "json"}

jobs:
- name: test
  public: true
  plan:
  - get: tiny
  - get: data

  - load_var: tv
    file: tiny/a.txt
  - load_var: ty
    file: data/data.yml
  - load_var: tj
    file: data/data.json
    reveal: true
  - load_var: tj2
    file: data/data.json
    format: raw

  - task: a-task
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: sh
        args:
         - -exc
         - |
           echo tiny shows ((.:tv))
           echo data contains ((.:ty.format)) and ((.:tj.format))
           echo data.json in raw format: ((.:tj2))

  - get: tiny2
    params:
      data: ((.:tv))

  - put: tiny2
    params:
      path: ((.:tv))
```

where `tj` is revealed. From the screen shot, we can see that `tj.format` is not redacted.

load_var

vars/vars_tracker.go
```diff
@@ -13,35 +14,64 @@ type CredVarsTrackerIterator interface {
...
-	if found {
+	var (
+		val   interface{}
+		found bool
+		err   error
+	)
+
+	redact := true
+	parts := strings.Split(varDef.Name, ":")
+	if len(parts) == 2 && parts[0] == "." {
```
vito: Maybe `credVars` could have an entry within its `NamedVariables` whose name is `.`, and have the same entry passed to the `*credVarsTracker` so it can directly set vars in it? I think that way we can avoid both this change and the change to `vars/template.go` that I commented on.

vars/vars_tracker.go
```diff
@@ -13,35 +14,64 @@ type CredVarsTrackerIterator interface {
...
+		varDef.Name = parts[1]
+		val, found, err = t.localVars.Get(varDef)
+		if found {
+			parts = strings.Split(varDef.Name, ".")
+			if _, ok := t.noRedactVarNames[parts[0]]; ok {
+				redact = false
```
vito: Ah, I guess this part would be kind of challenging, huh... :thinking:

web/elm/tests/BuildStepTests.elm
```diff
@@ -221,6 +221,13 @@ all =
                     >> when iAmLookingAtTheStepBody
                     >> then_ iSeeThePipelineName
             ]
+        , describe "load_var step"
```
vito: Awesome, thanks for adding Elm tests for this!

vars/template.go
```diff
@@ -142,8 +142,18 @@ type varsLookup struct {
 
 var ErrEmptyVar = errors.New("empty var")
 
+// Get value of a var. Name can be the following formats: 1) 'foo', where foo
+// is var name; 2) 'foo:bar', where foo is var source name, and bar is var name;
+// 3) '.:foo', where . means a local var, foo is var name.
 func (l varsLookup) Get(name string) (interface{}, bool, error) {
-	splitName := strings.Split(name, ".")
+	var splitName []string
```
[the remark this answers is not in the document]
evanchaoli: Unless we make some refactor, with current implementation, this change is not avoidable. This piece of code splits var name and its fields, for example: `foo.bar` will be split to `foo` and `bar`; `name:foo.bar` will be split to `name:foo` and `bar`, where `name:foo` will further be handled by `NamedVariables`. But for `.:foo`, the original code cannot handle, so I had to make this change.

vars/vars_tracker.go
evanchaoli: If later on we add `reveal` to `var_source`, then we can directly add local vars to `NamedVariables`.

evanchaoli:
@vito Thanks for catching those "var" I badly left in the code. I have fixed them all, and did a self review again. Hope this version is good enough to merge.

Just FYI, I'm going to start the Chinese New Year holiday vacation, so I'll not be responsive in next 10 days.

Please also help take a look https://github.com/concourse/concourse/pull/5034, that is an initial code change of OPA integration. Take your time, no hurry.

evanchaoli:
@vito I just back to work earlier. I realized a conflict due to the 6.0 merge, and I have fixed the conflict. Please see if any other change is required.

@pivotal-jamie-klassen Please take a look for the UI part. Basically I just cloned the changes of `set_pipeline` step.

vito:
@evanchaoli Welcome back - hope you had a nice break! Thanks for resolving the conflict, I'll take another look at this soon and aim to merge it in for v6.0.

evanchaoli:
@vito It seems something wrong with the CI backend Concourse. Yesterday I got all tests passed, but today I noticed "watsjs" reran and failed, but the error was not related to my change. Then I forced to push again but tests have not been triggered yet for long time.

PR #4988, opened 2020-01-09
worker/integration: improve local containerd integration testing flow

cirocosta:
### Why do we need this PR?

previously, one needed to go through a sequence of manual steps to get
the integration test running:

1. build an image
2. create a container from that
2. getting inside the container
3. run the tests

this new one changes the flow to leverage the `unit` image (that already
has all of the bits we need), and put into the script the manual work
that we needed to do before.

after this changes, the flow looks like:

1. `run containerd`
2. `run test`

**ps.: depends on https://github.com/concourse/ci/pull/253** - without it, `concourse/unit` (the image used as the base here) will not have the proper binaries for running `containerd`.

- have a script that sets up the environment for running the integration tests, leveraging the image that already has all the `containerd` contents

### Contributor Checklist

- ~Unit tests~

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


worker/backend/integration/run
```diff
@@ -0,0 +1,35 @@
...
+    ;;
+  esac
+}
+
+start_containerd() {
+  docker run --privileged -it \
+    --name $CONTAINER_NAME \
+    --volume /scratch \
+    --volume gopath:/gopath \
+    --volume $(realpath ../../..):/src \
```
pivotal-jamie-klassen: When I run `worker/backend/integration/run containerd` from the root of the repo, I get
```bash
docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "process_linux.go:449: container init caused \"mkdir /src/worker: permission denied\"": unknown.
```
but if I run it from `worker/backend/integration` I think it works. Maybe something like this?
```suggestion
    --volume $(realpath $(dirname $0)/../../..):/src \
```

worker/backend/integration/run
```diff
@@ -0,0 +1,35 @@
...
+    --workdir /src/worker/backend/integration \
+    concourse/unit \
+    /bin/bash -c '
+PATH=/usr/local/concourse/bin \
+	containerd --config <(echo "state = \"/scratch\"")
```
pivotal-jamie-klassen:
```suggestion
	/usr/bin/containerd --config <(echo "state = \"/scratch\"")
```
I think when you override the `PATH` env var, in my case I got
```
/bin/bash: line 1: containerd: command not found
```

cirocosta:
> Am I doing something wrong? I tried pulling concourse/unit first and also adding my suggested changes (before this I couldn't actually get anything running)

oh, sorry, I should've made more clear that this is assuming `concourse/unit` from concourse/ci#253 getting in.

worker/backend/integration/run
cirocosta: ooh yeah, totally! thanks!

worker/backend/integration/run
cirocosta: this is pretty much what I referred to in https://github.com/concourse/concourse/pull/4988#issuecomment-572697492 - the `concourse/unit` that you used did not yet have concourse's `containerd` binaries (which live under `/usr/local/concourse/bin`), only the `containerd` that comes from the `docker` installation (which is of a different version than the one we use).

worker/backend/integration/run
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env bash
+
+set -o errexit
+
+readonly CONTAINER_NAME="containerd-integration"
+readonly CONCOURSE_DIR="$(realpath $(dirname $0)/../../..)"
```
cirocosta: @pivotal-jamie-klassen thanks for the tip! btw, there's still the need for `realpath` because `--volume` needs an absolute path: https://github.com/docker/cli/issues/1203

pivotal-jamie-klassen:
Big question about `containerd` - do you need to run the tests in the same container? Can we instead do some docker port-forwarding or volume-mounting to allow the test process to run outside the container?

Even bigger - there are multiple scenarios, especially now, where our tests interact with `dockerd` and `containerd` processes that are just assumed to be running. Is it possible/does it make sense for our test process to actually control the lifecycle of them, in a similar way to https://github.com/concourse/concourse/blob/master/atc/postgresrunner/postgresrunner.go? The only interesting difference that `postgres` or `initdb` have from `dockerd` or `containerd`, as far as I can tell, is that they can run more easily in a container.

overall i'm interested in a simpler contract between this repo and `concourse/ci`, or "dev/ci parity".

WDYT??

cirocosta:
Hey @pivotal-jamie-klassen , thx for the thorough review!

> Can we instead do some docker port-forwarding or volume-mounting to allow the
> test process to run outside the container?

Unfortunately, no :disappointed:. The "core services" are expected to be
controlled from an agent in the same host.

```
[...] By design, we do not support the core plugins registering on the TCP
endpoint as these services are scoped to be used on a single host.
```

from [comment](https://github.com/containerd/containerd/issues/3466#issuecomment-516204803)
on making containerd remote-friendly.

---

(will get to the other points soon)

pivotal-jamie-klassen:
@cirocosta good point, I was poking around in that same issue but didn't understand just how "core" a  "core plugin" was! It's probably helpful to know (and maybe document...) that `containerd` and the test process must be running on the same host.

cirocosta:
> overall i'm interested in a simpler contract between this repo and
> concourse/ci, or "dev/ci parity".

that's a great concern to have, and I'm 100% on the idea of working towards that
:+1:

so far, the way that I've been developing the `containerd` code has been pretty
much the way that we have set up here: `containerd` is running in the background
(I keep it as a `systemd` service - see [cirocosta/containerd-install]),  and
then I target it through the tests (which is essentially what `ci` is doing
here), or manually through `ctr`.

my guess to why this is not much the case with the `db` tests is that (I guess)
we usually don't have `postgres` running in our machines in the background.

[cirocosta/containerd-install]: https://github.com/cirocosta/containerd-install

do you think we're good to go with the current way approach? (feel free to say
*no*)

thanks!

pivotal-jamie-klassen:
@cirocosta thanks for your thoughtful reflections! I have an observation and two questions.

Observation: I don't think we use that postgresrunner for quite the reason you've described. If I had to guess, I'd say it's because fully killing and reviving the postgres process means you can reliably have a clean DB/schema for every test, guaranteeing repeatable tests. I was seeing containerd as a stateful process where, as the test suite grows there might be some risk of tests leaving a muddied context/failing to run independently.

Question: am I off base? Maybe containerd is not so stateful as I thought. While the "clean operating room" guarantee is nice, I think it is also likely to slow down the test suite with so much shutdown/startup time, and if we also want multiple parallel container d processes for parallel test threads it could get quite expensive?

Second question: when we think of realistic user narratives for the containerd backend, is it likely folks will have a colocated systemd service? I know systemd itself uses cgroups to manage the capabilities of its managed processes, so your comment has me wondering if we should attempt to faithfully simulate that scenario.

cirocosta:
sorry for the long delay!

> I don't think we use that postgresrunner for quite the reason you've described.

yeahh... that's true

> FWIW also even though it's a small thing, there is a remarkable growth in
> developer Joy when I can push a single button to run my tests vs having to
> establish any latent context on my machine.

tootally!


> Maybe containerd is not so stateful as I thought. 

hmmm it actually kinda is :( e.g., when tests fail, I always end up running my
["prune containerd"] script, which, as you said, if we had a runner that takes
care of spinning it up & down etc, I'd not really have to do that :thinking:

["prune containerd"]: https://gist.github.com/cirocosta/b01d9869b38e73690bb02a6371c2b221


> While the "clean operating room" guarantee is nice, I think it is also likely
> to slow down the test suite with so much shutdown/startup time, and if we also
> want multiple parallel containerd processes for parallel test threads it
> could get quite expensive?

aaah .. I think that's something we could workaround with the type of things one
can do with Ginkgo's `SynchronizedSuite`, having every test targetting the same
containerd instance that was brought up as nothing would collide (every action
towards `containerd` *must* be towards a namespace anyway)

:thinking:

I'm very up to giving a try at having a runner that would take care of setting
containerd up for the tests exec :+1:

PR #5013, opened 2020-01-10
Vault credential manager supports lookup path templates

### LINKED ISSUE #4965: Path template support for Vault credential manager

The Vault integration asks operators to configure a particular path where secrets will live. By default, this is `/concourse`, so that the credential manager will look for team secrets under `/concourse/myteam`, and pipeline secrets under `/concourse/myteam/mypipeline`. Concourse itself will have read access to the whole hierarchy, and probably each team has more restrictive access control in place for their namespaces.

We have a use case for connecting to an existing Vault instance which is used for a variety of secrets (not just Concourse-facing ones). Team members want to separate their CI-related secrets from other data, so that they can exercise control over which values are accessible from a Concourse pipeline. Equally, Concourse itself should not have unrestricted access to the whole store. The ideal scheme would be for each team to use the following:

* `/secret/myteam/ci/ `- team secrets for Concourse
* `/secret/myteam/ci/mypipeline/` - pipeline secrets for Concourse
* `/secret/myteam/foo/bar/` - some other existing secrets, not for Concourse

Then the policy in Vault would be for Concourse to be able to read anything from `/secret/+/ci/*` (using Vault policy notation from https://www.vaultproject.io/docs/concepts/policies.html), but nothing else. Individual team members would continue to have fine-grained policies referring to particular paths within their spaces, as needed.

Unfortunately, the credential manager configuration on the Concourse side only supports a fixed path prefix. Some of the other credential managers support a template, specifically using the Go `text/template` library - for example, the SSM one allows configuring a template for team secrets, defaulting to `/concourse/{{.Team}}/{{.Secret}}`. If the Vault manager did the same, we could configure it to use `/secret/{{.Team}}/ci/{{.Secret}}`, and likewise for pipeline secrets, to get the desired setup.

While we could have a parallel Concourse hierarchy (most secrets under `/secret/myteam`, Concourse-facing ones under `/concourse/myteam`, etc.), this would be quite inconvenient for management of the secret values and associated policy. Some stuff would be duplicated between the trees, and it would be more error-prone to keep Vault policy in sync. Another idea is to give Concourse a prefix of `/secret`, but use policy to limit its actual access to whatever each team considers to be a "CI secret". But this sounds really awful to manage and reason about.

I suggest that the Vault manager could grow two new options, for team and pipeline secret templates, by analogy with the other credential managers that support templates. For compatibility, the [PathPrefix](https://github.com/concourse/concourse/blob/master/atc/creds/vault/manager.go#L19) option should be retained, and setting `CONCOURSE_VAULT_PATH_PREFIX=/foo` would be implemented internally by choosing `/foo/{{.Team}}/{{.Secret}}` as the team template, and `/foo/{{.Team}}/{{.Pipeline}}/{{.Secret}}` as the pipeline template. (But it would be an error to try to set the path prefix at the same time as setting the explicit templates.)

I think SharedPath should just be the same since it's a static string in any case.

Behind the scenes, there is [current logic](https://github.com/concourse/concourse/blob/master/atc/creds/vault/vault.go#L27) to smash the various strings together with `/` to end up with a list of lookup paths. That would need to change to instantiate templates instead. There is a refactoring opportunity as well, where the credential managers right now either use the shared `creds.NewSecretLookupWithPrefix` (Vault, k8s, CredHub) or duplicate template-instantiation logic (SSM, Secrets Manager) - I could imagine having a shared `creds.NewSecretLookupFromTemplate` or whatever, instead.

Yes 😃

agurney:
# Existing Issue

Fixes #4965.

# Why do we need this PR?

Some credential managers support lookup according to templates, but the Vault one didn't - the search paths were hardcoded, aside from the ability to customize a prefix. Now, with more general support for templated lookup paths, people whose secrets are organized in other ways can have an easier time integrating.

This will work for `var_sources` as well as for a globally configured credential manager. In #4965, the use case was for global configuration of lookups into
* `/secret/TEAM/ci/PIPELINE`
* `/secret/TEAM/ci/`

so that non-CI-relevant secrets in each `TEAM` namespace could be separated from Concourse's permissions. In the case of `var_sources`, somebody could now configure one or several sources in a pipeline pointing to specifically chosen sub-areas of a vault, without having to use a `TEAM/PIPELINE` scheme for that data.

* Refactor template support in all relevant credential managers (ssm, secretsmanager, conjur), which was previously duplicated. There is now a `creds.SecretLookupWithTemplate` type, analogously to `creds.SecretLookupWithPrefix`; when created, this gets a Go `template.Template`, and team and pipeline names. That template is instantiated with those, plus the variable name, on lookup. Some validation logic is also now transferred into the `BuildSecretTemplate` function.

* Alter the Vault manager to use templates. As [discussed](https://github.com/concourse/concourse/issues/4965#issuecomment-572162283), it can use an arbitrary list of lookup templates. The default is to have `/{{.Team}}/{{.Pipeline}}/{{.Secret}}` and `/{{.Team}}/{{.Secret}}`, which matches the existing logic.

These templates are interpreted subject to any provided path prefix and namespace in Vault configuration, so if I set a namespace of `/foo`, prefix of `/bar` and a template of `/baz/{{.Secret}}`, then I'll end up looking in `/foo/bar/baz/{{.Secret}}`. This is consistent with how a "shared path" is handled currently, though really such a path is just a template that doesn't vary with Team and Pipeline.

In the linked issue, there was an idea of exposing different config knobs for vault when it's exposed in `var_sources`, compared to when it's set globally. Having not thought _very_ deeply about it, my rough sense is in favor of being consistent and flexible, over cleaning up a set of options that's admittedly become a little cluttered. If we do want the config managers to take different options in general when appearing in `var_sources`, I'd appreciate any guidance about the overall cleanest way to express that structure.

Templates can be provided on the command line with `--vault-lookup-templates` or with an environment variable:
```sh
CONCOURSE_VAULT_LOOKUP_TEMPLATES=/foo/{{.Team}}/{{.Secret}},/bar/{{.Team}}/{{.Secret}}
```
For `var_sources` the configuration (I think?) ends up as `lookup_templates: [...]` under the vault config.

# Contributor Checklist
- [X] Unit tests
- [X] Updated documentation (located at https://github.com/concourse/docs)

Not sure what integration tests are needed; I saw Topgun has Vault-related tests, but really just for making sure auth works, as opposed to probing any other details of the configuration. I also noticed the "don't worry about this" note in CONTRIBUTORS so I'm not worrying about it until informed otherwise 😉 

~I'll update docs if all this seems correct and desirable - at least [the vault page](https://github.com/concourse/docs/blob/master/lit/docs/operation/creds/vault.lit) and [the var_sources one](https://github.com/concourse/docs/blob/master/lit/docs/config/dynamic-vars.lit).~
Documentation PR at https://github.com/concourse/docs/pull/292
concourse-bosh-release PR at https://github.com/concourse/concourse-bosh-release/pull/88

There should also be the same in [concourse-chart](https://github.com/concourse/concourse-chart), but as I'm not a Helm user I feel less confident on that one.

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


vito:
Sorry for the wait, taking a look at this now! :slightly_smiling_face:

atc/api/info_test.go
```diff
@@ -183,11 +183,12 @@ var _ = Describe("Pipelines API", func() {
...
 				vaultManager := &vault.VaultManager{
-					URL:        credServer.URL(),
-					Namespace:  "testnamespace",
-					PathPrefix: "testpath",
-					TLS:        tls,
-					Auth:       authConfig,
+					URL:             credServer.URL(),
+					Namespace:       "testnamespace",
+					PathPrefix:      "testpath",
+					LookupTemplates: []string{"/{{.Team}}/{{.Pipeline}}/{{.Secret}}", "/{{.Team}}/{{.Secret}}"},
```
vito: Side note, not a blocker: I'm iffy on using Go template syntax instead of `((vars))` for lookup paths, but I'm OK with it for now as it's what all the other credential managers use and we don't have a great alternative yet.

More detailed thoughts:

My main concern is whether users will be confused by seeing both `{{.Foo}}` and `((foo))` in the same pipeline config, assuming they're using `var_sources`. I think it would be better to just use the same templating syntax consistently throughout the product.

That being said, I don't think worth switching to `((vars))` for this yet. It would be ambiguous with regular static var interpolation: if the user had paths like `/((team))/((pipeline))/((secret))` but then does `fly set-pipeline -v team=foo` it would unintentionally fill in the `((team))` var.

This same concern came up in #4973 and now I'm considering adding a "local var" syntax, which would fix this problem. It might look something like `/((.team))/((.pipeline))/((.secret))`.

...so tl;dr: I'm OK with just using Go templates for now unless you happen to be interested in implementing this "local vars syntax" work. :sweat_smile:

We can revisit this if and when we transition to using [Prototypes](https://github.com/concourse/rfcs/pull/37) for `var_sources` - I'm working on an RFC for that. At that point I think we can pretty much have free reign on implementing a new Vault prototype for use with `var_sources` and we wouldn't have to worry so much about backwards-compatibility since `var_sources` is still experimental. That would also be a good time to clean up SharedPath and friends.

atc/api/info_test.go
```diff
@@ -255,6 +256,7 @@ var _ = Describe("Pipelines API", func() {
           "vault": {
             "url": "` + credServer.URL() + `",
             "path_prefix": "testpath",
+            "lookup_templates": ["/{{.Team}}/{{.Pipeline}}/{{.Secret}}", "/{{.Team}}/{{.Secret}}"],
```
vito: Another side note: in the future we'd probably just call this `lookup_paths` since the prototype wouldn't really know whether or not `((vars))` were used. This would be similar to how resources don't really know whether `((vars))` are used for their various parameters in `source` and `get`/`put` steps.

It's tempting to just call this `lookup_paths` from the get-go but I don't think it matters a whole lot. It might make it easier for folks to transition to the Vault prototype configuration scheme in the future, but they'd already have to swap out the `{{.Foo}}` syntax for `((.foo))` anyway. It might be better to just leave it as a different parameter and have the prototype complain loudly about an unknown `lookup_templates` field when users switch over. :slightly_smiling_face:

atc/creds/conjur/conjur.go
```diff
@@ -28,24 +27,11 @@ func NewConjur(log lager.Logger, client IConjurClient, secretTemplates []*templa
...
 	for _, template := range c.secretTemplates {
 		c.log.Info(" teamname: " + teamName + "pipeline: " + pipelineName)
-
-		lPath := &SecretLookupPathConjur{
-			NameTemplate: template,
-			TeamName:     teamName,
-			PipelineName: pipelineName,
-		}
-
+		lPath := creds.NewSecretLookupWithTemplate(template, teamName, pipelineName)
```
vito: Thanks for refactoring this!

atc/creds/vault/manager.go
```diff
@@ -13,12 +14,14 @@ import (
...
 
+
 type VaultManager struct {
 	URL string `mapstructure:"url" long:"url" description:"Vault server address used to access secrets."`
 
-	PathPrefix string `mapstructure:"path_prefix" long:"path-prefix" default:"/concourse" description:"Path under which to namespace credential lookup."`
-	SharedPath string `mapstructure:"shared_path" long:"shared-path" description:"Path under which to lookup shared credentials."`
-	Namespace  string `mapstructure:"namespace" long:"namespace"   description:"Vault namespace to use for authentication and secret lookup."`
+	PathPrefix      string `mapstructure:"path_prefix" long:"path-prefix" default:"/concourse" description:"Path under which to namespace credential lookup."`
+	LookupTemplates []string `mapstructure:"lookup_templates" long:"lookup-templates" default:"/{{.Team}}/{{.Pipeline}}/{{.Secret}}" default:"/{{.Team}}/{{.Secret}}" description:"Path templates for credential lookup"`
```
vito: Interesting, never seen this multi-default thing before. :eyes: ...Does that work?

atc/creds/vault/vault.go
```diff
@@ -18,18 +19,18 @@ type SecretReader interface {
...
+	SecretReader    SecretReader
+	Prefix          string
+	LookupTemplates []*template.Template
+	SharedPath      string
 }
 
 // NewSecretLookupPaths defines how variables will be searched in the underlying secret manager
 func (v Vault) NewSecretLookupPaths(teamName string, pipelineName string, allowRootPath bool) []creds.SecretLookupPath {
 	lookupPaths := []creds.SecretLookupPath{}
-	if len(pipelineName) > 0 {
```
vito: This is interesting. Will the templates will just be evaluated with an empty pipeline name now? :thinking:

release-notes/latest.md
```diff
@@ -1,3 +1,12 @@
 #### <sub><sup><a name="4938" href="#4938">:link:</a></sup></sub> feature
 
 * Include job label in build duration metrics exported to Prometheus. #4976
+
+#### feature
+
+* Support path templating for secret lookups in Vault credential
```
vito: Feel free to credit yourself! :slightly_smiling_face: e.g. '@agurney added support for ...'

Most release notes do this but it's not necessary if you don't want to. We usually link to the PR at the end though so people can see more details (see above release note).

atc/creds/vault/manager.go
agurney: Yeah, it reads pretty wacky, but https://github.com/jessevdk/go-flags/blob/master/parser_test.go#L35 shows it in action.

release-notes/latest.md
agurney: I'll pop that in, a bit chicken/egg initially for knowing the PR number ;-)

atc/api/info_test.go
agurney: The local var idea is interesting for being able to handle these things in the same mechanism as other vars, versus having a completely separate templating mechanism. I'm also a bit iffy on the stupendous power that `text/template` theoretically has - hopefully people aren't doing anything too adventurous here.

I guess the local-var model would be that it's an implicit var source, but otherwise functionally like any other, and populates a small namespace of team name, etc. They could be used in this structure or anywhere else. I anticipate people wanting `((.build-id))` too, looking at the [resource metadata](https://concourse-ci.org/implementing-resource-types.html#resource-metadata), albeit that can unlock some poor patterns.

atc/creds/vault/vault.go
agurney: Oh, good catch - I think there needs to be logic to suppress generated paths that have `//`, which is what other credential managers do in the same situation. Vault is perfectly happy to receive such paths, but they might end up pointing to a wrong place in the event I evaluate a `{{Pipeline}}`-having template in a non-`{{Pipeline}}` context.

agurney: Logic added, same as in the other cases. This still only catches the exact `/{{.Pipeline}}/` case, as opposed to e.g. `/pipeline_{{.Pipeline}}/` or something, should anyone use that.

It might be stronger to change to a test where we probe whether the template is pipeline-name-dependent (instantiate twice with different pipeline names and see if we get different strings), and refuse to use it for real if there is no pipeline name in the current context. I'm not sure on bundling that in as a general behavior change in this PR, though.

release-notes/latest.md
agurney: In it has popped.

atc/creds/vault/vault.go
vito: Maybe we could provide the template one of two structs:

```go
type PipelineEvaluation struct {
  Team     string
  Pipeline string
  Secret   string
}

type TeamEvaluation struct {
  Team   string
  Secret string
}
```

This way we can explicitly capture an error returned when the template tries to access the missing `.Pipeline` field just ignore the path. I'm not sure if there would be a nice error type to check for though.

agurney:
@vito I've added some subsidiary PRs for bosh and docs, and fixed that bug you noticed.

atc/creds/vault/vault.go
vito: Were you able to try the above approach? Seems like it'd be a little better than checking for `//` since it would also catch exotic scenarios like `/team:{{.Team}}/pipeline:{{.Pipeline}}/`.

agurney:
Thanks for getting back to this!
I've rebased and updated `latest.md` as requested. Plus, I was able to get the "proper" logic for detecting pipeline-dependent templates, in the newest commit https://github.com/concourse/concourse/pull/5013/commits/5f7d9201255670e871e14ffe7b43c787f5baa030 (and see its commit message for detail on my thinking). The implementation now tries to expand the template with no pipeline parameter given, at initialization time, and if that works with no error, it means the template didn't use `.Pipeline`. We save that information, so that where we previously tried to detect `//` when the pipeline name was an empty string, we _now_ just refer to that cached flag.

atc/creds/vault/vault.go
agurney: Yes, I ended up adding something much like this at template creation time. Details in the latest commit and much rambling in its commit message.

atc/creds/vault/manager.go
```diff
@@ -123,6 +130,13 @@ func (manager VaultManager) Validate() error {
 		return fmt.Errorf("path prefix must be a non-empty string")
 	}
 
+	for i, tmpl := range manager.LookupTemplates {
+		name := fmt.Sprintf("lookup-template-%d", i)
```
vito, line 134: This `name` argument seems a little awkward to use - is it required, or can it just be a made-up string in `creds.BuildSecretTemplate`? 🤔

atc/creds/vault/manager.go
```diff
@@ -89,6 +92,10 @@ func (manager *VaultManager) MarshalJSON() ([]byte, error) {
 func (manager *VaultManager) Config(config map[string]interface{}) error {
 	// apply defaults
 	manager.PathPrefix = "/concourse"
+	manager.LookupTemplates = []string{
```
vito, line 95: I think we might need to do something different here. I noticed in testing that when configuring as a var source, the defaults don't seem to get replaced - the configuration is additive. They're replaced properly if I configure `CONCOURSE_VAULT_LOOKUP_TEMPLATES` though.

If you need a quick way to try this out, here's my pipeline:

```yaml
var_sources:
- name: myvault
  type: vault
  config:
    url: https://vault:8200
    ca_cert: ((vault_ca_cert))
    client_cert: ((vault_client_cert))
    client_key: ((vault_client_key))
    lookup_templates: [/woahnelly]
    auth_backend: cert

jobs:
- name: use-creds
  plan:
  - task: print-secret
    config:
      platform: linux
      image_resource:
        type: mock
        source: {mirror_self: true}
      run:
        path: /bin/sh
        args:
          - -c
          - |
              echo ((myvault:some-secret))
```

Here's what I did:

```sh
# in concourse repo
$ ./hack/dc up
$ ./hack/vault/init
$ vault secrets enable -version=1 -path=concourse kv
$ vault write concourse/main/vault-repro/some-secret value=hello

# (replace these paths)
$ cat <<EOF > vault-values.yml
vault_ca_cert: $(cat ~/src/concourse/hack/vault/certs/vault-ca.crt | jq -Rs .)
vault_client_cert: $(cat ~/src/concourse/hack/vault/certs/concourse.crt | jq -Rs .)
vault_client_key: $(cat ~/src/concourse/hack/vault/certs/concourse.key | jq -Rs .)
EOF
$ fly -t l sp -p vault-repro -c vault-repro.yml -l vault-values.yml
```

When I triggered `use-creds`, I expected it to only look for `/concourse/woahnelly/some-secret` and result in an error, however it actually finds the credential in the default path and succeeds.

atc/creds/vault/manager.go
agurney: I think the only thing the names are good for, apart from maybe appearing in error messages, is for being invoked by name. Hopefully nobody is doing such a thing in the case of any of these templates! Errors are like `template: foo:1: unexpected unclosed action in command` for a template named `foo` that's defined as `"{{"`. So there's a bit of convenience value in having that bubble up with a name attached. We could create with an empty name, which yields the same sort of error message in the end, just with `template: :1:...`.

atc/creds/vault/manager.go
agurney: Thanks, trying to reproduce this.

agurney: Ah, got it. I didn't catch this earlier because it happens just when configuring <2 custom templates. They're individually overwriting the entries in the default array, as opposed to replacing it as one unit, or even appending. Looks like behavior from the `mapstructure` library which should be fine to tweak or work around.

agurney:
It now inserts `/` when joining prefix to template in a more clever way than not at all, and I fiddled with the defaulting logic. There is an old-ish issue https://github.com/mitchellh/mapstructure/issues/74 around the behavior for updating arrays and slices. This could possibly be achieved by a custom `DecodeHook` but I thought "eh, let's just write the logic directly instead of introducing a use-once abstraction".

On testing, I am very much relying on the existing suite for the _other_ cred managers touched here. I agree it would be nice to have more coverage in order to be really confident that this sort of change doesn't accidentally destroy the intended end-to-end behavior. The unit tests don't quite get us there since we are a little short of verifying the full journey from configuration to execution. 

Perhaps there's a way to get functional tests that invoke mock/stub versions of the various APIs? That might even be a little future-proof for when all this gets torn out in favor of the new model. I guess that means the cred managers become special resources? Then those tests could move fairly smoothly out of this repo into their future home. Or there's the material in `topgun/`, which is a bunch of setup, but apparently can test against the real backends.

PR #5015, opened 2020-01-11
atc: always emit a metric for each worker state

### LINKED ISSUE #4985: `worker_state` metric count off by one
## Bug Report

After upgrading to v5.8.0 the `worker_state` metric shows one less worker than actually running. For instance `fly -t main workers` and the database table `workers` both show 5 running workers but the metric `worker_state` shows 4. I suspect this is related to https://github.com/concourse/concourse/pull/4895

## Steps to Reproduce

- Spawn 2+ workers.
- Check running workers count with `worker_state` metric

## Expected Results

The `worker_state` metric matches the number of running workers.

## Actual Results

The `worker_state` metric doesn't match the number of running workers, it is missing one.

## Version Info

* Concourse version: 5.8.0
* Infrastructure/IaaS: AWS
* Did this used to work? Yes.

pivotal-jamie-klassen:
# Existing Issue

Fixes #4985.

The key change is to ensure that 0 values for state counts are emitted. I have a bit of a concern around the scalability of this - before this change we only emitted metrics for states where the count of workers was greater than zero, but now we're guaranteed 5 worker state metric events every tick. I guess a typical healthy cluster probably tends to have all running workers most of the time, so this could effectively mean 5 times as many worker state metrics events.

I don't know what the typical "utilization rate" of the metrics queue is (like how many events tend to be queued up as a proportion of the max length and in particular the typical proportion of the queue's max length that are worker state metrics), but if a cluster has a tendency to get that "queue-full" error I could see this change exacerbating it.

Outside of the "queue utilization" problem is the question of how the various metrics emitters are implemented -- if the emitter makes a request to an external (potentially costly or rate-limited) system, then increasing the number of events in this fashion could also put pressure on those systems. Obviously Prometheus wouldn't suffer from this problem, and I tried to figure out how the Datadog client worked, but couldn't see a non-noop concrete implementation of the `Gauge` method. Admittedly I didn't look for that long.

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/metric/metrics.go
+++ b/atc/metric/metrics.go
@@ -667,28 +667,22 @@ type WorkersState struct {
 }
 
 func (event WorkersState) Emit(logger lager.Logger) {
-	var (
-		perStateCounter = map[db.WorkerState]int{}
-		eventState      EventState
-	)
-
-	for _, workerState := range event.WorkerStateByName {
-		_, exists := perStateCounter[workerState]
-		if !exists {
-			perStateCounter[workerState] = 1
-			continue
-		}
-
-		perStateCounter[workerState] += 1
-	}
+	var eventState EventState
 
 	for _, state := range db.AllWorkerStates() {
-		count := perStateCounter[state]
+		count := 0
+		for _, workerState := range event.WorkerStateByName {
+			if workerState == state {
+				count += 1
+			}
+		}
+
 		if state == db.WorkerStateStalled && count > 0 {
 			eventState = EventStateWarning
 		} else {
 			eventState = EventStateOK
 		}
+
 		emit(
 			logger.Session("worker-state"),
 			Event{
--- a/atc/metric/metrics.go
+++ b/atc/metric/metrics.go
@@ -675,20 +675,20 @@ func (event WorkersState) Emit(logger lager.Logger) {
 	for _, workerState := range event.WorkerStateByName {
 		_, exists := perStateCounter[workerState]
 		if !exists {
-			perStateCounter[workerState] = 0
+			perStateCounter[workerState] = 1
 			continue
 		}
 
 		perStateCounter[workerState] += 1
 	}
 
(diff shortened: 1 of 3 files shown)


atc/metric/metrics.go
```diff
@@ -675,20 +675,20 @@ func (event WorkersState) Emit(logger lager.Logger) {
...
 			continue
 		}
 
 		perStateCounter[workerState] += 1
 	}
 
-	for state, count := range perStateCounter {
+	for _, state := range db.AllWorkerStates() {
+		count := perStateCounter[state]
 		if state == db.WorkerStateStalled && count > 0 {
```
kirillbilchenko: just wondering in this case if count can be < 0? does it make sense to check this count, since if we set `perStateCounter[workerState] = 1` to 1 means at least one will be in the map? correct me if I'm wrong?

pivotal-jamie-klassen: @kirillbilchenko I feel confident saying the count cannot be < 0, since the loop above only ever inserts values that are > 0. However, the value can be == 0.

This is because we are now iterating over `db.AllWorkerStates()` and it is quite possible (indeed, expected) that there will be some worker states that do not appear in the DB at the moment this metric is emitted (states like `landing` and `retiring` are intended to be ephemeral anyway and only occur when an operator intervenes).

From https://tour.golang.org/moretypes/22,
> If `key` is not in the map, then `elem` is the zero value for the map's element type.

Thus we can expect `count` to be zero in such cases. The other branch of this conditional is exercised in [this test case](https://github.com/concourse/concourse/pull/5015/files#diff-803b92969cc59c28c60a150bdc3c1b8fR55-R62), validating the need for that check.

Issue #3636, opened 2019-03-29
Concourse Admins should be able to send messages to the instance


Lindsayauchin:
Concourse Admins who operate a big Concourse with many teams often want a way to communicate to everyone that the system is unstable / is recovering. Today, there is no good way to do this other than to use e-mail/slack/smoke signals. 

Now that we have RBAC in place, we can restrict it so that only users with the `Concourse Admin` role have this functionality

Questions
- [ ] How do they write this message?
- [ ] What does it look like?
- [ ] How does an operator turn it off?

StevenArmstrong:
Could this be called something like announcements and have a message banner that is displayed across the top bar. New announcements could be broadcast to all user profiles and then acknowledged by users to turn them off. You could have a selectable icon on the dashboard page to read old read announcements.

marco-m:
> How do they write this message?

A possibility is to use `fly` and call the command `fly wall`, from good old `wall` Unix command line (see https://www.freebsd.org/cgi/man.cgi?query=wall) :-)

taylorsilva:
# Issue Bankruptcy

This issue is being closed as part of 2025's **Issue Bankruptcy** declaration. [See here for details](https://blog.concourse-ci.org/posts/2025-01-06-concourse-2024-review/)

If you think this issue still has merit then feel free to open a new issue and reference this one.

PR #5019, opened 2020-01-13
Add API endpoints for broadcasting a message to a concourse cluster

taylorsilva:
# Existing Issue
Partially addresses #3636

Adds an API endpoint for setting and getting a message meant to be consumed by clients (fly and web UI). The message can only be set by Admins. 

This PR does not include any changes to the fly or web components. Looking for feedback on the backend stuff I've got so far. 

I named everything `wall` because I really liked [@marco-m's comment](https://github.com/concourse/concourse/issues/3636#issuecomment-497979317) and wall is shorter than `announcements` or `broadcast`, though maybe those terms would be better to use. **wdyt?**

- ~`api/v1/info` has a `wall_message` field that will be set if a wall
message is set. The field will not appear if no message has been set.~
  - ~**Why**: The web and fly clients hit this endpoint frequently already (web every few seconds and fly at the start of every command). The thought of including `wall_message` in the info endpoint is to make it easier to implement in the clients. Not sure if this is worth adding more stuff to `info`. Currently info does not hit the database and adding `wall` will make it hit the db every time. **Should this be added to the `info` endpoint or not?**~
- All other wall related set and get's are under the `api/v1/wall` ~and
`api/v1/wall/expiration` endpoints.~
  - `api/v1/wall` requires a `message` (text) field and has an optional `expires_in` (time.Duration) field, allowing the client to set both at the same time. If expires is not set then the wall message will not disappear until either:
    - a DELETE is sent to the same endpoint
    - ~the `api/v1/wall/expiration` endpoint is updated~
  

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


atc/db/wall_test.go
```diff
@@ -0,0 +1,109 @@
...
+	})
+
+	Context("clear the message", func() {
+		It("returns no message", func() {
+			wall.SetMessage(testMessage)
+			msg, err := wall.GetMessage()
+			Expect(err).ToNot(HaveOccurred())
+			Expect(msg).To(Equal(testMessage), "ensure the message has been set")
+
+			wall.Clear()
```
pivotal-jamie-klassen:
```suggestion
			err = wall.Clear()
			Expect(err).ToNot(HaveOccurred())
```
just for consistency?

atc/db/wall_test.go
```diff
@@ -0,0 +1,109 @@
...
+			Context("the expiration is set", func() {
+				It("returns the expiration", func() {
+					wall.SetMessage(testMessage)
+					wall.SetExpiration(time.Minute)
+					expectedExpiration := time.Now().Add(time.Minute)
+
+					expiration, err := wall.GetExpiration()
+					Expect(err).ToNot(HaveOccurred())
+					// Compare time as strings because otherwise they're always off by a few nanoseconds
+					Expect(expiration.Format(time.ANSIC)).To(Equal(expectedExpiration.Format(time.ANSIC)))
```
pivotal-jamie-klassen: I think this test is probably resilient enough for almost all practical purposes, but I'm still pretty fond of externalizing your dependency on time with some kind of `Clock` interface and injecting a fake for cases like this. I know this isn't done super consistently across the codebase (especially because we tend to use the `now()` function in postgres at least as often as we use `time.Now()` in Go... 😬). If you've got the cycles, now (no pun intended) might be the moment to set a standard for externalizing your clocks!

atc/db/wall.go
```diff
@@ -0,0 +1,91 @@
...
+	SetExpiration(time.Duration) error
+	GetExpiration() (time.Time, error)
+	Clear() error
+}
+
+type wall struct {
+	conn Conn
+}
+
+func NewWall(conn Conn) Wall {
```
pivotal-jamie-klassen: For my clock idea below to work, probably something like
```suggestion
func NewWall(conn Conn, clock Clock) Wall {
```
would be called for

pivotal-jamie-klassen:
Just to muddy the waters a bit, I suspect that terminology like `TTL` might make more sense than `expiry`? Because it looks like (even though the DB is concerned with absolute timestamps) the API only speaks in terms of durations (`expires_in`).

atc/api/wallserver/message.go
```diff
@@ -0,0 +1,86 @@
...
+	logger := s.logger.Session("wall")
+
+	var wall atc.Wall
+	err := json.NewDecoder(r.Body).Decode(&wall)
+	if err != nil {
+		logger.Error("failed-to-decode-json", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	err = s.wall.SetMessage(wall.Message)
```
clarafu: Maybe we can have the `SetMessage` be a `SetWall` and combine the inserting of both the message and the expiration in one query? We are trying to have API queries use only one transaction in order to speed up the responses during times where the environment is pegged for connections, but it's definitely not a deal breaker.

clarafu:
I think adding one simple query to run every few seconds (whether its from the info or some other endpoint) won't have much affect on the performance of the system but adding it to the `info` endpoint might cause it the endpoint to be less snappy when the deployment is being pegged for connections.

To what @pivotal-jamie-klassen said about setting the expiration bug, I think that's a good point. I'm also not sure if we have already determined that we want to support the use case of having someone be able to set an expiration after they have already set the message?

atc/api/wallserver/message.go
taylorsilva: I'm down for that optimization

pivotal-jamie-klassen: I think it would be a great idea to simplify the `db.Wall` interface down to `Get` `Set` and `Clear`. Simple and CRUDdy

taylorsilva:
@pivotal-jamie-klassen @clarafu Incorporated the feedback you two gave.

- Consolidated the DB interface to three functions: SetWall, GetWall,
Clear. Set and Get take in and return atc.Wall
- Renamed atc.Wall.ExpiresIn to `TTL` which is readily recognized across
the industry
- Removed expiry API endpoints
- Removed `wall_message` from info endpoint

I'm going to work on adding the time/clock interface tomorrow for the db layer.

atc/db/wall.go
```diff
@@ -24,62 +23,87 @@ func NewWall(conn Conn) Wall {
...
+	}
+
+	query := psql.Insert("wall").
+		Columns("message")
+
+	if wall.TTL != "" {
+		ttl, err := time.ParseDuration(wall.TTL)
+		if err != nil {
+			return err
+		}
```
pivotal-jamie-klassen: It's hard to feel ethical rolling back a transaction because I failed to deserialize some data - feels a bit like cancelling a meeting because I have to go to the bathroom.

I think that the `TTL` field on `atc.Wall` should be a `time.Duration` instead of a string and this logic should get moved into the API layer, since it's really a serialization/validation issue.

atc/db/wall.go
```diff
@@ -24,62 +23,87 @@ func NewWall(conn Conn) Wall {
...
-			return expiration, nil
-		}
-		return expiration, err
+	if message.Valid {
+		wall.Message = message.String
+	}
+
+	if expiresAt.Valid {
+		ttl := time.Until(expiresAt.Time)
+		wall.TTL = ttl.Truncate(time.Second).String()
```
pivotal-jamie-klassen: ... then if TTL is a Duration you don't have to do this suspicious rounding.

atc/db/wall.go
```diff
@@ -0,0 +1,115 @@
...
+	)
+
+	err := scan.Scan(&message, &expiresAt)
+	if err != nil {
+		return err
+	}
+
+	if message.Valid {
+		wall.Message = message.String
+	}
```
pivotal-jamie-klassen: There's a not null constraint on the message column, right? I think you should be able to change `message` in this context to just be a string and then avoid this check since it's enforced by postgres?

atc/db/wall.go
taylorsilva: the bathroom analogy won me over 😆

taylorsilva:
@pivotal-jamie-klassen added a clock interface. Interested in what you think :)

atc/api/handler.go
```diff
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"github.com/concourse/concourse/atc/api/wallserver"
```
cirocosta: nit: we tend to separate the imports, keeping standard lib ones on the top, and                                                                                                      [0/1]
all-else in the bottom :grin:


```diff
diff --git a/atc/api/handler.go b/atc/api/handler.go
index 3be7caaa6..dc12a0a95 100644
--- a/atc/api/handler.go
+++ b/atc/api/handler.go
@@ -1,12 +1,9 @@
 package api

 import (
-       "github.com/concourse/concourse/atc/api/wallserver"
        "net/http"
        "path/filepath"

-       "github.com/concourse/concourse/atc/api/usersserver"
-
        "code.cloudfoundry.org/lager"
        "github.com/concourse/concourse/atc"
        "github.com/concourse/concourse/atc/api/artifactserver"
@@ -23,7 +20,9 @@ import (
        "github.com/concourse/concourse/atc/api/resourceserver"
        "github.com/concourse/concourse/atc/api/resourceserver/versionserver"
        "github.com/concourse/concourse/atc/api/teamserver"
+       "github.com/concourse/concourse/atc/api/usersserver"
        "github.com/concourse/concourse/atc/api/volumeserver"
+       "github.com/concourse/concourse/atc/api/wallserver"
        "github.com/concourse/concourse/atc/api/workerserver"
        "github.com/concourse/concourse/atc/creds"
        "github.com/concourse/concourse/atc/db"
```diff


e.g.: https://github.com/concourse/concourse/pull/4825/commits/c757369352f9bf1e79d9567f2e36611c7af27c97

File: atc/api/wall_test.go [hunk 9]
```diff
@@ -0,0 +1,185 @@
+package api_test
+
+import (
+	"bytes"
+	"encoding/json"
+	"github.com/concourse/concourse/atc"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
```
cirocosta: nit: imports could be separated

```diff
diff --git a/atc/api/wall_test.go b/atc/api/wall_test.go
index 4ed46ec18..a787d0e40 100644
--- a/atc/api/wall_test.go
+++ b/atc/api/wall_test.go
@@ -3,12 +3,14 @@ package api_test
 import (
        "bytes"
        "encoding/json"
-       "github.com/concourse/concourse/atc"
-       . "github.com/onsi/ginkgo"
-       . "github.com/onsi/gomega"
        "io/ioutil"
        "net/http"
        "time"
+
+       "github.com/concourse/concourse/atc"
+
+       . "github.com/onsi/ginkgo"
+       . "github.com/onsi/gomega"
 )

 var _ = Describe("Wall API", func() {
```

atc/api/wallserver/wall.go
```diff
@@ -0,0 +1,55 @@
+package wallserver
+
+import (
+	"encoding/json"
+	"github.com/concourse/concourse/atc"
```
cirocosta: nit: imports 😁

atc/db/db_suite_test.go
```diff
@@ -1,6 +1,7 @@
 package db_test
 
 import (
+	"github.com/concourse/concourse/atc/db/dbfakes"
```
cirocosta: nit: import 😁

taylorsilva:
I did a refactor of the db tests to reduce the number of assertions per test. Was 4-6 previously and got it down to 1-3 per spec. Cleaned up the imports and ran `go fmt`. Let me know if I missed anything. Thanks for all the feedback @clarafu @pivotal-jamie-klassen @cirocosta 😄

marco-m-pix4d:
@taylorsilva for my understanding, is anything using this endpoint ? I am interested especially in the web UI being able to display it. Thanks.

taylorsilva:
@marco-m-pix4d Nothing is using this endpoint right now. Never got around to figuring out what the web UI would look like. It is on the roadmap but it has a low priority so it's likely to keep getting bumped down for more urgent work that pops up.
https://github.com/concourse/concourse/issues/5714

marco-m-pix4d:
OK thanks!

PR #5023, opened 2020-01-14
web: dashboard no longer piles on slow API requests

### LINKED ISSUE #4885: Don't pile-on slow API requests when polling against the dashboard

zoetian:
# Existing Issue
Fixes #4885

Split up the APIData concept into each api call, if the previous fetch hasn't finished yet, there won't be any new fetch. As we made the changes, we also broke the `drag and drop`; So we ended up refactoring it as well. 

// Note: for acceptance criteria test, you may need to rebuild the backend after applying a patch that forces the jobserver to be slow, like
```diff
diff --git a/atc/api/jobserver/list_all.go b/atc/api/jobserver/list_all.go
index 392cc58ad..a34ba4237 100644
--- a/atc/api/jobserver/list_all.go
+++ b/atc/api/jobserver/list_all.go
@@ -3,6 +3,7 @@ package jobserver
 import (
 	"encoding/json"
 	"net/http"
+	"time"
 
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/api/accessor"
@@ -13,6 +14,8 @@ import (
 func (s *Server) ListAllJobs(w http.ResponseWriter, r *http.Request) {
 	logger := s.logger.Session("list-all-jobs")
 
+	time.Sleep(6 * time.Second)
+
 	acc := accessor.GetAccessor(r)
 
 	var dashboard db.Dashboard
```
We suggest this rather than filling up your database, which could be pretty time-consuming and delicate.

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


jomsie:
This is a biggie! I'd like to take a look and review it, but I'm out for a couple of days starting in about an hour, so either I can review it Tuesday or someone else might wanna get in on this?

Sorry :| Just saw it now :|

cirocosta:
Btw, would you mind giving a rough description of the approach?

I saw the addition of this "RequestBuffer", which seems interesting 😁  

Thanks!

zoetian:
> Btw, would you mind giving a rough description of the approach?
> 
> I saw the addition of this "RequestBuffer", which seems interesting 😁
> 
> Thanks!

Good point! 

Our end goal here is to make sure on a 5 second clock tick, the dashboard fires a new request only when the previous request is returned (so we won't pile up all these api requests) and update the ui when the model is not paused.
```
case delivery of
        ClockTicked FiveSeconds _ ->
            ( if isPaused model then
                shouldFire.set True model

              else
                shouldFire.set False model
            , if shouldFire.get model && not (isPaused model) then
                effect :: effects

              else
                effects
            )
```
We added another 4 fields on the dashboard model related to these api requests
`isJobsRequestFinished`, `isTeamsRequestFinished `, `isResourcesRequestFinished ` and `isPipelinesRequestFinished `.

We also extracted out this new module called `RequestBuffer` where `buffers` take in a list of Buffers with model, then we case of each callback that buffer takes in, if the callback matches with the effect, then it returns true (see the buffer explanation below); so when the dragState of the Dashboard model is being dragged, then change the result of `get` and `set` to update the model's field (i.e. `isTeamsRequestFinished` in this case).

```
buffers : List (Buffer Model)
buffers =
    [ Buffer FetchAllTeams
        (\c ->
            case c of
                AllTeamsFetched _ ->
                    True

                _ ->
                    False
        )
        (.dragState >> (/=) NotDragging)
        { get = \m -> m.isTeamsRequestFinished
        , set = \f m -> { m | isTeamsRequestFinished = f }
        }
...
```

`Buffer` is new type (also a function) we defined to take in an effect and returns a `function` that takes a callback and returns a bool; and if the condition hold here, it will update the record for `get` and `set` field here.

```
type Buffer a
    = Buffer Effect (Callback -> Bool) (a -> Bool) { get : a -> Bool, set : Bool -> a -> a }
```

PR #5026, opened 2020-01-14
Use all plans (nested ones) instead of only top level plan

clarafu:
Anything that loops over `(atc.JobConfig).Plan` is *probably* subtly broken, as it'll only be looping at the top level values and not recursing into `in_parallel`, job-level hooks, etc.

Instead everything should use `.Plans()` which collects all the build plans recursively into a flat slice.

There were two places where we found this was broken:

1. In `removeUnusedWorkerTaskCaches `
1. In `validateCredParams`

### DIFF
--- a/atc/job_config.go
+++ b/atc/job_config.go
@@ -20,7 +20,11 @@ type JobConfig struct {
 	Ensure  *PlanConfig `json:"ensure,omitempty"`
 	Success *PlanConfig `json:"on_success,omitempty"`
 
-	Plan PlanSequence `json:"plan"`
+	// The parent plan is only the TOP LEVEL plan in the job config. If there are
+	// any nested plans in the job config, the parent plan does not cover it!
+	//
+	// Use Plans() if you need all the plans within the job config.
+	ParentPlan PlanSequence `json:"plan"`
 }
 
 type BuildLogRetention struct {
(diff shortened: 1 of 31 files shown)


atc/job_config.go
```diff
@@ -20,7 +20,11 @@ type JobConfig struct {
 	Ensure  *PlanConfig `json:"ensure,omitempty"`
 	Success *PlanConfig `json:"on_success,omitempty"`
 
-	Plan PlanSequence `json:"plan"`
+	// The parent plan is only the TOP LEVEL plan in the job config. If there are
+	// any nested plans in the job config, the parent plan does not cover it!
+	//
+	// Use Plans() if you need all the plans within the job config.
+	ParentPlan PlanSequence `json:"plan"`
```
vito: I'm lukewarm on the name since it makes it sound like the plan is a parent of the job config (i.e. `(Volume).ParentVolume`). But I can't really think of a name that's intuitive enough in the happy path and clear enough to guide against misuse without being annoying to type out.

Maybe instead of renaming it to `ParentPlan` we can keep it as `Plan` and change the type to `PlanConfig`? This way it's more obvious that you shouldn't be looping over it directly, because the type won't allow it. We actually turn it into a `PlanConfig{Do: &config.Plan}` in a few places already, so this would let us skip that step.

vito:
superceded by #5497

PR #5033, opened 2020-01-15
use ImportStrategy when step output is cow volume

### LINKED ISSUE #4964: testflight's test for mounting input in the same place as output fails to materialize volume
### Bug Report

As observed by @vito and @clarafu, `testflight` running on nci.concourse-ci.org
recently started having the test where a task has the path of an `input` the
same as an `output`:

https://github.com/concourse/concourse/blob/f8ae6998684f05a1f93e0b9b3be03ab6f20007d5/testflight/volume_mounting_test.go#L39-L43

https://github.com/concourse/concourse/blob/f8ae6998684f05a1f93e0b9b3be03ab6f20007d5/testflight/fixtures/volume-mounting.yml#L175-L180

https://nci.concourse-ci.org/teams/main/pipelines/prs/jobs/testflight/builds/283#L5e0e4e39:1220

The timing of the failures (since this morning - 2020/01/03) seem to be quite
related to the moment of the bump from GKE 1.14.x to 1.15.4 which incurred in a
bump from Linux 4.15.x to 5.0.0.

Looking inside the task and inspecting the logs from the worker:

```json
{
  "timestamp": "2020-01-03T18:42:15.750634604Z",
  "level": "error",
  "source": "baggageclaim",
  "message": "baggageclaim.api.volume-server.create-volume-async.create-volume.failed-to-materialize-strategy",
  "data": {
    "error": "device or resource busy",
    "handle": "4af3ce40 -b08e-4481-6532-c63afd635f44",
    "session": "3.1.4500.1"
  }
}
```

```json
{
  "timestamp": "2020-01-03T18:42:15.750705655Z",
  "level": "error",
  "source": "baggageclaim",
  "message": "baggageclaim.api.volume-server.create-volume-async.failed-to-create",
  "data": {
    "error": "device or resource busy",
    "handle": "4af3ce40-b08e-4481-6532-c63afd635f44 ",
    "privileged": false,
    "session": "3.1.4500",
    "strategy": {
      "type": "cow",
      "volume": "a6ad87cf-6162-44cb-5c80-2fb8854b2e97"
    }
  }
}
```

### Steps to Reproduce

"reproduce":

1. run the `testflight` job in the pipeline
  - this will have the container running in our new linux 5.0-based machines on top
     of a privileged kubernetes pod
2. once it fails, hijack into it, run the docker daemon, and then run the test.

### Expected Results

test succeeds

### Actual Results

test fails

### Additional Context

- the test passes on Docker for Mac (Docker version 19.03.5, build 633a0ea) Linux
 4.19.76-linuxkit .
- the test passes on Linux 5.3.0-1009-gcp
- setting the same job in `nci` and `hh` then triggering the job (that fails to
  run on `testflight` works flawlessly

~my *guess* is that this is coming from a `mount(2)` that `runc` is trying to perform
when it tries to go from the OCI runtime spec to a container.~ err, this is from `baggageclaim`, so, no `runc` involved.

being from `mount(2)`, `EBUSY` would then interpreted as follows:

```

       EBUSY  An attempt was made to stack a new mount directly on top of an
              existing mount point that was created in this mount namespace
              with the same source and target.

```

~and yeah, indeed, we're trying to perform a stack mount there (as both input and
output would be mounted in the same directory, one would be "stacked" upon the
other).~

~what doesn't seem to make a lot of sense if the "with same source and target" -
why in `testflight` that'd be the case (assuming this is the problem)?~

### Version Info

* Concourse version: master (f8ae6998684f05a1f93e0b9b3be03ab6f20007d5)
* Deployment type (BOSH/Docker/binary): k8s on GKE (v1.15.4-gke.22)
  - `Linux gke-hush-house-ci-workers-79a0ea06-bkbz 5.0.0-1022-gke #22~18.04.3-Ubuntu SMP` Thu Oct 10 19:41:43 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
* Infrastructure/IaaS: GCP
* Browser (if applicable):
* Did this used to work? yes

kcmannem:

Fixes #4964  .

If you have a step that uses an output and input at the same mount path. The output will end up being a COW of the resource_cache volume. When this output is then used as an input to the next step in the plan, we try to create a COW of a COW(output volume of prev step). In newer versions of the kernel, this isn't something that is supported. With this change we can now achieve the same behavior by making a baggageclaim request using the ImportStrategy. Which will create a copy of the output volume by executing a `cp` command.
https://github.com/concourse/baggageclaim/blob/dc362b1f055e4836dc2afb16639383c61cee637a/volume/import_strategy.go#L29

We have a testflight test which catches this user behavior:
```
- step1:
    - input: some-resource
    - output: some-resource
- step2:
    - input: some-resource
```
https://github.com/concourse/concourse/blob/master/testflight/fixtures/volume-mounting.yml#L164

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/worker/volume.go
+++ b/atc/worker/volume.go
@@ -14,7 +14,6 @@ import (
 type Volume interface {
 	Handle() string
 	Path() string
-	ParentHandle() string
 
 	SetProperty(key string, value string) error
 	Properties() (baggageclaim.VolumeProperties, error)
@@ -24,8 +23,7 @@ type Volume interface {
 	StreamIn(ctx context.Context, path string, tarStream io.Reader) error
 	StreamOut(ctx context.Context, path string) (io.ReadCloser, error)
 
-	COWStrategy() baggageclaim.COWStrategy
-	ImportStrategy() baggageclaim.ImportStrategy
+	CopyStrategy() baggageclaim.Strategy
 
 	InitializeResourceCache(db.UsedResourceCache) error
 	InitializeTaskCache(logger lager.Logger, jobID int, stepName string, path string, privileged bool) error
@@ -110,16 +108,19 @@ func (v *volume) Destroy() error {
 	return v.bcVolume.Destroy()
 }
 
-func (v *volume) COWStrategy() baggageclaim.COWStrategy {
+// NOTE (runtime/#4964): in Linux kernel 5.x and greater,
+// We lose the ability the make a COW of a COW. Overlay will return an error.
+// The only time we do this is when the output of a prev step a COW and the
+// input we clone here tries to become a COW of it.
+func (v *volume) CopyStrategy() baggageclaim.Strategy {
+	if v.ParentHandle() != "" {
+		return baggageclaim.ImportStrategy{
+			Path:           v.bcVolume.Path(),
+			FollowSymlinks: true,
+		}
+	}
 	return baggageclaim.COWStrategy{
 		Parent: v.bcVolume,
-	}
-}
-
-func (v *volume) ImportStrategy() baggageclaim.ImportStrategy {
-	return baggageclaim.ImportStrategy{
-		Path: v.bcVolume.Path(),
-		FollowSymlinks: true,
 	}
 }
 
--- a/atc/worker/volume.go
+++ b/atc/worker/volume.go
@@ -14,6 +14,7 @@ import (
 type Volume interface {
 	Handle() string
 	Path() string
+	ParentHandle() string
 
 	SetProperty(key string, value string) error
 	Properties() (baggageclaim.VolumeProperties, error)
@@ -24,6 +25,7 @@ type Volume interface {
 	StreamOut(ctx context.Context, path string) (io.ReadCloser, error)
 
 	COWStrategy() baggageclaim.COWStrategy
+	ImportStrategy() baggageclaim.ImportStrategy
 
 	InitializeResourceCache(db.UsedResourceCache) error
 	InitializeTaskCache(logger lager.Logger, jobID int, stepName string, path string, privileged bool) error
@@ -76,6 +78,10 @@ func (v *volume) Handle() string { return v.bcVolume.Handle() }
 
 func (v *volume) Path() string { return v.bcVolume.Path() }
 
+func (v *volume) ParentHandle() string {
+	return v.dbVolume.ParentHandle()
+}
+
 func (v *volume) SetProperty(key string, value string) error {
 	return v.bcVolume.SetProperty(key, value)
 }
@@ -107,6 +113,13 @@ func (v *volume) Destroy() error {
 func (v *volume) COWStrategy() baggageclaim.COWStrategy {
 	return baggageclaim.COWStrategy{
 		Parent: v.bcVolume,
+	}
+}
+
+func (v *volume) ImportStrategy() baggageclaim.ImportStrategy {
+	return baggageclaim.ImportStrategy{
+		Path: v.bcVolume.Path(),
+		FollowSymlinks: true,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


atc/worker/worker.go
```diff
@@ -570,6 +577,15 @@ func (worker *gardenWorker) cloneLocalVolumes(
 	return mounts, nil
 }
 
+func avoidCOWCOWVolume(input mountableLocalInput) baggageclaim.Strategy {
```
kcmannem: was a pretty simple fix.

kcmannem:
other option is to cp in here the contents of the path specified by `parent` https://github.com/concourse/baggageclaim/blob/master/volume/driver/overlay_linux.go#L79-L84

^ That one is probably the easier way to fix it. I'm tending towards getting rid of this pr.

atc/worker/worker.go
```diff
@@ -549,11 +549,11 @@ func (worker *gardenWorker) cloneLocalVolumes(
 		inputVolume, err := worker.volumeClient.FindOrCreateCOWVolumeForContainer(
```
kcmannem: Note: We do a COW specific find or create here. This ensures that a parent handle gets stored in the child volumes db row. I don't think there's any side effects if we use an import strategy to copy over but still consider returned `inputVolume` a child.

atc/worker/volume.go
```diff
@@ -104,7 +108,17 @@ func (v *volume) Destroy() error {
 	return v.bcVolume.Destroy()
 }
 
-func (v *volume) COWStrategy() baggageclaim.COWStrategy {
+// NOTE (runtime/#4964): in Linux kernel 5.x and greater,
+// We lose the ability the make a COW of a COW. Overlay will return an error.
```
cirocosta: Hey,

Thanks for the PR!

> in Linux kernel 5.x and greater, we lose the ability the make a COW of a COW.
> Overlay will return an error.

reading through the details of how `overlayfs` works (see [overlayfs entry]), it
turns out that this is not really true (I originally thought it was, but there's
a catch), what can't be done is to have a `cow` of a `cow` where the `lowerdir`
of the "cow-cow" references an `upperdir` used somewhere else.

[overlayfs entry]: https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt

	FAIL

		mount \
			-t overlay \
			cow1 \
			-o lowerdir=./cow1-lower,upperdir=./cow1-upper,workdir=./cow1-work \
			./cow1-merged

		mount \
			-t overlay \
			cow1-cow \
			-o lowerdir=./cow1-upper,upperdir=./cow1-cow-upper,workdir=./cow1-cow-work \
			./cow1-cow-merged   |
					    |
					    '--> using an upper: PROBLEM!


	SUCCESS

		mount \
			-t overlay \
			cow1 \
			-o lowerdir=./cow1-lower,upperdir=./cow1-upper,workdir=./cow1-work \
			./cow1-merged

		mount \
			-t overlay \
			cow1-cow \
			-o lowerdir=./cow1-merged,upperdir=./cow1-cow-upper,workdir=./cow1-cow-work \
			./cow1-cow-merged   |
					    | 
					    '--> not an upper: just fine :)


I think we might be able to fix the problem that we currently have by changing
the way that we deal with ancestors :thinking:

I'll first give a try at changing the code in baggageclaim to see if I can get
anywhere, then come back to this one :+1: I'm 99% sure we can get this fixed
over there with very little code changes.

Thanks!

cirocosta: created https://github.com/concourse/baggageclaim/pull/35

kcmannem: SMART

kcmannem: should I close this pr?

cirocosta: yeaah 😁  thanks for the PR, though!

PR #5034, opened 2020-01-16
Policy enhancement - OPA integration

evanchaoli:
This PR is to integrate Concourse with OPA to enable policy enforcement ability.

The code change is based on RFC https://github.com/concourse/rfcs/pull/41

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


evanchaoli:
This calls for an initial review from the design perspective. Tests will be added later.

atc/api/policychecker/handler.go
```diff
@@ -0,0 +1,49 @@
...
+	policyChecker policy.PreChecker,
+) http.Handler {
+	return policyCheckerHandler{
+		handler:       handler,
+		action:        action,
+		policyChecker: policyChecker,
+	}
+}
+
+type policyCheckerHandler struct {
```
pivotal-jamie-klassen: Another English nit - if something is called a `thingHandler` I would expect it to handle `thing`s. So I would expect a `policyCheckerHandler` to handle policycheckers. But this struct is called a `Handler` because it fulfills the `http.Handler` interface!
```suggestion
type policyCheckingHandler struct {
```

atc/api/policychecker/handler.go
```diff
@@ -0,0 +1,49 @@
...
+	handler       http.Handler
+	action        string
+	policyChecker policy.PreChecker
+}
+
+func (h policyCheckerHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	acc := accessor.GetAccessor(r)
+	ctx := context.WithValue(r.Context(), "accessor", acc)
+
+	pass, err := h.policyChecker.CheckHttpApi(h.action, acc.UserName(), r)
```
pivotal-jamie-klassen: Instead of getting the accessor and extracting the username here, could we move that logic into the implementation of `CheckHttpApi`? Just allows for fewer parameters/smaller interface.
```suggestion
	pass, err := h.policyChecker.CheckHttpApi(h.action, r)
```

atc/policy/checker.go
```diff
@@ -0,0 +1,224 @@
...
+	realChecker Checker
+}
+
+func (c *preChecker) CheckHttpApi(action string, user string, req *http.Request) (bool, error) {
+	if c == nil {
+		return true, nil
+	}
+
+	// Ignore self invoked API calls.
+	if user == "system" {
```
pivotal-jwinters: https://github.com/concourse/concourse/pull/4950 changes the definition of what we consider to be a "system" request. 

These values are now configurable:
https://github.com/concourse/concourse/blob/0a2bff940f795bccb0d7c02d6c63e6f90456220f/atc/api/accessor/accessor.go#L59-L60

So we may have to do something similar here.

atc/api/policychecker/handler.go
evanchaoli: @pivotal-jamie-klassen I don't get this comment. To implement extracting username in `CheckHttpApi`, we have to pass `acc`, then it still have 3 parameters.

evanchaoli:
@pivotal-jamie-klassen Thanks for all of the naming suggestions. For `PreChecker`, I have taken option 1. Please take a look again. If you have no more big comments, I'll start to add unit tests.

atc/policy/checker.go
```diff
@@ -0,0 +1,225 @@
...
+	if len(f.Actions) == 1 {
+		f.Actions = strings.Split(f.Actions[0], ",")
+	}
+
+	if len(f.ActionsToSkip) == 1 {
+		f.ActionsToSkip = strings.Split(f.ActionsToSkip[0], ",")
+	}
+
+	return f
+}
```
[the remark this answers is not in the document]
evanchaoli: @pivotal-jamie-klassen  Each CLI option has an environment variable alternative, there is no way to specify multiple values with a single environment variable, so this logic is for that:

```
CONCOURSE_POLICY_CHECK_FILTER_ACTION_SKIP="RunTask,GetConfig"
concourse web ...
```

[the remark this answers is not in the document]
pivotal-jamie-klassen: @evanchaoli actually there is a way to specify multiple values in a single environment variable! https://github.com/vito/twentythousandtonnesofcrudeoil/blob/master/environment.go#L20-L23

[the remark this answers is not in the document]
pivotal-jamie-klassen: wow I totally thought that github.com/vito/twentythousandtonnesofcrudeoil handled this for us, but reading through the code now, I think you're right...

atc/engine/builder/builder.go
```diff
@@ -33,7 +34,7 @@ type StepFactory interface {
 type DelegateFactory interface {
 	GetDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.GetDelegate
 	PutDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.PutDelegate
-	TaskDelegate(db.Build, atc.PlanID, vars.CredVarsTracker) exec.TaskDelegate
+	TaskDelegate(db.Build, atc.PlanID, vars.CredVarsTracker, policy.Checker) exec.TaskDelegate
```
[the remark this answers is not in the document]
evanchaoli: @pivotal-jwinters Can you please be more specific? If you want to apply policy check to all steps?

evanchaoli:
I have just rebased and resolved conflicts. Unit tests have been added for the new code.

atc/api/policychecker/handler.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		rejector := NotPassRejector{msg: err.Error()}
+		rejector.Fail(w, r)
+		return
+	}
+	if !pass {
+		rejector := NotPassRejector{}
+		rejector.NotPass(w, r)
+		return
+	}
```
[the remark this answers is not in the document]
evanchaoli: I think I following `atc\api\auth\unauthorized_rejector.go`.

atc/engine/builder/builder.go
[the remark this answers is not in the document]
pivotal-jwinters: Didn't you outline some of these things in your RFC? Like being able to restrict where an image gets pulled from? Since an image can come from get step we would need to enforce these policies at that level.

I'm also curious why the policy.Checker gets passed through the delegate. Could we not pass it directly to the step itself?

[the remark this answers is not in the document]
evanchaoli: @pivotal-jwinters  I think I explained in the RFC: https://github.com/concourse/rfcs/pull/41/files#diff-920d7a06e64dac5ea2673094189fa570R38. With `set-pipeline` API, everything about a pipeline, expect task files, are available to check. So docker-images used for `get` and `put` can be checked at `set-pipeline`. Because task files are only fetched at runtime, to check images used in `task` step, I have to add a check point in `task` step.

I received a comment for a previous PR: https://github.com/concourse/concourse/pull/4708#pullrequestreview-311954465 from you that db related operation should be delegate, so I think policy related operations should be similar. Is there a guide for what should be in delegate and what should be in step itself?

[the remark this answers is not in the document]
pivotal-jwinters: Oh so you want to send the entire set-pipeline payload to be verified by the policy server? That means the policy server will need to know how to interpret a pipeline yaml? Wouldn't it be a lot easier to make these policy decisions at the get/put level instead?

There's no guide, but the way i look at it is - anything that controls the flow of logic should be within the step itself, and anything that deals with side effects (stdout, saving versions) should be in the delegate. So the delegate ends up dealing with a lot of db operations.

Overall the exec steps are very much a work in progress. In the past we've found that making large changes to this part of the code is very difficult (in part) because the steps are all very different. Over time we've been trying to align these steps to be more consistent, so these one off changes tend to raise red flags for me.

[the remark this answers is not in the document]
evanchaoli: > Oh so you want to send the entire set-pipeline payload to be verified by the policy server? 

Yes, all API calls' payload will be sent to the policy server, unless filtered (there are some filters to prevent from some APIs from going through the policy server.)

> That means the policy server will need to know how to interpret a pipeline yaml? Wouldn't it be a lot easier to make these policy decisions at the get/put level instead?

No, that's not true. To the policy server, every input is a JSON. So checking an entire pipeline is almost same easy as checking a task config.

But after thinking over, your comment is valid from the other perspective. Every get and put should also go through policy check, so that policy change may immediately take effective.  Otherwise if we only do policy check for `SetPipeline`, when policy is changed, say some image is no longer allowed to use, the policy change will not be able to apply to existing pipelines.

For that, I'll update the code as well as the RFC.

> There's no guide, but the way i look at it is - anything that controls the flow of logic should be within the step itself, and anything that deals with side effects (stdout, saving versions) should be in the delegate. So the delegate ends up dealing with a lot of db operations.
> 
> Overall the exec steps are very much a work in progress. In the past we've found that making large changes to this part of the code is very difficult (in part) because the steps are all very different. Over time we've been trying to align these steps to be more consistent, so these one off changes tend to raise red flags for me.

OK, I'll move policy check to step.

gowrisankar22:
We are also expecting this feature.

atc/engine/builder/builder.go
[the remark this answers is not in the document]
evanchaoli: @pivotal-jwinters I have pushed a new commit that moved `polichChecker` from delegate to step, and added policy check to `get` and `put`. I renamed the action name "RunTask" to "UsingImage", if you have a better name, feel free to advice. Please take a look again and see if you are ok with the new change.

evanchaoli:
@taylorsilva Just saw that you are current assignee of this PR, so can you please help move forward this PR?

atc/policy/opa/opa.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+type opa struct {
+	config OpaConfig
+	logger lager.Logger
+}
+
+func (c opa) Check(input policy.PolicyCheckInput) (bool, error) {
+	data := opaInput{input}
+	jsonBytes, err := json.Marshal(data)
```
pivotal-jwinters, line 47: Need to check this error.

atc/policy/opa/opa.go
```diff
@@ -0,0 +1,85 @@
...
+
+	c.logger.Debug("opa-check", lager.Data{"input": string(jsonBytes)})
+
+	req, err := http.NewRequest("POST", c.config.URL, bytes.NewBuffer(jsonBytes))
```
pivotal-jwinters, line 51: This one too.

atc/api/policychecker/handler.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+func (h policyCheckingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	acc := accessor.GetAccessor(r)
+	ctx := context.WithValue(r.Context(), "accessor", acc)
+
+	if h.policyChecker != nil {
+		pass, err := h.policyChecker.CheckHttpApi(h.action, acc, r)
+		if err != nil {
+			rejector := NotPassRejector{msg: err.Error()}
```
pivotal-jwinters, line 40: This rejector pattern is not very widely used across the api package, and it's something we're slowly moving away from, so if you could just inline the response as per @pivotal-jamie-klassen [suggestion](https://github.com/concourse/concourse/pull/5034#discussion_r370202935).

atc/exec/task_step_test.go
```diff
@@ -81,6 +81,7 @@ var _ = Describe("TaskStep", func() {
 		fakeDelegate.VariablesReturns(credVarsTracker)
 		fakeDelegate.StdoutReturns(stdoutBuf)
 		fakeDelegate.StderrReturns(stderrBuf)
+		fakeDelegate.PolicyCheckReturns(true, nil)
```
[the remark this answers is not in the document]
evanchaoli: Good catch!

evanchaoli:
@pivotal-jwinters I have addressed all of your comments, please review again. I'll add more unit tests in next commit.

evanchaoli:
I realized that I should add some usage explanation. Let me add a simple use case for sample.

_P.S. I'm not going to explain usage of OPA here, please refer to https://www.openpolicyagent.org/ for details of OPA._

First, I need to start an OPA server. For testing purpose, it's very easy to start a OPA server on my MacBook, opa is a single binary, I use the following command line:

```
$ ./opa run -l debug -f json -s -w ./opa_concourse/
```

Where `-w` option specifies a directory that stores policy rules, under my folder, I just put a very simple rule file:

```
package concourse
default allow = false
allow = true {
    input.action == "ListContainers"
}
```
The rule file is written in `Rego`, if you don't know `Rego`, that's fine, it's easy to understand that this rule defines that only `ListContainers` action is allowed. _This sample doesn't make much real case, I just want to demo the usage._

Then let's configure Concourse to integrate with the OPA. I added the following options to ATC:

```
CONCOURSE_OPA_URL: http://192.168.1.15:8181/v1/data/concourse/allow
CONCOURSE_POLICY_CHECK_FILTER_HTTP_METHODS: PUT,POST
CONCOURSE_POLICY_CHECK_FILTER_ACTION: ListWorkers,ListContainers
CONCOURSE_POLICY_CHECK_FILTER_ACTION_SKIP: PausePipeline,UnpausePipeline
```
In OPA URL, you can see that, `concourse` is the package name I define in the rule file, and `allow` is a variable I define in the rule file.

For `CONCOURSE_POLICY_CHECK_FILTER_HTTP_METHODS`, I define that only `PUT` and `POST` APIs will go through policy check, so that read-only actions, such as `ListAllPipelines` will not go through policy check. But if you want them to be checked, you may add `GET`.

For `CONCOURSE_POLICY_CHECK_FILTER_ACTION`, I add two `GET` actions `ListWorkers` and `ListContainers` as exceptions. Actions defined in this list will always go through policy check. This makes an use case easier that, say, you want only apply policy check to SavePipeline, then you may only define `CONCOURSE_POLICY_CHECK_FILTER_ACTION: SaveConfig`. _(SaveConfig is set pipeline.)_

For `CONCOURSE_POLICY_CHECK_FILTER_ACTION_SKIP`, you may define some actions that will always skip policy check. In this sample, I define `PausePipeline` and `UnpausePipeline`.

Then let's see what will happen. As the policy rule, only `ListContainer` is allowed:

```
$ fly -t dev workers # ListWorkers doesn't pass policy check, thus forbidden
error: forbidden

$ fly -t dev containers # ListContainer passes policy check, thus goes well
handle                                worker        pipeline  job   build #  build id  type   name  attempt
4cdac52d-3828-4e45-65ad-a1e3bf458c62  b0b10dd794f8  none      none  none     none      check  none  n/a

$ fly -t dev sp -c simple.yml -p s2 # SaveConfig doesn't pass policy check, thus forbidden
.. diffs ...
apply configuration? [yN]: y
error: forbidden

$ fly -t dev up -p simple # UnpausePipeline skips policy check, thus goes well
unpaused 'simple'
```

All actions can be configured in RBAC can apply to policy check. For a full list of actions, refer to https://concourse-ci.org/user-roles.html#action-matrix.

At OPA server side, it will receive policy check request with the following data:

```json
{"input":
    {"service": "concourse", 
      "cluster_name": "dev",
      "cluster_version": "0.0.0-dev",
       "http_method": "PUT",
       "action": "SaveConfig",
       "user": "test", 
       "team": "main",
       "pipeline": "s2",
       "data":{
           "jobs":[
              {"name": "test",
               "plan":[
                   {"get":
                    "tiny"},
                  {"config":{
                       "image_resource": {
                            "source":{"repository":"busybox"},
                            "type":"registry-image"},
                       "platform":"linux",
                       "run":{
                            "args": ["-exc","echo hello8"],
                             "path": "sh"}},
                   "task":"a-task"}]}
```

As you can see, the `input` includes a set of metadata of the action, and action specific `data` section. For this `SaveConfig` action, `data` is pipeline content. Based on `input`, you may write whatever policy rules, simple or complicated.

Besides API actions, there is a special action `UsingImage`, that can be policy checked before Concourse is going to start a container for `check/get/put/task`. `input` of `UsageImage` is like:

```json
"input": {
    "service": "concourse",
    "cluster_name": "dev",
    "cluster_version":"0.0.0-dev",
    "action": "UsingImage",
     "team": "main",
     "pipeline": "simple",
     "data":{
         "repository": "evanli/tiny2",
         "step": "check",
         "tag": "v4"
     }}}
```

evanchaoli:
@jamieklassen Please help move forward this PR. I really want it to be included in 6.2.0. I understand it takes time to merge this kind of big features, but it has been more than 4 months, and there is no more comment to address. Thanks a lot.

gowrisankar22:
@evanchaoli opa integration will help to disable task to run as root. Globally can we disable this so that users adding the `privileged`flag in the task will not have any effect ??

evanchaoli:
> can we disable this so that users adding the `privileged`flag in the task will not have any effect ??

TBH, I don't quite get your question. Basically, with this PR, Concourse just sends raw info to OPA, and at the OPA side, you may write rules to allow or disallow actions based the raw data.

vito:
@evanchaoli Sorry for the slow turnaround on this, it ended up being assigned to folks who were pretty busy. We're still getting the hang of balancing PR review. We assign them every morning now and dedicate half of each day to PR reviews, so that should help, but sometimes we have to re-assign.

Since the RFC is still open can we add an experimental warning somewhere? This feature seems like it'd be adopted by the exact folks I wouldn't want to surprise with breaking changes. 🙂 

We can try to get this in for v6.2.0 but I don't want to make any promises since we're aiming for that to be an LTS release for customers, and increasing scope adds risk, even though this is opt-in. If it's not in v6.2 it'll be in v6.3 which should be 2-3 weeks later - we've been improving our release turnaround time lately.

As a side question, would you find it beneficial to have easy access to nightly releases from our CI?

evanchaoli:
@vito 
> Since the RFC is still open can we add an experimental warning somewhere? This feature seems like it'd be adopted by the exact folks I wouldn't want to surprise with breaking changes. 🙂

This should not be a breaking feature. Without configuring an OPA server, it won't impact anything. I'm ok to add an experimental warning, but unlike `set_pipeline` etc. steps, this feature has no UI part, only place come to my brain to add the warning is when `concourse web` process is started, if OPA is configured, then print a warning log. Is it ok to you?

> We can try to get this in for v6.2.0 but I don't want to make any promises since we're aiming for that to be an LTS release for customers, and increasing scope adds risk, even though this is opt-in. If it's not in v6.2 it'll be in v6.3 which should be 2-3 weeks later - we've been improving our release turnaround time lately.

> As a side question, would you find it beneficial to have easy access to nightly releases from our CI?

Yes! Actually we have a small "preview" cluster. We upgrade it to your latest nightly build every Monday, so that we provide our team as well as users a place to know and play with the latest changes.

vito:
@evanchaoli 

> This should not be a breaking feature. Without configuring an OPA server, it won't impact anything. I'm ok to add an experimental warning, but unlike `set_pipeline` etc. steps, this feature has no UI part, only place come to my brain to add the warning is when `concourse web` process is started, if OPA is configured, then print a warning log. Is it ok to you?

Yeah, that sounds fine. I figured this doesn't break anything since it's opt-in, I just want to warn people about future breakage just in case we go a different direction with the RFC. (Not that I expect that to happen, it's more of a general rule with any work for an RFC that's not yet complete.)

evanchaoli:
@vito I just push a new commit that added an experimental warning log with the RFC URL attached.

gowrisankar22:
@evanchaoli

Somehow we want to block using privileged containers for our users. Is that possible via opa?
More details: 
`fly -t ci execute -c sample.yml --privileged`
https://concourse-ci.org/resource-types.html#schema.resource_type.privileged
https://concourse-ci.org/jobs.html#schema.step.task-step.privileged

evanchaoli:
@gowrisankar22 I believe it's possible.

allan-degnan-rft:
Tracking this as an important feature.  

> Since the RFC is still open can we add an experimental warning somewhere? This feature seems like it'd be adopted by the exact folks I wouldn't want to surprise with breaking changes. 🙂

Just wanted to make a comment on @vito timeline/risk vis-à-vis 6.2/6.3; which is that you typically want to write your Policy as Code against an LTS, because API changes will obviously require updates to relevant policies.

For what that's worth in the list of pros and cons.

veotax:
@evanchaoli @vito @gowrisankar22 Hi all. Just a quick question. Why don't we use Casbin? OPA has 3.4k stars but [Casbin](https://casbin.org/) has more.. nearly 6.8k stars.

Using OPA means we have to start a OPA server daemon besides the concourse binary. If the OPA process is dead for some reason, then the whole authz system doesn't work any more, which is a big security hole. Using Casbin as a library can make sure authz is compiled into concourse binary and the authz functionality will never go offline. So is there any specific consideration for using OPA?

evanchaoli:
> @evanchaoli @vito @gowrisankar22 Hi all. Just a quick question. Why don't we use Casbin? OPA has 3.4k stars but [Casbin](https://casbin.org/) has more.. nearly 6.8k stars.
> 
> Using OPA means we have to start a OPA server daemon besides the concourse binary. If the OPA process is dead for some reason, then the whole authz system doesn't work any more, which is a big security hole. Using Casbin as a library can make sure authz is compiled into concourse binary and the authz functionality will never go offline. So is there any specific consideration for using OPA?

@veotax Thanks for the info about Casbin. I'm not familiar with that, but just took a quick look at it.

To clarify, with this PR, OPA doesn't have to be involved in authz  process. Cluster admin may define which API (action) should go through OPA. For example, a cluster only wants to enforce a policy where all pre-commit pipeline should include unit-test, then the admin may configure that only `SaveConfig` action should go through OPA.

In other words, the purpose of this PR is not to refactor Concourse's authz and RBAC by leveraging a policy engine (though maybe Concourse dev folks want so). Instead, this PR will provides a chance to Concourse users (cluster admin, team admin, pipeline author, etc) to enforce their own policies.

With OPA, cluster admin may define cluster policies, team/pipeline admin may also define team/pipeline level of policies. As the policies are a separate set of files outside of Concourse, policy can be updated at anytime, without rebuild/restart Concourse.

I agree that as OPA server is a separate process besides the concourse process, it introduces a failure point, but I think that is okey. My team has been using OPA for a while, it is reliable. Usually OPA process runs as a sidecar container, thus even if an OPA process crashes, it impacts a single ATC. Also, in practice, we don't configure all actions to go through OPA, thus OPA process crash should not lead the entire Concourse cluster to crash.

Further more, my implementation of this PR allows to extend other policy engines in the future. As I'm not familiar with Casbin, I cannot tell if Casbin may archive same thing this PR wants to do. But at least, just like Concourse supports multiple secrets managers, users may choose Vault, K8S, or else to use, up to their own situations. OPA is one of the policy engines to support, more could be supported, and users may choose what to use.

At this point, as I'm contributing the PR, and my team has decided to use OPA, so OPA is here.

veotax:
> To clarify, with this PR, OPA doesn't have to be involved in authz process. Cluster admin may define which API (action) should go through OPA. For example, a cluster only wants to enforce a policy where all pre-commit pipeline should include unit-test, then the admin may configure that only SaveConfig action should go through OPA.

Sorry but I didn't quite get the point here.. If only one action like SaveConfig goes through OPA, then why it's worthy to start a whole new process to handle it? Using a light-weight authz library like Casbin instead of a new process seems to be easier.

> In other words, the purpose of this PR is not to refactor Concourse's authz and RBAC by leveraging a policy engine (though maybe Concourse dev folks want so). Instead, this PR will provides a chance to Concourse users (cluster admin, team admin, pipeline author, etc) to enforce their own policies.

> With OPA, cluster admin may define cluster policies, team/pipeline admin may also define team/pipeline level of policies. As the policies are a separate set of files outside of Concourse, policy can be updated at anytime, without rebuild/restart Concourse.

I think other authz products can do the policy editing, policy update too without restarting any process. Just reloading rules will be OK. 

> I agree that as OPA server is a separate process besides the concourse process, it introduces a failure point, but I think that is okey. My team has been using OPA for a while, it is reliable. Usually OPA process runs as a sidecar container, thus even if an OPA process crashes, it impacts a single ATC. Also, in practice, we don't configure all actions to go through OPA, thus OPA process crash should not lead the entire Concourse cluster to crash.

I think your solution works for you because you are already using OPA, so you may share concourse's OPA process with your other applications. But for those who didn't use OPA at all, it's an extra cost.

> Further more, my implementation of this PR allows to extend other policy engines in the future. As I'm not familiar with Casbin, I cannot tell if Casbin may archive same thing this PR wants to do. But at least, just like Concourse supports multiple secrets managers, users may choose Vault, K8S, or else to use, up to their own situations. OPA is one of the policy engines to support, more could be supported, and users may choose what to use.

There's a lot of code changes here in this PR. So we need to be cautious here. I think switching to another authz is very difficult in future. Until now, I didn't see any strong points that we have to go along this way. @vito what's your point?

evanchaoli:
> I think your solution works for you because you are already using OPA, so you may share concourse's OPA process with your other applications. But for those who didn't use OPA at all, it's an extra cost.
> 

If you don't configure an OPA to Concourse, it feature will not do anything, why is it an extra cost? It just provide a chance for users to do policy enforcement, users may choose to use or not to use it. That is it!

> There's a lot of code changes here in this PR. So we need to be cautious here. I think switching to another authz is very difficult in future. Until now, I didn't see any strong points that we have to go along this way. @vito what's your point?

To clarify again, this PR doesn't touch Concourse's authz and RBAC at all. If Concourse wants to switch to leverage Casbin for authz and RBAC, go ahead, there is no conflict.

allan-degnan-rft:
> > To clarify, with this PR, OPA doesn't have to be involved in authz process. Cluster admin may define which API (action) should go through OPA. For example, a cluster only wants to enforce a policy where all pre-commit pipeline should include unit-test, then the admin may configure that only SaveConfig action should go through OPA.
> 
> Sorry but I didn't quite get the point here.. If only one action like SaveConfig goes through OPA, then why it's worthy to start a whole new process to handle it? Using a light-weight authz library like Casbin instead of a new process seems to be easier.

It enables generic policy as code, i.e:

- I would like to ensure I'm using trusted images / resources.  
- I would like to enforce my git commits are signed. 
- I would like to enforce a pipeline run through specific tasks (testing phases, the above git commit checker, secret detector, etc)
- I would like to enforce my terraform changes (via the resource) does a plan and then further policy as code checks ( which again, you'd use OPA for )

Further to the point, you can embed OPA into Golang, so the argument to create another process or not is actually separate debate to whether or not Casbin would be a better. 

Stepping aside that 100% of your public activity seems to be trolling around projects suggesting they migrate to casbin.  

- Does it offer generic policy as code capabilities?  
- How would you actually achieve any of the above?
- Do you have someone willing to offer the code?
- Would it not make sense, as @evanchaoli suggested, to just make this a generic offering at that point?

Apologies for weighing in as an outside observer, but I've actually been tracking this since Jan as the above example use-cases are very important to me and OPA usage fits in with a generic CNCF stack.

vito:
To be honest, I'm a bit of a fly on the wall here and have no experience with Casbin or OPA, but from this conversation it seems like there are multiple competing solutions in this problem space, and we shouldn't necessarily expect Concourse to take an opinionated stance here that any one thing is The One True Solution for policy enforcement - whether that's OPA or Casbin.

We tried that with metrics - we wanted everyone to use Riemann because it was generic enough to emit to everything else. It didn't work out. Carrying these battles over from one project to another isn't worth it IMO.

That being said, having only one implementation isn't the same as taking an opinionated stance. It just means that's all that's been contributed so far.

The OPA RFC has a [stated goal](https://github.com/concourse/rfcs/pull/41/files#diff-920d7a06e64dac5ea2673094189fa570R31-R33) of supporting policy managers outside of just OPA, which I think is the Concourse-y thing to do. It makes total sense to me that @evanchaoli focused on OPA so far - because that's what they're familiar with. I don't think it's fair to task them with exploring all the other solutions.

Instead, those with experience with other policy managers should try to help out with the RFC. This seems like an opportunity for finding common ground and defining what that interface might look like - not for turning someone else's contribution into a battleground.

atc/policy/checker.go
```diff
@@ -0,0 +1,160 @@
...
+	return "policy check not pass"
+}
+
+type Filter struct {
+	HttpMethods   []string `long:"policy-check-filter-http-method" description:"API http method to go through policy check"`
+	Actions       []string `long:"policy-check-filter-action" description:"Actions in the list will go through policy check"`
+	ActionsToSkip []string `long:"policy-check-filter-action-skip" default:"UsingImage" description:"Actions the list will not go through policy check"`
+}
+
+func (f Filter) normalize() Filter {
```
vito, line 25: I think @jamieklassen was right about this in https://github.com/concourse/concourse/pull/5034#discussion_r373024220 - `twentythousandtonnesofcrudeoil` already configures `,` as the env delimiter so this shouldn't be needed:

https://github.com/vito/twentythousandtonnesofcrudeoil/blob/3b21ad808fcb867be679c4cad408a769a3977a1b/environment.go#L21-L23

atc/policy/checker.go
```diff
@@ -0,0 +1,160 @@
...
+	"code.cloudfoundry.org/lager"
+	"github.com/jessevdk/go-flags"
+)
+
+const ActionUsingImage = "UsingImage"
+
+type PolicyCheckNotPass struct{}
+
+func (e PolicyCheckNotPass) Error() string {
+	return "policy check not pass"
```
vito, line 16: This sounds a little better:

```suggestion
	return "policy check rejected"
```

(Plus renaming the type to match if that message is OK.)

atc/exec/policy_check.go
```diff
@@ -0,0 +1,57 @@
+package exec
+
+import (
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/atc/policy"
+)
+
+//go:generate counterfeiter . PolicyChecker
+
+type PolicyChecker interface {
```
vito, line 10: nit: this should probably be called `ImagePolicyChecker` so it doesn't get confused with the API policy checker

atc/exec/task_step.go
```diff
@@ -206,6 +210,20 @@ func (step *TaskStep) run(ctx context.Context, state RunState) error {
 
 	step.delegate.Initializing(logger)
 
+	if step.policyChecker != nil && config.ImageResource != nil {
```
vito, line 213: This could be circumvented by using [`image:`](https://concourse-ci.org/jobs.html#schema.step.task-step.image) and pointing to an image fetched by a `get` step or provided by a task output. Or by using a resource type that doesn't use `repository:` in its configuration.

atc/exec/policy_check.go
```diff
@@ -0,0 +1,57 @@
...
+	// Actions in skip list will not go through policy check.
+	if c.policyChecker.ShouldSkipAction(policy.ActionUsingImage) {
+		return true, nil
+	}
+
+	imageInfo := map[string]string{
+		"step": step,
+	}
+
+	if repository, ok := imageSource["repository"].(string); ok {
```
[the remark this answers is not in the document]
vito: Maybe we need a way for resource types/prototypes to provide Concourse with some sort of image identifier that can be passed to the policy checker.

In the future maybe this could be returned by the Prototype [`info` call](https://github.com/vito/rfcs/blob/prototypes/037-prototypes/proposal.md#prototype-info)? 🤔 I had a similar thought for a better design for webhooks, where `info` returns some sort of "hook tag" like `github:repo:foo` which can be mapped to incoming web hook payloads in order to know which resources to check.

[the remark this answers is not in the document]
vito: (P.S. I won't block the PR on this if there's really no better way, but I do think we should have a plan for the future and maybe even leave a TODO comment.)

atc/policy/checker.go
evanchaoli: Fixed.

atc/exec/policy_check.go
evanchaoli: Fixed.

atc/policy/checker.go
evanchaoli: Oh, yes. I just verified. Good to know.

atc/exec/policy_check.go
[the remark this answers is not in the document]
evanchaoli: Yeah, I think this is a valid comment, we should treat source as a black box. So let me change data sent to policy server to `image_source_type` + `image_source`, where `image_source_type` will be `registry-image`, `docker-image` etc., and `image_source` will be source of the type.

atc/exec/check_step.go
```diff
@@ -82,6 +86,22 @@ func (step *CheckStep) Run(ctx context.Context, state RunState) error {
 		return fmt.Errorf("timeout parse: %w", err)
 	}
 
+	if step.policyChecker != nil {
```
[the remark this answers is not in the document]
evanchaoli: This makes sense. After read through the container launch process, I now choose to hook the check in function `FindOrCreateContainer` of `atc/worker/worker.go`. This way, there is no need to add the check to every individual step. Please see if you feel ok with that. (P.S. I haven't added tests for the change in worker.go yet, because I want to get your confirm with the direction first.)

evanchaoli:
> Checking images feels like an invasive change that can be easily worked around, so that's my main concern with this. I'm pretty uncomfortable with breaking the convention of resource configuration being a black box. The rest looks great, and maybe there's really no better way to do this right now. In any case I wonder if it's worth it since it can be so easily circumvented.

@vito I agree that checking image can be circumvented, we thought over this problem before I added the check. But the following thoughts supported me to add that:

* Concourse is a CI/CD platform rather than a container runner. Containers are all launched via pipeline steps, which reduce the possibility of hack.
* As a CI/CD platform, Concourse faces to only CI/CD writers rather than all developers, which also reduces hacks.
* As a CI/CD platform, Concourse usually runs as an internal system behind firewall, in other words, Concourse usually faces to internal employees rather than hackers from Internet. 
* Main goal of policy check is prevent from unintentional policy breakages rather than hacker behaviors. For example, if the company set a security policy where `busybox` image is no longer allowed in any CI pipeline. Without an auto policy enforcement, users may 1) not be able to aware of the policy; 2) just ignore the policy; but with auto policy enforcement, pipelines with `busybox` would stop working, then users would most probably change `busybox` to a legal alternative rather than hack a way to continue to use `busybox`.
* Even if a Concourse runs on Internet, and the admin considers the check is fragile, as every action to check is configurable, the admin may disable the check.
* At `SaveConfig` API, images inuse can be checked, but that's not enough. Because a mature pipeline may silently run there for months and years without being re-set-pipeline, but policy may be created/updated anytime, thus a check has to be done before launching containers.

Our company has a security policy where docker images `busybox` and `alpine` are not allowed to use in anywhere, thus checking image would be a must for us to enforce the policy.

In summary, I agree image checking has the problem, but it's still useful in some use cases. What we should do is to document the problem, and let users decide if or not to use the check.

atc/exec/task_step.go
evanchaoli: Removed the check from task step.

atc/worker/worker.go
```diff
@@ -207,6 +211,72 @@ func (worker *gardenWorker) LookupVolume(logger lager.Logger, handle string) (Vo
 	return worker.volumeClient.LookupVolume(logger, handle)
 }
 
+var sensitiveKeys = []string{"password", "key", "token"}
```
vito, line 214: Ah, so this must have been one of the motivations for only checking certain fields. A blocklist approach is indeed kinda risky.

I wonder: if this check is performed before credentials are resolved (which I *think* may be the case - Concourse tries to evaluate creds as late as possible), do you think we could just pass the whole payload along? This won't help for people who aren't using credential management, but, well... they probably should be. There may be some other case I'm missing though.

More generally: is the network traffic to OPA expected to be secure and trusted? If all else fails, how worried are we if credentials are sent along?

Scope creep, not needed for this PR: I wonder if we should introduce another field alongside `source:` for storing sensitive fields so that in situations like this we can simply not send them along. Being able to distinguish which config "identifies" the resource and which config is necessary to access it would likely be helpful in other scenarios within Concourse.

atc/worker/worker.go
```diff
@@ -207,6 +211,72 @@ func (worker *gardenWorker) LookupVolume(logger lager.Logger, handle string) (Vo
...
+		}
+	} else {
+		// Ignore other images as policy checker cannot do much on them.
+		return true, nil
+	}
+
+	input := policy.PolicyCheckInput{
+		Action:   policy.ActionUsingImage,
+		Team:     containerSpec.TeamName,
+		Pipeline: metadata.PipelineName,
```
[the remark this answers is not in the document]
evanchaoli: Moved team to context. As container metadata already contains pipeline name, I don't store pipeline name to context.

atc/worker/worker.go
evanchaoli: > I wonder: if this check is performed before credentials are resolved (which I _think_ may be the case - Concourse tries to evaluate creds as late as possible), do you think we could just pass the whole payload along? This won't help for people who aren't using credential management, but, well... they probably should be. There may be some other case I'm missing though.
> 

I believe this check is done after var interpolation. Because var evaluation happens in beginning of a step, and then a step tries to launch a container.

> More generally: is the network traffic to OPA expected to be secure and trusted? If all else fails, how worried are we if credentials are sent along?
> 

Traffic between concourse and opa could be either secure or insecure, depends on how it's setup. But, policy checking might go to concourse debug log, so secrets might be printed to concourse logs, and concourse logs are not supposed to be secrets. In our clusters, concourse logs are redirected to log-insight, and everyone has access to log-insight. Similarly at OPA side, an OPA admin may turn on debug log, then all policy check requests are printed to OPA logs. That's the main concern I have.

> Scope creep, not needed for this PR: I wonder if we should introduce another field alongside `source:` for storing sensitive fields so that in situations like this we can simply not send them along. Being able to distinguish which config "identifies" the resource and which config is necessary to access it would likely be helpful in other scenarios within Concourse.

That's a good idea, but that would be a breaking change. Even if such a field is introduced, I guess we have to be back-compatible `source` for existing pipelines for a while, thus the new field doesn't help in short term. But for long term, I'm totally on this idea.

I just pushed a new commit, that addressed the other two comments but this one. Let me think more on this today.

evanchaoli: In the latest commit, I leverage secret redaction via build delegate to redact secrets in image source. Please see what you feel. I don't want to redact the entire message sent to OPA, because I have seen many users store non-secret in Vault, which causes unintentional redactions, and might lead to policy checking rejected.

evanchaoli:
Just rebased and resolved the conflicts.

gowrisankar22:
@evanchaoli Is that possible to mention examples to disable privileged containers in docs?

@vito Which version this feature gonna merged because I can see `6.3.0`is released yesterday.

vito:
@gowrisankar22 v6.3 ended up being our LTS instead of v6.2, so now we're aiming for v6.4.

atc/engine/builder/delegate_factory.go
```diff
@@ -389,6 +390,20 @@ func (delegate *buildStepDelegate) buildOutputFilter(str string) string {
 	return it.line
 }
 
+func (delegate *buildStepDelegate) ImageSourceRedaction(source atc.Source) (atc.Source, error) {
```
vito, line 393: This is a clever approach to the problem! There are a couple caveats though: it relies on the credentials coming from a credential manager (or `load_var`), and it relies on having secret redaction enabled.

I think these caveats are OK, but they are probably worth reflecting in the RFC.

We should also look into whether we can un-feature-flag secret redaction. I've opened an issue for that: https://github.com/concourse/concourse/issues/5771

All that aside, I think this might be a clearer name:

```suggestion
func (delegate *buildStepDelegate) RedactImageResource(source atc.Source) (atc.Source, error) {
```

atc/worker/image_factory.go
```diff
@@ -59,3 +61,6 @@ type NoopImageFetchingDelegate struct{}
 func (NoopImageFetchingDelegate) Stdout() io.Writer                                 { return ioutil.Discard }
 func (NoopImageFetchingDelegate) Stderr() io.Writer                                 { return ioutil.Discard }
 func (NoopImageFetchingDelegate) ImageVersionDetermined(db.UsedResourceCache) error { return nil }
+func (NoopImageFetchingDelegate) ImageSourceRedaction(source atc.Source) (atc.Source, error) {
+	return source, nil
```
vito, line 65: I'm not sure if it's OK for this to be a no-op - it's used during the `check` for an `image_resource:` image, so we may need credentials to be redacted in that case too.

atc/worker/worker.go
```diff
@@ -207,6 +212,64 @@ func (worker *gardenWorker) LookupVolume(logger lager.Logger, handle string) (Vo
 	return worker.volumeClient.LookupVolume(logger, handle)
 }
 
+func secureImageSource(delegate ImageFetchingDelegate, source atc.Source) atc.Source {
+	newSource, _ := delegate.ImageSourceRedaction(source)
```
vito, line 216: This error should probably not be dropped, even though it looks pretty unlikely to happen right now.

atc/engine/builder/delegate_factory.go
vito: FWIW, I'm willing to concede that what you had before was simpler and safer, even though it opened the "black box".

If this path starts to feel too treacherous I'd be OK with going back and opening an issue for something like https://github.com/concourse/concourse/pull/5034#discussion_r436028270 to clean it up.

atc/worker/client.go
```diff
@@ -46,7 +46,8 @@ type Client interface {
 		workerSpec WorkerSpec,
 		strategy ContainerPlacementStrategy,
 		containerMetadata db.ContainerMetadata,
-		resourceTypes atc.VersionedResourceTypes,
+		imageFetcherSpec ImageFetcherSpec,
```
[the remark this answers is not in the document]
evanchaoli: Thanks.

atc/worker/worker.go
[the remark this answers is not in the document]
evanchaoli: Okey, added pipeline to context.

atc/worker/worker.go
evanchaoli: Fixed.

atc/worker/image_factory.go
evanchaoli: yeah, good catch! As it's noop, we may just return an empty source as redacted.

atc/engine/builder/delegate_factory.go
evanchaoli: Renamed.

atc/worker/worker.go
```diff
@@ -207,6 +212,68 @@ func (worker *gardenWorker) LookupVolume(logger lager.Logger, handle string) (Vo
...
+	}
+
+	imageSpec := containerSpec.ImageSpec
+
+	imageInfo := map[string]interface{}{
+		"privileged": imageSpec.Privileged,
+	}
+
+	if imageSpec.ImageResource != nil {
+		imageInfo["image_source_type"] = imageSpec.ImageResource.Type
```
vito, line 238: nit: I find it confusing to say 'source type' when they're sibling fields.

```suggestion
		imageInfo["image_type"] = imageSpec.ImageResource.Type
```

atc/policy/checker.go
```diff
@@ -0,0 +1,144 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"code.cloudfoundry.org/lager"
+	"github.com/jessevdk/go-flags"
+)
+
+const ActionUsingImage = "UsingImage"
```
vito, line 11: Probably best to match the imperative tone of the API handlers:

```suggestion
const ActionUseImage = "UseImage"
```

atc/policy/checker.go
```diff
@@ -0,0 +1,144 @@
...
+type PolicyCheckNotPass struct{}
+
+func (e PolicyCheckNotPass) Error() string {
+	return "policy check rejected"
+}
+
+type Filter struct {
+	HttpMethods   []string `long:"policy-check-filter-http-method" description:"API http method to go through policy check"`
+	Actions       []string `long:"policy-check-filter-action" description:"Actions in the list will go through policy check"`
+	ActionsToSkip []string `long:"policy-check-filter-action-skip" default:"UsingImage" description:"Actions the list will not go through policy check"`
```
vito, line 22: I just noticed this default value here and found it kind of surprising. Can this be made opt-in instead, so there are no defaults here at all?

The concern I have is that I may not care about `UseImage` at all, and then list some HTTP actions to skip, not realizing that I've removed this default value along the way. Doing so may then cause all image usage checks to fail.

atc/worker/worker.go
evanchaoli: Renamed.

atc/policy/checker.go
evanchaoli: Renamed.

atc/policy/checker.go
evanchaoli: This is a good point. I have removed `UseImage` from default of `ActionsToSkip`, and I changed the way, only if `UseImage` is defined in `Actions` list, it will go through policy check. So that, users must explicitly define `UseImage` in `Actions` if they want to check `UseImage`.

evanchaoli: Now the behavior is, if a user only configures a OPA server, but without configuring `HttpMethods` and `Actions`, then no action will go through policy check. Meaning that, by default, no action will go through policy check. Users must explicitly configure which Http methods or actions should go through policy check.

marco-m-pix4d:
Hello @evanchaoli, is there any documentation available on how to use this feature, besides https://github.com/evanchaoli/rfcs/blob/opa/041-opa-integration/proposal.md ?

evanchaoli:
> Hello @evanchaoli, is there any documentation available on how to use this feature, besides https://github.com/evanchaoli/rfcs/blob/opa/041-opa-integration/proposal.md ?

So far, a sample usage is at https://github.com/concourse/concourse/pull/5034#issuecomment-630053739. I think Concourse team is working on the documentation.

marco-m-pix4d:
@evanchaoli  Ah, I missed that one. Thanks!

PR #4607, opened 2019-10-16
experiment: add tracing to ATC

cirocosta:
Hey,

In https://github.com/concourse/concourse/pull/4247, I brought up that perhaps
we should not think of our metrics as events, but rather as indicators of
system status, but I didn't really went with a proposal for what could be the
thing that would let operators have something that resambles events (with very
precise timing, high cardinality, etc).

some time ago though, @njbennett had brought up something quite interesting
(https://github.com/concourse/concourse/issues/3958): Concourse could emit build
events as traces!

and, that's essentially what I tried implementing here, trying to be as neutral
as possible with regards to vendors by going straight with opentelemetry.

here's the result of this PR:

Screen Shot 2019-10-15 at 9 17 21 PM

*(yesterday, I demoed the PR to the team - here's where you can find what we talked about: https://ops.tips/notes/tracing-builds-concourse/)*

Although the PR does work (yeah, please, try it out!), my goal with this one is
to gather some thoughts from all of us to see whether that's something that we'd
like to do, think about possible problems, and ... just reason about it, thus,
marking it as "draft".

The PR is broken down in a series of commits that kinda tell a story:

- first, I implement a little wrapper on top of opentelemetry to shield
  ourselves from the changes that are constantly happening underneath
- then, subsequently, I wire that to the `web` command, so that one can
  configure it
- then, I keep adding tracepoints from the very high level build tracker down to
  container creation.

Please let me know what you think! I'm quite excited about the possibilities
here, and, if we do this right, I can see this being a big addition to
Concourse.

Thank y'all!

---

here's a sample `docker-compose.yml` that spins up 3 workers and has Jaeger set up: https://gist.github.com/cirocosta/6a676a071d64e0076a30139e76f46a74

PR #5043, opened 2020-01-20
add tracing to ATC

cirocosta:
### Existing Issue

This PR is a slight refactor rebased on `master` of the functonality first
introduced in [experiment: add tracing to ATC #4607](https://github.com/concourse/concourse/pull/4607).

The PR adds a new package (`github.com/concourse/concourse/tracing`) that
provides a thin wrapper on top of [opentelemetry]l - the now de-facto standard
implementation for distr tracing - for us to consume throughout Concourse
without the need of being exposed to the internals of opentelemetry.

For instance, here's how we can tell how long a `get` step took, as well as
recording if there was an error or not:

https://github.com/concourse/concourse/compare/build-tracing-2?expand=1#diff-514296ff114a3c668658b1f40922d758R93-R108

Which, can be visualized like this in [jaeger]'s UI

Screenshot from 2020-01-19 20-55-36

To make possible to test (and use) the functionality locally, I included a
[docker-compose override] that lets one easily run [jaeger] locally with `web`
nodes sending their traces to it.

	docker-compose \
		-f ./docker-compose.yml \
		-f ./hack/overrides/jaeger.yml \
		up -d

[opentelemetry]: https://opentelemetry.io/
[jaeger]: https://www.jaegertracing.io/
[docker-compose override]: https://docs.docker.com/compose/extends/

ps.: differently from the initial PR, this one DOES NOT add tracepoints to every
single step and `RunScript`. I took the approach of limiting the tracepoints to
`get`, `task`, and `put` to minimize the amount of tracepoints so that we can
better reason about what we trace later in future PRs.

### Contributor Checklist

- [x] Unit tests
  - should I?

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


atc/exec/put_step.go
```diff
@@ -73,15 +73,31 @@ func NewPutStep(
...
+		"name":     step.plan.Name,
+	})
+
+	err := step.run(ctx, state)
+	tracing.End(span, err)
+
+	return err
+}
+
+// run chooses a worker that supports the step's resource type and creates a
```
vito: Might want to keep this godoc on the public method so it's exposed.

atc/exec/task_step.go
```diff
@@ -106,7 +107,22 @@ func NewTaskStep(
...
+		"name":     step.plan.Name,
+	})
+
+	err := step.run(ctx, state)
+	tracing.End(span, err)
+
+	return err
+}
+
+// run will first select the worker based on the TaskConfig's platform and the
```
vito: (here too)

go.mod
```diff
@@ -59,14 +59,12 @@ require (
...
 	github.com/jessevdk/go-flags v1.4.0
-	github.com/jonboulle/clockwork v0.1.0 // indirect
 	github.com/json-iterator/go v1.1.7 // indirect
-	github.com/konsorten/go-windows-terminal-sequences v1.0.2 // indirect
 	github.com/kr/pty v1.1.8
 	github.com/krishicks/yaml-patch v0.0.10
-	github.com/lib/pq v0.0.0-20181016162627-9eb73efc1fcc
-	github.com/mattn/go-colorable v0.1.1
-	github.com/mattn/go-isatty v0.0.7
+	github.com/lib/pq v1.2.0
```
vito: I guess we could bump straight to v1.3.0 if we're bumping this, no big deal though.

release-notes/latest.md
```diff
@@ -1,3 +1,19 @@
 #### <sub><sup><a name="4938" href="#4938">:link:</a></sup></sub> feature
 
 * Include job label in build duration metrics exported to Prometheus. #4976
+
+#### <sub><sup><a name="4607" href="#4607">:link:</a></sup></sub> feature
+
+* Add experimental support for exposing traces to [Jaeger] or [Stackdriver].
+
+With this feature enaled (via `--tracing-(jaeger|stackdriver)-*` variables in
```
vito:
```suggestion
With this feature enabled (via `--tracing-(jaeger|stackdriver)-*` flags in
```

atc/engine/engine.go
```diff
@@ -180,6 +180,14 @@ func (b *engineBuild) Run(logger lager.Logger) {
 
 	defer notifier.Close()
 
+	ctx, span := tracing.StartSpan(b.ctx, "build", tracing.Attrs{
+		"team":     b.build.TeamName(),
+		"pipeline": b.build.PipelineName(),
+		"job":      b.build.JobName(),
+		"build":    b.build.Name(),
```
vito: Should we add `build_id` here too? Also should we omit pipeline/job/build if it's a one-off build?

hack/overrides/jaeger.yml
```diff
@@ -0,0 +1,30 @@
+# jaeger.yml - a docker-compose override that adds 'jaeger to the stack.
+#
+# once running, head to `localhost:25000` to get access to the Jaeger UI.
```
taylorsilva: wondering why `25000` instead of jaegar's default port?

CONTRIBUTING.md
```diff
@@ -211,6 +211,25 @@ After this is done, the final step is to connect your IDE to the debugger with t
 For GoLand you can do so by going to Run | Edit Configurations… | + | Go Remote and fill in the parameters.
 
 
+### Trying out distributed tracing with Jaeger
+
+Under `./hack`, a docker-compose override file named `jaeger.yml` provides the
+essentials to get [Jaeger] running alongside the other components, as well tying
```
taylorsilva:
```suggestion
essentials to get [Jaeger] running alongside the other components, as well as tying
```
nit: a word

atc/engine/engine.go
cirocosta: 👍  for `build_id` 😁 

---

> Also should we omit pipeline/job/build if it's a one-off build?

I tried coming up with a reason for keeping / omitting, and only had a real reason for keeping - to filter down the list of builds who do not have a pipeline assigned to them (thus, one-offs). 

if omitting, then I think it might be good to add a field that'd mark the build as one-off (as the check for `pipeline=""` wouldn't be a thing anymore).

(also, without the condition, it looks nicer lol)

```diff
diff --git a/atc/engine/engine.go b/atc/engine/engine.go
index b74ba9746..59c57a25b 100644
--- a/atc/engine/engine.go
+++ b/atc/engine/engine.go
@@ -3,6 +3,7 @@ package engine
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"sync"
 	"time"
 
@@ -180,12 +181,18 @@ func (b *engineBuild) Run(logger lager.Logger) {
 
 	defer notifier.Close()
 
-	ctx, span := tracing.StartSpan(b.ctx, "build", tracing.Attrs{
+	attrs := tracing.Attrs{
 		"team":     b.build.TeamName(),
-		"pipeline": b.build.PipelineName(),
-		"job":      b.build.JobName(),
 		"build":    b.build.Name(),
-	})
+		"build_id": strconv.Itoa(b.build.ID()),
+	}
+
+	if b.build.PipelineID() != 0 {
+		attrs["pipeline"] = b.build.PipelineName()
+		attrs["job"] = b.build.JobName()
+	}
+
+	ctx, span := tracing.StartSpan(b.ctx, "build", attrs)
 	defer span.End()
 
 	step, err := b.builder.BuildStep(logger, b.build)
 ```

wdyt?

I really don't hold a strong opinion for either

File: hack/overrides/jaeger.yml (hunk 6 above)
Comment (cirocosta) [reply hack/overrides/jaeger.yml]: 25000 was a number that I could remember lol no big reason hahah `16686` in the middle of all of those other ports seem kinda hard:

```
0.0.0.0:25000->16686/tc
0.0.0.0:14268->14268/tcp	
0.0.0.0:5775->5775/udp
0.0.0.0:5778->5778/tcp
0.0.0.0:6831-6832->6831-6832/udp
0.0.0.0:9411->9411/tcp
14250/tcp
```

File: atc/engine/engine.go (hunk 5 above)
Comment (vito) [reply atc/engine/engine.go]: Fine with keeping them instead of adding a field just to explicitly say it's a one-off. 👍

**Comment by cirocosta** [comment]
Hey,

I just fixed the conflict w/ regards to release notes - just let me know if anything else still needs to be worked on @vito  @taylorsilva  😁  

thanks!

File: atc/engine/engine.go [hunk 8]
```diff
@@ -214,7 +224,7 @@ func (b *engineBuild) Run(logger lager.Logger) {
 
 	done := make(chan error)
 	go func() {
-		ctx := lagerctx.NewContext(b.ctx, logger)
+		ctx = lagerctx.NewContext(ctx, logger)
```
vito: Reassigning `ctx` in a goroutine here seems like it could be a race condition waiting to happen. :thinking: It's fine right now because nothing uses `ctx` after the goroutine but this seems like it might be hard to catch in the future. I'm not sure if we build and run Concourse with `--race` anywhere in the pipeline.

cirocosta: oh, we definitely don't need the definition there

>  I'm not sure if we build and run Concourse with --race anywhere in the pipeline.

I remember chatting about this with someone (tried searching the place, but found nothing lol) that we could perhaps have `smoke` running with `-race` (although, I think we'd need to build a version of `concourse` with `-race` as that's a build-time modification)

PR #4178, opened 2019-07-28
atc: Warn on unused resources vs erroring out, solves #4177

kallisti5:
fixes: #4177

Issue #4944, opened 2019-12-20
Only perform checks for resources that are used as inputs


vito:
Right now we perform `check`s across all resources. This is redundant for resources that are only ever used as `put` steps, since these versions won't be used for scheduling at all.

For better or for worse, lots of users have `put`-only resources for things like notifications and collecting build metadata, so this could be a nice short-term win now that we have this information readily available in the `job_inputs` table.

Some context in #4940

evanchaoli:
Per https://github.com/concourse/concourse/issues/4940#issuecomment-568083030, this change depends on table `job_inputs` and `job_inputs` will only be available from 6.0.

tkellen:
Yes please!

vito:
This was implemented by https://github.com/concourse/concourse/pull/5149 :slightly_smiling_face:

Issue #4177, opened 2019-07-28
Warn instead of fail on unused resources


kallisti5:
## What challenge are you facing?

It would be *great* to be able to leverage unused resources within pipelines.  We deploy multiple pipelines using the same yaml. These pipelines are "one per branch" (master, release1, release2, etc)

We template the branch out within our yaml.  It would be awesome to have our pipelines "select" the resource they want based on the branch name... however unused resources result in pipelines erroring out:

```
error: invalid pipeline config:
invalid resources:
	resource 'master-time-trigger' is not used
	resource 'r1beta2-time-trigger' is not used
```
Example:
```yaml

resources:
  - name: master-time-trigger
    type: time
    source:
      start: 1:00 AM
      stop: 6:00 AM
      location: America/Chicago
  - name: r1beta1-time-trigger
    type: time
    source:
      days: [Sunday]
      start: 1:00 AM
      stop: 6:00 AM
      location: America/Chicago
  - name: r1beta2-time-trigger
    type: time
    source:
      days: [Monday,Friday]
      start: 1:00 AM
      stop: 6:00 AM
      location: America/Chicago
.
.
# Jobs
jobs:
  - name: image-((branch))-((arch))
    public: true
    plan:
      - get: ((branch))-time-trigger
        trigger: true
      - get: haiku-git
        trigger: false
      - task: build ((branch)) anyboot
        config:
          platform: linux
          image_resource:
            type: registry-image
            source: { repository: haiku/toolchain-worker-((branch)) }
          inputs:
            - name: haiku-git
          outputs:
            - name: generated.((arch))
          run:
            path: /bin/bash
            args:
```

## What would make this better?

Concourse should warn the user about unused triggers instead of erroring out about them. Unused resources could be valuable in more complex pipelines.  Assuming the user knows what their doing and "just letting them know" vs blocking their attempts would be ideal.

## Are you interested in implementing this yourself?

Maybe depending on complexity.

kallisti5:
See PR https://github.com/concourse/concourse/pull/4178

kcmannem:
This isn't really my area in the code so I don't know what side effects unused resources will cause. Another approach we can try is to remove unused resources invisibly from the interpreted yaml before calling out to the ATC. This way we can spit out a warning such as: `Unused resources referenced. Excluding them from pipeline config.`

clarafu:
[...]

PR #5060, opened 2020-01-22
Warn on unused resources vs erroring out

xtremerui:
resubmitting of #4178

### DIFF
--- a/atc/configvalidate/validate.go
+++ b/atc/configvalidate/validate.go
@@ -36,10 +36,11 @@ func Validate(c Config) ([]ConfigWarning, []string) {
 		errorMessages = append(errorMessages, formatErr("resources", resourcesErr))
 	}
 
-	resourceTypesErr := validateResourceTypes(c)
+	resourceTypesWarnings, resourceTypesErr := validateResourceTypes(c)
 	if resourceTypesErr != nil {
 		errorMessages = append(errorMessages, formatErr("resource types", resourceTypesErr))
 	}
+	warnings = append(warnings, resourceTypesWarnings...)
 
 	varSourcesErr := validateVarSources(c)
 	if varSourcesErr != nil {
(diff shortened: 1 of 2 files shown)


cirocosta:
Hey @xtremerui,

@aoldershaw and I were taking a look at the changes, here are some things we
noticed:

1. without any changes to `fly`, the warning shows as a deprecation notice:


```console
$ fly -t local set-pipeline -p test -c ./pipeline.yml
resources:
  resource repository has been added:
+ name: repository
+ source:
+   uri: http://github.com/concourse/concourse
+ type: git

jobs:
  job test has been added:
+ name: test
+ plan:
+ - config:
+     container_limits: {}
+     image_resource:
+       source:
+         repository: busybox
+       type: registry-image
+     platform: linux
+     run:
+       path: "true"
+   task: test

apply configuration? [yN]: y

DEPRECATION WARNING:
  - resource 'repository' is not used in pipeline

pipeline created!
you can view your pipeline here: http://localhost:8080/teams/main/pipelines/test
```

It seems like some work could be done over here:
https://github.com/concourse/concourse/blob/8153ad08f7963b91fdf1f658820029341ea05edb/fly/commands/internal/displayhelpers/error_handling.go#L12-L43


2. having the check being implicitly performed inside `validateResourceTypes`
seems odd - perhaps it'd be better if not placed within `validateResourceTypes`
as it's not only about resource types?

https://github.com/concourse/concourse/pull/5060/files#diff-a86e8b6a0cbf3c90b8f1f94e47230d31R181


Wdyt?

thanks!

cirocosta:
With regards to @clarafu 's point here https://github.com/concourse/concourse/issues/4177#issuecomment-516100938 - should we perhaps make very clear to the user that they are about to have unnecessary checks going on?

xtremerui:
@cirocosta hmm thinking about it again I agee that unused resource should not check implicitly.

clarafu:
I think this PR might be feasible after we implement https://github.com/concourse/concourse/issues/4944! Which is to only check resources that are used as an input to a job.

xtremerui:
I will leave the original issue open and make it `blocked` until #4944 is done. Thx @clarafu !

rajatrj16:
Do we have a solution for the unused vars? 
```
WARNING:
Error loading existing config:
  - invalid resources:
        resource 'docker-image-dummy' is not used
        resource 'docker-test-image-dummy' is not used

error: configuration invalid
```
I am using this in the put resource

PR #5067, opened 2020-01-24
Revert "tf: temporary pend failing testflight test"

cirocosta:
### Why do we need this PR?

Previously, we had marked the test where we create a copy-on-write of a copy-on-write pending so that we could have our pipelines unblocked (see more here: https://github.com/concourse/concourse/issues/4964)

Now that we've got a fix (see https://github.com/concourse/baggageclaim/pull/35) , there's no
need to have this test skipped anymore.

ps.: I'm currently marking this as a draft because it's not complete - it must come with a commit that bumps the version of `baggageclaim` (as the PR didn't get in yet, we don't have a tag to reference in `go.mod`).

Reverts the commit where we skipped a failing test.

--

depends on https://github.com/concourse/concourse/pull/5160

### DIFF
--- a/testflight/volume_mounting_test.go
+++ b/testflight/volume_mounting_test.go
@@ -36,8 +36,7 @@ var _ = Describe("A job with nested volume mounts", func() {
 		Expect(sess).To(gbytes.Say("hello"))
 	})
 
-	// Pending this test for now, @cirocosta will be looking into it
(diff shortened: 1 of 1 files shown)


xtremerui:
Nice! Can't wait to see this got fixed. What is the targeted concourse version here? 5.8.1? Are we gonna back-port it to older version?

PR #5078, opened 2020-01-28
atc: log pipeline-id when failing to create check

cirocosta:
### Why do we need this PR?

Currently, when checkFactory fails to create a check for a pipeline, we
can't tell which pipeline had the problem.

Screen Shot 2020-01-28 at 10 00 21 AM

(^ can't tell who's causing that)

Include `lager.Data` containing the `pipeline-id` of that pipeline,
one is able to at least tell which pipline is the offending one.

### Contributor Checklist

- ~[ ] Unit tests~
- ~[ ] Integration tests (if applicable)~
- ~[ ] Updated documentation (located at https://github.com/concourse/docs)~
- ~[ ] Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)~

### Reviewer Checklist

- [ ] Code reviewed

--

cc @pivotal-jwinters

### DIFF
--- a/atc/lidar/scanner.go
+++ b/atc/lidar/scanner.go
@@ -105,7 +105,9 @@ func (s *scanner) check(checkable db.Checkable, resourceTypes db.ResourceTypes) 
 
 	_, created, err := s.checkFactory.TryCreateCheck(s.logger, checkable, resourceTypes, version, false)
 	if err != nil {
-		s.logger.Error("failed-to-create-check", err)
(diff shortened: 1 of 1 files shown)


cirocosta:
@pivotal-jwinters , having figured out where the problem comes from, do you think we should still get this in?

cirocosta:
@pivotal-jwinters , I'll actually close this one - we can create a new PR with its own purpose for the fix 😁  thxx!

Issue #3463, opened 2019-03-09
`put` should not insert versions directly - they should only come from `check`


vito:
## What challenge are you facing?

see https://github.com/concourse/git-resource/issues/18, https://github.com/concourse/git-resource/issues/196, https://github.com/concourse/git-resource/issues/221

This is already planned along with concourse/rfcs#1, but I'm gonna log an explicit issue for it as it has user-facing impact.

As of 5.0 we have decoupled build inputs and outputs from resource versions which means we don't have to directly add any entries to the version history - they'll be correlated by the `version_md5` once they're discovered by `check`. Having `put` insert directly into the resource versions table breaks expectations for users who currently use filtering rules on the resource in order to limit `check`, because if they perform a `put` to that same resource with a version that *normally* would have been ignored, the version still gets discovered because the `put` directly inserts it.

## What would make this better?

Stop doing that! Just record it as an output and move on.

This will be backwards-incompatible for any resources that rely on this behavior - for example by having a no-op `check` and a `put` that directly adds versions that propagate to later builds. I think that's OK and a necessary change as it's not really what resources are designed for. For additional context, this was discussed as part of concourse/rfcs#1.

Also related: concourse/rfcs#11, which also would solve this issue by allowing the versions to be recorded but have Concourse know not to trigger any builds for'em.

JohannesRudolph:
Sure - let me break my answer down into two parts because I have another question. What is the implication of this for the "implict get after put" workflow? 

Not stating that I was a particular fan of the implicit get by default but there's probably a lot of pipelines out there that rely on this. These will need a way to migrate to something else. I see two useful features that will be lost/need replacement: 

-`put` emits a new version without requiring a separate `check`, which means that concourse can start triggering other dependent jobs that `get` this resource with `trigger: true` immediately. This is great for reducing pipeline latency.
  - sidenote: the implicit get isn't "useless" from a global perspective because at least one worker will already now have a resource input version volume cached
- `put` with implicit `get` provides a way to generate a new resource version in a job and then immediately continue using that resource version in the same job.  If `put` no longer emits a new version `X`, the only way to work with `X` would be to have a separate `get` step. But how do we guarantee this step fetches `X` and not some other version?
- Related to the previous point, how does another job state a `passed` constraint on a resource emitted by a previous job? How do we achieve safe fan-in semantics (i.e. get me `X` emitted by the previous job)?

[...]

PR #5090, opened 2020-01-30
atc(feat): expose RESOURCE_NAME to steps get and put.

### LINKED ISSUE #5089: Expose RESOURCE_NAME to `put` step

Concourse exposes a set of build metadata, `BUILD_ID`, `BUILD_NAME`, etc. to `get` and `put`, but without current resource name.

When a resource is `putting`, it doesn't know who itself is, which leads to a weird situation, for example:

```yaml
resources:
- name: mr
   type: gitlab-merge-request
   ...

jobs:
- name: job1
  plan:
  - get: mr # get the merge-request metadata, and stores in files.

  - put: mr
     params:
        dir: mr
        status: running
```

The `put` on `mr` doesn't know where `mr` is, so user has to define a param `dir: mr` to tell the `put` where the resource is, which sounds and looks really weird.

If current resource name is exposed to `put`, then `put` can by default look for `/tmp/put/<current resource name>` for itself, so the above weird param `dir: mr` can be avoided.

Yes

evanchaoli:
Fixes #5089 .

Just expose environment variable `RESOURCE_NAME` to `get` and `put` containers.

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/exec/get_step.go
+++ b/atc/exec/get_step.go
@@ -137,12 +137,15 @@ func (step *GetStep) run(ctx context.Context, state RunState) error {
 		return err
 	}
 
+	env := step.metadata.Env()
+	env = append(env, fmt.Sprintf("RESOURCE_NAME=%s", step.plan.Name))
+
 	containerSpec := worker.ContainerSpec{
 		ImageSpec: worker.ImageSpec{
 			ResourceType: step.plan.Type,
 		},
 		TeamID: step.metadata.TeamID,
-		Env:    step.metadata.Env(),
+		Env:    env,
 	}
 
 	workerSpec := worker.WorkerSpec{
--- a/atc/exec/get_step_test.go
+++ b/atc/exec/get_step_test.go
@@ -182,7 +182,7 @@ var _ = Describe("GetStep", func() {
 					ResourceType: "some-resource-type",
 				},
 				TeamID: stepMetadata.TeamID,
-				Env:    stepMetadata.Env(),
+				Env:    append(stepMetadata.Env(), "RESOURCE_NAME=some-name"),
 			},
 		))
 	})
--- a/atc/exec/put_step.go
+++ b/atc/exec/put_step.go
@@ -2,6 +2,7 @@ package exec
 
 import (
 	"context"
+	"fmt"
 	"io"
 
 	"code.cloudfoundry.org/lager"
@@ -141,17 +142,17 @@ func (step *PutStep) run(ctx context.Context, state RunState) error {
 		return err
 	}
 
+	env := step.metadata.Env()
+	env = append(env, fmt.Sprintf("RESOURCE_NAME=%s", step.plan.Name))
+
 	containerSpec := worker.ContainerSpec{
 		ImageSpec: worker.ImageSpec{
 			ResourceType: step.plan.Type,
 		},
-		Tags:   step.plan.Tags,
-		TeamID: step.metadata.TeamID,
-
-		Dir: step.containerMetadata.WorkingDirectory,
-
-		Env: step.metadata.Env(),
-
+		Tags:           step.plan.Tags,
+		TeamID:         step.metadata.TeamID,
+		Dir:            step.containerMetadata.WorkingDirectory,
+		Env:            env,
 		ArtifactByPath: containerInputs,
 	}
 
--- a/atc/exec/put_step_test.go
+++ b/atc/exec/put_step_test.go
@@ -3,7 +3,6 @@ package exec_test
 import (
 	"context"
 	"errors"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/gbytes"
@@ -233,7 +232,7 @@ var _ = Describe("PutStep", func() {
 		}))
 		Expect(actualContainerSpec.Tags).To(Equal([]string{"some", "tags"}))
 		Expect(actualContainerSpec.TeamID).To(Equal(123))
-		Expect(actualContainerSpec.Env).To(Equal(stepMetadata.Env()))
+		Expect(actualContainerSpec.Env).To(Equal(append(stepMetadata.Env(),"RESOURCE_NAME=some-name")))
 		Expect(actualContainerSpec.Dir).To(Equal("/tmp/build/put"))
 
 		Expect(actualContainerSpec.ArtifactByPath).To(HaveLen(3))


evanchaoli:
> Do you mind moving this logic to where the step gets created in the builder instead of in the step itself: https://github.com/concourse/concourse/blob/master/atc/engine/builder/builder.go#L312

@pivotal-jwinters Not sure what you mean. Do you want to put `resource_name` in `stepMetadata`?

vito:
Sorry but I'm pretty opposed to this change. 🙁And sorry for not replying earlier on the original issue proposing this - I've been heads-down trying to get v6.0 ready.

I don't like the idea of resources changing their behavior based on what they're named because it violates the principle that resources are equivalent as long as they're configured the same way (i.e. `type`, `source`, `params`).

Conceptually, this conflates two boundaries. In the Concourse mental model, pipelines are a resource namespace, and resources themselves are just values. To use a programming language metaphor, this is the difference between variable names and variable values. If a language's values changed when you assigned them to a different name, that would feel really strange. 😛

This principle is also important for Concourse's ability to make fairly significant optimizations.

For example, imagine a resource type author implemented `get` to do something different based on the name of the resource, resulting in different bits. Now we would have to take the resource name into account for caching, which would ultimately lead to a lot of wasted disk space. Even if we decide to ignore the name for caching, cache hits would still result in the `get` step being skipped, which would lead to kind of weird semantics where it only runs with whatever the first resource's name was.

Imagining this change also exposed the name to `check`, resource type authors may be tempted to use the resource name to identify the resource in question so that there's less to configure. But then we wouldn't be able to efficiently consolidate version history either, which would lead to a lot of extra `check`ing load across the workers.

Exposing the resource name to `put` doesn't have such significant technical ramifications, especially once we implement #3463, but I still don't think this change is worth the saved line of YAML. I think the behavior proposed in #5089 (using the resource name as a default value for a `dir` param) is a bit too magical and could mislead new users into thinking that resources have some sort of internal state or communication mechanism. It also won't help you anyway if the `get` step is renamed:

```yaml
- get: concourse
  resource: concourse-pr
- put: concourse-pr
  params: {status: pending} # breaks as this needs dir: concourse, not concourse-pr
```

In the long run, I plan to remove all of the other environment variables that we expose to resource containers. That mechanism should be replaced with proper support for notifications, which are really the only things that should have such information.

Given that the original motivation for #5089 appears to be centred around notifications, and the small change here actually has pretty significant implications, I would rather just stay focused on the long-term solution as it'll have an even larger positive impact on the original use case (you wouldn't need the `put` step *at all*). 🙏

evanchaoli:
> And sorry for not replying earlier on the original issue proposing this - I've been heads-down trying to get v6.0 ready.

Understood!

> I don't like the idea of resources changing their behavior based on what they're named because it violates the principle that resources are equivalent as long as they're configured the same way (i.e. `type`, `source`, `params`).

I didn't mean to expose RESOURCE_NAME to `check` at all, nor `get` in #5089. Just when I wrote the code, I felt no harm to expose to `get`, so I added the logic to `get`. I'm totally ok to not expose RESOURCE_NAME to `check` and `get`.

> Exposing the resource name to `put` doesn't have such significant technical ramifications, especially once we implement #3463, but I still don't think this change is worth the saved line of YAML. I think the behavior proposed in #5089 (using the resource name as a default value for a `dir` param) is a bit too magical and could mislead new users into thinking that resources have some sort of internal state or communication mechanism. It also won't help you anyway if the `get` step is renamed:
> 
> ```yaml
> - get: concourse
>   resource: concourse-pr
> - put: concourse-pr
>   params: {status: pending} # breaks as this needs dir: concourse, not concourse-pr
> ```

I never used `resource` to rename a resource, also seldom see a pipeline on our clusters. So this example is valid, but I think it has the least impact.

We just expose a RESOURCE_NAME to `put`, it's up to resource type authors how to use it. A resource may still have a param for `put` to specify where is itself, but exposing RESOURCE_NAME  to `put` at least provides a way for default, it could be apply to most of use cases.

> In the long run, I plan to remove all of the other environment variables that we expose to resource containers. That mechanism should be replaced with proper support for notifications, which are really the only things that should have such information.
> 
> Given that the original motivation for #5089 appears to be centred around notifications, and the small change here actually has pretty significant implications, I would rather just stay focused on the long-term solution as it'll have an even larger positive impact on the original use case (you wouldn't need the `put` step _at all_). 🙏

#5089 isn't really centered around notifications. I have seen many resource types, for example, our home-brew resource types, gitlab, build-web, and so on... By removing redundant param of specifying resource itself's name help make pipeline code neat and easier to understand.

From a user's perspective, if I do get then put:
```yaml
 - get: resource-a
 - put: resource-a
```

It's totally operating the same resource object, it's hard to understand why I still need to specify a redundant param like:
```yaml
 - get: resource-a
 - put: resource-a
    param:
       dir: resource-a
```

What's reason why I think this PR will help pipeline to easier to understand. It's actually a burden for resource type authors to document a pretty explanation for such a param.

For long term, if environment variables are no longer exposed to `get` and `put`, then a number of resource types will have to rewrite, at that time, RESOURCE_NAME should not add much more burden. And what's ETA of the "long term"? I think it's worth letting current resource types enjoy RESOURCE_NAME before the "long term" comes.

In summary, I can remove the logic from `get`, but I still want to solve #5089. @vito, Are you ok with that?

vito:
> #5089 isn't really centered around notifications.

By notifications I mean things like GitHub status, GitLab status, PR status, Slack notifications, etc. - things that update or message some external location with the status of the build. Judging by the code sample in #5089 that seems like the use case:

```yaml
  - put: mr
    params:
      dir: mr
      status: running
```

Updating PR status is pretty much the only situation where I've ended up passing a resource to itself. The whole flow is pretty brittle. For example, in our use of the `github-pr` resource we had to be careful to give the `get` step and the `put` step different names so that the `put` step's implicit `get` step doesn't clobber our original `get` step's bits (which at one point we had given `params`):

https://github.com/concourse/ci/blob/00078c07486b18613f0c5d592e2ff782a2876ab3/pipelines/prs.yml#L105-L110

To be frank, the user experience for notifications is pretty terrible at the moment. It clutters the UI, it's verbose, and it's error-prone. If notifications are the primary motivation, I don't think this change really improves the situation enough to justify increasing the possibility of "gotchas" and misleading new users. The real problem to me is that you shouldn't have to write all those `put` steps in the first place.

With notification support, we would have a way to configure a resource with a notifier. The notifier would run for each state change to a build that uses the resource. The long-term plan is to allow Prototypes to implement a notifier interface, but if we want to take a shortcut we could just implement a few notifier types natively and mark them as 'experimental' like we did with `var_sources`. A basic experimental implementation wouldn't take long to implement and would have a much larger impact.

I'd be happy to help!

(P.S. Honestly it's kind of uncanny how your PRs keep straddling the line of things I haven't yet completed an RFC for but have put vague descriptions on the roadmap. 😅 This is the fourth one in a row - `set_pipeline` step, `var_sources`, `load_var` step, and now notifications. Sorry. You're just about driving v10 from the outside at this point. I owe you many beers.)

evanchaoli:
> (P.S. Honestly it's kind of uncanny how your PRs keep straddling the line of things I haven't yet completed an RFC for but have put vague descriptions on the roadmap. 😅 This is the fourth one in a row - `set_pipeline` step, `var_sources`, `load_var` step, and now notifications. Sorry. You're just about driving v10 from the outside at this point. I owe you many beers.)

Sorry about that. I'm going to stop after #5034. Also hopefully #5091 be merged to 6.0 as well that is a small feature, but I believe it benefits clusters.

vito:
> Sorry about that. I'm going to stop after #5034. Also hopefully #5091 be merged to 6.0 as well that is a small feature, but I believe it benefits clusters.

Haha, I'm the one that's sorry. I didn't mean that to discourage you. I just felt bad about always having some farther-off thing to promise instead of just merging in your contributions. I'm actually really grateful for your work and enjoy collaborating towards a common goal, I was just worried that this pattern is becoming frustrating for you. 😟

Cheers!

evanchaoli:
@vito TBH, I don't have a clear view of "v10" that you mentioned, I'm just driven by my own requirements, where #5034 is the last big feature in short term, that's reason why I said I'm going to stop.

For this PR, I didn't associate it to `notification` at all, I thought it's just an innocent small feature, as I mentioned, we do have some resource types not related to notification at all. Letting a `put` know its own resource name sounds reasonable. I don't think this feature conflicts to `notification`. From my understanding, notification mainly solves the implied get problem. Even if `notification` is released, it requires resource types authors to support the new interface, and pipeline authors to update pipelines, which takes time and effort, so some `put` might be there forever.

I won't argue more, though I'm still not convinced.

evanchaoli:
Closing as this PR is not going to be accepted.

PR #5091, opened 2020-01-30
atc(feat): add --resource-with-webhook-checking-interval to concouse web

### LINKED ISSUE #5088: Add a default resource check interval for webhooked resources

Many pipelines use Gitlab resources, and resource checks poll Gitlab every 1 minute, which generate significant API calls to Gitlab, thus we have received a concern from the Gitlab team. Plus https://github.com/concourse/concourse/issues/4940, I'm consider to prompt usage of webhook.

When a resource has webhook configured, resource checks are supposed to be driven by webhook. Though periodical checks are still there, they don't have to be as short as 1m, maybe 1h is good enough.

But I don't want to ask users to add two more parameters `webhook_token: xxxxx` and `check_every: 1h`. When `webhook_token` is configured, Cocnourse should default check interval to 1h (configured by another atc parameter).

Add a new CLI option to `concourse web`, named `--resource-with-webhook-checking-interval`, default to 1h. When a resource has webhook configured, use this value for its default check interval.

Yes, I can.

evanchaoli:
# Existing Issue

Fixes #5088 .

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/lidar/scanner.go
+++ b/atc/lidar/scanner.go
@@ -20,6 +20,14 @@ func NewScanner(
 	defaultCheckInterval time.Duration,
 	defaultWithWebhookCheckInterval time.Duration,
 ) *scanner {
+	// In case users are not aware of defaultWithWebhookCheckInterval and set defaultCheckInterval
+	// only, make sure defaultWithWebhookCheckInterval to not be smaller than defaultCheckInterval.
+	if defaultWithWebhookCheckInterval < defaultCheckInterval {
+		logger.Info("update-default-with-webhook-check-interface",
+			lager.Data{"oldValue": defaultWithWebhookCheckInterval, "newValue": defaultCheckInterval})
+		defaultWithWebhookCheckInterval = defaultCheckInterval
+	}
+
 	return &scanner{
 		logger:                          logger,
 		checkFactory:                    checkFactory,
--- a/atc/lidar/scanner.go
+++ b/atc/lidar/scanner.go
@@ -26,7 +26,7 @@ func NewScanner(
 	// reason webhooked resources take shorter checking interval than normal
 	// resources.
 	if defaultWithWebhookCheckInterval < defaultCheckInterval {
-		logger.Info("update-default-with-webhook-check-interface",
+		logger.Info("update-default-with-webhook-check-interval",
 			lager.Data{"oldValue": defaultWithWebhookCheckInterval, "newValue": defaultCheckInterval})
 		defaultWithWebhookCheckInterval = defaultCheckInterval
 	}
--- a/atc/lidar/scanner.go
+++ b/atc/lidar/scanner.go
@@ -18,23 +18,26 @@ func NewScanner(
 	secrets creds.Secrets,
 	defaultCheckTimeout time.Duration,
 	defaultCheckInterval time.Duration,
+	defaultWithWebhookCheckInterval time.Duration,
 ) *scanner {
 	return &scanner{
-		logger:               logger,
-		checkFactory:         checkFactory,
-		secrets:              secrets,
-		defaultCheckTimeout:  defaultCheckTimeout,
-		defaultCheckInterval: defaultCheckInterval,
+		logger:                          logger,
+		checkFactory:                    checkFactory,
+		secrets:                         secrets,
+		defaultCheckTimeout:             defaultCheckTimeout,
+		defaultCheckInterval:            defaultCheckInterval,
+		defaultWithWebhookCheckInterval: defaultWithWebhookCheckInterval,
 	}
 }
 
 type scanner struct {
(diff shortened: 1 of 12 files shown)


evanchaoli:
testflight failed, but it passed in my local, so I made a force push to trigger rerun of tests.

evanchaoli:
> To me this seems like a pretty clear [knob anti pattern](https://github.com/concourse/concourse/wiki/Anti-Patterns#knobs).
> 
> I understand that there could be a lot of of resources that have webhooks configured and it's tempting to add a blanket flag to cover them all. However I would argue that since you need to configure the webhook token for each resource anyway, it's probably better for users to make the check interval decision on a per resource basis at that time.

@pivotal-jwinters  If this new option is knob anti pattern, I would argue the existing default check interval option is also knob anti pattern. 

From what I see, most of resources don't configure `check_every`, most of pipelines just want system default behavior. With `webhook_token` configured to a resource, users actually don't care, even don't realize check interval of the resource, thus users wouldn't have motivation to add `check_every`.  IMO, asking users to define `check_every` together with `webhook_token` will lead to several issues:

* users will ask what would be suggested value
* users may forget to add `check_every`
* users may wrongly configure `check_every` to 1m again
* if users switch between webhook and non-webhook, they have to modify `check_every` also

Adding this new option will allow cluster admins to have control over default behavior of webhooked resources, which aims to reduce total checks.

evanchaoli:
I thought over and changed default of `ResourceWithWebhookCheckingInterval` to 1m. So that this PR will not change current behavior at all. It just offers a chance for cluster administrators to reduce checks for web-hooked resources.

evanchaoli:
@vito Can you please take a look at this PR? This is a small feature.

atc/lidar/scanner.go
```diff
@@ -92,6 +103,9 @@ func (s *scanner) check(checkable db.Checkable, resourceTypes db.ResourceTypes,
 	}
 
 	interval := s.defaultCheckInterval
+	if resource, ok := checkable.(db.Resource); ok && resource.HasWebhook() {
```
pivotal-jwinters, line 106: I'd like to avoid this type assertion. The `Checkable` interface should represent everything needed to perform a check, so if we need `HasWebhook` then can we implement this in resourceType as well and always return false.

evanchaoli: Added `HasWebhook` to resource type.

evanchaoli:
@pivotal-jwinters Addressed your comment, pls take a look again.

evanchaoli:
The failure of `concourse-ci/testflight-containerd` seems to not related to my change at all.

atc/lidar/scanner.go
```diff
@@ -18,23 +18,37 @@ func NewScanner(
...
 	defaultCheckInterval time.Duration,
+	defaultWithWebhookCheckInterval time.Duration,
 ) *scanner {
+	// In case that a user configures resource-checking-interval, but forgets to
+	// configure resource-with-webhook-checking-interval, keep both checking-
+	// intervals consistent. Even if both intervals are configured, there is no
+	// reason webhooked resources take shorter checking interval than normal
+	// resources.
+	if defaultWithWebhookCheckInterval < defaultCheckInterval {
+		logger.Info("update-default-with-webhook-check-interface",
```
pivotal-jwinters, line 29:
```suggestion
		logger.Info("update-default-with-webhook-check-interval",
```

atc/lidar/scanner.go
```diff
@@ -18,23 +18,37 @@ func NewScanner(
...
 	defaultCheckTimeout time.Duration,
 	defaultCheckInterval time.Duration,
+	defaultWithWebhookCheckInterval time.Duration,
 ) *scanner {
+	// In case that a user configures resource-checking-interval, but forgets to
+	// configure resource-with-webhook-checking-interval, keep both checking-
+	// intervals consistent. Even if both intervals are configured, there is no
+	// reason webhooked resources take shorter checking interval than normal
+	// resources.
+	if defaultWithWebhookCheckInterval < defaultCheckInterval {
```
pivotal-jwinters, line 28: Sorry one last thing before merging. Can you move this into the atccmd/command.go. It's configuration related, and we try to deal with all configuration at the top level.

evanchaoli:
[trigger ci]

evanchaoli:
@pivotal-jwinters The `watjs` should not relate to my change, please help re-trigger the build. Can you pls grant me pipeline-operator permission, so that I can re-trigger failed build on my own?

PR #5102, opened 2020-01-31
adding pipeline-overview command

reskin89:
# Existing Issue
The fly CLI is great to use, however it is lacking the same type of "view" that the GUI has, i.e. being able to see all the components of a pipeline in one quick press of a key/input of a command.

# Why do we need this PR?
This PR adds the `pipeline-overview` command which outputs in a manner that is comparable to looking at a pipeline in the GUI, with resources AND jobs involved, instead of having to repeatedly use both when using the CLI.  As a heavy user of the CLI myself, I've found this to be a pain point.

A new directive to `fly` has been added: `pipeline-overview alias: po` with one argument: `-p` for pipeline name.

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


clarafu:
Thanks for the PR! 👏 This does seem useful and the code looks pretty good to me. Would you be able to add an integration test for this new fly command in this [integration test suite](https://github.com/concourse/concourse/tree/master/fly/integration)?

reskin89:
You got it, I'll get some tests written and pushed up to this PR today

reskin89:
@clarafu tests added and pass locally :) lmk if there's any other updates I can make.  Thanks!

fly/commands/pipeline_overview.go
```diff
@@ -0,0 +1,147 @@
...
+	err = table.Render(os.Stdout, true)
+	var jobs []atc.Job
+
+	jobs, err = target.Team().ListJobs(pipelineName)
+	if err != nil {
+		return err
+	}
+
+	if command.Json {
+		err = displayhelpers.JsonPrint(jobs)
```
taylorsilva: This never executes when `--json` is used because the of the `return` on line 44. Currently only the resources are printed to stdout, jobs are never printed. 

It might also make sense to put the jobs and resources into one json object so people can do stuff with `jq` like `.jobs.jobName`.

```json
{
  "pipeline_name": "some-pipeline",
  "resources": ...,
  "jobs": ...,
}
```

fly/commands/pipeline_overview.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/concourse/concourse/fly/commands/internal/displayhelpers"
+	"github.com/concourse/concourse/fly/rc"
+	"github.com/concourse/concourse/fly/ui"
+	"github.com/fatih/color"
+)
+
+type PipelineOverviewCommand struct {
+	Pipeline string `short:"p" long:"pipeline" required:"true" description:"Get resources in this pipeline"`
+	Json     bool   `long:"json" description:"Print command result as JSON"`
+}
```
taylorsilva: What do you think about adding flags to allow users to only print either the jobs or resources? Maybe `-r`/`--resources` prints only the resources and `-j`/`--jobs` prints only the jobs.

fly/commands/pipeline_overview.go
```diff
@@ -0,0 +1,147 @@
...
+
+	if command.Json {
+		err = displayhelpers.JsonPrint(resources)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
+	headers = []string{"name", "type", "pinned"}
```
taylorsilva:
```suggestion
	headers = []string{"resource name", "type", "pinned"}
```

What do you think about explicitly stating that the following data is for resources and same for the jobs table? When I ran the command it took me a couple seconds to figure out "is this first table jobs or is the next one jobs..."

reskin89: You got it, I'll make that update, that's a good point

fly/commands/pipeline_overview.go
reskin89: I thought about this, it would almost make sense though, to then get rid of `fly jobs` and `fly resource` and have them all be part of this command.  Is that too big of a change at once though

fly/commands/pipeline_overview.go
reskin89: This is a good idea, I'll work on incorporating this

reskin89: This has been fixed, it now returns a json array listing resources as elements, and then the jobs as elements.  Stuffing them into their own object together seems to get a little dicey due to the amount and the data that returns.  I could potentially just create a new struct and map it to that so that:
```suggestion
	type overview struct {
	  pipeline_name string `json:"pipeline_name"`
	  resources []string      `json:"resources"`
	  jobs [] string               `json:"jobs"`
	  }
```

The only pitfall I see with this is the amount of data.  With how I just updated we're getting this, just in a less nested format.  If I did it the above way I personally would consider reducing the amount of data just for consumption purposes.

Thoughts @taylorsilva ?

fly/commands/pipeline_overview.go
taylorsilva: That would be quite a bigger diff. I'm fine just leaving it as-is

fly/commands/pipeline_overview.go
taylorsilva: Sorry about the late response!

The amount of data may or may not be a lot, always depends on the pipeline. I'm still in favour of making it easier for people to discover how to massage the data with a tool like jq. I think this would end up being the only command that returns two different objects at once (jobs and resources). I'd prefer to have the data mapped to a struct like you mention.

reskin89: Fair points.  I'll get this fixed up in the next day or so and push for final review.  Thanks @taylorsilva !

reskin89: All set @taylorsilva the json output is now the proposed object with a name attribute, and job/resource attributes that are a slice of their respective type.

fly/integration/pipeline_overview_test.go
```diff
@@ -0,0 +1,154 @@
...
+					Data: []ui.TableRow{
+						{{Contents: "resource-1"}, {Contents: "time"}, {Contents: "n/a"}},
+						{{Contents: "resource-2"}, {Contents: "custom"}, {Contents: "some:version", Color: color.New(color.FgCyan)}},
+					},
+				}))
+			})
+		})
+
+		pipelineName := "pipeline"
+		Context("when jobs are returned from the API", func() {
```
taylorsilva: Same as previous comment.

https://github.com/concourse/concourse/blob/aac7b2f461ed2a3ad6fa8d861d439d3774f9fde8/fly/integration/jobs_test.go#L89-L90

fly/integration/pipeline_overview_test.go
```diff
@@ -0,0 +1,154 @@
...
+				Expect(err).NotTo(HaveOccurred())
+
+				<-sess.Exited
+				Expect(sess.ExitCode()).To(Equal(1))
+
+				Expect(sess.Err).To(gbytes.Say("error: the required flag `" + osFlag("p", "pipeline") + "' was not specified"))
+			})
+		})
+
+		Context("when resources are returned from the API", func() {
```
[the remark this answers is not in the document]
reskin89: I can change that, I had replicated it since I'm calling "resources" essentially in `pipeline-overview`, but tackling it directly is probably a more idiomatic approach.

fly/commands/pipeline_overview.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+
+		err = displayhelpers.JsonPrint(resources_and_jobs)
+		if err != nil {
+			return err
+		}
+		// err = displayhelpers.JsonPrint(jobs)
+		// if err != nil {
+		// 	return err
+		// }
```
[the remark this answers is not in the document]
reskin89: :+1:

fly/commands/pipeline_overview.go
```diff
@@ -0,0 +1,156 @@
...
+
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/fly/commands/internal/displayhelpers"
+	"github.com/concourse/concourse/fly/rc"
+	"github.com/concourse/concourse/fly/ui"
+	"github.com/fatih/color"
+)
+
+type PipelineOverviewCommand struct {
+	Pipeline string `short:"p" long:"pipeline" required:"true" description:"Get resources in this pipeline"`
```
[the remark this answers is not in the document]
reskin89: :+1:

reskin89:
@taylorsilva I pushed a test re-write up along with the rest of your suggestions.  Sorry for the lag time, things have been a little nuts.  I'm going to see if the test suite passes, I have an environment issue somewhere in my local environment that's causing ginkgo to fail out on random things not related to my test.

reskin89:
@taylorsilva  I think I'm at an empass where I may need some ginkgo assistance.  My tests to me seem to work, but ginkgo doesn't seem to like them.  I'm not exactly understanding how one portion of the ginkgo "expectation" is different other than some indentation, which my command does not line up the columns of the resources with the jobs (It looks really funky and ambiguous when I do that).  Any tips here would be greatly appreciated

reskin89:
Also seems other tests not related to me are failing now too, I'm wondering if I need a rebase

taylorsilva:
Yeah rebasing might fix those errors. I re-ran them, we'll see if they pass because they all did look unrelated.

Right now I'm only getting one failing test locally on in the `fly/integration` test suite:

```
• Failure [0.148 seconds]
Fly CLI
/Users/taylor/workspace/concourse/concourse/fly/integration/pipeline_overview_test.go:17
  pipeline-overview
  /Users/taylor/workspace/concourse/concourse/fly/integration/pipeline_overview_test.go:18
    when resources and jobs are returned from the API
    /Users/taylor/workspace/concourse/concourse/fly/integration/pipeline_overview_test.go:37
      shows the pipeline's resources and jobs [It]
      /Users/taylor/workspace/concourse/concourse/fly/integration/pipeline_overview_test.go:90
```

Which I think might be fixed by adding a `Header` section like this test: https://github.com/concourse/concourse/blob/60a0724cd8d9006740c1fb92e8acf36397b5628f/fly/integration/containers_test.go#L128

taylorsilva:
Just re-read your comment. Indention has no effect on how the test runs, it shouldn't need to line up with the output, that's what all the Table functions are for.

taylorsilva:
@reskin89 hey, any interest in this? Totally fine if you don't have time for this right now. Wondering if there's anything I can do to unblock you?

A couple of folks on my team mentioned that it would be nice if `pipeline-overview` directly used the code used by `fly jobs` and `fly resources` to reduce the amount of duplicate code. wdyt?

reskin89:
@taylorsilva totally do it’s been a little crazy. I plan to fix the tests this weekend.

Are you referring to just calling the commands for them in the package directly and appending their output together as opposed to just using the same code? If so I think that’s a great idea and can dig into how to get that done properly

taylorsilva:
> Are you referring to just calling the commands for them in the package directly and appending their output together as opposed to just using the same code? If so I think that’s a great idea and can dig into how to get that done properly

Yes, something along those lines! The code in the existing commands (`jobs` and `resources`) might not be set-up to be called like that so you may have to do some refactoring and pull their functionality out into functions outside of the `Execute()` functions that get run by the `go-flags` library. Looking forward to how it goes for you 🙂

taylorsilva:
Hey @reskin89 it's been a while since we touched this. How are you feeling about it?

reskin89:
sorry @taylorsilva things have been super busy lately.  I'm hoping to be able to tackle a re-write of this fairly soon though, still very interested.

taylorsilva:
I'm going to closet this PR for now since it's been almost a year since it was initially opened. Feel free to submit another PR if you get around to working on this again :)

Issue #5063, opened 2020-01-23
Release v6.0


vito:
The outcome of this epic is to make a v6.0 beta available as a GitHub release (specifically a pre-release).

* [x] Upgrade old prod environment
* [x] Upgrade wings.pivotal.io
* [x] Merge `release/6.0.x` into `master`
* [x] Documentation
* [x] Release notes
* [x] Upgrade new prod environment (ci.concourse-ci.org)
  * Wait a bit and monitor the metrics
* [x] Upgrade Hush House environment (hush-house.pivotal.io)
* [x] Once a candidate makes it through the pipeline, ~~manually create a pre-release~~ publish a pre-release using the pipeline

vito:
Upgrading Prod was a little shakey:

```
E 2020-01-28T14:40:03.954240Z [1930268]: [3-1] db=atc,user=atc ERROR:  deadlock detected at character 26 
I 2020-01-28T14:40:03.954833Z [1930268]: [4-1] db=atc,user=atc DETAIL:  Process 1930268 waits for AccessShareLock on relation 16563 of database 16408; blocked by process 1930310. 
I 2020-01-28T14:40:03.955018Z 	Process 1930310 waits for AccessExclusiveLock on relation 16624 of database 16408; blocked by process 1930268. 
I 2020-01-28T14:40:03.955180Z 	Process 1930268: SELECT j.name, j.id FROM jobs j WHERE j.pipeline_id = $1 
I 2020-01-28T14:40:03.955428Z 	Process 1930310: ALTER TABLE pipelines 
I 2020-01-28T14:40:03.955688Z 	    DROP COLUMN schedule_requested, 
I 2020-01-28T14:40:03.955914Z 	    DROP COLUMN last_scheduled 
I 2020-01-28T14:40:03.956088Z [1930268]: [5-1] db=atc,user=atc HINT:  See server log for query details. 
I 2020-01-28T14:40:03.956384Z [1930268]: [6-1] db=atc,user=atc STATEMENT:  SELECT j.name, j.id FROM jobs j WHERE j.pipeline_id = $1 
E 2020-01-28T14:40:04.699603Z [1930130]: [19-1] db=atc,user=atc ERROR:  column "last_scheduled" does not exist at character 81 
I 2020-01-28T14:40:04.700359Z [1930130]: [20-1] db=atc,user=atc STATEMENT:   
I 2020-01-28T14:40:04.700710Z 			UPDATE pipelines 
[...]

PR #5103, opened 2020-01-31
algorithm: fix pagination of rerun builds, add tracing

### LINKED ISSUE #5094: Implement tracing for the algorithm
* [x] Trace algorithm recursion and etc. through the various resolvers
* [x] Trace spans or events for when incremental migrations are performed

vito:
# Existing Issue

Fixes #5094.

The actual bugfix here isn't captured in a separate issue, but there is (some) context in https://github.com/concourse/concourse/issues/5063#issuecomment-580271863.

# Why do we need this PR?

A bug with the algorithm's pagination logic (used for walking over chunks of 100 builds at a time) caused it to get stuck in a loop when looking for builds prior to a rerun.

This PR slings a bunch of code around in service of introducing tracing, which allowed this issue to be discovered and diagnosed pretty easily!

* fix pagination bug
* backfill unit tests for `VersionsDB` covering this scenario and a few others
* introduce tracing, and the ability to use tracing for algorithm tests
* threads `ctx` through `VersionsDB` and uses it to trace the incremental migrations
* adds support for `context.Context` in our `db.Conn` interface, though we don't actually make use of it yet (for e.g. canceling in flight queries). just figured I'd add this along the way.

# Contributor Checklist

- [x] Unit tests

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/versions_db.go
+++ b/atc/db/versions_db.go
@@ -398,8 +398,8 @@ func (versions VersionsDB) NextEveryVersion(ctx context.Context, buildID int, re
 	return nextVersion, false, true, nil
 }
 
-func (versions VersionsDB) LatestBuildPipes(ctx context.Context, buildID int) (map[int]int, error) {
-	rows, err := psql.Select("p.from_build_id", "b.job_id").
+func (versions VersionsDB) LatestBuildPipes(ctx context.Context, buildID int) (map[int]BuildCursor, error) {
+	rows, err := psql.Select("p.from_build_id", "b.rerun_of", "b.job_id").
 		From("build_pipes p").
 		Join("builds b ON b.id = p.from_build_id").
 		Where(sq.Eq{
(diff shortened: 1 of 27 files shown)


atc/db/versions_db.go
```diff
@@ -623,30 +590,73 @@ func (versions VersionsDB) migrateSingle(buildID int) (string, error) {
...
 
+type BuildCursor struct {
+	ID      int
+	RerunOf sql.NullInt64
+}
+
+func (cursor BuildCursor) OlderBuilds(idCol string) sq.Sqlizer {
+	if cursor.RerunOf.Valid {
+		return sq.Or{
+			sq.Expr("COALESCE(rerun_of, "+idCol+") < ?", cursor.RerunOf.Int64),
```
clarafu: I think this won't catch the original build of a rerun (ex, if there are builds 5.1 and 5, it won't grab build 5 if the cursor is 5.1). Do we care about grabbing all the builds that are all reruns of each other? Since they all have the same input versions? Maybe we only care about grabbing latest rerun of a build?

atc/db/versions_db.go
```diff
@@ -623,30 +590,73 @@ func (versions VersionsDB) migrateSingle(buildID int) (string, error) {
...
+	if cursor.RerunOf.Valid {
+		return sq.Or{
+			sq.Expr("COALESCE(rerun_of, "+idCol+") > ?", cursor.RerunOf.Int64),
+			sq.And{
+				sq.Eq{"rerun_of": cursor.RerunOf.Int64},
+				sq.Gt{idCol: cursor.ID},
+			},
+		}
+	} else {
+		return sq.Expr("COALESCE(rerun_of, "+idCol+") > ?", cursor.ID)
```
clarafu: I think this won't catch the rerun builds if the cursor is on the original build. For example, if we have builds 5.2, 5.1, 5 and the cursor id is 5, any reruns of that build will have a rerun_of of 5 (5.1 and 5.2 will have rerun_of = 5) so they won't be fetched from this query because we only grab anything larger than the cursor ID.

atc/db/versions_db.go
vito: > I think this won't catch the original build of a rerun (ex, if there are builds 5.1 and 5, it won't grab build 5 if the cursor is 5.1).

Good catch - I'll add a test for this!

> Do we care about grabbing all the builds that are all reruns of each other? Since they all have the same input versions? Maybe we only care about grabbing latest rerun of a build?

Thought about this - I think we do, because while they all have the same inputs they might end up having different outputs.

atc/db/versions_db.go
vito: > I think this won't catch the rerun builds if the cursor is on the original build. For example, if we have builds 5.2, 5.1, 5 and the cursor id is 5, any reruns of that build will have a rerun_of of 5 (5.1 and 5.2 will have rerun_of = 5) so they won't be fetched from this query because we only grab anything larger than the cursor ID.

:+1: will add a test

PR #5113, opened 2020-02-04
Record and expose pipelines last update date.

### LINKED ISSUE #4809: Record pipelines creation/update time
##  What challenge are you facing?

Sometimes users (or overzealous automation scripts) set pipelines and forget about them forever.
These ghost pipelines stays there, their resources running, but no job was ever triggered and thus is difficult for a human operator as well for an automated tool to distinguish between a pipeline that was created 5 minutes ago and one that's been there for 5 months.
For the above reason is difficult to cleanup these types of pipelines.

By simply adding a column in the `pipelines` table of the DB called `last_updated` that would keep track of when the pipeline is created/updated (e.g. when the set-pipeline has been invoked) and then exposing this datum through `fly pipelines` as well would simplify enormously the job of cleaning up forgotten pipelines.

Also consider that this could benefit #4673 which is now unable to distinguish between just-created-pipelines and long-forgotten-pipelines.
This is also related to #3281. 

Yes: I think I could do it quickly, I just want to gauge consensus :)

aledeganopix4d:
# Existing Issue
Record pipelines creation/update time #4809

Fixes #4809.

* Add a new column into the `pipelines` table called `last_updated`
  * When the pipeline is (re-)set the `last_updated` column is set to the current time
* Add a column in the output of `fly pipelines` to display the last updated date of the pipelines being displayed

# Contributor Checklist
- [X] Unit tests
- [X] Integration tests (if applicable)

# Reviewer Checklist

- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [x] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/migration/migrations/1580135204_add_pipelines_last_update.up.sql
+++ b/atc/db/migration/migrations/1580135204_add_pipelines_last_update.up.sql
@@ -1,0 +1,4 @@
+BEGIN;
+  ALTER TABLE pipelines
+    ADD COLUMN "last_updated" timestamp with time zone NOT NULL DEFAULT '1970-01-01 00:00:00';
+COMMIT;
--- a/atc/api/pipelines_test.go
+++ b/atc/api/pipelines_test.go
@@ -47,6 +47,7 @@ var _ = Describe("Pipelines API", func() {
 				Resources: []string{"resource3", "resource4"},
 			},
 		})
+		publicPipeline.LastUpdatedReturns(time.Unix(1, 0))
 
 		anotherPublicPipeline = new(dbfakes.FakePipeline)
 		anotherPublicPipeline.IDReturns(2)
@@ -54,6 +55,7 @@ var _ = Describe("Pipelines API", func() {
 		anotherPublicPipeline.PublicReturns(true)
 		anotherPublicPipeline.TeamNameReturns("another")
 		anotherPublicPipeline.NameReturns("another-pipeline")
+		anotherPublicPipeline.LastUpdatedReturns(time.Unix(1, 0))
 
 		privatePipeline = new(dbfakes.FakePipeline)
 		privatePipeline.IDReturns(3)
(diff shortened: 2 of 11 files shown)


atc/db/migration/migrations/1580135204_add_pipelines_last_update.up.sql
```diff
@@ -0,0 +1,4 @@
+BEGIN;
+  ALTER TABLE pipelines
+    ADD COLUMN "last_updated" timestamp with time zone NOT NULL DEFAULT '1970-01-01 00:00:00';
```
evanchaoli: Definitely this is a good start! I like this change. But I wonder if we can add a new table, say `pipeline_events` instead of just adding a column to table `pipeline`. I have the PR #4673, where I wanted to record pipeline pause and unpause time. If you add such a `pipeline_events` table, then I can reuse it. The new table will bring an ability of recording pipeline's entire lifecycle.

evanchaoli:
I think it would be great to show the info on GUI also. Do you have a plan to update GUI as well?

atc/db/migration/migrations/1580135204_add_pipelines_last_update.up.sql
pivotal-jamie-klassen: I haven't thought through the scale concerns, but I could imagine https://github.com/concourse/rfcs/pull/33 fitting into a similar extension mechanism.

aledeganopix4d: Thanks for the feedback!
I think that's an interesting idea, however I'm worried that by implementing an event logging for the pipelines the solution would have to be quite more complex: e.g. this log could not increase indefinitely so some rotation would have to be put in place etc...

This PR also comes from what agreed on #4809.
So, for now, I'd rather restraint increasing the scope of this PR so broadly.

aledeganopix4d:
> I think it would be great to show the info on GUI also. Do you have a plan to update GUI as well?

Nope, sorry, I've no idea where to put my hands in the front-end part...

aledeganopix4d:
Is there any blocker with this PR, can someone from Core/UI give some feedback please?

Issue #5099, opened 2020-01-30
containerd: ensure we're not leaking contexts


cirocosta:
### What challenge are you facing?

In our first iteration of using [`context.WithTimeout`] for guaranteeing that
our requests have a bounded amount of time to return, we end up doing so
without taking proper care of their `cancel` functions.

e.g., in 

https://github.com/concourse/concourse/blob/4d739fd1a91866fcd648593bc8be23d6aecbe3f0/worker/backend/backend.go#L103-L108

we're clearly ignoring `cancel` there, making the the context to "leak" until
the timer is reached, i.e.  in the way we're doing, we're keeping contexts
around for longer than they should:

[`context.WithTimeout`]: https://golang.org/pkg/context/#WithTimeout

> The WithCancel, WithDeadline, and WithTimeout functions take a Context (the
> parent) and return a derived Context (the child) and a CancelFunc. Calling the
> CancelFunc cancels the child and its children, removes the parent's reference
> to the child, and stops any associated timers. **Failing to call the CancelFunc
> leaks the child and its children until the parent is canceled or the timer
> fires**. The go vet tool checks that CancelFuncs are used on all control-flow
> paths.

https://github.com/golang/go/blob/0d34f5f09b155a9a7496490d6204f6f18179f098/src/context/context.go#L16-L23

### What would make this better?

Have `cancel` always captured, and call it when not needing that context anymore (the usual `defer cancel()`).

Related:

- https://github.com/concourse/concourse/issues/5051

pivotal-jamie-klassen:
how about a CI job that runs `go vet` to keep us honest?

cirocosta:
@pivotal-jamie-klassen that'd be greeeeat! I remember Topher had https://github.com/golangci/golangci-lint running in a job 👀

cirocosta:
@pivotal-jamie-klassen https://github.com/concourse/ci/issues/263

Issue #5070, opened 2020-01-27
containerd: ensure gracefulperiod is being honored


cirocosta:
### What challenge are you facing?

The way we're currently using `task.Kill` seems to not be really what we want;
apparently the way that `task.Kill` works, the signal gets delivered to the init
process that holds the whole task (`runc init`), and once that one dies.

> **If the "init" process of a PID namespace terminates, the kernel
> terminates all of the processes in the namespace via a SIGKILL**
> signal.  This behavior reflects the fact that the "init" process is
> essential for the correct operation of a PID namespace.
>
> from [`man pid_namespaces`](http://man7.org/linux/man-pages/man7/pid_namespaces.7.html)

meaning that the processes that we created in that container get no chance to be
properly terminated in a given grace period.

While at first this can seem like a "just fine" approach for a method like
`backend.Destroy()`, it's not really a good thing for those who expected Garden
implementors to deliver a SIGTERM so that they could perform some form of
graceful termination.

### What would make this better?

A possible approach would be to first deliver the graceful terminaton signal
(SIGTERM) to the processes that we created (through `task.Exec`), and only then
after we've dealt with the processes, then destroy the sandbox (hold by the init
process - `runc init`).

---

I'll work on a reproducible which we can then have in our integration tests to
make sure that we're really getting the details right and don't regress.

cirocosta:
adding some more context: at the moment, we create containers (which are based
of a rootfs), giving it an init process (that just sits there doing nothing),
and then exec new processes within that sandbox:


1. create the container

	container
		init proc

2. run something

	container
		init proc
		exec'd proc (e.g. /opt/resource/in)


when it comes to gracefully terminating this container, we'd expect the
following to happen:


- exec'd process to receive a signal
- wait few secs for it to go away (otherwise, ungracefully kill it)
- terminate the container (i.e., get rid of `init proc`)

and, it turns out that we can do that without having to list processes in the
system by ourselves:

```go
// taskProcesses retrieves a task's processes.
//
func taskProcesses(ctx context.Context, task containerd.Task) ([]containerd.Process, error) {
        pids, err := task.Pids(context.Background())
        if err != nil {
                return nil, fmt.Errorf("pid listing: %w", err)
        }

        procs := []containerd.Process{}
        for _, pid := range pids {
                if pid.Info == nil { // init
                        continue
                }

                info, err := typeurl.UnmarshalAny(pid.Info)
                if err != nil {
[...]

Issue #5051, opened 2020-01-21
containerd: per-request timeouts


cirocosta:
Hey, as we've seen in #5048, it can be tricky to get context cancellation
working properly in certain scenarios. At the same time, we must ensure that our
requests don't hang forever.

To make sure that we get it right, I'd like to propose a refactor on the way
that we're currently dealing with this aspect of the containerd backend
implementation.

Here are some approaches:

### 1. have our own grpc unary interceptor

```go
// WithDialOpts allows grpc.DialOptions to be set on the connection
func WithDialOpts(opts []grpc.DialOption) ClientOpt {
	return func(c *clientOpts) error {
		c.dialOptions = opts
		return nil
	}
}
```

https://github.com/containerd/containerd/blob/592a617f3d6e3c6aa0f1894c3edfe9a9cf890600/client_opts.go#L67-L73

which we could pass as a DialOption an UnaryInterceptor that does the call to `context.WithTimeout` for us:

```go
// WithUnaryInterceptor returns a DialOption that specifies the interceptor for
// unary RPCs.
func WithUnaryInterceptor(f UnaryClientInterceptor) DialOption {
	return newFuncDialOption(func(o *dialOptions) {
		o.unaryInt = f
	})
}
```

https://github.com/grpc/grpc-go/blob/3311b9ea68a8c6c1554506af68fcbd9fe3e1c74f/dialoptions.go#L425

which, gets used right before "fly"ing the request:

```go
// Invoke sends the RPC request on the wire and returns after response is
// received.  This is typically called by generated code.
//
// All errors returned by Invoke are compatible with the status package.
func (cc *ClientConn) Invoke(ctx context.Context, method string, args, reply interface{}, opts ...CallOption) error {
	// allow interceptor to see all applicable call options, which means those
	// configured as defaults from dial option as well as per-call options
	opts = combine(cc.dopts.callOptions, opts)


	if cc.dopts.unaryInt != nil {
		return cc.dopts.unaryInt(ctx, method, args, reply, cc, invoke, opts...)
	}
	return invoke(ctx, method, args, reply, cc, opts...)
}
```

https://github.com/grpc/grpc-go/blob/3311b9ea68a8c6c1554506af68fcbd9fe3e1c74f/call.go#L29-L38

the caveat here is that this can be *potentially* disruptive as it'll be applied
to literally *ANY* call that is performed, not leaving the possibility to
indivial methods to do things like have its own max request timeout (maybe this
is not a problem though).

[...]

Issue #5049, opened 2020-01-21
containerd: Use default namespace client option


cirocosta:
Currently, for each backend handler we end up populating `ctx` with the
namespace that has been configured for the backend.

https://github.com/concourse/concourse/blob/4bde6a2d8c77c9cb512c0ec641183112b3019603/worker/backend/backend.go#L103-L105

While that's not something computationally expensive, we end up having to test
that for every method of such interface.

https://github.com/concourse/concourse/blob/4bde6a2d8c77c9cb512c0ec641183112b3019603/worker/backend/backend_test.go#L89-L97

It turns out that `containerd` allows us to configure once a default namespace
that gets used for any further uses of that client.

```go
// WithDefaultNamespace sets the default namespace on the client
//
// Any operation that does not have a namespace set on the context will
// be provided the default namespace
func WithDefaultNamespace(ns string) ClientOpt {
	return func(c *clientOpts) error {
		c.defaultns = ns
		return nil
	}
}
```

https://github.com/containerd/containerd/blob/592a617f3d6e3c6aa0f1894c3edfe9a9cf890600/client_opts.go#L40-L49

By doing so, we can ensure that we only set that once, and not have to do that
all the time.

cirocosta:
I was curious about how this actually works under the hood, and, it turns out that just how I suggested that we could inject things into `ctx` from gRPCs interceptors (see https://github.com/concourse/concourse/issues/5051#issue-553018009), that's exactly the same mechanism that `containerd` uses under the hood:

```go
type namespaceInterceptor struct {
        namespace string
}

func (ni namespaceInterceptor) unary(ctx context.Context, method string, req, reply interface{}, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
        _, ok := namespaces.Namespace(ctx)
        if !ok {
                ctx = namespaces.WithNamespace(ctx, ni.namespace)
        }
        return invoker(ctx, method, req, reply, cc, opts...)
}

func (ni namespaceInterceptor) stream(ctx context.Context, desc *grpc.StreamDesc, cc *grpc.ClientConn, method string, streamer grpc.Streamer, opts ...grpc.CallOption) (grpc.ClientStream, error) {
        _, ok := namespaces.Namespace(ctx)
        if !ok {
                ctx = namespaces.WithNamespace(ctx, ni.namespace)
        }

        return streamer(ctx, desc, cc, method, opts...)
}
```

definitely convinced that we should go with this way 👍

PR #5114, opened 2020-02-04
containerd: add container lifecycle and networking

cirocosta:
### why do we need this pr

Most of the container-related methods of the garden backend were not yet
implemented - this PR gets us the ability to have a worker that implements just
enough of the rest of the garden interface to have a usable worker.

In overall, it brings:

- networking to containers (through cni)
- process management (create, signal, destroy, resize)
- proper use of user namespaces (we didn't configure the id mappings before)
- more integration tests

I'm not suuuper proud of how much this PR accumulated in itself though - it's a
culmination of many small chunks of work that I put together through slices of
time I got.

fixes https://github.com/concourse/concourse/issues/5048
fixes https://github.com/concourse/concourse/issues/5049
fixes https://github.com/concourse/concourse/issues/5051
fixes https://github.com/concourse/concourse/issues/5099
fixes https://github.com/concourse/concourse/issues/5070
fixes https://github.com/concourse/concourse/issues/4889

### changes proposed

- add  `killer` and `container_stopper` interfaces to providee the
  functionalities of killing processes, and stopping containers
  - ungraceful & graceful killers
- add a `network` interface, to bring networking to containers
  - CNI as the implementation
  - add file store for managing files that we gotta setup first before getting the
    containers ready (mostly for networking)
- collection of high uid & gid
  - in order to have user namespaces properly setup, we need to provide ID
    mappings in the OCI spec (something we missed before)
- add `cmd/init`, an init process to hold the sandbox together
- perform the setup of context deadlines within libcontainerd itself rather than
  at a higher level


Dockerfile
```diff
@@ -16,9 +16,11 @@ ARG RUNC_VERSION=v1.0.0-rc9
 ARG CNI_VERSION=v0.8.3
 ARG CONTAINERD_VERSION=1.3.2
 
-RUN apt update && apt install -y curl
+# make `ctr` target the default concourse namespace
+ENV CONTAINERD_NAMESPACE=concourse
 
 RUN set -x && \
+	apt install -y curl iptables && \
```
vito: For posterity: while the `firewall` plugin will use `firewalld` and then only fall back on `iptables` if it's not available, and I guess it's not available because we're running in a container. On a real VM this might use `firewalld`? :thinking: 

I wonder what it uses on K8s?

source: https://github.com/containernetworking/plugins/blob/f5c3d1b1bab6ee72a32702c22ebb0f008580ef2d/plugins/meta/firewall/README.md

Dockerfile
```diff
@@ -34,6 +36,12 @@ RUN go build -gcflags=all="-N -l" -o /usr/local/concourse/bin/concourse \
       ./cmd/concourse
 VOLUME /src
 
+
+# build the init executable for containerd
+RUN  set -x && \
+	gcc -O2 -static -o /usr/local/concourse/bin/init ./cmd/init/init.c
```
vito: I vaguely remember during my spike that I didn't actually need an `init` process in order to run things like `fly intercept` because containerd keeps the namespace alive and allows you to run multiple...tasks?... in it.

It looks like I did this by creating an initial task but just never running anything:

https://github.com/concourse/concourse/blob/3321b920f14acdfe12fecb87d0daec8a931233df/gcontainerd/backend.go#L190-L200

I guess this is a little hacky, but I'm just curious if that didn't work or had some weird behavior as a result of the hackiness.

worker/backend/backend.go
```diff
@@ -1,67 +1,102 @@
...
+	return b, nil
 }
 
 // Start initializes the client.
 //
 func (b *Backend) Start() (err error) {
 	err = b.client.Init()
 	if err != nil {
-		return ClientError{InnerError: fmt.Errorf("failed to initialize containerd client: %w", err)}
+		return fmt.Errorf("client init: %w", err)
```
vito: :+1: This kind of wrapping seems more Go-like

worker/backend/backend.go
```diff
@@ -74,225 +109,176 @@ func (b *Backend) Stop() {
...
 
-	container, err := b.client.GetContainer(ctxWithTimeout, handle)
+	err = b.containerStopper.GracefullyStop(ctx, container)
 	if err != nil {
-		return ClientError{InnerError: err}
+		return fmt.Errorf("stopping container: %w", err)
 	}
 
-	task, err := container.Task(ctxWithTimeout, nil)
+	task, err := container.Task(ctx, cio.Load)
```
vito: This will work even after the init process is killed, right? (`GracefullyStop` eventually kills it)

worker/backend/cni_network.go
```diff
@@ -0,0 +1,242 @@
...
+	Subnet string
+}
+
+// defaultCNINetworkConfig is the default configuration for the CNI network
+// created to put concourse containers into.
+//
+var defaultCNINetworkConfig = CNINetworkConfig{
+	BridgeName:  "concourse0",
+	NetworkName: "concourse",
+	Subnet:      "10.80.0.0/16",
```
vito: Do we expose a flag for this? I see this is the default, and there's a way to configure it (`WithCNINetworkConfig`), just wondering if we've bubbled that up, since we'll probably need to for e.g. CI integration tests.

worker/backend/container.go
```diff
@@ -1,146 +1,331 @@
...
+}
+
+// StreamOut - Not Implemented
+func (c *Container) StreamOut(spec garden.StreamOutSpec) (readCloser io.ReadCloser, err error) {
+	err = ErrNotImplemented
+	return
+}
+
+// SetGraceTime - Not Implemented
+func (c *Container) SetGraceTime(graceTime time.Duration) (err error) {
```
vito: We actually use this during gc for hijacked containers

worker/backend/errors.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (e ErrNotFound) Error() string {
+	return "not found: " + string(e)
+}
+
+var (
+	// ErrGracePeriodTimeout indicates that the grace period for a graceful
+	// termination has been reached.
+	//
+	ErrGracePeriodTimeout = errors.New("grace-period-timeout")
```
vito:
```suggestion
	ErrGracePeriodTimeout = errors.New("grace period timeout")
```

worker/backend/cni_network.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+const (
+	// fileStoreWorkDir is a default directory used for storing
+	// container-related files
+	//
+	fileStoreWorkDir = "/tmp"
```
vito: Do/should we ultimately point this at a subdirectory in the Concourse worker's `--work-dir`?

worker/backend/integration/sample/main.go
```diff
@@ -0,0 +1,85 @@
+package main
```
vito: (Can't comment on a binary file...)

Noticed the compiled binary is checked in as `worker/backend/integration/sample/sample` - can we `.gitignore` it and build it during testing instead?

worker/backend/spec/spec.go
```diff
@@ -17,7 +17,7 @@ import (
 // - masked paths
 // - rootfs propagation
 // - seccomp
-// - user namespaces: uid/gid mappings
+// x user namespaces: uid/gid mappings	// TODO test this shit
```
vito: lol

worker/backend/spec/spec.go
```diff
@@ -109,19 +116,41 @@ func OciSpecBindMounts(bindMounts []garden.BindMount) (mounts []specs.Mount, err
 	return
 }
 
+// OciIDMappings provides the uid/gid mappings for user namespaces (if
+// necessary, based on `privileged`).
+//
+func OciIDMappings(privileged bool, max uint32) []specs.LinuxIDMapping {
+	if privileged {
```
vito: I guess we can't make this unconditional until we make `containerd` the sole driver for Linux, because right now it's the ATC and BaggageClaim that does the permission remapping based on whether the container is privileged. If we made this unconditional we would have to change all that.

worker/backend/spec/uidgid.go
```diff
@@ -0,0 +1,121 @@
...
+
+	return MaxValid(f)
+}
+
+// MaxValid computes what the highest possible id in a permission map is.
+//
+// For example, given the following mapping from /proc/self/uid_map:
+//
+// 	0 1001 10
+//      | |    |
```
vito: these `|`s are misaligned :P

worker/workercmd/containerd.go
```diff
@@ -41,27 +55,57 @@ func containerdGardenServerRunner(logger lager.Logger, bindAddr, containerdAddr
...
+	//
+	// - CRI: we're not supposed to be targetted by a kubelet, so there's no
+	//        need to bring up kubernete's container runtime interface plugin.
+	//
+	// - AUFS/BTRFS/ZFS: since linux 3.18, `overlayfs` is in upstream, which
+	//                   most distros should include, so by keeping a focus
+	//                   on a single snapshotter implementation we can better
+	//                   reason about potential problems down the road.
+	//
+	const config = `disabled_plugins = ["cri", "aufs", "btrfs", "zfs"]`
```
vito: :+1: good idea

worker/backend/integration/sample/main.go
cirocosta: whoops hehe, totally!

cirocosta: rebased w/out the binary and added to gitignore 👍

Dockerfile
cirocosta: > I guess this is a little hacky, but I'm just curious if that didn't work or
> had some weird behavior as a result of the hackiness.

not that I remember :thinking:

going 1:1  with guardian, I tried following closely what they had, so it seemed
natural to go with our own init process.

howeeever, it seems like it's not really hacky to not bring our own init, even
according to the spec:


> status (string, REQUIRED) is the runtime state of the container.
> The value MAY be one of:
>
> **creating**: the container is being created (step 2 in the lifecycle) 
>
> **created**: the runtime has finished the create operation (after step 2 in the
> lifecycle), and the container process has neither exited nor executed the
> user-specified program 
>
> **running**: the container process has executed the user-specified program but
> has not exited (after step 5 in the lifecycle)
>
> see [oci/runtime.md](https://github.com/opencontainers/runtime-spec/blob/1ca19ac3226fbbba0ca30fe2d082bc0d08144152/runtime.md#state)

i.e., by not running `task.Start`, where's keeping oursselves there in the
`created` state, with `runc init` being PID 1.

which, is indeed what happens:


- in a container that has `task.Start`:

	cat /ns-proc/1/cmdline
	/tmp/gdn-init

	runc list (...)

	ID	       STATE     	BUNDLE
	b8240aff(..)   running   	/run/cont...d.runtime.v2.task/test15809...

- in a container that DOES NOT have `task.Start` issued


	cat /ns2-proc/1/cmdline | tr '\000' ' '
	runc init

	ID		STATE		BUNDLE
	a50ce69c(..)    created   	/run/cont...5f-daaf80c7ac5e...

Something that might be interesting to do then is even remove that `args`
section in the container config altogethr - from the spec's perspective, it
seems like a reasonable thing to do:

> process (object, OPTIONAL) specifies the container process. This property is
> REQUIRED **when start is called**.
>
> see [oci/config.md](https://github.com/opencontainers/runtime-spec/blob/1ca19ac3226fbbba0ca30fe2d082bc0d08144152/config.md#process)


(tl;dr: yeah, I think we can get rid of `task.Start` there and not have to bring
our own `init`, which is great!)

worker/backend/backend.go
cirocosta: oooooh good catch! the "trying to load the task" would "work" - it'd always return the "not found" that we catch right after the `if err`, which means that in practice we'd always skip the network tear down :sweat_smile:

Dockerfile
cirocosta: yeah .. I don't really know :thinking:  I think we can get our answer from https://github.com/containerd/cri - I'll add to the list of next steps :+1:  

my guess is that even within a container, we could get it set up, although that'd require us having that daemon up and running (could be an `ifrit.Runner` taking care of it) - if the perf is really convincing, I can totally see that being a valid approach (as far as I remember the big problem w/ iptables is how it has to reload all of its rules when you load something in, while with firewalld, that's on a "diff by diff" basis)

(btw, just found this: https://firewalld.org/documentation/concepts.html)

Dockerfile
cirocosta: @vito I was trying to see why you could get away w/out specifying `args` - as you were making the backend image aware already, it was fetching that from the image config (I think):

https://github.com/concourse/concourse/blob/3321b920f14acdfe12fecb87d0daec8a931233df/gcontainerd/backend.go#L160-L161

thus, not being empty / pointing to something nonexistent :thinking:

cirocosta: (the surprising behavior of `runc` trying to `start` during a `create`):

```
START?  github.com/opencontainers/runc/libcontainer.(*linuxContainer).start(0xc0000940f0, 0xc0000b4dc0, 0x0, 0x1)
                /go/src/github.com/opencontainers/runc/libcontainer/container_linux.go:348 +0x5c7
        github.com/opencontainers/runc/libcontainer.(*linuxContainer).Start(0xc0000940f0, 0xc0000b4dc0, 0x0, 0x0)
                /go/src/github.com/opencontainers/runc/libcontainer/container_linux.go:242 +0xaf
        main.(*runner).run(0xc0001215c0, 0xc0000ee2a0, 0x0, 0x0, 0x0)
                /go/src/github.com/opencontainers/runc/utils_linux.go:320 +0xcf7
        main.startContainer(0xc0000cc9a0, 0xc0000de080, 0x1, 0x0, 0x0, 0xc0000b16b0, 0x6f4185)
                /go/src/github.com/opencontainers/runc/utils_linux.go:452 +0x43a
        main.glob..func2(0xc0000cc9a0, 0x10100, 0xc0000cc9a0)
CREATE          /go/src/github.com/opencontainers/runc/create.go:65 +0xae
        github.com/opencontainers/runc/vendor/github.com/urfave/cli.HandleAction(0x83f500, 0x8db5c0, 0xc0000cc9a0, 0xc000088700, 0x0)
                /go/src/github.com/opencontainers/runc/vendor/github.com/urfave/cli/app.go:490 +0xc8
        github.com/opencontainers/runc/vendor/github.com/urfave/cli.Command.Run(0x8c336c, 0x6, 0x0, 0x0, 0x0, 0x0, 0x0, 0x8c7a88, 0x12, 0x0, ...)
                /go/src/github.com/opencontainers/runc/vendor/github.com/urfave/cli/command.go:210 +0x991
        github.com/opencontainers/runc/vendor/github.com/urfave/cli.(*App).Run(0xc00007d1e0, 0xc000094000, 0xf, 0xf, 0x0, 0x0)
                /go/src/github.com/opencontainers/runc/vendor/github.com/urfave/cli/app.go:255 +0x6ab
        main.main()
                /go/src/github.com/opencontainers/runc/main.go:145 +0xab6
```

worker/backend/container.go
cirocosta: https://github.com/concourse/concourse/issues/5134

worker/backend/backend.go
cirocosta: refactored the whole thing a bit to let us perform the deletion not during the killing phase (it was the "containerstopper" responsability), but instead, do after the network teardown is performed.

Dockerfile
cirocosta: https://github.com/concourse/concourse/issues/5135

Dockerfile
cirocosta: https://github.com/concourse/concourse/issues/5136

worker/backend/cni_network.go
cirocosta: https://github.com/concourse/concourse/issues/5140

worker/backend/rootfs_manager.go
```diff
@@ -0,0 +1,70 @@
...
+// configuration of a rootfsManager.
+//
+type RootfsManagerOpt func(m *rootfsManager)
+
+// WithMkdirAll configures the function to be used for creating directories
+// recursively.
+//
+func WithMkdirAll(f func(path string, mode os.FileMode) error) RootfsManagerOpt {
+	return func(m *rootfsManager) {
+		m.mkdirall = f
```
kcmannem: curious, when would you use a different mkdir function?

cirocosta: gooood question - the rationale was that I wanted to be able to verify that `mkdirall` was  called only a certain number of times:

https://github.com/concourse/concourse/blob/54991b2f1f17bd7d229a3ebbe6d6c16f3ff495a3/worker/backend/rootfs_manager_test.go#L41-L48

it could also be used to inject errors (I actually should've done that in the tests to test some failure scenarios 🤔)

worker/backend/spec/spec.go
```diff
@@ -109,19 +99,61 @@ func OciSpecBindMounts(bindMounts []garden.BindMount) (mounts []specs.Mount, err
...
+		{ // "root" inside, but non-root outside
+			ContainerID: 0,
+			HostID:      max,
+			Size:        1,
+		},
+		{ // anything else, not root inside & outside
+			ContainerID: 1,
+			HostID:      1,
+			Size:        max - 1,
+		},
```
aoldershaw: I'm not totally getting this part - why do we map to IDs `1 -> (max-1)` on the Host? I get that they're all just non-root uids, but does this not conflict with other containers' mappings (or are they all isolated)?

worker/backend/cni_network.go
```diff
@@ -0,0 +1,242 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("creating /etc/resolv.conf: %w", err)
+	}
+
+	return []specs.Mount{
+		{
+			Destination: "/etc/hosts",
+			Type:        "bind",
+			Source:      etcHosts,
+			Options:     []string{"rbind", "rw"},
```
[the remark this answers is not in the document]
cirocosta: [@aoldershaw  and I]: oh, this should actually be `bind` 😁

worker/backend/user_namespace.go
```diff
@@ -0,0 +1,131 @@
...
+		_, err := fmt.Sscanf(
+			scanner.Text(),
+			"%d %d %d",
+			&inside, &outside, &size,
+		)
+		if err != nil {
+			return 0, fmt.Errorf("scanf: %w", err)
+		}
+
+		val = minUint(maxUint(val, inside+size-1), math.MaxUint32)
```
aoldershaw: I think the `minUint` call is unneeded, since `maxUint(...)` returns a `uint32`, so it **can't** exceed `math.MaxUint32`

worker/backend/spec/spec.go
cirocosta: yeah, the mappings *will* indeed conflict: from the host perspective, it's
totally expected (in this implementation) that `user 123` in `containerA` (which
maps to `user 123` in the `host` user namespace) will be the same as `user 123`
in `containerB`.

While *not* overlapping is something that we *could* do,  that's the current
behavior of [`gdn`] - thus, wouldn't really fall into this PR, if we wanted to
get to it.

Also, [`baggageclaim`] performs the same assumptions (see
[`uidgid/mapper_linux.go`], so we'd probably need to make some changes there
too.

[`gdn`]: https://github.com/cloudfoundry/guardian
[`baggageclaim`]: https://github.com/concourse/baggageclaim
[`uidgid/mapper_linux.go`]: https://github.com/concourse/baggageclaim/blob/df945db3059d4f9ace9284994f37b625c3fd864e/uidgid/mapper_linux.go#L13-L41

pivotal-jamie-klassen:
Late to the party, but I wanted to suggest `SuperuserPath` instead of `RootPath`, to avoid conflicts with the (already-overloaded) notion of `privileged: true`.

worker/backend/backend_test.go
```diff
@@ -260,240 +251,55 @@ func (s *BackendSuite) TestLookupGetContainer() {
...
-	fakeTask := new(libcontainerdfakes.FakeTask)
-
-	fakeTask.WaitStub = exitBeforeTimeout
-	fakeContainer.TaskReturns(fakeTask, nil)
-	s.client.GetContainerReturns(fakeContainer, nil)
-
-	fakeTask.KillReturns(nil)
-	deleteStatus := containerd.NewExitStatus(1, time.Now(), errors.New("delete failed"))
-	fakeTask.DeleteReturns(deleteStatus, nil)
+// func (s *BackendSuite) TestDestroyGracefullyStopErrors() {
```
aoldershaw: Replace these tests to test the new Killer

worker/backend/container_test.go
```diff
@@ -0,0 +1,185 @@
...
+	s.killer = new(backendfakes.FakeKiller)
+
+	s.container = backend.NewContainer(
+		s.containerdContainer,
+		s.killer,
+		s.rootfsManager,
+	)
+}
+
+// func (s *ContainerSuite) TestStopWithKillUngracefullyStops() {
```
aoldershaw: Replace these tests with tests for the new Killer

worker/backend/backend_test.go
cirocosta: https://github.com/concourse/concourse/issues/5184

worker/backend/container_test.go
cirocosta: https://github.com/concourse/concourse/issues/5184

PR #5117, opened 2020-02-04
algorithm: prevent wasteful recursion on same candidates

vito:
# Existing Issue

n/a; encountered during Hush House environment upgrade to 6.0.0-rc.11

# Why do we need this PR?

Say you have a pipeline like so:

    {r1}--[j1]--{r1}--[j2]
    {r2}-/

* [j1] is a job with two inputs, {r1} and {r2}
* [j2] is a downstream job which only cares about the {r1} input.

If {r2} generates new versions much more frequently than {r1}, there
will be potentially many more builds of [j1] for [j2] to consider during
the algorithm, many of which having the same version of {r1} (because it
doesn't change as often).

With an unhappy pipeline, where the `passed` constraint can't be
satisified quickly, this can result in a lot of wasteful churn as the
algorithm keeps attempting the same version of {r1} over and over until
it eventually gets to a build that has a different version of it.

If you increase the complexity of this pipeline by adding many more jobs
upstream of [j2], each with many builds of their own each because they
also have another resource which makes them run more often, this can
lead to the algorithm running for a very very long time.

To mitigate this, we can just make note of whichever previous set of
candidates we tried and failed. If we're about to recurse with the exact
same set of candidates, we can just short-circuit, saving on a lot of
wasteful recursion.

A regression test has been added, which currently completes in ~30
seconds. I don't really know how long it would run for before because I
didn't have the patience.

(note: if you run the regression test with tracing enabled you can see it prevents recursing ~7k times)

* when we recurse and fail to resolve, mark the current set of candidates as "doomed"
* before recursing, check if the current candidates are "doomed" and just bail early

# Contributor Checklist

- [x] Unit tests

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


pivotal-jamie-klassen:
fwiw adding 30 seconds to a test suite sounds pretty painful.

I poked for a minute at the `unit` logs on ci.concourse-ci.org, it wasn't immediately obvious to me what the typical running time of the algorithm test suite is, so maybe that kind of running time is a necessary evil, but it raises a pretty big red flag for my TDD-based developer-experience spidey-sense.

For bystanders as much as myself: @vito does 30 seconds heavily impact your experience when hacking on algorithm stuff? Should we do something to move this sort of thing out of the way of a TDD loop? e.g. the elm code has a separate benchmark suite that only runs on CI.

vito:
@pivotal-jamie-klassen We can move these regression tests to a separate benchmark suite or make them only run if an env var is set or something. :+1: 

Most algorithm tests are table-driven with bespoke handcrafted scenarios. Those tests run pretty quickly, even though they depend on the database now.

The regression tests, however, are run against anonymized data from environments which exhibited either broken or extremely slow behavior in exotic edge cases. Having these tests around gives me great comfort in refactoring, redesigning, and optimizing the algorithm.

To me, the 30 seconds is a small price to pay; with something as mission-critical as the algorithm and so many different scenarios possible in the outside world, it's comforting to know that we have real-world examples captured and tested against.

I usually just run the full suite for good measure and focus the particular test I'm working on until the end, so the regression tests don't really affect me much, but it could make sense to make them opt-in for the sake of faster PR builds and such.

Issue #5123, opened 2020-02-06
Unify concurrency pattern across job scheduler, build tracker, and LIDAR checker


vito:
These three components are converging on a very similar flow:

1. Find all the things to do,
1. Kick them all off concurrently with an arbitrary max-in-flight to prevent overwhelming the system,
1. Make sure we're not "piling on" the same work through repeated component ticks.

Can we extract this into a common interface?

Currently the LIDAR checker and build tracker don't have a max in flight, but it probably makes sense to add one, for the same reason we added one for job scheduling.

As of https://github.com/concourse/concourse/issues/5092, all three components now perform the same `*sync.Map` trick to accomplish step 3.

Issue #5122, opened 2020-02-06
Refactor scheduler interfaces


vito:
The [scheduler runner](https://github.com/concourse/concourse/blob/8f75b035f398367468c58018df337f1cede61f59/atc/scheduler/runner.go) currently does a lot of fancy footwork to get data passed around in ways for the scheduler and algorithm to consume.

First, we get all the jobs to schedule...:

https://github.com/concourse/concourse/blob/8f75b035f398367468c58018df337f1cede61f59/atc/scheduler/runner.go#L50-L53

This part is fine, but then we make this complicated job-to-pipeline and pipeline-to-job mapping:

https://github.com/concourse/concourse/blob/8f75b035f398367468c58018df337f1cede61f59/atc/scheduler/runner.go#L171-L197

We later loop over all the *pipelines* and "schedule" each pipeline:

https://github.com/concourse/concourse/blob/8f75b035f398367468c58018df337f1cede61f59/atc/scheduler/runner.go#L61-L68

To schedule a pipeline, we fetch all of its resources and fetch all of its jobs, and then construct a job-name-to-id map:

https://github.com/concourse/concourse/blob/8f75b035f398367468c58018df337f1cede61f59/atc/scheduler/runner.go#L86-L89

We then pass the job we're scheduling, along with its pipeline, the pipeline's resources, and this job-name-to-id map down to the actual scheduler:

https://github.com/concourse/concourse/blob/8f75b035f398367468c58018df337f1cede61f59/atc/scheduler/runner.go#L141-L148

The first thing the scheduler does is to get the job's inputs:

https://github.com/concourse/concourse/blob/8f75b035f398367468c58018df337f1cede61f59/atc/scheduler/scheduler.go#L38-L41

This method returns a slice of `atc.JobInput` types, which have **names**, not **IDs** for its references resources and jobs. So the query converts the IDs in the database to names:

https://github.com/concourse/concourse/blob/8f75b035f398367468c58018df337f1cede61f59/atc/db/job.go#L194-L203

The algorithm then converts those **names** back into **IDs**:

https://github.com/concourse/concourse/blob/8f75b035f398367468c58018df337f1cede61f59/atc/scheduler/algorithm/resolver.go#L53-L60

https://github.com/concourse/concourse/blob/8f75b035f398367468c58018df337f1cede61f59/atc/scheduler/algorithm/resolver.go#L86-L90

Going back up the stack, the scheduler also passes the resources to two places:

[...]

PR #5124, opened 2020-02-06
scheduler: don't wait on jobs completing

### LINKED ISSUE #5092: Don't allow a stuck job scheduling to block other jobs from scheduling
Well, we thought this was unlikely, but bugs happen, and now it's happening in prod!

Let's have job scheduling behave the same as build tracking and resource checking - just keep track of which jobs are already being scheduled by tracking it in a `sync.Map`.

vito:
# Existing Issue

Fixes #5092.

# Why do we need this PR?

Currently, if a job reveals a regressive performance case with the new algorithm, that will prevent all other jobs from scheduling.

We originally made scheduling not wait, but that led to piling on and scheduling the same job a bunch of times concurrently if it took more than the scheduler tick interval.

Instead of waiting for all the jobs to complete scheduling within the tick, just prevent scheduling the same job by keeping track of them in a `*sync.Map`.

There is further refactor opportunity here, captured in #5122 and #5123.

# Contributor Checklist

- [x] Unit tests

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/scheduler/runner.go
+++ b/atc/scheduler/runner.go
@@ -8,7 +8,6 @@ import (
 
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc/db"
-	"github.com/concourse/concourse/atc/db/lock"
 	"github.com/concourse/concourse/atc/metric"
 	"github.com/concourse/concourse/atc/scheduler/algorithm"
 )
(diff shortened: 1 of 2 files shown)


atc/scheduler/runner.go
```diff
@@ -88,27 +93,37 @@ func (s *schedulerRunner) schedulePipeline(ctx context.Context, errGroup *multie
...
+			if err != nil {
+				jLog.Error("failed-to-acquire-lock", err)
+				return
+			}
+
+			if !acquired {
+				return
+			}
+
+			err = s.scheduleJob(ctx, logger, schedulingLock, pipeline, job, resources, jobsMap)
```
pivotal-jwinters: Can we defer release the schedulingLock so we don't need to pass it in here?

vito: sure can!

PR #5148, opened 2020-02-10
web: lazy-load non-essential javascript

### LINKED ISSUE #5131: web: eliminate render-blocking resources on page load
Based on a [Lighthouse report](https://googlechrome.github.io/lighthouse/viewer/?gist=f2cbe304e59dfd99b3ba25a456e8e466) run on the dashboard of our CI deployment (https://ci.concourse-ci.org/), we can see that the performance isn't great - 5.5 seconds to first contentful paint.

Lighthouse suggests that we can achieve a good deal of savings by "eliminating render-blocking resources". Anything that is not required for first render should ideally be deferred (if possible/not too difficult).

Screen Shot 2020-02-07 at 9 58 08 AM

## Which resources are necessary for initial render

* `main.css` - there's a case to be made for inlining critical styles and asynchronously loading the rest of the css, but given that our css is pretty small, it might not be worth the effort
* ~`index.js`~ - needed for rendering pipelines using d3. Could be dynamically imported if feasible
* ~`d3.v355.min.js`~ - needed for rendering pipelines into svgs. Could be dynamically imported if feasible
* ~`graph.js`~ - used by `index.js` to lay pipelines out. Could be dynamically imported or moved into `index.js` to have 1 fewer HTTP request
* ~`jquery-2.1.1.min.js`~ - I suspect we don't need to use jquery at all. We'll need to change some of the code in `index.js` and `elm-setup.js` to accomodate for this, though.
* `elm.min.js` - The compiled elm code is necessary for first render
* `elm-setup.js` - Sets up all the ports required for elm. Necessary for first render
*  `clipboard.min.js` - Probably not necessary, but Lighthouse doesn't seem to think the impact is large
* ~`mdi-svg.min.js`~ - needed for rendering icons on inputs/outputs in the pipeline graph. Could be dynamically imported. Additional note: somewhat counterintuitively, wrapping the massive `icon_paths` JSON object in a `JSON.parse(...)` (rather than treating it as a JS object literal) can improve performance: see https://www.youtube.com/watch?v=ff4fgQxPaO0

Note: by "dynamically importing", I had this feature in mind: https://v8.dev/features/dynamic-import#dynamic. It has decent browser support: https://caniuse.com/#feat=es6-module-dynamic-import. I'm definitely open to other suggestions for ways to achieve this.

aoldershaw:
# Existing Issue

Fixes #5131 .

# Why do we need this PR?

Currently, we load a lot of Javascript that isn't essential for the first render, and as a result, our [first contentful paint time](https://googlechrome.github.io/lighthouse/viewer/?gist=d6d0c36bd7e11829963bc7fb79673903) takes a significant hit. For instance, we load in all of the code required for rendering the pipeline graph (d3.js + custom logic) prior to rendering the dashboard, despite this code only being required after we click into a pipeline.

After lazy-loading these resources, we achieve a [much better first contentful paint time](https://googlechrome.github.io/lighthouse/viewer/?gist=f9227f86e80bcc48aa34cde157dfc7e6).

* Dynamically import (lazy-load) several modules that aren't essential for the critical rendering path (in particular: d3.js, graph rendering logic, mdi icons).
* Remove uses of jQuery

~**Note:** this change will (likely) reduce browser support - currently, [dynamic imports are only supported by ~86.1% of clients](https://caniuse.com/#feat=mdn-javascript_statements_import_dynamic_import). This support can be improved by running our Javascript through Babel and the [Babel dynamic import plugin](https://babeljs.io/docs/en/babel-plugin-syntax-dynamic-import), but I'm not sure whether or not this is necessary - dynamic imports are supported by Chrome 63 and beyond (Dec 2017), Safari 11.1 and beyond (March April 2018), Firefox 67 and beyond (May 2019).~

**EDIT:** I added a Babel transpilation step to support older browsers. See https://github.com/concourse/concourse/pull/5148/commits/2f6fe63fc3b98409545b51073cebcecb90b5597d

**Note:** in order to import d3.js in a module, I had to manually modify the d3.js bundle (replacing the final argument-less function call with `.apply(self)`). If we ever need to update d3.js, we'll need to keep this in mind.

# Contributor Checklist
- [ ] ~Unit tests~
- [ ] ~Integration tests (if applicable)~

# Reviewer Checklist
- [x] Code reviewed
- [ ] ~Tests reviewed~
- [ ] ~Documentation reviewed~
- [ ] Release notes reviewed
- [x] PR acceptance performed
- [ ] ~New config flags added? Ensure that they are added to the [BOSH](https://github.com/concourse/concourse-bosh-release) and [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for the [integration tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env) (for example, if they are Garden configs that are not displayed in the `--help` text).~


aoldershaw:
@pivotal-jamie-klassen thanks, missed that. The problem is that the all the Javascript did not previously run in strict mode, meaning variables could be defined out of nowhere (without declaring them using `var`, `let`, or `const`), but strict mode is always on when using modules.

jomsie:
this is awesome stuff! (also good snag jamie, and i know you're assigned this PR :) )

i just did some quick testing of this pr (just by running concourse locally and running some pipelines), and it seems to work fine

aoldershaw:
@pivotal-jamie-klassen I was actually a little worried about reducing browser support, so I added a Webpack config that generates a bundle that should work on old browsers. Note that it still uses the native ES6 modules where supported.

Let me know what you think of the change

jomsie:
looks like watsjs is taking a bit of time >_>

aoldershaw:
> looks like watsjs is taking a bit of time >_>

Weird, it's failed twice with

```
ERROR: Couldn't connect to Docker daemon at http+docker://localhost - is it running?
```

aoldershaw:
Figured out the issue, which I describe in more detail here: https://github.com/concourse/concourse/pull/5148/commits/6e558a8dcafa711ad4fd8aadbdb0beb2a684d093

In short, loading `elm-setup.js` as `type="module"` changes how the browser loads the script, which results in the Javascript side not being subscribed to ports when Elm sends messages to them, leading to lost messages. This results in HTTP 401s, since the CSRF token doesn't get saved to local storage.

@pivotal-jamie-klassen the PR's ready for re-review when you get the chance.

PR #5151, opened 2020-02-10
Assign pipeline operator role to RerunJobBuild route

### LINKED ISSUE #5115: non-admins get 403s when rerunning builds
## Bug Report

If you're not a cluster admin you get a 403 when hitting the `RerunJobBuild` endpoint.

## Steps to Reproduce

1. Set up a team with a `member`
1. Log in as the `member`
1. Create a build of a job on a pipeline in that team.
1. Click the re-run button

## Expected Results

re-run is queued 😎 

## Actual Results

403 in the browser console 😢 

## Additional Context

`atc.RerunJobBuild` does not appear in the role action map https://github.com/concourse/concourse/blob/master/atc/api/accessor/role_action_map.go.

## Version Info

* Concourse version: 6.0.0-pre
* Deployment type (BOSH/Docker/binary): k8s
* Infrastructure/IaaS: gke
* Did this used to work? No

clarafu:

# Existing Issue
The rerun build endpoint was not added to the action role map so only admin users were allowed to rerun builds.

Fixes #5115.

# Why do we need this PR?
The correct permissions for rerunning builds should be the same creating builds, which is a pipeline-operator role action so rerunning builds was added to the role action map for a pipeline-operator role.

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/api/accessor/role_action_map.go
+++ b/atc/api/accessor/role_action_map.go
@@ -1,9 +1,10 @@
 package accessor
 
 import (
+	"io/ioutil"
+
 	"code.cloudfoundry.org/lager"
 	"gopkg.in/yaml.v2"
-	"io/ioutil"
 
 	"github.com/concourse/concourse/atc"
 )
@@ -24,6 +25,7 @@ var requiredRoles = map[string]string{
 	atc.GetBuildPreparation:           "viewer",
 	atc.GetJob:                        "viewer",
 	atc.CreateJobBuild:                "pipeline-operator",
+	atc.RerunJobBuild:                 "pipeline-operator",
 	atc.ListAllJobs:                   "viewer",
 	atc.ListJobs:                      "viewer",
 	atc.ListJobBuilds:                 "viewer",
--- a/atc/api/accessor/accessor_test.go
+++ b/atc/api/accessor/accessor_test.go
@@ -1,15 +1,16 @@
 package accessor_test
 
 import (
-	"code.cloudfoundry.org/lager"
 	"crypto/rand"
 	"crypto/rsa"
 	"fmt"
+	"net/http"
+
+	"code.cloudfoundry.org/lager"
 	"github.com/dgrijalva/jwt-go"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
-	"net/http"
 
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/api/accessor"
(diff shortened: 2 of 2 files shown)


atc/api/accessor/role_action_map.go
```diff
@@ -17,6 +17,7 @@ var requiredRoles = map[string]string{
 	atc.GetCheck:                      "viewer",
 	atc.GetBuildPlan:                  "viewer",
 	atc.CreateBuild:                   "member",
+	atc.RerunJobBuild:                 "member",
```
cirocosta: Heey,

I'm not familiar with this part of the code, but, should we then increase the coverage in the accessor tests?

https://github.com/concourse/concourse/blob/991453c585ea542fe8ab76fd5f7ad22002fdfbb9/atc/api/accessor/accessor_test.go#L440-L450

thanks!

clarafu: I just added a test for ensuring that all endpoints have a role assigned to it! It actually found a few that didn't have a role lol

atc/api/accessor/role_action_map.go
```diff
@@ -75,6 +75,7 @@ var requiredRoles = map[string]string{
 	atc.ListWorkers:                   "viewer",
 	atc.DeleteWorker:                  "member",
 	atc.SetLogLevel:                   "member",
+	atc.ListActiveUsersSince:          "member",
```
pivotal-jamie-klassen: I may be wrong, but I think this is supposed to be for admins only?

atc/api/accessor/role_action_map.go
```diff
@@ -96,14 +97,17 @@ var requiredRoles = map[string]string{
 	atc.CreateArtifact:                "member",
 	atc.GetArtifact:                   "member",
 	atc.ListBuildArtifacts:            "viewer",
+	atc.GetWall:                       "viewer",
+	atc.SetWall:                       "member",
+	atc.ClearWall:                     "member",
```
pivotal-jamie-klassen: ditto, admin only

atc/api/accessor/role_action_map.go
clarafu: I also might be wrong haha but I thought the admin privileges were checked by another handler https://github.com/concourse/concourse/blob/e3a5020891b8f0858b31cf4f504a6789a0662c19/atc/wrappa/api_auth_wrappa.go#L116-L122

clarafu: @pivotal-jamie-klassen Ok I think you're right, I talked to Josh and he said that it doesn't make sense for these endpoints to even have a "team role" because it is more of a concourse-wide role as admin.

I think I'm going to change my test to ignore any that requires admin privileges, thanks a lot for your help!

pivotal-jwinters:
@clarafu a lot of this is changing in #4950, so I'm a little hesitant to make major changes here.

That giant role map is only used for team roles. Concourse-wide roles (system, admin) are different beasts and probably don't belong in here. I'm also mostly of the opinion that we don't need a catch all test since the default behaviour is to reject anything that isn't explicitly mapped here. However, if you really do want to add a test, can we map these handlers to an "admin" role, just for clarity, even though it won't ever be used in this context. 

Also do you mind targeting the `use-dex-tokens` branch with your changes.

clarafu:
@pivotal-jwinters I see, I made some changes that remove the admin required routes from the role action map and put them in a separate `AdminOnly` slice. The test will now check that all routes are covered by either the role action map or the admin only slice. But if these changes are too big and doesn't make sense to make because of the changes for #4950 then I don't mind removing them and turning this back into a simple PR for adding rerun builds to the role map.

I'd also like to target master for this because I want to get the change (adding rerun build route to the role map) into 6.0 to ship the release as soon as possible!

pivotal-jwinters:
@clarafu yeah makes sense to add the map fix to master. But can you make the test changes to a PR that targets use-dex-token?

clarafu:
@pivotal-jwinters Cool! Yup I can do that

clarafu:
So I split out the tests into #5175 that works ontop of the dex branch. Thanks for all the help everyone!

PR #5024, opened 2020-01-14
switch to pure go impl of zstd

kcmannem:
Signed-off-by: Alex Suraci <asuraci@pivotal.io>
Co-authored-by: Krishna Mannem <kmannem@pivotal.io>

Fixes #4621

# Changes proposed in this pull request

* converts datadog/zstd dependency to klauspost/compress/zstd

# Contributor Checklist
- [x] Unit tests
- [x] Integration tests (if applicable)
- [x] Updated documentation (located at https://github.com/concourse/docs)
- [ ] Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [ ] New config flags added? Ensure that they are added to the [BOSH](https://github.com/concourse/concourse-bosh-release) 
      and [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for the [integration tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env) (for example, if they are Garden configs that are not displayed in the `--help` text).

PR #5161, opened 2020-02-10
fly: fix use of zstd reader in tests

cirocosta:
### Why do we need this PR?

Some fly integration tests were not update when we switched from the
cgo-based zstd library to the new go-based one in #5024.

This made `fly`'s integration tests not compile (thus, not run lol)

Update the tests to properly leverage `zstd.NewReader`.


cirocosta:
(forced the merge as the PR tests that would run would not get to this suite)

PR #5171, opened 2020-02-12
atc: refactor lidar metrics

### LINKED ISSUE #5164: Revisit lidar metrics
### challange

The set of Lidar metrics that we currently can have its cardinality easily
growing to the 100s of thousands of timeseries on a nontrivial deployment.

For instance, with just 2 resource scopes, we get:

```
concourse_lidar_duration_seconds_bucket{scope_id="1",status="errored",le="0.001"} 0
concourse_lidar_duration_seconds_bucket{scope_id="1",status="errored",le="0.05"} 0
concourse_lidar_duration_seconds_bucket{scope_id="1",status="errored",le="0.1"} 0
concourse_lidar_duration_seconds_bucket{scope_id="1",status="errored",le="0.5"} 0
concourse_lidar_duration_seconds_bucket{scope_id="1",status="errored",le="1"} 0
concourse_lidar_duration_seconds_bucket{scope_id="1",status="errored",le="60"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="errored",le="180"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="errored",le="360"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="errored",le="720"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="errored",le="1440"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="errored",le="2880"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="errored",le="+Inf"} 6
concourse_lidar_duration_seconds_sum{scope_id="1",status="errored"} 11.033548
concourse_lidar_duration_seconds_count{scope_id="1",status="errored"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="started",le="0.001"} 0
concourse_lidar_duration_seconds_bucket{scope_id="1",status="started",le="0.05"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="started",le="0.1"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="started",le="0.5"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="started",le="1"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="started",le="60"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="started",le="180"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="started",le="360"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="started",le="720"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="started",le="1440"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="started",le="2880"} 6
concourse_lidar_duration_seconds_bucket{scope_id="1",status="started",le="+Inf"} 6
concourse_lidar_duration_seconds_sum{scope_id="1",status="started"} 0.042676
concourse_lidar_duration_seconds_count{scope_id="1",status="started"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="started",le="0.001"} 0
concourse_lidar_duration_seconds_bucket{scope_id="2",status="started",le="0.05"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="started",le="0.1"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="started",le="0.5"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="started",le="1"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="started",le="60"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="started",le="180"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="started",le="360"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="started",le="720"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="started",le="1440"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="started",le="2880"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="started",le="+Inf"} 6
concourse_lidar_duration_seconds_sum{scope_id="2",status="started"} 0.039638
concourse_lidar_duration_seconds_count{scope_id="2",status="started"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="succeeded",le="0.001"} 0
concourse_lidar_duration_seconds_bucket{scope_id="2",status="succeeded",le="0.05"} 0
concourse_lidar_duration_seconds_bucket{scope_id="2",status="succeeded",le="0.1"} 0
concourse_lidar_duration_seconds_bucket{scope_id="2",status="succeeded",le="0.5"} 0
concourse_lidar_duration_seconds_bucket{scope_id="2",status="succeeded",le="1"} 0
concourse_lidar_duration_seconds_bucket{scope_id="2",status="succeeded",le="60"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="succeeded",le="180"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="succeeded",le="360"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="succeeded",le="720"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="succeeded",le="1440"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="succeeded",le="2880"} 6
concourse_lidar_duration_seconds_bucket{scope_id="2",status="succeeded",le="+Inf"} 6
concourse_lidar_duration_seconds_sum{scope_id="2",status="succeeded"} 9.170402999999999
concourse_lidar_duration_seconds_count{scope_id="2",status="succeeded"} 6
```

which is just too much.

While using hitograms can be a contributor to it, the big problem is actually
having `scope_id` there, which can grow indefinitely. For instance, in the case
of `hush-house`, where we have ~3k resource config scopes, assuming those
`scope_id`s do not rotate at all, we'd end up with

	     buckets + total + sum
	     |
	3k * 14 * ~2 	~ 85k timeseries
	|	  |
	scopes	  started and (errored|succeeded)

### would be better

to fix that, we could drop that dimension (`scope_id`) as an operator would very
unlikely care about a specific `scope_id` (please correct me if I'm wrong).

https://github.com/concourse/concourse/blob/a2444ed7c16c0f54a26e6f2d5f653c504bbd24db/atc/metric/emitter/prometheus.go#L336-L357

it came to my mind that when dropping `scope_id`, we should also move towards
using a counter rather than a histogram as at that point, it'd be hard for an
operator to tell if something is wrong by looking at the worst time for checking
a resource across *all* checks (given that `scope_id` would be dropped).

for the high-cardinality case of "how long this specific resource is taking to
check", we could have that broken down in the form of traces: just like how
@vito instrumented the algorithm package (#5103), we could do something similar with
lidar.

refs:
- https://github.com/concourse/concourse/pull/5043: PR that added tracing
  instrumentation to `concourse`

cirocosta:
### existing issue

fixes #5164

~**This PR so far fixes only the Prometheus part of the issue mentioned above** - it might be good to think whether we'd like to go with changing the metric for the other emitters too (imo, we could go with the same approach for all of them).~

Remove `scope_id`, `check_name`, and convert the values from duration to values.

checks_total

	rather than making use of histograms (to capture duration) and having a
	dimension for `scope_id`, just keep track of how many checks we're
	performing

checks_enqueue_total

	instead of capturing the name of the check that's been enqueued, just
	count how many of them happened.

this way we don't end up exploding in terms of dimensionality, while still being
able to capture the ideas of

- "checks occurring" (and, failing/succeeding/etc)
  	- gives us utilization and error
- "checks being enqueued" (so that you can tell if there are more enqueues than
  actual checks)
	- gives us saturation

so that we end up with:

```
# HELP concourse_lidar_check_enqueue_total Counts the number of checks enqueued
# TYPE concourse_lidar_check_enqueue_total counter
concourse_lidar_check_enqueue_total 1
# HELP concourse_lidar_check_queue_size Records the size of check queue
# TYPE concourse_lidar_check_queue_size gauge
concourse_lidar_check_queue_size 1
# HELP concourse_lidar_checks_total Check time in seconds
# TYPE concourse_lidar_checks_total counter
concourse_lidar_checks_total{status="started"} 2
concourse_lidar_checks_total{status="succeeded"} 2
```

- [ ] Unit tests

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


cirocosta:
@pivotal-jwinters @xtremerui @vito how does this sound to y'all? specially w/ regards to removing the `duration`?

atc/metric/emitter/prometheus.go
```diff
@@ -324,28 +324,26 @@ func (config *PrometheusConfig) NewEmitter() (metric.Emitter, error) {
...
 
-	checksVec := prometheus.NewHistogramVec(
-		prometheus.HistogramOpts{
+	checksVec := prometheus.NewCounterVec(
+		prometheus.CounterOpts{
 			Namespace: "concourse",
 			Subsystem: "lidar",
-			Name:      "duration_seconds",
+			Name:      "checks_total",
 			Help:      "Check time in seconds",
```
vito: (not anymore)

atc/engine/engine.go
```diff
@@ -421,18 +421,12 @@ func (c *engineCheck) clearRunState() {
 
 func (c *engineCheck) trackStarted(logger lager.Logger) {
 	metric.CheckStarted{
-		CheckName:             c.check.Plan().Check.Name,
-		ResourceConfigScopeID: c.check.ResourceConfigScopeID(),
-		CheckStatus:           c.check.Status(),
-		CheckPendingDuration:  c.check.StartTime().Sub(c.check.CreateTime()),
+		CheckStatus: c.check.Status(),
```
vito: I don't think we need to include this field; it'll always be `started` which is kind of implied by the metric already.

xtremerui:
If I understand it correctly this PR removes the ability to track the duration of each check. The way it is now allows operator to set alert on check duration if one taking too long to finish, which includes scope id and name for operator to fast look up the problematic check. 

Could operator still do this with current changes?

cirocosta:
> Could operator still do this with current changes?

I don't think so :( in my mind, in a perfect world, that very specific
"per-<something>" information would come from tracing (in which case, one could
query for a span that's taking too long, or for those that errored.

*buuuut*, I recognize that not everyone would have that system in place.

however, the `checks` table does have all of the information to answer that
question in a given timeframe (check gc).

could we then assume that if we empower an operator to detect that the failure
rate of checks is going up (which we do with this PR by counting how many
errored), or that in the 99th percentile we have checks taking
1h to check (something we could have by keeping histogram-like duration
counting but without scope_id), and also give this person the ability to see the
per-check information through something like `fly checks` (which does not exist
yet), we could achieve that goal?

wdyt?

thanks!

xtremerui:
We didn't have the ability to track check duration in past with `radar`. If we identify this could be captured somewhere else I think the error rate should be good enough in metrics.

cirocosta:
>  If we identify this could be captured somewhere else I think the error rate should be good enough in metrics.

you mean, having a metric that is not per scope-id that let's you tell the overall duration of checks, of `fly checks`?

cirocosta:
> [@vito] It's tempting to use a periodically emitted counter instead of emitting metrics with a value of 1, just to make the 'push' flow a little more efficient, but I guess if it was already emitting this many before and the only problem was cardinality, maybe it's fine?

I think it's worth going for it TBH - the number of checks performed can be quiiiite big, and even if we do batching on our side (e.g., how `newrelic` and `datadog` do for `POST`ing less frequently), it's still a bunch of memory trashing going on for what could be just a few counters as you mentioneed

atc/metric/gauge.go
```diff
@@ -22,6 +22,19 @@ func (c *Gauge) Inc() {
 	}
 }
 
+func (c *Gauge) Set(val int64) {
```
cirocosta: @vito I'm not really sure if I did this right 😬 how does this sound?

cirocosta:
after the refactor, now it looks like this:

```
# HELP concourse_lidar_check_queue_size The size of the checks queue
# TYPE concourse_lidar_check_queue_size gauge
concourse_lidar_check_queue_size 1
# HELP concourse_lidar_checks_enqueued_total Total number of checks enqueued
# TYPE concourse_lidar_checks_enqueued_total counter
concourse_lidar_checks_enqueued_total 6
# HELP concourse_lidar_checks_finished_total Total number of checks finished
# TYPE concourse_lidar_checks_finished_total counter
concourse_lidar_checks_finished_total{status="error"} 0
concourse_lidar_checks_finished_total{status="success"} 12
# HELP concourse_lidar_checks_started_total Total number of checks started
# TYPE concourse_lidar_checks_started_total counter
concourse_lidar_checks_started_total 12
```

atc/metric/gauge.go
vito: I don't think this one *had* to change, since the previous metric was already emitted periodically by virtue of there only being one checker processor which itself runs periodically. So that's why this functionality didn't exist before.

I'm OK with either way though since from a pure code standpoint it's kind of nice to just have simple var definitions whose values get updated; it feels more Prometheus-style (...even though the way they currently end up getting to Prometheus is very roundabout). 👍

PR #5181, opened 2020-02-13
WIP: Introduction of Graceful Interrupt Timeout (Abort, Error, Timeout, etc)

### LINKED ISSUE #3737: Graceful Timeout Step Modifier
When using the timeout step modifier and the time limit is reached, the task is interrupted/exited without any graceful termination option.

This is problematic when using an infrastructure provisioning tool like Terraform which does not write the state back unless it gracefully exits. Therefore, we cannot do an on_abort/on_error/on_failure task to cleanup the half created resources.

The ability to annotate the current `timeout` step modifier with a similar interface to:

http://man7.org/linux/man-pages/man1/timeout.1.html

My current workaround is to use the above and not use the timeout capability built into Concourse itself.

The proposal would be to enable the follow:
```
timeout: duration
```
OR
```
timeout: { 
         duration: duration,
         signal: signal,
         signal_timeout: duration
}
```
This is a first pass thought, alternatively, potentially allowing `timeout` to be overridden by a `step` could achieve a similar behavior?

Yes, I would assist with implementation if there are no major concerns with:
1. Potentially breaking backward compatibility (would try not to)
2. Providing this functionality.

king-jam:
Opened to validate direction before diving too far into this.

* The only way I can see to get this to work is by adding a context.Value with this modifier and allowing it to pipe all the way down to the `atc/worker/client.go` code.
* I also added a worker owned context for this value to propagate in `atc/worker/interrupt.go`
* While this will likely work, I'm leaning towards this more cleanly being an extension in the gclient code. Specifically extending/changing the `Stop(kill bool) error` of the Container interface.

# Existing Issue
Fixes #3737 

# Why do we need this PR?
Per #3737 

When using the timeout step modifier and the time limit is reached, the task is interrupted/exited without any graceful termination option.

This is problematic when using an infrastructure provisioning tool like Terraform which does not write the state back unless it gracefully exits. Therefore, we cannot do an on_abort/on_error/on_failure task to cleanup the half created resources.

* Enable an annotation of `interrupt_timeout` which when present will change the process termination behavior in containers from SIGTERM -> wait 10s -> SIGKILL to SIGTERM -> wait interrupt_timeout -> SIGKILL.
* When not provided, the default behavor (SIGTERM -> wait 10s -> SIGKILL) would remain.

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/testflight/volume_mounting_test.go
+++ b/testflight/volume_mounting_test.go
@@ -36,7 +36,8 @@ var _ = Describe("A job with nested volume mounts", func() {
 		Expect(sess).To(gbytes.Say("hello"))
 	})
 
-	It("procceds through the plan with input same as output mounts", func() {
+	// Pending this test for now, @cirocosta will be looking into it
+	XIt("procceds through the plan with input same as output mounts", func() {
 		sess := fly("trigger-job", "-j", inPipeline("input-same-output"), "-w")
 		Expect(sess).To(gexec.Exit(0))
 		Expect(sess).To(gbytes.Say("hello"))


tjhiggins:
This is great @king-jam! Thank you for the quick turnaround. Will this interrupt timeout also apply if I manually abort?

king-jam:
Yes, it will be composable on ALL step types.

This needs a lot more testing which I will do after some maintainers sign
off on the approach.

PR #5191, opened 2020-02-17
atc(feat): graceful process termination when interrupted

### LINKED ISSUE #3737: Graceful Timeout Step Modifier

king-jam:
Signed-off-by: James King <james.r.king4@gmail.com>

# Existing Issue

Fixes #3737  .

# Why do we need this PR?
Per #3737

When using the timeout step modifier and the time limit is reached, the task is interrupted/exited without any graceful termination option.

This is problematic when using an infrastructure provisioning tool like Terraform which does not write the state back unless it gracefully exits. Therefore, we cannot do an on_abort/on_error/on_failure task to cleanup the half created resources.

* Enable an annotation of interrupt_timeout which when present will change the process termination behavior in containers from SIGTERM -> wait 10s -> SIGKILL to SIGTERM -> wait interrupt_timeout -> SIGKILL.
* When not provided, the default behavor (SIGTERM -> wait 10s -> SIGKILL) would remain.

# Contributor Checklist
- [X] Unit tests
- [x] Updated documentation (located at https://github.com/concourse/docs)

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/worker/client.go
+++ b/atc/worker/client.go
@@ -278,6 +278,58 @@ func (client *client) RunTaskStep(
 
 	select {
 	case <-ctx.Done():
+		// if we have an interrupt timeout, allow for a graceful termination
+		interruptTimeout, ok := InterruptTimeoutFromContext(ctx)
+		if ok {
+			// try to stop the process
+			err = process.Signal(garden.SignalTerminate)
+			if err != nil {
+				// if we can't send the signal, we need to do default behavior
+				logger.Error("process-terminate", err)
+
+				err = container.Stop(false)
+				if err != nil {
+					logger.Error("stopping-container", err)
+				}
+
+				status := <-exitStatusChan
+				return TaskResult{
+					ExitStatus:   status.processStatus,
+					VolumeMounts: container.VolumeMounts(),
+				}, ctx.Err()
+			}
+
+			// create a timer so we can clean this up on our terms
+			intTimer := time.NewTimer(interruptTimeout)
+
+			var status processStatus
+			select {
+			case status = <-exitStatusChan:
+				// process terminated with SIGTERM
+				// cleanup our timer
+				if !intTimer.Stop() {
+					<-intTimer.C
+				}
+				// break to our return flow
+				break
+			case <-intTimer.C:
+				// timed out, terminate with a SIGKILL
+				err = container.Stop(true)
+				if err != nil {
+					logger.Error("stopping-container", err)
+				}
+
+				// grab the status and break to our return flow
+				status = <-exitStatusChan
+				break
+			}
+
+			return TaskResult{
+				ExitStatus:   status.processStatus,
+				VolumeMounts: container.VolumeMounts(),
+			}, ctx.Err()
+		}
+
 		err = container.Stop(false)
 		if err != nil {
 			logger.Error("stopping-container", err)
--- a/atc/config.go
+++ b/atc/config.go
@@ -470,6 +470,9 @@ type PlanConfig struct {
 	// used on any step to interrupt the step after a given duration
 	Timeout string `json:"timeout,omitempty"`
 
+	// used on any step to force graceful terminal when interrupted
+	Interrupt string `json:"interrupt_timeout,omitempty"`
+
 	// not present in yaml
 	DependentGet string `json:"-" json:"-"`
 
--- a/atc/configvalidate/validate.go
+++ b/atc/configvalidate/validate.go
@@ -656,6 +656,14 @@ func validatePlan(c Config, identifier string, plan PlanConfig) ([]ConfigWarning
 		}
 	}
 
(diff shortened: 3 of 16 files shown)


xtreme-sameer-vohra:
Hey @king-jam 
Thank you for the submission and I'll be taking reviewing the PR shortly.

atc/worker/client.go
```diff
@@ -278,6 +278,58 @@ func (client *client) RunTaskStep(
...
+			select {
+			case status = <-exitStatusChan:
+				// process terminated with SIGTERM
+				// cleanup our timer
+				if !intTimer.Stop() {
+					<-intTimer.C
+				}
+				// break to our return flow
+				break
+			case <-intTimer.C:
```
xtreme-sameer-vohra, line 315: nit: could you simplify this using `time.After(duration)` ?

atc/worker/client.go
```diff
@@ -278,6 +278,58 @@ func (client *client) RunTaskStep(
 
 	select {
 	case <-ctx.Done():
+		// if we have an interrupt timeout, allow for a graceful termination
+		interruptTimeout, ok := InterruptTimeoutFromContext(ctx)
```
xtreme-sameer-vohra, line 282: Could you use the 
`if timeout,ok := InterruptTimeoutFromContext(ctx); ok {` 
style with an `else` to make the `else` case more explicit

xtreme-sameer-vohra:
Running acceptance testing, I wasn't able to validate the feature

Setup
- `docker-compose up --build -d`
- Set pipeline
```
jobs:
- name: test
  plan:
  - task: echo
    timeout: 31s
    interrupt_timeout: 15s
    config:
      image_resource:
        source:
          repository: busybox
        type: registry-image
      platform: linux
      run:
        path: sh
        args:
        - -c
        - |
          echo "foo"
          while true
          do
              echo "sleeps"
              sleep 5
          done
```
- Observed the build succeeds in the UI but expected it to fail. I didn't run a test to validate whether SIGINT was sent and the interrupt timeout was respected as this simpler case wasn't working.
- Also the UI isn't able to render the steps, its likely it can't parse the new format
- Using `fly`, the only output is the status which says `succeeded`
- Few more changes might be required to wire everything up to work (eg. atc/engine/builder/builder.go needs to be updated to include the new step )

@king-jam did this work for you ?

king-jam:
Hi @clarafu and @xtreme-sameer-vohra . This definitely requires a lot more work. I started it and sent it mainly to see if the `client.go` change would be acceptable. I knew the 6.0 release was incoming at the same time so stopped iterating pending major changes that I would have to go back and resolve.

I will start working on feedback and resolving conflicts.

clarafu:
@king-jam That makes total sense, v6.0 did come with a lot of moving around of code. I think the only PR I would be currently weary about conflicting with is https://github.com/concourse/concourse/pull/5504. #5504 changes how a build plan is created, so the only thing that will conflict with your PR is where we would need to add the new field `Interrupt` to the build plan. But I wouldn't worry about it too much since it'll probably be an easy edit, just wanted to let you know about it beforehand!

gowrisankar22:
Graceful termination for build interrupt and abort much useful feature for us. We are writing build abort / interrupt status to our monitoring stack when we interrupt or abort by trapping the SIGINT Signal. Our approach is not working all the time fine because sometimes due to non-graceful timeout. definitely we are gonna get 💯 % benefit out of it ..

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

clarafu:
@king-jam Hi! I'd like to know if you are still interested in continuing to work on this PR? If you don't think you will have time or are no longer interested in it, I'd like to close it for now and you can open a new PR in the future if needed. Thanks!

king-jam:
Yes, we can close for now.

PR #5192, opened 2020-02-17
feat(atc): retry build upon worker missing.

### LINKED ISSUE #5086: Retry steps upon worker disappearing

In any cloud infra, VM crashing is something "normal". As most of Concourse workers run in cloud, Concourse should be able to handle worker disappearing.

Currently, when a worker disappears, Concourse just fails the running steps, and log an error on the Web UI, which makes production pipelines unreliable.

In our production clusters, this is becoming a very serious issue.

Concourse should be able to handle errors of worker disappearing, and automatically rerun failed steps on other workers.

I also realize blindly rerunning some `put` steps may lead to problem, so we may add a flag to `put` step to allow user to turn off auto rerun.

I can try.

evanchaoli:
# Existing Issue

Fixes #5086 .

# Why do we need this PR?

This will reduce build failure rate on cloud infra where workers could die frequently.

The first commit is only call for design review. The change is actually very simple.

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


muntac:
Hi @evanchaoli ! Thanks for brining up and working on this issue. I'm an engineer on the Concourse team. I can help you with feedback on this if you'd like. Could you outline the approach you were thinking of?

Also, since this is a work in progress I'm going to convert the PR to a draft for now.

evanchaoli:
@muntac I think this is ready for review.

The approach is to let the build not fail if failed with worker missing error, so that ATC will auto retry to run the build again.

atc/engine/builder/step_factory.go
```diff
@@ -69,7 +69,7 @@ func (factory *stepFactory) GetStep(
 		factory.client,
 	)
 
-	return exec.LogError(getStep, delegate)
+	return exec.RetryError(getStep, delegate)
```
muntac, line 72: On https://github.com/concourse/concourse/issues/5086#issuecomment-606018381 it was recommended to not have the retry functionality turned on by default. That would mean having some sort of check here as well as allowing users to specify a flag to turn this feature on.

atc/exec/retry_error_step.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func RetryError(step Step, delegate LogErrorStepDelegate) Step {
+	return RetryErrorStep{
+		Step:     LogError(step, delegate),
+		delegate: delegate,
+	}
+}
+
+func (step RetryErrorStep) Run(ctx context.Context, state RunState) error {
```
muntac, line 31: Is there a way to unit test this functionality? You can look at any of the other step files for test examples. Here's one:
https://github.com/concourse/concourse/blob/master/atc/exec/ensure_step_test.go

atc/exec/retry_error_step.go
```diff
@@ -0,0 +1,53 @@
...
+	logger := lagerctx.FromContext(ctx)
+
+	runErr := step.Step.Run(ctx, state)
```
muntac, line 34: This is fine but took me a little bit of time to grasp, and I'm wondering if others who are unfamiliar with this part of the codebase will feel the same. 

Maybe the indirection introduced by having `RetryError` wrap `LogError` (which was itself wrapping a `Step`) is not necessary.  One alternative could be to move the essential logic of `RetryError` - conditionally setting error type to "Retriable" and returning that error - to `LogError` instead.

Another alternative is to check if the error message has "worker  .... disappeared" in the section of `engine.go` where you're checking if the error is of type `Retriable`.

I can maybe guess some reasons why those alternatives aren't as good and would like to know your thoughts.

atc/exec/retry_error_step.go
```diff
@@ -0,0 +1,53 @@
...
+		logger.Info("retriable", lager.Data{"error": runErr.Error()})
+		step.delegate.Errored(logger, "errored, will retry ...")
+		runErr = Retriable{}
+	}
+
+	return runErr
+}
+
+func (step RetryErrorStep) toRetry(logger lager.Logger, err error) bool {
+	re := regexp.MustCompile(`worker .+ disappeared`)
```
aledeganopix4d, line 45: Functionally speaking I agree this is the right thing to do.
I wonder if there is a more robust/abstract way to identify when a worker disappeared other than parsing the error log...

aledeganopix4d:
I really like the idea of this PR and it's 100% the right direction to make Concourse more resilient.
I just want to share our use case at Pix4D and why this PR would improve our operation: we run the workers on AWS Autoscaling Groups, sometimes it happens that we scale-in too soon and the worker disappear, often when doing something trivial like a GitHub status update after a long build.
Thus having this retry mechanism only on the failed step would improve significantly the efficiency of our system as we wouldn't have to rerun the entire job.

Thanks!

atc/engine/engine.go
```diff
@@ -234,6 +234,11 @@ func (b *engineBuild) Run(logger lager.Logger) {
 
 	case err = <-done:
 		logger.Debug("engine-build-done")
+		if err != nil {
+			if _, ok := err.(exec.Retriable); ok {
+				return
```
vito, line 239: This is pretty interesting. So the idea is to just bail on the build plan for now and allow the build tracker to pick it back up?

Don't really have any preference either way at the moment - it should be fine, and reinforces the notion that builds are idempotent and can be re-attached. It might make it difficult to implement things like "retry up to N times" though. 🤔

pivotal-jwinters: I like it! 

If we wanted to do "retry up to N times" or some type of backoff, it might make sense to do that in the build tracker anyway?

atc/worker/client.go
```diff
@@ -268,6 +268,9 @@ func (client *client) RunTaskStep(
 		return TaskResult{}, err
 	}
 
+	// TODO: delete before merge.
```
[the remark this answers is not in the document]
evanchaoli: Sorry for making you confused. It's not that I forgot to remove this comment. The commented line below is very important for my debugging. I left the code commented and put a TODO there, once all review comments are addressed, I'll remove them before merge.

atc/exec/task_step.go
```diff
@@ -1,15 +1,10 @@
...
-	"context"
-	"fmt"
-	"io"
-	"path"
-	"path/filepath"
-	"strings"
-
 	"code.cloudfoundry.org/lager"
 	"code.cloudfoundry.org/lager/lagerctx"
+	"context"
```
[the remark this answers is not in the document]
evanchaoli: I didn't manually modify that, it's done by `go format`. But, yes, I can exclude this change from this PR if my change don't touch this file.

atc/exec/retry_error_step.go
evanchaoli: I'll add unit test for this. Practically, for this kind of new feature changes, I usually called for an initial review first from design perspective, if you folks agree with the design, I added tests in the second commit. Because adding tests is kinda costly, if we added folks rejected the design, the time spent on writing unit test is wasted. I have done many PRs, and always did the way.

evanchaoli:
> Thus having this retry mechanism only on the failed step would improve significantly the efficiency of our system as we wouldn't have to rerun the entire job.

@aledeganopix4d Yes, that was in my plan, but the first commit didn't include that, because I wanted the Concourse dev team to take an initial review from the design perspective. My second commit has included that. But it leads to a lot of more complexities. Because when a worker disappeared, not only the current failed step, but some previous steps, like gets, and tasks having outputs ran on the worker, also need to rerun, because volumes disappeared together with the worker.

atc/engine/engine.go
evanchaoli: @vito @pivotal-jwinters please take a look at my second commit. In the second commit, I implemented a logic of only rerun necessary steps. With that, I just don't return to tracker at all, once a build fails, the engine retries it immediately, so that it holds the original `run_state`, from `run_state` it knows previous volumes.

evanchaoli:
In the second commit, I implemented a logic of only rerun necessary steps. To do that:

- Save worker name in build_event's payload. To decide if a step needs to rerun, we have to know on which worker it ran previously. If the worker is right the disappeared worker,  and the step generated some output volume, then the step has to be rerun, because the volume disappeared together with the worker.
- Add a method `HasDoneSuccessfully` in delegate to decide if the step has run successfully.

See my test result:
auto-rerun

In this test, I had two workers: `384dc5b3b263` and `8728505c556c`. At the last step `sleep-task`, I killed worker `8728505c556c`. Then we can see, only the get after put was rerun, because it ran on the killed worker, and of course, the failed step `sleep-task` was rerun.

NOTE: 
- Pls ignore unit test for now. I'll add unit tests once there is no more comment. for the design. 
- I should also add `HasDoneSuccessfull()` to `buidStepDelegate` and steps `set_pipeline` and `load_var`. Will add in next commit.

evanchaoli:
> @evanchaoli I think your second commit is moving in the wrong direction. Every step in concourse should already be re-entrant so your changes aren't necessary.

@pivotal-jwinters I'm not sure what you meant "every step in concourse should already be re-entrant", let me clarify with a sample.

Say, a job contains: steps-a/b/c/d in serial. Say a build, a/b have passed, and during running c:

- If the ATC crashes, then the other ATC will take over, and entire build (a/b/c/d) will rerun; 
- If the worker where c is running on crashes
  - Without this PR, the build will fail
  - With the first commit of this PR, the other workers will take over, and entire build (a/b/c/d) will rerun.
  - With the second commit of this PR, a/b will not rerun, only rerun c, then run d.

This sounds like build is re-entrant rather than step. A build holds a `run_state` in memory, and `run_state` holds a repo of artifacts (gets and outputs of tasks), which decides that steps cannot be re-entrant.

But yeah, having said that in the previous comment, I have realized that my second commit broke build re-entrant. With the second commit, if the ATC crashes, the build will fail. I don't need to store "worker name" to database at all. I'll fix that in next commit.

@pivotal-jwinters 
1. Does my above understanding correct?
2. Do you agree we should avoid to rerun a and b when the worker crashes when runs c?

atc/db/build.go
```diff
@@ -149,6 +149,8 @@ type Build interface {
 
 	IsDrained() bool
 	SetDrained(bool) error
+
+	FindWorker(string) (Worker, bool, error)
```
clarafu, line 153: I'm not sure if this method should belong on the `Build` interface. It seems to be pretty unrelated to a build and is only using the build's connection to find a worker that matches the name passed in as an argument. Do you think there a better place that this method could live on?

pivotal-jwinters:
@evanchaoli 

> If the ATC crashes, then the other ATC will take over, and entire build (a/b/c/d) will rerun;

Not exactly. The new ATC will start the build plan from the beginning, however each step that has already run will no-op because a container already exists for that step. The new atc will rebuild the `run_state` however the run state is typically just a reference to a volume on a worker (it doesn't hold any data itself) so rebuilding the `run_state` is cheap.

> Do you agree we should avoid to rerun a and b when the worker crashes when runs c?

This is a loaded question lol. Yes I do agree that this should be the result. However, I think there's another way we can achieve this result. There's a distinction between errored and failed builds. I think what we probably want is for concourse to retry errored builds (up to N times) before actually marking the build as errored. This covers the case of the worker missing, worker crashing, etc. When we retry a build plan for an errored build, all the previously successful steps will no-op as mentioned above, and all the errored steps will retry.

evanchaoli:
> Not exactly. The new ATC will start the build plan from the beginning, however each step that has already run will no-op because a container already exists for that step. The new atc will rebuild the `run_state` however the run state is typically just a reference to a volume on a worker (it doesn't hold any data itself) so rebuilding the `run_state` is cheap.

@pivotal-jwinters Good to know that. And I just did a test that proved you are right. However I still think my change is useful, especially for `put`, `load_var`, and `set_pipeline`. See below screen shot from my test:

discuss-step-rerun

In this test, I used 4 workers. And after `put tiny2`, I killed worker `5f67efc18be0` where the `put` ran on. Then we can see:

- The `put` chose a new worker and rerun
- The implied `get` chose the same worker and didn't rerun, as you explained.

But I expect `put` to NOT rerun either, which is a big problem. If the `put` is a `git push`, duplicate push would fail; if the `put` is a database write, it may leads to data conflict or duplicate data, and so on ...

For `load_var` and `set_pipeline`, they don't use container at all, meaning they will always rerun. As long as set_pipeline has done, even if the input volume is gone, no need to set the pipeline again, though cost of set-pipeline again is not high.

So in summary, IMO, making ATC to start over wouldn't completely avoid unnecessary step reruns, especially `put` rerun may cause unexpected build failure. WDYT?

atc/db/build.go
evanchaoli: @clarafu Thanks for your comment. I also feel not good to put `FindWorker` in `build`. But the problem is, this function needs to be called in step delegates, and a step delegate has only a `build` object that contains db conn, so I cannot find other place to add this function. Do you have any suggestion?

evanchaoli: No worries, I think this function is no longer needed, I'll delete it in next commit.

pivotal-jwinters:
@evanchaoli 

Yeah if a container goes away, concourse doesn't know its already run that step. @vito has talked about adding a table to track step status, which seems like it could be useful for what you're trying to achieve here. Querying build events just seems very fragile and quite expensive so I'm really hesitant in the current approach. 

Also, I'm really hoping that every step doesn't need to check if it's already ran. Maybe this is something we could move inside your the RetryStep?

vito:
> But I expect `put` to NOT rerun either, which is a big problem. If the `put` is a `git push`, duplicate push would fail; if the `put` is a database write, it may leads to data conflict or duplicate data, and so on ...

In principle, this should not be the case - all `put` steps *must* be idempotent, otherwise re-running builds (via clicking the re-run button in the UI) would fail if the build had previously run `put` steps. A `git push` for example is a no-op if the data is already there, and any `put` that does database operations *must* be idempotent in order to fit the resource design.

I would be completely comfortable doubling down on this, as it's always been a design intention.

evanchaoli:
> In principle, this should not be the case - all `put` steps _must_ be idempotent, otherwise re-running builds (via clicking the re-run button in the UI) would fail if the build had previously run `put` steps. A `git push` for example is a no-op if the data is already there, and any `put` that does database operations _must_ be idempotent in order to fit the resource design.

The principle, as well as the git "no-op" example you mentioned, fits to ATC crash handler. When ATC crashes, all workers are still there, by finding existing containers, `get`/`task`/`put` may not really rerun container works. But this PR is trying to handle worker crash, the principle might not work. For example, a job does the following:

```
- get: mygit
- task: generate a new commit
- put: mygit
- other steps ...
```
For a build, say, the task and put ran on same worker, and the worker crashed. Task will rerun, thus generate a new commit, though commit content is same as last run, but commit it is different, then put will fail, because previous push with different commit id has done.

So, to handle worker crash problem, Concourse needs to introduce some new principle. What I'm coding in this PR might not be the best solution/design. I'm open to accept your advice, but my goal is try to avoid unnecessary rerun. Like a lot of pipelines run tests in tasks, and a task may last hours. We don't want the test to rerun for hours just because a following step fails due to a worker crash. 

For `set_pipeline` may not hurt, but might make user confused, because the first run will show diffs, the second run will show no-diff, because the pipeline has been set.

My implementation handles worker crash at the engine layer, not raising up to the tracker level. So that `run_state` is there. If you consider query `build_event` table from db is fragile, maybe we can store step result in `run_state`.

> otherwise re-running builds (via clicking the re-run button in the UI) would fail if the build had previously run put steps.

With my above example, I doubt manual rerun would fail also, because task will generate a new commit. I'm on vacation today, so I won't make a pipeline to verify that today, but I'll do once I back to worker.

vito:
@evanchaoli Hmm, you bring up good points but I still think it would be OK to just re-run them. There may be examples where this would cause problems today but I think those problems should be addressed, rather than used to dissuade us from a core design goal.

In your example, the `put` step could be configured with `rebase: true` or `merge: true` which should result in a no-op when pushing the exact same commit. Maybe one of these should be the default, or a required parameter.

In at least some scenarios we would have to re-run the `task` step anyway. For example:

```yaml
plan:
- task: task-1
- put: put-1
- task: task-2
- put: put-2
```

Let's say...:

* there's only one worker, `worker-1`.
* `task-1` produces an output which is consumed by all subsequent steps.
* `task-1` and `put-1` run and succeed.
* `worker-1` crashes while running `task-2`.
* `worker-2` is provisioned, and the build can now re-run automatically.

One could argue that the `put: put-1` could be skipped when re-running, because it has already consumed its inputs and has yielded no new outputs. However in order for `put-2` to run, it will need the outputs produced by `task-1`. Given that the worker it ran on is gone, we would have to re-run it.

Now that we have to re-run it, it's possible that it generates a new commit. This means we may *also* have to re-run `put-1`. Maybe it's a repository mirror or something. Hard to think of a `git` specific example here, but I don't think it's a stretch to say this requirement would exist somewhere.

So, we might be able to address this if we're careful to track the complete flow of inputs/outputs through a build plan to identify whether a later step will need the step to re-run. But to me that feels like a lot of complexity. IMO a long-running task running again because the worker crashed is fine. Workers shouldn't crash often enough that it warrants building complexity that avoids core design principles.

The `set_pipeline` step doubling output is also, IMO, fine. Workers crashing is an unusual event. I think it's OK for things to look a little "unusual" as long as it makes sense once the user knows what happened (so we'd need some UI indicator).

> With my above example, I doubt manual rerun would fail also, because task will generate a new commit. I'm on vacation today, so I won't make a pipeline to verify that today, but I'll do once I back to worker.

Re-running the build will have all inputs pinned to the same versions they were before, so a second commit will be made against the same prior commit, so that should result in the same issue. (This is another reason we usually use `rebase: true` or `merge: true`.)

evanchaoli:
@vito Okey, let's just rerun all steps for now. But I can imagine that I'll be facing users challenge once this feature is in place.

Can you pls confirm your comment https://github.com/concourse/concourse/pull/5192#discussion_r413916938, how should I handle the comment?

vito:
@evanchaoli Cool! I don't think there are any problems to address from that comment. Exiting early from the build and retrying again just seemed like a simple and kind of novel approach which I've never tried but it might just work. 🙂

I mentioned "retry N times" as being a potential difficulty with that approach but that's not necessarily true; if we really want that we could increment a retry count on the build or something and have the build tracker notice when it's at the limit.

atc/engine/engine.go
evanchaoli: Per https://github.com/concourse/concourse/pull/5192#issuecomment-624827620, nothing to do with this comment, thus resolving.

evanchaoli: Per @vito, nothing to address for this comment, thus closing.

evanchaoli:
@muntac @pivotal-jwinters I have addressed all the comments, make the feature opt-in, removed debugging code, it's ready for merge. Please review it again. Thanks.

atc/exec/put_step.go
```diff
@@ -59,7 +59,7 @@ func NewPutStep(
 	strategy worker.ContainerPlacementStrategy,
 	workerClient worker.Client,
 	delegate PutDelegate,
-) *PutStep {
+) Step {
```
muntac, line 62: I'm assuming this change was unintentional from a replace command of some sort.

atc/exec/put_step_test.go
```diff
@@ -61,7 +61,7 @@ var _ = Describe("PutStep", func() {
 		repo  *build.Repository
 		state *execfakes.FakeRunState
 
-		putStep *exec.PutStep
+		putStep exec.Step
```
muntac, line 64: Assuming this was also unintentional like the other PutStep change.

atc/atccmd/command.go
```diff
@@ -239,6 +239,8 @@ type RunCommand struct {
 	SystemClaimKey        string   `long:"system-claim-key" default:"aud" description:"The token claim key to use when matching system-claim-values"`
 	SystemClaimValues     []string `long:"system-claim-value" default:"concourse-worker" description:"Configure which token requests should be considered 'system' requests."`
 	EnableArchivePipeline bool     `long:"enable-archive-pipeline" description:"Enable /api/v1/teams/{team}/pipelines/{pipeline}/archive endpoint."`
+
+	EnableBuildRerunWhenWorkerDisappear bool `long:"enable-rerun-when-worker-disappear" description:"Enable automatically build rerun when worker disappears"`
```
muntac, line 243: Small critique - I would add an "s" at the end `EnableBuildRerunWhenWorkerDisappears`

atc/engine/builder/step_factory.go
```diff
@@ -14,15 +14,16 @@ import (
...
+	pool                           worker.Pool
+	client                         worker.Client
+	resourceFactory                resource.ResourceFactory
+	teamFactory                    db.TeamFactory
+	resourceCacheFactory           db.ResourceCacheFactory
+	resourceConfigFactory          db.ResourceConfigFactory
+	defaultLimits                  atc.ContainerLimits
+	strategy                       worker.ContainerPlacementStrategy
+	lockFactory                    lock.LockFactory
+	enableRerunWhenWorkerDisappear bool
```
muntac, line 26: Small critique - again would add an "s" at the end `enableRerunWhenWorkerDisappears`

atc/exec/put_step.go
evanchaoli: No, this is a intentional change. If you look at other steps, like https://github.com/concourse/concourse/blob/master/atc/exec/get_step.go#L80, they all return `Step`, thus this is fixing a previous error. But this is also required to make the change `step_factory.go`, because, as the code:

```go
	putStep := exec.NewPutStep(
		plan.ID,
		*plan.Put,
		stepMetadata,
		containerMetadata,
		factory.resourceFactory,
		factory.strategy,
		factory.client,
		delegate,
	)

	putStep = exec.LogError(putStep, delegate)
	if factory.enableRerunWhenWorkerDisappear {
		putStep = exec.RetryError(putStep, delegate)
	}
```

If `NewPutStep` returns `*PutStep`, then `putStep` is type of `*PutStep`, then we cannot step return value of `exec.LogError(putStep, delegate)` to `putStep`, because of `LogError` returns a `Step` typed object.

atc/exec/put_step_test.go
evanchaoli: Same.

atc/atccmd/command.go
evanchaoli: fixed.

atc/engine/builder/step_factory.go
evanchaoli: fixed

atc/exec/put_step.go
muntac: Ah I see. Good catch!

evanchaoli:
@muntac As all review comments have been addressed, would approve and merge?

PR #5210, opened 2020-02-20
fly: Better error messages for --team

### LINKED ISSUE #5126: fly: passing a nonexistent `--team` gives a misleading error
## Bug Report

When I run `fly hijack`, `fly jobs`, `fly trigger-job`, `fly pause-job` or `fly unpause-job`  with the `--team` flag set to a nonexistent team, the error suggests that something is not found, but doesn't always make clear that the team in question doesn't exist.

## Steps to Reproduce

```bash
$ fly -t ci pause-job -j concourse/unit --team doesntexist
```

## Expected Results

something like 'no such team'.

## Actual Results

```
error: concourse/unit not found on team doesntexist
```

## Additional Context

the `Team` method on the `Team` interface in the `concourse` package under `go-concourse` is quite confusing.

## Version Info

* Concourse version: 6.0.0-pre (master)
* Deployment type (BOSH/Docker/binary): n/a
* Infrastructure/IaaS: n/a
* Did this used to work? No

## TODO

- [x] Trigger-job
- [x] hijack
- [x] job
- [x] pause-job
- [x] unpause-job

taylorsilva:
# Existing Issue

Fixes #5126 

# Why do we need this PR?
`--team` does not return any helpful error messages for when:

- User does not belong to a team
- The team does not exist

This PR adds error messages for both of those cases.

* Added an `httpAgent` which will replace `connection` in the `go-concourse` client library. Expanding `connection` was painful because it does too many things (send HTTP requests AND parse the response). `httpAgent` does one thing: send HTTP requests.
* Removed the mock HTTP server `adminATCServer`. It was not needed and was added in the context of Super Admin stories done in September.
* Refactored the existing `--team` tests so they make sense.
* With all the above refactoring we were able to add better error messages for when `--team` fails.
  * User is not on a team user gets `you do not have a role on team 'team-name'`
  * Team does not exist user gets `team 'team-name' does not exist`

# Contributor Checklist

- [x] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

cc @muntac @pivotal-izabela-gomes

### DIFF
--- a/fly/integration/login_insecure_test.go
+++ b/fly/integration/login_insecure_test.go
@@ -18,7 +18,7 @@ import (
 )
 
 var _ = Describe("login -k Command", func() {
-	var loginATCServer *ghttp.Server
+	var adminAtcServer *ghttp.Server
 
 	Describe("login", func() {
 		var (
@@ -27,23 +27,23 @@ var _ = Describe("login -k Command", func() {
 		)
 		BeforeEach(func() {
 			l := log.New(GinkgoWriter, "TLSServer", 0)
-			loginATCServer = ghttp.NewUnstartedServer()
-			loginATCServer.HTTPTestServer.Config.ErrorLog = l
-			loginATCServer.HTTPTestServer.StartTLS()
+			adminAtcServer = ghttp.NewUnstartedServer()
+			adminAtcServer.HTTPTestServer.Config.ErrorLog = l
+			adminAtcServer.HTTPTestServer.StartTLS()
 		})
 
 		AfterEach(func() {
-			loginATCServer.Close()
+			adminAtcServer.Close()
 		})
 
 		Context("to new target with invalid SSL with -k", func() {
 			BeforeEach(func() {
-				loginATCServer.AppendHandlers(
+				adminAtcServer.AppendHandlers(
 					infoHandler(),
 					tokenHandler(),
 				)
 
-				flyCmd = exec.Command(flyPath, "-t", "some-target", "login", "-c", loginATCServer.URL(), "-k", "-u", "some_user", "-p", "some_pass")
+				flyCmd = exec.Command(flyPath, "-t", "some-target", "login", "-c", adminAtcServer.URL(), "-k", "-u", "some_user", "-p", "some_pass")
 
 				var err error
 				stdin, err = flyCmd.StdinPipe()
@@ -68,7 +68,7 @@ var _ = Describe("login -k Command", func() {
 			Context("login to existing target", func() {
 				var otherCmd *exec.Cmd
 				BeforeEach(func() {
-					loginATCServer.AppendHandlers(
+					adminAtcServer.AppendHandlers(
 						infoHandler(),
 						tokenHandler(),
 					)
@@ -118,7 +118,7 @@ var _ = Describe("login -k Command", func() {
 		Context("to new target with invalid SSL without -k", func() {
 			Context("without --ca-cert", func() {
 				BeforeEach(func() {
-					flyCmd = exec.Command(flyPath, "-t", "some-target", "login", "-c", loginATCServer.URL(), "-u", "some_user", "-p", "some_pass")
+					flyCmd = exec.Command(flyPath, "-t", "some-target", "login", "-c", adminAtcServer.URL(), "-u", "some_user", "-p", "some_pass")
 
 					var err error
 					stdin, err = flyCmd.StdinPipe()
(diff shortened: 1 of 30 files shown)


fly/integration/login_insecure_test.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 var _ = Describe("login -k Command", func() {
-	var loginATCServer *ghttp.Server
+	var loginAtcServer *ghttp.Server
```
pivotal-jwinters: Small nit can we change this back to loginATCServer (capital ATC) to minimize the diff

fly/integration/login_test.go
```diff
@@ -24,7 +24,7 @@ import (
 
 var _ = Describe("login Command", func() {
 	var (
-		loginATCServer *ghttp.Server
+		loginAtcServer *ghttp.Server
```
pivotal-jwinters: same here

go-concourse/concourse/client.go
```diff
@@ -52,3 +59,42 @@ func (client *client) URL() string {
 func (client *client) HTTPClient() *http.Client {
 	return client.connection.HTTPClient()
 }
+
+func (client *client) FindTeam(teamName string) (Team, error) {
```
pivotal-jwinters: Should we move this into the team.go or teams.go file?

go-concourse/concourse/internal/connection_test.go
```diff
@@ -355,7 +357,7 @@ var _ = Describe("ATC Connection", func() {
 				BeforeEach(func() {
 					atcServer = ghttp.NewServer()
 
-					connection = NewConnection(atcServer.URL(), nil, tracing)
+					agent = NewHTTPAgent(atcServer.URL(), nil, tracing)
```
pivotal-jwinters: How come we're testing the http_agent in the connection_test file? Should we have a separate suite for http_agent_test? 

Overall a little confused why we have both the connection and agent? Should we just refactor the connection instead of deprecating it?

go-concourse/concourse/client.go
taylorsilva: Good point, forgot to move it out. In teams.go it goes!

go-concourse/concourse/internal/connection_test.go
taylorsilva: We want to use `HTTPAgent` everywhere but the diff on this refactor is already really large and this would've dramatically increased the scope. We're holding ourselves back and only making this one change while indicating where we want to see this change end up, with `connection` deleted and replaced by `HTTPAgent`

pivotal-jwinters: It would probably keep the diff small if we refactored `connection` to look like `agent` and then continued using `connection` everywhere.

taylorsilva: I disagree that the diff would be smaller  and we would still end up with a larger refactor (maybe I'm misunderstanding what you're suggesting?). This is how I'm interperting your suggestion and why I think it would result in a larger diff.

`connection` does two things: 1) send HTTP requests and 2) parse the response. 
`agent` does one thing 1) send HTTP requests. 
The response parsing still needs to happen though, who's responsible for that now? The caller has to do this now (and was already doing part of this but in a non-extensible way). This extra responsibility is what allowed us to return better error messages when `--team` is used.

For the one case (`FindTeam`) currently using `agent` you can see the response parsing happening here: 

https://github.com/concourse/concourse/blob/587483424a47f3eebb2ed8e861687b60cccb9ec7/go-concourse/concourse/teams.go#L122-L145

If we were to to switch everything over we'd have to change this section:
https://github.com/concourse/concourse/blob/587483424a47f3eebb2ed8e861687b60cccb9ec7/go-concourse/concourse/teams.go#L84-L97

 in every area that uses `connection` which is currently 71 places, every API call:
```
ag 'connection.Send' ./go-concourse/ | wc -l
      71
```

I think this will result in a much larger diff and will take a good chunk of time to refactor. I think this can be it's own story, finishing the refactor that was started in this story. wdyt?

pivotal-jwinters: yeah ok, I didn't realize that the response handling would be that different.

taylorsilva: Made an issue to complete this refactor: #5214

Issue #5209, opened 2020-02-20
experiment: kubernetes-based implementation of atc's worker interface


cirocosta:
### challenge

Without having [`kubernetes`] as a native scheduling target, the execution of
[`steps`] in that environment become opaque to Kubernetes, which:

- i. makes impossible for operators to have first-class support for dealing with
  units of work that Concourse creates (containers)

- ii. forces those operators to grant much broader permissions than they'd want
  to give to Concourse just for the sake of allowing workers to be able to do
  nested containerization.

[`kubernetes`]: https://kubernetes.io/
[`steps`]: https://concourse-ci.org/steps.html

### goal

Being able to target Kubernetes with native workloads, without forcing Concourse
users to change their pipelines / task configs (i.e., not breaking "userspace").

### current exploration

- #5223: "kubernetes as a worker" from ATC's perspective *(includes a demo 😁 )*
- https://github.com/concourse/baggageclaim/pull/36: extra capabilities added to `baggageclaim` to make it serve its volumes as images (makes baggageclaim a container image registry)
- https://github.com/concourse/concourse-chart/pull/77: adds to the chart the ability to let `web` have a serviceaccount that can manage a namespace's pods and other resources necessary

[oci distribution spec]: https://github.com/opencontainers/distribution-spec
[refactor exec/*_step.go to remove exposure to runtime/garden componentes]: https://github.com/concourse/concourse/issues/4265
[refactor check step]: https://github.com/concourse/concourse/issues/4957
[implement a kubernetes runtime for each type of step]: https://github.com/concourse/concourse/issues/4391

### context

the folks in the runtime domain have been working on making possible to have a
clear separation between the concerns of what "runtime" (materializing
"workloads"), and "core" (deciding when things should happen) of concourse are
responsible for.

	"concourse"
		|
		'------(workload)-------> worker

		
			VS

	"core"
		schedules work to be done

	"runtime"
		takes the work, makes it run, somewhere
			|
			'-----------------> worker

most of that started with the work of getting that separation regardless of
where those workloads would be scheduled to run on top of (see [RFC: runtime
interface]), while having in mind that we'd like to see how we could better
integrate with kubernetes ([RFC: how do we best leverage k8s as a runtime]).

in between that time and today, the team has also worked on making *possible*
for users to have Concourse running on Kubernetes, even if not in the best way
possible, by having a [helm] chart (see [concourse-chart]) that'd run the
current form of Concourse, running pods as workers, having Concourse scheduling
[...]

Issue #4957, opened 2020-01-02
Refactor Check Step


kcmannem:
Continuing the work from #4265. We need to slowly start converting the check step to also follow the new boundary we've placed between the exec/runtime/worker package. In its simplest form, it should be as simple as moving function calls around worker selection and container creation into a wrapper `client.RunCheckStep` on the worker.Client interface.
This won't be all as I'm sure we'll find more coupling as we explore the issue.

PR #5223, opened 2020-02-24
experiment: kubernetes worker

cirocosta:
### why do we need this pr?

This PR demonstrates that the approach described in #5209 is feasible.

It's **NOT** intended to ever be integrated *as is* - this is throwaway code
just for demonstration purposes to validate the design of #5209.

fixes https://github.com/concourse/concourse/issues/5209 (for the sake of github links lol)

### goal

The goal of this exploration is to be able to consume *any* pipeline composed of
jobs that are linux-based, without any changes needed to the pipeline and task
definitions.

### demo

Peek 2020-02-24 09-11

the demo showcases a task that's launched based of a container image
that's fetched by a `get` step (in `concourse`'s own `rootfs` format), listing the
input that it received from another `get` step that fetched a repository.

### trying it out

There are two ways of trying this out:

#### local build

You can do so using the `Makefile` included in here though:

```bash
# clone this branch to `concourse` and get inside it.
#
git clone \
	--branch k8s-iter-1 \
	https://github.com/concourse/concourse \
	concourse
cd $_


# start a kubernetes cluster using `kind` with a modified CRI (because ...
# containerd) that whitelists `http`-based registries.
#
make cluster


# run postgres in docker
#
make db


# build the ui
#
yarn install && yarn build


# build, install, and run Concourse (make sure you run `yarn build` 
# first to build the UI assets)
#
make run
```

as a result, you should have:

1. `postgres` running in docker
2. `kubernetes` running in docker
2. `web` running locally (with ATC available on `:8080`)

That done, you can now access Concourse on http://localhost:8080.

#### using the helm chart

I've uploaded an image (`cirocosta/concourse:k8s-iter-1`) to DockerHub for those
wanting to try out the "in-cluster" experience, and repurposed "the Helm chart of
today" to support the "runtime of tomorrow".

```bash
# clone the branch of the chart where it includes the necessary RBAC configs for
# setting up a service account that's powerful enough
#
git clone \
        --branch k8s-iter-1 \
        https://github.com/concourse/concourse-chart


# clone the branch of this PR
#
git clone \
        --branch k8s-iter-1 \
        https://github.com/concourse/concourse


# get the `init` configmap into the cluster so that we can hold containers alive
# until processes are meant to be executed in them.
#
NAMESPACE=default make -C ./concourse init


# install the chart / render the templates with the sample values under
# `hack/k8s/values.yaml` from this PR's branch
#
helm dependency update ./concourse-chart
helm upgrade \
        --install \
        --values=./concourse/hack/k8s/values.yaml \
        test \
        ./concourse-chart
```

with the chart installed, you can now port-forward ATC (http://localhost:8080)
and get going.

ps.: your cluster *must* have support for pulling images from http-based
registries served as pods. you can have a local cluster running `make -C
concourse cluster` if you'd like to use `kind`).

### how it works

The execution of each step follows the same pattern as we're accustomed with
when using Garden, where we set the sandbox up, and then run the desired process
inside it ( `db.Creating()` -> `container.Create()` -> `db.Created()`... ->
`container.Run()`)

Because a build plan can be seen as a directed acyclic graph when it comes to
dependencies, we can rely on that fact to dictate how each step either gathers
inputs, our supplies outputs.

For each step, as long as its dependencies (including transitive ones) were able
to fulfill what they should (e.g., a `get` successfully running `/opt/resource/in`), 
it'll be able to retrieve any artifact that it might need.

When running a step, `atc` communicates with Kubernetes to create pods that
represent those steps. 

Each step pod can be seen as a potential permutation of two configurations:

- i. having an "inputs fetcher" init container, responsible for fetching
  artifacts retrieves by dependnecies
- ii.  having an "output streamer" sidecar container, responsible for providing
  artifacts to those who might depend on this step

For instance, `task` in the example above would look like:

- having an "input fetcher" capable of retrieving data from `repository` and
  `image`
- having an "output streamer" capable of streaming the artifacts it produces to
  the "bucket" step ("bucket" would pull from the output streamer endpoint)

so that in the end, we have this form of "peer-to-peer" communication between
the pods themselves:

*(arrow direction indicating "depends on")*

Once done with the build, the regular internal Concourse container lifecycle
would take care of moving the pods in our DB for `CREATED` state to
`DESTROYING`, which the kubernetes implementation of a worker would then notice
the desire to not having certain pods, and then proceed with deleting them.

### concerns about the current design

The current approach presented here is *very* focused towards not changing much
of the current constructs in our codebase *as is of today*.

Despite demonstrating that *it is possible to run Concourse on Kubernetes*, the
current design might raise some eyebrows.

#### insecure registry whitelisting

Given that the container runtimes that `kubelet`s communicate with need to trust
the registries that they interact with, we have to rely on whitelisting the internal
pod domain as trusted insecure registries.

Interestingly, this is already done by default on GKE.

##### next iterations

To work around this, we can explore the avenue of having the images being pushed
to a central registry in the cluster that's trusted by the kubelets.

As long as we make the process of getting "input in" and "outputs out" pluggable
enough, we could have either/or.

#### `exec`ing via `apiserver`

It's very convenient to be able to just `exec` (or attach to) a process via
`apiserver`'s `exec` endpoints, making the current Concourse's container
lifecycle and process execution work very with pretty much no changes needed in
our code flow.

It might be that this does not scale though, with `apiserver` being hit so hard
with a system like `concourse`.

There can also be concerns about keeping that connection for a long time (e.g.,
steps that take a long time to finish their main executions), or the sheer size
of throughput necessary (steps that log tons to `stderr`).

##### next iterations

If we don't want to modify the imperative nature of ATC making the requests to
execute, we could have some form of "shim" that'd be mounted to every main
container in step pods, which would then be reached out to by ATC, performing
the task of `exec`in processes there, and dealing with the log streams /
re-attaching.

#### `sync`ing handles by fetching all concourse pods

Having a pod per Concourse container will necessarily mean that we'd have a pod
for each resource scope id (assuming no use of [ephemeral check containers]).

That means that an installation with 5000 of those would be fetching 5000 pods
on every worker tick.

[ephemeral check containers]: https://project.concourse-ci.org/projects/MDc6UHJvamVjdDM5Nzg3OTk=

##### next iterations

This could be improved by leveraging the same mechanisms that controllers do -
perhaps even making this a controller itself that gets informed on changes to
pods that match the label that we have?

#### `atc/worker` package

- added `RunCheckStep` to the `atc/worker.Client` interface + implemented the
method for the `atc/worker.client` (garden-based worker) concrete type

  - work that'd be done anyway: https://github.com/concourse/concourse/issues/4957

  - allows us to implement the kubernetes-based checking logic by implementing
  that method in the kubernetes client implementation

#### `atc/atccmd/kubernetes.go`

takes care of ingesting either a `kubeconfig` or the mountpoint for service
accounts to get "tenancy" to a namespace in a kubernetes cluster where it can
act upon.

e.g., when targetting an externla cluster

```
--kubernetes-worker-kubeconfig=~/.kube/config
```

or, with `web` being a pod in the same cluster:

```
--kubernetes-worker-service-account=/var/run/secrets/kubernetes.io/serviceaccount
```

ps.: in the case of `kubeconfig`, the current context is chosen, and a namespace
must be specified (no defaults are assumed).

#### `atc/worker/kubernetes` package

`Client`implements "the runtime interface" (`atc/worker.Client`)

- `target` concrete type
  - similar to what `beacon` is for garden
    - deals w/ the lifecycle of our kubernetes targets
      - heartbeating
      - saving its info in the workers table
      - retiring
    - deals w/ gc & syncing handles (atc/handles.Syncer)
    - used by `atc/atccmd` as a runner (target.NewTargetRunner)

#### `atc/worker/kubernetes/backend` package

implementation of the verbs to deal w/ concourse containers (create, run,
delete, lookup, list, etc)

- allows us to go with an approach that is very similar to how we already do
with `garden`:

  1. create a sandbox (pod)
  2. prepare inputs
  3. run a process in it
  4. delete

#### `atc/worker/kubernetes/backend/backend.go`

- `backend.New()`

in the initialization, I hardcoded a QPS (queries per second) and a burst, but
there's currently no way to tell whether we're hitting our limits or not.

it'd be good to be able to tell how often we reach that (or, at least reason
about possible strategies to handle backpressure, if that's a thing that we
should handle ourselves).

- `backend.Create`

it currently just waits until a pod is running.

given that the pod will go through its lifecycle (see [`pod phase`]), possibly spending quite some
time in the `pending` phase if autoscaling needs to be performed, it might be
good to have a way of notifying the user that that's going on (or not? seems
like a UX question too).

[`pod phase`]: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase

there's also no timeouts being set at the moment, but that should be something
an operator could tune.

- `backend.Lookup`

currently, it just searches for a pod that matches the handle that we're
searching for.

this means that it can also look at containers that are in phases that are not
recoverable (e.g., terminating, succeeded, failed, or possibly even unknown)

#### `atc/handles` package

deals with container/volumes handles

- Syncer interface

  - takes care of providing the ability to sync state between what the database
  has as the truth and what a worker thinks the truth is with regards to handles

    - volume_syncer & container_syncer
      - with kubernetes target being a concourse worker managed by `atc`,
      there's no need to have `atc` reaching out to itself for
      ReportWorkerContainers.

      - having this logic decloupled from the `api`
        allows us to have it used for k8s too

    - updated `atc/api/containerserver` and `atc/api/volumeserver` to take
    `handles.Syncer`.

#### `atc/exec/task_step.go`

refactored it a bit to make it not depend on volumes when performing the
artifacts registration, and instead, consume the set of artifacts that were
produced by the implementor of `atc/worker.RunTaskStep`.

this way the bahavior looks very similar to how we do with `get_step` - as part
of `result`, you give the artifacts.

#### `atc/worker/kubernetes/backend/container.go`

- `container.PortForward` provides an implementation for how we could get access
  to servers that a container has. While in the demo I useed that for streaming
  files neceessary for `task.config` to get to an ATC living outside the
  cluster, that could *potentially* be used for inter-cluster communication

- `container.Run` provides a way for us to run a given process inside the
  main container in the pod. Using `exec`, we're able to not only run any
  process we want, but also stream both stdout and stderr to ourselves at ease
  (very important for the way that we deal with the results of resources).

#### `atc/worker/kubernetes/backend/pod.go`

`backend.Pod` provides the ability to create a pod definition from a set of
options that bring the desired pieces.

e.g., if we want a pod which has baggageclaim, an inputs fetcher, and a main
container:

```
podDef := Pod(
        WithBase(handle),
        WithBaggageclaim(),
        WithInputsFetcher(),
        WithMain(imageUri),
)
```

#### `atc/worker/kubernetes/backend/runner.go`

implements the `runtime.Runner` interface, used by the `resource` package.

this is pretty much just building upon the what `backend/container.go`
provides.

#### `atc/worker/kubernetes/backend/creator.go`

provides the same kind of functionality that `FindOrCreateContainer` brings in
the Garden case:

- deals with the database side of it, as well as reaching out to the "container
  system".
  - in this case, it reaches out to kubernetes, to create pods.

- deals with having the base image available so that the container system can
  start the container from it (leverages `atc/worker/kubernetes/fetcher.go`)
  - `image`: resolves the location of the pod that performed the `get` for that
    image, providing the address of the baggageclaim server of that pod as the
    image uri
    - as `baggageclaim` itself is able to serve image volumes as container
      images, `kubelet`s can fetch from it.
  - `image_resource`: runs `check` and `get`, etc to fetch an image, which then
    becomes available just how we do for `image`
  - `rootfs_uri`: pases the name down to kubernetes directly

- deals with providing the inputs for the container
  - if there are inputs that need to be fetched before the main container
    starts, an `initContainer` is placed in the pod definition to fetch the
    necessary bits from those other pods
    - this means that we're performing direct streaming.
      - when thinking about connecting to other clusters / networks, it'll be
        important to be able to tell where those artifacts are coming from, so
        that we can use different methods to getting those bits in (might be in
        the form of a gateway).

#### `atc/worker/kubernetes/pod_artifact.go`

handles serialization of the location of artifacts in the cluster by creating a
string that contains all of the information necessary for the `exec` package to
pass around the details of where inputs live.

e.g.:

1. get step's `Run()` is executed, producing a result, with an artifact that has
in its `id`, all of the information of where the pod that contains the artifact
lives.
2. the artifact is registered in the artifact repository
3. a task step's `Run()` is executed
4. it looks up the artifact repository, and gets our string, passing that down
to our worker client
5. the worker client takes that information, deserializes it, and then creates
the task's pod with `initContainer` that is able to fetch its inputs.

ps.: builds upon https://github.com/concourse/baggageclaim/pull/36


zeusro:
Waiting for your good news !

cirocosta:
hey folks,

I'll close this PR for now - as I mentioned, it's meant to be just an exploration to prove that it's totally possible to get Concourse to run on Kubernetes using Kubernetes constructs, which was indeed demonstrated here.

That said, we'll continue with the work of first getting the `containerd` details fleshed out, we'll get back to the Kubernetes track of work and move full steam on it :+1:  

thank you!

Issue #5220, opened 2020-02-21
Pinning and `passed` constraints should be mutually exclusive


vito:
When a resource is pinned, it should only take effect for jobs that use the input without `passed` constraints.

Otherwise pinning would imply that all of those jobs need to run with a set of candidates which both use the version and satisfy the `passed` constraints. In practice, a user is not likely to want all of those jobs to fire at once; instead it makes more sense for the farthest upstream job to run with the pinned version and then propagate it downstream gradually.

vito:
Closing per https://github.com/concourse/concourse/issues/5274

PR #5224, opened 2020-02-24
Pinning and `passed` constraints should be mutually exclusive

clarafu:
[#5220]

Co-authored-by: Alex Suraci <suraci.alex@gmail.com>

When a resource is pinned, it should only take effect for jobs that use the input without passed constraints.

Otherwise pinning would imply that all of those jobs need to run with a set of candidates which both use the version and satisfy the passed constraints. In practice, a user is not likely to want all of those jobs to fire at once; instead it makes more sense for the farthest upstream job to run with the pinned version and then propagate it downstream gradually.

### DIFF
--- a/atc/scheduler/algorithm/group_resolver.go
+++ b/atc/scheduler/algorithm/group_resolver.go
@@ -61,26 +61,6 @@ func (r *groupResolver) Resolve(ctx context.Context) (map[string]*versionCandida
 	})
 	defer span.End()
 
-	for i, cfg := range r.inputConfigs {
-		if cfg.PinnedVersion == nil {
-			continue
-		}
-
-		version, found, err := r.vdb.FindVersionOfResource(ctx, cfg.ResourceID, cfg.PinnedVersion)
-		if err != nil {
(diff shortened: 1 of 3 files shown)


atc/scheduler/algorithm/group_resolver.go
```diff
@@ -61,26 +61,6 @@ func (r *groupResolver) Resolve(ctx context.Context) (map[string]*versionCandida
...
-			return nil, "", err
-		}
-
-		if !found {
-			notFoundErr := db.PinnedVersionNotFound{PinnedVersion: cfg.PinnedVersion}
-			span.SetStatus(codes.InvalidArgument)
-			return nil, notFoundErr.String(), nil
-		}
-
-		r.pins[i] = version
```
muntac: As the `pins` field in `groupResolver` is no longer being explicitly used does it make sense to remove it from the struct entirely? It would then also not need to be set on line 47.

clarafu:
I'm going to close this PR because we are backing out from this change #5274

PR #5237, opened 2020-02-25
contributing: add topgun section

cirocosta:
### why do we need this PR

- despite us not enforcing that contributors add topgun tests / run topgun when
  submitting PRs, sometimes we need to do so, but that flow isn't documente
  anywhere.

- BOSH lite seems to have problems with `btrfs`, making testing locally not
  possible w/out a manifest change.

### changes proposed

- update the contributing section with instructions to get started with running
  topgun locally

- update the base manifest to use `overlay` rather than the default (detect,
  which will prioritize btrfs)

---

cc @jtarchie

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -498,7 +498,50 @@ ginkgo .
 
 The `topgun/` suite is quite heavyweight and we don't currently expect most
 contributors to run or modify it. It's also kind of hard for ~~mere mortals~~
-external contributors to run anyway. So for now, ignore it.
+external contributors to run anyway.
+
+To run `topgun`, a BOSH director up and running is required - the only
+requirement with regards to where BOSH sits is having the ability to reach the
+instance that it creates (the tests make requests to them).
+
+You can have a local setup by leveraging the `dev` scripts in
+the [concourse-bosh-release] repo:
+
+[concourse-bosh-release]: https://github.com/concourse/concourse-bosh-release
+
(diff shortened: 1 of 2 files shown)


pivotal-jamie-klassen:
@pivotal-jwinters you said something about topgun 'feature-flagging itself' the other day - maybe something to add here (or in a follow-on PR)?

jtarchie:
As we are trying to run this locally, we've learned that there is a requirement on the `bosh` CLI version.
We are testing against `bosh` 6.2.0, but the `concourse/unit` docker image contains `bosh` 5.3.1.

This [regex](https://github.com/concourse/concourse/blob/7ec73b9be6203dd5e1ee7aaecd4e92daad9a6572/topgun/common/common.go#L594) does not match against the output from the newer bosh CLI.

cirocosta:
> We are testing against bosh 6.2.0, but the concourse/unit docker image contains bosh 5.3.1.
> This regex does not match against the output from the newer bosh CLI.

https://github.com/concourse/ci/issues/278

pivotal-jwinters:
Topgun is smart about what tests it runs:

https://github.com/concourse/concourse/blob/02ee0ee6ebdecec943e91e0d4e8e571908a2064f/topgun/core/credhub_test.go#L31-L33

I mentioned it in passing, more as an interesting note. It's probably not worth documenting because it's really straightforward for anyone looking at these tests.

PR #5239, opened 2020-02-25
atc: check step refactor (move from `exec` to `worker`)

### LINKED ISSUE #4957: Refactor Check Step

cirocosta:
### existing issue

fixes #4957 

### changes proposed in this pull request

- move the garden logic from the `exec` package to `worker`

ps.: this is strictly a refactor (no changes in behavior) 

### Contributor Checklist

- [x] Unit tests
- ~[ ] Integration tests (if applicable)~
- ~[ ] Updated documentation (located at https://github.com/concourse/docs)~
- ~[ ] Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)~

### Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/exec/check_step.go
+++ b/atc/exec/check_step.go
@@ -136,7 +136,7 @@ func (step *CheckStep) Run(ctx context.Context, state RunState) error {
 		return fmt.Errorf("run check step: %w", err)
 	}
 
-	err = step.delegate.SaveVersions(result)
+	err = step.delegate.SaveVersions(result.Versions)
 	if err != nil {
 		return fmt.Errorf("save versions: %w", err)
 	}
--- a/atc/exec/check_step.go
+++ b/atc/exec/check_step.go
@@ -11,7 +11,6 @@ import (
 	"github.com/concourse/concourse/atc/creds"
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/atc/resource"
-	"github.com/concourse/concourse/atc/runtime"
 	"github.com/concourse/concourse/atc/worker"
 )
 
@@ -25,6 +24,7 @@ type CheckStep struct {
 	pool              worker.Pool
 	delegate          CheckDelegate
 	succeeded         bool
+	workerClient      worker.Client
 }
 
 //go:generate counterfeiter . CheckDelegate
@@ -44,6 +44,7 @@ func NewCheckStep(
 	strategy worker.ContainerPlacementStrategy,
 	pool worker.Pool,
 	delegate CheckDelegate,
+	client worker.Client,
 ) *CheckStep {
 	return &CheckStep{
 		planID:            planID,
@@ -54,6 +55,7 @@ func NewCheckStep(
 		pool:              pool,
 		strategy:          strategy,
 		delegate:          delegate,
+		workerClient:      client,
 	}
 }
 
@@ -72,6 +74,12 @@ func (step *CheckStep) Run(ctx context.Context, state RunState) error {
 
 	resourceTypes, err := creds.NewVersionedResourceTypes(variables, step.plan.VersionedResourceTypes).Evaluate()
 	if err != nil {
+		return err
+	}
+
+	timeout, err := time.ParseDuration(step.plan.Timeout)
+	if err != nil {
+		logger.Error("failed-to-parse-timeout", err)
 		return err
 	}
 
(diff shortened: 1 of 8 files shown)


cirocosta:
ps.: this is a prerequisite for #5209

cirocosta:
(ready for review now :grin: )

atc/worker/client.go
```diff
@@ -36,6 +36,19 @@ type Client interface {
...
+		logger lager.Logger,
+		owner db.ContainerOwner,
+		containerSpec ContainerSpec,
+		workerSpec WorkerSpec,
+		strategy ContainerPlacementStrategy,
+		containerMetadata db.ContainerMetadata,
+		resourceTypes atc.VersionedResourceTypes,
+		timeout time.Duration,
+		checkable resource.Resource,
+	) (result []atc.Version, err error)
```
pivotal-jwinters: Should this return a CheckResult similar to all the other step interface methods?

atc/worker/client.go
```diff
@@ -156,6 +169,66 @@ func (client *client) CreateVolume(logger lager.Logger, volumeSpec VolumeSpec, w
...
+		logger,
+		&NoopImageFetchingDelegate{},
+		owner,
+		containerMetadata,
+		containerSpec,
+		resourceTypes,
+	)
+	if err != nil {
+		err = fmt.Errorf("find or create container: %w", err)
+		return
```
taylorsilva: Why are we setting `err` to a new error here? We're throwing it away if we end up here aren't we since we return on the second line? Should we be using `logger` here?

pivotal-jwinters: It uses named returns args, but none of the other implementations do this so maybe we should change this too @cirocosta?

cirocosta: oh hahah yeaah, sorry for that - I just have a tendency of always using named return args 😅  will address that

atc/worker/client.go
cirocosta: good ppoint!

atc/worker/client.go
cirocosta: https://github.com/concourse/concourse/pull/5239/commits/4431ba83fa376ad1cf5cfb0b537f237ae7d34e54

atc/worker/client.go
cirocosta: https://github.com/concourse/concourse/pull/5239/commits/c14cbf8230275ec71fd3bc124d89d4b4a7bffcff

atc/exec/check_step.go
```diff
@@ -105,59 +112,33 @@ func (step *CheckStep) Run(ctx context.Context, state RunState) error {
...
 
-	err = step.delegate.SaveVersions(versions)
+	err = step.delegate.SaveVersions(result.Versions)
 	if err != nil {
-		logger.Error("failed-to-save-versions", err)
-		return err
+		return fmt.Errorf("save versions: %w", err)
 	}
 
 	step.succeeded = true
```
pivotal-jwinters: Maybe we could give `CheckResult` an `ExitStatus` and make this similar to:
https://github.com/concourse/concourse/blob/aac7b2f461ed2a3ad6fa8d861d439d3774f9fde8/atc/exec/task_step.go#L248

atc/worker/client.go
```diff
@@ -156,6 +173,62 @@ func (client *client) CreateVolume(logger lager.Logger, volumeSpec VolumeSpec, w
...
+	versions, err := checkable.Check(deadline, checkProcessSpec, container)
+	if err != nil {
+		if err == context.DeadlineExceeded {
+			return CheckResult{}, fmt.Errorf("timed out after %v checking for new versions", timeout)
+		}
+
+		return CheckResult{}, fmt.Errorf("check: %w", err)
+	}
+
+	return CheckResult{Versions: versions}, nil
```
pivotal-jwinters: Can we give the CheckResult an ExitStatus similar to:
https://github.com/concourse/concourse/blob/aac7b2f461ed2a3ad6fa8d861d439d3774f9fde8/atc/worker/client.go#L553-L560

atc/exec/check_step_test.go
```diff
@@ -3,260 +3,176 @@ package exec_test
...
-			Expect(actualFromVersion).To(Equal(checkPlan.FromVersion))
+		It("errors", func() {
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("invalid duration"))
 		})
+	})
 
-		It("runs the check resource action", func() {
-			Expect(fakeResource.CheckCallCount()).To(Equal(1))
+	Context("with a resonable configuration", func() {
```
xtreme-sameer-vohra: Do you think it would be valuable to have a test case to ensure the worker.Client receives the correct args ?

cirocosta: yeahh, sounds like a good thing to test 😁

atc/worker/client.go
xtreme-sameer-vohra: Yes, please.
This allows us to differentiate script errors vs. Concourse errors

cirocosta: I'm a bit confused with how we should proceed here :thinking:

I see (and agree) that we want to be able to differentiate system errors from
executables returning a non-0 exit status, but, it seems like the current
behavior of "check error reporting" is tied to returning an error with that
stuff embedded into it, and I think changing that wouldn't be the scope of this
PR (which was meant to be structural only).

I say that because there seems to be a gap between the way that errors are
reported for `check`s and `get`s`/`put`s/`task`s:

- `check`s report *how* they errored through errors, being even recorded in the
  database under the `checks` table - the "reason" for the non-0 exit status is
  embedded in the error message from `runtime.ErrResourceScriptFailed`

https://github.com/concourse/concourse/blob/1ae69250ceef7480f56732377478c46642c8791a/atc/runtime/errors.go#L25-L31

- `get`s, `put`s, and `task`s have their "reason for a non-0" reported through
  their corresponding delegates: one can tell *how* a `get` failed by looking at
  the build logs, with `stderr` being directly tied to db events

wdyt?

thx!

atc/exec/check_step_test.go
cirocosta: ef039c956e532d58a8ba9ad53a43eba102a3b5ae

atc/worker/client.go
pivotal-jwinters: Yeah makes sense! 👍

PR #5244, opened 2020-02-26
Extract literals to constants and some code cleanup

kirillbilchenko:
# Existing Issue
N/A

Fixes # .
N/A
# Why do we need this PR?
With this PR just move some often repeated strings to constants, and added two wrapper to handle errors and result and add correspondence status code to response

* Add two wrapper functions to reduce number of code lines, fix some typos, extracted some string literals to constants
* Pr also fixing failing unit test `request to garden times out eventually`

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
- [x] Code reviewed
- [ ] Tests reviewed
- [ ] PR acceptance performed


atc/api/configserver/save.go
```diff
@@ -18,6 +18,11 @@ import (
 	"github.com/tedsuo/rata"
 )
 
+const (
+	JsonContent = "application/json"
```
muntac: HTTP headers are pretty well defined (i.e. does not change) and understood across languages and platforms. I think it's easier to understand the direct literals in this case than references.

The headers also appear across the system in at least 40 more files. To maintain consistency and hence readability all those would have to be changed as well, which would be a pretty big diff without adding much in the way of readability.

It would be better to retract the header related changes.

atc/api/jobserver/list_builds.go
```diff
@@ -22,7 +22,7 @@ func (s *Server) ListJobBuilds(pipeline db.Pipeline) http.Handler {
 			limit      int
 		)
 
-		logger := s.logger.Session("list-job-builds")
+		_ = s.logger.Session("list-job-builds")
```
aoldershaw: I notice you're ignoring this logger, which makes this line a no-op - the returned logger has additional context, which `s.logger` does not.

Rather than adding a `logger` parameter to `logIfErrorAndRespond`, I'd argue that it's better to get rid of `logIfErrorAndRespond` and `checkResultAndRespond`. IMO it's more idiomatic Go to make the error handling obvious and the code simple - even though the current checks are more lines of code, I find it easier to read. It's also easier to modify the behaviour of specific branches (for instance, changing the `logger` that gets used).

That said, if you have a contrary opinion, I'd be very happy to hear you out

kirillbilchenko: agree I will address this, I just wanted to display the approach to not duplicate this logger calls, but readability and style of course if affected by this, will address this 👍

kirillbilchenko:
@aoldershaw @muntac comments addressed please check

PR #5252, opened 2020-02-27
feat(atc): make pipeline and job badge API routes public

### LINKED ISSUE #996: Feature request: public badges for private jobs
# Feature Request

GitHub badges for private jobs don't show up in GitHub READMEs.
Here's how I see them:

image

## A Modest Proposal

It'd be lovely to have an option to make badges for a job available publicly, regardless of the job being private.

jhosteny:
fixes #996

Signed-off-by: Joe Hosteny <jhosteny@carnegierobotics.com>

# Existing Issue

Fixes #996.

# Why do we need this PR?

Make the badge routes public, so that we can see the build status for private pipelines and jobs without authentication (e.g., in a github README).

* Make PipelineBadge public 
* Make JobBadge public

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
- [x] Code reviewed
- [x] Tests reviewed
- [ ] Release notes reviewed
- [x] PR acceptance performed

### DIFF
--- a/atc/api/jobs_test.go
+++ b/atc/api/jobs_test.go
@@ -549,10 +549,11 @@ var _ = Describe("Jobs API", func() {
 			Context("and the pipeline is private", func() {
 				BeforeEach(func() {
 					fakePipeline.PublicReturns(false)
+					fakePipeline.JobReturns(fakeJob, true, nil)
 				})
 
-				It("returns 403", func() {
-					Expect(response.StatusCode).To(Equal(http.StatusForbidden))
+				It("returns 200", func() {
+					Expect(response.StatusCode).To(Equal(http.StatusOK))
 				})
 			})
 
--- a/atc/api/pipelines_test.go
+++ b/atc/api/pipelines_test.go
@@ -574,8 +574,8 @@ var _ = Describe("Pipelines API", func() {
 					BeforeEach(func() {
 						fakeaccess.IsAuthenticatedReturns(true)
 					})
-					It("returns 403", func() {
-						Expect(response.StatusCode).To(Equal(http.StatusForbidden))
+					It("returns 200", func() {
+						Expect(response.StatusCode).To(Equal(http.StatusOK))
 					})
 				})
 
@@ -584,8 +584,8 @@ var _ = Describe("Pipelines API", func() {
 						fakeaccess.IsAuthenticatedReturns(false)
 					})
 
-					It("returns 401", func() {
-						Expect(response.StatusCode).To(Equal(http.StatusUnauthorized))
+					It("returns 200", func() {
+						Expect(response.StatusCode).To(Equal(http.StatusOK))
 					})
 				})
 			})
--- a/atc/wrappa/api_auth_wrappa.go
+++ b/atc/wrappa/api_auth_wrappa.go
@@ -65,8 +65,6 @@ func (wrappa *APIAuthWrappa) Wrap(handlers rata.Handlers) rata.Handlers {
 		// pipeline is public or authorized
 		case atc.GetPipeline,
 			atc.GetJobBuild,
-			atc.PipelineBadge,
-			atc.JobBadge,
 			atc.ListJobs,
 			atc.GetJob,
 			atc.ListJobBuilds,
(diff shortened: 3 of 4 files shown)


pivotal-jamie-klassen:
Hmmm, thinking this through a bit I'm realizing it introduces a backwards-incompatible change that will reveal all pipeline badges that heretofore were hidden. I can imagine this would be an unpleasant side effect for some operators.

In the description of #996 it says
> have an option to make badges for a job available publicly

I think maybe we should think about how to leave the default behaviour in-place. I can think of a couple of options:
* (simple) a cluster-wide flag set by the operator that exposes all badge endpoints
* (complex) a field in the pipeline/job config, separate from `expose-pipeline` or the `public:` field, that determines badge visibility.

@jhosteny @gabro how do y'all feel about this?

jhosteny:
Hi @pivotal-jamie-klassen,

Because we'd probably make all badges public, and I'd prefer to have this when the next 6.0 rc or beta comes out, I'd opt for the global flag. However, I could understand why folks would want the per-pipeline setting.

Would allowing _both_ options be conceivable? That is, the flag could set the default visibility of badges (i.e., `public` or `repo-visibility`, to match the repo setting), and then the second change at some later date could override this setting, per pipeline? That might be nice if you generally wanted public badges for private repos, but there were one you wouldn't even want to leak the name of. For backwards compatibility, the flag would be set to `repo-visibility` if unset.

pivotal-jamie-klassen:
@jhosteny it sounds like we're on the same page. Ideally the 6.0 beta includes this option, and the least-risky change to make this viable would be the global flag. Maybe we just mark it as "experimental" and punt on figuring out the per-pipeline/per-job config once we can do a bit more user research.

Your proposal of planning with this future state in mind sounds great. We could call the flag `--default-badge-visibility` or something similar.

Would you be comfortable taking on this follow-on work? It's not in the direct roadmap for the full-time Concourse contributors, so if the time came to ship a 6.0 beta and we hadn't gotten around to it, I'm worried we might end up reverting this PR. If you can't commit the extra effort I can probably take it on in my own flex time, but I wouldn't want to step on your toes.

jhosteny:
Hi @pivotal-jamie-klassen,

I can give it a shot, but I am fairly new to the concourse codebase, so it might not make it in time. Do you know offhand if there are existing examples of API routes which are conditionally authorized based on a setting? If not, no problem - I'll start digging into the atc API auth code more.

pivotal-jamie-klassen:
Hi @jhosteny, I'm happy to give some guidance. I don't feel especially proud of the access control system in Concourse, so I'd be very curious to learn more about your contribution experience. Hopefully any WTF moments you have could be surfaced and ironed out!

As for driving the behaviour via a test (read: I'm a London-style TDD person), the [RBAC suite](https://github.com/concourse/concourse/blob/master/atc/integration/rbac_test.go#L115-L121) has some good examples of the level of integration test that could be used to exercise this functionality as a whole. Unfortunately the [`Team` interface](https://github.com/concourse/concourse/blob/master/go-concourse/concourse/team.go#L12-L75) in `go-concourse` doesn't have any badge-related methods (so the `ccClient` in those tests couldn't be used directly, but you can ask it for its [`http.Client`](https://github.com/concourse/concourse/blob/master/go-concourse/concourse/client.go#L16) to make your own requests), but maybe it should (!?).

I suspect that a good template for the kind of behavior you'd want to add is the [`checkPipelineAccessHandler`](https://github.com/concourse/concourse/blob/f3d4c0c/atc/api/auth/check_pipeline_access_handler.go). It is used to introduce a 'configuration-dependent access control' (similar in spirit to what you're looking for) based on whether the pipeline is exposed.

Here's my two cents on an architectural approach:

I could imagine adding a `checkBadgeAccessHandler` and `checkBadgeAccessHandlerFactory` - the factory could be instantiated [alongside the other handler factories of this kind](https://github.com/concourse/concourse/blob/f3d4c0c/atc/api/handler.go#L78-L80) (probably you'd need to add a new parameter to `NewHandler` for the default badge visibility, and pass it in the new configuration flag [here](https://github.com/concourse/concourse/blob/f3d4c0c22c8cd02c48f1ae95ae98d51d26e0f300/atc/atccmd/command.go#L1257-L1292) where the overall handler gets constructed) - maybe you'd pass the `checkPipelineAccessHandler`s [here](https://github.com/concourse/concourse/blob/f3d4c0c/atc/api/handler.go#L120) and [here](https://github.com/concourse/concourse/blob/f3d4c0c/atc/api/handler.go#L141) into your new handler factory.

If you settle on an approach like this, you'd probably be able to drive your change by modifying the existing tests under `atc/api`, plus probably unit tests for the brand new interfaces.

Thinking about it a bit longer, the above approach might be _slightly_ over-engineered, but it would be very consistent with the existing structures.

jhosteny:
@pivotal-jamie-klassen thanks, that's really helpful! I have a change that handles the global flag. I am working on tests for it next.

PR #5254, opened 2020-02-27
worker: fix container-sweeper test

muntac:
# Why do we need this PR?
Test started failing more often with upgrade to go 1.14. Error message is now different.

Updated error message.


kirillbilchenko:
@muntac can you check my pr it's already with those fix #5244

muntac:
@kirillbilchenko Thanks for the PR! I will take a deeper look into it soon. This broken test isn't related to the issue you brought up, and is blocking some of our pipelines, so we need to get the fix in quickly on its own.

PR #5257, opened 2020-02-27
Optimize user tracking

xtremerui:
# Why do we need this PR

This PR adds optimizations during the authorization flow for each api request. Every time concourse handles an api request it needs to pull down the list of all teams, so that it can iterate over the auth configuration to determine if the current user has access to that teams information. On every request it also tracks the user who made the request and writes this information to the db so that we can determine the last time the user was active in concourse (this is important for billing).

So this PR introduces two optimizations:
- caching the `GetTeams` call for 1 minute (which gets a db notification to invalidate on `SetTeam`)
- batching `CreateOrUpdateUser` writes to the db every 15s (or a batch size of 100 items)

# Benchmark results:

With batch and cache. Running 30s.
```
Running 30s test @ http://localhost:8080/api/v1/pipelines
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     7.62ms    3.60ms  46.72ms   84.30%
    Req/Sec   677.27     63.23   838.00     71.17%
  40466 requests in 30.02s, 10.88MB read
Requests/sec:   1348.02
Transfer/sec:    371.23KB
```

Master branch
```
Running 10s test @ http://localhost:8080/api/v1/pipelines
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     6.60ms    4.06ms  36.40ms   88.42%
    Req/Sec   812.27     85.47     0.97k    64.50%
  16174 requests in 10.00s, 4.35MB read
Requests/sec:   1617.05
Transfer/sec:    445.32KB
```

Without batch and cache `use-dex-token` branch
```
Running 10s test @ http://localhost:8080/api/v1/pipelines
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    16.93ms    7.44ms  83.86ms   79.78%
    Req/Sec   301.43     30.99   363.00     74.50%
  6007 requests in 10.01s, 1.62MB read
Requests/sec:    600.24
Transfer/sec:    165.30KB
```


cirocosta:
@xtremerui do you think we should also benchmark against 5.8 / 6.0 before the tokens pr just to have a baseline of how we're doing against what we had previously?

xtremerui:
yeah thats the plan. We need to know what improvement is now. The above benchmarks could be obslete already.

pivotal-jwinters:
@cirocosta this is ready for review

cirocosta:
Hey @pivotal-jwinters @xtremerui , 

could you give a brief description of how we're accomplishing what we want here? i see the use of the `batch` library here, but, could you elaborate on how it does the sync? e.g., is it writing in batch based on a given interval? buffer getting full / ratio of some sorts?

thanks!

pivotal-jwinters:
@ciro I updated the description with more info

cirocosta:
Hey,

I was doing some testing here of this, and wanted to share the results i've got.

The setup:

1. static rate of requests (1500/s), hold for 15s
2. `web` (varying its versions) and `db` (same version: 12.2) running in docker,
w/ no limits set, in the same machine
  - versions tested: 5.8.0, 6.0.0, master
    (5e315c6ccfe2fdb087ecb40a7a5339b5dc31feb5), and this PR
    (8f0d9efc84488fd61801125aeefd7df580b1486a)

3. Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz (turbo boost 5GHz)

Naturally, all four scenarios worked just fine hitting `/api/v1/info/pipelines`
with all of the requests authenticated.

```
5.8.0

        Requests      [total, rate, throughput]         22500, 1500.07, 1499.98
        Duration      [total, attack, wait]             15s, 14.999s, 906.663µs
        Latencies     [min, mean, 50, 90, 95, 99, max]  436.009µs, 931.217µs, 932.175µs, 1.259ms, 1.416ms, 2.358ms, 11.186ms
        Bytes In      [total, mean]                     67500, 3.00
        Bytes Out     [total, mean]                     0, 0.00
        Success       [ratio]                           100.00%
        Status Codes  [code:count]                      200:22500
        Error Set:


6.0.0
        Requests      [total, rate, throughput]         22500, 1500.08, 1500.03
        Duration      [total, attack, wait]             15s, 14.999s, 490µs
        Latencies     [min, mean, 50, 90, 95, 99, max]  433.87µs, 1.049ms, 995.682µs, 1.227ms, 1.413ms, 2.607ms, 88.059ms
        Bytes In      [total, mean]                     67500, 3.00
        Bytes Out     [total, mean]                     0, 0.00
        Success       [ratio]                           100.00%
        Status Codes  [code:count]                      200:22500
        Error Set:


master
        Requests      [total, rate, throughput]         22500, 1500.09, 1482.51
        Duration      [total, attack, wait]             15.177s, 14.999s, 177.851ms
        Latencies     [min, mean, 50, 90, 95, 99, max]  4.923ms, 198.958ms, 115.691ms, 412.016ms, 697.929ms, 1.458s, 4.69s
        Bytes In      [total, mean]                     67500, 3.00
        Bytes Out     [total, mean]                     0, 0.00
        Success       [ratio]                           100.00%
        Status Codes  [code:count]                      200:22500
        Error Set:


this pull request
        Requests      [total, rate, throughput]         22500, 1500.07, 1499.94
        Duration      [total, attack, wait]             15.001s, 14.999s, 1.243ms
        Latencies     [min, mean, 50, 90, 95, 99, max]  550.363µs, 976.507µs, 926.015µs, 1.296ms, 1.426ms, 2.318ms, 16.551ms
        Bytes In      [total, mean]                     67500, 3.00
        Bytes Out     [total, mean]                     0, 0.00
        Success       [ratio]                           100.00%
        Status Codes  [code:count]                      200:22500
        Error Set:
```

Note how `master`'s latencies are all over the place, from 4.93 millisecs up to
4.69 seconds - even its median is already higher than the max of any of the
other ones.

with regards to resource consumption, we're back to being very close to what we
have before too (look at the last bump there - this PR - and the previous -
`master`, where both DB and WEB utilization was higher).

Screenshot from 2020-04-17 14-21-48


good job!!

pivotal-jwinters:
@cirocosta I'm actually surprised that you got that throughput on `master`.

How many pipelines did you have configured? We noticed that the performance boost from this PR is much less noticeable once you have a certain number of pipelines (assuming you're hitting the /api/v1/pipelines endpoint)

cirocosta:
@pivotal-jwinters oh, I had 0 pipelines set - my rationale is that we wanted to just benchmark the effect of the accessor :thinking: was that wrong?

pivotal-jwinters:
@cirocosta I think that's a fair assumption. I just saw a much lower throughput on master than what you observed, so I was trying to rationalize.

atc/api/accessor/batcher.go
```diff
@@ -0,0 +1,88 @@
...
+
+	for item := range ps.Input {
+		if user, okParse := item.Get().(user); okParse {
+			users[user.sub] = user
+		} else {
+			ps.Errors <- errors.New("failed-to-parse-user")
+		}
+	}
+
+	err := b.userFactory.BatchUpsertUsers(users)
```
cirocosta, line 84: the whole `Process()` func is called asynchronously, but, should we get concerned about this taking longer than expected? 

e.g., if it turns out that postgres is having a really hard time and taking, say, 2m to respond to a simple exec statement, what would happen? would this block further calls to Process() that the library would make? could it block anything else?

atc/api/accessor/batcher.go
```diff
@@ -0,0 +1,88 @@
...
+	if err != nil {
```
cirocosta, line 85: this is (and the other error checking above) are currently not covered:

Screenshot from 2020-04-21 12-03-21

we *could* have that tested if we made the `batcher` type exported and then played around with having `fakeUserfactory` give an error there, and then us deal with the errors having access to the `.Errors` field of the PipelineStage object (that we'd supply). do you think that's something we should do?

atc/api/accessor/batcher.go
cirocosta: @pivotal-jwinters and I were looking at the details of how this method gets called by the `go-batch` library, and the conclusion is that even if this blocks for a long time, it'll not block further consumption of the channel where we writes users too (thus, will not block that path), and will eventually get it saved, doing an upsert, which is not a problem.

atc/api/accessor/batcher.go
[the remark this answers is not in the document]
cirocosta: @pivotal-jwinters I was thinking about this again, and we could actually assert that the `logger.Error` there is called when a failure from `BatchUpsertUsers` fails :thinking:  

wdyt of having a test for that?

We wouldn't be able to test that we're logging in the case of that parse error, which would be more of a programming mistake (we are the only ones that put items in that cache), but at least we'd be guaranteeing that we don't miss out on logging errors from the db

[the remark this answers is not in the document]
pivotal-jwinters: @ciro 45e199d3f873aa18ff62de8a6b37fca239863f66

pivotal-jamie-klassen:
👍 to release note - https://github.com/concourse/concourse/blob/master/release-notes/latest.md#4950 could probably be expanded to mention the DB impact.

release-notes/latest.md
```diff
@@ -1,26 +1,26 @@
...
-  NOTE: this is a breaking change. You'll neeed to add a couple required flags at startup `CONCOURSE_CLIENT_SECRET` and `CONCOURSE_TSA_CLIENT_SECRET`. And, yes, you will need to log out and log back in one last time.
+  NOTE: And yes, you will need to log out and log back in after upgrading.
 
 #### <sub><sup><a name="5305" href="#5305">:link:</a></sup></sub> feature
 
 * We've updated the way that hijacked containers get garbage collected
 
   We are no longer relying on garden to clean up hijacked containers. Instead, we have implemented this functionality in concourse itself. This makes it much more portable to different container backends.
 
+##### <sub><sup><a name="5305" href="#5305">:link:</a></sup></sub> feature
```
cirocosta, line 18: oh this is actually #5431 :thinking:

pivotal-jwinters: fixed

PR #5262, opened 2020-03-02
web: locally cache API responses for dashboard

### LINKED ISSUE #5193: web: preload pipelines/jobs from local storage on dashboard

One significant source of slowness for loading the dashboard on large deployments is fetching the list of all jobs/pipelines. Even though the pipelines and their shape (their jobs and the dependencies between them) will likely change infrequently, we currently have to wait for these network requests to resolve before rendering anything meaningful on the dashboard.

Since the pipelines on the dashboard are likely to look about the same (ignoring build statuses), we can get a much faster first render by locally caching (probably localStorage) the pipelines/jobs responses and first rendering the dashboard with that stale information - once the actual requests finish and we have up-to-date information, we can update the view.

This introduces a few risks/things to keep in mind:

1. Once the network request finishes, the rendered pipelines may shift - if this network request finishes quickly, the dashboard could feel fairly jumpy and erratic
2. Before the network request finishes, we're displaying stale info - we need to make it clear somehow that this isn't necessarily the current state of the world. Any recommendations on how we can make this obvious (and just feedback on the feature generally) would be appreciated @Lindsayauchin @matthewpereira 
3. Unlike the size/shape of the pipeline cards, the build statuses *are* likely to change between page loads. I'm not sure what the best approach is for displaying build statuses when we only have the cached value - it might even make sense to ignore the last known build status and render the jobs as having an "unknown" status (again, not sure how this should look)

Yep

aoldershaw:
# Existing Issue

Fixes #5193  .

# Why do we need this PR?

Previous efforts to improve the performance of the dashboard include https://github.com/concourse/concourse/pull/5118 and https://github.com/concourse/concourse/pull/5148. The former focuses on making the page feel more responsive once it's loaded by reducing the quantity of DOM the browser has to deal with, and the latter on optimizing the time to FCP (first contentful paint) by lazy-loading some Javascript.

This PR is related to https://github.com/concourse/concourse/pull/5148 in that it aims to make the page "load faster". Rather than just optimizing for the first contentful paint, however, here we try to optimize the first "meaningful" paint - that is, the time to render a dashboard that is usable in some way.

Typically, the biggest bottleneck for rendering meaningful content on the dashboard is *fetching* that meaningful content. Under the assumption that the pipelines on the dashboard don't change too frequently, we can render the dashboard quickly by displaying the last known state of these pipelines. This allows users to scroll through and search for pipelines before any network request completes. Once the network requests come in, we can swap in the fresh data in-place. Note that since build status is likely to change frequently, it doesn't make much sense to render this data when displaying the stale state.

For a quick demo of what this PR looks like:

ezgif-1-aeabaf97992f

Note: when running locally, I throttled my network to simulate a slow page load - it's just too fast when calling out to localhost. I used the "Slow 3G" setting on Chrome for this demo.

* Store/fetch the dashboard API responses (teams, pipelines, jobs) to/from local storage
* Introduce `Message.Storage` module to facilitate communication with `localStorage` (and `sessionStorage`)
* Introduce `FetchResult` module to represent data that can come from the network (`Fetched`), from cache (`Cached`), or nothing at all (`None`).

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


aoldershaw:
@pivotal-jamie-klassen Sorry about the large diff! I'll definitely try to find a more natural stopping point in the future rather than trying to jam everything into one big PR

Thanks for reviewing!

PR #5269, opened 2020-03-04
Add way to parse nested vault secret, in case if user request specific key from vault json data

### LINKED ISSUE #5253: CONCOURSE_VAULT_SHARED_PATH has no effect for secrets lookup
## Bug Report

After enabling CONCOURSE_VAULT_SHARED_PATH variable, Concourse keeps with the previous behavior as if this variable was not been seeing.

I have the following variables set:

```
CONCOURSE_VAULT_SHARED_PATH=defaults
CONCOURSE_VAULT_PATH_PREFIX=concourse
```
In Vault, I have the following structure:

```
concourse/main/defaults
concourse/defaults
```

I have the key `shared_secret` only in `concourse/defaults`.

My pipeline (test-shared-secrets) makes reference to `((defaults.shared_secret))` and fails when try to get this variable. When I create `shared_secret` in `/concourse/main/defaults` it works.

Checking API calls to Vault, I see concourse trying to fetch from `/concourse/main/test-shared-secrets/defaults` and also `/concourse/main/defaults` but never `/concourse/defaults` (as I was expecting).

## Steps to Reproduce

- Create a secret in `/concourse/defaults`, like shared_secret=geronimo
- Set a pipeline that refers to `((defaults.shared_secret))`
- Enable Vault and set the following variables:
-- CONCOURSE_VAULT_SHARED_PATH=defaults
-- CONCOURSE_VAULT_PATH_PREFIX=concourse
- Trigger the pipeline and you will see the failure

### Sample pipeline

```
---
resources:
- name: every-1m
  type: time
  source: {interval: 1m}

jobs:
- name: Echo
  plan:
  - get: every-1m
    trigger: true
  - task: testing-shared-secrets
    config:
      platform: linux
      image_resource:
        type: docker-image
        source: {repository: ubuntu}
      run:
        path: bash
        args:
          - -ec
          - |
            echo "my shared secret ((defaults.shared_secret))"
```

## Expected Results

I was expecting Concourse to find ((defaults.shared_secret)) in `/concourse/defaults/`

## Actual Results

Failed to interpolate task config: Finding variable 'defaults.shared_secret': missing field 'shared_secret' in var: defaults.shared_secret

## Additional Context

Now we use defaults per team but we want to use /concourse/defaults globally, taking precedence over /concourse/team/defaults.

## Version Info

* Concourse version: 5.8.0
* Infrastructure/IaaS: Kubernetes
* Browser (if applicable):
* Did this used to work? No

kirillbilchenko:
Signed-off-by: kirillbilchenko <kirya7@gmail.com>

# Existing Issue

Fixes #5253 

# Why do we need this PR?
In case if you vault secret key is a json object which  contains nested secrets, it's failing to lookup target secret across all available path. As example `somesecret.key` with current implementation will fail because the secret itself was found but nested search inside the json struct is not happening so there is no fallback to shared path, team path, if you have this secret in you pipeline but secret missed for specific key.

* if user specify with exactly secret he want to find in specific vault collection with dot separator, vault will try to find this secret in specified json  

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed

### DIFF
--- a/atc/creds/vault/vault.go
+++ b/atc/creds/vault/vault.go
@@ -62,8 +62,6 @@ func (v Vault) Get(secretPath string) (interface{}, *time.Time, bool, error) {
 		secretVal, found := secret.Data[secretName]
 		if found {
 			return secretVal, expiration, true, nil
-		} else {
-			return nil, nil, false, nil
 		}
 	}
 
--- a/atc/creds/vault/vault.go
+++ b/atc/creds/vault/vault.go
@@ -63,6 +63,7 @@ func (v Vault) Get(secretPath string) (interface{}, *time.Time, bool, error) {
 		if found {
 			return secretVal, expiration, true, nil
 		}
+		return nil, nil, false, nil
 	}
 
 	val, found := secret.Data["value"]
--- a/atc/creds/vault/vault.go
+++ b/atc/creds/vault/vault.go
@@ -2,6 +2,7 @@ package vault
 
 import (
 	"path"
+	"strings"
 	"time"
 
 	"github.com/concourse/concourse/atc/creds"
@@ -41,12 +42,29 @@ func (v Vault) NewSecretLookupPaths(teamName string, pipelineName string, allowR
 
 // Get retrieves the value and expiration of an individual secret
 func (v Vault) Get(secretPath string) (interface{}, *time.Time, bool, error) {
-	secret, expiration, found, err := v.findSecret(secretPath)
+	var secretName, secretRealPath string
+	if strings.Contains(secretPath, ".") {
+		parsed := strings.Split(secretPath, ".")
+		secretRealPath = parsed[0]
+		secretName = parsed[1]
+	} else {
+		secretRealPath = secretPath
+	}
+	secret, expiration, found, err := v.findSecret(secretRealPath)
 	if err != nil {
 		return nil, nil, false, err
 	}
 	if !found {
 		return nil, nil, false, nil
+	}
+
+	if len(secretName) != 0 {
+		secretVal, found := secret.Data[secretName]
+		if found {
+			return secretVal, expiration, true, nil
+		} else {
+			return nil, nil, false, nil
+		}
 	}
 
 	val, found := secret.Data["value"]
--- a/atc/creds/vault/vault.go
+++ b/atc/creds/vault/vault.go
@@ -42,14 +42,14 @@ func (v Vault) NewSecretLookupPaths(teamName string, pipelineName string, allowR
 
 // Get retrieves the value and expiration of an individual secret
 func (v Vault) Get(secretPath string) (interface{}, *time.Time, bool, error) {
-	var secretName, secretRealPath string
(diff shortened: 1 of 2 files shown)


atc/creds/vault/vault.go
```diff
@@ -41,14 +42,30 @@ func (v Vault) NewSecretLookupPaths(teamName string, pipelineName string, allowR
 
 // Get retrieves the value and expiration of an individual secret
 func (v Vault) Get(secretPath string) (interface{}, *time.Time, bool, error) {
-	secret, expiration, found, err := v.findSecret(secretPath)
+	var secretName string
+	secretRealPath := secretPath
+	if strings.Contains(secretPath, ".") {
```
vito: Hmm, I don't think this approach will work. Our `vars/` implementation is already supposed to handle `foo.bar` by splitting out the `foo`, so the `Get` call should only be called with `"foo"`. The extracting of fields off of the value is then applied to the secret that was returned by `Get` - the `secret.Data["value"]` is a special-case.

kirillbilchenko: actually for use as an example of secret we have json object with multiple secrets like:
{
git_token:"xxx"
webhook_token: "xxx"
concourse_webhook_pr: "xxx"
}
The problem with shared path which I triying to fix, if you have for example several defaults on different levels, like pipeline/team/root. And concourse is trying to find the key called default.concourse_webhook_pr in the vault, when it will lookup for default for example in the pipeline level it will find the secret, and will return as found, but when it will try to get  concourse_webhook_pr key from this map, it's failing with error  `failed to interpolate task config: Finding variable 'defaults.concourse_webhook_pr': missing field 'concourse_webhook_pr' in var: defaults.concourse_webhook_pr` and will not go through other available path. It still up to you if it suppose to work like this or not, but as for me in vault it's a pretty common case since secret.Data is returning map

vito: Yeah, I still don't really like that pattern; to me it's confusing for `((foo.bar))` to sometimes refer to `foo` in different locations. Utilizing fields in Vault is certainly common, but repeating the same definition name at multiple levels but with different sets of fields feels very error-prone and kind of unconventional. Wouldn't it be better to just explicitly name one `global_defaults` or something?

kirillbilchenko: the problem for me is like this, if you have a big organisation and you have team/pipeline/global level of defaults, for example git token per team can be in team defaults, and in the same time pipeline defaults namespace can contains pipeline specific defaults. With this level of lookups it's easier as for me to have some standardised pipelines definitions. And in the same time keep you vault name spaces more structured. And in scope of vault it will just try to find target key across all available folders, for me it's pretty much the same what's happening if you secret is looks like `{ value: "xxx"}`

kirillbilchenko:
Closing seems to work as expected.

PR #5270, opened 2020-03-04
The running state of a build that is a second rerun should show in the pipeline UI

### LINKED ISSUE #5236: Job has no next build when there are multiple build re-runs
## Bug Report

Jobs that have more than 1 build re-run are not rendered as "running" in the UI.

This doesn't seem to be a UI issue, however, since the API response for `/jobs` has `"next_build": null`

## Steps to Reproduce

1. Set a pipeline with any job
1. Run the job to completion (creates a `1` build)
1. Re-run the job to completion (creates a `1.1` build)
1. Re-run the job again (creates a `1.2` build)

## Expected Results

The UI displays the job as running and the `/jobs` endpoint includes a `next_build`

## Actual Results

`next_build` was `null`, so nothing was rendered in the UI

## Version Info

* Concourse version: master

clarafu:
There was a bug where the next build id of a job was not updated to the
correct build if the build was a rerun of a rerun. This was fixed by
having the rerun_of be compared to a non rerun build rather than the
latest completed build.

I also removed updating the next build id column on the Start method
because it should not be needed, since builds are added to the list of
possible "next build ids" when they are either pending or started and
since builds always go from pending -> started, we only need to update
the next build id when we create a new pending build (and then also
update it when we finish builds).

Fixes #5236

### DIFF
--- a/atc/db/build.go
+++ b/atc/db/build.go
@@ -340,13 +340,6 @@ func (b *build) Start(plan atc.Plan) (bool, error) {
 		return false, err
 	}
 
-	if b.jobID != 0 {
-		err = updateNextBuildForJob(tx, b.jobID)
-		if err != nil {
-			return false, err
-		}
-	}
-
 	err = tx.Commit()
 	if err != nil {
 		return false, err
@@ -514,12 +507,17 @@ func (b *build) Finish(status BuildStatus) error {
 			return err
 		}
 
-		err = updateLatestCompletedBuildForJob(tx, b.jobID)
+		latestNonRerunID, err := latestCompletedNonRerunBuild(tx, b.jobID)
 		if err != nil {
 			return err
 		}
 
-		err = updateNextBuildForJob(tx, b.jobID)
+		err = updateLatestCompletedBuildForJob(tx, b.jobID, latestNonRerunID)
+		if err != nil {
+			return err
+		}
+
+		err = updateNextBuildForJob(tx, b.jobID, latestNonRerunID)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


kcmannem:
LTGM

cirocosta:
Hey @clarafu 

> I also removed updating the next build id column on the Start method
because it should not be needed, since builds are added to the list of
possible "next build ids" when they are either pending or started and
since builds always go from pending -> started

I *think* this actually broke the way that `go-concourse` checks for "this
job has builds":

```
fly -t local builds -j test/test && fly -t local watch -j test/test
id   pipeline/job  build  status   start                     end  duration  team
147  test/test     1      started  2020-03-18@10:46:37-0400  n/a  2s+       main
error: job has no builds
```

clearly there's a build, but the detection says otherwise given the checks:

https://github.com/concourse/concourse/blob/4a46030a90d22e8b686b5d24ac745796a9751498/fly/       commands/helpers.go#L46-L52

that can indeed be verified by looking at the response that `jobs` is giving us
back:


```
2020/03/18 10:55:21 GET /api/v1/teams/main/pipelines/test/jobs/test HTTP/1.1
Host: localhost:8080
User-Agent: Go-http-client/1.1
Accept-Encoding: gzip


2020/03/18 10:55:21 HTTP/1.1 200 OK
Content-Length: 203
Content-Type: application/json
Date: Wed, 18 Mar 2020 14:55:21 GMT
Vary: Accept-Encoding
X-Concourse-Version: 0.0.0-dev
X-Download-Options: noopen
X-Frame-Options: deny

{  "id": 192,  "name": "test",
  "pipeline_name": "test",
  "team_name": "main",
  "next_build": null,
  "finished_build": null,
  "has_new_inputs": true,
  "inputs": [
    {
      "name": "test",
      "resource": "test",
      "trigger": true
    }
  ],
  "groups": null
}
```

is that really the case? I'm not very familiar with this part of the codebase :thinking: 

thanks!

clarafu:
@cirocosta Ah yeah good catch! Just to confirm with you, was that build created through fly execute? I think this might be because `CreateStartedBuild` does not update the next build ID? https://github.com/concourse/concourse/blob/4a46030a90d22e8b686b5d24ac745796a9751498/atc/db/pipeline.go#L957

cirocosta:
@clarafu, aaah actually, it wasn't - I was trying to debug why our SLIs are not doing well w/ regards to "create and run new pipeline" (which just sets the pipeline and waits for the pipeline to go from start to finish)

i.e., it was from `fly set-pipeline` and then waiting it to get going

clarafu:
@cirocosta Oh so it was from a regular build that was created by the scheduler?

cirocosta:
yesss

PR #5277, opened 2020-03-06
atc(feat): set_pipeline step to support var.

### LINKED ISSUE #5276: set_pipeline to support var

I'm building a pipeline that dynamically generates child pipelines and utilize the new `set_pipeline` step to submit child pipelines. But now `set_pipeline` can takes a static pipeline name, thus I'm blocked.

When I developed `set_pipeline` step, there was no `load_var` step. If we use var in pipeline name, then only way to pass value to var is via var_source or credential manager, which doesn't sound very useful.

As we 6.0 is going to introduce `load_var`, build a pipeline name dynamic by a previous step is very useful. Even the pipeline name can contain vars.

Yes

evanchaoli:
# Existing Issue

Fixes #5276  .

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/creds/set_pipeline_plan.go
+++ b/atc/creds/set_pipeline_plan.go
@@ -1,0 +1,28 @@
+package creds
+
+import (
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/vars"
+)
+
+type SetPipelinePlan struct {
+	variablesResolver vars.Variables
+	rawPlan           atc.SetPipelinePlan
+}
+
+func NewSetPipelinePlan(variables vars.Variables, plan atc.SetPipelinePlan) SetPipelinePlan {
+	return SetPipelinePlan{
+		variablesResolver: variables,
+		rawPlan:           plan,
+	}
+}
+
+func (s SetPipelinePlan) Evaluate() (atc.SetPipelinePlan, error) {
+	var plan atc.SetPipelinePlan
+	err := evaluate(s.variablesResolver, s.rawPlan, &plan)
+	if err != nil {
+		return atc.SetPipelinePlan{}, err
+	}
+
+	return plan, nil
+}
--- a/atc/creds/set_pipeline_plan.go
+++ b/atc/creds/set_pipeline_plan.go
@@ -19,10 +19,15 @@ func NewSetPipelinePlan(variables vars.Variables, plan atc.SetPipelinePlan) SetP
 
 func (s SetPipelinePlan) Evaluate() (atc.SetPipelinePlan, error) {
 	var plan atc.SetPipelinePlan
+
+	// Name should not be interpolated per #5277, thus backup name and restore
+	// after interpolation.
+	name := s.rawPlan.Name
 	err := evaluate(s.variablesResolver, s.rawPlan, &plan)
 	if err != nil {
 		return atc.SetPipelinePlan{}, err
 	}
+	plan.Name = name
 
 	return plan, nil
 }
--- a/atc/creds/set_pipeline_plan_test.go
+++ b/atc/creds/set_pipeline_plan_test.go
@@ -33,7 +33,7 @@ var _ = Describe("Evaluate", func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			Expect(result).To(Equal(atc.SetPipelinePlan{
-				Name:     "some-pn-is-ok",
+				Name:     "some-((pipeline-name))-ok", // Name should not be interpolated.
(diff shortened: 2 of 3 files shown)


vito:
This feels like something that should be addressed by pipeline instances (https://github.com/concourse/rfcs/pull/34) instead, since it's already on the horizon and provides a more generally useful approach. I've just proposed merging the RFC and will have a blog post announcing that tomorrow. 🙂

Until then I think it's best to just not do this; I don't think templating pipeline names is a design pattern we want to encourage, especially given likely conflicts in the future with concourse/rfcs#40. It'll be easier to just not add this than to rip it out later and break folks who relied on it. I would rather leave it as a known limitation while it's still experimental, if that's OK with you. I would suggest using the [`concourse-pipeline` resource](https://github.com/concourse/concourse-pipeline-resource) until then.

Thanks!

evanchaoli:
@vito But the `concourse-pipeline` resource has a bad restriction that it supports only local user authentication.

vito:
@evanchaoli OK, I can re-open, but I think we should add a warning if we detect that this functionality is used. Something like this:

```
WARNING: support for ((vars)) in the set_pipeline step name will be removed in a future release

the replacement for this pattern is being planned as RFC 34: https://github.com/concourse/rfcs/pull/34
```

evanchaoli:
@vito I just read through rfc#34, which would be a very useful feature, I left a comment there. But I think this feature is not conflict to rfc#34. This PR allows to modify pipeline name at runtime. Our use case is, there is a starter pipeline, it reads some input data, and generate sub-pipelines and dynamically set sub-pipelines, so we need to dynamically name sub-pipelines when set them. Different from instanced pipeline, in our case, sub-pipelines are not based on a template.

Even with instanced pipeline, this PR helps customize instanced pipeline name and `file`.

vito:
@evanchaoli Can you show an example of how you intend to use this?

vito:
I replied here which may clear things up: https://github.com/concourse/rfcs/pull/34#discussion_r421501216 - in short I can see why you'd want to parameterize `file:` but I still don't think we should allow parameterizing names. That's the old pattern that pipeline instances intends to replace. Paremeterizing names flawed because you have to carefully sanitize whatever values you're putting in there (see also concourse/rfcs#40), and in general it results in a cluttered dashboard with a ton of super long names. Instanced pipelines keep short names, allow instance vars to be used to identify the pipeline, de-clutter the UI, and keep related pipelines grouped together.

evanchaoli:
@vito From what you explained, sounds like the major reason to not support parameterized pipeline name is about UI display. In that case, can we set some restriction on pipeline name? Like pipeline name cannot be longer than X chars, then we can perform a check over interpolated pipeline name, if breaking the restriction, just fail the step. I think the restriction should also apply to static pipeline names, a crazy user may put a "descriptive" extremely long pipeline name and also clutter the UI.

As mentioned, instanced pipeline might not fit my use case (maybe fit, I need more thinking), I don't want to rework my pipeline in order to upgrade concourse to a version that contains instanced pipeline but deprecates this PR. So I don't want a temp solution.

vito:
@evanchaoli There's a bit more to it than just pipeline name length. The other problem is that if you configure a bunch of pipelines over time, the UI won't really help you keep them organized unless you keep running `fly order-pipelines --alphabetical` or something (which then doesn't let you maintain your own order).

You can literally just use pipeline instances as way to provide structured names for pipelines. There aren't any other special semantics behind it. They may share a common template, they may not; there's no difference from Concourse's perspective.

These two commands...:

```yaml
$ fly set-pipeline -p my-name-foo -c config-1.yml
$ fly set-pipeline -p my-name-bar -c config-2.yml
```

...and these two...:

```yaml
$ fly set-pipeline -p my-name --instance-var id=foo -c config-1.yml
$ fly set-pipeline -p my-name --instance-var id=bar -c config-2.yml
```

...both produce mechanically equivalent pipelines. The only difference is how they are identified, navigated, and shown in the UI. So from this perspective I'm having a hard time seeing how it wouldn't fit any particular use case. :)

evanchaoli:
@vito That makes sense. But I still have a confusing about automatically archiving un-configured instance, https://github.com/concourse/rfcs/pull/34/files#r421909756, once that confusion is addressed, I'm good to close this PR.

PR #5290, opened 2020-03-10
Switch back to original first occurrence behavior

### LINKED ISSUE #5274: Switch back to original first occurrence behavior
The new algorithm/scheduler in v6.0 changed the logic behind whether or not we will schedule a new build.

Prior to 6.0, we would only schedule a new build if the versions returned by the algorithm for triggerable resources have **never** been run by this job before. 

The new algorithm/scheduler was changed to schedule a new build if the versions returned for triggerable resource do not match the versions of the **previous** build. This change introduces a few problems that were pronounced by pinning and disabling of versions.

For example, if someone were to pin a resource to an old version and a build was created using that old version, when they unpin it all the jobs that rely on that resource would fire off a new build. This was because it would compare the now unpinned versions to the previous build that ran with the pinned versions, and since they do not match it will schedule a new build.

This feels pretty odd, which is why we added #5220 to have it so that only the job directly dependent on a resource (without any passed constraints) will be affected by pinning. This still doesn't completely solve the weirdness because this will mean that pinned resources do not directly affect jobs that use it as a passed constraint, and this is a very big difference to the current behaviour. This also applies to disabled versions, where the same situation of disabling and enabling a version will result in all the jobs dependent on the resources to all fire off. If we were to apply the same fix to disabling as for pinning, this will mean that disabled versions are not respected by jobs with passed constraints.. which this feels EVEN weirder.

So as a conclusion, we feel like we should back out from this change entirely and revert back to the old behavior, where we will only schedule a new build if the versions have never been run with this job before. We can further do more research on the proper behaviour that we want after we release the first beta 6.0, and have more time to slowly integrate it and get it right.

clarafu:
# Existing Issue

Fixes #5274.

# Why do we need this PR?
The scheduler will now schedule new builds only if all the versions
returned by the algorithm have not been used by any of the past builds.
Previously, we modified the behavior to have it compare only to be
previous build, but this proved to cause problems when combined with
pinned or disabled versions, so we backed out from this change.

First occurrence will behave the exact same way as before the new
algorithm was implemented.

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


atc/db/versions_db.go
```diff
@@ -32,6 +32,27 @@ func NewVersionsDB(conn Conn, limitRows int, cache *gocache.Cache) VersionsDB {
...
+		WITH builds_of_job AS (
+			SELECT id FROM builds WHERE job_id = $1
+		)
+		SELECT EXISTS (
+			SELECT 1
+			FROM build_resource_config_version_inputs i
+			JOIN builds_of_job b ON b.id = i.build_id
+			WHERE i.name = $2
+			AND i.version_md5 = $3
+		)`, jobID, inputName, versionMD5).
```
vito: In case anyone was following along as we discussed the potential need for adding a costly migration to add an index, we ended up not needing to add one; this query is fairly quick, much quicker than the one we originally replaced, and makes use of already existing indexes.

We also only schedule jobs when we *need* to schedule jobs, so even if this was kind of slow it wouldn't matter nearly as much as it did when we originally optimized this.

PR #5297, opened 2020-03-11
atc: create worker dir if not exist when using containerd

### LINKED ISSUE #5294: containerd runner does not create worker directory
## Bug Report

To run a worker using the Concourse binary CONCOURSE_WORK_DIR has to be provided. Garden runner creates the directory if it doesn't exist but containerd runner does not.

## Steps to Reproduce

Install and run the binary using CONCOURSE_GARDEN_USE_CONTAINERD set to "true".  

## Expected Results

Worker should come up normally.

## Actual Results

The following error shows up in the worker logs: `write default containerd config: ...... could not open file`

## Additional Context

We noticed this when we were working on https://github.com/concourse/ci/issues/284.

## Version Info

* Concourse version: 6.0.rc
* Did this used to work? No

xtremerui:
# Existing Issue
Fixes #5294

Create the worker dir if it does not exist, similar to what concoures does in https://github.com/concourse/concourse/blob/master/worker/workercmd/worker_linux.go#L108-L111

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/topgun/k8s/garden_config_test.go
+++ b/topgun/k8s/garden_config_test.go
@@ -7,6 +7,7 @@ import (
 	. "github.com/concourse/concourse/topgun"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
 )
 
 var _ = Describe("Garden Config", func() {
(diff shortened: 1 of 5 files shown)


taylorsilva:
Should we add a test (topgun/k8s maybe?) to ensure the CONCOURSE_WORK_DIR gets made when using the garden or containerd backends?

xtremerui:
@taylorsilva for now I dont see any test for containerd backend in topgun/k8s not sure if it is as easy as `passing a garden flag to use containerd` and test. I see the only tests for containerd is under https://github.com/concourse/concourse/blob/master/worker/backend/integration/integration_test.go

@cirocosta wdyt?

cirocosta:
@xtremerui yeaah, it should indeed be as simple as that 😁  

https://github.com/concourse/hush-house/blob/16424b7c3ecb1703f3c6b51c318b1f4c6160caa4/deployments/with-creds/ci-monitoring/values.yaml#L29-L33

xtremerui:
@cirocosta it might take more. Since if topgun test based on concourse/dev image, then the worker-dir is already defined and created https://github.com/concourse/ci/blob/1719a3d11bc6b70b44e97411849b27ab39410420/dockerfiles/dev/Dockerfile#L70-L72

unless there is a way to unset that environment var.

xtremerui:
@cirocosta @taylorsilva a topgun test is added as suggested. Please take a look. Thx!

xtremerui:
@cirocosta yes this was to fix the issue when deploying concourse with binary only. The k8s helm chart handles the dir already.

For the place of test, we have a similar test of atccmd here https://github.com/concourse/concourse/blob/master/atc/integration/configuration_test.go#L52-L62

Maybe we should have a `worker/integration`? I see there is also `worker/backend/integration` too.

taylorsilva:
@muntac and I created a new test suite in `worker/integration`.

The worker/integration test is now passing and fails when:

* you comment out the `os.MkDirAll()` in the `containerdRunner()`
* change the permission on the dir
* make a file instead of dir

To run the test we needed to be on linux. To test on linux while
developing on macOS we used docker by running the following commands:

```
$ docker run -it --entrypoint "/bin/bash" concourse/dev -v ~/workspace/concourse:/src
```

and then in the shell that it throws you in run:

```
$ cd /src
$ go test ./worker/integration
```

xtremerui:
@taylorsilva could you extract the instruction for running `worker/integration` into a readme file there? So developer would know how to run the suit without digging into the sea of issues of concourse.

taylorsilva:
Added a readme with steps on how to run the test on mac/windows using docker.

PR #5305, opened 2020-03-13
atc: simplify container gc

pivotal-jwinters:
This is PR is useful because it greatly simplifies our GC code (we no longer need to inject a worker provider). It also makes it more portable to other backends, since we no longer rely on a garden specific gracetime to clean up the containers behind the scenes.

It updates the way hijacked containers are gc'ed. Hijacking a container will now set its last_hijack timestamp in the database, and continuously update this value every minute until the connection to the container is closed. Containers will only get gc'ed if they haven't been hijacked in the
last configurable interval (default 10m).


pivotal-jwinters:
The other strategy I considered was reference counting hijack connections and only gc'ing the container when it has no more connected clients. The problem with this approach is that we could potentially get into state where if the atc dies mid connection we would never decrement the counter and the container would stay around forever.

cirocosta:
niiiiice!

I was trying to understand why we have that `discontinued` field, and found few
things:

1. we don't use `IsDiscontinued` anywhere!

```console
$ ag -Q "IsDiscontinued"
atc/db/container.go
284:    IsDiscontinued() bool
322:func (container *destroyingContainer) IsDiscontinued() bool { return container.isDiscontinued }
```

2. it seems like using `Discontinue()` there in `container_collector`, which not
only transitions the container to `destroying`, but also sets `discontinued` to
`true`, makes the whole sweeping to not occur:

https://github.com/concourse/concourse/blob/0f8503de3cb0948a19ba37e6752b182c3e6d5c35/worker/container_sweeper.go#L76-L77

which then ultimately calls

https://github.com/concourse/concourse/blob/0f8503de3cb0948a19ba37e6752b182c3e6d5c35/atc/api/containerserver/list_destroying_containers.go#L21

which then filters by those with `discontinued == false`:


https://github.com/concourse/concourse/blob/0f8503de3cb0948a19ba37e6752b182c3e6d5c35/atc/db/container_repository.go#L143-L156

making `contianer_sweeper` not sweep those that we're discontinuing
:sweat_smile:

from my understanding, we could just drop the `discontinued` field altogether.

<br />

please let  me know if I got it all wrong!

thankss!

cirocosta:
oh, also,

I noticed that in our docs we don't mention "for how long you can stay within
that hijacked container" - do you think we should have that there?

https://concourse-ci.org/builds.html#fly-intercept

kcmannem:
woah, so clean

pivotal-jwinters:
@cirocosta good catch on the discontinued state. Definitely looks like we don't need it, I'll remove it.

aoldershaw:
@cirocosta @pivotal-jwinters sorry, I missed your comment on Friday. IIRC the previous behaviour was that once a container was hijacked, destroying the container was left up to Garden (which it would only do 5 minutes after the hijack session ended). This was the reason for having `discontinued` - to indicate to the container sweeper that it's not responsible for cleaning it up

pivotal-jwinters:
@cirocosta @aoldershaw yeah garden was smart about its gracetime, but I don't think we need to be. I don't imagine we are going to have that many hijacked containers lying around, and 1hr seems like plenty of time to do what you need to do.

cirocosta:
thx for following up w/ the changes! lgtm!

I think now we're just missing release notes / possibly documenting more
explicitly this behavior in our docs.

---

> bomberman [...] is only unpaused after /bin/bash (or whatever command is run)
> exits.

so it seems like there's indeed a change in behavior here - what would
previously in theory allow you to stay hijacked until the worker would go away,
is not true anymore: we're putting a hard limit on "gc hijack grace period" +
time for the worker to sweep the container.

I agree that for most cases this is just fine, but I wonder if this could break
those exceptional cases where people relied on the fact that a hijack session
could stay for "as long as the worker exists".

I'm leaning towards just making sure that we call this out very explicitly in
the release notes, wdyt @pivotal-jwinters? cc @vito

pivotal-jwinters:
> I agree that for most cases this is just fine, but I wonder if this could break
those exceptional cases where people relied on the fact that a hijack session
could stay for "as long as the worker exists".

We've never had any guarantees (in the docs) around the lifecycle of hijacked containers, and that seems like a fairly obscure undocumented behaviour to be reliant upon. So yeah I agree that documenting the new behaviour should be sufficient.

cirocosta:
cooool! would you mind following up w/ the release note then? thxx!

pivotal-jwinters:
I'm going to hold off until @vito gives this the a-ok. I still have this nagging feeling in the back of my mind that we should still do some sort of reference counting + a configureable container gracetime to catch the cases where we fail to decrement the counter. But then I tell myself that it's not really necessary, and this simpler approach is better.

vito:
Awesome seeing that worker job runner code finally die!

My one concern is the same as @cirocosta pointed out, which is that this approach gives users 1 hour to figure out what's wrong in their container before it gets GCed, whereas previously we would keep it alive as there was an active connection.

While this behavior wasn't documented, it most definitely was intended. If someone's in the thick of debugging and it's taking a while - or maybe they went on a lunch break - having the container swept away from under them would be really frustrating, as it could take a long time before that particular failure can be reproduced.

Reference-counting would work but yeah, I'd have the same concerns over failing to decrement the counter. Maybe we could keep updating the 'last hijacked' value as long as the hijack endpoint is active?

@kcmannem Volumes won't always be enough to debug since there may be running processes to troubleshoot in the container.

pivotal-jwinters:
> Maybe we could keep updating the 'last hijacked' value as long as the hijack endpoint is active?

Oh nice! Yeah i'll try that.

pivotal-jwinters:
@vito @cirocosta 

The new approach is to heartbeat the last_hijack value at a 1m interval, and gc the container after 10m of inactivity.

PR #5350, opened 2020-03-23
5.2.x open redirect

xtremerui:
cherry picked from 5.8.x to escaped path in redirect URL


vito:
I'm kind of confused as to what this is and how it relates to https://github.com/concourse/concourse/pull/5337 :thinking:

xtremerui:
@vito unit job failed by the dependency `benchmark` of web is not available. That has been changed after 5.2.x.

PR #5371, opened 2020-03-27
fly: fix curl command when passing options

### LINKED ISSUE #5366: fly curl -- does not pass options to curl
## Bug Report

Running

```
fly curl <url_path> -- <curl_options>
```

does not seem to pass <curl_options> to the underlying curl command as documented here:

https://concourse-ci.org/administration.html#fly-curl

## Steps to Reproduce

```
fly -t ci curl /api/v1/teams/main/cc.xml -- -vvi
<Projects><Project activity="Sleeping" lastBuildLabel="1" lastBuildStatus="Success" lastBuildTime="2020-03-25T13:28:53Z" name="pytools/build" webUrl="http://localhost:8080/teams/main/pipelines/pytools/jobs/build"></Project><Project activity="Sleeping" lastBuildLabel="3" lastBuildStatus="Success" lastBuildTime="2020-03-25T21:52:05Z" name="bash-tools/build" webUrl="http://localhost:8080/teams/main/pipelines/bash-tools/jobs/build"></Project></Projects>
```

## Expected Results

You'd expect to see significant curl headers, such as:

```
...
> GET /api/v1/teams/main/cc.xml HTTP/1.1
> Host: localhost:8080
> User-Agent: curl/7.54.0
...
```

## Actual Results

Only the XML output is shown implying the curl command never received the `-vvi` options from the fly command, unless the fly command is somehow sanitising and removing the extra curl output they generate??

I realize I could work around via:

```
fly -t ci curl /api/v1/teams/main/cc.xml --print-and-exit
```

but it's not clear to me whether this is a bug in the `--` behaviour supposedly passing raw options to curl.

Update: the below command triggers curl's pipeline output to stderr which doesn't get silenced by the curl `-s` switch appended after `--` implying that it is not an output filtering issue but rather the `-s` switch never arriving on the curl command line:

```
fly -t ci curl /api/v1/teams/main/cc.xml -- -s | cat
```

## Version Info

* Concourse version: 5.8.0
* Deployment type: Docker

xtremerui:
# Existing Issue
Fixes #5366 

# Why do we need this PR?
#4048 introduced a `Query` arg that break passing extra curl options when doing `fly curl -- -some-option`

Parse the `Path` argument so it won't encode `?` to produce a wrong URL as mention in #4048 and thus allowing passing extra curl options back to normal 

# Contributor Checklist

# Reviewer Checklist
- [x] Code reviewed
- [x] Tests reviewed
- [ ] Release notes reviewed
- [x] PR acceptance performed

### DIFF
--- a/fly/commands/curl.go
+++ b/fly/commands/curl.go
@@ -12,9 +12,8 @@ import (
 
 type CurlCommand struct {
 	Args struct {
-		Path  string   `positional-arg-name:"PATH" required:"true"`
-		Query string   `positional-arg-name:"QUERY"`
-		Rest  []string `positional-arg-name:"curl flags" description:"To pass flags to curl, pass a -- argument, so that fly can distinguish them from its own flags"`
+		Path string   `positional-arg-name:"PATH" required:"true"`
+		Rest []string `positional-arg-name:"curl flags" description:"To pass flags to curl, pass a -- argument, so that fly can distinguish them from its own flags"`
 	} `positional-args:"yes"`
 	PrintAndExit bool `long:"print-and-exit" description:"Print curl command and exit"`
 }
@@ -30,7 +29,7 @@ func (command *CurlCommand) Execute([]string) error {
 		return err
 	}
 
-	fullUrl, err := command.makeFullUrl(target.URL(), command.Args.Path, command.Args.Query)
+	fullUrl, err := command.makeFullUrl(target.URL(), command.Args.Path)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pivotal-jamie-klassen:
I'll give a more thorough acceptance narrative, because you never know what we'll need to know about years from now.

I tried `fly -t dev curl /api/v1/jobs -- -s | grep foo` and was very pleased to _not_ see the usual curl output. For reference, that noise looks like this:
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   215  100   215    0     0  12002      0 --:--:-- --:--:-- --:--:-- 12647
```

The other half of this work is that I run `fly -t dev curl /api/v1/teams/main/pipelines/pipeline/resources/resource/versions?filter=version:version` to filter versions, rather than `fly -t dev curl /api/v1/teams/main/pipelines/pipeline/resources/resource/versions filter=version:version`, and I think this is an improvement, too. Although I think the fish shell is touchy about question marks, so you might have to quote the path/query part if you're using fish.

xtremerui:
Thank you for the acceptance narrative! PR is updated as required.

xtremerui:
Thank you for fast response!

PR #5367, opened 2020-03-26
fly: behaviour: add archive-pipeline command

aoldershaw:
# Existing Issue

branched of #5365 which should be merged first.
Fixes #5320 .

# Changes proposed in this pull request

* Add fly archive-pipeline
* fly archive-pipeline supports a single pipeline (`-p`) or all pipelines in a team (`--all`)
* fly archive-pipeline has a confirmation prompt before archiving any pipeline, overridable using `-n`

# Contributor Checklist
- [x] ~Unit tests~
- [x] Integration tests (if applicable)
- [ ] Updated documentation (located at https://github.com/concourse/docs)
- [ ] Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [ ] ~New config flags added? Ensure that they are added to the [BOSH](https://github.com/concourse/concourse-bosh-release) and [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for the [integration tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env) (for example, if they are Garden configs that are not displayed in the `--help` text).~

PR #5365, opened 2020-03-25
go-concourse: Add Archive to go-concourse

YoussB:
# Existing Issue
branched of #5359 which should be merged first.
Fixes #5319 

# Why do we need this PR?
Continuing the road to Archiving pipelines, this PR adds the archive functionality to go-concourse.

# Contributor Checklist

- [x] Unit tests
- [x] ~Integration tests (if applicable)~
- [x] ~Updated documentation (located at https://github.com/concourse/docs)~
- [x] ~Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)~

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [ ] New config flags added? Ensure that they are added to the [BOSH](https://github.com/concourse/concourse-bosh-release) 
      and [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for the [integration tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env) (for example, if they are Garden configs that are not displayed in the `--help` text).

PR #5359, opened 2020-03-24
atc: behaviour: archiving a pipeline also pauses it

aoldershaw:
Existing Issue

Branched off of #5352; it should be merged first
Fixes #5318 .

# Changes proposed in this pull request

* `db.Pipeline.Archive()` sets both "archived" and "paused" to true
* Setting an archived pipeline will unarchive it, but it will remain paused (see #5352)
* Note that you cannot get in a state where `archived == true` but `paused == false`, since #5351 prohibits unpausing an archived pipeline

# Contributor Checklist
- [x] Unit tests
- [x] Integration tests (if applicable)
- [ ] Updated documentation (located at https://github.com/concourse/docs)
- [ ] Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)

# Reviewer Checklist
> This section is intended for the core maintainers only, to track review progress. Please do not
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [x] ~New config flags added? Ensure that they are added to the [BOSH](https://github.com/concourse/concourse-bosh-release) and [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for the [integration tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env) (for example, if they are Garden configs that are not displayed in the `--help` text).~

PR #5352, opened 2020-03-23
atc: behaviour: setting a pipeline unarchives it

aoldershaw:
# Existing Issue

Branched off of #5351; it should be merged first
Fixes #5317 .

# Changes proposed in this pull request

* Updating an existing pipeline's config sets the `archived` field to `false`

Note: this does not currently work with `fly set-pipeline` if the pipeline config doesn't change - only through `go-concourse`/the API directly.

# Contributor Checklist
- [x] Unit tests
- [x] Integration tests (if applicable)
- [ ] Updated documentation (located at https://github.com/concourse/docs)
- [ ] Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [x] ~New config flags added? Ensure that they are added to the [BOSH](https://github.com/concourse/concourse-bosh-release) and [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for the [integration tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env) (for example, if they are Garden configs that are not displayed in the `--help` text).~

PR #5387, opened 2020-03-31
core/fly: add archive pipeline endpoint, backend functions, and fly command

### LINKED ISSUE #5368: Archived pipelines should not show up in `fly pipelines`
**Epic:** #5055

aoldershaw:
# Existing Issue

Fixes #5368

This PR combines the commits from the following PR's (which we closed): #5346, #5351, #5352, #5359, #5367, #5365

* Adds an `archived` column to the `pipelines` table
  * the `ArchivePipeline` endpoint strictly updates this field to `true`
  * access is limited to team owners. 
* unpausing an archived pipeline fails through the API, you cannot get in a state where `archived == true` but `paused == false`
* Updating an existing pipeline's config sets the archived field to false
* archiving a pipeline also pauses it
* Setting an archived pipeline will unarchive it, but it will remain paused
* adds `Archive()` to go-concourse
* adds an `archive-pipeline` command to fly
* Filter pipeline list to only include non-archived pipelines

Note: we'll eventually want another flag on `fly pipelines` to include archived pipelines, but this PR doesn't deal with that.

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
- [x] Code reviewed
- [x] Tests reviewed
- [x] Documentation reviewed
- [x] Release notes reviewed
- [x] PR acceptance performed


fly/commands/pipelines.go
```diff
@@ -28,19 +28,26 @@ func (command *PipelinesCommand) Execute([]string) error {
...
-		pipelines, err = target.Team().ListPipelines()
+		unfilteredPipelines, err = target.Team().ListPipelines()
 		headers = []string{"name", "paused", "public", "last updated"}
 	}
 	if err != nil {
 		return err
 	}
 
+	var pipelines []atc.Pipeline
+	for _, p := range unfilteredPipelines {
```
cirocosta, line 45: It's interesting how this creates a difference between the json that comes in from the API and the json that's displayed through `displayhelpers.JsonPrint`:

```console
$ fly -t local pipelines --json
null
```

vs

```console
$ fly -t local pipelines --verbose

GET /api/v1/teams/main/pipelines HTTP/1.1
HTTP/1.1 200 OK <...>
[
  {
    "id": 1,
    "name": "test",
    "paused": true,
    "public": false,
    "archived": true,
    "team_name": "main",
    "last_updated": 1585749902
  }
]
```

while I do understand that the purpose of `--json` is to literally just change the way that the information we want to show is displayed (instead of passing the response through), but it made me wonder if we should use a query parameter to limit the API's response in the first place :thinking:  have you thought about that? not sure how that'd affect future plans for the dashboard/fly. 

thanks!

aoldershaw: It did feel weird to me making `--json` differ from the API response, but it felt even weirder to get different data when specifying `--json` compared to the table view. If you have differing thoughts there I'd be curious to hear them - @jomsie and I were struggling a bit with getting the UX to feel right.

In terms of pagination, that definitely is something we've considered. @pivotal-jamie-klassen added a "tech debt" note [here](https://project.concourse-ci.org/projects/MDc6UHJvamVjdDM4Mzc3ODI=) to paginate all the "dashboard" APIs. I feel it would probably need to become its own epic since it'll likely require a fair bit of work to maintain feature parity with what we have now (e.g. filtering the dashboard becomes more difficult). It would be nice for the dashboard to refresh only the 200 jobs that are visible every 5 seconds, rather than the 30,000 jobs in the entire cluster.

cirocosta: >  If you have differing thoughts there I'd be curious to hear them 

tbh, I think it's great as is right now - I've always seen `json` as just a different view of the default `table`, which I think makes sense to have the behavior of "show the very same data but in a different way"

> In terms of pagination,

oh, sorry, I was actually referring to something like `?archived=false` (so that `archived:true` pipelines wouldn't even reach the client) or something like that (more like filtering, rather than paginating)

*(even though paginating would be great too!)*

aoldershaw: Oh haha sorry, I misunderstood.

The [RFC](https://github.com/concourse/rfcs/pull/33/files#diff-36359b59bfd438f0b68f9be6fefac7d0R28) specifies:

> `ListAllPipelines`, `ListPipelines`, and `GetPipeline` will continue to return archived pipelines. Hiding archived pipelines is the job of the web UI and `fly`, not the API.

Now, that's not to say having a query parameter on `ListAllPipelines` and `ListPipelines` wouldn't also make sense...

aoldershaw: PS I fixed `--json` to output `[]` when there are no non-archived pipelines rather than `null`

YoussB:
@cirocosta @vito this pr is ready for review. @taylorsilva and I have closed the other archiving prs for the sake of easier reviewing and merging.

vito:
FYI: noticed some compilation failures in the unit tests from the rebase so I fixed and force-pushed myself. The net change was just `fakeaccess` -> `fakeAccess` in `atc/api/pipelines_test.go`. I amended the commits that introduced the tests instead of adding additional commits.

YoussB:
All tests are passing now, just needs a review

atc/db/pipeline.go
```diff
@@ -621,6 +631,45 @@ func (p *pipeline) Unpause() error {
...
+			"id": p.id,
+		}).
+		RunWith(tx).
+		Exec()
+
+	if err != nil {
+		return err
+	}
+
+	err = p.clearConfigForJobsInPipeline(tx)
```
cirocosta, line 655: Might be important to point out in the notes that those who go through a down migration will end up with pipelines that, if unpaused, won't really work until the `set-pipeline` as failures like

```
web_1     | {"timestamp":"2020-04-09T15:29:11.132091560Z","level":"info","source":"atc","message":"atc.checker.errored","data":{"check":2,"error":"run check step: check: resource script '/opt/resource/check []' failed: exit status 1\n\nstderr:\ninvalid configuration: must configure either 'interval' or 'start' and 'stop'\n","session":"18"}}
```

would occur :thinking:  wdyt?

cirocosta: lol, after doing a down migration I just realized that a pipeline without resources like this actually works

```
jobs:
- name: ""
  plan: null
```

(you can get a successful build :thinking: not sure if the `name: ""` is really something we should accept when scheduling the build, @vito's https://github.com/concourse/concourse/issues/3985 would go against it)

---

if you're curious about the events of such build:

```
fly -t local curl /api/v1/builds/3/events
id: 0
event: event
data: {"data":{"status":"started","time":1586446873},"event":"status","version":"1.0"}

id: 1
event: event
data: {"data":{"status":"succeeded","time":1586446893},"event":"status","version":"1.0"}

id: 2
event: end
data
```

cirocosta:
@taylorsilva and I here:

for `2` from https://github.com/concourse/concourse/pull/5387#pullrequestreview-390894906, given that we can fix that with a refactor in go-concourse with no changes need in in `fly/commands`, we're going to leave this to a subsequent PR in the context of refactoring go-concourse.

atc/db/pipeline.go
cirocosta: https://github.com/concourse/concourse/pull/5387/commits/6ea4c55ba0fb56520454a76ec90f8e94ad56a4d0

cirocosta:
with regards to `3` from https://github.com/concourse/concourse/pull/5387#pullrequestreview-390894906, given that `pause-pipeline` behaves the same way, we'll leave it as is.

taylorsilva:
Once again, this PR is ready and...
bernie

release-notes/latest.md
```diff
@@ -40,3 +40,14 @@
 #### <sub><sup><a name="5410" href="#5410">:link:</a></sup></sub> feature
 
 * We've moved the "pin comment" field in the Resource view to the top of the page (next to the currently pinned version). The comment can be edited inline.
+
+#### <sub><sup><a name="5368" href="#5368">:link:</a></sup></sub> feature
+
+* Implented the core functionality for archiving pipelines [RFC #33].
```
zoetian, line 46: Implented -> Implemented

taylorsilva:
@zoetian we addressed [1] and [2] with our latest commit.

Regarding [3] there's an open issue to resolve that #5390

fly/commands/archive_pipeline.go
```diff
@@ -99,6 +100,8 @@ func (command ArchivePipelineCommand) confirmArchive(pipelines []string) bool {
 		command.printPipelinesTable(pipelines)
 	}
 
+	fmt.Printf("!!! archiving the pipeline will remove its configuration. Build history will be retained.\n\n")
```
zoetian, line 103: ! ! ! message received 👮

atc/api/pipelineserver/archive_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/concourse/concourse/atc/db/dbfakes"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+//go:generate counterfeiter code.cloudfoundry.org/lager.Logger
+
+var _ = Describe("Archive Handler", func() {
+	var (
+		fakeLogger *pipelineserverfakes.FakeLogger
```
pivotal-jamie-klassen, line 20:
```suggestion
		fakeLogger lagertest.TestLogger
```

atc/api/pipelineserver/archive_test.go
```diff
@@ -0,0 +1,91 @@
...
+		Expect(action).To(Equal("archive-pipeline"), "wrong action name")
+		Expect(actualError).To(Equal(expectedError))
+	})
+
+	It("write a debug log on every request", func() {
+		handler.ServeHTTP(recorder, request)
+
+		Expect(fakeLogger.DebugCallCount()).To(Equal(1))
+		action, _ := fakeLogger.DebugArgsForCall(0)
+		Expect(action).To(Equal("archive-pipeline"), "wrong action name")
```
pivotal-jamie-klassen, line 60:
```suggestion
		Expect(fakeLogger.Logs()).To(ConsistOf(
			gstruct.MatchFields(gstruct.IgnoreExtras, gstruct.Fields{
				"Message": "test.archive-pipeline",
				"LogLevel": lager.DEBUG,
			}),
		))
```

atc/api/pipelineserver/archive_test.go
```diff
@@ -0,0 +1,91 @@
...
+		fakeLogger *pipelineserverfakes.FakeLogger
+		server     *pipelineserver.Server
+		dbPipeline *dbfakes.FakePipeline
+		handler    http.Handler
+		recorder   *httptest.ResponseRecorder
+		request    *http.Request
+	)
+
+	BeforeEach(func() {
+		fakeLogger = new(pipelineserverfakes.FakeLogger)
```
pivotal-jamie-klassen, line 29:
```suggestion
		fakeLogger = lagertest.NewTestLogger("test")
```

taylorsilva:
@pivotal-jamie-klassen switched over the fakelogger from lager!

atc/api/pipelineserver/archive_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"code.cloudfoundry.org/lager"
+	"code.cloudfoundry.org/lager/lagertest"
+	"github.com/concourse/concourse/atc/api/pipelineserver"
+	"github.com/concourse/concourse/atc/db/dbfakes"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gstruct"
+)
+
+//go:generate counterfeiter code.cloudfoundry.org/lager.Logger
```
pivotal-jamie-klassen, line 18:
```suggestion
```

atc/api/pipelineserver/pipelineserverfakes/fake_logger.go
```diff
@@ -0,0 +1,451 @@
...
+	if fake.invocations == nil {
+		fake.invocations = map[string][][]interface{}{}
+	}
+	if fake.invocations[key] == nil {
+		fake.invocations[key] = [][]interface{}{}
+	}
+	fake.invocations[key] = append(fake.invocations[key], args)
+}
+
+var _ lager.Logger = new(FakeLogger)
```
pivotal-jamie-klassen, line 451:
```suggestion
```

pivotal-jamie-klassen, line 451:
```suggestion
```

xtreme-vikram-yadav:
Looks like the use case for erroring on set_pipeline step when the new pipeline has the same name as the archived pipeline is used is not yet implemented. Is this known or should we create an issue to address this?

chenbh:
^
> If the set_pipeline step notices that it's configuring an archived pipeline configured by a different job, or by no job, it will fail.

vito:
@xtreme-vikram-yadav Hmm sorry I don't follow - maybe open an issue and we'll hash it out there? (Just don't want to lose track of the conversation since this PR is already merged.)

PR #5392, opened 2020-04-01
web: behaviour: fix scroll on build history list

### LINKED ISSUE #5383: Scrolling build numbers list is broken
## Bug Report

Scrolling through the list of builds does not work - it always jumps to the most recent build

## Steps to Reproduce

1. Open a job with many builds
2. Use your mouse wheel to scroll

## Expected Results

I can scroll to older builds

## Actual Results

It fights me, and jumps back to the start of the list

## Version Info

* Concourse version: 6.0
* Deployment type (BOSH/Docker/binary): N/A
* Infrastructure/IaaS: N/A
* Did this used to work? Yes

aoldershaw:
# Existing Issue

Fixes #5383 .

* Introduce `scrollCoordsSimple` function for dealing with simple cases of computing scroll position (that don't require knowing the coordinates of the source and parent elements)
* Replace the deprecated DOM event `mousewheel` with `wheel`, which fixes scrolling on Firefox (which seems to have been broken for a while)

Note that this behaviour could be shoehorned in to the more complicated `scrollCoords` function, but there is a noticeable performance hit in scrolling due to the increased number of calls out to `Elm.Browser`.

# Contributor Checklist

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/web/elm/src/Build/Header/Header.elm
+++ b/web/elm/src/Build/Header/Header.elm
@@ -33,6 +33,7 @@ import Message.Subscription
         , Subscription(..)
         )
 import Routes
+import StrictEvents exposing (DeltaMode(..))
 import Time
 
 
@@ -394,12 +395,23 @@ update msg ( model, effects ) =
     case msg of
         ScrollBuilds event ->
             let
+                scrollFactor =
+                    case event.deltaMode of
+                        DeltaModePixel ->
+                            1
+
+                        DeltaModeLine ->
+                            20
+
+                        DeltaModePage ->
+                            800
+
                 scroll =
                     if event.deltaX == 0 then
-                        [ Scroll (Sideways event.deltaY) historyId ]
+                        [ Scroll (Sideways <| event.deltaY * scrollFactor) historyId ]
 
                     else
-                        [ Scroll (Sideways -event.deltaX) historyId ]
+                        [ Scroll (Sideways <| -event.deltaX * scrollFactor) historyId ]
 
                 checkVisibility =
                     case model.history |> List.Extra.last of
--- a/web/elm/src/Message/Effects.elm
+++ b/web/elm/src/Message/Effects.elm
@@ -11,7 +11,7 @@ import Api
 import Api.Endpoints as Endpoints
 import Assets
 import Base64
-import Browser.Dom exposing (Element, getElement, getViewport, getViewportOf, setViewportOf)
+import Browser.Dom exposing (Element, Viewport, getElement, getViewport, getViewportOf, setViewportOf)
 import Browser.Navigation as Navigation
 import Concourse exposing (encodeJob, encodePipeline, encodeTeam)
 import Concourse.BuildStatus exposing (BuildStatus)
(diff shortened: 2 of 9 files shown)


cirocosta:
Hey @aoldershaw ,

I was able to get scrolling working on Chrome, but not on Firefox :thinking:   - do you think it'd be hard to adapt for it too? the fix you provided doesn't seem to deal with anything that's platform dependent, which makes me wonder if it even worked before on Firefox at all :thinking:  

thanks!

aoldershaw:
Hey @cirocosta, thanks for testing in multiple browsers! Weird that it doesn't work in Firefox - this is essentially just reverting the scrolling behaviour to 495904da500188d89a9c9e62071d4ac0e3205ed7.

It also seems to not be the same scrolling issue as I set out to fix - on Chrome, you could scroll horizontally for a split second before having it forced back to the latest build, whereas on Firefox, it doesn't seem to move at all.

I'll take a look!

aoldershaw:
@cirocosta turns out we were using a deprecated event `onmousewheel` when we should have been using `onwheel` - https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event.

Seems to work on Chrome and Firefox now

jomsie:
scrolling looks good on both chrome/safari (on osx), but firefox (on osx) is very slow scrolling... takes a whole scroll to go back 1 or 1.5 builds. (might be a firefox issue?)

still, slow scrolling beats no scrolling at all!

aoldershaw:
> scrolling looks good on both chrome/safari (on osx), but firefox (on osx) is very slow scrolling... takes a whole scroll to go back 1 or 1.5 builds. (might be a firefox issue?)

Huh, weird...I hadn't noticed since I was just using my trackpad, but I am experiencing the same with a mouse. The `deltaY` values seem to be on completely different scales between Chrome and Firefox...not sure how best to deal with that

jomsie:
> > scrolling looks good on both chrome/safari (on osx), but firefox (on osx) is very slow scrolling... takes a whole scroll to go back 1 or 1.5 builds. (might be a firefox issue?)
> 
> Huh, weird...I hadn't noticed since I was just using my trackpad, but I am experiencing the same with a mouse. The `deltaY` values seem to be on completely different scales between Chrome and Firefox...not sure how best to deal with that. We could try just checking the sign of `deltaY` and scale it to a fixed value, but that could be annoying trying to scroll a small amount.
> 
> EDIT: ah, it may be the case that the `deltaMode` differs (Firefox is measuring in terms of lines, rather than pixels - https://stackoverflow.com/questions/20110224/what-is-the-height-of-a-line-in-a-wheel-event-deltamode-dom-delta-line). I'll take a look

👀 look tomorrow

aoldershaw:
Okay @jomsie, I've added some handling for `deltaMode`, which makes it feel more natural on Firefox. Let me know what you think!

jomsie:
i'll review this after lunch!!!

web/elm/src/Message/Effects.elm
```diff
@@ -657,19 +657,19 @@ scroll : ScrollDirection -> String -> Cmd Callback
...
-            scrollCoords id id (always 0) (.srcElem >> .viewport >> .y >> (+) -60)
+            scrollCoordsSimple id (always 0) (.viewport >> .y >> (+) -60)
 
         ToBottom ->
-            scrollCoords id id (always 0) (.parentElem >> .scene >> .height)
+            scrollCoordsSimple id (always 0) (.scene >> .height)
 
         Sideways delta ->
-            scrollCoords id id (.srcElem >> .viewport >> .x >> (+) -delta) (always 0)
+            scrollCoordsSimple id (.viewport >> .x >> (+) -delta) (always 0)
```
jomsie, line 672: the things firefox makes you do 😆

web/elm/src/StrictEvents.elm
```diff
@@ -28,6 +30,12 @@ type alias ScrollState =
     }
 
 
+type DeltaMode
```
jomsie, line 33: the things firefox makes you do 😆

jomsie:
Just some evidence! 

safari:
safari

chrome:
chrome

firefox:
firefox

web/elm/src/Build/Header/Header.elm
```diff
@@ -394,12 +395,23 @@ update msg ( model, effects ) =
     case msg of
         ScrollBuilds event ->
             let
+                scrollFactor =
+                    case event.deltaMode of
+                        DeltaModePixel ->
```
jomsie, line 400: Just a side comment, not super important, how'd you get the 20 / 800 for line / page?

aoldershaw: Totally fair question! Tried to give some justification for it in the commit message I added that in.

tl;dr 20 for Line is based on the font-size of the scroll bar. The 800 for Page is totally arbitrary - I took it from some Facebook library. I haven’t been able to produce a DeltaModePage event, so not sure if it feels natural, but it’s better than nothing!

Issue #5195, opened 2020-02-18
Make `zstd` compression opt-in; default to `gzip`


vito:
Supporting `zstd` has proven to be a little tricky. It's more performant, but the `cgo` package we use appears to not work with certain payloads (#4621), and switching to a pure Go implementation caused more problems which need to be looked into (#5183).

In hindsight we probably should have made this feature-flagged, so let's go back to `gzip` as the default for now and make `zstd` opt-in.

In making `zstd` opt-in let's un-revert the pure Go changes; I think that's what we want in the long run, but we need to investigate why it's not working.

kcmannem:
myabe use gzip as the fallback when zstd fails 🤷‍♂

xtremerui:
With going by making zstd opt-in, `fly` now needs a way to tell which one of gzip/zstd that concourse server is running with.

One option is making the compression info available by `/api/v1/info`. Or fly could just try using `gzip` first and fallback to zstd.

@vito wdyt?

vito:
@xtremerui I think we can have `fly` just use `gzip` by default. My understanding is that the biggest win with `zstd` is for when we have to stream between workers, i.e. in production pipelines - I don't think we really care as much about `fly execute`. 🤔 If it's "free" we could add a `--compression=gzip|zstd` flag?

cirocosta:
Hey,

I was trying to compare zstd (the datadog one, cgo-based) vs gozstd (from klauspost), here are some interesting stats from the cgroup where those baggageclaim servers ran:

![Screenshot from 2020-04-09 18-11-51](https://user-images.githubusercontent.com/3574444/78945240-d2924480-7a8d-11ea-939f-dff62e11226b.png)

first three (<<) are `zstd`, and last three (>>) are gozstd.

The test is essentially getting a baggageclaim server streaming stuff to another baggageclaim server, with that traffic going through a "streamer" (something that connects `stream-out` from one to `stream-in` from another, just how `atc` does). In this case, the setup is:

- 24GB of "random" data (a bunch of different concourse release tarballs and the PAS tile)
- each baggageclaim server limitted at 2cpus and 1GB mem through cgroups (`docker-compose` setup

```yaml
services:
  bc-provider:
    image: baggageclaim
    privileged: true
    volumes:
      - ./testdata/volumes:/volumes
    command:
      - baggageclaim
      - --volumes=/volumes
      - --bind-ip=0.0.0.0
      - --driver=naive
    deploy:
      resources:
        limits:
[...]

Issue #4621, opened 2019-10-21
fly execute sometimes returns 500 error when loading an input directory containing a terraform state file


simonxca:
## Bug Report

We were trying to run `fly execute -i terraform-state-file=$TERRAFORM_PATH` and we receive 500 errors around 80% of the time because it fails to upload the input directory. The other 20% of the time, the input directory uploads successfully.

## Steps to Reproduce

1. Run Concourse on Kubernetes.

1. Deploy Ops Manager on AWS  using terraform by following the instructions here: https://docs.pivotal.io/pivotalcf/2-6/om/aws/prepare-env-terraform.html

1. After deploying Ops Manager, you should end up with a `terraform.tfstate` file.

1. Put this `terraform.tfstate` file in an empty directory.

1. Use this directory as an input into a Concourse task (for example, name it `terraform-state-file`).

1. Run `fly execute -i terraform-state-file=$PATH_TO_DIRECTORY_ABOVE`.

## Expected Results

Input directory should upload successfully and say "Done".

## Actual Results

Uploading fails 80% of the time with 500 error.

## Additional Context

Does not happen with Bosh-deployed Concourse, error only occurs with Kubernetes-deployed Concourse.

We found a workaround for this issue. By creating a copy of the input directory inside the input directory, it uploads successfully 100% of the time. More specifically:

This **doesn't work**:
```
terraform-state-file/
|-- terraform.tfstate
```

This **does work**:
```
terraform-state-file/
|-- terraform.tfstate
|-- terraform-state-file/
    |-- terraform.tfstate
```

## Version Info

* Concourse version: 5.6.0
* Deployment type (BOSH/Docker/binary): Kubernetes (Helm)
* Infrastructure/IaaS: GCP
* Browser (if applicable): Chrome
* Did this used to work? Not sure.

xtremerui:
Could you provide the problematic terraform.tfstate? It seems a bit too much to do step 2 for reproducing this issue if the source of problem is the tfstate file and k8s concourse.

xtremerui:
Also when you see the 500 error do you have access to concourse logs so you could grap the detail error message?

simonxca:
Sure we can do that.

antonyoneill:
Hey guys, I've just fallen foul of this same issue.

Initially I saw it while uploading a git repo as an input to a job and have since managed to narrow it down to a reproducible example:

```bash
#!/usr/bin/env bash

# Navigate to a temporary dir
mkdir tmp-4621
[...]

PR #5398, opened 2020-04-02
Default to gzip compression

xtremerui:
# Existing Issue
#5195 

Add a flag `streaming-artifacts-compression` for switching between zstd/gzip compression in backend. To temporary solve #4621 , concourse API servers and `fly` are gzip only.

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [ ] New config flags added? Ensure that they are added to the [BOSH]


atc/worker/artifact_source.go
```diff
@@ -90,7 +93,7 @@ func (source *artifactSource) StreamFile(
 		reader: tarReader,
 		closers: []io.Closer{
 			out,
-			zstdReader,
+			CloseWithError{zstdReader},
```
aoldershaw: Looks like the library [provides a convenience method](https://github.com/klauspost/compress/blob/e2d2089232ece4911272106ae65a7a06746ea38a/zstd/decoder.go#L396-L402) for this purpose, so we don't need our own wrapper

```suggestion
			zstdReader.IOReadCloser(),
```

xtremerui: the `CloseWithError` wrapper is gone in the final changes. `zstdReader` is now a general interface `compressionReader`
https://github.com/concourse/concourse/pull/5398/files#diff-8643c2bdbcfa9111b8325ff14c713a65R104-R105

atc/compression/compression.go
```diff
@@ -0,0 +1,19 @@
...
+)
+
+//go:generate counterfeiter . Compression
+
+type Compression interface {
+	NewReader(io.ReadCloser) (Reader, error)
+	Encoding() baggageclaim.Encoding
+}
+
+type Reader interface {
```
cirocosta, line 16: This is literally [`io.ReadCloser`](https://golang.org/pkg/io/#ReadCloser)

```
type ReadCloser interface {
    Reader
    Closer
}

type Reader interface {
    Read(p []byte) (n int, err error)
}

type Closer interface {
    Close() error
}
```

which you used as the argument for `NewReader` :thinking:  was there a specific reason for having a separate iface?

atc/atccmd/command.go
```diff
@@ -179,6 +180,7 @@ type RunCommand struct {
 	BuildTrackerInterval time.Duration `long:"build-tracker-interval" default:"10s" description:"Interval on which to run build tracking."`
 
 	TelemetryOptIn bool `long:"telemetry-opt-in" hidden:"true" description:"Enable anonymous concourse version reporting."`
+	ZstdOptIn      bool `long:"zstd-opt-in" hidden:"true" description:"Enable zstd compression."`
```
cirocosta, line 183: Was there a specific reason to make it `hidden`? Also, I think we should go with `enable-*` instead given that we've been doing that for all of the other "opt-in" features:

```console
$ concourse web --help 2>&1 | grep enable
          --enable-global-resources                                                                             Enable equivalent resources
          --enable-build-auditing                                                                               Enable auditing for all api
          --enable-container-auditing                                                                           Enable auditing for all api
          --enable-job-auditing                                                                                 Enable auditing for all api
          --enable-pipeline-auditing                                                                            Enable auditing for all api
          --enable-resource-auditing                                                                            Enable auditing for all api
          --enable-system-auditing                                                                              Enable auditing for all api
          --enable-team-auditing                                                                                Enable auditing for all api
          --enable-worker-auditing                                                                              Enable auditing for all api
          --enable-volume-auditing                                                                              Enable auditing for all api
          --enable-redact-secrets                                                                               Enable redacting secrets in
          --enable-lets-encrypt                                                                                 Automatically configure TLS
          --secret-cache-enabled                                                                                Enable in-memory cache for
          --secret-cache-duration=                                                                              If the cache is enabled,
          --secret-cache-duration-notfound=                                                                     If the cache is enabled,
          --secret-cache-purge-interval=                                                                        If the cache is enabled,
```

xtremerui: nothing special here just follow the flag that has "OptIn" in it.
One reason I hesitated to make it a "enable" flag is we know there is still memory issue around it, which make me feel like we should not encourage user to opt in zstd until we figure it all out. I could make it not hidder for sure.

atc/compression/compression.go
xtremerui: Because the zstd Decoder has a different Close() method who doesn't return error

https://github.com/klauspost/compress/blob/4a2d40e4b07e5b3d333bc0569facd0f2dbf4ef39/zstd/decoder.go#L372

xtremerui: Good point I will update.

atc/atccmd/command.go
xtremerui: The way around is we make it a `enable` like feature flag and dogfooting in our own CI so we could observe and capture potential issues that mentioned in https://github.com/concourse/concourse/pull/5183

WDYT?

cirocosta: makes sense! thaanks!

xtremerui:
@cirocosta the flag is updated

atc/worker/artifact_source.go
```diff
@@ -73,13 +82,16 @@ func (source *artifactSource) StreamFile(
...
+	out, err := source.volume.StreamOut(ctx, filepath, source.compression.Encoding())
 	if err != nil {
 		return nil, err
 	}
 
-	zstdReader := zstd.NewReader(out)
-	tarReader := tar.NewReader(zstdReader)
+	compressionReader, err := source.compression.NewReader(out)
+	if err != nil {
+		return nil, runtime.FileNotFoundError{Path: filepath}
```
cirocosta, line 92: I'm not so sure if this is right :thinking:  

From what I understood, `NewReader` would error in multiple cases not really related to the file not having been found before (e.g., zstd failing to allocate buffers, or something like that).

On the other hand, `StreamOut` (few lines above), is something that is capable of returning a "file not found" type of error. e.g., consider the following call stack:

```
 0  0x0000000001a9be9b in github.com/concourse/baggageclaim/client.(*client).streamOut
    at /go/pkg/mod/github.com/concourse/baggageclaim@v1.6.5/client/client.go:288
 1  0x0000000001a9e6e1 in github.com/concourse/baggageclaim/client.(*clientVolume).StreamOut
    at /go/pkg/mod/github.com/concourse/baggageclaim@v1.6.5/client/volume.go:47
 2  0x0000000001adfa00 in github.com/concourse/concourse/atc/worker.(*volume).StreamOut
    at ./atc/worker/volume.go:92
 3  0x0000000001acbe82 in github.com/concourse/concourse/atc/worker.(*artifactSource).StreamFile
    at ./atc/worker/artifact_source.go:85
 4  0x0000000001ad3299 in github.com/concourse/concourse/atc/worker.(*client).StreamFileFromArtifact
    at ./atc/worker/client.go:709
 5  0x0000000001b9df3c in github.com/concourse/concourse/atc/exec.FileConfigSource.FetchConfig
    at ./atc/exec/task_config_source.go:84
 6  0x0000000001ba750b in github.com/concourse/concourse/atc/exec.(*FileConfigSource).FetchConfig
    at <autogenerated>:1
 7  0x0000000001b9e8ea in github.com/concourse/concourse/atc/exec.(*OverrideParamsConfigSource).FetchConfig
    at ./atc/exec/task_config_source.go:121
 8  0x0000000001b9f409 in github.com/concourse/concourse/atc/exec.InterpolateTemplateConfigSource.FetchConfig
    at ./atc/exec/task_config_source.go:168
 9  0x0000000001ba7862 in github.com/concourse/concourse/atc/exec.(*InterpolateTemplateConfigSource).FetchConfig
    at <autogenerated>:1
10  0x0000000001b9fdf9 in github.com/concourse/concourse/atc/exec.ValidatingConfigSource.FetchConfig
    at ./atc/exec/task_config_source.go:205
11  0x0000000001ba7ba9 in github.com/concourse/concourse/atc/exec.(*ValidatingConfigSource).FetchConfig
    at <autogenerated>:1
12  0x0000000001ba1799 in github.com/concourse/concourse/atc/exec.(*TaskStep).run
    at ./atc/exec/task_step.go:181
13  0x0000000001ba0dae in github.com/concourse/concourse/atc/exec.(*TaskStep).Run
```

looking at `github.com/concourse/baggageclaim/client.(*client).streamOut()`, we can see that the client does indeed perform some error checking:

```go
func getError(response *http.Response) error {
	var errorResponse *api.ErrorResponse
	err := json.NewDecoder(response.Body).Decode(&errorResponse)
	if err != nil {
		return err
	}

	if errorResponse.Message == api.ErrStreamOutNotFound.Error() {
		return baggageclaim.ErrFileNotFound
	}

	if response.StatusCode == 404 {
		return baggageclaim.ErrVolumeNotFound
	}

	return errors.New(errorResponse.Message)
}
```

https://github.com/concourse/baggageclaim/blob/aa42b1495315f94ff5ed8842df8b564d46cd572c/client/client.go#L298-L300

xtremerui: hmm this line comes from the next couple lines that handles the tarReader.Next(). Given that there are multiple cases here that could not be captured by a single runtime error, we might just want to bubble up the error then?

xtremerui: Since now we wrap around the reader from streamOut by `compression.NewReader()`, it is then decided by the underneath compression library that what error will be returned i.e. for now it is zstd.NewReader() and gzip.NewReader().

If the file is not found, the error would be captured on `compression.NewReader()` and it will never reach tar.NewReader() anymore.

I will remove the test and also remove the code to wrap error to be file not found err.

cirocosta: :+1: 

@xtreme-sameer-vohra,  I think you were the one most involved in the refactor that introduced the `runtime.FileNotFoundError`, do you see any problem with this approach? Ultimately, the UX is the same (in the UI, you see `task config 'config/something-else.yaml' not found`), and given that there's no place where we branch based on `runtime.FileNotFoundError`, it seems good to me

xtremerui: Removed the documentation of `runtime.FileNotFoundError`

xtremerui:
@xtreme-sameer-vohra if you also give a green light then I will go ahead and merge this.

atc/atccmd/command.go
```diff
@@ -124,6 +125,7 @@ type RunCommand struct {
 	InterceptIdleTimeout time.Duration `long:"intercept-idle-timeout" default:"0m" description:"Length of time for a intercepted session to be idle before terminating."`
 
 	EnableGlobalResources bool `long:"enable-global-resources" description:"Enable equivalent resources across pipelines and teams to share a single version history."`
+	EnableZstdCompression bool `long:"enable-zstd-compression" description:"Enable zstd compression for internal streaming."`
```
vito, line 128: Can we change this to something like `--streaming-artifacts-compression=[gzip|zstd]` with a default of `gzip`? Feels more future-proof if we want to experiment with various compression algorithms.

atc/worker/artifact_source.go
cirocosta: ping @xtreme-sameer-vohra

xtreme-sameer-vohra:
sorry folks, looking at it now

atc/worker/artifact_source.go
xtreme-sameer-vohra: Yes, agree with @cirocosta , the `out, err := source.volume.StreamOut(ctx, filepath, source.compression.Encoding())` would yield the FileNotFound error.

The [runtime.FileNotFoundError[(https://github.com/concourse/concourse/blame/master/atc/runtime/errors.go#L7) was introduced a long time ago and I can't think of a case where we need it for the `tarReader.Next` error.

xtreme-sameer-vohra: I haven't reviewed the entire PR, but this section looks good

xtreme-sameer-vohra: One option would be to move the encoding from the `stream out` call to the initialization of the baggageclaim client as the compression doesn't change once Concourse server is up. 

That would require bumping the baggageclaim library. However, since it doesn't change the contract between the client (baggageclaim) and the server (baggageclaim), it wouldn't require bumping worker version.

xtreme-sameer-vohra:
Dismissing @vito 's review as @xtremerui has updated baggageclaim to 1.6.6 in order to merge the PR

xtreme-sameer-vohra:
> We un-bumped the Baggageclaim dependency to v1.6.5 to resolve the errors with native Go zstd implementation, so I think we can bump it back to v1.6.6 as part of this PR. This will also pull in a fix for the `overlay` driver that will let us un-pend this `testflight` test:
> 
> https://github.com/concourse/concourse/blob/4e553cec96d8677d8482845d75b82a3a26aece6e/testflight/volume_mounting_test.go#L39-L44
> 
> @xtremerui Would you mind adding those changes to this PR? Thanks!
> 
> Also left a comment on the flag.

Dismissing @vito 's review as @xtremerui has updated baggageclaim to 1.6.6 in order to merge the PR

xtremerui:
deployed CI with https://github.com/concourse/hush-house/commit/d502e1e837ad17d83c95b1d9c28bffe768e61f90

PR #5416, opened 2020-04-09
cmd: generate client secret defaults in web cmd

pivotal-jwinters:
Generate random client secrets if not configured at startup.


docker-compose.yml
```diff
@@ -30,8 +30,6 @@ services:
       CONCOURSE_ADD_LOCAL_USER: test:test,guest:guest
       CONCOURSE_MAIN_TEAM_LOCAL_USER: test
       CONCOURSE_CLUSTER_NAME: dev
-      CONCOURSE_CLIENT_SECRET: Y29uY291cnNlLXdlYgo=
```
cirocosta, line 33: Given that `web` has both `tsa` and `tsa` together, and when running `docker-compose` for dev purposes only we have only a single web node, there's no problem in having these secrets being rotate on every restart of `web` - is that right?

pivotal-jwinters: Nope. Not a problem. These values get threaded through the system at startup, and saved in the db under the dex `client` table. So everything can stay in sync.

vito:
@pivotal-jamie-klassen These clients are actually only used internally when communicating with Dex during the log-in flow.

For example (and I'm partly writing this for myself), if you had a single `web` node and restarted it with newly generated client secrets, but the same signing token, all previously issued tokens would continue to work, and new log-ins would just use the new secret. With multiple `web` nodes however you would need to make sure they all have the same client secret, as the value ends up being recorded in the database on startup, and Dex would only respect the last one saved.

I just pushed a commit which will make it so that operators running a multi-node cluster don't have to configure this at all, which should make this a little less confusing - though I guess we should make the help text clearer anyway and guide most folks away from setting it. The new approach is to derive the value from the signing key - which already has to be configured the same across `web` nodes - rather than generate a random one every time.

PR #5420, opened 2020-04-13
Refactor scheduler

### LINKED ISSUE #5122: Refactor scheduler interfaces

clarafu:
# Existing Issue

Fixes #5122

# Why do we need this PR?
After the big changes that came along with the new v6.0 scheduler, there were some parts of the code that could be refactored to be easier to understand. 

There's quite a few changes made here and I tried to explain them is by stepping through our thought process and outlining each of the major pain points/how it was changed in the refactor.

## The three part problem
There were three problems that we were trying to tackle in this refactor. 

**1. Scheduling by pipeline**

https://github.com/concourse/concourse/blob/de50acac94f6658c4fffbb1af603150c382571f2/atc/scheduler/runner.go#L70

and then scheduling by jobs within each pipeline.

https://github.com/concourse/concourse/blob/de50acac94f6658c4fffbb1af603150c382571f2/atc/scheduler/runner.go#L123

Technically the scheduler does not need to rely on pipelines and schedules each job independently but we decided to have it schedule per pipeline first to avoid duplicate work. We need to grab out all the resources, resource types and related jobs (all the jobs within the same pipeline) for each job when we try to schedule it and if we only scheduled per job it would result in a lot of duplicated queries with multiple jobs within the same pipeline grabbing the same resources, resource types and jobs.

**2. The complicated mappings of pipelines and jobs.** 

We did the pipeline ID to jobs map in order to correlate each job to it's pipeline for scheduling.

https://github.com/concourse/concourse/blob/de50acac94f6658c4fffbb1af603150c382571f2/atc/scheduler/runner.go#L60

The name to ID map was for

https://github.com/concourse/concourse/blob/de50acac94f6658c4fffbb1af603150c382571f2/atc/scheduler/runner.go#L90-L93

**3. Passing of arguments through multiple interfaces**

One of the goals was to shrink the amount of arguments being passed around such that the scheduler runner is literally something that just does "get all jobs to schedule" -> "loop over them and schedule them with a max-in-flight". Currently, there are consistently the job to schedule, resources, related jobs and the pipeline being passed through from the runner -> scheduler -> buildstarter. 

## Attempted Solution

I'm going to split up how we were able to tackle the three goals into different parts, which are all related but is probably easier to understand if I section them off into just the major changes made.

#### Removing the need for resources to be passed around the interfaces

I started off by figuring out why resources are needed to be passed through the scheduler. Resources seemed to be used by these parts:

1. Resources are used by the algorithm to construct the `algorithm.InputConfigs` object. 

https://github.com/concourse/concourse/blob/e8998d5e387fc4f9b459be7374230b2a239fb3e6/atc/scheduler/algorithm/resolver.go#L53-L56

The algorithm uses a resource to grab it's resource ID and to figure out if the resource is pinned. Trying to move away from that, we found that we can also grab that information directly from the database. I moved the code for constructing the `algorithm.InputConfigs` from the algorithm to the place where we fetch out the `JobInputs` so that when we grab out the inputs for a job, rather than constructing these `JobInputs` that will later in the algorithm be turned into a `algorithm.InputConfigs`, we will directly construct the `algorithm.InputConfigs` object when we fetch the information from the database.

https://github.com/concourse/concourse/blob/b0d9a9507a849b451e60b87c7f1e519dcaf96474/atc/db/job.go#L221-L230

The new method that holds this logic is the `AlgorithmInputs()` method on the `Job` interface.

2. For manually triggered builds, the resources are used to check if they have been checked for new versions after the build creation time.

https://github.com/concourse/concourse/blob/e8998d5e387fc4f9b459be7374230b2a239fb3e6/atc/scheduler/build.go#L25-L41

To avoid this, we are able to use the `job_inputs` table to figure out which resources are used by the manually triggered job, and check through a single query if all of it's resources have been checked after the creation of the manually triggered build. 

https://github.com/concourse/concourse/blob/b0d9a9507a849b451e60b87c7f1e519dcaf96474/atc/db/build.go#L294-L309

When I was trying to write a single query for figuring out if all the resources have been checked after the creation of the build, I realized that I would need to make a change to how a config pinned version is represented in the database. This is because we don't want to require a resource that is pinned either through the API or through the pipeline config to need to be checked for manually triggered builds. Currently, the only way to figure out if a resource is pinned through the pipeline config is by querying out the resource config, unmarshal, decrypt and then pull of the `version` portion of the resource config, which cannot be done within a single query. So I made a change to have the `resource_pins` table also hold the information for resources that are pinned through the pipeline config which is represented through the `config` boolean. 

https://github.com/concourse/concourse/blob/b0d9a9507a849b451e60b87c7f1e519dcaf96474/atc/db/team.go#L1021-L1023

Now that both API and config pinned versions are represented through the `resource_pins` table, we can do a single query to figure out if all unpinned resources have been checked for manually triggered builds.

3. Construct `atc.ResourceConfigs` to create a build plan

https://github.com/concourse/concourse/blob/e8998d5e387fc4f9b459be7374230b2a239fb3e6/atc/scheduler/buildstarter.go#L223-L231

This is still necessary because we need to use the `ResourceConfigs` to create a build plan. But rather than using the entire Resource object to construct a `ResourceConfig` which only cares about a few fields, we can have the build plan take in a specialize object that contains only the fields it needs and query for only those pieces of data. This was implemented in the new `JobsToSchedule` method that will now grab out all the jobs that needed to be scheduled, the necessary information for all the resources associated to that job, and also the resource types for the pipeline of that job.

https://github.com/concourse/concourse/blob/b0d9a9507a849b451e60b87c7f1e519dcaf96474/atc/db/job_factory.go#L94-L101

#### Is the job name to ID mapping necessary?

This job name to ID mapping was done so that we could pass all the jobs that are in the same pipeline to the algorithm. The algorithm uses that map in order to find the job IDs for all the jobs that are passed constraints to the input that it is constructing.

https://github.com/concourse/concourse/blob/e8998d5e387fc4f9b459be7374230b2a239fb3e6/atc/scheduler/algorithm/resolver.go#L87-L90

In order to avoid doing this job mapping, I used the `job_inputs` table to grab out all the `passed_ job_id`s from the database directly when we grab out the `InputConfigs` for the job we want to schedule. 

https://github.com/concourse/concourse/blob/b0d9a9507a849b451e60b87c7f1e519dcaf96474/atc/db/job.go#L277-L282

#### Scheduling per pipeline

Now that we have removed the need to grab out all the resources and jobs per pipeline for each job that we want to schedule from the points outlined above, we no longer need to schedule per pipeline in the `runner.go`!

But there is one more change that is needed to make before that can happen, because we actually pass the pipeline through the scheduler interfaces in order to check if the pipeline is paused. So I made a change in the `ScheduleBuild` method where it is actually also responsible now for checking if the job or pipeline is paused, returning `false` if either of them are paused.

https://github.com/concourse/concourse/blob/b0d9a9507a849b451e60b87c7f1e519dcaf96474/atc/db/job.go#L539-L542

With the need for passing a pipeline through the interfaces removed, we no longer need to scheduler per pipeline and we no longer need to do that pipeline ID to jobs to schedule mapping!

## Behavioural changes that came with the refactor

Now that resources, resource types and jobs are fetched from the database at the very beginning where we find which jobs need to be scheduled, they will no longer be the most "up to date" as they were before this refactor. What I mean by that is, for example, before we would've grabbed out which jobs to schedule and we would grab out the details of the resource types only when we were just about to create a build plan. If someone made a change to the resource types between the time that the job was set to be scheduled and the time that the build plan is created, before the refactor we would have picked up on the change but with the refactor we would **not** pick up the change. This might actually be a good thing, because it means that all the data we pull out for the scheduler is consistent as it is all grabbed out at the same time. 

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


atc/db/job_factory.go
```diff
@@ -30,6 +35,148 @@ func NewJobFactory(conn Conn, lockFactory lock.LockFactory) JobFactory {
...
+
+	jobs, err := scanJobs(j.conn, j.lockFactory, rows)
+	if err != nil {
+		return nil, err
+	}
+
+	var schedulerJobs SchedulerJobs
+	pipelineResourceTypes := make(map[int]ResourceTypes)
+	for _, job := range jobs {
+		rows, err := psql.Select("r.name", "r.type", "r.config", "r.nonce").
```
pivotal-jwinters, line 94: Is looking up the resources for each job better than doing the top level resources query?

clarafu: @pivotal-jwinters You mean like using the `job_inputs` table to find what resources are for that job, check if they use any custom resource types and then grab out all the resource types that are used by that job?

pivotal-jwinters: I just mean we used to grab all the resources for a pipeline up front:
https://github.com/concourse/concourse/blob/master/atc/scheduler/runner.go#L80

But here we are making many queries to fetch resources per job. I was just wondering if you found that this was some way better?

But now that you mention resource types, It also seems weird because the only thing that needs the resource types is the the build starter (is that true?). But now we're fetching the resource types super early in the scheduler (which doesn't actually need them)? To me, the scheduler and build starter should be looked at as two different components but this seems to couple their behaviour.

clarafu: Ohh sorry I thought you were talking about the query for resource types. If you mean better as in faster or less duplicate work, I'm *pretty* sure doing it once per pipeline and then just passing all of them to each job is technically faster and less duplicate work because it is a single bigger query rather than multiple smaller queries that are probably grabbing the same resources.

I think what we were trying to achieve is to have the scheduler be completely job focused and no longer needs to care about pipelines. But I could definitely be convinced that the performance overhead is not worth it.

pivotal-jwinters: Oh i see. Yeah making it entirely job focused makes a lot of sense to me. What if we just query for ALL resources up front? Similar to what the lidar scanner does?

clarafu: Hmm if we did that we would still need to figure out which resources go with which jobs, so we would need to query to the `job_inputs` table for each job right?

pivotal-jwinters: What if we do that whole job_inputs joining resources query upfront for all jobs?

clarafu: Just a thought too, now that we only schedule jobs that need to be scheduled it might not be a problem doing a query for the resources for each job? I'm just worried that combining it all into one query might make the code more complicated especially if its for optimizing something that might not be a problem.

pivotal-jwinters: Hmm yeah good point. okey-dokey

atc/db/resource.go
```diff
@@ -584,13 +587,29 @@ func (r *resource) PinVersion(rcvID int) (bool, error) {
 		return false, err
 	}
```
aoldershaw, line 589: Ah, seems like we never rolled back the transaction on error cases...

```suggestion
	 defer Rollback(tx)
```

Can you also add this to `Resource.UnpinVersion`, `Job.GetNextBuildInputs`, `VolumeRepository.RemoveMissingVolumes`, `Pipeline.LoadDebugVersionsDB`?

(that's quite a bit more than I was expecting to find...I wonder if we can have some automated check for ensuring we always `defer Rollback(tx)`...)

atc/db/job_test.go
```diff
@@ -2109,186 +2109,447 @@ var _ = Describe("Job", func() {
...
+						{"api": "pinned"},
+					})
+					Expect(err).NotTo(HaveOccurred())
+
+					versionID, found, err := pinnedResource.ResourceConfigVersionID(atc.Version{"api": "pinned"})
+					Expect(err).NotTo(HaveOccurred())
+					Expect(found).To(BeTrue())
+
+					pinned, err := pinnedResource.PinVersion(versionID)
+					Expect(err).NotTo(HaveOccurred())
```
aoldershaw, line 2268: Wait, I thought this pinning a version through the API if it's already pinned in the config returned an error - https://github.com/concourse/concourse/pull/5420/files#diff-37e62c267e080752855dde75c8aa7d8eR602-R604

atc/db/migration/migrations/1585079293_add_config_pinned_version_to_resource_pins.up.go
```diff
@@ -0,0 +1,80 @@
...
+
+	for resourceID, version := range configPinnedVersions {
+		versionJSON, err := json.Marshal(version)
+		if err != nil {
+			return err
+		}
+
+		_, err = tx.Exec(`
+			INSERT INTO resource_pins (resource_id, version, config, comment_text)
+			VALUES ($1, $2, true, '')
```
aoldershaw, line 71: Should we add the same `comment_text` as we do here:

https://github.com/concourse/concourse/blob/b0d9a9507a849b451e60b87c7f1e519dcaf96474/atc/db/team.go#L1021-L1026

clarafu: Ya I only realized after that I don't think we actually show the comment text if it's pinned in from the config.. maybe I should just remove it from the `team.go`? 🤔

atc/db/resource.go
clarafu: Great catch! Thanks :D 

For the other ones do you think it would be worth making a separate PR for those?

atc/db/resource_test.go
```diff
@@ -1711,21 +1713,15 @@ var _ = Describe("Resource", func() {
...
-				found, err = resource.PinVersion(resConf.ID())
-				Expect(found).To(BeTrue())
-				Expect(err).ToNot(HaveOccurred())
-
-				found, err = resource.Reload()
-				Expect(found).To(BeTrue())
-				Expect(err).ToNot(HaveOccurred())
 			})
 
 			It("should return the config pinned version", func() {
```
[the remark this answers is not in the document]
clarafu: Yessir!

atc/db/job.go
```diff
@@ -533,6 +536,15 @@ func (j *job) ScheduleBuild(build Build) (bool, error) {
...
 
+	paused, err := j.isPipelineOrJobPaused(tx)
+	if err != nil {
+		return false, err
+	}
+
+	if paused {
+		return false, nil
+	}
+
```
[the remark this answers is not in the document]
clarafu: This is just to make sure that when we actually get to the point that we will start the build, the user did not pause the pipeline/job between the time that we decided this job should be scheduled and the build actually starting. This check to see if the pipeline/job is paused right before we start the build was moved from here https://github.com/concourse/concourse/blob/c15fb421208871c2763b13d497690cbe524f7d67/atc/scheduler/buildstarter.go#L166-L179
to this `ScheduleBuild` method.

atc/db/job_factory_test.go
```diff
@@ -732,14 +732,26 @@ var _ = Describe("Job Factory", func() {
...
+					for _, job := range jobs {
+						_, exists := jobsMap[job.Name()]
+						Expect(exists).To(BeFalse())
+
+						jobsMap[job.Name()] = job
+					}
+
+					job1, exists := jobsMap[job1.Name()]
+					Expect(exists).To(BeTrue())
+					Expect(job1.Name()).To(Equal(job1.Name()))
```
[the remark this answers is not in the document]
clarafu: Oh geez ya I hate this part of the test so much. I think your idea is good! At least I like it way better than mine haha I think I have this done in a few places across the codebase so a custom Gomega matcher should be something worth doing. I have definitely come into a couple of cases where I want to compare a slice of objects but not care about the ordering of a slice within the objects. I'll add that to my list of changes!

atc/db/job_test.go
clarafu: Ohh the test is for pinning through the get step. You can either pin the resource through it's resource config and also through the get step. Hm there might be a few things I should reword to make that more clear. I think I should rephrase the test to be more specific and possibly have it be `pinnedThroughResourceConfig` rather than `pinnedThroughConfig`? I might look around at rewording more things too to make a bigger distinction?

atc/db/job_factory_test.go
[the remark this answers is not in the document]
aoldershaw: The custom matcher could get a bit messy with reflection if I had to guess, something to keep in mind

atc/db/job_test.go
aoldershaw: Was not aware of that distinction! Thanks!

clarafu:
@aoldershaw I added the fixes you mentioned, there was only especially that I wanted you to go over too because I ended up rewriting the tests using case statements rather than the custom matcher that you proposed. I'd really like to hear your thoughts on the tests using case statements vs custom matcher? 

https://github.com/concourse/concourse/blob/be6d2e549a034c994bb188e2d78907cff297fd10/atc/db/job_factory_test.go#L736-L771

One thing left to do also is to create a separate PR for adding `tx.Rollback()` to the following methods: Resource.UnpinVersion, Job.GetNextBuildInputs, VolumeRepository.RemoveMissingVolumes, Pipeline.LoadDebugVersionsDB.

atc/db/job_factory_test.go
```diff
@@ -730,52 +736,39 @@ var _ = Describe("Job Factory", func() {
...
 					Expect(err).ToNot(HaveOccurred())
-					Expect(len(jobs)).To(Equal(3))
 
-					jobsMap := make(map[string]db.SchedulerJob)
 					for _, job := range jobs {
-						_, exists := jobsMap[job.Name()]
-						Expect(exists).To(BeFalse())
-
-						jobsMap[job.Name()] = job
+						switch job.Name() {
```
aoldershaw, line 741: I'm going to preface this by saying sorry for jumping around on this! Honestly, I think the test you had initially was mostly fine. I'm just curious to see if we can figure out a better way to do tests like these. Feel free to ignore me, though! Anyway:

I definitely feel this is better than a custom matcher, which would be almost certainly overkill. With that said, we do lose some coverage with this approach (at least the way it is now) - this test would pass in a few errant cases (there are no jobs; there are 3 jobs, but it's just "job1" repeated 3 times; etc). Now, we can add that coverage back by adding some additional assertions and logic here, but then we're basically back where we started!

I'm going to propose another possibility. Let me know what you think:

```go
...
// If we don't check this, we could have e.g. job1 present multiple times
Expect(len(jobs)).To(Equal(3))
jobResources := make(map[string][]db.SchedulerResource)
for _, job := range jobs {
    jobResources[job.Name()] = job.Resources
}
Expect(jobResources).To(ConsistOf(map[string]interface{}{
    job1.Name(): ConsistOf(
        db.SchedulerResource{
            Name: "some-resource",
            Type: "some-type",
        },
    ),
    job2.Name(): ConsistOf(...),
    job3.Name(): ConsistOf(...),
})
```

I haven't tested it, but it does appear `ConsistOf` can match on `maps`

clarafu: @aoldershaw Thats a really good point, using case statements can definitely give false positives! Haha don't apologize for that, its an iterative process and I'm really liking all the suggestions you are making. I think the approach you proposed is good, it might be unavoidable to be able to do this well without the weird mapping setup.

I was experimenting with using `MatchAllKeys` https://godoc.org/github.com/onsi/gomega/gstruct#MatchAllKeys rather than `ConsistOf` for comparing the map of scheduler resources? So it looks like this
```
Expect(jobResources).To(MatchAllKeys(Keys{
    job1.Name(): ConsistOf(
        db.SchedulerResource{
            Name: "some-resource",
            Type: "some-type",
        },
    ),
    job2.Name(): ConsistOf(...),
    job3.Name(): ConsistOf(...),
})
```
because I don't think the `ConsistOf` works with the nested `ConsistOf`s 🤔

aoldershaw:
Oh, and do you want me to open an issue for the missing `Rollbacks`?

clarafu:
@aoldershaw Thanks for all your feedback! Hm I know @vito has a lot on his plate but maybe if he has time he can do a quick look through the code since he is really familiar with it too? 

In terms of acceptance, I think basic acceptance of making sure scheduling still works in a few different random cases is good because this refactor is mostly just changing the structure of how scheduling is done. But if you would like some extra acceptance testing criteria, I would probably choose the parts of the code that I changed the most. What comes to mind would be:
- scheduling all jobs still works between multiple pipelines
- jobs that have passed constraints can still schedule builds appropriately
- manually triggered builds will wait on checks to finish and then schedule after checks have finished (excluding pinned resources)
- paused jobs or pipelines should not schedule any builds

Also there was some extra work that came about from this refactor where I had to store the resource config pinned versions into the `resource_pins` table. So I could see some acceptance criteria around that:
- pinning a version through the resource config still works
- pinning a version through the api and then pinning it to the resource config should overwrite the api pinned version

Ah yes I just added the issue for the `Rollbacks` here https://github.com/concourse/concourse/issues/5557!

aoldershaw:
@clarafu I've done some more thorough acceptance on this:

* multiple pipelines with `time` resources with different (and same) intervals, made sure can both can be scheduled
* pipeline with passed constraint
* manually triggering build resets the "checked 0s ago" timer on its resources
* paused jobs are paused, paused pipelines are paused
* pinning a resource through API and config works (and populates `resource_pins` table appropriately)
* pinning a resource through config overwrites API pin

atc/db/job_factory.go
```diff
@@ -30,6 +35,156 @@ func NewJobFactory(conn Conn, lockFactory lock.LockFactory) JobFactory {
...
+	ResourceTypes atc.VersionedResourceTypes
+}
+
+type SchedulerResources []SchedulerResource
+
+type SchedulerResource struct {
+	Name   string
+	Type   string
+	Source atc.Source
+	Tags   atc.Tags
```
vito, line 52: 🤔 Why are the tags necessary? I may be missing something. Just seems like the odd one out.

clarafu: Oh ya I don't think it's necessary, looks like we don't use it to construct the build plan!

PR #5429, opened 2020-04-13
atc/api: limit concurrent requests

### LINKED ISSUE #5421: API concurrency limiting
## operator configuration

```
concourse web \
  ...
  --concurrent-request-limit ListAllJobs=5 \
  --concurrent-request-limit ListAllPipelines=10
```

* values are `<api action>=<non-negative integer>`.
* what happens if the flag doesn't parse? barf with an error
* what happens if they specify a non-existent/illogical endpoint?
    * the only supported endpoint, for now, should be `ListAllJobs` - so we can avoid any footguns
* does setting a value of zero effectively disable the endpoint? should this replace the `--disable-list-all-jobs` implementation?

## API behaviour
* error log when the limit is hit
* give a `429 Too Many Requests` when the limit is hit. We may add a `Retry-After` in a later change.
* let's make sure that, for scoped handlers (like those returned by the `pipelineserver.ScopedHandlerFactory`) the concurrency-limiting logic takes place before any DB conns get opened.
* please ensure this is zero-cost -- any endpoints that do not have a concurrency limit in place should not get slower because of this change.
* Implementation reference: https://pauladamsmith.com/blog/2016/04/max-clients-go-net-http.html

* web UI should handle it gracefully - what this means is TBD, but let's track the work in #2935.

### LINKED ISSUE #5425: metric for API concurrency
use a gauge, tagging with the api endpoint. use the response time metric as an example. if this behaviour is expensive, only track concurrency for endpoints whose concurrency is being limited per #5421.

can we add a docker-compose override that lets us visualize this? more of a nice-to-have as delivering this in a timely fashion is a higher priority.

### LINKED ISSUE #5471: atc: return 501 when `--concurrent-request-limit` is 0

pivotal-jamie-klassen:
# Existing Issue

Fixes #5421.
Fixes #5425
Fixes #5471 

* if you pass the `--concurrent-request-limit <action>:<limit>` then any concurrent requests beyond the specified limit will be rejected with a 503 status. If that limit is zero, the request will be rejected with a 501.
* action needs to be valid, and only `ListAllJobs` is permitted. if you provide an malformed, invalid or forbidden value the web node will fail to start.
* if you configure multiple limits for the same action, the last flag wins
* there are now 'concurrent requests' (a gauge measuring the current/instantaneous number of requests currently being served by the given web node) and 'concurrent requests limit hit' (a counter for the number of times a request is rejected for concurrent-request-limiting reasons)

# Contributor Checklist
- [x] Unit tests
- [x] Updated documentation - concourse/docs#310
- [x] helm: concourse/concourse-chart#97
- [x] bosh: concourse/concourse-bosh-release#97, concourse/concourse-bosh-release#98

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


pivotal-jamie-klassen:
Still need to decide what to do if the same endpoint is limited twice, i.e.
```
concourse web \
  --concurrent-request-limit ListAllJobs=2 \
  --concurrent-request-limit ListAllJobs=3
```

cirocosta:
> Still need to decide what to do if the same endpoint is limited twice, i.e.

perhaps throw an error? I imagine it might be better to fail explicitly rather than implicitly decide on one or the other, wdyt?

atc/wrappa/concurrent_request_policy.go
```diff
@@ -0,0 +1,142 @@
...
+	for _, route := range atc.Routes {
+		if route.Name == flag.Action {
+			return nil
+		}
+	}
+	return formatError(
+		fmt.Sprintf("%s=%d", flag.Action, flag.Limit),
+		fmt.Sprintf("'%s' is not a valid action", flag.Action),
+	)
+}
```
pivotal-jamie-klassen, line 142: I think this should be moved into `UnmarshalFlag` above, and we should add the non-negative constraint there, too.

atc/wrappa/concurrent_request_limits_wrappa.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+	}
+
+	return wrapped
+}
+
+func wrapHandler(logger lager.Logger, limit int, handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusTooManyRequests)
+	})
```
pivotal-jamie-klassen, line 47:
```suggestion
	handlerPool := semaphore.NewWeighted(int64(limit))
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if !handlerPool.TryAcquire(1) {
			w.WriteHeader(http.StatusTooManyRequests)
		} else {
			defer handlerPool.Release(1)
			handler.serveHTTP(w,r)
		}
	})
```
This is generally the right implementation, although we'll probably create a `Pool` interface to accomodate the metric feature. There should be a `HandlerPool(string action) Pool` method on `ConcurrentRequestPolicy` and then the pool could get passed in
```suggestion
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if !handlerPool.TryAcquire() {
			w.WriteHeader(http.StatusTooManyRequests)
		} else {
			defer handlerPool.Release()
			handler.serveHTTP(w,r)
		}
	})
```

atc/wrappa/concurrent_request_limits_wrappa.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func (wrappa ConcurrencyLimitsWrappa) Wrap(handlers rata.Handlers) rata.Handlers {
+	wrapped := rata.Handlers{}
+
+	for name, handler := range handlers {
+		if wrappa.concurrentRequestPolicy.IsLimited(name) {
+			wrapped[name] = wrapHandler(
+				wrappa.logger,
+				wrappa.concurrentRequestPolicy.MaxConcurrentRequests(name),
```
pivotal-jamie-klassen, line 33:
```suggestion
				wrappa.concurrentRequestPolicy.HandlerPool(name),
```

atc/integration/concurrency_limit_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"net/http"
+
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/atc/wrappa"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = FDescribe("Concurrent request limits", func() {
```
pivotal-jamie-klassen, line 12:
```suggestion
var _ = Describe("Concurrent request limits", func() {
```

atc/atccmd/command_test.go
```diff
@@ -27,6 +27,20 @@ func (s *CommandSuite) TestLetsEncryptDefaultIsUpToDate() {
 	s.Equal(opt.Default, []string{autocert.DefaultACMEDirectory})
 }
 
+func (s *CommandSuite) TestInvalidConcurrencyLimit() {
```
pivotal-jamie-klassen, line 30: it probably then makes more sense for this test to assert on either the "no duplicates" rule.

atc/wrappa/concurrent_request_policy.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	return assignment[0], assignment[1], nil
+}
+
+func formatError(value string, reason string) error {
+	return fmt.Errorf("invalid concurrent request limit '%s': %s", value, reason)
+}
+
+type ConcurrentRequestPolicy interface {
+	MaxConcurrentRequests(action string) int
```
pivotal-jamie-klassen, line 43:
```suggestion
	HandlerPool(action string) Pool
```

atc/atccmd/command.go
```diff
@@ -401,6 +402,16 @@ func (cmd *RunCommand) Runner(positionalArguments []string) (ifrit.Runner, error
 		return nil, err
 	}
 
+	concurrentRequestPolicy := wrappa.NewConcurrentRequestPolicy(
+		cmd.ConcurrentRequestLimits,
+		[]string{atc.ListAllJobs},
+	)
```
pivotal-jamie-klassen, line 408: let's create a `concurrentRequestPolicy()` method on `RunCommand` for this, so we can pull the whole validation block into the `RunCommand.validate()` method.

pivotal-jamie-klassen: It means we'd construct the policy twice during startup, but it's super cheap, and it would mean we can avoid polluting these private methods with extra parameters.

atc/atccmd/command.go
```diff
@@ -1685,6 +1700,7 @@ func (cmd *RunCommand) constructAPIHandler(
 		logger,
 	)
 	apiWrapper := wrappa.MultiWrappa{
+		wrappa.NewConcurrencyLimitsWrappa(logger, concurrentRequestPolicy),
```
pivotal-jamie-klassen, line 1703:
```suggestion
		wrappa.NewConcurrencyLimitsWrappa(logger, cmd.concurrentRequestPolicy()),
```

atc/atccmd/command_test.go
pivotal-jamie-klassen: ... since it's really an integration test asserting on the wiring (most importantly that the `Validate()` method is being called on the concurrency policy.

atc/wrappa/concurrent_request_limits_wrappa.go
pivotal-jamie-klassen: `semaphore` is `golang.org/x/sync/sempahore`.

atc/wrappa/concurrent_request_limits_wrappa.go
```diff
@@ -0,0 +1,48 @@
...
+func (wrappa ConcurrencyLimitsWrappa) Wrap(handlers rata.Handlers) rata.Handlers {
+	wrapped := rata.Handlers{}
+
+	for action, handler := range handlers {
+		if wrappa.concurrentRequestPolicy.IsLimited(action) {
+			pool := wrappa.concurrentRequestPolicy.HandlerPool(action)
+			wrapped[action] = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if !pool.TryAcquire() {
+					wrappa.logger.Info("concurrent-request-limit-reached")
+					w.WriteHeader(http.StatusTooManyRequests)
```
pivotal-jwinters, line 37: I know this was discussed somewhere already, but is a 429 the right status code for this? In theory a user could submit a single request and the api responds telling them to rate limit themselves, which seems confusing.

I this this as more of a circuit breaker pattern, where the service would typically respond with a 503 Service Unavailable.

atc/wrappa/concurrent_request_policy.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+	return nil
+}
+
+func (crp *concurrentRequestPolicy) checkSupportedActions() error {
+	for _, limit := range crp.limits {
+		if !crp.supports(limit.Action) {
+			return formatError(
+				fmt.Sprintf("%s=%d", limit.Action, limit.Limit),
+				fmt.Sprintf("'%s' is not supported", limit.Action),
```
pivotal-jamie-klassen, line 117:
```suggestion
				fmt.Sprintf("action '%s' is not supported", limit.Action),
```
also we should probably mention which actions _are_ supported.

atc/wrappa/concurrent_request_policy.go
```diff
@@ -0,0 +1,145 @@
...
+	Limit  int
+}
+
+func (crl *ConcurrentRequestLimitFlag) UnmarshalFlag(value string) error {
+	variable, expression, err := parseAssignment(value)
+	if err != nil {
+		return err
+	}
+	limit, err := strconv.Atoi(expression)
+	if err != nil || limit < 0 {
```
pivotal-jamie-klassen, line 22: If we use https://golang.org/pkg/strconv/#ParseUint, I guess we don't need to check against negative numbers...

atc/wrappa/pool.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+type Pool interface {
+	TryAcquire() bool
+	Release() error
+}
+
+type pool struct {
+	size  int
+	state int
```
pivotal-jamie-klassen, line 15: obviously `uint64` is more annoying to write, but it is a more descriptive type...

atc/wrappa/concurrent_request_limits_wrappa.go
```diff
@@ -0,0 +1,56 @@
...
+	pool Pool,
+	handler http.Handler,
+) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if !pool.TryAcquire() {
+			wrappa.logger.Info("concurrent-request-limit-reached")
+			w.WriteHeader(http.StatusTooManyRequests)
+			return
+		}
+		defer pool.Release()
```
pivotal-jamie-klassen, line 53: it should never happen, but `Release` returns an error if you try to release more than held - maybe we should log this error if it (however unlikey) happens.

atc/wrappa/concurrent_request_limits_wrappa.go
pivotal-jamie-klassen: hmmm [the rfc defining the 429 code](https://tools.ietf.org/html/rfc6585#section-4) does pretty clearly indicate that "_the user_ has sent too many requests in a given amount of time" and this condition is not scoped or dependent on any particular user.

503, on the other hand, [says](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503) "Common causes are a server ... that is overloaded" and also allows for a `Retry-After` header, should we eventually choose to include one. It does sound like a better fit.

At first I was concerned about the web UI potentially giving special treatment to a 500-series status code, which anecdotally means "terrible bad stuff you can't possibly understand", but 503 seems like a scenario where retries are normal.

atc/wrappa/pool.go
cirocosta: this seems a lot like a semaphore where we just don't have to bother about notifying waiters :thinking:  was there a specific reason not to use https://godoc.org/golang.org/x/sync/semaphore , I guess, aside that?

pivotal-jamie-klassen: I made almost exactly the same comment in fe8ee20b47a8eba3a62 😁 -- this implementation is simpler, doesn't introduce a dependency, and returns an error instead of panicking when you try to release the semaphore when it is not held.

cirocosta: makes sense! and awesome to see that we're keeping that as part of the history :raised_hands: (and shame on me for not looking at the commit messages :sweat_smile: )

pivotal-jamie-klassen: No worries - I almost wonder if it's worth making a habit of sticking all the commit messages in the PR description so that you don't have to be vigilant about that. Otherwise your question is still valid - unless there's a strong reason for "gracefully" handling the "attempting to release an empty pool" case, it might be worth switching to the existing semaphore implementation because a) we already have a dependency on the library and b) it could allow removing a lot of code...

pivotal-jamie-klassen: We switched to the existing semaphore implementation!

pivotal-jamie-klassen:
Since Friday's refactorings, all my attempts at acceptance are failing. I've encountered the following problems:
* https://github.com/jessevdk/go-flags/issues/329 - the idea is really to use env vars to set a default, which will not be subject to the same validation rules as a flag -- I think. Maybe if we use a type alias we can get around this?
* not sure, but maybe when you create a `golang.org/x/sync/semaphore.Weighted` of sizer zero it kinda does nothing?

pivotal-jamie-klassen:
OK, it's fixed.

atc/metric/emitter/prometheus.go
```diff
@@ -151,6 +154,22 @@ func (config *PrometheusConfig) NewEmitter() (metric.Emitter, error) {
 	})
 	prometheus.MustRegister(buildsRunning)
 
+	concurrentRequestsLimitHit := prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: "concourse",
+		Subsystem: "builds",
+		Name:      "started_total",
+		Help:      "Total number of Concourse builds started.",
```
pivotal-jamie-klassen, line 161: maybe a copy-paste?

atc/metric/emitter/prometheus.go
```diff
@@ -151,6 +154,22 @@ func (config *PrometheusConfig) NewEmitter() (metric.Emitter, error) {
...
+	})
+	prometheus.MustRegister(concurrentRequestsLimitHit)
+
+	concurrentRequests := prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: "concourse",
+		Subsystem: "builds",
+		Name:      "running",
+		Help:      "Number of Concourse builds currently running.",
```
pivotal-jamie-klassen, line 169: also probably a copy-paste

cirocosta: total copy & paste :grin:  there were other issues, so I just left for a future commit :+1:

cirocosta:
doing some acceptance here:

```
# TYPE concourse_concurrent_requests gauge
concourse_concurrent_requests{action="ListAllJobs"} 1
# TYPE concourse_concurrent_requests_limit_hit_total counter
concourse_concurrent_requests_limit_hit_total{action="ListAllJobs"} 933
```

by hitting `localhost:9100` from within `concourse_web_1` after doing `for i in $(seq 1 1000); do curl localhost:8080/api/v1/jobs & done` having

```
CONCOURSE_CONCURRENT_REQUEST_LIMIT=ListAllJobs:1
```

with `hack/overrides/prometheus.yml` activated.

atc/metric/emitter/prometheus.go
cirocosta: https://github.com/concourse/concourse/pull/5429#issuecomment-616821387

atc/metric/emitter/prometheus.go
cirocosta: https://github.com/concourse/concourse/pull/5429#issuecomment-616821387

pivotal-jamie-klassen:
helpful feedback from @xtreme-sameer-vohra and @taylorsilva on the 5.5.x version of this: https://github.com/concourse/concourse/pull/5484#discussion_r413136423

atc/metric/emitter/prometheus.go
```diff
@@ -158,14 +158,14 @@ func (config *PrometheusConfig) NewEmitter() (metric.Emitter, error) {
 		Namespace: "concourse",
 		Subsystem: "concurrent_requests",
 		Name:      "limit_hit_total",
-		Help:      "Total number of Concourse builds started.",
+		Help:      "Total number of requests that failed to acquire the lock to be served.",
```
pivotal-jamie-klassen, line 161: https://github.com/concourse/docs/pull/310#discussion_r413833407

pivotal-jamie-klassen:
Also remember that stricter env var parsing affects the bosh release. We need to remove the default value for `worker_gateway.authorized_keys` and set the default for `baggageclaim.disable_user_namespaces` to `false` rather than `detect`.

pivotal-jamie-klassen:
I've also learned that k8s-topgun does some weird stuff with env vars: a1d5176ad0e30e8556e08c014254d68fe8268913, a1d5176ad0e30e8556e08c014254d68fe8268913 will need to be ported forward.

pivotal-jamie-klassen:
Furthermore, let's also port these nice release notes forward: 99e053d30bb35969356ba928142395c0859b1599

go.mod
```diff
@@ -125,3 +125,5 @@ require (
 go 1.13
 
 replace github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
+
+replace github.com/jessevdk/go-flags => github.com/vito/go-flags v1.4.1-0.20200421142706-6149d40139cb
```
pivotal-jamie-klassen, line 129: this particular version of vito/go-flags lacks an important feature: while it is nice that it gives errors on parsing env vars, it doesn't actually identify _which_ flag caused the error. this is pretty tough to sift through.

pivotal-jamie-klassen: vito/go-flags#1 addresses this, let's remember to update the dependency here.

pivotal-jamie-klassen:
currently this PR is blocked on
* [ ] concourse/ci#306 - so that the dev image can be safely used as a base image in testing
* [ ] vito/go-flags#1 and subsequent go.mod bump - just an important quality-of-life improvement
then it will be ready for review.

go.mod
pivotal-jamie-klassen: 6de27d2aa803323239eb77cc7216bd63ecbe2975

pivotal-jamie-klassen:
@clarafu now that concourse/ci#306 is merged, this is ready for review. Once https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/dev-image/builds/275 is finished and a new dev-image is built, the tests should be passing.

release-notes/latest.md
```diff
@@ -9,6 +9,24 @@
...
+
+  It is important to note that, if you use this configuration, it is possible for super-admins to effectively deny service to non-super-admins. This is because when super-admins look at the dashboard, the API returns a huge amount of data (much more than the average user) and it can take a long time (over 30s on some clusters) to serve the request. If you have multiple super-admin dashboards open, they are pretty much constantly consuming some portion of the number of concurrent requests your web node will allow. Any other requetss, even if they are potentially cheaper for the API to service, are much more likely to be rejected because the server is overloaded by super-admins. Still, the web node will no longer crash in these scenarios, and non-super-admins will still see their dashboards, albeit without nice previews. To work around this scenario, it is important to be careful of the number of super-admin users with open dashboards. #5429
```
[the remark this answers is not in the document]
pivotal-jamie-klassen: nice catch!

cmd/concourse/concourse_test.go
```diff
@@ -25,99 +25,93 @@ import (
...
+			Name:          "web",
 			AnsiColorCode: "32m",
 		})
 
-		concourseProcess = ginkgomon.Invoke(concourseRunner)
+		concourseProcess = ifrit.Background(concourseRunner)
+
+		select {
+		case <-concourseProcess.Ready():
+			// case err := <-concourseProcess.Wait():
```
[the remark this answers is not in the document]
pivotal-jamie-klassen: it shouldn't be there at all! e1e6c89b070355a9d0f54b1706ddc88c5c216514

atc/atccmd/command.go
```diff
@@ -119,8 +119,9 @@ type RunCommand struct {
 
 	Postgres flag.PostgresConfig `group:"PostgreSQL Configuration" namespace:"postgres"`
 
-	APIMaxOpenConnections     int `long:"api-max-conns" description:"The maximum number of open connections for the api connection pool." default:"10"`
-	BackendMaxOpenConnections int `long:"backend-max-conns" description:"The maximum number of open connections for the backend connection pool." default:"50"`
+	ConcurrentRequestLimits   map[wrappa.LimitedRoute]int `long:"concurrent-request-limit" description:"Limit the number of concurrent requests to an API endpoint (Example: ListAllJobs:5)"`
```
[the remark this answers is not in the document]
pivotal-jamie-klassen: Since it's a map, it can contain multiple limits, hence plural. However the way you do this is by passing the flag multiple times, so the long flag name is singular. This approach is consistent with [worker bundled resource types](https://github.com/concourse/concourse/blob/master/atc/atccmd/command.go#L159) and [extra metrics tags](https://github.com/concourse/concourse/blob/master/atc/atccmd/command.go#L164).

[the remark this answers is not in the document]
clarafu: @pivotal-jamie-klassen Ah I see, that makes sense!

PR #5431, opened 2020-04-13
Add timeout for hijacking failed containers for #3892

### LINKED ISSUE #3892: Configurable ceiling for how long containers can still be intercepted / hijacked

Containers hang around for a very long time after the build that they're related to fail.

Even though we only keep the last failed build as "interceptible" as `builds.interceptible` ends up being used as a check when determining whether the container should be marked for destruction or not, in some cases (where many many many containers are created per-build and builds are not often), the number of containers hanging around can become a problem as the number of containers that a worker has is a limiting resource.

E.g., consider `strabo` whose builds happen infrequently but, when they happen, they create a bunch of containers:

idle-containers
  

Possibly having a collector that would mark builds as non-interceptible after a configurable amount of time  🤔 

Thanks!

ebilling:
# Existing Issue
Fixes #3892.

* added timeout for hijacking failed containers

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


atc/atccmd/command.go
```diff
@@ -184,6 +184,7 @@ type RunCommand struct {
 		OneOffBuildGracePeriod time.Duration `long:"one-off-grace-period" default:"5m" description:"Period after which one-off build containers will be garbage-collected."`
 		MissingGracePeriod     time.Duration `long:"missing-grace-period" default:"5m" description:"Period after which to reap containers and volumes that were created but went missing from the worker."`
 		HijackGracePeriod      time.Duration `long:"hijack-grace-period" default:"10m" description:"Period after which hijacked containers will be garbage collected"`
+		FailedGracePeriod      time.Duration `long:"failed-grace-period" default:"4h" description:"Period after which failed containers will be garbage collected"`
```
xtremerui, line 187: @vito what is a better default value here that makes less impact to the GC?

Also @ebilling could you add a check for this flag so when it is set to '0' the failed containers of last build will still be kept as of today (i.e. user opt out this period) ? There might be users who got sufficient resource that dont mind keeping the failed containers longer. Thx!

ebilling: @xtremerui, made the change to go back to current behavior if failed-grace-period set to '0'.

vito: Hmm, I think 4 hours is too short at least. One of the reasons for keeping them forever was so that a build that failed on a Friday evening could still be debugged Monday morning. Build failures can be lightning in a bottle, so allowing the build to be debugged for longer periods seemed important at the time.

"Infinite" may be too costly for how infrequently this situation occurs though, and never being able to reproduce a failure again may in itself be feedback that your pipeline is too brittle and needs more sweeping changes. That won't help the sting go away at the time though.

@ebilling Not necessarily suggesting it, but what would be the highest theoretical value you would allow here? `72h` would allow a flake to survive a weekend, while not being infinite. 🤔

ebilling: Let's make it 96h.  That gives almost a week or a long weekend to get in.

xtremerui: Giving the case of keeping failed container to survive a weekend, I would say 24 hours after the weekend might still not be enough since Monday could be occupied by meetings plans etc. How about 120 hours (5 days) by default? So it will be available long enough for a work cycle of build failed->notified dev->planing fix->debugging->rerun build.

ebilling: Anything is better than infinite.  I'll push the change.

pivotal-jamie-klassen: just so it doesn't get forgotten (and in the interest of practicing responsible change), setting any default makes this a breaking change and it should be noted in the release notes!

release-notes/latest.md
```diff
@@ -1,9 +1,15 @@
-#### <sub><sup><a name="4950" href="#4950">:link:</a></sup></sub> feature, breaking
+##### <sub><sup><a name="5305" href="#5305">:link:</a></sup></sub> feature
+
+* We've updated the way that containers associated with failed runs get garbage collected
+
+  Containers associated with failed runs used to sit around until a new run is executed.  They now have a max lifetime (default - 4 hours), configurable via 'failed-grace-period' flag.
```
xtremerui, line 5:
```suggestion
  Containers associated with failed runs used to sit around until a new run is executed.  They now have a max lifetime (default - 120 hours), configurable via 'failed-grace-period' flag.
```

atc/atccmd/command.go
```diff
@@ -184,6 +184,7 @@ type RunCommand struct {
 		OneOffBuildGracePeriod time.Duration `long:"one-off-grace-period" default:"5m" description:"Period after which one-off build containers will be garbage-collected."`
 		MissingGracePeriod     time.Duration `long:"missing-grace-period" default:"5m" description:"Period after which to reap containers and volumes that were created but went missing from the worker."`
 		HijackGracePeriod      time.Duration `long:"hijack-grace-period" default:"10m" description:"Period after which hijacked containers will be garbage collected"`
+		FailedGracePeriod      time.Duration `long:"failed-grace-period" default:"120h" description:"Period after which failed containers will be garbage collected"`
```
xtremerui, line 187:
```suggestion
		FailedBuildGracePeriod      time.Duration `long:"failed-grace-period" default:"120h" description:"Period after which failed containers will be garbage collected"`
```

atc/db/build_factory.go
```diff
@@ -24,16 +24,18 @@ type BuildFactory interface {
 }
 
 type buildFactory struct {
-	conn              Conn
-	lockFactory       lock.LockFactory
-	oneOffGracePeriod time.Duration
+	conn                     Conn
+	lockFactory              lock.LockFactory
+	oneOffGracePeriod        time.Duration
+	maximumFailedGracePeriod time.Duration
```
xtremerui, line 30:
```suggestion
	failedGracePeriod time.Duration
```

atc/db/build_factory_test.go
```diff
@@ -68,7 +68,7 @@ var _ = Describe("BuildFactory", func() {
 			DescribeTable("completed and past the grace period",
 				func(status db.BuildStatus, matcher types.GomegaMatcher) {
 					//set grace period to 0 for this test
-					buildFactory = db.NewBuildFactory(dbConn, lockFactory, 0)
+					buildFactory = db.NewBuildFactory(dbConn, lockFactory, 0, 0)
```
xtremerui, line 71: Could you add tests under "pipeline builds" for the failed grace period similar to this one? The diff is you could have a new buildFactory with failedGracePeriod set to 0 or 1 second.

So then you could test under following situations:
build failed and check interceptable right away
build fialed and check interceptable after 1s

ebilling:
I think this is ready to be reviewed.  The TestFlight failure seems unrelated to my commits...

I have one question though and left the comment in the code.  Is there a convenient way to set the  end-time of the failed task back in time?  I prefer to not use a sleep for this test.  Especially since the timestamp is at second resolution.

xtremerui:
Nah it is hard in this case. The other way is to insert a record in DB with a given timestamp and verify the interceptable value which requires another query to the DB.

xtremerui:
Thx for the PR!

Issue #5139, opened 2020-02-07
containerd: support the use of concourse worker's DNS server


cirocosta:
### what challange are you facing

with the inability to configure the set of nameservers that containers should
use by default (see #5138), users who rely on
`CONCOURSE_GARDEN_DNS_PROXY_ENABLE` (e.g., to run `concourse` as part of a
`docker-compose` set up and reach other services in the docker network) can't
move to `containerd` because it'd never put Concourse's DNS server as a
nameserver under `/etc/resolv.conf` inside the containers.

### what would make this better

have Concourse's proxy DNS server listed as a nameserver in the set of
nameservers on the `/etc/resolv.conf` that gets injected into containers.

vito:
Sounds like we just need to change the `/etc/resolv.conf` content to include the IP address of the proxy server instead.

PR #5445, opened 2020-04-15
containerd: support custom dns configuration

### LINKED ISSUE #5138: containerd: support configuring default DNS nameservers
### what challange are you facing

the current implementation has a hardcoded set of values for `/etc/resolv.conf`:

https://github.com/concourse/concourse/blob/54991b2f1f17bd7d229a3ebbe6d6c16f3ff495a3/worker/backend/cni_network.go#L183-L186

while this is meant to work for those who rely only on external services (on ThE
InTeRwEbS), it's not something that would fit those who

- run jobs that require any form of internal DNS resolution
- do not trust 8.8.8.8

allow one to configure the set of nameservers listed under `/etc/resolv.conf`
(like `guardian` allows with the `--dns-server` flag)

https://github.com/cloudfoundry/guardian/blob/bcca70921740300398648046496790876981b158/guardiancmd/command.go#L157-L158

cirocosta:
### Existing Issue

Fixes #5138

This PR allows users to pass a custom DNS server for use in worker containers.

It does this by making CONCOURSE_GARDEN_DNS_SERVER flag have the same functionality for `containerd` runtime as it does when using `guardian`.

- [ ] Unit tests

### Reviewer Checklist

> fill out this section.
- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/worker/backend/cni_network.go
+++ b/worker/backend/cni_network.go
@@ -31,15 +31,6 @@ type CNINetworkConfig struct {
 	Subnet string
 }
 
-// defaultCNINetworkConfig is the default configuration for the CNI network
-// created to put concourse containers into.
-//
-var defaultCNINetworkConfig = CNINetworkConfig{
-	BridgeName:  "concourse0",
-	NetworkName: "concourse",
-	Subnet:      "10.80.0.0/16",
-}
-
 const (
 	// fileStoreWorkDir is a default directory used for storing
 	// container-related files
@@ -50,6 +41,21 @@ const (
 	// binaries in.
 	//
 	binariesDir = "/usr/local/concourse/bin"
+)
+
+var (
+	// nameServers is the default set of nameservers used.
+	//
+	nameServers = []string{"8.8.8.8"}
+
+	// defaultCNINetworkConfig is the default configuration for the CNI network
+	// created to put concourse containers into.
+	//
+	defaultCNINetworkConfig = CNINetworkConfig{
+		BridgeName:  "concourse0",
+		NetworkName: "concourse",
+		Subnet:      "10.80.0.0/16",
+	}
 )
 
 func (c CNINetworkConfig) ToJSON() string {
@@ -97,6 +103,15 @@ func WithCNIBinariesDir(dir string) CNINetworkOpt {
 	}
 }
 
+// WithNameServers sets the set of nameservers to be configured for the
+// /etc/resolv.conf inside the containers.
+//
+func WithNameServers(nameservers []string) CNINetworkOpt {
+	return func(n *cniNetwork) {
+		n.nameServers = nameservers
+	}
+}
+
 // WithCNIClient is an implementor of the CNI interface for reaching out to CNI
 // plugins.
 //
@@ -128,6 +143,7 @@ type cniNetwork struct {
 	client      cni.CNI
 	store       FileStore
 	config      CNINetworkConfig
+	nameServers []string
 	binariesDir string
 }
 
@@ -139,6 +155,7 @@ func NewCNINetwork(opts ...CNINetworkOpt) (*cniNetwork, error) {
 	n := &cniNetwork{
 		binariesDir: binariesDir,
 		config:      defaultCNINetworkConfig,
+		nameServers: nameServers,
 	}
 
 	for _, opt := range opts {
@@ -182,7 +199,7 @@ func (n cniNetwork) SetupMounts(handle string) ([]specs.Mount, error) {
 
 	resolvConf, err := n.store.Create(
 		filepath.Join(handle, "/resolv.conf"),
-		[]byte("nameserver 8.8.8.8"),
+		n.generateResolvConfContents(),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("creating /etc/resolv.conf: %w", err)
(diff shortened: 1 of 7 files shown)


worker/backend/cni_network.go
```diff
@@ -203,6 +220,15 @@ func (n cniNetwork) SetupMounts(handle string) ([]specs.Mount, error) {
 	}, nil
 }
 
+func (n cniNetwork) generateResolvConfContents() []byte {
+	contents := ""
+	for _, n := range n.nameServers {
+		contents = contents + fmt.Sprintf("nameserver %s\n", n)
```
aoldershaw, line 226: Nit: `n.nameServers` will never be large enough for this to have a meaningful performance impact, but could you change this to use a [`strings.Builder`](https://play.golang.org/p/Dv96oDHG0tF) or [`bytes.Buffer`](https://play.golang.org/p/b48TYKuEmiA) instead?

worker/backend/cni_network.go
```diff
@@ -52,6 +43,21 @@ const (
 	binariesDir = "/usr/local/concourse/bin"
 )
 
+var (
+	// nameServers is the default set of nameservers used.
+	//
+	nameServers = []string{"8.8.8.8"}
```
aoldershaw, line 49: Nit: for consistency with `defaultCNINetworkConfig`

```suggestion
	defaultNameServers = []string{"8.8.8.8"}
```

worker/backend/process_killer_test.go
```diff
@@ -3,10 +3,10 @@ package backend_test
 import (
 	"context"
 	"errors"
-	"fmt"
 	"math"
 	"syscall"
 	"time"
+	"fmt"
```
aoldershaw, line 9: Not sure why these tests are [failing to compile now](https://ci.concourse-ci.org/teams/main/pipelines/prs/jobs/unit/builds/899#L5e839dab:535:540), but it's claiming `fmt` is not used. Maybe run `goimports` on all the files? (I noticed some formatting issues in several files)

cirocosta:
ooh sooorrry, it was still a draft, but I forgot to marked it as such :facepalm:  will address the points for sure though! thanks!

worker/backend/process_killer_test.go
cirocosta: d01e3fd2c39929c5510c6ab2e6f43058ffe9f3eb

worker/backend/cni_network.go
cirocosta: 7e2246a76f9c6c66f2a31fc7b3faf066b7bc6d45

worker/backend/cni_network.go
cirocosta: aah we end up going with plain `+` for simplicity :grimacing:  in theory, for `strings.Builder` and `bytes.Buffer` we should check the error too

abf820a2291102912aff2cfbd98d38d57e949ead

worker/workercmd/worker_linux.go
```diff
@@ -132,6 +133,10 @@ func (cmd *WorkerCommand) gdnRunner(logger lager.Logger) (ifrit.Runner, error) {
 		"--no-image-plugin",
 	}
 
+	for _, dnsServer := range cmd.Garden.DNSServers {
+		gdnServerFlags = append(gdnServerFlags, "--dns-server", dnsServer)
```
aoldershaw, line 137: Was it not possible to set the list of DNS servers in Guardian before this?

cirocosta: goood question: it would be automagically forwarded through https://github.com/concourse/concourse/blob/aee8c5cbde7982361d109bc28a9db8a68e6f0664/worker/workercmd/worker_linux.go#L274-L277

we had to have that explicitly because the library that takes the env vars and fills the `cmd.*` structs unsets from the env, which means that just because we added there so support `containerd`, that'd helper would not get the var anymore

aoldershaw:
@cirocosta in doing PR acceptance, I noticed that there's different behaviour between Guardian and our containerd backend, and wanted to confirm this is known (and will be tackled in subsequent issues), or if there's more work to be done in this issue.

---

Using no override files (i.e. running Guardian with the default nameservers), containers have the following `/etc/resolv.conf`:

```
nameserver 172.18.0.4
```

...and are able to successfully `curl web:8080`.

Overriding `CONCOURSE_GARDEN_DNS_SERVER` env var to `"1.1.1.1"` (while still using Guardian) results in:

`/etc/resolv.conf`

```
nameserver 1.1.1.1
nameserver 172.18.0.4
```

...but now it is unable to resolve `web` (not sure why this is? I guess it fails to resolve with 1.1.1.1 but doesn't try 172.18.0.4??)

---

In contrast, using only the containerd override (i.e. setting `CONCOURSE_GARDEN_USE_CONTAINERD` to `"true"`) results in:

`/etc/resolv.conf`

```
nameserver 8.8.8.8
```

...which clearly can't resolve `web`.

Using containerd + overriding the DNS server to 1.1.1.1 results in:

`/etc/resolv.conf`

```
nameserver 1.1.1.1
```

...which also can't resolve `web`.

---

So, the feature does what it says on the tin, but doesn't seem to achieve parity with Guardian. I'm confused about where this `172.18.0.4` nameserver is coming from. Presumably, Guardian is populating it somehow, but I'm not sure what that's based on.

cirocosta:
Hey @aoldershaw, thanks for the detailed review!

That behavior would be captured by the next PR that'd come with regards to DNS resolution w/ containerd: https://github.com/concourse/concourse/issues/5139.

To tackle that, we'd end up implementing the behavior you mentioned 😁  

@muntac has been the one keeping most of the context w/ regards to the features, please correct me if I'm wrong

thanks!

aoldershaw:
@cirocosta ah, I thought I remember hearing something about that. Thanks for the clarification!

muntac:
Thanks @aoldershaw for the review, and @cirocosta for the responses. Sorry I didn't get around to responding in time.

PR #5448, opened 2020-04-16
atc: behaviour: emit tasks waiting prometheus metric

### LINKED ISSUE #5057: Provide metric for tasks queue when using limit-active-tasks placement strategy

We switched to the limit-active-tasks placement strategy and so far it solved a lot of our problems. We want to improve now by scaling our workers depending on the size of the task queue.
We are running our Concourse in a Kubernetes environment.

It would improve the scaling of workers.

Sure :)
We already saw that the first proposed implementation had this metric exported: https://github.com/concourse/concourse/pull/4612

tenjaa:
# Existing Issue
Fixes #5057

# Why do we need this PR?
Using the limit-active-tasks placement strategy gives the perfect opportunity
to dynamically scale workers based on tasks running. This metric makes
it easier for operators to implement such a scaling.

* add prometheus metric for waiting builds when using `limit-active-tasks` placement strategy 

# Contributor Checklist
- [x] Unit tests
Not sure if we need tests for Metrics. Looking through the code we didn't find any tests for metrics. Should we add some?

# Reviewer Checklist
- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [x] Release notes reviewed
- [x] PR acceptance performed


release-notes/latest.md
```diff
@@ -40,3 +40,7 @@
 #### <sub><sup><a name="5410" href="#5410">:link:</a></sup></sub> feature
 
 * We've moved the "pin comment" field in the Resource view to the top of the page (next to the currently pinned version). The comment can be edited inline.
+
+#### <sub><sup><a name="5057" href="#5057">:link:</a></sup></sub> feature
+ 
+ * @HannesHasselbring and @tenjaa added a prometheus metric of the amount of tasks that are currently waiting to be scheduled when using the `limit-active-tasks` placement strategy
```
pivotal-jamie-klassen, line 46: Always nice to link the PR that provided the change! Also I think this should work with all the metrics emitters, not that I've exhaustively tested.
```suggestion
 * @HannesHasselbring and @tenjaa added a metric for the amount of tasks that are currently waiting to be scheduled when using the `limit-active-tasks` placement strategy. #5448
```

tenjaa:
Hi @pivotal-jamie-klassen 
we added some tests and also fixed the merge conflict.

shyamz-22:
@pivotal-jamie-klassen we have added test as mentioned in the last review. However our test that test the task waiting metric is gauged correctly when waiting for worker to be available is bit [flaky](https://github.com/concourse/concourse/pull/5448/commits/82a042d7144c391513c167a514a1a98c81ca996d). It seem to succeed after 2 or 3 [attempts](https://ci.concourse-ci.org/teams/main/pipelines/prs/jobs/unit/builds/1099). Can you help us with that?
Thanks :)

pivotal-jamie-klassen:
tl;dr - making this properly testable may require some dependency-breaking refactors. Or maybe a more sophisticated stub on the `fakePool`'s `FindOrChooseWorkerForContainer` method.

@shyamz-22 I don't have time to go deeply into a solution right now, but I will come back to this and I just want to bring the challenge to the surface.

The full scenario we're trying to simulate is this:

1. the client cannot find a worker
1. the waiting tasks gauge is incremented
1. the step is aborted
1. the waiting tasks gauge is decremented

In the `JustBeforeEach`, `client.RunTaskStep` executes, and is blocking. When we look in the implementation, we are getting a race between the context being canceled and the metric being incremented. Somehow we want to wait long enough for the metric to actually be incremented before aborting the step.

For the record, I was able to get the test to consistently pass by writing it like this:

```go
Context("when waiting for worker to be available", func() {
	BeforeEach(func() {
		fakePool.FindOrChooseWorkerForContainerReturns(nil, nil)
		go func() {
			Eventually(metric.TasksWaiting.Max, 6*time.Second).Should(Equal(float64(1)))
			cancel()
		}()
	})

	It("metric task waiting is gauged and lock is released", func() {
		Expect(metric.TasksWaiting.Max()).To(Equal(float64(0)))
		Expect(fakeLock.ReleaseCallCount()).To(Equal(fakeLockFactory.AcquireCallCount()))
	})
})
```

However, I strongly **do not** recommend keeping this approach, because instead of failing it just hangs indefinitely. But it serves as an interesting illustration of how to orchestrate the concurrent processes to achieve the simulation described.

shyamz-22:
@pivotal-jamie-klassen  thanks for the reply. 

We were also in the same opinion as you. We also had an idea for the refactor to this behaviour better. Could we explore this refactor? or would you suggest a lead on how we can proceed 

thanks

pivotal-jamie-klassen:
@shyamz-22 if you have a good idea of how to refactor in that module, I'd say feel free to pursue it! Ideally the `RunTaskStep` method signature should be respected, but within that extracting and testing a collaborator independently is reasonable. After all, the complexity is really very high:

```bash
$ gocyclo atc/worker/client.go
22 worker (*client).chooseTaskWorker atc/worker/client.go:442:1
16 worker (*client).RunTaskStep atc/worker/client.go:236:1
...
```

Either way, I found another way to write your test that does the job - instead of aborting the step, the simulation can have the first attempt to find a worker fail and the second attempt succeed. This sidesteps all the race conditions and flakiness, but it's still quite a long-running test due to the hardcoded 5-second polling interval. If you can externalize and stub the ticker it could improve the running time by orders of magnitude.

My second revision:
```go
Context("when waiting for worker to be available", func() {
	BeforeEach(func() {
		fakePool.FindOrChooseWorkerForContainerReturnsOnCall(0, nil, nil)
		fakePool.FindOrChooseWorkerForContainerReturnsOnCall(1, fakeWorker, nil)
	})

	It("metric task waiting is gauged and lock is released", func() {
		Eventually(metric.TasksWaiting.Max(), 6*time.Second).Should(Equal(float64(1)))
		Eventually(metric.TasksWaiting.Max(), 6*time.Second).Should(Equal(float64(0)))
		Expect(fakeLock.ReleaseCallCount()).To(Equal(fakeLockFactory.AcquireCallCount()))
	})
})
```

shyamz-22:
@pivotal-jamie-klassen 
Hi, 

Thank you so much for the feedback. I have made some changes [here](https://github.com/volkswagen/concourse/tree/draft-waiting-tasks-refactor).

Please consider this as a draft.

## Changes:

1. https://github.com/volkswagen/concourse/blob/0f7d6d1040658596568ba9b6f28fb8a65370041d/atc/worker/client.go#L557
2. https://github.com/volkswagen/concourse/blob/0f7d6d1040658596568ba9b6f28fb8a65370041d/atc/worker/client.go#L112
3. https://github.com/volkswagen/concourse/blob/0f7d6d1040658596568ba9b6f28fb8a65370041d/atc/atccmd/command.go#L648
4. https://github.com/volkswagen/concourse/blob/0f7d6d1040658596568ba9b6f28fb8a65370041d/atc/atccmd/command.go#L876
5. https://github.com/volkswagen/concourse/blob/0f7d6d1040658596568ba9b6f28fb8a65370041d/atc/atccmd/command.go#L149
6. https://github.com/volkswagen/concourse/blob/draft-waiting-tasks-refactor/atc/worker/choose_task_worker_test.go

Some differences to the implementation and the draft here:

Previously the order of execution was
1. Acquire lock
2. Find Existing container
3. Choose worker

However the steps 1 and 2 are needed only when a worker is found. There are no tests to prove that the order matters. The current refactor translates roughly to the following (omitted release lock, you can find it in the code)

1. Choose Worker
2. Acquire lock
3. If worker is found increase active task and exit
4. If worker is not found
   a. increase waiting task metric
   b. wait for worker to be available and continue

## Observations:

- Tests are not flaky and faster [~2seconds]
- current gocyclo score is 

```bash
➜  concourse git:(draft-waiting-tasks-refactor) gocyclo atc/worker/client.go
16 worker (*client).RunTaskStep atc/worker/client.go:249:1
12 worker (*client).chooseTaskWorker atc/worker/client.go:557:1
```

> I also reordered the code in the order struts declaration, vars come first followed by Exported, private Receiver methods and at last functions 

Please let me know how to proceed further.

pivotal-jamie-klassen:
@shyamz-22 I quite like the reorganizing of that file, as well as the various `*Dummy()` methods in the test. The test is also much easier to read. Just for fun, I checked out the coverage, and while `atc/worker/client.go` has ~80% coverage, all of the functions involved in `RunTaskStep` are quite thoroughly covered.

I had one concern, and this mostly shows the limited degree to which I actually understand the affected components of Concourse. By acquiring the lock after choosing the worker, I can imagine this reducing the thread-safety of the worker-choosing routine. Since the worker-choosing part is no longer synchronized, can multiple web nodes race each other and do anything weird?

I think the answer is no, because the only code path that actually calls `RunTaskStep` is the one involving the `Run` method on a `TaskStep`, which ultimately only gets called from `engineBuild.Run`. This method itself involves acquiring a tracking lock. Since there can only be a single goroutine that possesses the tracking lock for a build, there is no risk of two webs accidentally scheduling the same task on different workers. It seems like the active task lock is really just to avoid two webs racing to increment/decrement the active tasks on a worker, and having that number potentially go out of sync with the actual number of task containers on the worker.

Since you are making the retry and reporting interval configurable, there are some jobs in our pipeline that will complain unless the appropriate configuration is exposed via the [helm chart](https://github.com/concourse/concourse-chart) and [bosh release](https://github.com/concourse/concourse-bosh-release). So it would be helpful to make PRs to those repos as well, though I can help with that. Otherwise, the configuration doesn't actually need to be passed all the way to the operator. One could still externalize them from the method under test but have them always be constant in production. I suggest considering this because it seems like the kind of configuration an operator would have no idea how or why to tune -- indeed I can't think of why I would ever change that variable! Maybe you can?

shyamz-22:
> One could still externalize them from the method under test but have them always be constant in production. 

My confusion to achieve this are:

1. chooseTaskWorker is not exported and is only called from the `RunTaskStep` which is an interface method. This method is the method under test which means the signature has to be modified.

2. If we Export the chooseTaskWorker method on the client, The `client` struct is again not exported and is implementing the Client interface

3. All the test packages are having `_test` and I do not want to add a test with the same package name as the production code breaking the convention.

This is why I decided to pass it as a configuration.  I cannot think of any other way as my understanding of all the components and code convention of Concourse is not strong

pivotal-jamie-klassen:
@shyamz-22 sorry, I was waving my hands there. What I mean was that instead of [adding these parameters](https://github.com/volkswagen/concourse/commit/0f7d6d1040658596568ba9b6f28fb8a65370041d#diff-67f9436106f0941622f0345dbf1c82a2R149-R150) you could make a change around here:

https://github.com/volkswagen/concourse/blob/0f7d6d1040658596568ba9b6f28fb8a65370041d/atc/atccmd/command.go#L86-L94

adding some lines defining constants:

```go
var workerAvailabilityPollingInterval = 5 * time.Second
var workerStatusPublishInterval = 1 * time.Minute
```

and then the code here:

https://github.com/volkswagen/concourse/blob/0f7d6d1040658596568ba9b6f28fb8a65370041d/atc/atccmd/command.go#L648

could instead look like

```go
	workerClient := worker.NewClient(pool, workerProvider, compressionLib, workerAvailabilityPollingInterval, workerStatusPublishInterval)
```

and similarly [here](https://github.com/volkswagen/concourse/commit/0f7d6d1040658596568ba9b6f28fb8a65370041d#diff-67f9436106f0941622f0345dbf1c82a2R876-R880).

Does this make sense? I'm also open to making those things operator-configurable if you think there's a precedent, I just can't think of any.

shyamz-22:
@pivotal-jamie-klassen I also cannot find a valid reason why an operator would need to configure the polling interval. I will do as you suggested by adding defaults and merge the changes. Hopefully this  closes the PR 🤞 😄

shyamz-22:
@pivotal-jamie-klassen thank you to you too for providing enough information in each review really appreciate it :)

PR #5451, opened 2020-04-16
atc/db: write operations to an archived pipeline fail

### LINKED ISSUE #5430: api/fly: write operations to an archived pipeline should fail
**Epic:** #5055

Generally speaking, given a pipeline is archived, the majority of write operations done via the API should fail. Below goes in full detail.

**Given**
```
$ fly -t dev set-pipeline -p test -c topgun/pipelines/custom-types.yml
$ fly -t dev archive-pipeline -p test
```

**Then**
* [x] API action `GetConfig` will `404`; when a pipeline is archived its config is removed to avoid leaking sensitive information.
  * Currently:
    ```
    HTTP/1.1 200 OK
		{"config":{"jobs":[{"name":"","plan":null}]}}
    ```
  * Desired:
    ```
    $ fly -t dev get-pipeline -p test
    <some error>
    ```
* [x] API action `CreateJobBuild` will error; archived pipelines must consume no scheduling resources, not even the build starter.

  * Currently:
    ```
    CreateJobBuild

    POST /api/v1/teams/main/pipelines/archived-2/jobs/test/builds HTTP/1.1
    2020/04/13 12:22:01 HTTP/1.1 200 OK
    ```
  * Desired:
    ```
    $ fly -t dev trigger-job -j test/job
    <some error>
    ```
* [x] API action `ScheduleJob` will error; archived pipelines must consume no scheduling resources, not even the build starter.

  * Currently:
    ```
    PUT /api/v1/teams/main/pipelines/archived-2/jobs/test/schedule HTTP/1.1
    HTTP/1.1 200 OK
    ```
  * Desired:
    ```
    $ fly -t dev schedule-job -j test/job
    <some error>
    ```
* [x] `CheckResource` and `CheckResourceType` will error; archived pipelines must consume no scheduling resources, including queuing checks.

  * Currently:
    ```
    CheckResource

		POST /api/v1/teams/main/pipelines/archived-2/resources/interval/check HTTP/1.1
    {"from":null}
		HTTP/1.1 201 Created

		CheckResourceType

    POST /api/v1/teams/main/pipelines/archived-2/resource-types/time/check HTTP/1.1
    {"from":null}
    HTTP/1.1 201 Created

    ```
  * Desired:
    ```
    $ fly -t dev check-resource -r test/resource-name
    <some error>
    $ fly -t dev check-resource-type -r test/resource-name
    <some error>
    ```
* [x] `PinResourceVersion`, `UnpinResource`, `EnableResourceVersion`, `DisableResourceVersion`, and `SetPinCommentOnResource` will error; archived pipelines are read-only.
	all should probably be revisited:
  * Currently (only tested one, assumed the rest behave the same):
    ```
    PinResourceVersion

    PUT /api/v1/teams/main/pipelines/archived-2/resources/interval/versions/3/pin
    HTTP/1.1 200 OK
    ```
  * Desired:
    ```
    $ fly -t dev pin-resource -r test/resource-name -v version --comment "some comment"
    <some error>
    $ fly -t dev unpin-resource -r test/resource-name
    <some error>
    $ fly -t dev enable-resource-version -r test/resource-name -v version
    <some error>
    $ fly -t dev disable-resource-version -r test/resource-name -v version
    <some error>
    ```

taylorsilva:
Existing Issue

Fixes #5430 .

* Return a 409 HTTP status code for certain write operations on archived pipelines
* Return a db.Conflict error at the db level for these operations
* TODO: handle 409s in fly to print nicer error messages

# Contributor Checklist
- [x] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/job.go
+++ b/atc/db/job.go
@@ -896,6 +896,17 @@ func (j *job) RequestSchedule() error {
 
 	defer tx.Rollback()
 
+	pipeline, ok, err := j.pipeline(tx)
+	if err != nil {
+		return err
+	}
+	if !ok {
+		return ErrPipelineDisappeared
+	}
+	if pipeline.Archived() {
+		return conflict("jobs part of an archived pipeline cannot be scheduled")
+	}
+
 	err = requestSchedule(tx, j.id)
 	if err != nil {
 		return err
--- a/atc/db/job.go
+++ b/atc/db/job.go
@@ -3,6 +3,7 @@ package db
 import (
 	"database/sql"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"sort"
 	"strconv"
@@ -14,6 +15,8 @@ import (
 	"github.com/concourse/concourse/atc/db/lock"
 	"github.com/lib/pq"
 )
+
+var ErrPipelineDisappeared = errors.New("pipeline disappeared")
 
 type InputVersionEmptyError struct {
 	InputName string
@@ -656,6 +659,17 @@ func (j *job) CreateBuild() (Build, error) {
 	}
 
 	defer Rollback(tx)
+
+	pipeline, ok, err := j.pipeline(tx)
+	if err != nil {
+		return nil, err
+	}
+	if !ok {
+		return nil, ErrPipelineDisappeared
+	}
+	if pipeline.Archived() {
+		return nil, conflict("jobs part of an archived pipeline cannot be triggered")
+	}
 
 	buildName, err := j.getNewBuildName(tx)
 	if err != nil {
--- a/atc/api/jobs_test.go
+++ b/atc/api/jobs_test.go
@@ -1480,7 +1480,19 @@ var _ = Describe("Jobs API", func() {
 						fakeJob.DisableManualTriggerReturns(false)
 					})
 
-					Context("when triggering the build fails", func() {
+					Context("when triggering the build fails due to a db conflict", func() {
+						BeforeEach(func() {
+							conflict := new(dbfakes.FakeConflict)
+							conflict.ConflictReturns("some conflict")
+							fakeJob.CreateBuildReturns(nil, conflict)
+						})
+						It("returns a 409", func() {
+							Expect(response.StatusCode).To(Equal(http.StatusConflict))
+							body, _ := ioutil.ReadAll(response.Body)
+							Expect(body).To(Equal([]byte("some conflict\n")))
+						})
+					})
+					Context("when triggering the build fails with an unexpected error", func() {
 						BeforeEach(func() {
 							fakeJob.CreateBuildReturns(nil, errors.New("nopers"))
 						})
--- a/atc/api/jobs_test.go
+++ b/atc/api/jobs_test.go
@@ -2560,6 +2560,20 @@ var _ = Describe("Jobs API", func() {
 						Expect(response.StatusCode).To(Equal(http.StatusInternalServerError))
 					})
 				})
+
+				Context("when scheduling fails due to a db conflict", func() {
+					BeforeEach(func() {
+						conflict := new(dbfakes.FakeConflict)
+						conflict.ConflictReturns("some conflict")
+						fakeJob.RequestScheduleReturns(conflict)
+					})
+
+					It("returns a 409 and an error", func() {
+						Expect(response.StatusCode).To(Equal(http.StatusConflict))
+						body, _ := ioutil.ReadAll(response.Body)
+						Expect(body).To(Equal([]byte("some conflict\n")))
+					})
+				})
 			})
 		})
 
--- a/atc/api/jobserver/create_build.go
+++ b/atc/api/jobserver/create_build.go
@@ -33,6 +33,11 @@ func (s *Server) CreateJobBuild(pipeline db.Pipeline) http.Handler {
 		}
 
 		build, err := job.CreateBuild()
+		if conflict, ok := err.(db.Conflict); ok {
+			logger.Error("failed-to-create-job-build", err)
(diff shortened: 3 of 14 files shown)


atc/db/job.go
```diff
@@ -657,6 +660,17 @@ func (j *job) CreateBuild() (Build, error) {
 
 	defer Rollback(tx)
 
+	pipeline, ok, err := j.pipeline(tx)
```
clarafu, line 663: Sorry if I'm commenting too early because this is still a draft 😅 I'm wondering if we should be guarding against write operations at a higher level rather than in the db suite? So what it would look like is that in the API handler for CreateJobBuild, before it calls CreateBuild 

https://github.com/concourse/concourse/blob/6eb64cdab4f4dee855ec5cd7c29ff3a607d7e94f/atc/api/jobserver/create_build.go#L35

it will check to make sure the pipeline is not archived first? I'm considering this because it feels like low level operations like creating a build for a job should not need to care about logic like archived pipelines cannot create builds. But I'm not too sure, wanted to hear your thoughts!

taylorsilva: Never too early for feedback!
I agree with you and we did it this way first. It lives here: https://github.com/concourse/concourse/commit/4a1693919c060904e609ffc2ad083f311908f5db

@aoldershaw and I had a lengthy debate about which method to go with: api or db layer. We saw merits to both and @pivotal-jamie-klassen pointed out that what we want is a policy layer. We kept debating about this and it came down to this: 

There are other components that interact directly with the `db` besides the API (e.g. lidar). This means any component that interacts with the `db`, like the API, potentially needs to be aware that a pipeline can be archived and guard against it. By guarding for this at the db layer we ensure other components don't need to worry about whether a pipeline is archived or not.

While in the middle of writing this version of the implementation though and I really don't like all the `if pipline.Archived()` that we're sprinkling everywhere 😞 Curious about your thoughts and anyone else's as well!

clarafu: Ya that perspective makes a lot of sense. I think the only thing I would push back with is the fact that I don't see too much difference between a paused pipeline and an archived pipeline? Maybe I'm missing something but I would see implementing archiving be really similar to how we deal with paused pipelines, where in the scheduler we would grab all unpaused pipelines and in lidar we grab all unpaused resources, etc.

I do know that one difference between the paused and archived is that paused pipelines can still accept requests that modify it's state (like creating a new build) but archived can't. Right, I personally would prefer not to have the `if pipeline.Archived()` implemented everywhere in the individual db methods? 🤔 Mostly because the high level component already knows about the pipelines and the way that we can think of each component could be something like this for the scheduler: "it grabs all unpaused and unarchived pipelines to schedule". Also, it can do the guarding earlier rather than have each db method that commits the write operations need to all understand the fact that archived pipelines cannot do the operation.

aoldershaw: I think part of the reason we wanted to implement it at the db layer is that it felt safer given that we weren't completely sure of all the internal components that would rely on this behaviour. Granted, you have a fair bit more context on these components than I do, so if you feel it's safe to add the conditions at a higher level, I'm sure you're right.

I can see how checking `pipeline.Archived()` in the higher-level components would be beneficial/necessary in some cases, so having the condition at the db level could be overkill. 
That said, it does feel like more cognitive overhead for the higher-level components to have to ensure they don't perform any illegal state changes - enforcing this at the db ensures we won't get into a bad state.

cirocosta: +1 on going with a higher level. 

From my point of view (of someone who baaarely works with the db package) is
that given that we've chosen to treat what goes into that package more as a
"this is how we persist & retrieve, and nothing more", leaving invariants to be
dealt with at higher levels, it seems to me that it'd indeed be more benefical
to follow the pattern we already have in place, and perform those checks there,
rather than having now to think about checks being performed *both* at higher
levels and db package.

not saying that there's no place for adopting a different paradigm where we
leave the validations to happen at the moment of getting down to persistence, I
just think that's not the approach we've chosen so far, and it'd seem a bit
confusing to now introduce just for archiving

cirocosta: with regards to how to perform the checks at that higher level,
`api_auth_wrappa` seems to be somewhat related to (from my understanding) what
@pivotal-jamie-klassen talked about when he mentions a "policy checking layer":

https://github.com/concourse/concourse/blob/421cf38c0d6df9576ff2e7d254f9f121c364a3af/atc/wrappa/api_auth_wrappa.go#L30-L56

which is ultimately the same as what @taylorsilva's mentioned (the other
approach they've tried before) but instead of individually wrapping the handlers
in `handler.go`, having a wrappa that deals in an action-by-action basis (which
might make the per-action visibility greater? might be worth trying)

clarafu: @aoldershaw I think I might argue that having the guard be at the database level might cause more cognitive overhead and more prone to refactor errors. But honestly this might just be an opinion thing, where theres good and bad to both and we just have to weigh our options. 😆 

At least this is how I think about it, if you think about a component like the scheduler it kind of "makes sense" that we don't need to schedule jobs that are from an archived pipeline (just like how we don't want to schedule job from a paused pipeline) and that is easily understandable, plus that logic most likely will not be touched often with contributors implementing new features. But if the guard was at the database level, when a contributor needs to implement a new feature that does write operations to the pipeline/job/resources tables, it would require them to know not to execute the write if the pipeline is archived.

But I definitely agree with what @cirocosta is saying, how we currently do have most of the logic around guards being on the higher-level components. I think if we want to start doing guarding in the db level, we should probably start making everything else follow the same pattern.

PR #5453, opened 2020-04-17
Remove duplicate migrations that not postgres 9.5 friendly

### LINKED ISSUE #5391: Launching/Migrating Concourse 6.0 fails with Postgres syntax errors
## Bug Report

I'm running Concourse via Docker Compose on a CentOS 7 EC2 instance. Here's my [`docker-compose.yaml`](https://gist.github.com/tprobinson/66eb6cbb4e4a45b8b5022c2a0262a69e).

I was unable to launch Concourse 6.0 with a Postgres 9.5 database, only a clean Postgres 12.2 database worked.

## Steps to Reproduce

* Run the linked Docker Compose document (`docker-compose up -d`)
* Change the `web` service's image version to 6.0
* `docker-compose up -d` again to update services

## Expected Results

Migrations to succeed, and Concourse to launch at version `6.0`

## Actual Results

I get the following errors when going to 6.0 via:
* 5.6 -> 6.0
* 5.6 -> 5.8.1 -> 6.0
* 6.0 with a fresh Postgres 9.5 database

```
db_1   | ERROR:  syntax error at or near "NOT" at character 41
db_1   | STATEMENT:  ALTER TABLE pipelines
db_1   |            ADD COLUMN If NOT EXISTS "last_updated" timestamp with time zone NOT NULL DEFAULT '1970-01-01 00:00:00'
web_1  | {"timestamp":"2020-03-31T00:11:37.295126928Z","level":"info","source":"atc","message":"atc.cmd.start","data":{"session":"1"}}
db_1   | ERROR:  syntax error at or near "NOT" at character 41
db_1   | STATEMENT:  ALTER TABLE pipelines
db_1   |            ADD COLUMN If NOT EXISTS "last_updated" timestamp with time zone NOT NULL DEFAULT '1970-01-01 00:00:00'
web_1  | {"timestamp":"2020-03-31T00:11:37.338327768Z","level":"info","source":"atc","message":"atc.cmd.finish","data":{"duration":74521,"session":"1"}}
web_1  | error: failed to migrate database: 1 error occurred:
web_1  |        * Migration '1583868323_add_pipelines_last_update.up.sql' failed: 2 errors occurred:
web_1  |        * Transaction ALTER TABLE pipelines
web_1  |     ADD COLUMN If NOT EXISTS "last_updated" timestamp with time zone NOT NULL DEFAULT '1970-01-01 00:00:00' failed, rolled back the migration
web_1  |        * pq: syntax error at or near "NOT"
```

If I try with Postgres at version 12.2, I get the following errors when going to 6.0 via:
* 5.6 -> 6.0
* 5.6 -> 5.8.1 -> 6.0

```
web_1  | {"timestamp":"2020-03-31T00:37:00.019293662Z","level":"info","source":"atc","message":"atc.cmd.finish","data":{"duration":65339,"session":"1"}}
concourse_web_1 exited with code 1
web_1  | {"timestamp":"2020-03-31T00:37:26.684459134Z","level":"info","source":"atc","message":"atc.cmd.start","data":{"session":"1"}}
db_1   | 2020-03-31 00:37:26.727 UTC [123] ERROR:  there is no unique constraint matching given keys for referenced table "builds"
db_1   | 2020-03-31 00:37:26.727 UTC [123] STATEMENT:  CREATE TABLE build_pipes (
db_1   |            "from_build_id" integer NOT NULL REFERENCES builds (id) ON DELETE CASCADE,
db_1   |            "to_build_id" integer NOT NULL REFERENCES builds (id) ON DELETE CASCADE
db_1   |          )
web_1  | {"timestamp":"2020-03-31T00:37:26.732123926Z","level":"info","source":"atc","message":"atc.cmd.finish","data":{"duration":90902,"session":"1"}}
web_1  | error: failed to migrate database: 1 error occurred:
web_1  |        * Migration '1579713177_create_table_build_pipes.up.sql' failed: 2 errors occurred:
web_1  |        * Transaction CREATE TABLE build_pipes (
web_1  |     "from_build_id" integer NOT NULL REFERENCES builds (id) ON DELETE CASCADE,
web_1  |     "to_build_id" integer NOT NULL REFERENCES builds (id) ON DELETE CASCADE
web_1  |   ) failed, rolled back the migration
web_1  |        * pq: there is no unique constraint matching given keys for referenced table "builds"
web_1  |
```

The only way I've launched 6.0 successfully is with a clean Postgres 12.2 database.

Let me know if there's any more information I can provide or tests I can perform. Thanks!

## Version Info

* Concourse version: 5.6, 5.8.1
* Infrastructure/IaaS: EC2

xtremerui:
# Existing Issue
Fixes #5391.

# Why do we need this PR?
To support postgres 9.5

The migration that caused the syntax error is a duplicate to fix an earlier migration error. We could just remove the migration.

# Contributor Checklist

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/migration/migrations/1583868323_add_pipelines_last_update.down.sql
+++ b/atc/db/migration/migrations/1583868323_add_pipelines_last_update.down.sql
@@ -1,4 +1,2 @@
 BEGIN;
-  ALTER TABLE pipelines
-    DROP COLUMN "last_updated";
 COMMIT;
--- a/atc/db/migration/migrations/1583868323_add_pipelines_last_update.up.sql
+++ b/atc/db/migration/migrations/1583868323_add_pipelines_last_update.up.sql
@@ -1,4 +1,2 @@
 BEGIN;
-  ALTER TABLE pipelines
-    ADD COLUMN If NOT EXISTS "last_updated" timestamp with time zone NOT NULL DEFAULT '1970-01-01 00:00:00';
 COMMIT;
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: '3'
 
 services:
   db:
-    image: postgres
+    image: postgres:${POSTGRES_TAG:-latest}
     shm_size: 1gb
     ports:
     - 6543:5432


xtremerui:
Is there a way to ignore this error in pr unit tests?
```
pr removed migrations:
1583868323_add_pipelines_last_update.down.sql
1583868323_add_pipelines_last_update.up.sql
and prs cannot remove migrations.

```

vito:
@xtremerui Maybe we should just make the file contents empty or comment out everything. Removing migrations does feel a little weird for ones that have already run. Not sure how our migrations package might deal with that now or in the future.

pivotal-jamie-klassen:
Could do `NULL;` or maybe something that will emit a message to the postgres logs https://www.postgresql.org/docs/9.5/plpgsql-statements.html#PLPGSQL-STATEMENTS-NULL

xtremerui:
@pivotal-jamie-klassen seems `NULL:` failed the unit tests.
Doing just
```
BEGIN;
COMMIT;
```
works though. So do we have a preference between no-op or raise a log?

pivotal-jamie-klassen:
@xtremerui I would prefer to see a message show up in the postgres logs, just in case of disaster. It's such a strange thing to have this no-op migration -- operators might be confused and I think it would be helpful to have the behaviour be self-documenting rather than requiring operators to look up the relevant GitHub issues that really only have to do with the internal problems faced by the Concourse team.

I'd love to see a `LOG` level message saying something like
```
Running no-op placeholder migration for adding the 'last_update' column to the 'pipelines' table. The column should already exist; this redundant migration was added for the sake of some internal environments used during development, where migrations may have been run out of order.
```

Maybe this is overkill - I'll defer to @vito's judgement.

vito:
@pivotal-jamie-klassen Hm, that would result in every new Concourse installation going forward printing a log message (at least until we eventually squash the migrations, which requires careful coordination). I would rather just comment it out or make it a no-op; a year from now I'd rather forget about this than have new users notice an intriguing log message for the foreseeable future. Making it a no-op should be fine, I'm pretty sure we've done something similar in the past with buggy migrations.

Issue #4983, opened 2020-01-08
Sync concourse/dex with upstream Dex


xtremerui:
The goal is to adopt changes from https://github.com/dexidp/dex as much as possible.

Meanwhile we could re-visit PRs in our fork to make sure they are not obslete.

xtremerui:
Notice https://github.com/concourse/dex/commit/5d302f25b524965d1b5ee677ab78ce2f35baf0ba in pr/oidc-groups-and-tls is covered by https://github.com/dexidp/dex/commit/8c1716d3568400f6f629822891149e1c398f194b

pivotal-jwinters:
I believe those two actually target two different claims.

dexidp/dex@8c1716d is the `name` claim.
concourse/dex@5d302f2 is the `username` claim.

pivotal-jwinters:
Before we do all this work I think the main starting point would be to get upstream dex to support the `client_credentials` and `password` grant types. Because until they do that we'll never be able to switch off our fork.

xtremerui:
- [x] `password grant` PR https://github.com/dexidp/dex/pull/1621
 - [ ] `client credential` PR https://github.com/dexidp/dex/pull/1629
 - [ ] `cf connector` PR https://github.com/dexidp/dex/pull/1624
----------------
 - `oidc and tls` 
 - [x] [configurable groups key PR](https://github.com/dexidp/dex/pull/1631)
 - [x] [TLS configure PR](https://github.com/dexidp/dex/pull/1632)
 - [x] [username claim PR](https://github.com/dexidp/dex/pull/1634)
----------------
 - [x] `generic oauth connector` PR https://github.com/dexidp/dex/pull/1630
 - [x] `http filesystem` PR https://github.com/dexidp/dex/pull/1416
 - [x] `optional prometheus and logger` PR https://github.com/dexidp/dex/pull/1625
----------------
- [x] `bcrypt client secrets` https://github.com/dexidp/dex/pull/1861
- [x] `static public key` https://github.com/dexidp/dex/pull/1802

xtremerui:
Regarding the fact that certain PRs will not be merged into upstream Dex in near future, we need to prepare a plan for switching back to upstream dex WITH deprecating some features in current Concourse. i.e.

 * oauth connector
 * cloudfoundry connector
 * tls config for oidc connector
 * brypt client secret
 * client credential grant type

[...]

PR #5459, opened 2020-04-17
skymarshal: bcrypt client secrets

### LINKED ISSUE #5428: Bcrypt client secrets in dex

pivotal-jwinters:
Fixes #5428 

- you can now configure dex with bcrypted client secrets
- if they are in plaintext concourse will bcrypt them and then pass them to dex.

corresponding dex commit:
https://github.com/concourse/dex/commit/dcbe94f

### DIFF
--- a/skymarshal/dexserver/dexserver.go
+++ b/skymarshal/dexserver/dexserver.go
@@ -74,10 +74,10 @@ func NewDexServerConfig(config *DexConfig) (server.Config, error) {
 		}
 	}
 
-	for clientId, clientSecret := range config.Clients {
+	for clientId, clientSecret := range newLocalClients(config) {
 		clients = append(clients, storage.Client{
 			ID:           clientId,
-			Secret:       clientSecret,
+			Secret:       string(clientSecret),
 			RedirectURIs: []string{config.RedirectURL},
 		})
 	}
@@ -215,5 +215,34 @@ func newLocalUsers(config *DexConfig) map[string][]byte {
 	}
 
 	return users
+
+}
+
+func newLocalClients(config *DexConfig) map[string][]byte {
+	clients := map[string][]byte{}
+
+	for clientID, clientSecret := range config.Clients {
+		if clientID != "" && clientSecret != "" {
+
+			var hashed []byte
+
+			if _, err := bcrypt.Cost([]byte(clientSecret)); err != nil {
+				if hashed, err = bcrypt.GenerateFromPassword([]byte(clientSecret), 0); err != nil {
+
+					config.Logger.Error("bcrypt-client-secret", err, lager.Data{
+						"client_id": clientID,
+					})
+
+					continue
+				}
+			} else {
+				hashed = []byte(clientSecret)
+			}
+
+			clients[clientID] = hashed
+		}
+	}
+
+	return clients
 
 }
--- a/go.mod
+++ b/go.mod
@@ -20,7 +20,7 @@ require (
 	github.com/cloudfoundry/go-socks5 v0.0.0-20180221174514-54f73bdb8a8e // indirect
 	github.com/cloudfoundry/socks5-proxy v0.0.0-20180530211953-3659db090cb2 // indirect
 	github.com/concourse/baggageclaim v1.6.5
-	github.com/concourse/dex v0.0.0-20191211162806-d04c1742d527
+	github.com/concourse/dex v0.0.0-20200417202922-dcbe94f28c88
 	github.com/concourse/flag v1.0.0
 	github.com/concourse/go-archive v1.0.1
 	github.com/concourse/retryhttp v1.0.2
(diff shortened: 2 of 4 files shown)


aoldershaw:
@pivotal-jwinters will the changes in https://github.com/concourse/dex/commit/dcbe94f make it more difficult to get to get our dex fork merged? It seems to me like it's a breaking change - correct me if I'm wrong.

Getting off our fork may also be a low-enough priority that keeping the changes to dex small and clean is more important than making that behaviour configurable, though.

pivotal-jwinters:
@aoldershaw yup you're totally right, it is breaking for upstream dex, but not for us, since we reconfigure all clients when concourse starts up. When we submit this to upstream we'll probably have to add some sort of backwards compatibility check for non-bcrypted clients, however I don't want to block our next release on figuring that out.

skymarshal/dexserver/dexserver.go
```diff
@@ -217,3 +217,32 @@ func newLocalUsers(config *DexConfig) map[string][]byte {
...
+			var hashed []byte
+
+			if _, err := bcrypt.Cost([]byte(clientSecret)); err != nil {
+				if hashed, err = bcrypt.GenerateFromPassword([]byte(clientSecret), 0); err != nil {
+
+					config.Logger.Error("bcrypt-client-secret", err, lager.Data{
+						"client_id": clientID,
+					})
+
+					continue
```
aoldershaw, line 236: I wonder if it makes sense to just log an error here. Admittedly, I'm not sure if/when this error would ever occur in practice, but maybe it would make sense to short circuit instead?

If we fail to create a client [here](https://github.com/concourse/concourse/pull/5459/files#diff-097bd19334894da24754e934c93e4638R159-R161), we return an error - I would expect the same behaviour for failing to generate the hash (since we'd end up without that client in that case).

WDYT? My guess is this error is very unlikely to occur in practice, so this may be a moot point.

xtremerui:
tagging [comment](https://github.com/concourse/concourse/pull/5459#issuecomment-616633651) to #4983 so when creating a PR for upstream dex we don't forget about adding the backward-compatibility support

skymarshal/dexserver/dexserver.go
pivotal-jwinters: I honestly didn't give it much thought. I basically copied the same implementation we had for user passwords. I can't remember how we originally settled on logging the error and continuing, it does seem a bit weird, but I think it was for a reason.

aoldershaw:
@pivotal-jwinters the code changes look good to me. Could you possibly give me some pointers on how to do acceptance on this, though? What's an example of a "local client" that I could try to authenticate with?

pivotal-jwinters:
The first thing you can check is looking at the `client` table in the DB, and seeing all the secrets are bcrypted.

Every form of logging into concourse uses a client. TSA -> ATC uses the `concourse-worker` client, logging in using the web ui uses the `concourse-web` client, and logging in with fly (using the -u -p flags) uses the `fly` client. So acceptance is making sure all those things still work. 

If you're really industrious you could even try overriding the `CONCOURSE_CLIENT_SECRET` and `CONCOURSE_TSA_CLIENT_SECRET` and make sure that any custom values passed in can be used to login, and also show up bcrypted in the db.

aoldershaw:
@pivotal-jwinters I took a look in the `client` DB table and modified `CONCOURSE_CLIENT_SECRET` and `CONCOURSE_TSA_CLIENT_SECRET` to be both plain text and a bcrypted hash.

When I set it to plain text, it bcrypts it in the DB, and I'm still able to login.

However, when I set it to a bcrypted hash, it's unchanged in the DB, but I'm getting `"Invalid client credentials"` on login.

I may have messed this part up - it's a bit finicky since `docker-compose` wants to interpolate `$` in environment variables. I added the following line to `docker-compose.yml`:

```
services:
  web:
    environment:
      ...
      CONCOURSE_CLIENT_SECRET: '$$2a$$10$$unh5UlJ7yFUlFmKZpYmPq.2Qdj0VpLshH748n8hdzJMLwd7uQTvl.'
```

In the `client` table, I see:
```
id              secret
concourse-web   $2a$10$unh5UlJ7yFUlFmKZpYmPq.2Qdj0VpLshH748n8hdzJMLwd7uQTvl.
```

which seems to match my unescaped raw secret. Any ideas what's going wrong?

pivotal-jwinters:
@aoldershaw ah yeah damn. 

It's because we do some fancy configuration behind the scenes for the `web` cmd. These values get passed to multiple places to make configuration easier for the end user: the web (which needs them in plaintext because it uses it for actually logging in), and dex (which does the whole bcrypt thing because it stores them in the db, and validates login attempts). 

So we can't actually support the bcrypted flow on the `web` command (but we can on the `atc` cmd). However, users aren't able to configure the `atc` cmd on its own, so effectively they can't use the bcrypted client secrets (yet). 

I would still consider this feature "complete" from the dex perspective (there are unit tests to cover the already bcrypted clients). But I wouldn't worry about acceptance testing that code path until we actually expose it to end users.

Hope this isn't too confusing...

Issue #3958, opened 2019-06-01
Emit build events as traces in order to visualize the behavior of complex jobs and pipelines


njbennett:
## What challenge are you facing?

The Cloud Foundry, PAS, and PKS release engineering teams have complex pipelines, and it's hard to understand their behavior holistically. Many jobs within the pipelines are "flaky" but it's difficult to prioritize work to reduce intermittent failures without knowing which intermittent failures are the most common. 

It's also difficult for us to understand the effects of changes to our pipelines. Concourse is our implementation layer, but we're using it to model high level concepts. We care about things like a "run" of a pipeline, or the status of a release train that might represent several partial pipeline runs.

We would like to be able to answer questions like: 
- "Which jobs fail the most?"
- "Which job failures are the most common?" 
- "Which tasks shared across many jobs fail the most?"
- "Which task failures are the most common?"
- "How are many failures in our pipeline are due to underlying problems with Concourse, and how many failures are due to problems with the software under test?"
- "How long does it take a particular version of a particular resource to make it from the start of the pipeline to the finish?"
- "Which jobs have a particular version of a resource passed, which jobs have a particular version of a resource failed, and what was the cause of those failures?"

Answering these questions would help us prioritize work to reduce intermittent failures, identify and report bugs to the Concourse team, and measure the impact of changes to our pipelines on their performance.

We can do some of this by exploring the GUI or scraping data from the Concourse API. The former is time-consuming, the latter requires building relatively fragile code that will break whenever that API changes. And, as I understand the API's authentication model, it requires a credential representing a user to access the most interesting data, and is difficult to poll with a machine.

We've been able to answer questions like this with fairly high precision about our acceptance tests for some time now, which has made it easier to keep intermittent failures from that source under control. However, many of the failures we're interested in occur in other steps of our release engineering process, especially deploy and teardown.

## What would make this better?

I'd like to be able to generate Honeycomb build events from Concourse tasks, and chain them together into traces.

These events would ideally contain:
- A span ID representing a particular PAS/cf-deployment/PKS version, so we could follow the progress of a particular version across the entire set of pipelines we manage
- A span ID representing each individual job, so we could link together tasks into jobs
- Fields with the version (and possibly other identifying information) from each resource used in the task
- The exit status of the task

This is currently possible with Travis-CI, CircleCI, and Jenkins, but this implementation depends on information that Concourse (I believe, might be wrong) does not currently make available to tasks. https://github.com/honeycombio/buildevents

[...]

PR #5461, opened 2020-04-19
opentelemetry ginkgo reporter

### LINKED ISSUE #5423: emit domain-specific spans from within tasks/resources

I have a task that makes a request to an internal web service that my team has written. Both my Concourse cluster and this web service are emitting spans to the same Jaeger cluster, but I would like some way to link the spans from Concourse and the spans from my service under a common trace (the request being made inside the task container).

Since the w3c defines a standard for the `traceparent` header, let's try the following: if tracing is enabled, any task/resource get/resource put script should have a `TRACE_PARENT` env var set where the value is the trace ID of the build in question.

:+1:

jamieklassen:
# Existing Issue

Fixes #5423

* `$TRACEPARENT` will be set in w3c-compatible format (`<version>-<trace id>-<span id>-<flags>`) in all build containers when tracing is enabled
    * I want to emphasize that I'm not married to this particular format/codec (as there are others including the B3 ones established with Zipkin)
* This means that if you `fly execute -c hack/tracing-task.yml` you can get the following awesome output from Jaeger:
    Screen Shot 2020-04-18 at 9 58 47 PM

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


hack/tracing-task.yml
```diff
@@ -0,0 +1,52 @@
...
+            'logging': True,
+        },
+        service_name='downstream-service',
+        validate=True,
+    )
+    # this call also sets opentracing.tracer
+    tracer = config.initialize_tracer()
+
+    version, trace_id, parent_id, flags = os.environ['TRACEPARENT'].split('-')
+    trace_id, span_id, parent_id, flags = span_context_from_string(trace_id + ':' + parent_id + ':' + parent_id + ':' + flags)
```
pivotal-jamie-klassen, line 45: This parsing is painful - the python jaeger client doesn't speak w3c tracecontext, but it _does_ speak b3. How many other client SDKs will have this bias, since b3 has been around longer? 🤔

pivotal-jamie-klassen:
Above I used the term "format/codec" -- I'm starting to get the sense that the official term for this "standard for how different systems can tell each other which trace they're part of" is called a "propagation format". OpenTracing seemed to call these ["Carrier formats"](https://opentracing.io/docs/overview/inject-extract/).

I'm interested in understanding this topic in a general fashion because I would like to enable a workflow where one web node starts a trace by creating a build, and potentially a different web node finishes it after starting/executing that build. The model feels very similar to what the Zipkin docs call [message tracing](https://zipkin.io/pages/instrumenting.html#message-tracing), but instead of an actual AMQP message (or other protocol), the "message" is a build in the DB.

pivotal-jamie-klassen:
In my commit message I allude that

> perhaps the codec should be configurable

Specifically what I'm imagining is a `--trace-propagation-format` startup flag with values like `w3c` or `b3` which would affect the way environment variables get set in build containers, and (looking forward) which headers will be used to extract trace context from API requests.

pivotal-jamie-klassen:
@cirocosta observed that a good way to experiment/dogfood the value of this feature would be for each test in a ginkgo suite to emit a span. This allows a user to get much more granular feedback about their testing jobs: instead of "how long does `bosh-topgun-both` take" or "how flaky is `bosh-topgun-both`" you can ask the corresponding question about each of the 50 tests in that suite.

pivotal-jamie-klassen:
After thinking about how to “trace a ginkgo suite”, I remembered @njbennett’s excellent discussion on #3958. I searched around and found https://github.com/cloudfoundry/honeycomb-ginkgo-reporter. So ginkgo actually has a pluggable mechanism for “instrumenting test runs”, amazing! https://onsi.github.io/ginkgo/#writing-custom-reporters. If I get the change, I will attempt to use the honeycomb-ginkgo-reporter as a model for a generic opentelemetry reporter, or, failing that, a jaeger one.

pivotal-jamie-klassen:
Today I implemented an experimental opentelemetry reporter, and when I ran

```bash
$ fly -t dev execute -i concourse=. -c <(cat <<EOF
platform: linux
image_resource:
  type: registry-image
  source: {repository: concourse/unit}
inputs:
- name: concourse
params:
  JAEGER_URL: http://jaeger:14268
run:
  path: bash
  args: 
  - -c
  - |
    go get github.com/onsi/ginkgo/ginkgo
    cd concourse
    ginkgo -nodes=8 atc/api
EOF
)
```

I was able to get jaeger output like:

Screen Shot 2020-05-01 at 6 11 07 PM

Which is really pretty cool.

pivotal-jamie-klassen:
Today I realized that it probably makes sense for concourse's garden client to simply set w3c tracecontext headers on its requests to garden. then our garden server implementation could propagate trace context from HTTP headers into containers via environment variables.

It was seeing a trace that started with `web` and went immediately to `ginkgo` that tipped me off to this -- the ginkgo process is actually running on a worker, so it's probably useful if the worker service appears in the trace output.

atc/api/api_suite_test.go
```diff
@@ -231,5 +237,24 @@ var _ = AfterEach(func() {
...
+		)
+		defer flush()
+		reporter := tracing.NewOpenTelemetryReporter(global.Tracer("ginkgo"))
+		RunSpecsWithDefaultAndCustomReporters(
+			t, "API Suite",
+			[]Reporter{reporter},
+		)
+	} else {
+		RunSpecs(t, "API Suite")
+	}
```
jamieklassen, line 259: this is noisy and should be extracted to a helper

atc/scheduler/algorithm/group_resolver.go
```diff
@@ -89,7 +89,7 @@ func (r *groupResolver) Resolve(ctx context.Context) (map[string]*versionCandida
 
 	if !resolved {
 		span.SetAttributes(key.New("failure").String(string(failure)))
-		span.SetStatus(codes.NotFound)
+		span.SetStatus(codes.NotFound, "")
```
jamieklassen, line 92:
```suggestion
		span.SetStatus(codes.NotFound, string(failure))
```

hack/tracing-task.yml
jamieklassen: also I think this task should not be included in this PR at all!

atc/scheduler/algorithm/group_resolver.go
jamieklassen: and maybe the `SetAttributes` call above can be removed

tracing/opentelemetry_reporter_test.go
```diff
@@ -0,0 +1,152 @@
...
+			)
+			Ω(specSpanAttrs["component_texts[1]"]).Should(
+				Equal(core.String("more context")),
+			)
+			Ω(specSpanAttrs["component_texts[2]"]).Should(
+				Equal(core.String("spec name")),
+			)
+		})
+
+		// TODO what about pending/skipped tests?
```
jamieklassen, line 137: and also this thing should really mark the status of the individual specs - passed/failed/etc.

tracing/opentelemetry_reporter.go
```diff
@@ -0,0 +1,107 @@
...
+func (reporter *OpenTelemetryReporter) BeforeSuiteDidRun(
+	setupSummary *types.SetupSummary,
+) {
+	beforeSuiteSpan := trace.SpanFromContext(reporter.currCtx)
+	switch setupSummary.State {
+	case types.SpecStateFailed:
+		beforeSuiteSpan.SetAttributes(core.KeyValue{
+			Key:   "state",
+			Value: core.String("failed"),
+		})
```
aoldershaw, line 74: I notice elsewhere, we use `span.SetStatus(...)`, e.g. https://github.com/concourse/concourse/blob/45bbca879cc33f790819f08f1fc29fd527030df5/atc/scheduler/algorithm/group_resolver.go#L77

Is there a recommended way for reporting successes/failures?

xtremerui:
Will close now. Feel free to reopen if you wish to continue this PR.

Issue #5310, opened 2020-03-16
Resource view: surface pin errors


jamieklassen:
As a concourse user capable of pinning/unpinning resource versions (a `pipeline-operator` or above, by default)
I want the UI to tell me when my attempts to pin or unpin have failed
So that I can re-attempt, diagnose, or otherwise avoid being misinformed about the state of my pipeline

Failing to pin:
<img width="499" alt="Screen Shot 2020-03-16 at 5 39 16 PM" src="https://user-images.githubusercontent.com/22056320/76802095-54ec5900-67ad-11ea-835a-e5caaf304881.png">

Failing to unpin:
<img width="818" alt="Screen Shot 2020-03-16 at 5 40 15 PM" src="https://user-images.githubusercontent.com/22056320/76802108-5c136700-67ad-11ea-8f66-3881288c22df.png">

### Acceptance criteria
Pinning failure:
**Given** i am on the resource page for a resource that is not currently pinned
**And** i have clicked the pin icon on one of the versions
**When** the underlying `PinResource` API call gives a non-200, non-401 status
**Then** i see a red outline around the check bar

Pinning failure, then success:
**Given** i am on the resource page for a resource that is not currently pinned
**And** i have previously failed to pin it
**And** i have clicked the pin icon on one of the versions an additional time
**When** the underlying `PinResource` API call succeeds
**Then** i see the check bar replaced by the pin tools, with no red outline

Unpinning failure:
**Given** i am on the resource page for a resource that is dynamically pinned
**And** i have clicked the unpin icon in the pin tools
**When** the underlying `UnpinResource` API call gives a non-200, non-401 status
**Then** i see a red outline around the pin tools

Unpinning failure, then success:
**Given** i am on the resource page for a resource that is dynamically pinned
**And** i have previously failed to unpin it
**And** i have clicked the unpin icon in the pin tools an additional time
**When** the underlying `UnpinResource` API call succeeds
**Then** i see the pin tools replaced by the check bar, with no red outline

### Notes
The conditions under which the red outline should disappear are somewhat underspecified. It may be desirable in the future for the red outline to disappear after a suitable time period, or even for it to gradually dim out. For the time being, though, it is safe to assume that the only conditions under which the red outline will be dismissed are if a subsequent pin/unpin attempt succeeds or if the page refreshes.

matthewpereira:
Last week @Lindsayauchin @zoetian @pivotal-jwinters and I briefly mobbed on a solution like this:

[...]

PR #5496, opened 2020-04-24
web: behaviour: recover from turbulence view

aoldershaw:
# Why is this PR needed?

If any network request fails once, the dashboard will be stuck in the
"show turbulence" page view forever (until you refresh the page). This
happens fairly frequently to me at least - if you have the dashboard
open, close your laptop for a little, and come back, the page will be
unusable due to the turbulence view.

Allow the turbulence view to disappear if all of the once failing network requests start succeeding.

Rather than just keeping track of whether we should display the turbulence view (which was previously never unset), keep track of whether each endpoint is failing. If any of the endpoints are still failing, display the turbulence view.

One change is that the dashboard no longer considers the "team pipelines" endpoint response for displaying the show turbulence view - if it ever errors, we'd still be stuck in the turbulence view forever (since nothing would trigger another `FetchPipelines` effect). I'd also argue that displaying turbulence if this endpoint fails is not that valuable - if there is truly turbulence, the next API calls (in < 5 seconds) will display this.

# Contributor Checklist

- [x] Unit tests
- [ ] Integration tests

# Reviewer Checklist

- [x] Code reviewed
- [x] Tests reviewed
- [x] Release notes reviewed
- [x] PR acceptance performed

### DIFF
--- a/release-notes/latest.md
+++ b/release-notes/latest.md
@@ -61,4 +61,7 @@
 
 * Add loading indicator on dashboard while awaiting initial API/cache response. #5458
 
+#### <sub><sup><a name="5496" href="#5496">:link:</a></sup></sub> fix
+
+* Allow the dashboard to recover from the "show turbulence" view if any API call fails once, but starts working afterward. This will prevent users from needing to refresh the page after closing their laptop or in the presence of network flakiness. #5496
 
--- a/web/elm/src/Dashboard/Dashboard.elm
+++ b/web/elm/src/Dashboard/Dashboard.elm
@@ -72,8 +72,7 @@ import Views.Styles
 
 init : Routes.SearchType -> ( Model, List Effect )
 init searchType =
-    ( { showTurbulence = False
-      , now = Nothing
+    ( { now = Nothing
       , hideFooter = False
       , hideFooterCounter = 0
       , showHelp = False
@@ -92,6 +91,10 @@ init searchType =
       , isTeamsRequestFinished = False
       , isResourcesRequestFinished = False
       , isPipelinesRequestFinished = False
+      , isJobsErroring = False
+      , isTeamsErroring = False
+      , isResourcesErroring = False
+      , isPipelinesErroring = False
       , viewportWidth = 0
       , viewportHeight = 0
       , scrollTop = 0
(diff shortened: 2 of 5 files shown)


cirocosta:
Oh, great catch! 

I just did some quick acceptance here w/ the following patch

```diff
diff --git a/atc/api/pipelineserver/list_all.go b/atc/api/pipelineserver/list_all.go
index 9042e202a..648166c53 100644
--- a/atc/api/pipelineserver/list_all.go
+++ b/atc/api/pipelineserver/list_all.go
@@ -9,10 +9,18 @@ import (
 	"github.com/concourse/concourse/atc/db"
 )
 
+var fail = 0
+
 // show all public pipelines and team private pipelines if authorized
 func (s *Server) ListAllPipelines(w http.ResponseWriter, r *http.Request) {
 	logger := s.logger.Session("list-all-pipelines")
 
+	fail ^= 1
+	if fail == 0 {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+
 	acc := accessor.GetAccessor(r)
 
 	var pipelines []db.Pipeline
```

and indeed, before the PR: once 500 gets returned, never gets back. After the patch, you can see the page going from turbulence to normal and back and forth 👍  

Will leave the actual code review for the folks who are ~way~ more familiar with the code than me 😁

cirocosta:
> One change is that the dashboard no longer considers the "team pipelines" endpoint response for displaying the show turbulence view [...]

does that mean that if you try to open the page for a given pipeline, and you get a 500 right there, you'll see nothing? i.e., all of the layout but w/out the pipeline graph?

aoldershaw:
@cirocosta speedy review, thanks! For anyone else who's interested, another way to do acceptance on this is to set the network preset to "Offline" in the network tab of developer tools (at least in Chrome).

> does that mean that if you try to open the page for a given pipeline, and you get a 500 right there, you'll see nothing

Good question - nope, that behaviour is still the same (it will display the turbulence view on the pipeline page). On the dashboard, after a user reorders pipelines, it will immediately trigger `GET /api/v1/teams/<team>/pipelines` (I believe as a convenience to correct the displayed pipeline order if the reordering fails - but correct me if there's more to it than that @pivotal-jamie-klassen or @zoetian - https://github.com/concourse/concourse/commit/8523c4434969fe4e454922b7d4ab42ffee1e6648). If this request fails, prior to this PR, we'd display turbulence on the dashboard (until the user refreshes the page). With the changes here, we won't display anything to indicate a failure until the next API calls come back as unsuccessful (which would be within 5 seconds, give or take). 

I think a better UX would be to have some sort of toast message system to indicate that the order pipeline request failed - there was some discussion of implementing that for the resource view here: https://github.com/concourse/concourse/issues/5310

PR #5497, opened 2020-04-24
web: behavior: fix non-recursive plan usage, refactor build factory

vito:
# Why is this PR needed?

I got tired of the code under `scheduler/factory` - it's become a bit of a broken window over the years and I've been itching to refactor it. 😅

This PR supercedes #5026

* closes #5026
* hopefully makes it easier to correctly traverse build plans

code

(j/k, each commit has details in the message)

# Contributor Checklist

- [ ] Unit tests
- [ ] Integration tests

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


atc/config.go
```diff
@@ -534,8 +526,52 @@ func (config PlanConfig) ResourceName() string {
 	panic("no resource name!")
 }
 
-func (config PlanConfig) Hooks() Hooks {
-	return Hooks{Abort: config.Abort, Error: config.Error, Failure: config.Failure, Ensure: config.Ensure, Success: config.Success}
+// Each walks through the PlanConfig and invokes f with it and any nested
+// PlanConfigs contained therein.
+func (config PlanConfig) Each(f func(PlanConfig)) {
```
pivotal-jwinters, line 531: neat!

atc/db/team.go
```diff
@@ -1331,18 +1331,20 @@ func (t *team) insertJobPipes(tx Tx, jobConfigs atc.JobConfigs, resourceNameToID
...
 	for _, jobConfig := range jobConfigs {
-		for _, plan := range jobConfig.Plans() {
+		var err error
+		jobConfig.PlanConfig().Each(func(plan atc.PlanConfig) {
+			if err != nil {
+				return
+			}
+
 			if plan.Get != "" {
 				err = insertJobInput(tx, plan, jobConfig.Name, resourceNameToID, jobNameToID)
```
pivotal-jwinters, line 1341: Maybe this should only overwrite err if the returned err is not nil? Otherwise this might clobber the earlier error?

atc/db/team.go
```diff
@@ -1331,18 +1331,20 @@ func (t *team) insertJobPipes(tx Tx, jobConfigs atc.JobConfigs, resourceNameToID
...
-				if err != nil {
-					return err
-				}
 			} else if plan.Put != "" {
 				err = insertJobOutput(tx, plan, jobConfig.Name, resourceNameToID, jobNameToID)
```
pivotal-jwinters, line 1343: same here

atc/task.go
```diff
@@ -20,7 +20,7 @@ type TaskConfig struct {
 	ImageResource *ImageResource `json:"image_resource,omitempty"`
 
 	// Limits to set on the Task Container
-	Limits ContainerLimits `json:"container_limits,omitempty"`
+	Limits *ContainerLimits `json:"container_limits,omitempty"`
```
pivotal-jwinters, line 23: how come?

atc/builds/planner_test.go
```diff
@@ -0,0 +1,702 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/atc/builds"
+	"github.com/concourse/concourse/atc/db"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
```
pivotal-jwinters, line 12: Can we stick with the same test suite within a given package?

atc/db/team.go
pivotal-jwinters: oh wait, I missed the err != nil at the beginning

atc/db/team.go
pivotal-jwinters: nvm

atc/task.go
pivotal-jwinters: nvm read the commit -> omitempty

atc/builds/planner_test.go
vito: Yeah it's a little hokey that it doesn't output the planner tests at all when they pass. I think I'll just convert the tracker tests.

pivotal-jwinters: Can we stick with ginkgo since that what most of concourse uses

pivotal-jwinters: I think using testify for small pluggable stuff is fine, but the core of concourse should be consistent.

vito: A long while back we discussed writing new test suites using `testify` instead as it's more accessible for outside contributors and it's generally harder to get wrong. The entire [`worker/backend/` package](https://github.com/concourse/concourse/tree/master/worker/backend) for example uses testify, and in various other places we're trending towards table-driven tests which is really not using a whole lot of Ginkgo anyway.

I don't agree that we should continue using Ginkgo just because the rest of the code uses Ginkgo. If that's the case then we're really just never going to switch. I don't want to end up with the "juicy bits" of Concourse being covered with Ginkgo and the "pluggable bits" being covered with `testify`. And if I had to pick one, after seeing how our test suites tend to grow with Ginkgo, right now I would choose `testify`.

I'm not saying we should go back and re-write all the Ginkgo tests in testify, but as long as we're writing new test suites I don't think we should have to keep using Ginkgo.

pivotal-jwinters: My understanding was that pluggable bits could/should use testify because there's a greater chance that those will see contributions from the community. 

Maintaining two different styles of test suites for all of concourse is a huge cognitive burden for core contributors and community members. The components of concourse are not well enough isolated to make changes in only one area, so you're going to be making changes across both suites. 

There are definitely cases where we abuse ginkgo, but there's an even greater chance of abusing testify because there is less structure to those tests. All the testitfy tests are new, so they might look better by comparison, but that's just because they haven't had time to suffer the same rot that our ginkgo tests have. You can have a look at dex to see how bad things can get without a good structure in place: https://github.com/concourse/dex/blob/master/server/server_test.go.

pivotal-jwinters: I kinda get the sense that we're "throwing the baby out with the bath water" because of a couple test suites that got out of hand. I feel like there are lots of other examples of test suites that have aged really well.

vito: This isn't about throwing away Ginkgo and switching to Testify because things didn't work out in the past. I don't expect Testify to fix all of our problems.

Here's where I'm at:

* We want more contributors to all parts of Concourse.
* Contributors outside of the Pivotal bubble aren't familiar with Ginkgo.
* Ginkgo is hard to get right, even for the core team.

My point isn't that Testify is better, it's that the choice of Ginkgo is suboptimal for the path we're trying to go. I think Ginkgo is great for sideprojects and walled gardens, but if our primary goal is contributor enablement it hinders more than it helps.

We talked about using Testify for pluggable bits, but that never really happened because we don't work on those bits ourselves. It was a reasonable idea to get the ball rolling, but I don't think that was ever the end goal. Compartmentalizing Testify usage like that sets the tone that we encourage contributions to some components but not others, which feels counterproductive and unrealistic given some of the recent contributions: `set_pipeline` step, `load_var` step, `var_sources:`, etc.

Maintaining Testify suites in addition to our Ginkgo suites in the interim is a burden for *us*, but keep in mind that community members are more likely to be familiar with `testing.T` -style testing. We don't feel a burden right now because we're the ones pushing it on everyone else. The collective burden of us learning Testify is a *lot* lower than the burden of all community contributors - current and future - having to learn Ginkgo. As we add more Testify tests we're at least directly benefiting those in the community, even before we're completely finished with the transition.

Testify takes a day to learn and slightly longer to internalize. It has good matchers, and it has suites for setup and teardown - it just doesn't let you go crazy with it. It provides just enough structure and gets out of the way.

Ginkgo's structure is a double-edged sword. I've seen plenty of test suites become "write-only" because they become harder and harder to refactor as the structure grows and grows. Everything starts off meticulously organized, and over time it gets worse - just like with any code. But by the time it's so bad that people can recognize how bad it is, it's *so much work* to refactor it that no one ever bothers.

To me, the `server_test.go` example you gave just looks like verbose, disorganized code. But it looks pretty easy to refactor.

In any case, we can bring this discussion to the team and check in with how they're feeling about testify on the `containerd` track. It honestly wouldn't take long to change my tests to Ginkgo, since they're table-driven, but to me that feels like a step back. I would rather convert the rest of the package to `testify` and begin adopting it for new code that we add under the `builds/` package.

vito:
I pulled on a bit of a thread. I started on a mission to get `atc/builds` to be fully one test framework (`testify`), which led me to the `TrackerRunner`. That component had some twisty logic that made it pretty painful to write tests for as-is, so I started splitting it into two smaller, easier-to-test parts which now live under an `atc/component/` package. 

Then I went and chased down all the other various component runners and started on a grand unifying adventure. Long story short, now all of our components are run using the `atc/component/`, and I was able to delete ~4 different implementations of the exact same code. 🥳

I ran into a couple odd things along the way and fixed them. They're tied into this PR because the code surrounding them had to change:

* One commit changes var source pool GC to *not* be represented as a component, and just GC in-memory. There's no cross-`web`-node coordination going on there so it really shouldn't be a component.
* Another commit removes the static worker registration flags and functionality, which to my knowledge hasn't been used in years.

Sorry the diff is getting so huge. 😅 I've tried to keep the commit history tidy. I don't mean to pile on pressure with such a large change - I can keep rebasing as long as it's just boring manual labor.

vito:
Also, would be particularly interested in constructive feedback for the `atc/component/` tests.

* `*component.Scheduler` is tested end-to-end style, covering the full behavior through one lifecycle. I tried a few other things but this approach felt like it had the most bang for the buck. It at least makes use of `s.Run("foo")` to provide sub-test descriptions.
* `*component.Coordinator` is table-tested, since there are a ton of conditionals involved multiplied by two different methods. This feels worthwhile as the end result is a pretty easy to skim list of test cases which have the description, setup, and result all in one entry, but the implied risk is whether `(CoordinatorTest).Run)` gets too complicated. There are some "invariant" assertions in there, but aside from that I tried to keep it pretty dumb.

I think these ended up in a good place but I can't really know that until someone else reads them. 🤷‍♂️

atc/scheduler/factory/build_factory_test.go
```diff
@@ -0,0 +1,679 @@
...
+}
+
+var factoryTests = []FactoryTest{
+	{
+		Title: "get step",
+		ConfigYAML: `
+			get: some-name
+			resource: some-resource
+			params: {some: params}
+			version: {doesnt: matter}
```
aoldershaw: Just to confirm my understanding: this version doesn't matter, since it's the responsibility of the scheduler (algorithm?) to determine the build inputs - so this pinned version in the config would be used by the scheduler, and the `BuildFactory` takes in whatever `BuildInputs` are given to it as the source of truth?

vito: Yep!

atc/atccmd/command.go
```diff
@@ -958,7 +957,7 @@ func (cmd *RunCommand) constructBackendMembers(
 				&scheduler.Scheduler{
 					Algorithm: alg,
 					BuildStarter: scheduler.NewBuildStarter(
-						factory.NewBuildFactory(
+						builds.NewPlanner(
```
aoldershaw, line 999: Really minor, and I don't really have strong opinions on this either way, but is there a reason you went with `builds` instead of `build`? I feel like Go packages tend to be singular (with the exception of `bytes`, `strings`, `errors`, etc., but if I had to guess, I'd say they mainly used the plural so as not to conflict with the built-in types of the same name). Plus, `build.Planner` feels just slightly easier to say than `builds.Planner` to me

vito: @aoldershaw Following the same pattern as those packages you mentioned. The pluralization also avoids clashing with local vars; `var build build.Build` actually shadows the `build` package from then on so it'd be a bit problematic. I suppose the same is true for `var builds []build.Build`, though I imagine that'd be less common.

I think the plural form also makes the package feel more like a "domain" than worrying about whether the contained types really deal with a single build or deal with the general concept of builds. That's all kind of subtle semantics though. 🤷‍♂️

vito:
@pivotal-jwinters Yeah no worries, all sorted - I cut it off just before the move from `scheduler.BuildFactory` to `builds.Planner` since that was the point of contention.

I split the component stuff out into a separate PR based on `master` so it can be reviewed independently: https://github.com/concourse/concourse/pull/5569

The rest of the commits have been moved into #5504 which has been freshly rebased off this PR.

aoldershaw:
Looks like #5420 getting merged introduced a bunch of conflicts 😬

vito:
😱 Yep, I'll rebase.

atc/scheduler/factory/build_factory_test.go
```diff
@@ -0,0 +1,679 @@
...
+						"duration": "1h"
+					}
+				}
+			]
+		}`,
+	},
+}
+
+func init() {
+	for _, hookType := range []string{"on_success", "on_failure", "on_abort", "ensure"} {
```
jamieklassen, line 484: this block is generating specifications from a template - this feels like a bit of a smell to me. If all these step types have such similar behaviour that specifications can be programatically generated, might it make sense to extract a common interface that they fulfill and somehow test against that? It just seems like "multiple types with the same behaviour" is sorta the definition of polymorphism. What do you think of this observation? I'm mostly speaking from the experience of 'generating tests from a template' in this fashion in the elm codebase and finding it got a bit unwieldy and difficult to read. In the elm case, using a polymorphism technique like extensible records ended up helping.

vito: Yeah this part was a bit hokey. These tests are really just trying to test precedence, and with https://github.com/concourse/concourse/pull/5504 (based on this PR) precedence is determined at the parsing stage, which has its own tests, including [one big precedence test](https://github.com/concourse/concourse/pull/5504/files#diff-3ec0d2bc70f6384123558ee5a728415aR307) that uses all modifiers that shows the precedence all at once.

I can make a follow-up commit to that PR that changes these tests to just have one test per step type without having to worry about precedence.

atc/plan.go
```diff
@@ -1,26 +1,29 @@
 package atc
 
 type Plan struct {
-	ID       PlanID `json:"id"`
+	ID       PlanID `json:"id,omitempty"` // not optional, but omitempty for tests
```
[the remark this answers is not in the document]
vito: Addressed in https://github.com/concourse/concourse/pull/5497/commits/0926a4aee6cfa190747b1bd093f9d05128c1f95e - this was just to make it a bit easier to write the test inputs, but I think this commit is decently more explicit and less hacky.

atc/scheduler/factory/build_factory_test.go
vito: Added this as a TODO in the other PR so this doesn't get lost in the merge.

PR #5500, opened 2020-04-24
Re-introduce mutex timeouts

### LINKED ISSUE #5385: Deadlock in notifications bus causes all builds to hang
## Bug Report

There is a deadlock scenario involving the "notifications bus" used in build management, where the delivery of notification messages contends with registration and deregistration of listeners. Both of those processes want to hold the same mutex. A deadlock occurs when the notification queue (a go channel internal to the pq library) is full. Briefly,
1. If the notification queue is full, then new notifications cannot be inserted by pq, and the goroutine that is meant to do that will block.
2. If it is blocked, then it cannot signal successful completion of a listen or unlisten request.
3. If that signal is not received, then a goroutine that is waiting for it (in ATC) will continue to wait. While it waits, it holds a mutex.
4. The ATC process that is meant to drain messages from the queue cannot do so until it acquires that same mutex. But it will never get it.

The outcome is that build initialization is stalled on that ATC, because there will eventually be many goroutines waiting to acquire that one mutex in order to register listeners. All of those builds appear "stuck" indefinitely. The number of stuck builds grows until the ATC process is shut down.

## Steps to Reproduce

We saw this after upgrading our Postgres DB to a more capable size in AWS RDS. My theory is that because the new DB is faster, it's filling up the queue more reliably. It may be possible to reproduce in a toy environment by blasting out a lot of notifications very quickly, but I haven't tried.

## Expected Results

Builds would run normally.

## Actual Results

For an affected ATC node, all builds become stuck. Build tracking locks are acquired, but not released, so we see that metric climbing consistently. Once this starts, all new builds on that ATC are affected; builds that had already started are able to "almost finish" but appear to hang after all steps are complete. Builds for which some other ATC is responsible, are fine.

We have been resolving this in practice by recreating the ATC node, which is a very forcible way to clear its mutexes and queues. This also entails some cleanup work in the database in order to mark affected builds as aborted, rather than have them appear "started". (That's not only cosmetic, but affects the max-in-flight calculations.)

## Additional Context

_Brace yourself, reader_, for stack dump trawling. The following excerpts were obtained from the `pprof` debugging endpoint on an affected ATC.

There are many instances of the following:
```
goroutine 1723146649 [semacquire, 30 minutes]:
sync.runtime_SemacquireMutex(0xc000dcbfdc, 0x423800, 0x1)
	/usr/local/go/src/runtime/sema.go:71 +0x47
sync.(*Mutex).lockSlow(0xc000dcbfd8)
	/usr/local/go/src/sync/mutex.go:138 +0xfc
sync.(*Mutex).Lock(...)
	/usr/local/go/src/sync/mutex.go:81
github.com/concourse/concourse/atc/db.(*notificationsBus).Listen(0xc000dcbfc0, 0xc06f6d0760, 0x13, 0x0, 0x0, 0x0)
	/tmp/build/1c3187db/concourse/atc/db/notifications_bus.go:48 +0x2c4
github.com/concourse/concourse/atc/db.newConditionNotifier(0x2879920, 0xc000dcbfc0, 0xc06f6d0760, 0x13, 0xc08530a9b0, 0xc06f6d0760, 0x13, 0xc0005803b0, 0xc0154d1a20)
	/tmp/build/1c3187db/concourse/atc/db/notifier.go:13 +0x49
github.com/concourse/concourse/atc/db.(*build).AbortNotifier(0xc062e92780, 0x2895f01, 0x0, 0x0, 0x282c0a0)
	/tmp/build/1c3187db/concourse/atc/db/build.go:462 +0x135
github.com/concourse/concourse/atc/engine.(*engineBuild).Run(0xc023ec1900, 0x2895fa0, 0xc0014eccc0)
	/tmp/build/1c3187db/concourse/atc/engine/engine.go:170 +0x411
created by github.com/concourse/concourse/atc/builds.(*Tracker).Track
	/tmp/build/1c3187db/concourse/atc/builds/tracker.go:49 +0x425
```
with variations on the length of time of waiting for the mutex. These are attempts to start tracking a build, part of which involves registering in the notifications bus. (There are three such buses, but this particular one is the interesting one.) The attempt to `Listen` requires acquisition of the `notificationsL` mutex, in order to put new entries in the `notifications` map which it guards. We also see parallel stuck attempts to `Unlisten`.

The actual holder of the mutex is:
```
goroutine 1722571740 [chan receive, 34 minutes]:
github.com/lib/pq.(*ListenerConn).ExecSimpleQuery(0xc00096cac0, 0xc036b43c48, 0x1d, 0xc0225f0100, 0x0, 0x0)
	/tmp/build/1c3187db/gopath/pkg/mod/github.com/lib/pq@v0.0.0-20181016162627-9eb73efc1fcc/notify.go:304 +0x141
github.com/lib/pq.(*ListenerConn).Unlisten(0xc00096cac0, 0xc08bfaff00, 0x12, 0x12, 0xc055ca0138, 0xc036b43c01)
	/tmp/build/1c3187db/gopath/pkg/mod/github.com/lib/pq@v0.0.0-20181016162627-9eb73efc1fcc/notify.go:224 +0x92
github.com/lib/pq.(*Listener).Unlisten(0xc00060d960, 0xc08bfaff00, 0x12, 0x0, 0x0)
	/tmp/build/1c3187db/gopath/pkg/mod/github.com/lib/pq@v0.0.0-20181016162627-9eb73efc1fcc/notify.go:560 +0x150
github.com/concourse/concourse/atc/db.(*notificationsBus).Unlisten(0xc000dcbfc0, 0xc08bfaff00, 0x12, 0xc052d90070, 0x0, 0x0)
	/tmp/build/1c3187db/concourse/atc/db/notifications_bus.go:78 +0x155
github.com/concourse/concourse/atc/radar.(*intervalRunner).Run(0xc0282b2820, 0x2879520, 0xc004fafc80, 0x2830840, 0xc0761d33e0)
	/tmp/build/1c3187db/concourse/atc/radar/interval_runner.go:73 +0x376
github.com/concourse/concourse/atc/radar.(*Runner).scanResources.func1(0xc006722460, 0x2895fa0, 0xc03c258540, 0x2879520, 0xc004fafc80, 0x28c4a00, 0xc045c2c160, 0xc0771cd910, 0x10)
	/tmp/build/1c3187db/concourse/atc/radar/runner.go:118 +0x147
created by github.com/concourse/concourse/atc/radar.(*Runner).scanResources
/tmp/build/1c3187db/concourse/atc/radar/runner.go:113 +0x316
```
In this case, it's an `Unlisten` attempt for the same bus (at `0xc000dcbfc0`). Behind the scenes, this goes off to https://github.com/lib/pq to run the `UNLISTEN` SQL command. The process is blocked on `chan receive` because it is waiting for a signal from pq that the `UNLISTEN` was successful. That signal is meant to come from another goroutine, which does the actual database communication, and hence processes _both_ `LISTEN`/`UNLISTEN` requests, and the delivery of notifications.

_That_ goroutine is here (note the same address for `ListenerConn`):
```
goroutine 122 [chan send, 34 minutes]:
github.com/lib/pq.(*ListenerConn).listenerConnLoop(0xc00096cac0, 0x0, 0x0)
	/tmp/build/1c3187db/gopath/pkg/mod/github.com/lib/pq@v0.0.0-20181016162627-9eb73efc1fcc/notify.go:148 +0x31d
github.com/lib/pq.(*ListenerConn).listenerConnMain(0xc00096cac0)
	/tmp/build/1c3187db/gopath/pkg/mod/github.com/lib/pq@v0.0.0-20181016162627-9eb73efc1fcc/notify.go:188 +0x2f
created by github.com/lib/pq.newDialListenerConn
	/tmp/build/1c3187db/gopath/pkg/mod/github.com/lib/pq@v0.0.0-20181016162627-9eb73efc1fcc/notify.go:81 +0x124
```
It is blocking on `chan send`, because it's trying to send a message over a different channel. Instead of using `replyChan`, which is the one hooked up waiting for the "I successfully unlistened" signal, it's using `notificationChan`, which is the one for delivering actual notifications from the DB into application code. This is a bounded channel of size 32, which means the go runtime will block if all 32 slots are in use. So this goroutine cannot make progress until somebody pops something off the channel. That somebody is:

```
goroutine 115 [chan send, 34 minutes]:
github.com/lib/pq.(*Listener).listenerConnLoop(0xc00060d960)
	/tmp/build/1c3187db/gopath/pkg/mod/github.com/lib/pq@v0.0.0-20181016162627-9eb73efc1fcc/notify.go:781 +0x1b7
github.com/lib/pq.(*Listener).listenerMain(0xc00060d960)
	/tmp/build/1c3187db/gopath/pkg/mod/github.com/lib/pq@v0.0.0-20181016162627-9eb73efc1fcc/notify.go:795 +0x2b
created by github.com/lib/pq.NewDialListener
	/tmp/build/1c3187db/gopath/pkg/mod/github.com/lib/pq@v0.0.0-20181016162627-9eb73efc1fcc/notify.go:464 +0x12c
```
which is another bit of internal pq plumbing. Here, we're in a tight loop of pulling messages off `notificationChan` and stuffing them into `l.Notify`, but we're blocked because the latter channel - also of bounded capacity - is full. This `Notify` is the channel which is _actually_ hooked up to the application-side code. In Concourse, it's used in `(*notificationsBus).wait`, which is the routine that takes those notifications from the DB and routes them to the appropriate waiting goroutines in ATC. Here's the relevant instance of `wait`:

```
goroutine 116 [semacquire, 34 minutes]:
sync.runtime_SemacquireMutex(0xc000dcbfdc, 0xc000016600, 0x1)
	/usr/local/go/src/runtime/sema.go:71 +0x47
sync.(*Mutex).lockSlow(0xc000dcbfd8)
	/usr/local/go/src/sync/mutex.go:138 +0xfc
sync.(*Mutex).Lock(...)
	/usr/local/go/src/sync/mutex.go:81
github.com/concourse/concourse/atc/db.(*notificationsBus).wait(0xc000dcbfc0)
	/tmp/build/1c3187db/concourse/atc/db/notifications_bus.go:92 +0x2a9
created by github.com/concourse/concourse/atc/db.NewNotificationsBus
	/tmp/build/1c3187db/concourse/atc/db/notifications_bus.go:33 +0x89
```
The goroutine is waiting on the very same `notificationsL` mutex, currently held elsewhere. It wants this mutex because it needs to traverse the `notifications` map of destination channels. But it will not get it: it's being held by goroutine 1722571740 above.

So we have a deadlock, because the holding goroutine won't release the mutex until it's informed that the `UNLISTEN` worked. It won't get that signal until the tracking goroutines inside pq deliver it, but that code is blocked waiting for messages to come off the `Notify` channel. The code which could do that, is waiting for its chance on the mutex.

## Version Info

* Concourse version: 5.7.0
* Deployment type (BOSH/Docker/binary): bosh workers, binary ATC
* Infrastructure/IaaS: AWS
* Did this used to work? yes, but since we bumped up the DB size, it's been recurring every couple of days

pivotal-jwinters:
Fixes #5385

This PR allows users to configure two different timeouts:
- The `NotificationTimeout`, for when listening for pub/sub messages from postgres
- The `LockTimeout`, for when acquiring locks from postgres

This features is marked as experimental, and the default behaviour is preserved from previous versions of concourse (no timeout).

### DIFF
--- a/atc/db/lock/mutex.go
+++ b/atc/db/lock/mutex.go
@@ -1,0 +1,45 @@
+package lock
+
+import "time"
+
+func NewMutex(timeout time.Duration) *mutex {
+	return &mutex{
+		c:  make(chan struct{}, 1),
+		to: timeout,
+	}
+}
+
+type mutex struct {
+	c  chan struct{}
+	to time.Duration
+}
+
+func (m *mutex) Unlock() {
+	<-m.c
+}
+
+func (m *mutex) Lock() bool {
+	return m.LockWithTimeout(m.to)
+}
+
+func (m *mutex) LockWithTimeout(timeout time.Duration) bool {
+	if timeout > 0 {
+		return m.lockWithTimeout(timeout)
+	} else {
+		return m.lockIndefinitely()
+	}
+}
+
+func (m *mutex) lockWithTimeout(timeout time.Duration) bool {
+	select {
+	case m.c <- struct{}{}:
+		return true
+	case <-time.After(timeout):
+	}
+	return false
+}
+
+func (m *mutex) lockIndefinitely() bool {
+	m.c <- struct{}{}
+	return true
+}
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -235,6 +235,9 @@ type RunCommand struct {
 	SystemClaimKey        string   `long:"system-claim-key" default:"aud" description:"The token claim key to use when matching system-claim-values"`
 	SystemClaimValues     []string `long:"system-claim-value" default:"concourse-worker" description:"Configure which token requests should be considered 'system' requests."`
 	EnableArchivePipeline bool     `long:"enable-archive-pipeline" description:"Enable /api/v1/teams/{team}/pipelines/{pipeline}/archive endpoint."`
+
+	LockTimeout         time.Duration `long:"lock-timeout" default:"0" description:"EXPERIMENTAL: Amount of time to wait before giving up acquiring locks"`
+	NotificationTimeout time.Duration `long:"notification-timeout" default:"0" description:"EXPERIMENTAL: Amount of time to wait before giving up listening for pub/sub messages"`
 }
 
 type Migration struct {
@@ -467,7 +470,7 @@ func (cmd *RunCommand) Runner(positionalArguments []string) (ifrit.Runner, error
 		return nil, err
 	}
 
-	lockFactory := lock.NewLockFactory(lockConn, metric.LogLockAcquired, metric.LogLockReleased)
+	lockFactory := lock.NewLockFactoryWithTimeout(lockConn, metric.LogLockAcquired, metric.LogLockReleased, cmd.LockTimeout)
 
 	apiConn, err := cmd.constructDBConn(retryingDriverName, logger, cmd.APIMaxOpenConnections, "api", lockFactory)
 	if err != nil {
@@ -1392,7 +1395,16 @@ func (cmd *RunCommand) constructDBConn(
 	connectionName string,
 	lockFactory lock.LockFactory,
 ) (db.Conn, error) {
-	dbConn, err := db.Open(logger.Session("db"), driverName, cmd.Postgres.ConnectionString(), cmd.newKey(), cmd.oldKey(), connectionName, lockFactory)
+	dbConn, err := db.OpenWithBusTimeout(
+		logger.Session("db"),
+		driverName,
+		cmd.Postgres.ConnectionString(),
+		cmd.newKey(),
+		cmd.oldKey(),
+		connectionName,
+		lockFactory,
+		cmd.NotificationTimeout,
+	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to migrate database: %s", err)
 	}
--- a/atc/db/lock/lock.go
+++ b/atc/db/lock/lock.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"code.cloudfoundry.org/lager"
 )
@@ -25,6 +26,8 @@ const (
 )
 
 var ErrLostLock = errors.New("lock was lost while held, possibly due to connection breakage")
+
+const lockTimeout = 10 * time.Second
 
 func NewBuildTrackingLockID(buildID int) LockID {
 	return LockID{LockTypeBuildTracking, buildID}
@@ -65,9 +68,9 @@ type LockFactory interface {
 }
 
 type lockFactory struct {
-	db           LockDB
-	locks        lockRepo
-	acquireMutex *sync.Mutex
+	db    LockDB
+	locks lockRepo
+	mutex *mutex
 
 	acquireFunc LogFunc
 	releaseFunc LogFunc
@@ -91,7 +94,7 @@ func NewLockFactory(
 			locks: map[string]bool{},
 			mutex: &sync.Mutex{},
 		},
-		acquireMutex: &sync.Mutex{},
+		mutex: &mutex{make(chan struct{}, 1)},
 	}
 }
 
@@ -102,21 +105,21 @@ func NewTestLockFactory(db LockDB) LockFactory {
 			locks: map[string]bool{},
 			mutex: &sync.Mutex{},
 		},
-		acquireMutex: &sync.Mutex{},
-		acquireFunc:  func(logger lager.Logger, id LockID) {},
-		releaseFunc:  func(logger lager.Logger, id LockID) {},
+		mutex:       &mutex{make(chan struct{}, 1)},
+		acquireFunc: func(logger lager.Logger, id LockID) {},
+		releaseFunc: func(logger lager.Logger, id LockID) {},
 	}
 }
 
 func (f *lockFactory) Acquire(logger lager.Logger, id LockID) (Lock, bool, error) {
 	l := &lock{
-		logger:       logger,
-		db:           f.db,
-		id:           id,
-		locks:        f.locks,
-		acquireMutex: f.acquireMutex,
-		acquired:     f.acquireFunc,
-		released:     f.releaseFunc,
+		logger:   logger,
+		db:       f.db,
+		id:       id,
+		locks:    f.locks,
+		mutex:    f.mutex,
+		acquired: f.acquireFunc,
+		released: f.releaseFunc,
 	}
 
 	acquired, err := l.Acquire()
@@ -147,42 +150,46 @@ type LockDB interface {
 type lock struct {
 	id LockID
 
-	logger       lager.Logger
-	db           LockDB
-	locks        lockRepo
-	acquireMutex *sync.Mutex
+	logger lager.Logger
+	db     LockDB
+	locks  lockRepo
+	mutex  *mutex
 
 	acquired LogFunc
 	released LogFunc
 }
 
 func (l *lock) Acquire() (bool, error) {
-	l.acquireMutex.Lock()
-	defer l.acquireMutex.Unlock()
+	if l.mutex.Lock(lockTimeout) {
+		defer l.mutex.Unlock()
 
-	logger := l.logger.Session("acquire", lager.Data{"id": l.id})
+		logger := l.logger.Session("acquire", lager.Data{"id": l.id})
 
-	if l.locks.IsRegistered(l.id) {
-		logger.Debug("not-acquired-already-held-locally")
-		return false, nil
+		if l.locks.IsRegistered(l.id) {
+			logger.Debug("not-acquired-already-held-locally")
+			return false, nil
+		}
+
+		acquired, err := l.db.Acquire(l.id)
+		if err != nil {
+			logger.Error("failed-to-register-in-db", err)
+			return false, err
+		}
+
+		if !acquired {
+			logger.Debug("not-acquired-already-held-in-db")
+			return false, nil
+		}
+
+		l.locks.Register(l.id)
+
+		l.acquired(logger, l.id)
+
+		return true, nil
 	}
 
-	acquired, err := l.db.Acquire(l.id)
-	if err != nil {
-		logger.Error("failed-to-register-in-db", err)
-		return false, err
-	}
-
-	if !acquired {
-		logger.Debug("not-acquired-already-held-in-db")
-		return false, nil
-	}
-
-	l.locks.Register(l.id)
-
-	l.acquired(logger, l.id)
-
-	return true, nil
+	l.logger.Debug("timed-out-acquiring-lock")
+	return false, nil
 }
 
 func (l *lock) Release() error {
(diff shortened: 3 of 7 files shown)


atc/db/lock/mutex.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (m *mutex) LockWithTimeout(timeout time.Duration) bool {
+	if timeout > 0 {
+		return m.lockWithTimeout(timeout)
+	} else {
+		return m.lockIndefinitely()
+	}
+}
+
+func (m *mutex) lockWithTimeout(timeout time.Duration) bool {
```
cirocosta, line 33: ha! neat :grin:

atc/db/lock/mutex.go
```diff
@@ -0,0 +1,45 @@
...
+	select {
+	case m.c <- struct{}{}:
+		return true
+	case <-time.After(timeout):
```
cirocosta, line 37: ha! neat :grin:

cirocosta:
hey @pivotal-jwinters,

I'm new to how we make use of the notication bus in our code, could you provide
some examples of how timeouts help us out with regards to lock that it tries to
acquire? or, perhaps, how things should break (or, partially work) to make the
timeout ... timeout?

thanks!

pivotal-jwinters:
@cirocosta the issue #5385 has a ton of useful info that describes the problem.

cirocosta:
indeed! could you share a bit how the timeout goes about solving the deadlock
mentioned there though? I'm having a hard time figuring out how one could go
about doing acceptance of this one to solve the linked issue (i.e., how could
one perhaps tweak part or the code to introduce the deadlock and then fix with
the timeout) :thinking:

thanks!

pivotal-jwinters:
This PR is a guess. I don't have any repro steps for getting concourse into a deadlocked state. All my changes are based on the fact that I believe concourse is resilient to failure during locking and when listening for notifications. But just to be sure it's off by default, and marked as experimental.

Let's briefly discuss the two scenarios:

### Locking timeout

In practice, I don't think we've actually seen any issues around the locking timeout since we added other synchronization mechanisms ([for example keeping track of running checks](https://github.com/concourse/concourse/blob/master/atc/lidar/checker.go#L48)) but the original PR was for this specific use-case so I've kept it around.

The original PR was reverted because `radar` was sensitive to not acquiring a lock, but now that `radar` is gone everything else should be able to handle this case gracefully.

### Notifications timeout

We care about two operations on the notification bus  (`Listen`, and `Unlisten`) since they share the same mutex. 

We have a couple runners (tracker_runner and interval_runner) that use the notification bus directly to `Listen` and `Unlisten` but these components only call these methods once at startup so it's not a big deal. 

The cases we're interested in are the abort notifier and events channel, because those `Listen` and `Unlisten` (via a `ConditionNotifier`) on demand. If a build can't listen for the abort channel, it errors. This isn't ideal but it is acceptable since we surface this to end users, and they can always re-run the build. The events channel listens for notifications in the web handler, so a failed request here is treated like any other failed http request, which is fine.

aoldershaw:
It seems to me that the root of the problem is holding the `bus.notificationsL` lock while waiting for `bus.listener.Listen` (sim. `.Unlisten`), since draining the notification channel (in `notificationBus.wait()` requires holding `bus.notificationsL` as well. However, it seems that the locks hold two different purposes:

1. In `Listen()` and `Unlisten()`, the lock is to ensure that only one Goroutine can modify `bus.notifications` at a time (otherwise, if two Goroutines called `Listen` at the same time, one could overwrite the other, and there'd only be one `sink`, and we could call `bus.listener.Listen()` twice).
2. in `wait()`, the lock is only to ensure we don't panic when iterating over the map, since it could be modified concurrently by `Listen()` or `Unlisten()` (correct me if I'm wrong here).

What if we changed `bus.notifications` to a `sync.Map` - for case 2, we wouldn't need to acquire any lock (ourselves), and so we'd still be able to drain the notification channel

Let me know if you think this won't work

pivotal-jwinters:
@aoldershaw you could achieve something similar with just a RWMutex, however at some point I convinced myself that this wasn't enough to solve the problem, now I just need to remember why...

aoldershaw:
@pivotal-jwinters although, with a `RWMutex` alone, we'd need to be careful about only locking around accesses to the `map` only. Thinking about it more, I think we could get away with just adding another layer of locking:

```go
func (bus *notificationsBus) Listen(channel string) (chan bool, error) {
	bus.registrationL.Lock()
	defer bus.registrationL.Unlock()

	bus.notificationL.Lock()
	numSinks := len(bus.notifications[channel])
	bus.notificationL.Unlock()
	if numSinks == 0 {
		err := bus.listener.Listen(channel)
		if err != nil {
			return nil, err
		}
	}

	notify := make(chan bool, 1)

	bus.notificationL.Lock()
	sinks, found := bus.notifications[channel]
	if !found {
		sinks = map[chan bool]struct{}{}
		bus.notifications[channel] = sinks
	}
	bus.notificationL.Unlock()

	sinks[notify] = struct{}{}

	return notify, nil
}
```

Essentially, add a `registrationL` lock that applies for the entire `Listen` and `Unlisten` calls. Only after that lock is acquired can `notificationL` be acquired, and `notificationL` is used to tightly guard only the accesses to `bus.notifications`.

pivotal-jwinters:
Neat!

In this case your initial suggestion of using a sync.map for the notifications might be simpler. It should have the same effect though right?

I still think we should timeout the registrationL since it's accessing a remote resource. Even if it doesn't deadlock anymore it could still block for a long time under heavy load.

aoldershaw:
Using `sync.Map` should have the same effect practically. That said, I'd be inclined to use mutexes so we don't lose type safety/need to do casting/type assertions all over the place.

Rather than having a timeout on `registrationL`, would it make more sense to have a timeout on the remote resource request itself? This might not actually be possible through `pq` given the existence of https://github.com/Kount/pq-timeouts

agurney:
Since running into the issue, we've looked at exactly @aoldershaw 's suggestion of replacing the 'big' mutex with a `sync.Map` - doing some local testing of that currently. As mentioned, it's a bit of an annoyance to try to reproduce, and I'm also anxious about the whole thing having correct behavior since there is so much concurrency-wrangling going on. I thought about an `RWMutex` but was concerned that it would still deadlock if somebody is holding the "W" lock to listen/unlisten, the `pq` internal buffers are full, and the goroutine that's hoping to iterate can't acquire an "R" lock.

As a side note, I believe timeouts in the DB side have no effect here, because there's no active DB queries while the deadlock occurs. The notifications were duly posted back from Postgres, and are now in a channel waiting to be `select`ed out. It could still be a useful thing in other scenarios (generically slow DB, etc.). We'd also done some thinking and prelim testing around setting those Postgres timeout settings "globally" for the Concourse user, as a sort of catch-all for unanticipated database-related oddities.

aoldershaw:
@agurney I think the key to using mutexes here is to not hold the mutex for the duration of the `Listen` and `Unlisten` calls, since those could block arbitrarily long - see my [2nd comment](https://github.com/concourse/concourse/pull/5500#issuecomment-620238688) for how this could look. `sync.Mutex` should work just as well.

Just checking, you mentioned you were running into the deadlock every couple of days before - have you run into it since using `sync.Map`?

agurney:
We haven't got the `sync.Map` alternative running just yet - but also seem to have "naturally" stopped having the deadlocks for the time being, for whatever reason. I'm keen to get a better resolution that just hoping it doesn't happen again 😄 

Agreed on having a tighter scope for holding the mutexes.

pivotal-jwinters:
closing in favour of https://github.com/concourse/concourse/pull/5519

PR #5050, opened 2020-01-21
use sigs.k8s.io/yaml everywhere

vito:
these few stragglers seem to have not been switched over

note that we still depend on gopkg.in/yaml.v2, through sigs.k8s.io/yaml:

    $ go mod why gopkg.in/yaml.v2
    # gopkg.in/yaml.v2
    github.com/concourse/concourse/atc
    sigs.k8s.io/yaml
    gopkg.in/yaml.v2

PR #5504, opened 2020-04-27
atc: behavior: introduce new step types

vito:
# TODOne

* [x] `godoc` (to some extent)
* [x] https://github.com/concourse/concourse/pull/5497#discussion_r436238491
* [x] ~~switch to yjson everywhere~~ nevermind, this is probably worth a separate refactor to replace #5050 

These changes should make steps easier to implement correctly and safely. Largely this comes down to removing `PlanConfig`.

* Generally `PlanConfig` was pretty intimidating and had a bunch of cobwebs - stale comments, unused fields... it just kept growing and never shrank.
* `PlanConfig` is a struct that contains literally all fields of all steps. This means we don't get much compiler assistance when adding a new step, because they're all the same type.
* This also means we need to do manual validation to check that you don't e.g. set a `file:` field on a `get:` step. Naturally, this was messy to implement and was not covered for all possible step combinations.
* All being one type makes it impossible for two steps to have a different type for the same field. For example, the `task:` step should really use `atc.TaskEnv`, not `atc.Params`.
* With `PlanConfig` containing all of the fields that were set in a "flat" structure, we have to carefully interpret precedence at runtime between `ensure:`, `on_success:`, `attempts:`, `retry:`, etc. It would be nice if this was just a matter of precedence at parse time.

This PR removes the `PlanConfig` megastruct and replaces it with a stronger set of types and primitives for recursively walking through steps.

Each step now has its own struct type (e.g. `atc.GetStep`) which implements the `StepConfig` interface and is otherwise just a regular old `json.Unmarshal` struct.

A new parsing technique is introduced in the first commit - I'd recommend reading the message and referencing the code. It's a little tricky but should be really low maintenance and hopefully not too slow (considering we only schedule jobs when necessary and don't unmarshal configs all that often now that `ListAllJobs` is straight from the database).

This PR also introduces a `StepVisitor` interface which is used by important things like the validator and build plan factory. Whenever a new step is added to the `StepVisitor` interface the compiler will helpfully tell us where things need to be updated.

*Note: this is marked `behavior:` because the validation messages are technically different now. Otherwise, everything should be the same.*

# Contributor Checklist

- [x] Unit tests
- [x] ~Integration tests~ (this is all covered in `testflight`)

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


atc/configvalidate/validate_test.go
```diff
@@ -661,214 +730,95 @@ var _ = Describe("ValidateConfig", func() {
...
 				config.Jobs = append(config.Jobs, job)
 			})
 
-			It("returns a single error", func() {
+			It("returns an error for each step", func() {
 				Expect(errorMessages).To(HaveLen(1))
 				Expect(errorMessages[0]).To(ContainSubstring("invalid jobs:"))
-				Expect(strings.Count(errorMessages[0], "has get steps with the same name: some-resource")).To(Equal(1))
+				Expect(errorMessages[0]).To(ContainSubstring("jobs.some-other-job.plan.do[1].get(some-resource): repeated name"))
+				Expect(errorMessages[0]).To(ContainSubstring("jobs.some-other-job.plan.do[1].get(some-resource): repeated name"))
```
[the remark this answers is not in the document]
vito: yep - good catch!

atc/configvalidate/validate_test.go
```diff
@@ -661,214 +730,95 @@ var _ = Describe("ValidateConfig", func() {
...
-							InParallel: nil,
-						})
-
-						config.Jobs = append(config.Jobs, job)
-					})
-
-					It("returns an error", func() {
-						Expect(errorMessages).To(HaveLen(1))
-						Expect(errorMessages[0]).To(ContainSubstring("invalid jobs:"))
-						Expect(errorMessages[0]).To(ContainSubstring("jobs.some-other-job.plan[0] has multiple actions specified (get, put)"))
```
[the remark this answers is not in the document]
vito: Good point. I pushed a change that puts the step type to the message, at least, so it now reads like so:

```
unmarshal converted json: malformed get step: json: unknown field "bogus"
```

Still not great, but it doesn't look like there's an easy way to show anything more. :/ We won't have an accurate line number to report since the YAML is first converted to JSON. It'd be nice if `encoding/json` returned more semantic information in these cases.

vito:
@aoldershaw Marking this as ready for review! I believe I've addressed all of the comments so far, and now I'm just going back and adding `godoc` to newly introduced types. All tests are passing.

@jamieklassen FYI, your concern from https://github.com/concourse/concourse/pull/5497#discussion_r433397082 is addressed by https://github.com/concourse/concourse/pull/5504/commits/ebf3065ee1422061965089ead6bf8d5b3053bd1f (`atc/builds: structure: don't test precedence` in case that link gets rebased away)

atc/steps.go
```diff
@@ -30,7 +30,10 @@ func (step *Step) UnmarshalJSON(data []byte) error {
 
 		err := step.ParseJSON(data)
 		if err != nil {
-			return err
+			return MalformedStepError{
+				StepType: "get",
```
aoldershaw, line 61: Should this be `s.Key`?

```suggestion
				StepType: s.Key,
```

atc/steps.go
```diff
@@ -8,10 +8,37 @@ import (
...
+// variable.
+//
+// First, the data is unmarshalled into a map[string]*json.RawMessage. Next,
+// UnmarshalJSON loops over StepPrecedence.
+//
+// For any StepDetector with a .Key field present in the map, .New is called to
+// construct an empty StepConfig, and then .ParseJSON is called on it to parse
+// the data. For step modifiers like `timeout:' and `attempts:', this will be a
+// non-strict parse that only parses their fields, ignoring the rest. For core
+// step types, this will be a strict parse, raising an error on any known
```
aoldershaw, line 32:
```suggestion
// step types, this will be a strict parse, raising an error on any unknown
```

atc/steps.go
vito: Hahaha. Yes. (Hey, it passed the tests!)

vito:
Thanks for the reviews!

PR #5511, opened 2020-04-28
topgun/k8s: run dns proxy tests against containerd

### LINKED ISSUE #5473: Enable topgun/k8s DNS proxy test suite with containerd as backend

muntac:
# Why is this PR needed?

Fixes #5473.

The entire DNS Proxy test suite can now be run either using `guardian` or `containerd` simply by passing a `useContainerd` flag which will set the `--use-containerd` flag on the worker.

# Reviewer Checklist

- [ ] Tests reviewed
- [ ] PR acceptance performed

### DIFF
--- a/topgun/k8s/dns_proxy_test.go
+++ b/topgun/k8s/dns_proxy_test.go
@@ -14,10 +14,13 @@ var _ = Describe("DNS Resolution", func() {
 		setReleaseNameAndNamespace("dp")
 	})
 
-	var setupDeployment = func(dnsProxyEnable, dnsServer string) {
+	var setupDeployment = func(useContainerd bool, dnsProxyEnable, dnsServer string) {
 		args := []string{
 			`--set=worker.replicas=1`,
 			`--set-string=concourse.worker.garden.dnsProxyEnable=` + dnsProxyEnable,
+		}
+		if useContainerd {
+			args = append(args, "--set=worker.garden.useContainerd=true")
 		}
 		if dnsServer != "" {
 			args = append(args,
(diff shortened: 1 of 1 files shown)


cirocosta:
iiinteresting approach, initially I thought a way of getting this would be to have the whole suite running against `containerd` and `garden` in different jobs (by populating the default flags  - https://github.com/concourse/concourse/blob/975ec8dcbfac2b04635cc5ca9648e09627a2bab7/topgun/k8s/k8s_suite_test.go#L265-L285 - it'd then use containerd everywhere in one, and not in the other)

did you explore that approach? it seems similar to what you outlined in the first approach there in the issue, but leveraging an environment variable that could indicate whether we use containerd or not, and then using that in the default set of `--set` params for helm; it'd not be specific to dns, but would cover it + possibly showcase other areas where we might not be getting containerd completely right yet :thinking:  

thanks!

muntac:
@cirocosta @YoussB  Sorry, I should've linked a [comment](https://github.com/concourse/concourse/issues/5473#issuecomment-620064070) we put on the issue after thinking through this further.

So we have two options 1) run all the suites under topgun-k8s with both `guardian` and `containerd`, and 2) run specific test suites OR specific specs in the suite against `containerd`.

We noticed that about half of the suites do not test any runtime specific functionality (e.g. `external_postgrest_test`, `web_scaling_test`). This means running them all against two different runtimes would not provide additional value, while slowing down the overall suite*.

Secondly, we noticed that it's possible some `guardian` specific tests might not be run in the same way with `containerd` (e.g. they need different flags). So having differences at the individual suite and spec level are ultimately unavoidable.

Lastly, that there will be `containerd` specific tests eventually in topgun-k8s as `containerd` specific functionality gets introduced. So even if we were succesfull in pursuing option 1 right now, we'd eventually have to refactor in a big way and it seemed easier to make specific changes at suite level as we move forward. 

* this was under the assumption that once containerd is production ready we wouldn't want a separate job in the pipeline for k8s-containerd. I forget the exact reasoning @pivotal-jamie-klassen provided for this at the time. but one of the advantages is being able to run the overall suite in one go locally (@YoussB this might answer your question).

YoussB:
@muntac sorry for the late response.

from what I understand, it seems that both tests are going to diverge eventually, would it make any sense to keep the 2 tests in their separate files (or maybe suites) and use helper functions with the tests?

not sure if this could be an anti-pattern, let me know what you think ?

taylorsilva:
Looking at the previous discussion I decided to see which tests suites under topgun/k8s would make sense to have both garden and containerd versions. (tldr: not many)

There are 15 test suites inside topgun/k8s. 
❌ = run with garden and containerd does not add value
✅ = run with garden and containerd does add value
* baggageclaim_drivers_test - ❌
* container_limits_test - ✅
* dns_proxy_test - ✅
* ephemeral_worker_test - ✅ (this might actually be an ❌, is there some container cleanup we care about? most of the things this suite may test may be on the web side)
* external_postgres_test - ❌
* external_worker_test - ❌
* garden_config_test - ✅ (would probably add another suite for containerd)
* https_web_tls_termination - ❌
* kubernetes_creds_mgmt_test - ❌
* mainteam_role_test - ❌ 
* prometheus_test - ❌ 
* rebalance_test - ❌ 
* tsa_node_port_test - ❌ 
* web_scaling_test -❌ 
* worker_lifecycle_test - ✅  (we care a lot about cleaning up volumes; not sure if we have the same concern for containers)

Glancing at this I'm inclined to agree with @muntac that running all tests with containerd and garden would greatly increase the time to run the test suite and not add much or any value at all. Therefore I'm fine to go along with the current approach. It feels very similar to what we did with the baggageclaim tests for GKE vs PKS.

I also do think that it's possible, though **extremely** unlikely, that containerd may cause some unknown behaviour with something like external workers. I don't think this thought warrants going with the topgun/k8s suite-wide approach of testing garden and containerd.

topgun/k8s/dns_proxy_test.go
```diff
@@ -52,51 +55,61 @@ var _ = Describe("DNS Resolution", func() {
...
+		)
+	}
+
+	Context("with gdn backend", func() {
+		expectedDnsProxyBehaviour(false)
+	})
+
+	Context("with containerd backend", func() {
+		expectedDnsProxyBehaviour(true)
+	})
```
taylorsilva, line 114: nit: could we move these up to the top so this test suite follows the flow of:

* BeforeEach()
* AfterEach()
* Context()

PR #5517, opened 2020-04-29
Redundant Volume Streaming

### LINKED ISSUE #5485: Image and Input Volumes provided by a resource maybe streamed redundantly for a Task
## Summary

If a Task references a previous step (eg. Get step fetching an image ) for its image and ends up being scheduled on worker `ABC`, while the previous step was scheduled on worker `123`, the volume for the image will be streamed from `123` to `ABC` even if the image cache exists on `ABC`.

This also impacted streaming of inputs that were backed by a resource.

## Steps to reproduce
- Setup Concourse with multiple workers ( was able to reproduce with docker compose )
- Fly a pipeline with a Task whose image references the output of a previous step
```
resources:
- name: apl
  type: registry-image
  source:
      repository: concourse/git-resource
      tag: dev-ubuntu
jobs:
- name: test-image-as-resource
  plan:
  - get: apl
    trigger: true
  - task: echo
    image: apl
    config:
        platform: linux
        run:
            path: sh
            args:
            - -c
            - |
                echo "foo"
```
- Execute builds until the resource cache volume exists on all workers
For example, with 3 workers
```
$ fly -t dev volumes
handle                                worker        type           identifier
2cbe1c52-4db0-4194-6ff6-9bcaeb18335c  31a442213347  resource       digest:sha256:cd47ec95bd17cd15382e4894c9a4be0e22cb2d3a7a60ea71aad9ace005fd1826
c7782d3c-cbb2-4320-77d0-3aa81416e63e  6ff99779506a  resource       digest:sha256:cd47ec95bd17cd15382e4894c9a4be0e22cb2d3a7a60ea71aad9ace005fd1826
26b6b62c-59bc-4be9-70da-abc3304a6743  9db63a79bb32  resource       digest:sha256:cd47ec95bd17cd15382e4894c9a4be0e22cb2d3a7a60ea71aad9ace005fd1826

```
- As the cache exists on all 3 workers, subsequent builds shouldn't require the `Task`'s image to be streamed. However, we observed via log statements and observing the initialization time for the `Task` step that it was indeed streaming the volume from the `123` that was chosen for the `Get` step to `ABC` chosen for the Task.

Relevant source code
- [client uses image referenced via previous step](https://github.com/concourse/concourse/blob/master/atc/worker/client.go#L250-L255)

[Original Discussion](https://github.com/concourse/concourse/discussions/5378)

## Expected results
When the Task's worker has the image volume required by the task, the volume is not streamed from the previous step's worker.

## Actual results
Concourse may redundantly stream volumes when
- Task's image is referencing the output of a previous Get step 
- For a given build, the Task and Get steps end up on different workers.

## Triaging info

* Concourse version: v6.0.0
* Did this used to work? Yes

xtreme-sameer-vohra:
closes #5485 

# Contributor Checklist

- [x] Unit tests
- [ ] Integration tests

# Reviewer Checklist

- [x] Code reviewed
- [x] Tests reviewed
- [x] Release notes reviewed
- [x] PR acceptance performed

### DIFF
--- a/atc/db/resource_cache_factory.go
+++ b/atc/db/resource_cache_factory.go
@@ -27,6 +27,8 @@ type ResourceCacheFactory interface {
 	// method can be removed at that point. See  https://github.com/concourse/concourse/issues/534
 	UpdateResourceCacheMetadata(UsedResourceCache, []atc.MetadataField) error
 	ResourceCacheMetadata(UsedResourceCache) (ResourceConfigMetadataFields, error)
+
+	FindResourceCacheByID(id int) (UsedResourceCache, bool, error)
 }
 
 type resourceCacheFactory struct {
@@ -119,6 +121,17 @@ func (f *resourceCacheFactory) ResourceCacheMetadata(resourceCache UsedResourceC
 	}
 
 	return metadata, nil
+}
+
+func (f *resourceCacheFactory) FindResourceCacheByID(id int) (UsedResourceCache, bool, error) {
+	tx, err := f.conn.Begin()
+	if err != nil {
+		return nil, false, err
+	}
+
+	defer Rollback(tx)
+
+	return findResourceCacheByID(tx, id, f.lockFactory, f.conn)
 }
 
 func findResourceCacheByID(tx Tx, resourceCacheID int, lock lock.LockFactory, conn Conn) (UsedResourceCache, bool, error) {
--- a/atc/worker/volume.go
+++ b/atc/worker/volume.go
@@ -26,6 +26,7 @@ type Volume interface {
 	COWStrategy() baggageclaim.COWStrategy
 
 	InitializeResourceCache(db.UsedResourceCache) error
+	GetResourceCacheID() int
 	InitializeTaskCache(logger lager.Logger, jobID int, stepName string, path string, privileged bool) error
 	InitializeArtifact(name string, buildID int) (db.WorkerArtifact, error)
 
@@ -112,6 +113,10 @@ func (v *volume) COWStrategy() baggageclaim.COWStrategy {
 
 func (v *volume) InitializeResourceCache(urc db.UsedResourceCache) error {
 	return v.dbVolume.InitializeResourceCache(urc)
+}
+
+func (v *volume) GetResourceCacheID() int {
+	return v.dbVolume.GetResourceCacheID()
 }
 
 func (v *volume) InitializeArtifact(name string, buildID int) (db.WorkerArtifact, error) {
--- a/atc/db/dbfakes/fake_created_volume.go
+++ b/atc/db/dbfakes/fake_created_volume.go
@@ -56,6 +56,16 @@ type FakeCreatedVolume struct {
 		result1 db.DestroyingVolume
 		result2 error
 	}
+	GetResourceCacheIDStub        func() int
+	getResourceCacheIDMutex       sync.RWMutex
+	getResourceCacheIDArgsForCall []struct {
+	}
+	getResourceCacheIDReturns struct {
+		result1 int
+	}
+	getResourceCacheIDReturnsOnCall map[int]struct {
+		result1 int
+	}
 	HandleStub        func() string
 	handleMutex       sync.RWMutex
 	handleArgsForCall []struct {
@@ -420,6 +430,58 @@ func (fake *FakeCreatedVolume) DestroyingReturnsOnCall(i int, result1 db.Destroy
 		result1 db.DestroyingVolume
 		result2 error
 	}{result1, result2}
+}
+
+func (fake *FakeCreatedVolume) GetResourceCacheID() int {
+	fake.getResourceCacheIDMutex.Lock()
+	ret, specificReturn := fake.getResourceCacheIDReturnsOnCall[len(fake.getResourceCacheIDArgsForCall)]
+	fake.getResourceCacheIDArgsForCall = append(fake.getResourceCacheIDArgsForCall, struct {
+	}{})
+	fake.recordInvocation("GetResourceCacheID", []interface{}{})
+	fake.getResourceCacheIDMutex.Unlock()
+	if fake.GetResourceCacheIDStub != nil {
+		return fake.GetResourceCacheIDStub()
+	}
+	if specificReturn {
+		return ret.result1
+	}
+	fakeReturns := fake.getResourceCacheIDReturns
+	return fakeReturns.result1
+}
+
+func (fake *FakeCreatedVolume) GetResourceCacheIDCallCount() int {
+	fake.getResourceCacheIDMutex.RLock()
+	defer fake.getResourceCacheIDMutex.RUnlock()
+	return len(fake.getResourceCacheIDArgsForCall)
+}
+
+func (fake *FakeCreatedVolume) GetResourceCacheIDCalls(stub func() int) {
+	fake.getResourceCacheIDMutex.Lock()
+	defer fake.getResourceCacheIDMutex.Unlock()
+	fake.GetResourceCacheIDStub = stub
+}
+
+func (fake *FakeCreatedVolume) GetResourceCacheIDReturns(result1 int) {
+	fake.getResourceCacheIDMutex.Lock()
+	defer fake.getResourceCacheIDMutex.Unlock()
+	fake.GetResourceCacheIDStub = nil
+	fake.getResourceCacheIDReturns = struct {
+		result1 int
+	}{result1}
+}
+
+func (fake *FakeCreatedVolume) GetResourceCacheIDReturnsOnCall(i int, result1 int) {
+	fake.getResourceCacheIDMutex.Lock()
+	defer fake.getResourceCacheIDMutex.Unlock()
+	fake.GetResourceCacheIDStub = nil
+	if fake.getResourceCacheIDReturnsOnCall == nil {
+		fake.getResourceCacheIDReturnsOnCall = make(map[int]struct {
+			result1 int
+		})
+	}
+	fake.getResourceCacheIDReturnsOnCall[i] = struct {
+		result1 int
+	}{result1}
 }
 
 func (fake *FakeCreatedVolume) Handle() string {
@@ -1099,6 +1161,8 @@ func (fake *FakeCreatedVolume) Invocations() map[string][][]interface{} {
 	defer fake.createChildForContainerMutex.RUnlock()
 	fake.destroyingMutex.RLock()
 	defer fake.destroyingMutex.RUnlock()
+	fake.getResourceCacheIDMutex.RLock()
+	defer fake.getResourceCacheIDMutex.RUnlock()
 	fake.handleMutex.RLock()
 	defer fake.handleMutex.RUnlock()
 	fake.initializeArtifactMutex.RLock()
--- a/atc/db/dbfakes/fake_resource_cache_factory.go
+++ b/atc/db/dbfakes/fake_resource_cache_factory.go
@@ -27,6 +27,21 @@ type FakeResourceCacheFactory struct {
 		result1 db.UsedResourceCache
 		result2 error
 	}
+	FindResourceCacheByIdStub        func(int) (db.UsedResourceCache, bool, error)
+	findResourceCacheByIdMutex       sync.RWMutex
+	findResourceCacheByIdArgsForCall []struct {
+		arg1 int
+	}
+	findResourceCacheByIdReturns struct {
+		result1 db.UsedResourceCache
+		result2 bool
+		result3 error
+	}
+	findResourceCacheByIdReturnsOnCall map[int]struct {
+		result1 db.UsedResourceCache
+		result2 bool
+		result3 error
+	}
 	ResourceCacheMetadataStub        func(db.UsedResourceCache) (db.ResourceConfigMetadataFields, error)
 	resourceCacheMetadataMutex       sync.RWMutex
 	resourceCacheMetadataArgsForCall []struct {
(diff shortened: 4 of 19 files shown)


pivotal-jwinters:
@pivotal-jamie-klassen a lot of what you described really resonates with me. 

We've actually talked about a blob store before, and we always rejected the idea because it was one additional thing that operators would have to deal with. I think the last time it was talked about we settle on the idea of a blobstore interface, whose initial implementation would be backed by volumes under the hood. 

We had a pretty comprehensive war on caches at one point, and a lot of the refactor work in the runtime was always with this goal in mind, but since we dissolved the core/runtime split it's kinda fallen by the wayside. I still think it's super important work, but it really needs someone to champion it.

I keep saying we but at this point I think I'm the only one left that was part of these discussions (@ddadlani @kcmannem 😢)

pivotal-jamie-klassen:
Also before I forget (and hopefully to save a feedback iteration) - please add a relase note!

cirocosta:
I wanted to use #5579 to see if I could visualize the original problem, and indeed, very helpful!

Having the following base `docker-compose.yml`:

```yaml
version: '3'

services:
  db:
    image: postgres:${POSTGRES_TAG:-latest}
    shm_size: 1gb
    ports:
    - 6543:5432
    environment:
      POSTGRES_DB: concourse
      POSTGRES_USER: dev
      POSTGRES_PASSWORD: dev

  web:
    build: .
    volumes:
    - .:/src
    command: web
    depends_on: [db]
    ports:
    - 8080:8080
    environment:
      CONCOURSE_CONTAINER_PLACEMENT_STRATEGY: random
      CONCOURSE_LOG_LEVEL: debug
      CONCOURSE_POSTGRES_HOST: db
      CONCOURSE_POSTGRES_USER: dev
      CONCOURSE_POSTGRES_PASSWORD: dev
      CONCOURSE_POSTGRES_DATABASE: concourse
      CONCOURSE_EXTERNAL_URL: http://localhost:8080
      CONCOURSE_CLUSTER_NAME: dev
      CONCOURSE_ENABLE_ARCHIVE_PIPELINE: "true"

  worker1: &worker
    build: .
    command: worker
    privileged: true
    stop_signal: SIGUSR2
    environment:
      CONCOURSE_LOG_LEVEL: debug
      CONCOURSE_TSA_HOST: web:2222

      # avoid using loopbacks
      CONCOURSE_BAGGAGECLAIM_DRIVER: overlay

      # so we can reach Garden/Baggageclaim for debugging
      CONCOURSE_BIND_IP: 0.0.0.0
      CONCOURSE_BAGGAGECLAIM_BIND_IP: 0.0.0.0

  worker2: *worker
```

which gives us two workers, using `random` as the container-placement strategy and the following pipeline

```yaml
resources:
  - type: registry-image
    name: image
    source: {repository: ubuntu}

jobs:
  - name: test
    plan:
      - get: image
      - task: test-task
        image: image
        config:
          platform: linux
          run:
            path: "true"
```

having ran the pipeline two times, getting the contents cached on both workers:

```console
$ fly -t local volumes | grep 5747
9dae7211-9b8e-45bb-7d73-3f8a3d41ca2d  0e690966f9b0  resource       digest:sha256:5747316366b8cc9e3021cd7286f42b2d6d81e3d743e2ab571f55bcd5df788cc8
8944f15a-ed3c-4299-6137-425a2e06368c  621ecba988ee  resource       digest:sha256:5747316366b8cc9e3021cd7286f42b2d6d81e3d743e2ab571f55bcd5df788cc8
```

I could then visualize the streaming taking place:

Screenshot from 2020-05-11 14-43-03

---

Will review the code soon :+1:

atc/worker/volume.go
```diff
@@ -26,6 +26,7 @@ type Volume interface {
 	COWStrategy() baggageclaim.COWStrategy
 
 	InitializeResourceCache(db.UsedResourceCache) error
+	GetResourceCacheId() int
```
cirocosta, line 29: was there a reason to make it `Id` instead of `ID`? I noticed that there's another `GetResourceCacheID` in this PR, but this one had that `d` lowercased 

```suggestion
	GetResourceCacheID() int
```

xtreme-sameer-vohra:
> Before getting into it, I'm marking this as 'Request changes' because even if the changes are all perfect, a few files need a good `gofmt` and at least one method could use a rename. Left as an exercise to the reader.
> 
> OK, I've spent a good hour and half poring over this and taking tons of notes as I try to understand the design. @xtreme-sameer-vohra Can you help me understand some of the choices here? In particular, I was really thrown by the addition of the `GetResourceCacheID` method on `CreatedVolume`:
> 
> https://github.com/concourse/concourse/pull/5517/files#diff-d48a5c0dafa10cc74fe02a6a204ed073R397-R399
> 
> In my mind it just sounds so wrong, like caches are a "high-level"/"domain" concept but volumes are a "low-level"/"solution" concept so caches could know about volumes but not the other way around. When I look at the terms I could swear a `resource cache` 'is-a' `volume`, or maybe 'has-a' `volume`, but surely not all volumes are resource caches, right? some volumes are inputs, some are outputs, some are rootfses, some are cert-bind-mounts, some are task caches. I'm guessing the `resourceCacheID` for those volume types would just be zero. What am I missing here? can you help me understand?
> 
> To give a bit more context, I instinctively feel like there's a kind of conceptual duplication between the `imageSpec` method on a `taskStep` in the `exec` package and the `ExistsOn` method on a `streamableArtifactSource` in the `worker` package. It seems like they both have this parallel boolean around whether the desired volume is "already around" due to a resource already having fetched it. Is there a way for these decisions to be made in the same place? This is a bit of a wild thought, but if there were a new way to get artifacts, like outside of resource caches or other steps in a build -- like maybe concourse had first-class support for a blobstore or registry to get them from -- I can imagine implementing support for it would require changes in both the `exec` and `worker` packages, which seems like shotgun surgery. It's a super long shot, sorry if that one doesn't make sense.

Our primary driver for the design was to maintain the boundary between `core` & `runtime` and not regress by having `core` depend on the `worker`. 
Specifically, this translated to trying to not making additional `core` abstractions such as `Steps`,`Resource Caches`, etc depend upon `worker` abstractions such as `Volumes` with a substantial interface. 

This creation of boundaries was started in the original step refactor by introducing `ArtifactSource` as an interface that is created & consumed by `runtime` but only ferried along by `core` in between steps. [Refactor Issue #4265](https://github.com/concourse/concourse/issues/4265)

This is still not complete and there still exist multiple `core` components that deal directly with `volumes`. [Issue #3607](https://github.com/concourse/concourse/issues/3607)
Our initial refactor was the smallest conceptual step we could foresee and it took a substantial amount of time. IMHO this is going to take a while to clean up unless we decide to halt everything else to deal with it. 

In the context of volumes and the view of Concourse as a single codebase, it is quite ugly for it to expose its volume type aka the ResourceCacheID. However, I propose that we maintain a boundary between `core` and `runtime`, where it would be preferred for `runtime` objects to depend and know about `core` ones but not the other way around. If that makes sense, then `runtime` needs to manage the one:many mapping between something `core` considers cacheable (eg. Resource Cache ) and the many copies of which may exist in the current runtime implementation (decentralized baggageclaim cluster). In the scope of this bug fix, this translated to exposing the ResoureCacheID on the volume. Also, this bug not only affects images for tasks but also all its inputs that may be on another worker. In the longer term, the tracking the relationship between a `volume` and a `Resource Cache` should be moved out of the volume and potentially into the `runtime.Artifact`. However, we'd also want to clean up the underlying data model which is quite coupled today. 

I agree with Josh that this refactoring is very important work, however, the driver for this was the [K8s Runtime](https://github.com/concourse/concourse/projects/16) which isn't currently prioritized. With only a single pair on runtime epics, this would get picked back up after the `containerd` epic.

atc/db/resource_cache_factory.go
```diff
@@ -27,6 +27,8 @@ type ResourceCacheFactory interface {
 	// method can be removed at that point. See  https://github.com/concourse/concourse/issues/534
 	UpdateResourceCacheMetadata(UsedResourceCache, []atc.MetadataField) error
 	ResourceCacheMetadata(UsedResourceCache) (ResourceConfigMetadataFields, error)
+
+	FindResourceCacheById(id int) (UsedResourceCache, bool, error)
```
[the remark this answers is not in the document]
cirocosta: >  [@pivotal-jamie-klassen]  like caches are a "high-level"/"domain" concept but volumes are a "low-level"/"solution" concept so caches could know about volumes but not the other way around

:+1: on that

in practice *as is* right now, volumes *do* know about the existence of caches (although directly only with regards to workers)

```
SELECT * FROM volumes;
 id |                handle                | worker_name  |  worker_resource_cache_id 
----+--------------------------------------+--------------+---------------------------
  8 | 8944f15a-ed3c-4299-6137-425a2e06368c | 621ecba988ee |                         1 
 10 | 9dae7211-9b8e-45bb-7d73-3f8a3d41ca2d | 0e690966f9b0 |                         2 
(6 rows)



SELECT * FROM worker_resource_caches;
 id | resource_cache_id
----+-------------------
  1 |                 1
  2 |                 1
(2 rows)


SELECT * FROM resource_caches;

 id | resource_config_id |               version                  |             metadata   
----+--------------------+----------------------------------------------------------------------------------
  1 |                  1 | {"digest": "sha256:5747316366b8cc..."} | [{"name":"repository","value":"ubuntu"}] 
```

*ps/: many columns redacted in those `select`s above*

but .. I completely agree on making volumes just the lower level thing that a cache reference (buut, I'm not suuuper familiar with its details, so, take that with a grain of salt)

cirocosta:
oh I didn't see @xtreme-sameer-vohra 's comment, but 

> [@xtreme-sameer-vohra] However, we'd also want to clean up the underlying data model which is quite coupled today.

is essentially what I wanted to bring up with that comment I made above showcasing the `worker_resource_cache_id` in the `volumes` table.

cirocosta:
@xtreme-sameer-vohra , is this approach different from what we had before 6.0? 

I'm curious if the reason why we regressed is that we didn't want a particular pattern and now we're getting that undesired pattern in by fixing the regression :thinking:

atc/worker/volume.go
xtreme-sameer-vohra: fixed

xtreme-sameer-vohra:
The interface `ArtifactSource` is quite similar between [5.8.x](https://github.com/concourse/concourse/blob/release/5.8.x/atc/worker/artifact_source.go#L14) & [6.1.x](https://github.com/concourse/concourse/blob/master/atc/worker/artifact_source.go#L27)

The previous [implementation getArtifactSource](https://github.com/concourse/concourse/blob/release/5.8.x/atc/exec/get_step.go#L250) was composed of 
* [resourceInstance](https://github.com/concourse/concourse/blob/release/5.8.x/atc/resource/resource_instance.go#L16) - Part of the `resource pkg`. This was used to query a volume for a resource cache on a worker and thus knew about `workers` & `volumes`
* [versionedSource] (https://github.com/concourse/concourse/blob/release/5.8.x/atc/resource/versioned_source.go#L14) -Part of the `resource pkg` and knew about `volumes`

The current [implementation artifactSource](https://github.com/concourse/concourse/blob/master/atc/worker/artifact_source.go#L39) is composed of
* [artifact](https://github.com/concourse/concourse/blob/master/atc/runtime/types.go#L38:1) 
* [volume](https://github.com/concourse/concourse/blob/master/atc/worker/volume.go#L14) 

The changes made from v5.8 to v6.0 were to reduce the references to `volume` in `core` components such as `resources`, which ultimately would set up for multiple `runtime` implementations. 
Its fair to consider extending the `artifactSource` or `artifact` to hold the relationship between a `Resource` and `Volume` which requires tracking `UsedResourceCache`, rather than exposing it on the `volume`. But just solely doing that belies the underlying data model.

jamieklassen:
I suspect that this PR is good, but I won't be satisfied until I can actually manually verify that it works. Doing that is much easier said than done. Here's what I have thought about so far:

1. I'll need a cluster with two workers, `wa` and `wb`.
2. I want a `get` step for the same resource `r`'s version `v` to have already run on both workers -- i.e. they both have warm caches.
3. I will run a task on one of the workers -- assume `wa` (wlog) -- where the `image` refers to `r`. This is the scenario which should be slow on master (because it will unnecessarily stream `v` from `wb`) but fast with this PR (which will reuse the cache on `wa` mentioned in 2.).

Now I'm trying the following concrete steps to achieve this:
1. ```bash
    docker-compose -f <(cat <<EOF
    version: '3'
    services:
      db:
        image: postgres:${POSTGRES_TAG:-latest}
        shm_size: 1gb
        ports:
        - 6543:5432
        environment:
          POSTGRES_DB: concourse
          POSTGRES_USER: dev
          POSTGRES_PASSWORD: dev
      web:
        build: .
        volumes:
        - .:/src
        command: web
        depends_on: [db]
        ports:
        - 8080:8080
        environment:
          CONCOURSE_LOG_LEVEL: debug
          CONCOURSE_POSTGRES_HOST: db
          CONCOURSE_POSTGRES_USER: dev
          CONCOURSE_POSTGRES_PASSWORD: dev
          CONCOURSE_POSTGRES_DATABASE: concourse
          CONCOURSE_EXTERNAL_URL: http://localhost:8080
          CONCOURSE_CLUSTER_NAME: dev
          CONCOURSE_ENABLE_ARCHIVE_PIPELINE: "true"
          CONCOURSE_CONTAINER_PLACEMENT_STRATEGY: random
      worker-1:
        build: .
        command: worker
        privileged: true
        depends_on: [web]
        stop_signal: SIGUSR2
        environment:
          CONCOURSE_TAG: worker-1
          CONCOURSE_LOG_LEVEL: debug
          CONCOURSE_TSA_HOST: web:2222
          CONCOURSE_BAGGAGECLAIM_DRIVER: overlay
          CONCOURSE_BIND_IP: 0.0.0.0
          CONCOURSE_BAGGAGECLAIM_BIND_IP: 0.0.0.0
      worker-2:
        build: .
        command: worker
        privileged: true
        depends_on: [web]
        stop_signal: SIGUSR2
        environment:
          CONCOURSE_TAG: worker-2
          CONCOURSE_LOG_LEVEL: debug
          CONCOURSE_TSA_HOST: web:2222
          CONCOURSE_BAGGAGECLAIM_DRIVER: overlay
          CONCOURSE_BIND_IP: 0.0.0.0
          CONCOURSE_BAGGAGECLAIM_BIND_IP: 0.0.0.0
    EOF
    ) up --build -d
    ```
1. wait for web node to listen
1. `fly -t dev login -c http://localhost:8080 -u test -p test`
1. ```bash
    fly -t dev set-pipeline -n -p pipeline -c <(cat <<EOF
    resources:
    - name: image
      type: mock
      tags: [worker-1]
      source: {mirror_self: true}
    jobs:
    - name: job
      plan:
      - get: image
        tags: [worker-1]
    EOF
    )
    ```
1. `fly -t dev unpause-pipeline -p pipeline`
1. `fly -t dev trigger-job -w -j pipeline/job`
1. Reconfigure the same pipeline:
    ```bash
    fly -t dev set-pipeline -n -p pipeline -c <(cat <<EOF
    resources:
    - name: image
      type: mock
      tags: [worker-1]
      source: {mirror_self: true}
    jobs:
    - name: job
      plan:
      - get: image
        tags: [worker-2] # <-- this is the diff
    EOF
    )
    ```
1. `fly -t dev trigger-job -w -j pipeline/job`
1. Now that both resource caches are populated, execute the task:
    ```bash
    fly -t dev execute --tag worker-1 -j pipeline/job --image image -c <(echo '{platform: linux, run: {path: "false"}}')
    ```

Here I thought I was being terribly clever by using tagged workers and `fly execute` to make the container placement deterministic, but it hasn't seemed to work. I will update if I think of a better way.

jamieklassen:
OK so I don't forget I'm publicly noticing that two of the test builds for this PR failed in an unusual way:
* https://ci.concourse-ci.org/teams/contributor/pipelines/prs/jobs/watsjs/builds/36
* https://ci.concourse-ci.org/teams/contributor/pipelines/prs/jobs/testflight/builds/36

I'm rerunning them now, but that error

```
pq: insert or update on table "volumes" violates foreign key constraint "volumes_parent_id_fkey"
```

is completely unfamiliar to me.

vito:
@jamieklassen Interesting, seems like a race condition - that error means we tried to insert a row, but the ID corresponded to a volume that doesn't exist anymore. I'll open a bug, we should be resilient to that.

PR #5520, opened 2020-04-30
Log db transaction query

xtremerui:
# Why is this PR needed?
when `log-db-queries` is enabled, only queries that made by db connection are logged. Those queries made by transactions are missing in the log output.

Show all db queries in log output

Add an intercept layer to log queries made by db transaction similar to the way it intercept db connection query.

# Reviewer Checklist
- [ ] Code reviewed
- [ ] PR acceptance performed


atc/db/log_conn.go
```diff
@@ -21,21 +21,41 @@ type logConn struct {
...
+	Tx
+
+	logger lager.Logger
+}
+
+func (t *logDbTx) QueryRow(query string, args ...interface{}) squirrel.RowScanner {
+	t.logger.Debug("tx-query-row", lager.Data{"query": strip(query)})
+	return t.Tx.QueryRow(query, args...)
+}
+
```
xtreme-sameer-vohra, line 34: Should we also wrap the `Tx.Query` func ? There are a few uses of it in the concourse code base.

xtreme-sameer-vohra: Ditto for `Tx.Exec` & `Tx.QueryRowContext`

xtremerui: OK If there are usage then we should probably wrap them too.

xtremerui: Updated

PR #5522, opened 2020-04-30
atc: fix race condition in notifications bus test

pivotal-jwinters:
Fixes a race condition in the notifications bus
Adds a regression test for the deadlock


pivotal-jwinters:
@aoldershaw fixed. Thanks!!

PR #5529, opened 2020-04-30
web: improve dashboard UX when ListAllJobs is disabled

### LINKED ISSUE #5470: web: dashboard: show blank pipeline cards when ListAllJobs is disabled
Given I am running my web node with `--concurrent-request-limit ListAllJobs=0`
When I load the dashboard from this web node
I see pipeline cards but no jobs displayed

pivotal-jamie-klassen:
# Why is this PR needed?

The raw functionality implemented in #5429, especially when you set the concurrency limit on the `ListAllJobs` endpoint to 0, provides pretty terrible UX - the dashboard will simply show the turbulence view all the time!

This change allows you to still see the pipeline cards on the dashboard when that endpoint is disabled, but with no job previews. Since there is a pipeline-scoped `ListJobs` endpoint, we added a 'sync' icon that tells you there is no job data available for each pipeline.

closes #5470.

* show empty placeholder
* show sync icon
* show 'no data' status text

# Contributor Checklist

- [x] Unit tests
- [x] ~Integration tests~
- [x] ~Updated documentation (located at https://github.com/concourse/docs)~ We still don't have a solid section of the docs dedicated to describing the UI -- hopefully the tooltip makes this change self-documenting enough.

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


web/elm/src/Dashboard/Models.elm
```diff
@@ -30,7 +31,7 @@ type alias Model =
             , isTeamsRequestFinished : Bool
             , isPipelinesRequestFinished : Bool
             , isResourcesRequestFinished : Bool
-            , isJobsErroring : Bool
+            , jobsError : Maybe FetchError
```
pivotal-jamie-klassen, line 34: It is probably worth making a similar change to the other `is<endpoint>Erroring` fields for consistency.

pivotal-jamie-klassen:
While describing the way the dashboard is rendered to @taylorsilva, I realized I consistently use the word 'Team' to describe what is called a 'Group' in the code - this is unnecessarily confusing. Now is the time to rename `Dashboard.Group` to `Dashboard.Team`.

pivotal-jamie-klassen:
While talking through the feature with @taylorsilva, we uncovered another unforeseen consequence of the design: There is not a clean way to represent having a pipeline that is paused _and_ has not job preview because the ListAllJobs endpoint is disabled. The existing designs suggest that, in this state, we should simply show the "jobs disabled" view. Conceptually, the problem with this is that whether or not the pipeline is paused is not affected by the pipeline's jobs. There's no good reason _not_ to show users that the pipeline is paused in this case.

We will now explore the architectural consequences of a design like this:
paused_with_jobs_disabled

Before we forget, there are similar concerns around interactions between stale data and disabled job previews - what happens if you cached some job data and then the endpoint gets disabled, or if you cache the 501 response. These are minor concerns, but any refactoring in the service of the above use-case could be relevant here.

pivotal-jamie-klassen:
While exploring the paused-and-unpreviewable flow, we discovered a different approach to the earlier refactoring around stale data. This is being pursed on https://github.com/concourse/concourse/pull/new/stale-field-refactor.

web/elm/src/Dashboard/Models.elm
pivotal-jamie-klassen: did it

aoldershaw:
A few minor styling comments:

Screen Shot 2020-05-11 at 1 17 51 PM

1. I think the sync icon is too bright - I think it should be more similar to the other icons when not hovering (which, looking at it more closely, appear to be different shades of grey from each other...)
2. The text doesn't quite appear to be vertically centered with the icon. This may be the case with the other icons as well, but it seems to be particularly noticeable with this one.

web/elm/src/Dashboard/Dashboard.elm
```diff
@@ -254,7 +255,29 @@ handleCallback callback ( model, effects ) =
...
+                    if status.code == httpStatusNotImplemented then
+                        ( { model
+                            | jobsError = Just Disabled
+                            , jobs = Fetched Dict.empty
+                            , pipelines =
+                                model.pipelines
+                                    |> Maybe.map
+                                        (List.map
+                                            (\p ->
+                                                { p | jobsDisabled = True }
```
aoldershaw, line 269: This will clobber any "click to sync" overrides of `jobsDisabled`, which obviously isn't a concern right now, but it reminded me: should we stop hitting the ListAllJobs endpoint after the first 501? These calls aren't expensive on the ATC, but it probably makes sense to stop trying if it's easy, which I think it should be. You have a better sense of the `Dashboard.RequestBuffer` module, but if I'm not mistaken you could modify this line:

https://github.com/concourse/concourse/blob/3ac5db508adb9db262f52857d6d23ea2c92f6c7e/web/elm/src/Dashboard/Dashboard.elm#L154

to something like:

```elm
        (\model -> model.dragState /= NotDragging && model.jobsError /= Just Disabled)
```

web/elm/src/Dashboard/Dashboard.elm
```diff
@@ -848,15 +855,7 @@ welcomeCard session { pipelines } =
...
-            case pipelines of
-                None ->
-                    False
-
-                Cached p ->
-                    List.isEmpty p
-
-                Fetched p ->
-                    List.isEmpty p
+            pipelines |> Maybe.map List.isEmpty |> Maybe.withDefault False
```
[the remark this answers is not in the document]
pivotal-jamie-klassen: haha nice catch. it doesn't matter what you put there because that function actually never gets called when `pipelines` is `Nothing` -- it's dead code! ccdfe5f167b37f37d0bd5ceb0f964a17856de6da

web/elm/src/Dashboard/Filter.elm
```diff
@@ -118,10 +118,10 @@ pipelineFilter pf jobs existingJobs pipeline =
         Status sf ->
             case sf of
                 PipelineStatus ps ->
-                    pipeline |> Pipeline.pipelineStatus False jobsForPipeline |> equal ps
```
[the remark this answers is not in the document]
pivotal-jamie-klassen: it took a few tries to get it right - even though everybody has an informal sense of it, a "pipeline status" is an ugly emergent notion that doesn't appear in any literal form in the core of concourse.

pivotal-jamie-klassen:
@aoldershaw

> 1. I think the sync icon is a bit too bright - it should be more similar to the other icons when not hovering IMO (which, looking at it more closely, appear to be different shades of grey from each other...)

The greys are more consistent in 4a8ab80a1a8d7326b37bf97cb328837f79d6fcdf. You're right that the pause and visibility toggles are different colors - this is because the visibility toggle actually has three states of opacity:

1. hidden and unhovered
1. hidden and hovered = exposed and unhovered
1. exposed and hovered

this has to do with the icon doing double duty as a status indicator and a call to action. This may be a bit clunky, but if we decide it needs to change I'd prefer it have its own focused design discussion.

web/elm/src/Dashboard/Dashboard.elm
```diff
@@ -254,7 +255,29 @@ handleCallback callback ( model, effects ) =
...
+                            , pipelines =
+                                model.pipelines
+                                    |> Maybe.map
+                                        (List.map
+                                            (\p ->
+                                                { p | jobsDisabled = True }
+                                            )
+                                        )
+                          }
+                        , effects ++ [ SaveCachedJobs [] ]
```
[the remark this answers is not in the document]
pivotal-jamie-klassen: It was not intentional! Delete is more like what I wanted -- 1a218b8ae8da733c030b10845305f13e2ab6b994.

web/elm/src/Dashboard/Dashboard.elm
pivotal-jamie-klassen: That feels reasonable. If I don't use that crazy over-engineered thing I made now, what is it even good for? 😝 942f5c74aff07f01dc6bf689be085b83946f9c69

pivotal-jamie-klassen:
@aoldershaw rebased!

aoldershaw:
Hmm, watsjs seems to be consistently failing on "shows pipelines in their correct order": https://ci.concourse-ci.org/teams/contributor/pipelines/prs/jobs/watsjs/builds/8#L5eb9f762:284

pivotal-jamie-klassen:
@aoldershaw See my latest saga of a commit message, which should fix that test.

PR #5546, opened 2020-05-05
atc/db: less cryptic error message

taylorsilva:
# Why is this PR needed?

When a resource's parent type has no version you see this error on the resource's page:

```
parent type has no version
```

Noticed this error message myself and it wasn't clear to me what was
wrong. Someone on discord also mentioned that the error message was not
helpful with their troubleshooting.
Screen Shot 2020-05-05 at 11 12 07 AM
This error message should guide the user to look at the likely source of
the error: the parent resource type's config.

Change the error message when no parent type exists to:

```
resource's parent type 'name-of-parent-resource-type' has no version
```

# Contributor Checklist

- [ ] ~Unit tests~
- [x] Integration tests

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/check_factory.go
+++ b/atc/db/check_factory.go
@@ -142,7 +142,7 @@ func (c *checkFactory) TryCreateCheck(logger lager.Logger, checkable Checkable, 
 	parentType, found := resourceTypes.Parent(checkable)
 	if found {
 		if parentType.Version() == nil {
-			return nil, false, fmt.Errorf("resource's parent type '%s' has no version", parentType.Name())
+			return nil, false, fmt.Errorf("resource type '%s' has no version", parentType.Name())
 		}
 	}
 
--- a/atc/db/check_factory.go
+++ b/atc/db/check_factory.go
@@ -3,7 +3,6 @@ package db
 import (
 	"database/sql"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"time"
 
@@ -143,7 +142,7 @@ func (c *checkFactory) TryCreateCheck(logger lager.Logger, checkable Checkable, 
 	parentType, found := resourceTypes.Parent(checkable)
 	if found {
 		if parentType.Version() == nil {
-			return nil, false, errors.New("parent type has no version")
+			return nil, false, fmt.Errorf("resource's parent type '%s' has no version", parentType.Name())
 		}
 	}
 
--- a/testflight/custom_resource_check_test.go
+++ b/testflight/custom_resource_check_test.go
@@ -16,7 +16,7 @@ var _ = Describe("When a resource type depends on another resource type", func()
 		check := spawnFly("check-resource", "-r", inPipeline("recursive-custom-resource"), "--shallow")
 		<-check.Exited
 		Expect(check).To(gexec.Exit(1))
-		Expect(check.Err).To(gbytes.Say("parent type has no version"))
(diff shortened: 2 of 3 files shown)


atc/db/check_factory.go
```diff
@@ -143,7 +142,7 @@ func (c *checkFactory) TryCreateCheck(logger lager.Logger, checkable Checkable,
 	parentType, found := resourceTypes.Parent(checkable)
 	if found {
 		if parentType.Version() == nil {
-			return nil, false, errors.New("parent type has no version")
+			return nil, false, fmt.Errorf("resource's parent type '%s' has no version", parentType.Name())
```
xtremerui, line 145:
```suggestion
			return nil, false, fmt.Errorf("resource type '%s' has no version", parentType.Name())
```
Would it be more aligned to Concourse docs this way?

taylorsilva: I think it's important to have the word `parent` in the error message. It makes it clear what the relationship is between the resource page the user is currently on and the _other_ (parent) resource that is causing the error.

In general we have a tendency to write very robotic sounding error messages, probably because we're in the middle of writing code which usually reads very robotic. This PR is trying to make the error message more readable to the human that will eventually read it.

xtremerui: In our doc we don't say "resource's parent type" but just "a resource", "a resource type" or "a type for the resource". In our common sence "resource's parent type" = "resource type" while we are concourse dev but for just concourse users the "parent" might cause a confusion.

More specifically IMHO the error message should align with the first sentence of [Resource Type's page](https://concourse-ci.org/resource-types.html#resource-types) in our doc, where it says
> Each resource in a pipeline has a type. 

However I am not good at UX so not 100% sure others would take the message the same way. How about summoning @matthewpereira to get some inputs? Thx!

taylorsilva: That convinced me!

testflight/custom_resource_check_test.go
```diff
@@ -16,7 +16,7 @@ var _ = Describe("When a resource type depends on another resource type", func()
 		check := spawnFly("check-resource", "-r", inPipeline("recursive-custom-resource"), "--shallow")
 		<-check.Exited
 		Expect(check).To(gexec.Exit(1))
-		Expect(check.Err).To(gbytes.Say("parent type has no version"))
+		Expect(check.Err).To(gbytes.Say("resource's parent type '.*' has no version"))
```
xtremerui, line 19:
```suggestion
		Expect(check.Err).To(gbytes.Say("resource type '.*' has no version"))
```

testflight/resource_type_test.go
```diff
@@ -44,7 +44,7 @@ var _ = Describe("Configuring a resource type in a pipeline config", func() {
 			checkResource := spawnFly("check-resource", "-r", inPipeline("my-resource"), "--shallow")
 			<-checkResource.Exited
 			Expect(checkResource).To(gexec.Exit(1))
-			Expect(checkResource.Err).To(gbytes.Say("parent type has no version"))
+			Expect(checkResource.Err).To(gbytes.Say("resource's parent type '.*' has no version"))
```
xtremerui, line 47:
```suggestion
			Expect(checkResource.Err).To(gbytes.Say("resource type '.*' has no version"))
```

PR #5548, opened 2020-05-05
automatically fill system-claim-values

cirocosta:
ps.: this should be cherry-picked to `master`

# Why is this PR needed?

The `system-claim-key` and `system-claim-value` flags exist so you can tell concourse which requests should have 'system' privileges. These system privileges grant access to perform various actions required by the TSA, mainly around the lifecycle of workers.

By default we grant these 'system' privileges to any request presenting a token that was issued using the `concourse-worker` client (the token will have the `aud` claim set to `concourse-worker`). 

If a user wants to change the `tsa-client-id` then they probably won't know that they need to change the `system-claim-value` as well. So we automagically configure it for them behind the scenes.

### DIFF
--- a/cmd/concourse/web.go
+++ b/cmd/concourse/web.go
@@ -118,22 +118,42 @@ func (cmd *WebCommand) populateSharedFlags() error {
 	cmd.RunCommand.Auth.AuthFlags.Clients[cmd.TSACommand.ClientID] = cmd.TSACommand.ClientSecret
 	cmd.RunCommand.Auth.AuthFlags.Clients[cmd.RunCommand.Server.ClientID] = cmd.RunCommand.Server.ClientSecret
 
-	if cmd.TSACommand.ClientID != "" {
-		if cmd.RunCommand.SystemClaimKey == "aud" {
-			found := false
-			for _, val := range cmd.RunCommand.SystemClaimValues {
-				if val == cmd.TSACommand.ClientID {
-					found = true
-				}
+	// if we're using the 'aud' as the SystemClaimKey then we want to validate
+	// that the SystemClaimValues contains our TSA Client. If it's not 'aud' then
+	// we can't validate anything
+	if cmd.RunCommand.SystemClaimKey == "aud" {
+
+		// if we're using the default SystemClaimValues then override these values
+		// to make sure they include the TSA ClientID
+		if len(cmd.RunCommand.SystemClaimValues) == 1 {
+			if cmd.RunCommand.SystemClaimValues[0] == "concourse-worker" {
+				cmd.RunCommand.SystemClaimValues = []string{cmd.TSACommand.ClientID}
 			}
-			if !found {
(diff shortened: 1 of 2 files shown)


aoldershaw:
Hey @cirocosta, you marked this as ready for review but I'm still not sure what this is trying to solve. Could you fill in the PR description when you get the chance? Thanks!

Also, this isn't blocking 6.1, is it?

cirocosta:
aaaah I think we actually need it, is that right, @pivotal-jwinters ? would you mind updating the desc as well, please? I think you can provide better links to the places that this relates to than I could :sweat_smile: 

thanks!

pivotal-jwinters:
@aoldershaw I've updated the description

aoldershaw:
@pivotal-jwinters I'm a bit confused - the description says

> If a user wants to change the tsa-client-id ... we automagically configure it for them behind the scenes

But this doesn't seem to be what happens:

https://github.com/concourse/concourse/pull/5548/files#diff-843540a525b9215ba65ce08088526745R120-R129
```go
			Context("with tsa-client-id specified", func() {
				BeforeEach(func() {
					args = append(args, "--tsa-client-id", "tsa-client-id")
				})

				It("errors", func() {
					Eventually(concourseRunner.Err()).Should(
						gbytes.Say("at least one systemClaimValue must be equal to tsa-client-id"),
					)
				})
```

pivotal-jwinters:
@aoldershaw not sure what we did to be honest, but I've updated this PR with what should be the correct behaviour.

cmd/concourse/web.go
```diff
@@ -117,12 +118,46 @@ func (cmd *WebCommand) populateSharedFlags() error {
...
 
+	// if we're using the 'aud' as the SystemClaimKey then we want to validate
+	// that the SystemClaimValues contains our TSA Client. If it's not 'aud' then
+	// we can't validate anything
+	if cmd.RunCommand.SystemClaimKey == "aud" {
+
+		// if we're using the default SystemClaimValues then override these values
+		// to make sure they include the TSA ClientID
+		if len(cmd.RunCommand.SystemClaimValues) == 1 {
+			if cmd.RunCommand.SystemClaimValues[0] == "concourse-worker" {
```
aoldershaw, line 129: It doesn't really make sense to do, but in theory, an operator *could* pass in:

```
--tsa-client-id something-else --system-claim-value concourse-worker
```

and it would replace `concourse-worker` with `something-else`. From what I can tell, doing this would basically just be trying to shoot yourself in the foot, so this is probably something we don't care about - just mentioning it anyway.

cirocosta:
@pivotal-jwinters are we cherry-picking this to master?

pivotal-jamie-klassen:
@cirocosta i leave the decision to y'all, but I think that merging `release/6.1.x` back to master once 6.1.0 ships is a viable option as well. We now know that we're planning to create a `support/6.2.x` branch for 6.2 LTS release.

Issue #5456, opened 2020-04-17
Pinning a resource to `version: latest` in the resource definition does not work


andy-paine:
## Summary

Pinning a resource to `version: latest` in the resource definition does not work, only in the `get` step

## Steps to reproduce

Set the following pipeline:
```
resources:
- name: timer
  type: time
  version: latest
  source:
    interval: 30s
jobs:
- name: test
  plan:
  - get: timer
    trigger: true
```

## Expected results
Pipeline sets okay and the latest version of the `timer` resource is used when the `test` job is run

## Actual results
Pipeline config fails validation as `version: latest` is not a valid `map[string]string`

## Additional context
The docs for the resource schema [here](https://concourse-ci.org/resources.html#schema.resource.version) state that:
> This has the same effect as setting [ `get` step **`version`** ](https://concourse-ci.org/jobs.html#schema.step.get-step.version) on every [ `get` step](https://concourse-ci.org/jobs.html#get-step) referencing the resource.

However using `version: latest` only works in the `get:` step and doesn't work in the resource definition.

Looking at the code it seems that the config for the resource uses the `Version` struct
https://github.com/concourse/concourse/blob/5e315c6ccfe2fdb087ecb40a7a5339b5dc31feb5/atc/config.go#L184
and the config for the get step uses the `VersionConfig` struct which supports the `every` and `latest` alternatives.
https://github.com/concourse/concourse/blob/5e315c6ccfe2fdb087ecb40a7a5339b5dc31feb5/atc/config.go#L486

I assume this _isn't_ intended behaviour and the two should be equivalent? Or is it just a docs problem and only specific pinning (e.g. `version: { ref: blah }`) is supported in resource definitions?

pivotal-jamie-klassen:
I guess this is a docs problem - as I understand it, the purpose of the `version:` field on the resource config is strictly for pinning. Indeed it says

> A version to pin the resource to across the pipeline.

in the same spot you linked. Maybe it's worth explicitly saying that that it's equivalent to setting the `get` step version across the pipeline, *except* the `version: every` and `version: latest` options are not valid here.

Stepping back, though, why were you wanting to set `version: latest` at the resource level? What were you expecting to happen otherwise?

DanielJonesEB:
@andy-paine and I have been working on this.

> Stepping back, though, why were you wanting to set version: latest at the resource level? What were you expecting to happen otherwise?

[...]

PR #5550, opened 2020-05-05
atc: structure: Make algorithm.Compute take only InputConfigs

andy-paine:
# Why is this PR needed?

There is a lot of logic going on inside the Algorithm.Compute method which makes it hard to decompose and test small things such as adding new `version` conditions (as part of fixing #5456).

Make it easier to understand the precedence of `version` flags and pinning between resources and job inputs.

Extract the logic around which is the correct pinned version and whether to be using every version into a separate, testable method. Add a set of table tests for the different set of inputs so we can easily and readably see what the expected outcomes of multiple configuration options are.

There is _one_ functional change in this PR which is that when a pinned version is set on an `InputConfig`, the `UseEveryVersion` flag is no longer explicitly set. This shouldn't be an issue as pinned versions take precedence over `version: every` and we now have tests if we ever want to change this behaviour.

# Notes
* This is my first foray into this codebase and I don't write a lot of Golang (that might be obvious) so feedback is more than welcome.
* Not a massive fan of table testing usually but seemed right given this is basically a matrix multiplication of possible configurations.
* Should this be in the algorithm package? Should it just be moved up into the scheduler package? The way `Scheduler` and `BuildStarter` interact means there's some duplicate code and so I stuck it in the `Algorithm` class since they both already had the same instance of that.

# Contributor Checklist

- [x] Unit tests
- [ ] ~Integration tests~ Hoping for a no-op to user-facing behaviour
- [ ] ~Updated documentation (located at https://github.com/concourse/docs)~ Not relevant, internal only
- [ ] ~Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)~ Not relevant, internal only

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/scheduler/algorithm/compute.go
+++ b/atc/scheduler/algorithm/compute.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/tracing"
 )
@@ -27,9 +26,7 @@ type Algorithm struct {
 func (a *Algorithm) Compute(
 	ctx context.Context,
 	job db.Job,
-	inputs []atc.JobInput,
-	resources db.Resources,
-	relatedJobs NameToIDMap,
+	inputConfigs InputConfigs,
 ) (db.InputMapping, bool, bool, error) {
 	ctx, span := tracing.StartSpan(ctx, "Algorithm.Compute", tracing.Attrs{
 		"pipeline": job.PipelineName(),
@@ -37,7 +34,7 @@ func (a *Algorithm) Compute(
 	})
 	defer span.End()
 
-	resolvers, err := constructResolvers(a.versionsDB, job, inputs, resources, relatedJobs)
+	resolvers, err := constructResolvers(a.versionsDB, inputConfigs)
 	if err != nil {
 		return nil, false, false, fmt.Errorf("construct resolvers: %w", err)
 	}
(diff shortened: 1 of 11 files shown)


DanielJonesEB:
Hey folks! Just an FYI that this is a preliminary commit on a [WIP PR](https://github.com/EngineerBetter/concourse/tree/version-every-on-resources) that will address https://github.com/concourse/concourse/issues/5456 by allowing resources to take `latest` as a version (and also then `every` for consistency). Once this is merged, we'll crack on with the next, more interesting, PR.

clarafu:
@andy-paine Thanks for the PR! I definitely agree think that having `algorithm.Compute` only take `InputConfigs` is reasonable and lessens the complexity of the algorithm. The funny and unfortunate thing is that we actually have a PR #5420 that refactors that part of the code 😓 and moved the logic for determining the input config into `job.AlgorithmInputs` 

https://github.com/concourse/concourse/blob/738371530925aaf48c0b03ad9cf780ca9139a702/atc/db/job.go#L248-L286

Im thinking that if you would still like to add the test cases, you could add them to the `job_test.go` suite that covers the `job.AlgorithmInputs` method 

https://github.com/concourse/concourse/blob/738371530925aaf48c0b03ad9cf780ca9139a702/atc/db/job_test.go#L2109

It might be harder to use a table testing setup for these db tests because they involve a database and prior setup for running each test.

andy-paine:
Closing this in favour of #5420 which supersedes a lot of this work

PR #5568, opened 2020-05-07
k8s/topgun: update prometheus dependencies and disable worker

cirocosta:
### Why is this PR needed?

With `stable/prometheus` going from having `kube-state-metrics` embedded in its set of templates to a separate dependency, `k8s-topgun` started failing:
```
Error: found in requirements.yaml, but missing in charts/ directory: kube-state-metrics
```

Without this PR, we can't have `k8s-topgun` running without having the version of `helm/charts` pinned to an old commit, which would force us to have that commit hardcoded in the release branches, as well as the main pipeline.

### What is this PR trying to accomplish?

It tries to make the installation of `prometheus` possible regardless of how its dependencies are set up (either through `requirements.yaml` - the new way of doing thing -, or the old one - embedded). 

### How does it accomplish that?

It does so by running `helm dependency update` before running `helm install`.

Given that such command will not fail in case of `requirements.yaml` not existing (which would be true for older versions of the `helm/charts` repo), it's safe to always do it.

### Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/topgun/k8s/prometheus_test.go
+++ b/topgun/k8s/prometheus_test.go
@@ -6,6 +6,7 @@ import (
 	"path"
 	"time"
 
+	. "github.com/concourse/concourse/topgun"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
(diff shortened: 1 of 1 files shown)


cirocosta:
I just realized this is very related to https://github.com/concourse/ci/issues/285.

IMO both coexist: I don't think that because this solves the point of having to `pin` that the other issue would be invalidated - it'd still be great to keep track of the versions to establish the matrix.

PR #5569, opened 2020-05-08
atc: behavior: consolidate component runners, minor cleanups

vito:
# Why is this PR needed?

We have four nearly word-for-word identical implementations of the same code for running various components on an interval. 

Consolidate these implementations and use it as a scenario for exploring `testify` usage for non-trivial code.

This PR consolidates them into a `component/` package, tested through `testify`, and also converts the `builds/` package over to `testify` in preparation for #5504 which migrates the build planner interface over to it.

Along the way, it also fixes two oddities:

1. Var source collection need not be a component.
1. Static worker registration code got in the way so I removed it; it should be dead code as it existed purely for a dev flow many years ago which we haven't used since.

# Contributor Checklist

- [ ] Unit tests
- [ ] Integration tests

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


atc/scheduler/runner.go
```diff
@@ -34,7 +35,7 @@ type schedulerRunner struct {
 	running            *sync.Map
 }
 
-func NewRunner(logger lager.Logger, jobFactory db.JobFactory, scheduler BuildScheduler, maxJobs uint64) Runner {
+func NewRunner(logger lager.Logger, jobFactory db.JobFactory, scheduler BuildScheduler, maxJobs uint64) component.Runnable {
```
pivotal-jwinters, line 38: If this returned `*schedulerRunner`, then we wouldn't need to depend on the component package here.

atc/component/scheduler.go
```diff
@@ -0,0 +1,76 @@
...
+// Schedulable represents a workload that is executed normally on a periodic
+// schedule, but can also be run immediately.
+type Schedulable interface {
+	RunPeriodically(context.Context)
+	RunImmediately(context.Context)
+}
+
+// Scheduler runs a workload periodically, or immediately upon receiving a
+// notification.
+type Scheduler struct {
```
pivotal-jwinters, line 29: Calling this a `Scheduler` is a bit confusing given the importance of the actual scheduler. Maybe just simply `Runner`?

atc/component/scheduler.go
```diff
@@ -0,0 +1,76 @@
...
+	"context"
+	"os"
+	"time"
+
+	"code.cloudfoundry.org/clock"
+	"code.cloudfoundry.org/lager"
+	"code.cloudfoundry.org/lager/lagerctx"
+)
+
+var Clock = clock.NewClock()
```
pivotal-jwinters, line 13: Is there a reason to make this a global variable? It could make testing more difficult.

atc/builds/tracker_test.go
```diff
@@ -1,123 +1,107 @@
...
 
-var _ = Describe("Tracker", func() {
-	var (
-		fakeBuildFactory *dbfakes.FakeBuildFactory
-		fakeEngine       *enginefakes.FakeEngine
-
-		tracker *builds.Tracker
-		logger  *lagertest.TestLogger
-	)
+type TrackerSuite struct {
```
pivotal-jwinters, line 18: Should this be part of this PR? Isn't this part of the other refactor?

atc/atccmd/command.go
```diff
@@ -965,25 +1004,21 @@ func (cmd *RunCommand) constructBackendMembers(
...
-			logger.Session(atc.ComponentBuildReaper),
-			gc.NewBuildLogCollector(
+		},
+		{
+			Component: atc.Component{
+				Name:     atc.ComponentBuildTracker,
+				Interval: cmd.BuildTrackerInterval,
+			},
+			Runnable: builds.NewTracker(dbBuildFactory, engine),
+			Drainer:  engine,
```
pivotal-jwinters, line 1014: I wonder if it makes sense to compose the Drainer into the Runnable interface. It would be nice for anyone implementing a component to just have to implement the Drain method alongside Run, and not have to know to do this. I'm thinking something along the lines of a RunDrainer (similar to a ReadCloser).

atc/atccmd/command.go
```diff
@@ -560,17 +563,58 @@ func (cmd *RunCommand) constructMembers(
...
-	return append(apiMembers, append(backendMembers, gcMembers...)...), nil
+	// technically these could come from any conn; the choice of apiConn is
+	// arbitrary
+	bus := apiConn.Bus()
+	componentFactory := db.NewComponentFactory(apiConn)
+
+	members := apiMembers
+	components := append(backendComponents, gcComponents...)
+	for _, c := range components {
+		dbComponent, err := componentFactory.CreateOrUpdate(c.Component)
```
pivotal-jwinters, line 584: Ah this makes way more sense

atc/component/scheduler.go
vito: It's actually exposed so that it can be reassigned in the tests. 🙂 I don't feel too strongly about this but I've seen that pattern in other projects and it feels like a decent way to keep objects from ending up "over-injected" with things that we'd never want to change IRL.

atc/scheduler/runner.go
vito: 👍  I can make the type public and just return it directly.

pivotal-jwinters: You don't even have to make it public, private structs still conform to all the necessary interfaces and it stops other packages from using those types.

atc/component/scheduler.go
vito: Yeah agreed, this is one of those wishy-washy names and we already use it elsewhere. I guess `Runner` gels well with `Runnable`. I can try that out!

atc/builds/tracker_test.go
vito: I put it here so that at the end the `atc/builds/` package can be fully one test framework, which is accomplished by the removal of the tracker runner, which is part of this PR.

atc/component/scheduler.go
pivotal-jwinters: Do the testify tests not run in parallel?

I feel like "over-injected" methods is a smell that we shouldn't be hiding away.

atc/atccmd/command.go
vito: I actually had it that way at one point, but we don't have any other cases where a component needs to drain, so it didn't feel worth making all of the interfaces (including `Coordinator` and `Scheduler`) more complicated.

Felt like putting it up here was better so we don't have 3/4 components with no-op `Drain` methods, and this particular functionality is already covered at the integration level anyway. More generally I would prefer to avoid relying on explicit draining too much since you never know when the ATC will get `kill -9`ed, and it often leads to annoying slowness when shutting down the `web` nodes.

atc/scheduler/runner.go
vito: Yeah, I just find that pattern kind of awkward. I've never really seen that in idiomatic Go codebases. 😕 Would be nice to just export the type and maybe even give it Godoc treatment. In cases like this I'd place more value on idiomatic than defensive code; we can catch that kind of stuff in PR review, and it seems unlikely to occur anyway.

atc/component/scheduler.go
vito: I definitely don't plan to do that for *all* injection; it's mainly that telling an object to use a "real clock" in production code when it's clearly just for testing feels a little goofy and adds noise to the real live code which we'll be reading forever.

`go test` runs tests in series unless you explicitly call `t.Parallel()` in the tests you want to run in parallel. (Side note: this looks like a good thing to watch to suss out any order dependence - https://github.com/golang/go/issues/28592)

atc/scheduler/runner.go
pivotal-jwinters: Why is it awkward? The whole point of having a constructor in the first place is because you want to control the creation of those objects.

atc/atccmd/command.go
pivotal-jwinters: Each component wouldn't need to have a no-op implementation of Drain. Instead we could wrap each in a NoopRunDrainer whenever it needs to fulfill that interface.

atc/scheduler/runner.go
vito: It's a pattern that seems to assume the codebase is a hostile environment with consumers of our APIs that don't know what they're doing. These seem like rare situations to catch in code review and tests.

We *could* abuse all kinds types exported by `stdlib` and third party packages by avoiding the constructor, but we don't. I've never really seen this pattern applied anywhere that we use.

It's awkward because it forces the caller to come up with an interface but only if they try to use the object in a certain way. Returning a private type means I can use the object so long as I do `foo := ...` and use `foo` within one function, but the moment I have to pass it as an argument somewhere (e.g. a private method) or if I use the `var foo` syntax, I can't be cause I can't refer to the type. So then the I have to explicitly declare an interface. Sure, it's nice having an explicit interface for the contract, but it really doesn't buy you anything if you weren't intending to have an interface already. The compiler will let you know if the interface breaks your expectations either way.

Returning a private type would also be a problem for the tests since they don't share the private namespace. They could just use the `component.Runnable` interface or declare their own, but then we've just moved the dependency from the package to the package's tests, or we've added an interface that we just have to keep changing to match the "real" interface.

To me, the upside of being more idiomatic code that can be documented with `godoc` outweighs the risk mitigated by having full control over the construction.

atc/atccmd/command.go
vito: Ahhh I see, you're not saying to add `Drain` to the `Runnable` interface, you're saying we can add another interface which components can conform to and if they do we'll do the draining automatically. Yeah I'll play with that!

atc/scheduler/runner.go
pivotal-jwinters: > It's awkward because it forces the caller to come up with an interface but only if they try to use the object in a certain way. Returning a private type means I can use the object so long as I do foo := ... and use foo within one function, but the moment I have to pass it as an argument somewhere (e.g. a private method) or if I use the var foo syntax, I can't because I can't refer to the type. So then the I have to explicitly declare an interface. Sure, it's nice having an explicit interface for the contract, but it really doesn't buy you anything if you weren't intending to have an interface already.

You say this as if it's a bad thing lol. Forcing people to be explicit about what they accept by defining an interface is good design. It's also probably the most idiomatic thing about go.

pivotal-jwinters: Honestly I don't really care about this specific case. I just want to make sure that we're being consistent with the rest of the codebase. And if we're not going to be consistent we need a good reason.

vito: I would prefer to be consistent with practices that are common in the Go community so that we can enforce them through tooling like `golint`, which now that I check it complains about this exact scenario:

```
atc/scheduler/runner.go:37:105: exported func NewRunner returns unexported type *scheduler.schedulerRunner, which can be annoying to use
```

Interfaces are great, but forcing someone to write an interface just for the sake of having one isn't intrinsically valuable. It can end up just being a maintenance burden and a mild headache as you can never 'go to definition' in your editor.

atc/component/scheduler_test.go
```diff
@@ -0,0 +1,158 @@
...
+	})
+
+	defer func() {
+		process.Signal(os.Interrupt)
+		<-process.Wait()
+	}()
+
+	s.Run("runs periodically on component interval", func() {
+		s.clock.WaitForWatcherAndIncrement(interval)
+		<-ranPeriodically
```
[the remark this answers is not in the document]
vito: Yeah, these test situations are kind of annoying since both approaches have downsides.

I stopped using `Eventually` even in Ginkgo/Gomega because it ended up causing way too many CI flakes that were just caused by the worker being under more load than usual.

Nowadays I prefer to accept the risk of the test hanging indefinitely - usually those failures don't escape developer machines, and if they make it to CI they tend to motivate people to fix the tests. 😄 I think that's the better tradeoff since it prevents us from becoming numb to flakiness.

vito:
@aoldershaw @clarafu Thanks for the reviews! Just rebased on master and addressed the comments. Noticed along the way that I deleted the scheduler runner tests, which was a mistake - those are back now too. (I only meant to delete the *interval* runner tests.)

Issue #5547, opened 2020-05-05
Take a look at `create-and-run-new-pipeline` in `monitoring-stats` to see if it's genuinely slower with v6.0


taylorsilva:
The [`create-and-run-new-pipeline` ](https://ci.concourse-ci.org/teams/main/pipelines/monitoring-stats/jobs/create-and-run-new-pipeline/builds/12942) SLI being run against our large scale deployment `hush-house` have been consistently failing. Investigate what is causing the SLI to fail.

As general knowledge of the SLI, it involves a few different steps for it to pass successfully:
1. Set the pipeline
1. Check the resource and save the new version
1. Auto-trigger a new build using that version
1. Run the task successfully

All these steps should finish within 70 seconds because of a timeout configured for the monitoring job.

https://github.com/concourse/oxygen-mask/blob/3ba002998bd924c6fa1818b45429c53110b9f425/monitor.yml#L227-L229

pivotal-jamie-klassen:
While broadly considering how to investigate and generate statistical hypotheses around this problem, we ended up talking a lot about tracing. This issue has now been kinda hijacked by an effort to add tracing to the `lidar` package -- maybe it should be renamed appropriately -- and the WIP commit above is a really good start.

The key insight was to add a `span_context` field to the `plan` JSON on a `db.Check`. This allows us to propagate span context between the `Scanner` and the `Checker` using the `propagators` package in `opentelemetry-go`. I learned how to use an approach like this in my work on #5461. This time, instead of using the `ChildOf` relationship type between spans, we have learned to use the `FollowsFrom` relationship type.

Initially we thought there was some bug because we only saw a single span in jaeger. We couldn't see any clear signals that anything was wrong, though. Today we decided to try checking the logs from the jaeger collector, and saw that elasticsearch was reporting some 403s, and continuing the investigation we saw that elasticsearch itself was rejecting write operations because our Docker VM was almost out of disk.

What fun it is to shave a yak! But more seriously, it might be worth having concourse do some logging if the `SpanSyncer` underlying the tracing package starts failing.

After some judicious docker pruning, the results in jaeger are exactly what we hoped!

<img width="995" alt="Screen Shot 2020-05-06 at 12 36 15 PM" src="https://user-images.githubusercontent.com/22056320/81204223-f23a6100-8f96-11ea-826d-fa5be878bf3a.png">

[...]

PR #5575, opened 2020-05-11
Lidar tracing

jamieklassen:
# Why is this PR needed?

There are a few recurring scenarios where concourse seems to "hang" and not give clear feedback about why. The work in this PR is focused around improving observability in the "I pushed a new commit, but my test job isn't triggering" workflow, which has sizeable overlap with the `create-and-run-new-pipeline` SLI.

The end goal is to enable a single execution of the `create-and-run-new-pipeline` SLI to correspond to a single trace - failing that, it would be nice to have a collection of traces that observe the flow of execution between all the exercised components: `scanner -> checker -> scheduler -> buildstarter -> exec` to enable a (stochastic) answer to the question "why is this SLI timing out?"

related to #5547.

The straightforward part is just emitting traces from the previously-less-observed components in the `lidar` package. The more interesting and subtle part is the use of span context propagation to create relationships between spans. Now the span context created by the scanner gets serialized in a `span_context` column on the `check` it creates - this context is picked up by the checker, and any resource versions it outputs are annotated with a `span_context` in a similar manner.

It still remains to create the link between the checker and scheduler/buildstarter/exec. This is a bit more subtle because there can potentially be many resource versions that "cause" a build to be scheduled. It will probably make more sense to use a `Link` than `Relationship` in opentelemetry for this purpose.

# Contributor Checklist

- [ ] Unit tests
- [ ] Integration tests

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [x] New config flags added? Ensure that they are added to the [BOSH](https://github.com/concourse/concourse-bosh-release) and [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for the [integration tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env) (for example, if they are Garden configs that are not displayed in the `--help` text).~

### DIFF
--- a/atc/db/check.go
+++ b/atc/db/check.go
@@ -11,7 +11,6 @@ import (
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db/lock"
 	"github.com/lib/pq"
-	"go.opentelemetry.io/otel/api/propagators"
 )
 
 type CheckStatus string
@@ -50,7 +49,7 @@ type Check interface {
 	AcquireTrackingLock(lager.Logger) (lock.Lock, bool, error)
 	Reload() (bool, error)
 
-	SpanContext() propagators.Supplier
+	SpanContext() SpanContext
 }
 
 var checksQuery = psql.Select(
@@ -327,7 +326,7 @@ func (c *check) SaveVersions(versions []atc.Version) error {
 	return saveVersions(c.conn, c.resourceConfigScopeID, versions)
 }
 
-func (c *check) SpanContext() propagators.Supplier {
+func (c *check) SpanContext() SpanContext {
 	return c.spanContext
 }
 
--- a/atc/db/check.go
+++ b/atc/db/check.go
@@ -11,6 +11,7 @@ import (
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db/lock"
 	"github.com/lib/pq"
+	"go.opentelemetry.io/otel/api/propagators"
 )
 
 type CheckStatus string
@@ -49,7 +50,7 @@ type Check interface {
 	AcquireTrackingLock(lager.Logger) (lock.Lock, bool, error)
 	Reload() (bool, error)
 
-	SpanContext() SpanContext
+	SpanContext() propagators.Supplier
 }
 
 var checksQuery = psql.Select(
@@ -326,7 +327,7 @@ func (c *check) SaveVersions(versions []atc.Version) error {
 	return saveVersions(c.conn, c.resourceConfigScopeID, versions)
 }
 
-func (c *check) SpanContext() SpanContext {
+func (c *check) SpanContext() propagators.Supplier {
 	return c.spanContext
 }
 
--- a/atc/db/check.go
+++ b/atc/db/check.go
@@ -11,6 +11,7 @@ import (
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db/lock"
 	"github.com/lib/pq"
+	"go.opentelemetry.io/otel/api/propagators"
 )
 
 type CheckStatus string
@@ -48,6 +49,8 @@ type Check interface {
 	AllCheckables() ([]Checkable, error)
 	AcquireTrackingLock(lager.Logger) (lock.Lock, bool, error)
 	Reload() (bool, error)
+
+	SpanContext() propagators.Supplier
 }
 
 var checksQuery = psql.Select(
@@ -62,6 +65,7 @@ var checksQuery = psql.Select(
 	"c.nonce",
 	"c.check_error",
 	"c.metadata",
+	"c.span_context",
 ).
 	From("checks c")
 
@@ -80,6 +84,8 @@ type check struct {
 	createTime time.Time
 	startTime  time.Time
 	endTime    time.Time
+
+	spanContext SpanContext
 }
 
 type CheckMetadata struct {
@@ -321,12 +327,16 @@ func (c *check) SaveVersions(versions []atc.Version) error {
 	return saveVersions(c.conn, c.resourceConfigScopeID, versions)
 }
 
+func (c *check) SpanContext() propagators.Supplier {
+	return c.spanContext
+}
+
 func scanCheck(c *check, row scannable) error {
 	var (
 		createTime, startTime, endTime  pq.NullTime
 		schema, plan, nonce, checkError sql.NullString
 		status                          string
-		metadata                        sql.NullString
+		metadata, spanContext           sql.NullString
 	)
 
 	err := row.Scan(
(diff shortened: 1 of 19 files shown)


atc/db/check.go
```diff
@@ -11,6 +12,7 @@ import (
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db/lock"
 	"github.com/lib/pq"
+	"go.opentelemetry.io/otel/api/propagators"
```
pivotal-jwinters, line 15: We probably don't want to use opentelemetry directly? It looks like the `tracing` package was designed to wrap this dependency.

pivotal-jwinters: Just doing a quick search it actually looks like the scheduler package uses it directly, so maybe we're not being strict about this anymore.

pivotal-jamie-klassen: yeah i hoped to make a clean adapter for that, just haven't gotten around to it yet.

pivotal-jamie-klassen:
This PR has a roughly-complete implementation for tracing the scanner and checker, linking their span contexts appropriately, but the changes involving the `span_context` column on the `resource_config_versions` table are still incomplete. @clarafu will take over this PR, narrowing its scope to the lidar package as named. The idea is to add at least this much telemetry to hush-house to partially instrument #5547 before deploying #5580 to confirm whether it fixes the root cause of our SLI degradations.

The scheduler/buildstarter piece, which has turned out to be a bit subtler, has been extracted to #5602. I simply added it to the roadmap rather than the [post-v6.0 feedback project](https://github.com/concourse/concourse/projects/46), following the precedent set by https://github.com/vito/cadet/commit/82ebaff63cf9ccd573d8c570cd881f9fd315b9ae -- @vito @scottietremendous hopefully I did this right.

clarafu:
I edited and cleaned up this PR to only include the tracing between lidar scanner and checker since we are going to split out adding tracing between saving a version in the checker to a build scheduling using those versions in a separate PR for #5602.

I also created a new branch called `tracing-wip` that holds the work we did for adding the `span_context` to resource versions.

atc/plan.go
```diff
@@ -111,16 +111,25 @@ type PutPlan struct {
...
-	Tags        Tags    `json:"tags,omitempty"`
-	Timeout     string  `json:"timeout,omitempty"`
-	FromVersion Version `json:"from_version,omitempty"`
+	Type        string            `json:"type"`
+	Name        string            `json:"name,omitempty"`
+	Source      Source            `json:"source"`
+	Tags        Tags              `json:"tags,omitempty"`
+	Timeout     string            `json:"timeout,omitempty"`
+	FromVersion Version           `json:"from_version,omitempty"`
+	SpanContext map[string]string `json:"span_context"`
```
pivotal-jwinters, line 120: Do we need this here if we're storing the span_context in its own column?

atc/db/check.go
pivotal-jwinters: @clarafu I think we can just change `propagators.Supplier` to `SpanContext` to get rid of this dependency?

atc/plan.go
clarafu: Ah yes you're right we can take that out now!

atc/db/check.go
clarafu: @pivotal-jwinters I think we need to keep it as `propagators.Supplier` because it needs to match the interface for WithSpanContext https://github.com/concourse/concourse/blob/b38487edf2d69b1fe8d75ed19b4eba566fd7ed46/tracing/tracer.go#L90-L92

as we pass call `StartSpanFollowing` in lidar checker https://github.com/concourse/concourse/blob/b38487edf2d69b1fe8d75ed19b4eba566fd7ed46/atc/lidar/checker.go#L52-L60
using that span context.

pivotal-jwinters: cool 👍

cirocosta:
it seems like we're consistently getting trouble with `-race` :thinking:  

```
==================

WARNING: DATA RACE

Read at 0x00c00012cfe0 by goroutine 56:

  github.com/concourse/concourse/atc/lidar.(*checker).Run.func1()

      /tmp/build/0df9eea0/concourse/atc/lidar/checker.go:56 +0xac


Previous write at 0x00c00012cfe0 by goroutine 21:

  github.com/concourse/concourse/atc/lidar.(*checker).Run()

      /tmp/build/0df9eea0/concourse/atc/lidar/checker.go:49 +0x22e

  github.com/concourse/concourse/atc/lidar_test.glob..func1.2()

      /tmp/build/0df9eea0/concourse/atc/lidar/checker_test.go:47 +0xa0

  ...


Goroutine 56 (running) created at:

  github.com/concourse/concourse/atc/lidar.(*checker).Run()

      /tmp/build/0df9eea0/concourse/atc/lidar/checker.go:51 +0x390

  github.com/concourse/concourse/atc/lidar_test.glob..func1.2()

      /tmp/build/0df9eea0/concourse/atc/lidar/checker_test.go:47 +0xa0

  github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync()

      /tmp/build/0df9eea0/gopath/pkg/mod/github.com/onsi/ginkgo@v1.10.3/internal/leafnodes/runner.go:113 +0xf3


```

(see https://ci.concourse-ci.org/teams/contributor/pipelines/prs/jobs/unit/builds/39#L5eba8d83:152:188)

atc/lidar/checker.go
```diff
@@ -47,6 +49,17 @@ func (c *checker) Run(ctx context.Context) error {
 	for _, ck := range checks {
 		if _, exists := c.running.LoadOrStore(ck.ID(), true); !exists {
 			go func(check db.Check) {
+				_, span := tracing.StartSpanFollowing(
```
cirocosta, line 52: it got me wondering if we should have a span right there in `exec.CheckStep.Run` :thinking:  

that way, we could follow the same pattern that we have for all of the other steps, and also be able to capture errors (something we wouldn't be able to do so just from here)

https://github.com/concourse/concourse/blob/0b79feeb5ab7fa205c60ae6d34334152390cc74c/atc/exec/task_step.go#L125-L138

jamieklassen: i think this is how we did it on the first try, but found that the exec part of the trace pretty much overlapped completely with the lidar part..

clarafu: I see, that's true that this won't capture any errors within the check step. I think we could add it back but maybe I can split that out into a different issue and have a separate PR for it?

clarafu:
@cirocosta Thanks for the review!! I addressed your comments, would you be able to review it one more time? Thank youuu :D

PR #5579, opened 2020-05-11
Tracing for Image Streaming

### LINKED ISSUE #5572: Add tracing to volume streaming
## Summary

With tracing we have the awesome ability to visualize complex flows and thus it would be very useful to apply tracing to volume streaming.

#### For Concourse developers
- it would make it very easy to observe under test scenarios where volumes are being streamed/not-streamed
  * 3 different ways of specifying task's image source

#### For Concourse users
- ability to observe their workloads in much more detail to be able to make more intelligent decisions with regards to the container placement strategy
- observe that task caches are never streamed
- add ability to account for total streaming of bytes
- ability to identify potential bottlenecks with regards to where the workers are and being able to optimize using tags or moving workers closer 

## Next Steps
- How might we observe the entire cache & then subsequent volume life cycle ?

* RFC: concourse/rfcs#
* Prior discussion: #
* Depends on #
* Part of #

xtreme-sameer-vohra:
Fixes #5572 

# Why is this PR needed? 

See #5572 

# Contributor Checklist

# Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

# Instruction for Acceptance Test
- Update the `docker-compose.yml` with two diff. workers to use diff. tags (e.g we used `workerb` in the following example)
- Run `docker-compose -f docker-compose.yml -f hack/overrides/jaeger.yml up --build -d` 
- You can use the following `pipeline.yml` for testing
```pipeline.yml
---
resources:
  - name: my-image
    type: registry-image
    source:
      repository: busybox

jobs:
  - name: create-and-consume
    public: true
    plan:
      - get: my-image
      - task: make-a-file
        tags: [workerb]
        image: my-image
        config:
          platform: linux
          run:
            path: sh
            args:
              - -exc
              - ls -la; echo "Created a file on $(date)" > ./files/created_file
          outputs:
            - name: files
      - task: consume-the-file
        config:
          platform: linux
          image_resource:
            type: registry-image
          inputs:
            - name: files
          run:
            path: cat
            args:
              - ./files/created_file
```
- Check jaeger UI on port `25000`

### DIFF
--- a/atc/worker/artifact_source.go
+++ b/atc/worker/artifact_source.go
@@ -8,6 +8,7 @@ import (
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc/compression"
 	"github.com/concourse/concourse/atc/runtime"
+	"github.com/concourse/concourse/tracing"
 	"github.com/hashicorp/go-multierror"
 )
 
@@ -60,6 +61,14 @@ func (source *artifactSource) StreamTo(
 	logger lager.Logger,
 	destination ArtifactDestination,
 ) error {
+	ctx, span := tracing.StartSpan(ctx, "artifactSource.StreamTo", nil)
+	defer span.End()
+
+	_, outSpan := tracing.StartSpan(ctx, "volume.StreamOut", tracing.Attrs{
+		"origin-volume": source.volume.Handle(),
+		"origin-worker": source.volume.WorkerName(),
+	})
+	defer outSpan.End()
 	out, err := source.volume.StreamOut(ctx, ".", source.compression.Encoding())
 	if err != nil {
 		return err
--- a/atc/worker/artifact_source.go
+++ b/atc/worker/artifact_source.go
@@ -70,17 +70,17 @@ func (source *artifactSource) StreamTo(
 	})
 	defer outSpan.End()
 	out, err := source.volume.StreamOut(ctx, ".", source.compression.Encoding())
+
 	if err != nil {
+		tracing.End(outSpan, err)
 		return err
 	}
 
 	defer out.Close()
 
 	err = destination.StreamIn(ctx, ".", source.compression.Encoding(), out)
-	if err != nil {
-		return err
-	}
-	return nil
+
+	return err
 }
 
 // TODO: figure out if we want logging before and after streams, I remove logger from private methods
--- a/atc/worker/worker.go
+++ b/atc/worker/worker.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/concourse/concourse/tracing"
 	"path/filepath"
 	"sort"
 	"strings"
@@ -23,6 +22,7 @@ import (
 	"github.com/concourse/concourse/atc/resource"
 	"github.com/concourse/concourse/atc/runtime"
 	"github.com/concourse/concourse/atc/worker/gclient"
+	"github.com/concourse/concourse/tracing"
 	"github.com/cppforlife/go-semi-semantic/version"
 	"golang.org/x/sync/errgroup"
 )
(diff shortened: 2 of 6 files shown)


atc/worker/artifact_source.go
```diff
@@ -60,6 +61,14 @@ func (source *artifactSource) StreamTo(
...
 	destination ArtifactDestination,
 ) error {
+	ctx, span := tracing.StartSpan(ctx, "artifactSource.StreamTo", nil)
+	defer span.End()
+
+	_, outSpan := tracing.StartSpan(ctx, "volume.StreamOut", tracing.Attrs{
+		"origin-volume": source.volume.Handle(),
+		"origin-worker": source.volume.WorkerName(),
+	})
+	defer outSpan.End()
```
cirocosta, line 71: oh, reminder: it'd be good to capture the error so that one can see from the traces when failures occur :grin:

scottietremendous:
@cirocosta @xtreme-sameer-vohra  - Is this issue still a WIP?

atc/worker/worker.go
```diff
@@ -8,6 +8,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/concourse/concourse/tracing"
```
cirocosta, line 11: whops, this should be in the other set of imports below \/

cirocosta:
@scottietremendous I'd say yes, but like, 99% done tbh (I don't see anything else other than addressing the comments above, but @xtreme-sameer-vohra might have a different view of it :grin: )

atc/worker/worker.go
```diff
@@ -491,16 +492,19 @@ func (worker *gardenWorker) createVolumes(
...
-		logger,
-		spec.TeamID,
-		isPrivileged,
-		creatingContainer,
-		nonlocalInputs,
-	)
-	if err != nil {
-		return nil, err
+	streamedMounts := []VolumeMount{}
+	if len(nonlocalInputs) > 0 {
```
clarafu, line 496: @cirocosta Is this adding new logic aside from adding tracing to image streaming? I'm not sure what effect only cloning remote volumes when there are non local inputs has?

atc/worker/artifact_source.go
clarafu: @cirocosta Hmm how exactly would I be able to go about doing that? Is there an example somewhere of a trace capturing when failures occur? I also have a question about why there is another span for `volume.StreamOut` created at the same time as the `artifactSource.StreamTo` span? Is the `volume.StreamOut` span supposed to be passed into the `source.volume.StreamOut` method?

zoetian:
Dev notes: 
- figure out the error pattern; (inside the Span interface there are 3 diff. methods that we can explore)
- create a separate issue for `bindMount/ volumePath`
- add release notes
- add some instruction for AC test

atc/worker/artifact_source.go
zoetian: In our latest change, we used `tracing.End()` to address the case when streaming fails. 

The reason we used 2 spans here to capture the `source` volume handle and worker name as well as the `destination` volume handle and worker name.  Furthermore, it also gives us the ability to treat these as 2 separate operations ( `source to atc` and `atc to destination` ) for performance monitoring as well as error handling.

xtreme-sameer-vohra:
> Dev notes:
> 
> * figure out the error pattern; (inside the Span interface there are 3 diff. methods that we can explore)
> * create a separate issue for `bindMount/ volumePath`
> * add release notes
> * add some instruction for AC test

Added https://github.com/concourse/rfcs/discussions/58 for the `bindMount/ volumePath` enhancement

Issue #2534, opened 2018-08-27
The Great Project Restructuring of 2018


vito:
Context: https://discuss.concourse-ci.org/t/psa-the-great-code-restructing-of-2018/543

* [x] Automate creation of "one big repo"
  * Done: https://github.com/concourse/one-big-repo/tree/maintenance
* [x] Get a proof-of-concept going using Go 1.11 modules
  * Done: https://github.com/concourse/one-big-repo/tree/master
* [x] Ship ~4.1.0~ 4.2.1
* [x] Prep workflow around "one big repo"
  * [x] Design pipeline for new structure
    * [x] Figure out how `concourse-bosh-release` should be built (solution: auto-bump concourse blob)
  * [x] Create a really nice `README.md` (finally)
  * [x] Preserve `CODE_OF_CONDUCT.md`, `LICENSE.md`, `NOTICE.md` (update attribution header? contact legal)
  * [x] Create a new `CONTRIBUTING.md` for the newer (hopefully much much simpler) workflow
  * [x] Create a better issue template
    * [x] Instruct users to open "ops" issues on deployment-specific repo (try to limit `concourse/concourse` to actual bugs and feature requests)
* [x] Create `concourse-bosh-release` repo containing the current content of `concourse/concourse`
* [x] Clean up all stale branches under `concourse/concourse`
  * [x] Migrate any ones we still need to `concourse-bosh-release`
* [x] Create "one big repo" based on prior automation, this time "for real"
* [x] Push the "one big repo" as a merge (with distinct parents) to the `concourse/concourse` `master` branch.
* [x] Archive `atc`, `fly`, `go-concourse`, `tsa`, `web`, `testflight`, ...
* [x] ~~Restructure "one big repo"; look to other "one big repos" for patterns (e.g. `pkg/...`, `cmd/...`, `web/...`)~~ (long-term)

jama-pivotal:
"blocked" on a release....but @vito will keep working on READMEs and what-not

vito:
@aegershman Hah, thanks!

---

I just realized we're in a bit of a pickle here. The original plan was to just `push -f` over the `concourse` repo, and this is so that we can keep all the existing issues here, since they can't be moved.

[...]

PR #5596, opened 2020-05-13
fix overriding provider auth when TeamAuth is empty.

### LINKED ISSUE #5595: Locking Users out of Concourse by accident
## Summary

I can lock all the users out of concourse by setting an empty config file in fly `set-team` command including the main team users

## Steps to reproduce

```bash
echo "{}" >> empty_config.yml
fly -t my_target set-team -n team_name -c empty_config.yml
setting team: team_name

apply team configuration? [yN]: y
team created

fly -t my_target set-team -n main -c empty_config.yml
setting team: main

apply team configuration? [yN]: y
team updated
```

## Expected results

Fly prevents me from doing this by throwing an error.

## Actual results

Now I have lost all the access rights to the pipeline. 

When we have a look at the teams database we see,

concourse=> select * from teams;
 id  |   name   | admin | legacy_auth | nonce | auth
----+----------+-------+-------------+-------+------
  1 | main             | t     |                      |             | null # <--- 🤓
  2 | team_name | f     |                      |             | null # <--- 🤓

## Additional context

We were able to figure out in the backend API how this happens. So we can also fix this issue.

## Triaging info

* Concourse version: 5.8.1
* Browser (if applicable):
* Did this used to work? We do not know

shyamz-22:
Signed-off-by: Syamala Umamaheswaran <shyam@oss.volkswagen.com>

Co-authored-by: Lukas Benedix <lbenedix@oss.volkswagen.com>

# Why is this PR needed?

I can lock all the users out of concourse by setting an empty config file in fly `set-team` command including the main team users

Detailed description can be found in this issue: https://github.com/concourse/concourse/issues/5595

Preventing the Overriding existing of TeamAuth to null which when happens locks all the users out of concourse. 

closes #5595.

We only call updateProviderAuth when the `TeamAuth` is not empty. 

# Contributor Checklist

- [x] Unit tests
- [x] ~Integration tests~

# Reviewer Checklist

- [x] Code reviewed
- [x] Tests reviewed
- [x] Release notes reviewed
- [ ] PR acceptance performed


shyamz-22:
I think the API should error when auth config is empty. We will update that tomorrow

atc/api/teamserver/set.go
```diff
@@ -43,11 +43,13 @@ func (s *Server) SetTeam(w http.ResponseWriter, r *http.Request) {
 
 	if found {
 		hLog.Debug("updating-credentials")
-		err = team.UpdateProviderAuth(atcTeam.Auth)
-		if err != nil {
-			hLog.Error("failed-to-update-team", err, lager.Data{"teamName": teamName})
-			w.WriteHeader(http.StatusInternalServerError)
-			return
+		if len(atcTeam.Auth) > 0 {
```
pivotal-jamie-klassen, line 46: I suspect we could invalidate the request without even looking up the team, and save an extra level of indentation while we're at it.

I'm just thinking out loud too (and this suggestion might constitute "over-engineering") but I could imagine a `Validate() error` method on the `atc.Team` type, that gets called right after decoding it [up here](https://github.com/concourse/concourse/blob/0b79feeb5ab7fa205c60ae6d34334152390cc74c/atc/api/teamserver/set.go#L24-L29), looking something like

```go
	err = atcTeam.Validate()
	if err != nil {
		hLog.Error("invalid-team", err)
		w.WriteHeader(http.StatusBadRequest)
		return
	}
```

shyamz-22: @jamieklassen  yes at first look yesterday I did not have better understanding of what the code does. After looking carefully I forsee changes in two places:
1. Fly set-team command is validating values for command line auth params. so it should do the same for config file too see commit [here](https://github.com/volkswagen/concourse/commit/7693ba2daf039ba18b94883521e0a16dd44b6d47#diff-d5c54d5d65e15d27255ae063f0a616a5)

2. In order to be defensive its also good idea to validate in the backend, to be sure that auth config is never empty. See commit [here](https://github.com/volkswagen/concourse/commit/32014c630ad4124e521130daa79487fc19bb49f1)

I really like the validate method, its more readable than check if length is zero. If the CI passes i will do this refactor as well :)

jamieklassen: nice catch on the server side! Smart to do the same validation for `--main-team-config`, as discussed in concourse/docs#319.

jamieklassen:
@shyamz-22 I know you're planning to do a refactor on this as well, but just so you don't forget you might as well resolve the conflicts in the release notes at the same time.

shyamz-22:
@jamieklassen  we have added the validation in both fly and backend for authConfig. However we have the following question

The AuthConfig in fly package is the same as atc.TeamAuth. and both fly and backend needs to validate the same thing. Do you think it is a good idea to make it common for both? If so can you point to me how to structure things shared between atc and fly.

Also I would like to validate the role is correct, but accessing `accessor.MemberRole` etc leads to cyclic dependency so I avoided that. It is anyway validated in accessor to determine if the user can execute particular command.

shyamz-22:
@jamieklassen any updates? Thank you in advance

skymarshal/skycmd/flags.go
```diff
@@ -144,6 +146,10 @@ func (flag *AuthTeamFlags) formatFromFile() (AuthConfig, error) {
 		}
 	}
 
+	if err := atc.TeamAuth(auth).Validate(); err != nil {
+		return nil, ErrAuthNotConfiguredFromFile
+	}
```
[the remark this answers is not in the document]
shyamz-22: @jamieklassen what do you think about the error message. Bit mouthful i think,

```Go
var ErrAuthNotConfiguredFromFlags = errors.New("team authentication config from flags does not contain valid users or groups for specified role(s)")
var ErrAuthNotConfiguredFromFile = errors.New("team authentication config from file does not contain valid users or groups for specified role(s)")

```

shyamz-22:
> Looking good! Small stuff, and the perennial release notes merge conflict.
> 
> I'm not sure I totally understand
> 
> > The AuthConfig in fly package is the same as atc.TeamAuth
> 
> I couldn't find anything called 'authconfig' in fly. But in general I think things that are shared between `atc` and `fly` tend to live on the `atc` side. Seeing additions in `atc/team.go` makes sense to me. As for the `MemberRole` thing, I would be OK with moving those constants into `atc/team.go`, but no big deal.

Sorry for not being clear in what I meant and the package name was wrong too in my comment.
[skycmd.AuthConfig](https://github.com/volkswagen/concourse/blob/10bf14652c9655cf6129229e544e2bdce8f7c7d0/skymarshal/skycmd/flags.go#L249) & [atc.TeamAuth](https://github.com/volkswagen/concourse/blob/10bf14652c9655cf6129229e544e2bdce8f7c7d0/atc/team.go#L22) are both same types. Which leads to code like [this](https://github.com/volkswagen/concourse/blob/10bf14652c9655cf6129229e544e2bdce8f7c7d0/skymarshal/skycmd/flags.go#L149), [this](https://github.com/volkswagen/concourse/blob/10bf14652c9655cf6129229e544e2bdce8f7c7d0/atc/atccmd/command.go#L1499) and [this](https://github.com/volkswagen/concourse/blob/10bf14652c9655cf6129229e544e2bdce8f7c7d0/fly/commands/set_team.go#L98) where we convert the type from `AuthConfig` to `TeamAuth` like 
```Go
	team := atc.Team{Auth: atc.TeamAuth(authRoles)}
```
Is this by design? otherwise I can remove this duplicate and keep only the atc.TeamAuth.

jamieklassen:
> Is this by design?

I just combed through the code pretty carefully and I'm fairly sure this is no longer serving any architectural need. Removing the `skycmd.AuthConfig` type and converging on `atc.TeamAuth` makes the most sense. Most likely these duplicate types exist because `atc`, `skymarshal` and `fly` used to exist in separate repositories before #2534.

I tried out this patch, which realizes that you were alluding to:

```diff
diff --git a/atc/atccmd/command.go b/atc/atccmd/command.go
index 52cd29e60..7bd7c19fd 100644
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -1496,7 +1496,7 @@ func (cmd *RunCommand) configureAuthForDefaultTeam(teamFactory db.TeamFactory) e
 		return fmt.Errorf("default team auth not configured: %v", err)
 	}
 
-	err = team.UpdateProviderAuth(atc.TeamAuth(auth))
+	err = team.UpdateProviderAuth(auth)
 	if err != nil {
 		return err
 	}
diff --git a/fly/commands/set_team.go b/fly/commands/set_team.go
index 496be1859..d0e6a73d3 100644
--- a/fly/commands/set_team.go
+++ b/fly/commands/set_team.go
@@ -95,7 +95,7 @@ func (command *SetTeamCommand) Execute([]string) error {
 		displayhelpers.Failf("bailing out")
 	}
 
-	team := atc.Team{Auth: atc.TeamAuth(authRoles)}
+	team := atc.Team{Auth: authRoles}
 
 	_, created, updated, err := target.Client().Team(teamName).CreateOrUpdate(team)
 	if err != nil {
diff --git a/skymarshal/skycmd/flags.go b/skymarshal/skycmd/flags.go
index c4ca95da8..e039beb9f 100644
--- a/skymarshal/skycmd/flags.go
+++ b/skymarshal/skycmd/flags.go
@@ -58,7 +58,7 @@ type AuthTeamFlags struct {
 	Config     flag.File `short:"c" long:"config" description:"Configuration file for specifying team params"`
 }
 
-func (flag *AuthTeamFlags) Format() (AuthConfig, error) {
+func (flag *AuthTeamFlags) Format() (atc.TeamAuth, error) {
 
 	if path := flag.Config.Path(); path != "" {
 		return flag.formatFromFile()
@@ -77,7 +77,7 @@ func (flag *AuthTeamFlags) Format() (AuthConfig, error) {
 // The github connector has configuration for: users, teams, orgs
 // The cf conncetor has configuration for: users, orgs, spaces
 
-func (flag *AuthTeamFlags) formatFromFile() (AuthConfig, error) {
+func (flag *AuthTeamFlags) formatFromFile() (atc.TeamAuth, error) {
 
 	content, err := ioutil.ReadFile(flag.Config.Path())
 	if err != nil {
@@ -91,7 +91,7 @@ func (flag *AuthTeamFlags) formatFromFile() (AuthConfig, error) {
 		return nil, err
 	}
 
-	auth := AuthConfig{}
+	auth := atc.TeamAuth{}
 
 	for _, role := range data.Roles {
 		roleName := role["name"].(string)
@@ -146,7 +146,7 @@ func (flag *AuthTeamFlags) formatFromFile() (AuthConfig, error) {
 		}
 	}
 
-	if err := atc.TeamAuth(auth).Validate(); err != nil {
+	if err := auth.Validate(); err != nil {
 		return nil, ErrAuthNotConfiguredFromFile
 	}
 
@@ -157,7 +157,7 @@ func (flag *AuthTeamFlags) formatFromFile() (AuthConfig, error) {
 // TeamConfig has already been populated by the flags library. All we need to
 // do is grab the teamConfig object and extract the users and groups.
 
-func (flag *AuthTeamFlags) formatFromFlags() (AuthConfig, error) {
+func (flag *AuthTeamFlags) formatFromFlags() (atc.TeamAuth, error) {
 
 	users := []string{}
 	groups := []string{}
@@ -189,7 +189,7 @@ func (flag *AuthTeamFlags) formatFromFlags() (AuthConfig, error) {
 		return nil, ErrAuthNotConfiguredFromFlags
 	}
 
-	return AuthConfig{
+	return atc.TeamAuth{
 		"owner": map[string][]string{
 			"users":  users,
 			"groups": groups,
@@ -245,5 +245,3 @@ func (con *Connector) newTeamConfig() (TeamConfig, error) {
 
 	return res, nil
 }
-
-type AuthConfig map[string]map[string][]string
```

and it worked great!

shyamz-22:
I already have made the changes. Just getting it reviewed

Issue #5578, opened 2020-05-11
tooltip on sync icon when job data unavailable


jamieklassen:
The sync icon looks _a lot_ like a button. Because people are probably going to want to click it, it's probably worth having a tooltip that explains what it means:

@matthewpereira's initial design looks like this:
![variations master](https://user-images.githubusercontent.com/6372810/79888100-fdec2c00-83c9-11ea-8f07-b6893d54178a.png)

but that design was assuming it would be clickable, as described in #5556. I would suggest that the icon stays transparent, and for consistency's sake I'm going to propose having a tooltip that matches the one on the expose button:

<img width="305" alt="Screen Shot 2020-05-12 at 3 00 54 PM" src="https://user-images.githubusercontent.com/22056320/81734394-6bd3c280-9461-11ea-846d-f0498458bdbc.png">

@matthewpereira if you want to take this time to redesign both tooltips to the white-with-black-text that you illustrated we can do that, but I think getting the tooltip present is a higher priority, and the two tooltips being compatible is non-negotiable. If you have time to think about these tooltips before this issue is complete let's definitely incorporate a new design for both, but otherwise I think it's safe to redesign later.

pivotal-jamie-klassen:
~30 minutes of effort maximally following my nose yielded

<img width="154" alt="Screen Shot 2020-05-13 at 2 47 17 PM" src="https://user-images.githubusercontent.com/22056320/81852262-bbc98c80-9528-11ea-925e-b490664e834a.png">

The obvious problems with this - the opacity and positioning - are caused by the fact that the tooltip is an absolutely-position child of the icon, thus inheriting opacity and some positioning properties from its parent. This is an implementation detail of the [`iconWithTooltip` function](https://github.com/concourse/concourse/blob/master/web/elm/src/Views/Icon.elm#L16-L33). Resolutions include:

1. Using a "custom" implementation for this icon with tooltip to account for the parent icon's opacity, and tweaking the position of the child tooltip
2. Parametrizing the `iconWithTooltip` function to allow it to accomodate this new scenario
3. Refactoring the usages of `iconWithTooltip` in the Dashboard modules to use the `Tooltip` module, and reusing it here as well.

[...]

PR #5603, opened 2020-05-13
web: behaviour: add tooltip to sync icon

jamieklassen:
# Why is this PR needed?

When the `ListAllJobs` endpoint is disabled, the 'sync' icon that appears on pipeline cards is just _begging_ to be clicked, but it's not clickable. This will very-probably frustrate users.

Add an explanatory tooltip when hovering over the icon.

closes concourse/concourse#5578.

Use the functions from the `Tooltip` module, and refactor the other tooltips in the Dashboard modules to follow suit.

At least for now (@matthewpereira may jump in and change this requirement), the style matches the current visibility tooltip:

Screen Shot 2020-05-13 at 4 21 19 PM

# Contributor Checklist

- [x] Unit tests
- [x] ~Integration tests~

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Release notes reviewed
- [x] PR acceptance performed

### DIFF
--- a/web/elm/src/Application/Application.elm
+++ b/web/elm/src/Application/Application.elm
@@ -434,9 +434,15 @@ view model =
     in
     { title = title ++ " - Concourse"
     , body =
-        [ Tooltip.view model.session
-        , Html.map Update body
-        ]
+        List.map (Html.map Update)
+            [ SubPage.tooltip model.subModel model.session
+                |> Maybe.map (Tooltip.view model.session)
+                |> Maybe.withDefault (Html.text "")
+            , SideBar.tooltip model.session
+                |> Maybe.map (Tooltip.view model.session)
+                |> Maybe.withDefault (Html.text "")
+            , body
+            ]
     }
 
 
--- a/web/elm/src/Resource/Resource.elm
+++ b/web/elm/src/Resource/Resource.elm
@@ -7,6 +7,7 @@ module Resource.Resource exposing
     , handleDelivery
     , init
     , subscriptions
+    , tooltip
     , update
     , versions
     , view
@@ -84,6 +85,7 @@ import StrictEvents
 import Svg
 import Svg.Attributes as SvgAttributes
 import Time
+import Tooltip
 import UpdateMsg exposing (UpdateMsg)
 import UserState exposing (UserState(..))
 import Views.DictView as DictView
@@ -881,6 +883,11 @@ view session model =
                     ]
             ]
         ]
+
+
+tooltip : Model -> a -> Maybe Tooltip.Tooltip
+tooltip _ _ =
+    Nothing
 
 
 header : Session -> Model -> Html Message
--- a/web/elm/src/Tooltip.elm
+++ b/web/elm/src/Tooltip.elm
@@ -2,6 +2,7 @@ module Tooltip exposing (Model, handleCallback, view)
 
 import Browser.Dom
 import Build.Styles
+import Dashboard.Styles
 import EffectTransformer exposing (ET)
 import HoverState exposing (TooltipPosition(..))
 import Html exposing (Html)
@@ -175,6 +176,13 @@ view { hovered } =
                 [ Html.div SideBar.Styles.tooltipArrow []
                 , Html.div SideBar.Styles.tooltipBody [ Html.text pipelineID.pipelineName ]
                 ]
+
+        HoverState.Tooltip (Message.PipelineStatusIcon _) target ->
+            Html.div
+                (Dashboard.Styles.jobsDisabledTooltip
+                    ++ position { direction = Top, alignment = Start } target Nothing Nothing
+                )
+                [ Html.text "automatic job monitoring disabled" ]
 
         _ ->
             Html.text ""
--- a/web/elm/src/Tooltip.elm
+++ b/web/elm/src/Tooltip.elm
@@ -1,9 +1,11 @@
 module Tooltip exposing (Model, handleCallback, view)
 
+import Browser.Dom
 import Build.Styles
 import EffectTransformer exposing (ET)
 import HoverState exposing (TooltipPosition(..))
 import Html exposing (Html)
+import Html.Attributes exposing (style)
 import Message.Callback exposing (Callback(..))
 import Message.Effects as Effects
 import Message.Message as Message exposing (DomID(..))
@@ -14,12 +16,28 @@ type alias Model m =
     { m | hovered : HoverState.HoverState }
 
 
-type TooltipPolicy
+type TooltipCondition
     = AlwaysShow
     | OnlyShowWhenOverflowing
 
 
-policy : DomID -> TooltipPolicy
+type alias AttachPosition =
+    { direction : Direction
+    , alignment : Alignment
+    }
+
+
+type Direction
+    = Top
+    | Right
+
+
+type Alignment
+    = Start
+    | Middle
+
+
+policy : DomID -> TooltipCondition
 policy domID =
     case domID of
         SideBarPipeline _ ->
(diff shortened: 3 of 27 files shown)


pivotal-jamie-klassen:
I used the `hub` CLI to open this PR, so it started its life "ready to review" instead of as a draft.

pivotal-jamie-klassen:
As noted in concourse/concourse#5578, the API for the `Tooltip` module was bit clunky to use. I had the following complaints:

* It was surprising that I had to modify `toHtmlID` -- once I realized it was necessary (by building elm in debug mode and inspecting the `GotViewport` callback wrapping an error), it was not very clear how I might drive the change with a test. So I just continued this work as a spike.
* The `GetViewportOf` effect is not always used for tooltip-related purposes, despite the fact that it has an associated `TooltipPolicy`. `AlwaysShow` seems to be used as a no-op sentinel value in some cases.
* While it was nice to be able to use the `Bottom` variant for the tooltip position (on my first try the tooltip appeared to the right of the icon), it felt awkward to copy paste from the 'first occurrence tooltip' positioning logic -- maybe we should have better language for the "point from which the tooltip's position is determined" and the "point on the 'parent/anchor' element to place the tooltip at (relative to the aforementioned position-determining point" -- I'm not sure this sentence really clarifies the issue :grimacing:... I'll write this up in a clearer fashion elsewhere.

Other things to consider while turning this spike into a properly-tested feature:

* test the tooltip styles -- `PipelineCardTests` is probably a reasonable place to do it.
* refactor the 'visibility' tooltip to use `Tooltip` module as well.

jamieklassen:
Last night I refactored away the ambiguity of `GetViewportOf/GotViewport`, and now I'm going to to attempt to clarify the tooltip positioning problem.

Every tooltip has a "target" or "parent" or "thing it is attached to" or "tooltipee". I want better terminology for how a tooltip is positioned relative to this thing. The words "align" and "attach" feel helpful. Existing examples:
* the sidebar tooltips attach to the right of their targets, middle aligning
* the visibility tooltip attaches above its target, middle aligning
* the 'sync' status icon tooltip attaches above its target, start aligning

Broadly, then, there are four possible `AttachDirection`s:
* `Top`
* `Right`
* `Bottom`
* `Left`

and three possible `Alignment`s:
* `Start`
* `Middle`
* `End`

and these compose in any combination to make a `TooltipPosition`. Examples include:
* `Top End` - visibility tooltip, duration tooltip
* `Top Start` - first occurrence tooltip, pin button tooltips
* `Right Middle` - sidebar team/pipeline tooltips, dashboard job preview tooltips & pipeline name tooltips
* `Bottom Middle` - rerun button tooltip

 If a tooltip has an arrow (not all do), then its arrow is assumed to align `Middle`.

**note** the `Alignment` should always be interpreted as perpendicular to the `AttachDirection`, so that for `Left` or `Right` attachments, `Start` = top and `End` = bottom, whereas for `Top` or `Bottom` attachments, `Start` = left and `End` = right. This is a convention biased in favour or programmers familiar with left-to-right languages, but the source code itself reads left-to-right so it seems like a reasonable default. If we end up pursuing internationalization more seriously, we should allow for some kind of global "flow direction" that allows the meaning of `Start` and `End` to be appropriately updated.

web/elm/src/Tooltip.elm
```diff
@@ -14,12 +16,28 @@ type alias Model m =
...
+
+
+type Direction
+    = Top
+    | Right
+
+
+type Alignment
+    = Start
+    | Middle
```
aoldershaw, line 38: Cool approach! Just as food for thought, I had previously implemented anchor-based positioning where entities had 9 possible anchor-points:

```
x--x--x
|  |  |
x--x--x
|  |  |
x--x--x
```

And to position elements, you would anchor one point to another.

For instance, with the tooltip example, you'd anchor the `BottomLeft` anchor-point of the tooltip to the `TopLeft` anchor-point of the sync icon:

```
*---------*---------*
|         |         |
*---------*---------*
|         |         |
@--x--x---*---------*
|  |  |
x--x--x
|  |  |
x--x--x
```

Each point was defined as an `x_0`, `y_0`, and `anchor` (where `x_0` and `y_0` are the coordinates of the top-left of the entity), and each anchor had a tuple of percentages along the `x` and `y` directions (e.g. `BottomMiddle` was `(0.5, 1.0)`). Then, computing the actual coordinates of a point (relative to the dimensions of its entity) is easy:

```
x = x_0 + width * x_pct
y = y_0 + height * y_pct
```

Anchoring a point to another becomes a matter of computing `x` and `y` above for the parent point, and producing a new point where `x_0 = x`, `y_0 = y`, and `anchor` is set appropriately,

My use-case required more flexibility than we probably need with tooltips: a configuration like "anchor `BottomLeft` to `TopRight`" isn't likely to be useful.

```
      *---------*---------*
      |         |         |
      *---------*---------*
      |         |         |
x--x--@---------*---------*
|  |  |
x--x--x
|  |  |
x--x--x
```

Your way might also be more intuitive, but figured I'd provide an alternative idea!

jamieklassen: when you say "had previously implemented" - do you mean elsewhere in this codebase, or on another project? I had toyed with an API similar to what you've described, but as you suggest, I found it a bit overkill. The "anchoring" formalism as you've described seems strictly more expressive than the attach/align one, so we could even implement attach/align in terms of anchors. If there's precedent.

aoldershaw: I meant on another project so you won't be able to reuse it here

web/elm/src/Application/Application.elm
```diff
@@ -434,9 +434,15 @@ view model =
     in
     { title = title ++ " - Concourse"
     , body =
-        [ Tooltip.view model.session
-        , Html.map Update body
-        ]
+        List.map (Html.map Update)
```
jamieklassen, line 437: filterMap might be cleaner here...

web/elm/src/Tooltip.elm
```diff
@@ -1,57 +1,152 @@
...
+                    [ style "bottom" <| String.fromFloat (viewport.height - target.y - target.height) ++ "px" ]
+
+        -- ( Bottom, _ ) ->
+        --     [ style "top" <| String.fromFloat (target.y + target.height) ++ "px" ]
+        -- ( Left, Start ) ->
+        --     [ style "top" <| String.fromFloat target.y ++ "px" ]
+        -- ( Left, Middle height ) ->
+        --     [ style "top" <| String.fromFloat (target.y + (target.height - height) / 2) ++ "px" ]
+        -- ( Left, End ) ->
+        --     [ style "bottom" <| String.fromFloat (viewport.height - target.y - target.height) ++ "px" ]
```
zoetian, line 99: Great PR! Can we remove this comment here?

web/elm/src/Tooltip.elm
```diff
@@ -1,57 +1,152 @@
...
+                    [ style "left" <| String.fromFloat (target.x + target.width) ++ "px" ]
+
+        -- ( Bottom, Start ) ->
+        --     [ style "left" <| String.fromFloat target.x ++ "px" ]
+        -- ( Bottom, Middle width ) ->
+        --     [ style "left" <| String.fromFloat (target.x + (target.width - width) / 2) ++ "px" ]
+        -- ( Bottom, End ) ->
+        --     [ style "right" <| String.fromFloat (target.x + target.width) ++ "px" ]
+        -- ( Left, _ ) ->
+        --     [ style "right" <| String.fromFloat (viewport.width - target.x) ++ "px" ]
```
zoetian, line 121: also remove the comment here maybe? =)

web/elm/src/Resource/Resource.elm
```diff
@@ -883,6 +885,11 @@ view session model =
         ]
 
 
+tooltip : Model -> a -> Maybe Tooltip.Tooltip
+tooltip _ _ =
+    Nothing
+
+
```
zoetian, line 892: Glad that we're adding this into multiple places! We can definitely make use of `tooltip` in future.

web/elm/src/Tooltip.elm
```diff
@@ -1,57 +1,135 @@
...
     { m | hovered : HoverState.HoverState }
 
 
+type alias Tooltip =
+    { body : Html Message
+    , arrow : Maybe Arrow
+    , attachPosition : AttachPosition
+    }
+
+
```
zoetian, line 30: Sorry I'm nagging here, can we add a short comment to explain what `Arrow` means here? Just so that when ppl read it, they can understand it refers to the pointer that points to the tooltip? =)

PR #5620, opened 2020-05-18
Make task vars to be interpolated from var_sources and load_var.

### LINKED ISSUE #5609: Task step `vars:` field should support interpolation
## Summary

e.g.:

```yaml
- name: do-something-with-image-digest
  serial: true
  plan:
  - get: image-name
  - load_var: image
    file: image-name/digest
  - task: write-image-digest-somewhere
    file: task.yml
    vars:
      image: ((.:image))
```

This currently does not expand `((.:image))` - it passes it verbatim to the task. It should!

## Context

* Prior discussion: #5607

evanchaoli:
Fixes #5609

# Contributor Checklist

- [x] Unit tests
- [x] Integration tests

# Reviewer Checklist

- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [x] PR acceptance performed

### DIFF
--- a/atc/exec/task_step.go
+++ b/atc/exec/task_step.go
@@ -157,8 +157,15 @@ func (step *TaskStep) run(ctx context.Context, state RunState) error {
 		// external task - construct a source which reads it from file
 		taskConfigSource = FileConfigSource{ConfigPath: step.plan.ConfigPath, Client: step.workerClient}
 
-		// for interpolation - use 'vars' from the pipeline, and then fill remaining with cred variables
-		taskVars = []vars.Variables{vars.StaticVariables(step.plan.Vars), variables}
+		// for interpolation - use 'vars' from the pipeline, and then fill remaining with cred variables.
+		// this 2-phase strategy allows to interpolate 'vars' by cred variables.
+		if len(step.plan.Vars) > 0 {
+			taskConfigSource = InterpolateTemplateConfigSource{
+				ConfigSource: taskConfigSource,
+				Vars:         []vars.Variables{vars.StaticVariables(step.plan.Vars)},
+			}
+		}
+		taskVars = []vars.Variables{variables}
 	} else {
 		// embedded task - first we take it
 		taskConfigSource = StaticConfigSource{Config: step.plan.Config}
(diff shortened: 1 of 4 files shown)


evanchaoli:
@vito It seems `watsjs` test is broken, I have rerun 8 times, but all failed. I don't think my change should impact `watsjs`.

atc/exec/task_step.go
```diff
@@ -157,8 +157,15 @@ func (step *TaskStep) run(ctx context.Context, state RunState) error {
...
 		taskConfigSource = FileConfigSource{ConfigPath: step.plan.ConfigPath, Client: step.workerClient}
 
-		// for interpolation - use 'vars' from the pipeline, and then fill remaining with cred variables
-		taskVars = []vars.Variables{vars.StaticVariables(step.plan.Vars), variables}
+		// for interpolation - use 'vars' from the pipeline, and then fill remaining with cred variables.
+		// this 2-phase strategy allows to interpolate 'vars' by cred variables.
+		if len(step.plan.Vars) > 0 {
+			taskConfigSource = InterpolateTemplateConfigSource{
+				ConfigSource: taskConfigSource,
+				Vars:         []vars.Variables{vars.StaticVariables(step.plan.Vars)},
```
aoldershaw, line 165: I'm just trying to understand what's going on exactly so bear with me (this part of the code is pretty new to me):

Suppose you had the following pipeline:

```yaml
...
jobs:
- name: job
  plan:
  ...
  - load_var: foo
    file: some-resource/foo
  - task: run
    file: some-other-resource/task.yml
    vars:
      foo: ((.:foo))
```

where `some-other-resource/task.yml` was something like:

```yaml
...
run:
  path: echo
  args:
  - ((foo))
```

---

Am I correct in thinking that this first phase of interpolation result in the following `task.yml`:

```yaml
...
run:
  path: echo
  args:
  - ((.:foo))
```

and now we're able to interpolate `((.:foo))` using `step.delegate.Variables()`? Or am I misunderstanding the flow here?

evanchaoli: Yes, your understanding is right. Actually, I thought over the solution. The other solution I tried initially was to directly interpolate at the first place, which also worked, but introduced more code changes, thus needed more changes on unit tests. Then I figured out this way is simpler.

aoldershaw: Makes sense, clever solution!

Could you add a release note when you get the chance? Then I'm happy to approve the changes

evanchaoli: @aoldershaw I have just added a release note.

PR #5622, opened 2020-05-19
Sort team by name on the UI.

evanchaoli:
# Why is this PR needed?

Per https://github.com/concourse/concourse/discussions/5412#discussioncomment-7929, the UI shows teams in order of creation time, which doesn't make much sense, because nobody would remember when teams are created, so that such order roughly equals to random order.

Some users might be many teams, such "random" team order makes users hard to find a team on the UI.

To sort teams on the UI by team name.

Just changing order by team id to order by team name.

There is already an index on team name, thus order by team name should not hurt sql query performance.

# Contributor Checklist

- [x] Unit tests
- [x] Integration tests

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/team.go
+++ b/atc/db/team.go
@@ -545,7 +545,7 @@ func (t *team) PublicPipelines() ([]Pipeline, error) {
 			"team_id": t.id,
 			"public":  true,
 		}).
-		OrderBy("team_id ASC", "ordering ASC").
+		OrderBy("t.name ASC", "ordering ASC").
 		RunWith(t.conn).
 		Query()
 	if err != nil {
--- a/atc/db/team_factory.go
+++ b/atc/db/team_factory.go
@@ -112,7 +112,7 @@ func (factory *teamFactory) FindTeam(teamName string) (Team, bool, error) {
 func (factory *teamFactory) GetTeams() ([]Team, error) {
 	rows, err := psql.Select("id, name, admin, auth").
 		From("teams").
-		OrderBy("id ASC").
+		OrderBy("name ASC").
 		RunWith(factory.conn).
 		Query()
 	if err != nil {
--- a/atc/db/pipeline_factory.go
+++ b/atc/db/pipeline_factory.go
@@ -35,7 +35,7 @@ func (f *pipelineFactory) VisiblePipelines(teamNames []string) ([]Pipeline, erro
 
 	rows, err := pipelinesQuery.
 		Where(sq.Eq{"t.name": teamNames}).
-		OrderBy("team_id ASC", "ordering ASC").
+		OrderBy("t.name ASC", "ordering ASC").
 		RunWith(tx).
 		Query()
 	if err != nil {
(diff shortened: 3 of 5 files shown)


atc/db/team_factory.go
```diff
@@ -112,7 +112,7 @@ func (factory *teamFactory) FindTeam(teamName string) (Team, bool, error) {
 func (factory *teamFactory) GetTeams() ([]Team, error) {
 	rows, err := psql.Select("id, name, admin, auth").
 		From("teams").
-		OrderBy("id ASC").
+		OrderBy("name ASC").
```
clarafu, line 115: Does this affect the order that the teams are displayed in the UI? It seems to be used by the `ListTeams` endpoint, does that endpoint get used to display the ordering?

I tried removing this change and spun up a server locally and I don't see it making any difference. Maybe I'm missing something about where this endpoint is being used?

atc/db/team.go
```diff
@@ -545,7 +545,7 @@ func (t *team) PublicPipelines() ([]Pipeline, error) {
 			"team_id": t.id,
 			"public":  true,
 		}).
-		OrderBy("team_id ASC", "ordering ASC").
+		OrderBy("t.name ASC", "ordering ASC").
```
clarafu, line 548: Same here! Not sure if this is used for the ordering?

evanchaoli: `PublicPipelines()` is called by `pipelineserver`, which lists exposed pipelines to un-logged-in users. This changes makes teams with public pipelines also ordered by team name.

atc/db/team_factory.go
evanchaoli: `GetTeams` is called by accessor to generate auth token, thus order is not an issue. But I think it's better to keep consistent for all places where needs to order teams, otherwise code reader might be confused why here order by id and there order by name. WDYT?

evanchaoli:
@clarafu Thanks for your comments, I have replied them all.

atc/db/team_factory.go
clarafu: @evanchaoli Ahh yeah I see it's used by the `fly teams` command. Then that does make sense for it to also be ordered by name. I think if it was just used by the accessor to generate an auth token, I would've been slightly unsure about having it be ordered by name? Just because I would rather than the queries be completely representative of what they are trying to fetch (if ordering by name matters or not) over being consistent. But this method is being used by a fly command to display the teams so I am 100% for having it be ordered by name too!

clarafu:
@evanchaoli Thanks for answering all my questions! The code looks good to me and I manually tested that the ordering works through the web UI and through fly! 

Before I can merge this PR, would you be able to add a quick release note for this change to `latest.md`? Thanks a lot!

evanchaoli:
> Before I can merge this PR, would you be able to add a quick release note for this change to `latest.md`?

@clarafu Added release note.

PR #5631, opened 2020-05-20
Send a notification to the build tracker when a build starts

### LINKED ISSUE #5612: Send a notification to the build tracker when a build is started by the scheduler
## Summary

There have been a few complaints about builds being slow to start, I have a theory as to why this might be happening.

When the scheduler starts a build, it shows up in the web UI in a `started` state but there is a delay before the build actually starts to run the steps. This is possibly caused by the fact that we no longer run the build from the `scheduler` itself, but rather the `scheduler` will start the build and have the `build tracker` pick it up in order to run it. The build tracker runs on it's own interval so it might take a few seconds for the tracker to pick up the build. The following commit is the change that removed the running of builds by the scheduler.

https://github.com/concourse/concourse/commit/3a6a9b8c94ffa8a104a692f58da2baeeecc488f0#diff-ec5d156bedf6973365a4b99c7af3d828L197

In order to avoid having the delay between the build being started and the build running, we should send a notification to the `build tracker` right after a build is started within the `buildstarter`.

https://github.com/concourse/concourse/blob/0b79feeb5ab7fa205c60ae6d34334152390cc74c/atc/scheduler/buildstarter.go#L253-L257

clarafu:
# Why is this PR needed?

Fixes #5612

There has been some noticeable slowness to the starting of builds after a change was made to the scheduler where it is no longer responsible to start the build that it scheduled 3a6a9b8#diff-ec5d156bedf6973365a4b99c7af3d828L197. This means that after a build is marked as `started` by the scheduler 

https://github.com/concourse/concourse/blob/c57de700a5a038ed7db3fe27f109cb0a7eda725f/atc/scheduler/buildstarter.go#L216-L220

it will have to wait for the next tick of the `build tracker` before the build is "actually" started (actually started as in the steps are run).

https://github.com/concourse/concourse/blob/c57de700a5a038ed7db3fe27f109cb0a7eda725f/atc/builds/tracker.go#L56-L60

The build tracker runs on it's own 10 second interval so users can see an extra 10 second delay after their build has technically started before any steps are run.

This PR is trying to get rid of that delay between the time that the build is started by the scheduler and the build tracker starting to run the steps of the build.

closes #5612.

A notification will be sent to the build tracker whenever a build is started. This will kick off the build tracker to run every time the scheduler starts a build. I manually tested triggering a build with and without this change and without the notification, I saw builds taking approximately 20 seconds to complete.

Screen Shot 2020-05-20 at 3 05 23 PM

Rather than with this change, the build completes within 2 seconds.

Screen Shot 2020-05-20 at 2 59 07 PM

# Contributor Checklist

- [ ] Unit tests
- [ ] Integration tests

# Reviewer Checklist

- [x] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/build.go
+++ b/atc/db/build.go
@@ -375,6 +375,11 @@ func (b *build) Start(plan atc.Plan) (bool, error) {
 		return false, err
 	}
 
+	err = b.conn.Bus().Notify(atc.ComponentBuildTracker)
+	if err != nil {
+		return false, err
+	}
+
 	return true, nil
 }
 


jamieklassen:
Merged this PR into #5659 and saw an encouraging result. Without this change, time between a build's creation and start is over 20s:
Screen Shot 2020-05-27 at 4 39 48 PM

With this change, it reduces to ~8s:
Screen Shot 2020-05-28 at 10 46 28 AM

PR #5642, opened 2020-05-22
atc/db: structure: don't differentiate between pinning sources

andy-paine:
Rather than setting one of these fields to nil and then nil checking in
CurrentPinnedVersion, only the pinned version needs storing. A flag for
if the source of the pinning is config is kept for the `api/present`
package to use for the UI

[Anti-patterns page]: https://github.com/concourse/concourse/wiki/Anti-Patterns

# Why is this PR needed?
Simplify the contract of the `db/resource` package slightly to remove `ConfigPinnedVersion` and `ApiPinnedVersion` in favour of `PinnedVersion`

Now that config pinned versions are in the DB and fetched the same way, there is not real need to store both the config pinned version and the API pinned version where at least one is always nil. The `api/present` package needs to know which one was pinned to set a `pinned_in_config` flag so just expose a boolean from the `db/resource` package to match what `api/present` expects.

This is prep work to do a little tidy up as part of implementing #5456 

Remove `ApiPinnedVersion` and `ConfigPinnedVersion` from `db/resource` package in favour of just the `PinnedVersion` and a boolean flag for `PinnedInConfig`

`PinnedVersion()` then replaces `CurrentPinnedVersion()` which means we only have to do logic about which version is pinned once in this class.

# Contributor Checklist
- [x] Unit tests
- [ ] ~Integration tests~

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/resource.go
+++ b/atc/db/resource.go
@@ -37,8 +37,8 @@ type Resource interface {
 	CheckSetupError() error
 	CheckError() error
 	WebhookToken() string
-	ConfigPinnedVersion() atc.Version
-	APIPinnedVersion() atc.Version
+	PinnedInConfig() bool
+	PinnedVersion() atc.Version
 	PinComment() string
 	SetPinComment(string) error
 	ResourceConfigID() int
@@ -46,8 +46,6 @@ type Resource interface {
 	Icon() string
 
 	HasWebhook() bool
-
-	CurrentPinnedVersion() atc.Version
 
 	ResourceConfigVersionID(atc.Version) (int, bool, error)
 	Versions(page Page, versionFilter atc.Version) ([]atc.ResourceVersion, Pagination, bool, error)
@@ -112,8 +110,8 @@ type resource struct {
 	checkSetupError       error
 	checkError            error
 	webhookToken          string
-	configPinnedVersion   atc.Version
-	apiPinnedVersion      atc.Version
+	pinnedInConfig        bool
+	pinnedVersion         atc.Version
 	pinComment            string
 	resourceConfigID      int
 	resourceConfigScopeID int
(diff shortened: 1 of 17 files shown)


atc/db/team_test.go
```diff
@@ -1832,8 +1830,7 @@ var _ = Describe("Team", func() {
 			resource, found, err = savedPipeline.Resource("some-resource")
 			Expect(err).ToNot(HaveOccurred())
 			Expect(found).To(BeTrue())
-			Expect(resource.ConfigPinnedVersion()).To(BeNil())
-			Expect(resource.APIPinnedVersion()).To(BeNil())
```
andy-paine, line 1836: I think that the `ConfigPinnedVersion` and `APIPinnedVersion` are being used as a proxy for what is in the database rows here. This test is still valid I think if you trust that `resource.PinnedVersion()` will return non-nil if the version was pinned either by API or config (which is unit tested so should hold)

atc/db/resource.go
```diff
@@ -156,35 +154,35 @@ func (resources Resources) Configs() atc.ResourceConfigs {
 			Source:       r.Source(),
 			CheckEvery:   r.CheckEvery(),
 			Tags:         r.Tags(),
-			Version:      r.ConfigPinnedVersion(),
+			Version:      r.PinnedVersion(),
```
clarafu, line 157: This should probably only return the `PinnedVersion()` if the version is pinned through the config. 🤔 This method is used by `pipeline.Config()` https://github.com/concourse/concourse/blob/6b3f6c8a556c2e30a9f3a1c7aed41e375fb85a52/atc/db/pipeline.go#L258
in order to fetch the config for a pipeline and we probably don't want to return a version pinned through the API as part as the pipeline configuration.

andy-paine:
Thanks for looking at this @clarafu - sorry I didn't get to update it! I think given that the original changes I wanted to make that this was the groundwork for (`version: every` on resources) isn't needed/possibly isn't wanted anymore, following up this PR isn't necessary. Also, since there are merge conflicts by now (quel surprise), I'm just going to close this.

PR #5645, opened 2020-05-22
atc,tf: structure: robustify active users test

### LINKED ISSUE #5628: Flaky Testflight test "track active user"
**What problem are you facing**

[Testflight](https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/testflight-postgres/builds/55) [suite](https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/testflight-postgres/builds/36) [fails](https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/testflight/builds/221) [semi-regularly](https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/testflight/builds/213) on the `track active user` test.

**What would make this better**

Make the test or feature it's testing more dependable, or if it's not significant enough remove the test entirely

jamieklassen:
# Why is this PR needed?

There are a few times when this particular test flakes, and this blocks up our delivery pipeline -- rerunning the whole testflight suite is quite time-consuming, adding tens of minutes to our cycle time.

Replace a flaky testflight test with a rock-solid (albeit a bit slow) atc integration test.

closes #5628.

I originally wrote this test as a throwaway during a discussion about an issue I wasn't actively working onand didn't expect it to stick around in exactly the same form -- famous last words.

I was able to consistently reproduce the testflight flake mentioned in the linked issue by always starting with an empty database. Initially I rewrote the test as an atc integration test just to make it easier to attach a debugger -- I was finding the `docker-compose down -v && docker-compose up -d && hack/trace web` dance a bit clunky. Conveniently once I did, the flake became completely reproducible because every atc integration test starts with a fresh database.

I didn't really thoroughly investigate, but I strongly suspect the flake came from the fact that there is an asynchronous `batcher` that is responsible for updating active user data - so there can be a delay between logging in and when the users table gets updated.

On the one hand, it may be worth asking "why doesn't the database get synchronously updated?" and the answer is, roughly, "optimization reasons".  Allowing for that, then, presumably there are some thorough unit tests for this carefully-optimized procedure. So why have an integration test at all if there's some non-determinism?

Because this feature is really, really important. It gets used to define billing in some cases. So it's worth having a potentially slow-running and complex integration test if it will verify that billing still works properly.

# Contributor Checklist

- [x] ~Unit tests~
- [x] Integration tests

# Reviewer Checklist

- [x] ~Code reviewed~
- [ ] Tests reviewed
- [ ] PR acceptance performed -- Try running the old testflight test with a fresh database and you'll see what I mean!


atc/integration/active_users_test.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+var _ = Describe("ATC Integration Test", func() {
+	It("active users are tracked", func() {
+		client := login(atcURL, "test", "test")
+		twoMonthsAgo := time.Now().AddDate(0, -2, 0)
+		Eventually(func() []atc.User {
+			users, _ := client.ListActiveUsersSince(twoMonthsAgo)
+			return users
+		}, atccmd.BatcherInterval).Should(
```
jamieklassen, line 20: Since this is a global now, can we modify it to be shorter to make the test go by faster?

aoldershaw, line 20: Since this is a global now, can we modify it to be shorter to make the test go by faster?

aoldershaw: That said, setting it to `1*time.Second` doesn't really seem to make the test much faster anyway - it typically takes ~3s for me with the default `atccmd.BatcherInterval` which isn't too bad

jamieklassen: yeah i love the idea of tuning it down, but i think that's a problem for another day -- in general i want to be especially careful and intentional making changes to this sort of thing because Billing.

PR #5651, opened 2020-05-24
experiment: atc: support Elasticsearch as a build event store

aoldershaw:

This PR is an experimental implementation of https://github.com/concourse/rfcs/pull/53, enabling  Elasticsearch to be used as a store for build events.

This PR doesn't address all of the possible applications mentioned in the RFC, but acts as a proof of concept for some of the ideas discussed.

The Elasticsearch implementation can be tested like so:

```
$ docker-compose -f docker-compose.yml -f hack/overrides/elasticsearch.yml up -d
```

This will configure Concourse to use the Elasticsearch Event Store. It will also spin up a Kibana UI at port 5601.

* Introduces an `EventStore` interface (more detail in the RFC) and two implementations of that interface - Postgres and Elasticsearch
  * The Postgres implementation behaves *similarly* to what we have now. However, rather than using database triggers to create/delete `pipeline_build_events_x` and `team_build_events_y` tables, this is done in code.
  * Rather than using database migrations, the database schema is setup in `Store.Setup`
* Extends the notification bus implementation to send build events using Postgres' `NOTIFY` command (previously, we just used `NOTIFY` for alerting, rather than sending data)

# Contributor Checklist

- [ ] Unit tests (some)
- [ ] Integration tests (not yet)

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


evanchaoli:
This is a great start! I just quickly went the code change, and get a few thoughts:

1. We should still allow small setups to use PG.
2. I'm thinking loudly. Maybe we can just wrap ES connection in `db.Conn`, so that we don't need to pass `EventStore` to everywhere. Which also enables a possibility where we can store more things to ES, such as trace logs.
3. Based on 2, we don't have to migrate existing build logs. Just load old build logs from PG and load new build logs from ES, maybe we can add a flag to `builds` table to distinguish where build events are stored.

aoldershaw:
Hey @evanchaoli, thanks for looking over the PR! With respect to some of your comments:

1. Postgres is currently the default - Elasticsearch is only used if `--elasticsearch-url` is provided
2. That's interesting. I don't know if it makes sense to have anything too specific to Elasticsearch on the `db.Conn`, since the idea of `EventStores` is that they should be interchangeable. I could try out storing the `EventStore` directly on the `db.Conn`, though!

Could you elaborate on your second point:
> Which also enables a possibility where we can store more things to ES, such as trace logs

By this, do you mean distributed tracing solutions like Jaeger? If so, from what I can tell, Concourse forwards spans to Jaeger directly, and Jaeger is configured to store the spans in something like ES. Take a look at `hack/overrides/jaeger.yml` which sets up Jaeger+ES locally:

https://github.com/concourse/concourse/blob/d0cb4f9bb61b05401f7ef2fb88aab3f5332903ca/hack/overrides/jaeger.yml#L18-L29

3. This kind of sounds similar to my comment on the RFC suggesting an "on the fly migration" technique: https://github.com/concourse/rfcs/pull/53#discussion_r423100193. It wasn't a fully formed idea, though - it always checks the "primary" event store first (in this case, ES), and if that returns no events, it would try to "fallback" event store (in this case, PG). The idea of storing in the DB in which event store the events live is possibly a better approach to this

evanchaoli:
> Could you elaborate on your second point:
> 
> > Which also enables a possibility where we can store more things to ES, such as trace logs

Oh, I just saw `tracing` was added but I had never read tracing related code yet, so I thought tracing data were just sent to logs. Actually I just thought loudly, I don't have a clear idea what else can be also sent to ES, maybe even logs. Now concourse logs, tracing logs, build logs go to three sinks, maybe they can all go to ES? Is it a future possibility? I don't know. My point was just, if ES connection is wrapped in `db.Conn`, then you don't have to pass `EventStore` to everywhere, actually only delegaters write build logs, but you now need to touch so many files; then the side benefit is to enable some future possibility. Sorry for the confusion.

clarafu:
Hey @aoldershaw! So sorry I've delayed looking at your PR because it's a draft and I've been swamped with stuff but I'll be taking the next few days to look at it!

aoldershaw:
@clarafu definitely no rush on this - it's a long way from being complete, and the approach is pretty up in the air at this point. There was an [interesting idea proposed on the RFC](https://github.com/concourse/rfcs/pull/53#discussion_r430196099) around not baking the implementations into Concourse core, and some other [concerns about the additional complexity](https://github.com/concourse/rfcs/pull/53#pullrequestreview-420275029) raised that definitely have merit, which I've been trying to think about some more (but haven't had too much time to work on it recently!)

aoldershaw:
Gonna close this for now. I think the overall approach needs a little more thought put into it, and haven't had much of a chance to work on this recently

deepakmohanakrishnan07:
Any thoughts if resuming this effort? This will significantly improve the PG performance, reduce cost and maintenance.

Issue #3551, opened 2019-03-20
Be able to specify team worker without the need of restart Concourse (TSA)


gaelL:
## What challenge are you facing?

As partially described in those both "related issues" :

  * https://github.com/concourse/concourse/issues/1105#issuecomment-473360524
  * https://github.com/concourse/concourse/issues/3549

I'm working with a mutualized Concourse server and dedicated workers (only) to teams. For each "users/customer" we create a dedicated team.
We don't provide default workers. A customer to run his jobs have to connect his own worker(s) to his team.

Let's say we have customer A and customer B.
Customer A will have a dedicated team A, on which he will assign Worker A to run his own jobs.

We want to prevent jobs of customer A running into Workers of customer B, and we also want to prevent Customer B to connect his workers on team A.

So to satisfy this issue, the current `tsa-team-authorized-keys` feature does perfectly the job.

BUT it has some drawbacks.

If we want to automate a bit the process, each time we create a new team in Concourse, we have to add a new param on the server and trigger a restart to take care of it.

eg 
```
--tsa-team-authorized-keys=teamA:/teamAfile
--tsa-team-authorized-keys=teamB:/teamBfile
....
--tsa-team-authorized-keys=teamN:/teamNfile
```

So if we create several teams by day, it becomes quickly hard to maintain, and creates potential downtime for customers.

**update** This would also improve k8s setup to put authorized keys in one configmap.

There ares 2 issues related to that :

  * 1) The need to add new command-line arguments for new Concourse team created.
  * 2) The need to restart Concourse when we add a key in those files (CF issue https://github.com/concourse/concourse/issues/1105#issuecomment-473360524)

## What would make this better?

I internally implemented a solution for 1) by adding a yaml authorized key flag :

```
--tsa-yaml-team-authorized-keys                              Path to file containing keys to authorize, in SSH authorized_keys yaml format. [$CONCOURSE_TSA_YAML_TEAM_AUTHORIZED_KEYS]
``` 

This flag targets a yaml file which contains authorized keys for each team in yaml format, following a bit this format https://bosh.io/jobs/tsa?source=github.com/concourse/concourse&version=4.2.1#p%3dteam_authorized_keys

```
  - team: teamA
    ssh_keys:
      - ssh-rsa keyA
  - team: teamB
    ssh_keys:
      - ssh-rsa keyB
```

So it solves the 2) issue, when I create a new Concourse team, I simply add a section for that team.
Then I tried the solution proposed in the issue for 1).

[...]

PR #5652, opened 2020-05-25
tsa: handling of SIGHUP to reload workers keys

### LINKED ISSUE #1105: TSA should support reloading authorized keys without restart
When changing the authorized_keys file to dynamically add a new worker, the TSA doesn't recognize the new key unless the TSA is restarted.  This makes dynamically provisioning a worker via automation difficult.

Please add the ability to automatically pick up changes to the authorized_key files or an API/fly command to trigger the reload.

---
Moved from concourse/tsa#5
/cc @iSynaptic

gaelL:
# Why is this PR needed?

This pull request fix the issue #1105 by adding TSA support for reloading authorized keys without restart based on signal `SIGUSR1`

As described in #1105 and #3551 it is really handy for HA purpose to be able to add worker keys without having to restart a Concourse web/TSA process.

It also rely on this pull request that have been merged https://github.com/concourse/flag/pull/5

Provide a solution for user to dynamically add or remove worker keys without having to restart Concourse.

closes #1105 .

I choose to use signal `SIGHUP` that any user can send with the following command:

```
kill -s SIGHUP <PID of Concourse web>
```

The implementation is really simple and for a user point of view it's really easy to send this signal to concourse when you update any authorized keys file.
Signal usage have been inspired by Concourse worker process https://concourse-ci.org/concourse-worker.html#restarting-a-worker

# Contributor Checklist

- [ ] Unit tests
- [ ] Integration tests

Let's discuss about those points if the technical solution is approved.

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/cloudfoundry/socks5-proxy v0.0.0-20180530211953-3659db090cb2 // indirect
 	github.com/concourse/baggageclaim v1.6.6
 	github.com/concourse/dex v0.0.0-20200417202922-dcbe94f28c88
-	github.com/concourse/flag v1.0.0
+	github.com/concourse/flag v1.0.1-0.20200521134421-2be0b225f922
 	github.com/concourse/go-archive v1.0.1
 	github.com/concourse/retryhttp v1.0.2
 	github.com/containerd/cgroups v0.0.0-20191220161829-06e718085901 // indirect
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/cloudfoundry/socks5-proxy v0.0.0-20180530211953-3659db090cb2 // indirect
 	github.com/concourse/baggageclaim v1.6.6
 	github.com/concourse/dex v0.0.0-20200417202922-dcbe94f28c88
-	github.com/concourse/flag v1.0.1-0.20200521134421-2be0b225f922
+	github.com/concourse/flag v1.1.0
 	github.com/concourse/go-archive v1.0.1
 	github.com/concourse/retryhttp v1.0.2
 	github.com/containerd/cgroups v0.0.0-20191220161829-06e718085901 // indirect
--- a/tsa/tsacmd/command.go
+++ b/tsa/tsacmd/command.go
@@ -144,21 +144,18 @@ func (cmd *TSACommand) Runner(args []string) (ifrit.Runner, error) {
 		httpClient:        httpClient,
 		sessionTeam:       sessionAuthTeam,
 	}
-	// Starts a goroutine that his purpose is to basically listen to the
-	// SIGHUP syscall to then reload the config.
-	// For now it only reload the TSACommand.AuthorizedKeys but any
-	// other configuration could be added
+	// Starts a goroutine whose purpose is to listen to the
+	// SIGHUP syscall and reload configuration upon receiving the signal.
+ 	// For now it only reloads the TSACommand.AuthorizedKeys but
+	// other configuration can potentially be added.
 	go func() {
-		// Set up channel on which to send signal notifications.
-		// We must use a buffered channel or risk missing the signal
-		// if we're not ready to receive when the signal is sent.
-		c := make(chan os.Signal, 1)
+		reloadWorkerKeys := make(chan os.Signal, 1)
 		defer close(c)
 		signal.Notify(c, syscall.SIGHUP)
 		for {
 
 			// Block until a signal is received.
-			_ = <-c
+			<-c
 
 			logger.Info("reloading-config")
 
@@ -174,7 +171,7 @@ func (cmd *TSACommand) Runner(args []string) (ifrit.Runner, error) {
 				continue
 			}
 
-			// compute again the config so it's updated
+			// Reconfigure the SSH server with the new keys
 			config, err := cmd.configureSSHServer(sessionAuthTeam, cmd.AuthorizedKeys.Keys, teamAuthorizedKeys)
 			if err != nil {
 				logger.Error("failed to configure SSH server: %s", err)
--- a/tsa/tsacmd/command.go
+++ b/tsa/tsacmd/command.go
@@ -149,13 +149,13 @@ func (cmd *TSACommand) Runner(args []string) (ifrit.Runner, error) {
 	// For now it only reload the TSACommand.AuthorizedKeys but any
 	// other configuration could be added
 	go func() {
+		// Set up channel on which to send signal notifications.
+		// We must use a buffered channel or risk missing the signal
+		// if we're not ready to receive when the signal is sent.
+		c := make(chan os.Signal, 1)
+		defer close(c)
+		signal.Notify(c, syscall.SIGHUP)
 		for {
-
-			// Set up channel on which to send signal notifications.
-			// We must use a buffered channel or risk missing the signal
-			// if we're not ready to receive when the signal is sent.
-			c := make(chan os.Signal, 1)
-			signal.Notify(c, syscall.SIGHUP)
 
 			// Block until a signal is received.
 			_ = <-c
--- a/tsa/tsacmd/command.go
+++ b/tsa/tsacmd/command.go
@@ -13,6 +13,9 @@ import (
 
 	yaml "gopkg.in/yaml.v2"
 
+	"os/signal"
+	"syscall"
+
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/skymarshal/token"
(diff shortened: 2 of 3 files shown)


gaelL:
As discussed in #1105 @muntac feel free to have a look

tsa/tsacmd/command.go
```diff
@@ -141,6 +144,46 @@ func (cmd *TSACommand) Runner(args []string) (ifrit.Runner, error) {
...
+	// For now it only reload the TSACommand.AuthorizedKeys but any
+	// other configuration could be added
+	go func() {
+		for {
+
+			// Set up channel on which to send signal notifications.
+			// We must use a buffered channel or risk missing the signal
+			// if we're not ready to receive when the signal is sent.
+			c := make(chan os.Signal, 1)
+			signal.Notify(c, syscall.SIGUSR1)
```
cirocosta, line 158: From my experience, SIGHUP tends to be more commonly used for "hey daemon, reload your config" overall, and SIGUSR1 has a very specific purpose for workers (which would mean that if we want to extend the idea of "reload on signal received", this would collide with the `land`-on-SIGUSR1 logic we currently have there).

Wdyt of going w/ SIGHUP?

gaelL: I agree. I updated the PR and commits to use SIGHUP

xescugc:
IDK how the documentation part works, but I would suggest to comment it somewhere as if not it's a hidden feature hehe.

Could potentially be commented on the `AuthorizedKeys` flag description at least :smile:

tsa/tsacmd/command.go
```diff
@@ -141,6 +144,46 @@ func (cmd *TSACommand) Runner(args []string) (ifrit.Runner, error) {
...
+	// For now it only reload the TSACommand.AuthorizedKeys but any
+	// other configuration could be added
+	go func() {
+		for {
+
+			// Set up channel on which to send signal notifications.
+			// We must use a buffered channel or risk missing the signal
+			// if we're not ready to receive when the signal is sent.
+			c := make(chan os.Signal, 1)
+			signal.Notify(c, syscall.SIGHUP)
```
aoldershaw, line 158: I think it's best to pull this out of the loop - otherwise, the number of `chan`s that are being used to send signals to will grow without bound

gaelL: GC would collect them as they would not be used anymore but it's maybe safer or more explicite to put it out of the for loop.

I will push a commit for it

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/cloudfoundry/socks5-proxy v0.0.0-20180530211953-3659db090cb2 // indirect
 	github.com/concourse/baggageclaim v1.6.6
 	github.com/concourse/dex v0.0.0-20200417202922-dcbe94f28c88
-	github.com/concourse/flag v1.0.0
+	github.com/concourse/flag v1.0.1-0.20200521134421-2be0b225f922
```
muntac, line 25: I went ahead and made a new [release](https://github.com/concourse/flag/releases/tag/v1.1.0) of `concourse/flag`. You can now change this to reference `v1.1.0` directly without a commit SHA.

gaelL:
@muntac this is how I test it:

This PR add a new goroutine on web/atc/tsa process. This goroutine will "run" each time it catch a `kill -s SIGHUP`.
Before we go forward what do we expect to see:

The goroutine code contain one logger line `logger.Info("reloading-config")`. So we expect to see `reloading-config` in one of the log of Concourse web container when we send the right kill signal.

# Basic test with AuthorizedKeys

I simply run docker compose in one term. And in another term I run a kill signal on the concourse web process :

```bash
docker-compose up
```

```bash
docker exec -it concourse_web_1 bash -c 'kill -s SIGHUP $(pgrep concourse)'
```

In my docker compose term I have the following log displayed each time I send the kill signal:

```
web_1     | {"timestamp":"2020-06-11T09:15:44.719067381Z","level":"info","source":"tsa","message":"tsa.reloading-config","data":{}}
```

As noticed, this log line is the expected one.


# Second test with authorizedKeys using team yaml file

For this it require a little patch to configure Concourse web to read the TEAM_AUTHORIZED_KEYS_FILE. But also a little change to generate the file for our test:

```
diff --git a/Dockerfile b/Dockerfile
index aad6300ba..8b5b8039c 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,6 +30,12 @@ RUN concourse generate-key -t ssh -b 1024 -f /concourse-keys/tsa_host_key
 RUN concourse generate-key -t ssh -b 1024 -f /concourse-keys/worker_key
 RUN cp /concourse-keys/worker_key.pub /concourse-keys/authorized_worker_keys

+RUN echo '- team: main\n\
+  ssh_keys:\n\
+    - '$(cat /concourse-keys/worker_key.pub)'\n\
+    - fake errorkey\n'\
+  > /concourse-keys/authorized_worker_keys.yml
+
 # 'web' keys
 ENV CONCOURSE_SESSION_SIGNING_KEY     /concourse-keys/session_signing_key
 ENV CONCOURSE_TSA_AUTHORIZED_KEYS     /concourse-keys/authorized_worker_keys
index ad221f201..0d64c20a8 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,6 +31,7 @@ services:
       CONCOURSE_CLUSTER_NAME: dev
       CONCOURSE_ENABLE_ARCHIVE_PIPELINE: "true"
+      CONCOURSE_TSA_TEAM_AUTHORIZED_KEYS_FILE: /concourse-keys/authorized_worker_keys.yml

   worker:
     build: .
```

I add a new env var in docker compose to configure the usage of TEAM_AUTHORIZED_KEYS_FILE.
And I add a `RUN` in the `Dockerfile` to create a valid yaml file.
As example as content, I used the public key generated for the worker and a fake one to trigger also the error logger.

Same as above we expect the log line `reloading-config`, also `logger.Info("load-team-authorized-keys-loaded", lager.Data{"team": t.Team, "key": k})` and the error one triggered by our fake key.

let's test it:

```
# Term 1 (adding --build to make sure we take into account our changes into Dockerfile)
docker-compose up --build

# Term 2
docker exec -it concourse_web_1 bash -c 'kill -s SIGHUP $(pgrep concourse)'
```

Output expected:

```
web_1     | {"timestamp":"2020-06-11T09:39:18.558549543Z","level":"info","source":"tsa","message":"tsa.reloading-config","data":{}}
web_1     | {"timestamp":"2020-06-11T09:39:18.558905580Z","level":"info","source":"tsa","message":"tsa.load-team-authorized-keys-loaded","data":{"key":"ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAAAgQDMadAtprsAjBXQzwK8u5HCEHOtYYQmtHCNEL1SzBBlS9SUegfKAosS4secV7HdAU3jA7MG4rvj3AFRHQFMHzGfj/IvV3JcHCs8cwrw+8yLqP8L/if2Fi3PtLmTeUBacyYLH2H7zRzFhv0ywih7+UI9d6FZfVphWcepoV0OOqn6nQ==","team":"main"}}
web_1     | {"timestamp":"2020-06-11T09:39:18.558954648Z","level":"error","source":"tsa","message":"tsa.load-team-authorized-keys-parse","data":{"error":"Invalid format, ignoring (fake errorkey): ssh: no key found"}}
```

Let me know if you need more details or help to test it.

muntac:
@gaelL Thanks a lot! I kept making the basic mistake of sending the signal to the `worker` (instead of the `web` where the TSA is) as I kept thinking "worker keys"! :-|

Anyway, it works for me now. I'm going to request some changes to the comments and then we should be good to go. Thanks for your patience.

tsa/tsacmd/command.go
```diff
@@ -141,6 +144,46 @@ func (cmd *TSACommand) Runner(args []string) (ifrit.Runner, error) {
...
+		// Set up channel on which to send signal notifications.
+		// We must use a buffered channel or risk missing the signal
+		// if we're not ready to receive when the signal is sent.
+		c := make(chan os.Signal, 1)
+		defer close(c)
+		signal.Notify(c, syscall.SIGHUP)
+		for {
+
+			// Block until a signal is received.
+			_ = <-c
```
muntac, line 161: I believe the idiomatic way doesn't require the blank identifier. I would change it to keep it consistent with how blocking is done [elsewhere](https://github.com/concourse/concourse/blob/c4261d94405c3a7cc200b7ed8bd3f0ed84b12b4b/worker/volume_sweeper.go#L92) in the code.
```suggestion
			<-c
```

muntac:
Lastly, this would benefit from being mentioned in the docs as @xescugc suggested. It seems most appropriate as a subsection under the [operating a web node](https://concourse-ci.org/concourse-web.html#web-operation) section.

This would involve a PR to the [docs repo](https://github.com/concourse/docs/) and the change would be in [here](https://github.com/concourse/docs/blob/c3ce45b93efcb2850e4912afb2feba7e4513c465/lit/docs/install/web.lit#L141).

@gaelL Let me know if you are able to get around to that. Otherwise, I can try updating it myself.

gaelL:
thanks for this review. I will give a try with the doc

xescugc:
On https://github.com/concourse/concourse/pull/5652/commits/ce04b9f7bbdeac6970454e16f4227a42bb09c1ad you renamed `reloadWorkerKeys := make(chan os.Signal, 1)` so all the references to `c` have to also be renamed.

When `defer close(c)` and `<- c`

gaelL:
Thanks, the code have been fixed and tested again

muntac:
Just realized we should write some sort of test for this feature. One option is to do it in the acceptance test suite, [topgun](https://github.com/concourse/concourse/tree/master/topgun).

In `topgun/k8s` we could bring up a Concourse deployment in a test, generate a new key, and then send the SIGHUP signal to the web process. Not exactly sure how to programatically test if the worker is using the new key 🤔 

Still thinking through if this is the best place to test.

jamieklassen:
I think you should actually be able to exercise this using similar setup/helpers to `atc/integration`. Maybe a `tsa/integration` suite would make sense. I haven't totally thought it through, but I know that `ifrit.Runner`s have a signal channel, which seems like a reasonable interface to test against.

gaelL:
FYI I created the Doc pull request related to this one : https://github.com/concourse/docs/pull/339

I will try to have a look on tests. It indeed does not look that easy to test. Any leads or help are welcome on this part :p

gaelL:
@muntac Thanks for your help on this pull request :+1:

Issue #5602, opened 2020-05-13
trace build creation


jamieklassen:
As a concourse operator
I want a way to link the [span context](https://github.com/concourse/concourse/pull/5575) associated with a resource version to the span context associated with a build
So that i can collect and reason about timing information for the complete causal flow of events through concourse

Given that a resource `r` has just checked
And produced a new version `v` with span context `sc`
And there is a job `j` with `r` as an input with `trigger: true`
When the scheduler creates a pending build `b` for `j`
Then `b`'s span context will be created with a [link](https://github.com/open-telemetry/opentelemetry-specification/blob/bdf39cb935a6f48badc687daeb78a66dc352761f/specification/overview.md#links-between-spans) to `sc`

**note** can we generalize the definition of `b` above to describe all cases where `b` is _caused by_ `v`? The definition of _cause_ has yet to be determined.

pivotal-jamie-klassen:
unassigning myself as I seem to still be somewhat occupied with #5405 - I'll pick it up if I'm done in time, but otherwise it's up for grabs!

clarafu:
When we add the span context to resource versions during the saving of versions from the check step, it would be nice if we can also create a new span within the same trace during the check step in order to capture any errors that comes from the check step. There is more context about the motivation behind this here https://github.com/concourse/concourse/pull/5575#discussion_r425823210.

There is some work that has been done for adding the span context to resource versions in the `wip-tracing` branch.

zoetian:
We have done most of the preliminary refactoring for tracing version creation (a dependency of this issue), and it should be possible to continue tomorrow with driving out the `SaveVersions` method on the `db.Check` interface.

PR #5657, opened 2020-05-25
atc: behaviour: check step runs in the same context as checker

### LINKED ISSUE #5654: check step should run in the same span context as the checker
## Summary

The `Run` method on an `exec.CheckStep` has a `context.Context` as a parameter. This parameter should include the SpanContext following the scan responsible for scheduling the check in question, as defined in #5575.

## Context

 Once this is done, the `CheckStep` can be used to
* propagate this span context into the check container via environment variables - #5423
* propagate this span context into the resource versions produced by the check - #5602

topherbullock:
# Why is this PR needed?

This change, while small in impact, should allow for more sophisticated tracing of the lidar package, including propagating span context into check containers (related to #5423) and propagating check-related span context into resource versions (as part of #5602).

closes #5654.

We were happy about the "sensing"/inspection possibilities provided by the `testtrace` package, and decided it would be worth the preliminary refactor of the `tracing.ConfigureTracer` method. This allowed us to write a pretty expressive checker test without relying too heavily on artisanal counterfeiter fakes. Overall this should improve testability of tracing-related functions for the future.

It remains to be seen whether the thread-unsafe mutation of the `tracing.Configured` variable causes any problems when running in parallel, say, on CI.

Implementation-wise, we noticed that the checker only interacted with a check step by way of an engineCheck, so adding a context parameter to the `Check` method on the `Engine` interface felt reasonable.

This change has no external user-facing impact, so it doesn't need documentation.

# Contributor Checklist

- [x] Unit tests
- [x] ~Integration tests~ change has very small scope

# Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -15,6 +15,8 @@ import (
 	"strings"
 	"time"
 
+	export "go.opentelemetry.io/otel/sdk/export/trace"
+
 	"code.cloudfoundry.org/clock"
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse"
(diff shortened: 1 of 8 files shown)


atc/atccmd/command.go
```diff
@@ -445,22 +447,21 @@ func (cmd *RunCommand) Runner(positionalArguments []string) (ifrit.Runner, error
...
+		exp, err = cmd.Tracing.Stackdriver.Exporter()
+	}
+	if err != nil {
+		return nil, err
+	}
+	tp, err := tracing.TraceProvider(exp)
+	if err != nil {
+		return nil, fmt.Errorf("failed to configure trace provider: %w", err)
 	}
+	tracing.ConfigureTraceProvider(tp)
```
aoldershaw, line 464: I'm not quite getting this change - assuming `tracing.TraceProvider` is fine with having a `nil` `export.SpanSyncer`, are we not still always setting `tracing.Configured` to `true` (even when neither Jaeger nor Stackdriver are configured)?

atc/engine/engine.go
```diff
@@ -22,7 +22,7 @@ import (
 
 type Engine interface {
 	NewBuild(db.Build) Runnable
-	NewCheck(db.Check) Runnable
+	NewCheck(context.Context, db.Check) Runnable
```
aoldershaw, line 25: I recognize this would result in a slightly bigger refactor, but is there a reason we pass the `Context` in here as opposed to modifying `Runnable` interface to take in the `Context`? To quote the [godocs](https://godoc.org/context) for the `context` package:

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.

Correct me if I'm wrong, but we always seem to call `NewBuild` and `NewCheck` immediately adjacent to `Run`, so it shouldn't matter which function we pass it into (in terms of what context we have available)

vito:
Let's resubmit this as the right person. :P

PR #5666, opened 2020-05-26
atc: structure: check step runs in the same context as checker

### LINKED ISSUE #5654: check step should run in the same span context as the checker

jamieklassen:

Refactor - closes #5654.

Some interesting refactors around the tracing package allowed us to make more effective use of the testtrace package from opentelemetry. Hopefully this will assist tracing-related testing efforts in the future.

Having the correct context available for the check step should unblock efforts to propagate span context into check containers (#5423) or resource versions (#5602).

## Notes to reviewer:

Commit messages are fairly thorough.

## Contributor Checklist

- [x] [Signed] all commits
- [x] ~Updated [Release notes]~

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/engine/engine.go
+++ b/atc/engine/engine.go
@@ -22,7 +22,7 @@ import (
 
 type Engine interface {
 	NewBuild(db.Build) Runnable
-	NewCheck(db.Check) Runnable
+	NewCheck(context.Context, db.Check) Runnable
 	ReleaseAll(lager.Logger)
 }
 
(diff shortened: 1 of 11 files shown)


atc/engine/engine.go
```diff
@@ -22,7 +22,7 @@ import (
 
 type Engine interface {
 	NewBuild(db.Build) Runnable
-	NewCheck(db.Check) Runnable
+	NewCheck(context.Context, db.Check) Runnable
```
aoldershaw, line 25: *Reposting from old PR*

---

I recognize this would result in a slightly bigger refactor, but is there a reason we pass the `Context` in here as opposed to modifying `Runnable` interface to take in the `Context`? To quote the [godocs](https://godoc.org/context) for the `context` package:

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.

Correct me if I'm wrong, but we always seem to call `NewBuild` and `NewCheck` immediately adjacent to `Run`, so it shouldn't matter which function we pass it into (in terms of what context we have available)

Issue #5360, opened 2020-03-24
v6.0 feedback


vito:
v6.0 was a big release with significant changes to pipeline semantics - something we rarely do. We'll be taking a bit of time after the release to refactor code and listen to feedback left on this issue before moving on to the next "big thing."

If your workflow was affected by v6.0, whether it's in a good way or a bad way, please leave a comment here! If this issue is closed before then, please just open a new issue or perhaps open a thread in the [forums](https://discuss.concourse-ci.org).

Thanks!

analytically:
Upgrading (from 5.8.0) went flawlessly, with one minor caveat: the order of build steps changed on the dashboard. While the pipeline still displays the correct order, the dashboard turns this around (sorting alphabetically?). Eg:

![image](https://user-images.githubusercontent.com/254434/77576187-8226be80-6eb3-11ea-958c-43bbfd0e47e6.png)

vs 

![image](https://user-images.githubusercontent.com/254434/77576258-ad111280-6eb3-11ea-9286-7955e1feb753.png)

vito:
@analytically FWIW, the dashboard pipeline grid and the pipeline UI have always been out-of-sync, and probably will remain that way until we [re-write the pipeline UI in Elm](https://project.concourse-ci.org/projects/MDc6UHJvamVjdDI5OTk4MjA=).

scuba10steve:
> ... We never got around to it because most of the motivation for it came from PR flows, which was (and still is) a pretty broken usage pattern which we were trying to address directly rather than have it act as a primary motivator for design decisions like this one.

I'm pretty curious if someone would be able to elaborate on this a little bit.  The reason I'm asking is we've began looking into potentially using concourse as an additional CI resource, but also have the capability of building pull requests through it to automatically run unit tests, compile code, etc.

I'm just trying to distinguish where the 'usage' part comes into play.

drahnr:
Upgrade from 5.8.0 to 6.0.0 went flawlessly :+1: , CPU usage dropped significantly (yay!), :zap: https://github.com/concourse/concourse/issues/5334 is still causing issues.

vito:
@scuba10steve Today most folks use the [`telia-oss/github-pr` resource](https://github.com/telia-oss/github-pr-resource), which kind of works but has a bunch of subtle problems in practice. Not the fault of the resource type at all - it's just a square peg trying to fit into a round hole.

[...]

PR #5676, opened 2020-05-27
Add a max in flight to resource checking

### LINKED ISSUE #4624: Investigate implications of "burst checking" behavior
LIDAR kicks off all the queued `check`s at once, which is different from Radar where each resource had its own interval. Theoretically this would mean spikier metrics.

What is the impact of this? Can/should we do something to mitigate it?

clarafu:
Adding a max in flight to resource checking will allow for less spiky behaviour during resource checking intervals. 

This is sort of a temporary fix, where the underlying problem of how the checker is structured will result in spiky behaviour due to the fact that it grabs all the started checks from the database and schedules them all at once. 

This spiky behaviour is tuned down by adding a max in flight, where there can only be a maximum number of checks that happen at once rather than an infinite amount. Having no upper bound to the number of checks that can happen at once can cause problems in large environments where there are thousands of resources, where users will see huge spikes in load. 

closes #4624.

Adds a max in flight to the number of resource checks that happen at once. This number is configurable but I have yet to find a sane default. If it is too low, users will complain about their resource checks being slow. But if it is too high then it will not help mitigate the load balancing.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -141,10 +141,12 @@ type RunCommand struct {
 
 	EnableGlobalResources bool `long:"enable-global-resources" description:"Enable equivalent resources across pipelines and teams to share a single version history."`
 
+	ComponentRunnerInterval time.Duration `long:"component-runner-interval" default:"10s" description:"Interval on which runners are kicked off for builds, locks, scans, and checks"`
+
 	LidarScannerInterval time.Duration `long:"lidar-scanner-interval" default:"1m" description:"Interval on which the resource scanner will run to see if new checks need to be scheduled"`
 	LidarCheckerInterval time.Duration `long:"lidar-checker-interval" default:"10s" description:"Interval on which the resource checker runs any scheduled checks"`
 
-	ComponentRunnerInterval time.Duration `long:"component-runner-interval" default:"10s" description:"Interval on which runners are kicked off for builds, locks, scans, and checks"`
+	ResourceCheckingMaxInFlight uint64 `long:"resource-checking-max-in-flight" default:"32" description:"Maximum number of resources to be checked at the same time"`
 
 	GlobalResourceCheckTimeout          time.Duration `long:"global-resource-check-timeout" default:"1h" description:"Time limit on checking for new versions of resources."`
 	ResourceCheckingInterval            time.Duration `long:"resource-checking-interval" default:"1m" description:"Interval on which to check for new versions of resources."`
@@ -952,6 +954,7 @@ func (cmd *RunCommand) constructBackendMembers(
 				logger.Session(atc.ComponentLidarChecker),
 				dbCheckFactory,
 				engine,
+				cmd.ResourceCheckingMaxInFlight,
 			),
 			cmd.ComponentRunnerInterval,
 			bus,
--- a/atc/lidar/checker.go
+++ b/atc/lidar/checker.go
@@ -16,12 +16,15 @@ func NewChecker(
 	logger lager.Logger,
 	checkFactory db.CheckFactory,
 	engine engine.Engine,
+	resourceCheckingMaxInFlight uint64,
 ) *checker {
+	newGuardResourceChecking := make(chan struct{}, resourceCheckingMaxInFlight)
 	return &checker{
-		logger:       logger,
-		checkFactory: checkFactory,
-		engine:       engine,
-		running:      &sync.Map{},
+		logger:                logger,
+		checkFactory:          checkFactory,
+		engine:                engine,
+		running:               &sync.Map{},
+		guardResourceChecking: newGuardResourceChecking,
 	}
 }
 
@@ -31,7 +34,8 @@ type checker struct {
 	checkFactory db.CheckFactory
 	engine       engine.Engine
 
-	running *sync.Map
+	guardResourceChecking chan struct{}
+	running               *sync.Map
 }
 
 func (c *checker) Run(ctx context.Context) error {
@@ -48,7 +52,13 @@ func (c *checker) Run(ctx context.Context) error {
 
 	for _, ck := range checks {
 		if _, exists := c.running.LoadOrStore(ck.ID(), true); !exists {
+			c.guardResourceChecking <- struct{}{}
+
 			go func(check db.Check) {
+				defer func() {
+					<-c.guardResourceChecking
+				}()
+
 				_, span := tracing.StartSpanFollowing(
 					check,
 					"checker.Run",
--- a/atc/lidar/checker_test.go
+++ b/atc/lidar/checker_test.go
@@ -40,6 +40,7 @@ var _ = Describe("Checker", func() {
 			logger,
 			fakeCheckFactory,
 			fakeEngine,
+			10,
 		)
 	})
 


jamieklassen:
I have been reviewing this PR and I haven't totally concluded what I think about it yet. Here is my investigation so far:

When @clarafu mentioned

> My main concern of having a default be too low is the time it will take to finish all the checks

My mind went to this: can I queue up a bunch of slow resource checks, then set a new pipeline and be forced to wait a long time before the newly-created resources get checked? The use-case i had in mind was the poor performance of the `create-and-run-new-pipeline` SLI: #5547. So I did a simulation, too.

## Preliminaries

I added a new docker-compose override called `registry.yml`:

```yaml
version: '3'
services:
  registry:
    image: registry
    ports:
    - 5000:5000
    restart: always
```

and a really basic custom resource type that will check slowly; `Dockerfile.sleep-resource`:

```dockerfile
FROM alpine
RUN mkdir -p /opt/resource && \
  echo -e '#!/bin/sh\nsleep 10\necho "[]"' > /opt/resource/check && \
  chmod +x /opt/resource/check
```

At the beginning of each test cycle I would `docker build . -f Dockerfile.sleep-resource -t localhost:5000/sleep-resource && docker push localhost:5000/sleep-resource` so that this custom resource would be accessible by concourse.

I used this `pipeline.yml`:

```yaml
resource_types:
- name: sleep-resource
  type: docker-image
  source:
    repository: "registry:5000/sleep-resource"
    insecure_registries:
    - registry:5000
resources:
- name: sleep
  type: sleep-resource
jobs:
- name: get
  plan:
  - get: sleep
```

I would `docker build . -t concourse/concourse:local && docker-compose -f docker-compose.yml -f registry.yml -f hack/overrides/jaeger.yml up -d`, then I would set and unpause the above pipeline with the name `pipeline`.

## Procedure

I saved the classic booklit pipeline in `booklit.yml` and my test was this:

```
for i in {1..10}; do fly -t dev cr -a --shallow -r pipeline/sleep; done && fly -t dev sp -p booklit -c booklit.yml -n && fly -t dev up -p booklit
```

Then I would examine the trace output in jaeger.

## Can slow checks + max-in-flight = reduced pipeline performance? YES

Before this change, the booklit check is run after 34 ms and finishes in ~12s:

Screen Shot 2020-05-29 at 5 29 33 PM

With this change and setting the max-in-flight to 1 (effectively serializing all resource checks), we see that check sitting around for ~100s while the other 10 checks (which take 10s each) happen until it gets run:

Screen Shot 2020-05-29 at 4 23 52 PM

This could be a bad thing when it occurs at scale -- as it stands, Concourse doesn't do much to show what is going on - the web UI will show that the resource is not checking, and I don't think you can even force a check, since that simply creates a db.Check which the checker will still have to get to eventually.

jamieklassen:
## Backwards compatibility

Having done the above experiment, I thought "well that's not great, but this is a 'temporary' or 'experimental' approach. can we maintain the old behaviour if we want?" The old behaviour was an unlimited number of checks in flight. I blindly tried to see if I could recreate that behaviour by setting the max-in-flight to zero.

Instead this resulted in the checker getting completely locked up - the first time the componentrunner executes it acquires a lock for the resource checker components, and that lock never gets released. That's because there ends up being a single goroutine stuck waiting forever to receive from a 0-size channel:

```
goroutine 183 [chan send, 2 minutes]: # 2 minutes will grow forever
github.com/concourse/concourse/atc/lidar.(*checker).Run(0xc00099bec0, 0x3e53fc0, 0xc000478040, 0x0, 0x0)
	/src/atc/lidar/checker.go:55 +0x441
github.com/concourse/concourse/atc/lidar.(*intervalRunner).run(0xc0003bb480, 0x3e53fc0, 0xc000478040, 0x0, 0x0, 0x0)
	/src/atc/lidar/interval_runner.go:130 +0x930
github.com/concourse/concourse/atc/lidar.(*intervalRunner).Run(0xc0003bb480, 0xc00010f740, 0xc00010f7a0, 0x0, 0x0)
	/src/atc/lidar/interval_runner.go:82 +0x524
github.com/tedsuo/ifrit.(*process).run(0xc0003bd400)
	/go/pkg/mod/github.com/tedsuo/ifrit@v0.0.0-20180802180643-bea94bb476cc/process.go:71 +0x6f
created by github.com/tedsuo/ifrit.Background
	/go/pkg/mod/github.com/tedsuo/ifrit@v0.0.0-20180802180643-bea94bb476cc/process.go:49 +0x64
```

so no resource checks will ever run!

I think we should probably be backwards-compatible by default if we're not sure about the big impacts of this.

jamieklassen:
## Multiple web nodes?

I didn't test with multiple web nodes, but based on the way the componentrunner works I have a bad feeling that this change will mean that there is a hard limit on the number of checks being run across the whole cluster, so that resource checking cannot meaningfully be parallelized across web nodes.

I don't feel quite right about this reasoning:

> since the job scheduling max in flight is 32 it seems to be alright to make the checks be 32 too

In my mind the scheduler is the thing that solves constraints and involves extensive access to the database, but the build tracker is the thing that involves extensive access to the worker.

## Synchronizing access to a bottleneck

In LIDAR, the scanner takes the place of the scheduler -- admittedly deciding how to insert a check in the database is not as DB-intensive as deciding the inputs to a pending build -- and the checker takes the place of the build tracker. Honestly maybe we should bring these names together so that we have a check scheduler and check tracker in lidar, and a build scheduler and build tracker elsewhere.

In any case, the reason for setting a max in flight on checks is not to synchronize access to the database (this was necessary for the scheduler), but rather to synchronize access to the _worker_ -- that's where checks get expensive. So the feature we want here is probably some kind of parallel to the `limit-active-tasks` placement strategy for build containers.

clarafu:
@jamieklassen That's a really good point that I should make it backwards compatible if we are to add this change, I'll go ahead and add the ability for resource checking max in flight to be able to take in `0` for the value of turned off!

I think that was definitely my similar concern about adding a max in flight, where if we have the max in flight set to 32 and there are 32 resources that are taking a long time to check, it will cause other checks to not be able to run. I think my assumption with adding a max in flight is that checks should be fairly lightweight and estimating each check should run in `100ms` or so for the average deployment. For slower or larger deployments, I would definitely advise pushing the max in flight to a higher value to allow for more breathing room for slow checks. This also is balanced by the number of resources that they have, which is why picking a default is a really hard task. The default I picked is definitely not optimized for a large deployment (such as hush-house), and is more for a simple averagely sized deployment. But maybe that is not the right thing to do?

For your comment on the most important part being the load on the workers, I also agree with this. I guess this max in flight is kinda a easy experimental bandaid to put on lidar, where a more solid solution might be to have a max number of check containers per worker? So I've been playing with the idea that in the long term I would like us not need to have any kind of hard limit on anything, and have the distributing of checks be more gradual and spread out over time rather than them only being able to happen all at once at these long intervals (1 minute because of the current lidar scanner interval). That will help remove the spikiness of resource checks and hopefully speed up the time it takes for a check to happen because the user will no longer need to wait for the check to be picked up by a specific interval of the scanner/checker. But on the other hand I can see us going in the direction of limiting the max number of check containers per worker, and possibly surface something to the user if their resource check is waiting for a worker to free up if the max in flight for the worker is hit.

vito:
A couple quick points:

* A per-worker max-in-flight will help avoid overloading workers, but it won't help reduce load on the external services being checked, which was one of the original problems reported by a user after upgrading to v6.0: https://github.com/concourse/concourse/issues/5360#issuecomment-609424802
* @jamieklassen brings up a good point that if this is enforced by max-in-flight, we'll be bottlenecked by the performance of the slowest `check`s. This makes it kind of hard to come up with a good number.

Maybe instead of having a hard max-in-flight, we could rate limit the number of checks we start in a given timeframe? That way we don't have to be dependent on `check`s finishing before moving on. This might still end up being a knob to tune with your deployment, but at least the downside of not tuning it isn't as bad. 🤔 It'll just move through a larger queue at the same rate, rather than halting all checking once you hit a certain threshold of "bad apples".

clarafu:
We have been playing around with rate limiting the number of checks rather than a hard max in flight. The way that we calculate the number that we should rate limit by is using a few different factors such as the checker interval, resource checking interval and the number of resource config scopes (checkable items) in the database. Here is a snippet of the logic we are currently using
```
func (c CheckRateCalculator) RateLimit(queueSize int) (rate.Limit, error) {
	var rateOfChecks rate.Limit
	if c.MaxChecksPerSecond == -1 {
		// UNLIMITED POWER
		rateOfChecks = rate.Inf
	} else if c.MaxChecksPerSecond == 0 {
		// The rate needed to get through the check queue within the configured
		// checker interval
		queueRate := float64(rate.Every(c.CheckerInterval / time.Duration(queueSize)))

		// Choose the minimum between the calculated rate to get through all the
		// checks in one interval and the maximum checks that can run per second
		checkableCount, err := c.CheckableCounter.CheckableCount()
		if err != nil {
			return 0, err
		}

		everythingRate := float64(checkableCount) / c.ResourceCheckingInterval.Seconds()

		rateOfChecks = rate.Limit(math.Min(queueRate, everythingRate))
	} else {
		rateOfChecks = rate.Limit(c.MaxChecksPerSecond)
	}

	return rateOfChecks, nil
}
```

We tried running the lidar simulation against this with 100 resources, and we saw the following metrics come up.

Screen Shot 2020-06-08 at 3 16 34 PM

It seemed to have distributed the checks across more checker intervals, approximately 16 checks per 10 seconds (the checker runs on a 10 second interval). The thing we are wondering now is if the `queueRate` is even necessary. If we want to flatten the curve of checks, the `everythingRate` should be the ideal number of checks we want to run in order to check all the resources in the deployment within the default resource checking interval. Maybe we should always be limiting by the `everythingRate`? This might cause some slowness especially at the beginning when it needs to redistribute the checks to other intervals? 

I also have not tested this at scale in a real environment, so I'm not sure the impacts of limiting the rate at which we start checks. There might be a concern of slowness with the checks.

clarafu:
We decided to remove the `queueRate` in favour of always enforcing the `everythingRate` because ideally, we want the checks to happen at the `everythingRate` so that the checks will be evenly spread out through the  resource checking interval. After removing the `queueRate` and allowing it to run for a while in the lidar simulation, we saw that the checks further spread out within the metrics.

Screen Shot 2020-06-08 at 5 30 50 PM

PR #5688, opened 2020-05-29
Release notes 6.2.x

taylorsilva:
Adding release notes to 6.2.x release branch


release-notes/latest.md
```diff
@@ -69,3 +69,23 @@ Currently the only API action that can be limited in this way is `ListAllJobs` -
...
+
+* Added a new Wrappa that returns HTTP 409 for endpoints listed in concourse/rfc#33 when the requested pipeine is archived
+
+#### <sub><sup><a name="5576" href="#5576">:link:</a></sup></sub> fix
+
+* @dcsg added support for AWS SSM for `var_sources`
+
+#### <sub><sup><a name="5575" href="#5575">:link:</a></sup></sub> feature
+
+* Added tracing to the lidar component, where a single trace will be emitted for each run of the scanner and the consequential checking that happens from the checker. The traces will allow for more in depth monitoring of resource checking through describing how long each resource is taking to scan and check. #5575
```
matthewpereira, line 87: It looks like this is the only item that has the link (#5575) tacked on to the end. We should take that out, since the heading will link to the issue already.

It might be easier to understand if we split it into multiple bullet points each expressing one idea, and link to lidar for those unfamiliar with it, like this:

```
#### <sub><sup><a name="5575" href="#5575">:link:</a></sup></sub> feature
* Added tracing to the [lidar component](https://github.com/concourse/concourse/tree/master/atc/lidar)
* A single trace will be emitted for each run of the scanner and the consequential checking that happens from the checker. 
* By describing how long each resource is taking to scan and check, these traces allow for more in depth monitoring of resource checking.
```

release-notes/latest.md
```diff
@@ -69,3 +69,23 @@ Currently the only API action that can be limited in this way is `ListAllJobs` -
...
 
 * Fixed a bug where fly would no longer tell you if the team you logged in with was invalid
+
+#### <sub><sup><a name="5192" href="#5192">:link:</a></sup></sub> fix
+
+* @evanchaoli changed the behaviour of the web to retry individual build steps that fail when a worker disappears.
+
+#### <sub><sup><a name="5549" href="#5549">:link:</a></sup></sub> fix
+
+* Added a new Wrappa that returns HTTP 409 for endpoints listed in concourse/rfc#33 when the requested pipeine is archived
```
matthewpereira, line 79: * If the `wrappa` part of the codebase is a wrapper for Rata, consider just calling it a 'wrapper' here so that people unfamiliar with the joke will get what we're talking about in this context. 😄

release-notes/latest.md
taylorsilva: Looking at past releases it seems like we usually stick the issue number at the end, otherwise a link to the issued doesn't appear anywhere
https://github.com/concourse/concourse/releases

taylorsilva: I'm going to add the issue numbers to the end of each note.

PR #5704, opened 2020-06-01
containerd: run task with the specified user

### LINKED ISSUE #5144: containerd: handle `user` field in proc spec
### what challange are you facing

it's currently not possible for someone to have a process run inside a container
with a particular user set by default:

https://github.com/concourse/concourse/blob/54991b2f1f17bd7d229a3ebbe6d6c16f3ff495a3/worker/backend/container.go#L290-L292

at no moment we look at garden's `User` property to set that user in
`containerd`'s process spec.

without this, one can't make use of `task.run.user`:

https://concourse-ci.org/tasks.html#task-run-user

have the `user` field of the OCI spec set:

> POSIX-platform User
>
> For POSIX platforms the user structure has the following fields:
>
>   uid (int, REQUIRED) specifies the user ID in the container namespace.
>   gid (int, REQUIRED) specifies the group ID in the container namespace.

https://github.com/opencontainers/runtime-spec/blob/39c287c415bf86fb5b7506528d471db5405f8ca8/config.md#user

ps.: it *might* also be important to have `USER`, `HOME`, etc environment
variables flag set, just how [`login(1)`] would. However, I'm not sure if that's
guardian's behavior - as we're in the phase of establishing a baseline, I'd
prefer going down the "follow guardian", at least for now.

[`login(1)`]: http://man7.org/linux/man-pages/man1/login.1.html

muntac:

closes #5144.

Look up user info from `/etc/passwd`, and plug it into the OCI spec. This is because the OCI spec requires the `uid` and `gid` rather than just the user name.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] ~Updated [Release notes]~

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/worker/runtime/container.go
+++ b/worker/runtime/container.go
@@ -15,6 +15,14 @@ import (
 
 const GraceTimeKey = "garden.grace-time"
 
+type UserNotFoundError struct {
+	User string
+}
+
+func (u UserNotFoundError) Error() string {
+	return fmt.Sprintf("user '%s' not found", u.User)
+}
+
 type Container struct {
 	container     containerd.Container
 	killer        Killer
@@ -75,8 +83,10 @@ func (c *Container) Run(
 		return nil, fmt.Errorf("container spec: %w", err)
 	}
 
-	procSpec := containerSpec.Process
-	setupContainerdProcSpec(spec, procSpec)
+	procSpec, err := c.setupContainerdProcSpec(spec, *containerSpec)
+	if err != nil {
+		return nil, err
+	}
 
 	err = c.rootfsManager.SetupCwd(containerSpec.Root.Path, procSpec.Cwd)
 	if err != nil {
@@ -91,7 +101,7 @@ func (c *Container) Run(
 	id := procID(spec)
 	cioOpts := containerdCIO(processIO, spec.TTY != nil)
 
-	proc, err := task.Exec(ctx, id, procSpec, cio.NewCreator(cioOpts...))
+	proc, err := task.Exec(ctx, id, &procSpec, cio.NewCreator(cioOpts...))
 	if err != nil {
 		return nil, fmt.Errorf("task exec: %w", err)
 	}
@@ -325,7 +335,9 @@ func procID(gdnProcSpec garden.ProcessSpec) string {
 	return id
 }
 
-func setupContainerdProcSpec(gdnProcSpec garden.ProcessSpec, procSpec *specs.Process) {
+func (c *Container) setupContainerdProcSpec(gdnProcSpec garden.ProcessSpec, containerSpec specs.Spec) (specs.Process, error) {
+	procSpec := containerSpec.Process
+
 	procSpec.Args = append([]string{gdnProcSpec.Path}, gdnProcSpec.Args...)
 	procSpec.Env = append(procSpec.Env, gdnProcSpec.Env...)
 
@@ -346,6 +358,20 @@ func setupContainerdProcSpec(gdnProcSpec garden.ProcessSpec, procSpec *specs.Pro
 			}
 		}
 	}
+
+
+	if gdnProcSpec.User != "" {
+		var ok bool
+		var err error
+		procSpec.User, ok, err = c.rootfsManager.LookupUser(containerSpec.Root.Path, gdnProcSpec.User)
+		if err != nil {
+			return specs.Process{}, fmt.Errorf("lookup user: %w", err)
+		}
+		if !ok {
+			return specs.Process{}, UserNotFoundError{User: gdnProcSpec.User}
+		}
+	}
+	return *procSpec, nil
 }
 
 func containerdCIO(gdnProcIO garden.ProcessIO, tty bool) []cio.Opt {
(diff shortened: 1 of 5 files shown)


worker/runtime/container.go
```diff
@@ -325,7 +335,9 @@ func procID(gdnProcSpec garden.ProcessSpec) string {
 	return id
 }
 
-func setupContainerdProcSpec(gdnProcSpec garden.ProcessSpec, procSpec *specs.Process) {
+func (c *Container) setupContainerdProcSpec(gdnProcSpec garden.ProcessSpec, containerSpec specs.Spec) (specs.Process, error) {
```
xtreme-sameer-vohra, line 338: Since this method doesn't rely on the `*Container`, do we want to want add behaviour on `Container` as opposed to keeping this a `static` method ?

worker/runtime/rootfs_manager.go
```diff
@@ -1,22 +1,31 @@
...
-// RootfsManager is responsible for mutating the rootfs of a container.
+// RootfsManager is responsible for mutating and reading from the rootfs of a
+// container.
 //
 type RootfsManager interface {
 	// SetupCwd mutates the root filesystem to guarantee the presence of a
 	// directory to be used as `cwd`.
 	//
-	SetupCwd(containerSpec *specs.Spec, cwd string) (err error)
+	SetupCwd(rootfsPath string, cwd string) (err error)
```
xtreme-sameer-vohra, line 23: 👍

worker/runtime/rootfs_manager_test.go
```diff
@@ -81,9 +72,90 @@ func (s *RootfsManagerSuite) TestSetupCwdWithoutIDMappings() {
...
+
+func (s *RootfsManagerSuite) TestLookupUserInvalidUID() {
+	mgr := runtime.NewRootfsManager()
+
+	s.writeEtcPasswd(`
+		some_user_name:*:NaN:0:System Administrator:/var/root:/bin/sh
+	`)
+
+	_, _, err := mgr.LookupUser(s.rootfsPath, "some_user_name")
+	s.Error(err)
```
xtreme-sameer-vohra, line 120: It would be clearer to assert not only an error was returned but it matched a type or error message "invalid uid....."
Same for the other two cases below

xtreme-sameer-vohra:
Acceptance testing

- [ ] defaulting to root
- [ ] specifying another user (eg. `nobody` in `busybox` )
- [ ] specifying unknown user ( eg. `foo` in `busybox` )
- [ ] checking env variables between garden vs containerd with a specified user (eg. `nobody` in `busybox` )

xtreme-sameer-vohra:
> Acceptance testing
> 
> * [x]  defaulting to root
> * [x]  specifying another user (eg. `nobody` in `busybox` )
> * [x]  specifying unknown user ( eg. `foo` in `busybox` )
> * [ ]  checking env variables between garden vs containerd with a specified user (eg. `nobody` in `busybox` )

For the last case, as [mentioned previously](https://github.com/concourse/concourse/issues/5144#issuecomment-634138837) Garden sets the $USER env variable, `containerd` doesn't set it. To maintain parity with `garden` and [man login](https://man7.org/linux/man-pages/man1/login.1.html) we should set it.

`env` in Garden
```
USER=nobody
SHLVL=1
HOME=/home
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/tmp/build/b1ade531
```

`env` in Containerd
```
SHLVL=1
HOME=/home
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PWD=/tmp/build/e8b2da71
```

worker/runtime/container.go
```diff
@@ -15,6 +15,14 @@ import (
 
 const GraceTimeKey = "garden.grace-time"
 
+type UserNotFoundError struct {
+	User string
+}
+
+func (u UserNotFoundError) Error() string {
+	return fmt.Sprintf("user '%s' not found", u.User)
```
xtreme-sameer-vohra, line 23: Gaurdian has the following error which might be more descriptive in helping the user rectify the issue
```
unable to find user SOME_MISSING_USER: no matching entries in passwd file
```

muntac: This error is thrown when we call `RootFSManager.LookupUser` and no user is found. The `LookupUser` looks up an user using the `etc/passwd`. However, we thought the Concourse UI facing user does not care about the details of how we find the user, and only that the user could not be found.

We can change it if you feel strongly about it.

worker/runtime/container.go
muntac: The method now uses Container's RootFSManager [here](https://github.com/concourse/concourse/blob/ac548408ddf8f7f0777eed85b78bca75f2875aab/worker/runtime/container.go#L366). So we thought it would be best to make it a method of `*Container`.

Do you feel it would be better to pass the `c.RootFSManager` in the signature?

worker/runtime/rootfs_manager_test.go
muntac: Good point! Done.

worker/runtime/container.go
muntac: Added details to message.

PR #5724, opened 2020-06-08
Check if version exists before using it as a candidate

### LINKED ISSUE #5655: Inputs algorithm should avoid choosing versions that are no longer available
## Summary

It's possible for the new algorithm to decide on a version which no longer exists, because `passed` constraints only reference version MD5s from build outputs.

To match the previous behavior, if we choose a version that no longer exists, we should skip it and move on to an older build.

It's probably a good idea to think about the performance implications of this, though any slowness should be greatly mitigated by the fact that we only schedule jobs which need to be scheduled anyway (as of v6.0).

## Context

* Prior discussion: https://github.com/concourse/concourse/issues/5477#issuecomment-633666451

clarafu:

It increases stability of the algorithm when trying to compute a version for it's inputs. When we determine a set of versions within the algorithm, we should only be using versions that exists in the `resource_config_versions` table. For the pinned and individual resolvers, we compute our version for the input using the `resource_config_versions` table so we will only compute a version that already exists. For the group resolver, we use the successful build outputs table so we might end up computing a version that exists in the `successful_build_outputs` but not in the `resource_config_versions` table.

closes #5655 

Before we set a version as a candidate for an input, we will do a check to see if that version exists in the `resource_config_versions` table. This will ensure that we only set a version as a candidate if it exists.

## Notes to reviewer:
The number of extra queries to the database that this change adds depending on the run of the algorithm. For example, if the algorithm is determining versions for a set of three inputs, if the versions that it chose the first time are all viable then this query will only be run three times (once for each input). But if the algorithm has to run through multiple builds which results in it setting multiple versions as a candidate for the input, that can result in us running this query for however many times we set the version as a candidate. The point I'm trying to make is that this change has the possibility of adding on to the load of the algorithm, and the scale of it depends on the amount of times that it needs to reset the version on a candidate. That being said, the query itself should be fairly lightweight. When we ran the query to check if a version exists on a large deployment, it ran in 0.05 ms. The following is the result of the explain analyze.
```
Result  (cost=17.04..17.05 rows=1 width=1) (actual time=0.033..0.033 rows=1 loops=1)
  InitPlan 1 (returns $1)
    ->  Nested Loop  (cost=0.96..17.04 rows=1 width=0) (actual time=0.033..0.033 rows=1 loops=1)
          ->  Index Scan using resources_pkey on resources r  (cost=0.41..8.46 rows=1 width=4) (actual time=0.013..0.013 rows=1 loops=1)
                Index Cond: (id = 79654)
          ->  Index Only Scan using resource_config_scope_id_and_version_md5_unique on resource_config_versions v  (cost=0.55..8.57 rows=1 width=4) (actual time=0.018..0.018 rows=1 loops=1)
                Index Cond: ((resource_config_scope_id = r.resource_config_scope_id) AND (version_md5 = '3895ed6ba0e0a864a2bc7eed3c72ad3e'::text))
                Heap Fetches: 0
Planning time: 0.379 ms
Execution time: 0.054 ms
```

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/scheduler/algorithm/group_resolver.go
+++ b/atc/scheduler/algorithm/group_resolver.go
@@ -256,6 +256,18 @@ outputs:
 				continue
 			}
 
+			if candidate == nil {
+				exists, err := r.vdb.VersionExists(ctx, output.ResourceID, output.Version)
+				if err != nil {
+					tracing.End(span, err)
+					return false, err
+				}
+
+				if !exists {
+					break outputs
+				}
+			}
+
 			// if this doesn't work out, restore it to either nil or the
 			// candidate *without* the job vouching for it
 			restore[c] = candidate
--- a/atc/db/versions_db.go
+++ b/atc/db/versions_db.go
@@ -212,6 +212,24 @@ func (versions VersionsDB) SuccessfulBuildOutputs(ctx context.Context, buildID i
 	return algorithmOutputs, nil
 }
 
+func (versions VersionsDB) VersionExists(ctx context.Context, resourceID int, versionMD5 ResourceVersion) (bool, error) {
+	var exists bool
+	err := versions.conn.QueryRowContext(ctx, `
+		SELECT EXISTS (
+			SELECT 1
+			FROM resource_config_versions v
+			JOIN resources r ON r.resource_config_scope_id = v.resource_config_scope_id
+			WHERE r.id = $1
+			AND v.version_md5 = $2
+		)`, resourceID, versionMD5).
+		Scan(&exists)
+	if err != nil {
+		return false, err
+	}
+
+	return exists, nil
+}
+
 func (versions VersionsDB) FindVersionOfResource(ctx context.Context, resourceID int, v atc.Version) (ResourceVersion, bool, error) {
 	versionJSON, err := json.Marshal(v)
 	if err != nil {
--- a/atc/scheduler/algorithm/algorithm_test.go
+++ b/atc/scheduler/algorithm/algorithm_test.go
@@ -3175,4 +3175,70 @@ var _ = DescribeTable("Input resolving",
 			},
 		},
 	}),
+
+	Entry("if the chosen version for an input with passed constraints does not exist, it will not select that version", Example{
+		DB: DB{
+			BuildInputs: []DBRow{
+				{Job: "another-job", BuildID: 100, Resource: "resource-x", Version: "rxv1", CheckOrder: 1},
+				{Job: "another-job", BuildID: 101, Resource: "resource-x", Version: "rxv2", CheckOrder: 2, DoNotInsertVersion: true},
+			},
+			Resources: []DBRow{
+				{Resource: "resource-x", Version: "rxv1", CheckOrder: 1},
+			},
+		},
+
+		Inputs: Inputs{
+			{
+				Name:     "resource-x",
+				Resource: "resource-x",
+				Passed:   []string{"another-job"},
+			},
+		},
+
+		Result: Result{
+			OK: true,
+			Values: map[string]string{
+				"resource-x": "rxv1",
+			},
+		},
+	}),
+
+	Entry("if there are multiple inputs with the same passed constraint job and the chosen version from a build does not exist, it will not use that build", Example{
+		DB: DB{
+			BuildInputs: []DBRow{
+				{Job: "another-job", BuildID: 100, Resource: "resource-x", Version: "rxv1", CheckOrder: 1},
+				{Job: "another-job", BuildID: 101, Resource: "resource-x", Version: "rxv2", CheckOrder: 2},
+
+				{Job: "another-job", BuildID: 100, Resource: "resource-y", Version: "ryv1", CheckOrder: 1},
+				{Job: "another-job", BuildID: 101, Resource: "resource-y", Version: "ryv2", CheckOrder: 2, DoNotInsertVersion: true},
+			},
+
+			Resources: []DBRow{
+				{Resource: "resource-x", Version: "rxv1", CheckOrder: 1},
+				{Resource: "resource-x", Version: "rxv2", CheckOrder: 2},
+				{Resource: "resource-y", Version: "ryv1", CheckOrder: 1},
+			},
+		},
+
+		Inputs: Inputs{
+			{
+				Name:     "resource-x",
+				Resource: "resource-x",
+				Passed:   []string{"another-job"},
+			},
+			{
+				Name:     "resource-y",
+				Resource: "resource-y",
+				Passed:   []string{"another-job"},
+			},
+		},
+
+		Result: Result{
+			OK: true,
+			Values: map[string]string{
+				"resource-x": "rxv1",
+				"resource-y": "ryv1",
+			},
+		},
+	}),
 )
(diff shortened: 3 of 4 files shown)


atc/scheduler/algorithm/group_resolver.go
```diff
@@ -256,6 +256,18 @@ outputs:
 				continue
 			}
 
+			if candidate == nil {
```
aoldershaw, line 259: What does it mean for `candidate` to be `nil` here? Does it mean there is currently no candidate version for a particular input?

If so, why don't we need to check for existence when `candidate != nil`? Could we not potentially overwrite the existing `candidate` with a version that doesn't exist?

aoldershaw: Looking at it more, I guess the `mismatch` case would apply if `candidate != nil` and the versions differed:

https://github.com/concourse/concourse/blob/e5f48f24ed9e7827547c1b71ee42b4df4523e15e/atc/scheduler/algorithm/group_resolver.go#L250-L253

Could you explain why we need to keep looking through builds if we already have a candidate version, and we don't even consider other versions? Or am I misunderstanding this part entirely?

clarafu: @aoldershaw For your first question, ya it is checking if there is no candidate set already for that version. If the version that we currently have is different than the one in the candidate, then that would be a mismatched version and we would restore that candidate to `nil`.

My original goal for having it only check if `candidate == nil` is so that when we vouch for that candidate with other jobs, we don't need to check if that version exists again. So the way that the algorithm works with passed constraints is that if you have one input with 3 passed constraints (let's say `job-1', `job-2` and `job-3`, it will recurse through the algorithm and "vouch for" the version we have determined to be the candidate with every passed constraint job. The flow is something like this
1. **candidate = nil, vouched for by = nil** : we run through the algorithm for the input and we run `trySatisfyPassedConstraintsForInput` for that input
https://github.com/concourse/concourse/blob/9f3c8970479647764fa1285e4574b08536eb1cbd/atc/scheduler/algorithm/group_resolver.go#L111-L116
then loop over the first passed constraint job `job-1`
https://github.com/concourse/concourse/blob/9f3c8970479647764fa1285e4574b08536eb1cbd/atc/scheduler/algorithm/group_resolver.go#L145
and determine a version as the candidate for that input (lets say the version is v1)
https://github.com/concourse/concourse/blob/9f3c8970479647764fa1285e4574b08536eb1cbd/atc/scheduler/algorithm/group_resolver.go#L270
then recurse
https://github.com/concourse/concourse/blob/9f3c8970479647764fa1285e4574b08536eb1cbd/atc/scheduler/algorithm/group_resolver.go#L282
1. **candidate = `v1`, vouched for by = `job-1`** : now we will `trySatisfyPassedConstraintsForInput` for that same input and this time when we loop over the passed constraint jobs, it will skip over `job-1` because it has already vouched for that version and it will move onto `job-2`
https://github.com/concourse/concourse/blob/9f3c8970479647764fa1285e4574b08536eb1cbd/atc/scheduler/algorithm/group_resolver.go#L148-L151
now `job-2` will then go through the same flow of finding a version for the candidate and if it finds the same version, then it will vouch for the version
https://github.com/concourse/concourse/blob/9f3c8970479647764fa1285e4574b08536eb1cbd/atc/scheduler/algorithm/group_resolver.go#L270
1. **candidate = `v1`, vouched for by = `job-1` and `job-2`** : the steps will be the same for `job-3` where it will skip over `job-1` and `job-2` and run `job-3`. If it produces `v1` then `job-3` will vouch for `v1`.
1. **candidate = `v1`, vouched for by = `job-1`, `job-2` and `job-3`** : we are finished! all the passed jobs have vouched for the version so now we can return out of this recursion. 
https://github.com/concourse/concourse/blob/9f3c8970479647764fa1285e4574b08536eb1cbd/atc/scheduler/algorithm/group_resolver.go#L180-L182

Based off of this, we will end up going over the code for vouching for a candidate version multiple times. I only wanted to check if the version exists if `candidate == nil` so that we only run the query for checking if the version exists once when we first determined a version for the candidate (which would be step 2 in the example above) rather if there was no check to see if candidate was nil then it would run the query every time we vouched for the candidate with all the passed constraint jobs.

clarafu: I think that comment might be a bit misleading?
```
// build contained a different version than the one we already have for 
// that candidate, so let's try a different build 
```

The `build` that it is referring to is a build of a different passed constraint job rather than the build of the passed constraint job that actually set the existing version on the candidate. So it would be when I have an input with two passed constraints, the first passed constraint job set the candidate to version `v1` but then when I recurse within the algorithm for the second passed constraint job, it finds `v2` as the version. This will end up being a mismatch and that will cause the `v1` version to be removed as a candidate.

clarafu:
@aoldershaw Don't worry it is a really confusing part of the codebase! If you have any questions or are really struggling with some parts of the code, just let me know and maybe we can go over it together :)

aoldershaw:
Ran through the repro steps you provided and, sure enough, it seems to work! I added an event to the `tryBuildOutputs` span to get some insight into when the versions get checked (and when they are found to not exist).

```diff
                        if candidate == nil {
                                exists, err := r.vdb.VersionExists(ctx, output.ResourceID, output.Version)
                                if err != nil {
                                        tracing.End(span, err)
                                        return false, err
                                }
+                               span.AddEvent(
+                                       ctx,
+                                       "checked version existence",
+                                       key.New("resourceID").Int(output.ResourceID),
+                                       key.New("version").String(string(output.Version)),
+                                       key.New("exists").Bool(exists),
+                               )

                                if !exists {
                                        break outputs
                                }
                        }
```

Upon performing step 9 (creating the second build for `job-2` with `v2` deleted), I see the following trace:

Screen Shot 2020-06-09 at 8 27 02 PM

followed by:

Screen Shot 2020-06-09 at 8 27 30 PM

Noice!

Still want to take some more time to understand the algorithm/group resolver a bit better. One thing that's confusing me: what's the purpose of steps 1-3? I would've thought we only need two versions:

1. trigger `job-1` with `v1`
2. trigger `job-1` with `v2`
3. delete `v1` from DB
4. trigger `job-2`, should run with `v2` (but first check `v1` and find that it's non-existent)

`job-2` runs with `v2`, but it never even checks `v1` - shouldn't it be seeing it as a candidate version since it's `version: every` and `job-1` `passed` on it?

As I'm writing this, I'm vaguely recalling something I read about `version: every` - something along the lines of the first time it's run, it runs with the current latest version, but uses every version after that initial version. Is that correct?

aoldershaw:
**Alternative repro** (let me know if this isn't quite testing the same thing - it seems to be based on the traces, though)

```yaml
resources:
- name: mock
  type: mock
  source:
    no_initial_version: true

jobs:
- name: job-1
  plan:
  - get: mock

- name: job-2
  plan:
  - get: mock
    # note: no "version: every"
    passed: [job-1]
```

1. `fly -t dev cr -r test/mock -f version:v1` in order to save `v1` as a version
1. create a build of `job-1` using `v1`
1. `fly -t dev cr -r test/mock -f version:v2` in order to save `v2` as a version
1. create a build of `job-1` using `v2`
1. delete `v2` from the database
1. create a build of `job-2` - it should pick up `v1`

PR #5729, opened 2020-06-09
Allow set_pipeline can be used across teams

tlwr:

Feature

Allow "team" parameter to be used for "set_pipeline" steps

- Omitting the parameter uses the current team
- Specifying it targets a particular team
- Specifying it requires the team to be an Admin team (main)

We currently use the [`concourse-pipeline-resource`](https://github.com/concourse/concourse-pipeline-resource) to manage a few pipelines for other teams.

It would be more elegant if this could be done via the `set_pipeline` step. We currently have a local user for each of our tenant teams, for this purpose.

Pertains to https://github.com/concourse/rfcs/pull/31:

- Requested: https://github.com/concourse/rfcs/pull/31#issuecomment-601621064
- Vito comments: https://github.com/concourse/rfcs/pull/31#issuecomment-625273371

I think admin teams (main) be able to set pipelines is reasonable.

### Examples

Set the pipeline `pipeline` of the current team
```
      - set_pipeline: pipeline
        file: pipeline/pipeline.yml
```

Set the pipeline `pipeline` of the team `another-team`
```
      - set_pipeline: pipeline
        team: another-team
        file: pipeline/pipeline.yml
```

This can only be done by the `main` team (or any team with `admin=t` in the database, which is only the `main` team)

## Notes to reviewer:

I have only written unit tests, I have not written an integration test

I have not updated the release notes yet

I have not raised a docs PR yet

## Contributor Checklist

- [x] [Signed] all commits
- [x] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/exec/set_pipeline_step.go
+++ b/atc/exec/set_pipeline_step.go
@@ -130,7 +130,42 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState) error {
 		return nil
 	}
 
-	team := step.teamFactory.GetByID(step.metadata.TeamID)
+	var team db.Team
+	if step.plan.Team == "" {
+		team = step.teamFactory.GetByID(step.metadata.TeamID)
+	} else {
+		currentTeam, found, err := step.teamFactory.FindTeam(step.metadata.TeamName)
+		if err != nil {
+			return err
+		}
+		if !found {
+			return fmt.Errorf("team %s not found", step.metadata.TeamName)
+		}
+
+		targetTeam, found, err := step.teamFactory.FindTeam(step.plan.Team)
+		if err != nil {
+			return err
+		}
+		if !found {
+			return fmt.Errorf("team %s not found", step.plan.Team)
+		}
+
+		permitted := false
+		if targetTeam.ID() == currentTeam.ID() {
+			permitted = true
+		}
+		if currentTeam.Admin() {
+			permitted = true
+		}
+		if !permitted {
+			return fmt.Errorf(
+				"only %s team can set another team's pipeline",
+				atc.DefaultTeamName,
+			)
+		}
+
+		team = targetTeam
+	}
 
 	fromVersion := db.ConfigVersion(0)
 	pipeline, found, err := team.Pipeline(step.plan.Name)
--- a/atc/exec/set_pipeline_step.go
+++ b/atc/exec/set_pipeline_step.go
@@ -134,6 +134,11 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState) error {
 	if step.plan.Team == "" {
 		team = step.teamFactory.GetByID(step.metadata.TeamID)
 	} else {
+		fmt.Fprintln(stderr, "\x1b[1;33mWARNING: specifying the team in a set_pipeline step is experimental and may be removed in the future!\x1b[0m")
+		fmt.Fprintln(stderr, "")
+		fmt.Fprintln(stderr, "\x1b[33mcontribute to discussion #5731 with feedback: https://github.com/concourse/concourse/discussions/5731\x1b[0m")
+		fmt.Fprintln(stderr, "")
+
 		currentTeam, found, err := step.teamFactory.FindTeam(step.metadata.TeamName)
 		if err != nil {
 			return err
--- a/atc/exec/set_pipeline_step_test.go
+++ b/atc/exec/set_pipeline_step_test.go
@@ -94,14 +94,7 @@ jobs:
 
 		credVarsTracker vars.CredVarsTracker
 
-		stepMetadata = exec.StepMetadata{
-			TeamID:       123,
-			TeamName:     "some-team",
-			BuildID:      42,
-			BuildName:    "some-build",
-			PipelineID:   4567,
-			PipelineName: "some-pipeline",
-		}
+		stepMetadata exec.StepMetadata
 
 		stdout, stderr *gbytes.Buffer
 
@@ -135,7 +128,18 @@ jobs:
 		fakeTeam = new(dbfakes.FakeTeam)
 		fakePipeline = new(dbfakes.FakePipeline)
 
-		fakeTeam.NameReturns("some-team")
+		stepMetadata = exec.StepMetadata{
+			TeamID:       123,
+			TeamName:     "some-team",
+			BuildID:      42,
+			BuildName:    "some-build",
+			PipelineID:   4567,
+			PipelineName: "some-pipeline",
+		}
+
+		fakeTeam.IDReturns(stepMetadata.TeamID)
+		fakeTeam.NameReturns(stepMetadata.TeamName)
+
 		fakePipeline.NameReturns("some-pipeline")
 		fakeTeamFactory.GetByIDReturns(fakeTeam)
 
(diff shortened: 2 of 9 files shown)


atc/exec/set_pipeline_step.go
```diff
@@ -105,7 +105,42 @@ func (step *SetPipelineStep) Run(ctx context.Context, state RunState) error {
...
+		}
+
+		permitted := false
+		if targetTeam.ID() == currentTeam.ID() {
+			permitted = true
+		}
+		if currentTeam.Admin() {
+			permitted = true
+		}
+		if !permitted {
```
tlwr, line 135: I think this could be simplified to

```
if !currentTeam.Admin() {
```

and not allow non-admin teams to specify a team

tlwr: Resolved in  ac6d54a

vito:
Thanks for working on this. 🙂 I'm hesitant to merge this though - I think there's more discussion to be had, either here or in RFCs.

As I mentioned in https://github.com/concourse/rfcs/pull/31#issuecomment-625273371 I don't think `set_pipeline` step should concern itself with other teams in the long run - it points to a broader problem to address.

There are pretty active discussions elsewhere in that RFC - namely this thread - https://github.com/concourse/rfcs/pull/31#discussion_r434347135. Right now the discussion is around a hierarchy of projects (which may *replace* teams) rather than cross-team configuration: a root project configures child projects which configure their own pipelines and further projects.

There's also a newly submitted RFC for completely static configuration, which may be of interest: https://github.com/concourse/rfcs/pull/59 - right now it assumes Projects, but in the meantime maybe it could configure Pipelines? Given that this PR is only functional for the admin team, this would be roughly equivalent to what you're trying to do since it's at the operator level - though it won't support configuring from a build plan.

---

All that being said, if this is the last thing that's keeping you (and presumably others) on `concourse-pipeline` resource, I can see value in *temporarily* adding a `team:` field while we figure this out - as long as the step is experimental. I'd like to get everyone off that resource type.

tlwr:
Hi, thank you for your detailed response!

I've been following the conversation in the RFC 31, and people have raised good points. I hope this PR is able to contribute to the discourse.

If RFC 59 supported pipelines, it would be applicable for this use-case, but would require a re-deployment of the web nodes just to change a pipeline according to static configuration. Although RFC 59 is a lovely change for other config, eg user management.

I can confirm, we are currently unable to stop using the `concourse-pipeline-resource` only because of the lack of `team` support.

One of the challenges of using the `concourse-pipeline-resource` is keeping the resource version in-line with the Concourse version, otherwise the resource becomes unreliable.

I completely understand that `main` team setting pipelines for other teams is a niche use-case. I also see this sort of `main` team management as a temporary local maximum until Projects are a thing. The advent of Projects will be a good opportunity for us to reconsider how we are currently using teams, and do a bit of re-tooling to better support our teams.

vito:
@tlwr 

> I completely understand that `main` team setting pipelines for other teams is a niche use-case. I also see this sort of `main` team management as a temporary local maximum until Projects are a thing. The advent of Projects will be a good opportunity for us to reconsider how we are currently using teams, and do a bit of re-tooling to better support our teams.

Happy to hear this, sounds like we're in alignment overall. 🙂 

> If RFC 59 supported pipelines, it would be applicable for this use-case, but would require a re-deployment of the web nodes just to change a pipeline according to static configuration. Although RFC 59 is a lovely change for other config, eg user management.

True - adding pipelines into the mix would probably result in a lot more re-configuration at the operator level, so maybe that's why it's omitted for now.

---

Here's a thought: since we already have the `set_pipeline` step marked "experimental", let's just go ahead and allow these experiments to happen, but let's also clearly document which ones are at risk for removal in the future so that users can expect to adjust. We should also provide an avenue for feedback so that we can know how people are using these experiments.

We could do this by adding a warning message when `team:` is used (in addition to the one that's already there), and include a link to a GitHub Discussion for providing feedback. Does that sound OK?

I've created a Discussion here so you can link to it if so: https://github.com/concourse/concourse/discussions/5731

tlwr:
Awesome

I've pushed up a new commit with the warning:
image

atc/exec/set_pipeline_step.go
```diff
@@ -130,7 +130,40 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState) error {
...
+
+		targetTeam, found, err := step.teamFactory.FindTeam(step.plan.Team)
+		if err != nil {
+			return err
+		}
+		if !found {
+			return fmt.Errorf("team %s not found", step.plan.Team)
+		}
+
+		if !currentTeam.Admin() {
```
evanchaoli, line 158: Only allowing admin user to set pipeline to other team doesn't help much for our clusters. Can we do this? If target team and current team have shared owner/member, then allow to set.

tlwr: Would you be able to confirm what you mean by "shared owner/member"?

I interpret this in two ways:
1. There is one or more shared owners in `team-a` and `team-b`
2. The users and groups of owners in `team-a` and `team-b` are the same

I think (1) is not a good idea, because then anyone owner in `team-a` can set pipelines in `team-b` without necessarily having access to do it themselves.

I think (2) is probably what you meant, so the resultant logic would be 

```
if !currentTeam.Admin() && !currentTeam.Auth().OwnersAreSameAs(targetTeam.Auth()) {
  // error not permitted
}

// continue to set pipeline
```

where `atc.TeamAuth.OwnersAreSameAs` is a new function which would check that the users and groups are the same, disregarding ordering.

---

I'm happy to update my change to include this, though first I'll wait for @vito to weigh in with his opinion

evanchaoli: @tlwr I meant 1. My theory is that, say user-1 and user-2 are owner/member (member role has permission to set pipeline) of team-a; and user-2 and user-3 are owner/member of team-b. Then user-2 can get a pipeline from team-a and set to team-b, though user-1 cannot, but that still implies that team-a can set pipeline to team-b. I agree that opens a hole where user-1 can set-pipeline to team-b via a `set-pipeline` step from team-a.

For 2, I feel that "exact same" hurts usability. For example, team-a has owner user-1, and team-b has owner user-1 and user-2, in this case, I think team-a should still be able to set pipeline to team-b.

So combining 1 and 2, can we do this: if owners/members of team-a is a sub-set of team-b's owners/members, then team-a can set pipeline to team-b. Which is a little bit more flexible than 2.

@vito A real use case, I don't want to mention name of the project here. The project has ~100 concourse teams, one team is pipeline maintainer, rest of teams reflect to modules of the project. Because every module has similar CI/CD workflows, the pipeline maintainer team maintains a set of pipeline templates via vars, and rolls out pipelines to other teams using templates+vars. So the owners of the maintainer team are owners of other teams, but they are not admin users. Now they use the concourse-fly resource type, this feature will make their maintainer pipeline a little bit simpler.

tlwr: @evanchaoli I get you now, I can see how that would work:

`team-a` can set `team-b`'s pipelines if every owner/member of `team-a` is an owner/member in `team-b`

I can update this PR to include this behaviour, but I'll let you and @vito find agreement first :)

vito: @evanchaoli Ah, got it, thanks! I think we have a similar scenario between our team and another team which maintains open-source license scanning pipelines for a bunch of other teams.

Since we don't want `team:` to be here forever, I'd like to plan ahead and see how this use case might be supported in a future without it just to make sure we don't paint ourselves into a corner.

Let's call the team that maintains the pipelines for other teams "upstream" and the teams that consume the pipelines "downstream". Given that the goal is for the upstream team to be able to configure pipelines for the downstream teams, I think this could be modeled as a Project that all the downstream teams configure within their own Project.

So the downstream team's Project config would look something like this:

```yaml
plan:
- set_pipeline: foo
- set_pipeline: bar
- set_project: upstream-baz
```

I'm glossing over a lot of things here, but architecturally this seems sound: it allows the downstream teams to opt-in to the upstream project, and allows the upstream project to self-update when changes are made without requiring any further maintenance from the downstream teams, and without requiring special access to it. In fact they could live on completely separate clusters.

Do you think this kind of flow would make sense for your use case?

tlwr: @vito and @evanchaoli the above discussed behaviour is a superset of what is currently implemented in this PR, so that this doesn't hang around I'm happy to raise a follow up PR implementing whatever behaviour meets the use case

I've documented the behaviour currently implemented in this PR here https://github.com/concourse/docs/pull/341

atc/exec/set_pipeline_step.go
```diff
@@ -130,7 +130,40 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState) error {
 		return nil
 	}
 
-	team := step.teamFactory.GetByID(step.metadata.TeamID)
+	var team db.Team
+	if step.plan.Team == "" {
+		team = step.teamFactory.GetByID(step.metadata.TeamID)
+	} else {
+		fmt.Fprintln(stderr, "\x1b[1;33mWARNING: specifying the team in a set_pipeline step is experimental and subject to change!\x1b[0m")
```
vito, line 137: I think we should use even stronger scare words here:

```suggestion
		fmt.Fprintln(stderr, "\x1b[1;33mWARNING: specifying the team in a set_pipeline step is experimental will likely be removed in the future!\x1b[0m")
```

atc/exec/set_pipeline_step.go
vito: @tlwr Sounds fair! This can be resolved in a later PR.

atc/exec/set_pipeline_step.go
```diff
@@ -130,7 +130,40 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState) error {
...
+		fmt.Fprintln(stderr, "")
+		fmt.Fprintln(stderr, "\x1b[33mcontribute to discussion #5731 with feedback: https://github.com/concourse/concourse/discussions/5731\x1b[0m")
+		fmt.Fprintln(stderr, "")
+
+		currentTeam, found, err := step.teamFactory.FindTeam(step.metadata.TeamName)
```
vito, line 142: To avoid an extra `FindTeam` query here, can we change the original team assignment (using `GetByID`) to `currentTeam` and then do `team = currentTeam` if `step.plan.Team` is not set?

i.e.:

```go
currentTeam := step.teamFactory.GetByID(step.metadata.TeamID)
if step.plan.Team == "" {
  team = currentTeam
} else {
  // ...
}
```

atc/exec/set_pipeline_step_test.go
```diff
@@ -305,6 +309,99 @@ jobs:
...
+
+					stepMetadata.TeamID = fakeUserCurrentTeam.ID()
+					stepMetadata.TeamName = fakeUserCurrentTeam.Name()
+					fakeTeamFactory.FindTeamReturnsOnCall(
+						0,
+						fakeUserCurrentTeam, true, nil,
+					)
+				})
+
+				Context("when team is empty", func() {
```
vito, line 332: nit: this Context sounds like the opposite of the outer Context which is a little confusing

vito:
Also: could a `testflight` test could be added for this?

Here's an example of a test that uses both an admin team and a regular team - I think that pattern could be applied here too:

https://github.com/concourse/concourse/blob/deb97522f480268b5cf9cbd1ef5fcea46b709fdd/testflight/list_builds_test.go#L38

atc/exec/set_pipeline_step.go
tlwr: AFAICT this represents a team, but doesn't make the team functional for use in the following checks

If we use `GetByID` then the `currentTeam.Admin()` call will return false, because it is the zero value of the struct

vito: Blargh, so it's partially instantiated? That's annoying. OK with this as-is then.

atc/exec/set_pipeline_step_test.go
tlwr: Force pushed so that the inner context is more specific "when team is set to the empty string"

atc/exec/set_pipeline_step.go
tlwr: Force pushed so the wording is now as you suggest

tlwr:
Thank you for your testflight suggestion, I have added two testflight tests in the two last commits. The set-pipeline testflight now tests the following:

- a team can use set-pipeline to create a new pipeline (existing)
- the main team can use set-pipeline to create a new pipeline for another team
- a team cannot use set-pipeline to create a new pipeline for the main team

LucaLanziani:
This would be super useful for my team too. It would be amazing to have it merged and released.

atc/exec/set_pipeline_step_test.go
```diff
@@ -305,6 +309,99 @@ jobs:
...
+					)
+				})
+
+				Context("when team is set to the empty string", func() {
+					BeforeEach(func() {
+						fakeTeam.PipelineReturns(fakePipeline, true, nil)
+						fakeTeam.SavePipelineReturns(fakePipeline, false, nil)
+					})
+
+					BeforeEach(func() {
```
vito, line 338: nit: two `BeforeEach`es here

testflight/set-pipeline-step_test.go
```diff
@@ -51,39 +55,56 @@ jobs:
         greetings: hello
 `
 
-	var fixture string
+	const pipelineContentWithTeam = pipelineContent + `
```
vito, line 58: nit: This looks kind of brittle; can we do this with a static `((var))` instead? We can run with `-y team=""`, `-y team=testflight`, and `-y team=main`.

tlwr: Hadn't thought of that, yes I will

Also I will move the inline template into a fixture

tlwr:
Thanks to your suggestion I've made a measurably improvement - I hope.

I was using the slightly janky/brittle constants rather than `fmt.Sprintf` because setting the team name to be the name of the team was an error - I dropped 5cebbde6bf982e5b6390251b1f6e97216dad08f5 so now if you are of the team: "team-a" and you set "team: team-a" it is no longer an error.

I also ended up templating the rest of the `set-pipeline` pipeline, which means we can remove the other `fmt.Sprintf`, which means we can just use a regular fixture file, and take advantage of the suite helpers.

atc/exec/set_pipeline_step_test.go
tlwr: Resolved in latest force push

tlwr:
I ran testflight locally and it was green after 42 mins

I've tried to debug the testflight failures but just got a Concourse build spinner

If they go read again I might badger someone in Discord

vito:
@tlwr Shot in the dark - maybe it passes in series but fails when run in parallel? You could try `ginkgo -nodes=4 ./testflight`.

tlwr:
You were correct, the most recent commit fixes a race condition where multiple tests were creating pipelines using the same name. All green!

PR #4857, opened 2019-12-06
feat(atc): support "set-pipeline: self".

evanchaoli:
This is a follow-up change of PR#4708 to add "set-pipeline: self".
The RFC is https://github.com/concourse/rfcs/pull/31.

This commit is just a revert of:
> 4b00001  Remove "set-pipeline: self". Will use a separate PR to add "self" back.

# Contributor Checklist
> Are the following items included as part of this PR? Please delete checkbox items that don't apply.
- [ ] Unit tests
- [ ] Integration tests (if applicable)
- [ ] Updated documentation (located at https://github.com/concourse/docs)
- [ ] Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)

# Reviewer Checklist
> This section is intended for the core maintainers only, to track review progress. Please do not
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [ ] New config flags added? Ensure that they are added to the [BOSH](https://github.com/concourse/concourse-bosh-release) 
      and [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for the [integration tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env) (for example, if they are Garden configs that are not displayed in the `--help` text).

PR #5737, opened 2020-06-10
feat(atc): support "set-pipeline: self"

evanchaoli:
This is a substitution of #4857 as it is too out-of-date, no way to rebase, so I just start over.

In #4857, "self" is replaced with real pipeline name in the first place, which leads to a lot of interface changes. In this change, I just do that in the set_pipeline step, so this is a tiny change. I take the way because:

1. I forget what was the comment that requested me to replace "self" in the first place.
2. If we do the replacement in the first place `build_factory`, then that also requires a lot of interface changes. As this is considered as an experimental feature, could be reverted in future, it's better to constraint the change scope.
3. Also, if we replace in the first place, as we need to print warning message in set_pipeline step, then we have to add a flag in SetPipelinePlan to indicate "self" is specified, which also sounds unnecessary.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/exec/set_pipeline_step.go
+++ b/atc/exec/set_pipeline_step.go
@@ -97,6 +97,15 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState) error {
 	fmt.Fprintln(stderr, "\x1b[33mfollow RFC #31 for updates: https://github.com/concourse/rfcs/pull/31\x1b[0m")
 	fmt.Fprintln(stderr, "")
 
+	if step.plan.Name == "self" {
+		fmt.Fprintln(stderr, "\x1b[1;33mWARNING: 'set_pipeline: self' is experimental and subject to change!\x1b[0m")
+		fmt.Fprintln(stderr, "")
+		fmt.Fprintln(stderr, "\x1b[33mcontribute to discussion #5732 with feedback: https://github.com/concourse/concourse/discussions/5732\x1b[0m")
+		fmt.Fprintln(stderr, "")
+
+		step.plan.Name = step.metadata.PipelineName
+	}
+
 	source := setPipelineSource{
 		ctx:    ctx,
 		logger: logger,
--- a/atc/exec/set_pipeline_step.go
+++ b/atc/exec/set_pipeline_step.go
@@ -98,7 +98,7 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState) error {
 	fmt.Fprintln(stderr, "")
 
 	if step.plan.Name == "self" {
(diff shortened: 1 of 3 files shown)


atc/exec/set_pipeline_step.go
```diff
@@ -94,6 +94,15 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState) error {
 	fmt.Fprintln(stderr, "\x1b[33mfollow RFC #31 for updates: https://github.com/concourse/rfcs/pull/31\x1b[0m")
 	fmt.Fprintln(stderr, "")
 
+	if step.plan.Name == "self" {
+		fmt.Fprintln(stderr, "\x1b[1;33mWARNING: 'set_pipeline: self' is experimental and subject to change!\x1b[0m")
```
vito, line 98: Stronger scare-words might be a good idea so we can get more feedback:

```suggestion
		fmt.Fprintln(stderr, "\x1b[1;33mWARNING: 'set_pipeline: self' is experimental and may be removed in the future!\x1b[0m")
```

vito:
@evanchaoli Just a poke in case you didn't see my other review comment about testing this with unit tests instead of `testflight`. I think this is good to go otherwise.

evanchaoli:
@vito Oh, I missed your last comment about unit test. Just added some tests, and rebased to code.

atc/exec/set_pipeline_step.go
```diff
@@ -97,6 +97,17 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState) error {
...
 
+	if step.plan.Name == "self" {
+		fmt.Fprintln(stderr, "\x1b[1;33mWARNING: 'set_pipeline: self' is experimental and may be removed in the future!\x1b[0m")
+		fmt.Fprintln(stderr, "")
+		fmt.Fprintln(stderr, "\x1b[33mcontribute to discussion #5732 with feedback: https://github.com/concourse/concourse/discussions/5732\x1b[0m")
+		fmt.Fprintln(stderr, "")
+
+		step.plan.Name = step.metadata.PipelineName
+		// self must be set to current team, thus ignore team.
+		step.plan.Team = ""
```
vito, line 108: Good catch 👍

Issue #5753, opened 2020-06-11
Task vars causing concourse to ignore the credential manager resolution


pn-santos:
## Summary

When a task specifies `vars` concourse seems to not even attempt to resolve task vars via the configured credential manager. 

If one doesn't specify `vars` all of the tasks vars are resolved via the credential manager as expected.

## Steps to reproduce

1) Setup concourse with a credential manager (in our case vault but not sure if specific to vault).

2) Create a task definition with vars defined (making sure they exist in vault), e.g.:
```
params:
  MY_VAR: ((some-var))
```

3) invoke the task with:
```
- task: run-task
  file: resource/task.yml
```

4) invoke the task with:
```
- task: run-task-with-var
  file: resource/task.yml
  vars:
    blah: hello
```

## Expected results

Both tasks invocations should run

## Actual results

Invoking the task with `vars` (step 4) fails with:
```
failed to interpolate task config: undefined vars: some-var
```

## Additional context

This was discovered after we upgraded from 6.0 to 6.2 (possible cause: https://github.com/concourse/concourse/pull/5620)

## Triaging info

* Concourse version: 6.2
* Browser (if applicable):
* Did this used to work? Yes (up to 6.0 at least)

evanchaoli:
I just tried to reproduce the issue, and confirmed the bug is introduced by #5620. The root cause is that, I used `InterpolateTemplateConfigSource` to interpolate task vars first: https://github.com/concourse/concourse/blob/master/atc/exec/task_step.go#L163-L166, but I didn't notice that `InterpolateTemplateConfigSource` expects all keys: https://github.com/concourse/concourse/blob/master/atc/exec/task_config_source.go#L179.

I'll submit a fix soon.

Jaypinto:
Hi,

I seem to be encountering the same problem on 6.4. I've configured my task like this:

- task: Extract 
      config:
        platform: linux
        image_resource:
          type: docker-image
          source: {repository: busybox}
        run:
          path: sh
          args:
          - -c
          - |
            env
            echo ((nexus))
      params:
        CONCOURSE_VAULT_URL: <url>
        CONCOURSE_VAULT_CLIENT_TOKEN: <token>
        CONCOURSE_VAULT_AUTH_BACKEND: <app role>
        CONCOURSE_VAULT_AUTH_PARAM: <auth param>
        CONCOURSE_VAULT_PATH_PREFIX: <prefix>

I see the following error: 

failed to interpolate task config: undefined vars: nexus

[...]

PR #5758, opened 2020-06-12
bugfix: fixed a bug where task_vars prevented credential-manager and …

evanchaoli:
…var_sources from being used.

Fixes: #5753

Add a field `ExpectAllKeys` to `InterpolateTemplateConfigSource`, and for task_vars interpolation, use, it should not expect all keys.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/testflight/fixtures/task_vars.yml
+++ b/testflight/fixtures/task_vars.yml
@@ -1,4 +1,11 @@
 ---
+var_sources:
+- name: vs
+  type: dummy
+  config:
+    vars:
+      echo_text: text-from-var-source
+
 resources:
 - name: some-resource
   type: mock
@@ -30,7 +37,7 @@ jobs:
     vars:
       image_resource_type: mock
 
-- name: extarnal-task-vars-from-load-var
+- name: external-task-vars-from-load-var
   plan:
   - get: some-resource
   - load_var: foo
@@ -42,3 +49,16 @@ jobs:
     vars:
       image_resource_type: mock
       echo_text: ((.:foo))
+
+- name: task-var-is-not-used
+  plan:
+  - get: some-resource
+  - load_var: foo
+    file: some-resource/foo.txt
+  - task: process-task-definition
+    file: some-resource/task_unwrap.yml
+  - task: run
+    file: unwrapped-task-resource/task.yml
+    vars:
+      image_resource_type: mock
+      foo: bar
--- a/testflight/fixtures/task_vars.yml
+++ b/testflight/fixtures/task_vars.yml
@@ -50,11 +50,9 @@ jobs:
       image_resource_type: mock
       echo_text: ((.:foo))
 
-- name: task-var-is-not-used
+- name: task-var-is-defined-but-task-also-needs-vars-from-var-sources
   plan:
   - get: some-resource
-  - load_var: foo
-    file: some-resource/foo.txt
   - task: process-task-definition
     file: some-resource/task_unwrap.yml
   - task: run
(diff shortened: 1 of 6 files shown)


jamieklassen:
@aoldershaw this may also fix #5765

aoldershaw:
@jamieklassen looks like this PR doesn't fix that issue - left a comment: https://github.com/concourse/concourse/issues/5765#issuecomment-644213496

aoldershaw:
Tested with the following pipeline:

```yaml
resources:
- name: repo
  type: git
  source:
    uri: https://github.com/aoldershaw/concourse-experiments.git

jobs:
- name: job
  plan:
  - get: repo
  - task: something
    file: repo/tasks/echo-with-vars.yml
    vars:
      some-var-from-task: world
```

and the following task config:

```yaml
platform: linux

image_resource:
  type: registry-image
  source: {repository: busybox}

run:
  path: sh
  args:
  - -c
  - |
      echo ((some-var-from-vault))
      echo ((some-var-from-task))
```

with `/concourse/main/some-var-from-vault` set to `hello`.

On master, this errors with:
```
failed to interpolate task config: undefined vars: some-var-from-vault
```

With this bugfix, it succeeds.

Going to do some more thorough acceptance soon, but looks good so far!

adityabyreddy74:
Hi, Can we please have this PR reviewed and processed.

Thank you !

aoldershaw:
@adityabyreddy74 if you're blocked on this, you can work around the issue by interpolating each credential manager variable in the pipeline local variables. For instance, suppose your task uses two variables `var1` and `var2`, where `var1` comes from the credential manager and `var2` is statically defined in local variables. Doing this:

```yaml
- task: something
  file: repo/some-task.yml
  vars:
      var2: some-value
```

won't work without this PR merged, but doing this:

```yaml
- task: something
  file: repo/some-task.yml
  vars:
    var1: ((var1))
    var2: some-value
```

should work for the time being

testflight/fixtures/task_vars.yml
```diff
@@ -42,3 +49,16 @@ jobs:
     vars:
       image_resource_type: mock
       echo_text: ((.:foo))
+
+- name: task-var-is-not-used
```
aoldershaw, line 53: Minor nit: this test (and the underlying issue) isn't really about task vars not being used - it's about referencing both task variables and `var_source` (or global credential manager) variables at the same time in a task. Could you rename this job and the associated test accordingly?

evanchaoli: The point of this task is, `task: run` takes a var named `foo` that is not used in the task file, which simulates #5753. So I think the job name reflects to goal of this task.

aoldershaw: Hmm, what I'm struggling with though is that if we were to either get rid of the unused variable `foo` or reference `foo` in the task, this test case would still capture the issue in #5753 (i.e. fail on master and pass with this fix). So, this isn't really about testing an unused variable

aoldershaw: my point being, unused variables in task vars aren't the issue. For instance, consider the following task:

```yaml
platform: linux
image_resource:
  type: registry-image
  source: {repository: busybox}
run:
  path: echo
  args:
  - ((A))
```

with the following pipeline:

```yaml
...
jobs:
- name: job
  plan:
  - get: repo
  - task: something
    file: repo/tasks/echo-with-vars.yml
    vars:
      A: this is used
      B: this is not used
```

This will work on without this bug fix.

evanchaoli: Okey, I got your point. Just renamed the task.

adityabyreddy74:
@aoldershaw Our secrets are scope under image_resource section of tasks config. Alternatively we can pull image before task gets called and point task to use that image using `image` tag. We are good now. Thanks.

PR #5770, opened 2020-06-16
fly: allow pasting long tokens into login

### LINKED ISSUE #5717: fly login does not allow enough characters to paste in bearer token
## Summary

When doing a `fly login` while unable to use the `login_port` feature (i.e., while ssh'd into another machine), the other option is to paste the token in by using the "Copy to Clipboard" button in the login prompt. However, when you try to copy and paste the token, the token gets trimmed due to a character limit. (You can't even hit "Enter").

## Steps to reproduce

1. `fly -t <some-concourse> login`
1. Navigate to the login url provided, but change the port number in the url so it doesn't log you in automatically
1. Press the "Copy to Clipboard" button in your browser
1. Attempt to paste the bearer token into the awaiting `fly-cli` prompt
1. Attempt to hit enter

## Expected results

I get logged in.

## Actual results

I cannot hit enter and the bearer token is trimmed

## Triaging info

* Concourse version: _6.2.0_
* Browser (if applicable): _Chrome (but should be irrelevant because the clipboard actually still has the whole token)_
* Did this used to work? _I assume so :|_

aoldershaw:

closes #5717 .

On OSX, there's some magic constant of 1024 characters, which is the
length of some TTY buffer. Not sure of the specifics, but essentially
you weren't able to paste tokens longer than 1024 characters in `fly
login` for the "or enter token manually" option.

Rather than doing a `fmt.Scanf`, use a module that enables raw mode and
doesn't impose any limits on the line length. It was surprisingly
difficult to find a module that did what I wanted in a nice way.

* vito/go-interact depends on golang.org/x/crypto/ssh/terminal which
  imposes a seemingly arbitrary line length limit of 4096. We could
  remove this limit, but then we'd need to maintain a fork.
* chzyer/readline is *really* slow when pasting long text. I think it
  re-renders for every character instead of recognizing that we're in
  paste mode
* tcnksm/go-input seemed to have the same 1024 character limit (despite
  using raw mode?)

Eventually I landed on peterh/liner, which works pretty nicely. It
renders the text in a single line (and trims what's not in the "frame")
by default - there is a MultiLineMode, but it seemed pretty finicky.

@vito recommended just using a password input, since we probably
shouldn't be printing it anyway. Where it's supported, use that -
elsewhere (e.g. in tests), use the regular prompt.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/fly/commands/login.go
+++ b/fly/commands/login.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net"
 	"net/http"
@@ -14,6 +15,7 @@ import (
 	"github.com/concourse/concourse/fly/rc"
 	"github.com/concourse/concourse/go-concourse/concourse"
 	semisemanticversion "github.com/cppforlife/go-semi-semantic/version"
+	"github.com/peterh/liner"
 	"github.com/skratchdot/open-golang/open"
 	"github.com/vito/go-interact/interact"
 	"golang.org/x/oauth2"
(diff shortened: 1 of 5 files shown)


aoldershaw:
@taylorsilva oh boy, I'm dumb! I think I must've run `stty -icanon` at some point and forgot to reset it which made it seem like this code change fixed the problem, but actually didn't.

Not sure of the best way to approach this. I did find a hacky way of fixing this by modifying the underlying `golang.org/x/crypto/ssh/terminal` package that `go-interact` relies on (https://github.com/concourse/concourse/issues/5717#issuecomment-644400064), but I'm not sure if there's a better way. Switching to using `go-interact` (without the suggested patch) will bump the the limit to 4096 characters right away, but it should really accept arbitrarily long tokens

aoldershaw:
@taylorsilva I've updated the PR. It now tries to use a hidden input field so it's a bit more awkward to manually verify.

My token to ci.concourse-ci.org is ~1500 characters - I imagine yours will be similar. If you open https://ci.concourse-ci.org/login?fly_port=61319, you'll be able to copy the token manually, and can verify that it indeed takes the token in (but it does not on master, since it's too long)

aoldershaw:
@taylorsilva ah damn! Fixed the conflict

Issue #5699, opened 2020-06-01
Investigate: what schema changes are necessary for automatic archival?


vito:
* [ ] How do we identify which pipelines are still active?
* [ ] What happens when the job is removed from the pipeline?
    * note: this doesn't imply the row is gone; it becomes `active: false`
* [ ] What happens when the parent pipeline is destroyed?
* [ ] What happens when a build sets a few pipelines but then fails before the rest?
* [ ] What happens when two builds which set pipelines run concurrently?

vito:
For point 1, here are some theoretical queries that the garbage collector could run:

```sql
WITH pipelines_to_keep AS (
    SELECT pipeline_id
    FROM build_pipelines bp
    JOIN jobs j ON j.id = bp.job_id AND bp.build_id = j.latest_succeeded_build_id
)
UPDATE pipelines p
SET archived = true
WHERE parent_job_id IS NOT NULL
AND id NOT IN (SELECT pipeline_id FROM pipelines_to_keep)
```

Alternatively (not sure if this full syntax works):

```sql
UPDATE pipelines p
SET archived = true
FROM jobs j WHERE j.id = p.parent_job_id
WHERE not archived
AND NOT EXISTS (
    SELECT 1
    FROM build_pipelines bp
    WHERE bp.pipeline_id = p.id
    AND bp.build_id = j.latest_succeeded_build_id
)
```

This assumes a few things:

* we add `jobs.latest_succeeded_build_id`
* we add a `pipelines.parent_job_id`, which points to the job that configured the pipeline
* we add a `build_pipelines` table, which tracks which pipelines were set by which builds

There is also a bit more work to do after setting `archived = true`: the config for each pipeline needs to be cleared out. Right now that's done by setting an encrypted empty config value, so if we were to keep that behavior we'd need to loop over all the pipelines and archive them. We're going to go on a side detour to see if we can have those values just become `NULL` instead, and handle that case in code.

taylorsilva:
>  How do we identify which pipelines are still active?

We tried a different schema:

- Add a `parent_job_id` and `parent_build_id` to the `pipelines table
- added nothing else

We were able to make this query to get the pipelines we want to archive:

```sql
select p.id, p.name from pipelines p
JOIN jobs j ON j.id = p.parent_job_id AND j.last_successful_build_id != p.created_by_build_id
WHERE created_by_build_id IS NOT NULL
AND NOT archived;
```

---

> What happens when the job is removed from the pipeline?
[...]

PR #5789, opened 2020-06-17
atc/exec & atc/db: track set pipelines

### LINKED ISSUE #5734: Track which pipelines set other pipelines
Given a pipeline with a `set_pipeline` step
when the `set_pipeline` step runs it should update the parent job and build id of the pipeline it set.

This information tells us which job and build set a pipeline.

taylorsilva:

closes #5734 

* Add two columns to the `pipelines` table to track which pipelines/jobs set another pipeline: `parent_job_id` and `parent_build_id`
* Add a `build.SavePipeline()` method that always updates a pipeline with a parent job and build ID
* Add a `build.SetParentIDs()` method for updating the parent job and build IDs of a pipeline when there's no diff and the pipeline has no reason to be re-saved in its entirety
* The previous two methods are used by the `set_pipeline` step to save pipelines
* The `team` parameter that was recently added to `set_pipeline` should still work as expected. The testflight tests still passed locally for me

For more context on when this information will be used see #5699 

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/pipeline.go
+++ b/atc/db/pipeline.go
@@ -130,7 +130,9 @@ var pipelinesQuery = psql.Select(`
 		p.paused,
 		p.public,
 		p.archived,
-		p.last_updated
+		p.last_updated,
+		p.parent_job_id,
+		p.parent_build_id
 	`).
 	From("pipelines p").
 	LeftJoin("teams t ON p.team_id = t.id")
--- a/atc/db/pipeline.go
+++ b/atc/db/pipeline.go
@@ -40,6 +40,8 @@ type Pipeline interface {
 	Name() string
 	TeamID() int
 	TeamName() string
+	ParentJobID() int
+	ParentBuildID() int
 	Groups() atc.GroupConfigs
 	VarSources() atc.VarSourceConfigs
 	ConfigVersion() ConfigVersion
@@ -99,6 +101,8 @@ type pipeline struct {
 	name          string
 	teamID        int
 	teamName      string
+	parentJobID   int
+	parentBuildID int
 	groups        atc.GroupConfigs
 	varSources    atc.VarSourceConfigs
 	configVersion ConfigVersion
@@ -142,6 +146,8 @@ func (p *pipeline) ID() int                  { return p.id }
 func (p *pipeline) Name() string             { return p.name }
 func (p *pipeline) TeamID() int              { return p.teamID }
 func (p *pipeline) TeamName() string         { return p.teamName }
+func (p *pipeline) ParentJobID() int         { return p.parentJobID }
+func (p *pipeline) ParentBuildID() int       { return p.parentBuildID }
 func (p *pipeline) Groups() atc.GroupConfigs { return p.groups }
 
 func (p *pipeline) VarSources() atc.VarSourceConfigs { return p.varSources }
--- a/atc/db/pipeline.go
+++ b/atc/db/pipeline.go
@@ -1148,7 +1148,7 @@ func (p *pipeline) Variables(logger lager.Logger, globalSecrets creds.Secrets, v
 
 func (p *pipeline) SetParentIDs(jobID, buildID int) error {
 	if jobID <= 0 || buildID <= 0 {
-		return errors.New("job and build id cannot be zero-value")
+		return errors.New("job and build id cannot be negative or zero-value")
 	}
 
 	tx, err := p.conn.Begin()
@@ -1164,7 +1164,7 @@ func (p *pipeline) SetParentIDs(jobID, buildID int) error {
 		Where(sq.Eq{
 			"id": p.id,
 		}).
-		Where(sq.Or{sq.Lt{"parent_build_id": buildID}, sq.Eq{"parent_build_id": sql.NullInt64{}}}).
+		Where(sq.Or{sq.Lt{"parent_build_id": buildID}, sq.Eq{"parent_build_id": nil}}).
 		RunWith(tx).
 		Exec()
 
--- a/atc/db/pipeline.go
+++ b/atc/db/pipeline.go
@@ -1147,6 +1147,10 @@ func (p *pipeline) Variables(logger lager.Logger, globalSecrets creds.Secrets, v
 }
 
 func (p *pipeline) SetParentIDs(jobID, buildID int) error {
+	if jobID <= 0 || buildID <= 0 {
+		return errors.New("job and build id cannot be zero-value")
+	}
+
 	tx, err := p.conn.Begin()
 	if err != nil {
 		return err
@@ -1154,17 +1158,26 @@ func (p *pipeline) SetParentIDs(jobID, buildID int) error {
 
 	defer Rollback(tx)
 
-	_, err = psql.Update("pipelines").
+	result, err := psql.Update("pipelines").
 		Set("parent_job_id", jobID).
 		Set("parent_build_id", buildID).
 		Where(sq.Eq{
 			"id": p.id,
 		}).
+		Where(sq.Or{sq.Lt{"parent_build_id": buildID}, sq.Eq{"parent_build_id": sql.NullInt64{}}}).
 		RunWith(tx).
 		Exec()
 
 	if err != nil {
 		return err
+	}
+
+	rows, err := result.RowsAffected()
+	if err != nil {
+		return err
+	}
+	if rows == 0 {
+		return ErrSetByNewerBuild
 	}
 
 	return tx.Commit()
--- a/atc/db/pipeline.go
+++ b/atc/db/pipeline.go
@@ -94,6 +94,8 @@ type Pipeline interface {
 	Rename(string) error
 
 	Variables(lager.Logger, creds.Secrets, creds.VarSourcePool) (vars.Variables, error)
+
+	SetParentIDs(jobID, buildID int) error
 }
 
 type pipeline struct {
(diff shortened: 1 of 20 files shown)


atc/api/configserver/save.go
```diff
@@ -96,7 +96,7 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	_, created, err := team.SavePipeline(pipelineName, config, version, true)
+	_, created, err := team.SavePipeline(db.PipelineMetadata{Name: pipelineName}, config, version, true)
```
vito, line 99: I'm iffy on the naming here, since metadata elsewhere is used for fields that are shown to the user but ultimately have no significant bearing on the object itself. For example, on a container we store metadata for things like the step name - these values are shown in the API and used for querying in `fly intercept`, but they have no significance to the container itself. This metadata, on the other hand, is pretty crucial as it determines the ownership and lifecycle of the object.

Is this primarily to avoid adding optional arguments? If so, maybe we could add a second `SavePipeline` method like `SaveBuildPipeline` or something? Or maybe we could add a `build.SavePipeline` method, which would get the team/job from its existing relationships? 🤔 Just throwing ideas out there.

atc/exec/set_pipeline_step_test.go
```diff
@@ -231,82 +233,94 @@ jobs:
...
+						Expect(fakeDelegate.FinishedCallCount()).To(Equal(1))
+						_, succeeded := fakeDelegate.FinishedArgsForCall(0)
+						Expect(succeeded).To(BeTrue())
+					})
 				})
 
-				It("should stdout have message", func() {
-					Expect(stdout).To(gbytes.Say("setting pipeline: some-pipeline"))
-					Expect(stdout).To(gbytes.Say("done"))
+				It("always saves Job and Build ID's", func() {
```
aoldershaw, line 314: Did some acceptance on this, and noticed it wasn't updating the Job and Build IDs when there is no diff - I think you'll need a few test cases here to make sure it gets called when:
1. There's no pipeline yet (what this is testing now, though not explicitly)
2. There is a pipeline and the config differs
3. There is a pipeline and the config is the same

atc/api/configserver/save.go
aoldershaw: +1 to having a separate method - `build.SavePipeline` sounds cool

atc/exec/set_pipeline_step.go
```diff
@@ -157,13 +165,17 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState) error {
 		logger.Debug("no-diff")
 
 		fmt.Fprintf(stdout, "no diff found.\n")
+		err := pipeline.SetParentIDs(step.metadata.JobID, step.metadata.BuildID)
```
aoldershaw, line 168: Hmm, I'm not sure this is the best approach as I don't think it'll handle concurrent updates very well, given that we only want to set the parent IDs if the build is "newer" (i.e. has a larger `BuildID`).

Right now, `build.SavePipeline` will ensure that the parent IDs only get updated if the build is newer than the previous parent, and returns `ErrSetByNewerBuild` if not (the `set_pipeline` step should just log a warning if it receives this error).

So to me it seems like we should always be calling `build.SavePipeline`, rather than adding this extra `SetParentIDs` method. Unless you came up with a different approach for handling concurrent `set_pipeline`s?

chenbh: That's a good point, we'll have to add a check for that.

I guess another question is if it's possible for a pipeline to be set by 2 different jobs (from the same or even different pipelines). There's nothing in the RFC that forbids this but I'm not sure if there's a use case in the wild.

aoldershaw: Yeah - we discussed this with @vito, and his thoughts were that that's a use case we probably shouldn't try to support (though should document what happens in that case).

If we follow the same "only update if the build ID is larger" approach, it shouldn't matter much if different jobs are setting the same pipeline, since build IDs are auto incrementing across all jobs.

chenbh: One possible issue is that we "lose" previous build history if another job configures the same pipeline:
```
job1 kicks off 2 builds, job 2 kicks off 1 build

job1 build2 finishes -> initial create
job2 build1 finishes -> overwrites previous config
job1 build1 finishes -> overwrites previous config
```

chenbh: Oh right, forgot that build_id is globally incremental

aoldershaw: With the increasing build ID approach, the idea is that the resultant pipeline config is at least always independent of the order the builds finish in - it depends on the order the builds were created in

taylorsilva: @aoldershaw we could change how `set_pipeline` works where it always calls `build.SavePipeline` regardless if there's a diff or not. I think that would be fine. The main benefit we have from calling `SetParentIDs` instead of `SavePipeline` in this scenario, where there's no diff, is less load against the database. We only have to make one query instead of 9+ queries (it's actually almost always more than nine because we issue one query per resource/resource type/job/etc.). I'm trying to be conscious of when we issue more queries than is actually necessary; I think this is one of those situations where we can stay ahead performance-wise.

To solve the race-condition you mentioned, we can add `WHERE parent_build_id < buildID` to the query in `SetParentIDs`, like we did in `SavePipeline` and return the same `ErrSetByNewerBuild`.

aoldershaw: @taylorsilva although, it's only 9+ queries if the pipeline will be set (i.e. the `buildID` is newer) - and in that case, we'd need to make those queries anyway.

So with `SetParentIDs`, we're making 10+ queries in the common case rather than 9+. I think the stronger argument for only updating in `build.SavePipeline` is so the update is all-or-nothing, rather than introducing the potential for inconsistent state in the DB by having 2 updates (what if the `SetParentIDs` update succeeds but the `SavePipeline` one fails? the `parent_build_id` and `parent_job_id` are no longer accurate)

taylorsilva: I feel like I'm not understanding something that you're trying to convey. Here's my response based on this assumption: you think we run both `SetParentIDs` and `SavePipeline` every time. Let me know if that assumption is wrong. The below may be invalid.

---

We would never run both `SetParentIDs` AND `SavePipeline`. It's one OR the other based on whether a diff exists or not.

If you look at `set_pipeline` currently, it exits early when there's no dif:
https://github.com/concourse/concourse/blob/becd8ab072ccaa6cdb513e73dd8e7dbe69c384f2/atc/exec/set_pipeline_step.go#L193-L200

We're trying to maintain that behaviour. When we fall into this `if` we do the minimum that we expect from this step which is updating the job and build id of the pipeline the step sets.

aoldershaw: Ah, your assumption is correct, thanks for the clarification. I didn't read the code carefully enough. Adding `WHERE parent_build_id < buildID` to `SetParentIDs` makes sense!

aoldershaw:
Hey @taylorsilva, is this ready to be reviewed? Also, looks like there are some merge conflicts

taylorsilva:
@aoldershaw not yet. I'll mark it "Ready to Review" when it is. It will no longer appear as a draft.

atc/exec/set_pipeline_step.go
```diff
@@ -217,7 +207,16 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState) error {
...
+		return err
+	}
+
+	if !found {
+		return fmt.Errorf("set_pipeline step not attached to a buildID")
+	}
+
+	pipeline, _, err = parentBuild.SavePipeline(step.plan.Name, team.ID(), atcConfig, fromVersion, false)
 	if err != nil {
 		return err
```
aoldershaw, line 221: Previously, we discussed special casing on `db.ErrSetByNewerBuild` here to log a warning but not error the build outright. Is that still something we want to do? e.g.

```go
	if err != nil {
		if err == db.ErrSetByNewerBuild {
			fmt.Fprintln(stderr, "\x1b[1;33mWARNING: the pipeline was set by a newer build, so this step is a no-op\x1b[0m")
			return nil
		}
		return err
	}
```

atc/db/pipeline.go
```diff
@@ -1147,26 +1147,39 @@ func (p *pipeline) Variables(logger lager.Logger, globalSecrets creds.Secrets, v
 }
 
 func (p *pipeline) SetParentIDs(jobID, buildID int) error {
+	if jobID <= 0 || buildID <= 0 {
+		return errors.New("job and build id cannot be zero-value")
```
aoldershaw, line 1151: Mismatch between this error and the test (test expects `job and build id cannot be negative or zero-value`)

atc/db/pipeline.go
```diff
@@ -1147,26 +1147,39 @@ func (p *pipeline) Variables(logger lager.Logger, globalSecrets creds.Secrets, v
...
 		}).
+		Where(sq.Or{sq.Lt{"parent_build_id": buildID}, sq.Eq{"parent_build_id": sql.NullInt64{}}}).
 		RunWith(tx).
 		Exec()
 
 	if err != nil {
 		return err
 	}
 
+	rows, err := result.RowsAffected()
```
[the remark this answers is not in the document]
taylorsilva: > Does Exec() not return an error if there are no rows affected?
That is correct. We initially used that `if` you referenced in `team.go`. The test that expects the error the `ErrSetByNewerBuild` was failing, which made us realized Exec() only returns an error if the query fails to run, and affecting zero rows isn't a failure at the sql level.

atc/db/pipeline.go
taylorsilva: weird, thought I changed that!

taylorsilva:
@aoldershaw updated with all changes! Thanks for the reminder about that title on that one commit. I initially left it blank because I couldn't think of a good title and then completely forgot to go back to it 😅

PR #5802, opened 2020-06-23
ListAllJobs endpoint can be watched for changes

aoldershaw:

Related RFC: concourse/rfcs#61

This PR aims to provide functionality similar to Kubernetes' `?watch` API (e.g. `kubectl get ... --watch`) for the `ListAllJobs` endpoint.

abort-job

set-pipeline

 The motivation for this is described more in depth in the RFC, but essentially, it should provide:

* (Hopefully) better performance on the ATC and DB
* Faster feedback in the UI dashboard

Eventually, I see this change applying to several endpoints - rather than polling in the UI, we have the ATC send us updates.

## Notes to reviewer:

To test: visit the dashboard, and then modify some jobs! (e.g. trigger builds, set/delete pipelines, expose/hide pipelines, etc)

This is feature flagged behind the `--enable-watch-endpoints` flag, which I've enabled by default in `docker-compose.yml`.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [x] PR acceptance performed
  the [integration
  `--help` text).


atc/db/watch/list_all_jobs_watcher_test.go
```diff
@@ -0,0 +1,364 @@
...
+					job1.CreateBuild()
+					Consistently(getInvokedEvents).Should(BeEmpty())
+				})
+
+				Context("when the notification is a delete", func() {
+					It("forwards the notification", func() {
+						pipeline.Destroy()
+						Eventually(getInvokedEvents).Should(ContainElement(DeleteEvent(job1.ID())))
+					})
+				})
```
chenbh, line 254: What's the scenario where I'm not part of the team but should be notified a job is deleted?

aoldershaw: Hmm yeah, this is really more of a limitation of the approach than a desired feature, so maybe this test shouldn't exist.

If we wanted to avoid sending notifications when jobs that aren't visible to a user are deleted, it would end up being pretty costly (in terms of additional queries and triggers) and probably pretty complicated - so I figured the tradeoff of sending every delete event to all clients was worth it (especially given that there's no sensitive info on delete events - just the job ID)

jamieklassen:
I have come back to this several times and never really convinced myself I understood how it works. The subtleties of queuing/not queuing notifications and handling dropped/interrupted connections or other edge-casey things seem like they would be easy to get wrong, and I couldn't grasp what `dirty` meant. If I was designing something like this from the ground up I'm sure I'd follow the example of kubernetes and use something like etcd which already has a "watch" system.

However, I don't want to block this potentially-valuable performance improvement just because my brain is too small. The feature flag seems to work properly so this is totally backwards-compatible, so I'd rather focus on getting it merged and load-tested instead of staring at it forever. So let's get it rebased, add the feature flag into the helm chart and bosh release, and turn it on in hush-house.

Maybe we can focus some effort on planning the load test experiment - what will we measure (memory on the web nodes, network i/o between the DB and web, number of DB connections? probably something else on the DB)? what do we expect to change? this exercise might motivate us to add some observability. A gauge for the number of open ListAllJobs connections might not be amiss.

jamieklassen (dismissed):

aoldershaw:
@jamieklassen thanks for the review! Hmm, I think if *you* are struggling to fully understand it then there's definitely a need to improve upon the clarity - I'll try to think about that before merging.

I'd also like to a load test experiment as you suggested, but haven't found the time. I think it's probably wise to validate the impact it has before adding so much more complexity to the codebase

aoldershaw:
Opened #6084 to track load testing

aoldershaw:
@jamieklassen I've done some load testing and documented it here: #6084

The skinny is that, under heavy load (tested with 100 active "dashboards"), enabling watch endpoints seems to result in:

* Reduced ATC CPU usage, and as a result, reduced build scheduling
  * Without watch endpoints, the ATCs CPU was pegged at 100%, and build scheduling dropped by ~50% as result
  * With watch endpoints, the ATCs CPU was around 85-95% (compared to 80% in the baseline), and build scheduling only dropped by ~10%
* Reduced network transmission (ATC -> browser dropped by ~70%, DB -> ATC dropped by ~60%)
* Slightly increased memory usage on both ATC/DB (~10-15%)

Not under client load (just the regular build scheduling load), having watch endpoints enabled resulted in slightly higher DB CPU and memory (~5-8% for each metric) - worth noting that at least some of this could be ascribed to the fact that the watch endpoint test was run after the vanilla concourse test, so the magnitude of data in the DB was a fair bit larger. The completed DB transaction rate didn't seem to be significantly affected, nor did DB connections.

---

In doing the load testing, I noticed some performance issues on the UI, primarily caused by writing the jobs cache to localStorage after every batch of events. I added some optimizations to mitigate that (9af0bc72458b96e3cb3ceb62f1c1a45052deb2f4, 100e067c602edc1f31b14aba0c53f2678f22e207).

> this exercise might motivate us to add some observability. A gauge for the number of open ListAllJobs connections might not be amiss

While looking into this, I noticed we have a `GaugeVec` for concurrent requests as part of the concurrent request limiting you worked on. I wonder if we should just always emit the number of `ListAllJobs` concurrent requests even if there is no concurrent request limit set.

It also makes me wonder about how the watch endpoints should play with the concurrent request limit - it's no longer a "you'll get your turn eventually, so just keep retrying" sort of thing. What do you think?

chenbh:
Closing as it looks like we won't be prioritizing this anytime soon and the merge conflicts look scary

Issue #5136, opened 2020-02-07
containerd: investigate use of firewalld rather than iptables


cirocosta:
### what challange are you facing

we have a hypotheesis that `iptables` can be quite resource intensive when the
number of containers get high (or the number of operations on it is large),
leading to problems like describe in https://github.com/concourse/concourse/issues/3694.

in the current networking implementation set up for containerd, we're defaulting
to use `iptables`, and that works fine, it might be that with `firewalld` (a
backend supported by the `firewall` CNI plugin) we achieve even better
performance (given that it can operate on rule diffs rather than having to
perform flushes for each change).

### what would make this better

be able to decide how we want to go with regards to the component that deals
with networking: iptables? firewalld?

references:

- cni: https://github.com/containernetworking/cni
- cli plugins: https://github.com/containernetworking/plugins
- iptables: http://man7.org/linux/man-pages/man8/iptables.8.html
- firewalld: https://firewalld.org/documentation/

vito:
We can probably close https://github.com/concourse/concourse/issues/3127 once this is done.

xtreme-sameer-vohra:
Hey folks,
@muntac and I were looked at this briefly. Given #3127 , we found a [source](https://www.tecmint.com/firewalld-vs-iptables-and-control-network-traffic-in-firewall/) that seemed to imply that we might experience the same issue with `firewalld` as with `iptables` as they both use the `netfilter` framework and weren't able to find any docs that explicitly confirmed the issue would be solved by switching

From source
```
Under the hood, both firewalld and the iptables service talk to the netfilter framework in the kernel through the same interface, not surprisingly, the iptables command. However, as opposed to the iptables service, firewalld can change the settings during normal system operation without existing connections being lost. 
```

Does anyone have more insight into this ? We haven't looked at the internals of `firewalld` to see how it leverages `netfilter`.

vito:
[...]

Issue #847, opened 2017-01-03
Periodic generates 'insufficient subnets remaining in the pool'


christophe-f:
* Concourse version: 2.5.0
* Deployment type (BOSH/Docker/binary): Aws Community image
* Infrastructure/IaaS: AWS EC2
* Browser (if applicable):
* Did this used to work? YES

Up to Dec 31 the periodic feature was working fine. After Jan 1st, not anymore. 
I get the error message 'insufficient subnets remaining in the pool'

Restarting the RDS and EC2 instances seems to resolve it.

vito:
What do you mean by periodic?

'insufficient subnets' means your workers hit the limit on the number of containers it can run. this usually means you'll need more workers for the pipelines you're running - this it can also happen if you have suddenly a bunch of failed jobs, since we'll keep those containers around for you to investigate, until the next build finishes

christophe-f:
Sorry, thought the type was periodic but it is type: time with interval: 2h

Make sense. After checking the build history, seems that we got 1 job lasting for 18 hours. I guess it could have caused the issue.

christophe-f:
Can I apply serial: true with the type time?

Thanks for your help Vito.

dmlemos:
We have seen this issue as well. 3 workers, two of them hit their limits. Recreating the workers each week seems to help, but we still encounter the issue from time to time.

They are nowhere near their capacity.
@vito I can either use less powerful aws ec2 instances and create more of them, or do you think it is wise to increase the max_container limit on the garden config?

miromode:
Seeing this as well. 3 workers with # of containers slowly creeping up to 256-ish, then start getting the `insufficient subnets ...` message. Restart fixes it but it starts happening again.

Is there a non-restart way to force the garbage containers to be collected? What can I do to see/provide more debugging context?

Sharor:
Just wanted to chip in, this happens for Kubernetes orchestration as well and is a bit painful. It happens even with green builds, although adding more workers helps a bit. We run a 1:2 ratio of pipelines to workers, and it happens every few days.

mariash:
Do you see this issue with 3.x? 3.x has improvements around container usage and garbage collection, so the number of containers should drop significantly. If you still see it with 3.x, please report what containers do you have.

Sharor:
We ended up following advice on Slack, and changing our infrastructural stack to BOSH. :slightly_smiling_face: Ill give it a test run though! Thanks.

robertjsullivan:
[...]

PR #5829, opened 2020-06-29
containerd: restrict traffic to specified networks

### LINKED ISSUE #5534: containerd: allow configuring deny networks
**What problem are you facing**
When using `containerd` as the worker backend it's not possible to specify "network ranges to which traffic from containers will be denied." This is possible with `guardian` using the `--deny-network` flag on the `gdn` binary (`CONCOURSE_GARDEN_DENY_NETWORK` flag on the `concourse worker` command). This is used on [Hush House](https://github.com/concourse/hush-house/blob/master/deployments/with-creds/workers/values.yaml#L35-L36) to restrict access to the IaaS services.

**What would make this better**
Having a flag that allows configuring a similar deny-network fucntionality for `containerd`. `containerd` itself is not concerned with container networking. This would likely have to be done through [CNI](https://github.com/containernetworking/cni). They maybe a way of doing this with one of the official [CNI plugins](https://github.com/containernetworking/plugins#plugins), or a [3rd party plugin](https://github.com/containernetworking/cni#3rd-party-plugins) may exist. Otherwise, a plugin may have to be implemented.

muntac:
Allows user to use `CONCOURSE_GARDEN_DENY_NETWORK` to pass in network ranges that containers (i.e. job builds) should not be able to access.

closes #5534 .

The CNI firewall [plugin](https://github.com/containernetworking/plugins/tree/master/plugins/meta/firewall) we are [using](https://github.com/concourse/concourse/blob/f2b45207df3ce0f0bd23094c363b6b65116b9566/worker/runtime/cni_network.go#L82) does not provide this functionality. Other plugins from 3rd party vendors usually were not lightweight, or did not have firewall capabilities.

Therefore, we decided to go about implementing this ourselves. The firewall plugin we're using makes use of [iptables](https://linux.die.net/man/8/iptables). It can use `firewalld` as a backend as well. After spending some time on #5136 we decided to [relegate](https://github.com/concourse/concourse/issues/5136#issuecomment-646872293) the choice of a different backend (firewalld, nftables) would be left for that future issue as the implementation is small enough. We also discovered that certain [issues](https://github.com/concourse/concourse/issues/847#issuecomment-647718453) like the failure to allocate IPs after a certain container amount are due to a misunderstanding of `guardian` allocates IPs and not a performance problem with `iptables`.

Initial state of iptables (run `iptables -L`) when worker starts up:

```
Chain INPUT (policy ACCEPT)
target     prot opt source               destination

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination
```

The firewall plugin creates the chains `CNI-ADMIN` and `CNI-FORWARD` once the first container is brought up by a build.

```
Chain INPUT (policy ACCEPT)
target     prot opt source               destination

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination
CNI-FORWARD  all  --  anywhere             anywhere             /* CNI firewall plugin rules */

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination

Chain CNI-ADMIN (1 references)
target     prot opt source               destination

Chain CNI-FORWARD (1 references)
target     prot opt source               destination
CNI-ADMIN  all  --  anywhere             anywhere             /* CNI firewall plugin rules */
ACCEPT     all  --  anywhere             10.80.0.2            ctstate RELATED,ESTABLISHED
ACCEPT     all  --  10.80.0.2            anywhere
```

As seen above, the plugin puts `CNI-FORWARD` as the first rule in the `FORWARD` chain, effectively routing all forwarding decisions to that chain. `CNI-FORWARD` will have routing rules for each container that comes up allowing traffic to and from anywhere.

However, it puts the `CNI-ADMIN` chain as its first rule. We [confirmed](https://github.com/containernetworking/plugins/issues/504#issuecomment-648907222) that this is mean for top level user defined rules that get consulted first before container specific rules. Hence, it's an appropriate spot for rules to restrict all container traffic.

We also [confirmed](https://github.com/containernetworking/plugins/issues/504#issuecomment-648907222) that `CNI-ADMIN` can be renamed to a name of our choice using this undocumented [field](https://github.com/containernetworking/plugins/blob/d0eeb27494b52c5e2ea623e9ea9d60aa3921f3ef/plugins/meta/firewall/firewall.go#L43) in our plugin [configuration](https://github.com/concourse/concourse/blob/2f62b6783db4d7ed2568e87d2789a0a49c6cfb0f/worker/runtime/cni_network.go#L86). We decided to rename it to `CONCOURSE-OPERATOR` as this would be an appropriate place for the operator to setup other overriding rules of their choice.

Upon the garden server starting we now setup this chain and add any network addresses that may have been passed via `CONCOURSE_GARDEN_DENY_NETWORK`. The output looks like this.

```
Chain INPUT (policy ACCEPT)
target     prot opt source               destination

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination

Chain CONCOURSE-OPERATOR (0 references)
target     prot opt source               destination
REJECT     all  --  anywhere             152.34.4.5           reject-with icmp-port-unreachable
```

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


worker/runtime/cni_network.go
```diff
@@ -220,6 +235,63 @@ func (n cniNetwork) SetupMounts(handle string) ([]specs.Mount, error) {
...
+			return err
+		}
+	}
+	return nil
+}
+
+// Create chain - ignore error if it already exists
+func createIptablesChain() error {
+	cmd := exec.Command("iptables", "-t", "filter", "-N", "CNI-ADMIN")
+	_, err := cmd.Output()
```
chenbh, line 256: Running a binary and parsing the output always feels kinda fragile to me

- https://godoc.org/github.com/coreos/go-iptables/iptables
- https://godoc.org/github.com/docker/libnetwork/iptables
- https://godoc.org/github.com/google/nftables (successor to iptables)

aoldershaw: Yeah, I agree - although, worth noting that at least the two iptables libraries above do essentially the same thing (but probably better to let them deal with it rather than us!)

muntac: I agree that it's preferable to avoid using binaries. There are a few reasons we made an exception. 

There are exactly four calls being made to `iptables` here. The two `iptables` libraries mentioned above would result in four analogous calls with similar parameters, which would then in turn call the binary. We felt this did not give enough benefits to import an extra library. Given the specificity of the error handling and optimization we're doing in table manipulation these direct calls also offer more transparency.

The libraries themselves, while from recognized orgs, are not officially affiliated with the Linux `netfilter` which is what `iptables` is making use of. They would break in the same way our code would if there was ever a breaking change on a newer version of `iptables`.

`nftables` is a good suggestion. We looked at whether to use it versus `iptables` or `firewalld` and decided further investigation is [needed](https://github.com/concourse/concourse/issues/5136#issuecomment-646872293) which will be done in #5136 in the future.

aoldershaw: Worth noting that just because `iptables` is on the `PATH`, doesn't mean it'll behave consistently everywhere. Looking briefly at coreos/ip-tables, it looks like they parse the `iptables` version and behave slightly differently based on that.

https://github.com/coreos/go-iptables/blob/521ee6c1700638dc6444e0a9a7511ad3fd7d966a/iptables/iptables.go#L476-L478
```
// getIptablesVersion returns the first three components of the iptables version
// and the operating mode (e.g. nf_tables or legacy)
// e.g. "iptables v1.3.66" would return (1, 3, 66, legacy, nil)
```

According to this issue https://github.com/coreos/go-iptables/issues/49, it also looks like `nf_tables` can be used as a backend for the `iptables` command (I guess?), and that affects the output format in some cases.

My point being it seems like there's a fair bit of nuance around interfacing with `iptables` - we may not experience it with how we're using it, but I'm not confident that we never will - so maybe it's worth not taking on that complexity ourselves.

muntac: @aoldershaw Digging into the code of `go-iptables` I see that it tries to detect the version of `iptables` but does not act upon this information in any significant way. The `mode` field which can be either `nftables` or `legacy` only gets acted upon for the where the `iptables` command is used in verbose mode (using `-v`). This command lists two new columns, `pkts` & `bytes`, which we don't have any use for.

Great point about the `PATH`! I'm not fully sure about the implications of this for the typical operator but it's something good to keep in mind.

I noticed that the [CNI network plugins](github.com/containernetworking/plugins), which we dependon,  use `go-iptables` under the hood as well. This somewhat counters my earlier point about breaking changes. We'd probably want to follow whatever versions of `go-iptables` that project is using, and if they decided it was worth updating against a breaking change might make sense for us to do so as well.

Given those two points I'm leaning on going with your suggestion of using `go-iptables`. Will let you know once the changes are in.

worker/workercmd/worker_linux.go
```diff
@@ -26,12 +26,12 @@ type GardenBackend struct {
...
-	DNS            DNSConfig     `group:"DNS Proxy Configuration" namespace:"dns-proxy"`
-
-	RequestTimeout time.Duration `long:"request-timeout" default:"5m" description:"How long to wait for requests to Garden to complete. 0 means no timeout."`
+	Bin                string        `long:"bin"        description:"Path to a garden backend executable (non-absolute names get resolved from $PATH)."`
+	Config             flag.File     `long:"config"     description:"Path to a config file to use for the Garden backend. Guardian flags as env vars, e.g. 'CONCOURSE_GARDEN_FOO_BAR=a,b' for '--foo-bar a --foo-bar b'."`
+	DNSServers         []string      `long:"dns-server" description:"DNS server IP address to use instead of automatically determined servers. Can be specified multiple times."`
+	DNS                DNSConfig     `group:"DNS Proxy Configuration" namespace:"dns-proxy"`
+	RestrictedNetworks []string      `long:"restricted-networks" description:"Network ranges to which traffic from containers will be restricted. Can be specified multiple times."`
```
aoldershaw, line 33: Maybe rename the flag so it's the singular form (matches `dns-server` above)

```suggestion
	RestrictedNetworks []string      `long:"restricted-network" description:"Network ranges to which traffic from containers will be restricted. Can be specified multiple times."`
```

worker/runtime/cni_network.go
```diff
@@ -220,6 +250,20 @@ func (n cniNetwork) SetupMounts(handle string) ([]specs.Mount, error) {
 	}, nil
 }
 
+func (n cniNetwork) SetupRestrictedNetworks() error {
+	const tableName = "filter"
+	err := n.ipt.CreateChainOrFlushIfExists(tableName, ipTablesAdminChainName)
+
+	// Create REJECT rules in empty admin chain
```
aoldershaw, line 257: I remember when exploring the iptables rules in CF as part of the networking onboarding week, we encountered a rule that looks like:

```
-A w--default -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
```

https://github.com/cloudfoundry/guardian/blob/b9201b81a0a5adfe9cafec2cb954e676a28e567b/kawasaki/iptables/global_chains.go#L294

I believe this was an optimization to prevent checking the destination of every packet once the connection is already established (we only check on the initial handshake packets).

We should probably do the same

worker/workercmd/worker_linux.go
aoldershaw: Also, was it a deliberate decision to change from guardian's `deny-network` flag? I'm guessing yes but I'm not sure where we landed on naming containerd flags yet

Worth noting that if you specify the old Guardian flag `CONCOURSE_GARDEN_DENY_NETWORK` with containerd enabled, it doesn't error - it's just a silent no-op (which seems dangerous - maybe this'll be addressed in the future "flags" epic)

worker/runtime/cni_network.go
```diff
@@ -79,13 +82,14 @@ func (c CNINetworkConfig) ToJSON() string {
       }
     },
     {
-      "type": "firewall"
+      "type": "firewall",
+	  "iptablesAdminChainName": "%s"
```
aoldershaw, line 86:
```suggestion
      "iptablesAdminChainName": "%s"
```

worker/workercmd/worker_linux.go
muntac: Good point. The flags epic should come up with proper naming, and take care of the error handling you mention. I'm going to switch this back to use `CONCOURSE_GARDEN_DENY_NETWORK` for now instead of introducing a new flag that might get changed.

worker/runtime/cni_network.go
muntac: Interesting point. I believe you're suggesting adding an "accept-all" rule for RELATED and ESTABLISHED connections before the REJECT rule. So the output using `CONCOURSE_GARDEN_DENY_NETWORK=152.34.4.5` would be

```
target     prot opt source               destination
ACCEPT     all  --  anywhere             anywhere             ctstate RELATED,ESTABLISHED
REJECT     all  --  anywhere             152.34.4.5      reject-with icmp-port-unreachable
```

I'm not aware of the exact mechanics of the kernel implementation of the connection tracking module. But from my understanding of `iptables`, for the first rule to work each packet, even when RELATED or ESTABLISHED, will be checked against the rule to ensure conformance. This results in an operation that likely involves checking the connection state.

Without this rule, each packet would be checked regardless of state against the `REJECT` rule to see if destination matches.

In that crude sense the performance should be roughly the same with two rules or one rule, as both cases involve a rule checking operation. If the connection status is NEW there would actually be an extra operation for the two rule case as two rules would have to be checked against.

Again, I say this without understanding the kernel [implementation](https://github.com/torvalds/linux/blob/f85c1598ddfe83f61d0656bd1d2025fa3b148b99/net/netfilter/nf_conntrack_core.c), and from not seeing performance benefits mentioned in the [docs](https://linux.die.net/man/8/iptables).

aoldershaw: I guess the idea would be that if there were 10 deny network ranges, we'd need to perform 10 checks for every packet that isn't denied, even though we could skip those checks for the vast majority of those packets.

e.g. if we're uploading a 1GB file, assuming an MTU of 1500 bytes/packet per https://stackoverflow.com/a/2614188/9994448, that means we're sending out few packets (maybe just 1?) to establish a connection, and ~600,000 packets with data. With the "accept all RELATED and ESTABLISHED" rule, we're doing ~600,000 checks - without it, we're doing ~6,000,000.

That said, it's probably a good idea to benchmark it in any case, or maybe talk to some Guardian folks to see why they went with that approach.

muntac: Great point! I was about to reply with a similar realization about checks being made until a match happens. Even with no deny network ranges, and 10 running containers, a packet from an ESTABLISHED/RELATED connection in the 10th container (e.g. outbound from 10.80.0.10) would go through 20 rules before it found a match.

```
Chain CNI-ADMIN (1 references)
target     prot opt source               destination

Chain CNI-FORWARD (1 references)
target     prot opt source               destination
CNI-ADMIN  all  --  anywhere             anywhere             /* CNI firewall plugin rules */
ACCEPT     all  --  anywhere             10.80.0.2            ctstate RELATED,ESTABLISHED
ACCEPT     all  --  10.80.0.2            anywhere
ACCEPT     all  --  anywhere             10.80.0.3            ctstate RELATED,ESTABLISHED
ACCEPT     all  --  10.80.0.3            anywhere
ACCEPT     all  --  anywhere             10.80.0.4            ctstate RELATED,ESTABLISHED
ACCEPT     all  --  10.80.0.4            anywhere
ACCEPT     all  --  anywhere             10.80.0.5            ctstate RELATED,ESTABLISHED
ACCEPT     all  --  10.80.0.5            anywhere
ACCEPT     all  --  anywhere             10.80.0.6            ctstate RELATED,ESTABLISHED
ACCEPT     all  --  10.80.0.6            anywhere
ACCEPT     all  --  anywhere             10.80.0.7            ctstate RELATED,ESTABLISHED
ACCEPT     all  --  10.80.0.7            anywhere
ACCEPT     all  --  anywhere             10.80.0.8            ctstate RELATED,ESTABLISHED
ACCEPT     all  --  10.80.0.8            anywhere
ACCEPT     all  --  anywhere             10.80.0.9            ctstate RELATED,ESTABLISHED
ACCEPT     all  --  10.80.0.9            anywhere
ACCEPT     all  --  anywhere             10.80.0.10            ctstate RELATED,ESTABLISHED
ACCEPT     all  --  10.80.0.10            anywhere
```

Like you said this would happen for the vast majority of checks and would be reduced to a single check if we add the bypass for ESTABLISHED/RELATED connections. Will update soon, thanks!

aoldershaw: Oh, neat! Good find

muntac:
@aoldershaw Amazing acceptance testing! I believe the worker is continuously restarting upon that error being thrown.

`backend.Start()` fails as a result of the error. The error bubbles up to this [Run()](https://github.com/concourse/concourse/blob/3295bbd95ad6b84d923fad6b777fb6970067f1a0/worker/workercmd/garden_server_runner.go#L15) function in `garden_server_runner.go` which is implementing [Run()](https://github.com/tedsuo/ifrit/blob/9a97d0632f00efd5508c1a211032f07de3255d23/process.go#L71) from`ifrit`. At no point does the error get logged, only bubbled up.

The particular `ifrit` construct we're [using](https://github.com/concourse/concourse/commit/d8092b5f0453cbb0b13b1a3586d32cad0f8f77f4#diff-e262c92980c14e2f98582980fd4fb446L30-R35) for `containerd` is a [Restarter](https://github.com/tedsuo/ifrit/blob/master/restart/restart.go). This commit [message](https://github.com/concourse/concourse/commit/d8092b5f0453cbb0b13b1a3586d32cad0f8f77f4) from @cirocosta mentions that this is to ensure that the garden server comes up after containerd by continuously restarting until containerd is up.

A quick fix is to make sure the error is printed in our implementation of Run() or in the Load() function of the restarter. This ensures an explicit failure. But the continuous restart isn't great. As the commit message points out a timeout for restarts should probably be put in. I'm going to look into that.

muntac:
Re: timeout
I think an easier solution would be to use the [Ordered](https://godoc.org/github.com/tedsuo/ifrit/grouper#NewOrdered) strategy instead of Parallel. That way Ifrit ensures the containerd process comes up before the garden server, and we won't have to use the Restarter.

muntac:
@aoldershaw Taking out the Restarter stops the worker process upon the garden server failure and causes the error to be logged. You should now see the following in the worker logs for the failure case:

```
worker_1  | time="2020-07-09T20:37:04.547191543Z" level=info msg="containerd successfully booted in 0.026184s"
worker_1  | {"timestamp":"2020-07-09T20:37:04.581087078Z","level":"error","source":"worker","message":"worker.garden-runner.logging-runner-exited","data":{"error":"Exit trace for group:\ncontainerd-garden-backend exited with error: setup restricted networks failed: appending reject rule for restricted network not-an-ip failed: running [/sbin/iptables -t filter -A CONCOURSE-OPERATOR -d not-an-ip -j REJECT --wait]: exit status 2: iptables v1.6.0: host/network `not-an-ip' not found\nTry `iptables -h' or 'iptables --help' for more information.\n\ncontainerd exited with nil\n","session":"8"}}
worker_1  | {"timestamp":"2020-07-09T20:37:04.581552911Z","level":"info","source":"worker","message":"worker.debug-runner.logging-runner-exited","data":{"session":"10"}}
.
.
.
worker_1  | error: Exit trace for group:
worker_1  | garden exited with error: Exit trace for group:
worker_1  | containerd-garden-backend exited with error: setup restricted networks failed: appending reject rule for restricted network not-an-ip failed: running [/sbin/iptables -t filter -A CONCOURSE-OPERATOR -d not-an-ip -j REJECT --wait]: exit status 2: iptables v1.6.0: host/network `not-an-ip' not found
worker_1  | Try `iptables -h' or 'iptables --help' for more information.
worker_1  |
worker_1  | containerd exited with nil
worker_1  |
.
.
.
concourse_worker_1 exited with code 1
```

muntac:
Thanks! I'm pretty sure the only such case is ensuring `containerd` is up which we now do through the Ordered strategy. In `guardian` they call the `garden/server` `Serve` function without any retry logic (we use `Start`which in turn calls `Serve`).

Issue #5712, opened 2020-06-03
Web UI Performance


matthewpereira:
### Problem

- Concourse’s dashboard UI is a sophisticated front end experience for monitoring and interacting with pipelines, jobs, resources, and build logs. However, as customers push their Concourse usage to extreme scale, performance on the dashboard suffers.
- Specific roles require more advanced workflow possibilities to reduce the amount of manual effort involved in navigating the dashboard UI

### Proposed Solutions
- Build logs that load only the most pertinent info by default, with more available on demand
- Stripped down, operations-focused version of the dashboard for maximum performance

### Historical context
#5032 - What information should we have on the dashboard?
#5193 - Preloading dashboard info

### Specific origin
#5341 - The dashboard page results in very high web memory consumption node when the number jobs is very large 

### Current efforts
- [x] #5405 - Implement and backport ListAllJobs endpoint capability
- [x] #5429 - Rate limit API requests
- [x] #5466 - Retry ListAllJobs endpoint on 503 error
- [ ] #5556 - Allow showing job info on demand

Issue #5648, opened 2020-05-22
Archived Pipelines in Web UI


matthewpereira:
### Problem

- Concourse’s dashboard UI is a sophisticated front end experience for monitoring and interacting with pipelines, jobs, resources, and build logs. However, as customers push their Concourse usage to extreme scale, performance on the dashboard suffers.
- Specific roles require more advanced workflow possibilities to reduce the amount of manual effort involved in navigating the dashboard UI

### Proposed Solutions

- Advanced loading, caching, and updating strategies to improve speed and decrease bandwidth
- Pipeline favouriting, **archiving**, and sorting capabilities to reduce the toil and cognitive overload of navigating thousands of pipelines on a daily basis
- Build logs that load only the most pertinent info by default, with more available on demand
- Stripped down, operations-focused version of the dashboard for maximum performance

### Related efforts

See #5434 - Favourite Pipelines

### Current Status

- Pipeline caching, lazy loading implemented in 6.0.0
- Favorite Pipelines and Archiving Pipelines initiative research and design phase underway
- Archiving Pipelines back-end engineering work in flight

### Tasks

- [x] #5635 - Alter pipeline tile for archived pipelines
- [x] #5637 - Alter sidebar pipeline list item for archived pipelines
- [x] #5640 - Create toggle switch for archived pipelines

aoldershaw:
This has been implemented (see https://github.com/concourse/concourse/pull/5715), but it was easy enough to go with the "view selector" approach, so I went ahead and opened the following PRs:

* https://github.com/concourse/concourse/pull/5760
  * archived pipelines don't display action buttons on pipeline/build/job/resource views
  * drag and drop is fixed when there are archived pipelines
  * **note:** it does not implement #5635 
* https://github.com/concourse/concourse/pull/5761
  * view selector in sidebar
  * sidebar style changes per #5478 
  * **note:** selecting archived pipelines in the view selector does not modify what's displayed in the dashboard (it only filters in the sidebar). do we want to also change the filtering in the dashboard @matthewpereira (in which case we'd want #5635)?

PR #5831, opened 2020-06-29
Add flag to enable or disable public pipelines

maxknee:

closes # .

As our concourse instance grows, we've found that
by hiding public pipelines via a command flag,
we are able to increase the performance of our system.

This should be used when needed to hide pipelines when
a large amount of users don't need or want to see other
teams' exposed pipelines.

## Notes to reviewer:

By spinning up and setting up public pipelines, then logging out and seeing that they are not visible. 
## Contributor Checklist

- [X] [Signed] all commits
- [ ] Updated [Documentation]
- [X] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -130,6 +130,7 @@ type RunCommand struct {
 	ConcurrentRequestLimits   map[wrappa.LimitedRoute]int `long:"concurrent-request-limit" description:"Limit the number of concurrent requests to an API endpoint (Example: ListAllJobs:5)"`
 	APIMaxOpenConnections     int                         `long:"api-max-conns" description:"The maximum number of open connections for the api connection pool." default:"10"`
 	BackendMaxOpenConnections int                         `long:"backend-max-conns" description:"The maximum number of open connections for the backend connection pool." default:"50"`
+	DisablePublicPipelines    bool                        `long:"disable-public-pipelines" description:"Disable the loading of public pipelines on the dashboard"`
 
 	CredentialManagement creds.CredentialManagementConfig `group:"Credential Management"`
 	CredentialManagers   creds.Managers
@@ -543,6 +544,8 @@ func (cmd *RunCommand) Runner(positionalArguments []string) (ifrit.Runner, error
 
 		atc.LoadBaseResourceTypeDefaults(defaults)
 	}
+
+	atc.DisablePublicPipelines = cmd.DisablePublicPipelines
 
 	//FIXME: These only need to run once for the entire binary. At the moment,
 	//they rely on state of the command.
--- a/atc/dashboard_job.go
+++ b/atc/dashboard_job.go
@@ -1,0 +1,44 @@
+package atc
+
+import "time"
+
+type Dashboard []DashboardJob
+
+type DashboardJob struct {
+	ID           int
+	Name         string
+	PipelineName string
+	TeamName     string
+	Paused       bool
+	HasNewInputs bool
+
+	FinishedBuild   *DashboardBuild
+	NextBuild       *DashboardBuild
+	TransitionBuild *DashboardBuild
+
+	Inputs  []DashboardJobInput
+	Outputs []JobOutput
+
+	Groups []string
+}
+
+type DashboardBuild struct {
+	ID           int
+	Name         string
+	JobName      string
+	PipelineName string
+	TeamName     string
+	Status       string
+
+	StartTime time.Time
+	EndTime   time.Time
+}
+
+type DashboardJobInput struct {
+	Name     string
+	Resource string
+	Passed   []string
+	Trigger  bool
+}
+
+var DisablePublicPipelines bool
--- a/atc/db/job_factory.go
+++ b/atc/db/job_factory.go
@@ -231,11 +231,18 @@ func (j *jobFactory) VisibleJobs(teamNames []string) ([]atc.JobSummary, error) {
 	}
 
 	defer Rollback(tx)
+	var predPublic interface{}
 
-	dashboardFactory := newDashboardFactory(tx, sq.Or{
+	predPublic = sq.Or{
 		sq.Eq{"tm.name": teamNames},
 		sq.Eq{"p.public": true},
-	})
+	}
+
+	if atc.DisablePublicPipelines {
+		predPublic = sq.Eq{"tm.name": teamNames}
+	}
+
+	dashboardFactory := newDashboardFactory(tx, predPublic)
 
 	dashboard, err := dashboardFactory.buildDashboard()
 	if err != nil {
--- a/atc/db/pipeline_factory.go
+++ b/atc/db/pipeline_factory.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	sq "github.com/Masterminds/squirrel"
+	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db/lock"
 )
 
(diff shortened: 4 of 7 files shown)


clarafu:
Hi @maxknee, thanks for your work on this PR! I wanted to discuss with you a bit more about the problems you are facing around dashboard performance and cluttering of dashboard views with public pipelines.

We currently have some tracks of work that tackle those similar problems. We have had many performance improvements with the work done in https://github.com/concourse/concourse/issues/5712, I was wondering what version of Concourse you are running where you are still seeing performance problems with the dashboard?

For your other problem of not needing to see public pipelines from other teams in order to reduce clutter, I think that is a very reasonable use case because I also agree with it. Most of the time in multi-tenant environments, you will never need to care about other team's pipelines. I'm wondering rather than having it be a flag set on the ATC, that can be tackled with some changes to the dashboard UI? I want to pull in @matthewpereira because I know he can probably provide a lot more insight into if there is an existing issue that addresses this problem already or if not, he might have some suggestions on what it might look like.

maxknee:
Hey @clarafu, so we run a REALLY big instance of concourse, where we have 17000+ public jobs, and across 8 load balanced ATCs, we were having severe latency issues where our `listalljobs` endpoint would be 5+ seconds to load. we have 4952 pipelines and 1045 public pipelines. 
image (2)

It's not that the UI is cluttered per se, BUT the payload and then parsing of json for everyone using concourse became clear that there were a lot of issues with dashboard performance lending to a degraded user experience. 
I'd also like to point out that rate limiting would probably not help in this case because the experience would be non-existent for a large sum of our users. 

Let me know what you think and thanks for looking at my PR

mdb:
To further color some of what @maxknee has shared (we are on the same team), it's perhaps also worth noting that, in addition to dramatically improving HTTP latency (and thus the time required to render a usable homepage), hiding public pipelines yielded other positive performance benefits throughout the system, including...

* reduced ATC CPU consumption
* notably increased database throughput (queries performed per second)
* reduced database load

Unfortunately, I don't have Grafana graphs illustrating all of this in great detail, though here's a graph illustrating the stabilization of database load following our May 20th deployment of @maxknee 's patch disabling the inclusion of public pipelines in `listAllJobs`:

image

clarafu:
@maxknee Thanks for the detailed explanation of your use case! We also noticed that the ListAllJobs endpoint was causing a lot of problems so we did a lot of work to improve it within the v6.0 release. You can find all the changes here https://github.com/concourse/concourse/releases#60-dashboard but I think one that is worth noting is that the endpoint no longer needs to decode the job config which we found to be a big source of the problem with how much computational resources the endpoint takes up. We also made improvements on lazy loading the dashboard so that users will not be stuck waiting for all the jobs/pipelines to load up before displaying anything.

The reason why I'm trying to figure out if there are other ways to solve the problems you are facing is because it seems like your main problem is not that you want to only see public pipelines (which could be solved through some changes to the dashboard UI) but more that the dashboard is very costly and this PR is a way to reduce the amount of data it needs to load. I'm worried that if we merge this PR and the problem goes away, as your deployment continues to grow in the number of pipelines that each team has, it might get to a point where the dashboard needs to load up a large number of pipelines again and you will see the same problem with degrading dashboard performance as you do now. I am hoping that we can tackle the underlying problem of the high dashboard CPU consumption rather than a possibly temporary solution to the problem.

I'd like to ask you to try out the latest version of Concourse with the existing dashboard improvements and see if it helps fix the performance. If after upgrading you don't see enough improvements, we can discuss further how we can make the dashboard more performant! 😄

maxknee:
@clarafu excellent, but what about database performance improvement? we have a very large RDS instance and coupled with lazily loading and other optimizations, would that also have a benefit?

we're working on trying to replicate what we have on a smaller scale to test 6.3 to see if this patch is needed or not. (but performance testing is hard)

ndmmcglynn:
@clarafu thanks for your replies and we are happy to see new the performance improvements and will test them. Personally, I tend to agree with your point below and do think there is pretty substantial UX clutter when the number of public pipelines reaches the 1k+ scale. It is no longer helpful to users (especially new ones) to have so much unrelated information hitting them at once which is why having the ability to disable this would be advantageous. Curious to hear thoughts from @matthewpereira as well!

> For your other problem of not needing to see public pipelines from other teams in order to reduce clutter, I think that is a very reasonable use case because I also agree with it. Most of the time in multi-tenant environments, you will never need to care about other team's pipelines. I'm wondering rather than having it be a flag set on the ATC, that can be tackled with some changes to the dashboard UI? I want to pull in @matthewpereira because I know he can probably provide a lot more insight into if there is an existing issue that addresses this problem already or if not, he might have some suggestions on what it might look like.

clarafu:
@maxknee There were some database CPU improvements with v6.0.0 because of the changes to the algorithm component. You can read about it here https://github.com/concourse/concourse/releases#3602, it also shows a before and after metric of the database CPU. 

@ndmmcglynn Right I definitely agree with the use case of being able to hide public pipelines because of the clutter it introduces in the dashboard especially for multi-tenant environments. I'm hoping if v6.x helps fix your ATC/DB CPU consumption, we can shift the focus of this conversation onto the details of that feature instead (if you would be interested)!

matthewpereira:
Hey folks, thanks for your patience with this issue. I've been tied up in other work all week, and didn't notice the Github mention.

Now that we've made the 6.x.x improvements to the scheduling algorithm, the DOM performance in the web UI, and added a few patches to give more options for limiting the ListAllJobs endpoint, the next two big initiatives we have underway for decluttering and improving the front-end experience are:
* #5648 Archiving pipelines - this should help simplify what you're seeing on the dashboard unless you specifically want to look for an archived pipeline, making it faster to navigate to important pipelines as part of your everyday workflows
* #5434 Favorite pipelines - likewise, this should make it faster to navigate to important and frequently-visited pipelines

However, neither of these really do too much to decrease bandwidth usage or make the interface more *responsive*. To that end we have a new concept of a view selector in the works that could allow you to switch between viewing different *types* of pipelines - eg, show only the failed-state pipelines, only the paused-state pipelines, only the error-state pipelines, only the archived pipelines, only the favorite pipelines - and we could probably extend that, but I'm wondering what the hierarchy should be. 

I like the idea of being able to hide public pipelines from other teams as well, but I'll have to think of the best place for this sort of switch or toggle to live in the UI. Should this be another *type* of pipeline that we can filter by, or would you folks think it's more useful for example to be able to `Filter to show only [Failed|Errored|Archived|Paused] Pipelines` and then `Show/Hide other teams' public pipelines` as two separate levels of filtering?

maxknee:
> Hey folks, thanks for your patience with this issue. I've been tied up in other work all week, and didn't notice the Github mention.
> 
> Now that we've made the 6.x.x improvements to the scheduling algorithm, the DOM performance in the web UI, and added a few patches to give more options for limiting the ListAllJobs endpoint, the next two big initiatives we have underway for decluttering and improving the front-end experience are:
> 
> * #5648 Archiving pipelines - this should help simplify what you're seeing on the dashboard unless you specifically want to look for an archived pipeline, making it faster to navigate to important pipelines as part of your everyday workflows
> * #5434 Favorite pipelines - likewise, this should make it faster to navigate to important and frequently-visited pipelines
> 
> However, neither of these really do too much to decrease bandwidth usage or make the interface more _responsive_. To that end we have a new concept of a view selector in the works that could allow you to switch between viewing different _types_ of pipelines - eg, show only the failed-state pipelines, only the paused-state pipelines, only the error-state pipelines, only the archived pipelines, only the favorite pipelines - and we could probably extend that, but I'm wondering what the hierarchy should be.
> 
> I like the idea of being able to hide public pipelines from other teams as well, but I'll have to think of the best place for this sort of switch or toggle to live in the UI. Should this be another _type_ of pipeline that we can filter by, or would you folks think it's more useful for example to be able to `Filter to show only [Failed|Errored|Archived|Paused] Pipelines` and then `Show/Hide other teams' public pipelines` as two separate levels of filtering?

^ I do like having extra filtering. If there's a way to ensure that we can set a default filtering as well, just to give our instance a some breathing room.

Haegi:
Is there any progress on this one?

xtremerui:
@clarafu could you provide an update please?

clarafu:
I'm going to close this PR for now because I think it was trying to solve a performance problem that will probably be better solved in a different way (either through performance optimizations or through a filter in the dashboard web UI). Feel free to open the issue if you feel differently and want to continue to work on this PR!

clarafu:
I'm going to close this PR for now because I think it was trying to solve a performance problem that will probably be better solved in a different way (either through performance optimizations or through a filter in the dashboard web UI). Feel free to open the issue if you feel differently and want to continue to work on this PR!

PR #5833, opened 2020-06-30
atc: renaming pipeline resources

### LINKED ISSUE #5811: Renaming pipeline resources
## Summary

Jobs can be renamed by configuring an [`old_name`](https://concourse-ci.org/jobs.html#schema.job.old_name) and configuring the pipeline. Afterwards, the `old_name` field can be removed.

This issue is for implementing the same functionality, but for resources. This way the data for disabled and pinned versions can be preserved through changing the resource name.

## Context

* Needed by #5810

mouellet:

closes #5811 .

- Add an `old_name` attribute to `atc.ResourceConfig`
- Add `team.updateResourcesName()` method for updating resources new names 
- Rename `team.updateNames()` to `team.updateJobsName()` & remove references to jobs in `team.checkCyclic()` and `team.sortUpdateNames()`

## Notes to reviewer:

Step's name or resource tag must also be updated with new resource name or pipeline validation will fail.

## Contributor Checklist

- [x] [Signed] all commits
- [x] Updated [Documentation]
- [x] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


atc/db/team_test.go
```diff
@@ -2161,6 +2161,262 @@ var _ = Describe("Team", func() {
...
+
+					versions, _, found, err := resource.Versions(db.Page{Limit: 3}, nil)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(found).To(BeTrue())
+					Expect(versions).To(HaveLen(1))
+					Expect(versions[0].Version).To(Equal(atc.Version{"disabled": "version"}))
+					Expect(versions[0].Enabled).To(BeFalse())
+				})
+
+				It("should successfully update resource name", func() {
```
aoldershaw, line 2323: Maybe this should test that the version is still disabled after the rename, rather than checking that the IDs match up?

atc/db/team_test.go
```diff
@@ -2161,6 +2161,262 @@ var _ = Describe("Team", func() {
...
+					Expect(err).ToNot(HaveOccurred())
+					Expect(found).To(BeTrue())
+					resID := resConf.ID()
+
+					found, err = resource.PinVersion(resID)
+					Expect(found).To(BeTrue())
+					Expect(err).ToNot(HaveOccurred())
+				})
+
+				It("should successfully update resource name", func() {
```
aoldershaw, line 2393: This tests the right thing (i.e. that the `APIPinnedVersion()` remains unchanged), but maybe you should change the test description? e.g.

```suggestion
				It("should not change the pinned version", func() {
```

testflight/rename_resource_test.go
```diff
@@ -0,0 +1,34 @@
...
+		Expect(build).To(gbytes.Say(pipelineName + "/without-params"))
+
+		setPipeline(
+			"fixtures/rename-resource-with-params.yml",
+			"-v", "unique_version="+guid.String(),
+		)
+
+		fly("trigger-job", "-j", inPipeline("without-params"), "-w")
+		build = fly("builds", "-p", pipelineName)
+		Expect(build).To(gbytes.Say(pipelineName + "/without-params"))
```
aoldershaw, line 32: Looks good, but maybe you could also test that it retains version history? e.g. something like

```go
resourceVersions := fly("resource-versions", "-r", inPipeline("new-git-resource"))
Expect(resourceVersions).To(gbytes.Say(originalGUID.String()))
```

In which case you'd also want to generate a new `guid` when you set the pipeline for the second time (that way, the original guid will only be a resource version if renaming the resource didn't clear the version history)

mouellet:
Hey @aoldershaw, I changed the resource's `atc.ResourceConfig` construction like you asked and a test on `pipeline.Archive()` fails... It expects a `atc.ResourceConfig` with `name` and `type`. Those where taken from the `resource` when they were passed get passed on as-is in constructor.

What's the impact of changing this? I can't see why the `emptyResourceConfigs` is not actually empty...

https://github.com/concourse/concourse/blob/a83786990a44e186fe71d1d03ae2bba2a3e62af1/atc/db/pipeline_test.go#L271-L274

Thanks

aoldershaw:
@mouellet when a pipeline is archived, it clears out the `config` column from the `resources` DB table. The `resources` table also has a `name` and `type` field (these are stored in the `config`, but are top-level fields as well).

Prior to the change you made, the `atc.ResourceConfig` was constructed mostly from fields from the `config`, but `Name` and `Type` came from the corresponding top-level database fields (which don't get cleared out when a pipeline is archived) - so that's why they're non-empty.

In this case, it's perfectly valid to have them be truly empty, though - that's actually the test we make for clearing out the jobs config when the pipeline is archived:

https://github.com/concourse/concourse/blob/a83786990a44e186fe71d1d03ae2bba2a3e62af1/atc/db/pipeline_test.go#L256-L263

So, I'd recommend modifying the test to look like the jobs test above.

In terms of impact, I don't think there should be any - `db.Resources.Configs()` is only ever called by `db.Pipeline.Config()` (other than some tests), and `db.Pipeline.Config()` is never called if the pipeline is archived (the API returns a 404 if you request the config of an archived pipeline).

testflight/rename_resource_test.go
mouellet: Had to re-work that test, gave me the chance to learned about the mock resource!

mouellet:
Thanks @aoldershaw, it's great and reassuring to have people who know the code base!

About the `db.resource` struct... I removed the top-level fields (those not in the `resources` table) and made the accessors call the `resource.Config()` method instead! I feared those top-level fields might not be in sync somehow and didn't want to decrypt and unmarshal at 2 places (in `db.scanResource` and in `resources.Config()`)

I can always revert if you think it goes too far, just let me know!

aoldershaw:
@mouellet awesome, thanks for the quick changes! The one thing I don't love about removing the decryption+unmarshaling of the config in `db.scanResource` is that a malformed/corrupted resource config won't necessarily cause an error anymore (since the getters just ignore the error) - it would error in the case where we call `Resource.Config()` directly, but not if we call e.g. `Resource.Public()`.

What if `db.scanResource` always decrypts+unmarshals the config (like we did before) and `Resource.Config()` just returns that config? The config lazy-loading you added is cool, but I think we pretty much always need to access fields in the config when we load a resource, so it's probably better to just do it up-front (that way, errors will always surface).

Then, the getters would just return e.g. `r.config.Public`, rather than having to call `r.Config()` and checking on the error. I do 

What do you think?

mouellet:
@aoldershaw I think we found the best solution!

Issue #5407, opened 2020-04-07
containerd: organize containerd flags prior to making containerd GA


muntac:
Currently, we're using some existing `CONCOURSE_GARDEN_*` flags for configuring the `containerd` runtime. This makes it unclear to the user which flags are actually used by `containerd`, and which have no effect. This might allow for ambiguous configurations that could lead to edge cases that are difficult to debug (e.g. an unused flag in a user's deployment in the past caused unpredictable behaviour and was hard to track down).

Some ideas to consider: 

- have separate set of flags for Guardian vs containerd runtimes
- ~have common flags for functionality common to all runtimes under the Concourse namespace. e.g. for setting custom DNS server use `CONCOURSE_DNS_SERVER`~
- having default values wherever possible to make it easy to get Concourse up and running
- have validation that prevents user from specifying Guardian and containerd flags simultaneously

--------------------------
EDIT:

**Requirements**

Add additional `containerd` specific flags (preferably prefixed `CONCOURSE_CONTAINERD`) for the following `CONCOURSE_GARDEN` flags:
- [ ] [CONCOURSE_GARDEN_DNS_PROXY_ENABLE](https://github.com/concourse/concourse/blob/88a99ac185df381987cb51a54a0f4d7ffbdf73ff/worker/workercmd/containerd.go#L151) - support added in https://github.com/concourse/concourse/pull/5465

- [ ] [CONCOURSE_GARDEN_DNS_SERVER](https://github.com/concourse/concourse/blob/88a99ac185df381987cb51a54a0f4d7ffbdf73ff/worker/workercmd/worker_linux.go#L31) - support added in https://github.com/concourse/concourse/pull/5445

- [ ] [CONCOURSE_GARDEN_DENY_NETWORK] - support added in #5829 

- [ ] [CONCOURSE_GARDEN_MAX_CONTAINERS] - support added in #5836 

Prevent mixed usage of Guardian and containerd flags
- [ ] Ensure [CONCOURSE_CONTAINER_NETWORK_POOL](https://github.com/concourse/concourse/blob/88a99ac185df381987cb51a54a0f4d7ffbdf73ff/worker/workercmd/containerd.go#L31) (support added in https://github.com/concourse/concourse/pull/5486) only works for `containerd`. https://github.com/concourse/concourse/pull/5486#issuecomment-620039187 mentioned that this is currently working when using Guardian as well.

[...]

PR #5836, opened 2020-06-30
5226 max container limit

### LINKED ISSUE #5226: containerd: support max containers limit

In https://github.com/concourse/concourse/pull/5225, we can support setting CPU and memory limits according to the Garden spec. However, `gdn` supports going beyond the Garden spec for resource limits. In particular, we should add support for the `max-containers` flag.

https://github.com/cloudfoundry/guardian/blob/7818ab761574cbcf345f9ac33f0c6d8cba9f376f/guardiancmd/command.go#L176

Support this functionality so that our containerd backend achieves parity with `gdn`

xtreme-sameer-vohra:
closes #5226 

We did some simple benchmarking
## Setup
Create 100 containers in parallel + backend setup and teardown
4x runs
 
### Without setting a limit
Avg - 31s
### With setting a limit of 100 and a timeout value of 5m
Avg - 30s

We don't expect the code to be faster than not setting the limit. However, we saw that the locking logic doesn't incur a severe time penalty.

## Contributor Checklist

- [x] [Signed] all commits
- [x] Updated [Documentation]

## Reviewer Checklist

- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

## Acceptance Tests
1. Deploy Concourse Worker with flag (eg. CONCOURSE_GARDEN_MAX_CONTAINERS) with a non zero value
1. Configure a pipeline job that would require more containers than the limit
1. See the steps error due to hitting the max container limit


worker/runtime/backend.go
```diff
@@ -28,6 +29,10 @@ type GardenBackend struct {
 	network       Network
 	rootfsManager RootfsManager
 	userNamespace UserNamespace
+
+	maxContainers  int
+	requestTimeout time.Duration
+	createLock     *semaphore.Weighted
```
aoldershaw, line 35: FWIW, when I suggested using semaphores here https://github.com/concourse/concourse/issues/5226#issuecomment-642094627, I didn't have a mutex in mind. I was thinking more along the lines of:

* On startup, check the number of containers by querying `containerd`
* Initialize a semaphore with value `MaxContainerLimit - NumContainersOnStartup`
* Whenever you create a container, acquire the semaphore
* Whenever you destroy a container, release the semaphore

If it's not too much of a performance burden querying `containerd` for the list of containers whenever you start a container, the approach you took is definitely safer, since we don't have to worry about containers going away without us releasing the semaphore. Figured I'd bring up the alternative approach, anyway

chenbh: We were thinking about that approach and @xtreme-sameer-vohra brought up the concern that there could be issues if the worker binary ever gets out of sync with containerd. In order to rectify that we would have to periodically query number of containers and update our count. But we would not be able to use `semaphore.Weighted` for that as it does not expose the count / weight, and if we implement our own, it eventually leads down a rabbit hole of details.

The only reason we used semaphores was that we wanted to put a timeout on lock acquisition and the semaphore is the only synchronization primitive that took a `context.Context`.

xtreme-sameer-vohra: Daniel covered the reasons for our approach - primarily making sure we're in sync
This was also the approach that the current [Guardian implementation uses](https://github.com/cloudfoundry/guardian/blob/6b021168907b2f2ae25cf54acae8b454b430332f/gardener/gardener.go#L230-L241) - albeit without a lock.

worker/workercmd/worker_linux.go
```diff
@@ -29,9 +29,11 @@ type GardenBackend struct {
 	Bin        string    `long:"bin"        description:"Path to a garden backend executable (non-absolute names get resolved from $PATH)."`
 	Config     flag.File `long:"config"     description:"Path to a config file to use for the Garden backend. Guardian flags as env vars, e.g. 'CONCOURSE_GARDEN_FOO_BAR=a,b' for '--foo-bar a --foo-bar b'."`
 	DNSServers []string  `long:"dns-server" description:"DNS server IP address to use instead of automatically determined servers. Can be specified multiple times."`
-	DNS            DNSConfig     `group:"DNS Proxy Configuration" namespace:"dns-proxy"`
+	DNS        DNSConfig `group:"DNS Proxy Configuration" namespace:"dns-proxy"`
 
 	RequestTimeout time.Duration `long:"request-timeout" default:"5m" description:"How long to wait for requests to Garden to complete. 0 means no timeout."`
```
aoldershaw, line 34: This flag is now responsible for more than just requests to Garden, now

Might it make sense to split it up into a separate flag?

aoldershaw:
I tried setting `CONCOURSE_GARDEN_MAX_CONTAINERS` to `2`, and ran the following pipeline:

```yaml
jobs:
- name: job
  plan:
  - in_parallel:
    - task: echo
      config: &task
        platform: linux
        image_resource:
          type: registry-image
          source: {repository: busybox}
        run:
          path: echo
          args: [hey]
    - task: echo2
      config: *task
    - task: echo3
      config: *task
```

When I run that job, all 3 `in_parellel` steps error with `max containers reached`. Running `fly -t dev containers` shows some `check` and `get` containers, as well as the `task` containers (despite there being no resources anywhere? are those checks for the base resource types?)

Also, side note - any idea why the task containers that failed to start are showing up in `fly containers`?

worker/workercmd/worker_linux.go
chenbh: I believe there's a larger story at moving all the containerd flags out of the `GARDEN` namespace at https://github.com/concourse/concourse/issues/5407

jamieklassen:
@aoldershaw I'd expect to see check containers for those anonymous resources defined by the `image_resource:` block of a task config.

chenbh:
> When I run that job, all 3 in_parellel steps error with max containers reached. Running fly -t dev containers shows some check and get containers, as well as the task containers (despite there being no resources anywhere? are those checks for the base resource types?)

Armchair-ing here but could that be the registry image resource? We have to get the image from somewhere and it feels like it could be modelled as an implicit `get` step?

aoldershaw:
@jamieklassen @chenbh good points, thanks. Not sure if the task containers showing up in `fly containers` is something unexpected or not, though it seems kind of weird at the very least. Haven't checked if this also happens with Guardian yet

worker/runtime/backend.go
```diff
@@ -140,6 +164,33 @@ func (b *GardenBackend) Ping() (err error) {
...
+
+	return NewContainer(
+		cont,
+		b.killer,
+		b.rootfsManager,
+	), nil
+}
+
+func (b *GardenBackend) createContainer(ctx context.Context, gdnSpec garden.ContainerSpec) (containerd.Container, error) {
+	acquired, err := b.checkContainerCapacity(ctx)
```
muntac, line 185: @jamieklassen and I were discussing this. We think separating the responsibility of acquiring the lock from checking the container limit might be more readable. 

In this case we would always acquire the lock and defer the release. Then separately check the container limit (whether we're below the limit OR if there's no limit).

xtreme-sameer-vohra:
> I tried setting `CONCOURSE_GARDEN_MAX_CONTAINERS` to `2`, and ran the following pipeline:
> 
> ```yaml
> jobs:
> - name: job
>   plan:
>   - in_parallel:
>     - task: echo
>       config: &task
>         platform: linux
>         image_resource:
>           type: registry-image
>           source: {repository: busybox}
>         run:
>           path: echo
>           args: [hey]
>     - task: echo2
>       config: *task
>     - task: echo3
>       config: *task
> ```
> 
> When I run that job, all 3 `in_parellel` steps error with `max containers reached`. Running `fly -t dev containers` shows some `check` and `get` containers, as well as the `task` containers (despite there being no resources anywhere? are those checks for the base resource types?)
> 
> Also, side note - any idea why the task containers that failed to start are showing up in `fly containers`?

The result of `fly containers` is state from the DB. The state of the container would be useful to distinguish whether it actually exists or is desired and concourse is working on making that happen.

worker/runtime/backend_test.go
```diff
@@ -141,9 +143,123 @@ func (s *BackendSuite) TestCreateContainerSetsHandle() {
...
+func (s *BackendSuite) TestCreateContainerLockTimeout() {
+	fakeTask := new(libcontainerdfakes.FakeTask)
+	fakeContainer := new(libcontainerdfakes.FakeContainer)
+
+	fakeContainer.IDReturns("handle")
+	fakeContainer.NewTaskReturns(fakeTask, nil)
+	s.client.NewContainerReturns(fakeContainer, nil)
+
+	network := new(runtimefakes.FakeNetwork)
+	network.SetupMountsStub = func(string) ([]specs.Mount, error) {
```
muntac, line 225: It might make more semantic sense to introduce the delay in `s.client.NewContainerReturns(fakeContainer, nil)` which actually creates the container.

worker/runtime/backend.go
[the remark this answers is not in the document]
aoldershaw: (I'm speculating that) the difficulty is that we don't want to acquire the lock if there is no container limit

worker/workercmd/worker_linux.go
aoldershaw: True, but I guess my point is that the flag is being used for two things - one is related to garden (and makes sense to be in the `GARDEN` namespace), while the other is containerd specific (and doesn't). But maybe we can address this in #5407

worker/runtime/backend.go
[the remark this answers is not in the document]
jamieklassen: but @aoldershaw it's worth exploring the cost of serializing container creation in this way -- like how expensive is the redundant locking in the "no limit" case?

worker/workercmd/worker_linux.go
xtreme-sameer-vohra: Yes, the flag is currently used for both `containerd` & `garden` and the reason was to limit scope to this PR to having a means of configuring max container limit for `containerd`. 
The scope of #5407 is to look at all flags that are used commonly between the 2 runtimes and dealing with them en masse.

worker/runtime/backend_test.go
xtreme-sameer-vohra: fixed

worker/runtime/backend.go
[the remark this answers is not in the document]
xtreme-sameer-vohra: From our final pseudo benchmarking, there didn't appear to be any noticeable overhead with acquisition under serial creation of containers, so the design is worth re-visiting if we can improve readability.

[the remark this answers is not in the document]
xtreme-sameer-vohra: refactored backend.go based on the suggestions.

muntac:
@xtreme-sameer-vohra @chenbh was looking at a stale version of this page and hadn't noticed the merge conflicts. Took a look. Just needs some clearing up in the flag name section and a function call as `deny-networks` stuff goes in the same places.

muntac:
The `validate-labels` test on CI is new and is failing because the validation script does not have the right permissions. I ran it manually with the right permissions and it asked for an relevant label for the PR, which I've gone ahead and added.

Screen Shot 2020-07-16 at 2 22 14 PM

muntac:
@xtreme-sameer-vohra @chenbh I'm still okay with this approach. However, after reading a bit about `select` in `go` I thought of sketching out a different approach that doesn't require a lock. Let me know if you have any thoughts about the readability of this.

```
type GardenBackend struct {
...
        // Signifies a turn.
	allowCreate chan struct{}
...
}

func NewGardenBackend(client libcontainerd.Client, opts ...GardenBackendOpt) (b GardenBackend, err error) {
...
	b.allowCreate = make(chan struct{})
...
}

func (b *GardenBackend) Start() (err error) {
...
        if b.maxContainers == 0 {
		b.allowCreate = nil
	} else {
		go func() {
			b.allowCreate <- struct {}{}
		}()
	}
...
}

// Create creates a new container.
//
func (b *GardenBackend) Create(gdnSpec garden.ContainerSpec) (garden.Container, error) {
	ctx := context.Background()
        
       if b.maxContainers == 0 {
		// create container as usual
	}

	select {
	case <-time.After(b.requestTimeout):
		return nil, fmt.Errorf("timeout waiting for turn to create container")
	case <-b.allowCreate:
                // Refill channel after this attempt is done to unblock others
		defer func() {
		  b.allowCreate <- struct{}{}
		}()
		if b.checkContainerCapacity(ctx) == nil {
			cont, err := b.createContainer(ctx, gdnSpec)
			err = b.startTask(ctx, cont)
			if err != nil {
				return nil, fmt.Errorf("starting task: %w", err)
			}

			return NewContainer(
				cont,
				b.killer,
				b.rootfsManager,
			), nil
		}
	}
	return nil, nil
}
```

xtreme-sameer-vohra:
> @xtreme-sameer-vohra @chenbh I'm still okay with this approach. However, after reading a bit about `select` in `go` I thought of sketching out a different approach that doesn't require a lock. Let me know if you have any thoughts about the readability of this.
> 
> ```
> type GardenBackend struct {
> ...
>         // Signifies a turn.
> 	allowCreate chan struct{}
> ...
> }
> 
> func NewGardenBackend(client libcontainerd.Client, opts ...GardenBackendOpt) (b GardenBackend, err error) {
> ...
> 	b.allowCreate = make(chan struct{})
> ...
> }
> 
> func (b *GardenBackend) Start() (err error) {
> ...
>         if b.maxContainers == 0 {
> 		b.allowCreate = nil
> 	} else {
> 		go func() {
> 			b.allowCreate <- struct {}{}
> 		}()
> 	}
> ...
> }
> 
> // Create creates a new container.
> //
> func (b *GardenBackend) Create(gdnSpec garden.ContainerSpec) (garden.Container, error) {
> 	ctx := context.Background()
>         
>        if b.maxContainers == 0 {
> 		// create container as usual
> 	}
> 
> 	select {
> 	case <-time.After(b.requestTimeout):
> 		return nil, fmt.Errorf("timeout waiting for turn to create container")
> 	case <-b.allowCreate:
>                 // Refill channel after this attempt is done to unblock others
> 		defer func() {
> 		  b.allowCreate <- struct{}{}
> 		}()
> 		if b.checkContainerCapacity(ctx) == nil {
> 			cont, err := b.createContainer(ctx, gdnSpec)
> 			err = b.startTask(ctx, cont)
> 			if err != nil {
> 				return nil, fmt.Errorf("starting task: %w", err)
> 			}
> 
> 			return NewContainer(
> 				cont,
> 				b.killer,
> 				b.rootfsManager,
> 			), nil
> 		}
> 	}
> 	return nil, nil
> }
> ```

Hey @muntac 
Is there a reason we'd want to avoid using a lock and use a channel instead ?
IMHO,  in this case, we want to ensure that only one create request is in flight at any time t, which a lock models more naturally.

While we can use a channel to model the required functionality;
- we don't leverage the message body `struct {}{}` 
- one consumer has to know about another consumer 
   ```
    // Refill channel after this attempt is done to unblock others
		defer func() {
		  b.allowCreate <- struct{}{}
		}()
   ```

muntac:
@xtreme-sameer-vohra Sorry, I meant to write that call as a goroutine so it doesn't block. The consumer in this case will be whoever is calling Create() next, or is already inside a Create() call and waiting for a receive on that channel.

```
// Refill channel after this attempt is done to unblock others
 	defer func() {
 	  go func(){
               b.allowCreate <- struct{}{}
          }()
 	}()
```

Empty structs are [recommended](https://golangbyexample.com/empty-struct-in-go/) whenever a channel is being used only for a notification, and the content of the channel doesn't matter. In this case the notification is "it is safe to attempt creating a container". Because it's an unbuffered channel only one request can receive on this channel at a time, and after receiving they do a send. This mimics the acquire and release functionality, respectively, of a lock.

I brought this up as an alternative to locking as it uses a built in construct of the language to help us manage concurrent access to a codepath. This way we don't need to implement and manage our own lock. Have not thought through how easy this would be to test.

xtreme-sameer-vohra:
> @xtreme-sameer-vohra Sorry, I meant to write that call as a goroutine so it doesn't block. The consumer in this case will be whoever is calling `Create()` next, or is already inside a `Create()` call and waiting for a receive on that channel.
> 
> ```
> // Refill channel after this attempt is done to unblock others
>  	defer func() {
>  	  go func(){
>                b.allowCreate <- struct{}{}
>           }()
>  	}()
> ```
> 
> Empty structs are [recommended](https://golangbyexample.com/empty-struct-in-go/) whenever a channel is being used only for a notification, and the content of the channel doesn't matter. In this case the notification is "it is safe to attempt creating a container". Because it's an unbuffered channel only one request can receive on this channel at a time, and after receiving they do a send. This mimics the acquire and release functionality, respectively, of a lock.
> 
> I brought this up as an alternative to locking as it uses a built in construct of the language to help us manage concurrent access to a codepath. This way we don't need to implement and manage our own lock. Have not thought through how easy this would be to test.

Thats fair, by using channels, we wouldn't need to add another import "golang.org/x/sync/semaphore". However, IMHO the sync package functionality better maps to our requirements. 
Also, we don't implement the lock, we simply wrap the lock provided by the sync package to be a no-op in scenarios where the max container limit is not set.

muntac:
That's a good point about wrapping and not implementing. Thanks for entertaining the idea! Feel free to go ahead and merge.

PR #5837, opened 2020-07-01
fly: fix fly login manual input on windows

aoldershaw:

There's more context in the first commit message, but tl;dr the library we switched to for reading in the token on stdin didn't behave so well on Windows (at least on git bash and in the test suite, seemed fine in Powershell).

The library tried to be smart about things based on the width of the terminal (because the library does smart things), but Windows was reporting that the terminal had 0 width for some reason. However, we don't really *need* a library that does smart things - we just need to read a line of text from the terminal and hide the output. 

* Ditch peterh/liner in favour of using raw mode ourselves and reading the input character-by-character, dealing with some special characters (ctrl+C, backspace, new line/carriage return).

## Notes to reviewer:

This doesn't shouldn't change the behaviour noticeably on Linux/Mac OS - only on Windows.

Also, putting the terminal into raw mode on Windows Git Bash also seems to be a no-go, so the input won't be hidden there (but will be elsewhere).

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/fly/commands/login.go
+++ b/fly/commands/login.go
@@ -115,7 +115,8 @@ func (command *LoginCommand) Execute(args []string) error {
 		return err
 	}
 
-	if pty.IsTerminal() && !command.BrowserOnly {
+	isRawMode := pty.IsTerminal() && !command.BrowserOnly
+	if isRawMode {
 		state, err := terminal.MakeRaw(int(os.Stdin.Fd()))
 		if err != nil {
 			return err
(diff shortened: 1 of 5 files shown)


aoldershaw:
@vito though raw mode doesn't seem to work on git bash (at least on my virtual desktop), if there is a character limit it seems to be pretty large on windows - put ~15000 characters into it and it didn't truncate at all.

aoldershaw:
@vito awesome, thanks! did you run into the same problem on GitBash (input wasn't hidden because enabling raw mode failed)?

vito:
@aoldershaw Yeah - I think it's fine, since at least the message was updated to not include the ` (output will be hidden)` part, and we didn't hide it before anyhow.

PR #5838, opened 2020-07-02
Refactor Artifact Source - remove logger

xtreme-sameer-vohra:
Simplify the interface

Remove logger which wasn't used

## Contributor Checklist
- [x] [Signed] all commits

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/worker/artifact_source.go
+++ b/atc/worker/artifact_source.go
@@ -30,11 +30,11 @@ type StreamableArtifactSource interface {
 	// StreamTo copies the data from the source to the destination. Note that
 	// this potentially uses a lot of network transfer, for larger artifacts, as
 	// the ATC will effectively act as a middleman.
-	StreamTo(context.Context, lager.Logger, ArtifactDestination) error
+	StreamTo(context.Context, ArtifactDestination) error
 
 	// StreamFile returns the contents of a single file in the artifact source.
 	// This is used for loading a task's configuration at runtime.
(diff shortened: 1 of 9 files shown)


aoldershaw:
Seems like a reasonable change, especially considering that if we need a logger, we can pass it through the context using `lagerctx`.

Is there a reason we don't have any logging here (even just at the `DEBUG` level)? Would that be useful for observability's sake? Looks like we have tracing in `StreamTo`, but nothing in `StreamFile`

xtreme-sameer-vohra:
> Seems like a reasonable change, especially considering that if we need a logger, we can pass it through the context using `lagerctx`.
> 
> Is there a reason we don't have any logging here (even just at the `DEBUG` level)? Would that be useful for observability's sake? Looks like we have tracing in `StreamTo`, but nothing in `StreamFile`

Yep, like you said we can introduce it again if we need to log anything. 
Completely agree, that it would be valuable to have observability specifically for the worker <-> worker volume streaming case. However, we haven't had logging for it so far and it hasn't been missed. Also, the tracing setup offers much more usable visualization.

kcmannem:
we use to have it, before the refactor I worked on. Never actually used it in practice hence I left the TODO to remove the stuff. There are other log indicators that you can use to figure out if streamings happening (from what I remember).

PR #5842, opened 2020-07-03
Add recover to concourse core to prevent panic

### LINKED ISSUE #3594: Panics in pipeline/build scheduling should not bring down the whole Concourse

xtremerui:

Feature 

closes #3594.

Add `recover()` to go routines around core scheduler, lidar and build engine. It will capture the panic and log it in stderr and logger.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/builds/tracker.go
+++ b/atc/builds/tracker.go
@@ -2,6 +2,10 @@ package builds
 
 import (
 	"context"
+	"fmt"
+	"os"
+	"runtime/debug"
+	"strconv"
 	"sync"
 
 	"code.cloudfoundry.org/lager"
@@ -46,6 +50,22 @@ func (bt *Tracker) Run(ctx context.Context) error {
 	for _, b := range builds {
 		if _, exists := bt.running.LoadOrStore(b.ID(), true); !exists {
 			go func(build db.Build) {
+				loggerData := lager.Data{
+					"build":    strconv.Itoa(build.ID()),
+					"pipeline": build.PipelineName(),
+					"job":      build.JobName(),
+				}
+				defer func() {
+					if r := recover(); r != nil {
+						err = fmt.Errorf("panic in tracker build run %s: %v", loggerData, r)
+
+						fmt.Fprintf(os.Stderr, "%s\n %s\n", err.Error(), string(debug.Stack()))
+						logger.Error("panic-in-tracker-build-run", err)
+
+						build.Finish(db.BuildStatusErrored)
+					}
+				}()
+
 				defer bt.running.Delete(build.ID())
 
 				metric.BuildsRunning.Inc()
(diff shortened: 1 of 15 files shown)


xtremerui:
@taylorsilva the differences for each handling of panic is one reason. The other important reason that we have do recover() colocated is because recover() has to be exact one stack up of the panic goroutine, which means putting the recover() in a recover/recover.go and invoking it from the panic goroutine doesn't work. It wont be able to catch the panic.

xtremerui:
@vito not sure if we want to have a release note for it or just update operation wiki to let users know to look for panic error in stderr?

vito:
@taylorsilva @xtremerui I wonder how `defer GinkgoRecover()` works? 🤔 



> @vito not sure if we want to have a release note for it or just update operation wiki to let users know to look for panic error in stderr?

I think this is definitely worth a release note - operators that have had to deal with a panicking cluster caused by a build would love to hear they don't need to worry anymore.

xtremerui:
@vito sry i wasn't very clear.

Our case is we need to get the reformed panic error to send it back to a channel or set for check/build run error. By running `defer Recover()` we cant get the error back so we tried
```
defer func() {
   error = OurRecover()
}()
```
where
```
func OurRecover() error {
   if r:= recover(); r!= nil{
      return r
    }
}
```
but it cant catch the panic in this case. Regarding all the difference we needs to handle the panic error and the above behaviour we then decided just have recover() colocated to where the goroutine runs.

xtremerui:
@vito sry i wasn't very clear.

Our case is we need to get the reformed panic error to send it back to a channel or set for check/build run error. By running defer Recover() we cant get the error back so we tried

defer func() {
   error = OurRecover()
}()
where

func OurRecover() error {
   if r:= recover(); r!= nil{
      return r
    }
}
but it cant catch the panic in this case. Regarding all the difference we needs to handle the panic error and the above behaviour we then decided just have recover() colocated to where the goroutine runs.

vito:
@xtremerui Ah, right, I didn't see that we actually do something with the result, so having a throwaway recover-er doesn't buy you a whole lot.

xtremerui:
@taylorsilva i added the release note.

Issue #5795, opened 2020-06-19
Show worker name on build page


evanchaoli:
## Summary

@vito As we discussed in the meeting, I was going to reopen #4519 to track the story, but the PR had been deleted, no way to reopen, thus I create a new issue. Please advice how to proceed.

## Context

Our worker VMs are distributed to multiple vSphere clusters, to debug some intermittent failures, e.g. transient network breakages, we have to know on which worker a step ran on, so that we can further identify on which vSphere cluster the failure occurred.

But now it's very hard to find out worker name, thus we want to request a feature to show worker name on build page.

* RFC: concourse/rfcs#
* Prior discussion: #4519
* Depends on #
* Part of #

PR #5846, opened 2020-07-06
feat: added a new event type SelectedWorker

evanchaoli:
Fixes: #5795

This PR adds an opt-in feature where worker name can be injected into get/put step's metadata. 

## Release Note

The build log page and `fly watch` now display the worker name for `get/put/task` steps

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/engine/builder/delegate_factory.go
+++ b/atc/engine/builder/delegate_factory.go
@@ -506,6 +506,17 @@ func (delegate *buildStepDelegate) Finished(logger lager.Logger, succeeded bool)
 	logger.Info("finished")
 }
 
+func (d *buildStepDelegate) SelectedWorker(logger lager.Logger, workerName string) {
+	err := d.build.SaveEvent(event.SelectedWorker{
+		Time:       time.Now().Unix(),
+		WorkerName: workerName,
+	})
+	if err != nil {
+		logger.Error("failed-to-save-selected-worker-event", err)
+		return
+	}
+}
+
 func (delegate *buildStepDelegate) Errored(logger lager.Logger, message string) {
 	err := delegate.build.SaveEvent(event.Error{
 		Message: message,
--- a/atc/engine/builder/delegate_factory.go
+++ b/atc/engine/builder/delegate_factory.go
@@ -506,9 +506,12 @@ func (delegate *buildStepDelegate) Finished(logger lager.Logger, succeeded bool)
 	logger.Info("finished")
 }
 
-func (d *buildStepDelegate) SelectedWorker(logger lager.Logger, workerName string) {
-	err := d.build.SaveEvent(event.SelectedWorker{
-		Time:       time.Now().Unix(),
+func (delegate *buildStepDelegate) SelectedWorker(logger lager.Logger, workerName string) {
+	err := delegate.build.SaveEvent(event.SelectedWorker{
+		Time: time.Now().Unix(),
+		Origin: event.Origin{
+			ID: event.OriginID(delegate.planID),
+		},
 		WorkerName: workerName,
 	})
 	if err != nil {
--- a/atc/worker/client.go
+++ b/atc/worker/client.go
@@ -100,13 +100,14 @@ func NewClient(pool Pool,
 	provider WorkerProvider,
 	compression compression.Compression,
 	workerPollingInterval time.Duration,
-	WorkerStatusPublishInterval time.Duration) *client {
+	workerStatusPublishInterval time.Duration,
+) *client {
 	return &client{
 		pool:                        pool,
 		provider:                    provider,
 		compression:                 compression,
 		workerPollingInterval:       workerPollingInterval,
-		workerStatusPublishInterval: WorkerStatusPublishInterval,
+		workerStatusPublishInterval: workerStatusPublishInterval,
 	}
 }
 
@@ -304,6 +305,8 @@ func (client *client) RunTaskStep(
 		return TaskResult{}, err
 	}
 
+	eventDelegate.SelectedWorker(logger, chosenWorker.Name())
+
 	// container already exited
 	exitStatusProp, _ := container.Properties()
 	code := exitStatusProp[taskExitStatusPropertyName]
@@ -427,6 +430,8 @@ func (client *client) RunGetStep(
 		return GetResult{}, err
 	}
 
+	eventDelegate.SelectedWorker(logger, chosenWorker.Name())
+
 	sign, err := resource.Signature()
 	if err != nil {
 		return GetResult{}, err
@@ -484,6 +489,8 @@ func (client *client) RunPutStep(
 	if err != nil {
 		return PutResult{}, err
 	}
+
+	eventDelegate.SelectedWorker(logger, chosenWorker.Name())
 
 	container, err := chosenWorker.FindOrCreateContainer(
 		ctx,
--- a/atc/worker/client_test.go
+++ b/atc/worker/client_test.go
@@ -453,6 +453,8 @@ var _ = Describe("Client", func() {
 
 			It("invokes the Starting Event on the delegate", func() {
 				Expect(fakeEventDelegate.StartingCallCount()).Should((Equal(1)))
+				_, actualWorkerName := fakeEventDelegate.SelectedWorkerArgsForCall(0)
+				Expect(actualWorkerName).To(Equal(fakeChosenWorker.Name()))
 			})
 
 			It("calls Fetch on the worker", func() {
@@ -643,6 +645,8 @@ var _ = Describe("Client", func() {
 			It("chooses a worker", func() {
 				Expect(err).ToNot(HaveOccurred())
 				Expect(fakePool.FindOrChooseWorkerForContainerCallCount()).To(Equal(1))
+				_, actualWorkerName := fakeEventDelegate.SelectedWorkerArgsForCall(0)
+				Expect(actualWorkerName).To(Equal(fakeWorker.Name()))
 			})
 
 			It("invokes the SelectedWorker Event on the delegate", func() {
@@ -1371,6 +1375,8 @@ var _ = Describe("Client", func() {
 
 		It("invokes the SelectedWorker Event on the delegate", func() {
 			Expect(fakeEventDelegate.SelectedWorkerCallCount()).Should((Equal(1)))
+			_, actualWorkerName := fakeEventDelegate.SelectedWorkerArgsForCall(0)
+			Expect(actualWorkerName).To(Equal(fakeChosenWorker.Name()))
 		})
 
 		Context("worker is chosen", func() {
--- a/atc/worker/client_test.go
+++ b/atc/worker/client_test.go
@@ -442,6 +442,10 @@ var _ = Describe("Client", func() {
 			Expect(actualStrategy).To(Equal(fakeStrategy))
 		})
 
+		It("invokes the SelectedWorker Event on the delegate", func() {
+			Expect(fakeEventDelegate.SelectedWorkerCallCount()).Should((Equal(1)))
+		})
+
 		Context("worker is chosen", func() {
 			BeforeEach(func() {
 				fakePool.FindOrChooseWorkerReturns(fakeChosenWorker, nil)
@@ -641,6 +645,10 @@ var _ = Describe("Client", func() {
 				Expect(fakePool.FindOrChooseWorkerForContainerCallCount()).To(Equal(1))
 			})
 
+			It("invokes the SelectedWorker Event on the delegate", func() {
+				Expect(fakeEventDelegate.SelectedWorkerCallCount()).Should((Equal(1)))
+			})
+
 			Context("when 'limit-active-tasks' strategy is chosen", func() {
 				BeforeEach(func() {
 					fakeStrategy.ModifiesActiveTasksReturns(true)
@@ -697,6 +705,10 @@ var _ = Describe("Client", func() {
 
 				It("returns the error", func() {
 					Expect(err).To(Equal(workerDisaster))
+				})
+
+				It("should not invokes the SelectedWorker Event on the delegate", func() {
+					Expect(fakeEventDelegate.SelectedWorkerCallCount()).Should((Equal(0)))
 				})
 			})
 
@@ -1355,6 +1367,10 @@ var _ = Describe("Client", func() {
 			Expect(actualContainerSpec).To(Equal(containerSpec))
 			Expect(actualWorkerSpec).To(Equal(workerSpec))
 			Expect(strategy).To(Equal(fakeStrategy))
+		})
+
+		It("invokes the SelectedWorker Event on the delegate", func() {
+			Expect(fakeEventDelegate.SelectedWorkerCallCount()).Should((Equal(1)))
 		})
 
 		Context("worker is chosen", func() {
--- a/fly/eventstream/render.go
+++ b/fly/eventstream/render.go
@@ -37,6 +37,10 @@ func Render(dst io.Writer, src eventstream.EventStream, options RenderOptions) i
 			dstImpl.SetTimestamp(e.Time)
 			fmt.Fprintf(dstImpl, "%s", e.Payload)
 
+		case event.SelectedWorker:
+			dstImpl.SetTimestamp(e.Time)
+			fmt.Fprintf(dstImpl, "\x1b[1mselected worker: %s\n", e.WorkerName)
+
 		case event.InitializeTask:
 			dstImpl.SetTimestamp(e.Time)
 			fmt.Fprintf(dstImpl, "\x1b[1minitializing\x1b[0m\n")
--- a/fly/eventstream/render.go
+++ b/fly/eventstream/render.go
@@ -39,7 +39,7 @@ func Render(dst io.Writer, src eventstream.EventStream, options RenderOptions) i
 
 		case event.SelectedWorker:
 			dstImpl.SetTimestamp(e.Time)
-			fmt.Fprintf(dstImpl, "\x1b[1mselected worker: %s\n", e.WorkerName)
+			fmt.Fprintf(dstImpl, "\x1b[1mselected worker:\x1b[0m %s\n", e.WorkerName)
 
 		case event.InitializeTask:
 			dstImpl.SetTimestamp(e.Time)
--- a/web/elm/src/Build/Output/Output.elm
+++ b/web/elm/src/Build/Output/Output.elm
@@ -145,6 +145,11 @@ handleEvent event ( model, effects ) =
             , effects
             )
 
+        SelectedWorker origin output time ->
+            ( updateStep origin.id (setRunning << appendStepLog ("selected worker: " ++ output ++ "\n") time) model
+            , effects
+            )
+
         Error origin message time ->
             ( updateStep origin.id (setStepError message time) model
             , effects
--- a/web/elm/src/Build/Output/Output.elm
+++ b/web/elm/src/Build/Output/Output.elm
@@ -146,7 +146,7 @@ handleEvent event ( model, effects ) =
             )
 
         SelectedWorker origin output time ->
-            ( updateStep origin.id (setRunning << appendStepLog ("selected worker: " ++ output ++ "\n") time) model
+            ( updateStep origin.id (setRunning << appendStepLog ("\u{001b}[1mselected worker: \u{001b}[0m" ++ output ++ "\n") time) model
             , effects
             )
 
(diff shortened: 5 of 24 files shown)


evanchaoli:
@xtreme-sameer-vohra Can you please kindly help review this PR as you are assigned? Hopefully it may catch up with 6.4.

xtreme-sameer-vohra:
Hey @evanchaoli 
Sorry for the delay in getting to this. I have been occupied with getting 6.4 to the release line. I should be able to look at this as soon as we have 6.4 out.
Unfortunately, we had the cut off for features for 6.4 last week but it should be good for 6.5.

xtreme-sameer-vohra:
Hey @evanchaoli 
Apologies once again for the delay in getting to this and thank you for submitting the PR.
Few things that come to mind;
- Can you help me understand in what scenarios you'd be wanting to leverage this data ? I am assuming the response from `fly containers` isn't sufficient or timely enough ?
- The PR is leveraging the resource version metadata to store the worker name, which is more ideally suited for metadata that is defined by a resource rather than runtime information. The current approach would also restrict the feature to only `get` & `put` steps. Would you be open to using the delegate (eg. [GetDelegate](https://github.com/concourse/concourse/blob/master/atc/exec/get_step.go#L39) ) for capturing the worker name which would allow it to be extended to the `task` step as well ?

I was able to test the feature and it works. However, I am hoping we can store and display the worker name in a more relevant way to differentiate it from resource metadata.

evanchaoli:
@xtreme-sameer-vohra This is a long story, refer to https://github.com/concourse/concourse/pull/4519.

The idea of showing worker name in metadata is from @vito. In a recent meeting with @vito, (btw, I'm from vmware) I raised this problem again, and @vito asked me to restart the issue, that's reason why I create this PR.

For now, task has no metadata. But @vito said Concourse may add metadata to task later.

xtreme-sameer-vohra:
> @xtreme-sameer-vohra This is a long story, refer to #4519.
> 
> The idea of showing worker name in metadata is from @vito. In a recent meeting with @vito, (btw, I'm from vmware) I raised this problem again, and @vito asked me to restart the issue, that's reason why I create this PR.
> 
> For now, task has no metadata. But @vito said Concourse may add metadata to task later.

Thanks for the context @evanchaoli 

Hey @vito 
I came to the same conclusion as @xtremerui [here](https://github.com/concourse/concourse/pull/4519#issuecomment-536671024). I seem to be missing the context behind using the resource meatadata to store the runtime metadata such as the worker name and also introducing metadata to the task step.

vito:
@xtreme-sameer-vohra Showing it in the metadata isn't ideal, but seems slightly better than printing it in the step output. If we can figure out a better way to do this that doesn't mix this metadata in with the "real" metadata, I think that'd be better, but I don't want to block this for too long on figuring out the perfect UI since the original ask was a long while back, the need is still apparent, and we've already got a lot of UI work on our plate.

If we don't want to put it in the metadata, one alternative idea would be to emit a new event type and have `fly watch` handle the event by printing out the relevant information. (We could even put it behind a `--trace` flag or something.) That way we can defer the UI work until later and rely on `fly watch` for seeing the relevant information until then. @evanchaoli would that be acceptable? I think it wouldn't be too hard to use since you can just copy-paste the build URL into `fly watch -u <url>`.

evanchaoli:
@vito TBH, I don't like to use `fly watch`. To me `fly watch` just prints build logs on terminal. I will have to copy-paster URL from the browser to a terminal, then scroll the terminal over mass build logs to find out worker name lines, or I want to search, I need to redirect `fly watch` outputs to a file, then open the file to search. All of those are just overhead, making issue support more difficult, which makes me to ask, why cannot we just print worker name to build log?

As Concourse prints image pulling logs to build log:

```
Pulling pcfseceng/email-resource@sha256:fb79dc8e3d7d03379d1592faa16ea46e6c7454e63688c94e28962262b45a9205 (attempt 2 of 3)...
sha256:fb79dc8e3d7d03379d1592faa16ea46e6c7454e63688c94e28962262b45a9205: Pulling from pcfseceng/email-resource
b0dc45cd432d: Pulling fs layer
364328af40b6: Pulling fs layer
9c7abf28af64: Pulling fs layer
635bab23d5f1: Pulling fs layer
054e7786c1b6: Pulling fs layer
5100e35a43b2: Pulling fs layer
3b0ecaaeac5e: Pulling fs layer
904fe70e658d: Pulling fs layer
5100e35a43b2: Waiting
635bab23d5f1: Waiting
3b0ecaaeac5e: Waiting
054e7786c1b6: Waiting
904fe70e658d: Waiting
```

why cannot we print worker name to build logs? If you feel workers should be transparent in most of cases, we can opt-in this feature, and only clusters like ours would want to turn on the feature.

vito:
> I will have to copy-paster URL from the browser to a terminal, then scroll the terminal over mass build logs to find out worker name lines, or I want to search, I need to redirect `fly watch` outputs to a file, then open the file to search.

Why not pipe to `grep`? Piping and the Unix philosophy are core to the design of `fly`, and it's one of the primary use cases for `fly watch`.

Here's a mockup for how it could look:

```sh
$ fly watch -u <url> | grep unit
step 0a3bef ("unit") selected worker worker-1
#      ^       ^- step name               ^- worker name
#      `- plan id

```

Or, to see the workers used by all the steps:

```sh
$ fly watch -u <url> | grep 'selected worker'
step deadbe ("build") selected worker worker-1
step 0a3bef ("unit") selected worker worker-1
```

This doesn't seem that bad to me, especially considering I'm likely to run more CLI commands anyway after finding this information.

Either way, the goal would still be to show this in the UI eventually, and having an event implemented would just be the first step. It just so happens that we can also use this event to provide the same information through `fly` until we do the UI work.

> why cannot we just print worker name to build log?

The build output `Log` event is only for output that is actually printed by the processes (tasks or resources) that the user ran. It is not meant to be a place to write arbitrary text. This is why the output events are [associated directly to the `stdout` and `stderr` of the process](https://github.com/concourse/concourse/blob/a822574e7d1e3e30a54d8cac416bc829c0103d02/atc/event/events.go#L93-L114), while every other type of progress or information that we show in the UI has its own event (e.g. Initialize, Start, Finish, Error).

By using an event we can show this in the build UI without confusing users as to why it looks like their process is printing this output, especially when they really shouldn't care most of the time.

> As Concourse prints image pulling logs to build log:

This was a "temporary" solution from the first implementation of using resources to fetch task images. We knew it wasn't perfect but figured "eh, the MVP is done, we'll get to the UI later." Here we are ~5 years later, and I still want to fix it (as mentioned in https://github.com/concourse/concourse/pull/4519#issuecomment-537087980). This is why I'm trying to avoid making the same mistake. 😅

@evanchaoli btw, I wonder if you've checked out [tracing](https://concourse-ci.org/tracing.html) yet? It would make a lot of sense to add this information to the trace for a build.

image

Obviously this isn't a replacement for showing in the web UI, but could be a viable approach from the ops side. I think we had this working with Wavefront too, but don't remember if any code changes were necessary.

evanchaoli:
@vito If your purpose is to keep event type "log" for only build log, then I agree that makes sense.

I have tried tracing in my local environment, but haven't decided to enable tracing on any of our clusters yet. Comparing the gain with burden, I just feel it's not time to enable tracing on the prod clusters yet. (also, tracing is still an experimental feature).

Till now, I haven't been using `fly watch` at all. To me, I try to use `fly` as less as possible. I face multiple clusters (prod, staging, dev, private), switching among targets, `fly login` everyday, which are really overhead. Thus I tend to get all info from only GUI.

Looks like, if I want to move forward, only way is to add a new event, and make `fly watch` to handle the new event. That's fine. Hopefully, after that, you may priority the UI part work. (My team is no longer able to contribute GUI part, because we lost @Twiknight :pray:).

evanchaoli:
@vito @xtreme-sameer-vohra I have re-implemented this PR, please take a look again. Thanks.

atc/engine/builder/delegate_factory.go
```diff
@@ -506,6 +506,17 @@ func (delegate *buildStepDelegate) Finished(logger lager.Logger, succeeded bool)
 	logger.Info("finished")
 }
 
+func (d *buildStepDelegate) SelectedWorker(logger lager.Logger, workerName string) {
+	err := d.build.SaveEvent(event.SelectedWorker{
+		Time:       time.Now().Unix(),
```
vito, line 511: In order to associate this event to its step in the UI, I think we'll need to add an `Origin` field.

Side note: ideally this would just be a field for the step's ID, since there's no stdout/stderr, but it looks like we just use `Origin` and leave the source blank in the `Initialize`/`Start`/`Finish` events, so I guess we should do the same here.

fly/eventstream/render.go
```diff
@@ -37,6 +37,10 @@ func Render(dst io.Writer, src eventstream.EventStream, options RenderOptions) i
 			dstImpl.SetTimestamp(e.Time)
 			fmt.Fprintf(dstImpl, "%s", e.Payload)
 
+		case event.SelectedWorker:
+			dstImpl.SetTimestamp(e.Time)
+			fmt.Fprintf(dstImpl, "\x1b[1mselected worker: %s\n", e.WorkerName)
```
vito, line 42: nit: It will be hard to tell which step selected which worker within an `in_parallel` step. Once the origin ID has been added, we could use it to retrieve the step name from the build plan, but there isn't great support in `go-concourse` for doing that kind of thing right now, so I'm OK with skipping it for this PR.

This is a kind of general problem with the `fly watch`/`fly execute` output - it might be cool to have a [`buildkit`-style UI](https://github.com/moby/buildkit) someday so all events can be associated to their origin in the output. 🤔

xtreme-sameer-vohra:
> @vito @xtreme-sameer-vohra I have re-implemented this PR, please take a look again. Thanks.
> 
> fly-watch-selected-worker

Thanks @evanchaoli . Taking a look

atc/worker/client.go
```diff
@@ -427,6 +430,8 @@ func (client *client) RunGetStep(
 		return GetResult{}, err
 	}
 
+	eventDelegate.SelectedWorker(logger, chosenWorker.Name())
```
xtreme-sameer-vohra, line 433: minor: Might make sense to put this just below the Signature check, as if that fails, this info won't be useful

atc/worker/client_test.go
```diff
@@ -442,6 +442,10 @@ var _ = Describe("Client", func() {
 			Expect(actualStrategy).To(Equal(fakeStrategy))
 		})
 
+		It("invokes the SelectedWorker Event on the delegate", func() {
+			Expect(fakeEventDelegate.SelectedWorkerCallCount()).Should((Equal(1)))
+		})
```
xtreme-sameer-vohra, line 447:
```suggestion
				_, actualWorkerName := fakeEventDelegate.SelectedWorkerArgsForCall(0)
				Expect(actualWorkerName).To(Equal(fakeWorker.Name()))
		})
```

atc/worker/client_test.go
```diff
@@ -641,6 +645,10 @@ var _ = Describe("Client", func() {
 				Expect(fakePool.FindOrChooseWorkerForContainerCallCount()).To(Equal(1))
 			})
 
+			It("invokes the SelectedWorker Event on the delegate", func() {
+				Expect(fakeEventDelegate.SelectedWorkerCallCount()).Should((Equal(1)))
+			})
```
xtreme-sameer-vohra, line 650:
```suggestion
				_, actualWorkerName := fakeEventDelegate.SelectedWorkerArgsForCall(0)
				Expect(actualWorkerName).To(Equal(fakeWorker.Name()))
			})
```

atc/worker/client_test.go
```diff
@@ -1357,6 +1369,10 @@ var _ = Describe("Client", func() {
 			Expect(strategy).To(Equal(fakeStrategy))
 		})
 
+		It("invokes the SelectedWorker Event on the delegate", func() {
+			Expect(fakeEventDelegate.SelectedWorkerCallCount()).Should((Equal(1)))
+		})
```
xtreme-sameer-vohra, line 1374:
```suggestion
			_, actualWorkerName := fakeEventDelegate.SelectedWorkerArgsForCall(0)
			Expect(actualWorkerName).To(Equal(fakeChosenWorker.Name()))
		})
```

fly/eventstream/render.go
evanchaoli: Okey, I'm going to skipping that for this PR.

atc/worker/client.go
evanchaoli: My idea was to emit worker info as early as possible. Because if everything goes fine, then we don't need to check worker name at all; but if something goes wrong, any error might be related to worker.

evanchaoli:
@vito Can you please help check why `watjs` test always failed, I have retried multiple times. I really have no idea with UI stuff.

xtremerui:
@evanchaoli I ran the failed test manually and see the log:
```
worker_1  | {"timestamp":"2020-07-28T14:25:31.356612500Z","level":"error","source":"guardian","message":"guardian.api.garden-server.get-property.failed","data":{"error":"property does not exist: concourse:exit-status","handle":"1ad28ded-a664-4a47-5fbc-a9938662d7b1","session":"3.1.48"}}
```
while in the UI the build finished but in task step `hello` it lonly prints out the image fetching log. No task event log is printed out. That why the test failed as it expect to see the task body to show `Hello World`

evanchaoli:
@xtremerui How does it relate to my change? I need help to fix the failure.

vito:
@xtremerui I think that log is a red herring - it shows up for every build as part of the idempotent step execution. That log occurs when a step checks to see if it has already run.

xtreme-sameer-vohra:
It seems to be failing consistently and If I manually visit the page I  seem to get the spinner without anything in the console. 

It seems to work via fly
```
/Users/svohra/go/bin/fly -t dev watch -u http://localhost:8080/teams/watsjs-team-adc8dfd5-cfdd-4a5c-b12a-7079717db992/pipelines/some-pipeline/jobs/say-hello/builds/1
selected worker: 0a6ac0c5ef43
INFO[0000] fetching version: mock
initializing
INFO[0000] fetching version: mock
INFO[0000] mirroring self image                          via_params=false
selected worker: 0a6ac0c5ef43
running echo Hello, world!
Hello, world!
selected worker: 0a6ac0c5ef43
INFO[0000] pushing version: put-version
selected worker: 0a6ac0c5ef43
INFO[0000] fetching version: put-version
succeeded
```

xtreme-sameer-vohra:
> @xtremerui I think that log is a red herring - it shows up for every build as part of the idempotent step execution. That log occurs when a step checks to see if it has already run.
> 
> @evanchaoli Maybe the web UI doesn't know how to handle (in this case, ignore) the new event type? Have you been able to test the build UI manually and see if everything renders OK?

I think it was the latter, where the UI doesn't know how to handle the new event type. This git patch got the UI rendering properly and for the test to pass. What would be the right way of handling it in `Output.elm` ?

```
diff --git a/web/elm/src/Build/Output/Output.elm b/web/elm/src/Build/Output/Output.elm
index 3e56908b6..50072e4f9 100644
--- a/web/elm/src/Build/Output/Output.elm
+++ b/web/elm/src/Build/Output/Output.elm
@@ -145,6 +145,12 @@ handleEvent event ( model, effects ) =
             , effects
             )
 
+        SelectedWorker origin output time ->
+            ( updateStep origin.id (setRunning << appendStepLog output time) model
+            , effects
+            )
+
+
         Error origin message time ->
             ( updateStep origin.id (setStepError message time) model
             , effects
diff --git a/web/elm/src/Build/StepTree/Models.elm b/web/elm/src/Build/StepTree/Models.elm
index d0858133d..bd1b7dc8f 100644
--- a/web/elm/src/Build/StepTree/Models.elm
+++ b/web/elm/src/Build/StepTree/Models.elm
@@ -123,6 +123,7 @@ type alias BuildEventEnvelope =
 type BuildEvent
     = BuildStatus BuildStatus Time.Posix
     | InitializeTask Origin Time.Posix
+    | SelectedWorker Origin String (Maybe Time.Posix)
     | StartTask Origin Time.Posix
     | FinishTask Origin Int Time.Posix
     | Initialize Origin Time.Posix
diff --git a/web/elm/src/Concourse/BuildEvents.elm b/web/elm/src/Concourse/BuildEvents.elm
index ef8fa7bde..1aa967f65 100644
--- a/web/elm/src/Concourse/BuildEvents.elm
+++ b/web/elm/src/Concourse/BuildEvents.elm
@@ -86,6 +86,16 @@ decodeBuildEvent =
                                 (Json.Decode.maybe <| Json.Decode.field "time" <| Json.Decode.map dateFromSeconds Json.Decode.int)
                             )
 
+                    "selected-worker" ->
+                        Json.Decode.field
+                            "data"
+                            (Json.Decode.map3 SelectedWorker
+                                (Json.Decode.field "origin" <| Json.Decode.lazy (\_ -> decodeOrigin))
+                                (Json.Decode.field "selected_worker" Json.Decode.string)
+                                (Json.Decode.maybe <| Json.Decode.field "time" <| Json.Decode.map dateFromSeconds Json.Decode.int)
+                            )
+
+
                     "error" ->
                         Json.Decode.field "data" decodeErrorEvent
```

evanchaoli:
@vito I tested with UI, as I triggered the build via UI. The UI rendered fine, see the screenshot:

ui-good

evanchaoli:
Greatly thank @xtreme-sameer-vohra for the piece of code. Now I can show selected worker on UI:

ui-selected-worker

xtreme-sameer-vohra:
Thanks @evanchaoli 
I tested the change via `fly` and the UI and it looks great. I am going to just wait for a peer to review the elm code as I am not the right person for that ;) 
Everything else looks good and is ready.

fly/eventstream/render.go
aoldershaw, line 42: Looks like this is missing the ANSI reset?

The following will bold only "selected worker:", not sure if we also want to bold the worker name

```suggestion
			fmt.Fprintf(dstImpl, "\x1b[1mselected worker:\x1b[0m %s\n", e.WorkerName)
```

web/elm/src/Build/Output/Output.elm
```diff
@@ -145,6 +145,11 @@ handleEvent event ( model, effects ) =
             , effects
             )
 
+        SelectedWorker origin output time ->
+            ( updateStep origin.id (setRunning << appendStepLog ("selected worker: " ++ output ++ "\n") time) model
```
aoldershaw, line 149: Might be nice to bold this as we do with the `fly watch` output. Should be pretty much the same, albeit with a different escape char:

```suggestion
            ( updateStep origin.id (setRunning << appendStepLog ("\u{001b}[1mselected worker: \u{001b}[0m" ++ output ++ "\n") time) model
```

aoldershaw: Wait, based on the discussion with @vito above (https://github.com/concourse/concourse/pull/5846#issuecomment-665121993), it looks like the decision was to ignore this information in the web UI and only log a message in `fly watch`

vito: Hmm, yeah that was the idea, but I was more assuming that @evanchaoli wouldn't be interested in doing any UI work so I suggested that we skip it for now.

My main concern was around the way the data is recorded. Now that it has its own event type, we can at least change the UI in the future, so I'm OK with doing this for now.

fly/eventstream/render.go
evanchaoli: Oh, yes, good catch!

web/elm/src/Build/Output/Output.elm
evanchaoli: I don't know ELM at all, so I am not able to contribute UI part.

The current implementation has met my needs, so it would relief pressure from your side, as I'll no longer push for UI part. 😄

fly/eventstream/render.go
```diff
@@ -39,7 +39,7 @@ func Render(dst io.Writer, src eventstream.EventStream, options RenderOptions) i
 
 		case event.SelectedWorker:
 			dstImpl.SetTimestamp(e.Time)
-			fmt.Fprintf(dstImpl, "\x1b[1mselected worker: %s\n", e.WorkerName)
+			fmt.Fprintf(dstImpl, "\x1b[1mselected worker:\x1b[0m %s\n", e.WorkerName)
```
aoldershaw, line 42: Looks like this broke the unit test actually: https://ci.concourse-ci.org/builds/2054579#L5f1da44d:550:553

evanchaoli: My bad. Forgot to update unit test, and after pushed the commit, didn't get a chance to come back to check build result. Just fixed.

PR #5834, opened 2020-06-30
atc/db: track last successful build id of a job

taylorsilva:
## What does this PR accomplish?
Bug Fix | **Feature** | Documentation

closes #5800

## Changes proposed by this PR:

* Add a column called `latest_successful_build_id` to the `jobs` table
* When a job finishes successfully, it logs the build ID in the new column

## Contributor Checklist
- [x] Followed [Code of conduct], [Contributing Guide] & avoided [Anti-patterns]
- [x] [Signed] all commits
- [x] Added tests (Unit and/or Integration)
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

[Code of Conduct]: https://github.com/concourse/concourse/blob/master/CODE_OF_CONDUCT.md
[Contributing Guide]: https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md
[Anti-patterns]: https://github.com/concourse/concourse/wiki/Anti-Patterns
[Signed]: https://help.github.com/en/github/authenticating-to-github/signing-commits
[Documentation]: https://github.com/concourse/docs
[Release notes]: https://github.com/concourse/concourse/tree/master/release-notes

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [ ] New config flags added? Ensure that they are added to the
  [BOSH](https://github.com/concourse/concourse-bosh-release) and
  [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for
  the [integration
  tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env)
  (for example, if they are Garden configs that are not displayed in the
  `--help` text).

PR #5854, opened 2020-07-07
Automatically archive pipelines

### LINKED ISSUE #5835: Automatically archive pipelines that are no longer set by their parent jobs
Create a new garbage collector that will archive pipelines that were previously set by a `set_pipeline` step and no longer are.

taylorsilva:

closes #5835 

* Added a db object, PipelineLifecycle, which find abandoned pipelines and archives them
* Added a gc object that acts as a thin wrapper around PipelineLifecycle
* Removes the feature flag for enabling archiving pipelines. Archiving pipelines is now always on

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [x] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


atc/atccmd/command.go
```diff
@@ -1113,6 +1114,7 @@ func (cmd *RunCommand) gcComponents(
 		atc.ComponentCollectorVolumes:           gc.NewVolumeCollector(dbVolumeRepository, cmd.GC.MissingGracePeriod),
 		atc.ComponentCollectorContainers:        gc.NewContainerCollector(dbContainerRepository, cmd.GC.MissingGracePeriod, cmd.GC.HijackGracePeriod),
 		atc.ComponentCollectorCheckSessions:     gc.NewResourceConfigCheckSessionCollector(resourceConfigCheckSessionLifecycle),
+		atc.ComponentCollectorPipelines:         gc.NewPipelineCollector(dbPipelineLifecycle),
```
aoldershaw, line 1117: Right now, archiving pipelines is feature flagged - so IMO either we should get rid of the feature flag, or only conditionally create the GC component

atc/db/pipeline_lifecycle.go
```diff
@@ -0,0 +1,87 @@
...
+		lockFactory: lockFactory,
+	}
+}
+
+type pipelineLifecycle struct {
+	conn        Conn
+	lockFactory lock.LockFactory
+}
+
+func (pl *pipelineLifecycle) ArchiveAbandonedPipelines() (int, error) {
```
aoldershaw, line 26: One alternative to having a GC component would be to perform the query whenever a build finishes successfully. That way, we wouldn't need to perform the `successful_builds` subquery in the first place - we could just use the job ID from the finished build.

Did you consider about going with that approach? One benefit is that the pipeline archiving would be immediate. Not sure what the performance impacts would be on an environment with lots of builds running constantly, though I imagine the query would be pretty cheap if just on a single `job_id`

atc/atccmd/command.go
aoldershaw: Might make sense to get rid of the flag altogether now that archiving pipelines is complete (I think?)

atc/db/pipeline_lifecycle.go
taylorsilva: Didn't think of doing this approach, though I'm hesitant to add a query that would be run at the end of **every** build. I think this would be problematic for people running large clusters where there are hundreds of builds finishing every minute. I feel like the query would be a no-op for ~99% of the builds too.

You can see our thought process here for some other alternatives: https://github.com/concourse/concourse/issues/5699#issuecomment-654475693

atc/atccmd/command.go
taylorsilva: Forgot there was even a flag for this! I think we can remove it since the RFC was merged.

atc/db/pipeline_lifecycle.go
chenbh: > One alternative to having a GC component would be to perform the query whenever a build finishes successfully.

That would have a problem on small systems if you delete a pipeline (or remove a job) and there's no builds running at the time. We can split it into 3 smaller queries that get triggered when one of the 3 conditions are met, but that would spread the logic over into multiple places.

aoldershaw: True that the query would be a no-op the majority of the time, but I imagine with an index on `p.parent_job_id`, the query would be really cheap - the `WHERE p.parent_job_id = $1` would result in 0 rows and short-circuit for all the no-op jobs, and would result in only a few rows when it does run `set_pipeline`. So the query for a build scales with the number of pipelines that the job sets, which seems appropriate.

With the GC/interval approach, it seems like there's a constant cost every GC tick (with some scary terms in the `EXPLAIN ANALYZE` like "Nested Loop Left Join" and "Sort Method: external merge"). Maybe that's okay (I'm definitely not an expert on SQL optimization), but it does seem like a pretty expensive query to be running every 30 seconds.

Worth noting that even with running the query in `Build.Finish()`, we'd still want to have some sort of GC interval going on to detect the *other* ways pipelines can get archived (parent pipeline destroyed, job no longer active) - though these would likely be much cheaper, given that we wouldn't need the `successful_builds` subquery.

Maybe this is something we can take a look into optimizing if we run into performance woes with this approach

atc/db/pipeline_lifecycle_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+	Context("the build succeeds once then fails", func() {
+		BeforeEach(func() {
+			By("first build is successful")
+			build, _ := defaultJob.CreateBuild()
+			child1Pipeline, _, _ = build.SavePipeline("child1-pipeline", defaultTeam.ID(), defaultPipelineConfig, db.ConfigVersion(0), false)
+			child2Pipeline, _, _ = build.SavePipeline("child2-pipeline", defaultTeam.ID(), defaultPipelineConfig, db.ConfigVersion(0), false)
+			build.Finish(db.BuildStatusSucceeded)
+
+			By("second build fails and sets no pipelines")
```
aoldershaw, line 61: Nit, and not that it really matters for the purpose of the test, but this claims it "sets no pipelines", but it does set one of the two pipelines

atc/db/pipeline_lifecycle.go
```diff
@@ -0,0 +1,89 @@
...
+	tx, err := pl.conn.Begin()
+	if err != nil {
+		return 0, err
+	}
+
+	defer Rollback(tx)
+
+	latestSuccessfulBuilds := `WITH successful_builds AS (
+      SELECT DISTINCT ON (job_id) id, job_id
```
clarafu, line 35: Hm I think I'm a little worried about this query when it is run against a really large deployment with thousands of builds. I quickly ran just this SELECT query against hush-house and it seemed to have taken around 2 minutes to run. I think this query worries me because it has to select out all of the builds that have a succeeded status, which most builds in the database are succeeded (hopefully haha). There is the results of the explain analyze:

```
Unique  (cost=768784.77..785318.94 rows=842 width=8) (actual time=125106.692..126360.440 rows=3023 loops=1)
  ->  Sort  (cost=768784.77..777051.86 rows=3306834 width=8) (actual time=125106.690..125927.008 rows=3351546 loops=1)
        Sort Key: job_id, id DESC
        Sort Method: external sort  Disk: 58984kB
        ->  Index Only Scan using succeeded_builds_ordering_with_rerun_builds_idx on builds  (cost=0.56..320277.94 rows=3306834 width=8) (actual time=9.970..119191.727 rows=3351546 loops=1)
              Heap Fetches: 666449
Planning time: 0.723 ms
Execution time: 126369.643 ms
```

I think what you're trying to achieve is to grab the latest successful build of each job, so I think in order to have the query not need to go through every successful build of a job, you might be able to do something that is like "for each job, order the successful builds in descending order of id but **LIMIT 1**`. Something similar would be like this query here

https://github.com/concourse/concourse/blob/ba4fcfdcad4f9343bfe4153a7539148dec67be52/atc/db/versions_db.go#L418-L423

But the difference is in that query I'm only looking for one build but here you are looking for multiple.

atc/db/pipeline_lifecycle.go
```diff
@@ -0,0 +1,89 @@
...
+      ORDER BY job_id, id DESC
+      )`
+	rows, err := pipelinesQuery.
+		Prefix(latestSuccessfulBuilds).
+		LeftJoin("jobs j ON j.id = p.parent_job_id").
+		LeftJoin("successful_builds ON successful_builds.job_id = p.parent_job_id").
+		Where(sq.Or{
+			// pipeline is no longer set by the parent pipeline
+			sq.And{
+				sq.NotEq{"parent_build_id": nil},
```
clarafu, line 47: I'm also wondering if we can add an index to this column? I don't see one added in the migration that introduced the column but maybe there's a reason for not adding one? Also same with the `archived` boolean.

https://github.com/concourse/concourse/blob/master/atc/db/migration/migrations/1591737494_add_parent_job_id.up.sql

atc/gc/pipeline_collector.go
```diff
@@ -0,0 +1,33 @@
...
+	}
+}
+
+func (pc *pipelineCollector) Run(ctx context.Context) error {
+	logger := lagerctx.FromContext(ctx).Session("pipeline-collector")
+
+	logger.Debug("start")
+	defer logger.Debug("done")
+
+	_, err := pc.pipelineLifecycle.ArchiveAbandonedPipelines()
```
aoldershaw, line 26: Maybe we should log the number of pipelines archived? Just logging start and done doesn't give much visibility into when things actually get archived

testflight/fixtures/archive-pipeline-3.yml
```diff
@@ -0,0 +1,39 @@
...
+        ---
+        name: somebody
+
+jobs:
+- name: sp-1
+  public: true
+  plan:
+    - get: some-resource
+    - set_pipeline: ((pipeline_name))
+      team: ((team_name))
```
aoldershaw, line 32: Is there a reason the `team` is configurable here (sim. in `archive-pipeline-1.yml`)? Based on the discussion here https://github.com/concourse/concourse/pull/5729#issuecomment-641392557, I believe the intention is to eventually get rid of `set_pipeline.team`, and it's just meant to be a temporary solution until we figure out the "right" way.

Maybe worth getting rid of it so we won't need to change these tests if/when we do get rid of the field.

atc/db/pipeline_lifecycle.go
clarafu: Hm I'm not 100% with my opinion but I think I would slightly prefer having the garbage collector archive pipelines purely in the sense that I would prefer having a separate component be the one that archives a pipeline rather than having the finishing of a build be overloaded with more responsibilities. But I also understand that builds can be the "owners" of a pipeline so maybe it does make sense?

I do think there is a concern about the responsiveness of the archiving if it was done by gc, where the users would have to wait for up to 10 seconds to see their pipeline be archived after a build finishes. I'm thinking maybe we can have a notification that gets sent to the component that archives the pipelines? But there is a worry that if we do keep a interval of 10 seconds, because there are so many builds finishing in a large deployment having a notification sent will not do much because the gc component can only run once every 10 seconds anyways. Maybe is there a way to make it only send a notification if it is a build that owns a pipeline? And that way we can trust the notification that comes in and run it everytime?

aoldershaw: @clarafu the thing that strikes me as odd if we were to send a notification to GC when a build finishes (and why I prefer performing the query in `Finish`) is that we already *know* what (single) `job_id` we need to check when the build finishes - with the separate GC component, we'd just be choosing to ignore that information and instead consider all the jobs.

I guess we could send along the notification bus the `job_id` that needs to be checked - I've implemented this a few times in my experimental PRs: https://github.com/concourse/concourse/pull/5651/commits/85e285272b1f38bf243cb1ace1d2386fa5a1b7ec (the difficulty being we then have to queue up notifications so we don't lose information). But that seems more complicated than it should be.

atc/db/pipeline_lifecycle.go
taylorsilva: Guess our data set we experimented with was a poor representation of real life, see here: https://github.com/concourse/concourse/issues/5699#issuecomment-654475693

In short, I agree with your concerns and this means we should probably go the denormalization route and bring back this PR: https://github.com/concourse/concourse/pull/5834

Regarding:
> do something that is like "for each job, order the successful builds in descending order of id but LIMIT 1"
How different is that from having `SELECT DISTINCT ON (job_id)` in the query, which also results in one result per job_id?

taylorsilva: I'm trying to think how we would add `LIMIT 1` to the query, but can't think of a way that wouldn't result in us hitting the build table per child pipeline...

atc/gc/pipeline_collector.go
taylorsilva: We debated this and figured that it's more valuable as a metric instead of a log line. We didn't discuss or dig into trying to add the metric though.

chenbh: I was thinking about emitting metrics vs logging the number, don't really feel strongly about either

aoldershaw: Good point, metrics makes sense - looks like there's precedence for this with e.g. the `ChecksCollector`:

https://github.com/concourse/concourse/blob/669f6b20a4b677443f70c997ae3eb80815726179/atc/gc/checks_collector.go#L36

atc/db/pipeline_lifecycle.go
clarafu: Something I was thinking about was something like `EXPLAIN ANALYZE SELECT ( SELECT bu.id FROM builds bu where bu.job_id = j.id AND bu.status='succeeded' ORDER BY bu.id LIMIT 1), j.id FROM jobs j;`. The thing is with the limit 1, it only really works with the correct index. So it would require something like `CREATE INDEX deleteme ON builds (job_id, id DESC) where status = 'succeeded';`. 

I am hesitant on adding another index to the builds table.. I'm wondering if we can use the successful_build_outputs table too? That way we don't need the `WHERE status = successful` but it would still require a new index on the table.

atc/db/pipeline_lifecycle.go
clarafu: Ya I agree with you, I think its unnecessary. My concern is probably it feeling a bit messy having a few different places where a pipeline can be archived (api, build finishing and gc) and it might feel weird as a user if sometimes their pipeline gets archived immediately (build finishing) but others it takes up to 10 seconds to archive (job is set as inactive). As a thought experiment, would it be possible to always archive pipelines always immediately? This means we no longer need a gc for archiving pipelines and we will always just archive when necessary (build finishing, parent pipeline destroyed, etc).

As a side note: I think if a build finishes and a pipeline is archived as a result, I think we will probably want to also immediately archive any pipelines that were set by that archived pipeline.

atc/db/pipeline_lifecycle.go
taylorsilva: Building off your query I made this query. I don't have a data set + schema to test its performance. Running it on hush-house returns zero records as expected:
```sql
SELECT p.id, p.parent_job_id, p.parent_build_id
FROM pipelines p
LEFT JOIN jobs j on j.id  = p.parent_job_id
WHERE
(parent_build_id IS NOT NULL 
 AND archived=false 
 AND p.parent_build_id < (SELECT bu.id FROM builds bu where bu.job_id = p.parent_job_id AND bu.status='succeeded' ORDER BY bu.id LIMIT 1)
)
OR
(parent_build_id IS NOT NULL AND j.id IS NULL)
OR
(j.active=false)
```

taylorsilva: Was curious about the data set size in hush house:
```
select count(*) from pipelines; 357
select count(*) from jobs; 3700
select count(*) from builds; 4,055,731
```
That's a lot of builds!

atc/db/pipeline_lifecycle.go
```diff
@@ -0,0 +1,89 @@
...
+      SELECT DISTINCT ON (job_id) id, job_id
+      FROM builds
+      WHERE status='succeeded'
+      ORDER BY job_id, id DESC
+      )`
+	rows, err := pipelinesQuery.
+		Prefix(latestSuccessfulBuilds).
+		LeftJoin("jobs j ON j.id = p.parent_job_id").
+		LeftJoin("successful_builds ON successful_builds.job_id = p.parent_job_id").
+		Where(sq.Or{
```
taylorsilva, line 44: Not seeing a comment for this yet so dropping it here before I forget. We should also archive child pipelines whose parent pipeline has been archived. This was asked by Aidan on slack

atc/db/pipeline_lifecycle.go
taylorsilva: I have the same thoughts as Clara
> ...having a few different places where a pipeline can be archived (api, build finishing and gc)...

Also makes it harder for us to maintain and think about the "automatic archiving feature".

We're going to stick with the gc approach. I think users are use to Concourse taking some time to do things, like job scheduling, running checks, etc.
Waiting max 30s probably won't feel abnormal to them and by the time they ask "why hasn't my pipeline been archived yet?" it'll probably be gone.

aoldershaw: My concerns are now less about the immediacy of archiving and more about choosing to ignore info that would make the query much cheaper (scoped to a single job rather than all jobs). If we make the query cheap across all jobs, though, then the extra complexity probably won't be worth it

clarafu: I think my only concern about the difference between Concourse taking some time to do things like job scheduling or running checks and archiving pipelines in gc is the fact that we show the users that something is happening in the background (build is pending or check is running). With archiving pipelines, the user won't see anything in the UI that shows that the archive is pending and it might just seem laggy that the archiving takes long to happen or maybe even broken because there is no indication that it is waiting to be archived. What do you think @taylorsilva? But maybe I'm just worried about something that could be a UI concern of being more transparent of when it is waiting to be archived?

taylorsilva: I agree with the UI/UX concerns here. Feels like automatic archiving is not very transparent about how it works or what it's doing. I'll be fine with that though if we detail how it works and when it GC's in the docs so users can find an answer or we can at least point to an answer when users ask.

Overall I still prefer the GC approach for its simplicity. I think we can still make a performant query with the GC approach. If we need to improve the performance of the query there are options. Right now we know the CTE part of the query is slow. The newer query (subquery with `LIMIT 1`) looks like it'll be faster but we need to test it which is what we plan to do today. We can also still go the denormalization route if we need even better performance (at the cost of data integrity and duplication). I'd prefer trying those routes first before splitting up each case where we should automatically archive into various parts of the code base.

kcmannem: if theres a delay in the UI, could I just kick off another build without knowing a piplines going to be archived the next min? (Might be a dumb question, Idk what archivable piplines looks like and if they let you kick off more than 1 build)

aoldershaw: @kcmannem yeah good point. I guess we could add a check in the “reject archived” wrappa that would determine if the pipeline would be archived in the next GC tick, and that would prevent performing any operations that are disallowed on archived pipelines (e.g triggering builds)

I hadn’t really thought about what to do about pipelines that have active builds when we try to archive them...maybe the GC could mark running pipelines as “awaiting archive”, and they could get swept in the next GC tick? We’d prevent scheduling on pipelines that are awaiting archived

aoldershaw: Or could we just archive the pipeline right away? I guess it may not be a problem archiving a pipeline with a running build, since the build could still complete

vito: Yeah, it should be OK to archive while a build is running; the build will eventually complete and the user will still be able to see the build if they navigate to the archived pipeline. It should also be safe to clear out the config, since by that point the build is independent of it and just running a build plan.

I do think that having the build complete and then the pipelines remain active and become archived some amount of time later doesn't feel great, which has already been pointed out. 🤔 The user may think Concourse decided not to archive it and wonder why. Seems like a potential source of repeated minor anxiety/uncertainty.

Archiving immediately seems a lot better since it removes that window of time, but TBH still not *super* great because it still may not be obvious that the build caused it to happen.

What I'm about to suggest is totally scope creep and belongs in a separate issue if we decide to look into it, but: if we could perform the archiving operations at the end of the build *and show it happening in the build plan*, that would be pretty awesome. It's a bit tricky because it's not just something you can do in the `set_pipeline` step - it has to be done after all `set_pipeline` steps have completed so we know the full set to diff. I'm tempted to try to frame this as a dynamic build plan, because such a mechanism could be helpful in other scenarios in the future (e.g. using the `across` step with a set of values that were generated within the build).

🤔 🤔 🤔

kcmannem: Like this? Being appended to the end of a build plan? Seems nice that it's explicit.

```
on_complete:
  - archive_pipeline
```

aoldershaw: @vito rather than using dynamic build plan, could we just create a new build event that represents archiving a pipeline? So the archiving doesn’t have a build plan, but we can still be explicit about what happened when looking at the build in the UI/Fly?

Can see the value in supporting dynamic build plans for across, though

taylorsilva: We decided to add onto `build.Finish()` so pipelines that are no longer being set by the parent pipeline are removed immediately at the end of a build.

All other cases are covered by the GC approach.

atc/db/pipeline_lifecycle.go
taylorsilva: We decided to cover the first case (which is causing all the performance concerns for the query) in `build.Finish()`. When a build finishes it will archive any pipelines that were previously being set by the job.

All other cases (parent pipeline deleted) are still done in GC by one query that no longer has to hit the `builds` table.

atc/db/pipeline_lifecycle.go
taylorsilva: We added an index to the `parent_build_id` and `parent_job_id`. Not sure if they're really needed though because we did an `EXPLAIN ANALYZE` and postgres never used either of the indexes. Not sure if it was due to the small dataset we had or not. wdyt?

atc/db/pipeline_lifecycle.go
vito: @kcmannem I was thinking that Concourse would automatically do it - by build plan I meant the internal `atc.Plan`, not the config `plan:`.

@aoldershaw Yeah that'd be a more lightweight option!

@taylorsilva Sounds good, let's look for feedback from users about this before committing to UI changes - at least it happens immediately!

atc/db/build.go
```diff
@@ -549,6 +549,26 @@ func (b *build) Finish(status BuildStatus) error {
 		if err != nil {
 			return err
 		}
+
+		pipelineRows, err := pipelinesQuery.
```
clarafu, line 553: Nice! I think having pipelines archive immediately makes sense because it won't have the user guessing as to when the expected archival will happen (also will feel much more responsive).

One thing I wanted to get your opinion on was the fact that this will only archive the pipeline that is the direct child of the build/job that is finished. So for example, let's say pipeline-1 has a job that sets pipeline-2 and pipeline-2 has a job that sets pipeline-3. If the build of pipeline-1 archives pipeline-2, pipeline-3 will not be archived until the gc tick. This would result in a user seeing their direct child pipeline being archived immediately, but any pipelines that were set by that archived pipeline would have to wait 10 seconds (or more depending on how nested the pipelines are).

I'm not sure if you think this is something that we want to fix? I think for now with pipelines it might not be a big use case because people probably won't have more than 1 level deep of nested pipelines? It's just something that would be possibly weird to the user if they do end up running into this situation.

If you do think it's something you want to explore, I wrote up a query that would cover that case. It recursively finds any pipelines that are children of any pipelines we found that need to be archived.
```
WITH RECURSIVE pipelines_to_archive AS (
	SELECT id from pipelines where archived = false AND parent_job_id = ? AND parent_build_id < ?
	UNION
	SELECT p.id from pipelines p join jobs j on p.parent_job_id = j.id join pipelines_to_archive on j.pipeline_id = pipelines_to_archive.id)
SELECT * from pipelines_to_archive;
```

aoldershaw: Woah, didn't know WITH RECURSIVE was a thing

taylorsilva: DAMN that's cool! We did briefly talk about that exact example you gave and thought "nahh who would do recursive pipelines?!". We'll give it a shot to see if we can implement it since you've done the hard work of making the query.

clarafu: Haha ya that's kinda where I was at with it too, I'm not sure who would do pipelines setting other pipelines?? But I guess if it's a light change that won't cost much then it might to worth to add?

atc/gc/pipeline_collector.go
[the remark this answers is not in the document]
taylorsilva: Doesn't feel like it would be a useful metric for people to consume. I decided to not return an `int` anymore, just an error from `pipelineLifecycle.ArchiveAbandonedPipelines()`

atc/db/pipeline_lifecycle.go
clarafu: If there are no rows that are fetched using the index or if the dataset was small enough that postgres determined it was faster to not use an idex, postgres explain analyze won't show that it uses the index. I just ran the query for archiving pipelines after a build is finished (https://github.com/concourse/concourse/pull/5854/files#diff-cd2dc5a827d8e3673211bb2ac1540801R556-R567) and it seemed to be using the index for `parent_job_id` but not `parent_build_id`

```
  CTE pipelines_to_archive
    ->  Recursive Union  (cost=4.16..178.36 rows=81 width=4) (actual time=0.055..0.967 rows=3 loops=1)
          ->  Bitmap Heap Scan on pipelines  (cost=4.16..9.51 rows=1 width=4) (actual time=0.036..0.063 rows=1 loops=1)
                Recheck Cond: (parent_job_id = 1)
                Filter: ((NOT archived) AND (parent_build_id < 2))
                Heap Blocks: exact=1
                ->  Bitmap Index Scan on parent_job_id  (cost=0.00..4.16 rows=2 width=0) (actual time=0.015..0.023 rows=1 loops=1)
                      Index Cond: (parent_job_id = 1)
          ->  Hash Join  (cost=1.42..16.72 rows=8 width=4) (actual time=0.233..0.274 rows=1 loops=3)
                Hash Cond: (j.pipeline_id = pipelines_to_archive.id)
                ->  Hash Join  (cost=1.09..16.29 rows=8 width=8) (actual time=0.098..0.174 rows=3 loops=3)
                      Hash Cond: (p_1.parent_job_id = j.id)
                      ->  Seq Scan on pipelines p_1  (cost=0.00..14.10 rows=410 width=8) (actual time=0.010..0.050 rows=4 loops=3)
                      ->  Hash  (cost=1.04..1.04 rows=4 width=8) (actual time=0.115..0.123 rows=4 loops=1)
                            ->  Seq Scan on jobs j  (cost=0.00..1.04 rows=4 width=8) (actual time=0.019..0.061 rows=4 loops=1)
                ->  Hash  (cost=0.20..0.20 rows=10 width=4) (actual time=0.042..0.045 rows=1 loops=3)
                      ->  WorkTable Scan on pipelines_to_archive  (cost=0.00..0.20 rows=10 width=4) (actual time=0.008..0.019 rows=1 loops=3)
  ->  Seq Scan on pipelines p  (cost=0.00..14.10 rows=410 width=167) (actual time=0.018..0.057 rows=4 loops=1)
  ->  Hash  (cost=1.62..1.62 rows=81 width=4) (actual time=1.072..1.079 rows=3 loops=1)
        ->  CTE Scan on pipelines_to_archive pa  (cost=0.00..1.62 rows=81 width=4) (actual time=0.072..1.037 rows=3 loops=1)
Planning Time: 0.339 ms
Execution Time: 1.531 ms
```

I'm not exactly sure why it isn't using the index for `parent_build_id`, maybe it's due to the fact that its not an equal operation but is actually a less than operation? Or it could be due to my dataset being too small that it was cheaper to not use the index. 

In any case, we can keep the indexes and monitor the usage in `hush-house`. If we find that the indexes are not useful, we can remove them.

aoldershaw: I'm guessing just a `parent_job_id` index enough. I'm guessing it'll be rare that a single job would have enough child pipelines to fill more than a few database pages, which is likely why Postgres chooses not to use the `parent_build_id` index.

clarafu: Ah yeah thats a good point! I definitely think we can remove the `parent_build_id` index then.

atc/db/migration/migrations/1595261696_add_parent_build_and_job_id_index.up.sql
```diff
@@ -0,0 +1,4 @@
+BEGIN;
+  CREATE INDEX parent_job_id ON pipelines USING btree (parent_job_id);
+  CREATE INDEX parent_build_id ON pipelines USING btree (parent_build_id);
```
clarafu, line 3: @aoldershaw brought up a really good point that the `parent_build_id` index will probably never need to be used.  https://github.com/concourse/concourse/pull/5854#discussion_r461695769 We can remove it!

taylorsilva:
@clarafu removed the `parent_build_id` index! Needs approval again

alonikomax:
Hey, I need to disable this for some jobs that run set_pipeline multiple times.

PR #5861, opened 2020-07-09
Allow `fly set-pipeline` to take configuration from stdin

evanchaoli:
feat: allow set-pipeline to take configure from stdin. this would make pipeline migration from one cluster to the other easier.

With this feature a user may migrate a pipeline from one cluster to the other with a single line command:

```bash
$ fly -t <target1> gp -p <pipeline-name> | fly -t <target2> sp -c - -p <pipeline-name> -n
```

Modified `PathFlag` to accept `-`. When `-` is given, it generate a temp file by reading from stdin.

## Notes to reviewer:

I see a couple of problems with this implementation:

1. The second `fly sp` must be given `-n`, otherwise `apply configuration? [yN]: bailing out`. As the question/answer is done by @vito's packet `interaction`, I wonder if @vito has idea with that.

2. `set-pipeline` command actually takes multiple `PathFlag`, `-c`, `-l`, only at most a single flag can use `-`, but it seems no way to restrict that.

Maybe we only add description about `-` to `-c`?

## Release Note

The `--config` flag of the `fly set-pipeline` command now supports `-` for reading pipeline config from stdin

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/path_flag.go
+++ b/atc/path_flag.go
@@ -2,6 +2,9 @@ package atc
 
 import (
 	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
 	"path/filepath"
 	"strings"
 
@@ -12,6 +15,21 @@ type PathFlag string
 
 func (path *PathFlag) UnmarshalFlag(value string) error {
 	if value == "" {
+		return nil
+	}
+
+	if value == "-" {
+		tempf, err := ioutil.TempFile("", "fly-set-pipeline")
+		if err != nil {
+			return fmt.Errorf("failed to create a temp file")
+		}
+		defer tempf.Close()
+
+		_, err = io.Copy(tempf, os.Stdin)
+		if err != nil {
+			return fmt.Errorf("failed to write temp file: %s", err.Error())
+		}
+		*path = PathFlag(tempf.Name())
 		return nil
 	}
 
--- a/atc/path_flag.go
+++ b/atc/path_flag.go
@@ -6,10 +6,13 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/jessevdk/go-flags"
 )
+
+const tempFilePattern = "___dash-stdin-path-flag___"
 
 type PathFlag string
 
@@ -18,8 +21,8 @@ func (path *PathFlag) UnmarshalFlag(value string) error {
 		return nil
 	}
 
-	if value == "-" {
-		tempf, err := ioutil.TempFile("", "fly-set-pipeline")
+	if value == "-" && runtime.GOOS != "windows" {
+		tempf, err := ioutil.TempFile("", tempFilePattern)
 		if err != nil {
 			return fmt.Errorf("failed to create a temp file")
 		}
@@ -59,4 +62,8 @@ func (path *PathFlag) Complete(match string) []flags.Completion {
 	}
 
 	return comps
+}
+
+func (path *PathFlag) FromStdin() bool {
+	return strings.Index(string(*path), tempFilePattern) >= 0
 }
--- a/atc/path_flag.go
+++ b/atc/path_flag.go
@@ -6,7 +6,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"runtime"
 	"strings"
 
 	"github.com/jessevdk/go-flags"
@@ -21,7 +20,13 @@ func (path *PathFlag) UnmarshalFlag(value string) error {
 		return nil
 	}
 
-	if value == "-" && runtime.GOOS != "windows" {
+	if value == "-" {
+		_, err := os.Stat("/dev/stdin")
+		if err == nil {
+			*path = PathFlag("/dev/stdin")
+			return nil
+		}
+
 		tempf, err := ioutil.TempFile("", tempFilePattern)
 		if err != nil {
 			return fmt.Errorf("failed to create a temp file")
@@ -65,5 +70,5 @@ func (path *PathFlag) Complete(match string) []flags.Completion {
 }
 
 func (path *PathFlag) FromStdin() bool {
-	return strings.Index(string(*path), tempFilePattern) >= 0
+	return *path == "/dev/stdin" || strings.Index(string(*path), tempFilePattern) >= 0
 }
--- a/release-notes/latest.md
+++ b/release-notes/latest.md
@@ -9,6 +9,7 @@
 #### <sub><sup><a name="5810" href="#5810">:link:</a></sup></sub> feature
 
 * Reduce the allowed character set for Concourse valid identifiers. Only prints warnings instead of errors as a first step. #5810
+
 * Add `--team` flag for `fly pause-pipeline` command. #5917
 * Add `--team` flag for `fly hide-pipeline` command. #5917
 
--- a/release-notes/latest.md
+++ b/release-notes/latest.md
@@ -9,7 +9,6 @@
 #### <sub><sup><a name="5810" href="#5810">:link:</a></sup></sub> feature
 
 * Reduce the allowed character set for Concourse valid identifiers. Only prints warnings instead of errors as a first step. #5810
-
 * Add `--team` flag for `fly pause-pipeline` command. #5917
 * Add `--team` flag for `fly hide-pipeline` command. #5917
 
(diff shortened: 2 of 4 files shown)


chenbh:
> The second fly sp must be given -n, otherwise apply configuration? [yN]: bailing out. 

This should be caused by the shell redirecting `stdin` to the pipe, which means the interactions library gets an EOF when it tries to read. We can probably get around that by reading from `/dev/tty` when doing the user prompt. While this wouldn't work for Windows I'm not even sure if Windows supports pipes in the first place. 

Tagging @aoldershaw since you recently worked with terminals and Windows.

aoldershaw:
@chenbh @evanchaoli maybe it makes sense to always be in non-interactive mode if `Stdin` isn't a terminal? We already have a helper for this:

https://github.com/concourse/concourse/blob/3295bbd95ad6b84d923fad6b777fb6970067f1a0/fly/pty/open_raw_term_unix.go#L13-L15

evanchaoli:
@aoldershaw I wanted to do so but didn't figure out a way. Thanks for the info.

evanchaoli:
Just pushed a commit to make non-interactive when reading pipeline config from stdin. Also updated help message and release note.

See my tests:

```
$ concourse.fork/fly/fly sp --help
Usage:
  fly [OPTIONS] set-pipeline [set-pipeline-OPTIONS]

Application Options:
  -t, --target=                     Concourse target name
  -v, --version                     Print the version of Fly and exit
      --verbose                     Print API requests and responses
      --print-table-headers         Print table headers even for redirected output

Help Options:
  -h, --help                        Show this help message

[set-pipeline command options]
      -n, --non-interactive         Skips interactions, uses default values
          --no-color                Disable color output
          --check-creds             Validate credential variables against credential manager
      -p, --pipeline=               Pipeline to configure
      -c, --config=                 Pipeline configuration file, "-" stands for stdin  <== This line is updated
      -v, --var=[NAME=STRING]       Specify a string value to set for a variable in the pipeline
      -y, --yaml-var=[NAME=YAML]    Specify a YAML value to set for a variable in the pipeline
      -l, --load-vars-from=         Variable flag that can be used for filling in template values in configuration from a YAML file
          --team=                   Name of the team to which the pipeline belongs, if different from the target default
```

```
$ concourse.fork/fly/fly -t dev gp -p simple | concourse.fork/fly/fly -t dev sp -p simple2 -c -
<...omit pipeline content...>

pipeline created!
you can view your pipeline here: http://127.0.0.1:8080/teams/main/pipelines/simple2

the pipeline is currently paused. to unpause, either:
  - run the unpause-pipeline command:
    concourse.fork/fly/fly -t dev unpause-pipeline -p simple2
  - click play next to the pipeline in the web ui
```

evanchaoli:
@xtremerui Can you please help review this PR? Hopefully it may catch up with 6.4 as this is a tiny change. If user doesn't use `-`, then it doesn't impact anything.

aoldershaw:
@evanchaoli sorry, I think we've already cutoff new changes for 6.4.0, and are in the process of releasing it - is that true @xtreme-sameer-vohra?

aoldershaw:
Looks like going to non-interactive mode when stdin isn't a terminal breaks the fly integration tests, since the test environment isn't a terminal. Hmm...

evanchaoli:
@xtremerui  @aoldershaw I have fixed the broken test and added a new test case. I changed to not check TTY, instead, only when pipeline configure is read from stdin, then use non-interactive mode. Please take a look again. Thanks.

agurney:
In `bash`, and I think in `zsh` as well, you can already say
```sh
  fly -t new set-pipeline -p foo -n -c <(fly -t old get-pipeline -p foo)
```

to get this effect ("process substitution"). But maybe the `-` is a convenience for people using different shells, or who want to have a more involved pipeline.

release-notes/latest.md
```diff
@@ -1,3 +1,7 @@
 #### <sub><sup><a name="5842" href="#5842">:link:</a></sup></sub> feature
 
 * Added recover for panic error that used to crash the cluster. Now it should be less easy to panic (we hope!) and if it does, panic error could be found on Stderr and log. #5842
+
+#### <sub><sup><a name="5861" href="#5861">:link:</a></sup></sub> feature
+
+* @evanchaoli Make `--config` flag of `fly set-pipelines` command to support `-` for reading pipeline config from stdin. #5861
```
agurney, line 7: I think this also affects other fly commands that use `PathFlag`:
* `format-pipeline`
* `validate-pipeline`
* `execute` (the task config, plus vars-from-file)
*  `login` (the CACert - this one's a bit obscure though)

atc/path_flag.go
```diff
@@ -2,19 +2,39 @@ package atc
...
+		}
+		defer tempf.Close()
+
+		_, err = io.Copy(tempf, os.Stdin)
+		if err != nil {
+			return fmt.Errorf("failed to write temp file: %s", err.Error())
+		}
+		*path = PathFlag(tempf.Name())
+		return nil
+	}
```
agurney, line 36: It might be possible on Linux and Mac to get away with
```go
if value == "-" {
  *path = PathFlag("/dev/stdin")
  return nil
}
```

At least, I can do `fly -t old gp -p foo | fly -t new sp -p foo -n -c /dev/stdin` with current `fly`, and it works; I didn't test making that change in `PathFlag`. I'm pretty sure this will not work on Windows though. Do you think it's worth having this simpler way if `GOOS` is non-Windows?

evanchaoli: Good point. TBH, I always ignore Windows because I don't have a windows env in hand. I'll add a non-windows check.

release-notes/latest.md
evanchaoli: Yes, exactly. But I only intended to modify `set-pipeline`, others are just side-effects. Maybe we can just consider others as a hide feature.

evanchaoli:
@agurney I chose to use `-` because `tar` and `curl` support `-` for reading from stdin, which might be more user friendly. `-c /dev/stdin` might look a little weird for non-advanced users.

evanchaoli:
@xtremerui Can you please help this one?

xtremerui:
@aoldershaw do you have anymore concerns?

atc/path_flag.go
```diff
@@ -2,19 +2,40 @@ package atc
...
+
 type PathFlag string
 
 func (path *PathFlag) UnmarshalFlag(value string) error {
 	if value == "" {
 		return nil
 	}
 
+	if value == "-" && runtime.GOOS != "windows" {
+		tempf, err := ioutil.TempFile("", tempFilePattern)
```
aoldershaw, line 25: I may be wrong, but I interpreted @agurney's [comment](https://github.com/concourse/concourse/pull/5861/files#r454470474) to mean that on Mac and Linux, we can use `/dev/stdin`, but on Windows we'll have to keep making a temp file. So, something like:

```go
if value == "-" {
	if runtime.GOOS != "windows" {
		*path = PathFlag("/dev/stdin")
		return nil
	}
	tempf, err := ioutil.TempFile("", tempFilePattern)
	if err != nil {
		return fmt.Errorf("failed to create a temp file")
	}
	defer tempf.Close()

	_, err = io.Copy(tempf, os.Stdin)
	if err != nil {
		return fmt.Errorf("failed to write temp file: %s", err.Error())
	}
	*path = PathFlag(tempf.Name())
	return nil
}
```

aoldershaw: Or maybe it would be better to check for the existence of /dev/stdin, rather than checking the GOOS

evanchaoli: Ah... I misunderstood  @agurney's comment. Just fixed. And checking existence of /dev/stdin also works. I managed to test on a windows VM, and the latest commit worked well.

evanchaoli:
@aoldershaw Addressed your comments, can you please take a look again?

evanchaoli:
> Tested on mac by running `pbpaste | fly -t dev sp -p test -c -`, and it's pretty nice. Haven't tested on windows. Will wait for @xtremerui's approval before merging.

I have tested on windows.

Issue #5808, opened 2020-06-24
Pipeline instances


vito:
## Summary

This issue is for implementing concourse/rfcs#34 as described.

## Context

* RFC: concourse/rfcs#34

mouellet:
Hey @vito, I can take on this too if no one else is working on it right now...

Here's my plan so far:
* add `instance_vars` to the `pipelines` table as a `jsonb` column.
* edit unique constraint to include `instance_vars` column.
  * [ ] check for constraint w/ null, maybe 2 partial indexes?
  * [ ] check for size limit, maybe index md5 fo jsonb?

* add `--instance-var` flag to `fly set-pipeline` command and use them as params in template.
* add `instance_vars` attribute to `set_pipeline` step and pipe those in `exec.setPipelineSource.FetchPipelineConfig()`.
* add request body to api requests for pipeline instance filtering (see `go-concourse.concourse.team.managePipeline`).

What do you think?

vito:
@mouellet 👍 Go for it!

> * check for constraint w/ null, maybe 2 partial indexes?

Good catch - we ran into this problem in the past. Two partial indexes sounds appropriate.

> * check for size limit, maybe index md5 fo jsonb?

MD5 index would work, and we do that for resource versions, but maybe we should enforce a limit instead - before it even goes to the database. We don't really expect or want folks to be using gigantic identifiers. It might be better to fail fast, since it might be accidental from e.g. passing a dynamic value that was unexpectedly large. Maybe we can limit it to ~1k? (Just to pick a random somewhat sensible number.)

mouellet:
Hey @vito, I can see now that changing this unique key is a bit of work and means lots of code changes and I would rather do it once, so I need your input!

I found that `pipelineName` is used almost everywhere as a parameter and I thought replacing it with a parameter object (something like `atc.PipelineFilter`) instead of adding another parameter (i.e. `instanceVars`) would be better. 

i.e.: In `db.Team.Pipeline(pipelineName string) (Pipeline, bool, error)` would become 
`Pipeline(filter atc.PipelineFilter) (Pipeline, bool, error)` 
instead of 
`Pipeline(pipelineName string, instanceVars atc.InstanceVars) (Pipeline, bool, error)`

What do you think?

There are also somethings that I'm not sure if they should apply on instanced pipelines, like access or check containers... I'll make a list so we can talk about it...

vito:
[...]

PR #5864, opened 2020-07-09
valid identifiers

### LINKED ISSUE #5810: Enforce restrictions for valid identifiers
## Summary

This issue is for implementing concourse/rfcs#40 as described.

## Context

* RFC: concourse/rfcs#40
* Depends on: #5811

mouellet:

closes #5810.

`fly` commands print warnings when `team`s, `pipeline`s, `job`s, `step`s, `group`s, `resource`s, `resource_type`s or `var_source`s identifier are not valid (see [rfc#40](https://github.com/concourse/rfcs/blob/master/040-valid-identifiers) for spec).

## Contributor Checklist

- [x] [Signed] all commits
- [x] Updated [Documentation]
- [x] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/api/configserver/save.go
+++ b/atc/api/configserver/save.go
@@ -3,7 +3,6 @@ package configserver
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -70,19 +69,15 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
 	}
 
 	pipelineName := rata.Param(r, "pipeline_name")
-	if err := atc.ValidateIdentifier(pipelineName, "pipeline"); err != nil {
-		var got *atc.InvalidIdentifierError
-		if errors.As(err, &got) {
-			warnings = append(warnings, got.ConfigWarning())
-		}
+	warning := atc.ValidateIdentifier(pipelineName, "pipeline")
+	if warning != nil {
+		warnings = append(warnings, *warning)
 	}
 
 	teamName := rata.Param(r, "team_name")
-	if err := atc.ValidateIdentifier(teamName, "team"); err != nil {
-		var got *atc.InvalidIdentifierError
-		if errors.As(err, &got) {
-			warnings = append(warnings, got.ConfigWarning())
-		}
+	warning = atc.ValidateIdentifier(teamName, "team")
+	if warning != nil {
+		warnings = append(warnings, *warning)
 	}
 
 	if checkCredentials {
--- a/atc/api/configserver/save.go
+++ b/atc/api/configserver/save.go
@@ -69,7 +69,22 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
 	}
 
 	pipelineName := rata.Param(r, "pipeline_name")
+	err := atc.ValidateIdentifier(pipelineName, "pipeline")
+	if err != nil {
+		warnings = append(warnings, atc.ConfigWarning{
+			Type:    "invalid_identifier",
+			Message: err.Error(),
+		})
+	}
+
 	teamName := rata.Param(r, "team_name")
+	err = atc.ValidateIdentifier(teamName, "team")
+	if err != nil {
+		warnings = append(warnings, atc.ConfigWarning{
+			Type:    "invalid_identifier",
+			Message: err.Error(),
+		})
+	}
 
 	if checkCredentials {
 		variables := creds.NewVariables(s.secretManager, teamName, pipelineName, false)
--- a/atc/api/configserver/save.go
+++ b/atc/api/configserver/save.go
@@ -3,6 +3,7 @@ package configserver
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -69,21 +70,19 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
 	}
 
 	pipelineName := rata.Param(r, "pipeline_name")
-	err := atc.ValidateIdentifier(pipelineName, "pipeline")
-	if err != nil {
-		warnings = append(warnings, atc.ConfigWarning{
-			Type:    "invalid_identifier",
-			Message: err.Error(),
-		})
+	if err := atc.ValidateIdentifier(pipelineName, "pipeline"); err != nil {
+		var got *atc.InvalidIdentifierError
+		if errors.As(err, &got) {
+			warnings = append(warnings, got.ConfigWarning())
+		}
 	}
 
 	teamName := rata.Param(r, "team_name")
-	err = atc.ValidateIdentifier(teamName, "team")
-	if err != nil {
-		warnings = append(warnings, atc.ConfigWarning{
-			Type:    "invalid_identifier",
-			Message: err.Error(),
-		})
+	if err := atc.ValidateIdentifier(teamName, "team"); err != nil {
+		var got *atc.InvalidIdentifierError
+		if errors.As(err, &got) {
+			warnings = append(warnings, got.ConfigWarning())
+		}
 	}
 
 	if checkCredentials {
--- a/atc/api/pipelineserver/rename.go
+++ b/atc/api/pipelineserver/rename.go
@@ -3,7 +3,6 @@ package pipelineserver
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io/ioutil"
 	"net/http"
 
@@ -46,22 +45,10 @@ func (s *Server) RenamePipeline(pipeline db.Pipeline) http.Handler {
 		}
 
 		w.Header().Set("Content-Type", "application/json")
-		w.WriteHeader(http.StatusOK)
-		s.writeResponse(w, atc.SaveConfigResponse{Warnings: warnings})
+		err = json.NewEncoder(w).Encode(atc.SaveConfigResponse{Warnings: warnings})
+		if err != nil {
+			s.logger.Error("failed-to-encode-response", err)
+			w.WriteHeader(http.StatusInternalServerError)
+		}
 	})
-}
-
-func (s *Server) writeResponse(w http.ResponseWriter, response atc.SaveConfigResponse) {
-	responseJSON, err := json.Marshal(response)
-	if err != nil {
-		w.WriteHeader(http.StatusInternalServerError)
-		fmt.Fprintf(w, "failed to generate error response: %s", err)
-		return
-	}
-
-	_, err = w.Write(responseJSON)
-	if err != nil {
-		w.WriteHeader(http.StatusInternalServerError)
-		return
-	}
 }
--- a/atc/api/pipelineserver/rename.go
+++ b/atc/api/pipelineserver/rename.go
@@ -2,7 +2,6 @@ package pipelineserver
 
 import (
 	"encoding/json"
-	"errors"
 	"io/ioutil"
 	"net/http"
 
@@ -30,11 +29,9 @@ func (s *Server) RenamePipeline(pipeline db.Pipeline) http.Handler {
 		}
 
 		var warnings []atc.ConfigWarning
-		if err = atc.ValidateIdentifier(rename.NewName, "pipeline"); err != nil {
-			var got *atc.InvalidIdentifierError
-			if errors.As(err, &got) {
-				warnings = append(warnings, got.ConfigWarning())
-			}
+		warning := atc.ValidateIdentifier(rename.NewName, "pipeline")
+		if warning != nil {
+			warnings = append(warnings, *warning)
 		}
 
 		err = pipeline.Rename(rename.NewName)
--- a/atc/api/pipelineserver/rename.go
+++ b/atc/api/pipelineserver/rename.go
@@ -2,6 +2,7 @@ package pipelineserver
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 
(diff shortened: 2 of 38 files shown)


aoldershaw:
Hey @mouellet, thanks for another awesome contribution! I haven't had much of a chance to look at this, and probably won't for another day or two - we're busy trying to get 6.4.0 out the door. But will try to review it soon!

atc/api/teams_test.go
```diff
@@ -606,8 +631,27 @@ var _ = Describe("Teams API", func() {
...
+						Expect(response.StatusCode).To(Equal(http.StatusOK))
+					})
+
+					Context("when a warning occurs", func() {
+
+						BeforeEach(func() {
+							requestBody = `{"name":"_some-new-name"}`
+						})
+
+						It("returns a waring in the response body", func() {
```
chenbh, line 644:
```suggestion
						It("returns a warning in the response body", func() {
```

atc/configwarning.go
```diff
@@ -1,6 +1,46 @@
...
+	}
+
+	if e.Context != nil {
+		warning.Message = fmt.Sprintf("%s: %s", strings.Join(e.Context, ""), e.Error())
+	}
+
+	return warning
+}
+
+var validIdentifiers = regexp.MustCompile(`^\p{L}[\p{L}\d\-.]*$`)
```
chenbh, line 36: The [RFC](https://github.com/concourse/rfcs/blob/master/040-valid-identifiers/proposal.md#proposal) mentions that

> All letters will be converted to lowercase. This is to enforce case-insensitivity and to present a consistent UI/UX while still supporting languages that don't have any casing (e.g. Japanese).

We should probably just emit a warning for now. 

@vito in the future, do you see this as the CLI automatically converting the case and logging a warning; or just fail outright?

atc/api/configserver/save.go
```diff
@@ -69,7 +69,22 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
 	}
 
 	pipelineName := rata.Param(r, "pipeline_name")
+	err := atc.ValidateIdentifier(pipelineName, "pipeline")
+	if err != nil {
+		warnings = append(warnings, atc.ConfigWarning{
```
aoldershaw: Based on the RFC, we should be able to save existing pipelines with invalid names, but we should disallow creating new pipelines:

> To ease this pain, we can allow existing data to stay as-is, and **only enforce the identifier rules for newly created teams and pipelines**

So, maybe we should add a check to determine whether it's a new pipeline or not (i.e. if the pipeline exists yet or not) - if it is new, we error with a 400 Bad Request, and if not, we add the warning

atc/api/pipelineserver/rename.go
```diff
@@ -28,13 +29,39 @@ func (s *Server) RenamePipeline(pipeline db.Pipeline) http.Handler {
 			return
 		}
 
+		var warnings []atc.ConfigWarning
+		err = atc.ValidateIdentifier(rename.NewName, "pipeline")
+		if err != nil {
+			warnings = append(warnings, atc.ConfigWarning{
```
aoldershaw: I think we should error explicitly here with a 400 Bad Request. I think the idea of the warnings is to remove the friction when existing pipelines already have invalid identifiers, but we should prevent renaming to an invalid identifier

atc/api/configserver/save.go
aoldershaw: Actually, this makes me realize that we should be strict about all the identifier rules for newly created pipelines (all of the changes to `configvalidate` and `step_validator` as well). Not sure what the best way to do this is - the easiest thing I can think of is to special case on `invalid_identifier` warnings and return a 400 error if the pipeline is new. Open to any ideas you have!

atc/api/pipelineserver/rename.go
```diff
@@ -28,13 +29,39 @@ func (s *Server) RenamePipeline(pipeline db.Pipeline) http.Handler {
...
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprintf(w, "failed to generate error response: %s", err)
+		return
+	}
+
+	_, err = w.Write(responseJSON)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
```
aoldershaw, line 66:
```suggestion
	err := json.NewEncoder(w).Encode(response)
	if err != nil {
		logger.Error("failed-to-encode-response", err)
		w.WriteHeader(http.StatusInternalServerError)
	}
```

Which you can maybe inline. By the way, according to the [godocs](https://pkg.go.dev/net/http?tab=doc#ResponseWriter) `WriteHeader` can only be called once, and there's an implicit `w.WriteHeader(http.StatusOK)` on first `Write` - so, I'd suggest getting rid of our `w.WriteHeader(http.StatusOK)` so that we can properly write the `http.StatusInternalServerError`

atc/api/teamserver/rename.go
```diff
@@ -35,6 +36,15 @@ func (s *Server) RenameTeam(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	var warnings []atc.ConfigWarning
+	err = atc.ValidateIdentifier(rename.NewName, "team")
+	if err != nil {
+		warnings = append(warnings, atc.ConfigWarning{
```
aoldershaw: Likewise, renaming a team to an invalid name should error

atc/api/teamserver/set.go
```diff
@@ -62,6 +69,16 @@ func (s *Server) SetTeam(w http.ResponseWriter, r *http.Request) {
 	} else if acc.IsAdmin() {
 		hLog.Debug("creating team")
 
+		var warnings []atc.ConfigWarning
+		err = atc.ValidateIdentifier(atcTeam.Name, "team")
+		if err != nil {
+			warnings = append(warnings, atc.ConfigWarning{
```
aoldershaw: Here, we should always error - we've already determined that the team is new, so we should prevent making a new team with an invalid name

atc/step_validator.go
```diff
@@ -66,7 +66,7 @@ func (validator *StepValidator) VisitTask(plan *TaskStep) error {
 	if plan.Config != nil && (plan.Config.RootfsURI != "" || plan.Config.ImageResource != nil) && plan.ImageArtifactName != "" {
 		validator.recordWarning(ConfigWarning{
 			Type:    "pipeline",
-			Message: "specifies image: on the step but also specifies an image under config: - the image: on the step takes precedence",
+			Message: validator.annotate("specifies image: on the step but also specifies an image under config: - the image: on the step takes precedence"),
```
aoldershaw, line 72: good catch!

atc/configwarning.go
```diff
@@ -3,21 +3,44 @@ package atc
...
+	if e.Context != nil {
+		warning.Message = fmt.Sprintf("%s: %s", strings.Join(e.Context, ""), e.Error())
+	}
+
+	return warning
+}
+
 var validIdentifiers = regexp.MustCompile(`^\p{L}[\p{L}\d\-.]*$`)
 
 func ValidateIdentifier(identifier string, context ...string) error {
```
aoldershaw, line 38: Does it make sense for this to just return a `*ConfigWarning` instead of an `error`? That way, we won't need to use `errors.As` to ensure the error type is correct everywhere?

atc/configwarning_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"github.com/concourse/concourse/atc"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("ValidateIdentifier", func() {
+	Context("when an identifier starts with valid letter", func() {
```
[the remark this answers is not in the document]
aoldershaw: No need to change it by the way, the way you have it already looks good and is easy to read - just showing an alternative approach

vito:
@aoldershaw We decided to go with warnings instead of errors until pipelines instances (#5808) lands. If we do this first, some users would become blocked as their previously-valid name generation schemes would no longer be functional and they would have to resort to pretty ugly sanitization schemes like `github.com-SLASH-foo`.

This concern was mentioned here: https://github.com/concourse/rfcs/pull/40#issuecomment-620611206

Right now the RFC [states](https://github.com/concourse/rfcs/blob/594ef3ca470514691b6cccca69e60bb2f7d2e135/040-valid-identifiers/proposal.md#easing-the-transition)...:

```
Existing data will still be fully functional and writable (i.e. updated with fly set-pipeline, fly set-team), and Concourse can emit warnings for any invalid identifiers (including the pipeline/team name itself) instead of erroring out completely.

After one year, we can turn these warnings into errors.
```

When I wrote this I originally intended this to apply only to already-existing data, but now that we're here it might be wise to just have all cases result in warnings and then switch to errors after some period of time so that folks have enough time to adapt their patterns to things like pipeline instances (...and so we have time to polish pipeline instances).

This would simplify the transition from 3 steps (Warn/Warn, Warn/Error w/ #5808 done, Error/Error after 1 year) down to 2 steps (Warn/Warn, Error/Error after 1 year) which may be easier to reason about.

What do y'all think? Should I amend the RFC?

atc/configwarning.go
mouellet: It does, but I thought it would ease the transition from warning to errors in the next phase. I see those error type checks as temporary for now. Let me know if you'd rather have it return `*ConfigWarning`, it would make it cleaner for now...

atc/configvalidate/validate.go
```diff
@@ -54,6 +56,13 @@ func Validate(c Config) ([]ConfigWarning, []string) {
 	return warnings, errorMessages
 }
 
+func validateIdentifier(identifier string) error {
+	if identifier != "" && !regexp.MustCompile(`^\p{L}[\p{L}\d\-.]*$`).MatchString(identifier) {
```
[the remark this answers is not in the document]
mouellet: Yeah your right! It should be an invalid identifier!

atc/configwarning_test.go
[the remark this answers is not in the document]
mouellet: No I'll change it! I really like that! It looks great! Thanks!!

atc/configwarning.go
mouellet: Isn't `\p{L}` short for `[\p{Ll}\p{Lu}\p{Lt}\p{Lm}\p{Lo}]`? If we remove `\p{Lu}` from that we would have valid identifier be lowercase only and support languages without case. Then there's no need for auto-converting case, right?

It's not pretty, but I think `^[\p{Ll}\p{Lt}\p{Lm}\p{Lo}][\p{Ll}\p{Lt}\p{Lm}\p{Lo}\d\-.]*$` would do that...

atc/configwarning.go
aoldershaw: I think I'd prefer if it returned `*ConfigWarning`, since then when we change it to `error` we'll get compiler errors wherever it's called. Right now, if `ValidateIdentifier` were to return an error that was not an `*InvalidIdentifierError`, we just ignore it (since `errors.As` would return `false`). But if we get compiler errors, we're forced to think about how we handle it (and there aren't that many places we'd need to change, so it shouldn't be too much effort!)

fly/commands/internal/setpipelinehelpers/atc_config.go
```diff
@@ -56,6 +58,18 @@ func (atcConfig ATCConfig) Set(yamlTemplateWithParams templatehelpers.YamlTempla
...
+			Type:    w.Type,
+			Message: w.Message,
+		})
+	}
+
+	if len(atcConfig.CommandWarnings) > 0 {
+		displayhelpers.ShowWarnings(atcConfig.CommandWarnings)
+	}
+
 	diffExists := diff(existingConfig, newConfig)
```
aoldershaw, line 74: Love this client side validation, but since we print it before the diff, it could be off screen when the pipeline yaml is long. What if we print it after we print out the diff?

atc/configwarning.go
```diff
@@ -1,6 +1,24 @@
...
 	Message string `json:"message"`
 }
+
+var validIdentifiers = regexp.MustCompile(`^[\p{Ll}\p{Lt}\p{Lm}\p{Lo}][\p{Ll}\p{Lt}\p{Lm}\p{Lo}\d\-.]*$`)
+
+func ValidateIdentifier(identifier string, context ...string) *ConfigWarning {
+	if identifier != "" && !validIdentifiers.MatchString(identifier) {
+		return &ConfigWarning{
+			Type:    "invalid_identifier",
+			Message: fmt.Sprintf("%s: %s", strings.Join(context, ""), fmt.Sprintf("'%s' is not a valid identifier", identifier)),
```
aoldershaw, line 20: I wonder if we can provide more context as to what a valid identifier is, since to a new user (or to an old user who isn't familiar with the valid identifiers RFC), it might not be obvious. While we will have this documented, perhaps it's better to tell the user what exactly was wrong with their identifier, or give a brief summary of what a valid identifier is. That could just be the responsibility of `fly` when it encounters an `invalid_identifier` warning, or we could encode this information into the `Message` itself...not sure what the best approach is.

Since the validation logic is pretty simple, maybe it's worth not using a regex at all (which looks a bit intimidating at this point) and instead do the validation manually. This way, we can give more fine-grained warning messages that can help users out (e.g. `'some_job' is not a valid identifier: illegal character '_'` or `'123abc' is not a valid identifier: must start with a letter`)

What do you think about that approach?

mouellet: Sure! Let's add a brief summary for _must start with_ and _illegal character_. Should we print a single item (the left most one) or a list?

> That could just be the responsibility of `fly` when it encounters an `invalid_identifier` warning, or we could encode this information into the `Message` itself...not sure what the best approach is.
> 

Encoding the details in the `Message` will keep consistency between client-side and server-side validation.

> Since the validation logic is pretty simple, maybe it's worth not using a regex at all (which looks a bit intimidating at this point) and instead do the validation manually.

If it's ok with you, I'd keep using a regex for now... There's a simple way to catch the details using them.

atc/configwarning.go
vito: @mouellet Neat! Yeah, that works as expected. I'll amend the RFC with this as well.

PR #5866, opened 2020-07-09
Allow usage of images from custom resource types

jvshahid:
# Existing Issue

Not that I am aware of.

# Why do we need this PR?

This PR adds support for Task's ImageResources that depend on a custom resource type and is a continuation of #318.  Prior to this change the fly CLI would error with the following error when the task container image used a custom resource type:

> initializing
> no workers satisfying: resource type 'resource-type, platform 'linux', tag 'offline'
> errored

* Set the VersionedResourceTypes on the TaskPlan.

# Contributor Checklist
- [ ] Unit tests

# Reviewer Checklist
> fill out this section.
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


jvshahid:
The issue can be reproduced using the following pipeline:

```yaml
resource_types:
- name: docker-image-resource
  type: docker-image
  privileged: true
  source:
    repository: concourse/docker-image-resource
    tag: latest

resources:
- name: ubuntu-docker-image
  type: docker-image-resource
  source:
    repository: ubuntu
    tag: latest

jobs:
- name: reproduce-bug
  plan:
  - get: ubuntu-docker-image
  - task: task
    image: ubuntu-docker-image
    config:
      platform: linux
      run:
        path: bash
        args:
        - -c
        - |
          echo hi
```

and using the following command to execute a task:

``` yaml
$ fly -t ci execute -c task.yml -j pipeline-name/reproduce-bug --image ubuntu-docker-image

executing build 22993 at https://eagle.ci.cf-app.com/builds/22993 
initializing
no workers satisfying: resource type 'docker-image-resource', platform 'linux'
errored
```

atc/testhelpers/walker.go
```diff
@@ -0,0 +1,168 @@
+package testhelpers
+
+import "github.com/concourse/concourse/atc"
+
+type PlanWalker func(atc.Plan) (atc.Plan, error)
+
+// Walks the plan tree structure calling walker for each encountered Plan,
+// including the root.
+func WalkPlan(plan atc.Plan, walker PlanWalker) (atc.Plan, error) {
```
aoldershaw, line 9: This looks a lot like `atc.Plan.Each`:

https://github.com/concourse/concourse/blob/0565fd9ebe7cbd86b1f85f6c49d0ea52e145b421/atc/plan.go#L36

Maybe just worth modifying `atc.Plan.Each` to take a `func(*atc.Plan) error` instead?

jvshahid: Good call. I don't strictly need to return an error in the new matcher that I introduced. I will modify both matchers to use `plan.Each`.

jvshahid: I pushed the changes. Please, take another look.

fly/integration/execute_test.go
```diff
@@ -246,6 +247,13 @@ run:
 			atcServer.RouteToHandler("POST", "/api/v1/teams/main/pipelines/some-pipeline/builds",
 				ghttp.CombineHandlers(
 					ghttp.VerifyRequest("POST", "/api/v1/teams/main/pipelines/some-pipeline/builds"),
+					testhelpers.VerifyResourceTypes(atc.VersionedResourceType{
```
clarafu, line 250: I'm wondering if we could test the addition of `VersionedResourceTypes` to the execute task plan by testing that the `atc.TaskPlan` contains the proper `atc.VersionedResourceTypes`? So we would add
`VerifyPlan(expectedPlan)` with the expectedPlan contain the proper `VersionedResourceTypes` rather than adding `VerifyResourceTypes(..`, and that would allow us to avoid adding this new resource types matcher. Let me know what you think!

atc/testhelpers/match_plan_matcher.go
```diff
@@ -79,73 +79,17 @@ func (matcher *PlanMatcher) NegatedFailureMessage(actual interface{}) string {
...
+		if plan.Get != nil {
+			if plan.Get.VersionFrom != nil {
+				planID := atc.PlanID("<stripped>")
+				plan.Get.VersionFrom = &planID
+			}
 		}
-	}
+	})
 
-	return plan, ids
```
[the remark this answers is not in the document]
jvshahid: I couldn't convince myself that this used to work. I believe the function always returned a `nil` value for `ids`.  I'm happy to fix it as part of this PR.

fly/integration/execute_test.go
jvshahid: I started writing the new matcher because I was being lazy and didn't want to figure out the structure of the pipeline in the test. It also felt brittle to break the test, for example, when the test fixture changes, because the expected pipeline didn't match. That said, I'm happy to get rid of the matcher if you do not want to maintain yet another matcher. I don't have a strong opinion.

atc/testhelpers/match_plan_matcher.go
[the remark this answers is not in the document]
clarafu: Oh yeah there was definitely a bug in the test 😅  It should be testing against the IDs that it stripped. I think it'd be great if we could fix it as part of this PR, I think the only thing needed to fix it is to append the ids to a list and return that? Thanks for catching that!

fly/integration/execute_test.go
jvshahid: Looking at this again, it looks like all the other tests share the same `expectedPlan`, so I could add the `VersionedResourceType` to the task config in the `BeforeEach`. I am still worried about the goal of the test not being clear over time. For example if a new Task or Get is added, the author of the change could simply modify the top-level `expectedPlan` to add the new task, without `VersionedResourceTypes`. This change would make this test trivially pass.

jvshahid:
Since I opened the PR, I have found another bug, that is somewhat related to this one. If you use the following pipeline:

```
resource_types:
- name: docker-image-resource
  type: docker-image
  privileged: true
  source:
    repository: concourse/docker-image-resource
    tag: latest

resources:
- name: ubuntu-docker-image
  type: docker-image-resource
  source:
    repository: ubuntu
    tag: latest

jobs:
- name: reproduce-bug
  plan:
  - get: ubuntu-docker-image
  - task: task
    image: ubuntu-docker-image
    tags:
    - offline
    config:
      platform: linux
      run:
        path: bash
        args:
        - -c
        - |
          echo hi
```

You won't be able to fly execute the task.  If you try to do it using the following command:

```
$ fly -t ci execute -j wrong-tag/reproduce-bug --image ubuntu-docker-image --tag offline -c task.yml
executing build 24022 at https://xxx.xxx.xxx.xxx/builds/24022
initializing
no workers satisfying: resource type 'docker-image-resource', platform 'linux', tag 'offline'
errored
```

I was able to fix the issue by using `GetPlan` instead of setting the taskConfig's `ImageResource` and setting `ImageArtifactName` on the TaskPlan instead.  You can take a look at the change here: https://github.com/jvshahid/concourse/commit/a8dd381e0c63d4da3c7f1041946ecc3e67b6df05.

Should I combine the two changes into one PR?

jvshahid:
I edited the above comment since when I initially wrote it, I used the wrong version of the fly CLI and got very confused.

fly/integration/execute_test.go
clarafu: I see, I definitely understand your concern. I think preferably we would want to avoid adding another matcher but if you feel like adding a resource type matcher would be the best method to testing this, then I would be for it.

I tried seeing what it would look like if we were to test using the `expectedPlan`, I'd love to get your opinion on it! The first thing I did was extracting out the `atc.TaskPlan` from the `expectedPlan`

```
taskPlan = planFactory.NewPlan(atc.TaskPlan{
	Name: "one-off",
	Config: &atc.TaskConfig{
		Platform: "some-platform",
		ImageResource: &atc.ImageResource{
			Type: "registry-image",
			Source: atc.Source{
				"repository": "ubuntu",
			},
		},
		Inputs: []atc.TaskInputConfig{
			{Name: "fixture"},
		},
		Params: map[string]string{
			"FOO":   "bar",
			"BAZ":   "buzz",
			"X":     "1",
			"EMPTY": "",
		},
		Run: atc.TaskRunConfig{
			Path: "find",
			Args: []string{"."},
		},
	},
})

expectedPlan = planFactory.NewPlan(atc.DoPlan{
	planFactory.NewPlan(atc.AggregatePlan{
		planFactory.NewPlan(atc.ArtifactInputPlan{
			ArtifactID: 125,
			Name:       filepath.Base(buildDir),
		}),
	}),
	taskPlan,
})
```

And then modifying the `VersionedResourceTypes` within the `Task` for the `taskPlan`.

```
Context("when there is a pipeline job with the same input", func() {
	BeforeEach(func() {
		taskPlan.Task.VersionedResourceTypes = atc.VersionedResourceTypes{
			atc.VersionedResourceType{
				ResourceType: atc.ResourceType{
					Name:   "resource-type",
					Type:   "s3",
					Source: atc.Source{},
				},
			},
		}
```

Then we would verify the plan is the same as the expectedPlan

```
atcServer.RouteToHandler("POST", "/api/v1/teams/main/pipelines/some-pipeline/builds",
	ghttp.CombineHandlers(
		ghttp.VerifyRequest("POST", "/api/v1/teams/main/pipelines/some-pipeline/builds"),
		VerifyPlan(expectedPlan),
		func(w http.ResponseWriter, r *http.Request) {
			http.SetCookie(w, &http.Cookie{
				Name:    "Some-Cookie",
				Value:   "some-cookie-data",
				Path:    "/",
				Expires: time.Now().Add(1 * time.Minute),
			})
		},
		ghttp.RespondWith(201, `{"id":128}`),
	),
)
```

Originally that would've been enough but because we are now verifying the build plan within the `"when there is a pipeline job with the same input"` test, it also has an extra change that adds build inputs to the plan so we would need to have the expectedPlan within the `BeforeEach` of the test respect that build inputs modification. I think this actually adds value to this test because it covers this [part of the codebase](https://github.com/concourse/concourse/blob/d64843ce1bbab2309213ff677e25aadb950fc33b/fly/commands/internal/executehelpers/inputs.go#L46-L89).

```
expectedPlan = planFactory.NewPlan(atc.DoPlan{
	planFactory.NewPlan(atc.AggregatePlan{
		planFactory.NewPlan(atc.GetPlan{
			Name: "fixture",
			VersionedResourceTypes: atc.VersionedResourceTypes{
				atc.VersionedResourceType{
					ResourceType: atc.ResourceType{
						Name:   "resource-type",
						Type:   "s3",
						Source: atc.Source{},
					},
				},
			},
		}),
	}),
	taskPlan,
})
```

jvshahid: I don't have a strong opinion either way. I will make the changes proposed above tomorrow morning in a separate commit.

jvshahid: I pushed the changes requested above.

atc/testhelpers/match_plan_matcher.go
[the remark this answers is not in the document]
jvshahid: Fixing this is proving to be a little tricky. The execute command creates a new plan factory using the current timestamp to seed the factory. I am not sure how we can test this in the integration test. Any ideas?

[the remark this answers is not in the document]
clarafu: You're totally right, that is pretty impossible to test by checking to see if the ids stripped in the actual vs expected is the same.

Maybe we can test the plan ids a different way, by just checking to see if the ids within the actual plan are all unique? It would be basically the exact same as the `planner_test.go` here

https://github.com/concourse/concourse/blob/350e79a1385df809cbabba41040635cc3d715b43/atc/builds/planner_test.go#L670-L682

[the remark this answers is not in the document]
jvshahid: I pushed the change.

clarafu:
Thanks for finding the other bug! Hm I think that can be a separate pull request because from first glance it seems to be a problem with it not grabbing the image from the get step?

clarafu:
Oh it looks like you have a few commits that need to be signed off, you can see which commits need to be fixed here https://github.com/concourse/concourse/pull/5866/checks?check_run_id=890802891.

jvshahid:
I think I fixed all commits to include the sign off header and force pushed the branch.

Issue #4089, opened 2019-07-02
A more coherent Build UI


iautom8things:
## What challenge are you facing?

The Build UI is cryptic and hard to understand, even for someone who works with Concourse on a regular basis.

Somehow this pipeline:
<img width="582" alt="image" src="https://user-images.githubusercontent.com/522040/60546563-30198980-9ce3-11e9-8e2d-ee7823a81ca6.png">

Yields this UI:
<img width="404" alt="Screen Shot 2019-07-02 at 12 38 34 PM" src="https://user-images.githubusercontent.com/522040/60546716-838bd780-9ce3-11e9-8619-10c028107995.png">

I'm not sure the whole ⬆️ with _nested_ ⬇️ is very clear to begin with, but having a case where it goes 3 levels deep (⬆️ version, ⬇️ opened-pr, ⬆️ opened-pr) -- what does that mean?  How do I explain this to someone that doesn't have the pipeline definition in front of them?

_I suspect_  this has something to do with the `on_success` and `on_failure`.  But this doesn't even make that much sense because _why would they be under the `version` step?_   They are actually hooks for the entire build, not just a single step.

I'm also afraid to see what it would look like if I added `on_error` and `ensure`.  I imagine we'd be playing Dance Dance Revolution at that point.

## What would make this better?

I really don't know what the perfect UI would look like.  I just know that _this_ isn't it.  I love Concourse, and I just wanted to spark the conversation around making the UI a little more friendlier, understandable and thus more approachable.

## Are you interested in implementing this yourself?

I'm not a designer/FE Developer so I would probably make things even worse if I tried.  I would be happy to help give feedback on any potential solutions.

vito:
Yeah I agree with a lot of this. The build plan started simple but as we added more step 'modifiers' it became harder and harder to understand.

There's some movement on this here, which you may be interested in: https://github.com/concourse/concourse/pull/4942

I think getting away from the cute UI metaphors and showing a distilled form of the YAML that yielded the build plan would be a fun direction to go.

taylorsilva:
# Issue Bankruptcy

This issue is being closed as part of 2025's **Issue Bankruptcy** declaration. [See here for details](https://blog.concourse-ci.org/posts/2025-01-06-concourse-2024-review/)

If you think this issue still has merit then feel free to open a new issue and reference this one.

PR #5887, opened 2020-07-16
Add experimental `across` step for running build plans across a matrix of values

aoldershaw:

Implements the `across` step for static lists as described in concourse/rfcs#29

Screen Shot 2020-07-25 at 1 19 09 AM

Screen Shot 2020-07-25 at 1 31 42 AM

* Adds a new experimental `across` step
  * The step runs as nested `in_parallel` steps
* In the build plan UI, values of `across` steps are displayed as key-value pairs in sub headers. Expanding the header reveals the substep
  * Previously, step headers stuck to the top of the screen while viewing the build log. Sub headers now stick to the bottom of their parent header so you don't lose context while viewing build logs for a substep (warning: a bit of scrappy Javascript required for that feature)

Screen Shot 2020-07-25 at 1 40 49 PM

## Release Note

* This feature must be enabled using `--enable-across-step` as it is subject to change - don't be alarmed if your pipelines with `across` stop working in a future release!
* There's no official documentation yet, but there are some examples of usage patterns in [the RFC](https://github.com/vito/rfcs/blob/spatial-resources/029-across-step/proposal.md)

## Notes to reviewer:

Happy to receive any feedback on the current approach/design. Mostly worked on this to familiarize myself with this part of the codebase (and because I think it's pretty cool!)

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [x] PR acceptance performed


aoldershaw:
p.s. not sure about `fly watch`...it's hard to make the output easy to follow when there's a `max_in_flight > 1`. There's currently no real indication about the vars that it's currently running with, since there's nothing baked into the build events (it's all derived from the plan in the UI)

aoldershaw:
Also realizing now that the colour indicator above each tab will likely be pretty useless for people with colour blindness. I think there's a lot of places in our UI lacking good accessibility, but probably better not to add on to the pile. We could maybe add an icon to each tab (like we have in the step headers)

jamieklassen:
maybe `fly watch` could just show the outputs in order -- show the output from the first combination, then the second, etc. While the first is running, stream its output. If the second finishes during this time just buffer its events and then when it's time you can show them all at once.

vito:
Just a quick note: I don't think existing steps like `in_parallel` and `aggregate` work nicely with `fly watch` either. It would obviously be great if all these steps did, but if it ends up being the last 20% that takes 80% of the time, it might be better to split it out to a separate PR that addresses all of them.

aoldershaw:
@jamieklassen re: point 2 - I thought about doing a flat list, but initially decided against it for a couple reasons:

1. I felt nesting might be nicer to see at a glance what var is causing problems. e.g. if you're running `across` go version and platform, if there's something breaking in a go version, you'd see right away which version is affected. On the flip side, if the issue lies in the platform, you would have to drill into the go versions to see that, so maybe that point is moot.
2. Related, but I was worried it could be a bit overwhelming showing every combination simultaneously, and could be hard to pinpoint what's causing failures

Not at all hard-set on this design, though - happy to change it if you think that'll be better

aoldershaw:
One uncertainty I have about the UX is displaying lots of complex objects in tabs - e.g. when we have a `github-prs` var source, I imagine it'd emit objects like `{"number": 1234, "author: "aoldershaw", "created_at": "2020-07-16T15:14:48Z"}`, which could be very noisy in the UI if we display the entire object (when often we'd only care about e.g. the number)

chenbh:
> p.s. not sure about fly watch...it's hard to make the output easy to follow when there's a max_in_flight > 1.

FWIW docker gets around this by colour coding and prepending each entry with the container name
```
web_1     | {"timestamp":"2020-07-16T15:53:34.027801900Z","level":"debug","source":"atc","message":"atc.metrics.metric-initialize","data":{"attributes":{},"buffer-size":1000,"host":"c8a3750baa60","session":"2"}}
db_1      | The files belonging to this database system will be owned by user "postgres".
worker_1  | {"timestamp":"2020-07-16T15:53:35.578405600Z","level":"info","source":"baggageclaim","message":"baggageclaim.using-driver","data":{"driver":"overlay"}}
```

vito:
@aoldershaw Version numbers are also key-value JSON objects, and we render them like this:

image

We could probably follow the same structure here. 🤔 

> * I felt nesting might be nicer to see at a glance what var is causing problems. e.g. if you're running `across` go version and platform, if there's something breaking in a go version, you'd see right away which version is affected. On the flip side, if the issue lies in the platform, you would have to drill into the go versions to see that, so maybe that point is moot.

The flip side here is what I'd be worried about - I think if I'm viewing a build to see a failure I'd rather just see them all as quickly as possible so my eye can travel to the failed combinations. I might even be able to visually identify patterns in the failures pretty quickly.

For example, let's say I'm faced with a red build containing the following combinations:

```yaml
- foo: 1, bar: a, baz: alpha
- foo: 1, bar: a, baz: beta
- foo: 1, bar: b, baz: alpha # red
- foo: 1, bar: b, baz: beta
- foo: 2, bar: a, baz: alpha
- foo: 2, bar: a, baz: beta
- foo: 2, bar: b, baz: alpha # red
- foo: 2, bar: b, baz: beta
```

From this I could determine that `bar: b, baz: alpha` is a "bad combination"- admittedly with a bit of mental gymnastics, but I at least wouldn't have to click around too much (and in the meantime lose track of which combinations I've identified as bad along the way).

> 2\. Related, but I was worried it could be a bit overwhelming showing every combination simultaneously, and could be hard to pinpoint what's causing failures

I wonder if we could somehow summarize failure rates for each value? 🤔

* `foo: 1` had 1/4 failures
* `foo: 2` had 1/4 failures
* `bar: a` had 0/4 failures
* `bar: b` had 2/4 failures
* `baz: alpha` had 2/4 failures
* `baz: beta` had 0/4 failures

(Note that these become /4 instead of /6 since one value is already determined.)

Just a rough idea, there may be better heuristics we could apply to help deducing the "root cause" though I'm sure there's some risk of false positives. For example, `foo: 1` and `foo: 2` had failures but they arguably weren't a root cause. Maybe you just sort by total failures?

aoldershaw:
@vito the idea of failure rates for each value is pretty cool!

> Version numbers are also key-value JSON objects, and we render them like this:
> ...
> We could probably follow the same structure here. 🤔

Would the idea then be to display all of the steps at once, where each step has the table of vars inside the header? (e.g. if you wrapped a `task` in a 3x3 `across` step, would you see 9 task step headers at once?)

I'm feel like that would take up a large amount of vertical space when the `across` vars are complex objects, which could make it difficult to see the status of all the steps.

One worry with this design (and the current design, for that matter) is that when we get `var_source` prototypes, authors will either be encouraged to not emit rich objects because it results in a more cluttered UI, or they won't care and there'll be a lot of noise in the UI. I think the main problem is that any given `across` step may only use one or two fields of an object, so displaying any other fields is just visual noise. Maybe we could figure out what fields are actually used by an `across` step and only display those (probably isn't feasible, since I guess task `file:`s can reference local vars)

vito:
> Would the idea then be to display all of the steps at once, where each step has the table of vars inside the header? (e.g. if you wrapped a `task` in a 3x3 `across` step, would you see 9 task step headers at once?)

Hmm, I think showing 9 task steps at once might be a bit noisy/confusing - I wonder if we could only show all the combination headers, and click them to toggle the actual step output? The combination headers could indicate the sub-step status, so a user could skim the combinations and just click any failed ones to reveal the steps it ran. The build plan may be a single `task` or `set_pipeline` step, but it may also be a more complicated sequence of steps, and in that case it makes even more sense to hide them by default IMO.

My thinking here is that it's not super interesting to see the same successful step output repeated for every possible combination. I *would* however be interested in knowing what combinations ran, which combinations were OK, and for the ones that failed, why - and at that point I'd just click the combination.

Whaddaya think?

> I'm feel like that would take up a large amount of vertical space when the `across` vars are complex objects, which could make it difficult to see the status of all the steps.
> 
> One worry with this design (and the current design, for that matter) is that when we get `var_source` prototypes, authors will either be encouraged to not emit rich objects because it results in a more cluttered UI, or they won't care and there'll be a lot of noise in the UI. I think the main problem is that any given `across` step may only use one or two fields of an object, so displaying any other fields is just visual noise. Maybe we could figure out what fields are actually used by an `across` step and only display those (probably isn't feasible, since I guess task `file:`s can reference local vars)

I think we'll need to make it clear to var source prototype authors that the `list` operation should only include minimal information, just like resource versions - they're effectively "keys", not "values". Fetching values would be done through the `get_var` step, which is analogous to the `get` step for resources.

This should at least be somewhat obvious for credential manager var sources, since the `list` operation really shouldn't directly return credential values, but I could see authors making this mistake for things like PR var sources.

That being said, if there truly are multi-field objects to deal with, I think we should still be able to handle it somehow, since users could always do that manually with `values:` too. I just don't know that I would dramatically alter the design to handle this case vs. the more common cases of combining a couple single-field vars or going across a large set of values from a dynamic source.

aoldershaw:
Ah, I wasn't too clear about what I meant - I did mean 9 headers only (not the build logs themselves) - but even with that, it's still pretty hard to see the "big picture" (at least on my laptop with minimal screen real estate). Suppose you ran a `task` `across` three vars: `pr`, `go-version`, and `platform`, and suppose the `pr` var source gave back the same `commit`, `committed`, and `pr` (renamed to `number`) fields in its list operation. That could look something like:

Screen Shot 2020-07-16 at 1 24 52 PM

I can only fit 2-3 step headers on my (small-ish) screen at a time, and if the build matrix is large, this could mean a lot of scrolling. Possibly the `pr` var source just gives back the PR number, though, which would reduce the vertical space.

This could just be the nature of the problem, and I'm guessing an `across` step like this isn't the best way to do things anyway (you'd probably want `across` prs which sets a pipeline instance, with an `across` go-version and platform in the pipeline instance)

---

I'll experiment with this approach (table in the step header with a flat list).

aoldershaw:
Here's a bit of a wonky design...rather than a flat list of tabs, have a table of clickable "tabs". When the dimensionality is >2, nest tables inside of tables:

### 1D

Screen Shot 2020-07-19 at 11 43 07 PM

### 2D

Screen Shot 2020-07-19 at 11 37 36 PM

### 3D

Screen Shot 2020-07-19 at 11 35 51 PM

---

Styling is really ugly as I was just hacking together some HTML manually - I really don't love the var name header along the rows (*maybe* vertical text would look better, but I'm guessing that's a bad move for readability) - but I think it makes it easy to see the data "at a glance". @vito's example from https://github.com/concourse/concourse/pull/5887#issuecomment-659511864 could look something like:

Screen Shot 2020-07-19 at 11 52 25 PM

Which makes it pretty easy to spot the bad combination.

What do you think @vito @jamieklassen @matthewpereira? Personally I don't love it, but it's kind of nice being able to see all of the step statuses in a compact way form. I'm also guessing that the majority of cases will have dimensionality of 1 or 2, so it feels nice to provide a proper matrix of values in those cases.

vito:
Haha, that's a cool demo but yeah I think it gets a bit too confusing for >2 dimensions.

I think I'd find it a bit difficult to skim the 2-dimentional case; my eye has to travel from each '❌' to two different places on the screen (the X and Y headers) to piece together the actual combination. Definitely a nitpick, but could get annoying day to day.

Concourse has thus far trained me to look for a list of ✔ or ❌ signs (or red header borders), and I think a regular old table layout works well for that even with N dimensions:

|||||
| - | - | - | - |
| `foo: 1` | `bar: a` | `baz: alpha` | ✔ |
| `foo: 1` | `bar: a` | `baz: beta` | ✔ |
| `foo: 1` | `bar: b` | `baz: alpha` | ❌ |
| `foo: 1` | `bar: b` | `baz: beta` | ✔ |
| `foo: 2` | `bar: a` | `baz: alpha` | ✔ |
| `foo: 2` | `bar: a` | `baz: beta` | ✔ |
| `foo: 2` | `bar: b` | `baz: alpha` | ❌ |
| `foo: 2` | `bar: b` | `baz: beta` | ✔ |

I've left the header empty in this case and put key-value pairs in each cell instead; this way I only have to read the failed rows.

Bonus: we could use failure rates to intensify background colors of the values in failed rows - e.g. giving `foo: 1` and `foo: 2` a 50% (1/2 failures) red background, and `bar: b` and `baz: alpha` a 100% red background (2/2 failures):

image

re: common-ness, I suspect 1-dimension will be the most common, 2 next-most common, but I wouldn't be surprised if we see a decent amount of 3 or more dimensions. All it takes is someone with a 2-dimension matrix who finds out they need to support some other set of dependency versions. 🤔

Might be a good idea to look for real-world examples of build matrix configs in other CI systems across GitHub.

matthewpereira:
Would it be helpful to have something like this, with a table in the header, so that users can get feedback about the whole across step at a glance without opening it up:

Build Page - Across Step - table in header

Or should we move that info to the tasks themselves, something like this, and open the Across Step 'group' by default?

Build Page - Across Step - table in task step

(we could adjust the styling to make it more table-like if necessary in either case)

matthewpereira:
We could also combine those two ideas to achieve a sort of collapsed table built up across each task step 'row':

Build Page - Across Step - table across task steps

aoldershaw:
@matthewpereira one difficulty with the second design - what if the step we're iterating `across` is a `get` or `put` (which admittedly is a bit odd)? Where would we put the version in the step header?

And for either design, how would we go about displaying e.g. a `do` step? Right now, a `do` step doesn't have its own step header - it's just a stack of the step headers it contains. I guess we could make `do` steps have a step header when they're nested in an `across` step 🤷

aoldershaw:
I guess one question is: if we have a table of permutations + the status for each one, is there value in also seeing the all of the substeps at once? Or would it be better to be able to select a particular permutation from the table and see just that one substep?

vito:
> I guess one question is: if we have a table of permutations + the status for each one, is there value in also seeing all of the substeps at once? Or would it be better to be able to select a particular permutation from the table and see just that one substep?

IMO, the latter. If we're already summarizing the results I don't think we need to show all the output at once.

However there may be a case where you want to view a couple of them, i.e. to compare a failed combination with a successful one. Might be getting a bit too involved, but some sort of toggle rather than a straight switch could be interesting to explore.

One way to do that would be to take @matthewpereira's example from https://github.com/concourse/concourse/pull/5887#issuecomment-663584590 and split the big `across` header into multiple individually toggle-able headers which reveal the step output for that combination.

matthewpereira:
We could either add rows to the header in the case that we need to show more information, or we could add columns - one column for the regular metadata we already show, one for the across-step metadata. I'm not sure yet which would work better or how we would account for small screen sizes where you don't have a lot of horizontal space. 😕 

I think that if you're debugging a problem, you might want to be able to compare back and forth between at least two permutations (eg, a failed one and a successful one), and having them both open would be helpful if the log output was noisy or difficult to read. If we had to click tab-style or accordion-style, you might end up having to switch back and forth a few times to compare things due to how much information the average person can keep in memory at a given time.

aoldershaw:
> One way to do that would be to take @matthewpereira's example from [#5887 (comment)](https://github.com/concourse/concourse/pull/5887#issuecomment-663584590) and split the big `across` header into multiple individually toggle-able headers which reveal the step output for that combination.

So kinda like https://github.com/concourse/concourse/pull/5887#issuecomment-663586419 but the task header would be nested under the combination header?

vito:
> So kinda like [#5887 (comment)](https://github.com/concourse/concourse/pull/5887#issuecomment-663586419) but the task header would be nested under the combination header?

Yep, and without putting the combination in the task headers themselves, since as you pointed out that won't work for more complicated sub-steps.

> I think that if you're debugging a problem, you might want to be able to compare back and forth between at least two permutations (eg, a failed one and a successful one), and having them both open would be helpful if the log output was noisy or difficult to read. If we had to click tab-style or accordion-style, you might end up having to switch back and forth a few times to compare things due to how much information the average person can keep in memory at a given time.

Yeah exactly; this is what's nudging me towards just reusing the click-to-toggle headers metaphor for this, so they can click-to-toggle the combinations that they want to compare.

matthewpereira:
> > One way to do that would be to take @matthewpereira's example from [#5887 (comment)](https://github.com/concourse/concourse/pull/5887#issuecomment-663584590) and split the big `across` header into multiple individually toggle-able headers which reveal the step output for that combination.
> 
> So kinda like [#5887 (comment)](https://github.com/concourse/concourse/pull/5887#issuecomment-663586419) but the task header would be nested under the combination header?

When I read that description and look at my screenshot I think they're the same thing! This makes me think maybe I'm not communicating things well enough in the screenshot. Maybe a clickable prototype would help: when you're talking about individually toggle-able headers to reveal step output for a given combination, are you thinking about this?

HaYh2E6i2S

aoldershaw:
I understood it as there being one more click before seeing the logs for the `task` - the `task` header would be nested under one more header that just has the vars and the overall status, so the `task` header itself wouldn't have vars in it. This is to accommodate for e.g. sequences of substeps that aren't represented as a single step header.

Alternatively, we could put the vars inside the substep directly, and could wrap more complicated substeps like `do` in a step header like I suggested here: https://github.com/concourse/concourse/pull/5887#issuecomment-663594152. So, optimize for the simple case (where there's just a single step), and require two clicks for more complex substeps

aoldershaw:
@vito @matthewpereira implemented the new design, updated the screenshots in the PR description if you want to take a look and let me know what you think

@jamieklassen re: your concern about max in flight (https://github.com/concourse/concourse/pull/5887#pullrequestreview-449953186) - I wonder if you still feel it's important to provide a message saying a substep is waiting for max in flight now that they're displayed as step headers. It now looks and feels a lot like in parallel steps which doesn't display a reason for why build steps are pending

matthewpereira:
This works for me. Thanks for all the attention and thought you put into this, @aoldershaw !

vito:
@aoldershaw This is pretty rad. Here's a fun test pipeline for anyone that wants to kick the tires:

```yaml
---
jobs:
- name: test-go
  plan:
  - across:
    - var: go
      values: [1.8, 1.14, rc]
      max_in_flight: 2
    - var: command
      values: [version, help, env]
      max_in_flight: 3
    task: run-command
    config:
      platform: linux
      image_resource:
        type: registry-image
        source:
          repository: golang
          tag: ((.:go))
      run:
        path: go
        args: [((.:command))]
```

Now that I'm using it though, it does feel like a bunch of clicks just to get to the output. One to expand the `across`, another to expand the combination, and another to expand the inner step. Makes me think you had the right idea from the start of just putting the info in the nested step's header.

The problem I had with that was `do` and `in_parallel`, which don't have their own step header, but now I wonder if we should give them one instead. Doing so would help resolve various UI quirks that result from `do` and `in_parallel`/`aggregate` appearing "inlined" into places that make the UI confusing/misleading (e.g. https://github.com/concourse/concourse/issues/4089).

One question this raises is whether giving them a header implies that they introduce another click-to-toggle level. 🤔 It might be annoying to have to expand `do`/`in_parallel` from now on. Maybe they could have a different header style that suggests that they're a 'wrapper(?)' step, and those types of steps are auto-expanded? I'm starting to feel like we need a full taxonomy for our step types, haha: core, wrapper(?), modifier.

Anyway, just thinking out loud and wondering what y'all think. One alternative to all this could be to just eliminate the outer click-to-toggle on the `across` step - then it'd just be two, one for the combination and then another for the inner step.

matthewpereira:
Lingering thought: are we rendering the key/value pairs as inline text, or as a table?

To ask a different way, would the beginning of a third variable tacked onto the end of this image:

screenshot

...look like:

```
go: 1.13  platform: linux  something_else: foo
go: 1.13  platform: darwin  something_else: foo
go: 1.13  platform: windows  something_else: foo
```

...or would we line up the start of the `something_else` variable so that it looks like this:

```
go: 1.13  platform: linux    something_else: foo
go: 1.13  platform: darwin   something_else: foo
go: 1.13  platform: windows  something_else: foo
```

aoldershaw:
@matthewpereira yeah, I initially wanted to line it up in some sort of table but couldn't really figure out how best to do that. Elm makes performing calculations based on the size of DOM elements kind of a pain

@vito we could get rid of the `across` header. I'm not sure where we'd display `across` step warnings (currently just the experimental warning, but based on [the current RFC](https://github.com/vito/rfcs/blob/spatial-resources/029-across-step/proposal.md#var-scoping-and-shadowing), we'll also want to do so for variable shadowing) - would we repeat the warnings for each substep?

I'm also happy to explore wrapper step headers. We *could* just have wrapper headers (for `do`/`in_parallel`/`aggregate`/nested `across`) when it's displayed as an `across` substep, but it may be confusing why other wrapper steps don't have headers.

There's not much value in having a header for the top-level implicit `do`. We could special case the top-level `do` and use an expanded-by-default wrapper header for all other wrapper steps, though. Any thoughts on how the header style could differ?

aoldershaw:
@vito one option to reduce the number of clicks to output could be to auto-expand the step header under a combination header when it's just a single step. So, the combination headers are still collapsed, and e.g. `task` or `set_pipeline` sub-steps still have a sub-header (hidden below the combination header), but clicking on the combination header expands both. When the sub-step is a `do` or `in_parallel`, we wouldn't auto-expand the sub-steps (which would be effectively the same as wrapping `do` and `in_parallel` in a "wrapper" header, at least within `across` steps).

aoldershaw:
@jamieklassen sorry, haven't had much time to work on this recently. Just pushed a change that adds the feature flag. I also added less restrictive validation to allow an `across` step variable to shadow a local variable (with a warning being logged during validate and at runtime) as described in [the RFC](https://github.com/vito/rfcs/blob/spatial-resources/029-across-step/proposal.md#var-scoping-and-shadowing). In doing so, I realized that the way I was trying to do variable scoping wasn't correct - I created a copy of the local vars in the `engine/builder`, but those local vars only get populated at runtime (meaning a `load_var` from outside the `across` step wouldn't have been accessible inside of it).

So, I implemented proper variable scoping, and along the way refactored the `CredVarsTracker` and renamed it to `BuildVariables` (since it has evolved to more than just tracking credentials for secret redaction). Apologies for adding even more complexity to the PR - if you'd like, I could try to break this PR up into smaller, more consumable chunks.

PR #5894, opened 2020-07-18
feat(load_var): add format trim and make it default

evanchaoli:

Some resource types as well as tasks generate output files with tailing "\n", for example https://github.com/concourse/docker-image-resource#in-fetch-the-image-from-the-registry all output files have tailing "\n".

When `load_var` reads file as `raw` format, tailing "\n" may make following steps hard to use the var. For example:

```yaml
- load_var: image_repo
  file: image/repository

- load_var: image_tag
  file: image/tag

- some-step:
   params:
      uri: ((.:image_repo)):((.:image_tag))
```

The `uri` expects an image uri, like `some-image:some-tag`, but as `load_var` carries tailing newlines, `uri` will actually be `some-image:\nsome-tag\n". 

Add a params `trim_raw` to `load_var` step, if true, then trim heading and trailing spaces of content reading from file.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/exec/load_var_step.go
+++ b/atc/exec/load_var_step.go
@@ -175,6 +175,8 @@ func (step *LoadVarStep) fetchVars(
 		if err != nil {
 			return nil, InvalidLocalVarFile{file, "yaml", err}
 		}
+	case "trim":
+		value = strings.TrimSpace(string(fileContent))
 	case "raw":
 		value = string(fileContent)
 	default:
@@ -197,12 +199,12 @@ func (step *LoadVarStep) fileFormat(file string) (string, error) {
 		return format, nil
 	}
 
-	return "raw", nil
+	return "trim", nil
 }
 
 func (step *LoadVarStep) isValidFormat(format string) bool {
 	switch format {
-	case "raw", "yml", "yaml", "json":
+	case "raw", "trim", "yml", "yaml", "json":
 		return true
 	}
 	return false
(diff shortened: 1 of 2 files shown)


evanchaoli:
Yeah, I believe users will love `load_var`. It does not only avoid `foo_file`, but it makes something used to be hard or even impossible how easy to archive. For example:

1. `semver` resource's parameter `bump` doesn't support `bump_file`, thus current pipelines can only do static version bump. But with `load_var`, a pipeline can dynamic determine bump type in a step prior to `semver`, like

```yaml
- put: semver-version
  params:
     bump: ((.:bump_type))
```

2. It can add prefix or suffix to a parameter, like:

```yaml
- put: something
  params:
     release: rel-((.:semver_version))
```

Actually, I have been enjoying `load_var` when I build some new pipelines recently.

Back to the topic, my initial thought was to make trim default, which requires minimum code change, and even not require doc update. But I was worrying that some pipeline may want to keep raw content, for example, though tailing space can be trimmed, but user may need to keep heading spaces, who knows. 

I think your idea of `format: trim/strip` is good. Maybe we can make `format: trim` default, and still support `format: raw`, so that users would not have to remember `trim` or `strip`; if they want to read raw content, they only need to remember `raw`.

How about that?

vito:
@evanchaoli Yeah, that sounds great! 👍

evanchaoli:
@vito Now the code change is very simple.

PR #5896, opened 2020-07-20
Instanced pipelines - Add instance vars as part of the pipeline identifier

### LINKED ISSUE #5808: Pipeline instances

mouellet:

> All pipelines are pipeline instances whether they have `instance_vars` or not...

### ATC
* Adds a `jsonb` column in the `pipelines` table with unique partial index.
* Replaced all `pipelineName` method parameter with `atc.PipelineRef`.
* Adds an `instance_vars` attribute to the `set_pipeline` step.

### API
* Adds an optional `instance_vars` query param as html encoded raw json to all endpoints based on `/api/v1/teams/:team_name/pipelines/:pipeline_name`.

### Fly commands
* Adds `--instance-var` flag to `set-pipeline` command (e.g.: `fly set-pipeline -p branch --instance-var branch=feature/foo`)
* Use json flatten in dot notation to identify pipeline instances in `flaghelpers.*Flag`
e.g.:
```
fly ... -p pipeline-name/version:6.0.0,os:windows
fly ... -j pipeline-name/version:6.0.0,os:windows/job-name
fly ... -r pipeline-name/version:6.0.0,os:windows/resource-name
```
Also support complex json and numeric values as strings or numbers:
```
{
  "foo": {
    "bar": {
      "baz": 1,
      "qux": "2"
    }
  },
  "bar": [1, "2"]
}
```
`fly ... -p 'pipeline-name/foo.bar.baz:1,foo.bar.qux:"2",bar.0:1,bar.1:"2"'`

## Notes to reviewer:

Pipeline Groups & UI changes are out of scope of this PR. I feature flagged the use of instance vars in the step and the api so we can merge asap as it is quite large.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


mouellet:
Hey @vito, I've been playing around with fly commands for instanced pipeline and I think I've found a way to simplify the fly UX.

Instead of adding `--instance-var` with a colon (`:`) for partial filtering to _almost_ all fly commands, I could modify the `flaghelpers.*Flag` to have an optional coma separated list of key:value pairs as suffix.

e.g.: 
```
fly get-pipeline -p pipeline-group/version:6.0.0,os:windows
fly builds -j pipeline-group/job-name/version:6.0.0,os:windows
fly check-resource -r pipeline-group/resource-name/version:6.0.0,os:windows
```

Then :
* The `flaghelpers.PipelineFlag`: could be like `PIPELINE_GROUP/PIPELINE_INSTANCE_VARS`
* The `flaghelpers.JobFlag` could be like `PIPELINE_GROUP/JOB/PIPELINE_INSTANCE_VARS`
* The `flaghelpers.ResourceFlag` could be like `PIPELINE_GROUP/RESOURCE/PIPELINE_INSTANCE_VARS`

This also simplifies group operations (like pausing all instances of a group). By just splitting the flag's value on "/", if there's an instance vars suffix we can easily distinguish between instanced and group operations.

What do you think?

vito:
@mouellet Good idea! Seems like a good use of the flag system, since in these cases the intent of the flag is to identify a single pipeline, and specifying the vars is necessary in order to do that.

One suggestion: the order of `PIPELINE/JOB/PIPELINE_VARS` feels unintuitive - could we do `PIPELINE/PIPELINE_VARS/JOB` instead?

aoldershaw:
Sorry for the lack of feedback on this @mouellet. It looks really good so far! Was playing around with setting some pipelines, and noticed a couple things:

1. You can create both a group of instanced pipelines and an individual pipeline with the same name:

```sh
$ fly -t dev sp -n -p prs -c pipeline.yml -i number=5896
... (succeeds)
$ fly -t dev sp -n -p prs -c pipeline.yml -i number=5897
... (succeeds)
$ fly -t dev sp -n -p prs -c pipeline.yml
... (succeeds)
```

I suspect we want to disallow the third `fly sp` (i.e. a team + pipeline name either refers to a single pipeline or to a group of pipeline instances). Is that true @vito?

2. Not sure what practical implications this has, but saving a pipeline without `instance_vars` doesn't result in `p.instance_vars` being `NULL` in Postgres - instead, it's the `jsonb` value of `null`. At the very least, I think this `UNIQUE INDEX` won't be used at all:

https://github.com/concourse/concourse/blob/a1371a06fc9105a93fd6f30910802c6f9685aeeb/atc/db/migration/migrations/1594400645_add_instancevars_to_pipelines.up.sql#L7-L9

vito:
> I suspect we want to disallow the third `fly sp` (i.e. a team + pipeline name either refers to a single pipeline or to a group of pipeline instances). Is that true @vito?

Yeah that's a good point, the RFC doesn't really make this clear. 🤔 I think we should disallow it; seems pretty sneaky.

mouellet:
> I suspect we want to disallow the third `fly sp` (i.e. a team + pipeline name either refers to a single pipeline or to a group of pipeline instances). Is that true @vito?

I had the same question, but for now I see all pipelines as pipeline instances whether they have instance vars or not... The Pipeline Groups concept hasn't been introduced yet and I think it's somewhat related to that. BTW, we should have a discussion about that soon!

> 1. Not sure what practical implications this has, but saving a pipeline without `instance_vars` doesn't result in `p.instance_vars` being `NULL` in Postgres - instead, it's the `jsonb` value of `null`. At the very least, I think this `UNIQUE INDEX` won't be used at all:
> 
Expect this to be working in the next commits (still WIP...)

aoldershaw:
> I had the same question, but for now I see all pipelines as pipeline instances whether they have instance vars or not

Hmm, that's interesting! All of the UI designs that @matthewpereira has mocked up are under the assumption that all pipeline instances sharing a name would have the same set of instance variables, but I'm realizing now that isn't necessarily the case. Pipeline instances sharing a name may not have the same pipeline configuration, either. Granted, the use case of the `across` step + `set_pipeline` with instance vars will result in uniform pipeline instances - but this isn't the only way to create pipeline instances, clearly. I wonder if this changes anything from a design perspective

mouellet:
> I wonder if this changes anything from a design perspective

Hey @aoldershaw , right now I'm focusing on replacing the `pipelineName` param for (almost) all method and api calls with a new `atc.PipelineRef` parameter object. This way we'll be able to query, visualize and take actions on individual pipeline instances (e.g. watch the build of a pipeline instance...).

It does allow for everything that you said, and it alone doesn't resolve the rfc... But I think it'll make it flexible enough for which ever design we pick fo the Pipeline Groups as well as pipeline group operations.

Maybe the `fly set-pipeline` command should remain as-is and a new fly command is needed for setting pipeline groups or templates..?

aoldershaw:
Hey @mouellet, sorry, I was mostly just thinking out loud as it's something I hadn't really thought of (and wasn't sure if our designer was aware of it or not).

I think the approach you're taking now makes sense!

mouellet:
> One suggestion: the order of `PIPELINE/JOB/PIPELINE_VARS` feels unintuitive - could we do `PIPELINE/PIPELINE_VARS/JOB` instead?

Hey @vito, I implemented this and it works great! Definitely more intuitive. 

There is one thing however, unmarshalling the flag is a little tricky...

Both pipeline name and job/resource name _should_ be valid identifiers, but not the instance vars with its colon. And dealing with the cases where someone forgot to add the job/resource name at the end (e.g. `pipeline-name/branch:master`)...

So I added an error when the job/resource name is not a valid identifier even if we only emit warnings for now. Is that ok or should we wait for the valid identifiers to be enforced?

vito:
> So I added an error when the job/resource name is not a valid identifier even if we only emit warnings for now. Is that ok or should we wait for the valid identifiers to be enforced?

Hmm, good question. I think we might have to leave it as a warning for now, since we don't want it to break for someone who has an existing job with invalid characters and was able to use the `foo/bar` notation up until this point. 🤔

fly/integration/builds_test.go
```diff
@@ -862,12 +863,12 @@ var _ = Describe("Fly CLI", func() {
 		Context("when passing the pipeline argument", func() {
 			BeforeEach(func() {
 				cmdArgs = append(cmdArgs, "-p")
-				cmdArgs = append(cmdArgs, "some-pipeline")
+				cmdArgs = append(cmdArgs, "some-pipeline/branch:master")
 
 				expectedURL = "/api/v1/teams/main/pipelines/some-pipeline/builds"
-				queryParams = "limit=50"
+				queryParams = []string{"instance_vars=%7B%22branch%22%3A%22master%22%7D", "limit=50"}
```
aoldershaw, line 868: Not a big deal, but I'm wondering if we could encode the `instance_vars` differently in the query parameters, since URL encoding raw JSON isn't all that easy to read

What if we were to encode it as we do in the `PipelineFlag`...for instance: `?instance_vars=branch:master`

We could still URL encode the var names + values if need be. e.g. we could have `instance_vars=foo:bar%3Abaz` to represent `{"foo": "bar:baz"}`

This might be totally unnecessary, what do you think?

mouellet: I tried that, but the query is automatically encode in `go-concourse/concourse/internal` and I'm not sure toying with it is a good idea!

But I'm glad you raised this point because I wanted to start a discussion about using the pipeline's Id instead of the alternate key... Or at least allowing both in the api.

I've spent _some_ time in Elm and learned that the nature of the instance vars (`Dict String String`) can't be used in another `Dict`'s key, so the `teamName/PipelineName/InstanceVars` combination is a dead-end... But the pipeline's Id is always available and could be used instead!

In other words, I think we should allow both:
* `/api/v1/teams/:team_name/pipelines/:pipeline_name?instance_vars=url_encode_json`
* `/api/v1/teams/:team_name/pipelines/:pipeline_id`

Or for backward compatibility `/api/v1/teams/:team_name/pipelines/:pipeline_name_or_id?instance_vars=url_encode_json`, where the `pipeline_name_or_id` is just a check on the parameter's type or some other simple mechanism.

Any thoughts?

aoldershaw: *If* we want to support querying pipelines in the API by their id, it might make more sense for the endpoint to be `/api/v1/pipelines/:pipeline_id`, since the ID is unique across teams. But I'm trying to understand if that's necessary - I think the dead-end you're running into can be avoided by converting the `Dict String String` into a datatype that is a `comparable` (like a `List (String, String)` via `Dict.toList`, possibly also sorting the keys if necessary).

Also, side note - I was under the impression that instance vars could be any type (not just `string`), and could even have nested objects. e.g.

```yaml
set_pipeline: something
file: some-file.yml
instance_vars:
  foo: {bar: baz}
  qux: [1.2.3, 4.5.6]
```

@vito is that true?

Haven't looked through all of the code - wondering if this is something you've considered

vito: @aoldershaw Hmm... that's true, though now that you mention it I don't know if we've even considered the potential complexity from that in other places like the UI. I suppose the simplest thing may be to just JSON encode the non-string values everywhere we show them in the UI.

I like the idea of adding a `/api/v1/pipelines/:pipeline_id` endpoint. That will be more future-proof, especially if big dramatic things happen like [teams being replaced by projects](https://github.com/vito/rfcs/blob/5679ee86b4e810a8ae8e00dcb21aca2ad745103e/032-projects/proposal.md#can-projects-replace-teams), and potentially allowing projects to be nested within each other. So any change that gets us away from these hierarchical paths feels like a step in the right direction.

aoldershaw: The approach I took for the [`across` step](https://github.com/concourse/concourse/pull/5887) was to go 1 level deep into objects - so, `foo: {bar: 1, baz: 2}` would be rendered as two "vars" in the combination headers: `foo.bar: 1`, `foo.baz: 2`. Similarly for lists: `foo: [1, 2]` -> `foo[0]: 1`, `foo[1]: 2`. Anything deeper just gets encoded as raw JSON: `foo: {bar: {baz: 1}}` -> `foo.bar: {"baz": 1}`.

I figured that it wouldn't be uncommon to have vars 1 level deep, so we should optimize for that case. You can see an example of this in the screenshots in the PR description (`upgrade_path.from`, `upgrade_path.to`)

mouellet: Yeah, nested objects in instance vars make sense.. How would it looks like in fly commands? I think any kind of raw JSON would be unusable... Maybe combining all headers? eg.: `fly get-pipeline -p pipeline-name/foo.bar.baz:1,foo.bar.qux:2,bar[0]:1,bar[1]:2`?

aoldershaw: Raw JSON would be a bit annoying with all the quoting, but perhaps we could use a dumbed-down form of YAML (basically just JSON without the quotes):

`fly get-pipeline -p pipeline-name/foo:{bar:{baz:1,qux:2}},bar:[1,2]`

To parse it, you could probably even just take the whole string `foo:{bar:{baz:1,qux:2}},bar:[1,2]`, wrap it in `{...}`, and call `yaml.Unmarshal`.

vito: @aoldershaw That'll get weird with shells since `{}` is part of glob expansion. You'd have to quote the whole argument.

I wonder if we should enforce that instance vars are just regular identifiers. There's hidden complexity even in simple scalar values - if someone uses numbers, for example, we'd need to actually handle '2' as a number everywhere in the CLI. But what if someone specified them as a string like `"2"`? Would they need to quote it in the CLI? Will that not confuse the shell? e.g. `echo foo"bar"` just prints `foobar`.

Forcing them to be identifiers is a bit more limiting, but honestly it might be for the best - they're being used to identify the pipeline, and other vars can still be specified with arbitrary values.

vito: Though that kind of throws a wrench in some of my sketches, e.g. `instance_vars: {pr_number: 2}`.

Hmm...

aoldershaw: Could be valid identifiers or numbers? But then you can't use semver since that's a string.

Maybe we just loosen the restriction that valid identifiers can't start with a number, at least for instance vars?

vito: Yeah, that's tempting. My only hesitation is that I think instanced pipelines will be largely driven through Concourse itself with the `set_pipeline` step, and as a result the `fly` CLI interaction should be less than usual. I'm trying to be conscious not to let the CLI dictate other parts of the design *too* much, since it's not really the primary usage pattern; users typically shouldn't be manually archiving, destroying, exposing, or hiding individual instances. (I can see the argument for pausing and unpausing, even though that can be done in the UI.)

The most common command would probably be `fly intercept`, but I think it'd be better if that just took a container handle copied from the UI anyway. There's also `fly intercept --url` already.

I guess we could also accept a pipeline ID which can be copied from the UI somehow, or discovered through `fly pipelines`. I'm not sure users would want to bother typing it all out in the first place.

Don't want to just toss in a bunch of uncertainty and then bail on the convo, so: at this point I'm leaning towards keeping arbitrary values and just coming up with some sort of syntax for specifying them, which is what y'all were already doing. My one suggestion would be to not worry too much about arbitrary complex values, and just support accepting a pipeline ID in those cases. We'll see how users use `fly` with instances - it might just not be a big deal.

mouellet: Just to be clear... Instance vars are vars like all others, right? 

We have a jsonb column in db, and I can confirm that using more complex json is feasible and works fine (with a minor update pending...) for `set_pipeline` step.

Now about the fly commands, I think flattening/unflattening the json into dotted notation (eg.:  `pipeline-name/foo.bar.baz:1,foo.bar.qux:2,bar.0:1,bar.1:2`) works best, and it's as-is for simple instance vars (eg.:  `pipeline-name/branch:issue/5432`). 

If we want something simpler, then I'd suggest using the pipeline's id.

vito: @mouellet That's right. The dot notation seems like a good idea, since it avoids curly braces. My only concern then would be whether those '2' values are parsed as strings or numbers, since it'll have to match the pipeline's var values. Maybe we should just parse the values as YAML? If someone had a string that contained a number, then they could do this to disambiguate:

```
fly ... -p 'pipeline-name/foo.bar.baz:"1"'
```

Single quotes being significant here so the shell doesn't eat the double quotes.

mouellet:
@vito, should I also update the logging, tracing and metrics w/ the pipeline instance vars?

We could use the whole `atc.PipelineRef` with `lager.Data` (or anything that takes and `interface{}` or json) and it would look something like this:
```
{...,"data":{"pipeline":{"name":"p1","instance_vars":{"branch":"master"}},...}}
```

Maybe could use the dot notations when it's only a string, like with `tracing.Attrs`...

What do you think?

vito:
@mouellet Hmm I wonder if it would be better to just log IDs. Logging pipeline names implies we need to also log e.g. team name/ID, project name/ID (later), and other things to disambiguate since names are non-unique at various levels.

It adds an extra step if you're e.g. trying to search for logs of pipeline XYZ, but on the other hand it's precise, less verbose, and won't involve false positives or additional filtering. 🤔

What do you think?

Side note: maybe we should start a pattern of adding a method like `LogData() lager.Data` to objects so there's a single place to change this and to make sure everything's using consistent data field names? (e.g. `pipeline_id` instead of `pipeline`) Then we can do e.g. `log.WithData(pipeline.LogData())` everywhere pipeline-related.

mouellet:
@vito, maybe this is out of scope for the current PR? It's already quite large and feature flagged...

I suggest we open an issue for that, replacing all _names_ with _IDs_. In logging, at the API level, etc.

vito:
@mouellet 👍 yeah sounds wise, that would probably be a bigger change

mouellet:
Hey @vito, @aoldershaw, I pushed some changes on the UI side. It's **not** inline with #5921, but at least we don't get 404 on pipeline instance pages! And I'm getting better with Elm!!

aoldershaw (commented):
Hey @mouellet, really sorry about the lack of engagement on this! To be honest, I completely forgot about it - my bad on that

I'm picking up the UI track now in #5921, so taking a look at the status of things. You've done a massive amount of work, so huge thanks!

I tested it out briefly and noticed something - migrating up causes problems with existing pipelines. Trying to hit the `/teams/:team_name/pipelines/:pipeline_name` endpoint for a pipeline that existed before migrating up gives a 404. I haven't looked into the code to figure out why, but I did notice that a pipeline created after migrating up has `instance_vars = 'null'` (as in, the JSON `null`), whereas pipelines created before have `instance_vars = NULL` (as in, Postgres `NULL`) - so, I'm guessing it has something to do with that.

It also made me think about your comment from a while back: https://github.com/concourse/concourse/pull/5896#discussion_r468586006

*What if* all the pipeline endpoints were under `/api/v1/pipelines/:id`? As @vito suggested here https://github.com/concourse/concourse/pull/5896#discussion_r468667929, it's a good idea for future-proofing to remove the hierarchy anyhow, and it means we don't need to identify pipelines in two distinct ways simultaneously (as was done in 1d2f65146bc9b77c10f533656bbd2b4769fd2c9b). We *could* keep the existing `/api/v1/teams/:team_name/pipelines/:pipeline_name` around for backwards compatibility, but we would not need to support pipelines with instance vars via query parameters. If we choose to do so, supporting both should be fairly straightforward - we just need another implementation of `pipelineserver/scoped_handler_factory.go`. Likewise, we could leave all of the `go-concourse.Team` methods that interact with a pipeline alone (i.e. bring them back to the state on `master`), mark them as deprecated, and add a `go-concourse.Pipelines` for interacting with pipelines using their ID.

This would mean that `fly` would first need to determine the pipeline ID. We could modify the `/api/v1/pipelines` endpoint to take in query parameters (`team`, `pipeline`, `instance_vars`) to search for specific pipeline(s). Each `fly` command could first query the pipeline to get the ID, then perform subsequent requests.

What do you think @mouellet and @vito?

aoldershaw:
Also, noticed an issue related to #6041 and as a result of #5898 - since dots and colons are now valid in var names, but we use the dot notation in `fly` to query based on instance vars, there's no way to access such instances in `fly`. e.g.

```
$ fly -t dev sp -p group -c pipeline.yml -i test.val=abcd
...
$ fly -t dev gp -p group/test.val:abcd
error: pipeline not found
```

Perhaps we could make something like `fly -t dev gp -p 'group/"test.val":abcd'` work - requires `'...'` surrounding the entire thing so the shell doesn't try to expand the `"`s, but better than nothing.

mouellet:
> Hey @mouellet, really sorry about the lack of engagement on this! To be honest, I completely forgot about it - my bad on that
> 
> I'm picking up the UI track now in #5921, so taking a look at the status of things. You've done a massive amount of work, so huge thanks!
> 
> I tested it out briefly and noticed something - migrating up causes problems with existing pipelines. Trying to hit the `/teams/:team_name/pipelines/:pipeline_name` endpoint for a pipeline that existed before migrating up gives a 404. I haven't looked into the code to figure out why, but I did notice that a pipeline created after migrating up has `instance_vars = 'null'` (as in, the JSON `null`), whereas pipelines created before have `instance_vars = NULL` (as in, Postgres `NULL`) - so, I'm guessing it has something to do with that.

Ah! You talked about that before and I didn't quite get it! I naively expected Go's `nil` to turn into Postgres `NULL`... But as we must have Postgres `NULL` for the partial index to work, I'll look into using `sql.Null*` for this...

> It also made me think about your comment from a while back: [#5896 (comment)](https://github.com/concourse/concourse/pull/5896#discussion_r468586006)
> 
> _What if_ all the pipeline endpoints were under `/api/v1/pipelines/:id`? As @vito suggested here [#5896 (comment)](https://github.com/concourse/concourse/pull/5896#discussion_r468667929), it's a good idea for future-proofing to remove the hierarchy anyhow, and it means we don't need to identify pipelines in two distinct ways simultaneously (as was done in [1d2f651](https://github.com/concourse/concourse/commit/1d2f65146bc9b77c10f533656bbd2b4769fd2c9b)). We _could_ keep the existing `/api/v1/teams/:team_name/pipelines/:pipeline_name` around for backwards compatibility, but we would not need to support pipelines with instance vars via query parameters. If we choose to do so, supporting both should be fairly straightforward - we just need another implementation of `pipelineserver/scoped_handler_factory.go`. Likewise, we could leave all of the `go-concourse.Team` methods that interact with a pipeline alone (i.e. bring them back to the state on `master`), mark them as deprecated, and add a `go-concourse.Pipelines` for interacting with pipelines using their ID.
> 
> This would mean that `fly` would first need to determine the pipeline ID. We could modify the `/api/v1/pipelines` endpoint to take in query parameters (`team`, `pipeline`, `instance_vars`) to search for specific pipeline(s). Each `fly` command could first query the pipeline to get the ID, then perform subsequent requests.

I'm all on board with this migration, but had more a 2 phase plan in mind with this PR being the first: to support existing endpoints via query parameters. Main reason is that the PR is already huge! But it also include all endpoints prefixed with `/api/v1/teams/:team_name/pipelines/:pipeline_name/`!

I would prefer to make a second PR if that's alright with you @aoldershaw , @vito?

mouellet:
> Also, noticed an issue related to #6041 and as a result of #5898 - since dots and colons are now valid in var names, but we use the dot notation in `fly` to query based on instance vars, there's no way to access such instances in `fly`. e.g.
> 
> ```
> $ fly -t dev sp -p group -c pipeline.yml -i test.val=abcd
> ...
> $ fly -t dev gp -p group/test.val:abcd
> error: pipeline not found
> ```
> 
> Perhaps we could make something like `fly -t dev gp -p 'group/"test.val":abcd'` work - requires `'...'` surrounding the entire thing so the shell doesn't try to expand the `"`s, but better than nothing.

The issue here as more to do with parsing the flag for nested/complex instance vars (I've been having problems with that!!). The reason why the `get-pipeline` doesn't work is because the instance var is parsed and saved in db as `{"test.val": "abcd"}` instead of `{"test": {"val": "abcd"}}`. We save in a jsonb column after all.

I understand we have to support it somehow, my first idea was to use the dot notation (same as the `get-pipeline` but for `set-pipeline`). However, it gets unusable pretty quickly especially with arrays and nested objects...
 
What do you think about using the 3 flags options that the vars uses?
1. for simple string/string key/value pairs (with a validation to exclude dot and colons)
1. for simple string/yaml key/value pairs (to support `int` and `bool`)
1. for loading nested/complex instance vars from a yaml file

aoldershaw:
> I'm all on board with this migration, but had more a 2 phase plan in mind with this PR being the first: to support existing endpoints via query parameters. Main reason is that the PR is already huge! But it also include all endpoints prefixed with /api/v1/teams/:team_name/pipelines/:pipeline_name/!

Makes sense to me. I was initially suggesting that we might not need to support instance vars on the `/teams/:team/pipelines/:pipeline` endpoints at all and instead make it so the only way to access pipelines with instance vars was through the new `/pipelines/:id` endpoints - but perhaps it makes the most sense to support both. I also don't mind adding the `/pipelines/:id` endpoints, as I'll need that for my UI work.

Speaking of which...would you be willing to revert the Elm changes you made? 😬

I apologize, I'm sure it was a fair bit of work, but if we are going to just use the `/pipelines/:id` endpoints in the UI, a lot of those changes will need to be reverted anyway. I've already done a refactor that replaces the `teamName` + `pipelineName` pipeline identifiers with just the pipeline ID - you can see the changes here: https://github.com/concourse/concourse/compare/issue/5921

mouellet:
> Speaking of which...would you be willing to revert the Elm changes you made? 😬

Of course!

You should have a look at this (https://github.com/concourse/concourse/pull/5896/commits/9158d346bdcba81d4bc8df29237ca71b8e5249b5#diff-0adb228ece585aa5df6c9151feef93d2R488) to convert the instance vars to dot notation if you want to use it instead of the format specified in #5821 (e.i. `<instance group name>-<instance var 1>[-<instance var 2>][...]`). The RFC wasn't clear on that...

mouellet:
Hey @vito, @aoldershaw,

I'd like to have your input on my revisited `--instance-var` flag for the `fly set-pipeline`.

Using the dot-notation for assignment can be tricky for complex objects, but it just works! And we could always add a flag to load the instance vars from a yaml file...

Just to make sure we all share the same understanding (and preparing for the documentation...), here's my definition of instance vars:

> A pipeline's `instance_vars` is a single JSON object written in key/value pairs where keys must be strings, and values must be a valid JSON data type (string, number, object, array, boolean).
>
> The object's key/value pairs will be injected in the pipeline as variables but should not contain secrets.
>
> With `fly` commands, the `instance_vars` is represented as a comma separated list of flattened JSON object properties using dot-notation, where `=` is used for assignment and `:` for partial filtering.
>
> e.g.: simple values
>```
>{
>  "branch": "master"
>}
>```
>```
>fly set-pipeline -c pipeline.yml -p pipeline --instance-var branch=master
>fly get-pipeline -p pipeline/branch:master
>```
>
> e.g.: nested values
>```
>{
>  "foo": {
>    "bar": "baz"
>  },
>  "quz": ["foo", 1, false]
>}
>```
>```
>fly set-pipeline -c pipeline.yml -p pipeline -i foo.bar=baz -i quz.0=foo,quz.1=1,quz.2=false
>fly get-pipeline -p pipeline/foo.bar:baz,quz.0:foo,quz.1:1,quz.2:false
>```

I've used dotted array index `quz.0` instead of the brace kind `quz[0]` only because it was easier the expand back into JSON...

fly/commands/internal/flaghelpers/job_flag.go
```diff
@@ -58,14 +76,53 @@ func (flag *JobFlag) Complete(match string) []flags.Completion {
...
+			kv := strings.SplitN(strings.TrimSpace(instanceVar), ":", 2)
+			if len(kv) == 2 {
+				instanceVars[kv[0]] = kv[1]
+			}
+		}
+
+		pipelineRef := atc.PipelineRef{Name: vs[0], InstanceVars: instanceVars}
+		for _, pipeline := range pipelines {
+			if strings.HasPrefix(pipeline.Ref().String(), pipelineRef.String()) {
+				comps = append(comps, flags.Completion{Item: pipeline.Ref().String() + "/"})
```
aoldershaw, line 85: Pretty awesome that you added flag completion for instance vars, too!

fly/commands/internal/flaghelpers/job_flag.go
```diff
@@ -12,19 +12,37 @@ import (
...
+	if jobNameIdx == -1 {
 		return errors.New("argument format should be <pipeline>/<job>")
 	}
 
-	job.PipelineName = vs[0]
-	job.JobName = vs[1]
+	flag.JobName = value[jobNameIdx+1:]
+	if warnings := atc.ValidateIdentifier(flag.JobName, "job"); warnings != nil {
+		return errors.New("argument format should be <pipeline>/<key:value>/<job>")
+	}
```
aoldershaw: Hmm I suspect we don't want to be strict about the job/resource identifier yet

aoldershaw (commented):
Only gone through about half of the commits, but have a few comments so far

atc/exec/set_pipeline_step.go
```diff
@@ -295,6 +299,14 @@ func (s setPipelineSource) FetchPipelineConfig() (atc.Config, error) {
 		staticVars = append(staticVars, sv)
 	}
 
+	if len(s.step.plan.InstanceVars) > 0 {
+		iv := vars.StaticVariables{}
+		for k, v := range s.step.plan.InstanceVars {
+			iv[k] = v
+		}
+		staticVars = append(staticVars, iv)
+	}
```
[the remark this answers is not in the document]
vito: Precedence is kind of funny to reason about here. Does this mean that if you you set an instance var, but then set the same var through `vars`/`var_files`, the template will be evaluated with values from `vars`/`var_files` however the pipeline will be identified by the `instance_vars`?

Seems kind of sneaky. I wonder if it should be a validation warning/error. Can we think of any situation where this might occur intentionally or accidentally?

fly/commands/internal/flaghelpers/job_flag.go
aoldershaw: Oh haha, think I just haven't read far enough into the commits, as it looks like this already may have changed

atc/exec/set_pipeline_step.go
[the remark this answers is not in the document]
aoldershaw: I suppose this *could* happen intentionally if you wanted to identify the pipeline using some memorable human-understandable value while the actual underlying value is more complex. For instance, a `ref` instance var could be some friendly name, while the `ref` var could be a commit/tag SHA. Or, an `image` instance var could be an image tag, while the `image` var could be the digest.

Those aren't such compelling examples, though, since you could just as easily use a different variable name to achieve the same thing.

> Seems kind of sneaky. I wonder if it should be a validation warning/error.

Would this just be config validation, or would we want to perform actual runtime validation?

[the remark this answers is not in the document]
vito: I think it would have to be at runtime because of `var_files`. This kind of check is unprecedented (heh) though and the implementation doesn't seem like it'd be very graceful. I guess you'd have to loop over `vars` and all `var_files` to look for conflicts with `instance_vars`?

I don't feel strongly about adding the validation unless we think it'll be a "gotcha." It doesn't seem particularly high-risk, and the current behavior seems like a reasonable interpretation.

aoldershaw:
Hey @mouellet, we really appreciate your patience 🙏  sorry it's taken so long, and sorry you've had to rebase so frequently to stay up-to-date with changes since first opening the PR!

We're planning to ship 6.6 this week, and after that I think we're pretty much ready to get this merged. I'm going to keep reviewing until then, but hopefully it won't take too much longer! Looks really good so far, and things seem to work as expected

aoldershaw:
I was thinking about how to support var names that have `.` in them. We could try to make use of the same logic we use for parsing var interpolation in pipelines:

https://github.com/concourse/concourse/blob/96fd3ac90267d3e255f38e735a0b9c2d2d56a34d/vars/template.go#L138-L159

This gives a `vars.VariableReference`, which has a `Path` and a list of `Fields`. That way, if (for whatever reason) you *need* to have an instance var with a `.` in it, you could do something like `-i '"some.var".field=123'`

aoldershaw:
Heads up, we're temporarily blocked on releasing 6.6, so it probably won't end up being this week - but soon! Will keep you posted

fly/commands/internal/flaghelpers/pipeline_flag.go
```diff
@@ -11,23 +12,71 @@ import (
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func unmarshalDotNotation(value string) (atc.DotNotation, error) {
+	dn := atc.DotNotation{}
+	for _, v := range strings.Split(value, ",") {
```
aoldershaw, line 64: It would also be nice to not *require* dot notation for querying, but that's not so obvious to me how to do that, so I'm fine with this

aoldershaw (approved):
Thanks again for the awesome contribution and sorry for the delay! We really appreciate your patience!

I'm going to merge this now - any updates can be handled in another PR (no pressure to address any of the comments I left - we're happy to do this)

PR #5897, opened 2020-07-20
Generate opaque access tokens

### LINKED ISSUE #5851: DEX auth token header length issues on v6.3.0
## Summary

Just upgraded our dev environment to v6.3.0 from v5.7.0, and 2 users are reporting problems authenticating via the web interface, due to Nginx (sitting in front of Concourse) throwing:

```
400 Bad Request

Request Header Or Cookie Too Large
```

And related Nginx logs:

```
2020/07/06 15:32:09 [info] 26472#26472: *897589 client sent too long header line: "Cookie: _gcl_au=1.1.1376805454.1588963645; _ga=GA1.2.778376655.1588963645; DEV_concourse-dt=ffffffgg79cc1e0445525d0f4f58155e4e5a4a4t378b; DEC=%7B%22SEGMENT%22%3A%22RES%22%2C%22JU%22%3A%22TRUNCATED%22%2C%22STOREBANNEREXPDATE%22%3A%2208%2F07%2F2020%2010%3A57%3A27%22%7D; TLTSID=74645591441348700406164839997852; _A_source=source=localhost:4200|campaign=(not%20set)|medium=referral|content=/; _A_time=64.1588963644472.1593545391369; _uetvid=fa33028-d34-4c55-50f0-268c13ccdf0c; skymarshal_auth0="bearer eyJhbGciOiJSUzI<TRUNCATED>..." while reading client request headers, client: 1.2.3.4, server: , request: "GET /api/v1/resources HTTP/1.1", host: "ourcluster.domain.com", referrer: "https://ourcluster.domain.com/"
```

Most users aren't reporting problems -- this used to be a large problem pre-5.7.0 where the amount of teams we were in, caused very large headers, but this isn't the case (5 teams vs I can login with over 140 teams). I believe this may be related to the AD groups being in the headers. The users in mention have over 700 AD groups (715 and 721).

What I am not sure of, if this is simply due to other headers being applied, or if the headers from Concourse are truly too large. I also am not sure how I can read the auth header.

Is there any way that I can open the header to see what's taking up the most space? (not familiar with jwt format is being used, or if salts are being used, from db, etc). I don't know if AD groups are embedded in the header... Is there any way I can ensure only the ad groups that map to a team are included (i.e. they'd have to relogin to get new access, but that's better than not being able to login at all).

We have temporarily made an Nginx adjustment (from an otherwise 90% stock Nginx config), to double the header length. I don't know if this is advised or not, but it seems to be an intermediary fix that works. The adjustment:

```nginx
large_client_header_buffers 8 16k; # defaults to 4 and 8k
```

## Steps to reproduce

1. Deploy v6.3.0
2. Setup team with ldap access via direct user (non-ad groups)
3. Have user return member of 700 AD groups (could be less, not very sure), via AD.
4. Attempt to login via the UI.

## Expected results

Login successful. Possibly stronger splitting of auth headers (I see only one skymarshal_auth[0-9], maybe it should be splitting it even further?).

## Triaging info

* Concourse version: v6.3.0
* Browser (if applicable): Chrome (latest), IE (latest), Edge (almost latest).
* Did this used to work? Yes.

### LINKED ISSUE #5912: Listing all teams takes a long time due to redundant auth checking
## Summary

When the web UI asks for `/api/v1/teams`, and you're on Concourse 6.1.0+, and there are a few hundred teams, and you're not an admin, the response can take a long time to come back. In our case I'm seeing it taking about 30s for a normal user who is in a handful of teams. Admin users can retrieve the full list very speedily (milliseconds).

Whichever web node happens to be receiving the bulk of load-balanced traffic can be taking about 40% CPU (in our case) serving these requests.

## Steps to reproduce

* We are on 6.4.0 but I think the same behavior will be seen on 6.1.0.
* Have lots of teams - we have about 600, and I expect the performance degradation to be linear in the number of teams, so it could still be subjectively "bad" with a smaller number
* Be a logged-in non-admin
* From the web UI or directly, look at response times for `/api/v1/teams`.

## Expected results

I don't know how critical this data is for the dashboard - and since it doesn't change frequently, things should typically be basically fine as soon as the browser has the results of the first call. In general, though, it would be nice to have snappy responses. If nothing else, that helps with our operational tooling where we'd like to have alerts for "HTTP responses are super slow" without distractions from this particular endpoint.

## Actual results

```sh
$ time fly -t ci curl /api/v1/teams
[...list of the 8 teams for this user comes back...]

real	0m29.448s
user	0m0.111s
sys	0m0.048s
```

In this flame graph, the `ListTeams` bar is taking 37% of CPU. The four `rolesForTeam` bars together account for the bulk of that, about 31%. Within each `rolesForTeam`, it's about half-and-half between the four calls that ultimately ask for federated claims, and some map iteration costs. The tiny sliver on the far right is DB access, which does not appear to be a big contributor to the time taken. (As evidenced by the fact that _admin_ users can get the full team list very quickly - in their case, we make the exact same DB call, but then only have to pass the first `IsAdmin()` test against each team.)

## Additional context

Inspection of 🔥 flame graphs 🔥 and the code shows that the time taken is driven by a few factors:
1. `access.IsAdmin` is called three times for each team
2. each of those, plus `access.hasRequiredRole`, invokes `access.rolesForTeam`
3. the resulting (number of teams) * 4 calls to `access.rolesForTeam` each involve several further calls to `access.federatedClaim`, and some iteration over maps

Much of this work seems to be redundant in a formal sense, and while it's not too expensive to do once, it is adding up over the larger number of teams. For example, `teamserver/list.go` has, for each team,

```go
		if acc.IsAdmin() || acc.IsAuthorized(team.Name()) {
```

but then `IsAuthorized()` also ends up asking for `IsAdmin()` first of all anyway. Then, within `IsAuthorized()`, the call to `TeamNames()` also checks for admin status before querying `hasRequiredRole()`. It seems like this could be simplified to remove the redundancy. (The same redundancy appears to be in place for some other endpoints, but this one is the expensive one because of the repeated cost of inspecting the _team_ structure.)

Within `rolesForTeam()`, we first ask for four of the federated claims (groups, connector ID, user ID and user name). The leaves of these seem taken up by invocations of `runtime.fastrand` and `runtime.duffzero`; I don't know enough about go internals to understand why some map lookups and typecasts should involve these. Other runtime functions used in `rolesForTeam()` are `runtime.mapaccess1_faststr`, `runtime.mapiterinit`, and `runtime.mapiternext`, presumably for the looping over `auth`, `auth["users"]` and `auth["groups"]`. There may be ways to make this more efficient, perhaps again by reducing the redundancy.

The caching logic and general changes in https://github.com/concourse/concourse/pull/5897 might alter the overall performance characteristics, but I haven't tested.

## Triaging info

* Concourse version: 6.4.0 but probably true on 6.1.0 as well, since that is when the token format change happened
* Browser (if applicable): not a browser issue since this can be seen with `curl` just as well
* Did this used to work? On 6.0.0 with the same number of teams, our metrics show the same call taking ~10ms. We upgraded yesterday and saw this behavior change immediately, although all other performance is very good. (In particular, builds/checks/etc. are going through speedily even though this is happening.)

aoldershaw:

closes #5851 .
closes #5912 .

Previously, we were using the ID tokens from dex as access tokens. While this allows for stateless validation of requests, we ran into several issues where the ID token was massive for users that were in many groups, which caused several problems in practice (e.g. load balancers rejecting requests with such large headers).

The approach this PR takes is:

* Intercept the `/sky/issuer/token` response from dex before it goes back to the client
  * Replace the access token from dex with an access token with 160 bits of entropy + the expiration timestamp
    * Rationale for 160 bit token: https://neilmadden.blog/2018/08/30/moving-away-from-uuids/
  * Store the mapping of the access token to claims in a new database table `access_tokens`
    * Sidenote: this table will make it super easy to implement concourse/rfcs#63
* On each API call, fetch the claims for the corresponding access token from the database
  * Cache this on the ATC so we don't need to always hit the DB (up to 1MiB of claims is stored in an LRU cache)
* GC component that clears out expired access tokens from the DB
  * Side note: a single user can have multiple valid access tokens if they log in multiple times, and indeed they'll have multiple rows in the database. This is consistent with dex, and makes sense - if you `fly login`, that shouldn't revoke your web UI token.

It also changes the user tracking behaviour - rather than updating a user's last login on every authenticated request, only update it on login. This affords us some simplification in the API auth flow (in particular, we no longer need to batch user update requests) - and #5915 indicates this approach has some serious performance implications.

## Release Note

* There were several issues that users encountered (particularly after v6.1.0) as a result of long access tokens. Concourse now generates much shorter access tokens rather than using the raw user data.
* Users' last activity is now tracked on login rather than on every request. Updating the last activity on every request caused database problems at scale. Note: last activity is only relevant to `fly active-users`

## Notes to reviewer:

TODO list:

- [x] caching for access token lookup
- [x] GC to clear expired access tokens from DB
- [ ] ~This may be scope creep, but maybe provide a mechanism for revoking tokens? Previously, operators could rotate the dex signing keys, and all existing sessions would be invalidated. With this change, we no longer validate against the JWT signature (since the token isn't a JWT), so rotating the signing keys will have no effect. With this change, we could introduce a `fly revoke-token (--all) (--user=connector:username) (--sub=sub)` without *too* much effort - I wonder if this is something people will want/need~
  - we can add this as an enhancement later if people want it

I left instructions for repro-ing the issue (on master) using an HAProxy load balancer here: https://github.com/concourse/concourse/issues/5851#issuecomment-661285419. Note that you may need to mess around with the `tune.bufsize` param so that it just barely allows unauthenticated requests to succeed, but fails when there's a long JWT as the access token. I found 2800 bytes was the sweet spot for me, but it may vary.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


aoldershaw:
Ran some benchmarks using `wrk` on the `/api/v1/user` endpoint to determine whether caching would be necessary.

On master, this is these are the results:

```
$ wrk -d5s -H 'Authorization: bearer eyJhbGciOiJSUzI1NiIsImtpZCI6IjgxNTY3ODRiZGVjODY1NWNjMmYwODljMjJjMzhlYjEyNmRiNDcwMGYifQ.eyJpc3MiOiJodHRwOi8vbG9jYWxob3N0OjgwODAvc2t5L2lzc3V
lciIsInN1YiI6IkNnUjBaWE4wRWdWc2IyTmhiQSIsImF1ZCI6ImZseSIsImV4cCI6MTU5NTQ2NTc1OSwiaWF0IjoxNTk1Mzc5MzU5LCJhdF9oYXNoIjoiaHNSeUdVaGZTdVk0NnNOS1FQQjFYQSIsImVtYWlsIjoidGVzdCIsImVtYWlsX3Zl
cmlmaWVkIjp0cnVlLCJmZWRlcmF0ZWRfY2xhaW1zIjp7ImNvbm5lY3Rvcl9pZCI6ImxvY2FsIiwidXNlcl9pZCI6InRlc3QiLCJ1c2VyX25hbWUiOiJ0ZXN0In19.WENjhqcTdNxl0lU_99RON3DY1inS3hCu_dPBYzIMptquhW_d0YGCKwuZ
Xh2xhFx2070BvumFMBF8K7ucx5xnIooV2QDnCDNk1Oz-0eZFFg-TbI4r5BPf2Pk6wW49oKd_OwWItbSpDuf5T0Q3K7eXKJgEgAQzmxKvKYbhJS3GtYA' http://docker.for.mac.localhost:8080/api/v1/user
Running 5s test @ http://docker.for.mac.localhost:8080/api/v1/user
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    17.94ms    7.83ms  76.28ms   73.87%
    Req/Sec   282.35     49.34   434.00     70.00%
  2829 requests in 5.06s, 1.23MB read
  Socket errors: connect 0, read 0, write 0, timeout 10
Requests/sec:    558.90
Transfer/sec:    248.34KB
```

On this branch:

```
$ wrk -d5s -H 'Authorization: bearer x' http://docker.for.mac.localhost:8080/api/v1/user
Running 5s test @ http://docker.for.mac.localhost:8080/api/v1/user
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    29.87ms   21.40ms 163.88ms   83.78%
    Req/Sec   183.26     45.68   280.00     67.00%
  1834 requests in 5.04s, 814.91KB read
Requests/sec:    364.20
Transfer/sec:    161.83KB
```

So it's certainly slower

aoldershaw:
After adding caching, I'm seeing much better performance:

```
$ wrk -d5s -H 'Authorization: bearer mepTHqP0WPpFEbEuUtdOj2oM/dZn3hhfAAAAAA' http://docker.for.mac.localhost:8080/api/v1/user
Running 5s test @ http://docker.for.mac.localhost:8080/api/v1/user
  2 threads and 10 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    14.67ms    5.40ms  93.56ms   75.67%
    Req/Sec   341.00     42.46   450.00     72.00%
  3404 requests in 5.02s, 1.48MB read
Requests/sec:    677.89
Transfer/sec:    301.21KB
```

Not sure if the increased requests/sec is indicative of better performance, or if my CPU was just a bit busier when running the previous benchmarks - but either way, at least it doesn't seem to negatively impact API performance, so I'm happy

atc/db/migration/migrations/1594926103_add_access_token_and_claims.down.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+    DROP TABLE IF EXISTS access_tokens;
```
jamieklassen, line 2: I couldn't remember if this worked on Postgres 9 so I deployed and then ran `migrate --migrate-db-to-version=1591737494` and it worked fine.

I remember now that it's `DROP COLUMN IF EXISTS` that only exists in certain versions of Postgres.

atc/atccmd/command.go
```diff
@@ -1808,7 +1821,7 @@ func (cmd *RunCommand) constructAPIHandler(
 			logger,
 			accessFactory,
 			tokenVerifier,
-			cacher,
+			teamsCacher,
 			batcher,
```
jamieklassen, line 1825: These three components `tokenVerifier`, `teamsCacher` and `batcher` are all consumed by the `AccessorHandler` and they mostly just get orchestrated to pass data to the `accessFactory`. Furthermore, they all contain optimizations intended to reduce stress on the database -- the `tokenVerifier` now wraps a 1MB-LRU `claimsCacher`, the `teamsCacher` is of fixed size with a time-based retention policy, and the `batcher` stores user info in memory until a certain batch size is reached or a certain amount of time has passed. All of these optimization strategies are slightly different, but they all seem to relate to the core problem that "figuring out token stuff is expensive".

Now that we're not only verifying the signature of tokens but we're storing them in the DB, can we consolidate these components so use a more consistent optimization pattern?

My most-aesthetic idea was to add columns to the `users` table for:
* the current access token -- which includes its expiry time (a string? or something more indexable?)
* the user's team roles (a jsonb? if it's too big maybe move into a join table??)
* whether the user has the system claim -- this might be where the idea breaks down because a _token_ can have a system claim without actually being associated to a _user_.

It doesn't have to be this exact solution, but it still probably makes sense to cache the results of computing "is this token valid? if so, what are its team memberships and is it a system token?" (which is all the DB-bound portions of constructing an `Accessor` I think) all together, rather than having separate caches for tokens and teams.

aoldershaw: I think it still makes sense for there to be an `access_tokens` table, rather than adding the current access token to the `users` table, since a user can have multiple active access tokens (e.g. otherwise, a `fly login` could invalidate the token from the UI).

If I'm understanding "user's team roles" correctly, that sounds like it'd be going against (one of the) initial motivations for the new auth introduced in 6.1 - team config changes shouldn't require logging in again. I think it still makes sense to store the claims from dex in some raw form in the DB (currently, it's a `jsonb` in the `access_tokens` table), but we could cache team memberships and whether it's a system token rather than just caching the claims themselves.

Also, it's worth noting we can probably get rid of the batcher optimization now (as well as updating `users.last_login` on every request), and instead use the `iat` claim of the user's latest access token to determine this information. This is a slight change in behaviour, but it's more accurate to the name `last_login`, anyway

jamieklassen: Yup all good points. The cache doesn't have to be an exact duplicate of the db table.

chenbh:
Small note that this will break community resources that use the undocumented api: https://github.com/alphagov/terraform-provider-concourse/blob/5ff605de4afef50551bf97ed56ba0c9f4c2730aa/pkg/client/client.go#L34-L37

I guess since it's undocumented we can get away with a minor release?

aoldershaw:
@chenbh huh, good find. Looks like they already need to play catch up with concourse versions (that `token.NewTokenSource` was added in https://github.com/alphagov/terraform-provider-concourse/commit/43515b09f91c295602c126b7052bb7bb70cb1a0f to support 6.1.0 changes). I'm inclined to say that it warrants a minor release for the reason you bring up, but we can probably mark it as breaking in the release notes. @vito do you agree?

vito:
@aoldershaw @chenbh Yeah sounds good. 👍

atc/atccmd/command.go
aoldershaw: I pushed a commit that removes the `batcher`, and updates the `last_login` when a user actually logs in (rather than on every request). Initially I suggested this could be done using the "`iat` claim on the user's latest access token" but realized this wouldn't work since we GC expired access tokens.

It makes sense to have the `accessFactory` consume the `tokenVerifier` and `teamsCacher` (instead of the `AccessorHandler`). I'm not exactly sure how best to go about consolidating those two caches into a precomputed team membership + metadata cache. It feels like doing so would just make both optimizations less effective - whenever we need to compute a new user's memberships, we need to fetch a list of all teams; and whenever we fetch a list of teams (or invalidate the list based on time or a DB notification), we'd have to invalidate the full cache.

So, unless you're opposed, I think I'll keep the two caches separate for now

skymarshal/token/access_token.go
```diff
@@ -0,0 +1,175 @@
...
+		}
+		newResp, err := json.Marshal(resp)
+		if err != nil {
+			logger.Error("marshal-new-response", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+		body = bytes.NewReader(newResp)
+	})
+}
```
agurney, line 114: I am not clear end-to-end on what happens if I happen to unilaterally POST to `/sky/issuer/token` with a body of my choosing - perhaps containing all sorts of claims. I haven't tested this PR but it looks like as written, there is no signature checking for the claims in the passed IDToken - we just go on to stuff the extracted data into the DB, under the newly chosen random string, and return that string. Later auth attempts using the token should succeed, because they look it up in the DB and retrieve the claims that were given. So I could use this to grant myself access to anything.

It is also likely that I haven't understood the entire dex interaction, and that there is some place where the authenticity of the dex payload is checked, or that there's a nuance I'm missing about how this handler is exposed.

Hoping that you can enlighten me :-)

aoldershaw: The `/sky/issuer/token` endpoint (and the OAuth2 token endpoint in general) supports several different flows, some of which require previous "hops" to get there (e.g. logging in via GitHub using the `authorization_code` grant type). For each of the possible flows, Dex validates that the flow was "successful" (whatever that means for the particular grant type - e.g. correct password for the `password` grant, valid auth code for the `authorization_code` grant, etc). None of the grant types, however, support providing your own claims - Dex tells you what your claims are.

My point being, Dex is still doing all the validation to determine whether the request to `/sky/issuer/token` is valid, and won't grant a token otherwise.

This new middleware delegates to Dex, parses and stores the claims without validation, and sends back a modified response. I didn't see any value in validating the claims against Dex's signing keys for the following reasons:

1. We aren't delegating to an unknown third-party - this is a piece of code that's bundled into the Concourse binary itself
2. We aren't hitting a network interface that could be tampered with - the Dex handlers and request/response handoff is all done in memory

That said, if you think there's still a vulnerability here, I'd be happy to hear you out. I'm far from a security expert, so I could be wrong on this.

agurney: Aha, I think I did not understand the "delegation to dex" part - that's the `httptest.NewRecorder` business where it has the inner `ServeHTTP`? So this ends up just tweaking the response from Dex, and Dex still does whatever it does to authenticate the total flow.

aoldershaw: Yep, exactly

atc/db/access_token.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func (c Claims) MarshalJSON() ([]byte, error) {
+	return json.Marshal(c.RawClaims)
+}
+
+func (c *Claims) UnmarshalJSON(data []byte) error {
+	type target Claims
+	var t target
+	if err := json.Unmarshal(data, &t); err != nil {
```
taylorsilva, line 38: This was neat to dig into. It wasn't obvious to me why Claims had to be redefined until I removed`target` and got a stack trace showing the recursive call to `Claims.UnmarshalJSON` that then occurs!

aoldershaw: Yup! Learned that trick from @vito's steps refactor:

https://github.com/concourse/concourse/blob/a822574e7d1e3e30a54d8cac416bc829c0103d02/atc/steps.go#L414-L415

chrisfarms:
I believe this may have introduced a race during concurrent access/auth :thinking:  https://github.com/concourse/concourse/pull/6019

PR #5898, opened 2020-07-21
Allow dot and colon in variable path

### LINKED ISSUE #4249: No way to reference credentials with `.` in their name
## Bug Report

When using the atc vault integration for vars in `((..))` format keys with dots are ignored.

A secret at ``concourse/main/some.secret:key`` can not be read through variables. I think the regex that looks for key names should use the last dot to split the path and the key.

it would be nice if concourse realize that` ((some.secret.key))` is a secret at `concourse/main/some.secret:key` or to have a way to escape the key `.`, something like `((some\.secret.key))`

Regex seems to explicitly ignore paths with `.`s.
https://github.com/concourse/concourse/blob/ca0fd4cb1ed0b619fd1e083b5b3c827ffb669e63/vars/template.go#L66

## Steps to Reproduce

Try to read a secret where the path contains the `.` character.

## Version Info

* Concourse version: 5.3.0
* Deployment type (BOSH/Docker/binary): bosh
* Infrastructure/IaaS: vsphere

xtremerui:

closes #4249  .

Allow dot `.` and `:` in variable path if quoted with `"`. Parse variable name in `((var_name))` to `VariableReference` so underneath implementation of `Variable` interface doesn't need to.

`((foo:"a_path_with.dot".bar.zoo))` ->
```
VariableReference{
  Name: "foo:\"a_path_with.dot\".bar.zoo",
  Source: "foo",
  Path: "a_path_with.dot",
  Fields: ["bar","zoo"],
}
```

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/creds/cached_secrets.go
+++ b/atc/creds/cached_secrets.go
@@ -3,6 +3,7 @@ package creds
 import (
 	"time"
 
+	"github.com/concourse/concourse/vars"
 	"github.com/patrickmn/go-cache"
 )
 
@@ -36,16 +37,16 @@ func NewCachedSecrets(secrets Secrets, cacheConfig SecretCacheConfig) *CachedSec
 	}
 }
 
-func (cs *CachedSecrets) Get(secretPath string) (interface{}, *time.Time, bool, error) {
+func (cs *CachedSecrets) Get(ref vars.VariableReference) (interface{}, *time.Time, bool, error) {
 	// if there is a corresponding entry in the cache, return it
-	entry, found := cs.cache.Get(secretPath)
+	entry, found := cs.cache.Get(ref.Name)
 	if found {
 		result := entry.(CacheEntry)
 		return result.value, result.expiration, result.found, nil
 	}
 
 	// otherwise, let's make a request to the underlying secret manager
-	value, expiration, found, err := cs.secrets.Get(secretPath)
+	value, expiration, found, err := cs.secrets.Get(ref)
 
 	// we don't want to cache errors, let the errors be retried the next time around
 	if err != nil {
@@ -66,9 +67,9 @@ func (cs *CachedSecrets) Get(secretPath string) (interface{}, *time.Time, bool, 
 				duration = itemDuration
 			}
 		}
-		cs.cache.Set(secretPath, entry, duration)
+		cs.cache.Set(ref.Name, entry, duration)
 	} else {
-		cs.cache.Set(secretPath, entry, cs.cacheConfig.DurationNotFound)
+		cs.cache.Set(ref.Name, entry, cs.cacheConfig.DurationNotFound)
 	}
 
 	return value, expiration, found, nil
--- a/atc/creds/cached_secrets.go
+++ b/atc/creds/cached_secrets.go
@@ -3,7 +3,6 @@ package creds
 import (
 	"time"
 
-	"github.com/concourse/concourse/vars"
 	"github.com/patrickmn/go-cache"
 )
 
@@ -37,16 +36,16 @@ func NewCachedSecrets(secrets Secrets, cacheConfig SecretCacheConfig) *CachedSec
 	}
 }
 
-func (cs *CachedSecrets) Get(ref vars.VariableReference) (interface{}, *time.Time, bool, error) {
+func (cs *CachedSecrets) Get(secretPath string) (interface{}, *time.Time, bool, error) {
 	// if there is a corresponding entry in the cache, return it
-	entry, found := cs.cache.Get(ref.Name)
+	entry, found := cs.cache.Get(secretPath)
 	if found {
 		result := entry.(CacheEntry)
 		return result.value, result.expiration, result.found, nil
 	}
 
 	// otherwise, let's make a request to the underlying secret manager
-	value, expiration, found, err := cs.secrets.Get(ref)
+	value, expiration, found, err := cs.secrets.Get(secretPath)
 
 	// we don't want to cache errors, let the errors be retried the next time around
 	if err != nil {
@@ -67,9 +66,9 @@ func (cs *CachedSecrets) Get(ref vars.VariableReference) (interface{}, *time.Tim
 				duration = itemDuration
 			}
 		}
-		cs.cache.Set(ref.Name, entry, duration)
+		cs.cache.Set(secretPath, entry, duration)
 	} else {
-		cs.cache.Set(ref.Name, entry, cs.cacheConfig.DurationNotFound)
+		cs.cache.Set(secretPath, entry, cs.cacheConfig.DurationNotFound)
 	}
 
 	return value, expiration, found, nil
--- a/atc/creds/conjur/conjur.go
+++ b/atc/creds/conjur/conjur.go
@@ -5,6 +5,7 @@ import (
 
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc/creds"
+	"github.com/concourse/concourse/vars"
 )
 
 type IConjurClient interface {
@@ -37,8 +38,8 @@ func (c Conjur) NewSecretLookupPaths(teamName string, pipelineName string, allow
 	return lookupPaths
 }
 
-func (c Conjur) Get(secretPath string) (interface{}, *time.Time, bool, error) {
-	secretValue, err := c.client.RetrieveSecret(secretPath)
+func (c Conjur) Get(ref vars.VariableReference) (interface{}, *time.Time, bool, error) {
+	secretValue, err := c.client.RetrieveSecret(ref.Name)
 	if err != nil {
 		return nil, nil, false, nil
 	}
--- a/atc/creds/conjur/conjur.go
+++ b/atc/creds/conjur/conjur.go
@@ -5,7 +5,6 @@ import (
 
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc/creds"
-	"github.com/concourse/concourse/vars"
 )
 
 type IConjurClient interface {
@@ -38,8 +37,8 @@ func (c Conjur) NewSecretLookupPaths(teamName string, pipelineName string, allow
 	return lookupPaths
 }
 
-func (c Conjur) Get(ref vars.VariableReference) (interface{}, *time.Time, bool, error) {
-	secretValue, err := c.client.RetrieveSecret(ref.Path)
+func (c Conjur) Get(secretPath string) (interface{}, *time.Time, bool, error) {
+	secretValue, err := c.client.RetrieveSecret(secretPath)
 	if err != nil {
 		return nil, nil, false, nil
 	}
--- a/atc/creds/conjur/conjur.go
+++ b/atc/creds/conjur/conjur.go
@@ -39,7 +39,7 @@ func (c Conjur) NewSecretLookupPaths(teamName string, pipelineName string, allow
 }
 
 func (c Conjur) Get(ref vars.VariableReference) (interface{}, *time.Time, bool, error) {
-	secretValue, err := c.client.RetrieveSecret(ref.Name)
+	secretValue, err := c.client.RetrieveSecret(ref.Path)
 	if err != nil {
 		return nil, nil, false, nil
 	}
--- a/vars/template.go
+++ b/vars/template.go
@@ -1,7 +1,6 @@
 package vars
 
 import (
-	"errors"
 	"fmt"
 	"regexp"
 	"sort"
@@ -51,7 +50,7 @@ func (t Template) Evaluate(vars Variables, opts EvaluateOpts) ([]byte, error) {
 
 func (t Template) interpolateRoot(obj interface{}, tracker varsTracker) (interface{}, error) {
 	var err error
-	obj, err = interpolator{}.Interpolate(obj, varsLookup{tracker})
+	obj, err = interpolator{}.Interpolate(obj, tracker)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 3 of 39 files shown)


vars/template.go
```diff
@@ -130,50 +129,81 @@ func (i interpolator) extractVarNames(value string) []string {
...
 }
 
-type varsLookup struct {
-	varsTracker
-}
+func parseVarName(name string) VariableReference {
+	var pathPieces []string
 
-var ErrEmptyVar = errors.New("empty var")
+	varRef := VariableReference{Name: name}
```
chenbh, line 141: I'm beginning to think that the Name shouldn't include Source, that way it'll just be Path + Fields and won't break all of our existing cred managers

vars/template.go
```diff
@@ -130,50 +129,81 @@ func (i interpolator) extractVarNames(value string) []string {
...
+	varRef.Path = strings.ReplaceAll(pathPieces[0], "\"", "")
+	if len(pathPieces) >= 2 {
+		varRef.Fields = pathPieces[1:]
 	}
 
-	val, found, err := l.varsTracker.Get(splitName[0])
+	return varRef
+}
+
+// var ErrEmptyVar = errors.New("empty var")
```
chenbh, line 161:
```suggestion
```

vars/template.go
```diff
@@ -130,50 +129,81 @@ func (i interpolator) extractVarNames(value string) []string {
...
+		expectAllUsed:  expectAllUsed,
+		missing:        map[string]struct{}{},
+		visited:        map[string]struct{}{},
+		visitedAll:     map[string]struct{}{},
+	}
+}
+
+// Get value of a var. Name can be the following formats: 1) 'foo', where foo
+// is var name; 2) 'foo:bar', where foo is var source name, and bar is var name;
+// 3) '.:foo', where . means a local var, foo is var name.
```
chenbh, line 187: Comment needs updating, the main purpose now is to extract Fields from a `map[string]interface{}`

atc/creds/conjur/conjur.go
```diff
@@ -37,8 +38,8 @@ func (c Conjur) NewSecretLookupPaths(teamName string, pipelineName string, allow
 	return lookupPaths
 }
 
-func (c Conjur) Get(secretPath string) (interface{}, *time.Time, bool, error) {
-	secretValue, err := c.client.RetrieveSecret(secretPath)
+func (c Conjur) Get(ref vars.VariableReference) (interface{}, *time.Time, bool, error) {
+	secretValue, err := c.client.RetrieveSecret(ref.Name)
```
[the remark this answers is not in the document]
xtremerui: Feels like it should be `ref.Path`

vars/template.go
xtremerui: The problem is if we do `path` + `fields` we will need to concatenate them back with `.` and wrap them properly with `"`. I think the parsing of `(())` makes us to use ref.Path for creds manager.

And the logic in template will take care of looking up fields.

chenbh: We can probably use `parts[1]`
https://github.com/concourse/concourse/blob/46efd51a6657287ba07b78b5627fb880f68fc32e/vars/template.go#L144-L145

chenbh: > And the logic in template will take care of looking up fields.

True, I guess another question is how much processing should the cred manager do vs Concourse. They technically have all the information (e.g. Fields) required to fully resolve the secret, but it might be less code duplication if Concourse handles that part

vars/template.go
xtremerui: I think the end goal of this `Get` is still finding a string value? Extracting fields is the first part. It then use the fields to do recursive look up.

aoldershaw:
After a bunch of fiddling, I eventually got a local Conjur connected to Concourse. I ended up mostly following the [quick start](https://www.conjur.org/get-started/quick-start), but made the few modifications to the quick start repo. These are the changes I made

1. Don't use SSL (was having trouble getting self-signed certs to pair with concourse)

```diff
diff --git a/conf/default.conf b/conf/default.conf
index 8aa7e18..0832396 100755
--- a/conf/default.conf
+++ b/conf/default.conf
@@ -1,11 +1,8 @@
 server {
-    listen              443 ssl;
+    listen              80;
     server_name         proxy;
     access_log          /var/log/nginx/access.log;

-    ssl_certificate     /etc/nginx/tls/nginx.crt;
-    ssl_certificate_key /etc/nginx/tls/nginx.key;
-
     location / {
       proxy_pass http://conjur;
     }
```

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 1332b86..56ad583 100755
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -53,10 +53,9 @@ services:
     image: nginx:1.13.6-alpine
     container_name: nginx_proxy
     ports:
-      - "8443:443"
+      - "8081:80"
     volumes:
       - ./conf/:/etc/nginx/conf.d/:ro
-      - ./conf/tls/:/etc/nginx/tls/:ro
     depends_on:
     - conjur
     - openssl
```

2. Use the following policy, rather than the provided `BotApp.yml`:

```yaml
- !policy
  id: concourse
  body:
  - !variable "secret.value"
  - !variable "secret"
```

3. Use the following docker-compose override for running concourse:

```yaml
version: '3'

services:
  web:
    environment:
      CONCOURSE_CONJUR_APPLIANCE_URL: http://docker.for.mac.localhost:8081
      CONCOURSE_CONJUR_ACCOUNT: myConjurAccount
      CONCOURSE_CONJUR_AUTHN_LOGIN: admin
      CONCOURSE_CONJUR_AUTHN_API_KEY: 1xkf4xh2efa4ja7k3ng0rhdqg52dajps63ks08h511q08ks1d8a4pd
      CONCOURSE_CONJUR_SECRET_TEMPLATE: concourse/{{.Secret}}
```

where the API_KEY is the admin key (in the `admin_data` file from the quick start).

---

Anyway, I tested it a bit with the changes in this PR and without. Something about this change seems to not work with (at least) Conjur, even in the case where there are no dots or colons in the path. e.g. consider the following pipeline:

```yaml
jobs:
- name: job
  plan:
  - task: echo
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: echo
        args: [((secret))]
```

On master, this works correctly (i.e. echos the value of `secret` in Conjur), but with this change, I get the following error:

```
failed to interpolate task config: 404 Not Found. Variable 'concourse/main/echo/secret' not found in account 'myConjurAccount'.
```

(note: I had to modify to modify this line https://github.com/concourse/concourse/blob/a822574e7d1e3e30a54d8cac416bc829c0103d02/atc/creds/conjur/conjur.go#L43 to get the error message to show up, since currently we just ignore it and assume the var wasn't found)

xtremerui:
Thx for the detail of integration test.

> (note: I had to modify to modify this line
> 
> https://github.com/concourse/concourse/blob/a822574e7d1e3e30a54d8cac416bc829c0103d02/atc/creds/conjur/conjur.go#L43
> 
> to get the error message to show up, since currently we just ignore it and assume the var wasn't found)

This error is strange since it outputs the correct path, which matches the unit test. Concourse doesn't implement `RetrieveSecret` so there is no change in that part. I wonder what the argument of `RetrieveSecret` is in master when it success to fetch the secret. Will try this locally.

xtremerui:
@aoldershaw the way that creds manager looking up secret is 
 - generate secret looks up path. If no templating ENV is set, it will use the default path template `concourse/team_name/pipeline_name/secret, concourse/team_name/secert, concourse/secret` in Conjur's case.
- try fetching the secret by EVERY path here
https://github.com/concourse/concourse/blob/c5e2ad3808f7229337c661a05dd4ce3c768e6df7/atc/creds/secret_var_lookup.go#L22
notice how it relies on the return of `Get` https://github.com/concourse/concourse/blob/c5e2ad3808f7229337c661a05dd4ce3c768e6df7/atc/creds/secret_var_lookup.go#L28
. It always expect no error so it could go to next iteration. So in your case if you change the line in Conjur to return an error. It will failed in the first attempt and stop the loop, while the correct path `concourse/secret` is ignored. Not sure if that is your case or not.

In my local testing, I rebased master and run the test pretty much follow your instruction. I have these values set up in Conjur.
```bash
docker-compose exec client conjur variable value concourse/secret 
blah
 docker-compose exec client conjur variable value concourse/secret.value
foo
```
In my test both ((secret)) and (("secret.value")) revolved to `blah` and `foo`, respectively. And ((secret.value)) give back error like
`failed to interpolate task config: cannot access field 'value' of non-map value ('string') from var: secret.value`.

Additionally, I use the setup to test static variable by `fly -t local sp -p echo -c echo.yml -v secret.value=buzz`.

I also used this pipeline to test named variable
```
var_sources:
- name: dumb
  type: dummy
  config:
    vars:
      simple: hello!
      user.value:
        username: big
        password: sekrit

jobs:
- name: job
  plan:
  - task: echo
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: echo
        args: [((dumb:"user.value".password))]
```
All seems working. Could you try this rebased version again? Thx!

aoldershaw:
@xtremerui just tried it out again, and it works as expected. Will take a closer look through the code now

atc/creds/cached_secrets.go
```diff
@@ -36,16 +37,16 @@ func NewCachedSecrets(secrets Secrets, cacheConfig SecretCacheConfig) *CachedSec
 	}
 }
 
-func (cs *CachedSecrets) Get(secretPath string) (interface{}, *time.Time, bool, error) {
+func (cs *CachedSecrets) Get(ref vars.VariableReference) (interface{}, *time.Time, bool, error) {
```
aoldershaw, line 40: What's the reason that all of the credential managers take in a `VariableReference`, rather than just the `Path`? They don't seem to consume the `Source` or `Fields` from what I can tell

vars/template.go
```diff
@@ -130,81 +129,44 @@ func (i interpolator) extractVarNames(value string) []string {
 	var names []string
 
 	for _, match := range interpolationRegex.FindAllSubmatch([]byte(value), -1) {
-		names = append(names, strings.TrimPrefix(string(match[1]), "!"))
```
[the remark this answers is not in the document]
xtremerui: yeah the whole mechanism was borrowed from bosh variable parsing. And `!` was a syntax that used there but in concourse it is not used.

vars/template.go
```diff
@@ -214,18 +176,23 @@ func newVarsTracker(vars Variables, expectAllFound, expectAllUsed bool) varsTrac
...
+func (t varsTracker) Get(varName string) (interface{}, bool, error) {
+	t.visitedAll[varName] = struct{}{}
+
+	varRef := parseVarName(varName)
 
-	val, found, err := t.vars.Get(VariableDefinition{Name: name})
+	val, found, err := t.vars.Get(VariableDefinition{Ref: varRef})
 	if !found {
-		t.missing[name] = struct{}{}
+		t.missing[varRef.Path] = err
```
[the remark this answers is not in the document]
xtremerui: I have changed it to use varRef.Name that is unique and added a test for that.

vars/template.go
```diff
@@ -266,8 +243,8 @@ func (t varsTracker) ExtraError() error {
 	unusedNames := map[string]struct{}{}
 
 	for _, def := range allDefs {
-		if _, found := t.visitedAll[def.Name]; !found {
-			unusedNames[def.Name] = struct{}{}
+		if _, found := t.visitedAll[def.Ref.Path]; !found {
+			unusedNames[def.Ref.Path] = struct{}{}
```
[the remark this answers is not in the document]
xtremerui: Good catch! I have updated it to use the combination of source and path as identifier.

atc/creds/cached_secrets.go
xtremerui: Guess we could still pass `path` as the argument.

vars/template.go
```diff
@@ -184,13 +184,13 @@ func newVarsTracker(vars Variables, expectAllFound, expectAllUsed bool) varsTrac
...
-	t.visitedAll[varName] = struct{}{}
-
 	varRef := parseVarName(varName)
 
+	t.visitedAll[identifier(varRef)] = struct{}{}
+
 	val, found, err := t.vars.Get(VariableDefinition{Ref: varRef})
 	if !found || err != nil {
-		t.missing[varRef.Path] = struct{}{}
+		t.missing[varRef.Name] = struct{}{}
```
aoldershaw, line 193: Just checking - is the reason we use `varRef.Name` rather than `identifier(varRef)` so that our error message matches the var they tried to reference? e.g. if they reference `((vs:var1.path.to.field))`, but `var1` doesn't exist in `vs`, we emit an error like:

```
undefined vars: vs:var1.path.to.field
```

rather than

```
undefined vars: vs:var1
```

??

xtremerui: yes I think `name` is the raw variable string that has everything. We'd like to show the error as detailed as possible.

In the other case for the unused error, we rely on the returned list of variable definitions from underlining variable type, where field info is not included. So we have to use `source` + `path` as identifier
https://github.com/concourse/concourse/blob/d8818bdbc1840e64253d3de6ec3c0273324618f3/vars/static_vars.go#L12-L20

aoldershaw:
Looks pretty good but FYI some of the unit tests are broken - missed reverting some test code in dfcc42783cc9abb80a414973f9dce002c5a37efe I think

mkkeffeler:
what version of concourse will this roll into?

aoldershaw:
@mkkeffeler should be the next release (6.5.0), which will probably be in around 2-3 weeks

mkkeffeler:
was this test case tested? seems to be faiiling for us on 6.6 

```
params:
  CP_CREDENTIALS: (("me/Nonprod/N_A_GOG_GCP_SVC_PFOP_NL/CloudService-N_A_GOG_GCP_NL-me.me.com-test@me.com/password"))

```

xtremerui:
was it working before? Seems to me the last part `com-test@me.com` needs to be quoted?

mkkeffeler:
actually no. its a conjur credential. all of those dots are dots in the actual secret name, not a sub attribute or anything like that

xtremerui:
do you mean the var is not evaluated to a value by conjur when you say "failed"? or there is an error.

mkkeffeler:
the var is not evaluated to a value in conjur, conjur logs show it isn't being looked up. when we print "env" we see the variable that we typed in the pipeline, rather than the value. other variables work though.

xtremerui:
@mkkeffeler Thx for the info. I think the interpolation regex failed to match the var name in your particular case. This is a bug we will fix it.

mkkeffeler:
hmm not sure i follow. Can you explain how it does that? and is there a workaround?

xtremerui:
we have a regex to first check if the node in yml file is a var that needed to be interpolated or not. For example, one condition is it needs to be surrounded by `((..))`. There are some other conditions as well.

The example you gave by current regex is not considered a var at all so it is treated as normal text. By allowing `@` in the regex will fix the problem. If you are more interested you can check the PR I submit to fix this later.

mkkeffeler:
got it

we also will have spaces in some vars. not sure if thats already working, but worth calling out :)

PR #5908, opened 2020-07-22
db: behaviour: save disable_manual_trigger on job

### LINKED ISSUE #5906: disable_manual_trigger not working as expected
## Summary

I was trying to configure a job which is not able to be triggered manually but it is possible to do it via UI and CLI.

## Steps to reproduce

- set the example pipeline with attached configuration
- unpause and trigger pipeline in frontend or via CLI

## Expected results

I expect to not see the trigger button or to be able to trigger the job.

## Actual results

The button is enabled and I'm able to start the job.

## Triaging info
[example.txt](https://github.com/concourse/concourse/files/4959480/example.txt)

* Concourse version: 6.3.0 and 6.4.0
* Browser (if applicable): Chrome 84.0.4147.89 & FF 77.0.1

jamieklassen:

closes #5906.

Very small change to a query, and added a test -- this feature seems to have regressed in #4933 when we did some refactoring of the database. While there _was_ a wats acceptance test covering this, I still think a DB test is much more economical.

## Notes to reviewer:

If you use the pipeline in the linked issue you should see that the code on master allows you to trigger the job, whereas with this change you will get the expected 409 Conflict when you attempt to trigger the job.

## Release Note

Fixed the [`disable_manual_trigger:`](https://concourse-ci.org/jobs.html#schema.job.disable_manual_trigger) field on jobs -- since v6.0.0 it had no effect and jobs with this setting could actually still be manually triggered.

## Contributor Checklist

- [x] [Signed] all commits
- [x] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

### DIFF
--- a/atc/db/job_test.go
+++ b/atc/db/job_test.go
@@ -84,6 +84,10 @@ var _ = Describe("Job", func() {
 				{
 					Name: "job-2",
 				},
+				{
+					Name: "non-triggerable-job",
+					DisableManualTrigger: true,
+				},
 			},
 			Resources: atc.ResourceConfigs{
 				{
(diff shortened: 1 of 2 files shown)


jamieklassen:
Bear with me while I get a bit philosophical -- I think we want to encourage idempotent builds, but not require "pure" builds (whose execution is perfectly reproducible and independent of environmental factors). If this practice is followed, then disabling manual triggering is really about controlling the possible sets of inputs rather than preventing builds from occurring on-demand -- builds would only be created when _triggering_ inputs have new versions, but rerunning should still be allowed in case the build flakes and you don't want to create a new resource version just to rerun.

If you really want to prevent on-demand execution that sounds more like an authorization thing rather than job configuration. But I might also be making excuses for an ambiguous design... 🤷‍♀️ 

I vaguely remember asking the same question ("does `disable_manual_trigger` also disable rerunning?") when we were planning the rerunning feature and I think we came up with the same answer above but I can't find a discussion to link -- @vito does this ring a bell?

vito:
@jamieklassen That does ring a (very faint) bell. In any case I think your rationale makes sense. 👍

PR #5911, opened 2020-07-24
worker: prefix containerd specific flags with CONCOURSE_CONTAINERD

### LINKED ISSUE #5407: containerd: organize containerd flags prior to making containerd GA

muntac:

closes #5407.

`containerd` specific flags have been put under their own namespace and are now prefixed with `CONCOURSE_CONTAINERD` to prevent confusing them with Guardian flags and features.

Guardian flags will still be prefixed with `CONCOURSE_GARDEN`. However, this PR changes many internal references to the terms Garden, Guardian, server, and backend to reduce ambiguity for developers.

Corresponding update to Helm chart: https://github.com/concourse/concourse-chart/pull/131
Corresponding update to BOSH release: https://github.com/concourse/concourse-bosh-release/pull/112
Corresponding updates to CI:
https://github.com/concourse/ci/pull/356
https://github.com/concourse/ci/pull/357

Remaining TODO:
- [x] Prevent mixed usage of containerd and guardian flags
- [x] Update Helm chart
- [x] Update BOSH release
- [x] Update containerd topgun/k8s test
- [x] Update CI `ignored-in-binary` and `ignored-in-distribution` lists to be compatible with updated binary, helm chart, and bosh release 

## Notes to reviewer:

We recommend going over individual commits and commit descriptions while reviewing.
 
## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/worker/workercmd/guardian.go
+++ b/worker/workercmd/guardian.go
@@ -104,9 +104,6 @@ func (cmd *WorkerCommand) guardianRunner(logger lager.Logger) (ifrit.Runner, err
 	return grouper.NewParallel(os.Interrupt, members), nil
 }
 
-// For historical reasons this is "CONCOURSE_GARDEN_" instead of "CONCOURSE_GUARDIAN_"
-var guardianEnvPrefix = "CONCOURSE_GARDEN_"
-
 func detectGuardianFlags(logger lager.Logger) []string {
 	env := os.Environ()
 
--- a/worker/workercmd/guardian.go
+++ b/worker/workercmd/guardian.go
@@ -17,7 +17,8 @@ import (
 	"github.com/tedsuo/ifrit/grouper"
 )
 
-// This runs Guardian using the gdn binary
+// This prepares the Guardian runtime using the gdn binary.
+// The gdn binary exposes Guardian's functionality via a Garden server.
 func (cmd *WorkerCommand) guardianRunner(logger lager.Logger) (ifrit.Runner, error) {
 	depotDir := filepath.Join(cmd.WorkDir.Path(), "depot")
 
@@ -30,9 +31,7 @@ func (cmd *WorkerCommand) guardianRunner(logger lager.Logger) (ifrit.Runner, err
 
 	members := grouper.Members{}
 
-	gdnFlags := []string{}
-
-	gdnServerFlags := []string{
+	gdnFlags := []string{
 		"--bind-ip", cmd.BindIP.IP.String(),
 		"--bind-port", fmt.Sprintf("%d", cmd.BindPort),
 
@@ -46,9 +45,9 @@ func (cmd *WorkerCommand) guardianRunner(logger lager.Logger) (ifrit.Runner, err
 		"--no-image-plugin",
 	}
 
-	gdnServerFlags = append(gdnServerFlags, detectGardenFlags(logger)...)
+	gdnFlags = append(gdnFlags, detectGuardianFlags(logger)...)
 
-	if cmd.Garden.DNS.Enable {
+	if cmd.Guardian.DNS.Enable {
 		dnsProxyRunner, err := cmd.dnsProxyRunner(logger.Session("dns-proxy"))
 		if err != nil {
 			return nil, err
@@ -67,18 +66,18 @@ func (cmd *WorkerCommand) guardianRunner(logger lager.Logger) (ifrit.Runner, err
 			),
 		})
 
-		gdnServerFlags = append(gdnServerFlags, "--dns-server", lip)
+		gdnFlags = append(gdnFlags, "--dns-server", lip)
 
 		// must permit access to host network in order for DNS proxy address to be
 		// reachable
-		gdnServerFlags = append(gdnServerFlags, "--allow-host-access")
+		gdnFlags = append(gdnFlags, "--allow-host-access")
 	}
 
-	gdnArgs := append(gdnFlags, append([]string{"server"}, gdnServerFlags...)...)
+	gdnArgs := append([]string{"server"}, gdnFlags...)
 
 	bin := "gdn"
-	if cmd.Garden.Bin != "" {
-		bin = cmd.Garden.Bin
+	if cmd.Guardian.Bin != "" {
+		bin = cmd.Guardian.Bin
 	}
 
 	gdnCmd := exec.Command(bin, gdnArgs...)
@@ -99,9 +98,10 @@ func (cmd *WorkerCommand) guardianRunner(logger lager.Logger) (ifrit.Runner, err
 	return grouper.NewParallel(os.Interrupt, members), nil
 }
 
-var gardenEnvPrefix = "CONCOURSE_GARDEN_"
+// For historical reasons this is "CONCOURSE_GARDEN_" instead of "CONCOURSE_GUARDIAN_"
+var guardianEnvPrefix = "CONCOURSE_GARDEN_"
 
-func detectGardenFlags(logger lager.Logger) []string {
+func detectGuardianFlags(logger lager.Logger) []string {
 	env := os.Environ()
 
 	flags := []string{}
@@ -115,11 +115,11 @@ func detectGardenFlags(logger lager.Logger) []string {
 		name := spl[0]
 		val := spl[1]
 
-		if !strings.HasPrefix(name, gardenEnvPrefix) {
+		if !strings.HasPrefix(name, guardianEnvPrefix) {
 			continue
 		}
 
-		strip := strings.Replace(name, gardenEnvPrefix, "", 1)
+		strip := strings.Replace(name, guardianEnvPrefix, "", 1)
 		flag := flagify(strip)
 
 		logger.Info("forwarding-garden-env-var", lager.Data{
--- a/worker/workercmd/guardian.go
+++ b/worker/workercmd/guardian.go
@@ -31,7 +31,13 @@ func (cmd *WorkerCommand) guardianRunner(logger lager.Logger) (ifrit.Runner, err
 
 	members := grouper.Members{}
 
-	gdnFlags := []string{
+	gdnConfigFlag := []string{}
+
+	if cmd.Guardian.Config.Path() != "" {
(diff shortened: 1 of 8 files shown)


worker/workercmd/guardian.go
```diff
@@ -31,13 +31,7 @@ func (cmd *WorkerCommand) guardianRunner(logger lager.Logger) (ifrit.Runner, err
 
 	members := grouper.Members{}
 
-	gdnFlags := []string{}
-
-	if cmd.Garden.Config.Path() != "" {
```
xtreme-sameer-vohra, line 36: We might want to keep this even though it could be passed directly to guardian via the detectGuardianFlags as it helps operators by showing up in `concourse worker -h` and it has been there [historically](https://github.com/concourse/concourse/commit/a5323c4de6d4e4d6b6d1fb783772e7a1b31cdf25)

worker/workercmd/guardian.go
```diff
@@ -51,25 +45,9 @@ func (cmd *WorkerCommand) guardianRunner(logger lager.Logger) (ifrit.Runner, err
...
-	if cmd.ContainerNetworkPool != "" {
-		gdnServerFlags = append(gdnServerFlags, "--network-pool", cmd.ContainerNetworkPool)
-	}
-
-	if cmd.Garden.MaxContainers != 0 {
-		gdnServerFlags = append(gdnServerFlags, "--max-containers", strconv.Itoa(cmd.Garden.MaxContainers))
-	}
-
-	gdnServerFlags = append(gdnServerFlags, detectGardenFlags(logger)...)
+	gdnFlags = append(gdnFlags, detectGuardianFlags(logger)...)
```
chenbh, line 48: Have we ever talked about if we want to automatically detect and pass on flags for `containerd` as well?

worker/workercmd/worker_linux.go
```diff
@@ -22,26 +22,39 @@ type Certs struct {
 	Dir string `long:"certs-dir" description:"Directory to use when creating the resource certificates volume."`
 }
 
-type GardenBackend struct {
-	UseHoudini    bool `long:"use-houdini"    description:"Use the insecure Houdini Garden backend."`
-	UseContainerd bool `long:"use-containerd" description:"Use the containerd backend. (experimental)"`
+type Runtime struct {
+	Type string `long:"type" default:"guardian" description:"Runtime to use with the worker. Possible values: guardian, containerd, houdini. Please note that Houdini is insecure and doesn't run 'tasks' in containers."`
```
chenbh, line 26: `--runtime=containerd` feels better than `--runtime-type=containerd`, WDYT?

worker/workercmd/worker.go
```diff
@@ -151,10 +155,10 @@ func (cmd *WorkerCommand) Runner(args []string) (ifrit.Runner, error) {
 
 	var members grouper.Members
 
-	if !cmd.gardenIsExternal() {
+	if !cmd.gardenServerIsExternal() {
```
chenbh, line 158: Does it make sense to treat external garden server as a runtime type as well?

worker/workercmd/guardian.go
chenbh: I agree, having it be explicitly stated as a flag makes it more trustworthy and easier to discover than having it be implicit.

Even the [docs](https://concourse-ci.org/concourse-worker.html#configuring-gdn-server) say that it's more for backwards compatibility than as the preferred way of configuring garden

muntac: Great catch! Changed it back to what it was. Also renamed `gdnConfigFlags` to explicitly call out that it's being used for the `--config` option.

worker/workercmd/guardian.go
muntac: Good question. I haven't had the conversation with anyone. However, we allow for a config file to be passed (`CONCOURSE_CONTAINERD_CONFIG`) which I feel is a reasonable option, and easier for us to maintain. There also aren't that many options on the binary.

Screen Shot 2020-07-29 at 9 41 23 AM

worker/workercmd/worker.go
muntac: Hm this is a good point and something we missed out on. I agree that consolidating it with runtime type would be neater. The only problem is we'd have to break the existing contract, which might be okay.

@xtreme-sameer-vohra have you run into many users who use an external garden server?

worker/workercmd/worker_linux.go
muntac: This choice came down to thinking about runtimes available for Linux vs non-Linux platforms. "Runtime" is defined at the `worker.go` level. If we want different behaviours for `worker_linux.go` and `worker_nonlinux.go` we had to make it a struct that could be defined differently in those two files. This mean we can easily enable only Houdini in non-Linux, and not have to write out our own validation.

Although I suppose the validation wouldn't may not be too difficult to do in each of those files simply based off the `cmd.Runtime` value. @xtreme-sameer-vohra any thoughts on this?

muntac: @chenbh the other issue here was the message we show for `Runtime` has to be different for linux vs non-linux.

e.g.
```
type Runtime struct {
	Type string `long:"type" default:"guardian" description:"Runtime to use with the worker. Possible values: guardian, containerd, houdini. Please note that Houdini is insecure and doesn't run 'tasks' in containers."`
}
```
for Linux
versus
```
type Runtime struct {
}
```
providing no option at all for non-linux as it defaults to Houdini. We need this structure as in the future we will likely see containerd enabled for Windows, and then have to introduce type here and a message saying only `containerd` and `houdini` are allowed values. (We will also likely have to break that out into a worker_mac and worker_windows file at that point with the worker_mac still not allowing a value and defaulting to houdini.)

chenbh: Should be possible with struct inheritance

worker.go:
```golang
type WorkerCommand struct {
...
	Runtime `group:"Runtime Configuration"`
...
}
```

worker_linux.go:
```golang
// +build linux
type Runtime struct {
	Runtime string `long:"runtime" default:"guardian" choice:"guardian" choice:"containerd" choice:"houdini" description:"Runtime to use with the worker. Please note that Houdini is insecure and doesn't run 'tasks' in containers."`
}
```

worker_nonlinux.go:
```golang
// +build !linux
type Runtime struct {}
```

muntac: @chenbh Amazing! We tried that initially but didn't realize you could set the parent struct and not have to provide the `namespace` attribute. Made the suggested change to `CONCOURSE_RUNTIME` in BOSH release and Helm chart as well. Thanks!

worker/workercmd/worker.go
muntac: @chenbh I've opened a new issue to track this https://github.com/concourse/concourse/issues/5963 and put it in https://project.concourse-ci.org/projects/MDc6UHJvamVjdDUxMTY3MTg=. As this is a small use case and this PR is facing scope creep I suggest we deal with it later.

chenbh:
Looks like the unit tests are failing due to https://github.com/concourse/concourse/blob/4795488830f6bd723b95cb7ed6345a658605bff5/worker/integration/integration_test.go#L47

muntac:
Looks like the `concourse/dev` image which is used to run all the integration tests has the `CONCOURSE_GARDEN_DNS_PROXY_ENABLE` set by default which will break our validation when running tests with `containerd`.

I opened this https://github.com/concourse/ci/pull/357 to fix that. But going to do some poking around first to be sure.

PR #5932, opened 2020-07-30
Enforce SetPipeline policy check in set_pipeline step

### LINKED ISSUE #5931: SaveConfig policy check is not enforced when using set_pipeline step
## Summary

`SaveConfig` policy check is not enforced when using `set_pipeline` step, which allows any policy enforcement of pipeline configuration to be circumvented

## Steps to reproduce

1. run OPA with policy that restricts which third-party/custom resource types can be used, for example:
  ```rego
  package concourse
  
  default allow = false
  
  # allow all actions without explicit rule mapping
  allow {
    actions != "SaveConfig"
    not actions[input.action]
  }
  
  # verify set-pipeline 
  allow {
    input.action == "SaveConfig"
    count(resource_violations) == 0
  }
  
  # extract list of custom & third party
  resource_violations[image] {
    image := input.data.resource_types[_].source.repository
    not allowed_resources[image]
  }
  
  # define list of approved custom & third-party resource images
  allowed_resources := {
    "teliaoss/github-pr-resource"
  }
  ```
2. run concourse web node with policy enforcement enabled, and using OPA backend
3. apply malicious pipeline, for example
  ```yaml
  jobs:
    - name: sync-pipeline
      plan:
        # generate new pipeline config that uses an unsupported resource type
        - task: gen-pipeline
          config:
            platform: linux
            image_resource:
              type: docker-image
              source:
                repository: ubuntu
                tag: bionic
            outputs:
              - name: pipeline
            run:
              path: /bin/bash
              args:
                - -c
                - |
                  cat <<PIPELINE > pipeline/pipeline.yml
                  jobs:
                    - name: sync-pipeline
                      plan:
                        - task: test
                          config:
                            platform: linux
                            image_resource:
                              type: docker-image
                              source:
                                repository: ubuntu
                                tag: bionic
                            outputs:
                              - name: pipeline
                            run:
                              path: /bin/bash
                              args:
                                - -c
                                - |
                                  echo "hello world"
                  resource_types:
                    - name: malicious-resource
                      type: registry-image
                      source:
                        repository: hacky/malicious-resource
                  PIPELINE
        # apply malicious pipeline
        - set_pipeline: self
          file: pipeline/pipeline.yml

  resource_types:
    # use allowed resource type when first applying pipeline
    - name: pull-request-resource
      type: registry-image
      source:
        repository: teliaoss/github-pr-resource

  ```
4. execute job that runs `set_pipeline` using a pipeline definition that includes unsupported resource_types 

## Expected results

`set_pipeline` should fail

## Actual results

`set_pipeline` succeeds and pipeline is updated with unsupported resource_type

## Triaging info

* Concourse version: 6.4.0
* Did this used to work? n/a

cludden:
Signed-off-by: Chris Ludden <chris.ludden@gmail.com>

Bug Fix - adds policy check to `set_pipeline` step

closes #5931.

Enforces `SaveConfig` policy check (when enabled) in `set_pipeline` step

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/exec/set_pipeline_step.go
+++ b/atc/exec/set_pipeline_step.go
@@ -19,10 +19,13 @@ import (
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/atc/exec/artifact"
 	"github.com/concourse/concourse/atc/exec/build"
+	"github.com/concourse/concourse/atc/policy"
 	"github.com/concourse/concourse/atc/worker"
 	"github.com/concourse/concourse/tracing"
 	"github.com/concourse/concourse/vars"
 )
+
+const ActionRunSetPipeline = "SetPipeline"
 
 // SetPipelineStep sets a pipeline to current team. This step takes pipeline
 // configure file and var files from some resource in the pipeline, like git.
@@ -34,6 +37,7 @@ type SetPipelineStep struct {
 	teamFactory     db.TeamFactory
 	buildFactory    db.BuildFactory
 	client          worker.Client
+	policyChecker   policy.Checker
 }
 
 func NewSetPipelineStep(
@@ -44,6 +48,7 @@ func NewSetPipelineStep(
 	teamFactory db.TeamFactory,
 	buildFactory db.BuildFactory,
 	client worker.Client,
+	policyChecker policy.Checker,
 ) Step {
 	return &SetPipelineStep{
 		planID:          planID,
@@ -53,6 +58,7 @@ func NewSetPipelineStep(
 		teamFactory:     teamFactory,
 		buildFactory:    buildFactory,
 		client:          client,
+		policyChecker:   policyChecker,
 	}
 }
 
@@ -220,8 +226,27 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState, delegate S
 		return true, nil
 	}
 
+	// conditionally check step
+	if step.policyChecker != nil && step.policyChecker.ShouldCheckAction(ActionRunSetPipeline) {
+		input := policy.PolicyCheckInput{
+			Action:   ActionRunSetPipeline,
+			Team:     team.Name(),
+			Pipeline: step.plan.Name,
+			Data:     &atcConfig,
+		}
+		result, err := step.policyChecker.Check(input)
+		if err != nil {
+			return false, fmt.Errorf("error checking policy enforcement")
+		}
+		if !result.Allowed {
+			return false, fmt.Errorf("policy check failed for set_pipeline: %s", strings.Join(result.Reasons, ", "))
+		}
+		logger.Debug("policy check passed for set_pipeline")
+	}
+
 	fmt.Fprintf(stdout, "setting pipeline: %s\n", pipelineRef.String())
 	delegate.SetPipelineChanged(logger, true)
+
 	parentBuild, found, err := step.buildFactory.Build(step.metadata.BuildID)
 	if err != nil {
 		return false, err
--- a/atc/engine/step_factory.go
+++ b/atc/engine/step_factory.go
@@ -180,6 +180,7 @@ func (factory *coreStepFactory) SetPipelineStep(
 		factory.teamFactory,
 		factory.buildFactory,
 		factory.client,
+		delegateFactory.policyChecker,
 	)
 
 	spStep = exec.LogError(spStep, delegateFactory)
--- a/atc/exec/set_pipeline_step_test.go
+++ b/atc/exec/set_pipeline_step_test.go
@@ -3,6 +3,7 @@ package exec_test
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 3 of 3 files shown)


atc/exec/set_pipeline_step.go
```diff
@@ -217,6 +221,25 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState) error {
 		return nil
 	}
 
+	// conditionally check step
+	if step.policyChecker != nil && step.policyChecker.ShouldCheckHttpMethod("PUT") || step.policyChecker.ShouldCheckAction(atc.SaveConfig) {
+		input := policy.PolicyCheckInput{
+			HttpMethod: "PUT",
```
vito, line 227: It feels a little awkward to be faking a HTTP request for this. I wonder if we should have this be its own action, similar to `UseImage`?

cludden: I agree, I mainly kept it for consistency with the current behavior of `SaveConfig`, but happy to refactor into its own action if that's preferred.

evanchaoli: I also think it's better to define a separate action for this set_pipeline step.

evanchaoli:
Yeah, I think I missed `set_pipeline` in my pr. Thank @cludden for the fix.

atc/exec/set_pipeline_step.go
cludden: Updated to omit `HttpMethod` and use new `RunSetPipeline` action

atc/exec/set_pipeline_step.go
```diff
@@ -19,22 +19,26 @@ import (
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/atc/exec/artifact"
 	"github.com/concourse/concourse/atc/exec/build"
+	"github.com/concourse/concourse/atc/policy"
 	"github.com/concourse/concourse/atc/worker"
 	"github.com/concourse/concourse/tracing"
 	"github.com/concourse/concourse/vars"
 )
 
+const ActionRunSetPipeline = "RunSetPipeline"
```
vito, line 28: Nice!

@evanchaoli @cludden Follow-up: do you think it would be valuable to have this single action run in both paths (set via API vs set via step)? Then the policy writer would only need to implement one policy. Right now it seems easy to forget to do one or the other.

If so, I think this could be renamed to just `SetPipeline`. Assuming the current name is from "run `set_pipeline` step".

A deeper follow-up topic: I wonder if it would be safer to move away from generic API policy actions and instead implement specific actions like this. It seems like API-level policy enforcement may not be the most flexible thing going forward, since there are ways for an action to occur without going through the API. This is a trend that I expect to continue as we make Concourse more capable of automating itself.

Are there other common situations where we want to vet the API request, including the request payload? Can those be turned into more specific actions as well?

I recognize that being able to enforce policies for all API endpoints is super flexible and a cheap way to make the policy enforcement really powerful, but it works by casting a pretty large net, and it may become really fiddly over time. 🤔 I'm just asking these questions now while it's still experimental.

Thanks!

cludden: FWIW, I agree. My original PR intended to keep the same behavior regardless of where `SaveConfig` is invoked (cli, api, set_pipeline step). A quick fix to move this PR forward would be to rename `SaveConfig` to `SetPipeline` and remove the associated HTTP method from the check input in both places (similar to `UseImage`). A medium effort fix, most likely outside the scope of this PR, would be to remove the HTTP method filters entirely, and simply rely on action based filters. A larger refactor would be to move the policy check away from the edge, closer to the database (where exactly, im not sure, as I'm not super familiar with the concourse codebase). Thoughts on how to proceed for this PR? I'd love to see this gap fixed in the next concourse release, but agree that the policy enforcement feature as a whole feels a bit half-baked. One larger change I'd like to see is a shift towards a deny based policy check, that allows for a list of violations to be returned and propagated downstream to the user. The current blanket policy failure error can be frustrating. Happy to take a stab at any of this if there's a preferred path forward.

cludden: @vito @evanchaoli any thoughts on how to proceed?

vito: @cludden Sorry for the wait on this - just got back from vacation.

@andy-paine has split the above out into a separate discussion: https://github.com/concourse/concourse/discussions/6013 (thanks!)

For the sake of avoiding scope creep, I think we can review this PR as-is now that the larger issue has been split out into a separate discussion so that the context doesn't get lost. 👍 I'll try to wrap this up today!

It looks like this needs a rebase now though.

andy-paine: I suspect the rebase is partly due to the "reasons" changes. Happy to help getting this sorted on top of that if needed.

cludden: Updated PR with:
- rebase to pick up changes
- rename `RunSetPipeline` to `SetPipeline`
- refactored to support new policy check output (thanks for adding @andy-paine !)

vito (changes requested):
Argh, sorry this grew stale again, I dropped the ball on a few PRs while figuring out how to deal with the Docker Hub rate limiting deadline. If you're down to rebase again I'll get this through more quickly.

Also there seem to be some rogue commits in here, probably from rebasing - could those be removed?

Thanks for your patience!

cludden:
I updated the PR, but it looks like the concourse was down when the PR checks ran..

vito:
@cludden Thanks! I'll manually kick off the builds.

vito (approved):
All set - many thanks!

PR #5942, opened 2020-08-04
feat(atc): allows to specify a registry-mirror url to ATC, and expose the url to resources via environment variable.

evanchaoli:

My ultimate goal is to make `registry-image` to transparently pull images from a mirror that is configured to ATC. To archive that, we need to:

1. Add a flag to `concourse web` to configure a registry mirror URL.
2. Expose the mirror URL to resource's `check/get/put` via environment variable.
3. Enhance `registry-image` resource type to check the environment variable, if there is a mirror passing from env, then pull from the mirror first and if pulling from mirror fails, then fail back to original registry.

This PR is addressing 1&2. 

## Notes to reviewer:

This PR is calling for review of the idea, no tests have been done yet. If the idea is acceptable, then I will add test cases and run tests.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -129,6 +129,8 @@ type RunCommand struct {
 
 	ExternalURL flag.URL `long:"external-url" description:"URL used to reach any ATC from the outside world."`
 
+	RegistryMirrorURL flag.URL `long:"registry-mirror-url" description:"URL of a registry mirror"`
+
 	Postgres flag.PostgresConfig `group:"PostgreSQL Configuration" namespace:"postgres"`
 
 	ConcurrentRequestLimits   map[wrappa.LimitedRoute]int `long:"concurrent-request-limit" description:"Limit the number of concurrent requests to an API endpoint (Example: ListAllJobs:5)"`
@@ -1581,6 +1583,7 @@ func (cmd *RunCommand) constructEngine(
 		stepFactory,
 		builder.NewDelegateFactory(),
 		cmd.ExternalURL.String(),
+		cmd.RegistryMirrorURL.String(),
 		secretManager,
 		cmd.varSourcePool,
 		cmd.EnableRedactSecrets,
--- a/atc/engine/builder/builder.go
+++ b/atc/engine/builder/builder.go
@@ -43,27 +43,30 @@ func NewStepBuilder(
 	stepFactory StepFactory,
 	delegateFactory DelegateFactory,
 	externalURL string,
+	registryMirrorURL string,
 	secrets creds.Secrets,
 	varSourcePool creds.VarSourcePool,
 	redactSecrets bool,
 ) *stepBuilder {
 	return &stepBuilder{
-		stepFactory:     stepFactory,
-		delegateFactory: delegateFactory,
-		externalURL:     externalURL,
-		globalSecrets:   secrets,
-		varSourcePool:   varSourcePool,
-		redactSecrets:   redactSecrets,
+		stepFactory:       stepFactory,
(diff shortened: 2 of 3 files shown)


vito:
Thanks for bringing this up! The use case totally makes sense and has come up before - it's definitely too much to ask of users to set a mirror everywhere. 👍 

I think setting it as an env var in all resource containers is a little too brittle though. It's a leaky abstraction; the entire point of using resources for image fetching is that Concourse doesn't have to know anything about it, but here we're adding a first-class flag and env var to thread all the way down to it, which effectively becomes part of the resource interface itself.

This will also get more complicated once more values need to be specified - not only is there more env to set, but what if someone has a registry mirror that requires credentials? We surely wouldn't want to propagate those values to every single resource container.

I understand that this is just an early idea to demonstrate the need though, and it does that well - I don't mean to sound harsh. 🙂 We just need to figure out a good way to go from here.

When this came up in the past I had thought about it as 'defaults for resource types.' The idea is that we can specify default config values for the resource type and let Concourse apply the defaults whenever it is used. This way the resource types don't even have to know that it's happening, and it can be applied to any resource type for any reason (e.g. `https_tunnel` on the `git` resource).

I don't have a clear idea of how this would work, but I'd be happy to work through it.

Here are some thoughts:

* It should work for base resource types (`registry-image`) in addition to types defined under `resource_types:`.
* It may be a good idea to set these defaults as *another* resource type, rather than modifying the type itself.
  * This way you can use the original, non-defaulted version alongside the defaulted version.
  * This way it doesn't matter whether the original type was a base type or a type defined in `resource_types:`.
  * It should still be possible to use this to *override* the original resource type, just as you're able to do with `resource_types:` today.
    * Deciding whether to override would at the discretion of the operator/user. In your case, you could do this to have `registry-image` "just work" everywhere.
* It should be possible to set cluster-wide, for your use case, though this may turn into project-wide in the future as we work towards concourse/rfcs#30 and concourse/rfcs#32.

Here's a quick mockup of how this could look:

```yaml
resource_types:
- name: registry-image
  alias: registry-image
  defaults:
    registry_mirror:
      host: my-mirror.com
      username: ...
      password: ...
```

The `alias:` field there effectively results in a copy of the original `resource-image` type, and the values under `defaults:` will act as if they are set in the `source:` of any resource which uses the type. This new type could be given a different `name:` value, but in this case we have chosen to override the old one so that the change is transparent to users.

I suppose `defaults:` could also be set directly on a resource type definition:

```yaml
resource_types:
- name: git
  type: registry-image
  source:
    repository: vito/my-git-resource
  defaults:
    https_tunnel:
      proxy_host: git.example.com
      proxy_port: 443
```

Once this is added, the last piece of the puzzle would be to add some way to configure cluster-wide resource types, which has already come up as a possible solution for concourse/rfcs#30 but this is the first thing to really drive it out. 🙂

What do you think?

evanchaoli:
> I understand that this is just an early idea to demonstrate the need though, and it does that well - I don't mean to sound harsh. 🙂 We just need to figure out a good way to go from here.

@vito Yeah. I just tried to use some quick code to test your opinion. I am happy you are with the idea.

I agree exporting a specific configuration to all resources via env is not ideal, making resource type to support default is a better idea. Let me think more then come back.

PR #5943, opened 2020-08-04
Optimize team roles computation in API

### LINKED ISSUE #5912: Listing all teams takes a long time due to redundant auth checking

aoldershaw:

closes #5912 .

* Precompute team roles once when we construct the accessor, rather than every time we try to check if a user has access to a team

## Notes to reviewer:

Opening a separate PR from #5897 since this is an independent change, and we are going to have a discussion about auth in general soon, so the approach is kind of up in the air.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/api/accessor/accessor.go
+++ b/atc/api/accessor/accessor.go
@@ -42,6 +42,8 @@ type access struct {
 	systemClaimKey    string
 	systemClaimValues []string
 	teams             []db.Team
+	teamRoles         map[string][]string
+	isAdmin           bool
 }
 
 func NewAccessor(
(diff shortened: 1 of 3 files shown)


aoldershaw:
Nevermind, going to close this and merge #5897 (which has since been already approved)

PR #5951, opened 2020-08-05
testflight: prevent pipeline from getting GC'd

aoldershaw:

Since we were calling `set_pipeline: self` but removing the job that set the pipeline initially,  the pipeline would occasionally get archived prior to triggering the "normal-job" job. This would cause fly to hang forever since archived pipelines are effectively paused.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [ ] Updated [Release notes]

## Reviewer Checklist

- [x] ~Code reviewed~
- [x] Tests reviewed
- [x] PR acceptance performed

### DIFF
--- a/testflight/fixtures/set-pipeline.yml
+++ b/testflight/fixtures/set-pipeline.yml
@@ -7,6 +7,11 @@ resources:
       pipeline.yml: |
         ---
         jobs:
+        # The sp job is required so the pipeline is not GC'd after `set_pipeline: self`
+        - name: sp
+          plan:
+          - set_pipeline: self
+            file: does-not-matter
         - name: normal-job
           public: true
           plan:


chenbh:
https://github.com/concourse/concourse/blob/54bf7721464ad67789497b606b663326dd601eb4/testflight/set-pipeline-step_test.go#L136-L143 
is the only test that had this issue right?

aoldershaw:
@chenbh yeah, all the other tests set another pipeline, and so the `sp` job on the parent pipeline wouldn't become inactive after setting the child pipeline (even if the fixture didn't have the `sp` job.

So we *are* adding an unnecessary job to the child pipelines in the other tests, but eh, I figured this was the easiest fix. I mean, one alternative could be parameterizing the fixture that's used in the test, and adding another fixture like `set-pipeline-self.yml` for this one test. But that fixture would also need the same comment explaining why the `sp` job was necessary, and I figured it wasn't worth it.

I also thought about just renaming both `sp` and `normal-job` to something generic like `job`, but that's a little too magical, and it's not obvious that the jobs need the same name

aoldershaw:
@jamieklassen good point, we may not need it. Going to merge this for now though, since the frequent failures on testflight are making it pretty annoying to get PRs to go green

PR #5957, opened 2020-08-06
Add flag ` --ignore-event-parsing-errors` to `fly watch` to ignore event parsing errors when an unknown event type or version is encountered. 

### LINKED ISSUE #5938: Fly fails when it encounters an unknown event type
## Summary

When fly encounters an unknown event type, it results in an error and fly exits.

Reference: [failing downgrade build](https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/downgrade/builds/352#L5f1e56ed:198:199)

In this case, a new event type was added to Concourse, a build was executed which resulted in the DB containing events of the new type. Subsequently, a down migration was performed and attempting to display the logs via `fly watch -u PREVIOUSLY_EXECUTED_BUILD` resulted in an error

```
failed to parse next event: unknown event type: NEW_EVENT_TYPE_NAME
```

## Steps to reproduce

1. Build & deploy Concourse with a new event type 
   ```
   cd concourse && git checkout eeeef8104d1b99532f18b1ba543a04424cdc6ef1 && docker-compose up --build -d 
   ```
  
1. Execute a pipeline job with a `get/put/task` step. Observe the UI or fly displays the new event (selected worker in this example)
   ```
   fly set-pipeline -p booklit -c booklit.yml
   fly trigger-job -j booklit/unit -w
   ```
1. Downgrade Concourse ( eg concourse/concourse:6.4.0 )
   ```
   docker-compose up --build -d 
   ```
1. Execute `fly watch -u PREVIOUSLY_EXECUTED_BUILD_URL`

## Expected results

Build output to be displayed & a warning that the build logs contain and unknown event type. While it might be overbearing to port this to older versions of Concourse and thus fly, it would make sense to put this in so it doesn't affect future versions.

## Actual results

Fly watch fails and exits with an error
```
failed to parse next event: unknown event type: selected-worker
```
## Triaging info

* Concourse version: Any version with a new event type being downgraded to a previous version without the event type. [Reference commit adds a new event type selected-worker](https://github.com/concourse/concourse/commit/eeeef8104d1b99532f18b1ba543a04424cdc6ef1)) downgraded to 6.4.x

* Did this used to work? No

xtreme-sameer-vohra:
Feature
closes #5938 

adding a toggle flag to `fly watch` that ignores parsing errors
of type;
- UnknownEventTypeError 
- UnknownEventVersionError

This prevents issues if a user were to trigger jobs on a version of
Concourse and then downgrade/upgrade Concourse to a version where an
event type is not recognized causing fly to bail. In such scenarios, the
user would have to determine the correct fly version to use to correctly
parse the build events, which seems quite cumbersome.

## Notes to reviewer:
For testing, easiest method is to trigger a job in a pipeline, edit a record in the build_events_X table with a bogus event and attempt to `fly watch` the build 

## Contributor Checklist

- [x] [Signed] all commits

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/fly/eventstream/render.go
+++ b/fly/eventstream/render.go
@@ -12,7 +12,8 @@ import (
 )
 
 type RenderOptions struct {
-	ShowTimestamp bool
+	ShowTimestamp            bool
+	IgnoreEventParsingErrors bool
 }
 
 func Render(dst io.Writer, src eventstream.EventStream, options RenderOptions) int {
@@ -25,11 +26,20 @@ func Render(dst io.Writer, src eventstream.EventStream, options RenderOptions) i
 		if err != nil {
 			if err == io.EOF {
 				return exitStatus
+			} else if _, ok := err.(event.UnknownEventTypeError); ok && options.IgnoreEventParsingErrors {
+				dstImpl.SetTimestamp(0)
+				fmt.Fprintf(dstImpl, "failed to parse next event: %s\n", ui.ErroredColor.Sprint(err))
+				continue
+			} else if _, ok := err.(event.UnknownEventVersionError); ok && options.IgnoreEventParsingErrors {
+				dstImpl.SetTimestamp(0)
+				fmt.Fprintf(dstImpl, "failed to parse next event: %s\n", ui.ErroredColor.Sprint(err))
+				continue
 			} else {
 				dstImpl.SetTimestamp(0)
 				fmt.Fprintf(dstImpl, "failed to parse next event: %s\n", err)
 				return 255
 			}
+
 		}
 
 		switch e := ev.(type) {
--- a/fly/eventstream/render.go
+++ b/fly/eventstream/render.go
@@ -26,20 +26,13 @@ func Render(dst io.Writer, src eventstream.EventStream, options RenderOptions) i
 		if err != nil {
 			if err == io.EOF {
 				return exitStatus
-			} else if _, ok := err.(event.UnknownEventTypeError); ok && options.IgnoreEventParsingErrors {
-				dstImpl.SetTimestamp(0)
-				fmt.Fprintf(dstImpl, "failed to parse next event: %s\n", ui.ErroredColor.Sprint(err))
-				continue
-			} else if _, ok := err.(event.UnknownEventVersionError); ok && options.IgnoreEventParsingErrors {
-				dstImpl.SetTimestamp(0)
-				fmt.Fprintf(dstImpl, "failed to parse next event: %s\n", ui.ErroredColor.Sprint(err))
+			} else if options.IgnoreEventParsingErrors && isEventParseError(err) {
 				continue
 			} else {
 				dstImpl.SetTimestamp(0)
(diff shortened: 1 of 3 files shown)


aoldershaw:
I wonder if we want this to be the default? If you were to downgrade with builds that contained a new event type, it would be annoying to always have to pass this flag in. Or similarly, if we were to remove event types (I think we just deprecate them now?), you'd have to always pass this in. WDYT?

xtreme-sameer-vohra:
> I wonder if we want this to be the default? If you were to downgrade with builds that contained a new event type, it would be annoying to always have to pass this flag in. Or similarly, if we were to remove event types (I think we just deprecate them now?), you'd have to always pass this in. WDYT?

My assumption is would only apply to a small subset of builds. This flag would only be limited to `watch` on older builds as `execute`, `trigger-job`, etc all would generate new build events corresponding to the running Concourse version. 

Also, if it defaulted to `true`, folks might not even notice the printed warning.

I dont hold this opinion strongly, so if we have set precedence in other places of doing something similar, I can make the change.

aoldershaw:
I guess it's more consistent with the UI to fail by default
https://github.com/concourse/concourse/blob/587688726f810fd57b985f9f0aae4e1b277fe6f0/web/elm/src/Concourse/BuildEvents.elm#L189-L190

fly/eventstream/render_test.go
```diff
@@ -318,4 +318,51 @@ var _ = Describe("V1.0 Renderer", func() {
...
+		})
+
+	})
+
 })
+
+type someNewEvent struct{}
+
+func (someNewEvent) EventType() atc.EventType  { return "some-new-event" }
+func (someNewEvent) Version() atc.EventVersion { return "1.0" }
```
aoldershaw, line 368: Looks like this is unused

fly/eventstream/render.go
```diff
@@ -25,9 +26,13 @@ func Render(dst io.Writer, src eventstream.EventStream, options RenderOptions) i
 		if err != nil {
 			if err == io.EOF {
 				return exitStatus
+			} else if options.IgnoreEventParsingErrors && isEventParseError(err) {
+				dstImpl.SetTimestamp(0)
+				fmt.Fprintf(dstImpl, "failed to parse next event: %s\n", ui.ErroredColor.Sprint(err))
```
aoldershaw, line 31: Do we need to log this? Presumably if they're passing the flag, they've already encountered that there is an unknown event type/version. Also, it is called `--ignore-event-parsing-errors`, so I think makes sense to ignore them completely

fly/eventstream/render_test.go
xtreme-sameer-vohra: whoops, good catch

fly/eventstream/render.go
xtreme-sameer-vohra: yea, thats fair 👍

PR #5958, opened 2020-08-06
Rerun builds with baggageclaim network errors

kirillbilchenko:
Signed-off-by: kirillbilchenko <kirya7@gmail.com>

closes # .

Add case to cover issue like this `Put "/volumes/a2dc0c2a-5057-4d0e-794f-ae7607ce8638/stream-in?path=.": EOF` which causing all job to fail
## Notes to reviewer:
Added test to cover this io.EOF case.

## Release Note

* The `--enable-rerun-when-worker-disappears` flag now supports rerunning builds after any network error from the ATC to the worker's baggageclaim. Such network errors are common when the worker disappears.
* Builds will now be rerun when this flag is enabled and the failing step is a nested step (e.g. within an `in_parallel`)

## Contributor Checklist

- [x] [Signed] all commits
- [x] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/engine/engine.go
+++ b/atc/engine/engine.go
@@ -240,7 +240,7 @@ func (b *engineBuild) Run(ctx context.Context) {
 	case err = <-done:
 		logger.Debug("engine-build-done")
 		if err != nil {
-			if _, ok := err.(exec.Retriable); ok {
+			if ok := errors.As(err, &exec.Retriable{}); ok {
 				return
 			}
 		}
--- a/atc/engine/engine.go
+++ b/atc/engine/engine.go
@@ -240,7 +240,7 @@ func (b *engineBuild) Run(ctx context.Context) {
 	case err = <-done:
 		logger.Debug("engine-build-done")
 		if err != nil {
-			if _, ok := err.(exec.Retriable); ok {
+			if _, ok := err.(exec.Retryable); ok {
 				return
 			}
 		}
--- a/atc/engine/engine.go
+++ b/atc/engine/engine.go
@@ -240,7 +240,7 @@ func (b *engineBuild) Run(ctx context.Context) {
 	case err = <-done:
 		logger.Debug("engine-build-done")
 		if err != nil {
-			if _, ok := err.(exec.Retryable); ok {
+			if _, ok := err.(exec.Retriable); ok {
 				return
 			}
 		}
--- a/atc/exec/in_parallel.go
+++ b/atc/exec/in_parallel.go
@@ -93,8 +93,8 @@ func (step InParallelStep) Run(ctx context.Context, state RunState) error {
 		return ctx.Err()
 	}
 
-	if merr, ok := result.(*multierror.Error); ok {
-		return fmt.Errorf("one or more parallel steps errored:\n%w", merr)
+	if result != nil {
+		return result
 	}
 
 	return nil
--- a/atc/exec/in_parallel.go
+++ b/atc/exec/in_parallel.go
@@ -6,7 +6,8 @@ import (
 	"fmt"
 	"os"
 	"runtime/debug"
-	"strings"
+
+	"github.com/hashicorp/go-multierror"
 )
 
 // InParallelStep is a step of steps to run in parallel.
@@ -77,14 +78,14 @@ func (step InParallelStep) Run(ctx context.Context, state RunState) error {
 		executedSteps++
 	}
 
-	var errorMessages []string
+	var result error
 	for i := 0; i < executedSteps; i++ {
 		err := <-errs
 		if err != nil && !errors.Is(err, context.Canceled) {
 			// The Run context being cancelled only means that one or more steps failed, not
 			// in_parallel itself. If we return context.Canceled error messages the step will
 			// be marked as errored instead of failed, and therefore they should be ignored.
-			errorMessages = append(errorMessages, err.Error())
+			result = multierror.Append(result, err)
 		}
 	}
 
@@ -92,8 +93,8 @@ func (step InParallelStep) Run(ctx context.Context, state RunState) error {
 		return ctx.Err()
 	}
 
-	if len(errorMessages) > 0 {
-		return fmt.Errorf("one or more parallel steps errored:\n%s", strings.Join(errorMessages, "\n"))
+	if merr, ok := result.(*multierror.Error); ok {
+		return fmt.Errorf("one or more parallel steps errored:\n%w", merr)
 	}
 
 	return nil
--- a/atc/exec/retry_error_step.go
+++ b/atc/exec/retry_error_step.go
@@ -1,16 +1,16 @@
 package exec
 
 import (
+	"code.cloudfoundry.org/lager"
+	"code.cloudfoundry.org/lager/lagerctx"
 	"context"
 	"errors"
 	"fmt"
+	"github.com/concourse/concourse/atc/worker/transport"
 	"net"
+	"net/url"
 	"reflect"
 	"regexp"
-
-	"code.cloudfoundry.org/lager"
-	"code.cloudfoundry.org/lager/lagerctx"
-	"github.com/concourse/concourse/atc/worker/transport"
 )
 
 type Retriable struct {
@@ -50,15 +50,12 @@ func (step RetryErrorStep) Run(ctx context.Context, state RunState) error {
 }
 
 func (step RetryErrorStep) toRetry(logger lager.Logger, err error) bool {
-	var transportErr *transport.WorkerMissingError
-	var unreachable *transport.WorkerUnreachableError
-	var netOpErr *net.OpError
-	re := regexp.MustCompile(`worker .+ disappeared`)
-	if ok := errors.As(err, &transportErr) || errors.As(err, &unreachable) || errors.As(err, &netOpErr); ok {
+	var urlError *url.Error
+	if errors.As(err, &transport.WorkerMissingError{}) || errors.As(err, &transport.WorkerUnreachableError{}) || errors.As(err, &urlError) {
 		logger.Debug("retry-error",
 			lager.Data{"err_type": reflect.TypeOf(err).String(), "err": err.Error()})
 		return true
-	} else if re.MatchString(err.Error()) {
+	} else if _, ok := err.(net.Error); ok || regexp.MustCompile(`worker .+ disappeared`).MatchString(err.Error()) {
 		logger.Debug("retry-error",
 			lager.Data{"err_type": reflect.TypeOf(err).String(), "err": err})
 		return true
--- a/atc/exec/retry_error_step.go
+++ b/atc/exec/retry_error_step.go
@@ -42,7 +42,7 @@ func (step RetryErrorStep) Run(ctx context.Context, state RunState) error {
 	logger := lagerctx.FromContext(ctx)
 	runErr := step.Step.Run(ctx, state)
 	if runErr != nil && step.toRetry(logger, runErr) {
-		logger.Info("retryable", lager.Data{"error": runErr.Error()})
+		logger.Info("retriable", lager.Data{"error": runErr.Error()})
 		step.delegate.Errored(logger, fmt.Sprintf("%s, will retry ...", runErr.Error()))
 		runErr = Retriable{runErr}
 	}
--- a/atc/exec/retry_error_step.go
+++ b/atc/exec/retry_error_step.go
@@ -2,20 +2,22 @@ package exec
 
 import (
 	"context"
+	"errors"
 	"fmt"
-	"github.com/concourse/concourse/atc/worker/transport"
+	"io"
 	"reflect"
 	"regexp"
 
 	"code.cloudfoundry.org/lager"
 	"code.cloudfoundry.org/lager/lagerctx"
+	"github.com/concourse/concourse/atc/worker/transport"
 )
 
-type Retriable struct {
+type Retryable struct {
 	Cause error
 }
 
-func (r Retriable) Error() string {
+func (r Retryable) Error() string {
 	return fmt.Sprintf("retriable: %s", r.Cause.Error())
 }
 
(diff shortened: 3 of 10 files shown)


atc/exec/retry_error_step.go
```diff
@@ -40,26 +42,30 @@ func (step RetryErrorStep) Run(ctx context.Context, state RunState) error {
...
 	return runErr
 }
 
 func (step RetryErrorStep) toRetry(logger lager.Logger, err error) bool {
-	switch err.(type) {
-	case transport.WorkerMissingError, transport.WorkerUnreachableError:
+	var transportErr *transport.WorkerMissingError
+	var unreachable *transport.WorkerUnreachableError
+	re := regexp.MustCompile(`worker .+ disappeared`)
+	if ok := errors.Is(err, transportErr) || errors.Is(err, unreachable); ok {
```
aoldershaw, line 56: https://blog.golang.org/go1.13-errors

> In the simplest case, the errors.Is function behaves like a comparison to a sentinel error, and the errors.As function behaves like a type assertion. When operating on wrapped errors, however, these functions consider all the errors in a chain.

I think you want to be using `errors.As` here, not `errors.Is`

atc/exec/retry_error_step.go
```diff
@@ -40,26 +42,30 @@ func (step RetryErrorStep) Run(ctx context.Context, state RunState) error {
...
+		logger.Debug("retry-error",
+			lager.Data{"err_type": reflect.TypeOf(err).String(), "err": err.Error()})
+		return true
+	} else if re.MatchString(err.Error()) {
+		logger.Debug("retry-error",
+			lager.Data{"err_type": reflect.TypeOf(err).String(), "err": err})
+		return true
+	} else if err == io.EOF {
```
aoldershaw, line 64: Similarly, I think you want to use `errors.Is` here (in case the `io.EOF` was wrapped)

atc/engine/engine.go
```diff
@@ -240,7 +240,7 @@ func (b *engineBuild) Run(ctx context.Context) {
 	case err = <-done:
 		logger.Debug("engine-build-done")
 		if err != nil {
-			if _, ok := err.(exec.Retriable); ok {
+			if _, ok := err.(exec.Retryable); ok {
```
aoldershaw, line 243: Not really a big deal, but I was curious about this. According to https://english.stackexchange.com/a/305433, there isn't a "correct" spelling, and both forms are pretty much equal in terms of usage.

kirillbilchenko: Ide was complaining about this, so I just changed according to suggestion, I can revert back

kirillbilchenko:
@aoldershaw also I observed that this retry is not always working for our cluster I posted the same issue in discord dev channel `by the way seems like feature with retrying step when worker disappear is not working 
Get "http://127.0.0.1:8080/containers": worker ip-172-16-78-20 disappeared while trying to reach it, will retry ...
 but nothing happening in reality`

kirillbilchenko:
Screenshot 2020-08-06 at 22 18 39
Recently observed the same on our real cluster

aoldershaw:
> Recently observed the same on our real cluster

@kirillbilchenko hmm interesting, is that with or without this patch?

kirillbilchenko:
@aoldershaw this is without with latest available version of concourse 6.4.x

aoldershaw:
Interesting...I was able to get the `EOF, will retry ...` error message again in the UI, but I'm not seeing any of the log messages that I'd expect to see. I see neither:

https://github.com/concourse/concourse/blob/549f60b6c204980e909621d5530410f332941e8d/atc/exec/retry_error_step.go#L65-L66

nor:

https://github.com/concourse/concourse/blob/549f60b6c204980e909621d5530410f332941e8d/atc/exec/retry_error_step.go#L45

yet I do see:

https://github.com/concourse/concourse/blob/549f60b6c204980e909621d5530410f332941e8d/atc/exec/retry_error_step.go#L46

so it's clearly hitting that code path...

kirillbilchenko:
@aoldershaw maybe @evanchaoli can help on this, cuz he was initially author of this feature.

aoldershaw:
Okay, I figured out why it wasn't working for me. I was testing something with `on_error` before this, and was trying to force the `EOF` error in the `on_error` hook. It looks like there's actually a bug in `atc/exec/on_error.go`:

https://github.com/concourse/concourse/blob/587688726f810fd57b985f9f0aae4e1b277fe6f0/atc/exec/on_error.go#L44

This should be:

```go
	errs = multierror.Append(errs, err)
```

(also, just noticed that https://github.com/concourse/concourse/blob/587688726f810fd57b985f9f0aae4e1b277fe6f0/atc/exec/on_error.go#L40 should be `errors.Is(stepRunErr, context.Canceled)`)

Additionally, this:

https://github.com/concourse/concourse/blob/587688726f810fd57b985f9f0aae4e1b277fe6f0/atc/engine/engine.go#L243

should be:

```go
	if errors.As(err, &exec.Retriable{}) {
```

With those changes, I'm able to get the retry with EOF to work for me.

I'm guessing your erroring `put` step in the screenshot you provided isn't in an `on_error` hook, though?

kirillbilchenko:
@aoldershaw no it's just a put step after generation of docker images we need to upload them, so it's in in parallel block as a separate put steps

kirillbilchenko:
> Okay, I figured out why it wasn't working for me. I was testing something with `on_error` before this, and was trying to force the `EOF` error in the `on_error` hook. It looks like there's actually a bug in `atc/exec/on_error.go`:
> 
> https://github.com/concourse/concourse/blob/587688726f810fd57b985f9f0aae4e1b277fe6f0/atc/exec/on_error.go#L44
> 
> This should be:
> 
> ```go
> 	errs = multierror.Append(errs, err)
> ```
> 
> (also, just noticed that
> 
> https://github.com/concourse/concourse/blob/587688726f810fd57b985f9f0aae4e1b277fe6f0/atc/exec/on_error.go#L40
> 
> should be `errors.Is(stepRunErr, context.Canceled)`)
> Additionally, this:
> 
> https://github.com/concourse/concourse/blob/587688726f810fd57b985f9f0aae4e1b277fe6f0/atc/engine/engine.go#L243
> 
> should be:
> 
> ```go
> 	if errors.As(err, &exec.Retriable{}) {
> ```
> 
> With those changes, I'm able to get the retry with EOF to work for me.
> 
> Is the erroring `put` step in the screenshot you provided in a hook (`on_success`, `on_failure`, `ensure`, etc.)?

@aoldershaw should I change this in scope of this pr?

aoldershaw:
> @aoldershaw no it's just a put step after generation of docker images we need to upload them, so it's in in parallel block as a separate put steps

Oooh interesting...it looks like the `in_parallel` step also won't work with the `Retriable` error type:

https://github.com/concourse/concourse/blob/587688726f810fd57b985f9f0aae4e1b277fe6f0/atc/exec/in_parallel.go#L95-L97

We could use a `multierror` here which would be better since it will work with `errors.As`. I wonder what other wrapper steps this won't work with 🤔 

> @aoldershaw should I change this in scope of this pr?

Up to you - if you wouldn't mind taking on some more scope to fix up the feature generally, we'd really appreciate it! But no pressure

kirillbilchenko:
> > @aoldershaw no it's just a put step after generation of docker images we need to upload them, so it's in in parallel block as a separate put steps
> 
> Oooh interesting...it looks like the `in_parallel` step also won't work with the `Retriable` error type:
> 
> https://github.com/concourse/concourse/blob/587688726f810fd57b985f9f0aae4e1b277fe6f0/atc/exec/in_parallel.go#L95-L97
> 
> We could use a `multierror` here which would be better since it will work with `errors.As`. I wonder what other wrapper steps this won't work with 🤔
> 
> > @aoldershaw should I change this in scope of this pr?
> 
> Up to you - if you wouldn't mind taking on some more scope to fix up the feature generally, we'd really appreciate it! But no pressure

@aoldershaw I will try to take a look on this tomorrow, thanks for helping with this!

kirillbilchenko:
@aoldershaw I pushed some changes, but interesting why it's not working with in parallel if type of step is RetryErrorStep there, I assume it should automatically retry nested step?

aoldershaw:
So there are a few problems with the code on master

1. In `atc/engine/engine.go` (the part that's responsible for actually running the steps), we check if the returned error was of type `exec.Retriable`:

https://github.com/concourse/concourse/blob/587688726f810fd57b985f9f0aae4e1b277fe6f0/atc/engine/engine.go#L243

With nested steps, however, we often wrap the error in another error type, forming a chain of errors. So, the way we were checking before (expecting `err` to be `exec.Retriable`) won't always be true if one of the nested steps is a `RetryErrorStep`, since we're only considering the head of the chain - it could be further down the chain. Changing that line to `errors.As(err, &exec.Retriable{})` actually traverses through the chain of errors and returns true if any of the errors are `exec.Retriable`.

2. The other problem is that we weren't always properly wrapping errors, which breaks the error chain (meaning we lose all the type information of the underlying errors). This is what was happening in the `in_parallel` step - we were just taking the error messages, without maintaining their type information:

https://github.com/concourse/concourse/blob/587688726f810fd57b985f9f0aae4e1b277fe6f0/atc/exec/in_parallel.go#L96

Fortunately, `go-multierror` maintains that information, and allows you to use it with `errors.As`:

https://github.com/hashicorp/go-multierror/blob/master/multierror.go#L54-L80

kirillbilchenko:
@aoldershaw so currently this one should be addressed in this PR I changed to use `go-multierror` the only place I'm not sure I keep `fmt.Errorf("one or more parallel steps errored:\n%w", merr)` this wrapped with new error cuz was not sure if something is linked to this type of error if not we can return just merr

aoldershaw:
I wonder if it makes sense to expand this beyond just `io.EOF` errors, and instead handle any `net.OpError`. I've seen an issue [on Discord](https://discordapp.com/channels/219899946617274369/413770960089382922/738450193862754419) where a Concourse user encountered the following `net.OpError`:

```
Put "/volumes/f30d582e-876b-4d4e-790c-ec1d64b6686f/stream-in?path=.": write tcp 100.98.0.6:43920->100.98.0.6:43925: write: broken pipe
```

that also resulted from workers disappearing, and they were wondering why `--enable-rerun-when-worker-disappears` wasn't working for them

atc/exec/in_parallel.go
```diff
@@ -77,23 +78,23 @@ func (step InParallelStep) Run(ctx context.Context, state RunState) error {
...
 	}
 
 	if ctx.Err() != nil {
 		return ctx.Err()
 	}
 
-	if len(errorMessages) > 0 {
-		return fmt.Errorf("one or more parallel steps errored:\n%s", strings.Join(errorMessages, "\n"))
+	if merr, ok := result.(*multierror.Error); ok {
+		return fmt.Errorf("one or more parallel steps errored:\n%w", merr)
```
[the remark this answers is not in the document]
kirillbilchenko: works with merr I just checked, so I can return it as well, I was not sure because some test may be were connected to this message, will change.

kirillbilchenko:
@aoldershaw will handle all suggestions above

kirillbilchenko:
> I wonder if it makes sense to expand this beyond just `io.EOF` errors, and instead handle any `net.OpError`. I've seen an issue [on Discord](https://discordapp.com/channels/219899946617274369/413770960089382922/738450193862754419) where a Concourse user encountered the following `net.OpError`:
> 
> ```
> Put "/volumes/f30d582e-876b-4d4e-790c-ec1d64b6686f/stream-in?path=.": write tcp 100.98.0.6:43920->100.98.0.6:43925: write: broken pipe
> ```
> 
> that also resulted from workers disappearing, and they were wondering why `--enable-rerun-when-worker-disappears` wasn't working for them

About this not sure, because for my case the most common error was this EOF

kirillbilchenko:
@aoldershaw added net.OpError to be handled in retriable, and test for this

atc/exec/retry_error_step.go
```diff
@@ -48,18 +51,23 @@ func (step RetryErrorStep) Run(ctx context.Context, state RunState) error {
...
+	re := regexp.MustCompile(`worker .+ disappeared`)
+	if ok := errors.As(err, &transportErr) || errors.As(err, &unreachable); ok {
+		logger.Debug("retry-error",
+			lager.Data{"err_type": reflect.TypeOf(err).String(), "err": err.Error()})
+		return true
+	} else if re.MatchString(err.Error()) {
+		logger.Debug("retry-error",
+			lager.Data{"err_type": reflect.TypeOf(err).String(), "err": err})
+		return true
+	} else if errors.Is(err, io.EOF) || errors.As(err, &netOpErr) {
```
aoldershaw, line 66: I think you can get rid of the `io.EOF` check now that we have the check for `*net.OpError` - in your case, the observed root error was an `io.EOF`, but the true root cause was a network issue because the worker went away.

So, just having the check for `*net.OpError` works for your case (where you were seeing `io.EOF`) and for other more general network errors, so you probably don't need both (unless there's some case where you'd observe an `io.EOF` not at the network layer - not sure what that would be, though)

kirillbilchenko: @aoldershaw agree in this case it's network level error so we will cover much more errors there. Removed this EOF and pulled latest master. Thanks again for all suggestions

kirillbilchenko:
@aoldershaw the only concern as I said maybe we need to rename the flag somehow? or it's ok to keep it name as it is?

aoldershaw:
> @aoldershaw the only concern as I said maybe we need to rename the flag somehow? or it's ok to keep it name as it is?

Hmm I'm not sure. I'm not sure how many people are using it, but the flag doesn't look like it was marked as experimental, so I wouldn't want to break their deployments. Also, there may be some cases I haven't considered, but I think network errors in builds will represent `ATC -> worker` errors - so if a network error fails, most likely it's because the worker isn't properly handling requests (probably because it disappeared).

So, I think it's probably okay to keep the flag name the same, at least for now. Maybe we could update the description, though. Something like:

```
Enable automatic build reruns when worker disappears or a network error occurs
```

kirillbilchenko:
@aoldershaw updated the flag description for now

atc/exec/retry_error_step.go
```diff
@@ -48,18 +50,18 @@ func (step RetryErrorStep) Run(ctx context.Context, state RunState) error {
 }
 
 func (step RetryErrorStep) toRetry(logger lager.Logger, err error) bool {
-	switch err.(type) {
-	case transport.WorkerMissingError, transport.WorkerUnreachableError:
+	var transportErr *transport.WorkerMissingError
+	var unreachable *transport.WorkerUnreachableError
+	var netOpErr *net.OpError
+	re := regexp.MustCompile(`worker .+ disappeared`)
+	if ok := errors.As(err, &transportErr) || errors.As(err, &unreachable) || errors.As(err, &netOpErr); ok {
```
aoldershaw, line 57: Rather than handling `*net.OpError`, I think it's better to handle `net.Error` (an interface that `*net.OpError` implements).

Additionally, I ran into a `*url.Error` during my testing, which apparently usually wraps a `net.Error` but not always. So, could you add a case for `*url.Error` as well?

kirillbilchenko:
@aoldershaw I update regarding you comments and for me right now it should work even better, I was aimed to fix only one problem, but we discovered a lot, thanks a lot for effort and help on this.

atc/exec/retry_error_step.go
```diff
@@ -48,18 +50,15 @@ func (step RetryErrorStep) Run(ctx context.Context, state RunState) error {
...
 			lager.Data{"err_type": reflect.TypeOf(err).String(), "err": err.Error()})
 		return true
-	default:
-		re := regexp.MustCompile(`worker .+ disappeared`)
-		if re.MatchString(err.Error()) {
-			logger.Debug("retry-error",
-				lager.Data{"err_type": reflect.TypeOf(err).String(), "err": err})
-			return true
-		}
+	} else if _, ok := err.(net.Error); ok || regexp.MustCompile(`worker .+ disappeared`).MatchString(err.Error()) {
```
aoldershaw, line 58: You should use the same `errors.As` for `net.Error` so it works with nested steps - it works for interfaces, too!

i.e.

```go
var netError net.Error
if errors.As(err, &netError) {
    ...
}
```

See the following demo: https://play.golang.org/p/yHvlZ2V7ApQ (note that `err.(net.Error)` returns false since it's a wrapped error)

kirillbilchenko: Let me try

kirillbilchenko:
@aoldershaw Thanks for babysitting this, fixed comment

PR #5975, opened 2020-08-11
Deflake metrics tests

jamieklassen:

This PR improves the flakiness of the metrics tests!

Now there are no more global variables in the tests (thus no more data races), and the clarity and thread-safety of the periodic tests has been improved. This required some light refactoring in the application code where global variables were being indirectly touched.

## Notes to reviewer:

I found this command helpful for fairly-quickly reproducing the data races and flaky assertions:
```
ginkgo -p -race -untilItFails atc/metric
```
and what was really helpful once I realized there was a deadlock (which actually appears even when the tests are run serially) to get rid of:
```
ginkgo -untilItFails -timeout=3s -v atc/metric
```
which would show the currently-running spec and stack trace when the timeout occurred -- which allowed me to narrow it down to deadlocking `semacquire` goroutines in the counterfeiter fakes.

## Contributor Checklist

- [x] [Signed] all commits
- [x] ~Added release note (Optional)~

## Reviewer Checklist

- [x] Code reviewed
- [x] Tests reviewed
- [ ] PR acceptance performed


atc/metric/emit.go
```diff
@@ -156,11 +160,6 @@ func (m *Monitor) Initialize(logger lager.Logger, host string, attributes map[st
 	return nil
 }
 
-func (m *Monitor) Deinitialize(logger lager.Logger) {
-	close(m.emissions)
```
aoldershaw: Getting rid of this means that the `m.emitLoop()` goroutines will leak in the tests, right? I guess that's fine, though

atc/metric/emit.go
```diff
@@ -36,9 +37,60 @@ type Monitor struct {
...
+	JobsScheduled:              Counter{},
+	JobsScheduling:             Gauge{},
+	BuildsStarted:              Counter{},
+	BuildsRunning:              Gauge{},
+	TasksWaiting:               Gauge{},
+	ChecksFinishedWithError:    Counter{},
+	ChecksFinishedWithSuccess:  Counter{},
+	ChecksQueueSize:            Gauge{},
+	ChecksStarted:              Counter{},
+	ChecksEnqueued:             Counter{},
```
aoldershaw: All of these can be omitted, since it's just their zero-values - is there a reason you explicitly instantiate them?

atc/metric/emit.go
jamieklassen: yup. that's a leak i'm willing to take on in order to pay down the test-inflicted damage of the extra method. we could actually move the concurrency part out of the metrics subsystem completely, and have the atccmd create the emitloop goroutine, which would be kinda cool. We'd then be responsible for calling `emitLoop` (or whatever public name it ends up getting) in tests, but I'd be OK with that since that would also make the tests fully synchronous.

I might take a stab at that another time.

atc/metric/emit.go
jamieklassen: copy-paste laziness! good catch

PR #5987, opened 2020-08-13
Prevent potential deadlock after pipeline deletion

### LINKED ISSUE #5287: Destroying a pipeline can get stuck
## Bug Report

When we tried to destroy a pipeline on our large Concourse deployment, there are times when the query to delete the pipeline in the database gets stuck. We saw two queries that have been running for 2 hours, one of them was the delete pipeline query

https://github.com/concourse/concourse/blob/a7716ae103636cdb15fbcfaca1211fd04d1cd749/atc/db/pipeline.go#L721-L726

and the other was one to delete build events https://github.com/concourse/concourse/blob/1756c865ae837c4141be4c0df98685c5ec7daf0e/atc/db/pipeline.go#L910-L911

We theorized that the two queries might have been deadlocked on each other, but that raises some suspicion because this normally would throw a deadlock error similar to the issue outlined in #3383.

## Expected Results

Deleting a pipeline does not get stuck.

## Actual Results

Deleting the pipeline gets stuck!

## Additional Context

As a result of this, the database CPU spiked to 100% and the API connection pool filled up. Because the API connection pool filled up, users would see slowness in the web UI and workers would stall because it is unable to heartbeat.

## Version Info

* Concourse version: 6.0
* Deployment type (BOSH/Docker/binary):
* Infrastructure/IaaS:
* Browser (if applicable):
* Did this used to work?

aoldershaw:

closes #5287  .

* Remove `pipeline_build_events_delete_trigger`, which previously dropped the `pipeline_build_events_x` table when the corresponding pipeline was deleted
  * Instead, deleting a pipeline marks it as deleted in a new `deleted_pipelines` table
* Modify the build log collector to drop the `pipeline_build_events_x` table for each pipeline that is marked as deleted in `deleted_pipelines`

## Notes to reviewer:

The motivation for this change is elaborated more in https://github.com/concourse/concourse/issues/5287#issuecomment-673472967, but essentially, having only one place where build events are deleted removes the risk of deadlocking here.

An alternative approach we considered was to compare the list of all (non-deleted) pipelines to the list of `pipeline_build_events_x` tables that are still around to determine which to delete. This would mean we wouldn't have to maintain the `deleted_pipelines` table. We went with the `deleted_pipelines` approach because it enables us to make decisions about *when* to remove pipeline build events for deleted pipelines. For instance, an operator may wish to keep build events around for 1 week after a pipeline is deleted. We don't currently support this, but it's a possibility with the chosen approach. Also, it kind of felt weird querying the list of tables in Postgres. Happy to reconsider the approach taken, though.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/pipeline_lifecycle.go
+++ b/atc/db/pipeline_lifecycle.go
@@ -110,6 +110,10 @@ func (p *pipelineLifecycle) RemoveBuildEventsForDeletedPipelines() error {
 
 	rows.Close()
 
+	if len(idsToDelete) == 0 {
+		return nil
+	}
+
 	for _, id := range idsToDelete {
 		_, err = p.conn.Exec(fmt.Sprintf("DROP TABLE IF EXISTS pipeline_build_events_%d", id))
 		if err != nil {
--- a/atc/db/pipeline_lifecycle.go
+++ b/atc/db/pipeline_lifecycle.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"database/sql"
+	"fmt"
 
 	sq "github.com/Masterminds/squirrel"
 	"github.com/concourse/concourse/atc/db/lock"
@@ -11,6 +12,7 @@ import (
 
 type PipelineLifecycle interface {
 	ArchiveAbandonedPipelines() error
+	RemoveBuildEventsForDeletedPipelines() error
 }
 
 func NewPipelineLifecycle(conn Conn, lockFactory lock.LockFactory) PipelineLifecycle {
@@ -83,6 +85,44 @@ func archivePipelines(tx Tx, conn Conn, lockFactory lock.LockFactory, rows *sql.
 		if err != nil {
 			return err
 		}
+	}
+
+	return nil
+}
+
+func (p *pipelineLifecycle) RemoveBuildEventsForDeletedPipelines() error {
+	rows, err := psql.Select("id").
+		From("deleted_pipelines").
+		RunWith(p.conn).
+		Query()
+	if err != nil {
+		return err
+	}
+
+	var idsToDelete []int
+	for rows.Next() {
+		var id int
+		if err := rows.Scan(&id); err != nil {
+			return err
+		}
+		idsToDelete = append(idsToDelete, id)
+	}
+
+	rows.Close()
+
+	for _, id := range idsToDelete {
+		_, err = p.conn.Exec(fmt.Sprintf("DROP TABLE IF EXISTS pipeline_build_events_%d", id))
+		if err != nil {
+			return err
+		}
+	}
+
+	_, err = psql.Delete("deleted_pipelines").
+		Where(sq.Eq{"id": idsToDelete}).
+		RunWith(p.conn).
+		Exec()
+	if err != nil {
+		return err
 	}
 
 	return nil
--- a/atc/db/pipeline_lifecycle_test.go
+++ b/atc/db/pipeline_lifecycle_test.go
@@ -91,8 +91,10 @@ var _ = Describe("PipelineLifecycle", func() {
 		)
 
 		BeforeEach(func() {
-			pipeline1 = defaultPipeline
-			pipeline2, _, _ = defaultTeam.SavePipeline("pipeline2", defaultPipelineConfig, 0, false)
+			pipeline1, _, err = defaultTeam.SavePipeline("pipeline1", defaultPipelineConfig, 0, false)
+			Expect(err).ToNot(HaveOccurred())
+			pipeline2, _, err = defaultTeam.SavePipeline("pipeline2", defaultPipelineConfig, 0, false)
+			Expect(err).ToNot(HaveOccurred())
 		})
 
 		pipelineBuildEventsExists := func(id int) bool {
@@ -104,8 +106,8 @@ var _ = Describe("PipelineLifecycle", func() {
 		}
 
 		It("drops the pipeline_build_events_x table for each deleted pipeline", func() {
-			pipeline1.Destroy()
-			pipeline2.Destroy()
+			destroy(pipeline1)
+			destroy(pipeline2)
 
 			err := pl.RemoveBuildEventsForDeletedPipelines()
 			Expect(err).ToNot(HaveOccurred())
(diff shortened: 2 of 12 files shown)


atc/db/pipeline_lifecycle_test.go
```diff
@@ -10,70 +12,124 @@ import (
...
-			defaultPipeline.Reload()
-			Expect(defaultPipeline.Archived()).To(BeFalse())
+	Describe("RemoveBuildEventsForDeletedPipelines", func() {
+		var (
+			pipeline1 db.Pipeline
+			pipeline2 db.Pipeline
+		)
+
+		BeforeEach(func() {
+			pipeline1 = defaultPipeline
```
chenbh, line 94: I was confused why we didn't save `pipeline1` until I realized `defaultPipeline` was saved in the suite setup. Maybe we should use it directly instead of renaming it (e.g. `defaultPipeline.Destroy()` down below)

atc/db/pipeline_lifecycle_test.go
```diff
@@ -10,70 +12,124 @@ import (
...
+			err := pl.RemoveBuildEventsForDeletedPipelines()
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(pipelineBuildEventsExists(pipeline1.ID())).To(BeFalse())
+			Expect(pipelineBuildEventsExists(pipeline2.ID())).To(BeFalse())
+		})
+
+		It("clears the deleted_pipelines table", func() {
+			pipeline1.Destroy()
+			pl.RemoveBuildEventsForDeletedPipelines()
```
chenbh, line 119:
```suggestion
			err := pl.RemoveBuildEventsForDeletedPipelines()
			Expect(err).ToNot(HaveOccurred())
```

atc/db/pipeline_test.go
```diff
@@ -1204,6 +1205,15 @@ var _ = Describe("Pipeline", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(found).To(BeFalse())
 		})
+
+		It("marks the pipeline ID in the deleted_pipelines table", func() {
+			pipeline.Destroy()
```
chenbh, line 1210: Just to be extra safe

```suggestion
			err := pipeline.Destroy()
			Expect(err).ToNot(HaveOccurred())
```

atc/db/pipeline_lifecycle.go
```diff
@@ -87,3 +89,41 @@ func archivePipelines(tx Tx, conn Conn, lockFactory lock.LockFactory, rows *sql.
...
+		if err := rows.Scan(&id); err != nil {
+			return err
+		}
+		idsToDelete = append(idsToDelete, id)
+	}
+
+	rows.Close()
+
+	for _, id := range idsToDelete {
+		_, err = p.conn.Exec(fmt.Sprintf("DROP TABLE IF EXISTS pipeline_build_events_%d", id))
```
clarafu, line 114: I think it's a great idea to not have all the queries run in a transaction, nice!

atc/db/pipeline_lifecycle_test.go
```diff
@@ -10,70 +12,124 @@ import (
...
+		pipelineBuildEventsExists := func(id int) bool {
+			var exists bool
+			err = dbConn.QueryRow(fmt.Sprintf("SELECT EXISTS (SELECT 1 FROM information_schema.tables WHERE table_name = 'pipeline_build_events_%d')", id)).Scan(&exists)
+			Expect(err).ToNot(HaveOccurred())
+
+			return exists
+		}
+
+		It("drops the pipeline_build_events_x table for each deleted pipeline", func() {
+			pipeline1.Destroy()
```
clarafu, line 107: nit: It's not a big deal but I might suggest to check the error for destroying the pipeline? 🤔  Just to make sure that it didn't error and that is actually what is causing `RemoveBuildEventsForDeletedPipelines` to not remove the build events for that pipeline.

aoldershaw: I guess what I'm struggling with is - what would cause `pipeline1.Destroy()` to fail? I might be wrong, but I think something would have to be pretty borked with the postgres runner for this to error, in which case we'd likely get a whole bunch of errors in other test cases, anyway.

IMO having the expectations focused on what the test case is actually testing is typically more valuable than erroring early, at least when the error is unlikely to happen/would affect a bunch of other tests. Feel free to challenge me on this - you've worked on this part of the codebase far longer than I have, and it's definitely possible that having more `Expect(err).ToNot(HaveOccurred())`s sprinkled around have made tests more robust/easy to debug for you.

cc @chenbh

atc/db/pipeline_lifecycle_test.go
aoldershaw: Fair, the asymmetric setup is a little weird. I'd rather keep the consistent naming of `pipeline1`/`pipeline2` though, rather than `defaultPipeline`/`pipeline2`, so I'll just save a new pipeline for `pipeline1` as well

atc/db/pipeline_lifecycle_test.go
aoldershaw: Going to push changes to add in the error checks since it's pretty minor anyway and I'd ideally like to get this in for 6.5. Found a decent compromise in adding a `destroy` helper that calls `pipeline.Destroy()` and ensures no error occurred. That way, at least the tests can remain easy to read while still giving us more helpful errors if an unexpected error occurs.

clarafu: Hm I don't think I have a very strong opinion on it, I've just seen times within db tests where because there was an error check missing it ended up taking much longer to figure out what went wrong. I think it might make your tests a bit more resilient to refactors because even though right now it is a simple `delete` query, it can be possible that someone in the future adds more logic within `pipeline.Destroy()` where an error can be thrown outside of the `delete` query.  But that being said, I really don't have a strong preference because I can also understand it taking away from the focus of the test. I think it's really up to you haha whatever you think is best! I just wanted to point it out just in case.

PR #5995, opened 2020-08-17
atc: Allow OPA integration to return reasons for failures (breaking change)

andy-paine:
Feature

Allow OPA documents to include an `allowed` and a `reasons` field that
gives a bit more feedback to the user about why something failed. This
allows OPA operators to write rules that are self explanatory as to why
things are getting denied.

Fly does not appear to print the full error message currently but this
works for the UI.

## Notes to reviewer:
This represents a breaking change to the OPA API (if you can call it that) - the document returned will no longer just be a boolean and will instead be objects (though the reasons are optional, it will need to be an object of at least `{ "allowed": true/false }`

I figured whilst this feature was new was probably the time to break the contract between Concourse and OPA though rather than later when it has general opt-in 😁

I'm also happy to write some documentation for the OPA feature at some point too 😄 

I used the following Rego:
```
package concourse

decision = {"allowed": true} {
  count(deny) == 0
}

decision = {"allowed": false, "reasons": reasons} {
  count(deny) > 0
  reasons := deny
}

deny["cannot use busybox"] {
  input.action = "UseImage"
  input.data.image_source.repository == "busybox"
}

deny["cannot use docker-image types"] {
  input.action = "UseImage"
  input.data.image_type == "docker-image"
}
```

Which in the UI produces the following output when trying to run a task with the `busybox` image from a `docker-image` resource:
Screenshot 2020-08-18 at 00 11 21

## Release Note
Operators using the OPA integration can now return reasons why a check failed. This is a breaking change and existing policies will need to be re-written to return an object `{}` containing an `allowed` and `reasons` fields. `allowed` is the original boolean value and `reasons` is an array of strings.

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/policy/opa/opa.go
+++ b/atc/policy/opa/opa.go
@@ -33,8 +33,13 @@ type opaInput struct {
 	Input policy.PolicyCheckInput `json:"input"`
 }
 
+type opaOuptut struct {
+	Allowed *bool    `json:"allowed,omitempty"`
+	Reasons []string `json:"reasons,omitempty"`
+}
+
 type opaResult struct {
-	Result *bool `json:"result,omitempty"`
+	Result *opaOuptut `json:"result,omitempty"`
 }
 
 type opa struct {
@@ -42,18 +47,18 @@ type opa struct {
 	logger lager.Logger
 }
 
-func (c opa) Check(input policy.PolicyCheckInput) (bool, error) {
+func (c opa) Check(input policy.PolicyCheckInput) (policy.PolicyCheckOutput, error) {
 	data := opaInput{input}
 	jsonBytes, err := json.Marshal(data)
 	if err != nil {
-		return false, err
+		return policy.FailedPolicyCheck(), nil
 	}
 
 	c.logger.Debug("opa-check", lager.Data{"input": string(jsonBytes)})
 
 	req, err := http.NewRequest("POST", c.config.URL, bytes.NewBuffer(jsonBytes))
 	if err != nil {
-		return false, err
+		return policy.FailedPolicyCheck(), nil
 	}
 	req.Header.Set("Content-Type", "application/json")
 
@@ -61,31 +66,34 @@ func (c opa) Check(input policy.PolicyCheckInput) (bool, error) {
 	client.Timeout = c.config.Timeout
 	resp, err := client.Do(req)
 	if err != nil {
-		return false, err
+		return policy.FailedPolicyCheck(), err
 	}
 	defer resp.Body.Close()
 
 	statusCode := resp.StatusCode
 	if statusCode != http.StatusOK {
-		return false, fmt.Errorf("opa returned status: %d", statusCode)
+		return policy.FailedPolicyCheck(), fmt.Errorf("opa returned status: %d", statusCode)
 	}
 
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
-		return false, fmt.Errorf("opa returned no response: %s", err.Error())
+		return policy.FailedPolicyCheck(), fmt.Errorf("opa returned no response: %s", err.Error())
 	}
 
 	result := &opaResult{}
 	err = json.Unmarshal(body, &result)
 	if err != nil {
-		return false, fmt.Errorf("opa returned bad response: %s", err.Error())
+		return policy.FailedPolicyCheck(), fmt.Errorf("opa returned bad response: %s", err.Error())
 	}
 
 	// If no result returned, meaning that the requested policy decision is
 	// undefined OPA, then consider as pass.
 	if result.Result == nil {
-		return true, nil
+		return policy.PassedPolicyCheck(), nil
 	}
 
-	return *result.Result, nil
+	return policy.PolicyCheckOutput{
+		Allowed: *result.Result.Allowed,
+		Reasons: result.Result.Reasons,
+	}, nil
 }
--- a/atc/policy/opa/opa.go
+++ b/atc/policy/opa/opa.go
@@ -51,14 +51,14 @@ func (c opa) Check(input policy.PolicyCheckInput) (policy.PolicyCheckOutput, err
 	data := opaInput{input}
 	jsonBytes, err := json.Marshal(data)
 	if err != nil {
-		return policy.FailedPolicyCheck(), nil
+		return policy.FailedPolicyCheck(), err
 	}
 
 	c.logger.Debug("opa-check", lager.Data{"input": string(jsonBytes)})
 
 	req, err := http.NewRequest("POST", c.config.URL, bytes.NewBuffer(jsonBytes))
 	if err != nil {
-		return policy.FailedPolicyCheck(), nil
+		return policy.FailedPolicyCheck(), err
 	}
 	req.Header.Set("Content-Type", "application/json")
 
@@ -86,10 +86,8 @@ func (c opa) Check(input policy.PolicyCheckInput) (policy.PolicyCheckOutput, err
 		return policy.FailedPolicyCheck(), fmt.Errorf("opa returned bad response: %s", err.Error())
 	}
 
-	// If no result returned, meaning that the requested policy decision is
-	// undefined OPA, then consider as pass.
-	if result.Result == nil {
-		return policy.PassedPolicyCheck(), nil
+	if result.Result == nil || result.Result.Allowed == nil {
+		return policy.FailedPolicyCheck(), fmt.Errorf("opa returned invalid response: %s", body)
 	}
 
 	return policy.PolicyCheckOutput{
--- a/hack/opa/policy.rego
+++ b/hack/opa/policy.rego
@@ -1,8 +1,24 @@
 package concourse
 
-# replace with 'false' to add rules
-default allow = true
+default decision = {"allowed": true}
 
-# allow {
-#   input.action == "ListContainers"
-# }
+# uncomment to include deny rules
+#decision = {"allowed": false, "reasons": reasons} {
+#  count(deny) > 0
+#  reasons := deny
+#}
+
+deny["cannot use docker-image types"] {
+  input.action == "UseImage"
+  input.data.image_type == "docker-image"
+}
+
+deny["cannot run privileged tasks"] {
+  input.action == "SaveConfig"
+  input.data.jobs[_].plan[_].privileged
+}
+
+deny["cannot use privileged resource types"] {
+  input.action == "SaveConfig"
+  input.data.resource_types[_].privileged
+}
--- a/atc/api/api_suite_test.go
+++ b/atc/api/api_suite_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/atc/db/dbfakes"
 	"github.com/concourse/concourse/atc/gc/gcfakes"
+	"github.com/concourse/concourse/atc/policy"
 	"github.com/concourse/concourse/atc/worker/workerfakes"
 	"github.com/concourse/concourse/atc/wrappa"
 
@@ -159,7 +160,7 @@ var _ = BeforeEach(func() {
 	checkWorkerTeamAccessHandlerFactory := auth.NewCheckWorkerTeamAccessHandlerFactory(dbWorkerFactory)
 
 	fakePolicyChecker = new(policycheckerfakes.FakePolicyChecker)
-	fakePolicyChecker.CheckReturns(true, nil)
+	fakePolicyChecker.CheckReturns(policy.PassedPolicyCheck(), nil)
 
 	apiWrapper := wrappa.MultiWrappa{
 		wrappa.NewPolicyCheckWrappa(logger, fakePolicyChecker),
--- a/atc/api/policychecker/checker.go
+++ b/atc/api/policychecker/checker.go
@@ -2,10 +2,11 @@ package policychecker
 
 import (
 	"bytes"
+	"encoding/json"
 	"io/ioutil"
 	"net/http"
+
 	"sigs.k8s.io/yaml"
-	"encoding/json"
 
 	"github.com/concourse/concourse/atc/api/accessor"
 	"github.com/concourse/concourse/atc/policy"
@@ -14,7 +15,7 @@ import (
 //go:generate counterfeiter . PolicyChecker
 
 type PolicyChecker interface {
-	Check(string, accessor.Access, *http.Request) (bool, error)
+	Check(string, accessor.Access, *http.Request) (policy.PolicyCheckOutput, error)
 }
 
 type checker struct {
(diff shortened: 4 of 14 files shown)


evanchaoli:
@andy-paine Thanks for the enhancement. Actually when I initially worked on the OPA integration feature, I wanted to provide reasons when policy check fails. However I didn't find how OPA could provide reasons. I just checked OPA doc again: https://www.openpolicyagent.org/docs/latest/integration/#integrating-with-the-rest-api, it still doesn't show OPA API would return `allows` and `reasons`. In your code change, you don't modify the API request, but just modify JSON unmarshal. Can you please point out any OPA doc that supports your change?

andy-paine:
So we call the `data` API for OPA (https://www.openpolicyagent.org/docs/latest/rest-api/#get-a-document-with-input) which will return a response containing `result` which includes whatever document is resolved for the query supplied by the URL path.

So if Concourse expects OPA to return a result which includes both an `allowed` flag and an optional `reasons` (or others going forwards), then authors just need to make sure that they return an object of that format. It is breaking because anyone using existing rules will need to modify them to resolve to `{ "allowed": true/false }` rather than `true/false`

I got the idea from https://github.com/open-policy-agent/opa/issues/1148

This playground shows the structure of the `decision` object (name of the document can be anything since it just forms the path of the `OPA_URL`) https://play.openpolicyagent.org/p/algBZBPdgN

andy-paine:
Woops! One of the integration tests causes my machine to hang so I can't run the entire test suite. I've updated (and added an extra test) and I _think_ I've run all the relevant ones now.

andy-paine:
I've updated the `hack` stuff - you'll need to go uncomment a bit if you want the rules to actually apply though. I included some rules that I think people may actually use 😄

atc/policy/opa/opa.go
```diff
@@ -33,59 +33,67 @@ type opaInput struct {
...
-		return false, fmt.Errorf("opa returned bad response: %s", err.Error())
+		return policy.FailedPolicyCheck(), fmt.Errorf("opa returned bad response: %s", err.Error())
 	}
 
 	// If no result returned, meaning that the requested policy decision is
 	// undefined OPA, then consider as pass.
 	if result.Result == nil {
-		return true, nil
+		return policy.PassedPolicyCheck(), nil
 	}
```
taylorsilva, line 93: If someone writes a policy that only returns `reasons` users get an EOF error via fly or no message at all on the build page. We find the following in the Concourse logs
```
2020/08/20 14:56:03 http: panic serving 172.24.0.1:58766: runtime error: invalid memory address or nil pointer dereference
goroutine 196255 [running]:
net/http.(*conn).serve.func1(0xc0007b52c0)
	/usr/local/go/src/net/http/server.go:1800 +0x149
panic(0x34c6ce0, 0x5105c80)
	/usr/local/go/src/runtime/panic.go:975 +0x41b
github.com/concourse/concourse/atc/policy/opa.opa.Check(0xc000056012, 0x2a, 0x12a05f200, 0x3e85d20, 0xc0007987e0, 0x37e7ad1, 0x9, 0xc0000520b7, 0x3, 0x37e6c70, ...)
	/src/atc/policy/opa/opa.go:96 +0xee8
github.com/concourse/concourse/atc/policy.(*Checker).Check(0xc000798900, 0x37e7ad1, 0x9, 0xc0000520b7, 0x3, 0x37e6c70, 0x9, 0xc0005d4b80, 0x3, 0x37eb46b, ...)
	/src/atc/policy/checker.go:166 +0x11b
github.com/concourse/concourse/atc/api/policychecker.(*checker).Check(0xc0007f6038, 0x37eb46b, 0xb, 0x3e85ea0, 0xc000bc8230, 0xc000945500, 0xc000824600, 0x0, 0x0, 0x0, ...)
	/src/atc/api/policychecker/checker.go:77 +0xc6b
github.com/concourse/concourse/atc/api/policychecker.policyCheckingHandler.ServeHTTP(0x3e85d20, 0xc0003fcf60, 0x3e11300, 0xc00085c210, 0x37eb46b, 0xb, 0x3e0c020, 0xc0007f6038, 0x3e67fa0, 0xc000bc81c0, ...)
	/src/atc/api/policychecker/handler.go:38 +0xf4
github.com/concourse/concourse/atc/api/auth.checkAuthHandler.ServeHTTP(0x3e10f00, 0xc0003bc8c0, 0x3e2f0a0, 0x527beb0, 0x3e67fa0, 0xc000bc81c0, 0xc000945500)
	/src/atc/api/auth/check_authentication_handler.go:28 +0x92
github.com/concourse/concourse/atc/wrappa.VersionedHandler.ServeHTTP(0x37e6c70, 0x9, 0x3e10dc0, 0xc0003a7700, 0x3e67fa0, 0xc000bc81c0, 0xc000945500)
	/src/atc/wrappa/versioned_handler.go:14 +0xa7
github.com/concourse/concourse/atc/api/accessor.(*accessorHandler).ServeHTTP(0xc0008115c0, 0x3e67fa0, 0xc000bc81c0, 0xc000945400)
	/src/atc/api/accessor/handler.go:64 +0x478
github.com/NYTimes/gziphandler.GzipHandlerWithOpts.func1.1(0x3e67b20, 0xc0006670a0, 0xc000945400)
	/go/pkg/mod/github.com/!n!y!times/gziphandler@v1.1.1/gzip.go:336 +0x2ba
net/http.HandlerFunc.ServeHTTP(0xc000871290, 0x3e67b20, 0xc0006670a0, 0xc000945400)
	/usr/local/go/src/net/http/server.go:2041 +0x44
github.com/bmizerany/pat.(*PatternServeMux).ServeHTTP(0xc0003c0c60, 0x3e67b20, 0xc0006670a0, 0xc000945400)
	/go/pkg/mod/github.com/bmizerany/pat@v0.0.0-20170815010413-6226ea591a40/mux.go:117 +0x313
github.com/concourse/concourse/atc/api/auth.csrfValidationHandler.ServeHTTP(0x3e0bbc0, 0xc0003c0c60, 0x3e8c800, 0xc000358ba8, 0x3e67b20, 0xc0006670a0, 0xc000945400)
	/src/atc/api/auth/csrf_validation_handler.go:60 +0x60b
net/http.(*ServeMux).ServeHTTP(0xc00087d6c0, 0x3e67b20, 0xc0006670a0, 0xc000945400)
	/usr/local/go/src/net/http/server.go:2416 +0x1a6
github.com/concourse/concourse/atc/api/auth.WebAuthHandler.ServeHTTP(0x3e0fa40, 0xc00087d6c0, 0x3e8c800, 0xc000358ba8, 0x3e67b20, 0xc0006670a0, 0xc000945400)
	/src/atc/api/auth/web_auth_handler.go:55 +0x300
github.com/concourse/concourse/atc/wrappa.SecurityHandler.ServeHTTP(0x35031d2, 0x4, 0x3e10d80, 0xc0007b3a60, 0x3e67b20, 0xc0006670a0, 0xc000945400)
	/src/atc/wrappa/security_handler.go:17 +0x1af
github.com/concourse/concourse/atc/wrappa.LoggerHandler.ServeHTTP(0x3e85d20, 0xc0003fcf60, 0x3e11520, 0xc0007b3a80, 0x3e67b20, 0xc0006670a0, 0xc0007e9700)
	/src/atc/wrappa/logger_handler.go:20 +0x2fb
net/http.serverHandler.ServeHTTP(0xc0001ca380, 0x3e67b20, 0xc0006670a0, 0xc0007e9700)
	/usr/local/go/src/net/http/server.go:2836 +0x20f
net/http.(*conn).serve(0xc0007b52c0, 0x3e6fda0, 0xc000403080)
	/usr/local/go/src/net/http/server.go:1924 +0x173a
created by net/http.(*Server).Serve
	/usr/local/go/src/net/http/server.go:2962 +0x931
```

This is definitely an edge-case, curious what your thoughts are? I was thinking we could add a case around here to check that `result.Result.Allowed == nil` as well? I think that was the original intention of this `if`-case.

taylorsilva: Looking at the original code:
```go
	type opaResult struct {
		Result *bool `json:"result,omitempty"`
	}
...
	if result.Result == nil {
		return policy.PassedPolicyCheck(), nil
	}

```
The intention was definitely to check if a bool was passed back. I think we should still do this check. The `if` in it's current state does not do this. I think it will always pass if `json.Unmarshal` was successful.

hack/opa/policy.rego
```diff
@@ -1,8 +1,24 @@
 package concourse
 
-# replace with 'false' to add rules
-default allow = true
+default decision = {"allowed": true}
```
taylorsilva, line 3: I really like this, I think it's a good starter example for people to work from!

atc/policy/opa/opa.go
```diff
@@ -33,59 +33,67 @@ type opaInput struct {
...
 	logger lager.Logger
 }
 
-func (c opa) Check(input policy.PolicyCheckInput) (bool, error) {
+func (c opa) Check(input policy.PolicyCheckInput) (policy.PolicyCheckOutput, error) {
 	data := opaInput{input}
 	jsonBytes, err := json.Marshal(data)
 	if err != nil {
-		return false, err
+		return policy.FailedPolicyCheck(), nil
```
andy-paine, line 54: This should be `err` instead of `nil`

atc/policy/opa/opa.go
```diff
@@ -33,59 +33,67 @@ type opaInput struct {
...
 	}
 
 	c.logger.Debug("opa-check", lager.Data{"input": string(jsonBytes)})
 
 	req, err := http.NewRequest("POST", c.config.URL, bytes.NewBuffer(jsonBytes))
 	if err != nil {
-		return false, err
+		return policy.FailedPolicyCheck(), nil
```
andy-paine, line 61: This should be `err` instead of `nil`

atc/policy/opa/opa.go
andy-paine: Yeah that makes sense. I thought that a `bool` would deserialize to a default of `false` as it does in some other languages but if it ends up as `nil` then we should probably check for that.

I'm unsure of what the standard is for OPA integrations about how strict the "API" of what is returned should be. Personally, I think that `result.Result` being `nil` as well as the response not containing an `allowed` field should also be treated as a failed check. If a check has been sent to a policy checker, that means an operator has specifically configured Concourse to do so for that action & http method so the policy checker (OPA) should *specifically* say that something is allowed rather than Concourse assuming things are allowed by default.

It's easy enough to put `default decision = { "allowed": "true" }` in your Rego and it prevents the potential foot-cannon of you configuring OPA wrongly and it just returns `{}` for everything which Concourse then inteprets as allowed due to that being the default. In this situation I think returning something like:
```
return policy.FailedPolicyCheck(), fmt.Errorf("opa returned invalid response: %+v", result.Result)
```
may be a more defensive approach. I can easily add some additional tests to cover this if we think that this approach is worthwhile.

andy-paine:
> Noting why this is a breaking change; this change results in the following error appearing in Concourse when using OPA rules written for the current implementation, which only returns a bool. You can use the current `policy.rego` on master to see this error on the build page:
> 
> ```
> opa returned bad response: json: cannot unmarshal bool into Go struct field opaResult.result of type opa.opaOuptut
> ```

I agree this isn't the prettiest but I figured since a) it is a brand new feature and people using it are opting-in so likely know where to go looking for OPA failures and b) it isn't a great user message but it does describe the problem properly!

atc/policy/opa/opa.go
taylorsilva: > I thought that a bool would deserialize to a default of false 

It does in Go, but not in this case because the bool is actually a pointer to a bool. If it was just a `bool` instead of `*bool` then you'd get the expected result of `result.Result == false`. I'm guessing this was done in case OPA returned `{}`. 

I think it's worth seeing if we can turn this pointer into a bool instead. I think it makes sense to have it as a pointer if we were going to `json.Marshal(opaOutput)` but I don't think we ever will. @evanchaoli is there a case/reason for keeping `opaOutput.Allowed` as a `*bool` instead of a `bool`?

> Personally, I think that result.Result being nil as well as the response not containing an allowed field should also be treated as a failed check

I agree with this 👍 I'd be happy to go with that approach, I think it's safer. I checked the initial OPA PR and didn't see this mentioned anywhere.

andy-paine: Actually if we made it a `bool` rather than `*bool` we'll not be able to differentiate between "didn't supply allowed and defaulted to false"  and "allowed was false" so might make it harder to alert an operator about what was wrong with their rules.

Thinking about the user-facing side of this - the error here is provided back to the user verbatim. Would it be preferable return an error just saying `opa returned invalid response` and log the actual response we got at INFO level or is it okay to display errors messages like `opa returned invalid response: &{Allowed:<nil> Reasons:[]}` to users?

taylorsilva: I think the latter would be fine, showing the full error to the user. It would not be helpful for users but it will be helpful for operators. Especially if the user reports the error to the operator, the operator doesn't have the take the extra step of digging through logs.

atc/policy/opa/opa.go
```diff
@@ -86,10 +86,8 @@ func (c opa) Check(input policy.PolicyCheckInput) (policy.PolicyCheckOutput, err
 		return policy.FailedPolicyCheck(), fmt.Errorf("opa returned bad response: %s", err.Error())
 	}
 
-	// If no result returned, meaning that the requested policy decision is
-	// undefined OPA, then consider as pass.
-	if result.Result == nil {
-		return policy.PassedPolicyCheck(), nil
+	if result.Result == nil || result.Result.Allowed == nil {
+		return policy.FailedPolicyCheck(), fmt.Errorf("opa returned invalid response: %s", resp.Body)
```
taylorsilva, line 90: I got a really intelligible response back by creating a rule that looked like this:

```rego
default decision = {"reasons": ["hey there"]}
``` 
In my build page I got this, which looks like the struct of an `io.ReadCloser`, which is what `Response.Body` is.
```
opa returned invalid response: &{%!s(func()=0xf65060) %!s(*http.bodyEOFSignal=&{0xc000a3d140 {0 0} false 0xc000118100 <nil> 0x15ff940}) %!s(func() bool=0x15f82a0)}
```

I changed it to return the `body` var on line 78
```suggestion
		return policy.FailedPolicyCheck(), fmt.Errorf("opa returned invalid response: %s", body)
```
Which gives a more human readable message:
```
opa returned invalid response: {"result":{"reasons":["hey there"]}}
```

andy-paine:
Ah good spot 👍 I've amended and force pushed rather than committing suggestions just to keep the git history a little cleaner. I've updated the release note in the PR description too. Documentation-wise there isn't really any for OPA at the moment but I will probably take a crack at it after this gets merged.

PR #5996, opened 2020-08-18
Return real JSON from "fly get-team --json"

agurney:

Bug Fix

This is documented, implemented-ish, and tested-ish, but didn't work: doing `fly -t tgt get-team --team-name foo --json` would always produce the syntactically-valid but boring response `{}`. This is because the code uses the go-concourse `Team` object, which has no data members and so does very little of interest when serialized.

There was originally a test for the YAML-formatted output, but that did not run for the reasons described in 2e5d954, and there were no tests for JSON. That commit also changed the test to look at the tabular format. It appears tests of the JSON path were accidentally skipped over in the refactoring.

We now synthesize a JSON object with team and auth contents, plus the team ID (originally discarded in `go-concourse` but now stashed for this purpose). That means the output will match the elements in `fly teams`.

## Notes to reviewer:

I added `.ID()` to the go-concourse `Team`, as a quick way to get this information through. In `get_team.go` I just make a little map with the desired data, which happens to mirror ATC's definition of `Team`. I feel a bit bad about not just using `atc.Team` properly, but was wary of getting caught up in too much messing about with refactoring of the different `Team`/`team` objects. Happy to give a stab at that if it seems worthwhile.

The changes in go-concourse/concourse/concoursefakes/fake_team.go are just whatever `go generate` did, and should be completely boilerplate.

## Release Note
The `--json` flag on `fly get-team` is meant to yield a structured JSON representation of the team data, including auth details, but instead just gave `{}`. Now it works as intended.

## Contributor Checklist

- [X] [Signed] all commits
- [ ] Updated [Documentation] - I didn't change this because it's already documented to work this way
- [X] Added release note (Optional)

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/go-concourse/concourse/team.go
+++ b/go-concourse/concourse/team.go
@@ -11,7 +11,7 @@ import (
 
 type Team interface {
 	Name() string
-
+    ID() int
 	Auth() atc.TeamAuth
 
 	CreateOrUpdate(team atc.Team) (atc.Team, bool, bool, []ConfigWarning, error)
@@ -77,6 +77,7 @@ type Team interface {
 
 type team struct {
 	name       string
+	id         int
 	connection internal.Connection //Deprecated
 	httpAgent  internal.HTTPAgent
 	auth       atc.TeamAuth
@@ -86,13 +87,17 @@ func (team *team) Name() string {
 	return team.name
 }
 
+func (team *team) ID() int {
+	return team.id
+}
+
+func (team *team) Auth() atc.TeamAuth {
+	return team.auth
+}
+
 func (client *client) Team(name string) Team {
 	return &team{
 		name:       name,
 		connection: client.connection,
 	}
-}
-
-func (team *team) Auth() atc.TeamAuth {
-	return team.auth
 }
--- a/go-concourse/concourse/team.go
+++ b/go-concourse/concourse/team.go
@@ -13,6 +13,7 @@ type Team interface {
 	Name() string
     ID() int
 	Auth() atc.TeamAuth
+	ATCTeam() atc.Team
 
 	CreateOrUpdate(team atc.Team) (atc.Team, bool, bool, []ConfigWarning, error)
 	RenameTeam(teamName, name string) (bool, []ConfigWarning, error)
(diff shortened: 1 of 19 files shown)


go-concourse/concourse/team.go
```diff
@@ -11,7 +11,7 @@ import (
 
 type Team interface {
 	Name() string
```
clarafu, line 13: I think it might be better if we had an accessor that constructs an `atc.Team` object (which can be used directly in order to return the json flag) and places that are using the `Name()` and `Auth()` accessors can pull it off the `atc.Team` object. I did see you mention this in your PR description so I wanna know your thoughts too!

clarafu (commented):
Thanks for submitting this PR! 😄  So sorry for taking so long to review it!

go-concourse/concourse/team.go
agurney: I think that could work, especially with changing to something like
```golang
type team struct {
    atc.Team
    connection internal.Connection
    httpAgent  internal.HTTPAgent
}
```

for the private type.

My hesitation is basically around not wanting to make too many far-reaching changes to the structures, which are used everywhere, _even though at the same time_ I do want to do something to straighten out `atc.Team` vs `db.Team` vs `concourse.Team` vs `db.team` vs `concourse.team`. I'll have a go at doing what you suggest, which would make the marshalling part more natural with the rest of the code.

clarafu: Right I definitely understand your concern, I took a quick peek to see how many places would need to change and I think it might only need to be adjusted in a few fly commands but I could be wrong. If it ends up being a more extensive change you can let me know and we can see where to go about from there?

agurney: I added a [new commit](https://github.com/concourse/concourse/pull/5996/commits/072c7ce9d1b096a51c57d20b573f01b41afb3c1a) with the suggested accessor implemented, but stopped short of removing the existing `team.Name()` and friends. On trying to push that through, it led to some pretty icky-looking changes to go through the additional layer of indirection in order to get the data. (I've also rebased.)

go-concourse/concourse/teams_test.go
```diff
@@ -41,7 +41,7 @@ var _ = Describe("ATC Handler Teams", func() {
 			It("returns the requested team", func() {
 				team, err := client.FindTeam(teamName)
 				Expect(err).NotTo(HaveOccurred())
-				Expect(team.Name()).To(Equal(teamName))
+				Expect(team.Name(),).To(Equal(teamName))
```
clarafu, line 44: I think you have an extra comma there!

clarafu (commented):
I like it! There are a lot of refactor changes but I think it is necessary for this approach and it doesn't feel like it's complicating the code, what do you think?

go-concourse/concourse/teams_test.go
agurney: Thanks for catching it, apologies for leaving it in.

clarafu (approved):
I tried running the `fly get-team --json` command against a concourse deployment with your changes and it works as intended. The output I get is 
```
{
  "id": 1,
  "name": "main",
  "auth": {
    "owner": {
      "groups": [],
      "users": [
        "local:test"
      ]
    }
  }
}
```

vs without your changes it gives me
```
{}
```

Everything looks great! Thanks so much for submitting the PR and making all those edits! :)

PR #6011, opened 2020-08-22
set_pipeline step header indicates whether changes were applied

### LINKED ISSUE #6000: set_pipeline step header should indicate whether changes were applied
When viewing a build that reconfigures pipelines, it's hard to tell which pipelines were actually changed:

image

Quick suggestion: highlight it in yellow, like the `get` step? Or the opposite - dim steps which had no changes to apply? No strong preference!

This would likely involve emitting an event from the `set_pipeline` step so that the UI can know to highlight the step.

andy-paine:

closes #6000  .

Add a new event for when a set_pipeline step runs called `SetPipelineChanged` that includes a boolean flag as to whether there was any diff. Frontend renders this as a yellow tooltip with a hover message of `pipeline config changed`
Screenshot 2020-08-22 at 00 14 38
Screenshot 2020-08-22 at 00 14 14
Screenshot 2020-08-22 at 10 10 15

## Notes to reviewer:
I didn't extensively test the tooltip bits as it is using the exact same event code as the `get` tooltips which are still covered. Never really written Elm before so feel free to point out if I'm doing something weird.

## Release Note
`set_pipeline` steps will now indicate if any changes were applied

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


web/elm/src/Build/StepTree/StepTree.elm
```diff
@@ -1074,16 +1074,22 @@ viewStepHeaderLabel : StepHeaderType -> StepID -> Html Message
...
+                    [ onMouseLeave <| Hover Nothing
+                    , onMouseEnter <| Hover <| Just <| ChangedStepLabel stepID "pipeline config changed"
+                    ]
+
+                _ ->
+                    []
     in
     Html.div
-        (id (toHtmlID <| FirstOccurrenceGetStepLabel stepID)
+        (id (toHtmlID <| ChangedStepLabel stepID "")
```
andy-paine, line 1092: `""` feels a bit weird but I think the alternative would have to be keeping the `FirstOccurrenceGetStepLabel` and creating a new `ChangedSetPipelineLabel` (in `DomID`) but in almost every situation this is just going to result in another branch in a `case` statement with almost identical contents.

matthewpereira:
I think this is a great quick win. In the long term it might be smart for us to come up with an icon or something that indicates what's going on rather than relying on colour, but since we already have the colour signal for changed get steps this feels like a smart way to capitalize on that.

My one suggestion is that we improve the readability of the tooltip at the same time. I think the existing background/foreground combination (top) was difficult to read, and we could improve that by switching to black text (middle), or changing both the foreground and background (bottom) for something that stands out better and is really easy to read:

Frame 11(2)

Here are the hex codes and contrast ratings for each of those combinations:

Tooltip contrast

That AAA rating (based on [WCAG guidelines](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast)) means that people with different visual impairments or who have low quality monitors or bad lighting conditions will all have an easier time reading the tooltip, and the rest of us won't have to try as hard either.

@jomsie @andy-paine What do you think? Would this be difficult to change?

andy-paine:
I went with the second one since Concourse has quite a dark theme, I felt that a lighter tooltip would help draw attention towards it. This affects the `get` step tooltip too but it also looks much better and shouldn't impact on anything else.

Screenshot 2020-08-25 at 15 11 23
Screenshot 2020-08-25 at 15 12 11

andy-paine:
In terms of bold vs. not bold, this is a side-by-side. The codebase uses a `700` default for font-weight and bold is `900`:
Screenshot 2020-08-25 at 15 12 11
Screenshot 2020-08-25 at 15 19 50

I think the bolder is probably easier to read. Thoughts on color + boldness? Both are trivial changes 👍

matthewpereira:
That's great - thanks for the quick turnaround time @andy-paine! I agree with you on the light tooltip background.

I think the `700` *looks* a little nicer, personally, but I could go either way. @jomsie do you have a preference?

andy-paine:
> shouldn't impact on anything else.

Turns out this was a wildly inaccurate statement 😆 - changing the default tooltip background color changes quite a bit. I can update everything else (I think being consistent about tooltips is a good plan) but it feels out of scope for this PR and I would rather raise it separately.

Do you mind if we treat this PR as "makes `set_pipeline` do the same as `get` does now" and we address the colors in a new one?

matthewpereira:
Haha, whoops. Sure - let's alter the colours separately. 👍 I'll make another issue and copy my comments above to it.

jomsie:
Howdy y'all!

So just to catch up with all the discussion, this PR is just to get the tooltip in for new changes on set_pipeline steps? Changing colours comes later (and separately)?

andy-paine:
Yep - I'm going to remove the last commit and stick it in a new PR

andy-paine:
Right should be in a reviewable state now - visually it will look like the original PR message again

atc/event/events.go
```diff
@@ -176,6 +176,14 @@ type FinishPut struct {
...
 func (FinishPut) Version() atc.EventVersion { return "5.1" }
 
+type SetPipelineChanged struct {
+	Origin  Origin `json:"origin"`
+	Changed bool   `json:"changed"`
+}
+
+func (SetPipelineChanged) EventType() atc.EventType  { return EventTypeSetPipelineChanged }
+func (SetPipelineChanged) Version() atc.EventVersion { return "1.0" }
+
```
agurney, line 186: More of a note for the future than something to bundle in here:

This new event type could also be rendered in `fly watch` by adding an extra case in `fly/eventstream/render.go`. That could be a quick follow-up for someone to add, or maybe you just want to slipstream that here. No strong feeling from me on that but I do think it would be nice to have rough parity between the textual and web views.

jomsie:
I apologize for taking so long reviewing this! I've been having some issues with my workstation lately. I haven't forgotten! :)

jomsie (approved):
Looks good! Thanks for this, adds some consistency to our use of tooltips :)

Issue #578, opened 2016-08-04
Support client side certificates for authentication


stepanstipl:
It would be great if Concourse would support TLS client side certificates for authentication (TLS mutual authentication).

This is widely supported by most web browsers and web servers (nginx, apache) and lot of projects support this as standard form of auth.

One way to implement this, that I was thinking about as possibly the easiest, would be by putting Concourse behind web server (like nginx) and reading client identity from http header (which would be populated by nginx from certificate's CN). This should mean minimal implementation on server side, support in fly cmd would have to be added. What I'm not sure about is functionality like fly intercept and what issues there might be with putting Concourse behind nginx.

xoebus:
I believe that this is fairly easy to do in pure Go without any termination in front.

cdutra:
We can set the ClientAuth and ClientCAs in the tls.Config on the server side: https://golang.org/pkg/crypto/tls/#Config

jtarchie:
I was investigating this.

It would need to be decided, feature wise, if you'd like to have mutual TLS for just the entire HTTPs server or just for authentication. For the first, it is as easy as providing a different TLS config and some command line arguments. The later would require two different HTTPs listeners on the same port.

In favor of the later, as it would provide a mechanism, besides basic auth, to do a non-interactive login for CI systems. We have clients that don't want to use basic auth as it is not "secure". This would probably require some changes to sky marshal and atc to support the bind of two HTTPs listeners.

acambera-nearsoft:
The link to pivotal-traker is broken. No idea of how to work around this issue.

acambera-nearsoft:
I made it work by excluding the `/api/` path from required cert authentication. It is not completely safe as it lefts `/api`/ unsecured, but it works.

Issue #5242, opened 2020-02-26
Separate .flyrc into two files - one with targets and one with bearer tokens


ari-becker:
## What challenge are you facing?

As an individual Concourse admin of a variety of Concourse instances with a variety of teams in each Concourse instance, I would like to keep my `.flyrc` in version control along with the rest of my `dotfiles`, for a) backup purposes and b) to have an easier transition when using different workstations.

## What would make this better?

Transition from `$HOME/.flyrc` to using `$XDG_CONFIG_HOME/concourse/fly.yaml` (listing the targets) and `$XDG_RUNTIME_DIR/concourse/fly-tokens.yaml` (listing the bearer tokens that are used when logged into specific targets). The `$XDG_CONFIG_HOME/concourse/fly.yaml` file can then be version-controlled.

This has the additional security benefit of ensuring that the user is effectively logged out of all of their targets when they log out of their machine (as `$XDG_RUNTIME_DIR` is deleted when the user logs out), and does the groundwork necessary to build a user daemon that could automatically log a user into each of the user's fly targets upon workstation login, and rotate the bearer tokens so that the user stays logged in while they stay logged into their workstation.

## Are you interested in implementing this yourself?

Sorry, I don't have the time at this moment.

aoldershaw:
I can see merit in splitting up the sensitive information from the configuration. With #6017, this sensitive information could extend beyond tokens, so maybe we can use `$XDG_RUNTIME_DIR/concourse/fly-sensitive.yaml` or something like that.

Rather than using `$XDG_CONFIG_DIR` directly, we could use [`os.UserConfigDir`](https://golang.org/pkg/os/#UserConfigDir) so it's portable to other OS's. I'm not sure what equivalents to `$XDG_RUNTIME_DIR` exist on OS X and Windows, though - any ideas there?

We should still support `$HOME/.flyrc` for a while to some extent. This might mean falling back to `$HOME/.flyrc` and/or having some conversion tool (e.g. `fly migrate-rc`?) to facilitate the transition.

avanier:
I think this is a more complex issue than meets the eye. By using [`os.UserConfigDir`](https://golang.org/pkg/os/#UserConfigDir), there will be a hit to discoverability, and thus debugging.

Properly implementing this would most likely require a few other changes like:

- Having the ability override the value set by `os.UserConfigDir`, either in the environment or in `.flyrc` to avoid leaving users with edge-case configurations stranded
[...]

PR #6017, opened 2020-08-26
Allow the fly http transport to use client certificates

avanier:

This PR adds the ability to declare client certificates when using the `login` command for use by fly's http transport.

This stems from the fact our Concourse deployment is imposed some security requirements that are not fulfilled natively by Concourse's authentication mechanisms. With client certificates attached to every request, this allows us to _pre-authenticate_ requests at the proxy or load-balancer level before requests reach ATC without having to resort to more traditional techniques like IP allow-lists.

Just to be clear, this feature is useful as-is without anything being implemented in ATC.

This works towards fulfilling some of the asks in #578 .

This PR does mainly two things:

- It adds the ability to tack-on client certificates to the http transport for consumption by a middleman before ATC
- It takes care of storing the settings in `.flyrc`

## Notes to reviewer:

It could be debated how the client-cert should be stored or be made reference to. I made the call to store the path at which the key-cert pair is made available and let the user deal with properly securing that file. That said, it could also be stored whole in `.flyrc`.

~Depending on how the review shakes out, I'll fill out the documentation afterwards.~

Weee documentation!

~As for tests, I'll need help.~ 

~Ok so it seems I've fixed the tests that were already there, but still need to write new tests for the new feature.~

Weee tests!

## Release Note

No idea, halp.

## Contributor Checklist

- [x] [Signed] all commits
- [x] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/fly/commands/login.go
+++ b/fly/commands/login.go
@@ -24,13 +24,15 @@ import (
 )
 
 type LoginCommand struct {
-	ATCURL      string       `short:"c" long:"concourse-url" description:"Concourse URL to authenticate with"`
-	Insecure    bool         `short:"k" long:"insecure" description:"Skip verification of the endpoint's SSL certificate"`
-	Username    string       `short:"u" long:"username" description:"Username for basic auth"`
-	Password    string       `short:"p" long:"password" description:"Password for basic auth"`
-	TeamName    string       `short:"n" long:"team-name" description:"Team to authenticate with"`
-	CACert      atc.PathFlag `long:"ca-cert" description:"Path to Concourse PEM-encoded CA certificate file."`
-	OpenBrowser bool         `short:"b" long:"open-browser" description:"Open browser to the auth endpoint"`
+	ATCURL         string       `short:"c" long:"concourse-url" description:"Concourse URL to authenticate with"`
+	Insecure       bool         `short:"k" long:"insecure" description:"Skip verification of the endpoint's SSL certificate"`
+	Username       string       `short:"u" long:"username" description:"Username for basic auth"`
+	Password       string       `short:"p" long:"password" description:"Password for basic auth"`
+	TeamName       string       `short:"n" long:"team-name" description:"Team to authenticate with"`
+	CACert         atc.PathFlag `long:"ca-cert" description:"Path to Concourse PEM-encoded CA certificate file."`
+	ClientCertPath atc.PathFlag `long:"client-cert" description:"Path to a PEM-encoded client certificate file."`
+	ClientKeyPath  atc.PathFlag `long:"client-key" description:"Path to a PEM-encoded client key file."`
+	OpenBrowser    bool         `short:"b" long:"open-browser" description:"Open browser to the auth endpoint"`
 
 	BrowserOnly bool
 }
@@ -63,6 +65,8 @@ func (command *LoginCommand) Execute(args []string) error {
 			command.TeamName,
 			command.Insecure,
 			caCert,
+			string(command.ClientCertPath),
+			string(command.ClientKeyPath),
 			Fly.Verbose,
 		)
 	} else {
@@ -71,6 +75,8 @@ func (command *LoginCommand) Execute(args []string) error {
 			command.TeamName,
 			command.Insecure,
 			caCert,
+			string(command.ClientCertPath),
+			string(command.ClientKeyPath),
 			Fly.Verbose,
 		)
 	}
@@ -153,7 +159,8 @@ func (command *LoginCommand) Execute(args []string) error {
 	err = command.verifyTeamExists(client.URL(), rc.TargetToken{
 		Type:  tokenType,
 		Value: tokenValue,
-	}, target.CACert())
+	}, target.CACert(), target.ClientCertPath(), target.ClientKeyPath())
+
 	if err != nil {
 		return err
 	}
@@ -165,6 +172,8 @@ func (command *LoginCommand) Execute(args []string) error {
 			Value: tokenValue,
 		},
 		target.CACert(),
+		target.ClientCertPath(),
+		target.ClientKeyPath(),
 	)
 }
 
@@ -306,7 +315,7 @@ func waitForTokenInput(tokenChannel chan string, errorChannel chan error, isRawM
 	}
 }
 
-func (command *LoginCommand) saveTarget(url string, token *rc.TargetToken, caCert string) error {
+func (command *LoginCommand) saveTarget(url string, token *rc.TargetToken, caCert string, clientCertPath string, clientKeyPath string) error {
 	err := rc.SaveTarget(
 		Fly.Target,
 		url,
@@ -317,6 +326,8 @@ func (command *LoginCommand) saveTarget(url string, token *rc.TargetToken, caCer
 			Value: token.Value,
 		},
 		caCert,
+		clientCertPath,
+		clientKeyPath,
 	)
 	if err != nil {
 		return err
@@ -505,13 +516,16 @@ func (command *LoginCommand) legacyAuth(target rc.Target, browserOnly bool, isRa
 	return "", "", nil
 }
 
-func (command *LoginCommand) verifyTeamExists(clientUrl string, token rc.TargetToken, caCert string) error {
+func (command *LoginCommand) verifyTeamExists(clientUrl string, token rc.TargetToken, caCert string, clientCertPath string,
+	clientKeyPath string) error {
 	verifyTarget, err := rc.NewAuthenticatedTarget("verify",
 		clientUrl,
 		command.TeamName,
 		command.Insecure,
 		&token,
 		caCert,
+		clientCertPath,
+		clientKeyPath,
 		false)
 	if err != nil {
 		return err
--- a/fly/commands/sync.go
+++ b/fly/commands/sync.go
@@ -10,14 +10,18 @@ import (
 	"github.com/vbauerster/mpb/v4/decor"
 
 	"github.com/concourse/concourse"
+	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/fly/commands/internal/displayhelpers"
 	"github.com/concourse/concourse/fly/rc"
 	"github.com/concourse/concourse/fly/ui"
 )
 
 type SyncCommand struct {
-	Force  bool   `long:"force" short:"f" description:"Sync even if versions already match."`
-	ATCURL string `long:"concourse-url" short:"c" description:"Concourse URL to sync with"`
+	Force          bool         `long:"force" short:"f" description:"Sync even if versions already match."`
+	ATCURL         string       `long:"concourse-url" short:"c" description:"Concourse URL to sync with"`
+	CACert         atc.PathFlag `long:"ca-cert" description:"Path to Concourse PEM-encoded CA certificate file."`
+	ClientCertPath atc.PathFlag `long:"client-certificate" description:"Path to a PEM-encoded client certificate file."`
+	ClientKeyPath  atc.PathFlag `long:"client-key" description:"Path to a PEM-encoded client key file."`
 }
 
 func (command *SyncCommand) Execute(args []string) error {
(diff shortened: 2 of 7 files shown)


avanier:
Hrm, ok. So I was trying to write some sensible test cases, and my inexperience at writing Go just met some things I had heard from other programmers: writing tests where file IO is concerned is pretty garbage.

Here's where I'm at. I'd like to add the ability to use client certs with `fly`. Cool beans. For convenience's sake, since _fly targets_ don't usually change, it makes sense to me that declaring where the client certificate and private key are located every time `fly` is invoked would not result in a particularly good user experience.

Looking at the current code, [storing a CAs to verify certificates against](https://github.com/concourse/concourse/blob/master/fly/rc/targets.go#L40) right within `.flyrc` doesn't seem too bad. I mean, if it's good enough for `kubectl`, it's good enough for `fly`, right? That said, storing a _client_ certificate is a whole different ballgame. It identifies _me_, and it's a secret that could be used beyond the scope of `fly`.

Reflexively, I figure that storing that certificate safely should be left to the user _(foreshadowing)_, and simply having paths pointing to the file in `.flyrc` makes perfect sense.

Also looking at `.flyrc`, we don't enforce permissions on that file. `0755` is rather permissive. The door is fairly well open for _fun things to happen here_.

So I go on my merry way, write my feature, unbreak already existing tests... and it hits me.

Writing relevant tests for this is not great. There's no quite obvious way to structure my interface to allow me to put the data in there when I'm testing.

This brings me to the following questions:

- Should I just power through, carry on with the current approach and figure out a way to write tests that make sense.
- Should I just change my approach and read the certificates at invoke time, then store them in `.flyrc` and be done with it. _(But the file permissions bother me and I'm afraid enforcing them will lead to other types of fun thing including cross-OS pleasures.)_
- Should I forgo `.flyrc` altogether and, say, read it from the environment every time. _(That feels icky.)_

Halp.

aoldershaw:
@avanier could you just write a new `.flyrc` in the tests pointing to some temp files that contain the cert+key? We already have a test helper `createFlyRc` for creating a new `.flyrc`, so it's just a matter of writing the files, e.g.

```go
const clientCert = "..."
const clientKey = "..."

certPath := filepath.Join(homeDir, "client.cert")
keyPath := filepath.Join(homeDir, "client.key")

err := ioutil.WriteFile(certPath, []byte(clientCert), 0600)

err = ioutil.WriteFile(keyPath, []byte(clientKey), 0600)

createFlyRc(rc.Targets{
	"some-target": {
		API:            "https://example.com/some-target",
		TeamName:       "main",
		Token:          &rc.TargetToken{Type: "Bearer", Value: validAccessToken(date(2020, 1, 1))},
		ClientCertPath: certPath,
		ClientKeyPath:  keyPath,
	},
})
```

avanier:
@aoldershaw Thanks for the feedback. I has a read at other people use of temp files tests and didn't notice we did them too. I'll stick with the current approach and pump out relevant tests.

I also noticed that #3476 addresses my permission issue. It just seems my `.flyrc` has been in my home for a long time and the current behaviour is to preserve file permissions. :stuck_out_tongue:

avanier:
@aoldershaw I think this should be ready for review.

ari-becker:
> Hrm, ok. So I was trying to write some sensible test cases, and my inexperience at writing Go just met some things I had heard from other programmers: writing tests where file IO is concerned is pretty garbage.
> 
> Here's where I'm at. I'd like to add the ability to use client certs with `fly`. Cool beans. For convenience's sake, since _fly targets_ don't usually change, it makes sense to me that declaring where the client certificate and private key are located every time `fly` is invoked would not result in a particularly good user experience.
> 
> Looking at the current code, [storing a CAs to verify certificates against](https://github.com/concourse/concourse/blob/master/fly/rc/targets.go#L40) right within `.flyrc` doesn't seem too bad. I mean, if it's good enough for `kubectl`, it's good enough for `fly`, right? That said, storing a _client_ certificate is a whole different ballgame. It identifies _me_, and it's a secret that could be used beyond the scope of `fly`.
> 
> Reflexively, I figured that storing that certificate safely should be left to the user _(foreshadowing)_, and simply having paths pointing to the file in `.flyrc` makes perfect sense.
> 
> Also looking at `.flyrc`, we don't enforce permissions on that file. `0755` is rather permissive. The door is fairly well open for _fun things to happen_ here.
> 
> So I go on my merry way, write my feature, unbreak already existing tests... and it hits me.
> 
> Writing relevant tests for this is not great. There's no quite obvious way to structure my interface to allow me to put the data of the client cert in there when I'm testing. All I have access to is the file paths I stored in `.flyrc`.
> 
> This brings me to the following questions:
> 
>     * Should I just power through, carry on with the current approach and figure out a way to write tests that make sense.
> 
>     * Should I just change my approach and read the certificates at invoke time, then store them in `.flyrc` and be done with it. _(But the file permissions bother me and I'm afraid enforcing them will lead to other types of fun thing including cross-OS pleasures.)_
> 
>     * Should I forgo `.flyrc` altogether and, say, read it from the environment every time. _(That feels icky.)_
> 
> 
> Halp.

Hi, I opened https://github.com/concourse/concourse/issues/5242 a while back which proposes a suggested solution to this. By splitting `.flyrc` into two files, one stored in `$XDG_CONFIG_HOME` and one stored in `XDG_RUNTIME_DIR`, this issue is neatly sidestepped. `$XDG_CONFIG_HOME` can be easily version-controlled, while the XDG standard guarantees that a) the `$XDG_RUNTIME_DIR` is only readable by the current user, and it's deleted upon logout, which solves the security issues surrounding the tokens. Additionally, this makes it substantially easier to later write automation that keeps the tokens in the `$XDG_RUNTIME_DIR` refreshed, so that users don't have to run `fly login` by hand for their list of targets.

avanier:
Just FYI, at this point all the changes I'm pushing are rebases to keep in sync with `master`.

aoldershaw:
@avanier sorry, I took some vacation so haven't had a chance to review this. Will take a look soon!

fly/rc/target.go
```diff
@@ -166,7 +191,13 @@ func LoadUnauthenticatedTarget(
 		return nil, err
 	}
 
-	httpClient := &http.Client{Transport: transport(insecure, caCertPool)}
+	var clientCertificate []tls.Certificate
+	clientCertificate, err = loadClientCertificate(clientCertPath, clientKeyPath)
+	if err != nil {
+		return nil, err
+	}
```
aoldershaw, line 198: With the CA Cert, we fall back to the existing values if left unspecified. In fact, it looks like there's a test for this: https://github.com/concourse/concourse/blob/9e29396f1a044a585fb2d84d1a7ff70d92212d3e/fly/integration/login_test.go#L284-L285.

Without that fallback, users would need to provide `--client-certificate` and `--client-key` every time they try to refresh their token, which could be a bit tedious. What do you think about adding a fallback here, e.g.

```go
if clientCertPath == "" {
	clientCertPath = targetProps.ClientCertPath
}
...
```

aoldershaw: On a similar note, maybe it's worthwhile to add a test to `login_test.go` and possibly `sync_test.go`

fly/commands/sync.go
```diff
@@ -10,14 +10,18 @@ import (
...
 	"github.com/concourse/concourse/fly/rc"
 	"github.com/concourse/concourse/fly/ui"
 )
 
 type SyncCommand struct {
-	Force  bool   `long:"force" short:"f" description:"Sync even if versions already match."`
-	ATCURL string `long:"concourse-url" short:"c" description:"Concourse URL to sync with"`
+	Force          bool         `long:"force" short:"f" description:"Sync even if versions already match."`
+	ATCURL         string       `long:"concourse-url" short:"c" description:"Concourse URL to sync with"`
+	CACert         atc.PathFlag `long:"ca-cert" description:"Path to Concourse PEM-encoded CA certificate file."`
```
aoldershaw, line 22: Good catch. Maybe it's also worth adding `Insecure` for consistency with `fly login`?

aoldershaw (changes requested):
Code and tests look good - just have a couple suggestions.

---

I did acceptance on this by setting up a docker-compose'd nginx with client auth following https://fardog.io/blog/2017/12/30/client-side-certificate-authentication-with-nginx/

I get 403 when trying to login without the client cert provided, and successful auth with it. I'm also able to make other requests to the server that make use of the cert.

fly/commands/sync.go
avanier: Oh, yeah. I forgot to mention that I did a drive-by on that too.

Yeah, I'll put insecure in too on the next push.

avanier:
@aoldershaw So, I put in all the changes from the review... except for adding an integration test for the client cert fallback, which was not explicitly asked for :smiling_imp: . I'm just punting. It would require me a bit more grokking of the integration tests to write it. Let me know if you think I should add it.

Other potential nitpick worthy point. The changes for the `insecure` flag is split across two commits. Let me know if I can also get away with not having to rebase that.

If can't be a weasel and get away with either, I'll do another push sometime in the latter half of the week.

aoldershaw:
> So, I put in all the changes from the review... except for adding an integration test for the client cert fallback, which was not explicitly asked for 😈 . I'm just punting. It would require me a bit more grokking of the integration tests to write it. Let me know if you think I should add it.
> 
> Other potential nitpick worthy point. The changes for the insecure flag is split across two commits. Let me know if I can also get away with not having to rebase that.

Hehe, no worries! More tests can never hurt, but it's not a huge deal IMO. With that said, I think I may not have been entirely clear in https://github.com/concourse/concourse/pull/6017#discussion_r485021033 - you added a fallback for `clientCertPath`, but we'll also want a fallback for `clientKeyPath`, since otherwise users still need to provide one of the flags every time they login!

avanier:
@aoldershaw Wow, ok, I'm a dummy! So, I just pushed the changes for that, and in this case I'll load the values from `.flyrc` if none of the flags are provided.

If only one of the two flags are provided, I'll assume the user is under-caffeinated and provide the usual error messages about using one flag and not the other.

avanier:
@aoldershaw I think this is good to go, but the `containerd-integration` test seems to have derped out. I can run the integration suite locally with no errors. I think this might be a transient error... or I need help understanding this failure.

aoldershaw:
@avanier yeah, looks like just a flake - I retriggered the job and it seemed to be happier.

p.s. you should have been invited to the contributor team on GitHub automatically after opening this PR which would grant you the ability to retrigger PR builds on https://ci.concourse-ci.org, but it looks like the automation there didn't work - I invited you manually, though.

Just going to do one last pass, then I think we're good to go!

fly/commands/login.go
```diff
@@ -24,13 +24,15 @@ import (
...
-	TeamName    string       `short:"n" long:"team-name" description:"Team to authenticate with"`
-	CACert      atc.PathFlag `long:"ca-cert" description:"Path to Concourse PEM-encoded CA certificate file."`
-	OpenBrowser bool         `short:"b" long:"open-browser" description:"Open browser to the auth endpoint"`
+	ATCURL         string       `short:"c" long:"concourse-url" description:"Concourse URL to authenticate with"`
+	Insecure       bool         `short:"k" long:"insecure" description:"Skip verification of the endpoint's SSL certificate"`
+	Username       string       `short:"u" long:"username" description:"Username for basic auth"`
+	Password       string       `short:"p" long:"password" description:"Password for basic auth"`
+	TeamName       string       `short:"n" long:"team-name" description:"Team to authenticate with"`
+	CACert         atc.PathFlag `long:"ca-cert" description:"Path to Concourse PEM-encoded CA certificate file."`
+	ClientCertPath atc.PathFlag `long:"client-certificate" description:"Path to a PEM-encoded client certificate file."`
```
aoldershaw, line 33: Maybe the flag should be `--client-cert` for consistency with `--ca-cert`

aoldershaw (changes requested):
Everything works as expected, but when I took a look at your docs PR, I noticed that the flag referenced there was `--client-cert` (rather than `--client-certificate`). One option could be to update the docs, but thinking about it, I think it makes more sense to rename the flag to `--client-cert` in `fly login` and `fly sync` for consistency with `--ca-cert`.

Sorry for adding yet another review cycle 😬

aoldershaw (approved):
Looks good, thanks for your patience and for the contribution!

Issue #4871, opened 2019-12-09
Record version history as a tree


vito:
Today we store versions with a `check_order` column which determines their order in the database. This means that when the version history "shifts around" we need to update a bunch of rows and give them a new `check_order`. This process is expensive and confusing as the `check_order` value continuously increments.

Instead, can we represent versions as a tree structure, similar to `git` commits, where each version points to its parent, and the pipeline resource just points to the latest version?

In this form, we can actually just leave 'deleted' versions around as no-longer-active branches in the version history.

taylorsilva:
I think is is a cool idea and worth keeping around.

PR #6018, opened 2020-08-27
Add feature of resource version reaper.

evanchaoli:

Concourse doesn't have a mechanism to reap resource versions, so that table `resource_config_versions` keeps growing unless some pipeline is destroyed.

On one of our clusters, the PG database crashed due to a memory issue. From the PG logs, I identified that a resource had 0.5M versions, so I had to destroy the pipeline in order to clean up those versions. Then I further dig into table `resource_config_versions` of the cluster, many resources have more than 100K versions.

Retain all versions of a resource seems to not make a lot of sense. It feels to me that retaining 1000 versions per resource should be good enough.

This PR adds a resource version reaper that will allow cluster administrator to configure `max-versions-per-resource`. If set to 0, then this feature is opt-out.

My algorithm includes 3 steps:

1. Get all `resource-config-scope-id` who has more than `max-versions-per-resource` versions.
2. For each such `resource-config-scope-id`, get `check_order` of #`max-versions-per-resource` version.
3. Delete versions whose `check_order` are smaller than `check_order` from step 2. Pinned version will not be deleted.

## Notes to reviewer:

I haven't run tests yet, but I have tested all SQL statements  in `db/version_reaper.go` that are main algorithm. Once the algorithm is confirmed by Concourse dev team, I'll add unit test cases and run tests.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/version_reaper.go
+++ b/atc/db/version_reaper.go
@@ -1,0 +1,124 @@
+package db
+
+import (
+	"code.cloudfoundry.org/lager"
+	"context"
+	"database/sql"
+	sq "github.com/Masterminds/squirrel"
+)
+
+type resourceVersionReaper struct {
+	conn                   Conn
+	logger                 lager.Logger
+	maxVersionsPerResource int
+}
+
+func NewResourceVersionReaper(logger lager.Logger, conn Conn, maxVersionsPerResource int) *resourceVersionReaper {
+	return &resourceVersionReaper{
+		conn:                   conn,
+		logger:                 logger,
+		maxVersionsPerResource: maxVersionsPerResource,
+	}
+}
+
+func (r *resourceVersionReaper) ReapVersions(ctx context.Context) error {
+	rcsIds, err := r.getResourceConfigScopesToReap(ctx)
+	if err != nil {
+		return err
+	}
+
+	for _, rcsId := range rcsIds {
+		checkOrder, err := r.getLastCheckOrderToRetain(ctx, rcsId)
+		if err != nil {
+			return nil
+		}
+
+		if checkOrder == 0 {
+			continue
+		}
+
+		n, err := r.deleteVersions(ctx, rcsId, checkOrder)
+		if err != nil {
+			r.logger.Error("failed-to-delete-verions", err, lager.Data{"resource-config-scope-id": rcsId})
+			continue
+		}
+		r.logger.Info("deleted-versions", lager.Data{"resource-config-scope-id": rcsId, "deleted-versions": n})
+	}
+
+	return nil
+}
+
+// getResourceConfigScopesToReap return a list of resource_config_scope_id who
+// has more than maxVersionsPerResource versions.
+func (r *resourceVersionReaper) getResourceConfigScopesToReap(ctx context.Context) ([]int, error) {
+	rows, err := psql.Select("count(*) a", "resource_config_scope_id").
+		From("resource_config_versions").
+		Where(sq.NotEq{
+			"check_order": 0,
+		}).
+		Having(sq.Gt{
+			"count(*)": r.maxVersionsPerResource,
+		}).
+		OrderBy("a DESC").
+		RunWith(r.conn).
+		QueryContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	rcsIds := []int{}
+	for rows.Next() {
+		var count, rcsId int
+		err = rows.Scan(&count, &rcsId)
+		if err != nil {
+			return nil, err
+		}
+
+		rcsIds = append(rcsIds, rcsId)
+	}
+
+	return rcsIds, nil
+}
+
+// getLastCheckOrderToRetain return a check_order of specified resource_config_scope_id.
+// Versions whose check_order are smaller than returned check_order can be deleted.
+func (r *resourceVersionReaper) getLastCheckOrderToRetain(ctx context.Context, rcsId int) (int, error) {
+	var checkOrder int
+	err := r.conn.QueryRowContext(ctx, `
+		SELECT check_order
+		FROM (
+			SELECT check_order
+			FROM resource_config_versions
+			WHERE resource_config_scope_id = $1 and check_order != 0
+            ORDER BY check_order DESC
+            LIMIT $2
+        ) AS rcv
+		ORDER BY check_order
+		LIMIT 1`, rcsId, r.maxVersionsPerResource).Scan(&checkOrder)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return 0, nil
+		}
+		return 0, err
+	}
+	return checkOrder, nil
+}
+
+// deleteVersions deletes unpinned versions whose check_order are smaller than lastCheckOrderToRetain.
+func (r *resourceVersionReaper) deleteVersions(ctx context.Context, rcsId int, lastCheckOrderToRetain int) (int64, error) {
+	result, err := r.conn.ExecContext(ctx, `
+		DELETE FROM resource_config_versions
+        WHERE resource_config_scope_id = $1 and check_order < $2 and check_order != 0 and version not in (
+			SELECT version
+			FROM resource_pins
+			WHERE resource_id in (
+				SELECT resource_id
+				FROM resource_config_scopes
+				WHERE id = $1
+			)
+        )`, rcsId, lastCheckOrderToRetain)
+	if err != nil {
+		return 0, err
+	}
+	return result.RowsAffected()
+}
--- a/atc/db/version_reaper.go
+++ b/atc/db/version_reaper.go
@@ -56,6 +56,7 @@ func (r *resourceVersionReaper) getResourceConfigScopesToReap(ctx context.Contex
 		Where(sq.NotEq{
 			"check_order": 0,
 		}).
+		GroupBy("resource_config_scope_id").
 		Having(sq.Gt{
 			"count(*)": r.maxVersionsPerResource,
 		}).
--- a/atc/gc/version_reaper.go
+++ b/atc/gc/version_reaper.go
@@ -15,10 +15,6 @@ type versionReaper struct {
 }
 
 func NewVersionReaper(conn db.Conn, maxVersionsPerResource int) *versionReaper {
-	if maxVersionsPerResource < 100 {
-		maxVersionsPerResource = 100
-	}
-
 	return &versionReaper{
 		conn:                   conn,
 		maxVersionsPerResource: maxVersionsPerResource,
--- a/atc/gc/version_reaper.go
+++ b/atc/gc/version_reaper.go
@@ -1,0 +1,42 @@
+package gc
+
+import (
+	"context"
+	"github.com/concourse/concourse/atc/db"
+	"time"
+
+	"code.cloudfoundry.org/lager/lagerctx"
+	"github.com/concourse/concourse/atc/metric"
+)
+
+type versionReaper struct {
+	conn                   db.Conn
+	maxVersionsPerResource int
+}
+
+func NewVersionReaper(conn db.Conn, maxVersionsPerResource int) *versionReaper {
+	if maxVersionsPerResource < 100 {
+		maxVersionsPerResource = 100
+	}
+
+	return &versionReaper{
+		conn:                   conn,
+		maxVersionsPerResource: maxVersionsPerResource,
+	}
+}
+
+func (r *versionReaper) Run(ctx context.Context) error {
+	logger := lagerctx.FromContext(ctx).Session("version-reaper")
+
+	start := time.Now()
+	defer func() {
+		metric.VersionReaperDuration{
+			Duration: time.Since(start),
+		}.Emit(logger)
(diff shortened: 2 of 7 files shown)


agurney:
From your DB, what resource types are implicated in excess versions? I'm guessing it's almost all "time".

evanchaoli:
> From your DB, what resource types are implicated in excess versions? I'm guessing it's almost all "time".

Well, `time` is a big source of excess versions, and I also see a lot of other resources having 10K+ versions. Some are user developed resource types. So I think, even if Concourse eventually eliminates `time` resource, it would be still necessary to have a mechanism to clean up excess versions.

atc/gc/version_reaper.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/concourse/concourse/atc/metric"
+)
+
+type versionReaper struct {
+	conn                   db.Conn
+	maxVersionsPerResource int
+}
+
+func NewVersionReaper(conn db.Conn, maxVersionsPerResource int) *versionReaper {
+	if maxVersionsPerResource < 100 {
```
clarafu, line 18: Was there a reason why we want to have a minimum of 100 versions per resource?

atc/db/version_reaper.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// getResourceConfigScopesToReap return a list of resource_config_scope_id who
+// has more than maxVersionsPerResource versions.
+func (r *resourceVersionReaper) getResourceConfigScopesToReap(ctx context.Context) ([]int, error) {
+	rows, err := psql.Select("count(*) a", "resource_config_scope_id").
+		From("resource_config_versions").
+		Where(sq.NotEq{
+			"check_order": 0,
+		}).
```
clarafu, line 58: I think we need a `GROUP BY resource_config_scope_id` to allow this query to work. But also when I ran this query on an environment that has a lot of resource versions, it took 12 seconds to run. 

```
Sort  (cost=135253.56..135255.41 rows=742 width=12) (actual time=12913.111..12913.138 rows=434 loops=1)
  Sort Key: (count(*)) DESC
  Sort Method: quicksort  Memory: 45kB
  ->  HashAggregate  (cost=135210.76..135218.18 rows=742 width=12) (actual time=12912.473..12913.007 rows=434 loops=1)
        Group Key: resource_config_scope_id
        Filter: (count(*) > 100)
        Rows Removed by Filter: 2363
        ->  Seq Scan on resource_config_versions  (cost=0.00..118432.66 rows=2237080 width=4) (actual time=0.633..12211.803 rows=2239886 loops=1)
              Filter: (check_order <> 0)
              Rows Removed by Filter: 1
Planning time: 0.116 ms
Execution time: 12913.192 ms
```

You might be able to add an index to `check_order` in order to speed it up? I'm not too sure though and I'm also hesitant to add that index to this table (if there is another option).

atc/gc/version_reaper.go
evanchaoli: This is just to prevent from unreasonable min value, say, 1 version to retain.

atc/db/version_reaper.go
evanchaoli: Good catch! My original SQL statement is:

```sql
select count(*) a, resource_config_scope_id from resource_config_versions where check_order != 0 group by resource_config_scope_id having count(*) > 1000 order by a desc
```

In my DB, table `resource_config_versions` has 3.3M tuples, and the query takes less than 2 seconds. I also tried to get rid of `check_order != 0` from the query, it didn't speed up significantly. But considering the following 2 statements also need to sort on `check_order`, I think it's worth adding an index on `resource_config_scope_id+check_order`. I guess the gain of keep `resource_config_versions` small would be bigger than the cost of adding a new index.

evanchaoli:
> My other big concern is having this new gc not introduce extra db load. I haven't looked too much into the current algorithm that this gc runs but I'll definitely spend some time on it. I'll let you know any feedback I have on it asap!

@clarafu This reaper may only run once 30 minutes or 1 hour or even longer, so I don't think it would add much db load.

I have tested the code locally, and unit tests have been added. So it's ready for final review and merge.

atc/gc/version_reaper.go
clarafu: I'm not sure how I feel about having a hardcoded strict minimum because there could be the case where someone only cares about having fewer than 100 versions per resource? And when they try to set less than 100 it is not clear why their set minimum isn't respected (if we do want to have a strict min maybe that should be a flag validation)? Is there any benefit to setting this min value?

evanchaoli: @clarafu  My thought is that we should prevent user from setting some unreasonable value, for example 1. If a user sets only to retain 1 version per resource, I am afraid Concourse not behavior properly. I agree 100 is a random coming out from my brain, I just feel it should be a reasonable limitation. If you believe 1 is a safe limitation, then I can remove this check.

clarafu: If someone sets it to be a really low value like 1, I don't think there will be necessarily any concourse problems that come about from it. There might be things that will incentivize users from not setting it to such a low value, such as a version not being able to run through the entire pipeline in time before another version comes in and ends up deleting the previous version? Maybe we can be clear in our documentation that if a low value is set, it can possibly result in your versions not making it completely through your pipelines because it gets deleted before it can do so.

evanchaoli: Okey, I just deleted the check. Please see if you have more comments.

evanchaoli:
@clarafu @muntac Please help review and move forward this PR, I wish 6.6, because `resource_config_versions` table has been too big on our prod cluster, I have to manually clean up tuples from the table recently.

evanchaoli:
@vito can you please help move forward this pr? I really want it as early as possible.

vito:
> I think the idea of having a resource version gcer is pretty interesting. I do understand how versions can get old and useless especially for resources like the time resource and they just end up being a waste of space. I wonder how this would work with the new resource prototypes where we want to emit all versions whenever we run a check. https://github.com/concourse/rfcs/blob/master/038-resource-prototypes/proposal.md#check-discover-new-versions-of-a-resource. Might want some insight from @vito about this?

This does seem to run counter to the goal of having resources detect/store all versions available. I think this approach is fine for now but I don't think we want this in the long run.

I don't think this will necessarily break anything though, as long as it deletes the oldest versions first. The first `check` will return all versions, and all but the latest N will be removed, and subsequent `check`s will just keep finding new versions from the latest version.

The intent of storing all versions is so that the version history does not get blown away due to a config change or credential rotation. Right now the history resets to the latest version, which causes automatic credential rotation to chronically break pipelines. Once we change `check` to collect all versions this will be fixed because versions won't disappear anymore - resources will "reset" to a fully populated version history generated from the new config.

This is all based on the expectation that resources have a bounded, finite history. Resources don't typically have millions or even thousands of versions. Even Git repos with a ton of commits won't store every single commit - `git log --first-parent` only returns 8k for the Concourse repo which has ~23k commits. So this should be fine for "proper" resource types (see ["What makes a resource?"](https://blog.concourse-ci.org/reinventing-resource-types/)).

Some resource types break this expectation - mainly `time`, which generates an infinitely growing version history. In my experience this is the number one source of the issue that this PR is addressing.

The v10 plan is to convert `time` from a resource type to a [var source prototype](https://github.com/concourse/rfcs/blob/master/027-var-steps/proposal.md) so that it doesn't have a version history at all, and instead is directly tied to triggering individual jobs. Once that's done I don't think we should have a version reaper - we should continue to ensure resource are only used for things that have a bounded history.

I know there's an argument of "users could write a resource type that saves 1,000,000 versions" but if we're trying to address that issue there are many many other aspects of Concourse that we would have to address. A user could also run `fly execute` in a loop and generate 1,000,000 builds. If saving 1,000,000 versions breaks Concourse, then I'd like to investigate why and try to fix it. In fact we kind of already did with v6.0's new scheduler.

@evanchaoli I'd like to know more about the memory issue. What Concourse version did you observe this behavior on? It's possible this issue is already fixed.

evanchaoli:
@vito First of all, this feature is a opt-in feature and by default off, so that it would not impact anything if users don't turn on the switch.

I believe this feature will be very useful for large deployment. 1 million version may not break Concourse, but how about 10 millions, 100 millions, and so on ... Currently, on our cluster, `resource_config_versions` table holds 3 million tuples, and 0.5 million of them are `time` versions. 

There are 255 resources having 5000+ versions; 449 resources having more than 1000+ versions. Once this feature is in place, I may use a loose limitation, maybe 5000 or 3000. I agree that a normal resource, like `git` might be hard to generate 3000 versions, so the limitation won't impact most of pipelines.

The other things, even when v10 is out, if you don't completely remove `time` resource, then people may still use it, which means we always need a mechanism to clean up old versions. Again, keeping all versions might not be a problem for small deployments, but I believe it's a potential risk for large deployments.

For the memory issue we hit, I'm not 100% sure that was due to the big version table. What happened is:
* Our cluster is on 5.8.0.
* One day, the database kept crashing.
* From PG's log, the db team found the following log:

```
2020-08-23 23:20:02 PDT -  -  -  DETAIL:  Failed process was running:
                        WITH max_checkorder AS (
                                SELECT max(check_order) co
                                FROM resource_config_versions
                                WHERE resource_config_scope_id = $1
                        )
                        UPDATE resource_config_versions
                        SET check_order = mc.co + 1
                        FROM max_checkorder mc
                        WHERE resource_config_scope_id = $1
                        AND version = $2
                        AND check_order <= mc.co;
```

* Following this clue, I found that `resource_config_versions` table had 3.5 million tuples at the time. I further found the no.1 resource has 90K+ versions, then I identified the pipeline, luckily, the pipeline owner is under same group as me, so I destroyed the pipeline after communicated with the pipeline owner. Immediately, 0.5 million tuples were deleted from `resource_config_versions` table. After that, the db stopped crashing.
* But later on, the db team told me that they also tuned a parameter of PG, which is why I am not 100% sure deleting 0.5 million tuples solved the problem. I think the problem is not about 0.5 tuples for the table, instead, I guess it's more about 0.5 million versions on a single resource.

I remember some other user from the community also complained his version table containing builds and versions since 2018. So this feature should be generic and needed by other users.

vito:
@evanchaoli When it comes to opt-in functionality, I try to imagine a sane default. I'm not a huge fan of leaving things opt-in: it makes it harder to know what product users are using, which makes it harder to guide the project. It also means users will still run into the same problem until they find out that the knob exists, which means we have more documentation work on our hands. I would prefer to be building an opinionated product that "just works" out of the box, so even if a proposed feature is opt-in, I pretend that it isn't so that I can consider all the implications.

When performance issues are identified in large-scale deployments, IMO reaping data should be the last resort. To me, the only reason to delete data should be to save disk space. To do so for the sake of optimization feels like eschewing engineering responsibility. It may fix the issue, but it does so at the expense of users (who have their data disappear) and operators (who have a knob they're responsible for tuning).

In this case, v6.0 fixes the query in question along with *many* other optimizations and architectural changes which will dramatically reduce database load. To fix this query in particular, it only took adding an index:

https://github.com/concourse/concourse/blob/a608c05e20672419cfaaa17332be102ba2f7041a/atc/db/migration/migrations/1579713185_add_indexes_to_builds_and_versions_tables.up.sql#L4

Looking beyond this simple optimization, #4871 suggests an architectural change that would eliminate the need for the query in the first place.

So, an apology: I suggested going ahead with this PR because I assumed this was an issue with v6.0+. Now that I know it's already fixed, and keeping our future plans in mind, I'm less supportive of this change. Sorry for going back and forth on this.

If instead your deployment had limits in place (assuming some default value), it's possible that we would have never identified this problem and your deployment would constantly be running an inefficient query while sacrificing user data. It's also possible that performance would be worse than if we had addressed the core issue, because every version saved would imply another database write to remove the oldest versions. This deletion query would have its own performance concerns, and the original query would still be inefficient.

I understand that users have complained about having a ton of build/version data in the past, but they're primarily complaining because it's *slow*. Rather than solving the issue by deleting the data that causes the problem, I would prefer to improve performance by reducing the amount of work to be done (through architectural improvements) or optimizing the work that's already done (by adding indexes and tweaking queries).

I recognize that there may be potential issues in the future when a user stores 1,000,000,000 versions - but I would prefer to wait and see if that actually happens so that we can plan a fix that addresses what actually goes wrong. I don't want to add this if it's purely based on a hypothetical, because it contradicts our future plans and adds more code to maintain.

evanchaoli:
@vito We will upgrade to 6.4.0 soon and let's see.

I'm also ok to give up this PR. This is a just a few SQL statements. If the problem is still there, I will run a clone job with the SQL statements to cleanup our database, which will be more flexible, no need to wait and upgrade to certain Concourse version, and we can cleanup db either periodically or on-demand.

PR #6020, opened 2020-08-27
[WIP] Allow pipeline to specify custom header bar color

andy-paine:
Allows users to quickly identify a particular pipeline visually for when pipelines are shown on dashboards and the structure may not be enough to differentiate them.

Adds a `display` config to the pipeline YAML and stores this in a `jsonb` column in the DB. This config is exposed via API and frontend renders it as a colored bar below main top bar in pipeline view.

Looks like this:
Screenshot 2020-08-27 at 14 42 07

## Notes to reviewer:
The user need attempting to be met here is "as an engineer with lots of similar pipelines on a monitor, I want to quickly identify which pipeline is which." I'm not particularly tied to the UI bar element as a means to do this but _something_ that serves this purpose would be useful - especially as instanced pipelines will make it much more common to have very similar looking pipelines that color may assist in differentiating between.

Most of this PR is useful if we want to provide any custom display stuff for a pipeline (e.g. a description) so even if the UI bit is nasty (it kinda is), it would be good to discuss if there is a useful next step for this work that could reuse a lot of the `DisplayConfig` stuff in the ATC.

This is a draft as I just want to gather feedback. Some `fly` tests are failing and the UX is a bit nasty where it says
```
display configuration has been added:
      + {}
```
which would be preferable to fix. 

## Alternatives
Some other designs I tried for reference.
Screenshot 2020-08-21 at 13 23 57
Screenshot 2020-08-21 at 13 25 51

## Release Note
N/A draft

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/api/pipelines_test.go
+++ b/atc/api/pipelines_test.go
@@ -45,6 +45,9 @@ var _ = Describe("Pipelines API", func() {
 				Resources: []string{"resource3", "resource4"},
 			},
 		})
+		publicPipeline.DisplayReturns(atc.DisplayConfig{
+			HeaderColor: "#FFFFFF",
+		})
 		publicPipeline.LastUpdatedReturns(time.Unix(1, 0))
 
 		anotherPublicPipeline = new(dbfakes.FakePipeline)
@@ -131,7 +134,10 @@ var _ = Describe("Pipelines API", func() {
 							"jobs": ["job3", "job4"],
 							"resources": ["resource3", "resource4"]
 						}
-					]
+					],
+					"display": {
+						"header_color": "#FFFFFF"
+					}
 				},
 				{
 					"id": 2,
@@ -140,7 +146,8 @@ var _ = Describe("Pipelines API", func() {
 					"public": true,
 					"archived": false,
 					"team_name": "another",
-					"last_updated": 1
+					"last_updated": 1,
+					"display": {}
 				}
 			]`))
 		})
@@ -283,7 +290,8 @@ var _ = Describe("Pipelines API", func() {
 								"jobs": ["job1", "job2"],
 								"resources": ["resource1", "resource2"]
 							}
-						]
+						],
+						"display": {}
 					},
 					{
 						"id": 1,
@@ -299,7 +307,10 @@ var _ = Describe("Pipelines API", func() {
 								"jobs": ["job3", "job4"],
 								"resources": ["resource3", "resource4"]
 							}
-						]
+						],
+						"display": {
+							"header_color": "#FFFFFF"
+						}
 					}
 				]`))
 			})
@@ -387,6 +398,9 @@ var _ = Describe("Pipelines API", func() {
 					Resources: []string{"resource3", "resource4"},
 				},
 			})
+			fakePipeline.DisplayReturns(atc.DisplayConfig{
+				HeaderColor: "#FFFFFF",
+			})
 			fakePipeline.LastUpdatedReturns(time.Unix(1, 0))
 		})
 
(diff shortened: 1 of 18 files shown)


aoldershaw:
Is the idea to just make the pipelines visually different, or is there significance to the chosen colours? Or is it a mix of the two?

If there are some cases of just wanting a pipeline to be visually distinguishable from others, how would you feel about supporting `header_color: random` (or something along those lines) in addition to `header_color: #abcdef`. The ATC would then pick a "random" colour for the pipeline (possibly derived from either the pipeline ID or the team name + pipeline name + eventually instance vars). We'd have to make sure that the random set of colours are visually different enough, so probably wouldn't want to just take any random 3 bytes and call it a day.

What do you think? I think it might useful when used in conjunction with the `across` step - e.g. if you're setting a pipeline instance for each PR, I'm guessing you don't really care what the colours are but you might want them to be look different.

That said, maybe it should just be the responsibility of the caller to generate a colour, since then they could e.g. assign a unique colour per PR-author or something more useful than totally random colours.

This is probably a silly idea haha, but just throwing it out there!

cc @matthewpereira

matthewpereira:
Since we use color to distinguish _status_ (success, failure, halted, errored, paused, etc) in the Concourse UI, we have to be pretty cautious about introducing new colors, allowing users to choose their own, and which elements on the UI we allow to change.

This issue further validates this idea that users want ways to amplify at-a-glance identification of pipelines, and it's insightful that new developments like instanced pipelines will make it more and more crucial for us to be able to visually 'fingerprint' a specific permutation so that users can better tell them apart.

We've had requests in the past for changing background colors and images (https://chrome.google.com/webstore/detail/sky-concourse-background/ppokkdhjcoaaafajohjcebifbneannec and https://chrome.google.com/webstore/detail/concourse-status-amplifie/hlocifdhegikgbaineimppincgbaeheg come to mind), and that might be a decent starting point here as well. Images are less likely to be mistaken for one of our existing key colors in the app.

It makes me think of the https://www.wikiwand.com/en/Identicon system; maybe something like this could be incorporated somewhere to provide some of this at-a-glance identifying information, without running the risk of being conflated with pipeline status.

I do really like the idea of being able to set descriptions or notes, and we have a parallel idea to this around letting administrators/operations teams set a banner message to give status updates or warnings before planned outages. It might be time to first revisit these concerns with a design studio to map out all the potential front end implementations for these ideas to find out where they overlap.

andy-paine:
I would be happy to pivot this into a custom background image - that's actually what this was originally going to be before we toned it down to just a custom color bar. Identicon seems like an interesting idea - whereabouts in the UI were you thinking of including it? Maybe a ~30%ish opacity background? Or a medium sized icon in one corner?

DanielJonesEB:
Background images would be great, and were the initial thought for this. We use Concourse a lot for deploying infrastructure environments, named after _things:_ fish, arthropods, foods, countries. A background [image of a vinegaroon](https://upload.wikimedia.org/wikipedia/commons/3/34/Mastigoproctus_giganteus_0004_L.D..jpg) would both make the pipeline's identity immediately and unmistakably obvious, whilst also having the benefit of giving people nightmares.

Plus, memes.

andy-paine:
Closing in favour of #6053. Thanks for you input all 🙂

PR #6022, opened 2020-08-27
Run checks as builds

vito:

Feature

This PR removes the `checks` table and LIDAR checker and runs checks as builds instead. They can now be viewed in the web UI, and `fly check-resource` and `fly check-resource-type` will stream the build logs. It's kind of neat!

image

image

## TODO

* [x] Check rate limiting
* [x] Migration to make `builds.id` a `bigint`
* [x] Restore span context flow so versions can be traced from builds that saved them to builds that were created from them

Side note: I would recommend reading each commit's message.

The following is a broad overview of the changes.

* Removes `resource_types.unique_version_history`. This is a breaking change but it should have little to no real-world impact.
  * **Motivation:** as I was going through the code for constructing resource config scopes, I noticed that this property means that we have to hoist `atc.VersionedResourceTypes` around. This felt unnecessarily complicated for something that's never really used.

* Modifies the `check` step to adopt more of the functionality previously implemented by the LIDAR checker or other components.
  * Enforce rate limiting for periodic checks.
  * No-op if the resource has already been checked within its interval.
  * Use the latest version as the cursor version if no cursor is specified.
  * Update the resource [type] to point to the resource config scope.
  * Update the resource config scope check start/end time.
  * Lock to ensure only one check is running per scope.
  * Set the resource's checking error.
  * **Motivation:** I want to be able to use the `check` step more liberally, for example for image resource fetching. By pushing more behavior down into it, we can reuse it more easily while having it "do the right thing."

* Removes the LIDAR checker, the checks API, the `checks` table, and the `db.Check` type.
  * **Motivation:** in order to use the `check` step for image fetching it needs to be able to exist in an `atc.Plan` which also has a `get` step (for fetching the image) and whatever step needs to run (a `task` or a resource using a resource type).

    They way the code was structured before, a `check` step could *only* be run for a `db.Check`, and the LIDAR checker could only run if the plan had a single `CheckPlan` as the only step. (It would actually panic if it was anything but.)

* Modifies the LIDAR scanner (and check factory) to create a `db.Build` instead of `db.Check`.
  * Admittedly it's a little strange that the `db.CheckFactory` type now creates a `db.Build`. I think this can be cleaned up in a future refactor which cleans up the job + resource/resource type build scheduling patterns.
  * **Motivation:** with the `check` step capable of doing all that on its own, all LIDAR has to do is queue up a build that runs it. The build will be run by the build tracker.

* Simplifies resource config GC - rather than referencing `resource_config_check_sessions`, add a time-based "last referenced" GC.
  * **Motivation:** now that the `check` step creates its own config + scope, I needed a way to ensure the config doesn't get GC'd before it's assigned to the resource. A simple timestamp seems to do the trick, and allows the config GC query to be simplified a bit - we don't need to reference `resource_config_check_sessions` anymore, because the "last referenced" timestamp will always be bumped by the `check` step anyway.

* Adds support for rendering `check` steps in the web UI and sending check output to the build event stream.
  * **Motivation:** 'cause it's cool.

* Consolidates build tracing/logging data into `(db.Build).TracingAttrs` and `(db.Build).LagerData`.
  * **Motivation:** previously we had many places that would set a build's job name in logging/tracing data. This has always been awkward because one-off builds do not have a job. Now that builds may also be for resources, this seemed like a good time to clean this up and add the resource to the data at the same time.

## Notes to reviewer:

Sorry this is so big - this is actually split out from a bigger refactor to replace the special-cased image fetching path with a composition of `check` and `get` steps. I found that this was not possible to do because `check` steps could only be run for a `db.Check` and not within a `db.Build`, so here I am!

There are a ton of changes here - I would recommend reviewing commit-by-commit. I spent a lot of time rebasing and amending to make sure each commit can be reviewed in isolation.

## Release Note

Resource `check` operations, which collect and save versions for pipeline resources, are now run as builds.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


vito:
@xtreme-sameer-vohra @aoldershaw @jomsie Just FYI this is ready to go now! I'll fill in a better release note before merging, but all the code changes should be done.

aoldershaw:
@vito sweet, good to know. Sorry for the lack of feedback thus far. I have skimmed through the code changes and tested it out briefly, and things seem to work really well! It's cool to see checks in the UI and in `fly`. It might also be nice for the resource page to link to the build for the latest check.

There's a lot to unpack here, and I'm not 100% familiar with `resource_configs`/`resource_config_scopes`/all the nuance around resource checking, so please bear with me while I try to wrap my head around it all!

vito:
@aoldershaw Cool! No worries, happy to answer any questions you have.

Yeah I had a similar thought for the resource page. It might be cool to even embed the build output directly there, since it should be pretty small (just a few steps). 🤔

xtreme-sameer-vohra:
> @xtreme-sameer-vohra @aoldershaw @jomsie Just FYI this is ready to go now! I'll fill in a better release note before merging, but all the code changes should be done.

Nice. 
I still haven't had the chance to go through it, but I had added myself to just peruse the change so don't block on my account :)

jomsie:
> > @xtreme-sameer-vohra @aoldershaw @jomsie Just FYI this is ready to go now! I'll fill in a better release note before merging, but all the code changes should be done.
> 
> Nice. 
> I still haven't had the chance to go through it, but I had added myself to just peruse the change so don't block on my account :)

I'll +1 this :D

atc/db/resource_check_rate_limiter.go
```diff
@@ -0,0 +1,105 @@
...
+		refreshConn:    refreshConn,
+		refreshLimiter: rate.NewLimiter(rate.Every(refreshInterval), 1),
+
+		mut: new(sync.Mutex),
+	}
+}
+
+func (limiter *ResourceCheckRateLimiter) Wait(ctx context.Context) error {
+	limiter.mut.Lock()
+	defer limiter.mut.Unlock()
```
aoldershaw, line 46: Is holding the lock while we wait going to cause issues while we wait for the `delay` to elapse? I suppose if if any check is rate limited, no other check will be able to run until at least `delay` passes, so perhaps it's fine - but curious about whether your reasoning was the same

atc/db/resource_check_rate_limiter.go
```diff
@@ -0,0 +1,105 @@
...
+	err := psql.Select("COUNT(id)").
+		From("resource_config_scopes").
+		RunWith(limiter.refreshConn).
+		QueryRow().
+		Scan(&count)
+	if err != nil {
+		return err
+	}
+
+	limit := rate.Limit(float64(count) / limiter.checkInterval.Seconds())
```
aoldershaw, line 92: Should this take into account `--max-checks-per-second` as we used to do? If that's specified, we don't even need to count the resource config scopes/refresh the limit - so maybe it'd be better as a separate implementation of `builder.RateLimiter` (e.g. `StaticCheckRateLimiter`)

aoldershaw (changes requested):
This is a great refactor, and things seem to be working quite nicely. Really appreciate all the effort you put into keeping the commits focused and leaving helpful commit messages. Have a few comments/questions

atc/db/resource_type.go
```diff
@@ -215,6 +226,23 @@ func (t *resourceType) Reload() (bool, error) {
...
 }
 
+func (r *resourceType) SetResourceConfigScope(scope ResourceConfigScope) error {
+	_, err := psql.Update("resource_types").
+		Set("resource_config_id", scope.ResourceConfig().ID()).
+		Where(sq.Eq{"id": r.id}).
+		Where(sq.Or{
+			sq.Eq{"resource_config_id": nil},
+			sq.NotEq{"resource_config_id": scope.ResourceConfig().ID()},
+		}).
```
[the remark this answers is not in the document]
vito: This is just to avoid performing an unnecessary write when the scope is already set to the same value. The `sq.Eq` `NULL` check is probably redundant with the `sq.NotEq` but it felt a little clearer to check for it explicitly, since `NULL` is the initial state.

atc/engine/builder/check_delegate.go
```diff
@@ -0,0 +1,238 @@
...
+
+	scope, err := config.FindOrCreateScope(resource) // ignore found, nil is ok
+	if err != nil {
+		return nil, fmt.Errorf("find or create scope: %w", err)
+	}
+
+	return scope, nil
+}
+
+func (d *checkDelegate) WaitAndRun(ctx context.Context, scope db.ResourceConfigScope) (lock.Lock, bool, error) {
```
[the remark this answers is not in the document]
vito: Haha, yeah this one's kind of hard to name. I agree that the current name isn't great. I almost went with `WaitAndShouldRun`. Happy to change it to `WaitToRun`.

atc/syslog/drainer.go
```diff
@@ -99,10 +94,12 @@ func (d *drainer) drainBuild(logger lager.Logger, build db.Build, syslog *Syslog
 				return err
 			}
 
-			payload := log.Payload
-			tag := build.TeamName() + "/" + build.PipelineName() + "/" + build.JobName() + "/" + build.Name() + "/" + string(log.Origin.ID)
-
-			err = syslog.Write(d.hostname, tag, time.Unix(log.Time, 0), payload)
+			err = syslog.Write(
+				d.hostname,
+				build.SyslogTag(log.Origin.ID),
```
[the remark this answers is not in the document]
vito: Hmm, yeah I think this will probably have to change more in the future. I kept it backwards-compatible for job builds, at least.

Given that we don't document the tag format, and it has only changed for one-off builds (which I don't think people care about syslogs for?), it I think it should be OK to change, but maybe @scottietremendous has more context on how customers use this, since I think it was originally a customer driven request.

FWIW I think this could be argued as a bugfix.

vito:
@aoldershaw Thanks for the review! I just merged `master` in and I think I've addressed all your feedback so far as separate commits.

atc/atccmd/command.go
```diff
@@ -1015,8 +1016,9 @@ func (cmd *RunCommand) backendComponents(
 	}
 
 	rateLimiter := db.NewResourceCheckRateLimiter(
-		dbConn,
+		rate.Limit(cmd.MaxChecksPerSecond),
```
aoldershaw, line 1021: The existing behaviour is:

* 0 or unset ==> default limit
* -1 ==> no limit
* anything else ==> static limit

I think we're just missing the `-1` case

vito: ah, yep - fixed!

atc/exec/check_step.go
```diff
@@ -93,11 +113,91 @@ func (step *CheckStep) run(ctx context.Context, state RunState) error {
...
+			return fmt.Errorf("save versions: %w", err)
+		}
+
+		_, err = scope.UpdateLastCheckEndTime()
+		if err != nil {
+			return fmt.Errorf("update check end time: %w", err)
+		}
+	}
+
+	err = step.delegate.PointToSavedVersions(scope)
```
aoldershaw, line 184: This is the only place we point a resource at its resource_config_scope, correct?

I set a pipeline locally whose resource script fails, and I was expecting to see the check error on the resource page in the UI, but it doesn't come up (it just always says `checked successfully`.

I realized that's because the check error is being set on the scope, but not the resource itself (is the resource check error for failing to start the check?). Since we never point the resource at the scope, though, the check error is "lost".

You mentioned we could stream the output of the build on the resource page - then I guess we wouldn't need `resource_config_scopes.check_error`, right?

vito: Ah, yeah good catch. I wonder if we're missing higher-level test coverage for this. Seems like it would prevent seeing check errors in general.

Some context: the scope is set after versions are found to make sure there isn't a window of time where a resource is pointing to an empty scope after e.g. a credential rotation or new resource type version.

I can change it to point to the new scope even if checking fails, the code might just look a little crusty.

In the long run, yeah, I think just showing the output/build status on the resource page would be a lot cleaner, since then we can get rid of this whole `check_error` thing. We could also return a resource's current build in the API to make the resource orange in the pipeline/dashboard UI. (Though maybe its color could instead reflect the build status for consistency's sake? 🤔)

This is definitely a blocker since it'll probably mean users don't see check failures at all in the UI. I think I'll push a quick fix first (set scope on check failure) to get the PR unblocked and then spike on the build UI stuff on another branch.

aoldershaw (approved):
I can't find anything wrong with it - seems like a pretty awesome cleanup, and I really like the new streaming check output and the implications this brings for running checks in regular build plans

vito:
@aoldershaw thanks for the feedback + review!

PR #6053, opened 2020-09-10
Custom background image

### LINKED ISSUE #1756: Feature Request: Background images
We often have several pipelines visible on one build monitor. I know there's the new dashboard view in beta, but this isn't quite as lovely as the main pipeline view.

How about allowing each pipeline YAML file to specify a background image URL? Most of a Concourse pipeline's view is actually the empty background, so it'd be helpful for those of us that work on multiple pipelines to differentiate them when many are visible, or to ensure that we're looking at the one we think we are.

Seems like it should be a small PR, but I figured there's no point raising one if there's no desire for such a feature to be merged.

andy-paine:
Allows users to quickly identify a particular pipeline visually for when pipelines are shown on dashboards and the structure may not be enough to differentiate them.
Bug Fix | _Feature- | Documentation

closes #1756
Replaces #6020 

Allow users to specify a custom background image which is put at 30% opacity, grayscaled and blended into existing background. This allows for some level of pipeline customisation without creating too many visual problems for the UI. Uses the CSS property `background-image: url('blah')` so either hyperlinks or `data:image/png;base64` etc. strings can be used. Background is rendered on update by Elm so can be hot-reloaded. 

Examples:
Screenshot 2020-09-10 at 14 14 47
plane

## Notes to reviewer:
I made the `Display` field a pointer so that a missing field wouldn't be included in the JSON output but this means some `Maybe` work has to happen in the frontend to deal with this. I think this isn't _too_ bad but felt like more work that it should have been?

I added some tests to the config diff-ing stuff as I was having trouble getting it to work right. I only covered `jobs` from the existing behaviour as all the other keys largely use the same codepaths and then I added new tests for the new display behaviour. Display is the first field in the pipeline config that is not a slice of some sort so doesn't match the current model of "equivalent item" hence I had to write some custom diff/render stuff to deal with it.

## Release Note
Pipeline authors can now include a custom `background_image` under the `display` key in pipeline config e.g:
```yaml
display:
  background_image: https://avatars1.githubusercontent.com/u/7809479?s=400&v=4
```

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/config_diff.go
+++ b/atc/config_diff.go
@@ -25,6 +25,11 @@ type Diff struct {
 	After  interface{}
 }
 
+type DisplayDiff struct {
+	Before *DisplayConfig
+	After  *DisplayConfig
+}
+
 func name(v interface{}) string {
 	return reflect.ValueOf(v).FieldByName("Name").String()
 }
@@ -49,6 +54,24 @@ func (diff Diff) Render(to io.Writer, label string) {
 
 		payloadB, _ := yaml.Marshal(diff.After)
 
+		renderDiff(to, "", string(payloadB))
+	}
+}
+
+func (diff DisplayDiff) Render(to io.Writer) {
+	label := "display configuration"
+	if diff.Before != nil && diff.After != nil {
+		fmt.Fprintf(to, ansi.Color("%s has changed:", "yellow")+"\n", label)
+		payloadA, _ := yaml.Marshal(diff.Before)
+		payloadB, _ := yaml.Marshal(diff.After)
+		renderDiff(to, string(payloadA), string(payloadB))
+	} else if diff.Before != nil {
+		fmt.Fprintf(to, ansi.Color("%s has been removed:", "yellow")+"\n", label)
+		payloadA, _ := yaml.Marshal(diff.Before)
+		renderDiff(to, string(payloadA), "")
+	} else {
+		fmt.Fprintf(to, ansi.Color("%s has been added:", "yellow")+"\n", label)
+		payloadB, _ := yaml.Marshal(diff.After)
 		renderDiff(to, "", string(payloadB))
 	}
 }
@@ -205,6 +228,20 @@ func diffIndices(oldIndex Index, newIndex Index) Diffs {
 	return diffs
 }
 
+func diffDisplay(oldDisplay, newDisplay *DisplayConfig) (DisplayDiff, bool) {
+	if oldDisplay == nil && newDisplay == nil {
+		return DisplayDiff{
+			Before: nil,
+			After:  nil,
+		}, false
+	}
+
+	return DisplayDiff{
+		Before: oldDisplay,
+		After:  newDisplay,
+	}, practicallyDifferent(oldDisplay, newDisplay)
+}
+
 func renderDiff(to io.Writer, a, b string) {
 	diffs := difflib.Diff(strings.Split(a, "\n"), strings.Split(b, "\n"))
 	indent := gexec.NewPrefixedWriter("\b\b", to)
@@ -291,5 +328,12 @@ func (c Config) Diff(out io.Writer, newConfig Config) bool {
 			diff.Render(indent, "job")
 		}
 	}
+
+	displayDiff, diff := diffDisplay(c.Display, newConfig.Display)
+	if diff {
+		diffExists = true
+		displayDiff.Render(indent)
+	}
+
 	return diffExists
 }
--- a/atc/configvalidate/validate.go
+++ b/atc/configvalidate/validate.go
@@ -3,6 +3,7 @@ package configvalidate
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"strings"
 
 	"github.com/concourse/concourse/atc"
@@ -54,6 +55,12 @@ func Validate(c Config) ([]ConfigWarning, []string) {
 		errorMessages = append(errorMessages, formatErr("jobs", jobsErr))
 	}
 	warnings = append(warnings, jobWarnings...)
+
+	displayWarnings, displayErr := validateDisplay(c)
+	if displayErr != nil {
+		errorMessages = append(errorMessages, formatErr("display config", displayErr))
+	}
+	warnings = append(warnings, displayWarnings...)
 
 	return warnings, errorMessages
 }
@@ -385,4 +392,29 @@ func validateVarSources(c Config) ([]ConfigWarning, error) {
 	}
 
 	return warnings, compositeErr(errorMessages)
+}
+
+func validateDisplay(c Config) ([]ConfigWarning, error) {
+	var warnings []ConfigWarning
+
+	if c.Display == nil {
+		return warnings, nil
+	}
+
+	url, err := url.Parse(c.Display.BackgroundImage)
+
+	if err != nil {
+		return warnings, fmt.Errorf("background_image is not a valid URL: %s", c.Display.BackgroundImage)
+	}
+
+	switch url.Scheme {
+	case "https":
+	case "http":
+	case "":
+		break
+	default:
+		return warnings, fmt.Errorf("background_image scheme must be either http, https or relative")
+	}
+
+	return warnings, nil
 }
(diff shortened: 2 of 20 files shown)


zoetian:
Heyy @andy-paine, this is a very interesting PR, thanks for submitting this! 

Can you please also run the `yarn analyse` for it? Seems like we got a test failed on CI. Thanks!

andy-paine:
Yeah I'm trying to solve it but it seems a little over-zealous in terms of what it will allow. It is complaining about:
```
type alias Display =
    { backgroundImage : Maybe String }
```
being a record with only 1 field but that's the structure of the object that the API is returning. It would be nice to add some more stuff into the `Display` object later so I don't really want to have to flatten both the record in the frontend and the API response in the ATC to just being `background_image` just to please the linter?

aoldershaw:
Haha this is awesome. 

I totally get why you made it a record (I would've done the same), but it doesn't appear there's any way of disabling a single check (https://github.com/stil4m/elm-analyse/issues/92), and I'd rather not disable it entirely! Plus, elm makes it pretty easy to refactor when we end up adding more fields to `Display`. It feels like an anti-pattern to do things just to make the linter happy, but I'm not sure what else we can do besides just tossing it, which would leave us worse-off IMO.

By the way, you don't have to flatten the API response to get this to work - you can use [`Json.Decode.at`](https://package.elm-lang.org/packages/elm/json/latest/Json-Decode#at)

aoldershaw (commented):
This is awesome - I never knew how much I needed this feature until I got to use it!

I was uncertain about how to go about displaying the background image without being too distracting, but I think you did a great job making it easy to distinguish pipelines while still having them look like Concourse pipelines.

One thing I'm a *bit* concerned about is the implications of allowing base64 encoded images. I think Postgres can do a decent job handling large row values using the TOAST table, so the actual storage isn't too concerning. However, I don't think the `/api/v1/pipelines` endpoint should return the `background_image` in this case - if there's 100 pipelines each with a `100KB` base64 encoded image, that's 10MB (every 5 seconds) (for every Concourse tab) (for every active user).

Granted, this is mostly an issue with the fact that the UI polls aggressively (which I've put some thought into getting rid of in #5802), but if we're only displaying the background image on the pipeline page, perhaps we can not include this display information in the `/api/v1/pipelines` endpoint. Alternatively, if we disallow base64 encoded background images, this isn't such an issue.

What do you think?

aoldershaw:
Oh also, it seems like opening the sidebar adds a scrollbar since the background image element is trying to take up 100% of the screen.

Screen Shot 2020-09-10 at 12 20 03 PM

Moving the background image element under the `pipeline-content` div seems to fix this.

----

p.s. this is SWEET, I can now have put a Concourse INSIDE of Concourse!

Screen Shot 2020-09-10 at 12 31 08 PM

atc/config_diff.go
```diff
@@ -53,6 +58,24 @@ func (diff Diff) Render(to io.Writer, label string) {
 	}
 }
 
+func (diff DisplayDiff) Render(to io.Writer) {
+	label := "display configuration"
+	if diff.Before != nil && diff.After != nil {
+		fmt.Fprintf(to, ansi.Color("%s has changed:", "yellow")+"\n", label)
+		payloadA, _ := yaml.Marshal(diff.Before)
+		payloadB, _ := yaml.Marshal(diff.After)
+		renderDiff(to, string(payloadA), string(payloadB))
```
chenbh, line 67: We probably shouldn't be rendering diffs if either Before or After is a base64 encoded image. A short summary or message like `((base64 encoded image omitted))` is a lot more meaningful than the actual value.

andy-paine: That's a good point (it is kinda ugly). I'm split as to whether to allow base64 encoded images as they have the benefit of not needing to egress to some random location as they are entirely contained within the pipeline config and the overhead of transferring that data. I think maybe just allowing links only is the simplest case and users who are running in an egress-controlled environment will either not be able to use the memes or will have to host them themselves (it is easy enough to just get the web node to host them on `/public/images` anyway.

web/elm/src/Pipeline/Pipeline.elm
```diff
@@ -454,6 +454,34 @@ isArchived p =
...
+    case pipeline of
+        RemoteData.Success p ->
+            case p.backgroundImage of
+                Just bg ->
+                    [ style "background-image" <|
+                        "url(\""
+                            ++ bg
+                            ++ "\")"
+                    , style "background-repeat" "no-repeat"
+                    , style "background-size" "100% 100%"
```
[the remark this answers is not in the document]
andy-paine: This looks way nicer! I don't do much frontend so this was me stumbling through css styles until it looked roughly right 👍

web/elm/src/Pipeline/Pipeline.elm
```diff
@@ -454,6 +454,34 @@ isArchived p =
...
+                    , style "background-blend-mode" "luminosity"
```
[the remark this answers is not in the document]
andy-paine: Seems not to - had a chat with @matthewpereira about this and I think we talked about the background blend mode but with the grayscale and the background being gray I don't think it does anything.

atc/config_diff.go
chenbh: I don't have a strong opinion either way. I guess that the teams that actually need this feature (i.e. differentiate between pipelines) are more likely to be in an enterprise environment. Which probably means that they have access to a s3-like bucket to serve the images.

DanielJonesEB:
Words cannot express how important this feature is for the future of humanity.

Subsequent generations will look back at this moment and wonder why no-one thought to bring about an eternity of peace, love and understanding via the medium of animated gifs sooner.

atc/configvalidate/validate.go
```diff
@@ -386,3 +393,23 @@ func validateVarSources(c Config) ([]ConfigWarning, error) {
...
+		return warnings, nil
+	}
+
+	url, err := url.Parse(c.Display.BackgroundImage)
+
+	if err != nil {
+		return warnings, fmt.Errorf("background_image is not a valid URL: %s", c.Display.BackgroundImage)
+	}
+
+	if url.Scheme == "data" {
```
aoldershaw, line 410: Okay, so I'm just being paranoid, but *maybe* we should be explicit about what schemes are valid (i.e. allow `http`/`https`/no scheme for relative URLs, rather than deny `data`).

Any sane browser will do this validation for us, but from what I can tell, certain older versions of IE are vulnerable to an XSS attack via the `javascript` scheme: https://owasp.org/www-community/xss-filter-evasion-cheatsheet#style-tag-using-background-image

Probably overkill, I know, but I figure it's better to add schemes in the future if people want them rather than trying to retroactively patch a vulnerability (that probably doesn't exist, but still) - or worse still, have the (unlikely) vulnerability go unnoticed by the "good guys"

Feel free to tell me I'm being crazy 😁  if you think this is totally unnecessary, lemme know

andy-paine: No I think that is reasonable - I was basing denying data off of what would work to show and image in that tag (from here https://developer.mozilla.org/en-US/docs/Web/CSS/url()) but handling the case of someone trying to do something naughty is also useful. Should also provide a bit of validation that the scheme is one that `url()` understands so may help users if they put in a URL with a scheme that wouldn't work (e.g. `tcp://`).

I've amended the last validation commit with that.

aoldershaw (approved):
Looks good to me! Excited for the meme potential this brings to the table.

@zoetian or @taylorsilva, did either of you want to take a deeper look?

taylorsilva:
@aoldershaw I read over the comments, I'm good with this 😁 
image

DanielJonesEB:
🍾

marco-m-pix4d:
Hello @andy-paine, I see that the Release Notes checkbox is checked, where do I find them ?

andy-paine:
Hi @marco-m-pix4d - I thought the `Release Note` section in the main PR comment magically got added when a release was cut. I've got a docs PR to raise as well to add it there.

marco-m-pix4d:
OK thanks!

Issue #5354, opened 2020-03-24
To support empty tags


evanchaoli:
## Bug Report

We received a use case from our users about pipeline templating. The users want to use var for tags, like:

```yaml
- task: some-task
  tags: [((worker-tag))]
  ...
```
Then the var will be given from a var file as:

```yaml
worker-tag: "a"
```
which works properly to direct the task to a "a" worker.

But if task needs to go to the default worker pool, specifying

```yaml
worker-tag: ""
```
doesn't work. The error is `no workers satisfying: resource type 'registry-image', platform 'linux', tag ''`

## Steps to Reproduce

1. Add `tags: [""]` to a task.
2. Then run the job.

## Expected Results

We expect the empty tag `""` to be considered as null tag, and be ignored.

## Actual Results

`""` is considered as a meaning tag, thus cannot match to any worker.

## Version Info

* Concourse version: 5.8.0
* Did this used to work? No

vito:
Would an acceptable alternative be to specify the array?:

```yaml
tags: ((worker_tags))
```

+

```yaml
worker_tags: []
```

evanchaoli:
I just tried, that works.

jlundy2:
@vito this is a great solution. I think this might be a helpful tip to put in the documentation somewhere.

PR #6057, opened 2020-09-10
Fix empty worker tags

### LINKED ISSUE #6056: Empty CONCOURSE_TAG environment variable stops workers from accepting jobs
## Summary

When rebuilding workers we set an empty environment variable `CONCOURSE_TAG`, which has the effect of making that worker unable to accept any jobs.

The error presented is 
```
run check step: run check step: find or choose worker for container: no workers satisfying: resource type 'git'
```

## Steps to reproduce

Set an empty `CONCOURSE_TAG` variable when running the `concourse worker` process

## Expected results

Having an empty variable should have the same behaviour as not setting the variable at all. i.e. the worker should be considered as having no tags at all

## Actual results

Worker is unable to accept any jobs 

## Triaging info

* Concourse version: 6.3.0, 6.5.1
* Did this used to work? 
Probably not

aholyoake-bc:
Empty worker tags should be treated the same as not having any tags at all
This makes having the CONCOURSE_TAGS being unset equivalent to CONCOURSE_TAGS being set and empty

This change should probably be handled in the environment variable wiring, but it's getting late and there's a lot of magic going on in there

Bug Fix

closes #6056 

If the tags list has length 1, but the only item is the empty string, then don't fail the match

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/worker/worker.go
+++ b/atc/worker/worker.go
@@ -764,8 +764,14 @@ func (worker *gardenWorker) Uptime() time.Duration {
 
 func (worker *gardenWorker) tagsMatch(tags []string) bool {
 	workerTags := worker.dbWorker.Tags()
-	if len(workerTags) > 0 && len(tags) == 0 {
-		return false
+	if len(tags) == 0 {
+		// If worker only has an empty tag due to user specifying "" instead of []
+		if len(workerTags) == 1 && workerTags[0] == "" {
+			return true
+		}
+		if len(workerTags) > 0 {
+			return false
+		}
 	}
 
 insert_coin:
--- a/atc/worker/worker_test.go
+++ b/atc/worker/worker_test.go
@@ -563,9 +563,40 @@ var _ = Describe("Worker", func() {
 				})
 			})
 
-			Context("when the worker has no tags", func() {
+			Context("when no tags are specified", func() {
+				BeforeEach(func() {
+					spec.Tags = []string{}
+				})
+
+				It("returns true", func() {
+					Expect(satisfies).To(BeFalse())
+				})
+			})
+
+			Context("when empty tags are specified", func() {
+				BeforeEach(func() {
+					spec.Tags = []string{""}
+				})
+
+				It("returns true", func() {
+					Expect(satisfies).To(BeFalse())
+				})
+			})
+
+			Context("when the worker has no tags and no requested tags", func() {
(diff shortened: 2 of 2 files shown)


chenbh:
Looks like there was a previous attempted fix at https://github.com/concourse/concourse/pull/5355 and https://github.com/concourse/concourse/issues/5354

aholyoake-bc:
#5355 looks like a sensible solution - any idea why it was closed?
#5354 solves a related problem of placing jobs on workers with empty tags, rather than configuring the workers themselves.

The question here is: is there any reason to consider an empty worker tag different to not having a worker tag at all?

atc/worker/worker.go
```diff
@@ -850,7 +850,7 @@ func (worker *gardenWorker) Uptime() time.Duration {
 
 func (worker *gardenWorker) tagsMatch(tags []string) bool {
 	workerTags := worker.dbWorker.Tags()
-	if len(workerTags) > 0 && len(tags) == 0 {
+	if len(workerTags) > 0 && !(len(workerTags) == 1 && workerTags[0] == "") && len(tags) == 0 {
```
muntac, line 853:
```suggestion
if len(tags) == 0 {
		// If worker only has an empty tag due to user specifying "" instead of []
		if len(workerTags) == 1 && workerTags[0] == "" {
			return true
		}
		if len(workerTags) > 0 {
			return false
		}
	}
```

muntac (changes requested):
@aholyoake-bc To answer your question - probably no. I tried hard to think of a way to do this at the environment variable level. I could only think of making a custom flag for the worker tags that would treat empty string as null, and then add it to the https://github.com/concourse/flag repo. But it seems like more trouble than it's worth.

We can go with this route. I made a small suggestion so that the condition is easier to read.

I would also add a test case somewhere around this [spot](https://github.com/concourse/concourse/blob/759aa59869586b1b3a974b9f101a52ff3e7973d8/atc/worker/worker_test.go#L566). Let me know if you need help with that.

muntac (approved):

PR #6060, opened 2020-09-11
Allow globs in groups

andy-paine:
Allow groups to list jobs via globs rather than just strict matches. It isn't uncommon to have an `all` group which you need to remember to add any new jobs to and it is easy to forget to `fly validate-pipeline` first. With this feature, the groups could just be (quoting to prevent being seen as YAML anchor):
```yaml
- name: all
  jobs:
  - "*"
- name: deploy
  jobs:
  - deploy-*
```

When validating/saving groups that a job is in, use `glob.Glob` rather than `Lookup`/map keys which strictly check equality. Add/modify some tests to ensure behaviour is covered.

## Notes to reviewer:
There are 2 commits here - the first uses a regular expression and the second converts it to using globs. I think globs seem nicer so I would _prefer_ to use globs and just squash these 2 commits but I left it there for reference. I don't think users need/necessarily want the full power of regular expressions just for selecting groups.

On my travels I also noticed that a group can contain resources - I'm pretty sure this field only exists on the object and in validation and isn't actually persisted to the DB anywhere. Doing some manual testing specifying `resources` in the pipeline config doesn't do anything so I'm pretty sure this is dead code. I say all this to qualify why I didn't bother extending the glob based matching to resources as well 😂 

## Release Note
`groups` in a pipeline can now match jobs based on globs e.g.:
```
groups:
- name: deploy
  jobs:
  - deploy-*
```

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/configvalidate/validate.go
+++ b/atc/configvalidate/validate.go
@@ -9,6 +9,7 @@ import (
 	"github.com/concourse/concourse/atc"
 	. "github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/creds"
+	"github.com/gobwas/glob"
 )
 
 func formatErr(groupName string, err error) string {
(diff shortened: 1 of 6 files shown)


agurney:
This looks useful for many of our teams who have fairly elaborate pipeline setups that include deployment and testing in multiple environments. I was just looking at some of the group configs our users currently have, and it should be nice if they can say stuff like `["terraform-*-staging"]` instead of `["terraform-plan-pod-staging", "terraform-apply-pod-staging", "terraform-plan-basic-staging", "terraform-apply-basic-staging"]`, and many other similar instances.

chenbh:
I wonder if we should use https://github.com/gobwas/glob and support the [full wildcards suite](https://tldp.org/LDP/GNU-Linux-Tools-Summary/html/x11655.htm) instead of just `*`. I can see a scenario where I want to group `deploy-to-{dev,qa,staging}` and `deploy-to-{uat,prod}` separately

andy-paine:
I could get behind that - I actually used that library to begin with but then didn't want to add a new dependency since the one I used is already included via the Vault dep

andy-paine:
I've squashed everything down to a single commit and it uses `gobwas` now

andy-paine:
I've fixed all the tests now. Can you let me know if there is anything more I need to do on this?

aoldershaw (approved):
Sorry for the delay getting this reviewed! Appreciate your patience 🙏 

Looks good to me! Seems like a useful feature - I have also felt the pain of forgetting to add new jobs to groups. While it's not *so* hard to add the job to the group after `fly` complains, it's certainly a QoL improvement. I also like how this encourages consistent naming of `jobs`!

Thanks!

PR #6079, opened 2020-09-18
feat(atc): add "defaults" to resource type.

### LINKED ISSUE #3360: --docker-registry doesn't work
## Bug Report

I'm trying to use our own docker artifactory as the default place to pull images. I did some try today on a 5.0 cluster.

By passing some environment variables to `concourse worker`, I can see that `gdn` has got `--docker-registry` option:

> root        26     6  0 07:45 ?        00:00:02 gdn server --bind-ip 0.0.0.0 --bind-port 7777 --depot /worker-state/depot --properties-path /worker-state/garden-properties.json --time-format rfc3339 --no-image-plugin --docker-registry my.artifactory.eng.vmware.com --insecure-docker-registry my.artifactory.eng.vmware.com --dns-server 172.18.0.4 --allow-host-access

Then I submit a basic pipeline to the cluster with specifying an image path of my docker artifactory, but Concourse failed to pull the image, so looks like `--docker-registry` doesn't work.

## Steps to Reproduce

Exports envs:

      CONCOURSE_GARDEN_DOCKER_REGISTRY=my.artifactory.eng.vmware.com
      CONCOURSE_GARDEN_INSECURE_DOCKER_REGISTRY=my.artifactory.eng.vmware.com

then start worker.

## Expected Results

Worker is expected to use my.artifactory.eng.vmware.com as the default registry.

## Actual Results

Cannot pull image from my.artifactory.eng.vmware.com.

## Version Info

* Concourse version: 5.0
* Infrastructure/IaaS: docker-compose
* Browser (if applicable): any
* Did this used to work? no

evanchaoli:

Feature

closes #3360

The initial motivation of this PR to allow a Concourse admin to configure a cluster to pull images from a non-DockerHub registry. To solve the issue the problem in a generic manner, the problem shifts to allow an admin to configure default sources to base resource types.

This is a new implementation of #5942, as this takes a totally different approach, so I'm starting a new PR.

This feature adds a `defaults` section to `resource_type` definition, and `defaults` will be applied to its directly child of resource or resource type.

A new column `defaults` is added to `base_resource_types` table, which allows cluster admin to configure `defaults` for built-in resource types.

My test pipelines:

```yaml
resource_types:
- name: rt-tiny
  type: registry-image
  source:
    repository: evanli/tiny2
    tag: v4
  defaults:
    key: ok-default

resources:
- name: tiny
  type: rt-tiny
  source:
    key2: ok

- name: resource
  type: mock
  source:
    create_files:
      task.yml:
        platform: linux
        image_resource:
          type: registry-image
          source: {repository: centos}
        run:
          path: echo
          args: ["hello world"]

jobs:
- name: test
  public: true
  plan:
  - get: tiny

  - task: task-should-succeed
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: sh
        args:
         - -exc
         - echo ok

  - get: resource
  - task: with-path
    file: resource/task.yml

  - put: tiny
``` 

Then I created a defaults file:

```yaml
registry-image:
  debug: true
```
And specify it to ATC:

```yaml
      CONCOURSE_BASE_RESOURCE_TYPE_DEFAULTS: /src/brt_defaults.yml
```

A build is like:

build_log

1. Add a CLI option `--base-resource-type-defaults <brt_defaults.yml>`
2. Add a new section `defaults` to resource type definition.
3. Apply default to its direct child's source as early as possible, so that vars can be used in defaults, and vars can be interpolated in step's run.

## Notes to reviewer:

@vito I don't implement `alias` that you suggested in https://github.com/concourse/concourse/pull/5942#issuecomment-669283962, because I don't see `alias` having any urgent use case. IMO, an admin should be very carefully to configure any defaults for base resource types, but once a default is configured, then it should be enforced. Given DockerHub deadline is approaching, can you please give priority to this PR? Hopefully target to 6.6.0.

## Release Note

Add `defaults` to resource types. With this feature, a cluster admin may configure default sources to based resource types when boots ATC. For example, when the following file is specified to ATC `--base-resource-type-defaults=brt_default.yml`:

```yaml
registry-image:
  registry_mirror:
    host: 192.168.1.12:5000
```

then all `registry-image` will pull images from the mirror `192.168.1.12:5000`. 

At pipeline level, `defaults` can be defined with resource types, for example:

```yaml
resource_types:
- name: rt-gitlab
  type: registry-image
  source:
    repository: <uri of a gitlab resource type>
  defaults:
    url: <gitlab url>
    project: <project-id>

resources:
- name: gitlab-mr
  type: rt-gitlab
  source:
    kind: MergeRequest

- name: gitlab-release
  type: rt-gitlab
  source:
    kind: Release
```

In this sample, when defines the gitlab resource-type, GitLab url and project-id are defined as defaults of the resource type, so that all resources of the type will inherit those defaults.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/base_resource_type_defaults.go
+++ b/atc/base_resource_type_defaults.go
@@ -1,0 +1,14 @@
+package atc
+
+var baseResourceTypeDefaults = map[string]Source{}
+
+func LoadBaseResourceTypeDefaults(defaults map[string]Source) {
+	baseResourceTypeDefaults = defaults
+}
+
+func FindBaseResourceTypeDefaults(name string) (Source, bool) {
+	if source, ok := baseResourceTypeDefaults[name]; ok {
+		return source, true
+	}
+	return nil, false
+}
--- a/atc/db/base_resource_type.go
+++ b/atc/db/base_resource_type.go
@@ -2,6 +2,8 @@ package db
 
 import (
 	"database/sql"
+	"encoding/json"
+	"github.com/concourse/concourse/atc"
 
 	sq "github.com/Masterminds/squirrel"
 )
@@ -13,6 +15,7 @@ import (
 // It is removed by gc.BaseResourceTypeCollector, once there are no references
 // to it from worker_base_resource_types.
 type BaseResourceType struct {
+	Id   int
 	Name string // The name of the type, e.g. 'git'.
 }
 
@@ -26,6 +29,7 @@ type UsedBaseResourceType struct {
 	ID                   int    // The ID of the BaseResourceType.
 	Name                 string // The name of the type, e.g. 'git'.
 	UniqueVersionHistory bool   // If set to true, will create unique version histories for each of the resources using this base resource type
+	Defaults             atc.Source
 }
 
 // FindOrCreate looks for an existing BaseResourceType and creates it if it
@@ -45,14 +49,20 @@ func (brt BaseResourceType) FindOrCreate(tx Tx, unique bool) (*UsedBaseResourceT
 
 func (brt BaseResourceType) Find(runner sq.Runner) (*UsedBaseResourceType, bool, error) {
 	var id int
+	var name string
 	var unique bool
-	err := psql.Select("id, unique_version_history").
-		From("base_resource_types").
-		Where(sq.Eq{"name": brt.Name}).
-		Suffix("FOR SHARE").
+	var defaultsString sql.NullString
+	sb := psql.Select("id, name, unique_version_history, defaults").
+		From("base_resource_types")
+	if brt.Id > 0 {
+		sb = sb.Where(sq.Eq{"id": brt.Id})
+	} else {
+		sb = sb.Where(sq.Eq{"name": brt.Name})
+	}
+	err := sb.Suffix("FOR SHARE").
 		RunWith(runner).
 		QueryRow().
-		Scan(&id, &unique)
+		Scan(&id, &name, &unique, &defaultsString)
 	if err != nil {
 		if err == sql.ErrNoRows {
 			return nil, false, nil
@@ -61,7 +71,15 @@ func (brt BaseResourceType) Find(runner sq.Runner) (*UsedBaseResourceType, bool,
 		return nil, false, err
 	}
 
-	return &UsedBaseResourceType{ID: id, Name: brt.Name, UniqueVersionHistory: unique}, true, nil
+	var defaults atc.Source
+	if defaultsString.Valid {
+		err := json.Unmarshal([]byte(defaultsString.String), &defaults)
+		if err != nil {
+			return nil, false, err
+		}
+	}
+
+	return &UsedBaseResourceType{ID: id, Name: name, UniqueVersionHistory: unique, Defaults: defaults}, true, nil
 }
 
 func (brt BaseResourceType) create(tx Tx, unique bool) (*UsedBaseResourceType, error) {
--- a/atc/db/base_resource_type.go
+++ b/atc/db/base_resource_type.go
@@ -2,8 +2,6 @@ package db
 
 import (
 	"database/sql"
-	"encoding/json"
-	"github.com/concourse/concourse/atc"
 
 	sq "github.com/Masterminds/squirrel"
 )
@@ -15,7 +13,6 @@ import (
 // It is removed by gc.BaseResourceTypeCollector, once there are no references
 // to it from worker_base_resource_types.
 type BaseResourceType struct {
-	Id   int
 	Name string // The name of the type, e.g. 'git'.
 }
 
@@ -29,7 +26,6 @@ type UsedBaseResourceType struct {
 	ID                   int    // The ID of the BaseResourceType.
 	Name                 string // The name of the type, e.g. 'git'.
 	UniqueVersionHistory bool   // If set to true, will create unique version histories for each of the resources using this base resource type
-	Defaults             atc.Source
 }
 
 // FindOrCreate looks for an existing BaseResourceType and creates it if it
@@ -49,20 +45,14 @@ func (brt BaseResourceType) FindOrCreate(tx Tx, unique bool) (*UsedBaseResourceT
 
 func (brt BaseResourceType) Find(runner sq.Runner) (*UsedBaseResourceType, bool, error) {
 	var id int
-	var name string
 	var unique bool
-	var defaultsString sql.NullString
-	sb := psql.Select("id, name, unique_version_history, defaults").
-		From("base_resource_types")
-	if brt.Id > 0 {
-		sb = sb.Where(sq.Eq{"id": brt.Id})
-	} else {
-		sb = sb.Where(sq.Eq{"name": brt.Name})
-	}
-	err := sb.Suffix("FOR SHARE").
+	err := psql.Select("id, unique_version_history").
+		From("base_resource_types").
+		Where(sq.Eq{"name": brt.Name}).
+		Suffix("FOR SHARE").
 		RunWith(runner).
 		QueryRow().
-		Scan(&id, &name, &unique, &defaultsString)
+		Scan(&id, &unique)
 	if err != nil {
 		if err == sql.ErrNoRows {
 			return nil, false, nil
@@ -71,15 +61,7 @@ func (brt BaseResourceType) Find(runner sq.Runner) (*UsedBaseResourceType, bool,
 		return nil, false, err
 	}
 
-	var defaults atc.Source
-	if defaultsString.Valid {
-		err := json.Unmarshal([]byte(defaultsString.String), &defaults)
-		if err != nil {
-			return nil, false, err
-		}
-	}
-
-	return &UsedBaseResourceType{ID: id, Name: name, UniqueVersionHistory: unique, Defaults: defaults}, true, nil
+	return &UsedBaseResourceType{ID: id, Name: brt.Name, UniqueVersionHistory: unique}, true, nil
 }
 
 func (brt BaseResourceType) create(tx Tx, unique bool) (*UsedBaseResourceType, error) {
--- a/atc/db/job_factory.go
+++ b/atc/db/job_factory.go
@@ -62,13 +62,6 @@ func (resources SchedulerResources) Lookup(name string) (SchedulerResource, bool
 }
 
 func (j *jobFactory) JobsToSchedule() (SchedulerJobs, error) {
-	tx, err := j.conn.Begin()
-	if err != nil {
-		return nil, err
-	}
-
-	defer tx.Rollback()
-
 	rows, err := jobsQuery.
 		Where(sq.Expr("j.schedule_requested > j.last_scheduled")).
 		Where(sq.Eq{
@@ -76,7 +69,7 @@ func (j *jobFactory) JobsToSchedule() (SchedulerJobs, error) {
 			"j.paused": false,
 			"p.paused": false,
 		}).
-		RunWith(tx).
+		RunWith(j.conn).
 		Query()
 	if err != nil {
 		return nil, err
@@ -90,7 +83,7 @@ func (j *jobFactory) JobsToSchedule() (SchedulerJobs, error) {
 	var schedulerJobs SchedulerJobs
 	pipelineResourceTypes := make(map[int]ResourceTypes)
 	for _, job := range jobs {
-		rows, err := tx.Query(`WITH inputs AS (
+		rows, err := j.conn.Query(`WITH inputs AS (
 				SELECT ji.resource_id from job_inputs ji where ji.job_id = $1
 				UNION
 				SELECT jo.resource_id from job_outputs jo where jo.job_id = $1
@@ -147,7 +140,7 @@ func (j *jobFactory) JobsToSchedule() (SchedulerJobs, error) {
 			rows, err := resourceTypesQuery.
 				Where(sq.Eq{"r.pipeline_id": job.PipelineID()}).
 				OrderBy("r.name").
-				RunWith(tx).
+				RunWith(j.conn).
 				Query()
 			if err != nil {
 				return nil, err
@@ -173,11 +166,6 @@ func (j *jobFactory) JobsToSchedule() (SchedulerJobs, error) {
 			Resources:     schedulerResources,
 			ResourceTypes: resourceTypes.Deserialize(),
 		})
-	}
-
-	err = tx.Commit()
-	if err != nil {
-		return nil, err
 	}
 
 	return schedulerJobs, nil
--- a/atc/db/job_factory.go
+++ b/atc/db/job_factory.go
@@ -51,17 +51,31 @@ type SchedulerResource struct {
 	Source atc.Source
 }
 
-func (resources SchedulerResources) Lookup(name string) (SchedulerResource, bool) {
+func (r *SchedulerResource) ApplySourceDefaults(resourceTypes atc.VersionedResourceTypes) {
+	parentType, found := resourceTypes.Lookup(r.Type)
+	if found {
+		r.Source = parentType.Defaults.Merge(r.Source)
+	}
+}
+
+func (resources SchedulerResources) Lookup(name string) (*SchedulerResource, bool) {
 	for _, resource := range resources {
 		if resource.Name == name {
-			return resource, true
+			return &resource, true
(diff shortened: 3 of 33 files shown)


atc/db/job_factory.go
```diff
@@ -62,21 +62,14 @@ func (resources SchedulerResources) Lookup(name string) (SchedulerResource, bool
 }
 
 func (j *jobFactory) JobsToSchedule() (SchedulerJobs, error) {
-	tx, err := j.conn.Begin()
```
evanchaoli, line 65: Hi @clarafu, I think you are the DB expert. Here I removed the transaction because it cause the db query to hang at https://github.com/concourse/concourse/pull/6079/files#diff-61e6f1d527ec479d4312e73f52b3f93aR99 for unit test. But I want to understand why transaction is used here, this function does read-only, how a transaction helps?

atc/db/resource_type.go
```diff
@@ -87,11 +90,24 @@ func (resourceTypes ResourceTypes) Deserialize() atc.VersionedResourceTypes {
 	var versionedResourceTypes atc.VersionedResourceTypes
 
 	for _, t := range resourceTypes {
+		// Apply source defaults to resource types
+		source := t.Source()
+		parentType, found := resourceTypes.Parent(t)
+		if found {
+			source = parentType.Defaults().Merge(source)
+		} else {
+			parentType, found, _ := t.FindParentBaseResourceType()
```
vito, line 99: This `err` should not be ignored - an error reaching the database shouldn't mean that defaults are skipped

atc/db/resource_type.go
```diff
@@ -87,11 +90,24 @@ func (resourceTypes ResourceTypes) Deserialize() atc.VersionedResourceTypes {
...
+			if found {
+				source = parentType.Defaults.Merge(source)
```
chenbh, line 101: Shadowing `parentType` confused me as to why it was `parentType.Defaults` in one place and `parentType.Defaults()` in another
		
```suggestion
			parentUbrt, found, _ := t.FindParentBaseResourceType()
			if found {
				source = parentUbrt.Defaults.Merge(source)
```

vito:
Thanks for the PR! Don't have time for a full review but the direction looks good - being able to adjust the defaults with `fly` seems nice, though I have some questions about it.

Had you given any thought to configuring defaults statically, e.g. `concourse web --resource-type-defaults foo.yml`? Just wondering how important static vs dynamic is since storing it in the database and having a `fly` command is more complex.

For example what happens when you try to set defaults before any workers are registered? Base resource types are created in the database when a worker registers with it - will setting defaults for a type fail until a worker creates it? If so, that might be a little awkward to automate cluster setup. Maybe it could just create the row? That would be little awkward since we don't know the `unique_version_history` value, but I think that gets upserted anyway.

Also what happens when all the workers go away? If the base resource types go away and come back, the defaults would need to be re-set with another `fly` command. I don't remember if base resource types are ever removed though - maybe this isn't an issue, but it does feel a little weird to have cluster configuration somewhat tied to worker lifecycle.

Static configuration has the obvious downside of needing a `web` node restart, but it doesn't seem like these defaults would change all that often, and having it be part of configuration would mean there's less bootstrapping to do in order to get a working cluster. The restart could also be avoided if we implement `SIGHUP` to reload configuration. (We'd like to switch everything to a single config file so this behavior would be more uniform - see https://github.com/concourse/concourse/discussions/5891.)

> @vito I don't implement `alias` that you suggested in [#5942 (comment)](https://github.com/concourse/concourse/pull/5942#issuecomment-669283962), because I don't see `alias` having any urgent use case. IMO, an admin should be very carefully to configure any defaults for base resource types, but once a default is configured, then it should be enforced. Given DockerHub deadline is approaching, can you please give priority to this PR? Hopefully target to 6.6.0.

Sounds good - I can't promise this will make it in to v6.6.0 since it's already past due, but v6.6 will at least have one other mitigation which should help: `check` will use HEAD requests which don't count towards the rate limit (https://github.com/concourse/registry-image-resource/pull/241 & https://github.com/concourse/docker-image-resource/pull/300).

Thanks again for picking this up!

atc/db/base_resource_type.go
```diff
@@ -45,14 +49,20 @@ func (brt BaseResourceType) FindOrCreate(tx Tx, unique bool) (*UsedBaseResourceT
...
-	err := psql.Select("id, unique_version_history").
-		From("base_resource_types").
-		Where(sq.Eq{"name": brt.Name}).
-		Suffix("FOR SHARE").
+	var defaultsString sql.NullString
+	sb := psql.Select("id, name, unique_version_history, defaults").
+		From("base_resource_types")
+	if brt.Id > 0 {
+		sb = sb.Where(sq.Eq{"id": brt.Id})
+	} else {
```
chenbh, line 59: As far as I can tell, this table always starts index at 1. What prompted this change?
https://github.com/concourse/concourse/blob/64a12a7023f7a68993730c3add624e38773eb9fd/atc/db/migration/migrations/1510262030_initial_schema.up.sql#L72-L77

chenbh (changes requested):
You'll also need to make changes to the [`exec` package](https://github.com/concourse/concourse/blob/master/atc/exec/task_step.go#L158-L171) to handle tasks that are defined via files
```yaml
resources:
- name: resource
  type: mock
  source:
    create_files:
      task.yml:
        platform: linux
        image_resource:
          type: alpine
        run:
          path: echo
          args: ["hello world"]

resource_types:
- name: alpine
  type: registry-image
  source:
    repository: concourse/registry-image-resource
  defaults:
    repository: alpine

jobs:
- name: foo
  plan:
  - get: resource
  - task: with-path
    file: resource/task.yml
```

chenbh:
I also don't see any changes for `fly` here, are they supposed to be part of a follow up commit? Otherwise looks good and works like a charm

evanchaoli:
> Had you given any thought to configuring defaults statically, e.g. `concourse web --resource-type-defaults foo.yml`? Just wondering how important static vs dynamic is since storing it in the database and having a `fly` command is more complex.
>

Yes, I considered static defaults. My decision was actually based on our OPs model. `fly` is more flexible for OPs. To restart a prod cluster, it would require a maintenance, needing to go through an approval process, announcing the maintenance to users N day ahead, etc., while `fly` may avoid such process. And if some mistake is made, `fly` can quickly fix the mistake. But at this point, I don't have strong opinion on `fly` as I can foresee that base resource type defaults should be used in very little cases, maybe only for the dockerhub issue. (Pipeline level resource type defaults could be used more widely.)

But I agree, from coding perspective, static configure is easier: avoiding changes on fly and db, thus less tests to add.
 
> For example what happens when you try to set defaults before any workers are registered? Base resource types are created in the database when a worker registers with it - will setting defaults for a type fail until a worker creates it? If so, that might be a little awkward to automate cluster setup. Maybe it could just create the row? That would be little awkward since we don't know the `unique_version_history` value, but I think that gets upserted anyway.
> 

I guess this is not a problem, because as we use `fly`, which implies `defaults` should be tuned after cluster is installed. Thus we should suggest automation cluster setup to not include this defaults configure.

At least, I never planned to include this in our automation setup.

> Also what happens when all the workers go away? If the base resource types go away and come back, the defaults would need to be re-set with another `fly` command. I don't remember if base resource types are ever removed though - maybe this isn't an issue, but it does feel a little weird to have cluster configuration somewhat tied to worker lifecycle.
> 

I didn't verify if `base_resource_types` will be cleaned up when all workers gone. But if that really matters, we can just store `defaults` in a separate table, say create a new table `base_resource_type_defaults`.

If switching to static configure can speed up this PR, make it more possible in 6.6, I'll be happy to change. 😃  My original plan was to make `fly` change today. Now I decide to switch to static configure. So I think I can complete all coding work including adding unit tests for this PR today.

atc/db/resource_type.go
evanchaoli: With the second commit, this is no longer an issue.

atc/db/job_factory.go
evanchaoli: With the second commit, this is no longer an issue.

atc/db/resource_type.go
evanchaoli: With the second commit, this is no longer an issue.

atc/db/base_resource_type.go
evanchaoli: With the second commit, this is no longer an issue.

evanchaoli:
> You'll also need to make changes to the [`exec` package](https://github.com/concourse/concourse/blob/master/atc/exec/task_step.go#L158-L171) 

Good catch! Fixed in the second commit.

evanchaoli:
@vito @chenbh In the second commit, I have done the following:

- Changed to do static configure
- Addressed review comments
- Refactored to avoid duplicate code

But unfortunately I have no time to add unit tests. I will add add unit tests tomorrow. Please go ahead to review the code change.

vito:
@evanchaoli Cool, thanks for adjusting! I think that'll make it easier to review.

> Yes, I considered static defaults. My decision was actually based on our OPs model. `fly` is more flexible for OPs. To restart a prod cluster, it would require a maintenance, needing to go through an approval process, announcing the maintenance to users N day ahead, etc., while `fly` may avoid such process. And if some mistake is made, `fly` can quickly fix the mistake. But at this point, I don't have strong opinion on `fly` as I can foresee that base resource type defaults should be used in very little cases, maybe only for the dockerhub issue. (Pipeline level resource type defaults could be used more widely.)

Yeah true, once you have established processes around configuration which requires redeploying, being able to change it "at runtime" is a lot more lightweight.

Maybe in the future this could be configurable at the project-level and inherited by all sub-projects. 🤔 Then it could be set in the "root" project and effectively be set cluster-wide without requiring a config change / deploy. We'd have to figure out how we want inheritance to work though - opt-in is the safest approach, but in this case it's not really what you want.

> I guess this is not a problem, because as we use `fly`, which implies `defaults` should be tuned after cluster is installed. Thus we should suggest automation cluster setup to not include this defaults configure.

IMO you want to automate as much as possible when it comes to configuring your cluster. This is a tiny detail that would be easily forgotten if it's manual and can't be done immediately. The problem is it's perfectly valid to have a cluster deployed with 0 global/shared workers, and then have the workers come in later. For example worker pools may be created on-demand for each team.

This is all resolved by making the config static, but it could also be resolved by pushing the config into projects in the future so that it's saved as config in a repo and doesn't require manually running any commands. I've left an internal note to consider this in the future.

evanchaoli:
I have tested the code change with a private today:

```yaml
registry-image:
  registry_mirror:
    host: 192.168.1.12:6000
```

All `check`, `get`, `put`, `task` (including task file) steps all pull images from the mirror.

chenbh (commented):
Just one last thing and I think the code changes are good to go!

I feel like the release note can be reworded so that it's more about generic defaults. Maybe you can provide an example with a resource-type that has a `registry_mirror` default?

Do you mind also updating the docs for this as well? The new attributes should be added to resources/resource-types and a section about configuring defaults on the web node

- https://github.com/concourse/docs/blob/master/lit/docs/resources.lit#L133
- https://github.com/concourse/docs/blob/master/lit/docs/resource-types.lit#L94
- https://github.com/concourse/docs/blob/master/lit/docs/install/web.lit#L275

atc/task.go
```diff
@@ -51,6 +51,22 @@ type ImageResource struct {
 	Version Version `json:"version,omitempty"`
 }
 
+func (ir *ImageResource) AppSourceDefaults(resourceTypes VersionedResourceTypes) {
```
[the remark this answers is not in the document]
evanchaoli: My bad, good catch. Should be "Apply" rather than "App".

atc/base_resource_type_defaults.go
```diff
@@ -0,0 +1,14 @@
+package atc
+
+var baseResourceTypeDefaults = map[string]Source{}
```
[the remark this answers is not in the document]
evanchaoli: I also struggled on if or not use the global `baseResourceTypeDefaults`. I even made a version that didn't use global vars. But comparing with the non-global way, passing the defaults object to everywhere, the global var way is much simpler. I think even we need to extend functionality to base resource type defaults, we may refactor the global var.

I also had concern on unit test. So I carefully added `AfterEach`, and tried to combine multiple `IT`s into a single `IT`. I also confirmed that, when `ginkgo` runs multiple nodes, tests are running in multiple processes, which made my feel more comfortable.

evanchaoli:
> Just one last thing and I think the code changes are good to go!
> 
> I feel like the release note can be reworded so that it's more about generic defaults. Maybe you can provide an example with a resource-type that has a `registry_mirror` default?
> 
> Do you mind also updating the docs for this as well? The new attributes should be added to resources/resource-types and a section about configuring defaults on the web node
> 
>     * https://github.com/concourse/docs/blob/master/lit/docs/resources.lit#L133
> 
>     * https://github.com/concourse/docs/blob/master/lit/docs/resource-types.lit#L94
> 
>     * https://github.com/concourse/docs/blob/master/lit/docs/install/web.lit#L275

@chenbh I'll update the release note with some details. But the docs, usually I only contributed code changes, and concourse team worked on docs. I'm willing to help on docs, but I'm not that confident with my English writing skill as my coding skill. 😸

chenbh (approved):
I'll do the docs changes then. lgtm!

chenbh:
TODO:

- [x] create PR for docs (https://github.com/concourse/concourse/pull/6079)
- [ ] create PR for BOSH
- [ ] create PR for helm

chenbh:
@evanchaoli it looks like there's some conflicts resulting from https://github.com/concourse/concourse/pull/6022

atc/exec/task_step.go
```diff
@@ -156,8 +156,11 @@ func (step *TaskStep) run(ctx context.Context, state RunState) error {
 	var taskVars []vars.Variables
 
 	if step.plan.ConfigPath != "" {
-		// external task - construct a source which reads it from file
-		taskConfigSource = FileConfigSource{ConfigPath: step.plan.ConfigPath, Client: step.workerClient}
+		// external task - construct a source which reads it from file, and apply base resource type defaults.
+		taskConfigSource = BaseResourceTypeDefaultsApplySource{
+			ConfigSource:  FileConfigSource{ConfigPath: step.plan.ConfigPath, Client: step.workerClient},
+			ResourceTypes: step.plan.VersionedResourceTypes,
+		}
```
agurney, line 163: I noticed in testing that this change does not apply to builds run with `fly execute`. In that case, the plan is already constructed (by fly, which doesn't know about the base resource type default settings), so we are in the `else` branch below. Locally, I changed line 177 below
```go
		taskConfigSource = StaticConfigSource{Config: step.plan.Config}
```
to
```go
		taskConfigSource = BaseResourceTypeDefaultsApplySource{
			ConfigSource: StaticConfigSource{Config: step.plan.Config},
			ResourceTypes: step.plan.VersionedResourceTypes,
		}
```
and this resulted in fly executed builds having the defaults applied, as expected. I did not check to see if this broke anything else, yet. 

I think it would be convenient to have parity between one-off builds and regular pipeline builds, as it may be confusing if they behave in subtly different ways - not limited to registry mirror settings. It's not about whether the volume of one-off builds potentially not using a configured mirror is problematic, but just the fact that different behavior is undesirable.

agurney (changes requested):

atc/exec/task_step.go
evanchaoli: Good catch! Your comment made me think over again. I think it makes sense to apply defaults after interpolate vars, because image_source itself could be defined by vars.

evanchaoli:
@chenbh I have rebased the code and resolved conflicts. Also addressed @agurney's comment. I have no time today to run end-to-end tests, will run all tests, including `fly exec` tomorrow.

evanchaoli:
@chenbh I have tested again the code, `check`, `get`, `put`, embedded task, task file, fly-exec, they all applied defaults.

vito:
Thanks!

PR #6085, opened 2020-09-23
Add `/pipelines/:pipeline_id/...` API endpoints

aoldershaw:

Add a `/pipelines/:pipeline_id/...` API as an alternative to the `/teams/:team_name/pipelines/:pipeline_name/...` API (which hasn't been removed!). This API will be consumed from the web UI after instanced pipelines are added, as it's more convenient than dealing with team name + pipeline name + instance vars.

It's also a good idea to break the hierarchy between teams/pipelines if we end up [replacing teams with projects](https://github.com/vito/rfcs/blob/projects/032-projects/proposal.md#can-projects-replace-teams).

This is also technically a breaking change as it removes the `/pipelines/:pipeline_name/jobs/:job_name/badge` endpoint, which has been around for a long time as a fallback for when teams were first introduced (it just redirects to the main team's badge endpoint). It's been several years now, though, so hopefully it doesn't have too much impact.

## Notes to reviewer:

This is marked as a draft as it currently breaks `go-concourse` and `fly`

This also doesn't expose the API via `go-concourse` - I don't have a need for this yet, as `fly` is still using the old endpoints (at least for the time being)

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


atc/api/auth/check_authorization_handler.go
```diff
@@ -30,6 +31,13 @@ func (h checkAuthorizationHandler) ServeHTTP(w http.ResponseWriter, r *http.Requ
 	}
 
 	teamName := r.URL.Query().Get(":team_name")
+	if teamName == "" {
+		pipeline, ok := r.Context().Value(PipelineContextKey).(db.Pipeline)
+		if !ok {
+			panic("missing pipeline")
+		}
+		teamName = pipeline.TeamName()
+	}
```
jamieklassen, line 40: something feels fishy to me about this -- it seems like this code is a bit obsessed with the implementation details of the fetchpipelinewrappa. It's not a huge deal, but aesthetically it feels funny. Would it make sense to have a fetchteamwrappa so that there's a uniform way to get the team from a requestcontext as well?

jamieklassen (commented):
There is some great code cleaning in here -- i love the removal of the main job badge stuff, very nice to see the redundant teamName parameter getting removed from the present functions, and removing a lot of duplicated "fetch a pipeline" logic is great.

I'm sensing some awkwardness around these rata and wrappa constructs, and I'm gonna try to share some wild thoughts without thinking them through too carefully, so that we can start the conversation early.

rata-wise: it feels like we have an abstract idea of a "scope" (like a set of endpoints associated to a group of logically-connected resources, in this case sub-objects/operations for a given pipeline) that can be addressed at multiple "base paths" (i.e. `/api/v1/pipelines/:id` or `/api/v1/teams/:team/pipelines/:pipeline`). I'd like to see this reflected in the code to avoid duplication -- e.g. it would be cool to be able to compose "routers" so that instead of a route pointing straight to a handler, it could point to another router. but rata is very "static" so you end up having to duplicate so many router entries... I suppose you could at least "dynamically" build up your list of routes like
```
for each pipeline-scoped route {
  append the "by pipeline ID" route
  append the "by name" route
}
```

wrappa-wise: I feel like the multiwrappa idea is a cool capability, but it seems to be more flexible than we need! my sense is that we're using wrappas to implement a kind of decorator pattern and we want to be able to chain decorators. however having an earlier decorator in the chain just write a value to the request context then forces later decorators to handle cases where the value is missing. I feel like there might be a more static/typed way to do this that avoids the panics, but I haven't thought it through very carefully...

atc/api/pipelineserver/scoped_handler_factory.go
```diff
@@ -8,45 +8,13 @@ import (
...
-			if err != nil {
-				w.WriteHeader(http.StatusInternalServerError)
-				return
-			}
-
-			if !found {
-				w.WriteHeader(http.StatusNotFound)
-				return
-			}
+			panic("missing pipeline")
```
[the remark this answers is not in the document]
aoldershaw: > In this case I imagine the server would just hang up, and clients would get an EOF error

Yup, exactly

> is there an automated test that would catch this sort of thing anyway

Not that I can think of generically - but any unit test that exercises the endpoint should fail if it's not configured properly

atc/api/auth/check_authorization_handler.go
aoldershaw: It is definitely fishy, but I was struggling to come up with a better solution.

One approach I had was to modify the `check_pipeline_access_handler` to support the case where we allow/disallow public pipelines, and use that for these endpoints instead, which is probably a slight improvement - although, I do the same awkward thing in `policychecker/checker.go`, which currently runs for every endpoint.

As you pointed out, I think the bigger problem is the implicit passing of data through the request context, since we lose all type safety. I'll try to think of alternative approaches, but definitely let me know if you can come up with anything

agurney:
Without having seen the future UI code that would invoke these, I wonder whether it's really needed to have an entire parallel version of all the pipeline-related API calls, vs. changing the existing ones to accept an ID as well as a name. So instead of adding `/pipelines/:pipeline_id/jobs/:job_name/schedule` alongside `/teams/:team_name/pipelines/:pipeline_name/jobs/:job_name/schedule`, why not change the latter paths to `/teams/:team_name/pipelines/:pipeline_name_or_id/jobs/:job_name/schedule` ? 

It can almost surely be assumed that a pipeline name won't consist only of digits (perhaps allowed historically, but forbidden by the current `identifier` schema, and I haven't seen those in practice). It feels like that would be a simpler adaptation, that would be more maintainable for future changes to the pipeline-related API which would otherwise touch multiple routes.

I imagine that in any invoking context, we already know the team name. I don't think it's a given that teams are going away, but if they do then the whole API needs to be revisited anyway, along with much else.

aoldershaw:
@agurney IMO `/teams/:team_name/pipelines/pipeline_name_or_id` is kind of an awkward API. It feels weird having the `:team_name` in the path, given that the pipeline ID alone is enough information to uniquely identify a pipeline - we could choose to error if the `:team_name` doesn't match the pipeline, or could ignore it.

I think you're right about maintainability concerns related to having multiple parallel endpoints with the current approach taken, but as @jamieklassen pointed out, there are ways we can do this without needing to update both routes every time. The simplest way would be the suggestion of programmatically adding routes for each pipeline endpoint. Perhaps a better way would be to using a smarter router - for instance, using gorilla/mux or similar, we could define a `pipelineRouter` (that would define a handler for e.g. `/jobs/:job_name/schedule`), and then handle both the `/teams/:team_name/pipelines/:pipeline_name` and `/pipelines/:pipeline_id` prefixes using the `pipelineRouter`.

aoldershaw:
Going to close this - changes to #6105 make it so this set of endpoints isn't required, and this was a slightly hacky approach anyway.

PR #6087, opened 2020-09-25
Only set Guardian flag default when flag not provided by user

muntac:

Bug Fix

CI currently breaking on this [test](https://github.com/concourse/concourse/blob/64a12a7023f7a68993730c3add624e38773eb9fd/topgun/k8s/garden_config_test.go#L61-L62). `config.ini` values get overridden by default values for Guardian.

Move `gdn` binary specific flags to own struct called, `GdnBinaryFlags`.

Using go-flags to set Concourse specific defaults for gdn binary flags overrides any value the user may pass via the config.ini file for Guardian. These are now set manually when the values are not present in the config and the defaults are still stated in the command description.

Following conditions are met in the order given
1. Sets GdnBinaryFlag when it has been set by user either via env or CLI option
2. Does nothing if flag is present in config.ini
3. Sets default value

There might be a simpler way to do this in the future using the go-flags IniParser that Guardian uses to parse the config.ini.

## Notes to reviewer:

To test the 3 scenarios:
1. `docker-compose up`
2. `docker exec -it concourse_worker_1 bash`
3. `ps auxww`
4. Check if the gdn process is running with the correct flags

**Scenario 1:** No config, no flags i.e. vanilla docker-compose
Result: max-containers set to 250 and network-pool set to `10.80.0.0/16`

**Scenario 2:** Config file provided but no flags
Example config file (at `/tmp/config.ini` on local):
```
[server]
max-containers=300
network-pool=10.80.0.0/16
```

Add a volume to `worker` section in `docker-compose.yml`
 
```
worker:
    ....
   volumes:
         - /tmp/config.ini:/concourse-config/config.ini
    environment:
         CONCOURSE_GARDEN_CONFIG: "/concourse-config/config.ini"
```

Result: max-containers and network-pool not set

**Scenario 3:** Config file provided AND flags provided (alternatively can also take out config file)
Same as before and add following to worker.environment section of `docker-compose.yml`
```
CONCOURSE_GARDEN_MAX_CONTAINERS: 500
CONCOURSE_GARDEN_NETWORK_POOL: 10.80.0.0/10
```

Result:  max-containers set to 500 and network-pool set to `10.80.0.0/10`

## Release Note
Only pass garden the configured defaults within Concourse for the guardian flags `max-containers` and `network-pool` if it is not set through the garden config file, environment variables or flags.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/worker/workercmd/guardian.go
+++ b/worker/workercmd/guardian.go
@@ -107,8 +107,9 @@ func (cmd *WorkerCommand) guardianRunner(logger lager.Logger) (ifrit.Runner, err
 
 	gdnArgs := append(gdnConfigFlag, append([]string{"server"}, gdnServerFlags...)...)
 
+	var configFlags GdnBinaryFlags
 	if cmd.Guardian.Config != "" {
-		configFlags, err := getGdnFlagsFromConfig(cmd.Guardian.Config.Path())
+		configFlags, err = getGdnFlagsFromConfig(cmd.Guardian.Config.Path())
 		if err != nil {
 			return nil, err
 		}
--- a/worker/workercmd/guardian.go
+++ b/worker/workercmd/guardian.go
@@ -3,9 +3,7 @@
 package workercmd
 
 import (
-	"bufio"
 	"fmt"
-	"github.com/concourse/flag"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -18,6 +16,7 @@ import (
 	concourseCmd "github.com/concourse/concourse/cmd"
 	"github.com/tedsuo/ifrit"
 	"github.com/tedsuo/ifrit/grouper"
+	flags "github.com/jessevdk/go-flags"
 )
 
 // Guardian binary flags - these are passed along as-is to the gdn binary as options.
@@ -27,15 +26,21 @@ import (
 // This is needed in order to avoid unintentional overrides of user values set in the optional config file.
 // See getGdnFlagsFromStruct for details.
 type GdnBinaryFlags struct {
-	Config         flag.File     `long:"config"     description:"Path to a config file to use for the Garden backend. e.g. 'foo-bar=a,b' for '--foo-bar a --foo-bar b'."`
-	MaxContainers  *int          `long:"max-containers" description:"Maximum container capacity. 0 means no limit. (default:250)"`
-	NetworkPool    string        `long:"network-pool" description:"Network range to use for dynamically allocated container subnets. (default:10.80.0.0/16)"`
+	Server struct {
+		Network struct {
+			Pool string `long:"network-pool" description:"Network range to use for dynamically allocated container subnets. (default:10.80.0.0/16)"`
+		} `group:"Container Networking"`
+
+		Limits struct {
+			MaxContainers *int `long:"max-containers" description:"Maximum container capacity. 0 means no limit. (default:250)"`
+		} `group:"Limits"`
+	} `group:"server"`
 }
 
 // Defaults for GdnBinaryFlags
 const (
 	defaultGdnMaxContainers = "250"
-	defaultGdnNetworkPool = "10.80.0.0/16"
+	defaultGdnNetworkPool   = "10.80.0.0/16"
 )
 
 // This prepares the Guardian runtime using the gdn binary.
@@ -54,8 +59,8 @@ func (cmd *WorkerCommand) guardianRunner(logger lager.Logger) (ifrit.Runner, err
 
 	gdnConfigFlag := []string{}
 
-	if cmd.Guardian.BinaryFlags.Config.Path() != "" {
-		gdnConfigFlag = append(gdnConfigFlag, "--config", cmd.Guardian.BinaryFlags.Config.Path())
+	if cmd.Guardian.Config.Path() != "" {
+		gdnConfigFlag = append(gdnConfigFlag, "--config", cmd.Guardian.Config.Path())
 	}
 
 	gdnServerFlags := []string{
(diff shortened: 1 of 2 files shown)


worker/workercmd/guardian.go
```diff
@@ -152,3 +177,64 @@ func detectGuardianFlags(logger lager.Logger) []string {
...
+		file, err := os.Open(configPath)
+		if err != nil {
+			return []string{}, err
+		}
+		defer file.Close()
+		bs := bufio.NewScanner(file)
+		for bs.Scan() {
+			line := bs.Text()
+
+			if len(line) == 0 || line[0] == '#' {
```
clarafu, line 194: Maybe we can also add a case for the semicolon being a comment line?

muntac: Good catch! Done.

worker/workercmd/guardian.go
```diff
@@ -152,3 +177,64 @@ func detectGuardianFlags(logger lager.Logger) []string {
...
+		file, err := os.Open(configPath)
+		if err != nil {
+			return []string{}, err
+		}
+		defer file.Close()
+		bs := bufio.NewScanner(file)
+		for bs.Scan() {
+			line := bs.Text()
+
+			if len(line) == 0 || line[0] == '#' || line[0] == ';' {
```
clarafu, line 194: One more thing I was thinking of is that it might not be the first character of the line that is a comment character, it can be indented or spaced in too. So maybe we should trim the spaces before we check for the comment character? Sorry I'm not sure if this is not really a reasonable concern though.

aoldershaw: Yeah, I think we should `TrimSpace` to match `go-flags`.

https://github.com/jessevdk/go-flags/blob/c17162fe8fd74f119ff938c5c67af63e3bac5ded/ini.go#L400

What's concerning to me is that we're only considering the `--long-flag` form of the value, whereas in `go-flags`, users can specify it in multiple ways. From the [godocs](https://godoc.org/github.com/jessevdk/go-flags#IniParser):

```
1. Compared to the ini-name tag on the option struct field (if present)
2. Compared to the struct field name
3. Compared to the option long name (if present)
4. Compared to the option short name (if present)
```

Is there any way we could `go-flags.IniParser` directly , rather than doing it ourselves?

muntac: I hear your concerns and understand that string parsing when we are trying to extract values for functional purposes needs a lot of consideration. But the scenario here is different.

All we are trying to do is see _if the words "max-containers" and "network-pool" are present as the first argument on a line_. This is so we know to not supply our own default values for them. We aren't even concerned with what values they have in the file.

Given all that, comments and even other flags showing up in the list do not matter. `max-containers` and `network-pool` both in the structs here and in Guardian only have a "long" name listed in the CLI help. Therefore, the user should not be trying to supply any other types of names - ini, struct, short.

muntac: That said, I'd prefer to use the IniParser as well but was not able to figure it out. I'm happy to take suggestions on how to use it but do not feel it's worth pursuing given the overall time the larger issue of containerd being blocked on BOSH has been taking.

aoldershaw: I think you're probably right that people won't use the `struct` version of the flags (and there are no `short` or `ini` versions).

One way that could possibly work (unverified) is to create a "mirror" of the [`GdnCommand` struct](https://github.com/cloudfoundry/guardian/blob/3837782d5e8fb8b7374e2b9109860690db15a5da/guardiancmd/command.go#L70) that contains only the fields we care about. e.g.

```go
type GdnCommand struct{
	ServerCommand  struct{
		Network struct{
			Pool string `long:"network-pool"`
		} `group:"Container Networking"

		Limits struct{
			MaxContainers *uint64 `long:"max-containers"`
		} `group:"Limits"
	} `command:"server"`
}
```

and then parsing into that struct:

```go
if *configFilePath != "" {
	var cmd GdnCommand
	parser := flags.NewParser(&cmd, flags.Default)
	parser.NamespaceDelimiter = "-"
	iniParser := flags.NewIniParser(parser)
	err := iniParser.ParseFile(*configFilePath)
	...
}
```

clarafu: Hm maybe @muntac if you don't have time, we could take over the PR and try to get it to use the iniParser?

muntac: @clarafu sure, I'd appreciate that.

clarafu: Okay I pushed to the branch! If @muntac and @aoldershaw can take a look at it that would be great! I ran the acceptance testing described in the PR description against the changes and it seems to work fine, but it might be smart if you can take a quick look at it too.

aoldershaw:
@clarafu I tried running scenario 1 and I'm not seeing the expected flags being applied

clarafu:
@aoldershaw Oh ya you're totally right, I moved it in there but that was my mistake. I'll change that now!

worker/workercmd/guardian.go
```diff
@@ -102,13 +107,14 @@ func (cmd *WorkerCommand) guardianRunner(logger lager.Logger) (ifrit.Runner, err
 
 	gdnArgs := append(gdnConfigFlag, append([]string{"server"}, gdnServerFlags...)...)
 
-
-	flagsInConfig, err := getGdnFlagsFromConfig(cmd.Guardian.BinaryFlags.Config.Path())
-	if err != nil {
-		return nil, err
+	if cmd.Guardian.Config != "" {
+		configFlags, err := getGdnFlagsFromConfig(cmd.Guardian.Config.Path())
```
aoldershaw, line 111: https://ci.concourse-ci.org/builds/2881001#L5f6df0f7:67

```suggestion
	var configFlags GdnBinaryFlags
	if cmd.Guardian.Config != "" {
		configFlags, err = getGdnFlagsFromConfig(cmd.Guardian.Config.Path())
```

aoldershaw (changes requested):

worker/workercmd/guardian.go
clarafu: 😅 thats why u should always build before pushing.. haha

aoldershaw (approved):
Looks good to me!

clarafu:
+1 the repro was super helpful!

muntac:
@clarafu @aoldershaw amazing work on this! This looks so much better. Thanks for putting in the time to do this the right way :pray:

PR #39, opened 2015-05-16
Documentation changes

zachgersh:
i made a mistake in the last pull-request.

Reopened here.  Right sort of changes.

Sorry @vito

Issue #5229, opened 2020-02-24
Extensible var sources


vito:
Just as `resource_types:` extends the supported types under `resources:`, we should provide a way to extend the supported types under `var_sources:` so that credential managers and other types of var sources may be implemented externally.

## Current proposal

Prerequisites:

* [x] concourse/rfcs#37

Come up with an interface that [prototypes](https://github.com/concourse/rfcs/pull/37) can implement so that they can be used as a var source:

```yaml
prototypes:
- name: vault-var-source
  type: registry-image
  source: {repository: concourse/vault-var-source-type}

var_sources:
- name: my-vault
  type: vault-var-source
  source: {...}

# ...
```

We must also ensure that the use of prototypes is secure and doesn't compromise the intent of using a credential manager in the first place.

WesleyHindle:
Are there any plans for taking this feature out of experimental?

PR #6115, opened 2020-10-05
atc: behavior: Implement support for Vault KV v2 backends

### LINKED ISSUE #2374: Enable to versioning credential with vault.
# Feature Request

I and my company users want to use vault's versioning function.
Vault has versioning function enabled from v0.10.0(https://github.com/hashicorp/vault/pull/4155).

But the version of vault currently used by this product is 0.9.6.

## A Modest Proposal
1. Upgrade vault
2. When using kv(ver2) as secret backend of vault, the path to read changes from `concourse/team/foo` to `concourse/data/team/foo`
3. In order to maintain compatibility, it is necessary to change the path to be read by secret backend.

daviddob:

Enables support for Vault's KV version 2 backends to enable versioning
and support default deployments of new vault instances. This implements a backwards
compatible change to detect kv2 backends and resolve the paths accordingly.

While RFC #39 was proposed to take this on it appears to still propose the use of a
`vault` type. As such hopefully this is not considered a `quick-fix` antipattern and
can be used to improve the credential-manager var sources as well.

Tested with all gingko + testflight cases.

Fixes #2374

## Notes to reviewer:

Leverages the existing vault test suite as the changes are made to the underlying api client.

## Contributor Checklist

- [x] [Signed] all commits
- [x] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


vito (changes requested):
Hey, thanks for picking this up!

It's unfortunate that we have to carry over so much internal code from Vault, but I get it, not sure what else to do there besides requesting upstream API changes.

There's a part of the code that mutates the client that I'm not sure will be safe though.

atc/creds/vault/vault_kvhelpers.go
```diff
@@ -0,0 +1,100 @@
...
+
+func kvPreflightVersionRequest(client *api.Client, path string) (string, int, error) {
+	// We don't want to use a wrapping call here so save any custom value and
+	// restore after
+	currentWrappingLookupFunc := client.CurrentWrappingLookupFunc()
+	client.SetWrappingLookupFunc(nil)
+	defer client.SetWrappingLookupFunc(currentWrappingLookupFunc)
+	currentOutputCurlString := client.OutputCurlString()
+	client.SetOutputCurlString(false)
+	defer client.SetOutputCurlString(currentOutputCurlString)
```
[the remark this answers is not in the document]
daviddob: Good call - those functions are mainly used for vault cli (OutputCurlString) and the [Cubbyhole Wrapping](https://www.vaultproject.io/docs/concepts/response-wrapping) functions of vault, neither of which concourse would need to use to implement kv_v2 support. Initially I wanted to avoid modifying upstream functions in case they need to be updated in the future, however this makes sense and does not impact functionality from the testing i've done.

vito (commented):
Thanks for waiting - getting back to this now that v6.7 is out!

How would you feel about adding tests for this? Originally the API client was meant to be a thin layer that can be faked out for tests, but now it's getting a little thick. 🤔 There aren't currently tests for the API client, but they could probably be added by faking out the API endpoint using [`ghttp.NewServer`](https://onsi.github.io/gomega/#ghttp-testing-http-clients).

daviddob:
I think adding tests are a solid idea as more complexity gets added - though with it being holiday season my time is a bit limited. I'll see what I can do in the meantime but it may have to sit for a while on my end.

vito:
@daviddob No worries - thanks for checking in!

Just as an FYI, we're working on a pretty big new feature that will allow all this code to be extracted into prototypes. It doesn't need to impact this PR either way, but I thought you might be interested in keeping it on your radar: #5229

daviddob:
@vito Been a little while, finally got some time and finished writing tests using ghttp for both KV2 and KV1 vault apis to test the api_client. The way the ghttp handlers work make it a bit awkward and verbose as the requests much match the order of the handlers but everything seems to be functioning as expected after bringing my changes up to date with master.

vito:
@daviddob Awesome, thanks for circling back to this! Just resolved a tiny conflict, commencing review now.

atc/creds/vault/vault_test.go
```diff
@@ -274,3 +297,687 @@ var _ = Describe("Vault", func() {
 		})
 	})
 })
+
+// The below tests use ghttp handlers to mock a real vault API to the api_client.
+// AppendHandlers has the following behavior which make the tests a bit messy.
```
vito, line 302: Just FYI, an alternative to `AppendHandlers` in situations where you don't care about the order is [`RouteToHandler`](https://pkg.go.dev/github.com/onsi/gomega/ghttp#Server.RouteToHandler).

atc/creds/vault/vault_test.go
```diff
@@ -31,6 +36,24 @@ func (msr *MockSecretReader) Read(lookupPath string) (*vaultapi.Secret, error) {
 	return nil, nil
 }
 
+func createMockV2Secret(value string) *vaultapi.Secret {
+	escapedInput, err := json.Marshal(&value)
+	Expect(err).To(BeNil())
+	secret := vaultapi.Secret{}
+	json.Unmarshal([]byte(fmt.Sprintf(`{"data":{"value":%s},"metadata":{"created_time":"2021-01-06T22:32:10.969537Z","deletion_time":"","destroyed":false,"version":3}}`, escapedInput)), &secret.Data)
```
vito, line 43: Seems like this could be a [`vaultapi.Secret`](https://pkg.go.dev/github.com/hashicorp/vault/api#Secret) literal:

```go
return &vaultapi.Secret{
  Data: map[string]interface{}{
    "data": map[string]interface{}{"value": value},
    "metadata": map[string]interface{}{
      "created_time": "2021-01-06T22:32:10.969537Z",
      "deletion_time": "",
      "destroyed": false,
      "version": 3,
    },
  },
}
```

atc/creds/vault/vault_test.go
```diff
@@ -31,6 +36,24 @@ func (msr *MockSecretReader) Read(lookupPath string) (*vaultapi.Secret, error) {
...
+
+	return &secret
+}
+
+func createMockV1Secret(value string) *vaultapi.Secret {
+	escapedInput, err := json.Marshal(&value)
+	Expect(err).To(BeNil())
+	secret := vaultapi.Secret{}
+	json.Unmarshal([]byte(fmt.Sprintf(`{"value":%s}`, escapedInput)), &secret.Data)
```
vito, line 52: Same feedback as above - mainly just nice to avoid marshaling/unmarshaling.

atc/creds/vault/vault_test.go
```diff
@@ -274,3 +297,687 @@ var _ = Describe("Vault", func() {
...
+	var v *vault.Vault
+	var variables vars.Variables
+	var vaultApi *vault.APIClient
+	var varFoo vars.Reference
+
+	BeforeEach(func() {
+		server = ghttp.NewServer()
+
+		var err error
+		vaultApi, err = vault.NewAPIClient(lagertest.NewTestLogger("test"), "http://"+server.Addr(), vault.TLSConfig{}, vault.AuthConfig{}, "")
```
vito, line 663: to catch up with `master`:

```suggestion
		vaultApi, err = vault.NewAPIClient(lagertest.NewTestLogger("test"), "http://"+server.Addr(), vault.TLSConfig{}, vault.AuthConfig{}, "", 0)
```

atc/creds/vault/vault_test.go
```diff
@@ -274,3 +297,687 @@ var _ = Describe("Vault", func() {
...
```
vito, line 326: to catch up with `master`:

```suggestion
		vaultApi, err = vault.NewAPIClient(lagertest.NewTestLogger("test"), "http://"+server.Addr(), vault.TLSConfig{}, vault.AuthConfig{}, "", 0)
```

vito (changes requested):
Looks good and verified that the tests fail if the v2 support is removed. Left some nitpicky suggestions. I would have just merged and touched them up myself but it looks like some small changes need to be made since `NewAPIClient` changed on `master` recently. (I suggested changes for them.)

Once those `NewAPIClient` calls are fixed up I'm happy to merge this as-is.

atc/creds/vault/vault_test.go
daviddob: Thanks! This is what I was looking for and apparently totally missed. Used it to clean up the mount calls that get made everywhere but left the others to show the proper flow of expected calls.

daviddob:
Updated as per your feedback - should be good to go. Thanks @vito!

vito (approved):
lgtm - thanks!

Issue #6111, opened 2020-10-02
Running a set_pipeline step against a file containing unexpected pipeline YML crashes ATC


mdb:
## Summary

In at least version 6.4.1, Concourse ATCs crash when a [set_pipeline](https://concourse-ci.org/jobs.html#schema.step.set-pipeline-step.set_pipeline) step is provided a YML file containing unexpected pipeline YML, such as a YML file containing a [task definition](https://concourse-ci.org/tasks.html), or even a YML file with no content.

**Update**: It appears this does not crash the ATC in versions >= 6.5.0. So far, I have only reproduced the ATC crash in version 6.4.1. However, in versions greater than 6.5.0, the job still errors and a `runtime error: invalid memory address or nil pointer dereference` is logged. For example, here is output from a `docker-compose up` performed in `master` (at commit `b5584f13bfe700b1c663a97f2e76afc9d406cd4a`):

```
web_1     | {"timestamp":"2020-10-04T11:13:46.010601527Z","level":"debug","source":"atc","message":"atc.tracker.notify.run.set-pipeline-step.starting","data":{"build":1,"job":"update-pipeline
-with-task-yml","job-id":1,"pipeline":"concourse-set-pipeline-dos","session":"24.24.2.4","step-name":"bad-pipeline"}}
web_1     | {"timestamp":"2020-10-04T11:13:46.011899176Z","level":"debug","source":"atc","message":"atc.tracker.notify.run.set-pipeline-step.no-diff","data":{"build":1,"job":"update-pipeline-
with-task-yml","job-id":1,"pipeline":"concourse-set-pipeline-dos","session":"24.24.2.4","step-name":"bad-pipeline"}}
web_1     | {"timestamp":"2020-10-04T11:13:46.014093778Z","level":"error","source":"atc","message":"atc.tracker.notify.run.panic-in-engine-build-step-run","data":{"build":1,"error":"panic in
engine build step run map[pipeline_name:concourse-set-pipeline-dos team_name:main]: runtime error: invalid memory address or nil pointer dereference","job":"update-pipeline-with-task-yml","pi
peline":"concourse-set-pipeline-dos","session":"24.24.2"}}
web_1     | {"timestamp":"2020-10-04T11:13:46.014280798Z","level":"debug","source":"atc","message":"atc.tracker.notify.run.engine-build-done","data":{"build":1,"job":"update-pipeline-with-tas
k-yml","pipeline":"concourse-set-pipeline-dos","session":"24.24.2"}}
[...]

PR #6116, opened 2020-10-05
Prevent set_pipeline runtime error

mdb:

This PR seeks to address issue #6111 by preventing a `runtime error: invalid memory address or nil pointer dereference` when `set_pipeline` is run with a `file:` whose YML contains unexpected content.

This PR attempts to prevent the runtime error in the most minimal way possible by _only_ attempting to invoke `pipeline.SetParentIDs(step.metadata.JobID, step.metadata.BuildID)` _if_ a pipeline is found by `team.Pipeline(pipelineRef)`, in effect avoiding a scenario where `pipeline` doesn't exist and where `pipeline.SetParentIDs()` causes a runtime error.

One could imagine other solutions, such as perhaps improved [configvalidate#Validate](https://github.com/concourse/concourse/blob/master/atc/configvalidate/validate.go#L25) functionality to protect against this. However, the solution implemented in this PR aspires to be a simple first pass at fixing the problem with a minimal code change.

## Notes to reviewer:

This fix can be verified by running `docker-compose up --build` against this code and `fly set-pipeline`-ing the following pipeline...

```yaml
resource_types:

- name: git-ubuntu
  type: registry-image
  source:
    repository: concourse/git-resource
    tag: ubuntu

resources:

- name: gist
  type: git-ubuntu
  source:
    uri: https://gist.github.com/ac6ae0980b268ae5ff2449fef3ab87c5.git

jobs:

- name: update-pipeline-with-task-yml
  plan:
  - get: gist
  - set_pipeline: bad-pipeline
    file: gist/task.yml

- name: update-pipeline-with-empty-yml
  plan:
  - get: gist
  - set_pipeline: bad-pipeline
    file: gist/empty.yml

- name: update-pipeline-with-valid-yml
  plan:
  - get: gist
  - set_pipeline: good-pipeline
    file: gist/valid-pipeline.yml
```

Running `update-pipeline-with-task-yml` and `update-pipeline-with-empty-yml` should not result in newly-set `bad-pipeline` pipelines. Running `update-pipeline-with-valid-yml` _should_ result in a newly-set `good-pipeline`.

## Release Note

`set_pipeline` of a YML pipeline configuration file with no `jobs:` or `resources:` no longer causes a `runtime error: invalid memory address or nil pointer dereference`.

## Contributor Checklist

- [X] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/exec/set_pipeline_step_test.go
+++ b/atc/exec/set_pipeline_step_test.go
@@ -31,18 +31,6 @@ jobs:
 - name:
 `
 
-	const badPipelineContentWithTaskSyntax = `
----
-platform: linux
-image_resource:
-  type: registry-image
-  source: {repository: busybox}
-run:
- path: echo
- args:
-   - hello
-`
-
 	const badPipelineContentWithEmptyContent = `
 ---
 `
@@ -261,30 +249,6 @@ jobs:
 			})
 		})
 
-		Context("when pipeline file exists but has task syntax", func() {
-			BeforeEach(func() {
-				fakeWorkerClient.StreamFileFromArtifactReturns(&fakeReadCloser{str: badPipelineContentWithTaskSyntax}, nil)
-			})
-
-			It("should not return error", func() {
-				Expect(stepErr).NotTo(HaveOccurred())
-			})
-
-			It("should log no-diff", func() {
-				Expect(stdout).To(gbytes.Say("no diff found."))
-			})
-
-			It("should send a set pipeline changed event", func() {
-				Expect(fakeDelegate.SetPipelineChangedCallCount()).To(Equal(1))
-				_, changed := fakeDelegate.SetPipelineChangedArgsForCall(0)
-				Expect(changed).To(BeFalse())
-			})
-
-			It("should not update the job and build id", func() {
-				Expect(fakePipeline.SetParentIDsCallCount()).To(Equal(0))
-			})
-		})
-
 		Context("when pipeline file exists but is empty", func() {
 			BeforeEach(func() {
 				fakeWorkerClient.StreamFileFromArtifactReturns(&fakeReadCloser{str: badPipelineContentWithEmptyContent}, nil)
@@ -343,6 +307,30 @@ jobs:
 
 				It("should stdout have message", func() {
 					Expect(stdout).To(gbytes.Say("done"))
+				})
+
+				Context("when no diff is found", func() {
+					BeforeEach(func() {
+						fakeWorkerClient.StreamFileFromArtifactReturns(&fakeReadCloser{str: badPipelineContentWithEmptyContent}, nil)
+					})
+
+					It("should not return error", func() {
+						Expect(stepErr).NotTo(HaveOccurred())
+					})
+
+					It("should log no-diff", func() {
+						Expect(stdout).To(gbytes.Say("no diff found."))
+					})
+
+					It("should send a set pipeline changed event", func() {
+						Expect(fakeDelegate.SetPipelineChangedCallCount()).To(Equal(1))
+						_, changed := fakeDelegate.SetPipelineChangedArgsForCall(0)
+						Expect(changed).To(BeFalse())
+					})
+
+					It("should not update the job and build id", func() {
+						Expect(fakePipeline.SetParentIDsCallCount()).To(Equal(0))
+					})
 				})
 			})
 
--- a/atc/exec/set_pipeline_step_test.go
+++ b/atc/exec/set_pipeline_step_test.go
@@ -31,6 +31,22 @@ jobs:
 - name:
 `
 
+	const badPipelineContentWithTaskSyntax = `
+---
+platform: linux
+image_resource:
+  type: registry-image
+  source: {repository: busybox}
+run:
+ path: echo
+ args:
+   - hello
+`
+
+	const badPipelineContentWithEmptyContent = `
+---
+`
+
 	const pipelineContent = `
 ---
 jobs:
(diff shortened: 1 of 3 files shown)


taylorsilva (commented):
Thanks for the PR! Just played around with it to see what the before and after looks like. 

I was surprised to see all the jobs go green; it felt misleading. With what's on master right now the jobs actually enter a failure state, not the right one, but it at least tells the user "hey something went wrong, start digging". With the jobs going green this information is now hidden and someone wouldn't realize that the job was actually a no-op until 1) they find their pipeline wasn't created or 2) their pipeline isn't reflecting their update.

Seeing as the nil pointer doesn't bring down the ATC anymore either I'd actually prefer to keep things as-is and work towards the ideal state of erroring and printing a useful message for the user. wdyt?

mdb:
@taylorsilva Thanks for the feedback (and it was nice meeting you on the office hours call this evening)!

> With the jobs going green this information is now hidden and someone wouldn't realize that the job was actually a no-op until 1) they find their pipeline wasn't created or 2) their pipeline isn't reflecting their update.

You make a good point. However, it's my understanding that the _reason_ no pipeline is created is because the YML results in a [atc.Config](https://github.com/concourse/concourse/blob/7f1c1ea44b38dcc94c34dbab9e6e7d767a101c7d/atc/config.go#L19-L26) whose fields (`Groups`, `VarSources`, `Resources`, `ResourceTypes`, `Jobs`, `Display`) are all empty.

So, with this in mind, I would expect `set_pipeline`-ing such a config would behave akin to its `fly set-pipeline` analog. However, `fly set-pipeline` of the same YML exits zero (success!) and reports "no changes to apply" (Is the inconsistent "no diff found" (`set_pipeline`) VS "no changes to apply" (`fly set-pipeline`) messaging perhaps another [issue](https://github.com/concourse/concourse/issues) worth documenting? Or is that intentional?):

```
./fly -t local set-pipeline --pipeline empty --config empty.yml
no changes to apply

echo $?
0

./fly -t local set-pipeline --pipeline empty --config task.yml
no changes to apply

echo $?
0
```

Furthermore `fly validate-pipeline` reports that the YML "looks good:"

```
./fly validate-pipeline --config empty.yml
looks good 

./fly validate-pipeline --config task.yml
looks good 
```

So, I would expect that `set_pipeline` and `fly set-pipeline` behave similarly when fed the same YML configuration.

But, if you'd prefer `set_pipeline` error in such a scenario, is it also worth reconsidering `fly set-pipeline` and `fly validate-pipeline`'s behavior? Such deeper reconsideration is some of what I was alluding to when I mentioned "One could imagine other solutions, such as perhaps improved `configvalidate#Validate` functionality to protect against this." However, I opted to punt on those more thorny product and code changes in favor of this relatively minimal change, which at least establishes a bit more consistency for now 😄 

Does that change your view?

taylorsilva:
@mdb That makes sense: fly is behaving one way currently and you're using that as the baseline for how `set_pipeline` should behave. Normally I would go with that too but I still think that brings use to an undesirable UX for `set_pipeline`. I think fly's current behaviour is mostly okay because the user does see specific feedback and can act on it if it doesn't meet their expectations (e.g. they `fly set-pipline` and get `no changes to apply` when they were expecting a diff). I think fly's behaviour could be improved too though to cover the examples mentioned in your PR and tests.

Not sure if we do this in the pipeline validation logic already but I'm thinking an easy way to see if a pipeline is valid is to check if it contains at least one job. I know I'm pulling you into another section of the codebase, but I'm thinking we could have both fly and `set_pipeline` behave the same by checking if a pipeline config contains zero jobs.

We could potentially add this logic here:

https://github.com/concourse/concourse/blob/7948c873d3e4c3949201f40e99c2570654df570c/atc/configvalidate/validate.go#L53-L58

Something like:
```go
if len(c.JobConfigs) == 0 {
    warnings = append(warnings, ConfigWarning{Type: "jobs", Message: "config contains zero jobs"})
}
```

I quickly scanned the validate test suite and didn't find anything doing this check. I think this could be an equally small change that may get us to our desired state. What do you think?

taylorsilva:
@vito curious about your thoughts here ^
trying to figure out if there are any gotchas here or what the historical reason is for not having this check already?

Chatted with @xtreme-sameer-vohra about it and he pointed out that `jobs` is optional according to our docs: https://concourse-ci.org/pipelines.html

The question I'm trying to answer is: what is considered the most minimal valid pipeline?
I think the answer is _a single valid job_. Is there a reason that is not true?

vito:
@taylorsilva Yeah, I'd say checking for a job to be present is sensible. In fact if you try to add a resource without adding a job you'll get a validation error saying the resource is unused.

Just a quick note: we expressly allow unknown toplevel keys, because folks use them to tuck away YAML `&anchors` to be referenced elsewhere in the config. So that's why weird things like a task config are valid: we ignore the unknown fields, and accept an empty pipeline.

The only practical reason I can think of to accept an empty config is if we want to allow someone to comment out the entire darn YAML and configure it. But it seems like you could also just pause the pipeline.

Beyond that, it's kind of a philosophical question: given a pipeline is just a set of N jobs, is a set of 0 jobs a valid pipeline? I think it feels kind of nice to allow it, since it's sort of the "base case" given our definition of a pipeline, but pragmatically speaking it's more likely a user error and it'd be nice to report those early - so it's not a hill I'd seek to die on. 😄

So I'd be OK with adding a validation for this, but I also think the code submitted so far is worth merging. Not checking the `found` value looks like a panic waiting to happen.

atc/exec/set_pipeline_step_test.go
```diff
@@ -241,6 +257,54 @@ jobs:
 			})
 		})
 
+		Context("when pipeline file exists but has task syntax", func() {
```
taylorsilva, line 260: For the tests, these won't be valid once we update the `configvalidate` package and they don't express the desired UX. The following is weird but it is essentially the scenario that your team ran into (kinda?).

There's a context `when specified pipeline not found`, and under that could be another context that says `when no diff is found`. That'll push us down the codepath to verify that we don't call `pipeline.SetParentIDs()` when no pipeline was found.
```go
Context("when no diff is found", func() {
  ...
}

**Review by taylorsilva — CHANGES_REQUESTED** [review]
Need to rebase this PR now. 
Left one comment about updating the tests and then this will be good to merge!

**Comment by taylorsilva** [comment]
> Just a quick note: we expressly allow unknown toplevel keys, because folks use them to tuck away YAML &anchors to be referenced elsewhere in the config. So that's why weird things like a task config are valid: we ignore the unknown fields, and accept an empty pipeline.

Thanks, that's exactly what I was trying to surface. Thankfully checking there's one job should not break this in any way.

File: atc/exec/set_pipeline_step_test.go (hunk 1 above)
Comment (mdb) [reply atc/exec/set_pipeline_step_test.go]: @taylorsilva Thanks! I _believe_ I've modified the tests as you're describing. Please let me know if I've misinterpreted.

**Comment by mdb** [comment]
@vito and @taylorsilva Thanks for the patient reviews. I've rebased, made what I believe to be the suggested test changes, and added a release note. 

While it's only peripherally related to this PR, I'm also curious to hear what you all think about this question (I'd be happy to submit a PR for it as well):

> Is the inconsistent "no diff found" (`set_pipeline`) VS "no changes to apply" (`fly set-pipeline`) messaging perhaps another issue worth documenting? Or is that intentional?

Thanks!

**Comment by vito** [comment]
Good catch - I think we should make the set_pipeline step consistent with the fly output.

File: release-notes/latest.md [hunk 2]
```diff
@@ -33,3 +33,7 @@
 
   A new metric called `tasks_wait_duration_bucket` is also added to express as quantiles the average time spent by tasks awaiting execution. PR: #5981
   ![Example graph for the task wait time histograms.](https://user-images.githubusercontent.com/40891147/89990749-189d2600-dc83-11ea-8fde-ae579fdb0a0a.png)
+
+#### <sub><sup><a name="6111" href="#6111">:link:</a></sup></sub> bug
+
+* @mdb Running `set_pipeline` against a pipeline config YML file with no `jobs:` no longer causes runtime error. Issue: #6111 PR: #6116
```
taylorsilva, line 39: No need to do this. When the next version of Concourse is released we read out whatever you put in the `Release Notes` header of this PR.

taylorsilva (changes requested):
Remove the release notes diif and then this will be good to go!

release-notes/latest.md
mdb: @taylorsilva Gotcha; I've removed this commit. I was following the instructions in the `CONTRIBUTING.md` doc. Perhaps it's worth updating these instructions?

taylorsilva (approved):

Issue #2687, opened 2018-10-16
Job boxes should allow descriptions


domdom82:
## What challenge are you facing?

This issue was raised in an open space discussion about concourse features during the CF Summit unconference at Basel 2018.

In large pipelines with many jobs it can become difficult for an ops person who did not design the pipeline to know what a specific job does. People are asking for some sort of "description" field that
allows the pipeline designer to explain more deeply the purpose of a job.

## What would make this better?

When I click on a job there should be an optional field that describes it in detail.

## Are you interested in implementing this yourself?

Sorry, not a UI guy :)

domdom82:
I'd say rather dup of #569 ?
We'd like to add descriptions to jobs, not single builds (which is also a great idea :)

taylorsilva:
Duplicate IMO, as per the previous comment

Issue #569, opened 2016-08-03
Proposal: add description into each jobs


fmy:
We can see the big picture of pipelines with D3.js UI.
But we can not understand what is done in each jobs from this UI.

Adding description field in yml files like below, and showing it anywhere in web UI.
```
jobs:
- name: test
  desc: unit-testing source-codes and put coverage reports into s3, then trigger build jobs.
```

For example
- show description tooltips with hovering each job boxes
- show descriptions in job pages.

keymon:
+1 

It is interesting to add some additional information of each job in the interface, so big pipelines created by big teams are more usable.

It would be nice to display this information when hovering the mouse over the job box.

keymon:
It would be nice to also integrate this feature with the variables, to enrich this description or meta information. Specially the new syntax #545. 

If these variables are later dynamic, this description should be enabled.

keymon:
Be able to parse links would be a plus: #180

keymon:
Also related: https://github.com/concourse/concourse/issues/310#issuecomment-233156116

shinji62:
@vito any thought about job description ? Something we can work on / help ?

vito:
I'm a little concerned that this adds an expected overhead for people to maintain pipelines, to also do their "due diligence" and add a description to each job. Are there patterns Concourse is leading to that make things not self-describing enough?

stickfigure:
Over here in services we have a lot of pipelines and they're complex enough to be hard on newbies (like me). It would be helpful to have some ability to express some basic documentation that displays in the UI.

I'm not sure it needs to be per-job, but certainly one big block of text that describes the entire pipeline would be helpful for navigation and getting oriented.

mockersf:
I am interested on working on this, and the related issues in Annotations project

I was thinking of adding an optional field "description" in resources, pipelines and jobs. When this field is defined, the UI would display a ["(i)"](https://materialdesignicons.com/icon/information-outline) in the breadcrumb that would display this description.

@vito how does that sounds to you ?

vito:
@mockersf :turtle: That sounds great!

/cc @Lindsayauchin

Lindsayauchin:
Something like this?

### Pipeline: 
![Screen Shot 2019-05-14 at 5 19 29 PM](https://user-images.githubusercontent.com/27746614/57733029-7c920280-766c-11e9-9da9-7b99c3e7c816.png)
[...]

PR #6155, opened 2020-10-14
Allow resources and jobs to have display names in the pipeline dashboard.

### LINKED ISSUE #3793: Resource/job display name

Jobs and resources in a pipeline need to have a unique name. 
However, we end up with multiple jobs/resources in a pipeline that have longer names than necessary in order to deal with this. eg, job1-abc, job2-def. This makes the pipeline harder to read and also makes the pipeline longer which then zooms out the pipeline.

Jobs/resource to have an optional display name attribute as well as a name. The display name, if it exists, is the one shown in the UI. If it doesn't exist, the name attribute is used instead.

Possibly

simonjohansson:
Screenshot 2020-10-14 at 13 57 52

Render of [1]

Feature
closes #3793

Due to https://github.com/concourse/concourse/issues/3985 , a lot of our pipelines throws deprecation warnings and we and our developers are very keen on continuing using `/_[A-Z]>....` for resource and job names as it's much more pleasant on the eyes.

This spike PR simply introduces a `display_name` field for Jobs and Resources to allow the user to define how it should be rendered.

Example for when it would be useful for us.
We have pipelines that gets triggered when other `pipeline/job` have finished/failed, some of these pipelines have > 10 pipeline triggers. Say we want to trigger when the job `Deploy to staging (Old CF environment)` in pipeline deploy-app have finished. You might have noticed the `(Old CF environment)` part, we currently have three CF environments and teams sometimes deploy to all three of them in the same pipeline, so its very convenient to allow arbitrary chars there as well.

With the restricted char set the resource would be rendered as (or some other permutation)

```
deploy-app.deploy-to-staging-old-cf-environment
```

whereas currently its shown as

```
deploy-app/Deploy to staging (Old CF environment)
```

## Notes to reviewer:
This is just a spike, meant to kickstart a discussion.

There are obvious UX issues, when you click a job/resource the name displayed is, well, `name`. How should the rendering behave when `display_name` is defined?

Screenshot 2020-10-14 at 14 10 53

Same goes for the gets and puts to a resource in a Job that have a `display_name`, should those be rendered with the real name or the display name?

Another issue that might arise

```
$ fly -t team hijack --check="my-pipeline/🛫 docs"
```

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]
- [x] Add support in ATC
- [x] Display the display name if present on the resource page
- [x] Display the display name if present on the job page
- [x] Display the display name if present on the build page
- [x] Display the jobs display name on the resource page if set under inputs to/ outputs of
- [ ] Display the resources display name if present on get/put on the job page?
- [ ] Display the resources display name if present on get/put on the build page?

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

[1]
```
---
resources:
  - name: concourse-docs-git
    display_name: "🛫 docs"
    type: git
    icon: github
    source:
      uri: https://github.com/concourse/docs
  - name: concourse-src-git
    type: git
    icon: github
    source:
jobs:
- name: compile-docs
  public: true
  plan:
    - get: concourse-docs-git
      trigger: true
    - task: list-files
      config:
        inputs:
          - name: concourse-docs-git
        platform: linux
        image_resource:
          type: registry-image
        run:
          path: ls
          args: ["-la", "./concourse-docs-git"]
- name: compile-src
  display_name: "compile 🛫 src"
  public: true
  plan:
    - get: concourse-src-git
      trigger: true
    - task: list-files
      config:
        inputs:
          - name: concourse-src-git
        platform: linux
        image_resource:
          type: registry-image
        run:
          path: ls
          args: ["-la", "./concourse-src-git"]

- name: release
  display_name: "😍 release docs and binaries 😍"
  public: true
  plan:
    - get: concourse-docs-git
      trigger: true
      passed:
        - compile-docs
    - get: concourse-src-git
      trigger: true
      passed:
        - compile-src
    - task: list-files
      config:
        inputs:
          - name: concourse-docs-git
        platform: linux
        image_resource:
          type: registry-image
        run:
          path: ls
          args: ["-la", "./concourse-docs-git"]
```


aoldershaw:
Thanks for the PR! Seems like a reasonable enhancement. I'll let @matthewpereira chime in if he has any objections design-wise.

> when you click a job/resource the name displayed is, well, `name`. How should the rendering behave when `display_name` is defined?

I think we should do the same thing you're doing on the pipeline view - try to use `display_name` and fallback to `name`. We already request the job/resource from the API when we load these pages, so we could have some logic to:
* Display the name from the URL before the API response resolves (current behaviour)
* Once the API response resolves, display either the `display_name` or the `name` from the response

simonjohansson:
@aoldershaw @matthewpereira 

Ive implemented the changes to the resource view, the job view and the builds view.

Question:

In the following examples, should the display_name be used as well?

Screenshot 2020-10-20 at 17 34 34
Screenshot 2020-10-20 at 17 35 05
Screenshot 2020-10-20 at 17 35 39

matthewpereira:
@simonjohansson Thanks for your patience with this PR. It looks great so far.

My inclination is to say that we should try to use the same term everywhere; once we have a `display_name` we should use that in each of the ❓ scenarios you've got a screenshot for in your message directly above this one, and fall back to `name` otherwise.

simonjohansson:
Alrighty, @aoldershaw @matthewpereira @vito . I have a working solution in 73f4ff389, but its not tested because we need to have a chat about if this is the right way forward.

Basically, in the resources and build view we populate the names (`//div[@class="build-step"]/div/div/h3/text()`) with the `name` from the fetched plan (`/api/v1/builds/$buildID/plan`)

We could use the same pattern as we do for setting `change` in `setupGetStep` to indicate that we are getting a new resource(`get` becomes yellow in the UI) and create `setupPutStep` and `setupCheckStep` to update the name from the fetched data (`/api/v1/teams/main/pipelines/$pipelineName/resources/$resourceName` on the resource view and `/api/v1/builds/$buildID/resources` on the build view)

But, this becomes an issue in the builds view, when you trigger a new build for a job that has a put step.

Basically what happens is
1. We click the build button
2. We get a plan (`api/v1/builds/$buildID/plan`) and the resources `R` (`/api/v1/builds/$buildID/resources`)
3. At this time the `R.output` is a empty list whereas `R.input[x]` might contain displayName[1] (which could be used to render)
4. Thus we have no way of knowing if the put has a display name and will be rendered with the name
5. If you refresh the build page after a build we will fetch the resources (`/api/v1/builds/$buildID/resources`) which will now contain the output with the display name and now everything renders correctly

So to circumvent this I decided  to simply attach the display name if present to the builds `public_plan` in the database, so when we get the plan in the build/resource view we already have all data we need to render. Does this seem like a valid approach? If so let me know and I'll make sure it's properly tested.

I will be off between 26th of October until 2nd of November so I will pick up the work again then.

[1] Since the endpoint called here is the same as the call made on the jobs page (`http://localhost:8080/teams/main/pipelines/big/jobs/$jobName` where we also want to render the display name

vito:
Sorry to be a stick in the mud here but I kind of want to surface some of the trade-offs involved with a feature like this.

Configuring a display name that changes how a pipeline component (resource, job) is shown everywhere in the UI disrupts the IMO very valuable feedback cycle pipeline developers have between the UI and the config. Whereas before the UI was a representation/visualization of the config, now it can look completely different, and I can't really 100% trust that the display names map to what I think they map to. This could result in subtle anxieties and extra mental cycles over time, and could present a small hurdle on top of everything else a new team member has to learn.

So: I hear the need to be able to express more information about a job/resource in the UI, but I'm not sure replacing the pipeline config names is the best way to do it. Could this be accomplished by adding a description instead? (#569 #2687) I'm not sure how it'd show up in the UI, but we could potentially allow Markdown so you could even link to useful information and such. To me, configuring a display name *and* a description is a bit redundant, and between the two a description seems more valuable and less invasive.

Thanks for your PR though, and for starting the conversation early - sorry I came in a bit late with this concern! 🙏

simonjohansson:
@vito I largely agree, with that said I still think its a reasonable feature to allow a pipeline to have display names in the pipeline view per `Changes proposed by this PR`

```
Example for when it would be useful for us.
We have pipelines that gets triggered when other pipeline/job have finished/failed, some of these pipelines have > 10 pipeline triggers. Say we want to trigger when the job "Deploy to staging (Old CF environment)" in pipeline deploy-app have finished. You might have noticed the (Old CF environment) part, we currently have three CF environments and teams sometimes deploy to all three of them in the same pipeline, so its very convenient to allow arbitrary chars there as well.

With the restricted char set the resource would be rendered as (or some other permutation)

"deploy-app.deploy-to-staging-old-cf-environment"

whereas currently its shown as

"deploy-app/Deploy to staging (Old CF environment)"
```

Virtually all of our jobs and resources have spaces, or slashes, or other random chars and the UI would look horrible when we are forced to make them adhere to the strict, what is essentially ID, rules. Loads of pipelines contain cognitive helpful information baked into the name so a developer can quickly see what job has failed when on the pipeline overview, se above example.

Personally I have no problems with making our job and resource names adhere to the new rules, but convincing hundreds of developers in our org that you cannot have names with spaces, or parentheses, or whether they see fit, and that they have to change thousands of pipelines is not a conversation Im looking forward to. :)

I do realise that this is for the better as its gonna be a never ending battle to try to make the API be able to serialize/deserialize every available permutation imaginable. And today we had lots of jobs where you cannot cancel builds due to random chars in the name(well `abort-build -b ..` to the rescue) in the UI but again, this is a change(initial change for the pipeline view, not so much the subsequent ones) I think will be a net positive.

simonjohansson:
@vito @matthewpereira @aoldershaw any thoughts?

As per https://github.com/concourse/concourse/issues/3985 there are more companies than ours that could benefit from this.

simonjohansson:
Ping pong :)

vito:
Sorry for the wait - dealing with an overlap of a lot of work-and-life deadlines/events right now! 🙏

I guess my argument is that folks are going to have to change their pipelines either way, and configuring a `description:` instead of a `display_name:` could be roughly the same net code change while maintaining the config <=> UI feedback loop and delivering a more powerful feature that has been requested for a long time.

(Sidebar: these identifier validations are warnings and not errors, and we'll hold off on them being errors for as long as it takes to learn from feedback like yours and iterate until everyone's as happy as possible - so I plan to carefully consider these changes and take as much time as needed, as they will have a lasting impact on Concourse's design. That might even mean silencing the warnings themselves for now if they're too annoying and we're taking too long to get there. Just want to make my expectations here clear. 🙂)

If you're still in favor of configuring a display name instead, I don't want to be the sole dissenting voice holding everything up, though regarding the changes themselves I would like to try and minimize the breadth of impact on the codebase. To me this seems like a web UI concern ("when displaying this resource/job/etc, show the display name instead of its real name"), but it looks like right now it has trickled all the way down into places like the `atc.Plan` types and API types where it feels a little redundant. I think it might be better to leave everything as the 'real' name, return the display name on the resource/job API types (`atc.Job`, `atc.Resource`), and let the web UI put 2 + 2 together. Does that sound reasonable?

Thanks!

simonjohansson:
@vito 

> Sorry for the wait - dealing with an overlap of a lot of work-and-life deadlines/events right now! 🙏

No worries, sorry for the impatience. :)

> guess my argument is that folks are going to have to change their pipelines either way

We are using a pre-processor that takes a [file like this](https://github.com/springernature/halfpipe-examples/blob/master/nodejs/.halfpipe.io) and turns it into a [valid concourse pipeline](https://github.com/springernature/halfpipe-examples/blob/master/nodejs/pipeline.yml) so this is not so much of a problem for us :) . But yeah, point taken

> configuring a description: instead of a display_name:

I don't think these two concepts are mutually exclusive? In my mind `display_name` is more controlling how a resource/job is rendered in the pipeline view whereas `description` contains human readable metadata?

If you agree Im happy to add the `description` functionality alongside with `display_name`!

> Sidebar: these identifier validations are warnings and not errors, and we'll hold off on making them errors for as long as it takes to learn from feedback like yours and iterate until everyone's as happy as possible

🙏 :thumbsup:

> though regarding the changes themselves I would like to try and minimize the breadth of impact on the codebase

 Yeah, definitely, the changes are very intrusive when it comes to displaying the `display_name` anywhere except the pipeline view(partly because I just bolted the functionality on in the quickest way to have something to show.). Tbh I prefer to render the actual name instead of `display_name` everywhere except the pipeline view. And the change for doing that is https://github.com/concourse/concourse/pull/6155/commits/b99aee36db20debfb1676486354be807d569bb42 all the subsequent commits Im more than happy to throw away.


But I guess that is up for discussion, what is easiets/least confusing for the end user.

So. To make the ball rolling again, are you in favour of the following?
1. Allow both `display_name` and `description` to be set for resources and jobs.
2. If `display_name` is defined on any of the resources/jobs use that to render in the pipeline view, if not just use the normal `name`
3. If `description` is defined for resources/jobs return it in the call to `/api/v1/teams/$TEAM/pipelines/$PIPELINE/jobs` and `/api/v1/teams/$TEAM/pipelines/$PIPELINE/resources`
4. Figure out in a later issue/PR about if/how to display the `display_name` in the resource, job and builds view
5. Figure out in a later issue/PR about how/where to display the `description`

vito:
@simonjohansson Thanks, but no need to add it - my suggestion wasn't to try to tack on more features; I'm trying to keep the surface area small. Configuring `name`, `display_name`, *and* `description` for potentially each job starts to feel like busywork. Between the two, `description` felt like it had more utility and it had been requested before the identifier validations were even added, so I had hoped to kill two birds with one stone by finding common ground.

My hesitation with `display_name` is that *replacing* the real name with another name in the pipeline UI is a pretty disruptive change. I would support a form that *doesn't* hide the real name and shows the display name separately, and that's what led me to the old `description` idea. I think it could also prevent the pipeline UI from stretching out as longer display names are added; horizontal screen real estate ain't cheap, and the wider a pipeline is the more it has to zoom out, making it harder to grok.

I guess I never asked explicitly: would a tooltip or some sort of additional text area in the job (truncated/expandable if too long) be sufficient for your use case? I think this approach would have much less downside in the long run, while having only slightly less upside for this use case, but much more upside for the `description` use case (i.e. self-documenting pipelines).

Again, sorry to be a stick in the mud; Concourse's entire conception is around identifying little things that can go wrong at all levels of the user's experience with the tool or with the team that they've joined, so I'm always in the mindset of trying to keep the it small and straightforward. Thanks for your patience; I know this can lead to a lot of friction. 🙏

robwhitby:
I think the reason this is a tricky issue is due to trying to force identifier semantics onto a field called "name". As IDs and names are two different things, perhaps we could approach from a slightly different angle by adding an `id` field (defaulting to the value of `name` for backwards compatibility) and allowing `name` to continue to be a "name".

simonjohansson:
@vito 

> I guess I never asked explicitly: would a tooltip or some sort of additional text area in the job (truncated/expandable if too long) be sufficient for your use case? I think this approach would have much less downside in the long run, while having only slightly less upside for this use case, but much more upside for the description use case (i.e. self-documenting pipelines).

Not really. To come back to the initial issue we see.

Here is a pretty standard pipeline that triggers on changes to git and when two other pipeline/jobs have succeeded. 

Screenshot 2020-12-01 at 12 23 45

If we follow the identifier schema

Screenshot 2020-12-01 at 12 28 32

And if we simplify a bit to remove as much noise as possible(`-`)

Screenshot 2020-12-01 at 12 32 10

I think we have lost contextual knowledge at this point. One could ofc argue that build implies that the tests have been run, or that notify would have contacted the right team based on the team you are in. Highly subjective, but I also think this makes things uglier and harder to read(Im a huge fan of spaces for delimiters, upper cases and ampersands, or when Im feely cute a smiley here or there) :)

> Again, sorry to be a stick in the mud; Concourse's entire conception is around identifying little things that can go wrong at all levels of the user's experience with the tool or with the team that they've joined, so I'm always in the mindset of trying to keep the it small and straightforward. Thanks for your patience; I know this can lead to a lot of friction. 🙏

No need to appologize, I totally get that and empathize with the main points of the 040-valid-identifiers RFC.

```
The success of tools like go fmt has shown that the engineers value global consistency and reduction of petty debates over personal stylistic preferences.

In the spirit of consistency and simplicity, this proposal is to dramatically reduce the allowed character set for Concourse identifiers.
```

This is obviously super sane and desirable for identifier, but I also think there needs to be atleast a superficial distinction between 
identifiers and a names(and I think with proper documentation the user would have no problem to discern between the two, if they so choose to use the proposed solution). 

Would this back and forwards maybe be better suited for a Github Discussion to see what the community at large thinks? Some more input would be helpful for moving the conversation forward.

If not, feel free to close the PR. 🙏

vito:
@simonjohansson A RFC would make sense here - it's a pipeline config change with potentially broad impact, which is typically a sign that it should go through the RFC process to get more eyes on it. 👍

For what it's worth, the RFC process doesn't have to hold up feature development - the feature should just be marked experimental, and warnings should be printed when it's used so that folks are directed to the RFC to provide feedback.

I have one observation based on a lot of the examples you've given so far: it seems like most of the information that you're placing in job names is to specify what the job relates to - deployments, releases, etc.

I would argue that that information is already present in a well-formed pipeline's UI, in the resources that the job interacts with.

image

I prefer short-and-sweet job names because resources are already there to provide context. Using shorter names makes jobs easier to use in anger with `fly`, and it allows the pipeline to fit in smaller windows without zooming out so much that it's annoying to interact with or difficult to identify config issues in passing. Using short names allows the screen real estate to be shared with resources, which are just as important as jobs for telling the full story. Resources even support icons, which your use of emoji indicates would be useful. :P

If I had to guess, I would say that most of these jobs that you'd like to give clearer names are actually performing side effects with *tasks* - which don't show up in the pipeline UI, because they really aren't meant to be used for performing mission-critical side-effects.

From my perspective, this is the pipeline UI doing its job: it's providing feedback that your pipeline is hiding side effects which are meant to be run via resources. If you don't use resources to represent all the effects of the pipeline, the job names carry a heavier burden, and the pipeline UI becomes significantly less useful.

Pulling on that thread, I think it'd be worth taking the time to see if these can be turned into resources - even if it means writing your own resource type - and then see if you have the same needs around job names.

@robwhitby Thanks, that's a good thought - but to me 'name' and 'identifier' don't really have a clear and obvious difference. Method names, for example, are extremely limited in their allowed characters - but we don't call them method identifiers. To me, Concourse jobs are very similar - the job names identify logical bits of work and are referenced elsewhere in the pipeline code (in `passed` constraints).

I think this is largely a difference in perspective - we both use the same tool but with different interpretations of `name` - but if there are any others like me, having `id` and `name` fields feels less clear than having `name` and `display_name`.

simonjohansson:
@vito
> @simonjohansson A RFC would make sense here - it's a pipeline config change with potentially broad impact, which is typically a sign that it should go through the RFC process to get more eyes on it. 👍

Alrighty, Ill give it a go! 

> I would argue that that information is already present in a well-formed pipeline's UI, in the resources that the job interacts with.

Your provided screenshot does indeed look nice and concise. Unfortunately the bulk of our pipelines are bigger/more complicated than that though :) 

> Pulling on that thread, I think it'd be worth taking the time to see if these can be turned into resources - even if it means writing your own resource type - and then see if you have the same needs around job names.

Right, I was lazy and just created a dummy pipeline to get the point across :) 

A real world pipeline(with some anonymization of the resource names) would look like

Screenshot 2020-12-02 at 10 01 15


Here we can see that we are doing a `deploy` to three different permutations of CF Envs, orgs and spaces. Furthermore we do a `deploy` to MarkLogic (which granted does not use a resource and simply relies on side effects in the task itself..)

I do agree with `....information is already present in a well-formed pipeline's UI, in the resources that the job interacts with` but currently we cannot simply rename the four logical deploys to short and sweet `deploy` as the `name` is unique across all jobs even if the resource names themselves are unique for each job.

And my point still stands that `CF-dev coolTeam/live` reads nicer than `cf-dev.coolteam-live`, furthermore `coolteam` does not exist in CF, `coolTeam` does. And we need to capture which env/org/space the developer deploys to as a pipeline can deploy to different permutations in the same pipeline. 

I think GitHub Actions solved this nicely.

```
...

jobs:
  my_first_job:
    name: My first job
    ...
  my_second_job:
    ...
```

Screenshot 2020-12-02 at 10 10 44

PR #6156, opened 2020-10-14
Introduce dbtest package for higher-level test setup

vito:

This PR introduces an `atc/db/dbtest` package for setting up test scenarios using higher-level (i.e. more stable) constructs. The aim of this pattern is to reduce the amount of churn whenever we make changes to how we model the internals, and also to reduce the amount of global test suite setup we've got going in under `atc/db`, primarily in `db_suite_test.go`.

* Introduce `dbtest.Builder` which provides higher-level methods (e.g. `WithPipeline`, `WithResourceVersions`) for constructing a `dbtest.Scenario`.
* Convert all tests which used to use `(db.Resource).SetResourceConfig` to use `dbtest` instead.
* Removes `(db.Resource).SetResourceConfig`, as it's no longer used in production code, and with this PR, no longer used in tests.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/dbtest/builder.go
+++ b/atc/db/dbtest/builder.go
@@ -180,6 +180,11 @@ func (builder Builder) WithResourceVersions(resourceName string, versions ...atc
 			return fmt.Errorf("save versions: %w", err)
 		}
 
+		_, err = scope.UpdateLastCheckEndTime()
+		if err != nil {
+			return fmt.Errorf("update last check end time: %w", err)
+		}
+
 		err = resource.SetResourceConfigScope(scope)
 		if err != nil {
 			return fmt.Errorf("set resource scope: %w", err)
--- a/atc/db/dbtest/builder.go
+++ b/atc/db/dbtest/builder.go
@@ -5,6 +5,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db"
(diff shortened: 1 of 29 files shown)


atc/db/dbtest/builder.go
```diff
@@ -77,6 +77,25 @@ func (builder Builder) WithWorker(worker atc.Worker) SetupFunc {
 	}
 }
 
+func (builder Builder) WithTeamWorker(worker atc.Worker) SetupFunc {
+	return func(scenario *Scenario) error {
+		if scenario.Team == nil {
+			err := builder.WithTeam(unique("team"))(scenario)
```
aoldershaw, line 83: Should this use `worker.Team`? Or, should we just error if the team is unset?

Also, might it make sense for `WithWorker` to just handle the case where `worker.Team` is set, and find the appropriate team to save the worker to? I'm not sure if there's a case where `worker.Team` is set but you don't want to `SaveWorker`

aoldershaw (approved):
Have one tiny comment, but don't think it needs to be a blocker. Tests are much nicer to read and should be better insulated from changes. Nice job @vito and @clarafu!

atc/db/dbtest/builder.go
vito: Yeah good call. I pushed https://github.com/concourse/concourse/pull/6156/commits/bded4c77fa3fdfa852f3f3eb185cb91b2f803dfc which removes `WithTeamWorker` and adjusts `WithWorker` to respect the `Team` field instead.

vito:
Thanks for the review!

PR #6170, opened 2020-10-17
Fix regression preventing the use of nested fields with `fly set-pipeline --var`

### LINKED ISSUE #6041: `set-pipeline --var` no longer supports nested keys
## Summary

In upgrading from 6.4.1 to 6.5.1, we saw that our use of `set-pipeline --var` no longer worked. We were previously specifying nested variables like `--var foo.bar=baz` and using them in the pipeline like `((foo.bar))`. However, with the introduction of https://github.com/concourse/concourse/pull/5896, the variable needs to be referred to as `(("foo.bar")`.

## Steps to reproduce

With this config:
```
jobs:
  - name: ((foo.bar))
    plan: []
```

Using fly 6.5.1, set the pipeline: `fly -t ci set-pipeline --pipeline foo --config foo.yaml --var foo.bar=baz`

## Expected results

```
jobs:
  job baz has been added:
+ name: baz
+ plan: []
```

## Actual results

```
jobs:
  job ((foo.bar)) has been added:
+ name: ((foo.bar))
+ plan: []
```

## Triaging info

* Concourse version: 6.5.1
* Did this used to work? Yes, in 6.4.1

aoldershaw:

closes #6041 .

This also fixes a small bug introduced by #5898 - if you reference a quoted var with a `:` that is not in a var source (e.g. `(("my:var"))`), the variable reference would be improperly parsed (as `{Source: "\"my", Path: "var"}`)

This also does some refactoring around pipeline instance vars, and changes their behaviour slightly (`set-pipeline`'s `-i/--instance-var` flag now just behaves the same as `-y/--yaml-var`, the pipeline flag allows for YAML instead of just dot notation, lists aren't traversed in dot notation - peep the commit messages for more details). Since pipeline instances have not been released yet, this is non-breaking

* Add the ability to "expand" a list of key-value pairs into a `vars.StaticVariables`
  * This follows the fields on var references and recursively creates maps
  * Note that @mouellet had added something similar for instance vars - this is a simpler version that doesn't traverse lists. The instance vars implementation now uses `KVPairs` and `StaticVariables` instead
  * e.g. 

```
KVPairs{
  {Ref: Reference{Path: "foo", Fields: []string{"a", "b"}}, Value: "1"},
  {Ref: Reference{Path: "foo", Fields: []string{"a", "c"}}, Value: "2"},
}.Expand()
```

gives

```
StaticVariables{
  "foo": map[string]interface{}{
    "a": map[string]interface{}{
      "b": "1",
      "c": "2",
    }
  }
}
```

* Parse var references "manually" (i.e. without using regex) - this makes it easier to control the parsing behaviour IMO
* Instance var flag now just takes a single value, making it more consistent
* Pipeline flag parsing now allows specifying YAML objects directly if you don't want to/can't use dot notation (e.g. for lists, now there is no dot notation)

## Release Note

* A regression was introduced in 6.5.0 that prevented the use of nested fields when setting a pipeline variable via `fly set-pipeline --var`
  * For instance, pre 6.5.0, `fly set-pipeline --var foo.bar=123 --var foo.baz=456` would create the variable `foo` with value `{bar: 123, baz: 456}` (that can be referenced in the pipeline config as `((foo.bar))`, `((foo.baz))`)
  * In 6.5.x/6.6.x, the same command would create variables `"foo.bar" = 123` and `"foo.baz" = 456` (that would have to be referenced in the pipeline config as `(("foo.bar"))` and `(("foo.baz"))`, respectively)
  * If you want to set a variable with a `.` in it, you can now quote the flag: e.g. `fly set-pipeline --var '"foo.bar"=123'` (note that this requires quoting the entire flag in `'...'` to avoid shell expansion)

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


fly/commands/internal/flaghelpers/pipeline_flag_test.go
```diff
@@ -17,52 +19,98 @@ var _ = Describe("PipelineFlag", func() {
...
+				desc: "errors if invalid ref is passed",
+				flag: `some-pipeline/"my.field".:bad`,
+				err:  `invalid var '"my.field".': empty field`,
+			},
+			{
+				desc: "errors if invalid YAML is passed as the value",
+				flag: `some-pipeline/hello:{bad: yaml`,
+				err:  `invalid value for key 'hello': error converting YAML to JSON: yaml: line 1: did not find expected ',' or '}'`,
+			},
+		} {
```
chenbh, line 102: Let's add some tests for special characters like `'pipeline/field:"value1,value2"'` and `'pipeline/field:"value1:value2"'`

fly/commands/internal/flaghelpers/resource_flag_test.go
```diff
@@ -15,83 +15,63 @@ var _ = Describe("ResourceFlag", func() {
...
+			desc: "resource name not specified",
+			flag: "some-pipeline/",
+			err:  "argument format should be <pipeline>/<resource>",
+		},
+		{
+			desc: "malformed instance var",
+			flag: "some-pipeline/branch=master/some-resource",
+			err:  "argument format should be <pipeline>/<key:value>/<resource>",
+		},
+	} {
```
chenbh, line 64: Same thing as job_flag_test

atc/pipeline.go
```diff
@@ -39,37 +37,19 @@ type RenameRequest struct {
...
-		}
+	kvPairs := vars.StaticVariables(iv).Flatten()
+	sort.Slice(kvPairs, func(i, j int) bool {
+		return kvPairs[i].Ref.String() < kvPairs[j].Ref.String()
+	})
+
+	var parts []string
+	for _, kvPair := range kvPairs {
+		rawVal, _ := yaml.Marshal(kvPair.Value)
+		val := strings.TrimRight(string(rawVal), "\n")
```
chenbh, line 49: If the `Value` contains a comma, the entire value will need to be quoted or else it can be confused with the separator for another instance var. 

I.e. `fly -t dev up -p p/foo:bar,baz` breaks while  `fly -t dev up -p "p/foo:\"bar,baz\""` is the correct form

fly/commands/execute.go
```diff
@@ -33,8 +33,8 @@ type ExecuteCommand struct {
 	Outputs        []flaghelpers.OutputPairFlag       `short:"o" long:"output"      value-name:"NAME=PATH"    description:"An output to fetch from the task (can be specified multiple times)"`
 	Image          string                             `long:"image" description:"Image resource for the one-off build"`
 	Tags           []string                           `          long:"tag"         value-name:"TAG"          description:"A tag for a specific environment (can be specified multiple times)"`
-	Var            []flaghelpers.VariablePairFlag     `short:"v"  long:"var"       value-name:"[NAME=STRING]"  description:"Specify a string value to set for a variable in the pipeline"`
-	YAMLVar        []flaghelpers.YAMLVariablePairFlag `short:"y"  long:"yaml-var"  value-name:"[NAME=YAML]"    description:"Specify a YAML value to set for a variable in the pipeline"`
+	Var            []flaghelpers.VariablePairFlag     `short:"v"  long:"var"       value-name:"[NAME=STRING]"  unquote:"false"  description:"Specify a string value to set for a variable in the pipeline"`
```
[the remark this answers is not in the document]
aoldershaw: Yeah, it's a bit awkward. The author mentioned it in an issue: https://github.com/jessevdk/go-flags/issues/185#issuecomment-238463021

atc/pipeline.go
aoldershaw: Ah, good catch. I also realized that marshaling values as YAML doesn't work for so well for lists, since it'll do the

```
- val1
- val2
```

syntax. Going to switch this to just marshaling as JSON, and we can explicitly unquote the value if it's a string and it doesn't contain any special characters

chenbh (approved):
lgtm

aoldershaw:
@chenbh FYI I re-hid the `--instance-var` flag, since it looks like #6105 probably won't make it into 6.7

PR #6177, opened 2020-10-19
Use dot notation in instanced pipelines API

zoetian:

With #6105, the query parameters for instanced pipelines in the UI changed from `?instance_vars=...ugly URL encoded JSON` to `vars.var1.field1="value"&vars.var2=123&...`. For consistency with the UI, let's also change the query parameters used in the API.

* `fly pipelines` includes pipeline ID column
* Instanced pipelines are now accessed using the API endpoints `/teams/:team_name/pipelines/:pipeline_name/...?vars.foo.bar=123&vars."baz.qux"="abc"` (corresponds with instance vars `{"foo": {"bar": 123}, "baz.qux": "abc"}`)
  * Previously, this would be `?instance_vars={"foo": {"bar": 123}, "baz.qux": "abc"}`
  * Note that `?vars={"foo":{"bar":123},"baz.qux":"abc"}` is also supported
* Quote `/` contained within an instance var value when converting to string form
  * e.g. `-j release/branch:"feature/foo"/shipit`
  * It still correctly parses if you provide `-j release/branch:feature/foo/shipit`, but that's confusing - when we generate these flags, we should quote it

TODO:

* [ ] ~Fix syslog drainer tag format~
  * The format isn't officially documented anywhere, but we want to be cautious about changing it so as not to break syslog consumers
  * We want to change it because it provides no information about pipeline instance vars
  * Nevermind, going to leave this for a future PR

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/api/auth/check_pipeline_access_handler.go
+++ b/atc/api/auth/check_pipeline_access_handler.go
@@ -2,7 +2,6 @@ package auth
 
 import (
 	"context"
-	"encoding/json"
 	"net/http"
 
 	"github.com/concourse/concourse/atc"
@@ -47,12 +46,11 @@ func (h checkPipelineAccessHandler) ServeHTTP(w http.ResponseWriter, r *http.Req
 	teamName := r.FormValue(":team_name")
 	pipelineName := r.FormValue(":pipeline_name")
 	pipelineRef := atc.PipelineRef{Name: pipelineName}
-	if instanceVars := r.URL.Query().Get("instance_vars"); instanceVars != "" {
-		err := json.Unmarshal([]byte(instanceVars), &pipelineRef.InstanceVars)
-		if err != nil {
-			w.WriteHeader(http.StatusInternalServerError)
-			return
-		}
+	var err error
+	pipelineRef.InstanceVars, err = atc.InstanceVarsFromQueryParams(r.URL.Query())
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		return
 	}
 
 	team, found, err := h.teamFactory.FindTeam(teamName)
--- a/atc/api/cc_test.go
+++ b/atc/api/cc_test.go
@@ -265,6 +265,44 @@ var _ = Describe("cc.xml", func() {
 					})
 				})
 
+				Context("when an instanced pipeline is found", func() {
+					var fakePipeline *dbfakes.FakePipeline
+					instanceVars := atc.InstanceVars{"hello": "world"}
+					BeforeEach(func() {
+						fakePipeline = new(dbfakes.FakePipeline)
+						fakePipeline.InstanceVarsReturns(instanceVars)
+						fakeTeam.PipelinesReturns([]db.Pipeline{
+							fakePipeline,
+						}, nil)
+
+						endTime, _ := time.Parse(time.RFC3339, "2018-11-04T21:26:38Z")
+						fakePipeline.DashboardReturns([]atc.JobSummary{
+							{
+								Name:                 "some-job",
+								PipelineName:         "something-else",
+								PipelineInstanceVars: instanceVars,
+								TeamName:             "a-team",
+								FinishedBuild: &atc.BuildSummary{
+									Name:    "42",
+									Status:  "succeeded",
+									EndTime: endTime.Unix(),
+								},
+							},
+						}, nil)
+					})
+
+					It("returns the proper web url in the CC.xml", func() {
+						body, err := ioutil.ReadAll(response.Body)
+						Expect(err).NotTo(HaveOccurred())
+
+						Expect(body).To(MatchXML(`
+<Projects>
+  <Project activity="Sleeping" lastBuildLabel="42" lastBuildStatus="Success" lastBuildTime="2018-11-04T21:26:38Z" name="something-else/hello:world/some-job" webUrl="https://example.com/teams/a-team/pipelines/something-else/jobs/some-job?var.hello=%22world%22"/>
+</Projects>
+`))
+					})
+				})
+
 				Context("when no pipeline is found", func() {
 					BeforeEach(func() {
 						fakeTeam.PipelinesReturns([]db.Pipeline{}, nil)
--- a/atc/api/cc_test.go
+++ b/atc/api/cc_test.go
@@ -267,7 +267,7 @@ var _ = Describe("cc.xml", func() {
 
 				Context("when an instanced pipeline is found", func() {
 					var fakePipeline *dbfakes.FakePipeline
-					instanceVars := atc.InstanceVars{"hello": "world"}
+					instanceVars := atc.InstanceVars{"branch": "feature/foo"}
 					BeforeEach(func() {
 						fakePipeline = new(dbfakes.FakePipeline)
 						fakePipeline.InstanceVarsReturns(instanceVars)
@@ -297,7 +297,7 @@ var _ = Describe("cc.xml", func() {
 
 						Expect(body).To(MatchXML(`
 <Projects>
-  <Project activity="Sleeping" lastBuildLabel="42" lastBuildStatus="Success" lastBuildTime="2018-11-04T21:26:38Z" name="something-else/hello:world/some-job" webUrl="https://example.com/teams/a-team/pipelines/something-else/jobs/some-job?var.hello=%22world%22"/>
+  <Project activity="Sleeping" lastBuildLabel="42" lastBuildStatus="Success" lastBuildTime="2018-11-04T21:26:38Z" name="something-else/branch:&#34;feature/foo&#34;/some-job" webUrl="https://example.com/teams/a-team/pipelines/something-else/jobs/some-job?var.branch=%22feature%2Ffoo%22"/>
 </Projects>
 `))
 					})
--- a/atc/api/cc_test.go
+++ b/atc/api/cc_test.go
@@ -297,7 +297,7 @@ var _ = Describe("cc.xml", func() {
 
 						Expect(body).To(MatchXML(`
 <Projects>
-  <Project activity="Sleeping" lastBuildLabel="42" lastBuildStatus="Success" lastBuildTime="2018-11-04T21:26:38Z" name="something-else/branch:&#34;feature/foo&#34;/some-job" webUrl="https://example.com/teams/a-team/pipelines/something-else/jobs/some-job?var.branch=%22feature%2Ffoo%22"/>
+  <Project activity="Sleeping" lastBuildLabel="42" lastBuildStatus="Success" lastBuildTime="2018-11-04T21:26:38Z" name="something-else/branch:&#34;feature/foo&#34;/some-job" webUrl="https://example.com/teams/a-team/pipelines/something-else/jobs/some-job?vars.branch=%22feature%2Ffoo%22"/>
 </Projects>
 `))
 					})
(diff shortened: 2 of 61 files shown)


aoldershaw:
@clarafu heads up, I'm going to be reverting the changes to the pipeline URL in #6105, so you can ignore the first two commits for now

aoldershaw:
@clarafu update, we ended up changing the web URL routes back to `/teams/:team_name/pipelines/:pipeline_name`, but changed the query parameters from `?instance_vars=...ugly URL encoded JSON...` to `?var.var1.field1="value"&var.var2=123&...`.

Considering changing the API query parameters to match the web's query parameters, but for now, at least what is pushed is reviewable.

clarafu:
@aoldershaw Sounds good! I have the next two days off so the quickest that I can finish reviewing it is on Thursday, sorry about that!

aoldershaw:
Hey @clarafu, curious to hear your thoughts on the API query parameters for instanced pipelines. To recap, the current API endpoints are e.g. `/api/v1/teams/main/pipelines/release?instance_vars={"version":"6.7.x"}`, and the web route for the same pipeline is `/teams/main/pipelines/release?var.version="6.7.x"`.

The reason for the web query parameters being in the form `var.xxx=yyy` is purely cosmetic - although the API endpoint may not look that bad, the reality is that the query string gets URI encoded, so ends up actually looking like `?instance_vars=%7B"version"%3A"6.7.x"%7D` (which is kinda brutal to look at, compared to `?var.version="6.7.x"` - and gets much worse as there are more instance vars).

I'm debating whether or not to change the API format to the `?var.xxx=yyy` format as well. I can see a few pros/cons to switching over:

Pros
* Consistency is nice
  * There's complexity/potential confusion in supporting both forms in different places

Cons
* Handling the `?instance_vars` form is simpler
  * This isn't the most convincing argument, though, since we have to handle the other form as well - 5ec490ea219edaf11d16aa9c9b56e385ebad6c4d
* It's easier for API clients to use the `?instance_vars=...` form
  * While this isn't a huge deal for the web UI and `fly` (since we already need the code written to generate the `?var.xxx=yyy` form), it makes any external clients more complicated
* The `?var.xxx=yyy` format implies partial filtering on instance vars, rather than an exact match, which I don't think we should support in the API

WDYT?

clarafu:
@aoldershaw Hm I have really minimal knowledge on API endpoint standards and applications so I've got a few questions I'd like to ask first 😅 

> * It's easier for API clients to use the `?instance_vars=...` form
>   
>   * While this isn't a huge deal for the web UI and `fly` (since we already need the code written to generate the `?var.xxx=yyy` form), it makes any external clients more complicated

I'm just wondering how is it easier for API clients to use the `instance_vars` rather than the other var form? I might be thinking about this completely wrong but I was thinking it might just be a difference between them encoding a struct into JSON and passing that as the instance_vars vs passing the fields directly as different parameters in the query string.

> * The `?var.xxx=yyy` format implies partial filtering on instance vars, rather than an exact match, which I don't think we should support in the API

This might be stupid question (because I know nothing of web practices) but I don't think I understand how the var format implies partial filtering vs instance vars? Is it just something that everything that writes APIs kind of "agrees" upon as a standard?

aoldershaw:
> I'm just wondering how is it easier for API clients to use the instance_vars rather than the other var form?

It's not *so* much harder to be fair, but there's some nuance that I didn't really get into around the `var.xxx=yyy` form - the instance vars are actually flattened using dot notation. For instance, if the instance vars are:

```json
{
    "foo": {
        "bar": "hello",
        "baz": 123
    }
}
```

this is encoded as:

```
?var.foo.bar="hello"&var.foo.baz=123
```

There's some added complexity when the path segments contain special characters like `.` e.g.

```
{
    "foo.bar": "hello"
}
```

is encoded as:

```
?var."foo.bar"="hello"
```

That said, I suppose we could allow the following to all be equivalent:

```
?var.foo.bar="hello"&var.foo.baz=123
?var.foo={"bar":"hello","baz":123}
?var={"foo":{"bar":"hello","baz":123}}
```

...which would give clients a more flexible calling option depending on what's most convenient, without adding much complexity to the API.

---

> This might be stupid question (because I know nothing of web practices) but I don't think I understand how the var format implies partial filtering vs instance vars? Is it just something that everything that writes APIs kind of "agrees" upon as a standard?

Yeah that's fair - this point was a bit subjective and perhaps misguided. The way I viewed query params was as a way to filter the results. Suppose a pipeline's instance vars are:

```
{
    "a": 1,
    "b": 2
}
```

To me, specifying `?var.a=1` suggests that you want to filter the pipeline instances to those where the instance var `a` has value `1` - but what if there are multiple pipelines with `a=1`, but different values of `b`? In contrast, using `?instance_vars={...}` is unambiguous - it's clear that you have to specify the full set of instance vars. However, we could specify as a contract that the full set of `var.xxx=yyy` params must define the instance vars in full.

Let me know if that doesn't make any sense. TBH I'm leaning towards using the `?var.xxx=yyy` form for the API

aoldershaw:
> To me, specifying ?var.a=1 suggests that you want to filter the pipeline instances to those where the instance var a has value 1

Put another way, `?var.a=1` feels like you're filtering instances by some criteria, whereas `?instance_vars={...}` feels like you're specifying a particular instance

clarafu:
> That said, I suppose we could allow the following to all be equivalent:
> 
> ```
> ?var.foo.bar="hello"&var.foo.baz=123
> ?var.foo={"bar":"hello","baz":123}
> ?var={"foo":{"bar":"hello","baz":123}}
> ```
> 
> ...which would give clients a more flexible calling option depending on what's most convenient, without adding much complexity to the API.

Does this mean basically supporting both the JSON formatted URLs and the var format (essentially going with the API accepting both the instance vars and var form with the only difference being that it is not passed as `instance_vars` but rather only as `var`)? Would supporting the JSON formatted URLs be purely for it to be more convenient for other people that are using our APIs? Because if we were to support the var form in the API, could we just pass it straight from the web to the API?

> To me, specifying `?var.a=1` suggests that you want to filter the pipeline instances to those where the instance var `a` has value `1` - but what if there are multiple pipelines with `a=1`, but different values of `b`? In contrast, using `?instance_vars={...}` is unambiguous - it's clear that you have to specify the full set of instance vars. However, we could specify as a contract that the full set of `var.xxx=yyy` params must define the instance vars in full.

Ah I see, I think I was assuming that you could pass a filter to the query param and that would show a search result of all the instances that match but that might be a totally different endpoint. I do think you're right, it isn't obvious that it must be defined in full but to be completely honest I can definitely see users thinking that they can pass in a subset of the vars with the `instance_vars={..}` too 😆   or maybe thats just me!

> Let me know if that doesn't make any sense. TBH I'm leaning towards using the `?var.xxx=yyy` form for the API

I think my initial reaction is to also have the var format for the API? I do like the consistency and most importantly, the simplicity of the URLs without having all those URI encoded special characters. I think what you were saying earlier, of supporting both might be an option too if it is simple enough and does not make it more complex?

aoldershaw:
> Does this mean basically supporting both the JSON formatted URLs and the var format ... Would supporting the JSON formatted URLs be purely for it to be more convenient for other people that are using our APIs?

Yeah, I realized that the var form is really just a generalization of the instance vars form - it's really just a way of providing JSON using dot notation, since if `a.b = 123`, then `a = {"b": 123}`. So, it's in part for convenience, and in part to generalize the format.

Side note - I think it probably makes sense to rename `var` to `vars` - so, `vars.version="1.2.3"` is assigning the `version` field of the `vars` object. A bit more natural than `var`

> if we were to support the var form in the API, could we just pass it straight from the web to the API?

Kind of...but we still need to parse the query params in Elm, so it doesn't buy us much (it's just as easy to reconstruct them).

> Ah I see, I think I was assuming that you could pass a filter to the query param and that would show a search result of all the instances that match but that might be a totally different endpoint

I think that definitely makes sense for the `/api/v1/pipelines` endpoint, but I don't want to do that for all of the `/api/v1/teams/:team_name/pipelines/:pipeline_name` endpoints - while it could be cool to, say, archive all pipelines in the `releases` instance group that have `var.rc=true`, I'd rather this not be implemented in the API (and maybe make it the responsibility of `fly`)

Thanks for your feedback!

aoldershaw:
@clarafu I've updated the query parameters to match the UI, and also updated #6105 accordingly. This is ready for review. I'd ask you try to review this before #6105 if possible, since #6105 relies on the backend changes here.

There is still one piece of outstanding backend work related to instanced pipelines - updating the syslog drainer tag format. The format is currently `team_name/pipeline_name/job_name/build_name/plan_id`, but that doesn't accommodate instanced pipelines. I'm considering changing the format *for instanced pipelines only* to `team_name/pipeline_name/pipeline_id/job_name/build_name/plan_id`, but I'm not sure if that'll make consuming syslog more difficult. Anyway, I think I'll leave that for a subsequent PR

clarafu (approved):
Everything looks great to me! 👍 I looked over the code and ran a few acceptance tests for each of the points you mentioned as the changes within the PR.

- `fly pipelines` has an ID column 
```
~/w/c/c/web (issue/5921) $ fly -t dev pipelines
id  name                                 paused  public  last updated
1   test                                 no      no      2021-01-04 13:58:13 -0500 EST
2   test2/foo:bar                        yes     no      2021-01-04 13:58:23 -0500 EST
3   test2/foo.bar:1,foo.bar2:2           yes     no      2021-01-04 14:12:58 -0500 EST
4   test2/foo.bar:1,foo.bar2:2,foo2:bar  yes     no      2021-01-04 14:13:36 -0500 EST
5   test2/foo:"bar/bar",foo2:bar         yes     no      2021-01-04 14:15:39 -0500 EST
6   test2/foo:"bar/bar"                  yes     no      2021-01-04 14:16:01 -0500 EST
7   test2/foo2:"bar/bar"                 yes     no      2021-01-04 14:16:31 -0500 EST
```

- Instanced pipelines are using the `vars:...` format for the API endpoints
```
~/w/c/c/web (issue/5921) $ fly -t dev get-pipeline -p test2/foo:bar --verbose
2021/01/04 14:11:25 GET /api/v1/info HTTP/1.1
...
Accept-Encoding: gzip


2021/01/04 14:11:25 HTTP/1.1 200 OK
...

{"version":"0.0.0-dev","worker_version":"2.2","external_url":"http://localhost:8080","cluster_name":"dev"}

2021/01/04 14:11:25 GET /api/v1/teams/main/pipelines/test2/config?vars.foo=%22bar%22 HTTP/1.1
Host: localhost:8080
User-Agent: Go-http-client/1.1
Accept-Encoding: gzip
```

- slashes are supported within instance var values
```
~/w/c/c/web (issue/5921) $ fly -t dev sp -p test2 -i foo="bar/bar" -c ~/workspace/example-pipelines/trigger-resource.yml --verbose
...
Accept-Encoding: gzip


2021/01/04 14:16:23 HTTP/1.1 200 OK
...

2021/01/04 14:16:23 GET /api/v1/teams/main/pipelines/test2/config?vars.foo=%22bar%2Fbar%22 HTTP/1.1
Host: localhost:8080
User-Agent: Go-http-client/1.1
Accept-Encoding: gzip
...
~/w/c/c/web (issue/5921) $ fly -t dev pipelines
id  name                                 paused  public  last updated
...
6   test2/foo:"bar/bar"                  yes     no      2021-01-04 14:16:01 -0500 EST
```

PR #6185, opened 2020-10-20
Fix topgun/k8s tests

xtreme-sameer-vohra:

Fixes various outstanding bugs with our topgun/k8s suite of tests. Please see each commit message for details on every fix.

## Notes to reviewer:
Review each commit individually. Each commit fixes one set of tests..

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/topgun/k8s/baggageclaim_drivers_test.go
+++ b/topgun/k8s/baggageclaim_drivers_test.go
@@ -8,6 +8,9 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/storage/v1"
+	k8sErrs "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -46,14 +49,20 @@ var _ = Describe("baggageclaim drivers", func() {
 			It("successfully recreates the worker", func() {
 				By("deploying concourse with ONLY one worker and having the worker pod use the gcloud disk and format it with btrfs")
 
+				scName := "btrfs"
+				createBtrfsStorageClass(scName)
+
 				setReleaseNameAndNamespace("real-btrfs-disk")
+				pvcName := "disk-" + namespace
 
 				deployWithDriverAndSelectors("btrfs", UBUNTU,
 					"--set=persistence.enabled=false",
 					"--set=worker.additionalVolumes[0].name=concourse-work-dir",
-					"--set=worker.additionalVolumes[0].gcePersistentDisk.pdName=disk-topgun-k8s-btrfs-test",
-					"--set=worker.additionalVolumes[0].gcePersistentDisk.fsType=btrfs",
+					"--set=worker.additionalVolumes[0].persistentVolumeClaim.claimName="+pvcName,
 				)
+
+				// We create the PVC after deploying because we need the namespace to be created first by helm
+				createPVC(pvcName, scName)
 
 				atc := waitAndLogin(namespace, releaseName+"-web")
 				defer atc.Close()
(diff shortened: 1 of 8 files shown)


topgun/k8s/baggageclaim_drivers_test.go
```diff
@@ -46,15 +49,21 @@ var _ = Describe("baggageclaim drivers", func() {
 			It("successfully recreates the worker", func() {
 				By("deploying concourse with ONLY one worker and having the worker pod use the gcloud disk and format it with btrfs")
 
+				scName := "btrfs"
+				createBtrfsStorageClass(scName)
```
YoussB, line 53: not sure about this change. 
- from one perspective, it allows to create the storage class if it is not found.
- but from a security perspective, in order to achieve that the service account that is used should have admin privileges on the cluster, I am not sure about the one that's used now in the test, but if it doesn't have enough privileges, and I think it shouldn't, then the test will fail.

as an alternative, I am thinking maybe we can only check if the storage class doesn't exist then we can either fail the test or skip it 🤔

topgun/k8s/baggageclaim_drivers_test.go
```diff
@@ -46,15 +49,21 @@ var _ = Describe("baggageclaim drivers", func() {
...
 				deployWithDriverAndSelectors("btrfs", UBUNTU,
 					"--set=persistence.enabled=false",
 					"--set=worker.additionalVolumes[0].name=concourse-work-dir",
-					"--set=worker.additionalVolumes[0].gcePersistentDisk.pdName=disk-topgun-k8s-btrfs-test",
-					"--set=worker.additionalVolumes[0].gcePersistentDisk.fsType=btrfs",
+					"--set=worker.additionalVolumes[0].persistentVolumeClaim.claimName="+pvcName,
 				)
 
+				// We create the PVC after deploying because we need the namespace to be created first by helm
+				createPVC(pvcName, scName)
```
YoussB, line 65: I might not remember correctly, but didn't we use to get the PVC created when deploying the chart? Has something changed here or am I missing something?

YoussB (changes requested):
added a few small comments

topgun/k8s/baggageclaim_drivers_test.go
xtreme-sameer-vohra: This test requires a volume that is using `btrfs`. It expected the volume to already be created and available. However, the volume didn't exist on the topgun cluster, we believe it might have been deleted by mistake. The commit message has some more details.

topgun/k8s/baggageclaim_drivers_test.go
xtreme-sameer-vohra: Fair point regarding the security  aspect. Our tests rely on being able to create and manage namespaces already. 
What would be the recommended way of managing  the storage class so that we minimize manual intervention when someone targets any vanilla cluster ?

topgun/k8s/baggageclaim_drivers_test.go
YoussB: I think something along the lines of having the cluster already bootstrapped with the storage class. We are considering a similar approach [here](https://github.com/concourse/greenpeace/blob/a2e1ff274c57f1166a1eae4a385e818e3c041cc0/terraform/production/cluster.tf#L32-L42).

topgun/k8s/baggageclaim_drivers_test.go
YoussB: I am not super familiar with the roles and privileges of accounts, if creating a namespace requires the same privileges as creating a storage class, then for sure we can have the test create the storage class if it is not found.

YoussB: I am assuming a vanilla cluster wouldn't have the storage class, so the test wouldn't be runnable, thus we can skip the tests that depend on such storage class, maybe 🤔

xtreme-sameer-vohra: I looked at it briefly and it should be possible to provide much more [granular permissions](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) and its probably a good idea to have more restricted service accounts. Currently, our service accounts have carte blanche access and this PR doesn't extend that. We'd be able to provide the ability to a service account to just create/manage storageclass without giving it other admin privileges.  Should we track/triage that as a separate story as that has a different objective of cleaning up permissions?

YoussB: sure .. sounds good to me

YoussB (approved):
🚀 🚀 
P.S. I am not going to run the tests locally since this would a long time, I'd rather wait and see them go through the pipeline.

taylorsilva:
We ran them on CI and they passed. Merging this in now. Thanks Bishoy!

PR #6189, opened 2020-10-21
Fix check container placement, and only use check sessions for pipeline resource checks

vito:

* When creating a resource container for a resource type, only create it on a worker which supports the bottom-most base resource type. This is effectively the same constraint that existed before, and fixes the mistaken placement of resource containers on darwin/windows workers.
* When running a `check` for a resource that is not associated to a pipeline resource, create a container that's associated to the build, rather than using a resource config check session. Otherwise there's no way to intercept the container, because `/api/v1/containers` only lists check containers that are associated to resources.
* Fix up a few topgun tests which failed for not-very-good reasons when all these changes were introduced.
* Improve error wrapping to make troubleshooting a bit easier.
* Migrate topgun, testflight, and watsjs to use the `mock` resource type for image fetching - previously they'd fetch `busybox` and `alpine` images all over the place, which is slower, adds a network dependency, and will probably stop working once Docker Hub adds rate limits.
* Nuke `web/acceptance/` from orbit - it hasn't been used in years.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


vito:
fyi @taylorsilva @xtreme-sameer-vohra this *might* require more fixes in `k8s-topgun`. This makes it so that image `check` containers are associated to the build, meaning they will appear in `fly containers` and `fly intercept` container selection. Similar fixes were made in #6185 to handle the `get` container being present, but now a `check` container may be present as well.

atc/versioned_resource_types_test.go
```diff
@@ -0,0 +1,66 @@
...
+					Type:   "some-type",
+					Source: atc.Source{"nested": "source"},
+				},
+				Version: atc.Version{"nested": "version"},
+			},
+			{
+				ResourceType: atc.ResourceType{
+					Name:   "overriding-type",
+					Type:   "overriding-type",
+					Source: atc.Source{"overriding": "source"},
```
xtremerui, line 34: what is the type that this type trying to "override"?

atc/versioned_resource_types_test.go
```diff
@@ -0,0 +1,66 @@
...
+			})
+		})
+
+		Context("when the type is nested", func() {
+			It("returns the bottom type", func() {
+				Expect(types.Base("nested-type")).To(Equal("some-base-type"))
+			})
+		})
+
+		Context("when the type overrides a base type", func() {
```
xtremerui, line 60: Like above, not exactly sure how the type overrides the base type.

atc/versioned_resource_types_test.go
vito: It overrides it by specifying the same type as its own type. This effectively overrides `overriding-type` everywhere *else* in the pipeline, while using the real "base" `overriding-type` for fetching the type itself.

As a real world use case, imagine someone overriding `registry-image` with a newer `registry-image` - they have to use the old `registry-image` to fetch the new one.

xtremerui: uhh, I see. Should the type be named/typed with "some-base-type" ? So it matches the test case with context "when the type overrides a base type"

vito: Doing that would change the meaning of the other types, since one of them uses `some-base-type` - if I override it here, the other type would then refer to the overridden one. I can rename it to `overridden-base-type` tho!

vito: (done)

vito:
Pushed some more commits which swap out all the images in topgun, testflight, and watsjs to use the `mock` resource type's `mirror_self: true` functionality rather than fetching `busybox` or `alpine` everywhere.

I ran the full topgun suite with the included `./hack/bosh-topgun` script. Only one test failed, and it should pass as soon as the latest `mock` resource type bump makes it in to the BOSH release, so I expect all the `bosh-topgun-*` jobs to pass once this goes through the pipeline. 👍

xtremerui (approved):

Issue #6194, opened 2020-10-22
Support more auth providers through shiny Dex


xtremerui:
With the implementation of #6128, there are some new Dex conectors we could utilize to make Concourse to support more auth providers.

We should evaluate our user groups and the need for those new auth providers so we know the priority of adding Concourse auth flags here https://github.com/concourse/concourse/tree/master/skymarshal/skycmd

 - atlassiancrowd
 - authproxy
 - gitea
 - google
 - keystone
 - linkedIn
 - openshift

PR #6195, opened 2020-10-22
Bump dex module to v0.2.0

### LINKED ISSUE #6128: Bump concourse/dex dependency to latest release version
## Summary

Currently in go.mod the dependency of concourse/dex is a psedo-version v0.0.0-xxxxx that is obsolete compared to upstream dex. Now the pipeline https://ci.concourse-ci.org/teams/main/pipelines/update-dex will generate an up-to-date version of concourse/dex whenever there a release of upstream dex. So we need to bump the concourse/dex dependency so Dependabot could pick up the version updates from concourse/dex later.

Since we havn't been synced with upstream dex for a while. Errors is expected when we switch to concourse/dex@v0.1.0. Fix compilation errors so it pass all tests and go through the release notes of dexidp/dex since April 2019 (when concourse/dex last synced) to make sure concourse is good with all those updates.

xtremerui:

closes #6128 .

Update go module dependency of concourse/dex from pesudo version to semver. Made some changes in skymarshal to adapte changes from upstream.

The two new templates are used for Device Authoriation Grant flow which is not currently used by Concourse. But we need them for sanity check.

## Release Note
Concourse uses a fork of [Dex](github.com/dexidp/dex) as its auth module. This Dex fork had not being synced up with upstream since [v2.16.0](https://github.com/dexidp/dex/releases/tag/v2.16.0). Now it tracks the latest [v2.25.0](https://github.com/dexidp/dex/releases/tag/v2.25.0). There are lots of improvements and bug fixes. Also some new connectors are available (Google, OpenShift, Atlassiancrowd etc,), we will evaluate and support them in future releases. An issue is created for tracking https://github.com/concourse/concourse/issues/6194.

We expect no impact for Concourse auth from this Dex bump.

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


go.mod
```diff
@@ -100,16 +109,21 @@ require (
...
-	mvdan.cc/unparam v0.0.0-20190917161559-b83a221c10a2 // indirect
+	k8s.io/utils v0.0.0-20200414100711-2df71ebbae66 // indirect
 	sigs.k8s.io/yaml v1.2.0
 )
 
 go 1.13
 
 replace github.com/docker/distribution v2.7.1+incompatible => github.com/docker/distribution v2.7.1-0.20190205005809-0d3efadf0154+incompatible
+
+replace github.com/concourse/dex => ./dex
```
chenbh, line 129: Don't think this line should be here

xtremerui: Good catch! Thx!

xtremerui:
@aoldershaw this PR is ready. Thx!

atc/api/accessor/accessor.go
```diff
@@ -211,14 +210,18 @@ func (a *access) claim(name string) string {
...
 func (a *access) userID() string {
 	return a.federatedClaim("user_id")
 }
 
+func (a *access) userName() string {
+	if a.claim("preferred_username") != "" {
+		return a.claim("preferred_username")
+	}
+
+	return a.claim("name")
```
aoldershaw, line 222: If a user doesn't have a preferred username set, and assuming they have the ability to change their name, could they not spoof any account this way?

e.g. if a team has access granted to `github:aoldershaw`, could someone change their name (not username) in GitHub to `aoldershaw` and gain access to that team?

Also, the OIDC spec says about `preferred_username`:

> This value MAY be any valid JSON string including special characters such as @, /, or whitespace. The RP MUST NOT rely upon this value being unique

https://openid.net/specs/openid-connect-core-1_0.html

...so maybe even that isn't safe for comparison

xtremerui: I think it is the responsibility of each individual connector to make sure it maps an unique value to `preferred_username` in the id token by default, e.g. cf connector
https://github.com/concourse/dex/blob/417c2f3e2edd95a3a1d83935401d34481e5bdfb4/connector/cf/cf.go#L351

github connector
https://github.com/concourse/dex/blob/417c2f3e2edd95a3a1d83935401d34481e5bdfb4/connector/github/github.go#L320

if `preferred_username` is not set, like saml connector, the fall back to `name` behaves the same as it was since now
`name` claim in the raw token is set from `claim.Username` https://github.com/concourse/dex/blob/417c2f3e2edd95a3a1d83935401d34481e5bdfb4/server/oauth2.go#L346 

that was used to be added into ferated claims in concourse dex fork. See the diff of file oauth2.go 
https://github.com/concourse/dex/commit/bce8d6e0bcaed0243374d3526596bfe3c10ac740

skymarshal/web/templates/device.html
```diff
@@ -0,0 +1,23 @@
+{{ template "header.html" . }}
+
+<div class="theme-panel">
```
chenbh, line 3: Should we replace these 2 with a "Something has gone horribly wrong, please file an issue on concourse/concourse" page? It might be better to render this as an error page if the user should never see it.

atc/api/accessor/accessor.go
aoldershaw: Ah, I think I was a bit confused about what the `name` claim is - I was thinking of it as the user's full name, but dex seems to use the `username` - which seems safer to fallback to - https://github.com/concourse/dex/blob/417c2f3e2edd95a3a1d83935401d34481e5bdfb4/server/oauth2.go#L346

I did notice that there are a few connectors that don't support the `preferred_username` claim, but we support in Concourse:

* BitBucket Cloud
* LDAP (though https://github.com/concourse/dex/commit/c1b421fa0460c5aca9c23c7b56501b64d577f5cc added a configuration option - perhaps we should expose it at some point?)
* Microsoft
* OAuth + OIDC
  * They do support it, but I don't think we expose the configuration in Concourse properly
  * We support configuring the `--user-name-key`, but if there's a `preferred_username` set, that will take precedence, meaning the `--user-name-key` won't have any effect I believe
  * Also, it looks like there's a bug in our OAuth connector - we have the config option `preferredUsernameKey` (https://github.com/concourse/dex/blob/417c2f3e2edd95a3a1d83935401d34481e5bdfb4/connector/oauth/oauth.go#L54), but we never use it for constructing the connector
* SAML

I suspect most of these are okay - if we don't set the `preferred_username`, we should fallback to the username (in the `name` claim). I'll take a look at how we were setting the `user_name` in the `federated_claims` to make sure it's the same.

I'm a bit concerned about the OAuth + OIDC connectors, though - do you think it'll cause problems?

skymarshal/web/templates/device.html
aoldershaw: Seems reasonable to me - I suspect this isn't something we'll choose to support, at least not for a while

atc/api/accessor/accessor.go
xtremerui: For OAuth+OIDC connector the user-name-key flag is flexible as it was before, so operators can stil keep the same config. IMHO exposing `preferredUsername` config is not neccessary for Concourse as it only needs an username to match the team config for auth (so we dont expose `email` config either). Concourse doesn't need both `name` and `preferredusername` claim to be available like some other apps to use `name` as a display name and use `preferredUsername` for auth.

The bug in OAuth connector is actually a miss sync between the pr and master branch in concourse/dex. In the oauth connector PR it uses a claim mapping as OIDC connector does already. I will fix this. The dex module will bump to 0.3.0 in this case. Thx!

aoldershaw: Suppose someone had configured the OAuth connector, where the downstream OAuth provider gives both a `name` claim and a `preferred_username` claim. Prior to bumping dex, Concourse would just ignore the `preferred_username` claim if I'm not mistaken - but now, I think we'd use the `preferred_username` instead, which is probably more correct, but does seem to be breaking.

Additionally, suppose they had set `--oauth-user-name-key=other_claim`. If the downstream OAuth provider sets the `preferred_username` field, we'd ignore the `other_claim` field as the username because Concourse first checks the `preferred_username`, and then falls back to `name` (which would be set to the value of `other_claim`). If we don't want to expose `--oauth-preferred-username-key`, should we at least configure the OAuth connector to use the `user-name-key` as the `preferred-username-key` as well?

i.e.
https://github.com/concourse/concourse/blob/4cb7612c5319d87208564f4ccd2199923b810d6e/skymarshal/skycmd/oauth_flags.go#L87

```diff
	return json.Marshal(oauth.Config{
		ClientID:           flag.ClientID,
		ClientSecret:       flag.ClientSecret,
		AuthorizationURL:   flag.AuthURL,
		TokenURL:           flag.TokenURL,
		UserInfoURL:        flag.UserInfoURL,
		Scopes:             flag.Scopes,
		GroupsKey:          flag.GroupsKey,
		UserIDKey:          flag.UserIDKey,
		UserNameKey:        flag.UserNameKey,
+		PreferredUsernameKey: flag.UserNameKey,
		RootCAs:            caCerts,
		InsecureSkipVerify: flag.InsecureSkipVerify,
		RedirectURI:        redirectURI,
	})
```

aoldershaw (approved):
It's hard for me to be confident that this won't break anything given the lack of integration tests for the various auth providers, and without a total understanding of all the changes that went into dex, but I can't find anything that's broken, so ... I guess we'll see 🤷

xtremerui:
we could probably try it on CI and hush-house

PR #6197, opened 2020-10-23
Prevent retrying on worker error when build is aborted

evanchaoli:

During my local test, I found that, in some case, when some workers dropped into bad state, steps (in my test it was `task` step) required volumes from the bad worker, then the build just fallen into a endless loop of retries, and the build couldn't be aborted.

To not retry a build if it's aborted.

## Notes to reviewer:

Nothing

## Release Note

Fixed a endless build retry bug.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/engine/engine.go
+++ b/atc/engine/engine.go
@@ -238,7 +238,7 @@ func (b *engineBuild) Run(ctx context.Context) {
 		logger.Info("releasing")
 
 	case <-done:
-		if errors.As(runErr, &exec.Retriable{}) {
+		if errors.As(runErr, &exec.Retriable{}) && !b.build.IsAborted() {
 			return
 		}
 
--- a/atc/engine/engine.go
+++ b/atc/engine/engine.go
@@ -238,7 +238,7 @@ func (b *engineBuild) Run(ctx context.Context) {
 		logger.Info("releasing")
 
 	case <-done:
-		if errors.As(runErr, &exec.Retriable{}) && !b.build.IsAborted() {
+		if errors.As(runErr, &exec.Retriable{}) {
 			return
 		}
 
--- a/atc/exec/on_success.go
+++ b/atc/exec/on_success.go
@@ -35,11 +35,5 @@ func (o OnSuccessStep) Run(ctx context.Context, state RunState) (bool, error) {
 		return false, nil
 	}
 
-	select {
-	case <-ctx.Done():
-		return false, context.Canceled
-	default:
-	}
-
 	return o.hook.Run(ctx, state)
 }
--- a/atc/exec/on_success.go
+++ b/atc/exec/on_success.go
@@ -35,5 +35,11 @@ func (o OnSuccessStep) Run(ctx context.Context, state RunState) (bool, error) {
 		return false, nil
 	}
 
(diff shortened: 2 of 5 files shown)


atc/engine/engine.go
```diff
@@ -238,7 +238,7 @@ func (b *engineBuild) Run(ctx context.Context) {
 		logger.Info("releasing")
 
 	case <-done:
-		if errors.As(runErr, &exec.Retriable{}) {
+		if errors.As(runErr, &exec.Retriable{}) && !b.build.IsAborted() {
```
vito, line 241: Makes sense, but I think it might be cleaner to push this logic down into the `RetryErrorStep` where it can check if `ctx.Err() != nil` to avoid wrapping it in a `Retriable` in the first place.

The `RetryStep` does something similar:

https://github.com/concourse/concourse/blob/0e8560d9b2ca285870a752a43fa0954642ac7297/atc/exec/retry_step.go#L30-L32

vito (changes requested):
Makes sense - thanks for the PR! Have one suggestion code-wise. (Will also need tests.)

atc/engine/engine.go
evanchaoli: Good point. Just updated the code change.

evanchaoli:
@vito please hold on a day to review my update. I'll try to reproduce the endless loop retry problem and see if this fix works tomorrow.

evanchaoli:
@vito Today I tried but failed to reproduce the endless build retry problem. But I added log that proved the newly added check could be hit and aborted builds as expected. So, please continue reviewing this PR. A test is added.

evanchaoli:
I also added a check of abort in OnSuccess step because it wires all steps, the check may ensure build to abort as early as possible.

vito (changes requested):
@evanchaoli I think I might have run into the same issue you found - I pushed a fix for it in #6237: https://github.com/concourse/concourse/pull/6237/commits/8c4b5d4a3d989ea4c7be426cd5e9fbea3e9ded06

The retry loop would happen if a timeout (e.g. `timeout: 5s`) was hit while talking to BaggageClaim - the `retryhttp` package would consider the `context.DeadlineExceeded` error to be retryable (because it's `.Temporary()`) and just keep retrying. Does this sound like it might be the root cause of your issue?

atc/exec/on_success.go
```diff
@@ -35,5 +35,11 @@ func (o OnSuccessStep) Run(ctx context.Context, state RunState) (bool, error) {
 		return false, nil
 	}
 
+	select {
```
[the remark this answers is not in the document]
evanchaoli: I am not 100% sure my issue was caused by `retryhttp`, but it could be. Let me remove this piece of code to unblock this PR.

vito (approved):
lgtm - thanks!

PR #6192, opened 2020-10-22
Update last check end time on check failure/error

vito:
## Changes proposed by this PR:

follow-up for #6022

previously the check step would only update the last check end time when
the check succeeded. this meant that if the check failed or errored,
we'd actually check again immediately, which is a bit too aggressive.

this brings it back to the old behavior, which is to update the last
check end time on failure/error.

## Contributor Checklist

- [ ] Followed [Code of conduct], [Contributing Guide] & avoided [Anti-patterns]
- [ ] [Signed] all commits
- [ ] Added tests (Unit and/or Integration)
- [ ] Updated [Documentation]
- [ ] Added release note (Optional)

[Code of Conduct]: https://github.com/concourse/concourse/blob/master/CODE_OF_CONDUCT.md
[Contributing Guide]: https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md
[Anti-patterns]: https://github.com/concourse/concourse/wiki/Anti-Patterns
[Signed]: https://help.github.com/en/github/authenticating-to-github/signing-commits
[Documentation]: https://github.com/concourse/docs

## Reviewer Checklist

- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [ ] New config flags added? Ensure that they are added to the
  [BOSH](https://github.com/concourse/concourse-bosh-release) and
  [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for
  the [integration
  tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env)
  (for example, if they are Garden configs that are not displayed in the
  `--help` text).

PR #6203, opened 2020-10-26
Migrate build_id references to bigint, batch event migration

### LINKED ISSUE #6199: Migrate remaining foreign keys pointing to `builds (id)` from `integer` to `bigint`
[This migration](https://gist.github.com/vito/05a210b79082906280baff013e972428) should do, except for one problem: `build_events` will take _**forever**_ to migrate naively like this.

Consider adding a new column instead and incrementally migrating it. See [this post](https://techcommunity.microsoft.com/t5/azure-database-for-postgresql/postgres-tips-how-to-convert-2-billion-rows-to-bigint-with-citus/ba-p/1490128) as an example. Ideally we can have this happen transparently through some component that runs in the background so end-users don't have to do anything.

vito:

fixes #6199

follow-up to #6022

This introduces a handful of migrations which convert `build_id` foreign key references to `bigint` in-place. For `build_events` that approach would grind everything to a halt because it can have millions of rows, so this PR adds a batch migration component which will migrate build events in batches of 100,000 (by default) every 10 seconds (by default). Flags are also added for both settings just in case operators feel that it's too aggressive or too slow depending on their situation.

The migration first renames `build_events.build_id` to `build_events.build_id_old`, and then adds a "new" `build_id bigint` column and updates the indexes. This rename dance ensures other `web` nodes which may be running mid-upgrade will write their events as `bigints`, so that the batch migrator doesn't have to play "catch-up." Build events are migrated from newest to oldest. Older builds will just not have any logs until the migrator gets to them. When all events are migrated, a final schema cleanup is performed and the primary key is updated.

## Release Note

* This migration can be quite slow if you have a ton of builds, meaning `web` nodes may take a while to start upon upgrading.
* Build events are migrated at runtime in batches of 100,000 every 10 seconds, newest to oldest. Until this migration completes, older build logs will not be present.
  * The batch size can be configured with `--build-events-bigint-batch-size` (integer).
  * The batch interval can be configured with `--build-events-bigint-interval` (duration).

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -38,6 +38,7 @@ import (
 	"github.com/concourse/concourse/atc/db/encryption"
 	"github.com/concourse/concourse/atc/db/lock"
 	"github.com/concourse/concourse/atc/db/migration"
+	"github.com/concourse/concourse/atc/db/migration/batch"
 	"github.com/concourse/concourse/atc/engine"
 	"github.com/concourse/concourse/atc/engine/builder"
 	"github.com/concourse/concourse/atc/gc"
@@ -141,6 +142,9 @@ type RunCommand struct {
 
 	EncryptionKey    flag.Cipher `long:"encryption-key"     description:"A 16 or 32 length key used to encrypt sensitive information before storing it in the database."`
 	OldEncryptionKey flag.Cipher `long:"old-encryption-key" description:"Encryption key previously used for encrypting sensitive information. If provided without a new key, data is encrypted. If provided with a new key, data is re-encrypted."`
+
+	BuildEventsBigintMigrationBatchSize int           `long:"build-events-bigint-batch-size" default:"100000" description:"Number of events to migrate in each batch."`
+	BuildEventsBigintMigrationInterval  time.Duration `long:"build-events-bigint-interval" default:"10s" description:"Interval on which to migrate each batch."`
 
 	DebugBindIP   flag.IP `long:"debug-bind-ip"   default:"127.0.0.1" description:"IP address on which to listen for the pprof debugger endpoints."`
 	DebugBindPort uint16  `long:"debug-bind-port" default:"8079"      description:"Port on which to listen for the pprof debugger endpoints."`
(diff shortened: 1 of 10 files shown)


aoldershaw (changes requested):
Getting the following error when trying to migrate up from some other dev version:

```
web_1     | error: failed to migrate database: 1 error occurred:
web_1     | 	* Migration '1603401316_alter_build_integers_to_bigints.up.sql' failed: 2 errors occurred:
web_1     | 	* Transaction ALTER TABLE jobs
web_1     |       ALTER COLUMN next_build_id TYPE bigint,
web_1     |       ALTER COLUMN latest_completed_build_id TYPE bigint,
web_1     |       ALTER COLUMN transition_build_id TYPE bigint failed, rolled back the migration
web_1     | 	* pq: cannot alter type of a column used in a trigger definition
```

Not sure if my DB got in a weird state for some reason as a result of jumping around between dev versions. Not sure what column/trigger is causing issues, will try to investigate

vito:
@aoldershaw hmm maybe it's the dashboard event polling stuff?

aoldershaw (approved):
The batch migration seems to work well - I happened to have ~200,000 build events just lying around, so was able to see it work in batches, and it appeared to handle the 100,000 row updates without a sweat. I can see this taking a fairly long time on some massive deployments with the default batch settings, but that's not the end of the world, especially considering it starts with the most recent builds. I migrated down and back up several times, and that also worked well.

As a side note - it does seem concerning that I have 200,000 build events, given that I have only 9 resources and haven't been running any "builds" (they're pretty much all check builds). To be fair, my local deployment has been up for a while, checks were erroring due to a misconfigured var, and #6192 wasn't merged yet.

I see you have a comment about this in the linked issue:

> Which led me to discovering that build events don't have a foreign key reference for the build_id column. 🤔 This probably means that resource checking builds are leaking their build events forever. This is probably so we can just truncate/drop the whole table partition when a pipeline is deleted - but maybe we could just truncate it first so the cascade doesn't fire? Maybe we need a new garbage collector to remove build events for deleted builds? ...might be challenging to even find which ones to remove efficiently though. Hmm....

Is that something we should look into before 6.7? Rather than adding a GC component, could we just delete the events when we delete the previous resource check build? I recognize this won't work for builds that are deleted elsewhere, but at least the build events won't pile up as resource checking occurs

EDIT: Opened #6211

vito:
@aoldershaw Woops I forgot to post the results of the benchmark tests - my finding was that 100k was a good balance of update throughput and not taking too long for each batch. On my machine 100k row batches took ~1 second. Smaller batch sizes were faster but it took longer to migrate everything. I don't think you want to set it too high since it might affect performance of writing new events. Haven't tested it though.

PR #6205, opened 2020-10-26
Fix race condition with check build creation + starting

vito:

follow-up for #6022

previously the check factory would create a build and then start it in a
separate database call. if the web node was shut down in between, the
pending build would be left there, but nothing would ever start it, and
checking would just stop until someone comes in and creates a new build.

now the build will be created and started in a single transaction.

a better refactor IMO would be to introduce a `builds.Starter` and use
it to simplify the job build starting and serial groups logic. but i
ain't got time for that!

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/resource.go
+++ b/atc/db/resource.go
@@ -67,7 +67,7 @@ type Resource interface {
 	SetResourceConfigScope(ResourceConfigScope) error
 
 	CheckPlan(atc.Version, time.Duration, ResourceTypes, atc.Source) atc.CheckPlan
-	CreateBuild(context.Context, bool) (Build, bool, error)
+	CreateBuild(context.Context, bool, atc.Plan) (Build, bool, error)
 
 	NotifyScan() error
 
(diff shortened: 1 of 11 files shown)


atc/db/resource.go
```diff
@@ -337,6 +342,16 @@ func (r *resource) CreateBuild(ctx context.Context, manuallyTriggered bool) (Bui
 		return nil, false, err
 	}
 
+	err = r.conn.Bus().Notify(atc.ComponentBuildTracker)
```
aoldershaw, line 345: Is there a reason we don't notify the build events channel when we start resource/resource type checks, but do when we start "regular" builds? I guess the only time you'd actually want to monitor the build events is when you manually trigger a check, and the first event that comes occurs (presumably the initialize event when the build is picked up) will notify the channel - just wondering why they're different

aoldershaw (commented):
Makes sense, but left one small question

atc/db/resource.go
vito: Yeah - the other one notifies because the build already existed, so a user may have the page open. In this case it's a newly created build so there's no one around to notify.

aoldershaw (approved):

PR #6208, opened 2020-10-26
feat(atc): support chained container placement strategies.

evanchaoli:

Currently if a container placement strategy results multiple candidate workers, then it will randomly pick up one from the candidates.

Say, if `volume-locality` strategy results 2 candidate workers, why choose the one that has less containers or tasks?

With PR implements a strategy chain, so that if one strategy results multiple candidate workers, then next strategy can be applied.

Organizes multiple strategies as a chain.

## Notes to reviewer:

Please review from the design perspective first. If you are ok with the design, I'll add more tests for chained strategies. This enhancement should not break any existing strategy.

## Release Note

* Enhanced container placement strategy to support chained strategies, for example `CONCOURSE_CONTAINER_PLACEMENT_STRATEGY=volume-locality,fewest-build-containers`

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/worker/placement.go
+++ b/atc/worker/placement.go
@@ -89,14 +89,10 @@ func NewRandomPlacementStrategy() ContainerPlacementStrategy {
 	return s
 }
 
-type VolumeLocalityPlacementStrategyNode struct {
-	rand *rand.Rand
-}
+type VolumeLocalityPlacementStrategyNode struct{}
 
 func newVolumeLocalityPlacementStrategyNode() ContainerPlacementStrategyChainNode {
-	return &VolumeLocalityPlacementStrategyNode{
-		rand: rand.New(rand.NewSource(time.Now().UnixNano())),
-	}
+	return &VolumeLocalityPlacementStrategyNode{}
 }
 
 func (strategy *VolumeLocalityPlacementStrategyNode) Choose(logger lager.Logger, workers []Worker, spec ContainerSpec) ([]Worker, error) {
@@ -130,14 +126,10 @@ func (strategy *VolumeLocalityPlacementStrategyNode) ModifiesActiveTasks() bool 
 	return false
 }
 
-type FewestBuildContainersPlacementStrategyNode struct {
-	rand *rand.Rand
-}
+type FewestBuildContainersPlacementStrategyNode struct{}
 
 func newFewestBuildContainersPlacementStrategy() ContainerPlacementStrategyChainNode {
-	return &FewestBuildContainersPlacementStrategyNode{
-		rand: rand.New(rand.NewSource(time.Now().UnixNano())),
-	}
+	return &FewestBuildContainersPlacementStrategyNode{}
 }
 
 func (strategy *FewestBuildContainersPlacementStrategyNode) Choose(logger lager.Logger, workers []Worker, spec ContainerSpec) ([]Worker, error) {
@@ -160,13 +152,11 @@ func (strategy *FewestBuildContainersPlacementStrategyNode) ModifiesActiveTasks(
 }
 
 type LimitActiveTasksPlacementStrategyNode struct {
-	rand     *rand.Rand
 	maxTasks int
 }
 
 func newLimitActiveTasksPlacementStrategy(maxTasks int) ContainerPlacementStrategyChainNode {
 	return &LimitActiveTasksPlacementStrategyNode{
-		rand:     rand.New(rand.NewSource(time.Now().UnixNano())),
 		maxTasks: maxTasks,
 	}
 }
--- a/atc/worker/placement.go
+++ b/atc/worker/placement.go
@@ -12,7 +12,7 @@ import (
 )
 
 type ContainerPlacementStrategyOptions struct {
-	ContainerPlacementStrategy []string `long:"container-placement-strategy" default:"volume-locality" choice:"volume-locality" choice:"random" choice:"fewest-build-containers" choice:"limit-active-tasks" description:"Method by which a worker is selected during container placement. If multiple methods are specified, they will be applied in order. Random strategy is deprecated, it is still accepted, but takes no effective. Now, no strategy equals to random strategy."`
+	ContainerPlacementStrategy []string `long:"container-placement-strategy" default:"volume-locality" choice:"volume-locality" choice:"random" choice:"fewest-build-containers" choice:"limit-active-tasks" description:"Method by which a worker is selected during container placement. If multiple methods are specified, they will be applied in order. Random strategy should only be used alone."`
 	MaxActiveTasksPerWorker    int      `long:"max-active-tasks-per-worker" default:"0" description:"Maximum allowed number of active build tasks per worker. Has effect only when used with limit-active-tasks placement strategy. 0 means no limit."`
 }
 
@@ -38,7 +38,7 @@ func NewContainerPlacementStrategy(opts ContainerPlacementStrategyOptions) (*con
 		strategy := strings.TrimSpace(strategy)
 		switch strategy {
 		case "random":
-			cps.nodes = append(cps.nodes, newRandomPlacementStrategyNode())
+			// Add nothing. Because an empty strategy chain equals to random strategy.
 		case "fewest-build-containers":
 			cps.nodes = append(cps.nodes, newFewestBuildContainersPlacementStrategy())
 		case "limit-active-tasks":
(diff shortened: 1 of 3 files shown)


xtreme-sameer-vohra (changes requested):

atc/worker/placement.go
```diff
@@ -1,31 +1,104 @@
...
+	for _, node := range strategy.nodes {
+		workers, err = node.Choose(logger, workers, spec)
+		if err != nil {
+			return nil, err
+		}
+		if len(workers) == 0 {
+			return nil, nil
+		}
+	}
+	if len(workers) == 1 {
```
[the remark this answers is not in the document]
xtreme-sameer-vohra: Nvm, I can see this needs to be where it is in case the limit active tasks strategy is being used.

xtreme-sameer-vohra:
Thanks for the PR @evanchaoli 
This looks great.

cc @kcmannem who had dreamed this up many 🌔s ago

evanchaoli:
@xtreme-sameer-vohra Thanks for your review. I'll add more test cases (I think I mentioned that in "Notes to reviewer"), just haven't got a chance to get back on this PR yet.

xtreme-sameer-vohra:
> @xtreme-sameer-vohra Thanks for your review. I'll add more test cases (I think I mentioned that in "Notes to reviewer"), just haven't got a chance to get back on this PR yet.

Sounds good. Feel free to include the changes here https://github.com/concourse/concourse/pull/6233

atc/worker/placement.go
```diff
@@ -1,31 +1,104 @@
...
+		case "random":
+			cps.nodes = append(cps.nodes, newRandomPlacementStrategyNode())
+		case "fewest-build-containers":
+			cps.nodes = append(cps.nodes, newFewestBuildContainersPlacementStrategy())
+		case "limit-active-tasks":
+			if opts.MaxActiveTasksPerWorker < 0 {
+				return nil, errors.New("max-active-tasks-per-worker must be greater or equal than 0")
+			}
+			cps.nodes = append(cps.nodes, newLimitActiveTasksPlacementStrategy(opts.MaxActiveTasksPerWorker))
+		default:
```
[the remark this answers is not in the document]
evanchaoli: I just copied the original code, so that is the current Concourse's behavior. But I agree explicit is better. Fixed.

atc/worker/placement.go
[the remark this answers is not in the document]
evanchaoli: Yes, in the loop, when there is only 1 worker left, next strategy should still be applied, which may further filter out workers.

evanchaoli:
@xtreme-sameer-vohra I took the test case you added from #6233. Also addressed the review comments. So I think this PR is ready to ship. Please take a look again.

atc/worker/placement.go
```diff
@@ -1,31 +1,104 @@
...
+	for _, strategy := range strategies {
+		strategy := strings.TrimSpace(strategy)
+		switch strategy {
+		case "random":
+			cps.nodes = append(cps.nodes, newRandomPlacementStrategyNode())
+		case "fewest-build-containers":
+			cps.nodes = append(cps.nodes, newFewestBuildContainersPlacementStrategy())
+		case "limit-active-tasks":
+			if opts.MaxActiveTasksPerWorker < 0 {
+				return nil, errors.New("max-active-tasks-per-worker must be greater or equal than 0")
```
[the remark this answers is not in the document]
evanchaoli: Added a test.

evanchaoli:
@muntac Can you please help move forward this PR?

atc/worker/placement.go
```diff
@@ -1,31 +1,106 @@
...
 	"math/rand"
+	"strings"
 	"time"
 
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc/db"
 )
 
+type ContainerPlacementStrategyOptions struct {
+	ContainerPlacementStrategy string `long:"container-placement-strategy" default:"volume-locality" choice:"volume-locality" choice:"random" choice:"fewest-build-containers" choice:"limit-active-tasks" description:"Method by which a worker is selected during container placement. Multiple methods can be chained by +, e.g. volume-locality+fewest-build-containers"`
```
aoldershaw, line 15: Perhaps it'd be better to make this a `[]string` (and rename the field to `ContainerPlacementStrategies`, but keep the flag as is).

You'd then configure `concourse web` with:

```
--container-placement-strategy=volume-locality --container-placement-strategy=fewest-build-containers
```

or

```
CONCOURSE_CONTAINER_PLACEMENT_STRATEGY=volume-locality,fewest-build-containers
```

That way, it's consistent with other flags that accept multiple values

evanchaoli: Make sense. I just didn't realize `[]string` can be used. 😅

atc/worker/placement.go
```diff
@@ -123,31 +194,27 @@ func (strategy *LimitActiveTasksPlacementStrategy) Choose(logger lager.Logger, w
...
+func newRandomPlacementStrategyNode() ContainerPlacementStrategyChainNode {
+	return &RandomPlacementStrategyNode{
 		rand: rand.New(rand.NewSource(time.Now().UnixNano())),
 	}
 }
 
-func (strategy *RandomPlacementStrategy) Choose(logger lager.Logger, workers []Worker, spec ContainerSpec) (Worker, error) {
-	return workers[strategy.rand.Intn(len(workers))], nil
+func (strategy *RandomPlacementStrategyNode) Choose(logger lager.Logger, workers []Worker, spec ContainerSpec) ([]Worker, error) {
+	return []Worker{workers[strategy.rand.Intn(len(workers))]}, nil
```
aoldershaw, line 215: Perhaps this should return all workers and let `containerPlacementStrategy.Choose()` deal with selecting at random.

Though there's no good reason to do it, if you specified `CONCOURSE_CONTAINER_PLACEMENT_STRATEGY= random,volume-locality`, the `volume-locality` strategy would have no effect, since we already filtered the list down to a single worker.

So, `random` is in a bit of a weird spot. It might be more valuable to view these as `XYZPlacementStrategyNodes` as filters - in which case, `RandomPlacementStrategyNode` is really an `IdentityFilter` or something like that - i.e. it just returns the full list of workers that it's provided.

atc/worker/placement.go
```diff
@@ -49,26 +123,24 @@ func (strategy *VolumeLocalityPlacementStrategy) Choose(logger lager.Logger, wor
...
 }
 
-func (strategy *VolumeLocalityPlacementStrategy) ModifiesActiveTasks() bool {
+func (strategy *VolumeLocalityPlacementStrategyNode) ModifiesActiveTasks() bool {
 	return false
 }
 
-type FewestBuildContainersPlacementStrategy struct {
+type FewestBuildContainersPlacementStrategyNode struct {
 	rand *rand.Rand
```
aoldershaw, line 134: Looks like we can get rid of `rand` here, as well as in `LimitActiveTasksPlacementStrategyNode` and possibly `RandomPlacementStrategyNode` (see my comment there)

evanchaoli: Good catch.

atc/worker/placement.go
evanchaoli: This makes sense. I thought to keep `random` strategy for backward-compatibility. Now, I think we can just deprecate `random` strategy, and states that in the option description.

aoldershaw: I don't know if we should get rid of random altogether - it just doesn't totally make sense with chaining. I think just making the RandomPlacementStrategyNode return the full list of workers is sufficient, though, since it won't lead to any unexpected behaviour with chaining

evanchaoli:
@aoldershaw Thanks for your review and comments. I have addressed your comments, please take a look again.

atc/worker/placement.go
evanchaoli: I don't remove "random" in this PR. I just added option description to state deprecation.

atc/worker/placement.go
```diff
@@ -1,31 +1,101 @@
...
 	"math/rand"
+	"strings"
 	"time"
 
 	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc/db"
 )
 
+type ContainerPlacementStrategyOptions struct {
+	ContainerPlacementStrategy []string `long:"container-placement-strategy" default:"volume-locality" choice:"volume-locality" choice:"random" choice:"fewest-build-containers" choice:"limit-active-tasks" description:"Method by which a worker is selected during container placement. If multiple methods are specified, they will be applied in order. Random strategy is deprecated, it is still accepted, but takes no effective. Now, no strategy equals to random strategy."`
```
aoldershaw, line 15: I think we should actually get rid of the deprecation notice - although it's true that no strategy is the same as a random strategy, I think it's not possible to specify no strategy, given that there's a default of `volume-locality`.

So, I don't think we'll even want to deprecate `random`, since operators should be allowed to use completely random placement if they so choose - but maybe we can mention that chaining `random` is redundant

evanchaoli: OK, make sense.

aoldershaw (approved):
Thanks! LGTM

PR #6222, opened 2020-10-27
go-concourse: Switch go-concourse/configs.go to use HTTPAgent

andy-paine:

Migrate go-concourse/configs.go away from deprecate Connection to HTTPAgent instead. This allows fly to correctly report the error messages returned when receiving a 403 which includes useful errors about why policy checks failed.

Switch to using team.httpagent rather than team.connection. Migrate logic that is missing from httpagent as consumers are expected to implement it.

## Notes to reviewer:
Not sure if there is a deprecation plan for the Connection object but this seems easy enough

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


jamieklassen (changes requested):
Nice change! There's no real deprecation plan for connection, and fly is the only consumer that we really take seriously. Those comments are in place mostly to remind anybody who blunders by this part of the code to try to treat fly as an HTTP client rather than a TCP client, since it was modelled after the (unix domain socket-compatible) [garden client](https://github.com/cloudfoundry/garden/blob/b404ff2d61e689c6510593cf75d39dc5311be663/client/connection/connection.go#L587-L593) way back when.

The CI job that failed was the result of some fairly brittle tests. I found a patch that should help with those, and actually collapses two assertions into one:
```patch
diff --git a/atc/integration/rbac_test.go b/atc/integration/rbac_test.go
index e13c92870..0799e7e5e 100644
--- a/atc/integration/rbac_test.go
+++ b/atc/integration/rbac_test.go
@@ -65,8 +65,7 @@ jobs:
 					ccClient := login(atcURL, "v-user", "v-user")
 
 					_, _, _, err := ccClient.Team(team.Name).CreateOrUpdatePipelineConfig(atc.PipelineRef{Name: "pipeline-new"}, "0", pipelineData, false)
-					Expect(err).To(HaveOccurred())
-					Expect(err.Error()).To(Equal("forbidden"))
+					Expect(err).To(MatchError(ContainSubstring("forbidden")))
 				})
 			})
 
@@ -83,8 +82,7 @@ jobs:
 					ccClient := login(atcURL, "po-user", "po-user")
 
 					_, _, _, err := ccClient.Team(team.Name).CreateOrUpdatePipelineConfig(atc.PipelineRef{Name: "pipeline-new"}, "0", pipelineData, false)
-					Expect(err).To(HaveOccurred())
-					Expect(err.Error()).To(Equal("forbidden"))
+					Expect(err).To(MatchError(ContainSubstring("forbidden")))
 				})
 			})

```

Finally, could you add a unit test for this new behaviour? I think a sibling context similar to

https://github.com/concourse/concourse/blob/e2831b531452357410a064dd4accce78fc6e87b5/go-concourse/concourse/configs_test.go#L360-L384

except with the name `when setting config returns forbidden` (or even `an unexpected status` instead of `forbidden`) would be handy, mainly as a signpost to unambiguously indicate that the `CreateOrUpdatePipelineConfig` method is supposed to return an error message containing the response body on unexpected responses codes.

Otherwise I spun up a docker-compose with `hack/overrides/opa.yml` and uncommented all the stuff I was supposed to and got exactly the behaviour you advertised -- and it's great!

go-concourse/concourse/configs.go
```diff
@@ -79,46 +75,38 @@ func (team *team) CreateOrUpdatePipelineConfig(pipelineRef atc.PipelineRef, conf
...
+		var validationErr atc.SaveConfigResponse
+		body, _ := ioutil.ReadAll(response.Body)
+		err = json.Unmarshal(body, &validationErr)
+		if err != nil {
+			return false, false, []ConfigWarning{}, internal.UnexpectedResponseError{
+				StatusCode: response.StatusCode,
+				Status:     response.Status,
+				Body:       string(body),
+			}
+		}
```
jamieklassen, line 95: Not your fault, but this a bit clunky, eh? We know enough about the config server to know that there are really two distinct scenarios here:
* you get a 400 status and an `atc.SaveConfigResponse`-shaped body
* you get a non-400 status and an unstructured body which [may](https://github.com/concourse/concourse/blob/e2831b531452357410a064dd4accce78fc6e87b5/atc/api/configserver/save.go#L113-L114) [be](https://github.com/concourse/concourse/blob/e2831b531452357410a064dd4accce78fc6e87b5/atc/api/configserver/save.go#L119-L120) [empty](https://github.com/concourse/concourse/blob/e2831b531452357410a064dd4accce78fc6e87b5/atc/api/configserver/save.go#L60-L61) or just contain an [error string](https://github.com/concourse/concourse/blob/e2831b531452357410a064dd4accce78fc6e87b5/atc/api/configserver/save.go#L126-L128)

How would you feel about casing on the specific status, adding a branch for `BadRequest`/`InvalidConfigError` and a branch for other non-200/`UnexpectedResponseError`?

andy-paine: That sounds good. I think there are a couple more things it can return due to the api wrappas like this https://github.com/concourse/concourse/blame/master/atc/api/policychecker/handler.go#L45-L50. The specific thing I wanted here was an actual error message for 403s since OPA can provide useful 403s now. Would you be open to that being a specific new case and handing back a slightly more strongly typed Forbidden error in that case too?

xtremerui (approved):
LGTM thx!

PR #6235, opened 2020-10-30
Check step only acquires lock for periodic pipeline resource checks

### LINKED ISSUE #6217: task check image hangs for long time
## Summary

During my debugging of some issue, workers were under heavy load, very slow, then on the GUI, I saw task's check image just hang for long time:

task-check-image-hang

@vito  PR #6153 is trying to show what is going on during image fetching, but in this case, there still didn't show any message, moreover, I couldn't abort the build.

My task just used `busybox` image that is tiny, and the image had been cached to an internal registry, thus DockerHub rate limit should not matter at all in my test.

One more confusion is, as I used a timer to trigger builds, and my tests used only `busybox` image, thus all workers should already cached the image, so that image fetching supposed to be skipped or quickly finished, why `check` still hang up here?

## Triaging info

* Concourse version: based on master branch
* Browser (if applicable):
* Did this used to work?

vito:

fixes #6217

* Cleans up all the panics from test output.
* Don't re-use resource type check containers, since they really shouldn't need caching semantics (they're almost always `registry-image`).
* Only acquires the resource config scope lock when checking a resource.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/builds/tracker_test.go
+++ b/atc/builds/tracker_test.go
@@ -2,6 +2,7 @@ package builds_test
 
 import (
 	"context"
+	"io/ioutil"
 	"testing"
 	"time"
 
@@ -11,9 +12,14 @@ import (
 	"github.com/concourse/concourse/atc/db/dbfakes"
 	"github.com/concourse/concourse/atc/engine"
 	"github.com/concourse/concourse/atc/engine/enginefakes"
+	"github.com/concourse/concourse/atc/util"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 )
+
+func init() {
+	util.PanicSink = ioutil.Discard
+}
 
 type TrackerSuite struct {
 	suite.Suite
--- a/atc/builds/tracker.go
+++ b/atc/builds/tracker.go
@@ -2,9 +2,6 @@ package builds
 
 import (
 	"context"
-	"fmt"
-	"os"
-	"runtime/debug"
 	"sync"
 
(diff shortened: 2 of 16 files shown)


atc/builds/tracker_test.go
```diff
@@ -11,10 +12,15 @@ import (
 	"github.com/concourse/concourse/atc/db/dbfakes"
 	"github.com/concourse/concourse/atc/engine"
 	"github.com/concourse/concourse/atc/engine/enginefakes"
+	"github.com/concourse/concourse/atc/util"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 )
 
+func init() {
+	util.PanicSink = ioutil.Discard
```
muntac, line 21: 👍

muntac, line 21: 👍

muntac (approved):
Looks good! Application-wise not fully sure what checking something that is not a resource means as I assumed the check step runs only to find new versions for a resource.

vito:
@muntac Thx for the review! Updated the title, hopefully that's more clear. Everything that is checked is indeed a resource by definition. The difference is that we only need to lock when performing the periodic check for a resource defined in a pipeline, because the check container is currently re-used.

PR #6237, opened 2020-10-30
Don't enforce timeouts during image fetching

vito:

Feature

* Adds a `timeout:` field to the `task`, `get`, and `put` steps which gets parsed before the more general `timeout:` step modifier. This allows the `task`/`get`/`put` steps to enforce the timeout more narrowly - around the running of the actual process, so that it doesn't include the time spent fetching images.
* Bumps `retryhttp` to v1.1.0 to fix an issue I noticed during testing: if the timeout is reached while making a request to Baggageclaim, `retryhttp` considered the `context.DeadlineExceeded` error to be retryable, so rather than timing out it'd go into a retry loop.
* Tweaks the `check` step timeout to be consistent with `task`/`get`/`put`, i.e. enforced at the step level rather than being passed to the `worker.Client`.

## Notes to reviewer:

I'm not 100% sure if this is a good idea. It feels more technically correct, because it's awkward to have to factor in time spent fetching images when determining an appropriate timeout, given that they'll be cached half the time. But it does make it harder to enforce timeouts on image fetching. You could work around it by explicitly using the `timeout:` step modifier, I guess.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/exec/get_step.go
+++ b/atc/exec/get_step.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"time"
 
 	"code.cloudfoundry.org/lager"
 	"code.cloudfoundry.org/lager/lagerctx"
@@ -205,8 +206,20 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
 
 	containerOwner := db.NewBuildStepContainerOwner(step.metadata.BuildID, step.planID, step.metadata.TeamID)
 
+	processCtx := ctx
+	if step.plan.Timeout != "" {
+		timeout, err := time.ParseDuration(step.plan.Timeout)
+		if err != nil {
+			return false, fmt.Errorf("parse timeout: %w", err)
+		}
+
+		var cancel func()
+		processCtx, cancel = context.WithTimeout(ctx, timeout)
+		defer cancel()
+	}
+
 	getResult, err := step.workerClient.RunGetStep(
-		ctx,
+		processCtx,
 		logger,
 		containerOwner,
 		containerSpec,
--- a/atc/exec/get_step.go
+++ b/atc/exec/get_step.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"time"
 
 	"code.cloudfoundry.org/lager"
 	"code.cloudfoundry.org/lager/lagerctx"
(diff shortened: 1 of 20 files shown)


vito:
Side note: this should make #3283 a lot easier to implement, since it can be done in the exact same way as `--global-resource-check-timeout`.

atc/exec/get_step.go
```diff
@@ -205,8 +207,20 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
 
 	containerOwner := db.NewBuildStepContainerOwner(step.metadata.BuildID, step.planID, step.metadata.TeamID)
 
+	processCtx := ctx
+	if step.plan.Timeout != "" {
```
xtreme-vikram-yadav, line 211: A minor issue and i have mixed feeling about this but this could be extracted into a shared method.

vito: Yeah, now that this is consistent between all the steps it might be easy to DRY it up. I'll take a look!

atc/exec/task_step.go
```diff
@@ -264,25 +264,20 @@ func (step *TaskStep) run(ctx context.Context, state RunState, delegate TaskDele
...
 
-	if err != nil {
-		if err == context.Canceled || err == context.DeadlineExceeded {
-			step.registerOutputs(logger, repository, config, result.VolumeMounts, step.containerMetadata)
-		}
-		return false, err
-	}
-
-	delegate.Finished(logger, ExitStatus(result.ExitStatus))
-
```
chenbh, line 262: At first I was hesitant about caching stuff from an errored task since it could've left the cache in an invalid state. But then I realized we also register caches on a failed task (which could also leave the cache invalid). Guess it's just up to the user to deal with potentially corrupted caches

chenbh (approved):
lgtm, I agree that it could be extracted into a shared method, but I'll approve this just in case

atc/exec/task_step.go
vito: Yeah, I think this is worth the trade-off since it's annoying if e.g. a task fetches a bunch of dependencies and then flakes out or something. We'll see if it's problematic. There's a `fly clear-task-cache` command now which should also help.

vito:
rebased + added a `MaybeTimeout` helper

PR #6254, opened 2020-11-09
Fix up test data race for build tracker

vito:
rather than doing a full DeepEqual on the entire fake build, just
compare IDs

this failed with --race because the goroutines call .ID(), which with a
fake object causes mutation for all the internal bookkeeping around call
count / call args, resulting in a data race.


chenbh (approved):
lgtm, I had trouble reproducing the data race until I used a docker container. Turns out there's differences between mac and linux versions of `go test`

PR #6262, opened 2020-11-11
Add docker-compose overrides for SAML, LDAP, OAuth, and OIDC

chenbh:
The SAML one might be a bit weird to work with, but the others should be straight forward and mimics what we can expect from a real life scenario.

### DIFF
--- a/hack/overrides/oidc.yml
+++ b/hack/overrides/oidc.yml
@@ -1,4 +1,4 @@
-# oidc.yml - a docker-compose override that adds a oidc IDP to the stack
+# oidc.yml - a docker-compose override that adds a OIDC auth to the stack
 #
 # There are 2 users and 1 group:
 # user1@example.com:user1pass;group1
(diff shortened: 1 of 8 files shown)


chenbh:
Going to mark this as draft while we try to get some more auth connectors docker-compose'd

chenbh:
@taylorsilva this is now ready for review

hack/overrides/oidc.yml
```diff
@@ -0,0 +1,33 @@
...
+version: '3'
+
+services:
+  web:
+    environment:
+      # CONCOURSE_MAIN_TEAM_OIDC_USER: user1
+      CONCOURSE_MAIN_TEAM_OIDC_USER: John
+      # CONCOURSE_MAIN_TEAM_OIDC_GROUP: group1
+
+      CONCOURSE_OIDC_ISSUER: http://oidc:9000
```
taylorsilva, line 19: A little bump in the dev experience here but not much we can do about it :(
Can we add a comment to the top explaining that when logging in with this override file they'll be directed to `http://oidc:9000` and they'll need to replace `oidc` with `localhost` to continue logging in. I can see someone that doesn't used docker often being confused by the behaviour

taylorsilva (changes requested):
Everything worked perfectly! One nit

chenbh:
Updated the comment, and also made the `./hack/dc` script not docker-compose up **EVERYTHING**

taylorsilva (approved):
lgtm

PR #6281, opened 2020-11-17
atc(feat): master - skip checking put-only resources

evanchaoli:

This is a re-submit of #5149 that was released in 6.0.0 and reverted in 6.6.0 because of two side effects:

1. Put-only resource's custom resource type is not auto checked.
2. Put-only resources don't show version history on the UI.

This PR is going to re-enable #5149 and solves the side effects.

* Pick up #5149
* As @vito suggested, I modified `db.build.SaveOutput()` to call `resource.SetResourceConfigScope()`. So that put-only resource get its `resource_config_scope_id` field populated.

## Notes to reviewer:

Based on my tests, problem 1 has been resolved by recent @vito 's refactor. Now every `put` embeds a `check` and a `get` for image fetching, so `put` will always fetch up-to-date image. I have tested: 1) update image without bumping image tag; 2) update image with bumping image tag, in both cases, `put` fetched correct images.

I made a code change in this PR to solve problem 2. Now put-only resource's version history can be showed up:

put-only-resource-show-version-list

I haven't added any tests yet. Just want @vito or someone to take a look first from the design perspective.

## Release Note

An optimization which should lower the resource checking load on some instances: instead of checking all resources, only resources which are actually used as inputs will be checked. This feature was released in 6.0.0 and reverted in 6.6.0 because of its side effects. Now after resolving those side effects, it's back.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/build.go
+++ b/atc/db/build.go
@@ -1216,15 +1216,15 @@ func (b *build) SaveOutput(
 	versionJSON := string(versionBytes)
 
 	if newVersion {
-		err = theResource.(*resource).setResourceConfigScopeInTransaction(tx, resourceConfigScope)
-		if err != nil {
-			return err
-		}
-
 		err = incrementCheckOrder(tx, resourceConfigScope.ID(), versionJSON)
 		if err != nil {
 			return err
 		}
+	}
+
+	err = theResource.(*resource).setResourceConfigScopeInTransaction(tx, resourceConfigScope)
+	if err != nil {
+		return err
 	}
 
 	_, err = psql.Insert("build_resource_config_version_outputs").
--- a/atc/db/build.go
+++ b/atc/db/build.go
@@ -1244,6 +1244,16 @@ func (b *build) SaveOutput(
 		return err
 	}
 
+	// resource.SetResourceConfigScope has to be called after the transaction,
+	// because if resourceConfigScope is newly created by findOrCreateResourceConfigScope,
+	// before it's committed, SetResourceConfigScope will violate foreign key constraint.
+	if newVersion {
+		err = resource.SetResourceConfigScope(resourceConfigScope)
+		if err != nil {
+			return err
+		}
+	}
+
 	return nil
 }
 
--- a/atc/db/build.go
+++ b/atc/db/build.go
@@ -1172,7 +1172,7 @@ func (b *build) SaveOutput(
 		return ErrBuildHasNoPipeline
 	}
 
-	resource, found, err := pipeline.Resource(resourceName)
+	theResource, found, err := pipeline.Resource(resourceName)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


atc/db/build.go
```diff
@@ -1244,6 +1244,16 @@ func (b *build) SaveOutput(
 		return err
 	}
 
+	// resource.SetResourceConfigScope has to be called after the transaction,
+	// because if resourceConfigScope is newly created by findOrCreateResourceConfigScope,
+	// before it's committed, SetResourceConfigScope will violate foreign key constraint.
+	if newVersion {
+		err = resource.SetResourceConfigScope(resourceConfigScope)
```
vito, line 1251: Can we just do this within the same transaction instead? Otherwise there's potentially a race condition wherein GC removes the scope because it's not (yet) referenced by a resource, which would lead to this failing.

I think this method call really just does one pretty simple query so it should be OK to just inline it so it can use `tx`.

vito (changes requested):
thanks for the PR! just have one suggestion, and then with tests this should be good to go.

atc/db/build.go
evanchaoli: Ah, I didn't read code of `resource.SetResourceConfigScope` when I made the initial change. Now I realize that, it always embeds a transaction, so that in my previous test, there were two different transactions.

evanchaoli: Fixed.

atc/db/build.go
```diff
@@ -1216,6 +1216,11 @@ func (b *build) SaveOutput(
 	versionJSON := string(versionBytes)
 
 	if newVersion {
+		err = theResource.(*resource).setResourceConfigScopeInTransaction(tx, resourceConfigScope)
```
vito, line 1219: I think in practice what you have here will work, but I think it'd be a slight improvement to make this unconditional, and to run it after `incrementCheckOrder`.

Technically the order of operations shouldn't make a difference, since it all runs in a transaction, but in `CheckStep` we only point the resource to the scope after the versions are saved in the proper order, and it matters in that scenario, so it would be nice for this to be consistent just so it doesn't raise questions for someone reading the code.

As for making it conditional, I just don't think it needs to be, and it'd be simpler to think of `put` and `check` the same way (determine config, save versions if necessary, point resource to config).

vito (changes requested):
Just nitpicks at this point, though a test should still be added.

Thanks!

atc/db/build.go
evanchaoli: Sure, fixed.

vito (approved):
lgtm!

PR #6321, opened 2020-11-26
`fly rename-pipeline` can rename an entire instance group

aoldershaw:

Previously, there was no way to rename an instance group - calling `fly rename-pipeline -o foo -n bar` would only rename the pipeline `foo` with no instance vars. Now, it renames all pipelines with the name `foo` into `bar`, keeping the instance vars as they were.

* Refactor some team-scoped API endpoints
* Replace `db.Pipeline.Rename(newName)` with `db.Team.RenamePipeline(oldName, newName)` - this new method renames all pipelines with a given `oldName`

## Notes to reviewer:

This came up in https://github.com/concourse/concourse/pull/6105#issuecomment-733762041

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/team_test.go
+++ b/atc/db/team_test.go
@@ -3289,6 +3289,62 @@ var _ = Describe("Team", func() {
 		})
 	})
 
+	Describe("RenamePipeline", func() {
+		It("renames individual pipelines", func() {
+			found, err := defaultTeam.RenamePipeline("default-pipeline", "new-pipeline")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(found).To(BeTrue())
+
+			_, err = defaultPipeline.Reload()
+			Expect(err).ToNot(HaveOccurred())
+			Expect(defaultPipeline.Name()).To(Equal("new-pipeline"))
+		})
+
+		Context("when multiple pipeline instances have the same name", func() {
+			var (
+				p1 db.Pipeline
+				p2 db.Pipeline
+			)
+
+			BeforeEach(func() {
+				var err error
+				p1, _, err = defaultTeam.SavePipeline(atc.PipelineRef{
+					Name:         "release",
+					InstanceVars: atc.InstanceVars{"version": "6.7.x"},
+				}, defaultPipelineConfig, db.ConfigVersion(0), false)
+				Expect(err).ToNot(HaveOccurred())
+
+				p2, _, err = defaultTeam.SavePipeline(atc.PipelineRef{
+					Name:         "release",
+					InstanceVars: atc.InstanceVars{"version": "7.0.x"},
+				}, defaultPipelineConfig, db.ConfigVersion(0), false)
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			It("renames every instance", func() {
+				found, err := defaultTeam.RenamePipeline("release", "new-pipeline")
+				Expect(err).ToNot(HaveOccurred())
+				Expect(found).To(BeTrue())
+
+				_, err = p1.Reload()
+				Expect(err).ToNot(HaveOccurred())
+				Expect(p1.Name()).To(Equal("new-pipeline"))
+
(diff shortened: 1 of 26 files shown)


atc/db/team_test.go
```diff
@@ -3289,6 +3289,62 @@ var _ = Describe("Team", func() {
...
+			found, err := defaultTeam.RenamePipeline("default-pipeline", "new-pipeline")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(found).To(BeTrue())
+
+			_, err = defaultPipeline.Reload()
+			Expect(err).ToNot(HaveOccurred())
+			Expect(defaultPipeline.Name()).To(Equal("new-pipeline"))
+		})
+
+		Context("when multiple pipeline instances have the same name", func() {
```
chenbh, line 3303: Can you add a test for renaming a set of pipelines that have the same name, but one of them doesn't have instance vars?

fly/commands/rename_pipeline.go
```diff
@@ -6,23 +6,20 @@ import (
...
+	if strings.Contains(command.OldName, "/") {
+		return errors.New("old pipeline name cannot contain '/'")
 	}
-
 	if strings.Contains(command.NewName, "/") {
-		return errors.New("pipeline name cannot contain '/'")
+		return errors.New("new pipeline name cannot contain '/'")
 	}
 	return nil
 }
```
chenbh, line 25: What is the point of checking for just `/` instead of using `atc.ValidateIdentifier`? The atc side does verify that the new name is a valid identifier, if we want to be extra safe we can check the old name here as well.

chenbh (commented):
Finally got around to reviewing this, just some minor comments and and rebasing onto master.

fly/commands/rename_pipeline.go
aoldershaw: The point of checking for `/` is to make it to make it clear to the user that they can't rename an individual instance. e.g. if they tried to do

```
$ fly rename-pipeline -o old/branch:master -n new/branch:master
```

we'd catch them and indicate that it expects a pipeline name.

Your idea of using `atc.ValidateIdentifier` is interesting, but I'm not sure we want to do that yet. Right now, you're allowed to rename a pipeline to an invalid identifier - you get a warning in the API response, but it doesn't stop you.

```
$ fly -t dev rp -o pipeline -n _illegal-name

DEPRECATION WARNING:
  - pipeline: '_illegal-name' is not a valid identifier: must start with a lowercase letter

identifier schema documentation: https://concourse-ci.org/config-basics.html#schema.identifier

pipeline successfully renamed to '_illegal-name'
```

Eventually, when we impose full restrictions, it could be better UX to just reject the identifier client-side, but for now, `fly` probably shouldn't be more restrictive than the API

That previous statement is a bit contradictory to this `/` validation, but at least we've had that validation for a while in `fly`

chenbh: In that case we should update the help text to be more informative
```
      -o, --old-name=        Pipeline to rename. For instance pipelines, this will be the name of the entire group
      -n, --new-name=        New name for the pipeline or instance group
```

aoldershaw: Good call, changed it to:

```
      -o, --old-name=        Existing pipeline or instance group to rename
      -n, --new-name=        New name for the pipeline or instance group
```

Worth noting that instanced pipelines are hidden away in `fly` currently since the feature is still experimental - we don't even show the `--instance-var` flag in the help text. But it's probably still a good idea to document this here since the behaviour isn't obvious

chenbh (approved):
lgtm

Issue #6246, opened 2020-11-05
Build failures when CI is deployed with containerd


muntac:
## Summary

Recently, the CI environment used by core Concourse team was redeployed with `containerd` enabled as were the workers used for testing PRs.

Several jobs failed with errors that went away when we switched back to Guardian.

1. [Postgres errors](https://ci.concourse-ci.org/teams/contributor/pipelines/prs/jobs/unit/builds/1179) when running unit tests

```
[e][initdb] initdb: removing contents of data directory "/tmp/concourse-pg-runner/postgres139775784"

createdb: could not connect to database template1: could not connect to server: No such file or directory

	Is the server running locally and accepting

	connections on Unix domain socket "/tmp/.s.PGSQL.5435"?


[e][postgres] 2020-11-03 22:09:57.856 UTC [39428] LOG:  database system was not properly shut down; automatic recovery in progress

[e][postgres] 2020-11-03 22:09:57.859 UTC [39428] LOG:  redo starts at 0/164E518

[e][postgres] 2020-11-03 22:09:57.872 UTC [39445] FATAL:  the database system is in recovery mode

psql: FATAL:  the database system is in recovery mode
```

```
/tmp/build/0df9eea0/concourse/atc/postgresrunner/postgresrunner.go:297

```

2. [OOM Score Error](https://ci.concourse-ci.org/teams/contributor/pipelines/prs/jobs/containerd-integration/builds/1098) In `containerd-integration`

```
starting task: new task: io.containerd.runc.v2: failed to adjust OOM score for shim: set shim OOM score: write /proc/4105/oom_score_adj: invalid argument
```

3. [Docker build failure](https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/dev-image/builds/603)

```
#5 2.727 container_linux.go:367: starting container process caused: process_linux.go:326: applying cgroup configuration for process caused: mkdir /sys/fs/cgroup/cpuset/buildkit: read-only file system
```

## Triaging info

* Concourse version: 6.7.1 (for PR workers), custom for CI (checks as builds on)
* Browser (if applicable):
* Did this used to work?

aoldershaw:
For 1., it's worth noting that the logs are riddled with `no space left on device`, e.g.

```
[o][initdb] running bootstrap script ... [e][initdb] 2020-11-03 22:09:49.289 UTC [39198] FATAL:  could not write to file "pg_wal/xlogtemp.39198": No space left on device
```

[...]

PR #6330, opened 2020-12-01
start containerd with low oom_score

muntac:

Bug Fix

part of #6246

It is [recommended](https://github.com/containerd/containerd/blob/master/docs/ops.md) that containerd be started with an `oom_score` of -999. We want it to be at the level of other system daemons. This is so that containerd never runs into an out of memory state before the containers it's managing are cleaned up. At the same time it should not be unkillable.

## Notes to reviewer:

To test change the following in the worker section `docker-compose.yml`:
- Change `CONCOURSE_GARDEN_DNS_PROXY_ENABLE` to `CONCOURSE_CONTAINERD_DNS_PROXY_ENABLE`
- Set `CONCOURSE_RUNTIME: "containerd"`
```
docker-compose up --build

#get onto the worker
docker exec -it concourse_worker_1 bash

ps auxww
#pick the PID of containerd daemon from the list
cat /proc/[CONTAINERD_PID]/oom_adj_score
```

The result should be `-999`. 
## Release Note

* It is [recommended](https://github.com/containerd/containerd/blob/master/docs/ops.md) that containerd be started with an `oom_score` of -999. We want it to be at the level of other system daemons. This is so that containerd never runs into an out of memory state before the containers it's managing are cleaned up. At the same time it should not be unkillable.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/worker/runtime/integration/integration_test.go
+++ b/worker/runtime/integration/integration_test.go
@@ -3,6 +3,7 @@ package integration_test
 import (
 	"bytes"
 	"fmt"
+	"github.com/concourse/concourse/worker/workercmd"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -38,10 +39,15 @@ func (s *IntegrationSuite) containerdSocket() string {
 }
 
 func (s *IntegrationSuite) startContainerd() {
+	configPath := filepath.Join(s.tmpDir, "containerd.toml")
+	err := workercmd.WriteDefaultContainerdConfig(configPath)
+	s.NoError(err)
+
 	command := exec.Command("containerd",
 		"--address="+s.containerdSocket(),
 		"--root="+filepath.Join(s.tmpDir, "root"),
 		"--state="+filepath.Join(s.tmpDir, "state"),
+		"--config="+configPath,
 	)
 
 	command.Stdout = &s.stdout
@@ -50,7 +56,7 @@ func (s *IntegrationSuite) startContainerd() {
 		Pdeathsig: syscall.SIGKILL,
 	}
 
-	err := command.Start()
+	err = command.Start()
 	s.NoError(err)
 
 	s.containerdProcess = command
--- a/worker/workercmd/containerd.go
+++ b/worker/workercmd/containerd.go
@@ -20,9 +20,9 @@ import (
 	"github.com/tedsuo/ifrit/grouper"
 )
 
-// writeDefaultContainerdConfig writes a default containerd configuration file
+// WriteDefaultContainerdConfig writes a default containerd configuration file
 // to a destination.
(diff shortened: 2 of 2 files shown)


aoldershaw:
I'm trying to figure out whether -999 is too extreme - it looks like the Docker default is `-500`:

https://github.com/moby/moby/pull/24516

> -500 is a good middle ground because it is less than the default for
most processes and still not -1000 which basically means never kill this
process in an OOM condition on the host machine. The only processes on
my machine that have a score less than -500 are dbus at -900 and sshd
and xfce( my window manager ) at -1000. I don't think docker should be
set lower, by default, than dbus or sshd so that is why I chose -500.

Though it looks like Kubernetes sets Docker's OOM score to -999?

https://github.com/kubernetes/kubernetes/issues/86420

I guess the difference with Docker is that Docker is often run on personal computers, whereas Concourse workers are typically dedicated, which would justify giving containerd a lower OOM score.

WDYT?

muntac:
@aoldershaw I had similar questions and came to similar conclusions. I think `containerd` ops docs themselves recommending -999 felt to me that would be the best way to go about it.

worker/runtime/integration/integration_test.go
```diff
@@ -38,10 +38,16 @@ func (s *IntegrationSuite) containerdSocket() string {
 }
 
 func (s *IntegrationSuite) startContainerd() {
+	configPath := filepath.Join(s.tmpDir, "containerd.toml")
+	config := `oom_score = -999`
```
aoldershaw, line 42: Thoughts on exposing `writeDefaultContainerdConfig` and using that here so that we're running the integration tests with the same (default) config that Concourse runs with? I guess we're running testflight against the default config, but can't hurt to keep the integration tests in sync too

muntac: @aoldershaw I like the idea of having the same default config. It seems safest to disable the same plugins we're disabling in production.

Trying to think of a neat way of exposing `writeDefaultContainerdConfig` as everything in that file is tied up with `WorkerCmd`, and we wouldn't want to put it in there. Do you think it makes sense to introduce a new `ContainerdRunner` interface in the `containerd.go` file, and expose everything through that? Would need a little bit of refactoring

The other option is to duplicate `writeDefaultContainerdConfig` in the test. The con being it would go out of sync when the default config gets updated.

muntac:
@aoldershaw made the requested changes.

aoldershaw (approved):
LGTM

PR #6339, opened 2020-12-03
atc(feat): add new container placement strategies: limit-max-containers and limit-max-volumes.

evanchaoli:

To minimum volume streaming, we are using the container placement strategy "volume-locality". But sometime, this strategy makes busy worker busier, the following "worker containers" chart is from our prod cluster:

worker-containers-spike

From the chart, we can see a spike. At that time, what we saw is, many build failed with error, "insufficient workers", spiked workers stalled, etc. In the meantime, there were a lot of other workers not busy, but they didn't share workload with those spiked workers.

Though garden support max containers restriction, the control is on worker, it's too late. Ideally, ATC should avoid to dispatch too many containers to a worker to make the worker overloaded.

With #6208, we will be able to chain multiple strategies. This PR is adding two new strategies `limit-max-containers` and `limit-max-volumes` that can filter out overloaded workers. So we may configure container placement strategy as `[limit-max-containers, limit-max-volumes, volume-locality, fewest-build-containers]`, which should be better balance workloads among workers.

## Notes to reviewer:

Calling for an initial review now. Tests have not been added yet.

## Release Note

Added two new container placement strategy: `limit-max-containers` and `limit-max-volumes`. That helps avoid workers to be overloaded. A strategy chain can be `[limit-max-containers, limit-max-volumes, volume-locality, fewest-build-containers]`, that may help better balance workloads among workers.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/worker/placement.go
+++ b/atc/worker/placement.go
@@ -18,6 +18,14 @@ type ContainerPlacementStrategyOptions struct {
 	MaxActiveVolumesPerWorker    int      `long:"max-active-volumes-per-worker" default:"0" description:"Maximum allowed number of active volumes per worker. Has effect only when used with limit-active-volumes placement strategy. 0 means no limit."`
 }
 
+type NoWorkerFitContainerPlacementStrategyError struct {
+	Strategy string
+}
+
+func (err NoWorkerFitContainerPlacementStrategyError) Error() string {
+	return fmt.Sprintf("no worker fit container placement strategy: %s", err.Strategy)
+}
+
 type ContainerPlacementStrategy interface {
 	//TODO: Don't pass around container metadata since it's not guaranteed to be deterministic.
 	// Change this after check containers stop being reused
@@ -28,6 +36,7 @@ type ContainerPlacementStrategy interface {
 type ContainerPlacementStrategyChainNode interface {
 	Choose(lager.Logger, []Worker, ContainerSpec) ([]Worker, error)
 	ModifiesActiveTasks() bool
+	StrategyName() string
 }
 
 type containerPlacementStrategy struct {
@@ -42,24 +51,24 @@ func NewContainerPlacementStrategy(opts ContainerPlacementStrategyOptions) (*con
 		case "random":
 			// Add nothing. Because an empty strategy chain equals to random strategy.
 		case "fewest-build-containers":
-			cps.nodes = append(cps.nodes, newFewestBuildContainersPlacementStrategy())
+			cps.nodes = append(cps.nodes, newFewestBuildContainersPlacementStrategy(strategy))
 		case "limit-active-tasks":
 			if opts.MaxActiveTasksPerWorker < 0 {
 				return nil, errors.New("max-active-tasks-per-worker must be greater or equal than 0")
 			}
-			cps.nodes = append(cps.nodes, newLimitActiveTasksPlacementStrategy(opts.MaxActiveTasksPerWorker))
+			cps.nodes = append(cps.nodes, newLimitActiveTasksPlacementStrategy(strategy, opts.MaxActiveTasksPerWorker))
 		case "limit-active-containers":
 			if opts.MaxActiveContainersPerWorker < 0 {
 				return nil, errors.New("max-active-containers-per-worker must be greater or equal than 0")
 			}
-			cps.nodes = append(cps.nodes, newLimitActiveContainersPlacementStrategy(opts.MaxActiveContainersPerWorker))
+			cps.nodes = append(cps.nodes, newLimitActiveContainersPlacementStrategy(strategy, opts.MaxActiveContainersPerWorker))
 		case "limit-active-volumes":
 			if opts.MaxActiveVolumesPerWorker < 0 {
 				return nil, errors.New("max-active-volumes-per-worker must be greater or equal than 0")
 			}
-			cps.nodes = append(cps.nodes, newLimitActiveVolumesPlacementStrategy(opts.MaxActiveVolumesPerWorker))
+			cps.nodes = append(cps.nodes, newLimitActiveVolumesPlacementStrategy(strategy, opts.MaxActiveVolumesPerWorker))
 		case "volume-locality":
-			cps.nodes = append(cps.nodes, newVolumeLocalityPlacementStrategyNode())
+			cps.nodes = append(cps.nodes, newVolumeLocalityPlacementStrategyNode(strategy))
 		default:
 			return nil, fmt.Errorf("invalid container placement strategy %s", strategy)
 		}
@@ -75,7 +84,7 @@ func (strategy *containerPlacementStrategy) Choose(logger lager.Logger, workers 
 			return nil, err
 		}
 		if len(workers) == 0 {
-			return nil, nil
+			return nil, NoWorkerFitContainerPlacementStrategyError{Strategy: node.StrategyName()}
 		}
 	}
 	if len(workers) == 1 {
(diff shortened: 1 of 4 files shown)


taylorsilva (commented):
This works pretty well so far! 👍 

I was curious about what error state we'd end up if no workers are chosen by the strategies. The build ends up error'ing (orange) because `nil` is returned instead of a `worker.Worker`. We end up with this stack trace from the web:

```
running build plan 7: runtime error: invalid memory address or nil pointer dereference","job":"simple-job","pipeline":"test","session":"21.472.2","team":"main"}}
web_1     | panic in running build plan 7: runtime error: invalid memory address or nil pointer dereference
web_1     |  goroutine 15978 [running]:
web_1     | runtime/debug.Stack(0x0, 0x0, 0x0)
web_1     | 	/usr/local/go/src/runtime/debug/stack.go:24 +0xa5
web_1     | github.com/concourse/concourse/atc/util.DumpPanic(0x3e3a680, 0x5663cd0, 0x420220f, 0x15, 0xc001ce68b8, 0x1, 0x1, 0x0, 0x0)
web_1     | 	/src/atc/util/panic_dump.go:19 +0x205
web_1     | github.com/concourse/concourse/atc/engine.(*engineBuild).Run.func2.1(0xc001b33e60, 0x45913a0, 0xc001b33f20, 0xc00155c340)
web_1     | 	/src/atc/engine/engine.go:218 +0x15c
web_1     | panic(0x3e3a680, 0x5663cd0)
web_1     | 	/usr/local/go/src/runtime/panic.go:975 +0x4c6
web_1     | github.com/concourse/concourse/atc/worker.(*client).RunGetStep(0xc000397ea0, 0x457b4c0, 0xc0014b3c20, 0x45913a0, 0xc00096d500, 0x45378c0, 0xc00148d160, 0x1, 0xc0000f03a8, 0x4, ...)
web_1     | 	/src/atc/worker/client.go:440 +0x323
web_1     | github.com/concourse/concourse/atc/exec.(*GetStep).run(0xc000d97a00, 0x457b4c0, 0xc0014b3c20, 0x459bb80, 0xc000901590, 0x4599300, 0xc001b83a90, 0xc0014b3c00, 0x0, 0x0)
web_1     | 	/src/atc/exec/get_step.go:208 +0x191d
web_1     | github.com/concourse/concourse/atc/exec.(*GetStep).Run(0xc000d97a00, 0x457b4c0, 0xc0014b3c20, 0x459bb80, 0xc000901590, 0x0, 0x0, 0x0)
web_1     | 	/src/atc/exec/get_step.go:107 +0x2fb
web_1     | github.com/concourse/concourse/atc/exec.LogErrorStep.Run(0x4513c40, 0xc000d97a00, 0x4519620, 0xc001186200, 0x457b4c0, 0xc0014b3c20, 0x459bb80, 0xc000901590, 0x0, 0x0, ...)
web_1     | 	/src/atc/exec/log_error_step.go:30 +0xda
web_1     | github.com/concourse/concourse/atc/exec.(*runState).Run(0xc000901590, 0x457b4c0, 0xc0014b3c20, 0xc001a5bc40, 0x12, 0x0, 0x0, 0x0, 0xc001a55e00, 0x0, ...)
web_1     | 	/src/atc/exec/run_state.go:96 +0x102
web_1     | github.com/concourse/concourse/atc/engine.(*buildStepDelegate).FetchImage(0xc00077b0a0, 0x457b4c0, 0xc0014b3c20, 0x0, 0x0, 0xc0000f03a8, 0x4, 0xc000901170, 0x0, 0x0, ...)
web_1     | 	/src/atc/engine/build_step_delegate.go:302 +0x802
web_1     | github.com/concourse/concourse/atc/exec.(*TaskStep).imageSpec(0xc000d98000, 0x457b4c0, 0xc0014b3c20, 0x459bb80, 0xc0014b3bc0, 0x4599480, 0xc000921320, 0xc0000f03e0, 0x5, 0x0, ...)
web_1     | 	/src/atc/exec/task_step.go:298 +0x46f
web_1     | github.com/concourse/concourse/atc/exec.(*TaskStep).run(0xc000d98000, 0x457b4c0, 0xc0014b3c20, 0x459bb80, 0xc0014b3bc0, 0x4599480, 0xc000921320, 0xc0014b3c00, 0x0, 0x0)
web_1     | 	/src/atc/exec/task_step.go:220 +0xdd8
web_1     | github.com/concourse/concourse/atc/exec.(*TaskStep).Run(0xc000d98000, 0x457b4c0, 0xc0014b3c20, 0x459bb80, 0xc0014b3bc0, 0xc001700600, 0x0, 0x0)
web_1     | 	/src/atc/exec/task_step.go:137 +0x275
web_1     | github.com/concourse/concourse/atc/exec.LogErrorStep.Run(0x4513d20, 0xc000d98000, 0x4519620, 0xc001c7eb00, 0x457b4c0, 0xc0014b3c20, 0x459bb80, 0xc0014b3bc0, 0x0, 0x0, ...)
web_1     | 	/src/atc/exec/log_error_step.go:30 +0xda
web_1     | github.com/concourse/concourse/atc/exec.OnSuccessStep.Run(0x45197e0, 0xc000e248c0, 0x4519780, 0x57fe6c0, 0x457b4c0, 0xc0014b3c20, 0x459bb80, 0xc0014b3bc0, 0x0, 0x0, ...)
web_1     | 	/src/atc/exec/on_success.go:29 +0x8c
web_1     | github.com/concourse/concourse/atc/exec.(*runState).Run(0xc0014b3bc0, 0x457b4c0, 0xc0014b3c20, 0xc001a69c28, 0x8, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
web_1     | 	/src/atc/exec/run_state.go:96 +0x102
web_1     | github.com/concourse/concourse/atc/engine.(*engineBuild).Run.func2(0xc00052ed80, 0xc001b33e60, 0x45913a0, 0xc001b33f20, 0xc00155c340, 0x459bb80, 0xc0014b3bc0, 0x457b400, 0xc001b50740, 0xc001a69e30)
web_1     | 	/src/atc/engine/engine.go:225 +0x1f5
web_1     | created by github.com/concourse/concourse/atc/engine.(*engineBuild).Run
web_1     | 	/src/atc/engine/engine.go:215 +0xfb9
```

This was also a problem with the limit active tasks strategy. They resolved this issue with this PR: https://github.com/concourse/concourse/pull/4148
I'm wondering if we should better handle this case since it's something that three strategies will run into now...

I was also able to push each worker past its limit. I had two workers with container limits set to 2. I was able to get both to 3 and 4 containers before I got the above error. I'm guessing this is due to the frequency that workers report how many containers they have, not 100% sure on that though, just my initial guess.

evanchaoli:
> I was curious about what error state we'd end up if no workers are chosen by the strategies. The build ends up error'ing (orange) because `nil` is returned instead of a `worker.Worker`. We end up with this stack trace from the web:

This is a good catch! I thought GUI should show something like "no worker". After reading related code, I think when `containerPlacementStrategy.Choose()` method should return an error if no worker pass strategy chain, so that https://github.com/concourse/concourse/blob/master/atc/worker/pool.go#L190 will forward the error, and it should not panic.

evanchaoli:
> I was also able to push each worker past its limit. I had two workers with container limits set to 2. I was able to get both to 3 and 4 containers before I got the above error. I'm guessing this is due to the frequency that workers report how many containers they have, not 100% sure on that though, just my initial guess.

That is true. I don't think this strategy can accurately restrict no worker having active containers to exceed the limit, as reporting of container count is not real time, and it may exist race condition (when multiple strategy checks come, limit not hit yet, so that all checks pass, then limit is exceeded). I think main purpose of this new strategies is to prevent busy workers from getting busier. In practice, say a worker can host 500 containers, then we should configure the limit to like 450 or 480, leaving some room to tolerant race condition cases.

I also realize, there is the other problem users may question. As workers may have different hardware configurations, some worker may host 100 container, some may host 1000 containers, but this limit is global. In that case, these new strategies are not usable. For for clusters like ours, all workers having same or similar configurations, this feature should benefit a lot for balancing workloads among workers.

taylorsilva:
> After reading related code, I think when containerPlacementStrategy.Choose() method should return an error if no worker pass strategy chain, so that https://github.com/concourse/concourse/blob/master/atc/worker/pool.go#L190 will forward the error, and it should not panic.

That should do it. Would be nice if it waited for a worker to be available instead of erroring out though. I can imagine a user being annoyed that they have to manually trigger a build because no worker was selected. I'm fine with doing the simplest thing first though.

> As workers may have different hardware configurations, some worker may host 100 container, some may host 1000 containers, but this limit is global.

All our strategies follow this paradigm so I don't think users expect anything different. It would be awesome to have each worker advertise its own limits though! Would be a much bigger change though :)

taylorsilva (approved):
This LGTM! Did a final acceptance, got the error when the workers hit max containers.

PR #6362, opened 2020-12-11
Allow configuring the login and query timeouts for Vault

### LINKED ISSUE #6347: Vault query hits "missing client token"
## Summary

Looking at the code https://github.com/concourse/concourse/blob/v6.7.2/atc/creds/vault/vault_factory.go#L34, `VaultFactory` waits 5 seconds for Vault login, but it doesn't check login result. Even if login timeout, it still proceeds.

We have seen many cases where Vault login took more than 5 seconds, the following is a real sample:

```
{"timestamp":"2020-12-07T07:27:07.106537957Z","level":"info","source":"atc","message":"atc.checker.running","data":{"check":1546682643,"cluster":"CI","session":"19”}}

{"timestamp":"2020-12-07T07:27:28.075060846Z","level":"info","source":"atc","message":"atc.checker.errored","data":{"check":1546682643,"cluster":"CI","error":"resource config creds evaluation: var lookup 'vmc-vault:vmc-dp/gitlab/api-token': Error making API request.\n\nURL: GET <...omit...>\nCode: 400. Errors:\n\n* missing client token","session":"19”}}

{"timestamp":"2020-12-07T07:27:29.437566900Z","level":"info","source":"atc","message":"atc.checker.login.succeeded","data":{"check":1546682643,"cluster":"CI","lease-duration":2764800,"policies”: <...omit...>
```

Also, looking of the code of cred pool https://github.com/concourse/concourse/blob/v6.7.2/atc/creds/pool.go#L106-L109, it hold a lock, but `NewSecret` may block (by `VaultFactory`), when there are many `var_sources`, this may cause builds to hang.

## Steps to reproduce

1. Use a slow Vault
2. In a pipeline, use a `var_source` to fetch a secret from the vault
3. Then you may hit the error

## Expected results

Should never hit "missing client token" error.

## Actual results

Intermittently sees the error.

## Triaging info

* Concourse version: all versions since `var_sources` was released
* Browser (if applicable): NA
* Did this used to work? No

evanchaoli:

closes #6347 .

1. Add some new Vault configuration parameter for timeout values.
2. For `VarSourcePool`, add a helper class `inPoolSecret` to wait for Vault to login, so that `varSourcePool.FindOrCreate` won't be blocked.

## Notes to reviewer:

The first commit is for your initial review from the design perspective. Code is not cleaned up yet, and tests not added yet.

EDIT: Now it's ready for review. Code cleaned up, tests added.

## Release Note

* These timeouts can be configured using `CONCOURSE_VAULT_LOGIN_TIMEOUT` and `CONCOURSE_VAULT_QUERY_TIMEOUT` respectively
* The new default login timeout is `60s`

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


atc/creds/dummy/manager_factory.go
```diff
@@ -49,5 +52,11 @@ func (factory *managerFactory) NewInstance(config interface{}) (creds.Manager, e
 		})
 	}
 
+	// delay should only be used in unit tests.
+	if delay, ok := configMap["delay"]; ok {
```
aoldershaw, line 56: It feels pretty hacky to have to pass a `delay` into the config for unit tests. Thinking about it more, I don't totally understand the reasoning for the modifications to `inPoolManager` - it's to avoid blocking long-blocking calls to `NewSecrets`, but seems to add a fair bit of complexity to support the `Secrets` not yet being ready. Looking at the credential managers, though, Vault is the only one that can block on `NewSecrets`.

So, what if we kept `inPoolManager` as it was, and just moved the "wait for login + timeout" code to `Vault.Get` instead? Is there a reason that won't work?

atc/creds/vault/manager.go
```diff
@@ -14,16 +14,17 @@ import (
...
-	Namespace       string `mapstructure:"namespace" long:"namespace"   description:"Vault namespace to use for authentication and secret lookup."`
+	PathPrefix      string        `mapstructure:"path_prefix" long:"path-prefix" default:"/concourse" description:"Path under which to namespace credential lookup."`
+	LookupTemplates []string      `mapstructure:"lookup_templates" long:"lookup-templates" default:"/{{.Team}}/{{.Pipeline}}/{{.Secret}}" default:"/{{.Team}}/{{.Secret}}" description:"Path templates for credential lookup"`
+	SharedPath      string        `mapstructure:"shared_path" long:"shared-path" description:"Path under which to lookup shared credentials."`
+	Namespace       string        `mapstructure:"namespace" long:"namespace"   description:"Vault namespace to use for authentication and secret lookup."`
+	LoginTimout     time.Duration `mapstructure:"login_timeout" long:"login-timeout" default:"60s" description:"Timeout value for Vault login."`
+	QueryTimout     time.Duration `mapstructure:"query_timeout" long:"query-timeout" default:"60s" description:"Timeout value for Vault query."`
```
aoldershaw, line 25: Typo in field names (`Timout` should be `Timeout`)

aoldershaw (changes requested):
Hey @evanchaoli, really sorry about the slow review! I was on vacation for much of December, so haven't had a chance to look at this until now

atc/creds/dummy/manager_factory.go
evanchaoli: @aoldershaw Thanks for the comment. Handing "login timeout" in Vault is a different direction from my current implementation. Please allow me to explain my idea and logic.

`inPoolManager` was only introduced by me for `var_sources`. When I implemented `var_sources`, I followed a principle that it should not touch logic of any individual credential manager.

For the system level credential manager, because "concourse web" takes time between startup and scheduling a build to run, thus login time in most of cases is not a problem. Only Vault credential manager add a sleep in `NewSecrets`, but that couldn't prove other credential managers to not requiring a login time. I only use Vault, I never tested other credential managers.

If kept `inPoolManager` as it was, it would imply that `NewSecrets` should return immediately without having any blocking operations. I'm afraid it would be hard to make all contributors to clearly aware of that, and thus it could be error prone, especially considering more credential managers may be added in future.

With my implementation of this PR, the complexity is contained in `inPoolManager`, thus it's testable and controllable. Changes in `inPoolManager` can be easily tested by unit tests, but changes in credential managers, say Vault, requires end-to-end tests with a real Vault, thus hard to be fully covered by unit tests. Thus I decided to avoid touching credential manager related code.

`dummy` was added @vito for testing `var_sources`, so I feel ok to add a new configuration option `delay`. Maybe I can delete the comment in line 55 that could be confused.

Putting all together, I feel enhancing `inPoolManager` is relative simpler and safer.

aoldershaw: I hear you, but to me, the fact that `NewSecrets` doesn't return an `error` is a pretty good indication that it shouldn't have blocking operations, as most blocking operations are error-prone - the fact that all other credential manager implementations follow this pattern is an indication of that. The blocking in `VaultFactory.NewSecrets` is super hacky as a result of not being able to return an `error` - if the login (which takes place elsewhere) "times out" (i.e. takes too long from when `NewSecrets` was called, rather than from when the login was actually initiated), we still return a Vault instance - it just won't work properly.

So IMO the Vault implementation should really just be made non-blocking. I don't think this should be that hard. The simplest implementation could just involve threading the `loggedIn` chan to the `Vault` struct, and in `Vault.findSecret`, waiting for the `loggedIn` chan to receive (with a timeout). Potentially a better design would be to pass in a modified `SecretReader` that handles the `ReAuther`'s behaviour and wraps the regular Vault client.

But either way, I think a slight change to the `Vault` implementation would be preferable.

evanchaoli: Okey.

atc/creds/vault/manager.go
```diff
@@ -95,6 +96,8 @@ func (manager *VaultManager) Config(config map[string]interface{}) error {
 	manager.PathPrefix = "/concourse"
 	manager.Auth.RetryMax = 5 * time.Minute
 	manager.Auth.RetryInitial = time.Second
+	manager.LoginTimeout = 60 * time.Second
+	manager.QueryTimeout = 60 * time.Second
```
aoldershaw, line 100: I was originally thinking that 60 seconds was a bit conservative, but looks like that's Vault's default timeout, so 👍 :

https://github.com/hashicorp/vault/blob/acecc949078f907e1daabe44712e37964b397687/api/client.go#L151

atc/creds/vault/vault.go
```diff
@@ -43,6 +51,15 @@ func (v Vault) NewSecretLookupPaths(teamName string, pipelineName string, allowR
 
 // Get retrieves the value and expiration of an individual secret
 func (v Vault) Get(secretPath string) (interface{}, *time.Time, bool, error) {
+	if v.LoggedIn != nil {
+		select {
+		case <-v.LoggedIn:
+			v.LoggedIn = nil
```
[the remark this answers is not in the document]
aoldershaw: Perhaps we should move this `select` anyway - if we fail to login to vault, there's no point blocking for 60 seconds every time we call `Get`

aoldershaw (changes requested):

atc/creds/vault/vault.go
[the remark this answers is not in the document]
evanchaoli: @aoldershaw  Good point. If Vault login fails, `ReAuther.authLoop()` will keep retrying login to Vault. 

If the error is something like wrong password, then login will never succeed. In this case, I think there is no much difference between `Get` returning immediately with a "missing client token" error and returning an timeout error after blocking 60 seconds.

If Vault login failure is due to some temp problem, like flaky network, or Vault restart, or etc., then blocking might reduce failures. 

WDYT?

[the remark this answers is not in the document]
evanchaoli: I don't have strong opinion of setting `v.LoggedIn = nil`. I can delete it.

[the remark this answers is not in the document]
aoldershaw: True, good point - I'm cool with keeping the `select` here then, but if you could delete `v.LoggedIn = nil`, that'd be great.

---

I tested out passing invalid credentials to Vault both on `master` and on this PR. On `master`, we get a vault specific error:

```
failed to interpolate task config: Error making API request.

URL: GET https://vault:8200/v1/concourse/main/bar/foo
Code: 400. Errors:

* missing client token
```

and on this branch, after 60 seconds, we get:

```
failed to interpolate task config: timed out to login to vault
```

At first I was thinking the less descriptive error on this PR is an issue, but I realize now that it's actually *better*, since the `missing client token` error is misleading - when we can't login, we just use an unauthenticated client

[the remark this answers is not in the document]
evanchaoli: Deleted `v.LoggedIn = nil`.

evanchaoli:
@aoldershaw I backport this PR to 6.7.x with #6413 because we keep seeing "missing client token" errors on our prod cluster, so I want to solve this problem in 6.7.3. Please also help review #6413.

aoldershaw (approved):

PR #6369, opened 2020-12-17
feat(atc): expose username of who manually triggered build to build metadata.

### LINKED ISSUE #4714: Ability to detect user who initiated build
## Context

I am part of a team that maintains a suite of software used for deployment and CI/CD internally within my company.  We maintain a set of hosted services that run the actual deployments, as well as a client application used to interact with those services.  This client application can be run locally or run within the context of a Concourse pipeline for automation of these deployments.  We are currently in the middle of a push to increase the visibility and auditability of deployments made using our tooling.

With the current stateless design of Concourse and the intentional limiting of information, I lack two very distinct capabilities that are limiting my team's ability to effectively use Concourse:

1) We do not have the ability to link a particular build to the user who triggered it (or if it was automatically triggered), preventing us from auditing builds or back-tracing issues that arise from an erroneous build.
2) We lack the ability to detect if we are running within Concourse, meaning we cannot enable selective behavior (such as only permitting certain actions within the context of a Concourse pipeline and not permitting those actions for a local run of the binary, or vice versa) without maintaining two separate editions of the binary in question (one for Concourse, one for local runs), a significant increase in project maintenance.

Of the two, the first one is the more critical for our use case.  Our Concourse server is integrated with our corporate SSO system, but the user information is not exposed to the pipelines, so our build process cannot detect who initiated a particular build (or which pipeline initiated it, or if it was automatically triggered).  This seriously limits our ability to audit deployments and builds, and do any sort of forensics if we encounter issues with a build down the line (especially since the pipeline-specific and build-specific information appears to be completely deleted if the pipeline itself is deleted).  We can collect this information if the deployment is triggered using a local run of the client binary, but not if it is triggered via Concourse.

Exposing the user information to the build process, perhaps through environment variables or some sort of local metadata service we could `curl` to.  Implementing a solution to 1) will most likely effectively solve 2) at the same time.

Unlikely I'd have the bandwidth to do so.

### LINKED ISSUE #6345: Add user info to the step metadata
Request to add user name to the step metadata list.
This would help determine and log which user has executed any particular job run and re-runs.

This info can also be used in the pipeline to notify the user who has triggered a job.

evanchaoli:

Feature

fixes #4714
closes #6345

Add a column `who_triggered` to `builds` table to store username of who manually triggered the build, then expose the username to build metadata via env.

## Notes to reviewer:

* Please feel free to suggest any renaming stuff.
* I found that, rerun build's `manually_triggered` field is not set to true, is that intended or a bug? I consider a bug, so I fixed it in this PR.

`fly builds` screenshot:

fly-builds-with-created-by

## Release Note

Expose username of who manually triggered build to build metadata.

* `fly builds` is added a new column `created by` that shows a user ID if a build is triggered manually.
* A new build metadata `BUILD_CREATED_BY` may be exposed to resource. But it is not exposed by default, you need to turn it by add `expose_build_created_by` when defining a resource:

```yaml
resources:
- name: some-resource
  type: some-type
  expose_build_created_by: true
  source:
    ...
```

As different authentication connectors populate different claims, a new `concourse web` CLI option `--concourse-display-user-id-per-connector` is added that allow cluster administrator to configure which claims field should be consider as unique user id. Values of this option should be in format `<connector>:<fieldname>`. Where `connector` should be an authentication connector the cluster, valid values are `ldap`, `github`, `cf`, `bitbucket-cloud`, `gitlab`, `microsoft`, `oauth`, `oidc` and `saml`. And `fieldname` should be one of the following 4 values:

* `user_id` mapping to claims' user id field
* `name` mapping to claims' username field
* `username` mapping to claims' preferred username field
* `email` mapping to claims' email field

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


atc/db/dbtest/builder.go
```diff
@@ -291,7 +291,7 @@ func (builder Builder) WithPendingJobBuild(assign *db.Build, jobName string) Set
 			return fmt.Errorf("job '%s' not configured in pipeline", jobName)
 		}
 
-		build, err := job.CreateBuild()
+		build, err := job.CreateBuild("some-user")
```
vito, line 294: It would be nice to convert more of the tests to use this new `dbtest` package so there doesn't have to be similar churn across `db/*_test.go` in the future, but def not a blocker for the PR (that's a lot of work!).

atc/api/jobserver/create_build.go
```diff
@@ -34,7 +36,8 @@ func (s *Server) CreateJobBuild(pipeline db.Pipeline) http.Handler {
 			return
 		}
 
-		build, err := job.CreateBuild()
+		acc := accessor.GetAccessor(r)
+		build, err := job.CreateBuild(acc.Claims().UserID)
```
vito, line 40: Just checking: is `UserID` the correct value to use here for most connectors? I wonder how this value compares between the various connectors we have. I think I would expect e.g. my GitHub login, `vito`, to be stored when I trigger a build, but I don't know if that's `UserID`, `Username`, or `PreferredUsername`.

@xtremerui Does using `UserID` seem sensible to you?

vito (changes requested):
Thanks for working on this! Can we mark this as 'fixes #4714' too?

> I found that, rerun build's manually_triggered field is not set to true, is that intended or a bug? I consider a bug, so I fixed it in this PR.

This is intended, though the name is admittedly not very descriptive. `manually_triggered` means the user triggered the job to create the build, and those builds behave differently from reruns. "Triggered" builds do not have their input versions determined at creation time and require a new `check` of every input resource before they can be computed. Rerun builds know their input versions immediately and do not require a `check`.

atc/api/jobserver/create_build.go
xtremerui: `UserID` could be a string that is not human readable so UserName or PreferredUserName should be used here. For example,

https://github.com/concourse/concourse/blob/d1dd60f7555f6e7f7d22b0e55289c24534c34b45/atc/api/accessor/accessor.go#L225-L231

evanchaoli: I tested with local user and LDAP:

* With local user, preferred_username is empty, `userid` and `username` are identical, in my test, both were `test`.
* With LDAP, preferred_username is also empty, `userid` is `chaol`, and `username` is `Chao (Evan) Li`. In LDAP case, `userid` is expected.

From https://github.com/concourse/concourse/blob/master/atc/api/accessor/accessor.go#L107-L118, both `userid` and `username` might be used. So I think the logic can be, if `userid` is not empty, then use `userid`, otherwise use `username`. @vito @xtremerui WDYT?

vito: Trying multiple makes sense to me - I'll leave it to @xtremerui to decide whether to include `preferred_username` as I don't know which connectors use it. :+1:

Devil's advocate: is there a standard convention for this kind of thing? Would it make sense to pass a structured value, since different resource types might need different fields? Should we just pass all the claims along? Is that crazy? (I don't hold any opinions here, I see pros and cons, but thought I'd bring it up in case anyone has a perspective to share.)

evanchaoli: Given we are not sure which id fits to all connectors and it's hard to test all connectors, passing entire claims sounds a good idea. In the first step, we can store entire claims in db, so that even if we only pass `userid` this time, it would be very easy to passing other fields without further db changes in future.

atc/db/dbtest/builder.go
evanchaoli: Thanks for the info. However I'd consider that is out of  scope of this PR.

atc/exec/step_metadata.go
```diff
@@ -62,5 +64,15 @@ func (metadata StepMetadata) Env() []string {
 		env = append(env, "ATC_EXTERNAL_URL="+metadata.ExternalURL)
 	}
 
+	if metadata.CreatedBy != nil {
+		if metadata.CreatedBy.UserID != "" {
```
vito, line 68: I don't know that we can be sure of this order of precedence - `UserID` with the GitHub connector returns the user's numeric ID which isn't super useful - seems like the login or email would be more useful there.

What do you think of just JSON encoding the claims and sending them along? Seems like UserID might be useful in some contexts and UserName useful in others, so maybe it should be up to the resource type to choose.

vito (commented):
Looking good but I'm afraid this has opened a can of worms around user info across connectors, so this may take a couple more iterations to figure out. 🤔

evanchaoli:
@vito I very hesitate to expose whole claims given there are a lot of connectors we don't know. But if we only expose a single field, e.g. for LDAP connector, expose userid; for GitHub connector, expose username, which should have no security concern. Then the problem is how to decide which field of claims to expose.

If we don't know, how about let cluster administrators to decide? Also I like your idea of opt-in. How about to add a global options like:

```
CONCOURSE_USERID_FIELD_OF_CLAIMS=ldap:userid,github:username,<some-connector>:preferredusername
```

If a cluster administrator decide to not expose trigger user, then just don't configure this option.

atc/db/dbtest/builder.go
```diff
@@ -276,7 +276,7 @@ func (builder Builder) WithResourceTypeVersions(resourceTypeName string, version
 	}
 }
 
-func (builder Builder) WithPendingJobBuild(assign *db.Build, jobName string) SetupFunc {
```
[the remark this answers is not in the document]
evanchaoli: Reverted the change.

atc/api/accessor/accessor.go
```diff
@@ -18,16 +18,7 @@ type Access interface {
...
-}
-
-type Claims struct {
-	Sub               string
-	UserID            string
-	UserName          string
-	PreferredUsername string
-	Email             string
-	Connector         string
+	Claims() atc.Claims
```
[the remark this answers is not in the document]
evanchaoli: Changed to use `atc.UserInfo`.

I never used `fly userinfo` before, but when I tried this command today, I noticed that, when I used an local account, "username" column showed nothing, which is because "local" connector doesn't populate `preferredUsername`. I have fixed this bug by adding `UserInfo.PreferredUserId()` function.

evanchaoli:
@vito I have addressed most of review comments. The only issues left are:

* how to opt-in the feature: I don't want to add a parameter at resource level. If we only expose user id, I don't think it need to be opt-in.
* expose a json string or a single user id. I would prefer expose a single user id, but I don't have strong opinion. I'm ok to expose a json string. If we expose a single user id, we will need to ask for help from other contributors to update

atc/exec/step_metadata.go
```diff
@@ -62,5 +64,15 @@ func (metadata StepMetadata) Env() []string {
 		env = append(env, "ATC_EXTERNAL_URL="+metadata.ExternalURL)
 	}
 
+	if metadata.CreatedBy != nil {
```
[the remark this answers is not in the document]
chenbh: Another solution could be the ability to mark a job (or step, but probably more appropriate for job) as `auditable: true`. This would make it so that the userinfo is passed along to every step, along with outputting a log entry with the userinfo every time a build is created.

Heck, maybe in the future we can expand on the audit ideas by instrumenting in OPA requests when certain builds are created.

vito:
@evanchaoli Cluster-wide configuration feels brittle to me. If everything is left up to the operator, resource type authors won't know what value they're getting (e.g. username or user ID). Also if resource type A is coded against user IDs, but resource type B is coded against usernames there would be no way to have both work.

Cluster-wide opt-in also doesn't feel great; sending potentially sensitive information to every resource without the user knowing feels wasteful. It seems like this is primarily for notifications, so configuring it on the notification resource feels more portable and safe. We could even expand on it to figure out what information to send:

```yaml
resources:
- name: some-notifier
  type: slack
  source: # ...
  user_info: # extract user's email for associating to Slack account to @mention
    github: email
    ldap: email

- name: github-status
  type: github-status
  source: # ...
  user_info: # extract user's GitHub username to @mention
    github: preferred_username
```

...this way we can even detect if the requested information isn't available (e.g. because the connector is not configured) and let the user know that their pipeline configuration needs to be updated for the cluster's connector setup.

Granted, configuring per-connector might be too complicated; I guess one alternative would be `user_info: email`, but that wouldn't allow for cases where you want a different field for different connectors. (But that's kind of deep in the weeds.)

Can we expand on the original use case for exposing this to resources? How do you see resource types using this information?

evanchaoli:
> Can we expand on the original use case for exposing this to resources? How do you see resource types using this information? Having a realistic scenario to talk about would probably be better than my own obtuse made-up use cases. 😅

The original request is #6345 that is from one of our VMC users. I asked him to file an GitHub issue directly.

The other request in our internal Jira:
"""
It would be very helpful to  have username logged in build log when pipeline gets re-triggered, in day-to-day operations in bigger teams. Today, it is hard for us to know who re-triggered the pipelines and need to keep checking via slack or mails.
"""

vito:
@evanchaoli That seems like a nice balance! I agree that it'd be great to see who triggered the build in the UI, and letting the operator control what to field to use makes sense to me (e.g. what if they use GitHub auth but want to use email addresses everywhere).

One UX concern - it would seem intuitive to me that whatever value we record for `created_by` and use for notifications is the same as the value shown in the upper right corner when you're logged in:

image

By tying the two together I think it becomes clear that it's meant to just be a human-readable descriptor, so that eliminates my concern about resource types needing numeric user IDs vs. user names - it should always be a username, real name, email address, etc. So +1 to your last snippet (`expose_build_created_by: true`).

If you agree, I think we should consider making that change in this PR as well to ensure they're consistent. In summary we'd be adding a operator-configurable mapping from connector to field to display/record, with sane defaults for each connector, and we'd respect it for `created_by`, the web UI, and `fly userinfo`.

I think this would involve a small API change to add a field to the `/api/v1/user` response; it currently exposes all the claims and it's up to the client to determine what to show, but with this change we'd be pushing that decision server-side.

https://github.com/concourse/concourse/blob/ed0493360b2a8da365f1a11d07595a8ac7eecad0/fly/commands/userinfo.go#L59

https://github.com/concourse/concourse/blob/37c4aebfb90fcbd1f7b7504846a9860f783b3a2d/web/elm/src/Login/Login.elm#L96-L102

Does this make sense? Thanks!

evanchaoli:
@vito In the latest two commits, I have

* added a CLI option to configure what field of claims to display
* added `expose_build_created_by` to resource
* added a new field `DisplayUserId` to `UserInfo`. I named this field as user id, because user id implies "unique". But I'm open to accept any renaming suggestion
* updated `fly userinfo` command to shows `DisplayUserId`
* updated `fly builds` command, so it can shows a new column `created-by`

I hesitate on one thing and want to hear your suggestion. Do we still need to store user info JSON in table `builds`  for column `created_by`? Or should be just store a string of `DisplayUserId`? Storing entire claims feels massive, occupying more db spaces, and the info is not used by other features. Only pros might be good for future using. I slightly keen on storing only string. WDYT?

Before reaching to a conclusion for storing claims or string, I haven't yet updated tests. I want to you take a quick look first from design perspective. Thanks.

vito:
> I hesitate on one thing and want to hear your suggestion. Do we still need to store user info JSON in table `builds` for column `created_by`? Or should be just store a string of `DisplayUserId`? Storing entire claims feels massive, occupying more db spaces, and the info is not used by other features. Only pros might be good for future using. I slightly keen on storing only string. WDYT?

I think we can just store the string. 👍🏻

evanchaoli:
@vito I have changed to store a string for "created_by", also fixed existing tests. Please go ahead review the code.

I will add tests for the new code later.

evanchaoli:
@vito I have added tests, so the PR is now ready for review and merge.

atc/builds/planner.go
```diff
@@ -95,13 +95,14 @@ func (visitor *planVisitor) VisitGet(step *atc.GetStep) error {
...
-		Tags:     step.Tags,
-		Timeout:  step.Timeout,
+		Type:                 resource.Type,
+		Resource:             resourceName,
+		Source:               resource.Source,
+		Params:               step.Params,
+		Version:              &version,
+		Tags:                 step.Tags,
+		Timeout:              step.Timeout,
+		ExposeBuildCreatedBy: resource.ExposeBuildCreatedBy,
```
vito, line 105: Can we avoid adding it to the `atc.GetPlan`? It seems like it should only be needed for `put`. 🤔 Passing it to `get` feels strange because it could technically end up caching stuff based on whoever triggered the build that generated the cache. I don't think it'll be useful to any resources anyway.

atc/builds/planner.go
```diff
@@ -125,11 +126,12 @@ func (visitor *planVisitor) VisitPut(step *atc.PutStep) error {
...
-		Resource: resourceName,
-		Type:     resource.Type,
-		Source:   resource.Source,
-		Params:   step.Params,
+		Name:                 logicalName,
+		Resource:             resourceName,
+		Type:                 resource.Type,
+		Source:               resource.Source,
+		Params:               step.Params,
+		ExposeBuildCreatedBy: resource.ExposeBuildCreatedBy,
```
vito, line 134: Ideally this can also be removed once we have a `atc.NotifyPlan` (or something). Makes sense for now though.

skymarshal/skycmd/flags.go
```diff
@@ -242,3 +242,68 @@ func (con *Connector) newTeamConfig() (TeamConfig, error) {
...
+		}
+	}
+
+	// For unconfigured connector, applies a default rule.
+	if username != "" {
+		return username
+	} else if userid != "" {
+		return userid
+	} else {
+		return preferredUsername
```
vito, line 307: I think we want preferredUsername to take precedence over the rest. With GitHub for example, preferred username is e.g. `vito` and username is e.g. `Alex Suraci`.

docker-compose.yml
```diff
@@ -32,6 +32,7 @@ services:
       CONCOURSE_CLUSTER_NAME: dev
       CONCOURSE_ENABLE_PIPELINE_INSTANCES: "true"
       CONCOURSE_ENABLE_ACROSS_STEP: "true"
+      CONCOURSE_DISPLAY_USER_ID_PER_CONNECTOR: "local:name,ldap:user_id,github:name"
```
vito, line 35: Does this need to be committed? Would rather leave the `docker-compose.yml` tidy if not just to make sure we're not straying too far from end users' configuration.

fly/commands/builds.go
```diff
@@ -192,6 +192,7 @@ func (command *BuildsCommand) displayBuilds(builds []atc.Build) error {
 			{Contents: "end", Color: color.New(color.Bold)},
 			{Contents: "duration", Color: color.New(color.Bold)},
 			{Contents: "team", Color: color.New(color.Bold)},
+			{Contents: "created-by", Color: color.New(color.Bold)},
```
vito, line 195: We use spaces in column titles elsewhere:

```suggestion
			{Contents: "created by", Color: color.New(color.Bold)},
```

skymarshal/skycmd/flags.go
```diff
@@ -242,3 +242,68 @@ func (con *Connector) newTeamConfig() (TeamConfig, error) {
...
+	mapConnectorUserid map[string]string
+}
+
+func NewSkyDisplayUserIdGenerator(config []string) (atc.DisplayUserIdGenerator, error) {
+	generator := &skyDisplayUserIdGenerator{
+		mapConnectorUserid: map[string]string{},
+	}
+
+	for _, str := range config {
+		parts := strings.Split(str, ":")
```
vito, line 256: This can be avoided by using a `map[string]string` instead of `[]string` for the flag struct field; our flag and env parser already handles `foo:bar,baz:buzz` syntax for parsing into maps.

vito (changes requested):
Looking good! Left some feedback, mostly nitpicks.

atc/atccmd/command.go
```diff
@@ -255,6 +255,8 @@ type RunCommand struct {
 	BaseResourceTypeDefaults flag.File `long:"base-resource-type-defaults" description:"Base resource type defaults"`
 
 	P2pVolumeStreamingTimeout time.Duration `long:"p2p-volume-streaming-timeout" description:"Timeout value of p2p volume streaming" default:"15m"`
+
+	DisplayUserIdPerConnector []string `long:"display-user-id-per-connector" description:"Define how to display username for each authentication connector. Format is <connector-id>:<filename>. Valid field names are user_id, name, username and email."`
```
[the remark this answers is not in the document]
evanchaoli: Yes, I followed https://github.com/concourse/concourse/blob/master/atc/api/usersserver/get.go#L19-L28, name maps to claims username, and username maps to claims preferred username.

atc/builds/planner.go
evanchaoli: @vito  Remember there are resource types to fetch build metadata, I wrote one for our internal usages, there are other community ones. For such resource type, `put` generates `build_id: xxx` as version, then implied `get` saves build metadata into files, so `get` needs it also.

I don't think cache is a problem. Because if a resource type doesn't store build metadata into files, then there is no cache problem there. If a resource type, like build metadata resource, BUILD_ID, BUILD_NAME may have same cache problem, so that it must use `put` to avoid cache problem.

I you have strong opinion to remove it from `get`, I'm fine. But in that way, build metadata resource's `put` will need to use include "create-by" in version, like `{build_id: 1234, created_by: someone}`, because version is the only way `put` can pass data to implied `get`.

WDYT?

atc/builds/planner.go
evanchaoli: Good to know.

skymarshal/skycmd/flags.go
evanchaoli: OK, changed the order to preferredUsername -> userId -> username, that fits at least GitHub and LDAP. With LDAP, preferred username is empty.

fly/commands/builds.go
evanchaoli: Actually I tried "created by", but "created-by" looked better to me. Anyway, I'm ok.

skymarshal/skycmd/flags.go
evanchaoli: Good to know. Fixed.

docker-compose.yml
evanchaoli: I intentionally added it, wanted to commit it. This configuration shows how to configure this option, and it fits to real end user's configuration, though github should change to use "username". Is it to commit this?

evanchaoli: Deleted.

evanchaoli:
@vito Expect the comment about if exposing created_by to `get`, I have addressed all other review comments.

atc/builds/planner.go
vito: That's a good point, but as far as the original request goes it sounded like this was only necessary for notifications, which would use the information directly in the `put` step. It seems risky to extend this to support patterns beyond that as they might become deprecated when proper support for notifications is introduced.

evanchaoli: I'm fine as that won't block our requirements. I'll remove the exposing from `get`.

atc/builds/planner.go
```diff
@@ -95,13 +95,13 @@ func (visitor *planVisitor) VisitGet(step *atc.GetStep) error {
...
 		Name: step.Name,
 
-		Type:     resource.Type,
-		Resource: resourceName,
-		Source:   resource.Source,
-		Params:   step.Params,
-		Version:  &version,
-		Tags:     step.Tags,
-		Timeout:  step.Timeout,
+		Type:                 resource.Type,
```
vito, line 98: Would you mind running `go fmt` to get rid of these? Just aiming to preserve `git blame` integrity since these don't need to change anymore.

atc/db/migration/migrations/1609959123_add_created_by_to_builds.up.sql
```diff
@@ -0,0 +1,4 @@
+BEGIN;
+ALTER TABLE builds
+    ADD COLUMN created_by text not null default ''::text;
```
vito, line 3: Should we just let this be `NULL`? I think setting a default will make the migration slow for deployments with a large number of builds, and I wouldn't expect any value to be present for builds which were created by Concourse.

skymarshal/skycmd/flags.go
```diff
@@ -242,3 +242,61 @@ func (con *Connector) newTeamConfig() (TeamConfig, error) {
...
+					valid = true
+					break
+				}
+			}
+		}
+		if !valid {
+			return nil, fmt.Errorf("invalid connector: %s", connectorId)
+		}
+
+		if fieldName != "user_id" && fieldName != "name" && fieldName != "username" && fieldName != "email" {
```
vito, line 271: nit: This might be easier to skim as a `switch`:

```go
switch fieldName {
case "user_id", "name", "username", "email":
default:
    return nil, fmt.Errorf(...)
}
```

vito (changes requested):
Just one concern about the migration (+ the `go fmt` request) and then I think we're good to go. Thanks!

skymarshal/skycmd/flags.go
evanchaoli: Thanks for the suggestion. Fixed.

atc/builds/planner.go
evanchaoli: @vito  I usually run `go fmt` before commit, sorry for missing this file.

I use IntelliJ IDEA. The IDEA has a feature of "format project" that can `go fmt` all files of the project, which is very convenient. But unfortunately I cannot use the feature, I tried, "format project" would cause many other files formating that I didn't touch, meaning that, there are many files in the project not `go fmt`ed. For this reason, I have to format changed files one by one, then potentially, missing some files.

Would you mind me to submit a PR to correct all missing formatted files? And afterwards, we may consider to add a format check in CI pipeline. The logic should be simple:

```
for each files in commit
     out_file = `go fmt original_file`
     if out_file != original_file
          fail
     end
end
```

which will make sure all changes well `go fmt`ed.

atc/db/migration/migrations/1609959123_add_created_by_to_builds.up.sql
evanchaoli: @vito  Actually I thought over that when I changed the field from `bson` to `text`. IMO, migration is an one-time effort, even if it's a little bit slow, it's acceptable. Even if a `builds` table contains millions of tuples, migration should be done within a few minutes (assuming that a large deployment must have a powerful db behind). In our case, we usually claim a maintenance window for each upgrade,  though Concourse may not really down. so I don't feel migration is a big problem.

On the other side, making this field `not null`, we will not need to scan this field into a `NullableString` first, which helps keep code clean, and is faster at runtime.

vito: I'm not super concerned about runtime performance of using `sql.NullString`; we already use it everywhere, and avoiding it for performance reasons seems like a micro-optimization. As for clean code, I don't think it's really that much worse; if you like we can keep things `string` everywhere instead of `*string` and handle the `NULL` value only in the one place where we read the value from the database.

I'm more concerned with having an precise representation to make troubleshooting easier; `""` vs. `NULL` can sometimes be a useful distinction. A `NULL` value means there was clearly no creator, while `""` could mean the operator-configured field was blank for the user in question (maybe they didn't configure their real name or email address).

evanchaoli: OK, I have made `created_by` nullable. And make API and `fly builds` to distinct "not set" and "empty:

* In API, if created by is not set, the field `created_by` absents. If the field is set with `""`, then it just takes `"created_by": ""`.
* In `fly builds`, if created by is not set, it shows "system".

fly-builds-w-system

vito: Neat! That feels pretty clear. Thanks!

vito (approved):
lgtm - thanks!

anijayn:
Hi @evanchaoli. Just curious on why exposing the `$BUILD_CREATED_BY` environment variable was removed for the get step in this [commit](https://github.com/concourse/concourse/pull/6369/commits/84160e823e7958e09fba1e6885d0a1c46272253c). Could you please share the reason if possible because I cannot find it anywhere ?

evanchaoli:
@anijayn @vito requested that, otherwise this PR couldn't get merged.

anijayn:
Thanks for replying. Can you please tell if there were any issues around adding that if you remember? We have a user who is requesting it but I'm not sure how to extract it from the `put` step.

karthikraina32:
@vito  is there is any reason why we are not exposing `$BUILD_CREATED_BY` environment variable in get step

anijayn:
@evanchaoli Sorry for tagging. Could you please reply to the query when you get time ?

evanchaoli:
@anijayn Refer to https://concourse-ci.org/implementing-resource-types.html#resource-metadata

`put` script can see the environment variable `BUILD_CREATED_BY`.

evanchaoli:
@karthikraina32 How we solved the problem was to put "created_by" in "version", so that when get step can generate "created_by" from "version".

karthikraina32:
@evanchaoli do you have any example for that ?

karthikraina32:
@evanchaoli sorry for tagging you again, can u please help on this

karthikraina32:
hi @evanchaoli  I have tried doing what you have informed but am still getting empty value in get step, help here is highly appreciated 
image

evanchaoli:
@karthikraina32 Let me show you my code. The code is written in Ruby, but I think it should be easy to understand:

```ruby
# out, put step includes "build_created_by" into "version"

new_version = {
    build_id: ENV['BUILD_ID']
}
if !ENV["BUILD_CREATED_BY"].nil?
    new_version["build_created_by"] = ENV["BUILD_CREATED_BY"]
end 

ret = {
    version: new_version,
    metadata: parse_metadata(source, params, input_dir)
}

puts ret.to_json
```

```ruby
# in: get step extracts "build_created_by" from "version"

# Parse the parameters passed to the script over STDIN
sSource = STDIN.read
parsed = JSON.parse(sSource)

# Get build_created_by from version, which is generated by put step.
version = parsed['version']
ENV["BUILD_CREATED_BY"] = version["build_created_by"]
```

Hope this helps.

anijayn:
Thanks for sharing this @evanchaoli. Could you please share how this resource is being used in a pipeline as well if possible?

karthikraina32:
Thanks @evanchaoli this is the piece of info needed.. And @anijayn I'll let you know that

taylorsilva:
For anyone wondering why this metadata field isn't provided in the `get` step, the answer is in this review discussion on this PR: https://github.com/concourse/concourse/pull/6369#discussion_r565558527

PR #6386, opened 2021-01-02
Disable resource checking for individual resources

taylorsilva:

This PR allows a user to disable automatic checking of a resource by setting `check_every: never` in a resource's definition. We've seen this request come up when people only want checking to happen via a webhook. The workaround has always been setting `check_every` to some large value like `99h`. 

## Notes to reviewer:
Lidar runs every ~10s. I tested using this pipeline:
```yaml
---
resources:
  - name: every-30s
    type: time
    check_every: never
    icon: clock-outline
    source:
      interval: 1s

jobs:
  - name: job
    public: true
    plan:
      - get: every-30s
        trigger: true
      - task: simple-task
        config:
          platform: linux
          image_resource:
            type: registry-image
          run:
            path: echo
            args: ["Hello, world!"]
```

## Release Note
* Automatic resource checking for individual resources can be disabled by setting `check_every: never` in a resource's definition

## Contributor Checklist
- [x] [Signed] all commits
- [x] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/api/config_test.go
+++ b/atc/api/config_test.go
@@ -638,7 +638,7 @@ jobs:
 											Name:         "some-resource",
 											Type:         "some-type",
 											Source:       nil,
-											CheckEvery:   "10s",
+											CheckEvery:   &atc.CheckEvery{Interval: 10 * time.Second},
 											CheckTimeout: "1m",
 										},
 									},
@@ -887,7 +887,7 @@ jobs:
 										Name:       "some-resource",
 										Type:       "some-type",
 										Source:     nil,
-										CheckEvery: "10s",
+										CheckEvery: &atc.CheckEvery{Interval: 10 * time.Second},
 									},
 								},
 									Jobs: atc.JobConfigs{
--- a/atc/api/resources_test.go
+++ b/atc/api/resources_test.go
@@ -798,7 +798,7 @@ var _ = Describe("Resources API", func() {
 				resourceType2.TypeReturns("type-2")
 				resourceType2.SourceReturns(map[string]interface{}{"source-key-2": "source-value-2"})
 				resourceType2.PrivilegedReturns(true)
-				resourceType2.CheckEveryReturns("10ms")
+				resourceType2.CheckEveryReturns(&atc.CheckEvery{Interval: 10 * time.Millisecond})
 				resourceType2.TagsReturns([]string{"tag1", "tag2"})
 				resourceType2.ParamsReturns(map[string]interface{}{"param-key-2": "param-value-2"})
 				resourceType2.VersionReturns(map[string]string{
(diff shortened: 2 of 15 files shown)


clarafu (commented):
Nice! This definitely would make it easier to disable resource checking without it feeling hacky. 

One thing I wanted your opinion on is having the `never` value be fetched at parse time (parsing the config) rather than at runtime (lidar)? So it would be similar to how `VersionConfig` is done where it accepts multiple forms of values (`every`, `latest` or a version). 

https://github.com/concourse/concourse/blob/d1dd60f7555f6e7f7d22b0e55289c24534c34b45/atc/steps.go#L644-L701

This might help for future usages of the `CheckEvery` value, where it will be very clear that there is a possible case of the `CheckEvery` holding a `never` string rather than always being a time duration. This will require a bit more work because it involves changing the `CheckEvery` type from a string to a struct of type something like
```
type CheckEvery struct {
  Never bool
  Interval time.Duration
}
```

https://github.com/concourse/concourse/blob/c19d076747579e4887cbd577428e79bf4cc02b9f/atc/config.go#L182

and having it be parsed out separate from the config within `scanResource()` in `db.Resource`.

Just wanted to get your thoughts first though!

taylorsilva:
@clarafu that is the EXACT kind of comment I was hoping to get! I did this quickly so I went path of least resistance, but I've been thinking that there's probably a different way of doing it along the lines you described, just didn't have time to dig into it.

I was also thinking of adding some extra validation checks for check_every as well because currently you can pass any string into `check_every` and it'll take it. The problem is this breaks checking and we don't surface any errors in the web UI so we could end up with frustrated and confused users! 😱

clarafu:
@taylorsilva Haha I'm glad to hear we're thinking along the same lines 😆 Hmm sorry I'm not exactly sure how that will break checking? Would the validations not be done during `fly -t <target> set-pipeline` since that is when the `UnmarshalJSON` would be run? I could totally be wrong about that.

taylorsilva:
@clarafu from my manual testing there's no validation. I was messing around with a pipeline and had set `check_every: 1hr`. Checking then broke but no error showed in the UI. I eventually remembered it's `1h` not `1hr` for go's duration strings. See my second commit for where things break: https://github.com/concourse/concourse/pull/6386/commits/3ef7b5208f52e210c70b8b00a0363f8971b09018

clarafu:
Oh sorry I meant that if we added the validation for `checkEvery` using the custom type JSON unmarshalers (like `VersionConfig`) then that would only break during the `fly set-pipeline` command and not during the check step right?

taylorsilva:
@clarafu made the changes! It now works like `VersionConfig`

aoldershaw (approved):
LGTM, not sure if @clarafu has any requests

PR #6410, opened 2021-01-08
Ensure task, set_pipeline, load_var steps have names

### LINKED ISSUE #6405: set_pipeline allows empty name and creates an unusable pipeline
## Summary

The set_pipeline step allows you to not enter a name for the pipeline. Concourse then creates a new pipeline with no name and you can't unpause, view, or delete the pipeline. Interestingly, both fly pipelines and the UI show it, but with no name.

Update: It looks like if you delete the job that uses set_pipeline, the broken pipeline it created disappears too.

## Steps to reproduce

...
plan:
  - set_pipeline:
     file: path/to/your/file.yml

## Expected results

Concourse should refuse to run the step with no name.

## Actual results

Concourse creates an unusable/untouchable pipeline.

## Triaging info

* Concourse version: 6.7.2

taylorsilva:

fixes #6405

These steps should always have names provided. Our docs note that names
are required but our validation doesn't return an error for empty
strings on these steps.

Put and Get steps are safe from this because the validator checks that
their resources exist.

## Release Note

* Return an error when no identifier is provided for task, set_pipeline, and load_var steps

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/api/config_test.go
+++ b/atc/api/config_test.go
@@ -407,7 +407,7 @@ var _ = Describe("Config API", func() {
 						Expect(ioutil.ReadAll(response.Body)).To(MatchJSON(`
 							{
 								"errors": [
-										"identifier cannot be an empty string"
+										"pipeline: identifier cannot be an empty string"
 								]
 							}`))
 					})
--- a/atc/api/config_test.go
+++ b/atc/api/config_test.go
@@ -353,25 +353,25 @@ var _ = Describe("Config API", func() {
 			})
 
 			Context("when an identifier is invalid", func() {
+				Context("and is a string", func() {
+					BeforeEach(func() {
+						var err error
+						request, err = requestGenerator.CreateRequest(atc.SaveConfig, rata.Params{
+							"team_name":     "_team",
+							"pipeline_name": "_pipeline",
+						}, nil)
+						Expect(err).NotTo(HaveOccurred())
 
-				BeforeEach(func() {
-					var err error
-					request, err = requestGenerator.CreateRequest(atc.SaveConfig, rata.Params{
-						"team_name":     "_team",
-						"pipeline_name": "_pipeline",
-					}, nil)
-					Expect(err).NotTo(HaveOccurred())
+						request.Header.Set("Content-Type", "application/json")
 
-					request.Header.Set("Content-Type", "application/json")
+						payload, err := json.Marshal(pipelineConfig)
+						Expect(err).NotTo(HaveOccurred())
 
-					payload, err := json.Marshal(pipelineConfig)
-					Expect(err).NotTo(HaveOccurred())
+						request.Body = gbytes.BufferWithBytes(payload)
+					})
 
-					request.Body = gbytes.BufferWithBytes(payload)
-				})
-
-				It("returns warnings in the response body", func() {
-					Expect(ioutil.ReadAll(response.Body)).To(MatchJSON(`
+					It("returns warnings in the response body", func() {
+						Expect(ioutil.ReadAll(response.Body)).To(MatchJSON(`
 							{
 								"warnings": [
 									{
@@ -384,7 +384,35 @@ var _ = Describe("Config API", func() {
 									}
 								]
 							}`))
+					})
 				})
+				Context("and is an empty string", func() {
+					BeforeEach(func() {
+						var err error
+						request, err = requestGenerator.CreateRequest(atc.SaveConfig, rata.Params{
+							"team_name":     "",
+							"pipeline_name": "",
+						}, nil)
+						Expect(err).NotTo(HaveOccurred())
+
+						request.Header.Set("Content-Type", "application/json")
+
+						payload, err := json.Marshal(pipelineConfig)
+						Expect(err).NotTo(HaveOccurred())
+
+						request.Body = gbytes.BufferWithBytes(payload)
+					})
+
+					It("returns warnings in the response body", func() {
+						Expect(ioutil.ReadAll(response.Body)).To(MatchJSON(`
+							{
+								"errors": [
+										"identifier cannot be an empty string"
+								]
+							}`))
+					})
+				})
+
 			})
 
 			Context("when a config version is specified", func() {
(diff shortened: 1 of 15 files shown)


aoldershaw (changes requested):
What do you think about moving this logic into `ValidateIdentifier` instead? The empty string certainly isn't a valid identifier, after all! Rather than a `*ConfigWarning`, it could return `(*ConfigWarning, error)`.

While this PR will fix the `set_pipeline` step, it doesn't fix e.g. `fly rename-pipeline`, which will let you rename a pipeline to `""`, but does not let you use it. `fly rename-team` fails with an internal server error due to some DB constraint not being met. Addressing the issue at `ValidateIdentifier` will at least force each caller to consider the error case, and will make it easy to migrate the soft identifier validation to strict validation when the time comes

taylorsilva:
@aoldershaw I thought about that and did wonder "is an empty string an invalid identifier?". I was hesitant to make the change there because I wasn't sure if this was true everywhere. I was worried about increasing the scope of that function. I guess that makes sense though; anywhere we expect an identifier we also expect it to no be an empty string. I'll move the implementation to that function.

taylorsilva:
@aoldershaw made the changes. Need to add tests for the api layer for the error being returned now still.

atc/api/configserver/save.go
```diff
@@ -69,19 +69,28 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
 	}
 
 	pipelineName := rata.Param(r, "pipeline_name")
-	warning := atc.ValidateIdentifier(pipelineName, "pipeline")
+	warning, err := atc.ValidateIdentifier(pipelineName, "pipeline")
+	if err != nil {
+		session.Info("ignoring-invalid-config", lager.Data{"error": err.Error()})
```
aoldershaw, line 74: Maybe change this log message:

```suggestion
		session.Info("ignoring-invalid-pipeline-name", lager.Data{"error": err.Error()})
```

similarly for the team name

atc/configwarning.go
```diff
@@ -15,7 +16,12 @@ var validIdentifiers = regexp.MustCompile(`^[\p{Ll}\p{Lt}\p{Lm}\p{Lo}][\p{Ll}\p{
 var startsWithLetter = regexp.MustCompile(`^[^\p{Ll}\p{Lt}\p{Lm}\p{Lo}]`)
 var invalidCharacter = regexp.MustCompile(`([^\p{Ll}\p{Lt}\p{Lm}\p{Lo}\d\-_.])`)
 
-func ValidateIdentifier(identifier string, context ...string) *ConfigWarning {
+func ValidateIdentifier(identifier string, context ...string) (*ConfigWarning, error) {
+	var err error
```
aoldershaw, line 20: Maybe get rid of this var - doesn't look like it's needed (and it's kind of misleading to see `return nil, err` at the end, given that the `err` is `nil`

atc/configwarning.go
```diff
@@ -15,7 +16,12 @@ var validIdentifiers = regexp.MustCompile(`^[\p{Ll}\p{Lt}\p{Lm}\p{Lo}][\p{Ll}\p{
...
+	if identifier == "" {
+		return nil, errors.New("identifier cannot be an empty string")
+	}
+
 	if identifier != "" && !validIdentifiers.MatchString(identifier) {
```
aoldershaw, line 25: Nit, but you can get rid of the first condition

```suggestion
	if !validIdentifiers.MatchString(identifier) {
```

atc/configwarning.go
```diff
@@ -15,7 +16,12 @@ var validIdentifiers = regexp.MustCompile(`^[\p{Ll}\p{Lt}\p{Lm}\p{Lo}][\p{Ll}\p{
 var startsWithLetter = regexp.MustCompile(`^[^\p{Ll}\p{Lt}\p{Lm}\p{Lo}]`)
 var invalidCharacter = regexp.MustCompile(`([^\p{Ll}\p{Lt}\p{Lm}\p{Lo}\d\-_.])`)
 
-func ValidateIdentifier(identifier string, context ...string) *ConfigWarning {
+func ValidateIdentifier(identifier string, context ...string) (*ConfigWarning, error) {
+	var err error
+	if identifier == "" {
+		return nil, errors.New("identifier cannot be an empty string")
```
aoldershaw, line 22: Perhaps this should incorporate the `context` as we do with warnings - that way you won't need all the `fmt.Sprintf("%s: %s", identifier, err.Error())` in `atc/configvalidate/validate.go`

```suggestion
		return nil, fmt.Errorf("%s: identifier cannot be an empty string", strings.Join(context, ""))
```

aoldershaw (changes requested):
Some minor changes, but it also [looks like there's a compilation issue](https://ci.concourse-ci.org/builds/67942986#L5ffe96d9:522):

```
fly/commands/set_pipeline.go:43:14: assignment mismatch: 1 variable but atc.ValidateIdentifier returns 2 values
```

atc/api/config_test.go
```diff
@@ -384,7 +384,35 @@ var _ = Describe("Config API", func() {
...
+						Expect(err).NotTo(HaveOccurred())
+
+						request.Body = gbytes.BufferWithBytes(payload)
+					})
+
+					It("returns warnings in the response body", func() {
+						Expect(ioutil.ReadAll(response.Body)).To(MatchJSON(`
+							{
+								"errors": [
+										"identifier cannot be an empty string"
```
aoldershaw, line 410: There seems to be a few failing tests now as a result of adding the context to the error: https://ci.concourse-ci.org/builds/69020323#L60011778:17:83

(just requires changing the asserted error message)

aoldershaw (changes requested):

aoldershaw (approved):

PR #6412, opened 2021-01-11
db: lidar checks put-only resources with failed checks

### LINKED ISSUE #6342: Check put-only resources as long as they are erroring
With upgrading prod to a dev image that had `registry-image` v1.0.0, various resource checks failed because their workers had not been upgraded to the new version and did not understand the `tag` field that is now present in versions.

So these resources all errored, but then they were never checked again. They're all fixed by manually going around and clicking 'check' once the workers are ugpraded, but it'd be nice if I didn't have to do that.

Can we instead skip only successfully-checked put-only resources? I think this can be done with a tweak to the query.

taylorsilva:
Bug Fix | Feature | Documentation | **Enhancement**

closes #6342

Previous behaviour: Lidar checks resoures that are inputs for jobs. This
naturally excluded put-only resoruces.

This commit makes a smal quality-of-life improvement by automatically
checking any put-only resources that have errored, aborted, or failed.
Otherwise, it should return the same result set as before; only resouces
that are inputs to jobs.

If a put-only resources fails as part of a job's build it will NOT get included
in Lidar's checking. It will only get checked by Lidar if it fails a check
build.

## Release Note

* Lidar now checks any put-only resources that ran a check which failed.

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/check_factory.go
+++ b/atc/db/check_factory.go
@@ -141,14 +141,17 @@ func (c *checkFactory) Resources() ([]Resource, error) {
 	rows, err := resourcesQuery.
 		LeftJoin("(select DISTINCT(resource_id) FROM job_inputs) ji ON ji.resource_id = r.id").
 		LeftJoin("(select DISTINCT(resource_id) FROM job_outputs) jo ON jo.resource_id = r.id").
+		Where(sq.And{
+			sq.Eq{"p.paused": false},
+			sq.Eq{"r.active": true},
+		}).
 		Where(sq.Or{
 			sq.And{
-				sq.Eq{"p.paused": false},
-				sq.Eq{"r.active": true},
+				// find all resources that are inputs to jobs
 				sq.NotEq{"ji.resource_id": nil},
 			},
 			sq.And{
-				sq.Eq{"p.paused": false},
+				// find put-only resources that have errored
 				sq.Expr("b.status IN ('aborted','failed','errored')"),
 				sq.Eq{"ji.resource_id": nil},
 			},
--- a/atc/db/check_factory.go
+++ b/atc/db/check_factory.go
@@ -139,8 +139,13 @@ func (c *checkFactory) Resources() ([]Resource, error) {
 	var resources []Resource
 
 	rows, err := resourcesQuery.
-		Join("(select DISTINCT(resource_id) FROM job_inputs) ji ON ji.resource_id = r.id").
+		LeftJoin("(select DISTINCT(resource_id) FROM job_inputs) ji ON ji.resource_id = r.id").
+		LeftJoin("(select DISTINCT(resource_id) FROM job_outputs) jo ON jo.resource_id = r.id").
 		Where(sq.Eq{"p.paused": false}).
+		// filter for resources that are job inputs
+		Where(sq.NotEq{"ji.resource_id": nil}).
+		// also inlude put-only resources that have errored
+		Suffix("OR (b.status IN ('aborted','failed','errored') AND ji.resource_id IS NULL)").
 		RunWith(c.conn).
 		Query()
 
--- a/atc/db/check_factory.go
+++ b/atc/db/check_factory.go
@@ -141,11 +141,18 @@ func (c *checkFactory) Resources() ([]Resource, error) {
 	rows, err := resourcesQuery.
 		LeftJoin("(select DISTINCT(resource_id) FROM job_inputs) ji ON ji.resource_id = r.id").
 		LeftJoin("(select DISTINCT(resource_id) FROM job_outputs) jo ON jo.resource_id = r.id").
-		Where(sq.Eq{"p.paused": false}).
-		// filter for resources that are job inputs
-		Where(sq.NotEq{"ji.resource_id": nil}).
-		// also inlude put-only resources that have errored
-		Suffix("OR (b.status IN ('aborted','failed','errored') AND ji.resource_id IS NULL)").
+		Where(sq.Or{
+			sq.And{
+				sq.Eq{"p.paused": false},
+				sq.Eq{"r.active": true},
+				sq.NotEq{"ji.resource_id": nil},
+			},
+			sq.And{
+				sq.Eq{"p.paused": false},
+				sq.Expr("b.status IN ('aborted','failed','errored')"),
+				sq.Eq{"ji.resource_id": nil},
+			},
+		}).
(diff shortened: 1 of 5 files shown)


atc/db/check_factory.go
```diff
@@ -139,8 +139,13 @@ func (c *checkFactory) Resources() ([]Resource, error) {
...
 
 	rows, err := resourcesQuery.
-		Join("(select DISTINCT(resource_id) FROM job_inputs) ji ON ji.resource_id = r.id").
+		LeftJoin("(select DISTINCT(resource_id) FROM job_inputs) ji ON ji.resource_id = r.id").
+		LeftJoin("(select DISTINCT(resource_id) FROM job_outputs) jo ON jo.resource_id = r.id").
 		Where(sq.Eq{"p.paused": false}).
+		// filter for resources that are job inputs
+		Where(sq.NotEq{"ji.resource_id": nil}).
+		// also inlude put-only resources that have errored
+		Suffix("OR (b.status IN ('aborted','failed','errored') AND ji.resource_id IS NULL)").
```
taylorsilva, line 148: I wanted to avoid doing this but I couldn't figure out a way to either 1) clear out the where and rebuild it or 2) add a where that `OR`'s instead of `AND`'s. Neither seemed possible with the `mastermind/squirrel` library

chenbh: I don't think you can clear out the `WHERE` in the `resourceQuery`, but you should be able to construct something like
```go
Where(sq.Or{
	sq.And{
		sq.Eq{"p.paused": false},
		sq.NotEq{"ji.resource_id": nil},
	},
	sq.And{
		sq.Expr("b.status IN ('aborted','failed','errored')"),
		sq.Eq{"ji.resource_id": nil},
	},
}).
```

taylorsilva: Yes, the issue is it will be `AND`'d with the `WHERE` in the base `resourcesQuery` var. See:

https://github.com/concourse/concourse/blob/df62ba0d8a2d4b1b2d99bff9a25bb00874e8eb6e/atc/db/resource.go#L76-L106

Thinking maybe I'll split it into two vars.

chenbh: Isn't it fine to `AND` that `Where(sq.Eq{"r.active": true})` since we only want to filter on active resources as well? My understanding is that the combined clause would look something like 
`... WHERE r.active = true AND ((p.paused = false AND ji.resource_id IS NOT NULL) OR (b.status IN ('aborted','failed','errored') AND ji.resource_id IS NOT NULL))`

We probably don't want to lidar check inactive resources anyways since an inactive resource indicates that it was "orphaned" as result of an existing pipeline config getting updated
https://github.com/concourse/concourse/blob/df62ba0d8a2d4b1b2d99bff9a25bb00874e8eb6e/atc/db/team.go#L1465-L1471
https://github.com/concourse/concourse/blob/df62ba0d8a2d4b1b2d99bff9a25bb00874e8eb6e/atc/db/team.go#L505

atc/db/check_factory.go
```diff
@@ -139,8 +139,20 @@ func (c *checkFactory) Resources() ([]Resource, error) {
...
 
 	rows, err := resourcesQuery.
-		Join("(select DISTINCT(resource_id) FROM job_inputs) ji ON ji.resource_id = r.id").
-		Where(sq.Eq{"p.paused": false}).
+		LeftJoin("(select DISTINCT(resource_id) FROM job_inputs) ji ON ji.resource_id = r.id").
+		LeftJoin("(select DISTINCT(resource_id) FROM job_outputs) jo ON jo.resource_id = r.id").
+		Where(sq.Or{
+			sq.And{
+				sq.Eq{"p.paused": false},
+				sq.Eq{"r.active": true},
```
vito, line 147: Pretty sure we want `r.active` to be `true` all the time - I believe 'active' means 'still configured', so if it's `false` that means someone removed it from their pipeline config. It seems safer to put that back on `resourcesQuery` since it seems like a safe assumption that we only ever want to deal with still-configured resources. (Unless I'm missing some edge case?)

atc/db/check_factory.go
```diff
@@ -139,8 +139,20 @@ func (c *checkFactory) Resources() ([]Resource, error) {
 	var resources []Resource
 
 	rows, err := resourcesQuery.
-		Join("(select DISTINCT(resource_id) FROM job_inputs) ji ON ji.resource_id = r.id").
-		Where(sq.Eq{"p.paused": false}).
+		LeftJoin("(select DISTINCT(resource_id) FROM job_inputs) ji ON ji.resource_id = r.id").
+		LeftJoin("(select DISTINCT(resource_id) FROM job_outputs) jo ON jo.resource_id = r.id").
+		Where(sq.Or{
+			sq.And{
+				sq.Eq{"p.paused": false},
```
vito, line 146: Might be clearer to extract this into a chained `Where` since it's the same in both branches.

vito (changes requested):

atc/db/check_factory.go
taylorsilva: Can't extract this into a chained where because the squirrel library only `AND`'s where's together, there's no way to `OR` two `where()` calls. That's why it's all stuffed into one `where()` and duplicated :(

taylorsilva: This is also why I refactored the `resourcesQuery` into `activeResourcesQuery`, so I could start with a query with no `where()`.

atc/db/resource.go
```diff
@@ -73,37 +73,41 @@ type Resource interface {
...
+		"b.end_time",
+	).
+		From("resources r").
+		Join("pipelines p ON p.id = r.pipeline_id").
+		Join("teams t ON t.id = p.team_id").
+		LeftJoin("builds b ON b.id = r.build_id").
+		LeftJoin("resource_config_scopes rs ON r.resource_config_scope_id = rs.id").
+		LeftJoin("resource_pins rp ON rp.resource_id = r.id")
+
+	activeResourcesQuery = resourcesQuery.
```
vito, line 108: I don't think this needs to be split out anymore

vito (approved):

PR #6436, opened 2021-01-18
Proposing fix for OIDC groups

### LINKED ISSUE #6435: OIDC authentication stopped working with 6.7.3
## Summary
We use OIDC authentication with Keycloak for Concourse. It used to work up until Concourse 6.7.3. Now, the corresponding dex event in the logs reports no groups ([]), albeit groups are present in the corresponding token.

## Steps to reproduce
Deploy Concourse 6.7.3, connect it to Keycloak (12.0.1), log in with any user that has any groups, see that groups are not there in the logs and that the user has no permissions.

## Expected results

`{"timestamp":"2021-01-16T00:44:27.810496550Z","level":"info","source":"atc","message":"atc.dex.event","data":{"fields":{},"message":"login successful: connector \"oidc\", username=\"wherearemygroups\", email=\"test@test.com\", groups=[\"CCADMINS\"]","session":"15"}}`
(This was generated with the same user with Concourse 6.7.2)

The user has permissions as defined by the corresponding groups.

## Actual results

`{"timestamp":"2021-01-16T00:43:25.392091032Z","level":"info","source":"atc","message":"atc.dex.event","data":{"fields":{},"message":"login successful: connector \"oidc\", username=\"wherearemygroups\", preferred_username=\"wherearemygroups\", email=\"test@test.com\", groups=[]","session":"14"}}`

The user has no permissions in concourse.

## Additional context

The claim for the user looks like that with my go-based OIDC tester:
`{
    "OAuth2Token": {
        "access_token": "*REDACTED*",
        "token_type": "Bearer",
        "refresh_token": "eyJhbGciOiJIUzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6ICJmMDUyZDkxYi04ZGYzLTQ1ZDEtYTBjNi04NjI0YTMwZDUzMzMifQ.eyJleHAiOjE2MTA3NTk3ODQsImlhdCI6MTYxMDc1Nzk4NCwianRpIjoiZTYzMDU0M2YtYzA0MC00ZGFiLThkZGEtOTkyMDQ0ZDQzNTRmIiwiaXNzIjoiaHR0cHM6Ly9hdXRoLmNsb3VkeWZtZS5kZS9hdXRoL3JlYWxtcy90cnVlZGV2b3BzIiwiYXVkIjoiaHR0cHM6Ly9hdXRoLmNsb3VkeWZtZS5kZS9hdXRoL3JlYWxtcy90cnVlZGV2b3BzIiwic3ViIjoiNDlmNjBmYmItY2NiNy00ZGJmLTgxMDctZmNjMDE3MjU3NGQ1IiwidHlwIjoiUmVmcmVzaCIsImF6cCI6InRlc3QiLCJzZXNzaW9uX3N0YXRlIjoiODYzNDFiNTQtMzI4YS00YjYyLTkyNjUtOGU0NzY5YTcwNzJkIiwic2NvcGUiOiJvcGVuaWQgcHJvZmlsZSBncm91cHMgZW1haWwifQ.RWdm5GhYWlOpzgHWkDNtcN8mPHiPHoNmdJ8mQUB6T4s",
        "expiry": "2021-01-16T00:51:24.818199987Z"
    },
    "IDTokenClaims": {
        "exp": 1610758284,
        "iat": 1610757984,
        "auth_time": 1610757630,
        "jti": "5846434b-b1d5-4692-be87-e8950077d36e",
        "iss": "****",
        "aud": "test",
        "sub": "49f60fbb-ccb7-4dbf-8107-fcc0172574d5",
        "typ": "ID",
        "azp": "test",
        "session_state": "86341b54-328a-4b62-9265-8e4769a7072d",
        "at_hash": "LQ4g8NPNCPGG6DYsdh6MqA",
        "acr": "0",
        "email_verified": true,
        "name": "Test Test",
        "groups": [
            "CCADMINS"
        ],
        "preferred_username": "wherearemygroups",
        "given_name": "Test",
        "family_name": "Test",
        "email": "test@test.com",
        "username": "wherearemygroups"
    }
}`

Keycloak itself evaluates the token like this:

`{
  "exp": 1610758417,
  "iat": 1610758117,
  "jti": "4fde8dfb-92f8-4bdb-aa05-3e2c4d191b0d",
  "iss": "****",
  "sub": "49f60fbb-ccb7-4dbf-8107-fcc0172574d5",
  "typ": "Bearer",
  "azp": "edge",
  "session_state": "2afc57a1-5c7d-4abc-8009-187046669348",
  "acr": "1",
  "scope": "openid profile groups email",
  "email_verified": true,
  "name": "Test Test",
  "groups": [
    "CCADMINS"
  ],
  "preferred_username": "wherearemygroups",
  "given_name": "Test",
  "family_name": "Test",
  "email": "test@test.com",
  "username": "wherearemygroups"
}`
## Triaging info

* Concourse version: 6.7.3
* Browser (if applicable): Any
* Did this used to work? yes

konstl000:

Bug Fix 

closes #6435

## Notes to reviewer:
Since Dex has the option InsecureEnableGroups set to false by default and concourse did not overwrite this, the OIDC connector ignores the groups coming from the upstream provider. This seems to be pretty bad for concourse, since its current RBAC model heavily relies on groups and many users (including myself) use those to manage access.
Adding an --oidc-enable-groups flag allows to override this behavior (if necessary) and thus to use groups again.

## Release Note
Added an --oidc-disable-groups flag that by default fetches groups claims from an upstream OIDC provider.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [x] Code reviewed
- [ ] Tests reviewed
- [x] Documentation reviewed
- [x] Release notes reviewed
- [x] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,9 +5,7 @@
   "repository": "https://github.com/concourse/concourse",
   "license": "Apache-2.0",
   "dependencies": {
-    "element-ui": "^2.15.0",
-    "request": "^2.86.0",
-    "vue-resource": "^1.5.1"
+    "request": "^2.86.0"
   },
   "devDependencies": {
     "@babel/core": "^7.8.4",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,9 @@
   "repository": "https://github.com/concourse/concourse",
   "license": "Apache-2.0",
   "dependencies": {
-    "request": "^2.86.0"
+    "element-ui": "^2.15.0",
+    "request": "^2.86.0",
+    "vue-resource": "^1.5.1"
   },
   "devDependencies": {
     "@babel/core": "^7.8.4",
--- a/skymarshal/skycmd/oidc_flags.go
+++ b/skymarshal/skycmd/oidc_flags.go
@@ -28,7 +28,7 @@ type OIDCFlags struct {
 	HostedDomains      []string    `long:"hosted-domains" description:"List of whitelisted domains when using Google, only users from a listed domain will be allowed to log in"`
 	CACerts            []flag.File `long:"ca-cert" description:"CA Certificate"`
 	InsecureSkipVerify bool        `long:"skip-ssl-validation" description:"Skip SSL validation"`
-  InsecureEnableGroups bool      `long:"enable-groups" description:"Enable OIDC groups"`
+  DisableGroups bool      `long:"disable-groups" description:"Enable OIDC groups"`
 }
 
 func (flag *OIDCFlags) Name() string {
@@ -76,7 +76,7 @@ func (flag *OIDCFlags) Serialize(redirectURI string) ([]byte, error) {
 		RootCAs:            caCerts,
 		InsecureSkipVerify: flag.InsecureSkipVerify,
 		RedirectURI:        redirectURI,
-    InsecureEnableGroups: flag.InsecureEnableGroups,
+    InsecureEnableGroups: ! flag.DisableGroups,
 	}
 
 	config.ClaimMapping.GroupsKey = flag.GroupsKey
--- a/skymarshal/skycmd/oidc_flags.go
+++ b/skymarshal/skycmd/oidc_flags.go
@@ -28,7 +28,7 @@ type OIDCFlags struct {
 	HostedDomains      []string    `long:"hosted-domains" description:"List of whitelisted domains when using Google, only users from a listed domain will be allowed to log in"`
 	CACerts            []flag.File `long:"ca-cert" description:"CA Certificate"`
 	InsecureSkipVerify bool        `long:"skip-ssl-validation" description:"Skip SSL validation"`
-  DisableGroups bool      `long:"disable-groups" description:"Enable OIDC groups"`
+  DisableGroups bool      `long:"disable-groups" description:"Disable OIDC groups claims"`
 }
 
 func (flag *OIDCFlags) Name() string {
--- a/skymarshal/skycmd/oidc_flags.go
+++ b/skymarshal/skycmd/oidc_flags.go
@@ -28,6 +28,7 @@ type OIDCFlags struct {
 	HostedDomains      []string    `long:"hosted-domains" description:"List of whitelisted domains when using Google, only users from a listed domain will be allowed to log in"`
 	CACerts            []flag.File `long:"ca-cert" description:"CA Certificate"`
 	InsecureSkipVerify bool        `long:"skip-ssl-validation" description:"Skip SSL validation"`
+  InsecureEnableGroups bool      `long:"enable-groups" description:"Enable OIDC groups"`
 }
 
 func (flag *OIDCFlags) Name() string {
@@ -75,6 +76,7 @@ func (flag *OIDCFlags) Serialize(redirectURI string) ([]byte, error) {
 		RootCAs:            caCerts,
 		InsecureSkipVerify: flag.InsecureSkipVerify,
 		RedirectURI:        redirectURI,
+    InsecureEnableGroups: flag.InsecureEnableGroups,
 	}
 
 	config.ClaimMapping.GroupsKey = flag.GroupsKey


package.json
```diff
@@ -5,7 +5,9 @@
   "repository": "https://github.com/concourse/concourse",
   "license": "Apache-2.0",
   "dependencies": {
-    "request": "^2.86.0"
+    "element-ui": "^2.15.0",
+    "request": "^2.86.0",
+    "vue-resource": "^1.5.1"
```
xtremerui, line 10: is this change required for the PR?

konstl000: No, I don't think so. It came about because I needed the UI to run the binary locally, but it also works without it.

skymarshal/skycmd/oidc_flags.go
```diff
@@ -28,6 +28,7 @@ type OIDCFlags struct {
 	HostedDomains      []string    `long:"hosted-domains" description:"List of whitelisted domains when using Google, only users from a listed domain will be allowed to log in"`
 	CACerts            []flag.File `long:"ca-cert" description:"CA Certificate"`
 	InsecureSkipVerify bool        `long:"skip-ssl-validation" description:"Skip SSL validation"`
+  InsecureEnableGroups bool      `long:"enable-groups" description:"Enable OIDC groups"`
```
aoldershaw, line 31: @xtremerui do we want this as an opt-in feature, or should it be opt-out instead? Making it opt-in means it'll be breaking (compared to 6.7.2 - 6.7.3 was breaking in an unintended way)

package.json
konstl000: More precisely, I followed the instructions [here](https://discuss.concourse-ci.org/t/error-in-dev-environment/992/2), since initially I was getting the same error message about main.css. But I tested the binary without these parameters too, it works fine.

skymarshal/skycmd/oidc_flags.go
aoldershaw: Or do we even need to provide the ability to opt-out?

konstl000: Alternatively, it can be true on default, but also overridable ...

package.json
aoldershaw: Could you revert the change? Those instructions aren't accurate - we don't use any those new dependencies (we don't even use Vue). I think the solution to the error message is just to run `yarn build`

skymarshal/skycmd/oidc_flags.go
konstl000: Just thought, it would make sense, since upstream dex does this ...

xtremerui: Good call. Making it opt-out is better for backward compatibility but at the same time opt-in flag makes user to aware that enabling groups is insecure. 

Given existing concourse users of OIDC connectors still have a chance to add the flag when they upgrading Concourse, IMO keep it opt-in is more aligned to upstream DEX.

package.json
konstl000: Done

skymarshal/skycmd/oidc_flags.go
aoldershaw: I don't think enabling groups is actually insecure here. See https://github.com/dexidp/dex/commit/21ab30d20767283b4696c992644432a38fc23849

> There's been some discussion in dexidp/dex#1065 regarding what to do about refreshing groups. As it stands today dex doesn't update any of the claims on refresh (groups would just be another one). The main concern with enabling it is that group claims may change more frequently.

So the insecure behaviour is around using the OIDC connector with refresh tokens, since those are long-lived and groups can change, but that change won't be propagated. But Concourse doesn't support refresh tokens, so it's no different from any other connector (when your groups change, you have to log out and log back in).

Also, being more aligned with upstream Dex doesn't necessarily make sense here - using groups is pretty central to Concourse's auth system, whereas Dex is just meant to be a general purpose identity provider

konstl000: I think, both opt-in and opt-out would make sense. Probably, a question of taste ...

xtremerui: @konstl000 in your use case, do you specified `--oidc-scopes` with `groups` ?

Maybe, we could set `InsecureEnableGroups` to true if `--oidc-scopes` contains `groups` since concourse needs both to work with groups claim.

konstl000: @xtremerui Actually not, because we just include groups into the openID scope on the Keycloak side. Probably, there are more people around who do that, considering that other tools (such as HashiCorp Vault) can rely on groups too ...

konstl000: Once I saw that it did not work, I tried to include groups via --oidc-scopes though :-)

aoldershaw: Makes sense to me - wasn't aware of `--oidc-scopes`. I just don't think it makes much sense to introduce a breaking change that forces operators to opt-in to a feature that, while potentially insecure in general, doesn't affect auth in Concourse

konstl000: I am quite fine with the opt-out option too. Should I change the PR accordingly?

konstl000: > Maybe, we could set `InsecureEnableGroups` to true if `--oidc-scopes` contains `groups` since concourse needs both to work with groups claim. WDYT @aoldershaw

That would be a breaking change too, at least for folks that are as lazy as me and just used to include groups in the openID scope.

xtremerui: @konstl000 ok thx!

konstl000: Now it is opt-out and is called --oidc-disable-groups

aoldershaw: > That would be a breaking change too, at least for folks that are as lazy as me and just used to include groups in the openID scope

Ah sorry, I didn't see https://github.com/concourse/concourse/pull/6436#discussion_r559676991 before I added my most recent comment - so just to be clear, in 6.7.2 and earlier, you didn't need to set `--oidc-scopes` to include `groups` - it picked them up by default?

skymarshal/skycmd/oidc_flags.go
```diff
@@ -28,6 +28,7 @@ type OIDCFlags struct {
 	HostedDomains      []string    `long:"hosted-domains" description:"List of whitelisted domains when using Google, only users from a listed domain will be allowed to log in"`
 	CACerts            []flag.File `long:"ca-cert" description:"CA Certificate"`
 	InsecureSkipVerify bool        `long:"skip-ssl-validation" description:"Skip SSL validation"`
+  DisableGroups bool      `long:"disable-groups" description:"Enable OIDC groups"`
```
xtremerui, line 31:
```suggestion
  DisableGroups bool      `long:"disable-groups" description:"Disable OIDC groups claims"`
```

xtremerui (changes requested):

skymarshal/skycmd/oidc_flags.go
konstl000: Yes, if the OIDC provider had `groups` in the openID scope, the earlier versions of Concourse would pick them up and use them without the `- -oidc-scopes` flag. This is also how our production Concourse is configured at the moment.

xtremerui: > Yes, if the OIDC provider had `groups` in the `openid` scope, the earlier versions of Concourse would pick them up and use them without the `--oidc-scopes` flag. 

So it depends on how OIDC provider returns claims based on scopes. In @konstl000 case Keycloak is configured to return groups claims by one of the default scopes [openid, profile, email] so `--oidc-scopes` is not needed. I agree that `--oidc-scopes` and disable groups claim should then be decoupled to achieve maximum flexibility.

xtremerui (approved):

taylorsilva:
@konstl000 FYI we're enabling 2FA for the Concourse org. We can see you don't have 2FA enabled. You'll be kicked from the Concourse org when we flip the switch in a few minutes. If you enable 2FA then you can re-join the Concourse org.

konstl000:
@taylorsilva, I just activated the 2FA. What are the further steps?

taylorsilva:
I actually don't think you'll need to do anything. You'll get added back automatically the next time the governance repo's github actions run https://github.com/concourse/governance/

PR #6441, opened 2021-01-18
[WIP] Add "limit-total-allocated-memory"

andy-paine:

* Refactored how the number of active containers is tracked/retrieved to use `containers` table as source of truth
* Add the configured container limits to the `containers` table
* Use the configured limits together with per-worker max resources flags to create new strategy that attempts to limit tasks landing on workers based on their resource usage (rather than just the number of tasks)
* Make "no workers" a retriable error so that large tasks do not immediately fail and instead wait for a worker with enough space

## Notes to reviewer:
There are quite a few changes to unpack that kind of go together here but could potentially be split if that would make things easier. I write pretty long commit messages but there are also some thoughts on bits that weren't necessarily the most obvious route below:

* Rather than tracking the number of active containers via the worker heartbeats, use the containers table and find all "creating" or "created" tasks. This then means `fly workers` reflects the state of the world that the ATC would use for any decision making as well as fixing a longstanding issue where the `fly workers` containers column showed a wildly different number to the `fly containers` output.
* By storing the containers configured limits in the `containers` table, we need to check that table in every strategy `Choose` call which could be slow. I decided (currently) to put this logic in the `ContainerRepository` code to keep the stuff that interacts with the `containers` table together but this may create an `O(n)` operation in the strategy code where `n = number of workers`. The alternative to this I think is to hydrate the "ActiveContainers" property of a worker object directly using a join in the main DB query so the worker object would contain the currently used resources directly without having to make another DB call per-worker.
* By making the error retriable we kind of lumped it in with the `enable-rerun-when-worker-disappears` flag which isn't quite right. I would like to go over this whole behaviour at some point and maybe have a command line option for "errors to retry" and have those come from some kind of known list of errors but for now it is _kind of related_ to disappearing workers (if you don't think about it too much!) so I've left it like that.

## Release Note
* A new placement strategy "limit-total-resources" can be used together with container limits so that large tasks will wait for available resources to become free before running

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/api/handler.go
+++ b/atc/api/handler.go
@@ -93,7 +93,7 @@ func NewHandler(
 	pipelineServer := pipelineserver.NewServer(logger, dbTeamFactory, dbPipelineFactory, externalURL)
 	configServer := configserver.NewServer(logger, dbTeamFactory, secretManager)
 	ccServer := ccserver.NewServer(logger, dbTeamFactory, externalURL)
-	workerServer := workerserver.NewServer(logger, workerTeamFactory, dbWorkerFactory)
+	workerServer := workerserver.NewServer(logger, workerTeamFactory, dbWorkerFactory, containerRepository)
 	logLevelServer := loglevelserver.NewServer(logger, sink)
 	cliServer := cliserver.NewServer(logger, absCLIDownloadsDir)
 	containerServer := containerserver.NewServer(logger, workerClient, secretManager, varSourcePool, interceptTimeoutFactory, interceptUpdateInterval, containerRepository, destroyer, clock)
--- a/atc/api/present/worker.go
+++ b/atc/api/present/worker.go
@@ -5,7 +5,7 @@ import (
 	"github.com/concourse/concourse/atc/db"
 )
 
-func Worker(workerInfo db.Worker) atc.Worker {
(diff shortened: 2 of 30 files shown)


DanielJonesEB:
For context, the folks at EngineerBetter are looking at this as part of a series of changes to get workers to only accept tasks that they have the available resources to run, so in our multi-tenant environment we can prevent folks from starting either: a bazillion smalls tasks at once; or, one massive task; and then OOMing their worker.

andy-paine:
We've done some more thinking about this and limiting CPU using this method doesn't make much sense since CPU just refers to shares rather than an absolute value. With this in mind we've made some refactors that just use memory as the limiting value and will probably just recommend using `limit-active-containers` for users who want to reduce CPU competition. I'll mark this as WIP until we've got those ready but don't feel pressured to review it quite yet.

andy-paine:
I'm going to close this in favour of breaking it out into #6469 and #6470 which tidy things up a bit and do one thing each

PR #6462, opened 2021-01-22
audit: log user IP address

taylorsilva:

From concourse/concourse#6456

Has the auditor also log a users IP address. Will log based on available headers (X-REAL-IP and X-FORWARDED-FOR) and fall back to `RemoteAddr` if no header is available.

## Release Note

* The audit logs now log a user's IP address

## Contributor Checklist
- [x] [Signed] all commits
- [x] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -214,15 +214,16 @@ type RunCommand struct {
 	DefaultMemoryLimit *string `long:"default-task-memory-limit" description:"Default maximum memory per task, 0 means unlimited"`
 
 	Auditor struct {
-		EnableBuildAuditLog     bool `long:"enable-build-auditing" description:"Enable auditing for all api requests connected to builds."`
-		EnableContainerAuditLog bool `long:"enable-container-auditing" description:"Enable auditing for all api requests connected to containers."`
-		EnableJobAuditLog       bool `long:"enable-job-auditing" description:"Enable auditing for all api requests connected to jobs."`
-		EnablePipelineAuditLog  bool `long:"enable-pipeline-auditing" description:"Enable auditing for all api requests connected to pipelines."`
-		EnableResourceAuditLog  bool `long:"enable-resource-auditing" description:"Enable auditing for all api requests connected to resources."`
-		EnableSystemAuditLog    bool `long:"enable-system-auditing" description:"Enable auditing for all api requests connected to system transactions."`
-		EnableTeamAuditLog      bool `long:"enable-team-auditing" description:"Enable auditing for all api requests connected to teams."`
-		EnableWorkerAuditLog    bool `long:"enable-worker-auditing" description:"Enable auditing for all api requests connected to workers."`
-		EnableVolumeAuditLog    bool `long:"enable-volume-auditing" description:"Enable auditing for all api requests connected to volumes."`
+		EnableBuildAuditLog     bool   `long:"enable-build-auditing" description:"Enable auditing for all api requests connected to builds."`
+		EnableContainerAuditLog bool   `long:"enable-container-auditing" description:"Enable auditing for all api requests connected to containers."`
+		EnableJobAuditLog       bool   `long:"enable-job-auditing" description:"Enable auditing for all api requests connected to jobs."`
+		EnablePipelineAuditLog  bool   `long:"enable-pipeline-auditing" description:"Enable auditing for all api requests connected to pipelines."`
+		EnableResourceAuditLog  bool   `long:"enable-resource-auditing" description:"Enable auditing for all api requests connected to resources."`
+		EnableSystemAuditLog    bool   `long:"enable-system-auditing" description:"Enable auditing for all api requests connected to system transactions."`
+		EnableTeamAuditLog      bool   `long:"enable-team-auditing" description:"Enable auditing for all api requests connected to teams."`
+		EnableWorkerAuditLog    bool   `long:"enable-worker-auditing" description:"Enable auditing for all api requests connected to workers."`
+		EnableVolumeAuditLog    bool   `long:"enable-volume-auditing" description:"Enable auditing for all api requests connected to volumes."`
+		ClientIPHeader          string `long:"client-ip-header" description:"A header that a reverse proxy will set to the client's IP address. If not set then X-REAL-IP and X-FORWARD-FOR will be used in that order."`
 	}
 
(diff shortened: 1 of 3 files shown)


atc/auditor/auditor.go
```diff
@@ -164,6 +164,6 @@ func (a *auditor) ValidateAction(action string) bool {
 func (a *auditor) Audit(action string, userName string, r *http.Request) {
 	err := r.ParseForm()
 	if err == nil && a.ValidateAction(action) {
-		a.logger.Info("audit", lager.Data{"action": action, "user": userName, "parameters": r.Form})
+		a.logger.Info("audit", lager.Data{"action": action, "user": userName, "ip": r.RemoteAddr, "parameters": r.Form})
```
aoldershaw, line 167: Small note: `r.RemoteAddr` includes both the IP address *and* the port (which is going to be some random higher-order ephemeral port) - I wonder if we want to strip that out? Or possibly include it as a separate field? Or just rename `ip` to `addr`?

We could use `net.SplitHostPort` if we want to extract the ip in isolation

taylorsilva: Seems that getting a users real ip can be tricky... https://golangbyexample.com/golang-ip-address-http-request/

I'm going to change it to check those headers in the order of `X-REAL-IP`, `X-FORWARDED-FOR`, and then RemoteAddr, parsing it if needed.

aoldershaw: Hm, does that mean you'd be able to spoof the IP that gets logged by providing one of those headers? Not sure if nginx or other reverse proxies would overwrite this header, but if not (or you aren't using a reverse proxy), it'd be easy to lie about your IP address (in which case, is there much value in including it in the audit log?)

aoldershaw:
I'm a little torn - the feature request was for security auditing, but using `X-Real-Ip` and `X-Forwarded-For` mean that the IP address can easily be spoofed by a malicious user, making the audit log unreliable. That said, it's likely pretty common to have a reverse proxy in front of Concourse, making `r.RemoteAddr` useless. However, in that case, maybe it's just the responsibility of the reverse proxy to keep an audit log of requests (something they do, anyway) - it won't be the same granularity as Concourse's audit log since it doesn't know about actions in Concourse, but at least it can't be spoofed and includes the requested path (which is enough to deduce the action).

Perhaps an alternative (higher-effort) solution is to allow logging a custom header (configured via flags). That way, users could choose to log `X-Real-Ip` or `X-Forwarded-For` (opting-in to spoofing attacks), or they could log something else entirely. For instance, the reverse proxy could set a unique request ID in `X-Request-ID`, and Concourse could log that - to find the actual IP, they could consult their reverse proxy's logs for that request ID.

taylorsilva:
> I'm a little torn - the feature request was for security auditing

I agree with this entirely, it's all I was thinking about as I was writing it and noted as much in my commit message

> This is not foolproof and is easily spoofed by a client.

Went ahead and added a config option for adding a custom header. It will still fallback to the other headers and remoteAddr.

atc/atccmd/command.go
```diff
@@ -214,15 +214,16 @@ type RunCommand struct {
...
+		EnableJobAuditLog       bool   `long:"enable-job-auditing" description:"Enable auditing for all api requests connected to jobs."`
+		EnablePipelineAuditLog  bool   `long:"enable-pipeline-auditing" description:"Enable auditing for all api requests connected to pipelines."`
+		EnableResourceAuditLog  bool   `long:"enable-resource-auditing" description:"Enable auditing for all api requests connected to resources."`
+		EnableSystemAuditLog    bool   `long:"enable-system-auditing" description:"Enable auditing for all api requests connected to system transactions."`
+		EnableTeamAuditLog      bool   `long:"enable-team-auditing" description:"Enable auditing for all api requests connected to teams."`
+		EnableWorkerAuditLog    bool   `long:"enable-worker-auditing" description:"Enable auditing for all api requests connected to workers."`
+		EnableVolumeAuditLog    bool   `long:"enable-volume-auditing" description:"Enable auditing for all api requests connected to volumes."`
+		ClientIPHeader          string `long:"client-ip-header" description:"A header that a reverse proxy will set to the client's IP address. If not set then X-REAL-IP and X-FORWARD-FOR will be used in that order."`
```
aoldershaw, line 226: I was thinking more along the lines of just adding some header(s), not necessarily for the IP address, that we just pass along as is and (don't try to infer any meaning from). In addition to logging the custom header(s), we'd log `r.RemoteAddr` as is.

So, if the operator configured logging the header `X-Forwarded-For`, then we'd log something like:

```
{"action": "...", "user": "...", "ip": "<remote proxy IP>", headers: {"X-Forwarded-For": "<local ip, possibly a chain, possibly spoofed>"}}
```

Or, if the operator configured logging the header `X-Request-Id`:

```
{"action": "...", "user": "...", "ip": "<remote proxy IP>", headers: {"X-Request-Id": "<some id that can be traced through the remote proxy logs>"}}
```

What do you think about that? I think always logging `r.RemoteAddr` is a good idea, since we can't possibly know what configuration of proxies is sitting in front of Concourse, if any. Allowing them to log any header(s) also seems like a good idea, since it allows you to log things like `X-Request-Id`, which can't be spoofed

cirocosta: > Allowing them to log any header(s) also seems like a good idea, since it allows you to log things like X-Request-Id, which can't be spoofed (if you've configured your reverse proxy correctly).

and (just to add to the "it's hard to get a meaningful ip"), nowadays it's just sooooo easy to use any of those browsers w/ VPN support built-in, or even [tor](https://www.torproject.org/) (one can literally just go to a private tab on brave and enable it or download the tor browser)

aoldershaw: @taylorsilva just pinging you about this - any thoughts about this approach?

taylorsilva:
I think the approaches mentioned make sense. At the same time, the scope on this is growing a bit much. I think it goes working towards a great final solution but I wasn't out to get to that point in this PR. I was looking to do a first pass. I'm gonna close this PR for now and make an issue from the discussion thread. I'll put it in the icebox.

PR #6465, opened 2021-01-24
Allow favoriting instance groups

aoldershaw:

Screen Shot 2021-01-23 at 9 41 49 PM

Previously, only individual instances could be favourited. When instances were favourited, the instance group card would be displayed in the favourites section, but with the set of instances filtered down to only those that were favourited.

I figured you'd typically only care about a few instances, and the old approach made it easy to find those instances at a glance - but there are definitely valid cases for wanting to see all instances at once, and this PR makes that possible (without needing to favourite each individual instance).

* Store favourited instance groups in `localStorage`
* Store sidebar state in `localStorage` (rather than `sessionStorage`)
  * Wasn't strictly necessary, but allowed simplifying the code, and if you use the sidebar, you'll probably want it to stay open when you open a new tab
* Display `PipelineCard` for each favourited instance, even when not viewing instance groups via the `group:...` filter
  * The `PipelineCards` contain both the group name + instance vars
  * In contrast, when viewing instance groups, the `PipelineCards` only contain the instance vars, because the headers outside of the cards contain the instance group name
* Display pipeline list items in the sidebar for favourited instances using the same `fly` notation
  * i.e. `group-name/var1:val1,var2:val2`

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/web/elm/benchmarks/Benchmarks.elm
+++ b/web/elm/benchmarks/Benchmarks.elm
@@ -683,6 +683,7 @@ sampleSession =
     , expandedTeamsInAllPipelines = Set.empty
     , collapsedTeamsInFavorites = Set.empty
     , favoritedPipelines = Set.empty
+    , favoritedInstanceGroups = Set.empty
     , hovered = HoverState.NoHover
     , sideBarState =
         { isOpen = False
--- a/web/elm/benchmarks/Benchmarks.elm
+++ b/web/elm/benchmarks/Benchmarks.elm
@@ -143,6 +143,7 @@ buildView session model =
                     |> Maybe.map
                         (\j ->
                             { pipelineName = j.pipelineName
+                            , pipelineInstanceVars = j.pipelineInstanceVars
                             , teamName = j.teamName
                             }
                         )
--- a/web/elm/src/Application/Application.elm
+++ b/web/elm/src/Application/Application.elm
@@ -89,6 +89,7 @@ init flags url =
             , screenSize = ScreenSize.Desktop
             , timeZone = Time.utc
             , favoritedPipelines = Set.empty
+            , favoritedInstanceGroups = Set.empty
             , route = route
             }
 
@@ -116,6 +117,7 @@ init flags url =
       , GetScreenSize
       , LoadSideBarState
       , LoadFavoritedPipelines
+      , LoadFavoritedInstanceGroups
       , FetchClusterInfo
       ]
         ++ handleTokenEffect
--- a/web/elm/src/Application/Application.elm
+++ b/web/elm/src/Application/Application.elm
@@ -456,9 +456,7 @@ view model =
 subscriptions : Model -> List Subscription
 subscriptions model =
     [ OnNonHrefLinkClicked
-    , OnTokenReceived
-    , OnSideBarStateReceived
-    , OnFavoritedPipelinesReceived
+    , OnLocalStorageReceived
     , OnWindowResize
     ]
         ++ (if model.session.draggingSideBar then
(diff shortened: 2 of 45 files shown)


matthewpereira:
One idea to think about and get your opinion on @aoldershaw:

In the design studio for this feature, folks seemed to like representing the 'tree structure' of a group that contains instances. Do you think that would be worth carrying over to the favorites, whether or not the group itself was favorited?

Something like this (forgive the unrealistic pipeline titles):
Screen Shot 2021-01-29 at 2 28 24 PM

aoldershaw:
@matthewpereira makes sense - if we chose to have allow expanding instance groups in the "all pipelines" section of the sidebar, it'd probably look similar to your mockup.

What should the badge display if both the group and instances within the group are favourited? e.g. if `a-favorited-group` had two favourited instances and was favourited itself, would we display `12` as is done now, or `2` (for consistency with unfavourited groups, at least in the mockup)?

chenbh (commented):
I'm not sure I like displaying the instance vars in the card header, it kinda looks off when there's a combination of favourited instanced and non-instanced pipelines
 
Screen Shot 2021-02-01 at 4 11 06 PM

Wild ideas:
1. display the pipeline name as `pipeline/var1:value1,var2:value2`. Con: this feels like it clutters up the text
1. display the pipeline name, but also have a greyed out `value1-value2` on the same line. Con: might not be immediately obvious what the extra text means
1. add the instanced group badge and show the instance vars when you hover over the name. Con: I can't tell at a glance what the pipeline actually refers to

aoldershaw:
@chenbh feel free to disagree, but IMO things don't look *so* wonky when there's a single extra row in the header (currently one instance var) - my screenshot in the PR description shows an example of that. What if we followed @matthewpereira's suggestion for the sidebar and displayed all instance vars underneath the group name

Screen Shot 2021-02-02 at 10 23 37 AM

It all fits in this cherry-picked example with short instance var names/values, but I imagine having more than ~2 instance vars will be rare. If we find that this often overflows, perhaps we can re-evaluate.

It's tricky because there's not much horizontal space to work with in the pipeline cards, but at least separating the instance vars from the name affords us some more space.

chenbh:
@aoldershaw I like that design, it's a great compromise between too much vertical space and too little horizontal space.

matthewpereira:
@aoldershaw For now, I'd recommend leaving out expand/collapse functionality, leaving out indenting, and just changing the icon to differentiate pipeline instances. My goal with including the 'group' along with pipeline instances that are favorited independently is just to simplify the labeling and make it a little easier to recognize what's going on.

We could explore nested functionality in the future, but I think right now having teams expand/collapse is more than enough complexity to work with. I'd prefer to roll out the most simple option and then see how that tests.

aoldershaw:
@chenbh @matthewpereira I've updated the sidebar to use the `/` icon and fixed a few edge-case bugs I noticed (header height issues with instanced pipelines w/ no instance vars and no other favorited instances, and a reordering bug when the first instance is archived). Still have yet to address https://github.com/concourse/concourse/pull/6465#issuecomment-771714672, though

aoldershaw:
@chenbh update - I have now addressed https://github.com/concourse/concourse/pull/6465#issuecomment-771714672, so now instance vars are displayed in the same row in the favourites section when *not* viewing an instance group:

Screen Shot 2021-02-10 at 5 54 58 PM

When viewing an instance group, however, the instance vars are allowed to span multiple rows, e.g.
Screen Shot 2021-02-10 at 5 55 55 PM

The thought behind this is that instances within a group should typically all have the same set of instance vars, so there shouldn't be empty space in the header and things will look uniform, so we might as well take advantage of the extra vertical space.

Let me know what you think

chenbh (approved):
lgtm!

Issue #4340, opened 2019-09-02
Exhausted container limit, contradicting container counts, container found for deletion constantly increasing


gerhard:
## Bug Report

This is an uber-issue (3 issues in 1) since it's been snowballing since at least 2017, specifically #1669. The issue was resolved in some Concourse versions, but it is [back with a vengeance since at least v5.4.0](https://github.com/concourse/concourse/issues/4150#issuecomment-514555817).

`fly containers` is reporting one number of containers (329), `fly workers` is reporting another (500). Which one is right? Since no more jobs can be scheduled, I would say `fly workers`. Attaching both files for inspection.

[fly-workers.txt](https://github.com/concourse/concourse/files/3566198/fly-workers.txt)
[fly-containers.txt](https://github.com/concourse/concourse/files/3566199/fly-containers.txt)

Lastly, the number of containers found for deletion is constantly increasing, it only resets itself when we delete workers.

## Steps to Reproduce

Not sure. Container counts are off, [containers are not GC'ed as documented](https://concourse-ci.org/garbage-collection.html#container-collection), [Caching & Retention is TODO](https://concourse-ci.org/caching-and-retention.html).

## Expected Results

I would expect containers to eventually be garbage collected.

## Actual Results

![image](https://user-images.githubusercontent.com/3342/64118129-e7d42180-cd8e-11e9-9d19-2da849799067.png)

## Additional Info

[Concourse Metrics](https://metrics.ci.rabbitmq.com/dashboard/snapshot/Nx81ufUxODMnbdbFbtgjj1hO12eFOelF?orgId=1)

![image](https://user-images.githubusercontent.com/3342/64121524-f6bed200-cd96-11e9-9b74-e3840d2bccc8.png)

## Version Info

* Concourse version: 5.4.1
* Deployment type: BOSH
* Infrastructure/IaaS: GCP
* Did this used to work? Yes, worked fine in v5.3 and prior v5.x versions, stopped working after 5.4.0

michaelklishin:
This issue is still relevant and plagues team RabbitMQ every few weeks.

vito:
@michaelklishin Super strange, I don't think we've seen others report this so sorry that this issue didn't get much response. We usually watch for higher activity since we don't really have the bandwidth to chase down everything since it can take a lot of time to get to the bottom of issues affecting a particular environment.

[...]

PR #6469, opened 2021-01-26
atc/db: Get worker.ActiveContainers from containers table

andy-paine:
Signed-off-by: Andy Paine <andy.paine@engineerbetter.com>

_Bug Fix_ | Feature | Documentation

Helps (partially) with #4340

Rather than tracking the active container count via the heartbeat, just use the containers table instead. This gives a more accurate representation of how many containers there are from the ATC's point of view since this is what is used for scheduling decisions and brings the `fly workers` and `fly containers` numbers in sync. This data is refreshed from the workers slightly less frequently than the heartbeats but is updated by the ATC as soon as a container is created so has faster initial feedback which may help with scheduling decisions.

## Notes to reviewer:
-

## Release Note
* The number of `containers` in the `fly workers` output will now match the containers returned by `fly containers`

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


clarafu:
Hi! I think I'm a little hesitant of changing the output of `fly containers` from the containers that are coming directly from the worker to the containers that are known on the database side. This is because the current representation of the containers from the worker is the only method that users can easily take a look at the current state of the containers within the worker. There are times when the containers within the workers gets out of sync with the containers in the database, and the users can use the `fly containers` command to try to figure out what is going on within the workers that are causing issues that they are seeing within Concourse. I think rather than changing the output of this fly command, maybe it would be more useful to add another command or flag that will show you a representation of what ATC thinks is the current state of your containers are?

andy-paine:
I don't think this PR changes the `fly containers` command - it just brings the count of containers output of `fly workers` in line with `fly containers`. Given that this is purely a number with no extra details, this output doesn't really help that much with debugging.

The only real use right now for it is "fly containers reports far more containers than fly workers - something in the ATC/DB must be wrong." To find out which containers are in the DB and shouldn't be, you still need to somehow exec onto the worker and query Garden directly. I could maybe see a place for `expected_containers` and `reported_containers` count fields in the worker API though?

clarafu:
Oh sorry about that! I misread your pull request description, totally my bad. I do agree that having the two container counts be different between `fly workers` and `fly containers` is confusing and only the users that know about the fact that `fly containers` is reporting containers directly from the worker and `fly workers` are reporting containers from the database would be able to use the information that comes from the commands correctly. I think your idea of having there be two columns of `expected_containers` and `reported_containers` would be a really good idea!

clarafu:
@andy-paine Not sure if you are busy but I was just wondering if you would be interested in adding the two columns rather than replacing the existing column with new data? Or if that's not something that you would like to do in this PR? It should be just keeping the existing column data and renaming it, then adding a new column with the new data that you added!

BooleanCat:
Hey @clarafu - I'm Tom from EngineerBetter. I'll make the changes Andy and yourself suggested here (renaming the existing column and adding a new column).

BooleanCat:
I've started with a shallow rename of the DB fields. Just to check, are we suggesting a deep rename of interfaces here such as [this interface](https://github.com/concourse/concourse/blob/9b1818ae67e2dfa7bc1082eb0e5d9152ca12c221/atc/db/worker.go#L62) too? Happy to do this but just wanted to point out that that will touch a few areas of the codebase wherever this interface is used and I'm not sure if that's considered a "breaking change" with versioning considerations.

clarafu:
Thanks for picking this PR up! I think renaming the DB fields along with the interfaces would be best, I don't think we need to consider this a breaking change because it is just adding a new column in the DB and renaming a field on the `fly workers` command. We usually consider something a breaking change when there is some large behavioural changes that users would need to change in their current workflow in order to match the new behaviour. The only case that I can think of that would make this breaking is if there was a user that depended on the specific `containers` string to be outputted by the `fly workers` command. But we don't have official support for fly or our API so I don't think we need to stress about that 👍

BooleanCat:
Renamed `active_containers` to `expected_containers`. I'll work on adding back Andy's change as `reported_containers` when I'm back at work next week 👍

atc/db/worker_factory.go
```diff
@@ -57,7 +58,8 @@ var workersQuery = psql.Select(`
 		w.ephemeral
 	`).
 	From("workers w").
-	LeftJoin("teams t ON w.team_id = t.id")
+	LeftJoin("teams t ON w.team_id = t.id").
+	LeftJoin("(SELECT worker_name, COUNT(handle) reported_containers FROM containers GROUP BY worker_name) cs ON w.name = cs.worker_name")
```
clarafu, line 62: This should be populating the `expected_containers` count rather than `reported_containers` because the `reported_containers` should be containing the number of containers that the worker heartbeating returns back. (but this is the same as what I mentioned in my previous review comment)

I was thinking here that maybe instead of having the `expected_containers` count be queried out everytime we select for a worker, we could have a method on the `db.Worker` interface called `ExpectedWorker() (int, error)` and it will run a query to find the number of containers for that worker in the containers table. The reason for this is that we can separately run the method to find the reported containers when we need it. If we put the query within the `workersQuery` then everytime we fetch out a `db.Worker` object it will run this query even in places that we don't care about the `expected_containers` (for example, grabbing out all the workers for container placement). The only time we will need the `expected_containers` is in the api so we can just have the api run the method to grab out the number of `expected_containers`. Similar to https://github.com/concourse/concourse/blob/0af59c6a895b1f6acfeaade85d359699ddc29e8b/atc/api/present/worker.go#L21-L24. Let me know what you think!

clarafu (changes requested):
For your question about `tsa/cmd/tsa`, I don't think you need to add any tests and your changes seem appropriate where it is just renaming the field. I think you don't need to add any new tests because we arn't changing any behaviour, since we are just renaming a field. As long as there are existing tests that test the new renamed field, I think that is sufficient!

I also see there are some merge conflicts, would you be able to merge or rebase your branch ontop of master?

Thanks! 😄

xtremerui:
closing the PR for now. Feel free to reopen it.

PR #6470, opened 2021-01-26
atc/worker: Add new "total memory" placement strategy

andy-paine:

Using the configured container limits for a task (or the global defaults), use the amount of memory currently running on a worker to decide if a task can can't be placed. This is similar to the active tasks/active volumes strategies but attempts to be a bit more refined by considering the size of tasks rather than just the number of them. Usage of the ContainerRepository may mean for large installations (100s of workers) that the strategy could be slow - if this appears to be the case we could implement this instead as a property of the worker and fetch it in a JOIN when we fetch all the compatible workers. Add some worker flags to allow for per-worker allocations so that the strategy can be used even when
workers are not a homogenous size. Not providing these flags on a worker makes this strategy _always_ treat it as a valid candidate.

This system is designed to be opt-in from both a user and operator perspective. Operators have to opt-in by configuring the strategy and the workers to report their memory limits and users can use it to try and ensure they don't have large tasks landing at the same time. Ideal workflow would be for these things to not turn a job orange as there is no worker found but I think that's a more general problem that I may well attempt to solve in another PR - this at least makes it so that you don't get all your large tasks scheduling at once and have them all fail.

## Release Note
* Add a new strategy to limit the total amount of memory assigned to an individual worker based on the `container_limits` configured for a task

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist
- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


atc/worker/placement.go
```diff
@@ -293,3 +304,58 @@ func (strategy *LimitActiveVolumesPlacementStrategyNode) ModifiesActiveTasks() b
...
+
+		if (allocatedMemory + requestedMemory) < *w.AllocatableMemory() {
+			candidates = append(candidates, w)
+		}
+	}
+
+	if len(candidates) == 0 {
+		return nil, NoWorkerFitContainerPlacementStrategyError{
+			Strategy: strategy.StrategyName(),
+			Reason:   message.NewPrinter(language.English).Sprintf("task requested %d bytes of memory and no workers had enough available memory", requestedMemory),
```
xtremerui, line 349: Is it necessary to indroduce new libraries (message and language) here?

atc/worker/placement_test.go
```diff
@@ -652,62 +669,62 @@ var _ = Describe("LimitActiveVolumesPlacementStrategyNode", func() {
...
-							)
-							Expect(chooseErr).ToNot(HaveOccurred())
-							return chosenWorker
-						}).Should(Or(Equal(compatibleWorker1), Equal(compatibleWorker2)))
-					})
-				})
+			BeforeEach(func() {
+				worker1AllocatedMemory = 1500
+				worker2AllocatedMemory = 1500
+				requestedMemory = 1000
```
xtremerui, line 702: wondering what if `requestedMemory` is below 500 in this case i.e. when there is worker with configured memory limit satisfying and also worker without configured allocatable memory available, which worker will be chosen?

xtremerui:
I am assuming a task without memory limit(i.e. requested memory is 0) configured would be able to choose a worker that has allocatable memory limit configured (together with all other workers that without memory limit configured) right?

 In such case, that worker could be running multiple containers(without memory limit configured) actually ALLOCATED certain amount of memory already. However, from the task pespective, the SQL that fetching the total allocated memory still returns 0 for the worker, which would make the worker available for a task requires large memory. When that happens, due to the unknown amount of allocated memory of the worker, that the task could still fail due to insufficient memory.

xtremerui:
I will close this PR due to inactivity. Feel free to reopen if it got picked up again.

Issue #6038, opened 2020-09-03
git resource cannot resolve host when using Docker for Mac


chenbh:
Tldr: `concourse/git-resource` is broken for Docker on Mac and we should use the ubuntu versions (e.g. docker compose with `concourse/concourse:ubuntu` or mirror `concourse/git-resource:ubuntu` as a custom resource type)

I'm using curl for all of the examples here, but the same thing occurs with git.

## Problem
So we use `alpine:edge` as the base for all of our resources and it seems like the latest version of `musl` ([1.2.1-r0](https://pkgs.alpinelinux.org/packages?name=musl&branch=edge)) somehow made it so that it no longer tries IPv4 lookup if IPv6 fails (maybe as part of [this fix?](https://git.musl-libc.org/cgit/musl/tree/WHATSNEW#n2223)). 
Unfortunately IPv6 is [not supported](https://docs.docker.com/docker-for-mac/troubleshoot/#known-issues) on Docker for Mac and so all DNS queries will fail to resolve if musl 1.2.1-r0 is installed.

```
$ docker run -it alpine:edge
# apk add curl
...
# curl https://github.com
<html>...</html>                                        <- works
# apk add --upgrade musl
(1/1) Upgrading musl (1.1.24-r9 -> 1.2.1-r0)
# curl https://github.com
curl: (6) Could not resolve host: github.com            <- breaks
```
Doing the exact same steps above but with `alpine:3` (which only supports musl 1.1.24-r9) works perfectly fine.

## Steps to reproduce

While the git resource would normally be created by Guardian as a container-in-container, the same effect still occurs if we run it as a normal Docker container:
```
$ docker pull concourse/git-resource
$ docker run -it concourse/git-resource
/usr/lib # curl https://github.com
curl: (6) Could not resolve host: github.com
```

## Workaround
The ubuntu version of git-resource is still working fine and comes packaged with `concourse/concourse:ubuntu` (or `6-ubuntu`, `6.5-ubuntu`, etc). Changing it at the `docker-compose.yml` level should fix it

Alternatively we can define the ubuntu resource as a resource type and use it at the pipeline level.
```
resource_types:
- name: git-ubuntu
  type: registry-image
  source: 
    repository: concourse/git-resource
    tag: ubuntu

resources:
- name: concourse
  type: git-ubuntu
  source:
    uri: https://github.com/concourse/concourse.git
```

## Fix
[...]

PR #6479, opened 2021-01-27
introduce new integration test suite

vito:

This new test suite aims to consolidate the following CI jobs / test suites:

* upgrade
* downgrade
* bosh-topgun-*
* k8s-topgun-*
* testflight

These tests are easy to run on a dev machine: just run `go test ./integration/...`. They work by using `docker-compose`, so you don't need BOSH or K8s.

Included so far are tests covering upgrade/downgrade, so we can replace the `upgrade` and `downgrade` jobs in the pipeline, along with a Vault test suite which should let us remove the Vault tests from TopGun.

We should add more test coverage for different auth flows before removing the Vault tests from TopGun, though - this suite doesn't provide equivalent coverage yet. Only enough to verify that Vault works. The TopGun failures are actually because it's running against a ridiculously old version of Vault (v0.7.3), because it uses the world's best Vault bosh release, [`vito/vault-boshrelease`](https://github.com/vito/vault-boshrelease). Now that we can test Vault using Docker Compose, we can eliminate our dependency on that release and even CI against new `vault` Docker images as they're released.

I think we should probably still have some sort of coverage for BOSH/K8s, but it should probably be reduced to only tests that are K8s or BOSH-specific - maybe they can even be unit tests. We should be able to test the same Concourse-specific lifecycle behavior through Docker Compose.

As for `testflight`, there's no immediate need to pull it in, but if we want to it could probably be tucked under `integration/core` as a suite that runs against a single Docker Compose deployment.

## Notes to reviewer:

I went through a bit of churn while figuring out the simplest and clearest way to structure all this, so it's probably best to look at the whole diff as some commits may be made irrelevant by later commits. Let me know if anything seems really confusing! I also want to write a README under `integration/` so any feedback you have would help me write it.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


Dockerfile
```diff
@@ -4,22 +4,6 @@
 ARG base_image=concourse/dev
 FROM ${base_image} AS base
 
-# generate keys (with 1024 bits just so they generate faster)
```
vito, line 7: These keys are now checked in, so they don't need to be generated in dev builds anymore. This makes it easier to run the tests against both 'dev' images and 'latest' Concourse images (e.g. upgrade/downgrade tests).

cmd/concourse/main.go
```diff
@@ -1,4 +1,4 @@
-//go:generate go run github.com/markbates/pkger/cmd/pkger
```
vito, line 1: this fix (and the moving `pkged.go` under `cmd/concourse`) should be backported to 6.7.x; without it, `fly` commands shell out to `go` on every run and seem to result in fetching and building `github.com/gobuffalo/here`, which is scary

integration/creds/creds_test.go
```diff
@@ -0,0 +1,172 @@
...
+			require.NoError(t, err)
+		})
+
+		t.Run("interpolates job builds", func(t *testing.T) {
+			// build will fail and return err if any values are wrong
+			err := fly.Run("trigger-job", "-w", "-j", "test/some-job")
+			require.NoError(t, err)
+		})
+
+		t.Run("interpolates one-off builds with job inputs", func(t *testing.T) {
```
aoldershaw, line 110: I get

```
missing inputs: some-resource
```

on this test. Not sure why - we trigger `test/some-job` right before, so I'd think it'd have the correct input

integration/ops/suite_test.go
```diff
@@ -0,0 +1,54 @@
...
+		err := fly.Run("trigger-job", "-j", "test/say-hello", "-w")
+		require.NoError(t, err)
+	})
+
+	t.Run("can still run checks", func(t *testing.T) {
+		err := fly.Run("check-resource", "-r", "test/mockery")
+		require.NoError(t, err)
+	})
+
+	t.Run("can still reach the internet", func(t *testing.T) {
```
aoldershaw, line 43: Hm, not sure why but I can't seem to get this test to work. My job...doesn't seem to be able to reach the internet (`unable to resolve host address 'example.com'`). Probably something misconfigured on my end but not sure what

aoldershaw (commented):
Just done a really quick pass, but looks really cool so far! Really like how the tests are structured. The `xyztest.Cmd` approach seems to be a really nice level of indirection - the tests are very readable and seem easy to write. I'm excited to extend this suite to add more high-level test coverage!

There are a couple tests that are failing for me when running locally - any ideas what might be causing that?

integration/ops/suite_test.go
vito: Yeah this one's finicky for me too, but in my case it was on my end - WSL2 sets a private IP in `/etc/resolv.conf` which isn't addressable within the container. Putting `8.8.8.8` in there and disabling WSL's auto-configuration fixed it.

Are you on Docker for Mac? You could try upgrading to 3.1.x: https://github.com/concourse/concourse/issues/6038

integration/creds/creds_test.go
vito: Weird, lemme take a look. I never saw that failure myself.

cmd/concourse/main.go
vito: backported: #6480

integration/ops/suite_test.go
aoldershaw: Currently seeing the issue on 3.1.0 - will try to skip the cleanup step so I can prod at the deployment after the test fails

integration/creds/creds_test.go
vito: Shot in the dark: make sure you run `docker-compose build` in the root of the repo. Maybe it was running against an old build?

aoldershaw: that was after running `docker-compose build`

it's weird though - I disabled the test cleanup, and tried running the same `fly execute` command against the deployment and it worked. will keep investigating

integration/internal/dctest/cmd.go
```diff
@@ -0,0 +1,96 @@
...
+	for _, env := range os.Environ() {
+		if strings.HasPrefix(env, "TEST_") {
+			dc = dc.WithEnv(env)
+		}
+	}
+
+	// clean up docker-compose when the test finishes
+	cleanupOnce(t, func() {
+		if t.Failed() {
+			err := os.Mkdir("logs", os.ModePerm)
```
aoldershaw, line 44:
```suggestion
			err := os.MkdirAll("logs", os.ModePerm)
```

Getting errors when it already exists

integration/creds/creds_test.go
aoldershaw: We looked into it, and it turns out I'm dumb and need to pull `concourse/dev:latest` every once in a while

cmd/concourse/main.go
aoldershaw: We merged the fix into `master` in #6482 so we could unblock the `release/7.0.x` branch

integration/internal/dctest/cmd.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	return strings.TrimRight(strings.Replace(out, "0.0.0.0", "127.0.0.1", 1), "\n"), nil
+}
+
+var onces = new(sync.Map)
+
+func cleanupOnce(t *testing.T, cleanup func()) {
```
aoldershaw, line 88: When could the `Cleanup` funcs get called more than once?

aoldershaw: Ignore me, I think I see it now - we spin up multiple `dctest.Cmd`'s, but they actually point to the same deployment, so we just want to `docker-compose down` once?

I wonder if we'll ever have a single test that needs two different deployments? Not sure why we would need to do that, rather than just adding overrides or using the `ypath` package to create a more complicated `docker-compose` deployment to suit our needs, though

integration/internal/cmdtest/cmd.go
```diff
@@ -0,0 +1,145 @@
...
+	Stderr io.Writer
+	Silent bool
+}
+
+func (cmd Cmd) WithEnv(env ...string) Cmd {
+	newEnv := make([]string, len(cmd.Env))
+	copy(newEnv, cmd.Env)
+	newEnv = append(newEnv, env...)
+
+	cmd.Env = newEnv
```
aoldershaw, line 29: Could this be written as

```go
cmd.Env = append(cmd.Env, env...)
```

?

Doesn't seem to mutate the original slice: https://play.golang.org/p/IrYwyrju3JV

integration/internal/dctest/cmd.go
vito: If a test calls `dctest.Init` twice - e.g. the upgrade/downgrade tests. It won't call each function twice, but it would call both functions, and in this case we should only need to run one `docker-compose` cleanup, and it should be whichever one was added last, which works since `t.Cleanup` is called in reverse order.

integration/internal/cmdtest/cmd.go
```diff
@@ -0,0 +1,145 @@
...
+	cmd.Stdin = in
+	return cmd
+}
+
+func (cmd Cmd) OutputTo(out io.Writer) Cmd {
+	cmd.Stdout = out
+	return cmd
+}
+
+func (cmd Cmd) Run(args ...string) error {
```
aoldershaw, line 62: Now I'm debating whether or not it makes sense to return an error here or to assert internally and add a `Cmd.ExpectFailure()` or something that you call prior to the `cmd.{Run,Output,OutputJSON}` that flips the assertion.

I expect that you'll nearly always want the command to succeed, so the extra `NoError` assertions just add clutter (though, I've admittedly gotten pretty used to skimming over `Expect(err).ToNot(HaveOccurred())` and `require.NoError(t, err)`, so perhaps it's not so bad). The `ExpectFailure` API is a bit awkward since you define it before actually running the command, e.g.:

```
fly.ExpectFailure().Run("trigger-job", ...)
```

As a possible alternative, we could have e.g. `Cmd.FallibleRun` that returns an `error` - also awkward (brutal name), and would presumably warrant `FallibleOutput`, etc.

---

What do you think? Perhaps we should just keep things as they are and return errors when we `Run` things - the clutter isn't *so* bad. IMO test helpers shouldn't return errors in general, but making sure a `Cmd.Run` actually succeeds often is the assertion of a test, so maybe it's better for the error checking to be explicit

integration/ops/pipelines/smoke-pipeline.yml
```diff
@@ -0,0 +1,36 @@
...
+        type: mock
+        source: {mirror_self: true}
+
+      run:
+        path: echo
+        args: ["Hello, world!"]
+  - put: mockery
+    params: {version: put-version}
+
+- name: use-the-internet
```
aoldershaw, line 24: I think we can remove this job now

hack/overrides/dev.yml
```diff
@@ -0,0 +1,8 @@
+version: '3'
```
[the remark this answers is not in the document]
vito: Woops, this can be deleted - this was part of working around the `build:` stuff I mentioned. Will clean up.

integration/internal/cmdtest/cmd.go
vito: Appending actually will mutate if the slice has capacity for it already. It's a subtle thing that's bitten me in the past.

Here's a demonstration: https://play.golang.org/p/42GnUNPhs8h

This blog post has more details: https://blog.allegro.tech/2017/07/golang-slices-gotcha.html

So I think code like this could potentially break:

```go
cmd := someCmdWithEnvSet
fly1 := cmd.WithEnv("FOO=1")
fly2 = cmd.WithEnv("FOO=2")
```

In this case it's possible for `fly1` to end up with `FOO=2` because the underlying slice was mutated.

integration/internal/cmdtest/cmd.go
vito: Yeah this is one thing I've been considering, since e.g. `WithTempDir` already calls a testing function which fails the test no any `err` instead of returning it. I can give it a try and see how the code looks!

integration/internal/cmdtest/cmd.go
aoldershaw: Yeah fair. I figured something like that would happen, but was surprised to see my playground example succeed, so generalized incorrectly

integration/internal/cmdtest/cmd.go
vito: One problem with `ExpectFailure`: sometimes you don't know if you can expect it.

```go
t.Eventually(func() {
  err := fly.Run("login", ...)
  if err != nil {
    return false
  }

  // ...
}, // ...)
```

In this case your second suggestion would probably be better - maybe something like `Try`, `TryOutput`, etc.?

vito: In combination with something like this:

```go
cmd.ExpectExit(2).Run(t, "foo")
```

This works nicely since I can just have a struct field initialized to `0` as the expected exit code.

aoldershaw: Ah good point. And haha Try is much better name. ExpectExit makes a lot of sense too, though I wonder if asserting on the specific exit code could be brittle?

vito: It's useful in some cases - for example, we expect `fly execute` to exit 2 for an error or 1 for a failure<sup>[1]</sup>. I think it's reasonable for tests to have strict expectations there in case commands start failing for the "wrong reasons" - e.g. what if the test misconfigured a task which it expected to fail? If it isn't precise enough, this could be an incorrectly passing test.

If you don't care about the exit code I guess you could just use `Try` and assert that an error was returned. I think typically you'd assume 1 and if you get something different back you'd verify that it's correct and then update the test.

[1]: https://github.com/concourse/concourse/blob/6fd0fcbe0fcb529e46e709fece52ba3741318edf/fly/eventstream/render.go#L78-L92

vito: Added all this as the last commit - lmk what you think!

aoldershaw (approved):
Having `Cmd.Run` assert on the error makes things really nice. Everything LGTM!

PR #6480, opened 2021-01-27
backport pkger fix to 6.7.x

vito:
`pkged.go` was in `github.com/concourse/concourse`, resulting in strange things happening in `fly` (e.g. shelling out to `go list`, fetching + compiling things, etc.)


aoldershaw (approved):
Don't really get what `pkger` is doing to cause issues but this seems to make senes 🤷

Will be nice to move to Go 1.16, though - eliminate some of this magic

vito:
@aoldershaw It's pretty hairy - by importing `github.com/markbates/pkger`, you run this code:

* https://github.com/markbates/pkger/blob/v0.17.1/pkger.go#L17-L21
* https://github.com/gobuffalo/here/blob/v0.6.0/current.go#L8
* which shells out to `go env GOMOD` 💥

To be honest I'm not really happy with the state of things because it still means that `concourse` shells out to `go env GOMOD` - just not `fly`. It's OK for the command to silently fail, but this still feels like bizarre behavior for operators.

PR #6487, opened 2021-01-29
hide "group:" dropdown suggestion in search

aoldershaw:
since instanced pipelines are still experimental and we need to figure
out the UX for them, it's confusing to suggest the "group:" filter in
the search bar, especially because instanced pipelines are disabled by
default.

with this commit, we hide the "group:" filter from the dropdown so you
don't stumble upon it. note that if you want to apply a "group:" filter,
you still can, and it'll still give suggestions after you've typed
"group:"

### DIFF
--- a/web/elm/src/Dashboard/Filter.elm
+++ b/web/elm/src/Dashboard/Filter.elm
@@ -47,6 +47,11 @@ type alias Filter =
     }
 
 
+filterTypes : List String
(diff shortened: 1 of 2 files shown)


vito (approved):
it is, indeed, hidden - unless i type `group:` and then it autocompletes. neat-o

PR #6489, opened 2021-02-01
Optimize postgres runner for db tests

aoldershaw:

Make tests that require running a local Postgres server faster (~2x faster on my machine)

```
# Before changes
ginkgo -p ./atc/db  560.68s user 1597.73s system 754% cpu 4:46.20 total

# After changes
ginkgo -p ./atc/db  172.06s user 645.53s system 777% cpu 1:45.15 total
```

On CI, it seems to be a *bit* faster, but not as drastic - the DB suite went from ~3 minutes to ~2 minutes

* Create a `TEMPLATE DATABASE` once after migrations were run and replace the `testdb` after each test with a copy of the template
  * `postgresRunner.Truncate` avoids needing to re-run migrations, but surprisingly to me, turned out to be pretty slow - it's a (relatively) constant time operation (regardless of the size of the tables), but for such small tables, it seems to be faster to just `DROP` the whole database and recreate it
  * Using a template also avoids needing to re-run migrations
* Add some minor optimizations based on the fact that durability is not a requirement (see https://www.postgresql.org/docs/13/non-durability.html)

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/integration/integration_suite_test.go
+++ b/atc/integration/integration_suite_test.go
@@ -26,20 +26,11 @@ import (
 var (
 	cmd            *atccmd.RunCommand
 	postgresRunner postgresrunner.Runner
-	dbProcess      ifrit.Process
 	atcProcess     ifrit.Process
 	atcURL         string
 )
 
-var _ = BeforeSuite(func() {
-	postgresRunner = postgresrunner.Runner{
-		Port: 5433 + GinkgoParallelNode(),
-	}
-
-	dbProcess = ifrit.Invoke(postgresRunner)
-
-	postgresRunner.InitializeTestDBTemplate()
-})
+var _ = postgresrunner.GinkgoRunner(&postgresRunner)
 
 var _ = BeforeEach(func() {
 	cmd = &atccmd.RunCommand{}
@@ -104,11 +95,6 @@ var _ = AfterEach(func() {
 	Expect(err).NotTo(HaveOccurred())
 
 	postgresRunner.DropTestDB()
-})
-
-var _ = AfterSuite(func() {
-	dbProcess.Signal(os.Interrupt)
-	<-dbProcess.Wait()
 })
 
 func TestIntegration(t *testing.T) {
--- a/atc/integration/integration_suite_test.go
+++ b/atc/integration/integration_suite_test.go
@@ -31,6 +31,16 @@ var (
 	atcURL         string
 )
 
+var _ = BeforeSuite(func() {
+	postgresRunner = postgresrunner.Runner{
+		Port: 5433 + GinkgoParallelNode(),
+	}
+
+	dbProcess = ifrit.Invoke(postgresRunner)
+
+	postgresRunner.InitializeTestDBTemplate()
+})
+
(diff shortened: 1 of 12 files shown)


aoldershaw:
Interestingly, it seems to make things slightly *slower* on CI - the DB suite took `3m42s` on this PR, whereas on previous PRs it took ~`3m10s`.

I wonder why it's so much faster on my laptop with these changes. Could be a result of some optimizations in `DROP DATABASE` or `CREATE DATABASE ... TEMPLATE ...` between Postgres versions (my laptop is running Postgres 13, while the unit image is using Postgres 11.

aoldershaw:
> I wonder why it's so much faster on my laptop with these changes

I didn't realize we use a tmpfs on CI for postgres - it's possible that with a tmpfs the truncates are actually pretty fast, whereas on a SSD it's faster to DROP+CREATE.

It could also have been a fluke that it was slower - I hijacked into the container and reran the DB suite and it ran in just `2m54s` - not much faster, but at least not slower. I also tried bumping postgres to 13, and it ran in `2m29s` - so maybe we should bump postgres in the unit image (I don't think there's any particular reason to test against Postgres 11 in unit)

aoldershaw:
I upgraded the Postgres version to 13 on the unit image, and the DB suite ran in `2m9s`, about 1 minute faster than the typical duration before bumping Postgres and these optimizations. I ran another PR with Postgres 13 (without the optimizations), and it's still around 3 minutes, so there's at least some improvement on CI!

atc/integration/integration_suite_test.go
```diff
@@ -98,10 +104,11 @@ var _ = AfterEach(func() {
 	Expect(err).NotTo(HaveOccurred())
 
 	postgresRunner.DropTestDB()
+})
 
+var _ = AfterSuite(func() {
 	dbProcess.Signal(os.Interrupt)
-	err = <-dbProcess.Wait()
-	Expect(err).NotTo(HaveOccurred())
+	<-dbProcess.Wait()
```
xtremerui, line 111: Is it a concern if the DB process can't be ended properly? I can see if the test is ran in a container in the pipeline it should be ok. How about running the tests locally? 

It seems some other suits are doing it this way
https://github.com/concourse/concourse/blob/3dc0c519c47f6f8017c81ea7ce97520e3b27c9f8/atc/db/migration/migration_suite_test.go#L39-L41

should them be the same?

aoldershaw: I went ahead and added some helpers to the `postgresrunner` package for common Ginkgo setup/teardown logic (and used the teardown with a timeout, so now all suites running postgres will timeout after 10 seconds). Lemme know what you think

xtremerui (approved):
LGTM!

Issue #6476, opened 2021-01-27
Containerd - Default Runtime


muntac:
Once `containerd` has been released (e.g. v7.0) without the experimental tag, pending feedback, it should be okay to switch to it as the default runtime in the subsequent release (e.g. v7.1).

_Other considerations_

- [ ] Update BOSH release to set appropriate CONTAINERD_INIT_BIN_PATH when no values are set for the runtime property.
- [ ] Update CI to have containerd-integration as a build constraint on build-concourse.

muntac:
Taking the milestone out as we don't necessarily have to make this change in a specific version.

Issue #6475, opened 2021-01-27
[SPIKE] Define what privileged means in context to containerd


muntac:
Figure out what the minimal set of permissions and capabilities for privileged containers when using containerd as a runtime.

vito:
Goals:

1. Folks should still be able to run containers-in-containers (e.g. `docker-compose up`).
2. Make image building faster.
    * Approach A: make it so we can build images without `privileged: true`.
    * Approach B: eliminate performance overhead of using `privileged: true`.
      * Caveat: this approach must not conflict with goal 1.

muntac:
Question to think about from a capabilities standpoint: can we get get away without [using](https://github.com/concourse/concourse/blob/370b34f7524696627c0047da39a71c2d114f9dd6/worker/runtime/spec/capabilities.go#L72) `CAP_SYS_ADMIN` as it is highly permissive?

https://lwn.net/Articles/486306/

muntac:
As @vito pointed out - Kubernetes and Docker **don't have user namespaces on by default**, and do not do any uid and gid mapping.

Docker [allows](https://docs.docker.com/engine/security/userns-remap/) user namespaces but only for non-privileged containers. This means that Docker does not support building of images or docker-in-docker when user namespaces are enabled (see [comment](https://github.com/concourse/concourse/issues/6537#issuecomment-781434714)).

For more on Kubernetes support: 
PR to support user namespaces: https://github.com/kubernetes/enhancements/pull/2101
Reading: https://kinvolk.io/blog/2020/12/improving-kubernetes-and-container-security-with-user-namespaces/#bringing-user-namespaces-to-kubernetes

muntac:
So far for scenarios where we are using user namespaces within a privileged container:
1. Image building works
2. Docker compose does **not** work
See: https://github.com/concourse/concourse/issues/6538#issue-809442105

For any scenario in non-privileged containers image building does **not** work.
See: https://github.com/concourse/concourse/issues/6537#issue-809438194

muntac:
Possible next steps:
1. Figure out how to make docker-compose work in a user namespace. This possibly requires better understanding of how uid/gid mapping is being done in `baggageclaim` but could also involve Kernel level issues.
2. Figure out how to make image building in non-privileged containers work.
- Using current oci-build-task that uses buildkit
- Use a buildkit alternative. Something that allows building images in non-privileged containers.
3. Get performance benefits some other way
[...]

Issue #4358, opened 2019-09-04
Figure out our new minimum requirements with `containerd`


vito:
* What OS distros work out of the box? Maybe we can just try the latest LTS release of various distros?
  * Debian
  * Ubuntu
  * Container-Optimized OS? Others?
* What is our minimum required kernel version?
* Can we invest in CI so we can ensure we don't break the platforms/Kernel versions we want to support?

For example, we know that on older Debian machines/images we saw an error relating to `seccomp_api_set` when trying to run the official `containerd` binary: https://github.com/concourse/concourse/issues/4130#issuecomment-519641631

I saw this on my Linux machine as well, and it was fixed by upgrading all my `apt` packages. Not sure which one - I'm guessing `libseccomp2` or something.

muntac:
@vito and I confirmed our CI smoke tests successfully run with `containerd` on the following:

1. Ubuntu 16.04 - Kernel 4.4 
2. Ubuntu 18.04 - Kernel 5.3
3. Ubuntu 20.04 - Kernel 5.4

Pending results:
4. Debian 9 - Kernel 4.9 
5. Debian 10 - Kernel ?

muntac:
> Can we invest in CI so we can ensure we don't break the platforms/Kernel versions we want to support?

The `bin-smoke-lts` [job](https://github.com/concourse/ci/blob/f596faad329753326cde0e46bcd08e9d78c8a728/pipelines/concourse.yml#L705) now tests against the above stated Ubuntu and Debian distros to account for this.

To Dos:
- [ ] Put updated requirements in https://concourse-ci.org/concourse-worker.html

PR #6492, opened 2021-02-02
Start non-privileged containers in their own cgroup namespace

### LINKED ISSUE #6474: Determine if containers should be started in their own cgroup namespaces
Docker starts containers in their own cgroup namespaces in [v20.10](https://docs.docker.com/engine/release-notes/#20100). Details [here](https://github.com/moby/moby/pull/38377). Figure out the pros and cons of doing the same with containerd.

muntac:
Feature

closes #6474 .

[cgroup namespaces]() give a process its own set of cgroup root directories. This allows for better isolation from the host or parent/ancestor processes.

Support for this feature was [introduced](https://phoronix.com/scan.php?page=news_item&px=CGroup-Namespaces-Linux-4.6) in Linux Kernel 4.6. The minimum Kernel version we support with containerd is 4.4 on Ubuntu 16.04. Fortunately, Ubuntu's own distribution of Kernel 4.4 (used on 16.04 Xenial) had the feature [backported](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1546775) via a patch.

This can be confirmed by running a VM with the latest Ubuntu 16.04 image (e.g. `generic/ubuntu1604` in Vagrant). `cgroup` is listed under namespaces.

```diff
vagrant@ubuntu1604:~$ uname -a
Linux ubuntu1604.localdomain 4.4.0-201-generic #233-Ubuntu SMP Thu Jan 14 06:10:28 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
vagrant@ubuntu1604:~$ ls -l /proc/self/ns
total 0
+ lrwxrwxrwx 1 vagrant vagrant 0 Feb  2 16:26 cgroup -> cgroup:[4026531835]
lrwxrwxrwx 1 vagrant vagrant 0 Feb  2 16:26 ipc -> ipc:[4026531839]
lrwxrwxrwx 1 vagrant vagrant 0 Feb  2 16:26 mnt -> mnt:[4026531840]
lrwxrwxrwx 1 vagrant vagrant 0 Feb  2 16:26 net -> net:[4026531957]
lrwxrwxrwx 1 vagrant vagrant 0 Feb  2 16:26 pid -> pid:[4026531836]
lrwxrwxrwx 1 vagrant vagrant 0 Feb  2 16:26 user -> user:[4026531837]
lrwxrwxrwx 1 vagrant vagrant 0 Feb  2 16:26 uts -> uts:[4026531838]
```

## Notes to reviewer:

We validated the change by checking that a Concourse container has a different cgroup namespace ID from the host i.e. worker. 

Namespaces for the worker the container is on:
```
$ docker exec -it concourse_worker_1 /bin/bash
root@bc53e72d5076:/src# ls -l /proc/self/ns
total 0
lrwxrwxrwx 1 root root 0 Feb  2 17:08 cgroup -> cgroup:[4026531835]
lrwxrwxrwx 1 root root 0 Feb  2 17:08 ipc -> ipc:[4026532822]
lrwxrwxrwx 1 root root 0 Feb  2 17:08 mnt -> mnt:[4026532820]
lrwxrwxrwx 1 root root 0 Feb  2 17:08 net -> net:[4026532825]
lrwxrwxrwx 1 root root 0 Feb  2 17:08 pid -> pid:[4026532823]
lrwxrwxrwx 1 root root 0 Feb  2 17:08 pid_for_children -> pid:[4026532823]
lrwxrwxrwx 1 root root 0 Feb  2 17:08 user -> user:[4026531837]
lrwxrwxrwx 1 root root 0 Feb  2 17:08 uts -> uts:[4026532821]
```

Namespaces inside the container:
```
# ls -l /proc/self/ns
total 0
lrwxrwxrwx    1 root     root             0 Feb  2 17:08 cgroup -> cgroup:[4026533154]
lrwxrwxrwx    1 root     root             0 Feb  2 17:08 ipc -> ipc:[4026533075]
lrwxrwxrwx    1 root     root             0 Feb  2 17:08 mnt -> mnt:[4026533073]
lrwxrwxrwx    1 root     root             0 Feb  2 17:08 net -> net:[4026533078]
lrwxrwxrwx    1 root     root             0 Feb  2 17:08 pid -> pid:[4026533076]
lrwxrwxrwx    1 root     root             0 Feb  2 17:08 pid_for_children -> pid:[4026533076]
lrwxrwxrwx    1 root     root             0 Feb  2 17:08 user -> user:[4026533072]
lrwxrwxrwx    1 root     root             0 Feb  2 17:08 uts -> uts:[4026533074]
```

## Release Note

* Start non-privileged containers in their own cgroup namespace

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/worker/runtime/spec/namespaces.go
+++ b/worker/runtime/spec/namespaces.go
@@ -13,6 +13,7 @@ var (
 
 	UnprivilegedContainerNamespaces = append(PrivilegedContainerNamespaces,
 		specs.LinuxNamespace{Type: specs.UserNamespace},
+		specs.LinuxNamespace{Type: specs.CgroupNamespace},
 	)
 )
 
--- a/worker/runtime/spec/namespaces.go
+++ b/worker/runtime/spec/namespaces.go
@@ -1,6 +1,7 @@
 package spec
 
 import (
+	"errors"
 	"os"
 
 	"github.com/opencontainers/runtime-spec/specs-go"
@@ -37,5 +38,8 @@ func OciNamespaces(privileged bool) []specs.LinuxNamespace {
 
 func cgroupNamespacesSupported() bool {
 	_, err := os.Stat("/proc/self/ns/cgroup")
-	return !os.IsNotExist(err)
+	if err != nil {
+		return !errors.Is(err, os.ErrNotExist)
+	}
+	return true
 }
--- a/worker/runtime/spec/namespaces.go
+++ b/worker/runtime/spec/namespaces.go
@@ -21,7 +21,10 @@ var (
 )
 
 func init() {
-	cgroupNamespacesEnabled()
+	if cgroupNamespacesSupported() {
+		UnprivilegedContainerNamespaces = append(UnprivilegedContainerNamespaces,
+			specs.LinuxNamespace{Type: specs.CgroupNamespace})
+	}
 }
 
 func OciNamespaces(privileged bool) []specs.LinuxNamespace {
@@ -32,11 +35,7 @@ func OciNamespaces(privileged bool) []specs.LinuxNamespace {
 	return PrivilegedContainerNamespaces
 }
 
-func cgroupNamespacesEnabled() {
-	if _, err := os.Stat("/proc/self/ns/cgroup"); os.IsNotExist(err) {
-		return
-	}
-
-	UnprivilegedContainerNamespaces = append(UnprivilegedContainerNamespaces,
-		specs.LinuxNamespace{Type: specs.CgroupNamespace})
+func cgroupNamespacesSupported() bool {
+	_, err := os.Stat("/proc/self/ns/cgroup")
+	return !os.IsNotExist(err)
 }
--- a/worker/runtime/spec/namespaces.go
+++ b/worker/runtime/spec/namespaces.go
@@ -1,6 +1,10 @@
 package spec
 
-import "github.com/opencontainers/runtime-spec/specs-go"
+import (
+	"os"
+
+	"github.com/opencontainers/runtime-spec/specs-go"
+)
 
 var (
 	PrivilegedContainerNamespaces = []specs.LinuxNamespace{
@@ -13,9 +17,12 @@ var (
 
 	UnprivilegedContainerNamespaces = append(PrivilegedContainerNamespaces,
 		specs.LinuxNamespace{Type: specs.UserNamespace},
-		specs.LinuxNamespace{Type: specs.CgroupNamespace},
 	)
 )
+
+func init() {
+	cgroupNamespacesEnabled()
+}
 
 func OciNamespaces(privileged bool) []specs.LinuxNamespace {
 	if !privileged {
@@ -23,4 +30,13 @@ func OciNamespaces(privileged bool) []specs.LinuxNamespace {
 	}
 
 	return PrivilegedContainerNamespaces
+}
+
+func cgroupNamespacesEnabled() {
+	if _, err := os.Stat("/proc/self/ns/cgroup"); os.IsNotExist(err) {
+		return
+	}
+
+	UnprivilegedContainerNamespaces = append(UnprivilegedContainerNamespaces,
+		specs.LinuxNamespace{Type: specs.CgroupNamespace})
 }


aoldershaw:
> The minimum Kernel version we support with containerd is 4.4 on Ubuntu 16.04

Just curious, what is this support based on? Do we intend to support other distros? And do those distros all have cgroup namespaces?

Is it possible/easy enough to check for cgroup ns support and only conditionally apply it?

taylorsilva:
> Just curious, what is this support based on? Do we intend to support other distros? And do those distros all have cgroup namespaces?

Some details here: https://github.com/concourse/concourse/issues/4358
Seems like we decided to see what was feasible for us to test and added support for that. We could more generally say we support any kernel with features x,y,z. That would probably be a more honest description of our support. Figuring out which distro's have back-ported what features is not easy. Muntac spent an entire afternoon trying to find the patch notes for where cgroup namespaces got backported to 4.4 in the ubuntu kernel. Doing this check for every feature across various distros would be... long and painful.

> Is it possible/easy enough to check for cgroup ns support and only conditionally apply it?

We weren't initially sure but it looks like you can check for the existence of `/proc/self/ns/cgroup` to see if you have cgroup namespaces. That's how docker does it: https://github.com/moby/moby/pull/38377
We didn't do that check since all the distro's we support have cgroup namespaces. I'm not against making cgroup namespaces optional though.

aoldershaw:
> Seems like we decided to see what was feasible for us to test and added support for that...Figuring out which distro's have back-ported what features is not easy

That's definitely true - trying to explicitly support a whole slew of distros/kernel versions would definitely cause us a lot of pain in the long run. But in this case, even if we don't advertise support for (or test compatibility with) such distro/kernel pairs, it seems like a pretty low-effort patch that at least won't prevent people from using Concourse (especially if we start defaulting to containerd #6476, and eventually remove `gdn` altogether). That said, I'm honestly not too familiar with the kernel landscape - it's possible that this is wholly unnecessary and that most distros are using kernel versions that have this feature. You and @muntac definitely have a better picture of this than me, so feel free to tell me it's not worth it

aoldershaw:
@muntac @taylorsilva just following up - any thoughts about https://github.com/concourse/concourse/pull/6492#issuecomment-776027424? Is it worth checking for `/proc/self/ns/cgroup` to not break older kernels, while still not explicitly supporting them?

muntac:
@aoldershaw sorry we never got back to you on this. My instinct is you're right, and it should be a small enough check. We'll make the update once we wrap up https://github.com/concourse/concourse/issues/6475.

taylorsilva:
@aoldershaw updated to only be enabled if supported by the kernel

worker/runtime/spec/namespaces.go
```diff
@@ -13,14 +17,26 @@ var (
 
 	UnprivilegedContainerNamespaces = append(PrivilegedContainerNamespaces,
 		specs.LinuxNamespace{Type: specs.UserNamespace},
-		specs.LinuxNamespace{Type: specs.CgroupNamespace},
 	)
 )
 
+func init() {
+	cgroupNamespacesEnabled()
```
aoldershaw, line 24: Nit: would it be possible to rename this to something more verbose like `enableCgroupNamespacesIfSupported`? `cgroupNamespacesEnabled` doesn't suggest to me that there is mutation, and `init` already makes things a bit cryptic, so making it clear there's a mutation is probably a good thing.

Alternatively, it could be something like:
```go
func init() {
	if cgroupNamespacesSupported() {
		UnprivilegedContainerNamespaces = append(UnprivilegedContainerNamespaces,
			specs.LinuxNamespace{Type: specs.CgroupNamespace})
	}
}
```

which is even more clear IMO

aoldershaw (changes requested):

worker/runtime/spec/namespaces.go
```diff
@@ -32,11 +35,7 @@ func OciNamespaces(privileged bool) []specs.LinuxNamespace {
...
-func cgroupNamespacesEnabled() {
-	if _, err := os.Stat("/proc/self/ns/cgroup"); os.IsNotExist(err) {
-		return
-	}
-
-	UnprivilegedContainerNamespaces = append(UnprivilegedContainerNamespaces,
-		specs.LinuxNamespace{Type: specs.CgroupNamespace})
+func cgroupNamespacesSupported() bool {
+	_, err := os.Stat("/proc/self/ns/cgroup")
+	return !os.IsNotExist(err)
```
aoldershaw, line 40: Actually, I think this logic is a bit off - if an error occurs that is not `os.ErrNotExist`, it'll return true. I think it should be something like:

```go
_, err := os.Stat("/proc/self/ns/cgroup")
if err != nil {
	return !os.IsNotExist(err)
}
return true
```

Also, looking at the [godoc](https://golang.org/pkg/os/#IsNotExist):

> This function predates errors.Is. It only supports errors returned by the os package. New code should use errors.Is(err, os.ErrNotExist).

So, I guess the preferred form is `errors.Is(err, os.ErrNotExists)`

aoldershaw (approved):
Thanks! LGTM

PR #6495, opened 2021-02-03
feat(atc): cache streamed volumes and try to get by looking for local cache

### LINKED ISSUE #6218: task image get and task are on different workers
## Summary

I found this problem in same test as #6217. See the screen shot:

task-get-different-worker

We can see that, task image get was done on a different worker than the worker where task ran.

I used "random" container placement strategy. But image fetching should be done on the worker where the task is going to run, right? @vito 

## Triaging info

* Concourse version:
* Browser (if applicable):
* Did this used to work?

evanchaoli:

closes #6218 .

This is a performance optimization PR. It will avoid `get` same version duplicately, which is not only reduce resource consumption  of Concourse itself, but also reduce Concourse counterpart systems' workloads. E.g. if many jobs need to get a common git repo, then only the first job should do `git clone`, rest of jobs will just get the `git` resource version from local workers, which reduces `git clone`s Concourse bringing to the Git system.

I think this could be an advantage that Concourse wins competitors.

I expect this feature plus the p2p stream feature should significantly improve large Concourse cluster's performance.

EDIT: more details is at https://github.com/concourse/concourse/pull/6495#issuecomment-791080524

1. Mark streamed volumes as resource cache.
2. `get` step will try to find resource cache first, if found, then don't run `get` container any more.
3. Change the way how remote inputs are streamed-in. Originally, when an input is at remote, it will create a volume then stream-in remote volume's contents into the new volume. In that way, the new volume may be written by a task. This PR changed the way to: stream-in remote volume and create a COW volume on it, which make local and remote inputs are handled in same way. @vito, this is a big change, so I want your comment as well.

I also studied impacts of this feature to volume-GC. Looks like there is no impact. Because volume GC relies on container GC and resource cache GC. When a container is destroyed, corresponding volumes' `container_id` will be set to NULL, so to resource cache, then once a volumes all xxxx_ids are NULL, it will be GC-ed. As this feature doesn't touch container and resource cache lifecycle at all, volume-GC should not be impacted either.

## Notes to reviewer:

Please help see if there are any potential issue with this solution.

I tested with the following pipeline:

```yaml
resources:
- name: mygit
  type: git
  source:
    uri: <a git repo>
    branch: master
    private_key: ((chaol-gitlab-private-key))

jobs:
- name: test
  public: true
  plan:
  - get: mygit
  - in_parallel:
    - task: t1
      config:
        platform: linux
        image_resource:
          type: registry-image
          source: {repository: busybox}
        inputs:
        - name: mygit
        outputs:
        - name: mygit
        run:
          path: sh
          args:
           - -exc
           - |
             echo ok
             sleep 5
             ls mygit/
             date > mygit/test.1.`date +%m-%d-%H-%M-%S`
    - task: t2
      config:
        platform: linux
        image_resource:
          type: registry-image
          source: {repository: busybox}
        inputs:
        - name: mygit
        outputs:
        - name: mygit
        run:
          path: sh
          args:
           - -exc
           - |
             echo ok
             sleep 5
             ls mygit/
             date > mygit/test.2.`date +%m-%d-%H-%M-%S`
    - task: t3
      config:
        platform: linux
        image_resource:
          type: registry-image
          source: {repository: busybox}
        inputs:
        - name: mygit
        outputs:
        - name: mygit
        run:
          path: sh
          args:
           - -exc
           - |
             echo ok
             sleep 5
             ls mygit/
             date > mygit/test.3.`date +%m-%d-%H-%M-%S`
    - task: t4
      config:
        platform: linux
        image_resource:
          type: registry-image
          source: {repository: busybox}
        inputs:
        - name: mygit
        outputs:
        - name: mygit
        run:
          path: sh
          args:
           - -exc
           - |
             echo ok
             sleep 5
             ls mygit/
             date > mygit/test.4.`date +%m-%d-%H-%M-%S`
  - task: show-result
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      inputs:
      - name: mygit
      run:
        path: sh
        args:
         - -exc
         - |
           ls mygit/
```

In the first run, `mygit` did `git clone`:

first-run

Then in the second run, `git mygit` did nothing but found the resource cache:

second-run

## Release Note

* 

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/resource_cache.go
+++ b/atc/db/resource_cache.go
@@ -177,6 +177,7 @@ func paramsHash(p atc.Params) string {
 type UsedResourceCache interface {
 	ID() int
 	Version() atc.Version
+	LoadVersionMetadata() ([]atc.MetadataField, error)
 
 	ResourceConfig() ResourceConfig
 
@@ -226,6 +227,26 @@ func (cache *usedResourceCache) BaseResourceType() *UsedBaseResourceType {
 	}
 
 	return cache.resourceConfig.CreatedByResourceCache().BaseResourceType()
+}
+
+func (cache *usedResourceCache) LoadVersionMetadata() ([]atc.MetadataField, error) {
+	r := &resourceConfigVersion{conn: cache.conn}
+
+	versionJSON, _ := json.Marshal(cache.Version())
+	row := resourceConfigVersionQuery.
+		Where(sq.Expr("v.version_md5 = md5(?)", versionJSON)).
+		RunWith(r.conn).
+		QueryRow()
+
+	err := scanResourceConfigVersion(r, row)
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	return r.metadata.ToATCMetadata(), nil
 }
 
 func mapHash(m map[string]interface{}) string {
--- a/atc/db/resource_cache.go
+++ b/atc/db/resource_cache.go
@@ -234,7 +234,9 @@ func (cache *usedResourceCache) LoadVersionMetadata() ([]atc.MetadataField, erro
 
 	versionJSON, _ := json.Marshal(cache.Version())
 	row := resourceConfigVersionQuery.
+		LeftJoin("resource_config_scopes rcs ON v.resource_config_scope_id = rcs.id").
 		Where(sq.Expr("v.version_md5 = md5(?)", versionJSON)).
+		Where(sq.Eq{"rcs.resource_config_id": cache.resourceConfig.ID()}).
 		RunWith(r.conn).
 		QueryRow()
 
--- a/atc/db/resource_cache.go
+++ b/atc/db/resource_cache.go
@@ -177,7 +177,6 @@ func paramsHash(p atc.Params) string {
 type UsedResourceCache interface {
 	ID() int
 	Version() atc.Version
-	LoadVersionMetadata() ([]atc.MetadataField, error)
 
 	ResourceConfig() ResourceConfig
 
@@ -227,28 +226,6 @@ func (cache *usedResourceCache) BaseResourceType() *UsedBaseResourceType {
 	}
 
 	return cache.resourceConfig.CreatedByResourceCache().BaseResourceType()
-}
-
-func (cache *usedResourceCache) LoadVersionMetadata() ([]atc.MetadataField, error) {
-	r := &resourceConfigVersion{conn: cache.conn}
-
-	versionJSON, _ := json.Marshal(cache.Version())
-	row := resourceConfigVersionQuery.
-		LeftJoin("resource_config_scopes rcs ON v.resource_config_scope_id = rcs.id").
-		Where(sq.Expr("v.version_md5 = md5(?)", versionJSON)).
-		Where(sq.Eq{"rcs.resource_config_id": cache.resourceConfig.ID()}).
-		RunWith(r.conn).
-		QueryRow()
-
-	err := scanResourceConfigVersion(r, row)
-	if err != nil {
-		if err == sql.ErrNoRows {
-			return nil, nil
-		}
-		return nil, err
-	}
-
-	return r.metadata.ToATCMetadata(), nil
 }
 
 func mapHash(m map[string]interface{}) string {
--- a/atc/db/team.go
+++ b/atc/db/team.go
@@ -173,6 +173,13 @@ func (t *team) FindWorkersForResourceCache(rcId int) ([]Worker, error) {
 			Where(sq.And{
 				sq.Eq{"rc.id": rcId},
 				sq.Eq{"w.state": "running"},
+			}).
+			Where(sq.Or{
+				sq.Eq{
+					"w.team_id": t.id,
+				}, sq.Eq{
+					"w.team_id": nil,
+				},
 			}))
 }
 
--- a/atc/db/team.go
+++ b/atc/db/team.go
@@ -172,14 +172,7 @@ func (t *team) FindWorkersForResourceCache(rcId int) ([]Worker, error) {
 			Join("resource_caches rc ON rc.id = wrc.resource_cache_id").
 			Where(sq.And{
 				sq.Eq{"rc.id": rcId},
-				sq.Eq{"w.state": "running"},
-			}).
-			Where(sq.Or{
-				sq.Eq{
-					"w.team_id": t.id,
-				}, sq.Eq{
-					"w.team_id": nil,
-				},
+				sq.Eq{"w.state": WorkerStateRunning},
 			}))
 }
 
--- a/atc/db/team.go
+++ b/atc/db/team.go
@@ -75,6 +75,7 @@ type Team interface {
 	FindCreatedContainerByHandle(string) (CreatedContainer, bool, error)
 	FindWorkerForContainer(handle string) (Worker, bool, error)
 	FindWorkerForVolume(handle string) (Worker, bool, error)
+	FindWorkersForResourceCache(rcId int) ([]Worker, error)
 
 	UpdateProviderAuth(auth atc.TeamAuth) error
 }
@@ -161,6 +162,18 @@ func (t *team) FindWorkerForVolume(handle string) (Worker, bool, error) {
 	return getWorker(t.conn, workersQuery.Join("volumes v ON v.worker_name = w.name").Where(sq.And{
 		sq.Eq{"v.handle": handle},
 	}))
+}
+
+func (t *team) FindWorkersForResourceCache(rcId int) ([]Worker, error) {
+	return getWorkers(
+		t.conn, workersQuery.
+			Join("volumes v ON v.worker_name = w.name").
+			Join("worker_resource_caches wrc ON v.worker_resource_cache_id = wrc.id").
+			Join("resource_caches rc ON rc.id = wrc.resource_cache_id").
+			Where(sq.And{
+				sq.Eq{"rc.id": rcId},
+				sq.Eq{"w.state": "running"},
+			}))
 }
 
 func (t *team) Containers() ([]Container, error) {
--- a/atc/exec/get_step.go
+++ b/atc/exec/get_step.go
@@ -194,7 +194,7 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
 
 	getResult, found, err := step.getFromLocalCache(ctx, logger, step.metadata.TeamID, resourceCache, workerSpec)
 	if found {
-		fmt.Fprintln(delegate.Stderr(), "\x1b[1;33mINFO: found resouce cache from local cache\x1b[0m")
+		fmt.Fprintln(delegate.Stderr(), "\x1b[1;36mINFO: found resouce cache from local cache\x1b[0m")
 		fmt.Fprintln(delegate.Stderr(), "")
 
 		state.StoreResult(step.planID, resourceCache)
--- a/atc/exec/get_step.go
+++ b/atc/exec/get_step.go
@@ -207,6 +207,10 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
 			getResult.GetArtifact,
 		)
 
+		if step.plan.Resource != "" {
+			delegate.UpdateVersion(logger, step.plan, getResult.VersionResult)
+		}
+
 		delegate.Finished(
 			logger,
 			ExitStatus(getResult.ExitStatus),
--- a/atc/exec/get_step.go
+++ b/atc/exec/get_step.go
@@ -192,7 +192,10 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
 		return false, err
 	}
 
-	getResult, found, err := step.getFromLocalCache(ctx, logger, step.metadata.TeamID, resourceCache, workerSpec)
+	getResult, found, err := step.getFromLocalCache(logger, step.metadata.TeamID, resourceCache, workerSpec)
+	if err != nil {
+		return false, err
+	}
 	if found {
 		fmt.Fprintln(delegate.Stderr(), "\x1b[1;36mINFO: found resouce cache from local cache\x1b[0m")
 		fmt.Fprintln(delegate.Stderr(), "")
@@ -294,17 +297,13 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
 }
 
 func (step *GetStep) getFromLocalCache(
-	ctx context.Context,
 	logger lager.Logger,
 	teamId int,
 	resourceCache db.UsedResourceCache,
 	workerSpec worker.WorkerSpec) (worker.GetResult, bool, error) {
-	volume, found, err := step.findResourceCache(logger, teamId, resourceCache, workerSpec)
-	if err != nil {
-		return worker.GetResult{}, false, err
-	}
+	volume, found := step.findResourceCache(logger, teamId, resourceCache, workerSpec)
 	if !found {
-		return worker.GetResult{}, false, err
+		return worker.GetResult{}, false, nil
 	}
 	metadata, err := resourceCache.LoadVersionMetadata()
 	if err != nil {
@@ -324,23 +323,23 @@ func (step *GetStep) findResourceCache(
 	logger lager.Logger,
 	teamId int,
 	resourceCache db.UsedResourceCache,
-	workerSpec worker.WorkerSpec) (worker.Volume, bool, error) {
+	workerSpec worker.WorkerSpec) (worker.Volume, bool) {
 	workers, err := step.workerPool.FindWorkersForResourceCache(logger, teamId, resourceCache.ID(), workerSpec)
 	if err != nil {
-		return nil, false, err
+		return nil, false
 	}
 
 	for _, sourceWorker := range workers {
 		volume, found, err := sourceWorker.FindVolumeForResourceCache(logger, resourceCache)
 		if err != nil {
-			logger.Error("ignore-error", err)
+			logger.Debug("ignore-error", lager.Data{"error": err})
 			continue
 		}
 		if !found {
 			continue
 		}
-		return volume, true, nil
+		return volume, true
 	}
 
-	return nil, false, nil
+	return nil, false
 }
--- a/atc/exec/get_step.go
+++ b/atc/exec/get_step.go
@@ -192,7 +192,7 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
 		return false, err
 	}
 
-	getResult, found, err := step.getFromLocalCache(ctx, logger, step.metadata.TeamID, resourceCache)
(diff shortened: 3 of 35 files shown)


clarafu:
@evanchaoli Sorry reviewing this PR is taking so long, I am new to the `worker` package and how all the code is intertwined so it is taking a bit of time to understand it! Just wanted to let you know that I am looking at it and it's not being ignored!

atc/db/team.go
```diff
@@ -163,6 +164,18 @@ func (t *team) FindWorkerForVolume(handle string) (Worker, bool, error) {
 	}))
 }
 
+func (t *team) FindWorkersForResourceCache(rcId int) ([]Worker, error) {
+	return getWorkers(
+		t.conn, workersQuery.
+			Join("volumes v ON v.worker_name = w.name").
+			Join("worker_resource_caches wrc ON v.worker_resource_cache_id = wrc.id").
+			Join("resource_caches rc ON rc.id = wrc.resource_cache_id").
+			Where(sq.And{
```
clarafu, line 173: We probably want to only fetch from a worker that the team has access to. So if there are team workers that are not part of this team or tagged workers that have the resource cache on them, we don't want to use that worker.

clarafu (changes requested):
This feature would be really cool! I also noticed a lot of test failures and I'm assuming that is just because you haven't gotten to fixing the tests with these changes?

atc/db/resource_cache.go
```diff
@@ -241,6 +242,26 @@ func (cache *usedResourceCache) BaseResourceType() *UsedBaseResourceType {
 	return cache.resourceConfig.CreatedByResourceCache().BaseResourceType()
 }
 
+func (cache *usedResourceCache) LoadVersionMetadata() ([]atc.MetadataField, error) {
+	r := &resourceConfigVersion{conn: cache.conn}
+
+	versionJSON, _ := json.Marshal(cache.Version())
+	row := resourceConfigVersionQuery.
+		Where(sq.Expr("v.version_md5 = md5(?)", versionJSON)).
```
clarafu, line 250: You would also need to filter on `resource_config_scope_id` because there can be multiple versions that have the same version strings but belong to different resource config scopes (which can result in different metadata)

evanchaoli:
> @evanchaoli Sorry reviewing this PR is taking so long, I am new to the `worker` package and how all the code is intertwined so it is taking a bit of time to understand it! Just wanted to let you know that I am looking at it and it's not being ignored!

No worries. Actually I just finished 2-week vacation for the Chinese New Year holiday, and back to work today. I believe this PR still need many more iterations to go.

atc/db/team.go
evanchaoli: I guess not really. I think "use that worker" means "run containers on the worker". But resource cache is completely distinguished by `source`, where `source` may include URL and credential to access the resource, it's very possible that multiple teams may share a common resource, like a docker image. So I feel `team` and `tag` don't really play significant role with resource cache.

atc/db/resource_cache.go
evanchaoli: Good point. Will add it.

evanchaoli: @clarafu Please take a look again.

atc/db/team.go
clarafu: Right, but I think we still don't want to use a team worker unless the get step was run from that team or a tagged worker if the step itself is tagged for that worker. For example, there can be the case where a team worker has access to a private network that other team do not have access to. They would expect that only their team would have access to that private network but if we shared volumes between team workers then that would not hold true. The resource cache being distinguished by source and type does not completely guarantee that it will always be unique (ex. a repo in two private networks can have contain the same items).

evanchaoli: That makes sense, though I would not be true in our deployments. I'll update the SQL.

evanchaoli:
> I also noticed a lot of test failures and I'm assuming that is just because you haven't gotten to fixing the tests with these changes?

@clarafu  Yes, I haven't fixed the tests yet. Basically this PR changed Concourse behaviors like:

1.`get` will no nothing if a local cached volume can be found
2. The way how remote volumes streamed in when creating a container is changed.

With those changes, fixing tests is a big effort. So I want Concourse team to review these changes from design perspective first. Once we aligned with the direction, I'll spend time fixing tests and adding new tests.

atc/db/team.go
evanchaoli: @clarafu fixed, please take a look.

evanchaoli:
@clarafu So far, this PR is almost ready. Missing part is to add some tests for the new code. I'll add tests in next few days.

The changes in `docker-compose.yml` as well as those logs marked with "EVAN" will be cleaned up before merge. Please just ignore them when you review the code.

vito:
Just chiming in to respond to this:

> Change the way how remote inputs are streamed-in. Originally, when an input is at remote, it will create a volume then stream-in remote volume's contents into the new volume. In that way, the new volume may be written by a task. This PR changed the way to: stream-in remote volume and create a COW volume on it, which make local and remote inputs are handled in same way. @vito, this is a big change, so I want your comment as well.

Makes sense! We technically need to initialize a pristine cache and then attach a copy of it to the task so it can't modify it. As long as the task's copy-on-write volume expires with the build, leaving only the pristine cache around, this sounds correct. 👍

atc/db/team.go
clarafu: I think we also don't want to use a cache on a tagged worker if the get step does not have a matching tag!

atc/db/team.go
```diff
@@ -163,6 +164,25 @@ func (t *team) FindWorkerForVolume(handle string) (Worker, bool, error) {
...
+				sq.Eq{"rc.id": rcId},
+				sq.Eq{"w.state": "running"},
```
clarafu, line 175: nit: It might be nice to use the `WorkerStateRunning` const within this query https://github.com/concourse/concourse/blob/0e8560d9b2ca285870a752a43fa0954642ac7297/atc/db/worker.go#L32

atc/db/team.go
```diff
@@ -163,6 +164,25 @@ func (t *team) FindWorkerForVolume(handle string) (Worker, bool, error) {
...
+			}).
+			Where(sq.Or{
+				sq.Eq{
```
clarafu, line 178: I think it might be a little more complicated than this. So I'm pretty sure that currently, if a team has a team worker then we will only ever use the team worker. This part of the code is where we will return only the team workers if there are any compatible team workers

https://github.com/concourse/concourse/blob/81c39e21d445069ebb95d8d1f63c9ed63eaf1a36/atc/worker/pool.go#L88-L92

With that in mind, I think we need to check to see if this team has any team workers and if it does, then we pull off the workers that have a resource cache that matches. If the team does not have any team workers, then we can grab a resource cache off any general workers (workers that are not part of a team and not tagged).

clarafu (changes requested):

atc/db/team.go
evanchaoli: Now I moved the team and tags matching to `atc/worker/pool.go`, so that it reuse same logic of `pool.allSatisfying()` for matching team and tags.

evanchaoli:
@clarafu I have added tests. So this PR is almost ready expect cleanup those private logs (marked with "EVAN"). I plan to run more tests, those logs will be useful for my tests. I'll clean up code once I finish the test. I hope this PR be included in 7.1.0.

atc/exec/get_step.go
```diff
@@ -191,6 +192,29 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
...
+			build.ArtifactName(step.plan.Name),
+			getResult.GetArtifact,
+		)
+
+		delegate.Finished(
+			logger,
+			ExitStatus(getResult.ExitStatus),
+			getResult.VersionResult,
+		)
+
```
clarafu, line 212: Do we want to update the version metadata like we do below?

```
		if step.plan.Resource != "" {
			delegate.UpdateVersion(logger, step.plan, getResult.VersionResult)
		}
```

I think this basically updates the metadata on the version that belongs to a `resource_config_scope_id`. There can be multiple `resource_config_scope_id`s that point to a single `resource_config_id`, which we probably want to update the metadata for the `resource_config_scope_id` that this resource points to even if it used a resource cache from the local cache.

atc/worker/artifact_source.go
```diff
@@ -149,12 +164,33 @@ func (source *artifactSource) StreamTo(
...
+
 	// Inc counter if no error occurred.
-	if err == nil {
-		metric.Metrics.VolumesStreamed.Inc()
+	metric.Metrics.VolumesStreamed.Inc()
+
+	usedResourceCache, found, err := source.resourceCacheFactory.FindResourceCacheByID(source.volume.GetResourceCacheID())
+	if err != nil {
+		logger.Error("artifactSource-StreamTo-failed-to-find-resource-cache", err)
+		return nil
```
clarafu, line 177: Is there a reason why we are returning nil instead of the err here? Is it because you are thinking that it is okay for this method to pass even if we don't initialize the resource cache? If that is your reason, I'm also wondering the same thing but the only thing that makes me feel like it should return an error is that this is not supposed to happen. Like there is something that is going wrong with the database if this was to happen so maybe it would make sense to just fail early rather than covering up the error.

atc/worker/artifact_source.go
```diff
@@ -149,12 +164,33 @@ func (source *artifactSource) StreamTo(
...
 	}
+	if !found {
+		logger.Info("artifactSource.StreamTo-not-find-resource-cache, this should not happen",
+			lager.Data{"rcId": source.volume.GetResourceCacheID(), "volumeHandle": source.volume.Handle()})
+		return nil
```
clarafu, line 182: If this should not happen, we probably want to return a constructed error like `ErrStreamingResourceCacheNotFound`. Same reasoning as my comment above, where if this is not supposed to happen then there is something that went wrong either in the ATCs or the database.

clarafu (changes requested):

atc/exec/get_step.go
```diff
@@ -191,6 +192,29 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
 		return false, err
 	}
 
+	getResult, found, err := step.getFromLocalCache(ctx, logger, step.metadata.TeamID, resourceCache, workerSpec)
```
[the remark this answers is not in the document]
evanchaoli: If fails to find the resource from local cache, we just want to ignore the error and go ahead with the normal `get` step. To address this comment, I added a comment line, and added a debug log to record the error for debugging purpose.

atc/worker/artifact_source.go
evanchaoli: This piece of code is trying to mark a streamed volume as a resource cache if source volume is a resource cache. If an error occurs, we just log the error and discard the error, so that it fall back to normal process.

atc/exec/get_step.go
evanchaoli: Make sense. Fixed.

clarafu (commented):
I'm doing acceptance and I noticed one thing about the `get` steps that were using a local cache. The get step in the UI seemed to be greyed out and I'm not familiar with what could be causing this? 

Screen Shot 2021-03-04 at 4 31 09 PM

I'm using a deployment with three workers and this pipeline

```
---
resources:
- name: some-resource
  type: git
  source:
    uri: https://github.com/concourse/concourse.git

jobs:
- name: task
  plan:
  - get: some-resource
  - task: task-1
    config:
      platform: linux
      image_resource:
        type: registry-image
      inputs:
      - name: some-resource
      run:
        path: echo
        args: ["Hello, world!"]
```

evanchaoli:
@clarafu The screen shot tells that your GUI is very old, even doesn't contain 7.0.0 changes. You need to run `yarn build` first, then `docker-compose build`.

evanchaoli:
@clarafu Let me rephrase what is motivation of the PR, so you may notice the problems during your acceptance test.

First, let me describe what 7.0.0 behaves. In 7.0.0, @vito changed the image fetch process, `task/put` steps (`get` is the same if it doesn't use a builtin resource type) embeds `check` and `get` steps to fetch a docker image that the `task`/`put` needs. I'll just use a task as example:

build-1

`get: image` selected worker-1, so it fetched the docker image to worker-1.
And the task itself selected worker-2, so Concourse will stream the image volume from worker-1 to worker-2.

In pre-7.x, if a task selects a worker, the image will just be fetched to the worker. So we can see that, 7.0.0 not only fetches the image but also streaming the image, which is a little bit performance burden.

build-2

In build-2, `get: image` selected worker-3, so it fetched the image again to worker-3.
And the task also selected worker-2. Ideally, given worker-2 already ran build-1, it has the image already, there is no need to streaming the image from worker-3 once again. But as streamed volume is not marked as a resource cache, 7.0.0 will streamed the image volume from worker-3 to worker-2.

build-3

In build 3, `get: image` again selected worker-3. As worker-3 already fetched the image in build 2, it will not fetch the image again.
And the task itself again selected worker-2. Unfortunately, 7.0.0 will streaming the image again the image volume from worker-3 to worker-2.

build-4

In build 4, `get: image` selected worker-2. Though worker-2 has been streamed the image volume multiple time, it will still fetch the image again.
The task still selected worker-2, and the image has been fetched on worker-2, it will no longer need to streaming the image volume.

With these 4 builds, Concourse will fetch the same image 3 times, and streaming the image volume 3 times. So we can see, 7.0.0 will take more efforts to warm up docker images over all workers, which is a performance decrease.

The initial motivation of this PR is to solve the performance decrease. With this PR, let's see the behavior changes for the 4 builds.

Build 1 will behave the same, the image will be fetched to worker-1, and streamed to worker-2. But the change is, the image volume will be marked as a resource cache on worker-2.

Build 2: `get: image` will find the image on worker-1 and worker 2, so the get will do nothing. And as worker-2 already contains the image, it will not stream the image again.

Build 3: `get: image` will again find the image on worker-1 and worker-2, so the get will do nothing. And as worker-2 already contains the image, it will not stream the image again.

Build 4: As worker-2 has the image already, there will be neither image fetching nor streaming.

So with this PR, the 4 build will do only 1 image fetching and 1 image volume streaming.

This feature not only optimize Concourse performance, it will also reduce external communications, like image pulling, git clones:

* Say a cluster has N workers, without this feature, an docker image may be pulled N times in order to warm up all workers. With this feature, the image will be pulled only once, and streamed N-1 time to get all workers warmed up.
* Say a common git repo that is `get` by a lot of pipelines, then each build of those pipelines will do `git clone`. With this feature, only 1 `git clone` needs to be done, then Concourse will use local cache for later builds.

As `get` will do nothing if a resource cache can be found, which will save a lot of containers, so that optimizes worker's performance.

evanchaoli:
@clarafu With the above explanation, for performance consideration, my plan is to skip 7.0.0, and use 7.x.0 (hopefully x is 1) where this feature is included.

clarafu:
@evanchaoli oh silly me, you're totally right I did a yarn build in the wrong repo when I was running your changes

clarafu:
Just an idea that is not necessary, it might be nice to show which worker the local resource cache the get step grabbed from? So it would also say `selected worker: <worker-with-cache>` in the build logs. 

Screen Shot 2021-03-08 at 1 02 03 PM

evanchaoli:
> Just an idea that is not necessary, it might be nice to show which worker the local resource cache the get step grabbed from? So it would also say `selected worker: <worker-with-cache>` in the build logs.

Notice that, `get` step actually doesn't grab a resource at all as long as it find a local cache on some worker. There are may be multiple workers containing the local cache, the following step may grab a cache resource from any worker. Thus I don't think it makes much sense to show a worker name from which the resource can be grabbed.

evanchaoli:
@clarafu I have cleaned up the code, removed my private logs and reverted changes in `docker-compose.yml`. Unless you come out some new comments, I think this PR is ready to merge.

atc/exec/get_step.go
```diff
@@ -268,3 +299,51 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
...
+func (step *GetStep) getFromLocalCache(
+	logger lager.Logger,
+	teamId int,
+	resourceCache db.UsedResourceCache,
+	workerSpec worker.WorkerSpec) (worker.GetResult, bool, error) {
+	volume, found := step.findResourceCache(logger, teamId, resourceCache, workerSpec)
+	if !found {
+		return worker.GetResult{}, false, nil
+	}
+	metadata, err := resourceCache.LoadVersionMetadata()
```
clarafu, line 312: @aoldershaw pointed out that we can actually use an existing method `ResourceCacheMetadata` rather than adding a new method here! https://github.com/concourse/concourse/blob/ff1d2fae5eb0e657987ec3af3f1162f5d63db698/atc/db/resource_cache_factory.go#L103

clarafu (changes requested):

atc/exec/get_step.go
evanchaoli: Fixed.

clarafu (approved):
👍  👍

chenbh:
@evanchaoli @clarafu 

I'm not sure the exact reason why, but [`WorkerResourceCaches`](https://github.com/concourse/concourse/blob/35e971a307d349be791178c5a3d17f24c966c8c1/atc/db/worker_resource_cache.go#L21-L26) must be associated to a [`WorkerBaseResourceType`](https://github.com/concourse/concourse/blob/35e971a307d349be791178c5a3d17f24c966c8c1/atc/db/worker_base_resource_type.go#L22-L36). This wasn't a problem before since resource caches are only initialized during a get step (which can only run on linux workers with at least 1 base resource type). But now that just steaming in a volume can initialize a resource cache, Windows and Darwin workers are now trying to register resource caches and erroring out.

clarafu:
Right, so to summarize for that issue we are seeing for `find or create container on worker <non-linux-worker>: worker base resource type disappeared` this is happening because it is trying to `InitializeResourceCache` on non linux workers but when it comes time to find the worker base resource type on the worker 

https://github.com/concourse/concourse/blob/37c4aebfb90fcbd1f7b7504846a9860f783b3a2d/atc/db/worker_resource_cache.go#L23-L26

it fails because the non-linux worker does not contain any base resource types.

clarafu:
There were three errors we saw on our ci deployment. I took some screenshots of them 

Screen Shot 2021-03-11 at 3 17 33 PM

Screen Shot 2021-03-11 at 3 17 57 PM
Screen Shot 2021-03-11 at 4 31 03 PM

chenbh:
From our convo on slack, it sounds like the resource cache is associated with a worker_base_resource_type_id is so that we can invalidate all caches if the worker gets bumped. If so then we can probably take the `worker_base_resource_type_id` from the source resource cache to populate the destination cache.

Also, I'm not 100% sure that last error was due to this PR, could just be an unfortunate flake as result of some other error.

evanchaoli:
@clarafu Will you follow up this issue? Do you expect me to do something?

PR #6507, opened 2021-02-04
worker/runtime: remove no-op device rule

muntac:
Misc

Related to #6475

CC @taylorsilva 

~~We did a bunch of testing to verify that this rule is truly a no-op. We
even wrote a garbage rule:~~

```
{Allow: false, Access: "akdsjasdjasd"},
```

~~and the container was created successfully. With or without that rule
the list of devices in
/sys/fs/cgroup/devices/[container-id]/devices.list was the same as the
unprivileged containers under
/sys/fs/cgroup/devices/garden/[container-id]/devices.list~~

~~We later learned that this default list of devices is very permissive
and "arguably insecure" in the words of a runc dev.~~

~~In future it might make sense to further restrict unprivileged
containers, specifically removing the first two rules which allow mknod
for all block and char devices, which basically gives you read-write for
any device...~~

It appears that runc adds all of these device rules by default with the exception of `/dev/fuse`.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/worker/runtime/spec/devices.go
+++ b/worker/runtime/spec/devices.go
@@ -3,29 +3,19 @@ package spec
 import "github.com/opencontainers/runtime-spec/specs-go"
 
 var (
+	// runc adds a list of devices by default.
+	// The rule below gets appended to that list.
+	// The rules along with some context can be found here:
+	// https://github.com/opencontainers/runc/blob/94ae7afa36cc3b8f551e0bc67cf83e5efdf2555f/libcontainer/specconv/spec_linux.go#L50-L192
+	// Currently these rules are highly permissive. We may want to re-visit them, but presently we don't know if they can
+	// be overriden.
+	// Linux docs about how cgroup device rules work:
+	// https://github.com/torvalds/linux/blob/master/Documentation/admin-guide/cgroup-v1/devices.rst
 	AnyContainerDevices = []specs.LinuxDeviceCgroup{
-		// runc allows these
-		{Access: "m", Type: "c", Major: deviceWildcard(), Minor: deviceWildcard(), Allow: true},
-		{Access: "m", Type: "b", Major: deviceWildcard(), Minor: deviceWildcard(), Allow: true},
-
-		{Access: "rwm", Type: "c", Major: intRef(1), Minor: intRef(3), Allow: true},          // /dev/null
-		{Access: "rwm", Type: "c", Major: intRef(1), Minor: intRef(8), Allow: true},          // /dev/random
-		{Access: "rwm", Type: "c", Major: intRef(1), Minor: intRef(7), Allow: true},          // /dev/full
-		{Access: "rwm", Type: "c", Major: intRef(5), Minor: intRef(0), Allow: true},          // /dev/tty
-		{Access: "rwm", Type: "c", Major: intRef(1), Minor: intRef(5), Allow: true},          // /dev/zero
-		{Access: "rwm", Type: "c", Major: intRef(1), Minor: intRef(9), Allow: true},          // /dev/urandom
-		{Access: "rwm", Type: "c", Major: intRef(136), Minor: deviceWildcard(), Allow: true}, // /dev/pts/*
-		{Access: "rwm", Type: "c", Major: intRef(5), Minor: intRef(2), Allow: true},          // /dev/ptmx
-		{Access: "rwm", Type: "c", Major: intRef(10), Minor: intRef(200), Allow: true},       // /dev/net/tun
-
-		// we allow this
-		{Access: "rwm", Type: "c", Major: intRef(10), Minor: intRef(229), Allow: true}, // /dev/fuse
-	}
-
-	PrivilegedOnlyDevices = []specs.LinuxDeviceCgroup{
-		{Allow: false, Access: "rwm"},
+		// This allows use of the FUSE filesystem. We are following Guardian in this case, and aren't sure of the
+		// exact use cases.
+		{Access: "rwm", Type: "c", Major: intRef(10), Minor: intRef(229), Allow: true}, 	// /dev/fuse
 	}
 )
 
 func intRef(i int64) *int64  { return &i }
-func deviceWildcard() *int64 { return intRef(-1) }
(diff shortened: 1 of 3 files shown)


taylorsilva:
The container limits tests are failing
```
[Fail] A job with a task that has container limits [It] sets the correct CPU and memory limits on the container 

/tmp/build/778af108/concourse/testflight/container_limits_test.go:39


[Fail] A job with a task that has container limits [It] successfully runs the build 

/tmp/build/778af108/concourse/testflight/container_limits_test.go:24
```
I think it's a flake. Another PR is also getting those errors on those two tests. Both pass locally for me when using containerd. Wondering if there's something about CI that's causing it to fail...

muntac:
Just noting here that the testflight failure was related to using the new runc `rc93` version in the dev image. The testflight tests have been passing since I pinned `concourse/dev` to use `rc92`. An [issue](https://github.com/concourse/concourse/issues/6512) for further investigation has been created.

I'm gonna kick off a new run of testflight for this PR with the newer `concourse/dev` image (by temporarily pinning the PR resource version).

chenbh:
@muntac were you able to modify any of those devices and have the change be reflected? I'm not sure if I'm doing something wrong, but I've tried to change all the permissions to `rw`, removing the `10:*` and `136:*` devices, and even removing all the `rwn` devices, but the `devices.list` for both privileged and non-privileged containers never changed.

taylorsilva:
@chenbh I don't think we changed any of the device rules in the code. We did manually mess around with writing our own cgroup device rules on a linux VM to help us understand what the rules mean and how they work. Based on your testing though, I wonder if these rules are getting ignored/overwritten 🤔 Guess we'll have to dig into this a bit more. Thanks for thorough testing!

chenbh:
And tracking down the runC version from [containerd](https://github.com/containerd/containerd/blob/v1.3.4/vendor.conf#L33), I would assume `{Allow: false, Access: "rwm"},` translates to `echo a *:* rwn > /sys/fs/cgroup/[container-id]/devices.deny` according to https://github.com/opencontainers/runc/blob/v1.0.0-rc10/libcontainer/specconv/spec_linux.go#L352-L365

taylorsilva:
@chenbh that's exactly what we thought!! But because we never saw the device list update to be `a *:* rw` we assumed it was a garbage rule.

chenbh:
@taylorsilva nvm it turns out there's a very large overlap between what we set and the containerd defaults. The only device we add/remove is the FUSE (10:229) device. 

Can we add a comment or some sort of documentation that all of these are default set by containerd and there's nothing we can do to modify this
https://github.com/concourse/concourse/blob/21ec03906934bb9c9852fbc46faa303105163aaa/worker/runtime/spec/devices.go#L13-L24

chenbh:
So I assumed the rule was originally set to clear out any `device.list` before the container is created. Which should've meant that if we remove it, any additional devices allowed to garden/containerd would've propagated to the container. 

1. Let's say we have a cgroup setup where garden is being run under a cgroup called `my-user`:

```
/sys/fs/cgroup/device/my-user
├── device.allow
├── device.deny
├── device.list
└── garden
    ├── device.allow
    ├── device.deny
    └── device.list
```

2. And the cgroup from `my-user` has access to some fancy device (let's say `/dev/kmsg` which is `1:11` ) in addition to the ones the container needs. I sorta emulated this by changing the [base cgroup path](https://github.com/concourse/concourse/blob/master/worker/runtime/spec/spec.go#L174-L179). (You'll also have to create these directory manually before any containers are run since `echo 'a' > devices.deny` is rejected if there's any child groups.)

```sh
cd /sys/fs/cgroup/devices
mkdir my-user
cd my-user
# required for container
echo 'a' > devices.deny
echo 'b *:* m' > devices.allow
echo 'c *:* m' > devices.allow
echo 'c 1:3 rwm' > devices.allow
echo 'c 1:5 rwm' > devices.allow
echo 'c 1:7 rwm' > devices.allow
echo 'c 1:8 rwm' > devices.allow
echo 'c 1:9 rwm' > devices.allow
echo 'c 5:* rwm' > devices.allow
echo 'c 10:* rwm' > devices.allow
echo 'c 136:* rwm' > devices.allow

# /dev/kmsg
echo 'c 1:11 rwm' > devices.allow
```
We can verify this `devices.list` propagates to any child group:
```sh
mkdir foo
cat foo/devices.list
b *:* m
c *:* m
c 1:3 rwm
c 1:5 rwm
c 1:7 rwm
c 1:8 rwm
c 1:9 rwm
c 1:11 rwm
c 5:* rwm
c 10:* rwm
c 136:* rwm
```

3. I would assume that because we no longer clear out `devices.list` aka `echo 'a *:* rwm > /sys/fs/cgroup/device/my-user/[container-id]/device.deny`, we should be able to inherit from the parent group and gain access to `/dev/kmsg`

4. However we don't!
```sh
$ fly -t dev i -b 1 -s 3
bash-5.1# cat /sys/fs/cgroup/devices/devices.list
b *:* m
c *:* m
c 1:3 rwm
c 1:5 rwm
c 1:7 rwm
c 1:8 rwm
c 1:9 rwm
c 136:* rwm
c 5:0 rwm
c 5:2 rwm
c 10:200 rwm
c 10:229 rwm
```

5. So somewhere something is resetting the `devices.list`, I'm guessing it's containerd which sets the default spec to be https://github.com/containerd/containerd/blob/v1.3.4/oci/spec.go#L230-L235

So yeah, pretty sure it's a redundant rule, but it's a huge hassle to track down, maybe we should leave a blurb somewhere so that if containerd (in addition to runc) changes their defaults in the future, we would know where to look.

muntac:
@chenbh we noticed that `runc` adds all these devices by default. so we took out all the rules and just kept `/dev/fuse` in.

chenbh (approved):

PR #6510, opened 2021-02-04
add lock for concourse migrate to latest version cmd

xtremerui:
Problem:
Previously it assumes `concourse migrate --migrate-to-version` runs only when web nodes are stopped and there are no other concurrent changes to DB ( so the cmd is used for running downgrade only).

`concourse migrate --migrate-to-latest-version` used the same code path. However, when the cmd runs in `initContainer` in concourse helm chart where migrations take certain amout of time due to huge amount of data, a race condition happens if two containers are initiating at the same time.

Change:
Add lock factory to `--migrate-to-latest-version` to ensure thread safe.

Unfortunately we don't have a proper way to test this since migrations in test environment finishes fast enough to mitigate the problem. The plan is to verify it during next hush-house version update.

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -257,6 +257,8 @@ type RunCommand struct {
 }
 
 type Migration struct {
+	lockFactory lock.LockFactory
+
 	Postgres               flag.PostgresConfig `group:"PostgreSQL Configuration" namespace:"postgres"`
 	EncryptionKey          flag.Cipher         `long:"encryption-key"     description:"A 16 or 32 length key used to encrypt sensitive information before storing it in the database."`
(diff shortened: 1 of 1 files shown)


atc/atccmd/command.go
```diff
@@ -380,10 +379,18 @@ func (cmd *Migration) rotateEncryptionKey() error {
...
+	}
+	defer lockConn.Close()
+
+	lockFactory := lock.NewLockFactory(lockConn, metric.LogLockAcquired, metric.LogLockReleased)
+
 	helper := migration.NewOpenHelper(
 		defaultDriverName,
 		cmd.Postgres.ConnectionString(),
-		nil,
+		lockFactory,
```
aoldershaw, line 393: Could you pass the lock factory in for *all* the `concourse migrate` commands (at the very least, the ones that mutate the DB - `migrateDBToVersion` and `rotateEncryptionKey` as well)

aoldershaw: (I recognize that the other commands aren't used in the chart, but it seems like something we should probably have anyway)

aoldershaw (approved):

PR #6511, opened 2021-02-04
Optimize check deletion

aoldershaw:

Reduces load on the database caused by frequent creation/deletion of checks.

* Re-adds the `build_image_resource_caches_job_build_id` index which was dropped in #5993 (it's chill, it hasn't been released)
* Adds an on `builds (resource_type_id)` to optimize deleting old checks for resource types (we already had the matching index for resource checks)
* Send check build events to their own partition of `build_events` (called `check_build_events`)
* Adds a new GC component for deleting builds/build events for completed checks, but only if they have a newer check for the same resource/resource type
  * This newer check doesn't also have to be completed
  * Without requiring at least one newer check, the build logs could disappear while you're trying to investigate a problem with resource checks, which would be annoying

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


atc/db/check_lifecycle.go
```diff
@@ -0,0 +1,41 @@
...
+        FROM builds b
+        WHERE completed AND resource_type_id IS NOT NULL
+        AND EXISTS (SELECT * FROM builds b2 WHERE b.resource_type_id = b2.resource_type_id AND b.id < b2.id)
+      ),
+      deleted_builds AS
+      (
+        DELETE FROM builds WHERE id IN (SELECT id FROM deletable_builds)
+        RETURNING id
+      )
+      DELETE FROM check_build_events WHERE build_id IN (SELECT id FROM deletable_builds)
```
clarafu, line 38: I think if we assume that the `deletable_builds` CTE will usually return a lot of rows, `DELETE FROM check_build_events USING deletable_builds WHERE deletable_builds.id = check_build_events.build_id` might perform better than `WHERE .. IN ( SELECT .. )`. I think because the `WHERE .. IN ...` might be slow when it has to compare against a large list of ids.

atc/db/check_lifecycle.go
```diff
@@ -0,0 +1,41 @@
...
+        AND EXISTS (SELECT * FROM builds b2 WHERE b.resource_id = b2.resource_id AND b.id < b2.id)
+          UNION ALL
+        SELECT id
+        FROM builds b
+        WHERE completed AND resource_type_id IS NOT NULL
+        AND EXISTS (SELECT * FROM builds b2 WHERE b.resource_type_id = b2.resource_type_id AND b.id < b2.id)
+      ),
+      deleted_builds AS
+      (
+        DELETE FROM builds WHERE id IN (SELECT id FROM deletable_builds)
```
clarafu, line 35: Same here!

atc/db/migration/migrations/1612457776_add_check_build_events_partition.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+  CREATE TABLE check_build_events () INHERITS (build_events);
+  CREATE INDEX check_build_events_build_id ON check_build_events (build_id);
+  CREATE INDEX check_build_events_build_id_event_id ON check_build_events (build_id, event_id);
```
clarafu, line 4: Does this need to be a unique index?

atc/db/check_lifecycle.go
aoldershaw: Oh sick, didn't know/forgot about `DELETE FROM ... USING`. Will make the change!

atc/db/migration/migrations/1612457776_add_check_build_events_partition.up.sql
aoldershaw: whoops thank you - fixed

atc/db/check_lifecycle.go
```diff
@@ -0,0 +1,41 @@
...
+        FROM builds b
+        WHERE completed AND resource_type_id IS NOT NULL
+        AND EXISTS (SELECT * FROM builds b2 WHERE b.resource_type_id = b2.resource_type_id AND b.id < b2.id)
+      ),
+      deleted_builds AS
+      (
+        DELETE FROM builds USING deletable_builds WHERE builds.id = deletable_builds.id
+        RETURNING builds.id
+      )
+      DELETE FROM check_build_events USING deletable_builds WHERE build_id = deletable_builds.id
```
clarafu, line 38: Oh were we going to use the `deleted_builds` table rather than the `deletable_builds`?

aoldershaw: I decided to use `deletable_builds` instead in the hopes that Postgres would realize "oh, you don't actually need the `RETURNING builds.id` part from `deleted_builds` and wouldn't store anything. I'm really just using `deleted_builds` as a hacky way to chain deletes in a single query

clarafu: Ohh could we just remove the returning part of the delete then?

aoldershaw: oh lol I just assumed that you couldn't use a straight `DELETE` in a CTE and didn't bother testing it, but yeah you're totally right. removed

aoldershaw: Actually I refactored it again to remove the `deletable_builds` CTE and just use a subquery within `deleted_builds`, and added back the `RETURNING id` clause - seems a bit more straightforward that way

atc/gc/check_collector_test.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"context"
+
+	"github.com/concourse/concourse/atc/db/dbfakes"
+	"github.com/concourse/concourse/atc/gc"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("AccessTokensCollector", func() {
```
vito, line 12:
```suggestion
var _ = Describe("ChecksCollector", func() {
```

aoldershaw: whoops, fixed

clarafu (approved):
Looks good!! I'm a little worried about the speed of the new query but not sure if we would want to go through stress testing or just hope that prod will be big enough to test it. 

Also wanted to get your opinion if you though 30 second gc interval for checks would be good enough?

aoldershaw:
@clarafu I'm inclined to test it on hush-house, which should be a bit more representative of reality than prod/a cooked-up stress test - if it hurts performance, we can rollback.

> Also wanted to get your opinion if you though 30 second gc interval for checks would be good enough?

Do you think it should be run more or less frequently? Do we have a sense of how many checks are run in a 30s interval?

risicle:
Sneak preview of an issue we'll likely open separately in the next few days: you were right to be worried about the speed of the new query.

Just had an incident where this query decided to start but would never complete, and every time we rolled our concourse-web nodes we'd end up with another copy of it trying to run alongside the zombie query that was previously running.

The problem is that it's just one big query. This means that:
 - its progress will never get persisted until the query completes in full
 - it potentially has to acquire a huge number of row locks (all at the same time) before it can complete - and of course, those row locks will interfere with the regular operation of the database.

In the end, we had to run this query manually in `build.id`-partitioned chunks of around 100k at a time. And even then we needed to stop the concourse-web processes to remove the lock-contention and make any headway in the more populated ranges.

PR #6517, opened 2021-02-08
Add index and mini refactor to the check lifecycle query

clarafu:

This PR adds an index to help speed up the deletion of the check builds and also adds a small refactor to the check lifecycle query. It should help speed up the query a little by using the `build_id` column on the resources table to figure out whether or not a check build needs to be gced or not.

The index will help speed up the `constraint resources_build_id_fkey on builds` from the checks gc query, here is the `EXPLAIN ANALYZE` of the query without the added index https://explain.depesz.com/s/Gl3M

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/check_lifecycle.go
+++ b/atc/db/check_lifecycle.go
@@ -18,13 +18,17 @@ func NewCheckLifecycle(conn Conn) CheckLifecycle {
 
 func (cl *checkLifecycle) DeleteCompletedChecks() error {
 	_, err := cl.conn.Exec(`
-      WITH deleted_builds AS
-      (
+      WITH resource_builds AS (
+        SELECT build_id
+        FROM resources
+        WHERE build_id IS NOT NULL
+      ),
+      deleted_builds AS (
         DELETE FROM builds USING (
           SELECT id
           FROM builds b
           WHERE completed AND resource_id IS NOT NULL
-          AND EXISTS (SELECT * FROM builds b2 WHERE b.resource_id = b2.resource_id AND b.id < b2.id)
+          AND NOT EXISTS ( SELECT 1 FROM resource_builds WHERE build_id = b.id )
             UNION ALL
           SELECT id
           FROM builds b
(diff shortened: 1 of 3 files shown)


atc/db/check_lifecycle.go
```diff
@@ -18,13 +18,17 @@ func NewCheckLifecycle(conn Conn) CheckLifecycle {
 
 func (cl *checkLifecycle) DeleteCompletedChecks() error {
 	_, err := cl.conn.Exec(`
-      WITH deleted_builds AS
-      (
+      WITH resource_builds AS (
```
aoldershaw, line 21: I tried to simplify this query a bit and came up with

```
SELECT b.id
FROM builds b
JOIN resources r ON r.id = b.resource_id
WHERE completed
AND b.id < r.build_id;
```

This query is a bit slower on hush-house than the one you provided, though - perhaps with some different indexes things would be a bit faster?

aoldershaw (approved):
LGTM!

I tried refactoring the query to hopefully optimize it further, but it ended up just making it slower 🤷

atc/db/check_lifecycle.go
clarafu: Ah yeaah I think if we added an index like `CREATE INDEX deleteme4 ON builds (resource_id, id) WHERE resource_id IS NOT NULL AND completed;`? Not sure how much it would speed up the query. I do know that there is a slow part on the `completed` and `resource_id` is not null filter here https://explain.depesz.com/s/xgPG#l6? Hm wondering if I should test out adding this index to hush-house to see how it affects the performance of this query!

The really unfortunate part is I think it is taking the most time going through the `check_build_events` table and I have no idea on how to speed that part up :(

aoldershaw: Yeah was just pointing out an alternative! I think the query you have now is good, and as you point out, further optimizations here are probably moot given that they aren’t really the bottleneck

PR #6522, opened 2021-02-08
Removes unnecessary indexes from build events tables

clarafu:

This PR removes a few indexes that should be unnecessary. Any queries that query on the `build_id` column can actually use the unique multi column index on `build_id` and `event_id`. Since the build events tables are updated really frequently, limiting the number of indexes that are applied to the table will help with the speed of inserts or updates.

These changes to existing migrations should be okay because it is not released yet with v7.0 but we did upgrade `ci` and `hush-house` with the migrations without this change so we will need to remove the indexes from those environments manually.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/migration/migrations/1603405319_prepare_build_events_for_bigint_migration.down.sql
+++ b/atc/db/migration/migrations/1603405319_prepare_build_events_for_bigint_migration.down.sql
@@ -3,7 +3,6 @@ BEGIN;
   UPDATE build_events SET build_id_old = build_id WHERE build_id_old IS NULL;
 
   -- drop the indexes for the bigint column
-  DROP INDEX build_events_build_id_idx;
   DROP INDEX build_events_build_id_event_id;
 
   -- drop the bigint column
@@ -11,6 +10,6 @@ BEGIN;
 
   -- rename everything back
   ALTER TABLE build_events RENAME COLUMN build_id_old TO build_id;
-  ALTER INDEX build_events_build_id_old_idx RENAME TO build_events_build_id_idx;
   ALTER INDEX build_events_build_id_old_event_id  RENAME TO build_events_build_id_event_id;
+  CREATE INDEX build_events_build_id_idx ON build_events (build_id);
 COMMIT;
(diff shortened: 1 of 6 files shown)


aoldershaw:
Are these indexes used by the `DeleteBuildEventsByBuildIDs` method?

https://github.com/concourse/concourse/blob/370b34f7524696627c0047da39a71c2d114f9dd6/atc/db/pipeline.go#L979-L982

Presumably that could just use the multi-column index, though (I'd think, at least). Would be really nice to get rid of these indexes, but just want to be sure it's not a huge performance regression (though it doesn't seem like it)

aoldershaw:
Ah just recalled that we had a test suite for ensuring that we're still doing index scans. Cool!

https://github.com/concourse/concourse/blob/7a1d9eba89fd921b4c951014867379d0f1ed409b/atc/db/migration/build_events_bigint_test.go#L59-L66

I'm going to try deleting all the `build_id` indexes from CI to see what effect it has on performance (if any)

aoldershaw (changes requested):
We'll also need to replace the trigger so we don't create the index for new pipelines: https://github.com/concourse/concourse/blob/370b34f7524696627c0047da39a71c2d114f9dd6/atc/db/migration/migrations/1530824006_create_pipelines_trigger.up.sql#L13

clarafu:
@aoldershaw Oh maybe I'm not understanding triggers properly but I originally thought that because we overwrote the `on_pipeline_insert()` trigger function here 

https://github.com/concourse/concourse/blob/7a1d9eba89fd921b4c951014867379d0f1ed409b/atc/db/migration/migrations/1606068653_build_events_partitions_bigint.up.sql#L37-L44

That it would replace the trigger function in the migration that you linked. Or am I totally wrong about that?

aoldershaw (approved):
Whoops I'm dumb! Was not reading closely enough.

clarafu:
Nah migrations are tricky and annoying 👎 , thanks!!

PR #6539, opened 2021-02-16
Adjust spacing and padding for elements in pipeline card view in Dashboard

### LINKED ISSUE #5971: CI monitor and Pipeline tile alignment adjustments
### Current state

Screen Shot 2020-08-21 at 11 14 46 AM

Inspecting the dom and adding gridlines, we can see that the CI monitor dashboard uses a variety of different spacings and paddings (7px, 12px, 30px, etc). The team names and section headers line up with the pipeline labels rather than the pipeline tiles themselves, and over time these values have shifted as different enhancements were introduced to the dashboard.

### Proposed state

Screen Shot 2020-08-21 at 11 07 36 AM

By using 32px spacing around all of our elements, and lining up the team and section headers with the pipeline tiles, we get a more consistent and orderly view.

### Current state

Screen Shot 2020-09-01 at 9 48 36 AM

Likewise, there are a number of different units dividing up the pipeline tile.

### Proposed state

Screen Shot 2020-09-01 at 9 45 25 AM

We can simplify the layout a fair bit and create symmetry between the header and footer by aligning everything in an orderly way.

## Figma mockups

[Mockups are available in Figma to get a detailed look at spacing and alignment](https://www.figma.com/file/fuoFSVcTGptiS21P2POA2XbN/Concourse-UI-kit-and-Experiments?node-id=1675%3A0). Look for the highlighted frames for the target variations.

## Acceptance Criteria

- [x] Line up team names with pipeline tiles so that dashboard has consistent margins
- [x] Pipeline tile spacing to be 32px, remove any alternative padding or margins wherever possible
- [x] Ensure icon sizing and stroke is consistent on pipeline icons
- [x] 16px, 8px spacing between elements in the pipeline tile

## Context

* Part of #5964 Web UI/UX Experience

xtremerui:
Feature

closes #5971  .

update spacing in pipeline card in dashboard view

## Notes to reviewer:
Certain number in tests is calculated by adding sizing and spacing from multiple cards. Since the horizontal spacing is increased a little bit from 24px to 32px, we have to increase the view port size from 600 to 700 for allowing multiple cards to show.

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed


web/assets/css/dashboard.less
```diff
@@ -62,9 +62,8 @@
     background: @grey80;
     z-index: 2;
     opacity: 0.9;
-    padding-top: (35px * @scale);
-    padding-bottom: (35px * @scale);
-    padding-left: ((50px + @name-padding) * @scale);
+    padding-top: (64px * @scale);
```
aoldershaw, line 65: Could you get rid of this padding - it results in the padding being doubled for team headers after the first one, since there's padding under the cards, too:

Screen Shot 2021-02-17 at 12 20 09 PM

You'll also need to add a bottom margin to the `all pipelines` header (that's done in Elm)

web/assets/css/dashboard.less
```diff
@@ -62,9 +62,8 @@
...
+    padding-left: ((39px + @name-padding) * @scale);
```
aoldershaw, line 66: Can we get rid of `@name-padding` altogether and just use a fixed `64px` here? `@name-padding` isn't used anywhere else.

Also, looks like the other variables here (`@banner-height`, `@footer-padding`) are also completely unused, so we can remove them too

web/elm/src/Dashboard/Grid/Constants.elm
```diff
@@ -21,17 +21,21 @@ cardBodyHeight =
...
 padding : number
 padding =
-    25
+    32
 
 
 cardHeaderHeight : number -> number
 cardHeaderHeight numRows =
-    20 + 30 * numRows
+    32 + 21 * numRows
```
aoldershaw, line 29: I don't think we want 21 here - we should give some padding between the rows when we have multiple header rows.

On `master`:
Screen Shot 2021-02-17 at 12 35 42 PM


With this change:
Screen Shot 2021-02-17 at 12 35 15 PM

web/elm/src/Dashboard/Grid/Constants.elm
```diff
@@ -21,17 +21,21 @@ cardBodyHeight =
...
+
+
+
+-- padding top 32 + line height 25 + spacing bottom 32
```
aoldershaw, line 33: Maybe we can encode these as variables and use them in `Dashboard/Styles.elm` so they can't get out of sync? Then we wouldn't need this comment

aoldershaw (changes requested):
Looks a lot better for the most part! Left some comments about places that should be adjusted spacing-wise.

Also, what are your thoughts on transitioning the styles from `dashboard.less` to `Dashboard/Styles.elm`? It's higher effort, but it'd be nice to not allow these to drift apart, especially with the tricky lazy rendering on the Dashboard that requires knowing the size of everything? I don't think it's a must, but would be nice to have

web/elm/src/Dashboard/Grid/Constants.elm
xtremerui: @matthewpereira from the Figma render we see the height for one row header is 21px. When there are multiple rows what vertical spacing should be in between?

web/elm/src/Dashboard/Grid/Constants.elm
xtremerui: Screen Shot 2021-02-19 at 9 40 56 AM
Screen Shot 2021-02-19 at 9 41 05 AM

these are some samples when vertical spacing between header rows is set to 8px

web/elm/src/Dashboard/Grid/Constants.elm
xtremerui: Screen Shot 2021-02-19 at 9 40 56 AM
Screen Shot 2021-02-19 at 9 41 05 AM

these are samples when vertical spacing is set to 8px between header rows

web/elm/src/Dashboard/Grid/Constants.elm
```diff
@@ -21,17 +21,27 @@ cardBodyHeight =
...
+
+carHeaderVerticalSpace : number
+carHeaderVerticalSpace =
+    8
 
 
 cardHeaderHeight : number -> number
 cardHeaderHeight numRows =
-    20 + 30 * numRows
+    32 + 21 * numRows + (numRows - 1) * carHeaderVerticalSpace
```
aoldershaw, line 34: I wonder if we really care to make the spacing between rows "exactly" 8px - the thing is, the 21px line height is already imprecise, since most characters aren't actually 21px tall. Maybe we can do a more even number instead, if for nothing else than to make the heights little more sane?

One thing I tried was removing the padding div between rows and just making the line-height `32px` - the problem there was that the top and bottom row looked weird since there was so much spacing on the top and bottom. But maybe we could absorb some of the 16px header padding into the rows - that is, there's still 16px of empty space between the edge of the header and the top of the text, but some of that is from the empty space created by the `line-height`, rather than the `padding` on the header.

```diff
diff --git a/web/elm/src/Dashboard/Grid/Constants.elm b/web/elm/src/Dashboard/Grid/Constants.elm
index 448fb1ce0..6b5387a6f 100644
--- a/web/elm/src/Dashboard/Grid/Constants.elm
+++ b/web/elm/src/Dashboard/Grid/Constants.elm
@@ -1,6 +1,9 @@
 module Dashboard.Grid.Constants exposing
     ( cardBodyHeight
     , cardHeaderHeight
+    , cardHeaderPadding
+    , cardHeaderPaddingOverlap
+    , cardHeaderRowHeight
     , cardWidth
     , groupHeaderHeight
     , padding
@@ -24,14 +27,27 @@ padding =
     32
 
 
-carHeaderVerticalSpace : number
-carHeaderVerticalSpace =
+cardHeaderRowHeight : number
+cardHeaderRowHeight =
+    32
+
+
+cardHeaderPadding : number
+cardHeaderPadding =
+    16
+
+
+cardHeaderPaddingOverlap : number
+cardHeaderPaddingOverlap =
+    -- this is the number of pixels of the padding defined in cardHeaderPadding
+    -- that cardHeaderRowHeight should contribute. this is to avoid having too
+    -- much padding on the top/bottom row
     8
 
 
 cardHeaderHeight : number -> number
 cardHeaderHeight numRows =
-    32 + 21 * numRows + (numRows - 1) * carHeaderVerticalSpace
+    2 * (cardHeaderPadding - cardHeaderPaddingOverlap) + cardHeaderRowHeight * numRows
 
 
 groupHeaderLineHeight : number
diff --git a/web/elm/src/Dashboard/Pipeline.elm b/web/elm/src/Dashboard/Pipeline.elm
index a21e683fc..626d19a85 100644
--- a/web/elm/src/Dashboard/Pipeline.elm
+++ b/web/elm/src/Dashboard/Pipeline.elm
@@ -277,12 +277,8 @@ transition =
 headerView : PipelinesSection -> Pipeline -> Bool -> Float -> Bool -> Bool -> Html Message
 headerView section pipeline resourceError headerHeight viewingInstanceGroups inInstanceGroup =
     let
-        verticalSpacer =
-            Html.div [ style "height" "8px" ] []
-
         rows =
-            List.intersperse verticalSpacer
-                (headerRows section viewingInstanceGroups pipeline inInstanceGroup)
+            headerRows section viewingInstanceGroups pipeline inInstanceGroup
 
         resourceErrorElem =
             Html.div
diff --git a/web/elm/src/Dashboard/Styles.elm b/web/elm/src/Dashboard/Styles.elm
index 59301c390..de693dd80 100644
--- a/web/elm/src/Dashboard/Styles.elm
+++ b/web/elm/src/Dashboard/Styles.elm
@@ -212,11 +212,18 @@ noPipelineCardHeader =
 
 pipelineCardHeader : Float -> List (Html.Attribute msg)
 pipelineCardHeader height =
+    let
+        paddingHorizontal =
+            String.fromFloat GridConstants.cardHeaderPadding ++ "px"
+
+        paddingVertical =
+            String.fromFloat (GridConstants.cardHeaderPadding - GridConstants.cardHeaderPaddingOverlap) ++ "px"
+    in
     [ style "background-color" Colors.card
     , style "color" Colors.dashboardPipelineHeaderText
     , style "font-size" "1.5em"
     , style "letter-spacing" "0.1em"
-    , style "padding" "16px"
+    , style "padding" <| paddingVertical ++ " " ++ paddingHorizontal
     , style "height" <| String.fromFloat height ++ "px"
     , style "box-sizing" "border-box"
     ]
@@ -230,7 +237,7 @@ instanceGroupCardHeader =
 pipelineName : List (Html.Attribute msg)
 pipelineName =
     [ style "width" "240px"
-    , style "line-height" "21px"
+    , style "line-height" <| String.fromFloat GridConstants.cardHeaderRowHeight ++ "px"
     , style "white-space" "nowrap"
     , style "overflow" "hidden"
     , style "text-overflow" "ellipsis"
```

aoldershaw (changes requested):
Looks great! One typo, and I have one other suggestion to make the numbers a little more uniform

xtremerui:
Screen Shot 2021-02-19 at 9 40 56 AM
Screen Shot 2021-02-22 at 1 51 30 PM

so there is a side by side comparison, top is a 8px dedicate spacer with 21px line height ; bottom is 32px heigh without spacer.

The most obivious difference I can tell from the screeshots is the vertical spacing between rows is bigger on bottom one. Pixel wise, the total heigh of header section is 80px on bottome one vs 82px on top one.

I think top one has clear size on each elements, which you can just use browser dev tool to fetch (but not as acurate); while the bottom one you'd have to go to code to find out the overlap size between the height and the header wrapper padding to figure out the actuall padding between A: text and top/bottom border, B: row texts. IMHO I think bottom one is less explicit than top.

Visually I don't have any preference. Maybe an opinion from @matthewpereira could be useful?

web/elm/src/Dashboard/Grid/Constants.elm
```diff
@@ -21,17 +21,27 @@ cardBodyHeight =
 
 padding : number
 padding =
-    25
+    32
+
+
+carHeaderVerticalSpace : number
+carHeaderVerticalSpace =
```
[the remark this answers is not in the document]
aoldershaw: Actually, can we rename this to something like `cardHeaderRowGap`? Vertical space suggests its the full space it takes up, but it's just the space between rows - `RowGap` is consistent with the CSS grid property `row-gap`: https://developer.mozilla.org/en-US/docs/Web/CSS/row-gap

web/elm/src/Dashboard/Grid/Constants.elm
aoldershaw, line 34: Can we extract these numbers to named constants, and use the constants when defining the styles in `Dashboard/Styles.elm`? e.g. this line becomes

```
    2 * cardHeaderPadding + cardHeaderRowLineHeight * numRows + (numRows - 1) * cardHeaderRowGap
```

and then in `Dashboard/Styles.elm` we use `cardHeaderPadding` in place of the hard-coded `16px` padding, and `cardHeaderRowLineHeight` in place of the hard-coded `21px` line height. Similarly in `Dashboard/Pipeline.elm`, we can use `cardHeaderRowGap` in place of the hard-coded `8px` height.

This way, there's an explicit dependency between the actual styles and their manually calculated heights, so the two can't drift apart

web/elm/src/Dashboard/Grid/Constants.elm
```diff
@@ -21,17 +21,27 @@ cardBodyHeight =
...
 
 cardHeaderHeight : number -> number
 cardHeaderHeight numRows =
-    20 + 30 * numRows
+    32 + 21 * numRows + (numRows - 1) * carHeaderVerticalSpace
+
+
+groupHeaderLineHeight : number
+groupHeaderLineHeight =
+    25
```
aoldershaw, line 39: Can this be used in the CSS for the header?

web/elm/src/Dashboard/Grid/Constants.elm
```diff
@@ -21,17 +21,27 @@ cardBodyHeight =
...
 
 
 groupHeaderHeight : number
 groupHeaderHeight =
-    60
+    padding + groupHeaderLineHeight
 
 
 sectionHeaderHeight : number
```
aoldershaw, line 47: This should be increased - it's now 94px. Perhaps we can split it up into two constants - `sectionHeaderLineHeight: 30px` + `2 * padding`. If we do this, we should use padding in `Dashboard/Styles.elm` for `pipelineSectionHeader`

web/elm/src/Dashboard/Grid.elm
```diff
@@ -278,7 +278,7 @@ computeFavoritesLayout params cards =
     in
     { cards = result.allCards |> List.filter (isVisible params)
     , headers = headers |> List.filter (isVisible params)
-    , height = result.totalHeight
+    , height = result.totalHeight - padding
```
aoldershaw, line 281: Rather than subtracting `padding`, maybe we should change this line:
https://github.com/concourse/concourse/blob/5530bdc6ee6461af68d6d98cf73b9c2f8f723b71/web/elm/src/Dashboard/Dashboard.elm#L1438
to
```
 , Views.Styles.separator 0 
```

and then get rid of `Dashboard.Grid.Constants.sectionSpacerHeight` entirely (including deleting https://github.com/concourse/concourse/blob/5530bdc6ee6461af68d6d98cf73b9c2f8f723b71/web/elm/src/Dashboard/Dashboard.elm#L1445). This is more consistent with how `computeLayout` works - the bottom padding is included in the `.dashboard-team-pipelines` div

web/elm/src/Dashboard/Grid/Constants.elm
```diff
@@ -21,17 +21,27 @@ cardBodyHeight =
...
+
+groupHeaderLineHeight : number
+groupHeaderLineHeight =
+    25
 
 
 groupHeaderHeight : number
 groupHeaderHeight =
-    60
+    padding + groupHeaderLineHeight
```
aoldershaw, line 44: Now that the padding is included in the `groupHeaderHeight` (which makes sense in general, I think), it's causing issues with a previous computation for lazy rendering. We need to get rid of

https://github.com/concourse/concourse/blob/5530bdc6ee6461af68d6d98cf73b9c2f8f723b71/web/elm/src/Dashboard/Dashboard.elm#L1470

...or when there are many teams, pipelines will start to disappear:


https://user-images.githubusercontent.com/26583442/108868309-1453ad00-75c4-11eb-999e-c69abfce6604.mov

aoldershaw (changes requested):

matthewpereira:
@xtremerui I don't see a huge difference in vertical spacing; I tried overlapping the images to inspect the differences, and each one appears to have about 36px in space between the heading, the team name, and the first pipeline tile. Maybe I'm misunderstanding you though.

I think spacers are a lot easier to understand and visualize for engineers (at least, they are for me) when they're reading a codebase, compared to line height changes. WDYT?

web/elm/src/Dashboard/Grid/Constants.elm
xtremerui: do you mean add a line-height 25px here ?https://github.com/concourse/concourse/blob/398ef03bbf45044912a3e8fda85200ba3ebb9480/web/assets/css/dashboard.less#L57

Without it the header line height is still 25px by the enforcement from (height 57px - margin-bottom 32px). Do you just want to explicity set it in CSS as well?

xtremerui:
@matthewpereira sry I should be more clear. Please try measure the spacing between text "3 text" and the next line. Also the spacing between text "no instance vars" and the grey border line below.  There should be a difference there. But it could be too tiny to be noticed. 

Anyway, if you are happy with both it'd be good enough for us. Thx.

matthewpereira:
Sorry for misunderstanding! I think that's within acceptable limits - feel free to go ahead with it. 🙇

aoldershaw (approved):
Looks great! Appreciate your patience

PR #6550, opened 2021-02-18
Switch migrations + web assets to Go 1.16 embedding

vito:
- tidier error check (thanks Go 1.16!)
- bump minimum Go to 1.16
- web: behavior: use go:embed, re-parse index.html
- migration: use go:embed
- remove packr

edit: woops this was created with `gh pr create -f` and it nixed all of the template. but this shouldn't really have any external impact at least. 🤔 further info available in the commit messages!

*note: this does not include Dex, which currently uses `pkger`*


xtremerui:
This will be backported to release/6.7.x and release/7.0.x branches.

tsa/tsacmd/server.go
```diff
@@ -86,7 +87,7 @@ func (server *server) Serve(listener net.Listener) {
 	for {
 		c, err := listener.Accept()
 		if err != nil {
-			if !strings.Contains(err.Error(), "use of closed network connection") {
+			if !errors.Is(err, net.ErrClosed) {
```
chenbh, line 90: Should also be applied to:
```
atc/worker/gclient/retryable_garden_connection.go
83:		if strings.Contains(err.Error(), "use of closed network connection") {
101:		if strings.Contains(err.Error(), "use of closed network connection") {
```

web/cmd/main.go
```diff
@@ -1,40 +0,0 @@
...
-
-	http.Handle("/api/", proxyHandler)
-	http.Handle("/auth/", proxyHandler)
-	http.Handle("/oauth/", proxyHandler)
-	http.Handle("/", webHandler)
-
-	if err = http.ListenAndServe(":8081", nil); err != nil {
-		logger.Error("server-error", err)
-	}
-}
```
chenbh, line 40: What was the purpose of this file? Was it possible to build the UI server by itself?

taylorsilva (commented):
LGTM outside of Daniel's one comment.

Everything loaded fine and didn't notice any errors in firefox's dev console. The embed package is pretty cool!

web/cmd/main.go
taylorsilva: Yup https://github.com/concourse/web/blob/8c95285e65efc043abcf6d285b011c01d309087b/cmd/main.go

vito: Yeah, I don't think we use it anymore now that the `docker-compose` flow is pretty smooth so I figured I'd just remove it.

go.mod
```diff
@@ -105,6 +104,6 @@ require (
 	sigs.k8s.io/yaml v1.2.0
 )
 
-go 1.13
+go 1.16
```
ringods, line 107: @vito don't forget to update the Go version here as well:

https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md#development-dependencies

vito: good call!

PR #6561, opened 2021-02-20
atc: Cleanup from staticcheck

taylorsilva:
Resolve a bunch of the suggestions from staticcheck, which should be similar to large portion of those reported by Deepsource as well. This should help make deepsource more useful during PR review. This PR mostly targets the `atc` package.

I tried to use the deepsource cli but I couldn't figure out how to get it to report directly in the console or where I could see the results in the web ui.

A nice next step would be to have deepsource ignore some of these that we know about but can't tackle right now, like the `atc/worker/gclient/connection/connection_hijacker.go`. And ignore one's that we're probably not going to change, like the `atc/postgresrunner` package. Looks like that can be configured via the ui([maybe?](https://deepsource.io/docs/access-control/people-access/ignore-permissions.html))

## Notes to reviewer:
All commits are very small, usually focused on the same type of problem.
These are the remaining messages for the `atc` package. They're not likely to pop-up in deepsource since they're rarely touched. Does deepsource respect `//lint:ignore` like staticcheck does?
```
$ staticcheck ./atc/...
# Maybe we should regenerate fakes. I think some are outdated?
atc/api/pipelineserver/pipelineserverfakes/fake_logger.go:137:58: error should be returned as the last argument (ST1008)
atc/api/pipelineserver/pipelineserverfakes/fake_logger.go:170:58: error should be returned as the last argument (ST1008)
atc/engine/check_delegate.go:133:9: empty branch (SA9003)
atc/postgresrunner/ginkgo.go:7:2: should not use dot imports (ST1001)
atc/postgresrunner/ginkgo.go:8:2: should not use dot imports (ST1001)
atc/postgresrunner/postgresrunner.go:19:2: should not use dot imports (ST1001)
atc/worker/gclient/connection/connection_hijacker.go:74:12: httputil.NewClientConn has been deprecated since Go 1.0: Use the Client or Transport in package net/http instead.  (SA1019)
# didn't dig into this one. Was worried the change would break something down stream of it
atc/worker/gclient/connection/connection_test.go:346:12: limits.LimitInShares is deprecated: Use Weight instead.  (SA1019)
atc/worker/placement.go:15:40: duplicate struct tag "choice" (SA5008)
```

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


atc/api/accessor/claims_cacher.go
```diff
@@ -49,13 +49,10 @@ func (c *claimsCacher) GetAccessToken(rawToken string) (db.AccessToken, bool, er
 		return db.AccessToken{Token: rawToken, Claims: entry.claims}, true, nil
 	}
 
-	token, found, err := c.accessTokenFetcher.GetAccessToken(rawToken)
+	token, _, err := c.accessTokenFetcher.GetAccessToken(rawToken)
```
taylorsilva, line 52: Checking `found` broke caching which is why this commit ignores it. I might dig into this again with a clearer head. Maybe the test is wrong?

aoldershaw: Yeah, looks like I was a bit lazy when I wrote this test, and a bit negligent when I wrote the code. This diff fixes the test and adds a missing test for the `!found` case

```diff
diff --git a/atc/api/accessor/claims_cacher.go b/atc/api/accessor/claims_cacher.go
index ffb5d6916..aa0daff41 100644
--- a/atc/api/accessor/claims_cacher.go
+++ b/atc/api/accessor/claims_cacher.go
@@ -49,10 +49,13 @@ func (c *claimsCacher) GetAccessToken(rawToken string) (db.AccessToken, bool, er
 		return db.AccessToken{Token: rawToken, Claims: entry.claims}, true, nil
 	}
 
-	token, _, err := c.accessTokenFetcher.GetAccessToken(rawToken)
+	token, found, err := c.accessTokenFetcher.GetAccessToken(rawToken)
 	if err != nil {
 		return db.AccessToken{}, false, err
 	}
+	if !found {
+		return db.AccessToken{}, false, nil
+	}
 	payload, err := json.Marshal(token.Claims)
 	if err != nil {
 		return db.AccessToken{}, false, err
diff --git a/atc/api/accessor/claims_cacher_test.go b/atc/api/accessor/claims_cacher_test.go
index 90f959667..206ee0a7b 100644
--- a/atc/api/accessor/claims_cacher_test.go
+++ b/atc/api/accessor/claims_cacher_test.go
@@ -29,11 +29,26 @@ var _ = Describe("ClaimsCacher", func() {
 	})
 
 	It("fetches claims from the DB", func() {
-		claimsCacher.GetAccessToken("token")
+		expectedToken := db.AccessToken{Claims: db.Claims{Username: "foo"}}
+		fakeAccessTokenFetcher.GetAccessTokenReturns(expectedToken, true, nil)
+
+		token, found, err := claimsCacher.GetAccessToken("token")
+		Expect(err).ToNot(HaveOccurred())
+		Expect(found).To(BeTrue())
+		Expect(token).To(Equal(expectedToken))
+
 		Expect(fakeAccessTokenFetcher.GetAccessTokenCallCount()).To(Equal(1), "did not fetch from DB")
 	})
 
+	It("returns not found if the token isn't found", func() {
+		fakeAccessTokenFetcher.GetAccessTokenReturns(db.AccessToken{}, false, nil)
+		_, found, err := claimsCacher.GetAccessToken("token")
+		Expect(err).ToNot(HaveOccurred())
+		Expect(found).To(BeFalse())
+	})
+
 	It("doesn't fetch from the DB when the result is cached", func() {
+		fakeAccessTokenFetcher.GetAccessTokenReturns(db.AccessToken{}, true, nil)
 		claimsCacher.GetAccessToken("token")
 		claimsCacher.GetAccessToken("token")
 		Expect(fakeAccessTokenFetcher.GetAccessTokenCallCount()).To(Equal(1), "did not cache claims")
```

aoldershaw:
Noice
Screen Shot 2021-02-22 at 10 34 02 AM

atc/db/dbtest/builder.go
```diff
@@ -79,6 +79,7 @@ func (builder Builder) WithWorker(worker atc.Worker) SetupFunc {
 				return fmt.Errorf("team does not exist: %s", worker.Team)
 			}
 
+			//lint:ignore SA4006 linter is wrong
```
aoldershaw, line 82: I think the linter is actually right here. On line 73, we create a new binding for `err` that only exists in this branch, and that's the `err` we replace on this line (while the check outside of this if statement is still checking the one defined on line 71)

aoldershaw (commented):
Looks good! Just the one suggestion I left about fixing up the `claims_cacher` tests and this should be good to merge. Thanks for taking this on!

atc/db/dbtest/builder.go
aoldershaw: https://play.golang.org/p/JhkCzCjBfa0

taylorsilva: Good catch! I totally missed that line all together!

aoldershaw (approved):
LGTM

PR #6585, opened 2021-02-24
6.7 remove packr backport

xtremerui:
backport of #6550


vito (approved):
for posterity: this also pulls in other changes to the migrations package (notably the simplification of SQL migration handling), which should be fine.

PR #6592, opened 2021-02-24
containerd: fix mount issues with certain images

### LINKED ISSUE #6578: Switching to containerd runtime breaks email resource
## Summary
Last week we updated to 7.0.0 what in general worked like a charm  👍🏻.
I also switched to the `containerd` runtime and initially it looked like everything still worked, what was true for 99% of the pipes but later on I realized that the email resource we are using a lot (in `put` steps) broke.

My initial thought was that this is a problem of the email resource (https://hub.docker.com/r/pcfseceng/email-resource) and stated in issue there:

https://github.com/pivotal-cf/email-resource/issues/41

Actually - it was the point in time that I remembered opening up this quite a while before, as upgrading to `containerd` earlier already failed with the same issue.

I tried to get a glimpse on what is going on but failed building the resource on my own (I am not a Go dev).

Thinking about this a bit more I came to the conclusion that this might not be an issue of the resource but rather with Concourse itself. Running the following command `docker run -it pcfseceng/email-resource:1.0.19` you can see that there is no file `/tmp/gdn-init` present in the container.

So I guess this is something special on how concourse bootstraps the containers. Digging around Google and other Concourse issues I could not find something meaningful, still there is some code in Concourse that deals with that file:

https://github.com/concourse/concourse/search?q=%2Ftmp%2Fgdn-init

Any help on this is really appreciated. Sounds crazy but the feature set of the offending email resource is very important to us.

## Steps to reproduce
Starting from `docker-compose up`, what steps can be followed to reproduce the
problem?

Did not actually try this but implementing a simple pipe using that resource should be sufficient to examine the problem.

## Expected results
The email resource still works in the pipeline as it should not depend on the container runtime

## Actual results
The email resource cannot be used any more.
```
run check step: run check step: find or create container: starting task: new task: OCI runtime create failed: container_linux.go:370: starting container process caused: exec: "/tmp/gdn-init": stat /tmp/gdn-init: no such file or directory: unknown
```

## Additional context
Garden runtime is scheduled to be dropped in 7.1. We heavily rely on the given resource in our pipes and there is no alternative out there. This might become a nasty one for us, as I am trying to keep our cluster up-to-date.

## Triaging info
* Concourse version: 7.0.0
* Did this used to work?: YES (with garden runtime)
* OS: Debian Buster

aoldershaw:

closes #6578 (more context in the issue comments)

* Remove the `/run` mount - it's neither mounted by Guardian nor Docker (from what I can tell)
* Mount `/tmp/gdn-init` after all the other default mounts
  * The bug was fixed by removing `/run`, but would have also been fixed by mounting `/tmp/gdn-init` last
  * I figure that, while not strictly necessary for this bug fix, mounting it after the other mounts is a bit safer since it mirrors how we mount the other "normal" mounts (regular filesystem mounts, e.g. task inputs)

## Release Note

* Fix an issue on the containerd runtime where processes fail to run with certain container images

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/worker/runtime/spec/mounts.go
+++ b/worker/runtime/spec/mounts.go
@@ -44,15 +44,9 @@ var (
 		},
 		{
 			Destination: "/sys/fs/cgroup",
-			Type: "cgroup",
-			Source: "cgroup",
-			Options: []string{"ro", "nosuid", "noexec", "nodev"},
-		},
-		{
-			Destination: "/run",
-			Type:        "tmpfs",
-			Source:      "tmpfs",
-			Options:     []string{"nosuid", "strictatime", "mode=755", "size=65536k"},
+			Type:        "cgroup",
+			Source:      "cgroup",
+			Options:     []string{"ro", "nosuid", "noexec", "nodev"},
 		},
 	}
 )
--- a/worker/runtime/spec/mounts.go
+++ b/worker/runtime/spec/mounts.go
@@ -52,17 +52,14 @@ var (
 )
 
 func ContainerMounts(privileged bool, initBinPath string) []specs.Mount {
-	mounts := append(
-		[]specs.Mount{
-			{
-				Source:      initBinPath,
-				Destination: "/tmp/gdn-init",
-				Type:        "bind",
-				Options:     []string{"bind"},
-			},
-		},
-		DefaultContainerMounts...,
-	)
+	mounts := make([]specs.Mount, 0, len(DefaultContainerMounts)+1)
+	mounts = append(mounts, DefaultContainerMounts...)
+	mounts = append(mounts, specs.Mount{
+		Source:      initBinPath,
+		Destination: "/tmp/gdn-init",
+		Type:        "bind",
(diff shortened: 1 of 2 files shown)


muntac (approved):
Mounting after all the others makes sense to me. I went down a bit of a rabbit hole trying to figure out why we we might want `/run` and looking into the `progrium/busybox` rootfs.

I suspect we got the idea from [containerd](https://github.com/containerd/containerd/blob/d8527867057daa93c17423ec7cd3438696b1c899/oci/spec.go#L202). But it's not easy to trace the git blame for that to a cause.

Anyway, seems we never had a problem with it not being mounted with Guardian so should be good to go like this.

PR #6619, opened 2021-03-01
Static vars should only resolve on "" sources

### LINKED ISSUE #6543: var interpolation ignores var source
## Summary

When setting a pipeline using static vars (including instances vars), var references to a var source's variable with the same name as the interpolated var are incorrectly interpolated

## Steps to reproduce

```
$ cat pipeline.yml
jobs:
- name: print
  plan:
  - across:
    - var: my-var
      values: [from local var source]
    task: print
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: busybox}
      run:
        path: echo
        args:
        - ((my-var))
        - ((.:my-var))

$ fly -t dev set-pipeline -p test -c pipeline.yml -v my-var="from static var"
$ fly -t dev trigger-job -j test/print -w
```

## Expected results

Prints:

```
from static var from local var source
```

## Actual results

```
from static var from static var
```

## Additional context

This happens because `vars.StaticVariables.Get` ignores `ref.Source`: https://github.com/concourse/concourse/blob/ab76fa8cf13879020112e0c18d56043fb1e1f114/vars/static_vars.go#L7

## Triaging info

* Concourse version: 7.0.0
* Browser (if applicable):
* Did this used to work? not sure

chenbh:

closes #6543 

## Release Note

- Fixed bug where static vars from `fly set-pipeline -v ... -y ...` were interpolated into local vars `((.:var))`

## Contributor Checklist
- [X] [Signed] all commits
- [X] Added release note (Optional)

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/vars/named_vars.go
+++ b/vars/named_vars.go
@@ -12,7 +12,7 @@ func (m NamedVariables) Get(ref Reference) (interface{}, bool, error) {
 	}
 
 	if vars, ok := m[ref.Source]; ok {
-		return vars.Get(ref)
+		return vars.Get(ref.WithoutSource())
 	}
 
 	return nil, false, MissingSourceError{Name: ref.String(), Source: ref.Source}
--- a/atc/exec/build_vars.go
+++ b/atc/exec/build_vars.go
@@ -32,7 +32,7 @@ func newBuildVariables(credVars vars.Variables, enableRedaction bool) *buildVari
 func (b *buildVariables) Get(ref vars.Reference) (interface{}, bool, error) {
 	if ref.Source == "." {
 		b.lock.RLock()
-		val, found, err := b.localVars.Get(ref)
+		val, found, err := b.localVars.Get(ref.WithoutSource())
 		b.lock.RUnlock()
 		if found || err != nil {
 			return val, found, err
--- a/vars/static_vars.go
+++ b/vars/static_vars.go
@@ -5,6 +5,10 @@ type StaticVariables map[string]interface{}
 var _ Variables = StaticVariables{}
 
 func (v StaticVariables) Get(ref Reference) (interface{}, bool, error) {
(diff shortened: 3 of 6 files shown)


vars/named_vars.go
```diff
@@ -12,7 +12,7 @@ func (m NamedVariables) Get(ref Reference) (interface{}, bool, error) {
 	}
 
 	if vars, ok := m[ref.Source]; ok {
-		return vars.Get(ref)
+		return vars.Get(ref.WithoutSource())
```
clarafu, line 15: Just wondering if it is necessary to remove the source from this `vars.Get` call?

chenbh: Since `NamedVariables` keeps track of multiple other implementations, `Source` is actually only meaningful to `NamedVariables` and nothing else. 

Let's say during a build we end up with a build var that looks like this:
```
NamedVar {
  "foo": StaticVar{
    "secret": "value"
  }
  "bar": Tracker{
    Vault: ....
  }
}
```

If we try to get `((foo:secret))`, it should pass `((secret))` to the StaticVar (as part of this PR, StaticVar will actually return false if the source is non-empty). Likewise if we try `((bar:path.field))`, it'll only pass `((path.field))` to Tracker/Vault.

clarafu: Just to make sure I completely understand, this change is purely refactoring and has no real user impact changes/behavioural changes right? Because the `NamedVariables` is only used for var source implementations so it will never contain a `StaticVar` which is used for build variables.

clarafu (approved):
COOL. Tested it, it works 👍

Issue #6206, opened 2020-10-26
limit-active-task: 2 tasks can land on the same worker


aliculPix4D:
## Summary

2 (or more) tasks can land on the same worker even though we selected limit-active-task container placement strategy option where max-active-task-per-worker is set to 1.

This bug is probably introduced after this refactoring commit:
https://github.com/concourse/concourse/commit/11a0216e1ef7c8094e634caec12784aeca16a38d

In my understanding two conditions are needed for bug to occur:
(`chooseTaskWorker` method in `atc/worker/client.go`)
1) `chosenWorker = client.pool.FindOrChooseWorkerForContainer` for the second task needs to happen before first task increases the number of active tasks on the same worker. In this case choose worker is not nil for the second task.
2) second task tries to acquire lock only after the first task released it already. Otherwise, we would sleep for 1 second , first task would already increased the number of active tasks and the bug would never occur

Bug occurs very rarely. For example, in the last three months we only experienced it around 15 times in our infra.

![bug_concourse](https://user-images.githubusercontent.com/50741304/97153508-e758d000-1772-11eb-886a-c541c7f4966b.png)

## Steps to reproduce

Add:
``` 
CONCOURSE_CONTAINER_PLACEMENT_STRATEGY: limit-active-tasks
CONCOURSE_MAX_ACTIVE_TASKS_PER_WORKER: 1
```

to your `docker-compose.yml` and run: 

```
docker-compose \                                                 
  -f ./docker-compose.yml \
  -f ./hack/overrides/prometheus.yml \
 up -d
```
Create a small pipeline template (`run_job.yml`) i.e.:
```
---
jobs:
- name: loop
  plan:
  - task: loop
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: 
          repository: busybox
      run:
        path: sh
        args:
          - -c
          - |
            echo "Executing on worker: `hostname`"
            for i in `seq 1 30`; do sleep 1; echo "Slept for ${i} seconds."; done

```
and executed the following shell script:

```
#!/bin/bash -ex

for i in `seq 1 60`; do fly -t ci sp -p parallel${i}-second-linux -c run_job.yml -n && fly -t ci up -p parallel${i}-second-linux; done

task(){
   fly -t ci tj -j parallel${i}-second-linux/loop;
}

for i in `seq 1 60`; do 
  task "$i" &
done
```

## Expected results

At all times, only a single task can land on the same worker (if max-active-task-per-worker is set to 1).

## Actual results

[...]

PR #6635, opened 2021-03-04
Wait for worker matching strategy when scheduling build steps

### LINKED ISSUE #6648: All workers busy when using container placement strategy limit-active-tasks in 7.0.0
## Summary
We are running a concourse environment with 1 web node and 3 workers using a web docker-compose config, which limits max running task per container to 1. One of the workers is using a tag to limit tasks that can be run on it.
```
      CONCOURSE_CONTAINER_PLACEMENT_STRATEGY: limit-active-tasks
      CONCOURSE_MAX_ACTIVE_TASKS_PER_WORKER: 1
```
We are experiencing an issue that didn't appear before updating to version 7.0.0 - workers can randomly become busy with some unknown tasks (nothing visible on a web UI), and no other job can be started:
```
All workers are busy at the moment, please stand-by.
```
This can happen on any worker. In particular, when it does happen on the one that uses a tag, it causes all jobs that require that tag to starve on the message shown above.
We have checked that required workers are indeed online (state: running).
```
fly -t ci workers
name                 containers  platform  tags  team  state    version  age
26a773549c67         46          linux     none  none  running  2.3      9d
Jans-Mac-mini.local  0           darwin    jan   none  running  2.3      7d
f0e17b5530aa         39          linux     none  none  running  2.3      10d
```
We have also checked using `fly -t ci builds` that there are no other blocking jobs. Only a single item, which is not a "check" resource is being shown:
```
1379786  some-project/some-step/114  started    2021-03-08@17:24:57+0100  n/a                       7m27s+    main  user
```
## Steps to reproduce
Run 1 web node, 1 worker node, use the limit-active-tasks and 1 task per worker. After some time the worker becomes locked, and nothing can be runned on it. Restarting the web node and/or the worker node does not help.
## Expected results
When there is no job running on the given worker, it should be possible to run a job on it, using limit-active-tasks and 1 task per worker.
## Actual results
Web node says that `All workers are busy at the moment, please stand-by.`, and no job can be run on the given worker.
## Additional context
We are using garden, because containerd caused issues described in https://github.com/concourse/concourse/issues/6613.
## Triaging info
* Concourse version: 7.0.0
* Did this used to work? Yes, has been working on all previous 6.x.x versions

multimac:

Fixes #6648

Similar to how the `task` step would wait for a worker to become available when using the `limit-active-tasks` strategy, this expands that functionality to all strategies (like the newly introduced `limit-active-containers` and `limit-active-volumes`), as well as applying it to all build step types (ie. `check/get/put/task`). Additionally, this PR adds a `waiting-for-worker` event to notify the user when a build is stuck waiting for a worker to become available.

## Notes to reviewer:
The existing `limit-active-tasks` strategy was a bit tricky to keep around in its current form, so I had to rewrite it and that has altered its behavior slightly. Before, it would lock and ensure that the number of active tasks never went above the configured limit. After the rewrite, `limit-active-tasks` now behaves more like `limit-active-containers` / `limit-active-volumes` in that it tries to cap the number of active tasks at the configured limit but may go over slightly due to race conditions.

Also I still need to add some additional tests around the `WaitForWorker` method, but wanted to open this PR to get some initial feedback on the implementation and also the tweaks to `limit-active-tasks` mentioned above.

## Release Note
* Previously, if no workers satisfied the container placement strategy for a step (with the exception of `task` steps when using the `limit-active-tasks` placement strategy), the step would simply error the build
* Now, all steps will wait for a worker to become available
* The metric `concourse_tasks_waiting` was removed and replaced with `concourse_steps_waiting{type="task"}`

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


aoldershaw:
Hi @multimac, thanks for the PR! Seems very useful.

I do have some concerns about removing the locking around limit-active-tasks, though - it would introduce a regression to https://github.com/concourse/concourse/issues/6206. While it's true that it's more consistent with limit-active-{containers,volumes}, I think it's likely more impactful since, depending on the workload, having a couple more containers around is much less impactful than having a couple more very expensive tasks. If needed, I think we'd be better off just adding locking to limit-active-{containers,volumes} rather than removing it from limit-active-tasks.

One alternative approach to the one taken that would allow for specialized behaviour for task steps (while still allowing for other steps to wait for a worker) is to add a `SelectWorker` (or `WaitForWorker`) method to the `engine.BuildStepDelegate`. `engine.TaskStepDelegate` could have its own specialized version (as we currently have) that increments/decrements active tasks (under a lock), while `engine.BuildStepDelegate` can have a more general version (like `worker.Pool.WaitForWorker` under your implementation. `engine.TaskStepDelegate` could shell out to `engine.BuildStepDelegate` when `limit-active-tasks` is not set, e.g.

```go
func (d *taskDelegate) WaitForWorker(...) (worker.Client, error) {
    if !strategy.ModifiesActiveTasks() {
        return d.BuildStepDelegate.WaitForWorker(...)
    }
    ...
}
```

What do you think about this approach? More than happy to chat about alternative approaches

multimac:
Hey @aoldershaw, thanks for the feedback.

That's fair. What do you think of the commit linked below?
https://github.com/multimac/concourse/commit/302fe0d246db94af5d7a824026cfdf01c49dd5a6

I was thinking about it a bit more overnight and thought we could make `worker.Pool.SelectWorker` responsible for calling `engine.BuildStepDelegate.SelectedWorker` (under a lock). This way, `engine.TaskStepDelegate` could put the logic to increment the active tasks in that method, and this would saves the waiting logic from needing to be duplicated in a method like `engine.BuildStepDelegate.WaitForWorker`.

aoldershaw:
@multimac I'm a bit worried about acquiring a global lock for every step - on large deployments, this would result in a lot of lock contention. Frankly, grabbing a global lock for every `task` also probably results in a fair bit of lock contention, but I would imagine far less.

We can possibly remove the need for the `limit-active-tasks` lock by leveraging atomicity in the database. Maybe we could select a list of all viable candidates increment the active tasks count for all of them atomically, and after we select a worker, somehow decrement all of the candidates that we did not select.

i.e. something like
```sql
UPDATE workers
SET active_tasks = active_tasks + 1
WHERE name IN (...candidates...)
  AND active_tasks < ...max_active_tasks...
```
and then after a worker is selected:
```sql
UPDATE workers
SET active_tasks = active_tasks - 1
WHERE name IN (...candidates...)
  AND name <> ...selected worker...
```

Then, `engine.TaskStepDelegate` is just responsible for decrementing the active tasks upon `Finished`. WDYT?

---

re: container and volume counts, that's a bit different - they aren't incremented/decremented in the same way. Currently, `limit-active-{containers,volumes}` takes the current number of containers/volumes from the last heartbeat from the worker. That is, the worker periodically (every 30s or so by default I think) pings the ATC saying "I'm still alive, and I have these containers and volumes" - the number of active containers/volumes is based on this heartbeat. However, #6469 improves upon this for containers by fetching the number of containers on-the-fly from the `containers` table (when a new container is created, it's first added to the `containers` table, and when it gets GC'd, it's removed from the `containers` table). But either way, we don't have a count we need to increment/decrement.

---

Oh, and side note - I realize this PR changes of the behaviour of `limit-active-tasks`. Previously, it had two responsibilities:
1. Choose the worker(s) with the fewest active tasks, and
2. (optionally) Filter out workers that have too many active tasks

Now, it's only doing 2, which I think makes more sense honestly. However, maybe we should add in a `fewest-active-tasks` strategy that only does 1 (and wouldn't necessarily need locking/atomicity, since it's less critical that you don't exceed some limit). This would be consistent with `limit-active-containers` and `fewest-build-containers`, for instance

multimac:
Yeah, that does sound a bit gross now you mention it. Instead of bumping the active task count for all containers, could we use `ContainerPlacementStrategyChainNode.Pick` to increase the active task count on the worker that is being checked and check that it isn't then over the configured limit?

I've pushed up a new commit https://github.com/concourse/concourse/pull/6635/commits/5f24ab265b8e7ca59207744112edcd4b7bf707a3 tries that out. I had to add a `Release` method which then decrements the active task count (similar to how the `taskStep` did that. This makes tracking the active task count purely the responsibility of the "limit-active-tasks" placement strategy.

---

Regarding the sorting of active containers, I think I removed this with the intent to add it back in but must have forgotten. I've put it back in in the commit mentioned above though.

In that commit I also removed the logic from the `fewest-build-containers` and `volume-locality` placement strategies where they would only return a single worker (that which had the fewest build containers or most volumes already present). Instead they now sort the list of candidates and allow later steps to further sort the list or reject candidate workers.

I imagine there could be a scenario currently where if you configured a strategy chain like `volume-locality,limit-active-tasks` then the `volume-locality` strategy could return the workers with the most volumes present but all these workers could fail the `limit-active-tasks` strategy, resulting in an error in placement of the container. I still need to fix all the tests in this PR, but I'll keep this in mind and make sure to add a test for this case.

atc/worker/placement.go
```diff
@@ -196,34 +258,51 @@ func newLimitActiveTasksPlacementStrategy(name string, maxTasks int) ContainerPl
...
+
+	if err != nil {
+		return err
+	}
 
-		workersByWork[activeTasks] = append(workersByWork[activeTasks], w)
-		if minActiveTasks == -1 || activeTasks < minActiveTasks {
-			minActiveTasks = activeTasks
+	// If the step is not a task or maxTasks == 0, ignore the number of active tasks and distribute the work evenly
+	if spec.Type == db.ContainerTypeTask && (strategy.maxTasks == 0 || activeTasks < strategy.maxTasks) {
```
aoldershaw, line 289: Should this condition be reversed? i.e.

```
if spec.Type != db.ContainerTypeTask || (strategy.maxTasks > 0 && activeTasks > strategy.maxTasks) {
```

Also, thoughts on checking the container type before calling `IncreaseActiveTasks`? i.e. something like:
```go
	if spec.Type != db.ContainerTypeTask {
		return nil
	}

	activeTasks, err := worker.IncreaseActiveTasks()
	if err != nil {
		return err
	}

	if strategy.maxTasks > 0 && activeTasks > strategy.maxTasks {
		// decrease and return error
		return ...
	}
	return nil
```

that way, we aren't increasing the active tasks count when we certainly don't need to

atc/worker/pool.go
```diff
@@ -202,14 +229,120 @@ dance:
...
+
 		if err != nil {
 			return nil, err
 		}
+
+		for _, candidate := range candidates {
+			err := strategy.Pick(logger, candidate, containerSpec)
+
+			if err != nil {
+				logger.Debug("Candidate worker rejected during selection", lager.Data{"reason": err.Error()})
```
aoldershaw, line 242:
```suggestion
				logger.Debug("candidate-worker-rejected-during-selection", lager.Data{"reason": err.Error()})
```

atc/worker/pool.go
```diff
@@ -41,7 +47,29 @@ type Pool interface {
 		ContainerSpec,
```
aoldershaw, line 47: Can we remove `SelectWorker` from the interface now that `WaitForWorker` is the entrypoint?

atc/worker/pool.go
```diff
@@ -202,14 +229,120 @@ dance:
...
+	var worker Client
+	var waiting bool = false
+	for {
+		var err error
+		worker, err = pool.SelectWorker(ctx, owner, containerSpec, workerSpec, strategy, callbacks)
+
+		if err != nil {
+			if errors.Is(err, ErrNoWorkers) {
+				// Could use these blocks to notify caller of the reason we're waiting
+			} else if errors.Is(err, ErrFailedToAcquireLock) {
```
aoldershaw, line 295: Can we get rid of this error case (and error?) now?

atc/worker/placement.go
```diff
@@ -196,34 +258,51 @@ func newLimitActiveTasksPlacementStrategy(name string, maxTasks int) ContainerPl
...
 	}
 
-	return workersByWork[minActiveTasks], nil
+	return nil
 }
 
-func (strategy *LimitActiveTasksPlacementStrategyNode) ModifiesActiveTasks() bool {
-	return true
+func (strategy *LimitActiveTasksPlacementStrategyNode) Release(logger lager.Logger, worker Worker) {
+	_, err := worker.DecreaseActiveTasks()
```
aoldershaw, line 302: this should take the `spec` in too, since we only want to decrease active tasks if `spec.Type == db.ContainerTypeTask`

aoldershaw (changes requested):
I think your approach with `Order` and `Pick` makes things a lot nicer! I've left some suggestions

multimac:
Awesome! Thanks for all the feedback. I don't know why my Google-fu didn't stumble over `rand.Shuffle` before but thanks for the heads up about it. I'm working on another commit which fixes up all the tests and addresses most of the points you brought up (still need to finish up the work).

One quick note, I noticed this comment (by you, coincidentally)...
https://github.com/concourse/concourse/blob/5f24ab265b8e7ca59207744112edcd4b7bf707a3/atc/worker/pool.go#L114-L118
I think I could restructure the code a bit to check for a valid team worker first and then, if all team-specific workers are rejected by the strategy, fall back to checking the general worker pool.

Do you have any thoughts on this? Would it maybe be out of scope of this PR? I can imagine a scenario where a team is assigned a single worker and when that worker fails the configured strategy then their jobs start failing (or stalling once this PR is merged).

multimac:
Can use this pipeline to test that the steps wait for a worker to be returned by the configured strategy, instead of erroring out like before. Tested using the following environment variables to set a limit on max active tasks...
* `CONCOURSE_CONTAINER_PLACEMENT_STRATEGY=limit-active-tasks,volume-locality,fewest-build-containers`
* `CONCOURSE_MAX_ACTIVE_TASKS_PER_WORKER=1`

```
---
jobs:
- name: job
  public: true
  plan:
  - across:
    - var: count
      values: ["1", "2", "3", "4", "5", "6", "7", "8", "9", "0"]
      max_in_flight: all
    do:
    - task: simple-task
      config:
        platform: linux
        image_resource:
          type: registry-image
        run:
          path: sh
          args:
          - -exc
          - |
            echo Waiting 5 seconds...
            sleep 5

            echo Hello World
```

atc/metric/periodic.go
```diff
@@ -181,11 +181,11 @@ func tick(logger lager.Logger, m *Monitor) {
...
+	for labels, gauge := range m.StepsWaiting {
 		m.emit(
-			logger.Session("tasks-waiting"),
+			logger.Session("steps-waiting"),
 			Event{
-				Name:  "tasks waiting",
+				Name:  "steps waiting",
 				Value: gauge.Max(),
 				Attributes: map[string]string{
 					"teamId":     labels.TeamId,
```
aoldershaw, line 191: We should add the `type` attribute here

atc/worker/pool.go
```diff
@@ -174,40 +286,86 @@ func (pool *pool) SelectWorker(
...
-		worker, err = strategy.Choose(logger, compatibleWorkers, containerSpec)
-		if err != nil {
-			return nil, err
+		select {
+		case <-ctx.Done():
+			logger.Info("aborted-waiting-for-worker")
+			return nil, 0, ctx.Err()
+		case <-pollingTicker.C:
+		case <-pool.waker:
+			break
```
aoldershaw, line 340: Can we get rid of this `break`? At first glance, I was wondering why it'd behave differently from `pollingTicker.C`, but then realized it doesn't since the `break` is a no-op (it just breaks out of the `select`, which it would do anyway)

aoldershaw (changes requested):
I really like what you've done, but have a few small comments! Thanks for your patience!

multimac:
No worries, thanks for all your help reviewing this. This is the most Go code I've written to date 😅 Just hope it works as expected.

Is there any documentation on how to build Concourse (specifically the web component) for production? Given this PR seems to be getting close to completion I'm thinking it might be worth deploying these changes to my company's Concourse instance and seeing if that shakes any bugs out.

atc/worker/placement.go
```diff
@@ -27,89 +34,176 @@ func (err NoWorkerFitContainerPlacementStrategyError) Error() string {
...
+		node.Release(logger, worker, spec)
+	}
 }
 
 func NewRandomPlacementStrategy() ContainerPlacementStrategy {
 	s, _ := NewContainerPlacementStrategy(ContainerPlacementStrategyOptions{ContainerPlacementStrategy: []string{"random"}})
 	return s
 }
 
+// Strategy which orders candidate workers based off the number of volumes which alread
```
odormond, line 205: Typo:
```suggestion
// Strategy which orders candidate workers based off the number of volumes which already
```

atc/worker/placement_test.go
```diff
@@ -15,796 +16,761 @@ import (
...
+					Expect(pickedWorker).To(Equal(workers[1]))
+				})
+			})
+
+			Context("when no workers are under the limit", func() {
+				BeforeEach(func() {
+					limit = 10
+				})
+
+				It("fails to pick workers with an equal or higher number of tasks", func() {
```
odormond, line 582: Copy-paste typo:
```suggestion
				It("fails to pick workers with an equal or higher number of containers", func() {
```

atc/worker/placement_test.go
```diff
@@ -15,796 +16,761 @@ import (
...
-					Consistently(func() Worker {
-						chosenWorker, chooseErr = strategy.Choose(
-							logger,
-							workers,
-							spec,
-						)
-						Expect(chooseErr).ToNot(HaveOccurred())
-						return chosenWorker
-					}).Should(Equal(compatibleWorker1))
+				It("orders workers by active container count", func() {
```
odormond, line 641:
```suggestion
				It("orders workers by active volume count", func() {
```

atc/worker/placement_test.go
```diff
@@ -15,796 +16,761 @@ import (
...
-			TeamID: 4567,
+			Context("when limit is zero", func() {
+				BeforeEach(func() {
+					limit = 0
+				})
 
-			Inputs: []InputSource{},
-		}
-	})
+				It("is able to pick and release the first worker, regardless of active containers", func() {
```
odormond, line 691:
```suggestion
				It("is able to pick and release the first worker, regardless of active volumes", func() {
```

atc/worker/placement_test.go
```diff
@@ -15,796 +16,761 @@ import (
...
+			Context("when limit is non-zero", func() {
+				BeforeEach(func() {
+					limit = 100
+				})
 
-			someWorker1.BuildContainersReturns(30)
-			someWorker2.BuildContainersReturns(20)
-			someWorker3.BuildContainersReturns(10)
-		})
+				It("fails to pick workers with an equal or higher number of containers", func() {
```
odormond, line 701:
```suggestion
				It("fails to pick workers with an equal or higher number of volumes", func() {
```

atc/worker/placement_test.go
```diff
@@ -15,796 +16,761 @@ import (
...
-				)
-				Expect(chooseErr).ToNot(HaveOccurred())
-				return chosenWorker
-			}).Should(Equal(someWorker3))
+			Context("when no workers are under the limit", func() {
+				BeforeEach(func() {
+					limit = 10
+				})
+
+				It("fails to pick workers with an equal or higher number of tasks", func() {
```
odormond, line 711:
```suggestion
				It("fails to pick workers with an equal or higher number of volumes", func() {
```

odormond (commented):
Great work! I'm looking forward to see this released.

I've pointed out a few copy-paste typos I noticed.

aoldershaw (changes requested):
Almost done! Just some cleanup now, and a few small bugs I noticed!

Re: building the Concourse image - it's all do-able, but we don't really have any documentation for it (besides the pipeline itself - https://github.com/concourse/ci/blob/master/pipelines/concourse.yml - but that does a lot more).

Once we squash all the remaining bugs that we've found, I'm cool to just merge it, which will create a production-ready image. We'll definitely deploy to our own cluster, and I'm happy to ping you when it's ready to go - the image is built and pushed here: https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/build-image/builds/415

Otherwise, if you'd feel better about deploying before merging, we can set up a pipeline to build an image based off of this branch - we've done it before, not too much trouble. Up to you!

atc/worker/placement.go
```diff
@@ -27,89 +34,176 @@ func (err NoWorkerFitContainerPlacementStrategyError) Error() string {
...
+
+	// Attempts to pick the given worker to run the specified container
+	Pick(lager.Logger, Worker, ContainerSpec) error
+
+	// Releases any resources acquired by any configured strategies as part of
+	// picking the candidate worker
+	Release(lager.Logger, Worker, ContainerSpec)
 }
 
 type ContainerPlacementStrategyChainNode interface {
```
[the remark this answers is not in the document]
aoldershaw: The implications of that would be that the individual strategies and the collective multi-strategy would all abide by the same interface, which feels a bit nicer

multimac:
Regarding the container image, I'm happy to wait until the normal CI pipelines pump out a production-ready image. Would be a bit easier than trying to push a custom container image to our cluster.

atc/exec/check_step.go
```diff
@@ -305,18 +305,29 @@ func (step *CheckStep) runCheck(
 
 	defer cancel()
 
-	chosenWorker, err := step.workerPool.SelectWorker(
+	chosenWorker, _, err := step.workerPool.SelectWorker(
```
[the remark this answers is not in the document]
aoldershaw: pretty much good to go, looks like the `atc/exec` tests need updating, though: https://ci.concourse-ci.org/builds/114970483#L605a440c:154:206

atc/worker/pool.go
```diff
@@ -151,16 +142,23 @@ func (pool *pool) findWorkerFromStrategy(
...
+		if err == nil {
 			return candidate, nil
 		}
+
+		formattedError := fmt.Errorf("worker: %s, error: %v", candidate.Name(), err)
+		if combinedError == nil {
+			combinedError = formattedError
+		} else {
+			combinedError = fmt.Errorf("%v\n%v", combinedError, formattedError)
+		}
```
aoldershaw, line 158: Not asking you to change it, but just for future reference, you can replace this with 
```go
combinedError = multierror.Append(combinedError, formattedError)
```

using [hashicorp/go-multierror](https://github.com/hashicorp/go-multierror) (which we already happen to depend on)

aoldershaw (approved):
Awesome work @multimac! Thanks again! Will let you know when this makes its way through the pipeline

atc/worker/placement.go
```diff
@@ -27,269 +33,400 @@ func (err NoWorkerFitContainerPlacementStrategyError) Error() string {
...
-	for _, w := range workers {
-		if strategy.maxContainers == 0 || w.ActiveContainers() <= strategy.maxContainers {
-			candidates = append(candidates, w)
-		}
+	for _, worker := range candidates {
+		counts[worker] = worker.ActiveContainers()
 	}
 
+	sort.SliceStable(candidates, func(i, j int) bool {
+		return counts[candidates[i]] < counts[candidates[j]]
```
aoldershaw, line 371: Consider the following placement strategy:
```
limit-active-containers,volume-locality
```

with max active containers is set to 100.

`worker1` has 50 containers, and `worker2` has 60 containers.
`worker1` has 1 volume locally, and `worker2` has 2 volumes locally.

Prior to this PR, Concourse would schedule on `worker2` - since both workers are both under the container limit, we resort to effectively only considering `volume-locality`. Since the `limit-active-containers` sort now takes precedence (which I think is what we want in general - the further up the chain, the more precedence), we'll pick `worker1` since it has fewer containers locally.

I think this can be fixed by not sorting on the number of containers, but sorting on whether the worker exceeds the number of containers or not (since `limit-active-containers` doesn't care about giving you the worker with the fewest containers active - that's what `fewest-build-containers` is for). e.g. something like:

```go
	exceedsThreshold := func(count int) int {
		if strategy.maxContainers > 0 && count > strategy.maxContainers {
			return 1
		} else {
			return 0
		}
	}
	sort.SliceStable(candidates, func(i, j int) bool {
		return exceedsThreshold(counts[candidates[i]]) < exceedsThreshold(counts[candidates[j]])
	}
```

So, we consider all counts less than the threshold to be 0, and all counts greater to be 1 - this way, we partition the slice while maintaining the original ordering.

**Note:** The same applies for `limit-active-volumes`, but **not** for `limit-active-tasks` (it's a weird one that both filters based on exceeding the threshold AND tries to give you the worker with the fewest tasks)

aoldershaw (changes requested):
I'm sorry 😬 , one more thing I noticed

atc/worker/placement.go
```diff
@@ -69,18 +74,30 @@ func NewChainPlacementStrategy(opts ContainerPlacementStrategyOptions) (*ChainPl
...
 		case "limit-active-tasks":
 			if opts.MaxActiveTasksPerWorker < 0 {
 				return nil, errors.New("max-active-tasks-per-worker must be greater or equal than 0")
 			}
 			cps.nodes = append(cps.nodes, newLimitActiveTasksStrategy(strategy, opts.MaxActiveTasksPerWorker))
 
+			// For backwards compatibility, "limit-active-tasks" implies "fewest-active-tasks"
+			cps.nodes = append(cps.nodes, newFewestActiveTasksStrategy(strategy))
+
+		case "fewest-active-containers":
```
aoldershaw, line 89: Would it be possible to revert most of the changes in this commit, and instead go with the partitioning approach in https://github.com/concourse/concourse/pull/6635#discussion_r600459319? The changes make sense, but I'd like to keep the scope of this PR limited. We'll need to think about the placement strategies a little more, first - e.g. `fewest-active-containers` is pretty much just `fewest-build-containers` now

aoldershaw (approved):
Thanks again (again)! Hopefully the review process wasn't *too* painful, but I really appreciate all the work you put into addressing my concerns in a timely manner. Let's get this merged and see if we can weed out any other issues

aoldershaw:
FYI @multimac the latest `concourse/concourse-dev` image has these changes in case you're looking to try it out (digest `sha256:fb41d72aa824e06ec712ca9f9adfd92a2ec25de99f09d576e1fd2484cc8a4d70` for alpine, `sha256:e4070a79dc90196b5463b454f1577a726da1c2a0a42a36b0fa75541330852f21` for ubuntu) - https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/build-image/builds/419

multimac:
No worries. Thanks for your help reviewing the changes. And the review process wasn't painful at all, was actually a very good experience. Hopefully I'll be able to do up some more PRs for Concourse in the future.

And awesome! Didn't get a chance to throw them in our cluster at work today, but will hopefully get the chance to tomorrow. Really keen to try out this functionality and hopefully be able to downscale our pool of workers a little bit.

aoldershaw:
Awesome! By the way, no pressure, but if you are interested in submitting more PRs in the future, you may want to add yourself to the contributors team by submitting a PR to [concourse/governance](https://github.com/concourse/governance) (for example, [this one](https://github.com/concourse/governance/pull/29/files)).

It doesn't change too much, but grants you access to the [PRs pipeline](https://ci.concourse-ci.org/teams/contributor/pipelines/prs?vars.branch=%22master%22) (so you can rerun flaky builds, etc.), and access to the `#dev` channel on Discord

PR #6652, opened 2021-03-09
runtime: check if swap limits is enabled

### LINKED ISSUE #6649: Don't write to memory.memsw.limit_in_bytes when memswap cgroup is disabled on the kernel
## Summary

The [testflight container limits tests](https://github.com/concourse/concourse/blob/master/testflight/container_limits_test.go) is skipped whenever the memswap cgroup is disabled on the kernel.

This test should pass regardless of this cgroup being disabled or not. What can be conditional is whether the test checks `memory.memsw.limit_in_bytes` or `memory.limit_in_bytes` to verify that container limits are set. It can probably just check `memory.limit_in_bytes`.

## Context

* Prior discussion: #6512

taylorsilva:

closes #6649

The containerd runtime has been updated to check if swap limits is
enabled and will conditionally set them if they're enabled.

These testflight tests will always fail for Guardian if swap limits are
disabled on the host, unless Guardian is told to not set the swap limit.

## Notes to reviewer:
This depends on using runc-93. Anything earlier will cause the tests to fail when they should be skipped, only for the Guardian runtime though.

## Release Note

* The containerd runtime will conditionally set memory swap limits if it detects that memory swap limits are enabled

## Contributor Checklist
- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

cc @EstebanFS

### DIFF
--- a/testflight/container_limits_test.go
+++ b/testflight/container_limits_test.go
@@ -1,7 +1,7 @@
 package testflight_test
 
 import (
-	"strings"
+	"regexp"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -16,8 +16,11 @@ var _ = Describe("A job with a task that has container limits", func() {
 		watch := spawnFly("trigger-job", "-j", inPipeline("container-limits-job"), "-w")
 		<-watch.Exited
 
+		match, _ := regexp.MatchString(`open /sys/fs/cgroup/memory/.*/memory\.memsw\.limit_in_bytes: (permission denied)?(no such file or directory)?`,
+			string(watch.Out.Contents()))
+
 		// Guardian runtime will always fail this test unless it's explicitly told not to set the swap limit
-		if strings.Contains(string(watch.Out.Contents()), "memsw.limit_in_bytes: no such file or directory") {
+		if match {
 			Skip("swap limits not enabled; skipping")
 		}
 
@@ -31,8 +34,11 @@ var _ = Describe("A job with a task that has container limits", func() {
 		watch := spawnFly("trigger-job", "-j", inPipeline("container-limits-failing-job"), "-w")
 		<-watch.Exited
 
+		match, _ := regexp.MatchString(`open /sys/fs/cgroup/memory/.*/memory\.memsw\.limit_in_bytes: (permission denied)?(no such file or directory)?`,
+			string(watch.Out.Contents()))
+
 		// Guardian runtime will always fail this test unless it's explicitly told not to set the swap limit
-		if strings.Contains(string(watch.Out.Contents()), "memsw.limit_in_bytes: no such file or directory") {
+		if match {
 			Skip("swap limits not enabled; skipping")
 		}
 
--- a/testflight/container_limits_test.go
+++ b/testflight/container_limits_test.go
@@ -16,7 +16,8 @@ var _ = Describe("A job with a task that has container limits", func() {
 		watch := spawnFly("trigger-job", "-j", inPipeline("container-limits-job"), "-w")
 		<-watch.Exited
 
-		if strings.Contains(string(watch.Out.Contents()), "memsw.limit_in_bytes: permission denied") {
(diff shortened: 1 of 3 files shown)


worker/runtime/spec/spec.go
```diff
@@ -150,7 +151,9 @@ func OciResources(limits garden.Limits) *specs.LinuxResources {
 	if memoryLimit > 0 {
 		memoryResources = &specs.LinuxMemory{
 			Limit: &memoryLimit,
-			Swap:  &memoryLimit,
+		}
+		if IsSwapLimitEnabled {
+			memoryResources.Swap = &memoryLimit
```
chenbh, line 156: I'm not a fan of making global variables public purely for testing, can we have `OciSpec` pass this in as a parameter instead?

worker/runtime/spec/spec.go
```diff
@@ -171,6 +174,18 @@ func OciResources(limits garden.Limits) *specs.LinuxResources {
...
 }
 
+var IsSwapLimitEnabled bool
+
+func init() {
+	IsSwapLimitEnabled = swapLimitEnabled()
+}
+
+func swapLimitEnabled() bool {
+	swapLimitFile := "/sys/fs/cgroup/memory/memory.memsw.limit_in_bytes"
```
chenbh, line 184: Can you move this to the top level var declaration? I like the pattern of defining any hard coded values at the top of the file.

chenbh (changes requested):

testflight/container_limits_test.go
```diff
@@ -16,7 +16,8 @@ var _ = Describe("A job with a task that has container limits", func() {
 		watch := spawnFly("trigger-job", "-j", inPipeline("container-limits-job"), "-w")
 		<-watch.Exited
 
-		if strings.Contains(string(watch.Out.Contents()), "memsw.limit_in_bytes: permission denied") {
+		// Guardian runtime will always fail this test unless it's explicitly told not to set the swap limit
+		if strings.Contains(string(watch.Out.Contents()), "memsw.limit_in_bytes: no such file or directory") {
```
chenbh, line 20: The Guardian runtime is failing because they're using an older version of runC with the `permission denied` error, maybe we should be doing regex matching instead? `failed to write "[0-9]+" to "/sys/fs/cgroup/memory/.*/memory.memsw.limit_in_bytes"`

The full error looks like
```
find or create container on worker fe03fcc43bdb: runc run: exit status 1: container_linux.go:367: starting container process caused: process_linux.go:459: container init caused: process_linux.go:422: setting cgroup config for procHooks process caused: failed to write "1073741824" to "/sys/fs/cgroup/memory/garden/b6ffb7e4-da29-462c-6514-3550aca2d4b6/memory.memsw.limit_in_bytes": open /sys/fs/cgroup/memory/garden/b6ffb7e4-da29-462c-6514-3550aca2d4b6/memory.memsw.limit_in_bytes: permission denied
```

taylorsilva: I'd actually prefer to let this fail since it will pass when bumping runc to rc-93. Will be a good indicator that we're using the latest version of runc. wdyt?

chenbh: I would argue that the test should not fail and be skipped regardless of the rc version. It's testing to see if the container limits are respected, if it's not possible to modify the cgroup, then this test shouldn't even run.

It's our fragile parsing of the error message that's causing it to fail instead of skip. The criteria for skipping should be "writing to the memory.limit_in_bytes cgroup failed (for whatever reason)", not "this file doesn't exist" or "you don't have permissions"


> Will be a good indicator that we're using the latest version of runc.

This might be very purist, but why should the test (or by extension us) care? We do not rely on any special features of rc93, so as long as it doesn't break our code, it should be treated like any other dependency and get bumped when it gets bumped.

chenbh (approved):
lgtm

PR #6659, opened 2021-03-11
Revert "feat(atc): cache streamed volumes and try to get by looking for local cache"

clarafu:
Reverts concourse/concourse#6495

After deploying this change to our ci deployment there were many build failures such as the following screenshots below

Screen Shot 2021-03-11 at 3 17 57 PM
Screen Shot 2021-03-11 at 3 17 33 PM
Screen Shot 2021-03-11 at 4 31 03 PM

@evanchaoli unfortunately we are not going to include the pr in the v7.1.0 release, we will need to do some much more thorough testing and reviewing of the pr with probably some other concourse team members that have more context than me about the workings of containers/volumes.


chenbh (approved):

evanchaoli:
> @evanchaoli unfortunately we are not going to include the pr in the v7.1.0 release, we will need to do some much more thorough testing and reviewing of the pr with probably some other concourse team members that have more context than me about the workings of containers/volumes.

@clarafu  It's absolution ok because nobody wants an unstable 7.1.0. I will resubmit the PR and call for more reviews.

Are error logs of ATC available for those error screen shots? I have no any non-Linux worker, if the problem only exists on non-Linux worker, then I'll hard to reproduce and fix.

PR #6660, opened 2021-03-12
RE:feat(atc): cache streamed volumes and try to get by looking for local cache

evanchaoli:

As reverted by #6659, this is a resubmit of #6495 for more reviews and tests.

From https://github.com/concourse/concourse/pull/6659#issue-591148375, there are 3 errors seen.

* The first one, "worker based resource type disappeared", as @clarafu analyzed, that could be because non-Linux workers don't have some base resource types. I think this error can be silently discard.
* The second one, "failed to stream in to volume". This is a volume stream error from baggage claim. Would need ATC log to further see why volume streaming failed.
* The third one, "$GOPATH/go.mod exists but should not", this could be related to task cache. My previous tests didn't cover task cache, I'll try to reproduce this error.

## Release Note

Optimize resource cache streaming and get step.

* Mark streamed resource cache volumes as resource cache, so that avoid duplicate streaming in next runs.
* If a resource to `get` can be found on some workers, then `get` step will do nothing. This will reduce times of Concourse connecting to external systems, such as git, docker hub, and so on.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


evanchaoli:
For the first error, I have discarded it.

For the third error, I managed to reproduce it and found the root cause. In the original code, task caches are handled in function `gardenWorker.cloneRemoteVolumes()`. #6659 changed the way how remote inputs are handled, but didn't take care of task caches. I have fixed the bug by handling task cache in a new function `gardenWorker.createCacheVolumes()`.

The second error is likely caused by task cache. But I'm not sure without ATC logs.

clarafu:
I think for the first error, we shouldn't ignore it but rather we should be putting in the `worker_base_resource_type_id` of the original worker resource cache into the `worker_resource_cache` for the streamed volume when we initialize it as a resource cache. This is because the `worker_base_resource_type_id` exists for us to know when to invalidate a worker resource cache when the worker base resource type gets a new version, and since the resource cache is actually fetched from a different worker when that different worker gets a new version of the base resource type it should invalidate all the resource caches that were created using that old version of the base type even if they now exist on another worker (because the caches were streamed and initialized on another worker). Ill try making an example to make sure I'm explaining my thoughts correctly:

```
Worker A ---> ResourceCache A [BaseType X = v1]
                     |
                     |  (streamed)
                     ▼
Worker B ---> ResourceCache B [BaseType X = v1]
```

So resource cache A is streamed onto worker B which creates resource cache B (it is basically the exact same as resource cache A). If worker A gets a new version of base resource type X, I think I would expect that resource cache A and resource cache B would be invalidated because they were both created using base resource type X.

evanchaoli:
> So resource cache A is streamed onto worker B which creates resource cache B (it is basically the exact same as resource cache A). If worker A gets a new version of base resource type X, I think I would expect that resource cache A and resource cache B would be invalidated because they were both technically created using base resource type X on worker A.

I doubt this process. When a worker joins ATC, https://github.com/concourse/concourse/blob/master/atc/db/worker_factory.go#L535-L543 will delete out-of-date `worker_base_resource_types` entries, then on rely db constraint `on delete cascade`, corresponding `worker_resource_caches` entries will be deleted as well, then Volume will be GC-ed.

But as worker B doesn't contains any base resource type, the invalidate process may not work as expected. Definitely, there are a lot of details behind that I might not fully understand yet. But the point is, there might be a lot of uncertain things hidden if we decide to introduce this new logic.

From the other perspective, creating resource cache on workers that don't contain base resource types will introduce some new complexities. Only non-Linux workers don't contain base resource types, and non-Linux worker is the minority. My implementation (discarding the error of "worker based resource type disappeared") just disable the feature on non-Linux workers, I think it's acceptable, because it just keeps current behavior on non-Linux workers. Why cannot we consider this feature is only bonus for Linux workers, which is the majority? 

If you really want to make non-Linux workers to mark resource cache, can we consider that as a separate enhancement with a new PR? I believe there will be a lot of things to investigate for that.

clarafu:
Right, currently the invalidating process does not work as expected because we currently tie together a streamed resource cache with the incorrect worker base resource types. I think we might have some disconnect in our understanding of the problem so let me try to explain my thoughts a little more thoroughly. I think the problem is that this is not only a bug with Linux and non-Linux workers, but with the streaming logic in general because of worker base resource types.

So lets say we have two workers, worker A and worker B. They both contain the git resource which makes it so there are two entries in the worker base resource types table. The entry for worker A's git base resource type has an id of 1 but with version `1.0.0` and the entry of worker B's git base resource type has an id of 2 with version `2.0.0`.
```
worker base resource types

| id       | base resource type    | version      | worker name         
---------------------------------------------------------------
| 1        | git                   | 1.0.0        | worker A
| 2        | git                   | 2.0.0        | worker B
```
Right now if we create a resource cache using `git` on worker A, it will point to `worker_base_resource_type_id` = 1. So there will be an entry in the `worker_resource_caches` that contains the `resource_cache_id`, worker and the `worker_base_resource_type_id`. This makes sense because we actually used the git base resource type on worker A to fetch the resource cache.
```
worker resource caches

| id     | resource_cache_id           | worker_base_resource_type_id
-----------------------------------------------------------------
| 1      | 1                           | 1
```
Now lets say we pick worker B to run a task that uses this git resource. The volume will be streamed from worker A to worker B and then we will initialize the resource cache on worker B. This will generate another row in the `worker_resource_caches` table but this time the `worker_base_resource_type_id` will point to the git base type on worker B. (We run FindOrCreate using the name of the worker we streamed to and the resource cache https://github.com/concourse/concourse/blob/28153d0cb8dd6f670d09a2001bd7b27d74ddf217/atc/db/volume.go#L363-L366)
```
worker resource caches

| id     | resource_cache_id           | worker_base_resource_type_id
-----------------------------------------------------------------
| 1      | 1                           | 1
| 2      | 1                           | 2
```
This is where it gets weird because the resource cache on worker B was actually **fetched** by worker A, just that the volume was streamed to worker B. So if worker A used version `1.0.0` of the git base resource type to fetch the resource cache, after streaming the volume and initializing the cache we will now think that we used version `2.0.0` of the git base resource type because that is the version that exists on worker B even though we never used version `2.0.0` to fetch the resource cache.

This is just one of the reasons why I think it is important to tie the worker resource cache to the **original** worker base resource type id that was used to create/fetch the resource cache. The other big reason is the one I mentioned in my previous comment, with invalidating the worker resource caches when it gets out of date. I don't think it makes sense to invalidate resource caches if those caches were not actually fetched using an out of date version of a base resource type. So I don't think this is just a problem that will be solved with ignoring the error when the worker doesn't contain the base resource type, I think this is just a bug that affects initializing resource caches due to a streamed volume between all types of workers regardless if they have the base resource type or not.

evanchaoli:
@clarafu I didn't notice that `InitializeResourceCache()` doesn't put version of worker base resource into consideration. I agree that, if a volume is fetched on worker A by git 1.0, then after the volume is streamed to worker B that contains git 2.0, we should not mark the volume as a resource cache on worker B.

aoldershaw:
Not an expert on the DB side of things, so this might not make complete sense, but it feels like a nicer solution would be for all workers with the cache to reference the *same* `worker_base_resource_type_id` (even though it only exists on the worker that actually ran the `in` script). That way, once the cache is invalidated, it is no longer used on *any* of the workers that reference it (i.e. there's a single source of truth for the cache). It also means workers without base resource types (e.g. darwin/windows workers) can keep resource caches around.

To do this, I think we'd need to extend `worker_resource_caches` - right now, a worker resource cache is identified by the pair `(worker_base_resource_type_id, resource_cache_id)`, but for this to work, I think we'd alternatively allow worker resource caches to be identified by `parent_worker_resource_cache_id` (which is a FK back to `worker_resource_caches`). When we `InitializeResourceCache`, we either initialize it using the `(worker_base_resource_type_id, resource_cache_id)` pair OR by referencing a parent `WorkerResourceCache`.

Does this seem sensible @clarafu @evanchaoli ?

clarafu:
@aoldershaw I completely agree with you, that is definitely the same direction of what I was proposing. I think the solution @evanchaoli that you came up with is slightly off from my original idea. The point I was trying to get across was not the version misalignment, but just the whole concept that we need to have the worker resource cache point to the original worker base resource type that was used to create it before it was streamed.

evanchaoli:
@clarafu @aoldershaw OK, finally I catch up with your idea. 😃 

However I'm afraid if you over thought the problem. All workers suppose to have same versions of base resource types, even if there is some slight version difference, versions should be compatible, otherwise Concourse would behave unpredictable.

Let's back to the basic, considering a case without this PR. If worker A has `git` resource type version 1.0, and worker B has 2.0. Assume the two versions generate incompatible artifacts. A `get: mygit` step may be dispatched to either worker A or worker B, meaning `get: mygit`'s behavior is unpredictable, which is definitely not expected.

As base resource type is wrapped in Concourse release binary, I don't think Concourse provide a way and encourage users to upgrade base resource type without upgrading worker.

In our case, the only possibility of two workers having different version of resource type is during small upgrading windows. But if two worker versions are incompatible from any aspect, we will claim downtime, destroying all old workers, then redeploy two workers.

So I don't think base resource type version should play an important role here. Otherwise, current Concourse should also handle it, considering the case worker A and worker B having different versions of `git` resource type, then which worker should be picked to run `git: mygit` step? Dose that make sense?

If you agree with what just talked, then I guess we don't even need the version comparison from my yesterday's commit.

chenbh:
> From the other perspective, creating resource cache on workers that don't contain base resource types will introduce new complexities. However, only non-Linux workers don't contain base resource types, and non-Linux worker is the minority. 

It's a perfectly valid use case to repackage the Concourse distribution with your own custom base resource types. Maybe I have 2 different worker deployments, one generic and one specialized with an additional custom base resource type (idk, maybe there's special hardware attached and is modelled as a resource). Both worker deployments are linux-based, but I would still like to benefit from streaming the output of my special worker over to a generic worker for processing.

evanchaoli:
> It's a perfectly valid use case to repackage the Concourse distribution with your own custom base resource types. Maybe I have 2 different worker deployments, one generic and one specialized with an additional custom base resource type (idk, maybe there's special hardware attached and is modelled as a resource). Both worker deployments are linux-based, but I would still like to benefit from streaming the output of my special worker over to a generic worker for processing.

@chenbh  If two workers contain different incompatible base type versions, shouldn't they be tagged differently? Otherwise `get: the-resource` may be dispatched to them randomly, generating incompatible artifacts, which is not a right behavior.

Also I would argue, in that case, should the custome base type be named differently? When ATC chooses a worker, it doesn't consider base resource type at all, which implies a base resource type should be all on same version across workers, or at lease on compatible versions.

https://github.com/concourse/concourse/blob/bb69967f7a59c175dc36fc95c617c5f1ddd71f50/atc/worker/worker.go#L715-L727

> Or maybe I have a linux worker that is completely cut off from the public internet and can only communicate with other workers. I can further enforce this by removing all the base resource types from the worker, which would guarantee no check/get/put steps will ever be scheduled there, but I can still use it for tasks.

In that case, this PR just doesn't benefit those internal workers. 

I try to keep this PR simple, make it benefits majority normal Linux workers. At least our clusters don't have such special use cases. I want to first step prove if this feature really improve performance in real environments. So for other use cases, if we really want to support, we can enhance later with a new PR.

evanchaoli:
@clarafu @aoldershaw @chenbh Thanks for your comments. But it seems we are stuck here. So let's first align with a root question: when a streamed volume should be marked as a resource cache on dest worker?

My answer should be,  if the same `get` step running on dest worker would generate exact same result (artifacts) as on source worker, then when the resource cache volume is streamed from source worker to dest worker, it can be marked as a resource cache.

Does that make sense?

chenbh:
> if the same get step running on dest worker would generate exact same result (artifacts) as on source worker, then when the resource cache volume is streamed from source worker to dest worker, it can be marked as a resource cache.

👍  I agree and can absolutely see the performance benefits of this.

> @chenbh If two workers contain different incompatible base type versions, shouldn't they be tagged differently? Otherwise get: the-resource may be dispatched to them randomly, generating incompatible artifacts, which is not a right behavior.

I think there's a misunderstanding, my argument was what happens when a worker has more base resource types than another worker,  *not* a different version of base resource type. I.e worker-1 has base resource types `git` and `registry-image`, but worker-2 has `git`, `registry-image`, and `my-special-resource` (v1). 

If I `get` `my-special-resource` and then stream that to worker-1 for use in a task, I would hope that a resource cache gets created with the `base_resource_type_version_id` for the `my-special-resource` (v1) on worker 2. So that if I upgrade worker-2 with a new version of `my-special-resource` (v2), it should invalidate the cache on worker-1.

This approach would basically solve the Windows and Darwin worker problem since there's no more distinction of linux vs non-linux workers in terms of resource caches.

> In that case, this PR just doesn't benefit those internal workers.
>
> I try to keep this PR simple, make it benefits majority normal Linux workers.

Good point, maybe this can be done in a future PR. My personal opinion was that I don't consider this feature "complete" without considerations to workers with different sets of base resource types. But I'm fine with spinning that out to a separate issue if you feel like that's scope creep.

clarafu:
I'm still on the same page as @chenbh with the fact that I believe when a volume gets streamed to a different worker and initialized on that different worker, the base resource type it should be initialized with should be the worker base resource type id of the **source** worker that the volume streamed **from**.

> My answer should be, if the same get step running on dest worker would generate exact same result (artifacts) as on source worker, then when the resource cache volume is streamed from source worker to dest worker, it can be marked as a resource cache.

I guess I don't really see the reason for only being able mark a volume on a dest worker if it would generate the exact same results as on the source worker. I mean that I'm not against it, I just don't see the benefits of imposing this rule versus the idea that I have been proposing (@chenbh and @aoldershaw too). The main difference I see is that the thing you are proposing will result in a very common use case of non-Linux workers not being able to have any resource caches initialized on it but I don't know if there is a strong enough argument for that. I think I feel so strongly about having the **source** worker's `worker_base_resource_type_id` be set on any resource caches that are initialized through streaming because that is the exact base resource type that was used to create that resource cache. If we tie the `worker_base_resource_type_id` of the destination worker to the resource cache, it seems technically wrong because we didn't actually use that worker base resource type to create this resource cache.

evanchaoli:
> I think I feel quite strongly about having the **source** worker's `worker_base_resource_type_id` be set on any resource caches that are initialized through streaming because that is the exact base resource type that was used to create that resource cache.

@clarafu We cannot simply point dest `worker_resource_cache` tuple  to source `worker_base_base_resource_type_id`.  Because `worker_base_base_resource_type_id` is the way to find out worker name. If you read:

https://github.com/concourse/concourse/blob/9b1818ae67e2dfa7bc1082eb0e5d9152ca12c221/atc/db/volume_repository.go#L428-L443

If we simply point to source `worker_base_base_resource_type_id`, then `volumeClient.FindVolumeForResourceCache()` will not be able to find the resource cache for dest worker. The method @aoldershaw suggested, adding a column won't solve the problem either. To solve the problem, more works need to do. How much more? I don't know. I guess it needs a lot of more investigation.

To proceed, I think there are two options:

Option 1: Just use my current implementation: discard `BaseResourceTypeMissingError` from `InitializeResourceCacher`. This option just prevents non-Linux workers who don't build in base resource types from enjoying streamed resource cache. This solution is straightforward, and simple. If we take this option, this PR is ready.

Option 2: To make a perfect solution, one solution I can come out is:

* Add a new table `streamed-worker-resource-caches` with fields `(id, source_worker_based_resource_type_id, worker_name, resource_cache_id)`. 
* Add a column `streamed-resource-cache-id` to table `volumes`.
* After a resource cache volume is streamed, insert a tuple to `streamed-worker-resource-caches` and set the tuple id to `volumes` tuple's `streamed-resource-cache-id`.
* When a worker's base resource type is invalidated, or a resource cache is invalidated, corresponding tuples in `streamed-worker-resource-caches` are invalidated too, then GC will destroy corresponding volumes.

This options would need a lot of more works, especially a lot of tests to run, which I believe deserves a separate PR.

So, I would prefer option 1 for this PR, and leave option 2 to a new PR. Are you ok with that?

@vito Would you mind comment a little bit here?

aoldershaw:
hey @evanchaoli, sorry for the late response! I was looking through the code to try to figure out how much work it'd be to handle Option 2 with the original solution I proposed. You're right that it wouldn't work by just adding the `parent_worker_resource_cache_id ` column by itself, but I think it would work if we also added the `worker_name` - instead of looking up the `base_resource_type` first, we just lookup the `(worker_name, resource_cache_id)` pair.

@clarafu would finding `worker_resource_caches` by `(worker_name, resource_cache_id)` work? I get that we keep track of the `worker_base_resource_type_id` in case the `worker_base_resource_type`'s version or image changes, but in that case, we `DELETE` the `worker_base_resource_type` which will `DELETE` the corresponding `worker_resource_caches` (because it's `ON DELETE CASCADE`).

https://github.com/concourse/concourse/blob/55563b0733a775b8c40bc259a522645496acd024/atc/db/worker_resource_cache.go#L104-L109

Since we `DELETE` any `worker_base_resource_types` besides the current one for a worker, this should be equivalent - right?

evanchaoli:
@aoldershaw If we added both `parent_worker_resource_cache_id` and `worker_name` to table, for a streamed resource cache, we will fill in these two columns, but leave `worker_base_resource_type_id` blank, which is almost same as my proposal of option2 of adding a new table, and I feel adding a new table is better.

aoldershaw:
@evanchaoli I think adding a new table would work too. I think I overcomplicated my proposed solution, though - I don't think we need `parent_worker_resource_cache_id` at all. If `InitializeResourceCache` took in `srcWorker string`, I think we could implement this pretty easily.

1. If `srcWorker == volume.Worker()`, then we `FindOrCreate` the `WorkerBaseResourceType` as before
2. If `srcWorker != volume.Worker()`, then we only `Find` the `WorkerBaseResourceType` (using the `srcWorker`) - if it doesn't exist, we error (because the base resource type disappeared)
3. In either case, we save the `worker_resource_cache` with `worker_name = volume.Worker()` and with the appropriate `worker_base_resource_type_id` from 1. or 2. (i.e. the `worker_base_resource_type_id` points to the actual base resource type used to fetch the volume, but it may not have originated from `volume.Worker()`
4. I think we'd want to change the `ON CONFLICT` here to be based on conflicts to `(resource_cache_id, worker_name)` instead https://github.com/concourse/concourse/blob/970c85cb2f2d27dbd189bb60958b62ad57aedf87/atc/db/worker_resource_cache.go#L56-L58
5. When we `Find` a `WorkerResourceCache`, we look it up by the `worker_name` and `resource_cache_id`, rather than by looking up a `WorkerBaseResourceType` first

Does that make sense? I feel like this solution would reduce the scope compared to adding a new table/concept - it just extends the `worker_resource_cache` concept slightly such that the `worker_base_resource_type` can point to a different worker.

WDYT? Tagging @clarafu as well - let me know if this won't work

clarafu:
@aoldershaw I totally agree with your idea, I think it also makes sense database schema wise to add the `worker_name` that this `worker_resource_cache` is for into the `worker_resource_caches` table if we are going to have the worker resource cache be able to be pointed to a different worker.

I think one thing I might change is that for step 2, I would run a query to grab the value of the `worker_base_resource_type_id` from the source resource cache rather than doing a findOrCreate of the WorkerBaseResourceType using the srcWorker. This is because the source worker could be a worker that had its resource cache streamed to it from a different worker. For example, 

```
resource cache created on Worker A---> streamed to Worker B ---> streamed to Worker C
```

This would be the schema I would expect to result from this

```
worker_resource_cache (original)
  -> worker_base_resource_type_id = 1
  -> worker_name = worker A

worker_resource_cache (streamed)
  -> worker_base_resource_type_id = 1
  -> worker_name = worker B

worker_resource_cache (streamed)
  -> worker_base_resource_type_id = 1
  -> worker_name = worker C
```

aoldershaw:
> I think one thing I might change is that for step 2, I would run a query to grab the value of the worker_base_resource_type_id from the source resource cache rather than doing a findOrCreate of the WorkerBaseResourceType using the srcWorker. This is because the source worker could be a worker that had its resource cache streamed to it from a different worker

Yup, totally makes sense, good catch!

evanchaoli:
@aoldershaw I still feel the thing is not as simple as adding a column `worker_name`. Let's consider this use case:

A simple job:

```yaml
jobs:
- name: some-job
  plan:
  - get: mygit
  - task: some-task
    config:
       inputs:
       - name: mygit
       run: [...]
```

Say, two builds (`build-1` and `build-2`) of this job are invoked at same time.

- `get: mygit` of `build-1` is on worker-1, and `get: mygit` of `build-2` is on worker-2
- Then after both `get`s, `worker-resource-caches` table contains:

  |id|worker-base-resource-type-id|resource-cache-id|worker-name|
  |--|------------------------------|-------------------|-------------|
  |1|20|100|worker-1|
  |2|30|100|worker-2|

- Tasks of both builds are scheduled to worker-3
  - As streaming `mygit` volume to worker-3 takes a while, both builds may streaming their `mygit` volumes to worker-3
  - Say, `build-1` finishes streaming `mygit` first, it inserts a new tuple:
  
    |id|worker-base-resource-type-id|resource-cache-id|worker-name|
    |--|------------------------------|-------------------|-------------|
    |1|20|100|worker-1|
    |2|30|100|worker-2|
    |3|20|100|worker-3|

  - After that, `build-2` finishes streaming `mygit`, per your theory, it should update tuple 3 due to ` on conflict`:
 
    |id|worker-base-resource-type-id|resource-cache-id|worker-name|
    |--|------------------------------|-------------------|-------------|
    |1|20|100|worker-1|
    |2|30|100|worker-2|
    |3|30|100|worker-3|

which doesn't sound correct. Because the first streamed `mygit` volume still points to tuple-3, but now tuple-3's `worker-base-resource-type-id` is updated to 30, while the first streamed `mygit` volume is generated by `worker-base-resource-type-id` 20 from `worker-1`. What do you think?

Also, with your proposal of `on conflict`, you also ignored base-resource-type's version difference between workers. But version difference is one of the concern @clarafu is having, right?

If base-resource-type version difference among workers is not a concern, I feel my current code change is good to merge, the only problem is that it doesn't handle non-Linux workers who don't embed base. But given non-Linux workers are minority, we may defer the support to a new PR.

aoldershaw:
@evanchaoli thanks for explaining that so clearly. That does make sense, and I do feel it's a bit concerning having multiple volumes on a worker that point to the same `worker_resource_cache_id` but originate from a different `worker_base_resource_type` (while only one of those `worker_base_resource_type_id`s is tracked).

However, I think that concern still holds true in the current state of this PR. Using the same example scenario, suppose we had:

|id|worker-base-resource-type-id|resource-cache-id|
  |--|------------------------------|-------------------|
  |1|20|100|
  |2|30|100|

Then we stream `mygit` from `build-1` to `worker3`:

  |id|worker-base-resource-type-id|resource-cache-id|
  |--|------------------------------|-------------------|
  |1|20|100|
  |2|30|100|
  |3|40|100|

Then we stream `mygit` from `build-2` to `worker3`:

|id|worker-base-resource-type-id|resource-cache-id|
  |--|------------------------------|-------------------|
  |1|20|100|
  |2|30|100|
  |3|40|100|

(no change here, since we're using the `base_resource_type_id` on `worker3` in each case).

In this case, we also have two volumes on `worker3` that originated from two different base resource types - the only difference I can see is that we aren't tracking either of the "actual" `worker_base_resource_type_id`s, so won't invalidate the cache it that base resource type changes

evanchaoli:
@aoldershaw I think the difference between your proposal and my current implementation is that, your proposal is trying to solve a concern (base resource type version difference among workers) but it actually doesn't resolve the concern. While, my implementation assumes there is no such a concern. My theory is, a base resource type should be on the same version (or at least at compatible versions) among all workers, otherwise Concourse will generate unpredictable results.

Still following the case you described. Now there are 2 streamed volumes on worker-3, both point to tuple-3 of `worker-resource-caches`. When resource cache 100 is invalidated, then the both volumes will be invalidated.

But yes, this scenario generates a duplicate resource cache volume, which waste storage of worker-3. To solve the problem, we can enhance volume-gc to destroy duplicate resource cache volumes on a worker. Yeah, from this perspective, the current PR is not "good enough". If you agree with the direction, I can start to worker on volume-gc enhancement.

clarafu:
That's a good point to bring up, I think my original thought was that we would change the unique constraint on the `worker_resource_caches` table to include the `resource_cache_id`, `worker_base_resource_type_id` and the `worker_name`. So in your example, I would actually see four rows in the table after `build-2` finishes streaming

    |id|worker-base-resource-type-id|resource-cache-id|worker-name|
    |--|------------------------------|-------------------|-------------|
    |1|20|100|worker-1|
    |2|30|100|worker-2|
    |3|20|100|worker-3|
    |4|30|100|worker-3|

This will result in two resource caches being created on that worker, which I think is fine. We would then use the `worker_name` and `resource_cache_id` to look up a volume on a worker within `FindResourceCacheVolume` and this would just return back any one of those rows that match the `worker_name` and `resource_cache_id`.

The `worker_base_resource_type_id` will purely be used for gc (gcing any resource caches that are using an old `worker_base_resource_type_id`) and no longer for lookups to **find** a worker resource cache. If we want to find a worker resource cache, we will use the `worker_name` and `resource_cache_id`. Doing it this way will result in duplicate `worker_resource_caches` but those will be automatically gced once we found it is not being used by any volumes. Let me know what you think @aoldershaw and @evanchaoli

evanchaoli:
> We would then use the `worker_name` and `resource_cache_id` to look up a volume on a worker within `FindResourceCacheVolume` and this would just return back any one of those rows that match the `worker_name` and `resource_cache_id`.

@clarafu From this statement, I believe you have agreed that base resource type version difference between workers doesn't matter. Then only intention of adding `worker-name` is to support non-Linux workers. Again, I still think we should defer non-Linux support to a separate PR. I would rephrase my reasons:

* non-Linux workers are the minority. Introducing such complexities of adding `worker-name` and changing existing logic of `worker-resource-caches` to support the minority sounds low ROI and risky. (where for "risky", I mean I may still yet fully understand the logic of `worker-resource-caches`, changing its logic may lead to potential issues.)
* I have no any non-Linux worker in hand, I don't want to spend time setting up one either. So it's hard for me to test and maintain non-Linux related features.
* I want this PR to catch up with 7.2, or 7.3 if it's too late for 7.2.
* We may add some comment in code to explain how non-Linux support can be done, and we can open a help-wanted issue. If some contributor wants this feature, s/he may help implement. If nobody cares that, which proves we make right decision of not investing on non-Linux support at this moment.

clarafu:
I'm not proposing that the difference in version doesn't matter, but I think more so that in this specific scenario we have no way of figuring out which one of those caches is using a "newer" base resource type version. A large reason why I'm so adamant towards putting this change forward within this PR is that without the change I am asking for, it will be a regression to the case of when we get a new version of worker base resource types, the resource caches from streamed volumes that use the old worker base resource type won't get gced and concourse will continue to use those resource caches using the old worker base resource type.

evanchaoli:
@clarafu I don't think there is any regression. Let's still use the above example.

With my proposal:

id | worker-base-resource-type-id | resource-cache-id
-- | -- | --
1 | 20 | 100
2 | 30 | 100
3 | 40 | 100

when the base resource type on worker-1 is upgraded, same upgrade should be done on worker-3 too, thus when 20 is invalidated, `40` will be invalidated as well. Otherwise Concourse will generate different result when a `get` of that resource is scheduled to worker-1 and worker-3.

With your proposal:

id | worker-base-resource-type-id | resource-cache-id | worker-name
-- | -- | -- | --
1 | 20 | 100 | worker-1
2 | 30 | 100 | worker-2
3 | 20 | 100 | worker-3
4 | 30 | 100 | worker-3

If the base resource type is only upgraded on worker-1, that will only invalidate tuple-1 and tuple-3, tuple-2 and tuple-4 will be still there. `get` step will find the resource cache on worker-2 and worker-3, so that newly upgraded base resource type on worker-1 will never be used.

So that, when all workers keep the same base resource type version, I don't think there is any regression.

aoldershaw:
> when the base resource type on worker-1 is upgraded, same upgrade should be done on worker-3 too

While that's often the case, it's not necessarily true. Different pools of workers can be managed independently by different groups of people, so upgrades aren't always simultaneous.

Also, it's not always the case that if workers have different base resource type versions, then the result must be non-deterministic. For instance, suppose that `worker-1` and `worker-2` are both a part of a separate pool of tagged workers, perhaps with access to some private infrastructure that's not available on other workers, and perhaps deployed separately from the primary pool of workers. `get: mygit` has a tag (referencing these workers) since it needs access to this private infrastructure.

If `worker-1` and `worker-2` are redeployed with a new base resource type version, we should invalidate the (streamed) worker resource cache(s) on `worker-3`, even if we never redeployed `worker-3`. Pointing to the original `worker_base_resource_cache_id` (on `worker-1`/`worker-2`) is probably the easiest way to accomplish that

evanchaoli:
@aoldershaw Thanks for the further explanation.

> For instance, suppose that worker-1 and worker-2 are both a part of a separate pool of tagged workers, 

That could be possible, but should be rare. At least, none of our clusters are managed in that way. Comparing the burden of adding `worker-name` with the gain, I don't think it's worthy.

It feels that we are hard to convince each other. How about to add a feature flag? Make this PR an experimental feature with a clear statement like "this feature can only be turned on when all workers keep the same version of base resource types".

clarafu:
I want to be straight with you that I don't agree with adding a feature flag for it just so that we can get this PR merged now. I think that the most important thing is to try our best improve the quality of the codebase. Concourse already has many stability issues, bugs and a **lot** of technical debt so anything going forward we want to either be helping improve or get rid of these debts. There is no guarantee that after feature flagging it that the work needed to be done to "finish" it will be completed, and adding additional half finished code will just be adding more technical debt into the codebase. 

I really want you to understand why I am being so adamant on this, since we are going to be working together a lot in the future. Even if a bug only occurs in rare situations, 99% of the time I would side on spending the time to fix it upfront. This is because if we leave the bug to exist and a user reports it, we will need to spend much more time trying to diagnose the reason and then having to implement the fix later. Typically the time it takes to fix a bug, most of the time is spent on the diagnoses rather than on the implementation. Secondly, if we let small bugs slide eventually they will build up and ruin the experience for users. What a user wants most is a stable and trustworthy product, trustworthy in a way that they know what they are going to get when they use a product. When the user is constantly faced with bugs even if they are small, that breaks the trust with the product. Right now, Concourse is already filled with many bugs and as maintainers we always try to do our best to avoid introducing more. I really want us to be aligned on our goal for improving Concourse as a product by improving both the stability and feature set at the same time. I apologize if I am stating points that you already know, I just want to be clear with you about my thoughts when I am working on Concourse.

I also think that finishing the implementation with the `worker_name` and supporting non-Linux workers won't be that much additional work. In hindsight, we probably could've finished this PR with the time we have spent going back on forth. I think the only time I would like a feature flag is after completing this caching streaming volumes feature, and it would be to protect us against unknown bugs.

evanchaoli:
@clarafu Quality is no doubt always the most important thing. Our difference is that, I consider adding `worker-name` might be more buggy, maybe I don't understand the whole picture of worker-resource-caches as deep as you do.

I can quickly implement `worker-name`. Please take a look at the latest commit. Some more works need to do:

* New tests need to add
* The migration scripts need to enhance. In up script, need to populate `worker-name` for existing tuples; In down script, need to delete streamed cache tuples.
* GC need to enhance to delete duplicate resource caches on a worker.

If the last commit looks good to you, I'll work on the rest of things.

atc/db/migration/migrations/1618553965_add_worker_name_to_worker_resource_caches.up.sql
```diff
@@ -0,0 +1,12 @@
+ALTER TABLE worker_resource_caches
+    ADD COLUMN worker_name text REFERENCES workers(name) ON DELETE CASCADE;
+
+DROP INDEX worker_resource_caches_uniq;
+
+CREATE UNIQUE INDEX worker_resource_caches_uniq
```
clarafu, line 6: I think it might be more beneficial to create this unique index in the order of `worker_name, resource_cache_id, worker_base_resource_type_Id`. This is because we will be querying for worker resource caches using `worker_name` and `resource_cache_id`, and not `worker_base_resource_type_id` so having  `worker_base_resource_type_id` as the first column in the index will result in those queries not being able to use this index. Another reason is that the index being split on `worker_name` first will help narrow down the scope of the queries using this index because it will be scoping down the rows in the table to the specific worker we are looking for, and then try to look for the resource cache within those rows.

atc/db/worker_resource_cache.go
```diff
@@ -70,54 +81,61 @@ func (workerResourceCache WorkerResourceCache) FindOrCreate(tx Tx) (*UsedWorkerR
...
+		err := rows.Scan(&id, &workerBasedResourceTypeId)
+		if err != nil {
+			return nil, nil, false, err
 		}
+		ids = append(ids, id)
+		workerBasedResourceTypeIds = append(workerBasedResourceTypeIds, workerBasedResourceTypeId)
+	}
 
-		return 0, false, err
+	if len(workerBasedResourceTypeIds) == 0 {
```
clarafu, line 136: Is there a reason that we need this check? 🤔  If the length of them is 0, wouldn't that mean that the value of `ids` and `workerBasedResourceTypeIds` will be nil as well (because there would be no rows fetched from the query)?

atc/worker/artifact_source.go
```diff
@@ -147,12 +163,39 @@ func (source *artifactSource) StreamTo(
...
+
+	// If the source volume is a resource cache, then mark dest volume as a resource cache too.
+	if source.volume.GetResourceCacheID() > 0 {
+		usedResourceCache, found, err := source.resourceCacheFactory.FindResourceCacheByID(source.volume.GetResourceCacheID())
+		if err != nil {
+			logger.Error("stream-to-failed-to-find-resource-cache", err)
+			return err
+		}
+		if !found {
+			logger.Info("stream-to-not-find-resource-cache--should-not-happen",
```
clarafu, line 181: Just a small typo!

```suggestion
			logger.Info("stream-to-not-find-resource-cache-should-not-happen",
```

atc/db/worker_resource_cache.go
```diff
@@ -70,54 +81,61 @@ func (workerResourceCache WorkerResourceCache) FindOrCreate(tx Tx) (*UsedWorkerR
...
 		return nil, false, err
 	}
 
 	if found {
-		return &UsedWorkerResourceCache{
-			ID: id,
-		}, true, nil
+		// Verify worker base resource type
+		for i, workerBasedResourceTypeId := range workerBasedResourceTypeIds {
+			_, foundBrt, err := WorkerBaseResourceType{}.FindById(runner, workerBasedResourceTypeId)
```
clarafu, line 96: I don't think we need to verify the worker base resource type here, unless I'm missing something? But from what I can tell, we grab the `worker_base_resource_type_id` from the `workerResourceCache.find` query right before. I think we can trust that `worker_base_resource_type_id` is valid because that column has a foreign key to the `worker_base_resource_types` table with an `ON DELETE CASCADE`.

clarafu (changes requested):
Thanks so much for working in the changes 😊  The relationship between objects and tables with workers and resource caches are quite complicated but incredibly delicate. I just have a few changes that I would like for you to take a look at but it mostly looks great!

I think you are correct in that I think we also don't need to GC duplicate resource caches on a worker. I think they can be gced naturally when there are no longer any uses for them and there isn't much harm in having duplicates.

atc/db/worker_resource_cache.go
```diff
@@ -19,10 +18,14 @@ type UsedWorkerResourceCache struct {
 var ErrWorkerBaseResourceTypeDisappeared = errors.New("worker base resource type disappeared")
 
 func (workerResourceCache WorkerResourceCache) FindOrCreate(tx Tx) (*UsedWorkerResourceCache, error) {
```
aoldershaw, line 20: Maybe it's better to just have `sourceWorkerName` as an argument to `FindOrCreate`, rather than adding it to the struct when its never used for `Find`

atc/db/worker_resource_cache.go
```diff
@@ -32,32 +35,40 @@ func (workerResourceCache WorkerResourceCache) FindOrCreate(tx Tx) (*UsedWorkerR
 		return nil, ErrWorkerBaseResourceTypeDisappeared
 	}
 
-	id, found, err := workerResourceCache.find(tx, usedWorkerBaseResourceType)
+	ids, workerBaseResourceTypeIds, found, err := workerResourceCache.find(tx, workerResourceCache.WorkerName)
```
aoldershaw, line 38: Can we get rid of the `workerName` param here?

atc/exec/get_step.go
```diff
@@ -281,3 +313,51 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
...
+	workerSpec worker.WorkerSpec) (worker.Volume, bool) {
+	workers, err := step.workerPool.FindWorkersForResourceCache(logger, teamId, resourceCache.ID(), workerSpec)
+	if err != nil {
+		return nil, false
+	}
+
+	for _, sourceWorker := range workers {
+		volume, found, err := sourceWorker.FindVolumeForResourceCache(logger, resourceCache)
+		if err != nil {
+			logger.Debug("ignore-error", lager.Data{"error": err})
```
aoldershaw, line 353: Can we give a little more context for this log message? Either adding a `Session` (indicating that we're `finding-resource-cache`) or changing the message to something like `ignore-find-resource-cache-error`. Maybe we could also add the worker name in the `lager.Data`

atc/worker/volume.go
```diff
@@ -38,6 +39,8 @@ type Volume interface {
 
 	WorkerName() string
 	Destroy() error
+
+	Volume() Volume
```
aoldershaw, line 43: Can we get rid of this?

atc/worker/volume.go
```diff
@@ -30,6 +30,7 @@ type Volume interface {
 	COWStrategy() baggageclaim.COWStrategy
 
 	InitializeResourceCache(db.UsedResourceCache) error
+	InitializeStreamedResourceCache(db.UsedResourceCache, string) error
```
aoldershaw, line 33: If we remove the condition of `""` as being valid for `db.CreatedVolume.InitializeResourceCache`, then we can get rid of this as well

atc/worker/worker.go
```diff
@@ -627,7 +658,7 @@ func (worker *gardenWorker) cloneRemoteVolumes(
 			},
 			container,
 			teamID,
-			nonLocalInput.desiredMountPath,
+			"",
```
aoldershaw, line 661: Does this work when there are multiple inputs that need to be streamed? `FindOrCreateVolumeForContainer` calls `db.VolumeRepository.FindContainerVolume` which looks for a volume on the worker with `c.container_id == container.ID()` and `c.path == mountPath` - but if we set the mount path to `""` for each non-local volume, if we have 2 remote inputs, won't  `FindOrCreateVolumeForContainer` think that the volume already exists and just create 1 volume?

aoldershaw (changes requested):
Sorry for taking so long to review this! Have some suggestions and a question about streaming multiple non-local volumes

atc/worker/worker.go
aoldershaw: If my thinking is correct, then *maybe* this could be hacked around by prefixing the `desiredMountPath` with some string that couldn't form a valid path (e.g. `/tmp/build/foo` -> `streamed:/tmp/build/foo`). But this isn't the most satisfying solution 😁

atc/db/team.go
```diff
@@ -162,6 +163,18 @@ func (t *team) FindWorkerForVolume(handle string) (Worker, bool, error) {
 	}))
 }
 
+func (t *team) FindWorkersForResourceCache(rcId int) ([]Worker, error) {
+	return getWorkers(
+		t.conn, workersQuery.
+			Join("volumes v ON v.worker_name = w.name").
```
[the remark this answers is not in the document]
evanchaoli: Good catch. This func was written before `worker_name` was added, and I forgot to update this function when adding `worker_name` to `worker_resource_caches`.

atc/db/worker_resource_cache.go
evanchaoli: Yes, `ids` and `workerBasedResourceTypeIds` are nil, but I want the third returned value to be `false`.

atc/worker/worker.go
evanchaoli: @aoldershaw Good catch! And a big issue. I missed to test multiple task inputs.

atc/db/volume.go
```diff
@@ -351,17 +351,25 @@ func (volume *createdVolume) findWorkerBaseResourceTypeByBaseResourceTypeID(base
 	}, nil
 }
 
-func (volume *createdVolume) InitializeResourceCache(resourceCache UsedResourceCache) error {
+// InitializeResourceCache creates a worker resource cache and point current volume's
+// worker_resource_cache_id to the cache. When initialize a resource cache that is
+// generated on current worker, then sourceWorkerName can be empty.
```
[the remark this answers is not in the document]
evanchaoli: @aoldershaw I'm not following. Can you be more specific?

atc/db/worker_resource_cache.go
evanchaoli: Good suggestion.

atc/db/worker_resource_cache.go
evanchaoli: Yes

atc/worker/volume.go
evanchaoli: Yes. Forgot to delete it.

atc/exec/get_step.go
evanchaoli: As we now mark streamed volume as resource cache, it may find multiple workers containing target resource cache. Then this `for` verify resource cache on each worker till find. A worker may crash (or other failures) in the same time, that should not break the current build, because other worker may also provide the resource cache. That's reason why we ignore the error here. I can revise the error message.

atc/db/volume.go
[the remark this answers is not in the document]
evanchaoli: @aoldershaw I intentionally keep two functions: `InitializeResourceCache` and `InitializeStreamedResourceCache`, because I consider they are logically different, one for initialize local generated resource cache, the other for initialize streamed-in resource cache. Keeping them separate helps make caller easier to decide how to use them.

With your comment, I think we should also make them different in module `db`, so I made the change in the last commit.

atc/worker/volume.go
evanchaoli: As replied to the previous comment, I intentionally added it.

atc/worker/worker.go
```diff
@@ -658,7 +658,7 @@ func (worker *gardenWorker) cloneRemoteVolumes(
 			},
 			container,
 			teamID,
-			"",
+			filepath.Join("/streamed", nonLocalInput.desiredMountPath),
```
aoldershaw, line 661: One problem with this is that `/streamed/desired/mount/path` *could* be a valid `desiredMountPath` for another input. Granted, it's unlikely to come up, but still a potential bug. I think my original suggestion of `streamed:/desired/mount/path` technically couldn't come up (at least with the current state of the code), but for a non-obvious reason (somewhere higher up, we `filepath.Join` the input path to the working directory, which is always an absolute path).

Either way, it's all kind of a hack. I guess one other way would be for there to be a separate `streamed: bool` column in `volumes` and `FindOrCreateVolumeForContainer` takes the `path` and whether or not it's streamed, and when we try to find the volume, we match on both fields. This is effectively doing the same thing as just prefixing the path with `/streamed/...` or `streamed:...`, but is a bit more explicit.

WDYT? cc @clarafu

evanchaoli: @aoldershaw I don't feel this would be a potential bug, because Concourse doesn't support absolution task input path, thus `/streamed/...` shouldn't be conflict to any input paths. Even if that's really an issue, we can workaround with a UUID `/streamed/<uuid>` as mount path, that won't be conflict to any path.

evanchaoli:
@clarafu @aoldershaw I have addressed all review comments, and reverted changes in `docker-compose.yml`. Unless you have further comment, this PR is ready to merge.

aoldershaw (changes requested):
Found a bug related to streaming image artifacts. With the following `docker-compose.yml` modifications:

```diff
diff --git a/docker-compose.yml b/docker-compose.yml
index 896626156..5dc90ad82 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -42,9 +42,6 @@ services:
     command: worker
     privileged: true
     depends_on: [web]
-    ports:
-    - 7777:7777
-    - 7788:7788
     volumes: ["./hack/keys:/concourse-keys"]
     stop_signal: SIGUSR2
     environment:
@@ -64,3 +61,33 @@ services:
 
       # work with docker-compose's dns
       CONCOURSE_CONTAINERD_DNS_PROXY_ENABLE: "true"
+
+      CONCOURSE_TAG: worker1
+
+  worker2:
+    build: .
+    image: concourse/concourse:local
+    command: worker
+    privileged: true
+    depends_on: [web]
+    volumes: ["./hack/keys:/concourse-keys"]
+    stop_signal: SIGUSR2
+    environment:
+      CONCOURSE_RUNTIME: containerd
+
+      CONCOURSE_TSA_PUBLIC_KEY: /concourse-keys/tsa_host_key.pub
+      CONCOURSE_TSA_WORKER_PRIVATE_KEY: /concourse-keys/worker_key
+
+      CONCOURSE_LOG_LEVEL: debug
+      CONCOURSE_TSA_HOST: web:2222
+
+      CONCOURSE_BIND_IP: 0.0.0.0
+      CONCOURSE_BAGGAGECLAIM_BIND_IP: 0.0.0.0
+
+      # avoid using loopbacks
+      CONCOURSE_BAGGAGECLAIM_DRIVER: overlay
+
+      # work with docker-compose's dns
+      CONCOURSE_CONTAINERD_DNS_PROXY_ENABLE: "true"
+
+      CONCOURSE_TAG: worker2
```

...and using the following pipeline:

```yaml
resources:
- name: some-image
  type: mock
  source:
    mirror_self: true
  tags: [worker1]

jobs:
- name: job
  plan:
  - get: some-image
    tags: [worker1]
  - task: task1
    image: some-image
    config:
      platform: linux
      run:
        path: sh
        args:
        - -c
        - echo hi > /shared
    tags: [worker2]
  - task: task2
    image: some-image
    config:
      platform: linux
      run:
        path: cat
        args: [/shared]
    tags: [worker2]
```

...you can reproduce the bug - `task2` shouldn't be able to access `/shared`, but it is able to.

The problematic bit of code is here: https://github.com/concourse/concourse/blob/286ce71f000025b417e9ad2586567a24f0934bee/atc/worker/image/image.go#L86-L110

Specifically, we shouldn't be using the fresh streamed volume as the rootfs for the container - we should create a COW volume of it and use that, exactly as you've done with the remote input volumes.

evanchaoli:
> Specifically, we shouldn't be using the fresh streamed volume as the rootfs for the container - we should create a COW volume of it and use that, exactly as you've done with the remote input volumes.

Thanks for the catching. Yeah, I missed to update this place.

Just fixed the bug and tested your test case:

test-image-volume

I think this is the expected result.

atc/worker/worker.go
clarafu: Hmm to be honest I'm fine with either, the advantages I could see with having a streamed boolean would be that it would allow us to easily query for streamed volumes vs non streamed volumes. Although, I can't really think of a need for that now other than to show the user in a `fly volumes` command.

clarafu:
I think this would benefit from adding some [integration tests](https://github.com/concourse/concourse/tree/master/integration/worker) for caching streamed volumes and also the gcing of streamed volumes when the parent worker disappears. We should probably get some integration tests for it in before merging, the integration test suite is pretty new so if you don't feel comfortable with adding them then we can help out with that.

evanchaoli:
@clarafu I have added a integration test case, please take a look.

atc/db/migration/migrations/1619180099_add_worker_name_to_worker_resource_caches.up.sql
```diff
@@ -0,0 +1,12 @@
...
+
+DROP INDEX worker_resource_caches_uniq;
+
+CREATE UNIQUE INDEX worker_resource_caches_uniq
+    ON worker_resource_caches (worker_name, resource_cache_id, worker_base_resource_type_id);
+
+-- populate worker-name for existing caches with current worker name
+UPDATE worker_resource_caches wrc
+    SET worker_name = (SELECT worker_name FROM worker_base_resource_types wbrt WHERE wbrt.id = wrc.worker_base_resource_type_id)
+    WHERE worker_name is null
```
clarafu, line 12: nit: 
```suggestion
    WHERE worker_name is null;
```

atc/db/worker_resource_cache.go
```diff
@@ -84,40 +95,45 @@ func (workerResourceCache WorkerResourceCache) Find(runner sq.Runner) (*UsedWork
...
-	id, found, err := workerResourceCache.find(runner, usedWorkerBaseResourceType)
-	if err != nil {
-		return nil, false, err
-	}
-
-	if found {
-		return &UsedWorkerResourceCache{
-			ID: id,
-		}, true, nil
+	// If there are multiple workers found, choose a random one.
```
clarafu, line 98:
```suggestion
	// If there are multiple worker resource caches found, choose a random one.
```

atc/exec/get_step.go
```diff
@@ -193,6 +194,37 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
 		return false, err
 	}
 
+	getResult, found, err := step.getFromLocalCache(logger, step.metadata.TeamID, resourceCache, workerSpec)
+	if err != nil {
+		return false, err
+	}
+	if found {
+		fmt.Fprintln(delegate.Stderr(), "\x1b[1;36mINFO: found resouce cache from local cache\x1b[0m")
```
clarafu, line 202:
```suggestion
		fmt.Fprintln(delegate.Stderr(), "\x1b[1;36mINFO: found resource cache from local cache\x1b[0m")
```

atc/worker/worker.go
```diff
@@ -619,6 +650,10 @@ func (worker *gardenWorker) cloneRemoteVolumes(
 	for i, nonLocalInput := range nonLocals {
 		// this is to ensure each go func gets its own non changing copy of the iterator
 		i, nonLocalInput := i, nonLocalInput
+		// create an empty volume to stream-in the remote volume. this volume will only
```
clarafu, line 653: Great comments! 😄

clarafu (commented):
Very small suggestions/nits, thanks for adding the integration test! It looks good, I think we can try it on CI to see if we can catch anything else.

evanchaoli:
@clarafu Pushed a new commit. Hopefully it's ready now.

aoldershaw (approved):
Thanks for your patience! The integration test looks good. Will wait for @clarafu before merging, but LGTM

clarafu (approved):
LGTM too, thanks so much for all your work @evanchaoli and taking all our feedback, it's super appreciated! Hopefully this time works well 🤞

marco-m:
@odormond FYI this is in the released 7.3.0 and could help us.

lrstanley:
One potential issue I have seen with this so far, is if the resource we're fetching in the get step, utilizes `$BUILD_` environment variables (e.g. a resource that generates html files for notification purposes), those things are cached, and thus would be inconsistently used. I am not sure if this applies for any situation or just for those using global resources like we are.

Is there anyway to specify disabling of resource volume caching for a specific resource? I could see this for situations where one doesn't want to cache very large resource volumes, but enable caching for things that are expected to be small.

evanchaoli:
@lrstanley I believe using BUILD environment when generating get artifacts is not a good idea. Even without this feature, `get` still have a chance to reuse cache via previous get container.

lrstanley:
Don't think we've ever had that be the case. Prior to enabling the cache functionality, and after we disabled it, the issues immediately went away. I guess I'm not sure which case that would be? What happens if the resource provisions credentials per call? Ones that don't have source provisionings defined? Those credentials would also be shared? Doesn't it make sense to allow resources to opt out of caching (I also personally feel like resources should be able to opt out of global resources too for the same reason..)? These things are relatively small in API scope, and I wouldn't think difficult in logic either.

As far as BUILD environment variables, and if they're recommended or not, I think it's a bit silly that they aren't recommended for use in get steps. They're not recommended in regular tasks, that's why they're not there (as per the website). If they're not recommended for get steps, **then they should be deprecated**. If they are a supported feature, then I think it makes sense to at least take them into consideration with other functionality. I.e. they're not experimental.

Furthermore, I really dislike that they aren't enabled for tasks. One has to try really hard to make some of these values "stateful", and if someone is already going through that much effort, they could do it regardless of if those variables are there or not. We've wanted to take advantage of these variables many times, for the specific usecase they were designed for (notifications, logging, purely metadata/addl info, nothing that would break a pipeline or cause state issues), but Concourse doesn't "trust" it's users, even though users can bypass the limitations they put in place anyway, and that's rather unfortunate.

Using put/ensure has been very painful for us, especially for those pipelines that have multiple if/elses for how notifications should be sent, where they're sent, and when they're sent. Additionally, with the new experimental instance vars, having the ability for one pipeline to `set_pipeline` other pipelines, it makes sense to expose the `BUILD` params more. Because if Concourse doesn't, users are just going to figure out a way of doing it anyway (passing in various variables as a replacement -- all inconsistent and more prone to things breaking). Why not allow it to happen, but control the way in which it's done?

At this point, our clusters run over 5800 (projected around 9000 in 1 year or so) pipelines with over 16000 jobs with many different varying usecases, but even though many pipelines reference they're own passed in versions of pipeline name/team/etc, I have yet to see state mapped externally that could really cause issues.

evanchaoli:
@lrstanley I think I just told a truth that you haven't encountered. If a job `get` same version as previous builds, it's possible to get cached artifacts, which randomly happens (e.g. the `get` is scheduled on the same worker of the previous build's `get` step, and the `get` container is still there), unlike this PR make that to always happen. That's reason why I said using BUILD environment in `get` is not an good idea.

I also requested a couple of times to expose BUILD environments to tasks, for that, I stand with you. :)

The PR is based on an assumption that, given a version, `get` should generate same artifacts.  Our cluster has 4000+ pipelines, which brings very heavy workload to our GitLab and our internal docker image registry. For example, given a git commit sha, `get` should always clone code on that sha. The original intention of this PR is to reduce workloads that Concourse drop to other systems.  Once a docker image is pulled, it will no longer need to pull it again.

I agree that, for notification stuff, build environment is useful, but for that, mostly we use `put`. I remember Concourse has an item in the roadmap to support notification, but before that, it may make sense to provide an option to opt-out local cache on specific resource. It should be easy to add. I'm current occupied by #7208. After that, I will try to add such an option on `get` step, and hopefully catch up with 7.4.0.

PR #6662, opened 2021-03-12
Add ability to navigate to resources page from build page

chenbh:

closes # .

Screen Shot 2021-03-12 at 5 36 40 PM

- add ability to open the Resource page to a specific version. If the version exists, then it will be displayed at the top of the page and is automatically expanded
- have the get/put steps in the Build page make use of this

## Release Note

UI: clicking on the version text for a get/put step in the Build page will now navigate directly to the Resource page with the corresponding version expanded

## Contributor Checklist
- [x] [Signed] all commits

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/web/elm/src/Build/StepTree/StepTree.elm
+++ b/web/elm/src/Build/StepTree/StepTree.elm
@@ -61,47 +61,48 @@ import Views.Spinner as Spinner
 
 
 init :
-    Highlight
+    Maybe Concourse.JobBuildIdentifier
+    -> Highlight
     -> Concourse.BuildResources
     -> Concourse.BuildPlan
     -> StepTreeModel
-init hl resources plan =
+init buildId hl resources plan =
     let
         step =
             constructStep plan
     in
     case plan.step of
         Concourse.BuildStepTask _ ->
-            step |> initBottom hl resources plan Task
+            step |> initBottom buildId hl resources plan Task
 
         Concourse.BuildStepCheck _ ->
-            step |> initBottom hl resources plan Check
+            step |> initBottom buildId hl resources plan Check
 
         Concourse.BuildStepGet name version ->
             step
                 |> setupGetStep resources name version
-                |> initBottom hl resources plan Get
+                |> initBottom buildId hl resources plan Get
 
         Concourse.BuildStepPut _ ->
-            step |> initBottom hl resources plan Put
+            step |> initBottom buildId hl resources plan Put
 
         Concourse.BuildStepArtifactInput _ ->
-            step |> initBottom hl resources plan ArtifactInput
+            step |> initBottom buildId hl resources plan ArtifactInput
 
         Concourse.BuildStepArtifactOutput _ ->
-            step |> initBottom hl resources plan ArtifactOutput
+            step |> initBottom buildId hl resources plan ArtifactOutput
 
         Concourse.BuildStepSetPipeline _ _ ->
-            step |> initBottom hl resources plan SetPipeline
+            step |> initBottom buildId hl resources plan SetPipeline
 
         Concourse.BuildStepLoadVar _ ->
-            step |> initBottom hl resources plan LoadVar
+            step |> initBottom buildId hl resources plan LoadVar
 
         Concourse.BuildStepInParallel plans ->
-            initMultiStep hl resources plan.id InParallel plans Nothing
+            initMultiStep buildId hl resources plan.id InParallel plans Nothing
 
         Concourse.BuildStepDo plans ->
-            initMultiStep hl resources plan.id Do plans Nothing
+            initMultiStep buildId hl resources plan.id Do plans Nothing
 
         Concourse.BuildStepAcross { vars, steps } ->
             let
@@ -119,7 +120,7 @@ init hl resources plan =
                         }
                    )
                 |> Just
-                |> initMultiStep hl resources plan.id (Across plan.id vars values) (Array.fromList plans)
+                |> initMultiStep buildId hl resources plan.id (Across plan.id vars values) (Array.fromList plans)
                 |> (\model ->
                         List.foldl
                             (\plan_ ->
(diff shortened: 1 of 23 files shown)


aoldershaw:
This is a very nice feature to have - frequently I've found myself copy/pasting the version I care about and pasting it into the browser's "find" feature, which is slow + annoying + tedious if the resource isn't on the first page.

A couple notes:
* Based on how we add arrows to tooltips (they're positioned centered on the element, regardless of where the tooltip actually is), the tooltips look weird when the version table is wide since the arrow is just hovering:
    Screen Shot 2021-03-15 at 11 48 13 AM
    * We can either try to center the tooltip box (which requires hard-coding its width), or just get rid of the arrow (which is the approach I took with other potentially wide elements)
* We shouldn't display this tooltip for anonymous `image_resources` - currently, it tries to take you to the resource page for a resource named `image`, since that's the step name
* Generally, we shouldn't be using the step name to determine which resource page to go to - we should use the `resource` field
  * e.g. if you have the following `get` step:
  ```yaml
  - get: foo
    resource: bar
  ```
  the link takes you to `.../resources/foo`, rather than `.../resources/bar`. I think we can also use this to determine if the resource is an `image_resource`, since `resource` should be empty

Note: we already expose the `resource` on the public plan: https://github.com/concourse/concourse/blob/28153d0cb8dd6f670d09a2001bd7b27d74ddf217/atc/public_plan.go#L183

...so just need to modify the JSON decoding to decode that in Elm:
https://github.com/concourse/concourse/blob/28153d0cb8dd6f670d09a2001bd7b27d74ddf217/web/elm/src/Concourse.elm#L665-L669

chenbh:
Good call, I didn't even think about anonymous gets

web/elm/src/Build/StepTree/StepTree.elm
```diff
@@ -1274,6 +1340,20 @@ tooltip model { hovered } =
...
 
+        HoverState.Tooltip (StepVersion _) _ ->
+            Just
+                { body =
+                    Html.div
+                        Styles.changedStepTooltip
+                        [ Html.text "view in resources page" ]
+                , attachPosition =
+                    { direction = Tooltip.Top
+                    , alignment = Tooltip.End
```
aoldershaw, line 1351: Could we actually to `Tooltip.Start`? Might just be me, but when I was trying to use this feature, I tended to put my cursor closer to the start of the version table (near where the keys are). When the version is particularly wide (like  with the `registry-image`), the tooltip appears pretty far away

web/elm/src/Routes.elm
```diff
@@ -188,9 +190,34 @@ resource =
...
         )
 
 
+resourceVersion : String -> Query.Parser (Maybe Concourse.Version)
+resourceVersion key =
+    let
+        split s =
+            case String.split ":" s of
+                [ k, v ] ->
+                    Just ( k, v )
```
aoldershaw, line 203: We'll need to rethink this a bit, as it doesn't work if the key/value contains a `:` (which happens a lot, actually, since the `digest` of the `registry-image` resource has a `:`) - if I click on the version of a resource using `registry-image`, it just strips out the `digest` from the URL and only keeps the `tag`.

The easiest solution would be to just encode a full JSON object and throw that in a single `filter` query param.

Another fairly easy solution is for `Concourse.versionQuery` to `URL.percentEncode` the key and value separately, and `URL.percentDecode` them separately here as well. That way, they end up getting doubly-encoded. A similar approach is to base64 encode them, but that makes the query string a little opaque.

If you want to avoid an ugly query string (because of all the special characters that need escaping in the other approaches), you could do something like I did with instance vars (`vars.key1="value1"&vars.key2="value2"`) - the serialization/parsing is all in https://github.com/concourse/concourse/blob/master/web/elm/src/DotNotation.elm. Since it's just a `Dict String String` it's even simpler - you don't need to worry about recursing through JSON objects or quoting strings with special characters. However, this approach requires a bit of a hack since the query parameter names are dynamic:
https://github.com/concourse/concourse/blob/ecda93e38208d0971e280076f9fbe8864d8e90dc/web/elm/src/Routes.elm#L521-L532

I personally think the JSON approach is the easiest and probably the best - I tried to make the instance vars query params more friendly since they appear all the time, but in this case you probably won't need to see it that often.

aoldershaw (changes requested):

web/elm/src/Routes.elm
chenbh: I originally had it in the JSON approach, but I took that out to make it more similar to the atc api logic. My thought was that if it fails here, then it'll also fail when the api server tries to decode it. What do you think of just taking the head of the list as the key and concatenate the rest into a string for the value?

https://github.com/concourse/concourse/blob/fc2fda3221c19f7ba39b2e9b96fa492463129e34/atc/api/resourceserver/versionserver/list.go#L34-L41

web/elm/src/Build/StepTree/StepTree.elm
chenbh: I briefly tried this before, but there was something weird about how it calculated the tooltip width and made it longer than necessary. I also liked the way the alignment was sorta consistent in style with the image fetching and step initialization timer tooltips.

Maybe I'll just bite the bullet and figure out the values to make it `Tooltip.Middle`

web/elm/src/Routes.elm
aoldershaw: Huh, wasn't aware of that! seems a bit problematic in general that the API server doesn't allow for keys that have a `:`, but oh well. I think your suggestion of taking the head as the key and the tail (joined by `:`) as the value is reasonable, since that'd match the behaviour of `SplitN(field, ":", 2)`

aoldershaw (approved):
LGTM!

web/elm/src/Build/StepTree/StepTree.elm
chenbh: Ended up doing `Tooltip.Middle`. 

> but there was something weird about how it calculated the tooltip width and made it longer than necessary.

This happens if the version text is short enough, the tooltip overflows the side of the screen into a newline, but the tooltip box doesn't take that into consideration and is longer than it need to be.
Screen Shot 2021-03-23 at 11 30 21 AM

But I'm going to say that versions text shorter than 9 characters (I think that's around the limit) are rare enough and we don't have to worry about it.

xtremerui:
if you approve fast enough would you beat the checks to merge it lol

chenbh:
> if you approve fast enough would you beat the checks to merge it lol

Not if the checks are marked as `Required`

PR #6705, opened 2021-03-22
enhance put.inputs detect to ingore prefixed . and ..

### LINKED ISSUE #6704: put step's "input: detect" fails to handle some cases
## Summary

We are in an effort to encourage all pipelines add "inputs: detect" to all put steps, which will help speed up put step's initialization, also saving Concourse cluster's resource, by avoid unnecessary volume streaming and mounts.

Some users reported `put` failures after adding `inputs: detects`. I checked those failures, and found that, some pipelines' `put` params are like:

```yaml
- put: some-resource
  params:
    file: ./build/some-file
    vars: ../vars/some-var-file
``` 

When a put needs multiple inputs, paths in params may be prefixed by `./` or `../`. In that case, as `detect` relies on the first path of paths, it will not be able to extract inputs from those paths.

## Steps to reproduce

As described in the above section.

## Expected results

Even if params paths are prefixed by `./` and `../`, inputs should be detected.

## Actual results

Inputs may not be detected.

## Triaging info

* Concourse version: 6.7.3 (other versions should be the same)
* Browser (if applicable): NA
* Did this used to work? No

evanchaoli:

closes #6704 .

Ignore prefixed `.` and `..` of put param paths.

## Release Note

Enhanced `put.inputs`:
* `input: detect` now can handle paths prefixed by `.` and `..`.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/exec/put_inputs.go
+++ b/atc/exec/put_inputs.go
@@ -83,8 +83,14 @@ func detectInputsFromParam(value interface{}) []build.ArtifactName {
 	switch actual := value.(type) {
 	case string:
 		input := actual
-		if idx := strings.IndexByte(actual, '/'); idx >= 0 {
-			input = actual[:idx]
+		if parts := strings.Split(actual, "/"); len(parts) > 1 {
+			for _, part := range parts {
+				if part == "." || part == ".." {
+					continue
+				}
+				input = part
+				break
+			}
 		}
 		return []build.ArtifactName{build.ArtifactName(input)}
 	case map[string]interface{}:
--- a/atc/exec/put_inputs.go
+++ b/atc/exec/put_inputs.go
@@ -45,13 +45,11 @@ func (i allInputs) FindAll(artifacts *build.Repository) (map[string]runtime.Arti
 
 type specificInputs struct {
 	inputs []string
-	params atc.Params
 }
 
-func NewSpecificInputs(inputs []string, params atc.Params) PutInputs {
+func NewSpecificInputs(inputs []string) PutInputs {
 	return &specificInputs{
 		inputs: inputs,
-		params: params,
 	}
 }
 
@@ -60,27 +58,14 @@ func (i specificInputs) FindAll(artifacts *build.Repository) (map[string]runtime
 
 	inputs := map[string]runtime.Artifact{}
 
-	for _, input := range i.inputs {
-		if input == atc.InputsDetect {
-			detectedInputs, err := NewDetectInputs(i.params).FindAll(artifacts)
-			if err != nil {
-				return nil, err
-			}
-
-			for k, v := range detectedInputs {
-				inputs[k] = v
-			}
-
-			continue
-		}
-
-		artifact, found := artifactsMap[build.ArtifactName(input)]
+	for _, i := range i.inputs {
+		artifact, found := artifactsMap[build.ArtifactName(i)]
 		if !found {
-			return nil, PutInputNotFoundError{Input: input}
+			return nil, PutInputNotFoundError{Input: i}
 		}
 
 		pi := putInput{
-			name:     build.ArtifactName(input),
+			name:     build.ArtifactName(i),
 			artifact: artifact,
 		}
 
--- a/atc/exec/put_inputs.go
+++ b/atc/exec/put_inputs.go
@@ -45,11 +45,13 @@ func (i allInputs) FindAll(artifacts *build.Repository) (map[string]runtime.Arti
 
 type specificInputs struct {
 	inputs []string
+	params atc.Params
 }
 
-func NewSpecificInputs(inputs []string) PutInputs {
+func NewSpecificInputs(inputs []string, params atc.Params) PutInputs {
 	return &specificInputs{
 		inputs: inputs,
+		params: params,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


taylorsilva (changes requested):
Could you add a test for this in `put_step_test.go`
https://github.com/concourse/concourse/blob/523345ca3b3e57a96e6cc8c53459579a0519b59c/atc/exec/put_step_test.go#L343-L348

One question that I still have after reading the issue; why would you put `../` as the path? Is this a quirk of the resource the `put` step represents in this instance? I'm worried that allowing users to specify `../` as the start of their path will confuse them when they try to think about how Concourse makes the inputs available to the resource. 

In the issue you gave this as the example
```yaml
- put: some-resource
  params:
    file: ./build/some-file
    vars: ../vars/some-var-file
```
So in the resource container, it's expected that both `build` and `vars` will be available at the same point in the filesystem.
```
/tmp/inputs
├── build
└── vars
```
Whereas the `../` syntax makes it seem like the artifact structure will look like:
```
/tmp
├── vars
└── inputs
    └── build
```

Are you able to explain why `../`?

evanchaoli:
> Could you add a test for this in `put_step_test.go`

Sure. I should have left a comment to say that, I planned to add new tests, just haven't got bandwidth to work on that yet. I'll add tests soon.

evanchaoli:
> Are you able to explain why `../`?

@taylorsilva I can list all 3 issues reported by our users after they adding `inputs: detect`.

1. The first one used `docker-image` resource:

```yaml
- put: img
  params:
     build: ./outputs
     dockerfile: ./outputs/Dockerfile
```

I asked the user to fix the problem by removing `./`. But we are not able to review all pipeline source code before they are submitted, and technically `./build/` equals to `build/`. So I think `detect` should support form of `./build`.

2. The second one use the built-in `cf` resource type:

```yaml
- put: app
  params:
    path: auth
    vars_files:
     - ../vars/vars.yml
     # other params
```
I didn't check source code of `cf-resource`, but I guess it changes current worker directory, so that var file has to use `../`. The user fixed the problem by changing `detect` to `inputs: [auth, vars]`.

3. The third error is caused by implicit path with `docker-image` resource:

```yaml
- put: docker-image
  params:
    build: "."
    dockerfile: "source/Dockerfile"
    # other params
```

The user fixed the error by changing `detect` to `inputs: [build, source, version]`. Here we can see that, `build` doesn't present in `params`, it should be a path that is used in Dockerfile. From this case, we can see that, it might be impossible to detect all dependent artifacts from `params`.

atc/exec/put_inputs.go
```diff
@@ -58,14 +60,27 @@ func (i specificInputs) FindAll(artifacts *build.Repository) (map[string]runtime
 
 	inputs := map[string]runtime.Artifact{}
 
-	for _, i := range i.inputs {
-		artifact, found := artifactsMap[build.ArtifactName(i)]
+	for _, input := range i.inputs {
+		if input == atc.InputsDetect {
+			detectedInputs, err := NewDetectInputs(i.params).FindAll(artifacts)
```
taylorsilva, line 65: What's the motivation for structuring it like this in the code, with `detect` being stored in `specificInputs`? What happens if a user has an input called `detect`? I realize this doesn't change how users need to write their pipelines, but if a user has:
```yaml
- put: resource
  inputs:
    - foo
    - bar
    - detect
```
and the input is actually called `detect` then won't we end up trying to "detect" inputs when we shouldn't?

Also feels odd to have `specificInputs` doing two things now: 1) finding specific inputs and 2) detecting inputs. I think this overloads what `specificInputs` does. Is there a good reason to do this? I don't see why we have to insert `detect` into the `inputs` array.

taylorsilva (changes requested):

taylorsilva:
> The second one use the built-in cf resource type:

I'm not at all familiar with the cf resource either. Not sure why they'd have to do that. I'm fine with keeping `../` in the detect logic since it appears it will help detect work better. Can't think of a case where it may break things for anyone or cause extra volumes to be streamed to the put step.

atc/exec/put_inputs.go
evanchaoli: @taylorsilva The motivation is from the third error I described above. Say, a `put` depends on 10 inputs (9 can be detected, 1 is implicit). After a user set `inputs: detect` to the `put`, it hits an error of "some folder not found" because the implicit input cannot be detected. Then the user can simply change to `inputs: [detect, implicit-one]`, instead of having to list all rest 9 inputs.

If a `put` step happens to have an input named "detect", both `inputs: detect` and `inputs: [detect, other-input]` would still work. Only cost of `inputs: [detect, other-input]` is parsing `params` which suppose to be a tiny effort, I won't worry about that.

Embedding `detect` into `specificInputs` is just an implementation detail, it doesn't impact user experience. This implementation avoids duplicate code if we want to support embedding "detect" in inputs array.

I also realize one thing that I didn't realize yesterday. Only problem of embedding "detect" in inputs array is, if a `put` happens to have an implicit input named "detect", that must be explicitly specified in inputs array, then my current design will not be able to support that. To solve that problem, maybe we can use an escape character, `inputs: [detect, \detect]`, where the first "detect" means to auto detect inputs from `params`, and the second "\detect" means the input named "detect" should be included. WDYT?

taylorsilva: This is when I wish I had a magic wand and could have concourse expose `inputs` as:
```yaml
- put: resource
    inputs:
      detect: true
      specific:
      - foo
      - bar
```
or something analogous to that.

The solution you proposed would work but it would be hacky and even documenting it feels awkward. https://concourse-ci.org/jobs.html#schema.step.put-step.inputs

I get that this commit gets "detect" closer to being a 100% solution for you, but I really don't like what it's doing to the code and the UX it gives users writing pipelines.

Currently, without this PR, the user and you as operator don't have zero options right now. Specifying specific inputs seems perfectly valid here, given some inputs are implicit. If the inputs do change the pipeline author will know on the first build with the change, given they don't catch it themselves. And if inputs do change they probably have to update the `params` anyways, right beside where the `inputs` are.

Using detect is convenient, you don't have to change the pipeline as much in the future. Though when I think about it, what's the workflow for someone using detect and changing the name of their inputs? They have to change the params anyways! They'll be modifying the step already. 

Right now I think adding `detect` as a keyword to the list of inputs would be confusing. It is neat but it complicates the code and people's understanding of pipelines. It won't be obvious on a first read if the `- detect` item is actually an input called detect or is it "detect"ing inputs. Someone will have to go into the docs or leave a comment in their pipeline about the special "detect" item. I'm not in favour of the change from this specific commit. You already need to advise your users to use `detect`, I don't think it's that much more of a stretch to tell them: "try detect first and then list all inputs if detect doesn't catch everything".

evanchaoli: @taylorsilva If that prevents this PR from proceeding, I comprise. Let me revert the change of adding `detect` to inputs list.

taylorsilva (approved):
Thanks Evan! LGTM

PR #6830, opened 2021-04-06
Re ordering instanced pipelines

### LINKED ISSUE #6617: Support re-ordering instanced pipelines
## Summary

Allow instanced pipelines to be re-ordered within their group from the UI/fly. Currently, instanced pipelines are displayed in the order they were created, and there's no way to manually re-arrange them.

Currently, it's required that instanced pipelines all have the same global ordering, and their IDs are used as a secondary ordering. This feature involves creating a new secondary ordering (perhaps a sub-ordering field in the DB?) and API endpoints to reorder an instance group

## Context

* Follow up to #6105, which omitted that feature for simplicity

EstebanFS:

closes #6617 

- Add functionality to allow instanced pipelines to be re-ordered within their group from the UI/fly

## Release Note

* Instanced Pipelines are allowed to be re-ordered with in their group through the UI (using the drag and drop functionality) or using the fly command: 
`fly -t dev oip -g groupName -p key1:var1 -p key2:var2`

## Contributor Checklist
- [x] [Signed] all commits
- [x] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


fly/commands/ordering_instanced_pipeline.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/concourse/concourse/go-concourse/concourse"
+
+	"github.com/concourse/concourse/fly/commands/internal/displayhelpers"
+	"github.com/concourse/concourse/fly/commands/internal/flaghelpers"
+	"github.com/concourse/concourse/fly/rc"
+)
+
+type OrderInstancedPipelinesCommand struct {
+	Pipeline     string                         `short:"g" long:"group" required:"true" description:"Name of the instance group"`
+	InstanceVars []flaghelpers.InstanceVarsFlag `short:"p" long:"pipeline" required:"true" description:"Instance vars identifying pipeline to order"`
```
chenbh, line 16: This can also be applied to `OrderPipelinesCommand`

```suggestion
	InstanceVars []flaghelpers.InstanceVarsFlag `short:"p" long:"pipeline" required:"true" description:"Instance vars identifying pipeline (can be specified multiple times to provide relative ordering)"`
```

fly/commands/internal/flaghelpers/instance_vars_flag.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (flag *InstanceVarsFlag) UnmarshalFlag(value string) error {
+	var err error
+	flag.InstanceVars, err = unmarshalInstanceVars(value)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
chenbh: Is it weird that we use `key=value` when setting a pipeline and `key:value` when referencing the pipeline?

web/elm/src/Dashboard/Dashboard.elm
```diff
@@ -675,24 +683,42 @@ update session msg =
 updateBody : Session -> Message -> ET Model
 updateBody session msg ( model, effects ) =
     case msg of
-        DragStart teamName cardId ->
-            ( { model | dragState = Models.Dragging teamName cardId }, effects )
+        DragStart card ->
+            ( { model | dragState = Models.Dragging card }, effects )
```
chenbh, line 687: Don't think it should be fixed in this PR, but I noticed a bug where the card doesn't render if I drag the boxes inside the card (i.e the jobs for pipelines or pipelines for instance groups).

https://user-images.githubusercontent.com/17577073/115045536-9796ce00-9ea4-11eb-8b4a-d5eefde737c5.mov

chenbh:
Tested both via UI and fly, everything looks good and just some minor comments

fly/commands/internal/flaghelpers/instance_vars_flag.go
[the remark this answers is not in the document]
aoldershaw: The `:`/`=` discrepancy was mentioned in the initial RFC: https://github.com/concourse/rfcs/blob/master/034-instanced-pipelines/proposal.md

> (Side note: `:` vs. `=` is a little weird but it's consistent with `fly check-resource` - we use `=` for assignment and `:` for partial filtering.)

web/elm/src/Api/Endpoints.elm
```diff
@@ -297,3 +309,12 @@ teamEndpoint endpoint =
...
     )
+
+
+instanceGroupEndpoint : InstanceGroupEndpoint -> RouteBuilder
+instanceGroupEndpoint endpoint =
+    ( case endpoint of
+        OrderInstanceGroupPipelines ->
+            [ "ordering" ]
+    , []
+    )
```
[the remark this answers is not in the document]
aoldershaw: `pipelineEndpoint` identifies individual pipelines via `teamName`, `pipelineName`, *and* `pipelineInstanceVars` - `instanceGroupEndpoint` is only used on instance groups, so has no `pipelineInstanceVars`

atc/db/team.go
```diff
@@ -431,22 +430,25 @@ func savePipeline(
...
 		if err != nil {
 			return 0, false, err
 		}
 		if ordering.Valid {
 			values["ordering"] = ordering.Int64
+			values["secondary_ordering"] = secondaryOrdering.Int64 + 1
 		} else {
 			values["ordering"] = sq.Expr("currval('pipelines_id_seq')")
+			values["secondary_ordering"] = 1
 		}
```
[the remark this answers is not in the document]
aoldershaw: Not really, we just need a consistent ordering here and it felt more natural that way, Also Ordering is not 0-base at the beginning it only change his ordering to 0-base after re-ordering some pipelines.

fly/commands/internal/flaghelpers/instance_vars_flag.go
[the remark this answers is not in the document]
aoldershaw: Moved most of the pipeline flag tests out to test just the instance vars flag

go-concourse/concourse/pipelines.go
```diff
@@ -57,6 +57,28 @@ func (team *team) OrderingPipelines(pipelineNames []string) error {
 	}, &internal.Response{})
 }
 
+func (team *team) OrderingPipelinesWithinGroup(groupName string, instanceVars []atc.InstanceVars) error {
```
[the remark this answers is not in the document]
aoldershaw: We switched go-concourse to use the now preferred `HttpAgent` rather than the `Connection`, and special case `http.StatusBadRequest` so that we can display a more friendly error message

atc/db/team.go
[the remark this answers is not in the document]
aoldershaw: To clarify the second part - we just take `currval('pipelines_id_seq')`, which will leave you with some value of `ordering` that is larger than any other pipeline currently.

`secondary_ordering` on the other hand is easy to keep as 1-based, and felt more natural to do so (since the `row_number` function we use in the migration is also 1-based, and it'd be a bit awkward to subtract 1 when all you really care about is the relative ordering).

That said, there isn't really a good reason for it - we're happy to reconsider this, I don't think either @EstebanFS or I feel particularly strongly about it one way or another

fly/commands/ordering_instanced_pipeline.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/concourse/concourse/go-concourse/concourse"
+
+	"github.com/concourse/concourse/fly/commands/internal/displayhelpers"
+	"github.com/concourse/concourse/fly/commands/internal/flaghelpers"
+	"github.com/concourse/concourse/fly/rc"
+)
+
+type OrderInstancedPipelinesCommand struct {
+	Pipeline     string                         `short:"g" long:"group" required:"true" description:"Name of the instance group"`
+	InstanceVars []flaghelpers.InstanceVarsFlag `short:"p" long:"pipeline" required:"true" description:"Instance vars identifying pipeline (can be specified multiple times to provide relative ordering)"`
```
clarafu, line 16: Was there a reason we wanted to have the flag name be `"pipeline"` instead of `"instance_var"` or `"i"` for short? I'm asking because for the `set-pipeline` command we use `"instance-var"` as the flag name where you pass in the instance vars that are to be set https://github.com/concourse/concourse/blob/34474b0f0cd5b6a46639b517e4b38d3d4250f6e7/fly/commands/set_pipeline.go#L29
With how the `set-pipeline` command is used, I thought that a `pipeline` can have multiple `instances` but for this `order-instance-pipelines` command it looks like a `group` (equivalent to `pipeline`) can have multiple `pipelines` (equivalent to `instances`). Just wondering what your thoughts are!

atc/db/team.go
```diff
@@ -732,6 +734,54 @@ func (t *team) OrderPipelines(names []string) error {
 	return nil
 }
 
+func (t *team) OrderPipelinesWithinGroup(groupName string, instanceVars []atc.InstanceVars) error {
```
clarafu, line 737: I guess this is more of a general question about ordering in general, but I was wondering what happens if you only order a few instances within a pipeline? For example, if I have 3 instances within a pipeline
```
instance1 -> secondary_ordering=1
instance2 -> secondary_ordering=2
instance3 -> secondary_ordering=3
```

And I run `fly -t <target> order-instance-pipelines -g <group> -p instance3 -p instance2`, it will set the `secondary_ordering` for `instance2` and `instance3`.
```
instance1 -> secondary_ordering=1
instance2 -> secondary_ordering=2
instance3 -> secondary_ordering=1
```

It will result in both `instance1` and `instance3` having `secondary_ordering` of 1, and in the UI I would see the ordering to be instance1 -> instance3 -> instance2. Do we want to just ignore the ordering of instances that are not passed into the fly command ordering? I know this is also a problem with the regular `fly order-pipelines` so it might not be necessary to address this here.

clarafu (commented):
Tested it out manually, with fly commands and reordering in the UI and it all looks great!! :) All the code looks 👍  too :D

atc/db/team.go
chenbh: My opinion of it is that the `order-[instance-]pipelines` command allows you to specify the relative ordering between pipelines you care about. If you don't include it in the list, then you probably don't care about where it actually appears.

But then this is also my opinion after reading the docs, this could be a case of "it's not a bug, it's a feature!"

atc/db/team.go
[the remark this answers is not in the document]
chenbh: I don't particularly care either, I guess it's just odd that what is (essentially) a list starting at `1`

chenbh (approved):
lgtm

fly/commands/ordering_instanced_pipeline.go
EstebanFS: We tried to leave the call to this command as similar as possible to the [order-pipeline](https://concourse-ci.org/managing-pipelines.html#fly-order-pipelines) command (so I could be friendly for the user), that's why we left the `pipeline` name flag to identify each instance pipeline and we added another flag called `group` to identify pipeline.

We checked also the syntax of the [set-pipeline](https://concourse-ci.org/setting-pipelines.html#fly-set-pipeline) command, but we though it could be more confusing for the user that way, because of the way [order-pipeline](https://concourse-ci.org/managing-pipelines.html#fly-order-pipelines) command was created.

Also we didn't want it to match [set-pipeline](https://concourse-ci.org/setting-pipelines.html#fly-set-pipeline) because `-i` it's used in a different way here, `-i` in [set-pipeline](https://concourse-ci.org/setting-pipelines.html#fly-set-pipeline) always sets a single instance var using `KEY=VALUE`, while `-p` in `order-instanced-pipelines` uses `KEY1:VALUE1,KEY2:VALUE2,...`
which could introduce its own confusion, because `-p` is usually `GROUP/KEY1:VALUE1,KEY2:VALUE2,...,` while here we omit the `GROUP/` because it's redundant, but still, it's closer to `-p` in other commands than `-i` in set-pipeline, we think.

Also, you're conceptually ordering pipelines (`-p`), not ordering instance vars (`-i`)

atc/db/team.go
EstebanFS: Exactly

fly/commands/ordering_instanced_pipeline.go
clarafu: I see, I think I understand! I always saw instance pipelines as a subset of a pipeline, but I guess I should more be thinking about it as an instance pipeline is equal to just a pipeline and pipelines can be grouped. Thanks for the explanation!

I think a small suggestion I have is to rename the `Pipeline` field to `Group` so that it is clear that the field is referencing a group rather than a pipeline?

atc/db/team.go
clarafu: Oh ya I see that comment in the doc haha Does seem weird to keep it like that but honestly its not a big deal 😄

fly/commands/ordering_instanced_pipeline.go
EstebanFS: That's right, I just made the change there

clarafu (approved):
NOICE!

PR #6848, opened 2021-04-08
Add exception handling in bigint migration

### LINKED ISSUE #6725: Database migration failed - from 6.7.3 to 7.1.0
## Summary

Database migration failed (in init-container, running with `migrate --migrate-to-latest-version` args) upgrading concourse 6.7.3 to 7.1.0.

## Steps to reproduce

Concourse 6.7.3 already deployed and started.
Upgrade concourse to 7.1.0 deploying with concourse helm chart v15.0.1.

## Expected results

Init container (running with `migrate --migrate-to-latest-version` args) successed migrating concourse DB

## Actual results

Init container (running with `migrate --migrate-to-latest-version` args) crashed (almost instantly) with this log: 
` error: migration '1606068653_build_events_partitions_bigint.up.sql' failed and was rolled back: pq: index "pipeline_build_events_14_build_id" does not exist                 │
│ stream closed  `

## Additional context

Helm chart deployment v15.0.1.

## Triaging info

* Concourse version: 7.1.0
* Browser (if applicable): NA
* Did this used to work?

xtremerui:
sumbitting as draft for getting feedbacks of the solution while waiting for user response.

## What does this PR accomplish
Change the migration so it can ignore the error that caused by race condition. We don't really care about the error(table not exist) since ultimately it should be an no-op when pipeline is deleted. 

fixes #6725 

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/migration/migrations/1606068653_build_events_partitions_bigint.up.sql
+++ b/atc/db/migration/migrations/1606068653_build_events_partitions_bigint.up.sql
@@ -7,12 +7,17 @@ BEGIN
 FOR pipeline IN
   SELECT id, name FROM pipelines
 LOOP
-  RAISE NOTICE 'renaming old indexes for pipeline % (%)', pipeline.id, pipeline.name;
-  EXECUTE format('DROP INDEX pipeline_build_events_%s_build_id', pipeline.id);
-  EXECUTE format('ALTER INDEX pipeline_build_events_%s_build_id_event_id RENAME TO pipeline_build_events_%s_build_id_old_event_id', pipeline.id, pipeline.id);
+  BEGIN
+    RAISE NOTICE 'renaming old indexes for pipeline % (%)', pipeline.id, pipeline.name;
+    EXECUTE format('DROP INDEX IF EXISTS pipeline_build_events_%s_build_id', pipeline.id);
+    EXECUTE format('ALTER INDEX IF EXISTS pipeline_build_events_%s_build_id_event_id RENAME TO pipeline_build_events_%s_build_id_old_event_id', pipeline.id, pipeline.id);
 
-  RAISE NOTICE 'creating new indexes for pipeline % (%)', pipeline.id, pipeline.name;
-  EXECUTE format('CREATE UNIQUE INDEX pipeline_build_events_%s_build_id_event_id ON pipeline_build_events_%s (build_id, event_id)', pipeline.id, pipeline.id);
+    RAISE NOTICE 'creating new indexes for pipeline % (%)', pipeline.id, pipeline.name;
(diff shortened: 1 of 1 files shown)


atc/db/migration/migrations/1606068653_build_events_partitions_bigint.up.sql
```diff
@@ -7,12 +7,17 @@ BEGIN
...
+    EXECUTE format('ALTER INDEX IF EXISTS pipeline_build_events_%s_build_id_event_id RENAME TO pipeline_build_events_%s_build_id_old_event_id', pipeline.id, pipeline.id);
 
-  RAISE NOTICE 'creating new indexes for pipeline % (%)', pipeline.id, pipeline.name;
-  EXECUTE format('CREATE UNIQUE INDEX pipeline_build_events_%s_build_id_event_id ON pipeline_build_events_%s (build_id, event_id)', pipeline.id, pipeline.id);
+    RAISE NOTICE 'creating new indexes for pipeline % (%)', pipeline.id, pipeline.name;
+    EXECUTE format('CREATE UNIQUE INDEX IF NOT EXISTS pipeline_build_events_%s_build_id_event_id ON pipeline_build_events_%s (build_id, event_id)', pipeline.id, pipeline.id);
+  EXCEPTION
+  WHEN undefined_table then
+    RAISE NOTICE '%', SQLERRM;
+  END;
```
aoldershaw, line 20: Maybe we can make the exception handling tighter to the commands that could trigger them? i.e. wrap each command in its own try-catch. Maybe the `CREATE UNIQUE INDEX` would need to handle `duplicate_table` instead (just guessing the error code, not sure that's correct)

aoldershaw (commented):
I definitely like the addition of `IF NOT EXISTS`, since without it (and without #6727), it makes sense that the migration would continually fail

I wonder if we can make the try-catches tighter to the command that they operate on, rather than running everything in a single try-catch. e.g. if a pipeline is concurrently deleted, if we just handle the exception when we try to `DROP INDEX`, we won't end up modifying the build events indexes and will end up in a bad state

xtremerui:
The `pipeline_build_events_%s_build_id_event_id` index that needs to be modified is also on table `pipeline_build_events_%s` so if that table is gone that index would have been deleted already. 

In such case we could only try-catch the `CREATE UNIQUE INDEX` one with exception `undefined_table`.

aoldershaw (approved):
Makes sense to me. I think this *technically* won't guarantee the migration will succeed if tables are being deleted concurrently since `IF NOT EXISTS` doesn't enforce synchronization, but there's no precedence for handling that in our migrations

PR #6854, opened 2021-04-09
atc(feat): enhance webhook triggered checks

evanchaoli:

Say there are 100 pipelines holding a common `git` resource, with global resource enabled, it suppose to run only a single check for the resource at a time. Say the 100 pipelines all configure webhooks on the resource. When there is a commit pushed, GitLab will post 100 webhook messages as there are 100 hooks configured in GitLab. Those 100 webhooks will create 100 check builds, and as webhook triggered build is marked as "manually triggered", so these 100 check builds will all be executed, which sounds incorrect. Ideally it should run the first check build, rest 99 should reuse result of the first one.

As those 100 checks belong to the same resource config scope, they will be invoke in serial. As 100 webhooks coming in at same time, they will have very similar check build `create-time`. After the first check build is executed, it's `check-start-time` should be later than `create-time`. So we just need to check if build `create-time` is earlier than last check start time, it can just reuse last check result.

The reason why I use `last-check-start-time` is that, consider situation:

* At time1, a commit push triggers a webhook call. A check is triggered, and started.
* At time2, very close to time1, a commit pushed again. A new check is triggered. But the first check hasn't finished yet.
* At time3, the first check finishes. And the second check is invoked. If we use `last-check-end-time`, the second check will not run, which is wrong.

In the offline discussion, @clarafu proposed an idea of avoiding queuing check builds if the scope has a pending or running check. I guess that is not feasible, because upon receiving webhook call, it cannot figure out resource config scope yet, because source is only evaluated in `check-step`, and only after source is evaluated, scope can be found.

With my solution, scope's last-check-start-time and build's create-time are there already, so it's very easy to implement.

## Notes to reviewer:

The first commit is for design review. If code looks good to reviewers, I'll quickly add tests. I wish this PR may catch up with 7.2.

## Release Note

* Enhance webhook triggered checks. When multiple pipelines hold a common resource, and webhook calls against the common resource are sent to all pipelines at same time, without this enhancement, each webhook call will cause a check to run. With this enhancement, only a single check will run, which is the expected behavior as global resource.

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/resource_config_scope.go
+++ b/atc/db/resource_config_scope.go
@@ -33,11 +33,11 @@ type ResourceConfigScope interface {
 		logger lager.Logger,
 	) (lock.Lock, bool, error)
 
-	LastCheckStartTime() (time.Time, error)
+	LastCheckStartTime() (time.Time, bool, error)
 	UpdateLastCheckStartTime() (bool, error)
 
-	LastCheckEndTime() (time.Time, error)
-	UpdateLastCheckEndTime() (bool, error)
+	LastCheckEndTime() (time.Time, bool, error)
+	UpdateLastCheckEndTime(bool) (bool, error)
 }
 
 type resourceConfigScope struct {
@@ -53,34 +53,36 @@ func (r *resourceConfigScope) ID() int                        { return r.id }
 func (r *resourceConfigScope) Resource() Resource             { return r.resource }
 func (r *resourceConfigScope) ResourceConfig() ResourceConfig { return r.resourceConfig }
 
-func (r *resourceConfigScope) LastCheckStartTime() (time.Time, error) {
+func (r *resourceConfigScope) LastCheckStartTime() (time.Time, bool, error) {
 	var lastCheckStartTime time.Time
-	err := psql.Select("last_check_start_time").
+	var lastCheckSucceeded bool
+	err := psql.Select("last_check_start_time, last_check_succeeded").
 		From("resource_config_scopes").
 		Where(sq.Eq{"id": r.id}).
 		RunWith(r.conn).
 		QueryRow().
-		Scan(&lastCheckStartTime)
+		Scan(&lastCheckStartTime, &lastCheckSucceeded)
 	if err != nil {
-		return time.Time{}, err
+		return time.Time{}, false, err
 	}
 
-	return lastCheckStartTime, nil
+	return lastCheckStartTime, lastCheckSucceeded, nil
 }
 
-func (r *resourceConfigScope) LastCheckEndTime() (time.Time, error) {
+func (r *resourceConfigScope) LastCheckEndTime() (time.Time, bool, error) {
 	var lastCheckEndTime time.Time
-	err := psql.Select("last_check_end_time").
+	var lastCheckSucceeded bool
+	err := psql.Select("last_check_end_time, last_check_succeeded").
 		From("resource_config_scopes").
 		Where(sq.Eq{"id": r.id}).
 		RunWith(r.conn).
 		QueryRow().
-		Scan(&lastCheckEndTime)
+		Scan(&lastCheckEndTime, &lastCheckSucceeded)
 	if err != nil {
-		return time.Time{}, err
+		return time.Time{}, false, err
 	}
 
-	return lastCheckEndTime, nil
+	return lastCheckEndTime, lastCheckSucceeded, nil
 }
 
 // SaveVersions stores a list of version in the db for a resource config
@@ -212,7 +214,7 @@ func (r *resourceConfigScope) UpdateLastCheckStartTime() (bool, error) {
 
 	updated, err := checkIfRowsUpdated(tx, `
 		UPDATE resource_config_scopes
-		SET last_check_start_time = now()
+		SET last_check_start_time = now(), last_check_succeeded = false
 		WHERE id = $1
 	`, r.id)
 	if err != nil {
@@ -231,7 +233,7 @@ func (r *resourceConfigScope) UpdateLastCheckStartTime() (bool, error) {
 	return true, nil
 }
 
-func (r *resourceConfigScope) UpdateLastCheckEndTime() (bool, error) {
+func (r *resourceConfigScope) UpdateLastCheckEndTime(succeeded bool) (bool, error) {
 	tx, err := r.conn.Begin()
 	if err != nil {
 		return false, err
@@ -241,9 +243,9 @@ func (r *resourceConfigScope) UpdateLastCheckEndTime() (bool, error) {
 
 	updated, err := checkIfRowsUpdated(tx, `
 		UPDATE resource_config_scopes
-		SET last_check_end_time = now()
-		WHERE id = $1
-	`, r.id)
+		SET last_check_end_time = now(), last_check_succeeded = $1
+		WHERE id = $2
+	`, succeeded, r.id)
 	if err != nil {
 		return false, err
 	}
--- a/atc/db/resource_config_scope.go
+++ b/atc/db/resource_config_scope.go
@@ -56,7 +56,7 @@ func (r *resourceConfigScope) ResourceConfig() ResourceConfig { return r.resourc
 func (r *resourceConfigScope) LastCheckStartTime() (time.Time, bool, error) {
 	var lastCheckStartTime time.Time
 	var lastCheckSucceeded bool
-	err := psql.Select("last_check_start_time, last_check_succeeded").
+	err := psql.Select("last_check_start_time", "last_check_succeeded").
 		From("resource_config_scopes").
 		Where(sq.Eq{"id": r.id}).
 		RunWith(r.conn).
@@ -72,7 +72,7 @@ func (r *resourceConfigScope) LastCheckStartTime() (time.Time, bool, error) {
 func (r *resourceConfigScope) LastCheckEndTime() (time.Time, bool, error) {
 	var lastCheckEndTime time.Time
 	var lastCheckSucceeded bool
-	err := psql.Select("last_check_end_time, last_check_succeeded").
+	err := psql.Select("last_check_end_time", "last_check_succeeded").
 		From("resource_config_scopes").
 		Where(sq.Eq{"id": r.id}).
 		RunWith(r.conn).
@@ -214,7 +214,7 @@ func (r *resourceConfigScope) UpdateLastCheckStartTime() (bool, error) {
 
 	updated, err := checkIfRowsUpdated(tx, `
 		UPDATE resource_config_scopes
-		SET last_check_start_time = now(), last_check_succeeded = false
+		SET last_check_start_time = now()
 		WHERE id = $1
 	`, r.id)
 	if err != nil {
--- a/atc/db/resource_config_scope.go
+++ b/atc/db/resource_config_scope.go
@@ -11,6 +11,12 @@ import (
 	"github.com/concourse/concourse/atc/db/lock"
 )
 
+type LastCheck struct {
+	StartTime time.Time
+	EndTime   time.Time
+	Succeeded bool
+}
+
 //counterfeiter:generate . ResourceConfigScope
 
 // ResourceConfigScope represents the relationship between a possible pipeline resource and a resource config.
@@ -33,10 +39,8 @@ type ResourceConfigScope interface {
 		logger lager.Logger,
 	) (lock.Lock, bool, error)
 
-	LastCheckStartTime() (time.Time, bool, error)
+	LastCheck() (LastCheck, error)
 	UpdateLastCheckStartTime() (bool, error)
-
-	LastCheckEndTime() (time.Time, bool, error)
 	UpdateLastCheckEndTime(bool) (bool, error)
 }
 
@@ -53,36 +57,24 @@ func (r *resourceConfigScope) ID() int                        { return r.id }
 func (r *resourceConfigScope) Resource() Resource             { return r.resource }
 func (r *resourceConfigScope) ResourceConfig() ResourceConfig { return r.resourceConfig }
 
-func (r *resourceConfigScope) LastCheckStartTime() (time.Time, bool, error) {
-	var lastCheckStartTime time.Time
+func (r *resourceConfigScope) LastCheck() (LastCheck, error) {
+	var lastCheckStartTime, lastCheckEndTime time.Time
 	var lastCheckSucceeded bool
-	err := psql.Select("last_check_start_time", "last_check_succeeded").
+	err := psql.Select("last_check_start_time", "last_check_end_time", "last_check_succeeded").
 		From("resource_config_scopes").
 		Where(sq.Eq{"id": r.id}).
 		RunWith(r.conn).
 		QueryRow().
-		Scan(&lastCheckStartTime, &lastCheckSucceeded)
+		Scan(&lastCheckStartTime, &lastCheckEndTime, &lastCheckSucceeded)
 	if err != nil {
-		return time.Time{}, false, err
+		return LastCheck{}, err
 	}
 
-	return lastCheckStartTime, lastCheckSucceeded, nil
-}
-
-func (r *resourceConfigScope) LastCheckEndTime() (time.Time, bool, error) {
-	var lastCheckEndTime time.Time
-	var lastCheckSucceeded bool
-	err := psql.Select("last_check_end_time", "last_check_succeeded").
-		From("resource_config_scopes").
-		Where(sq.Eq{"id": r.id}).
-		RunWith(r.conn).
-		QueryRow().
-		Scan(&lastCheckEndTime, &lastCheckSucceeded)
-	if err != nil {
-		return time.Time{}, false, err
-	}
-
-	return lastCheckEndTime, lastCheckSucceeded, nil
+	return LastCheck{
+		StartTime: lastCheckStartTime,
+		EndTime:   lastCheckEndTime,
+		Succeeded: lastCheckSucceeded,
+	}, nil
 }
 
 // SaveVersions stores a list of version in the db for a resource config
--- a/atc/db/resource_config_scope.go
+++ b/atc/db/resource_config_scope.go
@@ -33,6 +33,7 @@ type ResourceConfigScope interface {
 		logger lager.Logger,
 	) (lock.Lock, bool, error)
 
+	LastCheckStartTime() (time.Time, error)
 	UpdateLastCheckStartTime() (bool, error)
 
 	LastCheckEndTime() (time.Time, error)
@@ -51,6 +52,21 @@ type resourceConfigScope struct {
 func (r *resourceConfigScope) ID() int                        { return r.id }
 func (r *resourceConfigScope) Resource() Resource             { return r.resource }
 func (r *resourceConfigScope) ResourceConfig() ResourceConfig { return r.resourceConfig }
+
+func (r *resourceConfigScope) LastCheckStartTime() (time.Time, error) {
+	var lastCheckStartTime time.Time
+	err := psql.Select("last_check_start_time").
+		From("resource_config_scopes").
+		Where(sq.Eq{"id": r.id}).
+		RunWith(r.conn).
+		QueryRow().
+		Scan(&lastCheckStartTime)
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	return lastCheckStartTime, nil
+}
 
 func (r *resourceConfigScope) LastCheckEndTime() (time.Time, error) {
 	var lastCheckEndTime time.Time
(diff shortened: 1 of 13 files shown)


evanchaoli:
I have tested this PR locally. My test model is:

1. Start a dev cluster with global resource enabled
2. Use the same pipeline YAML that contains 1 `git` resource and a task to set 4 pipelines. The `git` resource is configured with `webhook_token`.
3. Set webhooks for the 4 pipelines on gitlab.
4. Push a commit to git, then 4 webhooks calls are sent to the 4 pipelines at same time.

Without this PR, 4 webhooks will create 4 manually-triggered check builds, and each build will start a real check.
With this PR, 4 webhooks will create 4 manually-triggered check builds, and only 1-2 builds will start a real check, rest will reuse previous check result. The reason why sometime 2 checks are started is that, there is only 1 ATC, check builds are created in serial, it's possible that when the first check build is created and started, the 4th check build is just created, in that case, the 4th check build will run.

atc/engine/check_delegate.go
```diff
@@ -125,8 +125,16 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
...
+		// ignore interval for manually triggered builds.
+		lastCheckStartTime, err := scope.LastCheckStartTime()
+		if err != nil {
+			return nil, false, err
+		}
+		if d.build.CreateTime().Before(lastCheckStartTime) || d.build.CreateTime().Equal(lastCheckStartTime) {
+			shouldRun = false
+		} else {
+			shouldRun = true
+		}
```
vito, line 137:
```suggestion
		// avoid running redundant checks
		shouldRun = d.build.CreateTime().After(lastCheckStartTime)
```

atc/engine/check_delegate_test.go
```diff
@@ -183,8 +183,37 @@ var _ = Describe("CheckDelegate", func() {
...
+					fakeBuild.CreateTimeReturns(time.Now().Add(-5*time.Second))
+					fakeResourceConfigScope.LastCheckStartTimeReturns(time.Now(), nil)
+				})
+
+				It("returns false", func() {
+					Expect(run).To(BeFalse())
+				})
+			})
+
+			Context("when the build create time earlier than last build start time", func(){
```
vito, line 208:
```suggestion
			Context("when the build create time after last check start time", func() {
```

atc/engine/check_delegate_test.go
```diff
@@ -183,8 +183,37 @@ var _ = Describe("CheckDelegate", func() {
...
+					fakeResourceConfigScope.LastCheckStartTimeReturns(time.Now(), errors.New("some-error"))
+				})
+
+				It("return the error", func(){
+					Expect(runErr).To(HaveOccurred())
+					Expect(runErr).To(Equal(errors.New("some-error")))
+				})
+			})
+
+			Context("when the build create time earlier than last build start time", func(){
```
vito, line 197:
```suggestion
			Context("when the build create time earlier than last check start time", func() {
```

vito (changes requested):
makes sense to me - just some nitpicks!

also could you run `go fmt`? it looks like there are a few things it'll clean up (mainly `func(){` -> `func() {`)

thanks!

evanchaoli:
@vito I have addressed all review comments. But `watjs` still failed. The error doesn't look like related to my change. I tried to create the pipeline locally, and see the error is exact same as #6868:

abort-error

But I don't know why other `watjs` builds passed, only my builds failed.

evanchaoli:
@vito Finally, I fixed the `wats` test. For get/put/task embedded checks, `checkDelegate` holds a normal build instead of a check build.

atc/engine/check_delegate.go
```diff
@@ -124,12 +124,22 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
 	runAt := end.Add(interval)
 
 	shouldRun := false
-	if d.build.IsManuallyTriggered() {
-		// ignore interval for manually triggered builds
-		shouldRun = true
-	} else if !d.clock.Now().Before(runAt) {
-		// run if we're past the last check end time
+	if d.build.Name() != db.CheckBuildName {
```
vito, line 127: I don't think we should overload the `.Name()` in this way - that's supposed to be for display purposes only. Could you explain the issue in the `watsjs` suite that this fixed?

evanchaoli: A failed `wats` build: https://ci.concourse-ci.org/teams/contributor/pipelines/prs/jobs/watsjs/builds/529?vars.branch=%22master%22. I couldn't exactly tell  why it failed, as I had no way to access ATC logs for the failed build. But when I debugged the failure, I found that, if I reverted my change from `check_delegate.go`, `wats` test would pass. Then I further realized that, for `task` embedded check, `check-delegate` actually holds the current running build rather than a check build. In any case, comparing current running build's creating with the resource's last check time doesn't make much sense.

So I added the logic of checking if `check-delegate` holds a normal build or a check build. If that's a normal build, that indicates a step embedded check, in that case, always run the check.

If `build.Name()` should not be used to distinguish normal build and check build, then other way I can see if is to check `build.ResourceId()`. Only check build returns a resource id.

vito: FWIW it's fairly easy to run the `watsjs` suite directly; there are [instructions in the contributing docs](https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md#running-the-web-acceptance-tests-webwats).

I ran the tests locally without this fix and saw the same failure. What's actually happening is that the test aborts the build, which interrupts the first `check` step that runs to find the task's image's version. Here's a screenshot:

image

All of these `image`s refer to the same resource. Because the first check step bumped the `LastCheckStartTime`, the later check steps decide not to run. But because the original check was interrupted, it didn't actually save any versions, and there isn't a version for them to return.

So I don't think this is caused by it being a check build vs. a 'regular' build - it's more because they're deciding not to run even when there has never been a previous successful run.

I think there are a couple possible fixes that would be a bit more accuratethan checking the build model:

* Check if the scope has a latest version and return `true` if it doesn't. This is technically correct, but it would cost an additional query.
* Instead of checking `build.ResourceID()`, check for `plan.Resource != ""`. This is a tiny change but makes it a bit more precise IMO, since it'll only affect the step we're actually trying to optimize, rather than all `check` steps in the build's plan. A resource checking build could contain a `check` for its resource type's image, too, and maybe more `check` steps in the future when we implement notifications. The check delegate already has similar logic [in other parts of `WaitToRun`](https://github.com/concourse/concourse/blob/0af59c6a895b1f6acfeaade85d359699ddc29e8b/atc/engine/check_delegate.go#L80) so it's at least a bit more consistent.

The one downside I can think of with the second approach is that it doesn't technically fix the core issue: if we ever have a build plan that has a `check` for a pipeline resource followed by a `get` of its version it's possible to end up with the same error if the resource never successfully checked before. No build plans currently do this, and I don't think there's anything relevant on the roadmap, so I'm OK with it given that it's kind of an edge case.

evanchaoli: @vito I have seen the contributor instruction for how to run `watsjs` test, but `yarn test` is just stuck. I don't know what is wrong with my environment.

Thanks for pointing out the root cause. To solve the core problem, I think we can just added a new boolean column `last_check_succeeded` to `resource_config_scopes`, so that we can easily how if last check succeeded. If last check failed, then it should run the check.

Even after solved the core problem, I still think we should distinguish regular build and check build in `checkDelegate.WaitToRun()`. Because `build.IsManuallyTriggered()` means different for regular build and check build.

For a check build, manual triggered check and lidar triggered check may behave differently.  But for a regular build, step embedded checks should behave in the same way no matter the build is triggered automatically or by manual.

The latest commit addressed the both problems. Please take a look again.

atc/db/resource_config_scope.go
```diff
@@ -52,19 +53,36 @@ func (r *resourceConfigScope) ID() int                        { return r.id }
 func (r *resourceConfigScope) Resource() Resource             { return r.resource }
 func (r *resourceConfigScope) ResourceConfig() ResourceConfig { return r.resourceConfig }
 
-func (r *resourceConfigScope) LastCheckEndTime() (time.Time, error) {
+func (r *resourceConfigScope) LastCheckStartTime() (time.Time, bool, error) {
+	var lastCheckStartTime time.Time
+	var lastCheckSucceeded bool
+	err := psql.Select("last_check_start_time, last_check_succeeded").
```
vito, line 59:
```suggestion
	err := psql.Select("last_check_start_time", "last_check_succeeded").
```

atc/db/resource_config_scope.go
```diff
@@ -196,7 +214,7 @@ func (r *resourceConfigScope) UpdateLastCheckStartTime() (bool, error) {
 
 	updated, err := checkIfRowsUpdated(tx, `
 		UPDATE resource_config_scopes
-		SET last_check_start_time = now()
+		SET last_check_start_time = now(), last_check_succeeded = false
```
vito, line 217: is it necessary to set it to `false` here? seems like it'd be safe to leave it as the previous value since it won't be checked until the lock is released (after `UpdateLastCheckEndTime`)

i could see other things using this column in the future to indicate checking status (besides the initial check - i guess technically we could allow `NULL` for that), so it may be noisy to have it flip-flop to 'false' all the time. not a huge deal though - just wondering if this is actually needed and i missed something.

atc/engine/check_delegate.go
```diff
@@ -77,7 +77,7 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
 
 	// rate limit periodic resource checks so worker load (plus load on external
 	// services) isn't too spiky
-	if !d.build.IsManuallyTriggered() && d.plan.Resource != "" {
+	if !d.build.IsManuallyTriggered() && (d.plan.Resource != "" || d.plan.ResourceType != "") {
```
vito, line 80: i think this is fine but i'm a little surprised to see this here - did something require it?

atc/engine/check_delegate.go
```diff
@@ -95,7 +95,7 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
 	}
 
 	var lock lock.Lock = lock.NoopLock{}
-	if d.plan.Resource != "" {
+	if d.plan.Resource != "" || d.plan.ResourceType != "" {
```
vito, line 98: now that we're checking for this combination in a few places, i wonder if adding a method like `(CheckPlan).IsPeriodic` would make it more readable?

atc/engine/check_delegate.go
```diff
@@ -111,24 +111,37 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
...
 	shouldRun := false
-	if d.build.IsManuallyTriggered() {
-		// ignore interval for manually triggered builds
-		shouldRun = true
-	} else if !d.clock.Now().Before(runAt) {
-		// run if we're past the last check end time
+	if d.plan.Resource == "" && d.plan.ResourceType == "" {
 		shouldRun = true
+	} else {
+		if d.build.IsManuallyTriggered() {
```
vito, line 118: would be nice to avoid extra indent level here:

```suggestion
	} else if d.build.IsManuallyTriggered() {
```

atc/engine/check_delegate.go
```diff
@@ -111,24 +111,37 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
...
+			if err != nil {
+				if releaseErr := lock.Release(); releaseErr != nil {
+					logger.Error("failed-to-release-lock", releaseErr)
+				}
+				return nil, false, err
+			}
+
+			// ignore interval for manually triggered builds.
+			// avoid running redundant checks
+			shouldRun = !lastCheckSucceeded || lastCheckStartTime.IsZero() || d.build.CreateTime().After(lastCheckStartTime)
```
vito, line 129: i think this'll read a little nicer without the `.IsZero` check since the build create time will always be after the unix epoch anyhow:

```suggestion
			shouldRun = !lastCheckSucceeded || d.build.CreateTime().After(lastCheckStartTime)
```

vito (changes requested):
The new approach makes sense to me 👍 - but I just had a question and a few nitpicky suggestions.

Thanks for your patience!

atc/db/resource_config_scope.go
evanchaoli: Make sense.

atc/engine/check_delegate.go
evanchaoli: Resource check will cause resource type check, so I consider resource type check should be handled in the same logic as resource check.

vito (changes requested):
just missing a test and then i think we're good to go!

atc/engine/check_delegate.go
```diff
@@ -111,25 +111,42 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
...
+			if releaseErr := lock.Release(); releaseErr != nil {
+				logger.Error("failed-to-release-lock", releaseErr)
+			}
+			return nil, false, fmt.Errorf("get last check end time: %w", err)
 		}
 
-		return nil, false, fmt.Errorf("get last check end time: %w", err)
+		runAt := end.Add(interval)
+		if !d.clock.Now().Before(runAt) {
+			// run if we're past the last check end time
```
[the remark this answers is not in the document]
evanchaoli: This piece of code is not mine. But I agree, you refactor is better, I can make the change.

atc/engine/check_delegate.go
```diff
@@ -111,25 +111,42 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
...
-	end, err := scope.LastCheckEndTime()
-	if err != nil {
-		if releaseErr := lock.Release(); releaseErr != nil {
-			logger.Error("failed-to-release-lock", releaseErr)
+	shouldRun := false
+	if !d.plan.IsPeriodic() {
+		shouldRun = true
+	} else if d.build.IsManuallyTriggered() {
+		// If a manually triggered check takes a from version, then it should be run.
+		if d.plan.FromVersion != nil {
```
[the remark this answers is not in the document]
evanchaoli: Yeah, that's for `fly check-resource --from`. I ran more tests yesterday, and noticed that `testflight` intermittently failed, then I further identified the failure was at `caching_test`, and found the root cause was about `fromVersion`. A manually triggered check with `fromVersion` should always run.

evanchaoli:
@vito In the second last commit, I have added a test case to address your comment.

In the last commit, I added some further optimization for step embedded check, if a step embedded check has been done successfully after build start, then it will not run again during the build. This optimization should save a lot of checks in a large deployment when some common task images (like busybox, ubuntu) are used. At least, I feel `testflight` runs faster with this optimization. However, if you don't think that makes sense, I'm ok to revert it.

atc/engine/check_delegate.go
```diff
@@ -111,25 +118,56 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
 		}
 	}
 
-	end, err := scope.LastCheckEndTime()
-	if err != nil {
-		if releaseErr := lock.Release(); releaseErr != nil {
-			logger.Error("failed-to-release-lock", releaseErr)
+	shouldRun := false
+	if !d.plan.IsPeriodic() {
+		lastCheckEndTime, lastCheckSucceeded, err := scope.LastCheckEndTime()
```
vito, line 123: querying the last check start time/end time/succeeded seems to be making this logic pretty noisy and hard to follow - wdyt of putting it one call above all the branching, since we can fetch all this in one query pretty cheaply, and we need at least some of it in all but one of these branches?

```go
lastCheck, err := scope.LastCheck()
// ...

lastCheck.StartTime
lastCheck.EndTime
lastCheck.Succeeded
```

atc/engine/check_delegate.go
```diff
@@ -72,12 +72,19 @@ func (d *checkDelegate) FindOrCreateScope(config db.ResourceConfig) (db.Resource
 	return scope, nil
 }
 
+// WaitToRun decides if a check should really run or just reuse a previous result, and acquires
+// a check log accordingly. There are three types of checks, each reflects to a different behavior:
```
vito, line 76:
```suggestion
// a check lock accordingly. There are three types of checks, each reflects to a different behavior:
```

atc/engine/check_delegate.go
```diff
@@ -111,25 +118,56 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
 		}
 	}
 
-	end, err := scope.LastCheckEndTime()
-	if err != nil {
-		if releaseErr := lock.Release(); releaseErr != nil {
-			logger.Error("failed-to-release-lock", releaseErr)
+	shouldRun := false
+	if !d.plan.IsPeriodic() {
```
vito, line 122: i know i suggested this name, but i just noticed it might be a little confusing - `IsPeriodic` and `IsManuallyTriggered` sound mutually exclusive, but they're not. i can't really think of a better name though. if you have any ideas by all means - won't block the PR on this either way, since someone can always rename it later.

atc/engine/check_delegate.go
```diff
@@ -111,25 +118,56 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
...
+			}
+			return nil, false, fmt.Errorf("get last check end time: %w", err)
 		}
 
-		return nil, false, fmt.Errorf("get last check end time: %w", err)
-	}
+		if !lastCheckSucceeded || lastCheckEndTime.Before(d.build.StartTime()) {
+			shouldRun = true
+		} else {
+			_, found, err := scope.LatestVersion()
```
vito, line 134: so: if the last check succeeded *after* this build started *and* it didn't produce a version, run another check.

this seems like a pretty specific scenario to ensure we run an extra check (vs. just assuming there's still no version) - did something break without this?

vito (changes requested):
thanks for the adjustments!

i'm OK with adding the extra optimization, but now that the logic is getting more complicated I think we should take special care to make it clear to someone reading it for the first time. the optimizations make sense, but this logic also went from 6 lines to ~48, so i think it's worth trying to bring that down a bit if possible. i left a suggestion for that!

atc/engine/check_delegate.go
evanchaoli: Good idea.

atc/engine/check_delegate.go
evanchaoli: Yeah, I feel the same. Sorry for blindly accepting your renaming suggestion. I just renamed the method to `IsStepEmbedded()`.

atc/engine/check_delegate.go
evanchaoli: Nothing broken without this. Maybe I over worried. I think this should only protect the case of manually cleaning up `resource_config_versions` table. I don't know if any users do so. For us, we only cleanup versions of `time` resources. I just removed the last version check.

atc/engine/check_delegate.go
vito: it's possible to be `IsPeriodic && IsManuallyTriggered` which is the case this PR addresses (a webhook-triggered check for a 'periodic' check plan). so for someone that doesn't know what those actually mean it might seem a little confusing. no big deal though.

vito (approved):
nice! that last refactor definitely made it easier to read. lgtm now - merging, thanks again!

atc/engine/check_delegate.go
```diff
@@ -111,25 +118,32 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
...
-	if d.build.IsManuallyTriggered() {
-		// ignore interval for manually triggered builds
-		shouldRun = true
-	} else if !d.clock.Now().Before(runAt) {
-		// run if we're past the last check end time
-		shouldRun = true
+	if !d.plan.IsPeriodic() {
+		if !lastCheck.Succeeded || lastCheck.EndTime.Before(d.build.StartTime()) {
+			shouldRun = true
+		}
```
aoldershaw, line 133: Hey @evanchaoli, would you mind explaining what this is for? I'm reading it as: any non-periodic checks (is this just checks for `task.config.image_resource`?) shouldn't run if the prior check succeeded and this build started after it finished. But I'm not totally understanding what this logic is for. Is this to prevent multiple checks for the same `image_resource` from running in the same build?

evanchaoli: @aoldershaw  Yes, your understanding is correct. When a job has multiple tasks that use same `image_resource`, before this change, it runs a `check` for each task, which is unnecessary. This logic prevents duplicate `check`s for same `image_resource` for the same build. Do you see any problem with this logic?

aoldershaw: Nope no problem - just wanted to make sure my understanding was correct! Thanks!

PR #6867, opened 2021-04-12
Allow override of container limits in task config

BooleanCat:
### Feature

[Allow users of reusable concourse task files to set container limits in pipeline definitions](https://github.com/concourse/concourse/discussions/6646). Since it is not possible to set container limits for tasks using the `file` syntax, it means that reusable tasks effectively cannot have container limits.

Allow pipeline authors to set container limits on `file` tasks like so:

```yaml
jobs:
- name: build
  plan:
  - get: concourse-tasks
  - task: build-image
    file: concourse-tasks/build-oci-image-unprivileged/task.yml
    input_mapping: {dockerfile-repo: concourse-tasks}
    container_limits:
      cpu: 2
      memory: 519430400
    params:
      CONTEXT: dockerfile-repo/build-oci-image-unprivileged/fixtures/example-dockerfiles/
      IMAGE_NAME: my-image
      TAG: "1.0"
```

## Release Note

Pipeline authors can now set `container_limits` for reusable tasks in pipelines. Any limits set in the pipeline will override the limits set within the reusable task file.

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/builds/planner.go
+++ b/atc/builds/planner.go
@@ -52,6 +52,7 @@ func (visitor *planVisitor) VisitTask(step *atc.TaskStep) error {
 		Name:              step.Name,
 		Privileged:        step.Privileged,
 		Config:            step.Config,
+		Limits:            step.Limits,
 		ConfigPath:        step.ConfigPath,
 		Vars:              step.Vars,
 		Tags:              step.Tags,
--- a/atc/builds/planner_test.go
+++ b/atc/builds/planner_test.go
@@ -275,6 +275,60 @@ var factoryTests = []PlannerTest{
 		}`,
 	},
 	{
+		Title: "task step with top level container limits",
+
+		Config: &atc.TaskStep{
+			Name:       "some-task",
+			Privileged: true,
+			Config: &atc.TaskConfig{
+				Platform: "linux",
+				Run:      atc.TaskRunConfig{Path: "hello"},
+			},
+			Limits: &atc.ContainerLimits{
+				CPU:    newCPULimit(456),
+				Memory: newMemoryLimit(2048),
+			},
+			ConfigPath:        "some-task-file",
+			Vars:              atc.Params{"some": "vars"},
+			Params:            atc.TaskEnv{"SOME": "PARAMS"},
+			Tags:              atc.Tags{"tag-1", "tag-2"},
+			InputMapping:      map[string]string{"generic": "specific"},
+			OutputMapping:     map[string]string{"specific": "generic"},
+			ImageArtifactName: "some-image",
+			Timeout:           "1h",
+		},
+
+		PlanJSON: `{
+			"id": "(unique)",
+			"task": {
+				"name": "some-task",
+				"privileged": true,
+				"config": {
+					"platform": "linux",
+					"run": {"path": "hello"}
+				},
+				"config_path": "some-task-file",
+				"vars": {"some": "vars"},
+				"container_limits": {"cpu": 456, "memory": 2048},
+				"params": {"SOME": "PARAMS"},
+				"tags": ["tag-1", "tag-2"],
+				"input_mapping": {"generic": "specific"},
+				"output_mapping": {"specific": "generic"},
+				"image": "some-image",
+				"timeout": "1h",
+				"resource_types": [
+					{
+						"name": "some-resource-type",
+						"type": "some-base-resource-type",
+						"source": {"some": "type-source"},
+						"defaults": {"default-key":"default-value"},
+						"version": {"some": "type-version"}
+					}
+				]
+			}
+		}`,
+	},
+	{
 		Title: "set_pipeline step",
 
 		Config: &atc.SetPipelineStep{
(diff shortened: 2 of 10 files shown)


aoldershaw:
FWIW it is possible for reusable task files to set container limits using `vars`, but there are some downsides to this approach:
* Cannot set a default limit *and* allow the limits to be configurable
* Requires task authors to add var interpolation to the task file
* Less discoverable

vito (approved):
The code looks good and the change makes sense to me!

I considered asking for an RFC for this but I think it [affects users in a clear and obvious way](https://github.com/concourse/rfcs#when-the-rfc-process-is-necessary), so it doesn't seem necessary. I think it still makes sense to reach agreement within the core team for approving this though, since it technically increases the core pipeline API surface area.

I'll request a few more reviews before merging (target approvals = 4, since we have [5 members](https://github.com/concourse/governance/blob/f9213c7474372023ab48602e72753c7d625b61a5/teams/core.yml#L19-L23)).

aoldershaw:
I think it makes sense to allow overriding the `container_limits` - I guess the question to me is whether it makes sense to special case `container_limits` in the top-level, or if we should provide a more general way to override any part of the task config (e.g. allowing you to specify `task.config` and `task.file`, where `task.config` doesn't need to be a full config, but takes precedence over what's set in the `file`)

I think that's not necessarily a good idea, though, as most parts of the task config shouldn't be overridable (e.g. `inputs` and `outputs` should be fixed), so it probably makes sense to bring the configurable bits up a level (as we already have for `params`). Just figured I'd bring it up in case anyone else has ideas around this

vito:
@aoldershaw Hmm overriding inputs might not be the strangest thing, given reusable tasks like [`oci-build-task`](https://github.com/vito/oci-build-task) which don't have a specific set of inputs. That particular example will become a moot point with the advent of prototypes though.

aoldershaw:
@vito in that case, do you think it makes sense to make this config override more general and no longer complain about specifying `file:` and `config:`?

chenbh (approved):
Seems like a reasonable change to me

@aoldershaw I feel like the separation between the task file and the task step is the scheduling concerns. The task config holds the information on how run the workload and the task step only has information about where to run. I would say that the only other field in the task file that could benefit from being overridden at the step level is the `rootfs_uri` field

aoldershaw:
> The task config holds the information on how run the workload and the task step only has information about where to run

Hm I guess I don't really see it this way - to me, the config is very much part of the task step, and `task.file` mainly feels like a compromise for making pipelines more maintainable and for reusing tasks. Prototypes will solve both of these concerns, though, at the cost of being more heavier to implement.

Letting `task.config` override fields set in `task.file` makes it easier to write reusable tasks, and could be a lighter-weight alternative to writing prototypes - e.g. the `oci-build-task` usage could be:

```yaml
- get: oci-build-task # the repo, not the image
- task: build
  file: oci-build-task/task.yml
  config:
    inputs: [{name: some-input, path: .}]
```

However, given that we want to start encouraging people to write prototypes instead, this might not be the approach we want to take.

Don't want to be a blocker on this, just wanted to open up the conversation - if @clarafu also thinks the top-level `container_limits` approach is the right call, I'm happy to approve this as well

clarafu:
Hmm I think I like the idea of allowing `config` be an override if the `file` is also set. It will help remove this problem we are running into (for `container_limits`) for any new fields being added to task configs in the future. My only concerns about it now is that with prototypes (like @vito mentioned) this won't really be relevant anymore and if it's a lot more work to implement/maintain it then it might not be worth doing. Also, the fact that our users have an existing preconception of how `file` and `config` cannot be used together for a task and if it is worth the effort of changing their knowledge of it when we know we want to have them use prototypes in the future. The names itself (`file` and `config`) also don't make it obvious that `config` will override anything in `file`.

That being said, I do still like the idea of it so I think if adding `config` as an override is an easy/lightweight thing to implement I would be all for it 👍

vito:
> @vito in that case, do you think it makes sense to make this config override more general and no longer complain about specifying `file:` and `config:`?

Pretty much what @clarafu said - I think it theoretically makes sense but I'm not sure it's worth the churn at the moment. I don't think we would necessarily want merging of `config:` and `file:` if we were in a world where prototypes existed already. I think any changes we make to the pipeline config should nudge the project towards how we want users to use it, and I think we would want folks using prototypes in the future instead.

clarafu (approved):
I'm going to approve this because I agree with @vito that I like the concept of `config` and `file` overwriting but Im not sure if it's worth spending time ourselves or asking the contributor to do. And I don't see it as much as a problem because prototypes will remove this concern in the future anyways.

aoldershaw (approved):
Wasn't aware Concourse allowed for both `file:` and `config:` back in the day. I think the `run:` step point makes a lot of sense, too!

Issue #6828, opened 2021-04-06
in_parallel set_pipeline 'no diff found' log output display behaviour


Krenair:
## Summary

Observed unexpected numbers of 'no diff found' output lines between multiple in_parallel set_pipeline steps with the same names (but different teams). Should've seen exactly 1 but sometimes saw 0 and sometimes saw 2. Suspect something in Concourse's web UI logic is associating output with the wrong step.

## Steps to reproduce

This is not reliable - after taking the below screenshot I refreshed the page and it looked fine, but...
`in_parallel` step like this:

```
$ fly -t cd-main get-pipeline -p deploy-info-pipelines | yq '.jobs[0].plan[1]'
{
  "in_parallel": {
    "steps": [
      {
        "file": "path/to/pipelines/info.yml",
        "set_pipeline": "info",
        "team": "team-a",
        "vars": {
          "deployment": "cd",
          "team": "team-a"
        }
      },
      {
        "file": "path/to/pipelines/info.yml",
        "set_pipeline": "info",
        "team": "team-b",
        "vars": {
          "deployment": "cd",
          "team": "team-b"
        }
      }
    ]
  }
}
```

And so on. We have several of these set_pipeline steps in an in_parallel.

I think you may have to expand the set_pipeline sections and have it stream the results to see this issue.

## Actual results

Before refresh:
![image](https://user-images.githubusercontent.com/826662/113742888-fb7a0380-96fa-11eb-8a07-49acc337ae51.png)

## Expected results

Exactly one 'no diff found' per step.
After refresh:
![image](https://user-images.githubusercontent.com/826662/113743997-2fa1f400-96fc-11eb-8417-58ab28f0d24d.png)

## Additional context

`no diff found` text was changed to `no changes to apply` in #6164

## Triaging info

* Concourse version: 6.7.4
* Browser (if applicable): Firefox
* Did this used to work? Don't think I've ever spotted this before, but under the circumstances it's hard to say this was never a problem in the past.

aoldershaw:
Thanks for the report. I was able to reproduce with a similar setup, seeing one of the `set_pipeline` steps not output `no changes to apply` (I guess the output changed between 6.7.4 and 7.1) - though I only got it to "work" once. When I inspected the response from the `/api/v1/builds/:build_id/events` endpoint, I noticed that the API sent the same log event twice in a row rather than sending the `no changes to apply` log event. Upon refresh, this fixed itself, and the correct event was sent.

[...]

PR #6873, opened 2021-04-13
Use cursor-based pagination for build events

aoldershaw:

...rather than LIMIT+OFFSET, which doesn't scale well with many build
events (see https://ivopereira.net/efficient-pagination-dont-use-offset-limit)

for instance, build 1139906 on CI has 600000 build events - fetching a
single batch with offset 500000 takes over 1s using LIMIT+OFFSET, while
only takes 10ms using the cursor-based approach (event_id > $2)

also, rather than incrementing the cursor for each row we scan, it feels
safer to set the cursor to the last event id - I'm not positive it'll
help, but I'm hoping it might prevent issues like #6828 - without a good
repro, though, it's hard to know

## Release Note

* Optimizes fetching build logs from the DB for builds with massive logs

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/db/build_event_source.go
+++ b/atc/db/build_event_source.go
@@ -3,6 +3,7 @@ package db
 import (
 	"encoding/json"
 	"errors"
+	"strconv"
 	"sync"
 
 	"github.com/concourse/concourse/atc"
@@ -81,10 +82,13 @@ func (source *buildEventSource) Close() error {
 	return source.notifier.Close()
 }
 
-func (source *buildEventSource) collectEvents(cursor uint) {
+func (source *buildEventSource) collectEvents(from uint) {
 	defer source.wg.Done()
 
-	var batchSize = cap(source.events)
+	batchSize := cap(source.events)
+	// cursor points to the last emitted event, so subtract 1
+	// (the first event is fetched using cursor == -1)
+	cursor := int(from) - 1
 
 	for {
 		select {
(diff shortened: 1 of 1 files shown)


chenbh (approved):
lgtm

Did we ever come to a conclusion on if we want to build sql via raw sql statements (`tx.Query('SELECT...')`) or via functions (`psql.Select().From().Where()`)?

aoldershaw:
@chenbh not sure in general - I think `sq` makes some queries nicer, but sometimes feels constraining (in which case raw SQL is better IMO). Don't think the queries here fall into that camp, though, so converted them to the `sq` form in https://github.com/concourse/concourse/pull/6873/commits/be11381ef36116d9f4289a3551774077514a75df

atc/db/build_event_source.go
```diff
@@ -108,25 +109,28 @@ func (source *buildEventSource) collectEvents(from uint) {
...
+		rows, err := psql.Select("event_id", "type", "version", "payload").
+			From(source.table).
+			Where(sq.Or{
+				sq.Eq{"build_id": source.buildID},
+				sq.Eq{"build_id_old": source.buildID},
+			}).
+			Where(sq.Gt{"event_id": cursor}).
+			Limit(uint64(batchSize)).
+			RunWith(tx).
+			Query()
```
chenbh: Is this missing a `.OrderedBy(....)`?

aoldershaw: yikes, good catch. fixed

chenbh (approved):
lgtm

PR #6893, opened 2021-04-15
Start up concourse using a config file

### LINKED ISSUE #6434: Using a config file to startup concourse
This is just a WIP dump of information about this. Also a place for me to track the progress!

# Motivation

Currently the only way to configure the concourse binary is through flags or environment variables. This makes it hard to keep a consistent state of values for each of the fields and to properly maintain a version history of each
deployment state. It also feels slightly out of touch with the other ways of using Concourse, where everything is built through yaml files as with pipeline configs or task configs.

- complex data structures are hard to write in scripts
- help text is hard to read because we have so many long flag names
- rather than maintaining all the bosh/helm chart properties, we can have it contain a single config file property which the chart/bosh release parses and then passes the values to the concourse binary during startup

Having a simple config file will also help with clarity of exactly what is being set in a deployment without needing to read through every flag/env var that you need to pass in or set when you are starting up concourse. 

Another motivation for this would be to have less work in transitioning values from a bosh or kubernetes Concourse manifest into the flags/env that the binaries use. Currently, the way that the bosh and helm chart propogates the
values is through setting each as individual environment variables and then starting up the concourse binary. If we were to have a single config file it can instead be passed straight through from bosh/helm chart to the concourse binary, eliminating the need to convert the bosh/helm chart YAML files to environment variables. This will also remove the extra burden of contributers needing to add any new configuration fields to both the bosh and helm chart everytime they are added to Concourse.

# Proposal

I would like to propose that we use a single configuration file in order to start up the Concourse binary. This means that there will be only one configuration file in YAML format for both the web and worker nodes. I imagine starting Concourse web node would look something like this:

```sh
concourse web --config=/tmp/concourse.yml
```

with a config file:

```yaml
web:
  tsa_host_key: /tmp/host-key.pem
  
  external_url: https://some-url.com

  auth:
    local_users:
    - test:password

    main_team:
    - test

  runtime:
    container_placement_strategy: fewest-build-containers
    max_active_tasks_per_worker: 1

  enable_global_resources: true
```

rather than the current command:

```sh
concourse web \
  --tsa-host-key=/tmp/host-key.pem \
  --external-url=https://some-url.com \
  --add-local-user=test:password \
  --main-team-local-user=test \
  --container-placement-strategy=fewest-build-containers \
  --enable-global-resources=true \
  --max-active-tasks-per-worker=1
```

I was hoping to continue to support flags for backwards compatibility sake, but hopefully they will become deprecated and be removed in the future. This means that in the future we will only support configuring with a config file and env vars. Having to support all three separate ways of configuring
Concourse seems like an overkill, especially when the advantages of using one method outweigh the others. This is still an unanswered question, as to whether or not we want to continue to support three methods of configuration or should
we only support configuring with a config file and env vars but deprecate flags support. 

# Implementation Details

## Switching CLI libraries

Alongside adding support for Concourse to be able to be configured with a config file, I was thinking that this might be the time that we can start using a new library for the command line parser. As Concourse gets deeper into building a strong open source community, the importance of using tools that other contributors are more familiar with might be a better idea for the project. Concourse currently uses the [Go-flags](https://github.com/jessevdk/go-flags) library in order to parse the
command line flags and a package named [twentythousandtonnesofcrudeoil](https://github.com/vito/twentythousandtonnesofcrudeoil) to parse the envrionrment variables as an option parser of `go-flags`.

After looking at a few CLI libraries, I chose to go with [Cobra](https://github.com/spf13/cobra) as it is really well known and used by multiple large open-source projects like kubernetes, hugo and github-cli. It supports parsing flags, environment variables and a config file. An example of what the concourse command file might look like:

```go
var concourseCmd = &cobra.Command{
  Use:   "concourse",
  Short: "Thing doer",
  Long:  `Concourse is a thing doer of anything!`,
}

func init() 
	concourseCmd.AddCommand(VersionCommand) // The version flag will be a command, unlike before where it was a flag
	concourseCmd.AddCommand(WebCommand)
  concourseCmd.AddCommand(workercmd.WorkerCommand)
  ...
}
```

Each existing command will be added to the root command `concourseCmd`. This should include commands like `web`, `worker`, `migrate`, etc. Within the `web` and `worker` commands, it will parse out the config file, flags and env vars to
start up the component. The precedence for each source of configuration settings will be flags < env < config. Values set from a config file has the highest precedence because we want to deprecate the usage of flags and env in the future. The following code snippet shows how/where the three sources of configuration can come together:

```go
func InitializeWeb(cmd *cobra.Command, args []string) error {
	// All the flag values are already parsed into the web struct
	web := webFlags

  // Fetch out env values. Will override any existing values in the struct
  err := envstruct.FetchEnv(web)
  if err != nil {
    return fmt.Errorf("fetch env: %w", err)
  }

	// Fetch out the values set from the config file and overwrite the flag/env
	// values
	if configFile != "" {
		file, err := os.Open(configFile)
		if err != nil {
			return fmt.Errorf("open config file: %w", err)
		}

		decoder := yaml.NewDecoder(file)
		err = decoder.Decode(&web)
		if err != nil {
			return fmt.Errorf("decode config: %w", err)
		}
	}

  ... // Validation of values in struct. See below code snippet.
}

```

## A single point for configuration value validation

One thing that I have not mentioned yet is how we can now use a proper validator library to validate the values passed in by the user. Previously, flag validation was implemented everywhere. There was validation happening when the value was parsed out of the flags and into the struct, and also validation within the code for starting the web command itself.

I was hoping to use the [validator](https://github.com/go-playground/validator) library to validate fields within the config file parsing, this will separate the validation portion of the fields into a proper validator. In the code snippet below, it shows that after we parse out the values from the flags/env/config we can pass that to the validator that will take care of all the configuration value validation checks.

```go
func InitializeWeb(cmd *cobra.Command, args []string) error {
  ... // Parsing of values into struct. See above code snippet.

	// Validate the values passed in by the user
	webValidator := atccmd.NewValidator()
	err := webValidator.Struct(web)
	if err != nil {
		validationErrors := err.(validator.ValidationErrors)

		var errOuts []string
		for _, err := range validationErrors {
			errOuts = append(errOuts, err)
		}

		return fmt.Errorf(`print out errors`)
	}

	err = web.Execute(cmd, args)
	if err != nil {
		return fmt.Errorf("execute web: %w", err)
	}

	return nil
}
```

Some important TODOs:

- [ ] use clarafu/envstruct to parse envrionment vars into struct
- [ ] document all fields within concourse docs (no longer have the help text within the concourse command to show all the field descriptions)
- [ ] have a "default" config that shows all the defaults. one thing that gets lost with config files is ability to see defaults

## What will the migration paths be?

Maybe we can provide a tool which you can run against your concourse deployment that will create a config file that represents the current state of your deployment. Is it possible? IDK

## Some implications I'm thinking about

- changing workflows, still supporting flags/env but have people switch to using config files instead
- some values are changing within the transition from flags/env -> config file. Maybe we can have a tool that will convert for you?
- document all fields within concourse docs (no longer have the help text within the concourse command to show all the field descriptions)

clarafu:

closes #6434.

depends on concourse/flag#6 & concourse/baggageclaim#45

### Important notes to reviewer

First of all, I'd like to apologize to the reviewer. This is an incredibly large PR and also involves several other PRs to be merged at the same time to get this fully working. My very deepest apologies for putting you down this torture. 🙇  I am in the process of creating an RFC for this so I will update this PR once I get that finished. I just wanted to push out this code as soon as I could in order to get some initial feedback on the general direction.

I would really much like to say that most of the code should be pretty straight forward and pretty repetitive so I would say that you don't have to go through all the code, just skimming would be fine. I will be performing some super extensive manual testing of all the converted fields. I think my major concerns would be if these major changes are something that we want to use and support going forward? Or is it a step backwards in terms of how well it feels in our codebase? I am 100% open to just completely scrapping this or redoing a bunch of it if we feel like it does not work.

**This is still a WIP!**

### Summary of what this PR does

I would recommend reading through my issue description because it basically outlines every major change that is made this PR. But I can summarize it up for you here:

The biggest user facing change is that we will be supporting starting up Concourse using a config file! Prior to this, the only way to pass in configuration into the concourse command is through flags and environment variables. We will still be supporting passing in configuration values through flags and environment variables but only for backwards compatibility sake and will be fully deprecated in the future. 

Along with this big change is that fact that we no longer use `go-flags` to fetch flag values. And no longer using `go-flags` means we don't use `vito/twentythousandtonsofcrudeoil` because that was the library that we used to fetch environment variables from flags set with `go-flags`. Instead of those libraries, we will be using `cobra` to set up each concourse command and to support the deprecated flags. I chose to go with [`cobra`](https://github.com/spf13/cobra) for the CLI library because it is really widely used within the open source community, but I do have my reservations about it (esp with the flag setup). There are a few files that are named something like `deprecated_flags.go`, and these are all files that contain the setup necessary to allow cobra to accept flags. These files can be deleted after we fully stop supporting flags and also any new configuration fields added will not need to be added as a flag, meaning that those files do not need to be touched after this PR is merged.

Another change is that in order to fetch values from environment variables and parse them into the configuration fields, I use a library called [`envstruct`](https://github.com/clarafu/envstruct) that I created. Basically it just goes through each field in the struct and builds up an environment variable name using the tags within the fields. Then it uses that built up env name to try to fetch a value from the environment variables.

Lastly, I also added a [validation](https://github.com/go-playground/validator) library so that we can consolidate most of the validation of configuration field values to a single layer. 

### Configuration file structure

One thing that I want a lot of feedback on is the structure of the configuration YAML file. This is because once we put this structure in place, it will be annoying to have to restructure it in later releases. So I want to try to find the most sane structure the first time around when we release this. For example, I created a config file for the local docker-compose:

```yaml
---
web:
  auth:
    main_team:
      local_user:
      - test
    session_signing_key: /concourse-keys/session_signing_key
    add_local_user:
      test: test
      guest: guest
  database:
    postgres:
      host: db
      user: dev
      password: dev
      database: concourse
  log_level: debug
  external_url: http://localhost:8080
  web_server:
    cluster_name: dev
  feature_flags:
    enable_pipeline_instances: true
    enable_across_step: true

worker_gateway:
  authorized_keys: /concourse-keys/authorized_worker_keys
  host_key: /concourse-keys/tsa_host_key

worker:
  runtime: containerd
  containerd:
    network:
      dns-proxy:
        enable: true
  tsa:
    host:
    - web:2222
    public_key: /concourse-keys/tsa_host_key.pub
    worker_private_key: /concourse-keys/worker_key
  log_level: debug
  bind_ip: 0.0.0.0
  baggageclaim:
    bind_ip: 0.0.0.0
    driver: overlay
```

This is an example of the basic structure of the configuration file. I placed both worker and web configuration in the same file and I think it would be nice to have that be the norm (because worker and web are different commands in the concourse CLI). I also renamed the `tsa` to `worker_gateway`. 

TODO: I'll try to write up a config file that contains all the configuration fields that can possibly be set in order to see if everything is structured the way we want it to be.

### Migration path from flags/env -> config file

Kind of an open question right now but there are a few ways that I can think of going forward. I'm going to create an RFC where it will outline those options and what I'm thinking. I think my number 1 candidate right now is to have a concourse command that you pass it the same flags/env as you would when you start up concourse and it will spit out a config file.

### No more help text for configuration descriptions

One thing that we need to do before we can get this merged is to update the docs to show all the descriptions for each configuration. Now that we no longer rely on flags, we won't have the help text to give us each configuration description. Rather we want the users to be able to see the descriptions for each configuration within the concourse docs. This will make it a lot more readable and easy to access and reference. One downside I can see is having to submit a PR to both the concourse and docs repo when you want to add a new configuration field. Something we were talking about is putting the docs into the concourse repo, which would help remove that issue.

### No need to update helm chart and bosh release for new configuration

Another great thing that using a config file provides for us is that we no longer need to update the helm chart and bosh release for every new configuration field added to concourse. I haven't started on the work for converting those to use the config file, so that is a big TODO. 

## Notes to reviewer:

A few open questions that I have about this approach:

- [ ] Do we want to deprecate supporting env vars too? I had a discussion about it https://github.com/concourse/concourse/discussions/5891 and we ended up deciding to just deprecate both env and flags but I want to keep this question be open in case anybody has any more opinions on the matter.
- [ ] How do we "enable" integrations with concourse? (ex. metrics emitters, creds managers, tracing providers, auth connectors). Right now we have a field where you write in the string of which one you want to use (except for auth providers where you use an enabled bool because you can have multiple). Is that best? Or should we do enabled bools for all? Another question is should these types be "nested" together in the config file?
  ```
  worker:
    runtime: containerd
    containerd:
      foo: bar
    garden:
      foo: bar

  vs

  worker:
    runtime:
      type: containerd
      containerd:
        foo: bar
      garden:
        foo: bar
  ```

In general, everything should be backwards compatible and there should be no behavioural changes. Here are the snags in this that I changed during this PR:

- Now that you configure auth providers that you want to use with setting `enabled: true` on the auth provider, it will error if the validation does not pass for the values that are set in that auth provider. Previously, we did not return with an error when an auth provider did not successfully `.Validate()`. This is because we wanted to find all the auth providers that the user configured and pass those to dex. But now that we know which ones are configured using the enabled boolean, we can use the `.Validate()` method to check if it is configured properly. https://github.com/concourse/concourse/blob/904ff186696903cef9df51cf1f47add5c5c4919a/skymarshal/dexserver/dexserver.go#L70-L76

There are still a lot of TODOs within the issue but I think most of the things to finish don't really affect the current code implementations, they're mostly add ons and manual testing. You can keep track of the TODOs in the issue.

If you want to docker compose this locally, you will need to grab the `config-file` branch from the `baggageclaim` and `flag` libraries and put them into this concourse repo. You can look at `go.mod` to see the replacements for those libaries.

## Release Note

* 

## Contributor Checklist

- [ ] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).


aoldershaw:
Jeeeeeez, this looks like it was pretty painful to write, so appreciate all the work you put in!

One thing that strikes me, though - do we really care about supporting flags/env vars in addition to the config file?

I get that we want the upgrade to be as painless as possible, but there are ways to make the migration easier, and so long as we make it very clear that you have to do an additional step before upgrading and document it well, I suspect that wouldn't be too much trouble for operators.

To me, it feels like the easiest solution is to add a separate `concourse migrate-config` command that accepts flags/env vars and spits out the equivalent config file. We'd take a copy of all the [legacy config structs](https://github.com/concourse/concourse/blob/22f2b4b1c99e42562a106206fa8d7bbcfaf7643b/atc/atccmd/command.go#L110) without any of the `defaults` and somehow convert that to the new structure (this could be as simple as adding `yaml` struct tags in addition to the `long` tags, but if the structure of the config file differs from the existing namespaces, we'd probably need to do more work).

Did you consider this approach? Is there a reason why it wouldn't work/is suboptimal?

clarafu:
@aoldershaw Ya thats definitely a good idea! I think I'm leaning the most towards that approach but the only reason why I had not solidified it yet was because I wanted to create an RFC to have some discussions about a few different ways I could see this happening. So the idea you mentioned was one of them and the other is actually having some sort of ATC/Worker endpoint (or concourse command?) that returns back a config. Not sure exactly how it would be implemented but I imagined you would point it to an existing concourse and it would spit out a config. This would mean that you would need to have already upgraded first, but the benefits I can think of is that it can be used not only when you want to migrate, but also in general purpose use for if the operator would like concourse to return back a config that contains all the most updated values that the concourse cluster is currently running with. Only annoying thing about this is that you would need to have upgraded first. Maybe there can be two migration paths? One for if you want to create a config file and use the config file during the upgrade and the other is to wait until after you upgrade to switch to a config file? I'll bring all this up in the RFC too.

For the flags and env, I don't think we want to fully support them. I do think keeping them for backwards compatibility is pretty important though and nice for the user so that they won't be forced to use a config file when they want to upgrade to a later version. My thoughts are to make it as easy as possible for users to upgrade so that they will feel less resistant to it by keeping the barrier to upgrading as low as possible. But I am totally open to other opinions!

aoldershaw:
I guess another question is: why the move to cobra? I get that it's a more canonical choice than go-flags, but given that we're trying to deprecate the flag usage anyway, I feel like it'd be easier/safer to stick with the existing flag structs we already know work rather than risk a regression in the switch.

clarafu:
@aoldershaw I think I mostly chose it because I wanted to switch concourse to use a CLI library that is well known in the open source community and since we still want to support flags for certain things, like `concourse web --config` and other concourse commands like `concourse migrate --latest-db-version`, I thought it would be best to just switch everything over to using cobra. 

Hm I agree with your concern about the risk of regression from switching all the flags over to cobra. I think when I weighed whether I wanted to keep using `go-flags` for the concourse flags vs switching it all over, I thought that the benefits of keeping all the deprecated flags isolated in files that contributors never need to touch would be better than having the struct fields be polluted with both the new tags and the `go-flag` tags. And in order to try to minimize the risk of regression, I will be doing some super extensive manual testing to ensure each flag/env is the same as before. That being said, I am definitely open to having the deprecated flags stay as `go-flags`, I think that will greatly reduce the risk of regression which is something I am worried about too.

atc/atccmd/command.go
```diff
@@ -96,387 +99,252 @@ const algorithmLimitRows = 100
...
-	} `group:"Authentication"`
-
-	ConfigRBAC flag.File `long:"config-rbac" description:"Customize RBAC role-action mapping."`
+	CredentialManagement creds.CredentialManagementConfig `yaml:"credential_management,omitempty"`
+	CredentialManager    string                           `yaml:"credential_manager,omitempty" validate:"creds_manager"`
+	CredentialManagers   CredentialManagersConfig         `yaml:",inline"`
 
-	SystemClaimKey    string   `long:"system-claim-key" default:"aud" description:"The token claim key to use when matching system-claim-values"`
-	SystemClaimValues []string `long:"system-claim-value" default:"concourse-worker" description:"Configure which token requests should be considered 'system' requests."`
+	ComponentRunnerInterval time.Duration `yaml:"component_runner_interval,omitempty"`
```
vito, line 130: Will this successfully parse a string duration value like `10s`?

atc/atccmd/command.go
```diff
@@ -96,387 +99,252 @@ const algorithmLimitRows = 100
...
-		nil,
-	)
+	BaseResourceTypeDefaults flag.File `yaml:"base_resource_type_defaults,omitempty"`
 
-	version, err := helper.CurrentVersion()
-	if err != nil {
-		return err
-	}
+	// NOT USED (HIDDEN)
+	TelemetryOptIn bool `yaml:"telemetry_opt_in,omitempty"`
```
vito, line 157: We can probably just get rid of this. I don't think anyone in the universe ever specified it. Not a big deal tho.

atc/atccmd/command.go
```diff
@@ -96,387 +99,252 @@ const algorithmLimitRows = 100
...
 
-func (cmd *Migration) currentDBVersion() error {
-	helper := migration.NewOpenHelper(
-		defaultDriverName,
-		cmd.Postgres.ConnectionString(),
-		cmd.lockFactory,
-		nil,
-		nil,
-	)
+	BaseResourceTypeDefaults flag.File `yaml:"base_resource_type_defaults,omitempty"`
```
vito, line 154: Rather than pointing this to a file, now that we have an actual config file I think it'd make sense to just have it inline.

atc/atccmd/command.go
```diff
@@ -96,387 +99,252 @@ const algorithmLimitRows = 100
...
+	BindPort    uint16    `yaml:"bind_port,omitempty"`
+	TLS         TLSConfig `yaml:"tls,omitempty"`
+	ExternalURL flag.URL  `yaml:"external_url,omitempty" validate:"url"`
 
-	JobSchedulingMaxInFlight uint64 `long:"job-scheduling-max-in-flight" default:"32" description:"Maximum number of jobs to be scheduling at the same time"`
+	Auth                      AuthConfig           `yaml:"auth,omitempty" ignore_env:"true"`
+	Server                    ServerConfig         `yaml:"web_server,omitempty"`
+	SystemClaim               SystemClaimConfig    `yaml:"system_claim,omitempty"`
+	LetsEncrypt               LetsEncryptConfig    `yaml:"lets_encrypt,omitempty"`
+	ConfigRBAC                flag.File            `yaml:"config_rbac,omitempty" validate:"rbac"`
```
vito, line 120: I think this can/should be inlined too.

atc/atccmd/config_types.go
```diff
@@ -0,0 +1,208 @@
...
+	BindPort uint16    `yaml:"bind_port,omitempty"`
+	Cert     flag.File `yaml:"cert,omitempty"`
+	Key      flag.File `yaml:"key,omitempty"`
+	CaCert   flag.File `yaml:"ca_cert,omitempty"`
+}
+
+// IMPORTANT! The env tags are only added for backwards compatibility sake. Any
+// new fields do NOT require the env tag
+type LetsEncryptConfig struct {
+	Enable  bool     `yaml:"enable,omitempty" env:"CONCOURSE_ENABLE_LETS_ENCRYPT,CONCOURSE_LETS_ENCRYPT_ENABLE"`
```
clarafu, line 33: If you are wondering why some struct values have `env` and some don't, its because I use my library called `envstruct` and it will build up an environment variable name based on the tags within the struct. For example, this `LetsEncryptConfig` is actually within a struct

```
type RunConfig struct {
  LetsEncrypt LetsEncryptConfig `yaml:"lets_encrypt"`
}
```

So the env that will be built up for the `Enable` field is `CONCOURSE_LETS_ENCRYPT_ENABLE` because it appends the prefix (which is set to `CONCOURSE`) to the `yaml` tag in the nested struct for `LetsEncrypt` (which is `lets_encrypt` but uppercased) and then to the value of the `yaml` tag for `Enable` (which is `enable` but uppercased). When we try to find an environment variable for this field, it will use that built up string `CONCOURSE_LETS_ENCRYPT_ENABLE`.

The problem is that fetching env is purely for backwards compatibility purposes and the env that we used before this config PR was `CONCOURSE_ENABLE_LETS_ENCRYPT` which is different than `CONCOURSE_LETS_ENCRYPT_ENABLE`. So setting a tag value for `env` will override the env variable that is fetched from `envstruct`.

atc/atccmd/config_types.go
```diff
@@ -0,0 +1,208 @@
...
+	"github.com/concourse/concourse/atc/worker"
+	"github.com/concourse/concourse/skymarshal/skycmd"
+	"github.com/concourse/flag"
+)
+
+type TLSConfig struct {
+	BindPort uint16    `yaml:"bind_port,omitempty"`
+	Cert     flag.File `yaml:"cert,omitempty"`
+	Key      flag.File `yaml:"key,omitempty"`
+	CaCert   flag.File `yaml:"ca_cert,omitempty"`
```
vito, line 27: Could make sense to support inlining these values rather than referencing files, but referencing files makes sense here too.

Maybe these could be called `*_file` and the inline form could be named `cert`, `ca_cert`, `key`?

atc/atccmd/command.go
```diff
@@ -96,387 +99,252 @@ const algorithmLimitRows = 100
...
-		}
-
-		if authGroup == nil && group.ShortDescription == "Authentication" {
-			authGroup = group
-		}
-
-		if metricsGroup != nil && credsGroup != nil && authGroup != nil && policyChecksGroup != nil {
-			break
-		}
+var CmdDefaults RunConfig = RunConfig{
```
vito, line 162: I like it - I think this is a lot easier to skim!

atc/atccmd/command.go
```diff
@@ -96,387 +99,252 @@ const algorithmLimitRows = 100
...
+		ClientID:      "concourse-web",
+	},
+
+	SystemClaim: SystemClaimConfig{
+		Key:    "aud",
+		Values: []string{"concourse-worker"},
+	},
+
+	LetsEncrypt: LetsEncryptConfig{
+		ACMEURL: ignoreErrParseURL("https://acme-v02.api.letsencrypt.org/directory"),
```
vito, line 206: tiny suggestion: a convention I've seen for this is to call the function `mustParseURL` and rather than ignoring it'll panic on any err. this convention comes from go stdlib: https://golang.org/pkg/regexp/#MustCompile

atc/atccmd/deprecated_flags.go
```diff
@@ -0,0 +1,506 @@
+package atccmd
+
+import (
+	"time"
+
+	"github.com/concourse/flag"
+	"github.com/spf13/cobra"
+)
+
+// XXX !! IMPORTANT !!
```
clarafu, line 10: There are a few added files throughout this PR that are called `deprecated_flags`. These files are supposed to be hidden away where any contributors wanting to add a new configuration field to any part of the config structs will not need to touch these files. This is because these files are only used to set up the flags and we are deprecating flags, they only exist now because we want to keep it backwards compatible.

This was a reason why I chose to overwrite all the flags, it is so that we can keep all the flag setup in separate files away from the main code. So any new changes will not need to touch these files at all and the only time we need to modify or look at this file is at the beginning where we are adding them within this PR and then also once in the future when we delete them because we no longer support flags.

That being said, these files are pretty uuuuugly with the number of flags we have.

cmd/concourse/validator/validator.go
```diff
@@ -0,0 +1,397 @@
...
+	"github.com/go-playground/validator/v10"
+	en_translations "github.com/go-playground/validator/v10/translations/en"
+	"gopkg.in/yaml.v2"
+)
+
+// NewValidator constructs a new validator that contains all the default and
+// custom validations that are necessary to run against the concourse config
+// fields. These include validations on the atc, tsa, worker and baggageclaim
+// fields.
+func NewValidator(trans ut.Translator) *validator.Validate {
```
clarafu, line 28: I added `go-playground/validator` to concourse because I wanted to try to consolidate as much validation as I could into one place. Previously, we just had the validation of configuration fields be spread out throughout the code. With this library, you would just add the `validate` tag to the struct field that you want to run a validation against, and you can use the built in validations that the library contains or you can add custom ones. 

I also used it to add custom error messages for the failure of the custom validations. I think that will be nice for showing a prettier error message to the user. I'd love any feedback on this though, not sure if this is better than before or not.

atc/integration/integration_suite_test.go
```diff
@@ -33,19 +32,12 @@ var (
 var _ = postgresrunner.GinkgoRunner(&postgresRunner)
 
 var _ = BeforeEach(func() {
-	cmd = &atccmd.RunCommand{}
-
-	// call parseArgs to populate flag defaults but ignore errors so that we can
```
vito, line 38: always nice seeing comments like this go away

atc/metric/emitter/lager.go
```diff
@@ -5,20 +5,24 @@ import (
...
-	metric.Metrics.RegisterEmitter(&LagerConfig{})
+	Enabled bool `yaml:"enabled,omitempty" env:"CONCOURSE_LAGER_ENABLE,CONCOURSE_EMIT_TO_LOGS"`
 }
 
+func (config *LagerConfig) ID() string          { return "lager" }
 func (config *LagerConfig) Description() string { return "Lager" }
-func (config *LagerConfig) IsConfigured() bool  { return config.Enabled }
+func (config *LagerConfig) Validate() error {
+	if !config.Enabled {
+		return errors.New("enabled needs to be true")
```
vito, line 21: does this even need to be here? seems like it could just always be valid, and we only use it if it's enabled?

cmd/concourse/quickstart.go
```diff
@@ -3,35 +3,133 @@ package main
...
+}
+
+func init() {
+	QuickstartCommand.Flags().StringVar(&quickStart.ConfigFile, "config", "", "path to the config file that will be used to quickstart the concourse cluster")
+
+	quickStart.WebConfig.RunConfig = atccmd.CmdDefaults
+	quickStart.WebConfig.TSAConfig = tsacmd.CmdDefaults
+	quickStart.WorkerCommand = workercmd.CmdDefaults
+
+	// IMPORTANT!: Can be removed when flags no longer supported
```
vito, line 46: nit: I've been using a `// XXX(branch-name/refactor-tag)` convention for comments like this, e.g. `// XXX(checks-refactor)`.

cmd/concourse/quickstart.go
```diff
@@ -3,35 +3,133 @@ package main
...
+	// IMPORTANT! This can be removed after we completely deprecate flags
+	fixupFlagDefaults(cmd, &quickStart.WebConfig)
+
+	// Fetch out env values
+	env := envstruct.Envstruct{
+		Prefix:        "CONCOURSE",
+		TagName:       "yaml",
+		OverrideName:  "env",
+		IgnoreTagName: "ignore_env",
+		StripValue:    true,
```
vito, line 69: What's this do? (I know I can look, just giving quick feedback from an API consumer's perspective to see if there's a clearer name we could use.)

cmd/concourse/quickstart.go
```diff
@@ -3,35 +3,133 @@ package main
...
+
+		Parser: envstruct.Parser{
+			Delimiter:   ",",
```
vito, line 72: What's this do? (same reasoning as above) What does it delimit? hm, maybe slices and maps?

worker/workercmd/worker.go
```diff
@@ -22,47 +23,92 @@ import (
 )
 
 type WorkerCommand struct {
-	Worker WorkerConfig
+	Worker WorkerConfig `yaml:",inline"`
 
-	TSA worker.TSAConfig `group:"TSA Configuration" namespace:"tsa"`
+	TSA worker.TSAConfig `yaml:"tsa,omitempty"`
```
vito, line 28:
```suggestion
	TSA worker.TSAConfig `yaml:"gateway,omitempty"`
```

Doesn't make sense to me to call the component's config `worker_gateway` but still refer to it as `tsa` here.

vito (commented):
skimmin'! leaving some quick back for now before I continue on

atc/atccmd/command.go
clarafu: I haven't tested it yet but it is just using `yaml.Unmarshal` or `yaml.Decode` (or pflags) to populate that so it should work? I'll confirm once I have tested it though.

atc/atccmd/command.go
clarafu: Maybe we can leave it to a different PR to get rid of it? I was hoping to keep this PR with as little behavioural changes as possible in the very slim chance that someone is actually still using it.

atc/atccmd/command.go
clarafu: Oh like have it be inline within the config file but there is a flag that still accepts the `base_resource_type_defaults` as a file?

atc/atccmd/command.go
clarafu: :D Really?

I was worried people might like seeing the default as a tag next to the field more

atc/integration/integration_suite_test.go
clarafu: yessir. even though my PR adds a ton more comments for deprecated lines of code. hehe

atc/metric/emitter/lager.go
clarafu: Oh so have it not fail during validation phase but push the checking of `enabled` to the initializing of the emitter?

cmd/concourse/quickstart.go
clarafu: Yup! Slices and maps

cmd/concourse/quickstart.go
aoldershaw: Not sure if there's a standard name for it, but I've heard to these be referred to as `Options` (e.g. https://pkg.go.dev/github.com/fatih/structtag#Tag).

So, maybe `StripOptions`?

clarafu:
Closing this for now because I don't have time to work on this anytime soon. This feature is still valid and hopefully can be submitted again in the future!

PR #6923, opened 2021-04-22
Scan unchecked resource-types

### LINKED ISSUE #6521: Changing resource type config results in mismatched version usage
## Summary

If I change a resource type's `source:` to point to a different image, steps will use the `version` associated to the resource type's existing `resource_config_id` until the `resource_config_id` gets updated. If the resource type is only ever used in put-only resources, it will never be updated.

## Steps to reproduce

1. `docker-compose up`
2. Configure a pipeline using `mockersf/concourse-slack-notifier` resource type **only in a `put` step**
3. Run a build that sends a notification
4. Reconfigure the resource type with a different resource type (e.g. `aoldershaw/concourse-slack-notifier`)
5. Run another build

## Expected results

The build should succeed, having used the latest version of `aoldershaw/concourse-slack-notifier`.

## Actual results

The build errors, trying to fetch `aoldershaw/concourse-slack-notifier` using a version associated to `mockersf/concourse-slack-notifier`.

image

## Additional context

Already confirmed that the version is associated to `mockersf/concourse-slack-notifier`.

```
$ docker pull mockersf/concourse-slack-notifier@sha256:d4ec74d5c6141685ed0245b8b0f05678e874c9b526a6a7ad91bb80199d765306
docker.io/mockersf/concourse-slack-notifier@sha256:d4ec74d5c6141685ed0245b8b0f05678e874c9b526a6a7ad91bb80199d765306: Pulling from mockersf/concourse-slack-notifier
2bafa357a9a5: Pull complete
754871f61ff7: Pull complete
428217b78094: Pull complete
bd217fb5bb19: Pull complete
aaae4c086bc2: Pull complete
f3b12d6136e0: Pull complete
Digest: sha256:d4ec74d5c6141685ed0245b8b0f05678e874c9b526a6a7ad91bb80199d765306
Status: Downloaded newer image for mockersf/concourse-slack-notifier@sha256:d4ec74d5c6141685ed0245b8b0f05678e874c9b526a6a7ad91bb80199d765306
docker.io/mockersf/concourse-slack-notifier@sha256:d4ec74d5c6141685ed0245b8b0f05678e874c9b526a6a7ad91bb80199d765306
```

This can be fixed by manually running `fly check-resource-type`. This issue more significantly affects put-only resources, because their resource types are not automatically checked, but it should also affect regular resources for the time interval until the resource type is automatically checked.

## Triaging info

* Concourse version: `370b34f7524696627c0047da39a71c2d114f9dd6` (`master`)
* Did this used to work? Pretty sure this was broken before too, but I haven't confirmed. #6281 just makes it more noticeable.

EstebanFS:

closes #6521

This covers the edge case where a put-only resource's parent-type would not be check because we were only looping over everything except put-only resources

## Notes to reviewer:
If you need extra help reproducing the issue, you can check Alex's comment https://github.com/concourse/concourse/issues/6521#issuecomment-813596246

## Release Note
* Fixed an edge case where a put-only resource's parent-type would not be checked

## Contributor Checklist
- [x] [Signed] all commits

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

cc @taylorsilva

### DIFF
--- a/atc/lidar/scanner.go
+++ b/atc/lidar/scanner.go
@@ -43,20 +43,15 @@ func (s *scanner) Run(ctx context.Context) error {
 		return err
 	}
 
-	waitGroup := new(sync.WaitGroup)
-	resourceTypesChecked := &sync.Map{}
-
-	s.scanResources(spanCtx, resources, resourceTypes, waitGroup, resourceTypesChecked)
-	waitGroup.Wait()
-	s.scanUncheckedResourceTypes(spanCtx, resourceTypes, waitGroup, resourceTypesChecked)
-
-	waitGroup.Wait()
+	s.scanResourceTypes(spanCtx, resourceTypes)
+	s.scanResources(spanCtx, resources, resourceTypes)
 
 	return nil
 }
 
-func (s *scanner) scanResources(ctx context.Context, resources []db.Resource, resourceTypes db.ResourceTypes, waitGroup *sync.WaitGroup, resourceTypesChecked *sync.Map) {
+func (s *scanner) scanResources(ctx context.Context, resources []db.Resource, resourceTypes db.ResourceTypes) {
 	logger := lagerctx.FromContext(ctx)
+	waitGroup := new(sync.WaitGroup)
 	for _, resource := range resources {
 		waitGroup.Add(1)
 
@@ -69,16 +64,17 @@ func (s *scanner) scanResources(ctx context.Context, resources []db.Resource, re
 			}()
 			defer waitGroup.Done()
 
-			s.check(ctx, resource, resourceTypes, resourceTypesChecked)
+			s.check(ctx, resource, resourceTypes)
 		}(resource, resourceTypes)
 	}
+	waitGroup.Wait()
 }
 
-func (s *scanner) scanUncheckedResourceTypes(ctx context.Context, resourceTypes db.ResourceTypes, waitGroup *sync.WaitGroup, resourceTypesChecked *sync.Map) {
+func (s *scanner) scanResourceTypes(ctx context.Context, resourceTypes db.ResourceTypes) {
 	logger := lagerctx.FromContext(ctx)
+	waitGroup := new(sync.WaitGroup)
 	for _, resourceType := range resourceTypes {
 		waitGroup.Add(1)
-
 		go func(resourceType db.ResourceType, resourceTypes db.ResourceTypes) {
 			defer func() {
 				err := util.DumpPanic(recover(), "scanning resource type %d", resourceType.ID())
@@ -87,15 +83,13 @@ func (s *scanner) scanUncheckedResourceTypes(ctx context.Context, resourceTypes 
 				}
 			}()
 			defer waitGroup.Done()
-			if _, exists := resourceTypesChecked.LoadOrStore(resourceType.ID(), true); !exists {
-				// only create a check for resource type if it has not been checked yet
-				s.check(ctx, resourceType, resourceTypes, resourceTypesChecked)
-			}
+			s.check(ctx, resourceType, resourceTypes)
 		}(resourceType, resourceTypes)
 	}
+	waitGroup.Wait()
 }
 
-func (s *scanner) check(ctx context.Context, checkable db.Checkable, resourceTypes db.ResourceTypes, resourceTypesChecked *sync.Map) {
+func (s *scanner) check(ctx context.Context, checkable db.Checkable, resourceTypes db.ResourceTypes) {
 	logger := lagerctx.FromContext(ctx)
 
 	spanCtx, span := tracing.StartSpan(ctx, "scanner.check", tracing.Attrs{
@@ -106,14 +100,6 @@ func (s *scanner) check(ctx context.Context, checkable db.Checkable, resourceTyp
 		"resource_config_scope_id": strconv.Itoa(checkable.ResourceConfigScopeID()),
 	})
 	defer span.End()
-
-	parentType, found := resourceTypes.Parent(checkable)
-	if found {
-		if _, exists := resourceTypesChecked.LoadOrStore(parentType.ID(), true); !exists {
-			// only create a check for resource type if it has not been checked yet
-			s.check(spanCtx, parentType, resourceTypes, resourceTypesChecked)
-		}
-	}
 
 	version := checkable.CurrentPinnedVersion()
 
--- a/atc/lidar/scanner.go
+++ b/atc/lidar/scanner.go
@@ -46,6 +46,17 @@ func (s *scanner) Run(ctx context.Context) error {
 	waitGroup := new(sync.WaitGroup)
 	resourceTypesChecked := &sync.Map{}
 
+	s.scanResources(spanCtx, resources, resourceTypes, waitGroup, resourceTypesChecked)
+	waitGroup.Wait()
+	s.scanUncheckedResourceTypes(spanCtx, resourceTypes, waitGroup, resourceTypesChecked)
+
+	waitGroup.Wait()
+
+	return nil
+}
+
+func (s *scanner) scanResources(ctx context.Context, resources []db.Resource, resourceTypes db.ResourceTypes, waitGroup *sync.WaitGroup, resourceTypesChecked *sync.Map) {
+	logger := lagerctx.FromContext(ctx)
 	for _, resource := range resources {
 		waitGroup.Add(1)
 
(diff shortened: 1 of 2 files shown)


atc/lidar/scanner.go
```diff
@@ -46,6 +46,17 @@ func (s *scanner) Run(ctx context.Context) error {
 	waitGroup := new(sync.WaitGroup)
 	resourceTypesChecked := &sync.Map{}
 
+	s.scanResources(spanCtx, resources, resourceTypes, waitGroup, resourceTypesChecked)
+	waitGroup.Wait()
```
aoldershaw, line 50: Is there a reason we want the two sets of "scans" to run sequentially? Or can we get rid of this `waitGroup.Wait()`?

atc/lidar/scanner.go
```diff
@@ -58,13 +69,30 @@ func (s *scanner) Run(ctx context.Context) error {
...
-	waitGroup.Wait()
+func (s *scanner) scanUncheckedResourceTypes(ctx context.Context, resourceTypes db.ResourceTypes, waitGroup *sync.WaitGroup, resourceTypesChecked *sync.Map) {
+	logger := lagerctx.FromContext(ctx)
+	for _, resourceType := range resourceTypes {
+		waitGroup.Add(1)
 
-	return nil
+		go func(resourceType db.ResourceType, resourceTypes db.ResourceTypes) {
+			defer func() {
+				err := util.DumpPanic(recover(), "scanning resource type %d", resourceType.ID())
```
aoldershaw, line 84: A bit confused about when this could panic. I guess it's just being defensive so as not to take down the ATC due to a programming mistake - but feels like that should get caught in tests, as it's just relatively simple control flow to create a build in the DB.

p.s. I recognize this panic recovery was present in the scanner beforehand - happy to keep it, just feels a bit odd to me

atc/lidar/scanner.go
```diff
@@ -58,13 +69,30 @@ func (s *scanner) Run(ctx context.Context) error {
...
+}
 
-	waitGroup.Wait()
+func (s *scanner) scanUncheckedResourceTypes(ctx context.Context, resourceTypes db.ResourceTypes, waitGroup *sync.WaitGroup, resourceTypesChecked *sync.Map) {
+	logger := lagerctx.FromContext(ctx)
+	for _, resourceType := range resourceTypes {
+		waitGroup.Add(1)
 
-	return nil
+		go func(resourceType db.ResourceType, resourceTypes db.ResourceTypes) {
```
aoldershaw, line 82: Minor thing, but we don't need to create a goroutine when the resource type is already checked. e.g. this could be:

```go
for _, resourceType := range resourceTypes {
	if _, exists := resourceTypesChecked.LoadOrStore(resourceType.ID(), true); exists {
		// check has already been created for resource type
		continue
	}
	waitGroup.Add(1)

	go func(resourceType db.ResourceType, resourceTypes db.ResourceTypes) {
		defer func() {
			err := util.DumpPanic(recover(), "scanning resource type %d", resourceType.ID())
			if err != nil {
				logger.Error("panic-in-scanner-run", err)
			}
		}()
		defer waitGroup.Done()

		s.check(ctx, resource, resourceTypes, resourceTypesChecked)
	}(resourceType, resourceTypes)
}
```

atc/lidar/scanner.go
```diff
@@ -46,6 +46,17 @@ func (s *scanner) Run(ctx context.Context) error {
 	waitGroup := new(sync.WaitGroup)
 	resourceTypesChecked := &sync.Map{}
 
+	s.scanResources(spanCtx, resources, resourceTypes, waitGroup, resourceTypesChecked)
+	waitGroup.Wait()
+	s.scanUncheckedResourceTypes(spanCtx, resourceTypes, waitGroup, resourceTypesChecked)
```
aoldershaw, line 51: Now that we check all resource types, does `s.check` still need to check the parent resource type? Or perhaps the order the checks are created in matters

taylorsilva: I think it still makes sense for `s.check` to check the parent resource-type first before running the resource's check. It feels out of order to run a check for a resource and then check if it's base resource-type has updated. I'm going to leave this as-is.

atc/lidar/scanner.go
taylorsilva: I say we keep it. Came from this PR: https://github.com/concourse/concourse/pull/5842

atc/lidar/scanner.go
taylorsilva: This gave me a different idea. Got rid of the `sync.Map` and put the `sync.WaitGroup` inside the `scan*` funcs.

aoldershaw: That's kind of what I had in mind as well, but wasn't totally convinced it would work - e.g. consider the following:

```yaml
resource_types:
- name: grandparent
  type: registry-image
- name: parent
  type: grandparent

resources:
- name: child
  type: parent
```

If we don't perform the recursive `s.check` for parent types, it's possible that we create a check for `parent` before its parent type `grandparent` - might that be an issue? (genuinely not sure, but it feels odd)

taylorsilva: Didn't even think about that use-case because I don't _think_ someone would use anything else besides `registry-image` as the `type` for `resource_types`. I think this would cause some unexpected behaviour depending on the order of the `resourceTypes` returned by the db. If `grandparent` appears before `parent` then everything is fine. If not, then for **one check interval**, an older version of `grandparent` would be used to fetch `parent`, instead of the most recently available version of `grandparent`. This would immediately be resolved on the next check for these resource-types though. Maybe this is fine? We _eventually_ get you to the expected result? wdyt? I'm a bit on the fence about this.

aoldershaw: I guess what's not clear to me is what `s.check` actually does. It doesn't actually run the check - it just creates a build in the DB (if it needs to) for the build tracker to pick up in order to actually run the check.

So, I *think* it shouldn't matter - I believe an older version of `grandparent` will always be used to fetch `parent` for a single check interval, even if we create the build for `grandparent` first

taylorsilva: Good point. I haven't tested this myself but I'm not even sure if there's a max-in-flight for all checks. I remember a PR about smoothing out the number of checks getting run; I forget where that is.
I'm fine with this PR as-is then. Up to you if you think we should still nudge this back to the recursive `s.check` or keep this simpler one.

aoldershaw: There's some rate limiting for running checks, but that's done within the check step itself (the check delegate, to be more precise):
https://github.com/concourse/concourse/blob/429eead503de082bd1eac1211c7afd0ed20418dc/atc/engine/check_delegate.go#L81

...however, that doesn't seem to rate limit resource types, just resources 🤔

In any event, I think the simplified code is nicer, I'm happy with it 👍

atc/lidar/scanner_test.go
```diff
@@ -168,6 +168,23 @@ var _ = Describe("Scanner", func() {
...
+						By("checkFactory.Resources should not return any put-only resources")
+						fakeResourceType.NameReturns("put-only-custom-type")
+						fakeResourceType.PipelineIDReturns(1)
+					})
+
+					It("creates a check for only the parent and not the put-only resource", func() {
+						Expect(fakeCheckFactory.TryCreateCheckCallCount()).To(Equal(2))
+
+						By("first check is for the default resource")
+						_, checkable, _, _, manuallyTriggered := fakeCheckFactory.TryCreateCheckArgsForCall(1)
```
aoldershaw, line 183: This test is failing https://ci.concourse-ci.org/builds/142874260#L6087a117:1177, should this be:
```suggestion
						_, checkable, _, _, manuallyTriggered := fakeCheckFactory.TryCreateCheckArgsForCall(0)
```

aoldershaw (approved):
LGTM

clarafu:
@taylorsilva Hey I noticed something small that we might want to address! Previously, we only creates checks for resource types when it is used by a resource and we only checked for resources in unpaused pipelines.

https://github.com/concourse/concourse/blob/08612f6c1a099176cb13536584ea074fd36057e5/atc/db/check_factory.go#L142-L144

But now that we create a check for all resource types, I think we probably want to continue to not create a check for resource types in `paused` pipelines. I think we just need to add a small`where` clause in this `checkFactory.ResourceTypes()` query

https://github.com/concourse/concourse/blob/08612f6c1a099176cb13536584ea074fd36057e5/atc/db/check_factory.go#L181-L183

taylorsilva:
Ah great catch! I'll PR and test it 👍 Thanks for call out!

PR #6949, opened 2021-04-28
Set Content-Security-Policy and Cache-Control Headers

taylorsilva:

- Add CSP headers that are configurable. Defaults to `frame-ancestors 'none'`
- Add a default cache-control header set to `no-store, private`

## Notes to reviewer:
~I'm not 100% sure about the cache-control values, but looking at the
header of some big sites (i.e. stackoverflow) I found they do set this header
to `private`.~

## Release Note

* A Content-Security-Policy header is now set with a default value that will block framing of the Concourse web UI. This was already possible with the default value of the X-Frames-Option header.
  * The CSP header value is configurable with `CONCOURSE_CONTENT_SECURITY_POLICY`
* A Cache-Control header is set on every page with a default value of `no-store, private`. The value of the header is overwritten for some paths (i.e. web assets)

## Contributor Checklist

- [x] [Signed] all commits
- [ ] Updated [Documentation]

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
  the [integration
  `--help` text).

### DIFF
--- a/atc/wrappa/security_handler.go
+++ b/atc/wrappa/security_handler.go
@@ -18,6 +18,6 @@ func (handler SecurityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 	w.Header().Set("X-XSS-Protection", "1; mode=block")
 	w.Header().Set("X-Content-Type-Options", "nosniff")
 	w.Header().Set("X-Download-Options", "noopen")
-	w.Header().Set("Cache-Control", "no-store, private")
+	w.Header().Set("Cache-Control", "private")
 	handler.Handler.ServeHTTP(w, r)
 }
--- a/atc/wrappa/security_handler.go
+++ b/atc/wrappa/security_handler.go
@@ -18,5 +18,6 @@ func (handler SecurityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 	w.Header().Set("X-XSS-Protection", "1; mode=block")
 	w.Header().Set("X-Content-Type-Options", "nosniff")
 	w.Header().Set("X-Download-Options", "noopen")
+	w.Header().Set("Cache-Control", "no-store, private")
 	handler.Handler.ServeHTTP(w, r)
 }
--- a/atc/wrappa/security_handler.go
+++ b/atc/wrappa/security_handler.go
@@ -3,13 +3,17 @@ package wrappa
 import "net/http"
 
 type SecurityHandler struct {
-	XFrameOptions string
-	Handler       http.Handler
+	XFrameOptions         string
+	ContentSecurityPolicy string
+	Handler               http.Handler
 }
 
 func (handler SecurityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if handler.XFrameOptions != "" {
 		w.Header().Set("X-Frame-Options", handler.XFrameOptions)
+	}
+	if handler.ContentSecurityPolicy != "" {
+		w.Header().Set("Content-Security-Policy", handler.ContentSecurityPolicy)
 	}
 	w.Header().Set("X-XSS-Protection", "1; mode=block")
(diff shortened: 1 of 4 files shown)


aoldershaw (changes requested):
@EstebanFS and I tried to better understand what these headers do. It sounds like the `frame-ancestors` in `Content-Security-Policy` obsoletes `X-Frame-Options`, which we also allow configuring - however, it isn't supported in IE, so probably makes sense to support both.

For `Cache-Control`, I guess these settings just mean don't cache anything. This makes sense for things that should be private/sensitive (like API responses), but probably isn't the best for things like assets (icons, stylesheets, javascript).

Also interesting that we explicitly set `Cache-Control` to cache these assets: https://github.com/concourse/concourse/blob/master/web/cache_handler.go. It seems like both values are sent back in the response - not sure which takes precedence

Screen Shot 2021-04-29 at 5 30 16 PM

I feel like a default value for `Cache-Control` of `private` might be a bit better, but not sure if that'll satisfy the vulnerability scanner.

analytically:
I'll be adding support for HSTS (https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) in a separate PR once this one is merged.

taylorsilva:
Interesting about both Cache-Control headers being returned. I thought later handlers would overwrite it. Not sure how browsers are meant to handle conflicting headers like that either. We can probably write smarter handler/wrapper so hard assets are cached and everything else is not.

Also thought about setting it to `private` instead of `private, no-store`. Didn't for the reason you gave, not sure if it will satisfy the vulnerability scanner :(

aoldershaw:
> I thought later handlers would overwrite it

I think we can get it to overwrite by using `w.Header.Set` rather than `w.Handler.Add` (in `web/cache_handler`): https://github.com/concourse/concourse/blob/63c757c1a8e8dc7054c0b18d3110c24540756b78/web/cache_handler.go#L14

taylorsilva:
I changed Cache-Control to only have `private`. Also changed the cache_handler to overwrite any previous `cache-control` headers as per your comment. I checked git blame history on that line but didn't see any reason why it **had** to be `Add` instead of `Set`, and I can't think of any myself. Didn't break any unit tests either.

atc/wrappa/security_handler.go
```diff
@@ -18,6 +18,6 @@ func (handler SecurityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 	w.Header().Set("X-XSS-Protection", "1; mode=block")
 	w.Header().Set("X-Content-Type-Options", "nosniff")
 	w.Header().Set("X-Download-Options", "noopen")
-	w.Header().Set("Cache-Control", "no-store, private")
```
aoldershaw: Actually 😁 could we add `no-store` back here? It does seem like a reasonable thing for API responses, assuming my understanding is correct, and I'm guessing `private` alone won't satisfy the vulnerability scanner if it still stores responses to disk.

Sorry about that 😬  I think I felt that if it was difficult overriding the `Cache-Control` header per route, then we'd be better off still allowing assets to be properly cached - but given that it's easy to cache assets while still preventing API responses from being cached, I think `no-store` is reasonable to keep

taylorsilva: Good point! I'll add it back in.

aoldershaw (approved):
LGTM! Note to self - we'll need to update CI to allow embedding iframes like we do with `XFrameOptions`: https://github.com/concourse/infrastructure/blob/dbf08cf0f07856beb2396c851edb531a1ed1f50b/terraform/environments/production/ci-values.yml.tpl#L14

taylorsilva:
I'm gonna go make the packaging PR's for this now.

PR #7002, opened 2021-05-11
containerd: add flag to allow host access

### LINKED ISSUE #6720: containerd: create flag to allow tasks to access host network
## Summary

Some users have a setup where the task needs to communicate with the host network. For Guardian they can do this currently by enabling `CONCOURSE_GARDEN_ALLOW_HOST_ACCESS`. For `containerd` this will have to be implemented, and set by a new flag `CONCOURSE_CONTAINERD_ALLOW_HOST_ACCESS`.

Guardian reference of flag can be found [here](https://github.com/cloudfoundry/guardian/blob/15a56cca9deaa1ad04c9c10f91054e7ce82bf3c0/guardiancmd/command.go#L152).

## Context

* Prior discussion: https://github.com/concourse/concourse/discussions/6608
* Part of #6476

muntac:

cc: @taylorsilva 

closes #6720

Chart PR: https://github.com/concourse/concourse-chart/pull/250
BOSH Release PR: https://github.com/concourse/concourse-bosh-release/pull/150

* [x] Block container from reaching host by default
* [x] Add a flag, `CONCOURSE_CONTAINERD_ALLOW_HOST_ACCESS`, to optionally allow container to reach host
* [x] Allow host access if DNS proxy is enabled

## Notes to reviewer

1. Default behaviour

Set `CONCOURSE_CONTAINERD_DNS_PROXY_ENABLE: "false"` in `docker-compose.yml`.

```
docker-compose up -d
docker inspect concourse_default
# Note down the IPv4 address of concourse_worker_1
fly -t dev login -u test -p test
fly -t dev sp -p tw -c /topgun/task-waiting.yml
fly -t dev up -p tw
fly -t dev tj -j tw/simple-job
fly -t dev i -u http://localhost:8080/teams/main/pipelines/task-waiting/jobs/simple-job/builds/1
# On container
ping [IPv4 address without the CIDR range]
```

Ping should see no response.

2. Optionally allow host network access

Set `CONCOURSE_CONTAINERD_ALLOW_HOST_ACCESS: "true"`. Repeat the steps above. Ping should now work.

## Release Note

* containerd: Previously containers were allowed to reach the host network by default. This is now prevented.
  * To allow traffic to the host network set `CONCOURSE_CONTAINERD_ALLOW_HOST_ACCESS` to true.
  * If DNS proxy is enabled host network access will be turned on implicitly.

### DIFF
--- a/worker/runtime/cni_network.go
+++ b/worker/runtime/cni_network.go
@@ -2,11 +2,8 @@ package runtime
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"net"
 	"path/filepath"
-	"regexp"
 	"strings"
 
 	"github.com/concourse/concourse/worker/runtime/iptables"
@@ -339,38 +336,12 @@ func (n cniNetwork) restrictHostAccess() error {
 		return fmt.Errorf("create chain or flush if exists failed: %w", err)
 	}
 
-	hostIp, err := GetHostIp()
+	err = n.ipt.AppendRule(filterTable, "INPUT", "-i", n.config.BridgeName, "-j", "REJECT", "--reject-with", "icmp-host-prohibited")
 	if err != nil {
-		return err
+		return fmt.Errorf("error appending iptables rule: %w", err)
 	}
-	n.ipt.AppendRule(filterTable, "INPUT", "-i", n.config.BridgeName, "-d", hostIp, "-j", "DROP")
 
 	return nil
-}
-
-func GetHostIp() (string, error) {
-	ifaces, err := net.Interfaces()
-	if err != nil {
-		return "", err
-	}
-	ethInterface := regexp.MustCompile("eth0")
-
-	for _, i := range ifaces {
-		if ethInterface.MatchString(i.Name) {
-			addrs, err := i.Addrs()
-			if err != nil {
-				return "", err
-			}
-			for _, address := range addrs {
-				if ipnet, ok := address.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
-					if ipnet.IP.To4() != nil {
-						return ipnet.IP.String(), nil
-					}
-				}
-			}
-		}
-	}
-	return "", errors.New("unable to find host's IP")
 }
 
 func (n cniNetwork) Add(ctx context.Context, task containerd.Task) error {
--- a/worker/runtime/cni_network.go
+++ b/worker/runtime/cni_network.go
@@ -157,9 +157,18 @@ func WithCNIFileStore(f FileStore) CNINetworkOpt {
 
 // WithRestrictedNetworks defines the network ranges that containers will be restricted
 // from accessing.
+//
 func WithRestrictedNetworks(restrictedNetworks []string) CNINetworkOpt {
 	return func(n *cniNetwork) {
 		n.restrictedNetworks = restrictedNetworks
+	}
+}
+
+// WithAllowHostAccess allows containers to talk to the host
+//
+func WithAllowHostAccess() CNINetworkOpt {
+	return func(n *cniNetwork) {
+		n.allowHostAccess = true
 	}
 }
 
@@ -178,6 +187,7 @@ type cniNetwork struct {
 	nameServers        []string
 	binariesDir        string
 	restrictedNetworks []string
+	allowHostAccess    bool
 	ipt                iptables.Iptables
 }
 
@@ -233,10 +243,14 @@ func (n cniNetwork) SetupHostNetwork() error  {
 	if err != nil {
 		return err
 	}
-	err = n.restrictHostAccess()
-	if err != nil {
-		return err
+
+	if !n.allowHostAccess {
+		err = n.restrictHostAccess()
+		if err != nil {
+			return err
+		}
 	}
+
 	return nil
 }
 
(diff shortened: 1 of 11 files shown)


worker/runtime/cni_network.go
```diff
@@ -303,6 +333,46 @@ func (n cniNetwork) generateResolvConfContents() ([]byte, error) {
...
+	err := n.ipt.CreateChainOrFlushIfExists(filterTable, "INPUT")
+	if err != nil {
+		return fmt.Errorf("create chain or flush if exists failed: %w", err)
+	}
+
+	hostIp, err := GetHostIp()
+	if err != nil {
+		return err
+	}
+	n.ipt.AppendRule(filterTable, "INPUT", "-i", n.config.BridgeName, "-d", hostIp, "-j", "DROP")
```
aoldershaw, line 346: Is there a reason for using `DROP` rather than `REJECT`? Seems like the main difference is that with `DROP`, trying to reach the host will timeout (rather than with `REJECT`, which will fail ASAP)

worker/runtime/cni_network.go
```diff
@@ -303,6 +333,46 @@ func (n cniNetwork) generateResolvConfContents() ([]byte, error) {
...
+
+	return nil
+}
+
+func GetHostIp() (string, error) {
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return "", err
+	}
+	ethInterface := regexp.MustCompile("eth0")
```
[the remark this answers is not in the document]
muntac: Great catch! Will change this.

worker/runtime/cni_network.go
muntac: `REJECT` explicitly responds to the attacker and lets them know that this IP exists. `DROP` prevents leaking that information.

aoldershaw: That brings up another question, actually - does this rule only `DROP` packets destined for the IP of the `eth0` (or equivalent) interface? Will it prevent access to any IP that's accessible within the host's network?

e.g. I was investigating a customer issue where a VM's BOSH DNS server was inaccessible to containers, which was causing DNS issues. This DNS server had IP `169.254.0.2` (which wasn't the same as `eth0` on the host). Would this rule handle that case?

muntac: Yup. It won't affect any other IPs. We tested this by bringing up a second worker and were able to ping that worker from the container. Guardian has the same behaviour.

Initially, we were wondering if we should try to stop that too but we concluded VM to VM network interactions in the network are up to the operator to determine through their firewall rules (they can technically prevent it through `CONCOURSE_CONTAINERD_RESTRICTED_NETWORK` as well).

muntac: Not quite sure why the DNS issue came up in the customer issue. Possible it was their internal firewall rules or some nuance about Guardian (if they're using that runtime).

aoldershaw: Sorry, should've clarified - it was a result of allow host access (i.e. with host access disabled, they couldn't access the internal DNS server, and could access it when enabled).

I guess the question I have is what's the reason for the different behaviour from Guardian? My interpretation of Guardian's `iptables` rules here are:
1. If packets are destined for the external network interface (eth0), then accept them
2. If packets are destined for any other interface, then reject if host access is disabled, and accept if enabled

...which seems like it should work more generally, as it seems local services can bind to other IP addresses (not sure how it works, but e.g. BOSH DNS binding to 169.254.0.2). Let me know if my interpretation is flawed, not 100% sure about how `iptables`/linux networking works evidently

taylorsilva:
@aoldershaw @muntac I made a change to the IP table rule being set. Read the commit message for an explanation: https://github.com/concourse/concourse/pull/7002/commits/a2629367cd006cacd1ea42f35928e50dc763cf7b

aoldershaw:
@taylorsilva I tried modifying the integration test to test the *opposite* case (i.e. with allow host access enabled, it should be able to reach the local server) - but it still fails. Like with guardian, do we need to add the opposite rule? (`-j ACCEPT` if allow host access is enabled)

taylorsilva:
@aoldershaw in that case the rule is not written at all and the `INPUT` chain's default policy of `ACCEPT` would allow the packet through.

I'm trying to test this right now but I'm getting a timeout issue, which shouldn't happen since I changed it to return a response upon rejection 😡  Will look into this more on Monday.

taylorsilva:
I think I see why it failed. The iptables rules are not flushed between tests, so it's possible for the env to become polluted and subsequent tests to fail. Will update the tests with a fix.

taylorsilva:
I ended up NOT being wrong, but in a different way 😅 

Tests weren't being cleaned up properly and for tests that used a custom backend they didn't cleanup stuff made in the `BeforeTest()` before making their own backend. They were starting in a dirty environment.

Everything should work as expected now!

aoldershaw (approved):
LGTM. I was a bit concerned the `REJECT` rule to the bridge interface would make https://github.com/concourse/rfcs/pull/84 impossible (without some additional service manager component running on the worker to selectively grant containers access to others). However, it doesn't seem to - I disabled the DNS proxy, spun up two tasks, and was able to communicate between them using `nc`. Is this because it's on the `INPUT` chain rather than the `FORWARD` chain, and packets from one container aren't "entering the system" from the outside?

taylorsilva:
> Is this because it's on the INPUT chain rather than the FORWARD chain, and packets from one container aren't "entering the system" from the outside?

This area is still kinda grey for me. Haven't figured out a good way to verify behaviour around interfaces + iptable rules, but the last part you said, the packet ""entering the system" from the outside", is the key part I think and is what causes it to not hit the REJECT rule in this case.

PR #7003, opened 2021-05-11
Fly clear-resource-cache command

### LINKED ISSUE #1038: Removing cached volumes

EstebanFS:

closes #1038

* [x] Added fly command `clear-resource-cache`

## Notes to reviewer
Added fly command `clear-resource-cache`
You could use this command following this structure  `fly -t ci clear-resource-cache -r pipeline/resource [--version some:version]`. 
The version flag is optional in case you want to specify the version of the resource-cache you would like to delete.
This command will return the numbers of caches deleted in the following format `1 caches removed`

## Release Note
* Added fly command `clear-resource-cache`, you could use this following the next format 
`fly -t ci clear-resource-cache -r pipeline/resource [--version some:version]`

### DIFF
--- a/atc/api/resources_test.go
+++ b/atc/api/resources_test.go
@@ -1613,6 +1613,25 @@ var _ = Describe("Resources API", func() {
 		var response *http.Response
 		var fakeResource *dbfakes.FakeResource
 
+		executeNoVersionConnection := func() {
+			request, err := http.NewRequest("DELETE", server.URL+"/api/v1/teams/a-team/pipelines/a-pipeline/resources/resource-name/cache", nil)
+			Expect(err).NotTo(HaveOccurred())
+
+			response, err = client.Do(request)
+			Expect(err).NotTo(HaveOccurred())
+		}
+
+		executeVersionConnection := func() {
+			reqPayload, err := json.Marshal(atc.VersionDeleteBody{Version: atc.Version{"ref": "fake-ref"}})
+			Expect(err).NotTo(HaveOccurred())
+
+			request, err := http.NewRequest("DELETE", server.URL+"/api/v1/teams/a-team/pipelines/a-pipeline/resources/resource-name/cache", bytes.NewBuffer(reqPayload))
+			Expect(err).NotTo(HaveOccurred())
+
+			response, err = client.Do(request)
+			Expect(err).NotTo(HaveOccurred())
+		}
+
 		Context("when authenticated ", func() {
 			BeforeEach(func() {
 				fakeAccess.IsAuthenticatedReturns(true)
@@ -1625,13 +1644,7 @@ var _ = Describe("Resources API", func() {
 
 				Context("when it tries to find a resource", func() {
 					JustBeforeEach(func() {
-						var err error
-
-						request, err := http.NewRequest("DELETE", server.URL+"/api/v1/teams/a-team/pipelines/a-pipeline/resources/resource-name/cache", nil)
-						Expect(err).NotTo(HaveOccurred())
-
-						response, err = client.Do(request)
-						Expect(err).NotTo(HaveOccurred())
+						executeNoVersionConnection()
 					})
 
 					It("calls the resource", func() {
@@ -1651,13 +1664,7 @@ var _ = Describe("Resources API", func() {
 
 						Context("when no version is passed", func() {
 							JustBeforeEach(func() {
-								var err error
-
-								request, err := http.NewRequest("DELETE", server.URL+"/api/v1/teams/a-team/pipelines/a-pipeline/resources/resource-name/cache", nil)
-								Expect(err).NotTo(HaveOccurred())
-
-								response, err = client.Do(request)
-								Expect(err).NotTo(HaveOccurred())
+								executeNoVersionConnection()
 							})
 
 							BeforeEach(func() {
@@ -1675,7 +1682,6 @@ var _ = Describe("Resources API", func() {
 							It("it send an empty version", func() {
 								version := fakeResource.ClearResourceCacheArgsForCall(0)
 								expectedVersion := atc.VersionDeleteBody{}.Version
-								fmt.Println(version)
 								Expect(version).To(Equal(expectedVersion))
 							})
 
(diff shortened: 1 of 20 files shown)


fly/commands/clear_resource_cache.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/fly/commands/internal/flaghelpers"
+	"github.com/concourse/concourse/fly/rc"
+)
+
+type ClearResourceCacheCommand struct {
+	Resource 	 flaghelpers.ResourceFlag `short:"r" long:"resource" required:"true" value-name:"PIPELINE/RESOURCE" description:"Name of a resource to clear cache"`
+	Version      *atc.Version             `short:"v" long:"version"                  value-name:"VERSION"           description:"Version of the resource to check from, e.g. digest:sha256@..."`
```
clarafu, line 13: Maybe we should mention in the description that if you don't add a Version, it will result in all the resource caches for that resource being deleted. I think this might be something we want to think about, whether we want to allow all the resource caches to be deleted when they don't pass a version. For example, a user can forget to add a `--version` flag when they are running the command and did not intend to delete all the resource caches.

If we are worried about this happening, maybe we can accept a special `all` value for the version flag and only if `all` is passed then we will delete all the caches. If the flag is not specified then it will error.

fly/commands/clear_resource_cache.go
```diff
@@ -0,0 +1,41 @@
...
+	target, err := rc.LoadTarget(Fly.Target, Fly.Verbose)
+	if err != nil {
+		return err
+	}
+
+	err = target.Validate()
+	if err != nil {
+		return err
+	}
+
```
clarafu, line 27: I'm wondering if we want to display some sort of warning message before we go ahead and delete the resource cache? Maybe something like the one for `clear-task-cache` https://github.com/concourse/concourse/blob/2000fa3b136cdd95fc0b8ad29bcfc7efb36273dd/fly/commands/clear_task_cache.go#L29-L35

What do you think? 🤔

clarafu (changes requested):
Wow great work so far!! 😄  I just have a few points that I'd like you to review and if you have any questions about any of them, feel free to bother me! I'm also open to taking some time to pair on them if you would prefer that as well.

I had quite a few reorganizing comments for the `api/resources_test.go` and they're a little difficult to relay the information well over text so don't stress if you don't quite understand what I'm saying 🤣

fly/commands/clear_resource_cache.go
EstebanFS: I think is a good idea, is better to keep safe with this kind of things

atc/api/resources_test.go
```diff
@@ -1608,4 +1608,260 @@ var _ = Describe("Resources API", func() {
...
+				})
+
+				Context("when finding the resource succeeds", func() {
+					BeforeEach(func() {
+						fakeResource = new(dbfakes.FakeResource)
+						fakeResource.IDReturns(1)
+						fakePipeline.ResourceReturns(fakeResource, true, nil)
+					})
+
+					Context("when clear cache succeeds", func() {
```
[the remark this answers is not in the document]
clarafu: You can actually move the `fakeResource.ClearResourceCacheReturns(1,nil)` on line 1662 to here underneath this context for `"when clear cache succeeds"` because that line for faking out the return value is actually mocking out the situation for when the clear cache succeeds (which is what the context is describing). So it would look something like

```
Context("when clear cache succeeds", func() {
  BeforeEach(func() {
    fakeResource.ClearResourceCacheReturns(1,nil)
  })
})
```

And then the `fakeResource.ClearResourceCacheReturns(1,nil)` line can be removed from line 1662!

atc/api/resourceserver/clear_resource_cache.go
```diff
@@ -0,0 +1,90 @@
...
+
+		resource, found, err := pipeline.Resource(resourceName)
+		if err != nil {
+			logger.Error("failed-to-get-resource", err)
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		var version atc.VersionDeleteBody
+		err = json.NewDecoder(r.Body).Decode(&version)
```
clarafu, line 28: We can actually move this code for decoding of the version to before we try to find a resource from the pipeline on line 20 https://github.com/concourse/concourse/blob/5db952798af1631e9d6f6396c2151dfbdbc005d8/atc/api/resourceserver/clear_resource_cache.go#L20 

That way the checking if the resource is not found https://github.com/concourse/concourse/blob/5db952798af1631e9d6f6396c2151dfbdbc005d8/atc/api/resourceserver/clear_resource_cache.go#L38-L51 is not split off from where the found boolean is set https://github.com/concourse/concourse/blob/5db952798af1631e9d6f6396c2151dfbdbc005d8/atc/api/resourceserver/clear_resource_cache.go#L20

fly/commands/clear_resource_cache.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/vito/go-interact/interact"
+
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/fly/commands/internal/flaghelpers"
+	"github.com/concourse/concourse/fly/rc"
+)
+
+type ClearResourceCacheCommand struct {
+	Resource 	 flaghelpers.ResourceFlag `short:"r" long:"resource" required:"true" value-name:"PIPELINE/RESOURCE" description:"Name of a resource to clear cache"`
+	Version      *atc.Version             `short:"v" long:"version"                  value-name:"VERSION"           description:"Version of the resource to check from, e.g. digest:sha256@..., in case a version is not specified the command will delete all the resource caches for that resource"`
```
clarafu, line 15: There seems to be a little bit of formatting weirdness, I think you can actually fix it by running `go fmt fly/commands/clear_resource_cache.go`!

fly/commands/clear_resource_cache.go
```diff
@@ -0,0 +1,65 @@
...
+		return err
+	}
+
+	err = target.Validate()
+	if err != nil {
+		return err
+	}
+
+	var version atc.Version
+	if command.Version != nil {
```
clarafu, line 31: Since we have a `command.Version != nil` check down below on line 38, we can actually combine this one with the one below. That way we don't need two separate `if command.Version != nil` statements!

clarafu (commented):
Thanks for addressing all my comments!! Looking great so far I just a few more small refactor things I'd like your opinion on! 👍

clarafu (approved):
It looks good to me!! :) I ran some quick manual tests for running the fly clear cache command with a simple pipeline and another time with multiple workers and everything seems to be working appropriately. Great job!!

PR #6982, opened 2021-05-05
worker: Set PATH based on UID instead of container's privileged state

taylorsilva:
## What does this PR accomplish?
**Bug Fix** | Feature | Documentation

## Changes proposed by this PR:
Set a users/processes PATH based on the UID inside the container instead of the privileged state of the container. This matches what guardian does: https://github.com/cloudfoundry/guardian/blob/bf4c737f2bc490aa82b8883e2c8964680236a8df/rundmc/processes/unix_env.go#L44-L46

## Notes to reviewer:
You can test by `fly execute`'ing this task

```yaml
platform: linux
image_resource:
  type: mock
  source: {mirror_self: true}
run:
  user: root
  path: sh
  args:
  - -c
  - |
    id
    echo $PATH
```
Path should include `/sbin`.
If you run this task which run under a different UID:

```yaml
platform: linux
image_resource:
  type: mock
  source: {mirror_self: true}
run:
  user: games
  path: sh
  args:
  - -c
  - |
    id
    echo $PATH
```
You should see a shorter Path that does not contain `/sbin`.

## Release Note

* Containerd: fixed a bug where PATH did not contain directories to system tools (i.e. `/sbin`) when a user/process was root. Only effects unprivileged containers.

## Contributor Checklist

- [x] Followed [Code of conduct], [Contributing Guide] & avoided [Anti-patterns]
- [x] [Signed] all commits
- [x] Added tests (Unit and/or Integration)
- [x] Added release note (Optional)

[Code of Conduct]: https://github.com/concourse/concourse/blob/master/CODE_OF_CONDUCT.md
[Contributing Guide]: https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md
[Anti-patterns]: https://github.com/concourse/concourse/wiki/Anti-Patterns
[Signed]: https://help.github.com/en/github/authenticating-to-github/signing-commits
[Documentation]: https://github.com/concourse/docs

## Reviewer Checklist

- [ ] Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [ ] New config flags added? Ensure that they are added to the
  [BOSH](https://github.com/concourse/concourse-bosh-release) and
  [Helm](https://github.com/concourse/helm) packaging; otherwise, ignored for
  the [integration
  tests](https://github.com/concourse/ci/tree/master/tasks/scripts/check-distribution-env)
  (for example, if they are Garden configs that are not displayed in the
[...]

PR #7023, opened 2021-05-17
worker/runtime: Don't overwrite $PATH

taylorsilva:
Fixes a bug from #6982. Borrowed from https://github.com/cloudfoundry/guardian/blob/bf4c737f2bc490aa82b8883e2c8964680236a8df/rundmc/processes/unix_env.go#L14-L20

If the image contains an existing $PATH then don't overwrite it. If no
existing $PATH is present then one will be set based on the UID of the
user

## Notes to reviewer

Tested this by running a task with the `concourse/unit` image and printing the `env`. I looked for the extra things the image sets in it's PATH like `/go/bin:/usr/local/go/bin`.


aoldershaw (approved):
Not sure if @muntac wants to take a look as well but I know the release is blocked on this so figured I'd give my review

PR #7024, opened 2021-05-17
Fix caching when streamed volume was also streamed

aoldershaw:

this fixes a bug with #6660 - if you streamed a resource cache from worker A -> B, and then streamed that volume from B -> C, the worker_resource_cache on C would previously be associated with the worker_base_resource_type on B (instead of A, which is where the volume originated).

what this would mean is that if the base resource type on A gets invalidated, it would not invalidate the cache on C, even though it should. it also means that if a resource cache was streamed to a windows/darwin worker, and we tried to stream the volume from the windows/darwin worker to another worker, we'd get an error (worker base resource type disappeared, or something like that)

* [x] Initialize the `WorkerResourceCache.FindOrCreate` now takes in the worker base resource type ID
    * For the "normal" `InitializeResourceCache` (i.e. when the get step runs on a worker), we `Find` the appropriate `WorkerBaseResourceType` on that worker and use its ID
    * For `InitializeStreamedResourceCache` (i.e. when we stream a resource cache from one worker to another), we `Find` the source worker's `WorkerResourceCache` and use the worker base resource type ID from there (which could itself be from another worker, if the source worker had its resource cache streamed!).
        * ~Note that we now error if the worker_resource_cache disappears from the source worker. However, if this happens, we'd want to invalidate the cache anyway~
        * If the `worker_resource_cache` disappears from the source worker, we simply don't mark the streamed volume as a resource cache
* [x] When finding a resource cache volume, randomize the order of workers with the resource cache. That way, it won't always be the same worker that needs to stream out volumes.

## Notes to reviewer

I also got rid of `worker_resource_cache_test.go` - it now has even less responsibility, and it's covered in `volume_test.go`


atc/db/volume.go
```diff
@@ -353,37 +354,84 @@ func (volume *createdVolume) findWorkerBaseResourceTypeByBaseResourceTypeID(base
...
 
+	sourceWorkerResourceCache, found, err := WorkerResourceCache{
+		WorkerName:    sourceWorkerName,
+		ResourceCache: resourceCache,
+	}.Find(tx)
+	if err != nil {
+		return err
+	}
+	if !found {
+		return ErrSourceResourceCacheDisappeared
```
aoldershaw, line 403: Is this a case that we want to error on? This could happen if while the volume was being streamed, the original worker (not necessarily the source worker) got a new base resource type version, invalidating the cache. Should we just give up and not initialize the resource cache, but not error?

clarafu: So that would mean that the volume was still able to stream correctly but we just can't initialize the resource cache for it? I think in that case we might be able to not error and just not initialize the cache

If we error that would mean that the `FindOrCreateContainer` for the container that we are trying to stream the volumes to will fail, and technically it shouldn't fail because it was able to stream the volume in correctly?

aoldershaw: Yeah - but it also means it's using a version of the volume that's likely "not the latest", since if the cache was invalidated, presumably there's a newer version of the base resource type. But it was the "latest" when the step started, so it's probably good enough. Will change to not error

clarafu (approved):
Great bug catches!! LGTM 😄

Issue #5135, opened 2020-02-07
containerd: validate if we really need to bring our own `init`


cirocosta:
### what challange are you facing

we're currently bringing our own implementation of an `init` executable that we
start during container setup, and keep around until container shutdown 

https://github.com/concourse/concourse/blob/54991b2f1f17bd7d229a3ebbe6d6c16f3ff495a3/cmd/init/init.c

while it works, and makes sense, **it'd be better if we could *not* bring our
own, as that's one more thing to maintain**.

our current implementation follows `guardian` pretty close when it comes to
getting a container up and running:

	CREATE
		1. create the sandbox
		2. start an init process there

	RUN
		1. run that process in the sandbox

but it turns out that when `runc create` is run, it runs `runc init`, which is
already able to be there holding the sandbbox (so that we can `exec` processes
within that namespace during `RUN`s).

ps.: in the [pr comment] I wrote about how this makes sense even spec-wise.

[pr comment]: https://github.com/concourse/concourse/pull/5114/files#r375319624

### what would make this better

if possible, not have our own `init`.

### references

- `pid_namespace`: http://man7.org/linux/man-pages/man7/pid_namespaces.7.html
- `oci runtime spec`: https://github.com/opencontainers/runtime-spec
- `guardian`: https://github.com/cloudfoundry/guardian
- `runc`: https://github.com/opencontainers/runc

taylorsilva:
# Issue Bankruptcy

This issue is being closed as part of 2025's **Issue Bankruptcy** declaration. [See here for details](https://blog.concourse-ci.org/posts/2025-01-06-concourse-2024-review/)

If you think this issue still has merit then feel free to open a new issue and reference this one.

PR #7042, opened 2021-05-20
containerd: recover from invalid check sessions more quickly on worker restart

aoldershaw:

this change was spurred by seeing checks error for a while after
redeploying our workers, e.g.:

```
run check: check: backend error: Exit status: 500, message: {"Type":"","Message":"task retrieval: no running task found: task 2adc1610-b027-4b0e-41e7-a7f9ebf74ee4 not found: not found","Handle":"","ProcessID":"","Binary":""}
```

the problem is that the ATC thinks the containers are still alive, so
does not invalidate the resource_config_check_session (until it expires
naturally, which can take up to an hour). digging into the worker, I
found that the Container actually *was* still alive - however, as the
error above suggests, the Task was not alive.

in containerd, a Container is just a metadata object that can contain a
Task. a Task is what we usually think of as a Container - a namespaced
process. so, what's happening when the worker restarts is the Container
persists, but the Task (i.e. the process) unsurprisingly gets
terminated. however, that Container doesn't get GC'd until the check
session expires because we report the Container as "alive", when it's in
fact not alive (if we consider the definition of Container used in the
rest of Concourse, i.e. an environment on which processes can be run)

so, with this change, the Garden `Containers` method only reports the
list of active Tasks, so when a Task dies, its corresponding Container
will eventually get GC'd.

note that we technically *could* start up a new Task on the Container if
we find the original task has been terminated, but this requires
additional bookkeeping to ensure we clean up the allocated network
namespaces for all of the tasks created for a Container. it's easier to
just treat a Container as a single-use thing and recreate them when
their process (task) dies

* [x] `Containers` reports only the active `Tasks` rather than all `Containers`
    * It still gives a list of usable `garden.Container`s, despite us only accessing the `Handle()` method currently
    * To avoid looking up each container every time, we wrap the Container in a `LazyContainer` that only looks up the container when a method is called on it

## Notes to reviewer

This mostly comes from how we use containerd, which doesn't seem to be very typical. When we create a Container, we spawn a Task running the `gdn-init` binary and wait for Concourse to "start a process in that task", in a sense.

What the more typical approach seems to be is to:
1. Create a Container (but not a Task up-front)
2. When Concourse tries to run something, only then spawn a Task that *runs the actual process*, not the `gdn-init` binary
3. When the process exits, remove the Task

There may be good reasons why we don't do this, I'm just not too sure what they are

## Release Note

* 

### DIFF
--- a/worker/runtime/backend.go
+++ b/worker/runtime/backend.go
@@ -327,7 +327,7 @@ func (b *GardenBackend) Containers(properties garden.Properties) ([]garden.Conta
 	containers := make([]garden.Container, len(tasks))
 	for i, task := range tasks {
 		containers[i] = &LazyContainer{
-			ID: task.ContainerID,
+			ID: task.ID,
 
 			client:        b.client,
 			killer:        b.killer,
--- a/worker/runtime/backend.go
+++ b/worker/runtime/backend.go
@@ -308,28 +308,34 @@ func (b *GardenBackend) Destroy(handle string) error {
 // Containers lists all containers filtered by properties (which are ANDed
 // together).
 //
-func (b *GardenBackend) Containers(properties garden.Properties) (containers []garden.Container, err error) {
+func (b *GardenBackend) Containers(properties garden.Properties) ([]garden.Container, error) {
 	filters, err := propertiesToFilterList(properties)
 	if err != nil {
-		return
+		return nil, err
 	}
 
-	res, err := b.client.Containers(context.Background(), filters...)
+	ctx := context.Background()
+	// note: we fetch Tasks rather than Containers deliberately. Tasks are the
+	// actual running processes on workers, while Containers are just metadata
+	// that can contain a Task. if the Task gets terminated (e.g. the worker
+	// reboots), the worker should not report this container as alive
+	tasks, err := b.client.Tasks(ctx, filters...)
 	if err != nil {
-		err = fmt.Errorf("list containers: %w", err)
-		return
+		return nil, fmt.Errorf("list tasks: %w", err)
 	}
 
-	containers = make([]garden.Container, len(res))
-	for i, containerdContainer := range res {
-		containers[i] = NewContainer(
-			containerdContainer,
-			b.killer,
-			b.rootfsManager,
-		)
+	containers := make([]garden.Container, len(tasks))
+	for i, task := range tasks {
+		containers[i] = &LazyContainer{
+			ID: task.ContainerID,
+
+			client:        b.client,
+			killer:        b.killer,
+			rootfsManager: b.rootfsManager,
+		}
 	}
 
-	return
+	return containers, nil
 }
 
 // Lookup returns the container with the specified handle.
--- a/worker/runtime/backend_test.go
+++ b/worker/runtime/backend_test.go
@@ -10,6 +10,7 @@ import (
 	"code.cloudfoundry.org/garden"
 	"code.cloudfoundry.org/garden/gardenfakes"
 	"github.com/concourse/concourse/worker/runtime"
+	"github.com/concourse/concourse/worker/runtime/libcontainerd"
 	"github.com/concourse/concourse/worker/runtime/libcontainerd/libcontainerdfakes"
 	"github.com/concourse/concourse/worker/runtime/runtimefakes"
 	"github.com/containerd/containerd"
@@ -258,11 +259,11 @@ func (s *BackendSuite) TestCreateContainerLockTimeout() {
 	}
 }
 func (s *BackendSuite) TestContainersWithContainerdFailure() {
-	s.client.ContainersReturns(nil, errors.New("err"))
+	s.client.TasksReturns(nil, errors.New("err"))
 
 	_, err := s.backend.Containers(nil)
 	s.Error(err)
-	s.Equal(1, s.client.ContainersCallCount())
+	s.Equal(1, s.client.TasksCallCount())
 }
 
 func (s *BackendSuite) TestContainersWithInvalidPropertyFilters() {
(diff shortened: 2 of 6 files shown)


aoldershaw:
@muntac I realize this doesn't work exactly as expected currently. Workers will periodically report their list of containers to the ATC, and I thought that if a container was thought to be alive in Concourse but the worker reports it as missing, it'd get GC'd from Concourse's internal state, but that doesn't seem to be the case. We call `DestroyUnknownContainers`:
https://github.com/concourse/concourse/blob/9290686709c3c2aa8e86643d171d3abe0967dfd6/atc/api/containerserver/report.go#L46

https://github.com/concourse/concourse/blob/9290686709c3c2aa8e86643d171d3abe0967dfd6/atc/db/container_repository.go#L391

...but it seems as though all that does is `INSERT` containers (marked as "destroying") in the DB if they exist on the worker but didn't yet exist in the DB. We add them to the DB as "destroying" so GC will kick in and actually destroy the missing containers.

What we *don't* do is mark containers that should exist (as they're "created" in the DB), but don't exist (on the worker), as "destroying". Not sure if this is deliberate or not

muntac:
@aoldershaw who calls the endpoint that invokes `ReportWorkerContainers` in `report.go`?

aoldershaw:
> Does the TSA routinely send a list of container handles from the worker to the ATC?

@muntac yup. Specifically, there's a `container_sweeper` on the workers that calls the `Containers` garden method and reports that to the TSA:
https://github.com/concourse/concourse/blob/cfa2200f7ce18fcd6bb6ebfcca617e41eb1f65fa/worker/container_sweeper.go#L61-L76

Then, the TSA hits that API endpoint:
https://github.com/concourse/concourse/blob/cfa2200f7ce18fcd6bb6ebfcca617e41eb1f65fa/tsa/worker_status.go#L41-L48

---

Btw, this feature is broken as its currently implemented. If a Task disappears from a worker, the ATC will eventually (i.e. after the 5m `MissingGracePeriod`) delete the container from the DB. However, even if the (containerd) Task is missing, the (containerd) Container might still exist - and if the ATC just deletes it from the DB, it won't instruct the runtime to delete the Container (on the worker side). So, there's the possibility of leaking containers.

Not sure of the best approach here tbh. If we can find a way to create Tasks on-the-fly (rather than at Container creation time - see the `Notes to Reviewer` section), this would probably be better than reporting active Tasks.

Alternatively, we could have some sort of reconciler process running on the workers that compares the list of Tasks to the list of Containers, and if there's a Container without a Task, it'll destroy it.

muntac:
I think the tasks approach can work. But got carried away trying to see if we can still use the containers object. Bit of a divergence below. Just wanted to write down an approach I was thinking through. No need to follow me down the rabbit hole.

-----------
**K8s and CRI**
I was trying to see what K8s does for container lifecycle management. Since K8s only talks to `containerd` via CRI I looked at that interface and containerd's CRI package. 

CRI has a [ListContainers](https://github.com/kubernetes/cri-api/blob/1a76359308d012205ec79297f166a47a49d93ddb/pkg/apis/runtime/v1alpha2/api.proto#L77) function which `containerd` [implements](https://github.com/containerd/containerd/blob/96df20cf770479393342047a62182961f9205d62/pkg/cri/server/instrumented_service.go#L181) in their CRI package.

This package uses a higher level Container object that wraps the standard `containerd` [Container](https://github.com/containerd/containerd/blob/96df20cf770479393342047a62182961f9205d62/pkg/cri/store/container/container.go#L39) metadata object we're familiar with. Importantly, the wrapper has a status field. When the ListContainers() returns containers to K8s in a [CRI container format](https://github.com/containerd/containerd/blob/96df20cf770479393342047a62182961f9205d62/pkg/cri/server/container_list.go#L50) this `status` field is passed along.

The `status` field as [dictated](https://github.com/kubernetes/cri-api/blob/1a76359308d012205ec79297f166a47a49d93ddb/pkg/apis/runtime/v1alpha2/api.pb.go#L185-L190) by the K8s api has four states:
```
const (
	ContainerState_CONTAINER_CREATED ContainerState = 0
	ContainerState_CONTAINER_RUNNING ContainerState = 1
	ContainerState_CONTAINER_EXITED  ContainerState = 2
	ContainerState_CONTAINER_UNKNOWN ContainerState = 3
)
```

In the containerd CRI implementation these get updated at different [points](https://github.com/containerd/containerd/blob/96df20cf770479393342047a62182961f9205d62/pkg/cri/store/container/status.go#L31-L61) but I don't understand the details. I just see that on [K8s side](https://github.com/kubernetes/kubernetes/blob/4d4b530114a6a160c0366673c73de607eeb54453/pkg/kubelet/stats/cri_stats_provider.go#L767) the states get used to make decisions like removing the containers.

----------
**Garden and Guardian**
Garden has a similar `state` field that would be returned in its [Container.Info](https://github.com/cloudfoundry/garden/blob/580cadd489d22cc37ec8fb9836ebb55cf04bb01a/container.go#L30) function. We don't [implement](https://github.com/concourse/concourse/blob/705b37778d2b361c68131d0a2e237f090766bed7/worker/runtime/container.go#L228) this on our side.

This Garden `state` field has two states - "active" and "stopped". Guardian implements the Info() function and [setting the state](https://github.com/cloudfoundry/guardian/blob/f2e380e2872d8a2c1aaffa95239a54492e290fcf/gardener/container.go#L62) based on container spec.

----------
So I was thinking if we can implement the Info() function and use read the state on ATC side to make the right decisions.

muntac:
Hm none of what I said addresses the issue you brought up around gdn-init.

1. FindOrCreateContainer() - ultimate sets off backend.Create() which creates the container metadata object and starts a task T that contains the gdn-init process.

2. RunTaskStep() - ultimately sets off container.Run() which starts off a process P in task T.

If I understand correctly you're asking why can't we avoid creating task T in step 1 and instead do it in step 2 directly with process P?

muntac:
There was this issue around not using our own init opened a while ago https://github.com/concourse/concourse/issues/5135

@cirocosta says that he was trying to follow what Guardian did when he tried the init approach. And then @vito noted in a PR [comment](https://github.com/concourse/concourse/pull/5114/files#r374734089), specifically in the Dockerfile line 42 - can't seem to link the convo directly, that he was able to get away without an init by creating an empty task for the container.

aoldershaw:
The idea of using `Container.Info` is interesting and is probably a better solution than modifying the `Containers` endpoint, which requires some additional synchronization on the worker side.

> If I understand correctly you're asking why can't we avoid creating task T in step 1 and instead do it in step 2 directly with process P?

Yup, that's the idea.

> he was able to get away without an init by creating an empty task for the container

That's interesting, I wonder how that works. As best I understand it, the task's pid is required to run it in a separate network namespace: https://github.com/concourse/concourse/blob/705b37778d2b361c68131d0a2e237f090766bed7/worker/runtime/cni_network.go#L407

Presumably an unstarted task wouldn't have a pid, right? Or if it does somehow and there is something actually running, would that empty task still be running after worker restart?

aoldershaw:
Taking a look through containerd, I found this:
https://github.com/containerd/containerd/blob/96df20cf770479393342047a62182961f9205d62/runtime/restart/restart.go#L17-L29
```go
// Package restart enables containers to have labels added and monitored to
// keep the container's task running if it is killed.
//
// Setting the StatusLabel on a container instructs the restart monitor to keep
// that container's task in a specific status.
```

Not sure if the restart monitor needs to be explicitly enabled, but this `SetStatus` may be interesting - we could set the desired status to `Running`, and if the task is ever terminated, containerd should bring it back up (until the container is removed).

aoldershaw:
Going to close this for now as I don't really have time to work on it.

PR #7050, opened 2021-05-20
atc: make resource cache streaming opt-in

taylorsilva:

Switch resource cache streaming from being opt-out to opt-in instead.

## Notes to reviewer

We found that the disk usage for non-linux workers increased by quite a
bit with this feature enabled. It actually filled the disk for our
darwin worker and forced us to disable the feature in order to unblock
our pipelines.


aoldershaw:
@clarafu and I beat you to it slightly 😁 https://github.com/concourse/concourse/pull/7049

PR #7058, opened 2021-05-20
Cleanup gomod and skip building fly by default in Dockerfile

chenbh:

- get rid of that random gomod replace directive since it's causing the build to skip `go mod download`
- building fly takes over a minute for me, it's also wasted time since it builds the linux version (I think the originally use case was for test suites)
- fly can still be built using the `--target=with-fly` during docker build
- update issue template for new releases. The newly created issue should be added to the release milestone for easier tracking

## Notes to reviewer

I _think_ this shouldn't break any of our pipelines because we now build `unit-image` and `dev-image` from dedicated dockerfiles in the ci repo and [pass it in directly](https://github.com/concourse/ci/blob/master/tasks/scripts/with-docker-compose#L9-L10) to docker-compose

### DIFF
--- a/.github/ISSUE_TEMPLATE/release-major-minor.md
+++ b/.github/ISSUE_TEMPLATE/release-major-minor.md
@@ -8,10 +8,12 @@ assignees: ''
 
 Steps for a new major/minor release:
 
+* [ ] Add this issue to the `v<M.m.x>` milestone
+
 * [ ] Create your release branch on the `concourse/concourse` github repo with the following format `release/M.m.x` (M being the major version and m being the minor version)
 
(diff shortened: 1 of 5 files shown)


aoldershaw (commented):
I think we'll need to update some places in CI, since we don't *always* use the Dockerfiles in CI for building the dev image.

For instance, in the prs pipeline, we use the Dockerfile in the concourse repo: https://github.com/concourse/ci/blob/02f50354eea0c4ffdd965f2273525e0bb3bde8ef/pipelines/prs.yml#L408-L420

Not sure if it's just limited to the prs pipeline, though it seems we use the Dockerfile in CI for building the dev-image in the main pipeline as you mentioned: https://github.com/concourse/ci/blob/02f50354eea0c4ffdd965f2273525e0bb3bde8ef/tasks/build-dev-image.yml#L6

chenbh:
Created https://github.com/concourse/ci/pull/401

aoldershaw (approved):

PR #7064, opened 2021-05-21
topgun: update regex for capturing instances

taylorsilva:

* [x] update regex to capture bosh instances with no jobs
* [x] store logs from failed tests in `/tmp`

## Notes to reviewer

We verified the regex fixes the problem by intercepting a container and changing the code and re-running the tests

### DIFF
--- a/topgun/common/common.go
+++ b/topgun/common/common.go
@@ -152,14 +152,13 @@ var _ = BeforeEach(func() {
 var _ = AfterEach(func() {
 	test := CurrentGinkgoTestDescription()
 	if test.Failed {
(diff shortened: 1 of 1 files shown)


muntac (approved):
LGTM.

Assumptions I made: 
1. I'm not exactly sure what the `--dns` output looks like but guessing we are expecting it to always be `z1` and that it shows up before the IP. 

2. The `|-` was added in case the process state is empty e.g. `some-instance/a123-b123-c123-d123-e123  -              -      z1      10.10.0.8`.

3. I know the IP is empty when it's an instance that's an errand. But since it says `LoadJobInstances()` where the instance command is called I'll assume we ignore errands.

Issue #6274, opened 2020-11-16
Implement a `read_var` step


vito:
When evaluating `((foo:bar.baz))`, if the value is already present in the build state's `foo` source, its value will be loaded. If not, a `read_var` step will be run.

The `read_var` step will use the var source prototype interface to read a var value from a var source.

When viewing the build in the UI, I would expect to see a step like `read_var: bar` run in the step's setup area (the 'gear' icon introduced for image fetching sub-steps).

The `read_var` step will store the value in a var source named `foo` in the build state, analogous to how the `load_var` step stores the value in a var source named `.` in the local build state.

clarafu:
We're going to approach this by trying to funnel all credential evaluation through the new `get_var` step. Any evaluation that happens in `exec/` should be OK already, because that happens in the context of a build, but we'll have to find other places (i.e. API handlers) and figure out how to change them, because running a `get_var` step probably won't be a good idea once it implies container creation and such.

* `api/resourceserver`: Resource checking w/ webhook token evaluates the webhook token
  * this is a little strange
* `api/configserver`: `fly set-pipeline --check-creds`
  * this could be done by `fly` instead, but submitting a one-off build with a bunch of `get_var` steps. this could maybe even happen *before* setting the config, if `fly` parses it and extracts out all the vars. but there'd need to be a way to include the var sources themselves in the build plan.
* `(db.Team).FindCheckContainer`: evaluates creds in order to find the scope that the check container points to. gross!
  * this would be much easier if we didn't re-use check containers 👀 
* `(db.Pipeline).Variables`: evaluates creds within var source config
  * this whole method might be able to go away if the `get_var` step handles this itself 🤔

clarafu:
Some implementation details:

- The structure of the `GetVarPlan` will be 
```
type GetVarPlan struct {
  Name string
  Path string
  Type string
  Source atc.Source
}
```
The `name` field represents the name of the var source, `path` is the path within the var source to fetch the var from, `type` is the type of var source and `source` is the source of the var source.

- There will be a global cache for caching credentials that are fetched from the var source. The cache key will be a hash of specific fields within the `GetVarPlan` consisting of the `path`, `type` and `source` of the var. In the future with prototypes, the `type` will need to be replaced with the nested properties of parent types if necessary. This is because the `type` string will become a user configured string of a `prototype` that is declared by the user, rather than now the `type` is a hardcoded value that Concourse understands (ex. `dummy` or `vault`). 

[...]

PR #7065, opened 2021-05-21
Implement skeleton for Prototypes

### LINKED ISSUE #7060: Implement Prototype support in pipeline config
## Summary

Allow users to configure `prototypes` in the pipeline config, much like how `resource_types` are configured, e.g.:

```yaml
prototypes:
- name: some-prototype
  type: registry-image
  source: {repository: aoldershaw/some-prototype}

...
```

Prototypes should be able to be checked for new versions, but needn't be checked periodically - they can instead be checked as needed (i.e. prior to a `run:` step executing, just like how we will be checking `resource_types` prior to a `get:` or `put:` step executing. note that we *currently* periodically check `resource_types`, but that is subject to change as part of #6274).

Implementation of the `run` to actually use prototypes in builds will be a part of another issue.

## Context

* RFC: concourse/rfcs#37
* Part of #6905

aoldershaw:

closes #7060 

* [x] Allow prototypes to be configured in the pipeline config
    * They are stored in the DB and are used for build planning
     * A prototype and a resource_type cannot share a name
* [x] Prototype images can be checked. However, there is no automated checking done for prototypes in e.g. lidar
     * The reason for this is that #6274 will change the way resource types are checked (they will no longer be checked directly in lidar, but will be checked when checking the resources that use the resource type). Similarly, resource prototypes will be checked when checking resources that use it, and regular prototypes will be checked when invoking the `run` step
     * I also didn't implement a `fly check-prototype` command for a similar reason - #6274 simplifies the recursive checking flow. After that work gets merged, we can easily add such a fly command
     * I did however implement an API endpoint for checking prototypes, so you can run e.g. `fly -t dev curl /api/v1/teams/main/pipelines/my-pipeline/prototypes/some-prototype/check -- -X POST -H "Content-Type: application/json" --data '{}'` to manually check a prototype
* [x] Allow configuring a `run:` step in the build plan
    * ...which currently doesn't actually do anything special - it just prints a message to the terminal
        Screen Shot 2021-05-21 at 3 18 08 PM
    * The reason for this "dumb" implementation is twofold:
        1. I'm waiting on #6597 to get merged, which will simplify the implementation
        2. We still need to figure out how best to represent inputs/outputs - see concourse/rfcs#103

## Notes to reviewer

You can test this out using the following pipeline:

```yaml
prototypes:
- name: some-prototype
  type: mock
  source: {mirror_self: true}

jobs:
- name: job
  plan:
  - run: some-message
    type: some-prototype
    params:
      foo: bar
```

Note that the config of the prototype doesn't really matter (unless you want to check the prototype image manually) since we never run anything on the prototype image.

### DIFF
--- a/atc/api/accessor/roles.go
+++ b/atc/api/accessor/roles.go
@@ -47,6 +47,7 @@ var DefaultRoles = map[string]string{
 	atc.CheckResource:                 OperatorRole,
 	atc.CheckResourceWebHook:          OperatorRole,
 	atc.CheckResourceType:             OperatorRole,
+	atc.CheckPrototype:                OperatorRole,
 	atc.ListResourceVersions:          ViewerRole,
 	atc.GetResourceVersion:            ViewerRole,
 	atc.EnableResourceVersion:         OperatorRole,
--- a/atc/api/handler.go
+++ b/atc/api/handler.go
@@ -165,6 +165,7 @@ func NewHandler(
 		atc.CheckResource:           pipelineHandlerFactory.HandlerFor(resourceServer.CheckResource),
 		atc.CheckResourceWebHook:    pipelineHandlerFactory.HandlerFor(resourceServer.CheckResourceWebHook),
 		atc.CheckResourceType:       pipelineHandlerFactory.HandlerFor(resourceServer.CheckResourceType),
+		atc.CheckPrototype:          pipelineHandlerFactory.HandlerFor(resourceServer.CheckPrototype),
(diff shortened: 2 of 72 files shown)


taylorsilva (approved):
LGTM

Did acceptance and tried out a few different pipelines to test stuff like validation. Looked over most of the code but skimmed some since a lot of it was extensions of existing patterns in the codebase  or plumbing-esque code. There were a couple nice refactors; I really liked the addition of the `location` type in the validation package 👍

taylorsilva:
Looks like there's some rebasing needed 😬 sorry for taking so long!

PR #7090, opened 2021-05-27
atc: across step logs errors

### LINKED ISSUE #6846: Across step errors don't show an orange outline in the builds page UI
## Summary

It looks like when there is an error within one of the across steps, it doesn't wrap the box with an orange line like it does when any other step errors. This is what I expect it to look like (but with a orange line)

Screen Shot 2021-04-08 at 4 35 36 PM

And this is what it looks like

Screen Shot 2021-04-08 at 4 33 08 PM

## Steps to reproduce

Create a job that uses an across step and have one of the steps error.

## Triaging info

* Concourse version: 7.1.0
* Browser (if applicable):
* Did this used to work? Not sure?

taylorsilva:

closes #6846

## Release Note

* Across step emits an `error` event when one of the sub-steps errors


integration/builds/events/events_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"encoding/json"
+	"strings"
+	"testing"
+
+	"github.com/concourse/concourse/integration/internal/dctest"
+	"github.com/concourse/concourse/integration/internal/flytest"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoggingErrorEvents(t *testing.T) {
```
aoldershaw, line 13: Is there a reason you went with `integration` instead of `testflight` for this? To me, integration is more focused on testing behaviour under different deployment configurations, whereas testflight is more focused on pipeline/execution semantics - which feels like a more natural fit. testflight is also a fair bit lighter weight than integration

integration/builds/events/events_test.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+	}
+
+	for description, test := range cases {
+		t.Run(description, func(t *testing.T) {
+			fly.ExpectExitCode = 2
+			fly.Run(t, "tj", "-j", "test/"+test.JobName, "--watch")
+
+			fly.ExpectExitCode = 0
+			buildsRaw := fly.Output(t, "curl", "api/v1/teams/main/pipelines/test/jobs/"+test.JobName+"/builds")
```
aoldershaw, line 40: Could you use `fly builds -j main/test/JOBNAME --json` instead?

integration/builds/events/events_test.go
taylorsilva: For some reason, I had the thought that we weren't trying to write more `testflight` tests since adding `integration`. Looking over testflight I do see your point though and it seems like a fine place to put this test.

aoldershaw: Perhaps, though I thought the intention was for integration to effectively replace most of `topgun`. I think `testflight` still makes sense as an integration suite for things that are independent of Concourse configuration (e.g. things that should apply to all runtimes)

taylorsilva: Ah I think that's where the mix-up was in my head. topgun != testflight 😅

aoldershaw (approved):
Nice! Cool that it also surfaces the errors to the top-level of the across step, without requiring you to click into the substeps

Screen Shot 2021-05-28 at 3 16 53 PM

Might be nice in the future to annotate which set of vars got resulted in which error rather than just dumping a list of all the errors, but this already is a nice improvement and bug fix.

PR #7102, opened 2021-05-28
atc(fix): fixed a bug in resource check rate limiter.

### LINKED ISSUE #7077: Slow duplicated check build with global resources enabled
## Summary

With global resources enabled, two pipelines with identical git resources will produce duplicated check builds every resource checking interval. One of which will run fine in a few seconds and the other will take ~1 minute to complete successfully but without any output. By "output", I mean that clicking on the check bar in the UI for the slow resource shows nothing.

## Steps to reproduce

Start a clean concourse install with a single linux worker. Started the ATC with `--enable-global-resource --resource-checking-interval 7m --container-placement-strategy limit-active-tasks --max-active-tasks-per-worker 1 --gc-failed-grace-period 1h` the rest is pretty much standard.

```
fly set-pipeline -p pipeline-1 -c simple-pipline-with-a-git-resource.yml
fly set-pipeline -p pipeline-2 -c simple-pipline-with-a-git-resource.yml
fly unpause-pipeline --all
watch -n 10 fly builds | grep check
```

## Expected results

Either a single check build for any pipeline defining the identical resource:
```
836  pipeline-2/resource.git/check      succeeded  2021-05-26@15:36:23+0200  2021-05-26@15:36:25+0200  2s        main  system
```

Or at least that none of the reported build check takes much longer than the other:
```
836  pipeline-2/resource.git/check      succeeded  2021-05-26@15:36:23+0200  2021-05-26@15:36:25+0200  2s        main  system
835  pipeline-1/resource.git/check      succeeded  2021-05-26@15:36:23+0200  2021-05-26@15:36:25+0200  2s        main  system
```

## Actual results

One of the two runs fast and the other stays there doing nothing for a minute or so:
```
836  pipeline-2/resource.git/check      succeeded  2021-05-26@15:36:23+0200  2021-05-26@15:36:25+0200  2s        main  system
835  pipeline-1/resource.git/check      succeeded  2021-05-26@15:36:23+0200  2021-05-26@15:37:23+0200  1m0s      main  system
```

## Additional context

Manually un-pausing the pipelines a few seconds apart resulted in a first check being effectively run on the worker and a second check build, which was not effectively executed, was reported as taking longer:
```
834  pipeline-2/resource.git/check      succeeded  2021-05-26@15:29:32+0200  2021-05-26@15:30:11+0200  39s       main  system
833  pipeline-1/resource.git/check      succeeded  2021-05-26@15:29:11+0200  2021-05-26@15:29:16+0200  5s        main  system
```

After waiting for the resource checking interval, two check builds are created at the same time:
```
836  pipeline-2/resource.git/check      succeeded  2021-05-26@15:36:23+0200  2021-05-26@15:36:25+0200  2s        main  system
835  pipeline-1/resource.git/check      succeeded  2021-05-26@15:36:23+0200  2021-05-26@15:37:23+0200  1m0s      main  system
```

Logs of the ATC (in debug) and the worker.
[concourse-web.log](https://github.com/concourse/concourse/files/6547531/concourse-web.log)
[concourse-worker.log](https://github.com/concourse/concourse/files/6547532/concourse-worker.log)

Disabling global resources fixes the problem but not for the very first check after the ATC is restarted. As can be seen in the logs below, check build ids 845 and 846 are initialized at 2021-05-26T14:12:09Z but 846 finishes at 2021-05-26T14:12:11Z while 845 is effectively started at 2021-05-26T14:13:01Z and finishes at 2021-05-26T14:13:05Z. Then they will each run in a few seconds, one minute apart. Note that they kept sharing the same container.
[concourse-web-no-g-r.log](https://github.com/concourse/concourse/files/6547670/concourse-web-no-g-r.log)

By the way, `fly containers` never update the build id it reports for a check container...
```
handle                                worker                       pipeline    job   build #  build id  type   name          attempt
b9e1736c-018c-4066-5b00-af3c76eafe2d  wrk-lin-1-0ea983261597099b9  pipeline-2  none  check    851       check  resource.git  n/a
```
While `fly builds` actually reports 860 after some times instead of 851:
```
860  pipeline-2/resource.git/check      succeeded  2021-05-26@17:02:22+0200  2021-05-26@17:02:24+0200  2s        main  system
```

## Triaging info

* Concourse version: 7.2.0
* Runtime: garden
* Browser (if applicable): Firefox but I don't think that matter as the `fly build` output exhibit the same delayed result for one of the pipeline.
* Did this used to work? Yes. We noticed these unresponsive check after updating from 6.7.2 straight to 7.2.0.

evanchaoli:

closes #7077

In pre-7.x, checks are created per resource scope, so check rate limit is calculated by `total resource scopes / check interval`.

In 7.x, check builds are created per resource, but rate limit is still calculated by resource scope count, which leads to smaller rate limit, and causes slow checks.

## Notes to reviewer

I also proposed the other solution in #7103, and I feel that one is better.

## Release Note

* Fixed a bug in check rate limiter that caused slow checks.

### DIFF
--- a/atc/db/resource_check_rate_limiter.go
+++ b/atc/db/resource_check_rate_limiter.go
@@ -3,6 +3,7 @@ package db
 import (
 	"context"
 	"fmt"
+	sq "github.com/Masterminds/squirrel"
 	"sync"
 	"time"
 
@@ -88,7 +89,8 @@ func (limiter *ResourceCheckRateLimiter) Limit() rate.Limit {
 func (limiter *ResourceCheckRateLimiter) refreshCheckLimiter() error {
 	var count int
 	err := psql.Select("COUNT(id)").
-		From("resource_config_scopes").
+		From("resources").
+		Where(sq.Eq{"active": true}).
 		RunWith(limiter.refreshConn).
 		QueryRow().
 		Scan(&count)
--- a/atc/db/resource_check_rate_limiter_test.go
+++ b/atc/db/resource_check_rate_limiter_test.go
@@ -2,10 +2,10 @@ package db_test
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"code.cloudfoundry.org/clock/fakeclock"
-	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -32,6 +32,11 @@ var _ = Describe("ResourceCheckRateLimiter", func() {
 		refreshInterval = 5 * time.Minute
 		fakeClock = fakeclock.NewFakeClock(time.Now())
 
+		_, err := psql.Delete("resources").
+			RunWith(dbConn).
+			Exec()
+		Expect(err).ToNot(HaveOccurred())
+
 		checkableCount = 0
 
 		ctx = context.Background()
@@ -56,14 +61,15 @@ var _ = Describe("ResourceCheckRateLimiter", func() {
 	}
 
 	createCheckable := func() {
-		config, err := resourceConfigFactory.FindOrCreateResourceConfig(
-			defaultWorkerResourceType.Type,
-			atc.Source{"some": "source", "count": checkableCount},
-			atc.VersionedResourceTypes{},
-		)
+		es := dbConn.EncryptionStrategy()
+		encryptedPayload, nonce, err := es.Encrypt([]byte("{}"))
 		Expect(err).ToNot(HaveOccurred())
 
-		_, err = config.FindOrCreateScope(nil)
+		_, err = psql.Insert("resources").
+			Columns("name", "pipeline_id", "config", "active", "nonce", "type").
+			Values(fmt.Sprintf("resource-%d", checkableCount), defaultPipeline.ID(), encryptedPayload, true, nonce, defaultWorkerResourceType.Type).
+			RunWith(dbConn).
+			Exec()
 		Expect(err).ToNot(HaveOccurred())
 
 		checkableCount++
--- a/atc/db/resource_check_rate_limiter_test.go
+++ b/atc/db/resource_check_rate_limiter_test.go
@@ -6,10 +6,14 @@ import (
 	"time"
 
 	"code.cloudfoundry.org/clock/fakeclock"
-	"github.com/concourse/concourse/atc/db"
+	"golang.org/x/time/rate"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
-	"golang.org/x/time/rate"
+
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/atc/db"
+	"github.com/concourse/concourse/atc/db/dbtest"
 )
 
 var _ = Describe("ResourceCheckRateLimiter", func() {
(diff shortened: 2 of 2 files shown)


atc/db/resource_check_rate_limiter_test.go
```diff
@@ -56,14 +61,15 @@ var _ = Describe("ResourceCheckRateLimiter", func() {
...
-			defaultWorkerResourceType.Type,
-			atc.Source{"some": "source", "count": checkableCount},
-			atc.VersionedResourceTypes{},
-		)
+		es := dbConn.EncryptionStrategy()
+		encryptedPayload, nonce, err := es.Encrypt([]byte("{}"))
 		Expect(err).ToNot(HaveOccurred())
 
-		_, err = config.FindOrCreateScope(nil)
+		_, err = psql.Insert("resources").
```
vito, line 68: can we change this to use the `dbtest` package instead? it's best to avoid running raw queries since they make the tests and schema hard to maintain.

something like this should work:

```go
var scenario *dbtest.Scenario

BeforeEach(func() {
  scenario = dbtest.Scenario() // no initial setup
})

createCheckable := func() {
  checkableCount++

  var resources atc.ResourceConfigs
  for i := 0; i < checkableCount; i++ {
    resources = append(resources, atc.ResourceConfig{
      Name: fmt.Sprintf("resource-%d", i)),
      Type: dbtest.BaseResourceType,
      Source: atc.Source{"some": "source"},
    })
  }

  scenario.Run(builder.WithPipeline(atc.Config{
    Resources: resources,
  })
}
```

atc/db/resource_check_rate_limiter_test.go
```diff
@@ -32,6 +32,11 @@ var _ = Describe("ResourceCheckRateLimiter", func() {
 		refreshInterval = 5 * time.Minute
 		fakeClock = fakeclock.NewFakeClock(time.Now())
 
+		_, err := psql.Delete("resources").
```
vito, line 35: guessing this is needed because of the global test suite setup

atc/db/resource_check_rate_limiter.go
```diff
@@ -88,7 +89,8 @@ func (limiter *ResourceCheckRateLimiter) Limit() rate.Limit {
 func (limiter *ResourceCheckRateLimiter) refreshCheckLimiter() error {
 	var count int
 	err := psql.Select("COUNT(id)").
-		From("resource_config_scopes").
+		From("resources").
+		Where(sq.Eq{"active": true}).
```
vito, line 93: hmm the tests don't cover this clause - maybe it could be extended to demonstrate that the rate changes when fewer resources are configured? should be easy enough with another function that just reduces `checkableCount` and saves a new (smaller) set of resources with dbtest.

vito (changes requested):
makes sense to me overall! would only like to change how the tests work.

atc/db/resource_check_rate_limiter_test.go
evanchaoli: Fixed.

atc/db/resource_check_rate_limiter_test.go
evanchaoli: Yes. To cover the test case of `rate.Inf`, have to delete all existing resources. But instead of deleting them, we can also set them inactive.

atc/db/resource_check_rate_limiter.go
evanchaoli: Fixed by resetting the pipeline with an empty resource list, so that all existing resources will be set to inactive.

vito (approved):
thanks!

PR #7103, opened 2021-05-30
atc(fix): fixed a bug for resource check rate limit. solution#2.

### LINKED ISSUE #7077: Slow duplicated check build with global resources enabled

evanchaoli:
Signed-off-by: Evan <chaol@vmware.com>

closes #7077 

This PR fixes the same bug as #7102 using a different solution.

Instead fixing how the rate is calculated, this PR change to only apply rate limit on `shouldRun` checks.

## Notes to reviewer

I feel this solution is better than #7102.

## Release Note

Fixed a bug in check rate limiter that caused slow checks.

### DIFF
--- a/atc/engine/check_delegate.go
+++ b/atc/engine/check_delegate.go
@@ -82,17 +82,7 @@ func (d *checkDelegate) FindOrCreateScope(config db.ResourceConfig) (db.Resource
 func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigScope) (lock.Lock, bool, error) {
 	logger := lagerctx.FromContext(ctx)
 
-	// rate limit periodic resource checks so worker load (plus load on external
-	// services) isn't too spiky
-	if !d.build.IsManuallyTriggered() && d.plan.IsPeriodic() {
-		err := d.limiter.Wait(ctx)
-		if err != nil {
-			return nil, false, fmt.Errorf("rate limit: %w", err)
-		}
-	}
-
 	var err error
-
 	var interval time.Duration
 	if d.plan.Interval != "" {
 		interval, err = time.ParseDuration(d.plan.Interval)
@@ -162,6 +152,18 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
 		}
 
 		return nil, false, nil
+	} else {
+		// rate limit periodic resource checks so worker load (plus load on external
+		// services) isn't too spiky
+		if !d.build.IsManuallyTriggered() && d.plan.IsPeriodic() {
+			err := d.limiter.Wait(ctx)
+			if err != nil {
+				if releaseErr := lock.Release(); releaseErr != nil {
+					logger.Error("failed-to-release-lock", releaseErr)
+				}
+				return nil, false, fmt.Errorf("rate limit: %w", err)
+			}
+		}
 	}
 
 	return lock, true, nil
--- a/atc/engine/check_delegate_test.go
+++ b/atc/engine/check_delegate_test.go
@@ -9,7 +9,6 @@ import (
 	. "github.com/onsi/gomega"
 
 	"code.cloudfoundry.org/clock/fakeclock"
-	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/atc/db/dbfakes"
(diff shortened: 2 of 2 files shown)


atc/engine/check_delegate.go
```diff
@@ -162,6 +152,18 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
 		}
 
 		return nil, false, nil
+	} else {
+		// rate limit periodic resource checks so worker load (plus load on external
+		// services) isn't too spiky
+		if !d.build.IsManuallyTriggered() && d.plan.IsPeriodic() {
```
vito, line 158: the rate limiting was originally outside the lock acquisition so that we don't sleep() with the lock held, since that'll prevent manually triggered checks from running. do you think moving it here is worth the tradeoff?

i guess folks don't manually trigger checks very often, but the thought is that when they do they probably want it to happen as quickly as possible. but if this approach is causing bigger problems i'm happy to change it.

side note: if we stop reusing check containers we shouldn't need the locking at all[1], but we'd have to do a bit more work to keep a check cache directory between runs so resources aren't running a fresh `git clone` all the time. we'd also want to verify the performance overhead of creating a fresh check container each time. the initial code change would be simple, at least. https://github.com/concourse/concourse/blob/5b7f0b9b5b1b9e287274e56b1e291be05965458b/atc/exec/check_step.go#L353-L355

[1]: might want to make sure the db code for saving versions is safe without an outer lock

evanchaoli: Yeah, that's a good point. Doing rate limit after acquiring lock is not a good idea.

PR #7112, opened 2021-06-01
Build page shows name of who triggered the build in header line of build page

evanchaoli:

#6369 added the ability to track username who triggered a build, and the info has been stored in the db, exposed via the build api, but not display in the UI yet.

This PR shows username of who triggered the build if the build is triggered manually in the header bar.

show-trggiered-by

If a build is triggered automatically, the it shows as usual:

no-triggered-by

## Notes to reviewer

I'm extremely new to ELM, actually, I just spent 1 hour reading some basic guide of ELM in order to enable me to understand the code. So, feel free to raise any comment.

I haven't updated tests yet, because I want to confirm the UI design first.

## Release Note

* The build page now shows the username of who triggers the build if the build is triggered manually.

### DIFF
--- a/web/elm/benchmarks/Benchmarks.elm
+++ b/web/elm/benchmarks/Benchmarks.elm
@@ -727,6 +727,7 @@ sampleOldModel =
                     , finishedAt = Nothing
                     }
                 , reapTime = Nothing
+                , createdBy = Nothing
                 }
             , prep = Nothing
             , output =
@@ -789,6 +790,7 @@ sampleModel =
     , hasLoadedYet = True
     , notFound = False
     , reapTime = Nothing
+    , createdBy = Nothing
     }
 
 
--- a/web/elm/src/Build/Build.elm
+++ b/web/elm/src/Build/Build.elm
@@ -123,6 +123,7 @@ init flags =
           , hasLoadedYet = False
           , notFound = False
           , reapTime = Nothing
+          , createdBy = Nothing
           }
         , [ GetCurrentTime
           , GetCurrentTimeZone
--- a/web/elm/src/Build/Header/Header.elm
+++ b/web/elm/src/Build/Header/Header.elm
@@ -201,6 +201,7 @@ historyItem model =
     , name = model.name
     , status = model.status
     , duration = model.duration
+    , createdBy = model.createdBy
     }
 
 
@@ -217,6 +218,7 @@ changeToBuild pageType ( model, effects ) =
                             , status = b.status
                             , duration = b.duration
                             , name = b.name
+                            , createdBy = b.createdBy
                         }
                     )
                 |> Maybe.withDefault model
@@ -518,6 +520,7 @@ handleCallback callback ( model, effects ) =
                      , name = b.name
                      , status = b.status
                      , duration = b.duration
+                     , createdBy = b.createdBy
                      }
                         :: model.history
                     )
@@ -564,6 +567,7 @@ handleBuildFetched b ( model, effects ) =
                     , name = b.name
                     , status = b.status
                     , duration = b.duration
+                    , createdBy = b.createdBy
                     }
                     model.history
             , fetchingHistory = True
@@ -608,6 +612,7 @@ handleHistoryFetched history ( model, effects ) =
                                         , name = b.name
                                         , status = b.status
                                         , duration = b.duration
+                                        , createdBy = b.createdBy
                                         }
                                     )
                            )
--- a/web/elm/src/Build/Header/Header.elm
+++ b/web/elm/src/Build/Header/Header.elm
@@ -59,7 +59,7 @@ header session model =
                     False
     in
     { leftWidgets =
-        [ Views.Title model.name model.job
+        [ Views.Title model.name model.job model.createdBy
         , Views.Duration (duration session model)
         ]
     , rightWidgets =
@@ -572,6 +572,7 @@ handleBuildFetched b ( model, effects ) =
             , job = b.job
             , id = b.id
             , name = b.name
+            , createdBy = b.createdBy
           }
         , effects
         )
--- a/web/elm/src/Build/Header/Models.elm
+++ b/web/elm/src/Build/Header/Models.elm
@@ -20,7 +20,7 @@ type alias Model r =
         , scrolledToCurrentBuild : Bool
         , history : List HistoryItem
         , duration : Concourse.BuildDuration
-        , createdBy: Concourse.BuildCreatedBy
+        , createdBy : Concourse.BuildCreatedBy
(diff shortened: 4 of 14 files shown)


aoldershaw:
Awesome, thanks for contributing the UI side of this feature as well!

Just to throw another design idea out there, how do you feel about something like:
Screen Shot 2021-06-01 at 11 03 14 AM

That way, we don't take up too much horizontal space (especially when the display name is relatively long), which is already pretty limited on smaller devices. With the extra space we have with this approach, we can be more explicit about what the `by username` actually means (e.g. to clarify that it's the user that created the build).

We can also add ellipses (pretty easily with CSS) if the display name is still too long:

Screen Shot 2021-06-01 at 11 12 46 AM

Will share some code for this design, but curious to hear your thoughts as well.

evanchaoli:
@aoldershaw Basically I'm not a big fan of UI development, so I usually don't have strong opinion on UI design either.

If we put the two designs together:

two-design-compare

I just feel the first one is clearer. I don't super concern about horizontal space occupied, because wide screen nowadays is already populate, and Concourse's users are developers and OPs, I don't think much of them use small screens.

But again, that's not a strong opinion. I'm ok to take up your design as well as your code (thanks a lot for the code sharing). I don't know if you guys have a process to finalize UI design. Maybe we can wait for more people's feedback, then decide which design to use.

vito:
IMO having it so big is a bit much. Folks do use small screens now and then, e.g. on a phone or making use of their wide monitors by having multiple windows side-by-side. I would prefer the smaller text myself.

aoldershaw:
Having it small also makes it feel more like build metadata, rather than feeling like it's part of the build name, which I think is a little more honest.

p.s. I also experimented with it appearing in (what is currently) the `duration` table as another row, but it looked really awkward, particularly for in-flight/pending builds.

evanchaoli:
As stated earlier, I don't have strong opinion on this UI design. Then I'm going to shamelessly borrow Aidan's code. 😸

aoldershaw (commented):
One more improvement we can make: switching builds looks pretty choppy at the moment, because even though we have enough information to immediately know who triggered a build when we switch builds, we don't currently use that information to update the view.

I added artificial network throttling to show you what I mean:

https://user-images.githubusercontent.com/26583442/121558670-dc8e3b00-c9e3-11eb-972c-e9851fc62697.mov

I've attached some code that makes things more snappy:

<details>

<summary>Code</summary>

```diff
diff --git a/web/elm/src/Build/Header/Header.elm b/web/elm/src/Build/Header/Header.elm
index 4191aa59b..970979505 100644
--- a/web/elm/src/Build/Header/Header.elm
+++ b/web/elm/src/Build/Header/Header.elm
@@ -201,6 +201,7 @@ historyItem model =
     , name = model.name
     , status = model.status
     , duration = model.duration
+    , createdBy = model.createdBy
     }
 
 
@@ -217,6 +218,7 @@ changeToBuild pageType ( model, effects ) =
                             , status = b.status
                             , duration = b.duration
                             , name = b.name
+                            , createdBy = b.createdBy
                         }
                     )
                 |> Maybe.withDefault model
@@ -518,6 +520,7 @@ handleCallback callback ( model, effects ) =
                      , name = b.name
                      , status = b.status
                      , duration = b.duration
+                     , createdBy = b.createdBy
                      }
                         :: model.history
                     )
@@ -564,6 +567,7 @@ handleBuildFetched b ( model, effects ) =
                     , name = b.name
                     , status = b.status
                     , duration = b.duration
+                    , createdBy = b.createdBy
                     }
                     model.history
             , fetchingHistory = True
@@ -608,6 +612,7 @@ handleHistoryFetched history ( model, effects ) =
                                         , name = b.name
                                         , status = b.status
                                         , duration = b.duration
+                                        , createdBy = b.createdBy
                                         }
                                     )
                            )
diff --git a/web/elm/src/Build/Header/Models.elm b/web/elm/src/Build/Header/Models.elm
index da082923b..de08797c5 100644
--- a/web/elm/src/Build/Header/Models.elm
+++ b/web/elm/src/Build/Header/Models.elm
@@ -20,7 +20,7 @@ type alias Model r =
         , scrolledToCurrentBuild : Bool
         , history : List HistoryItem
         , duration : Concourse.BuildDuration
-        , createdBy: Concourse.BuildCreatedBy
+        , createdBy : Concourse.BuildCreatedBy
         , status : BuildStatus.BuildStatus
         , disableManualTrigger : Bool
         , now : Maybe Time.Posix
@@ -35,6 +35,7 @@ type alias HistoryItem =
     , name : String
     , status : BuildStatus.BuildStatus
     , duration : Concourse.BuildDuration
+    , createdBy : Concourse.BuildCreatedBy
     }
 
 
diff --git a/web/elm/src/Build/Models.elm b/web/elm/src/Build/Models.elm
index 91ca92f97..36ae00850 100644
--- a/web/elm/src/Build/Models.elm
+++ b/web/elm/src/Build/Models.elm
@@ -49,6 +49,7 @@ type alias ShortcutsModel r =
         , status : BuildStatus.BuildStatus
         , isTriggerBuildKeyDown : Bool
         , duration : Concourse.BuildDuration
+        , createdBy : Concourse.BuildCreatedBy
     }
 
 
diff --git a/web/elm/src/Build/Shortcuts.elm b/web/elm/src/Build/Shortcuts.elm
index c4d13af84..e456d9b79 100644
--- a/web/elm/src/Build/Shortcuts.elm
+++ b/web/elm/src/Build/Shortcuts.elm
@@ -61,6 +61,7 @@ historyItem model =
     , name = model.name
     , status = model.status
     , duration = model.duration
+    , createdBy = model.createdBy
     }
 
```
</details>

Basically, we just need to add it to the `HistoryItem` data structure and then update the `Header` model using that info when we change builds. Without this, we need to wait for the "GET build" API request to resolve.

evanchaoli:
@aoldershaw Thanks a lot for the optimization code. Just added a commit with purely your code.

aoldershaw (changes requested):
Thanks for making those changes! One thing that'd be great is a test for this new behaviour. I wouldn't worry too much about testing the styles - just a test that given a build with a `createdBy` field, check that it renders the text `created by user` in the header.

Happy to help here if you'd like, just let me know

evanchaoli:
@aoldershaw Just updated the tests a little bit with adding "createdBy" in build data. If that's not enough, then I would like your help. :)

aoldershaw (approved):
Thanks!

ThibaultDelaune-pro:
Hi, shoudn't the trigger author be in the [build metadata](https://concourse-ci.org/implementing-resource-types.html#resource-metadata) (for example `$BUILD_TRIGGERER`) ?

Issue #533, opened 2016-07-13
No resource type out-of-the-box with windows worker


iar-wain:
Hi,

According to http://concourse.ci/resource-types.html, Concourse workers should support resource types such as `git` out of the box.

I probably missed something obvious somewhere but it looks like it might not the case for the Windows worker (everything went fine on Linux).

Both server and worker were started on a single Win7 SP1 64 machine with the following command lines:

Server:
`D:\>concourse web --development-mode --external-url http://127.0.0.1:8080 --session-signing-key session_signing_key --tsa-host-key host_key --tsa-authorized-keys authorized_worker_keys`

Worker:
`D:\>concourse worker --work-dir worker --tsa-host 127.0.0.1 --tsa-public-key host_key.pub --tsa-worker-private-key worker_key --peer-ip 127.0.0.1`

And the following pipeline was used:

``` yaml

---
resources:
- name: git-res
  type: git
  source:
    uri: https://github.com/concourse/concourse.git

jobs:
- name: hello-world
  plan:
  - get: git-res
  - task: hello-world
    config:
      platform: windows
      run:
        path: echo
        args: [hello world]
```

When executing the `hello-world` job, an error message was obtained:

```
error
no workers satisfying: resource type 'git'

available workers: 
  - platform 'windows'
```

What did I miss?

Additionally I couldn't find any documentation on how to add custom resource types for Windows workers. Where should I look for that?

Thanks in advance.

vito:
Resources are implemented as Docker images, so there aren't any for Windows yet. Someone could start porting them, though it doesn't sound very fun. At the moment you'll need at least one Linux worker to use resources. The bits will be fetched on Linux and streamed over to Windows.

iar-wain:
Thanks for the quick reply.

What would be my options at the moment for a build process that needs to run on Windows and that has some heavy dependencies (think of SDKs that take 1-2h + reboot to get installed)?

Is there a more convenient option beside:
- having all the dependencies installed beforehand on the Windows worker
- transgress the containers separation and have source/asset repositories ready outside of the containers (200+ GiB), that would only need to get synced by a task to get newer data
- use a single job that contains all the tasks and which gets triggered via the REST API and/or fly
- write the final output (40+ GiB) outside the containers (using an absolute path?)

Sorry if the questions sound silly, I'm quite new to that whole container thing. =)

Thanks in advance!

benmoss:
[...]

PR #7125, opened 2021-06-03
Add page to view all builds/resource versions downstream/upstream from a root resource version

### LINKED ISSUE #6904: Spike: Resource causality flow 2
Continuation of #1538

In scope:
- viewing by resource version
- viewing by timestamp (maybe indirectly by finding a build first?)
- 1 resource version generating multiple versions
- 1 resource version bumping itself

Out of scope:
- any pipeline config changes

chenbh:

closes #6904 

## Notes to reviewer

- it was too hard to adapt the pipeline graph rendering to this view, so we ended up using a separate renderer
- view is scoped to a single pipeline
- build -> resource version -> build relationship is established via ~`passed` constraints (might experiment with the more computationally expensive~ `versionmd5` based query)

## Release Note

- Disabled by default since computing causality for large datasets can be expensive, use `--enable-resource-causality` or `$CONCOURSE_ENABLE_RESOURCE_CAUSALITY=true` to enable the web UI and API endpoint.
    - Most datasets ([like the merge commit for this PR](https://ci.concourse-ci.org/teams/main/pipelines/concourse/resources/concourse/causality/1014864917/downstream)) have < 100 builds and/or resource versions and take < 100ms, but it's possible for some "slow paced" resource versions (i.e. very infrequent new versions) to generate extremely large datasets
    - There is an automatic cutoff at 5000 builds or 25000 resource versions. On our deployment, the call for our [slowest paced resource](https://ci.concourse-ci.org/teams/main/pipelines/concourse/resources/oci-build-task?filter=digest%3Asha256%3A77eee31c5b5f8c426bd4b391fb00c7eaa734593b54b51edfeb22960192056e0f&filter=tag%3A0.7.0) took about ~7 seconds to process, most of which is spent in the DB query
- The causality page can be navigated to from the resource page
Screen Shot 2021-06-03 at 11 37 08 AM
- The causality page displays all the builds and resource versions that was generated from (downstream) or resulted in (upstream) the creation of a particular resource version
Screen Shot 2021-06-03 at 11 25 03 AM
- The downstream graph will put the root resource version on the left whereas the upstream graph will put it on the right
- It takes into account all the intermediate resource versions when computing the final graph. In the picture above, while the resource page only shows that `git version: 123` is a direct input to integrate `#4 & #5`, there is also an indirect link from `git version: 123` -> `test #19` -> ... -> `intermediate-3 version:123` -> `integrate #6 & #6.1`


web/elm/src/Causality/Causality.elm
```diff
@@ -0,0 +1,584 @@
...
+        graph
+
+
+
+-- http://www.graphviz.org/doc/info/shapes.html#html. this should probably use Json.Encode.string to sanitize the output
+
+
+attributes : List ( String, String ) -> String
+attributes =
+    List.map (\( k, v ) -> k ++ "=\"" ++ v ++ "\"") >> String.join " "
```
aoldershaw, line 455: They value (and probably key to be safe, though we're the ones who pick the keys) need to be escaped since they are encoded in the pseudo-html, and can break things otherwise, e.g.

```elm
escape : String -> String
escape =
    String.replace "<" "&lt;"
        >> String.replace ">" "&gt;"
        >> String.replace "&" "&amp;"
        >> String.replace "\"" "&quot;"
        >> String.replace "'" "&#039;"


attributes : List ( String, String ) -> String
attributes =
    List.map (\( k, v ) -> escape k ++ "=\"" ++ escape v ++ "\"") >> String.join " "
```

Note that this no longer requires `Json.Encode.string` to avoid the value containing a quote since we escape it properly

---
Actually, maybe we want an `htmlAttributes` that does this escaping, since this escaping isn't valid for "normal" attributes - but I suppose it doesn't really matter too much

atc/db/resource.go
```diff
@@ -953,3 +956,244 @@ func requestScheduleForJobsUsingResource(tx Tx, resourceID int) error {
...
+	err = json.Unmarshal([]byte(versionStr), &root.Version)
+	if err != nil {
+		return root, false, err
+	}
+
+	buildIDs, err := r.getCausalityBuilds(versionMD5, query)
+	if err != nil {
+		return root, false, err
+	}
+	err = r.getCausalityResourceVersions(buildIDs, &root, updateInput, updateOutput)
```
cirocosta, line 1167: whoopsy? looks like it's unchecked 😁  not sure if intentional/not

web/elm/src/Causality/Causality.elm
chenbh: > Actually, maybe we want an htmlAttributes that does this escaping

Not quite sure I know what this means. Either way, I think we'll need to keep this function because even the cell text also needs to be escaped (i.e. `<TD>&lt;special&amp;name&gt;</TD>`)

web/assets/css/svg.less
```diff
@@ -1,7 +1,39 @@
 @import (reference) "_vars.less";
 
 svg {
-  .node rect { fill: @grey100; }
+  position: absolute;
```
aoldershaw, line 4: This breaks resource icons on the resource page
Screen Shot 2021-06-04 at 12 01 00 PM

web/elm/src/Causality/Causality.elm
aoldershaw: Right, I guess I just meant the way that characters get escaped when they are just regular attributes (as opposed to "HTML" attributes) is different. For instance, `"` should become `&quot;` in HTML, but `\"` in regular attributes.

That said, we don't currently really *need* to escape anything in the regular attributes, though, so that's why I say it doesn't matter too much

aoldershaw: But in any event, c85181d215df0676700d7ce89af89026d722a5c6 isn't sufficient for making things not break - e.g. consider https://ci.concourse-ci.org/teams/main/pipelines/aidan/resources/mock/causality/942884455/downstream

Screen Shot 2021-06-04 at 1 34 44 PM

web/assets/css/svg.less
chenbh: Should be fixed now

atc/db/resource.go
```diff
@@ -953,3 +956,243 @@ func requestScheduleForJobsUsingResource(tx Tx, resourceID int) error {
...
+	resourceVersions := make(map[int]*atc.CausalityResourceVersion)
+	// pre-populate the list with the root
+	resourceVersions[root.ResourceVersionID] = root
+
+	// go through all the inputs and construct the struct, this will partially construct the tree
+	rows, err = psql.Select("r.id", "rcv.id", "r.name", "rcv.version", "i.build_id").
+		From("build_resource_config_version_inputs i").
+		Join("resources r ON r.id = i.resource_id").
+		Join("resource_config_versions rcv ON rcv.version_md5 = i.version_md5 AND rcv.resource_config_scope_id = r.resource_config_scope_id").
+		Where(sq.Eq{"i.build_id": buildIDs}).
```
aoldershaw, line 1063: Is it feasible to do this without a `WHERE build_id IN (....)` condition? I imagine that the number of build ids could be pretty huge, so doing a `JOIN` would probably be more scalable.

e.g.

```sql
WITH RECURSIVE build_ids AS
(
...
)
SELECT ...
FROM build_resource_config_version_inputs i
JOIN ...
JOIN build_ids bi ON i.build_id = bi.id
...
UNION ALL
SELECT ...
FROM build_resource_config_version_outputs o
JOIN ...
JOIN build_ids bi ON o.build_id = bi.id
...
```

chenbh: You mean by embedding the `WITH RECURSIVE` query inside each of the queries for builds, build inputs, and build outputs? Is postgres capable of "learning" from the result of previous executions and make more informed plans?

My main concern is that the `WITH RECURSIVE` query should be the most expensive one, so we should only run it once and use the results of that in the 3 cheaper queries.

chenbh: Or is it possible to do this in a transaction, store the result of the `WITH RECURSIVE` as a temp view/table, and then refer to that table in subsequent queries in the same transaction?

aoldershaw: Well at least the inputs and outputs can be a single query with the `UNION ALL`, but yeah I was forgetting about the "build info" query

I know that the list passed to a `WHERE ... IN (...)`  has a limit of 65000 or so elements - but at that point, the performance is already probably pretty bad. Temp table is an option but feels pretty heavy weight, especially for the common case which shouldn't be too crazy.

Tagging @clarafu - WDYT?

chenbh: I just want to point out that it needs to execute different functions depending on if it's processing the inputs or outputs. Which should be easy to do by adding a `'input' AS type` to the select statement and then special casing on that in the code

https://github.com/concourse/concourse/blob/a7b862ea40fa15d308ff383784b46d9f5fa58345/atc/db/resource.go#L1176-L1183

chenbh:
Fun fact, it took 8.6s to load `oci-build-task`, which is probably the most used resource version on our CI outside of time related ones like `oxygen-mask`. 

https://ci.concourse-ci.org/teams/main/pipelines/concourse/resources/oci-build-task/causality/591570659/downstream

atc/db/resource.go
aoldershaw: re: https://github.com/concourse/concourse/pull/7125#issuecomment-855015082:

It looks like the recursive query is actually quite fast

```
CTE Scan on build_ids  (cost=694.57..709.99 rows=771 width=8) (actual time=1.119..196.107 rows=17056 loops=1)
  CTE build_ids
    ->  Recursive Union  (cost=8.59..694.57 rows=771 width=8) (actual time=1.118..191.759 rows=17056 loops=1)
          ->  Unique  (cost=8.59..8.59 rows=1 width=8) (actual time=1.117..1.278 rows=661 loops=1)
                ->  Sort  (cost=8.59..8.59 rows=1 width=8) (actual time=1.116..1.173 rows=661 loops=1)
                      Sort Key: i.build_id
                      Sort Method: quicksort  Memory: 55kB
                      ->  Index Scan using build_inputs_resource_versions_idx on build_resource_config_version_inputs i  (cost=0.56..8.58 rows=1 width=8) (actual time=0.022..0.973 rows=661 loops=1)
                            Index Cond: ((resource_id = 101) AND (version_md5 = '40fa80af72b9c40a7b218fd3fb5393c0'::text))
          ->  HashAggregate  (cost=66.29..67.06 rows=77 width=8) (actual time=26.320..26.764 rows=2342 loops=7)
                Group Key: bp.to_build_id
                ->  Nested Loop  (cost=0.98..66.09 rows=77 width=8) (actual time=0.566..22.632 rows=21790 loops=7)
                      Join Filter: (bi.build_id = i_1.build_id)
                      ->  Nested Loop  (cost=0.42..48.73 rows=14 width=24) (actual time=0.557..6.702 rows=2535 loops=7)
                            ->  WorkTable Scan on build_ids bi  (cost=0.00..0.20 rows=10 width=8) (actual time=0.000..0.463 rows=2437 loops=7)
                            ->  Index Only Scan using build_pipes_from_build_id_to_build_id_uniq on build_pipes bp  (cost=0.42..4.84 rows=1 width=16) (actual time=0.002..0.002 rows=1 loops=17056)
                                  Index Cond: (from_build_id = bi.build_id)
                                  Heap Fetches: 663
                      ->  Index Only Scan using build_resource_config_version_inputs_uniq on build_resource_config_version_inputs i_1  (cost=0.56..1.17 rows=6 width=8) (actual time=0.003..0.004 rows=9 loops=17745)
                            Index Cond: (build_id = bp.from_build_id)
                            Filter: (resource_id <> 101)
                            Rows Removed by Filter: 0
                            Heap Fetches: 12926
Planning time: 0.469 ms
Execution time: 197.186 ms
```

On the other hand, the massing `WHERE build_id IN (...)` condition seems to be really slow - the `EXPLAIN ANALYZE` has been running for over 3 minutes at this point, still not complete.

So, doing it as a `JOIN` seems like a better option, even if it means needing to run the query twice. For this query:

```sql
WITH RECURSIVE build_ids AS (
		SELECT DISTINCT i.build_id
			FROM build_resource_config_version_inputs i
			WHERE i.resource_id=101 AND i.version_md5='40fa80af72b9c40a7b218fd3fb5393c0'
	UNION ALL
		SELECT DISTINCT bp.to_build_id AS build_id
		FROM build_ids bi
		INNER JOIN build_pipes bp ON bi.build_id = bp.from_build_id
		INNER JOIN build_resource_config_version_inputs i ON i.build_id = bi.build_id
		WHERE i.resource_id!=101
)
SELECT r.id, rcv.id, r.name, rcv.version, i.build_id, 'input' as type
FROM build_resource_config_version_inputs i
JOIN resources r ON r.id = i.resource_id
JOIN resource_config_versions rcv ON rcv.version_md5 = i.version_md5 AND rcv.resource_config_scope_id = r.resource_config_scope_id
JOIN build_ids bi ON i.build_id = bi.build_id
UNION ALL
SELECT r.id, rcv.id, r.name, rcv.version, o.build_id, 'output' as type
FROM build_resource_config_version_outputs o
JOIN resources r ON r.id = o.resource_id
JOIN resource_config_versions rcv ON rcv.version_md5 = o.version_md5 AND rcv.resource_config_scope_id = r.resource_config_scope_id
JOIN build_ids bi ON o.build_id = bi.build_id
```

...here's the `EXPLAIN ANALYZE`:

```
Append  (cost=1535.45..54855.85 rows=3 width=113) (actual time=3.821..1212.134 rows=134774 loops=1)
  CTE build_ids
    ->  Recursive Union  (cost=8.59..694.57 rows=771 width=8) (actual time=1.654..215.981 rows=17090 loops=1)
          ->  Unique  (cost=8.59..8.59 rows=1 width=8) (actual time=1.653..1.983 rows=661 loops=1)
                ->  Sort  (cost=8.59..8.59 rows=1 width=8) (actual time=1.652..1.826 rows=661 loops=1)
                      Sort Key: i_1.build_id
                      Sort Method: quicksort  Memory: 55kB
                      ->  Index Scan using build_inputs_resource_versions_idx on build_resource_config_version_inputs i_1  (cost=0.56..8.58 rows=1 width=8) (actual time=0.026..1.472 rows=661 loops=1)
                            Index Cond: ((resource_id = 101) AND (version_md5 = '40fa80af72b9c40a7b218fd3fb5393c0'::text))
          ->  HashAggregate  (cost=66.29..67.06 rows=77 width=8) (actual time=28.771..29.864 rows=2347 loops=7)
                Group Key: bp.to_build_id
                ->  Nested Loop  (cost=0.98..66.09 rows=77 width=8) (actual time=0.472..24.682 rows=21834 loops=7)
                      Join Filter: (bi_2.build_id = i_2.build_id)
                      ->  Nested Loop  (cost=0.42..48.73 rows=14 width=24) (actual time=0.462..7.386 rows=2540 loops=7)
                            ->  WorkTable Scan on build_ids bi_2  (cost=0.00..0.20 rows=10 width=8) (actual time=0.000..0.441 rows=2441 loops=7)
                            ->  Index Only Scan using build_pipes_from_build_id_to_build_id_uniq on build_pipes bp  (cost=0.42..4.84 rows=1 width=16) (actual time=0.002..0.002 rows=1 loops=17090)
                                  Index Cond: (from_build_id = bi_2.build_id)
                                  Heap Fetches: 697
                      ->  Index Only Scan using build_resource_config_version_inputs_uniq on build_resource_config_version_inputs i_2  (cost=0.56..1.17 rows=6 width=8) (actual time=0.003..0.005 rows=9 loops=17779)
                            Index Cond: (build_id = bp.from_build_id)
                            Filter: (resource_id <> 101)
                            Rows Removed by Filter: 0
                            Heap Fetches: 13250
  ->  Nested Loop  (cost=840.88..39800.98 rows=2 width=113) (actual time=3.821..1111.347 rows=127307 loops=1)
        ->  Hash Join  (cost=840.33..4550.73 rows=4254 width=63) (actual time=3.704..432.316 rows=127835 loops=1)
              Hash Cond: (i.resource_id = r.id)
              ->  Nested Loop  (cost=0.56..3652.47 rows=4254 width=45) (actual time=1.679..379.326 rows=127835 loops=1)
                    ->  CTE Scan on build_ids bi  (cost=0.00..15.42 rows=771 width=8) (actual time=1.656..221.465 rows=17090 loops=1)
                    ->  Index Only Scan using build_resource_config_version_inputs_uniq on build_resource_config_version_inputs i  (cost=0.56..4.66 rows=6 width=45) (actual time=0.006..0.008 rows=7 loops=17090)
                          Index Cond: (build_id = bi.build_id)
                          Heap Fetches: 11169
              ->  Hash  (cost=807.12..807.12 rows=2612 width=22) (actual time=1.983..1.983 rows=2612 loops=1)
                    Buckets: 4096  Batches: 1  Memory Usage: 174kB
                    ->  Seq Scan on resources r  (cost=0.00..807.12 rows=2612 width=22) (actual time=0.011..1.488 rows=2612 loops=1)
        ->  Index Scan using resource_config_scope_id_and_version_md5_unique on resource_config_versions rcv  (cost=0.56..8.28 rows=1 width=92) (actual time=0.005..0.005 rows=1 loops=127835)
              Index Cond: ((resource_config_scope_id = r.resource_config_scope_id) AND (version_md5 = i.version_md5))
  ->  Nested Loop  (cost=1.25..14360.26 rows=1 width=113) (actual time=0.048..85.249 rows=7467 loops=1)
        ->  Nested Loop  (cost=0.70..3024.50 rows=1368 width=63) (actual time=0.032..50.585 rows=8122 loops=1)
              ->  Nested Loop  (cost=0.42..2545.01 rows=1368 width=45) (actual time=0.019..36.930 rows=8122 loops=1)
                    ->  CTE Scan on build_ids bi_1  (cost=0.00..15.42 rows=771 width=8) (actual time=0.001..3.806 rows=17090 loops=1)
                    ->  Index Only Scan using build_resource_config_version_outputs_uniq on build_resource_config_version_outputs o  (cost=0.42..3.26 rows=2 width=45) (actual time=0.002..0.002 rows=0 loops=17090)
                          Index Cond: (build_id = bi_1.build_id)
                          Heap Fetches: 185
              ->  Index Scan using resources_pkey on resources r_1  (cost=0.28..0.34 rows=1 width=22) (actual time=0.001..0.001 rows=1 loops=8122)
                    Index Cond: (id = o.resource_id)
        ->  Index Scan using resource_config_scope_id_and_version_md5_unique on resource_config_versions rcv_1  (cost=0.56..8.28 rows=1 width=92) (actual time=0.004..0.004 rows=1 loops=8122)
              Index Cond: ((resource_config_scope_id = r_1.resource_config_scope_id) AND (version_md5 = o.version_md5))
Planning time: 3.089 ms
Execution time: 1220.851 ms
```

clarafu: Hmm ya I definitely get the concern but right now I can't think of an easy way around it. 🤔  I was also thinking iof we can just combine the queries so we don't have to pass around the list of builds but I don't think we can attach fetching the resource/version/build info in the recursive query. The parent query that runs in order to do the `UNION ALL` will be super slow if it tries to union on all the resource/version/build info rather than just doing the union on the builds and then we fetch out the info after. 

As an example, I tried running this against `hush-house` with approx 200 builds and it took around 800 ms to run. https://explain.depesz.com/s/8rfJ

I think if we can think of a better way to do it that is easy enough then yes we should probably go ahead with that , but for now I think this current approach is good enough for this first iteration!

atc/db/resource.go
```diff
@@ -953,3 +956,243 @@ func requestScheduleForJobsUsingResource(tx Tx, resourceID int) error {
...
+	downStreamCausalityQuery = `
+WITH RECURSIVE build_ids AS (
+		SELECT DISTINCT i.build_id
+			FROM build_resource_config_version_inputs i
+			WHERE i.resource_id=$1 AND i.version_md5=$2
+	UNION ALL
+		SELECT DISTINCT bp.to_build_id AS build_id
+		FROM build_ids bi
+		INNER JOIN build_pipes bp ON bi.build_id = bp.from_build_id
+		INNER JOIN build_resource_config_version_inputs i ON i.build_id = bi.build_id
```
clarafu, line 970: I ran this query against `hush-house` on a resource version that had approx 200 builds, and it seemed to take around 500 ms to run https://explain.depesz.com/s/4Acp. I think a chunk of time is spent on this join onto `build_resource_config_version_inputs` in order to remove any builds that have the `resource_id` as an input. Just wanted to mention it in case we can think of anything to refactor this query in order to not need us to do that where clause!

atc/db/resource.go
```diff
@@ -953,3 +956,243 @@ func requestScheduleForJobsUsingResource(tx Tx, resourceID int) error {
...
 	return nil
 }
+
+var (
+	downStreamCausalityQuery = `
+WITH RECURSIVE build_ids AS (
+		SELECT DISTINCT i.build_id
+			FROM build_resource_config_version_inputs i
+			WHERE i.resource_id=$1 AND i.version_md5=$2
+	UNION ALL
```
clarafu, line 966: nit: maybe we can do just a UNION instead of a UNION ALL? Probably because we don't need duplicate resulting build_ids right

clarafu (commented):
Super duper cool :D

atc/db/resource.go
aoldershaw: @clarafu thoughts on just running the recursive query 2 (or 3) times, per https://github.com/concourse/concourse/pull/7125#discussion_r646657959? That last `EXPLAIN ANALYZE` is run on a resource version with 17000 builds and it doesn't get *too* slow

atc/api/resourceserver/versionserver/downstream.go
```diff
@@ -0,0 +1,56 @@
...
+		if !found {
+			logger.Info("resource-version-not-found", lager.Data{"resource-name": resourceName, "resource-config-version": versionID})
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		err = json.NewEncoder(w).Encode(causality)
+		if err != nil {
+			logger.Error("failed-to-encode", err, lager.Data{"resource-name": resourceName, "resource-config-version": versionID})
```
aoldershaw, line 50: According to the ATC logs, this resource version (https://ci.concourse-ci.org/teams/contributor/pipelines/prs/resources/concourse-pr/causality/939850057/downstream?vars.branch=%22master%22) contains a cycle, causing this encoding step to fail:

```
{"timestamp":"2021-06-08T13:42:14.522065439Z","level":"error","source":"atc","message":"atc.causality.failed-to-encode","data":{"error":"json: unsupported value: encountered a cycle via []*atc.CausalityBuild","resource-config-version":939850057,"resource-name":"concourse-pr","session":"4720233"}}
```

atc/db/resource.go
```diff
@@ -953,3 +956,243 @@ func requestScheduleForJobsUsingResource(tx Tx, resourceID int) error {
...
+			rName, versionStr string
+			version           atc.Version
+		)
+		rows.Scan(&rID, &rcvID, &rName, &versionStr, &bID)
+		err = json.Unmarshal([]byte(versionStr), &version)
+		if err != nil {
+			return err
+		}
+
+		rv, found := resourceVersions[rcvID]
```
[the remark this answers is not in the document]
chenbh: Good catch!

atc/api/resourceserver/versionserver/downstream.go
chenbh: Ah, this is caused by the `put` step not generating a new version, it just outputs the same version as the `get` step

clarafu:
Hmm do you think we should remove the `view all` button or tab for the `outputs of` (or for `inputs of` too) if it does not have any outputs/inputs for the user to view? This is because at first glance I thought the `view all` button was for the `metadata` section because it is really close to it and there was no divider between the `metadata` and `outputs of` section when there are no rows within the `outputs of` section. Honestly this is not a big deal though, just something I noticed! 

Screen Shot 2021-06-09 at 10 12 00 AM

atc/api/resourceserver/versionserver/downstream.go
chenbh: Ended up flattening the api response so it's no longer a tree. This has the side effect of allowing full cycles in the graph (in addition to the "partial" cycles for resources that bump themselves).

atc/db/resource.go
```diff
@@ -949,3 +1000,252 @@ func requestScheduleForJobsUsingResource(tx Tx, resourceID int) error {
...
+	resourceHasParent[resourceID] = true // the root resource is the only one that is not required to have a parent
+
+	buildAsChild := func(rv *atc.CausalityResourceVersion, build *atc.CausalityBuild) {
+		if !contains(build.ResourceVersionIDs, rv.ID) { // in case a build outputs the same resource multiple times
+			build.ResourceVersionIDs = append(build.ResourceVersionIDs, rv.ID)
+		}
+
+		if used, found := resourceHasParent[rv.ResourceID]; !used || !found {
+			resourceHasParent[rv.ResourceID] = false
+		}
```
aoldershaw, line 1100: Pretty sure this can just be removed - given that if a key isn't found in a `map`, the zero value is returned, the condition should be equivalent to:

```go
if !resourceHasParent[rv.ResourceID] {
	resourceHasParent[rv.ResourceID] = false
}
```

...which is redundant. We can just not store resources in the map if they don't have a parent for the same result.

web/public/causality.mjs
```diff
@@ -0,0 +1,49 @@
...
+
+export function renderCausality(dot){
+  const foundSvg = d3.select(".causality-graph");
+  const svg = createSvg(foundSvg)
+
+  graphviz.graphviz.layout(dot, "svg", "dot").then(content => {
+    svg.html(content);
+
+    svg.selectAll('title').remove();
+    svg.selectAll('*').attr('xlink:title', null);
```
aoldershaw, line 12: Could we add a comment explaining why we need this? IIRC it was to get rid of the HTML tooltips that can't be disabled with graphviz

web/elm/src/Views/TopBar.elm
```diff
@@ -194,17 +208,56 @@ jobBreadcrumb jobName =
...
+        component =
+            case direction of
+                Concourse.Downstream ->
+                    Assets.DownstreamCausalityComponent
+
+                Concourse.Upstream ->
+                    Assets.UpstreamCausalityComponent
+
+        name =
+            String.join "," <| Concourse.versionQuery version
```
aoldershaw, line 244: Wonder if we should do a similar truncation for resource version components here - otherwise, the breadcrumb can be really long and look a bit wonky
Screen Shot 2021-06-11 at 4 01 50 PM

chenbh: I feel like any non-css based solutions here would have trouble based on the size of the users device. We were able to use it in the SVG because the user can pan around, and we're only concerned with the minimum number of character to uniquely identify a resource.

Tried to solve this in https://github.com/concourse/concourse/pull/7125/commits/6786808e5d03fa53b2f8df8bbc41e0f7daf54a7f, but that only solves the problem if the very last breadcrumb overflows. Which should be fine if we assume pipeline and resource names to be somewhat short, and the only long breadcrumbs are the resource versions

aoldershaw: Maybe we could take the approach that we tried to take initially with the resource versions and wrap each key-value pair in a `span` that has a `max-width` and applies ellipsis to text if it overflows. So, the version breadcrumb may end up looking like:

```
digest:sha256:abc123foobar..., tag:branch-resource-causal...
```

Not a perfect solution as it can still wrap to multiple lines, though.

atc/db/resource.go
```diff
@@ -949,3 +993,263 @@ func requestScheduleForJobsUsingResource(tx Tx, resourceID int) error {
...
+	for _, v := range lst {
+		if e == v {
+			return true
+		}
+	}
+	return false
+}
+
+func causalityQuery(direction CausalityDirection) string {
+	switch direction {
```
aoldershaw, line 1010: I notice you deployed the updated version to CI. Performance doesn't seem too bad with the version_md5 based approach!

I haven't tried to reason about why this is happening yet (i.e. whether it's a bug with the UI or the API endpoints), but it seems like the graph is rendered in the wrong direction - e.g. when looking at the downstream graph, the resource version appears on the right of the graph with the job edges feeding into it, rather than out of it (e.g. https://ci.concourse-ci.org/teams/main/pipelines/concourse/resources/concourse/causality/999241219/downstream):

Screen Shot 2021-06-25 at 3 32 06 PM

chenbh: This, along with some other subtle bugs were fixed in https://github.com/concourse/concourse/pull/7125/commits/ff4a1a0c750471907aab693928b72630c029d5df

chenbh:
@aoldershaw deployed the latest build on CI, there is an interesting case with jobs that bump their own resources. The graph for https://ci.concourse-ci.org/teams/main/pipelines/concourse/resources/concourse/causality/1002942545/downstream shows 2 versions for the root resource `concourse` because of the `bosh-bump #533`. Because it `get` and `put` to the same resource, the previous version of `concourse-release-repo` was caused by the previous commit of `concourse`, leading it to show up multiple times.

Not sure how to address this, it might be confusing and takes a bit of time to track it down, but it does display the correct info and that could be more important to figure out edge cases like this one.

web/public/causality.mjs
```diff
@@ -0,0 +1,48 @@
...
+        ev.stopImmediatePropagation();
+      }
+    )
+    .call(zoom
+      .scaleExtent([0.5, 10])
+      .on("zoom", function() {
+        var ev = d3.event;
+        g.attr("transform", "translate(" + ev.translate + ") scale(" + ev.scale + ")");
+      })
+    );
```
aoldershaw, line 45: Minor annoyance, but if you zoom/pan on one page, then go to another causality page and try to zoom/pan, it jumps back to the previous state:

https://user-images.githubusercontent.com/26583442/123835347-6a1fc500-d8d6-11eb-8cd0-5bb2e8de7b57.mov

This can be fixed pretty easily here:

```diff
diff --git a/web/public/causality.mjs b/web/public/causality.mjs
index 87b1de232..e980dca77 100644
--- a/web/public/causality.mjs
+++ b/web/public/causality.mjs
@@ -43,6 +43,8 @@ function createSvg(svg) {
         g.attr("transform", "translate(" + ev.translate + ") scale(" + ev.scale + ")");
       })
     );
+    // Reset graph focus in case we previously modified it on another page.
+    zoom.translate([0,0]).scale(1).center(0,0);
   }
   return g;
 }
```

aoldershaw (commented):
Have one tiny suggestion about fixing the jumpiness when you zoom/pan and navigate to another causality graph, but otherwise I think it's pretty much good to go.

I don't think the "2 root versions" thing is such an issue, because as you say, it does display the correct information (albeit in a slightly confusing way). We could just prevent adding more resource versions to the root version, but not sure if that's necessarily the right approach

web/public/causality.mjs
chenbh: This should be fixed with https://github.com/concourse/concourse/commit/d453ff233f45aa8d68fb76bcd583859fb5e0c002#diff-2e1332d21e1a3253c1661d71595bca330a08e4722f22a8f9127be4101537bfccR23 and I don't see it when I do it, is this after hard reloading the page?

aoldershaw: Ah yeah i guess i had an old version 👍

aoldershaw (approved):
In that case, I'm good to approve. Merge at your own will!

chenbh:
I've decided to gate the api endpoint and the web UI behind a feature flag mainly so that a cluster with many curious users doesn't kill the database (although the better solution could've been to rate limit the endpoint). Most of it is contained in https://github.com/concourse/concourse/pull/7125/commits/b9d824ee2150d7609cc4d7a82b9254ca77676ee1

atc/api/infoserver/info.go
```diff
@@ -15,6 +15,15 @@ func (s *Server) Info(w http.ResponseWriter, r *http.Request) {
 		WorkerVersion: s.workerVersion,
 		ExternalURL:   s.externalURL,
 		ClusterName:   s.clusterName,
+		FeatureFlags: map[string]bool{
```
aoldershaw, line 18: Very sensible approach putting it in the `info` response, but I'll propose the idea that first came to my mind. What if instead we interpolate the (relevant) flags into the elm flags directly when we interpolate `index.html`:
https://github.com/concourse/concourse/blob/49b742951de35f289a5d752f6aa17a6d211f0a6a/web/public/index.html#L73-L79

If we did it this way, we wouldn't need to wait for a full round trip to the ATC to make decisions. This isn't such a big deal here (the causality buttons may take a split second to load if you load the resource page directly), but waiting for a round trip isn't ideal for features like #5802.

Anyway, not a big deal, happy to keep it how to have it - the API-based approach could be used in places like `fly` as well, so has its own benefits.

chenbh: Yeah, the main problem with this approach is that the feature flags are not known at subpage init time. I wonder if we can have some sort of pre-init that blocks until `ClusterInfo` has been fetched. But that's an idea for another time

aoldershaw: I think rather than baking more into Elm with a pre-init or something, an approach like I suggested might be better. But anyway, for now, I think the current approach is A-OK

aoldershaw (approved):
Going to re-approve but looks like there are some merge conflicts now 😬

PR #7135, opened 2021-06-07
Move garden config tests to integration suite

aoldershaw:

* [x] Remove k8s topgun tests for garden config
* [x] Add new test to integration suite to cover these tests
  * Note: it doesn't include the `CONCOURSE_GARDEN_DENY_NETWORK` test since Concourse places no special meaning to that, and is instead just focused on making sure it properly parses the config file and the flags
* [x] Fix the `group` for the `GdnBinaryFlags`, which otherwise breaks parsing the Garden config file


taylorsilva (approved):
LGTM
Another k8s/topgun gone! 🎉

PR #7144, opened 2021-06-08
Fix algorithm considering reruns as new builds

### LINKED ISSUE #6988: Jobs with both "passed" and "version:every" constraints fails to auto trigger build on a special case
## Summary
When a job is configured with `passed: [dependent-job]` and `version:every`, it fails to auto trigger if the `dependent-job` has duplicate builds i.e. two consecutive builds but with same input version.

## Steps to reproduce
Install Concourse V 7.1.0 using docker-compose file here [https://github.com/kirthiprakash/pipeline-example/blob/master/concourse_installation/docker-compose.yml](https://github.com/kirthiprakash/pipeline-example/blob/master/concourse_installation/docker-compose.yml)
Setup a two jobs with one job with `passed` constraint and `version:every` constraint
```
jobs:
  - name: stage-job
    plan:
      - get: pipeline-example-repo
        version: every
        trigger: true
      - task: display-file
        .
        .
        .
  - name: prod-job
    plan:
      - get: pipeline-example-repo
        trigger: true
        version: every
        passed: [stage-job]
      - task: display-file
```
Full configuration is available at [https://github.com/kirthiprakash/pipeline-example/blob/master/pipeline.yaml](https://github.com/kirthiprakash/pipeline-example/blob/master/pipeline.yaml)

Let the two jobs synchronise at the same input versions.

Assume the current state to be like this.
```
Resource v2 —> stage job build 2 ————> Resource v2 —> prod job build 2
Resource v1 —> stage job build 1 ————> Resource v1 —> prod job build 1
```

Now manually trigger a build on stage job from the Concourse UI which will result in
```
Resource v2 —> stage job build 3 ————> Scheduling Algorithm finds resource v2 is already built and doesn’t trigger new build
Resource v2 —> stage job build 2 ————> Resource v2 —> prod job build 2
Resource v1 —> stage job build 1 ————> Resource v1 —> prod job build 1
```
From now on, for any new resource versions, only stage job gets built and the prod job is never triggered.

Suppose there is new resource version v3.
```
Resource v3—> stage job build 4  ————> 
Resource v2 —> stage job build 3 ————> Algorithm finds resource v2 is already built and doesn’t trigger new build
Resource v2 —> stage job build 2 ————> Resource v2 —> prod job build 2
Resource v1 —> stage job build 1 ————> Resource v1 —> prod job build 1
```

The scheduling algorithm for prod job always resolves `stage-job/build3` and `resource v2`. Since `resource v2` is already built on `prod job` it never auto triggers the build. This was confirmed by enabling traces.

## Expected results

After two consecutive builds on upstream job with same inputs, the downstream job should trigger a new build when there is a new successful build on upstream job with a new input version.

## Actual results

After two consecutive builds on upstream job with same inputs, the downstream job never triggers even a single build even though there are new input versions and new successful builds on the upstream job.

## Additional context
The issue was not reproducible when `version:every` was removed from `prod job`

A note on traces: Traces really helped identified the issue, but it would be great if it had more details. For example, a span from the `Algorithm.compute` contains information about the version candidates but it was in md5 format. I did try a md5 hash of the commit id, but it wasn't matching. it took some time to figure out the input to the md5 should be `{"ref":"<commit id>"}` without any space. It would have been a lot easier if the version was displayed was from the resource type format.

## Triaging info

* Concourse version: V 7.1.0
* Browser (if applicable):
* Did this used to work? I never saw this issue on V.5.5.7

taylorsilva:

closes #6988

## Notes to reviewer

If you want to do acceptance: https://github.com/concourse/concourse/issues/6988#issuecomment-851581234

## Release Note

* Fixes pipelines getting stuck with the same inputs when a job upstream of a job with `version: every` succeeds and is rerun

### DIFF
--- a/atc/db/versions_db.go
+++ b/atc/db/versions_db.go
@@ -661,12 +661,7 @@ func (cursor BuildCursor) NewerBuilds(idCol string) sq.Sqlizer {
 			},
 		}
 	} else {
-		return sq.Or{
-			sq.Expr("COALESCE(rerun_of, "+idCol+") > ?", cursor.ID),
-
-			// include reruns of the build
-			sq.Eq{"rerun_of": cursor.ID},
-		}
+		return sq.Expr("COALESCE(rerun_of, "+idCol+") > ?", cursor.ID)
 	}
 }
 
--- a/atc/db/versions_db_test.go
+++ b/atc/db/versions_db_test.go
@@ -602,14 +602,7 @@ var _ = Describe("VersionsDB", func() {
 				}
 			})
 
-			It("returns the rerun builds, oldest to newest, followed by the newer builds, oldest to newest, followed by the cursor build and older builds, newest to oldest", func() {
-				for i := 0; i < len(rerunBuilds); i++ {
-					buildID, ok, err := paginatedBuilds.Next(ctx)
-					Expect(err).ToNot(HaveOccurred())
-					Expect(ok).To(BeTrue())
-					Expect(buildID).To(Equal(rerunBuilds[i].ID()))
-				}
-
+			It("returns the newer builds, oldest to newest, followed by the cursor build and older builds, newest to oldest", func() {
 				for i := 0; i < len(newerBuilds); i++ {
 					buildID, ok, err := paginatedBuilds.Next(ctx)
 					Expect(err).ToNot(HaveOccurred())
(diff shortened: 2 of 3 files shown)


clarafu (approved):
LGTM! Did acceptance and it fixes the bug :)

PR #7147, opened 2021-06-09
Add ability to comment on a build

### LINKED ISSUE #476: Feature Request: I would like the ability to leave a comment on a (failed) build
Our team currently documents the reasons why our builds have failed in a spreadsheet. I'd rather have this persisted on Concourse. This way, our developers can note the reason for a failed build next to the build number and other build information, and it'll be easier for us to keep track of common failures and flakes. This was the most useful thing about Jenkins, and it'd be nice to have similar functionality on Concourse.

I looked for a similar issue but wasn't able to find it.

multimac:

closes #476

This PR adds an editable comment field to builds of a job. A new "info" button is added to the header bar for a build and upon clicking it a text box will be show allowing for the build comment to be edited and saved.

Along with the UI work, there is a new API endpoint, `/api/v1/builds/:build_id/comment`, for updating the comment. This was written using the pin-comment endpoint for a resource as a guide.

I wasn't sure if there were any guidelines around sourcing icons, but found these two icons which are both licensed under Apache 2.0, the same as Concourse.

https://iconify.design/icon-sets/uil/info-circle.html
https://iconify.design/icon-sets/ic/round-save.html

Screenshot 2021-06-09 215141

I have no eye for design, so feedback is more than welcome. Some design decisions:
* Builds with a comment get a small triangle in the upper-right corner. Not sure how I feel about this, on one hand it's nice to know which builds have been commented on but also it could clutter the history bar.
* Edit / Save / Cancel buttons are placed in the header bar, separated from the new build / re-run build / abort build buttons by a small spacer element
* Cancel button for the build comment uses the same icon as the abort button. May be confusing, though it doesn't share the red background of the abort button.

**Still to do...**
* [x] Initial work
* [x] Fix weird unfocus bug
* [x] Move comments into separate `build_comments` table
* [x] Finalize design
* [x] Add tests

## Release Note

* You can now leave comments on builds. For instance, this can be used to give context to your coworkers about why a particular build failed:
  Screen Shot 2021-06-30 at 5 40 45 PM

* If a build has a comment, it is displayed with a small marker to help you quickly find builds of interest. Hovering over the build displays a portion of the comment:
  Screen Shot 2021-06-30 at 5 41 52 PM


aoldershaw (commented):
Great first pass at what I think will be a really useful feature! Just glanced at the code briefly, but left a comment on the database schema and have some thoughts on the UI design:

I like the idea of some sort of indication of which builds have a comment in the build list (currently a small triangle as you currently have) 👍 . Maybe we could even make it so that hovering over the build in the list displays the comment as a tooltip for even faster access? Not sure if that's necessarily a good idea, but just spit-balling.

I think it'd be best if the design for build comments was more similar to the design (but not necessarily the exact styles) for pinned resource comments. Here's a short video of what that looks like in case you aren't familiar:

https://user-images.githubusercontent.com/26583442/121573225-01d67580-c9f3-11eb-9c1e-d1e203c8f1b5.mov

We could apply a similar idea to what you have now - having a "comment" button in the header. For icons, we usually pull from https://materialdesignicons.com. Maybe the ones called `comment` (or `comment-*`) would be most intuitive.

The way I'm picturing this button working is similar to how it works currently, with some small differences. It'd act more like a toggle button to view/hide comments, and could behave something like the following:

* If the build does not currently have a comment...
  * On page load, the button appears as dim
  * On hover, the button appears as bright with the tooltip "add a comment"
  * On click, the button becomes bright, the comment section becomes visible, and you're directly editing the comment
  * For now, maybe the comment is pre-populated with the timestamp and author as text (though we may want to store this in a more structured way in the future)
* If the build already has a comment...
   * On page load, the button appears as bright. The comment section is visible
   * On hover, the button appears as dim with the tooltip "hide comment"
   * On click, the button becomes dim (perhaps with some additional indication that there is a comment, like a notification badge or something), and the comment section is hidden

The comment block could function in a similar way to the resource pin. When you aren't editing the comment, but the comment is visible, we display a pencil icon in the top-right corner. When we click that button, we are able to edit the text, and the pencil icon is replaced with a "save" button.

Curious to hear what you think about this! Feel free to push back on any of the ideas posed here that you don't agree with

web/elm/src/Views/Styles.elm
```diff
@@ -282,3 +288,97 @@ instanceGroupBadge backgroundColor =
     , style "align-items" "center"
     , style "justify-content" "center"
     ]
+
+
+commentBarWrapper : List (Html.Attribute msg)
+commentBarWrapper =
+    [ style "border" "medium dashed rgba(255, 255, 255, 0.5)"
+    , style "border-radius" "7px"
```
aoldershaw, line 296: Could we change these styles around a little?

1. There aren't any other cases of rounded rectangles in the UI (that I can think of), so it looks a little out-of-place - could we get rid of the `border-radius`
2. Could we also just make the border `1px solid`, rather than dashed

aoldershaw (changes requested):
Have a few suggestions, but looking great so far! Really like the new `CommentBar` module.

Thoughts on adding tooltips to the build tabs as well for a preview of the comment? Could be useful for scanning the history, e.g. so you can skip over short comments like "flake" If you're looking for something in particular. Doesn't need to be anything fancy, and we can just cut the comment off:

Screen Shot 2021-06-14 at 12 03 21 PM

Also, I noticed that when you're not authorized to see the build plan, you're still able to see the comment. I wonder if that's desirable, or if we should hide the comment from the API response if they aren't authorized.

Screen Shot 2021-06-14 at 12 07 23 PM

web/elm/src/Views/CommentBar.elm
```diff
@@ -0,0 +1,305 @@
...
+                ( { model
+                    | state =
+                        case model.state of
+                            Editing state ->
+                                Editing { state | content = content }
+
+                            state ->
+                                state
+                  }
+                , []
```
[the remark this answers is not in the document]
aoldershaw: When we click on the edit button, we also want to `Focus` on the textarea. Maybe it's easier for `CommentBar.update` to return a `List Effects.Effect` so we don't need these intermediate `CommentBar`-local `Effects` that just get mapped to the "actual" `Effects.Effect`

web/elm/src/Views/CommentBar.elm
```diff
@@ -0,0 +1,305 @@
+module Views.CommentBar exposing (Effect(..), Model, State(..), Style, ViewState, commentSetCallback, defaultStyle, getContent, setCachedContent, update, updateMap, view)
```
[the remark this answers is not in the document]
multimac: Thanks! I tried to keep it as similar to the pin comment's design, with the idea that in a follow-up PR I could update the pin comment to use this too.

There's some rough edges (see `resetState` in the latest commit), but it seems like it has been easy enough to encapsulate the logic.

web/elm/src/Views/Styles.elm
```diff
@@ -282,3 +288,97 @@ instanceGroupBadge backgroundColor =
...
+    [ style "padding" "5px 10px"
+    , style "margin" "5px 5px 7px 7px"
+    , style "outline" "none"
+    , style "border" "1px solid"
+    , style "transition" "border 200ms ease, color 200ms ease"
+    ]
+        ++ defaultFont
+
+
+commentBarSaveButton :
```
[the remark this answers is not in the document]
multimac: Whoops, nope I'd refactored this into the `commentBarTextButton` so it isn't needed anymore.

web/elm/src/Views/CommentBar.elm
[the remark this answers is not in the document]
multimac: Was already working on this after my last commit 😛

I'd originally had issues passing a `List Effects.Effect` out of `CommentBar.update`, because I couldn't figure out how to pass a lambda in to generate the correct "save" effect as it can differ depending on where the comment bar is used (eg. builds use `SetBuildComment` whereas pin comments use `SetPinComment`). I took another look at it in the latest commit and have created another type alias (the poorly named, `UpdateState`) that is passed into `CommentBar.update` with the lambda to call.

https://github.com/concourse/concourse/blob/435e3987bb10b67888a3766bad6f364816258490/web/elm/src/Views/CommentBar.elm#L43-L45

https://github.com/concourse/concourse/blob/435e3987bb10b67888a3766bad6f364816258490/web/elm/src/Views/CommentBar.elm#L319

web/elm/src/Views/CommentBar.elm
```diff
@@ -0,0 +1,305 @@
...
+           else
+            style "background-color" Colors.pinTools
+         ]
+            ++ Styles.commentBarEditButton
+        )
+
+
+loadingButton : Html Message.Message
+loadingButton =
+    Html.button
```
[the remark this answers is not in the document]
multimac: I was mainly copying how the pin comment is designed. It uses a button for the "loading" spinner, though I'm not sure why. I've updated it to just be a div and it seems to look the same.

multimac:
> Thoughts on adding tooltips to the build tabs as well for a preview of the comment? Could be useful for scanning the history, e.g. so you can skip over short comments like "flake" If you're looking for something in particular. Doesn't need to be anything fancy, and we can just cut the comment off:

👍 I've added tooltips which show the first 50 characters of the first line, and append `…` if either the first line is longer than 50 characters, or there are multiple lines.

> Also, I noticed that when you're not authorized to see the build plan, you're still able to see the comment. I wonder if that's desirable, or if we should hide the comment from the API response if they aren't authorized.

I've updated header to not show the build comment when the user isn't authorized, but it will still return the comment in the API response. Is this okay?

multimac:
Something for a future PR, but it would be neat if we could show the first line / 50-characters of the comment for the latest build of a job on the pipeline screen. I had a quick poke around the code but I'm not familiar with SVGs or D3 so I'll need to spend some time to understand it first.

web/elm/src/Views/CommentBar.elm
```diff
@@ -0,0 +1,345 @@
...
+
+            else
+                ( model, [] )
+
+        Message.EditCommentBar id content ->
+            if model.id == id then
+                case model.state of
+                    Editing state ->
+                        ( { model | state = Editing { state | content = content } }
+                        , [ SyncTextareaHeight (getTextAreaID model) ]
```
aoldershaw, line 335: One place where we're missing the `SyncTextareaHeight` calls is when we resize the window - not the end of the world, but would be nice

web/elm/src/Views/CommentBar.elm
```diff
@@ -0,0 +1,345 @@
...
+saveButton model state =
+    Html.button
+        (let
+            htmlID =
+                Message.CommentBarButton Message.Save model.id
+         in
+         [ id (toHtmlID htmlID)
+         , onMouseEnter (Message.Hover (Just htmlID))
+         , onMouseLeave (Message.Hover Nothing)
+         , onClick (Message.Click htmlID)
```
aoldershaw, line 218: Can we add `style "cursor" "pointer"`?

atc/api/present/build.go
```diff
@@ -22,6 +22,7 @@ func Build(build db.Build) atc.Build {
 		ID:                   build.ID(),
 		Name:                 build.Name(),
 		JobName:              build.JobName(),
+		Comment:              build.Comment(),
```
aoldershaw, line 25: I think it's best to exclude the comment from the API response if the viewer is authorized to get the build but isn't authorized to see any build details, rather than just hiding it from the UI.

The easiest way would be for there to be a `GET /builds/.../comment` endpoint (rather than adding the comment to the build directly), but this makes it impractical to have a tooltip for build comments in the history bar, which I think is a pretty useful feature.

One imperfect way of accomplishing this: changing the signature of `present.Build` to `(build db.Build, showComment bool)`.

Most endpoints would just set `showComment` to `false`, since we don't care about build comments for the most part. However, `buildserver/get.go`, `jobserver/get_build.go`, and `jobserver/list_builds.go` should set `showComment` to `true` *if* the following holds:

```
showComments := accessor.GetAccessor(r).IsAuthorized(job.TeamName()) || job.Public()
```

(note: for `buildserver/get.go`, we'll have for existence of the job as well - if the build isn't for a job, or the job isn't public, then `showComments` should be `false`)

It's a bit awkward sprinkling this logic throughout - maybe there's a nicer way to abstract it, but this is the first thing that came to mind.

web/elm/src/Build/Header/Header.elm
```diff
@@ -63,76 +68,128 @@ header session model =
...
-
-                            else
-                                Views.Light
-                        , backgroundColor = Concourse.BuildStatus.BuildStatusFailed
-                        }
-
-                 else if model.job /= Nothing then
+        Views.Button
+            (Just
+                { type_ = Views.ToggleComment
```
aoldershaw, line 73: If the user isn't authorized to set/read the build comment, we should hide this button

aoldershaw (changes requested):
Looking great! Think we're pretty close. Just a few things, particularly around not being authorized to view build comments. I feel like we should handle this at the API level, since if build comments are public, that may limit what you can write

web/elm/src/Build/Header/Header.elm
```diff
@@ -177,6 +270,37 @@ tooltip model session =
...
+
+                                                else
+                                                    text
+                                        in
+                                        { body = Html.text truncatedText
+                                        , attachPosition = { direction = Tooltip.Bottom, alignment = Tooltip.Start }
+                                        , arrow = Nothing
+                                        , containerAttrs = Nothing
+                                        }
+                                    )
```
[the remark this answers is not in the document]
multimac: Hmm, I thought about that, but then you lose the wrapping as the tooltip approaches the side of the window, like this...

image

We could avoid this by swapping from `alignment = Tooltip.Start` to `alignment = Tooltip.End` when we think that a tooltip would overflow the edge of the screen. I'll have a play, but can you think of any way to do this off the top of your head?

Also regarding the multiple lines, I did like the idea that only the first line of a comment would be shown which would make it kind of a heading, but happy to change if you think otherwise.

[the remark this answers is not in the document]
multimac: In the latest commit I've set it to show the first line of text, truncated by CSS, without worrying about handling when the tooltip overflows the screen. I've left that out for now unless we can think of a nice way to do it.

I had a crack by passing the max dimensions into the tooltip view code and adding logic in there to set the `max-width` and `max-height` property, then ensuring that the `div` always stays within the bounds of the screen. This might work but I wanted to wrap up the other changes first so I'll come back to it later.

atc/db/migration/migrations/1622031307_add_build_comment.up.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE builds
```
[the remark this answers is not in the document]
multimac: Good idea. I've make the `build_id` the primary key of the `build_comments` table, but this might prevent there from being multiple comments per build in the future. Would you like me to change it or is it okay for now?

atc/api/present/build.go
multimac: Instead of passing in a `showComments` and leaving it up to the caller to decide when to include comments in the build response, I thought it might be cleaner to pass in the job the build is for and an instance of `accessor.Access`, so `present.Build` could make the decision.

https://github.com/concourse/concourse/blob/00c44311db661129331621224139be9d6398f600/atc/api/present/build.go#L12

https://github.com/concourse/concourse/blob/00c44311db661129331621224139be9d6398f600/atc/api/present/build.go#L36-L46

If the caller cannot provide the job or `accessor.Access`, then the comments are not included. This way the logic for deciding when to show comments does not need to be duplicated across all the different call sites.

Also added `db/build.Job` to (try to) retrieve `Job` from ` Build`, to make the `buildserver/get.go` case simpler.

https://github.com/concourse/concourse/blob/00c44311db661129331621224139be9d6398f600/atc/db/build.go#L424-L447

multimac:
Added one last little bit of functionality in the latest commit, https://github.com/concourse/concourse/pull/7147/commits/d6621990934ab0850baf56cc792e8b6c2c37090e, showing a small triangle in the upper-right corner of a job when there is a comment on the latest build of that job. Not sure if it is a good idea or not, what do you think?

image

atc/api/present/build.go
aoldershaw: This definitely seems like a nicer solution than what I suggested 👍

web/elm/src/Build/Header/Header.elm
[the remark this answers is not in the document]
aoldershaw: Ah gotcha. To be honest I wouldn't worry too much about that as it currently impacts the rest of the tooltips in the UI. Maybe we can leave fixing tooltips overflowing over the edge of the window for a separate PR?

atc/db/migration/migrations/1622031307_add_build_comment.up.sql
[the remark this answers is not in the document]
aoldershaw: Happy to leave it as is - shouldn't be too hard to migrate if we end up wanting multiple comments per build in the future.

atc/db/build.go
```diff
@@ -447,32 +446,30 @@ func (b *build) Job() (Job, bool, error) {
...
 }
 
-func (b *build) SetComment(comment string) error {
-	var nullComment sql.NullString
-	if comment != "" {
-		nullComment = sql.NullString{String: comment, Valid: true}
-	}
+func (b *build) Comment() (string, error) {
+	var comment string
+	err := psql.Select("COALESCE(bc.comment, '')").
```
aoldershaw, line 451: I'm not 100% sure, but I'm guessing that making this query for each build will end up being more expensive than just always performing a `LEFT JOIN` in the `buildsQuery`, since with this we need to do N+1 round trips to the DB. Note that the `buildsQuery` isn't used for things like scheduling, which was my original concern around performance - but it's still a bit silly to fetch comments for builds if we would just end up ignoring them (e.g. when listing team builds).

I think the better approach conceptually would be to only perform the `LEFT JOIN` when we care about the comment, but the way our ORM is currently structured makes that a bit difficult. So, I think I'm happy just adding it into the `buildsQuery`.

atc/db/migration/migrations/1622031307_add_build_comment.up.sql
```diff
@@ -1,2 +1,7 @@
-ALTER TABLE builds
-ADD COLUMN comment text;
+CREATE TABLE build_comments (
+    build_id INTEGER PRIMARY KEY,
+    comment text
+);
+
+ALTER TABLE build_comments
+  ADD CONSTRAINT worker_artifacts_build_fkey FOREIGN KEY (build_id) REFERENCES builds(id) ON DELETE CASCADE;
```
aoldershaw, line 7:
```suggestion
  ADD CONSTRAINT build_comments_build_id_fkey FOREIGN KEY (build_id) REFERENCES builds(id) ON DELETE CASCADE;
```

web/public/graph.mjs
```diff
@@ -657,6 +659,10 @@ GraphNode.prototype.has_icon = function() {
   return typeof this.icon !== 'undefined';
 }
 
+GraphNode.prototype.has_comment = function() {
```
aoldershaw, line 662: I'm a bit mixed on adding the "comment triangle" on the pipeline graph.

To me, the pipeline graph is less concerned with why things are failing, and instead is mainly a view to show you what is failing - and if you want to know "why", you can easily drill into the builds page for the latest build.

On the flip-side, it could be useful to see at a glance what failing jobs have already been looked into (by the presence of a comment) without having to drill into the builds page.

But to be honest, I'm leaning toward not adding it to the pipeline page. Thoughts on at least leaving it out of this PR - perhaps this can be implemented further and discussed in a separate PR in the future?

aoldershaw (changes requested):
Nearly there! I like what you did with `present.Build`. Just have one more concern about fetching comments being an `N+1` query. Then, think it's just tests and fixing up merge conflicts

atc/api/buildserver/set_comment.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"encoding/json"
+	"net/http"
+
+	"code.cloudfoundry.org/lager"
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/atc/db"
+)
+
+func (s *Server) SetBuildComment(build db.Build) http.Handler {
```
aoldershaw, line 12: Could we add a few small tests for this?

atc/api/present/build.go
```diff
@@ -32,6 +33,19 @@ func Build(build db.Build) atc.Build {
...
+		showComments = showComments || job.Public()
+	}
+	if access != nil {
+		showComments = showComments || access.IsAuthorized(build.TeamName())
+	}
+
+	if showComments {
+		comment := build.Comment()
+		atcBuild.Comment = comment
+	}
```
aoldershaw, line 47: Could we add some tests for this, perhaps from the perspective of one of the endpoints that calls it. e.g. calling `GetJobBuild`, where you have access to view the build but not the comment

aoldershaw (changes requested):
Almost there! Appreciate your patience and your hard work! Just a few comments and some tests for the API would be nice

aoldershaw (approved):
Thanks for all your hard work on another awesome feature!

multimac:
No worries, thanks for the help getting it ready, tested, and merged in 🥳

Thinking I might keep trawling through the old issues to see what to work on next.

PR #7148, opened 2021-06-09
containerd: keep tasks running after `concourse worker` restarts gracefully

aoldershaw:

* Call `Process.CloseIO` after the process exits
* Properly start the garden server so that it can be gracefully stopped (currently, `gardenServer.Stop()` is a noop - see 2fa6b76aec201ec7b51ef4d9db5391bc08c962e0)
* `Setpgid` on the `containerd` process so that signals to the `concourse` process don't propagate to `containerd` (until we propagate the signal explicitly via ifrit)

## Release Note

* The containerd runtime is now more resilient to the `concourse worker` process gracefully restarting (e.g. via `monit restart`)
  * Tasks that were started prior to restart will continue to run when the worker process comes back up
  * This matches the behaviour of the Guardian runtime

### DIFF
--- a/integration/docker-compose.yml
+++ b/integration/docker-compose.yml
@@ -42,7 +42,9 @@ services:
     privileged: true
     depends_on: [web]
     ports: [7777, 7788]
-    volumes: ["../hack/keys:/concourse-keys"]
+    volumes:
+    - ..:/src
+    - ../hack/keys:/concourse-keys
     stop_signal: SIGUSR2
     environment:
       CONCOURSE_RUNTIME: containerd
--- a/integration/internal/cmdtest/cmd.go
+++ b/integration/internal/cmdtest/cmd.go
@@ -67,6 +67,14 @@ func (cmd Cmd) ExpectExit(code int) Cmd {
 	return cmd
 }
 
+func (cmd Cmd) Start(t *testing.T, args ...string) *exec.Cmd {
+	execCmd, err := cmd.TryStart(args...)
+	if err != nil {
+		t.Fatalf("Start: %v", err)
+	}
+	return execCmd
+}
+
 func (cmd Cmd) Run(t *testing.T, args ...string) {
 	err := cmd.Try(args...)
 	if err != nil {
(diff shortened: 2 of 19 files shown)


aoldershaw:
@vito alright this should be ready to go now. The issue with the integration test was that we occasionally ran the `touch /tmp/stop-waiting` *too soon* after restarting, such that the command went through successfully and terminated the `task` *before* the ATC had a chance to retry `Attach`ing to the running process. So, `fly execute` would error with `proc not running` not because the process was terminated by restarting the worker but because it exited successfully before it was able to `Attach`.

There was also an intermittent issue with the upgrade/downgrade tests that was fixed in 795dccbb72448a98445199a95798b56360600976 (this currently affects master)

vito (changes requested):
I did some manual testing and I think this causes the `git` resource type to hang, because it reads the full `stdin` buffer until EOF, and with this change it never sees EOF. On my machine I end up with a hanging `cat` process.

https://github.com/concourse/git-resource/blob/891c32f1b4b43beca81adff0890b8fea35bcbc0b/assets/in#L26

It looks like none of our tests catch this because the `mock` resource just decodes a single object and stops reading:

https://github.com/concourse/mock-resource/blob/8a49dd197f1fb99d903aa0492fe976a1b4c70a7c/cmd/in/main.go#L41

aoldershaw:
@vito good catch. Take a look at fa0a890486eb0b5b7840c08a9e0e52585a1a5ef4, which aims to solve this problem by treating processes with a TTY allocated (i.e. tasks, hijack processes) differently from processes without (i.e. resource containers).

It's a bit hacky, because it relies on resources not having a TTY attached, but seems to fix the issue. Let us know if you have a better idea of how to solve the problem

vito (approved):
Yeah at the surface level this is a bit coupled to how Concourse happens to run things at the moment, but I think it's better than nothing. You could even make the argument that when we're passing a request to something on `stdin` we wouldn't be doing it over a TTY, so it seems like a reasonable heuristic. The only reason we even run tasks with a TTY is to convince tools to use fancier colored output, since the web UI supports it, and some tools use that as a heuristic. The whole stack is pretty cludgy tbh.

aoldershaw:
Yeah - the main downside I see is that prototypes expect a request over stdin, but could also benefit from convincing tools to use fancier output and [modifying the terminal size](https://github.com/concourse/concourse/blob/b18397049baa5a3f20819fd1c83a4c31826ac27c/atc/worker/client.go#L206-L210) (given that they are intended to be used as "reusable tasks"). But who knows, maybe we'll have a better solution by then

PR #7154, opened 2021-06-10
Add teamName to concourse_steps_wait_duration* metrics

### LINKED ISSUE #6751: Metric tasks_waiting should be labeled with team name instead of team numeric ID
## Context

Concourse supports multiple teams. Each team is added with a name, and Concourse assigns to it a numeric ID that is stored in the Concourse database and is (correctly) persisted also if the associated team is removed.

The ID is auto incremented each time a team is added, so if two Concourse instances at a certain point in time have the same teams, but one of them in the past added and deleted some other teams, the corresponding team IDs will be different.

For example, assume Concourse P is in production and Concourse S is in staging.

Concourse P:
- add team `foo` (ID will be 1)
- delete team `foo`
- add team `bar` (ID will be 2)

Concourse S:
- add team `bar` (ID will be 1)

As long as the team ID is used only internally, this discrepancy has no effect. Unfortunately the team ID leaks to the outside world via Prometheus metrics.

## Problems

The metric `concourse_tasks_waiting` is not labelled by team name but by team ID. This causes two problems:

### 1. autoscaling decisions

Such metric can be used to take autoscaling decisions (spawn additional Concourse workers VMs).

Since the team ID is used, according to the “Context” section, this means that the staging and prod environment cannot share the same autoscaling rules, since the same team `bar` has ID 2 in prod and ID 1 in staging.

This causes either staging being unable to reproduce prod, or even worse, introduce a scaling bug in prod after having obtained a correctly working configuration in staging. (This is what happened to us and triggered this ticket).

### 2. display

The display of the metric on Grafana is not human understandable, since it will show teams with their numeric IDs instead than with their names. (This impacts everybody displaying these metrics).

## Proposed fix

All metrics that use the team ID should use the team name instead, starting with metric `concourse_task_waiting`.

Two approaches are possible:
1. Breaking change. Replace the team ID with the team name.
2. Non breaking change: Add the team name as an additional label. This has the drawback of increasing the cardinality of all the impacted metrics, so at least from our point of view this is not worth it.

Esysc:
## Rationale
The ID of a team is auto-increment and it generates discrepancy over time if an ID changes because of remove/add operations or related to different environments (dev/prod). The metric `councourse_steps_wait*` exposes `teamId` only that is valid for internal use. We need to have a unique identifier, it can be used for ASG escalation and dashboards.

Add field `teamName` to the metric `concourse_steps_wait*` , it's not a breaking change.
### Example
This line:
```concourse_steps_wait_duration_bucket{platform="",teamId="1",type="check",workerTags="",le="10"} 4```
 will change into:
```concourse_steps_wait_duration_bucket{platform="",teamId="1",teamName="main",type="check",workerTags="",le="10"} 4```

closes #6751

* Add new field `teamName` in label Subsystem `steps`, Name `waiting` and `wait_duration`.
* Add `TeamName` to  `ContainerSpec`.
 

## Notes for reviewers 

Thanks to @taylorsilva that pointed me in the right and clean direction.

### DIFF
--- a/atc/worker/container_spec.go
+++ b/atc/worker/container_spec.go
@@ -14,10 +14,12 @@ type WorkerSpec struct {
 	ResourceType string
 	Tags         []string
 	TeamID       int
+	TeamName     string
 }
 
 type ContainerSpec struct {
 	TeamID    int
+	TeamName  string
 	ImageSpec ImageSpec
 	Env       []string
 	Type      db.ContainerType
--- a/atc/worker/container_spec.go
+++ b/atc/worker/container_spec.go
@@ -14,7 +14,6 @@ type WorkerSpec struct {
 	ResourceType string
 	Tags         []string
 	TeamID       int
-	TeamName     string
 }
 
 type ContainerSpec struct {
--- a/atc/worker/db_worker_provider.go
+++ b/atc/worker/db_worker_provider.go
@@ -47,6 +47,10 @@ type WorkerProvider interface {
 		savedWorker db.Worker,
 		numBuildWorkers int,
 	) Worker
+
+	GetTeamByID(
+		teamID int,
+	) string
 }
 
 type dbWorkerProvider struct {
@@ -274,4 +278,18 @@ func (provider *dbWorkerProvider) NewGardenWorker(logger lager.Logger, savedWork
 		provider.dbResourceCacheFactory,
 		buildContainersCount,
 	)
+}
+
+func (provider *dbWorkerProvider) GetTeamByID(teamID int) (string) {
+	allTeams, err := provider.dbTeamFactory.GetTeams()
+	if err != nil {
+		return "Cannot get all teams"
+	}
+	teamName := "Not Found"
+	for i := range allTeams {
+		if allTeams[i].ID() == teamID {
+			teamName = allTeams[i].Name()
+		}
+	}
+	return teamName
 }
--- a/atc/worker/db_worker_provider.go
+++ b/atc/worker/db_worker_provider.go
@@ -47,10 +47,6 @@ type WorkerProvider interface {
 		savedWorker db.Worker,
 		numBuildWorkers int,
 	) Worker
-
-	GetTeamByID(
-		teamID int,
-	) string
 }
 
 type dbWorkerProvider struct {
@@ -278,18 +274,4 @@ func (provider *dbWorkerProvider) NewGardenWorker(logger lager.Logger, savedWork
 		provider.dbResourceCacheFactory,
 		buildContainersCount,
 	)
-}
-
-func (provider *dbWorkerProvider) GetTeamByID(teamID int) (string) {
-	allTeams, err := provider.dbTeamFactory.GetTeams()
-	if err != nil {
-		return "Cannot get all teams"
-	}
-	teamName := "Not Found"
-	for i := range allTeams {
-		if allTeams[i].ID() == teamID {
-			teamName = allTeams[i].Name()
-		}
-	}
-	return teamName
 }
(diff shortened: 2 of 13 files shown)

### CODE REVIEW

File: atc/worker/db_worker_provider.go [hunk 1]
```diff
@@ -275,3 +279,17 @@ func (provider *dbWorkerProvider) NewGardenWorker(logger lager.Logger, savedWork
 		buildContainersCount,
 	)
 }
+
+func (provider *dbWorkerProvider) GetTeamByID(teamID int) (string) {
```
taylorsilva, line 283: This seems like an odd place to have a function like this. I'd like to avoid turning the `dbWorkerProvider` into a catch-all place for orphaned functions, which is what this feels like it is doing.

Also, since we're making a db call with `provider.dbTeamFactory.GetTeams()` why not make a SQL query that gets us the exact data we need? (`SELECT ... FROM teams WHERE id=teamID`)
Then the `for` loop is not needed to find the team.

atc/worker/pool.go
```diff
@@ -281,6 +281,7 @@ func (pool *pool) SelectWorker(
 	labels := metric.StepsWaitingLabels{
 		Platform:   workerSpec.Platform,
 		TeamId:     strconv.Itoa(workerSpec.TeamID),
+		TeamName:   pool.provider.GetTeamByID(workerSpec.TeamID),
```
taylorsilva, line 284: When the team saw this PR during standup one concern that came up was "will this hit the db every time the metric is updated?". The db isn't being hit every time `Inc()` and `Dec()` are called, but it will get hit a lot more now (every single step that runs), even if Concourse isn't configured with any metric emitters. 

Is there a way to eliminate this db call completely? You'll notice that all the other fields on the struct are populated with information that's been passed into the function. Can we do the same for the `teamName`? Pass in the info we need?

Looking at what's being passed in already, the answer appears to be no. If we go up the call stack though, does someone else have this info?
Excluding tests, `SelectWorker()` is called in these places:
image

If we go to one of these functions, what info do we have at that level?

I found that each of these functions are part of a `Check/Task/Get/PutStep` struct which all contain a `StepMetadata` struct which contains the `TeamName`: https://github.com/concourse/concourse/blob/6841cd592dfe844599b73e333ea66c650f2f237b/atc/exec/step_metadata.go#L8-L20

Can we pass this in? Either the entire `StepMetadata` or the `TeamName`? This will avoid the need for any extra db calls.

taylorsilva (changes requested):

atc/worker/db_worker_provider.go
Esysc: This is what I did originally, I then thought it was better to use an existing function than add a new one (tried first with `GetByID`)

atc/worker/pool.go
Esysc: That's an interesting point. I agree it's not a good idea hit the db for that, especially for data that doesn't change so often over time.  I did a research in that direction, but I didn't find the origin myself, that's why I decide to hit the db. Thanks to point me to this.

atc/worker/container_spec.go
```diff
@@ -14,10 +14,12 @@ type WorkerSpec struct {
 	ResourceType string
 	Tags         []string
 	TeamID       int
+	TeamName     string
```
taylorsilva, line 17: I don't think we should add `TeamName` to `WorkerSpec`. I think adding it to `ContainerSpec` and using it from there is fine though.

I think adding `TeamName` to `WorkerSpec` will be a bit confusing later on in the worker selection part of the codebase, even though it's not being used in that area right now. `WorkerSpec` contains all the requirements that a worker needs to have in order to be used to run whichever step is executing. With that context, when I read `WorkerSpec.TeamName` I _think_ it represents a worker that belongs to a specific team, which is not what we're trying to convey here as far as usage goes.

From what I can tell there shouldn't be the same confusion with adding `TeamName` to `ContainerSpec`.

taylorsilva (changes requested):

atc/worker/container_spec.go
Esysc: This makes sense, absolutely. The fact of having already `TeamID` led me to add `TeamName` without thinking about the confusion that could generate.

taylorsilva (approved):

Issue #5215, opened 2020-02-21
Add --team to remaining fly commands


taylorsilva:
## Goal

Implement `--team` for remaining fly commands. `fly/integration` tests should be updated. Cases should be added to the `error_handling` file as well https://github.com/concourse/concourse/blob/587483424a47f3eebb2ed8e861687b60cccb9ec7/fly/integration/error_handling_test.go#L92

```bash
# use default team
fly -t ci <command> <command-options>
# use another team
fly -t ci <command> <command-options> --team other-team
```

- [ ] abort-build
- [ ] resources
- [ ] check-resource-type
- [ ] check-resource
- [ ] resource-versions
- [ ] watch
- [ ] checklist
- [ ] clear-task-cache
- [ ] execute
- [ ] expose-pipeline
- [ ] destroy-pipeline
- [ ] get-pipeline
- [ ] set-pipeline
- [ ] hide-pipeline
- [ ] pause-pipeline
- [ ] unpause-pipeline
- [ ] rename-pipeline
- [ ] order-pipelines
- [x] ~jobs~
- [x] ~trigger-job~
- [x] ~pause-job~
- [x] ~unpause-job~
- [x] ~hijack~

## Background
Follow-up story from #4196. You can look at `trigger-job` for guidance on how to implement the flag.

https://github.com/concourse/concourse/blob/587483424a47f3eebb2ed8e861687b60cccb9ec7/fly/commands/trigger_job.go#L40-L47

jroden:
Hi @taylorsilva hope you had a good weekend!  I would like to take one of these as a first contribution to concourse :smile: Picking one at random, I'll start on set-pipeline assuming its available?

taylorsilva:
@jroden `set-pipeline` is available to work on! Looking forward to your contribution :)

khng:
Hi @taylorsilva! 

Was hoping to pick up the `destroy-pipeline` command for this issue. 

Let me know if there are any issues!

taylorsilva:
All yours! Thanks for contributing :)

techgaun:
@taylorsilva is the `get-pipeline` available to pick? I would like to take that if its free. Same goes for `expose-pipeline` but would like to tackle `get-pipeline` first.

taylorsilva:
@techgaun they're both yours! Thanks for the help :)

techgaun:
[...]

PR #7162, opened 2021-06-14
 fly: add --team flag to checklist

tech-geek29:
document DCO process

Signed-off-by: rishabh.jain.cs14@gmail.com

## Why is this PR needed?
Currently the checklist command does not support teams option. This PR provides that functionality. See #5215

## How does it accomplish that?
Adds --team flag to checklist command which overrides the default configured for the target.

## Contributor Checklist
- [ ] Unit tests
- [x] Integration tests

##  Code reviewed
- [ ] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed
- [ ] New config flags added? Ensure that they are added to the BOSH
       and Helm packaging; otherwise, ignored for the integration tests (for example, if they are Garden configs that are not 
      displayed in the --help text).

### DIFF
--- a/fly/integration/checklist_test.go
+++ b/fly/integration/checklist_test.go
@@ -145,5 +145,59 @@ job-2: concourse.check %s main some-pipeline/branch:master job-2
 				})
 			})
 		})
+
+		Context("when a team name is specified alongwith pipeline name", func() {
+			JustBeforeEach(func() {
+				atcServer.AppendHandlers(
+					ghttp.CombineHandlers(
+						ghttp.VerifyRequest("GET", "/api/v1/teams/other-team"),
+						ghttp.RespondWithJSONEncoded(http.StatusOK, atc.Team{
+							Name: "other-team",
+						}),
+					),
+					ghttp.CombineHandlers(
+						ghttp.VerifyRequest("GET", "/api/v1/teams/other-team/pipelines/some-pipeline/config"),
+						ghttp.RespondWithJSONEncoded(http.StatusOK, atc.ConfigResponse{Config: config}, http.Header{atc.ConfigVersionHeader: {"42"}}),
+					),
+					ghttp.CombineHandlers(
+						ghttp.VerifyRequest("PUT", "/api/v1/teams/other-team/pipelines/some-pipeline/config"),
+						ghttp.RespondWithJSONEncoded(http.StatusOK, atc.ConfigResponse{Config: config}, http.Header{atc.ConfigVersionHeader: {"42"}}),
+					),
+				)
+			})
+
+			Context("when there are no groups", func() {
+				BeforeEach(func() {
+					config = atc.Config{
+						Jobs: atc.JobConfigs{
+							{
+								Name: "job-1",
+							},
+							{
+								Name: "job-2",
+							},
+						},
+					}
+
+				})
+
+				It("prints the config as yaml to stdout, and uses the pipeline name as header", func() {
+					flyCmd := exec.Command(flyPath, "-t", targetName, "checklist", "-p", "some-pipeline/branch:master", "--team", "other-team")
+
+					sess, err := gexec.Start(flyCmd, GinkgoWriter, GinkgoWriter)
+					Expect(err).NotTo(HaveOccurred())
+
+					<-sess.Exited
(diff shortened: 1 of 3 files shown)


tech-geek29:
@taylorsilva @izabelacg @aoldershaw , Can you please review the PR?

taylorsilva:
Hi @tech-geek29, we've assigned myself and Esteban to review it. Like I said in the issue, we assign PR's and get to them when we have time. We do allocate time everyday to review PR's but sometimes other things come up and it may take a few days for us to get to reviewing.

I'm going to ask for your patience here and stop pinging (`@'ing`) us as all you're doing is filling up our notifications and becoming a bit annoying. I understand waiting sucks and you want feedback NOW! We all do and it would be great if I could instantly give feedback, but we all have work and other things to do alongside reviewing PR's. Please give us some time and room to breath, I promise we will get to your PR.

You also pinged someone not on the team anymore (Izabela). This "shotgun" notification approach is generally not appreciated in any OSS community. It usually stresses people out, which helps no one.

As an example, I made a PR to an unrelated project two days ago: https://github.com/mrkwnzl/cyphersystem-foundryvtt/pull/134
The maintainer didn't reply until today. During the two days that I had to wait I did not ping them incessantly, I simply waited. If it ended up being a week with no update from the maintainer then pinging them and asking an open question like "any thoughts on this pr?" would probably be appropriate.

Thanks again for the PR. We'll probably be reviewing it this afternoon.

tech-geek29:
Hi @taylorsilva , Will keep this mind! I am really sorry about this!

fly/integration/checklist_test.go
```diff
@@ -141,6 +141,60 @@ some-orphaned-job: concourse.check %s main some-pipeline/branch:master some-orph
...
+							{
+								Name: "job-1",
+							},
+							{
+								Name: "job-2",
+							},
+						},
+					}
+
+				})
```
EstebanFS, line 182: I think could be a good idea to get ride of this context here because:
1) It is not clear what the test is doing at this point
2) Whether groups are part of the response has nothing to do with the team flag being used

You could just delete that context and just move the IT section outside of the context, at the same level of the JustBeforeEach function.

You could also move the `config` inside the BeforeEach to be part of the JustBeforeEach so the test can still pass.

fly/integration/checklist_test.go
```diff
@@ -141,6 +141,60 @@ some-orphaned-job: concourse.check %s main some-pipeline/branch:master some-orph
...
+
+				It("prints the config as yaml to stdout, and uses the pipeline name as header", func() {
```
EstebanFS, line 184: You should announce the effect of the team flag here, something like `prints the config as yaml to stdout, and uses the pipeline name and the specified team as header`

fly/integration/checklist_test.go
EstebanFS, line 182: I think could be a good idea to get ride of this context here because:
1) It is not clear what the test is doing at this point
2) Whether groups are part of the response has nothing to do with the team flag being used

You could just delete that context and just move the IT section outside of the context, at the same level of the JustBeforeEach function.

You could also move the `config` inside the BeforeEach to be part of the JustBeforeEach so the test can still pass.

fly/integration/checklist_test.go
EstebanFS, line 184: You should announce the effect of the team flag here, something like `prints the config as yaml to stdout, and uses the pipeline name and the specified team as header`

EstebanFS (changes requested):

fly/integration/checklist_test.go
tech-geek29: Thanks for the review! I have updated the test!

fly/integration/checklist_test.go
tech-geek29: Thanks for the review! I have updated the test!

EstebanFS (approved):
LGTM!

tech-geek29:
Hi @taylorsilva , Sorry but is there anything pending on this PR or it can be merged. Just not sure, I see it is approved but merged!

taylorsilva:
Forgot to hit the button :) it's in now, thanks for the reminder

PR #7176, opened 2021-06-15
Removing VersionedResourceTypes from get, check and put plans

clarafu:

This PR removes the need to include `VersionedResourceTypes` inside of the `get`, `check` and `put` plans. It is still necessary in the `task` plans because of the fact that the task config might not be available to the ATC during the time of the build planner due to the possibility of the task config being an external file (using `file: <file-path>`).

I've wanted to do this for a really long time because it always seemed weird how we always include the list of ALL the custom resource types in each step plan, even if the step is for a resource that does not use a custom resource type. In order to make this happen, we introduced a new field inside the step plans called `TypeImage`. This `TypeImage` struct includes a `GetPlan`, `CheckPlan`, `BaseType` and `Privileged`: 
- `GetPlan` : contains the get plan for fetching the image. Not nil when the resource step is using a custom resource type. 
- `CheckPlan`:  contains the check plan for grabbing the resource version for the image type. Not nil when the resource step is using a custom resource type.
- `BaseType`: contains the base resource type all the way down the dependency chain, so if you have a resource `foo` that has a parent custom resource type of `bar` and the `bar` resource type's parent is `git`, the `BaseType` field for the resource `foo` will contain `git` as the value.
- `Privileged` is used for indicating whether the parent resource type is privileged.

One of the reasons we needed the `VersionedResourceTypes` in the step plans was so that we could find or create resource configs or caches for the step. In order to remove the need for it, we are actually just directly using the resource cache returned back from fetching the image within the step (since the end result of why we needed the versioned resource types was to grab the resource cache of the parent). By doing this, we were able to remove the `Descriptor` layer of `ResourceConfigDescriptor` and `ResourceCacheDescriptor` (yay!) that was previously used to construct the resource config/cache using the versioned resource types. 

One other performance benefit is that the `ResourceTypes` query will no longer need to grab a version for each resource type when we fetch them from the database. So this query below is no longer there

https://github.com/concourse/concourse/blob/7fa86c1900311cd061203ee413b9c8113f8bf09c/atc/db/resource_type.go#L161-L167

That part of the query has actually been reported as one of the slowest queries for some of our users that most likely had a lot of custom resource types and versions.

**One important point to note**: There actually are behavioural changes in this PR, it is NOT a pure refactor. The behavioural change is that every time we run a step, we will actually run a check for the parent type. That check will still respect the checking interval/check every so that will ensure we don't check more times than the checking interval. This means that we will be removing the explicit checking for resource types in lidar. 

part of #6274 

- [x] Remove `container_id` from resource cache uses
- [x] Change checkfactory.ResourceTypes to return a map
- [x] Clean up/remove comments in atc.Plan
- [x] Web UI to use the new GetPlan and CheckPlan for images
- [x] Fix bug with setting `Interval` on check plans created in FetchImagePlan/ImageForType
       - Fixing this bug will result in fly execute builds to always run a check for any  custom resource types that are used as an image in the build. This is because we don't really have an easy way to propagate the default checking interval to fly for when we construct the check plan, also I think it makes sense for fly executed builds to run a check for any images used.
- [x] Fix bug with custom types acquiring checking lock. (can exit early if we check the now - lastcheck time is less than interval, and while we are waiting on the lock we can also periodically check lastcheck time so we can exit early if it hits)
- [ ] Try to consolidate merging of source defaults
- [ ] Try to refactor the check plan creations into one single place (rather than right now it can be created by db objects or the atc ResourceType object) 

## Release Note

* This PR is mainly a refactor but there is a behavioural change that comes along with it. If a resource uses a custom resource type, its `Put`, `Check`, `Get` and `Task` steps will now always create a check for its parent custom resource type. This check will still respect the resource checking interval and the `check_every` of the resource type. Because of this new feature, we no longer need to explicitly check the custom resource types in `lidar`. 


atc/db/team.go
```diff
@@ -955,33 +955,37 @@ func (t *team) FindCheckContainers(logger lager.Logger, pipelineRef atc.Pipeline
...
 	}
 
-	pipelineResourceTypes, err := pipeline.ResourceTypes()
-	if err != nil {
-		return nil, nil, err
-	}
+	// find any builds that have resource_id = resource.id
+	// Find any containers that have build_id ^
+
+	// XXX: THIS DOES NOT WORK!
```
aoldershaw, line 961: Leaving this as a reminder since it's not listed in the TODOs.

p.s. is there a reason we need to `FindOrCreateResourceConfig` here? Could we instead use the condition:
```
sq.Eq{"rccs.resource_config_id": resource.ResourceConfigID()}
```

I guess one implication of that is that we'd need to have called `PointToCheckedConfig` at least once in the check delegate (i.e. at least one check had to have completed, either successfully or unsuccessfully).

atc/engine/step_factory.go
```diff
@@ -20,6 +21,7 @@ type coreStepFactory struct {
 	resourceFactory       resource.ResourceFactory
 	teamFactory           db.TeamFactory
 	buildFactory          db.BuildFactory
+	lockFactory           lock.LockFactory
```
aoldershaw, line 24: Can we get rid of this?

atc/exec/build/variables.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+type Variables struct {
+	parentScope vars.Variables
+	localVars   vars.StaticVariables
+	tracker     *vars.Tracker
+	lock        sync.RWMutex
+}
+
+func NewVariables(tracker *vars.Tracker) *Variables {
```
aoldershaw, line 16: Was this part of the get var work? Looks like it's unused

atc/public_plan.go
```diff
@@ -14,6 +18,7 @@ func (plan Plan) Public() *json.RawMessage {
 		Check          *json.RawMessage `json:"check,omitempty"`
 		Task           *json.RawMessage `json:"task,omitempty"`
 		SetPipeline    *json.RawMessage `json:"set_pipeline,omitempty"`
+		GetVar         *json.RawMessage `json:"get_var,omitempty"`
```
aoldershaw, line 21: We can get rid of this for now

foo
```diff
@@ -0,0 +1,15 @@
+task:
```
aoldershaw, line 1: Can delete this

aoldershaw (changes requested):
Just did a quick skim, have some minor comments so far.

Also - I think we'll need some UI changes, as this seems to break the rendering of "image fetching" for check/get/put plans. Since we no longer emit the check/get plans in the build events for these types, we never end up rendering the cog icon. We'll need to decode the fetch image plans here: https://github.com/concourse/concourse/blob/d23db1bb0acb440238d8c8cd41ac81edb24aba0e/web/elm/src/Concourse.elm#L682-L687

atc/engine/build_step_delegate.go
```diff
@@ -225,139 +211,71 @@ func (delegate *buildStepDelegate) Errored(logger lager.Logger, message string)
...
-		return worker.ImageSpec{}, fmt.Errorf("image fetching failed")
+		return worker.ImageSpec{}, nil, fmt.Errorf("image fetching failed")
 	}
 
-	var cache db.UsedResourceCache
-	if !fetchState.Result(getID, &cache) {
-		return worker.ImageSpec{}, fmt.Errorf("get did not return a cache")
+	var result exec.GetResult
+	if !fetchState.Result(getPlan.ID, &result) {
+		return worker.ImageSpec{}, nil, fmt.Errorf("get did not return a result")
```
aoldershaw, line 249: Haven't looked into it too much, but I seem to be hitting this case pretty frequently
Screen Shot 2021-06-22 at 11 33 07 AM

clarafu:
Oh ya I never did the web UI portion of this 🤭  Will take a look at that now haha

atc/db/resource.go
```diff
@@ -255,22 +251,35 @@ func (r *resource) setResourceConfigScopeInTransaction(tx Tx, scope ResourceConf
...
+
+		Resource: r.name,
+	})
+
+	plan.Check.TypeImage = imagePlanner.ImageForType(plan.ID, r.type_, r.config.Tags, skipInterval && skipIntervalRecursively)
+	return plan
 }
 
+// CreateBuild creates a check build for this resource. It returns back the
+// build that was created, whether or not it was successful and any errors.
```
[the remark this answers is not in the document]
clarafu: Great suggestion!! :D

aoldershaw:
@clarafu just a reminder about the testflight failures - it seems like the issue is with the tests themselves rather than this PR breaking anything, but I'm not 100% sure

clarafu:
Hm I think the last testflight failures are caused by us no longer printing the `checking <resource> in build <number>` for each recursive resource/resource-type that we check because we no longer recursively check through fly. This is what you would used to see in fly

```
checking test/mock-resource-parent in build 5
selected worker: 0ed52d516c6f
succeeded

checking test/mock-resource-child in build 6
INFO: found resource cache from local cache

selected worker: 0ed52d516c6f
succeeded

checking test/mock-resource-grandchild in build 7
INFO: found resource cache from local cache

INFO: found resource cache from local cache

selected worker: 0ed52d516c6f
succeeded

checking test/recursive-custom-resource in build 8
INFO: found resource cache from local cache

INFO: found resource cache from local cache

INFO: found resource cache from local cache

selected worker: 0ed52d516c6f
succeeded
```

And now it is 

```
checking tf-pipeline-1-a819454e-e2a1-44be-65ed-cd1643f87e98/recursive-custom-resource in build 192
selected worker: 8f0763df7c8c
INFO: found resource cache from local cache

selected worker: 8f0763df7c8c
INFO: found resource cache from local cache

selected worker: 8f0763df7c8c
INFO[0000] fetching version: child-version              
INFO[0000] mirroring self image                          via_params=false
selected worker: 8f0763df7c8c
INFO: found resource cache from local cache

INFO: found resource cache from local cache

INFO: found resource cache from local cache

selected worker: 8f0763df7c8c
INFO[0000] fetching version: grandchild-version         
INFO[0000] mirroring self image                          via_params=false
selected worker: 8f0763df7c8c
succeeded
```

My dilemma is that we don't want to just print out the names of the resources in the check step output and since we no longer use fly to recursively check, there won't be the messages of what exactly it is currently checking in the fly output.

aoldershaw:
@clarafu An imperfect solution could just be to count the number of `selected worker` events we receive

clarafu:
It's weird, do the `selected worker` messages ever get lost? Sometimes when I run the tests it only prints a few of the `selected worker` messages but when I run the fly check resource command myself I always get 4 messages.

Also theres a test with resource type defaults that is failing so I'm looking into that. 🤔

aoldershaw:
It could be that a worker is only selected when a container actually runs, so if it's within the check interval it doesn't?

clarafu:
Don't we always run the check step if it is manually triggered?

clarafu:
So weird too, the defaults test is failing because it does not print out `"defaulted"` as a log for the resource type default 

```
checking tf-pipeline-1-e6a8c14d-6bfe-49b8-7390-f9af3257dfb8/some-resource in build 87
selected worker: 10b09931791e
INFO: found resource cache from local cache

succeeded
```

But when I run the check resource manually using the same pipeline, the log does show up. It's like the output gets cut out from the testflight tests..

```
checking test2/some-resource in build 98
selected worker: 10b09931791e
INFO: found resource cache from local cache

selected worker: 10b09931791e
INFO[0000] defaulted
succeeded
```

clarafu:
I've been trying to reproduce the testflight failure on my local machine but I've only got it to happen once after I repulled latest docker `concourse/dev` image and also added the two extra docker compose containerd override flags. But after it happened that one time it has continued to work properly after that, making me think it had nothing to do with those specific things that I changed and maybe the state of the workers/web when I was running the check resource commands? I was building docker compose and running the commands very shortly after it was up.

clarafu:
Ok I have a theory, I think this might be caused by 

https://github.com/concourse/concourse/blob/c867790fb5b924930ccb6d6f624a94846e414255/atc/engine/check_delegate.go#L148

where we avoid running a check if the build creation time is before the last check start time. I printed out the values when the test fails and this is what I get `shouldRun: false , lastCheck succeeded: true , create time after start boolean:  false , build create time:  2021-07-15 14:50:36.323729 +0000 +0000 , last check start time:  2021-07-15 14:50:37.155397 +0000 +0000`. I'm guessing this is only a problem now because now we run all the recursive parent resource type checks within the same build as the check for the resource, so the last check start time will always be later than the build creation time for any parent resource types. Rather than before we would create separate builds for each check we run for the recursive parent resource types.

aoldershaw:
@clarafu interesting, good find! I don't think I fully understand why this would occur, though. Why would the last check for the resource type's scope be after the build was created if we're in the process of determining whether we should run the check for the resource type?

Is this because other testflight tests are running with a `mock` resource with the exact same `resource_config` (and thus share a `resource_config_scope`)? i.e. there are other resources/resource types being checked concurrently with the same resource config as https://github.com/concourse/concourse/blob/fa6ec1a1b6bde3c25689abfcc1bceaeffdd1061f/testflight/fixtures/resource-type-defaults.yml#L3-L6

Or is something else going on?

clarafu:
Ya I actually thinnk I'm thinking about this wrong lol I think its the resource check that doesn't run.. right? Because that should be the second check that runs and the one that logs the `defaulted` output. I'm delirious!

aoldershaw:
Ohhhh interesting. When we unpause a pipeline, we kick off lidar immediately, so it's not that surprising that there would be an "interval" check for the resource right after the manual check starts.

I wonder why this wasn't a problem before...could it just be that because we now have to check the parent type, it just takes longer from after we create the check to actually run it?

clarafu:
Ohh yeah!! Ur probably right, we used to error when the resource type did not have a version.

aoldershaw:
Yeah, maybe we just get rid of the explicit fly check-resource in the test

clarafu:
Hm I see, we would need to get the build output from the last check then? Through `fly builds` then `fly watch`?

aoldershaw:
Oh i just mean don't test it for the check at all. We have unit tests for it, and if it works for the other step types than it's probably fine 😁  don't think it's worth blocking this PR, at least.

clarafu:
Ohh i see okay ill just do that then heh

atc/db/resource_type.go
```diff
@@ -48,16 +48,15 @@ type ResourceType interface {
...
 	CreateBuild(context.Context, bool, atc.Plan) (Build, bool, error)
 
-	Version() atc.Version
-
 	Reload() (bool, error)
 }
 
 type ResourceTypes []ResourceType
 
+// XXX: TODO Can remove the bool
```
aoldershaw, line 59: Meh, not sure if it's necessary. Returning the `bool` at least forces callers to deal with the fact that there may be no parent - it's possible to forget to check `!= nil`

aoldershaw (changes requested):

atc/lidar/scanner.go
```diff
@@ -107,7 +90,15 @@ func (s *scanner) check(ctx context.Context, checkable db.Checkable, resourceTyp
...
 
-	_, created, err := s.checkFactory.TryCreateCheck(lagerctx.NewContext(spanCtx, logger), checkable, resourceTypes, version, false)
+	// We don't want to pass in a nil list of resource types here if the pipeline
+	// does not have any resource types because we will be calling methods off of
+	// the db.ResourceTypes interface in the TryCreateCheck method
+	var resourceTypes db.ResourceTypes
+	if rts, found := resourceTypesMap[checkable.PipelineID()]; found {
+		resourceTypes = rts
+	}
+
```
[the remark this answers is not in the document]
clarafu: Oh sorry I meant a nil `db.ResourceTypes` rather than passing a `nil` to the `TryCreateCheck` method. Does that make sense? D:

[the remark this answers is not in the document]
aoldershaw: Still a bit confused tbh. `db.ResourceTypes` is just a `[]db.ResourceType` (not an interface), so `nil` should be fine, right? I don't see how this code is different from just:

```go
resourceTypes := resourceTypesMap[checkable.PipelineID()]
```

...since if it's not found, it'll just take the zero-value (a `nil` slice) anyway

aoldershaw (commented):
Leaving this review to say the last commit looks good! Happy to merge after we clean up the leftover stuff I mentioned in my last review

evanchaoli:
@clarafu @aoldershaw This PR looks like a "merge disaster" to #7208. If this one is merged first, then I would have to spend a lot of time understanding code changes of this PR. So can we shamelessly request to merge #7208 first?

aoldershaw:
@evanchaoli yeah, it'll be a pain resolving all the conflicts, but that'll be the case regardless of which gets merged first. One benefit to getting this merged first is that it'll simplify #7208 somewhat since we no longer have to worry about periodic checks for resource types, so the "in memory build" type only has to deal with resources.

That said, I get your concern about needing to understand all of these changes, too. If we do end up merging this PR first, I'm more than happy to handle the merge conflicts, and can send you a patch of the merge commit that you can apply to your branch.

Does that sound reasonable to you?

atc/engine/check_delegate.go
```diff
@@ -107,6 +121,29 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
 	var lock lock.Lock = lock.NoopLock{}
 	if d.plan.IsPeriodic() {
 		for {
+			lastCheck, err := scope.LastCheck()
```
clarafu, line 124: Please double check my logic here 😟  @aoldershaw

atc/engine/check_delegate.go
```diff
@@ -107,6 +121,29 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
...
+			if err != nil {
+				return nil, false, err
+			}
+
+			if d.build.IsManuallyTriggered() { // if the check was manually triggered
```
clarafu, line 129: Please double check my logic here 😟  @aoldershaw, I changed the if statements from being setting `shouldRun` to true, to returning false. 

This is what I converted it from:
https://github.com/concourse/concourse/blob/b047d257f2531c119e15d627e3799d9d931bad03/atc/engine/check_delegate.go#L136-L152

atc/engine/check_delegate_test.go
```diff
@@ -319,6 +329,29 @@ var _ = Describe("CheckDelegate", func() {
 						Expect(run).To(BeTrue())
 					})
 				})
+
+				Context("when the last check gets updated after the first loop", func() {
```
aoldershaw, line 333: What is this intended to test? I was expecting it to test that if it can't acquire the lock initially, it'll keep re-checking the `LastCheck` periodically, but it's really just testing the same thing as `when the interval has not elapsed since the last check`

atc/api/resourceserver/check.go
```diff
@@ -55,6 +55,7 @@ func (s *Server) CheckResource(dbPipeline db.Pipeline) http.Handler {
 			dbResourceTypes,
 			reqBody.From,
 			true,
+			!reqBody.Shallow,
```
evanchaoli, line 58: "Shallow" means check the resource itself. In that case, we suppose to not check resource types, then skipIntervalRecursively should to true. Not sure if my understanding is correct or not.

atc/builds/tracker.go
```diff
@@ -6,14 +6,27 @@ import (
...
 	"github.com/concourse/concourse/atc/db"
-	"github.com/concourse/concourse/atc/engine"
 	"github.com/concourse/concourse/atc/metric"
 	"github.com/concourse/concourse/atc/util"
 )
 
+//go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
+
+//go:generate counterfeiter . Engine
+type Engine interface {
```
evanchaoli, line 16: Looks like redefine `Engine` here just for defining a no-return `Runnable`. I feel not worthy doing so. If we want to silently ignore error returned from a run, `engine` may just not checking return value of `Run`. Adding a duplicate `Engine` and `Runnable` adds some confusion.

atc/config.go
```diff
@@ -296,6 +296,104 @@ func (types ResourceTypes) Without(name string) ResourceTypes {
...
+		// GetPlan for fetching the custom type's image and CheckPlan
+		// for checking the version of the custom type.
+		GetPlan:   &getPlan,
+		CheckPlan: checkPlan,
+	}
+}
+
+func FetchImagePlan(planID PlanID, image ImageResource, resourceTypes ResourceTypes, stepTags Tags, skipInterval bool, checkEvery *CheckEvery) (Plan, *Plan) {
+	// If resource type is a custom type, recurse in order to resolve nested resource types
+	getPlanID := planID + "/image-get"
```
evanchaoli, line 340: If there are nested resource types, plan id could be `some-id/image-get/image-get/image-check/...`. To make it clearer, can we change to `some-id/image-get/<type name>/image-get/<other type name>/...`.

Not sure how plan id is used. If the change would lead to complex changes in UI, please just ignore this comment.

atc/config.go
```diff
@@ -296,6 +296,104 @@ func (types ResourceTypes) Without(name string) ResourceTypes {
...
+			Tags: tags,
+		},
+	}
+
+	var maybeCheckPlan *Plan
+	if image.Version == nil {
+		checkPlanID := planID + "/image-check"
+		// don't know the version, need to do a Check before the Get
+		interval := DefaultCheckInterval
+		if checkEvery != nil && !checkEvery.Never {
```
evanchaoli, line 367: If `checkEvery.Never` is true, should that be a logic error? In that case, should we fail the build?

atc/lidar/scanner.go
```diff
@@ -70,26 +72,7 @@ func (s *scanner) scanResources(ctx context.Context, resources []db.Resource, re
...
-			}()
-			defer waitGroup.Done()
-			s.check(ctx, resourceType, resourceTypes)
-		}(resourceType, resourceTypes)
-	}
-	waitGroup.Wait()
-}
-
-func (s *scanner) check(ctx context.Context, checkable db.Checkable, resourceTypes db.ResourceTypes) {
+func (s *scanner) check(ctx context.Context, checkable db.Checkable, resourceTypesMap map[int]db.ResourceTypes) {
```
evanchaoli, line 75: I agree with @aoldershaw 's comment, empty array is just nil. So here we just just pass in `resourceTypes db.ResourceType` rather than a map.

atc/db/check_factory.go
```diff
@@ -30,15 +30,15 @@ type Checkable interface {
...
 	CreateBuild(context.Context, bool, atc.Plan) (Build, bool, error)
 }
 
 //counterfeiter:generate . CheckFactory
 type CheckFactory interface {
-	TryCreateCheck(context.Context, Checkable, ResourceTypes, atc.Version, bool) (Build, bool, error)
+	TryCreateCheck(context.Context, Checkable, ResourceTypes, atc.Version, bool, bool) (Build, bool, error)
 	Resources() ([]Resource, error)
-	ResourceTypes() ([]ResourceType, error)
+	ResourceTypes() (map[int]ResourceTypes, error)
```
evanchaoli, line 41: Maybe rename to `ResourceTypesByPipeline`.

atc/db/check_factory.go
```diff
@@ -102,23 +84,30 @@ func (c *checkFactory) TryCreateCheck(ctx context.Context, checkable Checkable,
...
+	if !skipInterval && time.Now().Before(checkable.LastCheckEndTime().Add(interval)) {
 		// skip creating the check if its interval hasn't elapsed yet
 		return nil, false, nil
 	}
 
-	checkPlan := checkable.CheckPlan(from, interval, resourceTypes.Filter(checkable), sourceDefaults)
-
-	plan := c.planFactory.NewPlan(checkPlan)
+	deserializedResourceTypes := resourceTypes.Filter(checkable).Deserialize()
+	if deserializedResourceTypes == nil {
```
evanchaoli, line 102: This check seems not needed. An empty array is equivalent to nil.

atc/db/resource.go
```diff
@@ -296,22 +296,34 @@ func (r *resource) setResourceConfigScopeInTransaction(tx Tx, scope ResourceConf
...
 }
 
-func (r *resource) CheckPlan(from atc.Version, interval time.Duration, resourceTypes ResourceTypes, sourceDefaults atc.Source) atc.CheckPlan {
-	return atc.CheckPlan{
-		Name:    r.Name(),
-		Type:    r.Type(),
-		Source:  sourceDefaults.Merge(r.Source()),
-		Tags:    r.Tags(),
-		Timeout: r.CheckTimeout(),
+type ImagePlanner interface {
```
evanchaoli, line 299: I doubt if this is a right place to define this interface. Maybe move to a separate file?

atc/db/resource_type.go
```diff
@@ -48,11 +48,9 @@ type ResourceType interface {
 
 	SetResourceConfigScope(ResourceConfigScope) error
 
-	CheckPlan(atc.Version, time.Duration, ResourceTypes, atc.Source) atc.CheckPlan
+	CheckPlan(planFactory atc.PlanFactory, imagePlanner ImagePlanner, from atc.Version, interval time.Duration, sourceDefaults atc.Source, manuallyTriggered bool, shallow bool) atc.Plan
```
evanchaoli, line 51: Should the last parameter `skipIntervalRecursively`?

atc/db/resource_type.go
```diff
@@ -83,8 +81,8 @@ func (resourceTypes ResourceTypes) Filter(checkable Checkable) ResourceTypes {
 	}
 }
 
-func (resourceTypes ResourceTypes) Deserialize() atc.VersionedResourceTypes {
-	var versionedResourceTypes atc.VersionedResourceTypes
+func (resourceTypes ResourceTypes) Deserialize() atc.ResourceTypes {
+	var atcResourceTypes atc.ResourceTypes
```
evanchaoli, line 85: If we init it with an empty array, then caller will not need to check nil.

atc/engine/check_delegate.go
evanchaoli: Why move acquire lock to be after interval check? I'm not sure if the logic change is correct without a deep thinking, but if this optimization is not related to this PR, I would suggest divide this change to a separate PR, so that people can put more attention on it.

evanchaoli:
@clarafu @aoldershaw I went through diffs of this PR, and tested it locally. Now I agree this is a good refactor, making code neater and logic cleaner. Now I see there should be not much conflicts to my PR, I can handle the merge easily. Also, I left a few comments, hope that helps.

atc/engine/check_delegate.go
clarafu: This actually is related because it fixes a bug that this PR introduced, I wrote a summary of it in the commit that adds it here https://github.com/concourse/concourse/pull/7176/commits/28982bddba3b1433ce4e345eecc0a56d7dda6aa0. Let me know if you have any more questions!

atc/engine/check_delegate_test.go
clarafu: Yaaaa right, it is exactly like that test. I don't know what I was thinking when I added this test haha 😆  

I think maybe I wanted to test that the last check value gets rechecked everytime it loops (but ofc I didn't write the test properly).

atc/builds/tracker.go
aoldershaw: At one point, I think it may have been an import cycle - but even with no import cycle, the [Go authors recommend](https://github.com/golang/go/wiki/CodeReviewComments#interfaces) defining interfaces where they're used, rather than where they're implemented:

> Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values. The implementing package should return concrete (usually pointer or struct) types: that way, new methods can be added to implementations without requiring extensive refactoring.

As you've no doubt noticed, we do a bad job following that advice, and honestly it's caused some pain 😁  but this tends to be a more sane way of defining dependencies.

atc/config.go
aoldershaw: The plan ID doesn't impact too much, so wouldn't be hard to change - that said, it's rarely surfaced to the user (the only place it appears is in the URL when you highlight lines on the build page)

atc/config.go
aoldershaw: `check_every: never` just means "don't periodically check this resource/type", not "never check this resource/type" - so I don't think this needs to result in a failed build.

But that's a good point - if you're setting `check_every: never` on a resource type, it'll now still check it every minute since it's no longer being scheduled via lidar, and every resource check with a custom type has an associated resource type check (that may or may not actually run a container).

Maybe we should just set a really long (practically infinite) interval? We still need to queue up a `check` step, but it shouldn't do anything.

atc/builds/tracker.go
clarafu: I actually would argue that it makes more sense to have it in the build tracker, this is because I found that the best practice is to create the interfaces where it is used rather where it is implemented. This makes it so that we can have more small and condensed interfaces while the implementation of the objects that conform to the interface be less tied to that interface. The interface should be directly tied to only what that specific component needs out of the object.

Throughout our codebase, we have interfaces really closely tied to the objects that implement them (really good example is the db package). But the only reason why we do that is to be able to test easily using fakes. I think if we can avoid tying the interface with the object, that would be best. What do you think about this?

atc/config.go
clarafu: I'm not sure if it is necessary to add the type name into the plan_id though because it is just used as a unique identifier. I don't know who would be needing to decode the plan ids, as the plan id of steps in the build plan are just incremented numbers that don't have anything to do with the actual step itself. Did you have anything in mind for using the plan ids?

atc/config.go
evanchaoli: I think here, `interval` is only used to determine if a new run should run (reaching to check interval). So as Aidan suggested, we can set a really long interval for `never`, so that, as long as a resource type has a version, it will never run auto check again.

atc/builds/tracker.go
evanchaoli: Yeah, make sense.

atc/config.go
clarafu: I do think that is a good idea, though I implemented sort of a rough draft first in order to get your opinion with the `Interval` field on the check plan be a struct that consists of both an `Interval time.Duration` and a `Never bool`, which basically is what the `CheckEvery` struct is when used in the pipeline config. I sort of liked having the `never` value be explicit but if you think the long interval approach would be better or theres a bug in the explicit never bool approach, I'm happy to change it!

atc/lidar/scanner.go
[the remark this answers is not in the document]
clarafu: Ohh my bad I thought that if you don't find a key in a map it will return a `nil` and not a nil slice. But I'm wrong! Looks like if you don't find a key in a map, it just returns the default nil value which is a nil slice of that type. I just wanted to make sure that it was never a real `nil` value being passed into `TryCreateCheck` because the `db.ResourceTypes` is being used as an interface in that method. Thanks for pointing that out haha

atc/db/resource_type.go
clarafu: I'm not exactly sure what you mean here, I don't think we have anywhere that requires us to check if the resource types is nil or not?

aoldershaw (approved):

PR #7184, opened 2021-06-17
Reduce number of `JOIN`s in builds query

aoldershaw:

* [x] Don't `JOIN` on `resource_types` or `prototypes` when `SELECT`ing a build
  * The reasoning is explained more thoroughly in add3a0b, but tl;dr, we were exceeding the default `join_collapse_limit` (of 8 relations) - if you exceed this limit, then Postgres refuses to optimize the order in which it performs the `JOIN`s, so you end up with a (potentially very) suboptimal execution plan
    * When we deployed to prod, the normal "get a check build" query that normally takes <100ms was taking >5 minutes 😬 
  * We previously performed the `JOIN`s to label builds as for a particular resource type or prototype, by name. Now, builds specifically for a resource type or prototype will appear as one-off builds (in tracing, logging, and the syslog drainer)
  * With #7176, we don't generate check builds for `resource_types` (or `prototypes`), so this will only apply to `fly check-resource-type`/`-prototype`
* [x] Add a wrapper `db.Conn` to fail our test suite if we introduce a query with "too many" `JOIN`s

### DIFF
--- a/atc/postgresrunner/join_limit_conn.go
+++ b/atc/postgresrunner/join_limit_conn.go
@@ -1,0 +1,96 @@
+package postgresrunner
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"strings"
+
+	"github.com/Masterminds/squirrel"
+	"github.com/concourse/concourse/atc/db"
+	"github.com/onsi/ginkgo"
+)
+
+func validateJoinLimit(query string) error {
+	// Each subquery, split by a UNION or a UNION ALL, can have up to 8 JOINs
+	subQueries := strings.Split(query, "UNION")
+	for _, subQuery := range subQueries {
+		numJoins := strings.Count(strings.ToUpper(subQuery), "JOIN")
+		if numJoins > 8 {
+			errMsg := "the following query has too many JOINs\n\n" + subQuery
+			if len(subQueries) > 1 {
+				errMsg += "\n\nit is part of this full query:\n\n" + query
+			}
+			errMsg += fmt.Sprintf("\n\njoin_collapse_limit defaults to 8 JOINs, while this query has %d. exceeding the limit can result in really slow queries", numJoins)
+
+			return fmt.Errorf(errMsg)
+		}
+	}
+	return nil
+}
+
+type joinLimitValidatorConn struct {
+	db.Conn
+}
+
+func (c joinLimitValidatorConn) Begin() (db.Tx, error) {
+	tx, err := c.Conn.Begin()
+	if err != nil {
+		return nil, err
+	}
+
+	return joinLimitValidatorTx{Tx: tx}, nil
+}
+
+func (c joinLimitValidatorConn) Query(query string, args ...interface{}) (*sql.Rows, error) {
+	if err := validateJoinLimit(query); err != nil {
+		ginkgo.Fail(err.Error())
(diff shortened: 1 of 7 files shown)


atc/postgresrunner/join_limit_conn.go
```diff
@@ -0,0 +1,96 @@
...
+	"strings"
+
+	"github.com/Masterminds/squirrel"
+	"github.com/concourse/concourse/atc/db"
+	"github.com/onsi/ginkgo"
+)
+
+func validateJoinLimit(query string) error {
+	// Each subquery, split by a UNION or a UNION ALL, can have up to 8 JOINs
+	subQueries := strings.Split(query, "UNION")
```
chenbh, line 16: nit: should we uppercase the query here incase we use lowercase `union`?

atc/postgresrunner/join_limit_conn.go
```diff
@@ -0,0 +1,96 @@
...
+	"fmt"
+	"strings"
+
+	"github.com/Masterminds/squirrel"
+	"github.com/concourse/concourse/atc/db"
+	"github.com/onsi/ginkgo"
+)
+
+func validateJoinLimit(query string) error {
+	// Each subquery, split by a UNION or a UNION ALL, can have up to 8 JOINs
```
chenbh, line 15: Should we split out CTEs as well? I'm assuming since they're executed independently of the main query, they shouldn't contribute to the limit
```
WITH foo AS (
SELECT ... JOIN ... JOIN ... JOIN ... JOIN ...JOIN
),
SELECT ... JOIN ... JOIN ... JOIN ... JOIN
```

aoldershaw: Added cd1c046f70b5f741db93e75bbd60b8acc4f48bef that addresses CTEs and any subqueries (I don't *think* subqueries count toward the `JOIN` limit...). Happy to revert it as it's probably too complex for the benefit it adds - I was mostly just having fun with it.

Honestly, keeping the naïve `UNION` only splitting maybe isn't such a big deal, since we're at least guaranteed there's no false negatives.

chenbh:
Do you know why it defaults to 8? While I think it's fine to get rid of the 2 joins for this particular query, it might be interesting to see how much performance hit we take if we bump the `join_collapse_limit` (and `from_collapse_limit` as well) up instead.

aoldershaw:
@chenbh we considered just bumping the `join_collapse_limit`, but decided against it for now just because there was this relatively low-hanging fruit.

The higher the `join_collapse_limit`, the longer planning will take. At least theoretically, if the limit is `n`, Postgres considers up to `n!` orderings, which can grow pretty quickly. That said, I'm guessing it doesn't actually need to consider all `n!` possibilities.

That said, if we continue to run into this problem (which hopefully we'll be able to detect now), we may consider bumping the limit.

chenbh (approved):
logic lgtm, I'll defer to @clarafu in case she has any db optimizations/tricks to help with these types of scenarios

aoldershaw:
@chenbh going to merge this for now since this is a blocker for deploying to CI. If you get the chance, can you rebase causality on to this? Curious to see how performant the `version_md5`-based algorithm is in practice.

chenbh:
done!

PR #7211, opened 2021-06-24
Treat resource types as checkable in rate limiter

taylorsilva:

Follow-up from #7102 
Will be made irrelevant by #7176

check builds are created for resources and custom resource types. The
resource types were not being considered in the rate limit, so as soon
as you have any custom resource types there would always be check builds
that would get stuck by the rate limiter.

### DIFF
--- a/atc/db/resource_check_rate_limiter.go
+++ b/atc/db/resource_check_rate_limiter.go
@@ -3,9 +3,10 @@ package db
 import (
 	"context"
 	"fmt"
-	sq "github.com/Masterminds/squirrel"
 	"sync"
 	"time"
+
+	sq "github.com/Masterminds/squirrel"
 
 	"code.cloudfoundry.org/clock"
 	"golang.org/x/time/rate"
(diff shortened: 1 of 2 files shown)


aoldershaw (approved):

aoldershaw:
Probably don't need the release note as the bug didn't exist in 7.3

aoldershaw:
Just based on the duration of testflight, it doesn't look like this totally fixes the issue - notice how before #7102 was merged, testflight took ~20mins (https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/testflight/builds/822), and after, it took ~40mins (or timed out) (https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/testflight/builds/824).

This testflight run is currently at 40mins and counting

taylorsilva:
Just noticed the timing too. Seems like we still got a mystery to solve 🤔

aoldershaw:
The following test took ~15m:

https://github.com/concourse/concourse/blob/58526cc97a5b8df2c5273bc37b608ef161ee7823/testflight/resource_config_versions_test.go#L24

I've seen the same thing without this fix:
https://ci.concourse-ci.org/teams/main/pipelines/branch/jobs/testflight/builds/17.2?vars.branch=%22resource-causality%22#L60ced7f3:503:509

taylorsilva:
We noticed these tests were taking over 400s to finish before:
https://github.com/concourse/concourse/blob/5558b97af40e9e367bae196b254bbc64675b463d/testflight/resource_type_test.go#L11

On my machine they could run under 100s. From this PR's checks they took under 400s now but still very slow.

PR #7218, opened 2021-06-25
Apply a minimum rate limit for resource checking

aoldershaw:

Primarily:

* [x] Apply a minimum rate limit of 1 check/second if no maximum rate limit is provided (i.e. we're using the default of `num resources / check interval` checks/second), and the calculated rate limit is very low
  * See 14e2fdb41ebb8da5694539d24a34ac0deebbe1e8 for more details

Also, some other small adjustments:
* [x] Don't rate limit resource type checks
* [x] Don't consider resources in paused/archived pipelines for resource count
* [x] Allow aborting checks while awaiting the checking lock

## Release Note

* If `CONCOURSE_MAX_CHECKS_PER_SECOND` is unset, Concourse will try to distribute checks evenly over the course of the check interval to reduce the concurrent load on external systems.
* If there are few resources in a Concourse deployment (~1-20), checks may have to wait a substantial amount of time to run in order to space the checks out evenly. However, there's no real benefit to doing this, since having just a few resources doesn't cause significant load in the first place.
* Now, Concourse ensures that at least one check is allowed to run per second

### DIFF
--- a/atc/db/resource_check_rate_limiter.go
+++ b/atc/db/resource_check_rate_limiter.go
@@ -3,6 +3,7 @@ package db
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	sq "github.com/Masterminds/squirrel"
@@ -23,6 +24,7 @@ type ResourceCheckRateLimiter struct {
 	refreshLimiter *rate.Limiter
 
 	clock clock.Clock
+	mut   *sync.Mutex
 }
 
 func NewResourceCheckRateLimiter(
@@ -36,6 +38,7 @@ func NewResourceCheckRateLimiter(
 	limiter := &ResourceCheckRateLimiter{
 		minChecksPerSecond: minChecksPerSecond,
 		clock:              clock,
+		mut:                new(sync.Mutex),
 	}
 
 	if checksPerSecond < 0 {
@@ -60,11 +63,9 @@ func NewResourceCheckRateLimiter(
 func (limiter *ResourceCheckRateLimiter) Wait(ctx context.Context) error {
 	logger := lagerctx.FromContext(ctx)
 
-	if limiter.refreshLimiter != nil && limiter.refreshLimiter.AllowN(limiter.clock.Now(), 1) {
-		err := limiter.refreshCheckLimiter()
-		if err != nil {
-			return fmt.Errorf("refresh: %w", err)
-		}
+	err := limiter.refreshCheckLimiterIfNeeded()
+	if err != nil {
+		return fmt.Errorf("refresh: %w", err)
 	}
 
 	reservation := limiter.checkLimiter.ReserveN(limiter.clock.Now(), 1)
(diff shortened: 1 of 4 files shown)


evanchaoli:
@aoldershaw I'm thinking that, with #7208, check rate limit might be no longer needed. I believe the rate limit was added together with the check refactor to mitigate performance drop. With #7208, lidar check will be almost the same cheap as 6.x, thus it might not need rate limit.

aoldershaw:
@evanchaoli the rate limiting was actually added in 6.3 (https://github.com/concourse/concourse/releases/tag/v6.3.0#4624), partially focused on reducing load in the Concourse cluster, but more focused on reducing load on external systems by not hitting them all at once.

So, while I think the improvements you've added in #7208 will probably help a lot, I think we still want to keep the resource rate limiting

atc/db/resource_check_rate_limiter.go
```diff
@@ -45,15 +42,16 @@ func NewResourceCheckRateLimiter(
...
+		limiter.checkLimiter = rate.NewLimiter(rate.Inf, 1)
 	}
 
 	return limiter
 }
 
 func (limiter *ResourceCheckRateLimiter) Wait(ctx context.Context) error {
-	limiter.mut.Lock()
-	defer limiter.mut.Unlock()
-
```
chenbh: While the only race condition I can think of is the `limter.checkLimiter.SetLimit()` getting called twice with different values, I still think it would be better keep this mutex just for sanity. If we get rid of the defer and explicitly call Unlock on line 58 and 63, it should solve the inconsistent delay.

atc/db/resource_check_rate_limiter.go
```diff
@@ -81,44 +88,35 @@ func (limiter *ResourceCheckRateLimiter) Wait(ctx context.Context) error {
...
 
-	count := resourceCount + resourceTypeCount
 	limit := rate.Limit(float64(count) / limiter.checkInterval.Seconds())
 	if count == 0 {
 		// don't bother waiting if there aren't any checkables
 		limit = rate.Inf
 	}
+	if limit < limiter.minChecksPerSecond {
+		limit = limiter.minChecksPerSecond
+	}
```
chenbh, line 117: Wild thought: can we setup a notifier to refresh the check limit every time a pipeline is saved/deleted/paused? To deal with larger installations we probably still want to rely on the periodic refresh. So it would be something like: if # of pipelines/resources < threshold, then use the notifier to refresh on pipeline changes, otherwise refresh every minute.

Haven't thought much about the complexities of that solution, happy to use the current solution if you think it's too much effort for such a small problem.

aoldershaw: Yeah I thought about that, but I felt like it wasn't worth the complexity it adds, especially considering the fact that if the number of resources is low then you don't really need to rate limit in the first place

atc/db/resource_check_rate_limiter.go
aoldershaw: I guess the implications of guarding the refresh with a mutex is that later calls to `Wait` will have to wait for the limit to be updated before proceeding, which seems reasonable. Updated

chenbh (approved):

PR #7220, opened 2021-06-28
Remove prototype_id from builds table

aoldershaw:

with concourse/concourse#7176, I realize we don't gain much from keeping
track of the prototype_id or resource_type_id on the builds table, as
the only time we will ever create checks for them is when we manually
fly check-{proto,resource-}type. this ID is used:

1. to ensure we don't create unnecessary checks (for non-manually
   triggered checks)
2. to GC old checks (i.e. where there is already a newer finished
   check).
3. to emit build events to the check_build_events partition for
   optimized GC

however, given that checks for these types will be limited to only
manually triggered checks, this is no longer really worthwhile.

1. all checks are manually triggered anyway, so no difference
2. there will be far fewer checks so GC'ing likely won't be worthwhile.
   if we do want to GC these checks, however, we could add a time-based
   GC (e.g. remove check builds older than 5 days)
3. same as 2

* [x] remove `prototype_id` from `builds` table

### DIFF
--- a/atc/db/build.go
+++ b/atc/db/build.go
@@ -69,7 +69,6 @@ var buildsQuery = psql.Select(`
 		b.job_id,
 		b.resource_id,
 		b.resource_type_id,
-		b.prototype_id,
 		b.team_id,
 		b.status,
 		b.manually_triggered,
@@ -129,8 +128,6 @@ type Build interface {
 	ResourceName() string
 
 	ResourceTypeID() int
-
-	PrototypeID() int
 
 	Schema() string
 	PrivatePlan() atc.Plan
@@ -224,7 +221,6 @@ type build struct {
 	resourceName string
 
 	resourceTypeID int
-	prototypeID    int
 
 	isManuallyTriggered bool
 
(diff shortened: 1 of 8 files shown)


xtremerui:
when we fly check resource, there is an implicit check of its resource type (when version is not available), does that kind of check build of resoure type count here?

aoldershaw:
@xtremerui currently on master, the answer is yes - we'll end up creating a separate build for the resource type as well. When #7176 is merged, however, we won't create a check build for the resource type explicitly. Instead, the check build for the resource will include a `check` for the resource type. Note that this doesn't affect `prototype_id` currently, since prototypes aren't checked automatically.

Once #7176 is merged, though, we should be able to also remove `resource_type_id` from the `builds` table. Note that `fly check-resource-type`/`fly check-prototype` will still create a build that is "for" a resource type/prototype - so we'll probably still want a way to GC old builds. But it doesn't need to be based on the builds table

xtremerui (approved):

PR #7239, opened 2021-07-02
Fixed a bug that caused time resource inaccurate.

evanchaoli:

When I tested #7208 with a lot of `time` resources, I hit an issue where a many `time` resources did never got checked.

Then I found that resource config scope acquires lock by using resource config id as lock id. For `unique version history` resources, like `time`, their resource configures are same but scope are different, so it should use scope id as lock id.

* [x] done

## Release Note

Fixed a bug that caused inaccurate `time` resource, especially on deployments that container big number of short interval `time` resources.

### DIFF
--- a/atc/db/lock/lock.go
+++ b/atc/db/lock/lock.go
@@ -15,7 +15,7 @@ import (
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
 
 const (
-	LockTypeResourceConfigChecking = iota
+	LockTypeResourceConfigScopeChecking = iota
 	LockTypeBuildTracking
 	LockTypeBatch
 	LockTypeVolumeCreating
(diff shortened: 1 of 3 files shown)


clarafu:
This will cause a different bug where with global resources turned off, multiple resources that point to the same resource config but have different resource config scopes will start running their checks at the same time using the same check container because check containers are per `resource_config_id` (due to the `resource_config_check_sessions`). So this lock was to avoid multiple resources that share the same resource config running at the same time because they use the same check container.

evanchaoli:
@clarafu I have noticed the problem you stated during testing #7208. I tried to change check owner to use scope id, but end up leading to a lot of extra containers, and huge amount of volumes that could not be GC-ed. So I will give up the change, leave it to future if it's really a problem. As far as I see, it should not be a big problem on our prod cluster.

PR #7247, opened 2021-07-05
Use browser cache API for dashboard caching

aoldershaw:

when I first implemented #5262, I wasn't aware of the browser cache API:
https://developer.mozilla.org/en-US/docs/Web/API/Cache

instead, I used localStorage, which isn't ideal since it has a pretty
small size capacity, forcing us to artificially bound the amount of data
that gets stored (#5363). this results in us displaying only a subset of
the data, which is unfortunate.

at the time, I also experimented with indexeddb which has a larger
capacity than localStorage, but decided against it since it was
noticably slower.

the browser cache API seems to be really fast, and has even more
generous storage limits than indexeddb. it also has fairly good browser
support - the only browser that doesn't support it is IE. given that
this isn't an application critical enhancement, I think it's probably
fine to not support this in IE.

## Release Note

* The cached API responses on the dashboard no longer need to get truncated, which was previously introduced to work around `localStorage` limits

### DIFF
--- a/web/elm/src/Application/Application.elm
+++ b/web/elm/src/Application/Application.elm
@@ -460,6 +460,7 @@ subscriptions : Model -> List Subscription
 subscriptions model =
     [ OnNonHrefLinkClicked
     , OnLocalStorageReceived
+    , OnCacheReceived
     , OnWindowResize
     ]
         ++ (if model.session.draggingSideBar then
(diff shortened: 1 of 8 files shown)


taylorsilva (approved):
LGTM

Adding some logging lines to see when things get called. Was able to save well above that 1k job limit.
image

I had to use the `mutual-tls.yml` override file since caching only works with https:
image
Spent a while wondering why it appeared to not be working until I read that little box 😛

aoldershaw:
> I had to use the mutual-tls.yml override file since caching only works with https:

Huh that's annoying, I hadn't noticed that when investigating the API. I guess that's not the case with Chrome, but is with Firefox (as I didn't have to do anything special). Maybe it's not the end of the world, though - large(ish) deployments are the ones that could really stand to benefit from caching at all, and they *should* be running https anyway. But still, not ideal...

PR #7249, opened 2021-07-05
Fix browser back button after selecting a group

### LINKED ISSUE #7248: Going back via browser button needs 2 clicks after navigating to group
## Summary

After selecting a pipeline in the UI, if I navigate to a group (i.e. `pipeline > group 1 > group 2 > group N`)  navigating in reverse using the browser's back button requires 2 clicks per group (i.e. two clicks to go from group N to group N-1)

## Steps to reproduce

1. Go to https://ci.concourse-ci.org/teams/main/pipelines/concourse
2. Click the `quickstart` group at the top
3. Click back (nothing changes)
4. Click back (goes back to 1.)

## Expected results

The first "back" click should go back to the previous view one just came from (before clicking a link). 

## Actual results

The first "back" click seems to do nothing (URI doesn't change), only after the 2nd click does the UI actually go "back".

## Additional context

Tried both firefox and chromium and had the same behaviour.

## Triaging info

* Concourse version: 7.3.2
* Browser (if applicable): Firefox 89.0.2 / Chromium 91.0.4472.114
* Did this used to work? Not sure only noticed it recently but I _think_ the issue has been there a while (at least in 7.2)

aoldershaw:

closes #7248

prior to this fix, we ended up emitting a NavigateTo effect in addition
to the href that's applied on the element, so we effectively pushed
history twice.

## Release Note

* Previously, if a pipeline group was selected in the UI, the back button would not work (you'd have to press it twice to go back)

### DIFF
--- a/web/elm/src/Message/Message.elm
+++ b/web/elm/src/Message/Message.elm
@@ -1,9 +1,10 @@
 module Message.Message exposing
-    ( DomID(..)
+    ( CommentBarButtonKind(..)
+    , DomID(..)
     , DropTarget(..)
     , Message(..)
     , PipelinesSection(..)
-    , VersionId,CommentBarButtonKind(..)
+    , VersionId
     , VersionToggleAction(..)
     , VisibilityAction(..)
     )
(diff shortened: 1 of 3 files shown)


xtremerui (approved):
LGTM! Tested with local build, now the back button will function correctly.

PR #7268, opened 2021-07-09
Cache the list of workers in memory

aoldershaw:

On hush-house and CI, the following queries take up a substantial amount of DB CPU time:

* `WorkerFactory.Workers`
* `WorkerFactory.BuildContainersCountPerWorker`

They are each called every time a container is scheduled - on hush-house, this amounts to ~250,000 times an hour. This worker data can easily fit into memory to significantly reduce the number of round-trips to the DB.

* [x] Extend notification bus to support queueing up multiple notifications and sending payloads
* [x] Add postgres triggers to the `workers` and `containers` tables - these triggers send a notification along the notification bus containing the type of operation (`UPDATE,` `INSERT`, or `DELETE`) and the affected worker name
* [x] Implement a `WorkerCache` that listens for these trigger events and modifies in-memory data structures
    * `WorkerCache` fetches this data every minute as well, to ensure the data doesn't get out of sync

## Release Note

* Scheduling containers should be more performant by reducing the number of required database calls


atc/db/migration/migrations/1625844436_add_worker_cache_triggers.down.sql
```diff
@@ -0,0 +1,4 @@
+DROP TRIGGER workers_upsert_or_delete_trigger ON workers;
```
xtremerui, line 1: could we add a `if exist` here?

xtremerui (commented):
is the extended notification bus queue for only notifications from that trigger or is it queueing for all? It seems like it is for all notifications. In this case, is there performance trade off (or gain) because of the queue?

atc/db/worker_cache.go
```diff
@@ -0,0 +1,278 @@
...
+		cache.removeWorker(name)
+		return nil
+	}
+
+	cache.mut.Lock()
+	defer cache.mut.Unlock()
+
+	cache.workers[name] = worker
+	if _, ok := cache.workerContainerCounts[name]; !ok {
+		cache.workerContainerCounts[name] = 0
```
xtremerui, line 268: when cache is un-sync with DB like this, should it be removed like line 259 or set to 0?

aoldershaw: This condition isn't necessarily about being out-of-sync - if a new worker is inserted, it won't yet have an entry in the `workerContainerCounts` table. Since it can't have containers yet, we add it to the table with 0 containers

aoldershaw:
> is the extended notification bus queue for only notifications from that trigger or is it queueing for all? It seems like it is for all notifications. In this case, is there performance trade off (or gain) because of the queue?

The queueing in the notification bus is really just a generalization of the existing behaviour. Previously, all uses of the notification bus were "tell me *if* something has changed". Since we ignore messages that come in when the queue size is exceeded, we achieved this desired behaviour using a queue size of 1 for the listening channel. All that to say, this doesn't impact the behaviour/performance of the existing notification bus users - it just makes setting the queue size explicit.

Now that we're sending messages that are more meaningful than "has something changed", and all of those messages need to be processed individually, we need to allow sending more messages along a single channel, since the consumer of the channel may be slower than the producer (depending on how Go schedules goroutines). This shouldn't be significantly different in terms of performance than with a queue size of 1

xtremerui (approved):

PR #7288, opened 2021-07-15
containerd: split long property values into multiple chunks

### LINKED ISSUE #7284: [containerd] Resource metadata that is too long can't be stored
## Summary

Containerd has a label length limit of 4096 bytes for both key + value for a single label.

We try to store the full version + metadata from get/put steps on the container as a label:
https://github.com/concourse/concourse/blob/2563979646ccc728b051d32a769f6a05deec2c00/atc/worker/container.go#L237

For resource versions that have a lot of metadata, this can result in errored builds:
Screen Shot 2021-07-14 at 1 24 34 PM

## Additional context

We only started noticing this after bumping containerd to 1.5.3, but don't see anything about it in the release notes.

An example "problematic" resource version is runc v1.0.0, which has a body of >4400 bytes:

* Build prior to bumping CI to a concourse version with containerd == v1.5.3: https://ci.concourse-ci.org/teams/main/pipelines/branch/jobs/dev-image/builds/65?vars.branch=%22rewrite-atc-worker%22#L60eb1a44:2
* Build after bumping CI to a concourse version with containerd == v1.5.3: https://ci.concourse-ci.org/teams/main/pipelines/branch/jobs/dev-image/builds/66?vars.branch=%22rewrite-atc-worker%22#L60ee1c94:9

## Triaging info

* Concourse version: master as of b396dfed6f10765ff70d7fbae1afe7453ac5d083
* Did this used to work? Yes

aoldershaw:

closes #7284

Our containerd runtime started erroring when resource versions would emit metadata strings that were too long. This is because we store the version + metadata (encoded as JSON) on the container as a garden.Property after the `in`/`out` script succeeds (so that if the ATC re-attaches to the build, we can check if we've already run the script, and if so, retrieve the result).

We previously stored garden properties as containerd labels in a 1:1 mapping. However, containerd v1.5.3 started rejecting labels that were "too long" (i.e. where the length of the key + value exceeded 4096 bytes).

As a workaround, we now split property values up into multiple chunks that get stored as sequential labels as needed. For instance, suppose we wanted to save the following property:

foo: aaaa....(5000 times)

This will be stored as the following labels in containerd:

foo.0: aaaa....(4091 times, so that the len(key) + len(value) == 4096)
foo.1: aaaa....(909 times)

Note that if a key + value does not exceed the limit, we still use the same `key.0` pattern. e.g.

foo: aaa

is stored as

foo.0: aaa


worker/runtime/integration/integration_test.go
```diff
@@ -909,6 +910,62 @@ func (s *IntegrationSuite) TestRequestTimeoutZero() {
...
+
+	containers, err = s.gardenBackend.Containers(garden.Properties{
+		"long1": longString,
+		"long2": longString,
+	})
+	s.NoError(err)
+	s.Len(containers, 1)
+
+	err = container.SetProperty(longString, "foo")
+	s.Error(err)
```
xtremerui, line 958: do we care about the err content here?

worker/runtime/properties.go
```diff
@@ -2,10 +2,106 @@ package runtime
...
+//
+func propertiesToLabels(properties garden.Properties) (map[string]string, error) {
+	// Hard restriction on the total length of key + value imposed by
+	// containerd on a per-label basis.
+	const maxLabelLen = 4096
+
+	labelSet := map[string]string{}
+	for key, value := range properties {
+		sequenceNum := 0
+		if len(key) > maxLabelLen/2 {
```
xtremerui, line 32: is the key size limited to 2048 by containerd?

xtremerui (commented):
I understand it is all runtime but just want to make sure after worker is bumped all container info is gone so we dont need to worry about obsolete data.

worker/runtime/properties.go
aoldershaw: No, there's no explicit restriction on key size - all containerd cares about is the combined key + value size.

However, certainly the key + the sequence number suffix (`.0`, `.1`, `.12345`, etc.) can't be longer than 4096, since then no value could be stored. Having a very long (but still <4096) key also means the value needs to be split over many labels, which could cause other problems. For instance, if the key + sequence number is 4095 bytes, and you try to store a 100 byte value, you'd end up needing 100 labels (consuming `400KB` of storage) - given that containerd's label system isn't designed for this, we didn't want to ever let it split this aggressively, so we picked an arbitrary key length limit of half the total length. Given that all of our property names are fixed strings (of ~30 bytes), we felt this was okay.

However, I think this probably deserves a comment, since it is just an arbitrary limit for our property system. Thanks for calling it out!

worker/runtime/integration/integration_test.go
aoldershaw: Sure, will update

aoldershaw:
@xtremerui good point about the obsolete data. While it's true that any existing properties stored on containers will no longer be readable, I don't think this is a huge concern. It means that if the ATC is restarted in the middle of builds running, it may have to re-run certain steps that it could otherwise just skip over (when it re-attaches to the build). Since this only affects the initial restart, though, maybe this isn't such a problem.

xtremerui (approved):

PR #7294, opened 2021-07-16
Fix prometheus emitter not setting default attributes

### LINKED ISSUE #6351: Default metric attributes aren't propagated to prometheus emitter
## Summary

Any additional attributes set by `--metrics-attribute=NAME:VALUE` are ignored by the prometheus emitter 

## Steps to reproduce

config.yml:
```
version: '3'

services:
  web:
    environment:
      CONCOURSE_METRICS_ATTRIBUTE: foo:bar
```

1. Run `docker-compose -f docker-compose.yml -f hack/overrides/prometheus.yml -f config.yml up -d`
1. Navigate to http://localhost:9090/graph?g0.expr=concourse_http_responses_duration_seconds_count

## Expected results

I should see the metrics with a `foo="bar"` attribute:
```
concourse_http_responses_duration_seconds_count{instance="web:9100", job="concourse", method="POST", route="RegisterWorker", status="200", foo="bar"}
```

## Actual results

```
concourse_http_responses_duration_seconds_count{instance="web:9100", job="concourse", method="POST", route="RegisterWorker", status="200"}
```

## Additional context

We handle prometheus in a weird way, we're defining all the metrics and possible tags at:
https://github.com/concourse/concourse/blob/a851d623b0cf4ea705ad3d3fa3af755b8a9627b2/atc/metric/emitter/prometheus.go#L325-L333

When the emitter is handling an event, `event.Attributes` should contain all the event attributes in addition to the default ones, but they're being ignored by the handler
https://github.com/concourse/concourse/blob/a851d623b0cf4ea705ad3d3fa3af755b8a9627b2/atc/metric/emitter/prometheus.go#L754-L774

## Triaging info

* Concourse version:
* Browser (if applicable): 
* Did this used to work? probably not

chenbh:

closes #6351 

## Notes to reviewer

I've mentioned it in the commit message, but while [best practices](https://prometheus.io/docs/instrumenting/writing_exporters/#target-labels-not-static-scraped-labels) is to configure const labels like this in the scrape config, the fact that the other emitters support default attributes makes this a bug imo.

## Release Note
Additional metrics attributes configured by `--metrics-attribute` now propagates to the prometheus emitter correctly.

### DIFF
--- a/atc/metric/emitter/influxdb.go
+++ b/atc/metric/emitter/influxdb.go
@@ -45,7 +45,7 @@ func init() {
 func (config *InfluxDBConfig) Description() string { return "InfluxDB" }
 func (config *InfluxDBConfig) IsConfigured() bool  { return config.URL != "" }
 
-func (config *InfluxDBConfig) NewEmitter(map[string]string) (metric.Emitter, error) {
+func (config *InfluxDBConfig) NewEmitter(_ map[string]string) (metric.Emitter, error) {
 	client, err := influxclient.NewHTTPClient(influxclient.HTTPConfig{
 		Addr:               config.URL,
 		Username:           config.Username,
--- a/atc/metric/emitter/influxdb.go
+++ b/atc/metric/emitter/influxdb.go
@@ -45,7 +45,7 @@ func init() {
 func (config *InfluxDBConfig) Description() string { return "InfluxDB" }
 func (config *InfluxDBConfig) IsConfigured() bool  { return config.URL != "" }
 
-func (config *InfluxDBConfig) NewEmitter() (metric.Emitter, error) {
+func (config *InfluxDBConfig) NewEmitter(map[string]string) (metric.Emitter, error) {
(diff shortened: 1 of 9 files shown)


atc/metric/emitter/influxdb.go
```diff
@@ -45,7 +45,7 @@ func init() {
 func (config *InfluxDBConfig) Description() string { return "InfluxDB" }
 func (config *InfluxDBConfig) IsConfigured() bool  { return config.URL != "" }
 
-func (config *InfluxDBConfig) NewEmitter() (metric.Emitter, error) {
+func (config *InfluxDBConfig) NewEmitter(map[string]string) (metric.Emitter, error) {
```
analytically, line 48: _?

chenbh: Good catch, I'm surprised there was no compiler errors, but turns out this is perfectly valid syntax and behaves the same! Still makes it harder to read so I fixed it anyways

aoldershaw (approved):

analytically (approved):

PR #7298, opened 2021-07-19
Add force_refresh to get step.

evanchaoli:

This is a follow-up of https://github.com/concourse/concourse/pull/6660#issuecomment-872750759, providing a flag `force_refresh` to `get` step to disable local cache of specific resource-get.

* [x] done

## Notes to reviewer

NA

## Release Note

Usually given a version, `get` steps on a resource should always generate same artifacts. But if that's not true for some resource, e.g. a resource's `get` step utilizes BUILD environment variables, then use `force_refresh` to disable local cache for the resource:

```yaml
- get: some-resource
  force_refresh: true
```

### DIFF
--- a/atc/builds/planner.go
+++ b/atc/builds/planner.go
@@ -122,13 +122,14 @@ func (visitor *planVisitor) VisitGet(step *atc.GetStep) error {
 	visitor.plan = visitor.planFactory.NewPlan(atc.GetPlan{
 		Name: step.Name,
 
-		Type:     resource.Type,
-		Resource: resourceName,
-		Source:   resource.Source,
-		Params:   step.Params,
-		Version:  &version,
-		Tags:     step.Tags,
-		Timeout:  step.Timeout,
+		Type:         resource.Type,
+		Resource:     resourceName,
+		Source:       resource.Source,
+		Params:       step.Params,
+		Version:      &version,
+		Tags:         step.Tags,
+		Timeout:      step.Timeout,
+		ForceRefresh: step.ForceRefresh,
 
 		VersionedResourceTypes: visitor.resourceTypes,
 	})
--- a/atc/builds/planner_test.go
+++ b/atc/builds/planner_test.go
@@ -80,12 +80,13 @@ var factoryTests = []PlannerTest{
 	{
 		Title: "get step",
 		Config: &atc.GetStep{
-			Name:     "some-name",
-			Resource: "some-resource",
-			Params:   atc.Params{"some": "params"},
-			Version:  &atc.VersionConfig{Pinned: atc.Version{"doesnt": "matter"}},
-			Tags:     atc.Tags{"tag-1", "tag-2"},
-			Timeout:  "1h",
+			Name:         "some-name",
+			Resource:     "some-resource",
+			Params:       atc.Params{"some": "params"},
+			Version:      &atc.VersionConfig{Pinned: atc.Version{"doesnt": "matter"}},
+			Tags:         atc.Tags{"tag-1", "tag-2"},
+			Timeout:      "1h",
+			ForceRefresh: true,
 		},
 		Inputs: []db.BuildInput{
 			{
@@ -104,6 +105,7 @@ var factoryTests = []PlannerTest{
 				"version": {"some":"version"},
 				"tags": ["tag-1", "tag-2"],
 				"timeout": "1h",
+				"force_refresh": true,
 				"resource_types": [
 					{
 						"name": "some-resource-type",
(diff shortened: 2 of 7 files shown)


clarafu:
Hm I find this a little awkward because this would only be useful if the user has both `enable-cache-streamed-volumes` to true and set this field on the `get` step. So this is purely a configuration for a volume caching mechanic behind a feature flag, but I might be able to look past it.

Another concern I have is that I think this won't actually solve the problem that this was intended to solve. You were right in your reply in the PR that it was just chance that they did not run into any problems with the local caching disabled, but it most definitely possible for it to reuse the get container (maybe their container strategy made it difficult to happen). This pr won't actually force a refresh on the get step, it will just make it so that the chances of it reusing an existing get container low again. Because if concourse happens to pick the same worker that already has that resource cache on it, it won't fetch the get container again.

aoldershaw:
Without passing judgment on whether this feature is the right way forward - I agree with Clara that it should apply when `EnableCacheStreamedVolumes` is disabled, too. #6597 consolidates the logic for checking the cache (which is currently spread out between the `exec/get_step.go` and the `worker/fetcher.go`) - so I'd suggest waiting for that to land (hopefully soon).

Also, if `force_refresh` is enabled, does it still make sense to initialize the fetched volume for the resource cache? Granted there could be other places where we `get` the same resource without `force_refresh`, and it could speed those up, but it seems like that should be pretty unlikely - and keeping a volume around only for it to be ignored seems wasteful

evanchaoli:
Actually, I also very hesitate in adding `force_refresh`, that is reason why I made limited code changes and waiting for  opinions.

Given a version, `get` should always generate the same artifacts. The principle should always be followed. `force_refresh` just comprises those custom resource types that doesn't follow the principle.

The most chances of breaking the principle is to use BUILD_ID/NAME in `get`. Should we just stop exposing build envs to `get` and `check`?

clarafu:
I wonder if there is a way to use an existing get container and apply the appropriate metadata by overwriting the environment variables before returning it at the end of the get step.

xtremerui:
Hi @evanchaoli, are you still on this or could we close it for now?

lrstanley:
I feel as though before this is closed, maybe we can discuss if a new issue should be created to address some of the issues that exist today with build environment variables and their usage within the get step?

> Given a version, get should always generate the same artifacts. The principle should always be followed.

If the above quote is true, then I see no scenario where these build variables should be used within `get` steps, especially because this leads to a potential vulnerability, if using global variables (and maybe other situations, but I'm still not versed on some of the internals).

I.e. if someone knew that the get step relied on `BUILD` environment variables, they could add multiple get steps that try and produce different team (or similar) information. If other pipelines under other teams use that same resource, those pipelines could pull in resources that use a different team name. If whatever the task does with the input data from that resource, this could lead to resources being generated with incorrect source/team data.

I'm not 100% sure what I said above would actually be the case, but from the testing I previously did with build metadata, seems like it is. I also understand that most resources don't use the build metadata in a get step, but I really think build environment variables shouldn't be allowed in `get` steps, just to be safe.

Additionally, I think for those who did want resources that had that information, it seems like one of the alternative solutions that exists in many different CI/CD platforms, and could still allow the "stateless way" that Concourse promotes, is passing in build parameters to all tasks/jobs (and ofc not `get` steps). I also mentioned this in [#6660](https://github.com/concourse/concourse/pull/6660#issuecomment-872655917), and I believe others would really like it, or have tried to do similar in the past ([discord user](https://discord.com/channels/219899946617274369/413770960089382922/934056274553884703), [resource 1](https://github.com/olhtbr/metadata-resource), [resource 2](https://github.com/SWCE/metadata-resource), [resource 3](https://github.com/vito/build-metadata-resource), and [my recent attempt](https://github.com/hangar-ci/metadata-resource)) -- has something like this been explicitly denied in the past, as a feature request? If not, would I be able to open a new feature request for this functionality? as well as another for removing `get`'s ability to obtain build info?

lrstanley:
Another thought -- if some of these parameters were exposed as some kind of var source (though treated differently, can't be cached if using build numbers and similar), and passed in as inputs to the get step, that could ensure there are always consistent inputs and outputs. I.e. you choose what you pass into the resource, and thus the user still controls the output. Additionally, I think that would reduce the chance on tasks that theoretically could use build variables (if my previous post were implemented in some way), i.e. they're not magic variables that are just "there" and available when running a script, rather variables that are passed to jobs/tasks by being explicit. I'm not sure how complex that would be to implement, but I think it keeps the stateless approach, but also allows folks to be much more dynamic in their resources and tasks.

We do something like this today, manually, but it's def. not a perfect solution. When a user initially sets the pipeline, our template passes in things like team-name, and pipeline-name, as these are passed to resources and/or tasks that need that for generating artifacts (e.g. artifacts being pushed to something like Artifactory/Nexus, that show where that .jar or similar was built from). The obvious downside is if the team name changes, the pipeline name changes, a bunch of other logic has to exist to ensure those are updated. Additionally, because we pass these things via the command line when we set the pipeline (rather than a yaml config, that will easily get outdated), whenever we do `set_pipeline`, we have to ensure all cli variables are re-defined there, otherwise all cli variables get lost when the pipeline updates itself, or updates child pipelines.

And I know this isn't the perfect place to discuss this, I'm just not sure if these ideas have already been shut down in the past, so not sure if I should create another issue.

xtremerui:
@lrstanley you can for sure create an issue to better track your ideas and thoughts if you are afraid of them being buried here. 

Due to the inactivity of this PR we will need to close it for now.

PR #7307, opened 2021-07-19
Propagate groups between subpages of a pipeline

clarafu:

If the user was viewing a group within a pipeline page, the group will be propagated when the user moves
between subpages (job, build, resource, causality) of a pipeline. To explain more clearly, lets say there is a pipeline with groups A and B. A user is on group B on the pipeline page then navigates to a job page and then clicks on the pipeline breadcrumb in the topbar, it will navigate the user to group B on the pipeline rather than before it would have navigated the user to the default group on the pipeline page.

## Release Note

* If a user was initially viewing a group in the pipeline page, this will be persisted in the pipeline breadcrumb when navigating between pipeline subpages.

### DIFF
--- a/web/elm/src/Application/Application.elm
+++ b/web/elm/src/Application/Application.elm
@@ -401,25 +401,36 @@ redirectToLoginIfNecessary err ( model, effects ) =
 urlUpdate : Routes.Route -> Model -> ( Model, List Effect )
 urlUpdate route model =
     let
+        oldRoute =
+            model.session.route
+
+        newRoute =
+            case route of
+                Routes.Pipeline _ ->
+                    route
+
+                _ ->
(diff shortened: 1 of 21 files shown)


web/elm/src/Application/Application.elm
```diff
@@ -401,25 +401,37 @@ redirectToLoginIfNecessary err ( model, effects ) =
 urlUpdate : Routes.Route -> Model -> ( Model, List Effect )
 urlUpdate route model =
     let
+        oldRoute =
+            model.session.route
+
+        newRoute =
+            case route of
+                Routes.Pipeline _ ->
+                    -- untested
```
aoldershaw, line 410: Can remove this 😁 it is in fact tested

chenbh (approved):
lgtm

PR #7310, opened 2021-07-20
Properly map `version_from` within `across` substep

aoldershaw:

previously, we didn't properly handle version_from in the across
substep. we would generate a template for the substep with dummy subplan
IDs, and when we render the template for each substep, we set each of
the subplan IDs to be unique value.

this however didn't take into account version_from for the GetPlan. for
instance, a `put` step will generate a PutPlan and a GetPlan, where the
`get` uses the `version_from` the put. since we weren't properly mapping
the version_from to the generated unique subplan ID, we weren't able to
use a put step as an across substep.

since this is such a subtle thing, and is very easy to miss, I added a
unit test that walks the atc.Plan type and verifies that any fields
containing an atc.PlanID must be properly handled. a field is considered
"handled" if it's explicitly added to a whitelist of handled fields -
any other fields will be considered unhandled


atc/engine/build_step_delegate.go
```diff
@@ -373,11 +373,29 @@ func (delegate *buildStepDelegate) ConstructAcrossSubsteps(templateBytes []byte,
...
 			return nil, fmt.Errorf("invalid template bytes: %w", err)
 		}
+
+		// Maps from the original subplan ID generated by the planner to the
+		// translated ID unique to the substep iteration.
+		mappedSubplanIDs := map[atc.PlanID]atc.PlanID{}
 		planIDCounter := 0
 		subPlan.Each(func(p *atc.Plan) {
-			p.ID = atc.PlanID(fmt.Sprintf("%s/%d/%d", delegate.planID, i, planIDCounter))
+			mappedID := atc.PlanID(fmt.Sprintf("%s/%d/%d", delegate.planID, i, planIDCounter))
```
xtremerui, line 382: is there a specific format of the mappedID? How is it decided?

atc/engine/build_step_delegate.go
```diff
@@ -373,11 +373,29 @@ func (delegate *buildStepDelegate) ConstructAcrossSubsteps(templateBytes []byte,
...
+			mappedSubplanIDs[p.ID] = mappedID
+			p.ID = mappedID
 			planIDCounter++
 		})
+
+		subPlan.Each(func(p *atc.Plan) {
```
xtremerui, line 388: we have two loops of subPlan, is it possible to combine them?

atc/engine/build_step_delegate.go
aoldershaw: There doesn't have to be a specific format, as long as it's unique within the build. Easiest way I thought to do that was this way - if the `across` step has plan ID `xyz`, then the `j`th subplan of the `i`th iteration gets ID `xyz/i/j`.

Note that the plan IDs don't need to be predictable, but it's kind of nice to have. The only place you'll see them is the URL when you highlight build logs in the UI, e.g. https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/testflight/builds/879#L60f3a721/2/0:132 (here, the `across` plan ID is `60f3a721`, this is iteration with index `2`, and the subplan with index `0`).

We do a similar thing with the image check and get subplans - e.g. https://dispatcher.concourse-ci.org/teams/main/pipelines/production/jobs/bump-concourse-to-rc/builds/8#L609c9bc9/image-get:1 (the plan ID is `609c9bc9`, which we suffix with `/image-get`).

atc/engine/build_step_delegate.go
aoldershaw: Combining them probably *would* still work, but it would introduce some coupling, since we'd need to depend on the order that `atc.Plan.Each` calls the function in. e.g. if `atc.Plan.Each` starting doing a pre-order/post-order traversal, this would break things.

When we do a first pass for re-mapping the IDs, we know that `mappedSubplanIDs` has processed *all* of the subplans. This makes it safe to rely on in the second pass for the `VersionFrom` mapping, regardless of how `atc.Plan.Each` traverses the plans.

That said, maybe that's fine - it's nice to know what type of traversal a function will use, so maybe that should just be a part of the contract of the method. But by the same token, iterating twice also isn't that big a deal, and won't really be a performance burden

taylorsilva (approved):
LGTM

PR #7338, opened 2021-07-26
Enable emitting dogstatsd metrics over uds

jmhwang7:
DogStatsD allows [sending metrics with a Unix Domain Socket (UDS)](https://docs.datadoghq.com/developers/dogstatsd/unix_socket/?tab=kubernetes#using-origin-detection-for-container-tagging), which has several advantages over UDP, like allowing the Agent to detect dropped packets and connection errors. Currently, when the datadog agent restarts, Concourse metrics are not correctly transmitted to the agent. This change adds a config variable for specifying uds file and uses the uds file if specified instead of host/port.

Signed-off-by: Jennifer Hwang <jmhwang7@gmail.com>

* [x] Adds `datadog-agent-uds-filepath` as an option - specified as `CONCOURSE_DATADOG_AGENT_UDS_FILEPATH` environment variable
* [x] Uses `datadog-agent-uds-filepath` to connect to statsd when specified

## Notes to reviewer

The underlying issue is https://github.com/DataDog/datadog-go/issues/54. It's fixed here: https://github.com/containernetworking/plugins/issues/123 , but we are at the mercy of GKE, which has an older version of portmap in their clusters (1.20.8-gke.900 and 1.21.2-gke.600 both have portmap `v0.8.5-gke.1`)

## Release Note

* The Datadog emitter can now be configured to communicate with the Datadog agent over Unix Domain Sockets

### DIFF
--- a/atc/metric/emitter/dogstatsd.go
+++ b/atc/metric/emitter/dogstatsd.go
@@ -19,6 +19,7 @@ type DogstatsdEmitter struct {
 type DogstatsDBConfig struct {
 	Host   string `long:"datadog-agent-host" description:"Datadog agent host to expose dogstatsd metrics"`
 	Port   string `long:"datadog-agent-port" description:"Datadog agent port to expose dogstatsd metrics"`
+	UDS    string `long:"datadog-agent-uds-filepath" description:"Datadog agent unix domain socket (uds) filepath to expose dogstatsd metrics"`
 	Prefix string `long:"datadog-prefix" description:"Prefix for all metrics to easily find them in Datadog"`
 }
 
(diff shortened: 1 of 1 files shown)


atc/metric/emitter/dogstatsd.go
```diff
@@ -31,8 +32,14 @@ func (config *DogstatsDBConfig) Description() string { return "Datadog" }
 func (config *DogstatsDBConfig) IsConfigured() bool { return config.Host != "" && config.Port != "" }
 
 func (config *DogstatsDBConfig) NewEmitter(_ map[string]string) (metric.Emitter, error) {
+	var client *statsd.Client
+	var err error
+	if config.UDS != "" {
+		client, err = statsd.New(config.UDS)
```
aoldershaw, line 38: I think this needs to be prefixed with `unix://`: https://github.com/DataDog/datadog-go/blob/f507940b85a2a97119a07db8be75bdf8544ead50/statsd/statsd.go#L272-L274

Given that the `--datadog-agent-uds-file` already specifies that it's for a unix domain socket, it's probably redundant to require that the user specify `unix://` there, too - so maybe we could just add the prefix here? If so, we'd want to document in the flag that it's just the filepath, without the `unix://` scheme.

Alternatively, we could rename the new flag to be more generic, e.g. `--datadog-agent-address`, and clarify that it could be prefixed with the `unix://` scheme to support UDS.

What do you think?

jmhwang7: Great point! When I was testing, I was prefixing `unix://` myself, but adding it as a prefix is definitely more convenient for everyone else. 

I've updated the parameter name and the description. Do you think that's sufficient or is there a way I can make it clearer?

jmhwang7:
@aoldershaw Thank you for taking a look at the PR so quickly! Is there anything I need to do to get this PR in the next release (7.4.0) or will it get automatically get cut for that release?

aoldershaw:
@jmhwang7 we'll get it in for 7.4.0. Thanks for the PR!

PR #7370, opened 2021-07-30
Don't allow empty identifiers when renaming pipelines/teams

### LINKED ISSUE #7365: fly rename-pipeline must not allow new name to be empty
## Summary

Renaming a pipeline with `fly -t some-target rename-pipeline -o "example" -n ""` succeeds and leaves a pipeline with an empty name. 

The renamed pipeline is visible in the web UI (with an empty name), but cannot be renamed again or even deleted.

## Steps to reproduce

```command
$ fly -t some-target rename-pipeline -o "example" -n ""
```

## Expected results

Error message from `fly` saying that the new name must not be empty.

## Actual results

`fly` accepted the empty string as new pipeline name. The renamed pipeline is visible in the web UI (with an empty name), but cannot be renamed again or even deleted.

## Additional context

We were able to rectify this situation by 

1. Finding the database id of the pipeline:

    ```command
    $ fly -t some-target pipelines --json
    ```

    It shows the pipeline id as `42` for the one pipeline with an empty name.

1. Updating the name using `psql`:

    ```sql
    # update pipelines set name='example' where id=42;
    ``` 

## Triaging info

* Concourse version: v7.3.2
* Did this used to work? don't know
* #6410 may be related

taylorsilva:

closes #7365

## Release Note

* Fixed a bug where a pipeline or team could be renamed to an empty string. The team/pipeline could not be deleted through fly. An error is now returned by the API if the identifier is blank

### DIFF
--- a/atc/api/pipelines_test.go
+++ b/atc/api/pipelines_test.go
@@ -1738,7 +1738,8 @@ var _ = Describe("Pipelines API", func() {
 							requestBody = `{"name":""}`
 						})
 
-						It("returns a warning in the response body", func() {
+						It("returns 400 Bad Request and an error in the response body", func() {
+							Expect(response.StatusCode).To(Equal(http.StatusBadRequest))
 							Expect(ioutil.ReadAll(response.Body)).To(MatchJSON(`
 							{
 								"errors": [
--- a/atc/api/configserver/get.go
+++ b/atc/api/configserver/get.go
@@ -8,6 +8,7 @@ import (
 	"code.cloudfoundry.org/lager"
 
 	"github.com/concourse/concourse/atc"
+	. "github.com/concourse/concourse/atc/api/helpers"
 	"github.com/tedsuo/rata"
 )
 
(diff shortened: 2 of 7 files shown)


atc/api/pipelines_test.go
```diff
@@ -1738,7 +1738,8 @@ var _ = Describe("Pipelines API", func() {
 							requestBody = `{"name":""}`
 						})
 
-						It("returns a warning in the response body", func() {
+						It("returns 400 Bad Request and an error in the response body", func() {
```
xtremerui, line 1741: Do we consider it a breaking change since the warning changed to a 400 ?

taylorsilva: I don't think so because it's the API. We don't support the API so the only clients we should worry about is fly and the web.

taylorsilva:
@xtremerui any other thoughts on this besides your one question?

xtremerui (approved):

PR #7445, opened 2021-08-20
Add team-scoped webhooks

### LINKED ISSUE #2230: Shared webhook endpoint for multiple resources

aoldershaw:

closes #2230
partially based on concourse/rfcs#110

Note: this differs from the RFC in that it doesn't handle automatic resource subscription to webhooks via prototypes - in order for a resource to use a webhook, it must be explicitly configured in the pipeline config.

* [x] Add a new webhook entity that can be managed via `fly`
    * `fly set-webhook` creates/updates a webhook within a team. Webhooks have a name and a type, both of which are arbitrary. Webhooks are also configured with a token, which is added to the webhook endpoint for verifying the caller.
    * `fly destroy-webhook` destroys an existing team webhook

    ```
    $ fly -t dev set-webhook --webhook concourse --type github
    
    warning: if the webhook already exists, this operation will overwrite the type and token
    apply webhook configuration? [yN]: y
    webhook created
    
    configure your external service with the following webhook URL:
    
    http://localhost:8080/api/v1/teams/main/webhooks/concourse?token=0C703AbDlSJnjxsgmpHQhQ
    
    $ fly -t dev destroy-webhook --webhook concourse
    !!! this will remove all data for webbook `concourse`
    
    please type the webhook name to confirm: concourse
    webhook destroyed
    ```
* [x] Allow configuring `resource.webhooks` - a list of webhooks that a resource "subscribes" to
    * Each webhook definition declares a `type` (that matches the type of webhook as created by `fly set-webhook --type`) and a `filter`
    * The `filter` is a JSON value that is used to restrict what webhook payloads will trigger a check for the resource. It uses Postgres' [containment logic](https://www.postgresql.org/docs/9.6/datatype-json.html#JSON-CONTAINMENT) to filter the payloads
    * Consider the following example that could be used in a [PR instanced pipeline](https://github.com/concourse/ci/blob/3f90d17b55b924bef948041acbfbd130294a1103/pipelines/pr.yml#L50-L56) to trigger a check whenever there's a push to the PR:
    ```yaml
    resources:
    - name: concourse-pr
      type: pull-request
      source:
        repository: concourse/concourse
        access_token: ((pull_requests_access_token))
        number: ((number))
      webhooks:
      - type: github
        filter:
          action: synchronize
          repository: {full_name: concourse/concourse}
          pull_request: {number: ((number))}
    ```

* [ ] Test the API handlers and fly commands (I got bored, it's so much boilerplate)

One motivation for this is to reduce checking frequency - you can configure a much longer `CONCOURSE_RESOURCE_WITH_WEBHOOK_CHECKING_INTERVAL` (e.g. `24h`), so resources with webhooks won't "cost" much in terms of resource checks. For more motivations, check the RFC.

## Notes to reviewer

* If you test this on localhost, you'll need to use something like [`ngrok`](https://ngrok.com/) to register as the webhook URL in external services
* To configure a webhook on GitHub, you can follow this guide: https://docs.github.com/en/developers/webhooks-and-events/webhooks/creating-webhooks
* For a list of all the webhook events that GitHub supports, check out https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads


ari-becker:
@aoldershaw any news?

aoldershaw:
@ari-becker sorry, not really. Unfortunately I'm no longer on the Concourse team so don't have too much time to dedicate to bring this forward. That said, I think it definitely has value, so if anyone's interested on taking this on (be it on the Concourse team or the wider community), that'd be awesome!

PR #7449, opened 2021-08-23
Allow team members to archive pipelines

wanderanimrod:
# Why
Team members [can delete pipelines](https://github.com/concourse/concourse/blob/b8658106da5ff6c7434eb17b2c7e60738afcb82b/atc/api/accessor/roles.go#L65) but they can't archive them. I think deleting pipelines should be more restricted than archiving them. 

# What
This PR gives team members `archive-pipeline` permissions.

## Release Note

* Users with the `member` role on a team can now archive pipelines by default. The "archive pipeline" action was previously assigned to the `owner` role. If you've [configured your own RBAC](https://concourse-ci.org/user-roles.html) this change will not effect you.

### DIFF
--- a/atc/api/accessor/roles.go
+++ b/atc/api/accessor/roles.go
@@ -66,7 +66,7 @@ var DefaultRoles = map[string]string{
 	atc.OrderPipelines:                 MemberRole,
 	atc.OrderPipelinesWithinGroup:      MemberRole,
 	atc.PausePipeline:                  OperatorRole,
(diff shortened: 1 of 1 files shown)


taylorsilva (approved):
LGTM - makes sense as the default

In case you didn't know, you can configure RBAC: https://concourse-ci.org/user-roles.html

wanderanimrod:
Thanks a lot @taylorsilva for the merge! Yes, we configure user roles but we wanted this to be the default.

PR #7454, opened 2021-08-24
Fix overlapping between inputs and jobs in UI

### LINKED ISSUE #7207: Jobs with no get/put steps are rendered in the resources column
## Summary

Reported by @khng

The most visible issue is that the job overlaps the resource, but I think the fundamental problem is that the job node is rendered in the resources column (or maybe it's the first column which happens to be for resources)

## Steps to reproduce

Using the following pipeline config:

```yaml
resources:
- name: a
  type: mock
- name: b
  type: mock

jobs:
- name: a
  plan:
  - get: a
- name: b
  plan:
  - get: a
  - get: b
- name: overlapping
```

## Expected results

UI is rendered normally

## Actual results

Screen Shot 2021-06-23 at 3 50 46 PM

## Additional context

I think the jobs are sorted at some point, just changing the names can result in a different graph

```yaml
jobs:
- name: c
  plan:
  - get: a
- name: b
  plan:
  - get: a
  - get: b
- name: overlapping
```

Screen Shot 2021-06-23 at 3 54 48 PM

## Triaging info

* Concourse version:
* Browser (if applicable):
* Did this used to work?

xtremerui:
Fixes #7207 

The .tug function will try to align input with its connecting job (to make the line straigh instead of bezier) while it doesn't consider the spacing of next node(job). This PR makes node in a rank group at least occupy 1 space even without any edge key (no input/output edge).

* [x] done
* [ ] todo

## Release Note

- Fixes an edge case that might overlap an input and job node in the pipeline view.

### DIFF
--- a/web/public/graph.mjs
+++ b/web/public/graph.mjs
@@ -514,7 +514,7 @@ RankGroup.prototype.layout = function() {
 
     node._keyOffset = rollingKeyOffset;
 
-    this.ordering.fill(rollingKeyOffset, node._edgeKeys.length);
+    this.ordering.fill(rollingKeyOffset, Math.max(node._edgeKeys.length, 1));
 
     rollingKeyOffset += Math.max(node._edgeKeys.length, 1);
   }


clarafu (approved):
Nice!! Tested it out manually and it fixes the bug :D

PR #7455, opened 2021-08-24
Validate if a Pipeline contains a cycle

EstebanFS:

Now we can detect if the pipelines contains a cycle before creating it.

* [x] Adding cycle detection on the validate pipeline method

## Notes to reviewer

if you execute `fly set-pipeline` with the following pipeline that contains a cycle

```yml
resources:
- name: get_1
  type: mock

jobs:
- name: step_1
  plan:
  - in_parallel:
      steps:
      - get: get_1
        passed:
        - step_1
```

you should see the following error `pipeline contains a cycle that starts at Job 'step_1'`

## Release Note

* The API will reject any pipelines that contains a cycle

### DIFF
--- a/atc/configvalidate/validate.go
+++ b/atc/configvalidate/validate.go
@@ -86,6 +86,12 @@ func Validate(c atc.Config) ([]atc.ConfigWarning, []string) {
 	}
 	warnings = append(warnings, displayWarnings...)
 
+	cycleErr := validateCycle(c)
+
+	if cycleErr != nil {
+		errorMessages = append(errorMessages, formatErr("jobs", cycleErr))
+	}
+
 	return warnings, errorMessages
 }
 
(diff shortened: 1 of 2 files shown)


atc/configvalidate/validate.go
```diff
@@ -490,3 +496,62 @@ func validateDisplay(c atc.Config) ([]atc.ConfigWarning, error) {
 
 	return warnings, nil
 }
+
+func detectCycle(j atc.JobConfig, visited map[string]int, initialJobConfig atc.Config) (bool, string) {
```
aoldershaw, line 500: Maybe rename `initialJobConfig` to `pipelineConfig`

atc/configvalidate/validate.go
```diff
@@ -490,3 +496,62 @@ func validateDisplay(c atc.Config) ([]atc.ConfigWarning, error) {
...
+	state := false
+	cycleErrorJob := ""
+	_ = j.StepConfig().Visit(atc.StepRecursor{
+		OnGet: func(step *atc.GetStep) error {
+			for _, nextJobName := range step.Passed {
+				nextJob := findJobByName(nextJobName, initialJobConfig.Jobs)
+				if visited[nextJobName] == semiVisited {
+					state = true
+					cycleErrorJob = nextJobName
+					return nil
```
aoldershaw, line 516: Not a big deal, but WDYT about changing the signature of `detectCycle` to just return an `error`. If we detect a cycle here, we can just return the error we currently construct here: https://github.com/concourse/concourse/blob/c948550fecc2a8df7fd178364aed7d145b15cc80/atc/configvalidate/validate.go#L554

Then, we can simply return the `error` that's returned by `j.StepConfig.Visit(...)`, and don't need to keep track of `state` or `cycleErrorJob`, simplifying the control flow a bit. Plus, it wasn't totally obvious to me what `state` referred to initially - but returning an `error` makes it more obvious that this is where the cycle occurs

atc/configvalidate/validate.go
```diff
@@ -490,3 +496,62 @@ func validateDisplay(c atc.Config) ([]atc.ConfigWarning, error) {
...
+
+func validateCycle(c atc.Config) error {
+	jobs := c.Jobs
+	cycleExist := false
+	cycleErrorJob := ""
+	emptyMap := make(map[string]int)
+	for _, job := range jobs {
+		cycleExist, cycleErrorJob = detectCycle(job, emptyMap, c)
+		if cycleExist == true {
+			break
```
aoldershaw, line 549: Similarly here, maybe we can just return the error here rather than `break`ing, so we can scope `cycleExist` and `cycleErrorJob` to within the `for` loop (or if you decide to return an `error` from `detectCycle` instead, this just becomes `if err != nil { return err }`)

aoldershaw (changes requested):
Had to sketch it out to understand how it works, but it's a neat algorithm!

I left some suggestions for making the control a bit simpler to follow, at least IMO. Let me know if you have any questions about it, and feel free to push back if you disagree

aoldershaw (approved):
LGTM! Thanks for making the changes I requested

PR #7459, opened 2021-08-25
Prevent open redirect to other hosts

taylorsilva:

We fell victim to this behaviour: https://github.com/golang/go/issues/38642

The solution in this commit ensures that the URI we redirect to is
always a path. The TrimLeft() removes all prefixed forward slashes,
which covers the case where a redirect uri starts with two or more
slashse.

We then always add a single forward slash which will ensure that
ParseRequestURI() interprets the URI as a path.

Now if someone visits `http://ci.concourse-ci.org/sky/login?redirect_uri=//google.com/`
they'll end up redirected to `http://ci.concourse-ci.org/google.com` which
will return a 404 and redirect one more time to the homepage.

In case the behaviour of ParseRequestURI() changes or is smarter in the
future, we added a check to ensure that if either Host or Scheme is
populated that the URI is not used for redirecting. The tests have also
been updated so this change in behaviour will cause the tests to fail.

## Release Note

* Prevent an open redirect vulnerability on the `/sky/login` path

### DIFF
--- a/skymarshal/skyserver/skyserver.go
+++ b/skymarshal/skyserver/skyserver.go
@@ -160,14 +160,14 @@ func (s *SkyServer) Callback(w http.ResponseWriter, r *http.Request) {
 func (s *SkyServer) Redirect(w http.ResponseWriter, r *http.Request, oauth2Token *oauth2.Token, redirectURI string) {
 	logger := s.config.Logger.Session("redirect")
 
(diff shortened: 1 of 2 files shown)


chenbh (approved):
lgtm!

It took a while for me to figure out that this particular url is only generated if you try to access a page while not being logged in. The redirect is to go back to the page after login in rather than as part of the auth flow.

PR #7461, opened 2021-08-25
Delete btrfs volume if it exists when using the overlay driver

### LINKED ISSUE #7430: Worker does not recover if btrfs volume is still mounted
## Summary

If a worker is initially deployed using the btrfs driver and then switches to overlay as the driver, there's nothing to remove the btrfs mount (to workdir/volumes), which seems to cause issues with the overlay driver.

## Steps to reproduce

1. Deploy concourse with `CONCOURSE_BAGGAGECLAIM_DRIVER: btrfs`
2. Set a pipeline with a resource and run a check
3. Re-deploy concourse with `CONCOURSE_BAGGAGECLAIM_DRIVER: overlay`
4. Run `mount`

**NOTE:** You can't use `docker-compose` for these deployments as the container would be recreated at this point, getting rid of all of the mounts

## Expected results

The btrfs mount to `/work-dir/volumes` is gone

## Actual results

It remains after deploying with overlay, causing issues

## Additional context

We observed this in our [bin-smoke](https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/bin-smoke/builds/648.1#L611b42e7/0/3:23) job. The worker process reported the following errors:

```
"worker.garden.garden-server.run.failed","data":{"error":"proc start: OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: read init-p: connection reset by peer: unknown"
```

```
error":"run check: check: exec failed: container_linux.go:380: starting container process caused: exec: \"/opt/resource/check\": stat /opt/resource/check: no such file or directory
```

When we took a look at the rootfs volume, we noticed that the lower dir (in this case, the volume for the `mock` base resource type) was empty, hence the missing `/opt/resource/check` binary.

Also, worth noting that the concourse chart manually clears out any existing btrfs volumes on startup: https://github.com/concourse/concourse-chart/blob/master/templates/worker-statefulset.yaml#L73-L76

## Triaging info

* Concourse version: 7.4.0
* Did this used to work? No

taylorsilva:

closes #7430

We found that if you ended up in a state where you were trying to run a
worker with overlay and created the overlay mounts intside a btrfs
mount, lots of weird stuff would happen with volume creation and COW
volumes specifically.

This commit ensures that if you're using overlay and the volume path is
a btrfs mount, then the btrfs mount is cleaned up.

## Release Note

* Made worker initialization more stable if you're switching from btrfs to overlay. The worker will remove the btrfs mount if it exists before creating overlay mounts

### DIFF
--- a/worker/baggageclaim/baggageclaimcmd/driver_linux.go
+++ b/worker/baggageclaim/baggageclaimcmd/driver_linux.go
@@ -20,8 +20,8 @@ import (
 const btrfsFSType = 0x9123683e
 
 func (cmd *BaggageclaimCommand) driver(logger lager.Logger) (volume.Driver, error) {
-	var fsStat syscall.Statfs_t
-	err := syscall.Statfs(cmd.VolumesDir.Path(), &fsStat)
+	var volMountInfo syscall.Statfs_t
+	err := syscall.Statfs(cmd.VolumesDir.Path(), &volMountInfo)
 	if err != nil {
 		return nil, fmt.Errorf("failed to stat volumes filesystem: %s", err)
 	}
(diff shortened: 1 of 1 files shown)


aoldershaw (commented):
In the chart we iterate over all of the subvolumes and delete them individually - is this necessary here too?

aoldershaw (approved):

PR #7463, opened 2021-08-26
indicate if a pipline is archived in pipeline view

### LINKED ISSUE #7163: Visual Indicator of Archived Pipeline
## Summary

Currently, if an archived pipeline is linked to, there's no visual indication that it is archived. 

There is only a lack of a "run job" play button, but everything else looks as if it is a normal operational pipeline.

The only minor differentiator otherwise is its absence from the sidebar.

Some type of visual indication of a pipeline that it is indeed archived would be helpful, since it still loads in the GUI if a link is provided, which is common in README's, and other documentation.

xtremerui:

* add "archived" to pipeline name
* change breadcrumbs background to grey

closes #7163

* [x] done

## Release Note

When viewing an archived pipeline (or any sub routes of it) in UI, the pipeline name now shows "archived" and the breadcrumbs background will change to grey so one won't confuse.

### DIFF
--- a/web/elm/benchmarks/Benchmarks.elm
+++ b/web/elm/benchmarks/Benchmarks.elm
@@ -133,7 +133,6 @@ buildView session model =
         [ Html.div
             (id "top-bar-app" :: Views.Styles.topBar False)
             [ SideBar.sideBarIcon session
-            , TopBar.concourseLogo
             , breadcrumbs session model
             , Login.view session.userState model
             ]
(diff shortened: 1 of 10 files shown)


clarafu:
I tested it out locally and the archived text looks great! I just noticed something small that I want your opinion on!

For archived pipelines, the grey header cuts off at the concourse icon on the top left

Screen Shot 2021-08-31 at 11 34 05 AM

While a paused pipeline the header extends past the concourse icon on the top left

image

Do you think they grey header should also extend past the concourse icon to keep it consistent?

xtremerui:
Good call I think that is more consistant. I will update it so the concourse icon also grey out when pipeline is archived.

xtremerui:
@clarafu i have upated the PR so the paused or archived state will also change the concourse logo. And also the pipeline state will persisted across all its sub routes etc. job, resource, builds.

clarafu (approved):
Just did some acceptance testing and it looks great!

Looks like there's some elm unit tests that need some fixing though https://ci.concourse-ci.org/builds/256387462

clarafu (approved):

PR #7487, opened 2021-08-30
worker: baggageclaim emits spans

### LINKED ISSUE #7358: Add spans to baggageclaim

taylorsilva:

closes #7358

Adds tracing to the worker generally and leverages it in the baggageclaim server.

## Notes to reviewer

All the commits under the `atc` package are for threading `ctx`'s down to the baggageclaim client calls. Nothing structurally changed in the `atc` package.

To start a Concourse with tracing:
```
docker-compose -f docker-compose.yml -f ./hack/overrides/jaeger.yml up -d --build
```
Then visit http://localhost:25000 (may take a few seconds to come up) to view traces. If you run a job you'll see a trace like this:

image

It's been interesting to see all this extra time on the web side compared to how fast the worker side is. Not sure what's up with that but it's very consistent, even for smaller volumes.

## Release Note

* Workers now emit traces from the baggageclaim server so one can see volumes being created and streamed as part of a build


EstebanFS:
Looks great so far!

EstebanFS:
Is there any particular reason why we don't add StartSpan to the DestroyVolumen method on `worker/baggageclaim/volume/repository.go`?

EstebanFS (approved):
LGTM!

taylorsilva:
@EstebanFS I did initially add `StartSpan()` to all the functions in this interface: https://github.com/concourse/concourse/blob/975b0439465e0036c342e59b3c026c0eaf7edcbd/worker/baggageclaim/volume/repository.go#L29-L44
I removed `StartSpan` from the functions that weren't super important or used at all during container initialization. The reduces the amount of traces generated, which can be a lot really fast.

PR #7494, opened 2021-09-01
Clear resource config scope ID if source field changes

### LINKED ISSUE #7468: Wrong image tag (registry-image) used after being updated
## Summary

In production we were experiencing situations where our pipelines would fail after we updated the relevant tags on a registry-image resource. We were deleting and re-creating the pipeline to resolve the issue but started to investigate why. 

## Steps to reproduce

We have tested these steps in our cluster, but also on a local, fresh docker-compose (concourse-docker) cluster. The commit SHA's are correct at time of writing, but, may not be the latest (version_depth=10) when this is used. The purpose of the commit SHA's it to trigger the pipeline when we update it.

- Set pipeline with the below configuration
```
resources:
  - name: image
    type: registry-image
    source: { repository: golang, tag: "1.16" }

  - name: git
    type: git
    source:
      branch: master
      version_depth: 10

jobs:
  - name: use-image
    plan:
      - in_parallel:
          - get: image
          - get: git
            trigger: true
            version: { ref: 1e24cfceeefd388d8c993ab9b26c71d05642579a}
      - task: task
        image: image
        config:
          platform: linux
          run:
            path: go
            args: ["version"]
``` 

- unpause and await build 1 to run
- update the pipeline, with a new golang image (1.13) and a new commit ref to cause the job to run
```
resources:
  - name: image
    type: registry-image
    source: { repository: golang, tag: "1.13" }

  - name: git
    type: git
    source:
      branch: master
      version_depth: 10

jobs:
  - name: use-image
    plan:
      - in_parallel:
          - get: image
          - get: git
            trigger: true
            version: { ref: 212509f7c8fd705f9dd7a26b41638ddfd8f744ba}
      - task: task
        image: image
        config:
          platform: linux
          run:
            path: go
            args: ["version"]

```

## Expected results

Build 2 is run with golang image 1.13

## Actual results

Build 2 is run with golang image 1.16

## Additional context

image

image

## Triaging info

* Concourse version: v7.4.0 (concourse-docker)
* Concourse version: v7.3.2 (production)

We have tried this with pr-resources instead of git and AWS ECR images instead of GoLang in docker hub and experienced the same behaviour.

taylorsilva:

closes #7468

when a resources config updates it will eventually point to a new
version history, which is tied to the resource through the config_id and
config_scope_id columns.

When a resource config is initally updated the relationship between the
new and old config is not made until a check for the resource is run.
This results in a period where the algorithm may run and use a version
from the previous resource's config.

This sql trigger will set the config id's to NULL for any resources who
has their config updated. If a resource's config is not updated then the
config id columns are left as-is.

## Notes to Reviewer

- The jsonb type was introduced in postgres 9.4 and we support 9.5+
  - https://www.postgresql.org/docs/9.4/datatype-json.html
- The third commit removes a sql statement that came from the resource renaming PR (`old_name`). It was added at the same time as the test I updated the descriptions for was. It's unclear to me why this query is necessary. I don't know why we would want to remove specifically the config_id's of inactive resources. I don't think removing that saves us resources in anyway, afaik.

### DIFF
--- a/atc/db/migration/migrations/1630514494_clear_resource_configs_on_update.down.sql
+++ b/atc/db/migration/migrations/1630514494_clear_resource_configs_on_update.down.sql
@@ -1,0 +1,4 @@
+
+DROP TRIGGER IF EXISTS resources_config_update_clears_config_ids_trigger ON resources;
+
+DROP FUNCTION IF EXISTS clear_resource_configs_ids();
--- a/atc/db/migration/migrations/1630514494_clear_resource_configs_on_update.up.sql
+++ b/atc/db/migration/migrations/1630514494_clear_resource_configs_on_update.up.sql
@@ -1,0 +1,14 @@
+
+CREATE OR REPLACE FUNCTION clear_resource_config_ids() RETURNS TRIGGER AS $$
+BEGIN
+        EXECUTE format('UPDATE resources SET resource_config_id = NULL, resource_config_scope_id = NULL WHERE id=%s', NEW.id);
+        RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+DROP TRIGGER IF EXISTS resources_config_update_clears_config_ids_trigger ON resources;
+CREATE TRIGGER resources_config_update_clears_config_ids_trigger
+	AFTER UPDATE on resources
+	FOR EACH ROW
+	WHEN ((NEW.config IS DISTINCT FROM OLD.config) AND (NEW.active IS TRUE))
+	EXECUTE PROCEDURE clear_resource_config_ids();
--- a/atc/db/migration/migrations/1630514494_clear_resource_configs_on_update.up.sql
+++ b/atc/db/migration/migrations/1630514494_clear_resource_configs_on_update.up.sql
@@ -10,5 +10,5 @@ DROP TRIGGER IF EXISTS resources_config_update_clears_config_ids_trigger ON reso
 CREATE TRIGGER resources_config_update_clears_config_ids_trigger
 	AFTER UPDATE on resources
 	FOR EACH ROW
-	WHEN ((NEW.config IS DISTINCT FROM OLD.config) AND (NEW.active IS TRUE))
+	WHEN ((NEW.config::jsonb->'source' IS DISTINCT FROM OLD.config::jsonb->'source') AND (NEW.active IS TRUE))
 	EXECUTE PROCEDURE clear_resource_config_ids();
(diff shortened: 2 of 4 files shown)


clarafu:
I'm wondering if there is a reason we didn't just clear the `resource_config_scope_id` along with the `resource_config_id` when we set a pipeline in `db/team.go`? 🤔  Because if that do that then it would fix this bug right? Is there a reason you would prefer using a trigger to clear the ids versus clearing it using a query?

taylorsilva:
I initially tried adding it to `saveResource`: https://github.com/concourse/concourse/blob/de4b0b2dbfba23182705737d38a08e2c051ebda3/atc/db/team.go#L1245
But from what I could tell it would clear it out the config id's for all resources of the pipeline even if their `source` didn't change. I didn't see an easy way to filter for only the resource's whose `source` changed without making more db calls. Did I miss something here though? I didn't look beyond the `saveResource` and `saveResources` functions.

I did try clearing out the config id's on this line: https://github.com/concourse/concourse/blob/de4b0b2dbfba23182705737d38a08e2c051ebda3/atc/db/team.go#L1261
but like I said, it resulted in the id's for all resource's being cleared. If the `source` did not change the resource got it's same config id's assigned to it again after a check. The con with this approach was that every resource needed to be checked again, which makes things feel kinda slow.

On a whim I decided to see if it was possible to do it as a trigger in postgres, thus this solution.

clarafu:
Ahhh yeah you're right, I missed that the updating `resource_config_id` query to NULL did it for all resources in the pipeline. I don't see a problem with this method then!

Just to discuss our options though with using a trigger vs completing this within go (which will probably require another query), I think theres trade offs for both. For using a trigger the tradeoff would be the maintainability where it is sort of invisible in the code that this trigger exists. And for clearing out the ids in the code that would mean more queries. I don't really have a preference, I think both methods are good!

taylorsilva:
I agree with all your points.

My reason for going with the trigger approach was to avoid adding another query. I'm not super familiar with all the query optimizations you and Aidan have done but I wanted to avoid adding any queries to things where they weren't 1000% necessary. I wouldn't consider pipeline saving a very hot path so I would estimate that adding one extra query wouldn't be horrible for db performance. 

I think I have a compromise solution, cause right now I do also see this trigger as a kind of "magic" happening the db that someone only looking at the code won't notice. I'll add a test for this trigger in the team_test, which at least codifies that we expect this behaviour. I think that would help with the maintainability aspect, wdyt?

taylorsilva:
Added a test. It fails when I remove the trigger from the list of migrations :)

clarafu:
@taylorsilva Sorry for the super late review! Oh one other thing that can be changed and we should reset the `resource_config_id` and the `resource_config_scope_id` is the `type`. So if either the `type` or the `source` of a resource changes then we should clear our the ids.

taylorsilva:
@clarafu done!

clarafu (approved):
Cool! Works great :D 👍

Issue #4279, opened 2019-08-14
Show 'recently viewed' pipelines first on the dashboard


vito:
I find myself accidentally going to the dashboard pretty often when I'm actually trying to go back to the pipeline for the build I'm viewing. When I land on the dashboard it can be kind of jarring as it can be hard to find where I was if there are a bunch of pipelines configured.

If we showed 'recently viewed' pipelines somewhere prominent it'd be easier to end up where I want to be without having to go back and then click the pipeline in the breadcrumb.

Related: https://github.com/concourse/concourse/issues/2647 which proposes manually marking pipelines as a 'favorite'. This would help a lot too, assuming I favorited the pipeline.

Lindsayauchin:
## Design Proposal 

Add a category for 'recently viewed' to the top of the list. 

![Screen Shot 2019-08-19 at 3 56 48 PM](https://user-images.githubusercontent.com/27746614/63295163-feac4b80-c299-11e9-941a-598b8e67ff6a.png)

taylorsilva:
# Issue Bankruptcy

This issue is being closed as part of 2025's **Issue Bankruptcy** declaration. [See here for details](https://blog.concourse-ci.org/posts/2025-01-06-concourse-2024-review/)

If you think this issue still has merit then feel free to open a new issue and reference this one.

Issue #6590, opened 2021-02-24
View Switcher


matthewpereira:

taylorsilva:
# Issue Bankruptcy

This issue is being closed as part of 2025's **Issue Bankruptcy** declaration. [See here for details](https://blog.concourse-ci.org/posts/2025-01-06-concourse-2024-review/)

If you think this issue still has merit then feel free to open a new issue and reference this one.

---

This would have been so cool to have 😭

PR #7506, opened 2021-09-05
web: behavior: Add "recently viewed pipelines" section to the sidebar

multimac:

relates to #4279

Adds a section to the sidebar listing the last 5 pipelines (or instances pipelines) that a user has visited. It doesn't include them on the dashboard, only in the sidebar, but should help navigation as I find myself at times jumping between 2 or 3 different pipelines to keep an eye on them.

image

* [x] UI work
* [ ] Tests

## Notes to reviewer

Just looking for thoughts and feedback at the moment. This change has borked a number of tests in `SideBarFeatures.elm` which I will need to fix up (and add relevant tests for the new "recently viewed" section).

While the "favorites" section could also be used to solve the problem of jumping between a couple of different pipelines, it requires the user to remember to favorite each pipeline they visit, and unfavorite them when they no longer need to bounce between them.

## Release Note

Sidebar now shows the 5 most recently visited pipelines.

### DIFF
--- a/web/elm/src/Application/Application.elm
+++ b/web/elm/src/Application/Application.elm
@@ -87,6 +87,26 @@ init flags url =
                 { isOpen = False
                 , width = 275
                 }
+            , recentlyViewed =
+                let
+                    toPipelineIdentifier =
+                        \p -> { teamName = p.teamName, pipelineName = p.pipelineName, pipelineInstanceVars = p.pipelineInstanceVars }
+                in
+                case route of
+                    Routes.Pipeline { id } ->
+                        [ id ]
+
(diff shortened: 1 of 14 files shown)


scottietremendous:
Hi @multimac! Having a look into this and I still think it still veers a bit too much into the same territory covered by Favourite Pipelines. We have had discussion about more functionality we can add to the sidebar view: https://github.com/concourse/concourse/issues/6590. Perhaps this is something we can investigate once we get to this epic of work. For now we'd probably err on the side of just leaving the favourite pipelines feature and place and not adding any more functionality.

Issue #7505, opened 2021-09-03
Build logs not deleted (RIP) - database size grows indefinitely since migration from 6.7.3 to v7.3.2


Berber31:
## Summary

For some jobs, build logs are not ripped even with `CONCOURSE_DEFAULT_BUILD_LOGS_TO_RETAIN` set to `10` and `CONCOURSE_MAX_BUILD_LOGS_TO_RETAIN` set to `30` since concourse migration from 6.7.3 to 7.3.2 (but not for every jobs)

## Steps to reproduce

- With concourse 6.7.3 trigger many job builds
- migrate to concourse 7.3.2 (with kubernetes init container)
- trigger many builds for the same job

## Expected results

- `CONCOURSE_DEFAULT_BUILD_LOGS_TO_RETAIN` set to `10` 
- `CONCOURSE_MAX_BUILD_LOGS_TO_RETAIN` set to `30`

A job triggered 184 times (with NO job.build_log_retention set) should show only the last 10 builds logs and build `#175` to `#1` logs should be RIP.

## Actual results

- Build logs `#1` to `#123` are RIP (OK)
BUT
- Build logs `#124` to `#184` are NOT RIP (and log could be displayed)

I expect build logs `#1` to `#174` to be ripped.

Database size was around 65Go (flat) and grows to 200 Go ( +135Go) in 1 month (since the concourse database migration)

**INFORMATION:** Concourse migration (database migration with the kubernetes init container) from 6.7.3 to 7.3.2 was done between build `#133` and `#134`

There is maybe other jobs with the same behavior, but don't know how to query it.

## Additional context

- Concourse 7.3.2 (migrated from 6.7.3 the 2021 july 21th )
- Run on kubernetes - deployed with concourse helm chart v15.5.1
- Concourse migration from 6.7.3 to 7.3.2 was done between build `#133` and `#134`
- Pipelines: 289
- Teams: 59

## Triaging info

* Concourse version: 7.3.2
* Browser (if applicable):
* Did this used to work?

taylorsilva:
This may be fixed by #7179 which is in 7.4.0. Could you try on 7.4.0 and let us know if the problem still exists?

Berber31:
> This may be fixed by #7179 which landed in 7.4.0. Could you try on 7.4.0 and let us know if the problem still exists?

we will try.

But do you have maybe some database query to run to try to understand the root cause (table not migrated correctly, columns not set, etc.) ?
[...]

PR #7530, opened 2021-09-08
atc/gc: make build reaper more robust

taylorsilva:

when looping over all the pipelines and jobs, if it ran into any errors
while parsing them the function would return and never finish checking
any remaining pieplines/jobs.

This commit ensures that all errors are logged but do not stop the
reaper from doing it's job for any remaining pipelines and jobs it
hasn't gone through yet

## Notes to reviewer

This might fix the issue #7505 but the OP noted they don't see the relevant logs to indicate that this is the problem they're facing.

## Release Note

* Make build log reaper more robust by not exiting early if it encounters an issue while iterating over pipelines/jobs. Before this change build logs for some pipelines could have accumulated endlessly even with a build retention policy

### DIFF
--- a/atc/gc/build_log_collector.go
+++ b/atc/gc/build_log_collector.go
@@ -1,11 +1,12 @@
 package gc
 
 import (
+	"context"
+	"time"
+
 	"code.cloudfoundry.org/lager"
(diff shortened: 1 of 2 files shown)


muntac (approved):
Nice catch! LGTM.

PR #7543, opened 2021-09-13
Close zstd reader in load var step

### LINKED ISSUE #7542: Memory leak in 7.5.0
## Summary

We have observed a memory leak in the unreleased 7.5.0 version of Concourse. This should not affect any currently released versions of Concourse (<= 7.4.0).

In our ci deployment, we have seen the memory increase at a steady pace until the web nodes restart. 

image

We also noticed that the goroutines were slowly increasing.

image

We also took a heap dump of the web nodes and saw a large portion of the memory being used by `zstd`.

image

We have a theory that the goroutines are increasing because of the load var step not closing the zstd reader which can end up leaking. 

https://github.com/concourse/concourse/blob/6a5d02890b66cfeaa9beca348f5331f700097a30/atc/exec/load_var_step.go#L141-L156

But we don't think this is the cause of the main memory leak because of a few factors. One is that we only observed this memory leak after we deployed ci with the 7.5.0 release branch and the bug in the load var step existed even in previous versions deployed to ci. Another reason is that in the metrics after we deployed ci with gzip instead of zstd, we observed the goroutines no longer increasing but the memory seems to be continuing to increase (it is still in the early stages of deploying so I could be wrong about this).
 
## Context

* RFC: concourse/rfcs#
* Prior discussion: #
* Depends on #
* Part of #

clarafu:

part of #7542

We observed a memory leak in our ci deployment with Concourse version 7.5.0 (unreleased). Most of the findings and information is documented in the issue #7542.

One of the potential problems we found was that the load var step does not close the file stream, which is the thing that closes the zstd reader. The zstd reader needs to be manually closed versus the gzip reader, which is closed for us through golang.

* [x] done
* [ ] todo


taylorsilva (approved):
LGTM - only concern I had was if we could end up calling `Close()` on a nil `stream` but doesn't like that should be the case if err is nil, as usual :)

clarafu:
Closed this because I accidentally made the change on top of master rather than `release/7.5.x` branch. The correct pr should be https://github.com/concourse/concourse/pull/7548.

PR #7641, opened 2021-10-04
7.4.x: add some DB optimizations

taylorsilva:

- [x] in-memory event ID's
- [x] only update resource config every minute
- [ ] ~don't update metadata if fetching from cache~
  - after reading over the code we determined this change isn't needed
- [x] close zstd in load_var_step to prevent memory leak

## Release Note

* DB optimizations
  * Increment an event ID in-memory instead of using a Postgres Sequence
  * Only update resource config if it hasn't been updated in a minute
  * Close zstd reader in load var step

### DIFF
--- a/atc/db/build.go
+++ b/atc/db/build.go
@@ -1775,26 +1775,6 @@ func newNullInt64(i int) sql.NullInt64 {
 	}
 }
 
-func (b *build) refreshEventIdSeq(runner sq.Runner) error {
-	var currentEventID sql.NullInt64
-	err := psql.Select("max(event_id)").
-		From(b.eventsTable()).
-		Where(sq.Eq{"build_id": b.id}).
-		RunWith(runner).
-		QueryRow().
-		Scan(&currentEventID)
-	if err != nil {
-		return err
-	}
-
-	var seqIDInit uint64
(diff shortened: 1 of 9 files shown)


taylorsilva:
We tried running some tests locally and found out that this one commit also relies on a bunch of other changes that aren't in 7.4.x as well. There's a lot we need to backport or we need to make the commit from scratch based on the current state of 7.4.x

muntac (changes requested):
Looks like the tests are failing on this [line](https://github.com/concourse/concourse/blob/35610af9f86663cd080399fadf7b5dc6a255fb3b/atc/db/build_factory_test.go#L443).

taylorsilva:
Test is failing because the `eventID` was added to the build struct. When the build is fetched via the build_factory the eventID is not read from the DB and is zero. This clashes with the `build4DB` we kept in-memory which has an `eventID` of 1 because of the call to `build.Finish()`.

This made me realize that when we create a build that may already have events we start with eventID zero. I'm guessing this not what we want and that we would end up overwriting events or events may not appear in order.

muntac (approved):
LGTM!

PR #7703, opened 2021-10-19
Simplify clearing resource config ids

### LINKED ISSUE #7697: Updating resources fail when db encryption is on
## Summary

Turn on db encryption with `CONCOURSE_ENCRYPTION_KEY`. Set a pipeline with a resource. Update the `source` of the resource and try setting the pipeline again. It will fail with
```
error: Unexpected Response
Status: 500 Internal Server Error
Body:
failed to save config: pq: invalid input syntax for type json
```

Caused by https://github.com/concourse/concourse/pull/7494
Does not affect any released version of concourse.

## Triaging info

* Concourse version: `master`

taylorsilva:

closes #7697 and #7468

The config column in the resource table contains a json object with
three top-level keys: name, type, source

We only want to clear the config when type and source change. When those
two change the row itself is updated, so the job has the same ID
whenever the type or source changes.

When the name changes that results in a new row being inserted with a
new ID. In that case this trigger will not run. Therefore we can
simplify the code to simply compare the new and old configs regardless
of it being encrypted or not.

### DIFF
--- a/atc/db/migration/migrations/1630514494_clear_resource_configs_on_update.up.sql
+++ b/atc/db/migration/migrations/1630514494_clear_resource_configs_on_update.up.sql
@@ -11,10 +11,7 @@ CREATE TRIGGER resources_config_update_clears_config_ids_trigger
 	AFTER UPDATE on resources
 	FOR EACH ROW
 	WHEN (
-		(
-			(NEW.config::jsonb->'source' IS DISTINCT FROM OLD.config::jsonb->'source') OR
-			(NEW.config::jsonb->'type' IS DISTINCT FROM OLD.config::jsonb->'type')
-		)
-		AND (NEW.active IS TRUE)
+		(NEW.config IS DISTINCT FROM OLD.config) AND
+		(NEW.active IS TRUE)
 	)
 	EXECUTE PROCEDURE clear_resource_config_ids();


taylorsilva:
The test failing is for when a resource is renamed it should retain its version history. This goes against the assumption I made that a change in the `config` column means the `type` or `source` changed. Many other things can cause that column to change.

Another thing I discovered is that all resources fields are saved into the `config` column. This is really annoying because now any change, even a change like update the `public` or `check_every` fields would cause the version history to clear. So annoying!

I would like to do a refactor where all those fields get their own column instead of all being stuffed in a json object in the `config` column. The `name` and `type` fields are already duplicated. We would have to get rid of the easy json unmarshalling that we do here: https://github.com/concourse/concourse/blob/f51bb4fe1bc8b111c43fb0bc1cc5527b676037ce/atc/db/resource.go#L863-L872
but that doesn't seem too bad imo. It would end up being a bunch of lines mapping values to their correct fields in the `ResouceConfig` struct 🤷 

Then the `config` column or maybe a new `source` column would represent JUST the `source` field of a resource, making it easier to detect this update.

This would also potentially make it easier to run queries seeing how many active resources one has, which resources are public, etc. Not sure if there's any real use case with that but at least this would make it easier to query for such information.

One big question I have is **why did we stick everything in a json blob in one column instead of starting with separate columns?**

PR #7726, opened 2021-10-25
[skip-migrations-check] Revert "Clear resource config scope ID if source field changes"

taylorsilva:
Reverts concourse/concourse#7494


muntac (approved):
LGTM

PR #7736, opened 2021-10-27
Refresh event id within build reload

### LINKED ISSUE #7683: duplicate key value violates unique constraint "pipeline_build_events_22_build_id_event_id"

clarafu:

closes #7683

We reload the db build object being starting the build within the engine. By refreshing the event id sequence here, we will load up the event id into memory before multiple steps try to increment the event id in order to create events. Without refreshing the event id in the reload, we might run into cases where concurrent steps will try to both reload the event id and end up trying to save their events using the same event id.

* [x] done
* [ ] todo

## Release Note

* 

### DIFF
--- a/atc/db/build.go
+++ b/atc/db/build.go
@@ -378,15 +378,31 @@ func (b *build) RerunNumber() int      { return b.rerunNumber }
 func (b *build) CreatedBy() *string    { return b.createdBy }
 
 func (b *build) Reload() (bool, error) {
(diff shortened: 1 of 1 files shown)


muntac (approved):
LGTM!

PR #7758, opened 2021-11-02
Fixes for bugs in 7.4.2

### LINKED ISSUE #7683: duplicate key value violates unique constraint "pipeline_build_events_22_build_id_event_id"

clarafu:

closes #7683

There were a few bugs reported by users for 7.4.2 and this PR hopefully addresses them.

The way that we are doing the in memory event ID and the updating of the last referenced in the resource config is refactored from master branch. I think if this works properly then we might consider porting this refactor to master as well. But if you think the risk is high in the changes that were made, we can get the same optimization benefits without the code refactor.

* [ ] done
* [ ] todo

## Notes to reviewer

One of the bugs that were reported was `save image get event: pq: duplicate key value violates unique constraint "pipeline_build_events_22_build_id_event_id"`.

I had a suspicion that it might have something to do with this https://github.com/concourse/concourse/blob/b8f9da5b8c00b6112a7a334b3dccdc49962859f9/atc/db/build.go#L1905 because just checking `0` will result in multiple goroutines still finding that the `eventID` is `0` and trying to refresh all together (even after we did `refreshEventId` in `Reload()`). We also don't update the `eventID` here https://github.com/concourse/concourse/blob/b8f9da5b8c00b6112a7a334b3dccdc49962859f9/atc/db/build.go#L1936 or fetch here https://github.com/concourse/concourse/blob/b8f9da5b8c00b6112a7a334b3dccdc49962859f9/atc/db/build.go#L1905 using the `sync/atomic` package 

I'm still trying to reproduce this issue but not having much luck yet :( It would be best if we are able to reproduce this locally so we can properly test against it.

Currently master branch is using https://github.com/concourse/concourse/blob/45621cb6dfd46ad9d75f822fb8b2d083913e7aaf/atc/util/seq_generator.go#L5-L7 to handle the concurrency. We could just switch back to that if we think it does have to do with the `atomic` package. Personally it is my first time using the package and Aidan and I were hoping we could use that instead of the `seq_generator` as a refactor, but maybe the risks are too high for us to use it.

The other bug was `create resource config: pq: deadlock detected","pipeline` which seems to be coming from Postgres and is probably related to the change that was made with updating the last referenced column for resource configs. In #7641, we made it so that we will SELECT a row from resource configs and then update the row after in order to update the last referenced column. This made it so that we introduced an area where a deadlock can occur between the time that we grabbed the `FOR SHARE` lock during the SELECT and the time that we need to grab the row lock for the update. I'm wondering if the `FOR SHARE` lock during the SELECT is really even necessary because removing it will remove the possibility of the deadlock. I'm still not 100% if this is the right thing to do so I will try looking into this a bit further. 

## Release Note

* 

### DIFF
--- a/atc/db/build.go
+++ b/atc/db/build.go
@@ -252,7 +252,7 @@ type build struct {
 
 	spanContext SpanContext
 
-	eventID uint64
+	eventIDCounter SeqCounter
 }
 
 func newEmptyBuild(conn Conn, lockFactory lock.LockFactory) *build {
@@ -1918,13 +1918,15 @@ func (b *build) saveEvent(tx Tx, event atc.Event) error {
 		return err
 	}
 
-	if b.eventID == 0 {
+	if b.eventIDCounter == nil {
 		if err := b.refreshEventIdSeq(tx); err != nil {
 			return err
 		}
 	}
 
-	eventID := atomic.AddUint64(&b.eventID, 1) - 1
+	// Subtract 1 because Next returns the incremented event counter but we
+	// actually want to insert the current event id
+	eventID := b.eventIDCounter.Next() - 1
 	_, err = psql.Insert(b.eventsTable()).
 		Columns("event_id", "build_id", "type", "version", "payload").
 		Values(eventID, b.id, string(event.EventType()), string(event.Version()), payload).
@@ -1949,7 +1951,7 @@ func (b *build) refreshEventIdSeq(runner sq.Runner) error {
 	if currentEventID.Valid {
 		seqIDInit = uint64(currentEventID.Int64) + 1
 	}
-	b.eventID = seqIDInit
+	b.eventIDCounter = &seqCounter{sequenceID: seqIDInit}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


xtremerui:
would this PR gonna be fixing the `FOR SHARE` lock issue or there will be a separate PR?

xtremerui:
Looks like its good to go. Let's merge this and load it in stress testing env.

xtremerui (approved):

clarafu:
We tested this out on our drills env called `baseline`. We first deployed 7.4.2 on it and ran a bunch of checks, we ended up seeing the deadlock issue that users were reporting `create resource config: pq: deadlock detected`. We deployed a concourse rc with this pr merged in and observed that the deadlock errors no longer appear.

lulukas86:
@clarafu: Many thanks for the update. Can you tell me when the bugfix will be released?

clarafu:
@lulukas86 Sorry for the delay, 7.4.3 should be released within the next few days!

PR #7827, opened 2021-11-22
web: behavior: support stringable bool for GetStep.Trigger

lrstanley:

* [x] support stringable bool for GetStep.Trigger (allow unmarshalling "true", "True", "false", "False" into GetStep.Trigger)

## Notes to reviewer

PR is related to [this Discord conversation](https://discord.com/channels/219899946617274369/413770960089382922/910563817161904169).

In short, today, it's not possible to interpolate the trigger field of a get step, when the source of the interpolated variable comes from sources like `--load-vars-from`, var-sources, vault, `--var`, instance vars, across steps, etc. The only exception I know of so far is `--yaml-var` which could let it be a boolean, but this clearly can't be used in all situations. This PR solves that by defining a custom unmarshaler.

Example, `((some-variable))`:

```yaml
[...]
- in_parallel:
  - { get: timer, trigger: ((some-variable)) }
```

With `fly -t target set-pipeline --var=some-variable=False` (or `false`, doesn't seem to make a difference), it causes:

```
error: error unmarshaling JSON: while decoding JSON: malformed in_parallel step: failed to unmarshal parallel steps: malformed get step: json: cannot unmarshal string into Go struct field GetStep.trigger of type bool
```

One of the primary benefits with this change, is you can now pass in a value for trigger, through across/set_pipeline, which would allow scenarios like a multi-branch workflow where non-master branches auto-run deployments but master doesn't, without having a separate pipeline or pipeline file.

This is the example pipeline I used to test, where `time-triggered.yaml` uses `((instance.trigger))` ([as seen here](https://github.com/lrstanley/concourse-test-pipeline/blob/master/time-triggered.yaml#L14)):

```yaml
deployments: &deployments
  - name: example
    trigger: true
  - name: example-2
    trigger: false

resources:
  - name: git-branch
    type: git
    icon: git
    source:
      uri: https://github.com/lrstanley/concourse-test-pipeline.git
      branch: master

jobs:
  - name: update-deployments
    public: false
    serial: true
    plan:
    - in_parallel:
        fail_fast: true
        steps:
        - get: git-branch
          trigger: true
    - across:
      - var: instance
        values: *deployments
      do:
      - set_pipeline: "example"
        file: git-branch/time-triggered.yaml
        instance_vars:
          name: ((.:instance.name))
        vars:
          instance: ((.:instance))
```

I have yet to test var-sources or vault interpolation, though suspect this should work as intended.

## Release Note

* Add better interpolation support for `trigger` within `get` steps (across, set_pipeline, etc). This is useful for multi-branch style pipelines that need to toggle things like auto-deployments for specific environments, as an example.

### DIFF
--- a/atc/db/job.go
+++ b/atc/db/job.go
@@ -22,7 +22,7 @@ type InputConfigs []InputConfig
 
 type InputConfig struct {
 	Name            string
-	Trigger         bool
+	Trigger         atc.Trigger
 	Passed          JobSet
 	UseEveryVersion bool
 	PinnedVersion   atc.Version
@@ -278,7 +278,7 @@ func (j *job) AlgorithmInputs() (InputConfigs, error) {
 			Name:       inputName,
 			ResourceID: resourceID,
 			JobID:      j.id,
-			Trigger:    trigger,
+			Trigger:    atc.Trigger(trigger),
 		}
 
 		if pinnedVersionString.Valid {
@@ -366,7 +366,7 @@ func (j *job) Inputs() ([]atc.JobInput, error) {
 		inputs = append(inputs, atc.JobInput{
 			Name:     inputName,
 			Resource: resourceName,
-			Trigger:  trigger,
+			Trigger:  atc.Trigger(trigger),
 			Version:  version,
 			Passed:   passed,
 		})
--- a/atc/job.go
+++ b/atc/job.go
@@ -31,7 +31,7 @@ type Job struct {
 type JobInput struct {
 	Name     string         `json:"name"`
 	Resource string         `json:"resource"`
-	Trigger  bool           `json:"trigger"`
+	Trigger  Trigger        `json:"trigger"`
 	Passed   []string       `json:"passed,omitempty"`
 	Version  *VersionConfig `json:"version,omitempty"`
 }
--- a/atc/steps.go
+++ b/atc/steps.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 )
 
@@ -292,13 +293,27 @@ var StepPrecedence = []StepDetector{
 	},
 }
 
+// Trigger represents if a step can be triggered by another event. Trigger
+// also supports unmarshalling from a boolean represented as a string, e.g.
+// "true", "True", "1", etc.
+type Trigger bool
+
+func (b *Trigger) UnmarshalJSON(data []byte) error {
+	converted, err := strconv.ParseBool(strings.Trim(string(data), "\""))
+	*b = Trigger(converted)
+	if err != nil {
+		return fmt.Errorf("Trigger unmarshal error: %w", err)
+	}
+	return nil
+}
+
 type GetStep struct {
 	Name     string         `json:"get"`
 	Resource string         `json:"resource,omitempty"`
 	Version  *VersionConfig `json:"version,omitempty"`
 	Params   Params         `json:"params,omitempty"`
 	Passed   []string       `json:"passed,omitempty"`
-	Trigger  bool           `json:"trigger,omitempty"`
+	Trigger  Trigger        `json:"trigger,omitempty"`
 	Tags     Tags           `json:"tags,omitempty"`
 	Timeout  string         `json:"timeout,omitempty"`
 }
--- a/atc/steps_test.go
+++ b/atc/steps_test.go
@@ -46,6 +46,45 @@ var factoryTests = []StepTest{
 		},
 	},
 	{
+		Title: "get step with stringable trigger: true",
+		ConfigYAML: `
+			get: some-name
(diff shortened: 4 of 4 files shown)


lrstanley:
is anything else needed on this @clarafu?

clarafu:
@lrstanley Hey! Sorry for the delay in reviewing, I haven't been able to get to this yet but I will definitely spend time on it today/tomorrow.

clarafu:
If I understand this correctly, what you want to do is to be able to support vars in the `trigger:` field in the pipeline config. If this is correct, I have a concern on whether this works or not because we actually use the value of the `trigger` field in the scheduler component in the ATC https://github.com/concourse/concourse/blob/3f70d5f29c92af4b3c688c23cffaa05e415d6815/atc/scheduler/scheduler.go#L92

And in this scheduler component, we have no yet evaluated any of the vars because the var evaluation happens within the run of the build itself. (For example here is the var evaluation in a `get` step https://github.com/concourse/concourse/blob/3f70d5f29c92af4b3c688c23cffaa05e415d6815/atc/exec/get_step.go#L140-L143) 

So my suspicion is that this does not work because with `trigger` being a var, the scheduler component does not know whether the resource is trigger true or false and without knowing that, it does not know whether to schedule a build for this job.

lrstanley:
Sorry for the delay in getting back to this.

> So my suspicion is that this does not work because with trigger being a var, the scheduler component does not know whether the resource is trigger true or false and without knowing that, it does not know whether to schedule a build for this job.

From my testing, I can confirm that these scenarios do in fact work with this PR:

- var provided by fly command line (`--var` and `--yaml-var`).
- var provided by parameter file (`--load-vars-from`).
- instance vars when writing pipelines (and with `--load-vars-from`).
- instance vars where the attribute in the setter pipeline is pulling from vault.

I think this is because in all of the above situations, the string version of the field does get converted, and the underlying type is inlined. The only thing that isn't working, is a single pipeline, pulling from vault, which is the scenario that I believe you're referring to, because it keeps the dynamic vault reference.

Thoughts on how I can work through this? Either we don't want to do this because we don't want to add additional complexity in the scheduler, or maybe a few other ideas:

1. Implement this PR, which would allow more dynamic instance-var and across-step functionality, just not support pulling from vault (which I believe many things do today, so I don't think it's unexpected that the trigger field also wouldn't be similar).
2. Pull all dynamic variables in the scheduler (seems like a lot of overhead, esp. if they don't have any variables that would change the schedulers decisions).
3. add an `EvaluateScheduleDependent()` (or similar) method that is specific to **just** "scheduler dependent" dynamic variables. Ideally, this might only add additional work on the scheduler if folks actually do use dynamic values in the very specific fields needed by the scheduler.
   - The other issue I see is if this were functionality that could happen (I'm not super familiar with the scheduler, so just probing for ideas), would fetching via the UI also depend on this new evaluate function that replaces only scheduler-dependent values?
   - A benefit of having this means we could allow other fields to be more dynamic, that are limited in the same way `trigger` is limited (see below for an example).

Similar to the above but a little side track, I'd be curious on a future that would dynamically allow enabling/disabling (or pausing/unpausing) a job, based off dynamic variables. As our team gets further into the usage of instance vars, one of the biggest things we've struggled with is with enabling/disabling auto-deploy and similar functionality (e.g. dev auto-deploys/runs, and prod doesn't, without having to maintain 10 different pipeline files). and in the UI, it would show in a specific fashion in the UI, preventing it from being toggled because it's being controlled by an external source.

clarafu:
@lrstanley Right, the problem I pointed out is only for credential manager vars because the interpolated vars from fly are actually filled in right when you set pipeline and not dynamically like the vars from credential managers. 

From my understanding of what you're trying to achieve, your use case is to have the `set_pipeline` able to interpolate vars for `trigger` that does not need to use the dynamic var interpolation. I think your first option is probably the one I'm leaning towards, just because it is the least complex and is something I think will be useful. How I can imagine this working is that we allow for vars in the pipeline `trigger` field but maybe we can do some sort of validation that the user is also passing in the value for the `trigger` field in the `--load-vars-from` or `--vars` flags if they are using fly and if it is through `set_pipeline` we validate that the `trigger` field is filled in through the `across` step values or the `instance_vars`. This way we know for sure that the user isn't trying to use the `trigger` var dynamically. It should also be properly documented on the website about how you can't use a var for `trigger` with credential manager vars.

I think I would prefer just having it not able to do dynamic var interpolation because once we do allow that, it gets a bit more messy/complicated. Right now, all dynamic var interpolation happens within a build run itself so we don't need to do any pulling from credential managers if there are no builds to run. But if we allow for dynamic var interpolation to be a part of figuring out when to schedule a build, it would mean a lot more pulling from credential managers (since we attempt to schedule jobs every 10 seconds).

clarafu:
I'm just going to close this PR for now but you can re-open it whenever you want to keep working at this! Thanks!

PR #7836, opened 2021-11-24
Use CONCOURSE_CONTAINERD_NAMESPACE_SUFFIX for containerdNamespace.

### LINKED ISSUE #7165: Allow setting containerdNamespace using environment variable (CONCOURSE_CONTAINERD_NAMESPACE)
It's currently hardcoded to 'concourse' here: https://github.com/concourse/concourse/blob/94b52c866b1704d6167b9d121d0a443a960d9d09/worker/workercmd/containerd.go#L26

When running multiple Concourse's on the same host it'd be handy to be able to customize the namespace.

analytically:

closes #7165

## Release Note

Allow setting a namespace for Containerd.

### DIFF
--- a/worker/workercmd/containerd.go
+++ b/worker/workercmd/containerd.go
@@ -23,7 +23,7 @@ import (
 	"github.com/tedsuo/ifrit/grouper"
 )
 
-const containerdNamespace = "concourse"
+var containerdNamespace = "concourse" + os.Getenv("CONCOURSE_CONTAINERD_NAMESPACE_SUFFIX")
 
 // WriteDefaultContainerdConfig writes a default containerd configuration file
 // to a destination.


worker/workercmd/containerd.go
```diff
@@ -23,7 +23,7 @@ import (
 	"github.com/tedsuo/ifrit/grouper"
 )
 
-const containerdNamespace = "concourse"
+var containerdNamespace = "concourse" + os.Getenv("CONCOURSE_CONTAINERD_NAMESPACE_SUFFIX")
```
taylorsilva, line 26: Thanks for the PR!

This implementation makes it hard for anyone to discover that this flag even exists. Also what if someone likes setting all their stuff with CLI flags instead of env vars? Overall, we should probably add this alongside the rest of the configuration fields that Concourse provides.

Could you add this as a new field on this struct:
https://github.com/concourse/concourse/blob/da6b3b68bc56884498cce9be31f6231b1d384ef2/worker/workercmd/worker_linux.go#L40

Then instead of directly calling `os.GetEnv` you can pull the suffix from the `ContainerdRuntime` struct. Probably around here, where the namespace is passed along:
https://github.com/concourse/concourse/blob/da6b3b68bc56884498cce9be31f6231b1d384ef2/worker/workercmd/containerd.go#L74-L77

taylorsilva (changes requested):

worker/workercmd/containerd.go
taylorsilva: @analytically Do you need any help with making the changes?

clarafu:
I'm going to close this PR for now because of inactivity but feel free to re-open whenever you want to continue working on it! Thanks!

Issue #7857, opened 2021-12-02
Downgrade to 7.4.0 database (1625844436) fails


Eeems:
## Summary

When attempting to downgrade the database to 7.4.0 as instructed in the 7.4.3 release notes it fails.

## Steps to reproduce

Run `concourse migrate --migrate-db-to-version=1625844436` on a 7.4.2 database.

## Expected results

The migration should work and allow us to upgrade to 7.4.3.

## Actual results

```
error: could not migrate to version: 1625844436 Reason: migration '1625844437_drop_build_event_id_seq.down.sql' failed and was rolled back: pq: invalid input syntax for type integer: "select max(event_id) from pipeline_build_events_"
```

## Triaging info

* Concourse version: 7.4.2
* Did this used to work? No idea.

clarafu:
Thanks for reporting this issue, there is a bug in the downgrade script 

https://github.com/concourse/concourse/blob/a421efde8fd9e572f6084188039aca213347ef28/atc/db/migration/migrations/1625844437_drop_build_event_id_seq.down.sql#L17

it is missing the second `|` in that line after `pipeline_build_events_'`.

Really sorry for the inconvenience, it seems like our upgrade/downgrade tests did not properly test out this downgrade script.

Eeems:
Is there anything I can do to work around this, or will I have to wait for a new version to fix it?

clarafu:
I think I'm going to create a script that you can run against your database that will fix it and then you will be able to downgrade back to 7.4.0!

Eeems:
It would probably be a good idea for someone to update the release notes with a reference to this just in case others are planning to downgrade.

Berber31:
@clarafu
> I think I'm going to create a script that you can run against your database that will fix it and then you will be able to downgrade back to 7.4.0!

Where the *.sql are located in the concourse docker image ? in didn't found them with a `find / -name *.sql` in the docker image

Berber31:
@clarafu 
> I think I'm going to create a script that you can run against your database that will fix it and then you will be able to downgrade back to 7.4.0!

Do you plan to provide a 7.4.4 with a migrate binary fixed to downgrade database to 7.4.0 ?

cause we face off the same issue with our production environnement.

Thanks

AndoniLarz:
> @clarafu
> 
> > I think I'm going to create a script that you can run against your database that will fix it and then you will be able to downgrade back to 7.4.0!
> 
[...]

PR #7860, opened 2021-12-06
Fix down migration for build event sequences

clarafu:

#7857

There was a typo in the down migration for constructing build event sequences in postgres but it was never caught because the migration only runs for started builds during the downgrade but the integration tests don't cover that use case.

* [x] done
* [ ] todo

## Release Note

* This PR fixes a typo in a down migration that affects Concourse version 7.4.1, 7.4.2 and 7.4.3. Therefore, you will be unable to downgrade from any of those listed versions.

### DIFF
--- a/atc/db/migration/migrations/1625844437_drop_build_event_id_seq.down.sql
+++ b/atc/db/migration/migrations/1625844437_drop_build_event_id_seq.down.sql
@@ -4,7 +4,7 @@ declare
     startValue int;
 begin
     for b in
-        select id, name, pipeline_id, team_id from builds where status = 'started' and completed = false and aborted = false
+        select id, name, pipeline_id, team_id from builds where status = 'started' and completed = false
     loop
         raise notice 'dropping sequence build_event_id_seq_% ...', b.id;
         execute 'drop sequence if exists build_event_id_seq_' || b.id;
--- a/atc/db/migration/migrations/1625844437_drop_build_event_id_seq.down.sql
+++ b/atc/db/migration/migrations/1625844437_drop_build_event_id_seq.down.sql
@@ -4,7 +4,7 @@ declare
     startValue int;
 begin
     for b in
-        select id, name, pipeline_id, team_id from builds where status = 'started' and completed = false
+        select id, name, pipeline_id, team_id from builds where (status = 'started' OR status = 'pending') and completed = false
     loop
         raise notice 'dropping sequence build_event_id_seq_% ...', b.id;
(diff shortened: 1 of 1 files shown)


xtremerui (approved):

Berber31:
what do you think about adding some tests to cover thoses lines ?
as mentioned here: https://github.com/concourse/concourse/issues/7857#issuecomment-985606916

Berber31:
@clarafu 
i think the query is too restrictive

`select id, name, pipeline_id, team_id from builds where status = 'started' and completed = false and aborted = false`

i suggest to remove `aborted = false` or add `or end_time is null`

we could face off a build aborted after a migration to 7.4.1. So aborted column could be 'true', so sequence will not be recreated.

we have this situation actually

 status  | completed | aborted | end_time |    id    | build_number |                  job_name                  |             name             |   name   
---------+-----------+---------+----------+----------+--------------+--------------------------------------------+------------------------------+----------
 started | f         | t       |          | 70973927 | 17           | controle_validite_changement_en_production | mars-acquisitionrecouvrement | der-grea


with theses logs:
```

concourse-postgresql 2021-12-07 11:15:48.696 GMT [4011] ERROR:  relation "build_event_id_seq_70973927" does not exist at character 96                                                                           │
│ concourse-postgresql 2021-12-07 11:15:48.696 GMT [4011] STATEMENT:  INSERT INTO pipeline_build_events_337 (event_id,build_id,type,version,payload) VALUES (nextval('build_event_id_seq_70973927'),$1,$2,$3,$4)
```

Berber31:
> ecause this is such a specific use case that adding one for it in the



> @Berber31 I think the reasoning behind that is because if the build is aborted then we won't need to apply more events to that build, therefore we don't need to migrate that event sequence?

So if a build started on 7.4.0 then aborted in 7.4.2 (and migrate DB back to 7.4.0) we won't be able to abort this build and will run forever (with the yellow highlight still blinking on this job)

clarafu:
Hm I see, you mean the build is set as aborted but the build itself has not actually been successfully been aborted yet and still continues to have some output. I see your case then, I agree I will change the query. Thanks!!

clarafu:
@xtremerui Could you review it one more time! Just for a sanity check haha because I made some changes

xtremerui:
@clarafu just checked again I think it is fine to me.

jmccann:
Was wondering if anyone could post some notes to https://github.com/concourse/concourse/discussions/8298 regarding how to upgrade from 7.4.3 to 7.7.1 with these DB migrations in mind.

PR #7876, opened 2021-12-09
Automatically pause pipelines

taylorsilva:

Adds a component that runs once every 24hrs and checks for any pipelines that
haven't run in more than X days. Setting the days to zero disables the component.

Thought of adding this feature after reading this discord message:
https://discord.com/channels/219899946617274369/413771057686773770/912836796348497932
image

Where the user created some scripts to do this for their Concourse instance.
Their reported paused/unpaused ratio was at 50%. This made me think there's a
lot of Concourse pipelines out there doing resource checks when they could be
doing nothing and saving compute power. Thus this feature, to bring it to all
Concourse users.

The database already has all the info needed to find pipelines to pause, so
users can get a benefit from this feature immediately as it will retroactively
pause pipelines on its first run.

## Release Note

* Adds a new component that will automatically pause pipelines that have not run in more than the configured number of days. The number of days can be configured with `CONCOURSE_PAUSE_PIPELINES_AFTER`. A value of zero (the default) disables this component. On first run it will retroactively pause pipelines that already fall out of the given day range.
  * Exceptions: Pipelines that have no builds falling within the day range **but** have been updated in <24hrs will not be automatically paused. This is to cover the case where someone sets a new pipeline and hasn't run any jobs for it yet.


maxknee:
Oh this will be great to have, we do this manually.

taylorsilva:
testflight is flaky af 😡

xtremerui:
the runc bump in our main CI caused the failure. Now it is pinned to previous version. Once `dev-image` is rebuilt it should fix PR pipelines.

xtremerui (approved):
LGTM!

atc/db/pipeline_pauser.go
```diff
@@ -0,0 +1,92 @@
...
+	activePipelines, err := getActivePipelines(p.conn, daysSinceLastBuild)
+	if err != nil {
+		return err
+	}
+
+	rows, err := pipelinesQuery.Where(sq.And{
+		sq.Eq{
+			"p.paused": false,
+		},
+		sq.NotEq{
```
clarafu, line 37: @taylorsilva Hmm does something like `Where(sq.Expr("SELECT p.id FROM pipelines AS p LEFT JOIN jobs AS j ON j.pipeline_id = p.id WHERE j.last_scheduled > CURRENT_DATE - ?::INTERVAL", strconv.Itoa(days) + " day)).` not work? This would help us not need to pass the pipelines over from one query to another, which can sometimes be slow if there are a lot active pipelines.

clarafu, line 37: Hmm does something like Where(sq.Expr("SELECT p.id FROM pipelines AS p LEFT JOIN jobs AS j ON j.pipeline_id = p.id WHERE j.last_scheduled > CURRENT_DATE - ?::INTERVAL", strconv.Itoa(days) + " day)). not work? This would help us not need to pass the pipelines over from one query to another, which can sometimes be slow if there are a lot active pipelines.

clarafu (commented):
@taylorsilva Sorry for my super late review, had to finish up some bug fix for the 7.4.x series! I know it is merged already but I had one thing I wanted to run by you!

atc/db/pipeline_pauser.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// Couldn't put a subquery inside a WHERE clause: https://github.com/Masterminds/squirrel/issues/258
+// This is a workaround. I really tried to put this in as a subquery similar to
+// how it's done in worker_lifecycle.go but the placeholder value for the days
+// was never getting parsed correctly, even after figuring out that it also
+// needs to be cast to INTERVAL.
+// Maybe I was doing something stupid; there may be a way to get this in as a subquery
+func getActivePipelines(conn Conn, days int) ([]int, error) {
+	stmt, err := conn.Prepare(`SELECT p.id FROM pipelines AS p
```
clarafu, line 68: One more thing to note is that if we want to keep this existing implementation, we can add a `DISTINCT` to the `p.id` because right now it will select out multiple of the same pipeline id from the jobs in the left join that have the same pipeline id.

atc/db/pipeline_pauser.go
taylorsilva: The query you have there is what I started with but squirrel would not interpolate the `'? day'` part EVER. I rewrote that query SO MANY TIMES (nearly lost my sanity) but sadly ended up with the two-part query that got merged.

The only thing that makes me somewhat okay with this implementation is the fact that this query is only run once every 24hrs.

That said, I'll give your suggestion a try because I would also REALLY like to get rid of the second query and have it as a subquery instead.

taylorsilva: yay it worked! PR incoming 😄

xtremerui:
@taylorsilva would you be able to also add `CONCOURSE_PAUSE_PIPELINES_AFTER` to bosh and helm release?

taylorsilva:
@xtremerui already did. The PRs should be referenced somewhere above.

atc/db/pipeline_pauser.go
clarafu: Yaay!! HMM I wonder what made it work this time with the `? day` 🤔

taylorsilva: I think the problem was with the way I was writing the query with squirrel vs when I was manually figuring out the query in postico. What I kept trying to do was get squirrel/postgres to interpolate the value for me (you know, avoid sql injection). So I kept writing the query as `... '? day'::INTERVAL`.

What your suggestion was changed the query from `... '? day'::INTERVAL` to `...?::INTERVAL` and TADA everything worked. I was too hung up on trying to pass ONLY the int instead of a string that contained the int.

clarafu: Ohh I see! Cool thats good to know haha! Thanks for explaining :)

clarafu:
Hmm @xtremerui @taylorsilva I'm curious to know if you have tried running this manually and saw it pause pipelines that have not run in the configured amount of days? I'm thinking that using `last_scheduled` on the job might not work because the `last_scheduled` column is updated every time the job scheduler component runs for this job. Even if the scheduler has run for the job, that does not guarantee that a build was created for it. That means the `last_scheduled` can be updated constantly without any builds actually being created and ran for the job.

taylorsilva:
No, did not manually test this.

I get what you're saying, that `last_scheduled` does not mean a build ran, but I also don't understand when that would be there case where the job was scheduled but a build was not created? I'd assume something else is going horribly wrong if jobs are being scheduled and no build is actually being made.

Okay diving into the code more cause I don't know the scheduler stuff very well, and this bit interests me:
https://github.com/concourse/concourse/blob/53949cdabe5d7c5131b2ae948b077a663bb7cf87/atc/scheduler/scheduler.go#L34-L59

These are various cases where the scheduler/algorithm can error and return a bool that indicates whether a retry is needed (true) or not (false). I guess these are the cases you are talking about?

And then diving over to
https://github.com/concourse/concourse/blob/53949cdabe5d7c5131b2ae948b077a663bb7cf87/atc/db/job_factory.go#L78-L87

where yes, this column does not look like the data I want! I'll make changes in #7914

clarafu:
Oh the cases where the scheduler was run and a build is not created is pretty common! For example, it could be when a resource version is disabled and it will trigger a request to schedule the job.

https://github.com/concourse/concourse/blob/53949cdabe5d7c5131b2ae948b077a663bb7cf87/atc/db/resource.go#L781-L784

This disabled resource version can be a super old one and when the scheduler runs, it determines the same set of inputs as it did in the latest build and therefore does not create a new build. Here lists all the cases where a request to schedule will be done https://concourse-ci.org/scheduler.html#scheduling-on-demand.

One way I can think of to get this to work is we might want to have this pipeline pausing component determine when it needs to be paused based off the creation time of the latest build of its jobs?

PR #7914, opened 2022-01-04
Follow-up for #7876

taylorsilva:
Follow-up for #7876

Implement Clara's suggestion. Got rid of the stupid second query! Thanks @clarafu!

TODO: 
- [x] add integration test but my cat wants to play so for another time!

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -152,6 +152,7 @@ type RunCommand struct {
 	ResourceWithWebhookCheckingInterval time.Duration `long:"resource-with-webhook-checking-interval" default:"1m" description:"Interval on which to check for new versions of resources that has webhook defined."`
 	MaxChecksPerSecond                  int           `long:"max-checks-per-second" description:"Maximum number of checks that can be started per second. If not specified, this will be calculated as (# of resources)/(resource checking interval). -1 value will remove this maximum limit of checks per second."`
 	PausePipelinesAfter                 int           `long:"pause-pipelines-after" default:"0" description:"The number of days after which a pipeline will be automatically paused if none of its jobs have run in more than the given number of days. A value of zero disables this component."`
+	PipelinePauserInterval              time.Duration `long:"pipeline-pauser-interval" default:"24h" hidden:"true" description:"The frequency on which the Pipeline Pauser component will be run to check if any pipelines need to be paused."`
 
 	ContainerPlacementStrategyOptions worker.PlacementOptions `group:"Container Placement Strategy"`
 
@@ -1083,7 +1084,7 @@ func (cmd *RunCommand) backendComponents(
 		{
 			Component: atc.Component{
 				Name:     atc.ComponentPipelinePauser,
-				Interval: 24 * time.Hour,
+				Interval: cmd.PipelinePauserInterval,
 			},
 			Runnable: pauser.NewPipelinePauser(
 				dbPipelinePauser,
--- a/atc/db/dbfakes/fake_pipeline_pauser.go
+++ b/atc/db/dbfakes/fake_pipeline_pauser.go
@@ -2,16 +2,18 @@
 package dbfakes
 
 import (
+	"context"
 	"sync"
 
 	"github.com/concourse/concourse/atc/db"
 )
 
 type FakePipelinePauser struct {
-	PausePipelinesStub        func(int) error
+	PausePipelinesStub        func(context.Context, int) error
 	pausePipelinesMutex       sync.RWMutex
 	pausePipelinesArgsForCall []struct {
-		arg1 int
+		arg1 context.Context
+		arg2 int
 	}
 	pausePipelinesReturns struct {
 		result1 error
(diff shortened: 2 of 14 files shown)


taylorsilva:
@clarafu made a change to the query to not use `last_scheduled`

I have an idea for integration tests that I'll add soon.

taylorsilva:
Vague idea of running an integration test that relies on changing postgres's time:
* https://github.com/wolfcw/libfaketime
* Somehow restart postgres with that library installed and configured^

evanchaoli:
@taylorsilva I just read the original PR #7876, can you please add a info log when pauses a pipeline? When user reports any issue, for example, if a user escalates that his pipeline is pauses, the first we do would search for ATC logs for clues.

taylorsilva:
@evanchaoli That's fine, can add that. In that PR though, when this component pauses a pipeline it does show that it was paused by this component:
https://github.com/concourse/concourse/blob/312aa55db483ccd42a3d6ebb25fa1f40fb33c586/atc/db/pipeline_pauser.go#L52

So a user will see `Paused by automatic-pipeline-pauser <some-date>` in the header of their pipeline overview.

taylorsilva:
@clarafu My idea to use libfaketime worked for postgres but doesn't work for golang: https://github.com/wolfcw/libfaketime/issues/195
I can't think of any other way of adjusting the time on the golang side without doing some re-work on the component stuff to fake out time. And even bubbling that up to an integration level feels pretty sus. Any ideas on how to write this integration test?
What I did write up is on this branch: https://github.com/concourse/concourse/tree/followup-7876-integration-test/integration/pauser

---

I did end up finding other issues while building out the integration. From my latest commit:
```
This commit contains a failing test.

I realized that a pipeline that was set but never ran, like a newly set
pipeline, would result in the pipeline being paused. I wrote a test and
made a change to the query to fix that.

I then realized that this change makes it so any pipeline that has a job
with zero builds impossible for the pauser to ever pause the pipeline. I
added a test for this and it is the failing test in this commit. I
currently can't think of a query to get around this issue. Depending on
discussion this test or the first test may be removed.

For the first test I added I see this as kinda an edge case. I feel like
it's extremely unlikely that someone will set a pipeline AND unpause it
at the exact time the pauser runs, which is currently once every 24hrs.
I ran into this when trying to setup an integration test for the pauser
and found that the pauser kept pausing my newly set pipeline before I
could trigger my pipelines job. That's the context on how I found this
issue.
```

Any ideas for a query that can pass both tests?

evanchaoli:
> So a user will see `Paused by automatic-pipeline-pauser <some-date>` in the header of their pipeline overview.

That's cool, didn't realize that, while I think a log line will be still helpful for daily OPs.

clarafu:
Hmm I'm not too familiar with how to fake it out but one thing I can think (but might not be the best if you don't want to add more flags) is to add a flag that allows the user to configure the interval time for this pipeline pauser component. For example, we have the build tracker that runs every 10s on default but you can change that interval time https://github.com/concourse/concourse/blob/2edd994963235fcb6acd9ee15fbdb1f5494cfdb7/atc/atccmd/command.go#L201 I'm not sure if you would want to do that though?

taylorsilva:
yeahhh I had that thought earlier too. I think there's a flag for that library that lets me hide it too. That'll probably be the way to go for that.

taylorsilva:
@clarafu I DID IT! This was the biggest PIA test to write 😭 but it works! It passes!

This is good to review and imo good to ship.

atc/atccmd/command.go
```diff
@@ -152,6 +152,7 @@ type RunCommand struct {
 	ResourceWithWebhookCheckingInterval time.Duration `long:"resource-with-webhook-checking-interval" default:"1m" description:"Interval on which to check for new versions of resources that has webhook defined."`
 	MaxChecksPerSecond                  int           `long:"max-checks-per-second" description:"Maximum number of checks that can be started per second. If not specified, this will be calculated as (# of resources)/(resource checking interval). -1 value will remove this maximum limit of checks per second."`
 	PausePipelinesAfter                 int           `long:"pause-pipelines-after" default:"0" description:"The number of days after which a pipeline will be automatically paused if none of its jobs have run in more than the given number of days. A value of zero disables this component."`
+	PipelinePauserInterval              time.Duration `long:"pipeline-pauser-interval" default:"24h" hidden:"true" description:"The frequency on which the Pipeline Pauser component will be run to check if any pipelines need to be paused."`
```
taylorsilva, line 155: this flag is hidden so it doesn't need to be added to the chart and bosh release

taylorsilva:
uhhh what's up with the db tests failing in unit? It looks like too many are running in parallel?

taylorsilva:
Went back to that edge case and fixed it. Added a clause so it only pauses pipelines that were updated more than 24hrs ago.

atc/atccmd/command.go
clarafu: @taylorsilva Does it need to be hidden? Just asking because I could see the user wanting to adjust the interval as well!

clarafu (commented):
Wow thanks for all the work for adding the integration test for this!! :D

taylorsilva:
Good catch @clarafu 
Added a test for that case and updated the query to pass the test.

xtremerui:
https://github.com/concourse/concourse/issues/8029 @taylorsilva

clarafu (approved):
Thanks for all the work on the tests and fixes!! I just ran some acceptance testing against it and everything works great :D 

These were the cases that I covered in my manual testing:
- pipeline was paused for job that had a build older than configured pause pipelines after interval
- pipeline was not paused when they had a running/pending build
- pipeline was not paused when configured within 1 day

PR #7922, opened 2022-01-05
Do not send check build events to syslog drainer

### LINKED ISSUE #7915: Resource check creates overflow logs files in external syslog server
## Summary
While Concourse syslog drainer is enabled, users notice huge amount of logs files are created in the external log server. 

An example of resource

```
- name: 24h-timer-trigger
  source:
    interval: 24h
  type: time
```
will causes many log files to be created as following

```
$ cat xxxxx_24h-timer-trigger_12724535_624fd02a___xxxx___2022-01-04___messages.log
2022-01-04T09:27:33+00:00 xxxx/24h-timer-trigger/12724535/624fd02a initializing
2022-01-04T09:27:38+00:00 xxxx/24h-timer-trigger/12724535/624fd02a selected worker: 26873e05-dae1-454c-b316-2b81d216de85

$ cat xxxxx_ 24h-timer-trigger_12724535____xxxx___2022-01-04___messages.log
2022-01-04T09:27:31+00:00 xxxx/24h-timer-trigger/12724535/ started
2022-01-04T09:27:39+00:00 xxxx/24h-timer-trigger/12724535/ succeeded
```

## Steps to reproduce
Enable syslog drain 

## Expected results
logs files are created for job logs 

## Actual results
huge amount of resource check logs files created

## Additional context
Its probably releated to the changes introduced in 7.0 that resource check runs as a `check build`. IN a way that resource check events are seen as build events and thus got drained to the external server. https://github.com/concourse/concourse/pull/6022

## Triaging info

* Concourse version: 7.4

* Did this used to work? Yes

xtremerui:
closes #7915 

* [x] done
* [ ] todo

## Release Note
 * Since v7.0, resouce and resource type checks are ran as builds. When syslog drainer is enabled, those check build events are also sent to external server, which requires storage space (depends on amount of resources and check interval). Now this type of events will be ignored by syslog drainer.

### DIFF
--- a/atc/db/build_factory.go
+++ b/atc/db/build_factory.go
@@ -212,10 +212,13 @@ func (f *buildFactory) constructBuildFilter() sq.Or {
 }
 
 func (f *buildFactory) GetDrainableBuilds() ([]Build, error) {
-	query := buildsQuery.Where(sq.Eq{
-		"b.completed": true,
-		"b.drained":   false,
-	})
+	query := buildsQuery.Where(
+		sq.Eq{
+			"b.completed":        true,
+			"b.drained":          false,
+			"b.resource_id":      nil,
+			"b.resource_type_id": nil,
+		})
 
 	return getBuilds(query, f.conn, f.lockFactory)
 }
--- a/atc/db/build_factory_test.go
+++ b/atc/db/build_factory_test.go
@@ -1,8 +1,10 @@
 package db_test
 
 import (
+	"context"
 	"time"
 
+	"code.cloudfoundry.org/lager/lagerctx"
 	"github.com/concourse/concourse/atc"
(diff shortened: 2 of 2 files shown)


clarafu:
What do you think about adding a unit test for checking that we no longer drain check builds?

xtremerui:
@clarafu unit test added. 

One interesting thing though, with the changes made in https://github.com/concourse/concourse/pull/7208 in v7.6, the lidar scanner created check build is not recorded in DB anymore, and thus `GetDrainableBuilds` won't include those check builds naturely.

In this case, I am not sure if this PR is really necessary or not.

clarafu (approved):
I think it's still worth having so that manual checks won't be sent to syslog drainer!

PR #7928, opened 2022-01-06
Shared (top level) secrets in AWS secrets manager

areller:

This PR introduces top level secrets when using AWS secrets manager. 

Pipeline specific secrets: `/concourse/{{.Team}}/{{.Pipeline}}/{{.Secret}}`
Team specific secrets: `/concourse/{{.Team}}/{{.Secret}}`
Top level secrets: `/concourse/{{.Secret}}`

This is useful for cases where there are secret that should be shared across teams. 

In my case, I have multiple teams that all use common images and resource types that are stored in a common ECR instance.  
The credentials (key/secret) for that ECR registry are stored in secret manager. I don't want to maintain a separate secret for each team if it's going to be the same.

### DIFF
--- a/atc/api/info_test.go
+++ b/atc/api/info_test.go
@@ -413,6 +413,7 @@ var _ = Describe("Pipelines API", func() {
 					AwsRegion:              "blah",
 					PipelineSecretTemplate: "pipeline-secret-template",
 					TeamSecretTemplate:     "team-secret-template",
+					TopLevelSecretTemplate: "top-level-secret-template",
 					SecretManager:          secretsManagerAccess,
 				}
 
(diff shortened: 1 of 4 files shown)


xtremerui (commented):
If `TopLevel` is not a term derived from AWS secrets manager, could you change it to `Shared` as it is done in vault manager [here](https://github.com/concourse/concourse/blob/6588dd6e7bc8de625cf87701f4ef3afffb1f0962/atc/creds/vault/manager.go#L22)?

areller:
> If `TopLevel` is not a term derived from AWS secrets manager, could you change it to `Shared` as it is done in vault manager [here](https://github.com/concourse/concourse/blob/6588dd6e7bc8de625cf87701f4ef3afffb1f0962/atc/creds/vault/manager.go#L22)?

no, it's not an AWS term. I'll change to shared.

xtremerui (approved):
LGTM. Thank you for the PR.

Issue #6956, opened 2021-04-29
Prometheus metric concourse_workers_containers falsely stuck on maximum value


ari-becker:
## Summary

The Prometheus metric `concourse_workers_containers` was falsely stuck on the maximum value, and not reporting for all workers. Restarting the web Kubernetes pods "unstuck" the reported metric.

## Steps to reproduce

Sorry, no idea! Logs don't seem to show anything.

## Expected results

The correct values reported in real time.

## Actual results

Grafana dashboard showing the stuck metric:
![image](https://user-images.githubusercontent.com/1388505/116545744-73ea7380-a8f9-11eb-80b3-5991435ac751.png)

Querying the exporter directly and seeing the stuck metric:
![image](https://user-images.githubusercontent.com/1388505/116545896-a72d0280-a8f9-11eb-9099-61112dc00500.png)

## Additional context
One of the `web` pods was fresh, the rest were old:
![image](https://user-images.githubusercontent.com/1388505/116546261-173b8880-a8fa-11eb-80fb-a7688e622232.png)

## Triaging info

* Concourse version: 5.8.1
* Browser (if applicable): N/A
* Did this used to work? N/A

pn-santos:
Don't want to highjack but could this be the same as: https://github.com/concourse/concourse/discussions/5992

We still see this behaviour where if a worker switches from ATC1 to ATC2 due to rebalancing then containers of that worker reported by ATC1 will become "stuck" at the last reported value while ATC2 will report the actual container count the worker has.

Obviously this makes it very tricky to properly track (via metrics) the real container counts (right now we are using a query that disregards a count if it's the same over a period of time, which kinda works but is very hacky).

It feels like the ATC never resets a worker's container count after the worker becomes registered with another ATC instance.

pn-santos:
Just to add another data point. We're seeing this happen with the `concourse_workers_tasks` and `concourse_workers_volumes` metrics too. So it seems it's an issue with all the per-worker metrics.

I'd assume that as soon as an (ATC) instance is aware that a worker is no longer registered with it it should reset/zero all it's per-worker metrics to ensure that only the ATC the worker is registered with reports the correct number.

taylorsilva:
[...]

PR #7932, opened 2022-01-07
Remove unused gc-container-collector-dropped metric

databus23:
While trying to find the root cause for #6956 I noticed that this metric seems to be not used anywhere in the code anymore


xtremerui:
Thx for the PR!

PR #7942, opened 2022-01-10
Handling huge volumes transfer in P2P streaming

### LINKED ISSUE #7910: Peer-to-peer streaming large volume across workers reliably fails
## Summary

I am trying to use the `mingc/android-build-box` Docker image to build an Android app, which is relatively large compared to the other images used in this cluster. The volume on disk is about ~10 GB:

```
[root@ip-172-31-38-220 062713c7-3d9a-4a76-5a65-8cf52b8ea200]# du -sh .
9.5G	.
```

Streaming this volume across workers (even if both have lots of disk/RAM) reliably fails.

## Steps to reproduce

Set up two workers with different tags, `a` and `b`.

Configure and run the following pipeline:

```yaml
---
resources:
- name: android-build-box
  type: registry-image
  public: true
  icon: android
  source:
    repository: mingc/android-build-box
    tag: 1.24.0

jobs:
- name: build-main
  public: true
  plan:
  - get: android-build-box
    tags:
    - a
  - task: print-gradle-version
    image: android-build-box
    config:
      platform: linux
      run:
        args:
        - -exc
        - echo hello world
        path: sh
    tags:
    - b
```

## Expected results

The volume is copied from worker `a` to worker `b` and the command executes.

## Actual results

Eventually the build errors with

```
find or create container on worker i-0555389473b7a1848: failed to stream p2p out from volume
```

Looking at the source worker `a` during the copy, there are several `tar` commands running concurrently started about a minute apart:

```
[root@ip-172-31-38-220 ~]# ps aux | grep -i tar
USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
4294967+  4806  1.4  0.0  20240  2884 ?        S    19:42   0:08 tar -C /dev/fd/3 -cz .
4294967+  4825  1.4  0.0  20240  2844 ?        S    19:43   0:07 tar -C /dev/fd/3 -cz .
4294967+  4866  1.2  0.0  20240  2816 ?        S    19:44   0:05 tar -C /dev/fd/3 -cz .
4294967+  4871  1.1  0.0  20240  2888 ?        S    19:45   0:04 tar -C /dev/fd/3 -cz .
4294967+  4891  1.1  0.0  20240  2872 ?        S    19:46   0:03 tar -C /dev/fd/3 -cz .
4294967+  4948  1.1  0.0  20276  2744 ?        S    19:47   0:03 tar -C /dev/fd/3 -cz .
4294967+  4970  1.2  0.0  20272  2652 ?        S    19:48   0:02 tar -C /dev/fd/3 -cz .
4294967+  5031  1.1  0.0  20272  2812 ?        S    19:49   0:01 tar -C /dev/fd/3 -cz .
4294967+  5051  1.2  0.0  20272  2732 ?        S    19:50   0:01 tar -C /dev/fd/3 -cz .
4294967+  5074  1.0  0.0  19948  2536 ?        S    19:51   0:00 tar -C /dev/fd/3 -cz .
```

It appears that each of these is copying the volume into RAM, so the worker eventually runs out of memory.

image

## Additional context

Source worker logs

```
Dec 31 19:59:55 ip-172-31-38-220.ec2.internal concourse[2850]: {"timestamp":"2021-12-31T19:59:55.638531288Z","level":"error","source":"baggageclaim","message":"baggageclaim.api.volume-server.stream-p2p-out.stream-p2p-out.failed-to-compress-volume","data":{"encoding":"gzip","error":"fork/exec /usr/bin/tar: cannot allocate memory","full-path":"/mnt/concourse/volumes/live/062713c7-3d9a-4a76-5a65-8cf52b8ea200/volume","session":"4.1.359.1","sub-path":".","volume":"062713c7-3d9a-4a76-5a65-8cf52b8ea200"}}
```

Source worker is a `r5dn.2xlarge` in AWS running only the Concourse worker under systemd and no other workloads inside Concourse. Memory usage is normal when it's not streaming the volume. The Concourse working directory is on a dedicated ~200 GB SSD (AWS instance store.)

Destination worker does not appear to affect anything.

## Triaging info

* Concourse version: 7.6.0
* Did this used to work? Not sure

evanchaoli:

closes #7910

After I enabled p2p streaming in a staging cluster that has real user pipelines running on it, I found two issues with p2p streaming:

First one is the issue reported in #7910, where source workers' memory usage get very high, because source worker compress volumes into memory buffer before sending to dest worker. This problem is solved by adding a pipeline on source worker, so that compressor writes to one end of the pipeline, and http request will read bytes from the other end of the pipeline.

Second is that if a volume is huge, couldn't be streamed within 3 minutes (configured by `concourse_baggageclaim_response_header_timeout`), the p2p streaming would hang for long time. Workflow of p2p streaming is as below:

Step1: ATC invokes HTTP API stream-p2p-out to source worker
Step2: source worker sending volume to dest worker
Step3: source worker response HTTP with status OK 

Where step2 may take very long time if volume is huge, when ATC's HTTP client has a 3-minutes timeout to wait for HTTP response. So that, if volume streaming takes longer than 3 minutes, ATC's HTTP client will timeout, and retry.

To solve the problem, now source will response ATC immediately, and sending a progress indicator every 10 seconds. After volume streaming finishes, it will close response with either an "ok" string or an error message string.

* [x] code changes
* [x] update tests

## Notes to reviewer

I have tested with fix with the test from #7910. The image is size of ~9GB, p2p streaming takes ~10 minutes in my test workers. Worker memory no longer has spikes when streaming large volumes.

## Release Note

* Fix a bug that P2P streaming would fail if streaming a volume takes longer than 3 minutes. This fix should be applied to both ATCs and workers.

### DIFF
--- a/worker/baggageclaim/api/volume_server.go
+++ b/worker/baggageclaim/api/volume_server.go
@@ -613,10 +613,10 @@ func (vs *VolumeServer) StreamP2pOut(w http.ResponseWriter, req *http.Request) {
 		}
 	}(doneChan)
 
-	// Send a white space to client as an indicator of in-progress every 10 seconds.
+	// Send a white space to client as an indicator of in-progress every 30 seconds.
 	// After source worker finishes sending volume to dest worker, send client "ok"
 	// for success or an error message.
-	tick := time.NewTicker(10 * time.Second)
+	tick := time.NewTicker(30 * time.Second)
 	for {
 		select {
 		case <-tick.C:
--- a/worker/baggageclaim/api/volume_server.go
+++ b/worker/baggageclaim/api/volume_server.go
@@ -614,7 +614,7 @@ func (vs *VolumeServer) StreamP2pOut(w http.ResponseWriter, req *http.Request) {
 	}(doneChan)
 
 	// Send a white space to client as an indicator of in-progress every 10 seconds.
-	// After source worker finishes sending volume ot dest worker, send client "ok"
+	// After source worker finishes sending volume to dest worker, send client "ok"
 	// for success or an error message.
 	tick := time.NewTicker(10 * time.Second)
 	for {
--- a/worker/baggageclaim/api/volume_server.go
+++ b/worker/baggageclaim/api/volume_server.go
@@ -4,18 +4,21 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"os"
 	"sync"
+	"time"
 
 	"code.cloudfoundry.org/lager"
 	"code.cloudfoundry.org/lager/lagerctx"
-	"github.com/concourse/concourse/tracing"
-	"github.com/concourse/concourse/worker/baggageclaim"
-	"github.com/concourse/concourse/worker/baggageclaim/volume"
 	uuid "github.com/nu7hatch/gouuid"
 	"github.com/tedsuo/rata"
 	"go.opentelemetry.io/otel/propagation"
+
+	"github.com/concourse/concourse/tracing"
+	"github.com/concourse/concourse/worker/baggageclaim"
+	"github.com/concourse/concourse/worker/baggageclaim/volume"
 )
 
 const httpUnprocessableEntity = 422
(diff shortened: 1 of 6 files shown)


evanchaoli:
@xtremerui @clarafu can you please help see how to handle the CodeQL scan error? I don't think that's a real problem, as worker gets streamInURL from ATC, and it's not a user-input URL.

worker/baggageclaim/api/volume_server.go
```diff
@@ -588,29 +592,47 @@ func (vs *VolumeServer) StreamP2pOut(w http.ResponseWriter, req *http.Request) {
...
+		if err != nil {
+			hLog.Error("failed-to-stream-out", err)
+			doneChan <- fmt.Errorf("%s: %w", ErrStreamP2pOutFailed, err)
+		} else {
+			close(doneChan)
+		}
+	}(doneChan)
+
+	// Send a white space to cline as an indicator of in-progress every 10 seconds.
+	// After source worker finishes sending volume ot dest worker, send either "ok"
```
analytically, line 617: to

worker/baggageclaim/api/volume_server.go
```diff
@@ -588,29 +592,47 @@ func (vs *VolumeServer) StreamP2pOut(w http.ResponseWriter, req *http.Request) {
...
+		if err != nil {
+			hLog.Error("failed-to-stream-out", err)
+			doneChan <- fmt.Errorf("%s: %w", ErrStreamP2pOutFailed, err)
+		} else {
+			close(doneChan)
+		}
+	}(doneChan)
+
+	// Send a white space to client as an indicator of in-progress every 10 seconds.
+	// After source worker finishes sending volume ot dest worker, send client "ok"
```
xtremerui, line 617:
```suggestion
	// After source worker finishes sending volume to dest worker, send client "ok"
```

worker/baggageclaim/api/volume_server.go
```diff
@@ -588,29 +592,47 @@ func (vs *VolumeServer) StreamP2pOut(w http.ResponseWriter, req *http.Request) {
...
+	// for success or an error message.
+	tick := time.NewTicker(10 * time.Second)
```
xtremerui, line 619: if the timeout is 3min, is ticking in 10 sec necessary here? I am assuming every tick will refrest the 3min timeout timer.

worker/baggageclaim/api/volume_server_test.go
```diff
@@ -1130,17 +1131,14 @@ var _ = Describe("Volume Server", func() {
...
 			request.Header.Set("Accept-Encoding", string(baggageclaim.GzipEncoding))
 			recorder := httptest.NewRecorder()
 			handler.ServeHTTP(recorder, request)
-			Expect(recorder.Code).To(Equal(404))
-
-			var responseError *api.ErrorResponse
-			err := json.NewDecoder(recorder.Body).Decode(&responseError)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(responseError.Message).To(Equal("no such file or directory"))
+			Expect(recorder.Code).To(Equal(200)) // status code should always be 200, error is in body
```
xtremerui, line 1140: Is it possible to maintain the way it returns status code as before like 403 404 and 500?

worker/baggageclaim/client_test.go
```diff
@@ -445,17 +446,34 @@ var _ = Describe("Baggage Claim Client", func() {
...
+					})
+					It("should succeed", func() {
+						err := vol.StreamP2pOut(context.TODO(), "some-dest-path", "http://some-url", "gzip")
+						Expect(err).ToNot(HaveOccurred())
+					})
 				})
-				It("should succeed", func() {
-					err := vol.StreamP2pOut(context.TODO(), "some-dest-path", "http://some-url", "gzip")
-					Expect(err).ToNot(HaveOccurred())
+				Context("when p2p streaming succeeds", func() {
```
xtremerui, line 463:
```suggestion
				Context("when p2p streaming failed", func() {
```

xtremerui (commented):
Thx for the PR!

worker/baggageclaim/api/volume_server.go
evanchaoli: 3 min is configurable. A p2p streaming may take uncertain time, maybe a few seconds, maybe 10 minutes or even longer. So I believe a periodical progress helps maintain the HTTP transaction stable. You may argue is 10 sec interval is short, that is fine, I don't consider much difference for 10 sec, 30 sec or 1 min, which should not be big overhead.

worker/baggageclaim/api/volume_server_test.go
evanchaoli: I don't think so. With the httplib, once we write to http response writer, if status is not set, the lib will auto set 200 as status. So, here 200 just means p2p streaming successfully started, actually result will be carried in response body. Actually, I think the new way is better. Because response body may take more descriptive error message than http code does, so that ATC log may provide more accurate error description.

worker/baggageclaim/api/volume_server.go
xtremerui: Could you make it 30sec(same as workers beacon to ATC)?

PR #7955, opened 2022-01-12
Do not cache secrets indefinitely in Vault

sprsquish:
Signed-off-by: Jeff Smick <jeff@fauna.com>

Change the cache behavior such that an expiration time resulting in a negative duration does not add the item to the cache. 

Currently a Vault secret with a lease duration of 0 will cause that secret to be cached indefinitely.

This change eliminates the ability to set an infinite cache duration, which may be a bug others are relying on. 

## Notes to reviewer

After pulling a secret from Vault, the expiration time is calculated from `time.Now()` and `secret.LeaseDuration`: https://github.com/concourse/concourse/blob/e3d9f5bb1d41e56b5de584e2b96505a90f1fe2b6/atc/creds/vault/vault.go#L85-L89

In the secrets cache, that's converted back to a duration via `time.Until`: https://github.com/concourse/concourse/blob/e3d9f5bb1d41e56b5de584e2b96505a90f1fe2b6/atc/creds/cached_secrets.go#L64-L67

Enough time will have passed that this number ends up being negative. For instance, I wrote a small test and ran it on my machine: 
```go
func main() {
	expiration := time.Now()
	duration := time.Until(expiration)
	fmt.Println("duration", duration)
}
```

```
$ go run go-time.go
duration -83ns
```

According to the [cache docs](https://pkg.go.dev/github.com/patrickmn/go-cache#Cache.Set) "If the duration is 0 (DefaultExpiration), the cache's default expiration time is used. If it is -1 (NoExpiration), the item never expires." However, the code doesn't look specifically for `-1`, rather it sets the duration only if it's been set to `0` or a number `> 0` (see [this snippet](https://github.com/patrickmn/go-cache/blob/46f407853014144407b6c2ec7ccc76bf67958d93/cache.go#L53-L59)).

### DIFF
--- a/atc/creds/cached_secrets.go
+++ b/atc/creds/cached_secrets.go
@@ -66,7 +66,14 @@ func (cs *CachedSecrets) Get(secretPath string) (interface{}, *time.Time, bool, 
 				duration = itemDuration
 			}
 		}
-		cs.cache.Set(secretPath, entry, duration)
+
+		// a negative duration will cause the cache to never expire the secret
+		// reduce the minimum cache duration to 1s and assume anything within that second is requesting
+		// a cache duration of 0 (do not cache).
+		seconds := duration.Seconds()
+		if seconds >= 1 || seconds <= -1 {
+			cs.cache.Set(secretPath, entry, duration)
+		}
 	} else {
 		cs.cache.Set(secretPath, entry, cs.cacheConfig.DurationNotFound)
 	}
--- a/atc/creds/cached_secrets.go
+++ b/atc/creds/cached_secrets.go
@@ -66,14 +66,7 @@ func (cs *CachedSecrets) Get(secretPath string) (interface{}, *time.Time, bool, 
 				duration = itemDuration
 			}
 		}
-
-		// a negative duration will cause the cache to never expire the secret
-		// reduce the minimum cache duration to 1s and assume anything within that second is requesting
-		// a cache duration of 0 (do not cache).
-		seconds := duration.Seconds()
-		if seconds >= 1 || seconds <= -1 {
-			cs.cache.Set(secretPath, entry, duration)
-		}
+		cs.cache.Set(secretPath, entry, duration)
 	} else {
 		cs.cache.Set(secretPath, entry, cs.cacheConfig.DurationNotFound)
 	}
--- a/atc/creds/cached_secrets_test.go
+++ b/atc/creds/cached_secrets_test.go
@@ -162,4 +162,19 @@ var _ = Describe("Caching of secrets", func() {
 		Expect(underlyingMisses).To(BeIdenticalTo(4))
 	})
 
+	It("should not cache forever if expiration is now", func() {
+		now := time.Now()
+		secretManager.GetStub = makeGetStub("foo", "value", &now, true, nil, &underlyingReads, &underlyingMisses)
+
+		// initial get should cause a read
+		_, expiration, _, _ := cachedSecretManager.Get("foo")
+		Expect(expiration).To(BeIdenticalTo(&now))
+		Expect(underlyingReads).To(BeIdenticalTo(1))
+		Expect(underlyingMisses).To(BeIdenticalTo(0))
+
+		// result should not be cached causing another read
+		_, _, _, _ = cachedSecretManager.Get("foo")
+		Expect(underlyingReads).To(BeIdenticalTo(2))
+		Expect(underlyingMisses).To(BeIdenticalTo(0))
+	})
 })
--- a/atc/creds/cached_secrets_test.go
+++ b/atc/creds/cached_secrets_test.go
@@ -162,19 +162,4 @@ var _ = Describe("Caching of secrets", func() {
 		Expect(underlyingMisses).To(BeIdenticalTo(4))
 	})
 
-	It("should not cache forever if expiration is now", func() {
-		now := time.Now()
-		secretManager.GetStub = makeGetStub("foo", "value", &now, true, nil, &underlyingReads, &underlyingMisses)
-
-		// initial get should cause a read
-		_, expiration, _, _ := cachedSecretManager.Get("foo")
-		Expect(expiration).To(BeIdenticalTo(&now))
-		Expect(underlyingReads).To(BeIdenticalTo(1))
-		Expect(underlyingMisses).To(BeIdenticalTo(0))
-
-		// result should not be cached causing another read
-		_, _, _, _ = cachedSecretManager.Get("foo")
-		Expect(underlyingReads).To(BeIdenticalTo(2))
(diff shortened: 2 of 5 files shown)


clarafu:
@sprsquish Is it wrong to let users cache their secrets indefinitely if that is what they want? Sorry I'm asking because I am unfamiliar with the best practices of caching and I'd like to know more about the motivations behind this change! Also just to note that this would be a breaking change. You might already know this but if you do not want to cache secrets, you could also set the `secret-cache-enabled` flag to `false`.

sprsquish:
> Is it wrong to let users cache their secrets indefinitely if that is what they want? Sorry I'm asking because I am unfamiliar with the best practices of caching and I'd like to know more about the motivations behind this change! 

@clarafu I don't think it's wrong, no. There may be legitimate need. It'd be helpful to have some sort of mechanism to bust the secrets cache in cases where the wrong value is set. I ran into this because I needed to rotate secrets and was expecting to be able to do so by simply updating the value in Vault.

> You might already know this but if you do not want to cache secrets, you could also set the secret-cache-enabled flag to false.

Indeed, but there are some secrets that we'd like to cache so we can't completely disable this.

After a bit more time with this, I think the core issue here is the conversion from lease duration to expiration time and back to duration. Maybe a better fix is to check if the recalculated value is within some negative delta and treat that as it was mean to be `time.Now()`

breathe:
I think I'm hitting the same issue as the op ...  As far as I'm understanding -- vault kv2 can never return any value for lease_duration other than 0 ... which means that concourse is always gonna compute a negative value for the cache expiration duration, which gets interpreted by the caching layer as 'cache forever' ... its pretty annoying not being able to update to a new version of the secret and have concourse pick it up...

xtremerui:
> vault kv2 can never return any value for lease_duration other than 0

@breathe could you point me the doc about this?

xtremerui:
> Currently a Vault secret with a lease duration of 0 will cause that secret to be cached indefinitely.

@sprsquish the author of cache secret actually mentioned that it is implemented to mimic go-cache that you linked the snippet with. https://github.com/concourse/concourse/commit/82facc423a929a607b91fb0bac67bc795befc539

In go-cache, when an expiration is 0, like you said, the cache never expired. To me, now the cache secret in Concourse is doing the same and I don't see a problem of it.

For your secret rotating problem, when you update the secret in vault, I am expecting the lease duration be also refreshed. If the secret has expired before your rotation, Vault should already invalid the secret and it should not be able to fetch. How is the case of 0 lease duration happening?

I think we might need to fully understand how lease duration works first.

databus23:
> In go-cache, when an expiration is 0, like you said, the cache never expired. To me, now the cache secret in Concourse is doing the same and I don't see a problem of it.

The problem with the current implementation is that for v2 k/v stores secrets are cached indefinitely by concourse with no way to prevent that.
At the moment we have to restart the concourse web components every time a user changes a secret in our vault (e.g. create a new version) because concourse continues to use the old version.

> For your secret rotating problem, when you update the secret in vault, I am expecting the lease duration be also refreshed. If the secret has expired before your rotation, Vault should already invalid the secret and it should not be able to fetch. How is the case of 0 lease duration happening?

No that is not case. There is no lease duration for the secret k,v-store in vault:

While the v1 k/v-store has the following mention in the [docs](https://www.vaultproject.io/docs/secrets/kv/kv-v1#ttls):

> Unlike other secrets engines, the KV secrets engine does not enforce TTLs for expiration. Instead, the lease_duration is a hint for how often consumers should check back for a new value.

This TTL section and any mention of leases is completely gone from the [v2 k/v documentation](https://www.vaultproject.io/docs/secrets/kv/kv-v2), also the instructions on how to set the `ttl` from the v1 docs don't work, the `lease_duration` for secrets in v2 stores always returns comes back as 0.

When this caching was implemented there was only the v1 k/v store available in vault and when support for v2 was implemented this detail was overlooked.

The think this really is a bug that needs to be addressed. We are only using v2 secret stores and we have quite bit of support overhead to to this caching behaviour with no permanent fix at the moment.

xtremerui:
@databus23  thx for the input. Just want to confirm, is https://www.vaultproject.io/docs/concepts/lease for kv v1 only?

breathe:
On that same page @xtremerui 

```
Note: The [Key/Value Backend](https://www.vaultproject.io/docs/secrets/kv) which stores arbitrary secrets does not issue leases although it will sometimes return a lease duration; see the documentation for more information.
```

I can't find official documentation on this -- but the `lease duration`  referred to by the snippet above was configurable metadata in a field named `ttl` with vault kv1.  That same field is exposed in the go kv2 vault client but it is not configurable with vault kv2.  That value is always 0 and you can't do anything to change it. https://github.com/hashicorp/vault/issues/6274

xtremerui:
> Indeed, but there are some secrets that we'd like to cache so we can't completely disable this.

@sprsquish how could you still cache some secrets against a kv v2 with the changes in this PR since the least duration is always 0? Isn't it equaivalent to set `secret-cache-enabled` to false?

Or you have kv v1 and v2 configured at the same time?

breathe:
@xtremerui I think that if this change was merged it would have the effect of allowing one to use a default cache time for kv2 wouldn't it?  I thought lease_duration == 0 signaled the caching layer to use the globally defined default.

I previously suggested disabling the secret cache entirely in our environment -- and I believe we are going to try that but we haven't yet.

xtremerui:
> I thought lease_duration == 0 signaled the caching layer to use the globally defined default caching time?

@breathe , lease_duration 0 will cause the expiration of the secret to be negative, but it won't allow global duration kicked in unless it set the duration of the cache secret to 0 like sth below, so underlying cache could use its`defaultExpiration` which is set by `cacheConfig.Duration`

```
 cs.cache.Set(secretPath, entry, 0)
```
but i dont find it in the implementation. So I am wondering how OP can still use some secrets with caching with this PR.

sprsquish:
@xtremerui 
> how could you still cache some secrets against a kv v2 with the changes in this PR since the least duration is always 0? Isn't it equaivalent to set secret-cache-enabled to false?

You're right. If the duration is between 1 and -1, this change won't set the item in the cache. It was a bit of a hack. What I'd really like to do is check if Vault is using kv2 and if so, ignore LeaseDuration to signal to the cache that it should use the default.

sprsquish:
@xtremerui I pushed a commit that takes a different approach and uses -1 as a sentinel in LeaseDuration. If the vault secret lookup sees that, it'll omit the expiration altogether and allow the cache to use its default.

It feels a bit hacky, but I wanted to post something that could start a discussion.

sprsquish:
I've added a test against kv1 and kv2. They feel out of character for the suite that they're in, but I didn't know where else to add them.

xtremerui (approved):

PR #7978, opened 2022-01-19
GC builds based on chronological order

### LINKED ISSUE #7951: Build logs get immediately reaped while re-running an already reaped build with the same inputs
## Summary

While re-running a build whose logs have been already reaped, the build runs but the newer build's logs gets reaped immediately.

## Steps to reproduce

1. Set `maxBuildLogsToRetain` to 10.
2. Create a sample pipeline with one or more jobs.
3. Run more than 10 builds. 
4. Run the 11th build so that the build # 1 gets reaped.
5. Re-running the build # 1  with the same inputs.

## Expected results
The build logs of the build # 1.1 should not be reaped.

## Actual results
The build logs of the build # 1.1 gets reaped.

## Additional context
Attaching screenshots from our actual builds:
Screenshot 2022-01-12 at 3 06 39 PM
Screenshot 2022-01-12 at 2 46 56 PM

## Triaging info

* Concourse version: v6.7.6
* Browser (if applicable): Microsoft Edge

xtremerui:
when getting builds for GC, instead of ordering builds by rerun group (which still be the default way), order builds chronologically so a rerun build won't be reaped immediately if the parent build is reaped already.

closes #7951 

* [x] done
* [ ] todo

## Release Note
 * Fix a bug that events of a rerun build be reaped immediately if its prarent build is already reaped. Now candidate builds for GC will be ordered chronologically.

### DIFF
--- a/atc/db/build_factory.go
+++ b/atc/db/build_factory.go
@@ -164,7 +164,7 @@ func (f *buildFactory) VisibleBuilds(teamNames []string, page Page) ([]BuildForA
 			f.lockFactory)
 	}
 	return getBuildsWithPagination(newBuildsQuery, minMaxIdQuery, page, f.conn,
-		f.lockFactory)
+		f.lockFactory, false)
 }
 
 func (f *buildFactory) AllBuilds(page Page) ([]BuildForAPI, Pagination, error) {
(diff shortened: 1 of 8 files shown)


clarafu (commented):
The implementation looks great!! Would you be able to add some unit tests in `job_test.go` for the new `ChronoBuilds` method? Thanks!

xtremerui:
Unit test added @clarafu

clarafu (approved):
Thanks! LGTM!

PR #7989, opened 2022-01-20
Garbage collect task caches from paused pipelines

### LINKED ISSUE #7988: Garbage collect task caches when pipeline is paused
## Summary

This is spawned off from https://github.com/concourse/concourse/issues/7986 where we realized that task caches were not being gced properly when a pipeline is archived. I think we might also want to garbage collect task caches from paused pipelines, but specifically only when there are no more `running` state builds for that job. 

## Context

* RFC: concourse/rfcs#
* Prior discussion: #
* Depends on #
* Part of #

xtremerui:
Garbage collect task caches from jobs that in a paused pipeline 

closes #7988

* [x] done
* [ ] todo

## Notes to reviewer
It seems finding jobs in a paused pipeline that has no running builds is non-trivial, the query becomes very slow if joining with builds table. IMHO if user gonna pause a pipeline, he either first making sure all jobs in the pipeline are done or he doesn't care the result of the running builds of a pausing pipeline. So I think we could ignore the job status and just GC all task caches from a paused pipeline.

## Release Note
 * When a pipeline or a job is paused, the task caches that used in the pipeline's job will be garbage collected. This should help free up worker disk space.

### DIFF
--- a/atc/db/dbtest/builder.go
+++ b/atc/db/dbtest/builder.go
@@ -5,6 +5,7 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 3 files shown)


clarafu:
Hmm to your comment in the description about assuming the user will make sure no jobs are running when they pause the pipeline, Im not sure if that is something that we should assume because I have definitely paused a pipeline many times when jobs were still running. 😆  I think it might be pretty dangerous for us to destroy task caches from underneath a build that is running even if the pipeline is paused. I was wondering if we have the query just do a quick check to see if the `next_build_id` for the job row is in `pending or started` state, that might not make the query really slow? The `next_build_id` shows the next build that should run for this job, so if the job has a build that is queued up and that build either has not run yet (in pending state) or is currently running (in started state) then we should not kill the task caches. WDYT?

xtremerui:
Only question is would `next_build_id` bump if the build change status from pending to started i.e. if a started build still count as next build of the job.

xtremerui:
Verified with the unit test that `next_build_id` will be `null` if a job build is finished. This allow the query to not need to join with builds table, which is ideal for us.

clarafu:
Nice!! One more thing is there can also be a time when the job is paused too. Should we just slip that in here too and have it remove task caches if the job is paused?

xtremerui:
Hmmm, thats interesting since pausing a job would then means all its task caches busted, which is an implicit way of clearing task cache from Web UI (as right now only `fly clear-task-cache` is available)

clarafu (approved):
Everything looks great!! Tested all the cases out (including paused pipeline, paused job, running/pending build) with manual acceptance testing and everything works great :)

PR #7998, opened 2022-01-23
WIP: remove global mutex from lock.go

evanchaoli:

closes # 

* [x] done
* [ ] todo

## Release Note

* 

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -254,6 +254,10 @@ type RunCommand struct {
 	} `group:"Feature Flags"`
 
 	BaseResourceTypeDefaults flag.File `long:"base-resource-type-defaults" description:"Base resource type defaults"`
+
+	AtcCount                  int `long:"atc-count" description:"Count of ATCs. This will help coordinate components." default:"1"`
+	MaxTrackingBuilds         int `long:"max-tracking-builds" description:"Max builds running per ATC" default:"0"`
+	MaxCheckingResourceScopes int `long:"max-checking-resource-scopes" description:"Max builds running per ATC" default:"0"`
 }
 
 type Migration struct {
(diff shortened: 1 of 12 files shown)


clarafu:
Hmm I think I know what the `acquireMutex` lock is for. 

https://github.com/concourse/concourse/blob/1b1dc768fd4fe124c5fde7ed82ee867954e15dc8/atc/db/lock/lock.go#L165-L166 

I think this is for making sure we don't have two threads trying to acquire the same lock at the same time. This is because if two threads try to acquire the same advisory lock (in the same session), they both will be able to acquire it. That is why we check if the lock is registered in the `lockRepo` in order to know if we have already acquired this lock https://github.com/concourse/concourse/blob/1b1dc768fd4fe124c5fde7ed82ee867954e15dc8/atc/db/lock/lock.go#L170-L173 because if we try to acquire the same lock again in the same session, it will successfully acquire. But if the two threads kick off at the same time, both will check that the lock is unregistered and then both select the advisory lock from postgres and both will be successful (which we don't want).

evanchaoli:
@clarafu You are right, as a lock factory use a single db connection, a mutex is needed to prevent multiple threads from acquiring the same advisory lock. But the mutex doesn't have to be a global one. When a build is trying to acquiring a creating-volume-lock, why it needs to wait for lidar scanner to acquire lidar-scanner-lock? Thus, from my test result, the global lock more like to slow down an ATC so that other ATC may share workloads.

In theory, each lock may use a separate mutex, but given the total amount of locks holding, there will be too many mutexes which is also costly. Given there is single db connection for lock factory, acquiring lock has to be in serial. We can use a db connection and a mutex for a type of locks, which is code change I proposed in this PR.

But from my observation over our prod cluster, the bottleneck is currently not at db-lock. So we don't have to patch 6.x with this lock change. For 7.x, I think this is still needed. I will raise a separate PR for the lock change on master branch.

clarafu:
I see, that definitely makes sense to have the mutex be on the lock type. 👍 

Unfortunately, I don't think we will be doing any further patches to the 6.7.x series because of our limited staff and all our efforts are focused on getting a stable version of 7.x series. Would you be comfortable with us working towards releasing a more stable version of the 7.x series and using that to upgrade runway?

clarafu:
I'm going to close this because we are not going to be doing any patches for 6.7.x series! https://github.com/concourse/concourse/pull/8198 is the same changes but on master branch

PR #8002, opened 2022-01-26
Reset resource config IDs on save

### LINKED ISSUE #7468: Wrong image tag (registry-image) used after being updated

ghost:
When a pipeline is updated, the resources and their config is updated in the database. However, the `resource_config_id` and `resource_config_scope_id` fields are not updated and so the build inputs algorithm considers "stale" versions (versions associated with an old resource config) as valid.

By setting these values back to NULL, we force LIDAR to FindOrCreate an existing resource_config and resource_config_scope (create new if config has changed) whilst also preventing the stale versions from being considered. As LIDAR is triggered immediately after SavePipeline and LIDAR does the FindOrCreate before trying to schedule any builds, this means that any existing versions for resources that haven't changed are quickly "reattached" and become valid again.

Fixes #7468

Co-Authored-By: Richard Inskip <richard.inskip@depop.com>
Signed-off-by: Andy Paine <andy.paine@engineerbetter.com>

closes #7468

## Notes to reviewer

We have revisited the suggested workaround for this issue in the PR (https://github.com/concourse/concourse/pull/7494) which was subsequently reverted. 

> I'm wondering if there is a reason we didn't just clear the resource_config_scope_id along with the resource_config_id when we set a pipeline in db/team.go? 
(https://github.com/concourse/concourse/pull/7494#issuecomment-912561016)

It was assumed that this would causes pipelines/jobs to wait for a check to occur if even there wasn't a change:

> The con with this approach was that every resource needed to be checked again before anything could be scheduled, which makes things feel kinda slow for no good reason.
(https://github.com/concourse/concourse/pull/7494#issuecomment-912568583)

However, upon testing this appears to be untrue. When a resource is not changed, LIDAR reuses the old config & config_scope

https://github.com/concourse/concourse/blob/d8d27b9bd997db0e5bce5ed4f3c8dbe9b1b43466/atc/engine/check_delegate.go#L99-L105

This fixes the original issue and no performance effects are visible when resources were not changed.

## Release Note

* Fixes #7468 where wrong resource config was used

### DIFF
--- a/atc/db/team.go
+++ b/atc/db/team.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strconv"
+	"strings"
 	"time"
 
 	"code.cloudfoundry.org/lager"
@@ -1230,113 +1232,6 @@ func registerSerialGroup(tx Tx, serialGroup string, jobID int) error {
 	return err
 }
 
-func saveResource(tx Tx, resource atc.ResourceConfig, pipelineID int) (int, error) {
-	configPayload, err := json.Marshal(resource)
-	if err != nil {
-		return 0, err
-	}
-
-	es := tx.EncryptionStrategy()
-	encryptedPayload, nonce, err := es.Encrypt(configPayload)
-	if err != nil {
-		return 0, err
-	}
-
-	onConflict := "ON CONFLICT (name, pipeline_id) DO UPDATE SET config = EXCLUDED.config, active = EXCLUDED.active, nonce = EXCLUDED.nonce, type = EXCLUDED.type"
-
-	oldConfig, oldType, err := existingResourceConfig(tx, resource.Name, pipelineID)
-	if err != nil {
-		return 0, err
-	}
-	if mapHash(resource.Source) != mapHash(oldConfig.Source) || resource.Type != oldType {
-		onConflict = onConflict + ", resource_config_id = NULL, resource_config_scope_id = NULL"
-	}
-
-	var resourceID int
-	err = psql.Insert("resources").
-		Columns("name", "pipeline_id", "config", "active", "nonce", "type").
-		Values(resource.Name, pipelineID, encryptedPayload, true, nonce, resource.Type).
-		Suffix(onConflict).
-		Suffix("RETURNING id").
-		RunWith(tx).
-		QueryRow().
-		Scan(&resourceID)
-	if err != nil {
-		return 0, err
-	}
-
-	_, err = psql.Delete("resource_pins").
-		Where(sq.Eq{
-			"resource_id": resourceID,
-			"config":      true,
-		}).
-		RunWith(tx).
-		Exec()
-	if err != nil {
-		return 0, err
-	}
-
-	if resource.Version != nil {
-		version, err := json.Marshal(resource.Version)
-		if err != nil {
-			return 0, err
-		}
-
-		_, err = psql.Insert("resource_pins").
-			Columns("resource_id", "version", "comment_text", "config").
-			Values(resourceID, version, "", true).
-			Suffix("ON CONFLICT (resource_id) DO UPDATE SET version = EXCLUDED.version, comment_text = EXCLUDED.comment_text, config = true").
-			RunWith(tx).
-			Exec()
-		if err != nil {
-			return 0, err
-		}
-	}
-
-	return resourceID, nil
-}
-
-func existingResourceConfig(tx Tx, name string, pipelineID int) (atc.ResourceConfig, string, error) {
-	var (
-		configBlob, nonce sql.NullString
-		oldType           string
-	)
-	row := psql.Select("config", "type", "nonce").
-		From("resources").
-		Where(sq.Eq{
-			"name":        name,
-			"pipeline_id": pipelineID,
-		}).
-		RunWith(tx).
-		QueryRow()
-	err := row.Scan(&configBlob, &oldType, &nonce)
-	if err != nil && err != sql.ErrNoRows {
-		if err == sql.ErrNoRows {
-			return atc.ResourceConfig{}, "", nil
-		}
-		return atc.ResourceConfig{}, "", err
-	}
-
-	var noncense *string
-	if nonce.Valid {
-		noncense = &nonce.String
-	}
-
-	oldConfig := atc.ResourceConfig{}
-	if configBlob.Valid {
-		decryptedConfig, err := tx.EncryptionStrategy().Decrypt(configBlob.String, noncense)
-		if err != nil {
-			return atc.ResourceConfig{}, "", err
-		}
-
-		err = json.Unmarshal(decryptedConfig, &oldConfig)
-		if err != nil {
-			return atc.ResourceConfig{}, "", err
-		}
-	}
-	return oldConfig, oldType, nil
-}
-
 func saveResourceType(tx Tx, resourceType atc.ResourceType, pipelineID int) error {
 	configPayload, err := json.Marshal(resourceType)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


atc/db/team.go
```diff
@@ -563,18 +563,6 @@ func savePipeline(
 		return 0, false, err
 	}
 
-	_, err = psql.Update("resources").
-		Set("resource_config_id", nil).
-		Where(sq.Eq{
-			"pipeline_id": pipelineID,
-			"active":      false,
```
xtremerui, line 570: why is this block removed? Is it conflicting with the changes?

rjinskidepop: The functionality is incorporated into the updated `saveResource` query: https://github.com/concourse/concourse/pull/8002/files#diff-bc775ddab7d4f0909d112312360cad5826a0958c9b5e83ce16911337ff8ecd48R1249

xtremerui (dismissed):

atc/db/team.go
```diff
@@ -1258,7 +1246,7 @@ func saveResource(tx Tx, resource atc.ResourceConfig, pipelineID int) (int, erro
 	err = psql.Insert("resources").
 		Columns("name", "pipeline_id", "config", "active", "nonce", "type").
 		Values(resource.Name, pipelineID, encryptedPayload, true, nonce, resource.Type).
-		Suffix("ON CONFLICT (name, pipeline_id) DO UPDATE SET config = EXCLUDED.config, active = EXCLUDED.active, nonce = EXCLUDED.nonce, type = EXCLUDED.type").
+		Suffix("ON CONFLICT (name, pipeline_id) DO UPDATE SET config = EXCLUDED.config, active = EXCLUDED.active, nonce = EXCLUDED.nonce, type = EXCLUDED.type, resource_config_id = NULL, resource_config_scope_id = NULL").
```
xtremerui, line 1249: I stand corrected here that @clarafu just pointed out that the implementation here will set ALL existing resources' resource_config_id to null. In this case, there will be a window of resource config collector to GC those un-linked resource configs and causing all existing versions to be deleted.

https://github.com/concourse/concourse/blob/d8d27b9bd997db0e5bce5ed4f3c8dbe9b1b43466/atc/db/resource_config_factory.go#L236-L243

Note the delete cascade chain is resource_config_versions(resource_config_scopes.id) -> resource_config_scopes(resource_configs.id)->resource_configs

andy-paine:
That kind of makes sense. Whilst it isn't ideal to make unchanged resources have to re-check to re-fetch versions, at the very least that should eventually be consistent. The current behaviour is a bug with potential security implications (if I put a specific version in my resource I really should get that version in my job).

To us, it looks like there are 2 options from here going forwards:
* In `saveResources` we get the current configuration of each resource before saving it and then only set the `resource_config_id` and `resource_config_scope_id` columns to `NULL` for the resources that have changed config/type. This adds a bunch of extra queries to this code path but means we only affect changed resources.
* Refactor how the `resource_config_scope_id` relationship works by adding the `source_hash` field to the `resources` table and making the primary key of the `resource_config_scopes` table that hash. This would allow `saveResources` to _always_ calculate and set the `source_hash` column in the DB and if it hasn't changed, there will still be a `resource_config_scopes` row to match. If it has changed then the LIDAR scanner needs to run to create a new `resource_config_scopes` row and discover new versions. This is a much bigger change but avoids lots of these "which async process will get there first" questions and doesn't require extra DB queries. Would also need to work out how to deal with global resources vs. non-global resources (probably include the pipeline_id in the hash when global is turned off)

Any thoughts on these approaches? Ideally we would get something to fix this bug merged ASAP

xtremerui:
> * Refactor how the `resource_config_scope_id` relationship works by adding the `source_hash` field to the `resources` table

I'd only consider this one if there is no other way. I totally agree that this is a much bigger change.

> * only set the `resource_config_id` and `resource_config_scope_id` columns to `NULL` for the resources that have changed config/type.

Please take a look at [this issue](https://github.com/concourse/concourse/issues/7697)for the potential problems of the approach. And there is also this [PR](https://github.com/concourse/concourse/pull/7703) who I believe is trying to solve the same problem. Since the PR is still a draft could you coordinate with @taylorsilva ? Thx!

We'd love to get this bug fixed if possible. The targeted date for v7.7 is mid Feb or so. Hopefully we can find a solution by then. Thank you again for contributing on this issue.

taylorsilva:
I have no intention of picking up that PR again but happy to help hand it off to someone to work on, if it's worth working off of.

andy-paine:
What's your current thoughts on using trigger vs. doing a `SELECT` in the `saveResources` block? Splitting the fields would help the trigger since it could just diff the encrypted blob but having a trigger still feels a bit awkward. Potentially we could index the `config` column and then use the `ON CONFLICT` part of the SQL insert but that also sounds hacky.

taylorsilva:
Been a while since I thought about this issue, but I think going with a `SELECT` in something like `saveResources` is a better way to go potentially. I'm not recalling why I was hesitant to do it that way, but it seems like the best way to do given the way things are currently structured.

xtremerui:
If possible I would avoid doing migrations to change the DB structure. If trigger means we have to split the field in migration then i think `select` is fine.

andy-paine:
Cool. I've pushed a new commit that does this. It isn't the prettiest thing in the world but it works. I tried using `scanResource` from `atc/db/resource.go` but that requires a `newEmptyResource` to scan into (rather than just `resource{}`) which then requires a connection & lock factory. Passing these all the way down from `team.SavePipeline` felt a bit ugly so I just replicated the relevant bits of `scanResource` in `team.go` as we don't need the connection or lock factory stuff anyway. This allows the query to be a bit leaner too as we only fetch the config, type and nonce this way.

atc/db/team.go
```diff
@@ -1286,6 +1296,47 @@ func saveResource(tx Tx, resource atc.ResourceConfig, pipelineID int) (int, erro
...
+		From("resources").
+		Where(sq.Eq{
+			"name":        name,
+			"pipeline_id": pipelineID,
+		}).
+		RunWith(tx).
+		QueryRow()
+	err := row.Scan(&configBlob, &oldType, &nonce)
+	if err != nil && err != sql.ErrNoRows {
+		if err == sql.ErrNoRows {
```
clarafu, line 1314: We will never reach here because there is the `&& err != sql.ErrNoRows` check right above it

atc/db/team.go
```diff
@@ -1286,6 +1296,47 @@ func saveResource(tx Tx, resource atc.ResourceConfig, pipelineID int) (int, erro
...
+	}
+
+	var noncense *string
+	if nonce.Valid {
+		noncense = &nonce.String
+	}
+
+	oldConfig := atc.ResourceConfig{}
+	if configBlob.Valid {
+		decryptedConfig, err := tx.EncryptionStrategy().Decrypt(configBlob.String, noncense)
```
clarafu, line 1327: Hmm would it be possible for us to compare the encrypted payloads against each other to avoid needing to decrypt every config for every resource? Doing that will also cause there to be a diff if the db encryption key is rotated so maybe after we compare the encrypted payloads to be different, we could then check for a diff in the underlying config old vs new? That being said I'm not sure 100% if we can just compare the encrypted configs haha

atc/db/team.go
```diff
@@ -1286,6 +1296,47 @@ func saveResource(tx Tx, resource atc.ResourceConfig, pipelineID int) (int, erro
 	return resourceID, nil
 }
 
+func existingResourceConfig(tx Tx, name string, pipelineID int) (atc.ResourceConfig, string, error) {
+	var (
+		configBlob, nonce sql.NullString
+		oldType           string
+	)
+	row := psql.Select("config", "type", "nonce").
```
clarafu, line 1304: Just a small refactor suggestion here!

```
err := psql.Select("config", "type", "nonce").
       From("resources").
       Where(sq.Eq{
		"name":        name,
		"pipeline_id": pipelineID,
       }).
       RunWith(tx).
       QueryRow().
       Scan(&configBlob, &oldType, &nonce)
```

clarafu (commented):
I think this is probably the best way to approach this, even though it adds some extra select queries. It would be best if we can figure out a way to not need to decrypt the config every time too though haha

atc/db/team.go
andy-paine: This would be a nice-to-have I think - it would require splitting out the fields in the encrypted blob that _are_ allowed to change (icon etc.) from the source which would be a much bigger change. I don't disagree that it would be nice but I don't want to scope creep the fixing of this bug too much for a performance improvement

andy-paine:
I've added a new commit that mitigates the extra DB querying by doing the SELECT and INSERT for all resources at once. This means we always make only 4 DB queries (SELECT existing, INSERT new/updated resources, DELETE resource pins, INSERT new/updated pins) so we go from `O(n)` queries to `O(1)` 🙂 

Had to do some ugly-ish things by creating a new struct for `existingResource` but the structs/functions in `atc/db/resource.go` just don't play nicely with transactions and the config IDs aren't nillable so required a bunch of extra `== 0` checks

atc/db/team.go
```diff
@@ -1567,19 +1499,158 @@ func inactivateTableForPipeline(tx Tx, pipelineID int, tableName string) error {
...
+		return nil, err
+	}
 	return resourceNameToID, nil
 }
 
+type existingResource struct {
+	Type                  string
+	Config                atc.ResourceConfig
+	ResourceConfigID      interface{}
+	ResourceConfigScopeID interface{}
```
andy-paine, line 1574: These only exist to be used in a squirrel query `.Values()` function so the type isn't that relevant - the main thing is we want them to be nil-able (rather than 0 which is the default for `int`)

atc/db/team.go
clarafu: I could be missing something but I don't think I completely understand why it will require splitting out the fields in the encrypted blob? Would we not be able to straight up compare the old vs new encrypted payloads and most of the time they will be the same because nothing has changed. But if there is a diff then we decrypt it and do the same comparisons that we do right now (comparing the source and the type)?

andy-paine: The encrypted blob is the whole config - not just the source. We don't necessarily want to reset the versions if e.g. the icon (which is part of the config) has changed. If we just compared the encrypted bit it would change if the icon changed I think

andy-paine: Yep that works - it is a bit of added complexity but should improve the throughput as _most_ resources probably don't change on your average SavePipeline call 👍

atc/db/team.go
```diff
@@ -1567,19 +1500,178 @@ func inactivateTableForPipeline(tx Tx, pipelineID int, tableName string) error {
...
+			return nil, err
+		}
+		values := []interface{}{resource.Name, pipelineID, encryptedPayload, true, nonce, resource.Type}
+
+		existing, exists := existingResources[resource.Name]
+
+		if !exists {
+			values = append(values, nil, nil)
+		} else {
+			configsDiffer, err := configsDifferent(resource, encryptedPayload, existing, tx.EncryptionStrategy())
```
clarafu, line 1535: nit: 
```suggestion
			configsDiffer, err := configsDifferent(resource, encryptedPayload, existing, es)
```

atc/db/team.go
clarafu: Is there a reason why you want to avoid doing a `.Valid` check on them?

atc/db/team.go
```diff
@@ -1567,19 +1500,178 @@ func inactivateTableForPipeline(tx Tx, pipelineID int, tableName string) error {
...
+	rows, err := insertQuery.RunWith(tx).Query()
+	if err != nil {
+		return nil, err
+	}
+	for rows.Next() {
+		var (
+			name string
+			id   int
+		)
+		rows.Scan(&name, &id)
```
clarafu, line 1569:
```suggestion
		err = rows.Scan(&name, &id)
		if err != nil {
		         return nil, err
	         }
```

clarafu (commented):
It's looking great! I just had a few nits and some small fixes. Once we get those fixed I can do some acceptance testing and then hopefully it should be good to go :)

atc/db/team.go
```diff
@@ -1567,19 +1500,178 @@ func inactivateTableForPipeline(tx Tx, pipelineID int, tableName string) error {
...
+
+		r.ResourceConfigID, _ = rcID.Value()
+		r.ResourceConfigScopeID, _ = rcScopeID.Value()
+		resources[name] = r
+	}
+
+	return resources, nil
+}
+
+func configsDifferent(resourceConfig atc.ResourceConfig, encryptedBlob string, existing *existingResource, es encryption.Strategy) (bool, error) {
```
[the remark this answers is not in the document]
andy-paine: I think we can just scrap the pointer altogether - we know if we've added it to the map then it must exist so just make the map `map[string]existingResource` 🙂

andy-paine (commented):
I've addressed all those points except the `Valid` stuff - let me know if you need me to change that or if it is okay as it is

andy-paine:
Hi @clarafu just a heads up that this is my last day on this engagement with Depop so I will have limited capacity to keep PRs up to date and make changes after today. Is there anything else I need to do to get this PR merged?

rymir:
@clarafu 👋 what are the current blockers before getting this one in? we were hoping to get this one in for 7.7.0 😿 , but I see we didn't make it, can we try targeting 7.8.0 as we are running a custom fork in Depop atm 😨 thanks!

clarafu (approved):
Super sorry it took me so long to review this, this was pushed back in priority for the release so I had to get through other prs/issues first.

I just did some acceptance testing and ran through the code again and everything looks like it is running great! Thanks so much for being patient and for this work!!

PR #8015, opened 2022-01-27
Show var source error on resource and build page

### LINKED ISSUE #7927: Invalid var_sources causes build to error without any hint
## Summary

If a pipeline contains an invalid `var_sources`, then build will immediately fail without any error message. As users cannot get hint of why build fails, they may have to ask support from OPs, which would significantly increase OPs cost.

## Steps to reproduce

Submit the following pipeline to a cluster:

```yaml
var_sources:
- name: vs
  type: vault
  config:
    url: http://fake-vault.com
    path_prefix: some
    client_token: ((my-vault-token))

jobs:
- name: job
  plan:
  - task: task1
    config:
      platform: linux
      image_resource:
        type: mock
        source: {mirror_self: true}
      run:
        path: echo
        args: [((vs:arg))]
```

Then trigger a build of the job, the job will immediately fail without any hint message:

build-error-no-log

From ATC logs, I can find an error:

```
web_1     | {"timestamp":"2022-01-06T03:18:54.401787200Z","level":"error","source":"atc","message":"atc.tracker.notify.run.failed-to-create-run-state","data":{"build":"3","build_id":34,"error":"evaluate var_source 'vs' error: undefined vars: my-vault-token","job":"job","pipeline":"tiv","session":"26.7.2","team":"main"}}
web_1     | {"timestamp":"2022-01-06T03:18:54.412388000Z","level":"info","source":"atc","message":"atc.tracker.notify.run.finish.errored","data":{"build":"3","build_id":34,"error":"evaluate var_source 'vs' error: undefined vars: my-vault-token","job":"job","pipeline":"tiv","session":"26.7.2.3","team":"main"}}
```

## Expected results

If an error is caused by pipeline definition, users should be able to view the error, so that users get idea of how to fix the error.

## Actual results

Users cannot see any error.

## Triaging info

* Concourse version: 7.6.0/master branch
* Browser (if applicable): NA
* Did this used to work?

xtremerui:

On build page, add 'do' step to step trees so error event could
be linked to it correctly.

Screen Shot 2022-01-27 at 11 59 55 AM

On resource page, the check was created async and hoping backend
to finish the check build. But if var source failed in first, the
check build was never be able to run the `check` script. Now after
the check request, ui will ask for check build plan explicitly to
get the errored build events.

Screen Shot 2022-01-27 at 11 59 44 AM

Note: the auto ticking to udpate resource check status relys on if the resource having a build linked. In this case, the check build errored on the first stage by invalid var interpolation, so the build is never ran and linked to the resource. One will need to manually trigger the check to see to error.

closes #7927

* [x] done
* [ ] todo

## Release Note
 * Now error caused by variable interpolation can be shown correctly on resource and build page.

### DIFF
--- a/atc/db/build_in_memory_check.go
+++ b/atc/db/build_in_memory_check.go
@@ -419,10 +419,6 @@ func (b *inMemoryCheckBuild) AcquireTrackingLock(logger lager.Logger, interval t
 }
 
 func (b *inMemoryCheckBuild) Finish(status BuildStatus) error {
-	if !b.runningInContainer {
-		return nil
-	}
-
 	b.status = status
 	b.endTime = time.Now()
 
@@ -431,6 +427,13 @@ func (b *inMemoryCheckBuild) Finish(status BuildStatus) error {
 		return err
 	}
 	defer Rollback(tx)
+
+	if !b.dbInited {
+		err := b.initDbStuff(tx)
+		if err != nil {
+			return err
+		}
+	}
 
 	err = b.saveEvent(tx, event.Status{
 		Status: atc.BuildStatus(status),
--- a/atc/db/build_in_memory_check.go
+++ b/atc/db/build_in_memory_check.go
@@ -277,17 +277,19 @@ type inMemoryCheckBuild struct {
 }
 
 func newRunningInMemoryCheckBuild(conn Conn, lockFactory lock.LockFactory, checkable Checkable, plan atc.Plan, spanContext SpanContext, seqGen util.SequenceGenerator) (*inMemoryCheckBuild, error) {
+	timeNow := time.Now()
+
 	build := &inMemoryCheckBuild{
 		inMemoryCheckBuildForApi: inMemoryCheckBuildForApi{
 			id:        0,
 			conn:      conn,
 			checkable: checkable,
 			plan:      plan,
-			startTime: time.Now(),
+			startTime: timeNow,
 			status:    BuildStatusStarted,
 		},
 		lockFactory: lockFactory,
-		createTime:  time.Now(),
+		createTime:  timeNow,
 		spanContext: spanContext,
 		preId:       seqGen.Next(),
 		eventIdSeq:  util.NewSequenceGenerator(0),
@@ -302,7 +304,7 @@ func newRunningInMemoryCheckBuild(conn Conn, lockFactory lock.LockFactory, check
 
 	build.SaveEvent(event.Status{
 		Status: atc.StatusStarted,
-		Time:   time.Now().Unix(),
+		Time:   timeNow.Unix(),
 	})
 
 	return build, nil
--- a/atc/db/build_in_memory_check.go
+++ b/atc/db/build_in_memory_check.go
@@ -419,6 +419,10 @@ func (b *inMemoryCheckBuild) AcquireTrackingLock(logger lager.Logger, interval t
 }
 
 func (b *inMemoryCheckBuild) Finish(status BuildStatus) error {
+	if !b.runningInContainer && status == BuildStatusSucceeded {
+		return nil
+	}
+
 	b.status = status
 	b.endTime = time.Now()
 
--- a/atc/db/build_in_memory_check.go
+++ b/atc/db/build_in_memory_check.go
@@ -206,6 +206,7 @@ func (b *inMemoryCheckBuildForApi) Events(from uint) (EventSource, error) {
 					err = psql.Select("1").
 						From("resources").
 						Where(sq.Eq{"id": b.resourceId}).
+						Where(sq.Eq{"in_memory_build_status": BuildStatusStarted}).
 						Where(sq.Eq{"in_memory_build_id": buildID}).
 						RunWith(tx).
 						QueryRow().
@@ -286,7 +287,7 @@ func newRunningInMemoryCheckBuild(conn Conn, lockFactory lock.LockFactory, check
 			checkable: checkable,
 			plan:      plan,
 			startTime: timeNow,
-			status:    BuildStatusStarted,
+			status:    BuildStatusPending,
 		},
 		lockFactory: lockFactory,
 		createTime:  timeNow,
@@ -314,7 +315,7 @@ func (b *inMemoryCheckBuild) RunStateID() string {
 	return fmt.Sprintf("in-memory-check-build:%v", b.preId)
 }
 
-func (b *inMemoryCheckBuild) IsRunning() bool           { return b.status == BuildStatusStarted }
+func (b *inMemoryCheckBuild) IsRunning() bool           { return b.endTime.IsZero() }
 func (b *inMemoryCheckBuild) IsManuallyTriggered() bool { return false }
 func (b *inMemoryCheckBuild) CreateTime() time.Time     { return b.createTime }
 
@@ -437,6 +438,18 @@ func (b *inMemoryCheckBuild) Finish(status BuildStatus) error {
 	})
 	if err != nil {
 		return err
+	}
+
+	// Update in memory build status in resources table
+	if b.resourceId != 0 {
+		_, err := psql.Update("resources").
+			Set("in_memory_build_status", status).
+			Where(sq.Eq{"id": b.resourceId}).
+			RunWith(tx).
+			Exec()
+		if err != nil {
+			return err
+		}
 	}
 
 	// Release the containers using in this build, so that they can be GC-ed.
@@ -577,6 +590,7 @@ func (b *inMemoryCheckBuild) initDbStuff(tx Tx) error {
 			Set("in_memory_build_id", b.id).
 			Set("in_memory_build_start_time", b.StartTime()).
 			Set("in_memory_build_plan", b.PublicPlan()).
+			Set("in_memory_build_status", BuildStatusStarted).
 			Where(sq.Eq{"id": b.resourceId}).
 			RunWith(tx).Exec()
 		if err != nil {
--- a/atc/db/build_in_memory_check_test.go
+++ b/atc/db/build_in_memory_check_test.go
@@ -127,6 +127,46 @@ var _ = Describe("Build", func() {
 				Expect(err.Error()).To(Equal("no build event"))
 			})
 
+			Context("BeforeCheckBuildStart", func() {
+				Context("Finish", func() {
+					Context("when build finishes with status succeed", func() {
+						BeforeEach(func() {
+							err := build.Finish(db.BuildStatusSucceeded)
+							Expect(err).ToNot(HaveOccurred())
+						})
+
+						It("does not init DB", func() {
+							Expect(build.ID()).To(Equal(0))
+						})
+					})
+
+					Context("when build finishes with status errored", func() {
+						BeforeEach(func() {
+							err := build.Finish(db.BuildStatusErrored)
+							Expect(err).ToNot(HaveOccurred())
+						})
+
+						It("update in-memory build status for resource", func() {
+							reloaded, err := defaultResource.Reload()
+							Expect(reloaded).To(BeTrue())
+							Expect(err).ToNot(HaveOccurred())
+							Expect(defaultResource.BuildSummary().Status).To(Equal(atc.StatusErrored))
+						})
+
+						It("save build status", func() {
+							eventSource, err := build.Events(3)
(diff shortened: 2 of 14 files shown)


clarafu:
Sorry, I just wanted to confirm with you that I'm understanding this properly. Is the fix here that the user will have to manually check the resource in order to see the error? Is there no way we can show the error without having the user need to manually run a check?

I think it would be best if we can expose the error to the user without requiring them to manually check the resource since this is how it worked before the 7.6.0 changes that caused this bug. I think all the parts are there but we just need to modify some stuff to make it save the in memory build events in the database and then tie it to the resource using the `in_memory_build_id` on the resource/resourcetype/prototypes column. 

There must be something different between a regular build and an in memory build where we don't tie the build to the resource until after the resource check is run/finished. Because if I go to a version of concourse prior to 7.6 (without the in memory check changes), I can see that the error appears because the check build is tied to the resource through the `build_id` column on the resources table and we save the error into the `check_build_events` table when the var fails to interpolate or whenever a build fails to run during the setup stage. But for in memory checks (post 7.6) the `in_memory_build_id` column is empty on the resources table when a check fails so there is no way to tie the check build events that saves the error to the resource check build.

xtremerui:
Yeah I noticed the comment in 
https://github.com/concourse/concourse/blob/4ec23fd7c6011d3ba6be7f6ba019d411ec53129e/atc/exec/check_step.go#L180-L183

xtremerui:
@clarafu I have done a deep dive into this issue and found the root cause. I am writing up a summary.

xtremerui:
I have pushed a commit that *partly* fixes the resource page var source error displaying issue. But there are caveats.

 1 . By now with the latest fix, the error will show properly as normal. However, the error will be flicking every 5-15 secs due to the check running in server. The reason is when a check in memory happens and it fails var source interpolation, there is no resource config scope created the the check. So there wont be any build infos like it used to have in the resource query

https://github.com/concourse/concourse/blob/223485fe795e1eb9e2bc3892ce3bcc878c1cc4a2/atc/db/resource.go#L130-L134

everytime scanner ticks, it will consider the check never ran and it will create the check again. From the UI you then see  the resource state is flicking and refreshing every 5-15 sec or so as frontend is quering backend for resource and check build info.

2. In the pipeline view, a resource with var source error won't be shown as yellow. Same reason as 1, since there is no resource config scope created for such resource check, the check build status is lost. In fact, for all in memory check builds, there is no buidl status afterwards. It is inferred from a logic here

https://github.com/concourse/concourse/blob/223485fe795e1eb9e2bc3892ce3bcc878c1cc4a2/atc/db/resource.go#L977-L988

thus for a check build that failed by interpolation (that only build start time is available), the check build status read from UI will be always `started` (versus the error status shown in resource page comes from the build events status).

For caveat 1, I don't think it is really a bug since thats how it works.
For caveat 2, to fix it we have to add a new column `in_memory_build_end_time` to `resource` table. Then we could apply similar method as shown in 2 to determine a check build status to be errored if it only has start time but no end time (as var source interpolation failed in between).

I am not sure if caveat 2 worth a fix that needs migration in this case.

clarafu:
Great job having the error message be able to be shown even without manually triggering! I just tested it out and it works great, the refreshing/flickering of the error message is not a problem at all. For the second problem, I'm not sure if just an end time will work because if the check is currently running then it will have a started time and not end time. I feel like we would need something more like a column for status on the resources table. We could think about adding that rather than doing all this extra logic here https://github.com/concourse/concourse/blob/223485fe795e1eb9e2bc3892ce3bcc878c1cc4a2/atc/db/resource.go#L977-L1013
I think most of that logic is for coming up with a build status because I think we were trying to avoid needing to update the resources table after an in memory check has finished. But I can't think of any way around it if we want to show the proper status 🤔

atc/db/build_in_memory_check.go
```diff
@@ -206,6 +206,7 @@ func (b *inMemoryCheckBuildForApi) Events(from uint) (EventSource, error) {
 					err = psql.Select("1").
 						From("resources").
 						Where(sq.Eq{"id": b.resourceId}).
+						Where(sq.Eq{"in_memory_build_status": BuildStatusStarted}).
```
xtremerui, line 209: Without this line, the build events of  a resource that is with var source error will not be closed from server, so UI will keep showing `loading ...` . The reason is no resource config scope was created due to var error and this query will actually find one record, thus it will return completed as false.
https://github.com/concourse/concourse/blob/4e6ad40006517d4dcf61cc8146125bf82831456f/atc/db/build_in_memory_check.go#L215

atc/exec/check_step.go
```diff
@@ -99,6 +99,10 @@ func (step *CheckStep) run(ctx context.Context, state RunState, delegate CheckDe
 		"step-name": step.plan.Name,
 	})
 
+	if err := delegate.BeforeSelectWorker(logger); err != nil {
```
evanchaoli, line 102: I have concern to pull up `BeforeSelectWorker` to before `delegate.Initializing()`. Looking at

https://github.com/concourse/concourse/blob/fa6ef67835be414278cba484305a8e656ce7a9b3/atc/engine/build_step_delegate.go#L171-L185

`BeforeSelectWorker` means a worker is needed, which implied a real check needs to run. So when only when an in-memory-check really needs to run check, it interacts with DB. If pull up `BeforeSelectWorker`, then all in-memory-check will init db, which will completely eliminate performance improvement of in-memory-check.

atc/db/build_in_memory_check.go
```diff
@@ -416,10 +419,6 @@ func (b *inMemoryCheckBuild) AcquireTrackingLock(logger lager.Logger, interval t
 }
 
 func (b *inMemoryCheckBuild) Finish(status BuildStatus) error {
-	if !b.runningInContainer {
```
xtremerui, line 419: @evanchaoli I think your concern is valid so I have updated it to only init the DB while it indeed need to record the build event.

The only thing here is `runningInContainer`, while an error like var source interpolation happens, that value here is false (since the build hasn't really being ran in a container yet). So I have to remove this check to allow it to init the DB.

evanchaoli: If you remove `runningInContainer`, then all non-run checks will also write to db, which still doesn't resolve my concern. I think you can do something like:

```golang
if !b.runningInContaienr && status == SUCCEED
```

atc/db/build_in_memory_check_test.go
```diff
@@ -245,6 +254,14 @@ var _ = Describe("Build", func() {
...
 					})
 
+					It("update in-memory build status for resource", func() {
+						reloaded, err := defaultResource.Reload()
+						Expect(reloaded).To(BeTrue())
+						Expect(err).ToNot(HaveOccurred())
+
+						Expect(defaultResource.BuildSummary().Status).To(Equal(atc.StatusSucceeded))
+					})
+
```
evanchaoli, line 264: I think we need to add two test cases when `OnCheckBuildStart` is not called:

1. Finish with "succeeded", no db operation should be done, this is for non-run checks.
2. Finish with an error, db operation should be done, this is for the problem this pr is fixing.

atc/db/build_in_memory_check.go
xtremerui: In engine.go it only calls b.finish when there is error that needs to be recorded. Why `all non-run checks` would call finish since they haven't started to run yet? Is there somewhere else that method is used?

atc/db/build_in_memory_check_test.go
xtremerui: I am a bit confused, when you say `non-run checks` do you mean a succeeded check build that saves its build events to memory? If yes, call it `non-db checks` might be better.

`non-run checks` to me means a check build that picked up by engine but failed to run the actual `check` script in `exec/check_step.go`, like the check with var error in this case.

atc/db/build_in_memory_check.go
xtremerui: Also, `Finish` clean up entries in containers and resource caches table, if we gate it by `succeeded`, what will happen to the containers and caches used for the succeeded check build?

evanchaoli: We don't simply gate it by `succeeded`. When 'runningInContainer` is true, it's not gated, so clean up will still run.

atc/db/build_in_memory_check_test.go
evanchaoli: I meant the cases where `checkDelegate.WaitToRun()` returns false, thus no need to start a container and run check.

xtremerui: > `checkDelegate.WaitToRun()` returns false

I suppose you meant this here
https://github.com/concourse/concourse/blob/0aeac8807b40acfcc9b806ab37e6286cc16ed785/atc/exec/check_step.go#L146

if `run` is fales, it goes to else block to store the found version and then
https://github.com/concourse/concourse/blob/0aeac8807b40acfcc9b806ab37e6286cc16ed785/atc/exec/check_step.go#L233

since check delegate doesn have an implementation of `Finished`, it goes to the one in build step delegate
https://github.com/concourse/concourse/blob/0aeac8807b40acfcc9b806ab37e6286cc16ed785/atc/engine/build_step_delegate.go#L151

which will then save the finish event to the build using `SaveEvent` here
https://github.com/concourse/concourse/blob/0aeac8807b40acfcc9b806ab37e6286cc16ed785/atc/db/build_in_memory_check.go#L518

I might be missing some important part, but the non-run checks you meant are not relevant to the `Finish` method that I changed. Could you give some more details?

atc/db/build_in_memory_check.go
xtremerui: I see. So if `runningInContainer` is false and it is succeeded, it means the check build finds a version so it doesn't need to create a container to run the check, and thus it won't need to run clean up in `Finish`.

atc/engine/check_delegate.go
```diff
@@ -70,6 +70,13 @@ func (d *checkDelegate) Initializing(logger lager.Logger) {
 	logger.Info("initializing")
 }
 
+func (d *checkDelegate) BeforeSelectWorker(logger lager.Logger) error {
```
evanchaoli, line 73: Why this is needed? It has been implemented in build_step_delegate.

xtremerui: Ah it was added by previous commit that for init DB before var source interpolation. It is not needed anymore.

clarafu (changes requested):
I tested this out manually and it all works really nicely! I tested it out with the following cases:
- a resource with failing var in source config
- a resource sharing the same resource config (with global resources) + failing var
- a resource type with failing var

The only kind of weird thing that I don't think is a problem is that the error from the resource type var interpolation shows up in the resource's error

image

Other than that, I think after we clean up the code and also adding those test cases that Evan mentioned, it should be good to merge!

atc/db/build_in_memory_check_test.go
```diff
@@ -126,6 +127,46 @@ var _ = Describe("Build", func() {
 				Expect(err.Error()).To(Equal("no build event"))
 			})
 
+			Context("BeforeCheckBuildStart", func() {
```
xtremerui, line 130: @clarafu these are the two tests that evanchaoli mentioned.

atc/db/resource_test.go
```diff
@@ -195,7 +195,6 @@ var _ = Describe("Resource", func() {
 					Expect(buildSummary.Name).To(Equal(db.CheckBuildName))
 					Expect(buildSummary.TeamName).To(Equal(scenario.Team.Name()))
 					Expect(buildSummary.PipelineName).To(Equal(scenario.Pipeline.Name()))
-					Expect(buildSummary.Status).To(Equal(atc.StatusFailed))
```
clarafu, line 198: Should we be testing it to return the status that is in the db? (Pending build?)

xtremerui: There is no such status in db since the build (id 99) never exists. The status in the removed line comes from 
https://github.com/concourse/concourse/blob/a188c2a9d671597cd3f724807800731a6df563be/atc/db/resource.go#L1003

and 

https://github.com/concourse/concourse/blob/a188c2a9d671597cd3f724807800731a6df563be/atc/db/resource.go#L133

since in the test when initializing the resource config scope the field never got set to true. And there is no proper way to set it othan than DB surgery..

clarafu (approved):

PR #8031, opened 2022-02-02
Add build event for volume streaming

### LINKED ISSUE #6575: Missing information when resources are streamed
## Summary

Currently it is not obvious which resources are streamed and which are local, yet the worker is stalling for some reason.

Currently the UI only shows the worker name, but often things are stalling for a considerable time with large~ish images relative to the available bandwidth.

## Expected results

Shows from which worker a resource is streamed and the progress.

## Actual results

No output, no visible progress.

## Triaging info

* Concourse version: 7.0.0
* Browser (if applicable): Firefox 85.0.1
* Did this used to work?  No

andy-paine:

Often build steps appear to "hang" in the UI after a worker is selected
which is commonly due to volumes being streamed. As these volumes can be
quite large, this can take a significant amount of time that would be
helpful to tell the user about.

This adds a "streaming-volume" event to the build logs which outputs the
"input name" (directory that it will be mounted at) and the source and
destination worker names to help users understand when jobs are waiting
for volumes to be streamed.

The current worker structure did not really have a particularly good
place to save this event so I kept the "delegate" pattern and passed
that down into the atc/worker code. Due to cyclic dependencies this has
to be in the atc/runtime package but I think that is fine anyway as the
runtime.BuildStepDelegate only specifies the delegate methods needed at
runtime etc.

closes #6575

* [x] done
* [ ] todo

## Notes to reviewer
The delegate being passed all the way down seems like an elegant-ish hack but open to other suggestions

## Release Note
* Build logs will now contain new events when a volume is being streamed to a worker
Screenshot 2022-02-02 at 11 39 01

### DIFF
--- a/atc/runtime/types.go
+++ b/atc/runtime/types.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"strings"
 
+	"code.cloudfoundry.org/lager"
 	"github.com/concourse/concourse/atc/compression"
 	"github.com/concourse/concourse/atc/db"
 	"go.opentelemetry.io/otel/propagation"
@@ -27,7 +28,7 @@ type Worker interface {
 	// It can be thought of as declaratively saying "I want a container
 	// matching these specifications" - and the Worker implementation will
 	// "make it so", regardless of what Containers/Volumes already exist.
-	FindOrCreateContainer(context.Context, db.ContainerOwner, db.ContainerMetadata, ContainerSpec) (Container, []VolumeMount, error)
+	FindOrCreateContainer(context.Context, db.ContainerOwner, db.ContainerMetadata, ContainerSpec, BuildStepDelegate) (Container, []VolumeMount, error)
 	// CreateVolumeForArtifact creates a new empty Volume to be used as a
 	// WorkerArtifact. This is used for uploading local inputs to a worker via
 	// `fly execute -i ...`.
@@ -121,6 +122,10 @@ type ContainerSpec struct {
 	// CertsBindMount indicates whether or not to mount the worker's Certs
 	// volume onto the container.
 	CertsBindMount bool
+}
+
+type BuildStepDelegate interface {
+	StreamingVolume(lager.Logger, string, string, string)
 }
 
 // ContainerSpec must implement propagation.TextMapCarrier so that it can be
--- a/atc/runtime/types.go
+++ b/atc/runtime/types.go
@@ -281,13 +281,23 @@ type ContainerLimits struct {
 // Artifact represents an output from a step that can be used as an input to
 // other steps.
 type Artifact interface {
+	// StreamOut converts the contents of the Artifact under path to a compressed
+	// tar stream. The result of StreamOut can be passed in to StreamIn for
+	// another Artifact.
+	//
+	// path is a relative path - "." indicates using the root of the Artifact.
 	StreamOut(ctx context.Context, path string, compression compression.Compression) (io.ReadCloser, error)
+
+	// Handle gives the globally unique ID of the Volume.
+	Handle() string
+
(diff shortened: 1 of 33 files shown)


atc/worker/gardenruntime/worker.go
```diff
@@ -43,6 +44,7 @@ type DB struct {
 	WorkerTaskCacheFactory        db.WorkerTaskCacheFactory
 	ResourceCacheFactory          db.ResourceCacheFactory
 	WorkerBaseResourceTypeFactory db.WorkerBaseResourceTypeFactory
+	BuildFactory                  db.BuildFactory
```
andy-paine, line 47: This is a mistake

atc/runtime/types.go
```diff
@@ -276,26 +281,30 @@ type ContainerLimits struct {
...
+
+	// Source gives the original source of the artifact e.g. worker name
+	Source() string
 }
 
 // Volume represents a data volume that can be mounted to a Container.
 type Volume interface {
-	// Handle gives the globally unique ID of the Volume.
-	Handle() string
+	Artifact
```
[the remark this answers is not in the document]
andy-paine: Currently in lots of places that we use `runtime.Artifact` we end up casting it to `runtime.Volume` to be able to call functions to get the handle and source worker. The `runtime.Artifact` endpoint is designed to be a future abstraction that only `gardenruntime.Volume` currently implements and it seems reasonable that any future implementation should also be able to say what its unique identifier is `Handle()` and where it came from `Source()`. By moving these functions up to the `runtime.Artifact` interface, we avoid having to cast it to `runtime.Volume` in places like this https://github.com/concourse/concourse/pull/8031/files#diff-2034974614561f2b31bff2d4e9572aa96a162533d58b49a6cc06cfd74625d905R522

atc/syslog/drainer.go
```diff
@@ -173,6 +173,16 @@ func (d *drainer) sendEvent(logger lager.Logger, build db.Build, syslog *Syslog,
...
+	case event.EventTypeStreamingVolume:
+		var streamingVolumeEvent event.StreamingVolume
+		err := json.Unmarshal(*ev.Data, &streamingVolumeEvent)
+		if err != nil {
+			logger.Error("failed-to-unmarshal", err)
+			return err
+		}
+		ts = time.Unix(streamingVolumeEvent.Time, 0)
+		tag = build.SyslogTag(streamingVolumeEvent.Origin.ID)
+		message = fmt.Sprintf("streaming volume %s from %s to %s", streamingVolumeEvent.Volume, streamingVolumeEvent.SourceWorker, streamingVolumeEvent.DestWorker)
```
xtremerui, line 185:
```suggestion
		message = fmt.Sprintf("streaming volume %s from worker %s to %s", streamingVolumeEvent.Volume, streamingVolumeEvent.SourceWorker, streamingVolumeEvent.DestWorker)
```
I like the way `streaming volume xxx` so i know `xxx` is a volume. Could we also do that for worker?

Another question is the `to %s` part, is there a case that destination worker is different to the worker in `selected worker` event just above? If not, I feel `streaming volume xxx from worker xxx` might be good enough.

andy-paine:
Any idea why this PR's pipeline is archived - the checks won't run 😞 ? https://ci.concourse-ci.org/teams/contributor/pipelines/pr?vars.number=8031

xtremerui (approved):

wanderanimrod:
@andy-paine , do we know the size of the volume being streamed at the location we are logging this? It would be very nice to indicate the size in human-readable units so that users have more data while optimizing their pipelines. Just asking before I open a new issue for this. If it is too hard, I'd rather not open an issue that will be stale for long. Thanks.

PR #7991, opened 2022-01-22
WIP: dirty changes of 7.6.0 performance optimization.

evanchaoli:
## Changes proposed by this PR

Optimize 7.6.0 performance.

* [ ] done
* [ ] todo

## Notes to reviewer

This might should have been submit to `master` branch, but as my work is done on branch `7.6.x`, this PR for now is just showing my idea of changes. 

## Release Note

* 

[imperative mood]: https://chris.beams.io/posts/git-commit/#imperative

PR #8042, opened 2022-02-06
reduced in-memory checks

evanchaoli:

This is derived from #7991. Instead of putting everything in a messy PR, I decided to wrap each piece of enhancement into a separate PR.

This PR reduced total count of Lidar checks by pulling check-interval test to earlier before creating a check go-routine.

The reason why testing check-interval is placed inside `checkDelegate.WaitToRun()` is because resource config scope is needed. 

This PR is based a theory that if a resource is not updated, then it's resource config scope will not be changed. If a pipeline is not updated, then resources of the pipeline will not be updated. In Lidar scanner, if a resource is not updated, and it doesn't reach to next check time, then skip the resource.

This enhancement will save huge amount of unnecessary checks (go-routines, db locks, etc.)

* [x] code changes
* [x] add tests

## Notes to reviewer

NA

## Release Note

Enhanced Lidar to reduce checks so that improved ATC performance.

### DIFF
--- a/atc/db/check_factory.go
+++ b/atc/db/check_factory.go
@@ -26,7 +26,7 @@ type Checkable interface {
 	Tags() atc.Tags
 	CheckEvery() *atc.CheckEvery
 	CheckTimeout() string
-	LastCheckEndTime() time.Time
+	TimeToCheck() bool
 	CurrentPinnedVersion() atc.Version
 
 	HasWebhook() bool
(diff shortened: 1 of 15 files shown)


xtremerui:
> n Lidar scanner, if a resource is not updated, and it doesn't reach to next check time, then skip the resource.

why does a resource is updated or not matters here? In the `TryCreateCheck` method, it only checks if it is manually triggered and if the check interval time elapse

evanchaoli:
> why does a resource is updated or not matters here? In the `TryCreateCheck` method, it only checks if it is manually triggered and if the check interval time elapse

The logic is:

- "check interval time elapse" is based on last check end time
- Last check end time is stored on resource configure scope
- Resource configure scope is only populated check `checkStep.Run()`

Lidar has no ability to populate a resource's configure scope. But if a resource is not updated, then Lidar may consider its current configure scope is up-to-date. That's reason why resource update time matters in Lidar.

evanchaoli:
@xtremerui I did a refactor, but code is dirty. Please take a preview. I'll clean up the code and update tests tomorrow.

clarafu:
Hmm I think your assumption of "if a pipeline is not updated, the scope will be unchanged" is partly true for most cases. If the resource is using a base resource type then this is true but if the resource is using a custom resource type then this could not be true. It can not be true for custom resource types because if a custom resource type gets a new version then all the resources that use that custom type will get new resource config scopes. And a custom resource type can get a new version even without a pipeline being updated.

evanchaoli:
> And a custom resource type can get a new version even without a pipeline being updated.

@clarafu I was aware of that. Not only a custom resource type getting new version, but if a resource uses vars from Vault (or other cred manager), scope may also change without updating the pipeline. But those should be rare cases, and in that cases, it would just take one check interval to refresh the resource's scope. Comparing the performance gain, I think it's worthy.

evanchaoli:
@xtremerui @clarafu I have updated tests. So this PR is ready for review. Basically, this PR does the best thing over the few PRs I submitted recently.

PR #8043, opened 2022-02-06
Change nested checks to reuse check containers

evanchaoli:

This is derived from https://github.com/concourse/concourse/pull/7991. Instead of putting everything in a messy PR, I decided to wrap each piece of enhancement into a separate PR.

When `put: some-custome-resource`, the `put` step will check a nested `check` for the custom resource-type, then implied `get` should not duplicately run the check again. But without this fix, `put` and implied `get` will run check twice.

https://github.com/concourse/concourse/commit/7ef20bb8dcd623e87c8399a96c9a4377a626c7b4 has broken the logic of 

https://github.com/concourse/concourse/blob/2fc810fcc27d6edcdeb5a651b3b657b45458e739/atc/plan.go#L303-L305

because the commit inject `resource_type` to check plan in all cases.

As there could still be manual resource type checks issued by `fly check-resource-type`, and manually 

* [x] done

## Release Note

Fixed a bug where nested checks were run duplicately.

### DIFF
--- a/atc/engine/check_delegate.go
+++ b/atc/engine/check_delegate.go
@@ -125,7 +125,7 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
 	interval := d.plan.Interval.Interval
 
 	var lock lock.Lock = lock.NoopLock{}
-	if d.plan.IsPeriodic() {
+	if d.build.Name() == db.CheckBuildName {
 		for {
 			run, err := d.shouldPeriodicCheckRun(scope, interval)
 			if err != nil {
--- a/atc/engine/check_delegate_test.go
+++ b/atc/engine/check_delegate_test.go
@@ -381,6 +381,7 @@ var _ = Describe("CheckDelegate", func() {
 
 		Context("when not running for a resource", func() {
 			BeforeEach(func() {
+				fakeBuild.NameReturns("1")
 				plan.Check.Resource = ""
 			})
 
--- a/atc/exec/check_step.go
+++ b/atc/exec/check_step.go
@@ -136,11 +136,9 @@ func (step *CheckStep) run(ctx context.Context, state RunState, delegate CheckDe
 	// Point scope to resource before check runs. Because a resource's check build
 	// summary is associated with scope, only after pointing to scope, check status
 	// can be fetched.
-	if step.plan.Resource != "" {
-		err = delegate.PointToCheckedConfig(scope)
-		if err != nil {
-			return false, fmt.Errorf("update resource config scope: %w", err)
-		}
+	err = delegate.PointToCheckedConfig(scope)
+	if err != nil {
+		return false, fmt.Errorf("update resource config scope: %w", err)
 	}
 
 	lock, run, err := delegate.WaitToRun(ctx, scope)
--- a/atc/exec/check_step_test.go
+++ b/atc/exec/check_step_test.go
@@ -462,6 +462,7 @@ var _ = Describe("CheckStep", func() {
 				Context("when the plan is nested", func() {
 					BeforeEach(func() {
 						checkPlan.Resource = ""
+						checkPlan.ResourceType = "some-resource-type"
 
 						expectedOwner = db.NewBuildStepContainerOwner(
 							501,
@@ -485,8 +486,8 @@ var _ = Describe("CheckStep", func() {
 						fakePool.FindOrSelectWorkerReturns(chosenWorker, nil)
 					})
 
-					It("not points the resource or resource type to the scope", func() {
-						Expect(fakeDelegate.PointToCheckedConfigCallCount()).To(Equal(0))
+					It("points the resource type to the scope", func() {
+						Expect(fakeDelegate.PointToCheckedConfigCallCount()).To(Equal(1))
 					})
 
 					It("uses delegate's container owner", func() {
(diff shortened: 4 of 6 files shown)


evanchaoli:
@clarafu For this PR, before you review the code change, let's rephrase the logic.

In current implementation, periodical check (triggered by Lidar) will acquire `resource-checking-lock`, and reuse check containers (resource config + scope as container owner). For nested check, it doesn't acquire `resource-checking-lock` and always create a new container.

I'm afraid the current logic is not ideal. Say two pipelines use the same custom resource type, say `github`, and each pipeline triggers a build at almost the same time, say pipeline1 is slightly earlier. As pipeline1 has been running a check for `github` resource type, for pipeline2, is waiting for the existing check to finish or creating a new container faster? I guess former should be faster.

My idea of enhancement:

In general there are 4 types of checks:

- Lidar triggered resource checks
- Manually trigged resource checks or resource type checks
- Normal build nested checks (should only be resource type checks)
- Check build nested checks (should only be resource type checks)

more to add ...

clarafu:
I think I'm a little confused because I would assume that the second check for the same custom type would end up not actually running the check because of this code path

https://github.com/concourse/concourse/blob/b67033906d8ef2b126b0dce0d52c8ef29b37f086/atc/engine/check_delegate.go#L203-L207

evanchaoli:
@clarafu when the second check reaches to the code piece you mentioned, it is the bug. A nested check should go with:

https://github.com/concourse/concourse/blob/e1e184ee7caaf0b244209fb004d69cf074c62f77/atc/engine/check_delegate.go#L170-L181

With this PR, I'm afraid I have went beyond the original scope that it supposes to be. With this PR, I am now actually wanting to do an enhancement of letting nested checks also reusing check containers. But this needs more tests to prove if it is really helpful.

So let me submit a new PR for the simple bug fix, and leave this PR for the enhancement.

clarafu:
I think this is less of a bug and more of a result of the changes with https://github.com/concourse/concourse/pull/7176. I actually mention in the pr description that 

> The behavioural change is that every time we run a step, we will actually run a check for the parent type. That check will still respect the checking interval/check every so that will ensure we don't check more times than the checking interval. 

I think the code path that you linked is actually never used because now, every check plan will include either a resource, resource type or prototype (because of the #7176 pr changes).

I actually have a second question to what exactly this PR is trying to solve because from my understanding, if we have a put step for a resource that is using a resource type, only one check for the resource type from both the get and put step should actually run. This is because the put step should first do a check for its image, which this check will actually run and result in a new version. Then the get step will run and do a check, but it will see that the interval has not elapsed yet because of the check that ran in the put step incremented the `LastCheck` value, so it will actually not run a check and just use the latest result.

evanchaoli:
@clarafu Let's clarify the logic in `WaitToRun()`:

https://github.com/concourse/concourse/blob/8eecb168189ea6adcb574ef32f800953becb1768/atc/engine/check_delegate.go#L128-L181

where, if a check if for Lidar triggered resource, it should go to branch of line 129. A nested resource type check should go to line 171. The theory behind is that, Lidar triggered resource check reuses container, it needs to acquire lock. But nested resource type check always creates a new container, thus no need to acquire lock:

https://github.com/concourse/concourse/blob/8eecb168189ea6adcb574ef32f800953becb1768/atc/exec/check_step.go#L313-L331

Also, for nested build, we should not use check interval, instead, if a build, a check should run only once. For example, default check interval is 1 minute, many `put` may take longer than 1 minute, if we use check interval, then implied `get` will run `check` again, which is not much reasonable. 

But with #7176, check plan is always added the field `ResourceType`, thus

https://github.com/concourse/concourse/blob/8eecb168189ea6adcb574ef32f800953becb1768/atc/plan.go#L303-L305

will always return true. This is the bug that #8048 is fixing.

clarafu:
I'm pretty sure it was https://github.com/concourse/concourse/pull/6854 that added the locking around resource types in order to avoid having multiple resource type checks running at the same time, and then we can just use the result of the one check for the rest of the checks. I think that is useful so that we arn't running multiple resource type checks at the same time.

I do agree with you that within one build, we probably don't need to check a resource type multiple times. I say that because I believe that within a single build, resources and resource types should only be using a single version and that should be trusted. If the thing you are trying to fix is the fact that we might run a check for a resource type multiple times within the same build because the build takes longer than the check interval, can we just add a little more logic so that we still respect the check interval but only run if the last check time is earlier than the build start time? For example, in this else https://github.com/concourse/concourse/blob/e3d9f5bb1d41e56b5de584e2b96505a90f1fe2b6/atc/engine/check_delegate.go#L202-L208 it can also do a check similar to this https://github.com/concourse/concourse/blob/e3d9f5bb1d41e56b5de584e2b96505a90f1fe2b6/atc/engine/check_delegate.go#L178-L180 for anything other than resources (resource-types and prototypes) and it exits early if the last check time is later than build start time. What do you think about that or am I misinterpreting something?

evanchaoli:
@clarafu The fix of #8048 is right to let nested check to go through 

https://github.com/concourse/concourse/blob/e3d9f5bb1d41e56b5de584e2b96505a90f1fe2b6/atc/engine/check_delegate.go#L178-L180

From this perspective, we are aligned. Please focus on #8048 only for 7.7 and leave this PR open.

clarafu:
Going to close this because https://github.com/concourse/concourse/pull/8048 applies the same changes to master!

PR #8047, opened 2022-02-08
Do not duplicate streaming of volumes in parallel when EnableCacheStreamedVolumes is true

andy-paine:
When many steps start in parallel on a single worker that share a volume, they will currently all attempt to stream the volume in parallel. With `atc.EnableCacheStreamedVolumes = true`, these streamed volumes will then be marked as resource caches. As these caches would then be valid local copies of the volume, we can just stream a volume to a worker once and make other steps wait for that stream to finish. This removes the likelyhood of the "ton of streaming out" from this comment as it will only stream once per worker that requires it rather than once per started task: 
https://github.com/concourse/concourse/blob/b67033906d8ef2b126b0dce0d52c8ef29b37f086/atc/exec/get_step.go#L275-L280

* [x] merge #8031 
* [x] use a lock to ensure only one step streams a volume to a worker
* [x] extend to image streaming
* [x] add a new build event so that users know what is happening  

## Notes to reviewer
Draft as it is currently based off of #8031 to reuse delegate code.

This is a relatively large refactor of the volume streaming code with a couple of changes which we think are safe to do but could do with validating. The basic idea is:
* With `atc.EnableCacheStreamedVolumes = true`, streamed volumes are marked as caches and so are effectively the same as local volumes after streaming
* Rather than have multiple steps stream an identical volume, use a global lock
* "Streaming remote volumes" is now basically "either wait or stream volumes until they all become local"

Because of this, we no longer separate between volumes that were already on the worker (no streaming or waiting necessary) and volumes that were streamed either directly or by another step. This used to be represented by the `inputDestinationPaths` map which seems to track volumes that were already local as they can be reused as outputs. It seems that this couldn't be done with streamed volumes in the past:
https://github.com/concourse/concourse/blob/b67033906d8ef2b126b0dce0d52c8ef29b37f086/atc/worker/gardenruntime/worker.go#L421-L425
but we believe that all the streamed volumes are COW-ed anyway so it should be fine to reuse the _mounted_ volume (since that is not a cache). As such, we add all inputs to the `inputDestinationPaths` rather than separate "already local" and "streamed but now local"

Other than that, things are mostly just a refactor to abstract away the idea of "already local" and "streamed but now local" into a single `findOrStreamVolume` method and goroutines are used to parallelise the streaming and/or waiting (as streaming already did).

## Release Note
* Steps that stream volumes will now use a global (per worker) lock to ensure identical volumes are not streamed more times than they need to be
* A new `waiting-for-streamed-volume`/`waiting for volume <name> to be streamed by another step` event is included in build step logs where this behaviour occurs

### DIFF
--- a/atc/worker/gardenruntime/gardenruntimetest/baggageclaim.go
+++ b/atc/worker/gardenruntime/gardenruntimetest/baggageclaim.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strings"
+	"sync"
 
 	"github.com/concourse/concourse/atc/runtime/runtimetest"
 	"github.com/concourse/concourse/worker/baggageclaim"
@@ -14,9 +15,12 @@ import (
 
 type Baggageclaim struct {
 	Volumes []*Volume
+	Mutex   sync.Mutex
 }
 
 func (b *Baggageclaim) FindVolume(handle string) (*Volume, int, bool) {
+	b.Mutex.Lock()
+	defer b.Mutex.Unlock()
 	for i, v := range b.Volumes {
 		if v.handle == handle {
 			return v, i, true
@@ -37,6 +41,8 @@ func (b *Baggageclaim) FilteredVolumes(pred func(*Volume) bool) []*Volume {
 
 func (b *Baggageclaim) AddVolume(volume *Volume) *Volume {
 	_, i, ok := b.FindVolume(volume.handle)
+	b.Mutex.Lock()
+	defer b.Mutex.Unlock()
 	if ok {
 		b.Volumes[i] = volume
 		return volume
@@ -67,6 +73,8 @@ func (b *Baggageclaim) LookupVolume(_ context.Context, handle string) (baggagecl
 }
 
 func (b *Baggageclaim) DestroyVolumes(_ context.Context, handles []string) error {
+	b.Mutex.Lock()
+	defer b.Mutex.Unlock()
 	for _, handle := range handles {
 		b.DestroyVolume(context.Background(), handle)
 	}
--- a/atc/worker/gardenruntime/image.go
+++ b/atc/worker/gardenruntime/image.go
@@ -3,12 +3,10 @@ package gardenruntime
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 	"io"
 	"net/url"
 	"path"
 
-	"code.cloudfoundry.org/lager"
 	"code.cloudfoundry.org/lager/lagerctx"
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db"
@@ -40,19 +38,48 @@ func (worker *Worker) fetchImageForContainer(
 	delegate runtime.BuildStepDelegate,
 ) (FetchedImage, error) {
 	logger := lagerctx.FromContext(ctx)
+
 	if imageSpec.ImageArtifact != nil {
-		volume, ok, err := worker.findVolumeForArtifact(ctx, teamID, imageSpec.ImageArtifact)
+		volume, err := worker.findOrStreamVolume(ctx, imageSpec.Privileged, teamID, container, imageSpec.ImageArtifact, "/", delegate)
 		if err != nil {
-			logger.Error("failed-to-locate-artifact-volume", err)
+			logger.Error("failed-to-find-or-stream-volume-for-image", err)
 			return FetchedImage{}, err
 		}
-		if ok && volume.DBVolume().WorkerName() == worker.Name() {
-			// it's on the same worker, so it will be a baggageclaim volume
-			volumeOnWorker := volume.(Volume)
-			return worker.imageProvidedByPreviousStepOnSameWorker(ctx, logger, imageSpec.Privileged, teamID, container, volumeOnWorker)
-		} else {
-			return worker.imageProvidedByPreviousStepOnDifferentWorker(ctx, imageSpec.Privileged, teamID, container, imageSpec.ImageArtifact, delegate)
+
+		imageVolume, err := worker.findOrCreateCOWVolumeForContainer(
+			ctx,
+			imageSpec.Privileged,
+			container,
+			volume,
+			teamID,
+			"/",
+		)
+		if err != nil {
+			logger.Error("failed-to-create-cow-volume-for-image", err)
+			return FetchedImage{}, err
 		}
+
+		imageMetadataReader, err := worker.streamer.StreamFile(ctx, imageSpec.ImageArtifact, ImageMetadataFile)
+		if err != nil {
+			logger.Error("failed-to-stream-metadata-file", err)
+			return FetchedImage{}, err
+		}
+
+		metadata, err := loadMetadata(imageMetadataReader)
+		if err != nil {
+			return FetchedImage{}, err
+		}
+
+		imageURL := url.URL{
+			Scheme: RawRootFSScheme,
+			Path:   path.Join(imageVolume.Path(), "rootfs"),
+		}
+
+		return FetchedImage{
+			Metadata:   metadata,
+			URL:        imageURL.String(),
+			Privileged: imageSpec.Privileged,
+		}, nil
 	}
 
 	if imageSpec.ResourceType != "" {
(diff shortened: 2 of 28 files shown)


evanchaoli:
@andy-paine Thank you so much for this PR. I lost sleep last night for thinking how to make the change, and found this PR this morning.

@clarafu @xtremerui All my load tests have proven that performance of 7.x (with p2p-streaming and enable-cache-streamed-volume) is X times better than 6.7.8 **once workers warmed up**. However, on a busy cluster, workers warming up is really a problem, that was one of reasons we failed to upgrade to 7.6 last time.

After upgrade from 6.x to 7.x, all workers are fresh. A common image will be fetched by one worker, then rest workers will all be streaming the image volume from the first worker. In theory, if there are 100 workers in total, then total streaming session should be no more than 99. However, say worker A has 10 builds started, all of them need the image, then worker A may initiate 10 streaming sessions. In my load tests, I have observed that, the source worker ran hundreds `gzip` and `tar` processes, the worker CPU was exhausted, so that image streaming might take very long time, 10 minutes, even longer, and led to various errors. Though slow, and some failures, some steaming would succeed, workers slowly got warmed up, then 7.x ran much faster then 6.x.

This PR should prevent too much streaming sessions from a worker. So I think this PR is critical.

atc/worker/gardenruntime/worker.go
```diff
@@ -713,3 +640,89 @@ func (worker *Worker) GardenClient() gclient.Client {
...
+		if err != nil {
+			logger.Error("failed-to-find-volume-for-artifact", err)
+			return Volume{}, err
+		}
+		if ok && localVolume.DBVolume().WorkerName() == worker.Name() {
+			volume := localVolume.(Volume)
+			return volume, nil
+		}
+
+		lock, acquired, err := worker.acquireVolumeStreamingLock(ctx, artifact.Handle(), container.WorkerName())
```
evanchaoli, line 687: I think we should use resource cache id for the lock.

Say, worker-1 and worker-2 both have an image, and now 2 builds on worker-3 need the image. The two builds' nested `image-get` may return different artifacts from worker-1 and worker-2, but they actually point to the same image. In this case, if we use volume handle for lock, worker-3 will still duplicate streaming the image.

atc/worker/gardenruntime/gardenruntimetest/baggageclaim.go
```diff
@@ -7,16 +7,20 @@ import (
...
 	"strings"
+	"sync"
 
 	"github.com/concourse/concourse/atc/runtime/runtimetest"
 	"github.com/concourse/concourse/worker/baggageclaim"
 )
 
 type Baggageclaim struct {
 	Volumes []*Volume
+	Mutex   sync.Mutex
```
evanchaoli, line 18: Should `FilteredVolumes()` also be mutex protected? In that case, `RWMutex` might be a better fit. Or maybe we can try https://github.com/cornelk/hashmap. As Volumes are compared by handle, I guess using a hashmap with handle as key might be more performant.

evanchaoli:
I applied the current diff to my load tests. After redeployed all 100 workers, workers got warmed up much quicker than before. I logged in a few workers, and count total `tar`s, and counts never exceed 99.

atc/worker/gardenruntime/worker.go
andy-paine: Yeah that's a nice extra optimization. It means casting from `runtime.Artifact` to `runtime.Volume` again which feels pretty bad (all artifacts currently are volumes) but we can make it work easily enough. I know you bumped into similar issues with the original cache-streamed-volumes work.

atc/worker/gardenruntime/gardenruntimetest/baggageclaim.go
andy-paine: Probably can yeah. To be honest since this is test code I just made it thread-safe enough to pass the `-race` tests - the real baggageclaim deals with this with proper locks and things.

evanchaoli: @andy-paine To clarify, if you mean that eventually this PR will not include changes in this file?

atc/worker/gardenruntime/worker.go
```diff
@@ -713,3 +640,97 @@ func (worker *Worker) GardenClient() gclient.Client {
...
 }
+
+func (worker *Worker) acquireVolumeStreamingLock(
+	ctx context.Context,
+	resourceCacheID int,
+	destination string,
+) (lock.Lock, bool, error) {
+	// if caching of streamed volumes is not enabled then we should not globally
+	// lock the volume streaming since steps cannot share the streamed volumes as caches
+	if !atc.EnableCacheStreamedVolumes {
```
evanchaoli, line 651: Say a `get` is followed by a `in_parallel`, and `in_parallel` has 3 `task`s that all take the `get` resource as input. When the 3 tasks happen to place on the same worker, it will initiate 3 volume streams. In that case, should this lock also help?

atc/worker/gardenruntime/worker.go
```diff
@@ -713,3 +640,97 @@ func (worker *Worker) GardenClient() gclient.Client {
...
+		var resourceCacheLock lock.Lock
+		var acquired bool
+		if ok {
+			resourceCacheLock, acquired, err = worker.acquireVolumeStreamingLock(ctx, volume.dbVolume.GetResourceCacheID(), container.WorkerName())
+		} else {
+			// If the artifact is not a volume we cannot cache it so always stream
+			resourceCacheLock, acquired = lock.NoopLock{}, true
+		}
+		if err != nil {
+			logger.Error("failed-to-acquire-lock-for-volume-streaming", err)
```
evanchaoli, line 697: Maybe add some data with the error, like resource cache id.

atc/worker/gardenruntime/worker.go
```diff
@@ -713,3 +640,97 @@ func (worker *Worker) GardenClient() gclient.Client {
...
+				logger.Error("failed-to-create-image-artifact-replicated-volume", err)
+				return Volume{}, err
+			}
+
+			delegate.StreamingVolume(logger, inputPath, artifact.Source(), streamedVolume.DBVolume().WorkerName())
+			if err := worker.streamer.Stream(ctx, artifact, streamedVolume); err != nil {
+				logger.Error("failed-to-stream-artifact", err)
+				return Volume{}, err
+			}
+			logger.Debug("streamed-non-local-volume")
```
evanchaoli, line 721: Add some data like streamedVolume handle, dest worker name.

atc/worker/gardenruntime/worker.go
```diff
@@ -713,3 +640,97 @@ func (worker *Worker) GardenClient() gclient.Client {
...
+			return streamedVolume, nil
+		}
+
+		if tick == nil {
+			delegate.WaitingForStreamedVolume(logger, inputPath, container.WorkerName())
+			tick = time.NewTicker(WaitingForStreamedVolumePollingInterval)
```
evanchaoli, line 727: Do we need to `defer tick.Stop()`.

atc/worker/gardenruntime/worker.go
```diff
@@ -23,6 +25,8 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+var WaitingForStreamedVolumePollingInterval = 5 * time.Second
```
evanchaoli, line 28: Is 5 seconds too long? Some volume could be very small, especially with p2p streaming, a volume streaming could be very fast. With a 5 seconds timer, once a step needs to wait for streaming, it would take more than 5 seconds. I guess 2 seconds might be better.

atc/worker/gardenruntime/gardenruntimetest/baggageclaim.go
andy-paine: The PR tests failed due to race conditions https://ci.concourse-ci.org/teams/contributor/pipelines/pr/jobs/unit/builds/1?vars.number=8047#L61fdb20d:1096 so that's why I introduced this in this PR as otherwise it wouldn't pass checks

atc/worker/gardenruntime/worker.go
andy-paine: If `EnableCacheStreamedVolumes` is not enabled then they should all stream since the volume streamed for the first task isn't a valid cache for the second task. In the case where `EnableCacheStreamedVolumes` is enabled then the lock should kick in and only one of those tasks will actually do the streaming and the other 2 will wait.

atc/worker/gardenruntime/worker.go
andy-paine: This is all logged by the streamer already - not sure if we really need it again here? https://github.com/concourse/concourse/blob/master/atc/worker/streamer.go#L41-L46

atc/worker/gardenruntime/worker.go
andy-paine: I wasn't sure since each loops does the `findVolumeForArtifact` and potentially tries to acquire the lock which could get a bit aggressive on the DB if many tasks are waiting for streaming in parallel?

atc/worker/gardenruntime/gardenruntimetest/baggageclaim.go
evanchaoli: That's weird. I applied this PR to my local and ran unit test, but didn't hit the race error at all. Can you try remove the change and see if unit test still hit the error?

atc/worker/gardenruntime/worker.go
evanchaoli: ok, then just ignore this comment.

atc/worker/gardenruntime/worker.go
evanchaoli: Is it possible to change to 1 or 2 seconds and try in your environment to see that would lead to more db loads?

andy-paine:
Could someone un-archive and retrigger https://ci.concourse-ci.org/teams/contributor/pipelines/pr/jobs/integration/builds/5?vars.number=8047 please?

xtremerui (approved):

xtremerui:
TBO I can't completly understand all the implementation details in this PR. However, I think the problem it tries to solve is solid and it covers the scenarios in tests.

Given it only affects the behaviour when `EnableCacheStreamedVolumes ` is true, I tend to merge it and deploy it to hush-house to verify further.

atc/worker/gardenruntime/worker.go
```diff
@@ -382,55 +386,42 @@ type mountableLocalInput struct {
...
 
-		if ok && volume.DBVolume().WorkerName() == worker.Name() {
-			localInputs = append(localInputs, mountableLocalInput{
-				cowParent: volume.(Volume),
-				mountPath: input.DestinationPath,
-			})
-		} else {
-			remoteInputs = append(remoteInputs, remoteInput{
-				volume:    input.Artifact,
+		g.Go(func() error {
```
clarafu, line 410: Do we need to pass in the `input`/`i` variables into the argument of the goroutines? Just so that it will get the correct value within every loop.

```suggestion
		g.Go(func(i int, input []runtime.Input) error {
```

atc/worker/gardenruntime/worker.go
```diff
@@ -713,3 +640,100 @@ func (worker *Worker) GardenClient() gclient.Client {
...
+	artifact runtime.Artifact,
+	mountPath string,
+	delegate runtime.BuildStepDelegate,
+) (Volume, error) {
+	logger := lagerctx.FromContext(ctx)
+	var tick *time.Ticker
+
+	_, inputPath := path.Split(mountPath)
+	if inputPath == "" {
+		inputPath = "for image"
```
clarafu, line 673: Sorry I'm not very familiar with the worker code, could you explain a bit as to what this is used for just for my own curiousity? Thanks!

atc/worker/gardenruntime/image.go
```diff
@@ -40,19 +38,48 @@ func (worker *Worker) fetchImageForContainer(
...
+			ctx,
+			imageSpec.Privileged,
+			container,
+			volume,
+			teamID,
+			"/",
+		)
+		if err != nil {
+			logger.Error("failed-to-create-cow-volume-for-image", err)
+			return FetchedImage{}, err
```
clarafu, line 59: nit! Just something that might be useful that we had in the `imageProvidedByPreviousStepOnSameWorker` method previously. If you think itll be helpful, can also add extra error printing info for error handling of `loadMetadata` and `StreamFile`.

```suggestion
			return FetchedImage{}, fmt.Errorf("create COW volume: %w", err)
```

clarafu (commented):
Hi! I think it makes a lot of sense to lock around streaming a volume for the same resource cache, thanks for your work :)

I had a quick question though, about your comment in the description
> but we believe that all the streamed volumes are COW-ed anyway so it should be fine to reuse the mounted volume (since that is not a cache). As such, we add all inputs to the inputDestinationPaths rather than separate "already local" and "streamed but now local"

I don't think I fully understand what you mean by reusing the mounted volume. I was wondering if you meant that there were any behavioural changes that came from that? Because when I look at the code I only really see code being refactored but no real behavioural changes (other than the locking). I totally could have missed something though!

atc/worker/gardenruntime/worker.go
evanchaoli: I think this is just for display in UI.

atc/worker/gardenruntime/image.go
xtremerui: @evanchaoli I am not sure if @andy-paine could still pick this up, would you be able to update this PR to move it forward? thx

evanchaoli: @xtremerui I'm willing to move forward this PR, but I'm not aware of how to update other's PR. Only way I'm realizing is to check-pick all commits and create a new PR, but that sounds stealing @andy-paine 's work. The other way is that, as the comments are all nit issues, you just go ahead merge this PR, then I can use a new PR to address the comments. What's your suggestion?

xtremerui: I think you can checkout this PR.

```
git checkout -b depop-avoid-duplicating-parallel-volume-streams master
git pull https://github.com/depop/concourse.git avoid-duplicating-parallel-volume-streams
```
Then you can create a new branch from it and make changes. Submit your branch as a new PR. All commit history will be there and we just need to close this PR and continue on yours.

evanchaoli: @xtremerui Okey, your way is better than cherry-pick. Following your way, I have created #8322.

@clarafu Addressed this comment in #8322.

atc/worker/gardenruntime/worker.go
evanchaoli: @clarafu This is not needed. If you look at a few lines back, `i` and `input` have been made as local copy within the `for` loop. But anyway, I have changed it to a more comprehensive implementation in #8322.

atc/worker/gardenruntime/worker.go
evanchaoli: Resolved.

atc/worker/gardenruntime/worker.go
evanchaoli: Resolved.

atc/worker/gardenruntime/worker.go
evanchaoli: Skip this comment.

xtremerui:
Close for #8322

PR #8048, opened 2022-02-09
Fix bug of resource type checks

evanchaoli:

Refer to https://github.com/concourse/concourse/pull/8043#issue-1125126046
And https://github.com/concourse/concourse/pull/8043#issuecomment-1033418261

This bug causes nested resource type check to need acquiring resource-config-check-lock. One issue during our last 7.6 upgrade failure was that some step nested check took very long time to start, I think that should be due to this bug.

* [x] done

## Release Note

- Remove lock on checking resource types and prototypes, this will result in a small behaviour change. For example, if you have multiple resources that use the same resource type and those resources run a check for that resource type at the same time, the resource type will be checked multiple times. This was the behaviour before 7.3.0 so we are reverting back to this behaviour.
- Fixes a bug introduced in 7.6.0 where resource type `resource_config_id` were never updated.

### DIFF
--- a/atc/engine/check_delegate.go
+++ b/atc/engine/check_delegate.go
@@ -72,7 +72,7 @@ func (d *checkDelegate) Initializing(logger lager.Logger) {
 
 func (d *checkDelegate) FindOrCreateScope(config db.ResourceConfig) (db.ResourceConfigScope, error) {
 	var resourceIDPtr *int
-	if d.plan.Resource != "" {
+	if d.plan.IsResourceCheck() {
 		pipeline, err := d.pipeline()
 		if err != nil {
 			return nil, fmt.Errorf("get pipeline: %w", err)
@@ -125,7 +125,7 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
 	interval := d.plan.Interval.Interval
 
 	var lock lock.Lock = lock.NoopLock{}
-	if d.plan.IsPeriodic() {
+	if d.plan.IsResourceCheck() {
 		for {
 			run, err := d.shouldPeriodicCheckRun(scope, interval)
 			if err != nil {
--- a/atc/engine/check_delegate.go
+++ b/atc/engine/check_delegate.go
@@ -110,7 +110,7 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
 		if d.plan.Interval.Never {
 			// exit early if user specified to never run periodic checks
 			return nil, false, nil
-		} else if d.plan.Resource != "" {
+		} else if d.plan.IsResourceCheck() {
 			// rate limit periodic resource checks so worker load (plus load on
 			// external services) isn't too spiky. note that we don't rate limit
 			// resource type or prototype checks, because they are created every time a
--- a/atc/exec/check_step.go
+++ b/atc/exec/check_step.go
@@ -170,7 +170,7 @@ func (step *CheckStep) run(ctx context.Context, state RunState, delegate CheckDe
 
 		metric.Metrics.ChecksStarted.Inc()
 
-		_, buildId, err := delegate.UpdateScopeLastCheckStartTime(scope, (step.plan.Resource == ""))
+		_, buildId, err := delegate.UpdateScopeLastCheckStartTime(scope, !step.plan.IsResourceCheck())
 		if err != nil {
 			return false, fmt.Errorf("update check start time: %w", err)
 		}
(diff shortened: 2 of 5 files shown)


clarafu:
@evanchaoli I left a comment https://github.com/concourse/concourse/pull/8043#issuecomment-1042109991 in the other PR, please take a quick look at it! I think I'd like to figure out exactly what the problem is and what you are seeing in your deployment.

clarafu:
We can talk through it in this PR! I'm not sure if we are completely aligned because I still think that we should be locking on resource types/prototypes because of the reasons mentioned in this comment https://github.com/concourse/concourse/pull/8043#issuecomment-1049210977. I think with the changes in this PR it will cause the resource types to be checked multiple times when they are run at the same time. For example, in a build that has 4 in parallel resource gets for the same resource type it ends up checking the same resource type for each check 

image

vs with the resource lock it will only check once

image

So what I was hoping to do is while it is trying to acquire the lock, within `shouldPeriodicCheckRun(` it will exit early for resource-types and prototypes if the last check end time is later than the build start time.

evanchaoli:
> I think with the changes in this PR it will cause the resource types to be checked multiple times when they are run at the same time.

@clarafu This is true but I think this the design since @vito added nested check in 7.0.0. If you look at 7.0.0 code:

https://github.com/concourse/concourse/blob/35e971a307d349be791178c5a3d17f24c966c8c1/atc/engine/check_delegate.go#L99-L113

Locks are only acquired for resource checks. The other evidence is:

https://github.com/concourse/concourse/blob/35e971a307d349be791178c5a3d17f24c966c8c1/atc/exec/check_step.go#L328-L350

Resource check reuses containers, thus acquiring lock has to be done. But nested check doesn't reuse container (`BuildStepContainerOwner`), thus no need to acquire lock.

Actually, in our last 7.6 upgrade failure, one thing we observed that, a lot of builds were stuck at nested check for long time, that is reason why I found this bug.

I always have the concern that nested checks will generate too many containers. I am also with you that nested check may also acquire lock and reuse check containers, that's reason why I proposed https://github.com/concourse/concourse/pull/8043#issuecomment-1032156348. A potential risk that, when we acquire lock and reuse lock, if a nested check happens to hang due to some reason (maybe worker falls into a bad state, maybe a transient network problem, maybe whatever other problem...) all builds that need the resource type will be blocked. Thus, though I expect that is a big enhancement, but it's a big change. Without plenty of tests, I have no confident to merge it. Thus I won't expect #8043 to catch up with 7.7 or even 7.8.

This PR is only to keep the current design and fixed the small bug. I have been running my load tests by applying this PR in my private build.

atc/engine/check_delegate.go
```diff
@@ -125,7 +125,7 @@ func (d *checkDelegate) WaitToRun(ctx context.Context, scope db.ResourceConfigSc
 	interval := d.plan.Interval.Interval
```
clarafu, line 125: nit: Seems like I can't reach the line but on line 113, we can probably change 
```
} else if d.plan.Resource != "" {
```
to 
```
} else if d.plan.IsResourceCheck() {
```

atc/exec/check_step.go
```diff
@@ -136,11 +136,9 @@ func (step *CheckStep) run(ctx context.Context, state RunState, delegate CheckDe
 	// Point scope to resource before check runs. Because a resource's check build
 	// summary is associated with scope, only after pointing to scope, check status
 	// can be fetched.
-	if step.plan.Resource != "" {
```
clarafu, line 139: Is this a bug fix for something else? Where we never update the `resource_config_id` of resource types because of this if statement?

clarafu (commented):
Right, its actually always been that we never acquired resource type check locks until this pr https://github.com/concourse/concourse/pull/6854 made the change. I saw that you made this comment here https://github.com/concourse/concourse/pull/6854#discussion_r619198579 about adding the checking lock for optimization reasons? I think if the purpose was adding for optimization reasons and it won't break any existing logic, I think it makes sense to remove the checking lock then. I was only concerned about the resource type checking lock being added for some architectural reason and also that if we have multiple resource type checks happening at once, it might clobber the versions being saved if they end up finding different versions. But I don't think that will be much of a problem because this was the behaviour before. Also, I agree that removing the resource type checking lock will help speed up nested checks when workers are being slow so I think the benefits outweigh the negatives. I just have a few small questions left and then we can get this merged in for 7.7!

atc/engine/check_delegate.go
evanchaoli: Yeap.

atc/exec/check_step.go
evanchaoli: Yes, I think so. This is a bug introduced by #7208 (in-memory check) in 7.6.0. With this bug, `resource_config_id` is never updated in table `resource_types`.

evanchaoli: @clarafu BTW, #8049 is an optimization that avoids duplicate updating scopes.

clarafu (approved):
LGTM! Thanks!! :)

PR #8049, opened 2022-02-09
db optimization: if scope is unchanged, check step can skip scope update

evanchaoli:

A check step always update scope of a resource, but in most of case, a resource's scope should be unchanged. There is an optimization at SQL level:

https://github.com/concourse/concourse/blob/e1e184ee7caaf0b244209fb004d69cf074c62f77/atc/db/resource.go#L275-L279

But that there is still a SQL request sending to PG server. By recording current scope id in check plan, then check step may know if scope is changed, if not, then don't call the db function at all.

Given count of check builds is far more than normal builds, this optimization should save a lot of db transactions.

* [x] code for main logic
* [x] add tests

## Notes to reviewer

I have done a load test, comparing 6.7.8, current master branch and this PR.

In the test, I used 3 ATCs and 100 workers. There are 180 pipelines, each has two resources, all resources have different scopes, so totally 360 resources and 360 scopes. Each pipelines triggers a build every 10 minutes.

The test shows that 7.x has much better performance than 6.7.8. We can see that, with 6.7.8, CPU usages are almost 100%, but 7.x is mostly below 50%. For DB queries, QPS of 7.x is also much lower than 6.7.8. With this PR, QPS looks a little lower than master branch.

## Release Note

Optimized db performance by reducing unnecessary resource scope updates.

### DIFF
--- a/atc/config.go
+++ b/atc/config.go
@@ -200,7 +200,7 @@ type ResourceType struct {
 	Tags       Tags        `json:"tags,omitempty"`
 	Params     Params      `json:"params,omitempty"`
 
-	CurrentResourceConfigScope int
+	CurrentResourceConfigScope int `json:"-"`
 }
 
 type Prototype struct {
--- a/atc/config.go
+++ b/atc/config.go
@@ -199,6 +199,8 @@ type ResourceType struct {
 	CheckEvery *CheckEvery `json:"check_every,omitempty"`
 	Tags       Tags        `json:"tags,omitempty"`
 	Params     Params      `json:"params,omitempty"`
+
+	CurrentResourceConfigScope int
 }
 
 type Prototype struct {
@@ -314,6 +316,8 @@ func (types ResourceTypes) ImageForType(planID PlanID, resourceType string, step
 		Source: parent.Source,
 		Params: parent.Params,
 		Tags:   parent.Tags,
+
+		CurrentResourceConfigScope: parent.CurrentResourceConfigScope,
 	}
 
 	getPlan, checkPlan := FetchImagePlan(planID, imageResource, types.Without(parent.Name), stepTags, skipInterval, parent.CheckEvery)
@@ -386,6 +390,8 @@ func FetchImagePlan(planID PlanID, image ImageResource, resourceTypes ResourceTy
 				Tags: tags,
 
 				SkipInterval: skipInterval,
+
+				CurrentResourceConfigScope: image.CurrentResourceConfigScope,
 			},
 		}
 		maybeCheckPlan = &checkPlan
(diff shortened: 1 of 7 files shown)


clarafu:
I think I'm a little conflicted about this change for a few reasons. One of them is because we are adding this new field onto the resource type config and the check plan purely for optimization reasons. IMO I think if possible, the atc plan and config structs should only contain fields that define the object itself. But I can see myself swaying on this if the optimization does improve things a lot and there is no other option. But I find that  the only real improvement is less db queries and I'm not sure if that is enough to change my mind because a slightly higher db query count does not really affect much unless we see performance degradations from the high number of queries being run.

Is there a reason you chose this specific query to optimize by not having to run this? Did you run any db schema queries that showed you that this query was the bottleneck in your test environment? Whenever I run the query for finding the queries that take up the most total time, I see this updating scope query to be really low on the list.

evanchaoli:
@clarafu 

> Is there a reason you chose this specific query to optimize by not having to run this? 

I happened to notice this query, and realized it could be an optimization. I wanted to do that because:

* Unlike both ATCs and workers being horizontally scalable, PG is hard to scale out. And Concourse is highly rely on PG, thus I think every optimization for PG is worthwhile.
* This optimization is inside check step. Total count of checks are much bigger than other steps.

To your concern, I think `json:"-"` helps indicate the field doesn't belong to resource type config and check plan.

clarafu:
Do you think you could share the output of the schema inspection query for that query? It should be something like this

```
SELECT
  total_time / calls AS avg_time,
  calls,
  total_time,
  rows,
  100.0 * shared_blks_hit / nullif(shared_blks_hit + shared_blks_read, 0) AS hit_percent,
  regexp_replace(query, '[\s\t\n]+', ' ', 'g')
FROM pg_stat_statements
WHERE query NOT LIKE '%EXPLAIN%'
AND query NOT LIKE '%INDEX%'
AND query LIKE '%UPDATE resources SET resource_config_id = $1, resource_config_scope_id = $2 WHERE name = $3 AND pipeline_id = $4 AND (resource_config_id IS NULL OR resource_config_scope_id IS NULL OR resource_config_id <> $5 OR resource_config_scope_id <> $6) RETURNING id%';
```

Because when I take a look at this query in my stress tests, I don't see it anywhere near the top queries that are taking up the most time out of the database. The reason is because just sending the db request is very lightweight, almost like a select query because we are not actually updating a row. I think if you can show me that the query, even with it not actually needing to update the rows most of the time, does still have a big impact on the database then I can be convinced to change the struct signature. I just feel that if the query does not cause any problems then I don't see the need to add these additional fields to the struct.

evanchaoli:
@clarafu I agree this is a tiny performance improvement. Let's not spending too much time no this PR for now. There are other more important PRs, let's leave it for future discussion.

PR #8057, opened 2022-02-10
atc: Consider image volumes in volume-locality strategy

andy-paine:

Often the largest volume that may need streaming is the image that the
container executes in. By considering this volume too, this may allow
for slightly better placement/reduced streaming for some builds (in
particular builds with no inputs or caches will now prefer to be on the
same worker as their image volume rather than a random one)

I think this should also go some of the way to resolving issues like #6218 as volume-locality should do a better job of using the same worker

* [x] look at the image artifact when considering volumes in volume-locality placement strategy

### DIFF
--- a/atc/worker/placement.go
+++ b/atc/worker/placement.go
@@ -140,16 +140,16 @@ type volumeLocalityStrategy struct{}
 func (strategy volumeLocalityStrategy) Order(logger lager.Logger, pool Pool, workers []db.Worker, spec runtime.ContainerSpec) ([]db.Worker, error) {
 	counts := make(map[string]int, len(workers))
 
-	for _, input := range spec.Inputs {
-		volume, ok := input.Artifact.(runtime.Volume)
+	updateCountsForArtifact := func(artifact runtime.Artifact, destinationPath string) error {
+		volume, ok := artifact.(runtime.Volume)
 		if !ok {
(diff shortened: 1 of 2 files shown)


evanchaoli:
I think this PR should only be merged once #8070 is fixed. Otherwise it may end up that all containers go to the same worker and image gets no chance to be warmed up on other workers.

evanchaoli:
@andy-paine Can you please rebase this PR? Because it doesn't contain change of #8061 but Github seems not detecting that.

atc/worker/placement.go
```diff
@@ -175,12 +175,25 @@ func (strategy volumeLocalityStrategy) Order(logger lager.Logger, pool Pool, wor
...
+				return err
 			}
 			if found {
 				counts[worker.Name()]++
 			}
 		}
+		return nil
+	}
+
+	err := updateCountsForArtifact(spec.ImageSpec.ImageArtifact, "/")
```
evanchaoli, line 187: Usually image is much than other artifacts, can we give image 5 counts?

That way will especially benefit those use case where some job uses huge task images, but the job doesn't run frequently. Current image-get runs on a worker, task runs on the other worker, image streaming take longer time than directly pulling image from registry, because volume streaming involves tar->gzip->ungzip->untar.

taylorsilva:
Superseded by #9188

PR #8061, opened 2022-02-12
Ignore cached input from volume-locality's consideration

### LINKED ISSUE #8070: Volume-locality strategy should ignore inputs of cached resource
## Summary

With `EnableCachedtreamedVolumes` turning on, if `get` step finds a resource cache it's gonna fetch, then it will not do anything, just returned the found volume.

Say, a following task needs the `get` step as an input. If container placement strategy is `volume-locality`, then the task will be placed on the worker where the resource cache volume locates, which is bad, if the resource cache is widely used by many pipelines, because it will end up:

- All containers the need the resource as input will go to the same worker
- The resource volume will get no chance to be streamed to other workers, thus no way to sharing loads across workers.

Let's think about if `EnableCachedtreamedVolumes` off. Then the `get` should go to a `fewest-build-container` worker (#8061), then the `task` should go to the same worker per `volume-locality`.

Thus, when `EnableCachedtreamedVolumes` is on, as `get` does nothing, the `task` should go to a `fewest-build-container` worker, then stream the input volume from source worker. This is still more performant, because streaming a volume between workers should anyway cheaper than running a real `get` step. But for those common resource caches, which are shared by a lot of pipelines, once they are warmed up on most of workers, then build should run much faster.

## Triaging info

* Concourse version: 7.6.0
* Browser (if applicable):
* Did this used to work?

evanchaoli:

Closes #8070

* [x] Ignore cached input from `volume-locality`'s `Order()` 

## Release Note

* When `EnableCacheStreamedVolume` is enabled and container placement strategy is `volume-locality`, as `get` step may not fetch a resource if the resource is found in cache, following step containers may all be placed to the worker where cached resource is found. That worker might be overloaded when there are other workers available. This PR fixes the problem.

### DIFF
--- a/atc/exec/put_step_test.go
+++ b/atc/exec/put_step_test.go
@@ -392,14 +392,14 @@ var _ = Describe("PutStep", func() {
 				})
 			})
 
-			Context("when only some inputs are from cache ", func() {
+			Context("when only inputs are from cache ", func() {
 				BeforeEach(func() {
 					putPlan.Inputs = &atc.InputsConfig{
 						Specified: []string{"input2"},
 					}
 				})
 
-				It("runs with specified inputs", func() {
+				It("runs with cached inputs", func() {
 					Expect(chosenContainer.Spec.Inputs).To(ConsistOf([]runtime.Input{
 						{
 							Artifact:        volume2,
--- a/atc/exec/put_step_test.go
+++ b/atc/exec/put_step_test.go
@@ -139,9 +139,9 @@ var _ = Describe("PutStep", func() {
 		volume2 = runtimetest.NewVolume("volume2")
 		volume3 = runtimetest.NewVolume("volume3")
 
-		repo.RegisterArtifact("input1", volume1)
-		repo.RegisterArtifact("input2", volume2)
-		repo.RegisterArtifact("input3", volume3)
+		repo.RegisterArtifact("input1", volume1, false)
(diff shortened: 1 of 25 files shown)


evanchaoli:
@xtremerui I have reduced scope of this PR to focus on "ignore cached input from volume-locality's consideration". Hope that helps quicker review.

atc/exec/put_step_test.go
```diff
@@ -371,10 +381,30 @@ var _ = Describe("PutStep", func() {
...
+			})
+
+			Context("when only some inputs are from cache ", func() {
+				BeforeEach(func() {
+					putPlan.Inputs = &atc.InputsConfig{
+						Specified: []string{"input2"},
+					}
+				})
+
+				It("runs with specified inputs", func() {
```
xtremerui, line 402: How is a specified input related to cache?

evanchaoli: I think this is a typo. Goal of this test is to make sure when only input is from cache, container spec is well populated, `FromCache` should be true.

xtremerui:
<p float="left">



</p>

<p float="left">



</p>

In acceptance test, the task `unit` is placed in expected workers in first 3 builds while the deployment has 3 workers. In build 4, it reuse a worker with local volume of input and task image so no streaming.

xtremerui (approved):

PR #8063, opened 2022-02-12
Enhance volume-locality strategy

### LINKED ISSUE #8062: Volume-locality strategy is too costly
## Summary

Where there are a big number of worker, 100 workers in my tests, volume-locality strategy is very cost, which exhausts ATC, leads to a lot of more db queries, and eventually makes builds much slower. In my tests, if I use fewest-build-containers, builds can finishes in 1 minutes, but if volume-locality, some builds may take 1 hour even longer.

image

ATC CPUs were almost 100%, but after changed to fewest containers, CPU usages dropped to below 50%.

image

DB queries also dropped a lot after changed to fewest containers.

Looking at code of volume-locality, it iterates all workers, which I think is really a bad implementation:

https://github.com/concourse/concourse/blob/5a97ab10e27a5c74c2a95e24c283bc98cbdd351e/atc/worker/placement.go#L171-L183

## Steps to reproduce

Use 100 workers, and run some workloads, the compare "volume-locality" with "fewest-build-containers".

## Expected results

There should not be much difference wrt metrics.

## Actual results

volume-locality causes much higher ATC CPU usages, a lot of more db queries, slower builds, and so on.

## Triaging info

* Concourse version:
* Browser (if applicable):
* Did this used to work?

evanchaoli:

closes #8062

* [x] done

## Notes to reviewer

Enhanced performance of `volume-locality` strategy.

## Release Note

* Optimize database queries for `volume-locality` container placement strategy

### DIFF
--- a/atc/db/volume_repository.go
+++ b/atc/db/volume_repository.go
@@ -30,6 +30,9 @@ type VolumeRepository interface {
 	FindVolumesForContainer(container CreatedContainer) ([]CreatedVolume, error)
 	GetOrphanedVolumes() ([]CreatedVolume, error)
 
+	FindWorkersForResourceCache(resourceCache ResourceCache) ([]string, error)
+	FindWorkersForTaskCache(taskCache UsedTaskCache) ([]string, error)
+
 	DestroyFailedVolumes() (count int, err error)
 
 	GetDestroyingVolumes(workerName string) ([]string, error)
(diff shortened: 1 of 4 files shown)


atc/db/volume_repository.go
```diff
@@ -465,6 +468,58 @@ func (repository *volumeRepository) FindResourceCacheVolume(workerName string, r
 	return createdVolume, true, nil
 }
 
+func (repository *volumeRepository) FindWorkersForResourceCache(resourceCache ResourceCache) ([]string, error) {
+	rows, err := psql.Select("distinct(c.worker_name)").
```
clarafu, line 472: Hmm I think this will need some changes because it disregards the workers that are passed into the `Order(` method in the `placement.go` file. The list of workers that are passed into that method are a list of workers that are "compatible" (ex, non-team workers, tagged workers, etc) https://github.com/concourse/concourse/blob/dbbed7a7875441fe01784e05e485edb2079779c9/atc/worker/pool.go#L234-L237

evanchaoli: Good cache.

evanchaoli:
> Thx for the PR. Do you have the metrics comparison? Wondering the performance improvement over it.

Yes I have. In my load tests, initially I used "fewest-build-container", then I switched to "volume-locality". From then metrics, ATC CPU Usages were bumped to almost 100%, DB queries were also increase. When used "fewest-build-container", a build took around 1 minute to finish; after switched to "volume-locality", as ATC CPU usages were too high, builds got very slow, some builds took more than 1 hour. 

After I applied this PR and #8061 to the load test, ATC CPU usages dropped a lot, steadily below 50%. DB queries also decreased. And build duration reduced to less than 1 minute.



> Also could you add tests for those two new methods?

I think tests have been covered by `placement_test.go`. Looking at `volume_repository_test.go`, not all functions have tests in it. So I guess, we don't have duplicate tests in `volume_repository_test.go`.

To address @clarafu 's review comment, I found that current `volume-locality` strategy also has a bug. I fixed that, and added some tests in `placement_test.go`.

xtremerui (approved):

evanchaoli:
@xtremerui can you pls merge this one as it's approved?

xtremerui:
@clarafu is also reviewing.

atc/db/volume_repository.go
clarafu: Oh I was actually trying to describe that there is a problem that the list of workers that we query might not be a compatible worker and we were incrementing the `count[worker]++` in the map even if the worker is not a compatible worker. But I noticed you fixed it in your latest commit with https://github.com/concourse/concourse/blob/332618afd26014f612ecf199236e7201d9a16181/atc/worker/placement.go#L187 so great job!

evanchaoli: @clarafu As mentioned in https://github.com/concourse/concourse/pull/8063#discussion_r806320698, incrementing the `count[worker]++` was actually not a bug, because sort is only done on candidate workers, and `count` map only provides supporting data to the sort.

clarafu: Oh I'm sorry I didn't read that part properly earlier, thanks a lot for clarifying it to me again!

clarafu (approved):
The improvements to the queries seem great!

The only thing I can think of that might cause a bit of inaccuracy with this change is the fact that sometimes there are worker resource caches or worker task caches that don't have a volume. Typically this should never happen but I'm guessing when something goes wrong on the worker side and the volume ends up getting deleted, we will end up with worker task caches or worker resource caches that have no volume associated to it. I saw this in a few workers in our deployments, but the workers I saw them on were usually ones that were special cased. But I don't think that should be a huge problem because worst case is that we choose a worker that doesn't actually have the volume on it, which is not that big of a deal. I would much rather have the performance improvements with this change.

PR #8069, opened 2022-02-14
Render nested across steps

### LINKED ISSUE #7564: nested across step display bug
## Summary

When nesting across steps inside across steps, the steps inside the nested across step do not get rendered in the UI, and their output cannot be seen. Based on my testing, the nested steps are still running, and properly report error/success statuses up to the overall build, so it seems like this is just a display issue.

Worked around this by using an additional variable in the across matrix, since that was probably more appropriate for my case, but maybe there's a use case that makes sense for this?

## Steps to reproduce

1. Fly a pipeline that has a job with a nested across step. Here's a pretty contrived example, when I found the issue, the pipeline was a little less apparent that there was an across inside an across:

```
groups:
- name: main
  jobs:
  - unit-tests

jobs:
- name: unit-tests
  plan:
  - across:
    - var: release
      values:
      - {name: cf-networking, db-types: [mysql, postgres]}
      - {name: silk,          db-types: [postgres]}
      max_in_flight: all
    do:
    - task: echo-release
      config:
        image_resource:
          type: registry-image
          source: { repository: alpine }
        platform: linux
        run:
          path: echo
          args:
          - ((.:release.name))
    - across:
      - var: db-type
        values: ((.:release.db-types))
        max_in_flight: all
      do:
      - task: echo-release-and-db-type
        config:
          image_resource:
            type: registry-image
            source: { repository: alpine }
          platform: linux
          run:
            path: echo
            args:
            - ((.:release.name)) on ((.:db-type))

resources: []
```

2. Run the pipeline
3. View the build and verify that the steps inside the nested across do not display in the UI

## Expected results

Either nested across steps should have their steps displayed.

## Actual results

The nested across step itself is displayed, but none of its variables, or sub-steps.

## Triaging info

* Concourse version:
* Browser (if applicable):
* Did this used to work?

areller:

closes #7564 

This PR fixes a rendering issue with nested across steps.

Inside `updateTreeNodeAt`, when matching an Across node, we also update all of the nodes that are nested within that across node. When doing that, we should pass the `id` that we want to update, instead of the `id` (`stepId`) of the across node.

### Before

image

### After

image

## Release Note

 * Fix a rendering issue with nested across steps.

### DIFF
--- a/web/elm/src/Build/StepTree/Models.elm
+++ b/web/elm/src/Build/StepTree/Models.elm
@@ -364,7 +364,7 @@ updateTreeNodeAt id fn tree =
         Across stepId vars vals trees ->
             let
                 withUpdatedChildren =
-                    Across stepId vars vals <| Array.map (updateTreeNodeAt stepId fn) trees
+                    Across stepId vars vals <| Array.map (updateTreeNodeAt id fn) trees
             in
             if stepId == id then
                 fn withUpdatedChildren


xtremerui:
Thx for the fix!

PR #8111, opened 2022-02-21
Allow to specify port fly login will listen for the callback

### LINKED ISSUE #4837: Specify fly_port on fly login command

*Now* I am trying to dockerize some pipeline generation tool for my team to use.
I am trying to include `fly login` in the wrapper docker image, but it is not possible to map a dynamic port statically.

It would be better if we can specify the `fly_port` when executing the `fly_login` command.

Yes. But I barely have spare time nowadays, so it would be better is someone takes this —given it is approved.

databus23:
This PR adds a flag `token-listener` flag to `fly login` to specify the local listen address explicitly.

This enables workflows where fly is running remotely (e.g. via ssh or docker) and the port needs
to be known a priori to configure a static port forward.

closes #4837

## Notes to reviewer
I didn't bother to support this new flag in the legacy auth code path as it seems this is only for really old concourse versions and seems unused at this point.

## Release Note

fly: Add optional --token-listener flag to login command

### DIFF
--- a/fly/commands/login.go
+++ b/fly/commands/login.go
@@ -33,6 +33,7 @@ type LoginCommand struct {
 	ClientCertPath atc.PathFlag `long:"client-cert" description:"Path to a PEM-encoded client certificate file."`
 	ClientKeyPath  atc.PathFlag `long:"client-key" description:"Path to a PEM-encoded client key file."`
 	OpenBrowser    bool         `short:"b" long:"open-browser" description:"Open browser to the auth endpoint"`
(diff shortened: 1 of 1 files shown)


xtremerui:
Hi, thx for the PR. Is it possible to add a test udner https://github.com/concourse/concourse/blob/13fb1e94127635deecba2eb1d7a2357ba9fe85cc/fly/integration/login_test.go#L415 ?

taylorsilva:
Test never added by @databus23 
Feel free to open the PR again if you have time to add tests.

PR #8142, opened 2022-03-03
Force get and check to use "fewest-build-containers" strategy.

evanchaoli:

Currently, `check` step use random strategy to choose worker. If a check container, especially for nested checks, happened to be placed on a overloaded worker, then the entire build might be stuck.

This PR will force `get` and `check` to use `fewest-build-containers` strategy. Because `get` and `check` steps takes no input, they can always go to a least loaded worker.

* [x] done

## Notes to reviewer

@xtremerui @clarafu Code change of this PR is very simple. If possible, please make it in 7.7.

Optimize container placement strategy:
* Force `get` and `check` steps to use "fewest-build-containers" strategy. `check` step used to use random strategy. If an nested check step happens to be sent to a busy worker, that might block entire build. For `get` step, if strategy is `volume-locality`, as `get` takes no input, it will directly pass through `volume-locality` strategy and end up with random strategy. If a `get` happens to be sent to a busy worker, the following `task` and `put` steps that need the `get` as inputs may also be sent to the busy worker due to `volume-locality`, then the busy worker will get busier. With this optimization, `check` and `get` will always be placed on a relatively idle worker.
* Change default strategy from `volume-locality` to `volume-locality,fewest-build-containers`. When a `task` or `put` step takes inputs from a multiple workers, the combined strategy will choose a relatively idle one. More over, when a `task` or `put` takes no input, the combined strategy will ensure the step is placed on a relatively idle worker.

## Release Note

* Reverted in https://github.com/concourse/concourse/pull/8144

### DIFF
--- a/atc/engine/step_factory.go
+++ b/atc/engine/step_factory.go
@@ -24,6 +24,7 @@ type coreStepFactory struct {
 	resourceConfigFactory db.ResourceConfigFactory
 	defaultLimits         atc.ContainerLimits
 	strategy              worker.PlacementStrategy
+	noInputStrategy       worker.PlacementStrategy
 	defaultCheckTimeout   time.Duration
 	defaultGetTimeout     time.Duration
 	defaultPutTimeout     time.Duration
@@ -40,6 +41,7 @@ func NewCoreStepFactory(
 	resourceConfigFactory db.ResourceConfigFactory,
 	defaultLimits atc.ContainerLimits,
 	strategy worker.PlacementStrategy,
+	noInputStrategy worker.PlacementStrategy,
 	defaultCheckTimeout time.Duration,
 	defaultGetTimeout time.Duration,
 	defaultPutTimeout time.Duration,
@@ -55,6 +57,7 @@ func NewCoreStepFactory(
 		resourceConfigFactory: resourceConfigFactory,
 		defaultLimits:         defaultLimits,
 		strategy:              strategy,
+		noInputStrategy:       noInputStrategy,
 		defaultCheckTimeout:   defaultCheckTimeout,
 		defaultGetTimeout:     defaultGetTimeout,
 		defaultPutTimeout:     defaultPutTimeout,
@@ -77,7 +80,7 @@ func (factory *coreStepFactory) GetStep(
 		containerMetadata,
 		factory.lockFactory,
 		factory.resourceCacheFactory,
-		factory.strategy,
+		factory.noInputStrategy,
 		delegateFactory,
 		factory.pool,
 		factory.defaultGetTimeout,
@@ -130,7 +133,7 @@ func (factory *coreStepFactory) CheckStep(
 		stepMetadata,
 		factory.resourceConfigFactory,
 		containerMetadata,
-		nil,
+		factory.noInputStrategy,
 		factory.pool,
 		delegateFactory,
 		factory.defaultCheckTimeout,
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -1027,7 +1027,7 @@ func (cmd *RunCommand) backendComponents(
 		return nil, err
 	}
 
-	buildContainerStrategy, err := cmd.chooseBuildContainerStrategy()
+	buildContainerStrategy, noInputBuildContainerStrategy, err := cmd.chooseBuildContainerStrategy()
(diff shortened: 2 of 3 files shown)


xtremerui:
I think the default worker container strategy is `volume-locality` ?

atc/engine/step_factory.go
```diff
@@ -77,7 +80,7 @@ func (factory *coreStepFactory) GetStep(
 		containerMetadata,
 		factory.lockFactory,
 		factory.resourceCacheFactory,
-		factory.strategy,
```
xtremerui, line 80: I believe this one is using the default strategy `volume-locality`

evanchaoli:
> I think the default worker container strategy is `volume-locality` ?

Yes, it is. But `voume-locality` works based on a step's inputs and `get`/`check` steps have no input, thus `volume-locality` strategy will just pass through.

Looking at the code:

https://github.com/concourse/concourse/blob/c7d131be613089b4973d76cdb3e7a69f2be8950c/atc/worker/placement.go#L79-L107

At line 88, it shuffles candidate workers, then iterate all strategies. As `volume-locality` will just pass through, `get` and `check` step will end up choose the first worker from the shuffled list, which is a random one.

This is true to those `put` and `task` steps that take no input. From this perspective, it would be better to change default strategy to `volume-locality,fewest-build-containers`.

atc/engine/step_factory.go
evanchaoli: Yes. But as I explained in the other comment, `volume-locality` doesn't work for `get`.

atc/engine/step_factory.go
```diff
@@ -130,7 +133,7 @@ func (factory *coreStepFactory) CheckStep(
 		stepMetadata,
 		factory.resourceConfigFactory,
 		containerMetadata,
-		nil,
```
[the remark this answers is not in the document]
evanchaoli: I believe `nil` will works no strategy, and ends up a random worker chosen.

clarafu:
Hmm I'm not sure if this will work well because the fewest build container strategy does not take into account check containers so when you have a worker that has less build containers than others, all the checks will be scheduled (since checks are scheduled in batches at the same time) onto that worker and that worker will end up having x number of check containers. When we first introduced the strategy, we actually had check containers scheduled using the fewest-build-containers strategy but we saw workers getting overloaded by check containers https://github.com/concourse/concourse/issues/3251

evanchaoli:
@clarafu Thanks for pointing out the check strategy problem and #3251, I wasn't aware of that check container is not counted for `fewest-build-containers`. But I think at the time of #3251, there was no nested checks yet. Nested checks have `build_id`, so they should be able to apply `fewest-build-containers`.

>  If we have it be fewest build containers and maybe a bunch of builds kick off at once, they could all choose the same worker for the get and then all the get, put and tasks would then be placed all on those workers.

I think this point is also valid in theory. But only running in a production environment may really prove that.

I also think current fewest container strategy is not good enough. Because container count of workers are not really real-time, which leads to the situation as you described: if a bunch of steps start at once, then they may placed on the same worker. 

We may enhance `fewest-build-containers` in this way:
* Sort candidate workers by container count as the current implementation
* Divide sorted workers into 2 groups
* Randomize the first group of workers
* Choose the first worker

This way `fewest-build-container` strategy will randomly choose a worker from relative idle workers, which prevents a lot of containers from being placed on the same worker, and which mitigate inaccuracy of non-real-time worker container counts.

>  I can think of is lowering the interval for the checking interval for resource types?

This can be considered in #8043. But without #8043, nested resource type doesn't depend on check-interval, right? We only ensure a resource type is checked only once in a build.

But we can also increase resource check interval in a large deployment. We have increase default resource check interval to 2 minutes in our prod cluster.

In summary:

I agree to revert this PR. I will create a new PR with adding the following things:
 
* Enhance `fewest-build-contaienrs` strategy to consider first half of fewest container counts workers as candidates.
* Add a new ATC option: `no-input-container-placement-strategy`. Instead of force `get` to use `fewest-container`, make it configurable. I think only running the strategies in production environments for some days, we can really conclude which one is better. And maybe different clusters have different fits.
* Change resource  checks back to use `random` strategy, and nested checks to use `no-input-container-placement-strategy`.

PR #8149, opened 2022-03-08
Fix topgun resource types test failure

xtremerui:
reverting previous changes https://github.com/concourse/concourse/pull/7707 as it turned out to be a fix on a flake that happened to be worked for a while. Now with 7.7 performance improvement on lidar, the flake seems gone. Thus the topgun test started to fail (in a good way).

Plus the resource will not do lidar check anymore it should reduce the chance of flake to minimum.


clarafu (approved):

xtremerui:
@clarafu the test is still failing even by setting the resource to never check. 

https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/bosh-topgun-both/builds/1093#L622a4254:2761:2771

Issue #8167, opened 2022-03-14
All pipelines set with `set_pipeline` step archived after upgrade to 7.7.0


mymasse:
## Summary

We have setup our Concourse to follow the gitops for setting pipelines (https://blog.concourse-ci.org/gitops-for-pipelines/) and have a pipeline which makes use of the `set_pipeline` step to manage all our pipelines. After the upgrade from 7.6.0 to 7.7.0 at the first run of the master pipeline all child pipelines were archived.

## Steps to reproduce

- Follow blog mentioned above to create self updating pipelines with Concourse 7.6.0.
- Upgrade to 7.7.0
- Run the self-updating pipeline

## Expected results

Pipelines are updated (if required)

## Actual results

All pipelines were archived

## Additional context

Wondering if related to #8102 

## Triaging info

* Concourse version: 7.7.0
* Browser (if applicable):
* Did this used to work? Yes

mymasse:
We downgraded our concourse back to 7.6.0 and re-ran our master pipeline and the all came back.

evanchaoli:
@mymasse Can you describe more details on how that happened? Did your first run succeed or fail? Why do you consider the problem is caused by #8137?

xtremerui:
pinning @clarafu just in case.

clarafu:
@mymasse Do you know if the parent job that is responsible for all the `set_pipeline` steps for the child pipeline had a failed/errored build as the latest build when you saw all the child pipelines get archived?

clarafu:
Oh btw https://github.com/concourse/concourse/pull/8137 never made it into 7.7, so I think that pr is supposed to fix what you saw happen here.

gid0:
We're also impacted by this issue, self-updating pipelines were archived the first time they ran after the 7.6.0 to 7.7.0 upgrade.
The pipelines use set_pipeline: self.

gid0:
Re. the fix here: https://github.com/concourse/concourse/commit/79cbc87931e08ab00f17e7fa3daa4f78b8406866
And its description:

> If the last build of the parent pipeline was a failed/errored build that
> did not get to the set_pipeline step for the child pipelines...

[...]

PR #8179, opened 2022-03-18
Automatic pipeline archiver archiving child pipelines when parent job has a failed/errored build

clarafu:

related #8167

If the last build of the parent pipeline was a failed/errored build that did not get to the `set_pipeline` step for the child pipelines, that previously would have caused the pipeline archiver to archive the child pipeline because the `parent_build_id` would be less than the `latest_completed_build_id` for the job since failed/errored builds count
towards the `latest_completed_build_id`. So I changed the query to also only count if the `latest_completed_build_id` is a successful build.

Also found another bug where the pipeline archiver was archiving pipelines that have already been archived. This would result in a lot of unnecessary queries being run by the automatic archiver because it is updating the rows for pipelines that are already archived.

* [x] done
* [ ] todo

## Release Note

* Also fixes a bug where the pipeline archiver re-archives pipelines that are already archived

### DIFF
--- a/atc/db/pipeline_lifecycle.go
+++ b/atc/db/pipeline_lifecycle.go
@@ -40,6 +40,8 @@ func (pl *pipelineLifecycle) ArchiveAbandonedPipelines() error {
 		Where(sq.And{
 			// pipeline was set by some build
 			sq.NotEq{"p.parent_job_id": nil},
+			// pipeline is not already archived
+			sq.Eq{"p.archived": false},
 			sq.Or{
 				// job (that set child pipeline) from parent pipeline is
 				// removed, Concourse marks job as inactive
@@ -51,9 +53,20 @@ func (pl *pipelineLifecycle) ArchiveAbandonedPipelines() error {
 				// build that set the pipeline is not the most recent for the job.
 				// parent_build_id can be later than latest_completed_build_id if this
 				// gc query runs during a run of a build, specifically between the time
-				// of a completed set pipeline step and the build finishing
-				sq.Expr("p.parent_build_id < j.latest_completed_build_id"),
-			}}).
+				// of a completed set pipeline step and the build finishing. Also only
+				// take into account successful builds in order to know if this child
+				// pipeline had its set_pipeline step removed from parent job.
+				sq.And{
+					sq.Expr("p.parent_build_id < j.latest_completed_build_id"),
+					sq.Expr(`EXISTS (
+						SELECT 1
+						FROM builds lb
+						WHERE lb.id = j.latest_completed_build_id
+						AND lb.status = ?
+					)`, BuildStatusSucceeded),
+				},
+			},
+		}).
 		RunWith(tx).
 		Query()
 	if err != nil {
--- a/atc/db/pipeline_lifecycle_test.go
+++ b/atc/db/pipeline_lifecycle_test.go
@@ -2,6 +2,7 @@ package db_test
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db"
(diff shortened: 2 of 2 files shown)


evanchaoli:
I think this is a correct fix but I still want to understand why #8167 happened.

atc/db/pipeline_lifecycle.go
```diff
@@ -40,20 +40,31 @@ func (pl *pipelineLifecycle) ArchiveAbandonedPipelines() error {
...
-				// build that set the pipeline is not the most recent for the job.
-				// parent_build_id can be later than latest_completed_build_id if this
-				// gc query runs during a run of a build, specifically between the time
-				// of a completed set pipeline step and the build finishing
-				sq.Expr("p.parent_build_id < j.latest_completed_build_id"),
-			}}).
+				// // build that set the pipeline is not the most recent for the job.
+				// // parent_build_id can be later than latest_completed_build_id if this
+				// // gc query runs during a run of a build, specifically between the time
+				// // of a completed set pipeline step and the build finishing
```
wuhuizuo, line 56: should revert the changed lines?

clarafu: Hmm I don't think the changed lines here introduced the bug 🤔  I feel like the bug was just caused by the thing this pr is fixing but I could be wrong and missing something?

wuhuizuo: I mean the double commented lines. @clarafu

clarafu: Ohhhh haha yes good catch 😆

atc/db/pipeline_lifecycle.go
```diff
@@ -51,9 +53,20 @@ func (pl *pipelineLifecycle) ArchiveAbandonedPipelines() error {
...
 				// gc query runs during a run of a build, specifically between the time
-				// of a completed set pipeline step and the build finishing
-				sq.Expr("p.parent_build_id < j.latest_completed_build_id"),
-			}}).
+				// of a completed set pipeline step and the build finishing. Also only
+				// take into account successful builds in order to know if this child
+				// pipeline had its set_pipeline step removed from parent job.
+				sq.And{
+					sq.Expr("p.parent_build_id < j.latest_completed_build_id"),
+					sq.Expr(`EXISTS (
```
evanchaoli, line 61: I still don't think this is good enough. For example, `set_pipeline` is wrapped under `on_failure`, then after a success run, child pipeline will be archived.

`p.parent_build_id < j.latest_completed_build_id` is used to handle a case where a job originally had `set_pipeline` then removed `set_pipeline` from the job, then all child pipelines set by this job would be archived. As we have seen that, this logic is hard to implement, it relates pipeline logic, there are a lot of cases.

I still cannot understand why many users reported auto archive bug after upgrade to 7.7. You changed "!=" to "<", the condition actually became to less aggressive.

So I think maybe we just remove `p.parent_build_id < j.latest_completed_build_id`. Only auto archive child pipeline when parent destroyed or archived.

clarafu: I agree this is a hard case to implement, there's many edge cases to it that are hard to completely flush out. I do think that this is an important feature to get down though because we want to be able to support child pipelines getting set and archived when it is no longer a step within the parent job. Our use case for this is our pr pipelines within our prod environment. https://ci.concourse-ci.org/teams/main/pipelines/reconfigure-pipelines/jobs/reconfigure-prs/builds/792

The case that you mentioned with `on_failure` is a very good point, I agree that will break this behaviour. I'm not sure I can think of a reason why someone would want to do a `set_pipeline` during an `on_failure` though because the purpose of `on_failure`, `on_error`, etc is for notifications. But we do not restrict a user from using it in this way. Maybe there is a better way to achieve the result we want without using this current method of comparing `parent_build_id` and `latest_completed_build_id`, I'll think about it some more because we do want the end result feature.

For the auto archive bug after upgrade, the PR https://github.com/concourse/concourse/pull/8137 never made it into 7.7.0.

clarafu: I merged this pr so that I can include it in the 7.7.1 patch release but I created an issue here https://github.com/concourse/concourse/issues/8201 for the concerns we have mentioned in this thread and for rethinking the logic for auto archiving. Once we get that fixed, we will get it into 7.8.0.

PR #8183, opened 2022-03-21
Add no-input-strategy and let get/check use it.

evanchaoli:

Currently `check` step use random container placement strategy, which is because of issue #3251. But at the time of #3251, there was not step nested check yet. If a nested check happens to be placed on a busy worker, then the build would be stuck at the nested `check` step.

The other thing is, `get` and `check` step has no input, thus `volume-locality` doesn't fit them at all, but `volume-locality` is good for `put` and `task` steps as they need input volumes, `volume-locality` helps avoid volume streaming, thus reduce build time.

This PR adds a new ATC CLI opiton: `--no-input-container-placement-strategy` which allows to specify a second placement strategy, and the strategy will be used for `get` and nested `check` steps. Resource checks will be still using random strategy.

* [x] done

## Notes to reviewer

This is a resubmit of #8142 as I consider the change is important for nested `checks`.

@clarafu To address your concern of https://github.com/concourse/concourse/pull/8142#issuecomment-1059486155:

1. This PR doesn't force `get`, `check` to use `fewest-build-containers`, instead, it's configurable.
2. Resource check will be still using random strategy.
3. For `get` step, this PR make `get` step's strategy configurable, so that we can try in real deployment to see which strategy is better. Also I'm thinking for `get` step, a new strategy `fewer-build-containers` (worker having containers less than mean are all candidate) might be better, I may create the new strategy in a separate PR.

## Release Note

Enhanced container placement strategy:

* Add optional flag `--no-input-container-placement-strategy` for configuring a container placement strategy used for only `get` and **nested** `check` steps.

### DIFF
--- a/atc/engine/step_factory.go
+++ b/atc/engine/step_factory.go
@@ -24,6 +24,7 @@ type coreStepFactory struct {
 	resourceConfigFactory db.ResourceConfigFactory
 	defaultLimits         atc.ContainerLimits
 	strategy              worker.PlacementStrategy
+	noInputStrategy       worker.PlacementStrategy
 	defaultCheckTimeout   time.Duration
 	defaultGetTimeout     time.Duration
 	defaultPutTimeout     time.Duration
@@ -40,6 +41,7 @@ func NewCoreStepFactory(
 	resourceConfigFactory db.ResourceConfigFactory,
 	defaultLimits atc.ContainerLimits,
 	strategy worker.PlacementStrategy,
+	noInputStrategy worker.PlacementStrategy,
 	defaultCheckTimeout time.Duration,
 	defaultGetTimeout time.Duration,
 	defaultPutTimeout time.Duration,
@@ -55,6 +57,7 @@ func NewCoreStepFactory(
 		resourceConfigFactory: resourceConfigFactory,
 		defaultLimits:         defaultLimits,
 		strategy:              strategy,
+		noInputStrategy:       noInputStrategy,
 		defaultCheckTimeout:   defaultCheckTimeout,
 		defaultGetTimeout:     defaultGetTimeout,
 		defaultPutTimeout:     defaultPutTimeout,
@@ -77,7 +80,7 @@ func (factory *coreStepFactory) GetStep(
 		containerMetadata,
 		factory.lockFactory,
 		factory.resourceCacheFactory,
-		factory.strategy,
+		factory.noInputStrategy,
 		delegateFactory,
 		factory.pool,
 		factory.defaultGetTimeout,
@@ -130,7 +133,7 @@ func (factory *coreStepFactory) CheckStep(
 		stepMetadata,
 		factory.resourceConfigFactory,
 		containerMetadata,
-		nil,
+		factory.noInputStrategy,
 		factory.pool,
 		delegateFactory,
 		factory.defaultCheckTimeout,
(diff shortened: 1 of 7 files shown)


atc/exec/check_step.go
```diff
@@ -273,7 +273,12 @@ func (step *CheckStep) runCheck(
 		return nil, runtime.ProcessResult{}, err
 	}
 
-	worker, err := step.workerPool.FindOrSelectWorker(ctx, containerOwner, containerSpec, workerSpec, step.strategy, delegate)
+	strategy := step.strategy
+	if step.plan.IsResourceCheck() {
+		// Resource check containers should be placed randomly. Refer to PR#3251.
```
clarafu, line 278:
```suggestion
		// Resource check containers should be placed randomly. Refer to issue #3251.
```

atc/engine/step_factory.go
```diff
@@ -77,7 +80,7 @@ func (factory *coreStepFactory) GetStep(
 		containerMetadata,
 		factory.lockFactory,
 		factory.resourceCacheFactory,
-		factory.strategy,
+		factory.noInputStrategy,
```
clarafu, line 83: Would we be able to make it so that if you don't pass in a `no-input-strategy` it will continue to use the `factory.strategy`? This way the addition of this new flag won't affect users that don't configure it.

atc/worker/placement.go
```diff
@@ -14,6 +14,7 @@ import (
 
 type PlacementOptions struct {
 	Strategies                   []string `long:"container-placement-strategy" default:"volume-locality" choice:"volume-locality" choice:"random" choice:"fewest-build-containers" choice:"limit-active-tasks" choice:"limit-active-containers" choice:"limit-active-volumes" description:"Method by which a worker is selected during container placement. If multiple methods are specified, they will be applied in order. Random strategy should only be used alone."`
+	NoInputStrategies            []string `long:"no-input-container-placement-strategy" default:"fewest-build-containers" choice:"volume-locality" choice:"random" choice:"fewest-build-containers" choice:"limit-active-tasks" choice:"limit-active-containers" choice:"limit-active-volumes" description:"Method by which a worker is selected during container placement. If multiple methods are specified, they will be applied in order. Random strategy should only be used alone."`
```
clarafu, line 17:
```suggestion
	NoInputStrategies            []string `long:"no-input-container-placement-strategy" default:"fewest-build-containers" choice:"volume-locality" choice:"random" choice:"fewest-build-containers" choice:"limit-active-tasks" choice:"limit-active-containers" choice:"limit-active-volumes" description:"A second container placement strategy that will only be used for get and nested check steps."`
```

With my suggestion above having it so that if we don't pass anything in to this flag, it will just use the original strategy we will need to remove the default here then so that it can be a nil value if it is not passed in.

atc/engine/step_factory.go
evanchaoli: Changed noInputStrategy to default to strategy.

clarafu (approved):
It looks great! I think we just need to add the new flag to the concourse-chart and concourse-bosh-release and it should be all good to go!

evanchaoli:
> I think we just need to add the new flag to the concourse-chart and concourse-bosh-release and it should be all good to go!

Here comes them: 

* https://github.com/concourse/concourse-chart/pull/299
* https://github.com/concourse/concourse-bosh-release/pull/166

PR #8184, opened 2022-03-21
optimize worker selection

evanchaoli:

https://github.com/concourse/concourse/blob/4647d76409117172e4e265b068d5e9bcb58db5c8/atc/worker/pool.go#L111-L141

This function is not implement efficiently. It finds out worker containing the container, then list all compatible workers, then uses a double loop to decide if a found worker should be returned.

This PR optimize the function. The change should be easy to understand. Though the optimization is small, but given all resource checks suppose to reuse containers, this optimization will get rid of listing workers for most of resource check steps. Given resource check builds are the most significant part of total workloads, this optimization should improve big. 

* [x] done

## Release Note

* Optimized worker selection when `global-resources` is enabled

### DIFF
--- a/atc/worker/pool.go
+++ b/atc/worker/pool.go
@@ -231,17 +231,15 @@ func (pool Pool) findWorkerForContainer(logger lager.Logger, owner db.ContainerO
 		return nil, nil, false, err
 	}
 
+	for _, w := range workersWithContainer {
+		if pool.isWorkerCompatibleAndRunning(logger, w, workerSpec) {
+			return w, nil, true, nil
+		}
+	}
+
 	compatibleWorkers, err := pool.allCompatibleAndRunningWorkers(logger, workerSpec)
 	if err != nil {
 		return nil, nil, false, err
-	}
-
-	for _, w := range workersWithContainer {
-		for _, c := range compatibleWorkers {
-			if w.Name() == c.Name() {
-				return w, compatibleWorkers, true, nil
-			}
-		}
 	}
 
 	return nil, compatibleWorkers, false, nil
--- a/atc/worker/pool.go
+++ b/atc/worker/pool.go
@@ -232,8 +232,9 @@ func (pool Pool) findWorkerForContainer(logger lager.Logger, owner db.ContainerO
 		return nil, nil, false, err
 	}
 
-	// When global-resources is enabled, a check-container may run on any team
-	// worker, which enables this optimization. Details refer to PR#8184.
+	// When global-resources is enabled, a check-container may run on any worker
+	// as long as the scope is the same, which enables this optimization. Details
+	// refer to PR#8184.
 	if atc.EnableGlobalResources {
 		for _, w := range workersWithContainer {
 			if pool.isWorkerCompatibleAndRunning(logger, w, workerSpec) {
--- a/atc/worker/pool.go
+++ b/atc/worker/pool.go
@@ -10,6 +10,7 @@ import (
 
 	"code.cloudfoundry.org/lager"
 	"code.cloudfoundry.org/lager/lagerctx"
+	"github.com/concourse/concourse/atc"
 	"github.com/concourse/concourse/atc/db"
 	"github.com/concourse/concourse/atc/metric"
 	"github.com/concourse/concourse/atc/runtime"
@@ -231,15 +232,29 @@ func (pool Pool) findWorkerForContainer(logger lager.Logger, owner db.ContainerO
 		return nil, nil, false, err
 	}
 
-	for _, w := range workersWithContainer {
-		if pool.isWorkerCompatibleAndRunning(logger, w, workerSpec) {
-			return w, nil, true, nil
+	// When global-resources is enabled, a check-container may run on any team
+	// worker, which enables this optimization. Details refer to PR#8184.
+	if atc.EnableGlobalResources {
+		for _, w := range workersWithContainer {
+			if pool.isWorkerCompatibleAndRunning(logger, w, workerSpec) {
+				return w, nil, true, nil
+			}
 		}
 	}
 
 	compatibleWorkers, err := pool.allCompatibleAndRunningWorkers(logger, workerSpec)
 	if err != nil {
 		return nil, nil, false, err
+	}
+
+	if !atc.EnableGlobalResources {
+		for _, w := range workersWithContainer {
+			for _, c := range compatibleWorkers {
+				if w.Name() == c.Name() {
+					return w, compatibleWorkers, true, nil
+				}
+			}
+		}
 	}
 
 	return nil, compatibleWorkers, false, nil


atc/worker/pool.go
```diff
@@ -231,19 +231,17 @@ func (pool Pool) findWorkerForContainer(logger lager.Logger, owner db.ContainerO
 		return nil, nil, false, err
 	}
 
+	for _, w := range workersWithContainer {
```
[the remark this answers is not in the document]
evanchaoli: @clarafu I don't think this is a problem. In theory, it should never find more than 1 container, and when the found container was created, it should have checked team comparability.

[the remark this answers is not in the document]
clarafu: I think it becomes a problem for users that don't have global-resources enabled. Since they won't be sharing the same version history, it does matter where the check is run because they will all have individual scopes and will need to be checked individually. We also documented the feature that if they use team workers, all work/build will be scheduled onto the team workers.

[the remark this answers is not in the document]
evanchaoli: I believe Concourse has goal of enabling global-resources by default, and I don't see a reason of not enabling global-resources, thus I believe disabling global-resources is minority. So, if this is the only concern, we can have this optimization only when global-resources is enabled.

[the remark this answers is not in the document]
clarafu: I really want it to be on by default 😢  But we still have the problem of https://github.com/concourse/concourse/issues/3023 that we need to fix before we can turn it on by default

atc/worker/pool.go
```diff
@@ -231,15 +232,27 @@ func (pool Pool) findWorkerForContainer(logger lager.Logger, owner db.ContainerO
 		return nil, nil, false, err
 	}
 
+	// When global-resources is enabled, a check-container may run on any team
+	// worker, which enables this optimization. Details refer to PR#8184.
+	if atc.EnableGlobalResources {
+		for _, w := range workersWithContainer {
+			if pool.isWorkerCompatibleAndRunning(logger, w, workerSpec) {
```
clarafu, line 239: Is there a way we can have it choose a team worker if it is available? And there isn't one, then we can use any worker?

I know some users use team workers to split up the workloads and it would be best if we can keep the same behaviour as before where it will prefer to use team workers if available and if not then it will use general workers.

atc/worker/pool.go
```diff
@@ -231,15 +232,27 @@ func (pool Pool) findWorkerForContainer(logger lager.Logger, owner db.ContainerO
 		return nil, nil, false, err
 	}
 
+	// When global-resources is enabled, a check-container may run on any team
```
clarafu, line 235: I think the check container will only run on either its team worker or a general worker but it will never run on another team's worker.

evanchaoli: I might not state clearly. What I really meant was that: say resource-1 of team-1, resource-2 of team-2 and resource-3 of team-3 has the same scope, and the 3 teams all have team workers. Then the 3 checks will run on all 3 team workers with the existing logic, which is actually a waste, as they share the same version history and acquiring the same lock. With this PR, the check may run on any team worker.

atc/worker/pool.go
evanchaoli: With this PR, it should never find more than 1 worker with target container. Do you agree?

With the example I used in the other comment. If resource-1 happens to run before the other two, then the check container will be placed on team-1's worker. When resource-2 runs, it will still reuse the check container from team-1's worker. If the team-1's worker is stalled, and resource-3 gets to run, then the check container will be recreated on team-3's worker. It should never find two containers.

evanchaoli:
@clarafu any more comments?

atc/worker/pool.go
clarafu: Right, I think this PR would make it choose the worker that has the target container on it, even if it is not from their team's worker. But I think before this PR, we would only choose the container if it was on one of the team workers. I'm just a little worried that the users will now see that they are running workloads on general workers rather than their own team workers. For example, previously when there is a team with team workers users would see that all their checks are run only on their team workers but after this pr, they might see their checks run on general workers. I do think global resources blurs this line with sharing of versions/check results, but this pr will take it a step further and actually run a team's resource checks on a non-team worker. We do define that team workers will take all workloads from that team in our docs https://concourse-ci.org/concourse-worker.html#team-workers so if we change this then we might want to add a note in there for users with team workers and global-resources.

atc/worker/pool.go
clarafu: Hmm are you sure about that? I think if there are 3 resource check containers on 3 different team workers, it won't end up using a team worker from a different team. For example, team-1 will never use the check container for the same scope on team-2 worker or team-3 worker. This is because we have a check in the `isWorkerCompatibleAndRunning` method that will check to see if the worker's `TeamID` is the same as the `TeamID` in the spec.

https://github.com/concourse/concourse/blob/72cb3661374529b6fb4ea80d98654f804e13a78d/atc/worker/pool.go#L421-L425

evanchaoli: @clarafu What you described is true with the existing logic, and that's the waste I mentioned. As the 3 resources share the same version history, it should not be a matter on which team's worker the check container runs, right? With this PR, the check container may run on any team's worker, there is only a single check container for the 3 resources, which saves 2 containers comparing to the existing logic.

atc/worker/pool.go
evanchaoli: > I'm just a little worried that the users will now see that they are running workloads on general workers rather than their own team workers.

@clarafu I don't think that should happen. This PR doesn't modify the logic of initial check container creation, so if there is a team worker, a check container will be created on the team container.

What this PR changes is, a check container now may locate on other team's worker in case of the same scope. We can just enhance the doc to explain that.

clarafu: Are you sure? I don't think I was describing the initial check container creation, I think my example was explaining if there is a check container that exists for the scope on a general worker, before this change it would not use that check container and create a new container on a team worker. But after this change it will just use the check container on the general worker even if the team has a team worker. I'm just unsure about this fact because we give a guarantee that if you have team workers, all workloads will be scheduled onto those workers and we would be breaking that with this change.

atc/worker/pool.go
clarafu: I do think it still matters which team worker the check container runs on because we make a guarantee that if you have a team worker, all workloads will be scheduled on it. Users might think its a bug when they see their checks selecting other team's workers.

atc/worker/pool.go
evanchaoli: @clarafu I understood what you described, I just don't consider that's a problem. Let's use an example:

* teamA has resourceA and teamB has resourceB, and the two resources belong to the same scope
* both teams have no team worker
* a check container will be created on a general worker
* after some time, teamB got a team worker

Then our difference comes:

* My opinion is that, resourceB should still reuse the original check container
* Your opinion is that, resourceB should create a new check container on teamB's worker

Does that state clearly our difference?

As resourceB will still acquire the same resource-check-lock, share the same version history, and test the same scrope last check time as resourceA, what sense does it make to create a second check container on teamB's worker? If resourceA's check is always run a little bit earlier than resourceB, then due to the scope's last check time, resourceB might never get a chance to run check container.

So I think we should update the doc to explain that, when global resource is enabled, a check container might not strictly follow the team worker rule.

atc/worker/pool.go
evanchaoli: Same response as the last comment.

atc/worker/pool.go
xtremerui: According to what you describe above, i think the comment should be `a check-container may run on any worker` as long as the scope is same?

xtremerui (approved):
Given there is a way to opt-out (disable Global resources) and the check container will only be reused if their scopes are the same, I think it is safe to have this optimization. Plus in Concourse doc the global resources are still marked as experimental feature.

We will need to update the doc here https://concourse-ci.org/global-resources.html#complications-with-reusing-containers

EDIT: in concourse-bosh-release global resources flag is by default false while in concourse-chart it is by default true. Its better to make it consistent.

evanchaoli:
@xtremerui Thanks for pointing out the doc, I created a PR to update the doc: https://github.com/concourse/docs/pull/499.

PR #8196, opened 2022-03-24
Add fly clear-versions command

### LINKED ISSUE #5418: Clear resource versions fly command

I've had a few times where users have come to me with a problem about either their database being slow because of a query that has to look through their large `resource_config_versions` table and is causing slowness or their version history is wrong because of some bug with the resource type producing incorrect versions.

In those situations, I would ask them to clear out some of their resource version history manually through the database. This step would be a lot easier if there was a fly command that would clear out the version history.

This command should only be able to be used by an admin. This is because multiple teams can be using the same resource version history so clearing out one history could result in many teams having their version history wiped.

There are a few things to take into consideration when implementing this. For example, what happens if there is a resource that has pinned a version (or disabled) and you try to clear out the version history? Or should it clear out the entire history or leave the latest version remaining?

clarafu:

closes #5418

I added a fly command that will clear the version history for a resource or a resource type. It will print out a warning message before it executes the deletion in order to show the user which other resources/resource-types are affected if they are sharing the same version history.

Only admin users will be able to use this command because of how it can be used to delete version histories of resources/resource-types across pipelines/teams. 

Through adding this fly command, I had to add 4 API endpoints. 2 are for listing out the resources/resource-types that share the same `resource_config_scope_id` (meaning that they are sharing the same version history). The other 2 endpoints are for clearing the version history for the resource or resource type.

* [x] done
* [ ] todo

## Release Note

* Can be used to clear version history of a resource or resource type
* Can only be used by an admin user
* If you have `global-resources` enabled, it can possibly delete version histories of other resources/resource-types in other pipelines so there is a warning message that will show any resources or resource types that are affected.


go-concourse/concourse/team.go
```diff
@@ -51,8 +51,12 @@ type Team interface {
 
 	Resource(pipelineRef atc.PipelineRef, resourceName string) (atc.Resource, bool, error)
 	ListResources(pipelineRef atc.PipelineRef) ([]atc.Resource, error)
+	ListSharedForResource(pipelineRef atc.PipelineRef, resourceName string) (atc.ResourcesAndTypes, bool, error)
+	ListSharedForResourceType(pipelineRef atc.PipelineRef, resourceTypeName string) (atc.ResourcesAndTypes, bool, error)
 	ResourceTypes(pipelineRef atc.PipelineRef) (atc.ResourceTypes, bool, error)
 	ResourceVersions(pipelineRef atc.PipelineRef, resourceName string, page Page, filter atc.Version) ([]atc.ResourceVersion, Pagination, bool, error)
+	ClearResourceVersions(pipelineRef atc.PipelineRef, resourceName string) (int64, error)
+	ClearResourceTypeVersions(pipelineRef atc.PipelineRef, resourceName string) (int64, error)
```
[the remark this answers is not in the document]
clarafu: Great catch! I made the edits :D

atc/api/resourceserver/list_shared_for_resource.go
```diff
@@ -0,0 +1,45 @@
...
+			w.WriteHeader(http.StatusInternalServerError)
+			return
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		err = json.NewEncoder(w).Encode(shared)
+		if err != nil {
+			logger.Error("failed-to-encode-shared-resources-and-types", err)
+			w.WriteHeader(http.StatusInternalServerError)
```
xtremerui, line 42: somewhere it returns error in the response body so fly cmd will show the error together with status code, which I think might be helpful. Have you considered to do the same for these new endpoints?

https://github.com/concourse/concourse/blob/c0e0782bfac8f345820c9a1631631fa03129e322/atc/api/resourceserver/check.go#L64

clarafu: Thats a good idea, let me add that!

clarafu:
@xtremerui Should be ready to be re-reviewed! :)

atc/api/resourceserver/list_shared_for_resource.go
xtremerui: @clarafu are you going to add the error mesage to the response still?

xtremerui (approved):
Manually run the concourse and build fly to test the new command. Works good for me!

PR #8198, opened 2022-03-25
Enhance locks

evanchaoli:

Concourse highly relies on db locks to coordinate works across ATCs. But all locks shares a single db connection and a single mutex.

For example, when GC component is acquiring a BatchLock, the other go-routing that is trying to acquire a VolumeCreating lock has to wait the global mutex.

The global mutex is a performance bottleneck on a busy deployment. For example, on our prod cluster, step initialization were always slow, and I noticed that the more resource checking locks hold, then less volume creating locks could be acquired. Thus I created a private build that give a volume-create lock a separate mutex. The private build has been significantly reduced step initialization on our prod cluster.

This PR will give certain lock types dedicate mutex, so that one lock type won't impact others.

During my load tests, I also found that, after giving each lock type a dedicate mutex, locks could be quickly acquired, but one ATC could get extremely busy while other ATCs were idle. That was because some component might be always run on a single ATC. For example, if Lidar scanner is running on ATC1, then ATC1 needs to acquire a lot of resource checking locks, then ATC1 has less chance to acquire `BatchLock("lidar")` again because of the global mutex, so that other ATC will be more likely to run Lidar scanner in next round.

So we need to enhancement component as well to ensure each ATC to have the same chance to run a component.

* [x] Enhance locks

## Release Note

* 

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -279,13 +279,17 @@ type Migration struct {
 }
 
 func (m *Migration) Execute(args []string) error {
-	lockConn, err := constructLockConn(defaultDriverName, m.Postgres.ConnectionString())
+	lockConns, err := constructLockConns(defaultDriverName, m.Postgres.ConnectionString())
 	if err != nil {
 		return err
 	}
-	defer lockConn.Close()
+	defer func() {
+		for _, conn := range lockConns {
+			conn.Close()
+		}
+	}()
 
-	m.lockFactory = lock.NewLockFactory(lockConn, metric.LogLockAcquired, metric.LogLockReleased)
+	m.lockFactory = lock.NewLockFactory(lockConns, metric.LogLockAcquired, metric.LogLockReleased)
 
 	if m.MigrateToLatestVersion {
 		return m.migrateToLatestVersion()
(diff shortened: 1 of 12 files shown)


clarafu:
Hmm what kind of component enhancement were you thinking about adding? Because if this does cause one ATC to be overloaded then we might need to think about balancing the workloads out before getting this lock change merged

evanchaoli:
> Hmm what kind of component enhancement were you thinking about adding? Because if this does cause one ATC to be overloaded then we might need to think about balancing the workloads out before getting this lock change merged

@clarafu  First of all, this PR is very important. You know that our prod was unstable in Feb. We tried many ways to scale out: tuned container placement strategy, added 1 more ATC as well as some more workers, added more cpu and memory to the host where the db ran on. However the instability still lasted for a few months until I applied a private patch that gave VolumeCreationLock a separate mutex.

I found one ATC overloaded initially with a change that gave each lock type a separate mutex. But in this PR, I only give some lock types dedicate mutexes. I have been running this PR in a load test for a few months, it works well. Workloads can be distributed across ATC.

atc/db/lock/lock.go
```diff
@@ -24,8 +24,39 @@ const (
...
+	factoryInMemoryCheckBuildTracking
+	factoryBuildTracking
+	factoryJobScheduling
+
+	// FactoryCount must be the last item
+	FactoryCount
+)
+
+func mapLockTypeToFactory(lockType int) int {
+	switch lockType {
```
xtremerui, line 44: Based on what it decides one lock type to have its own factory e.g. locks count, lock held duration ...?

atc/atccmd/command.go
```diff
@@ -1646,17 +1652,21 @@ type Closer interface {
...
 }
 
-func constructLockConn(driverName, connectionString string) (*sql.DB, error) {
-	dbConn, err := sql.Open(driverName, connectionString)
-	if err != nil {
-		return nil, err
-	}
+func constructLockConns(driverName, connectionString string) ([lock.FactoryCount]*sql.DB, error) {
+	conns := [lock.FactoryCount]*sql.DB{}
+	for i := 0; i < lock.FactoryCount; i++ {
```
xtremerui, line 1657: Where is `FactoryCount` got initialized? What is an example value?

atc/db/lock/lock.go
evanchaoli: Based on metrics. I just give those locks that are acquired big amount of times dedicate mutexes.

atc/db/lock/lock.go
```diff
@@ -81,23 +116,29 @@ type lockFactory struct {
 type LogFunc func(logger lager.Logger, id LockID)
 
 func NewLockFactory(
-	conn *sql.DB,
+	conn [FactoryCount]*sql.DB,
```
xtremerui, line 119:
```suggestion
	conns [FactoryCount]*sql.DB,
```

xtremerui (approved):

PR #8200, opened 2022-03-25
Fix automatic pipeline archive bug

clarafu:

~~Port over the code from https://github.com/concourse/concourse/pull/8137 and https://github.com/concourse/concourse/pull/8179 that help fix a few bugs found in the automatic pipeline archiver logic.~~

Actually I'm just going to remove the line that introduced the bugs for now and have more time to think about how we want to achieve the behaviour we want. The `sq.Expr("p.parent_build_id != j.latest_completed_build_id")` is not super necessary because we already have a query when a build successfully finishes that will help gc child pipelines that are no longer set.

https://github.com/concourse/concourse/blob/4f3bb8fb4916cf8ff8cc918ed81580f84e418420/atc/db/build.go#L751-L762

It is not perfect because it relies on a build finishing successfully to archive the child pipelines, which is what the line that introduced the bug was trying to improve. But since the line has brought up so much risk, let's take a step back and think about it and for now we can remove it from the 7.7.x release.

## Release Note

* Removes the line that introduced a bug in 7.7.0 where child pipelines can get archived accidentally. 
* Fixes a bug where pipelines that are already archived get re-archived

### DIFF
--- a/atc/db/pipeline_lifecycle.go
+++ b/atc/db/pipeline_lifecycle.go
@@ -48,8 +48,11 @@ func (pl *pipelineLifecycle) ArchiveAbandonedPipelines() error {
 				sq.Eq{"j.id": nil},
 				// parent pipeline was archived
 				sq.Eq{"parent.archived": true},
-				// build that set the pipeline is not the most recent for the job
-				sq.Expr("p.parent_build_id != j.latest_completed_build_id"),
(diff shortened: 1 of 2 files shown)


vito (approved):
👻

gid0:
❤️

divyamudundi:
Thanks for the release.

PR #8243, opened 2022-04-06
Update topgun resource type test to match actual behaviour

xtremerui:
Refer to https://github.com/concourse/concourse/pull/8236

With `check_every: never` set to resource, the triggering job will first waiting for a version of `10m`, which then trigger the check on 10m first. So there will be 2 check containers for `10m`  and `my-time`. And results resource cache of `my-time` created in DB. 

Then the build will start. In the `get` step, it sees `10m`  has a custom type `my-time` ,  so it will try to check the image and thus getting a resource cache hit. This will result 1 check container.

In the next step there is the check for the task image. So in total 4 check containers.

Removing `check_every: never` from `my-time` as it will make the build hangs forever as `10m` couldn't find a version.

### DIFF
--- a/topgun/both/resource_types_test.go
+++ b/topgun/both/resource_types_test.go
@@ -22,11 +22,13 @@ var _ = Describe("A pipeline-provided resource type", func() {
 		<-buildSession.Exited
 		Expect(buildSession.ExitCode()).To(Equal(1))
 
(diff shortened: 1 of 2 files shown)


clarafu:
Waiting on https://github.com/concourse/concourse/pull/8253 before commiting to this!

Issue #7761, opened 2021-11-03
First build number in build page is shifted


clarafu:
## Summary

I noticed that the first build number in the build page looks to be much further to the right compared to the other build tabs.

![image](https://user-images.githubusercontent.com/19290703/140072610-be532760-8c55-47eb-af80-57b36e338496.png)

The reason for it being shifted is because it is so that it can be aligned with the build name text above it, but I find that the larger and shifted tab for the first build number looks a bit more awkward than being unaligned with the build name text above.

This is what it looked like before

![image](https://user-images.githubusercontent.com/19290703/140072879-0ada4918-7af3-451e-94b6-8e45840a47e5.png)

I'd just like to get some more input on this or if anyone feels differently, because they both have their advantages and disadvantages!

## Triaging info

* Concourse version:
* Browser (if applicable):
* Did this used to work?

mkocus:
In my opinion the numbers should have the same padding from left and right.

Also, the previous color style IMHO had better readability and the colors were more human, less Linux console ;)

kberzinch:
Regarding the colors - white on yellow (used at the top of running builds) does not meet WCAG contrast recommendations: https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=FAD43B

clarafu:
@kberzinch We do have this merged in https://github.com/concourse/concourse/pull/7756 where we change the header font to be black instead of white

matthewpereira:
WDYT about making the padding on build numbers larger?

![image](https://user-images.githubusercontent.com/6372810/151188498-bfc6e823-02ea-4d35-ab34-7f895d8303e4.png)
![image](https://user-images.githubusercontent.com/6372810/151188541-13fd9243-eaf2-4dd1-a1bc-0fdf1ffce32e.png)

This would solve the alignment issue (I think) and have the side benefit of giving the smaller builds a friendlier click target by increasing their surface area (build 1 used to be 20px wide, this would make it 40px wide, which is a lot closer but still not quite meeting WCAG 2.1's 44px minimum accessibility target).

clarafu:
I updated the description with the most updated images of the new vs old build page, please check it out!

lrstanley:
For me, the aborted colors in the updated screenshots for 7.7.0 are still quite hard to read. Have to really focus to read them. Though, it's also possible the screenshot quality is a bit lower than what it would actually look like.

matthewpereira:
[...]

PR #8254, opened 2022-04-08
UI build page tweaks

xtremerui:

address concerns raised in https://github.com/concourse/concourse/issues/7761

* set right padding to 16px for history build number so there is more spacing between text while maintaining the left align between latest build number and build title.
  
* add right boarder to history build number. It used to have the right boarder in older versions.
   
* set build title to white for aborted build. It should increase the readability. The contrast boost from 2.51 to 8.35, as shown in below screenshots.

Before:
Screen Shot 2022-04-07 at 9 42 04 PM

After:
Screen Shot 2022-04-07 at 9 55 58 PM

* [x] done
* [ ] todo

## Release Note

* Increase contrast on Build page title when build is aborted. Add more spacing and boarder for history build numbers.

### DIFF
--- a/web/assets/css/production.less
+++ b/web/assets/css/production.less
@@ -113,8 +113,7 @@
   line-height: 24px;
   text-align: center;
   margin: 0;
-  padding: 4px 6px 4px 4px;
-  min-width: 24px;
+  padding: 4px 16px;
   height: 24px;
   display: block;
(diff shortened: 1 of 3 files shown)


clarafu (approved):
image

It looks pretty good! Great job 😄

PR #8219, opened 2022-03-28
Bump cloudfoundry dependencies in go mod

xtremerui:
## Changes proposed by this PR
bump deps `credhub-cli` `garden` `localip` `clock` from cloudfoundry org. Most of the updates from these libraries are also deps bumping and security patching.

closes #8215

PR #8265, opened 2022-04-12
Update credhub var lookup logic

xtremerui:
After [bummping credhub-cli](https://github.com/concourse/concourse/pull/8219) to 2.9.3 it failed toggun creds test by
not being able to fall back to team secret when pipeline secret is
not found. It is due to this commit
https://github.com/cloudfoundry/credhub-cli/commit/e3951663d25cdc3a50c796937762e50eba4921ef

Changes in this PR:

 * update credhub var look up to find by name instead of find by path
 so it could avoid the following get request i.e. we dont need to do
 the get request if we can't find one secret already.
 * skip credhub no-credential-found error so it can continue
 searching through available paths.

* [x] done
* [ ] todo

## Notes to reviewer

It is hard to test this as running a standalone credhub instance is non-trivial (as it needs to run with UAA server or within a bosh env). The only place will be in topgun `bosh-topgun-core` where concourse test creds.

### DIFF
--- a/atc/creds/credhub/credhub.go
+++ b/atc/creds/credhub/credhub.go
@@ -6,8 +6,8 @@ import (
 
 	"github.com/concourse/concourse/atc/creds"
 
-	"code.cloudfoundry.org/credhub-cli/credhub"
 	"code.cloudfoundry.org/credhub-cli/credhub/credentials"
+	"code.cloudfoundry.org/credhub-cli/errors"
 	"code.cloudfoundry.org/lager"
 )
 
@@ -58,16 +58,16 @@ func (c CredHubAtc) findCred(path string) (credentials.Credential, bool, error) 
 		return cred, false, err
 	}
 
-	_, err = ch.FindByPath(path)
+	_, err = ch.FindByPartialName(path)
 	if err != nil {
+		if err.Error() == errors.NewNoMatchingCredentialsFoundError().Error() {
+			return cred, false, nil
+		}
+
 		return cred, false, err
 	}
 
 	cred, err = ch.GetLatestVersion(path)
-	if _, ok := err.(*credhub.Error); ok {
-		return cred, false, nil
-	}
-
 	if err != nil {
 		return cred, false, err
 	}
--- a/topgun/core/creds_test.go
+++ b/topgun/core/creds_test.go
@@ -58,7 +58,7 @@ func testCredentialManagement(
 			Fly.Run("unpause-pipeline", "-p", "pipeline-creds-test")
 		})
 
-		It("parameterizes via Vault and leaves the pipeline uninterpolated", func() {
+		It("parameterizes via creds manager and leaves the pipeline uninterpolated", func() {
 			By("triggering job")
 			watch := Fly.Start("trigger-job", "-w", "-j", "pipeline-creds-test/some-job")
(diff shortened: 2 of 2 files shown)


atc/creds/credhub/credhub.go
```diff
@@ -58,16 +58,16 @@ func (c CredHubAtc) findCred(path string) (credentials.Credential, bool, error)
 		return cred, false, err
 	}
 
-	_, err = ch.FindByPath(path)
+	_, err = ch.FindByPartialName(path)
```
clarafu, line 61: Are these able to be used interchangeably? Wouldn't the path be different than the name in credhub? For example, path would be `/credhub/concourse/username` versus name would be `username`?

atc/creds/credhub/credhub.go
```diff
@@ -58,16 +58,16 @@ func (c CredHubAtc) findCred(path string) (credentials.Credential, bool, error)
...
 	if err != nil {
+		if err.Error() == errors.NewNoMatchingCredentialsFoundError().Error() {
+			return cred, false, nil
+		}
+
 		return cred, false, err
 	}
 
 	cred, err = ch.GetLatestVersion(path)
-	if _, ok := err.(*credhub.Error); ok {
```
clarafu, line 67: Was there a reason this needed to be deleted?

clarafu (commented):
Sorry I don't know much about credhub/this pr so I just had a few questions about the changes made!

atc/creds/credhub/credhub.go
clarafu: It looks like the `FindByPath` is more restrictive than `FindByPartialName`? 🤔  I'm surprised that `FindByPath` did not find anything for `/concourse/main/team_secret` when there is a var in that exact path?

Was there anything that changed in the `FindByPath` method in credhub in this new bump that is making us need to use `FindByPartialName` instead now?

atc/creds/credhub/credhub.go
clarafu: Oh yess I see that, thanks!

atc/creds/credhub/credhub.go
xtremerui: > I'm surprised that `FindByPath` did not find anything for `/concourse/main/team_secret` when there is a var in that exact path?

the reason is credhub take "path" as a strictly meaning a path to the cred i.e. the parent folder of where the cred is, but not the complete uri (in Concourse semantics) to the cred.

thats also the main reason I made the switch of the method. 
The second reason is `FindByPartialName` failed earlier so we could skip the next api call if not found. 
Third reason is the linked commit from Credhub CLI actually made the error comparison harder.  `FindByParthialName` is at least returning and EXPOSING an error `NewNoMatchingCredentialsFoundError` that we can use. If not, we will have to do an awkward error comparison by comparing the error message, which would be broken if next time they change the error message.



> Was there anything that changed in the `FindByPath` method in credhub in this new bump that is making us need to use `FindByPartialName` instead now?

No, the behaviour of both method are the same. Its just us were using the `FindByPath` differently.

clarafu: Hmm I see, thanks for explaining everything!! :)

clarafu (approved):

PR #8277, opened 2022-04-15
Expose ATC_EXTERNAL_URL to task env.

evanchaoli:

Expose ATC_EXTERNAL_URL to task env.

* [x] done

## Notes to reviewer

It has been a long discussion for whether or not expose step metadata to task. Many users have requested that but Concourse team considered that is an anti-pattern.

This PR exposes only ATC_EXTERNAL_URL. Hopefully, Concourse team would not consider anti-pattern.

## Release Note

* Expose ATC_EXTERNAL_URL to task env.

### DIFF
--- a/atc/exec/step_metadata.go
+++ b/atc/exec/step_metadata.go
@@ -69,3 +69,11 @@ func (metadata StepMetadata) Env() []string {
 
 	return env
 }
+
+func (metadata StepMetadata) TaskEnv() []string {
+	env := []string{}
+	if metadata.ExternalURL != "" {
+		env = append(env, "ATC_EXTERNAL_URL="+metadata.ExternalURL)
+	}
+	return env
+}
--- a/atc/exec/task_step_test.go
+++ b/atc/exec/task_step_test.go
@@ -56,9 +56,10 @@ var _ = Describe("TaskStep", func() {
 		}
 
 		stepMetadata = exec.StepMetadata{
-			TeamID:  123,
-			BuildID: 1234,
-			JobID:   12345,
+			TeamID:      123,
+			BuildID:     1234,
+			JobID:       12345,
+			ExternalURL: "http://foo.bar",
 		}
 
 		planID = atc.PlanID("42")
@@ -174,6 +175,10 @@ var _ = Describe("TaskStep", func() {
 			chosenContainer = chosenWorker.Containers[0]
 			fakePool = new(execfakes.FakePool)
 			fakePool.FindOrSelectWorkerReturns(chosenWorker, nil)
+		})
+
+		It("Task env includes atc external url", func(){
+			Expect(chosenContainer.Spec.Env).To(ContainElement("ATC_EXTERNAL_URL=http://foo.bar"))
 		})
 
 		Context("before running the task", func() {
--- a/atc/exec/task_step_test.go
+++ b/atc/exec/task_step_test.go
@@ -178,7 +178,7 @@ var _ = Describe("TaskStep", func() {
 		})
 
 		It("Task env includes atc external url", func(){
-			Expect(chosenContainer.Spec.Env).To(ContainElement("ATC_EXTERNAL_URL=http://foo.bar"))
+			Expect(chosenContainer.Spec.Env).To(ConsistOf("ATC_EXTERNAL_URL=http://foo.bar", "SECURE=secret-task-param"))
 		})
 
 		Context("before running the task", func() {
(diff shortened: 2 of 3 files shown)


xtremerui:
What is the use case for ACT_EXTERNAL_URL here? Thx

taylorsilva:
The anti-pattern we're usually against with these env vars is the build ID and similar metadata. Stuff that auto-increments and users could potentially use as an ID themselves. We don't want users to use Concourse's internal ID's for anything outside of Concourse, and then get mad at us for when the ID's reset for some reason, like they renamed the job or the pipeline.

The external URL itself doesn't fit this anti-pattern. Users can't really use it for crafting ID's in a useful way. So maybe it's fine to insert it into tasks?

Still would be nice to know what the exact use-case is like Rui asked for.

evanchaoli:
We have a normal cluster and a secured cluster. The secured cluster is deployed inside firewall with restricted network accesses. And we have a `task`, similar to [oci-build-task](https://github.com/concourse/oci-build-task), the task depends on an internal site, say site-I. site-I has deployments in both normal network and secured network. On the secured cluster, the task needs to connect to the secured site-I. We don't want users to configure site-I's URL in pipeline, because site-I is a kind of infra resource, pipeline authors don't have to be aware of site-I's URLs. So that the task needs to know where it runs.

Currently, we use a param of the task to indicate if secured or not. Pipeline authors need set the param properly, which adds complexity of pipeline code, and it's error-prone. With this PR, the task will be able to get cluster URL, in turn to decide which site-I to connect automatically.

atc/exec/step_metadata.go
```diff
@@ -69,3 +69,11 @@ func (metadata StepMetadata) Env() []string {
 
 	return env
 }
+
+func (metadata StepMetadata) TaskEnv() []string {
+	env := []string{}
+	if metadata.ExternalURL != "" {
```
xtremerui, line 75: is there a case that this is not set?

evanchaoli: I am not sure. But I just copied the piece of existing code:

https://github.com/concourse/concourse/blob/c49d370b904af1950c18b916d0e19fa9ab872d93/atc/exec/step_metadata.go#L62-L64

taylorsilva: `ExternalURL` will not be set if it wasn't specified here: https://github.com/concourse/concourse/blob/c49d370b904af1950c18b916d0e19fa9ab872d93/atc/atccmd/command.go#L127

atc/exec/task_step_test.go
```diff
@@ -176,6 +177,10 @@ var _ = Describe("TaskStep", func() {
 			fakePool.FindOrSelectWorkerReturns(chosenWorker, nil)
 		})
 
+		It("Task env includes atc external url", func(){
+			Expect(chosenContainer.Spec.Env).To(ContainElement("ATC_EXTERNAL_URL=http://foo.bar"))
```
xtremerui, line 181:
```suggestion
			Expect(chosenContainer.Spec.Env).To(ConsistOf("ATC_EXTERNAL_URL=http://foo.bar", "SECURE=secret-task-param"))
```

taylorsilva:
@evanchaoli Not sure if you tried this, but for your task in your example, could you not ping the internal site to see if you can reach it or not? Feels like there should be a way to test if you're in the secure or non-secure network.

I also don't think this decreases complexity at all, unless there's a division of ownership between the pipeline authors and the authors of this task with the `SECURED` param. Then this PR would decrease complexity for the pipeline authors, but not at all for the task authors. The task authors would be switching out this `SECURED` param for another. They still need to code in the logic of `IF secure network THEN action a ELSE action b`.

---

Another way to solve this too could be with storing info like this in a shared path: https://concourse-ci.org/vault-credential-manager.html#vault-shared-path

You could store the external url, or maybe another variable like `INTERNAL` that's `true` or `false`. Then your task authors can add this to their task config

```yaml
platform: linux

...

params:
  INTERNAL: ((internal))
```

The benefit of this way is that you can implement it today and don't need to do an upgrade of Concourse to get this. You can configure shared paths with vault, setup that shared var, and change your task, all without upgrading Concourse.

I don't know all the requirements your task has, but I think storing this info in a vault secret is better than using the `ATC_EXTERNAL_URL` as a "proxy" for determining a tasks network location. wdyt?

taylorsilva (changes requested):
The suggested use case for this is looking like an anti-pattern. There are other ways their individual problem can be tackled without relying on this metadata.

evanchaoli:
@taylorsilva Thanks for suggesting solutions, however I don't think they work for us.

We have many internal resource types facing the similar situation. Storing those data to Vault just making Vault hard to maintain, and error-prone, as which require each resource type author to request populate some data into Vault shared path.

As task doesn't have build metadata, pipelines have been using a separate resource to fetch build metadata and input to task; and some thing can be implemented as task, but instead implemented as resource just in order to get build metadata, which just adds complexity of pipelines. Many users had requested the feature of expose same build meta to `task` as `put`/`get`. This PR is just a compromise with only exposing atc url.

Maybe I provided a bad use case, which made you consider anti-pattern. But general idea of this PR is, make a task to be aware of where it runs, similar to put/get.

xtremerui:
@clarafu wdyt?

PR #8306, opened 2022-04-29
Inherite env proxy configure when tls enabled

### LINKED ISSUE #8296: Proxy settings for skyHttpClient are lost when tls is enabled
Our Credentials provider sits behind a proxy, and if the TLS is enabled on concourse, then the HTTP Client that speaks to the credentials provider, fails to connect. This is because the `HTTPTransport` used when creating the client, replaces the default transport while enabling TLS [here](https://github.com/concourse/concourse/blob/master/atc/atccmd/command.go#L1416).

xtremerui:
closes #8296

* [x] done
* [ ] todo

## Release Note

 * Fix a bug that proxy setting through env var got lost when TLS is enabled by [--tls-bind-port](https://github.com/concourse/concourse/blob/master/atc/atccmd/command.go#L117)

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -1414,6 +1414,7 @@ func (cmd *RunCommand) skyHttpClient() (*http.Client, error) {
 		}
 
 		httpClient.Transport = &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
 			TLSClientConfig: &tls.Config{
 				RootCAs: certpool,
 			},


atc/atccmd/command.go
```diff
@@ -1414,6 +1414,7 @@ func (cmd *RunCommand) skyHttpClient() (*http.Client, error) {
 		}
 
 		httpClient.Transport = &http.Transport{
+			Proxy: http.ProxyFromEnvironment,
```
asehra, line 1417: Would it be better to copy `DefaultTransport` and only update the `TLSClientConfig`?
DefaultTransport as defined [here](https://pkg.go.dev/net/http#RoundTripper) appears to have other defaults that we'd lose if we build our own Transport:

```go
var DefaultTransport RoundTripper = &Transport{
	Proxy: ProxyFromEnvironment,
	DialContext: defaultTransportDialContext(&net.Dialer{
		Timeout:   30 * time.Second,
		KeepAlive: 30 * time.Second,
	}),
	ForceAttemptHTTP2:     true,
	MaxIdleConns:          100,
	IdleConnTimeout:       90 * time.Second,
	TLSHandshakeTimeout:   10 * time.Second,
	ExpectContinueTimeout: 1 * time.Second,
}
```

asehra: In fact, we already use `DefaultTransport` in the line 1423 below. So maybe we do that before the `if` condition, and only assign the tls config within the `if` condition. That way the `else` can also be removed. 

```go
	httpClient.Transport = http.DefaultTransport // <- here

	if cmd.isTLSEnabled() {
		certpool, err := x509.SystemCertPool()
		if err != nil {
			return nil, err
		}

		if !cmd.LetsEncrypt.Enable {
			cert, err := tls.LoadX509KeyPair(string(cmd.TLSCert), string(cmd.TLSKey))
			if err != nil {
				return nil, err
			}

			x509Cert, err := x509.ParseCertificate(cert.Certificate[0])
			if err != nil {
				return nil, err
			}

			certpool.AddCert(x509Cert)
		}

		httpClient.Transport.TLSClientConfig: &tls.Config{
			RootCAs: certpool,
		}
	} // <- no `else` needed here
```

You could also optionally make a copy of the `http.DefaultTransport` instead of directly assigning to httpClient, to avoid messing with a Transport instance that we didn't actually create. (although not strictly necessary)

xtremerui: This is exact the concern I have here. 

With or without the `Proxy` field, it does feel like the way you suggested is more consistent and cleaner. However, we don't know why the code was like that when it was first written. It is impossible to figure out the intend of the author.

So instead of changing the behaviour when TLS is enabled (by using the default `Transport`), I think keep it the same but only adding `Proxy` field is a safer bet.

clarafu (approved):

PR #8322, opened 2022-05-06
Depop avoid duplicating parallel volume streams

evanchaoli:

This is PR pulled #8047 and addressed review comments in order to move forward the work.

Closes #8047

* [x] done

## Release Note

* Steps that stream volumes will now use a global (per worker) lock to ensure identical volumes are not streamed more times than they need to be
* A new waiting-for-streamed-volume/waiting for volume <name> to be streamed by another step event is included in build step logs where this behavior occurs

### DIFF
--- a/atc/db/lock/lock.go
+++ b/atc/db/lock/lock.go
@@ -60,6 +60,10 @@ func NewInMemoryCheckBuildTrackingLockID(checkableType string, checkableId int) 
 	return LockID{LockTypeInMemoryCheckBuildTracking, lockIDFromString(fmt.Sprintf("%s-%d", checkableType, checkableId))}
 }
 
+func NewVolumeStreamingLockID(volume string, worker string) LockID {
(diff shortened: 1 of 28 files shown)


xtremerui (approved):
LGTM

andy-paine:
Thanks for picking this up @evanchaoli - I've been working with another client for a few months now so didn't have the time to look at this so it is very much appreciated you taking it forward. The original work was done on behalf of Depop so if they could be properly credited in the release notes @xtremerui that would be great. Thanks all!

evanchaoli:
@andy-paine You may go ahead address review comments in #8047 and merge that one if you have time. I had to create this new PR as you didn't have response at the time.

evanchaoli:
@xtremerui Even if this one instead of #8047 is merged, pls only mention @andy-paine in the release note.

clarafu (approved):

PR #8324, opened 2022-05-06
Bump vault api and add srv lookup flag

### LINKED ISSUE #8304: Concourse's vault client does a ton of SRV lookups
## Steps to reproduce

- Deploy concourse with the vault integration
- configure the vault URL without the port, e.g.: `https://vault.enterprise.tld`
- ensure you have no SRV record for `_http._tcp.vault.enterprise.tld`
- check the DNS SRV requests that go out from concourse / hit your DNS server

=> you should see a ton of requests, all of them not returning anything
=> concourse is, however, still working correcly

## Expected results

When concourse runs, it doesn't do too many (sorry for being vague here) SRV lookups.

## Actual results

We see a lot of SRV lookups, and they all fail, because we don't have the SRV records in DNS.

On one concourse we've seen ~14 of those lookups per second (would amount to ~1.2m per day); and we have about 10ish conci around.

## Additional context

We have concourse talking to vault and we've set the vault URL to `https://vault.enterprise.tld` (note: we don't specify the port explicitly here).
We see a lot of DNS SRV queries for `_http._tcp.vault.enterprise.tld` which fail.
Everything still works, but the team that manages the resolvers is not happy about us, rather our conci, hammering their systems.

We think this is related to [this issue](https://github.com/hashicorp/vault/issues/5540), which shows, that if the vault URL has not explicitly specified a port, a SRV request is issued, not matter what.
There is a new version of the package (`github.com/hashicorp/vault/api@v1.5.0`) which still seems to show the same behaviour, but gives us a switch (`VAULT_SRV_LOOKUP`) to turn of SRV lookups.

I am not sure what we can / should do about that, but some thoughts:
- In the [concourse docs](https://concourse-ci.org/vault-credential-manager.html) explicitly call out, that vault will do SRV lookups when there is no port in the vault URL
- Update to `github.com/hashicorp/vault/api@v1.5.0` and document the option to dsiable SRV requests via `VAULT_SRV_LOOKUP`

## Triaging info

* Concourse version: `v7.4.0` -- but we think pretty much any concourse with the same vault package will show that behaviour
* Did this used to work? Didn't check if there was different behaviour in older versions of `github.com/hashicorp/vault/api` :shrug:

xtremerui:

closes #8304 

 * Bump vualt api to version 1.5.0.
 * Add `--disable-srv-lookup` flag to vault configure

* [x] done
* [ ] todo

## Release Note
 * Bump Vault API package to latest version and add `--disable-srv-lookup` flag to Vault configure. If your current Vault URL contains a port number, this change makes no impact. If your Vault URL dose not contain port number, by default SRV lookup is opt-in for backward compatibility. In this case, one can use the flag to disable the feature to avoid unnecessary requests from Vault client.


clarafu (approved):

clarafu (approved):
Sorry im not sure what you changed in the force push haha but it looks the same to me 😆

PR #8342, opened 2022-05-13
Force checks on nested resource types when manually triggered build

clarafu:

Previously we had it that manually triggered builds would not force any nested resource types or images to skip its interval for its check plan. This is because we did not want all the nested resource types to check everytime a build is manually triggered. This might cause a lot of unnecessary checks because most users don't care about its resource type versions. 

But Rui brought up a good point that image resources cannot be manually checked through fly so the only way to force a check for them is through a manual build. This PR makes it so that a manually triggered build will make the nested resource types and images skip its interval. One thing that will help alleviate the increased number of checks that this change will is result in is that we will only ever check a resource type once in the build, so forcing the checks wont result in the same resource type getting checked multiple times.

* [x] done
* [ ] todo

## Release Note

* When a build is manually triggered, it will cause any nested resource types or images to skip its checking interval, essentially forcing a check. This will not result in the same resource type getting checked multiple times if it appears multiple times in a build.

### DIFF
--- a/atc/api/jobserver/create_build.go
+++ b/atc/api/jobserver/create_build.go
@@ -75,10 +75,7 @@ func (s *Server) CreateJobBuild(pipeline db.Pipeline) http.Handler {
 					resourceTypes,
 					version,
 					true,
-					// Don't recursively skip the interval for checks for parent resource
-					// types - doing so would result in a ton of checks running
-					// unnecessarily.
-					false,
+					true,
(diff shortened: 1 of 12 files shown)


xtremerui:
@clarafu are you willing to fixt the test failure or could I continue on your branch?

xtremerui (approved):
LGTM!

Issue #3486, opened 2019-03-12
OIDC with Okta doesn't get username or groups


MartinLeedotOrg:
## Bug Report
I have followed the steps at:
https://concourse-ci.org/generic-oidc-auth.html

I have created an app in Okta with the implicit flow and mapped all groups.

## Steps to Reproduce

Following environment variables:
export CONCOURSE_OIDC_DISPLAY_NAME=Okta
export CONCOURSE_OIDC_CLIENT_ID=clientid
export CONCOURSE_OIDC_CLIENT_SECRET=clientsecret
export CONCOURSE_OIDC_ISSUER=https://issuer.issuer.com
export CONCOURSE_MAIN_TEAM_OIDC_GROUP=dev-concourse
export CONCOURSE_OIDC_GROUPS_KEY='groups'
export CONCOURSE_OIDC_SCOPE='openid groups'

## Expected Results

Username and groups to be passed to concourse to allow user to join the main team

## Actual Results

HTTP 401 returned to clients, following in logs:
```
{"timestamp":"2019-03-12T17:03:17.480427687Z","level":"info","source":"atc","message":"atc.dex.event","data":{"fields":{},"message":"login successful: connector \"oidc\", username=\"\", email=\" foo@bar.com (unverified)\", groups=[]","session":"6"}}
{"timestamp":"2019-03-12T17:03:17.532387726Z","level":"error","source":"atc","message":"atc.sky.callback.failed-to-issue-concourse-token","data":{"error":"user doesn't belong to any team","session":"5.40"}}
```
## Version Info

* Concourse version: 5.0.0
* Deployment type (BOSH/Docker/binary): binary
* Infrastructure/IaaS: debian
* Browser (if applicable): N/A
* Did this used to work? N/A

mattskone:
I'm experiencing and troubleshooting the same, but with 4.2.2 on Kubernets via Helm.  Possibly some overlap with https://github.com/concourse/concourse/issues/2683?

MartinLeedotOrg:
In okta you can specify a regex for which groups get passed through in the groups claim. I've specified .* (and verified that works in another application).

`dev-concourse` is a group I've set up for this test.

I can see that username might not be supported - looks like it falls back to sub? That's fine for me - less than ideal but works. Groups not working is a bigger problem though.

I can have okta only allow authentication for certain users, is there a combination of the configuration that would allow me to say "All users authenticating by OIDC should be in the main team"?

[...]

PR #8353, opened 2022-05-17
Introduce OIDC get user info flag

danpilch:

* [x] Add `CONCOURSE_OIDC_DISABLE_GET_USER_INFO` flag

## Notes to reviewer

This PR adds the `dex` GetUserInfo flag which is required to get extra user info e.g. groups claim from an OIDC provider (Okta in this case). Please see https://github.com/dexidp/dex/blob/9cd29bdee04ca38104a4488b211208694501d54f/connector/oidc/oidc.go#L48-L51 without this flag, we cannot get groups from Okta and fail to authenticate to `fly` successfully.

Notably, you can see from this issue https://github.com/concourse/concourse/issues/3486 that other people have the same problem.

## Release Note

* Add `CONCOURSE_OIDC_DISABLE_GET_USER_INFO` flag. OIDC connector will now fetch additional claims from [OpenID UserInfo endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo). This should fix the problem of configuring Concourse team auth by OIDC user groups due to groups claims missing in some identity providers' auth response.

### DIFF
--- a/skymarshal/skycmd/oidc_flags.go
+++ b/skymarshal/skycmd/oidc_flags.go
@@ -30,7 +30,7 @@ type OIDCFlags struct {
 	InsecureSkipVerify        bool        `long:"skip-ssl-validation" description:"Skip SSL validation"`
 	DisableGroups             bool        `long:"disable-groups" description:"Disable OIDC groups claims"`
 	InsecureSkipEmailVerified bool        `long:"skip-email-verified-validation" description:"Ignore the email_verified claim from the upstream provider, treating all users as if email_verified were true."`
-	GetUserInfo               bool        `long:"get-user-info" description:"When enabled, the OpenID Connector will query the UserInfo endpoint for additional claims, e.g. groups"`
+	DisableGetUserInfo        bool        `long:"disable-get-user-info" description:"When disabled, the OpenID Connector will not query the UserInfo endpoint for additional claims, e.g. groups"`
 }
 
 func (flag *OIDCFlags) Name() string {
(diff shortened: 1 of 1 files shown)


xtremerui:
Thank you for the PR! Could you fix the DCO error by signing your commit?

skymarshal/skycmd/oidc_flags.go
```diff
@@ -79,6 +80,7 @@ func (flag *OIDCFlags) Serialize(redirectURI string) ([]byte, error) {
 		RedirectURI:               redirectURI,
 		InsecureEnableGroups:      !flag.DisableGroups,
 		InsecureSkipEmailVerified: flag.InsecureSkipEmailVerified,
+		GetUserInfo:               flag.GetUserInfo,
```
xtremerui, line 83: Do you think it is really necessary to provide this configure or we can just set it to true? IMO, groups claim is always (for most users) something we want for Concourse auth for teams. 

Would `oidc.provider.UserInfo()` require certain config in IDP (e.g. Okta) to enable the userinfo endpoint? Would there be any risk by always quering this endpoint?

danpilch: I think we could always set it to `true`. I added it as a flag in case it causes any edge case issues so there's always the option of disabling it. I don't think you need to enable or configure anything in Okta to enable the userinfo endpoint.

danpilch:
> Thank you for the PR! Could you fix the DCO error by signing your commit?

Fixed!

skymarshal/skycmd/oidc_flags.go
danpilch: I'm happy to set it to `true` if you want

xtremerui: hmm, with further looking into the OpenID spec and the impl of oidc.UserInfo(). I think your concern of the edge case is valid. Though [userinfo endpoint](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo) is a standard, there might still be issues when querying it, depends on the IDP server.

In this case, I think we could make the flag to
```suggestion
		GetUserInfo:               !flag.DisableGetUserInfo,
```

danpilch:
OK I have made the relevant changes

xtremerui (approved):

PR #8362, opened 2022-05-19
Fix default username prompt for local logins

### LINKED ISSUE #8361: Bad textbox caption on login page
## Summary

We upgraded a cluster to 7.7.1. When I login with "username/password", it prompts "email address":

bad-login-form

## Steps to reproduce

Open 7.7.1 UI, click on "login" at top-right corner, then choose "username/password".

## Expected results

The first textbox should prompt "Username".

## Actual results

It now prompts "email".

## Triaging info

* Concourse version: 7.7.1
* Browser (if applicable): Any
* Did this used to work? Yes

phan-do:

closes #8361

## Notes to reviewer

64e504b introduced an incorrect request path check which caused the default value of the username prompt for local logins (`username`) to never be set. Specifically, the path being checked was only a prefix of the full request path.

## Release Note

- Ensure the default username prompt for local logins is properly set.

### DIFF
--- a/skymarshal/dexserver/web/templates/password.html
+++ b/skymarshal/dexserver/web/templates/password.html
@@ -1,7 +1,7 @@
 {{ template "header.html" . }}
 
 {{ $usernamePrompt := .UsernamePrompt | lower }}
-{{ if eq .ReqPath "/sky/issuer/auth/local" }}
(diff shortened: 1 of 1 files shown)


xtremerui (approved):
Thank you for the fix!

PR #8374, opened 2022-05-23
set_pipeline step supports detach mode

evanchaoli:

If jobA of pipelineA call `set_pipeline` step to set pipelineB, then jobA is pipelineB's parent job. Currently Concourse has a logic that, if jobA sets another pipeline than pipelineB, then pipelineB will be automatically archived, which helps prevent from orphan pipelines.

`set_pipeline` step supports parameter `team` only when parent job runs from `main` team. We have a use case where a service pipeline running from `main` team may set child parent pipelines per needs. In other words, build-1 may set a child pipeline to teamA, then build-2 may set another child pipeline to teamB, and so on ... The behavior will trigger auto pipeline archive, thus this use case doesn't work with `set_pipeline` step.

I think my use case makes sense from `main` team. To not break the current logic, I am adding a `detach` flag to `set_pipeline` step. A parent job needs to explicitly turn on the flag so that child pipeline will be set in detach mode (same as set via `fly set-pipeline`, so that child pipeline will never be auto archived.

* [x] done

## Release Note

* Add `detach` flag to `set_pipeline` step. Only the `main` team can set child pipelines in detach mode. In detach mode, child pipelines will never be automatically archived.

### DIFF
--- a/atc/builds/planner.go
+++ b/atc/builds/planner.go
@@ -264,6 +264,7 @@ func (visitor *planVisitor) VisitSetPipeline(step *atc.SetPipelineStep) error {
 		Vars:         step.Vars,
 		VarFiles:     step.VarFiles,
 		InstanceVars: step.InstanceVars,
+		Detach:       step.Detach,
 	})
 
 	return nil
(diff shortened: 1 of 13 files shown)


timota:
any plans to add this feature to concourse? It will be extremely useful.

taylorsilva:
Not closing this, even though I'm guessing Evan may not be around to help get this through. Would still be nice to have so leaving this open so someone can refer to the code.

PR #8390, opened 2022-05-31
Fix DB integer out of range error

### LINKED ISSUE #6354: pq: nextval: reached maximum value of sequence "resource_config_versions_id_seq"
## Summary

Concourse was unable to check for any new resources, no pipelines were able to run. Displaying the error:

```
run check step: save versions: pq: nextval: reached maximum value of squence "resource_config_versions_id_seq" (2147483647)
```

Although we only had `141431` entries:

```
concourse=> SELECT count(*) FROM resource_config_versions;
 count
--------
 141431
(1 row)
```

The data:

```
concourse=> select * from resource_config_versions order by resource_config_versions desc limit 10;
     id     |                   version                    |           version_md5            | metadata | check_order | resource_config_scope_id | span_context
------------+----------------------------------------------+----------------------------------+----------+-------------+--------------------------+--------------
 2147286180 | {"time": "2020-12-10T05:19:14.134465595Z"}   | 4d0d977d2d300856c4ba8914b1d646fb | []       |       27167 |                     1484 | {}
 2147216221 | {"time": "2020-12-10T05:12:23.072239534Z"}   | d85659bb1965d9b03ad2ab46e4ed8394 | []       |        1747 |                     2322 | {}
...
```

## Steps to reproduce

```
concourse=> \d+ resource_config_versions_id_seq
          Sequence "public.resource_config_versions_id_seq"
  Type   | Start | Minimum |  Maximum   | Increment | Cycles? | Cache
---------+-------+---------+------------+-----------+---------+-------
 integer |     1 |       1 | 2147483647 |         1 | no      |     1
Owned by: public.resource_config_versions.id
```

`resource_config_versions_id_seq` is an int, although `id` is a bigint:

```
concourse=> \d+ resource_config_versions
                                                         Table "public.resource_config_versions"
          Column          |  Type   | Collation | Nullable |                       Default                        | Storage  | Stats target | Description
--------------------------+---------+-----------+----------+------------------------------------------------------+----------+--------------+-------------
 id                       | bigint  |           | not null | nextval('resource_config_versions_id_seq'::regclass) | plain    |              |
 version                  | jsonb   |           | not null |                                                      | extended |              |
 version_md5              | text    |           | not null |                                                      | extended |              |
 metadata                 | jsonb   |           | not null | 'null'::jsonb                                        | extended |              |
 check_order              | integer |           | not null | 0                                                    | plain    |              |
 resource_config_scope_id | integer |           | not null |                                                      | plain    |              |
 span_context             | jsonb   |           |          |                                                      | extended |              |
 old_id                   | bigint  |           |          |                                                      | plain    |              |
 compacted_id             | bigint  |           |          |                                                      | plain    |              |
Indexes:
    "resource_config_versions_pkey" PRIMARY KEY, btree (id)
    "resource_config_scope_id_and_version_md5_unique" UNIQUE CONSTRAINT, btree (resource_config_scope_id, version_md5)
    "resource_config_versions_check_order_idx" btree (resource_config_scope_id, check_order DESC)
    "resource_config_versions_version" gin (version jsonb_path_ops) WITH (fastupdate='false')
Foreign-key constraints:
    "resource_config_versions_resource_config_scope_id_fkey" FOREIGN KEY (resource_config_scope_id) REFERENCES resource_config_scopes(id) ON DELETE CASCADE
Access method: heap
```

We added `old_id` and `compacted_id` while investigating.

## Expected results

- No error message.
- `resource_config_versions` `resource_config_versions_id_seq` does not grow in an unbounded way

## Additional context

Upon resolving the issue the id column in resource_config_versions began to skip entries immediately:
Screenshot 2020-12-10 at 11 12 18

We resolved the issue by:

```
BEGIN;  
  -- create a temporary sequence to compact the ids
  CREATE SEQUENCE temp_resource_config_versions_id_seq;
  LOCK TABLE resource_config_versions IN EXCLUSIVE MODE;
  -- bump all ids above the sequence value
  UPDATE resource_config_versions SET id = id + CURRVAL('resource_config_versions_id_seq');
  -- update the main id column 
  UPDATE resource_config_versions SET id = NEXTVAL('temp_resource_config_versions_id_seq');
  -- reset the main sequence to be the next value
  ALTER SEQUENCE resource_config_versions_id_seq RESTART WITH NEXTVAL('temp_resource_config_versions_id_seq');
  DROP SEQUENCE temp_resource_config_versions_id_seq;
COMMIT;
```

This allowed us to continue using Concourse but we are unaware of the long term consequences.

- `enableGlobalResources: false`

## Triaging info

- Seems similar to https://github.com/concourse/concourse/issues/5667, but this issue relates to `resource_config_versions_id_seq` not `resource_config_versions_id`. 

* Concourse version: v6.6.0
* Did this used to work? N/A

### LINKED ISSUE #8364: Concourse cannot handle more than 2147483647 checks/builds
## Summary

When build_id id in the builds table exceeds 2147483647, all future builds fail to create.

Errors in the logs like:
```
concourse-web-1     | {"timestamp":"2022-05-19T20:54:03.056255000Z","level":"error","source":"atc","message":"atc.tracker.tick.run.lock.acquire.failed-to-register-in-db","data":{"build":"26","build_id":2147483660,"error":"pq: value \"2147483660\" is out of range for type integer","id":[1,2147483660],"job":"job","pipeline":"example","session":"26.486.4.1.1","team":"main"}}
concourse-web-1     | {"timestamp":"2022-05-19T20:54:03.056311307Z","level":"error","source":"atc","message":"atc.tracker.tick.run.failed-to-get-lock","data":{"build":"26","build_id":2147483660,"error":"pq: value \"2147483660\" is out of range for type integer","job":"job","pipeline":"example","session":"26.486.4","team":"main"}}
concourse-db-1      | 2022-05-19 20:54:03.056 UTC [70] ERROR:  value "2147483660" is out of range for type integer
concourse-db-1      | 2022-05-19 20:54:03.056 UTC [70] CONTEXT:  unnamed portal parameter $2 = '...'
concourse-db-1      | 2022-05-19 20:54:03.056 UTC [70] STATEMENT:  SELECT pg_try_advisory_lock($1,$2)
concourse-db-1      | 2022-05-19 20:54:03.056 UTC [70] ERROR:  value "2147483656" is out of range for type integer
concourse-db-1      | 2022-05-19 20:54:03.056 UTC [70] CONTEXT:  unnamed portal parameter $2 = '...'
```

## Steps to reproduce

1. Start up the dev env: `docker-compose up`
2. Start the every 30s pipeline: `fly -t dev set-pipeline -p example -c examples/pipelines/time-triggered.yml`
3. Unpause the pipeline
4. Let it run for a few minutes, but may not be strictly necessary
5. Log into the DB: ./hack/db
6. Restart the builds_id_seq sequence to start at 2147483647.
> ```
> concourse=# alter sequence builds_id_seq restart with 2147483647;
> ALTER SEQUENCE
> ```
> 

## Expected results

Builds continue to work

## Actual results

Concourse system stops executing new builds

## Additional context

1. Unupdated schema columns
We noticed there were a few build_id and related fields that were not updated in the transition to bigint:

https://github.com/concourse/concourse/blob/8c8eb0565101b287107fe9d36ef970505ad166b7/atc/db/migration/migrations/1603401316_alter_build_integers_to_bigints.up.sql

```
table: build_comments
column: build_id

table: builds
column: rerun_of

table: successful_build_outputs
column: rerun_of

table: jobs
column: first_logged_build_id
```
There may be more.

2. In our production Concourse cluster on which this originally happened,
one of the errors we saw was:
```
LOG: execute <unnamed>: SELECT completed FROM builds WHERE id = $1
DETAIL: parameters: $1 = '2147493588'
ERROR: value "2147493588" is out of range for type integer
```
What’s weird is that builds and id are already bigint, including the sequence.

That error message, happens if Postgres is told to typecast from a string to an integer, as can be seen reproduced below:
```
testdb=# CREATE TABLE foobar (test_id bigint);
CREATE TABLE
 
testdb=# SELECT * FROM foobar WHERE test_id = 2147483648;
test_id
---------
(0 rows)
 
testdb=# SELECT * FROM foobar WHERE test_id = '2147483648'::integer;
ERROR: value "2147483648" is out of range for type integer
```

So likely the case, that this is somewhere in code a string is being passed to Postgres and told to typecast to an integer,
before storing to a bigint field.

## Triaging info
Concourse version: 7.4.4.  Test was against master Concourse branch.
Browser (if applicable):
Did this used to work? Not likely

xtremerui:

closes #8364 
closes #6354

 * convert build id related columns to type bigint in DB
 * build tracking lock id will be computed from build_id%max_int
 * for build events, select on build id only when it is beyond max_int so it wont query old build id (integer type) with a beyond max_int value and fail the query

* [x] done
* [ ] todo

## Release Note

 * Fix DB out of range error due to build numbers exceed the integer limit

### DIFF
--- a/atc/db/build_event_source.go
+++ b/atc/db/build_event_source.go
@@ -3,12 +3,12 @@ package db
 import (
 	"encoding/json"
 	"errors"
+	"math"
 	"strconv"
 	"sync"
 
 	sq "github.com/Masterminds/squirrel"
 	"github.com/concourse/concourse/atc"
-	"github.com/concourse/concourse/atc/db/lock"
 	"github.com/concourse/concourse/atc/event"
 )
 
@@ -126,7 +126,7 @@ func (source *buildEventSource) collectEvents(from uint) {
 			From(source.table)
 
 		var query sq.SelectBuilder
-		if source.buildID > lock.MaxInt {
+		if source.buildID > math.MaxInt32 {
 			query = eventsQuery.Where(sq.Eq{"build_id": source.buildID})
 		} else {
 			query = eventsQuery.Where(sq.Or{
--- a/atc/db/build_event_source.go
+++ b/atc/db/build_event_source.go
@@ -8,6 +8,7 @@ import (
 
 	sq "github.com/Masterminds/squirrel"
 	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/atc/db/lock"
 	"github.com/concourse/concourse/atc/event"
 )
 
@@ -121,12 +122,20 @@ func (source *buildEventSource) collectEvents(from uint) {
 			return
 		}
 
-		rows, err := psql.Select("event_id", "type", "version", "payload").
-			From(source.table).
-			Where(sq.Or{
+		eventsQuery := psql.Select("event_id", "type", "version", "payload").
+			From(source.table)
+
+		var query sq.SelectBuilder
+		if source.buildID > lock.MaxInt {
+			query = eventsQuery.Where(sq.Eq{"build_id": source.buildID})
+		} else {
+			query = eventsQuery.Where(sq.Or{
 				sq.Eq{"build_id": source.buildID},
 				sq.Eq{"build_id_old": source.buildID},
-			}).
+			})
+		}
+
+		rows, err := query.
 			Where(sq.Gt{"event_id": cursor}).
 			OrderBy("event_id ASC").
 			Limit(uint64(batchSize)).
--- a/atc/db/lock/lock.go
+++ b/atc/db/lock/lock.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"hash/crc32"
+	"math"
 	"strconv"
 	"strings"
 	"sync"
@@ -15,7 +16,6 @@ import (
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
 
 const (
-	MaxInt                         = 2147483647
 	LockTypeResourceConfigChecking = iota
 	LockTypeBuildTracking
 	LockTypeBatch
@@ -30,7 +30,7 @@ const (
 var ErrLostLock = errors.New("lock was lost while held, possibly due to connection breakage")
 
 func NewBuildTrackingLockID(buildID int) LockID {
-	return LockID{LockTypeBuildTracking, buildID % MaxInt}
+	return LockID{LockTypeBuildTracking, buildID % math.MaxInt32}
 }
 
 func NewResourceConfigCheckingLockID(resourceConfigID int) LockID {
@@ -42,7 +42,7 @@ func NewTaskLockID(taskName string) LockID {
 }
 
 func NewVolumeCreatingLockID(volumeID int) LockID {
-	return LockID{LockTypeVolumeCreating, volumeID}
+	return LockID{LockTypeVolumeCreating, volumeID % math.MaxInt32}
 }
 
 func NewDatabaseMigrationLockID() LockID {
--- a/atc/db/lock/lock.go
+++ b/atc/db/lock/lock.go
@@ -29,6 +29,14 @@ const (
 
 var ErrLostLock = errors.New("lock was lost while held, possibly due to connection breakage")
 
+/*
+	When adding a new lock type or update existing ones, consider if
+	the ID will be exhausting max int32 ID pool quickly. If yes,
+	use ID % math.MaxInt32 to prevent pg_try_advisory_lock(int, int)
+	query from failing by "integer out of range" error.
+	Refer to https://github.com/concourse/concourse/pull/8390
+*/
+
 func NewBuildTrackingLockID(buildID int) LockID {
 	return LockID{LockTypeBuildTracking, buildID % math.MaxInt32}
 }
--- a/atc/db/lock/lock.go
+++ b/atc/db/lock/lock.go
@@ -15,6 +15,7 @@ import (
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
 
 const (
+	MaxInt                         = 2147483647
 	LockTypeResourceConfigChecking = iota
 	LockTypeBuildTracking
 	LockTypeBatch
@@ -29,7 +30,7 @@ const (
 var ErrLostLock = errors.New("lock was lost while held, possibly due to connection breakage")
 
 func NewBuildTrackingLockID(buildID int) LockID {
-	return LockID{LockTypeBuildTracking, buildID}
+	return LockID{LockTypeBuildTracking, buildID % MaxInt}
 }
 
 func NewResourceConfigCheckingLockID(resourceConfigID int) LockID {
--- a/atc/db/migration/migrations/1653924132_int_to_bigint.down.sql
+++ b/atc/db/migration/migrations/1653924132_int_to_bigint.down.sql
@@ -1,0 +1,16 @@
+ALTER SEQUENCE resource_config_versions_id_seq AS integer;
+
+ALTER TABLE build_comments
+    ALTER COLUMN build_id TYPE integer;
+
+ALTER TABLE builds
+    ALTER COLUMN rerun_of TYPE integer;
+
+ALTER TABLE successful_build_outputs
+    ALTER COLUMN rerun_of TYPE integer;
+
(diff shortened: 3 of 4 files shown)


atc/db/lock/lock.go
```diff
@@ -15,6 +15,7 @@ import (
 //go:generate go run github.com/maxbrunsfeld/counterfeiter/v6 -generate
 
 const (
+	MaxInt                         = 2147483647
```
evanchaoli, line 18: I don't think this const is needed, we can just use `math.MaxInt32`.

atc/db/build_event_source.go
```diff
@@ -121,12 +122,20 @@ func (source *buildEventSource) collectEvents(from uint) {
...
 		}
 
-		rows, err := psql.Select("event_id", "type", "version", "payload").
-			From(source.table).
-			Where(sq.Or{
+		eventsQuery := psql.Select("event_id", "type", "version", "payload").
+			From(source.table)
+
+		var query sq.SelectBuilder
+		if source.buildID > lock.MaxInt {
```
evanchaoli, line 129: As the other comment, `lock.MaxInt` is not needed, we can just use `math.MaxInt32` here.

atc/db/lock/lock.go
```diff
@@ -29,7 +30,7 @@ const (
 var ErrLostLock = errors.New("lock was lost while held, possibly due to connection breakage")
 
 func NewBuildTrackingLockID(buildID int) LockID {
-	return LockID{LockTypeBuildTracking, buildID}
+	return LockID{LockTypeBuildTracking, buildID % MaxInt}
```
evanchaoli, line 33: Like I suggested in https://github.com/concourse/concourse/issues/8364#issuecomment-1141557700, I still think the change should be in `LockID.toDBArgs()`. This is not just a problem with build id, I would consider volume id, container id to be larger than build id. As the restriction comes from PG, the fix should be as close as possible to db related code.

If you change `LockID.toDBArgs()`, please add some tests accordingly.

xtremerui: For hush-house (being up for 4 years), we have the maximum ids as following
```
1081477231 build id
214934158  container id
410257222  volumn id
```

Could you list the ids in runway for comparation? Ideally I want to keep the impact as minimum as possible, until we see the same problem reported for container/volumn ID.

evanchaoli: Our prod cluster has been running for ~3 years. To roughly calculate, say 1000 days, `10422031 / 1000 = 10422` builds per day, that's a reasonable number. As most of builds are triggered by code check-in. 

Remember that, 6.x uses `checks` table, so that checks doesn't consume any build id. 

From the other perspective, majority workloads on our cluster are coming from checks, that's reason why I spent so much efforts on optimizing checks.

xtremerui: I don't really want to add this overhead dividend operation to not needed lock ids so to me LockID.toDBArgs() is the least place I want to solve the problem. Plus the way you suggested will also calculate the lock_type/maxint32 which is totally a waste.

 I agree that volume id and container id should be considered though. However it seems there is no lock ID that depends on container ID.

evanchaoli: Comparing to the following DB operation `pg_try_advisory_lock`, cost of `%` operator can be ignored.

With your current implementation, there are still potential problems:

* When container_id related lock is added, or other lock types that need build id or volume id are added, future developers have to remember `% MaxInt32`, which is error-prone.
* Job id, resource configure id may eventually reach to MaxInt32 though possibility is low.

The following implementation will not `%` lock type:

```golang
func (l LockID) toDBArgs() []interface{} {
	result := []interface{}{l[0]}
	if len(l) > 1 {
             result = append(result, l[1] % math.MaxInt32
        }

	return result
}
```

xtremerui: > * When container_id related lock is added, or other lock types that need build id or volume id are added, future developers have to remember `% MaxInt32`, which is error-prone.

I have added a block of comment for furture reference of developing lock type. Plus I can hardly imagine there will be a change of existing lock type or adding a new one.

evanchaoli: TBH, I am not convinced. I still think the change should be in `toDBArgs()`. But anyway I don't want to block you.

evanchaoli:
@xtremerui It seems I have no permission to approve this PR, thus just putting a comment here: LGTM, Approved.

PR #8416, opened 2022-06-10
fixed a bug of error invalidated-worker-resource-cache-exists

### LINKED ISSUE #8413: Upgrading to 7.8.0 introduced `invalidated-worker-resource-cache-exists` in some of our workers
## Summary

We recently upgraded from 7.7.1 to 7.8.0 and it appeared to go fine, however pretty soon teams started complaining that their resource GETs were failing due to "invalidated-worker-resource-cache-exists", restarting the underlying node and clearing the resource cache did nothing, however restarting the pod (with no ephemeral storage) worked for a short time.
We can not reproduce this in our dev environment either, it only affects some teams too, not all of them.

## Steps to reproduce

Once upgraded, each team has it's own worker pod on it's own worker node with ephemeral storage as an EBS volume for duration of the underlying node. When left as is, some teams reported seeing "invalidated-worker-resource-cache-exists" - nothing changed after upgrade.

## Expected results

Before upgrade to 7.8.0, these issues were not happening - resource caches appeared to work fine

## Actual results

"invalidated-worker-resource-cache-exists" printed for several teams on resource "GET" step

## Additional context

From looking at the [code](https://github.com/concourse/concourse/blob/v7.8.0/atc/db/worker_resource_cache.go#L40), it appears as though:
```
// If the found worker resource cache's worker base resource type id is
// 0, then means the original source has been invalidated. But as this
// worker resource cache can still be found and used, so before this cache
// is GC-ed, let's not recreate the cache.
if uwrc.WorkerBaseResourceTypeID == 0 {
	return nil, false, errors.New("invalidated-worker-resource-cache-exists")
}
```
is the cause. Getting teams to flush their caches independently did not work.
We have upgraded the underlying EC2 instance for the database as it was looking a bit overutilised but this issue still occurred.

I have looked at the breaking Vault change as we did use that and although it sounded similar, I believe they use separate caches and we haven't got the creds cache enabled - so I think it is separate.

## Triaging info

* Concourse version: 7.8.0
* Browser (if applicable):
* Did this used to work? yes

evanchaoli:

closes #8413

* [x] done

## Release Note

Fixed a bug of error "invalidated-worker-resource-cache-exists".

### DIFF
--- a/atc/db/volume_test.go
+++ b/atc/db/volume_test.go
@@ -241,6 +241,7 @@ var _ = Describe("Volume", func() {
 			createdVolume = volumeOnWorker(scenario.Workers[0])
 			err = createdVolume.InitializeResourceCache(resourceCache)
 			Expect(err).ToNot(HaveOccurred())
+			Expect(createdVolume.Type()).To(Equal(db.VolumeTypeResource))
 		})
 
 		Context("when initialize created resource cache", func() {
@@ -356,15 +357,9 @@ var _ = Describe("Volume", func() {
 		})
 
 		Context("when streaming a volume cache that has been invalidated on the source worker", func() {
-			It("leaves the volume owned by the container", func() {
-				scenario.Run(
-					builder.WithBaseWorker(), // workers[1]
-				)
+			var streamedVolume db.CreatedVolume
 
-				cachedVolume := volumeOnWorker(scenario.Workers[0])
-				err := cachedVolume.InitializeResourceCache(resourceCache)
-				Expect(err).ToNot(HaveOccurred())
-
+			BeforeEach(func() {
 				scenario.Run(
 					builder.WithWorker(atc.Worker{
 						Name:          scenario.Workers[0].Name(),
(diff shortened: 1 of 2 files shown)


atc/db/volume_test.go
```diff
@@ -374,17 +369,76 @@ var _ = Describe("Volume", func() {
...
+			It("initializing resource cache on the new volume should not fail", func() {
+				err := newVolume.InitializeResourceCache(resourceCache)
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			It("leaves the volume owned by the container", func() {
+				Expect(newVolume.Type()).To(Equal(db.VolumeTypeContainer))
+			})
+
+			It("original invalidated resource cache volume still be found", func() {
```
xtremerui, line 435: Isn't this one the repeat of line line 418 - 421?

maybe you want to remove those in the before block?
```suggestion
			It("resource cache volume still be found on worker 1 after original invalidated on worker 0", func() {
```

xtremerui (approved):
LGTM

PR #8425, opened 2022-06-15
Allow SSM Parameters to specify a format using tags

ConnorWidtfeldt:

Currently, there is no way to access fields from vars sourced via SSM. This is because Concourse has no understanding of the format the parameter value is in. This change allows Concourse to read SSM parameter tags (with ssm:ListTagsForResource) and check for a tag called "format" which will behave the exact same as the load_vars format. Thus allowing JSON and YAML formats to be loaded. 

* [x] implement unmarshalling using a value set in the tags
* [ ] create tests for various formats, and failures in retrieving tags from AWS
* [ ] check for any remaining steps I could have missed in this...

## Notes to reviewer

Most of the code is a modified version of the load_vars handler. Maybe these should be moved into a shared location?

## Release Note

* The AWS SSM credential manager can now load different file formats when a format is specified in a tag named `format` on the parameter

### DIFF
--- a/atc/creds/ssm/ssm.go
+++ b/atc/creds/ssm/ssm.go
@@ -1,10 +1,13 @@
 package ssm
 
 import (
+	"encoding/json"
+	"fmt"
 	"strings"
 	"time"
 
 	"github.com/concourse/concourse/atc/creds"
+	"gopkg.in/yaml.v2"
 
 	"code.cloudfoundry.org/lager"
 	"github.com/aws/aws-sdk-go/aws"
@@ -42,6 +45,7 @@ func (s *Ssm) NewSecretLookupPaths(teamName string, pipelineName string, allowRo
 func (s *Ssm) Get(secretPath string) (interface{}, *time.Time, bool, error) {
 	// Try to get the parameter as string value, by name
 	value, expiration, found, err := s.getParameterByName(secretPath)
+
 	if err != nil {
 		s.log.Error("unable to retrieve aws ssm secret by name", err, lager.Data{
 			"secretPath": secretPath,
@@ -71,8 +75,17 @@ func (s *Ssm) getParameterByName(name string) (interface{}, *time.Time, bool, er
 		WithDecryption: aws.Bool(true),
 	})
 	if err == nil {
-		return *param.Parameter.Value, nil, true, nil
+		format, err := s.getParameterFormat(&name)
+		if err != nil {
+			return nil, nil, false, err
+		}
 
+		value, err := unmarshalParameterFormat(*param.Parameter.Value, format)
+		if err != nil {
+			return nil, nil, false, err
+		}
+
+		return value, nil, true, nil
 	} else if errObj, ok := err.(awserr.Error); ok && errObj.Code() == ssm.ErrCodeParameterNotFound {
 		return nil, nil, false, nil
 	}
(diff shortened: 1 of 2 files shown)


taylorsilva:
You may not be able to see the logs from the failed tests but they're all under the SSM test suite. You should be able to replicate the failures by running: `go test atc/creds/ssm/`

ConnorWidtfeldt:
> You may not be able to see the logs from the failed tests but they're all under the SSM test suite. You should be able to replicate the failures by running: `go test atc/creds/ssm/`

Is there anything special I need to do to see the output of the unit tests on ci.concourse-ci.org? I'll run them locally too, but I figured I would check in case it was an easy add.

taylorsilva:
Add yourself as a contributor in this repo: https://github.com/concourse/governance

taylorsilva:
The downgrade test is failing. That's in `./integration/ops`.
The integration tests are slower. They bring up concourse locally using docker compose to test stuff at a higher level.

ConnorWidtfeldt:
I should probably try running all the tests locally to see what's up. I'm surprised to see it run slower, it should only process a parameter if it's not a raw value, which is the only thing I can think of that would cause slowdowns outside of the secondary request to AWS to get the tags (these unit tests have no ways of testing for network latency, right?) 

Glad I started with a simpler PR for my first contribution to Concourse. I'm learning a ton about the setup.

atc/creds/ssm/ssm.go
```diff
@@ -99,5 +112,63 @@ func (s *Ssm) getParameterByPath(path string) (interface{}, *time.Time, bool, er
...
+		ResourceType: aws.String(ssm.ResourceTypeForTaggingParameter),
+		ResourceId:   id,
+	})
+
+	if err == nil {
+		for i := range tags.TagList {
+			tag := strings.ToLower(*tags.TagList[i].Key)
+			if tag == "format" {
+				format = strings.ToLower(*tags.TagList[i].Value)
+			}
```
taylorsilva, line 140: Once you enter this case you can `break` out of the `for` loop

atc/creds/ssm/ssm.go
```diff
@@ -99,5 +112,63 @@ func (s *Ssm) getParameterByPath(path string) (interface{}, *time.Time, bool, er
...
+		}
+	case "yml", "yaml":
+		err := yaml.Unmarshal([]byte(input), &value)
+		if err != nil {
+			return nil, err
+		}
+	case "trim":
+		value = strings.TrimSpace(string(input))
+	case "raw":
+		value = string(input)
```
taylorsilva, line 168:
```suggestion
		value = input
```
Redundant type casting

atc/creds/ssm/ssm.go
```diff
@@ -99,5 +112,63 @@ func (s *Ssm) getParameterByPath(path string) (interface{}, *time.Time, bool, er
...
+	default:
+		return nil, fmt.Errorf("unknown format %s, should never happen, ", format)
```
taylorsilva, line 170:
```suggestion
		return nil, fmt.Errorf("unknown format %w, should never happen, ", format)
```
Use `%w`: https://pkg.go.dev/fmt#Errorf

taylorsilva (changes requested):
Didn't look over tests yet but made some comments for small improvements. Mostly nits to make things more inline with the rest of the code base. This is looking great so far though! 🎉

atc/creds/ssm/ssm.go
```diff
@@ -99,5 +112,63 @@ func (s *Ssm) getParameterByPath(path string) (interface{}, *time.Time, bool, er
...
+			tag := strings.ToLower(*tags.TagList[i].Key)
+			if tag == "format" {
+				format = strings.ToLower(*tags.TagList[i].Value)
+			}
+		}
+
+		if !isValidFormat(format) {
+			return "nil", fmt.Errorf("invalid format %s", format)
+		}
+	}
```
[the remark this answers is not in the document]
ConnorWidtfeldt: Oh, that looks way better. I like to remove as much nesting as possible, and you make a good point about it correctly handling the error case.

PR #8453, opened 2022-06-27
Add a drift to component interval.

evanchaoli:

We have successfully upgraded one of our heaviest loaded cluster to 7.8.1.

During the upgrade, we initially encountered one problem where one ATC got overloaded, but other ATCs were idle. Which meant that workloads were not distributed evenly.

Some metrics show that:

image

One ATC's goroutes went to very high, nearing to 1 million, but other ATCs were idle. But after 14:00, I applied this patch, goroutines dropped significantly, and multiple ATCs had similar goroutine counts.

image

Before 14:00, each bar contained one 1 or 2 colors, meaning only 1 or 2 ATCs doing in-momory-checks; after 14:00, I applied this patch, each bar had 4 colors, each all ATCs were running in-memory-check.

The change of this PR is very simple, I just added a tiny drift time [-1 second .. 1 second] to component interval, so that other ATCs will get more chance to acquire the component's batch lock.

* [x] done

## Notes to reviewer

Without this PR, 7.8.1 couldn't work on our prod cluster, thus I submit to branch 7.8.x, and wish a release 7.8.2. I will port it to master branch once it is merged.

## Release Note

 - Enhancement of component scheduling so that workloads are distributed across ATCs more evenly.

### DIFF
--- a/atc/db/component.go
+++ b/atc/db/component.go
@@ -2,6 +2,7 @@ package db
 
 import (
 	"database/sql"
+	"math/rand"
 	"time"
 
 	sq "github.com/Masterminds/squirrel"
@@ -56,8 +57,17 @@ func (c *component) Reload() (bool, error) {
 	return true, nil
 }
 
+// IntervalElapsed adds a random tiny drift [-1, 1] seconds to interval, when
+// there are multiple ATCs, each ATC uses a slightly different interval, so
+// that shorter interval gets more chance to run. This mechanism helps distribute
+// component to across ATCs more evenly.
 func (c *component) IntervalElapsed() bool {
-	return time.Now().After(c.lastRan.Add(c.interval))
+	interval := c.interval
+	drift := time.Duration(rand.Int())%(2*time.Second) - time.Second
+	if interval+drift > 0 {
+		interval += drift
+	}
+	return time.Now().After(c.lastRan.Add(interval))
 }
 
 func (c *component) UpdateLastRan() error {
--- a/atc/db/component.go
+++ b/atc/db/component.go
@@ -31,6 +31,7 @@ type component struct {
 	interval time.Duration
 	lastRan  time.Time
 	paused   bool
+	rander   *rand.Rand
 
 	conn Conn
 }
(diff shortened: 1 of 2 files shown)


evanchaoli:
> I am more curious for root cause of the problem. Why the tick just happens in a way that only one ATC keep getting the component lock?

I did once debug on the problem by adding more debug messages.

Multiple ATCs anyway start at different time, though time differences are tiny.

Say ATC1 started at 1.000 second, and ATC started at 1.500 second.

ATC1 will get, say scanner-lock. And scanner may finish in 100ms, so ATC1 will try to acquire the lock again at 10.100 second.

ATC2 initially didn't get the lock, and it will try to acquire the lock again at 10.500 second.

So we can see that, ATC1 will get the lock again.

By adding a drift, ATC1 may acquire lock at 11.000 second, and ATC2 may acquire lock at 10.000 second, so ATC2 will get the lock.

atc/db/component.go
```diff
@@ -56,8 +57,17 @@ func (c *component) Reload() (bool, error) {
...
 }
 
+// IntervalElapsed adds a random tiny drift [-1, 1] seconds to interval, when
+// there are multiple ATCs, each ATC uses a slightly different interval, so
+// that shorter interval gets more chance to run. This mechanism helps distribute
+// component to across ATCs more evenly.
 func (c *component) IntervalElapsed() bool {
-	return time.Now().After(c.lastRan.Add(c.interval))
+	interval := c.interval
+	drift := time.Duration(rand.Int())%(2*time.Second) - time.Second
```
xtremerui, line 66: Maybe it should seed the rand to make it truely random?

evanchaoli: Without a seed, when each ATC starts, they will all start with the same random number. But in this case, I think that is ok, because we just want a random drift. Our prod cluster's metrics prove the current code is working fine.

But to honor your comment, I added a seed.

xtremerui (approved):

xtremerui:
LGTM! Please create a PR to master. Thx!

Issue #5194, opened 2020-02-18
containerd: oci spec misses seccomp


cirocosta:
### What challenge are you facing?

Containers created by the `containerd` garden backend miss OCI's [`LinuxSeccomp`] field being set. 

This means that the current set of syscalls that a process can execute is only being limitted by the set of capabilities that we grant (which is not enough) - [`seccomp`] bridges that gap. 

[`LinuxSeccomp`]: https://github.com/opencontainers/runtime-spec/blob/7cddf4f49ffc4012ca5f4ec5608c4c5514965bab/specs-go/config.go#L596-L602
[`seccomp`]: https://lwn.net/Articles/656307/

ps.: given that at this phase we're trying to mimic [`gdn`] as closely as possible,
I'd suggest us going straight with the same spec as they produce - this can either
be retrieved from the code, or by inspecting the OCI bundles from a running worker.

[`gdn`]: https://github.com/cloudfoundry/guardian

### What would make this better?

Have the OCI spec being filled with a set of allowed/disallowed syscalls via `seccomp`.

PR #8457, opened 2022-06-27
Allow AARCH64 syscalls for worker runtime

rdrgmnzs:
Signed-off-by: Rodrigo Menezes <brdude@gmail.com>

As AARCH64 becomes more popular allow the worker runtime to execute syscalls for it (ex. Mac M1 or AWS Graviton machines). 

Another option is to completely remove the seccomp architectural limitation as per https://github.com/opencontainers/runtime-spec/blob/72c1f0b44f7976960febe9e48e36c88d12ec9b72/specs-go/config.go#L635 by default only the native architecture of the kernel is permitted.

Relates to https://github.com/concourse/concourse/discussions/8451

### DIFF
--- a/worker/runtime/spec/seccomp.go
+++ b/worker/runtime/spec/seccomp.go
@@ -8,6 +8,7 @@ var seccomp = &specs.LinuxSeccomp{
 		specs.ArchX86_64,
 		specs.ArchX86,
 		specs.ArchX32,
+		specs.ArchAARCH64,
 	},
 	Syscalls: []specs.LinuxSyscall{
 		AllowSyscall("accept"),


xtremerui:
Thx for the PR! Is there a way to test this change? Unfortunately we don't have a M1 machine here.

rdrgmnzs:
@xtremerui the only ways I can things to test this is either having access to a Mac M1  / AWS Graviton instance or if install QEMU create a AARCH64 VM and try to run the docker-compose example ([using the ARM64 docker image](https://hub.docker.com/r/rdclda/concourse)) inside the emulated AARCH64 VM.

rdrgmnzs:
@xtremerui would the option of removing the seccomp CPU architecture limit work for you? That way you could simply test out the regular binary and ensure that by default seccomp is really limited just to the default hosts architecture like mentioned in https://github.com/opencontainers/runtime-spec/blob/72c1f0b44f7976960febe9e48e36c88d12ec9b72/specs-go/config.go#L635 . That would also have the effect of no having to maintain this for any possible future CPU architectures that come out (not that it happens often).

xtremerui:
This should be answered to the discussion but I would posted here too. The seccomp arch limit we have came from [here](https://github.com/concourse/concourse/pull/5228). 

Reading through the issue I feel if [guardien](https://github.com/cloudfoundry/guardian/blob/master/guardiancmd/seccomp_linux.go#L16-L18) doesn't add AARCH64 as a supporting arch, we probably can't either, especially without being properly tested. 

It is also mentioned in the original issue https://github.com/concourse/concourse/issues/5194 that the grant from Concourse itself is too broad so we should not remove the seccomp to use the default hosts architecture.

taylorsilva:
@rdrgmnzs I have an M1 that I can test this on. How would I test it? Do you have general steps I can follow? I can probably figure out the details if you have some guidelines.

PR #8461, opened 2022-06-30
Enhance Vault API client to auto retry upon rate limit.

evanchaoli:

After we upgraded to 7.8.1, Concourse became to faster, but the Vault became to a problem. We had to set a rate limit on Vault, thus some build may fail due to hit Vault rate limit.

It's very sad when a build has run 1 hour and fails at the last step because of hitting Vault rate limit. And every day there are a few such builds.

This PR enhanced Vault API client to auto retry upon rate limit error.

* [x] done

## Notes to reviewer

There is the other place that does retry https://github.com/concourse/concourse/blob/master/atc/creds/retryable_secrets.go. I decided to put auto retry in vault's api client level because retryable_secrets.go will lead to more retries which may make rate limit situation even worse. 

That is because, each secret will query multiple paths: pipeline path, team team, shared path, root path. Say if a secret suppose to be found under shared path, pipeline path and team path have been queried, and hit rate limit when query shared path. If retry is at vault api client level, it will only retry to query shared path; but if retry is in retryable_secrets.go, then it may retry query pipeline path and team path.

Also, Vault api client level already has retry logic and that retry cannot be turned off. Thus if a deployment only uses Vault as secret manager, then retryable_secrets can be turned off.

Once merged, I'll port it to master branch.

## Release Note

 - Enhanced Vault credential manager to auto retry when hitting Vault rate limit error. Vault started to support [rate limit](https://learn.hashicorp.com/tutorials/vault/resource-quotas) since 1.5. When setting rate limit on Vault, it's better to [enable](https://www.vaultproject.io/api-docs/system/quotas-config#enable_rate_limit_response_headers) rate limit HTTP response header by `vault write sys/quotas/config enable_rate_limit_response_headers=true`, so that the response header `Retry-After` may guide the Vault API client to retry after a reasonable duration.

### DIFF
--- a/atc/creds/vault/api_client.go
+++ b/atc/creds/vault/api_client.go
@@ -221,9 +221,9 @@ func (ac *APIClient) baseClient() (*vaultapi.Client, error) {
 	}
 
 	config.CheckRetry = func(ctx context.Context, resp *http.Response, err error) (bool, error) {
-		retryable, err := retryablehttp.DefaultRetryPolicy(ctx, resp, err)
-		if err != nil || retryable {
-			return retryable, err
+		retryable, retryErr := retryablehttp.DefaultRetryPolicy(ctx, resp, err)
+		if retryErr != nil || retryable {
+			return retryable, retryErr
 		}
 		// We also want to retry when hitting Vault rate limit.
 		if resp.StatusCode == 429 {
@@ -234,8 +234,8 @@ func (ac *APIClient) baseClient() (*vaultapi.Client, error) {
 
 	// Let's define retry backoff parameters explicitly
 	config.MinRetryWait = time.Millisecond * 1000
-	config.MaxRetryWait = time.Millisecond * 2000
-	config.MaxRetries = 3
+	config.MaxRetryWait = time.Millisecond * 3000
+	config.MaxRetries = 2
 
 	client, err := vaultapi.NewClient(config)
 	if err != nil {
--- a/atc/creds/vault/api_client.go
+++ b/atc/creds/vault/api_client.go
@@ -1,6 +1,7 @@
 package vault
 
 import (
+	"context"
 	"crypto/tls"
 	"fmt"
 	"io/ioutil"
@@ -11,6 +12,7 @@ import (
 	"time"
 
 	"code.cloudfoundry.org/lager"
+	"github.com/hashicorp/go-retryablehttp"
 	"github.com/hashicorp/go-rootcerts"
 	vaultapi "github.com/hashicorp/vault/api"
 )
@@ -216,6 +218,18 @@ func (ac *APIClient) baseClient() (*vaultapi.Client, error) {
 	err := ac.configureTLS(config.HttpClient.Transport.(*http.Transport).TLSClientConfig)
 	if err != nil {
 		return nil, err
+	}
+
+	config.CheckRetry = func(ctx context.Context, resp *http.Response, err error) (bool, error) {
+		retryable, err := retryablehttp.DefaultRetryPolicy(ctx, resp, err)
+		if err != nil || retryable {
+			return retryable, err
+		}
+		// We also want to retry when hitting Vault rate limit.
+		if resp.StatusCode == 429 {
+			return true, nil
+		}
+		return false, nil
 	}
 
 	client, err := vaultapi.NewClient(config)
--- a/atc/creds/vault/api_client.go
+++ b/atc/creds/vault/api_client.go
@@ -232,6 +232,11 @@ func (ac *APIClient) baseClient() (*vaultapi.Client, error) {
 		return false, nil
 	}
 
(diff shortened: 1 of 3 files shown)


atc/creds/vault/api_client.go
```diff
@@ -218,6 +220,23 @@ func (ac *APIClient) baseClient() (*vaultapi.Client, error) {
 		return nil, err
 	}
 
+	config.CheckRetry = func(ctx context.Context, resp *http.Response, err error) (bool, error) {
+		retryable, retryErr := retryablehttp.DefaultRetryPolicy(ctx, resp, err)
+		if retryErr != nil || retryable {
+			return retryable, retryErr
+		}
+		// We also want to retry when hitting Vault rate limit.
+		if resp.StatusCode == 429 {
```
xtremerui, line 229: I could only find https://www.vaultproject.io/api-docs/system/health where it says 

`[429](https://www.vaultproject.io/api-docs/system/health#429) if unsealed and standby`

is 429 a status code for rate limit only ? If not we might want to add some extra check to make sure the scope of this change.

xtremerui:
Would there be conflict when genenral retry and vault retry are both on?  How do them work together ? If not, we need to document it clearly.

atc/creds/vault/api_client.go
evanchaoli: Good catch. Per https://github.com/hashicorp/vault/blob/main/api/response.go#L32, we can further check request path.

evanchaoli:
> Would there be conflict when genenral retry and vault retry are both on? How do them work together ? If not, we need to document it clearly.

I guess there should be no conflict. It just leads to more retries on connection errors. For 429, was general retry won't retry on it, only Vault API layer will retry on 429.

But per our experience, when Vault went to down for a while, secrets caches expired on Concourse side, Concourse will invokes huge amount of request to Vault, then Vault may be slow to response, then Concourse may retry too many time, which might be caused by these two redundant retry levels. Thus I plan to disable the general retry on our clusters.

xtremerui:
> Thus I plan to disable the general retry on our clusters.

how do you disable this general retry?

evanchaoli:
> how do you disable this general retry?

https://github.com/concourse/concourse/blob/281d3e2ff5a072fcef267177ca686ba697f5f02c/atc/creds/retryable_secrets.go#L10-L13

By configuring `Attempts` to 0.

evanchaoli:
@xtremerui I have applied this patch to our prod cluster yesterday, so far it looks like reduced Vault rate limit error in Concourse side. But I realized that it's better to make the retry parameter configurable, because in the Vault side, rate limit settings include `interval` and `block-interval`. If `block-interval` is 10 seconds, as Vault API default retry interval is 1-3 seconds, retry would not work at all.

xtremerui:
since we already have the configure available for general creds manager, I would prefer to disable the retry on Vault client. Adding another set of config is confusing.

xtremerui:
Does that mean we need to update [documentation](https://concourse-ci.org/creds-retry-logic.html) ? If user is using Vault what is recommendation?

evanchaoli:
> Does that mean we need to update [documentation](https://concourse-ci.org/creds-retry-logic.html) ? If user is using Vault we'd recommend to set the attemps to 0 to disable it.

OK, I'll submit a PR to docs.

evanchaoli:
https://github.com/concourse/docs/pull/501 created.

xtremerui (approved):

xtremerui:
We will need to PR it back to master.

evanchaoli:
@xtremerui It's arguable if this is a bug fix. Because we bump Vault module, so that Vault API retry on rate limit error which is actually an enhancement, right? Without this PR, Vault API won't retry on rate limit error at all.

xtremerui:
@evanchaoli i am trying to update the release note for the draft release. I can't find a vault doc about `Retry-after` in header when `enable_rate_limit_response_headers` is true. Could you help? Thx.

evanchaoli:
@evanchaoli  https://learn.hashicorp.com/tutorials/vault/resource-quotas is only page I found `enable_rate_limit_response_headers`, but I cannot find any Vault doc about `Retry-After`. I got that by reading Vault code.

evanchaoli:
> Could you give the link to the source code? Thx

From Vault server side:

1. Here `Retry-After` is evaluated:
https://github.com/hashicorp/vault/blob/1915ca0334af51f8a6661d4da73622a1fd305e9c/vault/quotas/quotas_rate_limit.go#L292

2. Here `Retry-After` is inserted to response header:
https://github.com/hashicorp/vault/blob/1915ca0334af51f8a6661d4da73622a1fd305e9c/http/util.go#L77-L81

From Vault API client, here is `Retry-After is handled:
https://github.com/hashicorp/go-retryablehttp/blob/master/client.go#L512-L518

PR #8467, opened 2022-07-06
Don’t use ‘here’ as link text

quis:
Using ‘here’ or ‘click here’ as link text is bad because such links don’t mean anything out of context.

The Web Content Accessibility guidelines say:
> Whenever possible, provide link text that identifies the purpose of the link without needing additional context

– https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html

This commit moves the word ‘login’ inside the link, so it makes sense in and of itself.


xtremerui (approved):
LGTM!

PR #8485, opened 2022-07-18
remove "check build started/finished" metrics.

evanchaoli:

After several weeks of upgrading to 7.8, we found that the InfluxDB are quickly filling up. Then we identified the "check build started" and "check build finished" are two significant contributors.

Actually we don't monitor those two metrics at all. Because check builds are per resources, but real checks are per resource config scopes. Most of check builds just quickly terminate without actually running anything. 

* [x] done

## Release Note

Removed "check build started" and "check build finished" metrics. To monitor checks, use "check started" and "check finished" instead.


xtremerui (approved):
LGTM

PR #8486, opened 2022-07-19
Fixed a bug where invalidated worker resource caches are not GC-ed

### LINKED ISSUE #8473: Resource cache broken in 7.8.1?
## Summary

After upgrading to `7.8.1` (coming from `7.7.1`) every single step, be it a `task`, `get` or `put` will always download the required image to run instead of re-using a previously downloaded image. 

A `task` seems to always download the `image_resource`, and `get`/`put` to always download the corresponding `resource_type` image (obviously only applies to custom resource types).

For whatever reason it never re-uses an resource image from cache. In `7.7.1` it was often the case one would see the message: 
```
INFO: found existing resource cache
```
whenever a `get` was performed on an image of a `resource_type`. Same for `get` on a task's `image_resource`.

In `7.8.1` one of two things happen, either:
- The `get` for the `resource_type` image (or for the task's `image_resource`) immediately downloads the image
- The `get` for the `resource_type` image logs `INFO: waiting to acquire resource lock` and gets stuck there anywhere between a couple of seconds and several minutes, _then_ downloads the image

This happens even if, either within the same pipeline or across builds the `get` for the `resource_type` image lands on the same worker.

## Steps to reproduce

The only thing I could think of that might be relevant is that we are using global resources. 

_TODO: I'll try and setup a repo with docker compose, but maybe the bug report raises some flags for people more familiar with recent changes that might have caused this_ 

## Expected results

Once a `resource_type` image is downloaded, subsequent uses of that image (assuming no changes to the `resource_type` definition nor to the image sha256) re-use an image from cache.

## Actual results

Concourse is always re-downloading the same `resource_type` images and `image_resource` (from `tasks`) over and over again on every step.

## Additional context

Here's an example:
resource-lock

It's pulling the _same_ image that was pulled in a previous step and it's doing it on the _same worker_.

I can provide any relevant info and metrics if necessary.

## Triaging info

* Concourse version: `7.8.1`
* Browser (if applicable):
* Did this used to work? Yes, in `7.7.1` we never had this issue

evanchaoli:

closes #8473 

* [x] db schema change (up)
* [x] db schema change (down)
* [x] update `db.WorkerResourceCaches` 
* [x] update `InitResourceCache` and `InitStreamedResourceCache`
* [x] update unit tests
* [x] add a GC component to clean up invalidated worker resource caches 
* [x] run end-to-end tests   
* [x] done

In summary, the solution is:

* Add `where` clause to unique index `worker_resource_caches_uniq`, so that invalidated caches will not conflict to valid caches.
* Add `invalid_since` column to table `worker_resource_caches`, it will be auto set by a db trigger, so that GC component may delete invalidated caches based on time.
* Updated `db.WorkerResourceCaches`, so that for `worker name` + `resource cache id`, there can be only one valid cache, and multiple invalidated caches.
* If a build started before `invalid_since`, then the cache can still be used by the build; otherwise the build cannot use the cache.
* Add a GC component to delete invalidated worker resource caches based on `invalid_since` time.

## Release Note

* 


xtremerui (commented):
Just couple questions

atc/worker/gardenruntime/worker.go
```diff
@@ -715,6 +715,12 @@ func (worker *Worker) findOrStreamVolume(
 				return Volume{}, err
 			}
 
+			// TODO: find artifact again
+			// From `get` to where it artificat is used, it may take time, and during
+			// the period, original worker may have been pruned, but the artifact might
+			// have been streamed to other workers, thus find the artifact again can make
+			// sure the artifact exist.
```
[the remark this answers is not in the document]
evanchaoli: With my latest commit, this TODO is no longer needed. Even if original worker is pruned, worker resource cache will be marked as invalidated, but it can still be used by running builds.

But if the source worker is pruned, then the build may still fail, which is a general problem of volume streaming, I am not going to solve the problem in this PR.

atc/db/volume.go
```diff
@@ -462,25 +472,26 @@ func (volume *createdVolume) initializeResourceCache(tx Tx, resourceCache Resour
...
-		return false, err
+		return nil, false, err
 	}
 
 	if affected == 0 {
-		return false, ErrVolumeMissing
+		return nil, false, ErrVolumeMissing
 	}
 
+	volume.workerResourceCacheID = workerResourceCache.ID
```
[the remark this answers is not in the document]
evanchaoli: Before this PR, it uses `resource cache id` + `worker name` to look for `worker resource cache`, it supposed to have a single tuple found by the criteria. Now, that criteria cannot uniquely identify a worker resource cache, because it may be 1 valid cache and multiple invalidated caches. Thus I changed to use  `workerResourceCache.ID` to identify a specific cache.

atc/db/worker_resource_cache.go
```diff
@@ -79,38 +89,101 @@ func (workerResourceCache WorkerResourceCache) FindOrCreate(tx Tx, sourceWorkerB
...
+	}, true, nil
 }
 
-func (workerResourceCache WorkerResourceCache) find(runner sq.Runner) (*UsedWorkerResourceCache, bool, error) {
+// find should only return a valid worker resource cache if volumeShouldBeValidBefore is nil.
+// When volumeShouldBeValidBefore is not nil, if a cache's invalid_since is later than
+// volumeShouldBeValidBefore, then the cache is considered as valid. If there is a valid cache,
+// then the valid cache should be returned; if there are multiple invalid caches with
+// invalid_since later than volumeShouldBeValidBefore, the newest invalid_since one should be
+// returned.
```
xtremerui, line 132:
```suggestion
// find should return a valid worker resource cache if 
//   volumeShouldBeValidBefore is nil OR
//   a cache's invalid_since is later than volumeShouldBeValidBefore.
// If there are multiple invalid caches with invalid_since later than volumeShouldBeValidBefore, 
// the cache with newest invalid_since should be returned.
```

atc/db/worker_resource_caches_test.go
```diff
@@ -0,0 +1,240 @@
...
+			})
+
+			It("should create a cache", func() {
+				Expect(findErr).ToNot(HaveOccurred())
+				Expect(valid).To(BeTrue())
+				Expect(usedWorkerResourceCache).ToNot(BeNil())
+			})
+
+			Context("Create a worker resource cache again on worker0 again with it's own base resource type", func() {
+				var uwrc2 *db.UsedWorkerResourceCache
```
xtremerui, line 101:
```suggestion
				var uwrc *db.UsedWorkerResourceCache
```

atc/db/worker_resource_caches_test.go
```diff
@@ -0,0 +1,240 @@
...
+				It("should find a cache", func() {
+					Expect(findErr).ToNot(HaveOccurred())
+					Expect(valid).To(BeTrue())
+					Expect(uwrc2).ToNot(BeNil())
+					Expect(*uwrc2).To(Equal(*usedWorkerResourceCache))
+				})
+			})
+
+			Context("Create a worker resource cache again on worker0 with worker1's base resource type", func() {
+				var uwrc2 *db.UsedWorkerResourceCache
```
xtremerui, line 121:
```suggestion
				var uwrc *db.UsedWorkerResourceCache
```

atc/db/worker_resource_caches_test.go
```diff
@@ -0,0 +1,240 @@
...
+				It("should not create a cache, but find the existing cache", func() {
+					Expect(findErr).ToNot(HaveOccurred())
+					Expect(valid).To(BeFalse()) // valid is false as this is not the cache to create
+					Expect(uwrc2).ToNot(BeNil())
+					Expect(*uwrc2).To(Equal(*usedWorkerResourceCache))
+				})
+			})
+
+			Context("Create a worker resource cache on worker1 with worker0's base base resource type", func(){
+				var uwrc2 *db.UsedWorkerResourceCache
```
xtremerui, line 141:
```suggestion
				var uwrc *db.UsedWorkerResourceCache
```

xtremerui (commented):
I'd like to make the docs and tests more clear and readable.

xtremerui (approved):
LGTM!

evanchaoli:
@xtremerui Please hold merge this PR. We have tried to deployed this PR to our prod cluster and found it might cause ATC CPU usages to increase ~10%. I will enhance this PR to solve the CPU usage increasing problem.

atc/db/worker_resource_cache.go
```diff
@@ -124,65 +124,64 @@ func (workerResourceCache WorkerResourceCache) FindByID(runner sq.Runner, id int
...
+		Prefix(
+			`WITH candidates AS (
+                     SELECT id, worker_base_resource_type_id, invalid_since
+                       FROM worker_resource_caches
+                       WHERE resource_cache_id = $1 and worker_name = $2
+                 ),
+				 valid_caches AS (
+                     SELECT id, worker_base_resource_type_id, invalid_since, 1 as priority
+                       FROM candidates
+                       WHERE worker_base_resource_type_id is not null
```
deepakmohanakrishnan07, line 151: This is great, I have been following this change in the recent past to overcome this issue in our production cluster. 

I have a quick question on this psql query. IIUC, Does this WHERE clause check for `worker_base_resource_type_id != 0`? Because, I think the `worker_base_resource_type_id` is set to `0` when a worker is either pruned or the cache is invalidated.

evanchaoli: @deepakmohanakrishnan07 No, `worker_base_resource_type_id` is set to NULL. You may check with your db to verify that.

xtremerui:
@evanchaoli just a follow up on this issue as we are now in progress of releasing 7.9.

evanchaoli:
@xtremerui We are planning to apply this patch to stage this week and prod next week. So if 7.9 is planned for Nov., there should be enough time to prove this patch in our clusters.

atc/db/worker_resource_cache.go
xtremerui: @evanchaoli looks like the comment here is changed back to the old version

PR #8492, opened 2022-07-22
Support a way to skip implied get after put.

### LINKED ISSUE #3299: Add an option to skip implied get of put

I have realized the closed issue https://github.com/concourse/concourse/issues/837, as well as @vito 's post on stackoverflow https://stackoverflow.com/questions/38964299/why-does-concourse-get-a-resource-after-puting-it

As our users keep raising the questions to me, I'd like to reopen the discussion.

It looks to me that most of pipelines on our clusters end up with notification `put`s (Email, Slack, Pagerduty, and so on ...), those are pure `put` resources. They don't do anything with `get`, but each implied `get` consumes a container. Invoking a container takes seconds and ever longer, thus avoiding those implied `get` will not only save cpu/memory/disk of workers, but more importantly save overall pipeline execution time.

I understand @vito 's points that are explained in the stackoverflow post, and I respect the original Concourse design. I think adding an option to `put` step should not break anything, and it also meets the "explicit" principle of software design. By default, `put` will do implied `get`. But if one explicitly turns on the flag of a `put`, then skip implied `get` of the `put`.

Adding a flag to `put`, for example:

    - put: email
      skip_get: true # Add this option, default to false
      params:
          ...

This is a generic solution to allow pipeline developers to explicitly bypass implied `get`, which will save pipeline execution time as well as worker's cpu/memory/disk resources.

Yes, if we reach to an agreement of the design, we'd like to implement this.

evanchaoli:

closes #3299

* [x] done

## Notes to reviewer

There have been much such requests, and I fully understood why implied `get` is needed in order to keep a resource fresh.

But we may also honor reality. There are a lot of put only resources running in pipelines, mostly about notifications, like `put: email`, `put: slack`, etc. Implied `get` just run an empty container, and plug nested `image-check` + `image-get`. We have noticed that Intermittently implied `get` happened to be placed on a busy worker thus stuck for some time, but they didn't do anything, which hurts stability of build execution. 

Adding such an option to opt-out implied `get` will

* eliminate unnecessary failure points, thus improve build stability
* reduce build execute time
* save worker resources

The other reason I reopen the issue is that, Concourse proposed other approaches to solve the problem, but now those approaches seem to be no ETA anymore.

If you agree to merge this PR, I can update docs as well.

## Release Note

- Added `no_get` option to `put` step to skip implied `get`. For example:
    ```yaml
    - put: email
      no_get: true
      params:
        ...
    ```

### DIFF
--- a/atc/builds/planner.go
+++ b/atc/builds/planner.go
@@ -170,6 +170,11 @@ func (visitor *planVisitor) VisitPut(step *atc.PutStep) error {
 
 	plan.Put.TypeImage = visitor.resourceTypes.ImageForType(plan.ID, resource.Type, step.Tags, visitor.manuallyTriggered)
 
+	if step.NoGet {
+		visitor.plan = plan
+		return nil
+	}
+
 	dependentGetPlan := visitor.planFactory.NewPlan(atc.GetPlan{
 		Type:        resource.Type,
 		Name:        logicalName,
--- a/atc/builds/planner_test.go
+++ b/atc/builds/planner_test.go
@@ -1169,6 +1169,77 @@ var factoryTests = []PlannerTest{
 		}`,
 	},
 	{
+		Title: "put step with no_get",
+		Config: &atc.PutStep{
+			Name:      "some-name",
+			Resource:  "some-resource",
+			Params:    atc.Params{"some": "params"},
+			Tags:      atc.Tags{"tag-1", "tag-2"},
+			Inputs:    &atc.InputsConfig{All: true},
+			NoGet: true,
+		},
+		Inputs: []db.BuildInput{
+			{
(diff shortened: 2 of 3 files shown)


simonjohansson:
This is such a simple fix to an annoying thing when you don't need a get for put only resources... :)

Please merge!

simonjohansson:
Thx @xtremerui :)

evanchaoli:
@xtremerui I just rebased the code, hopefully tests would pass.

xtremerui (approved):
LGTM!

evanchaoli:
@xtremerui As you have approved this PR, please go ahead merge it. I am not going to apply a patch of this PR to our prod.

PR #8500, opened 2022-07-27
Fix a bad SQL for check gc.

evanchaoli:

closes #8454 

* [x] done

We have noticed deadlock problems in our prod db, then we found this slow query, in the meantime we noticed `check_build_events` had reached to 20 million tuples. 

Explain the SQL, we can see its cost is huge:

image

Then I worked out a new SQL:

image

On our db, the new SQL takes only 2 seconds to run, while the old SQL might take hours.

Now, I have disabled  component `collector_checks`, and use a 1 minute timer triggered pipeline to run the new SQL to do check GC. So far, it's running well.

## Release Note

* Optimized performance of check-build-events collector.

### DIFF
--- a/atc/db/check_lifecycle.go
+++ b/atc/db/check_lifecycle.go
@@ -61,16 +61,12 @@ func (cl *checkLifecycle) DeleteCompletedChecks(logger lager.Logger) error {
 	}
 
 	_, err2 := cl.conn.Exec(`
-      WITH resource_builds AS (
-        SELECT distinct(last_check_build_id) as build_id
-        FROM resource_config_scopes
-        WHERE last_check_build_id IS NOT NULL
-		  UNION ALL
-        SELECT distinct(in_memory_build_id) as build_id
-        FROM resources
-        WHERE in_memory_build_id IS NOT NULL
+      WITH expired_imb_ids AS (
+	    SELECT distinct(build_id) AS build_id FROM check_build_events cbe 
+		WHERE NOT EXISTS (SELECT 1 FROM resources WHERE in_memory_build_id = cbe.build_id) AND 
+		      NOT EXISTS (SELECT 1 FROM builds WHERE id = cbe.build_id) 
       )
-      DELETE FROM check_build_events WHERE build_id NOT IN (SELECT build_id FROM resource_builds)
+      DELETE FROM check_build_events cbe2 USING expired_imb_ids WHERE cbe2.build_id = expired_imb_ids.build_id;
     `)
 
 	if err1 != nil {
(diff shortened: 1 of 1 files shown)


xtremerui:
what's the bad of the old and the good for the new? Is it just a refactor or there is a behaviour change?

evanchaoli:
> what's the bad of the old and the good for the new? Is it just a refactor or there is a behaviour change?

This is a db performance optimization. I just added PR description.

evanchaoli:
@xtremerui Why this PR cannot trigger concourse-ci to run?

atc/db/check_lifecycle.go
```diff
@@ -61,16 +61,12 @@ func (cl *checkLifecycle) DeleteCompletedChecks(logger lager.Logger) error {
...
-        FROM resource_config_scopes
-        WHERE last_check_build_id IS NOT NULL
-		  UNION ALL
-        SELECT distinct(in_memory_build_id) as build_id
-        FROM resources
-        WHERE in_memory_build_id IS NOT NULL
+      WITH expired_imb_ids AS (
+	    SELECT distinct(build_id) AS build_id FROM check_build_events cbe 
+		WHERE NOT EXISTS (SELECT 1 FROM resources WHERE in_memory_build_id = cbe.build_id) AND 
+		      NOT EXISTS (SELECT 1 FROM builds WHERE id = cbe.build_id)
```
taylorsilva, line 67:
```suggestion
      WITH expired_imb_ids AS (
	    SELECT distinct(build_id) AS build_id
		FROM check_build_events cbe
		WHERE NOT EXISTS (SELECT 1 FROM resources WHERE in_memory_build_id = cbe.build_id)
			AND NOT EXISTS (SELECT 1 FROM builds WHERE id = cbe.build_id)

```
Just a nit for the formatting.

atc/db/check_lifecycle.go
```diff
@@ -61,16 +61,12 @@ func (cl *checkLifecycle) DeleteCompletedChecks(logger lager.Logger) error {
...
-        SELECT distinct(in_memory_build_id) as build_id
-        FROM resources
-        WHERE in_memory_build_id IS NOT NULL
+      WITH expired_imb_ids AS (
+	    SELECT distinct(build_id) AS build_id FROM check_build_events cbe 
+		WHERE NOT EXISTS (SELECT 1 FROM resources WHERE in_memory_build_id = cbe.build_id) AND 
+		      NOT EXISTS (SELECT 1 FROM builds WHERE id = cbe.build_id) 
       )
-      DELETE FROM check_build_events WHERE build_id NOT IN (SELECT build_id FROM resource_builds)
+      DELETE FROM check_build_events cbe2 USING expired_imb_ids WHERE cbe2.build_id = expired_imb_ids.build_id;
```
taylorsilva, line 69:
```suggestion
      DELETE FROM check_build_events cbe2 USING expired_imb_ids
	  WHERE cbe2.build_id = expired_imb_ids.build_id;
```
Similar nit

xtremerui (approved):
LGTM!

PR #8504, opened 2022-07-28
Reduce redundant job schedule

### LINKED ISSUE #8498: Scheduler slowness
## Summary

Looking at 

https://github.com/concourse/concourse/blob/4ca828100f9ab862b7d96d888e34e8f2ab3635cf/atc/db/resource_config_scope.go#L320-L323

when a new version is 
created, it will request schedule for all jobs that take the resource as an input. But should it only request scheduler for trigger resource? Meaning:

```golang
     Where(sq.Eq{
        "r.resource_config_scope_id": rcsID,
        "j.passed_job_id":            nil,
        "j.trigger:                   true, // add this
     }).
```

Our prod cluster is experiencing scheduler slowness, because a common git resource that is used in 4000 pipelines are generating a lot of new versions, which make scheduler always have 4000 jobs to schedule. But the git resource is not a trigger resource of any pipeline.

@clarafu This code was written by you. Can you please help me understand why `j.trigger: true` was not used?

## Triaging info

* Concourse version: I think all versions
* Browser (if applicable): NA
* Did this used to work?

SimonXming:

closes #8498 

* [ ] done

## Notes to reviewer

Our prod cluster is experiencing scheduler slowness, because a common git resource that is used in 4000 pipelines are generating a lot of new versions, which make scheduler always have 4000 jobs to schedule. But the git resource is not a trigger resource of any pipeline, so there are many redundant job scheduling for this case.

This PR is for reduce the redundant job scheduling when a new version detected for non-trigger resource without pending/running build.

## Release Note

* Avoid job schedule when a new version detected for non-trigger resource without pending/running build.

### DIFF
--- a/atc/db/build_test.go
+++ b/atc/db/build_test.go
@@ -1259,13 +1259,67 @@ var _ = Describe("Build", func() {
 				Expect(buildOutputs[0].Version).To(Equal(outputVersion))
 			})
 
-			Context("with a job in a separate team downstream of the same resource config", func() {
+			Context("with a job in a separate team downstream of the same non-trigger resource config", func() {
 				var otherScenario *dbtest.Scenario
 
 				var beforeTime, otherJobBeforeTime time.Time
 				var otherTeamBeforeTime, otherTeamOtherJobBeforeTime time.Time
 
 				BeforeEach(func() {
+					otherScenario = dbtest.Setup(
+						builder.WithTeam("other-team"),
+						builder.WithPipeline(pipelineConfig),
+						builder.WithResourceVersions("some-resource", atc.Version{"some": "version"}),
+						builder.WithResourceVersions("some-other-resource", atc.Version{"some": "other-version"}),
+					)
+
+					beforeTime = scenario.Job("some-job").ScheduleRequestedTime()
+					otherTeamBeforeTime = otherScenario.Job("some-job").ScheduleRequestedTime()
+
+					otherJobBeforeTime = scenario.Job("some-other-job").ScheduleRequestedTime()
+					otherTeamOtherJobBeforeTime = otherScenario.Job("some-other-job").ScheduleRequestedTime()
+				})
+
+				It("requests schedule on jobs which use the same config", func() {
+					Expect(scenario.Job("some-job").ScheduleRequestedTime()).To(BeTemporally(">", beforeTime))
+					Expect(otherScenario.Job("some-job").ScheduleRequestedTime()).To(BeTemporally("==", otherTeamBeforeTime))
+
+					Expect(scenario.Job("some-other-job").ScheduleRequestedTime()).To(BeTemporally("==", otherJobBeforeTime))
+					Expect(otherScenario.Job("some-other-job").ScheduleRequestedTime()).To(BeTemporally("==", otherTeamOtherJobBeforeTime))
+				})
+			})
+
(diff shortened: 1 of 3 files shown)


xtremerui:
@SimonXming you might want to check the concourse-ci/unit tests there are consistent failures.

SimonXming:
> @SimonXming you might want to check the concourse-ci/unit tests there are consistent failures.

Thanks @xtremerui, unit test updated.

atc/db/resource_config_scope.go
```diff
@@ -314,14 +314,26 @@ func incrementCheckOrder(tx Tx, rcsID int, version string) error {
...
+		From("job_inputs ji").
+		Join("resources r ON r.id = ji.resource_id").
+		Join("jobs j ON j.id = ji.job_id").
+		Where(sq.Or{
+			sq.Eq{
+				"r.resource_config_scope_id": rcsID,
+				"ji.passed_job_id":           nil,
+				"ji.trigger":                 true,
+			},
+			sq.And{
```
xtremerui, line 327: why do we need this block?

SimonXming: For non-trigger resource, normally new version shouldn't trigger job schedule. 

But if there is a pending build waiting for versions of the non-trigger resource in the `preparing build` phase, the trigger job schedule is necessary, otherwise it will stuck in `pending` state.

atc/db/resource_config_scope_test.go
```diff
@@ -138,7 +139,7 @@ var _ = Describe("Resource Config Scope", func() {
 				Expect(latestVR.CheckOrder()).To(Equal(2))
 			})
 
-			Context("when a new version is added", func() {
+			Context("when a new version is added for non-trigger resource", func() {
 				It("requests schedule on the jobs that use the resource", func() {
```
xtremerui, line 143:
```suggestion
				It("should not request schedule on the jobs that use the resource", func() {
```

atc/db/build_test.go
```diff
@@ -1259,13 +1259,67 @@ var _ = Describe("Build", func() {
...
+
+					beforeTime = scenario.Job("some-job").ScheduleRequestedTime()
+					otherTeamBeforeTime = otherScenario.Job("some-job").ScheduleRequestedTime()
+
+					otherJobBeforeTime = scenario.Job("some-other-job").ScheduleRequestedTime()
+					otherTeamOtherJobBeforeTime = otherScenario.Job("some-other-job").ScheduleRequestedTime()
+				})
+
+				It("requests schedule on jobs which use the same config", func() {
+					Expect(scenario.Job("some-job").ScheduleRequestedTime()).To(BeTemporally(">", beforeTime))
```
xtremerui, line 1284: why this one is scheduled? Is there somewhere a pending build is created for the job?

SimonXming: Yes, there is a `pending` job build for `some-job` in `scenario`.

* https://github.com/concourse/concourse/pull/8504/files#diff-f1d897e0cae526cb4f902a4aa326a976ebd9f07411cd6e7eb2896149c73168edL1210-L1215

xtremerui (approved):
LGTM!

SimonXming:
Hi @xtremerui, `concourse-ci/containerd-integration` is failed... could you help take a look? Thanks

PR #8511, opened 2022-08-02
Add renovate rule for k8s client-go lib

xtremerui:
this lib has a history of tags like v1.5.2 but its currently latest
is 0.24.3. I believe it was released as a whole with other kubebetes
packages but now it has its own versioning.

The problem is renovate will always bump it to v1.5.2 despite it is not
the latest.

Adding a rule to exclude those legacy versions.

Current tags https://github.com/kubernetes/client-go/tags

Example of legacy tag
```
⟩ curl -s https://proxy.golang.org/k8s.io/client-go/@v/v1.5.2.info
{"Version":"v1.5.2","Time":"2021-02-25T12:24:29Z"}⏎
```

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -16,8 +16,8 @@
       "groupSlug": "go"
     },
     {
-      "matchPackagePatterns": ["k8s.io/client-go"],
-      "matchCurrentVersion": "<1.0.0"
+      "matchPackageNames": ["k8s.io/client-go"],
+      "allowedVersions": "!/1\\.(4\\.0|5\\.0|5\\.1|5\\.2)$/"
     }
   ],
   "labels": ["dependencies", "misc", "release/undocumented"]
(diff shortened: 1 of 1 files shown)


markstokan (commented):
It seems like [retracted tags](https://go.dev/ref/mod#go-mod-file-retract) in Go Modules presents a problem with renovate.  There is an [open issue](https://github.com/renovatebot/renovate/issues/13012) to improve the logic around them.  The work hasn't been done but the issue includes a renovate.json workaround for the k8s.io/client-go dependency that improves on this PR.

```
{
  "packageRules": [
    {
      "matchPackageNames": ["k8s.io/client-go"],
      "allowedVersions": "!/1\\.(4\\.0|5\\.0|5\\.1|5\\.2)$/"
    }
  ]
}
```
* `matchPackageName` instead of matchPackagePattern
* `allowedVersions` exclude the specific retracted tags (1.4.0, 1.5.0, 1.5.1, 1.5.2)

PR #8525, opened 2022-08-16
Chunk build reaping by max database params

### LINKED ISSUE #8524: Build reaper broken with Postgres if more than 65535 builds to reap
## Summary

Noticed the following in our logs recently:

```json
{
    "level": "error",
    "source": "atc",
    "message": "atc.reaper.tick.build-reaper.failed-to-delete-build-events",
    "data": {
        "error": "pq: got 247760 parameters but PostgreSQL only supports 65535 parameters",
        ...
    },
    ...
}
```

## Steps to reproduce

Deploy with Postgres. Create a lot of builds. Configure build retention.

## Expected results

Builds to be removed.

## Actual results

The error above.

## Additional context

See also #5258 - it seems in general that there aren't checks in the database code for creating queries with an excess number of parameters.

## Triaging info

* Concourse version: v7.8.2

ghost:

closes #8524 

* [x] Adds chunking function
* [x] Uses it for build reaping

## Notes to reviewer

There may be more elegant solutions - this felt simple and an improvement over the current code.

It may be helpful to review all instances of dynamically generated queries for a similar class of bug.

### DIFF
--- a/atc/db/chunker.go
+++ b/atc/db/chunker.go
@@ -1,0 +1,18 @@
+package db
+
+const (
+	maxDatabaseParameterLimit = 65000 // Postgres will error if more than 65535 are used.
+)
+
+func chunkByMaxParams(a []int, f func([]int) error) error {
+	for {
+		if len(a) <= maxDatabaseParameterLimit {
+			return f(a)
+		}
+		err := f(a[:maxDatabaseParameterLimit])
+		if err != nil {
+			return err
+		}
+		a = a[maxDatabaseParameterLimit:]
+	}
+}
--- a/atc/db/pipeline.go
+++ b/atc/db/pipeline.go
@@ -989,6 +989,10 @@ func (p *pipeline) LoadDebugVersionsDB() (*atc.DebugVersionsDB, error) {
 }
 
 func (p *pipeline) DeleteBuildEventsByBuildIDs(buildIDs []int) error {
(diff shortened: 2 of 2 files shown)


ae-govau:
> Given that it might be multiple iterates on the params list, in current impl it will create multiple transactions that requires lots of DB connections. If `chunkByMaxParams` can take a `tx` then whoever calls it can then just do one tx.Commit() in the end.

I'm happy to make that change if desired, but I don't quite understand the reasoning. The chunks are processed sequentially (not in parallel), so while it's true there will be multiple transactions, I don't understand why that would require more DB connections, as each is completed before the next is opened?

xtremerui:
I think acquiring connection from DB connection pool and releasing it on each of the iteration introduces certain overheads comparing to using only one connection for the whole batch

ae-govau:
> I think acquiring connection from DB connection pool and releasing it on each of the iteration introduces certain overheads comparing to using only one connection for the whole batch

Although I remain a little sceptical that any meaningful overhead is added, and whether it's relevant for code like this which will infrequently called, I've gone ahead and refactored to do in a single transaction anyway.

On reflection though, I wonder if using `pq.Array()` is a better approach.

ae-govau:
#8528 is likely a better approach than this PR.

ae-govau:
Abandoning in favour of #8528

PR #8534, opened 2022-08-22
Fix: job should get scheduled when job inputs/resources get resolve error

### LINKED ISSUE #8498: Scheduler slowness

SimonXming:

closes #8498 

* [ ] done

## Notes to reviewer

With PR #8504, we noticed a bug for job scheduling in following scenario
- A pipeline contains a trigger resource `a` and a non-trigger resource `b`, no build get scheduled with following conditions
    - For trigger resource `a`, if the check succeed with new versions
    - For non-trigger resource `b` , if the check failed/errored(resolve error)
- When user fix the configuration of non-trigger resource `b` and check succeed with new versions, new build still **NOT** be scheduled.
  - But the **correct** behavior should be new build get scheduled.

The root cause is, when a job inputs/resources get succeed from previous failed/errored state, job schedule got ignored. This PR is for fix this bug, when job inputs/resources get failed/errored(`BuildInput.ResolveError` is not empty or `job.inputs_determined` is false), job schedule should be triggered for checking if job inputs got resolved(`job.inputs_determined` is true).

## Release Note

* Fix: Job should get scheduled, when job inputs get resolve error

### DIFF
--- a/atc/db/resource_config_scope.go
+++ b/atc/db/resource_config_scope.go
@@ -330,7 +330,10 @@ func requestScheduleForJobsUsingResourceConfigScope(tx Tx, rcsID int) error {
 					"ji.passed_job_id":           nil,
 					"ji.trigger":                 false,
 				},
-				sq.NotEq{"j.next_build_id": nil},
+				sq.Or{
+					sq.NotEq{"j.next_build_id": nil},
+					sq.Eq{"j.inputs_determined": false},
+				},
 			},
 		}).
 		OrderBy("ji.job_id DESC").
--- a/atc/db/build_test.go
+++ b/atc/db/build_test.go
@@ -1271,6 +1271,11 @@ var _ = Describe("Build", func() {
 						builder.WithPipeline(pipelineConfig),
 						builder.WithResourceVersions("some-resource", atc.Version{"some": "version"}),
 						builder.WithResourceVersions("some-other-resource", atc.Version{"some": "other-version"}),
+						builder.WithNextInputMapping("some-job", dbtest.JobInputs{
+							{
+								Name: "some-resource",
+							},
+						}),
 					)
 
 					beforeTime = scenario.Job("some-job").ScheduleRequestedTime()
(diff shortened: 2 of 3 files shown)


atc/db/resource_config_scope.go
```diff
@@ -330,7 +330,10 @@ func requestScheduleForJobsUsingResourceConfigScope(tx Tx, rcsID int) error {
 					"ji.passed_job_id":           nil,
 					"ji.trigger":                 false,
 				},
-				sq.NotEq{"j.next_build_id": nil},
+				sq.Or{
+					sq.NotEq{"j.next_build_id": nil},
+					sq.Eq{"j.inputs_determined": false},
```
xtremerui, line 335: where is it checking `BuildInput.ResolveError is not empty` as you mentioned?

SimonXming: 1. `job.SaveNextInputMapping` will update `jobs.inputs_determined` with variable `resolved`
    -  https://github.com/concourse/concourse/blob/master/atc/scheduler/scheduler.go#L51 
2. For variable `resolved`, it determined by existence of resolve error
    - https://github.com/concourse/concourse/blob/7cff39e2086fcfb8c6d23f8b72fe8dc113eebab7/atc/scheduler/algorithm/compute.go#L61

xtremerui (approved):
LGTM!

xtremerui:
@SimonXming sorry for this but could you fix the conflict? Once that is done we are good to go,

evanchaoli:
@xtremerui As #8504 is reverted, I believe this PR is no longer needed.

PR #8544, opened 2022-09-06
Change host to event_host tag for Datadog integration

pablokbs:
While trying to use the concourse integration with Datadog, I noticed that concourse is adding the `host` tag by default, and in the case of Kubernetes, it uses the pod name as the value.

This messes up with the inherited tags for the instances that the pod is running, as it's overriding the host tag and thus the new host value doesn't have any tag attached to it.

Unless I'm missing something, we don't need to have this tag added to the metric, as the datadog agent will add it anyway. So I'm renaming the `host` tag to `event_host` (just in case someone still wants to use it)

## Notes to reviewer

Sadly, I wasn't able to test as I couldn't for the life of me compile the binary + web files, I know this is not ideal, but if someone can build this binary for me, I'll test it and confirm that it fixes the issues I was seeing.

EDIT:

I've confirmed that adding manually a tag into Datadog to the host (pod name) adds the tags to all metrics attached to that host. So this should fix the issue.

### DIFF
--- a/atc/metric/emitter/dogstatsd.go
+++ b/atc/metric/emitter/dogstatsd.go
@@ -66,7 +66,7 @@ func (emitter *DogstatsdEmitter) Emit(logger lager.Logger, event metric.Event) {
 	name := specialChars.ReplaceAllString(strings.Replace(strings.ToLower(event.Name), " ", "_", -1), "")
 
 	tags := []string{
-		fmt.Sprintf("host:%s", event.Host),
+		fmt.Sprintf("event_host:%s", event.Host),
 	}
 
 	for k, v := range event.Attributes {


pablokbs:
```
➜  ~/git/concourse (248606d42) ✗ yarn install
yarn install v1.22.19
warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.39s.
➜  ~/git/concourse (248606d42) ✗ yarn build
yarn run v1.22.19
$ yarn run build-less && yarn run build-elm && yarn run build-js
$ lessc web/assets/css/main.less web/public/main.out.css && cleancss -o web/public/main.css web/public/main.out.css && rm web/public/main.out.css
$ cd web/elm && elm make --optimize --output ../public/elm.js src/Main.elm && uglifyjs ../public/elm.js --compress "pure_funcs=[F2,F3,F4,F5,F6,F7,F8,F9,A2,A3,A4,A5,A6,A7,A8,A9],pure_getters,keep_fargs=false,unsafe_comps,unsafe" | uglifyjs --mangle --output=../public/elm.min.js
Success!

    Main ───> ../public/elm.js

ERROR: invalid option --output=../public/elm.min.js
node:events:505
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at afterWriteDispatched (node:internal/stream_base_commons:160:15)
    at writeGeneric (node:internal/stream_base_commons:151:3)
    at Socket._writeGeneric (node:net:823:11)
    at Socket._write (node:net:835:8)
    at writeOrBuffer (node:internal/streams/writable:390:12)
    at _write (node:internal/streams/writable:331:10)
    at Socket.Writable.write (node:internal/streams/writable:335:10)
    at print (/Users/kbs/git/concourse/node_modules/uglify-js/bin/uglifyjs:603:20)
    at run (/Users/kbs/git/concourse/node_modules/uglify-js/bin/uglifyjs:474:9)
    at Object.<anonymous> (/Users/kbs/git/concourse/node_modules/uglify-js/bin/uglifyjs:298:5)
Emitted 'error' event on Socket instance at:
    at emitErrorNT (node:internal/streams/destroy:164:8)
    at emitErrorCloseNT (node:internal/streams/destroy:129:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -32,
  code: 'EPIPE',
  syscall: 'write'
}

Node.js v17.9.1
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

blyles:
Would this apply similar to a dd-agent-bosh? We have tags outputted to datadog.yml that don't get attached to the metrics that are pushed through and we only see 3 instead.  So we can't filter by say "region: us-west-2" in data dog dashboards.

We see host is generated and placed in datadog.yml but if i read your problem correctly its likely that the host between the emitter and what datadog places in are disconnected thus no custom tags are pushed along with the dogstatsd metrics.

xtremerui (approved):

PR #8565, opened 2022-09-20
[POC] Expose docker socket (if it exists) for privileged containers

jryan128:
If docker is installed on the worker and a task container is privileged, bind mount the docker socket into the task container.

This allows the container to run docker-compose or create containers (using docker installed inside the task container itself) on the worker without using docker-in-docker.

I tried using `nerdctl` instead of `docker` but that doesn't work remotely over a socket yet.

The only problem with this approach is garbage collection. That can be mitigated with a fairly complicated logic that only reaps orphan containers if their creation time is behind the start time of all currently running tasks.

Since docker and containerd do not share the networking concepts, I imagine the user will have to manage the networking themselves. I feel like that makes this approach not worth it.

This was asked for as a feature in [discord]( https://discord.com/channels/219899946617274369/604330794898554956/1010236295664107570 ) while I tackle the greater services RFC.

Signed-off-by: Jonathan Ryan <jonathan_ryan@comcast.com>

closes # 

* [x] done
* [ ] todo

## Release Note

* 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,3 +41,5 @@ FROM base
 # set up a volume so locally built web UI changes auto-propagate
 VOLUME /src
 ENV CONCOURSE_WEB_PUBLIC_DIR=/src/web/public
+
+RUN apt-get -y install docker.io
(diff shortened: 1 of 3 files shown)


jryan128:
After testing the networking it doesn't work very well. Trying to mix docker and containerd, and networks doesn't seem to work. May be best to abandon this approach.

PR #8577, opened 2022-09-28
New pipelines without build should be paused automatically with a configurable interval.

SimonXming:
## Notes to reviewer

For auto-pipeline-pauser, the current behaviour is for newly created pipeline if no build triggered in 24 hours, the pipeline will be paused automatically. This `24 hours` interval is hardcode value and not configurable. Some of our users found it annoying to keep unpause pipelines which recently created but haven't trigger a build yet.

This PR is for use existing params `pause-pipelines-after` to make this behaviour configurable.

## Release Note

* Using params `pause-pipelines-after`, so pipelines that were just set could be paused automatically with configurable interval.

### DIFF
--- a/atc/db/pipeline_pauser.go
+++ b/atc/db/pipeline_pauser.go
@@ -48,7 +48,7 @@ func (p *pipelinePauser) PausePipelines(ctx context.Context, daysSinceLastBuild 
 						)`,
 			strconv.Itoa(daysSinceLastBuild)+" day"),
 		// Covers edge case where pipelines that were just set could be paused automatically
-		// Only pauses the pipeline if it was last updated more than $daysSinceLastBuild days ago
+		// Only pauses the pipeline if it was last updated more than ${daysSinceLastBuild} days ago
 		sq.Expr(`p.last_updated < CURRENT_DATE - ?::INTERVAL`, strconv.Itoa(daysSinceLastBuild)+" day"),
 	}).RunWith(p.conn).Query()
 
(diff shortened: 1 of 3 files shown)


atc/db/pipeline_pauser.go
```diff
@@ -48,8 +48,8 @@ func (p *pipelinePauser) PausePipelines(ctx context.Context, daysSinceLastBuild
 						)`,
 			strconv.Itoa(daysSinceLastBuild)+" day"),
 		// Covers edge case where pipelines that were just set could be paused automatically
-		// Only pauses the pipeline if it was last updated more than 24hrs ago
-		sq.Expr(`p.last_updated < CURRENT_DATE - '1 day'::INTERVAL`),
+		// Only pauses the pipeline if it was last updated more than $daysSinceCreatedWithoutBuild days ago
+		sq.Expr(`p.last_updated < CURRENT_DATE - ?::INTERVAL`, strconv.Itoa(daysSinceCreatedWithoutBuild)+" day"),
```
evanchaoli, line 52: Auto pauser works in a way that when a pipeline has no build for N days, then pause it. So that, if a pipeline is newly created, why only give it 1 day to create the first build? Should we also give N days to new pipeline to create the first build? @xtremerui WDYT?

xtremerui: Seems reasonable to me to make them the same. I don't have the context of current value set.

xtremerui (approved):
LGTM!

PR #8584, opened 2022-10-06
atc/db/lock: preallocate memory

florianl:
**What this PR does / why we need it:**

Preallocate memory instead of enforcing an incremental growth. This will result in less work for the garbage collector.


evanchaoli:
LGTM

xtremerui (approved):
thx @evanchaoli

Issue #8585, opened 2022-10-06
ParseEvent return deprecated eventstream struct


gaelL:
## Summary

Using go-concourse to get Events from eventstream using `client.BuildEvents` returns deprecated event structure.

## investigation (dump of relevant discord messages)

I'm working with go-concourse, getting eventstream from it.
But during my tests, I got something strange.

On Concourse UI I can find such events `{"data":{"origin":{"id":"633c3ca0"},"time":1664895906,"succeeded":true},"event":"finish","version":"1.0","event_id":"4"}` (notice we have an origin field).

But the event I got from go-concourse does not contain this field.

Looking deeper in `go-concourse`, here is my results from `NextEvent()` https://github.com/concourse/concourse/blob/3c5e57c540609ed22aaa520f5f9121a4a09fae0d/go-concourse/concourse/eventstream/stream.go#L29

I did the following changes in the code in order to print some debug:

```diff
func (s *SSEEventStream) NextEvent() (atc.Event, error) {
        se, err := s.sseReader.Next()
        if err != nil {
                return nil, err
        }
+        fmt.Printf(">>>>%v\n", string(se.Data))
        switch se.Name {
        case "event":
                var message event.Message
                err = json.Unmarshal(se.Data, &message)
                if err != nil {
                        return nil, err
                }
+                spew.Dump(message)
                return message.Event, nil

        case "end":
                return nil, io.EOF

        default:
                return nil, fmt.Errorf("unknown event name: %s", se.Name)
        }
}

```

It produce the following output

```
>>>>{"data":{"origin":{"id":"633c3ca0"},"time":1664895906,"succeeded":true},"event":"finish","version":"1.0","event_id":"4"}
(event.Message) {
 Event: (event.FinishV10) {
  Time: (int64) 1664895906,
  ExitStatus: (int) 0
 }
}
```

As you can see the raw json event contain the `origin` field.
But Unmarshal seems to be done using `FinishV10` struct which does not contains it. I feel it should use the `Finish` struct instead:

  * `FinishV10` https://github.com/concourse/concourse/blob/f72eb75a59aad5205953273d3a37b4c1117dcc7f/atc/event/deprecated_events.go#L157
  * `Finish` https://github.com/concourse/concourse/blob/eaab65373b1cb35b62c859e69099c330452a46ca/atc/event/events.go#L247

actually what I have discovered is that 2 struct has been defined for the same event type. With the same Version

```
events.go
type Finish struct {
[...]

PR #8588, opened 2022-10-10
stream: Adding new NextEventRaw method

gaelL:

This method allow go-concourse users to get raw sse.Event from the eventstream.

This offers a handy way to fully get an event (payload and event_id), Instead of a decoded atc.Event.

This function also provides a workaround for https://github.com/concourse/concourse/issues/8585

## Notes to reviewer

This PR proposal offers more flexibility for go-concourse and eventstream users. It also gave a workaround for issue #8585


xtremerui (approved):
LGTM!

PR #8606, opened 2022-10-26
Change id of table resource_config_scopes to bigint

### LINKED ISSUE #8605: pq: integer out of range
## Summary

May users reported an error:

```
create resource config scope: find or create scope: pq: integer out of range
```

Then I found that table `resource_config_scopes`'s `id` has reached to max value of `int4`.

## Triaging info

* Concourse version: 7.8.1
* Browser (if applicable): NA
* Did this used to work? No

evanchaoli:

closes #8605

* [x] done

## Notes to reviewer

Looks like resource_config_scope's id has been handle with `int` type in Concourse code, thus only db schema change is needed.

## Release Note

Convert `id` column of the `resource_config_scopes` table and all tables referencing `resource_config_scope_id` to a `bigint`.


xtremerui (approved):
LGTM!

PR #8613, opened 2022-11-04
Performance optimize on accessor.

evanchaoli:

This is a performance optimization based on a `gprof` torch graph we captured from our prod cluster:

runway-atc-69d89c9fc4-8rxcv

We can see that `fmt.Sprint()` has consumed a significant amount of CPU times in function `(*access).roleOnTeam()`. Reading the code, I see two problems:

* `fmt.Sprint()` are duplicate called.
* Once a role is confirmed by userId/userName, then it doesn't need to check against groups.

This PR fixed the two problems.

* [x] done

## Release Note

Optimized performance of the login authentication process, which will benefit large deployments that has a lot teams and a lot of UI/fly accesses.

### DIFF
--- a/atc/api/accessor/accessor.go
+++ b/atc/api/accessor/accessor.go
@@ -93,51 +93,64 @@ func contains(arr []string, val string) bool {
 }
 
 func (a *access) rolesForTeam(auth atc.TeamAuth) []string {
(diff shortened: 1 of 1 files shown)


xtremerui (approved):
LGTM!

PR #8620, opened 2022-11-09
Fixed a bug of leaking resource cofig scope ids.

### LINKED ISSUE #8618: Leaking resource_config_scopes id
## Summary

When global resources are enabled, it will leak IDs of table `resource_config_scopes`. Our prod cluster's `resource_config_scopes` table had reached to `int4` limitation, which is very weird, as pipelines don't change frequently. In theory, only when a resource's config is changed, it should generate a new scope. We had to apply #8606 on our cluster.

## Steps to reproduce

This is easy to reproduce.

1. Start a local cluster with global resources enabled. (`docker-compose build` with `CONCOURSE_ENABLE_GLOBAL_RESOURCES: "true"` when `docker-compose up`.)

2. Set the following simple pipeline twice, so that the two resources will be the same scope:

```yaml
resources:
- name: data
  type: mock
  source:
    create_files:
      data.json: |
        [{"key": "a"}, {"key": "b"}]

jobs:
- name: something
  public: true
  plan:
    - get: data
```

Don't remember to unpause them after set-pipeline.

3. Wait for a few check runs.

4. Then check resource config scopes:

```sql
select id, resource_config_id, resource_id, last_check_build_id, last_check_start_time, last_check_end_time from resource_config_scopes order by id desc limit 10
```

and check current seq of resource config scopes' id:

```sql
select last_value from resource_config_scopes_id_seq;
```

## Expected results

1. There should be only 1 scope, and the scope id should never change.
2. `resource_config_scopes_id_seq` should never change.

## Actual results

For 1, actual result is as expected.

For 2, after each check round, `resource_config_scopes_id_seq` will increase by 2.

## Additional context

The root cause is at:

https://github.com/concourse/concourse/blob/5a9b0ca8800605dbc3da61aeea891538d65699fb/atc/db/resource_config.go#L128-L135

When global resources is enabled, `resource_id` column is set to `null` if a resource doesn't need unique version history (only `time` resource). But this query uses function parameter resource id in where clause, so that the query wouldn't return a row. Then the logic will jump to

https://github.com/concourse/concourse/blob/5a9b0ca8800605dbc3da61aeea891538d65699fb/atc/db/resource_config.go#L214-L224

The `insert` has `ON CONFLICT (resource_config_id) WHERE resource_id IS NULL`, so that it won't insert a new row, but `resource_config_scopes_id_seq` still gets increased. 

__NOTE:__ This bug doesn't only cause leak of scope id, it also leads to an unneeded `insert`. Because the function should have returned against `select`, but it actually runs an `insert on conflict` after `select`. Thus this bug has a big performance impact.

## Triaging info

* Concourse version: 7.x
* Browser (if applicable):
* Did this used to work?

### LINKED ISSUE #8619: After enabling global resources, old resoures won't be affected
## Summary

Say a cluster is not enabled global resources, there are two resources having the same config, each of them will have a different scope.

After enabling global resource, the two resources suppose to have the same scope, but actually they will still have the old scopes.

## Steps to reproduce

1. Start a local cluster with global resources enabled. (`docker-compose build` with `CONCOURSE_ENABLE_GLOBAL_RESOURCES: "false"` when `docker-compose up`.)

2. Set the following simple pipeline twice, so that the two resources will be the same scope:

```yaml
resources:
- name: data
  type: mock
  source:
    create_files:
      data.json: |
        [{"key": "a"}, {"key": "b"}]

jobs:
- name: something
  public: true
  plan:
    - get: data
```

Don't remember to unpause them after set-pipeline.

3. Wait for a few check runs.

4. Then check resource config scopes:

```sql
select id, resource_config_id, resource_id, last_check_build_id, last_check_start_time, last_check_end_time from resource_config_scopes order by id desc limit 10
```

5. Enable global resource and restart. (`docker-compose build` with `CONCOURSE_ENABLE_GLOBAL_RESOURCES: "true"` when `docker-compose up`.)

6. Wait a while for resource checks to run.

7. Then check resource config scopes using the same SQL as step 4.

## Expected results

At step 4, there should be two scopes.

At step 7, there should be only 1 scope and `resource_id` column should be `null`.

## Actual results

At step 7, there are still two scopes same as step 4, which means global resources feature doesn't apply to existing resources.

If we destroy the two pipelines and reset them, there will be only 1 scope.

## Additional context

The root cause is at

https://github.com/concourse/concourse/blob/5a9b0ca8800605dbc3da61aeea891538d65699fb/atc/db/resource_config.go#L128-L135

When global resources is not enabled, scope's `resource_id` will be filled in a resource's ID.

After global resource is enabled, the `select` will still return the old scope row, so that it won't create a new scope as global scope.

## Triaging info

* Concourse version: 7.x
* Browser (if applicable):
* Did this used to work?

evanchaoli:

closes #8618
closes #8619

* [x] code change
* [x] add unit tests
* [x] run more tests
* [x] done

## Notes to reviewer

This fix will lead to a breaking change, when switching global-resources from off to on, all resource histories will be lost. I think that should not be a big issue. Because:

1. Global-resources should always be turned on, actually Concourse wants to turn on the feature by default.
2. Global-resources is not such a switch that would be switched on/off frequently.

## Release Note

Fixed bugs about resource config scopes:
* When global-resources is enabled, `resource_config_scopes` tables leaked IDs. A side effect of the bug is that unnecessary `insert` will be performed (see #8618 for details). So, this fix will fix the ID leaking problem, also improve performance.
* When global-resources is enabled, old resources wouldn't be affected. This fix will ensure old resources to switch to global scopes.

__BREAKING:__ when switching global-resources from off to on, all resource histories will be lost. Losing version history usually not a big problem, for example, if you change `source` of a resource, for example, you change password/private-key of a `git` resource, its version history will be lost. But depending on a resource's `check` behavior, versions may be regenerated.

If your deployment has turned ON global-resources, or you have stay with global-resources OFF, this "breaking" won't impact your deployment.

Note that, if your cluster has turned ON global-resources, and you plan to turn OFF. In that case, after turning OFF global-resources, each resource will have an unique version history, thus shared version history will be lost. The behavior is always there, this PR doesn't change the behavior.

If you upgrade this fix, and turn on global-resources, as described, version histories will be lost. If you turn off global-resources, then old version histories will come back.

### DIFF
--- a/atc/db/resource_config_test.go
+++ b/atc/db/resource_config_test.go
@@ -9,16 +9,66 @@ import (
 
 var _ = Describe("ResourceConfig", func() {
 	var resourceConfig db.ResourceConfig
+	var otherResource db.Resource
 
 	Context("when non-unique", func() {
 		BeforeEach(func() {
 			var err error
+
+			// Adding a resourceTypeConfig to create a scope. The reason behind is that,
+			// when table "resource_config_scopes" is empty (just created), currentResourceConfigScopesIdSeq()
+			// will return 1; after insert the first tuple, currentResourceConfigScopesIdSeq()
+			// will also return 1. To work around the problem, adding a fake scope.
+			resourceTypeConfig, err := resourceConfigFactory.FindOrCreateResourceConfig(
+				defaultWorkerResourceType.Type,
+				atc.Source{"some": "fake-source-type"},
+				nil,
+			)
+			Expect(err).ToNot(HaveOccurred())
+			_, err = resourceTypeConfig.FindOrCreateScope(nil)
+			Expect(err).ToNot(HaveOccurred())
+
 			resourceConfig, err = resourceConfigFactory.FindOrCreateResourceConfig(
 				defaultWorkerResourceType.Type,
 				atc.Source{"some": "source"},
 				nil,
 			)
 			Expect(err).ToNot(HaveOccurred())
+
+			// Add a resource with the same config.
+			otherPipeline, created, err := defaultTeam.SavePipeline(
+				atc.PipelineRef{Name: "other-pipeline-with-resources"},
+				atc.Config{
+					Resources: atc.ResourceConfigs{
+						{
+							Name:   "other-resource",
+							Type:   "some-base-resource-type",
+							Source: atc.Source{"some": "source"},
+						},
+					},
+					Jobs: atc.JobConfigs{
+						{
+							Name: "some-job",
+							PlanSequence: []atc.Step{
+								{
+									Config: &atc.GetStep{
+										Name: "other-resource",
+									},
+								},
+							},
+						},
+					},
+				},
+				0,
+				false,
+			)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(created).To(BeTrue())
+
+			var found bool
+			otherResource, found, err = otherPipeline.Resource("other-resource")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(found).To(BeTrue())
 		})
 
 		Describe("FindOrCreateScope", func() {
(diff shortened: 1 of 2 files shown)


xtremerui:
> * BREAKING: when switching global-resources from off to on, all resource histories will be lost.

If global resource is ON and one upgrades to a version with this change, would resource histories gone?

xtremerui:
I think tests should be added to make sure:

when global resource is ON, with existing resource config, it tries to create resource config scope twice for default resource then tries to create a resource config scoep for a new resource. Verify the new resource config scope id is not increased by 2.

evanchaoli:
> If global resource is turned ON in previous version of Concourse, and one upgrades to a version with this change, would resource histories gone for those old resources that have same resource config (or would they opt-in global resource)?

The breaking will only happen when switch global resources from OFF to ON. Thus if global resource is ON before upgrade to this PR, then the "breaking" thing won't apply at all.

evanchaoli:
> 2\. when global resource is ON, create 2 resources with same config. Then [turn global resource off](https://github.com/concourse/concourse/blob/7d92ba9c6487ac7977969cf509ccfd3090cb77eb/atc/db/resource_config_test.go#L60) and verify those 2 resources are now sharing the resource config.

After turn off global resources, do you mean "those 2 resource are now NOT sharing the resource config scope"?

atc/db/resource_config_test.go
```diff
@@ -93,10 +108,15 @@ var _ = Describe("ResourceConfig", func() {
...
 					Expect(createdScope.ResourceConfig().ID()).To(Equal(resourceConfig.ID()))
+					seqAfterCreate, err := currentResourceConfigScopesIdSeq()
+					Expect(err).ToNot(HaveOccurred())
 
 					foundScope, err := resourceConfig.FindOrCreateScope(nil)
 					Expect(err).ToNot(HaveOccurred())
 					Expect(foundScope.ID()).To(Equal(createdScope.ID()))
+					seqAfterFind, err := currentResourceConfigScopesIdSeq()
+					Expect(err).ToNot(HaveOccurred())
+					Expect(seqAfterCreate).To(Equal(seqAfterFind))
```
xtremerui, line 119: I would expect further testing on when you use a different resource config to create scope, its scope id is only increased by 1 since the second time you try to create the scope on line 114, it will get rejected (where the bug was happenning)

xtremerui:
> After turn off global resources, do you mean "those 2 resource are now NOT sharing the resource config scope"?

yes sorry it should be `sharing the same resource config scope` like you said.

atc/db/resource_config_test.go
evanchaoli: Added more tests. I incidentally committed new tests with `--amend`, thus not adding a new commit.

xtremerui (approved):
LGTM

PR #8706, opened 2023-03-13
Support "raw" encoding for volume streaming.

evanchaoli:

A volume streaming workflow is like:

```
source worker                     dest worker
tar -> gzip     -> ship volume -> unzip -> untar
```

_NOTE: there are two compression methods supported, gzip and zstd, but that doesn't impact the problem this PR is trying to resolve._

I have noticed that `gzip` is usually much slower than tar. On a worker of our prod cluster, a VM with 8 vCPU and 32GB memory, I did a test:

- tar a dir with ~900MB content took ~1s
- gzip the tar file took ~48s, and shrink the tar file to ~300MB
- scp the tgz file to the other worker took 1s
- unzip the tgz file took ~8s
- untar tar file took ~4s

The above steps simulates a p2p volume streaming, so that you can see, streaming such a 900MB volume would take 62s, and gzip and unzip contributed the major portion.

The other thing we noticed during our daily operation is that, when a worker raises an alert about high CPU usage, if we ssh into the worker, we usually see a lot `tar` and `gzip` processes.

Our clusters all turn "p2p volume streaming" on, and out network infra is KGB Ethernet, thus network bandwidth is cheaper than VM's CPU times.

This PR adds support of a new compression method "nozip" for volume streaming.

## Notes to reviewer

A `docker-compose` local test shows this PR dramatically improves volume streaming speed:

With changes of this PR, the image streaming took only 12 seconds:

nozip

After switching back to "gzip" mode, the same image streaming took 39 seconds:

gzip

* [x] Basic workflow
* [x] Update non-linux streamer
* [ ] Done

## Release Note

Add a new compression method `raw` to `CONCOURSE_STREAMING_ARTIFACTS_COMPRESSION`. The new method will cost more network bandwidth of workers but save a lot of worker's CPU times, and make volume streaming dramatically faster. The bigger volume under streaming the more dramatic improvement on streaming speed.


atc/atccmd/command.go
```diff
@@ -1172,6 +1172,8 @@ func (cmd *RunCommand) backendComponents(
 func (cmd *RunCommand) compression() compression.Compression {
 	if cmd.StreamingArtifactsCompression == "zstd" {
 		return compression.NewZstdCompression()
+	} else if cmd.StreamingArtifactsCompression == "nozip" {
```
analytically, line 1175: `nozip` is confusing IMHO, how about `uncompressed`? Or `raw`?

xtremerui: I prefer `raw` here. WDYT @evanchaoli ?

evanchaoli: I am ok.

xtremerui (approved):
LGTM!

PR #8709, opened 2023-03-16
Add a drift based number of goroutines to component scheduler.

### LINKED ISSUE #8710: Interface fake implementations auto-generate for unit test seems broken
## Summary
Interface fake implementations auto-generate for unit test seems broken for `atc/db/worker_factory.go`.

## Steps to reproduce
```
> cd concourse
> go generate ./...
...
/concourse/concourse/atc/db/worker_factory.go:141:29: cannot use rows (variable of type *sql.Rows) as scannable value in argument to scanWorker: wrong type for method Scan
atc/db/resource_cache.go:24: running "counterfeiter": exit status 1
```

## Expected results
`go generate` should run successfully.

## Actual results
`go generate` failed for `atc/db/worker_factory.go`.

## Triaging info

* Concourse version: current master branch
* Browser (if applicable): None
* Did this used to work?

evanchaoli:

Close #8710

Currently component scheduler has a drift between -1 second to 1 second:

https://github.com/concourse/concourse/blob/5d2f4d3dcb24dfe9738d4552ca39458d41ad7122/atc/db/component.go#L65-L76

But on our clusters, we still frequently see some ATCs are very busy while some ATCs are idle. Sometime an ATC may have 2000K goroutes while the other ATC has only 2K goroutines.

This PR changes the drift value to be based current goroutine count, so that an idle ATC gets more chance to run components.

* [x] initial code
* [x] make `DriftBase` configurable 
* [x] Add tests
* [x] done

## Notes to reviewer

I have applied this change to one of our production clusters. Look at the goroutine metrics:

image

Before applied this change, some ATC's Goroutine count climbed to very high, which led to various of weird issues. After we applied a patch of this change with threshold of 50000, once an ATC's goroutine count reached to 50000, goroutine count quickly went down, that because once reaching to threshold, the ATC got less possibility to take any component, while idle ATCs got more possibilities to run components.

## Release Note

Add a new ATC option `--num-goroutine-threshold` to specify a threshold of goroutine count. If set, when a ATC's goroutine count reaches to threshold, then it will get less possibility to run workloads than other ATCs that have less goroutines. This option will help distribute workloads across ATCs evenly.


atc/atccmd/command.go
```diff
@@ -73,6 +73,7 @@ import (
 	"golang.org/x/crypto/acme/autocert"
 	"golang.org/x/oauth2"
 	"golang.org/x/time/rate"
+	"gopkg.in/square/go-jose.v2/jwt"
```
xtremerui, line 76: This one is not maintained anymore. Could you switch to go-jose/go-jose please?

atc/db/component_test.go
```diff
@@ -33,31 +33,233 @@ var _ = Describe("Component", func() {
...
+				It("should drift 1 second", func() {
+					Expect(elapsed).To(BeFalse())
+
+					fakeCompClock.Increment(1 * time.Second)
+					elapsed = component.IntervalElapsed()
+					Expect(elapsed).To(BeTrue())
+				})
+			})
+
+			Context("when there are 100000 goroutine", func() {
```
xtremerui, line 173: Duplicated test.

atc/db/component.go
```diff
@@ -68,28 +85,57 @@ func (c *component) Reload() (bool, error) {
...
-
+	c.lastRan = lastRan
 	return nil
 }
 
+const maxDrift = 5 * time.Second
+
+// ComputeDrift computes a drift for components scheduler, the drift should help
+// distribute workloads more evenly across ATCs. When numGoroutineThreshold
+// is not set, drift will be a random value in range of [-1, 1] second. When
```
xtremerui, line 116: just trying to understand why it drift differently when num Goroutine set/unset.

evanchaoli: For random drift, all ATCs are in the equal places, thus it doesn't make sense to add a 5-second drift. For numGoroutine based drift, when a ATC has 20K goroutines, obviously it should get bigger drift than a ATC that has only 0.1K goroutines.

xtremerui (approved):
LGTM!

PR #8712, opened 2023-03-22
atc: Support more than one Vault path prefix simultaneously

davidkuster:

Support for Vault kv2 was added in https://github.com/concourse/concourse/pull/6115. That supports both kv1 and kv2, but as [mentioned in the docs](https://concourse-ci.org/vault-credential-manager.html#configuring-the-secrets-engine), only a single prefix path is supported:
> Concourse is currently limited to looking under a single path, meaning enabling only one secrets engine is supported: `kv`, or `kv_v2`.

This PR enables multiple prefix paths and then does the lookup of the secrets in each, for each [path template](https://concourse-ci.org/vault-credential-manager.html#vault-lookup-templates) configured.

The idea in our org is to enable a smooth transition of teams from kv1 -> kv2. Without this we'd need to do a big bang release of ensuring all teams have migrated their secrets to the new engine, then update the Concourse config to point at kv2. Instead with this change we can support both, and let someone else herd the cats. 🙂 

Note this would potentially double the number of APIs to Vault. I don't see that's a reason to not do it, but something to call out in the docs at a minimum.

* [x] support multiple Vault prefix paths
* [x] manual testing 
* [ ] unit or other tests
* [ ] update docs

## Notes to reviewer

Creating this as a draft to get feedback on the overall idea, and specifically around changing the datatype of the existing `PathPrefix` var. This seems like the shortest path but I believe technically it's a breaking change for the `/api/v1/info/creds` endpoint.

Example of current response:
```json
    "path_prefix": "/kv1",
```
Example after this change:
```json
    "path_prefix": [
        "/kv2",
        "/kv1"
    ],
```

I'm not sure how widely used this endpoint is. I believe it's only available to admins under the `main` team, so it seems a bit more plausible to sneak it in.

The alternative I'm thinking of is to create a parallel `--path-prefixes` config option, although presumably the code would have to only allow either that or `--path-prefix`, but not both.

Entirely possible there are other aspects I'm not thinking of, hence the ask for feedback. Thanks!

Btw, quick example of local testing:
```yaml
jobs:
- name: hiya
  plan:
  - task: simple-task
    config:
      platform: linux
      image_resource:
        type: registry-image
      run:
        path: echo
        args: ["Hello world!", ((kv2-test.hello)), ((kv1-test.hello))]
```
Where `kv2-test` secret only exists in the kv2 path, and `kv1-test` secret only exists in the kv1 path. This results in `Hello world! Hello from kv2! Hello from kv1!`

## Release Note

TBD

* 

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -481,7 +481,7 @@ and suite runner of choice for Go code. You'll need to install the `ginkgo` CLI
 to run the unit tests and `testflight`:
 
 ```sh
-$ go get github.com/onsi/ginkgo/ginkgo
+$ go install github.com/onsi/ginkgo/ginkgo
 ```
 
 We use [Counterfeiter](https://github.com/maxbrunsfeld/counterfeiter) to generate
--- a/atc/api/info_test.go
+++ b/atc/api/info_test.go
@@ -323,7 +323,7 @@ var _ = Describe("Pipelines API", func() {
 
 					credhubManager := &credhub.CredHubManager{
 						URL:        credServer.URL(),
-						PathPrefix: "some-prefix",
+						PathPrefix: []string{"some-prefix"},
 						TLS:        tls,
 						UAA:        uaa,
 						Client:     &credhub.LazyCredhub{},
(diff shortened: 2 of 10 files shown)


davidkuster:
FWIW I'm not sure if the unit tests on this actually failed or if it's pending as [build 87](https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/unit/builds/87). I see [build 86](https://ci.concourse-ci.org/teams/main/pipelines/concourse/jobs/unit/builds/86) has been going for about a week.

davidkuster:
Closing in favor of https://github.com/concourse/concourse/pull/8714

PR #8713, opened 2023-03-22
Hermetic for task container

xtremerui:
This PR implements a new field for task step `Hermetic: bool`. When set to true, the container that running the task will NOT have external network access, aka the container's outgoing traffic through worker host network gateway will be dropped.

Noted, only `containerd` runtime supports hermetic container. For other runtimes (houdini and guardian) it will be just no-op.

* [ ] update Concourse doc on concourse-ci.org

## Notes to reviewer

It is possible to implement hermetic container for `guardian` runtime. Based on `deny_networks` flag [available](https://github.com/cloudfoundry/garden/issues/34#issuecomment-97941187) on garden backend. We could deny networks access for all IPs when starting up garden backend, and whitelist those IPs (like the change in atc/worker in this PR) later per container creation. 

## Release Note
 - add `Hermetic: bool` to task step configuration. When set to true, the task container will be running without external network access. Only worker runtime `containerd` supports this feature. There will be a reminder as warning when setting a pipeline contains task step that sets `hermetic: true`.

### DIFF
--- a/atc/worker/gardenruntime/worker.go
+++ b/atc/worker/gardenruntime/worker.go
@@ -597,12 +597,16 @@ func (worker *Worker) getBindMounts(ctx context.Context, volumeMounts []runtime.
 	return bindMounts, nil
 }
 
+// All outgoing network traffic will be dropped during runtime
+// depends on if NetOutRul is set
 func (worker *Worker) getNetOut(hermetic bool) []garden.NetOutRule {
-	// All outgoing network traffic will be dropped during runtime
+	// set NetOutRule to nil so worker runtime knows nothing is allowed
+	// to reach outside
 	if hermetic {
 		return nil
 	}
 
+	// set NetOutRule to whitelist all range of IPs
 	return []garden.NetOutRule{
 		{
 			Networks: []garden.IPRange{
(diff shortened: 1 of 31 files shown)


atc/worker/gardenruntime/worker.go
```diff
@@ -217,6 +218,7 @@ func (worker *Worker) createGardenContainer(
 			Properties: garden.Properties{
 				userPropertyName: fetchedImage.Metadata.User,
 			},
+			NetOut: worker.getNetOut(containerSpec.Hermetic),
```
evanchaoli, line 221: Can we only set `NetOut` when `hermetic` is true? And do nothing when it's false?

atc/worker/gardenruntime/worker.go
```diff
@@ -594,6 +597,28 @@ func (worker *Worker) getBindMounts(ctx context.Context, volumeMounts []runtime.
 	return bindMounts, nil
 }
 
+// All outgoing network traffic will be dropped during runtime
+// depends on if NetOutRul is set
+func (worker *Worker) getNetOut(hermetic bool) []garden.NetOutRule {
```
evanchaoli, line 602: If you take my other comment, then this function is not needed.

atc/worker/gardenruntime/worker.go
xtremerui: Yes thx for the review.

PR #8714, opened 2023-03-23
add Vault path-prefixes config

davidkuster:

_Closing https://github.com/concourse/concourse/pull/8712 in favor of this approach. Copying the bulk of the description here because the intent is the same._

Support for Vault kv2 was added in https://github.com/concourse/concourse/pull/6115. That supports both kv1 and kv2, but as [mentioned in the docs](https://concourse-ci.org/vault-credential-manager.html#configuring-the-secrets-engine), only a single prefix path is supported:
> Concourse is currently limited to looking under a single path, meaning enabling only one secrets engine is supported: `kv`, or `kv_v2`.

This PR enables multiple prefix paths and then does the lookup of the secrets in each, for each [path template](https://concourse-ci.org/vault-credential-manager.html#vault-lookup-templates) configured.

The idea in our org is to enable a smooth transition of teams from kv1 -> kv2. Without this we'd need to do a big bang release of ensuring all teams have migrated their secrets to the new engine, then update the Concourse config to point at kv2. Instead with this change we can support both, and let someone else herd the cats. 🙂 

Note this would potentially double the number of API calls to Vault. I don't see that's a reason to not do it, but something to call out in the docs at a minimum.

* [x] support multiple Vault prefix paths
* [x] limited manual testing 
* [x] unit or other tests
* [ ] update docs

## Notes to reviewer

Believe the approach in #8712 isn't viable due to `var_sources`. Instead trying the creation of a separate `--path-prefixes` config option. I tested locally with the following config:
```yaml
      CONCOURSE_VAULT_PATH_PREFIX: ""
      CONCOURSE_VAULT_PATH_PREFIXES: "/kv2,/kv1"
```

Quick example of local testing:
```yaml
jobs:
- name: hiya
  plan:
  - task: simple-task
    config:
      platform: linux
      image_resource:
        type: registry-image
      run:
        path: echo
        args: ["Hello world!", ((kv2-test.hello)), ((kv1-test.hello))]
```
Where `kv2-test` secret only exists in the kv2 path, and `kv1-test` secret only exists in the kv1 path. This results in `Hello world! Hello from kv2! Hello from kv1!`

## Release Note

* Add support for multiple Vault path prefixes. This allows for simultaneous support of multiple secret engines, allowing teams to transition between Vault locations at their own pace.

### DIFF
--- a/atc/creds/vault/manager.go
+++ b/atc/creds/vault/manager.go
@@ -18,6 +18,7 @@ type VaultManager struct {
 	URL string `mapstructure:"url" long:"url" description:"Vault server address used to access secrets."`
 
 	PathPrefix      string        `mapstructure:"path_prefix" long:"path-prefix" default:"/concourse" description:"Path under which to namespace credential lookup."`
+	PathPrefixes    []string      `mapstructure:"path_prefixes" long:"path-prefixes" description:"Multiple paths under which to namespace credential lookup."`
 	LookupTemplates []string      `mapstructure:"lookup_templates" long:"lookup-templates" default:"/{{.Team}}/{{.Pipeline}}/{{.Secret}}" default:"/{{.Team}}/{{.Secret}}" description:"Path templates for credential lookup"`
 	SharedPath      string        `mapstructure:"shared_path" long:"shared-path" description:"Path under which to lookup shared credentials."`
 	Namespace       string        `mapstructure:"namespace" long:"namespace"   description:"Vault namespace to use for authentication and secret lookup."`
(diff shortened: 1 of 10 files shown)


atc/creds/vault/manager.go
```diff
@@ -141,14 +144,20 @@ func (manager VaultManager) Validate() error {
...
-	if manager.PathPrefix == "" {
+	if manager.PathPrefix == "" && len(manager.PathPrefixes) == 0 {
 		return fmt.Errorf("path prefix must be a non-empty string")
 	}
 
-	for i, tmpl := range manager.LookupTemplates {
-		name := fmt.Sprintf("lookup-template-%d", i)
-		if _, err := creds.BuildSecretTemplate(name, manager.PathPrefix+tmpl); err != nil {
-			return err
+	if manager.PathPrefix != "" && len(manager.PathPrefixes) > 0 {
```
xtremerui, line 151: Why it is not allow for Vault to set both, compared to Credhub?

taylorsilva:
@davidkuster It's been a while since anyone has touched this PR. Do you have interest in working on this? Totally fine if you've moved on. I imagine the cats may have been herded in other ways at your org :)

taylorsilva:
Just peeked at your profile and then your linkedin. I see you were at comcast and are now at Procore. I'm going to assume you don't care about this PR anymore and will close it. I'll happily re-open it if you ask, or feel free to re-submit.

PR #8736, opened 2023-04-28
Optimize db notify.

evanchaoli:
### Problem

When you open a running build on the Concourse UI, you can see build logs keep growing, new logs keep scrolling down. How that is archived? In short, the workflow is like:

 * When UI open a build, the API will call `build.Events()`, and `build.Events()` will return an `EventSource` object.
 * The `EventSource` will issue a db query `LISTEN build_event_<build-id>`, upon receiving a notification over the listen, it will query the build event table to fetch new logs and streams new logs to UI over API.
 * While a build is running, for each line of log newly generated, it will send a db notification `NOTIFY build_event_<build-id>`, so that all clients that are listening will get notified to fetch new logs.

This workflow works fine, but with db performance wastes:

 * It highly relies on the db operations of `NOTIFY build_event_<build-id>` and `LISTEN build_event_<build-id>`. On our deployments, we have noticed db queries of those two statements.
 * A build may generate a lot of lines of logs in a second, it's not necessary to execute a `NOTIFY build_event_<build-id>` for every log line.
 * When a running build is not being watched over UI or `fly`, it's no need to execute a `NOTIFY build_event_<build-id>` , because no one is listening.
 * If you are watching a completed build, as no `NOTIFY` will come, it doesn't need to `LISTEN`.

### Solution

This PR optimizes usages of `NOTIFY` and `LISTEN` in the following ways:

 * For completed builds, `build.Events()` should not `LISTEN`.
 * For running builds not being watched, as there is no `LISTEN`, don't execute `NOTIFY`.
 * For running builds being watched, cache `NOTIFY`s and only `NOTIFY` once every 500ms. Based on my tests, 500ms keeps UI as smooth as no `NOTIFY` cache.

### Status

* [x] functionality
* [x] add tests
* [x] clean up code
* [x] done

## Notes to reviewer

@xtremerui We tried to apply this patch to our biggest cluster for 15 minutes during last weekend maintenance of upgrading to 7.9.1 from 7.8.2.  

From 14:30 ~ 14:45, it ran 7.9.1; and from 14:50pm to 15:05 it ran with a patch of this PR. During the entire test period, we turned off scheduler and tracker, thus no build were running, there were only Lidar checks running.

From the DB metrics, we can see, TPS was ~18K without this patch, and reduced to ~10K with this patch:

image

With this PR, the DB's CPU usage was also lower than without this PR:

image

image

There were not much impact on the DB's memory usage:

image

## Release Note

Optimized the database notifications, which will reduce TPS/QPS in the database side. A new ATC option `--db-notification-bus-queue-size` is added, defaults to 10000. If the UI doesn't load logs of running builds in time, then consider to increase value of the option.

### DIFF
--- a/atc/db/notifications_bus.go
+++ b/atc/db/notifications_bus.go
@@ -51,10 +51,11 @@ type notificationsBus struct {
 
 func NewNotificationsBus(listener Listener, executor Executor) *notificationsBus {
 	bus := &notificationsBus{
-		listener:       listener,
-		executor:       executor,
-		notifications:  newNotificationsMap(),
-		notifyChan:     make(chan string, 1000000),
+		listener:      listener,
+		executor:      executor,
+		notifications: newNotificationsMap(),
+		// TODO: make the chan size configurable
+		notifyChan:     make(chan string, 100000),
 		notifyDoneChan: make(chan struct{}, 1),
 		notifyCache:    map[string]struct{}{},
 	}
@@ -67,6 +68,11 @@ func NewNotificationsBus(listener Listener, executor Executor) *notificationsBus
 }
 
 func (bus *notificationsBus) Close() error {
+	close(bus.notifyChan)
+	bus.notifyChan = nil
+	close(bus.notifyDoneChan)
+	bus.notifyDoneChan = nil
+	bus.notifyCache = nil
 	return bus.listener.Close()
 }
 
@@ -130,15 +136,13 @@ func (bus *notificationsBus) wait() {
 			bus.handleReconnect()
 		}
 	}
-	close(bus.notifyChan)
-	close(bus.notifyDoneChan)
 }
 
 func (bus *notificationsBus) cacheNotify() {
 	for {
 		channel, ok := <-bus.notifyChan
 		if !ok {
-			break
+			return
 		}
 
 		bus.notifyCacheLock.Lock()
@@ -154,7 +158,6 @@ func (bus *notificationsBus) cacheNotify() {
 func (bus *notificationsBus) asyncNotify() {
 	ticker := time.NewTicker(500 * time.Millisecond)
 	go func() {
-	async:
 		for {
 			select {
 			case <-ticker.C:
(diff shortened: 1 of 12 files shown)


xtremerui:
Not sure if you still have the screen shot available, but the CPU usage one is showing DB connection count.

evanchaoli:
@xtremerui I added a wrong chart. I just added CPU usage chart. You may still start to review this PR, but you can hold on merge. I plan to apply this patch to our prod cluster in two weeks.

atc/db/notifications_bus.go
```diff
@@ -52,14 +53,26 @@ type notificationsBus struct {
 	wg *sync.WaitGroup
 }
 
+var notificationBusQueueSize = 10000
+
+func SetNotificationBusQueueSize(size int) error {
+	if size <= 0 {
+		return nil
+	}
+	if size < 1000 || size > 1000000 {
```
xtremerui, line 62: what is the reasoning behind these two numbers? What will happen if the size is too small or too big?

I would like to know the gains and cost while setting the queue size with different number. If setting default to 10000 works for most cases, we should avoid adding this flag.

evanchaoli: The reason why I added this parameter is I don't know what value is proper, so that after we apply this patch to our prod, the queue size must be tunable. If queue size is too small, running builds will not be able to refreshed in time (note that, only impact running builds); if queue size is too big, it will consume a lot of ATC memories. So I set a wide range which allows us to tune the parameter with enough flexibility.

xtremerui:
Is it good to merge this PR now?

evanchaoli:
> Is it good to merge this PR now?

I think so. We have applied this patch to our cluster for 1 week, nothing abnormal reported. And I do see reduced TPS and QPS from the db metrics.

xtremerui:
@evanchaoli i am wondering if the memory and CPU usage of web node increased due to this change?

Spimtav (approved):
i'm approving these changes on behalf of @xtremerui, who's reviewed these changes in the past.

evanchaoli:
> i am wondering if the memory and CPU usage of web node increased due to this change?

No.

PR #8746, opened 2023-05-15
Fixed a race condition in component factory.

evanchaoli:

Fixed a race condition introduced by #8709.

* [x] done

## Release Note

Fixed a race condition introduced by #8709.


xtremerui (approved):
LGTM!

PR #8756, opened 2023-05-26
Added a maximum volume size that can be streamed

evanchaoli:

Our prod cluster experienced worker unstable issues, the cause is like:

1 A pipeline incidentally generated a very large volume (>130GB), which filled up the current step's worker's disk
2 The worker is auto retired by our worker manager as its disk usage is close to full
3 The next step is scheduled to the other worker, as the first worker is retired
4 The next step requires the large volume as input, so that the large volume is streamed to the worker, so that the other worker is filled up also

So, as we can see, a large volume may bring up multiple workers, and even worse, there is no way to find out the pipeline that generates such large volumes.

This feature adds a threshold for maximum size of volumes that can be streamed. So that, in above step 4, volume streaming should fail, so that prevents from filling up another worker, and when users report the issue to us, we know which pipeline is the bad boy.

UI will show the error:

Non-p2p streaming:
image

P2p streaming:
image

* [x] poc
* [x] functionality 
* [x] add tests
* [x] clean up code

## Release Note

Add a new ATC option `CONCOURSE_STREAMING_SIZE_LIMITATION" that restricts maximum size in MB of volumes can be streamed between workers. This is a mechanism to prevent rogue pipeline pipelines from hurting multiple workers.

### DIFF
--- a/atc/worker/streamer.go
+++ b/atc/worker/streamer.go
@@ -145,8 +145,8 @@ func (s Streamer) p2pStream(ctx context.Context, src runtime.P2PVolume, dst runt
 	if err != nil {
 		return fmt.Errorf("invalid stream-in-url: %w", err)
 	}
-	// If stream limit is set, append the limit to stream-in url
-	if s.limitInMB > 0.00000094 {
+	// If stream limit is set to greater than 1 byte, append the limit to stream-in url
+	if s.limitInMB > float64(1)/1024/1024 {
 		query := rawUrl.Query()
 		query.Add("limit", fmt.Sprintf("%f", s.limitInMB))
 		rawUrl.RawQuery = query.Encode()
--- a/atc/worker/streamer.go
+++ b/atc/worker/streamer.go
@@ -5,7 +5,7 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"os"
+	"net/url"
 	"time"
 
 	"code.cloudfoundry.org/lager/v3"
@@ -54,7 +54,6 @@ func (s Streamer) Stream(ctx context.Context, src runtime.Artifact, dst runtime.
 
 	err := s.stream(ctx, src, dst)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "EVAN:atc:%s\n", err)
 		return err
 	}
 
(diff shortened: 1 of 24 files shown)


atc/worker/streamer.go
```diff
@@ -136,6 +140,19 @@ func (s Streamer) p2pStream(ctx context.Context, src runtime.P2PVolume, dst runt
 		return err
 	}
 
+	// Verify stream-in url
+	rawUrl, err := url.Parse(streamInUrl)
+	if err != nil {
+		return fmt.Errorf("invalid stream-in-url: %w", err)
+	}
+	// If stream limit is set, append the limit to stream-in url
+	if s.limitInMB > 0.00000094 {
```
xtremerui, line 149: Is there a specific reason to use this number other than `0.0`?

worker/baggageclaim/api/volume_server.go
```diff
@@ -495,6 +506,13 @@ func (vs *VolumeServer) StreamIn(w http.ResponseWriter, req *http.Request) {
 			return
 		}
 
+		var errExceedStreamLimit volume.ErrExceedStreamLimit
+		if errors.As(err, &errExceedStreamLimit) {
+			hLog.Error("exceeded-stream-limit", err)
```
xtremerui, line 511: Any reason to use Error instead of Info? Seems like same kind of error like others here.

worker/baggageclaim/api/volume_server.go
```diff
@@ -541,6 +559,11 @@ func (vs *VolumeServer) StreamOut(w http.ResponseWriter, req *http.Request) {
 			return
 		}
 
+		var errExceedStreamLimit volume.ErrExceedStreamLimit
+		if errors.As(err, &errExceedStreamLimit) {
```
xtremerui, line 563: Why isn't it logging it and returning just like others?

worker/baggageclaim/client_test.go
```diff
@@ -229,7 +229,7 @@ var _ = Describe("Baggage Claim Client", func() {
 						ghttp.RespondWith(http.StatusNoContent, ""),
 					),
 				)
-				err := vol.StreamIn(context.TODO(), ".", baggageclaim.GzipEncoding, strings.NewReader("some tar content"))
+				err := vol.StreamIn(context.TODO(), ".", baggageclaim.GzipEncoding, 0, strings.NewReader("some tar content"))
```
xtremerui, line 232: We should verify the request query param `limit` here

atc/worker/streamer.go
evanchaoli: I just want the limit to be bigger than 1 byte. I just updated the code to look clearer.

worker/baggageclaim/api/volume_server.go
evanchaoli: This error check is not needed at all. I initially put the limitation check at StreamOut side, but then moved to StreamIn side. So the error check should be only done in StreamIn side.

worker/baggageclaim/api/volume_server.go
evanchaoli: Because I want to always log this error. Because this is a protection mechanism, it doesn't suppose to happen frequently, but we want to see all violations.

xtremerui (approved):
LGTM!

xtremerui:
We will need to update concourse chart and concourse bosh release with this new flag

evanchaoli:
> We will need to update concourse chart and concourse bosh release with this new flag

Yes, I will do that. Please go ahead merge this PR.

PR #8795, opened 2023-07-25
topgun/k8s: fix dropped test error

alrs:
This picks up a dropped test `err` variable in `topgun/k8s`.


taylorsilva (approved):
Thanks for the PR!

PR #8800, opened 2023-08-22
feat/cli: allow to extract the default seccomp profile

drahnr:

* Adds a subcommand to allow extracting the default seccomp filter set

## Notes to reviewer

The main use-case is to extend/modify/limit these based on requirements, but provide them as a reasonable baseline for custom modifications. Prime use case is performance counters.

## Release Note

* Add a cli sub-command to allow extracting the default seccomp filter set

### DIFF
--- a/cmd/concourse/generate_key.go
+++ b/cmd/concourse/generate_key.go
@@ -4,9 +4,13 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/json"
 	"encoding/pem"
 	"fmt"
+	"io/ioutil"
 	"os"
+
+	bespec "github.com/concourse/concourse/worker/runtime/spec"
 
 	"golang.org/x/crypto/ssh"
 )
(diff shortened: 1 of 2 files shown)


cmd/concourse/generate_key.go
```diff
@@ -74,3 +78,27 @@ func (cmd *GenerateKeyCommand) Execute(args []string) error {
 
 	return nil
 }
+
+type ExtractInternalConfigCommand struct {
+	FilePath string `short:"f"  long:"filename"  required:"true"  description:"File path where the key shall be created. When generating ssh keys, the public key will be stored in a file with the same name but with '.pub' appended."`
```
taylorsilva, line 83: The `description` here appears to not be updated. It talks about ssh keys but it looks like it's actually the destination you want the seccomp profile written to.

cmd/concourse/generate_key.go
```diff
@@ -74,3 +78,27 @@ func (cmd *GenerateKeyCommand) Execute(args []string) error {
 
 	return nil
 }
+
+type ExtractInternalConfigCommand struct {
```
taylorsilva, line 82: Feels weird having this sub-command in this file. Would make more sense to have it in its own file. wdyt?

cmd/concourse/generate_key.go
```diff
@@ -74,3 +78,27 @@ func (cmd *GenerateKeyCommand) Execute(args []string) error {
...
+	FilePath string `short:"f"  long:"filename"  required:"true"  description:"File path where the key shall be created. When generating ssh keys, the public key will be stored in a file with the same name but with '.pub' appended."`
+
+	Seccomp bool `long:"seccomp" required:"false"  description:"Extract the default builtin seccomp filter"`
+}
+
+func (cmd *ExtractInternalConfigCommand) Execute(args []string) error {
+	var dest = cmd.FilePath
+	if cmd.Seccomp {
+		seccompfilter := bespec.GetDefaultSeccompProfile()
```
taylorsilva, line 91: So the whole point of this command is to print the `seccomp` variable from https://github.com/concourse/concourse/blob/master/worker/runtime/spec/seccomp.go in json format?

cmd/concourse/generate_key.go
taylorsilva: Maybe change the `longname` to something else to, like `output-file`.

taylorsilva (changes requested):
this is a pretty low-stakes PR, but implementation is a bit hacky. Can see some bits were copied and the whole thing was quickly put together. If this PR gets cleaned up I don't mind merging it in.

cmd/concourse/generate_key.go
drahnr: Yes, because that's what then could be modified and re-used as baseline for custom trim down or extension based on use case and re-supplied to the containerd runner

taylorsilva:
Closing because this has gone stale. Feel free to re-submit if you find time to address the comments.

PR #8810, opened 2023-09-12
Bump imdario/mergo to v1.0.0

taylorsilva:

Bump imdario/mergo to v1.0.0

They changed their module path to their personal domain so automated bumping failed


xtremerui:
thx @taylorsilva ! Good to see you!

PR #8851, opened 2023-10-30
Update release issue template during 7.11.x

wayneadams:


taylorsilva:
@wayneadams are the process changes outlined in this PR true? Or is what's currently published still true?

I see the `merge-dev-into-master` is gone, so I'm guessing the changes proposed here are true?
image

PR #8856, opened 2023-11-21
`fly`: add `background` option to `execute` command

KoltesDigital:

This patch adds a new `-b/--background` option to `fly execute`. This option tells `fly` to create the build, and exit just before watching the logs. It also means that outputs can't be retrieved. Instead, the build ID is printed, and logs can then be followed with `fly watch`.

* [x] added feature with acceptance test
* [ ] updated documentation
* [ ] left trail in contributors

## Rationale

I self-host Gitea and Drone. The application I'm making also needs some worker pool to spawn short-lived processes, and I thought it would be cool to only have one generic job framework to do both. But Drone is not flexible enough for that; besides it has been bought and shut down. I don't want to use frameworks that would lock me in K8S.

So, let's give Concourse a try! CICD works fine, however my jobs are triggered out of nowhere so I can't rely on resource checks to launch them. Although... maybe a custom resource type listing pending jobs with `version: every` could have somewhat worked, but can't be parallelized... Anyway, I feel spawning one-off tasks fits more to the idea.

I've seen on some pages that ATC's REST API is private, and on some other pages it's somewhat private only because it hasn't been documented yet. Well, my app will communicate through the API for sure, regardless of its privacy! I can't rely on spawning a process and waiting for it, my app can be restarted anytime, and jobs should not depend of that. So I prefer to independently spawn a job, and retrieve its logs.

I thought this workflow could be useful to CLI users too. Build ID is printed so that spawner process can use it (works with Bash and PowerShell).

Note that I don't need to fetch outputs back (yet), but CLI users could benefit from the complete split workflow: `fly execute --background` + `fly watch` + to-be-done `fly outputs`.

## Notes to reviewer

I don't master `flyIn`. It would have been better to test afterwards that `fly watch <buildID>` would have yield the same outputs as the preceding test, but then test times out. Maybe it's expected that `watch` keeps running if it's spawned while the build is running too, but I vaguely recall that it exits as soon as the build is completed.

Besides I'm new to Go, so I tried to fit in the code style, but have no idea how idiomatic it is.

Alternative name could have been `-d/--daemon` (like `docker run` does), but it implies the process is long-lasting, while it shouldn't.

## Release Note

* `fly execute` gets new `-b/--background` option to create builds without watching them.

### DIFF
--- a/fly/commands/execute.go
+++ b/fly/commands/execute.go
@@ -24,6 +24,7 @@ import (
 )
 
 type ExecuteCommand struct {
+	Background     bool                               `short:"b" long:"background"                            description:"Create the build and exit, i.e. neither watch logs nor retrieve outputs."`
 	TaskConfig     atc.PathFlag                       `short:"c" long:"config" required:"true"                description:"The task config to execute"`
 	Privileged     bool                               `short:"p" long:"privileged"                            description:"Run the task with full privileges"`
 	IncludeIgnored bool                               `          long:"include-ignored"                       description:"Including .gitignored paths. Disregards .gitignore entries and uploads everything"`
@@ -85,6 +86,10 @@ func (command *ExecuteCommand) Execute(args []string) error {
 		return err
 	}
 
+	if len(outputs) > 0 && command.Background {
+		return fmt.Errorf("background execution cannot withstand outputs")
+	}
+
 	plan, err := executehelpers.CreateBuildPlan(
 		planFactory,
 		target,
@@ -128,6 +133,11 @@ func (command *ExecuteCommand) Execute(args []string) error {
 	}
 
 	fmt.Printf("executing build %d at %s\n", build.ID, clientURL.ResolveReference(buildURL))
+
+	if command.Background {
+		os.Exit(0)
+		return nil
+	}
 
 	terminate := make(chan os.Signal, 1)
 
--- a/testflight/flying_test.go
+++ b/testflight/flying_test.go
@@ -80,6 +80,11 @@ run:
 		Expect(execS).To(gbytes.Say("ARGS are SOME ARGS"))
 	})
 
+	It("works in background", func() {
+		execS := flyIn(tmp, "execute", "-c", "task.yml", "-i", "fixture=./fixture", "-i", "input-1=./input-1", "-i", "input-2=./input-2", "--background", "--", "SOME", "ARGS",)
+		Eventually(execS).Should(gbytes.Say("executing build"))
+	})
+
 	Describe("hijacking", func() {
 		It("executes an interactive command in a running task's container", func() {
 			err := os.WriteFile(


taylorsilva:
> however my jobs are triggered out of nowhere so I can't rely on resource checks to launch them

This feels like a pipeline design problem. Are you still facing this pain? Could you share more details on how your pipelines are structured?

KoltesDigital:
@taylorsilva Well I meant, it's not out of nowhere, that's very deterministic: every time users click on the button. However, now that I have more experience with Concourse, I would model my need with a queue resource. I.e. when a user wants to run a job, instead of instructing Concourse to run it, the request is pushed into a queue, which will appear as a version to Concourse. A `serial` pipeline would pop and execute each request. This has the benefit of letting Concourse take care of limiting parallel jobs and other stuffs. The drawback, rather an extra effort, is that I'll have to monitor all build creations and match them back to the original requests. Regarding resource checks, a webhook trigger would force almost-immediate processing.

Still, I think that the flag could be useful. Usually log CLIs have a follow flag. In your case you're already following logs, but I don't understand why not-following wouldn't be allowed.

fly/commands/execute.go
```diff
@@ -127,6 +132,13 @@ func (command *ExecuteCommand) Execute(args []string) error {
...
 	}
 
+	if command.Background {
+		fmt.Printf("%d", build.ID)
+
+		os.Exit(0)
+		return nil
+	}
+
 	fmt.Printf("executing build %d at %s\n", build.ID, clientURL.ResolveReference(buildURL))
```
taylorsilva, line 142: The only thing I want to think about a bit is the output when you run `fly execute` with this flag. Right now it's printing out the build ID because, I assume, it gets you what you want for your use-case.

What if someone wants the full URL to the build though? By not printing the URL I think it becomes harder for users to discover that URL. The current output is not unparsable as well. `fly execute ... | awk '{print $3}'` would easily get you the build ID.

I think we should prefer to keep the current output as-is and just exit early. WDYT?

```suggestion
	fmt.Printf("executing build %d at %s\n", build.ID, clientURL.ResolveReference(buildURL))
	
	if command.Background {
		os.Exit(0)
		return nil
	}
```

taylorsilva (changes requested):
> Still, I think that the flag could be useful. Usually log CLIs have a follow flag. In your case you're already following logs, but I don't understand why not-following wouldn't be allowed.

Agreed, I can't think of a reason to not have this feature and am fine with merging it in. See below for feedback about the command output with this flag.

fly/commands/execute.go
KoltesDigital: Agreed, it was really specific 😄 I've changed the test accordingly.

Though someday, I suggest adding an arg like `--output=json` to all commands to dump multiple data from which parent processes can pick. Even in the tests, you rely on a regex to get the build id back. But that may be a big change that you'd better handle.

taylorsilva: Thanks for the quick turnaround! And totally agree, I had the same thought about having an `--output` for all commands. `fly` itself probably deserves a v2 of or at least some major internal re-working.

taylorsilva:
Looks like there's an issue with the watjs tests. Seeing it on `master` too. We'll have to look into that to unblock.

taylorsilva (approved):

taylorsilva:
Thanks for the PR @KoltesDigital!

PR #8895, opened 2024-01-23
Rebase master 7.11.1

xtremerui:
rebasing master for releasing v7.11.1


atc/creds/ssm/ssm_test.go
```diff
@@ -164,6 +164,19 @@ var _ = Describe("Ssm", func() {
 			Expect(err).To(BeNil())
 		})
 
+		It("should get shared parameter if exists", func() {
```
Spimtav, line 167: do you think there's merit to adding a test for the opposite case, ie: the shared parameter doesn't exist?

Spimtav (changes requested):
Everything seemed straightforward to my naive, contextless eyes.  Just a couple of small feature requests / replacements and package improvements, that i could see.  i don't really have anything to add design or implementation wise.  

The only thing i'm curious about is if it's worth testing the SharedPath feature a bit more rigorously.  For example, one test for the negative case (ie: SharedPath is empty), and maybe some corner cases too just in case (ie: the SharedPath exists but can't be found when looked up, the SharedPath exists but there's a network/read/whatever error on lookup, etc.).

Just an idea though, curious if you feel it would be necessary.

atc/creds/ssm/ssm_test.go
Spimtav: Rui and i discussed this synchronously, and i wrote up the summary here: https://github.com/concourse/concourse/issues/8897

Gonna close this review now and we'll figure out the tests later.

Spimtav (approved):

PR #9010, opened 2024-10-10
Split go-concourse client tests

aliculPix4D:
Split go-concourse client tests

* Connection client is deprecated. It is important to keep the test for the `HTTPAgent` separate and clean. If tests are mixed as before this PR, it will be easy to remove the `HTTPAgent` test by accident at some later stage when connection client gets removed. This is especially true because there is only 1 test for `HTTPAgent`.

The test coverage percentage remained the same. 

**commit 1:** Connection fakes are unused. Let's remove them and never generate them
**commit 2:** Remove the HTTP agent test from the connection test suite
**commit 3:** Create go-concourse/concourse/internal/http_agent_test.go and add back the test removed in commit 2


marco-m-pix4d (approved):
@taylorsilva FYI these PRs by @aliculPix4D are a series of fixes/cleanups we did in our Concourse deployment.

xtremerui (approved):
LGTM!

marco-m-pix4d:
Thanks @xtremerui !

PR #9011, opened 2024-10-10
go-concourse:connection client prints response body to the end user

aliculPix4D:
go-concourse:connection client prints response body to the end user

* if policy checker (e.g. OPA) does not allow the action the reason is never shown to the end user. The end user saw only Forbidden without the message why it is forbidden. To see more details end user was required to run the command with `--verbose` or inspect the request in web UI (if possible e.g. by clicking on expose pipeline button).
* this PR improves the error message the end user sees by printing also the reason from the response body. It is very important that a user sees the message of why the action was forbidden e.g by policy checker. 

Almost every fly command (except e.g. set-pipeline) still uses the deprecated HTTP client called connection client, so this PR affects many commands.

**Example:** `fly expose-pipeline` (and many other commands) when policy checker (OPA) returned `200` and denied the action (correct JSON body).

What users saw before:

```
$ fly -t main expose-pipeline -p test
error: forbidden
```

What users see now?

```
$ fly -t main expose-pipeline -p test
error: forbidden: policy check failed: 
 * <actual deny reason from OPA policy>
```

### DIFF
--- a/go-concourse/concourse/internal/connection_test.go
+++ b/go-concourse/concourse/internal/connection_test.go
@@ -351,6 +351,31 @@ var _ = Describe("ATC Connection", func() {
 				})
 			})
 
+			Describe("403 response", func() {
+				BeforeEach(func() {
+					atcServer = ghttp.NewServer()
+					connection = NewConnection(atcServer.URL(), nil, tracing)
+					atcServer.AppendHandlers(
+						ghttp.CombineHandlers(
+							ghttp.VerifyRequest("DELETE", "/api/v1/teams/main/pipelines/foo"),
+							ghttp.RespondWith(http.StatusForbidden, "problem"),
+						),
+					)
+				})
+				It("returns back ForbiddenError", func() {
+					err := connection.Send(Request{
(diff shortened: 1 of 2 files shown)


go-concourse/concourse/internal/connection_test.go
```diff
@@ -381,6 +381,31 @@ var _ = Describe("ATC Connection", func() {
 				})
 			})
 
+			Describe("403 response", func() {
```
aliculPix4D, line 384: Not a big deal but this will require a little bit of care when merging in combination with: https://github.com/concourse/concourse/pull/9010

In other PR, I am removing the test just above that is very similar to the one I am adding here. The difference being: one uses `HTTPAgent` client and the other the `connection` client.

marco-m-pix4d (approved):

analytically (approved):

xtremerui:
if you could resolve the conflict we are good to go. Thank you!

aliculPix4D:
> if you could resolve the conflict we are good to go. Thank you!

Done. 

Thanks for taking care of these PRs...

PR #9013, opened 2024-10-10
Log policy checker error messages

aliculPix4D:
Log policy checker error messages.

* minor improvements to OPA agent error messages
* atc: log errors returned by policy checker agent

This will ensure that:
- we return a meaningful error message to the Concourse end user (`policy-checker: unreachable or misconfigured`)
- Concourse operator/maintainer has all the required information in logs and can configure automated alerting on ERROR logs...

Originally, we wanted to go a step further and also log the success and log the actual policy checker decision and reason but that turned out to be too complicated with the current code to be done in a clean way. To improve this situation, in the future we plan to open a new issue to start the discussion on the proposed changes first.

## Release Note

* Log detailed OPA error messages in web nodes logs and show a friendlier error message to the end user

### DIFF
--- a/atc/api/policychecker/handler.go
+++ b/atc/api/policychecker/handler.go
@@ -34,10 +34,15 @@ type policyCheckingHandler struct {
 func (h policyCheckingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	acc := accessor.GetAccessor(r)
 
+	logger := h.logger.Session("policy-checker")
 	result, err := h.policyChecker.Check(h.action, acc, r)
 	if err != nil {
 		w.WriteHeader(http.StatusBadRequest)
-		fmt.Fprintf(w, "policy check error: %s", err.Error())
+		userMsg := "policy-checker: unreachable or misconfigured"
+		// error should be logged and meaningful to the Concourse operator/maintainer
+		logger.Error("policy-checker", err, lager.Data{"user-msg": userMsg})
+		// actual error is not useful and can be confusing to Concourse user
+		fmt.Fprint(w, userMsg)
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


atc/api/policychecker/handler.go
```diff
@@ -34,10 +34,15 @@ type policyCheckingHandler struct {
 func (h policyCheckingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	acc := accessor.GetAccessor(r)
 
+	logger := h.logger.Session("policy-checker")
 	result, err := h.policyChecker.Check(h.action, acc, r)
 	if err != nil {
 		w.WriteHeader(http.StatusBadRequest)
-		fmt.Fprintf(w, "policy check error: %s", err.Error())
+		userMsg := "policy-checker: unreachable or misconfigured"
```
marco-m-pix4d, line 41: I remember that in the fork I mentioned method `SessionName()` of lager, hoping it would return a unique session ID. Since I do not see it here, does it mean that it returns instead "policy-checker" ?

aliculPix4D: Ah... I am almost 100% sure I tried to use it but there were no ID... but honestly now I am confused and don't remember exactly...

From the source code:
https://github.com/cloudfoundry/lager/blob/8d31fc126ae626bbf33d7a05e405a2f4be414345/logger.go#L56
there is no mention of ID.

marco-m-pix4d (approved):

atc/api/policychecker/handler.go
marco-m-pix4d: OK, too bad. Thanks.

taylorsilva:
Wow didn't realize how many PRs you guys had opened! I'll get around to them once 7.12.0 is out

taylorsilva (approved):
LGTM

Thanks for these small PRs. I really love seeing folks make small improvements like this.

aliculPix4D:
> Thanks for these small PRs. I really love seeing folks make small improvements like this.

Thanks for taking care of these PRs...

PR #9019, opened 2024-10-28
Allow for varying the Prometheus namespace, defaults to "concourse".

analytically:
Currently the prometheus namespace (or prefix) is hardcoded as "concourse". This PR makes this a config setting, that defaults to "concourse" so it's backwards compatible.


taylorsilva (changes requested):
Test needs to be updated to specify the Namespace now:

https://github.com/concourse/concourse/blob/67b05c396eb899f3ba0ba21ff159b0c8f4cbbec1/atc/metric/emitter/prometheus_test.go#L184-L187

PR #9059, opened 2025-01-24
Bump Dex and other dependencies

### LINKED ISSUE #9006: CF API v2 End of Life
## Summary

[[RFC] CF API v2 End of Life](https://github.com/cloudfoundry/community/pull/941) - No longer able to use the CF authentication connector soon.

## Additional context

Based on the provided [RFC](https://github.com/cloudfoundry/community/pull/941) and [this documentation](https://v2-apidocs.cloudfoundry.org/#users), soon we will not be able to use the CF authentication connector, if we don't switch to `v3` API.
The migration should be smooth and easy. Based on the details provided by the CF community, the new API should have backward compatibility. Issue spotted at least [here](https://github.com/concourse/dex/blob/master/connector/cloudfoundry/cloudfoundry.go#L352-L355) and [here](https://github.com/concourse/dex/blob/8337b8fa07291fb9efe61010e51220ece78a9694/connector/cloudfoundry/cloudfoundry.go#L97). So the [test cases](https://github.com/concourse/dex/blob/8337b8fa07291fb9efe61010e51220ece78a9694/connector/cloudfoundry/cloudfoundry_test.go) should be fixed as well

taylorsilva:

Closes #9006

* Update to the latest version of our dex fork which supports the CF v2 API
  * Dex started using Go's new structured logging (slog) so we needed to handle passing that in. Thankfully Lager added a lager.NewHandler to make it easy to migrate to slog.
* Switch to using our own fork of houdini: https://github.com/concourse/houdini
  * We're the only one's that care and use houdini so it makes sense for us to just own this dependency fully instead of pretending that we don't

## Release Note

* CF API v3 is now supported by our fork of Dex which will ensure CF Auth does not break when the CF v2 API is officially gone. Work done by @Kump3r @IvanChalukov in https://github.com/concourse/dex/pull/148
* Bumped our fork of Dex to [v2.41.1](https://github.com/dexidp/dex/releases/tag/v2.41.1) of upstream Dex


Kump3r:
Maybe a quick remark here. We switch from CF API `v2` to `v3`. Not from `v1`.

taylorsilva:
Thanks for the nit! Updated the release note.

PR #9069, opened 2025-02-18
Run svgo -r against all SVG's.

taylorsilva:

This is optimizing all SVG's so they don't take up as much space. We used https://svgo.dev/

## Release Note

* Ran all SVG's through CLI tool [svgo](https://svgo.dev/) to reduce the size of all SVG's


taylorsilva:
No actual code changes in this so merging it straight in.

PR #9081, opened 2025-02-24
fix the policy-checker-handler logger session name

aliculPix4D:
fix the policy-checker-handler logger session name

Very minor log name correction.

Currently in the logs one sees:
```
{"timestamp":"2024-10-09T10:05:31.196402292Z","level":"error","source":"atc","message":"atc.policy-checker.policy-checker","data":{"error":"OPA server: connecting: Post "[http://localhost:8181/v1/data/concourse/decision\](http://localhost:8181/v1/data/concourse/decision%5C)": dial tcp [::1]:8181: connect: connection refused","session":"134","user-msg":"policy-checker: unreachable or misconfigured"}}
```

note the msg: `atc.policy-checker.policy-checker`. The `atc.policy-checker` is already there. So the session name should just be `handler`.

With the proposed change we will see: `atc.policy-checker.handler`

### DIFF
--- a/atc/api/policychecker/handler.go
+++ b/atc/api/policychecker/handler.go
@@ -34,7 +34,7 @@ type policyCheckingHandler struct {
 func (h policyCheckingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	acc := accessor.GetAccessor(r)
 
-	logger := h.logger.Session("policy-checker")
+	logger := h.logger.Session("handler")
 	result, err := h.policyChecker.Check(h.action, acc, r)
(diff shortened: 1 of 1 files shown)


marco-m-pix4d (approved):
(is this going to upstream on purpose? Ah, it is the one you mentioned missing?)

aliculPix4D:
> (is this going to upstream on purpose? Ah, it is the one you mentioned missing?)

Yes, its on purpose... we had this merged for few months now, but I forgot to include it in the PRs to upstream...

taylorsilva (approved):

PR #9089, opened 2025-02-28
Replace min/max helpers with built-in min/max

Juneezee:

We can use the built-in `min` and `max` functions since Go 1.21.

Reference: https://go.dev/ref/spec#Min_and_max


taylorsilva (approved):
Thanks for the PR!

PR #9102, opened 2025-03-07
Fix `--team` flag in `order-pipelines` command 

### LINKED ISSUE #9099: fly --team flag is nonfunctional for `order-pipelines` command
## Summary

In https://github.com/concourse/concourse/issues/5215 the community came together to add the teams option flag to most cli commands. However, it appears the order-pipeline commands (along with others?) fail to respect the flag passed.

## Steps to reproduce

Make two teams "main" and "other" with pipelines in both. Issue the below command and observe the results.

## Expected results

`fly -t main order-pipelines --team=other-team` should order pipelines for `other-team`.

## Actual results

`fly -t main order-pipelines --team=other-team` orders pipelines for `main`.

## Additional context

Tracing the code it looks like https://github.com/concourse/concourse/blob/master/fly/commands/ordering_pipeline.go#L42 and https://github.com/concourse/concourse/blob/master/fly/commands/ordering_pipeline.go#L64-L69 appears to be only targeting pipelines from the current team, not others.

Could the fix be related to the fix that was done for `builds` when `--team` flag was enabled there? see https://github.com/concourse/concourse/issues/8722

## Triaging info

* Concourse version: 7.11.2
* Did this used to work? Unclear

IvanChalukov:
 
The order-pipelines command with the --team option does not seem to work as expected. If you attempt to order pipelines from a team different from the one you are logged into, the operation will fail.

closes #9099
related to item in #5215

* [x] implement fix
* [x] add tests

## Notes to reviewer
Sample scenario: If you are logged into the main team (which has two pipelines: sample1 and sample2) and attempt to order pipelines from other-team (which has test1 and test2), the command will fail with the following error:
```sh
> fly -t lc order-pipelines  --alphabetical --team other-team
failed to order pipelines: pipeline 'sample' not found
```
This suggests that the --team option is not correctly switching context to the specified team, causing the command to reference pipelines from the currently logged-in team instead.

## Release Note

* Fix `order-pipelines` command with `--team` Option

### DIFF
--- a/fly/commands/ordering_pipeline.go
+++ b/fly/commands/ordering_pipeline.go
@@ -36,10 +36,16 @@ func (command *OrderPipelinesCommand) Execute(args []string) error {
 		return err
 	}
 
+	var team concourse.Team
+	team, err = command.Team.LoadTeam(target)
+	if err != nil {
+		return err
+	}
+
 	var orderedNames []string
 	if command.Alphabetical {
 		seen := map[string]bool{}
-		ps, err := target.Team().ListPipelines()
+		ps, err := team.ListPipelines()
 		if err != nil {
 			return err
 		}
@@ -58,12 +64,6 @@ func (command *OrderPipelinesCommand) Execute(args []string) error {
 			}
 		}
 		orderedNames = command.Pipelines
-	}
-
-	var team concourse.Team
-	team, err = command.Team.LoadTeam(target)
-	if err != nil {
-		return err
 	}
 
 	err = team.OrderingPipelines(orderedNames)
--- a/fly/integration/error_handling_test.go
+++ b/fly/integration/error_handling_test.go
@@ -194,6 +194,8 @@ var _ = Describe("Fly CLI", func() {
 				exec.Command(flyPath, "-t", targetName, "destroy-pipeline", "-p", "pipeline", "--team", otherTeam)),
 			Entry("get-pipeline command returns an error",
 				exec.Command(flyPath, "-t", targetName, "get-pipeline", "-p", "pipeline", "--team", otherTeam)),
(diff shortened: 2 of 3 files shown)


Kump3r (approved):
Not sure how to trigger the checks, but in any case I tested both scenarios with the old fly and your changes. The issue seems to be correctly handled by your change, that the old `order-pipelines --alphabetical` fetches the pipelines for the team it is currently in. I also confirmed that the code won't some how regress and break the functionality with `order-pipelines --pipeline <some-pipeline-name> --pipeline <other-pipeline-name>`. Thanks for the PR!
Screenshots from testing
Screenshot 2025-03-08 at 13 12 24
Screenshot 2025-03-08 at 13 24 04

IvanChalukov:
Thanks for the validation @Kump3r. I am not sure how to trigger checks too.

taylorsilva (approved):
Thanks for the fix and the additional tests 🙏

taylorsilva:
The checks trigger automatically. There's a `reconfigure` pipeline here (which I just made public): https://ci.concourse-ci.org/teams/main/pipelines/reconfigure
That creates all the PR pipelines.

Each job may take a while to report in but they will eventually. I made a change to the PRs pipeline: https://github.com/concourse/ci/blob/master/pipelines/pr.yml
so all the testing jobs don't run in parallel. I did this because right now there is only one PR worker 😞 See the one `ci-pr-worker-*` below:
image

I've found if I have two or more PRs in-flight at the same time that the chance of tests flaking due to CPU stress greatly increases. I would add more PR workers but that's a Broadcom decision at this time.

IvanChalukov:
Thanks for the detailed explanation! I was wondering if I had missed something, but now I understand how it works.

PR #9108, opened 2025-03-11
feat: increase footer hide delay from 4 to 8 seconds

analytically:
Users reported the footer disappearing too quickly. The extended delay provides a better experience.

# Release Note

* Web footer timeout has been increased from 4 seconds to 8 seconds


taylorsilva (changes requested):
There's a yarn test that expects the footer to disappear after 6 seconds. Should be updated to wait 10s. Can replicate locally by running:

```
yarn install
yarn test
```

taylorsilva (approved):

PR #9111, opened 2025-03-13
Add ESC key shortcut to hide keyboard help

analytically:
- Modified handleKeyPressed function to detect Escape key presses
- Added conditional logic to hide help panel when ESC is pressed and help is visible
- Follows common UI patterns where ESC dismisses overlays rather than toggling them

# Release Note

* Can use the Escape key to close the Help menu in the Web UI


taylorsilva (approved):
Took a quick look to see if we have any tests for these shortcuts, but couldn't find any. LGTM.

taylorsilva:
Unit test timed out, re-running.

PR #9112, opened 2025-03-14
chore: improve README

analytically:


taylorsilva:
I've updated the README and took in a few of your suggestions from this PR. Thanks!

PR #9114, opened 2025-03-14
Use math/rand/v2

analytically:
This commit updates the rander.Int() implementation to use Go 1.22's math/rand/v2 package. The new rand package provides per-goroutine random generators, eliminating lock contention that existed with the previous global random source.


analytically:
LGTM

taylorsilva (approved):

PR #9115, opened 2025-03-14
Fix race condition in NotificationsBus

analytically:
This commit addresses a race condition in the `asyncNotify` method of the NotificationsBus. Previously, the lock was held while both reading the notification cache and sending notifications, which could lead to deadlocks or delays if the notification process blocked.

The fix keeps the critical section as small as possible by:
1. Creating a local copy of channels to notify while holding the lock
2. Clearing the cache while still under lock
3. Releasing the lock before performing the actual notifications

This approach reduces lock contention and creates a more resilient notification system while maintaining the original behavior.

### DIFF
--- a/atc/db/notifications_bus.go
+++ b/atc/db/notifications_bus.go
@@ -193,14 +193,23 @@ func (bus *notificationsBus) asyncNotify() {
 		for {
 			select {
 			case <-ticker.C:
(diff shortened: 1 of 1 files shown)


taylorsilva (approved):
LGTM

PR #9117, opened 2025-03-14
Support custom pipeline background image styling

analytically:
Add support for configurable filter setting for pipeline background images. This enhances the visual customization options for pipelines by allowing users to set:

- Custom filter effects (defaults to grayscale(100%) opacity(30%) as before)

# Release Note

* Add `background_filter` option for pipeline background images which takes in string of [CSS filters](https://developer.mozilla.org/en-US/docs/Web/CSS/filter). Defaults to the current filters `opacity(30%) grayscale(100%)`

### DIFF
--- a/atc/config.go
+++ b/atc/config.go
@@ -213,7 +213,8 @@ type Prototype struct {
 }
 
 type DisplayConfig struct {
-	BackgroundImage string `json:"background_image,omitempty"`
+	BackgroundImage  string `json:"background_image,omitempty"`
+	BackgroundFilter string `json:"background_filter,omitempty"`
 }
 
 type CheckEvery struct {
--- a/web/elm/src/Concourse.elm
+++ b/web/elm/src/Concourse.elm
@@ -1117,6 +1117,7 @@ type alias Pipeline =
     , teamName : TeamName
     , groups : List PipelineGroup
     , backgroundImage : Maybe String
+    , backgroundFilter : Maybe String
     }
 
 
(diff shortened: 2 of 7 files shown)


taylorsilva:
Just a note-to-self, we should document these: https://github.com/concourse/docs/blob/70e8fd6ada55064074bc0d98666ca0219fa8d637/lit/docs/pipelines.lit#L109-L135

taylorsilva (changes requested):
Very nice! I think we can actually remove the `background_opacity` field because `background_filter` can be used to combine multiple filters like `opacity()`: https://developer.mozilla.org/en-US/docs/Web/CSS/filter

I made this filter:
```yaml
background_filter: blur(5px)grayscale(100%)
```

and got:

image

Adding `opacity()`
```yaml
background_filter: blur(5px)grayscale(100%)opacity(50%)
```
image

`background_filter` is all we really need :)

analytically:
Currently opacity is set to 30%. If we'd just allow the filter to be set, what do we do with the default value? Not set it? People might expect it to be kept at 30% when just setting the filter?

analytically:
Having both opacity and filter is more flexible IMO. You can just set the image and opacity without thinking about filter. Or just image and filter, then an opacity of 30% is applied by default.

analytically:
Or, we remove the default for opacity into the filter.

taylorsilva (approved):

PR #9119, opened 2025-03-14
atc: allow identifiers to start with numbers

analytically:
This change updates the identifier validation to permit identifiers that start with
numbers (e.g., "1pipeline") while keeping other validation rules intact. All identifiers
must now start with either a lowercase letter or a number, and the error messages have
been updated accordingly.

# Release Note

* Identifiers (names of pipelines/jobs/resources/steps/etc.) can start with numbers. Previously they could only start with `a-z`, otherwise Concourse would warn you that the identifier is invalid. An identifier cannot be only numbers though. `543` is invalid, but `2-days` is valid.

### DIFF
--- a/atc/api/config_test.go
+++ b/atc/api/config_test.go
@@ -376,11 +376,11 @@ var _ = Describe("Config API", func() {
 								"warnings": [
 									{
 										"type": "invalid_identifier",
-										"message": "pipeline: '_pipeline' is not a valid identifier: must start with a lowercase letter"
+										"message": "pipeline: '_pipeline' is not a valid identifier: must start with a lowercase letter or a number"
 									},
 									{
 										"type": "invalid_identifier",
-										"message": "team: '_team' is not a valid identifier: must start with a lowercase letter"
+										"message": "team: '_team' is not a valid identifier: must start with a lowercase letter or a number"
 									}
 								]
 							}`))
--- a/atc/api/pipelines_test.go
+++ b/atc/api/pipelines_test.go
@@ -1727,7 +1727,7 @@ var _ = Describe("Pipelines API", func() {
 								"warnings": [
 									{
 										"type": "invalid_identifier",
-										"message": "pipeline: '_some-new-name' is not a valid identifier: must start with a lowercase letter"
+										"message": "pipeline: '_some-new-name' is not a valid identifier: must start with a lowercase letter or a number"
 									}
 								]
 							}`))
--- a/atc/api/teams_test.go
+++ b/atc/api/teams_test.go
@@ -361,7 +361,7 @@ var _ = Describe("Teams API", func() {
 									"warnings": [
 										{
 											"type": "invalid_identifier",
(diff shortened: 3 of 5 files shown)


taylorsilva (changes requested):
Why do we need to identify `time.Duration`-like strings? It feels like we're trying to resolve a very specific pain here instead of trying to determine if there's a larger ask/pain here.

Also, instead of rolling our own regex, why not use [`time.ParseDuration()`](https://pkg.go.dev/time#ParseDuration)?

After looking up where identifier gets used I'm guessing you have resources whose names are `time.Duration`-like strings?

Back to the larger ask, because I don't want this file to be committed full of exceptions for strings that look like X, then Y, then Z, etc.

**Do we want to allow identifiers that start with numbers?**
Having context for why we have our current limitation would be helpful to answer this question. Let's go explore the project history.

I looked through the commit history for this file, specifically looking at the lines where the regex are defined.
I looked at this commit https://github.com/concourse/concourse/commit/e6c2ede816c7425cd2b257cd612eb87268820471.
Then the PR it came from: https://github.com/concourse/concourse/pull/5864
Ah, that PR references an RFC: https://github.com/concourse/rfcs/pull/40
And the original motivation(s): https://github.com/concourse/concourse/issues/3985

So taking in all this context, it looks like identifier restrictions were introduced to ensure folks don't put weird characters in places that appear in URL's.

I'm not seeing any great reason to not allow identifiers to start with numbers, though it also seems like no one is really asking for that. That's always dangerous with yaml too because a bunch of numbers are going to be parsed as a number instead of a string.

```yaml
a_number: 123
a_string: "123"
```

Though I don't think anyone would name a pipeline/job/resource/task step JUST numbers.

I think it's reasonable to want to allow identifiers to start with numbers, given the one example here of having a time resource that you want to name `1h`.

Would it break anything on the CLI level?
```bash
fly -t ci set-pipeline -p 1-pipeline-name
```
Pipeline names look weird, but I don't think this will break anything at the CLI level.

At the URL level this would be totally fine too. Having numbers at the start of any part of a URL path isn't breaking anything at the browser level or making anything harder for us to do within the web frontend, is it?

Do we want to allow an identifier to be ONLY numbers? A pipeline, job, or resource that's only numbers feels _very_ strange and now I kinda worry it might break something...

I would be fine with allowing the identifier to start with numbers, but let's not allow the identifier to be _only_ numbers. I think that would be a good middle ground here. What do you think?

analytically:
> I would be fine with allowing the identifier to start with numbers, but let's not allow the identifier to be only numbers. I think > that would be a good middle ground here. What do you think?

Agreed. I'll change the PR.

taylorsilva:
Docs here will need to be updated: https://github.com/concourse/docs/blob/70e8fd6ada55064074bc0d98666ca0219fa8d637/lit/docs/config/basic-schemas.lit#L79-L105

taylorsilva (approved):

PR #9121, opened 2025-03-14
Remove 'set_pipeline' experimental warnings

analytically:
The set_pipeline feature is now considered stable and widely used, so remove the warning messages that marked it as experimental. This affects both the self-referential pipeline syntax and team specification warnings.

# Release Note

* Removed "experimental" warnings from the `set_pipeline` step

### DIFF
--- a/atc/exec/set_pipeline_step.go
+++ b/atc/exec/set_pipeline_step.go
@@ -84,11 +84,6 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState, delegate S
 	stderr := delegate.Stderr()
 
 	if step.plan.Name == "self" {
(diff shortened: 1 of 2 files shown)


taylorsilva (approved):
Makes sense, it's not going anywhere or changing behaviour. It's pretty settled.

taylorsilva (changes requested):
Forgot to remove some tests:
```
Summarizing 2 Failures:

  [FAIL] SetPipelineStep when file is configured when pipeline file is good when team is configured when team exists when the target team is the current team [It] should print an experimental message

  github.com/concourse/concourse/atc/exec/set_pipeline_step_test.go:519

  [FAIL] SetPipelineStep when file is configured when pipeline file is good when set-pipeline self [It] should print an experimental message

  github.com/concourse/concourse/atc/exec/set_pipeline_step_test.go:443
```

analytically:
Removed the test checks.

taylorsilva (approved):

PR #9126, opened 2025-03-15
Remove unused 'visited' field from varsTracker struct

analytically:


taylorsilva (approved):
Nit: the unused field was `visited` and this PR renames `visitedAll` to `visited`, and then removes `visitedAll`.

PR #9127, opened 2025-03-15
Add default.pgo PGO file (300s profile) from web instance

analytically:
This commit adds a 300-second Profile-Guided Optimization (PGO) file captured from our production Concourse web instance. Adding this PGO profile will help improve runtime performance by optimizing code paths based on real-world usage patterns.


taylorsilva (changes requested):
I'm rejecting this PR because our current environment has very few real pipelines in it. I'd prefer to create a PGO profile from a cluster with thousands of real pipelines and 10's to hundreds of users actively using. I'd also prefer to ensure we capture this during peak load of a cluster.

Quoting from https://go.dev/doc/pgo
> For best results, it is important that profiles are representative of actual behavior in the application’s production environment. Using an unrepresentative profile is likely to result in a binary with little to no improvement in production.

taylorsilva:
I found this article: https://theyahya.com/posts/go-pgo/
and performed the same test they did. There was no difference at all based on the pgo profile you generated here. Attaching the files here.

[pgo-off.txt](https://github.com/user-attachments/files/19303314/pgo-off.txt)
[pgo-on.txt](https://github.com/user-attachments/files/19303316/pgo-on.txt)

taylorsilva:
Had another thought that maybe the profiling only works if you're compiling against the exact same platform (Linux & amd64 in this case).
Ran the test again with `GOOS=linux GOARCH=amd64` and
image

Still no difference with or without pgo 😞

taylorsilva:
Also I think the `default.pgo` needs to be in the `cmd/concourse` directory. I left it in the root of the repo and had `pgo=default.pgo` set when I ran `go build`.

PR #9130, opened 2025-03-16
atc: enhance container memory unit parsing with IEC notation support

analytically:
Refactor memory unit parsing to improve code clarity and maintainability.
The existing implementation was already using the correct binary multipliers
(2^10, 2^20, 2^30) but with a less clear approach of using powers.

Add support for IEC standard notation (KiB, MiB, GiB) alongside the traditional
units (KB, MB, GB) and abbreviated forms (K, M, G). Fix case sensitivity
issues with a proper case-insensitive regex pattern.

### DIFF
--- a/atc/container_limits.go
+++ b/atc/container_limits.go
@@ -8,7 +8,7 @@ import (
 	"strings"
 )
 
-var memoryRegex = regexp.MustCompile(`^([0-9]+)([GMK]?[B])?$`)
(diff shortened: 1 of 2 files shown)


analytically:
Should we validate limits too? Like there doesn't seem a point in setting limits < 256kb?

taylorsilva (approved):
Docs should be updated: https://github.com/concourse/docs/blob/c7b8d360dfe2ca7ddef04b359dcfa42a6b8d7811/lit/docs/tasks.lit#L287-L290

They don't even mention the current state correctly of being able to use `KB/MB/KB`

PR #9132, opened 2025-03-16
fix: add O_TRUNC flag when writing volume metadata file

analytically:
When writing metadata files, the code was using O_WRONLY|O_CREATE flags without O_TRUNC. This could cause file corruption when writing shorter content than what existed previously, as the remaining bytes from the old content would persist in the file.

The fix adds O_TRUNC to ensure clean overwrites and maintain data integrity.


taylorsilva (approved):
Likely never caused any issues since I don't think there's a case where baggageclaim would try to write the metadata file twice. Definitely doesn't hurt to add this. Thanks!

PR #9134, opened 2025-03-17
Update TLS/SSH configs to balance modern hardware acceleration with broad compatibility, added extensive documentation

analytically:
This commit revises our crypto configuration to better utilize hardware acceleration while maintaining compatibility with a range of hardware:

TLS changes:
- Add X25519 curve and prioritize it (fastest and most secure) while keeping NIST curves for compatibility
- Reorder cipher suites to prioritize AES-GCM algorithms that benefit from hardware acceleration on modern CPUs
- Add ChaCha20-Poly1305 ciphers as fallback for environments without AES acceleration
- Remove deprecated PreferServerCipherSuites flag

SSH changes:
- Retain default Go SSH ciphers which already prioritize AES-GCM appropriately
- Expand MAC algorithm selection to include SHA-512 variants for stronger security where resource constraints aren't an issue
- Reorder key exchange algorithms to prefer curve25519 over NIST curves for better security and performance characteristics

These changes provide optimal performance on modern hardware with AES acceleration (like EC2 i4i/c8g instances) while gracefully adapting to older hardware from ~2020 and maintaining our security standards.

Previous config was overly restrictive with cipher choices and didn't take full advantage of hardware acceleration capabilities.

### DIFF
--- a/atc/config.go
+++ b/atc/config.go
@@ -435,48 +435,87 @@ func (config Config) JobIsPublic(jobName string) (bool, error) {
 
 func DefaultTLSConfig() *tls.Config {
 	return &tls.Config{
+		// Enforce minimum TLS version 1.2 which provides adequate security
+		// TLS 1.0 and 1.1 are considered insecure due to various vulnerabilities
 		MinVersion: tls.VersionTLS12,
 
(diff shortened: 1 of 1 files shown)


analytically:
https://www.amazon.science/blog/better-performing-25519-elliptic-curve-cryptography (this curve now has priority, this change will reduce CPU usage and latency during connection establishment)

PR #9136, opened 2025-03-18
Add S2 compression option for streaming container volumes

analytically:
This commit adds support for S2 compression, optimized for throughput at the cost of some compression ratio.

Key changes:
- Implement S2 compression in the compression package
- Add streaming support for S2 in both Linux and non-Linux environments
- Update encoding types to include S2Encoding

This implementation follows the same pattern as existing zstd and gzip implementations.


analytically:
For comparison between zstd and s2: https://github.com/grafana/mimir/issues/8522

analytically:
https://blog.min.io/transparent-data-compression/

PR #9140, opened 2025-03-19
Fix user namespace ID mapping performance on Linux

analytically:
This commit addresses a significant performance issue with user namespace UID/GID mappings in Linux containers. The baggageclaim component uses these mappings to provide security isolation between processes, but the current implementation was inefficiently handling ID range mappings.

The core issue was in the findMapping function, which only looked at single-entry mappings (Size=1) while ignoring range mappings that cover most UIDs/GIDs. This caused the system to unnecessarily change ownership of almost every file when remapping user namespaces, even when those files didn't actually require changes.

For large volumes with many files, this resulted in thousands of redundant chown and chmod system calls, significantly slowing down container creation and volume operations.

The fix properly handles the full range of ID mappings, ensuring that file ownership is only changed when actually necessary according to the mapping rules. This dramatically reduces the number of system calls during volume operations, particularly for unprivileged containers with large directory hierarchies.

### DIFF
--- a/worker/baggageclaim/uidgid/mapper_linux.go
+++ b/worker/baggageclaim/uidgid/mapper_linux.go
@@ -51,16 +51,21 @@ func (m uidGidMapper) Apply(cmd *exec.Cmd) {
 }
 
 func findMapping(idMap []syscall.SysProcIDMap, fromID int) int {
-	for _, id := range idMap {
-		if id.Size != 1 {
-			continue
-		}
+	for _, mapping := range idMap {
+		// Check if the fromID is within this mapping range
+		startID := mapping.ContainerID
+		endID := mapping.ContainerID + mapping.Size - 1
 
-		if id.ContainerID == fromID {
-			return id.HostID
+		if fromID >= startID && fromID <= endID {
+			// Calculate the offset within the range
+			offset := fromID - mapping.ContainerID
+
+			// Apply the offset to the host ID base
+			return mapping.HostID + offset
 		}
 	}
 
+	// If no mapping found, return the original ID
 	return fromID
 }
 


taylorsilva (commented):
Dug into the commit history and found the prior commit: https://github.com/concourse/baggageclaim/commit/552c40e2b302ace8317b12ea40e9c68416e940e8#diff-1f8974e5e4138c509fa7c5c158699de9910d44356c0b8e890faa7f2e078609f3

So we actually used to do it the way you're proposing in this PR, and switched away from it to the current state. Sadly, the earlier commit doesn't explain _why_ it was changed in this way. The upstream issue the commit links to also doesn't explain why the code changed in this way.

I decided to do some local tests to see if this PR decreases the amount of chown'ing that we do.

I created this pipeline and ran it twice, once with `privileged` set to true and again with `privileged` set to false.

```yaml
jobs:
- name: simple
  plan:
  - task: simple-task
    # privileged: true #comment/uncomment as needed
    config: &config
      platform: linux
      image_resource:
        type: mock
        source: {mirror_self: true}
      run:
        path: echo
        args: ["Hello, world!"]
```

I added a `fmt.Println()` line right before the `chown`: https://github.com/concourse/concourse/blob/c5a058d6ccd71000095c5d73085a5e7de8f447a7/worker/baggageclaim/uidgid/translator.go#L50

I set and ran the pipeline, for both `master` and this PR, and then counted the number of times the `fmt.Println()` statement appeared in the logs of the worker. I brought Concourse up and down between each tests, ensuring I always started with a fresh Concourse instance.

```
master-unprivileged: 714
master-privileged: 1420

PR-unprivileged: 714
PR-privileged: 1420
```

Huh, chown was called the exact same number of times in all states. So we're not seeing a performance improvement with this at all then, are we?

I'm going to stop my review here and leave it in your hands to figure this out, if you think it's worth your time. Feel free to close the PR otherwise.

Let me know if I also missed measuring something else here. My understanding about this stuff feels a tad shaky 🫤

PR #9144, opened 2025-03-19
kernel: Use unix.ByteSliceToString for safe release string handling

analytically:
Replace manual byte conversion and null terminator handling with the
standard unix.ByteSliceToString function from golang.org/x/sys/unix.
This improves cross-platform compatibility and eliminates potential
panic conditions when processing kernel release strings.

The previous implementation manually converted syscall.Utsname.Release
byte-by-byte and required explicit null terminator handling, which
could fail on certain platforms due to type differences or if the
string filled the entire buffer.

See https://github.com/moby/moby/blob/master/pkg/parsers/kernel/kernel_unix.go

### DIFF
--- a/worker/baggageclaim/kernel/kernel_unix.go
+++ b/worker/baggageclaim/kernel/kernel_unix.go
@@ -5,7 +5,9 @@
 // versions for different platforms.
 package kernel
 
-import "bytes"
+import (
+	"golang.org/x/sys/unix"
+)
 
 // GetKernelVersion gets the current kernel version.
 func GetKernelVersion() (*VersionInfo, error) {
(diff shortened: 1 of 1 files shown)


worker/baggageclaim/kernel/kernel_unix.go
```diff
@@ -14,18 +16,8 @@ func GetKernelVersion() (*VersionInfo, error) {
...
-	for _, c := range uts.Release {
-		release[i] = byte(c)
-		i++
-	}
-
 	// Remove the \x00 from the release for Atoi to parse correctly
-	release = release[:bytes.IndexByte(release, 0)]
-
-	return ParseRelease(string(release))
+	return ParseRelease(unix.ByteSliceToString(uts.Release[:]))
```
taylorsilva, line 20: Failing to compile

```
worker/baggageclaim/kernel/kernel_unix.go:20:56: cannot use uts.Release[:] (value of type []int8) as []byte value in argument to unix.ByteSliceToString
```

taylorsilva (changes requested):

taylorsilva:
Doesn't compile and no further movement on this, therefore closing this PR.

PR #9145, opened 2025-03-19
fix(runtime): optimize rootfs_manager SetupCwd to reduce unnecessary syscall

analytically:
Removes redundant os.Stat call in SetupCwd by directly attempting directory creation with MkdirAll and handling the "already exists" error. This eliminates an extra syscall.

### DIFF
--- a/worker/runtime/rootfs_manager.go
+++ b/worker/runtime/rootfs_manager.go
@@ -81,13 +81,8 @@ func NewRootfsManager(opts ...RootfsManagerOpt) *rootfsManager {
 func (r rootfsManager) SetupCwd(rootfsPath string, cwd string) error {
 	abs := filepath.Join(rootfsPath, cwd)
 
-	_, err := os.Stat(abs)
-	if err == nil { // exists
-		return nil
-	}
-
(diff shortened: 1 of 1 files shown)


worker/runtime/rootfs_manager.go
```diff
@@ -81,13 +81,8 @@ func NewRootfsManager(opts ...RootfsManagerOpt) *rootfsManager {
...
 
-	_, err := os.Stat(abs)
-	if err == nil { // exists
-		return nil
-	}
-
-	err = r.mkdirall(abs, 0777)
-	if err != nil {
+	err := r.mkdirall(abs, 0777)
+	if err != nil && !os.IsExist(err) {
```
taylorsilva, line 85: It seems like the `os.IsExist()` check is not necessary at all. Looking at `os.MkdirAll()`: https://github.com/golang/go/blob/dd687c386016445b2f4abc2a958963dd2056d95f/src/os/path.go#L12-L18

It calls `os.Stat()` for us already! I think we can just do this then:

```suggestion
	if err != nil {
```

Wdyt?

PR #9151, opened 2025-03-22
network: fix nil pointer bug in DNS server

analytically:
This commit fixes a potential nil pointer dereference bug in the DNS server's error handling code. The bug occurred when all upstream DNS servers failed, creating an uninitialized *dns.Msg then calling SetRcode() on it.

The fix properly initializes the DNS message before calling methods on it:
- Changes `var m *dns.Msg` to `m := new(dns.Msg)`
- Adds `m.SetReply(r)` to ensure proper ID and question data in error responses
- Adds proper error tracking with a lastErr variable
- Ensures the server returns immediately after sending a successful response
- Adds a test

Additionally, the code is refactored to improve testability by separating the configuration loading from the server creation logic through a new DNSServerWithConfig function that accepts a dns.ClientConfig parameter.

### DIFF
--- a/worker/network/dns.go
+++ b/worker/network/dns.go
@@ -2,32 +2,49 @@ package network
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/miekg/dns"
 )
 
+// DNSServer creates a DNS server that forwards requests to the resolvers
+// specified in /etc/resolv.conf.
 func DNSServer() (*dns.Server, error) {
 	resolvConf, err := dns.ClientConfigFromFile("/etc/resolv.conf")
 	if err != nil {
 		return nil, err
 	}
 
-	var client dns.Client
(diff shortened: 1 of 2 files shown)


worker/network/dns.go
```diff
@@ -2,32 +2,49 @@ package network
...
+			lastErr = err
 		}
 
-		if err != nil {
-			var m *dns.Msg
+		// We only get here if all servers failed
+		if lastErr != nil {
+			m := new(dns.Msg)
+			m.SetReply(r)
 			m.SetRcode(r, dns.RcodeRefused)
```
analytically, line 48: in a later PR we should improve on this, we should not reply with Refused here, but be more specific depending on the type of response we got from upstream

analytically:
TODO: fix the test

PR #9154, opened 2025-03-23
fly/commands: remove placeholder handling from yamlpatch package

analytically:
Removes the external dependency on the abandoned yamlpatch package, remove {{ and }} handling from format_pipeline.


analytically:
https://github.com/krishicks/yaml-patch/blob/master/placeholder_wrapper.go

taylorsilva (changes requested):
If we're copying the code, shouldn't we also carry over the tests? https://github.com/krishicks/yaml-patch/blob/master/placeholder_wrapper_test.go

analytically:
Not sure if there is much point to this as for version 8.0 the `{{` and `}}` are removed?

taylorsilva:
Good point. Should this dependency and code be removed then? Do any other tests fail when they're removed?

analytically:
Removed it, doesn't seem like any tests fail.

PR #9159, opened 2025-03-25
perf: optimize metrics map allocation in emit.go

analytically:
- Pre-allocate attribute maps with exact capacity to avoid resizing
- Improve memory efficiency during high-volume metric collection
- Reduce garbage collection pressure from repeated map allocations


taylorsilva:
Changed by #9216

PR #9162, opened 2025-03-25
perf: optimize countingWaitGroup for faster Count() operations

analytically:
Improve performance of the countingWaitGroup implementation while
maintaining the original behavior:

- Use atomic.Int64 for the counter to enable lock-free Count() reads
- Keep mutex synchronization for Add() and Done() to ensure atomicity
  between WaitGroup operations and counter updates
- Significantly reduce contention when multiple goroutines read the count
  while operations are in progress

This change improves performance when the beacon rebalancing logic
frequently checks the Count() value.

### DIFF
--- a/worker/counting_wait_group.go
+++ b/worker/counting_wait_group.go
@@ -1,31 +1,37 @@
 package worker
 
-import "sync"
+import (
+	"sync"
+	"sync/atomic"
+)
 
+// countingWaitGroup extends sync.WaitGroup with a count that can be queried.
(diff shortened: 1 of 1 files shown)


taylorsilva:
The Beacon rebalance operation is called every 4hrs by default. Not exactly "frequent".
https://github.com/concourse/concourse/blob/a84c7388cbc3d0da5eb28afefcc7bb48fff1df60/worker/workercmd/worker.go#L48

Going to leave the code as-is.

PR #9021, opened 2024-11-01
Replace md5 with sha256 algorithm

IvanChalukov:
[add yourself]: https://github.com/concourse/governance#individual-contributors
[PR requirements]: https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md#pull-request-requirements

## Changes proposed by this PR

closes #9020

Switching `md5` algorithm to `sha256` across the code base and add migration for database naming convention.

* [x] done

## Notes to reviewer

Tested on FIPS compliant database and on already existing database.

## Release Note

* Switches `md5` algorithm to `sha256` across the code base
* Renames database related entries from `md5` to `sha256` suffix

[imperative mood]: https://chris.beams.io/posts/git-commit/#imperative

PR #9165, opened 2025-03-27
Switch use of md5 hashing to sha256

### LINKED ISSUE #9122: Upgrading after switching md5 to sha256 creates duplicate versions
## Summary

Initial report here: https://github.com/concourse/concourse/pull/9021#issuecomment-2725149695

After upgrading, all resources will have their latest version duplicated. This happens because when we save the version here: https://github.com/concourse/concourse/blob/e5f2552c18b00858de492ca73337eb98f4f2bf36/atc/db/resource_config_scope.go#L261-L290

The `ON CONFLICT` won't conflict on the old version because the old version was stored with an md5 digest, not a sha256 digest.

This results in the latest version of a resource to be duplicated in the database and triggers new builds for any jobs that have `trigger: true` for that resource.

## Steps to reproduce

1. Using the `docker-compose.yml` in this repo, change the image to `concourse/concourse:7.12`.
2. Do `docker compose up -d`
3. Set the following pipeline, which will create one version and trigger the job once
```yaml
resources:
- name: some-resource
  type: mock

jobs:
- name: some-passing-job
  plan:
  - get: some-resource
    trigger: true
```
4. Upgrade to the latest commit on `master`. Change the image back to `concourse/concourse:local` and then `docker compose up -d --build`.

## Expected results

The version is not duplicated and jobs are not triggered after upgrading.

## Actual results

The latest version for a resource is duplicated and triggers downstream builds with `trigger: true` set on them.

## Additional context

I gave some initial thoughts on a solution here: https://github.com/concourse/concourse/pull/9021#issuecomment-2725274967

@IvanChalukov - Would you have time to work a solution for this? I may have to revert this otherwise until a solution for this can be made.

## Triaging info

* Concourse version: `master`

IvanChalukov:

Related to #9021 and #9133

closes #9122

FIPS-compliant PostgreSQL instances restrict the use of `md5` hashing, requiring a transition to `sha256`. To accommodate this, a migration has been implemented to convert all existing `md5` hashes in the affected tables to `sha256`. For large databases, this process may take some time to complete.

* [x] Replace MD5 with SHA-256 in all relevant areas
* [x] Update test cases accordingly
* [x] Perform manual testing to ensure UI functionality remains intact, including features like disabling resources and pinning specific resource versions

## Notes to reviewer
Results from tests on highly loaded database:
#### Setup

| **Component**         | **Details**             |
|----------------------|------------------------|
| **Database Instance** | PostgreSQL on AWS      |
| **Instance Class**    | `db.t4g.xlarge`        |
| **vCPU**             | 4                       |
| **RAM**              | 16 GB                   |
| **Engine Version**   | 16.8                    |
| **Storage**          | 250 GiB                 |

#### Data
Tables | Rows
-- | --
resource_config_versions | 10 337 812
build_resource_config_version_inputs | 18 626 661
build_resource_config_version_outputs | 2 745 473
next_build_inputs | 41 482
resource_caches | 531
resource_disabled_versions | 46
builds | 9 804 696
jobs | 40 006

Migration was done in ~19 mins.

## Release Note

* Must be executed during a scheduled downtime window. Depending on the size of your database the migration could take ten's of minutes or even a few hours. Plan accordingly.
* Replaces MD5 with SHA-256 across the codebase
* Migrates all existing MD5-hashed data to SHA-256 in the database

### DIFF
--- a/atc/db/migration/migrations/1743084615_switch_md5_to_sha256.down.sql
+++ b/atc/db/migration/migrations/1743084615_switch_md5_to_sha256.down.sql
@@ -1,0 +1,133 @@
+SELECT clock_timestamp(); -- Start time
+
+-- Step 1: Revert column renames from version_sha256 back to version_md5
+ALTER TABLE resource_config_versions
+RENAME COLUMN version_sha256 TO version_md5;
+
+ALTER TABLE build_resource_config_version_inputs
+RENAME COLUMN version_sha256 TO version_md5;
+
+ALTER TABLE build_resource_config_version_outputs
+RENAME COLUMN version_sha256 TO version_md5;
+
+ALTER TABLE next_build_inputs
+RENAME COLUMN version_sha256 TO version_md5;
+
+ALTER TABLE resource_caches
+RENAME COLUMN version_sha256 TO version_md5;
+
+ALTER TABLE resource_disabled_versions
+RENAME COLUMN version_sha256 TO version_md5;
+
+--- Drop Indexes
+-- resource_config_versions
+ALTER TABLE resource_config_versions 
+    DROP CONSTRAINT IF EXISTS resource_config_scope_id_and_version_sha_unique;
+DROP INDEX IF EXISTS resource_config_versions_check_order_idx;
+DROP INDEX IF EXISTS resource_config_versions_version;
+
+-- resource_disabled_versions
+DROP INDEX IF EXISTS resource_disabled_versions_resource_id_version_sha_uniq; 
+
+-- resource_caches
+DROP INDEX IF EXISTS resource_caches_resource_config_id_version_sha_params_hash_uniq;
+DROP INDEX IF EXISTS resource_caches_resource_config_id;
+
+-- build_resource_config_version_inputs
+DROP INDEX IF EXISTS build_inputs_resource_versions_idx;
+DROP INDEX IF EXISTS build_resource_config_version_inputs_uniq;
+
+-- build_resource_config_version_outputs
+DROP INDEX IF EXISTS build_resource_config_version_outputs_build_id_idx;
+DROP INDEX IF EXISTS build_resource_config_version_outputs_resource_id_idx;
+DROP INDEX IF EXISTS build_resource_config_version_outputs_uniq;
+
+-- next_build_inputs
+ALTER TABLE next_build_inputs 
+    DROP CONSTRAINT IF EXISTS next_build_inputs_unique_job_id_input_name;
+DROP INDEX IF EXISTS next_build_inputs_job_id;
+
+-- Step 2: Revert all rows to their original md5 values
+WITH json_string_cte AS (
+    SELECT 
+        rcv.id,
+        rcv.version_md5 AS old_version_md5,
+        '{' || string_agg('"' || kv.key || '":"' || kv.value || '"', ',' ORDER BY kv.key) || '}' AS json_string
+    FROM resource_config_versions rcv
+    JOIN jsonb_each_text(rcv.version::jsonb) AS kv ON true
+    WHERE jsonb_typeof(rcv.version::jsonb) = 'object'
+    GROUP BY rcv.id, rcv.version_md5
+),
+hashed_json_string_cte AS (
+    SELECT 
+        json_string_cte.id,
+        json_string_cte.old_version_md5,
+        md5(json_string_cte.json_string) AS new_version_md5
+    FROM json_string_cte
+),
+update_resource_versions AS (
+    UPDATE resource_config_versions rcv
+    SET version_md5 = hjs.new_version_md5
+    FROM hashed_json_string_cte hjs
+    WHERE rcv.id = hjs.id
+),
+update_resource_disabled_versions AS (
+    UPDATE resource_disabled_versions rdv
+    SET version_md5 = hjs.new_version_md5
+    FROM hashed_json_string_cte hjs
+    WHERE rdv.version_md5 = hjs.old_version_md5
+),
+update_build_resource_config_version_inputs AS (
+    UPDATE build_resource_config_version_inputs bri
+    SET version_md5 = hjs.new_version_md5
+    FROM hashed_json_string_cte hjs
+    WHERE bri.version_md5 = hjs.old_version_md5
+),
+update_build_resource_config_version_outputs AS (
+    UPDATE build_resource_config_version_outputs bro
+    SET version_md5 = hjs.new_version_md5
+    FROM hashed_json_string_cte hjs
+    WHERE bro.version_md5 = hjs.old_version_md5
+),
+update_resource_caches AS (
+    UPDATE resource_caches rc
+    SET version_md5 = hjs.new_version_md5
+    FROM hashed_json_string_cte hjs
+    WHERE rc.version_md5 = hjs.old_version_md5
+)
+UPDATE next_build_inputs nbi
+SET version_md5 = hjs.new_version_md5
+FROM hashed_json_string_cte hjs
+WHERE nbi.version_md5 = hjs.old_version_md5;
+
+--- Recreate Indexes
+-- resource_config_versions
+ALTER TABLE resource_config_versions
+    ADD CONSTRAINT resource_config_scope_id_and_version_md5_unique UNIQUE (resource_config_scope_id, version_md5);
+CREATE INDEX resource_config_versions_check_order_idx ON resource_config_versions (resource_config_scope_id, check_order DESC);
+CREATE INDEX resource_config_versions_version ON resource_config_versions USING gin(version jsonb_path_ops) WITH (FASTUPDATE = false);
+
+-- build_resource_config_version_inputs
+CREATE INDEX build_inputs_resource_versions_idx ON build_resource_config_version_inputs (resource_id, version_md5);
+CREATE UNIQUE INDEX build_resource_config_version_inputs_uniq
+ON build_resource_config_version_inputs (build_id, resource_id, version_md5, name);
+
+-- build_resource_config_version_outputs
+CREATE INDEX build_resource_config_version_outputs_build_id_idx ON build_resource_config_version_outputs (build_id);
+CREATE INDEX build_resource_config_version_outputs_resource_id_idx ON build_resource_config_version_outputs (resource_id);
+CREATE UNIQUE INDEX build_resource_config_version_outputs_uniq
+ON build_resource_config_version_outputs (build_id, resource_id, version_md5, name);
+
+-- next_build_inputs
+CREATE INDEX next_build_inputs_job_id ON next_build_inputs USING btree (job_id);
+ALTER TABLE ONLY next_build_inputs
+    ADD CONSTRAINT next_build_inputs_unique_job_id_input_name UNIQUE (job_id, input_name);
+
+-- resource_caches
+CREATE INDEX resource_caches_resource_config_id ON resource_caches USING btree (resource_config_id);
+CREATE UNIQUE INDEX resource_caches_resource_config_id_version_md5_params_hash_uniq
+ON resource_caches (resource_config_id, version_md5, params_hash);
+
+-- resource_disabled_versions
+CREATE UNIQUE INDEX resource_disabled_versions_resource_id_version_md5_uniq
+ON resource_disabled_versions (resource_id, version_md5);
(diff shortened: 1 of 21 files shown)


Kump3r (approved):
Tested with a local Concourse and a similar external Postgres database to what @IvanChalukov described. In my case the migration took around ~20min., but in the end I was able to confirm the previous regressions are no longer applicable. 
- No new builds were rescheduled with the same versions ✅ 
- Disabled versions and pinned versions were not duplicated and re-executed. ✅

taylorsilva (commented):
Thanks for doing the timing tests guys. I'm going to do some local upgrade/downgrade testing just to see if I can find any other breaking scenarios before approving.

taylorsilva (approved):
This LGTM. I was trying to think of versions that might cause key ordering to differ between Go and Postgres again. I made the mock resource output a custom version with this value:

```json
{
  "version_one": "first",
  "version_two": "second",
  "version_three": "third",
  "versionOne": "one",
  "versionTwo": "two",
  "versionThree": "three",
  "version-one": "once",
  "version-two": "twice",
  "version-three": "thrice"
}
```

Go ordered it like this:
```
version-one	once
version-three	thrice
version-two	twice
versionOne	one
versionThree	three
versionTwo	two
version_one	first
version_three	third
version_two	second
```

And Postgresql ordered it like this:
```json
{
  "versionOne": "one",
  "versionTwo": "two",
  "version-one": "once",
  "version-two": "twice",
  "version_one": "first",
  "version_two": "second",
  "versionThree": "three",
  "version-three": "thrice",
  "version_three": "third"
}
```

I made a simple pipeline that would trigger on any new versions found.

I `docker compose`'d what's currently on `master`, set the pipeline, let the job trigger once with the custom version.
I then switched over to this PR branch, `docker compose`'d to upgrade my dev instance. No new version was created and the job didn't trigger a second time.

image
(ignore the errors, they're happening because I'm forcing the mock resource to output a version it doesn't know how to read)

image
(second row is the relevant version to track. First row is the custom mock resource type I made).

I'm going to leave this PR open for one more week just in case someone can think of something else related to the resource versions that might cause issues. I think the one I thought of is unlike anything the majority of resources would ever output as a version. Hopefully it is edge-case-y enough to give us confidence that this does work now and won't cause a bunch of versions to be recreated.

Thanks for the hard work @IvanChalukov!

atc/db/migration/migrations/1743084615_switch_md5_to_sha256.down.sql
```diff
@@ -0,0 +1,131 @@
...
+DROP INDEX IF EXISTS build_resource_config_version_outputs_resource_id_idx;
+DROP INDEX IF EXISTS build_resource_config_version_outputs_uniq;
+
+-- next_build_inputs
+ALTER TABLE next_build_inputs 
+    DROP CONSTRAINT IF EXISTS next_build_inputs_unique_job_id_input_name;
+DROP INDEX IF EXISTS next_build_inputs_job_id;
+
+-- Step 2: Revert all rows to their original md5 values
+WITH json_string_cte AS (
```
aliculPix4D, line 49: @IvanChalukov FYI the following is probably safer...

We had some resources in database with: `{}` reported as version and also some with NULL, but the version_md5 (or version_sha256) is still calculated for them.

So we should cover at least the  `{}`, since they are much more common than we expected.
```
WITH json_string_cte AS (
    SELECT 
        rcv.id,
        rcv.version_md5 AS old_version_md5,
        COALESCE(
            '{' || string_agg('"' || kv.key || '":"' || kv.value || '"', ',' ORDER BY kv.key) || '}',
            '{}'
        ) AS json_string
    FROM resource_config_versions rcv
    LEFT OUTER JOIN jsonb_each_text(rcv.version::jsonb) AS kv ON true
    WHERE jsonb_typeof(rcv.version::jsonb) = 'object'
    GROUP BY rcv.id, rcv.version_md5
),
```

IvanChalukov: @aliculPix4D Does this cause any anomalies in the UI or trigger unexpected resource checks after the migration?

Also, could you provide a sample resource that is stored with `{}` as its version so I can validate the behavior?

IvanChalukov: Hi @aliculPix4D, I tried setting up a resource with empty version but wasn’t able to get it working. Do you have any updates on the questions from my previous comment? Regarding the resources that have `{}` as their version—are they still in use, or just leftovers?

aliculPix4D: Hi @IvanChalukov 

Sorry for the late response. I did double check this, and I "think" (not 100% sure) the reason for it is because these resources all had `check_every: never` (a "put-only" resource). This was the only similarity that I could find in the resource config in those resources.

> Does this cause any anomalies in the UI or trigger unexpected resource checks after the migration?

No, not really. Let me give you a bit more context..

The reason we noticed this is because we deployed in our production our own custom release of Concourse which contained the original "broken" PR that did the switch from md5 to sha256 (See: https://github.com/concourse/concourse/pull/9021#issuecomment-2725149695). We were running this custom release in production, so all our new version hashes in database tables were in sha256 format (we had a mix in md5 from before deployment and sha256 from after deployment). Since the PR got reverted, we had to "recover" and revert this change by doing the same recalculation you are doing in this PR, just in the opposite direction (basically your down script).  We started with your down.sql script, but as it turns out, it was a bit more complex than this and my colleague (@odormond ) had to almost completely rewrite it.

The reason it was more complicated for us, are the duplicated entries that appeared. We were exposed to the problem because the old up-script (from the original PR) was too simple and did not try to convert the hashes. Then, when our concourse run, resource checks produced new versions whose hashes were not found in the DB (sha256 != md5) so they were inserted but the actual version they discovered (the real JSON object) was already there (they were not really "new") -> duplicates.

Originally, I thought you would require the same "improvements" here, and hence I started to report them, but after further discussion, we don't think your PR would need to change the up/down script because simply your up script from this PR would do the proper recalculation from the get-go, and hence you will never get those duplicated entries and the same troubles we were in.

aliculPix4D: Note that I also noticed that in our database we have 12 versions that have are not covered due to:
`WHERE jsonb_typeof(rcv.version::jsonb) = 'object'` filter.
```
     id     | version |           version_md5            
------------+---------+----------------------------------
 3578861433 | null    | 37a6259cc0c1dae299a7866489dff0b
...
```
where the 37a6259cc0c1dae299a7866489dff0b is just md5 of null string.

I am unsure how common is this occurance in other deployments, but maybe something to also keep in mind.

IvanChalukov: Thanks for the detailed explanation, @aliculPix4D. I’ll go ahead and resolve this conversation since there are no action items. Feel free to reach out if you come across anything that might be problematic. I really appreciate your findings and suggestions.

IvanChalukov: We expect the version to be valid JSON rather than `null`, so it can be properly structured as key/value pairs. Handling `null` objects doesn’t seem appropriate on my side. You could clean up the versions for this resource.
And again I do not know how this version could be used in any pipeline.

taylorsilva: Thanks for hashing this out guys! I think I'm going to try and do some data collection from the community here. I'm worried about breaking people's databases or making the upgrade very hard for folks to do. Right now I agree with you Ivan, it appears there's nothing we need to change here. The super cautious side of me wants to further validate how likely it is that users will run into these edge-cases. I'll open up an issue for this though and make call-out for how community users can help us. I'm going to look at the ci.concourse-ci.org database first.

taylorsilva:
Following up on the previous conversation in this PR currently marked as resolved: I locally ran a copy of the `ci.concourse-ci.org` database, looking for any `{}` or NULL versions in `resource_config_versions` and `resource_caches` tables. I did not find any in there. I used this query:

```sql
select count("version") from resource_config_versions where "version" = '{}'::jsonb or "version" IS NULL;
select count("version") from resource_caches rc where "version" = '{}'::jsonb or "version" is NULL;
```

I'm going to make a call out to the community to run the above query to see if this is also the case for others or if Pix4D's case is more common.

---

Since I had a local copy of our prod db, I ran the migration just to get a sense of how long it might take to run in prod and got a 6min runtime. This is on an M1 macbook pro with 32GB ram, running Postgres 17 natively (not in docker).

taylorsilva:
@IvanChalukov Based on the data so far from https://github.com/orgs/concourse/discussions/9265 it looks like empty versions `{}` are present in people's databases.

I think these empty versions must come from custom resources. I modified the `mock` resource type to return an empty version. You can use that image here: https://hub.docker.com/r/concourseteam/mock-resource/tags

image

```yaml
resource_types:
  - name: mock-empty
    type: registry-image
    source:
      repository: concourseteam/mock-resource
      tag: empty-version

resources:
  - name: mock
    type: mock-empty

jobs:
  - name: job
    plan:
      - get: mock
```

image

This is probably something we shouldn't allow. I'll open a separate issue for that.

Anyways, what happens when we upgrade a db running Concourse 7.14 to `master`?
I setup a local Concourse running 7.14 and set the above pipeline. Db looks like this:
image

Then I upgraded it the latest code on `master`. Db looked like this after:
image

Empty versions are not recalculated, as previously stated in prior comments.

I've also just discovered another issue related to 7.14. The latest migration number comes after the one in this PR: 
image

Which currently causes upgrades to fail.

## Action Items

Based on the above, the following work needs to be done:
- Migration number needs to be updated so it's the latest migration. I'll go ahead and fix this myself.
- The migration needs to be updated to handle `{}` empty json versions. I'd like to assign this to you @IvanChalukov 
- Open new issue to potentially filter out or error when resources return empty versions.

aliculPix4D:
> I've also just discovered another issue related to 7.14. The latest migration number comes after the one in this PR:

@taylorsilva does this mean that people who deployed v7.14.X all have this issue with migration history and that you will have to release v7.14.2? or that happened only in your experiment because you first deployed the master?

Kump3r:
> > I've also just discovered another issue related to 7.14. The latest migration number comes after the one in this PR:
> 
> @taylorsilva does this mean that people who deployed v7.14.X all have this issue with migration history and that you will have to release v7.14.2? or that happened only in your experiment because you first deployed the master?

No it means that in master the md5_sha256 migration needs to come after the other, since it uses timestamp names to schedule the order of migrations, but due to the older migration still not being released it needs to be renamed.

taylorsilva:
What Kalin said! There will be no issue for users, this was only an issue present on `master` and doesn't affect any released version of Concourse.

ragaskar:
👋 

Hi folks -- not following all of this, but I see that in this thread there was possibly some concern that these changes could cause builds to be incorrectly re-run, I thought I'd add some anecdotal data: 

we just upgraded to 8.0.0 and it looks like we had re-runs for jobs that triggered off of the gcs-resource. That is, it _looks like_ after our deployment we had two jobs get triggered by inputs that we had already processed. I'm not really up-to-speed on how version-ing etc works with resources so my statement might appear nonsensical, apologies. 

I see two jobs where this has happened (both related to [gcs-resource](https://github.com/frodenas/gcs-resource)). Those two jobs do not have a "version" (maybe?? I'm not sure if the "version" returned as part of the resource `in` is the same thing as the concourse `version` for the resource) because we don't set a regex for them -- they just have a "generation".  Curiously previous builds never show this resource appearing as "changed" -- even when the generation ticks up. 

There are another few jobs that use the gcs-resource that didn't re-trigger and they _did_ have a regex set and thus had a `version` value. If this is the same `version` as a concourse `version` it sounds like that explains our problem ... 

After reading more of the gcs-resource README I think the "why didn't it change previously" makes more sense to me: 

"Objects will be found via the pattern configured by regexp or versioned_file. The versions will be used to order them (using [semver](http://semver.org/)). Each object's filename is the resulting version." 

The filename would never have changed in this case. Doesn't necessarily explain why it triggered now, but maybe that comes down to the hash approach changing? I dunno. It's not that big of a deal for us, but maybe I'll come back if we see future triggering fail for some reason. 

No request for action, just noting this in case someone else shows up having observed similar behavior looking for anecdotes.

taylorsilva:
We did find issues related to jobs re-running or not running at all with versions they previously ran with. Thanks for trying to explain your situation. It sounds like there may have been a bug of some kind but there's not enough info for me to speculate. As you said, you don't fully get how the resource versioning stuff works, so maybe things are behaving as expected?

@ragaskar 
If you want to share the job history of the version that triggered the re-run, that would be help. This page specifically when you click on a resource in a pipeline:
image

And highlight the jobs that got re-run. Then on the build logs for that job, if you could show the `get` steps, like this:
image

If the `get` step has the text highlighted yellow, that means it was the cause of the trigger. I'm curious if that's the case for your gcs resource's version.

PR #9166, opened 2025-03-28
refactor: simplify hasPermission method in accessor

analytically:
Improve code readability by removing redundant boolean operation and temporary variable in the hasPermission method. No functional change as this simplification would likely be performed by the compiler anyway.


taylorsilva (approved):
whoa, that was a weird way to do that before. Feels like someone overthinking things 😅

Issue #9147, opened 2025-03-20
Release 7.13.0


taylorsilva:
Steps for a new major/minor release:

* [x] Add this issue to the `v<M.m.x>` milestone

* [x] Create your release branch on the `concourse/concourse` github repo with the following format `release/M.m.x` (M being the major version and m being the minor version)

* [x] Create the release branch on `concourse/concourse-docker` repository.

* [x] Create the release branch on the `concourse/concourse-bosh-release` repository. Make any missing changes to the spec of `web` or `worker` depending on if the release contains any changes that adds or modifies any flags.

  * Any changes you make on the branch will not get automatically merged back to master so try to make the changes on master and then create the branch from there.
  * We should really have something that will merge the branch back to master. (like we do for the `concourse/concourse` branches)

* [x] Create the release branch on `concourse/concourse-bosh-deployment` repository.

* [x] Create the release branch on `concourse/concourse-chart` repository from the `dev` branch. Make any missing changes to `values.yaml` or `templates/web-deployment.yaml` for changes to flags on web or `templates/worker-deployment.yaml` for changes to flags on the worker.

* [x] Bump the appropriate versions for resource types. Go to the releases page `https://project.concourse-ci.org/releases` and take a look at which resource type repositories have had new commits or PRs. Take a look at what those changes entail and bump the version in their respective pipeline in `ci.concourse-ci.org`.

  * If the changes were only README or repo restructuring changes with no user impact, you don't need to bump the version
  * If the changes were small bug fixes or changes, you can do a patch version bump
  * If the changes were adding of features, you can do a minor version bump
  * If the changes involve a breaking changes, that should be a major version bump
  * [x] https://github.com/concourse/git-resource/ - Minor bump
  * [x] https://github.com/concourse/registry-image-resource - Minor bump
  * [x] https://github.com/concourse/time-resource/ - Minor bump
  * [x] https://github.com/concourse/s3-resource - Major bump. V2 signing removed.
  * [x] https://github.com/concourse/github-release-resource/ - Minor bump
  * [x] https://github.com/concourse/pool-resource/ - Patch bump
  * [x] https://github.com/concourse/semver-resource - Minor bump
  * [x] https://github.com/concourse/docker-image-resource/ - Patch bump
  * [x] https://github.com/concourse/hg-resource - Patch bump
  * [x] https://github.com/concourse/bosh-io-release-resource - Patch bump
  * [x] https://github.com/concourse/bosh-io-stemcell-resource - Patch bump

[...]

PR #9173, opened 2025-04-01
runtime: Seccomp update

taylorsilva:

Related to #9147

Fixes the issue we found on workers running on v5 of the linux kernel. When running `curl` we'd see `getaddrinfo() thread failed to start`.

The seccomp profile now has `clone3` always returning `ENOSYS` which glibc uses to know it should fallback to regular `clone`.

## Release Note

* Update seccomp profile to be in sync with Docker/Containerd's default profiles


worker/runtime/spec/seccomp.go
```diff
@@ -356,22 +412,34 @@ var seccomp = &specs.LinuxSeccomp{
 				Op:       specs.OpMaskedEqual,
 			},
 		),
+		{
+			// Make clone3 return ENOSYS instead of the default EPERM so
+			// callers (glibc) fallback to calling clone, instead of error'ing
+			Names:    []string{"clone3"},
+			Action:   specs.ActErrno,
+			ErrnoRet: &enosys,
+		},
```
analytically, line 421: so it's not just allowing clone3 that would work?

PR #9177, opened 2025-04-06
Warn instead of error on unused resources

### LINKED ISSUE #4177: Warn instead of fail on unused resources

IvanChalukov:

resubmitting of #5060
closes #4177

The `validateResourcesUnused` function is now called within `validateResources` instead of `validateResourceTypes` as it was in #5060. This placement is more convenient since it is related to `resources`.

* [x] Implement support for unused resources
* [x] Update tests

## Release Note

* Allow configuring pipelines even when they contain unused resources.

### DIFF
--- a/atc/configvalidate/validate.go
+++ b/atc/configvalidate/validate.go
@@ -208,7 +208,7 @@ func validateResources(c atc.Config) ([]atc.ConfigWarning, error) {
 		}
 	}
 
-	errorMessages = append(errorMessages, validateResourcesUnused(c)...)
+	warnings = append(warnings, validateResourcesUnused(c)...)
 
 	return warnings, compositeErr(errorMessages)
 }
@@ -287,18 +287,20 @@ func validatePrototypes(c atc.Config, seenTypes map[string]location) ([]atc.Conf
 	return warnings, compositeErr(errorMessages)
 }
 
-func validateResourcesUnused(c atc.Config) []string {
+func validateResourcesUnused(c atc.Config) []atc.ConfigWarning {
+	warnings := []atc.ConfigWarning{}
 	usedResources := usedResources(c)
 
-	var errorMessages []string
 	for _, resource := range c.Resources {
 		if _, used := usedResources[resource.Name]; !used {
-			message := fmt.Sprintf("resource '%s' is not used", resource.Name)
-			errorMessages = append(errorMessages, message)
+			warnings = append(warnings, atc.ConfigWarning{
+				Type:    "resources",
+				Message: "resource '" + resource.Name + "' is not used in pipeline",
+			})
 		}
 	}
 
-	return errorMessages
+	return warnings
 }
 
 func usedResources(c atc.Config) map[string]bool {
--- a/atc/configvalidate/validate_test.go
+++ b/atc/configvalidate/validate_test.go
@@ -167,8 +167,9 @@ var _ = Describe("ValidateConfig", func() {
 			})
 
 			It("returns a warning", func() {
-				Expect(warnings).To(HaveLen(1))
+				Expect(warnings).To(HaveLen(2))
 				Expect(warnings[0].Message).To(ContainSubstring("'_some-resource' is not a valid identifier"))
+				Expect(warnings[1].Message).To(ContainSubstring("resource '_some-resource' is not used"))
 			})
 		})
 
@@ -583,7 +584,7 @@ var _ = Describe("ValidateConfig", func() {
 						Type: "some-type",
 					},
 					{
-						Name: "get-alias",
+						Name: "unused-get-alias",
 						Type: "some-type",
 					},
 					{
@@ -595,7 +596,7 @@ var _ = Describe("ValidateConfig", func() {
 						Type: "some-type",
 					},
 					{
-						Name: "put-alias",
+						Name: "unused-put-alias",
 						Type: "some-type",
 					},
 					{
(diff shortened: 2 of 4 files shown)


Kump3r:
So from what I read in the original issue, please correct me if I am wrong, the check will not be executed on unused resources right? But still their versions/configuration would be kept in the database. So what worries me is that users that don’t know what they are doing could end up with a lot of garbage. Also possibly they could be confused as to why their pipeline was not fetching “repo-a”, or triggered in “time-a”. I understand the benefits here, just trying to think on possible side effects. I will try to allocate time to review and test this.

IvanChalukov:
I've managed to test it out and I can confirm that the unused resources versions/configurations are kept in the database until pipeline is deleted. Based on the issue and previous PRs, I assume this is the intended behavior, but it might not be ideal for all users — particularly those unaware of the implications.
I could extend the warning message during pipeline setup to inform users that unused resource data will be preserved. Additionally, the `validate-pipeline` command with the `--strict` flag can be used to ensure all defined resources are actually used in the pipeline before applying it. 
@taylorsilva, could you share your thoughts on this?

taylorsilva:
Hey, I'll try to review it this week. Thanks as usual for the work guys!

kcbimonte:
> I've managed to test it out and I can confirm that the unused resources versions/configurations are kept in the database until pipeline is deleted. Based on the issue and previous PRs, I assume this is the intended behavior, but it might not be ideal for all users — particularly those unaware of the implications.
> I could extend the warning message during pipeline setup to inform users that unused resource data will be preserved. Additionally, the `validate-pipeline` command with the `--strict` flag can be used to ensure all defined resources are actually used in the pipeline before applying it. 
> @taylorsilva, could you share your thoughts on this? 

If the resource itself is still stored in the DB, I'd honestly prefer the inverse of the strict flag. Maybe make strict take a boolean to disable it to allow unusued resources and keep the existing behavior as default. 

In other words, treat this as opt-in functionality instead of opt-out functionality.

taylorsilva:
> If the resource itself is still stored in the DB, I'd honestly prefer the inverse of the strict flag

Can you explain your position some more please @kcbimonte? What is your concern with the resource still being in the DB even if it's not used in the pipeline?

My initial concern was if unused resources would still have checks run for them. That would be a waste of compute. From #5060 we can see in this comment https://github.com/concourse/concourse/pull/5060#issuecomment-577784162 that #4944 makes this not possible.

What are your concerns with the resource config and versions still being stored in the DB?

taylorsilva (changes requested):
There are a few things with overall behaviour that I don't think is ideal still. This is one of those problems where it seems simple (just change this one thing!) but the higher integration-level effect is wider impacting.

#6281 does stop unused resources from being _continuously_ checked, but unused resources do still get checked _once_ when they are initially created. After that first check they aren't checked again. So there is some compute time and space being used up by these unused resources.

Now, onto the issues I have with the current behaviour introduced by this PR.

1) The warning prints twice and says "Deprecation Warning". Wording is odd here as we're not deprecating anything, we're only warning. I think it should just say "Warning".
    image

2) If an unused resource is incorrectly configured, which will result in it error'ing when it's checked:
    ```yaml
    - name: time-3
      type: time
      source:
        interval: not an interval
    ```
    The "error triangle" shows up in the web UI:
    image
    Which will be confusing to users since you can't see where the error is coming from if you click into the pipeline:
    image
    I don't have a suggestion for how to deal with this. You could 1) don't store the unused resource in the db at all, or 2) filter out unused resources at some other layer (frontend, API, Db, etc.).


Lastly, I agree with @kcbimonte's opinion. I also prefer when Concourse stops me from setting a pipeline with an unused resource. For me and the way I write pipelines, it usually means I've made a mistake of some kind.

A middle ground might be to add an opt-in flag to allow for unused resources if desired. Should we add this to the `set_pipeline` step as well then?

Given all of that, I'm starting to doubt the use-case of #4177. Is it worth making this behavioural change? Do you guys at SAP have a strong reason to want this?

taylorsilva:
Sample pipeline I used for review:

```yaml
resources:
- name: time-1
  type: time
  check_every: 10s
  source:
    interval: 10s

- name: time-2
  type: time
  source:
    interval: 30s

- name: time-3
  type: time
  source:
    interval: not an interval

jobs:
- name: job-a
  plan:
    - get: time-1
```

taylorsilva:
Thanks Ivan.

I'm going to close the related issue as well. I don't think anyone is strongly asking for this.

PR #9179, opened 2025-04-07
Replace setTimeout with requestAnimationFrame for visual operations

analytically:
Improve rendering performance by synchronizing visual DOM operations with the browser's rendering cycle. This change specifically targets operations that directly influence what the user sees, such as:

- Pipeline SVG rendering
- Textarea height adjustment
- Element visibility detection
- SVG icon and causality graph rendering

Non-visual operations like localStorage interaction remain using setTimeout for better compatibility with the event loop. Operations requiring more aggressive polling like sticky header syncing remain using setInterval.

### DIFF
--- a/web/public/elm-setup.js
+++ b/web/public/elm-setup.js
@@ -20,7 +20,7 @@ app.ports.pinTeamNames.subscribe(function(config) {
   const lowestHeaderTop = (section) => body(section).offsetTop + body(section).scrollHeight - header(section).scrollHeight;
 
   const pageHeaderHeight = () => config.pageHeaderHeight;
-  const viewportTop = () => window.pageYOffset + pageHeaderHeight();
+  const viewportTop = () => window.scrollY + pageHeaderHeight();
 
   const updateHeader = (section) => {
     var scrolledFarEnough = section.offsetTop < viewportTop();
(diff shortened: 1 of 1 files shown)


web/public/elm-setup.js
```diff
@@ -62,8 +62,7 @@ app.ports.renderPipeline.subscribe(function (values) {
   const jobs = values[0];
   const resources = values[1];
   renderingModulePromise.then(({renderPipeline}) =>
-    // elm 0.17 bug, see https://github.com/elm-lang/core/issues/595
-    setTimeout(() => renderPipeline(jobs, resources, app.ports.newUrl), 0)
```
analytically, line 66: we should investigate if this ELM bug is still relevant... https://github.com/elm/core/issues/595

analytically: so this is still relevant on Safari, doesn't render without the animationFrame

taylorsilva: To confirm, you could reproduce the issue on Safari (Version 18.5 20621.2.5.11.8)? I'm trying to do the same right now, but everything appears to be working. I made sure the js file wasn't being cached by the browser and then did the following:

1. Open Concourse dashboard, no pipelines visible
2. Set a pipeline with fly
3. went back to dashboard, pipeline appeared. Pressing Pause/Unpause works
4. Clicking into the pipeline renders correctly
5. Starting a new build of the job in the pipeline works. Build logs render as expected.

Did you do anything else to reproduce the issue?

analytically: Yes I could reproduce the issue with Safari 18.5. It rendered a blank screen. But worked fine on Firefox/Chrome. Not sure what I did to reproduce it. Will need to try again.

Issue #1588, opened 2017-09-13
Add BUILD_URL metadata


gregarcara:
# Feature Request

Hi, it would be great if along with the current metadata made available in PUT and GET there was also a BUILD_URL environment variable that would have the URL that is for the current build. All the pieces are there now to build the URL but it would be a great for notification resources.

PR #9187, opened 2025-04-17
Add BUILD_URL in StepMetadata.Env()

analytically:
This commit:
- Adds a BUILD_URL env parameter to step metadata
- Adds proper handling for one-off jobs with a different URL pattern

Was also requested here: https://github.com/concourse/concourse/issues/1588

### DIFF
--- a/atc/exec/step_metadata.go
+++ b/atc/exec/step_metadata.go
@@ -61,6 +61,33 @@ func (metadata StepMetadata) Env() []string {
 
 	if metadata.ExternalURL != "" {
 		env = append(env, "ATC_EXTERNAL_URL="+metadata.ExternalURL)
+
+		if metadata.TeamName != "" && metadata.BuildID != 0 {
+			var buildURL string
+
+			if metadata.PipelineName != "" && metadata.JobName != "" && metadata.BuildName != "" {
+				// Regular job build URL
+				detailedURL := fmt.Sprintf("%s/teams/%s/pipelines/%s/jobs/%s/builds/%s",
+					metadata.ExternalURL,
+					metadata.TeamName,
+					metadata.PipelineName,
+					metadata.JobName,
+					metadata.BuildName)
+
+				// Check if URL exceeds 255 characters
+				if len(detailedURL) > 255 {
+					// Use short format instead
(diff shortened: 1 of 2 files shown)


analytically:
Also discussed at length here: https://github.com/concourse/concourse/pull/2747

kcbimonte:
So, we developed a resource internally that allows for Concourse to "run" external pipelines and report them as such to GitLab. GitLab has a limit of 255 characters for Target URLs ([Commit Status API](https://docs.gitlab.com/api/commits/#set-the-pipeline-status-of-a-commit)), so we instead went with using the Build ID path style instead (e.g. https://ci.concourse-ci.org/builds/411969621).

Since certain tools have path length limitations on where we'd realistically use this BUILD_URL variable, does it make more sense to instead just compose the BUILD_URL based off the `$ATC_EXTERNAL_URL/builds/$BUILD_ID` so that we can stay within the sensible path limits while also providing a ready-made URL?

analytically:
@kcbimonte I've added a test for > 255 char length limit and used the shorter version when that's the case

atc/exec/step_metadata.go
```diff
@@ -61,6 +61,33 @@ func (metadata StepMetadata) Env() []string {
...
+					metadata.JobName,
+					metadata.BuildName)
+
+				// Check if URL exceeds 255 characters
+				if len(detailedURL) > 255 {
+					// Use short format instead
+					buildURL = fmt.Sprintf("%s/builds/%d", metadata.ExternalURL, metadata.BuildID)
+				} else {
+					buildURL = detailedURL
+				}
```
taylorsilva, line 83: I don't like how character length determines whether you get a long or short URL. What if you always want the long or short URL? You have no guaranteed way to get either now.

I'd prefer going back to what #2747 was doing, setting a `BUILD_URL` and `BUILD_URL_SHORT`.

taylorsilva (changes requested):
Would be nice to add an acceptance level test as well: https://github.com/concourse/concourse/pull/2747#issuecomment-439919525

atc/exec/step_metadata.go
analytically: Happy to add BUILD_URL_SHORT but I disagree about length determining the URL, personally I think this is a valid workaround to support the second largest Git site out of the box

PR #9188, opened 2025-04-18
atc: Consider image volumes in volume-locality strategy

analytically:
See https://github.com/concourse/concourse/pull/8057

Often the largest volume that may need streaming is the image that the container executes in. By considering this volume too, this may allow for slightly better placement/reduced streaming for some builds (in particular builds with no inputs or caches will now prefer to be on the same worker as their image volume rather than a random one)

### DIFF
--- a/atc/worker/placement.go
+++ b/atc/worker/placement.go
@@ -166,20 +166,17 @@ func (strategy volumeLocalityStrategy) Order(logger lager.Logger, pool Pool, wor
 		counts[worker.Name()] = 0
 	}
 
-	for _, input := range spec.Inputs {
-		if input.FromCache {
-			continue
-		}
+	updateCountsForArtifact := func(artifact runtime.Artifact, destinationPath string) error {
+		volume, ok := artifact.(runtime.Volume)
 
-		volume, ok := input.Artifact.(runtime.Volume)
 		if !ok {
 			// Non-volume artifacts don't live on workers, so don't affect
 			// volume locality decisions.
-			continue
+			return nil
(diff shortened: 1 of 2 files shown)


taylorsilva (changes requested):
I'm not sure this is an assumption we want to bake into the placement strategy. We're assuming that the image artifact is _always_ the volume that takes up the most disk space. We know this is not always the case though.

I'd be open to the latter change this PR makes though, where if there are no other inputs then the image weight is taken into account. That makes sense for all containers in this situation.

analytically:
So you're saying remove the additional (5) weight for the image artifact?

analytically:
The weight was requested on the original PR by @evanchaoli 
screentshot-2025-05-13 at 20 42 17@2x

taylorsilva:
Ah, no one ever reviewed that PR. Evan was also never a maintainer. My feedback is the same.

I'd also like for Andy to get credit for the work done by him. Updating commit message to include him as Co-Author

atc/worker/placement.go
```diff
@@ -215,6 +213,23 @@ func (strategy volumeLocalityStrategy) Order(logger lager.Logger, pool Pool, wor
 				counts[worker]++
 			}
 		}
+
+		return nil
+	}
+
+	err := updateCountsForArtifactWithWeight(spec.ImageSpec.ImageArtifact, "/")
```
analytically, line 220: Rename the method if weight is removed

taylorsilva (approved):

PR #9202, opened 2025-05-15
Return Error for invalid identifiers instead of Warning

### LINKED ISSUE #8475: Step `set-pipeline` can take invalid/problematic target pipeline name (in conjunction with instanced pipelines)
## Summary
We introduced a new pipeline for building hotfixes leveraging the new instanced pipeline feature.
Everything works so far, the issue arose as we gave the new instance pipeline a name containing a `/` without further thinking about it (pipeline name: `hotfix/foo-bar-baz`.

The pipeline then was created successfully, but could not be accessed via UI as the `/` broke the frontend URL handling. Using an URL escaped `/` also didn't help and would only be workaround.

Also deleting the instanced pipeline was a bit messy.

## Steps to reproduce
* build a pipeline the the creates an instanced pipeline
* give the target pipeline a name containing a `/``
* Let the instanced pipeline be created by Concourse
* Try to access the new instanced pipeline via UI
* Try to get rid of the instanced pipeline

## Expected results
The `set-pipeline` step should somehow sanitise or validate the target name of the instance pipeline to create and make sure it fits the required naming conventions. An easy solution could be to just disallow `/` in the name and presenting a proper error message to the user.

## Actual results
Target name of instanced pipeline is not validated and can contain `/`.

## Triaging info

* Concourse version: `7.8.1`
* Browser (if applicable): `Chrome Mac 03.0.5060.114 (Official Build) (x86_64)`

### LINKED ISSUE #8725: Fly cannot delete Instanced Pipelines with Special Characters
## Steps to reproduce

Create a instanced-pipeline with commas or other special characters:
```
fly -t my-team set-pipeline --pipeline my-group -c pipeline.yml --instance-var env=prod,staging,dev
```

then run

```
fly -t my-team destroy-pipeline --pipeline my-group/env:"prod,staging,dev"
```

## Expected results

Either I should not be allowed to create a pipeline instance that contains commas, or there should be a documented way to delete it through fly.

Double-quoting or using escape characters should also allow me to delete it.

## Actual results

```
fly -t my-team destroy-pipeline --pipeline my-group/env:"prod,staging,dev"
!!! this will remove all data for pipeline `my-group/env:prod`

fly -t my-team destroy-pipeline --pipeline my-group/env:"prod\,staging\,dev"
!!! this will remove all data for pipeline `my-group/env:prod\\`
```

## Triaging info

* Concourse version: v7.8.3

### LINKED ISSUE #9271: Resolve RFC-040 - Error on invalid identifiers
### Summary

Return an error instead of a warning when an invalid identifier is detected, as proposed by [the RFC](https://github.com/concourse/rfcs/blob/master/040-valid-identifiers/proposal.md#easing-the-transition). Users have had 5 years to transition their identifiers to meet these requirements.

The current PR that will resolve this issue already notes some issues with completely switching over to this. There are some edge-cases we need to think about related to vars `((...))` being used.

### Context

Discussion: https://github.com/orgs/concourse/discussions/9184
RFC: https://github.com/concourse/rfcs/blob/master/040-valid-identifiers/proposal.md
Current PR: https://github.com/concourse/concourse/pull/9202

ceco556:

closes #9271
closes https://github.com/concourse/concourse/issues/8725
closes https://github.com/concourse/concourse/issues/8475

* [x] When there is invalid identifiers in a fly command it will return Error instead of Warning.
* [x] Check for invalid identifiers in the set_pipeline step and return Error if there is. 

## Notes to reviewer

I changed Warnings to be ConfigErrors. Let me know if you are okay with this. If needed i will rename it to IdentifierErrors or something else.

## Release Note

Fly commands now will return Error when there is invalid identifiers found.
If there is set_pipeline step with invalid identifier in a pipeline and you try to set it, it will return Error.

### DIFF
--- a/atc/step_validator.go
+++ b/atc/step_validator.go
@@ -7,14 +7,10 @@ import (
 )
 
 // StepValidator is a StepVisitor which validates each step that visits it,
-// collecting warnings and errors as it goes.
+// collecting errors as it goes.
 type StepValidator struct {
-	// Warnings is a slice of warning messages to show to the user, while still
-	// allowing the pipeline to be configured. This is typically used for
-	// deprecations.
-	//
 	// This field will be populated after visiting the step.
-	Warnings []ConfigWarning
+	ConfigErrors []ConfigErrors
 
 	// Errors is a slice of critical errors which will prevent configuring the
 	// pipeline.
@@ -36,8 +32,8 @@ type scope map[string]bool
 // The Config specified is used to validate the existence of resources and jobs
 // referenced by steps.
 //
-// The context argument contains the initial context used to annotate error and
-// warning messages. For example, []string{"jobs(foo)", ".plan"} will result in
+// The context argument contains the initial context used to annotate error
+// messages. For example, []string{"jobs(foo)", ".plan"} will result in
 // errors like 'jobs(foo).plan.task(bar): blah blah'.
 func NewStepValidator(config Config, context []string) *StepValidator {
 	return &StepValidator{
@@ -64,12 +60,9 @@ func (validator *StepValidator) VisitTask(plan *TaskStep) error {
 	validator.pushContextf(".task(%s)", plan.Name)
 	defer validator.popContext()
 
-	warning, err := ValidateIdentifier(plan.Name, validator.context...)
-	if err != nil {
-		validator.recordError(err.Error())
-	}
-	if warning != nil {
-		validator.recordWarning(*warning)
+	configError := ValidateIdentifier(plan.Name, validator.context...)
+	if configError != nil {
+		validator.recordError(configError.Error())
 	}
 
 	if plan.Config == nil && plan.ConfigPath == "" {
@@ -81,14 +74,14 @@ func (validator *StepValidator) VisitTask(plan *TaskStep) error {
 	}
 
 	if plan.Config != nil && (plan.Config.RootfsURI != "" || plan.Config.ImageResource != nil) && plan.ImageArtifactName != "" {
-		validator.recordWarning(ConfigWarning{
+		validator.recordConfigErrors(ConfigErrors{
 			Type:    "pipeline",
 			Message: validator.annotate("specifies image: on the step but also specifies an image under config: - the image: on the step takes precedence"),
 		})
 	}
 
 	if plan.Hermetic {
-		validator.recordWarning(ConfigWarning{
+		validator.recordConfigErrors(ConfigErrors{
 			Type:    "pipeline",
 			Message: validator.annotate("specifies `hermetic:` only works against worker containerd runtime"),
 		})
@@ -117,12 +110,9 @@ func (validator *StepValidator) VisitGet(step *GetStep) error {
 	validator.pushContextf(".get(%s)", step.Name)
 	defer validator.popContext()
 
-	warning, err := ValidateIdentifier(step.Name, validator.context...)
-	if err != nil {
-		validator.recordError(err.Error())
-	}
-	if warning != nil {
-		validator.recordWarning(*warning)
+	configError := ValidateIdentifier(step.Name, validator.context...)
+	if configError != nil {
+		validator.recordError(configError.Error())
 	}
 
 	if validator.seenGetName[step.Name] {
@@ -178,14 +168,10 @@ func (validator *StepValidator) VisitPut(step *PutStep) error {
 	validator.pushContextf(".put(%s)", step.Name)
 	defer validator.popContext()
 
-	warning, err := ValidateIdentifier(step.Name, validator.context...)
-	if err != nil {
-		validator.recordError(err.Error())
+	configError := ValidateIdentifier(step.Name, validator.context...)
+	if configError != nil {
+		validator.recordError(configError.Error())
 	}
-	if warning != nil {
-		validator.recordWarning(*warning)
-	}
-
 	resourceName := step.ResourceName()
 
 	_, found := validator.config.Resources.Lookup(resourceName)
@@ -200,15 +186,9 @@ func (validator *StepValidator) VisitRun(step *RunStep) error {
 	validator.pushContextf(".run(%s.%s)", step.Type, step.Message)
 	defer validator.popContext()
 
-	warning, err := ValidateIdentifier(step.Message, validator.context...)
-	if err != nil {
-		validator.recordError(err.Error())
-	}
-	if warning != nil {
-		// To prevent people from writing prototypes with invalid message
-		// names, we'll explicitly error on invalid identifiers rather than
-		// emitting a warning.
-		validator.recordError(warning.Message)
+	configError := ValidateIdentifier(step.Message, validator.context...)
+	if configError != nil {
+		validator.recordError(configError.Error())
 	}
 
 	_, found := validator.config.Prototypes.Lookup(step.Type)
@@ -223,12 +203,9 @@ func (validator *StepValidator) VisitSetPipeline(step *SetPipelineStep) error {
 	validator.pushContextf(".set_pipeline(%s)", step.Name)
 	defer validator.popContext()
 
-	warning, err := ValidateIdentifier(step.Name, validator.context...)
-	if err != nil {
-		validator.recordError(err.Error())
-	}
-	if warning != nil {
-		validator.recordWarning(*warning)
+	configError := ValidateIdentifier(step.Name, validator.context...)
+	if configError != nil {
+		validator.recordError(configError.Error())
 	}
 
 	if step.File == "" {
@@ -242,12 +219,9 @@ func (validator *StepValidator) VisitLoadVar(step *LoadVarStep) error {
 	validator.pushContextf(".load_var(%s)", step.Name)
 	defer validator.popContext()
 
-	warning, err := ValidateIdentifier(step.Name, validator.context...)
-	if err != nil {
-		validator.recordError(err.Error())
-	}
-	if warning != nil {
-		validator.recordWarning(*warning)
+	configError := ValidateIdentifier(step.Name, validator.context...)
+	if configError != nil {
+		validator.recordError(configError.Error())
 	}
 
 	validator.declareLocalVar(step.Name)
@@ -426,8 +400,8 @@ func (validator *StepValidator) VisitEnsure(step *EnsureStep) error {
 	return validator.Validate(step.Hook)
 }
 
-func (validator *StepValidator) recordWarning(warning ConfigWarning) {
-	validator.Warnings = append(validator.Warnings, warning)
+func (validator *StepValidator) recordConfigErrors(configError ConfigErrors) {
+	validator.ConfigErrors = append(validator.ConfigErrors, configError)
 }
 
 func (validator *StepValidator) recordError(message string) {
@@ -479,7 +453,7 @@ func (validator *StepValidator) declareLocalVar(name string) {
 	if validator.currentLocalVarScope()[name] {
 		validator.recordError("repeated var name")
 	} else if validator.localVarIsDeclared(name) {
-		validator.recordWarning(ConfigWarning{
+		validator.recordConfigErrors(ConfigErrors{
 			Type:    "var_shadowed",
 			Message: validator.annotate(fmt.Sprintf("shadows local var '%s'", name)),
 		})
--- a/atc/step_validator.go
+++ b/atc/step_validator.go
@@ -62,7 +62,7 @@ func (validator *StepValidator) VisitTask(plan *TaskStep) error {
 
 	configError := ValidateIdentifier(plan.Name, validator.context...)
 	if configError != nil {
-		validator.recordError(configError.Error())
+		validator.recordConfigErrors(*configError)
 	}
 
 	if plan.Config == nil && plan.ConfigPath == "" {
@@ -112,7 +112,7 @@ func (validator *StepValidator) VisitGet(step *GetStep) error {
 
 	configError := ValidateIdentifier(step.Name, validator.context...)
 	if configError != nil {
-		validator.recordError(configError.Error())
+		validator.recordConfigErrors(*configError)
 	}
 
 	if validator.seenGetName[step.Name] {
@@ -170,7 +170,7 @@ func (validator *StepValidator) VisitPut(step *PutStep) error {
 
 	configError := ValidateIdentifier(step.Name, validator.context...)
 	if configError != nil {
-		validator.recordError(configError.Error())
+		validator.recordConfigErrors(*configError)
 	}
 	resourceName := step.ResourceName()
 
@@ -205,7 +205,7 @@ func (validator *StepValidator) VisitSetPipeline(step *SetPipelineStep) error {
 
 	configError := ValidateIdentifier(step.Name, validator.context...)
 	if configError != nil {
-		validator.recordError(configError.Error())
+		validator.recordConfigErrors(*configError)
 	}
 
 	if step.File == "" {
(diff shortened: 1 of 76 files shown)


ceco556:
I have only on question here. Do we need a fallback mechanism for the delete-pipeline, delete-team and ge-pipeline?
If there are already set pipelines with invalid identifiers and try to delete them with this change it will return Error.

For me it would be good to have such for some time. @taylorsilva , what do you think?

ceco556:
@taylorsilva , could you please give me access to the PR checks so i can fix them?

ceco556:
Sorry for the commits, my forked repo was not synced.
If it is a problem i will open new PR.

taylorsilva:
Initial thought for the users that have pipelines with invalid identifiers, since I think the validation only happens client side, would be to add a flag to skip the validation check, `--skip-validation`.

This would be a final escape hatch for users that have ignored the warnings.

ceco556:
I will implement it. Thanks for the suggestion.

After log out and log in, still i don't see the PR checks. Other pipelines are visible now. Did i miss something?

ceco556:
Ah, yes, found it!
Thank you!

taylorsilva:
Ignore the `[containerd-integration](https://ci.concourse-ci.org/builds/464632242)` failure. It's flaking, I'll get it passing.

The unit test failures are legit though: https://ci.concourse-ci.org/builds/464630948#L6821cb41:627:654

ceco556:
The unit job timed out right at the end. I've retriggered it. Is it normal to time out ?

I think i managed to fix the unit test, now i have to take a look at the testflight tests.

taylorsilva:
Sadly the time out is normal. There's only one worker for all PR's right now which creates a bottleneck and these timeouts are one side effect. Hopefully we can get more PR workers soon.

ceco556:
While fixing the testflight tests i found one issue. If you try to set a pipeline with set_pipeline step in it such as:

```
jobs:
- name: sp-1
  public: true
  plan:
    - get: some-resource
    - set_pipeline: ((pipeline_name))
      file: some-resource/pipeline.yml
      var_files:
      - some-resource/name.yml
      vars:
        greetings: hello
```
It will return error because the name ((pipeline_name)) has invalid identifier - "(". 
My suggestion here is to change the regex to not check for "(". What do you think?

atc/step_validator.go
```diff
@@ -81,19 +74,12 @@ func (validator *StepValidator) VisitTask(plan *TaskStep) error {
...
 			Message: validator.annotate("specifies image: on the step but also specifies an image under config: - the image: on the step takes precedence"),
 		})
 	}
 
-	if plan.Hermetic {
-		validator.recordWarning(ConfigWarning{
-			Type:    "pipeline",
-			Message: validator.annotate("specifies `hermetic:` only works against worker containerd runtime"),
-		})
-	}
```
taylorsilva, line 95: oh this is unfortunate, but I don't think we should keep the warning stuff just for this one warning. We warn about this in the docs already.
I'm mostly commenting so folks know I saw this and accepted that we won't issue this warning for tasks that set `hermetic` anymore.

atc/step_validator.go
```diff
@@ -223,12 +196,9 @@ func (validator *StepValidator) VisitSetPipeline(step *SetPipelineStep) error {
...
 
-	warning, err := ValidateIdentifier(step.Name, validator.context...)
-	if err != nil {
-		validator.recordError(err.Error())
-	}
-	if warning != nil {
-		validator.recordWarning(*warning)
+	configError := ValidateIdentifier(step.Name, validator.context...)
+	if configError != nil {
+		validator.recordConfigErrors(*configError)
```
taylorsilva, line 201: I'm guessing it's here that we're getting the error when the step name is a var:
```yaml
- set_pipeline: ((pipeline_name))
  file: pipeline.yml
```

Let me know what you think of this. Before calling `ValidateIdentifier()` on the `step.Name`, check if it contains any vars in the name (you could re-use your regex here!). In that case we don't call `ValidateIdentifier()` and leave the pipeline name to be checked later during step runtime after the var(s) are resolved.

If the `step.Name` does not contain any vars, then we proceed with calling `ValidateIdentifier()` on it and returning any errors.

This allows us to provide users with early feedback about the pipeline name when they're not using vars, and still allowing users to use vars in the pipeline name if they choose to.

Wdyt?

atc/step_validator.go
taylorsilva, line 1: I think we an remove `ConfigErrors` and just use `Errors`. We could then remove the `recordConfigErrors()` function and only use `recordError()` everywhere.

fly/commands/internal/setpipelinehelpers/atc_config.go
```diff
@@ -28,7 +28,7 @@ type ATCConfig struct {
 	SkipInteraction  bool
 	CheckCredentials bool
 	DryRun           bool
-	CommandWarnings  []concourse.ConfigWarning
+	CommandErrors    []atc.ConfigErrors
```
taylorsilva, line 31: I don't think this field is will be needed anymore if you consolidate to just return `error` instead of using `[]atc.ConfigErrors` everywhere still.

fly/commands/set_pipeline.go
```diff
@@ -34,26 +35,37 @@ type SetPipelineCommand struct {
 	Team flaghelpers.TeamFlag `long:"team"              description:"Name of the team to which the pipeline belongs, if different from the target default"`
 }
 
-func (command *SetPipelineCommand) Validate() ([]concourse.ConfigWarning, error) {
-	var warnings []concourse.ConfigWarning
+func (command *SetPipelineCommand) Validate() ([]atc.ConfigErrors, error) {
+	var configErrors []atc.ConfigErrors
```
taylorsilva, line 39: dup: consolidate to return `error` only

```go
func (command *SetPipelineCommand) Validate() error {
```

fly/integration/error_handling_test.go
```diff
@@ -101,10 +101,8 @@ var _ = Describe("Fly CLI", func() {
 					),
 				)
 				flyCmd.Path = flyPath // idk why but the .Path is not getting set when we run exec.Command even though flyPath is available...
-				sess, err := gexec.Start(flyCmd, GinkgoWriter, GinkgoWriter)
-				Expect(err).NotTo(HaveOccurred())
-
-				Eventually(sess.Err.Contents).Should(ContainSubstring(`error: team 'doesnotexist' does not exist`))
```
taylorsilva, line 107: Can we keep the `ContainSubstring()` checks in these tests?

fly/integration/rename_pipeline_test.go
```diff
@@ -164,11 +163,10 @@ var _ = Describe("RenamePipeline", func() {
 			flyCmd := exec.Command(flyPath, "-t", targetName, "rename-pipeline", "-o", "some-pipeline", "-n", newName, "--team", "other-team")
 
 			sess, err := gexec.Start(flyCmd, GinkgoWriter, GinkgoWriter)
-			Expect(err).NotTo(HaveOccurred())
+			Expect(err).ToNot(HaveOccurred())
 
-			Eventually(sess).Should(gexec.Exit(0))
-			Expect(sess.Out).To(gbytes.Say(fmt.Sprintf("pipeline successfully renamed to '%s'", newName)))
```
taylorsilva, line 170: nit: can we keep the output check here?

go-concourse/concourse/teams_test.go
```diff
@@ -174,18 +174,15 @@ var _ = Describe("ATC Handler Teams", func() {
...
+					expectedResponse.Errors = []string{
+						"team: 'new venture' is not a valid identifier",
 					}
 				})
 
 				It("returns back false for created, and true for updated", func() {
 					_, _, _, warnings, err := team.CreateOrUpdate(desiredTeam)
 					Expect(err).NotTo(HaveOccurred())
-					Expect(warnings).To(HaveLen(1))
-					Expect(warnings[0].Message).To(ContainSubstring("team: 'new venture' is not a valid identifier"))
```
taylorsilva, line 188: nit: can we still check for the string?

taylorsilva (changes requested):
Overall we're on the right track here.

The majority of my feedback basically boils down to: delete more code! 😀 

I think anywhere where we were returning an `atc.ConfigWarning` can be consolidated into the `error` that was returned along with the `ConfigWarning`. That should result in code being deleted and simplify method signatures to return only `error` instead of warnings and errors.

Also great job adding the `SkipValidation` escape hatch for various fly commands 👏

atc/step_validator.go
ceco556: I think this is a good idea. 
Will fix it.

ceco556:
Hi @taylorsilva,

Thanks for the code review. While addressing your comments i encountered with the integration tests for fly.(set_pipeline_test.go)
I get a lot of errors for invalid groups, resources, etc. Should we raise errors for these cases?
If yes, could you please assist with the test. Let me know if you'd like to commit my latest changes.

taylorsilva:
Thanks @ceco556! I'm a bit backlogged right now so it'll take a while for me to circle back here, hopefully next week when you're back. Once I have time I'll dig into the tests.

This comment here also pointed out a similar issue that you ran into with the `set_pipeline` step, but for `task` steps as well: https://github.com/concourse/concourse/discussions/9184#discussioncomment-13348471

Not sure how we'll tackle this right now. We can dig into it more next week hopefully.

ceco556:
Hi @taylorsilva, did you have time to dig in this PR?

I have concerns regarding the failing tests and how we will handle the placeholders in the set_pipeline step. Even we skip the check for invalid identifiers in the set_pipeline step, there are a lot failing tests which have to be rewritten.

taylorsilva:
I haven't had a chance yet, but this PR has been in my head for the last few weeks. I have the same concerns as you and I'm not feeling good about the possible paths in front of us.

I've been feeling a bit tempted to have fields like the name of tasks not be considered `identifiers` at all and just skip checking them. The [RFC](https://github.com/concourse/rfcs/blob/master/040-valid-identifiers/proposal.md) does mention step names, but the supporting arguments only apply to job and resource names. Maybe we have a more lax check on step names, like making sure they don't have newline characters?

taylorsilva:
Heads up, I'm going to focus on the next release (7.14), so I won't get back to this PR until after that is done. This is top of the list though for me to review after the next release is out.

taylorsilva:
Hey, FINALLY circling back to this. So it seems our biggest issue is with handling vars `(( ))` in identifiers. You [previously shared](https://github.com/concourse/concourse/pull/9202#issuecomment-2898603607) this example:

```yaml
jobs:
- name: sp-1
  public: true
  plan:
    - get: some-resource
    - set_pipeline: ((pipeline_name))
      file: some-resource/pipeline.yml
      var_files:
      - some-resource/name.yml
      vars:
        greetings: hello
```

What confuses me about this is: when is `((pipeline_name))` expected to be interpolated?
I _think_ `((pipeline_name))` would be interpolated when this pipeline is set. If I make your example a bit more portable:

```yaml
resources:
- name: mock
  type: mock
  source:
    create_files:
      pipeline.yml: |
        resources:
          - name: mock
            type: mock
        jobs:
          - name: job
            plan:
              - get: mock

jobs:
- name: sp
  plan:
    - get: mock
    - set_pipeline: ((pipeline_name))
      file: mock/pipeline.yml
```
and `fly set-pipeline` and run it, I get an error `undefined vars: pipeline_name`
image
This error is interesting for other reasons, but the main point for now is that it fails.

If I set this pipeline again with `--var pipeline_name=sp-test-ch` set, everything works:
image

What workflow were you conveying when you shared this pipeline?

---

I decided to see where vars currently work through out a pipeline. I think I found some bugs related to how and when Concourse interpolates vars, but I think we can get away with ignoring those for now for the sake of this PR.

First, there are three "types" of vars that Concourse can interpolate:

- `(( ))` - the classic. Can be interpolated when a pipeline is being set and used anywhere in a pipeline. Can also be interpolated during runtime, where the value is then expected to come from the configured secrets manager. Runtime interpolation is limited though (e.g. step names cannot be interpolated, but a `task.config` can be)
- `((.:))` - the "local" var source, created by `load_var` and `across` steps
- `((src:))` - var sources, come from `var_sources` defined in a pipeline

I created the following pipeline to play around with these different types of vars in various situations.

<details>
<summary>Click to view pipeline</summary>

```yaml
var_sources:
  - name: varsrc
    type: dummy
    config:
      vars:
        task-name: my-vars
        child-pipe: child_pipe

resources:
  - name: mock
    type: mock

jobs:
  - name: job
    plan:
      - get: image
        resource: mock
        params:
          mirror_self_via_params: true

      - get: mock
        params:
          create_files_via_params:
            vars.yml: |
              task_name: some_task_name
              invalid_task_name: invalid task name
              child-pipe: child_pipe
            pipe.yml: |
              resources:
                - name: mock
                  type: mock
              jobs:
                - name: job
                  plan:
                    - get: mock

      - load_var: my-vars
        reveal: true
        file: mock/vars.yml

      - across:
          - var: acr
            values:
              - 123
              - abc
        in_parallel:
          # is currently allowed and should remain so
          # should we error if the identifier is not valid???
          - task: ((.:acr))
            image: image
            config:
              platform: linux
              run:
                path: bash
                args:
                  - -ec
                  # referncing var_sources does not work ((varsrc:child-pipe)) inside across steps
                  # results in error: missing source 'varsrc' in var: varsrc:child-pipe
                  # referencing ((.:my-vars)) does work! it is interpolated
                  - |
                    echo ((.:my-vars.child-pipe))-((.:acr))
          # does not warn, will create a pipeline with the literal var `(( ))` as the name
          # ((.:my-vars)) is not interpolatd
          # ((.:acr)) is interpolated
          # ((varsrc:)) results in an error during runtime: missing source 'varsrc'. I think this means lookups from secrets manager may also fail...
          - set_pipeline: ((.:my-vars.child-pipe))-((.:acr))
            file: mock/pipe.yml

      # currently warns, and should not be allowed moving forward
      - task: ((.:task-name))
        image: image
        config:
          platform: linux
          run:
            path: bash
            args:
              - -ec
              - |
                echo ((.:my-vars.child-pipe))-((varsrc:child-pipe))

      # it seems we detect the var, but it's never actually interpolated. We end
      # up with a bad pipeline name. The step should error.
      - set_pipeline: ((.:my-vars.child-pipe))
        file: mock/pipe.yml
```

</details>

There are a number of different cases we have to deal with here. Overall, I found the variable interpolation within `across` steps isn't working as expected. There are some bugs here. Thankfully, the core part works as expected and I think the fact that no one has logged any of the bugs I just found means I've stumbled across a very unusual use-case that we can ignore for now. Read my comments in the pipeline above for details on what works and doesn't work throughout the pipeline.

What does this all mean for this PR? There are various places where "validation" happens and we can check identifiers. I think when setting a pipeline, we can ignore any vars found inside `across` steps. Any vars found in identifiers outside `across` steps should result in an error.

During runtime, I think I would like to see the `across` properly validate identifiers for `task` and `set_pipeline` steps. How both of those are handled may be different. My expertise in that area of the code base is a bit foggy. I consider this part a bit out of scope for this PR, feels a bit like scope-creep potentially.

Those are my thoughts right now. I'm interested in your response to the first section of my post here. Let me know your thoughts! Sorry for the longggg post 😬

ceco556:
Hi @taylorsilva, 

I am sorry for the late response. Last couple of months were very interesting for me and i didn't have much time to work on this. I am not sure i will be able to continue the work on this PR next weeks, maybe months.

From what you've said, i totally agree with that to validate only what is outside of the across step. My only concern here is the tests. There are a tons of failing tests after the requested changes from you which must be fixed to handle these errors. I am not the best GO test developer and i am not sure i can handle them 😄  I will be very grateful for any help on this PR.

I am totally fine if someone wants to continue the work and fix the tests. Let me know what do you think.

taylorsilva:
Totally cool if you want to drop this PR. It is a tricky issue and was an informative exercise for the project. Feel free to close the PR.

PR #9208, opened 2025-05-27
Skip renewal for non-renewable Vault tokens

IvanChalukov:

This PR, developed in collaboration with @Kump3r, addresses scenarios where Vault tokens cannot be self-renewed, such as [Batch tokens](https://developer.hashicorp.com/vault/docs/concepts/tokens#batch-tokens). In our environment, over 2,000 pipelines across 60 teams use Batch Vault tokens to collect information. This results in millions of renewal attempts each month, all of which fail because Batch tokens are non-renewable.

* [x] ensure that vault token is renewable

## Notes to reviewer
Service tokens and Batch tokens comparison could be found in [Token type comparison](https://developer.hashicorp.com/vault/docs/concepts/tokens#token-type-comparison)

## Release Note

* Improved Vault token renewal logic to skip non-renewable tokens like Batch tokens, reducing unnecessary renewal attempts.

### DIFF
--- a/atc/creds/vault/api_client.go
+++ b/atc/creds/vault/api_client.go
@@ -184,6 +184,31 @@ func (ac *APIClient) Renew() (time.Duration, error) {
 
 	client := ac.client()
 
+	// Retrieve token information to determine if it is renewable before attempting renewal.
+	tokenInfo, err := client.Auth().Token().LookupSelf()
+	if err != nil {
+		logger.Error("failed-to-lookup-token", err)
+		return time.Second, err
+	}
+
+	if tokenInfo == nil {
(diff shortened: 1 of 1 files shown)


Kump3r (approved):
I tested the implementation with both Batch and service tokens:
1. `batch` tokens are skipped after the lookup and never try to be renewed. After restart of the web additional lookup is done for each token, but again only once and no renew requests are done
2. `service` tokens continue to work as before and we see a successful renew info log in the output logs. No further requests are made until the time duration requires a new retry request.
We haven't executed any integration/unit tests as we failed to find such. Let us know if we just missed them and they need to be addressed.
Good job @IvanChalukov and thanks for the PR!

taylorsilva (approved):

PR #9214, opened 2025-06-18
Restrict algorithms for host key exchange

### LINKED ISSUE #9212: TSA offers "ssh-rsa" as host key algorithm
## Summary

"ssh-rsa" is offered as host key algorithm by TSA for worker registration on port 2222. This seems to be flagged at least by some security scanners as "using deprecated SHA1 cryptographic settings to communicate."

This is a bit misleading and I think conflates the host key and signing algorithm used for the hashes. And I'm not sure SHA1 is even used when using go crypto even though client requests it (it might default to the sha2 versions anyway, I think I read something along these lines in some github issue, but haven't traced the code to the very end).

But I do think this will be flagged in more contexts going forward. And whether it's a false positive or not, well, either way it has to be addressed somehow.

I am guessing the particular scanner uses something like this under the hood:

```
nmap -Pn --script ssh2-enum-algos -sV -p 2222 tsa.example.com
```

This returns (for 7.13.1):

```
|   server_host_key_algorithms: (3)
|       rsa-sha2-256
|       rsa-sha2-512
|       ssh-rsa
```

And the "ssh-rsa" is what is suspicious.

I do think this can be addressed by explicitly allowing sha2-256 and sha2-512 only, this rather straight forward:

```
-	config.AddHostKey(signer)
+	signerWithAlgorithms, err := ssh.NewSignerWithAlgorithms(signer.(ssh.AlgorithmSigner), []string{ssh.KeyAlgoRSASHA256, ssh.KeyAlgoRSASHA512})
+	if err != nil {
+		log.Fatal("Failed to create private key with restricted algorithms: ", err)
+	}
+
+	config.AddHostKey(signerWithAlgorithms)
```

Problem is explained in: https://github.com/golang/go/issues/52132

Results in:

```
|   server_host_key_algorithms: (2)
|       rsa-sha2-256
|       rsa-sha2-512
```

In my experiments with quickstart this causes no problems (both use the same client library so we know they support sha2-xxx). 

Is it worth it if I submit a pr for this? 

Theoretically there could be issues when connecting from truly ancient worker versions (that actually rely on ssh-rsa).  
For that case I can imagine allowing a list of key signing algorithms as a parameter(so one could enable it explicitly), and defaulting to sha2-256 and sha2-512 only?

## Context

* RFC: concourse/rfcs#
* Prior discussion: #
* Depends on #
* Part of #

neumayer:
Currently, "ssh-rsa" is offered as host key algorithm by TSA for worker registration on port 2222. This seems to be flagged at least by some security scanners as "using deprecated SHA1 cryptographic settings to communicate."

This can be verified with:

nmap -Pn --script ssh2-enum-algos -sV -p 2222 tsa.example.com

And results in:

|   server_host_key_algorithms: (3)
|       rsa-sha2-256
|       rsa-sha2-512
|       ssh-rsa

This can be addressed by specifying the allowed host key algorithms on the server side as outlined here:

https://github.com/golang/go/issues/52132

closes https://github.com/concourse/concourse/issues/9212

### DIFF
--- a/tsa/tsacmd/command.go
+++ b/tsa/tsacmd/command.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"log"
 	"net/http"
 	"os"
 	"sync"
@@ -281,7 +280,7 @@ func (cmd *TSACommand) configureSSHServer(sessionAuthTeam *sessionTeam, authoriz
 
 	signerWithAlgorithms, err := ssh.NewSignerWithAlgorithms(signer.(ssh.AlgorithmSigner), []string{ssh.KeyAlgoRSASHA256, ssh.KeyAlgoRSASHA512})
 	if err != nil {
-		log.Fatal("Failed to create private key with restricted algorithms: ", err)
+		return nil, fmt.Errorf("failed to create private key with restricted algorithms: %s", err)
 	}
 
 	config.AddHostKey(signerWithAlgorithms)
--- a/tsa/tsacmd/command.go
+++ b/tsa/tsacmd/command.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"log"
 	"net/http"
 	"os"
 	"sync"
@@ -278,7 +279,12 @@ func (cmd *TSACommand) configureSSHServer(sessionAuthTeam *sessionTeam, authoriz
 		return nil, fmt.Errorf("failed to create signer from host key: %s", err)
 	}
 
-	config.AddHostKey(signer)
+	signerWithAlgorithms, err := ssh.NewSignerWithAlgorithms(signer.(ssh.AlgorithmSigner), []string{ssh.KeyAlgoRSASHA256, ssh.KeyAlgoRSASHA512})
+	if err != nil {
+		log.Fatal("Failed to create private key with restricted algorithms: ", err)
+	}
+
+	config.AddHostKey(signerWithAlgorithms)
 
 	return config, nil
 }


tsa/tsacmd/command.go
```diff
@@ -278,7 +279,12 @@ func (cmd *TSACommand) configureSSHServer(sessionAuthTeam *sessionTeam, authoriz
 		return nil, fmt.Errorf("failed to create signer from host key: %s", err)
 	}
 
-	config.AddHostKey(signer)
+	signerWithAlgorithms, err := ssh.NewSignerWithAlgorithms(signer.(ssh.AlgorithmSigner), []string{ssh.KeyAlgoRSASHA256, ssh.KeyAlgoRSASHA512})
+	if err != nil {
+		log.Fatal("Failed to create private key with restricted algorithms: ", err)
```
taylorsilva: Instead of `log.Fatal()`, return `nil, fmt.Errorf()`. See line 279 for an example. I'll merge once the change is made.

taylorsilva (changes requested):
FYI, I rebased your branch onto `master`. One of the PR checks was failing, seemingly due to a change in `goimport`'s behaviour.

tsa/tsacmd/command.go
neumayer: Done.

neumayer:
I can't access the logs of that failure. But in line with your comment before, I assume it's unrelated to my change.

taylorsilva:
You are correct. There's been a single test in that test suite that's been flaking in CI when there's high CPU usage. Re-running it.

Edit: Yup, it's green now.

taylorsilva (approved):
Approved! Just going to let the integration tests go green and then I'll merge. Thanks again for the PR 🙏

taylorsilva:
@neumayer I added you to the `all` team in the Concourse org here, you should get an email from Github to accept the invite. That'll allow you to see the logs from the PR checks. Will make any future contributions you make easier :)

PR #9224, opened 2025-06-30
Improve idtoken credential provider compatibility (with AWS and Azure)

dbaumgarten:

While working on the documentation for the new idtoken credential provider (and testing the provided examples) I discovered that it has compatibility issues with AWS and Azure.

This PR aims to fix these.

- Azure requires the key-usage of the JWKs to be "sig". Currently it is set to "sign" (which is incorrect according to the [RFC](https://datatracker.ietf.org/doc/html/rfc7517#section-4.2) )

- AWS requires more details in the OIDC-Discovery-Endpoint: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_oidc.html#manage-oidc-provider-prerequisites

With the changes proposed in this PR, compatibility with AWS and Azure is ensured and the examples provided in the upcoming docs ( https://github.com/concourse/docs/pull/563 ) work flawlessly.

### DIFF
--- a/atc/api/idtokenserver/openid_discovery.go
+++ b/atc/api/idtokenserver/openid_discovery.go
@@ -3,18 +3,28 @@ package idtokenserver
 import (
 	"encoding/json"
 	"net/http"
+
+	"github.com/go-jose/go-jose/v3"
 )
 
 func (s *Server) OpenIDConfiguration(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


dbaumgarten:
@taylorsilva Would you please take a look at this? Thanks :)

taylorsilva (approved):
LGTM - I'll get to the docs next. Thanks for doing that extra integration testing!

PR #9238, opened 2025-07-23
containerd: add flag to add additional-hosts

Kump3r:

Based on #6549, we want to maintain the adopt the functionally of adding additional DNS host entries to the `/etc/hosts` file of the container.

* [x] main functionallity
* [x] docker-compose local test of the env variable `CONCOURSE_CONTAINERD_ADDITIONAL_HOSTS`
* [x] unit/integration tests
* [x] bosh deployment introduction of env variable - https://github.com/concourse/concourse-bosh-release/pull/185
* [x] helm chart introduction of env variable - https://github.com/concourse/concourse-chart/pull/381
* [x] documentation in the runtime section of configuring worker - https://github.com/concourse/docs/pull/566

## Notes to reviewer
N/A

### DIFF
--- a/worker/runtime/cni_network.go
+++ b/worker/runtime/cni_network.go
@@ -287,6 +287,13 @@ func WithRestrictedNetworks(restrictedNetworks []string) CNINetworkOpt {
 	}
 }
 
+// WithAdditionalHosts defines the additional hosts that will be added to the /etc/hosts file in containers.
+func WithAdditionalHosts(additionalHosts []string) CNINetworkOpt {
+	return func(n *cniNetwork) {
+		n.additionalHosts = additionalHosts
+	}
+}
+
 // WithAllowHostAccess allows containers to talk to the host
 func WithAllowHostAccess() CNINetworkOpt {
 	return func(n *cniNetwork) {
@@ -319,6 +326,7 @@ type cniNetwork struct {
 	store              FileStore
 	config             CNINetworkConfig
 	nameServers        []string
+	additionalHosts    []string
 	binariesDir        string
 	restrictedNetworks []string
 	allowHostAccess    bool
@@ -406,6 +414,10 @@ func (n cniNetwork) SetupMounts(handle string) ([]specs.Mount, error) {
 		return nil, fmt.Errorf("creating /etc/hosts: %w", err)
 	}
 
+	// Adding Additional hosts as in guardian original implementation
+	// https://github.com/cloudfoundry/guardian/blob/main/kawasaki/dns/hosts_file_compiler.go
+	n.addHostsFileEntries(handle)
+
 	etcHostName, err := n.store.Create(
 		filepath.Join(handle, "/hostname"),
 		[]byte(handle+"\n"),
@@ -483,6 +495,17 @@ func (n cniNetwork) generateResolvConfContents() ([]byte, error) {
 	contents = strings.Join(resolvConfEntries, "\n") + "\n"
 
 	return []byte(contents), err
+}
+
+func (n cniNetwork) addHostsFileEntries(handle string) {
+	if len(n.additionalHosts) > 0 {
+		for _, entry := range n.additionalHosts {
+			err := n.store.Append(filepath.Join(handle, "/hosts"), []byte(entry+"\n"))
+			if err != nil {
+				fmt.Errorf("appending entry to /etc/hosts: %w", err)
+			}
+		}
+	}
 }
 
 func (n cniNetwork) restrictHostAccess() error {
--- a/worker/runtime/cni_network.go
+++ b/worker/runtime/cni_network.go
@@ -500,10 +500,7 @@ func (n cniNetwork) generateResolvConfContents() ([]byte, error) {
 func (n cniNetwork) addHostsFileEntries(handle string) {
 	if len(n.additionalHosts) > 0 {
 		for _, entry := range n.additionalHosts {
-			err := n.store.Append(filepath.Join(handle, "/hosts"), []byte(entry+"\n"))
-			if err != nil {
-				fmt.Errorf("appending entry to /etc/hosts: %w", err)
-			}
+			n.store.Append(filepath.Join(handle, "/hosts"), []byte(entry+"\n"))
 		}
 	}
 }
--- a/worker/runtime/cni_network.go
+++ b/worker/runtime/cni_network.go
@@ -505,7 +505,7 @@ func (n cniNetwork) addHostsFileEntries(handle string) error {
 		if len(fields) < 2 {
 			return fmt.Errorf("invalid host entry %q: must have IP and hostname", entry)
 		}
-		ip := fields[1]
+		ip := fields[0]
 		if net.ParseIP(ip) == nil {
 			return fmt.Errorf("invalid IP in host entry: %q", entry)
 		}
--- a/worker/runtime/cni_network.go
+++ b/worker/runtime/cni_network.go
@@ -503,7 +503,7 @@ func (n cniNetwork) addHostsFileEntries(handle string) error {
 	for _, entry := range n.additionalHosts {
 		fields := strings.Fields(entry)
 		if len(fields) < 2 {
-			return fmt.Errorf("invalid host entry %q: must have IP and hostname", entry)
+			return fmt.Errorf("invalid host entry %q: must have IP and hostname separated by a space", entry)
 		}
 		ip := fields[0]
 		if net.ParseIP(ip) == nil {
--- a/worker/runtime/cni_network.go
+++ b/worker/runtime/cni_network.go
@@ -505,9 +505,9 @@ func (n cniNetwork) addHostsFileEntries(handle string) error {
 		if len(fields) < 2 {
 			return fmt.Errorf("invalid host entry %q: must have IP and hostname", entry)
 		}
-		ip := fields[0]
+		ip := fields[1]
 		if net.ParseIP(ip) == nil {
-			return fmt.Errorf("invalid IP in host entry %q", entry)
+			return fmt.Errorf("invalid IP in host entry: %q", entry)
 		}
 		if err := n.store.Append(filepath.Join(handle, "hosts"), []byte(entry+"\n")); err != nil {
 			return fmt.Errorf("failed to append host entry %q: %w", entry, err)
--- a/worker/runtime/cni_network.go
+++ b/worker/runtime/cni_network.go
@@ -416,7 +416,9 @@ func (n cniNetwork) SetupMounts(handle string) ([]specs.Mount, error) {
 
 	// Adding Additional hosts as in guardian original implementation
 	// https://github.com/cloudfoundry/guardian/blob/main/kawasaki/dns/hosts_file_compiler.go
-	n.addHostsFileEntries(handle)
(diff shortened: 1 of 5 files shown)


taylorsilva (commented):
Wow, so easy! Was looking to see when `/etc/hosts` gets added to the file store and found it here: https://github.com/concourse/concourse/blob/8419676d0b3b4c81130ec0d03cefb9fe08ccf976/worker/runtime/cni_network.go#L401-L407

This looks good! Add the tests and this will be good to merge 🚀

taylorsilva:
Thanks for doing all the packaging stuff as well! ❤️

worker/runtime/cni_network.go
```diff
@@ -485,6 +497,14 @@ func (n cniNetwork) generateResolvConfContents() ([]byte, error) {
 	return []byte(contents), err
 }
 
+func (n cniNetwork) addHostsFileEntries(handle string) {
+	if len(n.additionalHosts) > 0 {
+		for _, entry := range n.additionalHosts {
+			n.store.Append(filepath.Join(handle, "/hosts"), []byte(entry+"\n"))
+		}
+	}
+}
```
IvanChalukov: You can skip the `len(additionalHosts)` check - the for loop is safe to run on an empty slice and will simply do nothing.

Also, it's worth handling the error returned from `FileStore.Append`. Not doing so could lead to silent failures when writing to the `/etc/hosts` file.
```suggestion
func (n cniNetwork) addHostsFileEntries(handle string) error {
	for _, entry := range n.additionalHosts {
		if err := n.store.Append(filepath.Join(handle, "hosts"), []byte(entry+"\n")); err != nil {
			return fmt.Errorf("failed to append host entry %q: %w", entry, err)
		}
	}
	return nil
}
```

worker/runtime/cni_network.go
```diff
@@ -485,6 +497,14 @@ func (n cniNetwork) generateResolvConfContents() ([]byte, error) {
...
+
```
IvanChalukov, line 518: Another thing that might be worth considering: validating the additionalHosts entries before writing them(as its done in `/etc/resolv.conf`). For example, each entry could be split into IP and hostname parts, and validate the IP using `net.ParseIP`. This would help avoid malformed entries being written. wdyt?

worker/runtime/cni_network.go
```diff
@@ -406,6 +414,10 @@ func (n cniNetwork) SetupMounts(handle string) ([]specs.Mount, error) {
 		return nil, fmt.Errorf("creating /etc/hosts: %w", err)
 	}
 
+	// Adding Additional hosts as in guardian original implementation
+	// https://github.com/cloudfoundry/guardian/blob/main/kawasaki/dns/hosts_file_compiler.go
+	n.addHostsFileEntries(handle)
```
IvanChalukov: Handle potential error returned by `addHostsFileEntries`:

```suggestion
	if err := n.addHostsFileEntries(handle); err != nil {
		return nil, fmt.Errorf("adding additional hosts: %w", err)
	}
```

worker/runtime/cni_network_test.go
```diff
@@ -176,6 +176,38 @@ func (s *CNINetworkSuite) TestSetupMountsCallsStoreWithoutNameServers() {
...
+	_, err = network.SetupMounts("some-handle")
+	s.NoError(err)
+
+	_, firstHost := s.store.AppendArgsForCall(0)
+	_, secondHost := s.store.AppendArgsForCall(1)
+
+	s.Equal(firstHost, []byte("example 1.2.3.4\n"))
+	s.Equal(secondHost, []byte("another-example 5.6.7.8\n"))
+}
+
```
IvanChalukov, line 197: Add test if `addHostsFileEntries` fails to append additional host entry.

IvanChalukov (commented):
I’ve tested this locally and it behaves as expected. Nice job! I’ve added a few small suggestions around error handling to help make it a bit more resilient.

worker/runtime/cni_network.go
Kump3r: Thought about that, but is there a point raising an error, should the addtional_hosts append fail. You would still have the hosts from the create? Maybe you are right was just wondering on that and decided to start without it and adjust based on the review

worker/runtime/cni_network.go
Kump3r: I like it, I opted against it based on the [guardian approach](https://github.com/cloudfoundry/guardian/blob/main/kawasaki/dns/hosts_file_compiler.go) and again, wanted to see the review

worker/runtime/cni_network.go
IvanChalukov: I see where you're coming from, and it makes sense to start simple. That said, I do think it's worth surfacing the error if an additional host fails to write - it likely points to a misconfiguration, and catching it early can save debugging time later from my perspective.

Kump3r:
Thanks for your feedback, we considered all of them previously as I mentioned, so it was a no brainer in adding them. It was just after a long day, so I totally messed up the previous unit tests 😅

worker/runtime/cni_network.go
```diff
@@ -485,6 +499,23 @@ func (n cniNetwork) generateResolvConfContents() ([]byte, error) {
 	return []byte(contents), err
 }
 
+func (n cniNetwork) addHostsFileEntries(handle string) error {
+	for _, entry := range n.additionalHosts {
+		fields := strings.Fields(entry)
+		if len(fields) < 2 {
+			return fmt.Errorf("invalid host entry %q: must have IP and hostname", entry)
+		}
+		ip := fields[1]
```
IvanChalukov: First entry of the list is the IP. 
```suggestion
		ip := fields[0]
```

worker/runtime/cni_network_test.go
```diff
@@ -176,6 +176,59 @@ func (s *CNINetworkSuite) TestSetupMountsCallsStoreWithoutNameServers() {
 	s.Equal(resolvConfContents, []byte(contents))
 }
 
+func (s *CNINetworkSuite) TestSetupMountsCallsStoreWithAdditionalHosts() {
+	network, err := runtime.NewCNINetwork(
+		runtime.WithDefaultsForTesting(),
+		runtime.WithCNIFileStore(s.store),
+		runtime.WithAdditionalHosts([]string{"example 1.2.3.4", "another-example 5.6.7.8"}),
```
IvanChalukov: Wrong host entry.
```suggestion
		runtime.WithAdditionalHosts([]string{"1.2.3.4 example", "5.6.7.8 another-example"}),
```

worker/runtime/cni_network_test.go
```diff
@@ -176,6 +176,59 @@ func (s *CNINetworkSuite) TestSetupMountsCallsStoreWithoutNameServers() {
...
+	s.NoError(err)
+
+	_, err = network.SetupMounts("some-handle")
+	s.NoError(err)
+
+	_, firstHost := s.store.AppendArgsForCall(0)
+	_, secondHost := s.store.AppendArgsForCall(1)
+
+	s.Equal(firstHost, []byte("example 1.2.3.4\n"))
+	s.Equal(secondHost, []byte("another-example 5.6.7.8\n"))
```
IvanChalukov: Update assert.
```suggestion
	s.Equal(firstHost, []byte("1.2.3.4 example\n"))
	s.Equal(secondHost, []byte("5.6.7.8 another-example\n"))
```

worker/runtime/cni_network_test.go
```diff
@@ -176,6 +176,59 @@ func (s *CNINetworkSuite) TestSetupMountsCallsStoreWithoutNameServers() {
...
+	s.NoError(err)
+
+	s.Equal(0, s.store.AppendCallCount())
+}
+
+func (s *CNINetworkSuite) TestSetupMountsFailsToAppendAdditionalHost() {
+	network, err := runtime.NewCNINetwork(
+		runtime.WithDefaultsForTesting(),
+		runtime.WithCNIFileStore(s.store),
+		runtime.WithAdditionalHosts([]string{"example 1.2.3.4"}),
```
IvanChalukov: Tests fails due to line 222: `s.store.AppendReturns(errors.New("append-error"))`. Host entry is not correct:
```suggestion
		runtime.WithAdditionalHosts([]string{"1.2.3.4 example"}),
```

worker/runtime/cni_network_test.go
```diff
@@ -176,6 +176,59 @@ func (s *CNINetworkSuite) TestSetupMountsCallsStoreWithoutNameServers() {
...
+
+	// Simulate failure on first Append call
+	s.store.AppendReturns(errors.New("append-error"))
+
+	// Act
+	_, err = network.SetupMounts("some-handle")
+
+	// Assert
+	s.Error(err)
+	s.EqualError(errors.Unwrap(err), "failed to append host entry \"example 1.2.3.4\": append-error")
```
IvanChalukov: Update assert:
```suggestion
	s.EqualError(errors.Unwrap(err), "failed to append host entry \"1.2.3.4 example\": append-error")

```

worker/runtime/cni_network_test.go
```diff
@@ -176,6 +176,59 @@ func (s *CNINetworkSuite) TestSetupMountsCallsStoreWithoutNameServers() {
...
+}
+
```
IvanChalukov, line 248: Add tests for invalid IP:
```suggestion
func (s *CNINetworkSuite) TestSetupMountsFailsWithInvalidIP() {
	network, err := runtime.NewCNINetwork(
		runtime.WithDefaultsForTesting(),
		runtime.WithCNIFileStore(s.store),
		runtime.WithAdditionalHosts([]string{"example.com 1.2.3.4.5"}),
		runtime.WithIptables(s.iptables),
	)
	s.NoError(err)

	_, err = network.SetupMounts("some-handle")

	s.Error(err)
	s.Contains(err.Error(), "invalid IP in host entry")
}
```

IvanChalukov (changes requested):

worker/runtime/cni_network_test.go
Kump3r: I will add a case for that

Kump3r:
You are absolutely right, when I was adjusting the tests, I switched the order and didn't think it matters as I forgot I already prepared the other packaging stuff. Nice catch!

IvanChalukov (approved):
Tested locally with the following configurations:
- No `CONCOURSE_CONTAINERD_ADDITIONAL_HOSTS` set → ✅ no changes to containers, behaves as expected
- Invalid values: `CONCOURSE_CONTAINERD_ADDITIONAL_HOSTS: 1.2.3.4 example.com,5.6.7.8.1 another-example.com` → ✅ fails to create container and print appropriate error message 
- Valid values: `CONCOURSE_CONTAINERD_ADDITIONAL_HOSTS: 1.2.3.4 example.com,5.6.7.8 another-example.com` → ✅ entries applied successfully to containers

Everything works as expected from my perspective - nice work!

taylorsilva (approved):
Quickly repeated Ivan's tests and got the same results. I added an additional commit that clarifies the format we expect the IP and hostname to be passed in with.

Will let the PR checks go through one last time and then merge.

Kump3r:
Thanks guys! Feel free to check the packaging stuff as they should be pretty straightforward as well.

PR #9240, opened 2025-07-25
Add /download-fly page for downloading the fly cli

taylorsilva:

Concourse initially shipped with 3 variants of fly, one for each OS. With the next release, Concourse will now ship with 5 variants of fly; arm64 variants for Linux and macOS.

This PR adds a new page to the web UI, accessible at `/download-fly`. This page contains a dropdown with all variants of fly that a user can download. Upon selecting a platform variant, a set of installation instructions is shown, along with a direct download link to fly.

This page replaces our previous "download fly" links, which were three icon links to the 3 OS-specific variants of fly.

I also took the opportunity to add a link to the Concourse docs to the footer.

## Footer
Before
image

After
image

## Welcome Card
Before
image

After
image

## Pipeline Footer
Before
image

After
image

## Download Page `/download-fly`

Initial view:
image

Linux Platform selected (macOS is the same):
image

Windows Platform selected. The instructions also configure the Windows PATH for the user:
image

## Fly Platform Options

I chose `x86_64` and `arm64` as the cpu arch labels. I think most technical users will recognize those two cpu architecture labels. I chose `x86_64` over `amd64` because lexically "amd64" is looks similar to "arm64" and I wanted to avoid users clicking the wrong platform and wondering why they get `exec format` errors.
image


analytically:
OSX intel vs Mac?

taylorsilva:
@analytically Are you talking about the naming of things in the dropdown? Not exactly sure what you mean by your statement or if you're suggesting a change...

taylorsilva:
I'm merging this now. Feel free to share any feedback here after I merge though. I'm happy to keep tweaking this page, though I think this current setup will be sufficient for everyone.

PR #9251, opened 2025-08-04
Fixing Prometheus Emitter for one-off builds

### LINKED ISSUE #9250: Pipeline attribute is not passed when using fly execute
## Summary

Pipeline is not **required** for builds, hence the metric emitter should not treat it as such: https://github.com/concourse/concourse/blob/2f6ceeef369c2f65d2205c6879d7ed8332a6c19f/atc/metric/emitter/prometheus.go#L1079-L1083

During a `fly execute`, no pipeline is required. However the emit event requires a pipeline.
## Steps to reproduce

Have concourse with prometheus and run:

```
cat <<EOF > "test.yml"
---
platform: linux
image_resource:
  type: registry-image
  source:
    repository: busybox
run:
  path: echo
  args: [hello world]
EOF
```
`fly execute --config ./test.yml`

## Expected results

No errors in web.stdout

## Actual results

```
{"timestamp":"2025-08-02T22:41:04.533900847Z","level":"error","source":"atc","message":"atc.tracker.notify.run.build-finished.emit.failed-to-find-pipeline-in-event","data":{"build":"1","build_id":1"error":"expected pipeline to exist in event.Attributes","session":"29.174.3.25.1","team":"main"}}
```

## Triaging info

* Concourse version: 7.13.1
* Did this used to work? - I am uncertain, possibly if the validations for team/pipeline etc. were added later on

Kump3r:

Fixing Prometheus Emitter for one-off builds with `fly execute` that do not require pipelineName and jobName
closes #9250 

## Tested code with:
```go
package main

import (
	"fmt"
	"time"
)

type Event struct {
	Name       string
	Value      float64
	Attributes map[string]string
	Host       string
	Time       time.Time
}

func main() {
	event := Event{
		Name:  "build finished",
		Value: 1,
			"team":     "team1",
			"status":   "succeeded",
			"pipeline": "test", // and without this line
		},
	}

	pipeline := ""
	if val, ok := event.Attributes["pipeline"]; ok {
		pipeline = val
	}
	fmt.Println(pipeline)
}

```

Also tested on a bosh Concourse with prometheus.


Kump3r:
Tests seem to be broken due to the development of release 7.14.0. Once the problems with the base image are fixed I will try and trigger the relevant jobs again.

taylorsilva:
Problem was actually with Garden, see https://github.com/concourse/concourse/pull/9252

Re-triggering jobs now.

Kump3r:
I merged master to the branch to resolve the failing test, ready for review and merge

taylorsilva (commented):
LGTM - I have one question, which may be a silly one: is there any issue with passing empty strings `""` as labels? That's the only big change I see happening with this PR and I don't want to break people's metrics. If you can answer that and it won't break things then I'm good to merge this.

Kump3r:
I haven’t seen issues with it. It just sets them as empty if the task is without a pipeline/job. We could always set them as ”undefined” or some other string, but imho that might be more misleading. Let me know. The metrics in the end look like in the test I introduced btw.

taylorsilva (approved):

PR #9273, opened 2025-08-29
Add broadcast message to Concourse cluster

### LINKED ISSUE #9241: Broadcast message to Concourse cluster
## Summary

This is a follow-up to #3636. We want a way for Concourse operators to broadcast a message to all users. Work for this was started in #5019, which added API endpoints for this feature. No further work has been done.

To finish this feature, we need to update both fly and the web UI.

### Fly Commands

The following fly commands should be added:

- `fly get-wall` - Fetches and prints the current message being broadcast, if any
- `fly set-wall` - Sets the contents of the Wall and optional expiry
- `fly clear-wall` - Clears the message currently being broadcast

(This is where I wish we had designed fly with a verb-based command syntax like kubectl)

### Web UI

The web UI should check the wall endpoint for any new messages. If a new message is found, its contents should be rendered as a banner at the top of the page until the TTL is hit or the message is cleared by the Concourse operator.

## Context

* Prior discussion: https://github.com/orgs/concourse/discussions/9239
* PR: https://github.com/concourse/concourse/pull/5019

ceco556:

closes #9241
closes #9239 

* [x] `fly set-wall` with required parameters `-m` as message and `--ttl` as time to live to set a broadcast message to appear on the top.
* [x] `fly get-wall` to get the already set broadcast message with it ttl.
* [x] `fly clear-wall` to clear the already set broadcast message.
* [x] Broadcast message appears at the top. After executing `fly set-wall` within few seconds the message will appear.
* [x] Executing `fly clear-wall` will remove the message of the top of the page within few seconds.
* [x] Working links if added to the broadcast message. If not it will appear as simple text at the top.
* [x] ~Added a warning icon before the message.~
* [x] Tests

## Notes to reviewer

- I've made the `--ttl` property for `fly set-wall` required, since I think it’s good to always specify it. If you feel it doesn’t need to be required, I can set a default value.
- I haven’t added tests yet—I’d like to get approval on the code before proceeding with them.
- I used a lighter gray as the container’s background color. I tried using orange, but it felt too distracting on the screen. That's why i added the warning icon.
- I am not sure how good is to have directly pasted icon in the elm code. I can remove it and replace it with an alternative if needed.

## Release Note
By using the `fly` commands `set-wall, get-wall, and clear-wall`, we can easily manage a broadcast message that appears at the top of the page.

broadcast_message

### DIFF
--- a/fly/commands/set_wall.go
+++ b/fly/commands/set_wall.go
@@ -9,8 +9,8 @@ import (
 )
 
 type SetWallCommand struct {
-	Message string        `short:"m" long:"message" required:"true" description:"Message to broadcast"`
-	TTL     time.Duration `long:"ttl" required:"true" description:"Time-to-live for the message (e.g. 1h30m)"`
+	Message string        `short:"m" long:"message" required:"true" description:"Message to broadcast. Supports emojis and links."`
+	TTL     time.Duration `long:"ttl" required:"false" description:"Time-to-live for the message (e.g. 1h30m). Zero values will result in a message with no expiration."`
 }
 
 func (command *SetWallCommand) Execute([]string) error {
@@ -35,7 +35,11 @@ func (command *SetWallCommand) Execute([]string) error {
 	}
 
 	fmt.Println("Wall message set successfully")
-	fmt.Printf("Message will expire in %v\n", command.TTL)
+	if command.TTL == 0 {
+		fmt.Println("No expiration set")
+	} else {
+		fmt.Printf("Message will expire in %v\n", command.TTL)
+	}
 
 	return nil
 }
--- a/fly/commands/set_wall.go
+++ b/fly/commands/set_wall.go
@@ -1,0 +1,41 @@
+package commands
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/fly/rc"
+)
+
+type SetWallCommand struct {
+	Message string        `short:"m" long:"message" required:"true" description:"Message to broadcast"`
+	TTL     time.Duration `long:"ttl" required:"true" description:"Time-to-live for the message (e.g. 1h30m)"`
+}
+
+func (command *SetWallCommand) Execute([]string) error {
+	target, err := rc.LoadTarget(Fly.Target, Fly.Verbose)
+	if err != nil {
+		return err
+	}
+
+	err = target.Validate()
+	if err != nil {
+		return err
+	}
+
+	wall := atc.Wall{
+		Message: command.Message,
+		TTL:     command.TTL,
+	}
+
+	err = target.Client().SetWall(wall)
+	if err != nil {
+		return fmt.Errorf("failed to set wall message: %w", err)
+	}
+
+	fmt.Println("Wall message set successfully")
+	fmt.Printf("Message will expire in %v\n", command.TTL)
+
+	return nil
+}
--- a/fly/integration/clear_wall_test.go
+++ b/fly/integration/clear_wall_test.go
@@ -1,0 +1,33 @@
+package integration_test
+
+import (
+	"os/exec"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+	"github.com/onsi/gomega/gexec"
+	"github.com/onsi/gomega/ghttp"
+)
+
+var _ = Describe("FLY CLI", func() {
+	Describe("clear-wall", func() {
+		Context("when clearing succeeds", func() {
+			It("sends DELETE /api/v1/wall and prints success", func() {
+				atcServer.AppendHandlers(
+					ghttp.CombineHandlers(
+						ghttp.VerifyRequest("DELETE", "/api/v1/wall"),
+						ghttp.RespondWith(200, ""),
+					),
+				)
+
+				flyCmd := exec.Command(flyPath, "-t", targetName, "clear-wall")
+				sess, err := gexec.Start(flyCmd, GinkgoWriter, GinkgoWriter)
+				Expect(err).NotTo(HaveOccurred())
+				<-sess.Exited
+				Expect(sess.ExitCode()).To(Equal(0))
+				Expect(sess.Out).To(gbytes.Say("Wall message cleared successfully"))
+			})
+		})
+	})
+})
--- a/fly/integration/clear_wall_test.go
+++ b/fly/integration/clear_wall_test.go
@@ -29,5 +29,22 @@ var _ = Describe("FLY CLI", func() {
 				Expect(sess.Out).To(gbytes.Say("Wall message cleared successfully"))
 			})
 		})
+
+		Context("when the API returns an error", func() {
+			It("exits non-zero and shows an error", func() {
+				atcServer.AppendHandlers(
+					ghttp.CombineHandlers(
+						ghttp.VerifyRequest("DELETE", "/api/v1/wall"),
+						ghttp.RespondWith(500, ""),
+					),
+				)
+				flyCmd := exec.Command(flyPath, "-t", targetName, "clear-wall")
+				sess, err := gexec.Start(flyCmd, GinkgoWriter, GinkgoWriter)
+				Expect(err).NotTo(HaveOccurred())
+				<-sess.Exited
+				Expect(sess.ExitCode()).To(Equal(1))
+				Expect(sess.Err).To(gbytes.Say("failed to clear wall message"))
+			})
+		})
 	})
 })
--- a/web/assets/css/wall.less
+++ b/web/assets/css/wall.less
@@ -1,7 +1,8 @@
 @import (reference) "_vars.less";
 
 #wall-banner {
-  position: fixed;
+  position: sticky;
+  padding-block: clamp(0.4rem, 1.2vw, 0.75rem);
   top: 55px;
   left: 0;
   right: 0;
@@ -9,10 +10,8 @@
   background-color: @grey70;
   color: #ffff;
   padding: 6px 16px;
-  font-size: 14px;
-  font-family: sans-serif;
+  font-size: 1.25em;
+  font-family: 'Inconsolata', monospace;
   text-align: center;
-  display: flex;
-  justify-content: center;
-  align-items: center;
 }
+
--- a/web/assets/css/wall.less
+++ b/web/assets/css/wall.less
@@ -16,8 +16,3 @@
   justify-content: center;
   align-items: center;
 }
-
-#wall-banner .wall-icon {
-  font-size: 16px;
-  margin-right: 4px;
-}
--- a/web/assets/css/wall.less
+++ b/web/assets/css/wall.less
@@ -1,0 +1,23 @@
+@import (reference) "_vars.less";
+
+#wall-banner {
+  position: fixed;
+  top: 55px;
+  left: 0;
+  right: 0;
+  z-index: 998;
+  background-color: @grey70;
+  color: #ffff;
+  padding: 6px 16px;
+  font-size: 14px;
+  font-family: sans-serif;
+  text-align: center;
+  display: flex;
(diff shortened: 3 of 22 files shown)


ceco556:
@taylorsilva  Thanks for reviewing the code! I’ve added the aliases and fixed the link issue.
Now starting work on the test.

fly/commands/set_wall.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+	"time"
+
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/fly/rc"
+)
+
+type SetWallCommand struct {
+	Message string        `short:"m" long:"message" required:"true" description:"Message to broadcast"`
+	TTL     time.Duration `long:"ttl" required:"true" description:"Time-to-live for the message (e.g. 1h30m)"`
```
analytically: why not a default for this, say 10m?

analytically: also, is an unlimited time possible? eg. "There was an issue processing your recent payment."

analytically: aka until manually cleared

web/elm/src/Application/Application.elm
```diff
@@ -454,12 +474,23 @@ view model =
             , SideBar.tooltip model.session
                 |> Maybe.map (Tooltip.view model.session)
                 |> Maybe.withDefault (Html.text "")
+            , case model.wallMessage of
+                Just msg ->
+                    Html.div [ id "wall-banner", style "white-space" "pre-wrap" ] <|
+                        Html.span [ Html.Attributes.class "wall-icon" ] [ Html.text "⚠️" ]
```
analytically: So a warning icon is the default?

fly/commands/set_wall.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/concourse/concourse/atc"
+	"github.com/concourse/concourse/fly/rc"
+)
+
+type SetWallCommand struct {
+	Message string        `short:"m" long:"message" required:"true" description:"Message to broadcast"`
```
analytically: Can the message be HTML?

fly/commands/set_wall.go
taylorsilva: `9999h` lol

taylorsilva: Decided to go look at the backend code again and found that it is setup to support messages with no expiration: https://github.com/concourse/concourse/blob/16c3ab7e4675f390dda8d06d489569e7d29743bb/atc/db/wall_test.go#L25-L40

So we probably should have a way of allowing users to set a message with no TTL. I tried setting a message with `--ttl 0s`. The UI never showed the message, but `fly get-wall` did return the message:

```
> fly -t l set-wall -m "some message" --ttl 0s
Wall message set successfully
Message will expire in 0s

> fly -t l get-wall
Wall Message: some message
Expires in: 0s
```

And it is correctly stores in the database with no expiration:
image

So we could:
1. Update the `--ttl` flag description to state that zero values (e.g. 0s) will result in a message with no expiration
2. When there is no expiration for a message, have `fly get-wall` state so (e.g. "This message will not expire")
3. Fix the web UI to show messages with no expiration

web/elm/src/Application/Application.elm
taylorsilva: I was wondering if we should force this icon upon all messages as well. I decided to try and see if you can pass in an emoji into `--message` and found that you can!

```
> fly -t l set-wall -m "❌ another emoji" --ttl 1h
> fly -t l get-wall
Wall Message: ❌ another emoji
Expires in: 59m53.649798028s
```

image

I think we should drop the ⚠️ emoji, gives users more flexibility with how they want their messages conveyed. The ⚠️ emoji will make it seem like every message is some kind of problem.

fly/commands/set_wall.go
taylorsilva: I think parsing links is the farthest we should go here. I'd get worried about XSS attacks if we allow users to inject random html onto the page.

taylorsilva (commented):
Summarizing some of the feedback and suggested changes from the comments left by @analytically, combined with my thoughts.

- Remove the ⚠️ warning icon from the message. Maybe update the description for the `--message` flag letting folks know they can pass emoji's into the message. Letting users know that links will be parsed would be nice too
- In the description for `--ttl`, let the user know that a zero value (e.g. 0s) will result in a message never expiring
  - Make sure the web UI properly shows message's with no expiration. Currently doesn't but `fly get-wall` does fetch the message, so I think the backend parts are all good.
  - See my comment here for more details: https://github.com/concourse/concourse/pull/9273/files#r2311013676

I'll look over the tests later. Thanks for all the work so far!

fly/commands/set_wall.go
Kump3r: Since the backend doesn’t require `ttl`, doesn’t it make more sense to be optional in fly as well? Just document it as an option and that is it.

ceco556:
Hello, apologies for the delay on the PR. 
I recently became a father to a baby boy on September 2nd. I’ll review the comments and start working on it as soon as I can.

ceco556:
Hello! Again, apologies for the delay on the PR. 

I've addressed the comments and works as expected. Let me know if you need something to change.

taylorsilva:
Thank you and welcome back! I'll probably get around to reviewing this again next week, it's near the top of my pile for PR's right now :)

fly/integration/clear_wall_test.go
```diff
@@ -0,0 +1,33 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+	"github.com/onsi/gomega/gexec"
+	"github.com/onsi/gomega/ghttp"
+)
+
+var _ = Describe("FLY CLI", func() {
+	Describe("clear-wall", func() {
+		Context("when clearing succeeds", func() {
```
taylorsilva, line 15: Add a test for when the API returns an error. You've written a similar test in `get_wall_test.go`

web/assets/css/wall.less
```diff
@@ -0,0 +1,18 @@
...
+  z-index: 998;
+  background-color: @grey70;
+  color: #ffff;
+  padding: 6px 16px;
+  font-size: 14px;
+  font-family: sans-serif;
+  text-align: center;
+  display: flex;
+  justify-content: center;
+  align-items: center;
```
taylorsilva: Found this visual bug when writing a long message that includes a link. The `display: flex` seems to be the cause of this.

```
fly -t l set-wall -m "Hello world! ✈️ 🌎. See https://concourse-ci.org/quick-start.html\#docker-compose-concourse and now I'm making a really long message just to see what happens when I make things overflow a bit even though most people will read their Concourse instances on desktops"
```

image

Renders the same way on Firefox and Chrome. I think we can remove these three CSS attributes to resolve the issue.
```suggestion
```

I don't know CSS well at all and not sure what your reasons were for adding `display: flex` was in the first place.

web/assets/css/wall.less
```diff
@@ -0,0 +1,18 @@
...
+  position: fixed;
+  top: 55px;
+  left: 0;
+  right: 0;
+  z-index: 998;
+  background-color: @grey70;
+  color: #ffff;
+  padding: 6px 16px;
+  font-size: 14px;
+  font-family: sans-serif;
```
taylorsilva:
```suggestion
  font-size: 1.25em;
  font-family: 'Inconsolata', monospace;
```

nit-picking here. We set font-family to `'Inconsolata', monospace` (almost) everywhere. Trying to keep the theme consistent.

web/assets/css/wall.less
```diff
@@ -0,0 +1,18 @@
+@import (reference) "_vars.less";
+
+#wall-banner {
```
taylorsilva, line 3: Some other CSS-relate things that need to be fixed.

## Build Logs & Job page
Wall message is clipping the top of the builds and jobs pages.

image

image

## Resource checks/versions page
Same issue as previous, header of checks page is being clipped.
image

## Dashboard
I'm a bit iffy about this one. I don't like the message breaking through the sidebar. Even if we leave it breaking through the sidebar, the `all pipelines` header doesn't have enough padding between itself and the message.

image

taylorsilva (changes requested):
Looking good so far. A few visual nit-picky things but then we'll be good.

It would be nice to add a few more Elm tests verifying the visual elements of the banner. See these tests for examples:
- https://github.com/concourse/concourse/blob/master/web/elm/tests/DownloadFlyTests.elm
- https://github.com/concourse/concourse/blob/master/web/elm/tests/JobTests.elm

The tricky part might be figuring out how to manipulate the model within the test. Once you do that though, writing the tests will be easy. I won't block the PR on this though. I'm putting it in the "nice to have" category.

web/elm/src/Application/Application.elm
```diff
@@ -121,6 +123,7 @@ init flags url =
       , LoadFavoritedPipelines
       , LoadFavoritedInstanceGroups
       , FetchClusterInfo
+    , FetchWall
```
analytically: spacing

fly/integration/clear_wall_test.go
ceco556: Tests added.

web/assets/css/wall.less
ceco556: I don't remember what was the idea, but i am pretty sure it is from testing some things.
Removed.

web/assets/css/wall.less
ceco556: Fixed.

web/assets/css/wall.less
ceco556: Nice catch! 
Fixed in the latest commit.

ceco556:
All comments addresed in my latest commit.
I've added few tests in the `ApplicationTests.elm` - when there is no message, when there is a message set and when WallFetched returns error. Let me know if there are more cases you would like to cover.

One thing i found in the `fly set-wall` command. When you pass an empty message, the command returns that the wall message was set and the banner doesn't appear in the UI. I think we should have a check in `set_wall.go` which will return error message if the message is empty. What do you think?

web/elm/src/Application/Application.elm
ceco556: Fixed.

taylorsilva:
I think that's fine to error if the message is empty. I would do this at the API layer instead though, then nothing else needs to change in fly: https://github.com/concourse/concourse/blob/615a5305c486551efe8cead20c2e2acd580eebbd/atc/api/wallserver/wall.go#L30

Can return HTTP 400. Something like this _might_ work:
https://github.com/concourse/concourse/blob/615a5305c486551efe8cead20c2e2acd580eebbd/atc/api/pipelineserver/order.go#L31-L32
and fly might just print the error returned? Give it a try, see what works :)

taylorsilva (commented):
This LGTM now that the CSS issues are resolved. Will wait for your final change on adding a check for empty messages before final approval and merge.

ceco556:
> I think that's fine to error if the message is empty. I would do this at the API layer instead though, then nothing else needs to change in fly: https://github.com/concourse/concourse/blob/615a5305c486551efe8cead20c2e2acd580eebbd/atc/api/wallserver/wall.go#L30
> 
> Can return HTTP 400. Something like this _might_ work:
> https://github.com/concourse/concourse/blob/615a5305c486551efe8cead20c2e2acd580eebbd/atc/api/pipelineserver/order.go#L31-L32
> and fly might just print the error returned? Give it a try, see what works :)

Addressed in my latest commit. If there is no message the API will return 400. 

I think the PR is ready. Let me know if there is something else to address.

taylorsilva (approved):
LGTM - I'm going to add an error message if the API detects an empty message, just so anyone encountering the error knows why it errored. Just HTTP 400 on its own is a tad unclear. I'll do this after merging though. Thank you SO MUCH for completing this feature! Operators are going to really appreciate it! 🎉

PR #9278, opened 2025-09-04
update migration number for switch_md5_to_sha256 [skip-migrations-check]

taylorsilva:

Related to #9165 

7.14 does not include this migration, but the latest migration it shipped with does have a higher migration number. This would result in the next release not applying the switch_md5_to_sha256 migration and then various errors occurring as the web node tries to query columns that don't exist.


Kump3r (approved):
Nice catch! Was there a migration in the minor version release?

taylorsilva:
There was, from https://github.com/concourse/concourse/pull/9035

PR #9285, opened 2025-09-06
Fix error logging level for missing resource in ListJobInputs

analytically:
Change log level from Debug to Error when a required resource cannot be found during job input listing. This is a legitimate error condition that should be visible in production logs for debugging failed requests.

A missing resource causes the endpoint to return 500 Internal Server Error, which warrants error-level logging rather than debug-level.


taylorsilva (approved):
LGTM

PR #9288, opened 2025-09-08
driver: use -R flag for recursive btrfs subvolume deletion

analytically:
Replace manual subvolume traversal with native recursive deletion using btrfs subvolume delete -R. This simplifies the code and uses the more efficient built-in recursive deletion mechanism.

### DIFF
--- a/worker/baggageclaim/volume/driver/btrfs.go
+++ b/worker/baggageclaim/volume/driver/btrfs.go
@@ -6,36 +6,49 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"code.cloudfoundry.org/lager/v3"
 	"github.com/concourse/concourse/worker/baggageclaim/volume"
 )
 
 type BtrFSDriver struct {
-	logger   lager.Logger
-	btrfsBin string
+	logger                lager.Logger
+	btrfsBin              string
+	supportsRecursiveFlag bool
 }
 
 func NewBtrFSDriver(
 	logger lager.Logger,
 	btrfsBin string,
 ) *BtrFSDriver {
-	return &BtrFSDriver{
+	driver := &BtrFSDriver{
 		logger:   logger,
 		btrfsBin: btrfsBin,
 	}
+
+	// Check if btrfs supports -R flag for recursive deletion
+	driver.supportsRecursiveFlag = driver.checkRecursiveFlagSupport()
+
+	return driver
 }
 
 func (driver *BtrFSDriver) CreateVolume(vol volume.FilesystemInitVolume) error {
 	_, _, err := driver.run(driver.btrfsBin, "subvolume", "create", vol.DataPath())
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return err
 }
 
 func (driver *BtrFSDriver) DestroyVolume(vol volume.FilesystemVolume) error {
+	// If btrfs supports -R flag, use it for recursive deletion
+	if driver.supportsRecursiveFlag {
+		driver.logger.Debug("using-recursive-flag", lager.Data{"path": vol.DataPath()})
+		_, _, err := driver.run(driver.btrfsBin, "subvolume", "delete", "-R", vol.DataPath())
+		return err
+	}
+
+	// Fall back to manual recursive deletion
+	driver.logger.Debug("using-manual-recursion", lager.Data{"path": vol.DataPath()})
+
 	volumePathsToDelete := []string{}
 
 	findSubvolumes := func(p string, f os.FileInfo, err error) error {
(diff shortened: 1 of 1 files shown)


taylorsilva:
What version of btrfs will this require users to use? Does this also effect the minimum kernel version that Concourse would support if users want to use btrfs?

analytically:
6.12 - https://github.com/kdave/btrfs-progs/commit/a4d2d1b498c121ebcaddc800e1d98ee285c1568f

btrfs-progs-6.12 (2024-11-29)[](https://btrfs.readthedocs.io/en/latest/CHANGES.html#btrfs-progs-6-12-2024-11-29)

analytically:
It's available since kernel 4.16 - https://github.com/kdave/btrfs-progs/commit/678da5a7f71ed7fbc887e40cc8ef91008c6fe843

taylorsilva (changes requested):
Thanks for doing that research! So the feature has been available in the kernel since 4.16, but the cli tool has only had it more recently, since 2024/11.

It seems possible that someone running on a kernel <6.12 would not easily have access to `btrfs` cli with the `--recursive` option then, unless they went out of their way to compile it themselves, correct?

Quickly testing this with Docker

```bash
$ docker run --entrypoint /bin/bash -it ubuntu:22.04
#running inside the container now
$ apt update
$ apt install btrfs-progs
$ btrfs version
btrfs-progs v5.16.2
$ btrfs subvolume delete --help
usage: btrfs subvolume delete [options] <subvolume> [<subvolume>...]
       btrfs subvolume delete [options] -i|--subvolid <subvolid> <path>

    Delete subvolume(s)

    Delete subvolumes from the filesystem, specified by a path or id. The
    corresponding directory is removed instantly but the data blocks are
    removed later.
    The deletion does not involve full commit by default due to
    performance reasons (as a consequence, the subvolume may appear again
    after a crash). Use one of the --commit options to wait until the
    operation is safely stored on the media.

    -c|--commit-after      wait for transaction commit at the end of the operation
    -C|--commit-each       wait for transaction commit after deleting each subvolume
    -i|--subvolid          subvolume id of the to be removed subvolume
    -v|--verbose           deprecated, alias for global -v option

    Global options:
    -v|--verbose       increase output verbosity
    -q|--quiet         print only errors
```

I think we need to do the safe thing here, which sadly means not getting rid of the old code and adding more code 😞 

I think this PR should try and detect if `btrfs subvolume delete` supports the `--recursive` flag. If it does, then use it and take the short-happy code path. If we don't detect support for `--recursive` then take the old-long-sad code path.

Maybe add a comment stating that the old-long-sad code path can be deleted once the minimum kernel version Concourse supports is >=6.12 (probably 5+ years from now 😅 )

analytically:
> It seems possible that someone running on a kernel <6.12 would not easily have access to `btrfs` cli with the `--recursive` option then, unless they went out of their way to compile it themselves, correct?

That's correct yeah. But I think manually scanning the files hides problems with large volumes. So we could potentially scan the CLI version and apply the approach needed.

analytically:
We could potentially shop a btrfs binary ourselves.

taylorsilva:
> We could potentially shop a btrfs binary ourselves.

I don't think it's wise for the project to take on that maintenance burden of building and shipping the btrfs binary at this time.

analytically:
Ok I'll change it so it detects the version of btrfs-progs available

worker/baggageclaim/volume/driver/btrfs.go
```diff
@@ -81,6 +96,80 @@ func (driver *BtrFSDriver) CreateCopyOnWriteLayer(
...
+		if err != nil {
+			// If parsing fails, treat as 0
+			num = 0
+		}
+		numbers = append(numbers, num)
+	}
+
+	return numbers
+}
+
```
taylorsilva: Instead of writing all this version checking logic that also has no tests, you can pull in one of the dependencies that we already use for semver checks:

- From fly github.com/cppforlife/go-semi-semantic
  - Used here: https://github.com/concourse/concourse/blob/master/fly/version/version.go
- Indirect dependency, but looks like most popular Go lib for semver parsing: https://github.com/Masterminds/semver

worker/baggageclaim/volume/driver/btrfs.go
```diff
@@ -81,6 +96,80 @@ func (driver *BtrFSDriver) CreateCopyOnWriteLayer(
 	return err
 }
 
+func (driver *BtrFSDriver) checkRecursiveFlagSupport() bool {
+	// Get btrfs version
+	stdout, _, err := driver.run(driver.btrfsBin, "version")
```
taylorsilva: I was thinking of running `btrfs subvolume delete --help` and checking for `--recursive`. Would avoid having to parse and check the version number at all. Not a strict suggestion, but would like your thoughts on that approach vs parsing and checking the semver number.

taylorsilva (changes requested):
Left some thoughts about the semver parsing.

For transparency, is this PR LLM-assisted? The redundant commenting is what's tipping me off:
```go
// Check if version is >= 6.12
supportsFlag := isVersionAtLeast(version, "6.12")
```

worker/baggageclaim/volume/driver/btrfs.go
analytically: Yeah I prefer this approach, implemented.

taylorsilva (approved):
LGTM - Thanks again for the PR

PR #9293, opened 2025-09-10
Prevent resources from emitting empty versions

### LINKED ISSUE #9279: Resources can emit empty versions
### Summary

From https://github.com/concourse/concourse/pull/9165 and https://github.com/orgs/concourse/discussions/9265

Concourse resource types can emit an empty json object `{}` as a version. This does render in the UI as an empty rectangle and is stored in the database as `{}`.

image

### Steps to Reproduce

I modified the mock resource type to emit an empty version `{}`: https://github.com/concourse/mock-resource/tree/empty-version

```yaml
resource_types:
  - name: mock-empty
    type: registry-image
    source:
      repository: concourseteam/mock-resource
      tag: empty-version

resources:
  - name: mock
    type: mock-empty

jobs:
  - name: job
    plan:
      - get: mock
```

Click on the resource, you'll see the empty version:
image

### Expected Results

Empty version is not stored by Concourse. Unsure if we should error when we encounter empty versions or just filter them out. I'm leaning towards filtering them out since I don't think anyone really wants them.

### Actual Results

Empty version object is accepted by Concourse and stored in the database as `{}` and rendered in the UI as an empty version.

### Web Node(s) configuration

_No response_

### Worker(s) configuration

_No response_

### Concourse Version

All versions as of 7.14

### Browser (if applicable)

_No response_

### Did this use to work?

No

IvanChalukov:

closes #9279

Resources should not be allowed to emit empty versions (`{}`). Currently, Concourse accepts them, which results in confusing behavior:

- They render in the UI as a blank/empty rectangle.
- They are stored in the database as {}, which has no semantic meaning.

This PR filters out empty versions before they are persisted and provide a clear error message in the UI, so users know their resource is misconfigured and can update its configuration.

## Notes to reviewer

To validate it:
1. Add it to a pipeline:
```yml
resource_types:
  - name: mock-empty
    type: registry-image
    source:
      repository: concourseteam/mock-resource
      tag: empty-version

resources:
  - name: mock
    type: mock-empty

jobs:
  - name: job
    plan:
      - get: mock
```
2. Output:
image

### DIFF
--- a/atc/db/resource_config_scope.go
+++ b/atc/db/resource_config_scope.go
@@ -261,7 +261,7 @@ func (r *resourceConfigScope) UpdateLastCheckEndTime(succeeded bool) (bool, erro
 
 func saveResourceVersion(tx Tx, rcsID int, version atc.Version, metadata ResourceConfigMetadataFields, spanContext SpanContext) (bool, error) {
 	if len(version) == 0 {
-		return false, errors.New("resource output version is missing")
+		return false, errors.New("resource output version is empty. Version must contain at least one key-value pair")
 	}
 
 	versionJSON, err := json.Marshal(version)
--- a/atc/db/resource_config_scope.go
+++ b/atc/db/resource_config_scope.go
@@ -3,6 +3,7 @@ package db
 import (
 	"database/sql"
 	"encoding/json"
+	"errors"
 	"time"
 
 	"code.cloudfoundry.org/lager/v3"
@@ -259,6 +260,10 @@ func (r *resourceConfigScope) UpdateLastCheckEndTime(succeeded bool) (bool, erro
 }
 
 func saveResourceVersion(tx Tx, rcsID int, version atc.Version, metadata ResourceConfigMetadataFields, spanContext SpanContext) (bool, error) {
+	if len(version) == 0 {
+		return false, errors.New("resource output version is missing")
+	}
+
 	versionJSON, err := json.Marshal(version)
 	if err != nil {
 		return false, err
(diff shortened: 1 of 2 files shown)


atc/db/resource_config_scope.go
```diff
@@ -259,6 +260,10 @@ func (r *resourceConfigScope) UpdateLastCheckEndTime(succeeded bool) (bool, erro
 }
 
 func saveResourceVersion(tx Tx, rcsID int, version atc.Version, metadata ResourceConfigMetadataFields, spanContext SpanContext) (bool, error) {
+	if len(version) == 0 {
+		return false, errors.New("resource output version is missing")
```
taylorsilva:
```suggestion
		return false, errors.New("resource output version is empty. Version must contain at least one key-value pair")
```
nit: make the error more descriptive and tell the user what's wrong.

taylorsilva (changes requested):
I am good with this being the solution to #9279. Thank you Ivan! Only one suggestion, then I'm happy to merge this in.

atc/db/resource_config_scope.go
IvanChalukov: Thanks for the suggestion - I’ve updated it. Messaging isn’t my strong suit 😅

IvanChalukov:
One more note: this is technically a breaking change for anyone using resources that emit empty versions(I hope no one is doing this). It’s probably worth calling out in the upcoming release notes.

taylorsilva:
That is a fair point. I'll make sure it's listed under breaking changes.

Kump3r (approved):
Tested with local Concourse:
Without the change (code from master):
without-pr-code
With the change (code from PR):
with-pr-code
Also I see the following info message in the logs:
```
web-1     | {"timestamp":"2025-09-12T10:55:59.811931047Z","level":"info","source":"atc","message":"atc.tracker-imb.run.errored","data":{"build":"check","error":"save versions: resource output version is empty. Version must contain at least one key-value pair","pipeline":"test-empty","pre_build_id":12,"resource":"mock","session":"23.11","team":"main"}}
```
LGTM!

taylorsilva (approved):

PR #9297, opened 2025-09-16
Add ptrace syscall to seccomp profile. Kernel requirement >= 4.8. 

### LINKED ISSUE #9174: containerd worker seccomp profile doesn't allow `ptrace`
The containerd seccomp profile doesn't include `ptrace` in its seccomp profile. This prevents `strace` and some other useful tools from working:

```
root@9f4415aa-8ad1-4490-6c49-f5417ffb2f83:/tmp/build/get# strace ls   
strace: test_ptrace_get_syscall_info: PTRACE_TRACEME: Operation not permitted
strace: ptrace(PTRACE_TRACEME, ...): Operation not permitted
strace: PTRACE_SETOPTIONS: Operation not permitted
strace: detach: waitpid(250): No child processes
strace: Process 250 detached
```

This is of course not the same thing as adding `CAP_SYS_PTRACE` - allowing ptrace in seccomp would only enable the safer uses.

moby/docker added default support for own-process `ptrace` in 2019 via https://github.com/moby/moby/commit/1124543ca8071074a537a15db251af46a5189907 - note it used some extra conditional machinery to restrict it to kernels >= 4.8, but such old kernels (ubuntu 16.x era) are extremely unlikely to still be in use in 2025 and I'd argue such a check could probably be skipped.

Users can of course always set a custom seccomp profile since #8044, but it would a lot handier for this to work out of the box and not require users to keep a custom seccomp profile in-sync with concourse's default.

I should admit that I've got extra skin-in-the-game here because we use a custom resource https://github.com/alphagov/concourse-bag-resource that internally uses the `proot` tool to allow resource types to be "stacked" (I haven't publicized this much yet and it may offend your sensibilities) - it would be neat if it worked out-of-the-box on both guardian and containerd workers.

analytically:
Resolves #9174.

### DIFF
--- a/worker/runtime/spec/seccomp.go
+++ b/worker/runtime/spec/seccomp.go
@@ -232,6 +232,7 @@ var seccomp = &specs.LinuxSeccomp{
 				"process_mrelease",
 				"pselect6",
 				"pselect6_time64",
+				"ptrace",
 				"pwrite64",
 				"pwritev",
 				"pwritev2",


taylorsilva (approved):
LGTM - Will run the PR checks and make sure they go green and then merge.

PR #9298, opened 2025-09-17
Fix time-based filtering returning incorrect builds

### LINKED ISSUE #8998: fly builds does not respect given time range
## Summary

It seems that `fly builds -t ... -p ... --since ...` does not return _"all builds (created) in the given time range"_ but instead returns _"all builds with an id equal or greater than the id of the builds with the oldest creation date in the given time range"_. If I understand correctly, this is expected as [getBuildsWithDates](https://github.com/concourse/concourse/blob/62d9c6786619ec049f94c26eb0d60b24a5685eec/atc/db/build_factory.go#L277) selects the minimums and maximum build **by time**, but then calls [getBuildsWithPagination](https://github.com/concourse/concourse/blob/62d9c6786619ec049f94c26eb0d60b24a5685eec/atc/db/build_factory.go#L362) to return all inbetween builds **by id**. This can return a lot of builds that do not fall within the specified time-range.

I am experiencing this with a build that has been paused for a long time (a month or so) but somehow was triggered (manually a month a go) and got assigned a (now month-old) build id. After unpausing this build, it got a recent start and finish time. Now, when a time range of `fly builds ... --since ...` of, say, the last to hours includes this build, then all builds (a whole lot of them) of the past month are returned even though these builds do not fall into the specified time range.

## Steps to reproduce

I think that you can reproduce this by

1. pause a job
2. trigger that job
3. wait some time, trigger jobs inbetween
4. wait some more time
5. unpause the job and let it finish
6. request logs with `fly builds ... --since ...` just including _the long-paused-recently-finished job_

## Expected results

just the builds after the time specified with `--since` should be returned

## Actual results

all builds that were created after _the long-paused-recently-finished job_ are returned

## Triaging info

* Concourse version: 7.11.2
* Did this used to work? I guess not

analytically:
Fly's `--since` flag was returning all builds created after a given time rather than builds that actually started after that time. This occurred because getBuildsWithDates converted time boundaries to build IDs, incorrectly assuming IDs correlate with start times.

When a paused build was triggered long ago but started recently, it would have an old ID but recent start_time, causing `--since` to return months of unrelated builds.

This fix removes the ID conversion and filters directly on start_time, ensuring only builds that actually started within the time range are returned.

Fixes #8998

NOTE: this does cause a table scan when there is no index on start_time. This might be desirable, as the date option isn't used often and adding an index might be problematic. If not we should add a migration to run `CREATE INDEX CONCURRENTLY idx_builds_start_time ON builds(start_time);`

### DIFF
--- a/atc/db/build_factory.go
+++ b/atc/db/build_factory.go
@@ -4,7 +4,6 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
-	"strconv"
 	"time"
 
 	"code.cloudfoundry.org/lager/v3"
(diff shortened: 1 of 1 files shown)


atc/db/build_factory.go
```diff
@@ -275,88 +274,34 @@ func getBuilds(buildsQuery sq.SelectBuilder, conn DbConn, lockFactory lock.LockF
 }
 
 func getBuildsWithDates(buildsQuery, minMaxIdQuery sq.SelectBuilder, page Page, conn DbConn, lockFactory lock.LockFactory) ([]BuildForAPI, Pagination, error) {
```
taylorsilva, line 276: `minMaxIdQuery` is no longer used and can be removed.

taylorsilva (changes requested):
LGTM - Just one small nit and then it's good to merge.

taylorsilva (approved):
LGTM - Will remove the unused arg later

PR #9300, opened 2025-09-18
skip container limits test when using guardian runtime

taylorsilva:
on systems with only cgroup v2 enabled.

See upstream issue https://github.com/cloudfoundry/garden-runc-release/issues/384

### DIFF
--- a/testflight/container_limits_test.go
+++ b/testflight/container_limits_test.go
@@ -29,6 +29,10 @@ var _ = Describe("A job with a task that has container limits", func() {
 	})
 
 	It("sets the correct CPU and memory limits on the container", func() {
+		if config.Runtime == "guardian" && cgroupsV2Only() {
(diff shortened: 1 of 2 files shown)


taylorsilva:
containerd-integration has one failing test. The test suite is passing in the main pipeline, but failing for all PR's:

```
=== NAME  TestSuite

    integration_test.go:492: 

        	Error Trace:	github.com/concourse/concourse/worker/runtime/integration/integration_test.go:492

        	Error:      	Not equal: 

        	            	expected: 0

        	            	actual  : 1

        	Test:       	TestSuite

        	Messages:   	Process in container should be able to reach the host network

=== NAME  TestSuite/TestContainerAllowsHostAccess

    testing.go:1811: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test

--- FAIL: TestSuite (33.05s)
```

My gut is telling me this is due to a difference between the PR worker and the normal workers. Digging through the infra code to try and find the difference.

taylorsilva:
Found the diff, was networking stuff!

PR #9301, opened 2025-09-20
Fix overlay driver metacopy detection for kernels without support

analytically:
Previously checked only for file existence, not actual enablement. Now reads /sys/module/overlay/parameters/metacopy and verifies it contains 'Y' before using metacopy=on mount option.

Fixes mount failures on systems with CONFIG_OVERLAY_FS_METACOPY unset.

### DIFF
--- a/worker/baggageclaim/volume/driver/overlay_linux.go
+++ b/worker/baggageclaim/volume/driver/overlay_linux.go
@@ -25,11 +25,8 @@ func init() {
 // file then it will not create a copy of the file. Files will only be copied
 // when they are written to.
 func metacopySupported() bool {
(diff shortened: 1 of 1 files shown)


analytically:
More details on metacopy and K8s here: https://kinvolk.io/blog/2023/11/tips-and-tricks-for-user-namespaces-with-kubernetes-and-containerd

analytically:
Actually this is wrong. CONFIG_OVERLAY_FS_METACOPY just sets the default. `/sys/module/overlay/parameters/metacopy` exists when the overlay module is loaded. So Concourse will ALWAYS set metacopy=on. So we shouldn't test metacopySupported unless we're supporting a kernel that didn't have that feature?

PR #9303, opened 2025-09-30
fix: use tolerance for scroll bottom detection to prevent auto-scrolling breaking

analytically:
The exact equality check for detecting scroll position at bottom was failing due to sub-pixel rounding differences between scrollHeight, scrollTop, and clientHeight. This caused autoScroll to incorrectly disable during rapid content updates.

Added 1px tolerance to account for browser rounding behavior.

### DIFF
--- a/web/elm/src/Build/Build.elm
+++ b/web/elm/src/Build/Build.elm
@@ -478,7 +478,7 @@ update msg ( model, effects ) =
         Scrolled { scrollHeight, scrollTop, clientHeight } ->
             ( { model
                 | autoScroll =
-                    (scrollHeight == scrollTop + clientHeight)
+                    (scrollHeight - (scrollTop + clientHeight) <= 1)
                         && not model.isScrollToIdInProgress
               }
             , effects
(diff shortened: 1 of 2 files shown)


taylorsilva (changes requested):
LGTM - I made this test pipeline to validate the fix:

```yaml
jobs:
  - name: job
    plan:
      - task: print
        config:
          platform: linux
          image_resource:
            type: mock
            source: {mirror_self: true}
          run:
            path: bash
            args:
              - -c
              - |
                sleep 3s
                for i in {1..10000}; do
                  sleep 0.5s
                  echo "Line $i"
                done
```

Once the text hit the bottom of the screen, shortly after the auto-scroll would stop. Re-building the frontend with this PR fixed the issue, auto-scrolling kept working.

Funnily, the related test for this PR is failing now:
```
$ yarn test
Compiling > Starting tests

elm-test 0.19.1-revision16
--------------------------

Running 2953 tests. To reproduce these results, run: elm-test --fuzz 100 --seed 159141073039854

↓ BuildTests
↓ build page
✗ when build is running but the user is not scrolled to the bottom it does not scroll on build event

    [Scroll ToBottom "build-body"]
    ╷
    │ Expect.equal
    ╵
    []



TEST RUN FAILED

Duration: 637 ms
Passed:   2952
Failed:   1
```

Please update the test so it passes. Then this is good to merge. Thank you!

taylorsilva (approved):

PR #9309, opened 2025-10-02
fix: close SQL result sets immediately in JobsToSchedule loop

analytically:
The JobsToSchedule method was using defer Close() inside the job iteration loop, causing all result sets to remain open until function completion. With many jobs to schedule, this exhausts database connections and memory.

Changed to close resourceRows, resourceTypeRows, and prototypeRows immediately after processing each job instead of deferring closure until function return. Added rows.Err() checks after each iteration loop for defensive error handling.

### DIFF
--- a/atc/db/job_factory.go
+++ b/atc/db/job_factory.go
@@ -102,6 +102,8 @@ func (j *jobFactory) JobsToSchedule() (SchedulerJobs, error) {
 		return nil, err
 	}
 
+	es := j.conn.EncryptionStrategy()
+
 	var schedulerJobs SchedulerJobs
 	pipelineResourceTypes := make(map[int]ResourceTypes)
 	pipelinePrototypes := make(map[int]Prototypes)
@@ -130,8 +132,6 @@ func (j *jobFactory) JobsToSchedule() (SchedulerJobs, error) {
 				return nil, err
 			}
 
-			es := j.conn.EncryptionStrategy()
-
 			var noncense *string
 			if nonce.Valid {
 				noncense = &nonce.String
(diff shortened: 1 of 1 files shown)


atc/db/job_factory.go
```diff
@@ -155,6 +154,11 @@ func (j *jobFactory) JobsToSchedule() (SchedulerJobs, error) {
 				ExposeBuildCreatedBy: config.ExposeBuildCreatedBy,
 			})
 		}
+		if err = resourceRows.Err(); err != nil {
+			resourceRows.Close()
+			return nil, err
+		}
+		resourceRows.Close()
```
taylorsilva, line 162:
```suggestion
		Close(resourceRows)
```

atc/db/job_factory.go
```diff
@@ -177,6 +180,11 @@ func (j *jobFactory) JobsToSchedule() (SchedulerJobs, error) {
 
 				resourceTypes = append(resourceTypes, resourceType)
 			}
+			if err = resourceTypeRows.Err(); err != nil {
+				resourceTypeRows.Close()
+				return nil, err
+			}
+			resourceTypeRows.Close()
```
taylorsilva, line 189:
```suggestion
			Close(resourceTypeRows)
```

atc/db/job_factory.go
```diff
@@ -202,6 +209,11 @@ func (j *jobFactory) JobsToSchedule() (SchedulerJobs, error) {
 
 				prototypes = append(prototypes, prototype)
 			}
+			if err = prototypeRows.Err(); err != nil {
+				prototypeRows.Close()
+				return nil, err
+			}
+			prototypeRows.Close()
```
taylorsilva, line 219:
```suggestion
			Close(prototypeRows)
```

taylorsilva (changes requested):
By not calling `Close()` with `defer` this presents a new risk: resource leaks if the function returns before we can call `.Close()`.

See https://github.com/golang/go/issues/46863

Good news is, we can call `.Close()` multiple times without any side-effects! Please add back the `defer Close()` lines, then this will be good to merge.

taylorsilva (approved):
LGTM

Nit: the `.Close()` inside the `rows.Err()` if statements is redundant because the deferred `Close()` will be called when we hit the `return nil, err`.

PR #9311, opened 2025-10-02
fix: prevent panic in teams cache when notification listener fails

analytically:
Return early from waitForNotifications goroutine if Listen() returns an error, avoiding nil channel dereference panic in the receive loop.


taylorsilva (approved):
LGTM - Triggering checks again since they likely failed for unrelated reasons.

PR #9312, opened 2025-10-02
fix: use case-insensitive comparison for cloudfoundry connector normalization

analytically:
Use strings.EqualFold instead of == when checking if connector is "cloudfoundry" to ensure consistent case-insensitive behavior with the rest of the auth matching logic in roleOnTeam.

### DIFF
--- a/atc/api/accessor/accessor.go
+++ b/atc/api/accessor/accessor.go
@@ -95,7 +95,7 @@ func contains(arr []string, val string) bool {
 func (a *access) rolesForTeam(auth atc.TeamAuth) []string {
 	connectorID := a.connectorID()
 
-	if connectorID == "cloudfoundry" {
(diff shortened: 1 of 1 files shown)


IvanChalukov:
I believe this change is safe to merge, although it is unlikely to have a practical impact. Since the value is generated by the Dex server rather than user input, casing inconsistencies should not occur under normal circumstances. I was not able to identify any scenario in which the value would differ from `cloudfoundry`.

taylorsilva:
Thanks for digging into that Ivan. Sounds like this is overkill then. It would result in a performance hit as well compared to the `==` operator.

analytically:
A performance hit? You can't be serious. The EqualFold is used everywhere else in the file for comparison of external data.

taylorsilva:
eh you're right, it would be very benign

PR #9318, opened 2025-10-06
feat: display team name in set_pipeline step header

### LINKED ISSUE #9255: Display team for `set_pipeline` step
## Summary

When using the `team` field on `set_pipeline` step, the team should be displayed in the collapsible black bar after the pipeline name. If no team name is specified, only the pipeline name is displayed, which is the current state.  

## Context

* Prior discussion: https://github.com/orgs/concourse/discussions/5731#discussioncomment-50736

analytically:
Add team field to BuildStepSetPipeline type and display it in the step header after the pipeline name when specified. Backend already sends the team field; frontend was ignoring it.

Fixes #9255

### DIFF
--- a/web/elm/src/Build/StepTree/StepTree.elm
+++ b/web/elm/src/Build/StepTree/StepTree.elm
@@ -102,7 +102,7 @@ init buildId hl resources plan =
         Concourse.BuildStepArtifactOutput _ ->
             step |> initBottom buildId hl resources plan ArtifactOutput
 
-        Concourse.BuildStepSetPipeline _ _ ->
+        Concourse.BuildStepSetPipeline _ _ _ ->
             step |> initBottom buildId hl resources plan SetPipeline
 
         Concourse.BuildStepLoadVar _ ->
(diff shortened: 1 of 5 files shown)


taylorsilva (changes requested):
When you're making UI changes, can you please include before/after images in your PR description. It helps speed up reviewing the PR.

Before
image

After
image

---

So after the pipeline name you've appended `[TEAM]`. It doesn't feel like this meshes with the style we usually see. For example, with the across step:

image

Which visually feels easier to parse. Initially I couldn't even tell where the team names were rendering.

I'm thinking this:
image

Instead of what you currently have here:
image

I'm open to other visual styles if you've got any other ideas. WDYT?

analytically:
Okay I'll change the style to be more in line with other places. I'll post screenshots.

analytically:
Looks like this now: 

screentshot-2025-11-15 at 17 25 14@2x

analytically:
screentshot-2025-11-15 at 17 28 20@2x

web/elm/src/Build/StepTree/StepTree.elm
```diff
@@ -1247,17 +1247,25 @@ viewStepHeader step =
...
-                    :: (if Dict.isEmpty instanceVars then
-                            []
-
-                        else
-                            [ Html.span [ style "margin-left" "10px", style "margin-right" "4px" ] [ Html.text "/" ]
-                            , viewKeyValuePairHeaderLabels (Dict.toList instanceVars)
+        Concourse.BuildStepSetPipeline name team instanceVars ->
+            headerWithContent "set_pipeline:" (Just "pipeline config changed")
+                ([Html.span [] [ Html.text name ]]
+                 ++ (case team of
```
taylorsilva: `yarn analyse` failing with:

```
Messages:

- src/Build/StepTree/StepTree.elm

  > Use `::` instead of `++` at ((1252,18),(1268,18))
```

```suggestion
                (Html.span [] [ Html.text name ]
                 :: (case team of
```

taylorsilva (changes requested):
Yes, looks good now! I like that it only shows the `team` field if it's different from the current team.

image

Just one small thing to fix from `yarn analyse` and then this is good to merge.

taylorsilva (approved):

PR #9320, opened 2025-10-07
Fix active_tasks going negative when MaxTasks == 0

analytically:
Release() decremented without matching increment when MaxTasks == 0. Add same guard to Release() that exists in Approve().

Fixes logs like:
ERROR: new row for relation \"workers\" violates check constraint \"workers_active_tasks_check\"


taylorsilva (approved):
LGTM

PR #9323, opened 2025-10-09
fix: pass correct target id when recursing into retry steps

analytically:


taylorsilva (changes requested):
_sigh_ - what does this fix exactly? Looking at the code immediately around the diff, it isn't obvious what this is fixing. I'm guessing you did this because something was visually off in the web ui? Can you show a before/after to showcase what this changes, or include steps to reproduce the bug please?

I gave this a shot trying to reproduce with this pipeline:

```yaml
jobs:
  - name: attempts
    plan:
      - task: randomly-fails
        attempts: 4
        config: &task
          platform: linux
          image_resource: &image
            type: mock
            source: {mirror_self: true}
          run:
            path: bash
            args:
              - -c
              - |
                echo "will randomly fail or succeed"
                sleep 1s
                exit $((RANDOM % 2))

  - name: attempt-multi-steps
    plan:
      - attempts: 4
        do:
          - task: try-1
            config: *task
          - task: try-2
            config: *task
```

But my before and after screenshots look the same. Only sharing screenshots of the `attempt-multi-steps` job. The other job was also visually the same.

### Before
image

### After
image

Having detailed PR messages helps speed up reviews. Seeing an empty PR message is (often) quite annoying. You assume I can glean all the context from the title of the PR and this isn't always the case. While trying to understand what this change fixes I dug into the commit history of the file and found this PR which I think is a good example of a detailed PR message, and is very similar to your PR: https://github.com/concourse/concourse/pull/8069

Detailed explanations of the bug also help those on older versions find out if newer versions fix issues they're having.

Please update the PR message explaining/showing what this fixes. Screenshots, reproducibles, _something_ 🙏

analytically:
It's nested Retries within Retry nodes. I'll improve the PR, sorry, we had this on a branch for ages and I didn't bring up the web diff again. Thought it would be obvious enough from the code.

taylorsilva:
Giving it another go, changed my second job so the top `do` step has `attempts` and the nested `task` steps have attempts too. Nested retries feels odd:

```yaml
jobs:
  - name: attempts
    plan:
      - task: randomly-fails
        attempts: 2
        config: &task
          platform: linux
          image_resource: &image
            type: mock
            source: { mirror_self: true }
          run:
            path: bash
            args:
              - -c
              - |
                echo "will randomly fail or succeed"
                sleep 1s
                exit $((RANDOM % 2))

  - name: attempt-multi-steps
    plan:
      - attempts: 4
        do:
          - task: try-1
            attempts: 2
            config: *task
          - task: try-2
            attempts: 2
            config: *task
```

Looking at `attempt-multi-steps` on `master`:
image

Using this PR:
image

No diff 😕 
These definitely don't look great as-is, but ignoring that for now 😅 

Added a third job:

```yaml
  - name: attempt-across
    plan:
      - attempts: 4
        across:
          - var: somevar
            values: [1, 2, 3, 4]
        do:
          - task: try-1
            attempts: 2
            config: *task
```

On `master`:
image

With this PR:
image

Still no diff. I'll wait for you to provide an example pipeline showcasing the issue.

PR #9324, opened 2025-10-09
Fix HTML injection vulnerability in build names (causality graph)

analytically:
Escape build names in causality graph job labels to prevent HTML/XSS injection. Build names containing <, >, &, ", or ' characters were being rendered unescaped, breaking the GraphViz HTML table structure.


IvanChalukov:
Is it possible to build name to containes any of `<, >, &, ", or '`? Do you have any samples how that could be achieved?

taylorsilva (commented):
Great question by Ivan, please answer it. Again, for UI changes, can you show before/after images and include reproducible steps. TY!

analytically:
Yeah this is a non-issue after deeper inspection. Other names in the same file were escaped so I thought it was an oversight.

PR #9325, opened 2025-10-09
replace left/right quotes in no builds message with straight quotes

analytically:


taylorsilva (approved):
Screenshots 🙏 

Before
image

After
image

Previous is not a "corrupt" quotation mark, it's just not the standard one anyone can easily type on their keyboard.

I do prefer the way it looks with the straight quotation mark 👍

PR #9326, opened 2025-10-09
allow CDN caching for hashed static assets

analytically:
Replace "private" with "public, immutable" in Cache-Control header. Assets use content hashing, safe for aggressive caching.


analytically:
The `immutable` directive tells browsers to never revalidate a cached resource, even on hard refresh, because the content at that URL will never change. This is safe for Concourse because assets use content-hash URLs like `/public/elm.min.js?id=a1b2c3d4`, so when the file changes the hash changes and the browser sees it as a completely new URL. Without immutable, browsers waste time sending "has this changed?" requests to the server even though the answer is always "no" for a given hash, whereas with it they skip those requests entirely and load instantly from cache.

taylorsilva (approved):

PR #9327, opened 2025-10-09
fix: buffer template output before response

analytically:
Use pooled buffers to render templates before writing HTTP response. Fixes incorrect "200 OK" status on template errors. Eliminates buffer allocations via sync.Pool.

### DIFF
--- a/web/index_handler.go
+++ b/web/index_handler.go
@@ -1,10 +1,12 @@
 package web
 
 import (
+	"bytes"
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
 	"html/template"
+	"io"
 	"io/fs"
 	"net/http"
 	"sync"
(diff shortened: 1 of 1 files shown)


web/index_handler.go
```diff
@@ -61,28 +63,55 @@ func (ac *assetCache) asset(path string) (string, error) {
...
-			},
-		}
+		var t *template.Template
+		var err error
 
-		t, err := template.New("web").Funcs(funcs).ParseFS(publicFS, "index.html")
-		if err != nil {
-			log.Error("failed-to-parse-templates", err)
-			w.WriteHeader(http.StatusInternalServerError)
-			return
```
taylorsilva, line 82: I'm trying to understand what was wrong with the previous code and how you would get 200 OK on template errors. From what I can see here it appears we return HTTP 500 if the template errors.

Do you have a reproducible or any logs from this log line `failed-to-parse-templates`?

taylorsilva:
Closing because no follow-up from contributor.

PR #9330, opened 2025-10-11
Nit: Remove sha1 usage and make task pwd less predictable

taylorsilva:

users could determine the working dir of the task because it was based
on the sha1 of the task name. The intention has always been to encourage
users to not rely knowing the working dir of their tasks and write portable scripts.

This also removes usage of sha1 within the Go codebase of Concourse, though its usage here was fine.

### DIFF
--- a/atc/engine/step_factory.go
+++ b/atc/engine/step_factory.go
@@ -1,7 +1,7 @@
 package engine
 
 import (
-	"crypto/sha1"
+	"crypto/sha256"
 	"fmt"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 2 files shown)


atc/engine/step_factory.go
```diff
@@ -177,7 +177,7 @@ func (factory *coreStepFactory) TaskStep(
 	containerMetadata db.ContainerMetadata,
 	delegateFactory DelegateFactory,
 ) exec.Step {
-	sum := sha1.Sum([]byte(plan.Task.Name))
+	sum := sha256.Sum256(fmt.Appendf([]byte{}, "%d", time.Now().Unix()))
```
analytically, line 180: This is second granularity, and will cause same working directory when tasks are started in same second! I think you meant to use UnixNano or UnixMilli?

taylorsilva: I know, but it doesn't matter because it would be in two separate containers. The main point is that the pwd is not as predictable now since there's no way to know the exact time that this value is calculated at and therefore figure out the sha.

PR #9335, opened 2025-10-16
Add `--non-interactive` flag to `clear-versions` command

mstandley-tempus:
Allow users to clear versions for a pipeline resource or resource type without being prompted for confirmation

* [x]  add non-interactive flag to `clear-versions`
* [x]  integration tests

## Release Note
* Added non-interactive flag to fly command clear-versions. Use:
  `fly -t dev clear-versions --resource=pipeline/resource --non-interactive`


mstandley-tempus:
I built this locally and tested against a concourse instance.  The --non-interactive flag works as expected, and it still prompts for confirmation and works as expected when that flag is omitted.

Kump3r (approved):
Again tested by building fly and targeting an internal instance. Default behaviour asks for confirmation new `-n/--non-interactive` works as on the other commands.

PR #9337, opened 2025-10-17
Retry LoadProcess with fixed sleep for exec registration races

analytically:
Under load, exec processes may not be registered immediately in containerd. Add retry loop (max 5 attempts, 10ms fixed sleep) to handle transient "no running process found" errors during Attach.

Fixes timeouts when spawning concurrent tasks on loaded workers.

### DIFF
--- a/worker/runtime/container.go
+++ b/worker/runtime/container.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"regexp"
 	"slices"
+	"strings"
 	"time"
 
 	"code.cloudfoundry.org/garden"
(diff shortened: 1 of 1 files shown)


worker/runtime/container.go
```diff
@@ -191,11 +192,30 @@ func (c *Container) Attach(pid string, processIO garden.ProcessIO) (process gard
...
 
-	proc, err := task.LoadProcess(ctx, pid, ioAttach)
-	if err != nil {
-		return nil, fmt.Errorf("load proc: %w", err)
+	// Retry LoadProcess - exec may not be initialized yet under load.
+	const maxRetries = 5
+	var proc containerd.Process
+	var lastErr error
+
+	for attempt := 0; attempt < maxRetries; attempt++ {
```
taylorsilva, line 201: nit: can use `range` instead of c-style for loop

```suggestion
	for attempt := range maxRetries {
```

taylorsilva (approved):
LGTM - I'll fix the nit after merging

PR #9342, opened 2025-10-28
Bump elm-ansi to 11.0.1 that fixes CursorColumn (CSI G) to use 1-based indexing as per ANSI spec.

analytically:
Other fixes:
- Fix hyperlink URL parsing to preserve semicolons in URLs by only splitting on first two semicolons
- Fix non-ASCII character encoding in URLs to encode individual characters rather than entire string

### DIFF
--- a/web/elm/elm.json
+++ b/web/elm/elm.json
@@ -26,7 +26,7 @@
             "matthewsj/elm-ordering": "2.0.0",
             "ryan-haskell/date-format": "1.0.0",
             "truqu/elm-base64": "2.0.4",
-            "vito/elm-ansi": "11.0.0"
+            "vito/elm-ansi": "11.0.1"
         },
         "indirect": {
             "BrianHicks/elm-trend": "2.1.3",


analytically:
Pipeline to test: 

```
---
jobs:
  - name: csi-g-test
    plan:
      - task: test-csi-g
        config:
          platform: linux
          image_resource:
            type: registry-image
            source:
              repository: python
              tag: "3.11"
          run:
            path: sh
            args:
              - -exc
              - |
                pip install rich
                cat > test.py << 'EOF'
                from rich.console import Console
                console = Console(width=180, force_terminal=True)
                
                # Test CSI G positioning
                for col in [1, 30, 60, 90, 120, 150, 180]:
                    console.file.write(f"\033[{col}G")
                    console.file.flush()
                    console.print(f"|{col}", end="")
                console.print()
                EOF
                
                python3 test.py
```

Before:
screentshot-2025-10-28 at 18 08 44@2x

After:
screentshot-2025-10-28 at 18 10 09@2x

analytically:
DO NOT MERGE. Newer version of elm-ansi is coming with a lot more improvements.

taylorsilva:
I see v12 of elm-ansi is out: https://github.com/vito/elm-ansi/commit/264ff3ce3462a8825972009dcab4f4e863082e13

Is it safe to bump to the new major?

analytically:
It should work yeah. I still have some issue with width calculation.

PR #9343, opened 2025-10-29
web: behaviour: use async XHR for fly login to fix browser shield blocking

ramonskie:
Synchronous XMLHttpRequest is deprecated and blocked by privacy-focused
browsers like Brave when shields are enabled. This prevented fly login
from retrieving the authentication token via the localhost callback.

Switch to asynchronous XHR which is compatible with modern browser
security features and still properly handled by existing event listeners.

Signed-off-by: ramonskie <ramonmakkelie@gmail.com>

### DIFF
--- a/web/public/elm-setup.js
+++ b/web/public/elm-setup.js
@@ -341,13 +341,10 @@ app.ports.rawHttpRequest.subscribe(function(url) {
     app.ports.rawHttpResponse.send('success');
   });
 
-  xhr.open('GET', url, false);
+  xhr.open('GET', url, true);
 
   try {
     xhr.send();
-    if (xhr.readyState === 1) {
-      app.ports.rawHttpResponse.send('browserError');
-    }
   } catch (error) {
     app.ports.rawHttpResponse.send('networkError');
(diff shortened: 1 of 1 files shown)


taylorsilva:
Thanks for the PR! Will review next week most likely.

analytically (approved):
LGTM

web/public/elm-setup.js
```diff
@@ -341,13 +341,10 @@ app.ports.rawHttpRequest.subscribe(function(url) {
     app.ports.rawHttpResponse.send('success');
   });
 
-  xhr.open('GET', url, false);
+  xhr.open('GET', url, true);
```
analytically, line 344: this defaults to true, so we could just remove the third param?

ramonskie: hmm i think you are right. although it is now pretty explicit.
its up to you. i don't really have a opinion on it now

taylorsilva (approved):
LGTM - Tested on Chrome, Safari, and Firefox on macOS.

PR #9345, opened 2025-10-31
Fix [containerd]: make container attach idempotent and handle exited processes gracefully

### LINKED ISSUE #9344: Entire task restarts if web nodes restart when task is nearly done running
_Originally posted by @Kump3r in https://github.com/concourse/concourse/issues/9234#issuecomment-3468149678_

We are observing a really unfortunate side effect of this. More or less what the problem is and how you can reproduce:

## Problem:
When you are running a task and you are towards the end of the execution, if the webs restart at that specific time, the process on the worker has already completed and the variant of the task we are trying to attach to fails with (`"worker.garden.garden-server.attach.failed","data":{"error":"load proc: no running process found: process does not exist task: not found"`) and the backend manager, spawns a new task re executing the deploy/script once again. It is a really rare corner case and not sure if you planned it this way (works as designed), however it leads to executing the same script twice in a single build task.

## Build log outputs:
```
selected worker: 68451fde511c
+ echo 'pid: 7'
pid: 7
+ for i in {1..30}
+ echo 'Welcome 1 times'
Welcome 1 times
+ sleep 1
+ for i in {1..30}
+ echo 'Welcome 2 times'
Welcome 2 times
+ sleep 1
+ for i in {1..30}
+ echo 'Welcome 3 times'
Welcome 3 times
....
+ sleep 1
+ for i in {1..30}
+ echo 'Welcome 26 times'
Welcome 26 times
+ sleep 1
+ for i in {1..30}
+ echo 'Welcome 27 times'
Welcome 27 times
+ sleep 1
selected worker: 68451fde511c
+ echo 'pid: 43'
pid: 43
+ for i in {1..30}
+ echo 'Welcome 1 times'
Welcome 1 times
+ sleep 1
+ for i in {1..30}
+ echo 'Welcome 2 times'
Welcome 2 times
+ sleep 1
+ for i in {1..30}
+ echo 'Welcome 3 times'
Welcome 3 times
+ sleep 1
+ for i in {1..30}
+ echo 'Welcome 4 times'
Welcome 4 times
+ sleep 1
+ for i in {1..30}
+ echo 'Welcome 5 times'
Welcome 5 times
...
+ sleep 1
+ for i in {1..30}
+ echo 'Welcome 27 times'
Welcome 27 times
+ sleep 1
+ for i in {1..30}
+ echo 'Welcome 28 times'
Welcome 28 times
+ sleep 1
+ for i in {1..30}
+ echo 'Welcome 29 times'
Welcome 29 times
+ sleep 1
+ for i in {1..30}
+ echo 'Welcome 30 times'
Welcome 30 times
+ sleep 1
```

## Steps to reproduce:
1. Pipeline: 
```yaml
jobs:
- name: hello-world
  plan:
  - task: say-hello
    config:
      platform: linux
      image_resource:
        type: mock
        source: {mirror_self: true}
      run:
        path: bash
        args:
        - -exc
        - |
          echo "pid: $$"
          for i in {1..30}; do
            echo "Welcome $i times" && sleep 1
          done
```
2. When you reach output "Welcome 25/26 times" run `docker restart <web-container-id>`

## Context

Concourse version: Tested on 7.13.2, likely present in all versions of Concourse
Container Runtimes: containerd tested and confirmed. Need to test and verify guardian and houdini runtimes.

IvanChalukov:

closes #9344

This change refactors process attach behavior to handle exited container processes gracefully.
Instead re-running task after web restart, the runtime now reuses the recorded exit status when a process no longer exists in containerd.

## Notes to reviewer
1. Create pipeline:
```yml
jobs:
- name: hello-world
  plan:
  - task: say-hello
    config:
      platform: linux
      image_resource:
        type: mock
        source: {mirror_self: true}
      run:
        path: bash
        args: 
        - -exc
        - |
          echo "pid: $$"
          for i in {1..15}; do
            echo "Welcome $i times" && sleep 1
          done
          exit 1
```
2. Trigger pipeline
3. Restart the Concourse web process a few seconds before the loop completes.
4. Expected result
    - The worker should not re-run the task after the web restart.
    - The task should complete normally and report the correct exit status (1).
    - The pipeline UI should show the task as finished.

### DIFF
--- a/worker/runtime/container.go
+++ b/worker/runtime/container.go
@@ -24,8 +24,8 @@ const (
 	SuperuserPath = "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 	Path          = "PATH=/usr/local/bin:/usr/bin:/bin"
 
-	GraceTimeKey            = "garden.grace-time"
-	ContainerProcessExitKey = "garden.process-exit-code"
+	GraceTimeKey         = "garden.grace-time"
+	ProcessExitStatusKey = "garden.process-exit-status"
 )
 
 var (
@@ -492,7 +492,7 @@ func isNoSuchExecutable(err error) bool {
 }
 
 func (c *Container) lookupStoredExit() (int, bool) {
-	val, err := c.Property(ContainerProcessExitKey)
+	val, err := c.Property(ProcessExitStatusKey)
 	if err != nil || val == "" {
 		return 0, false
 	}
--- a/worker/runtime/container.go
+++ b/worker/runtime/container.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"regexp"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
@@ -23,7 +24,8 @@ const (
 	SuperuserPath = "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 	Path          = "PATH=/usr/local/bin:/usr/bin:/bin"
 
-	GraceTimeKey = "garden.grace-time"
+	GraceTimeKey            = "garden.grace-time"
+	ContainerProcessExitKey = "garden.process-exit-code"
 )
 
 var (
@@ -175,7 +177,7 @@ func (c *Container) Run(
 		}
 	}
 
-	return NewProcess(proc, exitStatusC), nil
+	return NewProcess(proc, exitStatusC, *c), nil
 }
 
 // Attach starts streaming the output back to the client from a specified process.
@@ -215,6 +217,11 @@ func (c *Container) Attach(pid string, processIO garden.ProcessIO) (process gard
 	}
 
 	if lastErr != nil {
+		if errdefs.IsNotFound(lastErr) {
+			if code, ok := c.lookupStoredExit(); ok {
+				return NewFinishedProcess(pid, code), nil
+			}
+		}
 		return nil, fmt.Errorf("load proc: %w", lastErr)
 	}
 
@@ -232,7 +239,7 @@ func (c *Container) Attach(pid string, processIO garden.ProcessIO) (process gard
 		return nil, fmt.Errorf("proc wait: %w", err)
 	}
 
-	return NewProcess(proc, exitStatusC), nil
+	return NewProcess(proc, exitStatusC, *c), nil
 }
 
 // Properties returns the current set of properties
@@ -482,4 +489,16 @@ func containerdCIO(gdnProcIO garden.ProcessIO, tty bool) []cio.Opt {
 
 func isNoSuchExecutable(err error) bool {
 	return noSuchFile.MatchString(err.Error()) || executableNotFound.MatchString(err.Error())
+}
+
+func (c *Container) lookupStoredExit() (int, bool) {
+	val, err := c.Property(ContainerProcessExitKey)
+	if err != nil || val == "" {
+		return 0, false
+	}
+	code, convErr := strconv.Atoi(val)
+	if convErr != nil {
+		return 0, false
+	}
+	return code, true
 }
--- a/worker/runtime/finished_process.go
+++ b/worker/runtime/finished_process.go
@@ -6,16 +6,16 @@ import (
 	"code.cloudfoundry.org/garden"
 )
 
-type finishedProcess struct {
+type FinishedProcess struct {
 	id         string
 	exitStatus int
 }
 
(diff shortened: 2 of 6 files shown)


Kump3r:
## Testing
### Parameters:
* 1 worker
* 1 web
* Process structure - idea is to have nested processes to ensure we propagate status code:
```
containerd-shim(303427)-+-bash(303551)---some-wrapper(303682)-+-bash(308832)---sleep(308911)
```

Tested this with the following scenarios:
- Usual scenario without web restart - ✅ - Works as before, no regressions
- Web restart in the beginning/middle of task - ✅ - Works as before (joins ongoing task, logs visible after attach), no regressions
- Web restart at the end:
  - Where the sleep succeeds - ✅ - Green job and no duplication of execution, after `selected worker` the build is only marked
  - Where the sleep fails - ✅- Red job and no duplicated execution, after `selected worker` the build is marked if retries present we continue with the next attempt

## Code review
Keeping in mind I am not that of an expert at the codebase, looking at the code I do not see anything that I would block, your suggestion is a lot simpler and native to this codebase, then what I was going for/looking at. Maybe it would be cool if @taylorsilva can check and give you some nit suggestions, or ideas for improvement. One thing we discussed is the introduction of the `finishedProcess` interface and that maybe there is a cleaner way to achieve what you were going for, but none that I know of.

## Additional details
Thanks for the work on this @IvanChalukov! And if you guys have anything else you would like validated I would be more than happy to do it on my end, as I have the changes locally and can do it rather easily.

worker/runtime/container.go
```diff
@@ -23,7 +24,8 @@ const (
 	SuperuserPath = "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 	Path          = "PATH=/usr/local/bin:/usr/bin:/bin"
 
-	GraceTimeKey = "garden.grace-time"
+	GraceTimeKey            = "garden.grace-time"
+	ContainerProcessExitKey = "garden.process-exit-code"
```
analytically: The proper technical name according to POSIX is 'exit status'

worker/runtime/finished_process.go
```diff
@@ -0,0 +1,21 @@
...
+
+package runtime
+
+import (
+	"code.cloudfoundry.org/garden"
+)
+
+type finishedProcess struct {
+	id       string
+	exitCode int
```
analytically: We use exitStatusC so let's keep the naming consistent?

worker/runtime/container.go
IvanChalukov: I will change it to "garden.process-exit-status".

worker/runtime/finished_process.go
IvanChalukov: I’ll rename it to `exitStatus` for clarity instead of `exitStatusC`, since the `C` makes it look like it refers to a chan from my perspective.

worker/runtime/process.go
```diff
@@ -58,6 +61,7 @@ func (p *Process) Wait() (int, error) {
 		return 0, fmt.Errorf("delete process: %w", err)
 	}
 
+	p.c.SetProperty(ProcessExitStatusKey, fmt.Sprintf("%d", status.ExitCode()))
```
taylorsilva: Yes, this is exactly what I was thinking of using! This looks good, continue down this path.

worker/runtime/process.go
```diff
@@ -15,15 +15,18 @@ import (
 type Process struct {
 	process     containerd.Process
 	exitStatusC <-chan containerd.ExitStatus
+	c           Container
```
taylorsilva: nit: naming style at the struct level is usually not single letters.
```suggestion
	container           Container
```

taylorsilva (changes requested):
This looks great. We should add a test for this. Looks like `worker/runtime/integration/` might be a good spot. Copying and modify this test might be a good start: https://github.com/concourse/concourse/blob/276676663fe0d06b5bbe5bc1687beb06eb483bd0/worker/runtime/integration/integration_test.go#L766-L769

Kump3r:
There is already a test for attach to a non-existent process - https://github.com/concourse/concourse/blob/276676663fe0d06b5bbe5bc1687beb06eb483bd0/worker/runtime/integration/integration_test.go#L743-L744. I think we can do similarly attach to a non-existent process that has an exit status property set to confirm that it is correctly propagated I guess.

worker/runtime/process.go
IvanChalukov: done

taylorsilva (approved):
LGTM - Thanks for fixing this! Really making containerd more robust here 💪

PR #9346, opened 2025-11-05
Added More Docs

hayden3456:
This PR adds source-linked, auto-generated docs for the codebase. Before you dismiss it: it includes inline source references, architecture diagrams, and I think it does a solid job. No hard feelings if you’d prefer not to include it.

If you find it useful, I’m building a documentation copilot to help explain why code is the way it is. It will:
- Auto-sync docs 
- Asks a few questions on PR's to gather understanding as to why
- Visualize the system at different levels of abstraction

I'm just building this out and would love to chat about what would help most so I can build something genuinely useful.
docforge.net

### DIFF
--- a/docs/autogenerated_docs/1-overview.md
+++ b/docs/autogenerated_docs/1-overview.md
@@ -1,0 +1,277 @@
+# Overview
+
+<details>
+<summary>Relevant source files</summary>
+
+The following files were used as context for generating this wiki page:
+
+- [.github/ISSUE_TEMPLATE/bug-report.md](https://github.com/concourse/concourse/blob/301f8064/.github/ISSUE_TEMPLATE/bug-report.md)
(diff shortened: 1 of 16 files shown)


taylorsilva:
Interesting... I can see you've been busy spamming similar PR's across various open-source projects. Not sure this is a good tactic.

This is a lot of stuff to just throw at a project. I don't want to accept it because it would take a lot of time for me to validate all the information in here and https://github.com/godotengine/godot/pull/112407#issuecomment-3491024390 indicates to me that this PR is probably full of similar issues, which is what I'd expect from an LLM-generated set of docs that hasn't been validated by a human.

Throwing a +1 to this comment as well from another PR: https://github.com/AssemblyScript/assemblyscript/pull/2959#issuecomment-3494087821

PR #9358, opened 2025-11-12
add `--oidc-issuer-url` flag to decouple OIDC issuer from external URL

b1n9s:

The IDToken/OIDC feature is currently unusable for Concourse deployments in private networks because cloud providers (AWS IAM OIDC, GCP Workload Identity, Azure Federated Credentials) need to fetch OIDC discovery metadata from the issuer URL, but private Concourse instances are not publicly routable.

  **Current limitation:**
  - Concourse generates JWT tokens with `iss` claim = `--external-url`
  - Cloud providers must fetch `<issuer>/.well-known/openid-configuration` and `<issuer>/.well-known/jwks.json`
  - For private deployments, these endpoints are unreachable from cloud provider networks
  - This makes workload identity federation impossible for security-conscious organizations that deploy Concourse privately

[Discussion](https://github.com/orgs/concourse/discussions/9349) 
  ### Solution

  Add an optional `--oidc-issuer-url` flag that allows the OIDC issuer to be configured separately from Concourse's external URL. This enables:

  1. Concourse runs privately at `https://concourse.internal.example.com`
  2. OIDC discovery served publicly at `https://oidc.example.com`
  3. Tokens contain `iss=https://oidc.example.com` for cloud provider verification
  4. Operators sync `.well-known` files to the public URL via their preferred method (CDN, reverse proxy, etc.)

  ### Changes

  **Core Implementation:**
  - [x] Added `--oidc-issuer-url` CLI flag to `atc/atccmd/command.go`
  - [x] Extended `ManagerFactory` to support separate OIDC issuer with fallback to external URL
  - [x] Updated IDToken server to use OIDC issuer in discovery endpoints
  - [x] Wired configuration through command → handler → token generator → API server

  **Testing:**
  - [x] Unit tests for token generation with custom issuer
  - [x] Unit tests for ManagerFactory fallback logic
  - [x] Integration test for OpenID discovery endpoint
  - [x] All existing tests continue to pass


taylorsilva:
Thanks for the PR! I've added it to my PR backlog to review.

atc/creds/idtoken/doc.go
taylorsilva, line 1: Having a `doc.go` file is not a pattern we follow. We don't write user/operator docs alongside the code. This will just get lost and forgotten in here. PR'ing this content into the docs repo would be great though, probably to this page: https://github.com/concourse/docs/blob/master/lit/docs/operation/creds/idtoken.lit

taylorsilva (changes requested):
Gave it a very quick look over to see if there was anything crazy in here, but looks solid. I'll give it a deeper look later.

atc/creds/idtoken/doc.go
b1n9s: Thank you very much for the review @taylorsilva, I've removed the doc file here and will create a PR in the docs repo later.

b1n9s:
And the PR for the documentation update here. https://github.com/concourse/docs/pull/573

analytically (approved):

atc/api/idtokenserver/server.go
```diff
@@ -10,17 +10,26 @@ import (
...
 	logger              lager.Logger
 	externalURL         string
+	oidcIssuer          string
 	dbSigningKeyFactory db.SigningKeyFactory
 }
 
 func NewServer(
 	logger lager.Logger,
 	externalURL string,
+	oidcIssuer string,
```
dbaumgarten, line 18: How about only passing oidcIssuer here and it either contains the value of the flag or of externalURL.
The server does not need to know where the issuer value is coming from, just what it is supposed to be

atc/creds/idtoken/manager_factory.go
```diff
@@ -30,6 +31,10 @@ func (factory *ManagerFactory) SetIssuer(issuer string) {
 	factory.issuer = issuer
 }
 
+func (factory *ManagerFactory) SetOIDCIssuer(oidcIssuer string) {
```
dbaumgarten: Ther is already a SetIssuer method. Just use it.

atc/atccmd/command.go
```diff
@@ -657,6 +658,9 @@ func (cmd *RunCommand) Runner(positionalArguments []string) (ifrit.Runner, error
 
 	idtoken.UpdateGlobalManagerFactory(func(f *idtoken.ManagerFactory) {
 		f.SetIssuer(cmd.ExternalURL.String())
```
dbaumgarten: Check if OIDCIssuerURL is set. If yes call SetIssuer(cmd.OIDCIssuerURL.String()), otherwise call f.SetIssuer(cmd.ExternalURL.String())

atc/api/handler.go
```diff
@@ -112,7 +113,7 @@ func NewHandler(
 	artifactServer := artifactserver.NewServer(logger, workerPool)
 	usersServer := usersserver.NewServer(logger, dbUserFactory)
 	wallServer := wallserver.NewServer(dbWall, logger)
-	idTokenServer := idtokenserver.NewServer(logger, externalURL, dbSigningKeyFactory)
+	idTokenServer := idtokenserver.NewServer(logger, externalURL, oidcIssuer, dbSigningKeyFactory)
```
dbaumgarten: Suggestion:

`
if oidcIssuer == "" {
oidcIssuer=externalURL
}
idTokenServer := idtokenserver.NewServer(logger, oidcIssuer, dbSigningKeyFactory)
`

dbaumgarten (changes requested):
Maybe it is just nitpicking, but I would very much like moving that "use configured issuer and fallback to externalURL"-logic out of the Server and the Factory.
They don't need to know where the value for the issuer is coming from. They just use the passed value.
Which value is passed to them is then decided based on Concourse's configuration.

atc/api/idtokenserver/server.go
b1n9s: updated

atc/creds/idtoken/manager_factory.go
b1n9s: updated.

atc/atccmd/command.go
b1n9s: updated.

atc/api/handler.go
b1n9s: updated.

b1n9s:
> Maybe it is just nitpicking, but I would very much like moving that "use configured issuer and fallback to externalURL"-logic out of the Server and the Factory. They don't need to know where the value for the issuer is coming from. They just use the passed value. Which value is passed to them is then decided based on Concourse's configuration.

Appreciate your review, @dbaumgarten. That's a reasonable call. I've updated the PR to resolve the URL before passing it to the server and factory. Thank you.

dbaumgarten (approved):
LGTM 👍

b1n9s:
Thank you everyone for your review. 
@taylorsilva I think this PR is at a good state now, would you give it another look when you have a sec, please? thank you.

taylorsilva:
It's on my list for tomorrow :)

taylorsilva (approved):

taylorsilva:
Thanks again for the PR! This will really improve things for users of this feature :)

PR #9362, opened 2025-11-17
fix(set_pipeline): evaluate InstanceVars ahead of other variables

### LINKED ISSUE #9356: Instance variables have lower precedence than var files when using set_pipeline steps
### Summary

I have a pipeline with some static variables I use to customize it. Defaults for all variables are provided in a YAML file. Some overrides are provided as instance variables via either `-i` for `fly` or `instance_vars` for `set_pipeline`.

If I set the pipeline via `fly` the pipeline is templetized with instance vars taking precedence over values from files with the same name. But if I use a `set_pipeline` step, values from var files actually have higher precedence, and instance overrides are not applied (they still qualify the name of the pipeline as expected).

A simple workaround it to specify overrides twice via both `vars` and `instance_vars`, which is pretty trivial using YAML anchors. But this lack of symmetry between the CLI and the step can be confusing.

### Steps to Reproduce

1. Fire up a local Concourse instance via `docker-compose` (I actually used `podman-compose` but I don't think this is relevant here)
2. Login to it as the `test` user
3. Use fly to apply [this sample pipeline](https://gist.github.com/binary-manu/78205678792dd77f83e8eb6365ad4492) with [this vars file ](https://gist.github.com/binary-manu/754499d7032199b95533556a68e5f092)
4. Manually trigger the `print` job
5.  Expect it to print exactly what the var file says.
6. Reapply the pipeline, this time adding `-i var2=something`
7. Manually trigger the `print` job of the new pipeline instance
8.  Expect it to print `something` for `var2`
9. Now apply [this parent pipeline](https://gist.github.com/binary-manu/260558ed6635f256fdc07c8bc04a36fc), which creates the same two instances as above, but via `set_pipeline`
10.  Manually trigger the `print` jobs in the new instances.
11. Now observe that in the instanced version of the pipeline, the override specified to `set_pipeline` via `instance_vars` is ignored and the output message in the task uses the text from the file.

### Expected Results

I expected the pipeline instance created via `set_pipeline` to give instance variables higher priority than vars from files, just like `fly` does.

### Actual Results

The instance variable is used to discriminate pipeline instances, but is ignored in the templatization, as the var file value is used instead.

### Web Node(s) configuration

_No response_

### Worker(s) configuration

_No response_

### Concourse Version

7.14.2

### Browser (if applicable)

Firefox 144.0.2 64-bit (Arch Linux Package)

### Did this use to work?

I just started with Concourse, thus I have no experience with previous versions

aeijdenberg:

closes #9356 

* [x] adds test to demonstrate broken behaviour
* [x] fix code make test pass

## Notes to reviewer

It's a shame the code used for `fly set-pipeline` and the `set_pipeline` step seem quite divergent - but that feels like a much bigger PR to look at, so this makes a smaller change to address the reported issue.

## Release Note

* `set_pipeline` step now allows `instance_vars` to override vars set via `vars`/`var_file`. This matches the behaviour of `fly set-pipeline`

### DIFF
--- a/atc/exec/set_pipeline_step.go
+++ b/atc/exec/set_pipeline_step.go
@@ -266,6 +266,12 @@ func (s setPipelineSource) FetchPipelineConfig() (atc.Config, error) {
 	}
 
 	staticVars := []vars.Variables{}
+
+	// add instance vars first so that they take precedence during evaluation later
+	if len(s.step.plan.InstanceVars) > 0 {
+		staticVars = append(staticVars, vars.StaticVariables(maps.Clone(s.step.plan.InstanceVars)))
+	}
+
 	if len(s.step.plan.Vars) > 0 {
 		staticVars = append(staticVars, vars.StaticVariables(s.step.plan.Vars))
 	}
@@ -281,10 +287,6 @@ func (s setPipelineSource) FetchPipelineConfig() (atc.Config, error) {
 		}
 
 		staticVars = append(staticVars, sv)
-	}
-
-	if len(s.step.plan.InstanceVars) > 0 {
-		staticVars = append(staticVars, vars.StaticVariables(maps.Clone(s.step.plan.InstanceVars)))
 	}
 
 	if len(staticVars) > 0 {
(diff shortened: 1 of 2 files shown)


analytically (approved):

taylorsilva:
Thanks for the PR! Adding it to my backlog.

taylorsilva (approved):
LGTM - Thanks for fixing this!

PR #9368, opened 2025-11-20
fix: vault userpass auth cannot be used

nouseforaname:
when trying to exchange the user / pass combination for a token, the userpass auth expects the PUT to be done against a path that has the desired username as its last element.
e.g.:

```
vault login -output-curl-string -method=userpass username="HEREIAM" password="notinpath"
curl -X PUT -H "X-Vault-Request: true" -d '{"password":"notinpath"}' https://my-vault-host.com/v1/auth/userpass/login/HEREIAM
```

this is already done for okta and ldap, this commit enables the appending logic that exists for okta and ldap to also be executed for the userpass engine.


taylorsilva (approved):
Thanks for the PR! LGTM

PR #9372, opened 2025-11-25
set containerd as the default runtime for linux workers

### LINKED ISSUE #9266: Change default runtime to `containerd`
### Summary

As per the title, change the default container runtime used by `concourse worker` to `containerd`.

### Why?

The speed of development is faster with containerd. We actually get contributions for our worker/runtime package. The Guardian runtime only just got support for cgroups v2 working this year while containerd has had it for years. There have been a bunch of issues where people say they can't start Concourse and the solution is "set the runtime to containerd". The OSS cluster has been using containerd for years now without issue. The containerd runtime feels very stable at this point and I feel confident setting it as the default runtime for Concourse moving forward.

### Context

Discussion: https://github.com/orgs/concourse/discussions/9184

taylorsilva:

closes #9266 

Makes containerd the default runtime for linux workers. See related issue for rationale.

Did a quick look over the codebase to see if this would break any tests; nothing obvious popped out to me. We'll see how the PR checks do. I expect no issues.

### DIFF
--- a/worker/workercmd/worker_linux_amd64.go
+++ b/worker/workercmd/worker_linux_amd64.go
@@ -1,5 +1,5 @@
 package workercmd
 
 type RuntimeConfiguration struct {
-	Runtime string `long:"runtime" default:"guardian" choice:"guardian" choice:"containerd" choice:"houdini" description:"Runtime to use with the worker. Please note that Houdini is insecure and doesn't run 'tasks' in containers."`
+	Runtime string `long:"runtime" default:"containerd" choice:"guardian" choice:"containerd" choice:"houdini" description:"Runtime to use with the worker. Please note that Houdini is insecure and doesn't run 'tasks' in containers."`
 }


Kump3r:
Just as a reminder I assume we would need to fix the bosh, helm, docs packaging stuff as well, after this is merged, or?

taylorsilva:
Good question, hadn't thought about packaging because I don't think anything needs to change at the packaging layer.

Helm chart has actually been defaulting to containerd: https://github.com/concourse/concourse-chart/blob/cfaf5f7e98845dc7df72d651c4f9f442e95aa648/values.yaml#L1743-L1745
Looks like I did that for test reasons though. Can probably revert it.

Same thing for Bosh: https://github.com/concourse/concourse-bosh-release/blob/6aecfd31d1b10b1bbc2b9a9f470c4d0c79caf766/jobs/worker/spec#L195-L200

Am I missing anything about the packaging stuff related to this PR?

You are correct about the docs. I'll make a note about that in the issue.

Kump3r:
Looking at https://github.com/concourse/concourse-bosh-release/blob/6aecfd31d1b10b1bbc2b9a9f470c4d0c79caf766/jobs/worker/spec#L195-L200, I assume that the default is still guardian for bosh, unless you change it with an OPS file. So I believe this should be changed as well for bosh release v8.x.x.
In the deployment there is nothing in the cluster, or light `concourse.yml` which sets the runtime to containerd. So I assume for a bosh VM if not overwritten by an OPS file, you will get guardian as default.

Kump3r:
So based on a little digging, the example we are setting to `guardian`, will "Example value, to be displayed in the [release index](https://bosh.io/docs/releases)." Sadly the link doesn't work and I am not sure what this index is, so I assume since we do not have default, we are good to go. Sorry for the noise, was not sure how this works. I assume it takes the default from the binary then.

Source: https://bosh.io/docs/jobs/#spec

taylorsilva:
The `example` field is there to show what kind of value that field accepts. It's not used as a default value. At the packaging level, we always try to fall back to whatever the binary's default's are so we can avoid situations where packaging and binary defaults drift.

Issue #4839, opened 2019-11-28
Progress bars and other indicators should not be 'buffered' with secret redaction enabled


vito:
## What challenge are you facing?

Same as https://github.com/concourse/concourse/issues/4799 - we have a regression fix in #4817 but it would be nice in the long run if secret redaction could handle these cases too.

## What would make this better?

Not sure, I think we'd need to test out a few situations.

For one example, a lot of test suites print out a bunch of dots on a single line (i.e. Ginkgo, Rspec). I'm not sure what heuristic to use here but it would be nice if I could actually see those printed one-by-one, as otherwise it'll look like my tests are hanging.

The other example I can think of is CLIs that print a progress bar. These tend to be a bunch of punctuation with `\r` printed to re-draw the progress bar. `fly sync` is probably a good example of this?

## Are you interested in implementing this yourself?

nope

Issue #4939, opened 2019-12-20
Executing tasks does not stream output


wezzynl:
## Bug Report

`fly execute` does not stream task output while running task.

## Steps to Reproduce

Run a task using `fly execute`.

## Expected Results

The output of the running task to stream into the terminal, seeing output as it happens.

## Actual Results

Output looks like being buffered and only shows when a task completes or errors. So `executing build X at https://Y` is what you see when you start the task and only after the task finishes or errors you see the output that the task has generated. Previously, in 5.7.0 and before, task output would stream into the terminal.

## Version Info

* Concourse version: 5.7.2
* Deployment type (BOSH/Docker/binary): Docker
* Infrastructure/IaaS: On-premise, Kubernetes
* Did this used to work? Yes, versions before 5.7.2.

pivotal-jamie-klassen:
@wezzynl can you give an example task? Does the same task in a pipeline have this issue too, or just `fly execute`? I also have to wonder - do you have [secret redaction](https://concourse-ci.org/creds-redacting.html) enabled?

It is striking that this worked for you in v5.7.0, since we noted that v5.7.0 suffered from #4799 and the point of putting #4817 into v5.7.2 was to fix issues like this!

wezzynl:
@pivotal-jamie-klassen I do not have secret redaction enabled (it's off by default, right?). Running the job in the web interface it works correctly, streaming output as it comes in.

Now that you mention 5.7.0 I'm not 100% sure anymore, the version before I upgraded might be indeed before 5.7.0. I am sure I'm running 5.7.2 now though. ;)

I ran a quick test with a simple task:

tasks/ci/tasks/test.yml
```yml
---
platform: linux

image_resource:
  type: registry-image
  source:
    repository: private.local.repo/some-image
    tag: latest
    username: ((registry-username))
    password: ((registry-password))

inputs:
  - name: tasks

run:
  path: ./tasks/ci/build/scripts/test.sh

```

tasks/ci/build/scripts/test.sh

```sh
#!/bin/sh

set -e

echo "line 1"
sleep 10
echo "line 2"
sleep 10
echo "line 3"
```

When running `fly -t ss execute --input tasks=../../ -c tasks/test.yml --load-vars-from=vars.yml` first it shows this:

```
uploading tasks done
executing build 96 at https://xxx/builds/96
initializing
running ./tasks/ci/build/scripts/test.sh
```

[...]

PR #9378, opened 2025-11-26
redact secrets is always enabled

### LINKED ISSUE #9376: Enable redact secrets by default
### Summary

The redact secrets feature should always be enabled. Similar to #9268 and #9269 

We've been using it for a whole and have seen no negative side-effects. We've also seen no reports from the community that there's any problems with the feature.

Other CI/CD platforms have the same feature and don't even give users a way to opt-out of it.

### Context

_No response_

taylorsilva:

closes #9376

This PR also tries to improve a few things in situations where logs were indefinitely buffered:

- Properly resolve the issue reported in https://github.com/concourse/concourse/issues/4799 by flushing when `\r` is detected
- Trying to improve on situations noted in https://github.com/concourse/concourse/issues/4839 by flushing any cached logs if the last write was > 1 second ago


analytically:
I read that it will stop 'streaming output' ... should we look into this deeper? There are also outstanding issues with redaction not working in all cases.

taylorsilva:
> I read that it will stop 'streaming output'

Redact secrets will stop streaming output? If you're referring to an existing issue _please_ link to it. I did a quick search and nothing came up in this repo.

> There are also outstanding issues with redaction not working in all cases.

If you're referring to the PR you just opened here: https://github.com/concourse/concourse/pull/9379
Those are edge-cases and not reason, imo, to not have redact secrets always enabled.

taylorsilva:
Should see if https://github.com/concourse/concourse/issues/4799 is an issue since it seems it was resolved by not redacting secrets.

This comment has some interesting context as well: https://github.com/concourse/concourse/issues/4749#issuecomment-555090809

Similar logs not streaming issue which got closed by the stale bot: https://github.com/concourse/concourse/issues/4939

taylorsilva:
Was able to reproduce https://github.com/concourse/concourse/issues/4799 on this branch! That is a blocker for getting this merged.

atc/engine/log_event_writer.go
```diff
@@ -94,11 +41,16 @@ func (writer *dbEventWriterWithSecretRedaction) Write(data []byte) (int, error)
...
+		writer.dangling = nil
 	}
 
 	payload := string(text)
 	if data != nil {
 		idx := strings.LastIndex(payload, "\n")
+		idxR := strings.LastIndex(payload, "\r")
+		if idx < idxR {
+			idx = idxR
+		}
```
taylorsilva: This should address #4799. Should see if there are some tests we can add though.

taylorsilva:
@Kump3r Have you guys seen any issues with the redact secrets feature not working as expected? Or logs not appearing and then suddenly appearing all at once?

Kump3r:
Not to my knowledge, and our operators even have a scan for secrets leaking in the build logs, so it should have caught it. We can try to reproduce something, are there steps defined somewhere we should try? And with which version of Concourse

taylorsilva:
Not aware of any reproducibles at this time. #4799 was the only one I could scrounge up but I think I've got that fixed in this PR. Trying to see if there are any known issues but sounds like there aren't. You guys do have redact secrets enabled, right?

taylorsilva:
Redacting secrets is definitely a hard problem. Even GH Actions makes no guarantees about redacting all secrets: https://docs.github.com/en/actions/reference/security/secure-use#use-secrets-for-sensitive-information

Kump3r:
Yes, since as far back as I can remember

atc/engine/log_event_writer.go
```diff
@@ -94,20 +44,29 @@ func (writer *dbEventWriterWithSecretRedaction) Write(data []byte) (int, error)
...
-			// No new-line found, then cache the log.
-			writer.dangling = text
-			return len(data), nil
+			// Avoid holding onto the buffer indefinitely by flushing if the
+			// last flush was more than 1 second ago.
+			if writer.clock.Since(writer.lastFlush) < time.Second {
+				// No new-line found, then cache the log.
+				writer.dangling = text
+				return len(data), nil
+			}
```
taylorsilva, line 68: I've added this logic to address #4839. The writer will flush whatever was passed in and whatever was in `writer.dangling`.

I manually tested this using this pipeline:

```yaml
jobs:
  - name: job
    plan:
      - in_parallel:
        - task: spinner
          config: &task
            platform: linux
            image_resource: &image
              type: mock
              source: { mirror_self: true }
            run:
              path: bash
              args:
                - -c
                - |
                  spin='-\|/'

                  i=0
                  for e in {1..100}; do
                    i=$(( (i+1) %4 ))
                    printf "\r${spin:$i:1}"
                    sleep .1
                  done
        - task: dots
          config:
            platform: linux
            image_resource: *image
            run:
              path: bash
              args:
                - -c
                - |
                  for e in {1..100}; do
                    echo -n "$e | "
                    sleep 0.1
                  done

```

Before this PR, neither of these steps print anything until they finished. With this PR, both steps print logs at regular intervals. The first step relies on us flushing if `\r` is detected. The second step relies on the 1 second flush being hit.

taylorsilva: I'm not sure if this "flush if last flush was >1s" ago is a good idea or not. It seems fine from what I've been able to read (and asking multiple LLM's 😅). Probably going to leave this open for a bit to see if anyone can think of some reason that this is a bad idea.

analytically: So this wasn't there before, but it did stream as secret redaction was off? But when it's on it doesn't stream?

analytically: Breaks secret redaction - The whole point of buffering until newline is to ensure secrets spanning chunk boundaries get redacted.

analytically: Image

taylorsilva: > Breaks secret redaction - The whole point of buffering until newline is to ensure secrets spanning chunk boundaries get redacted.

I'm going to assume you mean the 1s flushing. The use-case I'm trying to resolve here is

> For one example, a lot of test suites print out a bunch of dots on a single line (i.e. Ginkgo, Rspec). I'm not sure what heuristic to use here but it would be nice if I could actually see those printed one-by-one, as otherwise it'll look like my tests are hanging.

From #4839. I reference this in the PR description. The second step in the example I gave is what I'm trying to improve. I do want to improve those situations if I can.

I have been thinking about this over the weekend, wondering if 1s is too short. Having a hard time limit like this kinda sucks, but I want to avoid having us hold onto an ever growing line of text and users staring at a blank screen wondering if their step is running or not.

I wonder how much data gets passed into one call of `Write()`? I don't know if something upstream from `Write()` would split the data or not?

Would a higher flush limit be safer? Like 5 or 10 seconds? It feels like this would be a more conservative backstop. Wdyt?

I'm trying to push for defaults that keep users safe (default to redacting secrets) while also not making things look broken (holding onto text in the buffer until `Close()` is called). Those are the things we're trying to balance here.

---

Also, I don't think we can ever guarantee secrets won't be leaked in every scenario. Even Github Actions does not guarantee this:

> Because there are multiple ways a secret value can be transformed, automatic redaction is not guaranteed. Adhere to the following best practices to limit risks associated with secrets.

Source: https://docs.github.com/en/actions/reference/security/secure-use

And this has been discussed extensively that we'll never cover all scenarios.

taylorsilva: @drich10 and I had a chat about this PR. Derek felt fine proceeding with what's proposed in this PR, acknowledging the fact that yes, a secret may leak if we flush when we've only received a partial secret.

Another thing I realized was that without the time-based flushing, it's possible for the buffer to grow indefinitely. For example if you do:

```bash
while true; do
  echo -n "."
done
```

in a task, this will print an endless stream of dots that will never get printed to the console and will grow the buffer on the web node indefinitely. Not ideal!

There's also the user's perspective of not seeing anything getting printed and wondering if Concourse is hanging or not.

We'll adjust in the future based on user feedback.

PR #4526, opened 2019-09-30
atc: use ((redacted)) syntax, update release notes

vito:
Signed-off-by: Alex Suraci <suraci.alex@gmail.com>

PR #9379, opened 2025-11-27
fix(atc): prevent partial redaction of overlapping secrets

analytically:
Secrets redaction had two bugs:

1. When one secret was a substring of another (e.g., "pass" and
   "password"), map iteration order determined replacement order.
   If "pass" replaced first, "password" became "((redacted))word".

2. Secrets containing substrings of "((redacted))" (e.g., "redact")
   corrupted previously inserted markers, producing mangled output
   like "((((redacted))ed))".

Fix: use a bitmask to track redacted byte positions in the original
string. Consecutive masked regions collapse to a single "((redacted))"
marker in one final output pass.

Also use utf8.RuneCountInString for the single-char check to correctly
handle multi-byte characters.

### DIFF
--- a/atc/engine/build_step_delegate.go
+++ b/atc/engine/build_step_delegate.go
@@ -8,6 +8,7 @@ import (
 	"maps"
 	"strings"
 	"time"
+	"unicode/utf8"
 
 	"code.cloudfoundry.org/clock"
 	"code.cloudfoundry.org/lager/v3"
@@ -424,7 +425,7 @@ func (delegate *buildStepDelegate) checkPolicy(input policy.PolicyCheckInput) er
 func (delegate *buildStepDelegate) buildOutputFilter(str string) string {
 	it := &credVarsIterator{line: str}
 	delegate.state.IterateInterpolatedCreds(it)
-	return it.line
+	return it.Line()
 }
 
 func (delegate *buildStepDelegate) redactImageSource(source atc.Source) (atc.Source, error) {
@@ -446,15 +447,55 @@ func (delegate *buildStepDelegate) ContainerOwner(planId atc.PlanID) db.Containe
 }
 
 type credVarsIterator struct {
-	line string
+	line    string
+	secrets []string
 }
 
 func (it *credVarsIterator) YieldCred(name, value string) {
 	for _, lineValue := range strings.Split(value, "\n") {
 		lineValue = strings.TrimSpace(lineValue)
-		// Don't consider a single char as a secret.
-		if len(lineValue) > 1 {
-			it.line = strings.Replace(it.line, lineValue, "((redacted))", -1)
+		if utf8.RuneCountInString(lineValue) > 1 {
+			it.secrets = append(it.secrets, lineValue)
 		}
 	}
+}
+
+func (it *credVarsIterator) Line() string {
+	if len(it.secrets) == 0 {
+		return it.line
+	}
+
+	mask := make([]bool, len(it.line))
+
+	for _, secret := range it.secrets {
+		slen := len(secret)
+		start := 0
+		for {
+			idx := strings.Index(it.line[start:], secret)
+			if idx == -1 {
+				break
+			}
+			idx += start
+			for i := idx; i < idx+slen; i++ {
+				mask[i] = true
+			}
+			start = idx + 1
+		}
+	}
+
+	var out strings.Builder
+	out.Grow(len(it.line))
+	i := 0
+	for i < len(it.line) {
+		if mask[i] {
+			out.WriteString("((redacted))")
+			for i < len(it.line) && mask[i] {
+				i++
+			}
+		} else {
+			out.WriteByte(it.line[i])
+			i++
+		}
+	}
+	return out.String()
 }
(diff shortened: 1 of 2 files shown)


atc/engine/build_step_delegate.go
```diff
@@ -446,15 +448,26 @@ func (delegate *buildStepDelegate) ContainerOwner(planId atc.PlanID) db.Containe
...
+		if utf8.RuneCountInString(lineValue) > 1 {
+			it.secrets = append(it.secrets, lineValue)
 		}
 	}
 }
+
+func (it *credVarsIterator) Line() string {
+	sort.Slice(it.secrets, func(i, j int) bool {
+		return len(it.secrets[i]) > len(it.secrets[j])
+	})
```
taylorsilva: Should we do the sorting in `YieldCred()` instead, which I expect gets called a lot less often than the `Line()` func here does. Even though the slice will be sorted after the first call to `Line()`, each call will still have CPU costs because `sort.Slice()` will go through the entire slice to ensure it's sorted.

From what I can tell, `YieldCred()` will be called every time a new secret is used in a build, while `Line()` will get constantly called as logs are generated.

atc/engine/build_step_delegate_test.go
```diff
@@ -1050,6 +1050,56 @@ var _ = Describe("BuildStepDelegate", func() {
...
+					writer = delegate.Stdout()
+					writer.Write([]byte("the mysecret value"))
+					writer.(io.Closer).Close()
+
+					Expect(fakeBuild.SaveEventCallCount()).To(Equal(1))
+					payload := fakeBuild.SaveEventArgsForCall(0).(event.Log).Payload
+					Expect(payload).To(Equal("the ((redacted)) value"))
+					Expect(payload).ToNot(ContainSubstring("(((("))
+				})
+			})
```
taylorsilva, line 1101: I don't think it's worth the CPU cycles (the two calls to `strings.ReplaceAll`) to deal with this edge case. At worse, someone has a secret with the value `redacted`. Even less likely, someone has a string that is a substring of `redacted` like `red` or `act`, or even a single letter of the word "redacted". I doubt anyone has as a secret like that or cares about actually trying to hide it.

It also feels like you traded one edge case for another by using `\x00` as the intermediate string. Logs contain mostly programmatic output and some programs may use `\x00`. A quick search tells me that `\x00` can be used as a delimiter in some CLI programs. This could result in a bunch of `((redacted))` appearing everywhere in someone's logs in seemingly random places.

taylorsilva (changes requested):

atc/engine/build_step_delegate_test.go
analytically: It's only a single extra call. The x00 approach works and is the best option IMO.

taylorsilva: Okay here's another idea. Instead of `((redacted))` being the string we replace things with, we use `((***))` or `***` instead? It reduces the surface area of the edge case from "all substrings of redacted" to three asterisks. A lone secret having a value of `*`, `**`, or `***` feels _incredibly_ unlikely.
Wdyt?

taylorsilva: I am very against doing a string replacement of `\x00` because it makes things less predictable for users. As-is, users can understand what happens if they see `((((redacted))ed))`. They can change the value of their secret to avoid this issue.

With this PR, if any of the log output contains `\x00` a users log output would become a complete mess of `((redacted))`'s and it would be out of the users control, unlike in the current situation.

analytically: Valid UTF-8 never contains \x00 as part of a multi-byte character - it only appears as the literal null character. But null has no legitimate use in text output. Go strings are byte slices, not null-terminated like C. So \x00 is just another byte - safe to use in strings operations, but guaranteed not to appear in real credential values or log output.

analytically: Ok I'll change the marker. Make it more strange.

taylorsilva: Was curious how we even settled on `((redacted))`. The history:
- Started as `[***redacted**]` in the original PR https://github.com/concourse/concourse/pull/4311
- Became `((redacted))` in https://github.com/concourse/concourse/pull/4526 because Marie Kondo vibes ✨

analytically: I've used a bitmask instead of a market.

taylorsilva: For example:

```yaml
resources:
  - name: file
    type: mock
    source:
      create_files:
        some_file: the-secret

jobs:
  - name: job
    plan:
      - get: file
      - load_var: secret
        file: file/some_file
      - task: echo
        config: &task
          platform: linux
          image_resource: &image
            type: mock
            source: { mirror_self: true }
          run:
            path: bash
            args:
              - -c
              - |
                echo ((.:secret))
                echo -e "Hello\x00World"
                find . -print0
```

Results in:
image

analytically: Yes, that's the whole point. X00 doesn't transmit, doesn't show. So use it as marker in a byte seq. But I've moved on, it's a bitmask now, better in all cases.

analytically: Bitmask now.

atc/engine/build_step_delegate.go
analytically: Sort not needed with bitmask approach.

PR #9380, opened 2025-11-27
fix(engine): clear dangling bytes after flush in redacting writer

analytically:
When Close() calls Write(nil), the dangling bytes are saved but never cleared. Calling Close() twice would duplicate the final output.


taylorsilva:
Closing because I think we already resolved this in #9378

PR #9381, opened 2025-12-01
Show pipeline last updated tooltip on pipeline name hover

analytically:
Add tooltip to pipeline breadcrumb name showing "pipeline last updated at <datetime>" when user hovers over the pipeline name in the top bar.

### DIFF
--- a/web/elm/src/Concourse.elm
+++ b/web/elm/src/Concourse.elm
@@ -1143,6 +1143,7 @@ type alias Pipeline =
     , public : Bool
     , teamName : TeamName
     , groups : List PipelineGroup
+    , lastUpdatedAt : Time.Posix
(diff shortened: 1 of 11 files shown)


analytically:
screentshot-2025-12-01 at 09 43 22@2x

taylorsilva (commented):
Works fine on instance groups too:
image

Looks like we're just missing tests. I think they would go here: https://github.com/concourse/concourse/blob/master/web/elm/tests/PipelineTests.elm
and
https://github.com/concourse/concourse/blob/master/web/elm/tests/TopBarTests.elm

Looks good otherwise!

analytically:
Done!

taylorsilva (approved):

PR #9385, opened 2025-12-02
Windows/baggageclaim: Retry initializing volumes on "Access is denied" errors

### LINKED ISSUE #9352: Windows Worker Volume Creation error
### Summary

Creating resources on Windowns intermittenly results in an access denied error trying to move the newly created folder from init to 

Image

```
error: rename H:\\work\\volumes\\init\\29b4d346-3fa4-4355-aab8-13d683c5ebd3 H:\\work\\volumes\\live\\29b4d346-3fa4-4355-aab8-13d683c5ebd3: Access is denied.
```

### Steps to Reproduce

Configure concourse with a windows worker
create a pipeline with a job that has a resources and runs steps on windows
run the pipeline alot.  you might or might not be able to reproduce.  I can reliably reproduce it on my VMs in my environment.

### Expected Results

Resource created and build continues

### Actual Results

Error saying the resource couldn't be created.

### Web Node(s) configuration

_No response_

### Worker(s) configuration

_No response_

### Concourse Version

7.14.2

### Browser (if applicable)

_No response_

### Did this use to work?

Yes but many versions ago and on a previous VM

taylorsilva:

closes #9352 (I hope)

Also looked at adjacent code to the main change made in this PR for anything we could be forgetting to close and found two cases where we fail to close our file handles.

I don't have a Windows worker setup to reproduce this issue myself, so this is mostly hoping it's fixed based on this comment: https://github.com/concourse/concourse/issues/9352#issuecomment-3512728939

### DIFF
--- a/go-archive/tarfs/extract.go
+++ b/go-archive/tarfs/extract.go
@@ -98,6 +98,7 @@ func ExtractEntry(header *tar.Header, dest string, input io.Reader, chown bool) 
 
 		_, err = io.Copy(file, input)
 		if err != nil {
+			file.Close()
 			return err
 		}
 
--- a/go-archive/tgzfs/extract.go
+++ b/go-archive/tgzfs/extract.go
@@ -23,6 +23,7 @@ func Extract(src io.Reader, dest string) error {
 	if err != nil {
 		return err
 	}
+	defer gr.Close()
 
 	tarReader := tar.NewReader(gr)
 
--- a/worker/baggageclaim/volume/filesystem.go
+++ b/worker/baggageclaim/volume/filesystem.go
@@ -3,6 +3,8 @@ package volume
 import (
 	"os"
 	"path/filepath"
+	"strings"
+	"time"
 )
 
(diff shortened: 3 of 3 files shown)


analytically:
Could this be relevant? https://github.com/golang/go/issues/76360

taylorsilva:
Interesting find! I don't think it's related though from looking at the referenced PR's on Gerrit. It looks like they only added stuff that gets used by `os.RemoveAll` which we are not using.

Just to be sure, I tracked down what `os.Rename` calls and it ends up here: https://cs.opensource.google/go/go/+/refs/tags/go1.25.5:src/internal/syscall/windows/syscall_windows.go;l=355-365;drc=433c01e94efe72f985096839d0631103b3163be6
Which ends up calling here: https://cs.opensource.google/go/go/+/refs/tags/go1.25.5:src/internal/syscall/windows/zsyscall_windows.go;l=418-424;drc=433c01e94efe72f985096839d0631103b3163be6

No interaction with any of the functions used by `os.RemoveAll`. Good find though, thanks for sharing.

PR #9387, opened 2025-12-04
fix(skymarshal): use HMAC-signed stateless OAuth state tokens

### LINKED ISSUE #9036: unexpected state token on successful auth callback
## Summary

We use OIDC to authenticate to our Concourse instance. When logging in to Concourse a cookie is set called `skymarshal_state`, this cookie contains a base64 encoded json containing redirect URI and Entropy, for example:

```json
{"redirect_uri":"/teams/example-team-name/pipelines/example-pipeline-name","entropy":"1b3068d653cfe7aa6c2b184421ad7cd52f17f95258a6cdb9afd6411853a1a456"}
```

When a user next successfully reauthenticates after the authentication cookie times out this `skymarshal_state` cookie is compared with the state passed in the URL:

```json
https://concourse-url.com/sky/callback?code=the-code&state=state-that-looks-like-the-above-json
```

[The following code then compares the two](https://github.com/concourse/concourse/blob/67b05c396eb899f3ba0ba21ff159b0c8f4cbbec1/skymarshal/skyserver/skyserver.go#L134-L138), resulting in a failed comparison 
```go
	if stateToken != r.FormValue("state") {
		logger.Error("failed-with-unexpected-state-token", errors.New("state token does not match"))
		http.Error(w, "unexpected state token", http.StatusBadRequest)
		return
	}
```

This results in `unexpected state token` being returned to the user.

Additionally, if the cookie expires then [the check above this](https://github.com/concourse/concourse/blob/67b05c396eb899f3ba0ba21ff159b0c8f4cbbec1/skymarshal/skyserver/skyserver.go#L128-L132):
```go
	if stateToken == "" {
		logger.Error("failed-with-invalid-state-token", errors.New("state token is empty"))
		http.Error(w, "invalid state token", http.StatusBadRequest)
		return
	}
```
Will fail, resulting in `invalid state token`.

**Additional Information**
The redirect URI in the cookie contains the path set upon first login.

The state being passed in the URL contains a reference to the pipeline you are viewing when the authentication expiration occurs.

It is likely that the pipeline that the user is looking at differs from the pipeline they were looking at when they previously logged in (we have many pipelines that our users travel between in the course of their work).

This happens frequently due to the configured lifetime of the tokens from our OIDC provider.

Once the issue occurs all open Concourse tabs are impacted with `invalid state token` 

## Steps to reproduce

1. Configure a login with OIDC (this may not have an impact, just our situation)
2. Login to set cookies in the browser
3. Navigate to a different pipeline / URL
4. Wait until reauthentication is required
5. Reauthenticate
6. See `unexpected state token` after authenticating successfully

## Expected results
Users should not encounter errors, after logging in users should be redirected to the URL presented in the query part of the URL.

## Actual results
`unexpected state token` error presented with a HTTP 400.

## Triaging info

* Concourse version: 7.8.2, but this check is present in master.
* Browser (if applicable): Firefox, Edge (Chromium)
* Did this used to work? No

analytically:
Two bugs caused intermittent "unexpected state token" errors:

1. base64.StdEncoding uses +, /, = which get mangled in URLs
2. Multiple browser tabs race to overwrite the single state cookie

Replace cookie-based state validation with self-validating HMAC-signed
tokens. State now includes timestamp, entropy, and signature - verified
entirely from the URL parameter without cookie comparison.

Signing key derived from OAuth client credentials (HMAC of clientID
with clientSecret as key), avoiding new configuration.

Fixes #9036

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -2,6 +2,8 @@ package atccmd
 
 import (
 	"context"
+	"crypto/hmac"
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
 	"database/sql"
@@ -1996,12 +1998,25 @@ func (cmd *RunCommand) constructSkyHandler(
 		HTTPClient:         httpClient,
 		ClaimsCacher:       claimsCacher,
 		AccessTokenFactory: accessTokenFactory,
+		StateSigningKey: deriveStateSigningKey(
+			oauth2Config.ClientID,
+			oauth2Config.ClientSecret,
+			cmd.Postgres.User,
+			cmd.Postgres.Password),
 	})
 	if err != nil {
 		return nil, err
 	}
 
 	return skyserver.NewSkyHandler(skyServer), nil
+}
+
+func deriveStateSigningKey(clientID, clientSecret, dbUser, dbPassword string) []byte {
+	mac := hmac.New(sha256.New, []byte(clientSecret))
+	mac.Write([]byte(clientID))
+	mac.Write([]byte(dbUser))
+	mac.Write([]byte(dbPassword))
+	return mac.Sum(nil)
 }
 
 func (cmd *RunCommand) constructTokenVerifier(claimsCacher accessor.AccessTokenFetcher) accessor.TokenVerifier {
--- a/skymarshal/skyserver/skyserver.go
+++ b/skymarshal/skyserver/skyserver.go
@@ -2,7 +2,9 @@ package skyserver
 
 import (
 	"context"
+	"crypto/hmac"
 	"crypto/rand"
+	"crypto/sha256"
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
@@ -29,6 +31,7 @@ type SkyConfig struct {
 	HTTPClient         *http.Client
 	AccessTokenFactory db.AccessTokenFactory
 	ClaimsCacher       accessor.AccessTokenFetcher
+	StateSigningKey    []byte
 }
 
 func NewSkyHandler(server *SkyServer) http.Handler {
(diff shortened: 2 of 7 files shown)


atc/atccmd/command.go
```diff
@@ -1991,6 +1993,7 @@ func (cmd *RunCommand) constructSkyHandler(
 		HTTPClient:         httpClient,
 		ClaimsCacher:       claimsCacher,
 		AccessTokenFactory: accessTokenFactory,
+		StateSigningKey:    deriveStateSigningKey(oauth2Config.ClientID, oauth2Config.ClientSecret),
```
analytically: hmm not sure how this acts when client id and secret are not unique, aka the default

skymarshal/skyserver/skyserver.go
```diff
@@ -236,20 +234,66 @@ func (s *SkyServer) Logout(w http.ResponseWriter, r *http.Request) {
 type stateToken struct {
 	RedirectURI string `json:"redirect_uri"`
 	Entropy     string `json:"entropy"`
+	Timestamp   int64  `json:"ts"`
+	Signature   string `json:"sig,omitempty"`
 }
 
-func encode(token stateToken) string {
-	json, _ := json.Marshal(token)
+const stateTokenMaxAge = 3600 // 1 hour
```
taylorsilva: What happens after 1hr? Default login time is 24hrs and configured with `CONCOURSE_AUTH_DURATION`

taylorsilva (dismissed):
I confirmed that this does fix #9036

LGTM - Resolve the merge conflicts and I'll merge

taylorsilva:
I was able to push to your branch and fix the conflicts 😁 Will let the PR checks run and then merge

PR #9389, opened 2025-12-05
Add SameSite attribute to authentication and CSRF cookies

### LINKED ISSUE #9388: Cookies are missing the SameSite attribute
### Summary

Cookies are attached to cross-site requests in a number of cases. When nothing prevents a browser from attaching the cookies and there are no security controls against CSRF attacks, an attacker may perform the CSRF attack. The SameSite attribute defines the scope of a cookie, i.e., determines when the browser adds the cookie to HTTP requests. Secure configurations restrict cookies to the same-site context. The SameSite attribute has three possible values defining the browser's behavior: - Strict - A cookie with the Strict attribute value is attached only to same-site requests. - Lax - Works similarly to Strict with one addition: such cookie is added to cross-site requests that are top-level navigation using "safe "HTTP methods." - None - A cookie with the None attribute value is attached to any request. When the attribute value is not specified explicitly, modern browsers imply that the attribute is set to Lax. Modern browsers allow cookies with the SameSite attribute explicitly set to None only along with the Secure attribute set.

In the `skymarshal/token/middleware.go`, the cookies are created without the safer SameSite attribute. Although modern browsers assume and block requests against CSRF, we should still add this layer of protection imho.

### Steps to Reproduce

docker compose up and create a cookie

### Expected Results

_No response_

### Actual Results

_No response_

### Web Node(s) configuration

_No response_

### Worker(s) configuration

_No response_

### Concourse Version

master code

### Browser (if applicable)

_No response_

### Did this use to work?

Do not think so

Kump3r:

closes #9388

* Add SameSite attribute to authentication and CSRF cookies, to add a layer of CSRF protection

Docs: https://pkg.go.dev/net/http#SameSite

### DIFF
--- a/skymarshal/token/middleware.go
+++ b/skymarshal/token/middleware.go
@@ -39,6 +39,7 @@ func (m *middleware) UnsetAuthToken(w http.ResponseWriter) {
 		MaxAge:   -1,
 		Secure:   m.secureCookies,
 		HttpOnly: true,
+		SameSite: http.SameSiteLaxMode,
 	})
 }
 
@@ -50,6 +51,7 @@ func (m *middleware) SetAuthToken(w http.ResponseWriter, tokenStr string, expiry
 		Expires:  expiry,
 		HttpOnly: true,
 		Secure:   m.secureCookies,
+		SameSite: http.SameSiteLaxMode,
 	})
 
 	return nil
@@ -70,6 +72,7 @@ func (m *middleware) UnsetCSRFToken(w http.ResponseWriter) {
 		MaxAge:   -1,
 		Secure:   m.secureCookies,
 		HttpOnly: true,
+		SameSite: http.SameSiteLaxMode,
 	})
 }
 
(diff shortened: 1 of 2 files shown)


Kump3r:
### Before:
Before

### After:
After

### Tested:

Fly and normal login - no regressions - ✅

taylorsilva (approved):
LGTM

Even though we didn't explicitly set it, by default browsers use Lax mode if `SameSite` is not specified.

> If no SameSite attribute is set, the cookie is treated as Lax by default.

Src: https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies#controlling_third-party_cookies_with_samesite

Kump3r:
Yeah I mentioned it in the description of the issue, but I guess if you use a custom browser, or a really old one it would be nice to have this extra layer of security hardening for the users. That is why I marked is as enhancement, not a bug. Thanks for merging it!

PR #9406, opened 2025-12-28
fix: across step showing failed when nested retry succeeds

### LINKED ISSUE #7565: across step retry-status bug
## Summary

When an across's nested step has attempts on it, the across gets marked as failed, if any of the attempts fail, rather than if all of them fail.

## Steps to reproduce

1. Fly this pipeline:
```
groups:
- name: main
  jobs:
  - across-with-retries

jobs:
- name: across-with-retries
  plan:
  - across:
    - var: i
      values: [1]
      max_in_flight: all
    do:
    - task: echo-release
      attempts: 3
      config:
        image_resource:
          type: registry-image
          source: { repository: bash }
        platform: linux
        run:
          path: bash
          args:
          - -c
          - |
            echo "Seeing if we're on an even second or odd"
            exit $(expr $(date +%s) % 2)
resources: []
```
2. Run the across-with-retries job a bunch of times
3. Look through the results until you see a job that had multiple attempts.

## Expected results

The `across` step should be marked as green/successful

## Actual results

The across step is incorrectly marked as red, while at the same time the pipeline is marked as green, and the nested step is marked as green (correctly).

## Triaging info

* Concourse version: 7.4.0
* Browser (if applicable):
* Did this used to work?
Screen Shot 2021-09-17 at 2 43 50 PM

analytically:
activeStepIds for Retry used takeWhile to collect attempts until success, which excluded the successful attempt itself. Parent steps then only saw failed attempts when computing severity.

Changed to only include the last active attempt's steps, since that represents the retry's actual outcome.

Fixes #7565

### DIFF
--- a/web/elm/src/Build/StepTree/Models.elm
+++ b/web/elm/src/Build/StepTree/Models.elm
@@ -160,6 +160,7 @@ mostSevereStepState model stepTree =
                 case step.buildStep of
                     Concourse.BuildStepDo _ ->
                         state
+
                     _ ->
                         case stepStateOrdering step.state state of
                             LT ->
@@ -323,8 +324,10 @@ activeStepIds model tree =
         Retry _ trees ->
             trees
                 |> Array.toList
-                |> List.Extra.takeWhile (mostSevereStepState model >> (/=) StepStateSucceeded)
-                |> List.concatMap (activeStepIds model)
+                |> List.filter (treeIsActive model)
+                |> List.Extra.last
+                |> Maybe.map (activeStepIds model)
+                |> Maybe.withDefault []
 
 
 updateTreeNodeAt : StepID -> (StepTree -> StepTree) -> StepTree -> StepTree


analytically:
screentshot-2025-12-28 at 11 59 03@2x

analytically:
Integration failure seems unrelated

IvanChalukov (approved):
Looks good to me. I tested it locally using the example from #7565 and everything worked as expected:
```yaml
groups:
- name: main
  jobs:
  - across-with-retries

jobs:
- name: across-with-retries
  plan:
  - across:
    - var: i
      values: [1]
      max_in_flight: all
    do:
    - task: echo-release
      attempts: 3
      config:
        image_resource:
          type: registry-image
          source: { repository: bash }
        platform: linux
        run:
          path: bash
          args:
          - -c
          - |
            echo "Seeing if we're on an even second or odd"
            exit $(expr $(date +%s) % 2)
```

PR #9419, opened 2026-01-08
Add metadata from get steps to the local build var

PentaHelix:

Metadata from `get` steps are added to the local build var `((.:))` under the name of the get step. E.g.:

```yaml
- get: repo
```
Will then have all it's metadata available under the var `((.:repo.<metadata-field-name))`.

Example from CI:
image
Will then have the following vars available in the build for later steps to use:
* `((.:repo.commit))`
* `((.:repo.author))`
* `((.:repo.author_date))`
* etc.

### DIFF
--- a/atc/exec/get_step.go
+++ b/atc/exec/get_step.go
@@ -253,6 +253,8 @@ func (step *GetStep) run(ctx context.Context, state RunState, delegate GetDelega
 			delegate.UpdateResourceVersion(logger, step.plan.Resource, versionResult)
 		}
 
+		state.AddLocalVar(step.plan.Name, AsMap(versionResult.Metadata), false)
+
 		succeeded = true
 	}
 
@@ -525,4 +527,12 @@ func (step *GetStep) resourceMountVolume(mounts []runtime.VolumeMount) runtime.V
 		}
 	}
 	return nil
+}
+
+func AsMap(metadata []atc.MetadataField) map[string]any {
+	result := make(map[string]any, len(metadata))
+	for _, kv := range metadata {
+		result[kv.Name] = kv.Value
+	}
+	return result
 }
--- a/atc/exec/get_step_test.go
+++ b/atc/exec/get_step_test.go
@@ -734,6 +734,11 @@ var _ = Describe("GetStep", func() {
 			Expect(fakeDelegate.UpdateResourceVersionCallCount()).To(Equal(1))
 		})
 
+		It("adds metadata to the build variables", func() {
+			value, _, _ := runState.Get(vars.Reference{Source: ".", Path: getPlan.Name, Fields: []string{"some"}})
+			Expect(value).To(Equal("metadata"))
+		})
+
 		It("does not return an err", func() {
 			Expect(stepErr).ToNot(HaveOccurred())
 		})


atc/exec/get_step.go
```diff
@@ -526,3 +528,11 @@ func (step *GetStep) resourceMountVolume(mounts []runtime.VolumeMount) runtime.V
 	}
 	return nil
 }
+
+func AsMap(metadata []atc.MetadataField) map[string]any {
```
PentaHelix, line 532: There is probably a better place for this function

taylorsilva:
@PentaHelix Thanks for the PR! I have added you the `all` team in the `concourse` github org, which will allow you to see the build logs for the PR checks. You'll probably get an email to accept the invite.

taylorsilva:
Thanks for the PR! Will review after I get 8.0 out.

atc/exec/get_step.go
taylorsilva: I'd add it onto a new type of `[]MetadataField` here: https://github.com/concourse/concourse/blob/e81401c0b9bffc7107a30df0e6fcb189b4ad7d75/atc/resource_types.go#L10-L13

Something like:

```go
type Metadata []MetadataField

func (m Metadata) AsMap() map[string]any {
	result := make(map[string]any, len(m))
	for _, v := range m {
		result[v.Name] = v.Value
	}
	return result
}
```

Then `VersionResult` can also use that type: https://github.com/concourse/concourse/blob/e81401c0b9bffc7107a30df0e6fcb189b4ad7d75/atc/resource/resource.go#L19-L22

Then `AsMap(versionResult.Metadata)` would become `versionResult.Metadata.AsMap()`!

taylorsilva (approved):
LGTM! I suggested how you can refactor `AsMap()` if you want to. As-is though, this works as expected.

Given this pipeline:

```yaml
resources:
  - name: metadata
    type: mock
    source:
      metadata:
        - name: hello
          value: from metadata
        - name: hope
          value: it works!

jobs:
  - name: job
    plan:
      - get: metadata
      - task: echo
        config:
          platform: linux
          image_resource:
            type: mock
            source:
              mirror_self: true
          run:
            path: bash
            args:
              - -c
              - |
                echo ((.:metadata.hello))
                echo ((.:metadata.hope))
```

It worked!

image

If you don't have time to refactor I'll merge it as-is and refactor it later. Let me know. Thanks again for the PR!

kcbimonte:
To prevent possible collisions with Local Vars, wonder if it'd be beneficial to use the resource name instead of a `.` to denote that it's coming from a resource instead of a local var. 

Like if I have a local var in the job of metadata and the mock resource providing metadata, which one "wins"?

taylorsilva:
@PentaHelix asked about that too. My response: https://github.com/orgs/concourse/discussions/9417#discussioncomment-15479153

After this PR, if someone currently has a pipeline that has a get step with the same name as `load_var` step, nothing will functionally change for them. The `load_var` will overwrite the metadata.

The part before the colon is also meant to be used to refer to `var_sources`, so this would introduce a new risk of step names colliding with `var_sources` which I think is a worse situation to put users in.

PR #9422, opened 2026-01-12
db: optimize IsFirstOccurrence query for jobs with many builds

### LINKED ISSUE #8160: Improve `builds_of_job` bottleneck query
## Summary

In my stress environment, the database is pinned to around 95% CPU and it seems like the most troublesome query is the following: 

```
WITH builds_of_job AS (
 	SELECT id FROM builds WHERE job_id = $1
 ) 
 SELECT EXISTS ( 
 	SELECT 1
 	FROM build_resource_config_version_inputs i 
 	JOIN builds_of_job b ON b.id = i.build_id 
 	WHERE i.name = $2
 	AND i.version_md5 = $3
 	AND i.resource_id = $4
 );
```

The following image is a screenshot of the db inspection query I ran against my stress env 

image

You can see that the `builds_of_job` query takes significantly more avg and total time than the other queries below it.

This query has been worked on many times and we even had an PR that I created a while ago that reworks it further https://github.com/concourse/concourse/pull/6403. I think we ended up not completing the work because the environment we were seeing the query be slow on no longer showed the slowness, but now that I have the stress environment I can do the rest of the work to improve this query.

When I run the query on my stress env, I get the following explain analyze results:

```
Result  (cost=25768.26..25768.27 rows=1 width=1) (actual time=2334.258..2334.259 rows=1 loops=1)
  CTE builds_of_job
    ->  Index Only Scan using builds_ordered_by_job_id_idx on builds  (cost=0.56..24932.55 rows=34333 width=8) (actual time=0.870..2159.823 rows=35377 loops=1)
          Index Cond: (job_id = 1)
          Heap Fetches: 4713
  InitPlan 2 (returns $1)
    ->  Hash Join  (cost=20.30..835.72 rows=1 width=0) (actual time=2334.257..2334.257 rows=1 loops=1)
          Hash Cond: (b.id = i.build_id)
          ->  CTE Scan on builds_of_job b  (cost=0.00..686.66 rows=34333 width=8) (actual time=0.880..2226.518 rows=35377 loops=1)
          ->  Hash  (cost=20.29..20.29 rows=1 width=8) (actual time=3.416..3.416 rows=3 loops=1)
                ->  Index Scan using build_inputs_resource_versions_idx on build_resource_config_version_inputs i  (cost=0.56..20.29 rows=1 width=8) (actual time=1.365..3.412 rows=3 loops=1)
                      Index Cond: ((resource_id = 1) AND (version_md5 = 'b5f0011a6d168c264ea8595b1696d62a'::text))
                      Filter: (name = '1m'::text)
Planning time: 10.428 ms
Execution time: 2336.828 ms
```

The slowest part of the query comes from us needing to fetch all the builds for that job. I chose a job that had a lot of builds, which is why you can see the time it took to run was very long.

I think what I'm going to do is implement the thing we discussed in the old PR. I'm going to first check the version against the previous build and if it exists then we are finished there. But if it does not exist then I will run a reworked query 

```
EXPLAIN ANALYZE WITH builds_of_version AS ( 
	SELECT build_id
	FROM build_resource_config_version_inputs
	WHERE name = $1
	AND version_md5 = $2
	AND resource_id = $3
) SELECT EXISTS ( 
	SELECT 1 
	FROM builds b
	JOIN builds_of_version v ON b.id = v.build_id
	WHERE job_id = $4
);
```

Where the difference is basically it narrows down the rows of builds by first filtering on the versions and then checking to see if that build exists for that job. The problem with this query is that if you have a version that is used by a ton of builds, it will end up having to grab all those builds that use the query and that can be just as slow as the previous query. That is why if we add the logic of first checking to see if the version is the same as the previous build, that might avoid us needing to do this long query. Also, I mentioned in the PR that if I add this new index `create index someindex on build_resource_config_version_inputs (resource_id, version_md5, name, build_id);` then it will help speed up that slow query.

analytically:
The original query fetched all builds for a job before filtering by version, causing 2+ second queries for jobs with 35k+ builds.

Add fast-path that checks the last successful build first. This handles the common case where a version is reused across builds, avoiding the expensive fallback query entirely.

Change fallback to use nested EXISTS instead of CTE. This filters by version first then checks job membership, and allows Postgres to stop at the first match.

Fixes #8160 but without index. TBD later on if we'd like to add an index.

### DIFF
--- a/atc/db/versions_db.go
+++ b/atc/db/versions_db.go
@@ -34,20 +34,52 @@ func NewVersionsDB(conn DbConn, limitRows int, cache *gocache.Cache) VersionsDB 
 	}
 }
 
-func (versions VersionsDB) IsFirstOccurrence(ctx context.Context, jobID int, inputName string, versionSHA256 ResourceVersion, resourceId int) (bool, error) {
-	var exists bool
+func (versions VersionsDB) IsFirstOccurrence(ctx context.Context, jobID int, inputName string, versionDigest ResourceVersion, resourceID int) (bool, error) {
+	var lastBuildID sql.NullInt64
 	err := versions.conn.QueryRowContext(ctx, `
-		WITH builds_of_job AS (
-			SELECT id FROM builds WHERE job_id = $1
-		)
+		SELECT id FROM builds
+		WHERE job_id = $1 AND status = 'succeeded'
+		ORDER BY id DESC
+		LIMIT 1`, jobID).Scan(&lastBuildID)
+	if err != nil && err != sql.ErrNoRows {
+		return false, err
+	}
+
+	if lastBuildID.Valid {
+		var exists bool
+		err := versions.conn.QueryRowContext(ctx, `
+			SELECT EXISTS (
+				SELECT 1
+				FROM build_resource_config_version_inputs
+				WHERE build_id = $1
+				AND name = $2
+				AND version_digest = $3
+				AND resource_id = $4
+			)`, lastBuildID.Int64, inputName, versionDigest, resourceID).
+			Scan(&exists)
+		if err != nil {
+			return false, err
+		}
+		if exists {
+			return false, nil
+		}
+	}
+
+	// Fallback: nested EXISTS, filters by version first then checks job.
+	var exists bool
+	err = versions.conn.QueryRowContext(ctx, `
 		SELECT EXISTS (
 			SELECT 1
 			FROM build_resource_config_version_inputs i
-			JOIN builds_of_job b ON b.id = i.build_id
-			WHERE i.name = $2
-			AND i.version_digest = $3
-			AND i.resource_id = $4
-		)`, jobID, inputName, versionSHA256, resourceId).
+			WHERE i.resource_id = $1
+			AND i.version_digest = $2
+			AND i.name = $3
+			AND EXISTS (
+				SELECT 1 FROM builds b
+				WHERE b.id = i.build_id
+				AND b.job_id = $4
+			)
+		)`, resourceID, versionDigest, inputName, jobID).
 		Scan(&exists)
 	if err != nil {
 		return false, err
--- a/atc/db/versions_db.go
+++ b/atc/db/versions_db.go
@@ -35,10 +35,11 @@ func NewVersionsDB(conn DbConn, limitRows int, cache *gocache.Cache) VersionsDB 
 }
 
 func (versions VersionsDB) IsFirstOccurrence(ctx context.Context, jobID int, inputName string, versionDigest ResourceVersion, resourceID int) (bool, error) {
+	// Shortcut: check if previous non-pending build used the same resource version
 	var lastBuildID sql.NullInt64
 	err := versions.conn.QueryRowContext(ctx, `
 		SELECT id FROM builds
-		WHERE job_id = $1 AND status = 'succeeded'
+		WHERE job_id = $1 AND status <> 'pending'
 		ORDER BY id DESC
 		LIMIT 1`, jobID).Scan(&lastBuildID)
 	if err != nil && err != sql.ErrNoRows {


atc/db/versions_db.go
```diff
@@ -34,20 +34,52 @@ func NewVersionsDB(conn DbConn, limitRows int, cache *gocache.Cache) VersionsDB
...
-func (versions VersionsDB) IsFirstOccurrence(ctx context.Context, jobID int, inputName string, versionSHA256 ResourceVersion, resourceId int) (bool, error) {
-	var exists bool
+func (versions VersionsDB) IsFirstOccurrence(ctx context.Context, jobID int, inputName string, versionDigest ResourceVersion, resourceID int) (bool, error) {
+	var lastBuildID sql.NullInt64
 	err := versions.conn.QueryRowContext(ctx, `
-		WITH builds_of_job AS (
-			SELECT id FROM builds WHERE job_id = $1
-		)
+		SELECT id FROM builds
+		WHERE job_id = $1 AND status = 'succeeded'
```
taylorsilva: I'm guessing you filtered by `status` because we need to avoid looking up the current job we're trying to schedule. The problem with this is that we want to know if this is the first occurrence of this resource for any build of the job, not just successful builds. We also need to consider any in-flight builds that are already running. I think flipping this to ignore any pending builds should satisfy that.


```suggestion
		WHERE job_id = $1 AND status <> 'pending'
```

ref: https://github.com/concourse/concourse/blob/8a921e9a3eb08598ca9457b11872a0596c42cdd3/atc/db/build.go#L59-L66

taylorsilva (changes requested):
This looks good to me. Dig a bunch of digging into the issue and previous PR made (#6403). Also ran the queries against CI with a job that has over a million builds.

I was mostly interested in the fallback query and how it compared to the query it's replacing.

Original query:
```sql
explain analyze
WITH builds_of_job AS (
	SELECT id FROM builds WHERE job_id = 490
	)
SELECT EXISTS (
	SELECT 1
	FROM build_resource_config_version_inputs i
	JOIN builds_of_job b ON b.id = i.build_id
	WHERE i.name = 'every-30s'
	AND i.version_digest = '7f8b0bcb41db44b2efda116d3b3740a894df412a73c212e4c6c70bb3094e36c9'
);
```

```
Result  (cost=581.43..581.44 rows=1 width=1) (actual time=3.457..3.459 rows=1.00 loops=1)
  Buffers: shared hit=1544
  InitPlan 1
    ->  Nested Loop  (cost=1.12..4643.61 rows=8 width=0) (actual time=3.453..3.455 rows=1.00 loops=1)
          Buffers: shared hit=1544
          ->  Index Scan using build_inputs_resource_versions_idx on build_resource_config_version_inputs i  (cost=0.56..4424.46 rows=47 width=8) (actual time=3.423..3.424 rows=1.00 loops=1)
                Index Cond: (version_digest = '7f8b0bcb41db44b2efda116d3b3740a894df412a73c212e4c6c70bb3094e36c9'::text)
                Filter: (name = 'every-30s'::text)
                Index Searches: 332
                Buffers: shared hit=1538
          ->  Index Only Scan using builds_ordered_by_job_id_idx on builds  (cost=0.56..4.66 rows=1 width=8) (actual time=0.022..0.022 rows=1.00 loops=1)
                Index Cond: ((job_id = 490) AND (id = i.build_id))
                Heap Fetches: 1
                Index Searches: 1
                Buffers: shared hit=6
Planning:
  Buffers: shared hit=14
Planning Time: 0.639 ms
Execution Time: 3.512 ms
```

PR query
```sql
explain analyze SELECT EXISTS (
SELECT 1
FROM build_resource_config_version_inputs i
WHERE i.resource_id = 874
AND i.version_digest = '7f8b0bcb41db44b2efda116d3b3740a894df412a73c212e4c6c70bb3094e36c9'
AND i.name = 'every-30s'
AND EXISTS (
	SELECT 1 FROM builds b
	WHERE b.id = i.build_id
	AND b.job_id = 490
)
);
```

```
Result  (cost=65.01..65.02 rows=1 width=1) (actual time=0.050..0.051 rows=1.00 loops=1)
  Buffers: shared hit=11
  InitPlan 1
    ->  Nested Loop  (cost=1.12..65.01 rows=1 width=0) (actual time=0.048..0.049 rows=1.00 loops=1)
          Buffers: shared hit=11
          ->  Index Scan using build_inputs_resource_versions_idx on build_resource_config_version_inputs i  (cost=0.56..38.12 rows=5 width=8) (actual time=0.031..0.032 rows=1.00 loops=1)
                Index Cond: ((resource_id = 874) AND (version_digest = '7f8b0bcb41db44b2efda116d3b3740a894df412a73c212e4c6c70bb3094e36c9'::text))
                Filter: (name = 'every-30s'::text)
                Index Searches: 1
                Buffers: shared hit=5
          ->  Index Only Scan using builds_ordered_by_job_id_idx on builds b  (cost=0.56..5.38 rows=1 width=8) (actual time=0.014..0.014 rows=1.00 loops=1)
                Index Cond: ((job_id = 490) AND (id = i.build_id))
                Heap Fetches: 1
                Index Searches: 1
                Buffers: shared hit=6
Planning:
  Buffers: shared hit=14
Planning Time: 0.789 ms
Execution Time: 0.101 ms
```

It's a great improvement!

I checked the two new queries that check if the previous build is the first occurrence, and they both seem fast. Their query plans used indexes and ran in a few milliseconds.

atc/db/versions_db.go
taylorsilva: Wondering if we even should filter by status since the previous query did not.

Locally checked the triggering behaviour using this pipeline

```yaml
resources:
  - name: time
    type: time
    source:
      interval: 1s

jobs:
  - name: regular-trigger
    plan:
      - get: time
        trigger: true

  - name: every-trigger
    plan:
      - get: time
        trigger: true
        version: every
```

The `every` trigger worked as expected, and normal triggering did as well, only triggering on the latest version when multiple versions were emitted at once. I think we're good here!

taylorsilva (approved):

PR #9436, opened 2026-01-23
Handle json null values in the resource_config_versions table during the sha256 migration

### LINKED ISSUE #9432: Migration script error upgrading to `8.0.0`: `error: migration '1747084615_switch_md5_to_sha256.up.sql' failed and was rolled back: ERROR: column "version_sha256" of relation "resource_config_versions" contains null values (SQLSTATE 23502)`
### Summary

Greetings,

I recently attempted upgrading a Concourse instance from `7.14.3` to `8.0.0` and encountered the following error:
```bash
error: migration '1747084615_switch_md5_to_sha256.up.sql' failed and was rolled back: ERROR: column "version_sha256" of relation "resource_config_versions" contains null values (SQLSTATE 23502)
```

This Concourse instance is deployed inside of a Kubernetes `1.31` cluster and this error was reported in the `concourse-migration` container in one of the `concourse-web` pods when the upgrade was underway. There are unfortunately no other logs when this error happens - the `concourse-migration` container just keeps logging that same message each time it attempts to apply the migration.

Looking at [the migration script](https://github.com/concourse/concourse/blob/master/atc/db/migration/migrations/1747084615_switch_md5_to_sha256.up.sql), my guess is that the `ALTER TABLE resource_config_versions ALTER COLUMN version_sha256 SET NOT NULL` statement is failing because some row(s) in this table have a null value for this column after the `WITH json_string_cte ...` statement from earlier in the script runs and generates the data for that column. I'm not sure what data exists in these tables currently that are causing this to go awry. This Concourse instance has some history and was initially deployed with version `7.1.0` (I think) in case that matters.

Deployment environment:
- Kubernetes `1.31.11` (RKE2)
- Helm Chart `20.0.0`
- Concourse `7.14.3` -> attempting `8.0.0`

### Steps to Reproduce

1. Start with an existing `7.14.3` Concourse deployment that has existed for years and gone through numerous upgrades.
2. Attempt to upgrade to `8.0.0`.
3. Observe errors in `concourse-web` pod in the `concourse-migration` container.

### Expected Results

Upgrade to `8.0.0` completed successfully.

### Actual Results

New `8.0.0` web nodes fail to start due to migration script failing to succeed. 

### Web Node(s) configuration

Two web nodes. This is an air-gapped installation so I cannot easily provide these configuration values; if there are values that are relevant to the issues I will be happy to provide them.

### Worker(s) configuration

Four worker nodes. This is an air-gapped installation so I cannot easily provide these configuration values; if there are values that are relevant to the issues I will be happy to provide them.

### Concourse Version

7.14.3

### Browser (if applicable)

_No response_

### Did this use to work?

_No response_

taylorsilva:

closes #9432

No clue how these values got in there, but we've had multiple reports of users having versions with null json values in this table. This is a different type from SQL NULL. The `version` column has a `NOT NULL` constraint on it already.

The migraton was not computing these null json values. They got filtered out by the WHERE clause that's now been removed. This would then result in an error later on when we tried to apply a `NOT NULL` constraint on the new version_sha256 column because we'd have these rows that we filtered out and never computed values for.

Now any versions that aren't already json objects are converted to empty json objects, allowing us to properly compute them now.

## Notes to reviewer

Using a local instance of Concourse 7.14, I made my `resource_config_versions` table look like this. I manually inserted the second record with the `some_md5` value. The value in this column doesn't matter for reproducing the problem or verifying the fix. We don't modify the values in this column during the migration.

```
id|version            |version_md5                     |metadata|check_order|resource_config_scope_id|span_context|
--+-------------------+--------------------------------+--------+-----------+------------------------+------------+
 1|{"version": "mock"}|8b60d2cd04609ea4cf585ee063babf48|null    |          1|                       1|{}          |
 4|null               |some_md5                        |null    |          0|                       1|            |
```

I then ran this slightly stripped down version of the problem query from the migration:

```sql
WITH json_string_cte AS ( 
     SELECT 
         rcv.id, 
         COALESCE( 
             '{' || string_agg('"' || kv.key || '":"' || kv.value || '"', ',' ORDER BY kv.key) || '}', 
             '{}' 
         ) AS json_string 
     FROM resource_config_versions rcv 
     LEFT JOIN jsonb_each_text(rcv.version::jsonb) AS kv ON true 
     WHERE jsonb_typeof(rcv.version::jsonb) = 'object' 
     GROUP BY rcv.id
 ), 
 hashed_json_string_cte AS ( 
     SELECT 
         json_string_cte.id, 
         encode(digest(json_string_cte.json_string, 'sha256'), 'hex') AS new_version_sha256 
     FROM json_string_cte 
 ) 
 select * FROM hashed_json_string_cte;
```

Which only returned one row, when it should return two:

```
id|new_version_sha256                                              |
--+----------------------------------------------------------------+
 1|dc5503a03b7a16ff473d3628846a4f7d62fac2d7f69b16a5808cd3d9270474c3|
```
We've missed converting a record!

Running the updated query + function:

```sql
CREATE OR REPLACE FUNCTION jsonb_coalesce_empty(value jsonb)
RETURNS jsonb AS $$
  SELECT
      CASE WHEN jsonb_typeof($1) != 'object'
          THEN '{}'::jsonb
          ELSE $1
  END
$$ LANGUAGE sql;

WITH json_string_cte AS ( 
     SELECT 
         rcv.id, 
         COALESCE(
            '{' || string_agg('"' || kv.key || '":"' || kv.value || '"', ',' ORDER BY kv.key) || '}',
            '{}'
        ) AS json_string
     FROM resource_config_versions rcv
     LEFT JOIN jsonb_each_text(jsonb_coalesce_empty(rcv.version::jsonb)) AS kv ON true 
     GROUP BY rcv.id
 ), 
 hashed_json_string_cte AS ( 
     SELECT 
         json_string_cte.id, 
         json_string_cte.json_string, --added for verification
         encode(digest(json_string_cte.json_string, 'sha256'), 'hex') AS new_version_sha256 
     FROM json_string_cte 
 ) 
 select * FROM hashed_json_string_cte;
```

It now returns all the records:
```
id|json_string       |new_version_sha256                                              |
--+------------------+----------------------------------------------------------------+
 1|{"version":"mock"}|dc5503a03b7a16ff473d3628846a4f7d62fac2d7f69b16a5808cd3d9270474c3|
 4|{}                |44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a|
```

### DIFF
--- a/atc/db/migration/migrations/1747084615_switch_md5_to_sha256.down.sql
+++ b/atc/db/migration/migrations/1747084615_switch_md5_to_sha256.down.sql
@@ -8,8 +8,8 @@ WITH json_string_cte AS (
             '{}'
         ) AS json_string
     FROM resource_config_versions rcv
-    LEFT JOIN jsonb_each_text(rcv.version::jsonb) AS kv ON true
-    WHERE jsonb_typeof(rcv.version::jsonb) = 'object' AND rcv.version_md5 IS NULL
+    LEFT JOIN jsonb_each_text(jsonb_coalesce_empty(rcv.version::jsonb)) AS kv ON true
+    WHERE rcv.version_md5 IS NULL
     GROUP BY rcv.id, rcv.version_sha256
 ),
 hashed_json_string_cte AS (
@@ -96,3 +96,5 @@ RENAME TO resource_disabled_versions_resource_id_version_md5_uniq;
 -- resource_caches
 ALTER INDEX IF EXISTS resource_caches_resource_config_id_version_digest_params_hash_uniq
 RENAME to resource_caches_resource_config_id_version_md5_params_hash_uniq;
+
+DROP FUNCTION IF EXISTS jsonb_coalesce_empty(value jsonb);
--- a/atc/db/migration/migrations/1747084615_switch_md5_to_sha256.down.sql
+++ b/atc/db/migration/migrations/1747084615_switch_md5_to_sha256.down.sql
@@ -94,7 +94,7 @@ ALTER INDEX IF EXISTS resource_disabled_versions_resource_id_version_digest_uniq
 RENAME TO resource_disabled_versions_resource_id_version_md5_uniq;
 
 -- resource_caches
-ALTER INDEX IF EXISTS resource_caches_resource_config_id_version_digest_params_hash_uniq
+ALTER INDEX IF EXISTS resource_caches_rsc_config_id_version_digest_params_hash_uniq
 RENAME to resource_caches_resource_config_id_version_md5_params_hash_uniq;
 
 DROP FUNCTION IF EXISTS jsonb_coalesce_empty(value jsonb);
--- a/atc/db/migration/migrations/1747084615_switch_md5_to_sha256.up.sql
+++ b/atc/db/migration/migrations/1747084615_switch_md5_to_sha256.up.sql
@@ -49,6 +49,15 @@ RENAME TO resource_disabled_versions_resource_id_version_digest_uniq;
 ALTER INDEX IF EXISTS resource_caches_resource_config_id_version_md5_params_hash_uniq
 RENAME to resource_caches_rsc_config_id_version_digest_params_hash_uniq;
 
+CREATE OR REPLACE FUNCTION jsonb_coalesce_empty(value jsonb)
+RETURNS jsonb AS $$
+  SELECT
+      CASE WHEN jsonb_typeof($1) != 'object'
+          THEN '{}'::jsonb
+          ELSE $1
+  END
+$$ LANGUAGE sql;
+
 -- Convert all rows to their new sha256 values
 WITH json_string_cte AS (
     SELECT
@@ -59,8 +68,7 @@ WITH json_string_cte AS (
             '{}'
         ) AS json_string
     FROM resource_config_versions rcv
-    LEFT JOIN jsonb_each_text(rcv.version::jsonb) AS kv ON true
-    WHERE jsonb_typeof(rcv.version::jsonb) = 'object'
+    LEFT JOIN jsonb_each_text(jsonb_coalesce_empty(rcv.version::jsonb)) AS kv ON true
     GROUP BY rcv.id, rcv.version_sha256
 ),
 hashed_json_string_cte AS (


Kump3r (commented):
Not sure If I can test this any other way, than what you suggested (manually adding such objects to the table. One thing on my mind, do we add to the down migration: DROP FUNCTION IF EXISTS jsonb_coalesce_empty(value jsonb); so it is correctly reverted? Not sure if the function was replaced, so we need to replace it back. Wdyt?

taylorsilva:
Added the `DROP FUNCTION` to the down migration and also updated the MD5 calculation query as well since it would've had the same issue as the up currently does.

atc/db/migration/migrations/1747084615_switch_md5_to_sha256.down.sql
```diff
@@ -96,3 +96,5 @@ RENAME TO resource_disabled_versions_resource_id_version_md5_uniq;
 -- resource_caches
 ALTER INDEX IF EXISTS resource_caches_resource_config_id_version_digest_params_hash_uniq
 RENAME to resource_caches_resource_config_id_version_md5_params_hash_uniq;
```
IvanChalukov: In the migration up we create `resource_caches_rsc_config_id_version_digest_params_hash_uniq` instead of `resource_caches_resource_config_id_version_digest_params_hash_uniq`. I've missed that in previous PR.

```suggestion
ALTER INDEX IF EXISTS resource_caches_rsc_config_id_version_digest_params_hash_uniq
RENAME to resource_caches_resource_config_id_version_md5_params_hash_uniq;
```

IvanChalukov (changes requested):
I requested one additional change to align the up and down migrations for the index `resource_caches_resource_config_id_version_md5_params_hash_uniq `. I missed this in the previous review.

---

Everything else appears to work as expected.

I tested this by creating a sample pipeline (to reuse an existing `resource_config_scope_id`) and then inserting a `null` version into the `resource_config_versions` table.

For the UP migration:
```psql
INSERT INTO resource_config_versions (
    version,
    version_md5,
    metadata,
    check_order,
    resource_config_scope_id,
    span_context
)
SELECT
    'null'::jsonb,
    md5('null'),
    metadata,
    check_order,
    resource_config_scope_id,
    span_context
FROM resource_config_versions
WHERE id = 1;
```
then trigger the migration.

For the down migration, I ran a similar test using a sample pipeline:
```psql
INSERT INTO resource_config_versions (
    version,
    metadata,
    check_order,
    resource_config_scope_id,
    span_context,
    version_sha256
)
SELECT
    'null'::jsonb,
    metadata,
    check_order,
    resource_config_scope_id,
    span_context,
    encode(digest('null', 'sha256'), 'hex')
FROM resource_config_versions
WHERE id = 1;
```
---

The only potential concern I can think of is that the hash representation of null (MD5/SHA256 of `null`) differs from the representation of an empty object (`{}`). But I hope no one is actually using these null versions, and this should not result in unintended job triggering or similar side effects.

aliculPix4D:
> No clue how these values got in there

hunch... maybe when developing a new resource type and immediately testing it in Concourse (basically testing only the docker image invoking the check, get and put), even though it was not even nearly ready... 

Anyway, just guessing... 

As you mentioned, this is not possible anymore due to: https://github.com/concourse/concourse/pull/9293

taylorsilva:
Thanks Ivan! I'll make that additional change.

> The only potential concern I can think of is that the hash representation of null (MD5/SHA256 of null) differs from the representation of an empty object ({}). But I hope no one is actually using these null versions, and this should not result in unintended job triggering or similar side effects.

I thought about that too. I'm also hoping that no one is using these versions and like Aleksandar said, they're probably just from people developing their own resource types 🤞

PR #9438, opened 2026-01-24
Make container selection for fly intercept interactive and searchable

taylorsilva:

Using the wonderful TUI library from [Charm](https://github.com/charmbracelet/bubbletea), users can now:

* use arrow keys or j/k to select a container via an interactive list
* filter the list using any search term
* pagination for when the list is very long

Removed the "build id#" from the front of the container strings because it felt redundant and not useful information when trying to find a container in the list.

## Before
image

## After
image
image
image

### DIFF
--- a/fly/integration/hijack_test.go
+++ b/fly/integration/hijack_test.go
@@ -3,9 +3,9 @@ package integration_test
 import (
 	"fmt"
 	"net/http"
-	"os"
 	"os/exec"
 	"strings"
+	"time"
 
 	"github.com/concourse/concourse/atc"
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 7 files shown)


kcbimonte:
How does this look with tasks with attempts? I don't remember exactly, but doesn't fly intercept / hijack show all attempts for a failed job? Like job test, run 3, task unit attempt 1, 2, and 3.

fly/integration/hijack_test.go
```diff
@@ -395,14 +395,18 @@ var _ = Describe("Hijacking", func() {
...
-			Eventually(sess.Out).Should(gbytes.Say("1. resource: banana, type: check"))
-			Eventually(sess.Out).Should(gbytes.Say("2. build #2, step: some-input, type: get, attempt: 1.1.1"))
-			Eventually(sess.Out).Should(gbytes.Say("3. build #2, step: some-output, type: put, attempt: 1.1.2"))
-			Eventually(sess.Out).Should(gbytes.Say("4. build #2, step: some-output, type: task, attempt: 1"))
-			Eventually(sess.Out).Should(gbytes.Say("choose a container: "))
+			Eventually(sess.Out).Should(gbytes.Say("Select a container:"))
+			Eventually(sess.Out).Should(gbytes.Say("resource: banana, type: check"))
+			Eventually(sess.Out).Should(gbytes.Say("step: some-input, type: get, attempt: 1.1.1"))
+			Eventually(sess.Out).Should(gbytes.Say("step: some-output, type: put, attempt: 1.1.2"))
+			Eventually(sess.Out).Should(gbytes.Say("step: some-output, type: task, attempt: 1"))
```
taylorsilva, line 402: @kcbimonte It looks the same. I'm highlighting the updated tests here which include steps with multiple attempts.

kcbimonte: Cool. Attempts is at the end instead of where I thought it was. All good here

Issue #9274, opened 2025-09-01
Overlay volume layers leaking


Kump3r:
# Original issue - #8159 

This is reproduced with newer versions of Concourse, somewhat rarely, but nevertheless to the same extend. Volumes start piling up and the garbage collector never manages to clean the `/overlays` and `/dead` volumes.

## Summary
In my stress environment running 7.6.0, I have seem specifically 2 workers that have seen their disk usage climb until it gets full. 

![image](https://user-images.githubusercontent.com/19290703/157898710-cc6d1941-6bef-4bb7-ac17-346d279670f2.png)

The weird thing is that this only happened on 2 of my workers and the rest of them (10 others) have a very consistent disk usage. So when I dug into where all that disk space is coming from, it looks like it is being taken up mostly by the `concourse-work-dir/overlays` folder. The following image shows the output of `du -h --max-depth=1 .` and the disk usage limit is 50 GB.

![image](https://user-images.githubusercontent.com/19290703/157905452-6422cc23-9fd3-4268-8bac-0249a5ba0cd0.png)

I took a look into the `concourse-work-dir/overlays/` folder and there were approximately 14000 folders in it, while the `overlays/work/` folder had around 4000 and the `concourse-work-dir/volumes/live/` had around 400. The `concourse-work-dir/volumes/dead/` also had around 4000. 

Then I tried to dig into what all those extra folders were in the `overlays` directory and I found some that were created on February 18th (which was almost a month ago) and no longer had a `volumes` row for it in the database. 

This screenshot is the logs that I found for that specific volume handle that the folder in `overlays` was for 
![Screen Shot 2022-03-09 at 12 10 53 PM](https://user-images.githubusercontent.com/19290703/157903840-cadcadf0-c9c7-4f43-922b-df396fb71cf4.png)

It seems like it tried to destroy the volume but failed from a timeout to baggageclaim, so I'm suspecting that something was going wrong in baggageclaim where it ended up deleting the volume from `volumes/live/` but not from the other overlay layers such as `overlays/`.

Digging into the code a little, it looks like for overlay driver we delete a volume by first renaming the folder in `volumes/live/` to `volumes/dead/`

https://github.com/concourse/concourse/blob/dabb94d93947d061c3037658da46b3c5a1098ebb/worker/baggageclaim/volume/filesystem.go#L254-L257

And then we delete the folder in the upper dir `overlays/`, then work dir `overlays/work/` and lastly the dead volume `volumes/dead/`

[...]

PR #9451, opened 2026-02-07
Fix some ways that overlay volumes leak and add more logging

taylorsilva:

Related to #9274

* Pass a logger into the Filesystem component to capture errors that we ignore
* Fixed our wrapper of `cp/robocopy` to return detailed errors. Before we would just return the exit status which isn't helpful. Now we'll return the error printed to `stderr`.
  * The Windows implementation was doing what we wanted, but missed the fact the it needed to call `cmd.Output()` instead of `cmd.Run()` in order for stderr to be captured.
* Fixed a leak in `ImportStrategy.Materialize()`. We'd create the volume and then try to populate it with existing files. If that failed we would error and not clean up the volume which would get orphaned in the `init` directory. Now when it fails we destroy the volume.
* Removed the `cleanup()` function from Filesystem which could result in overlay volumes leaking
    * I haven't been able to verify this, but I think it may have been possible, particularly with the overlay driver, to create volumes and then not clean them up properly. The overlay drive is prone to this error because it does all it's work outside of the `volumes/[init/live/dead]` folders. It does all of its work in the `overlays/` folder and then mounts the final result in `init/live/dead`. When calling `cleanup()` all we do is remove the folder in `init`, but we don't give the overlay driver a chance to remove whatever it's done in `overlays/`. Therefore, we potentially leak volumes
* The Destroy() function in the Overlay driver now runs all deletion steps and logs any errors, only returning an error from the last deletion step. Worst-case now, data inside the volume will be removed and the folder/volume will still be present. At least disk space will be cleared up!
* I've added test coverage for the files I've changed. This part of the codebase hasn't changed in a long time. It's mostly stayed this way since it was created. We didn't write sufficient test coverage to begin with!


worker/baggageclaim/baggageclaimcmd/driver_linux.go
```diff
@@ -133,8 +133,6 @@ func supportsFilesystem(fs string) (bool, error) {
 			return true, nil
 		}
 	}
-
-	return false, nil
```
taylorsilva, line 137: This was unreachable and called out by the go lsp.

worker/baggageclaim/volume/filesystem.go
```diff
@@ -347,6 +366,5 @@ func (vol *deadVolume) Destroy() error {
 	if err != nil {
 		return err
 	}
-
-	return vol.cleanup()
+	return os.RemoveAll(vol.dir)
 }
```
taylorsilva, line 379: I've been thinking about this function. I've been debating if I should make the call to  `driver.DestroyVolume()` and log any errors. I want this function to always make a best effort to clean up as much as possible, regardless of the state that the underlying volume is in (partial or fully created state).

PR #9452, opened 2026-02-08
Use @lydell/elm to support building concourse on arm

akshaymankar:

* Use unofficial elm package so concourse can be built on arm computers. Following recommendation from  here: https://github.com/elm/compiler/pull/2287#issuecomment-2389450649

## Notes to reviewer

I'm packaging concourse in nix, this is the only change I need to do to get it to work.


akshaymankar:
The CLA thing wants me to sign using my work email even if the commit uses my private email. So I'm waiting for the support to help me sort this out.

taylorsilva:
You'll need to rebase on master and resolve the conflict before the PR checks can run.

---

I'm not keen to move the project off to a fork for such a large dependency of the project. I get that it's a fork in the lightest sense since it doesn't contain any new code, it's just re-publishing and adding new binaries (I read the linked comment and the related issues).

This is not an issue for Concourse's own build process though. We build the frontend on linux/amd64 and then use the resulting artifact later when we build the Go binaries for linux/amd64 and linux/arm64. That's why we haven't run into this issue.

This doesn't do anything for the project, it just makes your own build process easier at the expense of the project having to trust a fork/republish of a major dependency.

Can you explain why, as a project maintainer of Concourse, I should move the project to this fork of elm?

akshaymankar:
> You'll need to rebase on master and resolve the conflict before the PR checks can run.

Done. The yarn.lock changed more than I expected, I couldn't figure out how to avoid that. Things are only upgraded so hopefully its fine.

> Can you explain why, as a project maintainer of Concourse, I should move the project to this fork of elm?

I'm creating a nix flake for concourse. The WIP repo is here: https://codeberg.org/axeman/concourse-flake
Eventually I'd like to merge the derivation and the module into nixpkgs.  

I run all my infra with nixos, some people in my bubble do too. I'd like concourse to be as easy (or even easier) to deploy on nixos as it is with K8s, BOSH and ansible. I'm hoping you'd want concourse to be more accessible too and so help me with this :)

yarn.lock
taylorsilva, line 1: Outside of elm, why are the version constraints lowered for a bunch of other dependencies?

taylorsilva (commented):
Still mulling this over. Talked about this in the working group meeting yesterday. Folks had similar feelings as me.

Actually looked at the diff. Curious about what happened with the version constraints.

yarn.lock
akshaymankar: I can't explain it, but I'm not very familiar with yarn.

package.json
```diff
@@ -10,26 +10,27 @@
...
     "uglify-js": "^3.19.3",
-    "webpack": "^5.104.1",
+    "webpack": "^5.105.2",
     "webpack-cli": "6.0.1"
   },
   "resolutions": {
-    "less/request": "^2.86.0"
+    "less/request": "^2.86.0",
+    "elm": "npm:@lydell/elm@0.19.1-14"
   },
```
taylorsilva, line 33: I've changed how we pull in `@lydell/elm` to make it obvious that this is a workaround and that we ideally want to target the official elm package.

https://yarnpkg.com/configuration/manifest#resolutions

This basically overrides the `elm` dependency. In the future we can simply remove this override.

taylorsilva: I've locally verified that I can do yarn install/build on linux/arm64 after making this change.

taylorsilva (commented):
I decided to do some due diligence. I wanted to understand `lydell`'s relationship to the Elm community. After going through recent community posts in the Elm slack and forum, I feel fine with pointing to this fork. [Simon Lydell](https://github.com/lydell/) is a very active member of the Elm community and has many different Elm packages that folks rely on.

I'll feel less confident about this if this fork does not keep up to date with the official package. Not a big issue since the official package hasn't updated in two years now and it's unclear if Elm maintainer (Evan) will release an update anytime soon.

I will open an issue as a reminder that we should remove this override.

package.json
akshaymankar: This also works in the nix derivation :+1:

PR #9463, opened 2026-02-19
Implement `disable_reruns` property for jobs

### LINKED ISSUE #5292: disable rerun of a specific job
### challenge

it's currently not possible to avoid one to re-run old builds of a particular
job.

similar to how to we have [`disable_manual_trigger`] as a job configuration, we
could have `disable_reruns` (bool) which would then prevent people from manually
triggering re-runs of a particular job.

[`disable_manual_trigger`]: https://concourse-ci.org/jobs.html#job-disable-manual-trigger

in those cases where one does not want to have builds running with old versions
of resources that are not considered "latest", having a
`disable_manuaL_triggegr` would allow one to have a job that always "runs
forward" with regards to the resource versions.

---

I'm afraid I might be falling into the XY problem here - to make it more explicit, what
I'd be trying to solve is the latest sentence there: making "impossible" to have a
particular job running new builds with old versions of its resources.

ceco556:

closes #5292 

* [x] Create `disable_reruns` property
* [x] DB migration
* [x] Rerun disabled in the Web UI
* [x] Tests

disable_rerun_job_trigger

## Notes to reviewer

* Implemented `disable_reruns` using the same approach as `disable_manual_trigger`.

## Release Note

By using `disable_reruns`, you can prevent rerunning a job when the flag is set to `true`.


web/elm/src/Build/Header/Header.elm
```diff
@@ -234,7 +234,14 @@ tooltip model session =
                 )
 
         HoverState.Tooltip RerunBuildButton _ ->
-            Just (buttonTooltip "re-run with the same inputs")
+            Just
+                (buttonTooltip <|
+                    if model.disableRerunJobTrigger then
+                        "re-run disabled in job config"
```
taylorsilva: Should add tests for this. Tests for the original hover text are here: https://github.com/concourse/concourse/blob/0e3ce86dda1dcb9d0db5f80a88526466debf4816/web/elm/tests/BuildTests.elm#L2240-L2253

taylorsilva (changes requested):
Another awesome PR! Overall looks good, just some nit-picky things.

- I'd prefer the new job field be `disable_reruns` instead of `disable_rerun_job_trigger`. The latter feels more descriptive of the backend implementation and overly verbose. `disable_reruns` is also what was suggested in the issue.
- Tests for the UI, specifically the tooltip text and CSS changes to the button (verify that it's not clickable), should be added
  - I debated having a better visual indication that reruns are disabled. I'm sadly not a designer and didn't like any of the suggestions I could come up with. I think what you've done is sufficient.

ceco556:
I will add the tests and change the property to `disable_reruns`.
Regarding the better visual indication, i can try to change the color to red when disabled or just hide it?

ceco556:
I think i've addressed the comments.
Hopefully didn't miss something.

taylorsilva (approved):
LGTM - Will merge once PR checks are done

PR #9472, opened 2026-02-26
Enable hot-reloading of ATC HTTPS listener

### LINKED ISSUE #9468: Implement ATC TLS Hot Reload
### Summary

trap `SIGHUP` signals in ATC, and when received, if TLS is enabled, reload the TLS key and certificate with as little interruption to the HTTPS listener as possible. 

This should, as reasonably as possible, mimic the behavior of TSA authorized key hot reloading

### Context

Issue [#1105](https://github.com/concourse/concourse/issues/1105)
PR [#5652](https://github.com/concourse/concourse/pull/5652)

jghiloni:

closes #9468 

- Wraps `*github.com/tedsuo/ifrit/http_server.httpServer` for TLS calls with new `ifrit.Runner` -- `*github.com/concourse/concourse/atc/tls.ReloadableHTTPSListener` that can react to `syscall.SIGHUP` signals, reload the `*tls.Config`, and restart the https listener
- Adds tests for same
- Updates `atccmd.RunCommand` to use `tls.ReloadableHTTPSListener` instead of `http_server.NewTLSListener` and adds a method to reload the TLS config when the signal is received

## Notes to reviewer

Like the TSA hot-reloading, this doesn't make a ton of sense to use in the BOSH or K8s deployment model 

## Release Note

When running ATC in a self-managed way (like `systemd` or `docker`), sending a `SIGHUP` to the running process will now reload TLS configuration data for the web listener as well as the authorized keys for the TSA.

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -49,6 +49,7 @@ import (
 	"github.com/concourse/concourse/atc/scheduler"
 	"github.com/concourse/concourse/atc/scheduler/algorithm"
 	"github.com/concourse/concourse/atc/syslog"
+	atctls "github.com/concourse/concourse/atc/tls"
 	"github.com/concourse/concourse/atc/util"
 	"github.com/concourse/concourse/atc/worker"
 	"github.com/concourse/concourse/atc/wrappa"
@@ -1055,11 +1056,12 @@ func (cmd *RunCommand) constructAPIMembers(
 		if err != nil {
 			return nil, err
 		}
-		members = append(members, grouper.Member{Name: "web-tls", Runner: http_server.NewTLSServer(
-			cmd.tlsBindAddr(),
-			httpsHandler,
-			tlsConfig,
-		)})
+		members = append(members, grouper.Member{
+			Name: "web-tls",
+			Runner: atctls.NewReloadableListener(
+				cmd.tlsBindAddr(), httpsHandler, tlsConfig, cmd.reloadTLSEnv(logger, dbConn), logger,
+			),
+		})
 	}
 
 	return members, nil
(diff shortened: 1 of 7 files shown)


jghiloni:
Forgot to note: In addition to the unit tests that were provided, I manually tested by running the concourse binary with the `web` command, updating the contents of the files pointed to by the `CONCOURSE_TLS_CERT` and `CONCOURSE_TLS_KEY` variables, then sending `SIGHUP` to the pid of the binary. I then confirmed that the PID had not changed, and that the cert offered by the https endpoint had indeed changed.

atc/atccmd/command.go
```diff
@@ -2195,6 +2197,16 @@ func (cmd *RunCommand) isMTLSEnabled() bool {
...
 }
 
+func (cmd *RunCommand) reloadTLSEnv(logger lager.Logger, dbConn db.DbConn) atctls.ConfigReloader {
+	return func() (*tls.Config, error) {
+		if cmd.isTLSEnabled() && !cmd.LetsEncrypt.Enable {
+			return cmd.tlsConfig(logger, dbConn)
+		}
+
+		return nil, nil
+	}
```
taylorsilva: I think the if-check here is redundant. `cmd.tlsConfig()` makes similar checks to both configs, the difference being it will return a default `tlsConfig`, and this function returns `nil`.

I think we can simplify this function by returning the result of `cmd.tlsConfig()`.


```suggestion
		return cmd.tlsConfig(logger, dbConn)
	}
```

atc/tls/runner.go
```diff
@@ -0,0 +1,120 @@
...
+	r.activeMu.Lock()
+	defer r.activeMu.Unlock()
+	if r.active != nil {
+		r.active.Signal(os.Interrupt)
+		err = <-r.active.Wait()
+	}
+
+	r.active = nil
+	return err
+}
```
taylorsilva: Seems like we don't use or need the returned error by `Stop()`. Maybe just ignore it?

```diff
-- err = <-r.active.Wait()
++ <-r.active.Wait()
```

taylorsilva (changes requested):
Tested locally and it works as expected! Just some small nit-pick stuff and then this is good to merge.

taylorsilva (approved):
TY!

PR #9477, opened 2026-03-05
Skip Lookup for batch tokens

Kump3r:

* Skip renewal for batch tokens, but instead of look-up determine the token type by inspecting the value of the token

## Notes to reviewer

Currently if token type is batch we do a look-up that fails and we determine the token is not renewable and skip the renew call errors, but cause look-up errors in our case **~620K**, this aims to avoid that by inspecting the token itself, rather than causing server-side issues.
The method I used to check the token is defined in the server side go code of vault, but unfortunately are NOT available in the Vault API client library.

- [ ] - To Be Tested still

### DIFF
--- a/atc/creds/vault/api_client.go
+++ b/atc/creds/vault/api_client.go
@@ -104,6 +104,12 @@ func (ac *APIClient) loginParams() map[string]any {
 	return loginParams
 }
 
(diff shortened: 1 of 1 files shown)


Kump3r:
batch tokens could be looked-up, there is another issue I assume. Need to dive deeper into the logic here.

PR #9483, opened 2026-03-13
[chore] regenerate counterfeiter fakes

alicul:
[chore] regenerate counterfeiter fakes

- counterfeiter v6.12.1 has an updated code-generation template compared to the version that was previously used to generate those fakes.
- the new version of counterfeiter removed all the individual read-locks (e.g., fake.claimsMutex.RLock()) from the .Invocations() method.


taylorsilva (approved):
Thanks for the PR! I appreciate the help with these low hanging tasks.

I locally ran `go generate` just to be sure nothing was being snuck in.

PR #9484, opened 2026-03-15
Make the Broadcast Message feature UI operable

ceco556:

closes #9478 

* [x] Added a new icon - bullhorn.
* [x] Added a button(bullhorn icon) which opens an editor bar where admin can type and submit a broadcast message
* [x] Buttons in the editor bar to set and clear the wall.
* [x] Bullhorn button is only visible for admins. It is not visible for users.
* [x] Tests

broadcast-message-edit-ui

> The button is visible if the user is admin.

not-visible-if-not-admin

> The button is not visible if the user is not admin.

## Notes to reviewer

- The wall banner is hidden while the editor is open to avoid overlap.
- The `WallSet (Ok ())` closes the editor as a safety measure, even though the save button closes it immediately.
- The bullhorn button closes the editor panel even though there is `x` button. I think it will be good to keep the `x` button, but i am ok with removing it if needed.
- I think i've covered all test cases.

## Release Note

Using this feature, concourse admins now can edit the broadcast message in the UI.


taylorsilva (changes requested):
Verified that the icon only appears for admin users:

Non-admin user:
Image

Admin user:
Image

---

I noticed, after setting a message, when pressing the "clear" button to remove the message, the edit text stays in place. I think when the user presses the clear button, the visual result in the UI should be the same as pressing the close "X" button.

https://github.com/user-attachments/assets/e869eef2-9cf9-4238-960f-3e2ce25cfe43

---

When pressing the broadcast icon, the textfield should get immediate text focus. Currently, after pressing the broadcast button, the user has to click the text field before they can start typing their message.

---

When hovering over the broadcast icon, there should be a hover text stating something like "Broadcast a message for all users to see".

---

When typing in the text field, if a user presses "ENTER" it puts them on a new line, which on this single-line text field, looks like the message got deleted. The text field should either expand to show the full message or enforce everything being on one line. I think I'd prefer the latter, the entire message being on one line.

https://github.com/user-attachments/assets/1ee31b41-fbc7-4a8d-b6de-726f6d30280e

---

That's my feedback. Overall this looks great and will be very handy to have! TY!

ceco556:
Thank you for the review!
I will work on this next days to address them.

Regarding the clear button, it was intended to work this way. I was thinking about this during the implementation and asked my self - what if someone clicks it by mistake? We don't have a confirmation for this?. So i decided to keep the message in the container and this way if someone clicks it by mistake, he can set it again. I am okay with removing it, but should we then have a confirmation for clearing the wall?

Thank you for the videos! It is very helpful for me to see what should i fix!

taylorsilva:
> Regarding the clear button, it was intended to work this way. I was thinking about this during the implementation and asked my self - what if someone clicks it by mistake? We don't have a confirmation for this?. So i decided to keep the message in the container and this way if someone clicks it by mistake, he can set it again

Very considerate of you! I think it's fine to clear the message and close the box though. The information someone would put in there is not critical and I doubt it would be very long either. To me at least, pressing clear doesn't "feel" like it does anything in its current state.

ceco556:
I've addressed the comments.

One thing came to my mind during fixing the issue with ENTER creating a new line. Now it doesn't, but should i change it to set the message directly when pressed?

Thanks again for the videos, it was very helpful!

ceco556:
The unit test job is failling and it is seems to be an issue with one of the tests in the DB Suite, but it is unrelated to this changes. Is there a global issue?
@taylorsilva, am i missing something?

taylorsilva:
Yup, it's unrelated, see https://github.com/concourse/concourse/pull/9492#discussion_r2983742548

I'll take a look into it. I haven't been able to reproduce it locally. Only happens in CI. Hopefully I'll figure it out with this second look.

taylorsilva:
I figured it out! Will push a fix and then re-trigger the PR check for you.

taylorsilva (approved):
Manually tested everything again and LGTM! TY!

PR #9485, opened 2026-03-15
Max timeout global configuration for tasks

ceco556:

closes #9480 

* [x] Introduce a new configuration option - `CONCOURSE_MAX_TASK_TIMEOUT`
* [x] Validation when user sets a pipeline
* [x] Tests

max_timeout

## Notes to reviewer

- Followed the same code pattern written for `CONCOURSE_DEFAULT_TASK_TIMEOUT`.
- If set to zero, it will be disabled.

## Release Note

Operators now can configure a maximum allowed timeout for tasks. Any pipeline that sets a timeout which exceeds the maximum allowed will be rejected.

### DIFF
--- a/atc/api/api_suite_test.go
+++ b/atc/api/api_suite_test.go
@@ -223,6 +223,7 @@ var _ = BeforeEach(func() {
 		credsManagers,
 		interceptTimeoutFactory,
 		time.Second,
+		time.Duration(0),
 		dbWall,
 		fakeClock,
 		dbSigningKeyFactory,
--- a/atc/api/configserver/save.go
+++ b/atc/api/configserver/save.go
@@ -61,7 +61,7 @@ func (s *Server) SaveConfig(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	warnings, errorMessages := configvalidate.Validate(config)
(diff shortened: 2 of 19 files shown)


taylorsilva:
After reading @alicul's comment reminding me about OPA, I'm currently not in favour of adding this as a feature. I'm definitely a lot more lenient when it comes to adding [more knobs](https://github.com/concourse/concourse/wiki/Anti-Patterns#knobs), but I'd still like to avoid adding new stuff if an existing feature can do this already.

As Aleksandar showed, OPA can easily do this. The hard part is setting up OPA, no denying that. But once you have it setup, the number of rules you can enforce is endless. It's not reasonable for Concourse to try and ingest every potential rule that that an operator may want to enforce; we'll never satisfy everyone.

I don't think we'll proceed with merging this PR in because:
- OPA integration can do this for operators already
- It goes against the [knob anti-pattern](https://github.com/concourse/concourse/wiki/Anti-Patterns#knobs)

Sorry for not thinking about this more deeply when you opened the thread. I should have flagged these concerns then. Being a little lenient to myself, I was on vacation with my partner at the time and was mostly trying to make sure I responded to any outstanding issues/threads 😅

PR #9488, opened 2026-03-16
Add --team flag to fly pipelines command

izabelacg:

Part of #5215

- Adds `--team` flag to `fly pipelines` to list pipelines for a team other than the target default
- Rejects --all and --team when used together


taylorsilva (approved):
Thanks Izabela! Nice to see a PR from you :)

PR #9492, opened 2026-03-18
Smarter resource checking - Only check triggering resources

### LINKED ISSUE #1445: Resource that don't trigger jobs should not check.
# Feature Request

We have a job that takes in 33 resources, but we always want to manually trigger the job. The resource checking coming from all of these resource is a massive drain on our workers, and we get no benefit out of the checking, as every time we trigger the job, it automatically pulls the newest version of each resource.

This often leads to an `insufficient subnets` error, as our workers will hit the max number of containers if more than one pipeline is un-paused. 

To give context, we have 6 jobs that similar to the one below. 

screen shot 2017-08-04 at 2 28 42 pm

## A Modest Proposal

Resources that do not trigger any jobs should have version checking turned off by default. This will save many unnecessary containers from being created.

taylorsilva:

closes #1445

The goal of this PR is to reduce the number of resource checks that currently happen in Concourse. I'm going to describe the current behaviour of Concourse and then explain what this PR changes.

### Current Behaviour

There are three main ways a build is started: manually triggered, automatic trigger, and rerun of previous build.

In parallel, there is also Lidar running every 10s, looking for any resources that have surpassed their `check_every` interval. Lidar performs this check for _every_ resource in unpaused pipelines, except resources that are only used in put steps.

**Manually triggered builds**: These happen when a user starts a new build via the web UI or `fly trigger-job`. In both of these cases, Concourse will create resource checks for all inputs of the job, overriding the check interval of all resources. The job will only start once all input resources have been checked.

**Automatically triggered builds**: These happen when a check started by Lidar finds a new version of an input to a job. In this case, no extra resource checks are made and the job is started with the latest set of input versions available from all resources of the job.

**Rerun builds**: These never create checks because they use the previous builds input versions. These are manually created.

With this setup, we end up with a system that is constantly running checks, more than it really needs to. The benefit of this is that when a job is automatically triggered, we don't need to wait for any other resource checks to run, we can just start the build. The con with this approach is that we waste a lot of compute checking resources and end up hammering external systems constantly.

### Behaviour Proposed by this PR

Lidar only creates checks for resources that are triggers for jobs, excluding triggers that also have passed constraints. This alone should drop the number of checks regularly occurring in a Concourse cluster.

The API endpoint used to manually trigger new builds no longer creates checks. That is now done by the BuildStarter.

We now move the creation of most resource checks into the `atc/scheduler/` package, specifically the BuildStarter.

Behaviour is the same for rerun builds, and mostly the same for manually triggered builds. Automatically triggered builds are slightly different.

**Manually triggered builds**: The check interval of all input resources is overridden, same as previous behaviour when checks were made in the `atc/api/` layer. The build is started once all input resources have been checked. The big change here is which resources the job waits to check; only resources that are inputs with no passed constraints.

**Automatically triggered builds**: These builds start after all non-triggering input resources have been checked, if they have passed their check interval. We don't check triggering resources because Lidar will check them, which is the same behaviour as before.

In all cases, when the `BuildStarter` creates checks, it won't create checks for resources that are pinned to a specific version, and any that have passed constraints.

With checks not constantly running for as many resources, the baseline amount of compute required for a Concourse cluster should be lower on average. External systems will also be hammered much less, though they will see spikes whenever a job starts and all it's non-triggering resources are checked. The con with this approach is that jobs will take longer to start if they need to wait for any resources to be checked.

#### Migration

There is a migration in this PR that adds a column to help us track when the next check of a resource should occur. The default value this column has for all resources after the migration is applied is Unix epoch. I don't _think_ this should have any adverse affects. Lidar will only check `trigger: true` jobs, and whenever a job is scheduled, then all of its inputs will be checked. I don't expect there to be a "stampeding herd" of checks being run after the migration is applied.


taylorsilva:
I wanted to track three things to evaluate how this change in scheduling of jobs and checks works. I started a local instance of Concourse running these three pipelines (same one's I used in the issue):

- [p-01.yml](https://github.com/user-attachments/files/26195256/p-01.yml)
- [p-02.yml](https://github.com/user-attachments/files/26195257/p-02.yml)
- [p-03.yml](https://github.com/user-attachments/files/26195258/p-03.yml)
- [p-04.yml](https://github.com/user-attachments/files/26195259/p-04.yml)
- [p-05.yml](https://github.com/user-attachments/files/26195260/p-05.yml)

The three metrics I want to track over a ~1hr period are:
1. Rate of checks being scheduled
2. Total checks run
3. Total builds for jobs run

My hope is that 1 and 2 go down (less checks being scheduled) and 3 stays the same. Initialization of jobs should slow down a bit since we have to wait for checks to run now (not in all cases), but overall I'd expect Concourse to be able to get through roughly the same number of builds before and after this PR.

So I set the pipelines and let them run for 1hr. They all use `time` resources with different `source.intervals`, and different triggering and passed constraints setup. I used the default check interval of one minute for all of them.

## Current Behaviour

image

- Rate: ~0.7 checks per second
- Total Checks run: 2617
- Total Job Builds Ran: 441


## PR Behaviour

image

- Rate: ~0.24 checks per second (lower is better, means less periodic checks occurring)
- Total Checks run: 824 (less is better)
- Total Job Builds Ran: 442 (similar is good, no change in scheduling rate)

## Summary

It appears this PR achieves the desired goal. The rate and total number of checks run dropped by 68%, but the number of builds run over the 1hr period was the same at ~442 builds successfully run.

I'm still cautiously making this change, so I'm planning to run it on our CI env first as a dev release after merging.

atc/db/check_factory_test.go
```diff
@@ -491,33 +500,35 @@ var _ = Describe("CheckFactory", func() {
...
+						builder.WithResourceVersions("some-resource-trigger", time.Minute),
+						builder.WithFailingResourceCheck("some-put-only-resource", time.Minute),
+					)
 				})
-				It("returns the resource", func() {
+
+				It("returns the triggering, and errored put-only resource", func() {
 					Expect(resources).To(HaveLen(2))
+					Expect(resources[0].Name()).To(Equal("some-resource-trigger"))
+					Expect(resources[1].Name()).To(Equal("some-put-only-resource"))
```
taylorsilva: For some reason this test is flaky in CI. I haven't gotten it to fail locally (`ginkgo --until-it-fails`) and even `fly intercept`'ing and re-running the test has always resulted in it passing.

My gut says this is related to CPU/Memory pressure causing the test to flake, but I'm not 100% sure about that. We also set `--flake-attempts` to 3 and this test fails all three times, which is VERY surprising to me. This makes me think the setup of the test is bad, but I can't see where it could be messing up.

The error I keep seeing is that `resources` has a length of three instead of two. I'd expect this to happen if the `BeforeEach()` didn't run. Maybe ginkgo is messing up the parallelization some how??

I'm going to mark the test as `Serial`; maybe we're somehow conflicting with other tests?? I don't see how that would happen, but having it flake like this is very annoying. If setting it to `Serial` makes it pass consistently then I'll do it.

taylorsilva:
I updated ci.concourse-ci.org to see if anything breaks from this PR.

So far everything seems normal. Upgrade was done at 11:08am UTC-4.

PR #9390, opened 2025-12-09
worker: add fallback to ATC streaming when P2P fails

kurtmc:
## Changes proposed by this PR

Related to #7017

This PR implements an improvement for volume streaming by adding automatic fallback to ATC (web node) streaming when P2P streaming fails.

### Key changes:
* When P2P streaming fails (e.g., workers on different networks, connectivity issues), automatically fall back to streaming through the ATC web node
* Added a log for when the fallback occurs
* Add new metric `volumes_streamed_via_fallback` to track fallback usage for observability

### Rationale:
I want to be able to run a small number of nodes in a different network and still have the majority of the nodes run in a seperate network and use p2p for peformance reasons.

## Notes to reviewer

* The fallback preserves the same error handling contract - if both P2P and ATC streaming fail, the error from ATC streaming is returned

## Release Note

Volume streaming now automatically falls back to streaming through the web node when peer-to-peer streaming fails. This makes the `enable-p2p-volume-streaming` flag more resilient in environments where workers may be on different networks.

PR #9503, opened 2026-03-24
Fix missing volumesStreamedViaFallback Prometheus metric

alicul:
The volumesStreamedViaFallback metric was being tracked internally but not exposed via Prometheus. This metric is incremented when volume streaming falls back to streaming through the ATC instead of direct worker-to-worker streaming. I think this was introduced in: https://github.com/concourse/concourse/pull/9390

The metric is exposed as:
concourse_volumes_volumes_streamed_via_fallback

I was preparing the (next) PR to extend a bit the streaming metrics like:
````
1. Counters
- `concourse_volumes_streaming_p2p_success_total` - Successful P2P streams
- `concourse_volumes_streaming_p2p_failure_total` - Failed P2P streams
- `concourse_volumes_streaming_atc_success_total` - Successful ATC streams
- `concourse_volumes_streaming_atc_failure_total` - Failed ATC streams

2. Histograms
- `concourse_volumes_streaming_duration_seconds{method,status}` - Stream duration by method and status
- `concourse_volumes_streaming_size_bytes` - Volume size distribution

3. Gauges
- `concourse_volumes_streaming_in_progress{method}` - Currently active streams
`````
when I noticed the issue with pre-existing: volumesStreamedViaFallback.

I did some local manual test and checked that metrics increase as expected by the test:
````
$ curl -s http://localhost:9090/metrics | grep -E "concourse_volumes_volumes_streamed|concourse_volumes_volumes_streamed_via_fallback" | grep -v "^#"
concourse_volumes_volumes_streamed 0
concourse_volumes_volumes_streamed_via_fallback 0

$ curl -s http://localhost:9090/metrics | grep -E "concourse_volumes_volumes_streamed|concourse_volumes_volumes_streamed_via_fallback" | grep -v "^#"
concourse_volumes_volumes_streamed 1
concourse_volumes_volumes_streamed_via_fallback 0

$ curl -s http://localhost:9090/metrics | grep -E "concourse_volumes_volumes_streamed|concourse_volumes_volumes_streamed_via_fallback" | grep -v "^#"
concourse_volumes_volumes_streamed 1
concourse_volumes_volumes_streamed_via_fallback 1
````
I know it doesn't show much, but if you wish, I can post the full steps (docker compose with multiple workers in different networks with P2P enabled, + test pipeline.)

### DIFF
--- a/atc/metric/emitter/prometheus.go
+++ b/atc/metric/emitter/prometheus.go
@@ -78,7 +78,8 @@ type PrometheusEmitter struct {
 
 	checksEnqueued prometheus.Counter
 
-	volumesStreamed prometheus.Counter
+	volumesStreamed            prometheus.Counter
+	volumesStreamedViaFallback prometheus.Counter
 
 	getStepCacheHits       prometheus.Counter
 	streamedResourceCaches prometheus.Counter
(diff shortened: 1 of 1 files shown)


taylorsilva (approved):
LGTM - I trust you when you say you verified this locally :)

PR #9509, opened 2026-04-02
worker/runtime: wrap containers from NewContainer and Containers with request timeout

Kump3r:
### Description
`NewContainer` and `Containers` were returning raw `containerd.Container` values,
bypassing the `requestTimeout` wrapper applied in `GetContainer`. This meant
downstream calls to `Spec`, `Task`, and `NewTask` during `container.Run` could block
indefinitely with a background context, even when --containerd-request-timeout
was configured.

This caused intermittent hangs where `gdn-init` was alive but the user process
was never exec'd - a userspace handshake deadlock that would only resolve when
the outer step timeout was reached, leaking a lidar check in the process.

closes:
#9398 
#9511

### What I changed?
* Updated `client.go` so `NewContainer` now returns the timeout-wrapped container type instead of the raw containerd container.
* Updated `client.go` so `Containers` also returns timeout-wrapped container objects for every listed container.

### Why this targets the hang?
* The stuck path is consistent with unbounded `containerd` calls during early process startup.
* Before this change, containers from create/list paths could bypass the wrapper and run calls like `Spec`, `Task`, and `NewTask` with effectively unbounded contexts.
* After this change, those calls consistently honor the configured worker --containerd-request-timeout, so deadlocked handshakes should fail fast instead of leaking long-running checks/tasks.

### Validation
* docker compose && run a pipeline

### Notes to reviewer
trade-off to be aware of: When unit tests use `FakeClient.NewContainerReturns(fakeContainer, nil)`, the backend gets back a raw `FakeContainer` without the timeout wrapper. So `container_test.go` and `backend_test.go` don't exercise the timeout wrapping. That's intentional for unit tests as the interfaces (`Client-FakeClient` and `Container-FakeContainer`) remain unchanged.

The wrapping is exercised in the integration tests under `integration_test.go` which uses the real `libcontainerd.New(...)` with a real `containerd` client that now returns wrapped containers from both `NewContainer` and `Containers`.

### Steps to reproduce
Still not actually 100% sure as it is quite intermittent, but I assume heavier load, IO on the machine might cause the leak 

### Possible verification:
As explained in the discussion on a problematic web, we see:
1. Attaching to the webs of an impacted instance
2. Checking the lidar metrics with curl http://localhost:9090 | grep concourse_lidar
3. Doing the math STARTED − (SUCCESS + ERROR) should eventauly equal 0
4. That was true for all web hosts, except for a single one which came out to 1

So I expect this works if eventually we see that the STARTED − (SUCCESS + ERROR) is always 0

### DIFF
--- a/worker/runtime/integration/integration_test.go
+++ b/worker/runtime/integration/integration_test.go
@@ -1223,3 +1223,80 @@ func (s *IntegrationSuite) TestNetworkMountsAreRemoved() {
 	s.NoError(err)
 	s.Len(networkFiles, 0)
 }
+
+// TestNewContainerEnforcesTimeoutOnTask is a regression test verifying that
+// containers returned by client.NewContainer enforce the requestTimeout on
+// follow-on operations (Task, NewTask, Spec).
+
+// How this test works:
+//  1. Create a libcontainerd client with a 1s requestTimeout
+//  2. Create a container directly via client.NewContainer with a minimal
+//     OCI runtime spec (no init binary or running task needed)
+//  3. Freeze containerd with SIGSTOP (making all gRPC calls hang)
+//  4. Call Task() on the container from NewContainer with context.Background()
+//     (no deadline from the caller — the wrapper must add it)
+//  5. Expect a "deadline exceeded" error within ~1s
+//  6. Also verify GetContainer returns a container that behaves identically
+//  7. Unfreeze containerd and verify recovery
+//
+// Step 4 will blocks indefinitely and the test times out without the fix
+func (s *IntegrationSuite) TestNewContainerEnforcesTimeoutOnTask() {
+	requestTimeout := 1 * time.Second
+
+	client := libcontainerd.New(
+		s.containerdSocket(),
+		"test-timeout-enforcement",
+		requestTimeout,
+	)
+	err := client.Init()
+	s.NoError(err)
+	defer client.Stop()
+
+	handle := uuid()
+	minimalSpec := &specs.Spec{
+		Version: specs.Version,
+		Process: &specs.Process{
+			Args: []string{"/bin/sh"},
+		},
+		Root: &specs.Root{
+			Path: s.rootfs,
+		},
+	}
+
+	contViaNew, err := client.NewContainer(
+		context.Background(), handle, map[string]string{}, minimalSpec,
+	)
+	s.NoError(err, "NewContainer should succeed")
+	defer func() {
+		_ = s.containerdProcess.Process.Signal(syscall.SIGCONT)
+		contViaNew.Delete(context.Background())
+	}()
+
+	_, err = contViaNew.Task(context.Background(), nil)
+	s.Error(err, "Task should return 'not found' error (no task started)")
+
+	contViaGet, err := client.GetContainer(context.Background(), handle)
+	s.NoError(err, "GetContainer should succeed while containerd is responsive")
+
+	_, err = contViaGet.Task(context.Background(), nil)
+	s.Error(err, "Task via GetContainer should also return 'not found'")
+
+	s.NoError(s.containerdProcess.Process.Signal(syscall.SIGSTOP))
+
+	start := time.Now()
+	_, err = contViaNew.Task(context.Background(), nil)
+	elapsed := time.Since(start)
+
+	s.Error(err, "Task must fail when containerd is frozen")
+	s.Contains(err.Error(), "deadline exceeded",
+		"Expected 'context deadline exceeded', got: %v", err)
+	s.Less(elapsed.Seconds(), 3.0,
+		"Task should timeout in ~1s (requestTimeout), but took %v", elapsed)
+
+	s.NoError(s.containerdProcess.Process.Signal(syscall.SIGCONT))
+
+	_, err = contViaNew.Task(context.Background(), nil)
+	s.Error(err)
+	s.NotContains(err.Error(), "deadline exceeded",
+		"After unfreeze, Task should not timeout")
+}
--- a/worker/runtime/integration/integration_test.go
+++ b/worker/runtime/integration/integration_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/concourse/concourse/worker/workercmd"
 	containerd "github.com/containerd/containerd/v2/client"
 	"github.com/jackpal/gateway"
+	"github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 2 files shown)


Kump3r:
## Steps to reproduce
1. Add `CONCOURSE_CONTAINERD_REQUEST_TIMEOUT: 10s` to worker `docker-compose.yml` section.
2. ~Add a `time.Sleep(15 * time.Second)` temporarily at the top of `container.Run()` in `container.go`~
2. Add a `time.Sleep(15 * time.Second)` to the top of `startTask` method in `backend.go`
3. Trigger a check so a container gets created and directly do a `CTRD_PID=$(pidof containerd) && kill -SIGSTOP $CTRD_PID` from the worker while sleep is running. Wait for the sleep to expire — Spec() is called against a frozen `containerd`.

Pre-fix: hangs indefinitely despite `--containerd-request-timeout=10s`, because Spec/Task are called with an unwrapped background context. Post-fix: times out after 10s with container `spec: context deadline exceeded`.

In production the timing was driven by the TTY handshake race under load rather than an artificial sleep — but the effect was identical.

## Commands used
1. Watch the containerd task list — wait until a check container appears
```
watch -n1 'ctr -n concourse tasks list'
```

2. Kill containerd from another `docker exec -it concourse-worker-1 sh`
```
CTRD_PID=$(pidof containerd) && kill -SIGSTOP $CTRD_PID
```
### **BEFORE FIX**: continuously ongoing check not timeouted after 10s., until eventualy we see `no workers error` as we killed the only one

### **AFTER FIX**:  Behaviour with the fix that was missing, was clearly seen in the UI with:
```
run check: start process: backend error: Exit status: 500, message: {"Type":"","Message":"container spec: context deadline exceeded","Handle":"","ProcessID":"","Binary":""}
```

It is pretty hard to reproduce without an actual load on the scheduler and system, but that is the closest I was able to get to the actual behaviour we observe in prod and would love your feedback. I know it is quite a small change but for us it would be actually really helpful in ensuring the robustness of `containerd` and having a smoother flow.

Kump3r:
Screenshot 2026-04-03 at 10 40 01


As you can see in the time above it is 10:40, while the check itself was triggered in 10:38:42, well over a minute ago so the actual 10s policy is not respected and defaults to whatever the upstream DEFAULT_CHECK_TIMEOUT, or check_timeout is passed, when in fact the containerd_request_timeout should be respected and fail.

IvanChalukov:
I was able to test it locally. We had a quick sync with @Kump3r on how to test it manually and we update steps a little bit:
1. Add `time.Sleep(15 * time.Second)` at the top of `startTas` method in `worker/runtime/backend.go`
2. Start concourse via docker compose(Set `CONCOURSE_CONTAINERD_REQUEST_TIMEOUT: 10s`) - `docker-compose up --build -d`
3. Set pipeline. Do not unpause it.
4. Connect to worker with `docker exec -it concourse-worker-1 bash`
5. Trigger check from UI and stop freeze containerd with `CTRD_PID=$(pidof containerd) && kill -SIGSTOP $CTRD_PID`
6. Await for result into UI. Without fix check will timeout eventually. With fix will respect set REQUEST_TIMEOUT to 10s
---

Additionally I think I came up with an integration test that would test behaviour in `worker/runtime/integration/integration_test.go`:
```go
// TestNewContainerEnforcesTimeoutOnTask is a regression test verifying that
// containers returned by client.NewContainer enforce the requestTimeout on
// follow-on operations (Task, NewTask, Spec).

// How this test works:
//  1. Create a libcontainerd client with a 1s requestTimeout
//  2. Create a container directly via client.NewContainer with a minimal
//     OCI runtime spec (no init binary or running task needed)
//  3. Freeze containerd with SIGSTOP (making all gRPC calls hang)
//  4. Call Task() on the container from NewContainer with context.Background()
//     (no deadline from the caller — the wrapper must add it)
//  5. Expect a "deadline exceeded" error within ~1s
//  6. Also verify GetContainer returns a container that behaves identically
//  7. Unfreeze containerd and verify recovery
//
// Step 4 will blocks indefinitely and the test times out without the fix
func (s *IntegrationSuite) TestNewContainerEnforcesTimeoutOnTask() {
	requestTimeout := 1 * time.Second

	client := libcontainerd.New(
		s.containerdSocket(),
		"test-timeout-enforcement",
		requestTimeout,
	)
	err := client.Init()
	s.NoError(err)
	defer client.Stop()

	handle := uuid()
	minimalSpec := &specs.Spec{
		Version: specs.Version,
		Process: &specs.Process{
			Args: []string{"/bin/sh"},
		},
		Root: &specs.Root{
			Path: s.rootfs,
		},
	}

	contViaNew, err := client.NewContainer(
		context.Background(), handle, map[string]string{}, minimalSpec,
	)
	s.NoError(err, "NewContainer should succeed")
	defer func() {
		_ = s.containerdProcess.Process.Signal(syscall.SIGCONT)
		contViaNew.Delete(context.Background())
	}()

	_, err = contViaNew.Task(context.Background(), nil)
	s.Error(err, "Task should return 'not found' error (no task started)")

	contViaGet, err := client.GetContainer(context.Background(), handle)
	s.NoError(err, "GetContainer should succeed while containerd is responsive")

	_, err = contViaGet.Task(context.Background(), nil)
	s.Error(err, "Task via GetContainer should also return 'not found'")

	s.NoError(s.containerdProcess.Process.Signal(syscall.SIGSTOP))

	start := time.Now()
	_, err = contViaNew.Task(context.Background(), nil)
	elapsed := time.Since(start)

	s.Error(err, "Task must fail when containerd is frozen")
	s.Contains(err.Error(), "deadline exceeded",
		"Expected 'context deadline exceeded', got: %v", err)
	s.Less(elapsed.Seconds(), 3.0,
		"Task should timeout in ~1s (requestTimeout), but took %v", elapsed)

	s.NoError(s.containerdProcess.Process.Signal(syscall.SIGCONT))

	_, err = contViaNew.Task(context.Background(), nil)
	s.Error(err)
	s.NotContains(err.Error(), "deadline exceeded",
		"After unfreeze, Task should not timeout")
}
```

IvanChalukov (approved):
Great job! I've tested it locally as I already mention it comment, as well as through integration test.

taylorsilva:
Just read over everything, this is great! Very simple fix, but it's all the research and debugging beforehand that shows the true effort of finding what feels like a "needle in a haystack" type of bug.

Thanks for reviewing and testing Ivan! Glad you were able to write a test for this too 🙏

PR #9521, opened 2026-04-10
Allow pending builds in paused pipelines to be aborted and automatically abort pending builds in archived pipelines

### LINKED ISSUE #9519: Abort builds for paused pipelines
### Summary

Allow users the ability to abort builds that are pending on paused pipelines

### Context

When a pipeline is paused, users are able to trigger a build. However, they are unable to abort the build. The workaround is to unpause the pipeline, then abort the build.

taylorsilva:

closes #9519

This PR fixes a bug and adds an enhancement.

* Enhancement: automatically abort pending builds in archived pipelines
  * I tested and it was possible for a user to manually abort the build
* Bug: pending builds can be aborted even when the pipeline is paused
  * This bug was introduced in v6.0.0, when a lot of changes were made to the scheduler and adjacent components. See the note for more details.

## Notes to reviewer

This bug was introduced in v6.0.0 when big changes were made to the
scheduler and related db components. It was made by a change to the
JobFactory query that fetches jobs that need scheduling. The query would
filter out jobs and pipelines that were paused, making it impossible for
us to abort any builds that were part of paused jobs or pipelines.

Changing this query alone fixes the bug without any weird scheduling
regressions. The rest of the changes are to ensure we don't consume any
more resources than we need to, ensuring performance doesn't degrade
since we'll definitely fetch more builds now.

First, I had to update db.Job to include info about its pipeline being
paused or not. That leverages the existing job query already used across
the codebase. That information is exposed via the PipelineIsPaused()
func.

With this info, the scheduler can now detect if the job or pipeline is
paused and take a shorter path, skipping all the algorithm calls.

We then rely on db.Job.ScheduleBuild() to see that the job or pipeline
is paused, and not schedule any pending builds. I was also able to
remove one query from db.Job.isPipelineOrJobPaused() since the info it
queried for was now part of db.Job.pipelinePaused.

Overall there should be very minimal impact on performance from this
change.

### DIFF
--- a/atc/db/job.go
+++ b/atc/db/job.go
@@ -83,6 +83,7 @@ type Job interface {
 	PausedAt() time.Time
 	// Returns true if the pipeline the job belongs to is paused
 	PipelineIsPaused() bool
+	PipelineIsArchived() bool
 	Pause(pausedBy string) error
 	Unpause() error
 
@@ -136,7 +137,8 @@ var jobsQuery = psql.Select(
 	"j.disable_reruns",
 	"j.paused_by",
 	"j.paused_at",
-	"p.paused").
+	"p.paused",
+	"p.archived").
 	From("jobs j").
 	LeftJoin("pipelines p ON j.pipeline_id = p.id").
 	LeftJoin("teams t ON p.team_id = t.id")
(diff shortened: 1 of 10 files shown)


atc/db/job.go
```diff
@@ -607,21 +616,16 @@ func (j *job) ScheduleBuild(build Build) (bool, error) {
 		return true, nil
 	}
 
-	tx, err := j.conn.Begin()
-	if err != nil {
-		return false, err
+	if j.paused || j.pipelinePaused {
+		return false, nil
```
taylorsilva, line 620: Don't need to check the archive status of a pipeline because an archived pipeline is also marked as paused.

PR #9523, opened 2026-04-12
feat(atc): natively inject BUILD_STATUS into resource metadata

### LINKED ISSUE #8541: Make the build state available to resources via metadata
When using a Concourse resource as notification mechanism (say to update the GitHub Commit status, or to send a chat message), the configuration is verbose and redundant, also if "otpimized" with YAML anchors. For example:

```yaml
cogito-handlers: &cogito-handlers
    on_success:
      put: cogito
      # redundant: I am telling the resource that the state is `success`, but the handler is `on_success`
      params: {state: success}
    on_failure:
      put: cogito
      # same redundancy for all the other handlers
      params: {state: failure}
    on_error:
      put: cogito
      params: {state: error}
    on_abort:
      put: cogito
      params: {state: abort}
```

Then this YAML anchor is referenced in the rest of the pipeline:

```yaml
jobs:
  - name: autocat
    <<: *cogito-handlers
    plan:
      - get: repo.git
    ...
```

When running the `put` step, the resource has already access to the following [resource metadata](https://concourse-ci.org/implementing-resource-types.html#resource-metadata):

- `$BUILD_ID`
- `$BUILD_NAME`
- `$BUILD_JOB_NAME`
- `$BUILD_PIPELINE_NAME`
- `$BUILD_PIPELINE_INSTANCE_VARS`
- `$BUILD_TEAM_NAME`
- `$BUILD_CREATED_BY`
- `$ATC_EXTERNAL_URL`

By looking at the metadata, I realized that the resource has enough information to query the ATC, get the build state associated to the `$BUILD_ID` and allow the configuration to be less verbose and more natural:

```yaml
cogito-handlers: &cogito-handlers
    on_success:
      put: cogito
    on_failure:
      put: cogito
    on_error:
      put: cogito
    on_abort:
      put: cogito
```

But querying the ATC adds another possibility for failure and especially requires to provide to the resource the local credentials of a Concourse user. Also if the user can be restricted to read-only access via the [viewer role](https://concourse-ci.org/user-roles.html#team-viewer-role), it is still giving the resource more permissions than it really needs.

Thus the proposal: if we added to the resource metadata the build state, say
```
- `$BUILD_STATE`
```
we could gain the reduced verbosity shown above without the complexity and risk of giving user credentials to the resource.

Regarding #6600, in my understanding it would require so many other changes (prototypes and notification itself) that I don't see it happening soon.

arakakimarcelo:
When using a resource within a hook step (e.g., `on_success`, `on_failure`, `on_error`, `on_abort`), the execution step is now context-aware of the build state sequence that triggered it. This modifies the builder tree to propagate the hook's execution state down to the resource environment variable natively, reducing configuration redundancy.

Resolves #8541

closes #8541

* Modified [atc/exec/step_metadata.go](cci:7://file:///Users/arakaki/projects/open-source/concourse/atc/exec/step_metadata.go:0:0-0:0) by adding a `BuildState` string that natively generates the `BUILD_STATE` environment variable.
* Refactored step initialization throughout [atc/engine/builder.go](cci:7://file:///Users/arakaki/projects/open-source/concourse/atc/engine/builder.go:0:0-0:0) to thread a contextual tracking of the overarching `BuildState`.
* Modified the specific `on_success`, `on_failure`, `on_error`, and `on_abort` builder wrappers to appropriately pass `"succeeded"`, `"failed"`, `"errored"`, and `"aborted"` identifiers downstream when parsing nested plans.
* Updated related [atc/engine/builder_test.go](cci:7://file:///Users/arakaki/projects/open-source/concourse/atc/engine/builder_test.go:0:0-0:0) and [atc/exec/step_metadata_test.go](cci:7://file:///Users/arakaki/projects/open-source/concourse/atc/exec/step_metadata_test.go:0:0-0:0) test suites with explicit assertions checking that `$BUILD_STATE` parses correctly within nested pipeline hooks.

## Notes to reviewer
The engine passes the state by mapping hook-specific constants directly (e.g., `db.BuildStatusSucceeded`) dynamically when establishing the hierarchical step tree layout inside `.Next` fields. This elegantly isolates the behavior down recursively regardless of how deep the `in_parallel`/[do](cci:1://file:///Users/arakaki/projects/open-source/concourse/atc/exec/put_step.go:31:1-31:19)/[try](cci:1://file:///Users/arakaki/projects/open-source/concourse/atc/engine/builder.go:264:0-275:1) blocks scale. 
All unit checks passed successfully validating this nested environment allocation. 

## Release Note
* Added contextual `$BUILD_STATE` metadata generation globally natively exposed when instantiating arbitrary hooks (simplifying parameter configurations for notification resources out-of-the-box).

### DIFF
--- a/atc/engine/builder.go
+++ b/atc/engine/builder.go
@@ -60,387 +60,352 @@ type stepperFactory struct {
 	lockFactory     lock.LockFactory
 }
 
+type planBuilder struct {
+	factory    *stepperFactory
+	build      db.Build
+	buildState string
+}
+
+func (pb *planBuilder) withBuildState(state string) *planBuilder {
+	return &planBuilder{factory: pb.factory, build: pb.build, buildState: state}
+}
+
 func (factory *stepperFactory) StepperForBuild(build db.Build) (exec.Stepper, error) {
 	if build.Schema() != supportedSchema {
 		return nil, errors.New("schema not supported")
 	}
 
+	pb := &planBuilder{factory: factory, build: build, buildState: build.Status().String()}
 	return func(plan atc.Plan) exec.Step {
-		return factory.buildStep(build, plan, string(build.Status()))
+		return pb.buildStep(plan)
 	}, nil
 }
 
-func (factory *stepperFactory) buildDelegateFactory(build db.Build, plan atc.Plan) DelegateFactory {
+func (pb *planBuilder) buildDelegateFactory(plan atc.Plan) DelegateFactory {
 	return DelegateFactory{
-		build:           build,
+		build:           pb.build,
 		plan:            plan,
-		rateLimiter:     factory.rateLimiter,
-		policyChecker:   factory.policyChecker,
-		dbWorkerFactory: factory.dbWorkerFactory,
-		lockFactory:     factory.lockFactory,
+		rateLimiter:     pb.factory.rateLimiter,
+		policyChecker:   pb.factory.policyChecker,
+		dbWorkerFactory: pb.factory.dbWorkerFactory,
+		lockFactory:     pb.factory.lockFactory,
 	}
 }
 
-func (factory *stepperFactory) buildStep(build db.Build, plan atc.Plan, buildState string) exec.Step {
+func (pb *planBuilder) buildStep(plan atc.Plan) exec.Step {
 	if plan.InParallel != nil {
-		return factory.buildParallelStep(build, plan, buildState)
+		return pb.buildParallelStep(plan)
 	}
 
 	if plan.Across != nil {
-		return factory.buildAcrossStep(build, plan, buildState)
+		return pb.buildAcrossStep(plan)
 	}
 
 	if plan.Do != nil {
-		return factory.buildDoStep(build, plan, buildState)
+		return pb.buildDoStep(plan)
 	}
 
 	if plan.Timeout != nil {
-		return factory.buildTimeoutStep(build, plan, buildState)
+		return pb.buildTimeoutStep(plan)
 	}
 
 	if plan.Try != nil {
-		return factory.buildTryStep(build, plan, buildState)
+		return pb.buildTryStep(plan)
 	}
 
 	if plan.OnAbort != nil {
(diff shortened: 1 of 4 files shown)


arakakimarcelo:
@taylorsilva can't seem to check what's wrong with validate-labels

taylorsilva:
Need to add you to the contributor team so you can read the build logs.

That check looks for an appropriate label, which I've added now.

taylorsilva:
Thanks for kicking off this PR!

I've reviewed the code and made some refactors to clean things up. I didn't like adding a third argument to all the functions, so I cleaned things up by introducing a `planBuilder` which takes 2/3 of the initial set of args. It's also clearer in the code that we only set build status for the `on_*` hook steps now.

The bigger UX decisions I made were the following:
- Change env var name from `BUILD_STATE` to `BUILD_STATUS` because we refer to it as "build status" everywhere else in the code base
- `BUILD_STATUS` is only set for the `on_[failure/success/error/abort]` hook steps. All other steps only ever saw `BUILD_STATUS=started`

I tried to see if there was an easy way to dynamically set `BUILD_STATUS` during execution of all the steps. Sadly it would involve a larger refactor. Because we're setting the status during the creating of the in-memory 'exec.Stepper` object, there's no way for us to easily inject the build status during runtime.

---

I manually verified this work by running this pipeline:
```yaml
resources:
  - name: env
    type: mock

jobs:
  - name: job
    plan:
      - &print_env
        put: env
        no_get: true
        params:
          version: ok
          print_env: true
      - task: ref-local-vars
        config:
          platform: linux
          image_resource:
            type: mock
            source: {mirror_self: true}
          run:
            path: sh
            args:
              - -c
              - |
                echo Hello
                exit 1
    on_success: *print_env
    on_failure: *print_env
    ensure: *print_env
    on_error: *print_env
    on_abort: *print_env
```
image
image
image
image

arakakimarcelo:
Very nice your solution @taylorsilva.
Only thing left is to resolve the unit tests, right?

taylorsilva:
I'm expecting all PR checks to pass. Once they pass I'll merge.

PR #9525, opened 2026-04-12
fix(atc): retain var_sources visibility inside across interpolation steps

### LINKED ISSUE #8191: var_sources not accessible when inside an across task
## Summary

I have 2 dummy var_sources defined. I am using across to iterate over one of the dummy vars but I get the following error when trying to also access the other dummy var inside the task. The same thing happens with accessing vault var_sources inside across.

```
failed to interpolate template: missing source 'dummyvar2' in var: dummyvar2:testkey
```

## Steps to reproduce

Create and run the following pipeline:

```
var_sources:
- name: dummyvar
  type: dummy
  config:
    vars:
      testkey: ["test1", "test2"]
- name: dummyvar2
  type: dummy
  config:
    vars:
      testkey: "test3"

jobs:
- name: placeholder
  plan:
  - across:
    - var: testiter
      values: ((dummyvar:testkey))
    task: use-vars
    config:
      platform: linux

      image_resource:
        type: mock
        source: {mirror_self: true}

      run:
        path: sh
        args:
        - -exc
        - |
          echo ((.:testiter))
          echo ((dummyvar2:testkey))
```

Running this without ` echo ((dummyvar2:testkey))` works fine - it iterates over `testkey` and echoes `test1` and `test2`.

## Expected results

The task should be able to access the other variable.

## Actual results

This error is produced when running the pipeline

```
failed to interpolate template: missing source 'dummyvar2' in var: dummyvar2:testkey
```

## Triaging info

* Concourse version:  v7.7.0
* Browser (if applicable): Chromium
* Did this used to work? No

arakakimarcelo:

closes #8191

* When the across step was interpolating the local vars introduced by its own step, it would erroneously error on all non-local `((.:local-var))` source vars
* This PR catches these `MissingSourceError`'s and ignores them
* A final interpolation gets done by the concrete step(s) inside the `across` step, which will resolve or error on any remaining vars

### DIFF
--- a/atc/engine/build_step_delegate.go
+++ b/atc/engine/build_step_delegate.go
@@ -307,7 +307,7 @@ func (delegate *buildStepDelegate) ConstructAcrossSubsteps(templateBytes []byte,
 		for j, v := range acrossVars {
 			localVars[v.Var] = values[j]
 		}
-		
+
 		interpolatedBytes, err := template.Evaluate(ignoreMissingSourceVars{vars.NamedVariables{".": localVars}}, vars.EvaluateOpts{})
 		if err != nil {
 			return nil, fmt.Errorf("failed to interpolate template: %w", err)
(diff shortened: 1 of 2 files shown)


taylorsilva (approved):
Verified I could reproduce the issue on `master` first using the pipeline from the referenced issue. I then built and tested with this PR and it worked:

Image

Also verified that if a source doesn't exist it'll error still:

Image

I added a nit commit. Will merge once checks pass.

taylorsilva:
The PR body doesn't match the change. Seems like AI summary from a previous attempt. I've updated the PR body to describe the actual changes.

PR #9533, opened 2026-04-15
Add TTL/expiry for task caches and `CONCOURSE_DEFAULT_TASK_CACHE_TTL`

taylorsilva:

This PR adds a `ttl` param to the `task.caches` object and a web node flag `CONCOURSE_DEFAULT_TASK_CACHE_TTL` which defaults to a value of `0s` (no expiry).

Task caches currently only expire if a pipeline or job is paused/archived/deleted. For an active pipeline/job with task caches, the caches will grow forever and never be GC'd. This can result in worker disks slowly filling up over several days or weeks unless the cache user takes deliberate steps to manage their task cache.

When upgrading to a version with this feature all existing task caches will receive a TTL of 0 (no expiry). If a default TTL is set, as task caches are used, their TTL will be updated to the default or the one configured in the task. The task-level configuration takes precedence over the web-node level setting.

Example pipeline:
```yaml
jobs:
- name: job
  plan:
  - task: cache
    config:
      platform: linux
      image_resource:
        type: mock
        source: {mirror_self: true}
      caches:
        - path: /cache/never-expires
        - path: /cache/expires
          ttl: 1m

      run:
        path: sh
        args:
        - -exc
        - |
          echo hello
```

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -280,6 +280,8 @@ type RunCommand struct {
 	DefaultPutTimeout  time.Duration `long:"default-put-timeout" description:"Default timeout of put steps"`
 	DefaultTaskTimeout time.Duration `long:"default-task-timeout" description:"Default timeout of task steps"`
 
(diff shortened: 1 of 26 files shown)


taylorsilva:
I've manually tested this to make sure task-level settings are respected and that the web-level default setting applies when no task-level setting is present.

PR #9537, opened 2026-04-17
Don't allow pipelines or teams containing '/' to be set with the set_pipeline step

### LINKED ISSUE #8475: Step `set-pipeline` can take invalid/problematic target pipeline name (in conjunction with instanced pipelines)

taylorsilva:

closes #8475 

The `set_pipeline` step will now error if a pipeline name contains forward slashes `/`, which is what `fly set-pipeline` currently does.

```
ERROR: pipeline name cannot contain '/'
```

We do this same check in the `fly set-pipeline` command: https://github.com/concourse/concourse/blob/6e4c92efde567111733698ffd49660762fbbaf68/fly/commands/set_pipeline.go#L40-L42

I verified that we don't need to enforce this for resource or job names. Fly currently let's you set these and the UI does not break when resources or jobs have `/` in their name.

image

### Update

`fly set-team` and the `set_pipeline` step will error on teams containing `/` as well now. Teams with this character already didn't work in the web UI. The fact we have no bug reports regarding this leads me to think that no one has tried making a team with a `/` in its name, so I don't think we're breaking anything for users.

Marking this as a breaking change though and will warn users to rename their teams before upgrading.

### DIFF
--- a/atc/exec/set_pipeline_step.go
+++ b/atc/exec/set_pipeline_step.go
@@ -98,6 +98,12 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState, delegate S
 		return false, nil
 	}
 
+	if strings.Contains(step.plan.Team, "/") {
+		fmt.Fprintln(delegate.Stderr(), "ERROR: team name cannot contain '/'")
+		delegate.Finished(logger, false)
+		return false, nil
+	}
+
 	source := setPipelineSource{
 		ctx:      ctx,
 		logger:   logger,
--- a/atc/exec/set_pipeline_step.go
+++ b/atc/exec/set_pipeline_step.go
@@ -92,6 +92,12 @@ func (step *SetPipelineStep) run(ctx context.Context, state RunState, delegate S
 		step.plan.Team = ""
 	}
 
+	if strings.Contains(step.plan.Name, "/") {
+		fmt.Fprintln(delegate.Stderr(), "ERROR: pipeline name cannot contain '/'")
+		delegate.Finished(logger, false)
+		return false, nil
+	}
+
 	source := setPipelineSource{
 		ctx:      ctx,
 		logger:   logger,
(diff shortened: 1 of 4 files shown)


ceco556:
I've tested the PR and it works as expected!

While testing, I also checked whether we have similar validation for `fly set-team`. It turns out we don’t, and this causes similar issue to the one with `set_pipeline` step.

## Example
```bash
fly -t local set-team -n test/main --config team.yml
```
I get deprecation warning:
```bash
DEPRECATION WARNING:
  - team: 'test/main' is not a valid identifier: illegal character '/'
```
After creating a pipeline in the new team and attempting to open it, nothing happens. The URL is:
```
http://localhost:8080/teams/test/main/pipelines/test-pipeline
```

Should we add the same validation to `fly set-team` with this PR or open a new one?

taylorsilva:
Thanks for checking that! Let's add it to this PR, it's small and similar enough to pack together.

ceco556 (approved):
LGTM!

Thank you!

PR #9542, opened 2026-04-20
atc: add base64 and hex encryption key flags

### LINKED ISSUE #1410: --encryption-key and --old-encryption-key flags are interpreted as RAW bytes, limiting available entropy
# Bug Report

(I apologize ahead if this should be better classified as a feature request, since this would requiring breaking compatibility or adding a new feature)

When parsing `--encryption-key` and `--old-encryption-key`, the provided string is just [converted to raw bytes](https://github.com/concourse/atc/blob/ff7a0687605e6d2ba1d61e1c3d9dadf1cc82c23a/atccmd/cipher_flag.go#L14), instead of being parsed as base64 or hex as is usually the case with encryption keys. This means that some byte values (e.g. the NUL byte) can't be easily supplied, and other byte values (1-31, 127-255) are hard to supply in a consistent way. With only 95 safe character values per byte, we lose 1.5 bits of entropy on the byte, so a nominally 128-bit key will only have 104 bits of entropy available and a 256-bit key would have 208 bits of entropy.

Parsing keys as hex-encoded or base64-encoded values will easily solve the issue, but is going to break down backward compatibility quite spectacularly, so Concourse will need to support a new set of arguments, e.g. `--encryption-key-hex` and `--old-encryption-key-hex` or `--encryption-key-base64` and `--old-encryption-key-base64`.

samuelBarreto:

closes #1410

- [x] Added `--encryption-key-base64` and `--encryption-key-hex` flags to `RunCommand` struct
- [x] Added `--old-encryption-key-base64` and `--old-encryption-key-hex` flags to `RunCommand` struct
- [x] Added the same flags to `Migration` struct
- [x] Updated `RunCommand.newKey()` and `RunCommand.oldKey()` to resolve keys from all three formats (raw → base64 → hex)
- [x] Added `Migration.resolveNewKey()` and `Migration.resolveOldKey()` helper functions
- [x] Updated `atc/scripts/db_decrypt.go` to support the new key formats

**Reasoning:** The existing `--encryption-key` flag interprets the key as raw bytes (`[]byte(val)`), limiting entropy to ~95 printable ASCII characters per byte. A nominally 256-bit key only provides ~208 bits of entropy. The new flags allow users to supply keys encoded in base64 or hex, enabling the full byte range (0x00-0xFF) and full 256-bit entropy.

The `CipherBase64` and `CipherHex` types used by these flags were already merged into `flag/cipher.go` via #9541.

## Notes to reviewer

- Backward compatible: the existing `--encryption-key` and `--old-encryption-key` flags are unchanged
- Key resolution priority: raw → base64 → hex (first non-nil wins)
- Only one format should be provided at a time; if multiple are set, ~the priority order above applies~ an error is returned
- Validated locally via `docker compose up --build` — new flags appear in `concourse web --help`

## Release Note

Added `--encryption-key-base64`, `--encryption-key-hex`, `--old-encryption-key-base64`, and `--old-encryption-key-hex` flags, allowing encryption keys to be provided in base64 or hex encoding for full 256-bit entropy support.

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -150,7 +150,7 @@ type RunCommand struct {
 	EncryptionKey          flag.Cipher       `long:"encryption-key"            description:"A 16 or 32 length key used to encrypt sensitive information before storing it in the database."`
 	EncryptionKeyBase64    flag.CipherBase64 `long:"encryption-key-base64"     description:"A base64-encoded 16 or 32 byte key used to encrypt sensitive information before storing it in the database."`
 	EncryptionKeyHex       flag.CipherHex    `long:"encryption-key-hex"        description:"A hex-encoded 16 or 32 byte key used to encrypt sensitive information before storing it in the database."`
-	OldEncryptionKey       flag.Cipher       `long:"old-encryption-key"        description:"Encryption key previously used for encrypting sensitive information. If provided without a new key, data is encrypted. If provided with a new key, data is re-encrypted."`
+	OldEncryptionKey       flag.Cipher       `long:"old-encryption-key"        description:"Encryption key previously used for encrypting sensitive information. If provided without a new key, data is decrypted. If provided with a new key, data is re-encrypted."`
 	OldEncryptionKeyBase64 flag.CipherBase64 `long:"old-encryption-key-base64" description:"Base64-encoded encryption key previously used. If provided without a new key, data is decrypted. If provided with a new key, data is re-encrypted."`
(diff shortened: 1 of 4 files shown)


taylorsilva (approved):
LGTM - I pushed a commit that refactors some of the code from `command.go` into `encryption_key.go`. I also made one behavioural change: if multiple key formats are provided an error is returned instead of just picking the first one. We don't want to assume things and potentially confuse operators about which key the db was encrypted with.

Thanks for working on this! Once the PR checks pass again I'll merge it.

taylorsilva:
TODO: will need to add these env vars to the helm chart and bosh packaging

Issue #2726, opened 2018-10-24
USABILITY: Make a missing task image/image_resource have a better error message


jghiloni:
> Please remove all lines starting with '>' before submitting!
>
> Bug reports on this repo should be related to core Concourse functionality.
>
> If the bug has more to do with *running* Concourse, consider opening the
> issue on a platform-specific repository instead:
>
>    BOSH: https://github.com/concourse/concourse-bosh-release
>    Docker: https://github.com/concourse/concourse-docker
>    Kubernetes: https://github.com/concourse/charts
>
> It's OK if you open it here, but we may move it to a more appropriate repo.
>
> If you're not sure that this is a bug and just need help with something,
> please ask in our support forum instead:
>
>     https://discuss.concourse-ci.org/c/support

## Bug Report

When creating a task whose `image` or `image_resource` has been omitted, the error message is not useful: 

```
Backend error: Exit status: 500, message: {"Type":"","Message":"runc exec: exit status 1: exec failed: container_linux.go:348: starting container process caused \"no such file or directory\"\n","Handle":"","ProcessID":"","Binary":""}
```

This is hard to determine whether or not the error is a missing an image or the task definition itself is missing.

## Steps to Reproduce

1. Create a task in a pipeline whose `image_resource` or `image` field is missing
1. ???
1. Profit!

## Expected Results

1. A useful error message that no image was available

## Actual Results

1. A not useful error message that file was not found

## Version Info

* Concourse version: 4.2.1
* Deployment type (BOSH/Docker/binary): BOSH
* Infrastructure/IaaS: vSphere
* Browser (if applicable):
* Did this used to work? Probably not ¯\_(ツ)_/¯?

edtan:
On v5.1.0, it now says "no image plugin configured" with this sample pipeline:

```
jobs:
- name: test
  plan:
  - task: missing-image
    config:
      platform: linux
      run:
        path: sh
        args:
          - -exc
          - |
            ls
```

vito:
The error bubbles up from the Garden backend. Doing anything more here is kind of tricky. On Windows and Darwin there is no valid `image` or `image_resource` to set, so it's up to the backend to handle it. In these two cases there is no notion of an image or rootfs so they just run your task in an ephemeral directory. The Linux backend we use (Garden-runC) is equipped to also handle this by using a default rootfs, if configured, however we don't have such a thing configured in the binary distribution and don't even have an image plugin configured. Hence the error.

[...]

PR #9563, opened 2026-05-01
Emit a more useful error if we get the "rootfs empty" error

taylorsilva:

If a user doesn't specify `image` or `image_resource`, the containerd runtime will return "rootfs must not be empty" as the error. This PR emits the error from containerd still, but appends a useful message suggesting the user check their job or task config for an image. The most common cause of this error is a user no specifying an image for the task.

An image is not required for all tasks. For workers using the houdini runtime, not image is required. That's why we can't just error if no image is specified or even do validation of the step/task config.


taylorsilva:
Closes #2726

PR #9564, opened 2026-05-06
Add SimpleIcons support in pipeline resources

### LINKED ISSUE #8803: Material Design - Brand Icons deprecated
## Summary

At some point all brand icons were tagged as deprecated as explained in https://pictogrammers.com/docs/library/mdi/guides/brand-icons/

The brand icons are now part of 
https://github.com/Pictogrammers/brand-icons

but not of the original material design icons set anymore.

## Steps to reproduce

This isn't yet a problem, but will become one with a future™ update and I'd prefer to maintain the descriptive "where does this live" nature of my pipelines.

jghiloni:
Closes #8803

Material Design Icons are well and good, but folks like using brand-specific icons, which are largely deprecated in MDI. This feature adds support for Simple Icons, which include brand icons that are not deprecated.

This includes the code to use the icons, as well as a hack script to generate the `si-svg.min.js` which differs slightly from its MDI counterpart by including the full SVG, not just the paths, allowing SI variants to have alt-text in the UI. The hack script also generates a markdown file in the root of the repository with links to the icons along with their titles and slugs. Simply run `yarn update-si-svg` when the version of the `simple-icons` npm package generates and everything will be updated.

## Release Note

View `simple-icons.md` for a list of icons supported, as well as links to preview them

### DIFF
--- a/hack/update-si-svg/main.go
+++ b/hack/update-si-svg/main.go
@@ -1,0 +1,153 @@
+package main
+
+import (
+	"encoding/json"
+	"encoding/xml"
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+type clArgs struct {
+	pathToNodeModule string
+	pathToOutputJS   string
+	pathToMarkdown   string
+}
+
+type iconMetadata struct {
+	Title string `json:"title"`
+	Slug  string `json:"slug"`
+}
+
+type nodeVersion struct {
+	Version string `json:"version"`
+}
+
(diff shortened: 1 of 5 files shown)


taylorsilva (approved):
LGTM - In the issue this PR closes, there were concerns raised about the size of the file generated, which is about 5MB. This does get compressed to about 2.6MB when sent to the browser and is then cached. So we're not constantly sending this over the wire on every full load of the UI. I am fine with this trade-off overall. It didn't seem to make the UI load slower.

Issue #9449, opened 2026-02-02
Migrate to netfilter (nftables)


taylorsilva:
### Summary

We should move off of invoking `iptables` in our containerd runtime and use `nftables` (netfilter) instead. iptables has been labelled as legacy for MANY years now and are not enabled on newer kernels. Also seeing errors like this more often in our containerd integration tests:

```
starting task: network add: cni net setup: 
  plugin type="bridge" failed (add):
  failed to list chains: running [/usr/bin/ip6tables -t nat -S --wait]:
  exit status 3: ip6tables v1.8.11 (legacy):
  can't initialize ip6tables table `nat': Table does not exist (do you need to insmod?)
Perhaps ip6tables or your kernel needs to be upgraded.
```

Will probably change behaviour to detect if nftables is present and prefer using it over iptables, with iptables being present as a fallback.

### Context

Past issues:
* https://github.com/concourse/concourse/issues/5136
* https://github.com/concourse/concourse/issues/4557

Docs for research:
https://wiki.nftables.org/wiki-nftables/index.php/Moving_from_iptables_to_nftables
https://github.com/google/nftables

taylorsilva:
Reviewing old issues, it looks like we thought about using the CNI plugin for firewalld. firewalld is present on most major linux distributions now. https://firewalld.org/ lists:

- RHEL 7 and newer
- CentOS 7 and newer
- Fedora 18 and newer
- SUSE 15 and newer
- OpenSUSE 15 and newer

There are some big ones missing from this list, like Ubuntu and Debian. Ubuntu uses ufw and Debian appears to not come with anything on top of netfilter. You can install firewalld easily on all major distributions. Vito noticed the same thing https://github.com/concourse/concourse/issues/5136#issuecomment-645488765 and he referenced the nftables go pkg from Google too. Funny, looks like the landscape hasn't changed that much.

I think I prefer directly building on nftables instead of on firewalld right now. Doesn't add a pre-req when setting up a Worker. Going to start proceeding in that direction using the Go package from Google: https://github.com/google/nftables

taylorsilva:
Looking at https://github.com/concourse/concourse/blob/1dfdf121f7245360722f693ccd46d5bd15e6ece5/worker/runtime/iptables/iptables.go#L24
[...]

PR #9565, opened 2026-05-07
fix: DNS resolution for Ubuntu 24 stemcells

Kump3r:
# Fix DNS resolution on Ubuntu 24 (Noble) stemcells

## Problem

On Concourse workers with Ubuntu Noble (24.04) stemcells, container DNS resolution was failing with:
```
dial tcp: lookup some.address.com on [::1]:53: read udp [::1]:58930->[::1]:53: read: connection refused
```

### Root Cause

**Upstream resolv.conf behavior changed between Ubuntu Jammy and Noble:**

- **Jammy** (22.04LTS): `/etc/resolv.conf` directly contained a routable DNS IP (e.g., `nameserver 172.18.96.9`)
- **Noble** (24.04LTS): `/etc/resolv.conf` is now a systemd-resolved stub resolver symlink with only loopback nameservers:
  ```
  nameserver 127.0.0.53
  nameserver ::1
  ```

**Concourse's DNS parser issue:**

The regex patterns in `worker/runtime/resolvconf_parser.go` only matched IPv4 loopback (`127.0.0.x`), not IPv6 loopback (`::1`). When systemd-resolved's stub was detected as loopback-only, the parser would:
1. Fall back to `localip.LocalIP()` (the worker VM IP)
2. Write that to the container's `/etc/resolv.conf`
3. Containers then tried to query that VM IP, which was unreachable from the container network namespace
4. This cascaded to DNS defaulting to `[::1]:53`, which didn't exist, causing connection refused

## Solution

Updated `worker/runtime/resolvconf_parser.go` to:

1. **Add IPv6 loopback to regex patterns** — both `127.0.0.x` and `::1` are now recognized as loopback and handled uniformly
2. **Implement systemd-resolved upstream fallback** — when host `/etc/resolv.conf` is loopback-only on Noble, parser now:
   - Attempts to read upstream resolvers from `/run/systemd/resolve/resolv.conf` 
   - Uses those if available (correct for Noble)
   - Falls back to worker IP only if upstream file is unavailable (backward compatibility for older/edge cases)
3. **Preserve non-loopback passthrough** — if `/etc/resolv.conf` contains any usable non-loopback nameserver, it passes through unchanged (Jammy behavior preserved)

### Modified Files

- **`worker/runtime/resolvconf_parser.go`**
  - Updated `loopbackNameserverRegexp` to match both `127.0.0.x` and `::1` with multiline mode
  - Updated `loopbackIPRegexp` to strip both IPv4 and IPv6 loopback
  - Added `systemdResolvedResolvConfPath` constant pointing to systemd upstream resolv.conf
  - Refactored logic to:
    - Parse entries first, return immediately if any valid nameserver exists
    - Only attempt fallbacks when all nameservers are loopback
    - Try systemd-resolved upstream file before falling back to worker IP
  - Added testable hooks (`readFileFn`, `localIPFn`) for deterministic testing without exporting test-only APIs

- **`worker/runtime/resolvconf_parser_internal_test.go`** (new file)
  - Tests systemd-resolved upstream fallback behavior (Noble scenario)
  - Tests local IP fallback when upstream is unavailable (backward compatibility)
  - Uses internal package scope to access unexported test helpers

### Why Separate Test File?

The internal tests need access to unexported helpers (`readFileFn`, `localIPFn`, default path constants) to stub I/O deterministically. Placing them in package `runtime` (not `runtime_test`) allows this without polluting the public API. The existing `resolvconf_parser_test.go` remains in package `runtime_test` for black-box behavior validation.

## Testing

- ✅ Validated with manually built binary on **Ubuntu Noble (24.04)** stemcell — containers now resolve DNS correctly
- ✅ Validated backward compatibility on **Ubuntu Jammy (22.04)** stemcell — existing behavior unchanged
- ✅ All existing runtime tests pass: `ginkgo -r ./worker/runtime`
- ✅ New internal tests confirm both fallback paths work correctly

## Behavior Summary

| Scenario | Before | After |
|----------|--------|-------|
| **Jammy** with upstream DNS in `/etc/resolv.conf` | ✓ Passes through | ✓ Passes through |
| **Noble** with systemd stub only | ✗ Tries `[::1]:53` → connection refused | ✓ Uses systemd upstream resolvers |
| **Edge case**: No upstream available | Uses worker IP | Uses worker IP (fallback) |

## Backward Compatibility

- Jammy and older stemcells with routable DNS IPs in `/etc/resolv.conf` are unaffected
- If systemd upstream file is missing (legacy/custom setups), behavior reverts to worker IP (same as before)
- No configuration changes required; this is fully transparent

## Related Issue

This fixes DNS resolution failures introduced by Ubuntu 24's systemd-resolved stub resolver changes for Concourse workers on Noble stemcells.

### DIFF
--- a/worker/runtime/resolvconf_parser.go
+++ b/worker/runtime/resolvconf_parser.go
@@ -12,30 +12,77 @@ import (
 )
 
 var (
+	defaultHostResolvConfPath     = "/etc/resolv.conf"
+	systemdResolvedResolvConfPath = "/run/systemd/resolve/resolv.conf"
+	readFileFn                    = os.ReadFile
+	localIPFn                     = localip.LocalIP
+
 	// Regular expressions used for DNS resolution parsing
-	loopbackNameserverRegexp = regexp.MustCompile(`^\s*nameserver\s+127\.0\.0\.\d+\s*$`)
-	loopbackIPRegexp         = regexp.MustCompile(`127\.\d{1,3}\.\d{1,3}\.\d{1,3}`)
+	loopbackNameserverRegexp = regexp.MustCompile(`(?m)^\s*nameserver\s+(?:127\.0\.0\.\d+|::1)\s*$`)
+	loopbackIPRegexp         = regexp.MustCompile(`(?:127\.\d{1,3}\.\d{1,3}\.\d{1,3}|::1)`)
 )
 
 // Parse resolve.conf file from the provided path.
 // implementation is based on guardian's implementation
 // here: https://github.com/cloudfoundry/guardian/blob/master/kawasaki/dns/resolv_compiler.go
 func ParseHostResolveConf(path string) ([]string, error) {
-	resolvConf, err := os.ReadFile(path)
+	resolvContents, err := readResolvConf(path)
 	if err != nil {
-		return nil, fmt.Errorf("unable to read host's resolv.conf: %w", err)
+		return nil, err
 	}
 
-	resolvContents := string(resolvConf)
+	entries := parseResolvEntries(resolvContents)
+	if hasNameserverEntry(entries) {
+		return entries, nil
+	}
 
 	if loopbackNameserverRegexp.MatchString(resolvContents) {
-		ip, err := localip.LocalIP()
+		if path == defaultHostResolvConfPath {
+			entries, err := parseResolvConf(systemdResolvedResolvConfPath)
+			if err == nil && hasNameserverEntry(entries) {
+				return entries, nil
+			}
+		}
+
+		ip, err := localIPFn()
 		if err != nil {
 			return nil, err
 		}
 		return []string{"nameserver " + ip}, nil
 	}
 
+	return entries, nil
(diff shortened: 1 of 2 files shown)


Kump3r:
Well the `containerd-integration` is failing due to the flakey issue https://github.com/concourse/concourse/issues/9449
@taylorsilva can we prioritise this as newer ubuntu stemcells are impacted. Would it be possible to include this in v8.2.0?

taylorsilva:
Thanks Kalin! Fine with getting this into v8.2.0

Yeah that flaky test error is so weird. I re-deployed the workers last night which always causes this issue. I don't know why, but when re-deployed we'll see this error, but after a few hours it just stops, as if someone ran `insmod` to install the missing module(s) 😕

taylorsilva:
Yup and now it just worked! Weirdddddd

Will review and merge shortly.

taylorsilva (approved):

PR #9569, opened 2026-05-11
fix: handle all IPv4 loopback resolvers

immanuwell:
Small follow-up to #9565. The parser already drops any `127.x.x.x` nameserver from container resolv.conf, but the fallback check only noticed `127.0.0.x`. So `127.1.2.3` got dropped and then... no nameserver. tiny DNS papercut, imo.

Repro, before this patch:
```sh
go test ./worker/runtime -run TestParseHostResolveConf_FallsBackToLocalIPForAnyIPv4LoopbackNameserver -count=1
```

It failed with `actual: []string(nil)`, because fallback never kicked in.

Fix is just making the loopback detector match the same IPv4 loopback range the parser already filters: `127.0.0.0/8`.

Checked:
```sh
go test ./worker/runtime -count=1
git diff origin/master...HEAD --check
```

### DIFF
--- a/worker/runtime/resolvconf_parser.go
+++ b/worker/runtime/resolvconf_parser.go
@@ -18,7 +18,7 @@ var (
 	localIPFn                     = localip.LocalIP
 
 	// Regular expressions used for DNS resolution parsing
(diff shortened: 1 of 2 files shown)


Kump3r (approved):
Nice catch! Thanks I missed this as I was focused on 127.0.0.1 and it was this way in the past implementation as well. Didn’t consider the alternative.

PR #9576, opened 2026-05-18
Remove legacy `X-XSS-Protection` response header

IvanChalukov:

* Removes the deprecated `X-XSS-Protection` header from ATC responses.
* Updates tests to reflect the header removal.
* Aligns Concourse with current browser guidance by dropping a non-standard legacy header that is no longer recommended - https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-XSS-Protection

## Notes to reviewer

`X-XSS-Protection` is a legacy header that modern browsers do not meaningfully rely on. Removing it should not change behavior for supported modern browsers.

## Release Note

* Concourse no longer sends the deprecated `X-XSS-Protection` response header.

### DIFF
--- a/atc/wrappa/security_handler.go
+++ b/atc/wrappa/security_handler.go
@@ -19,7 +19,6 @@ func (handler SecurityHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 	if handler.StrictTransportSecurity != "" {
 		w.Header().Set("Strict-Transport-Security", handler.StrictTransportSecurity)
 	}
-	w.Header().Set("X-XSS-Protection", "1; mode=block")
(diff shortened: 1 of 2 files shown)


taylorsilva (approved):
LGTM

Read these two pages:
- https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Headers_Cheat_Sheet.html#x-xss-protection
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-XSS-Protection

Both recommend not setting this header and using CSP instead. We added support for configuring CSP in https://github.com/concourse/concourse/pull/6949 with `CONCOURSE_CONTENT_SECURITY_POLICY`

PR #9583, opened 2026-05-22
atc: behaviour: add origin_pipeline to OPA payload

### LINKED ISSUE #9507: Enhance SetPipeline OPA input to include both source and target pipeline metadata
### Summary

The [recent discussion we had about OPA](https://github.com/orgs/concourse/discussions/9480#discussioncomment-16184457) reminded me of a minor issue we noticed in our Concourse/OPA deployment.

I will try to prepare the PR to address this, but let's have it officially reported as an issue before.

**Context:**

When Concourse is integrated with Open Policy Agent (OPA), it sends request data to an external policy engine to authorize actions. This includes metadata about the action being performed, the pipeline involved, and the configuration data being submitted.

Currently, Concourse handles two primary actions for updating configurations:

SaveConfig: Triggered via fly set-pipeline.

SetPipeline: Triggered via the set_pipeline step within an existing pipeline (often used in "bootstrap" or "parent" pipeline patterns).

**The Problem**

There is an inconsistency in how the pipeline context is reported in the OPA request body between these two actions.

In SaveConfig, the pipeline field represents the pipeline being created or updated.

**In SetPipeline, the pipeline field represents the calling pipeline (the one running the step), while the configuration in the data payload refers to the target pipeline.**

This creates a limitation for fine-grained security policies. If a user wants to whitelist a specific task or resource within a specific target pipeline, they cannot do so uniquely because the request context doesn't explicitly distinguish the target pipeline name from the source. This forces policies to be overly broad (e.g., whitelisting a task name globally across all pipelines managed by a specific bootstrap pipeline).

**Proposed Solution**

Modify the Concourse OPA input payload for the SetPipeline action to include both the source (setter) and the target (being set).

I suggest adding a specific field to clarify the relationship. For example:

Current Payload Structure:

```json
{
  "action": "SetPipeline",
  "pipeline": "parent-pipeline",
  "data": { ... }
}
```
Proposed Payload Structure:

```json
{
  "action": "SetPipeline",
  "pipeline": "target-pipeline-name",
  "origin_pipeline": "parent-pipeline",
  "data": { ... }
}
```

### Steps to Reproduce

1. Set up the Concourse with OPA
2. Configure a pipeline with a set-pipeline task that sets another pipeline
3. Inspect the payload OPA receives from Concourse. 

### Expected Results

The payload OPA receives from Concourse should include both parent pipeline name and the pipeline name we are trying to set. This allows to write the OPA rules that target the correct pipeline (the one we are trying to set) and not the parent pipeline.

### Actual Results

Inspect the payload OPA receives from Concourse. At the moment, only the parent pipeline name is shown, not the actual pipeline we are trying to set.

### Web Node(s) configuration

_No response_

### Worker(s) configuration

_No response_

### Concourse Version

7.14.3 (not applicable actually)

### Browser (if applicable)

_No response_

### Did this use to work?

No

Petahhh:

Pipelines can be set via fly or in a step. In the latter case, there are two pipelines: the origin pipeline that has been run and the target pipeline actually being set.

The payload sent to OPA server now distinguishes the two. `pipeline` will indicate the pipeline being set where `origin_pipeline` refers to the pipeline doing the set-pipeline.

Docs PR: https://github.com/concourse/docs/pull/608

closes #9507

## Release Note

* Breaking Change: Set-Pipeline OPA payload improvement
  * the payload sent to an OPA server when setting pipeline now includes `origin_pipeline`
  * In the case of pipeline being set by another pipeline, `pipeline` previously referred to the setting pipeline and not the target pipeline
  * the setting pipeline is now available in `origin_pipeline` and `pipeline` refers to the pipeline being set

### DIFF
--- a/atc/engine/set_pipeline_delegate.go
+++ b/atc/engine/set_pipeline_delegate.go
@@ -50,15 +50,16 @@ func (delegate *setPipelineStepDelegate) SetPipelineChanged(logger lager.Logger,
 	logger.Debug("set pipeline changed")
 }
 
-func (delegate *setPipelineStepDelegate) CheckRunSetPipelinePolicy(atcConfig *atc.Config) error {
+func (delegate *setPipelineStepDelegate) CheckRunSetPipelinePolicy(targetPipeline string, atcConfig *atc.Config) error {
 	if !delegate.policyChecker.ShouldCheckAction(policy.ActionRunSetPipeline) {
 		return nil
 	}
 
 	return delegate.checkPolicy(policy.PolicyCheckInput{
-		Action:   policy.ActionRunSetPipeline,
-		Team:     delegate.build.TeamName(),
-		Pipeline: delegate.build.PipelineName(),
-		Data:     atcConfig,
+		Action:         policy.ActionRunSetPipeline,
+		Team:           delegate.build.TeamName(),
+		Pipeline:       targetPipeline,
+		OriginPipeline: delegate.build.PipelineName(),
+		Data:           atcConfig,
 	})
 }
--- a/atc/engine/set_pipeline_delegate_test.go
+++ b/atc/engine/set_pipeline_delegate_test.go
@@ -83,7 +83,7 @@ var _ = Describe("SetPipelineStepDelegate", func() {
 				},
 			}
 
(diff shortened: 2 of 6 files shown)


taylorsilva (approved):
LGTM - TY Peter!

PR #9588, opened 2026-05-27
feat: Add health endpoint basic structures and routes

Kump3r:

Initial work based on https://github.com/concourse/rfcs/pull/141

- __`atc/health.go`__ - Health DTOs and constants 
- __`atc/health_test.go`__ - Comprehensive unit tests
- __`atc/routes.go`__ - Modified with health route constants and mapping

## Notes to reviewer
This is the very beginning of the `/api/v1/health` endpoint feature. The focus of this PR is to establish basic structure and data types. Actual logic will follow in separate PR's following the "Implementation plan" I created in the rfc PR. The focus on the review here that I desire is the design and validity of the data structures and JSON schema. Since I will follow-up with the actual health server in the next PR, I used stubs for now in places, where the handler was expected.

### DIFF
--- a/atc/api/api_suite_test.go
+++ b/atc/api/api_suite_test.go
@@ -59,6 +59,7 @@ var (
 	dbWorkerFactory         *dbfakes.FakeWorkerFactory
 	dbWorkerTeamFactory     *dbfakes.FakeTeamFactory
 	dbWorkerLifecycle       *dbfakes.FakeWorkerLifecycle
+	fakeDbConn              *dbfakes.FakeDbConn
 	build                   *dbfakes.FakeBuild
 	dbBuildFactory          *dbfakes.FakeBuildFactory
 	dbUserFactory           *dbfakes.FakeUserFactory
@@ -138,6 +139,7 @@ var _ = BeforeEach(func() {
 
 	dbWorkerFactory = new(dbfakes.FakeWorkerFactory)
 	dbWorkerLifecycle = new(dbfakes.FakeWorkerLifecycle)
+	fakeDbConn = new(dbfakes.FakeDbConn)
 
 	fakeWorkerPool = new(apifakes.FakePool)
 
@@ -209,6 +211,8 @@ var _ = BeforeEach(func() {
 		dbResourceConfigFactory,
 		dbUserFactory,
 		dbComponentFactory,
+		fakeDbConn,
+		1, // minWorkerCount
 
 		constructedEventHandler.Construct,
 
--- a/atc/api/api_suite_test.go
+++ b/atc/api/api_suite_test.go
@@ -212,7 +212,8 @@ var _ = BeforeEach(func() {
 		dbUserFactory,
 		dbComponentFactory,
 		fakeDbConn,
-		1, // minWorkerCount
+		1,   // minWorkerCount
+		2.0, // componentStaleMultiplier
 
 		constructedEventHandler.Construct,
 
--- a/atc/api/api_suite_test.go
+++ b/atc/api/api_suite_test.go
@@ -167,16 +167,13 @@ var _ = BeforeEach(func() {
 
 	build = new(dbfakes.FakeBuild)
 
-	checkPipelineAccessHandlerFactory := auth.NewCheckPipelineAccessHandlerFactory(dbTeamFactory)
-
-	checkBuildReadAccessHandlerFactory := auth.NewCheckBuildReadAccessHandlerFactory(dbBuildFactory)
-
-	checkBuildWriteAccessHandlerFactory := auth.NewCheckBuildWriteAccessHandlerFactory(dbBuildFactory)
-
-	checkWorkerTeamAccessHandlerFactory := auth.NewCheckWorkerTeamAccessHandlerFactory(dbWorkerFactory)
-
 	fakePolicyChecker = new(policycheckerfakes.FakePolicyChecker)
 	fakePolicyChecker.CheckReturns(policy.PassedPolicyCheck(), nil)
+
+	checkPipelineAccessHandlerFactory := auth.NewCheckPipelineAccessHandlerFactory(dbTeamFactory)
+	checkBuildReadAccessHandlerFactory := auth.NewCheckBuildReadAccessHandlerFactory(dbBuildFactory)
+	checkBuildWriteAccessHandlerFactory := auth.NewCheckBuildWriteAccessHandlerFactory(dbBuildFactory)
+	checkWorkerTeamAccessHandlerFactory := auth.NewCheckWorkerTeamAccessHandlerFactory(dbWorkerFactory)
 
 	apiWrapper := wrappa.MultiWrappa{
 		wrappa.NewPolicyCheckWrappa(logger, fakePolicyChecker),
@@ -263,6 +260,50 @@ var _ = AfterEach(func() {
 	os.Remove(cliDownloadsDir)
 	server.Close()
 })
+
+// buildTestServer constructs and starts a test HTTP server with the given minWorkerCount.
+// Use this in tests that need a non-default minWorkerCount (e.g. to exercise the degraded worker path).
+// Callers are responsible for closing the returned server.
+func buildTestServer(workerCount int) *httptest.Server {
+	checkPipelineAccessHandlerFactory := auth.NewCheckPipelineAccessHandlerFactory(dbTeamFactory)
+	checkBuildReadAccessHandlerFactory := auth.NewCheckBuildReadAccessHandlerFactory(dbBuildFactory)
+	checkBuildWriteAccessHandlerFactory := auth.NewCheckBuildWriteAccessHandlerFactory(dbBuildFactory)
+	checkWorkerTeamAccessHandlerFactory := auth.NewCheckWorkerTeamAccessHandlerFactory(dbWorkerFactory)
+
+	apiWrapper := wrappa.MultiWrappa{
+		wrappa.NewPolicyCheckWrappa(logger, fakePolicyChecker),
+		wrappa.NewAPIAuthWrappa(
+			checkPipelineAccessHandlerFactory,
+			checkBuildReadAccessHandlerFactory,
+			checkBuildWriteAccessHandlerFactory,
+			checkWorkerTeamAccessHandlerFactory,
+		),
+	}
+
+	handler, err := api.NewHandler(
+		logger,
+		externalURL, "", clusterName,
+		apiWrapper,
+		dbTeamFactory, dbPipelineFactory, dbJobFactory, dbResourceFactory,
+		dbWorkerFactory, dbWorkerTeamFactory, fakeVolumeRepository, fakeContainerRepository,
+		fakeDestroyer, dbBuildFactory, dbCheckFactory, dbResourceConfigFactory,
+		dbUserFactory, dbComponentFactory, fakeDbConn,
+		workerCount,
+		2.0,
+		constructedEventHandler.Construct,
+		fakeWorkerPool, sink, isTLSEnabled, cliDownloadsDir,
+		"1.2.3", "4.5.6",
+		fakeSecretManager, fakeVarSourcePool, credsManagers,
+		interceptTimeoutFactory, time.Second, dbWall, fakeClock, dbSigningKeyFactory,
+	)
+	Expect(err).NotTo(HaveOccurred())
+
+	accessorHandler := accessor.NewHandler(
+		logger, "some-action", handler, fakeAccessor,
+		new(auditorfakes.FakeAuditor), map[string]string{},
+	)
+	return httptest.NewServer(wrappa.LoggerHandler{Logger: logger, Handler: accessorHandler})
+}
 
 func TestAPI(t *testing.T) {
 	RegisterFailHandler(Fail)
--- a/atc/api/api_suite_test.go
+++ b/atc/api/api_suite_test.go
@@ -85,6 +85,11 @@ var (
 	client *http.Client
 )
 
+func TestAPI(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "API Suite")
+}
+
 type fakeEventHandlerFactory struct {
 	build db.BuildForAPI
 
@@ -170,6 +175,31 @@ var _ = BeforeEach(func() {
 	fakePolicyChecker = new(policycheckerfakes.FakePolicyChecker)
 	fakePolicyChecker.CheckReturns(policy.PassedPolicyCheck(), nil)
 
+	server = newApiTestServer()
+
+	client = &http.Client{
+		Transport: &http.Transport{},
+	}
+})
+
+var _ = AfterEach(func() {
+	os.Remove(cliDownloadsDir)
+	server.Close()
+})
+
+func newApiTestServer(opts ...apiConfigOpt) *httptest.Server {
+	testServerConfig := &apiServerConfig{
+		workerCount:              1,
+		componentStaleMultiplier: 2.0,
+		concourseVersion:         "1.2.3",
+		workerVersion:            "4.5.6",
+		interceptUpdateInterval:  time.Second,
+	}
+
+	for _, opt := range opts {
+		opt(testServerConfig)
+	}
+
 	checkPipelineAccessHandlerFactory := auth.NewCheckPipelineAccessHandlerFactory(dbTeamFactory)
 	checkBuildReadAccessHandlerFactory := auth.NewCheckBuildReadAccessHandlerFactory(dbBuildFactory)
 	checkBuildWriteAccessHandlerFactory := auth.NewCheckBuildWriteAccessHandlerFactory(dbBuildFactory)
@@ -189,7 +219,7 @@ var _ = BeforeEach(func() {
 		logger,
 
 		externalURL,
-		"",
+		"", // OIDC issuer
 		clusterName,
 
 		apiWrapper,
@@ -209,8 +239,8 @@ var _ = BeforeEach(func() {
 		dbUserFactory,
 		dbComponentFactory,
 		fakeDbConn,
-		1,   // minWorkerCount
-		2.0, // componentStaleMultiplier
+		testServerConfig.workerCount,
+		testServerConfig.componentStaleMultiplier,
 
 		constructedEventHandler.Construct,
 
@@ -221,13 +251,13 @@ var _ = BeforeEach(func() {
 		isTLSEnabled,
 
 		cliDownloadsDir,
-		"1.2.3",
-		"4.5.6",
+		testServerConfig.concourseVersion,
+		testServerConfig.workerVersion,
 		fakeSecretManager,
 		fakeVarSourcePool,
 		credsManagers,
 		interceptTimeoutFactory,
-		time.Second,
+		testServerConfig.interceptUpdateInterval,
 		dbWall,
 		fakeClock,
 		dbSigningKeyFactory,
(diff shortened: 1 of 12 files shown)


Kump3r:
## PR Checklist: `health-endpoint-base`

### Implementation

#### Health Server Package (`atc/api/healthserver/`)
- [x] `server.go` — `Server` struct with `dbConn`, `workerFactory`, `minWorkerCount`
- [x] `health.go` — `GetHealth` handler with DB + worker logic
- [x] DB health: `SELECT pg_is_in_recovery()` with 2s timeout — detects unreachable and read-only (standby)
- [x] Worker health: `total==0 || running==0` → failing; `running < minWorkerCount` → degraded; `running >= minWorkerCount` → ok
- [x] HTTP 503 on `failing`, HTTP 200 on `ok`/`degraded`

#### DTOs (`atc/health.go`)
- [x] `Health`, `DatabaseHealth`, `WorkerHealth` structs with correct JSON tags
- [x] Overall status constants: `ok`, `degraded`, `failing`
- [x] Subsystem status constants: `healthy`, `unhealthy`

#### HTTP Wiring (`atc/api/handler.go`)
- [x] `http.NotFound` stub replaced with `healthserver.GetHealth`
- [x] `dbConn` and `minWorkerCount` added to `NewHandler` signature

#### Auth / Middleware
- [x] `atc/wrappa/api_auth_wrappa.go` — `GetHealth` confirmed in unauthenticated bucket (alongside `GetInfo`)
- [x] `atc/wrappa/reject_archived_wrappa.go` — `GetHealth` confirmed in non-pipeline passthrough case
- [x] `atc/auditor/auditor.go` — `GetHealth` confirmed under `EnableSystemAuditLog` (alongside `GetInfo`)

#### Configuration (`atc/atccmd/command.go`)
- [x] `--health-min-worker-count` CLI flag added (default `1`, group `Health Endpoint`)
- [x] `dbConn` threaded through `constructAPIHandler` → `NewHandler` → `healthserver.NewServer`

---

### Tests

- [x] `atc/api/health_test.go` — integration tests via the real HTTP test server
  - [x] DB healthy + workers meet threshold → 200, status `ok`
  - [x] Workers exactly at threshold → 200, status `ok`
  - [x] No running workers (stalled registered) → 503, status `failing`
  - [x] No workers at all → 503, status `failing`
  - [x] DB unreachable → 503, status `failing`, error message present
  - [x] DB read-only (standby replica) → 503, status `failing`, error contains "read-only"
  - [x] Worker factory returns error → 503, status `failing`
  - [x] Timestamp is recent
- [x] `atc/health_test.go` — DTO unit tests updated to match new struct shapes
- [x] `atc/api/api_suite_test.go` — `fakeDbConn` wired in
- [x] All 1046 existing API tests still pass

---

### Still To Do

- [x] Add other componenets, besides workers and DB, like scheduler, runtime, GC etc.
- [x] Integration test scenario: custom `minWorkerCount` value (e.g. `minWorkerCount=3`, 2 running → degraded)
- [x] `degraded` path exercised in a test (requires test suite to support non-default `minWorkerCount`)
- [x] Manual testing instructions / curl examples in PR description
- [ ] Packaging: documentation, bosh release, helm chart updates
- [ ] go-concourse client library: `go-concourse/concourse/health.go` + `client.go` interface + tests

---

### Open questions
#### Should we include the scheduler in the health check?

Right now the endpoint checks two things: can we talk to the database, and are there workers available. But there's a third signal we could add: is Concourse actually scheduling builds?

If the scheduler component has stopped running (e.g. it's stuck or stalled), the ATC looks healthy from the outside — database is up, workers are connected — but builds are silently not being picked up. An operator watching the health endpoint would have no idea.

We already track when the scheduler last ran in the database, so we could flag it as `degraded` if it hasn't run within a reasonable time window (e.g. more than 2× its configured interval). It would still return HTTP 200 so load balancers don't drop traffic, but the JSON would surface the problem.

The one concern worth discussing: every call to the health endpoint would do an extra database read to check scheduler state. Under aggressive probing (e.g. k8s checking every few seconds across many ATC instances) that adds load. We could mitigate it by caching the result for a few seconds.

### Proposition
#### Component Health Reporting

After reviewing `atc/component.go`, there are 23 registered components across three groups:

- **Runtime** (`scheduler`, `tracker`, `scanner`) — control whether builds start and run
- **Garbage Collection** (`reaper`, `collector_*`, `pipeline_pauser`) — clean up containers, volumes, resource caches etc.
- **System** (`signing_key_lifecycler`, `being_watched_build_marker`) — infrastructure maintenance

#### Decision

All components will be reported in the health response so operators have full visibility. However, only **runtime component stalls affect the overall status**.

| Group | Stale behaviour | HTTP impact |
|-------|----------------|-------------|
| Runtime (`scheduler`, `tracker`, `scanner`) | Promotes overall status to `degraded` | 200 with `"status": "degraded"` |
| GC + System | Visible in JSON body | No change to status or HTTP code |

A component is considered stale if it hasn't run within `staleMultiplier × interval` (default multiplier: `2.0`, configurable via `--health-component-stale-multiplier`). Paused components are excluded — pausing is an intentional operator action.

#### Rationale

A stale scheduler means builds silently stop being picked up — that is a meaningful degradation operators and load balancers should know about. A stale GC collector is a problem that accumulates over time but does not immediately break builds, so it should not cause noisy status changes for tooling that keys off the top-level status or HTTP code.

#### Example response (runtime component stale)

```json
{
  "status": "degraded",
  "database": { "status": "healthy" },
  "workers": { "status": "healthy", "total": 3, "running": 3 },
  "components": [
    { "name": "scheduler", "healthy": false, "paused": false, "last_ran": "2026-06-08T10:00:00Z", "stale": true },
    { "name": "tracker",   "healthy": true,  "paused": false, "last_ran": "2026-06-08T10:05:00Z", "stale": false },
    { "name": "collector_volumes", "healthy": false, "paused": false, "last_ran": "2026-06-07T08:00:00Z", "stale": true }
  ]
}
```

Regarding this:
> The one concern worth discussing: every call to the health endpoint would do an extra database read to check scheduler state. Under aggressive probing (e.g. k8s checking every few seconds across many ATC instances) that adds load. We could mitigate it by caching the result for a few seconds.

To my current design idea, I think we can use one single cheap query for all:

`ComponentFactory.All()` is a single `SELECT id, name, interval, last_ran, paused FROM components` - no joins, no subqueries, no aggregation. The `components` table has at most ~23 rows. It returns everything needed to compute staleness (`last_ran`, `interval`, `paused`) in one round trip.

So the full health check becomes exactly **3 queries total**:

| Query | Purpose |
|-------|---------|
| `SELECT pg_is_in_recovery()` | DB writability check |
| `SELECT ... FROM workers` | Worker count + states |
| `SELECT ... FROM components` | All 23 component states |

All three are already used elsewhere on hot paths - no joins, no aggregation. The staleness calculation (`time.Since(lastRan) > multiplier × interval`) happens in Go after the fetch, not in SQL.
The only scenario worth watching is high-frequency probing (e.g. k8s liveness every 2s × 10 ATC replicas ≈ 15 queries/second against the same 3 tables). Still cheap in absolute terms, but we can add a short in-memory cache later if metrics show it adding up.

taylorsilva:
_Just throwing my thoughts down as I'm skimming._

- Reading your "Pr Checklist" comment
  - "is Concourse actually scheduling builds?" - There are existing metrics that operators can watch that would tell them this.
    - That said, I'm not against having the granular component health checks. I think they could be useful info for operators.
  - When is the "last ran" time updated for components? Do we update it even if the component failed?
    - Looks like it doesn't update "last ran" if the component failed, so your setup should work: https://github.com/concourse/concourse/blob/22ccaf2a34385115db24c7e7e0568f1fae85ed0e/atc/component/coordinator.go#L64-L72
  - In the example response, `components[*]` each object has a `healthy` and `stale` field. The `stale` field feels redundant. Is there ever a time where `healthy` is true and `stale` is true?
  - All these tables are very small. I doubt we'd see performance issues stemming from any of these queries. Only one I'm not familiar with is `pg_is_in_recovery()`. Running a quick `EXPLAIN ANALYZE` I get `0.029ms` on the database backing ci.concourse-ci.org.
    - Caching the results of these queries adds complexity and all the problems of figuring out when to invalidate the cache. I'd skip adding any caching for now and wait for real world feedback from users.

_Haven't look at the code yet._

Kump3r:
>   * In the example response, `components[*]` each object has a `healthy` and `stale` field. The `stale` field feels redundant. Is there ever a time where `healthy` is true and `stale` is true?

Good catch. The original thinking was to keep `status` as a uniform field across all subsystems (database, workers, components) so clients only need one field to check health, while `stale` was meant to carry the *reason* - distinguishing "unhealthy because stale" from potential future failure modes like a component reporting its own error.

In practice that future hasn't arrived. Right now there is exactly one way a component becomes unhealthy: it goes stale. So `stale: true` and `status: "unhealthy"` are always identical, and `stale` adds noise without information. `last_ran` is already in the response so any client that wants to reason about staleness can do so themselves.

Dropping `stale`, keeping `status` with `healthy`/`unhealthy`. If we add a second failure mode later we can introduce a `reason` field at that point.

Also thanks for confirming the other implementation details and decisions so far!

Kump3r:
## Manual Testing

Connect to the database:

```bash
docker compose exec db psql -U dev concourse 
```

### Trigger overall `degraded` (runtime component stale)

```sql
UPDATE components SET last_ran = NOW() - INTERVAL '5 minutes' WHERE name = 'scheduler';
```

The scheduler is a runtime dependency — a stale scheduler promotes the top-level `status` to `"degraded"`.

### Trigger component-level `unhealthy` without affecting overall status (GC component stale)

```sql
UPDATE components SET last_ran = NOW() - INTERVAL '5 minutes' WHERE name = 'collector_volumes';
```

The collector stalling is surfaced in the `components` array (`"status": "unhealthy"`) but the top-level `status` stays `"ok"` — GC stalls don't break builds immediately.

### Trigger overall `failing` (no running workers)

```bash
docker compose stop worker
```

### Reset

```sql
UPDATE components SET last_ran = NOW() WHERE name IN ('scheduler', 'collector_volumes');
```

```bash
docker compose start worker
```

### Check the endpoint

```bash
curl -s http://localhost:8080/api/v1/health | jq .
```

Expected shape:

```json
{
  "status": "ok | degraded | failing",
  "timestamp": "...",
  "database": { "status": "healthy | unhealthy", "error": "..." },
  "workers":  { "status": "healthy | degraded | unhealthy", "total": 1, "running": 1 },
  "components": [
    { "name": "scheduler", "status": "healthy | unhealthy", "paused": false, "last_ran": "..." },
    ...
  ]
}
```

Kump3r:
Okay the overall idea got its shape, what is left to be done from my POV is:

- [ ] - Packaging: documentation, bosh release, helm chart updates
- [ ] - go-concourse client library: `go-concourse/concourse/health.go` + `client.go` interface + tests

Both of which I prefer to do after the review, as it doesn't make sense to invest time if stuff needs to be reworked based on that.

taylorsilva:
I will review this! Keep getting pulled off on side quests as you can probably tell 😅 I'll stop putting it off and review over the next few days.

atc/atccmd/command.go
```diff
@@ -204,6 +204,11 @@ type RunCommand struct {
 		ClientSecret            string `long:"client-secret" required:"true" description:"Client secret to use for login flow"`
 	} `group:"Web Server"`
 
+	Health struct {
+		MinWorkerCount         int     `long:"health-min-worker-count" default:"1" description:"Minimum number of running workers for the health endpoint to report healthy. Below this threshold the status is degraded (still 200). Zero workers is always failing."`
```
taylorsilva: nit: What is meant by "Zero workers is always failing"? Do you mean that setting this to a value of zero will result in the health endpoint always returning "failing"? or do you mean "regardless of the value here, zero connected workers will be regarded as failing"?

atc/health.go
```diff
@@ -0,0 +1,46 @@
...
+	HealthStatusOK       = "ok"
+	HealthStatusDegraded = "degraded"
+	HealthStatusFailing  = "failing"
+)
+
+// Per-subsystem status values — used in DatabaseHealth.Status, WorkerHealth.Status, ComponentHealth.Status.
+const (
+	HealthStatusHealthy   = "healthy"
+	HealthStatusUnhealthy = "unhealthy"
+)
```
taylorsilva: Instead of comments, make a type declaration and reference the type declaration in the above structs. Making the type private ensures this package retains control over the allowed values, similar to how an Enum would.

```suggestion
type systemStatus string

const (
	HealthStatusOK       systemStatus = "ok"
	HealthStatusDegraded systemStatus = "degraded"
	HealthStatusFailing  systemStatus = "failing"
)

type healthStatus string

const (
	HealthStatusHealthy   healthStatus = "healthy"
	HealthStatusUnhealthy healthStatus = "unhealthy"
)
```

Using Enums is possible here here, but then the resulting JSON will have int's instead of human-readable words, which makes it harder for users to use this endpoint. Not as self-documenting.

atc/health.go
```diff
@@ -0,0 +1,46 @@
...
+	Status string `json:"status"`
+	Error  string `json:"error,omitempty"`
+}
+
+// WorkerHealth represents the aggregate health of registered workers.
+type WorkerHealth struct {
+	Status  string `json:"status"`
+	Total   int    `json:"total"`
+	Running int    `json:"running"`
+}
```
taylorsilva, line 47: There was some debate about how this part should be structured. The information does feel limited in this format. I figure a health endpoint is useful if it can quickly surface the problem areas faster (_which_ workers are unhealthy?).

I know in the RFC I said we could approach this over multiple PR's, but now the maintainer in me is like "whatever you ship becomes the API people depend on!", and I want to try limiting the number of times we need to "break" things.

Detailed information about workers is available at `/api/v1/workers`, though that endpoint is limited by authentication and authorization checks (the full payload from the endpoint has a lot of details). Depending on your team/role, you'll see a subset of all workers connected to the cluster. Would be bad of us to backdoor a way to this info via an unauthenticated endpoint!

Maybe there's a middle ground here. We could list the names of any workers that are "unhealthy". So the json could look like:

```json
{
  ...
  "workers": {
    "status": "unhealthy",
    "total": 10,
    "running": 10,
    "unhealthy_workers": [
      "worker-03",
      "worker-04",
      "worker-08"
    ]
  }
}
```



I think the worker name is the minimal amount of information we could provide via this unauthenticated endpoint. WDYT?

```suggestion
// WorkerHealth represents the aggregate health of registered workers.
type WorkerHealth struct {
	Status           string   `json:"status"`
	Total            int      `json:"total"`
	Running          int      `json:"running"`
	UnhealthyWorkers []string `json:"unhealthy_workers,omitempty"`

}
```

atc/api/healthserver/health.go
```diff
@@ -0,0 +1,147 @@
...
+
+func countWorkers(workers []db.Worker, state db.WorkerState) int {
+	n := 0
+	for _, w := range workers {
+		if w.State() == state {
+			n++
+		}
+	}
+	return n
+}
```
taylorsilva: func isn't exported, why have a separate input for the state? This is only used in one place to count the number of workers with a state of `WorkerStateRunning`.


```suggestion
func countHealthyWorkers(workers []db.Worker) int {
	n := 0
	for _, w := range workers {
		if w.State() == WorkerStateRunning {
			n++
		}
	}
	return n
}
```

Wondering if we should count all workers that don't have a state of `stalled`? I wouldn't count workers that are `retiring/landing/landed` as unhealthy. Maybe this should be `w.State() != db.WorkerStateStalled`? WDYT?

atc/api/healthserver/health.go
```diff
@@ -0,0 +1,147 @@
...
+		health.Status = atc.HealthStatusFailing
+	case workerHealth.Status == atc.HealthStatusDegraded || runtimeDegraded:
+		health.Status = atc.HealthStatusDegraded
+	default:
+		health.Status = atc.HealthStatusOK
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	if health.Status == atc.HealthStatusFailing {
+		w.WriteHeader(http.StatusServiceUnavailable)
```
taylorsilva, line 45: Interesting, didn't think about what status code the endpoint should return!

Did some quick googling and seeing good arguments for both sides. Lots of semantics here.

On one side, folks say returning a 5xx is appropriate as a very simple way to convey to any system (i.e. loadbalancer) that the system is "down" in some sense. This also works for k8s HTTP probes, which only understand status codes: https://kubernetes.io/docs/concepts/workloads/pods/probes/#http-probes

On the other side, folks say HTTP status codes are meant to represent the status of the request being sent to the server. In this case, we're conflating "status of us completing this HTTP request" with "status of the Concourse system". This side seems more aligned with the spirit of the status code as defined in the RFC from IETF: https://datatracker.ietf.org/doc/html/rfc9110#name-503-service-unavailable 

> The 503 (Service Unavailable) status code indicates that the server is currently unable to handle the request due to a temporary overload or scheduled maintenance, which will likely be alleviated after some delay. The server MAY send a [Retry-After](https://datatracker.ietf.org/doc/html/rfc9110#field.retry-after) header field ([Section 10.2.3](https://datatracker.ietf.org/doc/html/rfc9110#field.retry-after)) to suggest an appropriate amount of time for the client to wait before retrying the request.

I think both make sense and which path we go down depends on who the consumer of this endpoint is.

Is this endpoint useful for health probes like the k8s one I linked to above? I don't think it is. If k8s was probing this endpoint it would be using it to determine if it needs to kill the web node container, which would be pointless if the issue is say "database connection" and the database is not reachable. No amount of web node restarts/recreations would resolve that.

I think a more intelligent consumer is the only valid consumer of this endpoint as we're currently describing it. I could imagine someone writing a [k8s operator](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/) to read this endpoint and take more intelligent action.

Looking at the RFC again, we had listed four initial motivations.

1. Monitoring and alerting - I assume the body payload is the most important bit to these systems. I think in this case it probably doesn't matter to the monitoring/alerting system what HTTP status code we return
2. Automation and self-healing - Depends _who_ is doing the self-healing. Right now this endpoint describes system-wide health, not the health of one component.
3. User experience - Similar as 1), they probably just want to read the body payload
4. Industry standards - Besides [k8s health endpoints](https://kubernetes.io/docs/reference/using-api/health-checks/) (`/healthz` is actually deprecated) I'm not sure what the rest of the industry generally does.

I guess overall it doesn't hurt returning HTTP 503 here. It's an easy way to quickly indicate to systems/users that something is wrong. We just need to make sure we have this documented that this endpoint **does not** represent the health of the individual web node fulfilling the request.

atc/api/healthserver/health.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if err := json.NewEncoder(w).Encode(health); err != nil {
+		logger.Error("failed-to-encode-health", err)
+	}
```
taylorsilva, line 51: We should probably print an error message to the body and change the HTTP status code to `HTTP 500`.

atc/api/api_suite_test.go
```diff
@@ -259,6 +261,50 @@ var _ = AfterEach(func() {
 	server.Close()
 })
 
+// buildTestServer constructs and starts a test HTTP server with the given minWorkerCount.
+// Use this in tests that need a non-default minWorkerCount (e.g. to exercise the degraded worker path).
+// Callers are responsible for closing the returned server.
+func buildTestServer(workerCount int) *httptest.Server {
```
taylorsilva: Having to do this feels like a hint that we've got the wrong abstraction somewhere.

Instead of this custom function, could we maybe add an interface, I think in this package `atc/api/`, and then the healthserver handler takes that interface as an input. The interface would wrap our two config values for the healthserver. Hmmm feels a bit odd too.

I don't have a better suggestion than that right now. We should try an eliminate this test func though. It feels like it goes against the patterns laid out in this. I do totally see why you added it as I'm not seeing an obvious alternative besides the interface path I mentioned.

taylorsilva (commented):
Reviewed most of this. Looking good so far! Let me know your thoughts on the things I called out (WDYT).

atc/atccmd/command.go
Kump3r: My ideas is if to have 3 statuses:
1. If healthy workers are above the minimum worker threshold
2. If the workers are below the threshold, but at least a single one is working, then status will be degraded as build scheduled will potentially be slower, due to lack of resources, but in the end still working
3. 3. If no working workers are found we fail, as there is nothing to process the requests

Setting this value to 0 practically means we would not have a degraded state. Either we have a single worker and it will report healthy, or no workers and the response will change to failed.

atc/health.go
Kump3r: For the initial draft I opted out against it, as even listing the workers endpoint requires permission and I didn’t want to give the information out in an unauthenticated endpoint, to not make a mistake, or bypass a valuable reason for not giving away workers information. With that being said, I am fine with implementing the failed workers list, or even each worker reporting its state. 2 questions come to mind:
1. since we will provide information like worker UID anyway, I assume it will not be a problem to try and add a reason/failure etc., which will tell the user whats going on with the specific worker, if there is such a reason.
2. I started thinking whether the worker endpoint list should also be unauthenticated, but even typing this a few red flags pop up, because as far as I remember there was other info like IP’s etc, that should still require the correct privileges. Feel free to skip this point, only leaving it here if you had some similar thoughts and we need to reiterate it.

atc/api/healthserver/health.go
Kump3r: I thought that if the user configures a minimum threshold and due to draining etc, the actual processing workers is bellow that threshold, the status should also degrade, so the operator double-checks the configuration such as max in flight/ in parallel strategies, to drain less workers at a time. Maybe they just didn’t notice this. For the method I agree. I refactored locally a couple of times and switched directions. It looks unnecessary currently so +1.

atc/api/api_suite_test.go
Kump3r: I agree with this, couldn’t think of anything cleaner when writing the tests. I might try to brainstorm some options and add them as comments, so we can decide on the best direction. I prefer tested and unclean then not adding a test 😅

atc/atccmd/command.go
taylorsilva: Gotcha! So setting to zero means we'll never enter the "degraded" state because of the number of workers, we'll only go from "healthy" to "failing".

atc/health.go
taylorsilva: The workers endpoint does have a lot of extra info about the worker. No secrets, but also no point in allowing anyone to see all that info anyways. I think the name of workers is okay to expose. Worker names aren't "secrets" and are usually arbitrary things like "worker-01". I think we should include this info. Without it, I don't think this endpoint is very useful to operators.

taylorsilva:
Read your replies and replied where necessary. Feel free to ping me when you've pushed changes to this again. No rush! :)

atc/atccmd/command.go
Kump3r: Okey, I reworded: setting to 0 skips the degraded state entirely - only healthy or failing.

atc/health.go
Kump3r: systemStatus and subsystemStatus are now unexported named string types - the compiler rejects invalid values instead of a comment doing it.

atc/health.go
Kump3r: Non-running worker names are now listed in the response so operators know exactly which workers are the problem.

Screenshot 2026-07-02 at 10 50 08

atc/api/healthserver/health.go
Kump3r: Inlined into checkWorkers - it was only ever called once and the loop now collects names and counts in a single pass.

atc/api/healthserver/health.go
Kump3r: Kept 503 - maybe it's the simplest signal for load balancers and k8s probes; the body explains the actual cause. Added a code note that this reflects system health, not the health of the individual web node serving the request, but maybe worth to add it somewhere to the docs as well.

atc/api/healthserver/health.go
Kump3r: http.Error is now called after the log line so the client gets an error body instead of a silent empty response.

atc/api/api_suite_test.go
Kump3r: Looked into this more sadly moving handler construction to `JustBeforeEach` (which runs after all `BeforeEach` blocks) would fix this, but roughly 20 test files in this suite make their HTTP requests inside `BeforeEach`, so that refactor would touch the whole suite and create a lot of noise. Based on this trade-off I suggest going with the helper method, albeit "unclean", until a global refactor is made. Wdyt?

Kump3r:
Added a note to each section with my 2 cents, I more or less tried addressing everything, leaving the test suit refactor as a hanging. Think it is a bit out of scope here and am not sure how to make it any better sadly. Sorry for that, if you have any concrete suggestions I will dive deeper into that, based on my current understanding and knowledge, I would have to refactor a lot of tests just to make this work and not sure whether that brings any value, besides code clarity :/

taylorsilva (approved):
I've reviewed the latest changes and they LGTM.

I pushed a commit resolving our issue with how the test setup worked. I used the options pattern to make it easier for us to introduce similar test changes in the future. Now all the setup code is consolidated into one set of functions instead of having duplicate funcs.

If someone wants to configure something in the `api.NewHandler()`, they just add fields to the `apiServerConfig` struct and new `apiConfigOpt` func for the new fields.

I added a few fields as examples. Didn't include `apiConfigOpt` funcs for them since we don't need to change them from the default. Also helped make those values more descriptive when reading the code.

Kump3r:
Thanks @taylorsilva for the support in RFC and in reviewing. I will address the follow-up packaging (helm/bosh/fly/client lib) in separate PR's.

PR #9593, opened 2026-06-08
Chown bind-mount volumes when running as non-root user

### LINKED ISSUE #403: When executing a task with non-root user, inputs should be owned by that user
We are using a Docker image with a non-root user (Dockerfile contains a "USER" command). All inputs (e.g. obtained by the git resource) are owned by root. Therefore, to make a commit we need to chown all files first. It seems like Concourse should do that for me when mounting the inputs.

taylorsilva:

closes #403

When a container runs as a non-root user, any input/output/cache volumes will be `chown`'d to the non-root user. This means every file and directory inside the volume will be owned by the non-root user, as well as the mount point the volume is mounted at inside the container. This will allow tasks running with non-root users to easily interact with volumes made from other steps that would previously remain owned by `root`. See the issue for workarounds users made. These workarounds should no longer be necessary after this change.

I'm labelling this as a **breaking change** since this is something Concourse will do automatically that may break existing workflows IF you're running tasks as a nonroot user. We can add an opt-out toggle in the future if deemed necessary. Based on #403 though, it _seems_ like most users would appreciate the change proposed in this PR.

For anyone that runs stuff as `root` all the time, things should continue to work as-is.

This does introduce a potential footgun for any tasks that output a volume.
```yaml
# this is a task config, I'm omitting most fields
platform: linux

outputs:
- name: some-output

run:
  user: nonroot
```

If the step(s) that follow this task run as `root`, they'll receive `some-output` unchanged. It won't be owned by root, it'll still be owned as `nonroot`. If the user doesn't exist in their container then they'll see the raw UID/GID of the `nonroot` user.

I made this pipeline to test some basic scenarios. All jobs pass when using this PR:

<details>
<summary>Click to expand and  view pipeline</summary>

```yaml
resources:
  - name: image
    type: mock
    source:
      mirror_self: true
    version:
      version: mock

  - name: version
    type: mock
    source:
      no_initial_version: true

jobs:
  - name: root
    plan:
      - get: image
      - task: root
        image: image
        config:
          platform: linux
          inputs:
            - name: image
          outputs:
            - name: empty
          run:
            path: bash
            args:
              - -c
              - |
                set -x
                whoami
                ls -lah .
                ls -lah /
                echo "hello world" > ./empty/file
                echo "hello image" > ./image/image

  - name: root-privileged
    plan:
      - get: image
      - task: root
        privileged: true
        image: image
        config:
          platform: linux
          inputs:
            - name: image
          outputs:
            - name: empty
          run:
            path: bash
            args:
              - -c
              - |
                set -x
                whoami
                ls -lah .
                ls -lah /
                echo "hello world" > ./empty/file
                echo "hello image" > ./image/image

  - name: nonroot
    plan:
      - get: image
      - task: nonroot
        image: image
        config:
          platform: linux
          inputs:
            - name: image
          outputs:
            - name: empty
          run:
            user: nonroot
            path: bash
            args:
              - -c
              - |
                set -x
                whoami
                ls -lah .
                ls -lah /
                echo "hello world" > ./empty/file
                echo "hello image" > ./image/image

  - name: nonroot-privileged
    plan:
      - get: image
      - task: nonroot
        privileged: true
        image: image
        config:
          platform: linux
          inputs:
            - name: image
          outputs:
            - name: empty
          run:
            user: nonroot
            path: bash
            args:
              - -c
              - |
                set -x
                whoami
                ls -lah .
                ls -lah /
                echo "hello world" > ./empty/file
                echo "hello image" > ./image/image

  - name: different-users
    plan:
      - get: image
      - task: nonroot
        image: image
        config:
          platform: linux
          outputs:
            - name: empty
          run:
            user: nonroot
            path: bash
            args:
              - -c
              - |
                set -x
                whoami
                ls -lah .
                echo "hello from nonroot" > ./empty/file

      - task: adm
        image: image
        config:
          platform: linux
          inputs:
            - name: empty
          outputs:
            - name: empty
          run:
            user: adm
            path: bash
            args:
              - -c
              - |
                set -x
                whoami
                ls -lah .
                ls -lah ./empty
                cat ./empty/file
                echo "hello from adm" > ./empty/file

      - task: daemon
        image: image
        config:
          platform: linux
          inputs:
            - name: empty
          run:
            user: daemon
            path: bash
            args:
              - -c
              - |
                set -x
                whoami
                ls -lah .
                ls -lah ./empty
                cat ./empty/file
                echo "hello from daemon" > ./empty/file

      - task: news
        image: image
        config:
          platform: linux
          inputs:
            - name: empty
          outputs:
            - name: empty
          run:
            user: news
            path: bash
            args:
              - -c
              - |
                set -x
                whoami
                ls -lah .
                ls -lah ./empty
                cat ./empty/file
                echo "hello from news" > ./empty/file

  - name: put
    plan:
      - get: image
      - task: nonroot
        image: image
        config:
          platform: linux
          outputs:
            - name: empty
          run:
            user: nonroot
            path: bash
            args:
              - -c
              - |
                set -x
                whoami
                ls -lah .
                ls -lah /
                echo "hello from nonroot" > ./empty/file
      - put: version
        params:
          file: empty/file

  - name: nonroot-to-root
    plan:
      - get: image
      - task: nonroot
        image: image
        config:
          platform: linux
          outputs:
            - name: empty
          run:
            user: nonroot
            path: bash
            args:
              - -c
              - |
                set -x
                whoami
                ls -lah .
                echo "hello from nonroot" > ./empty/file

      - task: root
        image: image
        config:
          platform: linux
          inputs:
            - name: empty
          run:
            user: root
            path: bash
            args:
              - -c
              - |
                set -x
                whoami
                ls -lah .
                ls -lah ./empty
                cat ./empty/file
                echo "hello from root" > ./empty/file
```

<details>

### DIFF
--- a/atc/db/volume_repository.go
+++ b/atc/db/volume_repository.go
@@ -52,6 +52,8 @@ type VolumeRepository interface {
 
 const noTeam = 0
 
+var _ VolumeRepository = (*volumeRepository)(nil)
+
 type volumeRepository struct {
 	conn DbConn
 }
--- a/atc/exec/task_step.go
+++ b/atc/exec/task_step.go
@@ -452,6 +452,7 @@ func (step *TaskStep) containerSpec(logger lager.Logger, state RunState, imageSp
 
 		Dir:      metadata.WorkingDirectory,
 		Hermetic: step.plan.Hermetic,
+		User:     config.Run.User,
 	}
 
 	var err error
--- a/atc/exec/task_step_test.go
+++ b/atc/exec/task_step_test.go
@@ -295,6 +295,16 @@ var _ = Describe("TaskStep", func() {
 			})
 		})
 
+		Context("when a user is configured", func() {
+			BeforeEach(func() {
+				taskPlan.Config.Run.User = "some-user"
+			})
+
+			It("sets the user on the container spec", func() {
+				Expect(chosenContainer.Spec.User).To(Equal("some-user"))
+			})
+		})
+
 		Context("when a timeout is configured", func() {
 			BeforeEach(func() {
 				taskPlan.Timeout = "1ms"
--- a/atc/runtime/types.go
+++ b/atc/runtime/types.go
@@ -126,6 +126,10 @@ type ContainerSpec struct {
 
 	// Hermetic indicates whether or not the container has external network access.
 	Hermetic bool
+
+	// User that bind mounts will be owned by and that processes will run under
+	// inside the container
+	User string
 }
 
 type BuildStepDelegate interface {
@@ -363,4 +367,11 @@ type VolumeMount struct {
 	// MountPath is the absolute path in the Container at which the Volume is
 	// mounted.
 	MountPath string
+}
+
+const ExistingOwner, ExistingGroup = -1, -1
+
+type VolumeOwnership struct {
+	Uid int
+	Gid int
 }
--- a/atc/worker/gardenruntime/container.go
+++ b/atc/worker/gardenruntime/container.go
@@ -18,6 +18,8 @@ const userPropertyName = "user"
 
 const exitStatusPropertyName = "concourse:exit-status"
 
+var _ runtime.Container = (*Container)(nil)
+
 type Container struct {
 	DBContainer_    db.CreatedContainer
 	GardenContainer gclient.Container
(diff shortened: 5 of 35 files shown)


taylorsilva:
I added an integration test under the `testflight` suite that's a more refined version of the pipeline in my PR description, testing for specific folder permissions and ability to write to various volumes.

PR #9596, opened 2026-06-10
Remove inline scripts and styles to enable strict CSP

Kump3r:
Move keyframe animations from inline <style> in index.html into animation.less so they are compiled into main.css at build time.

Replace the inline <script> that injected window.elmFlags with <meta> tags rendered server-side, and update elm-setup.js to read those values via document.querySelector. This eliminates 'unsafe-inline' from script-src, allowing a clean default-src 'self' CSP policy.

I can compile and add screenshots later, but there should be no functional changes, the idea is to allow Concourse to use CSP: `"frame-ancestors 'none'; default-src 'self'"` as strict security hardening, removing only the background_image and custom fonts section, but allowing concourse to continue to run normally

## Notes to reviewer
Steps to reproduce, add
```
CONCOURSE_CONTENT_SECURITY_POLICY: "frame-ancestors 'none'; default-src 'self'" 
```
to `docker-compose.yml` `web` section.

Before:

Screenshot 2026-06-10 at 18 19 57

After:
Screenshot 2026-06-10 at 18 22 38
The font is blocked as expected, with this strict policy, also pipeline images would be blocked.


Kump3r:
After demo:

https://github.com/user-attachments/assets/335b674d-8c2b-4553-a323-49d901d46179

Important: The error visible in the beginning during login, I addressed in the last commit, now there are no more JS errors and we can safely use: "frame-ancestors 'none'; default-src 'self'" to block scripts, images, fonts that we do not expect.

Kump3r:
Also tested the auth flow trough the CF connector in our environments and it works normally without any issue. There are no scripts there, just thought it is worth mentioning 😅

taylorsilva:
I went ahead and embedded the UnifrakturCook font within Concourse. I prefer having Concourse include everything that it needs to run the web UI.

ceco556 (approved):
Tested locally and it is working as expected. The only thing i didn't test is the CF connector, but i see that Kalin already tested it.

Great job!

PR #9600, opened 2026-06-12
feat: add validation for team names to prevent '/' characters in rename-team command

IvanChalukov:

Follow up of #9537
Teams should not be able to contain "/" in their names. This change handle rename-pipeline command.

* add validation for "/" in rename-team command
* extend tests for the command

## Notes to reviewer

Sample test could be executed by running:
```sh
fly -t dev set-team -n new/team --local-user=test
```

---

I've also test to rename teams and pipelines through API. Both operation succeeds and I was able to rename pipeline from `vault2` to `vault/2`, as well as team `ivan` to `ivan/2`. Sample calls:
- pipeline rename with `/`
  ```sh
  curl --request PUT \
    --url http://localhost:8080/api/v1/teams/ivan/pipelines/vault2/rename \
    --header 'authorization: Bearer <token>' \
    --header 'content-type: application/json' \
    --data '{
    "name": "vault/2"
  }'
  ```
  The response of that call contains only warning:
  ```sh
  {"warnings":[{"type":"invalid_identifier","message":"pipeline: 'vault/2' is not a valid identifier: illegal character '/'"}]}
  ```

- team rename with `/`:
  ```sh
  curl --request PUT \
    --url http://localhost:8080/api/v1/teams/ivan/rename \
    --header 'authorization: Bearer <token>' \
    --header 'content-type: application/json' \
    --data '{
    "name": "ivan/2"
  }'
  The response of that call contains only warning:
  
  ```sh
  {"warnings":[{"type":"invalid_identifier","message":"team: 'ivan/2' is not a valid identifier: illegal character '/'"}]}
  ```

However, I believe those exceptions would be handled in #9271

---

### DIFF
--- a/fly/commands/rename_team.go
+++ b/fly/commands/rename_team.go
@@ -2,6 +2,8 @@ package commands
 
 import (
 	"fmt"
+	"strings"
+	"errors"
 
 	"github.com/concourse/concourse/fly/commands/internal/displayhelpers"
 	"github.com/concourse/concourse/fly/commands/internal/flaghelpers"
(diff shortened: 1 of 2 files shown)


Kump3r:
Screenshot 2026-06-13 at 12 13 33

Works as expected!

Kump3r (approved):

PR #9601, opened 2026-06-15
Use os.Root when extracting files

### LINKED ISSUE #9597: PR #9486 breaks volumes containing symlinks on Windows
### Summary

_Thanks to @aliculPix4D for reporting this via email._

When streaming an input directory, such as a git or S3 resource, that contains a symlink (pointing to a file still inside the same input directory/git repository) to a Windows worker, Concourse fails with the following error:

```
find or create container on worker wrk-win2025-082c64ee436d0eb9c: failed to stream in to volume
```

Looking deeper into worker logs, one can identify that the root cause is a symlink file in the input volume.

```
{"level":"info","source":"worker","message":"worker.baggageclaim.api.volume-server.stream-in.bad-stream-payload","data":{"error":"entry './workers/[images-vars.tf](http://images-vars.tf/)' links outside of target directory: ../shared/[images-vars.tf](http://images-vars.tf/)"}}
```

### Difference Between PR Intent and Observed Platform Behavior

While the intent of PR #9486 was to secure symlink streaming for Windows and Darwin, the actual behavior is highly inconsistent across platforms—e.g on Windows it is simultaneously too strict in safe scenarios and too loose in risky ones:

- Windows (Inconsistent Security): While it blocks the safe intra-repo symlinks mentioned above, the fix completely ignores absolute paths in symlinks (e.g., C:\Windows\System32\drivers\etc\hosts). These are allowed to stream without any issue, meaning the fix is too loose where it actually matters.
- Darwin (Bypasses Fix): Contrary to what the PR states, streaming from Linux to Darwin bypasses these code changes entirely. Because tar is pre-installed and on the PATH in any fresh MacOS installation, Concourse invokes the tar executable directly. Consequently, the "security fix" does not apply, and Darwin ALLOWS anything. It only behaves like Windows if tar is missing.
- Linux to Linux (Bypasses Fix): As expected with the current implementation, Linux invokes the tar executable directly, bypassing the PR's code changes and ALLOWING anything.

Additional Observations

- Insufficient Build Logs: When a build errors out due to this streaming issue, the build logs do not provide any clue as to why the streaming failure occurred, leaving pipeline operators in the dark.
- Incomplete ATC/Worker Logs: The ATC/worker logs available to Concourse operators only report the first symlink issue. If an input contains multiple problematic symlinks, the logs hide the subsequent failures.

We realized that part of our confusion stems from the fact that commit https://github.com/concourse/concourse/pull/9486/changes/8b7a4131ce1b49eb41f53a3c3a70a9a3e688bcc7 considers symlinks NOT exiting from the root directory of the input volume a potential vulnerability.

### Steps to Reproduce

Stream any volume that contains a symlink to a Windows worker.

### Expected Results

- Safe Symlink Handling: Symlinks that point to a file inside the same input volume/directory (e.g., within the same git repository) are entirely safe and should stream without issue.

- Outside Symlinks: Absolute and relative symlinks that point outside the input directory should be blocked. In most cases, they will inherently be broken at the destination anyway. Therefore, blocking them seems reasonable regardless of their perceived security risk.

### Actual Results

In Concourse >=v8.1, a regression prevents valid inputs from streaming. This was introduced by https://github.com/concourse/concourse/pull/9486, which aimed to patch a tar security issue for Windows and Darwin workers.

Instead of streaming, the current implementation strictly blocks (what we consider) valid use cases. Specifically, safe intra-repo symlinks (pointing inside the input directory) are flagged as a security risk and fail to stream, resulting in a false positive. This happens when streaming to a Windows worker only.

We prepared a few demo examples (pipeline, build logs, input examples, etc.) in https://github.com/Pix4D/concourse-tar-streaming-issue for you to have a look.

### Web Node(s) configuration

_No response_

### Worker(s) configuration

_No response_

### Concourse Version

>= 8.1.1

### Browser (if applicable)

_No response_

### Did this use to work?

<= 8.1.0

taylorsilva:

closes #9597 

Basically reverting #9486, but adding one safe-guard against path escapes from symlinks when creating a relative symlink. [`os.Root`](https://pkg.go.dev/os#Root) is doing the heavy-lifting for us here. It'll protect us from extraction attacks that try to write outside the destination directory. It'll still allow us to create the symlink, but it'll fail if someone tries to do an extraction attack using the symlink.

I left a comment in the code path where we create symlinks explaining this a bit more.

I manually setup a pipeline using the test repo from Pix4D: https://github.com/Pix4D/concourse-tar-streaming-issue
```yaml
resources:
  - name: repo
    type: git
    source:
      uri: https://github.com/Pix4D/concourse-tar-streaming-issue.git

jobs:
  - name: job
    plan:
      - get: repo
      - task: windows
        config:
          platform: windows
          inputs:
            - name: repo
          run:
            path: powershell
            args:
              - -Command
              - |
                ls .
```

and streamed it from a linux to windows worker. I verified I could reproduce the issue from `master` first:
```
{"timestamp":"2026-06-15T16:14:50.724027700Z","level":"info","source":"worker","message":"worker.baggageclaim.api.volume-server.stream-in.bad-stream-payload","data":{"error":"entry './safe-intra-repo-windows/workers/images-vars.tf' links outside of target directory: ../shared/images-vars.tf","session":"2.2.1.11","volume":"bb0f01bf-1b35-4206-a0bc-9a40840081be"}}
```
image

Redeploying the windows worker with this PR, it worked:
image

### DIFF
--- a/go-archive/tarfs/extract.go
+++ b/go-archive/tarfs/extract.go
@@ -2,8 +2,10 @@ package tarfs
 
 import (
 	"archive/tar"
+	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -23,6 +25,12 @@ func Extract(src io.Reader, dest string) error {
 
 	chown := os.Getuid() == 0
 
+	root, err := os.OpenRoot(dest)
+	if err != nil {
+		return err
+	}
+	defer root.Close()
+
 	for {
 		hdr, err := tarReader.Next()
 		if err == io.EOF {
@@ -37,7 +45,7 @@ func Extract(src io.Reader, dest string) error {
 			continue
 		}
 
-		err = ExtractEntry(hdr, dest, tarReader, chown)
+		err = extractEntry(root, hdr, tarReader, chown)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


aliculPix4D:
@taylorsilva first impressions: it doesn't seem to fully resolve the issue for us.

This is all the logs I have so far:

from ATC:
```
atc.tracker.notify.run.task-step.find-or-create-cow-volume-for-container.failed-to-create-volume-in-baggageclaim: "error":"failed to create volume"
atc.tracker.notify.run.task-step.failed-to-create-volume-mounts-for-container: "error":"failed to create volume"
atc.tracker.notify.run.run.errored","data":{"build":"1","build_id":15318,"error":"find or create container on worker wrk-vs2019-win-0-0ae43318ab7cb4760: failed to create volume
```
From worker:
```
{"level":"error","source":"worker","message":"worker.baggageclaim.api.volume-server.create-volume-async.create-volume.failed-to-materialize-strategy","data":{"error":"exit status 9\n","handle":"4921c41c-495a-4aa9-ba1a-cdae80b0d8ac","privileged":false,"session":"2.2.1.1401.1","strategy":{"type":"cow","volume":"6731f982-9fb8-4051-a136-178d816bbdd8"}}}
{"level":"error","source":"worker","message":"worker.baggageclaim.api.volume-server.create-volume-async.failed-to-create","data":{"error":"exit status 9\n","handle":"4921c41c-495a-4aa9-ba1a-cdae80b0d8ac","privileged":false,"session":"2.2.1.1401","strategy":{"type":"cow","volume":"6731f982-9fb8-4051-a136-178d816bbdd8"}}}
```

the `exit status 9\n` doesn't give us any real clue to what the problem is.

What I noticed is the following:

- It seems to work fine on Windows Server 2025 AMI
- It consistently fails on Windows Server 2019 AMI

I will get back to you as soon as I have more information and confirmation that this is still related to the symlink issue.

**UPDATE:** this is definitely still related to symlinks. Doing a simple test, just by removing the symlinks from the repo, and everything passes.
**UPDATE2:** some extra notes:
- The exit code 9 is 1 (some files copied successfully) + 8 (failed to copy some files or directory)
- streaming works when symlinks are files, but not directories (both on 2019 and 2025)
- file symlinks are broken and unreadable  (both on 2019 and 2025)
- ~~there must be something related also to permissions:~~
```
      - task: check-privileges-<windows-os-version>
        tags: [win2025 or win20219]
        config:
          platform: windows
          run:
            path: cmd.exe
            args: ["/c", "whoami /priv"]
```
~~shows:~~

~~SeCreateSymbolicLinkPrivilege Create symbolic links     Disabled~~

- task: https://github.com/Pix4D/concourse-tar-streaming-issue/blob/fcd72d3832d94bbbd860b23e1cd803a88dde2c64/ci/pipeline.yml#L58
fails on both 2025 and 2019 when trying to read symlink to file

**UPDATE 3:** Thanks to @odormond who noticed that contrary to: `os.Symlink`, unfortunately, `root.Symlink()` doesn't really work properly on Windows (and backward and forward slashes). I will prepare small examples that show the exact problem.

taylorsilva:
The "exit status 9" indicates to me that we're in another part of the code where we are exec'ing some binary. The immediate spot that comes to mind is when we call `robocopy` when creating COW volumes, which appears to be what is happening here. My guess is that robocopy is having issues with copying the symlink? Maybe this goes back to the creation of the symlink that you mentioned in your "UPDATE 3"?

Are you still able to reproduce the issue with your example repo: https://github.com/Pix4D/concourse-tar-streaming-issue ?

aliculPix4D:
> Are you still able to reproduce the issue with your example repo: https://github.com/Pix4D/concourse-tar-streaming-issue ?

So your example in the PR description would reproduce the issue, if instead of `ls` you just tried to `cat` the file, like I do here:
https://github.com/Pix4D/concourse-tar-streaming-issue/blob/fcd72d3832d94bbbd860b23e1cd803a88dde2c64/ci/pipeline.yml#L58

Sure, in this case, it will successfully stream, but the created symlink will be unusable (due to forward slashes on Windows). 
We think that the root cause of the issue is `root.Symlink` which doesn't work properly on Windows if the symlink uses forward slash (everything works fine with backward slash).

This just manifested itself differently when you also have a directory symlink in the repository. In that case, it will fail during the streaming, as the error in my original post shows... but the root cause remains the same: root.Symlink (we think; otherwise there are really 2 issues)

Please allow me some time to prepare the examples.

taylorsilva:
Feel free to open a new issue when you have a final report.

aliculPix4D:
> Feel free to open a new issue when you have a final report.

Finally done in: https://github.com/concourse/concourse/issues/9609

It took a lot more time than expected :)

PR #9602, opened 2026-06-15
fix: split container property labels on rune boundaries

kurtmc:
Property values over containerd's 4096-byte label limit are split into multiple labels. The split used a raw byte offset, so a multi-byte UTF-8 rune straddling the boundary was bisected, leaving a chunk that is invalid UTF-8. containerd then rejects it ("set label: grpc: error while marshaling: string field contains invalid UTF-8"), erroring the build.

Back the chunk boundary off to the nearest rune start before slicing.

Discussion from closed git-resource PR: https://github.com/concourse/git-resource/pull/475


taylorsilva (approved):
LGTM - I verified that the test fails when commenting out the new code. Thanks for figuring this out and making a PR Kurt! I'm shipping 8.2.4 tomorrow so I'll get this in that release.

PR #9604, opened 2026-06-17
Add configurable stall timeout to auto-prune long-stalled workers

### LINKED ISSUE #9603: Add configurable stall timeout to auto-prune long-stalled workers (ephemeral is too aggressive for cache-heavy spot fleets)
## Summary

Add a configurable grace period after which **stalled** (non-ephemeral) workers are automatically pruned, e.g. `CONCOURSE_WORKER_STALL_TIMEOUT` / `--worker-stall-timeout`. Today operators on autoscaling/spot infrastructure are forced to choose between two unsatisfying options: the `ephemeral` flag (which deletes workers on a ~60–90s heartbeat gap and wipes their cache state) or leaving genuinely-dead workers in `stalled` forever and pruning them by hand / via external tooling.

## Current behavior

The GC worker collector (`atc/gc/worker_collector.go`) handles unresponsive workers in one of two ways depending on the `ephemeral` flag (`atc/db/worker_lifecycle.go`):

- `ephemeral = true` → `DeleteUnresponsiveEphemeralWorkers()`:
  `DELETE FROM workers WHERE ephemeral AND expires < NOW()`
- `ephemeral = false` → `StallUnresponsiveWorkers()`:
  `UPDATE workers SET state='stalled' WHERE state='running' AND expires < NOW()`

A worker's `expires` is set to `NOW() + ttl` on each heartbeat, where `ttl = heartbeat-interval * 2` (default heartbeat interval 30s → 60s TTL, `tsa/heartbeater.go`). The GC sweep runs every 30s. So a worker is eligible for stall/delete ~60–90s after its last successful heartbeat.

Crucially, there is **no path that ever deletes a non-ephemeral worker once it is `stalled`** — it lingers indefinitely until an operator runs `fly prune-worker`.

## The problem

On EC2 autoscaling groups using spot instances, workers regularly disappear *without* completing a graceful `retire`/`land` (hard spot reclamation, instance failure, etc.). Those workers go `stalled` and accumulate. Stalled workers are excluded from new placement (`atc/worker/pool.go` `isWorkerCompatibleAndRunning`), but their `worker_resource_caches` rows persist and can still be selected as streaming sources, adding build friction until they are pruned.

### Why `ephemeral` is not an acceptable answer for cache-heavy fleets

`ephemeral` deletes the worker row, and every worker-scoped table (`containers`, `volumes`, `worker_base_resource_types`, `worker_task_caches`, `worker_resource_caches`) is `ON DELETE CASCADE` on `workers(name)`. So deletion wipes the worker's entire DB cache footprint; baggageclaim then sweeps the now-"orphaned" on-disk volumes. For workers using `volume-locality` placement with large caches (hundreds of GB to TB), that is very expensive to rebuild.

The damaging part is a timing asymmetry:

- **Delete trigger:** ~60–90s after a heartbeat gap.
- **Worker reconnect detection:** up to **5 minutes** — the worker→TSA keepalive timeout is hardcoded to `time.Minute * 5` (`tsa/client.go`). On an *unclean* break (network partition, web/TSA node hang, TSA CPU saturation delaying heartbeat delivery) there's no TCP RST, so the worker doesn't notice the dead tunnel and re-dial for up to 5 minutes.

This means a perfectly healthy worker whose TSA connection is briefly disrupted — e.g. when a web/TSA node is rotated or restarted — can be **deleted and cache-wiped while it is still running builds**, only to reconnect moments later and re-register empty. In a fleet where the web nodes themselves autoscale/recycle, this happens routinely, producing worker "flapping" and cache thrash.

By contrast, a `stalled` worker recovers for free: `HeartbeatWorker` (`atc/db/worker_factory.go`) maps any non-landing/landed/retiring state back to `running`, so a transiently-disconnected worker returns to service with its container/volume/cache rows intact and zero re-streaming. The only thing missing is cleanup of workers that *never* come back.

## Proposed solution

Add a stall-prune grace period so the GC deletes workers that have been `stalled` longer than a configurable duration. Something like:

```
CONCOURSE_WORKER_STALL_TIMEOUT (e.g. default 0 = disabled / never, opt-in)
```

Implementation sketch: track when a worker entered `stalled` (a `stalled_since` / reuse of a timestamp column), and add a `DeleteStalledWorkers(timeout)` step to `workerCollector.Run` analogous to `DeleteUnresponsiveEphemeralWorkers`, deleting rows where `state='stalled' AND stalled_since < NOW() - timeout`.

This gives operators the cleanup behavior of `ephemeral` (no manual pruning, no indefinite accumulation) **without** the hair-trigger cache wipe, because the grace period (minutes/hours, operator's choice) comfortably exceeds the worst-case reconnect window. Transiently-disconnected workers recover; only genuinely-dead ones are reaped.

## Workaround we're using

In the meantime we run an external reconciliation reaper on the web nodes: a systemd timer that lists `stalled` workers via `fly workers --json`, cross-checks each against EC2 (the worker's hostname-derived private IP vs. the set of still-running worker instances), and only `fly prune-worker`s those whose backing instance is confirmed gone. This is safe but is bespoke per-cloud glue that we think the scheduler could express natively with a stall timeout.

Happy to contribute a PR if the maintainers are open to the approach — would appreciate guidance on the preferred config surface (a global `--worker-stall-timeout` vs. a per-worker setting) and on whether reusing `expires`/adding a `stalled_since` column is preferred for tracking stall age.

blarghmatey:

closes #9603

Adds an opt-in stall timeout so operators can have stalled (unresponsive,
non-ephemeral) workers automatically pruned after a configurable grace period,
without resorting to `ephemeral` and its aggressive cache wipe.

* **New config `CONCOURSE_STALLED_WORKER_TIMEOUT` / `--stalled-worker-timeout`**
  (default `0` = disabled, preserving today's behavior). When set, the GC
  worker collector deletes workers that have been `stalled` longer than the
  timeout.
* **New `stalled_since` column on `workers`** (migration
  `1781568000_add_stalled_since_to_workers`). `StallUnresponsiveWorkers` stamps
  it when a worker transitions `running` → `stalled`; `HeartbeatWorker` clears
  it when a worker reconnects, so a recovered worker resets its grace period and
  keeps all of its container/volume/cache rows. The migration backfills existing
  stalled workers with `now()` so they aren't immortal due to a `NULL` value.
* **`DeleteStalledWorkers(timeout)`** added to `WorkerLifecycle` and wired into
  `workerCollector.Run`, analogous to `DeleteUnresponsiveEphemeralWorkers`. It
  only runs when the timeout is greater than zero.

### Why

On EC2 autoscaling/spot fleets, workers regularly disappear without a graceful
`retire`/`land` and accumulate in `stalled` forever — nothing ever deletes a
non-ephemeral worker once stalled. The only existing automatic cleanup,
`ephemeral`, deletes a worker ~60–90s after a heartbeat gap and cascades away
its entire cache footprint. That's far too aggressive: the worker→TSA reconnect
window is hardcoded to up to 5 minutes (`tsa/client.go`), so on an unclean break
a healthy worker can be deleted and cache-wiped while still running builds, only
to reconnect moments later and re-register empty.

A grace-period-based stall timeout (minutes/hours, operator's choice) comfortably
exceeds the worst-case reconnect window: transiently-disconnected workers recover
for free, and only genuinely-dead workers are reaped. This matches the approach
agreed with @taylorsilva in
https://github.com/concourse/concourse/issues/9603#issuecomment-4723686750
(build on `atc/gc/worker_collector.go`, add a column to track stall age).

## Notes to reviewer

* Behavior is unchanged unless the operator opts in by setting a non-zero
  `--worker-stall-timeout`.
* New unit tests cover the collector (timeout disabled vs. set, error path) and
  the DB lifecycle (`DeleteStalledWorkers` respects the timeout, leaves running
  workers alone, and a heartbeat clears the stall before the timeout).
* The config is exposed as a top-level flag to match the
  `CONCOURSE_STALLED_WORKER_TIMEOUT` name requested in the issue; happy to move it
  under the `gc` namespace (`CONCOURSE_GC_WORKER_STALL_TIMEOUT`) if maintainers
  prefer it grouped with the other grace periods.

## Release Note

* Added `CONCOURSE_STALLED_WORKER_TIMEOUT` to automatically prune workers that have
  been stalled longer than the configured duration. Defaults to `0` (disabled).

### DIFF
--- a/atc/atccmd/command.go
+++ b/atc/atccmd/command.go
@@ -218,6 +218,8 @@ type RunCommand struct {
 		VarSourceRecyclePeriod time.Duration `long:"var-source-recycle-period" default:"5m" description:"Period after which to reap var_sources that are not used."`
 	} `group:"Garbage Collection" namespace:"gc"`
 
+	WorkerStallTimeout time.Duration `long:"worker-stall-timeout" default:"0" description:"Period after which stalled (unresponsive, non-ephemeral) workers will be automatically pruned along with their cache state. 0 (the default) means stalled workers are never automatically pruned and must be removed manually with 'fly prune-worker'."`
+
 	BuildTrackerInterval time.Duration `long:"build-tracker-interval" default:"10s" description:"Interval on which to run build tracking."`
 
 	TelemetryOptIn bool `long:"telemetry-opt-in" hidden:"true" description:"Enable anonymous concourse version reporting."`
@@ -1391,7 +1393,7 @@ func (cmd *RunCommand) gcComponents(
 
 	collectors := map[string]component.Runnable{
 		atc.ComponentCollectorBuilds:            gc.NewBuildCollector(dbBuildFactory),
-		atc.ComponentCollectorWorkers:           gc.NewWorkerCollector(dbWorkerLifecycle),
+		atc.ComponentCollectorWorkers:           gc.NewWorkerCollector(dbWorkerLifecycle, cmd.WorkerStallTimeout),
 		atc.ComponentCollectorResourceConfigs:   gc.NewResourceConfigCollector(dbResourceConfigFactory, unreferencedConfigGracePeriod),
 		atc.ComponentCollectorResourceCaches:    gc.NewResourceCacheCollector(dbResourceCacheLifecycle),
 		atc.ComponentCollectorTaskCaches:        gc.NewTaskCacheCollector(dbTaskCacheLifecycle),
--- a/atc/db/migration/migrations/1781568000_add_stalled_since_to_workers.down.sql
+++ b/atc/db/migration/migrations/1781568000_add_stalled_since_to_workers.down.sql
@@ -1,0 +1,1 @@
+ALTER TABLE workers DROP COLUMN stalled_since;
--- a/atc/db/migration/migrations/1781568000_add_stalled_since_to_workers.up.sql
+++ b/atc/db/migration/migrations/1781568000_add_stalled_since_to_workers.up.sql
@@ -1,0 +1,6 @@
+ALTER TABLE workers ADD COLUMN stalled_since timestamp with time zone;
+
+-- Backfill existing stalled workers so they are subject to the stall timeout
+-- starting from the moment this migration runs, rather than being immortal due
+-- to a NULL stalled_since.
+UPDATE workers SET stalled_since = now() WHERE state = 'stalled';
--- a/atc/db/worker_lifecycle.go
+++ b/atc/db/worker_lifecycle.go
@@ -2,6 +2,8 @@ package db
 
 import (
 	"database/sql"
+	"fmt"
+	"time"
 
 	sq "github.com/Masterminds/squirrel"
 )
@@ -10,6 +12,7 @@ import (
 type WorkerLifecycle interface {
 	DeleteUnresponsiveEphemeralWorkers() ([]string, error)
 	StallUnresponsiveWorkers() ([]string, error)
+	DeleteStalledWorkers(timeout time.Duration) ([]string, error)
 	LandFinishedLandingWorkers() ([]string, error)
 	DeleteFinishedRetiringWorkers() ([]string, error)
 	GetWorkerStateByName() (map[string]WorkerState, error)
@@ -47,11 +50,38 @@ func (lifecycle *workerLifecycle) DeleteUnresponsiveEphemeralWorkers() ([]string
 func (lifecycle *workerLifecycle) StallUnresponsiveWorkers() ([]string, error) {
 	query, args, err := psql.Update("workers").
 		SetMap(map[string]any{
-			"state":   string(WorkerStateStalled),
-			"expires": nil,
+			"state":         string(WorkerStateStalled),
+			"expires":       nil,
+			"stalled_since": sq.Expr("NOW()"),
 		}).
 		Where(sq.Eq{"state": string(WorkerStateRunning)}).
 		Where(sq.Expr("expires < NOW()")).
+		Suffix("RETURNING name").
+		ToSql()
+	if err != nil {
+		return []string{}, err
+	}
+
+	rows, err := lifecycle.conn.Query(query, args...)
+	if err != nil {
+		return nil, err
+	}
+
+	return workersAffected(rows)
+}
+
+// DeleteStalledWorkers deletes workers that have been in the 'stalled' state
+// for longer than the given timeout. This gives operators the automatic
+// cleanup of ephemeral workers without the hair-trigger cache wipe: a
+// transiently-disconnected worker that reconnects within the timeout returns
+// to 'running' (clearing stalled_since) and keeps its caches, while genuinely
+// dead workers are eventually reaped.
+func (lifecycle *workerLifecycle) DeleteStalledWorkers(timeout time.Duration) ([]string, error) {
+	query, args, err := psql.Delete("workers").
+		Where(sq.Eq{"state": string(WorkerStateStalled)}).
+		Where(sq.Expr(
+			fmt.Sprintf("stalled_since < NOW() - '%d second'::INTERVAL", int(timeout.Seconds())),
+		)).
 		Suffix("RETURNING name").
 		ToSql()
 	if err != nil {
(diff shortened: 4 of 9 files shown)


atc/db/migration/migrations/1781568000_add_stalled_since_to_workers.up.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE workers ADD COLUMN stalled_since timestamp with time zone;
+
+-- Backfill existing stalled workers so they are subject to the stall timeout
+-- starting from the moment this migration runs, rather than being immortal due
+-- to a NULL stalled_since.
+UPDATE workers SET stalled_since = now() WHERE state = 'stalled';
```
taylorsilva, line 6: Is fine, but not sure how they would be "immortal". Wouldn't the Worker GC component set the `stalled_since` once it runs? Guess I'll find out as I review this.

atc/db/migration/migrations/1781568000_add_stalled_since_to_workers.down.sql
taylorsilva, line 1: Appreciating how the LLM knew the migration needed to be the "latest" but obviously didn't do a `date +%s`. Highly doubt it got a unix timestamp with 3 trailing zeros!

Just to make sure it didn't use a timestamp far into the future and messes up future migrations, I checked the current result of `date +%s` on my machine and got `1781713555`, which is after this migrations timestamp. We're all good here!

atc/db/worker_lifecycle.go
```diff
@@ -47,8 +50,9 @@ func (lifecycle *workerLifecycle) DeleteUnresponsiveEphemeralWorkers() ([]string
 func (lifecycle *workerLifecycle) StallUnresponsiveWorkers() ([]string, error) {
 	query, args, err := psql.Update("workers").
 		SetMap(map[string]any{
-			"state":   string(WorkerStateStalled),
-			"expires": nil,
+			"state":         string(WorkerStateStalled),
+			"expires":       nil,
+			"stalled_since": sq.Expr("NOW()"),
```
taylorsilva, line 55: Ah I see, so we set `stalled_since` when the worker is marked as stalled, not in the GC component. This seems fine then, previous comment in the migration can be ignored.

atc/atccmd/command.go
```diff
@@ -218,6 +218,8 @@ type RunCommand struct {
 		VarSourceRecyclePeriod time.Duration `long:"var-source-recycle-period" default:"5m" description:"Period after which to reap var_sources that are not used."`
 	} `group:"Garbage Collection" namespace:"gc"`
 
+	WorkerStallTimeout time.Duration `long:"worker-stall-timeout" default:"0" description:"Period after which stalled (unresponsive, non-ephemeral) workers will be automatically pruned along with their cache state. 0 (the default) means stalled workers are never automatically pruned and must be removed manually with 'fly prune-worker'."`
```
taylorsilva, line 221:
```suggestion
	WorkerStallTimeout time.Duration `long:"worker-stall-timeout" default:"0s" description:"Period after which stalled (unresponsive, non-ephemeral) workers will be automatically pruned along with their cache state. 0s (the default) means stalled workers are never automatically pruned and must be removed manually with 'fly prune-worker'."`
```

Because it's type `time.Duration` we should indicate it accepts a duration format by having the default value be `0s`. The other `time.Duration` flags do this.

taylorsilva:
@blarghmatey Can you complete the EasyCLA check please? TY!

atc/atccmd/command.go
taylorsilva: I wonder if the name of the flag here will cause issues when using the `quickstart` command. That command namespaces all the worker commands from `CONCOURSE_` to `CONCOURSE_WORKER_`. The web commands are not namespaced like this and are left as-is.

Ah easy fix, name it `StalledWorkerTimeout`!

taylorsilva (commented):
Solid PR overall. Only minor nits, but I can resolve those after merging. Once your complete the EasyCLA I can merge this. TY!

blarghmatey:
Thanks for the quick review! I'm happy to address the nita if you prefer, otherwise it should be clear to merge as I've signed the CLA

taylorsilva:
Thanks! I've got a commit with the changes I want, so no worries, I'll fix my nits on main :)

taylorsilva:
I've update the PR description with the newer flag name `CONCOURSE_STALLED_WORKER_TIMEOUT`/`--stalled-worker-timeout`.

PR #9607, opened 2026-06-18
Make aborting jobs faster when the worker disappears

taylorsilva:

This is covering the case where a build is running and the worker it's running on disappears. In this case the web node would eventually error after ~5mins due to the GardenRequest timeout. If a user tries to abort a build in such a state before the GardenRequest timeout, nothing will happen as we would get stuck waiting for GardenContainer.Stop() to hit its own 5min timeout.

We wrap these calls in a shorter timeout of 10s. In this scenario we know the user has called abort, so a shorter timeout makes sense to override GardenRequest's longer timeout of 5mins.

There is an integration tests that I've added which reproduces the abort state I observed on `master`.

### DIFF
--- a/atc/worker/gardenruntime/container.go
+++ b/atc/worker/gardenruntime/container.go
@@ -56,12 +56,18 @@ func (c Container) Run(_ context.Context, spec runtime.ProcessSpec, io runtime.P
 	if err != nil {
 		return nil, fmt.Errorf("get properties: %w", err)
 	}
-	// Using context.Background() since GardenContainer.Run uses the passed in
-	// ctx to stream the process output. Since we send a SIGTERM on ctx
-	// cancellation, using the same ctx for streaming results in those logs not
-	// showing up.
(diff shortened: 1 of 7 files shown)


taylorsilva:
one testflight test failing but only with the guardian runtime:
```
  [FAILED] Got stuck at:
      checking tf-pipeline-3-3e85c7f0-43fd-4a1f-a54e-f2750d4932c8/my-resource in build 138
      initializing check: my-resource
      selected worker: 8edb88aa7572
      failed
      
  Waiting for:
      timeout exceeded
  In [It] at: github.com/concourse/concourse/testflight/resource_check_timeout_test.go:35
```

On containerd I get this output from that fly command:
```
initializing check: my-resource
selected worker: ci-workers-worker-1
timeout exceeded
failed
```

taylorsilva:
Weird, I can't locally reproduce this. I get the "timeout exceeded" string with both runtimes...

taylorsilva:
Yup just some weirdness in CI. Couldn't reproduce it locally and starting a fresh build instead of a re-run worked.

PR #9611, opened 2026-06-19
add --force flag to fly's abort-build command

taylorsilva:

**Users should NOT use this flag unless they've tried regular aborting first and it has failed to abort for more than 1min.**

When the API handler sees this flag set it'll go ahead and mark the build as aborted in the database. It still sends the abort command to the atc/engine running the build as usual, so container's in-flight still get interrupted and `on_*` hook steps still run.

This is meant to help resolve some edge-cases I've seen where builds are stuck "running", even though the worker(s) they were running on are very much dead and gone. In these situations the build is stuck in the running state and pressing abort does nothing. I've not been able to reproduce these cases (I have seen them myself two times over the years), but they are very frustrating as it feels like Concourse is just "hanging" dealing with these "zombie" builds. This `--force` will help users kill off these zombie builds.

This flag should only be used as a last resort. I did some testing to see what happens to normal in-flight builds. `on_*` hook steps still fire, but streaming logs does get a bit messed up if you look at them with `fly watch`. You won't get any logs after the "abort". If you look in the web UI you'll see any logs that happened after "abort". Not horrible and probably preferable to having the build stuck in a running state.

Visually here's what happens.

**Build aborted normally**
image

**Build aborted with `--force`**
image

You can see the "interrupted" build event isn't logged. If you do `fly watch` on the `--force`'d build, you won't see any logs from the `on_*` hook steps either. I did confirm that in-flight containers are still interrupted though, so still gets you the end result desired.

I've added a warning message in the `--help` text for `fly abort-build` to urge users to do a regular abort first before trying again with `--force`.

### DIFF
--- a/atc/api/builds_test.go
+++ b/atc/api/builds_test.go
@@ -1027,12 +1027,22 @@ var _ = Describe("Builds API", func() {
 	Describe("PUT /api/v1/builds/:build_id/abort", func() {
 		var (
 			response *http.Response
+			force    bool
 		)
+
+		BeforeEach(func() {
+			force = false
+		})
 
 		JustBeforeEach(func() {
 			var err error
 
-			req, err := http.NewRequest("PUT", server.URL+"/api/v1/builds/128/abort", nil)
+			url := server.URL + "/api/v1/builds/128/abort"
+			if force {
+				url += "?force="
+			}
+
+			req, err := http.NewRequest("PUT", url, nil)
 			Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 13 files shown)


Kump3r:
We experienced the same issues, when workers became overloaded and stalled eventually needing a recreate to be unblocked. The web is still trying to track containers/processes that are no longer there. The current way we mitigated it, was by restarting the web. So we would highly appreciate a feature such as this cc: @IvanChalukov

Issue #9062, opened 2025-02-11
Formalize Shared Secrets Support across all supported managers


kcbimonte:
## Summary
Shared Secrets have gotten partial support depending on the Secrets Managers that people prefer. In the latest couple of releases, SSM has been added to that list that already included Vault and AWS Secrets Manager. However, the remaining Secrets Managers, most notably CredHub, have not had the functionality extended to them. 

This is a Backlog item request to add the Shared Secrets Path functionality to the remaining secrets managers so that usage can be consistent across the secrets managers when possible. The current list is the following:
- CredHub
- Kubernetes
- Conjur

## Context

* RFC: concourse/rfcs# N/A
* Prior discussion: #8687, #3221, #6226
* Depends on # N/A
* Part of # N/A

VikranthBala:
Hey @taylorsilva, I'd like to take a crack at this if it's still up for grabs. I'm fairly new to the Concourse codebase, so I'd like to start by scoping just the CredHub piece first, get a PR up, and then look at Kubernetes/Conjur as follow-ups once I've got a feel for the pattern.

VikranthBala:
Hey @taylorsilva I have added my PR for credhub, do let me know if it looks fine or if you have any comments

taylorsilva:
Thanks @VikranthBala! All contributions are much appreciated. I'll take a look at the PR this week. Got a few others I need to look at first. It's on my radar :)

VikranthBala:
Hi @taylorsilva, I have added my PR for shared path support for Kubernetes. Can you please review it & let me know if any changes are required. Thank you in Advance :D

taylorsilva:
Final PR made and merged that completes this issue.

PR #9615, opened 2026-06-23
add shared path support to credhub `CONCOURSE_CREDHUB_SHARED_PATH`

VikranthBala:

As of now, it is not possible to configure shared paths for Credhub cred manager as opposed to other cred managers such as Vault,AWS. So I propose to add a new parameter --credhub-shared-path similar to the Vault cred manager. This parameter can be optional (as with Vault). This PR provides all the changes needed for this functionality.

### Notes to reviewer

Release Note

    Added `--credhub-shared-path` to configure shared secret paths for Credhub cred manager similarly to the one for Vault, SSM.

PR is for the issue: #9062

### DIFF
--- a/atc/creds/credhub/credhub.go
+++ b/atc/creds/credhub/credhub.go
@@ -11,9 +11,10 @@ import (
 )
 
 type CredHubAtc struct {
-	CredHub *LazyCredhub
-	logger  lager.Logger
-	prefix  string
+	CredHub    *LazyCredhub
+	logger     lager.Logger
+	prefix     string
+	sharedPath string
 }
 
 // NewSecretLookupPaths defines how variables will be searched in the underlying secret manager
@@ -25,6 +26,9 @@ func (c CredHubAtc) NewSecretLookupPaths(teamName string, pipelineName string, a
 	lookupPaths = append(lookupPaths, creds.NewSecretLookupWithPrefix(path.Join(c.prefix, teamName)+"/"))
 	if allowRootPath {
 		lookupPaths = append(lookupPaths, creds.NewSecretLookupWithPrefix(c.prefix+"/"))
+	}
+	if c.sharedPath != "" {
+		lookupPaths = append(lookupPaths, creds.NewSecretLookupWithPrefix(c.sharedPath+"/"))
 	}
 	return lookupPaths
 }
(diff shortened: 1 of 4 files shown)


atc/creds/credhub/credhub.go
```diff
@@ -26,6 +27,9 @@ func (c CredHubAtc) NewSecretLookupPaths(teamName string, pipelineName string, a
 	if allowRootPath {
 		lookupPaths = append(lookupPaths, creds.NewSecretLookupWithPrefix(c.prefix+"/"))
 	}
+	if c.sharedPath != "" {
+		lookupPaths = append(lookupPaths, creds.NewSecretLookupWithPrefix(c.sharedPath+"/"))
+	}
```
taylorsilva, line 32: I think this should also include the `c.prefix`. I think the credhub semantics are similar to Vault's. Both var sources have a `PathPrefix` that is, as per their descriptions:

> Path under which to namespace credential lookup.

Which in Vault means all secret lookups happen under the `PathPrefix`. See https://github.com/concourse/concourse/blob/6d8aee76225529b1727dcf39e148a042ad61328b/atc/creds/vault/vault.go#L43-L45

Contrasting that, we don't do that in the SSM cred manager: https://github.com/concourse/concourse/blob/6d8aee76225529b1727dcf39e148a042ad61328b/atc/creds/ssm/ssm.go#L49-L51
but that's because it has no `PathPrefix`: https://github.com/concourse/concourse/blob/6d8aee76225529b1727dcf39e148a042ad61328b/atc/creds/ssm/manager.go#L22-L31

atc/creds/credhub/manager.go
```diff
@@ -18,6 +18,8 @@ type CredHubManager struct {
 
 	PathPrefix string `long:"path-prefix" default:"/concourse" description:"Path under which to namespace credential lookup."`
 
+	SharedPath string `long:"shared-path" description:"CredHub parameter path used for shared parameters"`
```
taylorsilva: nit: change description to match Vault's
```suggestion
	SharedPath string `long:"shared-path" description:"Path under which to namespace credential lookup."`
```

taylorsilva (changes requested):
This looks good! Change the lookup path logic for it and then this looks good to merge.

VikranthBala:
Thanks for the review, I will have a look at the comments & update the PR

taylorsilva (approved):
LGTM - Thank you @VikranthBala

PR #9618, opened 2026-06-24
Add `CONCOURSE_CUSTOM_HTTP_HEADERS` allowing operators to set multiple headers on all HTTP responses

AtanasNikov:
**Changes proposed by this PR:**

Adds support for configuring additional baseline security HTTP headers in Concourse by configuring `CONCOURSE_ADDITIONAL_HTTP_HEADERS`. When setting with the env var, each Header+Value is separated by a comma, like so: `CONCOURSE_ADDITIONAL_HTTP_HEADERS: "Referrer-Policy:no-referrer,Cross-Origin-Opener-Policy:same-origin"`

Unit tests are included to verify headers are set correctly and omitted when empty.

The change was validated on a dev Concourse deployment with a custom build:

- Verified headers via curl
- Validated browser behavior using DevTools (Network + Console)
- Tested cross-origin scenarios (navigation, popups, resource loading)
- Confirmed no regressions in UI/API flows

***We will open PRs for `concourse-bosh-release`, `concourse-chart` and `docs` as well.***


Kump3r:
Unit test failed with unrelated errors, I am retriggering

ceco556:
Yes and it still failling.
There is an issue opened #9619.

Kump3r:
Could you guys sync master in your fork and merge it to your feature branch, so the unit test pass? Thanks!

taylorsilva:
Should be no need rebase, we do that automatically with the PR resource. You just hit the resource cache when you re-ran it. I re-triggered it now and it's doing a fresh rebase in CI now.

taylorsilva (commented):
This PR has me wondering if we need to change our approach here. I feel like instead of endlessly adding new flags for HTTP headers, we should have some way for operators to configure a set of headers that are then attached to the security handler here. This also ensures operators aren't gated on a new Concourse release when all they want to do is add a header.

I know this probably isn't how you were thinking this would go when you made this PR, but I think I'd prefer we re-evaluate how we approach these configurations flags. I'm not keen on adding the same kind of flags over and over again.

Are you open to exploring an alternative approach to this? It doesn't have to happen in this PR. We can open an issue and figure out the details in there and then create a new PR. WDYT?

Kump3r:
Since they are more or less set the same, what you have in mind is remove the different params and let operators provide an array/map with the headers. I think this is actually a reasonable way forward and agree if I understand it correctly. Seems simple enough to refactor, the problem would be deprecating the old properties and notifying operators of the change. Or am I missing something?

taylorsilva:
Yeah you mostly got what I'm saying. I wouldn't remove the existing flags, still use them. But yes to adding a way for operators to pass in an array of HTTP headers + values.

ceco556:
We worked together with @AtanasNikov and changed it to be an array of HTTP headers. 

Tested locally and it works as expected:

 `docker-compose.yml`:
`CONCOURSE_HTTP_HEADERS: "Referrer-Policy:no-referrer,Cross-Origin-Opener-Policy:same-origin"`

```
> curl -skI http://localhost:8080/

HTTP/1.1 200 OK
Cache-Control: no-store, private
Content-Security-Policy: frame-ancestors 'none'
Cross-Origin-Opener-Policy: same-origin
Referrer-Policy: no-referrer
X-Download-Options: noopen
X-Frame-Options: deny
Date: Tue, 30 Jun 2026 17:31:54 GMT
Content-Length: 1413
Content-Type: text/html; charset=utf-8
```

taylorsilva:
Just need @AtanasNikov to do the EasyCLA check please.

Integration test actually passed, but Github gave us a 401 back when trying to update the PR status, which triggered the `on_failure` hook step which DIDN'T get a 401 error. Flakey Github responses >_<
Running again, should pass.

taylorsilva (dismissed):
Added a "nit" commit on top. I preferred the flag name matching the struct field's name, so made it `CONCOURSE_ADDITIONAL_HTTP_HEADERS`. Also makes it clear that we set it _in addition_ to what we also set.

I will also say that I recognize for some HTTP headers, it's common to set them to a value with commas in them `value1,value2`. For this case, users HAVE to use the flag version `--additional-http-headers`, setting it once for each header they want to add.

Otherwise this LGTM. TY for changing direction here! Once the EasyCLA is signed off we can merge this.

IvanChalukov (commented):
Hey folks, I really like the idea of having a property that allows adding additional HTTP headers.

That being said, I do have a concern about using a comma as the delimiter here:

CONCOURSE_ADDITIONAL_HTTP_HEADERS: "Referrer-Policy:no-referrer,Cross-Origin-Opener-Policy:same-origin"

Several HTTP headers legitimately contain comma-separated values (e.g. Cache-Control, Accept-encoding, etc.), so using “,” as a separator makes the format ambiguous and could limit which headers can be configured in the future. Would it make sense to use a less ambiguous format or delimiter?

Also, I assume this property can override predefined headers such as CSP. If that’s the case, it could be a source of confusion, so I think the documentation should be expanded to clarify the behavior and any precedence rules.

taylorsilva:
Good points Ivan. Current code does make it possible to override the "legacy" header flags (CSP,Strict-Transport-Security, and X-Frame-Options) but funnily enough, it can't override these three: https://github.com/concourse/concourse/blob/0a51dc47799f1eea3e694b7a10c5bac3a816815b/atc/wrappa/security_handler.go#L22-L24

We should probably figure out the intention here. How do we expect all these flags to work together?

---

Another thought just occurred to me as well. Couldn't someone put a load balancer in front of their web nodes and have the load balancer add the extra headers for them?

It's nice to have Concourse have some safe defaults out of the box, but these feel like "I'm really going out of my way to _really_ lock things down" headers that most people won't set. Is there something I'm missing though that makes it more practical to have Concourse set these headers? The original set of headers were:

- `Cross-Origin-Opener-Policy`
- `Cross-Origin-Resource-Policy`
- `Cross-Origin-Embedder-Policy`

AtanasNikov:
Nice catch.

We are introducing a new --additional-http-headers flag that accepts a JSON object, allowing operators to configure arbitrary HTTP response headers. We chose JSON object because it correctly handles header values containing commas (for example, Cache-Control: no-cache, private), which wasn't working with go-flags' built-in delimiter parsing.

We are applying the AdditionalHTTPHeaders map after all existing header assignments in ServeHTTP, including the three currently hardcoded headers. This allows operators to override the currently default headers(X-Content-Type-Options, X-Download-Options, Cache-Control) if they explicitly choose to.
If --additional-http-headers is not specified, Concourse continues to behave exactly as it is and uses the existing default values.

Operators who want to use these or any other response headers can opt in by configuring --additional-http-headers, while everyone else continues to get the existing behavior unchanged.

atc/http_headers_flag.go
Kump3r, line 1: Why are we introducing this as a separate module and not add it to the [security_handler.go](https://github.com/concourse/concourse/pull/9618/changes/a981a692926c9d2b6b6801a2124941bb66d9e6e0#diff-2e9b10dd95a158bbe82ebef5d8b340233cd53de1e028e3296b09c4f7fc2ca43c). Also there are no tests for this atm.

Kump3r: since we are introducing validatio, should we also validate that this additional headers do not have XFrameOptions, ContentSecurityPolicy, StrictTransportSecurity as those should be configured trough the dedicated properties to not have confusion?

Kump3r (commented):
Some thoughts based on the latest commits from my side.

atc/http_headers_flag.go
Kump3r: seems like other such flags are validated in the command.go directly, so you should follow that pattern. Also when you change that you wouldn't need a dedicated structure and can directly let operators pass map[string]string which would then have some validations.

taylorsilva (dismissed):
We discussed this PR during the Working Group meeting today and decided on suggesting the following changes for this PR:

- Rename flag/env var to `--custom-http-headers`/`CONCOURSE_CUSTOM_HTTP_HEADERS` and have the behaviour be to override any previously set headers. That means having it be the last one to set headers in the Security handler
- The flag should take in a path to a JSON/YAML file that we then parse. We realized there are a lot of random characters that can go into the values of headers which makes it very difficult to parse via flags/env vars. Parsing headers from a file will be safer and easier for operators to manage.

Example of headers file:
```json
{
  "header-1": "some; value; string;",
  "header-2": "some, value, string"
}
```
```yaml
header-1: "some; value; string;"
header-2: "some, value, string"
```

AtanasNikov:
We've contributed with @IvanChalukov on replacing `--additional-http-headers` with `--custom-http-headers` (CONCOURSE_CUSTOM_HTTP_HEADERS), which now accepts a path to a YAML or JSON file instead of an inline JSON string, making it safer and easier for operators to manage header values containing special characters.

Following the flag.AuthorizedKeys pattern, a new CustomHTTPHeaders type parses and validates the file at startup via UnmarshalFlag, applying RFC 7230 validation using golang.org/x/net/http/httpguts. 

Custom headers are applied last in SecurityHandler, allowing them to override any previously set header.

atc/http_headers_flag.go
```diff
@@ -0,0 +1,18 @@
+package atc
```
IvanChalukov: [http_headers_flag.go‎](https://github.com/concourse/concourse/pull/9618/changes#diff-4892490b18f83b6d5f5f21ce7fc90303d1e0ce390961ccbcba7f67e3c9012c17) is no longer needed. Please delete it.

atc/atccmd/command.go
```diff
@@ -1921,13 +1920,19 @@ func (cmd *RunCommand) constructHTTPHandler(
 	webMux.Handle("/.well-known/", apiHandler)
 	webMux.Handle("/", webHandler)
 
+	var additionalHTTPHeaders atc.HTTPHeadersFlag
```
IvanChalukov: that validation is no longer needed with CustomHTTPHeaders implementation.

atc/wrappa/security_handler_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 
+	"github.com/concourse/concourse/atc"
```
IvanChalukov: import is not used. remove it.

IvanChalukov (changes requested):

IvanChalukov (approved):
Looks good to me! I've tested it locally with both YAML and JSON files:
- JSON sample:
```json
{
  "X-Content-Type-Options": "some; value; string;",
  "Referrer-Policy": "no-referrer",
  "Cross-Origin-Opener-Policy": "same-origin",
  "Cache-Control": "no-store",
  "dummy-header": "123asd"
}
```
- YAML sample:
```yaml
X-Content-Type-Options: "some; value; string;"
Referrer-Policy: "referrer"
Cross-Origin-Opener-Policy: "same-orig"
Cache-Control: "private,max-age=10"
```

It also fails to start web when try to set invalid header like `Cache- Control: "private,max-age=10"`(uses whitespaces instead of `-`):
```err
error: invalid argument for flag `--custom-http-headers' 
(expected flag.CustomHTTPHeaders): invalid header name "Cache- Control" in custom HTTP headers file (/headers.yaml)
```

PR #9624, opened 2026-06-30
Prevent wall banner from overlapping page content when scrolling

### LINKED ISSUE #9623: Broadcast message breaks pipeline/job names in build view
### Summary

The Broadcast message UI feature, sometimes breaks the view of the pipeline/job name in the build view:

Image

### Steps to Reproduce

1. Create a long broadcast message
2. Open the build view

### Expected Results

I see the name of the pipeline and job

### Actual Results

The broadcast message covers the names

### Web Node(s) configuration

_No response_

### Worker(s) configuration

_No response_

### Concourse Version

8.2.4

### Browser (if applicable)

_No response_

### Did this use to work?

_No response_

ceco556:

closes #9623 

The wall banner (`#wall-banner`) is using `position: sticky`, which requires a scrolling ancestor to work correctly. The banner sits outside the scroll container and because of that `position: sticky` had no effect. At the end the banner scrolled with the page content and overlapped job names and team panel.

## Notes to reviewer

* `Application.elm` - `#wall-banner` and `.wall-editor-bar` moved inside `#page-wrapper`, which now has `display: flex; flex-direction: column`. The banner is a natural flex item above the page content.

## Release Note

Prevent wall banner from overlapping page content when scrolling.

test_broadcast_message_fix

### DIFF
--- a/web/elm/src/Application/Application.elm
+++ b/web/elm/src/Application/Application.elm
@@ -582,24 +582,12 @@ view model =
             , SideBar.tooltip model.session
                 |> Maybe.map (Tooltip.view model.session)
                 |> Maybe.withDefault (Html.text "")
-            , case model.wallMessage of
-                Just msg ->
-                    if model.wallEditor.isOpen then
-                        Html.text ""
-
-                    else
-                        Html.div [ id "wall-banner", style "white-space" "pre-wrap" ] <|
-                            wallLinks msg
-
-                Nothing ->
-                    Html.text ""
-            , if model.wallEditor.isOpen then
-                wallEditorView model.wallEditor model.wallMessage model.session.hovered
-
-              else
-                Html.text ""
             , Html.div
-                ([ id "page-wrapper", style "height" "100%" ]
+                ([ id "page-wrapper"
+                 , style "height" "100%"
+                 , style "display" "flex"
+                 , style "flex-direction" "column"
+                 ]
                     ++ (if model.session.draggingSideBar then
                             Styles.disableInteraction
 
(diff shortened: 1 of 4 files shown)


web/public/elm-setup.js
```diff
@@ -20,6 +20,23 @@ const app = Elm.Main.init({
...
+    const bannerHeight = banner ? banner.offsetHeight : 0;
+    page.style.paddingTop = (NAV_HEIGHT + bannerHeight) + 'px';
+  };
+
+  const observer = new MutationObserver(applyWallOffset);
+  observer.observe(document.body, { childList: true, subtree: true, attributes: true, attributeFilter: ['style'] });
+  window.addEventListener('resize', applyWallOffset);
+  applyWallOffset();
+})();
+
```
taylorsilva: This feels like a hack. Is there no way for us to do this with just CSS? Maybe we need to move things around a bit more in the layout, but I'd prefer a CSS fix for this instead of including this anonymous function that may break if we make other layout changes in the future and don't notice because we don't always test how the page layout looks when the banner is set.

taylorsilva (changes requested):

web/public/elm-setup.js
ceco556: Sorry for this, yesterday i tried to move things around but couldn't find a solution so decided to use an observer to check when the wall is set. I didn't like it too, but it was a easy fix using it.

I did some research today and found a way easier fix. Thank you for stopping the PR and again sorry for the approach. 

> Note: Updated the description of the PR with the changes.

taylorsilva (changes requested):
Code looks good, thanks for changing it up. Sadly I'm not seeing a difference in behaviour here. Tried on Chrome and Firefox, same results on both:

https://github.com/user-attachments/assets/4710b972-ad9b-4854-a6ba-74e341e9d136

I'm using this pipeline here in the video: https://github.com/concourse/concourse/blob/master/testflight/fixtures/hooks.yml

ceco556:
Nice catch!
Thank you for the video, it was way easier to reproduce it. 

Fixed in the latest commit.

taylorsilva (approved):
Added a nit commit to update a test and fix a small formatting thing.

Otherwise LGTM - Tested with a few of the testflight pipelines and everything looks good now! No extra scroll bar is introduced when the banner is set.

I made sure to test a few other pages as well to ensure things don't look wonky. Looked at:
- Resource versions page
- Builds page of a job
- Resource causality page
- Download fly page

Issue #9605, opened 2026-06-17
Make transition from pipeline to pipeline using instance variables possible


benjaminguttmann-avtq:
### Summary

Make it possible to transition a pipeline that is currently not using instances variables to a pipeline using instance variables via `fly rename-pipeline`.

Like that you could keep your task/ run history when you are refactoring your pipelines to use instance variables.

benjaminguttmann-avtq:
Would be interested to work on it

benjaminguttmann-avtq:
First question that comes to my mind would be, how should fly rename-pipeline be adjusted?

For example should it allow using `/` in the pipeline name and then extract the values after the `/` to instance variables or should the instance variables be a dedicated cli parameter, .e.g you have oldName, newName, and oldInstanceVariable,NewInstanceVariable?

taylorsilva:
I'm pretty sure there are some helper functions in the fly code that can be used to parse the instance vars for you.

There's the `PipelineFlag` type: https://github.com/concourse/concourse/blob/b4c4b7e2bc2c82076eb2ea9bd9aa107ee44a7fea/fly/commands/internal/flaghelpers/pipeline_flag.go

You can use that type here instead of `string`: https://github.com/concourse/concourse/blob/b4c4b7e2bc2c82076eb2ea9bd9aa107ee44a7fea/fly/commands/rename_pipeline.go#L14-L15

It may also be useful to add a `--instance-var` flag like we have in `set-pipeline`: https://github.com/concourse/concourse/blob/b4c4b7e2bc2c82076eb2ea9bd9aa107ee44a7fea/fly/commands/set_pipeline.go#L30

Then someone could rename a pipeline in two ways ([docs for syntax](https://concourse-ci.org/docs/pipelines/grouping-pipelines/#managing-instanced-pipelines)):
```
fly rename-pipeline \
  --old-name pipeline \
  --new-name pipeline/var1:value1,var2:value2
```
or
```
fly rename-pipeline \
  --old-name pipeline \
  --new-name pipeline \
  --instance-var var1=value1 --instance-var var2=value2
```

Could even rename an existing instance pipeline.
```
fly rename-pipeline \
  --old-name pipeline/var1:old1 \
  --new-name pipeline/var1:new1
```

Actually, instead of one `--instance-var` flag, maybe there should be two? `--old-instance-var` and `--new-instance-var`? That might make it clearer for the user how to rename instanced pipelines.

[...]

PR #9629, opened 2026-07-06
atc/db: Introduce possibility to rename pipelines from non instance…

benjaminguttmann-avtq:
… variables to instance_variables

Introduced DB part of the functionality to rename pipeline using non instance_variable to a pipeline using instance_variables (#9605)

### DIFF
--- a/atc/db/team.go
+++ b/atc/db/team.go
@@ -667,12 +667,43 @@ func (t *team) SavePipeline(
 }
 
 func (t *team) RenamePipeline(oldName, newName string) (bool, error) {
-	result, err := psql.Update("pipelines").
-		Set("name", newName).
+	return t.renamePipeline(atc.PipelineRef{Name: oldName}, atc.PipelineRef{Name: newName})
+}
+
+func (t *team) renamePipeline(oldRef, newRef atc.PipelineRef) (bool, error) {
+	query := psql.Update("pipelines").
+		Set("name", newRef.Name).
 		Where(sq.Eq{
 			"team_id": t.id,
-			"name":    oldName,
-		}).
+			"name":    oldRef.Name,
+		})
+
+	if oldRef.InstanceVars != nil {
+		oldInstanceVarsBytes, err := json.Marshal(oldRef.InstanceVars)
+		if err != nil {
+			return false, err
+		}
+
+		query = query.Where(sq.Eq{
+			"instance_vars": sql.NullString{String: string(oldInstanceVarsBytes), Valid: true},
+		})
+	}
+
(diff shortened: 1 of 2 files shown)


atc/db/team.go
```diff
@@ -667,12 +667,43 @@ func (t *team) SavePipeline(
 }
 
 func (t *team) RenamePipeline(oldName, newName string) (bool, error) {
-	result, err := psql.Update("pipelines").
-		Set("name", newName).
+	return t.renamePipeline(atc.PipelineRef{Name: oldName}, atc.PipelineRef{Name: newName})
+}
+
+func (t *team) renamePipeline(oldRef, newRef atc.PipelineRef) (bool, error) {
```
taylorsilva, line 673: I think it would be fine to update the public function's signature to just be the private's. Though I suppose doing so would immediately force you to include code changes for everything else in the stack (api, fly, go-concourse, etc.).

You know, I think I'll prefer seeing the entire change in one PR. I have a decent sense what the shape of it will be so I don't think it'll be terribly challenging for me to review.

atc/db/team.go
```diff
@@ -667,12 +667,43 @@ func (t *team) SavePipeline(
...
+			return false, err
+		}
+
+		query = query.Where(sq.Eq{
+			"instance_vars": sql.NullString{String: string(oldInstanceVarsBytes), Valid: true},
+		})
+	}
+
+	if oldRef.InstanceVars != nil || newRef.InstanceVars != nil {
+		newInstanceVars := sql.NullString{Valid: false}
```
taylorsilva, line 693: nit: style thing here, in Go we don't usually initialize structs with their zero values. Another way of writing this would be `newInstanceVars := sql.NullString{}`
```suggestion
		var newInstanceVars sql.NullString
```

Similar to this code in the same file: https://github.com/concourse/concourse/blob/b0a1474ca97a6fad9a0aca1285570c764d357320/atc/db/team.go#L420-L433

benjaminguttmann-avtq:
@taylorsilva finally were able to pick this up again. So my local tests worked fine beside the different test runs

taylorsilva:
One of the `atc/db` tests are still failing: https://ci.concourse-ci.org/teams/contributor/pipelines/pr/jobs/unit/builds/2?vars.number=%229629%22#L6ab24524:261:266

```
  [FAILED] Expected
      <atc.InstanceVars | len:1>: {
          "version": <string>"6.7.x",
      }
  to be nil
  In [It] at: github.com/concourse/concourse/atc/db/team_test.go:4028
```

PR #9634, opened 2026-07-08
fix: Hermetic doesn't configure hosts on the containers

### LINKED ISSUE #9633: Hermetic containers stall on DNS timeouts instead of failing fast
### Summary

When a task runs with `hermetic: true`, Concourse blocks outbound traffic via iptables DROP rules. However, the container still gets a fully populated `/etc/resolv.conf`. Any process that performs a DNS lookup will send UDP packets that silently hit the DROP rules and wait for the full resolver timeout (typically 5 seconds per attempt × retries × number of nameservers) before the lookup fails.

Root cause:
`SetupMounts` in `worker/runtime/cni_network.go` always writes a populated `resolv.conf` regardless of hermeticity. `DropContainerTraffic` adds the DROP rules only after the container is already configured with nameservers to query.

### Steps to Reproduce

1. docker compose up (with `containerd`) runtime
2. Create the following pipeline:
```
jobs:
  - name: hermetic-dns-timeout
    plan:
      - task: dns-lookup
        hermetic: true
        config:
          platform: linux
          image_resource:
            type: registry-image
            source:
              repository: alpine
          run:
            path: sh
            args:
              - -c
              - |
                echo "Starting DNS lookup..."
                time nslookup google.com
                echo "Done."
```
3. Trigger the job and observe how long nslookup example.com takes to return

### Expected Results

DNS lookups in a hermetic container should fail immediately. Since all outbound traffic is intentionally blocked, there is no point sending DNS packets at all.

### Actual Results

Tools that do DNS lookups at startup (e.g. package managers, SDKs, manifest validators) can take 30–40 minutes to begin actual work, spending all that time exhausting DNS retry cycles against a silent firewall.

### Web Node(s) configuration

_No response_

### Worker(s) configuration

CONCOURSE_RUNTIME: containerd

### Concourse Version

v8.2.4

### Browser (if applicable)

_No response_

### Did this use to work?

no

Kump3r:

closes #9633 

* `network.go`- added `hermetic` bool parameter to the `SetupMounts` interface method
* `cni_network.go` - when `hermetic` is true, write an empty `/etc/resolv.conf` instead of copying nameservers from the host
* `backend.go` - compute hermetic once and pass it through to `SetupMounts`
* `runtimefakes/fake_network.go` - updated generated fake to match the new signature
* `cni_network_test.go` - updated all existing calls to pass `hermetic: false`; added new test asserting that a hermetic container gets an empty `resolv.conf` even when nameservers are configured on the worker

### Without fix:
Screenshot 2026-07-08 at 13 29 19

### With fix:
Screenshot 2026-07-08 at 13 34 27

### DIFF
--- a/worker/runtime/backend.go
+++ b/worker/runtime/backend.go
@@ -309,12 +309,14 @@ func (b *GardenBackend) Ping() (err error) {
 func (b *GardenBackend) Create(gdnSpec garden.ContainerSpec) (garden.Container, error) {
 	ctx := context.Background()
 
-	cont, err := b.createContainer(ctx, gdnSpec)
+	hermetic := b.isHermetic(gdnSpec)
+
+	cont, err := b.createContainer(ctx, gdnSpec, hermetic)
 	if err != nil {
 		return nil, fmt.Errorf("new container: %w", err)
 	}
 
-	err = b.startTask(ctx, cont, b.isHermetic(gdnSpec))
+	err = b.startTask(ctx, cont, hermetic)
 	if err != nil {
 		return nil, fmt.Errorf("starting task: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


taylorsilva (approved):
LGTM - Good quality of life change for tasks that use `hermetic`.

PR #9638, opened 2026-07-09
skymarshal: behaviour: handle RFC 8693 token-exchange responses

### LINKED ISSUE #9636: RFC 8693 token exchange against embedded Dex always returns an empty 500 — StoreAccessToken middleware assumes every token response contains id_token
### Summary

Concourse's embedded Dex supports the RFC 8693 token-exchange grant
(urn:ietf:params:oauth:grant-type:token-exchange) and advertises it in
/sky/issuer/.well-known/openid-configuration. The exchange itself succeeds —
Dex verifies the subject token against the upstream connector and mints a
token — but the client always receives HTTP 500 with an empty body, and the
minted token is discarded.

Root cause: skymarshal/token/access_token.go (StoreAccessToken) intercepts
every 2xx response from /sky/issuer/token, unmarshals the JSON, and parses
the id_token field to convert it into a stored opaque access token. Per
RFC 8693 section 2.2.1, a token-exchange response carries the issued token in
access_token and has no id_token member (Dex's handleTokenExchange
puts the token in access_token for both requested_token_type values and
never sets IDToken). The middleware therefore calls
claimsParser.ParseClaims(""), which fails, and the error path writes a bare
500 without a body, dropping Dex's successful response.

Proposed fix: when id_token is absent from Dex's response, parse the claims
from access_token instead (it is the same RS256-signed Dex JWT in the
exchange case). This converts exchanged tokens into stored opaque Concourse
access tokens exactly like every other grant. I have a small candidate patch
for skymarshal/token/access_token.go and am happy to open a PR.

### Steps to Reproduce

The bug is in Concourse's post-processing of the token endpoint response, so
it reproduces with any OIDC connector for which the token exchange itself
succeeds (i.e. any connector whose TokenIdentity accepts the subject token).
We hit it with Microsoft Entra; a locally run Concourse (docker compose up)
pointing its oidc connector at any IdP reproduces it the same way:

docker compose up -d a Concourse whose web node has an OIDC connector

configured (CONCOURSE_OIDC_ISSUER etc. pointing at your IdP), with
 CONCOURSE_OIDC_DISABLE_GET_USER_INFO=true so that id_token subject
 verification is done locally.

Obtain a valid token from the upstream IdP whose issuer/signature Dex can

verify (for Entra: an access token with the Concourse app registration as
 audience and a v2.0 issuer).

Perform the exchange against Concourse's token endpoint:

 curl -i http://localhost:8080/sky/issuer/token \
   -u 'fly:Zmx5' \
   --data-urlencode 'grant_type=urn:ietf:params:oauth:grant-type:token-exchange' \
   --data-urlencode 'connector_id=oidc' \
   --data-urlencode 'subject_token_type=urn:ietf:params:oauth:token-type:id_token' \
   --data-urlencode "subject_token=${UPSTREAM_TOKEN}" \
   --data-urlencode 'requested_token_type=urn:ietf:params:oauth:token-type:access_token' \
   --data-urlencode 'scope=openid profile email federated:id groups'

Observe the response and the web node logs.

### Expected Results

A 200 response containing a Concourse-usable token, as for any other grant
against /sky/issuer/token — i.e. the RFC 8693 response rewritten to carry a
stored opaque Concourse access token, the same post-processing applied to
authorization-code and password grants.

### Actual Results

HTTP 500 with an empty body, on every attempt, for both
requested_token_type values. The web node logs (with the timestamp matching
the request):

{"source":"atc","message":"atc.dex-server.token-request.parse-id-token",
 "data":{"error":"go-jose/go-jose: compact JWS format must have three parts","session":"..."}}

Dex itself succeeded: the subject token was verified and a token was minted
(no atc.dex error precedes the middleware error). The middleware discards
that successful response. Requesting
requested_token_type=urn:ietf:params:oauth:token-type:id_token does not
help — per RFC 8693 the issued ID token still arrives in access_token, and
the id_token field the middleware reads remains absent.

### Web Node(s) configuration

One web node (deployed via the Concourse Helm chart 19.0.2). Relevant auth
configuration, secrets omitted:

CONCOURSE_OIDC_DISPLAY_NAME: Azure Entra
CONCOURSE_OIDC_ISSUER: https://login.microsoftonline.com/<tenant-id>/v2.0
CONCOURSE_OIDC_SCOPE: openid email profile
CONCOURSE_OIDC_USER_NAME_KEY: preferred_username
CONCOURSE_OIDC_GROUPS_KEY: roles
CONCOURSE_OIDC_SKIP_EMAIL_VERIFIED_VALIDATION: "true"
CONCOURSE_OIDC_DISABLE_GET_USER_INFO: "true"

### Worker(s) configuration

Three workers (CONCOURSE_RUNTIME: containerd,
CONCOURSE_BAGGAGECLAIM_DRIVER: overlay). Not relevant to this bug — the
failure is entirely in the web node's token endpoint middleware.

### Concourse Version

7.14.x (Helm chart 19.0.2). The affected code in skymarshal/token/access_token.go is unchanged on current master.

### Browser (if applicable)

N/A — API-only, reproducible with curl.

### Did this use to work?

No. StoreAccessToken predates Dex's token-exchange support (inherited when the Dex fork was rebased), and it has always assumed an id_token field in every successful token response, so this grant has never worked against the embedded Dex on any Concourse version.

ibaileyStingray:

closes #9636

* Dex's token-exchange grant returns the issued token in access_token and, per RFC 8693 section 2.2.1, has no id_token member. StoreAccessToken assumed every 2xx from /sky/issuer/token carried an id_token, so it parsed an empty string, failed, and converted Dex's successful response into an empty 500 - making token exchange against the embedded Dex unusable. Fall back to parsing claims from access_token so exchanged tokens are converted to stored Concourse access tokens like every other grant.

## Notes to reviewer

- **Why rewrite rather than pass through?** Simply skipping the middleware for
  the exchange grant would hand the client a raw Dex JWT, which the ATC no
  longer accepts (Concourse >= 7 validates opaque DB-backed tokens). Rewriting
  keeps exchange clients on the same token model as `fly` login, including the
  user row creation (`CreateOrUpdateUser`) and the short-header benefit that
  motivated opaque tokens in the first place.
- **Why not gate on `grant_type` from the request?** The response-side check
  (`id_token` empty) is more robust — the request body has already been
  consumed by the Dex handler, and any 2xx response lacking `id_token`
  previously produced a guaranteed 500, so the fallback is strictly an
  improvement.
- The two struct additions are cosmetic but keep the rewritten response
  honest: `omitempty` stops an empty `"id_token": ""` appearing in exchange
  responses, and `issued_token_type` is preserved instead of silently dropped
  on re-marshal.
- A test belongs in `skymarshal/token/access_token_test.go`: feed the
  middleware a fake Dex 200 with `access_token` = signed JWT and no
  `id_token`, assert a 200 with an opaque token and stored claims.

### DIFF
--- a/skymarshal/token/access_token.go
+++ b/skymarshal/token/access_token.go
@@ -70,11 +70,12 @@ func StoreAccessToken(
 			return
 		}
 		var resp struct {
-			AccessToken  string `json:"access_token"`
-			TokenType    string `json:"token_type"`
-			ExpiresIn    int    `json:"expires_in"`
-			RefreshToken string `json:"refresh_token,omitempty"`
-			IDToken      string `json:"id_token"`
+			AccessToken     string `json:"access_token"`
+			TokenType       string `json:"token_type"`
+			ExpiresIn       int    `json:"expires_in"`
+			RefreshToken    string `json:"refresh_token,omitempty"`
+			IDToken         string `json:"id_token,omitempty"`
+			IssuedTokenType string `json:"issued_token_type,omitempty"`
 		}
 		err := json.Unmarshal(rec.Body.Bytes(), &resp)
 		if err != nil {
@@ -82,7 +83,11 @@ func StoreAccessToken(
 			w.WriteHeader(http.StatusInternalServerError)
 			return
 		}
-		claims, err := claimsParser.ParseClaims(resp.IDToken)
+		rawIDToken := resp.IDToken
+		if rawIDToken == "" {
+			rawIDToken = resp.AccessToken
+		}
+		claims, err := claimsParser.ParseClaims(rawIDToken)
 		if err != nil {
 			logger.Error("parse-id-token", err)
 			w.WriteHeader(http.StatusInternalServerError)
--- a/skymarshal/token/access_token_test.go
+++ b/skymarshal/token/access_token_test.go
@@ -79,6 +79,16 @@ var _ = Describe("Access Tokens", func() {
 				expectBody:       `{"access_token":"123abc","token_type":"bearer","expires_in":1234,"id_token":"a.b.c"}`,
 			},
 			{
+				it: "converts RFC 8693 token-exchange responses, which have no id_token",
+
+				path:       "/sky/issuer/token",
+				statusCode: 200,
+				body:       `{"access_token":"a.b.c","issued_token_type":"urn:ietf:params:oauth:token-type:access_token","token_type":"bearer","expires_in":1234}`,
+
+				expectStatusCode: 200,
+				expectBody:       `{"access_token":"123abc","token_type":"bearer","expires_in":1234,"issued_token_type":"urn:ietf:params:oauth:token-type:access_token"}`,
+			},
+			{
 				it: "forwards failure response",
 
 				path:       "/sky/issuer/token",
@@ -168,6 +178,33 @@ var _ = Describe("Access Tokens", func() {
 				Expect(rec.Body.String()).To(Equal(t.expectBody))
 			})
 		}
+
+		Describe("choosing the token to parse claims from", func() {
+			var serveTokenResponse = func(body string) {
+				baseHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					w.WriteHeader(200)
+					w.Write([]byte(body))
+				})
+				handler := token.StoreAccessToken(dummyLogger, baseHandler, generator, claimsParser, accessTokenFactory, userFactory, displayUserIdGenerator)
+				r, _ := http.NewRequest("GET", "/sky/issuer/token", nil)
+				generator.GenerateAccessTokenReturns("123abc", nil)
+				handler.ServeHTTP(httptest.NewRecorder(), r)
+			}
+
+			It("parses claims from the id_token when present", func() {
+				serveTokenResponse(`{"access_token":"123","token_type":"bearer","expires_in":1234,"id_token":"a.b.c"}`)
+
+				Expect(claimsParser.ParseClaimsCallCount()).To(Equal(1))
+				Expect(claimsParser.ParseClaimsArgsForCall(0)).To(Equal("a.b.c"))
+			})
+
+			It("parses claims from the access_token when id_token is absent (RFC 8693 token exchange)", func() {
+				serveTokenResponse(`{"access_token":"x.y.z","issued_token_type":"urn:ietf:params:oauth:token-type:access_token","token_type":"bearer","expires_in":1234}`)
+
+				Expect(claimsParser.ParseClaimsCallCount()).To(Equal(1))
+				Expect(claimsParser.ParseClaimsArgsForCall(0)).To(Equal("x.y.z"))
+			})
+		})
 	})
 
 	Describe("Token Generation", func() {


taylorsilva:
Thanks for opening an issue and opening a PR to fix it. I'll spend some time validating the issue and your proposed fix. I like to be extra cautious around any auth changes. The proposed change does seem good and I appreciate you adding tests for it as well!

When you get a chance can you please do the EasyCLA sign-off by clicking the link in the previous comment. TY!

ibaileyStingray:
Thanks you for you feedback. Being an corporate contributor, the easyCLA should take a few days. But I am actively working for it to reach the good person desk :). Meanwhile, I'll stay available for any question or suggestion from your side.
thanks again.

ibaileyStingray:
easyCLA step approved. thanks for your patience

taylorsilva:
I'm going to need your help reproducing the original issue and verifying that this PR fixes it. I tried puzzling it out on my own (and with Claude) but couldn't get this working.

In the issue you opened, you mention getting an `${UPSTREAM_TOKEN}`. I tried doing this by logging into a local Concourse and intercepting the `http://web:8080/sky/issuer/callback?code=SOME_CODE&state=xyz` URL from the OIDC provider and pasting it into the provided curl command:

```
curl -i http://localhost:8080/sky/issuer/token \
-u 'fly:Zmx5' \
--data-urlencode 'grant_type=urn:ietf:params:oauth:grant-type:token-exchange' \
--data-urlencode 'connector_id=oidc' \
--data-urlencode 'subject_token_type=urn:ietf:params:oauth:token-type:id_token' \
--data-urlencode "subject_token=SOME_CODE" \
--data-urlencode 'requested_token_type=urn:ietf:params:oauth:token-type:access_token' \
--data-urlencode 'scope=openid profile email federated:id groups'
```
But I got back HTTP 401 and `{"error":"access_denied"}`. The web logs also showed a slightly different but similar error to the one in the issue:

```
{"timestamp":"2026-07-13T20:09:26.215517544Z","level":"error","source":"atc","message":"atc.dex.failed to verify subject token",
  "data":{"err":"oidc: failed to verify token: oidc: malformed jwt: go-jose/go-jose: compact JWS format must have three parts","session":"17"}}
```

Not sure what you're using for your local OIDC. The one we used to use in the `./hack/overrides/oidc.yml` is outdated and doesn't work for me at the moment. I tried another one I found. General auth with Concourse works, but can't get this to work.

More precise steps to reproduce would be helpful.

taylorsilva:
I pushed the local oidc setup I had to `master`. You can run it with:
```
docker compose -f docker-compose.yml -f ./hack/overrides/oidc.yml up -d
```

ibaileyStingray:
Thanks for pushing the local OIDC setup — I reproduced the issue and verified the fix with it. Two things explain what you saw:

1. The subject_token must be a JWT from the upstream IdP, not the callback code. The code=SOME_CODE value from the callback URL is an opaque authorization code. Dex's OIDC connector tries to verify the subject token as a JWT, fails (compact JWS format must have three parts — your error came from atc.dex, i.e. inside Dex), and returns 401 access_denied before the bug in the Concourse middleware is ever reached. The bug only triggers when Dex succeeds: the middleware then discards Dex's successful response and returns an empty 500.

2. One env var is needed for the ID-token flow. With getUserInfo enabled (the default), Dex also sends the subject token to the IdP's UserInfo endpoint, which rejects an ID token with Invalid token use. This matches our production setup (Entra with CONCOURSE_OIDC_DISABLE_GET_USER_INFO=true, as described in #9636). Add under web.environment in hack/overrides/oidc.yml:

`CONCOURSE_OIDC_DISABLE_GET_USER_INFO: "true"`

Repro steps (from a checkout of master, then again with this PR applied):

`docker compose -f docker-compose.yml -f ./hack/overrides/oidc.yml up -d --build`

Mint an upstream ID token headlessly via local-idp's login API (no browser needed):

```
CHALLENGE_ID=$(curl -s http://localhost:9000/login/init \
  -H 'Content-Type: application/json' \
  -d '{"username":"user1","password":"password1","client_id":"client1","scopes":"openid profile email"}' \
  | python3 -c 'import json,sys; print(json.load(sys.stdin)["challenge_id"])')
UPSTREAM_TOKEN=$(curl -s http://localhost:9000/login/complete \
  -H 'Content-Type: application/json' \
  -d "{\"challenge_id\":\"${CHALLENGE_ID}\",\"challenge_data\":\"x\"}" \
  | python3 -c 'import json,sys; print(json.load(sys.stdin)["identity_token"])')
```

Then run the exchange from the issue:

```
curl -i http://localhost:8080/sky/issuer/token \
  -u 'fly:Zmx5' \
  --data-urlencode 'grant_type=urn:ietf:params:oauth:grant-type:token-exchange' \
  --data-urlencode 'connector_id=oidc' \
  --data-urlencode 'subject_token_type=urn:ietf:params:oauth:token-type:id_token' \
  --data-urlencode "subject_token=${UPSTREAM_TOKEN}" \
  --data-urlencode 'requested_token_type=urn:ietf:params:oauth:token-type:access_token' \
  --data-urlencode 'scope=openid profile email federated:id groups'
```

On current master: HTTP 500 with an empty body, and the web log shows the failure in the Concourse middleware (not Dex):

```
atc.dex-server.token-request.parse-id-token:
go-jose/go-jose: compact JWS format must have three parts
```

With this PR applied: HTTP 200 with the RFC 8693 response intact and an opaque Concourse token:

```
{
  "access_token": "<opaque>",
  "token_type": "bearer",
  "expires_in": 86399,
  "issued_token_type": "urn:ietf:params:oauth:token-type:access_token"
}
```

and the returned token authenticates successfully against GET /api/v1/user.

taylorsilva (approved):
TY for the detailed steps. Was able to reproduce it locally and verify the output is what you said it would be. Code changes and tests look good to me as well. Thanks again for the PR!

PR #9640, opened 2026-07-10
add: Health endpoint go-client methods and fly command

Kump3r:

Follow-up to #9588 — adds client library support and a CLI command for the health endpoint.

- Added `GetHealth() (atc.Health, error)` to the `go-concourse` client interface
- Implemented `GET /api/v1/health` client method using `httpAgent` so the response body is decoded on both 200 and 503 (a 503 from this endpoint means the cluster is failing, not that the request itself failed)
- Regenerated `concoursefakes/fake_client.go` with the new method
- Added `fly health` command with table output and `--json` flag, ~~registered as alias `h`~~
- Table shows overall status, database, workers (total/running + unhealthy worker names), and all components with their last run time

## Notes to reviewer

- **`httpAgent` over `connection.Send`**: the health endpoint deliberately returns 503 with a JSON body when the cluster is failing. `connection.Send` treats non-2xx as an error and skips decoding — `httpAgent.Send` lets us decode 503 responses the same as 200, so callers always get a populated `atc.Health` struct regardless of cluster state.

- **No `target.Validate()` in the fly command**: the health endpoint is unauthenticated. Calling `Validate()` would hit `/api/v1/info` and do a version check before the actual health call, which is unnecessary overhead. Follows the same pattern as `fly status`.

- **Exit code is always 0**: a failing cluster is informational, not a fly error. The cluster state is conveyed through the JSON body and table output. Consistent with how `fly status` works.

- **Integration tests use `createFlyRc` + `atcServer.Reset()`**: the suite `BeforeEach` runs a full login sequence and queues 4 handlers. Since `fly health` skips login, we write the target directly to `~/.flyrc` and reset the server to discard those unused handlers. Same pattern as `fly status` integration tests.

### DIFF
--- a/fly/rc/target.go
+++ b/fly/rc/target.go
@@ -170,6 +170,43 @@ func LoadTarget(selectedTarget TargetName, tracing bool) (Target, error) {
 	), nil
 }
 
+// Use for unauthenticated endpoints of targets, like (health, info, etc.).
+func LoadTargetWithoutAuth(selectedTarget TargetName, tracing bool) (Target, error) {
+	var clientCertificate []tls.Certificate
+
+	targetProps, err := selectTarget(selectedTarget)
+	if err != nil {
+		return nil, err
+	}
+
+	caCertPool, err := loadCACertPool(targetProps.CACert)
+	if err != nil {
+		return nil, err
+	}
+
+	clientCertificate, err = loadClientCertificate(targetProps.ClientCertPath, targetProps.ClientKeyPath)
+	if err != nil {
+		return nil, err
+	}
+
+	httpClient := defaultHttpClient(nil, targetProps.Insecure, caCertPool, clientCertificate)
+	client := concourse.NewClient(targetProps.API, httpClient, tracing)
+
+	return NewTarget(
+		selectedTarget,
+		targetProps.TeamName,
+		targetProps.API,
+		targetProps.Token,
+		targetProps.CACert,
+		caCertPool,
+		targetProps.ClientCertPath,
+		targetProps.ClientKeyPath,
+		clientCertificate,
+		targetProps.Insecure,
+		client,
+	), nil
+}
+
 func LoadUnauthenticatedTarget(
 	selectedTarget TargetName,
 	teamName string,
--- a/fly/commands/fly.go
+++ b/fly/commands/fly.go
@@ -21,6 +21,7 @@ type FlyCommand struct {
 	Login  LoginCommand  `command:"login" alias:"l" description:"Authenticate with the target"`
 	Logout LogoutCommand `command:"logout" alias:"o" description:"Release authentication with the target"`
 	Status StatusCommand `command:"status" description:"Login status"`
+	Health HealthCommand `command:"health" alias:"h" description:"Check cluster health"`
 	Sync   SyncCommand   `command:"sync"  alias:"s" description:"Download and replace the current fly from the target"`
 
 	ActiveUsers ActiveUsersCommand `command:"active-users" alias:"au" description:"List the active users since a date or for the past 2 months"`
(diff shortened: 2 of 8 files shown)


taylorsilva (changes requested):
Tried manually testing it first and got HTTP 401. Can reproduce by doing:

```
$ fly -t l logout
$ fly -t l health
error: Unexpected Response
Status: 401 Unauthorized
Body: not authorized
```

I had to `fly login` first to get it to work. I think you need to use `LoadUnauthenticatedTarget()` instead of `LoadTarget()` to avoid this scenario: https://github.com/concourse/concourse/blob/3c3fc8d348f5d1cbf3bf3b723679dce0863332f5/fly/rc/target.go#L173

Not sure if that function is appropriate here either, func signature is much longer! Might need a new one to mirror the usability of `LoadTarget()`. Feel free to rename/refactor the functions in `target.go`, `LoadUnauthenticatedTarget()` is only used in one place.

Table and json output looks good otherwise.

Kump3r:
I noticed this as well during my testing, but I couldn’t find any other fly commands without authentication, so decided I should follow that pattern. I will take a look at your recommendation.

Kump3r:
Okay it was a simple enough change, I decided to just mirror the LoadTarget with nil for token and switch the health command to use it instead. Also added a test with a specifically expired token, so I can check that the token is not required.

taylorsilva (approved):

fly/rc/target.go
```diff
@@ -170,6 +170,43 @@ func LoadTarget(selectedTarget TargetName, tracing bool) (Target, error) {
...
+		targetProps.CACert,
+		caCertPool,
+		targetProps.ClientCertPath,
+		targetProps.ClientKeyPath,
+		clientCertificate,
+		targetProps.Insecure,
+		client,
+	), nil
+}
+
```
IvanChalukov, line 209: nit: suggestion for code duplication
```suggestion
func LoadTarget(selectedTarget TargetName, tracing bool) (Target, error) {
	targetProps, err := selectTarget(selectedTarget)
	if err != nil {
		return nil, err
	}
	return buildTarget(selectedTarget, targetProps, targetProps.Token, tracing)
}

// LoadTargetWithoutAuth loads a target without attaching credentials to the
// HTTP transport. Use for public endpoints (health, etc.) where a missing or
// expired token must not block the call.
func LoadTargetWithoutAuth(selectedTarget TargetName, tracing bool) (Target, error) {
	targetProps, err := selectTarget(selectedTarget)
	if err != nil {
		return nil, err
	}
	return buildTarget(selectedTarget, targetProps, nil, tracing)
}

func buildTarget(selectedTarget TargetName, targetProps TargetProps, token *TargetToken, tracing bool) (Target, error) {
	var clientCertificate []tls.Certificate

	caCertPool, err := loadCACertPool(targetProps.CACert)
	if err != nil {
		return nil, err
	}

	clientCertificate, err = loadClientCertificate(targetProps.ClientCertPath, targetProps.ClientKeyPath)
	if err != nil {
		return nil, err
	}

	httpClient := defaultHttpClient(token, targetProps.Insecure, caCertPool, clientCertificate)
	client := concourse.NewClient(targetProps.API, httpClient, tracing)

	return NewTarget(
		selectedTarget,
		targetProps.TeamName,
		targetProps.API,
		token,
		targetProps.CACert,
		caCertPool,
		targetProps.ClientCertPath,
		targetProps.ClientKeyPath,
		clientCertificate,
		targetProps.Insecure,
		client,
	), nil
}

```

taylorsilva: @IvanChalukov feel free to do a follow-up PR with these refactors if you've got time.

Kump3r: https://github.com/concourse/concourse/pull/9644

PR #9643, opened 2026-07-13
add shared path support to kubernetes, CONCOURSE_KUBERNETES_SHARED_PATH

VikranthBala:
Changes proposed by this PR

As of now, it is not possible to configure shared paths for Kubernetes secret manager as opposed to other cred managers such as Vault,AWS. So I propose to add a new parameter --kubernetes-shared-path similar to the Vault cred manager. This parameter can be optional (as with Vault). This PR provides all the changes needed for this functionality.
Notes to reviewer

Release Note

Added `--kubernetes-shared-path` to configure shared secret paths for Kubernetes cred manager similarly to the one for Vault, SSM.

PR is for the issue: https://github.com/concourse/concourse/issues/9062

### DIFF
--- a/atc/creds/kubernetes/manager.go
+++ b/atc/creds/kubernetes/manager.go
@@ -13,19 +13,19 @@ import (
 )
 
 type KubernetesManager struct {
-	InClusterConfig bool   `long:"in-cluster" description:"Enables the in-cluster client."`
-	ConfigPath      string `long:"config-path" description:"Path to Kubernetes config when running ATC outside Kubernetes."`
-	NamespacePrefix string `long:"namespace-prefix" default:"concourse-" description:"Prefix to use for Kubernetes namespaces under which secrets will be looked up."`
-	SharedPath      string `long:"shared-path" description:"Path under which to look up shared credentials. This value is appended to the configured namespacePrefix to determine the shared namespace."`
+	InClusterConfig       bool   `long:"in-cluster" description:"Enables the in-cluster client."`
+	ConfigPath            string `long:"config-path" description:"Path to Kubernetes config when running ATC outside Kubernetes."`
+	NamespacePrefix       string `long:"namespace-prefix" default:"concourse-" description:"Prefix to use for Kubernetes namespaces under which secrets will be looked up."`
+	NamespaceSharedSuffix string `long:"shared-namespace-suffix" description:"Appended to the namespace-prefix, which combined should match an existing Kubernetes namespace used to lookup shared secrets in."`
 }
 
 func (manager *KubernetesManager) MarshalJSON() ([]byte, error) {
 	// XXX: Get Health
 	return json.Marshal(&map[string]any{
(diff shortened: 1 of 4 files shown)


VikranthBala:
@taylorsilva I'd like your thoughts on how **sharedPath** should map to Kubernetes.

Other credential managers naturally support hierarchical paths, but Kubernetes secrets are namespaced rather than path-based. To support sharedPath, the current implementation maps it to a dedicated namespace:

```
sharedPath = shared
↓
concourse-shared/<secret-name>
```

This has the benefit of preserving Kubernetes' namespace-based RBAC model. Teams can be granted access to their own namespace (concourse-<team>) and the shared namespace (concourse-shared), while the root namespace remains isolated.

An alternative would be to store shared secrets in the root namespace (e.g. concourse-/shared.secret-name), but that either requires broader RBAC permissions on the root namespace or explicit resourceNames entries for every shared secret, which becomes harder to manage.

Does mapping sharedPath to a dedicated namespace seem like the right Kubernetes interpretation, or would you prefer a different approach?

taylorsilva (commented):
My first thought is what you've done here: operators can define a shared k8s namespace and this new config option points to that namespace.

The help text should also note that the given shared path is appended to the namespace prefix.

VikranthBala:
@taylorsilva I have updated the PR, please review it & let me know if any changes needs to be done

atc/creds/kubernetes/manager.go
```diff
@@ -16,6 +16,7 @@ type KubernetesManager struct {
 	InClusterConfig bool   `long:"in-cluster" description:"Enables the in-cluster client."`
 	ConfigPath      string `long:"config-path" description:"Path to Kubernetes config when running ATC outside Kubernetes."`
 	NamespacePrefix string `long:"namespace-prefix" default:"concourse-" description:"Prefix to use for Kubernetes namespaces under which secrets will be looked up."`
+	SharedPath      string `long:"shared-path" description:"Path under which to look up shared credentials. This value is appended to the configured namespacePrefix to determine the shared namespace."`
```
taylorsilva: I kept thinking about the implied semantics given the name of the flag and I don't like it. `SharedPath` implies I can set this to some kind of system-like path (e.g. `some/path/here`) like the other cred managers, when that's not the case here.

I'd prefer we be explicit about what this does so I'll suggest the following change:


```suggestion
	NamespaceSharedSuffix      string `long:"shared-namespace-suffix" description:"Appended to the namespace-prefix, which combined should match an existing Kubernetes namespace used to lookup shared secrets in."`
```

Update the name and related vars, then this looks good to merge.

taylorsilva (changes requested):

VikranthBala:
@taylorsilva I have Updated the PR with renamed flag & vars. Please review & let me know if these changes look fine

taylorsilva (approved):
LGTM - Thanks again for your work on this!

PR #9654, opened 2026-07-23
fix tarfs absolute symlink test for Windows

aliculPix4D:
Why was https://github.com/concourse/concourse/commit/146ab3352284e09afe823cf81e326d1316adb048 the wrong fix?

Code path for absolute path `/absolute/path/file` on Linux:

linux

Code path for path `/absolute/path/file` on Windows:

623207706-326c2525-12c6-4b7b-8ddc-f7272118954a

We must ensure we are testing the same code path on each platform. Hence, we must test an absolute path (`/absolute/path/file`)  on Unix and an absolute path on Windows (`C:\foo`).

The fact that the test was green at the end (after `filepath.FromSlash(header.Linkname)` manipulation ) is just due to the fact that at the end, by chance, in both cases, we end up calling the same L134 in production code:
```
err = root.Symlink(filepath.FromSlash(header.Linkname), filePath)
```

### DIFF
--- a/go-archive/tarfs/extract_test.go
+++ b/go-archive/tarfs/extract_test.go
@@ -219,10 +219,15 @@ var _ = Describe("ExtractEntry", func() {
 
 	Context("symlinks", func() {
 		It("does not modify absolute paths", func() {
+			linkname := "/absolute/path/file"
+			if runtime.GOOS == "windows" {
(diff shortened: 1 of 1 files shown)


aliculPix4D:
@odormond @marco-m-pix4d FYI

taylorsilva (approved):

Issue #9609, opened 2026-06-19
Windows volume streaming still broken for intra-repo symlinks due to `os.Root.Symlink` path separator bug (Go #80073)


aliculPix4D:
## Summary

[v8.2.4](https://github.com/concourse/concourse/releases/tag/v8.2.4) includes [#9601](https://github.com/concourse/concourse/pull/9601), which addressed [#9597](https://github.com/concourse/concourse/issues/9597) by switching Windows tar extraction to `os.Root`. That unblocks the false-positive `BreakoutError` for safe intra-repo symlinks, but **symlinks are still broken on Windows** because [`os.Root.Symlink` does not normalize path separators on Windows](https://github.com/golang/go/issues/80073).

Tar headers from Linux/git use Unix-style forward slashes (e.g. `../shared/images-vars.tf`). `os.Symlink` normalizes these to backslashes on Windows; `os.Root.Symlink` writes them verbatim into the NTFS reparse point. Windows requires backslashes for relative symlink targets, so the symlinks are created but **unreadable**.

This affects any input streamed to a Windows worker (git resources, task inputs, COW volumes via robocopy).

## Regression timeline

| Version | Behaviour |
|---------|-----------|
| ≤ 8.1.0 | Intra-repo symlinks work on Windows |
| ≥ 8.1.1 ([#9486](https://github.com/concourse/concourse/pull/9486)) | Safe intra-repo symlinks rejected with `BreakoutError` ([#9597](https://github.com/concourse/concourse/issues/9597)) |
| 8.2.4 ([#9601](https://github.com/concourse/concourse/pull/9601)) | Stream-in no longer rejected, but symlinks unreadable on Windows (Go [#80073](https://github.com/golang/go/issues/80073)) |

## Root cause

In `go-archive/tarfs/extract.go`, PR #9601 calls:

```go
err = root.Symlink(header.Linkname, filePath)
```

When `header.Linkname` contains forward slashes (typical for git repos created on Linux), `root.Symlink` stores them unchanged. On Windows, resolution fails with:

> The filename, directory name, or volume label syntax is incorrect.

Standalone reproducers and build logs: https://github.com/Pix4D/concourse-tar-streaming-issue

Upstream Go issue: https://github.com/golang/go/issues/80073

### Steps to Reproduce

1. Deploy Concourse **v8.2.4** with Linux and Windows workers.
[...]

PR #9655, opened 2026-07-23
go-archive:tarfs: test relative and absolute symlink extraction

aliculPix4D:
go-archive:tarfs: test relative and absolute symlink extraction

- verify extracted symlinks preserve their expected targets across platforms.
- regression test for: https://github.com/concourse/concourse/issues/9609 and https://github.com/golang/go/issues/80073

### DIFF
--- a/go-archive/tarfs/extract_test.go
+++ b/go-archive/tarfs/extract_test.go
@@ -64,6 +64,12 @@ var _ = Describe("Extract", func() {
 			Link: "/some-file",
 			Mode: 0755,
 		},
+		{
+			// Relative dir symlink: forward-slash targets were unreadable and broke robocopy on Windows (#9609).
+			Name: "./symlink-dir/relative-dir-symlink",
+			Link: "../nonempty-dir",
+			Mode: 0755,
+		},
 	}
 
 	BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


go-archive/tarfs/extract_test.go
```diff
@@ -118,16 +124,18 @@ var _ = Describe("Extract", func() {
 
 		Expect(emptyDirInfo.IsDir()).To(BeTrue())
 
-		target, err := os.Readlink(filepath.Join(extractionDest, "some-symlink"))
```
aliculPix4D, line 121: These 3 lines are not removed, they are just generalized in the `verifySymlink` helper function to avoid repetition.

e.g as you can see, these checks were done only for 1 out of 3 symlinks in the archive.

go-archive/tarfs/extract_test.go
```diff
@@ -118,16 +124,18 @@ var _ = Describe("Extract", func() {
...
-		Expect(err).NotTo(HaveOccurred())
-		Expect(target).To(Equal("some-file"))
-
-		symlinkInfo, err := os.Lstat(filepath.Join(extractionDest, "some-symlink"))
-		Expect(err).NotTo(HaveOccurred())
-
-		if runtime.GOOS != "windows" {
-			Expect(symlinkInfo.Mode() & 0755).To(Equal(os.FileMode(0755)))
```
aliculPix4D, line 129: This was catched by @odormond during the review in Pix4D fork: https://github.com/Pix4D/concourse/pull/829#discussion_r3630641725

If symlink was successfully created (already confirmed with `os.Readlink`) the mode must also match, so there is no need to even call `os.Lstat`.


I am quoting him here:
> there is no way to change the mode bits of symlink. So if it is a symlink as os.Readlink succeeded, then its mode is 0755.

> chmod() changes the mode of the file specified, whose pathname is given in pathname, which is dereferenced if it is a symbolic link.
-- man 2 open

aliculPix4D:
@odormond @marco-m-pix4d FYI

taylorsilva (approved):

PR #9489, opened 2026-03-16
Opaque User Data Feature (Issue #8426)

alicul:
Support for opaque `user_data` field in pipeline YAML configurations, allowing users to store arbitrary metadata that is preserved through the set-pipeline/get-pipeline cycle.

The feature allows users to store arbitrary metadata in pipeline configurations, addressing the use case described in issue #8426 where teams need to attach organizational information to pipelines for programmatic processing.

**Closes GitHub Issue**: https://github.com/concourse/concourse/issues/8426

### Pipeline YAML with user_data
```yaml
jobs:
  - name: build
    plan:
      - task: test
        config:
          platform: linux
          run:
            path: echo
            args: ["hello"]

user_data:
  organization: Platform Team
  contact: platform@example.com
  labels:
    - production
    - critical
  metadata:
    cost_center: CC-1234
    region: us-west-2
```

### Workflow
```bash
# Set pipeline with user_data
fly -t target set-pipeline -p my-pipeline -c pipeline.yml

# Get pipeline - user_data is preserved
fly -t target get-pipeline -p my-pipeline
# Output includes the same user_data block
```

PR #9661, opened 2026-07-23
Add pipeline-level description field

blarghmatey:

Surfaces a pipeline's `user_data` in the web UI, on its own page behind a button. Proposed in https://github.com/orgs/concourse/discussions/9660; reworked from the original approach per review — see [this comment](https://github.com/concourse/concourse/pull/9661#issuecomment-5135175495) for the before/after.

Frontend-only. `user_data` already exists end-to-end from #9489 — `db.Pipeline.UserData()`, the presenter, and `user_data` on the API pipeline payload — so there are no `atc/` changes here, and no new pipeline config surface.

* New route `/teams/:team/pipelines/:pipeline/info` and page module (`web/elm/src/PipelineInfo/`).
* Reached from an `information-outline` button in the pipeline page's existing top-right button group, alongside the pin/favorite/pause controls. The button is only rendered when the pipeline actually has `user_data`.
* `user_data` is an opaque JSON value, so the page picks a rendering from its shape:

  | `user_data` | rendering |
  |---|---|
  | a string | markdown |
  | object with a string `description` | that key as markdown, remaining keys as YAML below |
  | anything else | YAML |

* YAML rendering is a small local module (`web/elm/src/Yaml.elm`) rather than a new dependency. It renders multi-line strings as literal block scalars and quotes conservatively — anything that would otherwise read back as a number, bool, null, or YAML syntax. It doesn't re-sort keys; note they arrive alphabetized already, since `user_data` is an `any` on the Go side and `encoding/json` sorts map keys.
* Markdown renders via `dillonkearns/elm-markdown`'s `defaultHtmlRenderer`, left unmodified (no custom `html` renderer) so raw HTML embedded in the source doesn't render — `user_data` is visible to anyone with pipeline view access, including anonymous viewers on public pipelines. Parse/render failure falls back to the raw, still-escaped text.
* Decoding treats an explicit `"user_data": null` as unset, the same as an absent key; `Json.Decode.value` on its own accepts `null` and breaks the encode/decode round-trip.
* Also fixes the production Elm build, which this PR's new dependency broke — see "Build fix" below.
* Dashboard tiles are out of scope — they use a separate `Pipeline` type that doesn't carry `user_data`.

## Screenshots

All captured against a local Concourse (`docker-compose`) running this branch.

**The button, in the existing top-right group.** Only rendered when the pipeline has `user_data` — with it (left) and without (right):

| has `user_data` | no `user_data` |
|---|---|
| top bar with info button | top bar without info button |

The pipeline page is otherwise untouched:

pipeline page with info button

**Object with a `description`** — markdown, with the remaining keys as YAML below:

info page with description and yaml

**A bare string** — rendered wholly as markdown:

info page with string markdown

**Anything else** — rendered wholly as YAML (note nesting, sequences, and `"#platform"` quoted because `#` would otherwise start a comment):

info page with yaml only

**No `user_data`** — the route is still reachable by URL, so it has an empty state:

info page empty state

## Build fix

Adding `dillonkearns/elm-markdown` silently broke `yarn build-elm`, producing a **zero-byte `elm.min.js`** — which is the file `index.html` loads, so the entire web UI rendered blank. Two faults combined:

1. elm-markdown's `Markdown.InlineParser.isWhitespace` matches U+2028 and U+2029. The Elm compiler emits Char literals verbatim, and both are line terminators in ECMAScript before ES2019, so a literal one inside a string literal is a syntax error to an ES5-era parser. `uglify-js` is ES5-only and rejected the bundle with `Unterminated string constant`.
2. `build-elm` piped `uglifyjs | uglifyjs --mangle --output`. A shell pipeline reports the exit status of its **last** command, so the first stage dying left the second reading empty stdin and writing an empty file — exit 0, no output, broken bundle.

`hack/escape-js-line-separators` rewrites those two characters to their escape sequences before minification (semantically identical), and the pipe is replaced with an intermediate file and `&&` so any stage's failure is fatal.

**This is what failed `concourse-ci/watsjs`** on the earlier revision: with no Elm loaded the page never rendered, so every test timed out in `beforeEach` waiting for `#user-id`, including `smoke > reaching the internet`. I originally read that as unrelated CI flake and said so in a comment — that was wrong, and it's fixed here.

## Notes to reviewer

* Test coverage for the YAML renderer, the routing, the info button's visibility and link target, and each of the three rendering shapes, plus the raw-HTML case. Full Elm suite passes: 3084 tests (was 3036).
* Button visibility, its link target, and the empty state were also verified in a real browser against a running server, not just in the unit tests.
* Rebased onto latest `master`; no conflicts.

## Release Note

* A pipeline's `user_data` is now viewable in the web UI, from an info button on the pipeline page.


taylorsilva (changes requested):
Thanks for the PR and happy to see your excitement around adding such a feature!

As I mentioned in the discussion post, I think using the existing `user_data` field would be useful here. https://github.com/concourse/concourse/pull/9489

`user_data` feels like a good fit for data like "pipeline description" since it was meant as a catch-all for arbitrary information. We could have heuristics where if the value of `user_data` is a string we render it in markdown, or the UI can look for a `description` field and only display and render that field as markdown. Otherwise, render the entire `user_data` as YAML.

Another motivation I have to use the `user_data` config instead of adding a new top-level config option to the pipeline is that it reduces the amount of code needed to implement the feature. You can eliminate all the `atc/` code and focus purely on the frontend changes. Means less future maintenance for the project as well.

Regarding the design, I don't like the way this looks. It feels like a sticky note has been attached to the screen. I'm also generally wary about adding new visual stuff to the pipeline overview page. It can already get really noisy with a lot of resources, jobs, and groups.

I'd prefer to not display this information by default and instead hide it behind a button. I'd be fine with rendering it as a modal (I don't think we do modals anywhere else in the UI at the moment, so new ground to be broken here if going in that direction) or on a new page. A new page would be simplest.

I'd use the `information-outline` icon for the button: https://pictogrammers.com/library/mdi/icon/information-outline/
I'd probably add the button to the top-right with the existing set of buttons.

blarghmatey:
Thanks for the thorough review — this was the right call on both counts. Reworked in 2200eda and rebased onto latest `master` (18 commits, no conflicts).

## `user_data` instead of a new config field

You were right that this needed no `atc/` code at all. `user_data` is already plumbed end-to-end from #9489 — `db.Pipeline.UserData()`, the presenter, and `user_data` on the API pipeline payload — and nothing on the Elm side was consuming it yet. So the entire backend side of this PR is gone: no `Config.Description`, no `description` column or migration, no `db.Pipeline` getter, no presenter field, no regenerated fake. The diff against `master` is now frontend-only:

```
$ git diff origin/master...HEAD --name-only -- atc/
(empty)
```

## Behind a button, on its own page

Agreed on the sticky-note critique — and on not adding always-on visual weight to a page that's already dense. It now lives at `/teams/:team/pipelines/:pipeline/info`, reached from an `information-outline` button in the existing top-right button group, next to the pin/favorite/pause controls. The button only renders when the pipeline actually has `user_data`, so pipelines without it are pixel-identical to today.

I went with the page rather than the modal, per your "simplest" steer. Two things reinforced it once I started: `user_data` is arbitrary and potentially long, and a page scrolls without a height ceiling where a modal needs a fixed max-height and an inner scroll container; and a real URL is genuinely useful for something read as documentation.

One note in case it's useful for future work — you mentioned we don't do modals anywhere. That's true for modals proper, but `Pipeline/PinMenu/PinMenu.elm` is already a floating panel anchored to a top-bar icon on this exact page, with click-to-expand and hover state. It's not a modal (no backdrop, escape handling, or focus trap), but it would be a decent starting point if a modal is ever wanted. Happy to switch this over if you'd rather.

## Rendering

Implemented all three heuristics you floated, since `user_data` is opaque and any of the shapes are legal:

| `user_data` | rendering |
|---|---|
| a string | markdown |
| object with a string `description` | that key as markdown, remaining keys as YAML below |
| anything else | YAML |

The YAML renderer is a small local module (`web/elm/src/Yaml.elm`) rather than a new dependency. It preserves the author's key order instead of sorting — this is read as documentation — renders multi-line strings as literal block scalars, and quotes conservatively: anything that would otherwise read back as a number, bool, null, or YAML syntax.

Markdown still goes through `dillonkearns/elm-markdown`'s `defaultHtmlRenderer` unmodified, so raw HTML in the source doesn't render (`html = Markdown.Html.oneOf []`). `user_data` is visible to anyone with pipeline view access, including anonymous viewers on public pipelines, so opting into raw-HTML passthrough would be a mistake here. Parse/render failure falls back to the raw, still-escaped text.

One decoder subtlety worth flagging: an explicit `"user_data": null` is now treated as unset, the same as an absent key. `Json.Decode.value` accepts `null` happily, which broke the existing `pipeline encoding/decoding are inverses` test — caught by the suite rather than by inspection.

## Tests

Added coverage for the parts that now have real logic — the YAML renderer, the routing, the button's visibility and link target, and each of the three rendering shapes, plus the raw-HTML case. 3084 Elm tests pass (was 3036).

## `concourse-ci/watsjs`

This one is unrelated to the PR — worth a look if it's recurring. Every test in every file failed identically in `beforeEach` with `TimeoutError: Waiting for selector '#user-id' failed`, including `smoke › reaching the internet`, before the job hit `timeout exceeded`. `#user-id` is the login/top-bar user menu, and this PR touches no auth, top-bar, or login code. Reads like the test cluster never came up rather than anything in the diff. It'll re-run against the reworked code on this push.

## Still open

The dashboard tiles still don't surface `user_data` — they use a separate `Pipeline` type that doesn't carry it. Deliberately out of scope here, happy to follow up if it's wanted.

I don't have a Concourse instance handy to grab fresh screenshots of the new page — the ones in the description are of the old inline panel and no longer reflect the UI. Glad to add current ones if that's useful for review.

blarghmatey:
Correction to my last comment, plus screenshots.

## I was wrong about `concourse-ci/watsjs`

I said that failure was unrelated to this PR. It wasn't — it was caused by it, and I should have dug in rather than pattern-matching "every test failed identically, therefore infra."

Adding `dillonkearns/elm-markdown` silently broke `yarn build-elm`, producing a **zero-byte `elm.min.js`**. That's the file `index.html` actually loads, so the whole UI rendered blank — which is exactly why every test timed out in `beforeEach` waiting for `#user-id`, `smoke > reaching the internet` included. The uniformity of the failures was the clue that it was a real, total breakage rather than flake; I read it as the opposite.

Two faults had to line up:

1. elm-markdown's `Markdown.InlineParser.isWhitespace` matches U+2028 and U+2029. The Elm compiler emits Char literals verbatim, and both are line terminators in ECMAScript before ES2019 — a literal one inside a string literal is a syntax error to an ES5-era parser. `uglify-js` is ES5-only and bailed with `Unterminated string constant`. The emitted JavaScript is valid; the minifier just predates the spec change.

2. That alone would have been loud, except `build-elm` piped `uglifyjs | uglifyjs --mangle --output`. A shell pipeline reports the exit status of its **last** command, so the first stage dying left the second reading empty stdin and writing an empty file — exit 0, no output, broken bundle.

Reproduced locally against `master` as a control: master's bundle minifies to ~916 KB, this branch's produced 0 bytes.

Fixed in b720fad — `hack/escape-js-line-separators` rewrites those two characters to their escape sequences before minification (semantically identical), and the pipe becomes an intermediate file plus `&&` so any stage's failure is fatal. `elm.min.js` is now ~376 KB. Worth noting the pipe fault is pre-existing and independent of this PR: any dependency carrying those characters would have hit the same silent-empty-bundle failure.

## Screenshots

Stood up a local Concourse via `docker-compose` and captured the reworked UI. The old inline-panel screenshots are gone from the description.

The button only appears when the pipeline has `user_data` — with (left) and without (right):

| has `user_data` | no `user_data` |
|---|---|
| with | without |

Object with a `description` — markdown plus the remaining keys as YAML:

description plus yaml

The other two shapes (bare string → markdown, anything else → YAML) and the empty state are in the description.

## Two smaller things from testing against a real server

- The YAML block sat flush against the last markdown paragraph, because the markdown's own last-child margin is zeroed. Fixed in 71a065b.
- I'd claimed the YAML renderer preserves the author's key order. It preserves the order it *receives*, but that isn't the authored order — `user_data` is an `any` on the Go side, so `encoding/json` has already sorted the map keys alphabetically before the frontend sees them. Comment and PR text corrected; you can see the alphabetization in the screenshots.

Behaviour verified in-browser as well as in unit tests: button present with `user_data`, absent without, correct link target, and the empty state on `/info` for a pipeline with none.

bcersows:
Love it! One tiny nitpick, is the missing vertical dash between the pin and info buttons intended? :)

blarghmatey:
CI update on 21c5044 — both bugs I introduced are now confirmed fixed by CI, not just locally.

| check | result |
|---|---|
| `concourse-ci/watsjs` | **pass** — confirms the zero-byte `elm.min.js` diagnosis |
| `concourse-ci/unit` | **pass** — confirms the `elm-analyse` fix |
| `concourse-ci/containerd-integration` | pass |
| `concourse-ci/testflight` | pass |
| `concourse-ci/compile-check`, `go-format-check`, `validate-labels`, EasyCLA | pass |
| `concourse-ci/integration` | **fail** — unrelated, details below |

## One more real bug, found and fixed

`unit` was failing on `yarn analyse` from my very first reworked push, and I missed it for three pushes — my fault for not running `elm-analyse` locally alongside `elm-test`:

```
src/Routes.elm                     Record has only one field
src/PipelineInfo/PipelineInfo.elm  Record has only one field
```

Both were records out of habit, copying neighbouring routes that happen to carry more than one field. `Routes.PipelineInfo` now takes `Concourse.PipelineIdentifier` positionally, and `PipelineInfo`'s single-field `Flags` alias is gone. Fixed in 21c5044; `elm-analyse` reports `Found 0 message(s)`.

## The `integration` failure looks unrelated

`TestVaultV2WithUnmountPath` (`integration/creds/vault_test.go:99`) timed out after 100s with `Condition never satisfied: should have a running worker` — a Vault credential-manager test that stands up docker-compose and waits for a worker to register.

The diff between the build where it **passed** (71a065b) and where it **failed** (21c5044) is exactly the elm-analyse fix:

```
web/elm/src/Pipeline/Pipeline.elm          |  2 +-
web/elm/src/PipelineInfo/PipelineInfo.elm  | 25 +++++-----------
web/elm/src/Routes.elm                     | 14 ++++-------
web/elm/src/SubPage/SubPage.elm            |  8 +++----
web/elm/src/Views/TopBar.elm               |  2 +-
web/elm/tests/PipelineInfoTests.elm        |  8 ++-----
```

Six Elm files, zero Go. An Elm record-vs-positional constructor change can't reach Go worker registration, and `integration` passed on 1aecf591 and 71a065b with identical Go.

`containerd-integration` behaved the same way earlier in this PR — `TestNewContainerEnforcesTimeoutOnTask` failed once (it `SIGSTOP`s containerd and races a ~1s request timeout, and got `task not found` instead of `deadline exceeded`), then passed again on the next push with the same Go code.

I'm deliberately **not** pushing an empty commit to retrigger — I'd rather not burn a full cycle or re-roll the other jobs. Happy to if you'd like, but a maintainer re-running just `integration` is cheaper. Flagging both tests as flaky in case that's useful independent of this PR; I'm glad to open a separate issue with the evidence.

Given my earlier mistake of calling watsjs "unrelated flake" when it was my bug, I want to be explicit that this call rests on a different footing: a same-PR A/B where the diff between pass and fail contains no code the failing test executes.

blarghmatey:
@bcersows Not intentional — good catch, thanks. Fixed in d3f836f.

Two things were wrong, and the second is why it wasn't just a one-liner.

The container had no `border-left` at all, which every other item in that group carries:

```elm
-- favoritedIcon, pauseToggle, and PinMenu all do this
style "border-left" <| "1px solid " ++ Colors.background
style "background-color" Colors.topBarBackground
```

But adding the border alone gave a divider only as tall as the icon, still visibly different from its neighbours. The others size themselves with a `17px` margin on the icon rather than padding on the container — 17 + 20 + 17 is exactly the 54px top bar, which is what makes their dividers full-height. I'd used horizontal padding on the container instead, so mine was 20px tall.

Now matching the sibling pattern rather than a padding-based one of my own:

top bar with divider

All four dividers in the group are identical now. Verified against a local Concourse rather than by inspection — the intermediate short-divider state looked fine in code and only showed up in the screenshot.

taylorsilva:
Thanks for all the work on this and TY @bcersows for reviewing it! I like everything I'm seeing in the screenshots. Super simple and no frills 👍 I'll play around with it a bit and review the code before merging. Thanks for your patience.

hack/escape-js-line-separators
taylorsilva, line 1: Hard no on this weird hacky workaround the LLM came up with. Why is this even needed? What change made the addition of this script necessary?

Okay, read the commit message. Taking a look at uglify-js now.

taylorsilva: uglify-js looks like abandonware: https://github.com/mishoo/UglifyJS

Not going to dig into git history, but I'd wager we started using this years ago. Probably doesn't make sense to anymore.

Looks like https://github.com/terser/terser is the modern choice. Looks like an easy swap too.

taylorsilva (changes requested):

hack/escape-js-line-separators
taylorsilva: I tried to push changes to your PR branch here but it looks like your org has it locked down.

Update this PR to remove the usage of uglify-js and replace it with terser. It looks like we should have moved off uglify-js anyways.

web/elm/src/Yaml.elm
taylorsilva, line 1: I'm not crazy about this as well. I wonder if we could do json -> Elm type -> yaml. Options for a package to use is not great though.

Most recent package is https://package.elm-lang.org/packages/miniBill/elm-yaml/latest/ which is a fork of a fork...

I worry about playing wac-a-mole with bugs, that's what makes me hesitant to add this.

Then again, all we need is to convert from Elm type to yaml. That part probably works fine with any of the available packages. I imagine the hard part of the decode/encode problem is the decoding part(??) and we only need the encoding part.

I really don't like either option here. This hand-rolled encoder is probably _fine_. We don't need it to produce perfect yaml, just yaml-looking strings that we can render.

Okay, I think I'm settling on being _okay_ with this file and its code as long as we keep in mind that we are not promising that this returns 100% valid yaml.

web/elm/src/PipelineInfo/Styles.elm
```diff
@@ -0,0 +1,59 @@
...
+    [ style "word-wrap" "break-word" ]
+
+
+yaml : List (Html.Attribute msg)
+yaml =
+    [ style "background-color" Colors.groupsBarBackground
+    , style "padding" "15px"
+    , style "margin" "0"
+    , style "overflow-x" "auto"
+    , style "font-family" "Courier, monospace"
```
taylorsilva: Use `Views.Styles.fontFamilyDefault`
https://github.com/concourse/concourse/blob/9f3f88fcf88bff2972865905ae8f67d7fb648cf1/web/elm/src/Views/Styles.elm#L49-L51

web/elm/src/PipelineInfo/Styles.elm
```diff
@@ -0,0 +1,59 @@
...
+    , style "white-space" "pre"
+    ]
+
+
+empty : List (Html.Attribute msg)
+empty =
+    [ style "color" Colors.dashboardText
+    , style "opacity" "0.5"
```
taylorsilva: Why set opacity? Makes it harder to read the text:

As-is:
Image

Without opacity set:

Image

taylorsilva (changes requested):
Thanks for your patience! This looks pretty good, just two cosmetic changes I'd like to change and of course swapping out uglify-js for terser.

I left my musings about `Yaml.elm`, no change needed there though. I appreciate the tests added alongside it.

nit: like all LLM things, tons of comments. The one's explaining "why" aren't bad, but lots of "this function does xyz" type of comments. If you can remove them I'd appreciate it. Not a blocker.

hack/escape-js-line-separators
blarghmatey: Fixed in 9fb10ecc0: swapped uglify-js for terser and dropped hack/escape-js-line-separators -- terser's parser handles the U+2028/U+2029 Elm emits natively, so the escape step isn't needed. Verified with a real elm make -> terser build (392K minified output, parses as valid JS) plus the debug build path.

web/elm/src/Yaml.elm
blarghmatey: No change made per your last comment ("I think I'm settling on being okay with this file").

web/elm/src/PipelineInfo/Styles.elm
blarghmatey: Fixed in badd07628: now uses Views.Styles.fontFamilyDefault.

web/elm/src/PipelineInfo/Styles.elm
blarghmatey: Fixed in badd07628: opacity removed from the empty-state style.

blarghmatey:
Addressed the round from 21c5044c/79f88fcf: swapped uglify-js for terser and dropped hack/escape-js-line-separators (9fb10ecc0), and fixed the two Styles.elm nits -- shared monospace font, no more opacity on the empty state (badd07628). Both pushed to pipeline-description and all four threads replied to and resolved. Re-checked the comment density across the diff (PipelineInfo.elm, Yaml.elm, Styles.elm, tests) for the 'tons of comments' note in the last review -- every comment left is a why-explanation (e.g. the null-vs-absent user_data decode, the 404-vs-transient-error handling, the YAML quoting rules), nothing restating what the code does, so no changes made there.

package.json
```diff
@@ -40,8 +40,8 @@
     "build-js": "webpack --mode production",
     "test": "cd web/elm && elm-test",
     "build-less": "lessc web/assets/css/main.less web/public/main.out.css && cleancss -o web/public/main.css web/public/main.out.css && rm web/public/main.out.css",
-    "build-elm": "cd web/elm && elm make --optimize --output ../public/elm.js src/Main.elm && uglifyjs ../public/elm.js --compress \"pure_funcs=[F2,F3,F4,F5,F6,F7,F8,F9,A2,A3,A4,A5,A6,A7,A8,A9],pure_getters,keep_fargs=false,unsafe_comps,unsafe\" | uglifyjs --mangle --output ../public/elm.min.js",
-    "build-elm-debug": "cd web/elm && elm make --debug --output ../public/elm.js src/Main.elm && uglifyjs < ../public/elm.js > ../public/elm.min.js",
+    "build-elm": "cd web/elm && elm make --optimize --output ../public/elm.js src/Main.elm && terser ../public/elm.js --compress \"pure_funcs=[F2,F3,F4,F5,F6,F7,F8,F9,A2,A3,A4,A5,A6,A7,A8,A9],pure_getters,keep_fargs=false,unsafe_comps,unsafe\" -o ../public/elm.compressed.js && terser ../public/elm.compressed.js --mangle --output ../public/elm.min.js && rm ../public/elm.compressed.js",
```
taylorsilva:
```suggestion
    "build-elm": "cd web/elm && elm make --optimize --output ../public/elm.js src/Main.elm && terser ../public/elm.js --compress \"pure_funcs=[F2,F3,F4,F5,F6,F7,F8,F9,A2,A3,A4,A5,A6,A7,A8,A9],pure_getters,keep_fargs=false,unsafe_comps,unsafe\" | terser --mangle --output ../public/elm.min.js",
```

Not sure why it decided to save the compress file instead of passing the output over stdout/stdin like the initial command. Reverted command back to previous version and replaced `uglifyjs` with `terser`.

web/elm/src/Pipeline/Pipeline.elm
```diff
@@ -531,9 +534,67 @@ view session model =
         ]
 
 
+{-| Links to the pipeline's `user_data`, and is only shown when there is any
+to link to.
+-}
```
taylorsilva: Comment culling
```suggestion
```

Should add a test for this in: web/elm/tests/PipelineTests.elm
Test should verify that the icon is present when there's `user_data` and not present when there is no `user_data`.

web/elm/src/PipelineInfo/PipelineInfo.elm
```diff
@@ -0,0 +1,284 @@
...
+            )
+
+        PipelineFetched (Err err) ->
+            case err of
+                Http.BadStatus { status } ->
+                    if status.code == 404 then
+                        -- the only failure that means "there is no such
+                        -- pipeline"; anything else leaves the model alone so
+                        -- the poll can recover rather than bouncing to the
+                        -- not-found page on a transient error
```
taylorsilva:
```suggestion
```

404's are pretty self-explanatory...

web/elm/src/PipelineInfo/PipelineInfo.elm
```diff
@@ -0,0 +1,284 @@
...
+        [ Html.text (Yaml.fromJson value) ]
+
+
+{-| Rendered with elm-markdown's `defaultHtmlRenderer`, which does not render
+raw HTML embedded in the markdown source (`html = Markdown.Html.oneOf []`) --
+`user_data` is visible to anyone with pipeline view access, including
+anonymous viewers on public pipelines, so we deliberately don't opt into a
+raw-HTML-passthrough renderer. Falls back to the raw (still-escaped) text on
+any parse or render failure.
+-}
```
taylorsilva:
```suggestion
```

web/elm/tests/PipelineInfoTests.elm
```diff
@@ -0,0 +1,219 @@
...
+
+userData : String -> Json.Decode.Value
+userData json =
+    Json.Decode.decodeString Json.Decode.value json
+        |> Result.withDefault Json.Encode.null
+
+
+{-| Opens /teams/team/pipelines/pipeline/info with the given `user_data`
+already fetched.
+-}
```
taylorsilva:
```suggestion
```

web/elm/tests/YamlTests.elm
```diff
@@ -0,0 +1,141 @@
...
+
+import Expect
+import Json.Decode
+import Test exposing (Test, describe, test)
+import Yaml
+
+
+{-| Parsing from a JSON string keeps these cases readable and exercises the
+same path the API response takes.
+-}
```
taylorsilva:
```suggestion
```

web/elm/src/Pipeline/Pipeline.elm
```diff
@@ -531,9 +534,67 @@ view session model =
...
+                    )
+                :: onMouseEnter (Hover <| Just TopBarInfoIcon)
+                :: onMouseLeave (Hover Nothing)
+                :: Styles.infoIcon
+            )
+            [ Icon.icon
+                { sizePx = 20, image = Assets.InformationOutlineIcon }
+                -- 17px each side of a 20px icon fills the 54px top bar, so
+                -- the border-left reads as a full-height divider like the
+                -- favourite and pause controls
```
taylorsilva:
```suggestion
```
Should add a test for this in: web/elm/tests/PipelineTests.elm
There are tests for the other icons in there already.
Removing or changing `margin: 17px` should fail the test.

web/elm/src/PipelineInfo/Styles.elm
```diff
@@ -0,0 +1,58 @@
...
+body =
+    [ style "max-width" "800px"
+    , style "padding" "20px"
+    , style "color" Colors.dashboardText
+    , style "font-size" "14px"
+    , style "line-height" "1.5"
+
+    -- the markdown and YAML blocks are siblings; a gap keeps them apart
+    -- without either needing a margin that would double up on the padding
+    -- when it's the only child
```
taylorsilva:
```suggestion
```

Add a test for this instead in: web/elm/tests/PipelineInfoTests.elm
So when `line-height: 1.5` is removed, the test should fail.

taylorsilva (changes requested):
New day, fresh eyes and mind!

I realized there is some untested behaviour for the changes made in `Pipeline/Pipeline.elm`. Tests should be added in `web/elm/tests/PipelineTests.elm`:

- Add tests that verify the info icon is present when `user_data` is present, and the inverse
- The info icon has `margin: 17px` set on it. A test should fail if this is changed/removed

You can refer to the existing icon tests for examples: https://github.com/concourse/concourse/blob/9f3f88fcf88bff2972865905ae8f67d7fb648cf1/web/elm/tests/PipelineTests.elm#L751


In `web/elm/tests/PipelineInfoTests.elm` add a test that ensures `line-height` is set to create the gap between the markdown and yaml blocks.

Why leave comments when you can have the code + tests do all the explaining :)

blarghmatey:
Tests added in afb8dd1dd for all three asks.

`web/elm/tests/PipelineTests.elm` — new `top bar info icon` describe, next to the existing star-icon tests:

- `is shown when the pipeline has user_data` / `is hidden when the pipeline has no user_data`
- `links to the pipeline's info page`
- `icon has the same 17px margin as its neighbours` — asserts `margin: 17px` alongside `iconSelector` for the 20px `information-outline` image

Three of these existed but lived in `PipelineInfoTests.elm`, which was the wrong home for `Pipeline/Pipeline.elm` behaviour. Moved rather than duplicated, so there's one place to look.

`web/elm/tests/PipelineInfoTests.elm` — `the markdown and YAML blocks are separated by a gap`, asserting `display: flex`, `flex-direction: column`, `gap: 20px`, and `line-height: 1.5` on the info page body. The body div picked up an `id` (`pipeline-info-body`) so the test has something to target. `gap` is what actually separates the two blocks; `line-height` is pinned too since you named it.

Confirmed each assertion fails against unpinned code rather than just passing against the current one:

| change | failing test |
|---|---|
| `margin: 17px` → `10px` | `icon has the same 17px margin as its neighbours` |
| drop `gap: 20px` and `line-height: 1.5` | `the markdown and YAML blocks are separated by a gap` |

Reverted both after checking. Full suite: 3086 pass, `elm-analyse` reports 0 messages.

On the comment nit — your `nit: remove comments` commit already took care of it, and I've left it that way.

taylorsilva (changes requested):
Tooltip doesn't work. Hovering over the info icon prints a warning to the console.

Here's a patch that will fix it and add tests. Includes a bonus wiring up that we missed:

```patch
diff --git a/web/elm/src/Application/Application.elm b/web/elm/src/Application/Application.elm
index 68a5e12a7..27170b32b 100644
--- a/web/elm/src/Application/Application.elm
+++ b/web/elm/src/Application/Application.elm
@@ -666,6 +666,9 @@ routeMatchesModel route model =
         ( Routes.Dashboard _, SubPage.DashboardModel _ ) ->
             True
 
+        ( Routes.PipelineInfo _, SubPage.PipelineInfoModel _ ) ->
+            True
+
         _ ->
             False
 
diff --git a/web/elm/src/Message/Effects.elm b/web/elm/src/Message/Effects.elm
index 63d227c85..24f511f27 100644
--- a/web/elm/src/Message/Effects.elm
+++ b/web/elm/src/Message/Effects.elm
@@ -943,6 +943,9 @@ toHtmlID domId =
         TopBarPinIcon ->
             "top-bar-pin-icon"
 
+        TopBarInfoIcon ->
+            "top-bar-info-icon"
+
         CommentBar id ->
             "comment-bar-" ++ toHtmlID id
 
diff --git a/web/elm/tests/PipelineTests.elm b/web/elm/tests/PipelineTests.elm
index 15fd65879..6aa0b2d1f 100644
--- a/web/elm/tests/PipelineTests.elm
+++ b/web/elm/tests/PipelineTests.elm
@@ -895,6 +895,25 @@ all =
                                         , image = Assets.InformationOutlineIcon
                                         }
                                 )
+                , test "renders under the DOM id that hovering looks up" <|
+                    -- a DomID with no toHtmlID branch falls through to the
+                    -- catch-all "", so hovering calls getElementById("") and
+                    -- the tooltip never resolves
+                    \_ ->
+                        withUserData
+                            |> Common.queryView
+                            |> Query.has [ id <| Effects.toHtmlID TopBarInfoIcon ]
+                , test "hovering it asks for the viewport of that element" <|
+                    \_ ->
+                        withUserData
+                            |> Application.update
+                                (Msgs.Update <| Hover <| Just TopBarInfoIcon)
+                            |> Tuple.second
+                            |> Common.contains (Effects.GetViewportOf TopBarInfoIcon)
+                , test "hovering it shows the pipeline info tooltip" <|
+                    \_ ->
+                        withUserData
+                            |> Common.expectTooltip TopBarInfoIcon "pipeline info"
                 ]
             , describe "pipeline name tooltip" <|
                 let

```

blarghmatey:
Applied your patch verbatim in 014947bbf. Confirmed the `toHtmlID` branch is what makes it work: with that hunk reverted, "renders under the DOM id that hovering looks up" fails and the other two still pass, since they go through the `DomID` rather than the string.

No test for the `routeMatchesModel` addition. Both paths end at `PipelineInfo.init`, so a navigation test passes with or without the branch; the observable difference is only that `SubPage.urlUpdateValid`'s existing `Routes.PipelineInfo` case is now reachable.

3089 tests pass, `elm-analyse` reports 0 messages, `elm-format` is a no-op on the three files.

taylorsilva (approved):

PR #9666, opened 2026-07-28
worker-pinning: Add a flag to the job to allow all tasks to run on one worker

kallisti5:

This adds a flag named "pin_worker" to the job which instructs the planner to keep job tasks on the chosen worker for that buildrun.

(aka, once a job starts on a worker, keep all tasks on the selected worker)

Current behavior is tasks can run on "any worker" and stream data around (which may be an extremely poor choice if workers are in differing locations).

The workaround in the past has been abuse of tags["workera"] to keep pipelines rigid on one worker (which makes jobs less-reliable as they can't choose better workers)

## Notes to reviewer

This needs tested in a big way.   Definitely do not merge until it has been validated (i'm working on this now)

## Release Note

* Add ```pin_worker``` job flag to instruct the planner to keep all tasks of a job within the chosen worker. This can be helpful when workers are located in geographically diverse locations and streaming data between them is slow and unnecessary.

### DIFF
--- a/atc/exec/check_step.go
+++ b/atc/exec/check_step.go
@@ -27,6 +27,7 @@ type CheckStep struct {
 	resourceConfigFactory db.ResourceConfigFactory
 	noInputStrategy       worker.PlacementStrategy
 	checkStrategy         worker.PlacementStrategy
+	pinWorker             bool
 	delegateFactory       CheckDelegateFactory
 	workerPool            Pool
 	defaultCheckTimeout   time.Duration
@@ -58,6 +59,7 @@ func NewCheckStep(
 	containerMetadata db.ContainerMetadata,
 	noInputStrategy worker.PlacementStrategy,
 	checkStrategy worker.PlacementStrategy,
+	pinWorker bool,
 	pool Pool,
 	delegateFactory CheckDelegateFactory,
 	defaultCheckTimeout time.Duration,
@@ -71,6 +73,7 @@ func NewCheckStep(
 		workerPool:            pool,
 		noInputStrategy:       noInputStrategy,
 		checkStrategy:         checkStrategy,
+		pinWorker:             pinWorker,
 		delegateFactory:       delegateFactory,
 		defaultCheckTimeout:   defaultCheckTimeout,
 	}
@@ -195,7 +198,7 @@ func (step *CheckStep) run(ctx context.Context, state RunState, delegate CheckDe
 			ctx = lagerctx.NewContext(ctx, logger)
 		}
 
-		versions, processResult, runErr := step.runCheck(ctx, logger, delegate, imageSpec, resourceConfig, source, fromVersion)
+		versions, processResult, runErr := step.runCheck(ctx, logger, state, delegate, imageSpec, resourceConfig, source, fromVersion)
 		if runErr != nil || processResult.ExitStatus != 0 {
 			metric.Metrics.ChecksFinishedWithError.Inc()
 
@@ -250,6 +253,7 @@ func (step *CheckStep) run(ctx context.Context, state RunState, delegate CheckDe
 func (step *CheckStep) runCheck(
 	ctx context.Context,
 	logger lager.Logger,
+	state RunState,
 	delegate CheckDelegate,
 	imageSpec runtime.ImageSpec,
 	resourceConfig db.ResourceConfig,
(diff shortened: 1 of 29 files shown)


taylorsilva:
Just read the PR title and I'm SO HAPPY someone else did this before me! Your initial pain around this has been in the back of my head since you mentioned this in whatever issue it came up in.

kallisti5:
Yeah, I've been fighting our single concourse worker for a release and decided it was time.  Even if you didn't want the change, I was planning on a fork for us to leverage it :laughing:

taylorsilva:
No fork needed, will happily work with you here to get this in 😁

atc/exec/export_test.go
taylorsilva, line 1: Not a convention we follow. Feels like a structural smell...

atc/exec/check_step.go
```diff
@@ -343,3 +347,29 @@ func (step *CheckStep) containerOwner(delegate CheckDelegate, resourceConfig db.
 		expires,
 	)
 }
+
+// selectWorker picks a worker for the check step, honoring the build's pinned
+// worker (if the job has pin_worker: true). The first step to need a
+// worker sets the pinned worker; subsequent steps are forced to use the
+// same one. See selectStepWorker for the race-handling details.
+func (step *CheckStep) selectWorker(
```
taylorsilva, line 355: checks run in their own builds. You can't have a check step in a job. Doesn't make sense for a check step to run through the same code, does it? Might be fine to leave as-is but comment is terribly misleading.

taylorsilva (commented):
It's a lot of code to take in, but I'm getting the general shape of the change.

One thing I'm noticing is that we have to pass a `pinnedWorker` bool through a lot of the code. It's an extra input on a lot of funcs now.

We also had to add a `selectWorker()` func to every step type. This is not feeling very [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

When thinking about this feature, the approach I would have taken would have been to have the engine see that `pin_worker` is true and return a new struct, `worker.PinnedPool` or something that wraps the real `worker.Pool` that the engine usually passes into each step. I think this approach would reduce a lot of the duplicate calls we have going on right now.

The code you have in `atc/exec/worker_selector.go` could probably move into `atc/worker/`, becoming the code for `FindOrSelectWorkerOnPinned`. That function could be on the `worker.PinnedPool` struct I suggested, renamed and overriding `FindOrSelectWorker` from the underlying `worker.Pool`.

This `PinnedPool` struct could even live in `atc/engine/`. Might make more sense there?

I think that would result in a cleaner diff overall. `worker.Pool` would become an interface and we put all this scheduling stuff behind it. What do you think?

---

Edit:
- FYI, existing Pool interface: https://github.com/concourse/concourse/blob/9f3f88fcf88bff2972865905ae8f67d7fb648cf1/atc/exec/step.go#L73-L79

taylorsilva:
Unit tests are broken. Ran it again because I thought the timeout was just the worker overloaded. Looks like it gets stuck running the `atc/worker/` suite.

kallisti5:
I'll run over these changes.  Do you prefer clean commits (editing the existing commits for a clean commit) or additional commits making the recommended changes on top of the first commit?

taylorsilva:
I prefer clean commits. Thanks for asking.

ArjanSchouten:
Out of interest:
1. Does volume-locality strategy not work in this case?
2. Why did you decide to add an extra field to the job? Isn't adding a new placement strategy or ATC flag a cleaner approach?

taylorsilva:
> Out of interest:
> 
>     1. Does volume-locality strategy not work in this case?
> 
>     2. Why did you decide to add an extra field to the job? Isn't adding a new placement strategy or ATC flag a cleaner approach?

I think I can answer both of these as the reviewer. They're good questions.

tldr: Feature is similar to tags so it follows a similar implementation path outside of placement strategies.

1. Correct, it wouldn't really work. The placement strategies are still used, but since the worker gets chosen at the start of the job, volume-locality won't have any influence over which worker gets chosen since the build has zero volumes.
2. This feature doesn't make sense as a placement strategy unless we do a major overhaul of placement strategies that would allow for user/pipeline-level input. The only example of user/pipeline input affecting is `tags` which doesn't use a placement strategy for scheduling, it happens a layer above the placement strategies. I think that's also the right place for this feature to be implemented.
  This also isn't laid out in the PR description, but this feature is meant for ad-hoc situations where users know their job's workload ahead of time and know that volume streaming is expensive time-wise. Think of jobs that have volumes with multiple GB's of volumes.

PR #9668, opened 2026-07-31
fly: copy stdin buffer before sending hijack input

lodener:

* `stdinWriter.Write` handed `io.Copy`'s buffer straight to the channel that `handleInput` serializes from. `io.Copy` reuses a single 32KB buffer, so the next read could overwrite bytes that hadn't been written to the websocket yet, and the JSON that went out held whatever was in the buffer at marshal time. Fixed by cloning the slice in `Write`.
* Fixing it in `stdinWriter` rather than at the call site, since `io.Writer` says implementations must not retain `p`, so that's the side in the wrong.
* Added an integration test that writes 8MB to fly's stdin in one shot and checks the reassembled `HijackInput` frames match what was sent.

## Notes to reviewer

Typing never hits this because each read is one byte and the round trip finishes long before the next keystroke. Pasting does: the terminal delivers the payload in a burst, the reads come back to back, and the input can land in the container reordered and duplicated once the paste spans more than one read.

How often you see it depends on how your terminal feeds the pty. In my case, Ghostty and Alacritty on macOS push a paste in large writes with no gaps, which trips it easily. Terminals like iTerm that split a paste into smaller writes, or pace them (some rate limit paste input, and anything routing through tmux or ssh gets chunked along the way), leave enough time between reads for the previous frame to go out, so the corruption stays hidden. The race is there either way.

The fake ATC in the integration test sleeps for a second before reading input frames. That fills the socket buffers so fly's `WriteJSON` blocks, which is what opens the race window. Without it the race is 50/50, causing it to pass against the broken code.

To confirm: revert the `bytes.Clone` and the test fails on the byte comparison.

## Release Note

* Fixes corrupted input when pasting into `fly hijack` / `fly intercept`. Pasted text longer than a single terminal read could arrive at the container reordered or duplicated.

### DIFF
--- a/fly/commands/hijack.go
+++ b/fly/commands/hijack.go
@@ -1,6 +1,7 @@
 package commands
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -470,7 +471,7 @@ type stdinWriter struct {
 
 func (w *stdinWriter) Write(d []byte) (int, error) {
 	w.inputs <- atc.HijackInput{
-		Stdin: d,
+		Stdin: bytes.Clone(d),
 	}
 
 	return len(d), nil
--- a/fly/integration/hijack_test.go
+++ b/fly/integration/hijack_test.go
@@ -1,6 +1,7 @@
 package integration_test
 
 import (
+	"bytes"
 	"fmt"
 	"net/http"
 	"os/exec"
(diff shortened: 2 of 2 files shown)


taylorsilva (approved):
LGTM - Thank you so much for fixing this! Was also one of those minor things I'd notice while busy troubleshooting something else and would forget about.

PR #9651, opened 2026-07-22
add support for shared path in Conjur

VikranthBala:
Changes proposed by this PR

As of now, it is not possible to configure shared paths for Conjur secret manager as opposed to other cred managers such as Vault,AWS. So I propose to add a new parameter --conjur-shared-path similar to the Vault cred manager. This parameter can be optional (as with Vault). This PR provides all the changes needed for this functionality.
Notes to reviewer

Release Note

Added --conjur-shared-path to configure shared secret paths for Conjur secrets manager similarly to the one for Vault, SSM.

PR is for the issue: https://github.com/concourse/concourse/issues/9062

PR #4693, opened 2019-10-31
Conjur credential manager

AndrewCopeland:
# Why do we need this PR?
Conjur is a password vault that allows the ability to obtain secrets, this PR allows the ability to use the conjur password vault within concourse as a credential manager.

# Changes proposed in this pull request
> What are the changes included in this PR? Feel free to add as many lines as needed below.
* A new credential manager has been added.

# Contributor Checklist
> Are the following items included as part of this PR? Please delete checkbox items that don't apply.
- [ x] Unit tests
- [ ] Updated documentation (located at https://github.com/concourse/docs)
- [ x] Updated release notes (located at https://github.com/concourse/concourse/tree/master/release-notes)

# Reviewer Checklist
- [x] Code reviewed
- [x] Tests reviewed
- [ ] Documentation reviewed
- [ ] Release notes reviewed
- [ ] PR acceptance performed

PR #9675, opened 2026-08-04
Revert "add support for shared path in Conjur"

taylorsilva:
Reverts concourse/concourse#9651

Kevin noted here: https://github.com/concourse/docs/pull/616#discussion_r3700257850
That the existing config `CONCOURSE_CONJUR_SECRET_TEMPLATE` appears to be functional as a shared path. The only difference I see is in the default values. Actually it probably shouldn't even have a default value since that could catch an operator by disguise.

@AndrewCopeland You initially added this code https://github.com/concourse/concourse/pull/4693
I know you're probably very far removed from Concourse and potentially Conjur now too, but are you able to recall what the intention of `SecretTemplate` is: https://github.com/concourse/concourse/blob/0ccc29ab5ef3ddd8423609e3e0c3758df7b74b46/atc/creds/conjur/manager.go#L28

**Is `SecretTemplate` meant to be a "shared path" lookup?**

### DIFF
--- a/atc/creds/conjur/conjur.go
+++ b/atc/creds/conjur/conjur.go
@@ -15,15 +15,13 @@ type Conjur struct {
 	log             lager.Logger
 	client          IConjurClient
 	secretTemplates []*creds.SecretTemplate
-	sharedPath      string
 }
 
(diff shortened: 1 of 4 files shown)


taylorsilva:
My current plan is to revert this since it _looks_ like `SecretTemplate` functions as a shared path lookup.

PR #9626, opened 2026-07-03
web: instance groups: add list view

bcersows:
This commit adds a list view capability to the web frontend. It's available only for instance groups with a toggle similar to the HD one. Related to that, the former HD route was replaced with a session-stored one, so its setting along with the new list view one are restored on reload and independent of the route flow.

## Changes proposed by this PR

closes https://github.com/orgs/concourse/discussions/9608 (#9608)

_Make sure to follow all [PR requirements](https://github.com/concourse/concourse/blob/master/CONTRIBUTING.md#pull-request-requirements)._

* new list view capability for instance groups
  * can be toggled
  * still allows for dragging, etc
  * fully responsive
* the HD route was removed and warped into a session-stored value along with the new list view toggle status
  * main pro: is being restored now even after switching to details view and back

[...]

PR #9678, opened 2026-08-06
web: instance groups: add list view favorites

bcersows:
Extend the list view function with a favorites list similar to the grid view.

Added a favorites list similar to the grid view's.

Follow-up to https://github.com/concourse/concourse/pull/9626 and as discussed in https://github.com/orgs/concourse/discussions/9608#discussioncomment-17895632.

image

### DIFF
--- a/web/elm/src/Dashboard/Dashboard.elm
+++ b/web/elm/src/Dashboard/Dashboard.elm
@@ -1441,8 +1441,61 @@ cardsView session params teamCards =
             params.highDensity && not noPipelines
 
         ( headerView, offsetHeight ) =
-            if (highDensity && not viewingInstanceGroups) || (params.groupListView && viewingInstanceGroups) then
+            if highDensity && not viewingInstanceGroups then
                 ( [], 0 )
+
+            else if params.groupListView && viewingInstanceGroups then
+                let
+                    favoritedSections =
+                        teamCards
+                            |> List.map
+                                (\section ->
+                                    { section
+                                        | cards =
+                                            List.filter
+                                                (\c ->
+                                                    case c of
+                                                        PipelineCard p ->
+                                                            Favorites.isPipelineFavorited session p
+
+                                                        InstancedPipelineCard p ->
+                                                            Favorites.isPipelineFavorited session p
+
+                                                        InstanceGroupCard p _ ->
+                                                            Favorites.isInstanceGroupFavorited session (Concourse.toInstanceGroupId p)
+                                                )
+                                                section.cards
+                                    }
+                                )
+                            |> List.filter (\s -> not (List.isEmpty s.cards))
+
+                    allPipelinesHeader =
+                        Html.div Styles.pipelineSectionHeader [ Html.text "all pipelines" ]
+                in
+                if List.isEmpty teamCards then
+                    ( [], 0 )
+
+                else if List.isEmpty favoritedSections then
+                    ( [ allPipelinesHeader ], 0 )
+
+                else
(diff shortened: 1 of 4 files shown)


bcersows:
Hi @taylorsilva , there's again a failed build I'm not able to access. Could you enlighten me? :)

taylorsilva:
validate labels looks for certain labels: `bug`, `misc`, `enhancement`, `breaking`

I made the pipeline public so you should be able to see any future errors.

taylorsilva (changes requested):
Playing around with this in UI first. Works well and as expected!

- Favourites are surfaced at both the instance and dashboard levels
- Favourites work in both grid and list views at the instance level

The only visual feedback I have is to remove one of the pipeline name headers. With this being displayed twice it takes up a lot of vertical screen space:

Image

One other thing I just noticed, which I don't think was introduced by this PR because I see a similar result in the current release, is that there seems to be more padding at the top of the instance view page. Here's a video showing the area I'm talking about. I show it on my local instance first and then again in ci.concourse-ci.org:

https://github.com/user-attachments/assets/05abb225-b9e6-4a0a-b5b1-1e08acd7d76e

Going back to 8.2.x we don't see the same shifting of the UI when clicking into an instance group:

https://github.com/user-attachments/assets/0c99f66c-d317-4a66-9ad8-393e31503aa8

Would be nice if we could fix that up here :)

bcersows:
Hi!

For the padding I've got a fix (not pushed yet), stupid little thingy.

As for the header.. I could do that, but is that what we (you 😄) really want? It's not a regular user flow, I'd say, but in theory it's possible that the user manually updates the filter and deletes the `team` one, leaving only the `group` filter. And iiif there's the same group across multiple teams (maybe using teams for environment separation, I dunno) that could be confusing. See the following screenshot with two duplicated groups across teams:

image

If that kinda constructed edge case is fine for you, I'll throw it out, no biggie. :)

taylorsilva:
Interesting call out! I didn't consider multiple teams with the same pipeline instance groups. I definitely think that's a bit of an edge case since a user would have to manually change the search bar's value. I like seeing when little edge cases like that work though.

You convinced me, leave it as-is! I appreciate the push back :)

Hope that header padding issue isn't giving you too much trouble! 🙏

bcersows:
Fixed the padding issue. Also decreased the `margin-bottom` a bit for the team name header thingy, so at least it's a bit less wasted space now. So looking like this now:
image

My IDE warned me that two styles are unused as of now, `instanceGroupCardBadge` and `instanceGroupCardBannerHd`, want me to get rid of them while I'm at it?

The integration test seemed to have failed with my last merge commit push; any idea why?

taylorsilva:
> My IDE warned me that two styles are unused as of now, instanceGroupCardBadge and instanceGroupCardBannerHd, want me to get rid of them while I'm at it?

Clean up like that would be appreciated! I prefer removing dead code when possible.

> The integration test seemed to have failed with my last merge commit push; any idea why?

Unrelated to your PR; worker's recycled and there's some weird issue with that test suite whenever the workers restart. Retriggering, it should pass.

bcersows:
Those styles are gone. :)

taylorsilva (approved):
I rebased the PR branch on latest `master` just to remove the merge commit that was on the branch. I've also added a few tests to cover the new code.

Played around with it in the UI and it looks great now! Will merge after the latest checks complete.

bcersows:
Thanks for the final touches!

PR #9681, opened 2026-08-07
Fix flaky timing in containerd and vault integration tests (#9669)

### LINKED ISSUE #9669: Flaky CI tests: TestNewContainerEnforcesTimeoutOnTask and TestVaultV2WithUnmountPath
## Summary

Two tests in the CI suite fail intermittently on unchanged code. Both surfaced on a frontend-only PR ([#9661](https://github.com/concourse/concourse/pull/9661)) whose diff contains no Go at all, which makes the evidence unusually clean: the same Go code passed and failed the same job across pushes.

Filing this separately from the PR since it's independent of that change.

### 1. `TestSuite/TestNewContainerEnforcesTimeoutOnTask` — `containerd-integration`

`worker/runtime/integration/integration_test.go:1293`

```
Error: "no running task found: task c74ee145-f777-4462-a48c-673be99c6739 not found" does not contain "deadline exceeded"
Test:  TestSuite
Messages: Expected 'context deadline exceeded', got: no running task found: task ... not found
```

The test `SIGSTOP`s the containerd process and then asserts that a `Task()` call fails with `context deadline exceeded` within ~1s (`requestTimeout`). When it flakes, the call instead returns the ordinary "task not found" error — i.e. it completed before the freeze took effect. The other 14 tests in the suite passed in the same run.

Results across pushes on PR #9661, with byte-identical Go:

| build | result |
|---|---|
| [571370468](https://ci.concourse-ci.org/builds/571370468) | pass |
| [571370698](https://ci.concourse-ci.org/builds/571370698) | pass |
| [571371932](https://ci.concourse-ci.org/builds/571371932) | **fail** |
| next push | pass |

### 2. `TestVaultV2WithUnmountPath` — `integration`

`integration/creds/vault_test.go:99`, via `integration/internal/flytest/cmd.go:88`

```
Error: Condition never satisfied
Test:  TestVaultV2WithUnmountPath
Messages: should have a running worker
--- FAIL: TestVaultV2WithUnmountPath (100.77s)
```

The test stands up docker-compose (vault + db + web + worker) and waits for a worker to register; it timed out at 100s. Build [571383233](https://ci.concourse-ci.org/builds/571383233).

Results on PR #9661:

| commit | result |
|---|---|
| `1aecf591` | pass |
| `71a065b` | pass |
| `21c5044` | **fail** |

The diff between the passing `71a065b` and the failing `21c5044` is six `.elm` files and zero `.go` files — a route constructor changed from a single-field record to a positional argument, and a one-field type alias removed. Nothing in that diff is reachable from Go worker registration.

## Steps to Reproduce

Both are intermittent, so there's no deterministic reproduction. They reproduce by re-running the affected jobs against unchanged commits — which is how they were identified here.

## Expected Results

Both tests pass deterministically, or are marked/retried if they depend on timing that CI can't guarantee.

## Actual Results

Both fail intermittently, turning the PR check bar red on changes that cannot affect them.

## Notes

Both look timing-dependent by construction rather than genuinely broken:

- the containerd one races a `SIGSTOP` against a 1s timeout;
- the Vault one waits a fixed 100s for a worker to register in a four-container docker-compose stack.

Worth noting for outside contributors: Concourse CI check runs have no re-run affordance from the GitHub UI, so the only way to clear a flake is to push a new commit (PR #9661 has an empty "Retrigger CI" commit for exactly this reason). That makes these more disruptive than they'd otherwise be.

## Concourse Version

`master` (`0.0.0-dev`), CI as of 2026-07-30.

djedi-knight:
## What

Fixes two intermittently-failing integration tests called out in #9669. Both
are timing races in the tests themselves, not product bugs.

### `TestNewContainerEnforcesTimeoutOnTask` (containerd-integration)

The test `SIGSTOP`s containerd and expects the next `Task()` call to hang and
hit the 1s request timeout (`context deadline exceeded`). But `SIGSTOP` is
delivered asynchronously — containerd could service one more gRPC request and
return a fast `no running task found ... not found` before the kernel actually
stopped it, so the assertion saw the wrong error.

Fix: after sending `SIGSTOP`, poll `/proc/<pid>/stat` and wait until the process
reports the stopped state (`T`) before starting the timing assertion. The call
then races the request timeout, not signal delivery.

### `TestVaultV2WithUnmountPath` (integration)

`WaitForRunningWorker` gave a four-container docker-compose stack
(db + web + worker, plus vault) only 60s to register a worker. On a loaded CI
runner cold-start can exceed that, producing `should have a running worker`.

Fix: raise the ceiling in the shared `WaitForRunningWorker` helper to 3 minutes.
A healthy worker still returns within seconds — the wider ceiling only absorbs
slow startup under load.

## Testing

`go vet` and `gofmt` pass on both changed packages. The tests themselves require
a live containerd and a docker-compose stack, so they were not executed in this
change; the fixes are deterministic by construction.

Fixes #9669

### DIFF
--- a/fly/integration/suite_test.go
+++ b/fly/integration/suite_test.go
@@ -1,9 +1,16 @@
 package integration_test
 
 import (
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"fmt"
+	"math/big"
+	"net"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


taylorsilva:
Thanks for the PR. I ended up pushing some other changes that now conflict with this PR. I re-did everything over in #9688 so I'll close this PR.

PR #9684, opened 2026-08-13
Widen volumes and containers id columns to bigint

### LINKED ISSUE #9672: Migrate other ID columns from int to bigint
Customer ran into an issue where web nodes started returning:

```
ERROR: integer out of range (SQLSTATE 22003)
```

This was quickly traced back to the `volumes` table `id` column. This column has type `integer`: https://github.com/concourse/concourse/blob/439d59e521dfc5db83a75aa15e8b6e75d5197231/atc/db/migration/migrations/1510262030_initial_schema.up.sql#L637-L638

The contents of this table don't grow, there aren't 2+ billion records in it. Each new record does cause the ID to increase and it can't increase forever. This customer had officially made 2,147,483,647 volumes! They've been running on this particular database since 2018.

The fix for this is straight forward:

```sql
ALTER TABLE volumes ALTER COLUMN id TYPE bigint
```

I want to add migrations for this and probably the `containers` table as well which has a similar column. I'll review other tables that may potentially face this issue and migrate their ID columns as well.

djedi-knight:
Fixes #9672.

`volumes.id` and `containers.id` are backed by ever-incrementing sequences and can exhaust the 32-bit `integer` range (2,147,483,647) on high-churn deployments even though row counts stay small, causing web nodes to fail with `ERROR: integer out of range (SQLSTATE 22003)`.

This migrates `volumes.id` and `containers.id` — and the columns that reference them through a foreign key — from `integer` to `bigint`:

- `containers.id`, `containers.image_check_container_id`, `containers.image_get_container_id`
- `volumes.id`, `volumes.parent_id`, `volumes.container_id`

Only the `volumes` and `containers` tables are rewritten. The `volumes` composite self-FK `(parent_id, parent_state)` → `(id, state)` is dropped and recreated around the `ALTER`s — otherwise altering the `id` side fails with `could not find cast from volume_state to anyenum` — and the `volumes(id, state)` unique constraint is preserved throughout.

### Scope

Following review feedback, this is intentionally limited to the two high-churn tables named in the issue. Other sequence-backed `id` columns can overflow in principle, but widening them rewrites large tables (e.g. `builds`) under an `ACCESS EXCLUSIVE` lock, so they're better discussed and sequenced separately. Happy to open a follow-up issue to work through which additional tables warrant it.

### Notes

- `volumes`/`containers` are themselves large, churny tables, so the `ALTER ... TYPE` rewrites them under an exclusive lock; on very large deployments this can take a while. Let me know if this should ride along with a major version bump / release note.
- The down migration reverts to `integer` and will (expectedly) fail if a live id already exceeds the int range — the standard limitation when narrowing back.

### Testing

Verified an up → down → up round-trip through Concourse's own migrator on PostgreSQL 16: the migration history records every step `passed`, the six columns flip `integer` ↔ `bigint` in both directions, and the `volumes` self-FK and `volumes(id, state)` unique constraint stay intact at each step. `go build ./atc/db/...` and `go vet ./atc/db/...` pass.

### DIFF
--- a/atc/db/migration/migrations/1786574734_widen_volumes_and_containers_ids_to_bigint.down.sql
+++ b/atc/db/migration/migrations/1786574734_widen_volumes_and_containers_ids_to_bigint.down.sql
@@ -1,0 +1,19 @@
+-- Reverse of the volumes/containers id-widening migration: bigint -> integer.
+-- This can fail if a live sequence value already exceeds the 32-bit integer
+-- range, which is the expected/standard limitation for narrowing back to
+-- integer.
+ALTER TABLE volumes DROP CONSTRAINT volumes_parent_id_fkey;
+
+ALTER TABLE containers
+    ALTER COLUMN id TYPE integer,
+    ALTER COLUMN image_check_container_id TYPE integer,
+    ALTER COLUMN image_get_container_id TYPE integer;
+
+ALTER TABLE volumes
+    ALTER COLUMN id TYPE integer,
+    ALTER COLUMN parent_id TYPE integer,
+    ALTER COLUMN container_id TYPE integer;
+
+ALTER TABLE volumes ADD CONSTRAINT volumes_parent_id_fkey
+    FOREIGN KEY (parent_id, parent_state)
+    REFERENCES volumes (id, state) DEFERRABLE;
--- a/atc/db/migration/migrations/1786574734_widen_volumes_and_containers_ids_to_bigint.up.sql
+++ b/atc/db/migration/migrations/1786574734_widen_volumes_and_containers_ids_to_bigint.up.sql
@@ -1,0 +1,27 @@
+-- Widen the volumes and containers id columns (and the columns that reference
+-- them through a foreign key) from integer to bigint, so their high-churn
+-- sequences cannot overflow the 32-bit integer range and produce
+-- "ERROR: integer out of range (SQLSTATE 22003)". See
+-- https://github.com/concourse/concourse/issues/9672.
+--
+-- Only the volumes and containers tables are rewritten by this migration.
+--
+-- The volumes composite self-foreign-key (parent_id, parent_state) ->
+-- (id, state) is dropped first: altering the id side otherwise fails with
+-- "could not find cast from volume_state to anyenum". The volumes(id, state)
+-- unique constraint it depends on is preserved throughout.
+ALTER TABLE volumes DROP CONSTRAINT volumes_parent_id_fkey;
+
+ALTER TABLE containers
+    ALTER COLUMN id TYPE bigint,
+    ALTER COLUMN image_check_container_id TYPE bigint,
+    ALTER COLUMN image_get_container_id TYPE bigint;
+
+ALTER TABLE volumes
+    ALTER COLUMN id TYPE bigint,
+    ALTER COLUMN parent_id TYPE bigint,
+    ALTER COLUMN container_id TYPE bigint;
+
+ALTER TABLE volumes ADD CONSTRAINT volumes_parent_id_fkey
+    FOREIGN KEY (parent_id, parent_state)
+    REFERENCES volumes (id, state) DEFERRABLE;


atc/db/migration/migrations/1786574734_migrate_remaining_ids_to_bigint.down.sql
```diff
@@ -0,0 +1,150 @@
...
+
+ALTER TABLE build_image_resource_caches
+    ALTER COLUMN job_id TYPE integer,
+    ALTER COLUMN resource_cache_id TYPE integer;
+
+ALTER TABLE build_resource_config_version_inputs ALTER COLUMN resource_id TYPE integer;
+
+ALTER TABLE build_resource_config_version_outputs ALTER COLUMN resource_id TYPE integer;
+
+ALTER TABLE builds
```
IvanChalukov: I’m a bit hesitant to migrate the columns from build and build_resource_config_version_inputs, considering that they contain millions of rows. Similar migrations have caused a lot of pain in past releases.

taylorsilva (changes requested):
This PR looks like a very naive "find every ID column using int and change it to bigint". Changing the `component` table is pointless, it has a static list of components currently at 20 items. I wouldn't have considered touching that table at all since it'll never have 2+ billion records inserted into it over its lifetime. It's a similar story for the `base_resource_types` table.

I was mostly thinking of updating the `volumes` and `containers` tables, the two that have a lot of churn (adding/deleting records) but never a lot of individual records at any given moment. I wasn't planning to update every table that has an `int` for its ID type. I said as much in the issue:

> I want to add migrations for this and probably the containers table as well which has a similar column. I'll review other tables that may potentially face this issue and migrate their ID columns as well.

I'd prefer to take a cautious approach here and avoid touching the tables that grow a lot like the `builds` table. Ivan is right that tables like `builds` can be **massive** in a production database. Doing a migration on that table can easily take hours on large databases. Whenever we include such large impact migrations we usually end up doing a major bump. I wouldn't expect to do such a migration until we want to make a v9 release of Concourse. I know, kinda weird, but it's been the project's convention for the last few versions.

As a general collaboration thing, before opening this PR that has a much wider scope than the referenced issue, I would have preferred discussing which other tables to include within the issue. Everyone has this trigger-happy sense of wanting to fire off PR's and THEN have a discussion about the scope of the PR/issue. That's putting the "cart before horse".

In its current state I won't accept this PR. Its scope is much wider than the referenced issue and it doesn't seem like a careful evaluation of which tables to modify has been done. If you want to re-scope this PR to align with the issue you are free to do so.

I will happily review a PR that modifies the `volumes` and `containers` tables. Other tables can also be updated if they're deemed to have a low migration impact, similar to scale as the `volumes` and `containers` tables.

djedi-knight:
Thanks for the review, and fair point - apologies for jumping straight to a broad PR instead of discussing scope first.

I've rescoped this down to just `volumes` and `containers`, which is what the issue was about and where the churn actually is. The migration now widens only `volumes.id` and `containers.id` plus the columns that reference them via FK (`containers.image_check_container_id`, `containers.image_get_container_id`, `volumes.container_id`, `volumes.parent_id`), so **only the `volumes` and `containers` tables are rewritten** - nothing touches `builds`, `build_resource_config_version_inputs`, or static tables like `components`/`base_resource_types` any more. I also dropped the `lock.go` change, since it only mattered for the `jobs`/`resource_configs` ids that are no longer in scope.

`volumes`/`containers` are themselves large, churny tables, so the `ALTER ... TYPE` will still rewrite them under an exclusive lock - let me know if you'd like this to ride along with a version bump / release note and I'll adjust.

If it's useful, I'm happy to open a separate issue to discuss which other high-churn tables (resource caches, `worker_*`, etc.) might warrant the same treatment on their own merits, rather than bundling them here.

Force-pushed the rescoped version.

atc/db/migration/migrations/1786574734_migrate_remaining_ids_to_bigint.down.sql
djedi-knight: Please review new changes if possible

taylorsilva:
Thanks for updating it. I started reviewing it and had some suggestions/questions, but realized I didn't want to relay all this through a meat-proxy. I did also want to update other similar tables, but again don't want to push that exercise through a meat-proxy.

I've opened my own PR #9691 which resolves the issue. My head was in the problem space from reviewing this so I just tackled it on my own. Cheers! 🍻

djedi-knight:
@taylorsilva you could just say thanks w/o trying to be a turd about it, but I guess to each its own.  This comment brought to you straight from the meat-source ;)

PR #9686, opened 2026-08-16
Support quoted SAML groups containing commas in main team env var

### LINKED ISSUE #8597: Unable to specify main team saml group descriptions which contain the comma character
## Summary

I'm unable to specify the `CONCOURSE_MAIN_TEAM_SAML_GROUP` flag if the group name I want to use contains a `,` character.

https://github.com/concourse/dex/blob/master/connector/saml/saml.go#L424 The saml group delimiter is ',' and there is no logic for escaping the character or quoting it so that you can include it in the group name.

In our setup we end up using LDAP group descriptions which contain commas..
e.g.

`CN=my_concourse_admin,OU=SecurityGroups,DC=example,DC=com`

This group name works when using `fly set-team -c teamconfig.yaml` where teamconfig.yaml contains the following:

```yaml
roles:
- name: owner
  saml:
    groups: ["CN=my_concourse_admin,OU=SecurityGroups,DC=example,DC=com"]
```

## Steps to reproduce

set the MAIN_TEAM_SAML_GROUP flag using an environment variable:

` CONCOURSE_MAIN_TEAM_SAML_GROUP='CN=my_concourse_admin,OU=SecurityGroups,DC=example,DC=com`

Connect to saml in an environment where the groups attr passed to concourse is formatted like this:

```
[\"CN=my_concourse_publish,OU=SecurityGroups,DC=example,DC=com\"
\"CN=my_concourse_admin,OU=SecurityGroups,DC=example,DC=com\"
]
```
etc.

Attempt to log in using the main team.

## Expected results

`CONCOURSE_MAIN_TEAM_SAML_GROUP='CN=my_concourse_admin,OU=SecurityGroups,DC=example,DC=com'`
should show the following when running `fly get-team -n main`
`saml:cn=my_concourse_admins,ou=securitygroups,dc=example,dc=com`

## Actual results

Each comma delimits a separate group and the main team config becomes
`saml:cn=my_concourse_admins,saml:ou=securitygroups,saml:dc=example,saml:dc=com`

## Additional context

As a workaround: setting the group using a flag seems to work. `concourse web --main-team-saml-group=CN=my_concourse_admin,OU=SecurityGroups,DC=example,DC=com`

## Triaging info

* Concourse version: v7.6.0
* Did this used to work? no (but works using `fly set-team`)

mubashir1osmani:
Fixes #8597.

The environment adapter treats slice flags as comma-delimited, which breaks LDAP-style SAML group DNs like `CN=my_concourse_admin,OU=SecurityGroups,DC=example,DC=com`.

An earlier revision of this PR declared an explicit env key for the flag, which stopped comma-splitting entirely — a breaking change for anyone setting multiple groups via `CONCOURSE_MAIN_TEAM_SAML_GROUP` (they'd be locked out of the main team). This revision takes a different, backwards-compatible approach:

- The env var keeps its default comma-split behavior — **no change for existing clusters**.
- A group **wrapped in double quotes** is rejoined after the split and the quotes stripped, so commas inside it are preserved:

  ```
  CONCOURSE_MAIN_TEAM_SAML_GROUP='"CN=my_concourse_admin,OU=SecurityGroups,DC=example,DC=com"'
  ```

  yields a single group. Quoted and unquoted groups can be mixed. Quoting is opt-in, and no valid group name starts with a `"`, so existing configurations parse identically.
- An unclosed quote leaves the values untouched rather than erroring.

Tests are now part of the existing Ginkgo suite and cover the existing comma-split behavior, quoted DNs, mixed quoted/unquoted groups, unclosed quotes, and the `--main-team-saml-group` flag path.

Tested with:
- `go test ./skymarshal/skycmd -count=1`


skymarshal/skycmd/saml_flags_test.go
taylorsilva, line 1: Existing test suite uses Ginkgo: https://github.com/concourse/concourse/blob/master/skymarshal/skycmd/skycmd_suite_test.go

Please add the tests to the existing Ginkgo test suite.

taylorsilva (changes requested):

skymarshal/skycmd/saml_flags.go
```diff
@@ -74,7 +74,7 @@ func (flag *SAMLFlags) Serialize(redirectURI string) ([]byte, error) {
 
 type SAMLTeamFlags struct {
 	Users  []string `json:"users" long:"user" description:"A whitelisted SAML user" value-name:"USERNAME"`
-	Groups []string `json:"groups" long:"group" description:"A whitelisted SAML group" value-name:"GROUP_NAME"`
+	Groups []string `json:"groups" long:"group" env:"CONCOURSE_MAIN_TEAM_SAML_GROUP" description:"A whitelisted SAML group" value-name:"GROUP_NAME"`
```
taylorsilva, line 77: Need to think if this is safe to do for existing clusters. Will this break anything? We do some env wrangling too with `CONCOURSE_`, like in the `quickstart` command we add `CONCOURSE_WORKER_`. I think all the web commands still get the regular prefix in that situation.

taylorsilva (changes requested):
Locally tested this out to see what the behaviour was. It does end up parsing the entire env value as one string. The problem is that it's a breaking change. Anyone that was setting this env var expecting it to break on commas now wouldn't get that and would be locked out of the main team.

This fixes the issue's use case, but breaks the existing use case we currently support.

This PR cannot be merged for that reason, along with the weird test.

skymarshal/skycmd/saml_flags_test.go
mubashir1osmani: Done — the standalone test is replaced with Ginkgo specs in the existing SkyCmd suite.

skymarshal/skycmd/saml_flags.go
mubashir1osmani: Good catch — the explicit env key was indeed breaking for anyone setting multiple groups via the env var. I've reworked the fix: the env tag is gone, so comma-splitting (and the quickstart/prefix wrangling) behaves exactly as before. Instead, `GetGroups()` rejoins any group wrapped in double quotes after the split, so `CONCOURSE_MAIN_TEAM_SAML_GROUP='"CN=admins,OU=Groups,DC=example,DC=com"'` yields one group. Quoting is opt-in and no valid group name starts with a quote, so existing configs parse identically.

taylorsilva: Cool! That might work as a workaround for this and avoids the breaking change issue too. I'll review this when I have time.

PR #9689, opened 2026-08-22
Updates Running Concourse section in Docs

bespokeeagle:

* Update CONTRIBUTING.md to instruct contributors to build the image before running docker compose up.

## Notes to reviewer

This avoids a race condition where the web and worker services build and tag the same image simultaneously, causing the build to fail.


taylorsilva (approved):
Thanks, I had noticed this too and never got around to updating this doc. I think there are a few other spots this could be updated as well but I'll do those on main.

PR #9693, opened 2026-08-29
 idtoken: emit a jti claim

### LINKED ISSUE #9692: idtoken var source: emit a jti claim so relying parties can do replay prevention
### Summary

The `idtoken` credential provider does not put a `jti` (JWT ID) claim into the tokens it mints. Relying parties that implement replay prevention for JWT bearer assertions key it on `jti`, so with Concourse tokens they cannot do it at all, and a leaked token stays replayable until `exp`. Adding a random `jti` per generated token would let them.

#### Why this matters

RFC 7523 section 3 defines `jti` as the claim a relying party uses to reject a replayed assertion, and it is what implementations key their single-use caches on.

The concrete case we hit is Keycloak's federated client authentication (Keycloak 26.6 and later), where a Concourse `idtoken` is presented as a `client_assertion` to authenticate an OAuth client.

### Context

Discussion: https://github.com/orgs/concourse/discussions/9690

baloo42:

Closes #9692 

* idtoken: emit a `jti` claim
* idtoken: advertise `jti` in the OpenID discovery document

## Notes to reviewer

Discussion: https://github.com/orgs/concourse/discussions/9690

A config knob is deliberately not implemented. This was discussed in the above discussion but rejected.

### DIFF
--- a/atc/api/idtokenserver/openid_discovery.go
+++ b/atc/api/idtokenserver/openid_discovery.go
@@ -20,7 +20,7 @@ func (s *Server) OpenIDConfiguration(w http.ResponseWriter, r *http.Request) {
 	}{
 		Issuer:                           s.oidcIssuer,
 		JWKSUri:                          s.oidcIssuer + "/.well-known/jwks.json",
-		ClaimsSupported:                  []string{"aud", "iat", "iss", "sub"},
+		ClaimsSupported:                  []string{"aud", "iat", "iss", "jti", "sub"},
 		ResponseTypesSupported:           []string{"idtoken"},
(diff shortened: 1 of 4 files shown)


taylorsilva:
Final decision regarding adding a knob: don't want a knob, always include a `jti`.

Mark the PR as ready to review at your leisure. TY!

taylorsilva (approved):
I pushed a commit to change how the uuid is generated. Using `uuid.NewString()` could result in a panic if there's an error generating the uuid. Everywhere else we use uuid in the code we instead call `uuid.NewRandom()` to catch and handle the error ourselves.

Otherwise this LGTM! I'll merge once the PR checks pass.

PR #9695, opened 2026-08-30
worker: keep in-flight overlay dirs during orphan sweep

### LINKED ISSUE #9685: worker: volume sweeper removes overlay dirs of volumes being created, breaking container create
## Summary

The volume sweeper can delete the overlay dirs of a volume that is being created at that moment. The volume stays mounted, but creating a file in it then fails with ENOENT, so the container create on that volume dies with:

```
find or create container on worker <id>: starting task: new task: failed to create shim task: OCI runtime create failed:
runc create failed: unable to start container process: error during container init: error mounting
"/usr/local/concourse/bin/init" to rootfs at "/tmp/gdn-init": create mountpoint for /tmp/gdn-init mount:
make mountpoint "/tmp/gdn-init": openat gdn-init: no such file or directory
```

We get this about 4 times a day. Most hits are resource check containers, which run again on their own, but a hit on a step of a job errors the whole build.

## Steps to Reproduce

Hard to trigger on purpose, the window is a few milliseconds per sweep. We ran the worker at debug level and captured 2 minutes of log around each failure. In 8 captures, all 9 failures were preceded by an orphan removal of the same handle, out of ~1200 `create-cow` with only 10 orphan removals in total.

One of them:

```
02:08:26.141 volume-server.cleanup-orphaned-volumes.start
02:08:26.143 overlay-driver.creating-overlay-mount        upperdir=/worker-state/overlays/27c32ab5-...
02:08:26.144 overlay-driver.removing-orphaned-overlay-work-dir  handle=27c32ab5-...
02:08:26.144 volume-server.cleanup-orphaned-volumes.done
02:08:27.966 garden-server.create.failed                  openat gdn-init: no such file or directory
```

`CleanupOrphanedEntries()` reads `live/` and `init/` into `knownHandles`, then `RemoveOrphanedResources()` lists `overlays/` and `overlays/work/` and removes everything that is not in that snapshot. Two ways to fall through it:

* creation: `initRawVolume()` creates `init/<handle>` before the driver creates the overlay dirs, so a volume created after the snapshot is missing from it and loses its work dir.
* promotion: `Initialize()` renames `init/<handle>` to `live/<handle>`. A snapshot taken during the rename finds the handle in neither dir, and the whole layer is removed.

We see both: 4 of the 9 are work-dir removals with the overlay mount inside the sweep window, 5 are layer removals 1-5 ms after `filesystem.init-volume`. In every case the removal comes after the volume's own `creating-overlay-mount`, and neighbouring volumes mounted a few ms earlier or later are fine.

The ENOENT can be reproduced without Concourse:

```
mount -t overlay overlay -o lowerdir=$B/lower,upperdir=$B/upper,workdir=$B/work $B/merged
rm -rf $B/work
touch $B/merged/tmp/gdn-init   # No such file or directory, while ls $B/merged/tmp still works
```

Same result when upperdir and workdir are both removed. That is what runc hits: the init binary is a file, so runc creates the mountpoint with `O_CREAT|O_EXCL` inside the rootfs.

## Expected Results

A volume created or promoted while a sweep is running keeps its overlay dirs.

## Actual Results

Its layer or work dir is removed, the overlay is broken while mounted, and the container create fails.

## Worker(s) configuration

One worker, containerd runtime, overlay driver, `CONCOURSE_CONTAINERD_MAX_CONTAINERS=0`, default sweep interval of 30s, a few hundred containers at a time.

## Concourse Version

8.2.3. `volume/filesystem.go` and `volume/driver/overlay_linux.go` are identical in 8.2.4, 8.3.0 and master, so it should still be there.

## Did this use to work?

Yes. `RemoveOrphanedResources` does not exist in 8.0.0, it came in 8.1.0.

## Possible fix

Drop the snapshot and let the driver ask at removal time, e.g. pass `isKnown func(handle string) bool` instead of the map, which stats `init/<handle>` first and `live/<handle>` second. That order is safe against the rename: the entry is always in one of the two, and the `live` check happens after the `init` check, so it cannot miss both.

ShravanthReddy:

closes #9685

The overlay sweeper deleted layer/work dirs of volumes that were being created or promoted during the sweep. The volume stayed mounted, then runc hit ENOENT creating `/tmp/gdn-init` and the container create failed.

* Drop the `live/`+`init/` snapshot. `RemoveOrphanedResources` now takes a live `isKnown(handle)` callback.
* `CleanupOrphanedEntries` stats `init/` first, then `live/` at delete time, so an `init→live` rename cannot miss both, and a volume created after sweep start is still known.
* Fail closed: a `Stat` error other than `ErrNotExist` (permission, IO) is treated as known, so a transient stat failure cannot mass-delete overlays.
* Overlay test covers the real race: handle absent from a stale snapshot, present in `init/` via live `isKnown`. Work-only orphans are still deleted. `btrfs`/`naive` implement the new signature.

## Notes to reviewer

The in-flight overlay test used to pass the handle as known (`knownHandles(inFlight)`), which is the snapshot case, not the race. It now uses a live `Stat` of `init/` while the snapshot is empty.

To confirm:

```
go test ./worker/baggageclaim/volume/driver -ginkgo.focus "RemoveOrphanedResources" -count=1 -timeout 60s
go test ./worker/baggageclaim/volume -ginkgo.focus "CleanupOrphanedEntries" -count=1 -timeout 60s
```

## Release Note

* Fix worker overlay sweeper deleting overlay dirs of volumes being created, which caused runc to fail with ENOENT on `/tmp/gdn-init`.

### DIFF
--- a/worker/baggageclaim/volume/driver/overlay_linux.go
+++ b/worker/baggageclaim/volume/driver/overlay_linux.go
@@ -257,7 +257,7 @@ func (driver *OverlayDriver) workDir(vol volume.FilesystemVolume) string {
 
 func (driver *OverlayDriver) RemoveOrphanedResources(isKnown func(handle string) bool) error {
 	if isKnown == nil {
-		isKnown = func(string) bool { return false }
+		return errors.New("must pass in a function")
 	}
 
 	_, err := os.Stat(driver.OverlaysDir)
--- a/worker/baggageclaim/volume/driver/overlay_linux.go
+++ b/worker/baggageclaim/volume/driver/overlay_linux.go
@@ -255,7 +255,11 @@ func (driver *OverlayDriver) workDir(vol volume.FilesystemVolume) string {
 	return filepath.Join(driver.OverlaysDir, "work", vol.Handle())
 }
 
-func (driver *OverlayDriver) RemoveOrphanedResources(knownHandles map[string]struct{}) error {
+func (driver *OverlayDriver) RemoveOrphanedResources(isKnown func(handle string) bool) error {
+	if isKnown == nil {
+		isKnown = func(string) bool { return false }
+	}
+
 	_, err := os.Stat(driver.OverlaysDir)
 	if err != nil {
 		if errors.Is(err, fs.ErrNotExist) {
@@ -279,7 +283,7 @@ func (driver *OverlayDriver) RemoveOrphanedResources(knownHandles map[string]str
 			continue
 		}
 
-		if _, known := knownHandles[name]; known {
+		if isKnown(name) {
 			continue
 		}
 
@@ -307,7 +311,7 @@ func (driver *OverlayDriver) RemoveOrphanedResources(knownHandles map[string]str
 
 	for _, entry := range workEntries {
 		name := entry.Name()
-		if _, known := knownHandles[name]; known {
+		if isKnown(name) {
 			continue
 		}
 
--- a/worker/baggageclaim/volume/driver/overlay_linux_test.go
+++ b/worker/baggageclaim/volume/driver/overlay_linux_test.go
@@ -1,6 +1,7 @@
 package driver_test
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -12,6 +13,17 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 )
+
+func knownHandles(handles ...string) func(string) bool {
+	set := make(map[string]struct{}, len(handles))
+	for _, h := range handles {
+		set[h] = struct{}{}
+	}
+	return func(handle string) bool {
+		_, ok := set[handle]
+		return ok
+	}
+}
 
 var _ = Describe("Overlay", func() {
 	Context("Driver", func() {
@@ -135,12 +147,7 @@ var _ = Describe("Overlay", func() {
 			// orphan-vol-2 has no corresponding work dir
 			Expect(os.Mkdir(filepath.Join(overlaysDir, "orphan-vol-2"), 0755)).To(Succeed())
 
-			knownHandles := map[string]struct{}{
-				"known-vol-1": {},
-				"known-vol-2": {},
-			}
-
-			err := overlayDrv.RemoveOrphanedResources(knownHandles)
+			err := overlayDrv.RemoveOrphanedResources(knownHandles("known-vol-1", "known-vol-2"))
 			Expect(err).ToNot(HaveOccurred())
 
 			// Known handles should still exist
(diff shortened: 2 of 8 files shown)


taylorsilva:
Thanks for the PR! One of the baggageclaim unit tests is failing: https://ci.concourse-ci.org/teams/contributor/pipelines/pr/jobs/unit/builds/1?vars.number=%229695%22#L6a8ffa44:119:121

This is in my backlog to review.

worker/baggageclaim/volume/driver/overlay_linux_test.go
```diff
@@ -161,16 +168,53 @@ var _ = Describe("Overlay", func() {
...
+				}
+				return true
+			}
+
+			err := overlayDrv.RemoveOrphanedResources(isKnown)
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(workPath).To(BeADirectory())
+			Expect(orphanPath).ToNot(BeADirectory())
+		})
```
taylorsilva: Not sure this is actually testing anything. We should probably just test that we call the passed in func and the behaviour based on what that func returns.

worker/baggageclaim/volume/driver/overlay_linux.go
```diff
@@ -255,7 +255,11 @@ func (driver *OverlayDriver) workDir(vol volume.FilesystemVolume) string {
 	return filepath.Join(driver.OverlaysDir, "work", vol.Handle())
 }
 
-func (driver *OverlayDriver) RemoveOrphanedResources(knownHandles map[string]struct{}) error {
+func (driver *OverlayDriver) RemoveOrphanedResources(isKnown func(handle string) bool) error {
+	if isKnown == nil {
+		isKnown = func(string) bool { return false }
+	}
```
taylorsilva, line 261: Wouldn't this result in every volume being deleted? Should probably just return an error.

worker/baggageclaim/volume/driver/overlay_linux_test.go
taylorsilva: Actually there is no behaviour change at the driver-level. There's nothing really to test that isn't already covered by existing tests. I think this test can be deleted.

worker/baggageclaim/volume/filesystem_test.go
```diff
@@ -251,11 +251,52 @@ var _ = Describe("Filesystem", func() {
...
+			Expect(os.RemoveAll(initDir)).To(Succeed())
+			Expect(os.RemoveAll(liveDir)).To(Succeed())
+			Expect(os.WriteFile(initDir, nil, 0644)).To(Succeed())
+			Expect(os.WriteFile(liveDir, nil, 0644)).To(Succeed())
+			defer func() {
+				_ = os.Remove(initDir)
+				_ = os.Remove(liveDir)
+				_ = os.Mkdir(initDir, 0755)
+				_ = os.Mkdir(liveDir, 0755)
+			}()
```
taylorsilva: Not needed, the BeforeEach() that these files/dirs are made in will do the cleanup for us.

worker/baggageclaim/volume/filesystem_test.go
```diff
@@ -251,11 +251,52 @@ var _ = Describe("Filesystem", func() {
...
+			Expect(err).ToNot(HaveOccurred())
+
+			isKnown := driver.RemoveOrphanedResourcesArgsForCall(0)
+			Expect(isKnown("unknown-vol")).To(BeFalse())
+
+			initDir := filepath.Join(parentDir, "init")
+			liveDir := filepath.Join(parentDir, "live")
+			// chmod 0 is not a Stat error as root (CI unit-baggageclaim).
+			// Replace the dirs with files so Stat(init/any-handle) returns
+			// ENOTDIR, which is not ErrNotExist even as uid 0.
```
taylorsilva: Terribly confusing comment. More plain way of saying this:

```
removing the init/ and live/ dirs and replacing them with files. This will cause os.Stat to return ENOTDIR
```

taylorsilva (commented):
Thanks again for the PR! Overall I think this is the right direction to take this. Most of my feedback is regarding the tests.

I'm going to push a commit in a bit addressing my comments, assuming github doesn't block me, which it sometimes does. If it does block me I'll post my diff here for you to apply.

taylorsilva:
Pushed my commit successfully! I address all my comments in that commit. Biggest changes are with the tests. Original tests were mostly trying to test the right thing, they just had a few gaps and were overly verbose in a few spots.

Assuming the PR checks pass and you also don't see any further issues with this, I'll merge it.

taylorsilva (approved):

PR #9696, opened 2026-09-01
atc: add Google Secret Manager credential manager

sammirza-mo:

* **Adds a Google Secret Manager credential manager** (`gcpsecretmanager`) so `((vars))` can be resolved from GCP Secret Manager.
* **Authentication follows GCP conventions.** Credentials can be supplied via an inline service-account key (`--gcp-secretmanager-credentials-json`) or a key file (`--gcp-secretmanager-credentials-file`); when neither is set the manager
  falls back to Application Default Credentials / Workload Identity. 
* **Configurable secret-ID templates** for pipeline-, team-, and shared-scoped lookups. Because Secret Manager IDs cannot contain slashes, segments are delimited with a double hyphen by default (e.g. `concourse--{{.Team}}--{{.Pipeline}}--{{.Secret}}`). Templates are validated at config time to ensure they only expand to characters Secret Manager permits (`[A-Za-z0-9_-]`).
* **Always resolves the latest secret version** consistent with the other credential managers
* **Security-conscious lookups.** Every resolved secret ID is validated against the allowed character set *before* the resource name is built. Secret payloads are never logged (only the secret ID and project are), and the server-supplied CRC32C checksum is verified so a corrupted payload is caught.
* **JSON payload support.** A payload that parses as a JSON object is returned as a map so `((secret.field))` works; any other payload (scalars, arrays, plain strings) is returned verbatim as a string.
* **Health check + info integration.** Implements `Health()` (via `AccessSecretVersion` against a health-check secret ID) and a `MarshalJSON` that surfaces config/health to `/api/v1/info/creds` without ever serialising credentials. Configurable per-request `--gcp-secretmanager-request-timeout` (default `10s`).
* **Wiring:** registered the manager in `atc/atccmd/command.go` and `fly/commands/validate_pipeline.go`, and added `gcpsecretmanager` to the list of managers allowed as a pipeline `var_source` in `atc/configvalidate/validate.go`.
* **Dependencies:** added `cloud.google.com/go/secretmanager` and promoted a few related transitive deps (`gax-go`, `google.golang.org/api`, `iam`) in `go.mod`/`go.sum`.
* **Local dev tooling:** added `hack/overrides/gcp-secretmanager.yml` (docker compose override wiring ADC into the `web` node) plus `hack/gcp-secretmanager/` scripts and sample pipelines (`seed-secrets.sh`, `destroy-secrets.sh`, `pipeline.yml`, `pipeline-var-source.yml`) for exercising the manager end-to-end.

## Notes to reviewer

* **Test coverage:** unit tests are included using a counterfeiter fake for the Secret Manager API — `gcpsecretmanager_test.go` (lookup, JSON vs string payloads, NotFound/FailedPrecondition fall-through, CRC mismatch, secret-ID validation) and `manager_test.go` (config validation, template validation, credential-source handling). No `testflight`/`topgun` changes, consistent with how the other cloud creds managers are tested.
* **Docs:** per CONTRIBUTING, user-facing documentation should land as a separate PR against `concourse/docs`. 
* **Manual verification:** the manager was exercised locally via `hack/overrides/gcp-secretmanager.yml` against a real GCP sandbox project using ADC, seeding/destroying secrets with the scripts under `hack/gcp-secretmanager/` and resolving them through both cluster-wide config and a pipeline `var_source`.

## **Licensing** 
_This contribution is made by employees of Mechanical Orchard, Inc. under the terms of the project’s license._


kcbimonte:
I'm working on getting a documentation page to describe this new Secrets Manager functionality. 

Just had one question though. What's the use case of having a secret version flag? Seems to be new across the existing secrets manager support. And that it wouldn't allow secrets to be independently updated without operator input. 

Thinking along the use case of having a mix of secrets that are auto rotated and some that are constants. If the auto rotate runs, then you'd get version 2 of those but still be on version 1 of the constants. 

Now, I've never used GCP, so I'm probably missing something in it's secrets manager support that would've prompted this version flag.

sammirza-mo:
Thanks @kcbimonte -- this is a good catch and thank you for picking up the docs page.

Honestly, I included it just because the GCP API exposes a version parameter, not because there was a real need; I hadn't checked it against the other managers. Also, I didn't catch that the `--gcp-secretmanager-secret-version` flag is global, not per-secret. Thanks for pointing that out. 

I've cross-checked against the existing credential managers. Like you said, they resolve to the latest value with no operator-facing version selector.

So I'll remove the secret-version flag and always resolve `latest`. That (a) brings this in line with every other manager and (b) removes the confusing all-or-nothing behaviour. 

I'll push an update shortly.

taylorsilva:
Thanks for the PR! This is in my PR backlog to review, but will probably be 1-2 weeks until I get to it.

There is a merge conflict from some dependency updates that just merged FYI.

Based on the scope of this PR, you can ignore the `containerd-integration` check if it fails.

sammirza-mo:
Thanks @taylorsilva! I have resolved the merge conflict

PR #9712, opened 2026-09-10
fix: race condition during web restart + integration test

### LINKED ISSUE #9711: Race condition in HA: concurrent ATC instances attempt to create the same container, causing "already exists" error and silent duplicate task execution
### Summary

In an HA deployment with multiple ATC (web) instances, a rolling restart can cause two ATCs to concurrently attempt to create the same container on a worker. The second attempt fails with `new container: container "X": already exists`, erroring the build and triggering a retry — but the original container continues running silently on the worker, causing downstream side effects (e.g. conflicting Terraform state on the retry).

Image

### Steps to Reproduce

1. Run Concourse with two web (ATC) instances sharing the same PostgreSQL database
2. Trigger a build that runs a task step
3. While the build is in the container-creation phase (between `FindContainer` returning `state=creating` and `creatingContainer.Created()` being called), perform a rolling restart — stop web1, allowing web2 to acquire the build tracking advisory lock
4. After web2 picks up the build and enters `createGardenContainer`, restart web1
5. Web1 re-acquires the build tracking lock, calls `Lookup` (returns nil — container not yet in containerd), and also enters `createGardenContainer`
6. Both ATCs call `gardenClient.Create` with the same container handle concurrently

### Expected Results

Only one ATC creates the container. The other either attaches to the existing container or waits. The build completes once without duplicate execution.

### Actual Results

```
atc.tracker.tick.run.failed-to-create-container-in-garden
  error: "new container: container \"4330d7d3-941c-4ca1-8dea-5b8175c7284a\": already exists"

atc.tracker.tick.run.errored
  error: "run check: find or create container on worker db634eed7c74:
          new container: container \"4330d7d3-941c-4ca1-8dea-5b8175c7284a\": already exists"
```
The build is marked as errored and a retry is scheduled on a different worker. The original container continues executing on the first worker unobserved, causing side effects (e.g. Terraform state lock conflicts, duplicate deployments).

### Web Node(s) configuration

1. The build tracking advisory lock (`pg_try_advisory_lock(1, buildID)`) is session-level and releases when the ATC's PostgreSQL connection is closed — this occurs during a rolling restart or when the connection pool recycles an idle connection
2. A second ATC acquires the now-free lock and calls `indOrCreateContainer`
3. The DB shows `state=creating`; `gardenClient.Lookup` returns nil (container not yet in containerd)
4. Both ATCs proceed into `createGardenContainer` → `gardenClient.Create`
5. The Garden backend's `createLock` serializes the two HTTP requests — first succeeds, second calls `containerd.NewContainer` and receives `ErrAlreadyExists`
6. The error propagates as `"new container: container X: already exists"` with no recovery path — `markContainerAsFailed` is called and the build errors

Local changes for easier reproduce:
`docker-compose.yml` — add a second web service:
```
web2:
  # copy of web service
  ports: [8081:8080]
  environment:
    CONCOURSE_PEER_ADDRESS: web2
    # ... same as web
```
Worker `CONCOURSE_TSA_HOST: web:2222,web2:2222`

`atc/worker/gardenruntime/worker.go` — sleep before `createGardenContainer` to widen the race window:
```
if gardenContainer == nil {
    time.Sleep(60 * time.Second) // gives time to stop/start web instances
    gardenContainer, err = worker.createGardenContainer(...)
}
```

`worker/runtime/backend.go` — sleep inside `createContainer` to hold `createLock` while the second ATC queues its request:
```
// before b.client.NewContainer(...)
time.Sleep(60 * time.Second)
return b.client.NewContainer(ctx, gdnSpec.Handle, labels, oci)
```

### Worker(s) configuration

_No response_

### Concourse Version

8.3.0

### Browser (if applicable)

_No response_

### Did this use to work?

_No response_

Kump3r:
# Changes proposed by this PR

closes #9711

## What

When two ATC instances both try to create the same container at the same time, one of them would fail with "handle already exists". This left the build in a broken state even though the container was successfully created by the other instance.

## Why

In an HA setup, both ATCs can see a container marked as `creating` in the database and both try to create it in Garden. The second one loses the race and gets an error. Instead of treating this as a real failure, we now check if the container already exists in Garden before giving up — if it does, we just use it.

## How

In `createGardenContainer`, after a failed `Create` call, we attempt a `Lookup` on the same handle. If the container is there, we return it. If not, we fall back to the original error path.

A new test covers the concurrent case: two goroutines both race to create a container with the same handle and both must succeed without returning an "already exists" error.

## Test helper changes

To make the race deterministic, the test needs to pause both goroutines inside `Garden.Create` at the same time before either one proceeds. This required two small additions to the test double:

- `BeforeCreate` hook on `grt.Garden` — called just before the lock is acquired, giving the test a place to synchronise both goroutines
- `WithGarden` on `grt.Worker` — the only way to hand a pre-configured `*Garden` to the worker is before `Build()` is called, since `scenario.Worker()` returns a plain interface with no path back to the underlying garden instance

Both follow the same pattern already used for `WithGardenContainers` in the same file.

## Test

- Test fails without the fix (one goroutine gets "already exists")
- Test passes with the fix (both goroutines succeed)

### DIFF
--- a/atc/worker/gardenruntime/gardenruntimetest/garden.go
+++ b/atc/worker/gardenruntime/gardenruntimetest/garden.go
@@ -16,6 +16,7 @@ import (
 
 type Garden struct {
 	ContainerList  []*Container
+	BeforeCreate   func(spec garden.ContainerSpec)
 	containersLock sync.Mutex
 }
 
@@ -55,14 +56,19 @@ func (g *Garden) Create(spec garden.ContainerSpec) (gclient.Container, error) {
 	if handle == "fail-to-create" {
 		return nil, errors.New("failed to create (because handle is fail-to-create)")
 	}
-	if _, _, ok := g.FindContainer(handle); ok {
-		return nil, fmt.Errorf("handle %s already exists", handle)
-	}
 
-	container := NewContainer(handle).WithSpec(spec)
+	if g.BeforeCreate != nil {
+		g.BeforeCreate(spec)
+	}
 
 	g.containersLock.Lock()
 	defer g.containersLock.Unlock()
+	for _, c := range g.ContainerList {
+		if c.handle == handle {
+			return nil, fmt.Errorf("handle %s already exists", handle)
+		}
+	}
+	container := NewContainer(handle).WithSpec(spec)
 	g.ContainerList = append(g.ContainerList, container)
 	return container, nil
 }
--- a/atc/worker/gardenruntime/gardenruntimetest/worker.go
+++ b/atc/worker/gardenruntime/gardenruntimetest/worker.go
@@ -30,6 +30,7 @@ type WorkerSetupFunc func(*atc.Worker)
 type Worker struct {
 	WorkerName       string
 	Containers       []*Container
+	GardenClient     *Garden
 	Volumes          []*Volume
 	SetupFuncs       []SetupFunc
 	WorkerSetupFuncs []WorkerSetupFunc
@@ -60,9 +61,13 @@ func (w *Worker) Setup(s *workertest.Scenario) {
 }
 
 func (w Worker) Build(db worker.DB, dbWorker db.Worker) runtime.Worker {
+	gdn := w.GardenClient
+	if gdn == nil {
+		gdn = &Garden{ContainerList: w.Containers}
+	}
 	return gardenruntime.NewWorker(
 		dbWorker,
-		&Garden{ContainerList: w.Containers},
+		gdn,
 		&Baggageclaim{Volumes: w.Volumes, Mutex: sync.Mutex{}},
 		db.ToGardenRuntimeDB(),
 		worker.NewStreamer(db.ResourceCacheFactory, compression.NewGzipCompression(), 0, worker.P2PConfig{
@@ -76,6 +81,12 @@ func (w Worker) WithGardenContainers(containers ...*Container) *Worker {
 	w2.Containers = make([]*Container, len(w.Containers)+len(containers))
 	copy(w2.Containers, w.Containers)
 	copy(w2.Containers[len(w.Containers):], containers)
+	return &w2
+}
+
+func (w Worker) WithGarden(g *Garden) *Worker {
+	w2 := w
+	w2.GardenClient = g
 	return &w2
 }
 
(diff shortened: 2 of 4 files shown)


Kump3r:
The fix here handles the behaviour when the same worker is reselected, which based on my testing is the 99% of the time. I still think there could be potential double execution, but the side case is rather rare. Based on my testing and going trough the code, we would need to hit this exact same scenario and during that time the originally selected worker needs to change to failing/unresponsive, causing the scheduler to select a new worker (which will start the process), but the original worker (failing) one could continue to run the process on the container in the VM (potentially causing git conflicts, terraform state mess etc.). This would only be possible if the workers are in a bad (overloaded) state and rolling restart of the web is triggered. Sadly this side case might not be that simple to handle and might require a bit of "over-engineering" for a rare side case, as I assume during overloaded workers we will notice a whole bunch of other unexpected side effects.

Kump3r:
Since that behaviour slightly diverges from the original issue (aka. "container already exists", without retrying to join it), I think that could/should be a separate issue. @taylorsilva I would love to hear your thoughts if we should consider this based on the explained details above. Without writing anything I am thinking of a simple proposition/implementation, something like when a worker begins executing a step, it acquires a PostgreSQL advisory lock scoped to that step. Because these locks are tied to the database connection, they are automatically released if the worker loses connectivity - at which point the new worker can safely take over. If two workers ever try to run the same step at the same time, the second one will detect the lock and fail fast rather than proceeding.
I do not think it is worth blocking this, as it is more or less the more "common" scenario, just wanted to make things as robust as possible.

taylorsilva:
Hey, I'll review this a bit deeper next week. I just read the first line of the PR description and something feels off:

> When two ATC instances both try to create the same container at the same time, one of them would fail with "handle already exists".

Based on my knowledge, this should not be possible. Only one web node can execute a build at a time (we use db locks to ensure this), therefore two web nodes should _never_ be racing to create a container because they can't **both** have a lock on the build.

I'll read the issue and your comments more at the start of next week when my IRL stuff is over (family wedding).

Kump3r:
> Based on my knowledge, this should not be possible. Only one web node can execute a build at a time (we use db locks to ensure this), therefore two web nodes should _never_ be racing to create a container because they can't **both** have a lock on the build.

No worries and no rush. Exactly, I thought so too, but as it turns out during a rolling update the first web releases the lock, however the container is running on the work and we hit the container already exists error. It is “tricky” to reproduce, but sadly it occurs in production during redeploy of webs for example.

PR #9714, opened 2026-09-10
fix: Vault URL validation scheme broken with placeholder

### LINKED ISSUE #9706: regression: var_source validation scheme breaks when using ((some_var.containing_scheme))
### Summary

https://github.com/concourse/concourse/commit/d7ce1bebd2c3e21a8c39b535ab6ebc75afb0368e

seems to have broken setting pipelines containing a var_source that has it's url set via a ((var)).

the actual value behind the var is prefixed with `https://`

### Steps to Reproduce

produce a pipeline with this diff:

```
variable source:
  variable source secrets has changed:
  config:
    auth_backend: userpass
    auth_params:
      password: ((v.password))
      username: ((v.username))
    path_prefix: ((v.mount_point))
-   url: https://((v.vault_addr))
+   url: ((v.vault_addr))
  name: secrets
  type: vault

pipeline name: pipeline

apply configuration? [yN]: y
error: invalid pipeline config:
invalid variable sources:
        credential manager secrets is invalid: invalid scheme "": must be http or https
```

### Expected Results

_No response_

### Actual Results

_No response_

### Web Node(s) configuration

_No response_

### Worker(s) configuration

_No response_

### Concourse Version

8.3.0

### Browser (if applicable)

_No response_

### Did this use to work?

yes. 8.2.4

Kump3r:

closes #9706

Tested on local Concourse with example pipeline:
```
var_sources:
- name: secrets
  type: vault
  config:
    url: ((v.vault_addr))
    auth_backend: userpass
    auth_params:
      username: ((v.username))
      password: ((v.password))
    path_prefix: /concourse

jobs:
- name: hello
  plan:
  - task: say-hello
    config:
      platform: linux
      image_resource:
        type: registry-image
        source: {repository: alpine}
      run:
        path: echo
        args: ["hello world"]
```

### DIFF
--- a/atc/creds/vault/manager.go
+++ b/atc/creds/vault/manager.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/url"
 	"path"
+	"strings"
 	"time"
 
 	"code.cloudfoundry.org/lager/v3"
@@ -144,14 +145,16 @@ func (manager VaultManager) Validate() error {
 		return fmt.Errorf("invalid URL: %s", err)
 	}
 
-	switch u.Scheme {
-	case "http", "https":
-	default:
-		return fmt.Errorf("invalid scheme %q: must be http or https", u.Scheme)
-	}
+	if !strings.Contains(manager.URL, "((") {
+		switch u.Scheme {
+		case "http", "https":
+		default:
+			return fmt.Errorf("invalid scheme %q: must be http or https", u.Scheme)
+		}
 
-	if u.Hostname() == "" {
-		return fmt.Errorf("URL must specify a host")
+		if u.Hostname() == "" {
+			return fmt.Errorf("URL must specify a host")
+		}
 	}
 
 	if manager.PathPrefix == "" {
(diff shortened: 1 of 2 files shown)


taylorsilva (approved):
I think this is good enough. It means someone could pass in a bad URL, but this really is just a "best-effort" check we're doing here.

Kump3r:
Yeah, I agree. Again the problem with validations when there are placeholders arises. I can try to think of something, but as the variables are resolved later during runtime, we either need to accept this risk and not break it, or force operators to placeholder only the interchangeable part of the URL, without the prefix. I guess if we opt for that we would need to note it down explicitly.

Edit: I would assume the Vault library will panick further down the line and raise an Exception as well, it would have been just nicer to handle it before creating the pipeline, but afaik the vars will be interpolated later during a build.

Edit-2: Yup, with invalid URL in placeholder users will see:
```
"invalid/v1/sys/internal/ui/mounts/namespace/path/to/secret/test": unsupported protocol scheme ""
```

PR #9723, opened 2026-09-17
fix: baggageclaim: fix overlay driver stacking mounts on worker restart

### LINKED ISSUE #9722: baggageclaim: overlay driver Recover() stacks bind-mounts on each in-place restart → ENOSPC crash loop
### Summary

On workers using the overlay baggageclaim driver, every in-place process restart
(e.g. BOSH in-place update, monit restart) adds a full new layer of bind/overlay
mounts on top of the previous run's mounts without checking whether they already
exist. Over N restarts with V on-disk volumes, the mount table grows to N × V
entries. Once it hits the kernel per-mount-namespace limit the next mount(2) call
returns ENOSPC, Recover() aborts, and the worker crash-loops under monit, which
restarts it again and stacks another layer. Self-reinforcing.

Related: #7430 ("worker does not recover if volume is still mounted")

### Steps to Reproduce

1. Run a Concourse worker with --baggageclaim-driver=overlay.
2. Execute several pipelines to create volumes under $CONCOURSE_WORK_DIR/volumes/live/.
3. Without destroying the VM, restart only the worker process: sudo /var/vcap/bosh/bin/monit restart worker
Repeat step 3 five or more times.
4. Observe: mount | grep "$CONCOURSE_WORK_DIR" | wc -l Each restart multiplies the count by (1 + previous).
5. Continue restarting until ENOSPC is returned; worker enters crash loop.

### Expected Results

_No response_

### Actual Results

Example:
```bash
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/bosh_ssh/bosh_2ab78393b77346a# mount | grep "$CONCOURSE_WORK_DIR" | wc -l
182
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/bosh_ssh/bosh_2ab78393b77346a# monit restart worker
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/bosh_ssh/bosh_2ab78393b77346a# mount | grep "$CONCOURSE_WORK_DIR" | wc -l
386
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/bosh_ssh/bosh_2ab78393b77346a# monit restart worker
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/bosh_ssh/bosh_2ab78393b77346a# mount | grep "$CONCOURSE_WORK_DIR" | wc -l
758
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/bosh_ssh/bosh_2ab78393b77346a# monit restart worker
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/bosh_ssh/bosh_2ab78393b77346a# mount | grep "$CONCOURSE_WORK_DIR" | wc -l
```

### Web Node(s) configuration

_No response_

### Worker(s) configuration

CONCOURSE_BAGGAGECLAIM_DRIVER=overlay

### Concourse Version

8.3.0

### Browser (if applicable)

_No response_

### Did this use to work?

_No response_

Kump3r:

closes #9722

### What

`OverlayDriver.Recover()` called `bindMount`/`overlayMount` for every
on-disk volume on startup without checking whether those paths were
already mounted. Linux allows stacking bind mounts silently, so each
in-place process restart (BOSH update, monit, etc.) added a full extra
copy of all volume mounts. After enough restarts the mount table hit
its kernel limit, `mount(2)` returned ENOSPC, and the worker entered a
crash loop that made things worse on every iteration.

### How

- `bindMount` and `overlayMount` now check `mountinfo.Mounted()` before
  calling `syscall.Mount` and skip silently if already present.
- `Recover()` tracks which paths it mounts in the current pass and
  unmounts them if any later step fails, preventing partial mount state
  from compounding on the next restart.
- `moby/sys/mountinfo` was already an indirect dependency; promoted to
  direct.

### Tests

- Unit: `Recover()` called N times on already-mounted volumes — mount
  count stays flat.
- Unit: `Recover()` with a bad volume mid-loop — previously mounted
  volumes are rolled back.
- Integration: worker bounced without prior unmount (dirty restart) —
  volumes recover correctly and mount count does not grow across
  repeated bounces.


kristianstanchev:
Example error message in production: 
```
{"timestamp":"2026-09-17T11:27:05.773012935Z","level":"error","source":"worker","message":"worker.baggageclaim.failed-to-recover-volume-driver","data":{"error":"recover bind mount: no space left on device","session":"2"}}
{"timestamp":"2026-09-17T11:27:26.028301997Z","level":"error","source":"worker","message":"worker.baggageclaim.failed-to-recover-volume-driver","data":{"error":"recover bind mount: no space left on device","session":"2"}}
{"timestamp":"2026-09-17T11:27:46.180025145Z","level":"error","source":"worker","message":"worker.baggageclaim.failed-to-recover-volume-driver","data":{"error":"recover bind mount: no space left on device","session":"2"}}
{"timestamp":"2026-09-17T11:28:06.427830771Z","level":"error","source":"worker","message":"worker.baggageclaim.failed-to-recover-volume-driver","data":{"error":"recover bind mount: no space left on device","session":"2"}}
```

worker/baggageclaim/volume/driver/overlay_linux.go
```diff
@@ -123,10 +126,22 @@ func (driver *OverlayDriver) Recover(fs volume.Filesystem) error {
...
+	// can be rolled back if a later step fails.
+	var mountedThisPass []string
+	rollback := func() {
+		for _, path := range mountedThisPass {
+			if err := syscall.Unmount(path, 0); err != nil && !errors.Is(err, syscall.EINVAL) {
+				driver.logger.Error("rollback-unmount", err, lager.Data{"path": path})
+			}
+		}
+	}
+
```
taylorsilva: I'm not sure about the utility of doing this `rollback`. It assumes the state prior to `Recover()` being called is a better place to leave things in if there's a failure of any kind. Without knowing what error we're dealing with this may or may not be useful. **Why is unmounting what we've successfully mounted so far a good thing to do?**

Another question: why do we ignore `EINVAL` errors ("invalid argument") during rollback?

---

I think it would make sense to do this if we know why. I see in the issue you wrote:

> Defect 2 - Recover() does not roll back mounts on error (line 115–162)

Why is this a defect?

taylorsilva: This `rollback` function also seems to be the entire reason for adding the boolean return value to `bindMount()` and `overlayMount()`.

worker/baggageclaim/integration/baggageclaim/overlay_mounts_test.go
```diff
@@ -124,4 +125,61 @@ var _ = Describe("baggageclaim restart", func() {
...
+			Expect(dataExistsInVolume(dataFile, rootVolume.Path())).To(BeTrue())
+
+			cowVolume, err = client.CreateVolume(ctx, "cow-handle", baggageclaim.VolumeSpec{
+				Strategy:   baggageclaim.COWStrategy{Parent: rootVolume},
+				Properties: map[string]string{},
+			})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(dataExistsInVolume(dataFile, cowVolume.Path())).To(BeTrue())
+
+			runner.Bounce()
```
taylorsilva: Why do we bounce the runner in the `BeforeEach()`? If we want to test multiple restarts we could do a `for` loop in the `It()` instead where we restart the runner multiple times.

worker/baggageclaim/integration/baggageclaim/overlay_mounts_test.go
```diff
@@ -124,4 +125,61 @@ var _ = Describe("baggageclaim restart", func() {
...
+		})
+
+		AfterEach(func() {
+			_ = syscall.Unmount(rootVolume.Path(), 0)
+			_ = syscall.Unmount(cowVolume.Path(), 0)
```
taylorsilva: nit: why ignore errors? We check for errors in the other tests in this file.

worker/baggageclaim/volume/driver/overlay_linux.go
taylorsilva: If we end up keeping this, we should also rollback in the reverse order of the list to correctly rollback any parent->child relationships. Current setup would rollback a parent before a child.

Kump3r: Let me provide some clarity although I do agree with the challenge and you are right to question, also the reversed order is an actual bug. 

To the point why do it in the first place - The rollback was added as defense-in-depth for the case where Recover() mounts volumes 1..K successfully and then fails on K+1, leaving a partial state. The theory was: better to clean up what we added than leave half-mounted volumes.vBut given that the primary fix is the idempotency check (`mountinfo.Mounted()` before every syscall.Mount), partial mounts from a failed pass are already harmless on the next restart — the idempotency check will skip them. So rollback doesn't actually add safety, just complexity. Removing it now.

The bool return only exists for rollback - Yup, without rollback there's no reason to distinguish "we mounted it this pass" from "it was already mounted". Both `bindMount` and `overlayMount` go back to returning plain `error`.

Rollback order is wrong (parent before child) - Yes, I missed this and its a real bug: the list was ordered roots first then CoWs, so rollback would try to unmount a root bind-mount while its CoW overlay was still live - that would fail with EBUSY. The fix would be to reverse the list before iterating. But since we're dropping rollback entirely, this doesn't need to be fixed - it goes away.

worker/baggageclaim/integration/baggageclaim/overlay_mounts_test.go
Kump3r: Valid point - the test is clearer if the full scenario (create → bounce → assert → bounce again → assert) lives inside the `It` block rather than split across BeforeEach and It. Both bounces are now inside the It, so the reader can follow the whole story without jumping between blocks.

worker/baggageclaim/integration/baggageclaim/overlay_mounts_test.go
Kump3r: I used `_ =` here because AfterEach runs even when the test body fails, and
at that point I wasn't certain whether the mounts would still be present -
if the test failed mid-bounce, cleanup might be in an unknown state and a
hard `Expect` would produce a noisy secondary failure that masks the real one.
That said, the test is structured so both volumes are always mounted by the
time AfterEach runs, so the uncertainty doesn't actually exist. Fixed to use
`Expect(err).NotTo(HaveOccurred())`

Kump3r:
Validated the fix by compiling the binary and switching to the bosh VM, with the same steps described in the issue:
```
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/packages/concourse/bin# mount | grep "$CONCOURSE_WORK_DIR" | wc -l
225
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/packages/concourse/bin# monit restart worker
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/packages/concourse/bin# mount | grep "$CONCOURSE_WORK_DIR" | wc -l
229
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/packages/concourse/bin# mount | grep "$CONCOURSE_WORK_DIR" | wc -l
232
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/packages/concourse/bin# monit restart worker
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/packages/concourse/bin# mount | grep "$CONCOURSE_WORK_DIR" | wc -l
238
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/packages/concourse/bin# monit restart worker
worker/fa3d0fb7-3650-4e4c-a375-830664466ce8:/var/vcap/packages/concourse/bin# mount | grep "$CONCOURSE_WORK_DIR" | wc -l
238
```
As you can see with the new binary the cound is not multiplied on each restart

taylorsilva (approved):
Thanks for making the changes Kalin! LGTM now. I'll make sure this makes it over to 8.3.1

Kump3r:
Thanks for the quick response as always and for taking a look

